PR #4186, opened 2024-05-18
Set default value of `raw.theme` to `auto`, and allow setting `raw.theme` to `auto`

### LINKED ISSUE #4182: Cannot set raw theme to `none`
### Description

image

The document says `raw.theme` can be `none` or `str`, but when I compile the following code:

```typst
#set raw(theme: none)
```

it emits the error:

```plaintext
error: expected string, found none
```

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

Coekjan:
This pr closes #4182 .

I found out the bug in `parse_theme`. The function `parse_theme` always wants to parse the argument `theme` into `EcoString`, which raises error when meeting `none`.

As https://github.com/typst/typst/pull/4186#pullrequestreview-2071726156 , we set default value of `raw.theme` to `auto`, and allow setting `raw.theme` to `auto`. To implement it, I redesigned the function signature (return value type):
1. `Err(...)` indicates argument `theme` found but error met when casting it into `Smart<EcoString>`;
2. `Ok((None, None))` indicates argument `theme` not found;
3. `Ok((Some(Smart::Auto), None))` indicates argument `theme` found, and it is `auto`;
4. `Ok((Some(Smart::Custom(path)), Some(data)))` indicates argument `theme` found, and it is `str` (path to the theme file).


laurmaedje (commented):
I think `theme: none` doesn't really make sense like this, since it's still a theme, just the default one. We should probably switch the default to `auto` instead.

We could also support `theme: none` to disable highlighting. Not sure whether we need that.

Coekjan:
> I think `theme: none` doesn't really make sense like this, since it's still a theme, just the default one. We should probably switch the default to `auto` instead.

Agree.

> 
> We could also support `theme: none` to disable highlighting. Not sure whether we need that.

Maybe, in this PR, we can switch to `auto`, and leave whether we need `none` for future discussion.

laurmaedje:
Thanks!

PR #4190, opened 2024-05-19
Do not show tracepoints outside of root by default

MDLC01:
This PR changes the default way tracebacks are displayed to exclude tracepoints that are within a package. This can be reverted to show full tracebacks using by passing `--diagnostic-location source` to the CLI.

The goal is to reduce the noise for the end user when compiling a Typst file via the CLI.

I'm not sure if "diagnostic location" is the right name for that. I'm not sure either how to properly test that.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -209,6 +209,14 @@ pub struct SharedArgs {
         value_parser = clap::value_parser!(DiagnosticFormat)
     )]
     pub diagnostic_format: DiagnosticFormat,
+
+    /// Where diagnostics should be reported
+    #[clap(
+        long,
+        default_value_t = DiagnosticLocation::WithinRoot,
+        value_parser = clap::value_parser!(DiagnosticLocation)
+    )]
+    pub diagnostic_location: DiagnosticLocation,
 }
 
 /// Parses a UNIX timestamp according to <https://reproducible-builds.org/specs/source-date-epoch/>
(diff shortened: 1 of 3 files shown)


laurmaedje:
Can you show an example of a before/after? Without that, it's hard to judge if/why this is desirable.

MDLC01:
Consider the following file.

```typ
#import "@preview/example:0.1.0": div
// `div` just divides its two arguments, which is not possible for strings.
#div("this", "errors")
```

Full error message with `--diagnostic-location source` (same as before the PR):
```
error: cannot divide string by string
  ┌─ @preview/example:0.1.0\util\math.typ:4:17
  │
4 │ #let div(x, y) = x / y
  │                  ^^^^^

help: error occurred in this call of function `div`
  ┌─ \\?\D:\typst\test.typ:3:1
  │
3 │ #div("this", "errors")
  │  ^^^^^^^^^^^^^^^^^^^^^
```

Full error message without `--diagnostic-location source`:
```
error: cannot divide string by string
  ┌─ \\?\D:\typst\test.typ:3:1
  │
3 │ #div("this", "errors")
  │  ^^^^^^^^^^^^^^^^^^^^^
```

PgBiel:
I have a feeling this makes the error less intuitive, at least if this is the default behavior. I could see a new user being confused by this error (whereas it could be the package author's fault instead of the user's fault).

Maybe a way to indicate the tracepoint was suppressed would be good.

laurmaedje:
I think the same reasoning can apply if the panic is in a utility file in your project. While the package boundary is common, it's not necessarily the clear-cut boundary.

PR #4196, opened 2024-05-20
`layout` documentation improvements

MDLC01:
I rephrased some parts and added a mention that `layout` provides context.

Sorry for the slight spam of documentation PRs from me.

### DIFF
--- a/crates/typst/src/layout/layout.rs
+++ b/crates/typst/src/layout/layout.rs
@@ -60,7 +60,7 @@ pub fn layout(
     ///
     /// This function is called once for each time the content returned by
     /// `layout` appears in the document. This makes it possible to generate
-    /// content that depends on the size of its container.
+    /// content that depends on the dimensions of its container.
     func: Func,
 ) -> Content {
     LayoutElem::new(func).pack().spanned(span)
--- a/crates/typst/src/layout/layout.rs
+++ b/crates/typst/src/layout/layout.rs
@@ -27,7 +27,7 @@ use crate::syntax::Span;
 /// ])
 /// ```
 ///
-/// If the `layout` call is placed inside a box width a width of `{800pt}` and a
+/// If the `layout` call is placed inside a box with a width of `{800pt}` and a
 /// height of `{400pt}`, then the specified function will be given the parameter
 /// `{(width: 800pt, height: 400pt)}`. If it placed directly into the page, it
 /// receives the page's dimensions minus its margins. This is mostly useful in
--- a/crates/typst/src/layout/layout.rs
+++ b/crates/typst/src/layout/layout.rs
@@ -29,7 +29,7 @@ use crate::syntax::Span;
 ///
 /// If the `layout` call is placed inside a box with a width of `{800pt}` and a
 /// height of `{400pt}`, then the specified function will be given the parameter
-/// `{(width: 800pt, height: 400pt)}`. If it placed directly into the page, it
+/// `{(width: 800pt, height: 400pt)}`. If it is placed directly into the page, it
 /// receives the page's dimensions minus its margins. This is mostly useful in
 /// combination with [measurement]($measure).
 ///
--- a/crates/typst/src/layout/layout.rs
+++ b/crates/typst/src/layout/layout.rs
@@ -28,7 +28,7 @@ use crate::syntax::Span;
 /// ```
 ///
 /// If the `layout` call is placed inside a box with a width of `{800pt}` and a
-/// height of `{400pt}`, then the specified function will be given the parameter
+/// height of `{400pt}`, then the specified function will be given the argument
 /// `{(width: 800pt, height: 400pt)}`. If it is placed directly into the page, it
 /// receives the page's dimensions minus its margins. This is mostly useful in
 /// combination with [measurement]($measure).
--- a/crates/typst/src/layout/layout.rs
+++ b/crates/typst/src/layout/layout.rs
@@ -12,8 +12,11 @@ use crate::syntax::Span;
 /// Provides access to the current outer container's (or page's, if none)
 /// dimensions (width and height).
 ///
-/// Accepts a function that receives a single parameter, which is a
-/// dictionary with keys `width` and `height`, both of type [`length`].
+/// Accepts a function that receives a single parameter, which is a dictionary
+/// with keys `width` and `height`, both of type [`length`]. The function is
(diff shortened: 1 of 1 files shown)


crates/typst/src/layout/layout.rs
```diff
@@ -27,11 +27,11 @@ use crate::syntax::Span;
 /// ])
 /// ```
 ///
-/// If the `layout` call is placed inside of a box width a width of `{800pt}`
-/// and a height of `{400pt}`, then the specified function will be given the
-/// parameter `{(width: 800pt, height: 400pt)}`. If it placed directly into the
-/// page it receives the page's dimensions minus its margins. This is mostly
-/// useful in combination with [measurement]($measure).
+/// If the `layout` call is placed inside a box width a width of `{800pt}` and a
```
EpicEricEE, line 30:
```suggestion
/// If the `layout` call is placed inside a box with a width of `{800pt}` and a
```
I know you didn't change it, but might as well fix it while you're at it :)

crates/typst/src/layout/layout.rs
```diff
@@ -27,11 +27,11 @@ use crate::syntax::Span;
...
+/// height of `{400pt}`, then the specified function will be given the parameter
+/// `{(width: 800pt, height: 400pt)}`. If it placed directly into the page, it
```
EpicEricEE, line 32:
```suggestion
/// `{(width: 800pt, height: 400pt)}`. If it is placed directly into the page, it
```
Or "If placed..."

crates/typst/src/layout/layout.rs
```diff
@@ -56,8 +59,8 @@ pub fn layout(
     /// and `height`.
     ///
     /// This function is called once for each time the content returned by
-    /// `layout` appears in the document. That makes it possible to generate
-    /// content that depends on the size of the container it is inside of.
+    /// `layout` appears in the document. This makes it possible to generate
+    /// content that depends on the size of its container.
```
EpicEricEE, line 63:
```suggestion
    /// content that depends on the dimensions of its container.
```
The other mentions of "size" have also been replaced, so this one should probably too.

crates/typst/src/layout/layout.rs
MDLC01: Good catch!

crates/typst/src/layout/layout.rs
```diff
@@ -43,8 +43,11 @@ use crate::syntax::Span;
 /// })
 /// ```
 ///
-/// Note that this function will provide an infinite width or height if one of
-/// the page width or height is `auto`, respectively.
+/// Note that the provided width or height will be infinite in case the
+/// corresponding page dimension is set to `{auto}`.
```
PgBiel, line 47: I think this might be a slight bit better (not 100% sure):

```suggestion
/// Note that the width or height provided by `layout` will be infinite if the
/// corresponding page dimension is set to `{auto}`.
```

PgBiel: Could also be worthwhile to add an example of this

crates/typst/src/layout/layout.rs
```diff
@@ -43,8 +43,11 @@ use crate::syntax::Span;
...
+///
+/// Moreover, this function also provides [context]($context) to its argument,
+/// so you don't need to use it in combination with the `context` keyword.
```
PgBiel, line 50: Not 100% sure on this one. I feel like there could be some better way to describe this behavior.

An example (e.g. using `here()`) could be helpful.

```suggestion
/// Moreover, `layout` also provides [context]($context) to its argument's
/// function, meaning you don't need to use it in combination with the
/// `context` keyword.
```

crates/typst/src/layout/layout.rs
```diff
@@ -9,10 +9,10 @@ use crate::introspection::Locatable;
 use crate::layout::{Fragment, LayoutMultiple, Regions, Size};
 use crate::syntax::Span;
 
-/// Provides access to the current outer container's (or page's, if none) size
-/// (width and height).
+/// Provides access to the current outer container's (or page's, if none)
+/// dimensions (width and height).
 ///
-/// The given function must accept a single parameter, `size`, which is a
+/// `layout` accepts a function that receives a single parameter, which is a
```
PgBiel, line 15: Perhaps this makes more sense given the preceding phrase (feel free to disagree - just brainstorming):

```suggestion
/// Accepts a function that receives a single parameter, which is a
```

PgBiel (commented):
Thanks for this! Leaving some initial opinions.

PgBiel:
> Sorry for the slight spam of documentation PRs from me.

Keep 'em coming! :)

crates/typst/src/layout/layout.rs
MDLC01: I agree with the rephrasing. Not sure this is worth adding an example, though. I fear it would increase the length of the - already long - documentation too much for nothing (and this sentence is clear enough in my opinion).

crates/typst/src/layout/layout.rs
```diff
@@ -27,11 +27,11 @@ use crate::syntax::Span;
 /// ])
 /// ```
 ///
-/// If the `layout` call is placed inside of a box width a width of `{800pt}`
-/// and a height of `{400pt}`, then the specified function will be given the
-/// parameter `{(width: 800pt, height: 400pt)}`. If it placed directly into the
-/// page it receives the page's dimensions minus its margins. This is mostly
-/// useful in combination with [measurement]($measure).
+/// If the `layout` call is placed inside a box with a width of `{800pt}` and a
+/// height of `{400pt}`, then the specified function will be given the parameter
```
laurmaedje, line 31:
```suggestion
/// height of `{400pt}`, then the specified function will be given the argument
```
I think generally, `parameter` is the variable at the def-site and `argument` is the actual value at the call-site, so I think this makes more sense.

crates/typst/src/layout/layout.rs
laurmaedje: `[context]($context)` can just be `[context]` if I'm not mistaken

laurmaedje: The wordings seem very similar, I think either is fine.

laurmaedje (commented):
> Sorry for the slight spam of documentation PRs from me.

Absolutely nothing to be sorry for, on the contrary! :)

laurmaedje:
Thanks!

Issue #4188, opened 2024-05-19
Corner unicodes (⌜⌝⌞⌟) are positioned wrong in math mode


Zeta611:
### Description

Corner unicodes (U+231C--U+231F) are positioned wrong in math mode:
```typst
$⌜a⌟⌞b⌝$ vs $⌜$$a$$⌟$$⌞$$b$$⌝$
```
<img width="331" alt="Screenshot 2024-05-19 at 14 45 37" src="https://github.com/typst/typst/assets/9553691/f10c6ecd-8829-477a-b497-b75981d973ee">

Compare this with LaTeX:
```tex
\documentclass{article}
\usepackage{unicode-math}
\begin{document}
$⌜a⌟⌞b⌝$ vs $⌜$$a$$⌟$$⌞$$b$$⌝$
\end{document}
```
<img width="198" alt="Screenshot 2024-05-19 at 14 47 34" src="https://github.com/typst/typst/assets/9553691/b9b96b22-03b5-4c06-901e-8e3b98673f3e">

both giving the same results.

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

Enivex:
The issue goes away if at least one of them is escaped (for instance `$⌜a\⌟$`), so I assume this has to do with delimiter matching somehow.

Enivex:
Resolved by https://github.com/typst/typst/pull/4200

PR #1756, opened 2023-07-20
Center delimiters vertically about the axis

damaxwell:
Closes #540

This is the same song, different verse of the recently merged #1735.  The XITS Math parentheses have a low baseline.

image

However, the baseline plays no role in math layout:

TeXbook, Appendix G, Rule 19 (deep in the paragraph): "Shift the delimiters up or down so that they are vertically centered with respect to the axis."

XeLaTeX and MS Word have no trouble with delimiters with XITS Math.

This PR ensures that delimiters (which arise in vector, matrix, binomial coefficient, and LR constructions) are centered on the axis.

There were 9 test files that had changes.  All were very minor (2% or less pixel variations) associated with minor shifts associated with delimiters.

```
#set text(size:12pt)
#show math.equation: set text(font:"XITS Math", fallback:false)
$
( A / (sum_c^d k) )
$
```

## Before
image

## After
image

PR #4200, opened 2024-05-20
Align corner brackets to the top and bottom when scaling

bluebear94:
See #4188

This broke scaling of the corner brackets, U+231C – U+231F. Hopefully this didn’t happen to be needed? Right?


PgBiel:
> This broke scaling of the corner brackets, U+231C – U+231F.

Could you provide some images comparing the old and the new behavior? Should make it easier to tell if that's a problem or not (e.g. if the previous behavior was unintended).

Enter-tainer:
I think there are images in #4188 but i wonder if it's the correct place to apply "fix". And i'm also surprised to see non of other tests fail

Enivex:
> I think there are images in #4188 but i wonder if it's the correct place to apply "fix". And i'm also surprised to see non of other tests fail

Presumably because these are some of the few delimiters that arent vertically symmetric.

Enivex:
https://github.com/typst/typst/pull/1756

This was done purposely

bluebear94:
Both LuaTeX and XeTeX have incorrect results with `\left` and `\right` (though less so with LuaTeX):

```tex
\usepackage{unicode-math}

\[\ulcorner c \urcorner \ulcorner \frac{a}{b} \urcorner\]

\[\left\ulcorner c \right\urcorner \left\ulcorner \frac{a}{b} \right\urcorner\]
```

gives the following in LuaLaTeX:

image

and the following in XeLaTeX:

image

Enivex:
Okay, so I guess we'll have to come up with a solution ourselves.

Enivex:
Can these specific delimiters be top and bottom aligned?

tests/suite/math/delimited.typ
```diff
@@ -62,3 +62,7 @@ $ 1/(2 y (x) (2(3)) $
 // Test ignoring weak spacing immediately after the opening
 // and immediately before the closing.
 $ [#h(1em, weak: true)A(dif x, f(x) dif x)sum#h(1em, weak: true)] $
+
+--- math-lr-corner-brackets ---
+// Test positioning of U+231C to U+231F
```
laurmaedje, line 67: could you add a comment here that they are supposed to look the same? the "vs" had me briefly thinking that they are supposed to look _different._

could also rename it to `issue-4188-lr-corner-brackets` to link it to that issue

crates/typst/src/math/fragment.rs
```diff
@@ -408,9 +410,16 @@ impl VariantFragment {
...
+    }
+
+    /// Vertically adjust the fragment's frame so that it is aligned
+    /// to the given alignment on the axis.
+    pub fn align_on_axis(&mut self, ctx: &MathContext, align: VAlignment) {
         let h = self.frame.height();
         let axis = ctx.constants.axis_height().scaled(ctx, self.font_size);
-        self.frame.set_baseline(h / 2.0 + axis);
+        let align = (2 - align as u8) as f64;
+        self.frame.set_baseline((h / 2.0 + axis) * align);
```
laurmaedje, line 422: can this use `align.position(..)`? The `(2 - align as u8)` is too clever for my liking and also creates an implicit dependence on the order of the enum.

tests/suite/math/delimited.typ
bluebear94: I’ve changed the `vs` to an `=`.

crates/typst/src/math/fragment.rs
bluebear94: This hasn’t been defined for `VAlignment` yet, but I’ve done this.

laurmaedje:
Thanks!

istudyatuni:
I guess linked issue could be closed

PR #2517, opened 2023-10-30
Do not create a block frame for out-of-flow items only

MDLC01:
Fixes #2297

This PR makes it so that, when a breakable block is broken into multiple frames, a frame containing only out-of-flow items (which correspond to `place` elements, counter updates, etc.) will move its items to the next frame if there is one.

***Please review this carefully!*** I was not familiar with this part of the codebase when I started working on that, so please make sure I did not write anything stupid. All tests pass, so it should be ok, but there are still some things I'm not fully confident about (see comments). In particular, because this fix is implemented by modifying the way `FlowElem` is laid out, I think it may affect other things (although I could not think of any).

PR #4209, opened 2024-05-21
Fix widow pad frames

### LINKED ISSUE #4234: Element with no content followed by `pad` can force breakable block to create an empty frame
### Description

This bug has a similar effect as #2297, which was fixed by #2517. Sadly, the fix does not work in the specific case where the element with no content is directly followed by a `pad` element (which happens somewhat often given that `block` used `pad` under the hood to make `inset` work).

### Example

```typ
#set page(height: 3cm, margin: 0.5cm)

Not enough space remaining on this page to fit the whole block.

#block(stroke: 1pt)[
  #counter("dummy").update(1)
  #pad(left: 1cm, top: 1cm, rect(height: 1cm, fill: gray))
]
```

An empty frame is created for the counter update.

Typst produces the same output without an explicit `pad`:

```typ
#set page(height: 3cm, margin: 0.5cm)

Not enough space remaining on this page to fit the whole block.

#block(stroke: 1pt)[
  #counter("dummy").update(1)
  #block(inset: (left: 1cm, top: 1cm), rect(height: 1cm, fill: gray))
]
```

Note that removing the counter update removes the undesired rectangle.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

MDLC01:
Fixes #4234.

It turns out the empty frames created by `pad` have a non-zero size (they had padding applied to them). This made #2517 not work with those frames.

This PR solves the issue by applying the following transformation: each of the empty frames generated by `pad` that come before any content become zero-sized.[^1] In case of a `pad` element that generates only empty frames, the last one is always kept in order to make sure the element never "disappears".

[^1]: It is actually a tiny bit more complicated than that, due to some frames having negative dimensions (see [this Discord message](https://discord.com/channels/1054443721975922748/1088371867913572452/1242205845102657586)).


MDLC01:
Turns out this is harder than I thought. I'm marking this as a draft for now.

MDLC01:
Actually, I just forgot to add a test ref! So this PR is now ready for review.

laurmaedje:
Is there an associated issue for this?

The whole approach of skipping zero-sized frames in the flow is a bit of hack in my opinion. Adding to this with more complexity in padding seems like a potential maintainance burden to me. It is already very hard to change this in the flow already due to the various ad-hoc bug fixes in it.

MDLC01:
> Is there an associated issue for this?

There is not. I can create one.

Another idea I had after making this PR, which may feel less hacky, was to allow fragments to contain "fillers" instead of frames, which act as zero-sized frames, and are used instead of an empty frame that would have been created to fill a region when nothing fits into it. I don't know how this related this would be to #4212 (if at all).

MDLC01:
I created #4234.

laurmaedje:
Let's wait with this until after the layout refactor (if I don't fix it myself during it). Every extra somewhat ad-hoc fix makes incremental refactoring harder.

MDLC01:
Ok. I'm closing this PR, then.

PR #4211, opened 2024-05-21
Allow somewhat arbitrary characters as `mat`, `vec` and `cases` `delim`

MDLC01:
This PR implement the design that was briefly [discussed on Discord](https://discord.com/channels/1054443721975922748/1176478139757629563/1242520922318180372). `delim` can now be specified in one of the following form.

- `none`, for no delimiters.

- A delimiter to use as the opening delimiter. The closing delimiter is inferred automatically.
  - In case of an "opening", the corresponding closing delimiter is the corresponding "closing".
  - In case of a "closing", the corresponding closing delimiter is the corresponding "opening".
    - For example, specifying `delim: ")"` results in inverted parentheses.
  - In case of a "fence", the corresponding closing delimiter is the same as the opening delimiter.

- An array of two delimiters to use respectively as the opening delimiter and as the closing delimiter.

In the definition above, the term "delimiter" refers to a value of one of th following forms.

- `none`, for no delimiter.

- A `symbol`, or a single-character `str`.
  - Must be a Unicode math "opening", "closing", or "fence".

"somewhat arbitrary" in the PR title refers to the fact that only "opening", "closing" and "fence" characters are allowed.

## Original description.

This PR adds double brackets, triple bars, and angles as valid `mat`, `vec`, and `cases` `delim`s.

I also made it possible to use Unicode characters to specify the delimiters (specifically, `"⟦"` for double brackets, `"‖"` for double bars, `"⦀"` for triple bars, and `"⟨"` for angles).

Regarding other potential delimiters, I tried double angles, but they did not seem to work in the tests (instead resulting in a box glyph), and fences and double fences, which do not scale.

For reference, below are all the possible delimiters:
Parentheses, brackets, double brackets, curly braces, vertical bars, double vertical bars, angles.


Enivex:
Is there a specific reason why there is so much duplication when it comes to delimiters? To me, it seems like typst should "know" what all the scalable delimiters are, and allow them where appropriate (matrices, vectors, cases, possibly more?)

MDLC01:
I'm not sure exactly, but I figured we needed an enum like `Delimiter` because a delimiter is inputted as a string by the user, but actually corresponds to two Unicode codepoints (one for the opening delimiter, one for the closing delimiter).

Enivex:
> I'm not sure exactly, but I figured we needed an enum like `Delimiter` because a delimiter is inputted as a string by the user, but actually corresponds to two Unicode codepoints (one for the opening delimiter, one for the closing delimiter).
> 
> Also, matrices, vectors and cases already share the same `Delimiter` type.

Does that mean cases automatically gets these as well?

MDLC01:
> > I'm not sure exactly, but I figured we needed an enum like `Delimiter` because a delimiter is inputted as a string by the user, but actually corresponds to two Unicode codepoints (one for the opening delimiter, one for the closing delimiter).
> > Also, matrices, vectors and cases already share the same `Delimiter` type.
> 
> Does that mean cases automatically gets these as well?

Yes, it does!

Enivex:
While you're at it, you may as well add triple bars as well.

(There are plenty of other delimiters too, but most are very obscure)

MDLC01:
This PR is quickly becoming a mess, mainly due to me trying to do too many things at once. Maybe we should close it and discuss a design that would allow specifying any pair of delimiters?

Enivex:
I didn't mean to derail this PR. The other changes could be a future one?

MDLC01:
If we opt for a design similar to [`smartquote`'s `quote` parameter](https://typst.app/docs/reference/text/smartquote/#parameters-quotes), I can implement it as part of this PR, especially because this PR would be pretty much overridden by one that implements the other changes.

MDLC01:
I implemented the ability to use any Unicode "opening", "closing" or "fence" character. See the updated PR description for more information.

crates/typst/src/math/matrix.rs
```diff
@@ -330,59 +318,119 @@ impl LayoutMath for Packed<CasesElem> {
...
+
+    v: Array => {
+        let v = v.as_slice();
+        if v.len() != 2 {
+            bail!("expected 2 delimiters, found {}", v.len());
+        }
+        let open = v[0].clone().cast()?;
+        let close = v[1].clone().cast()?;
+        Self { open, close }
+    },
```
laurmaedje, line 404:
```suggestion
    v: Array => match v.as_slice() {
        [open, close] => Self {
            open: open.clone().cast()?,
            close: close.clone().cast()?,
        },
        _ => bail!("expected 2 delimiters, found {}", v.len())
    },
```

MDLC01:
> `||` is not supported anymore, correct?

Indeed. Moreover, `mat(delim: ||, ..)` does not work either, probably due to some parsing limitations. This has to be done with `math(delim: bar.double, ..)` or `math(delim: "‖", ..)`.

laurmaedje:
> Moreover, mat(delim: ||, ..) does not work either, probably due to some parsing limitations. This has to be done with math(delim: bar.double, ..) or math(delim: "‖", ..).

Not a big deal I think. Thank you!

PR #4216, opened 2024-05-22
Improving error message for invalid file types

### LINKED ISSUE #2486: Better error message when trying to compile wrong file type.
### Description

Current behavior:
```bash
~$ typst compile main.pdf
error: file is not valid utf-8.
```
New behavior:
```
~$ typst compile main.pdf
error: file is not valid utf-8.
help: a file with suffix `.pdf` is usually not a typst file. Maybe you meant `main.typ` instead?
```
`main.pdf` is a stand-in for any file that is not a typst file and has the wrong file extension.

The additional help message would only appear when the file ending is wrong **and** another error in the file happens. There wouldn't be any complaints if a different file extension was used but the file was valid typst source.

### Use Case

Today I tried several times to compile `main.pdf` before I realized that I meant to type `main.typ`. A help message would have nudged me to the correct solution earlier.

PepinhoJp:
Closes #2486 

Refactored code for loading main source file with utf-8 error detection, providing a helpful message. See examples below:

```
> ./typst compile typst

error: file is not valid utf-8
 = hint: a file without an extension (`.something`) is not usually a Typst file. Check if you meant to add `.typ` to its name.
```

```
> ./typst.exe compile typst.exe

error: file is not valid utf-8
 = hint: a file with the `.exe` extension is not usually a Typst file. Check if you meant to use `.typ` instead.
```

### DIFF
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -100,42 +100,45 @@ pub fn compile_once(
     }
 
     // Check if main file can be read and opened.
-    if let Err(error) = world.source(world.main()) {
+    let main_result = world.source(world.main());
+    if main_result.is_err() {
         set_failed();
         if watching {
             Status::Error.print(command).unwrap();
         }
 
-        let was_invalid_utf8 = matches!(error, FileError::InvalidUtf8);
-        let mut errors =
-            FileResult::<Source>::Err(error).at(Span::detached()).unwrap_err();
-
-        if was_invalid_utf8 {
-            match &command.common.input {
+        let errors = match main_result {
+            Err(FileError::InvalidUtf8) => match &command.common.input {
                 Input::Path(path) => {
                     let extension = path.extension();
+                    let mut errors = main_result.at(Span::detached()).unwrap_err();
                     if let Some(extension) = extension {
                         if extension != "typ" {
                             errors
                                 .make_mut()
                                 .first_mut()
                                 .unwrap()
-                                .hint(eco_format!("a file with the `.{}` extension is not usually a Typst file. Check if you meant to use `.typ` instead.", extension.to_string_lossy()))
+                                .hint(eco_format!("a file with the `.{}` extension is not usually a Typst file. Check if you meant to use `.typ` instead.", extension.to_string_lossy()));
                         }
                     } else {
                         errors
                             .make_mut()
                             .first_mut()
                             .unwrap()
-                            .hint("a file without an extension (`.something`) is not usually a Typst file. Check if you meant to add `.typ` to its name.")
+                            .hint("a file without an extension (`.something`) is not usually a Typst file. Check if you meant to add `.typ` to its name.");
                     }
+
+                    errors
                 }
                 _ => {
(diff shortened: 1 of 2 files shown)


MDLC01:
I'm not convinced the second part of the hint ("Check if you meant to add `.typ` to its name." / "Check if you meant to use `.typ` instead.") is useful. Changing the file's extension won't magically make its content valid UTF-8. Otherwise, I really like this hint!

crates/typst-cli/src/compile.rs
```diff
@@ -98,12 +98,46 @@ pub fn compile_once(
...
             Status::Error.print(command).unwrap();
         }
 
+        let errors = match main_result {
+            Err(FileError::InvalidUtf8) => match &command.common.input {
+                Input::Path(path) => {
+                    let extension = path.extension();
+                    let mut errors = main_result.at(Span::detached()).unwrap_err();
+                    if let Some(extension) = extension {
+                        if extension != "typ" {
```
MDLC01, line 114:
```suggestion
                        if extension != "typ" && extension != "typst" {
```
Some people may use `.typst` instead of `.typ`.

Coekjan: `.typc` is also a valid typst extension, using for typst code mode.

MDLC01: I don't think it is used as a file extension. Moreover, a file with the `typc` extension would not be interpreted as code by the compiler (it is treated just like any regular Typst file: as markup). So I don't think if would make sense to add `typc` as an exception as well.

laurmaedje: Agreed, typc ist just a highlighting tag.

Coekjan: Thanks for your correction.

PgBiel: > Some people may use `.typst` instead of `.typ`.

I've only seen that in one specific occasion so far, but is this something we want to officially support?

PgBiel:
> I'm not convinced the second part of the hint ("Check if you meant to add `.typ` to its name." / "Check if you meant to use `.typ` instead.") is useful. Changing the file's extension won't magically make its content valid UTF-8.

I believe the rationale for this is in the original issue, where you may accidentally (for example) use tab autocomplete towards a previously generated PDF file with the same name instead of the source file.

With that said, it could be removed for the case without an extension, or even kept just for `png` / `svg` / `pdf`.

PepinhoJp:
The compiler will now only suggest adding .typ if the .typ file exists. @MDLC01 thanks for your input.

laurmaedje (commented):
I think the implementation could still be simplified a bit, I've left a few comments to that effect. 

I now also saw why the `Hint<T> for SourceResult<T>` impl wasn't in place yet, but I think with a few other changes I proposed, it won't be needed anymore anyway. Sorry for sending you down the wrong path with this!

laurmaedje:
Thank you!

SebastianJL:
Thanks y'all for doing this ❤️ I originally opened the issue but wasn't able back then to come up with a satisfying solution in code.

PR #4226, opened 2024-05-22
Make block equations breakable

EpicEricEE:
Allows breaking block equations at page (or column) boundaries while keeping alignment in place.

The behaviour can be configured via a show-set rule
```typ
```
and is currently set to `false` by default. I'm not sure if this would be better as a property on equation itself, as there isn't really a block element underneath it.

The numbering is repeated for every section with the same alignment as set via the `number-align` parameter. This would ideally be a bit more configurable, but I don't know how that should look like.

As equations can only be split at line breaks, an equation containing a breakable block will not break inside that block. Also, gradients are not continuous across multiple sections, but are instead repeated (as is the fill of breakable blocks).

image image


crates/typst/src/math/equation.rs
```diff
@@ -248,26 +249,87 @@ impl Packed<EquationElem> {
...
+
+        let equation_builders = if BlockElem::breakable_in(styles) {
+            let mut rows = full_equation_builder.frames.into_iter().peekable();
+            let mut equation_builders = vec![];
+            let mut last_first_pos = Point::zero();
+
+            for region in regions.iter() {
+                // Keep track of the position of the first row in this region,
+                // so that the offset can be reverted later.
+                let Some(first_pos) = rows.peek().map(|(_, pos)| *pos) else { break };
```
laurmaedje, line 279:
```suggestion
                let Some(&(_, first_pos)) = rows.peek() else { break };
```

crates/typst/src/math/equation.rs
```diff
@@ -248,26 +249,87 @@ impl Packed<EquationElem> {
...
+                    let mut pos = *pos;
+                    pos.y -= first_pos.y;
+
+                    if !region.y.fits(sub.height() + pos.y) {
+                        // Allow first row to overflow to prevent infinite
+                        // creation of new regions which may all be too small.
+                        if !(frames.is_empty() && regions.in_last()) {
+                            break;
+                        }
+                    }
```
laurmaedje, line 294:
```suggestion
                    // Finish this region if the line doesn't fit. Only do it if
                    // we placed at least one line _or_ we still have non-last
                    // regions. Crucially, we don't want to infinitely create
                    // new regions which are too small.
                    if !region.y.fits(sub.height() + pos.y)
                        && (!frames.is_empty() || !regions.in_last())
                    {
                        break;
                    }
```
This is a matter of personal preference, but I find expressions like `!(a && b)` super hard to understand, so I'd prefer to apply De-Morgan's law here.

laurmaedje:
Thank you!

PR #4227, opened 2024-05-22
Add more environment control parameters to CLI

### LINKED ISSUE #3892: More control over typst compilation environment
### Description

To improve reproducibility across systems compiling a Typst document, it would be beneficial to have complete control over the parts of the compilation environment which are implicitly used.

The following things are implicitly taken and not part of inputs of the typst binary:
- fonts
- package data directory
- package cache directory

`TYPST_FONTS` is currently additive, there is no way to configured Typst to not include system fonts to my knowledge. It would be great if this was not additive, but an absolute configuration of how to retrieve fonts. If the need for more than one path exists, then it could be split alongside a common inter path separator like `:` (similar to `PATH`).

Setting `XDG_DATA_DIR`/`XDG_CACHE_DIR` can have side effects on other programs run in the same context as Typst. While they are good fallbacks, there should be a first-class `TYPST_*` environment variable for the respective package cache and package data directories which is considered first before fallback to the respective `$XDG_*/typst` occurs.

During discussion about this in the Typst community discord, it was suggested that these should, similar to other options, also be optional compiler flags.

### Use Case

This aids in isolating the effect of environment variables on the rest of the system and helps reproducibility by removing implicit assumptions on the compilation environment.

LuizAugustoPapa:
Closes #3892

PR co-authored by @PgBiel , @tulio240 , @PepinhoJp

This PR adds a few parameters to the Typst CLI which lets you control the environment accessed by Typst:

1. `--package-cache-path` or (env) `TYPST_PACKAGE_CACHE_PATH`: Set this to change where downloaded Typst packages are installed to. For example, `#import "@preview/cetz:0.2.2"` would download `cetz:0.2.2` to `$TYPST_PACKAGE_CACHE_PATH/preview/cetz/0.2.2` (instead of the usual `$XDG_CACHE_HOME/typst/packages/preview/cetz/0.2.2`, on Linux as an example)

2. `--package-path` or (env) `TYPST_PACKAGE_PATH`: Set this to change the root folder where local packages are searched. For example, `#import "@local/tablex:0.0.8"` would search for `tablex:0.0.8` at `$TYPST_PACKAGE_PATH/local/tablex/0.0.8` (instead of the usual `$XDG_DATA_HOME/typst/packages/local/tablex/0.0.8`, on Linux as an example)

3. `--ignore-system-fonts` (no env): Specify this to disable searching for system fonts, only use embedded fonts or fonts in specified in `--font-path`

4. (for `typst update` only) `--backup-path` or (env) `TYPST_UPDATE_BACKUP_PATH`: Set this to change the folder where the Typst binary is backed up to before an update

We suggest giving these features a quick test to ensure they work in different environments.

A quick note: We used an `Arc` in the `FileSlot` field with package storage path information to avoid having to use a reference with a lifetime generic parameter in the `FileSlot` struct (since this information is shared between all `FileSlot`, so didn't want to clone). Let us know if this should be changed

Credit to @bluebear94 for the idea of grouping up arguments related to font settings together.


crates/typst-cli/src/update.rs
```diff
@@ -40,7 +40,13 @@ pub fn update(command: &UpdateCommand) -> StrResult<()> {
         }
     }
 
-    let backup_path = backup_path()?;
+    let backup_dir = command.backup_path.clone().map(Ok).unwrap_or_else(backup_path)?;
+
+    fs::create_dir_all(&backup_dir)
+        .map_err(|err| eco_format!("failed to create backup directory ({err})"))?;
+
+    let backup_path = backup_dir.join("typst_backup.part");
```
laurmaedje, line 48: I wonder, if the user specifies `--backup-path`, shouldn't that be the full path rather than a directory to which `typst_backup.part` is joined?

PgBiel:
> I wonder if ` --packages-path` adding a new search path for packages and having ` --ignore-default-packages-path` as a way to avoid the default path would be interesting.
> 
> The usecase would be adding organization wide packages or templates using shared folders, while keeping the user's default package path for other packages. The ignore-default-packages-path, on the other hand, would allow some form of isolation and ease testing.

Currently, the default path is no longer searched. We considered adding a way to have multiple search paths, which seems useful in principle, but we think this is probably out of scope for this particular PR. Perhaps there could be some per-namespace option, e.g. use `@project` for the project's local packages and `@local` for system-wide local packages.

crates/typst-cli/src/update.rs
PgBiel: I think this is unnecessary in principle. Also adds a bit of a burden to remember the correct filename when using `--revert`.

laurmaedje: I was confused mostly by the variable naming here. It points towards an inconsistency, either in the CLI arg naming or in the source code.

PgBiel: Fair. I think my argument also doesn't really hold because, in practice, you'd have to remember the directory regardless.

I considered that keeping it as a directory would probably be more "future-proof" since we can add more files to the backup without problems, but I guess we could just generate a `tar` instead then.

I thought about renaming it to  `--backup-file-path` or `--backup-file` to make the difference from other parameters more apparent, but wasn't sure. Maybe changing it to a file isn't that bad anyway?

pachi:
> Currently, the default path is no longer searched. We considered adding a way to have multiple search paths, which seems useful in principle, but we think this is probably out of scope for this particular PR. Perhaps there could be some per-namespace option, e.g. use `@project` for the project's local packages and `@local` for system-wide local packages.

I'm not sure if distinguishing between those two namespaces is worth it. That would mean that the template import changes not due to network use but due to local path. IMHO, @local meaning no network access vs @preview is interesting, but more finegrained control over where the local packages would be fetched should done with the env vars.

PgBiel:
Sorry, I didn't mean to imply that those namespace choices I mentioned would be hardcoded or available by default - I meant that you'd be able to pick whichever namespace names you wanted for your own directories of local packages, as is possible today by creating folders other than "local" under `(DATA)/typst/packages`. With an additional parameter, you could define a namespace named "@supercoolproject" pointing to a custom folder in your computer for example.

laurmaedje:
Though with that configuration we're nearing typical compiler flags for linking external packages, rather than typical package manager style configuration for external packages. I wonder whether such path dependencies aren't better put into a `typst.toml`.

pachi:
> Though with that configuration we're nearing typical compiler flags for linking external packages, rather than typical package manager style configuration for external packages. I wonder whether such path dependencies aren't better put into a `typst.toml`.

Being able to configure in a typst.toml the paths for fonts, (extra)packages, etc could be great for some projects.

Being able to use env vars IMHO is still convenient when using typst from another app, which would probably send the content to the compiler using stdin, so no document specific files would be used on disk, only resources (fonts, packages, images, etc).

crates/typst-cli/src/update.rs
PgBiel: Alright, I've made the following changes:

1. Renamed `--packages-path` to `--package-path`.
2. Changed `--backup-path` to point to the backup file and updated the relevant arg docs.

I think this should be ok now.

laurmaedje:
Thank you all!

drupol:
Excellent !!! Looking forward to using this in Nix :)

CC @loqusion (context: https://discourse.nixos.org/t/typix-a-nix-library-for-compiling-typst-projects/40257)

matthiasbeyer:
I came across while searching for a solution to the problem that `typst` seems to ignore the `XDG_DATA_HOME` and always tries `~/.local/share` first.

This would solve that issue, from what I read. Cannot wait to have it available in a release!

PgBiel:
> `typst` seems to ignore the `XDG_DATA_HOME` and always tries `~/.local/share` first.

That would be a bug; if you can still reproduce it in Typst 0.12-rc2 (the latest release candidate), please open an issue.

matthiasbeyer:
Just to be sure: With `XDG_CACHE_HOME`, everything seems to work. Above I tried `XDG_DATA_HOME`, which didn't work.
Would that still be a bug?

PR #4228, opened 2024-05-22
Add nested import syntax

### LINKED ISSUE #2148: Convenient syntax for nested import
### Description

It would be highly convenient to allow full folders to exist as local packages:

```typ
#import "@local/my-templates/proposal/nsf.typ": *
```

Currently, this results in the following error:
```
error: `templates/proposal/nsf` is not a valid package name
```

### Use Case

I have one templates folder with several options:
```
templates
├───slides
|   ├───dark.typ
|   └───light.typ
└───proposal
    ├───nsf.typ
    └───sbir.typ
```
Currently, the best workaround is to use a relative import:
```typ
#import "../../../templates/slides/dark.typ": *
```

This is not ideal because (1) it is sensitive to the location of the main document and (2) requires typst to be run with `--root` to work.

I would love to symlink this folder to `typst/packages/local` and access it as a "package" to increase reusability. Even if I were to "publish" `templates` as a package to make this happen, it would still require "flattening" each template into something individually accessible since only one entrypoint can be specified:

```typ
#import "@local/templates:0.1.0": slides-dark
// Even so, this does not allow for star-imports of *only* the "slides/dark" template helpers
```

It seems like potential solutions are to:
- Allow multiple entrypoints
- Allow the entrypoint to be a keyword like "folder" or "none" or something similar that indicates there are subpackages
- Allow arbitrary folders to be placed in `local` without complaining that they are nonconformant packages

tulio240:
Closes #2148

PR co-authored by @PgBiel, @LuizAugustoPapa, @PepinhoJp

Adds the syntax below to import names from submodules more easily (supports any amount of submodule nesting):

```typ
#import "@preview/cetz:0.2.2": draw.rect
#import std: array.zip as azip

// that's almost equivalent to
// (but without creating the 'draw' and 'array' variables)

#import "@preview/cetz:0.2.2": draw
#import std: array

#let rect = draw.rect
#let azip = array.zip
```


crates/typst-syntax/src/kind.rs
```diff
@@ -275,6 +273,11 @@ pub enum SyntaxKind {
     Error,
     /// The end of token stream.
     End,
+
+    /// A path to an imported name from a submodule: `a.b.c`.
+    ImportItemPath,
+    /// A renamed import item: `a as d`.
+    RenamedImportItem,
```
laurmaedje, line 280: After you've merged with main, these can be moved back up to the original spot. I've moved the problematic comment syntax kinds instead, in #4287.

laurmaedje:
Thank you all!

PR #4183, opened 2024-05-18
Hint for language-region pair on `text.lang`

Coekjan:
This pr closes #4142 . If I get it correctly, `text.lang` is the only place parsing ISO language code.

I added a new function `parse_lang` to parse an argument for `text.lang`:
1. Use `Lang::from_str` to parse the arg. If success, all done.
2. Check whether the arg is a lang-region pair. If so, give a hint.

PR #4229, opened 2024-05-22
Refactor casts to use `HintedStrResult`

PgBiel:
Related: https://github.com/typst/typst/pull/4183

- `FromValue::from_value` now returns a `HintedStrResult` instead of a `StrResult` (and so does `Value::cast` as a consequence).
  - This makes it easier to display detailed errors resulting from bad casts.
- Experimented with changing `HintedString`'s representation to a single `EcoVec` containing both the message and hints. This makes it smaller in the stack, but it may be potentially slower to access the message within. Still, I assume reading error messages isn't in the hot path anyway. (The main goal is making `HintedStrResult` smaller in the stack. It now occupies a minimum of two words.)
   - This does have the disadvantage that each `HintedString` created will always incur a heap allocation, even if there are no hints. But maybe this isn't very relevant since we don't create warnings and errors that often.
   - One alternative solution would be to simply convert the `hints` field of `HintedString` into `EcoVec` instead of `Vec`; that'd already help reduce its stack size overall (and/or combine this with `Box` in `HintedStrResult`).


crates/typst-cli/src/main.rs
```diff
@@ -37,17 +38,25 @@ fn main() -> ExitCode {
     let timer = Timer::new(&ARGS);
 
     let res = match &ARGS.command {
```
laurmaedje, line 40: maybe we could move the match into a separate function, so that we can use `?` instead of all the `map_err`? could be called `run` or `dispatch`.

crates/typst/src/diag.rs
```diff
@@ -161,6 +161,18 @@ impl SourceDiagnostic {
         }
     }
 
+    /// Create a new error with hints.
+    pub fn hinted_error(span: Span, hinted_message: HintedString) -> Self {
```
laurmaedje, line 165: I think it's nicer if we keep this in the `At` impl, so that there is only one way to do this for API consumers.

crates/typst/src/diag.rs
```diff
@@ -298,27 +310,51 @@ pub type HintedStrResult<T> = Result<T, HintedString>;
 /// A string message with hints.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct HintedString {
+    /// The string's components as a single vector of strings.
+    /// The first element of the vector contains the message.
+    /// The remaining elements are the hints.
+    /// This is done to reduce the size of a HintedString.
+    /// The vector is guaranteed to not be empty.
+    components: EcoVec<EcoString>,
```
laurmaedje, line 318: Since it's just a single field, we can use `pub struct HintedString(EcoVec<EcoString>);`

crates/typst/src/diag.rs
```diff
@@ -298,27 +310,51 @@ pub type HintedStrResult<T> = Result<T, HintedString>;
...
+}
+
+impl HintedString {
+    /// Creates a new hinted string with the given message and hints.
+    pub fn new(message: EcoString, hints: impl IntoIterator<Item = EcoString>) -> Self {
```
laurmaedje, line 323: I think the constructor should not take the hints. Rather, we would replicate the full set of `hint`, `with_hint`, and `with_hints` from `SourceDiagnostic`. That seems like a more convenient API to me.

PgBiel:
Done. I think we might have some other things we want to use hinted casts for as well.

For example, I remember I wanted to have the "use `grid.header` instead" part of the error below be a hint when I first wrote it, but couldn't due to the lack of hinted string in casts; do you think it'd make sense to convert it to a hint as well?

https://github.com/typst/typst/blob/23746ee18901e08852306f35639298ad234d3481/crates/typst/src/layout/grid/mod.rs#L428-L436

PR #4233, opened 2024-05-23
Check ill body of contextual element early

Myriad-Dreamin:
When editing the code from `#{context text.font}` to `#{context}`, the `typst_ide`'s `analyze_expr` panics at `src/eval/call.rs:286:68`.
```
thread 'analysis::track_values::tests::test_broken_expr' panicked at src/eval/call.rs:286:68:
```

https://github.com/typst/typst/blob/34f1a232463bafeb3d8d3fa9adeae33c4bb95b23/crates/typst/src/eval/call.rs#L284-L287

I printed the contextual element, and found that it is caused by the body of contextual expression `Eof` is not an expression.

```
context closure: Closure { node: Eof: "", defaults: [], captured: Scope {}, num_pos_params: 0 }
```

It is strange that `typst-cli` and webapp don't panic with this editing, but I indeed eliminate the panic by checking the legality of the body when constructing the contextual element.

### DIFF
--- a/crates/typst/src/eval/code.rs
+++ b/crates/typst/src/eval/code.rs
@@ -335,18 +335,21 @@ impl Eval for ast::Contextual<'_> {
     type Output = Content;
 
     fn eval(self, vm: &mut Vm) -> SourceResult<Self::Output> {
-        let body = self.body();
+        let body = self.body().to_untyped();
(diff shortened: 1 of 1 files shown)


laurmaedje:
Fixed more generally via https://github.com/typst/typst/pull/4288

PR #3953, opened 2024-04-18
Support highlight, overline, underline, and strike for math equations

LuxxxLucy:
Regarding #2200, it turns out the highlight is supported for normal text but not for math equations.  This PR implement the highlight for math equations.

Like highlight, other line-based decorations are supported now as well, including overline, underline, strikethrough. Result

image

(tests added in `deco.typ`)

## Old

This is the second of two PR for solving #2200  (and shoud be easier to review if #3941 is merged first) (or Just merge this PR directly)

It continues on #3941 that add support for overline underline and strike for math equations. Some refactoring and corresponding tests are added as well.

Issue #2200, opened 2023-09-21
highlight function not working with equations


cprieto:
### Description

Highlight is not highlighting equations but it works with text.

Example:

```
This is a #highlight[highlight sample], it works.

This is an equation: #highlight[$a_n = C_0 a_(n-1)$], it doesn't
```

Produces: 
![image](https://github.com/typst/typst/assets/94194/ad908f6a-5063-46ed-aba6-2998d9b49a32)

### Reproduction URL

https://typst.app/project/ww-AG8xGryot5E-agddCDs

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

8LWXpg:
similar issue also happens to `strike` and `text(baseline: length)`

```typst
$
#highlight[x $x$ xx]\
#strike[x $x$ xx]\
#text(baseline: -5pt)[x $x$ xx]\
$

#highlight[x $x$ xx]\
#strike[x $x$ xx]\
#text(baseline: -5pt)[x $x$ xx]\
```

<img width="99" alt="image" src="https://github.com/typst/typst/assets/105704427/ddd0369e-12de-4f02-ab12-9525e2cbfb06">

LuxxxLucy:
It seems that for now only TextElem would apply the line decorations (highlight, strikethrough, etc). It shoud not be difficult, I will work on it. 

The only problem I have is that the metrics provided as arguments to the line decoration, such as baseline, ascender, may only make sense to a TextElem, but does not really make sense for a math expression. For example, if we have a fraction surrounded by some normal text, 

```#highlight[We have $inline(1/2 < (x+1)/2)$ here]```

<img width="394" alt="image" src="https://github.com/typst/typst/assets/19356905/74f311ac-5fe6-4f31-80e4-9ec71ef07e6d">

how should we decide the height of the highlight rectangle for the inline math. As you can see, the math expr can easily go beyond the height of normal text. Should we simply place a highlight rectangle that covers the entirety of the math expression, or should be make it consistent to the surrounding text?

davidleejy:
Not sure if this quick and dirty workaround could help any Typst users out there:

If you find short character sequences in math equations not getting highlighted while longer ones do, for example:

If `$ a + b = #highlight[c] $` does not result in the `c` symbol being highlighted, and lengthening the argument of the highlight command by adding more characters: `$ a + b = #highlight[ccc] $` successfully highlights `ccc`.

[...]

PR #4236, opened 2024-05-23
Use line height based spacing model

### LINKED ISSUE #1028: rendering: vector and square root
### Description

The square root is rendered through the vector. See [HTI Dictaat Wiskunde 3.pdf](https://github.com/typst/typst/files/11352078/HTI.Dictaat.Wiskunde.3.pdf).

### Reproduction URL

https://typst.app/project/pxKIAuoZZl7hgc5cpu4nnH

### Operating system

Web app

### Typst version

### LINKED ISSUE #4224: Proposal: change `leading` option to `line-height`
### Description

I propose deprecating the `par.leading` option, and add a `par.line-height` option instead. 

### Use Case

# Motivation

## The current leading model is broken

Current leading model specifies the space between bottom edge of last line and top edge of the next line. This is fine when every line frame is of the same height but becomes undesirable when some lines become taller.  See https://github.com/typst/typst/issues/1028#issuecomment-2029983751 for example. To alleviate the problem, currently Typst hacks into the bounding box of inline formula. However, this workaround is not ideal as it leads to the exact issues described in #1028 and lacks a straightforward solution.

Furthermore, line spacing becomes uneven when text with different size appeared in paragraph. No proper solution exists under current leading model. 

Typst:
image

LaTeX:
image

## The current leading model is confusing and inconsistent with precedence

Typst define leading as “The spacing between lines”. However, other typesetting software such as Adobe InDesign and Affinity Publisher use the term “leading” to refer to the distance between text baselines. Other typesetting systems like CSS, LaTeX, and Microsoft Word do not employ the term “leading”, yet they all relate line spacing to the distance between baselines.

Typography guidelines ([example](https://practicaltypography.com/line-spacing.html)) consider line spacing in terms of baseline distance too.

# Proposal

Deprecate the `leading` option and introduce a new option called `line-height`, which represents the basic distance between baselines. `leading` can be removed in the future version along with `locate` function.

For each line frame $L_i$, define $H_i$ as the distance from the top edge to the baseline (height), and $D_i$ as the distance from the baseline to the bottom edge (depth). During layout processes, Typst should insert a blank frame of height $x$ between two lines, $L_1$ and $L_2$. The height of this frame is determined by:
$$x = \max(0, D_1 + H_2 - s)$$
where $s$ represents the `line-height` value. Essentially, we maintain consistent baseline spacing unless there is an collision.

In the future, an additional option `min-line-gap`$=g$ can be added, changing the above formula to
$$x = \max(g, D_1 + H_2 - s).$$
See below for the explanation of $g$.

# Alternative

## The TeX model
(Reference: _The TeXBook_, p.77-80)

TeX employs the following formula:

$$x = \begin{cases}D_1 + H_2 - s, &D_1 + H_2 -s \geq l\\\\g, &\text{otherwise}\end{cases}$$

where $x$ is the space inserted between lines, and $s$, $l$, $g$ corresponds to `\baselineskip`, `\lineskiplimit` and `\lineskip` respectively. Notably the first formula presented in Proposal section is a special case where $l = g = 0$. So the proposed formula is compatible to the LaTeX algorithm.

## The InDesign model
From [InDesign's document](https://helpx.adobe.com/indesign/using/leading.html):

> The default auto-leading option sets the leading at 120% of the type size. […] By default, leading is a character attribute, which means that you can apply more than one leading value within the same paragraph. The largest leading value in a line of type determines the leading for that line.

This model is not suitable for Typst because it do not play well with inline formulas.

Enter-tainer:
fix #1028 fix #4224 

The ci is expected to fail since the line height is changed, all of ref images are needed to be updated

TODO:

- [x] bring back orphan and widows prevention
- [ ] fix TODOs. specifically about the spacing of term/enum/list --> How to calculate row gutter for grids
- [x] Find a good default for line height in normal text and footnote
- [ ] figure out what it TIGHT_LEADING in math


peng1999:
> * Find a good default for line height in normal text and footnote

My suggestion is to find a value that keeps the line spacing unchanged in most tests...

Enter-tainer:
> > * Find a good default for line height in normal text and footnote
> 
> My suggestion is to find a value that keeps the line spacing unchanged in most tests...

image

It's impossible. Previously, the line height in typst is...(guess what?) 7.15(leading) + 7.240234375(asc) = 14.390234375pt. And this means that if we want it looks the same, the line-height would be 1.3082031250000001em or so.

But when i set it to 1.3082031250000001, the calculated leading is 7.150000000000002. And if i set it to the next float using `np.nextafter`, it becomes 7.149999999999999

crates/typst/src/math/row.rs
```diff
@@ -162,27 +159,31 @@ impl MathRun {
...
         let rows: Vec<_> = self.rows();
         let row_count = rows.len();
         let alignments = alignments(&rows);
 
-        let leading = if EquationElem::size_in(styles) >= MathSize::Text {
-            ParElem::leading_in(styles)
-        } else {
-            let font_size = scaled_font_size(ctx, styles);
-            TIGHT_LEADING.at(font_size)
-        };
```
Enter-tainer, line 174: anyidea what is tight leading? It is a workaround or something?

LuxxxLucy:
I just want to comment that #4318 may be a possible predecessor for the line height adjust solution can avoid breaking the test. This PR (#4318) is an attempt to support #2200 (draft PR #3953), and I think it should be relevant so I think you might be interested to take a look at it.

laurmaedje:
I think until we have a solution that properly fixes the equation problem (for the first line and the last line as well), I'd refrain from making any changes. (In the interest of only changing things once.) So I'll close this for now.

PR #4251, opened 2024-05-25
feat: pngs use background color of the main frame

### LINKED ISSUE #4088: PNG Output does not support transparency
## Synopsis
Setting no background results in a transparent svg, but opaquely white png.
## Example
```ts
#lorem(20)
```
Results in
### SVG
transparent

---
### PNG
transparent

### Reproduction URL

_No response_

### Operating system

Web app, Linux

### Typst version

benlubas:
Closes #4088

Allows rendering to PNG with a transparent background by explicitly setting the page fill color like: `#set page(fill: rbg(0, 0, 0, 0))` at the top of the document.

Documents that do not have `fill` set at the top are given a white background.

I'm missing tests. I can add those as well, but I'd need a few days.


benlubas:
Something is different about the test environment such that this change isn't effective in the test environment. Any help would be awesome.

Here's the test that I'm trying to add:
```typst
--- page-fill-transparent ---
// Pages with transparent background at the top should produce a transparent png
#set page(fill: rgb(0, 0, 0, 0), width: auto, height: auto, margin: 0pt)
#text(fill: olive)[hi]
```
here's the image it produces in the test:

page-fill-transparent

and here's the image produced by `cargo run -- compile test.typ test.png` on my branch (light mode users might want to turn on dark mode to see that's transparent):

test

---

To me it seems like there's some global test set rules getting applied to make the font size smaller among other things, and this is ruining this particular test. Is there a way to bypass these?

laurmaedje:
I'll need to discuss the details of dealing with background in different formats with @reknih again. Can't promise anything, but I'll try to do that soon.

Once we're sure that these are the semantics we want, I can take a look at the specifics in the code that fail the test runner.

laurmaedje:
I am considering whether we should simply make PNG export transparent by default, there is some more discussion on [Discord](https://discord.com/channels/1054443721975922748/1259074835850723328).

laurmaedje:
Closing in favor of https://github.com/typst/typst/pull/4586

PR #4256, opened 2024-05-25
Allow setting `raw.theme` to `none` to disable syntax highlighting

### LINKED ISSUE #4175: Disable syntax highlighting when raw theme is set to `none`
### Description

There does not seem to be an obvious way to remove syntax highlighting from code (raw) blocks without altering the source. 
It would be great if there is a way to disable syntax highlighting in Typst itself: e.g.,
```#set raw(lang: none)``` 

### Use Case

 The current solution (removing the language specifier from the source) interferes with other uses of that specifier, e.g., in  mdbook, code blocks need to be annotated with the language specifier to enable e.g., testing the code blocks. The other solution ```#show raw.line: it => it.text``` seems hacky and not obvious. Most users will expect setting raw(lang:none) to disable syntax highlighting.

### PR DESCRIPTION
**Pull request by Coekjan**
This pr closes #4175 .

As https://github.com/typst/typst/pull/4186#pullrequestreview-2071726156 , we allow setting default value of `raw.theme` to `none` to disable syntax highlighting. To implement it, I redesigned the function signature (return value type):
1. `Err(...)` indicates argument `theme` found but error met when casting it into `Smart<Option<EcoString>>`;
2. `Ok((None, None))` indicates argument `theme` not found;
3. `Ok((Some(Smart::Auto), None))` indicates argument `theme` found, and it is `auto`;
4. `Ok((Some(Smart::Custom(None)), None))` indicates argument `theme` found, and it is `none`;
5. `Ok((Some(Smart::Custom(Some(path))), Some(data)))` indicates argument `theme` found, and it is `str` (path to the theme file).

### CODE REVIEW

File: crates/typst/src/text/raw.rs [hunk 1]
```diff
@@ -852,6 +862,13 @@ pub static RAW_THEME: Lazy<synt::Theme> = Lazy::new(|| synt::Theme {
     ],
 });
 
+/// The theme used to disable syntax highlighting
+pub static RAW_THEME_NONE: Lazy<synt::Theme> = Lazy::new(|| synt::Theme {
```
laurmaedje, line 866: Maybe instead having a "nothing theme", we should rather just not highlight in the first place?

DrGo: Agree. In a big book there will be many instances of raw elements, so this is not a case of premature optimization.

Coekjan: Agree. I have fixed it. Now it will early exit when `raw.theme` is `none`.

crates/typst/src/text/raw.rs
```diff
@@ -30,6 +30,7 @@ use crate::{syntax, World};
 type StyleFn<'a> =
     &'a mut dyn FnMut(usize, &LinkedNode, Range<usize>, synt::Style) -> Content;
 type LineFn<'a> = &'a mut dyn FnMut(usize, Range<usize>, &mut Vec<Content>);
+type ThemeArgType = Smart<Option<EcoString>>;
```
DrGo, line 33: You might just want to mention in the docs that setting the `theme=none` disables syntax highlighting.

DrGo (commented):
Thanks.. looks good to me

crates/typst/src/text/raw.rs
Coekjan: Thanks. I have mentioned it in the docs.

crates/typst/src/text/raw.rs
```diff
@@ -231,12 +232,22 @@ pub struct RawElem {
...
+    ///
+    /// Additionally, you can set the theme to `none` to disable highlighting.
+    ///
+    /// ````example
+    /// #set raw(theme: none)
+    ///
+    /// ```typ
+    /// #let this = "is not highlighted"
+    /// ```
+    /// ````
```
laurmaedje, line 244: While it'd be nice to have an example, unfortunately parameters can have only one example (which is by default hidden in a drop down). But I think people can imagine how it looks without highlighting, so I guess it's not critical to have this.

tests/suite/text/raw.typ
```diff
@@ -648,6 +648,22 @@ fn main() {
...
+
+#set raw(theme: auto)
+```typ
+#let foo = "bar"
+```
+
+#set raw(theme: none)
+```typ
+#let foo = "bar"
+```
```
laurmaedje, line 665: Since auto is already tested above, it suffices to have just the first of these three here, in my opinion.

laurmaedje:
Thanks!

Issue #4232, opened 2024-05-23
Allow `measure` to accept available size


MDLC01:
### Description

With this proposal, `measure` would accept an optional `available` dictionary with `width` and `height` keys, specifying available width and height respectively. If `available` is specified, the content passed to measure is laid out without having infinite available space, but instead with restricted space, as defined by `available`.

### Use Case

When you have, e.g., a paragraph, you might want to know its dimensions when it is laid out on the page, instead the dimensions it would have on an infinite page.

There are workaround currently, but for more complex use cases, they quickly become very ugly.

Myriad-Dreamin:
You may try this:

```
#context measure(lorem(100)) \
#context measure(block(lorem(100), width: 400pt))
```

![image](https://github.com/typst/typst/assets/35292584/3e70850d-7a3b-485e-9402-e00a491a5597)

MDLC01:
Yes, but this only works if you are trying to measure the height with given an available width. I don't think there are workarounds as simple as this one for other cases, such as measuring the width given an available width.

laurmaedje:
Fixed by https://github.com/typst/typst/pull/4264

PR #4264, opened 2024-05-26
Allow `measure` to accept available size

bluebear94:
See #4232.

Use separate width and height parameters instead of a single dictionary in order to allow setting only one of them.


tests/suite/layout/measure.typ
```diff
@@ -7,3 +7,10 @@
 }
 #text(10pt, f(6pt, 8pt))
 #text(20pt, f(13pt, 14pt))
+
+--- measure-given-area ---
+// Test `measure` given an area.
+#let text = lorem(100)
+
+#context measure(text)
+#context measure(width: 400pt, height: auto, text)
```
EpicEricEE, line 16: This could probably be made into a compile-only test without an output image, similar to the test case above

crates/typst/src/layout/measure.rs
```diff
@@ -60,7 +73,12 @@ pub fn measure(
         None => context.styles().at(span)?,
     };
 
-    let pod = Regions::one(Axes::splat(Abs::inf()), Axes::splat(false));
+    let available = Axes::new(
+        width.resolve(context.styles().at(span)?).unwrap_or(Abs::inf()),
+        height.resolve(context.styles().at(span)?).unwrap_or(Abs::inf()),
```
laurmaedje, line 78: Why not just `styles` instead of `context.styles().at(span)?`?

crates/typst/src/layout/measure.rs
```diff
@@ -2,17 +2,18 @@ use comemo::Tracked;
...
 ///
-/// The `measure` function lets you determine the layouted size of content. Note
-/// that an infinite space is assumed, therefore the measured height/width may
-/// not necessarily match the final height/width of the measured content. If you
-/// want to measure in the current layout dimensions, you can combine `measure`
-/// and [`layout`].
+/// The `measure` function lets you determine the layouted size of content.
+/// By default an infinite space is assumed, so the measured dimensions may
+/// not necessarily match the final dimensions of the content.
+/// If you want to measure in the current layout dimensions, you can combine `measure` and [`layout`].
```
laurmaedje, line 16: please wrap this at 80 columns

crates/typst/src/layout/measure.rs
```diff
@@ -48,6 +49,18 @@ pub fn measure(
     context: Tracked<Context>,
     /// The callsite span.
     span: Span,
+    /// The width available to layout the content.
+    ///
```
laurmaedje, line 53: Maybe we should have an example showcasing the difference here?

crates/typst/src/layout/measure.rs
bluebear94: Good point. Didn’t notice that variable, so now I’ve changed the code to use it.

crates/typst/src/layout/measure.rs
bluebear94: I’ve added an example. How does it look?

laurmaedje:
Thanks!

laurmaedje:
Fixes #4232

PR #4267, opened 2024-05-27
Add named symbols: `⌈`, `⌉`, `⌊`, `⌋`

### LINKED ISSUE #4265: floor and ceiling symbols missing from `sym`
### Description

I expected to find floor and ceiling symbols in the `sym` module.

### Use Case

while the symbols can be generated with `floor(x)` and `ceil(x)` it would be useful to access them directly via `sym`.
e.g. `floor.l   floor.r   ceil.l   ceil.r`

Coekjan:
This pr closes #4265 .

```typ
#grid(
  columns: (1fr, 1fr),
  align: (start + horizon, end + horizon),
  align(center, table(
    columns: 4,
    align: center + horizon,
    table.cell(colspan: 2)[In Markup Mode], table.cell(colspan: 2)[In Math Mode],
    ```typst #sym.floor```, sym.floor, ```typst $floor$```, $floor$,
    ```typst #sym.ceil```, sym.ceil, ```typst $ceil$```, $ceil$,
    ```typst #sym.floor.l```, sym.floor.l, ```typst $floor.l$```, $floor.l$,
    ```typst #sym.floor.r```, sym.floor.r, ```typst $floor.r$```, $floor.r$,
    ```typst #sym.ceil.l```, sym.ceil.l, ```typst $ceil.l$```, $ceil.l$,
    ```typst #sym.ceil.r```, sym.ceil.r, ```typst $ceil.r$```, $ceil.r$,
  )),
  [
    $ round(A) $
    $ floor.l A ceil.r $
    $ ceil.l A floor.r $
    $ floor.l A floor.r = floor.l(A) = floor(A) $
    $ ceil.l A ceil.r = ceil.l(A) = ceil(A) $
  ],
)
```

image

### DIFF
--- a/crates/typst/src/math/mod.rs
+++ b/crates/typst/src/math/mod.rs
@@ -187,8 +187,6 @@ pub fn module() -> Module {
     math.define_elem::<PrimesElem>();
     math.define_func::<abs>();
     math.define_func::<norm>();
-    math.define_func::<floor>();
-    math.define_func::<ceil>();
     math.define_func::<round>();
     math.define_func::<sqrt>();
     math.define_func::<upright>();
--- a/crates/typst/src/symbols/sym.rs
+++ b/crates/typst/src/symbols/sym.rs
@@ -64,6 +64,14 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
         spheric.rev: '⦠',
         spheric.top: '⦡',
     ],
+    ceil: [
+        #[call(crate::math::ceil)] l: '⌈',
+        r: '⌉',
+    ],
+    floor: [
+        #[call(crate::math::floor)] l: '⌊',
+        r: '⌋',
+    ],
 
     // Punctuation.
     amp: ['&', inv: '⅋'],
--- a/docs/src/lib.rs
+++ b/docs/src/lib.rs
@@ -542,7 +542,7 @@ fn group_page(
     let mut outline_items = vec![];
     for name in &group.filter {
         let value = group.module().scope().get(name).unwrap();
-        let Value::Func(func) = value else { panic!("not a function") };
+        let Ok(ref func) = value.clone().cast::<Func>() else { panic!("not a function") };
         let func = func_model(resolver, func, &path, true);
         let id_base = urlify(&eco_format!("functions-{}", func.name));
         let children = func_outline(&func, &id_base);


laurmaedje:
I prefer the names `floor.l` and `floor.r`. To use these without conflict we would need to generalize the callable symbols from just accents to something more generic.

Coekjan:
> I prefer the names `floor.l` and `floor.r`. To use these without conflict we would need to generalize the callable symbols from just accents to something more generic.

If I got it correctly, we need to distinguish different objects from the same name depending on their different types.

laurmaedje:
What I meant is that symbols could get a generic way to be callable just like functions and we would put the current floor function implementation into the floor symbol.

Coekjan:
> What I meant is that symbols could get a generic way to be callable just like functions and we would put the current floor function implementation into the floor symbol.

Thanks! I got it. I will try to implement it in another PR and mark this one as draft.

Coekjan:
> symbols could get a generic way to be callable just like functions

Before implementation, some details I wonder:
1. Shall we get **all** symbols callable? Like `floor.l()`, `floor.r()`. Or we just get some specific symbols callable? Like `floor()`.
2. If we define `floor` in `SYM` table, shall `floor` (without paren) be allowed? Or we just allow `floor()`.

PgBiel:
My opinions:

> 1. Shall we get **all** symbols callable? Like `floor.l()`, `floor.r()`. Or we just get some specific symbols callable? Like `floor()`.

No, just a few ones. For `floor()` it makes sense due to how it surrounds the content. For accents like `arrow()`, it makes sense as they appear above the content. But I don't think this would apply to e.g. `square`, `gradient` and so on.

> 2. If we define `floor` in `SYM` table, shall `floor` (without paren) be allowed? Or we just allow `floor()`.

This is something I wondered as well, because naturally `floor` (without paren)  would then correspond to `floor.l` if we added the symbols. But IMHO `floor` by itself shouldn't really correspond to anything, maybe just returns the "symbol function".

> 3. What syntax would you like to implement function in (or out of) the `SYM` table?

I don't have strong opinions here, but here are two possible approaches I could think of:

1. Move "functional symbols" out of the symbol table and use a separate macro to declare them in the compiler's source code. This, however, might make it harder or less consistent to declare "sub-symbols".
2. Add a way to specify a function's name in the symbol table in place of the symbol's string contents (could be some special notation or keyword for it). I think this might compose better with the rest of symbols.

Coekjan:
I see current `floor` function is defined only for math mode. If we merge its definition into symbol table, shall we allow `floor` be called in markup mode?

Coekjan:
> I see current `floor` function is defined only for math mode. If we merge its definition into symbol table, shall we allow `floor` be called in markup mode?

Any opinions about this?

crates/typst/src/eval/call.rs
```diff
@@ -129,6 +129,18 @@ impl Eval for ast::FuncCall<'_> {
...
+            Value::Symbol(sym) => {
+                if let Some(handler) = sym.func() {
+                    Value::Func(handler)
+                } else {
+                    Value::Symbol(sym)
+                }
+            }
+            callee => callee,
+        };
+
```
wrzian, line 143: Seeing this file changed, I'd like to note that I think the `Eval<FuncCall>::eval` function is really ugly in general, and deserves an overhaul. I'd be interested in discussing this if @laurmaedje agrees. I think splitting out the massive `let (...) = if { /* huge mess of 80+ lines */ } else {/* one single line */}` into a separate function would be a nice start, but there are a few early returns in the huge mess that make it complicated in the context of the overall function. (This is also somewhat relevant for the new VM refactor, since it completely redoes the func call system)

Coekjan: Agree. When I saw this function, it was a bit hard to understand the whole behavior of it.

And I would suggest sending another PR to refactor it (let every PR do its own job).

Coekjan: By the way, this change actually originates from #4299 on which this PR depends.

laurmaedje: @wrzian I agree as well. Not my proudest work 😂

Enivex:
Is this one finished?

laurmaedje:
Oh, now I see why you did this in the first place. I'll check whether your change had the intended effect on the rendered docs.

Coekjan:
> Oh, now I see why you did this in the first place. I'll check whether your change had the intended effect on the rendered docs.

Would you like me to reapply the change, or will you handle it?

laurmaedje:
Probably it will work, so feel free to make the change and if it turns out that it doesn't work, I'll come up with something when I test it.

laurmaedje:
Docs look good. Thanks!

Issue #4191, opened 2024-05-19
Allow `scale` to tolerate absolute lengths


kram-tm:
### Description

Currently, the `scale` function only supports values of the type `ratio`, making it somewhat difficult to resize content to an absolute width/height such as that of the current page or container. This is possible in LaTeX using `\resizebox{width}{height}{Foo}` (for possible reference, not to add any pressure) and can currently be circumvented in Typst by "norming" the size of the content to be scaled and multiplying it by a given length as in:
```typ
#layout(container => {
  let contents = $a^2 = b^2 + c^2$
  let contents-dimensions = measure(contents)
  let scale-ratio = container.width / contents-dimensions.width * 100%
  scale(reflow: true, x: scale-ratio, y: scale-ratio, contents)
})
```
![image](https://github.com/typst/typst/assets/34284981/3394a958-8de3-4346-8540-d41f78c393c5)

### Use Case

While I don't like tautologies, I'm feeling the urge to simply say that this would make scaling to an absolute length easier, which in turn would facilitate the use case presented above among others without needing to resort to using math explicitly. Admittedly, the uses of this are limited by the fact that most people resize images, where absolute lengths are indeed valid; however, I've seen a question from three different people on the Discord server about resizing text, a fletcher diagram as well as an equation to fit exactly in a page. This may not be a lot in terms of demand, but is indicative of utility.

bluebear94:
Would it also make sense for `scale` to allow at most one of `x` and `y` to be `auto` to indicate that that axis should be scaled with the same ratio as the other?

PR #4271, opened 2024-05-27
Allow absolute lengths in `scale`

bluebear94:
See #4191.

This is a draft PR because the code is messy. In particular:

* It’s already more code than the rest of the `ScaleElem`’s `LayoutSingle` implementation, and I’m not sure if this is worth the additional complexity.
* If an absolute length is used, then we have to measure the content. This is at least done lazily.
* `scale` with `reflow: true` behaves differently if either `page.width` or `page.height` is `auto`. This problem already exists and arises from `Ratio::of` returning zero if the result would not be finite.
* It would make sense for `scale.x` and `scale.y` to default to `auto` now, but I decided not to do this because it’s a breaking change.


laurmaedje:
> It’s already more code than the rest of the ScaleElem’s LayoutSingle implementation, and I’m not sure if this is worth the additional complexity.

I think the feature is cool, but we should definitely try to simplify the implementation as much as possible.

> If an absolute length is used, then we have to measure the content. This is at least done lazily.

What I'm wondering is: Do we measure twice now sometimes? Because we measure while resolving the ratio and then again in `measure_and_layout`?

> scale with reflow: true behaves differently if either page.width or page.height is auto. This problem already exists and arises from Ratio::of returning zero if the result would not be finite.

What's the exact problem here? Can we fix it independently of this?

> It would make sense for scale.x and scale.y to default to auto now, but I decided not to do this because it’s a breaking change.

We can break that at the same time where we change `reflow`'s default to `true`.

bluebear94:
> What I'm wondering is: Do we measure twice now sometimes? Because we measure while resolving the ratio and then again in `measure_and_layout`?

Yes. We also use different values for the `regions` parameter’s size each time, so reusing the calculation might pose a problem.

> > scale with reflow: true behaves differently if either page.width or page.height is auto. This problem already exists and arises from Ratio::of returning zero if the result would not be finite.
>
> What's the exact problem here? Can we fix it independently of this?

With `reflow: true` and `page.width` or `height: auto`, the body is measured in a region with zero width or height, leading to the measured dimensions having zero width or height, even though the body might not really have these dimensions.

Here’s an example in Typst 0.11.1:

```typst
#let aaa = {
  set text(size: 28pt)
  let scaled(body) = box(stroke: 0.5pt + red, scale(
    x: 60%,
    y: 40%,
    box(stroke: 0.5pt, clip: false, body)
  ))

  set scale(reflow: false)
  [Hello #scaled[World]!\ ]

  set scale(reflow: true)
  [Hello #scaled[World]!]
}

#set page(width: 200pt, height: 100pt)
#aaa

#set page(width: 200pt, height: auto)
#aaa

#set page(width: auto, height: 100pt)
#aaa

#aaa
```

which results in these four pages:

| Page size | Result |
|-|-|
| `(width: 200pt, height: 100pt)` | Page 1 |
| `page(width: 200pt, height: auto)` | Page 2 |
| `page(width: auto, height: 100pt)` | Page 3 |
| `#set page(width: auto, height: auto)` | Page 4 |

laurmaedje:
> With reflow: true and page.width or height: auto, the body is measured in a region with zero width or height, leading to the measured dimensions having zero width or height, even though the body might not really have these dimensions.

Okay, that sounds like a problem that's fixable separately. We should probably rather measure in an infinite space?

Other than that, how should we proceed with this PR? I think the feature is desirable in principle. I agree that the code could be a bit nicer, but it's also not _that_ messy.

bluebear94:
> Other than that, how should we proceed with this PR? I think the feature is desirable in principle. I agree that the code could be a bit nicer, but it's also not _that_ messy.

I guess I agree with that.

laurmaedje:
I tried to merge with main, but the reference image changed. (The first scaled cylinder got smaller.) Not sure what the cause is.

bluebear94:
> I tried to merge with main, but the reference image changed. (The first scaled cylinder got smaller.) Not sure what the cause is.

I rebased the branch onto a10e332, and the tests passed. When I rebased to 20b8d2c, I got the same test failure, so it must be caused by #4354.

laurmaedje:
Okay, it's probably fine then. You can probably just accept the change.

laurmaedje:
Thank you!

istudyatuni:
I think linked issue could be closed?

Issue #3882, opened 2024-04-05
Support for Greek numbering


rousbound:
### Description

Add the option "g" and "G" to numbering with greek letters:
```typst
set enum(numbering:"g)")
+ first
+ second
+ third
(...)
```

Would render to
```
sym.alpha first
sym.beta second
sym.gamma third
(...)
```

The same for uppercase.

### Use Case

When you have high levels of identation, we "only" have 1, a, A, i, I and some more "paragraph" like characters like sym.pilcrow and sym.section. It would be cool to have greek letters as an option, specially on deep nested documents to avoid repetition of the numbering types mentioned above.

I could create a package to create a wrapper for numbering for such a thing, but I think that it could be desirable and trivial to have the numbering pattern on the core Typst itself.

Edit:
Having in mind that western thinking is highly rooted in greek culture, I think this feature can be desirable for other language speaking users aside from greek ones too.

Enivex:
Surely a Greek letter should be used if this is added, not `g` and `G`?

Alpha and beta are out, since the capital letters are indistinguishable from their latin counterparts. Gamma could work.

rousbound:
You mean to use the greek numbering pattern in some way like this:
```typst
#set enum(numbering:"Gamma.i.1")

#set enum(numbering:"gamma.i.1")
```

Instead of:

```typst
#set enum(numbering:"G.i.1")
#set enum(numbering:"g.i.1")
```

I see your point, but if we are going to be more verbose then I think maybe its better like this:

```typst
#set enum(numbering:"Greek.i.1")
#set enum(numbering:"greek.i.1")
````

**Comment by Enivex** [comment]
No, I mean the actual Unicode symbols γ and Γ.

**Comment by Leedehai** [comment]
I prefer γ and Γ for greek counting numbers because (1) it doesn't take the place of "g"/"G" and (2) it aligns with precedents.

**Comment by rousbound** [comment]
Oh, ok, then I think you mean:
```typst
#set enum(numbering:"Γ.i.1")
#set enum(numbering:"γ.i.1")
```
The thing that crosses my mind is:
I would need to copy an paste a unicode symbol from docs/google or memorise unicode code and use the Ctrl+Shift+u method, right?

I'm not very used to this so it feels a lit bit too much of extra work at first sight.

But I see the point of "(2) aligns with precendent".

"(1) doesn't take the place of 'g/G'" I'm not sure why that could be a problem in the future, can you think of an example?

Enivex:
[...]

Issue #1177, opened 2023-05-11
[Feature Request] Escaping interpreted characters in numbering patterns


edhebi:
### Description

A number of characters are interpreted in numbering patterns (since that's the whole point). Because there's no escape hatch, this makes them impossible to use in the various prefix of suffix. I propose that a backlash character should stop the next character from being interpreted in this way. For example a patter of `"Section 1.1 )"` would be written `"Sect\\ion 1.1 )"`

I can see two limitations:

* It's not obvious that you need a double backslash instead of of a single one to someone not familiar with escape sequences, and that makes an actual backslash be `"\\\\"` but I suspect that's not a common character at all to put in there.
* It's added complexity for something that can currently be solved with numbering functions. I do think this makes a bunch of patterns nicer for the user on the long run tho.

I'd actually be willing to implement this if there's interested, that'd be a good way to get familiar with the codebase :p

### Use Case

For more involved numbering patterns, it can get frustrating to work around interpreted characters, especially `aAiI`. For example, `numbering("Section 1.1 )")` doesn't produce the expected result because the `i` is a special character in a numbering pattern.

With this feature it would be `numbering("Sect\\ion  1.1 )")` instead.

```c
// Before
#set heading(numbering: (..nums) => "Section " + numbering("1.1 )", ..nums.pos()))
// After
#set heading(numbering: "Sect\\ion 1.1 )")
```

laurmaedje:
This is certainly an issue although I think having "Section" be part of the numbering is misuse. The "Section" part is a supplement and should be handled separately. 

I am not really convinced by the backslash solution though. It doesn't fix the problem that we cannot add new numbering kinds without it being a breaking change and, to be honest, it's just not particularly nice. 

A clean solution would of course be to "escape" the actual numberings (i.e. `"{1} / {1}"` or maybe with just a single letter), but that would make normal simple patterns so complicated. I am unsure how to proceed here.

We also have similar patterns for dates soon and there has been discussion about patterns for citations. We should find a general solution for this that is consistent across all these usages.

edhebi:
supplement aren't really related to how the header/thing is displayed tho, if I'm understanding correctly. Right now, displaying something *before* the number in a heading requires a number prefix. My example of the word "Section" was a bit weird I'll admit; but being to reuse headings for other things in documents is useful. I had the case of doing homework sheets and trying to number the things "Exercice n -- [the heading here]".

[...]

PR #4273, opened 2024-05-28
Lucy numbering br

LuxxxLucy:
Regarding Issue #3882,

I mainly took references from [Greek Number Converter](https://www.russellcottrell.com/greek/utilities/GreekNumberConverter.htm) and [Greek Number](https://www.unco.edu/fraternity-sorority/resources/greek-alphabet.aspx)

I am for now using the lowercase and upper case Alpha as the symbol.

Some test cases (unicode level) are added in `numbering.rs`
```
    greek_number_tests! {
...
        long_complex_0: (97_554, "αΜθʹ, ͵ζφνδ", Case::Lower),
        long_complex_1: (2_056_839_184, "βΜκʹ, αΜ͵εχπγ, ͵θρπδ", Case::Lower),
        long_complex_2: (12_312_398_676, "βΜρκγʹ, αΜ͵ασλθ, ͵ηχοϛ", Case::Lower),
...
    }
```

Png rendered 
image


JamesxX:
Would it not make more sense for the numbering to cover the entire alphabet rather than the first 10 characters? Base 26 (or however many characters greek has) rather than base 10

LuxxxLucy:
> Would it not make more sense for the numbering to cover the entire alphabet rather than the first 10 characters? Base 26 (or however many characters greek has) rather than base 10

Hi @JamesxX I think the main issue is definition on what is the Greek numeral system. The thing you just mentioned is called [lower-greek](https://w3c.github.io/predefined-counter-styles/#lower-greek) which represent 11 as λ (the 11-th char of the alphabet), while the greek numbering system is a decimal-like system (though with a different set of  chars), 11 is represented by "ιαʹ" (where "ι" is 10 and "α" is 1).

it seems to me that both should be implemented if permited.

tests/suite/model/numbering.typ
```diff
@@ -24,6 +24,25 @@
   [ for #i \ ]
 }
 
+--- numbering-greek ---
```
laurmaedje, line 27: New numbering tests should not use reference images anymore, but rather assertions. The old ones should also be updated some time.

crates/typst/src/model/numbering.rs
```diff
@@ -208,8 +208,11 @@ impl FromStr for NumberingPattern {
             };
 
             let prefix = pattern[handled..i].into();
-            let case =
-                if c.is_uppercase() || c == '壹' { Case::Upper } else { Case::Lower };
+            let case = if c.is_uppercase() || c == '壹' || c == 'Α' {
```
laurmaedje, line 211: why `c == 'Α'`? isn't `c.is_uppercase()` true for it?

probably we need to update `c.to_ascii_lowercase()` above since we have more than ASCII here

laurmaedje (commented):
Here are a few comments, but not really relevant ones, since I can't judge the correctness of the Greek numbering itself. And then there is of course also still the question of the letter etc.

I'm not sure how best to proceed here.

crates/typst/src/model/numbering.rs
LuxxxLucy: > why `c == 'Α'`? isn't `c.is_uppercase()` true for it?

Oh that is true; we do not need to do this. Updated.

> probably we need to update `c.to_ascii_lowercase()` above since we have more than ASCII here

Changed to use general to_lowercase()

tests/suite/model/numbering.typ
LuxxxLucy: Update the test case, now only assertions, not sure whether it is what you want it to be. If so, do you want all the numbering test to be change to assertions in the numbering.type file like this?

(I previously thought that the typ and ref png should be the test on rendering results, and because of so for the numbering I added some local test to `numbering.rs` where the function is defined.)

laurmaedje: >  If so, do you want all the numbering test to be change to assertions in the numbering.type file like this?

In principle yes, but that can be done separately. But open a PR if you'd like!

> I previously thought that the typ and ref png should be the test on rendering results, and because of so for the numbering I added some local test to numbering.rs where the function is defined.

I think the tests should just be where it's most convenient. If it's easier to test in Typst code, test it there, if it's simpler to test in Rust, test it there. This applies to the logic itself. The interface must be tested in Typst code to ensure that it works in integration.

laurmaedje:
I merged with main and tidied up the tests. While doing that I noticed that the number 2000000001 comes out differently with your code than with the [online converter](https://www.russellcottrell.com/greek/utilities/GreekNumberConverter.htm):
- Your code: "βΜκʹ, αΜαʹ
- Online converter: βΜκʹ, αʹ

laurmaedje:
Regarding the letter. It think we should use `α` and `Α`. Yes, the uppercase variant looks just like a normal A, but it is the only consistent choice. You can always write `sym.Alpha` and I also like the idea (forget where it was presented) to allow long-form numberings like `{greek}` and as soon as there is a brace-style numbering, implicit numbering is disabled.

LuxxxLucy:
> I merged with main and tidied up the tests. While doing that I noticed that the number 2000000001 comes out differently with your code than with the [online converter](https://www.russellcottrell.com/greek/utilities/GreekNumberConverter.htm):
> 
> * Your code: "βΜκʹ, αΜαʹ
> * Online converter: βΜκʹ, αʹ

Sorry, I did not realize it is so long ago and almost forgot this PR. I have just updated the PR with a fix and added more test cases.

> Regarding the letter. It think we should use `α` and `Α`.

the current way should be this case, using the `α` and `Α`.

>  I also like the idea (forget where it was presented) to allow long-form numberings like {greek} 

I like this idea as well!  I am interested it supporting it but do not know how to start with it, do you have some hint/high level ideas on how to do it? that would be great.

laurmaedje:
> I have just updated the PR with a fix and added more test cases.

Thanks!

> the current way should be this case, using the α and Α.

Yep.

> I like this idea as well! I am interested it supporting it but do not know how to start with it, do you have some hint/high level ideas on how to do it? that would be great.

Responded here: https://github.com/typst/typst/issues/1177#issuecomment-2451572160

laurmaedje:
Thank you!

Issue #2634, opened 2023-11-09
Zero argument functions without parentheses in math mode


yaksher:
### Description

Treat the name of a function in math mode as an invocation with no arguments (erroring if that is not valid for the function in question).

This would not extend to code mode, where referring to the function itself as a value happens often. In math mode, though, referring to a function directly simply produces a useless rendering of the function itself.

### Use Case

Would allow users to accomplish things similar to the `tilde` and `hat` functions, which can be symbols or accent functions, themselves.

It seems unlikely to cause any new confusion or bugs because it's functionality that already exists in the built in functions and functions people write themselves will, in general, not be valid without arguments.

Leedehai:
I'd like to caution against having diverging syntax behavior in math vs code modes, though it would be certainly better if there's a concrete use case or user pain point to demonstrate the benefits.

yaksher:
This isn't new divergent behavior though. Function call behavior is already distinct between math and code mode (no square bracket content calls and I think another thing?) and, more importantly, this behavior already exists in special cases. I think generalizing behavior makes it less confusing than having the special case of accent symbols behaving that way while everything else doesn't.

laurmaedje:
Symbols aren't functions though. The fact that symbols are callable doesn't mean that function should be implicitly callable. There is no implicit calling in Typst so far.

yaksher:
There are things which can either be used as function calls or be used as constants; the fact that they're technically not functions when they're used on their own under the hood or whatever does not seem like it should be a meaningful distinction to me, except insofar as `tilde()` is invalid rather than equivalent to `tilde`.

I don't think this would add any confusion since:
- functions are pure, so a function of no arguments will always be the same thing, there's no tricky behavior with the same "constant" changing appearance, unless they use state, which can create constants that have variable behavior already (and in fact, invisibly call a function in a constant under the hood through `State::display`)
- most user-defined functions are not valid to be called with 0 arguments, and the error message could be made to be whatever the current one is rather than not enough arguments or whatever. If people are doing advanced enough stuff to use `..args` and accidentally write it without arguments and instead of an error it produces the 0 argument output, I feel like they can probably figure it out, especially if this is universal and documented behavior of math mode/functions
- again, in terms of experience of use, this behavior already exists in some symbols. Even if symbols are technically "not functions", I feel like if something is used with function call syntax and does something that can't be implemented without a function by a user, it's fair to say it's functionally (hah) a function.

[...]

Issue #422, opened 2023-03-29
Fake accents


Enivex:
Currently https://typst.app/docs/reference/math/accent/ will only display correctly if supplied with one of the actually supported accents.

In LaTeX I use https://ctan.org/pkg/accents?lang=en, which can be used to typeset more or less anything as an accent. It would be useful if Typst's accent function would revert to similar behavior when supplied with something other than the few supported symbols.

user202729:
You can already stack the boxes manually, but I don't know how to do that reliably while correctly handle the spacing.

alerque:
Don't go down this road. The approach LaTeX used is a bandaid on a hack. The correct technology exists today: OpenType fonts can specify the placement of accents on characters in a robust and correct way. If you want the support, use a font that supports it. The typesetter should not be responsible for faking it. The low level tools can be provided to fake something up, but it should not be provided as a general purpose tool to handle accented characters.

Enivex:
> Don't go down this road. The approach LaTeX used is a bandaid on a hack. The correct technology exists today: OpenType fonts can specify the placement of accents on characters in a robust and correct way. If you want the support, use a font that supports it. The typesetter should not be responsible for faking it. The low level tools can be provided to fake something up, but it should not be provided as a general purpose tool to handle accented characters.

I think you misunderstand. The point is that even symbols that aren't normally used as accents can be used as them. Obviously regular accents wouldn't be faked.

Here's an example:

![image](https://user-images.githubusercontent.com/2191754/228509013-b354fae8-185e-4b28-a84e-f38a0c033cd9.png)

Enivex:
> You can already stack the boxes manually, but I don't know how to do that reliably while correctly handle the spacing.

Yes, that's the hard part. I've tried playing around with it, but since I lack the understanding of Typst's intracicies when it comes to spacing and positioning I couldn't make it work properly.

PR #4299, opened 2024-05-30
Make symbols callable like functions & migrate callable accents to callable symbols

Coekjan:
## Introduction

This pr makes symbols callable just like functions. https://github.com/typst/typst/pull/4267#issuecomment-2133469679

Briefly, this pr allows developers to specify the handler function of a symbol in the symbol table, so that the symbol can be called like a function and the handler function will be called.

Additionally, this pr migrates existing callable accents to callable symbols.

### To Make Top Level Symbols Callable

If the symbol only has itself as the top level symbol:

```rs
floor: #[call(crate::math::floor)] '⌊',
```

image

Or the symbol has lots of variants:

```rs
floor: [
    #[call(crate::math::floor)] '⌊',
    r: '⌋',
],
```

image

### To Make Symbol Variants Callable

```rs
floor: [
    #[call(crate::math::floor)] l: '⌊',
    r: '⌋',
],
```

image


Enivex:
@Coekjan is it only the top level that can be callable?

As in, could I conceivably make `brace.b` callable for instance?

Coekjan:
> @Coekjan is it only the top level that can be callable?

Yes.

> As in, could I conceivably make `brace.b` callable for instance?

Currently it doesn't support that.

Enivex:
> > @Coekjan is it only the top level that can be callable?
> 
> Yes.
> 
> > As in, could I conceivably make `brace.b` callable for instance?
> 
> Currently it doesn't support that.

@laurmaedje suggested to replace `underbrace` with a callable `brace.b` etc. That would require a more general solution than this PR.

Coekjan:
Ok, I will try to implement it in this PR.

Do we have any other needs about this feature?

laurmaedje:
> @laurmaedje suggested to replace underbrace with a callable brace.b etc. That would require a more general solution than this PR.

`$ arrow.l(x) $` already exists. this should be similar.

Enivex:
> > @laurmaedje suggested to replace underbrace with a callable brace.b etc. That would require a more general solution than this PR.
> 
> `$ arrow.l(x) $` already exists. this should be similar.

I'm confused. What is currently missing then?

laurmaedje:
`brace.b` is not marked as an accent, so it doesn't get the behaviour.

Enivex:
Oh okay, so it's specifically that it's possible for accents but not general symbols

Leedehai:
(No comment, but I'd like to mention it appears this PR fixes #2634)

Coekjan:
I updated the implementation just now, you can see its behavior via some examples in the PR description.

Enivex:
> (No comment, but I'd like to mention it appears this PR fixes #2634)

The question is whether that is desirable

Coekjan:
Hi everyone, do you like current implementation? Please send your comments if I didn't do it well enough :open_mouth:

laurmaedje:
Sorry for the slow response -- I'm quite busy with working on the compiler myself at the moment.

One thing that I would like to see us at least try is to somehow also use this infrastructure for the existing callable accent symbols. Currently, it is handled in `eval/call.rs`, right below the code you added.

@Enivex I would assume that https://github.com/typst/typst/issues/422 doesn't really affect this, right? Since it would only apply to `accent`, not to the call shorthand.

I'm also not sure whether the docs generation works as implement right now, but that's details and also hard to check for you, so I would just check that before merging.

Enivex:
> Sorry for the slow response -- I'm quite busy with working on the compiler myself at the moment.
> 
> One thing that I would like to see us at least try is to somehow also use this infrastructure for the existing callable accent symbols. Currently, it is handled in `eval/call.rs`, right below the code you added.
> 
> @Enivex I would assume that https://github.com/typst/typst/issues/422 doesn't really affect this, right? Since it would only apply to `accent`, not to the call shorthand.
> 
> I'm also not sure whether the docs generation works as implement right now, but that's details and also hard to check for you, so I would just check that before merging.

It shouldn't affect #422 no.

Coekjan:
> Sorry for the slow response -- I'm quite busy with working on the compiler myself at the moment.

Thanks for your reply. I completely understand and appreciate the hard work you're putting into this open-source project :heart: 

> One thing that I would like to see us at least try is to somehow also use this infrastructure for the existing callable accent symbols.

In my evaluation of the existing infrastructure for callable accent symbols, I have identified a couple of issues that need to be addressed to effectively integrate this feature:
1. My current implementation does not support the invocation of `NativeElement`. This is a capability that _I plan to develop later_.
2. Attempting to directly invoke `AccentElem` within the `SYM` table, as in the example `arrow: [#[call(crate::math::AccentElem)] self, ],`, has proven unsuccessful. This is due to the omission of the second argument, `accent`, required by `AccentElem`. At this juncture, I still have no idea to deal with it (**need help** :cry: ).

> I'm also not sure whether the docs generation works as implement right now, but that's details and also hard to check for you, so I would just check that before merging.

Concerning the docs generation, I acknowledge that this is an area I am not very familiar with, so I would appreciate it if you could check that aspect before we proceed with the merge.

laurmaedje:
> This is due to the omission of the second argument, accent, required by AccentElem. 

What if we have an inline closure with the correct accent symbol in `#[call(..)]`?

Coekjan:
> What if we have an inline closure with the correct accent symbol in `#[call(..)]`?

Thanks for your idea. Do you mean we inline a closure like this?

```rs
arrow: [
    #[call(|| {
        let func: Func = AccentElem::elem().into();
        func.bind_2nd_arg(Accent::new('→')) // <--- ?
    })]
    self,
],
```

If so, we may need a general function to bind _k_-th arg for `Func`.

laurmaedje:
yeah something like this. not sure whether we should opt for a generic binding mechanism or just some manual (possibly macro-assisted) `#[func]` definitions.

Coekjan:
> yeah something like this. not sure whether we should opt for a generic binding mechanism or just some manual (possibly macro-assisted) `#[func]` definitions.

I think manual `#[func]` definitions could be enough, because:
1. the number of accent symbols is small
2. a generic binding mechanism might introduce big changes in `func.rs`

Coekjan:
Hi, I have migrated the callable accents to callable symbols:
1. The 1st commit makes symbols callable;
2. The 2nd commit migrates callable accents.

Coekjan:
Notice that current implementation may enable accessing accents (which is only accessible in math mode before) via symbols, e.g. `#sym.arrow(A)`

crates/typst-macros/src/lib.rs
```diff
@@ -340,17 +340,45 @@ pub fn derive_cast(item: BoundaryStream) -> BoundaryStream {
...
+///     acute: [
+///         #[call(crate::math::acute)] '´',
+///         double: '˝',
+///     ],
+///
+///     // A callable symbol where each variant has a modifier.
+///     arrow: [
+///         #[call(crate::math::arrow)] r: '→',
+///         r.long.bar: '⟼',
+///         #[call(crate::math::arrow)] l: '←',
```
laurmaedje, line 371:
```suggestion
///         #[call(crate::math::arrow_l)] l: '←',
```

crates/typst/src/eval/call.rs
```diff
@@ -129,23 +128,20 @@ impl Eval for ast::FuncCall<'_> {
...
+                    Value::Func(handler)
+                } else {
+                    Value::Symbol(sym)
                 }
             }
+            callee => callee,
+        };
+
+        // For non-functions in math, we wrap the arguments in parentheses.
+        if in_math && !matches!(callee, Value::Func(_)) {
```
laurmaedje, line 144: We could try doing the `let callee = callee.cast::<Func>().at(callee_span);` above here, but not yet early return if it's `Err(_)`, instead first checking the non-function in math case.

Then, we could add `Symbol -> Func` to the cast implementation of `Func` in `foundations/value.rs` (like `Type` has). and get rid of the symbol special case above.

Not 100% whether it'll work, but worth a try.

crates/typst/src/symbols/symbol.rs
```diff
@@ -46,47 +46,63 @@ pub use typst_macros::symbols;
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Symbol(Repr);
 
+/// The character of a symbol, possibly with a function.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum SymChar {
```
laurmaedje, line 51: I think a struct with fields of type `char` and `Option<fn() -> Func>` would be a bit simpler.

crates/typst/src/symbols/symbol.rs
```diff
@@ -215,6 +204,41 @@ impl Display for Symbol {
...
+    /// Get the function associated with the symbol.
+    pub fn func(&self) -> Option<Func> {
+        match self {
+            Self::Pure(_) => None,
+            Self::WithFunc(_, func) => Some((func)()),
+        }
+    }
+}
+
+impl From<&SymChar> for char {
```
laurmaedje, line 227: I think an accessor would be clearer for this. The `char -> SymChar` from impl is fine, but this one feels a bit off.

crates/typst/src/symbols/symbol.rs
```diff
@@ -46,47 +46,63 @@ pub use typst_macros::symbols;
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Symbol(Repr);
 
+/// The character of a symbol, possibly with a function.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
```
laurmaedje, line 50: This could implement `Copy` and then we can probably get rid of a few `cloned` and the `ref` in the destructuring.

crates/typst/src/math/accent.rs
```diff
@@ -63,6 +117,29 @@ pub struct AccentElem {
     pub size: Smart<Rel<Length>>,
 }
 
+// Keep it synced with the documenting table above.
+accents! {
```
laurmaedje, line 121: I see how it's nice to have this close to the table, but I think I'd still prefer for it to be in its canonical place, below `impl Accent`.

crates/typst/src/math/accent.rs
```diff
@@ -1,18 +1,71 @@
...
+/// //  |            |      +-- The name of the function.
+/// //  |            +--------- The other characters that represent the accent.
+/// //  +---------------------- The default character that represents the accent.
+/// }
+/// ```
+///
+/// When combined with the `Accent::combine` function, accent characters can be normalized
+/// to the default character.
+macro_rules! accents {
+    ($($($ch:literal)|+ => $name:ident),+ $(,)?) => {
```
laurmaedje, line 31:
```suggestion
    ($($primary:literal $(| $alt:literal)* => $name:ident),* $(,)?) => {
```

crates/typst/src/math/accent.rs
```diff
@@ -1,18 +1,71 @@
...
+        impl Accent {
+            /// Normalize an accent to a combining one.
+            pub fn combine(c: char) -> Option<char> {
+                Some(match c {
+                    $(
+                        $($ch)|+ => accents!(@__private first $($ch),+),
+                    )+
```
laurmaedje, line 38:
```suggestion
                    $($primary $(| $alt)* => $primary,)*
```

crates/typst/src/math/accent.rs
```diff
@@ -1,18 +1,71 @@
...
+                /// The base to which the accent is applied.
+                base: Content,
+                /// The size of the accent, relative to the width of the base.
+                #[named]
+                size: Option<Smart<Rel<Length>>>,
+            ) -> Content {
+                let mut accent = AccentElem::new(
+                    base,
+                    Accent::new(accents!(@__private first $($ch),+)),
+                );
```
laurmaedje, line 57:
```suggestion
                let mut accent = AccentElem::new(base, Accent::new($primary));
```

crates/typst/src/math/accent.rs
```diff
@@ -1,18 +1,71 @@
...
+                    $(
+                        $($ch)|+ => accents!(@__private first $($ch),+),
+                    )+
+                    _ => return None,
+                })
+            }
+        }
+
+        $(
+            /// The accent function for callable symbol definitions.
```
laurmaedje, line 45: Would be nice if this was customized for each symbol. Could be something like:
```rs
/// The accent function for
#[doc = stringify!($primary)]
/// .
```
(Something like this. Didn't test whether this works correctly...)

crates/typst-macros/src/symbols.rs
```diff
@@ -29,19 +33,37 @@ pub fn symbols(stream: TokenStream) -> Result<TokenStream> {
...
 struct Variant {
     name: String,
     c: syn::LitChar,
+    handler: Handler,
+}
+
+enum Handler {
+    None,
+    Path(Path),
 }
```
laurmaedje, line 67: In my experience, redefined `Option` is a bit of an anti-pattern because you don't get the option methods with it (e.g. `is_none` which you had to redefine). You could try having a newtype `struct Handler(Option<Path>)` instead and see whether it ends up nicer or not.

crates/typst-macros/src/symbols.rs
```diff
@@ -68,16 +94,50 @@ impl Parse for Kind {
...
             name.push_str(&input.call(Ident::parse_any)?.to_string());
             while input.peek(Token![.]) {
                 input.parse::<Token![.]>()?;
                 name.push('.');
                 name.push_str(&input.call(Ident::parse_any)?.to_string());
             }
+            if name == "self" && !handler.is_none() {
+                return Ok(Self {
+                    name: "".to_string(),
+                    c: LitChar::new('\0', Span::call_site()),
```
laurmaedje, line 108: If I use `floor` directly as a symbol / cast it to a string, do I get the null-terminator? That would be a bit of a problem.

laurmaedje:
> Notice that current implementation enables accessing accents (which is only accessible in math mode before) via symbols, e.g. #sym.arrow("A").

I don't think it's a problem.

crates/typst-macros/src/symbols.rs
Coekjan: I tested the final pdf before. This symbol definition 

```rs
floor: [
    #[call(crate::math::floor)] self,
    l: '⌊',
    r: '⌋',
],
```

renders

```typ
$ A floor B $
```

into

image

Not sure if the `\0` does something unexpected.

crates/typst/src/eval/call.rs
Coekjan: I tried it just now. If we simply do something like:

```rs
primitive! {
    // ...
    // NOTE: `try_func()` wraps `func()` to give `StrResult`.
    // The error message is "symbol {} is not callable"
    Symbol(symbol) => symbol.try_func()?,
}
```

It cannot pass the tests. It seems because the compiler tries to cast some symbols into function, even if the symbol is not called. For example, when the following typst file gets compiled, the compiler emits an error: "symbol ∫ is not callable".

```typ
$ ∫_a^b iota_a^b $
```

To implement the cast without affecting current behavior of the compiler, I tried to add a `if` guard to the `primitive` macro, so that `Symbol -> Func` is not always castable:

```rs
primitive! {
    // ...
    Symbol(symbol) if symbol.func().is_some() => symbol.func().unwrap(),
}
```

I will push the code soon, and you can see my implementation.

crates/typst-macros/src/symbols.rs
Coekjan: Okay, let's remove the `self` notation

Coekjan:
Thanks for your kind and deep review. I have refined my code :smile:

crates/typst/src/eval/call.rs
Coekjan: Sorry for my mistake. The full case is:

```typ
#let eq = $ ∫_a^b iota_a^b $
#eq
#show "∫": math.limits
#show math.iota: math.limits.with(inline: false)
#eq
$iota_a^b$
```

It seems the show rule results in the cast

Coekjan: Here is the minimal case:

```typ
#show math.iota: math.limits.with(inline: false)
$ iota_a^b $
```

Coekjan: The stack backtrace indicates that `Selector` tries to cast anything to `Func`, so this is the reason why `Symbol -> Func` is applied in this case.

```txt
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_display
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:197:5
   3: core::panicking::panic_str
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:172:5
   4: core::option::expect_failed
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/option.rs:1995:5
   5: core::option::Option<T>::expect
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/option.rs:896:21
   6: typst::foundations::value::<impl typst::foundations::cast::FromValue for typst::foundations::func::Func>::from_value
             at ./crates/typst/src/foundations/value.rs:665:23
   7: <typst::foundations::selector::Selector as typst::foundations::cast::FromValue>::from_value
             at ./crates/typst/src/foundations/selector.rs:278:1
   8: <typst::foundations::selector::ShowableSelector as typst::foundations::cast::FromValue>::from_value
             at ./crates/typst/src/foundations/selector.rs:450:24
   9: typst::foundations::value::Value::cast
             at ./crates/typst/src/foundations/value.rs:155:9
  10: typst::eval::rules::<impl typst::eval::Eval for typst_syntax::ast::ShowRule>::eval::{{closure}}
             at ./crates/typst/src/eval/rules.rs:37:24
  11: core::option::Option<T>::map
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/option.rs:1073:29
  12: typst::eval::rules::<impl typst::eval::Eval for typst_syntax::ast::ShowRule>::eval
             at ./crates/typst/src/eval/rules.rs:35:24
  13: typst::eval::markup::eval_markup
             at ./crates/typst/src/eval/markup.rs:42:30
  14: typst::eval::markup::<impl typst::eval::Eval for typst_syntax::ast::Markup>::eval
             at ./crates/typst/src/eval/markup.rs:19:9
  15: typst::eval::eval::{{closure}}
             at ./crates/typst/src/eval/mod.rs:74:18
```

laurmaedje: I see. In that case, you can move the `text: EcoString` arm upwards to prefer casting a symbol to a text selector over a function selector (in particular since symbols won't generate element functions anyway).

Coekjan: Thanks! Now it passes all the tests :heart:

laurmaedje:
Thank you!

PR #4305, opened 2024-05-30
Add symbols `{gt,lt}.neq`

T0mstone:
Adds two new symbols `⪈` and `⪇`.

Unicode has these and I sometimes feel like using them to spite the existence of `⊊`.
Disregarding that, I think it's good to have these anyway for consistency.


Enivex:
I'm surprised these weren't already included

laurmaedje:
> I'm surprised these weren't already included

My first thought was that it's because both U+2A88 and U+2269 exist and I couldn't decide which one to take / how to name them. But actually U+2269 already has a name `gt.nequiv`, which goes against Unicode's naming ...

T0mstone:
> U+2269 already has a name `gt.nequiv`, which goes against Unicode's naming ...

Whatever we decide to do with this, I think U+2A88 ⪈ should be consistent with U+228B ⊋ (as well as their respected mirrored versions). The latter is currently called `supset.neq`, which is why I chose the names I did.

T0mstone:
(the force push just now was just a rebase since gh complained)

Enivex:
I think neq and nequiv are consistent with the rest of the naming. It would only be a problem if there was yet another symbol with three bars under.

laurmaedje:
Makes sense. Thanks!

Issue #563, opened 2023-04-04
Spanned diagnostics for non-Typst files


reknih:
The BibLaTeX error should have a trace in the BibLaTeX file so that the web app jumps there on click etc.

laurmaedje:
Closed by https://github.com/typst/typst/pull/6308

PR #4309, opened 2024-05-31
Basic Definition Finder for IDE

Myriad-Dreamin:
This PR introduces a `typst_ide::definition` function. Its signature simulates that of `typst_ide::tooltip`. With given a cursor, `typst_ide::definition` finds the definition of item at cursor:
- `DefinitionKind::Module`: any module definitions with a span.
- `DefinitionKind::Item`: any variables with a span or any builtin items (with a detached span).
- `DefinitionKind::Label`: any `<label>` definitions with a span.

`typst_ide::definition` doesn't find:
- a label in bibliography, whose name doesn't have a span, see #563.
- the real span of definition of an imported items, since we need to change the structure of scopes and tracer. I think @laurmaedje can design and improve them better than me.
- a definition of field, same reason as above.

Instead of completeness, we may focus on setting up workflow for `typst_ide::definition`. 
- I added two common matchers for all ide functions. The design of these matchers might be arguable.
- I find we cannot test the functionality of finding definitions in other files well, since `TestWorld` provides an only main source file.


crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,240 @@
...
+            value: Some(Value::Module(module.clone())),
+            range: Span::detached(),
+            name_range: Span::detached(),
+        }
+    }
+}
+
+/// The kind of a definition.
+#[derive(Debug, Clone, PartialEq, Hash)]
+pub enum DefinitionKind {
```
laurmaedje, line 170: These types aren't reexported at the moment, so they're inaccessible.

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,240 @@
...
+use typst::foundations::{Label, Module, Selector, Value};
+use typst::model::Document;
+use typst::syntax::ast::AstNode;
+use typst::syntax::{ast, FileId, LinkedNode, Side, Source, Span, SyntaxKind};
+use typst::World;
+
+use crate::analyze::analyze_import;
+use crate::{deref_target, named_items, DerefTarget, NamedItem};
+
+/// Find the definition of item under the cursor.
```
laurmaedje, line 11:
```suggestion
/// Find the definition of the item under the cursor.
```

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,240 @@
...
+///
+/// Passing a `document` (from a previous compilation) is optional, but enhances
```
laurmaedje, line 13: copy-pasta from autocomplete

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// `<foo>`
+    ///   ^^^
+    Label,
+}
+
+#[derive(Debug, Clone, PartialEq, Hash)]
+pub enum ModuleDefinitionKind {
+    /// `import ("fo" + "o.typ")`
+    ///           ^^^^^^^^
+    Path(FileId),
```
laurmaedje, line 186: Since this is constructed from a `Module`, why are we just providing the FileId, instead of the full module?

crates/typst-ide/src/definition.rs
laurmaedje: I'm not sure having three enums pulls its weight. Couldn't we just have 
```
enum DefinitionKind {
  Variable,
  Function,
  Label,
  Module(Module),
}
```

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// The name of the definition.
+    pub name: EcoString,
+    /// The kind of the definition.
+    pub kind: DefinitionKind,
+    /// A possible instance of the definition.
+    pub value: Option<Value>,
+    /// The source range of the entire definition.
+    pub range: Span,
+    /// The range of the name of the definition.
+    pub name_range: Span,
```
laurmaedje, line 123: The name_range seems to always be iehter detached or the same as `range`. Do we need it?

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Debug, Clone)]
+pub struct Definition {
+    /// The name of the definition.
+    pub name: EcoString,
+    /// The kind of the definition.
+    pub kind: DefinitionKind,
+    /// A possible instance of the definition.
+    pub value: Option<Value>,
+    /// The source range of the entire definition.
+    pub range: Span,
```
laurmaedje, line 121: Since it's not yet a range, I think just calling it `span` would be clearer.

crates/typst-ide/src/matchers.rs
```diff
@@ -0,0 +1,178 @@
+use ecow::EcoString;
```
laurmaedje, line 1: Please flatten the imports.

crates/typst-ide/src/matchers.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub(crate) fn value(&self) -> Option<Value> {
+        match self {
+            NamedItem::Var(..) | NamedItem::Fn(..) => None,
+            NamedItem::Module(value) => Some(Value::Module((*value).clone())),
+            NamedItem::Import(_, _, value) => value.cloned(),
+        }
+    }
+}
+
+/// Find an expression that can be dereferenced on the given node.
```
laurmaedje, line 129: I'm not yet entirely clear on what exactly a "matcher" is and what the specific contract/idea of `deref_target` is.

crates/typst-ide/src/matchers.rs
```diff
@@ -0,0 +1,178 @@
...
+            continue;
+        }
+
+        break;
+    }
+
+    None
+}
+
+pub enum NamedItem<'a> {
```
laurmaedje, line 103: Please add at least minimal doc comments for all items.

laurmaedje (commented):
Thanks for taking this on! It's really needed. Also, sorry for the long wait time. I left some comments now.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: I actually copied it from the documentation of `tooltip`.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: We may not be able to provide a `Module` if evaluating the module fails. It looks better if we resolve as a `FileId` with evaluation whenever possible. I did it in `tinymist-query` but I deleted the related code to reduce cost to the first PR.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: The (definition) range is larger or equal to the name range. Providing a range will help editor preview the definition. Providing a name range will help the features like renaming.

I haven't make the definition range yet, since it might introduce much more code.

Should we delete it currently?

image

crates/typst-ide/src/matchers.rs
Myriad-Dreamin: `deref_target` is a complete or incomplete "expression" node that can be operated by ide. It is a common matcher on AST that could be used by all ide functions, like `completion`, `tooltip` or `definition`. Two 



We can change the name of this function if we don't like it.

Myriad-Dreamin: `deref_target` is a complete or incomplete "expression" node that can be operated by ide. It is a common matcher on AST that could be used by all ide functions, like `completion`, `tooltip` or `definition`. Three advantages:
- make pattern matching in the ide functions more declarative and reusable:
  ```rs
  match deref_target(node) { 
    Callee(..) | VarAccess(..) => {}, // identifiers or identifier paths
    Label(..) => {}, // label expressions
    Normal(..) => {}, // normal expressions
    _ => ..,
  }
  ```
  v.s.
  ```rs
  if let ast::FuncCall(call) = node {
    // extract information from AST call
    do_identifier(..)
  } else if let ast::FieldAccess(fa) = node {
    do_identifier(..)
  } else if let ast::FieldAccess(fa) = node {
    do_label(..)
  } else if let ast::Error(err) = node {
    // recover errors
    do_identifier(..) or do_label(..)
  }
  ```
- resist changes of `typst_syntax` (AST).
- recover some common errors on imcomplete AST, like some opening labels.
  
We can change the name of this function if we don't like it.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: it is indeed simpler by flattening the enums, but I will not be able to ignore some more general kind when I'm processing the definition kinds:
```rs
// either
match def.kind { 
  Var(Variable) => {},
  Var(Function) => {},
}
// or
match def.kind { 
  Var(..) => {},
}
// or
match def.kind { 
  _ => {},
}
```
I'm sure it is a personal preference.

crates/typst-ide/src/definition.rs
laurmaedje: Yeah, let's only add it once it's needed.

crates/typst-ide/src/matchers.rs
laurmaedje: Well, the second version could also be written with match. But I see how it's nice to be able to process e.g. a FuncCall and a Set rule callee in a single match arm.

crates/typst-ide/src/matchers.rs
```diff
@@ -0,0 +1,178 @@
...
+            } else {
+                DerefTarget::Normal(expr_node.kind(), expr_node)
+            }
+        }
+        _ if expr.hash()
+            || matches!(expr_node.kind(), SyntaxKind::MathIdent | SyntaxKind::Error) =>
+        {
+            DerefTarget::Normal(expr_node.kind(), expr_node)
+        }
+        _ => return None,
```
laurmaedje, line 165: What is the exact definition of `Normal` vs `None`. Why `hash()`?

Myriad-Dreamin: It isn't well defined. The condition to check is borrowed on some existing code:

```rs
let mut ancestor = leaf;
while !ancestor.is::<ast::Expr>() {
    ancestor = ancestor.parent()?;
}

let expr = ancestor.cast::<ast::Expr>()?;
if !expr.hash() && !matches!(expr, ast::Expr::MathIdent(_)) {
    return None;
}
```

Perhaps some markup expressions are excluded, but I'm not quite sure an accurate list of them.

https://github.com/Myriad-Dreamin/typst/blob/3e1156787f1f05c514a80f3a7fa237f3dbf5c8d5/crates/typst-ide/src/tooltip.rs#L51-L59

laurmaedje: ah okay, yeah, there it excludes markup expression because that would get annoying real quick on hover

crates/typst-ide/src/definition.rs
Myriad-Dreamin: I implemented it.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: As discussed, we should add an overall README.md to describe all about these arguments together.

Myriad-Dreamin:
Resolved all discussion other than the one to add a `README.md`.

crates/typst-ide/src/lib.rs
```diff
@@ -92,7 +96,7 @@ fn summarize_font_family<'a>(variants: impl Iterator<Item = &'a FontInfo>) -> Ec
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
laurmaedje, line 99: why is this made public?

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,243 @@
...
+                span,
+                name_span: Span::detached(),
+            });
+        }
+        DerefTarget::Label(..) | DerefTarget::Code(..) => {
+            return None;
+        }
+    };
+
+    let mut paths = eco_vec![];
```
laurmaedje, line 53: Why do we collect the path here and continue just to early return below when the path is non-empty?

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// ```
+    Function,
+    /// ```plain
+    /// import ("fo" + "o.typ")
+    ///          ^^^^^^^^
+    /// ```
+    ///
+    /// IDE will always resolve a path instead of a module whenever possible.
+    /// This allows resolving a module containing errors.
+    ModulePath(FileId),
```
laurmaedje, line 188: I understand the desire to be resilient, but since the code currently only generates a `ModulePath` from an already known `Module` value it feels unnecessary for the time being.

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,243 @@
...
+    use typst::syntax::{Side, Span};
+
+    use super::{definition, Definition};
+    use crate::tests::TestWorld;
+
+    fn var(text: &str, value: bool) -> Option<Definition> {
+        Some(Definition::item(
+            &text.into(),
+            Span::detached(),
+            Span::detached(),
```
laurmaedje, line 223: why are the test spans all detached?

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// ```
+    Variable,
+    /// ```plain
+    /// let foo(it) = it;
+    ///     ^^^
+    /// ```
+    Function,
+    /// ```plain
+    /// import ("fo" + "o.typ")
+    ///          ^^^^^^^^
```
laurmaedje, line 183: this squiggle seems to be off. would also be nice if the docs mentioned what the squiggles describe. I assume it is `name_span`?

crates/typst-ide/src/definition.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// ```plain
+    /// import calc: *
+    ///        ^^^^
+    /// ```
+    ///
+    /// Some modules are not associated with a file, like the built-in modules.
+    Module(Module),
+    /// ```plain
+    /// <foo>
+    ///  ^^^
```
laurmaedje, line 198: does the span not also contain the angle brackets?

laurmaedje (commented):
I refactored a little bit, but got some new questions in the process.

An aside: The `node.find(some_expr.span())` is an interesting trick/workaround for the lack of a LinkedNode-style typed AST! ^^

crates/typst-ide/src/lib.rs
Myriad-Dreamin: This is related to the [discord message](https://discord.com/channels/1054443721975922748/1229699205786898485/1248923723298242560). But yeah, I should open another PR if I want to use `TestWorld` outside of crate.

> For typst-ide, internal functions just prevent developers from constructing their code analysis based on existing code of typst-ide. In typst crate it looks sensible to prefer private, but typst-ide we could prefer pub.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: I can continue my work on field definition and I don't think they cause any notable overhead, but I have removed them.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: They are not used, and I have no idea on test spans.

crates/typst-ide/src/definition.rs
Myriad-Dreamin: Ideally they shouldn't contain the angle bracket, but we cannot have this when we are using spans.

laurmaedje: I think it's okay if it contains the angle brackets. They are part of the label. But it would be good if the comment would reflect the exact span the compiler also emits.

Myriad-Dreamin: I updated the comment, like this:
```rs
/// ```plain
/// let foo(it) = it;
///     ^^^
/// ^^^^^^^^^^^^^^^^^ span
///     ^^^ name_span
/// ```
````

Myriad-Dreamin:
I have updated the PR according to the last conversation.

laurmaedje:
Thank you!

Issue #3015, opened 2023-12-20
Simultaneous query & compile


ntjess:
### Description

I have a large document that I would like to compile and see the results of some labeled content. Currently, I query and compile separately for 2 total calls to typst. It would be ideal to allow querying *during* compilation and print the query result to stdout while exporting files as usual.

### Use Case

Any documents where both metadata and outputs are equally desirable. A common use case is slideshows. In one compilation, the svg slides can be generated and their metadata for transitions, speaker notes, etc. will also be available.

SillyFreak:
I think there are many generalizations that are worth considering from the start:

- compile and do one query (the initial feature request)
- (compile and) do multiple queries
- do multiple compilations, e.g. with different `--input`s or output formats
- do any of these, even with different input documents (presumably, the input documents would be similar enough that a lot of memoized information would be reused between them)

I think the last one may be attainable, but may also have the fewest legitimate use cases and/or it might be an anti-pattern to structure one's workflow in a way where this saves a lot of work.

There has been some discussion on Discord in this direction: https://discord.com/channels/1054443721975922748/1245821833308012614

One of the suggestions was the option to start Typst as a agent and access it using an ID:

```bash
id=$(typst agent diagrams.typ)
typst query --id "$id" --field value metadata
typst c --id "$id"
typst c --id "$id" 'diagram{n}.svg'
typst agent --kill "$id"
```

this would allow the first three scenarios in a natural way.

PR #4312, opened 2024-06-01
Allow compile query

remimimimimi:
This PR adds the ability to compile query via CLI. It extends compile sub-command. Implementation of [#3015](https://github.com/typst/typst/issues/3015).

Currently, it lacks support of (1) keeping context values like counters and states, and (2) styles of target content - result of query which we render.

To resolve (1) I tried to rebuild the `Locator` from the frames of the first compilation, but results were incorrect. I also tried to keep `Locator` from the entire build of the document. This attempt was also unsuccessful.

I could solve issue (2) by abusing `Tracer`, but looking for a cleaner solution.

To see those experiments, look at [this branch](https://github.com/remimimimimi/typst/commits/query-compile/).

List of unsolved problems:
- [ ] Preserve context values
- [ ] Keep styles
- [ ] Decide which sub-command will implement this functionality - `compile` vs `query`

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -80,6 +80,14 @@ pub struct CompileCommand {
     #[clap(required_if_eq("input", "-"), value_parser = ValueParser::new(output_value_parser))]
     pub output: Option<Output>,
 
+    /// Defines which element to retrieve.
+    #[arg(long = "selector", default_value = None)]
+    pub selector: Option<String>,
+
+    /// Specify which element to render after selection.
+    #[arg(long = "element_number", short = 'n', default_value_t = 0)]
+    pub element_number: usize,
+
     /// Which pages to export. When unspecified, all document pages are exported.
     ///
     /// Pages to export are separated by commas, and can be either simple page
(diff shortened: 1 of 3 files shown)


laurmaedje:
I think the intent of #3015 was something else. Basically just to compile a document and at the same time run a query on the same document, rather than compiling the _output of the query,_ as implemented here. Did you have a specific use case for that in mind?

cc: @ntjess

remimimimimi:
@laurmaedje I misunderstood that issue, so I've created a [new one](https://github.com/typst/typst/issues/4313) that describes my use case.

laurmaedje:
I think there are a lot of challenges to properly support something like this.

- Correctly keeping contextual things is not possible at the moment, but might become easier in the future, due to other things I'm working on
- Keeping styles is also impossible at the moment
- The CLI interface. The `--element-number` interface feels a bit weird to me.

Can you elaborate more on the use case? I understood that it's for extracting compilations of specific equations or something like that? I can see how that can be useful (as it can work with calculations etc. correctly as opposed to just extracting the equation syntax), but it's also quite specific.

PR #2220, opened 2023-09-24
Fixing the baseline in math content

LuxxxLucy:
Per #2214 , it turns out that in `fn layout_text` of `math/ctx.rs`, which handles the layout of math content, regular text is  handled by the typst standard text layout (which will be calling the shaping engine), while the a single letter in a math font or a number is handled without going into the standard layout routines, but instead with an ad-hoc layout function in MathFragment::GlyphFragment.

The baseline shift is not taken consideration by the GlyphFragment which causes the issue of #2214. This PR fixes the issue as well as providing the bug's test case `tests/typ/bugs/math-shift.typ`.

```
hello #text(baseline: -5pt)[123 #sym.WW\orld]\
hello #text(baseline: -5pt)[$123 WW#text[or]$ld]\
```

result: 
image

PR #2186, opened 2023-09-18
Add gaps for matrix, case, and vector

SekoiaTree:
Fixes #1696. I'm not sure if I can have row-gaps be a parameter in math mode, but I don't know enough about how math mode works to get it to accept it easily.

PR #4318, opened 2024-06-02
Adding a ascent/descent setter in frame so that the original frame size is not lost in equation slackness adjustment

LuxxxLucy:
In equations, there is an adjustment of the frame size, where the frame height (size().y) is changed so that a large equation will be adjusted to not effect the later line layout (as it effects the line height)

```
let MathParItem::Frame(frame) = item else { continue };
 
             let font_size = scaled_font_size(&ctx, styles);
             let top_edge = TextElem::top_edge_in(styles).resolve(font_size, &font, None);
             let bottom_edge =
                 -TextElem::bottom_edge_in(styles).resolve(font_size, &font, None);
 
            let ascent = top_edge.max(frame.ascent() - slack);
            let descent = bottom_edge.max(frame.descent() - slack);
             frame.translate(Point::with_y(ascent - frame.baseline()));
             frame.size_mut().y = ascent + descent;
         }
```

However, if the frame.size().y is modified here, we will not be able to do decoration of the frame (see #2200) , because the frame size is not correct anymore. This PR adds the interface to set ascent/descent of the frame, and so that the line height determination will rely to the value of ascent and descent.

Below shows the result after this PR's change, the red line is the baseline, and the blue line is the ascent, while the background is true frame size. The frame size is not modified.

image

## Note 1:
The change of this PR should  also be relevant to #1028 #4236

## Note 2:
this PR does not break any other test, except for #2220, the reason being the descent being changed.  Before:
Snipaste_2024-06-01_20-04-18
After
Snipaste_2024-06-01_20-02-51


crates/typst/src/layout/frame.rs
```diff
@@ -31,6 +31,12 @@ pub struct Frame {
     items: Arc<LazyHash<Vec<(Point, FrameItem)>>>,
     /// The hardness of this frame.
     kind: FrameKind,
+    /// The ascent of the frame. If this is `None`, the
+    /// frame's implicit ascent is the baseline.
+    ascent: Option<Abs>,
+    /// The descent of the frame. If this is `None`, the
+    /// frame's implicit descent is size minus baseline.
+    descent: Option<Abs>,
 }
```
Enter-tainer, line 40: I think `size` + `baseline` already determines ascent and descent. The box model should look like this

```
          ────┬──────────────────────────┬───    
            ▲ │                          │ ▲     
            │ │                          │ │     
            │ │                          │ │     
            │ │                          │ │     
            │ │                          │ │     
            │ │                          │ │     
            │ │                          │ │     
          asc │                          │ │     
            │ │                          │ │     
            │ │                          │ │     
            │ │                          │ size.y
            │ │                          │ │     
            │ │                          │ │     
            │ │                          │ │     
baseline  ──┴─┼──────────────────────────┤ │     
            ▲ │                          │ │     
            │ │                          │ │     
       desc │ │                          │ │     
            ▼ │                          │ ▼     
          ────┴──────────────────────────┴────   
```

LuxxxLucy: The point is that ascent+descent != size.y, as in the case of the equation slackness adjustment. In the slackness adjustment, the size (the original size of the frame) is larger than the ascent+descent.

The ascent and descent will be used by inline layout for determining the line height while the size still refers to the original size.

Enter-tainer: > The point is that ascent+descent != size.y

I highly suspect this is not a good idea. `asc + desc = size.y` should be a invariant.

LuxxxLucy: what the slackness adjust does
```
          ────┬──────────────────────────┬───    
              │                            ▲     
              │                            │
              │                            │     
              │                            │     
          ────┬─────────────┬              │  
            ▲ │                            │     
            │ │                            │     
          asc (adjusted by slackness)      │     
            │ │                            │     
            │ │                            │     
            │ │                            size.y
            │ │                            │     
            │ │                            │     
            │ │                            │     
baseline  ──┴─┼──────────────────────────┤ │     
            ▲ │                            │     
            │ │                            │     
       desc │ │                            │     
            ▼ │                            ▼     
          ────┴──────────────────────────┴────   
```

previously the ascent/descnet is not changed, but the size y is directly changed.



```
          ────┬──────────────────────────┬───────
              │                                 ▲
              │                                 │
              │                                 │
              │                                 │
          ────┬──────────────────────────┬───   │ 
            ▲ │                            ▲    true size.y
            │ │                            │    │ 
            │ │                            │    │ 
            │ │                            │    │ 
            │ │                            │    │ 
            │ │                            size.y (modified by the slackness)
            │ │                            │    │ 
            │ │                            │    │ 
            │ │                            │    │ 
baseline  ──┴─┼──────────────────────────┤ │    │ 
            ▲ │                            │    │ 
            │ │                            │    │ 
       desc │ │                            │    │ 
            ▼ │                            ▼    ▼ 
          ────┴──────────────────────────┴───────   
```

LuxxxLucy: > > The point is that ascent+descent != size.y
> 
> I highly suspect this is not a good idea. `asc + desc = size.y` should be an invariant.

In fonts, I mean in reality, ascent and descent does not really means the max top/bottom edge of the shape, a particular glyph can easily reach higher than the ascent or lower than descent, it is not an invariant.

Example: many, but the easiest example is accented uppercase À.

The point is if we should not modify the original frame size when doing slackness, otherwise when we do inline layout we lost track of the true size of the frame and cannot use it anymore.

LuxxxLucy: > Slackness adjust for inline equations adjust is... a quick & dirty workaround in my opinion, and hopefully I want to remove it my PR #4236 (I'm still waiting @laurmaedje 's comment on the line height proposal #4224). There is no such adjust in proper typesetting systems like LaTeX.

I think it would be impossible to make #4236 pass the test without making an explicit handling of the ascent/descent explicit setter. And tslackness is properly handled in Latex and I think @laurmaedje has explained and provided examples.

Or if you are  concerned with the naming, we can easily rename the ascent() descent() to something like line_height_above_baseline() and line_height_below_baseline(), the usage remains the same.

Enter-tainer: > In fonts, I mean in reality, ascent and descent does not really means the max top/bottom edge of the shape, a particular glyph can easily reach higher than the ascent or lower than descent, it is not an invariant.

Sorry i don't know much about this. All I learn about baseline/height/depth comes from TeXBook and typst's source code. For this specific question, typst [do calculate the true bbox for `À`](https://docs.rs/typst/latest/src/typst/math/fragment.rs.html#294) and use it to determin size, height and depth, by using the path data stored in the font.

For the box model, I think typst borrow the idea from TeX. Here is relevant explaination in TeXBook page 63.
image

> The point is if we should not modify the original frame size when doing slackness, otherwise when we do inline layout we lost track of the true size of the frame and cannot use it anymore.

Agree. This is the root cause of #1028. It is a workaround for line spacing issue. I'd hope we can totally eliminate slackness adjust. And we shouldn't have another workaround for a workaround.

LuxxxLucy: > , typst [do calculate the true bbox for `À`](https://docs.rs/typst/latest/src/typst/math/fragment.rs.html#294) and use it to determin size, height and depth, by using the path data stored in the font.

This is true, the point is what the height (used for line height determination) is because we are noting treating only a sequence of pure text but text mixed with math frames and even inline image. And I think this height is the ascender (or for the math equations, the slackness adjusted height), instead of the actual size of this particular glyph.

Enter-tainer: > it would be impossible to make https://github.com/typst/typst/pull/4236 pass the test

Sorry, maybe there is a gap. I don't really get this point though. Currently there are about ~200 test doesn't pass. This is because I haven't fix line height for termlist/unordered list/ordered list yet. Another take is that you can always "pass" the test by updating the reference images..?

> And slackness is properly handled in Latex and I think @laurmaedje has explained and provided examples.

believe me, tex doesn't have this adjust. The tex solution is described in #4224. Namely, tex determines line gap in 2 stages: in normal cases, it tries to make the distance between lines equals to `\baselineskip`. But if this makes lines overlap, it will ensure that the distance of the bottom edge of the previous line and the top edge of current line is at least 1pt.

Enter-tainer: > And I think this height is the ascender (or for the math equations, the slackness adjusted height), instead of the actual size of this particular glyph.

This is unfortunately not true. slackness adjust only happens in equations that is to be inserted into normal text flows. They don't exist in other cases. For example:

```typ
has slacknessadjust: $a^2+b^2=c^2$

has slackness adjust as a whole, but no slackness adjust for each individual case arms. 

$cases(a^2+b^2=c^2, a=3, b=4, c=5)$
```

LuxxxLucy: Okay I learn a lot. Thank you.

I am okay as long as the original size is not lost as I am interested in #2200; that being said, I still think it would be difficult to make #4236 pass the test (and I mean by preserving the existing result without breaking the tests, but I am not very familiar with it so I can be wrong). The line height adjustment seems to be complicated, if we are really going for it, a complete collision detection procedure needs to be implemented and some heuristic would be inevitable. I hope it can be done soon. 

Do you think it is possible to get a predecessor PR to #4236  that can preserve the frame size?

Enter-tainer: > I am okay as long as the original size

#4236 removes the slackness adjust, so inline math has correct bbox.

> I still think it would be difficult to make https://github.com/typst/typst/pull/4236 pass the test

Agree, the term list thing is not easy to fix. One thing to notice about typst's test system is that the reference image can be update with `--update` when running tests. This is especially useful when rendering result is expected to change.

> The line height adjustment seems to be complicated, if we are really going for it, a complete collision detection procedure needs to be implemented and some heuristic would be inevitable. 

To my surprise, it is not that hard(at least for most cases). Typst already maintains the exact bbox for each element, and the collision test code can be seen in https://github.com/typst/typst/pull/4236/files#diff-6c11c99cb19f6dd03f76013fec99eb36081d56355d9b5ed8546f1d785bd49af4L323-L325 .

You can see the inline equation overlap issue solved in this image: https://github.com/typst/typst/pull/4236/files#diff-d5c29466f38994d7e7c7337b0fee01231dc6a45d0a65b37807d38e3a2647cae1

> I hope it can be done soon.

Me too! I didn't work on it last week because I'm (a) working on #2186 (b) waiting laurmaedje's comment on the proposal. I will continue to work on this pr.

> Do you think it is possible to get a predecessor PR to https://github.com/typst/typst/pull/4236 that can preserve the frame size?

I'm not really sure. I haven't think of this in detail(because I bet on the line height solution!), there might be no "non-hacky" way to do this.

laurmaedje: It depends on what we settle on in #4224, but in principle I agree with @Enter-tainer that `ascent + descent = size.y` is a good invariant to have, and that the equation slack is more of a hack.

LuxxxLucy:
Closing in favour of waiting for #4224

PR #4319, opened 2024-06-02
Hint for labels in code mode

### LINKED ISSUE #4317: Improve error message for labels in code mode
### Description

Currently, you cannot add labels to elements constructed in code mode, requiring you to use a markup block instead. However, the error isn't very helpful, which leads to this being a common source of confusion for beginners. Even if it could potentially be interesting to support label attachment syntax in code mode in the future, I believe improving the error message could be a good first step for the time being.

For example, consider the code below:

```typ
#{
  figure([a]) <b>
}
```

This will currently fail with "expected semicolon or line break". Would be nice to have it suggest using a `#[ ]` markup block instead. (Could even suggest a fix like Rust does in the future, if we ever get support for that.)

### Use Case

Mostly just prevent a common source of confusion for beginners.

Coekjan:
This PR closes #4317 .

To add hints during parsing, I added a new function `Parser::hint(&mut self, hint: &str)`.

### DIFF
--- a/crates/typst-syntax/src/parser.rs
+++ b/crates/typst-syntax/src/parser.rs
@@ -633,6 +633,10 @@ fn code_exprs(p: &mut Parser, mut stop: impl FnMut(&Parser) -> bool) {
             code_expr(p);
             if !p.end() && !stop(p) && !p.eat_if(SyntaxKind::Semicolon) {
                 p.expected("semicolon or line break");
+                if p.at(SyntaxKind::Label) {
+                    p.hint("labels can only be applied in markup mode");
+                    p.hint("try wrapping your code in a markup block (`[ ]`)");
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/typst-syntax/src/parser.rs
```diff
@@ -633,6 +633,9 @@ fn code_exprs(p: &mut Parser, mut stop: impl FnMut(&Parser) -> bool) {
             code_expr(p);
             if !p.end() && !stop(p) && !p.eat_if(SyntaxKind::Semicolon) {
                 p.expected("semicolon or line break");
+                if p.at(SyntaxKind::Label) {
+                    p.hint("you can use labels only in markup mode");
```
PgBiel, line 637: Maybe we can improve this message a bit? Something like "labels can only be applied in markup mode", followed by a second hint "try wrapping your code in a markup block (\`[ ]\`)"

(I'm open to further suggestions)

PgBiel (commented):
Wow, that was fast! Thanks!

laurmaedje:
Thanks!

PR #4327, opened 2024-06-03
Add missing math symbols (U+0000 – U+257F)

bluebear94:
At least the ones that I think make sense to include and have a good idea of what it should be called.

Thanks to @Enivex for their spreadsheet of operators with their LaTeX and Typst symbol names.

Unresolved questions:

* [ ] How should symbols with over- or underlines be named?


Enivex:
Just some thoughts:

1. I think most of the APL symbols should be their own category (also what's up with the `r#box`?), as they're quite special. Especially the ones where the latex name includes APL.
2. Instead of `≖` `≗` being `eq.circle.small` and `eq.circle.small.above` I'd suggest `eq.circle.c` and `eq.circle.t`
3. Unless there's yet more minus symbols to distinguish between, I'd go with `minus.dots` for ∺ instead of `minus.dot.quad`
4.  For `≑` `≒` `≓` I suggest `eq.dots`, `eq.dots.down`, `eq.dots.up` due to presedence
5. Move `≍` and `≭` to `asymp`and `asympt.not` instead of putting them under eq.
6. `≏` and `≎` should be `eq.bump` and `eq.bumps` (or the more verbose `eq.bump.double`). In addition there is the related symbol `⪮`, which I don't think you've added. While the official name is "equals sign with bumpy above" I'd suggest using `equiv.bump` for this.
7. ⋺ is misnamed `in.rev.double`, when it should be `in.rev.long`
8. I don't think `not.inv` is a good name for `⌙`. Inversion brings to mind mirroring (like for `rev`). I'd go with `not.turned`, which matches both the latex and official Unicode name ("Turned Not Sign").
9. ⋕ should be `parallel.eq`, not `parallel.equal`
10. The important distinction between ⊿ (U+22BF) and ◿ (U+25FF) is that the former has class Normal while the latter has class Binary. This is why the Latex name includes `var`. This distinction is not at all clear from the typst name. Though I'm not sure what the solution should be.
11. `lozenge.square` should be `lozenge.sq` I believe
12. All the `twostroke` variants of `arrow` should surely be `stroke.double`
13. The naming of `arrow.t.b.bar` `arrow.tl.bar` isn't entirely consistent with the other `bar` variants.
14. I do not think the naming of the turning arrows (↰ etc) is particularly good
15. Same comment about `iota.inv` as I had for `not.inv`. It should be `iota.turned`
16. Regarding the "// Miscellaneous non-letter-like symbols." section. We should investigate this more closely and see if they can fit into other categories.

I'm sure I missed some things. There's a lot of stuff happening here.

Enivex:
An alternative to `turned` is `rot` (for `rotated`)

Enivex:
In general I might add that my purpose for creating the spreadsheet was that I wanted to discuss naming as a whole (also for existing symbols). I think a PR is a bit premature, at least for some of these.

bluebear94:
> 6\. `≏` and `≎` should be `eq.bump` and `eq.bumps` (or the more verbose `eq.bump.double`). In addition there is the related symbol `⪮`, which I don't think you've added. While the official name is "equals sign with bumpy above" I'd suggest using `equiv.bump` for this.

`⪮` doesn’t occur until later in the code point range, but I’ve added this here as well.

> 8\. I don't think `not.inv` is a good name for `⌙`. Inversion brings to mind mirroring (like for `rev`). I'd go with `not.turned`, which matches both the latex and official Unicode name ("Turned Not Sign").
>
> 15\. Same comment about `iota.inv` as I had for `not.inv`. It should be `iota.turned`

This name matches `inv` being used for turned symbols in `{amp, excl, quest}.inv`.

> 10\. The important distinction between ⊿ (U+22BF) and ◿ (U+25FF) is that the former has class Normal while the latter has class Binary. This is why the Latex name includes `var`. This distinction is not at all clear from the typst name. Though I'm not sure what the solution should be.

I’ll remove it for now.

Enivex:
> This name matches `inv` being used for turned symbols in `{amp, excl, quest}.inv`.

I'd suggest changing those to `.rot` as well. While the word inverted can be used in this sense, it's somewhat ambiguous because it can also involve reversion.

bluebear94:
I think this is going to need further discussion, so closing this for now.

PR #4330, opened 2024-06-04
Add hint when string is used in place of label

### LINKED ISSUE #4329: Hint when string is used in place of label
### Description

The error message `expected label, found string` should have a hint explaining how a label differs syntactically from a string.

From web app feedback.

### Use Case

Help users out, especially since it's `#cite(<label>)` rather than `#cite(@label)` which isn't immediately obvious.

bluebear94:
Closes #4329.

This also refactors `CastInfo::error` to return a hinted string and the `FromValue` implementation generated by the `primitive` macro to forward to `Reflect::error`.

### DIFF
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -920,10 +920,11 @@ fn is_math_id_continue(c: char) -> bool {
 }
 
 #[inline]
-pub fn is_valid_in_label_literal(c: char) -> bool {
+fn is_valid_in_label_literal(c: char) -> bool {
     is_id_continue(c) || matches!(c, ':' | '.')
 }
 
+/// Returns true if this string is valid in a label literal.
 pub fn is_valid_label_literal_id(id: &str) -> bool {
     !id.is_empty() && id.chars().all(is_valid_in_label_literal)
 }
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -408,7 +408,7 @@ impl Lexer<'_> {
     }
 
     fn ref_marker(&mut self) -> SyntaxKind {
-        self.s.eat_while(is_valid_in_label);
+        self.s.eat_while(is_valid_in_label_literal);
 
         // Don't include the trailing characters likely to be part of text.
         while matches!(self.s.scout(-1), Some('.' | ':')) {
@@ -419,7 +419,7 @@ impl Lexer<'_> {
     }
 
     fn label(&mut self) -> SyntaxKind {
-        let label = self.s.eat_while(is_valid_in_label);
+        let label = self.s.eat_while(is_valid_in_label_literal);
         if label.is_empty() {
             return self.error("label cannot be empty");
         }
@@ -920,6 +920,10 @@ fn is_math_id_continue(c: char) -> bool {
 }
 
 #[inline]
-pub fn is_valid_in_label(c: char) -> bool {
+pub fn is_valid_in_label_literal(c: char) -> bool {
     is_id_continue(c) || matches!(c, ':' | '.')
+}
+
+pub fn is_valid_label_literal_id(id: &str) -> bool {
+    !id.is_empty() && id.chars().all(is_valid_in_label_literal)
 }
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -408,7 +408,7 @@ impl Lexer<'_> {
     }
 
     fn ref_marker(&mut self) -> SyntaxKind {
-        self.s.eat_while(|c| is_id_continue(c) || matches!(c, ':' | '.'));
+        self.s.eat_while(is_valid_in_label);
 
         // Don't include the trailing characters likely to be part of text.
         while matches!(self.s.scout(-1), Some('.' | ':')) {
@@ -419,7 +419,7 @@ impl Lexer<'_> {
     }
 
     fn label(&mut self) -> SyntaxKind {
-        let label = self.s.eat_while(|c| is_id_continue(c) || matches!(c, ':' | '.'));
+        let label = self.s.eat_while(is_valid_in_label);
         if label.is_empty() {
             return self.error("label cannot be empty");
         }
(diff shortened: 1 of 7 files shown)


crates/typst/src/foundations/cast.rs
```diff
@@ -328,13 +330,24 @@ impl CastInfo {
...
-            if parts.iter().any(|p| p == "length") && !matching_type {
-                write!(msg, ": a length needs a unit - did you mean {i}pt?").unwrap();
+            if !matching_type && parts.iter().any(|p| p == "length") {
+                msg.hint(eco_format!("a length needs a unit - did you mean {i}pt?"));
+            }
+        }
+        if let Value::Str(s) = found {
+            if !matching_type && parts.iter().any(|p| p == "label") {
+                if is_valid_in_label_literal(s) {
+                    msg.hint(eco_format!("use `<{s}>` to create a label or the label function to convert a string to a label"));
```
Coekjan, line 343: Maybe we can improve this message a bit? I would suggest "use `<{s}>` or `#label("{s}")` to create a label".

(I'm open to further suggestions)

crates/typst/src/foundations/cast.rs
```diff
@@ -328,13 +330,24 @@ impl CastInfo {
             write!(msg, "{}", found.ty()).unwrap();
         }
 
+        let mut msg = HintedString::new(msg.into());
```
Coekjan, line 333:
```suggestion
        let mut msg: HintedString = msg.into();
```

tests/suite/layout/length.typ
```diff
@@ -45,7 +45,8 @@
 }
 
 --- length-unit-hint ---
-// Error: 1:17-1:19 expected length, found integer: a length needs a unit - did you mean 12pt?
+// Error: 1:17-1:19 expected length, found integer
+// Hint: 1:17-1:19 a length needs a unit - did you mean 12pt?
```
Coekjan, line 49: Line numbers here could be removed.

```suggestion
// Error: 17-19 expected length, found integer
// Hint: 17-19 a length needs a unit - did you mean 12pt?
```

tests/suite/model/cite.typ
```diff
@@ -114,3 +114,17 @@ B #cite(<netwok>) #cite(<arrgh>).
 @mcintosh_anxiety
 #show bibliography: none
 #bibliography("/assets/bib/works.bib", style: "chicago-author-date")
+
+--- cite-type-error-hint ---
+// Test hint for cast error from str to label
+// Error: 7-15 expected label, found string
+// Hint: 7-15 use `<netwok>` to create a label or the label function to convert a string to a label
+#cite("netwok")
+#bibliography("/assets/bib/works.bib")
```
Coekjan, line 123: Maybe it is no need to add `#bibliography(..)` here.

crates/typst/src/foundations/label.rs
```diff
@@ -78,3 +78,7 @@ impl From<Label> for PicoStr {
 
 /// Indicates that an element cannot be labelled.
 pub trait Unlabellable {}
+
+pub(crate) fn is_valid_in_label_literal(id: &str) -> bool {
+    !id.is_empty() && id.chars().all(typst_syntax::is_valid_in_label)
+}
```
Coekjan, line 84: It seems that this function is simple and only used once in `CastInfo::error`.

I'm not sure if this is over-encapsulation.

Coekjan:
By the way, you may want to change the PR description to something like "Closes #4329", so that GitHub shall close that issue when this PR is merged.

crates/typst/src/foundations/cast.rs
laurmaedje: I like that! But we should probably drop the hash since this error will appear in code mode.

bluebear94: What if the offending expression is not a literal?

laurmaedje: Fair point. But in any case the current wording imo contains the word "label" too many times.

Coekjan: Can we first check if it is string literal? And if so, we can have `label("{s}")` as the hint. Otherwise, we fallback to a general hint.

bluebear94: Right now, not all values keep span info. The other solution is to make `FromValue::from_value` take span info, which seems icky.

laurmaedje: How about ``use the label syntax `<{s}>` or create a label from a string with the `label` function``?

laurmaedje: I didn't see that the code was already changed. I'm also okay with leaving it as-is, with the `"label({})", s.repr()`

bluebear94: Anyway, we could change it to show a different message for non-literals in a future PR, perhaps once Typst moves to a bytecode interpreter.

crates/typst/src/foundations/label.rs
bluebear94: I think having this is fine, too, since the implementation isn’t exactly trivial, but I’ve thought about whether to put it here or in `typst_syntax`.

laurmaedje: The naming is a bit tricky. Right now the naming doesn't make the difference clear.

We could go with
- `typst_syntax::is_valid_in_label` -> `typst_syntax::is_valid_char_in_label_literal`
- `is_valid_in_label_literal` -> `typst_syntax::is_valid_in_label_literal`

Or we could go with @Coekjan's reasoning and have only the char-version as `typst_syntax::is_valid_in_label_literal`. Makes the naming easier.

bluebear94: I opted for:

* `typst_syntax::is_valid_in_label` → `typst_syntax::is_valid_in_label_literal`
* `is_valid_in_label_literal` → `typst_syntax::is_valid_label_literal_id`

crates/typst-syntax/src/lexer.rs
```diff
@@ -918,3 +918,13 @@ fn is_math_id_start(c: char) -> bool {
 fn is_math_id_continue(c: char) -> bool {
     is_xid_continue(c) && c != '_'
 }
+
+#[inline]
+fn is_valid_in_label_literal(c: char) -> bool {
```
Coekjan, line 923: You may want to document this function.

laurmaedje:
Thank you!

PR #4345, opened 2024-06-06
interrupt_styles: Check all potential style matches

### LINKED ISSUE #4340: `set document()` resets `set page()` (paper size, fill) on the last empty page
### Description

If I set `set document(author: "")` before `set page(paper: "presentation-16-9", fill: gray)` then the last page *if empty* will have paper size set to A4.

```typ
#set document(author: "")
#set page(paper: "presentation-16-9", fill: gray)
text
```

So, if I do

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

bluebear94:
Fixes #4340.

Previously, a style list with a style affecting, say, DocumentElem, would not continue with a potential match with a PageElem. This causes a style list affecting both to only handle the DocumentElem case, thus failing to call interrupt_page when necessary.


laurmaedje:
I think only the first change is actually necessary since `interrupt_page` implies `interrupt_par` etc. But the document interruption just doesn't do anything, so it doesn't work that way for document styles. 

But the performance difference is negligible and so I think it's maybe clearer and doesn't hurt to remove all the `else`s.

laurmaedje:
Thanks!

Issue #3727, opened 2024-03-19
`math.class` is not taken into account within a `context` expression


loutr:
### Description

I maintain the tiny `xarrow` package and the latest version 0.11 of the compiler seems to have brought a regression, which [impacts the package](https://codeberg.org/loutr/typst-xarrow/issues/6). A minimal example is the following:

```typst
#let test = math.class("relation", "--")
$ a test b $
```
displays correctly, with an appropriate spacing around `--`, but

```typst
#let test = style(_ => math.class("relation", "--"))
// or the new #let test = context math.class("relation", "--")
$ a test b $
```
doesn't (it has an extra space between `--` and `b`).

This is easily fixable in the case of xarrow for instance, where it boils down to swapping operations as in:
```
// where expr depends on context of course
#let test = math.class("relation", context expr)
```
but I figured it would be best to signal it anyway. Sorry if this is a duplicate or my mistake or anything. Thanks for your amazing work as always!

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

wrzian:
This seems like it might be related to the issue with the `math-call-pass-to-box` test I added? See here: https://github.com/typst/typst/pull/3786#issuecomment-2061864950

mkorje:
On main, all three examples look identical. Bisect says #4348 fixed this.

Is there still an issue here? @EpicEricEE you mentioned in #4348 that the "--" is not being measured correctly.

EpicEricEE:
I did, but that is probably a separate issue

![image](https://github.com/user-attachments/assets/03250e36-fa2c-4037-a2e3-d11e67003d22)

PR #4348, opened 2024-06-06
Avoid spaces around ignorant and invisible elements in math

### LINKED ISSUE #4326: Unexpected space in math between invisible element and text
### Description

_(Partially) a regression on main, using 084a0614_

```typ
#let size1 = context measure($ "abc" $)
#let size2 = context measure($ #metadata(none) "abc" $)
#let size3 = context measure($ #metadata(none)"abc" $) // Note missing space
```

In 0.11.1, `size1 == size2 == size3`, but on main `size2 > size1` and `size3 == size1`. This is also visible when rendered:

image

As the metadata element isn't visible, the space between it and the text should not be kept.
The same happens for other invisible elements, such as counter or state updates. _However, with those elements, the space is also there in 0.11.1._

**Addendum:** Another instance of this issue is the fact that the following to sizes are not equivalent. This was also already the case in 0.11.1.
```typ
#let size1 = context measure($ #metadata(none) = 0 $)
#let size2 = context measure($ = 0 $)
```

Maybe related to #4234

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

EpicEricEE:
Fixes #4326

I don't know what the intended behaviour is, but I feel like invisible and ignorant elements should not be accounted for when the space between two elements is inserted in an equation.

I have to admit that I don't really like the growing list of fields on the `FrameFragment` struct but it seemed to be the easiest way.


bluebear94:
Does this fix #3727 as well? In particular, do all three of the following lines display identically?

```typ

#let test = context math.class("relation", "--")
$ a test b $

$ a class("relation", "--") #metadata("hi") b $

$ a class("relation", "--") b $
```

EpicEricEE:
They all look identical, but I'm not sure if it's correct as it doesn't look centered.

image

laurmaedje:
Why another field instead of just setting `spaced` to `false` for ignorant/invisible?

EpicEricEE:
Because then it wouldn't be truly ignorant but just like a `normal` class frame which influences the spacing between neighboring frames, e.g. `vary` class operators would become binary in the test cases
```typ
// Operator spacing
$#counter("test").update(3) + b$ \
$#place(dx: 5em)[a] + b$
```

laurmaedje:
Okay.

> I have to admit that I don't really like the growing list of fields on the FrameFragment struct but it seemed to be the easiest way.

Me neither, but I guess one more field won't change anything.

If you could fix the conflict, I'll merge.

laurmaedje:
Thanks!

Issue #1841, opened 2023-08-01
Element function to freeze state


edgarogh:
## Context / use-case

I would like to create a template for PDF slideshows, with a `#fragment[]` function that allows me to hide an element at first and reveal it later on a duplicate page (like reveal.js fragments or Beamer's `\pause`s).

## Description

To do so, I'd need the ability to duplicate a page, freezing almost all state (including `state()` variables, heading numbering, page number, etc.), except for one "iteration variable" that tells me which copy I'm laying out currently.

Essentially, I'm looking for a function that would "instantiate" a `content` at a specific location (regarding state) in a way that resolves/freezes implicit and explicit state inside it, so that any attempt to lay it out multiple times results in the same state materialization.

### Possible implementation 1: `freeze(content, exclude: list or state)`

```ocaml
#let idx = state("idx", 0)
#let body = [= Heading #idx.display() #idx.update(v => v+1)] // depends on implicit heading state

#let body_frozen = freeze(body, exclude: (idx,))

// Headings have the same number
body_frozen
body_frozen
body_frozen
```

Here, we explicitly exclude a state variable from the freezing, so it can be used and mutated inside.

Note: if a heading is inserted between the call to `freeze` and the layouts of `body_frozen` (L5), we would expect the numbering to be in the wrong order, which I guess would be a feature?

### Possible implementation 2: `clone(content, count: number, exclude: list or state)`

This approach would be more "managed" and tailored to my specific case, leading to less room for weird behavior such as unordered numbering:

```ocaml
#let idx = state("idx", 0)
#let body = [= Heading #idx.display() #idx.update(v => v+1)]

clone(body, exclude: (idx,), count: 3) // Creates 3 copies of body, with only "idx" being mutated
```

Another possible approach would be to provide an map-like usage:

```ocaml
#let idx = state("idx", 0)
#let body = [= Heading #idx.display()]

clone(body, list: (0, 1, 2), state: idx)
```

Here, the list elements would be mapped to the body, passing their values through the `idx` state variable. The state variable could maybe have its original value stored and restored after the iteration, so that it isn't observable outside?

## Soundness

I must admit, I haven't thought that much about the "soundness" of such a feature, and how it would interact with all the weird "meta" Typst features (locate, query, etc.).

## Last note

Maybe this idea is a bad solution to the project I have initially (XY problem). Maybe it should be restricted in other ways. Or made even more generic. Maybe my initial problem should just be a Typst builtin.

sitandr:
[...]

PR #4352, opened 2024-06-07
Pure location assignment

### LINKED ISSUE #1910: Duplicated figure counter for customized figures in in #place blocks
### Description

I used sample customized figure from documentation: https://typst.app/docs/reference/meta/figure/#modifying-appearance

```typst
#show figure: it => align(center)[
  #it.caption |
  #emph[
    #it.supplement
    #it.counter.display(it.numbering)
  ]
  #v(10pt, weak: true)
  #it.body
]
```

Then I added 3 figures, with a single figure in a place block:

```typst
#outline(target: figure)

#figure(caption: "Sample figure 1", text("Figure content."))

#place[
  #figure(caption: "Sample figure 2", text("Figure content."))
]

#figure(caption: "Sample figure 3", text("Figure content."))
```

Displayed counters are wrong (`Figure 2.` is displayed two times), `#outline` displays them correctly though.

image

### Reproduction URL

https://typst.app/project/rQrYcfNZdQKjJ79KmAMxDK

### Operating system

Web app

### Typst version

### LINKED ISSUE #2480: Counter reset between usages in containers
### Description

While helping with a document on the discord server, I stumbled upon a curious bug:
```typst
#let q = counter("question")
#let step-show = {
    q.step()
    q.display("1")
}

#grid(step-show, step-show)

#pagebreak()
#step-show

#q.update(10)
#grid(step-show, step-show)
```
Will result in the following pages:

Page one:
image

Page two:
image

Removing the pagebreak, or moving the outer `step-show` before it solves the issue.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

### LINKED ISSUE #3661: Automatic numbering bug introduced when interacting with blocks
### Description

I have code to write a Mathematics test with questions and subquestions. I have boiled it down to a minimal working example to replicate the bug -- basically the subquestions should go a, b, c but they go -, -, a.

### Reproduction URL

https://typst.app/project/rmunz8o4lFgvM2UUeznpJZ

### Operating system

Web app, macOS

### Typst version

### LINKED ISSUE #4023: Citation formatting fails in table.header
### Description

When I try to compile a typst document that has a citation in a `table.header` ([link to offending commit][the-doc]), I get the following error message:
```sh
[15:03:11] compiled with errors

error: failed to format citation (this is a bug)
   ┌─ main.typ:69:42
   │
69 │   [Summary of forward vs. inverse problems @söderholm2024peeling],
   │
```
I don't know if this is a problem with `table.header` itself or in the way that I've defined a `frame` template function in the file `tau-presentation.typ`, but it would be cool if this could get resolved.

[the-doc]: https://gitlab.com/SeSodesa/tau-typst-presentation/-/commit/cefd297354da39d0f778431177c80ae15f7d9bfa

### Reproduction URL

https://gitlab.com/SeSodesa/tau-typst-presentation/-/commit/cefd297354da39d0f778431177c80ae15f7d9bfa

### Operating system

macOS

### Typst version

laurmaedje:
This PR changes how locations (essentially just unique element IDs) are assigned to elements. This has two primary motivations:

- **Correctness:** The current system is subtly broken in various ways and produces wrong results in edge cases (see bugs linked below).
- **Performance:** The new system is 100% free of side effects, making layout pure and removing the final blocker for multi-threaded layout.

The principle ideas in this PR are a few months old and were touched on in my blog post on [frozen state](https://laurmaedje.github.io/posts/frozen-state/). However, so far, progress was blocked on problems related to measurement. Ultimately, measurement of content in combination with introspection is a bit ill-defined, since the measured content doesn't necessarily end up in the document. One possible solution would have been to expose all this complexity to the user of `measure` and let them deal with it. I seriously considered this, but in the end decided that it is too much to ask from users.

Instead, I came up with a solution that I call "introspector-assisted location assignment". Essentially, during measurement we use the introspector to try to match up the locations of measured elements with real ones based on a key hash and the context's location. This works correctly almost always. Note that compiler-internal measurement does not use this best-effort mode, since it can manage the locators manually. The `content.measure` method was removed since it's now the same as `content.layout`.

For the details of the new location assignment method and also the "measurement mode", see the extensive doc comment in `src/introspection/locator.rs`.

Bugs fixed:
- Fixes #1910 (not reproducible on main, but probably some variation still was)
- Fixes #2480
- Fixes #3661
- Fixes #4023

Also makes significant progress for #1841, which can be tackled in a future PR.


crates/typst/src/introspection/locator.rs
```diff
@@ -1,117 +1,350 @@
...
+///   the same content. Using it to e.g. layout the same piece of figure content
+///   will yield the same figure number both times. Typically, when we layout
+///   something multiple times using `relayout`, only one of the outputs
+///   actually ends up in the document, while the other outputs are only used
+///   for measurement and then discarded.
+///
+/// The `Locator` intentionally does not implement `Copy` and `Clone` so that it
+/// can only be used once. This ensures that whenever we are layouting multiple
+/// things, we make an explicit decision whether we want to:
+///
```
PgBiel, line 45: Seems like something is missing or was moved here :eyes:

wrzian:
I read a bit of that massive doc comment and was thinking that it might be desirable to move it into a README at some point. Not sure what the best structure for that would be, but there should probably be a system FOR READMEs of both somewhat-polished architecture descriptions and more ad-hoc developer notes.

anechaev2109-debug:
привет_

PR #4354, opened 2024-06-08
Don't change DPI in resvg

### LINKED ISSUE #3733: SVG is wrongly cropped both horizontally and vertically when inserted via `image`
### Description

This is a regression with respect to 0.10.0, consider the following code (black background is set so that the slight white shadow is visible below the logo):
```typst
#set page(paper: "a6", margin: 0pt, fill: black)
#image("./images/python-logo.svg", fit: "contain")
```
where we want to render [python logo](https://s3.dualstack.us-east-2.amazonaws.com/pythondotorg-assets/media/files/python-logo-only.svg). This results in the following document after compilation with 0.11: 
test_11

 The same document compiled with 0.10: 
test_10

It looks like the logo is also cut if you use `#image("./images/python-logo.svg", width: 5%)` in which case we respectively get 
test_11_width
test_10_width

and I'm not sure if this is a single bug or are these two separate bugs (one regarding this weird cutting and the second one regarding `fit: "contain"` not working correctly).

 

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

LaurenzV:
Fixes #3733.

So basically, after looking into this a bit, I think that the DPI logic in `resvg` is a bit broken, or rather, it doesn't really do what we would like to do it. In `resvg`, if you change the DPI, all that is changed is the way how units such as `mm` and `pt` are interpreted, it doesn't actually affect the outlines. Here is the result of rendering the Python logo with `resvg` in the CLI when setting the DPI to 72:

test

As you can see, it applies the exact same crop that was observed inside of the issue. So the issue is that since we decreased the DPI from 97 to 72, it only affects the size of the SVG but it doesn't scale the actual outlines.

Apart from that, we are already applying the DPI inside of `typst` by changing `w` and `h` depending on what the DPI is, so I think we are kind of applying it twice, too? In any case, I think it's best to just use the default here.

(Also, how does adding tests work now? Do I just submit a PR at `typst-dev-assets`?)


LaurenzV:
Ok so I changed the image DPI to the default of usvg now.

With the following SVG:

```svg
<svg id="svg1" width="200pt" height="200pt" xmlns="http://www.w3.org/2000/svg">
    <title>Simple case</title>

    <rect id="rect1" x="0" y="0" width="200pt" height="200pt" fill="green"/>
</svg>
```

and the following document:
```typ
#set page(width: 202pt, height: 202pt, margin: 0pt)
#image("test.svg")
```

we now get the following with a small white border on the bottom-right, as expected:
typst

laurmaedje:
Thanks!

PR #4364, opened 2024-06-10
Parser trivia refactor

wrzian:
Draft refactor for discussing how the parser should handle trivia.

See detailed explanation in this thread: <https://discord.com/channels/1054443721975922748/1088371867913572452/1249492570380435496>


wrzian:
At @Myriad-Dreamin's request I have copied everything I wrote from the discord thread below:

Re, this comment: https://github.com/typst/typst/pull/3786#issuecomment-2053579667

> > Also, and this is out of scope, I would be interested in simplifying the parser interface somewhat (and adding comments/documentation) in a future PR if that would be desired.
> 
> In principle yes, but let's best first discuss it a bit, so that we're on the same page about what changes would be desirable. If you're on the Discord, doing it there in <#1088371867913572452> would be simplest.

I think the current `Parser` impl (bottom of `parser.rs`) has one particular issue that echoes out into the rest of the parser and has both performance and cleanliness implications: it doesn't abstract trivia properly for math and code mode. This core issue is exacerbated by small function naming ambiguities that make the parser harder to reason about than it should be in math/code. I don't have all the answers, but I have some partially-implemented ideas for changes to simplify the parser and would like to discuss them!

Concretely, the parser tries to abstract trivia (spaces/newlines/comments) from Math/Code mode by keeping track of `prev_end` and `current_start` indices into the overall `text` string which represent the end of the last non-trivia token (so the start of any trivia token) and the start of the current non-trivia token. (Note that Markup mode handles trivia manually during parsing, so isn't really relevant here)

The invariant that Math and Code shouldn't need to think about trivia when parsing is only enforced ad-hoc by the parser and has been the cause of a number of bugs (one of which I fixed!). I think we could improve this by representing the presence of leading trivia over the current node in the type-system via an `Option` type, and also representing the current token's text and kind together in an actual `Token` struct as the current element.

An example issue with the trivia invariant is the `wrap()` function. To get the end marker for wrapping, `wrap()` calls `before_trivia()` which does a full reverse iteration over `nodes` in order to find the last non-trivia node's index before calling `wrap_within()` and then `nodes.drain()` on the indices. If we instead just keep that index inside of a trivia Option type, then we can avoid this loop. Similar issues exist in the `unskip()` function, which is only called in three locations, that could each be simplified.

---

My main desired outcome is to make it impossible to represent the state where the parser is in math/code and `Parser:current` is a trivia node.

I'm thinking of achieving this with the following changes:
1. Remove the `skip()`, `unskip()`, `lex()` functions, and fold their behavior into a new `lex_past_trivia()` function which updates the nodes and skips trivia correctly, paired with a `re_lex()` function, that just truncates `nodes` to before the trivia, then calls `lex_past_trivia()` again from that point.
2. Replace `prev_end`, `current_start`, and `current` with two new elements
    a. `prev_end` tracked the existence of leading trivia, we instead start to track trivia in math/code with a value `trivia_start: Option<Triv{node: Marker, text_offset: usize}>`, now representing whether there was leading trivia at all, the starting index of the leading trivia node, and the starting position of the trivia in `text` for jumping the lexer.  
    b. `current_start` and `current` (in combination with the implicit `lexer.cursor()`) tracked the string slice for the current token and its kind. I think it would be better to give the parser a new struct: `current: Token{kind, text}` that greedily gets `lexer.cursor()` and is always valid. I think these `Token` structs would be well placed in the Lexer, but aren't necessary there.  
    
    (Note that these are effectively identical to leaf nodes in `nodes.rs` when they're first constructed, lacking only a default constructed Span, and with their text still referencing the parser's text, instead of as a new EcoString)

---

As for function naming, I think the ambiguity of `Parser::current()`, `Parser::eat()`, and `Parser::wrap()` adds a lot to this annoyance, especially with `Parser::eat()` not having any comments and its body literally just being `{ save(); lex(); skip(); }`.

My main note on these is that `current()` really acts like a `peek()` function, since its token *is not yet* part of the `nodes` vector. Then `save()`, just saves the current node as a leaf node, and `eat()` combines `save()` with moving the lexer forward, but also skipping trivia. In my changes, `lex()` and `skip()` would no longer be separate, and I think every place that would normally call `save()` could then be replaced entirely by just `eat()`.

I don't know that the functions need to be changed, doc-comments about their behavior will likely suffice 😁, but slight naming changes could be helpful!

wrzian:
As an aside, I was also looking into the following patch to improve the `Parser::finish()` function, but wasn't able to make it work cleanly due to failing some tests in how I believe code-mode gets re-parsed. I think it might be fixed if the parser is instantiated with `text` as the exact range that it needs to re-parse, but I haven't looked into how that's used elsewhere in-detail yet.

<details><summary>Here's the patch, apply with `git apply`.</summary>

```diff
diff --git a/crates/typst-syntax/src/parser.rs b/crates/typst-syntax/src/parser.rs
index 4a2fd5c6..d0470548 100644
--- a/crates/typst-syntax/src/parser.rs
+++ b/crates/typst-syntax/src/parser.rs
@@ -15 +15 @@ pub fn parse(text: &str) -> SyntaxNode {
-    p.finish().into_iter().next().unwrap()
+    p.finish_as_root()
@@ -21,4 +21,3 @@ pub fn parse_code(text: &str) -> SyntaxNode {
-    let m = p.before_trivia();
-    code_exprs(&mut p, |_| false);
-    p.wrap_within(m, p.marker(), SyntaxKind::Code);
-    p.finish().into_iter().next().unwrap()
+    code(&mut p, |_| false);
+    // Note that trailing trivia right here is an annoying edge case.
+    p.finish_as_root()
@@ -31 +30 @@ pub fn parse_math(text: &str) -> SyntaxNode {
-    p.finish().into_iter().next().unwrap()
+    p.finish_as_root()
@@ -828,2 +827 @@ pub(super) fn reparse_block(text: &str, range: Range<usize>) -> Option<SyntaxNod
-    (p.balanced && p.prev_end() == range.end)
-        .then(|| p.finish().into_iter().next().unwrap())
+    (p.balanced && p.prev_end() == range.end).then(|| p.finish_as_root())
@@ -1618,0 +1617,7 @@ impl<'s> Parser<'s> {
+    /// Consume the parser, yielding the root SyntaxNode of the parsed syntax tree.
+    fn finish_as_root(self) -> SyntaxNode {
+        assert!(self.current.kind == SyntaxKind::End);
+        assert!(self.nodes.len() == 1);
+        self.nodes.into_iter().next().unwrap()
+    }
+
@@ -1764 +1769,5 @@ impl<'s> Parser<'s> {
-        self.wrap_within(from, self.before_trivia(), kind);
+        if self.at(SyntaxKind::End) {
+            self.wrap_within(from, self.before_trivia(), kind);
+        } else {
+            self.wrap_within(from, self.marker(), kind);
+        }
```

</details>

crates/typst-syntax/src/parser.rs
```diff
@@ -1494,23 +1497,70 @@ fn pattern_leaf<'s>(
...
+    offset: usize,
+}
+
+/// Manages parsing of a stream of tokens into a tree of SyntaxNodes.
+///
+/// This acts like a bottom-up (shift-reduce) parser, where we investigate a current token
+/// and either:
+/// 1. Eat (shift) another token, putting `current` into the `nodes` vector.
+/// 2. Wrap (reduce) the nodes up to a marker (excluding `current`) into a corresponding
+///    SyntaxKind.
```
laurmaedje, line 1522: I'm not sure this is accurate. As far as I known, recursive descent parser is top-down and very different from shift-reduce/bottom-up. How we manage the node collection is sort of conceptually separate from the parsing structure.

crates/typst-syntax/src/parser.rs
```diff
@@ -1494,23 +1497,70 @@ fn pattern_leaf<'s>(
...
+///
+/// In the end we have a nested tree of SyntaxNodes as a Concrete Syntax Tree. This will
+/// have all irrelevant tokens (whitespace, comments, code parens, commas in function
+/// args, etc) removed when interpreted as an Abstract Syntax Tree for evaluation. The
```
laurmaedje, line 1526: "removed" here sounds like the AST is built eagerly rather than being a lazy view in top of the CST

crates/typst-syntax/src/parser.rs
```diff
@@ -1494,23 +1497,70 @@ fn pattern_leaf<'s>(
...
+    /// Nodes representing the structural syntax tree of anything previously parsed. Does
+    /// include the nodes from `prev_trivia`, but does not include `current`
     nodes: Vec<SyntaxNode>,
+    /// The stack of our lexer modes, these are pushed and popped by the `enter()` and
+    /// `exit()` functions when we switch modes.
     modes: Vec<LexMode>,
+    /// A stack of newline modes used by Code Mode to determine when an expression should
+    /// end or continue in the presence of newlines.
     newline_modes: Vec<NewlineMode>,
+    /// Stored parser checkpoints at a given text index for efficient math paren-expr
```
laurmaedje, line 1552: "math" here is a bit misleading since it sounds like math mode. Perhaps we could rather emphasize that this is for efficient parser backtracking.

crates/typst-syntax/src/parser.rs
```diff
@@ -1520,94 +1570,118 @@ enum NewlineMode {
...
-    nodes: usize,
+    current: Token<'s>,
+    prev_trivia: Option<TriviaStart>,
+    node_len: usize,
 }
 
 impl<'s> Parser<'s> {
+    /// Create a new parser starting at the given text offset and lexer mode.
+    ///
+    /// TODO idea: why not just pass in text with the offset already? (i.e. text[offset..]?)
```
laurmaedje, line 1593: because that would change `lexer.cursor()`

crates/typst-syntax/src/parser.rs
```diff
@@ -1494,23 +1497,70 @@ fn pattern_leaf<'s>(
...
+    /// parsing, see comments in `expr_with_paren` (also known as packrat parsing).
     memo: HashMap<usize, (Range<usize>, Checkpoint<'s>)>,
+    /// The stored parse results at each checkpoint.
     memo_arena: Vec<SyntaxNode>,
 }
 
 /// How to proceed with parsing when seeing a newline.
-#[derive(Clone)]
+///
+/// TODO idea: Make LexMode::Math contain the newline mode internally, then we can remove
```
laurmaedje, line 1561: do you mean `LexMode::Code`? Perhaps, not sure whether it would work, but feel free to try.

crates/typst-syntax/src/parser.rs
```diff
@@ -1628,38 +1702,46 @@ impl<'s> Parser<'s> {
...
+    /// Convert the current token's SyntaxKind and eat it.
+    fn convert_and_eat(&mut self, kind: SyntaxKind) {
+        self.current.kind = kind;
         self.eat();
     }
 
+    /// Whether there was a newline in the previous token.
+    /// Warning: only use this in Markup mode due to `prev_trivia` in Math/Code mode.
+    ///
+    /// TODO idea: maybe emit newline tokens in Code mode based on NewlineMode?
```
laurmaedje, line 1722: Since a newline is never really a proper token in the final AST, I would prefer not to have it as a syntax kind.

crates/typst-syntax/src/parser.rs
```diff
@@ -1628,38 +1702,46 @@ impl<'s> Parser<'s> {
...
     fn newline(&mut self) -> bool {
         self.lexer.newline()
     }
 
+    /// The column of the current node.
     fn column(&self, at: usize) -> usize {
+        // TODO idea: keep track of current column in the parser to make this faster?
```
laurmaedje, line 1729: probably not really in the hot path. but we can experiment with it.

crates/typst-syntax/src/parser.rs
```diff
@@ -539,8 +534,10 @@ fn math_args(p: &mut Parser) {
             ")" => break,
             ";" => {
                 maybe_wrap_in_math(p, arg, named);
-                p.wrap(array, SyntaxKind::Array);
-                p.convert(SyntaxKind::Semicolon);
+                // Must use wrap_within to include wrapping the empty math element when
+                // trivia is present.
+                p.wrap_within(array, p.marker(), SyntaxKind::Array);
```
laurmaedje, line 539: is this and the below a bug fix? if yes, it would be good to have a regression test.

crates/typst-syntax/src/parser.rs
```diff
@@ -657,7 +655,12 @@ fn embedded_code_expr(p: &mut Parser) {
     p.enter_newline_mode(NewlineMode::Stop);
     p.enter(LexMode::Code);
     p.assert(SyntaxKind::Hash);
-    p.unskip();
+    if p.had_trivia() {
+        p.expected_at(p.before_trivia(), "expression");
+        p.exit();
+        p.exit_newline_mode();
+        return;
+    }
```
laurmaedje, line 663: why did this change?

laurmaedje (commented):
One thing I've talked about with @PgBiel recently is allowing the lexer to return not just tokens but full subtrees for specific cases where it's simpler to handle at the lexer stage. This currently concerns only raw blocks, but might expand to other stuff in the future.

Do you think this makes sense and, if yes, how would it integrate with these changes, in particular with the `Token` struct?

wrzian:
> One thing I've talked about with @PgBiel recently is allowing the lexer to return not just tokens but full subtrees for specific cases where it's simpler to handle at the lexer stage. This currently concerns only raw blocks, but might expand to other stuff in the future.
> 
> Do you think this makes sense and, if yes, how would it integrate with these changes, in particular with the `Token` struct?

I would be a fan of that change. It would also be really convenient for math parsing as it could simpify both field access and primes.

I think an interesting direction could be to make `lexer::next()` just return a full `SyntaxNode`, which the parser would place as its `current` node, then we don't need a token struct and don't have to convert the node in the Parser. I think this would be reasonably clean and could probably remove the `error: Option<SyntaxError>` field from the Lexer. 

Not sure if there would be any big downsides, although this might end up adding a constant `Span::detatched()` that would be needlessly copied, so I could see this either improving or hurting performance. Would need to measure.

crates/typst-syntax/src/parser.rs
wrzian: I definitely meant to write 'code' there, I did this accidentally in a few comments.

How about this?
> Parser checkpoints for a given text index. Used for efficient parser backtracking similar to packrat parsing. See comments above in `expr_with_paren`.
Feel free to reword.

crates/typst-syntax/src/parser.rs
wrzian: Yep lol.

I ended up trying this and I think it worked out well. I'm pushing this as a new commit, see below.

crates/typst-syntax/src/parser.rs
wrzian: I'll leave this for a later date or for someone else if anyone's interested. Should I keep the TODO in the code or remove it?

crates/typst-syntax/src/parser.rs
wrzian: Did some other work on NewlineMode that removes this comment. Agreed that this isn't desirable.

crates/typst-syntax/src/parser.rs
wrzian: Agreed, was conflating the way we parse with the way we build the nodes when writing this. Trying to note the similarity to shift-reduce parsers isn't really helpful here.

How about this rewrite?
```
/// Manages parsing of a stream of tokens into a tree of SyntaxNodes.
///
/// The implementation presents an interface that can investigate a current token and take
/// one of the following actions:
/// 1. Eat a token, placing `current` into the `nodes` vector as a `LeafNode` and
///    preparing a new `current` by calling into the lexer.
/// 2. Wrap nodes from a marker to the end of `nodes` (excluding `current`) into an
///    `InnerNode` of a specific SyntaxKind.
/// 3. Produce or convert nodes into an `ErrorNode` when something expected is missing or
///    something unexpected is found.
///
/// Overall the Parser produces a nested tree of SyntaxNodes as a *Concrete* Syntax Tree.
/// The raw Concrete Syntax Tree should contain the entire source text, and is used as-is
/// for e.g. syntax highlighting and IDE features. In `ast.rs` the CST is interpreted as a
/// lazy view over an *Abstract* Syntax Tree. The AST module skips over irrelevant tokens
/// -- whitespace, comments, code parens, commas in function args, etc. -- as it iterates
/// through the tree.
```

crates/typst-syntax/src/parser.rs
wrzian: Agreed, addressed in the rewrite in the other comment.

crates/typst-syntax/src/parser.rs
wrzian: The previous code handled trivia following a hash (i.e. `#/**/var`) by just doing an `unskip` and expecting that the subsequent `code_expr_prec` call would create an error for an expected expression. In removing `unskip`  we now manually handle this case. I added a comment in the new commit

wrzian:
Just commited a change that moves NewlineMode inside the LexMode struct. I think this is a lot cleaner than it was before and really centralizes the relevant code, although there are still a few areas I'm not sure about. Let me know what you think.

PgBiel:
The `next() -> SyntaxNode` change has been implemented for #4459 , @wrzian feel free to share any possible suggestions :)

I took the approach of not changing the internal lexer stuff too much for now. That is, `take_error` is still there, as most functions were not changed (still return `SyntaxKind`). But `take_error` is now a private function in the lexer, since it eventually (at `next()`) converts the syntax kinds to nodes without any intervention from the parser.

laurmaedje:
Sorry for the slowness here, but I'd prefer to wait for https://github.com/typst/typst/pull/4459 to be merged before proceeding here.

crates/typst-syntax/src/parser.rs
wrzian: Yes? I'm not fully sure why these were needed tbh, but it fixed an issue in the math-call tests (that *I* added :upside_down_face: ). I really hate this function in particular and I'm planning a separate math-parsing PR for later anyways (3 weeks?).

wrzian:
Ok, I think this is in a good spot to review now. I didn't fully make the lexer return SyntaxNodes, but I did make the motion of the error case more explicit, although I couldn't find a better method to organize them, and I feel like the cost of making `current` into a full syntaxnode (thus requiring checking the type on every access) outweighs the benefits.

crates/typst-syntax/src/parser.rs
laurmaedje: I would prefer to have it removed unless we have some substantial reasoning or testing that this would actually have a noticable performance impact.

wrzian:
Just rebased onto main and amended a change for #4869 in the lexer mode -> closures commit.

wrzian:
Ok, I put together a simple pair of scripts to measure parser timing.

My test file is the concatenated sources of
[typst-examples-book](https://github.com/sitandr/typst-examples-book).
You can make your own version by cloning the repo and using ripgrep:
```bash
git clone git@github.com:sitandr/typst-examples-book.git examples-book

OUTFILE="typst-examples-all.typ"
rg --sort=path --pcre2 -I -U '^(`+)\s*typ(((?!\1)(?s:.))+)\1' -r '$2' > $OUTFILE
```

This gives about 5000 lines of code and markup, which feels about right
for a parser testbench. Note that the file doesn't actually compile, but
it does parse without error, which is all we need here.

On main, my laptop takes around 4400 ms to parse this. With the current
changes in this PR there's no notable difference.

I got these from the following scripts by using the existing `--timings`
flag and pulling out the delta for the "parsing file" item.

<details><summary>

Initial `run-timings.sh` script.

</summary>

Note: Requires installing `jq` for JSON parsing.

```sh
#!/usr/bin/sh
# Run timing compilation
# Usage: ./run-timings.sh BUILD_DIR

# generated from typst-examples-book sources.
BIG_FILE="typst-examples-all.typ"

echo "Finding timings for the build in: '$1'"
EXE="$1/target/release/typst"

if test ! -e "$EXE" ; then
    echo "'$EXE' doesn't exist, exiting."
    echo "Rebuild with: cargo build --release"
    exit 1
fi

TIMING_JSON="time-$(basename "$1").json"

CMD="$EXE compile --ignore-system-fonts $BIG_FILE --timings $TIMING_JSON"
echo Running: "$CMD"
$CMD
echo Output is in: "$TIMING_JSON"

NAME_SEARCH='parsing file'
echo "Getting timing for item: '$NAME_SEARCH'"
MILLIS=$(jq "map(select(.name == \"$NAME_SEARCH\") | .ts) | (.[1] - .[0]) | round" $TIMING_JSON)
echo "Ran in $MILLIS ms"
```

</details>

<details><summary>

`avg-timings.sh` to run 100 times and get the average.

</summary>

```sh
#!/usr/bin/sh
# Average over many timing runs
# Usage: ./avg-timings.sh BUILD_DIR

if test ! -e $1 ; then
    echo "'$1' doesn't exist, exiting."
    exit 1
fi

N=100
echo "Running $N times..."

OUTFILE="times-$N-$(basename "$1").json"
echo > "$OUTFILE"
for i in $(seq 1 $N); do
    MILLIS=$(./run-timings.sh "$1" 2> /dev/null | sed -E -n 's/Ran in (.+) ms/\1/p');
    echo $MILLIS >> "$OUTFILE"
done

# Why is this the best way to get an average from the shell? Ugh.
AVG=$(awk '{sum += $1; count += 1} END {printf "%i\n", sum/count}' "$OUTFILE")
echo "Average: $AVG ms"
rm "$OUTFILE"
```

</details>

crates/typst-syntax/src/parser.rs
wrzian: I'm going to commit this new wording for now.

wrzian:
Completed the change to make the Lexer return full SyntaxNodes and implemented this for Raw nodes. With the timing check I've found that returning SyntaxNode adds about a 6% performance hit to the parser, but this is likely acceptable given that eval and layout are usually the dominating performance issue.

I'll also copy another [note I made on Discord:](https://discord.com/channels/1054443721975922748/1262525590175617085/1281989305509609572)

> One problem I ran into was an annoying lifetime issue with SyntaxNode:
>
> The `Parser::current_text()` function currently returns a `&str` borrow from the original source text (the `'s` lifetime that the parser *contains*), but since SyntaxNodes own their string, changing `current_text` to use `SyntaxNode::text()` now has a lifetime matching the parser itself, which differs and causes issues.
>
> My temporary solution is to convert to an EcoString and clone it in a number of places where the `current_text()` method is used (it still returns a &str when possible). This mostly affects the hashmaps used for function arg named-parameters. It might be possible to reduce this with a lot of extra lifetime annotations in the right spots, but it would be a bit of work.

But for now, I think this is ready for another round of review!

crates/typst-syntax/src/lexer.rs
```diff
@@ -96,23 +137,46 @@ impl Lexer<'_> {
...
+    }
 
-        self.newline = false;
-        self.error = None;
+    /// Proceed to the next token and return its [`SyntaxKind`] plus a potential error.
+    /// Note the token could be a [trivia](SyntaxKind::is_trivia).
+    pub fn next(&mut self) -> SyntaxKind {
+        assert_eq!(self.raw.len(), 0);
+        assert_eq!(self.error, None);
         let start = self.s.cursor();
```
laurmaedje, line 176: Wouldn't it be cleaner if
- `next` returned a `SyntaxNode`
- The error and raw logic was moved into `next()` (or even further down into the places where it would be created, though then all the parser function would need to return nodes, which might slow things down, not sure)
- The trivia stuff stays in `lex_past_trivia` and moves back into the parser?

This way, the lexer has a clearer public API and is not tied to the `nodes` vector and `Token` struct.

crates/typst-syntax/src/lexer.rs
```diff
@@ -16,12 +16,35 @@ pub(super) struct Lexer<'s> {
...
     newline: bool,
-    /// The state held by raw line lexing.
-    raw: Vec<(SyntaxKind, usize)>,
+    /// The state held by raw line lexing. This is emptied into an inner SyntaxNode before
+    /// returning from [`lex_past_trivia`].
+    raw: Vec<SyntaxNode>,
     /// An error for the last token.
+    ///
+    /// This is present to increase convenience when returning an error by avoiding the
+    /// need to manually return an error from most lexer functions.
```
laurmaedje, line 25: Like mentioned in another comment, it feels somewhat unfortunate that we still keep `raw` and `error` even though they are not really necessary with the new design.

crates/typst-syntax/src/lexer.rs
```diff
@@ -127,24 +191,43 @@ impl Lexer<'_> {
...
+                    }
+                    _ => self.code(start, c),
+                },
             },
 
             None => SyntaxKind::End,
-        }
+        };
+        use NewlineMode::MaybeContinue;
+        if matches!(self.mode, LexMode::Code(MaybeContinue)) && kind.is_trivia() {
```
laurmaedje, line 213:
```suggestion
        if matches!(self.mode, LexMode::Code(NewlineMode::MaybeContinue))
            && kind.is_trivia()
```

crates/typst-syntax/src/lexer.rs
```diff
@@ -127,24 +191,43 @@ impl Lexer<'_> {
...
+            // Preserve `self.newline` when in MaybeContinue mode and parsing trivia.
+            self.newline = self.newline || newlines > 0;
```
laurmaedje, line 215:
```suggestion
            self.newline |= newlines > 0;
```

crates/typst-syntax/src/lexer.rs
```diff
@@ -127,24 +191,43 @@ impl Lexer<'_> {
             Some(c) => match self.mode {
                 LexMode::Markup => self.markup(start, c),
                 LexMode::Math => self.math(start, c),
-                LexMode::Code => self.code(start, c),
-                LexMode::Raw => unreachable!(),
+                LexMode::Code(mode) => match mode {
+                    // Code mode might produce false `SyntaxKind::End` tokens on newlines.
```
laurmaedje, line 195: The "might" here is a bit confusing. It sounds like that's maybe happening somewhere else and not definitely right here.

crates/typst-syntax/src/parser.rs
```diff
@@ -1824,16 +1766,28 @@ impl<'s> Parser<'s> {
...
         } else {
             self.balanced &= !kind.is_grouping();
             self.expected(kind.name());
         }
         at
     }
 
+    /// Eat the current node and convert to expected or unexpected for Some/None expected.
+    #[track_caller]
+    fn eat_as_expecting(&mut self, expected: Option<&str>) {
```
laurmaedje, line 1781: It doesn't really feel worth it to me to take `Option` here and make every caller wrap in `Some(_)` just to not copy three lines into `unexpected`.

crates/typst-syntax/src/parser.rs
```diff
@@ -1509,120 +1489,139 @@ fn pattern_leaf<'s>(
     }
 }
 
-/// Manages parsing of a stream of tokens.
+/// Manages parsing a stream of tokens into a tree of SyntaxNodes.
```
laurmaedje, line 1492: General note: Please wrap comments at 80 rather than 90 columns.

crates/typst-syntax/src/lexer.rs
wrzian: I could that, but if we did, then we would likely make most of the other lexer functions return a Result type or other SyntaxNode item. I could be convinced of this, but it would be more effort.

crates/typst-syntax/src/lexer.rs
wrzian: :cry:

crates/typst-syntax/src/lexer.rs
wrzian: How about: "Maybe produce false `SyntaxKind::End` tokens on newlines." ?

crates/typst-syntax/src/lexer.rs
laurmaedje: I would have thought just a SyntaxNode. Not sure how it affects performance, but overall I think it would reap the rewards of the whole refactor.

crates/typst-syntax/src/parser.rs
wrzian: :+1: Updated my comment wrapping extension to wrap at 80. Will fix these.

crates/typst-syntax/src/parser.rs
wrzian: I realized that this can actually be even further simplified now that we have access to the node itself by just calling `current.node.unexpected()` or `current.node.expected(thing)` *before* calling `p.eat()`. I've removed the function and pulled those out to the three callers in the new "small changes" commit.

crates/typst-syntax/src/lexer.rs
wrzian: I don't think it's going to help that much at this stage. I tried converting `Lexer::error()` into creating a  syntax node and it just adds a ton of noise everywhere. 
I also did a draft of moving the raw Vec into the raw function and having it return a SyntaxNode, but it has annoyances due to how `Lexer::code` is currently called when checking for the newline mode.

wrzian: I do think the change for math field accesses is a win. Curous what your thougts are there.

wrzian:
Ok, did some more refactoring and now the parser feels much better. I moved newline modes back into the parser and expanded them to be active within markup mode and have more options (`Indented`, `NoParBreak`). This means that the parsing of trivia is standardized across all 3 modes.
I was then able to factor out  a common 'delims()' function across various delimited groups (strong, emph, headings, lists, code blocks, etc.) that just takes the newline mode and a few syntax kinds for opening/closing delimiters. Pretty happy with the overall direction of these changes.

Also did a few miscellaneous things:
- Added a subcommand to the test runner to generate a directory of the ast outputs (run with: `test-runner write-syntax`, and diff with `git diff --no-index --stat <correct_from_main_somehow> tests/store/syntax`
- Squashed a few commits (finally got JJ working properly!), although I didn't try to simplify all that much yet.
- Made the Lexer return an enum of the `TokenType`. I'm not sure if this is necessary, but it helps make lex_past_trivia look really clean.
- Removed the Raw vector from the Lexer by moving it into the raw function itself.

I haven't tested performance too much, but with the same method as before I'm getting a roughly 5% performance hit in the parser. Would need to actually profile the code to figure out where the difference actually comes from.

wrzian:
FYI I've been working on reorganizing these commits and am effectively rewriting everything. Would you prefer if I did a force-push on this branch or create a new pull request for this?

laurmaedje:
Whichever you prefer is fine with me.

wrzian:
Alright, I'm going to close this one and reopen a new PR so we can start fresh.

PR #4366, opened 2024-06-10
Basic multi-threading

laurmaedje:
This PR adds basic support for multi-threading. Specifically, it allows Typst to multi-thread at explicit pagebreak boundaries. This means that longer documents with many explicit pagebreaks (e.g. at chapter boundaries) can saturate the CPU pretty well, while small documents see little benefit for now. This is only the initial, simplest way to multi-thread, there are also possibilities for multi-threading within page runs. This PR is actually very small, it reaps the fruits of labor done in https://github.com/typst/typst/pull/4352.

The speed-up is very document- and system-dependant. Looking at some very unscientific benchmarks for _The Thesis™_ on my MacBook Air (M1):

|        | Main (a68a241) | Threaded (152f934) |
|-------:|---------------:|-------------------:|
| Total  | ~5800 ms       | ~3000 ms           |
| Eval[^1] | ~250 ms        | ~230 ms            |
| Layout | ~4750 ms       | ~2000 ms           |
| Export[^2] | ~770 ms        | ~770 ms            |

This parallelization affects only layout, not parsing, evaluation, realization, and export. Eval still changed a bit, which demonstrates that there are some fluctuations in these naive benchmarks.

Still, we can see almost 2x total speedup and almost 2.5x speed up in layout perf. For M1, Geekbench shows ca 3.5x multi-core score compared to single-core, so if I'm interpreting this number correctly, that's the upper bound one could expect when all cores are perfectly saturated all the time. If the page runs aren't perfectly schedulable, we lose a bit of core saturation to that, a bit probably goes to synchronization, and then of course to the OS and other apps. (I didn't close any other apps ...)

I think that's a pretty nice win overall. I'd be curious to see how this plays out on other machines, in particular, desktop computers with more cores.

Note that incremental compilation is pretty much unaffected by these changes as all but one page run are typically memoized anyway.

[^1]: Note that this is only initial eval, not eval triggered by layout, which is parallel. In the thesis, eval triggered by layout is actually the most time-consuming thing because it's still using tablex.
[^2]: Export is already quite parallel on main. It's mostly busy encoding images ...

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -217,6 +217,11 @@ pub struct SharedArgs {
     /// Arguments related to storage of packages in the system
     #[clap(flatten)]
     pub package_storage_args: PackageStorageArgs,
+
+    /// Number of parallel jobs spawned during compilation,
+    /// defaults to number of CPUs.
+    #[clap(long, short)]
+    pub jobs: Option<usize>,
 }
 
 /// Arguments related to where packages are stored in the system.
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -56,6 +56,11 @@ pub struct SystemWorld {
 impl SystemWorld {
     /// Create a new system world.
     pub fn new(command: &SharedArgs) -> Result<Self, WorldCreationError> {
+        // Set up the thread pool.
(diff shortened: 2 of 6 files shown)


Dherse:
Sub 2s compile time on *The thesis™️* on a 7950x3d with SVG export, crazy results 🚀

PR #4371, opened 2024-06-11
Add support for COLRv1 emojis

LaurenzV:
Note that this still doesn't support the whole COLRv1 spec (such as some blend modes and sweep gradients), since the way we implement COLRv1 support right now is by converting it into a SVG and then rendering it.

Here is how the .png output of a file looks for example:
typst

Here is how the generated PDF looks:
[typst.pdf](https://github.com/user-attachments/files/15784522/typst.pdf)

Yeah, unfortunately it is a bit big (1.1MB), and also emojis with transparencies don't show up nicely on Apple Preview because of a bug on their side, but otherwise I think it looks pretty nice, and it's definitely better than not supporting COLRv1 at all! We can always make more optimizations later on.

P.S.: Should I create a PR at `typst-dev-assets` to add the COLRv1 Noto Color Emoji font and a test case?


LaurenzV:
I think it's around 24MB...

laurmaedje:
Yeah, subset it is.

LaurenzV:
PR is out.

LaurenzV:
@laurmaedje I linked it to the newest `typst-dev-assets` now!

crates/typst/Cargo.toml
```diff
@@ -68,6 +68,7 @@ unicode-script = { workspace = true }
 unicode-segmentation = { workspace = true }
 unscanny = { workspace = true }
 usvg = { workspace = true }
+xmlwriter = {workspace = true}
```
laurmaedje, line 71:
```suggestion
xmlwriter = { workspace = true }
```

Cargo.lock
```diff
@@ -2657,7 +2658,6 @@ dependencies = [
 [[package]]
 name = "typst-dev-assets"
 version = "0.11.0"
-source = "git+https://github.com/typst/typst-dev-assets?rev=ee8ae61cca138dc92f9d818fc7f2fc046d0148c5#ee8ae61cca138dc92f9d818fc7f2fc046d0148c5"
```
laurmaedje, line 2660: strange that it removes the information from the lock file

crates/typst/src/text/font/color.rs
```diff
@@ -33,15 +35,88 @@ pub fn frame_for_glyph(font: &Font, glyph_id: u16) -> Frame {
...
+    svg.end_element();
+
+    let mut data = svg.end_document().into_bytes();
+
+    let mut decoded = vec![];
+    if data.starts_with(&[0x1f, 0x8b]) {
+        let mut decoder = flate2::read::GzDecoder::new(data.as_bytes());
+        decoder.read_to_end(&mut decoded).ok()?;
+        data = decoded;
+    }
```
laurmaedje, line 102: This seems unnecessary since you've just encoded the document and it's surely not compressed.

crates/typst/src/text/font/color.rs
```diff
@@ -266,3 +272,356 @@ fn make_svg_unsized(svg: &mut String) {
...
+}
+
+trait XmlWriterExt {
+    fn write_color_attribute(&mut self, name: &str, ts: ttf_parser::RgbaColor);
+    fn write_transform_attribute(&mut self, name: &str, ts: ttf_parser::Transform);
+    fn write_spread_method_attribute(&mut self, method: ttf_parser::colr::GradientExtend);
+}
+
+impl XmlWriterExt for xmlwriter::XmlWriter {
```
laurmaedje, line 318: Why are these not just on `GlyphPainter`?

laurmaedje (commented):
Do any of the updated tests use COLRv1 features? They look unchanged to me.

LaurenzV:
> Do any of the updated tests use COLRv1 features? They look unchanged to me.

Yeah, otherwise I wouldn't have had to regenerate the png tests. 😄 But since they are both Noto Color Emoji, they look the same.

crates/typst/src/text/font/color.rs
LaurenzV: Good point, I copy-pasted too much. 😅

crates/typst/src/text/font/color.rs
LaurenzV: Mainly because I copy-pasted it from the `resvg` implementation, but I will change it.

Cargo.lock
LaurenzV: Weird, it got readded now, not sure what happened.

laurmaedje:
Alright. Thank you!

Issue #2123, opened 2023-09-13
Read all images from a folder and render.


RuofengX:
### Description

I want use typst to compose type that "3 screenshots per page, landscape, images from a folder".

I want typest to behavior like css but the output is pdf. When I trying to achieve this, there are some missing
future:
-  walk through a folder by given path, return a iterator which yields all file path
- render a length-unknown image path list

### Use Case

Use typst as a universal type set tool.

RuofengX:
```typst
#set page("a4", flipped: true)
#set text(10pt, style: "italic")

#let cell = rect.with(
  inset: 8pt,
)

#grid(
  columns: (auto, auto, auto),
  rows: (100%),
  gutter: 8pt,
  cell[#image("./images/1.png")],
  cell[#image("./images/2.png")],
  cell[#image("./images/3.png")],
)
```

Such code will compose a paper of 3 images in it, but I have a foldre of many screenshots.
Is there some tool could handle these scanery?

LuxxxLucy:
I recommend try with the preprocessing approach (using "typst-pyimage") first.

LuxxxLucy:
@laurmaedje do we want to implement a function for directory? 

Maybe adding a interface named "list_dir" which takes a path and returns an array?

laurmaedje:
I'm not yet completely decided, but I'm open to discussing it. It should of course be restricted to the project/package directory like file loading. And there are further design decisions about what exactly it returns (what about folders for instance: is there a way to tell whether an array item is a file or folder). And then we need a good and efficient interface for the `World` trait.

LuxxxLucy:
How about a simple design like this (two new function)

1. list_dir(path: str): returns a list of strings under the given path, which includes files and sub-folders.
2. is_dir(path: str): returns whether a path is a directory.

I think this should be a minimum design; if a path is not a directory but a filename, then we should use functions such as "read" (we might also need to add some debug log and throw the error should we `read` a directory, instead of a file).

DrakeAxelrod:
I saw within feat requests that walking a dir was requested. My question is until or if that feature gets added it there a way I could do a hacky rust implementation? My use case for moving from lualatex to typst requires me to automatic iterate a directory and sort the files based on content and then import in that order. (copypaste question from discord)

little more details:
my use case is for writing a penetration testing report in which I need to walk through a directory and import/include/sort findings based on cvss scores.

NZRosto:
[...]

PR #4380, opened 2024-06-12
Implement `read-dir`

lucifer1004:
This PR works as a POC solution for #2123, which I think is a missing basic feature, especially when writing a thesis/novel or editing a collection of works.

The design is very primitive and the implementation may introduce some bugs, but all these can be discussed and changed. A major purpose of this PR is to support those who are in favor of the directory walking feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2686,6 +2686,7 @@ name = "typst-fuzz"
 version = "0.11.0"
 dependencies = [
  "comemo",
+ "ecow",
  "libfuzzer-sys",
  "typst",
  "typst-assets",
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -204,8 +204,14 @@ impl World for SystemWorld {
         self.slot(id, |slot| slot.file(&self.root, &self.package_storage))
     }
 
-    fn directory(&self, id: FileId) -> FileResult<EcoVec<EcoString>> {
-        self.slot(id, |slot| slot.directory(&self.root, &self.package_storage))
+    fn directory(
+        &self,
+        id: FileId,
+        dir_trailing: Option<EcoString>,
+    ) -> FileResult<EcoVec<EcoString>> {
+        self.slot(id, |slot| {
+            slot.directory(&self.root, &self.package_storage, dir_trailing)
+        })
     }
 
     fn font(&self, index: usize) -> Option<Font> {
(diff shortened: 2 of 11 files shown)


Andrew15-5:
If anything, the name should be changed. It contains 2 words without any separator. The standard library has only single word names. So this is unusual. 

I have 2 propositions: `read-dir`, `directory`.
But first one is just a function like `calc.max`, while the second one is more like an element (`image`).

If we would go with the second, then it would have to be an `#[element]` that would have `children` (at the very least, maybe `readable` too).

Since directory is not a building block, it means that ultimately making it an element is useless. But I just thought of another approach. Like `image` has `image.decode` (IIRC, this can be changed in the near future), `read` can have `read.dir` or `read.directory`.

I don't know if you can add functions/methods to a function (`read` is a function while `image` is an element), but this to me feels like the best way to go in the current design.

lucifer1004:
On the Rust side, `#[func(name="read.dir")]` is viable but the exported function cannot be used in Typst since it will not be interpreted as a whole.

lucifer1004:
I renamed the function to `read-dir` in accordance with the naming convention of the standard library.

laurmaedje:
Thanks for the PR! A remaining concern of mine is the one from https://github.com/typst/typst/issues/2123#issuecomment-1717189889: How does one distinguish files from folders into which one might want to recurse?

I'm also not certain how `typst watch` would efficiently watch for file system changes that affect this. It might not turn out to be problematic, but I'd like to know for sure before introducing the functionality.

There is also the problem of non-UTF-8 paths (which I think fall into an `unwrap_or_default()` here), but it's maybe not a huge deal (though we should definitely handle them more gracefully that returning an empty string).

Andrew15-5:
> How does one distinguish files from folders into which one might want to recurse?

I think the simplest solution will be to use `/` identifier at the end (`ls -p`). It won't complicate the code (just a bit), but the output of the function will be a bit harder to parse. Like if you want to get just the names, then you would have to strip the `/` manually (although this can be a switch that is turned off by default).

> I'm also not certain how `typst watch` would efficiently watch for file system changes that affect this.

I think it's exactly the same as using `inotifywait -r .` which outputs `Setting up watches.  Beware: since -r was given, this may take a while!`. It will be a performance problem if you watch your home directory, but even a big Typst project wouldn't be that big to notice a performance hit, I think. In fact, `typst watch` already acts like `inotifywait -r .`. Right now, I can only think of a problem that this function call can't be memoized. Or you would have to store metadata about currently tracked files, and when `typst watch` sees a new/changed/removed file and `#read-dir()` is checking the path where the change occurred, then it will rerun the function. If the function (1) checks a different path or (2) a new change is just some file content has been changed (no new/removed files), then it won't be rerun.

I can't say anything concrete about the UTF-8 problem, but I think that one of those `lossy` methods can be used and a warning/note in the documentation can be added that the invalid UTF-8 names will drop the invalid bytes. Maybe the other solution would be to add (another) switch that will control whether the output will be an array of the lossy UTF-8 `str`ings, or an array of `bytes` elements.

lucifer1004:
1. I just added a `dir-trailing` option which defaults to `none`. This can be used to distinguish sub-directories from files. I tried to add some tests, but found that the handling of assets is specialized. I may need some help here.

2. For `typst watch`, when there is no change to the source file, change of directory contents will not trigger a re-build. But I think this is minor, since we only need to perform a no-op (type in something and delete, then save) to trigger it.

3. For non UTF-8 directory name, the problem is that we cannot type in the exact name (using other encodings) in a Typst source file.      

4. In the current implementation, `FileError` is reused, but adding `FileError::IsFile` is kind of ridiculous. I would like your ideas on how to express this type of error (trying to read a file as a directory) more clearly.

istudyatuni:
I don't like `dir-trailing`, I think it won't play well with path/resource type, when it will be implemented

I suggest returning some `path` type, which will hold internally `PathBuf`, will have `is-dir`, `is-file`, will be loosely converted to string, and when passed to functions like `#image`, will be treated as-is, allowing for non-utf8 paths

laurmaedje:
I agree that a `path` type is the better option. I think we should defer this until we have a solution for paths/resources worked out and implemented. I'm sorry to be blocking this, but I think a little patience will reward us with a cleaner solution in the end. Still, thanks for your work @lucifer1004!

Issue #4174, opened 2024-05-17
Export target elements


tingerrr:
### Description

I propose adding `document`-like `set`-rule only elements for export targets to control target specific settings:
- `png`
  - controlling PPI from within Typst documents
    - although this may open a small DOS attack surface from template packages choosing astronomical values
- `pdf`
  - PDF document metadata, which is currently set using `document` 

Use `document` either for only the common settings or as a convenient default for various settings from multiple targets.

Alongside this, a variable such as `target` should be added, which is set depending on the current export target. This allows code to act according to the export target, preventing incorrect output or panics when a target specific feature is used, such as Forms #2421 or OCGs #2613.

The newly added elements and variable could be put into an `export` module, i.e. `set export.pdf(...)`, `if export.target == export.pdf { ... }`, as they should not require easy access for the majority of users.

While I thought about the target variable being the element of the active export target for nice looking checks (`target == pdf`), this comes with some downsides:
- elements with no fields must be available too (`svg`)
- sub-targets, such as #2942 are not easily accounted for unless they receive potentially useless sub-elements like `pdf.a2` or `pdf.x`
- if sub-elements are added, they would complicate the simple checks as the equality would no longer hold `target == pdf` would be false for `target == pdf.a2`, users would need to use `target in (pdf, pdf.a2, pdf.x)`, listing all possible sub-targets

Using strings solves this more easily, as these sub-target checks could use string methods like `target.starts-with("pdf")` or even `target.starts-with("pdf/a")` for all versions of PDF/A.

Most of the sub-target specific settings will likely live on the "encompassing" element anyway (i.e. pdf for PDF/A).

### Use Case

1. Export target specific features which are likely to be more "low-level" than normal Typst features if they can't be easily generalized will likely need packages to be made usable for the average Typst user.

   Such packages need to be able to:
   - use export target specific features without panics (if they do when used on an unsupported target)
   - provide reasonable fallback behavior, when the target specific feature is not supported
     - this is especially important for preview of features which are not available for PNGs, such as forms
     - provide example images of target specific features without the need to take screenshots of PDF viewers to show of export target specific packages

1. Export targets such as HTML #721 likely need more export target specific code, as it is fundamentally different from how Typst currently works.

emilyyyylime:
`target` sounds like it'd fit inside the `sys` module, but I'm not too against `export` either

laurmaedje:
I think the question of a `target` variable and elements for setting target-specific metadata is fairly separate.

[...]

PR #1904, opened 2023-08-09
Add targets

SekoiaTree:
This PR adds support for targets, which allow a document to know what it's being compiled into. 

This is part of the basics that will eventually allow HTML compilation.

Currently missing:
- tests (might not be needed? They don't really work with the existing testing pipeline, since they're meta)
- clippy. So much clippy (probably)

PR #4385, opened 2024-06-12
Add `sys.target` export targets

tingerrr:
This PR adds a `sys.target`, a string variable containing `pdf`, `svg` or `raster` depending on the export target.

It should be noted that, with this PR in its current stage, compilation is run 3 times per average test on `--extended`. It may be more sensible, but less intuitive to contributors, if we assume a default target of `raster` and then only override the target for those test which have an appropriate annotation.

### TODO
- [ ] allow custom export targets?
- [x] finalize design for CLI interface
- [x] write documentation
- [ ] allow contextual override

Related to #4174, partially resolves it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request, merge_group]
 env:
   RUSTFLAGS: "-Dwarnings"
   RUSTDOCFLAGS: "-Dwarnings"
-  TYPST_TESTS_EXTENDED: true
+  TYPST_TESTS_ALL: true
 
 jobs:
   # This allows us to have one branch protection rule for the full test matrix.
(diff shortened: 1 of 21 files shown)


laurmaedje:
Relevant prior art: https://github.com/typst/typst/pull/1904

tingerrr:
I have renamed `--format` on `query` to `--output-format` to avoid the clash with `--format` while maintaining a consistent CLI.

tingerrr:
I've changed the test runner to have three options, one for each export target, running only those that are selected. By default, `--raster` is implied, but when another export target is requested `--raster` no longer is. `--extended` was changed to `--all` to reflect this change in behavior.

The README contains a note explaining that only render tests are compared.

laurmaedje:
As discussed on Discord, we probably need a bit more fine-grained approach (see [here](https://github.com/typst/typst/issues/4174#issuecomment-2165246219)), so I'll close this for the time being.

PR #4390, opened 2024-06-13
Paragraph spacing property

laurmaedje:
This introduces a new `par.spacing` property. Paragraphs are [not considered `block`s anymore](https://github.com/typst/typst/pull/4310) and are not affected by `block.spacing`. Instead `block.spacing` now delegates to `par.spacing`, when it is `auto` (the new default). If, however, it is set to a non-auto value, it has precedence over paragraph spacing.

This change has the following benefits:
- Changing the paragraph spacing (and by extension the default block spacing) becomes simpler in the common case: `#set par(spacing: ..)` instead of `show par: set block(spacing: ..)`.
- Previously, `block.above` and `block.below` created spacing with higher priority (and thus possibly different results) than `block.spacing` because otherwise it was impossible to override a larger spacing (e.g. for a heading to reduce the spacing below it). This was confusing. In the new system, `spacing` is really just a shorthand for `above` and `below` since these naturally have higher priority than paragraph spacing (and thus also `auto` block spacing, which is the same).
- Because we can remove the internal hack, `block.above` and `block.below` are now available in context expressions. `par.spacing` is of course available as well.

Moreover, I changed `list.spacing`, `enum.spacing`, and `terms.spacing` to select `leading`/`spacing` based on their tightness when set to `auto`, but to always respect the set value otherwise. Previously tight lists always ignored the `spacing` property. I believe the original reasoning was that `set list(spacing: ..)` shouldn't break tight lists, but this problem is better solved with `show list.where(tight: false): set list(spacing: ..)` in my opinion. 

I also improved the docs on `par.leading` and fixed a small bug with `quote` related to its block spacing.


istudyatuni:
> If, however, it is set to a non-auto value, it has precedence over paragraph spacing

When both `block.spacing` and `par.spacing` are defined, which one will take the precedence?

laurmaedje:
> When both block.spacing and par.spacing are defined, which one will take the precedence?

I'm not sure what exactly you mean. Maybe this explains it:

`block.spacing` has type `auto | length` (default: `auto`), while `par.spacing` has type `length` (default: `1.2em`). `block.spacing` falls back to `par.spacing` when it is `auto`. Both are always defined, it's just that `block.spacing` can be `auto`.

When two paragraphs touch, the larger `par.spacing` wins. When two blocks touch, the larger of the first's `block.below` and the second's `block.above` wins. When a paragraph and a block touch, the block spacing always wins (unless it is `auto`, then it's as if two paragraphs touched).

Andrew15-5:
This was mentioned in https://forum.typst.app/t/why-bottom-pad-in-paragraph-breaks-left-pad-in-quote/600/12.

memeplex:
The new behavior of list is not clear from https://staging.typst.app/docs/reference/model/list/#parameters-spacing:

> If this is false, the items are spaced apart with [list spacing](https://staging.typst.app/docs/reference/model/list/#parameters-spacing). If it is true, they use normal [leading](https://staging.typst.app/docs/reference/model/par/#parameters-leading) instead. This makes the list more compact, which can look better if the items are short.

I think the documentation should be improved, especially because the behavior has changed. An example like `show list.where(tight: false): set list(spacing: ..)` would be helpful.

Andrew15-5:
You have linked to the wrong parameter. The quote is from https://staging.typst.app/docs/reference/model/list/#parameters-tight.

PR #4394, opened 2024-06-13
Allow "imprecise" bounds for the compiler version

### LINKED ISSUE #4379: Allow "imprecise" version numbers for specifying a package's compiler compatibility
### Description

The `compiler` field in `typst.toml` currently requires a "full" version number of the form `X.Y.Z`. Since this is actually a lower bound, allowing e.g. `X.Y` would make sense. There are even richer forms of specifying semver ranges, but they are probably overkill, especially as long as Typst is in the `0.Y.Z` range.

### Use Case

currently, trying to import a package with a line like `compiler = "0.11"` in `typst.toml` will result in an error like this:

```
error: package manifest is malformed (version number is missing patch version)
  ┌─ gallery/example.typ:1:8
  │
1 │ #import "@local/package:0.0.1"
  │         ^^^^^^^^^^^^^^^^^^^^^^
```

Other than allowing this reasonable version specification, this change would allow for more descriptive errors specific to version _bounds_ in the manifest.

SillyFreak:
Fixes #4379. As suggested there, this is a direct implementation of the comparison logic without using a library. The comparisons are described and tested with doc tests, and I did a few manual end-to-end tests:

When compiling a file containing a single import to a local package, the following `compiler` field values in that packages have the following error messages:
- `compiler = "0.11.0"` - _no error_
- `compiler = "0.11"` - _no error_
- `compiler = "0.11.1"` - package requires typst 0.11.1 or newer (current version is 0.11.0)
- `compiler = "0.12.0"` - package requires typst 0.12.0 or newer (current version is 0.11.0)
- `compiler = "0.12"` - package requires typst 0.12 or newer (current version is 0.11.0)
- `compiler = "0.11.0.0"` - package manifest is malformed (version bound has unexpected fourth component: `0`)
- `compiler = "0.11.x"` - package manifest is malformed (`x` is not a valid patch version bound)
- `compiler = ""` - package manifest is malformed (`` is not a valid major version bound)

I didn't put "compiler" anywhere into the error messages, as the `SemVerBound` is not specific to that - so the error messages are only slightly improved, but the original error condition is removed. The remaining error messages contain a part of the version bound, so it's more likely that one will realize what field the error is about.

Some remarks regarding the code:
- the fields of `SemVerBound` are public, like with `PackageVersion`. It's possible to initialize an invalid value by giving a patch version but not a minor one, so it might make sense to change these fields to private and add a constructor.
- the error message for `compiler = "0.12"` says "typst 0.12 or newer", not "0.12.0 or newer". If that's considered a problem, I could add a conversion to `PackageVersion` for this.

### DIFF
--- a/crates/typst-syntax/src/package.rs
+++ b/crates/typst-syntax/src/package.rs
@@ -40,7 +40,7 @@ pub struct PackageInfo {
     /// The path of the entrypoint into the package.
     pub entrypoint: EcoString,
     /// The minimum required compiler version for the package.
-    pub compiler: Option<PackageVersion>,
+    pub compiler: Option<SemVerBound>,
 }
 
 impl PackageManifest {
@@ -62,7 +62,7 @@ impl PackageManifest {
 
         if let Some(required) = self.package.compiler {
             let current = PackageVersion::compiler();
-            if current < required {
+            if !current.matches_ge(&required) {
                 return Err(eco_format!(
                     "package requires typst {required} or newer \
                      (current version is {current})"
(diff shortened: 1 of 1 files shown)


bluebear94:
Is there a practical difference between `compiler = "0.11.0"` and `compiler = "0.11"`?

SillyFreak:
There isn't, but for version bounds the more specific parts (minor and patch) are usually optional. As the error message for a missing part wasn't great (see #4379), some improvement seemed useful and this was easier than trying to improve the error message.

See also [this Discord message](https://discord.com/channels/1054443721975922748/1088371867913572452/1250189635087368263) for context.

SillyFreak:
@frozolotl that would be a breaking change though. I did it this way to preserve the current behavior, except for making the toml field more relaxed and slightly improving the messages. Changing the way how the `compiler` field works should imo be a separate issue.

laurmaedje:
> I don't think greater or equal is the bound to choose here. It would mean that Typst 0.12.0 is fully compatible with Typst 0.11.0, which is not the case. I think we should instead use the ^ operator by default, like Cargo does.

Changing the `compiler` field to work like a `^` dependency would be _very_ bad. It would mean that releasing a new compiler version would automatically break the package ecosystem for no reason. Sure, a package _might_ be broken by a compiler update, but this would mean that it's guaranteed to be broken.

frozolotl:
> > I don't think greater or equal is the bound to choose here. It would mean that Typst 0.12.0 is fully compatible with Typst 0.11.0, which is not the case. I think we should instead use the ^ operator by default, like Cargo does.
> 
> Changing the `compiler` field to work like a `^` dependency would be _very_ bad. It would mean that releasing a new compiler version would automatically break the package ecosystem for no reason. Sure, a package _might_ be broken by a compiler update, but this would mean that it's guaranteed to be broken.

True, that makes sense.

SillyFreak:
> It would mean that releasing a new compiler version would automatically break the package ecosystem for no reason.

I was thinking about whether this only applies as long as Typst is pre-1.0, but even after that, caret semantics would not be a good fit for compiler compatibility: If there ever is Typst 2.0, you _still_ wouldn't want to need to update all packages just to "certify" that they're compatible. It would not only split the ecosystem in the inevitable way that breaking changes bring, but also artificially.

A caret dependency allows _automatically_ using a newer version of something without worrying about breaking changes. Since newer compilers aren't used automatically, using caret semantics for the compiler field doesn't actually provide a benefit.

crates/typst-syntax/src/package.rs
```diff
@@ -216,6 +216,171 @@ impl PackageVersion {
...
+    /// assert!(!v1_1_1.matches_gt(&SemVerBound::from_str("1.1.1").unwrap()));
+    /// assert!(!v1_1_1.matches_gt(&SemVerBound::from_str("1.1.2").unwrap()));
+    /// ```
+    pub fn matches_gt(&self, bound: &SemVerBound) -> bool {
+        if self.major != bound.major {
+            return self.major > bound.major;
+        }
+        let Some(minor) = bound.minor else {
+            return false;
+        };
```
laurmaedje, line 283:
```suggestion
        let Some(minor) = bound.minor else { return false };
```
if we remove the semicolon, it formats a little nicer in my opinion

crates/typst-syntax/src/package.rs
```diff
@@ -216,6 +216,171 @@ impl PackageVersion {
     }
 }
 
+impl PackageVersion {
+    // the code in this impl block is based on https://github.com/dtolnay/semver/blob/1.0.23/src/eval.rs
+    // author: David Tolnay <https://github.com/dtolnay>
+    // licensed under Apache 2.0
```
laurmaedje, line 222: If the code is really close enough to the original, then we'd also need an entry in the NOTICE file. though to me it looks fairly different.

crates/typst-syntax/src/package.rs
```diff
@@ -216,6 +216,171 @@ impl PackageVersion {
...
+    /// assert!(v1_1_1.matches_eq(&SemVerBound::from_str("1").unwrap()));
+    /// assert!(!v1_1_1.matches_eq(&SemVerBound::from_str("2").unwrap()));
+    ///
+    /// assert!(!v1_1_1.matches_eq(&SemVerBound::from_str("1.0").unwrap()));
+    /// assert!(v1_1_1.matches_eq(&SemVerBound::from_str("1.1").unwrap()));
+    /// assert!(!v1_1_1.matches_eq(&SemVerBound::from_str("1.2").unwrap()));
+    ///
+    /// assert!(!v1_1_1.matches_eq(&SemVerBound::from_str("1.1.0").unwrap()));
+    /// assert!(v1_1_1.matches_eq(&SemVerBound::from_str("1.1.1").unwrap()));
+    /// assert!(!v1_1_1.matches_eq(&SemVerBound::from_str("1.1.2").unwrap()));
```
laurmaedje, line 242: The doc tests are a bit much imo. The total PR diff is a lot of lines for such a small thing. Also doc tests are slow to compile. I'd prefer a few (like ten assertions) in a `#[cfg(test)]` module at the bottom.

crates/typst-syntax/src/package.rs
```diff
@@ -216,6 +216,171 @@ impl PackageVersion {
...
+        if bound.minor.map_or(false, |minor| self.minor != minor) {
+            return false;
+        }
+        if bound.patch.map_or(false, |patch| self.patch != patch) {
+            return false;
+        }
+        true
+    }
+
+    /// Performs a `>` match with the given version bound. The match only succeeds if some version
```
laurmaedje, line 257: Please wrap at 80 columns. Thanks. :)

crates/typst-syntax/src/package.rs
```diff
@@ -265,3 +430,73 @@ impl<'de> Deserialize<'de> for PackageVersion {
         string.parse().map_err(serde::de::Error::custom)
     }
 }
+
+/// A version bound for compatibility specification.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct SemVerBound {
```
laurmaedje, line 436: I'd prefer the name `VersionBound`, since our version type is also not strictly semver (less powerful).

crates/typst-syntax/src/package.rs
```diff
@@ -216,6 +216,171 @@ impl PackageVersion {
...
+        if self.major != bound.major {
+            return false;
+        }
+        if bound.minor.map_or(false, |minor| self.minor != minor) {
+            return false;
+        }
+        if bound.patch.map_or(false, |patch| self.patch != patch) {
+            return false;
+        }
+        true
```
laurmaedje, line 254:
```suggestion
        self.major == bound.major
            && bound.minor.map_or(true, |minor| self.minor == minor)
            && bound.patch.map_or(true, |patch| self.patch == patch)
```

crates/typst-syntax/src/package.rs
SillyFreak: yeah, I wanted to be on the safe side and at least give you the opportunity to judge for yourself. If you agree that it's fairly different (algorithmically it's similar, but the code itself in many parts isn't) then I'll remove that note.

laurmaedje: yeah, afaik copyright doesn't cover algorithms, just code, so I think we don't need it. that aside, the algorithm is also pretty obvious.

laurmaedje:
Thanks!

Issue #4398, opened 2024-06-14
`raw()` function doesn't add `block` field to the `raw` element (unlike the raw syntax sugar)


Andrew15-5:
### Description

```typ
#let a = ``.block
#let b = raw("").block
```

Both `a` and `b` should be `false`, but the second line gives the next error:

```
error: content does not contain field "block"
  ┌─ a.typ:2:17
  │
2 │ #let b = raw("").block
  │                  ^^^^^
```

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

Andrew15-5:
If it's easy to fix, then I want to do it myself (with a hint at the problem place).

laurmaedje:
This is working as intended. Fields are only present if explicitly set, so that they can still be affected by set rules otherwise. It's the same for all other elements. In a show rule, the field will also be populated since the style context is known by then.

PgBiel:
Maybe we can have a hint for this specific case (the field normally exists but was omitted by the constructor, so it's not yet known)?

laurmaedje:
Sure

Andrew15-5:
> In a show rule, the field will also be populated since the style context is known by then.

I don't know what you mean, since this is not true:

![image](https://github.com/typst/typst/assets/37143421/ceb34401-e8cd-402d-af97-e2d3717485c8)
![image](https://github.com/typst/typst/assets/37143421/9e7ebe1a-a609-4059-831f-4f2fd8552714)

PgBiel:
`elem` doesn't seem to be the subject of the show rule in that case (Laurenz was referring specifically to the case where the `raw` comes from `it` in `#show raw: it => ...` instead of from `#let myraw = raw(...)`).

bluebear94:
> Maybe we can have a hint for this specific case (the field normally exists but was omitted by the constructor, so it's not yet known)?

Maybe something like `field {} in content is not set at this point`?

Andrew15-5:
> Laurenz was referring specifically to the case where the `raw` comes from `it` in `#show raw: it => ...`.

Mmm, yeah, I have a general document show rule instead.

PR #4399, opened 2024-06-15
Add message when trying to access a field that is not set

bluebear94:
See #4398 for discussion.


crates/typst/src/foundations/element.rs
```diff
@@ -124,7 +126,11 @@ impl Element {
     }
 
     /// Extract the field name for the given field ID.
```
bluebear94, line 128: This function seems to get the field’s value, not its name; is that correct?

crates/typst-macros/src/elem.rs
```diff
@@ -977,27 +977,27 @@ fn create_fields_impl(element: &Elem) -> TokenStream {
...
             }
 
-            fn field(&self, id: u8) -> Option<#foundations::Value> {
-                let id = Fields::try_from(id).ok()?;
+            fn field(&self, id: u8) -> Result<#foundations::Value, #foundations::FieldAccessError> {
+                let id = Fields::try_from(id)?;
                 match id {
                     #(#field_arms,)*
-                    _ => None,
+                    _ => Err(#foundations::FieldAccessError::Internal),
```
laurmaedje, line 984: The `None` in these functions can also indicate that the field doesn't exist because depending on the function the matchers above aren't exhausive. If you try `#hide[].hidden` for instance, then you'll get the "this is a bug" message. I think we can just return `MissingField` and remove `Internal`.

crates/typst/src/foundations/content.rs
```diff
@@ -961,18 +965,39 @@ pub trait PlainText {
...
+            }
+            FieldAccessError::MissingFieldNoDefault => {
+                eco_format!(
+                    "content does not contain field {} and \
+                 no default value was specified",
+                    field.repr()
+                )
+            }
+            FieldAccessError::FieldNotSet => {
+                eco_format!("field {} in content is not set at this point", field.repr())
```
laurmaedje, line 993: the wording here could be slightly confusing if a person has written a set rule right above and thinks that this will have an effect. perhaps "is not known" at this point? we could also add a hint, but I have no idea how to explain where it is available in the length of a hint.

laurmaedje: We might also want to reword the variants and message perhaps:

- `MissingField: content does not contain field {}` could be `Unknown: heading does not have field {}`

- `FieldNotSet: field {} in content is not set at this point` could be `Unset: field {} in heading is not yet known at this point`

crates/typst-macros/src/elem.rs
bluebear94: I’ve fixed it for `Fields::{field, field_with_styles, field_from_styles}`, but I’m not sure whether this can happen for `impl TryFrom<u8> for Fields`.

laurmaedje: It can't, this would indicate a bug. I'm okay with keeping the variant.

laurmaedje:
Thank you!

PR #4400, opened 2024-06-15
Add hints to array destructuring error messages

MDLC01:
This small PR adds hints to array destructuring error messages indicating the length of the array. For example: "the provided array has a length of 3, but the pattern expects 2 elements".

### DIFF
--- a/crates/typst/src/eval/binding.rs
+++ b/crates/typst/src/eval/binding.rs
@@ -89,7 +89,6 @@ fn destructure_array<F>(
 where
     F: Fn(&mut Vm, ast::Expr, Value) -> SourceResult<()>,
 {
-    let destruct_len = destruct.items().count();
     let len = value.as_slice().len();
     let mut i = 0;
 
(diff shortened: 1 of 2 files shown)


crates/typst/src/eval/binding.rs
```diff
@@ -89,23 +89,30 @@ fn destructure_array<F>(
 where
     F: Fn(&mut Vm, ast::Expr, Value) -> SourceResult<()>,
 {
+    let destruct_len = destruct.items().count();
```
laurmaedje, line 92: why was this moved up here? it doesn't seem to be used anywhere but `Spread`, but now we unnecessarily traverse the destructuring AST twice.

crates/typst/src/eval/binding.rs
```diff
@@ -89,23 +89,30 @@ fn destructure_array<F>(
...
     let mut i = 0;
 
     for p in destruct.items() {
         match p {
             ast::DestructuringItem::Pattern(pattern) => {
                 let Ok(v) = value.at(i as i64, None) else {
-                    bail!(pattern.span(), "not enough elements to destructure");
+                    bail!(
+                        pattern.span(), "not enough elements to destructure";
+                        hint: "the provided array has a length of {}", len
```
laurmaedje, line 102:
```suggestion
                        hint: "the provided array has a length of {len}",
```

crates/typst/src/eval/binding.rs
MDLC01: This is indeed unnecessary.

laurmaedje:
Thanks!

Issue #4263, opened 2024-05-26
Add hint to type errors related to shadowed built-in names


PgBiel:
### Description

Some errors, especially type errors, can be trivially connected to an accidentally overridden built-in name, as seen in https://github.com/typst/typst/issues/3176, or more recently in a Discord support thread[^1]. For example, if you write

```typ
#let image = "something"

// later

#image("file.typ")
```

You will get an error message saying that `image` should have been a function if we wanted to call it, but it's actually a string. However, this only happened because it's a built-in function used very often, so we expected it to be callable.

Therefore, I propose such errors (whichever errors we can unambiguously trace back to the wrong usage of a shadowed built-in name, so, mostly type errors) include, as a hint, that they have overridden the built-in name, and may wish to either use the new `std.image` syntax, or rename their variable which is shadowing the built-in name.

Of course, this can produce a few false-positives, but I think the overall benefit might outweigh that rare chance, especially because it is only a hint to an error (the error happens and the compilation is interrupted regardless). In addition, it'd be nice to have this hint for packages' functions too (or, more generally, any shadowed functions), but that'd be much harder to implement in an unambiguous and consistent way, so I think implementing this only for shadowed built-in functions and types might be enough for now.

[^1]: https://discord.com/channels/1054443721975922748/1088371919725793360/1244335344778870785

### Use Case

It can be pretty easy to forget you shadowed a built-in name at the top of your file after writing a lot of stuff under it, so I think this could be helpful.

laurmaedje:
How so? Can't we still look at the identifier for the error span and check it against the std scope?

bluebear94:
You’re right. Almost forgot about that.

Andrew15-5:
> use the new `std.image` syntax

There is a new syntax?

PgBiel:
> > use the new `std.image` syntax
> 
> There is a new syntax?

See: https://github.com/typst/typst/pull/4038

Andrew15-5:
Wait, it's not in the v0.11.1? The release was 12 days ago, but the merge was last month.

PgBiel:
v0.11.1 was a patch release and only included bug fixes.

Andrew15-5:
Ohh, it has a separate commit history, that explains it.

laurmaedje:
Fixed by https://github.com/typst/typst/pull/4402

PR #4402, opened 2024-06-15
Hint for shadowed std functions

Coekjan:
Related: #4263 

This PR adds a simple hint for shadowed std functions:

```typ
#let image = "image";

#image("ABC") // Hint here
```

We may need to hint for other shadowed things in the future, e.g. builtin variables, modules, so this PR shall not close that issue.

### DIFF
--- a/crates/typst/src/eval/call.rs
+++ b/crates/typst/src/eval/call.rs
@@ -128,7 +128,7 @@ impl Eval for ast::FuncCall<'_> {
             (callee.eval(vm)?, args.eval(vm)?.spanned(span))
         };
 
-        let func_result = callee.clone().cast::<Func>().at(callee_span);
+        let func_result = callee.clone().cast::<Func>();
         if in_math && func_result.is_err() {
             // For non-functions in math, we wrap the arguments in parentheses.
             let mut body = Content::empty();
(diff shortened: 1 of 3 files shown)


crates/typst-syntax/src/ast.rs
```diff
@@ -1563,6 +1563,11 @@ node! {
 }
 
 impl<'a> FuncCall<'a> {
+    /// The function identifier.
+    pub fn ident(self) -> Ident<'a> {
```
laurmaedje, line 1567: The callee may not be an identifier, so this operation is a bit ill-defined. You should match on `callee` instead.

laurmaedje (commented):
I wonder whether we can somehow solve this more generally, so that we don't have to spread this all over the place for functions, modules, etc. Seems tricky though.

Coekjan:
> I wonder whether we can somehow solve this more generally, so that we don't have to spread this all over the place for functions, modules, etc. Seems tricky though.

Thanks for your feedback!

I believe that situations involving shadowed functions are more common and easier to trigger than those involving shadowed variables or modules. Therefore, I think we should prioritize providing hints for shadowed functions. After all, a general solution may not be as straightforward to conceive and implement.

crates/typst-syntax/src/ast.rs
Coekjan: Thanks for the review.

I see `#"A"("ABC")` can be parsed to `FuncCall` and it has no identifier, so this operation is not good indeed. To make it better, I turned to let this function return `Option<Ident<'a>>` instead just now.

laurmaedje: why not just `match call.callee() { ast::Expr::Ident(ident) => .., _ => .. }` ?

Coekjan: Indeed. I've modified the code according to your suggestion.

laurmaedje:
> Therefore, I think we should prioritize providing hints for shadowed functions.

Fair enough. Thanks!

crates/typst/src/eval/call.rs
```diff
@@ -148,21 +148,21 @@ impl Eval for ast::FuncCall<'_> {
...
                 if let ast::Expr::Ident(ident) = self.callee() {
                     let ident = ident.get();
                     if vm.scopes.check_std_shadowed(ident) {
                         err.hint(eco_format!(
-                            "use `std.{}` to access the shadowed standard library function",
-                            ident,
-                        ));
+                        "use `std.{}` to access the shadowed standard library function",
+                        ident,
+                    ));
```
Coekjan, line 159: @laurmaedje indentations here seems broken :ghost:

PR #4412, opened 2024-06-18
docs: add reference for list of latex math symbols and typst equivalents

tobiasdiez:
I stumbled upon http://mirrors.ctan.org/info/typstfun/typstfun.pdf which I found pretty helpful in migrating from Latex to typst. Hence, I've added it to the docs.
Upstream repo: https://github.com/lvjr/typstfun

### DIFF
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -499,6 +499,8 @@ their name. Some symbols are additionally available through shorthands, such as
 Refer to the [symbol pages]($reference/symbols) for a full list of the symbols.
 If a symbol is missing, you can also access it through a
 [Unicode escape sequence]($syntax/#escapes).
+A pretty complete list of LaTeX math symbols and their Typst equivalent can be
+found in the community-managed [typstfun PDF](http://mirrors.ctan.org/info/typstfun/typstfun.pdf).
 
 Alternate and related forms of symbols can often be selected by
 [appending a modifier]($symbol) after a period. For example,


Enivex:
In my opinion it makes more sense to have this information in the symbol list instead https://typst.app/docs/reference/symbols/sym/

tobiasdiez:
I agree, this would be better - but also way more work that I currently cannot invest.

Coekjan:
Not sure if the upstream repo (typstfun) will be actively maintained.

Typst is still young, and currently the symbol names are not very stable (you can see the breaking changes on symbol names since 0.11 in https://github.com/mitex-rs/mitex/issues/165). In the future, if that repo is not synchronized with typst, the document here could be misleading. And I would suggest to maintain the symbol-equivalents document officially by typst team, if it is possible.

laurmaedje:
What's the best list for Unicode <-> LaTeX mapping? Given that, I can add it to the symbol docs page.

Enivex:
> What's the best list for Unicode <-> LaTeX mapping? Given that, I can add it to the symbol docs page.

https://github.com/wspr/unicode-math/blob/master/unicode-math-table.tex

Same list I used for the spreadsheet I created

laurmaedje:
I added the LaTeX names to the symbol list and to the search index.

tobiasdiez:
> I added the LaTeX names to the symbol list and to the search index.

I used it a couple of times last week, and it works really well! Thanks a lot 🚀

PR #3728, opened 2024-03-19
Localization with custom key-value format

Heinenen:
This is a follow-up PR/replacement for #3094 and aims to improve the handling of localization in Typst. 
It relies on https://github.com/typst/typst-assets/pull/2 to be merged.
The main benefit of this PR is the separation between translation strings and code.
Previously, the localized strings of one language were spread across 7 files.

As discussed in #3094, this PR implements a simple, custom key-value format for translations.
This is less complex than using Fluent or other localization systems, which are overkill for our current needs.
If this changes, switching to Fluent is now easier than without this PR.

The format is the following, whitespace around `key` and `value` are ignored:
```
# this is a comment, only valid on its own line
key = value
```

PR #4414, opened 2024-06-18
Add Galician & catalan translations

JonPichel:
This PR adds Galician translations to the Typst compiler. Furthermore, it recovers the Catalan translations from #3181 that got lost during the change to the custom key-value localization format in #3728.


laurmaedje:
Thank you!

PR #4424, opened 2024-06-20
Fix color mixing and gradients in luma space

EpicEricEE:
The four components of luma colors are `[luma, luma, luma, alpha]`. In color mixing, it was mistakenly assumed that the second component was the alpha value, leading to some unexpected mixing behaviour:

```typ
#rect(fill: gradient.linear(black, silver, space: luma))
```
image


laurmaedje:
Thanks!

PR #2661, opened 2023-11-11
Fix luma → cmyk conversion

frozolotl:
Previously, the luma → cmyk conversion produced incorrect results.
This new conversion matches the values gained from luma → rgba → cmyk.

This PR also fixes a bug where `repr(luma(0%))` printed `rgb(0%, 0%, 0%, 0%)`.

PR #4425, opened 2024-06-20
Fix luma to CMYK conversion

EpicEricEE:
The luma value is basically the "amount of white" but the CMYK values are rather an "amount of darkness" so it should be inverted. Previously `cmyk(black)` would return white (and the other way around).

This was also included in #2661 but then forgotten.


laurmaedje:
Thanks!

PR #4429, opened 2024-06-22
Better error message when package version not exists

### LINKED ISSUE #4355: Better error message when trying to add a package version that doesn't exist
### Description

It would be great if a bit more detail could be shown on why a package was not downloadable.  Consider the following error message:

```shell
downloading @preview/finite:0.3.1
error: package not found (searched for @preview/finite:0.3.1)
  ┌─ test.typ:1:8
  │
1 │ #import "@preview/finite:0.3.1": automaton
  │         ^^^^^^^^^^^^^^^^^^^^^^^
```

This is a bit misleading - my first impression is that the package `@preview/finite` was not found, not that that version doesn't exist.  A better error might be:

```shell
downloading @preview/finite:0.3.1
error: package found, but version 0.3.1 does not exist.  Latest version is 0.3.0. 
(searched for @preview/finite:0.3.1)
  ┌─ test.typ:1:8
  │
1 │ #import "@preview/finite:0.3.1": automaton
  │         ^^^^^^^^^^^^^^^^^^^^^^^
```

### Use Case

Consider the following document:
```typst
#import "@preview/finite:0.3.t": automaton

#automaton((
  q0: (q1:0, q0:"0,1"),
  q1: (q0:(0,1), q2:"0"),
  q2: (),
))
```

If I try to compile this as of the date of writing, I get the following result:

```shell
downloading @preview/finite:0.3.1
error: package not found (searched for @preview/finite:0.3.1)
  ┌─ test.typ:1:8
  │
1 │ #import "@preview/finite:0.3.1": automaton
  │         ^^^^^^^^^^^^^^^^^^^^^^^
```

This doesn't tell me whether the package `@preview/finite` does not exist, or if it is an issue with the version (currently v0.3.1 is not released).  I then have no way to know how to go forward and resolve the issue, which was that I had to write `0.3.0` instead.

Coekjan:
Closes #4355

### DIFF
--- a/crates/typst-cli/src/package.rs
+++ b/crates/typst-cli/src/package.rs
@@ -5,6 +5,7 @@ use std::path::{Path, PathBuf};
 use crate::args::PackageStorageArgs;
 use codespan_reporting::term::{self, termcolor};
 use ecow::eco_format;
+use once_cell::sync::OnceCell;
 use termcolor::WriteColor;
 use typst::diag::{bail, PackageError, PackageResult, StrResult};
 use typst::syntax::package::{
@@ -21,6 +22,7 @@ const DEFAULT_PACKAGES_SUBDIR: &str = "typst/packages";
 pub struct PackageStorage {
     pub package_cache_path: Option<PathBuf>,
     pub package_path: Option<PathBuf>,
+    index: OnceCell<Vec<PackageInfo>>,
 }
 
 impl PackageStorage {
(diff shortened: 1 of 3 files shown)


laurmaedje (commented):
`determine_latest_version` will download the `@preview` package index over and over again. I think we should avoid this. Perhaps it could live in a `OnceCell` in `PackageStorage`.

Coekjan:
> `determine_latest_version` will download the `@preview` package index over and over again. I think we should avoid this. Perhaps it could live in a `OnceCell` in `PackageStorage`.

Agree. The package index is now cached.

laurmaedje:
Private is fine. Thank you, this is nice!

laurmaedje:
one thing is noticed is that the "downloading .." doesn't go away when there's an error. but that's also on main, so it's unrelated.

Bildschirmfoto 2024-06-22 um 12 33 47

PR #4430, opened 2024-06-22
Open with (detached) custom viewers and raise error on failure

### LINKED ISSUE #4386: The `open` CLI option can fail silently
### Description

I'm creating a downstream application that uses the CLI. However, the `open` option fails most of the time. It seems likely that the cause of failure is on my machine, but the failure is difficult to debug because no error or warning is displayed. If I'm understanding the code correctly, it seems like potential errors are ignored on this [line](https://github.com/typst/typst/blob/ad4ef68a112dedabf80f885a02bcb574eb9af9e4/crates/typst-cli/src/compile.rs#L482). Could it instead display a warning message?

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

Coekjan:
Closes #4386 .

I just tested the behavior on my linux KDE desktop and not sure if it works on other platform (but I think it would work :smile: )

---

#### `typst compile --open` and the default resource openers failed

```console
$ PATH= ./target/debug/typst compile ../test.typ --open
error: failed to open file with any of these resource openers: xdg-open, gio, gnome-open, kde-open (No such file or directory (os error 2))
```

#### `typst watch --open` and the default openers failed

```console
$ PATH= ./target/debug/typst watch ../test.typ --open
watching ../test.typ
writing to ../test.pdf

[14:01:55] compiled successfully in 17.73ms

error: failed to open file with any of these resource openers: xdg-open, gio, gnome-open, kde-open (No such file or directory (os error 2))
```

#### `typst compile --open /path/nonexistent`

```console
$ ./target/debug/typst compile ../test.typ --open /path/nonexistent
error: failed to open file with /path/nonexistent (No such file or directory (os error 2))
```

#### `typst watch --open /path/nonexistent`

```console
$ ./target/debug/typst watch ../test.typ --open /path/nonexistent
watching ../test.typ
writing to ../test.pdf

[14:03:36] compiled successfully in 17.72ms

error: failed to open file with /path/nonexistent (No such file or directory (os error 2))
```

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -100,9 +100,11 @@ pub struct CompileCommand {
     #[arg(long = "format", short = 'f')]
     pub format: Option<OutputFormat>,
 
-    /// Opens the output file using the default viewer after compilation.
-    /// Ignored if output is stdout
-    #[arg(long = "open")]
+    /// Opens the output file with the default viewer or a specific program after
+    /// compilation
+    ///
+    /// Ignored if output is stdout.
+    #[arg(long = "open", value_name = "VIEWER")]
     pub open: Option<Option<String>>,
 
     /// The PPI (pixels per inch) to use for PNG export
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -472,14 +472,30 @@ fn write_make_deps(world: &mut SystemWorld, command: &CompileCommand) -> StrResu
 /// Opens the given file using:
 /// - The default file viewer if `open` is `None`.
 /// - The given viewer provided by `open` if it is `Some`.
+///
+/// If the file could not be opened, an error is returned.
 fn open_file(open: Option<&str>, path: &Path) -> StrResult<()> {
+    // Some resource openers require the path to be canonicalized.
+    let path = path
+        .canonicalize()
+        .map_err(|err| eco_format!("failed to canonicalize path ({err})"))?;
     if let Some(app) = open {
-        open::with_in_background(path, app);
+        open::with_detached(&path, app)
+            .map_err(|err| eco_format!("failed to open file with {} ({})", app, err))
     } else {
-        open::that_in_background(path);
+        open::that_detached(&path).map_err(|err| {
+            let openers = open::commands(path)
+                .iter()
+                .map(|command| command.get_program().to_string_lossy())
+                .collect::<Vec<_>>()
+                .join(", ");
+            eco_format!(
+                "failed to open file with any of these resource openers: {} ({})",
+                openers,
+                err,
+            )
+        })
     }
-
-    Ok(())
 }
 
 /// Adds useful hints when the main source file couldn't be read


jakobjpeters:
Where Typst v0.11.1 succeeded to open the PDF 1/20 times on my machine, this PR succeeded 20/20 times running `typst compile ../test.typ --open`.

Coekjan:
> Where Typst v0.11.1 succeeded to open the PDF 1/20 times on my machine, this PR succeeded 20/20 times running `typst compile ../test.typ --open`.

Thanks for your testing. I updated the code just now. Could you please help test the newest implementation on your machine? Best regards.

jakobjpeters:
I'm pleased to report another 20/20 successes with the updated code!

tbethe:
As the comments in the code also reflect, the program specified as an argument in `--open` refers to a viewer, not an opener. That is the point of it.

1. What use cases require specifying a different opener than your standard one? I cannot think of any, but I am curious, perhaps there are some.
2. Having multiple pdf viewer is common, at least for me, and `--open` should allow you to specify which one you want to use.

The issue is that we have to take into account that the  command specified by `--open` can block AND we want to know if it failed. I'm not sure how to reconcile those two.

Coekjan:
I turned to use `open::{with,that}_detached` to open the output file, and now the compiler:
1. does not block on `--open <custome-viewer>` CLI option.
2. emits errors when it fails to open.

laurmaedje:
> As the comments in the code also reflect, the program specified as an argument in --open refers to a viewer, not an opener.

I'm not familiar with the exact distinction, but if this is indeed the case, perhaps we should adjust the error messages to say "viewer" instead of "opener"? Or we just drop it completely and say `failed to open file with any of ...`

Coekjan:
Current PR implements that:
1. `typst compile A.typ --open`: opens the file with resource openers in the system (see https://docs.rs/crate/open/5.2.0/source/src/unix.rs for examples).
    ```rs
    pub fn commands<T: AsRef<OsStr>>(path: T) -> Vec<Command> {
        let path = path.as_ref();
        let mut commands: Vec<(&str, Vec<&OsStr>)> = vec![];

        let wsl_path = wsl_path(path);
        if is_wsl::is_wsl() {
            commands.push(("wslview", vec![&wsl_path]));
        }

        commands.extend_from_slice(&[
            ("xdg-open", vec![&path]),
            ("gio", vec![OsStr::new("open"), path]),
            ("gnome-open", vec![path]),
            ("kde-open", vec![path]),
        ]);

        commands
            .iter()
            .map(|(command, args)| {
                let mut cmd = Command::new(command);
                cmd.args(args);
                cmd
            })
            .collect()
    }
    ```
2. `typst compile A.typ --open <specific-viewer>`: opens the file with specific viewer.

Both of them will give error messages on failure. And with the help of `open::{with,that}_detached`, `typst watch A.typ --open <viewer>` will not block the re-compilation.

Coekjan:
When `typst compile A.typ --open` fails, its error message will be "failed to open file with any of these resource openers: xdg-open, gio, gnome-open, kde-open (( THE OS ERROR HERE ))"

laurmaedje:
I see, that makes sense. Thank you for your attention to detail on this PR!

PR #4435, opened 2024-06-23
Refactor `impl Eval for ast::FuncCall<'_>`

Coekjan:
See https://github.com/typst/typst/pull/4267#discussion_r1623711869 .

This PR aims to extract some code from `impl Eval for ast::FuncCall<'_>`, which makes it easier for developers to understand the implementation.

Co-authored-by: @wrzian

### DIFF
--- a/crates/typst/src/eval/call.rs
+++ b/crates/typst/src/eval/call.rs
@@ -1,7 +1,9 @@
 use comemo::{Tracked, TrackedMut};
 use ecow::{eco_format, EcoVec};
 
-use crate::diag::{bail, error, At, HintedStrResult, SourceResult, Trace, Tracepoint};
+use crate::diag::{
+    bail, error, At, HintedStrResult, HintedString, SourceResult, Trace, Tracepoint,
+};
 use crate::engine::{Engine, Sink, Traced};
 use crate::eval::{Access, Eval, FlowEvent, Route, Vm};
 use crate::foundations::{
@@ -10,7 +12,7 @@ use crate::foundations::{
 };
 use crate::introspection::Introspector;
 use crate::math::LrElem;
-use crate::syntax::ast::{self, AstNode};
+use crate::syntax::ast::{self, AstNode, Ident};
 use crate::syntax::{Span, Spanned, SyntaxNode};
 use crate::text::TextElem;
 use crate::utils::LazyHash;
@@ -32,33 +34,28 @@ impl Eval for ast::FuncCall<'_> {
         }
 
         // Try to evaluate as a call to an associated function or field.
-        let (callee, mut args) = if let ast::Expr::FieldAccess(access) = callee {
-            let target = access.target();
-            let target_span = target.span();
+        let (callee, args) = if let ast::Expr::FieldAccess(access) = callee {
+            let target_expr = access.target();
             let field = access.field();
-            let field_span = field.span();
-
-            let target = if is_mutating_method(&field) {
-                let mut args = args.eval(vm)?.spanned(span);
-                let target = target.access(vm)?;
-
-                // Only arrays and dictionaries have mutable methods.
-                if matches!(target, Value::Array(_) | Value::Dict(_)) {
-                    args.span = span;
-                    let point = || Tracepoint::Call(Some(field.get().clone()));
-                    return call_method_mut(target, &field, args, span).trace(
-                        vm.world(),
-                        point,
-                        span,
-                    );
-                }
-
-                target.clone()
-            } else {
-                access.target().eval(vm)?
-            };
-
             let mut args = args.eval(vm)?.spanned(span);
+
+            // Evaluate the target expression.
+            let target = if is_mutating_method(&field) {
+                match target_expr.access(vm)? {
+                    // Only arrays and dictionaries have mutable methods.
+                    target @ (Value::Array(_) | Value::Dict(_)) => {
+                        let point = || Tracepoint::Call(Some(field.get().clone()));
+                        return call_method_mut(target, &field, args, span).trace(
+                            vm.world(),
+                            point,
+                            span,
+                        );
+                    }
+                    target => target.clone(),
+                }
+            } else {
+                target_expr.eval(vm)?
+            };
 
             // Handle plugins.
             if let Value::Plugin(plugin) = &target {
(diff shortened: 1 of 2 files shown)


crates/typst/src/eval/call.rs
```diff
@@ -32,135 +34,50 @@ impl Eval for ast::FuncCall<'_> {
...
+                        return call_method_mut(
+                            target_expr.access(vm)?,
+                            &field,
+                            args,
+                            span,
+                        )
+                        .trace(vm.world(), point, span);
+                    }
+                    target => target.clone(),
                 }
```
PgBiel, line 57: We should perhaps not access twice, if possible:

```suggestion
                match target_expr.access(vm)? {
                    // Only arrays and dictionaries have mutable methods.
                    target@(Value::Array(_) | Value::Dict(_)) => {
                        let point = || Tracepoint::Call(Some(field.get().clone()));
                        return call_method_mut(
                            target,
                            &field,
                            args,
                            span,
                        )
                        .trace(vm.world(), point, span);
                    }
                    target => target.clone(),
                }
```

crates/typst/src/foundations/args.rs
```diff
@@ -76,8 +76,20 @@ impl Args {
...
+            Arg {
+                span: self.span,
+                name: None,
+                value: Spanned::new(value, span),
+            },
+        )
+    }
+
+    /// Push a positional argument to the back.
+    pub fn push_back(&mut self, span: Span, value: Value) {
```
PgBiel, line 92: These names are slightly confusing due to the similarity with `VecDeque`. I'd suggest the following instead (while replacing earlier usage with `args.insert(0, span, value)`):

```suggestion
    /// Insert a positional argument at a specific index.
    pub fn insert(&mut self, index: usize, span: Span, value: Value) {
        self.items.insert(
            index,
            Arg {
                span: self.span,
                name: None,
                value: Spanned::new(value, span),
            },
        )
    }

    /// Push a positional argument.
    pub fn push(&mut self, span: Span, value: Value) {
```

crates/typst/src/eval/call.rs
```diff
@@ -368,6 +285,60 @@ pub(crate) fn call_closure(
...
+/// A function call on a field is only allowed for functions, types, modules (because
+/// they are scopes), and symbols (because they have modifiers).
+///
+/// For dictionaries, it is not allowed because it would be ambiguous (prioritizing
+/// associated functions would make an addition of a new associated function a breaking
+/// change and prioritizing fields would break associated functions for certain
+/// dictionaries).
+fn eval_callee_args(
+    target: Value,
+    target_expr_span: Span,
```
PgBiel, line 300:
```suggestion
    target_span: Span,
```

crates/typst/src/eval/call.rs
```diff
@@ -368,6 +285,60 @@ pub(crate) fn call_closure(
...
+                    field.as_str(),
+                ));
+            }
+            _ if target.field(&field).is_ok() => {
+                error.hint(eco_format!(
+                    "did you mean to access the field `{}`?",
+                    field.as_str()
+                ));
+            }
+            _ => (),
```
PgBiel, line 334: very nitpicky nitpick but i think this is more usual

```suggestion
            _ => {}
```

PgBiel (commented):
Very nice, thanks! Here are my thoughts:

Coekjan:
> Very nice, thanks! Here are my thoughts:

Thanks for your kind review. I applied your suggestions!

crates/typst/src/eval/call.rs
```diff
@@ -32,135 +34,48 @@ impl Eval for ast::FuncCall<'_> {
...
+            let mut args = args.eval(vm)?.spanned(span);
 
+            // Evaluate the target expression.
             let target = if is_mutating_method(&field) {
-                let mut args = args.eval(vm)?.spanned(span);
-                let target = target.access(vm)?;
-
-                // Only arrays and dictionaries have mutable methods.
-                if matches!(target, Value::Array(_) | Value::Dict(_)) {
-                    args.span = span;
```
PgBiel, line 47: Paging @laurmaedje : I think this `args.span = span;` was at odds with the original `.spanned(span)`, right? It does seem that `.spanned(span)` only changes the span if there wasn't already one, while `.span = span` will always change it. I think `.spanned(span)` is enough here, since `.eval` creates an `Args` with a detached span, so this should be ok to be removed, correct? (Just wanted to make sure we're doing the right thing here)

PgBiel (approved):
Looks great, thank you!

crates/typst/src/eval/call.rs
Coekjan: When investigating this code, I did think the extra `.span = span` is no need here, because `.spanned()` already set it.

If I am wrong, please correct me. :smile:

wrzian:
I made a new child PR with some refactors of the refactor, see here: https://github.com/Coekjan/typst/pull/1

I took a stab at refactoring the main `eval()` function a bit further. I put the early return cases into a new type `EvaluatedMethod` which let me simplify the big `(callee, args)` if-else statement. I also factored out the missing method case into its own function.

However, in doing so I discovered an important change that the current refactor made: **The current refactor swaps the order of evaluation of the callee and the arguments when the callee is a method call.**

I made a long comment explaining the status quo in the file, but the difference would arise from code like this:
```typ
// Typst v0.11.1
#{
  let a = (1, 2, 3)
  let b = a.enumerate(start: a.pop())
  // => a: (1, 2)
  // => b: ((3, 1), (4, 2), (5, 3))
}
#{
  let a = (1, 2, 3)
  let start = a.pop()
  let b = a.enumerate(start: start)
  // => a: (1, 2)
  // => b: ((3, 1), (4, 2))
}
```

Here the difference between the two blocks is whether `a.pop()` is called within the arguments or outside them. The current behavior evaluates the *callee first* and the *arguments second*, so the value of `a` is cloned for the enumeration, and `a.pop()` does not affect it, whereas the second block has the pop happen first so the enumeration is affected.

The refactor as-is would change this behavior by evaluating args first in the first block, making the two blocks the same. However I believe this is *not* desired, and my commit returns to the current behavior.

I haven't added this as a test yet, but we probably should.

laurmaedje:
How should I proceed here? Shall I review this as-is or will the refactor of the refactor be upstreamed to the original refactor?

Coekjan:
> How should I proceed here? Shall I review this as-is or will the refactor of the refactor be upstreamed to the original refactor?

I believe we (@wrzian and I) will merge https://github.com/Coekjan/typst/pull/1 soon and then we could continue discussing the refactor here.

Coekjan:
I think if we expect `callee` to be evaluated before `args`, shall we add some regression test about it?

crates/typst/src/eval/call.rs
```diff
@@ -368,6 +261,111 @@ pub(crate) fn call_closure(
...
+        Ok(MethodCall::Special(value))
+    } else if let Some(callee) = target.ty().scope().get(&method) {
+        args.insert(0, target_expr.span(), target);
+        Ok(MethodCall::Normal(callee.clone(), args))
+    } else if matches!(
+        target,
+        Value::Symbol(_) | Value::Func(_) | Value::Type(_) | Value::Module(_)
+    ) {
+        // Certain value types may have their own ways to access method fields.
+        // e.g. `$arrow.r(v)$`, `func.with(fill: red)`
```
laurmaedje, line 331: `func.with` is a wrong example here, I think. This branch covers things like `table.cell[..]`. `func.with` is handled by `target.ty().scope().get(&method)`.

crates/typst/src/eval/call.rs
```diff
@@ -368,6 +261,111 @@ pub(crate) fn call_closure(
     Ok(output)
 }
 
+/// Method calls have some special cases which we convert to values eagerly. Otherwise
+/// this contains the evaluated method to call and the arguments to call with.
+enum MethodCall {
```
laurmaedje, line 266: It's not really always a `MethodCall`. It could be a function call on a field (e.g. `table.cell`) and then the special case in particular can also be other stuff.

How about `FieldCall` and renaming `Special` to `Resolved` since the call is already resolved?

crates/typst/src/eval/call.rs
```diff
@@ -368,6 +261,111 @@ pub(crate) fn call_closure(
...
+        "type {} has no method `{}`",
+        target.ty(),
+        method.as_str(),
+    );
+
+    match target {
+        Value::Dict(ref dict) if matches!(dict.get(&method), Ok(Value::Func(_))) => {
+            error.hint(eco_format!(
+                "to call the function stored in the dictionary, surround \
+                    the field access with parentheses, e.g. `(dict.{})(..)`",
```
laurmaedje, line 353: could you align "to" and "the" as they were before?

crates/typst/src/eval/call.rs
```diff
@@ -376,6 +374,47 @@ fn in_math(expr: ast::Expr) -> bool {
...
+    vm: &mut Vm,
+    callee: &ast::Expr,
+    mut err: HintedString,
+) -> HintedString {
+    if let ast::Expr::Ident(ident) = callee {
+        let ident = ident.get();
+        if vm.scopes.check_std_shadowed(ident) {
+            err.hint(eco_format!(
+                "use `std.{}` to access the shadowed standard library function",
+                ident,
```
laurmaedje, line 411: I think `ident` could be directly in the string here. for `format!` clippy even complains afaik, but not for `eco_format`.

laurmaedje:
Thanks to both of you!

PR #4444, opened 2024-06-24
Reword docs for `math.cancel`

### LINKED ISSUE #3785: Misleading documentation for the angle parameter of cancel
https://typst.app/docs/reference/math/cancel/

The current documentation says the following:
```
How much to rotate the cancel line.

    If auto, the line assumes the default angle; that is, along the diagonal line of the content box.
    If given an angle, the line is rotated by that angle clockwise w.r.t the y-axis.
    If given a function angle => angle, the line is rotated by the angle returned by that function. The function receives the default angle as its input.

Default: auto
```

It is not clear that second option results in an absolute rotation from a vertical line. It's also not specified _which_ diagonal is the default.

I would go with something like this:
```
Controls the angle between the line and the y-axis, measured clockwise.

The default angle is that corresponding to the rising diagonal of the content's bounding box.

If set to auto, the angle is the default.
If given a function, the function receives the default angle and must return a new angle.

Default: auto
```

AnarchistHoneybun:
resolves #3785 

given that the ambiguity in the docs arises from 

1. the arguments to cancel mentioning a line rotation as default, then saying it's rotated by some degree, giving the impression it's rotated by default + some degree, not directly from y-axis
2. the docs not specifying which diagonal the default angle aligns the line to

I have made these changes:

1. move the line about the angle argument above the auto/default argument. this clears up most of the ambiguity without having to make any real changes to these lines, so we don't have to worry about further misinterpretations
2. change the word diagonal to anti-diagonal ([reference](https://en.wikipedia.org/wiki/Main_diagonal)) to more clearly denote what diagonal we mean

(first time trying to contribute to this project! this was the first entry on good first issues so I took it. not a lot of work but I'm trying to get familiar with the code more)


crates/typst/src/math/cancel.rs
```diff
@@ -65,13 +65,13 @@ pub struct CancelElem {
 
     /// How much to rotate the cancel line.
     ///
-    /// - If `{auto}`, the line assumes the default angle; that is, along the
-    ///   diagonal line of the content box.
     /// - If given an angle, the line is rotated by that angle clockwise w.r.t
```
bluebear94, line 68: I think `w.r.t` should also be expanded to `with respect to`.

AnarchistHoneybun: I used it since
- that's what was used originally
- doesn't seem that obscure of an initialism
- changing it here would essentially make the way of writing different from all the other places this initialism is used, and, like, we might want to open a separate issue for something like that(?)

,what do you think?

(ESL, so apologies if any sentence structure seems awkward)

Enivex: I would definitely write it out in full.

Enivex:
I do not like the use of anti-diagonal. That makes it no more clear, as we're not talking about matrices here (and even for matrices that's arguably not common terminology).

crates/typst/src/math/cancel.rs
AnarchistHoneybun: Should I open a separate issue for that or just lump the changes with this one

Enivex: Do you have examples of other instances of "w.r.t." or other abbreviations in the documentation?

AnarchistHoneybun:
> I do not like the use of anti-diagonal. That makes it no more clear, as we're not talking about matrices here (and even for matrices that's arguably not common terminology).
> 
> I also do not think just changing the order is sufficient to clear up all the ambiguity.

- the other option would be to add a bottom of the page annotation which says for a diagonal we mean the one running from top right of the element to the bottom left. The broader question here is, do we expect the average user to understand/look up slightly obscure terms like these,  or we explain ourselves on each step, potentially littering the docs with annotations.

- I quite liked the current formatting where the three argument types are addressed separately, and tried to maintain that for my changes. Any other sentence I try to use to explain it starts sounding very awkward very soon. But I'll try to rephrase

crates/typst/src/math/cancel.rs
AnarchistHoneybun: Apart from the file changed for this pr, these are other two using w.r.t. away from my pc currently but I will try to run a search for other abbreviations if needed, when I get back

Coekjan: I prefer to write "with respect to" fully.

I believe if the document is written for users, especially the API reference, it should be readable and easy to understand for everyone (including non-native English users & inexperienced developers).

BTW, here is the documentation for users, whereas the "w.r.t" in your posted image is for developers, but not for users.

AnarchistHoneybun: I see. I'll change it to the full words

bluebear94:
I too prefer the term “rising diagonal” over “anti-diagonal”, though the former could still be unclear.

Both of these terms are dependent on the writing direction, too; I don’t know whether we want to mirror the stroke for e.g. Arabic math notation, but I guess this is a rare case to begin with.

AnarchistHoneybun:
> I too prefer the term “rising diagonal” over “anti-diagonal”, though the former could still be unclear.
> 
> Both of these terms are dependent on the writing direction, too; I don’t know whether we want to mirror the stroke for e.g. Arabic math notation, but I guess this is a rare case to begin with.

this brings us back to a previous question, do we trust people to look up definitions of stuff like this, if needed, or we litter the docs with explanations and then explanations for those. We could also cop out a bit and put a reference to the wiki article/math stack exchange for any of these terms if we use them in the docs so people have a direct way of checking what the words exactly mean

laurmaedje:
I think "rising diagonal" doesn't sound bad. It's clearer to the average person than anti-diagonal and we don't support Arabic math notation anyway.

AnarchistHoneybun:
that seems to be the general consensus then I'll change it o7

laurmaedje:
I don't really have an opinion on this, but @Enivex still seemed to have concerns with just changing the ordering.

laurmaedje:
I think it's a bit better than before and relatively clear, so I'll just merge. Thanks!

PR #4446, opened 2024-06-25
fix(array.slice): Clarify and correct function description

flauschpantoffel:
Updated the description of the `slice` function to explicitly mention both start and end index. This ensures the documentation accurately describes the behavior of the function.

### DIFF
--- a/crates/typst/src/foundations/array.rs
+++ b/crates/typst/src/foundations/array.rs
@@ -259,7 +259,7 @@ impl Array {
             .ok_or_else(|| out_of_bounds_no_default(index, self.len()))
     }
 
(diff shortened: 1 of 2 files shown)


crates/typst/src/foundations/array.rs
```diff
@@ -259,7 +259,7 @@ impl Array {
             .ok_or_else(|| out_of_bounds_no_default(index, self.len()))
     }
 
-    /// Extracts a subslice of the array. Fails with an error if the start or
+    /// Extracts a subslice of the array. Fails with an error if the start or end
```
Coekjan, line 262: Would it be better if we say "slice" instead of "subslice"?

laurmaedje:
Thanks.

PR #4456, opened 2024-06-26
Fix footnote-reference numbering

### LINKED ISSUE #4454: Footnote references do not respect numbering format
### Description

using typst 0.11.1 (50115102)

@MDLC01 (just fyi)

When generating a footnote and then referencing it to have multiple references pointing to the same footnote the original formatting will be ignored and replaced with default numbering e.g., 1.

```typst
#set page(width:auto,height:auto)
I love mangos #footnote(numbering: "*")[These are yummy! ]<fn1>, apples @fn1 , bananas@fn1 and rubix Cubes.
```

image

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

Coekjan:
Fixes #4454.


crates/typst/src/model/reference.rs
```diff
@@ -177,8 +177,14 @@ impl Show for Packed<RefElem> {
 
         let elem = elem.at(span)?;
 
-        if elem.func() == FootnoteElem::elem() {
-            return Ok(FootnoteElem::with_label(target).pack().spanned(span));
+        if elem.is::<FootnoteElem>() {
```
bluebear94, line 180: Why not use `Content::get_ref` here to avoid the unwrap later on?

Coekjan: I cannot find the `Content::get_ref` function. Could you please tell me where it is?

Coekjan: I see `Packed::<T>::from_ref`. Guess you mean we can use this?

laurmaedje: The canonical version is `elem.to_packed`.

bluebear94 (commented):
Looks good to me, but I’m concerned that there might be other fields that need to be copied from the source footnote in the future.

Coekjan:
> Looks good to me, but I’m concerned that there might be other fields that need to be copied from the source footnote in the future.

I defined `FootnoteElem::into_ref(&self, Label)`, so it would be better now.

laurmaedje:
Looks good, thank you!

Issue #4452, opened 2024-06-25
Crate for downstream tooling


tingerrr:
### Description

Lift parts of `typst-cli` and the typst/package `bundler` into a dedicated crate for implementing tooling related to Typst.

A non-exhaustive list of features I believe should be added to such a crate:
- manifest parsing
  - parsing manifest files into a well typed struct
- manifest validation
  - ensuring the validity of a manifest for the universe
- manifest writing/editing
  - editing a manifest in place like cargo does (see `toml-edit`)
  - I think this is not required now, but essential later when git packages and custom registries arrive
- tool section reading and writing
  - in a similar vein as above
- package downloading
  - making the implementation of a custom registry easier
  - exposing the default locations and fallbacks
- font loading
  - loading fonts with the correct priorities
- package download locking
  - being able to lock downloading of a package consistently to avoid interference with other tools running in parallel, provided they also use the same locking mechanism
- project locking
  - as above but for doing automation inside the project

If these features are sufficiently abstracted, most of them could be optional feature flags, reducing possible code bloat.

My reasoning for keeping manifest parsing and validation separate is that some things are fixed, like what the Typst Universe allows as valid manifests, but there are valid use cases where these need not be valid for the Universe: company internal registries may have different requirements for licenses, keywords, etc. Furthermore, tools should be able to ignore some errors where they don't matter while still being able to read other parts of the manifest, like their own tool section, or to fix the manifest like a linter would.

### Prior Art
Some prior art would be [`typst-project`](https://github.com/tingerrr/typst-project), a crate I made for `typst-test` and by some extent `utpm`. It provides manifest parsing, reading the tool section for custom configuration, and a quite frankly overengineered API for locating the project root.

Another example of where this was done is jj, which actually exposes the whole CLI as a library. They provide a `jj_lib` library crate for the core building blocks, a `jj_cli` lib crate for users to implement [custom backends](https://github.com/martinvonz/jj/blob/main/cli/examples/custom-backend/main.rs) and commands for company internal usage while maintiang the rest of the default CLI and the `jj_cli` binary crate as the default jj implementation. While this is certainly overengineered from the perspective of Typst, it can serve as a good example of how to find the right API abstractions.

### Use Case

Various downstream tooling projects for Typst written in Rust exist, many of which needed to reimplement some part of the functionality contained in `typst-cli`.

Some projects which had to reimplement features include:
|tool|reimplemented feature|
|---|---|
[...]

PR #4457, opened 2024-06-26
Make typst-cli crate usable as library

zinifier:
Like others in #4452 i would like to use typst as library. This is the simplest way i found to do it, turning typst_cli into a library.

### DIFF
--- a/crates/typst-cli/src/lib.rs
+++ b/crates/typst-cli/src/lib.rs
@@ -1,0 +1,69 @@
+pub mod args;
+pub mod compile;
+pub mod download;
+pub mod fonts;
+pub mod init;
+pub mod package;
+pub mod query;
+pub mod terminal;
+pub mod timings;
+#[cfg(feature = "self-update")]
+pub mod update;
+pub mod watch;
+pub mod world;
+
+use std::cell::Cell;
+use std::io::{self, Write};
+use std::process::ExitCode;
+
+use clap::Parser;
+use codespan_reporting::term;
+use codespan_reporting::term::termcolor::WriteColor;
+use once_cell::sync::Lazy;
+
+use crate::args::CliArguments;
+
+/// Ensure a failure exit code.
+pub fn set_failed() {
+    EXIT.with(|cell| cell.set(ExitCode::FAILURE));
+}
+
+
+thread_local! {
+    /// The CLI's exit code.
+    pub static EXIT: Cell<ExitCode> = const { Cell::new(ExitCode::SUCCESS) };
+}
+
+/// The parsed commandline arguments.
+pub static ARGS: Lazy<CliArguments> = Lazy::new(CliArguments::parse);
+
+/// Used by `args.rs`.
+pub fn typst_version() -> &'static str {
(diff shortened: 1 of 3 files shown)


crates/typst-cli/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+
+thread_local! {
+    /// The CLI's exit code.
+    pub static EXIT: Cell<ExitCode> = const { Cell::new(ExitCode::SUCCESS) };
+}
+
+/// The parsed commandline arguments.
+pub static ARGS: Lazy<CliArguments> = Lazy::new(CliArguments::parse);
```
tingerrr, line 38: This is implicitly accessed in various parts of the library and will exit if the cli arguments don't match.

crates/typst-cli/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+pub mod fonts;
+pub mod init;
+pub mod package;
+pub mod query;
+pub mod terminal;
+pub mod timings;
+#[cfg(feature = "self-update")]
+pub mod update;
+pub mod watch;
+pub mod world;
```
tingerrr, line 13: Exposing all of these comes with a big semver hazard, if any of these APIs which are specific to typst-cli alone like `update` are changed, a new major version of the library would be required.

In theory, no one in their right mind would use the `update` module, but this is not enough of an argument to ignore the semver hazard I'm afraid.

tingerrr (commented):
I'm afraid turning typst-cli into a library crate just like this will not be enough. The implicit dependency on typst-cli specific arguments (`ARGS`) must be removed from all publically exported functions.

crates/typst-cli/src/lib.rs
zinifier: I think it's only used in terminal.rs for color management, and download.rs for cert management, and main.rs of course. It works fine when calling compile_once from outside typst.

zinifier:
> I'm afraid turning typst-cli into a library crate just like this will not be enough.

Yes sorry i should have made it more clear: i'm not saying this is the good way to do it. I'm just showing that's what i did that works for me. Feel free to close the pull request if you have better plans.

crates/typst-cli/src/lib.rs
tingerrr: But that's the point, a consumer might try to use download and then get a failure about their cli arguments being incorrect when they might not be making a cli program at all.

tingerrr:
> Yes sorry i should have made it more clear: i'm not saying this is the good way to do it. I'm just showing that's what i did that works for me. Feel free to close the pull request if you have better plans.

I see, I can't manage pull requests, but I think it would be good to turn this into a draft (see the upper right next to reviewers) to clearly indicate that this is more of an example than a finished proposed change to the repo.

laurmaedje:
I'm on board with a `typst-kit` crate with a proper API to help with downstream tooling (as discussed in https://github.com/typst/typst/issues/4452, @tingerrr is already working on it). I think that's the cleaner solution, so I'll close this. Thanks still!

Issue #4217, opened 2024-05-22
Assigning Diagnostics with Identifiers


Myriad-Dreamin:
### Description

The diagnostics from Typst should be able to be identified or referred via a precise way, e.g. by assigning identifiers, [MSVC Error LNK2019](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk2019?view=msvc-170).

### Use Case

Here are four possible use cases:
- Make it possible to attach ["Fix Me" code actions](https://code.visualstudio.com/docs/editor/refactoring#_code-actions-quick-fixes-and-refactorings) to the diagnostics.
- Operating the diagnostics programmatically, for example, changing the [hint to solve out of root errors](https://github.com/Myriad-Dreamin/tinymist/issues/117), whose hint is only useful for `typst-cli`.
- Further detailed documentation on specific diagnostics, like [MSVC explaining the solutions to LNK2019 errors](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk2019?view=msvc-170).
- Filtering the unwanted errors by ID.

Enter-tainer:
Similar thing in rust: https://doc.rust-lang.org/error_codes/error-index.html

PgBiel:
Hi, this is something we intend to work on, in particular to make features such as warning suppression easier to implement. I have been considering something like Rust's slugs (see https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html), which would make a reference to a warning immediately readable (for example, `unnecessary_import_rename`), but we could also use numeric codes for compiler warnings (e.g. `W0123`). This can also be expanded to errors. What do you think?

Myriad-Dreamin:
IMHO, it is not necessary to enforce diagnostics to be immediately readable, as the IDs usually come with the diagnostics messages, which explain the meaning of IDs. This is also not quite useful as the English string is not quite readable for non-English users, which means we will have to show them by localization in the end.

But string IDs do have an advantage that I like, that we can have a namespace on diagnostics, e.g. `@preview/some_package:some_warning_id`, so we can maintain and operate warnings emitted by packages easily.

PgBiel:
[...]

PR #4459, opened 2024-06-26
Warning suppression

PgBiel:
**UPDATE (2024-08-06):** After lots of discussion (internally, over Discord etc.) regarding not only the ideal syntax for annotations, but also the potential use cases for warn suppression, we've decided to prioritize other, more important improvements for diagnostics, so this PR won't move forward for the time being.

---

This PR implements warning suppression within Typst source code, and all the underlying features that it requires:
1. Warning identifiers (partially solves #4217 - does not add identifiers for errors yet)
2. ~~Decorator~~ Annotation syntax (used to annotate where warn suppression occurs, indicating which warnings are suppressed)

Regarding internals, also implements some important lexing changes (might conflict a bit with #4364 , but hopefully not very much).

The idea is to make it possible to add more warnings without hurting UX too much, by allowing users to suppress them.

**Draft:** There are some rough edges in some places, so the PR shouldn't be merged just yet. (Also, many more tests are needed.)

## Overview

This is what it looks like:

```typ
// @allow unnecessary-stars
#[**]
```

**(Note: Used to be `/! allow("unnecessary-stars")` in the initial version of the PR)**

We can see all three main elements of this PR here:
1. ~~Decorator~~ **Annotation syntax:** `// @name("argument", "argument2")` annotates something relevant to the compiler at the source code.
     - This is bikesheddable still, even the name ("directive" could perhaps be a more appropriate name). **(Edit: The name has been changed to "Annotation")**
     - The idea is that ~~decorators~~ annotations should be nearly as flexible as comments so that they can be inserted in various locations within the source code. Hence why the syntax resembles that of comments as well.
2. **Warning identifiers:** The warning regarding `**` having no effect now has an identifier, `"unnecessary-stars"`. This is bikesheddable and I chose some (in principle, reasonable) initial names for all existing warnings.
3.  **Warning suppression:** The `allow` decorator, which takes warning identifiers, causes the given warning to be suppressed if it is raised under the decorator. That is, it won't be raised by the compiler at that point.

Of note, warning suppression applies to warnings raised not only directly under the decorator (or nested within what's under it), but also warnings raised in other files (or elsewhere in the same file) which have tracepoints that lead back to the decorator. For now, only calling functions (during eval) adds tracepoints to warnings, meaning that warnings raised during layout won't have tracepoints, or the tracepoints will be incomplete (e.g. warnings in a `context { }`).

This means that the `unnecessary-stars` warning will be suppressed in all of the following cases:

```typ
// @allow unnecessary-stars
#[**]

// @allow unnecessary-stars
#{
   [**]
   [**]
}

#let func() = {
    [**]
}

// Warning will have a tracepoint to this call
// @allow unnecessary-stars
#func()
```

But not in the cases below (must be up to one line above):

```typ
// @allow unnecessary-stars

#[**]

// @allow unnecessary-stars

#{
   [**]
   [**]
}
```

## Implementation highlights

1. Identifiers were fairly simple to implement, and are contained entirely within `diag.rs`.
2. For decorator syntax, the following main changes were done:
     1. The lexer was updated such that `next()` returns a `SyntaxNode`. It's almost always a leaf node, except when there is an error (then an error is returned - this means `take_error` and such are now internal) or when lexing a decorator (which returns a node with a subtree).
           - I tried to make very minimal changes here for this PR. This means that most functions still return `SyntaxKind`, and as such I didn't change raw elements to use this new lexer API. I think this is better suited to a separate PR to better gauge if there could be any possible regressions when implementing that.
     2. The former change was because all decorator parsing is done entirely within the lexer, as it's simpler to work with characters directly in that case.
     3. The parser had to be updated accordingly.
3. For warning suppression itself:
     1. Added `Engine::tracepoint` which is used to run some code and then immediately attach a tracepoint to the returned errors (through a returned `SourceDiagnostic`) and warnings (through a temporary `Sink`, which is created in the function and then merged into the outer engine's `Sink`).
     2. This is used in `impl Eval for FuncCall` to add a tracepoint to all warnings raised during a function's evaluation. This allows tracking which function calls led to a warning being raised.
     3. In the `Sink`, we now store duplicate warnings (relaxed warning deduplication), but only as long as they have differing tracepoints. This is because we need to keep track of all sets of tracepoints for a warning, since an `allow` decorator above a particular tracepoint might suppress that warning at that point, but the warning might still be raised because it wasn't suppressed at other tracepoint chains.
     5. Created `LinkedNode::prev_attached_decorator` which searches for the nearest sibling decorator _before_ the given syntax node, but only up to one line before it. That is, decorators two lines above a given node don't apply to it. (To properly implement this, syntax nodes now keep track of if they have 0, 1 or 2 newlines within them.)
     6. Created `Sink::suppress_and_deduplicate_warnings` which performs suppression by searching the syntax tree near where each warning was raised - as well as near each tracepoint - by checking if there is any decorator sibling before the node at the warning's span that suppresses it (see the previous item). If not, checks the node's parent, and keeps checking parents and their attached decorators until either a decorator suppressing this warning is found, or we reached the root node of the source file.
           - This function also deduplicates warnings which differ only by their tracepoints, which we can do since we now know which sets of tracepoints were suppressed or not (see item 3).
           - There are certainly performance improvements to be had here, e.g. maybe by caching/memoizing the search somehow. In principle, however, this is an initial implementation which seems to work.
      7. That sink function is used at `compile` to take and return the warnings, so warning suppression always occurs after a compilation.

## TODO

A few TODOs before undrafting:

- [ ] Adjust syntax
- [x] Why are Windows tests broken?!? (Fixed: we weren't handling `\r\n` correctly. Thanks @LaurenzV!)
- [ ] More tests, including testing `context { }` and tracepoints
- [x] ~~Improve the `Trace` trait stuff in `diag.rs` (I feel like there might be a less convoluted way to do this)~~ Keeping as is for now
- [x] Optimize newline counting in nodes
- [x] Improve the tree search code
- [x] Update `Decorator::arguments` considering `DecoratorName`
- [ ] Other possible minor improvements


JamesxX:
Would a directive only apply to the line that immediately follows, the next line of code that isn't trivial (tab, cr, etc)?

If it applies to more than one line, is it bounded by scope? Is it possibly to bound it with curly braces?

PgBiel:
I've updated the description with a few more examples!

It applies to the syntax node in the line immediately below the decorator in the source code. This includes a block that is directly under it.

crates/typst-syntax/src/ast.rs
```diff
@@ -80,6 +80,40 @@ impl<'a> Markup<'a> {
...
+
+    /// Get the decorator name as a string slice.
+    pub fn as_str(self) -> &'a str {
+        self.get()
+    }
+}
+
+node! {
+    /// A decorator: `/! allow("warning")`.
+    Decorator
```
laurmaedje, line 101: As discussed, I prefer the naming "annotation". We also discussed the syntax `// @allow warning`, but haven't 100% settled yet.

crates/typst-syntax/src/lexer.rs
```diff
@@ -132,6 +138,31 @@ impl Lexer<'_> {
             },
 
             None => SyntaxKind::End,
+        };
+
+        self.emit_token(token, start)
+    }
+
+    /// Constructs an error node with the given message.
```
laurmaedje, line 146: Please wrap comments at 80 columns.

crates/typst-syntax/src/lexer.rs
```diff
@@ -182,6 +213,151 @@ impl Lexer<'_> {
     }
 }
 
+/// Decorator lexing and auxiliary methods.
+impl Lexer<'_> {
+    /// Lexes and parses a decorator into a complete syntax subtree.
+    /// The lexer is fully responsible by the decorator, as it is simpler to
```
laurmaedje, line 219:
```suggestion
    /// The lexer is fully responsible for the decorator, as it is simpler to
```

crates/typst-syntax/src/lexer.rs
```diff
@@ -182,6 +213,151 @@ impl Lexer<'_> {
...
+    /// parse them at the character level, given they follow a very simple
+    /// and rigid structure, in the form
+    /// `/! decorator-name("string argument1", "string argument2")`
+    /// with optional whitespaces and comments between arguments.
+    fn decorator(&mut self, start: usize) -> SyntaxNode {
```
laurmaedje, line 224: I'll skip reviewing this for now as the syntax is subject the change.

crates/typst-syntax/src/node.rs
```diff
@@ -381,6 +398,11 @@ struct InnerNode {
     descendants: usize,
     /// Whether this node or any of its children are erroneous.
     erroneous: bool,
+    /// The (capped) amount of newlines in this node's descendants.
+    /// This is solely used to tell whether this node contains 0, 1, 2 or more
+    /// newlines. As such, this number is capped at 2, even though there may be
+    /// more newlines inside this node.
+    capped_newlines: u8,
```
laurmaedje, line 405: do we really need a field for this? is it too slow otherwise?

crates/typst/src/diag.rs
```diff
@@ -115,12 +115,14 @@ macro_rules! __error {
...
     (
         $span:expr,
+        $id:ident,
         $fmt:literal $(, $arg:expr)*
         $(; hint: $hint:literal $(, $hint_arg:expr)*)*
         $(,)?
     ) => {
         $crate::diag::SourceDiagnostic::warning(
             $span,
+            std::option::Option::Some($crate::diag::CompilerWarning::$id),
```
laurmaedje, line 125:
```suggestion
            ::std::option::Option::Some($crate::diag::CompilerWarning::$id),
```

crates/typst/src/diag.rs
```diff
@@ -228,6 +239,49 @@ impl From<SyntaxError> for SourceDiagnostic {
...
+pub enum Identifier {
+    /// Identifier for a built-in compiler warning.
+    Warn(CompilerWarning),
+    /// Identifier for a warning raised by a package.
+    User(EcoString),
+}
+
+/// Built-in compiler warnings.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum CompilerWarning {
```
laurmaedje, line 253: what's the pro/con for using an enum here instead of just using string keys as well?

crates/typst/src/diag.rs
```diff
@@ -228,6 +239,49 @@ impl From<SyntaxError> for SourceDiagnostic {
     }
 }
 
+/// The identifier of a [`SourceDiagnostic`].
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum Identifier {
+    /// Identifier for a built-in compiler warning.
+    Warn(CompilerWarning),
```
laurmaedje, line 246: The `Warn` variants is strangely named since both are warnings. Perhaps it should rather be `Compiler`.

crates/typst/src/diag.rs
```diff
@@ -228,6 +239,49 @@ impl From<SyntaxError> for SourceDiagnostic {
     }
 }
 
+/// The identifier of a [`SourceDiagnostic`].
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum Identifier {
```
laurmaedje, line 244: The name of this enum is maybe a bit too generic.

crates/typst/src/engine.rs
```diff
@@ -186,6 +308,9 @@ impl Sink {
     }
 
     /// Extend from another sink.
+    /// Using `Sink::extend_tracked` is preferable as it avoids a call to this
```
laurmaedje, line 311: A general nit about comment style: When a line is ended early due to a conceptual break (i.e. after "Extend from another sink" there should be a blank line as it's a new paragraph). Same above actually with "Later, during warning suppression [...]".

So either it is close enough to wrap at 80 columns or it should be a paragraph break.

crates/typst/src/engine.rs
```diff
@@ -146,20 +186,90 @@ impl Sink {
...
+            // If this warning wasn't suppressed, any further duplicates (with
+            // different tracepoints) should be removed.
+            should_raise && unsuppressed_warning_set.insert(hash)
+        });
+
+        self.warnings
+    }
+
+    /// Adds a tracepoint to all warnings outside the given span.
+    pub fn trace_warnings<F>(
```
laurmaedje, line 263: Not sure whether this method pulls its weight or whether we should just inline its definition at the only call site.

crates/typst/src/engine.rs
```diff
@@ -146,20 +186,90 @@ impl Sink {
...
+    /// For example, calling the same function twice but suppressing a warning
+    /// on the first call shouldn't suppress on the second, so each set of
+    /// tracepoints for a particular warning matters. If at least one instance
+    /// of a warning isn't suppressed, the warning will be returned, and the
+    /// remaining duplicates are discarded.
+    pub fn suppress_and_deduplicate_warnings(
+        mut self,
+        world: &dyn World,
+    ) -> EcoVec<SourceDiagnostic> {
+        let mut unsuppressed_warning_set = HashSet::<u128>::default();
```
laurmaedje, line 226: I think variable names could be less verbose sometimes. E.g. `unsuppressed` would be sufficient imo. At least the `_set` suffix is unnecessary since we can see the type.

crates/typst/src/engine.rs
```diff
@@ -202,6 +327,61 @@ impl Sink {
     }
 }
 
+/// Checks if a given warning is suppressed given one span it has a tracepoint
+/// in. If one of the ancestors of the node where the warning occurred has a
+/// warning suppression decorator sibling right before it suppressing this
+/// particular warning, the warning is considered suppressed.
+fn check_warning_suppressed(
```
laurmaedje, line 334: I think `is_warning_suppressed` is a better name as it makes immediately clear what `true` means and what `false` means.

crates/typst/src/engine.rs
```diff
@@ -202,6 +327,61 @@ impl Sink {
...
+    span: Span,
+    world: &dyn World,
+    identifier: &diag::Identifier,
+) -> bool {
+    let Some(file) = span.id() else {
+        // Don't suppress detached warnings.
+        return false;
```
laurmaedje, line 341: For such simple let-else, I generally prefer the single-line style (which rustfmt only does if the semicolon is removed.)

crates/typst/src/engine.rs
```diff
@@ -202,6 +327,61 @@ impl Sink {
...
+    };
+
+    // The source must exist if a warning occurred in the file,
+    // or has a tracepoint in the file.
+    let source = world.source(file).unwrap();
```
laurmaedje, line 346: better safe than sorry imo, I would just skip it. could also combine it with the previous thing:

```rs
let Some(source) = span.id().and_then(|id| world.source(id)) else { return false };
```

crates/typst-syntax/src/parser.rs
```diff
@@ -1562,8 +1564,8 @@ impl<'s> Parser<'s> {
         self.prev_end
     }
 
-    fn current(&self) -> SyntaxKind {
-        self.current
+    fn current_kind(&self) -> SyntaxKind {
```
laurmaedje, line 1567: I think we could keep the name `fn current` for now. I would thus like to keep the diff as plain as possible for now since the parser isn't the focus.

crates/typst/src/engine.rs
```diff
@@ -146,20 +186,90 @@ impl Sink {
...
+    /// This is so we can attempt to suppress each tracepoint separately,
+    /// without having one suppression discard all other attempts of raising
+    /// this same warning through a different set of tracepoints.
+    ///
+    /// For example, calling the same function twice but suppressing a warning
+    /// on the first call shouldn't suppress on the second, so each set of
+    /// tracepoints for a particular warning matters. If at least one instance
+    /// of a warning isn't suppressed, the warning will be returned, and the
+    /// remaining duplicates are discarded.
+    pub fn suppress_and_deduplicate_warnings(
```
laurmaedje, line 222: I think this would be better as a free-standing function in diag.rs. The `fn deduplicate` in lib.rs could also be moved there. Perhaps have `fn deduplicate_errors` and `fn deduplicate_and_suppress_warnings` (swapping the naming order to align the two).

crates/typst/src/diag.rs
PgBiel: This was mostly because there used to be an `Error` variant, but I removed it since I deemed it out of scope for the PR; we can keep it open to adding error identifiers in the future.

Though I'd still accept a change in the name, however just `Compiler` would become ambiguous in that context. Maybe `CompilerWarn` or similar.

crates/typst/src/diag.rs
PgBiel: The thought I had is that it's already scoped under `diag::`, so `diag::Identifier` does it justice. We could, however, prefix it with `Diagnostic` or similar.

crates/typst/src/diag.rs
PgBiel: I believe this is the simplest way to have our set of warnings in one place, in a way we can add features to. I had implemented identifier categories using this enum, for example, though I removed it to keep the PR's scope small. We could even generate the warning messages using the enum, but I think this would require some further design discussion (and it could also not be worth it), so I haven't made any efforts in that direction for this PR.
There is also the pro of type-safety when invoking `warn!`, so we can be sure to not accidentally create new identifiers without listing them here first, ensuring that would be an intentional change.

The con is that the identifier and other information are detached from where the warning is actually emitted, but I think it's fine as long as it's in `diag`, or in an appropriate location in general.

crates/typst/src/engine.rs
PgBiel: With your other suggestion, the line got long enough for `rust-analyzer` to insist on breaking it regardless.

crates/typst-syntax/src/node.rs
PgBiel: Hmmm. I didn't really measure it, but it seemed like a waste to compute this twice. However, right now, we're _always_ computing this, even when there are no warnings, which is evidently worse. Something like an `OnceCell` for each node could be ideal, but I'll probably just remove this field for now.

crates/typst-syntax/src/ast.rs
PgBiel: The syntax has been updated, though we may want to confirm it through further discussion.

crates/typst/src/diag.rs
PgBiel: Switched to just string keys for now after further discussion. We can reconsider enums if we need them in the future.

crates/typst/src/diag.rs
PgBiel: Resolving due to https://github.com/typst/typst/pull/4459#discussion_r1688231225

crates/typst/src/diag.rs
PgBiel: Resolving due to https://github.com/typst/typst/pull/4459#discussion_r1688231225

crates/typst-syntax/src/lexer.rs
```diff
@@ -182,6 +224,197 @@ impl Lexer<'_> {
...
+    fn annotation_name(&mut self, start: usize) -> SyntaxKind {
+        self.s.eat_while(is_id_continue);
+        let ident = self.s.from(start);
+
+        if ident == "allow" {
+            SyntaxKind::AnnotationName
+        } else {
+            let error = self.error(eco_format!("invalid annotation name"));
+            self.hint("must be 'allow'");
+            error
```
istudyatuni, line 364:
```suggestion
            self.error(eco_format!("invalid annotation name"))
                .hint("must be 'allow'")
```
Not sure if it works like this

crates/typst/src/diag.rs
```diff
@@ -410,6 +437,119 @@ impl<T> Hint<T> for HintedStrResult<T> {
...
+        return false;
+    }
+
+    for argument in annotation.arguments() {
+        if warning == argument.get() {
+            return true;
+        }
+    }
+
+    false
```
istudyatuni, line 550:
```suggestion
    annotation.name().as_str() == "allow" && annotation.arguments().iter()
        .any(|a| a.get == warning)
```

laurmaedje:
Based on the update posted by @PgBiel above, I'll close this for the time being:

> UPDATE (2024-08-06): After lots of discussion (internally, over Discord etc.) regarding not only the ideal syntax for annotations, but also the potential use cases for warn suppression, we've decided to prioritize other, more important improvements for diagnostics, so this PR won't move forward for the time being.

PR #4462, opened 2024-06-27
Fix approximated size of `reflow: true` transformations

### LINKED ISSUE #4237: `measure()` treats objects in `rotate(reflow: true, …)` as zero-width
### Description

The dimensions of an object that has been rotated with `reflow: true` seem to be different according to 1) an enclosing `rect` and 2) the contextual `measure` function. Specifically, the bounding box reported by measure is that of a rotated zero-width object.

MWE:
```typ
#set rect(inset: 0pt, stroke: red)

#let obj = rotate(30deg, reflow: true)[hello]

#rect(obj)

#context rect(
  width: measure(obj).width,
  height: measure(obj).height,
  obj,
)
```

Screenshot 2024-05-24 at 00 44 06

This is also observed for rotated objects other than text (see reproduction URL).

Perhaps related: the dimensions seem to be incorrect when the parent container sizes automatically to content:

```typ
#set rect(inset: 0pt, stroke: red)

Auto-sizing page:
#rect(rotate(30deg, reflow: true)[hello])
```
Screenshot 2024-05-24 at 02 19 18

### Reproduction URL

https://typst.app/project/rUM1E564AvpxiKtmw50JMW

### Operating system

### Typst version

shinyfelix:
- use the inverse transform for approximating sizes
- handle infite sizes
- adjust the rotate approximation

more information: https://discord.com/channels/1054443721975922748/1088371867913572452/1255820917616017449
Attempts to fix #4237 and the additional weird behavior shown in the link above.

I am not pleased with lines 156-158 right now, because I am unsure, how we should proceed, if only one size is infinite.


shinyfelix:
this pr: test
main: main
```
#set rect(inset: 0pt, stroke: red)

Auto-sizing page:
#rect(rotate(30deg, reflow: true)[hello])

#set text(size: 32pt)
#let scaled(body) = box(scale(
  x: 20%,
  y: 40%,
  body
), stroke: 1pt)

#set scale(reflow: false)
Hello #scaled[World]!

#set scale(reflow: true)
Hello #scaled[World]!

#let obj = rotate(30deg, reflow: true)[hello]

#rect(obj)

#context rect(
  width: measure(obj).width,
  height: measure(obj).height,
  obj,
)

#import "@preview/cetz:0.2.2"

#let obj = rect(rotate(30deg, reflow: true)[hello])

#obj

#cetz.canvas({
  cetz.draw.content(
    (0,0),
    obj
  )
})
```
Code from https://github.com/cetz-package/cetz/issues/604#issue-2312747274, #4237 and the transform-scale test.

shinyfelix:
The size when rotating is still not prefect (and I don't think it can be, because it would require being able to layout in rotated regions.)
```
Text rotated by 45° breaks out of bounds:\
#box(width: 100pt, height: 100pt, stroke: 1pt)[
  #rotate(45deg, lorem(50), reflow: true)
]
#v(6em)
Text rotated by 90° is good:\
#box(width: 30pt, height: 120pt, stroke: 1pt)[
  #rotate(-90deg, lorem(14), reflow: true)
]
```
test1

I want to at least fix it, so that the first example looks something like this:

test2

laurmaedje:
Thank you!

> The size when rotating is still not prefect (and I don't think it can be, because it would require being able to layout in rotated regions.)

> I want to at least fix it, so that the first example looks something like this: [...]

I think the changes here so far are definitely an improvement, so I'll merge this now. But feel free to open another PR if you come up with further improvements.

Issue #3585, opened 2024-03-08
More consistent json output in cli query


LDemetrios:
### Description

I propose to include type descriptors during serialization of the values of some types, the same way as it happens with different content types, for example:
```json
{
    "type": "datetime",
    "year": 2024,
    "month": 3,
    "day": 8
}
```
instead of
```json
"datetime(year: 2024, month: 3, day: 8)"
```

The same applies to `duration`, `regex`, `label` and `version` types, and, maybe, for `fraction`, `length`, `ratio`, `relative` and `alignment`:
```json
{
    "type": "relative",
    "percent": 50,
    "pt": -2,
    "em": 1,
}
```
instead of
```json
"50% + -2pt + 1em",
```

Things get even worse with parsing colors. I played with it a bit and got the number of possible variants scares me. It's at least 14. So, again, 
```json
{
    "type": "rgb",
    "hex": "#0b1621"
}
```
instead of
```json
"rgb(\"#0b1621\")",
```

### Use Case

I assume, one decides to use json because it is one of the most used format, and json libraries exist for every language. Thus, there is no need to write parsers, just work with the data. But when there's a JSON strig like `color.linear-rgb(100%, 49.8%, 49.8%, 50%)"`, additional parsing is required.

Moreover, there are many libraries that could convert JSON not to suitable structure (list, map, etc), but directly to the object model. Depending on the language, it may use reflection (JVM-based languages) or macros (Rust), or something else. But as far as it isn't Prolog, it is hard to provide arbitrary templates for values. It is common practice to have so-called type descriptors: parser gets it first, and then decides exactly which object to construct. It is already present in `content` representation, named `func`:
```json
 {
    "func": "sequence",
    "children": [
        {
            "func": "emph",
            "body": {
                "func": "text",
                "text": "aaa"
            }
        },
        {
            "func": "space"
        },
        {
            "func": "text",
            "text": "bbb"
        },
    ]
}
```

Thus I can create my own object model for typst content:

```kt
interface Content
data object Space : Content
data class Text(val text: String) : Content
data class Emph(val body: Content) : Content
```
Plus few annotation which handle inheritance, differences in naming case and default values. And then very simple call:
```kt
val content = json.deserialize<Content>("...")
```
It's important to note that this kind of thing is possible in many languages, so this will make life easier for many developers.

[...]

PR #4466, opened 2024-06-28
Typed serialization (#3585)

LDemetrios:
Pretty much described in the issue (https://github.com/typst/typst/issues/3585). What is done:

- implemented `Serialize` for `Args`, `Auto`, `Datetime`, `Duration`, `Label`, `Selector`, `Regex`, `Type`, `Version`, `Counter`, `State`, `Abs`, `Align`, `Angle`, `Axes`, `Dir`, `Fr`, `Length`, `Ratio`, `Rel`, `Color`, `Gradient`, `Paint`, `Stroke`. The resulting way of serialization is as in the file I attached (the result of `query "<a>" --pretty` on typ file)
[serialization-test.txt](https://github.com/user-attachments/files/16029333/serialization-test.txt)
(Github doesn't seem to support attaching .typ)
[otuput.json](https://github.com/user-attachments/files/16029338/otuput.json)

- added stubs for types `Module`, `Plugin`, `Location`, `CslStyle`, `Pattern`. 
- added `erased_serde::Serialize` to `Bounds` for `Value::Dyn` to dispatch serialization of dynamic values.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,6 +612,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
+name = "erased-serde"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "24e2389d65ab4fab27dc2a5de7b191e1f6617d1f1c8855c0dc569c94a4cbb18d"
+dependencies = [
+ "serde",
+ "typeid",
+]
+
+[[package]]
 name = "errno"
 version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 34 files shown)


LDemetrios:
I find some cases to be hard to handle in a consistent way, `state` being one of them. I tried to add some meaningful information where I could, but obviously there are cases when it's impossible at all (e.g. `plugin`). 

May be it would worth using only stubs for some types, and producing a warning about attempt to serialize unserializable. That would be (for all I understand) `plugin`, `location`, `pattern`, `CslStyle`. Regarding the pattern, it's especially sad, but I don't see a way to meaningfully serialize `Frame`... 

Speeking about `state` specifically. I was guided by idea that you should be able not only _take_ information from the document, but also process and _put it back_, by for example writing some code in nearby file and recompiling. And `state` constructor requires `key`. And I don't agree that `key` isn't accessible from the Typst code:
```typ
#let s = state("a", 0)
#repr(s).split("\"").at(1) 
// Obviously there is more going on with escapings in strings,
// but you get the idea
```
 ~~pls don't kl me~~
I'm by the way not that sure one would need to query state... If not, `state` can be the one of the types I talked in previous par

I'm not sure I understand what you are saying about "approach down the road". Is it "I should rewrite my PR in different approach", or "We'll see when all the values are unified"?

As for time, now that the academic year started, I will probably have less time to respond, sorry for that, but I'll try to keep up

LDemetrios:
Then I think I'll close this PR and may be return in the future, once type system is set. 

Don't mind if I keep fork open and (as much as I have time) up to date? Hope some people might wanna us it

laurmaedje:
Thanks for your understanding.

> Don't mind if I keep fork open and (as much as I have time) up to date?

Sure, it's all free and open source after all!

PR #4468, opened 2024-06-28
Fix duplicated and incomplete Thai characters at line endings

### LINKED ISSUE #3427: Repetition of Thai Characters
### Description

I has been typing a long paragraph in Thai, but when there are some floating characters, the base repeat itself before line break. As you can see, in the first line it should be "ผู้" then a linebreak and "นำ" as the first word of second line. However, from the image, the "นํ" repeat itself. How I can prevent this? 
Screen Shot 2567-02-11 at 21 51 28

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

OverflowCat:
Fixes #3427


crates/typst/src/layout/inline/shaping.rs
```diff
@@ -563,19 +563,6 @@ impl<'a> ShapedText<'a> {
...
-            Side::Right => usize::checked_add,
-        };
-
-        // Search for the outermost glyph with the text index.
-        while let Some(next) = next(idx, 1) {
-            if self.glyphs.get(next).map_or(true, |g| g.range.start != text_index) {
-                break;
-            }
-            idx = next;
-        }
```
OverflowCat, line 577: @laurmaedje Can you help me understand the scenarios where this deleted code would be used? Git blame seems not helpful. Based on my results from running all tests, this code is only triggered and produces incorrect output when encountering Thai language.

Marisada:
At find_safe_to_break function will call self.glyphs.binary_search_by to find an index of glyph with start range equal to text_index. But binary_search_by says “If there are multiple matches, then any one of the matches could be returned.” so when we find “end” to the “right”, we get any index of the next grapheme cluster and break at that point, getting some extra glyph from next grapheme cluster.
If we find “end” and to the “left”, we will get a first glyph’s index of next grapheme cluster to break.

https://github.com/typst/typst/blob/main/crates/typst/src/layout/inline/shaping.rs#L518-L519

so change from:

let right = self.find_safe_to_break(end, Side::Right)?;

to:

let right = self.find_safe_to_break(end, Side::Left)?;

may be solve?

OverflowCat:
Given this, passing in `towards` doesn't make a difference. Do I still need to keep this part of the logic?

Marisada:
When `binary_search_by` on `start` without `towards` also break grapheme cluster.

ex: glyph ranges of นกยางบินกันกิน 

นก : 0..3, 3..6
ยาง : 6..9, 9..12, 12..15
บิน : 15..15, 15..21, 21..24
กัน : 24..24, 24..30, 30..33
กิน : 33..33, 33..39, 39..42

`binary_search_by` 24 will be 24..30 (not 24..24) and break to 
นกยางบินก
 ันกิน

laurmaedje:
> At find_safe_to_break function will call self.glyphs.binary_search_by to find an index of glyph with start range equal to text_index. But binary_search_by says “If there are multiple matches, then any one of the matches could be returned.”

That is indeed the reason. But as the issue demonstrates, it means the `15..15` range will be picked up by both lines.

I'll need to think some more to figure out how best to fix this.

laurmaedje:
I'll close this as the change set as-is doesn't fix the issue, it just shifts it.

PR #4470, opened 2024-06-28
Fix suggestions for '.' after code mode has exited.

### LINKED ISSUE #2733: Incorrect suggestions for period after code mode has ended
### Description

This is quite easy to reproduce in a number of ways. Write for instance 

`#"hello" .`

A list will pop up with suggestions
image

However, whitespace is not allowed in this situation, so the suggestions are inappropriate

image

### Reproduction URL

_No response_

### Browsers

_No response_

### OS

_No response_

shinyfelix:
Fixes #2733
- Add `with_trivia` parameter to `LinkedNode::prev_sibling`
- Use this parameter in `compete_field_accesses` to disallow trivia nodes before '.' during an inline expression in markup mode.

Questions:
- Should such a parameter also be added to `LinkedNode::next_sibling` for consistency?
- Should a separate function (for example `LinedNode::prev_sibling_with_trivia`) be used instead, since `LinkedNode::prev_sibling` is mostly called with `with_trivia: false`


laurmaedje:
We could also, in the place where we need it, check that the end offset of the previous node and the start offset of the following node are equal. That might be simpler.

If we decide that we don't like that, I think I'd prefer an additional method over the parameter.

Could you please add a unit test as well?

laurmaedje:
Thanks!

PR #4471, opened 2024-06-28
Add `--fill-transparent` option

T0mstone:
This PR adds a simple switch `--fill-transparent` (name to be bikeshed) to `typst compile` that changes the background to be transparent instead of white.

I don't see a reason why typst shouldn't be able to generate PNGs with a transparent background.

Cf. https://github.com/typst/typst/discussions/3914#discussioncomment-9076050

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -109,6 +109,11 @@ pub struct CompileCommand {
     #[arg(long = "ppi", default_value_t = 144.0)]
     pub ppi: f32,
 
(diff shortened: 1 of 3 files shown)


laurmaedje:
See https://github.com/typst/typst/pull/4251#issuecomment-2152848695

laurmaedje:
Still not sure whether we'll make PNG transparent by default or add `#set page(fill: auto | none | paint)`, but I want to avoid CLI arguments as much as possible (as they are external and less portable), so I'll go ahead and close this already. Thanks still!

PR #4477, opened 2024-06-29
Wrap the body of outline in a box

### LINKED ISSUE #4476: RTL headings with LTR ending rendered incorrectly in Outline
### Description

Hi!

The following code reproduces the issue I'm encountering:

```typst
#set text(lang: "he")

#outline()

= הוקוס Pocus
```

which is rendered using the latest Typst version as:

outline-rendering-incorrectly

i.e. the "Pocus", which is part of the heading, "switches places" with the page number.

A workaround (which probably doesn't fix all cases) for now is adding:
```typst
#show outline.entry: it => {
  box(it.body)

  box(width: 1fr, it.fill)

  it.page
}
```

(The workaround is known to break when an outline entry needs multiple lines, see #4477).

in which case we get the desired result:

outline-rendering-correctly

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

noamzaks:
Fixes #4476 like the workaround I presented in it - by wrapping the Outline's body inside a box.

### DIFF
--- a/crates/typst/src/model/outline.rs
+++ b/crates/typst/src/model/outline.rs
@@ -500,7 +500,14 @@ impl Show for Packed<OutlineEntry> {
         };
 
         // The body text remains overridable.
-        seq.push(self.body().clone().linked(Destination::Location(location)));
+        seq.push(
+            BoxElem::new()
+                .with_body(Some(
+                    self.body().clone().linked(Destination::Location(location)),
+                ))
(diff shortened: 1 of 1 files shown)


noamzaks:
OK, the test is failing because this fix breaks multiline elements in the outline:

Figure 1 here is rendered incorrectly, the dots don't look good.

outline

PR #4478, opened 2024-06-29
Do not conflate Archivo Narrow and Archivo (Black) fonts

matze:
Archivo Narrow is a distinct typeface and not to be conflated with Archivo (Black). This fixes the issue not being able to select Archivo and Archivo Narrow and being stuck with one.

### DIFF
--- a/crates/typst/src/text/font/exceptions.rs
+++ b/crates/typst/src/text/font/exceptions.rs
@@ -48,6 +48,16 @@ static EXCEPTION_MAP: phf::Map<&'static str, Exception> = phf::phf_map! {
     // See https://corefonts.sourceforge.net/.
     "Arial-Black" => Exception::new()
         .weight(900),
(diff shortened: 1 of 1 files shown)


laurmaedje:
I downloaded Archivo Narrow from Google Fonts and it looks like it also has semibold variants which wouldn't be covered by this. 

This PR really shows again that the style trimming we currently do is really problematic: 

https://github.com/typst/typst/blob/2df138a507860dc0cb610943ad6e608a33d03d7d/crates/typst/src/text/font/book.rs#L339-L399

The exceptions were mostly intended for fixing fonts, not Typst :/

matze:
So, how do we go forward? Close this and somehow solve the style trimming issue more holistically? As a workaround, I removed one of the fonts but it's annoying to say the least.

laurmaedje:
We can revisit this later and merge the fix for now. Thanks!

PR #4480, opened 2024-06-30
Add horizonal fractions

shinyfelix:
- Add a `horizontal` parameter, which will layout fractions horizontally.
- Automatically surround the numerator/denominator with bracktes if they contain multiple elements
- Add `bracket-horizontal` (name could be changed) in order to switch off the behavior above, or force numerator and denominator to always be delimited

I'm not sure whether the way multiple elements are detected is the best way (or works in all expected cases). 
I'm also open to adding a way to support switching to horizontal mode if the math size would be to high otherwise (see https://discord.com/channels/1054443721975922748/1088371867913572452/1256637712182542377). 
Though that may be better in a separate pr.

### DIFF
--- a/crates/typst/src/math/frac.rs
+++ b/crates/typst/src/math/frac.rs
@@ -1,5 +1,5 @@
 use crate::diag::{bail, SourceResult};
-use crate::foundations::{elem, Content, Packed, StyleChain, Value};
+use crate::foundations::{elem, Content, Packed, SequenceElem, Smart, StyleChain, Value};
 use crate::layout::{Em, Frame, FrameItem, Point, Size};
 use crate::math::{
     scaled_font_size, style_for_denominator, style_for_numerator, FrameFragment,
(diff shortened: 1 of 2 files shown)


laurmaedje:
I think `bracket-horizontal` isn't really great UX. It seems to me that it would be better to keep the parentheses the user used for horizontal fractions (though that's harder to implement). I also wonder whether `inline` is a better term than `horizontal` (even though it is perhaps a little more overloaded).

shinyfelix:
> I think `bracket-horizontal` isn't really great UX. It seems to me that it would be better to keep the parentheses the user used for horizontal fractions (though that's harder to implement). I also wonder whether `inline` is a better term than `horizontal` (even though it is perhaps a little more overloaded).

I agree that when writing `$ (a + b) / c $` the user provided parentheses should be used. But we also have to consider the case where `math.frac` is called. $ frac(a + b,  c) $ should in my opinion also group the `a+b`, but the user doesn't provide any parenthesis.

laurmaedje:
> $ frac(a + b, c) $ should in my opinion also group the a+b, but the user doesn't provide any parenthesis.

I sort of agree, but it's also feels a bit implicit.

laurmaedje:
I think this needs more design (ideally in a forge thread on Discord) before we can continue on the implementation side, so I'll close this PR for now. Thanks still!

Issue #1030, opened 2023-04-28
Optimize optimized paragraph layout


rpitasky:
Notwithstanding some confusion about what “runtime complexity” is, this should definitely be looked at.

### Discussed in https://github.com/typst/typst/discussions/489

<div type='discussions-op-text'>

<sup>Originally posted by **TS40** March 31, 2023</sup>
I looked a bit how Typst scales when more than a few pages are created.
While it is much faster than PdfTeX for a few pages, it will be ca. 2 times slower with many pages.
So, I was interested what complexity it has.

Here are my first ideas using `#lorem` and `#set par(justify: true)`.
I edited the source code to print the calls to `line` function and at the end iterated over the lines result and printed the widths to verify I got it correct.

```
lorem5000(default width):     200882 calls to line;    345 final lines; runtime=582*N
lorem50000(default width):   1992536 calls to line;   3442 final lines; runtime=579*N
lorem500000(default width): 20018244 calls to line;  34323 final lines; runtime=583*N
```

With the default settings and additionally `#set par(justify: true)` the complexity will be <= 580*N where N is the number of text lines in the document (because the document is not only text lines and other things are probably cheaper).

I have not yet understood the algorithm good enough. But I think it is possible to make the constant 580 quite a bit smaller by either some kind of binary searching or by filtering impossible candidates without shaping them (but maybe this is already done, not sure).

Note that the smaller the text width and the less text lines, the faster it will be:

```
lorem500(width: 5cm):           5443  calls to line;    157 final lines; runtime=35*N
lorem5000(width: 5cm):         51576  calls to line;   1570 final lines; runtime=33*N
lorem50000(width: 5cm):       511380  calls to line;  15686 final lines; runtime=33*N
```

Update:
[Here](https://github.com/jaroslov/knuth-plass-thoughts/blob/master/plass.md#further-optimizations--features) it says, the runtime of the algorithm is approximately `max_number_of_words_in_a_line * number_of_linebreaks` in an optimized variant of the algorithm. In the created pdf with default width, we have approximately 20 words per line, so this is quite a bit different from 580. But maybe I mixed up some parameters like "lines" and "line breaks".</div>

peng1999:
The current algorithm implemented in typst call `line` function $c=\alpha mn$ times, where $m$ is words per line, $n$ is the number of total words in text, and $\alpha$ is some constant. Actually we can calculate $\alpha$ from your data:

[...]

PR #4483, opened 2024-07-01
Optimize optimized paragraph layout

laurmaedje:
This PR optimizes the optimal paragraph layout used when justification is enabled. Typst uses the Knuth-Plass (TeX) algorithm for producing optimal paragraphs. This dynamic programming algorithm minimizes a cost function across the paragraph.

## Background

To compute these costs, we must determine the widths of many line attempts. In the original TeX implementation, computing these widths was simple: Simply add up the precomputed widths of the individual boxes and glues in the line. This process can even be further accelerated by building an accumulated list of the widths up to a certain point. Then, we can get the width of a range through subtraction in $O(1)$.

Unfortunately, this approach of computing widths falls down with proper text shaping. Shaping two subruns of text and summing up their widths doesn't necessarily give the same width as shaping the whole run due to various contextual adjustments and replacements that can occur.

For this reason, Typst must properly build a line to know its width. We do reuse shaping results from an initial shaping pass if possible, but even if that works out, there is still quite a bit of overhead in the line computations.

As observed in https://github.com/typst/typst/issues/1030#issuecomment-1528464333, for a paragraph with $n$ words, on average $m$ words line per line and hyphenation enabled, we make roughly $\alpha m n$ line attempts (with $\alpha \approx 3$ due to hyphenation). For example, `lorem(200)` on normal A4 paper results in 7775 line attempts. This is costly.

## Approach

This PR brings the number of line attempts in the exact same `lorem(200)` scenario down to 444, using the following approach:

1. Observe that directly taking the width of a sliced line still gives us not the exact width, but still a good approximation of it.
2. Build a TeX-like accumulated list for the paragraph to very cheaply compute width and cost estimates for lines.
3. Optimize the paragraph with these estimates (`linebreak_optimized_approximate`).
4. Compute the _real, non-approximate_ cost $c$ of the resulting paragraph.
5. Optimize the paragraph with exact layout, but pruning the search space using $c$ as an upper bound on the possible cost (`linebreak_optimized_bounded`).

We use the upper bound $c$ in two ways:

- When a line's exact cost plus that of its optimal predecessor exceeds the bound, we do not consider it as a candidate predecessor for future lines and don't even push it into our dynamic programming table.

- When a line attempt was underfull, we know that any shorter line will have higher cost. Thus, we can skip any line attempt for which the optimal predecessor cost plus the line's lower bound exceeds the paragraph's upper bound. (The exception are lines with negative spacing, which we exclude.)

## Results

This optimization is most effective for short paragraphs. The longer the paragraph gets, the less likely it is that a single bad line already exceeds the upper bound. As a result, we build more bad lines. The graphs below show how many lines we build with and without the bound approach.

The X axis denotes the number of words $n$ and the Y axis shows the number of lines we built to find the optimal layout of the document `lorem(n)` (with hyphenation enabled). The left graph is zoomed in to $n \leq 1000$ while the right graph is zoomed out to $n \leq 3000$. Both show the same data.

For 50 words, we are ~20x as efficient. For 250 words, we are at ~15x. For 500 words, it's still ~10x. For 1000 words, we are down to ~3x. And for 2000 words, it's only ~1.15x. Close to 2500 words our bound fails (because the estimate was bad and we produced an invalid line [^1]).

Number of line attempts

## Performance

Finally, here are some raw performance measurements on a document with 100,000 words spread across lots of page runs and a paragraph size of 300 words. That's roughly 2.5 paragraphs per page, so on the longer side as far as paragraphs go.

<details>
<summary>Source code</summary>

```typ
#set heading(numbering: "1.")
#set par(justify: true)

#outline()

#{
  lorem(100000)
    .split()
    .chunks(300)
    .map(c => c.join(" "))
    .chunks(50)
    .map(c => {
    pagebreak()
    [= Heading]
    c.join(parbreak())
    })
    .join()
}
```
</details>

| Version              | Execution Time |
|:---------------------|---------------:|
| 0.11.1               | 5.75s          |
| main (e90c30903)[^2] | 5.57s          |
| PR (9cbc92dce)       | 0.68s          |

Closes https://github.com/typst/typst/issues/1030

[^1]: This failure case can also happen with fewer words. But we try to avoid it by allowing less space shrinking in approximate layout. If it occurs, it's as if the optimization didn't exist, but no further bad things happen.

[^2]: Interestingly, main gives barely any improvement over 0.11.1 even though the document has lots of page runs and is thus multi-threading-friendly. Runtime seems to be bounded by something else, perhaps memory.

### DIFF
--- a/crates/typst/src/layout/inline/line.rs
+++ b/crates/typst/src/layout/inline/line.rs
@@ -3,7 +3,7 @@ use unicode_bidi::BidiInfo;
 use super::*;
 use crate::engine::Engine;
 use crate::layout::{Abs, Em, Fr, Frame, FrameItem, Point};
-use crate::text::TextElem;
+use crate::text::{Lang, TextElem};
 use crate::utils::Numeric;
 
 /// A layouted line, consisting of a sequence of layouted paragraph items that
@@ -129,7 +129,7 @@ pub fn line<'a>(
     p: &'a Preparation,
     mut range: Range,
     breakpoint: Breakpoint,
-    prepend_hyphen: bool,
+    pred: Option<&Line>,
 ) -> Line<'a> {
     let end = range.end;
     let mut justify =
@@ -148,6 +148,8 @@ pub fn line<'a>(
             dash: None,
         };
     }
+
+    let prepend_hyphen = pred.map_or(false, should_insert_hyphen);
 
     // Slice out the relevant items.
     let (mut expanded, mut inner) = p.slice(range.clone());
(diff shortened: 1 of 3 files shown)


laurmaedje:
> binary searching candidates 

This works for simple (first-fit) linebreaking. I thought about doing it but skipped it for now because for simple linebreaks the bottleneck is `prepare`, not `linebreak`.

For optimized linebreaking it's much harder. I think it is possible to skip even more lines with something in that direction, but it would need a lot of care to implement it correctly.

> median words per line calculation and median interword spacing  ... from the previous text block to approximate the following

I'm not sure this would help much. We want an optimal solution after all. The approximation pass is already very fast and the exact pass can't really do such things.

Issue #1537, opened 2023-06-20
`datetime.display` long representations do not take the text language into consideration


chloe-the-catgirl:
### Description

> I would let it depend on whatever is set in `text`. So for example if someone sets `#set text(lang: "de")`, the date should use the German version as well.
> _Originally posted by @LaurenzV in https://github.com/typst/typst/issues/303#issuecomment-1489005819_

This is not working right now. Minimal working example:

```typst
#set text(lang: "de")
#datetime.today().display("[month repr:long]")

// Output is: June
```

For me, it's not critical, just wanted to write it down, so it does not get forgotten.

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
Yes, that's a known limitation. The formatting to string happens at a stage where the text language is not yet known. Hopefully we can improve this in the future.

LaurenzV:
Even if we would know the text language, it's still going to be tricky because the underlying library we use doesn't support  translating the formatted months/weeks into a different language. Will probably require us to somehow implement this on our side.

GoldsteinE:
There needs to be some way to specify declension for this to work with some languages, e.g. it’s «13 ноября», not «13 ноябрь».

TeX’s datetime2 does this by redefining date display comments, which is IMO not ideal.

https://www.dickimaw-books.com/faq.php?action=view&category=datetime2#datedeclension

FNG-2002:
Hello everyone, is there any update on this?

laurmaedje:
This is currently blocked on larger work to enable show rules on types like `datetime`, so it will still take some time.

LaurenzV:
FWIW there is currently a package that allows you to format datetimes using different locales: https://typst.app/universe/package/icu-datetime

Perhaps we should also rework the Typst implementation to use `icu4x` instead of the `time` crate for date formatting. The main challenge would probably be how to lazily load the `icu4x` data.

laurmaedje:
How much data are we talking?

LaurenzV:
I'm not sure, but according to the package it's 8MB... Although I guess some of it will also be the wasm code for the library.

laurmaedje:
Okay that's a lot.

LaurenzV:
But from what I've seen icu data should be pretty modular, so it should be possible to load whatever you need on demand. But I haven't looked into it more deeply.

PR #1355, opened 2023-05-28
Use icu4x for linebreaking algorithm

peng1999:
Remove `xi-unicode`, which is discontinued and hard to tailor.

Fix #1009, #335

Currently I use LSTM model rather than dictionary for East Asian language support. In `icu_segmenter`'s document:

> The LSTM, or Long Term Short Memory, is a machine learning model. It is smaller than the full dictionary but more expensive during segmentation (inference).

Also made #1164 easier to solve.

PR #4485, opened 2024-07-02
Allow displaying datetime in any locale

### LINKED ISSUE #2840: Datetime should be settable, and/or formatted by language
### Description

While it is currently possible to set the format of ``` #datetime.display() ```, it is not possible to use set rules, as it is not an ```Element```. Datetime format also does not adapt to the text language.

### Use Case

Useful for multi-language templates or just using a specific date format more then one time.

lynn:
Fixes #2840.

### DIFF
--- a/crates/typst/src/foundations/datetime.rs
+++ b/crates/typst/src/foundations/datetime.rs
@@ -3,6 +3,10 @@ use std::hash::Hash;
 use std::ops::{Add, Sub};
 
 use ecow::{eco_format, EcoString, EcoVec};
+use icu_datetime::options::length;
+use icu_datetime::{DateFormatter, DateTimeFormatter, TimeFormatter};
+use icu_locid::Locale;
+use length::Bag;
 use time::error::{Format, InvalidFormatDescription};
 use time::macros::format_description;
 use time::{format_description, Month, PrimitiveDateTime};
@@ -33,6 +37,9 @@ use crate::World;
 ///   "y:[year repr:last_two]"
 /// )
 ///
+/// #date.display(locale: "de", date: "full") \
+/// #date.display(locale: "en-u-ca-japanese")
+///
 /// #let time = datetime(
 ///   hour: 18,
 ///   minute: 2,
@@ -43,6 +50,8 @@ use crate::World;
 /// #time.display(
 ///   "h:[hour repr:12][period]"
 /// )
+///
+/// #time.display(locale: "ar", time: "short")
 /// ```
 ///
 /// # Datetime and Duration
@@ -69,9 +78,9 @@ use crate::World;
 ///
 /// # Format
 /// You can specify a customized formatting using the
-/// [`display`]($datetime.display) method. The format of a datetime is
-/// specified by providing _components_ with a specified number of _modifiers_.
-/// A component represents a certain part of the datetime that you want to
+/// [`display`]($datetime.display) method. The format of a datetime is specified
+/// by providing _components_ with a specified number of _modifiers_. A
+/// component represents a certain part of the datetime that you want to
 /// display, and with the help of modifiers you can define how you want to
 /// display that component. In order to display a component, you wrap the name
 /// of the component in square brackets (e.g. `[[year]]` will display the year).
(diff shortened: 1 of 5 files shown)


crates/typst/src/foundations/datetime.rs
```diff
@@ -341,30 +430,52 @@ impl Datetime {
...
+        /// The length of the localized time. Can be `"medium"` (hour, minute)
+        /// or `"short"` (hour, minute, second).
+        #[named]
+        #[default(length::Time::Medium)]
+        time: length::Time,
     ) -> StrResult<EcoString> {
         let pat = |s| format_description::parse_borrowed::<2>(s).unwrap();
-        let result = match pattern {
-            Smart::Auto => match self {
+        match (pattern, locale) {
```
Coekjan, line 460: I think `locale` here could be `Smart<Str>`, and when it is `auto`, we could infer the `locale` from `text.lang` & `text.region` (if it is ok).

Coekjan (commented):
Great feature, and here are my thoughts.

LaurenzV:
What's the impact on binary size for this? See also https://github.com/typst/typst/issues/1537#issuecomment-2156418750.

lynn:
Before this PR:

```
 % wc -c target/release/typst
 29554984 target/release/typst
```

After this PR:
```
 % wc -c target/release/typst
 41625624 target/release/typst
```
Ouch. OK, I should definitely think about reducing it :) Maybe we can drop support for non-Gregorian calendars.

laurmaedje:
I'm not a big fan of having more or less the raw time + icu4x APIs shoehorned into a single function. Both don't really feel natively Typst to me. 

- The `time` format strings work very differently from other stuff in Typst (e.g. numbering, though these will also likely change)
- The icu single-string `locale` is at odds with Typst's approach of specifying lang and region separately. The `date` and `time` widths enums feel a bit ad-hoc.

I don't have a good proposal how to deal with this. I think the capability added by this is very nice, but the API design is hard.

I also agree that this should integrate with context (ideally, it will become `datetime`'s show rule in a future where types and content are the same). Then, the configuration would likely be something on `#set datetime(..)`. This will be a breaking change anyway, but we can still keep it a bit in mind when designing the API.

On another note, we usually generate the icu4x data and check it into [typst-assets](https://github.com/typst/typst-assets/blob/4ee794cf8fb98eb67194e757c9820ab8562d853b/src/lib.rs#L25). **However,** I must admit I'm not 100% sure why we do that instead of using `compiled_data`, it just started that way in https://github.com/typst/typst/pull/1355. Maybe it results in a leaner build? cc @peng1999

peng1999:
> **However,** I must admit I'm not 100% sure why we do that instead of using `compiled_data`, it just started that way in #1355. Maybe it results in a leaner build? cc @peng1999

We don't use `compiled_data` because the linebreak algorithm needs customized data provider. After unicode-org/icu4x#3284, maybe we can switch to `compiled_data`. But for now, I suggest stick on generated data and avoid `compiled_data` to prevent binary duplication.

laurmaedje:
> We don't use compiled_data because the linebreak algorithm needs customized data provide

couldn't we use a custom blob specifically for linebreak data, but compiled_data for everything else?

peng1999:
> couldn't we use a custom blob specifically for linebreak data, but compiled_data for everything else?

Maybe yes. Need to adjust flags of icu-datagen.

lynn:
Another approach is to only leave only the `time` format string API, and then use find-and-replace to localize the result.

So, `datetime.today().display("[weekday repr:long], [day] [month repr:long]")` would do something like

* Pass the string to `time-rs` and get back `s = "Sunday, 7 July"`.
* Look up the lang/region in typst and see that it is e.g. `nl`
* Ask `icu` for the month and weekday names for that language.
* Make a new string using the localized names: `s.replace("Sunday", "zondag").replace("July", "juli")`

laurmaedje:
I think this needs more design (ideally in a forge thread on Discord) before we can continue on the implementation side, so I'll close this PR for now. With types and thus show rules for datetimes coming in the future, there's also more to consider API-design-wise.

 Thanks still!

PR #4488, opened 2024-07-02
Rename `notes.*` to `note.*` & Add some music symbols

### LINKED ISSUE #4481: Add the flat (♭) and sharp (♯) symbols
### Description

Adding two symbols to the existing symbol catalogue.

### Use Case

These symbols are commonly used in music notation and differential geometry, which makes them a useful addition.

Coekjan:
Closes #4481 .

This PR adds some music symbols about note, rest and pitch.


crates/typst/src/symbols/sym.rs
```diff
@@ -495,6 +494,27 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
...
+        fr8: '𝅘𝅥𝅮',
+        fr8.alt: '♪',
+        fr8.beamed: '♫',
+        fr16: '𝅘𝅥𝅯',
+        fr16.beamed: '♬',
+        fr32: '𝅘𝅥𝅰',
+        fr64: '𝅘𝅥𝅱',
+        fr128: '𝅘𝅥𝅲',
+    ],
+    natual: '♮',
```
Enter-tainer, line 514: typo？

Coekjan: sorry for that, fixed :cry:

Enivex:
I'm not sure the naming of the notes is in line with other naming. Wouldn't it be better to use more verbose names with for instance "half" and "quarter"?

Coekjan:
> I'm not sure the naming of the notes is in line with other naming. Wouldn't it be better to use more verbose names with for instance "half" and "quarter"?

I did name them "half" (current "fr2") and "quarter" (current "fr4") first, but I then thought "thirtysecond" (current "fr32"), "sixtyfourth" (current "fr64") and "hundredtwentyeighth" (current "fr128") are too long 🤣. So I finally name them all to "frx", where the meaning of "fr" is like `fr` function and "x" is the arabic numeral.

Enivex:
> > I'm not sure the naming of the notes is in line with other naming. Wouldn't it be better to use more verbose names with for instance "half" and "quarter"?
> 
> I did name them "half" (current "fr2") and "quarter" (current "fr4") first, but I then thought "thirtysecond" (current "fr32"), "sixtyfourth" (current "fr64") and "hundredtwentyeighth" (current "fr128") are too long 🤣. So I finally name them all to "frx", where the meaning of "fr" is like `fr` function and "x" is the arabic numeral.

Is the fr supposed to stand for fraction?

Coekjan:
> Is the fr supposed to stand for fraction?

Ah, yes. "fr" stands for `fraction` (not `fr` function).

Coekjan:
> Why not just `notes.4`, `notes.8`, etc.?

Unfortunately, current symbol macro infrastructure only support parsing valid rust identifiers to the symbol names (and variant names).

Coekjan:
> I guess one compromise could be to (additionally?) use full names for 1,2,4,8 and 16? They're likely the most used anyway.

Maybe this should be discussed with team. Let's move to discord.

> Would it also make sense to add rest symbols?

It would make sense. I will add them later.

alerque:
> Let's move to discord.

This is such an anti-pattern...

Coekjan:
See discord ( https://discord.com/channels/1054443721975922748/1088371867913572452/1257675683090534460 ). I think we can:
1. allow variant names starting with numbers
2. name the notes: `whole`, `half`, `quarter`, `8th`, `16th`, etc.

max397574:
> This is such an anti-pattern...



> > Let's move to discord.
> 
> This is such an anti-pattern...

also wouldn't make sense to discuss about design of symbol names in a pr which isn't directly related to this

Coekjan:
> See discord ( https://discord.com/channels/1054443721975922748/1088371867913572452/1257675683090534460 ). I think we can:
> 
> 1. allow variant names starting with numbers
> 2. name the notes: `whole`, `half`, `quarter`, `8th`, `16th`, etc.

Now we saw that the first step cannot be a good choice, because it requires big modification on lexer/parser foundation.

Coekjan:
Finally, we thought `whole`, `half`, `quarter`, `v8`, `v16`, `v32`, `v64`, `v128` could be okay.

max397574:
and what would the `v` mean?

Coekjan:
> and what would the `v` mean? would there be sth useful to memorize this?

https://en.wikipedia.org/wiki/Note_value said they are note value.

max397574:
well but the value wouldn't be 16 but rather 1/16

Enivex:
What is your opposition to the naming @alerque ?

hunkyjimpjorps:
> well but the value wouldn't be 16 but rather 1/16

I made the suggestion based on the way [Lilypond](https://lilypond.org/doc/v2.23/Documentation/notation/writing-rhythms) specifies note durations; it's understood that the value is the reciprocal of the duration

Enivex:
> well but the value wouldn't be 16 but rather 1/16

As long as there aren't notes not of the form 1/n in Unicode I don't see a big issue with this.

StarlightScribe:
Probably best to follow the Unicode names closely for consistency. "BeamedAscending" for example. 
https://www.compart.com/en/unicode/U+1F39C#UNC_DB

thecaralice:
Another suggestion for note/rest lengths is using british names: maxima, longa, breve, semibreve, minim, quaver, semiquaver, demisemiquaver etc.

Coekjan:
There may be some controversial aspects of the current design (note/rest names) in this PR. I propose to put those controversial parts on hold for now and leave them for further discussion in the future. In the meantime, I would like to request merging the non-controversial parts:
1. Rename `notes.{up,down}` to `note.beamed.{ascend,descend}`: follows names in https://www.compart.com/en/unicode/U+1F39C (as @StarlightScribe  suggested)
2. Add some music symbols about note, rest and pitch: adds `{note,rest}.{whole,half,quarter,eighth,sixteenth}` but the 1/32, 1/64 and 1/128 note/rest is left to be discussed in the future.

Enivex:
> 1. Rename `notes.{up,down}` to `note.beamed.{ascend,descend}`: follows names in https://www.compart.com/en/unicode/U+1F39C (as @StarlightScribe  suggested)

Note that there's already `dots.up` and `dots.down`.

Coekjan:
> I guess the naming doesn't _have_ to be consistent, but I'm not sure what we gain from the long `ascend` and `descend`. There are plenty of places where typst differs from the exact wording in the Unicode name.

In music, "up"/"down" may have different meanings. If we use "up"/"down" here, would it be confusing?

See `sharp.up` and `sharp.down`: https://www.compart.com/en/unicode/U+1D130, https://www.compart.com/en/unicode/U+1D131

Also see `kievan.quarter.down` and `kievan.quarter.up`: https://www.compart.com/en/unicode/U+1D1E4, https://www.compart.com/en/unicode/U+1D1E5

Coekjan:
> Following typst's naming convention those would be `sharp.t` and `sharp.b` I believe

I updated their names following the convention and additionally renamed `note.grace.noslash` to `note.grace`.

crates/typst/src/symbols/sym.rs
```diff
@@ -495,6 +494,59 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
...
+    note: [
+        up: '🎜',
+        down: '🎝',
+        whole: '𝅝',
+        half: '𝅗𝅥',
+        quarter: '𝅘𝅥',
+        quarter.alt: '♩',
+        eighth: '𝅘𝅥𝅮',
+        eighth.alt: '♪',
+        eighth.beamed: '♫',
```
laurmaedje, line 507: shouldn't this `beamed` also be `up` and `down`?

crates/typst/src/symbols/sym.rs
```diff
@@ -495,6 +494,59 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
...
+        sixteenth.beamed: '♬',
+        grace: '𝆕',
+        grace.slash: '𝆔',
+        kievan.final: '𝇠',
+        kievan.whole: '𝇢',
+        kievan.half: '𝇣',
+        kievan.quarter.t: '𝇥',
+        kievan.quarter.b: '𝇤',
+        kievan.eighth.t: '𝇧',
+        kievan.eighth.b: '𝇦',
```
laurmaedje, line 518: I can see the ones above, but `kievan` seems very obscure. I don't think we need everything.

crates/typst/src/symbols/sym.rs
```diff
@@ -495,6 +494,59 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
...
+        half: '𝄼',
+        quarter: '𝄽',
+        eighth: '𝄾',
+        sixteenth: '𝄿',
+    ],
+    natural: [
+        '♮',
+        t: '𝄮',
+        b: '𝄯',
+    ],
```
laurmaedje, line 533: `sym.rest` and `sym.natural` aren't obviously tied to music. I think that's a problem.

Enivex: > `sym.rest` and `sym.natural` aren't obviously tied to music. I think that's a problem.

Isn't that an argument that could be made for half the symbols? Does it actually matter unless there's any danger of ambiguity?

Rest and natural are the common names for them.

laurmaedje: > Isn't that an argument that could be made for half the symbols in sym?

can it? I just scrolled through the file and for most symbols it felt fairly obvious what they are but if I would see `sym.natural` I'd have no clue what it is.

Coekjan: AFAIK, LaTeX has `\natural`, `\sharp`, `\flat`.

Enivex: > > Isn't that an argument that could be made for half the symbols in sym?
> 
> can it? I just scrolled through the file and for most symbols it felt fairly obvious what they are but if I would see `sym.natural` I'd have no clue what it is.

This would depend on the reader no?

I guess you could put them under `pitch` or something, as in `pitch.natural`, `pitch.sharp` and `pitch.flat`. Though that makes them quite verbose.

laurmaedje: Grouping natural/sharp/flat together sounds nice!

crates/typst/src/symbols/sym.rs
Coekjan: These `*.beamed` symbols do not indicating the `up` or `down` in their (standard) unicode names. Naming them `up` or `down` can be over-interpretation.

crates/typst/src/symbols/sym.rs
Coekjan: I will remove them later.

crates/typst/src/symbols/sym.rs
laurmaedje: Nevermind, I confused beamed with ascend/descend

crates/typst/src/symbols/sym.rs
Coekjan: If we group them, which one shall be the primary symbol? And I don't think `flat.*` (and `sharp.*`) are the variants of `natural.*`.

Enivex: > If we group them, which one shall be the primary symbol? And I don't think `flat.*` (and `sharp.*`) are the variants of `natural.*`.

Does there need to be a primary symbol?

Coekjan: In my knowledge to Typst symbols, they always have a primary symbol and may have some variants.

Enivex: > In my knowledge to Typst symbols, they always have a primary symbol and may have some variants.

You already have examples here where that's not the case, like `rest`

Coekjan: By "primary symbol", I am wondering what should users get when they use `pitch` (without something like `.natural` following it).

Enivex: > By "primary symbol", I am wondering what users shall get when they use `pitch` (without something like `.natural` following it).

The only logical choice would be natural.

(That being said, I'm not a super-fan of relying on fallback like this)

Coekjan: :heart_hands: Done. Please review my newest code.

laurmaedje: According to Wikipedia, it looks like these things are called [accidentals](https://en.wikipedia.org/wiki/Accidental_(music)).

laurmaedje: Symbol naming is so hard ... since symbol variants are also not really intended as modules, but more really as variants I'm not 100% sure what to do here. If both of you actually prefer having the three top-level I think I would also be okay with that, given the LaTeX and HTML precedent. (Sorry for the back and forth, I'm just unsure.)

Enivex: > According to Wikipedia, it looks like these things are called [accidentals](https://en.wikipedia.org/wiki/Accidental_(music)).

It's been a long time since I read sheet music, but as far as I understand it's a specific use of them that is known as accidentals. Not if they they're used in a key signature for instance.

Coekjan: :heart_hands: Done again. Please review the newest code.

laurmaedje:
We definitely have the nicest looking bike shed in town. Thanks!

PR #4490, opened 2024-07-02
Conversion from int to bytes and back

### LINKED ISSUE #2868: Conversion from int to bytes and back
### Description

Currently `bytes` can only be constructed from a string or an array of `int` < 256.

For some use cases, especially when interacting with plugins, it would be useful to convert bigger `int` to `bytes` and back (given the length of the `bytes` is correct)

Note that this would require choosing a byte ordering, as to not leave the behavior platform dependent.

### Use Case

Interacting with plugins

chicoferreira:
Closes #2868.

Implements conversion from int to bytes and vice-versa.

Example:
```typ
#array(10000.to-bytes("big"))
#int.from-bytes(bytes((0, 0, 0, 0, 0, 0, 0, 1)), "little")
```
Output:
```
(0, 0, 0, 0, 0, 0, 39, 16)
72057594037927936
```

Maybe this should be extended to the float type as well? Waiting on implementation feedback.

### DIFF
--- a/crates/typst/src/foundations/int.rs
+++ b/crates/typst/src/foundations/int.rs
@@ -214,40 +214,44 @@ impl i64 {
         }
     }
 
-    /// Converts bytes to an integer. The bytes will be treated as a 64-bit signed integer.
-    /// The bytes should be at most 8 bytes (64 bits) in size to fit in a 64-bit integer, otherwise an error will occur.
-    ///
-    /// If the signed parameter is true and the most significant bit is set,
-    /// the number will be treated as negative, filling the remaining bytes with 0xFF (two's complement).
+    /// Converts bytes to an integer.
     ///
     /// ```example
-    /// #int.from-bytes(bytes((0, 0, 0, 0, 0, 0, 0, 1)))
+    /// #int.from-bytes(bytes((0, 0, 0, 0, 0, 0, 0, 1))) \
     /// #int.from-bytes(bytes((1, 0, 0, 0, 0, 0, 0, 0)), endian: "big")
     /// ```
     #[func]
     pub fn from_bytes(
         /// The bytes that should be converted to an integer.
+        ///
+        /// Must be of length at most 8 so that the result fits into a 64-bit
+        /// signed integer.
         bytes: Bytes,
-        /// Endianness of the conversion.
+        /// The endianness of the conversion.
         #[named]
         #[default(Endianness::Little)]
         endian: Endianness,
-        /// True if the bytes should be treated as signed, false if they should be treated as unsigned.
+        /// Whether the bytes should be treated as a signed integer. If this is
+        /// `{true}` and the most significant bit is set, the resulting number
+        /// will negative.
         #[named]
         #[default(true)]
         signed: bool,
     ) -> StrResult<i64> {
         let len = bytes.len();
-        if len > 8 {
+        if len == 0 {
+            return Ok(0);
+        } else if len > 8 {
             bail!("too many bytes to convert to a 64 bit number");
         }
 
-        let mut buf = [0u8; 8];
-
-        // `decimal` will hold the part of the buffer that should be filled with the input bytes,
+        // `decimal` will hold the part of the buffer that should be filled with
+        // the input bytes, `rest` will remain as is or be filled with 0xFF for
+        // negative numbers if signed is true.
+        //
         // – big-endian: `decimal` will be the rightmost bytes of the buffer.
         // - little-endian: `decimal` will be the leftmost bytes of the buffer.
-        // `rest` will remain as is or be filled with 0xFF for negative numbers if signed is true.
+        let mut buf = [0u8; 8];
         let (rest, decimal) = match endian {
             Endianness::Big => buf.split_at_mut(8 - len),
             Endianness::Little => {
@@ -258,14 +262,16 @@ impl i64 {
 
         decimal.copy_from_slice(bytes.as_ref());
 
-        let has_most_significant_bit_on = match endian {
-            Endianness::Big => decimal[0] & 0b10000000 != 0,
-            Endianness::Little => decimal[len - 1] & 0b10000000 != 0,
-        };
+        // Perform sign-extension if necessary.
+        if signed {
+            let most_significant_byte = match endian {
+                Endianness::Big => decimal[0],
+                Endianness::Little => decimal[len - 1],
+            };
 
-        if signed && has_most_significant_bit_on {
-            // If the number is negative, fill the remaining bytes with 0xFF (two's complement).
-            rest.fill(0xFF);
+            if most_significant_byte & 0b1000_0000 != 0 {
+                rest.fill(0xFF);
+            }
         }
 
         Ok(match endian {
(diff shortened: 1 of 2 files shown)


MDLC01:
Why using a string to specify endianness instead of a boolean?

istudyatuni:
I think because you can't tell which is used if you don't know function signature

MDLC01:
What about using a named argument? Something like `big-endian`, which is `false` by default (because Wasm uses little endian).
Or maybe having two distinct functions, just like Rust does.

chicoferreira:
I think specifying  `"little"` and `"big"` is way more descriptive than a `true` or `false`. The same can be said for `big-endian: true/false`.

crates/typst/src/foundations/int.rs
```diff
@@ -1,9 +1,8 @@
-use std::num::{NonZeroI64, NonZeroIsize, NonZeroU64, NonZeroUsize, ParseIntError};
-
 use ecow::{eco_format, EcoString};
+use std::num::{NonZeroI64, NonZeroIsize, NonZeroU64, NonZeroUsize, ParseIntError};
```
PgBiel, line 3: i believe this change should be undone

chicoferreira: Done

crates/typst/src/foundations/int.rs
```diff
@@ -145,7 +145,6 @@ impl i64 {
     #[func(title = "Bitwise Left Shift")]
     pub fn bit_lshift(
         self,
-
```
Coekjan, line 148: I think this change (empty line deletion) can be undone. Also on line 171 and line 181 below. You can check them in git diff.

crates/typst/src/foundations/int.rs
```diff
@@ -214,6 +211,50 @@ impl i64 {
...
+    /// ```example
+    /// #int.from-bytes(bytes((0, 0, 0, 0, 0, 0, 0, 1)), "little")
+    /// ```
+    #[func]
+    pub fn from_bytes(
+        /// The bytes that should be converted to an integer.
+        bytes: Bytes,
+        /// Endianness
+        /// - Little endian: the bytes are interpreted from the least significant to the most significant.
+        /// - Big endian: the bytes are interpreted from the most significant to the least significant.
```
Coekjan, line 227: Seems this document does not tell users how to specify the endianness? I would suggest clearly document the acceptable strings here, instead of indicating them in examples.

crates/typst/src/foundations/int.rs
```diff
@@ -222,6 +263,23 @@ impl Repr for i64 {
...
+}
+
+cast! {
+    Endianness,
+    v: Str =>
+        match v.as_str() {
+            "big" => Ok(Self::Big),
+            "little" => Ok(Self::Little),
+            _ => Err("invalid endianness"),
+        }?,
```
Coekjan, line 280: I think we can implement `FromStr` for `Endianness`, and call `Endianness::from_str` here. Using `FromStr` here might be more friendly to developers.

crates/typst/src/foundations/int.rs
```diff
@@ -222,6 +263,23 @@ impl Repr for i64 {
     }
 }
 
+/// The endianness of a bytes to and from int conversion.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+enum Endianness {
```
Coekjan, line 268: I am not sure if we should make this type public.

crates/typst/src/foundations/int.rs
laurmaedje: We should rather use the `Cast` proc macro like for `enum Linebreaks`. Then, the docs will render a [nice table](https://typst.app/docs/reference/model/par/#parameters-linebreaks) completely auotmatically with the doc comments from the enum.

crates/typst/src/foundations/int.rs
PgBiel: I also found it weird at first, but I think other bitwise operation functions don't have this empty line, so this probably makes it more consistent.

MDLC01:
What about providing two different functions?

crates/typst/src/foundations/int.rs
chicoferreira: I will keep it then!

crates/typst/src/foundations/int.rs
chicoferreira: Improved the documentation in the latest commit

crates/typst/src/foundations/int.rs
chicoferreira: I guess it could be, but I am waiting on response if we are also implementing bytes to float conversion. That way it, this enum would be better placed in the Bytes file.

crates/typst/src/foundations/int.rs
chicoferreira: Done in the latest commit. Didn't know that ty!

chicoferreira:
@MDLC01 I think having it in the same function would be better. In `int.from-bytes(b, "big")` vs `int.from-bytes-big(b)` and `1.to-bytes("little")` vs `1.to-bytes-little()`, I think the former feels better. Also, for documentation purposes, having two different functions that do the same effectively (and thus will have the same documentation) would be repetitive. On top of that, as the endianness would be grouped together in a table the differences between the two would be more explicit.

laurmaedje:
I think only supporting 8-byte signed integers is a bit limiting. There wouldn't really be a problem with support for <8 byte integers.

chicoferreira:
@laurmaedje Added support for smaller than 8 bytes integers. Should we make `to-bytes` have an optional argument to specify the target bytes size? 

Something like: `256.to-bytes("big", size: 2)` would return `((1, 0))`

If so, should the result be truncated if the conversion result can't fit the specified size?

chicoferreira:
I've landed the functionality mentioned in the last comment.

Now `#array(256.to-bytes("big", 2))` results in `(1, 0)`. 

The endianness specifies if either left or right should be truncated. It is also possible to make it bigger than 8, which makes those bytes with leading zeros in the left or right depending on the endianness.

crates/typst/src/foundations/int.rs
```diff
@@ -1,9 +1,10 @@
 use std::num::{NonZeroI64, NonZeroIsize, NonZeroU64, NonZeroUsize, ParseIntError};
 
 use ecow::{eco_format, EcoString};
+use typst_macros::Cast;
```
laurmaedje, line 4: please import this from `foundations` instead

crates/typst/src/foundations/int.rs
```diff
@@ -214,6 +212,78 @@ impl i64 {
...
+    #[func]
+    pub fn from_bytes(
+        /// The bytes that should be converted to an integer.
+        bytes: Bytes,
+        /// Endianness of the conversion.
+        endianness: Endianness,
+    ) -> StrResult<i64> {
+        let len = bytes.len();
+        if len > 8 {
+            return Err("Bytes too large to convert to a 64 bit number".into());
```
laurmaedje, line 231:
```suggestion
            bail!("too many bytes to convert to a 64 bit number");
```
(imported from `crate::diag`)

crates/typst/src/foundations/int.rs
laurmaedje: should be public

laurmaedje (commented):
Looks quite good now, though I think the arguments should be named. `size` definitely, but `endian` maybe too (though it would force us to pick a default).

I still wonder how unsigned numbers play into this.

chicoferreira:
Done the suggestions in the latest commit.

> Looks quite good now, though I think the arguments should be named. `size` definitely, but `endian` maybe too (though it would force us to pick a default).

I chose `"little"` as the default as it is the expected endianness of web assembly (https://webassembly.org/docs/portability/).

> I still wonder how unsigned numbers play into this.

Good catch, we might need to add another argument `signed: bool` to the `from-bytes`, but it may be confusing. When converting bytes with 8 bytes there is no problem, because the most significant bit is 1 in the i64. There is a problem when converting smaller than 8 bytes to an integer because we can't assume if the last bit signifies the sign or a decimal part. Also right now, there is no way to generate negative numbers with bytes lower than 8 bytes, it will be treated every time as unsigned, so the sign as an argument must exist in some way.

chicoferreira:
@laurmaedje implemented the signed parameter which will have an effect for bytes smaller than 8 bytes. If the most significant bit is set on those bytes, it will treat the number as negative, filling the remaining bytes with 0xFF to make the two's complement.

Also, as this may be confusing to grasp I've also added some unit tests which should tell the correct behavior of the function. I may also need help documenting the functions better to accommodate these changes.

laurmaedje:
Sounds sensible. I'll need to take a closer look to understand the logic.

laurmaedje:
The logic looks all good. Thanks!

Issue #3696, opened 2024-03-16
Inline math fragments are reversed while text is set to an RTL language


yonatanmgr:
### Description

This source code:
`תהא סדרה $a_n$: $[a_n: 1, 1/2, 1/3, dots]$`
Produces this:
![image](https://github.com/typst/typst/assets/31913495/5826cc1f-c198-4204-bba6-0589840e71ee)

Instead of:
![image](https://github.com/typst/typst/assets/31913495/ddaaf803-3984-41dd-a49d-3f43bc6ab1c8)

My template contains the following set rule: `set text(font: "", lang: "he")`. Changing the language to `"en"` shows the brackets in the correct order (but messes up the rest of the document as it is written in Hebrew).

**This occurred only after the new update.**

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
Oh, that doesn't look good. I think it's a result of https://github.com/typst/typst/pull/2938.

yonatanmgr:
It seems to be not only brackets, but inline math in RTL as a whole:
A paragraph in 0.10.0:
![image](https://github.com/typst/typst/assets/31913495/fd9edd68-4d87-4e65-925f-83e3c2be3d94)

The same paragraph in 11.0.0:
![image](https://github.com/typst/typst/assets/31913495/5e775750-c04c-4278-9781-abf52f8e3727)

laurmaedje:
Yeah, previously an inline equation would be one atomic box. But now, it consists of multiple breakable parts and those are affected by BiDi reordering in unexpected ways. This piece of code might be a temporary workaround:
```typ
#show math.equation.where(block: false): it => box({
  set text(dir: ltr)
  it
})
```
("Might" because I'm not 100% sure how it should really look.)

emilyyyylime:
Generally `show math.equation: set text(lang: "en")` should give correct results

laurmaedje:
I tried that, but without the box, it didn't work.

ocantor:
Is there a way to disable line breaks in inline equations altogether (i.e. restore 0.10.0 behavior)? I personally prefer it over TeX's behavior in RTL documents, at least most of the time.

laurmaedje:
@ocantor `#show math.equation.where(block: false): box` should do the trick

ocantor:
[...]

PR #4491, opened 2024-07-02
Wrap outline entry body in LRE/RLE + make smart quotes ignore directional control characters

### LINKED ISSUE #4476: RTL headings with LTR ending rendered incorrectly in Outline

bluebear94:
Fixes #4476.


crates/typst/src/model/outline.rs
```diff
@@ -500,7 +500,18 @@ impl Show for Packed<OutlineEntry> {
...
+                match TextElem::dir_in(styles) {
+                    crate::layout::Dir::RTL => "\u{202B}",
+                    _ => "\u{202A}",
+                }
+                .into(),
+            )
+            .pack(),
+        );
         seq.push(self.body().clone().linked(Destination::Location(location)));
+        seq.push(TextElem::new("\u{202C}".into()).pack());
```
laurmaedje, line 514: I think this kind of isolation could be moved into a helper function, e.g. `crate::text::isolate`.

laurmaedje: I also wonder whether we want LRE or LRI. Difference:

```typ
#set text(lang: "he")
#let val = "זוהי כותרת ארוכה מאוד שתורגמה על ידי תוכנת מחשב ועל כן עשויה להיות שגיאות דקדוקיות"
\u{202B} #val Pocus \u{202C} #box(width: 1fr, repeat[.]) 1 \
\u{2066} #val Pocus \u{2069} #box(width: 1fr, repeat[.]) 1
```

See also: https://github.com/typst/typst/issues/3696#issuecomment-2115547827

crates/typst/src/layout/inline/collect.rs
```diff
@@ -201,7 +202,7 @@ pub fn collect<'a>(
                 );
                 let peeked = iter.peek().and_then(|(child, _)| {
                     if let Some(elem) = child.to_packed::<TextElem>() {
-                        elem.text().chars().next()
+                        elem.text().chars().find(|c| !is_directional_control(*c))
```
laurmaedje, line 205: could you explain when you use `is_directional_control` over `is_ignorable` / `is_exterior_to_quote`? I don't quite understand the difference.

bluebear94: `is_directional_control` returns true for characters that have no effect on the smart quote algorithm – it’s as if they aren’t there in the first place. This is different from `is_ignorable`, which determines which characters lie immediately outside a quote, hence the renaming.

laurmaedje: Sounds like `is_ignorable` would be a good name for what `is_directional_control` is doing. I wonder whether we can apply the `Default_Ignorable` Unicode property to filter out more stuff. This is already used in paragraph layout.

crates/typst/src/text/smartquote.rs
```diff
@@ -168,10 +168,14 @@ impl Default for SmartQuoter {
     }
 }
 
-fn is_ignorable(c: char) -> bool {
+fn is_exterior_to_quote(c: char) -> bool {
     c.is_whitespace() || is_newline(c)
 }
 
+pub(crate) fn is_ignorable(c: char) -> bool {
+    icu_properties::sets::default_ignorable_code_point().contains(c)
```
laurmaedje, line 176: currently, we don't use the compiled_data from icu4x, but rather generated one (see [here for how it works](https://github.com/typst/typst/blob/59374f737079472aa5694b69ebced8c4cd5f9dfc/crates/typst/src/layout/inline/linebreak.rs#L55-L59)) and [here for the reason](https://github.com/typst/typst/pull/4485#issuecomment-2211732974). This might change in the future. Either we investigate whether we should change that before merging this PR or it should switch to the blob data.

crates/typst/src/model/outline.rs
laurmaedje: So, this leaves this as the only remaining question before merge.

crates/typst/src/text/smartquote.rs
bluebear94: Changed to use the generated `DEFAULT_IGNORABLE_DATA`.

crates/typst/src/model/outline.rs
bluebear94: > ```typst
> #set text(lang: "he")
> #let val = "זוהי כותרת ארוכה מאוד שתורגמה על ידי תוכנת מחשב ועל כן עשויה להיות שגיאות דקדוקיות"
> \u{202B} #val Pocus \u{202C} #box(width: 1fr, repeat[.]) 1 \
> \u{2066} #val Pocus \u{2069} #box(width: 1fr, repeat[.]) 1
> ```

I think you mean `\u{2067}` (assuming you meant to put RLI instead of LRI). With that change, the two lines look identical.

laurmaedje: My bad. But the lines being the same in this case doesn't mean it's always the same, so the question remains.

laurmaedje:
Thanks!

istudyatuni:
Strange, #4476 doesn't exist

PR #4494, opened 2024-07-03
Add missing keys to manifest types

tingerrr:
This PR is the first step towards #4452 and adds the missing keys used in the `typst/package` bundler to the `typst-syntax` manifest types as fully optional. I opted to make `authors`, `license` and `description` optional too, as these are only relevant for the universe itself, but not necessarily for the compiler or other tooling.

I've also added a `unknown_fields` map which collects the field names which were parsed but not expected, this can be used by downstream tooling for diagnostics on misspelled fields. This also means that manifests are no longer `Eq` or `Hash` as these ignored values are not parsed and can't be compared.

Ideally a downstream consumer would be able to actually get values out of these unknown keys to support custom keys and such, but to enable this we'd have to take a hard dependency on `toml` to add its `Table` and actually deserialize the values in question. A similar thing applies to `ToolInfo`, in both cases we expect to only ever parse toml documents and `Tables` can be used to deserialize directly after the fact. See here: https://github.com/tingerrr/typst-project/blob/e19fb3d68b10fce7d2366f4e5969edac6e2f7d34/src/manifest/tool.rs#L11-L39

I think without getting any values out of them, people will end up recreating the manifest structs 1:1 to parse their own tool's config or company internal keys, which is what we're trying to prevent.

### DIFF
--- a/crates/typst-syntax/Cargo.toml
+++ b/crates/typst-syntax/Cargo.toml
@@ -17,14 +17,12 @@ typst-utils = { workspace = true }
 ecow = { workspace = true }
 once_cell = { workspace = true }
 serde = { workspace = true }
+toml = { workspace = true }
 unicode-ident = { workspace = true }
 unicode-math-class = { workspace = true }
 unicode-script = { workspace = true }
 unicode-segmentation = { workspace = true }
 unscanny = { workspace = true }
-
-[dev-dependencies]
-toml = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/typst-syntax/Cargo.toml
+++ b/crates/typst-syntax/Cargo.toml
@@ -23,5 +23,8 @@ unicode-script = { workspace = true }
 unicode-segmentation = { workspace = true }
 unscanny = { workspace = true }
 
+[dev-dependencies]
+toml = { workspace = true }
+
 [lints]
 workspace = true
--- a/crates/typst-syntax/src/package.rs
+++ b/crates/typst-syntax/src/package.rs
@@ -26,19 +26,51 @@ pub struct PackageManifest {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub template: Option<TemplateInfo>,
     /// The tools section for third-party configuration.
-    #[serde(default, skip_serializing)]
-    pub tool: Option<ToolInfo>,
+    #[serde(default)]
+    pub tool: ToolInfo,
     /// All parsed but unknown fields, this can be used for validation.
     #[serde(flatten, skip_serializing)]
     pub unknown_fields: UnknownFields,
 }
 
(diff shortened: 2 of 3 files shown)


crates/typst-syntax/src/package.rs
```diff
@@ -423,4 +482,34 @@ mod tests {
...
+    fn minimal_manifest() {
+        assert!(toml::from_str::<PackageManifest>(
+            r#"
+            [package]
+            name = "package"
+            version = "0.1.0"
+            entrypoint = "src/lib.typ"
+        "#,
+        )
+        .is_ok());
```
laurmaedje, line 496: I think if we're gonna have this test, it should validate that at least some part of the manifest actually parses as intended.

tingerrr:
Do you have any opinion on the stuff mentioned in the comment? While a hard dependency on the TOML format itself would be a bit of a leaky abstraction, it would no doubt be more useful than simply swallowing the values and only collecting the keys.

laurmaedje:
I'm not sure whether we want to encourage people to add extra keys outside of the `tool` section. Isn't that precisely where third parties should put their stuff? This way, we also won't get collision with future compiler or universe keys.

If you agree, this of course still leaves us with the `tool` section. I don't think it would be a big problem to take on a hard dep on `toml`. Unless serde has some way to keep the data around for later deserialization without upfront knowledge of the deserializer?

tingerrr:
> I'm not sure whether we want to encourage people to add extra keys outside of the tool section. Isn't that precisely where third parties should put their stuff? This way, we also won't get collision with future compiler or universe keys.

Yeah, I would only allow this kind of late deserialization for the tool section.

> Unless serde has some way to keep the data around for later deserialization without upfront knowledge of the deserializer?

All mentions of this I found are using `serde_json` as they strictly ask for json examples, and I used `toml` myself in `typst-project`. I don't think there's a sort of "can-hold-anything" intermediate serde format, at least I didn't find any format agnostic crate for this.

laurmaedje:
Okay, we can depend on toml then.

tingerrr:
I've added an example for the tool section in the docs too. I think that should be ready to merge now.

crates/typst-syntax/Cargo.toml
```diff
@@ -17,11 +17,15 @@ typst-utils = { workspace = true }
...
 serde = { workspace = true }
+toml = { workspace = true }
 unicode-ident = { workspace = true }
 unicode-math-class = { workspace = true }
 unicode-script = { workspace = true }
 unicode-segmentation = { workspace = true }
 unscanny = { workspace = true }
 
+[dev-dependencies]
+toml = { workspace = true }
```
laurmaedje, line 28: dev dep can be skipped now I think

laurmaedje:
Thanks!

PR #4496, opened 2024-07-04
Save and restore graphics state for every frame

### LINKED ISSUE #4361: Transparency leak in PDF
### Description

The transparency of text or shapes can leak into subsequent images in PDF as the external graphics state is not reset properly. 

```typ
#rect(fill: red.transparentize(50%))
#image("image.png", width: 45pt) // Image of yellow rectangle
```

| PDF (Chrome) | Preview |
| --- | --- |
| PDF | Preview |

The yellow image inherits the transparency of the red rectangle in PDF but not in the preview.

_Tried with 0.11.1 on the webapp, and on main with f91cad7_

### Reproduction URL

_No response_

### Operating system

Web app, Windows

### Typst version

EpicEricEE:
Fixes #4361

The graphics state was previously only saved and restored when writing groups or images, but not for text and shapes. I don't know if there was a reason behind that, but that made it possible for the transparency state to leak outside the current text/shape writing.

I hope this doesn't break anything, PDF tests would be useful here😅


laurmaedje:
Saving and restoring around every single text item seems a bit wasteful in terms of the generated output. A slightly more surgical approach might be nice.

EpicEricEE:
There would of course be the simple option of just explicitly setting the opacity to fully opaque when writing images

laurmaedje:
All other elements already handle it? If yes, sounds good. And we can skip internally in Builder writing a fully opaque state if it's already opaque (maybe that happens naturally)?

EpicEricEE:
Color glyphs also don't handle it, but the result depends on the viewer and probably on the font (the red color in vscode-pdf seems to be an unrelated issue):

| Edge (Acrobat) | Chrome | vscode-pdf (PDF.js) |
| --- | --- | --- |
| image | image | image |

```typ
#text(fill: red.transparentize(80%))[ASDF]#emoji.face
#text(fill: black.transparentize(50%))[ASDF]#emoji.face
#text(fill: black)[ASDF]#emoji.face
```

When setting the graphics state explicitly for them too, it works. However, when using a very outdated version of acrobat pro that can't even render color glyphs, it becomes visible that the fill color also leaked outside the text element:

image

This probably shouldn't matter much though, as color glyphs normally don't react to the fill color.
The other frame types are handled though:

- Shapes and normal text set their own opacity with custom fill/stroke
- Links and Tags don't write any graphics so they don't need to do anything
- Groups are just other frames of the above types

And yes, there already is a check in `set_external_graphics_state` so that it is only written if it differs from the previous one.

laurmaedje:
Looks like `write_color_glyphs` should also `set_opacities`.

I wonder whether we can change `external_graphics_state: Option<ExtGState>` to `external_graphics_state: ExtGState` since it should have a clear default.

An alternative design would be that we do rely on save/restore, but only do it if actually necessary because of transparency.

laurmaedje:
The test `show-selector-element-or-label` among a few others seems to have regressed.

EpicEricEE:
It seems like the `external_graphics_state` field of the builder struct was never actually updated, but as it was always None, it didn't have much of an impact before.

laurmaedje:
Thanks! I added your snippet as a test. The PNG test doesn't validate it, but at least we can discover regressions on the PDF output with a PDF comparison script. And if we add PDF tests down the road, it's good to already have it in the test suite.

PR #2378, opened 2023-10-12
Add `csch` and `sech` operators

Weissnix4711:
Adds the following maths operators:
- sech
- csch
- arsinh
- arcosh
- artanh

And the calc functions:
- asinh
- acosh
- atanh

PR #1090, opened 2023-05-03
Add more math operators

StrangeGirlMurph:
Added some missing ops.

`id`: identity (very important)
`im`: image (if there is `ker` we also need `im`; also doesn't collide with `Im`)
`tr`: trace (also relevant for linear algebra and kind of in the same class as `det` and `dim`) 
`cl`: closure (important in topology and analysis)
`int`: interior (important in topology and analysis)

PR #4504, opened 2024-07-05
Add arctg and arcctg predefined operators

blt-r:
We already have tangent and cotangent in the tg/ctg notation, and we have inverse trig functions with arc prefix, so it would make sense to have arctg and arcctg.

These functions are commonly used where I am from.

There are more trig function with missing inverse versions (cot, sec, csc), so maybe it would make sense to add them too, but I don't know how often they are used if at all.

### DIFF
--- a/crates/typst/src/math/op.rs
+++ b/crates/typst/src/math/op.rs
@@ -17,7 +17,7 @@ use crate::text::TextElem;
 /// ```
 ///
 /// # Predefined Operators { #predefined }
-/// Typst predefines the operators `arccos`, `arcsin`, `arctan`, `arg`, `cos`,
+/// Typst predefines the operators `arccos`, `arcctg`, `arcsin`, `arctan`, `arctg`, `arg`, `cos`,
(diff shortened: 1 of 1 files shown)


Enivex:
> Prior discussions about adding operators: https://github.com/typst/typst/pull/1090, https://github.com/typst/typst/pull/2378.
> 
> I've honestly lost track of what we decided should and shouldn't be included at this point.
> 
> cc: @Enivex 

My stance is that, if anything, I'd remove `tg` and `ctg`, because `tan` and `cot` are way more common. It's super easy to define your own operators anyway.

If we _have_ to add something, I'd argue that it should be only `arctg` and not `arcctg`

(Note that LaTeX doesn't predefine any of the `tg` variants either.)

laurmaedje:
I'll defer to @Enivex's judgement on this one, so I'll close this. Thanks still!

Regarding `tg` and `ctg`. I think those have been around forever from the original list I took from LaTeX. I don't have any particular attachment to keeping them around.

Issue #4407, opened 2024-06-16
Adding invisible data (`metadata()`) after setting `page.paper` inside `content` (sub-document) creates an empty page


Andrew15-5:
### Description

```typ
// 2 pages
// #set page(paper: "a5") // Doesn't matter
#metadata("") <begin-marker>
#[
  #set page(paper: "a5")
  Text
]
#metadata("") <end-marker>

// 1 page
// #set page(paper: "a5") // Doesn't matter
#metadata("") <begin-marker>
#set page(paper: "a5")
Text
#metadata("") <end-marker>
```

The first block creates one page with text and an empty page. The second doesn't add an unnecessary page at the end.

I am creating a bundle document that `include`s several other documents. This way I can have all reference links working and have an outline of all sub-document (with working links). But including a full document means including set/show rules, which creates a `styled()` content block/wrapper for each sub-document.

<details><summary>General AST structure</summary>

```rs
...
    [(
      metadata(value: ""), <- "sub-document starts" marker
      [ ],
      [ ],
      styled(child: [([Text], [ ])], ..), <- sub-document
      metadata(value: ""), <- "sub-document ends" marker
      [ ],
    )],
...
```

</details>

It would be normal in _some_ cases if I replace seconds `metadata()` call with visible content (text). But still, maybe you want the text from outer (bundle document) be inserted into the same last sub-document's page (where there is still a lot of empty space) and not into a whole new page. This aspect can be potentially a new feature, but I don't know how exactly this can be controlled (maybe a new `page.append-parbreak-if-inside-content`).

I would like that _at least_ an invisible data such as `metadata()` (or `hide()`?) won't create a new empty page as this is totally not a desirable behavior. But I have a feeling that this is impossible with the current (v0.11.0) design (maybe?). Or maybe a new check can be added that will "diff" the previous (bundle/global) `set page()` and the current (sub-document/nested) `set page()`. If they are identical then the last one will be removed and this way no new empty page at the end will appear. IMO, this is an easier route. But what if `page.footer` don't much?

To fix this, I can remove the sub-document's `set page()` by using a condition, but that condition would require a `context` so that I can `query()` the start/end sub-document labels/markers to see if the current document is inside a bundle document or not. At which point it will become #2326.

Similar but still different problem exists with `set document` (instead of `set page`) which is inside a context (to also check for `is-sub-document()`): #4065.

A general "embed sub-document" issue: #3723.

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

[...]

Issue #4065, opened 2024-05-03
Can't set document properties based on context


frozolotl:
### Description

There currently is no way to set any document properties using context-dependent functions.
If one tries to define a show-rule as follows, the PDF export will not have a title set. Using `context document.title` does still output the set title, though.

```typ
// NOTE: You probably don't want to actually do such a `show: body => context …` construction.
#show: body => context {
  set document(title: [Title]) if query(<my-label>).len() == 0
  body
}
#context document.title
// uncomment if desired: <my-label>
```

### Proposal
There should probably be a way to define document properties based on context.

If that is not feasible, then `context set document()` should throw a compiler error.

### Typst version

- [X] I am using the latest version of Typst

Andrew15-5:
Here is the minified version:

```typ
#let template(doc) = {
  // context if state("bundle").get() != true {
  context if query(<sub-document-begin>).len() == 0 {
    set document(title: [Title])
  }
  // set ...
  // set ...
  // set ...
  // show ...
  // show ...
  // show ...
  doc
}
#show: template
Text
```

Andrew15-5:
Similar issue: https://github.com/typst/typst/issues/2326.
(`set page()` inside `context` vs. `set document()` (inside `if` statement that is) inside `context`)

Issue #2326, opened 2023-10-06
`set page` inside `context` appends empty page


EpicEricEE:
### Description

I'm trying to adjust the page margin depending on the maximum heading numbering width, so I need to use `#set page` inside `locate`.

```typ
#set page(numbering: "1", width: 6cm, height: 2.5cm, margin: 2em)

First Page

#locate(loc => [
  #set page(numbering: "i")
  Second Page

  Third Page
])
```

![image](https://github.com/typst/typst/assets/7191192/c98ea92e-97fb-4f0b-93e6-e9b4616c76d7)

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

alexanderwiller:
I am trying something similar, I want both different header content and different margins depending on if I'm on the first page or not.

First, I did this to achieve a different header between the first and following pages:

```
set page(paper: "a4",
  header: locate(loc => {
    if counter(page).at(loc).first() == 1 { header_1stpage } else { header }
  }),
)
```

That approach worked, but I wanted to have margins that differ per-page too, so I tried to make the margin dynamic based on the page too:

```
set page(paper: "a4",
  margin: locate(loc => {
    if counter(page).at(loc).first() == 1 { (left: 25mm, top: 98.46mm, rest: 20mm) } else { (left: 25mm, top: 40mm, rest: 20mm) }
  }),
  header: locate(loc => {
    if counter(page).at(loc).first() == 1 { header_1stpage } else { header }
  }),
)
```

That approach did not work, because `locate` does return content (related to https://github.com/typst/typst/issues/1613 I think), not the type of what is contained inside the locate (dict of margins).

In general, repeating the `locate` seems quite verbose, so I tried something different:

```
set page(paper: "a4", margin: (left: 25mm, rest: 20mm))

locate(loc => {if counter(page).at(loc).first() == 1 {
  set page(header: header_1stpage, margin: (top: 98.46mm))
} else {
  set page(header: header, margin: (top: 40mm))
}})
```

While I really like the cleanliness, this approach works only partially:
- The first page header and top margin is set according to the rule
- There is no page body on the first page, it starts only on page two
- Neither second page header nor top margin are set according to the rule

I created a complete example that shows the situation: https://typst.app/project/rjX6qQRpfnYDu9HZJIoK4l

[...]

PR #4513, opened 2024-07-06
[PoC/WIP] Added `entrypoint()` function

### LINKED ISSUE #4502: Entry point (main file) detection within Typst document for nested projects
### Description

I have a few standalone documents that are self-contained:

sub-document

But I also want to join them into a bundle while preserving their internal look and structure, plus adding an outline and other inter-document links/references so that I can quickly jump to any of the sub-document:

bundle document

The core problem now is less flexible project configuration in the web app:

```typst
#let is-bundle-document = false
// #let is-bundle-document = true

...

#let settings(doc, n: auto) = {
  show: set-rules
  show: show-rules

  show: it => {
    import "info.typ": is-bundle-document
    if is-bundle-document { return it }
    if n == auto { return it }
    let n = if n == auto { n } else { n.find(regex("\d+")) }
    show: set-document.with(n: n)
    show: set-page
    set-bibliography
    if n != auto { state("lab-number").update(n) }
    it
  }

  show: turn-on-first-line-indentation
  doc
}
```

I have to manually toggle the `is-bundle-document` boolean to be able to successfully compile either bundle or any of the sub-documents. And this problem exist only in the web app. Locally I have this:

```typst
#let settings(doc) = {
  show: set-rules
  show: show-rules

  show: it => {
    import "stuff.typ": get-lab-number
    // FS hack to successfully compile both bundle and individual documents.
    let is-bundle-document = get-lab-number() == none
    if is-bundle-document { return it }
    show: set-document
    show: set-page
    set-bibliography
    it
  }

  show: turn-on-first-line-indentation
  doc
}
```

where `get-lab-number()` is just a wrapper for `read("/src/lab-number").find(regex("\d+"))`. Here I utilize the "feature" of absolute file paths as `"/src/lab-number"` for bundle document and for a sub-document mean/point to different files! All because the `TYPST_ROOT` or simply `PWD` are different. So with "realpath" they become `project/src/lab-number` and `project/src/sub1/src/lab-number` respectively. And since they (files) can hold different content, I can just `read()` it without any context and instantly know what is the current "entry point" or which document is being compiled: a specific sub-document or the whole bundle. _This one simple trick_ automates the whole process, so I don't have to touch the content of any document whatsoever and just execute `typst c` at the right dir or with the right `root` specified. Pretty neat, huh?

But in order to use the same "global settings" in all (sub-)documents I had to symlink stuff from `project/global/` to `project/src/sub1/global/` (because the root changes and this way global stuff will be immune to it, i.e., to the root change).

Why not just use `context` and `query` or whatever? The main reasons why it should be avoided:
1. `read()` must receive `str` (not `content`/`context()`);
2. `context` conditions are not allowed with set/show rule, at least generally speaking (here are issues for the [`page`](https://github.com/typst/typst/issues/4407) (https://github.com/typst/typst/issues/2326) and [`document`](https://github.com/typst/typst/issues/4065#issuecomment-2093864598));
3. global show rule that inserts indentations in correct spots throughout the document will break if the whole sub-document is wrapped in a context.

In the web app, you can't configure `TYPST_ROOT` or use symlinks. So something that will provide some form of identification of which "main file" or which "document" is being watched/compiled to the web is necessary.

The easiest (maybe not) solution is to add something like `if /..../file.typ is main then TYPST_ROOT=/..../ else if ...` maybe in `typst.toml` or just in the web project settings. Or add `get-main-file-path()` which would return a string with absolute path to the main file (without context). Or something else. Adding symlinks would also be nice (depending on the solution).

This is probably a niche request, but this "configurability" of the project should be a general issue, a thing that can and should be improved. So that web app users can have a bit more control, making it more competitive to the local/CLI approach.

But for now, I literally thought of all sorts of crazy ideas, but none of them would work, so I have to use a simple global boolean variable. Toggling it depending on which document is being worked on. This is actually what I come up with initially when building locally, but then I realized that it can be fully automated. Unlike the web.

Related: https://github.com/typst/typst/issues/3723.

### Use Case

.

Andrew15-5:
Here is my minimal proof of concept implementation (WIP).

I have no idea about how performant the solution is or whether it added the file (`entrypoint.rs`) in the right spot. I also modified the `World` trait by adding a new function. It only was implemented for the `SystemWorld` struct and I don't know if other "worlds" exist. For those it will be a breaking change.

Currently the function doesn't have any arguments, but I definitely will add a few. Namely: return entry point dir instead of a file (file's parent), return just a file name without absolute path (maybe something else too).

There is a lot to discuss (well, maybe not _that_ much). For example, I don't think there is a single standard function that has 2 words, so I used `entrypoint` instead of `entry-point` (kinda inspired by Dockerfile's `ENTRYPOINT`). Also which arguments other users would use/like to have. The security aspect (although I think everything is good here). Documentation, tests.

The discussion can also be held [here](https://discord.com/channels/1054443721975922748/1258375840933216321). Adjacent conversion was [here](https://discord.com/channels/1054443721975922748/1259234897935994900).

CC @laurmaedje

Closes #4502.

P.S. I added `bundle` dir and `justfile` to see what I want in practice. Use `just`/`just sub 1` to build bundle or 1st sub- document. Use `just see` to open PDF (using `nohup` and `xdg-open`).


Andrew15-5:
`cr -- c --root bundle ./bundle/src/main.typ out.pdf`:
<details><summary>screenshot</summary>

bundle

</details>

`cr -- c --root bundle ./bundle/src/sub1/src/main.typ out.pdf`:
<details><summary>screenshot</summary>

sub1

</details>

`cr -- c --root bundle ./bundle/src/sub2/src/main.typ out.pdf`:
<details><summary>screenshot</summary>

sub2

</details>

Andrew15-5:
In https://github.com/typst/typst/pull/4513/commits/83884f1049f87764d165a90f1fe057e0066e3e11 changed to opt-in approach in the `Source` struct. By default, the main is never `stdin`. But this can be changed separately. So now the footprint of changes is much less. I assume, that for the web app nothing would have to be changed, since it (should) never uses `stdin`.

Andrew15-5:
https://github.com/typst/typst/pull/4513/commits/841bb95cff3567827526ed9099f2912fd4a806f4 Adds `raw` named argument (default `false`), which will make the function return `bytes` type instead of `str`. All because the `str` is lossy, since it's converted from `Path`/`OsStr`.

<details><summary>Screenshots</summary>

utf-8non-utf-8


</details>

laurmaedje:
I'm not really convinced of adding this. Independently of the str/bytes thing (which seems similar to the problem [here](https://github.com/typst/typst/pull/4380#issuecomment-2212390468)), it generally feels like a sort of ad-hoc solution. I think context is the "right thing" here since the question is literally "is this subdocument built in the context of another document". Putting large parts of the document into context currently has adverse effects, but I think we can fix that.

Andrew15-5:
> it generally feels like a sort of ad-hoc solution.

Well, it kinda is. But also it can be improved later and any point, for example, when `path` type or whatever will be added. If I could instead make a plugin for this, I would, but currently the only way is to modify the compiler.

> I think context is the "right thing" here since the question is literally "is this subdocument built in the context of another document".

You are ignoring the fact that the "context" wraps its content in `context()` element which in turn breaks **a lot** of stuff, practically making more problems than it _tries_ to solve. If it didn't do it, then I think I would definitely make it work without any `entrypoint`s.

> Putting large parts of the document into context currently has adverse effects, but I think we can fix that.

How exactly?

`context` currently creates several problems (idk how to divide them):
1. it will break `show: turn-on-first-line-indentation` unless I can set it inside/before(?) the context wrapper;
2. using it like `context if is-sub-document()` would break `show: set-document` because it...either wouldn't apply at all, or give an error that `document` must be set before any `content` (https://github.com/typst/typst/issues/4065#issuecomment-2093864598);
3. setting `page` inside context (or something else?) also has a bug: #2326 / #4407;
4. very brittle `show` rules because (1) it's very easy to lose (not apply) nested things (show rule or `panic()` etc.), (2) we don't yet have conditional show rules.

In my symlink/different-root-dir local hack, the `show: settings` already looks hacky and "barely gets the job done":

image

I would love to hear how we can get around the current `context` restrictions.

laurmaedje:
> You are ignoring the fact that the "context" wraps its content in context() element which in turn breaks a lot of stuff.

I'm not, that's why I said that it has adverse effects.

> How exactly?

You pretty much listed the problems right below.

Andrew15-5:
> You pretty much listed the problems right below.

The issue with those is that (1) there are a lot of problems (bugs **and** not yet implemented features), (2) some or all of them depend on some underlying limitation, i.e., the full/`.flatten()` list of problems is even longer, (3) resolving **all** of them requires a huge amount of work and time.

We already had ad-hoc solution and just things that are now deprecated. One good example that comes to mind is `locate()` and `loc` argument to `query`. It's OK to have stuff that are not best for a while until a better solution is added. Typst is in beta for a reason. The `read-dir()` PR you've mentioned is targeted at a general use that will 100% be better with a `path` type added, though I would argue that it would be nice to have it now in its current implementation, rather than (again) wait for a better solution (that has unknown ETA). The `entrypoint()` on the other hand has a more narrow usage, so you wouldn't gain much by waiting for either `path` type (and/)or all other mentioned issues to be resolved. In the nested example project I added, it is strictly/mainly used here:

https://github.com/typst/typst/blob/1f5008881f76bc75d9042e9eefca4bece77ea0dc/bundle/global/settings.typ#L33

Which is to have that sweet-sweet ability to differentiate between which document is being compiled (if all of them have the same root dir, i.e., like in the web app). And a simple hard-codded path string is all it needs.

I of course can squeeze out even more out of it by using it something like this:

https://github.com/typst/typst/blob/1f5008881f76bc75d9042e9eefca4bece77ea0dc/bundle/global/settings.typ#L1-L4

This way, I can further automate things like "which project is used exactly" (by using the unique path value) and "figuring out the correct readable file path" (the problem with `read()` and paths like `/src/sub1/src/code/file.sh`).

I don't see a problem providing a temporary (but also "more than enough") feature solution to "give time for a better solution to emerge". IMO, the scale "wait for a very long time and suffer" vs. "have an OK solution and maybe a better one will appear at some point" says that this `entrypoint()` PR will result in "net gain" (or more gain) than the other way around.

mattfbacon:
> It's OK to have stuff that are not best for a while until a better solution is added. 

I think you're right, as long as there are enough users and use cases. In this case it seems like it's just you. I can't think of any case where I would use this and honestly I don't even understand your use-case—it just seems like a big XY problem to me.

Andrew15-5:
It's "just me" because we don't have a ton of users and most of them don't do nested documents. I'm just a pioneer in this type of projects (in fact, there is a chance that I have to make another nested project soon). I'm sure if we had a lot more people using Typst, I wouldn't be the only one. But still, such type of projects isn't your normal/typical "thesis" or "study" or "assignment" or whatever else is very common, so obviously it (`entrypoint()`) wouldn't be popular or used by many. More like "used by a few".

But nevertheless, Typst is supposed to be a tool to make creative documents, i.e., not limit what a user can do/create. So adding it "in the meantime" would be really nice. After an alternative solution appears (probably should be a **better** solution), this function (if the solution is completely different) or its behavior (if, e.g., `path` type is added) can be simply deprecated and removed after a few versions. Since the same thing already happened with other parts of Typst, it shouldn't be a big deal to do it again. We still have a long way to go until v1.0.0.

I faced a lot of major problems/roadblocks before, but all of them can be solved without changing the compiler (even if it means creating a bunch of huge `show` rules etc.). And in this case...well, you already know.

Andrew15-5:
> it just seems like a big XY problem to me.

What does this mean?

> I don't even understand your use-case

Simply put (talking about the web app): I need to have a few separate documents, but they all share the same global settings. To keep those settings synced across all those separate documents, I either have to use symlinks (not possible) or an external sync software (not possible) or put them all in the same root dir (possible). But I also have to create a document that will glue them all together (with an additional top-level outline that lists those document, but not their content). So, I have to also put the "bundling" version of the document in the same root dir as all separate documents. And now we end up with the problem of having the same rood dir but different entry points, a.k.a. main/input files. `entrypoint()` is the only (non-context) solution that can solve the issue.

All other solutions either make sub-document unmaintainable or just harder to read/edit + looks unnatural (like   [the `#import` hack](https://discord.com/channels/1054443721975922748/1258375840933216321/1259771966798299198)), or completely broken (like any use of `context`).

Andrew15-5:
> 1. That's because `turn-on-first-line-indentation` itself is a hack since it depends on inspecting content recursively, which is fundamentally problematic

Exactly. But it's the only way to not add any `#h(...)` at the start of almost every single paragraph, and it is the only solution that won't add vertical spacing. So it's "readable noise-free document with a huge hack in the back" or "no hacks and an ugly/repeatable/noisy snippet throughout the document". _It's what it's_.

> 1. Originally, I wanted to make content completely opaque for this reason. I ended up opening it up, at the cost of some hacky/easy-to-break things becoming possible.

Well, in the end it's a good thing that you did, because without `show` rule hacks you wouldn't be able to create A LOT of documents (if they have specific visual guidelines). I can't imagine my life without this "feature" in the current iteration of Typst.

> 3. in fact, 2 and 3 are closely related.

I know.

> 2. That's fixable.

Would this be a general "make context transparent, maybe just a little" fix, or a specific "allow applying `set document` from within `context` and/or `if`"?

> 3. That's also fixable

Is there an ETA for 2/3? Or do they have the lowest priority? Do they have any dependency issues that have to be solved first (some layout or context logic etc.)?

> 4. I don't really get what you mean here.

I don't have an example, but I faced this problem before, where, e.g., `panic()` wouldn't be called even when it should be, judging by the source code. So it can "get lost" and the document will be compiled like it was nothing.

> 4. Doesn't seem like a blocker for what you want to do.

It isn't, but this list isn't about that specifically.

The 5 would be the `#bibliography` issue. I didn't actually tested this (or I don't remember), but can I do `context if ... { bibliography() }`? From the screenshot you can see that I basically isolated the thing that currently have some issues: `set document()`, `set page()`, can't call the exact same `bibliography()` twice. And it's all wrapped into the ad-hoc of the `show-if` rule (which should have an open issue?). Fixing everything in this block would be an "alternative solution" to the `entrypoint()`. At least I hope so, this needs to be tested to know for sure.


> > We already had ad-hoc solution and just things that are now deprecated. One good example that comes to mind is locate() and loc argument to query.
> 
> That's different because when `locate` was added, there wasn't yet a better solution. In contrast, here, the better solution already exists (context), it just has a few bugs that need ironing out.

You can't really call it a solution if it has "a few bugs" that preventing me from using said "solution".

> Moreover, `entrypoint` adds more capability than is necessary to fix your use case, and in my opinion a bad/easy to abuse capability.

Since I used it in 2 ways and both are a perfect fit for fixing the problem, I don't see how it adds more capability than is necessary to fix my use case. Abuse? I think it's useless in 99% of other documents. If it will be used in a different way, then it means there is no better solution for that problem. Or, of course, because a beginner doesn't know "de (right) wae".

> I don't want people to start relying on it, making it hard to take away.

Well, it depends on the next solution. It can be an upgraded version of `entrypoint()` or, well, fixed all the mentioned issues. So in some cases, I agree. But with the help of the "deprecated" label, it shouldn't be hard to switch to an alternative as happend(-ing) with `locate()` and the deprecated types' comparison.

> All `locate` use cases were completely covered by context, so we took nothing away there.

Yeah, the new solution isn't much different, generally speaking.

Andrew15-5:
Oh, actually, there is another problem:

```typ
#let get-number() = {
  let file = "/" + entrypoint().split("/").slice(0, -1).join("/") + "/number"
  read(file).find(regex("\d+"))
}

#let set-document(doc, number: auto) = {
  set document(title: [Document #if number == auto {
      "№ " + get-number()
    } else {
      if type(number) != int { "№" } + "№ " + number
    }])
  doc
}
```

Which is basically `set document(title: entrypoint())` with extra steps. Which will have to be translated to either `context set document(title: state().get())` or `set document(title: context state().get())`. Which can be the same as (2) or a completely separate problem. What do you think?

laurmaedje:
> Would this be a general "make context transparent, maybe just a little" fix, or a specific "allow applying set document from within context and/or if"?

It would be a fix for set document being brittle. Context would remain opaque, that's a fundamental thing.

> Is there an ETA for 2/3? Or do they have the lowest priority? Do they have any dependency issues that have to be solved first (some layout or context logic etc.)?

No ETA, but also not low priority.

> You can't really call it a solution if it has "a few bugs" that preventing me from using said "solution".

The bugs may make it unusable for your case and that is unfortunate, but that doesn't stop it from being the conceptual solution. That's just how bugs work --- sometimes they break your use case.

In summary: I tried to argue my position objectively, but in the end it's also in part a gut feeling that this feature is not right for Typst. My experience has been that, with language design, the gut feeling often turns out to be right down the road. I'm sorry that your use case is currently blocked by bugs. :/

Andrew15-5:
> It would be a fix for set document being brittle. Context would remain opaque, that's a fundamental thing.

Can you define "being brittle" in terms of syntax/behavior? Which aspects of that will change after the fix?

laurmaedje:
Brittle in the sense that not even invisible metadata may appear before it. Just like how the invisible metadata generated by context breaks set page.

laurmaedje:
> So, the context set document() will in fact work after the fix, correct?

That's the plan.

> But if context creates a visible wrapper in the AST now, how would this set rule be applied to the outer scope, outside the context wrapper, i.e., to the whole document? Will this specific set rule be able to "penetrate" the "opaqueness" of the context? Because this would be a completely new concept, I think.

The set document check doesn't happen at the AST or eval level. It happens during realization where context is resolved. It's not a problem.

> Which one will be a working example: context if ... { set document() }, context set document() if ...? Both?

The second. The first only if the nothing follows after the if (because set document must apply to everything). Just like without `context`.

Andrew15-5:
> The first only if the nothing follows after the if

I didn't get this part. Can you illustrate what you mean by this? E.g., one example when "the nothing" follows "the if", and another one when it doesn't.

laurmaedje:
This would be ok:
```typ
#show: it => context if condition {
  set document(..)
  it
} else {
  it
}

Text here
```

This doesn't make sense because the set rule is scoped:
```typ
#context if condition {
  set document(..)
}

Text here
```

Andrew15-5:
Got it.

Andrew15-5:
It's [the same](https://github.com/typst/typst/issues/4502#issuecomment-2608448298) for the web app, nothing has changed. For the CLI, you can use [the same symlink + root hack](https://github.com/typst/typst/issues/4502#issue-2391472203).

Issue #1622, opened 2023-06-30
Typst don't apply `footnote` style when `text` is called before hand


FirelightFlagboy:
### Description

Typst version: `0.5.0 (3a8b9cce)`

I'm trying to configure the style of the generate footnote, but it don't seems to work when I call `text` to set the style of a specific block of text.

Here is the minimal input where I'm able to have the issue.

```tyspt
#text(weight: "bold", size: 1.5em)[Styled title]

#show footnote.entry: set text(blue)
#show footnote: set text(red)

== Hello world

This is a simple String with a footnote#footnote[A footnote is a text that is put at the bottom of a page describing an element]
```

I expected the footnote text to be `blue` but it's not.

> The issue is gone when I remove the first line with `#text...`

Result in the following [test.pdf](https://github.com/typst/typst/files/11915114/test.pdf)

<details>

<summary>Generated Image</summary>

![test-1](https://github.com/typst/typst/assets/30697622/aaa7c720-b1f0-4d13-a775-aa7bc4b5ca06)

</details>

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
Duplicate of <https://github.com/typst/typst/issues/1467>

FirelightFlagboy:
@laurmaedje By looking at the documentation <https://typst.app/docs/reference/meta/footnote/>

The information that the `#show footnote.entry` need to be at the top of the document is missing. (https://github.com/typst/typst/issues/1348#issuecomment-1566316463)

Could it be added ?

PR #4516, opened 2024-07-07
Line numbers

### LINKED ISSUE #352: Line Numbers
Typst seems like a good potential choice for people in the legal community to be able to write court filings. However, some courts require that legal documents include line numbers in the left-hand column. LaTeX has `lineno` for this, but I don't see the ability to do it with Typst. If a developer needs some examples of courts' local rules that shows the format I'd be happy to post them.

PgBiel:
Closes #352

Adds line numbers.

~~The feature is still WIP (see the TODO at the bottom). Opening this PR to share a bit of the initial progress for now.~~

## Overview

- You will be able to enable (or locally disable/customize) line numbers with a set rule, such as `#set par.line(numbering: "1")` (bikesheddable).
- You'd also be able to control alignment of line numbers, spacing between line numbers and the text, and so on.
- In principle, the line numbers go in the left or right margin. This should also be configurable.

Here's what we have so far with the current code: the Typst code below

```typ
#set par.line(numbering: "1")
#show figure: set par.line(numbering: none)

#align(center, text(52pt)[Lorem Ipsum])

#text(32pt, grid(
  columns: (1fr,)*3,
  [Some\ Author],
  [Authorus\ Johnson],
  [Authorus\ Mister\ Johnson]
))

#lorem(100)

#figure[
  No line numbers here\
  No line numbers here
]

#lorem(100)
```

produces (**note that this isn't final**)

sample

### TL;DR

We basically perform a "manual query" of line elements while laying out the root flow of the document, by recursively searching inside frames and subframes produced by layout. On each searched subframe, we add its relative `y` to a counter. When we find the line, we save the final value of the counter, and thus we will have the height of the line relative to the parent frame. We then lay out `counter(par.line).update() + counter(par.line).display()` on the left margin at each detected line number height (we have the relative heights to each parent frame, so we lay out each parent frame and then add the parent frame's final height to the line numbers' relative heights).

This "manual query" mechanism could potentially be reused for other features, such as margin notes, paragraph numbers, and so on. Maybe even some more general mechanism of absolute alignment between elements (e.g. easily place an element at the same height as another). But we'd need more exploration to know.

We also synthesize the `par.line` set rules in the queried line elements, so that you can `#set par.line(numbering: ...)` somewhere deep within your document in order to e.g. have some lines not be numbered (by setting numbering to `none`), or have them produce different numbering from the rest.

A similar mechanism could be potentially used to enable local styling of footnotes (see https://github.com/typst/typst/issues/1622), but some design discussion would be needed to determine whether that'd be desirable behavior.

### Details

1. There are two new elements: `ParLine` (exposed as `par.line` to the user), which is only used for styles (and thus isn't `show`able), and `ParLineMarker`, which isn't exposed to the user and is only used for `Show`.
      - In principle, we'd ideally only have one, but I didn't want the user to be able to `#show par.line` as that'd result in some unexpected behavior, since we don't wrap entire lines in `ParLine`, but rather just add a marker at the start of each line, so it wouldn't be replacing anything. But we can change this (and it's easier to replace 2 elements with 1 than go the other way).
2. Within `commit` in inline paragraph layout, we add a `ParLineMarker` to the start of each line's frame.
      - It saves the current value of `par.line.numbering`. If it is `none`, the par line marker is not placed, and the line isn't numbered.
3. Within a flow (`flow.rs`):
    1. Created a new flow item, `LineNumber`. It holds a frame (counter update + counter display) and its height (relative to the top-level flow element the referenced line is contained in). As an invariant, it must always appear in the `flow.items` vector _right after the parent of its referenced line_ so that we may correctly determine its final `y` position in `finish_region` (by just adding parent's final `y` position to the relative height of the line number).
    2. Created `collect_par_lines`, which searches for a tag element containing a par line marker. It has a counter (relative height of the searched line) which starts at zero. On each searched subframe, its relative `y` position is added to the counter. When we find the tag, we add its own `y` position to the counter, and that becomes the final relative height of the searched line (and thus of the line number) within the frame we were initially searching for lines.
    3. Created `handle_par_lines`, which takes a vector of searched lines and their relative heights and adds the corresponding line numbers into the flow as flow items. Must be called **right after** the parent was pushed as a flow item. This function has a **deduplication procedure** (since many lines may have the same height) which basically sorts the found lines by relative height and then excludes from numbering any lines that have less than 1pt of height difference from a numbered line.
        - There is a similar deduplication procedure in `finish_region` later, so I'm probably gonna be simplifying this still.
        - Added calls to this function within `handle_item`, `layout_block` etc.
     4. Within `finish_region`, we save the height of each flow item we push to the output frame, and then, as soon as we find line numbers, we add their relative heights to the latest saved height, as that was the height of the corresponding lines' parent frame. Then we push the line numbers to the output frame, with another deduplication mechanism since the initial deduplication procedure only runs within each parent frame, but not between frames.
        - This still needs some fixing since line numbers may be placed out of order due to `#place`.
      5. Thus, we have line numbers.

## TODO

- [x] Fix line number ordering with `#place`
- [x] Adjust to differing line heights
- [x] Add line number alignment option (default to right alignment)
- [x] Add line number spacing option
- [x] Allow picking which margin (left or right, start or end) the line numbers appear in
- [x] Add proper support for line numbers in 2-column documents
- [x] Add tests
- [x] Per-page number reset
- [x] Auto line number alignment based on (opposite to) margin
- [x] Prohibit intersecting line numbers (to our best ability)
- [x] Cleanup


crates/typst/src/model/par.rs
```diff
@@ -224,3 +231,62 @@ pub enum Linebreaks {
 pub struct ParbreakElem {}
 
 impl Unlabellable for Packed<ParbreakElem> {}
+
+/// A paragraph line.
+/// This element is exclusively used for the line number counter, and cannot
+/// be placed.
```
bluebear94, line 237:
```suggestion
/// This element is exclusively used for the line number counter and cannot
/// be placed.
```

Dherse:
That's a surprisingly small implementation, awesome 🎉

laurmaedje:
Could we do the frame search fully in `finish_region` and do nothing before? That would simplify things quite a lot.

crates/typst/src/layout/flow.rs
```diff
@@ -5,18 +5,23 @@
 //! inline-level layoutable elements.
 
 use std::fmt::{self, Debug, Formatter};
+use std::num::NonZeroUsize;
+
+use typst_syntax::Span;
+use typst_utils::NonZeroExt;
```
laurmaedje, line 11: rust-analyzer sadly doesn't do it, but the general style is to import these from `crate::syntax` and `crate::utils` reexports instead of the raw crates (similar problem frequently happens with `typst_macros`).

crates/typst/src/model/par.rs
```diff
@@ -224,3 +231,62 @@ pub enum Linebreaks {
...
+#[elem(name = "line", title = "Paragraph Line", Construct, Locatable, Count)]
+pub struct ParLine {
+    /// How to number each line. Accepts a
+    /// [numbering pattern or function]($numbering).
+    ///
+    /// ```example
+    /// #set par.line(numbering: "1.")
+    ///
+    /// #lorem(100)
```
laurmaedje, line 246: lorem(100) is too long for the docs style

crates/typst/src/layout/flow.rs
```diff
@@ -856,6 +960,44 @@ impl<'a, 'e> FlowLayouter<'a, 'e> {
 
         Ok(())
     }
+
+    fn layout_line_number(&mut self, numbering: &Numbering, y: Abs) -> SourceResult<()> {
+        let line_counter = Counter::of(ParLineMarker::elem());
```
laurmaedje, line 965: This is a general style nit and mostly just my personal preference, but (as also noted in the other PR), I think many variable names could be shortened. This makes the code a bit less verbose.

E.g.:
- `line_counter` => `counter` (we don't care about any other counter here, so it's clear enough)
- `line_counter_update` => either `update` or `counter_update`
- `line_number` => `number`

crates/typst/src/layout/flow.rs
```diff
@@ -856,6 +960,44 @@ impl<'a, 'e> FlowLayouter<'a, 'e> {
...
+        let mut line_counter_update = line_counter
+            .clone()
+            .update(Span::detached(), CounterUpdate::Step(NonZeroUsize::ONE));
+
+        let update_hash = crate::utils::hash128(&line_counter_update);
+        let location = self.locator.next_location(self.engine.introspector, update_hash);
+        line_counter_update.set_location(location);
```
laurmaedje, line 972: This manual location management is a bit unidiomatic. `next_location` is not supposed to be used outside of the core realization code. 

I don't think it's necessary. Wouldn't just a normal `update() + display()` work?

crates/typst/src/layout/flow.rs
laurmaedje, line 965: Why do we use the `ParLineMarker` counter instead of `ParLine`?

crates/typst/src/layout/flow.rs
```diff
@@ -875,3 +1017,35 @@ fn collect_footnotes(notes: &mut Vec<Packed<FootnoteElem>>, frame: &Frame) {
...
+) {
+    for (pos, item) in frame.items() {
+        match item {
+            FrameItem::Group(group) => {
+                collect_par_lines(lines, &group.frame, prev_y + pos.y)
+            }
+            FrameItem::Tag(tag)
+                if !lines
+                    .iter()
+                    .any(|(_, line)| line.location() == tag.elem.location()) =>
```
laurmaedje, line 1040: will this happen in practice? It makes the thing $O(n^2)$ (footnotes have that problem as well, but possibly a bit less pronounced in practice) and since we deduplicate by Y anyway it's perhaps unnecessary.

crates/typst/src/layout/inline/line.rs
```diff
@@ -486,6 +492,19 @@ pub fn commit(
...
+    if let Some(numbering) = ParLine::numbering_in(styles) {
+        let par_line = ParLineMarker::new(numbering).pack();
+        let par_line_region =
+            Regions::one(Size::new(width, full), Axes::new(true, false));
+        frames.push((
+            offset,
+            par_line
+                .layout(engine, locator.next(&par_line), styles, par_line_region)?
+                .into_frame(),
+        ));
```
laurmaedje, line 505: IIUC this only generates a Frame with a single `Tag`. Could we generate the tag directly?

Going through the full layout machinery feels a bit wasteful just to generate a `Tag` whose location doesn't even really interest us. If I'm not mistaken, the location deduplication in the flow is unnecessary (as mentioned in another comment). I that case we could also generate a somewhat fake Tag.

I'm not sure here, I think the other comments should be addressed first and we can discuss this again later.

crates/typst/src/layout/flow.rs
laurmaedje: You can replace `display_at_loc` with `CounterDisplayElem` to not require a location.

crates/typst/src/layout/flow.rs
PgBiel: Good catch, I was going so fast with my prototype that I didn't even notice this :joy:

crates/typst/src/layout/flow.rs
PgBiel: This was actually an accident (a "bug"), which has now been promoted into a feature (:sunglasses:), as we don't want to expose the line number counter for now, until we find a good way to allow consistent counter value fetching and updating.

crates/typst/src/layout/flow.rs
PgBiel: Yeah, removing this seems to be totally fine, especially with deduplication later on.

crates/typst/src/model/par.rs
PgBiel: ty :)

crates/typst/src/layout/inline/line.rs
PgBiel: I inlined the location generation for the tag, and it seems to be working. Let me know if the new solution is appropriate.

crates/typst/src/model/par.rs
```diff
@@ -206,3 +215,166 @@ impl ParbreakElem {
...
+
+cast! {
+    ParLineNumberingScope,
+    self => match self {
+        Self::Document => "document".into_value(),
+        Self::Page => "page".into_value(),
+    },
+    "document" => Self::Document,
+    "page" => Self::Page,
+}
```
laurmaedje, line 338: You can use the `Cast` derive macro for this. Then the doc comments will also show up the in rendered Typst docs.

crates/typst/src/model/par.rs
```diff
@@ -206,3 +215,166 @@ impl ParbreakElem {
...
+    pub numbering_scope: ParLineNumberingScope,
+}
+
+impl Construct for ParLine {
+    fn construct(_: &mut Engine, args: &mut Args) -> SourceResult<Content> {
+        bail!(args.span, "cannot be constructed manually");
+    }
+}
+
+impl Count for Packed<ParLine> {
```
laurmaedje, line 311: Can probably skip count here since the counter we use is the one of `ParLineMarker`

tests/ref/line-numbers-page-scope-with-columns.png
laurmaedje, line 1: the overflow looks a bit bad and makes it seems like the test failed

tests/ref/line-numbers-peek-into-grid.png
laurmaedje, line 1: with this one as well, it's hard to say what the intent of the test is (and it's fairly large)

tests/suite/layout/line-numbers.typ
```diff
@@ -0,0 +1,202 @@
...
+--- line-numbers-columns-override ---
+#set columns(gutter: 1.5em)
+#set page(columns: 2, margin: (x: 1.5em))
+#set par.line(numbering: "1", number-margin: end, number-clearance: 0.5em)
+
+Hello \
+Beautiful \
+World
+#colbreak()
+Birds\
```
laurmaedje, line 118: Please add spaces before `\` :)

crates/typst/src/layout/flow.rs
```diff
@@ -1214,6 +1259,90 @@ impl<'a, 'e> FlowLayouter<'a, 'e> {
...
+                FixedAlignment::End => size.x + number_clearance,
+                // Shouldn't be specifiable by the user due to 'OuterHAlignment'.
+                FixedAlignment::Center => unreachable!(),
+            };
+            let number_pos = Point::new(number_x, line.y);
+            max_number_width.set_max(number.width());
+            line_numbers.push((number_pos, number, number_align, number_margin));
+        }
+
+        for (mut pos, number, align, margin) in line_numbers {
```
laurmaedje, line 1328: Any particular reason to first collect into `line_numbers` and then immediately iterate? Could we do it directly?

crates/typst/src/layout/flow.rs
```diff
@@ -1214,6 +1259,90 @@ impl<'a, 'e> FlowLayouter<'a, 'e> {
             }
         }
 
+        if self.should_reset_page_scoped_line_numbers() {
```
laurmaedje, line 1262: Could you extract the whole line number logic into its own function?

crates/typst/src/layout/flow.rs
```diff
@@ -1082,6 +1106,13 @@ impl<'a, 'e> FlowLayouter<'a, 'e> {
             && !self.items.is_empty()
             && self.items.iter().all(FlowItem::is_out_of_flow)
         {
+            let mut output = Frame::soft(self.initial);
+
+            if self.should_reset_page_scoped_line_numbers() {
+                // Reset line numbers at the first column if requested.
+                output.push_frame(Point::zero(), self.layout_line_number_reset()?);
+            }
```
laurmaedje, line 1114: Then we can call that here as well.

PgBiel: You mean the line number logic would apply here too, but with an empty vector of lines?

tests/ref/line-numbers-page-scope-with-columns.png
PgBiel: I'm having a bit of trouble with this... maybe I should just increase the page size more?

tests/ref/line-numbers-peek-into-grid.png
PgBiel: The intent was to ensure line numbers are still emitted for things nested deep within blocks , grids etc., but yeah it could be reduced

crates/typst/src/layout/flow.rs
PgBiel: This is due to collecting `max_number_width` which is needed for alignment.

An artifact of this, as can be seen in the `line-numbers-start-alignment` test, is that line numbers with `start` alignment in the `start` margin are probably aligned between themselves in the same page, but not between different pages. I'm not sure of the best way to fix this, or even if it would be worth it to do so, frankly (I expect `start` alignment on a `start` margin, or `end` on an `end` margin, to be a rare style choice anyway).

tests/suite/layout/line-numbers.typ
PgBiel: whoops

crates/typst/src/layout/flow.rs
PgBiel: okay i think that works, although I haven't added an explicit test for it yet. Resolving for now.

crates/typst/src/layout/flow.rs
laurmaedje: makes sense

crates/typst/src/layout/flow.rs
PgBiel: Added a test for this now.

crates/typst/src/model/par.rs
```diff
@@ -206,3 +214,149 @@ impl ParbreakElem {
...
+    ///   numbering: "1",
+    ///   number-clearance: 0.5pt
+    /// )
+    ///
+    /// Typesetting \
+    /// Styling \
+    /// Layout
+    /// ```
+    #[ghost]
+    #[default(Length::from(Abs::cm(1.0)))]
```
PgBiel, line 279: Is 1cm fine as a default? This gave me some headaches for test images since they have a small margin, but i think this looks ok in actual documents.

crates/typst/src/model/par.rs
```diff
@@ -206,3 +214,143 @@ impl ParbreakElem {
...
+    /// new page.
+    Page,
+}
+
+/// A marker used to indicate the presence of a line.
+///
+/// This element is added to each line in a paragraph and later searched to
+/// find out where to add line numbers.
+#[elem(Construct, Locatable, Count)]
+pub struct ParLineMarker {
```
PgBiel, line 327: I've removed `Show` from `ParLineMarker` as it didn't seem to be necessary - we aren't laying it out directly, only its tag. Let me know if I should undo this.

daskol:
It's great to see support for line numbers. I wanted to try it out and built `typst` from commit 92ec56601. Everything is great, but how can I customize the font? The following rule doesn't work.

```typst
show par.line: set text(size: 6pt)
```

PgBiel:
You can use `#set par.line(numbering: n => text(size: 6pt)[#n])`. `par.line` is a "virtual" element in the sense that it only exists for `#set par.line(numbering: ..., /* and other options */)` to work, but you can't do anything else with it.

daskol:
I see. Thank you.

joshuabeny1999:
Nice work. Would need this for an assigment in my master. How can i use it now on macOS? Or when the next release will be? The last one was in May...

laurmaedje:
The next release will be fairly soon. If you want to use the development version, you need to install Rust and then follow the instructions in the README.

PR #4517, opened 2024-07-08
Allow non-utf8 values for `input` and `output`

T0mstone:
When we started accepting `-`, the implementation accidentally required paths to be UTF-8 now. This fixes that.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -5,7 +5,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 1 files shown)


T0mstone:
> FYI, it is the default behavior, the stdin PR just added stdin support. You have to opt out of it, e.g., the way you did it.

No, the default value parser for `PathBuf` accepts non-utf8 paths too. The issue was the custom value parser requiring `&str` as the input.

Andrew15-5:
Ohh, really? That's very cool how the value parser adapts to the type of the argument.

laurmaedje:
Thanks!

PR #4522, opened 2024-07-08
typst-pdf: Build *_DEFLATED in build script instead of lazily at runtime

bluebear94:


laurmaedje:
Have you benchmarked the performance impact of this? It feels a bit like premature optimization to me.

bluebear94:
* Binary size for `release`: 32535304 → 32530720
* PDF export time: 17665106 ns → 17410679 ns with 1 run using the `--timings` flag

So the performance impact is pretty small, but I don’t see any reason why the code should perform worse after this change. And getting rid of these particular `Lazy` statics was easy compared to some of the ones elsewhere in the codebase.

laurmaedje:
> So the performance impact is pretty small, but I don’t see any reason why the code should perform worse after this change.

I'm not concerned about the code performing worse, mostly about unnecessary complexity with `build.rs` and slower builds because some dependencies are built for the build script (and potentially even twice if host and target mismatch, like for wasm).

laurmaedje:
For me, personally, a `build.rs` needs to bring meaningful benefits to justify its existence. This does not feel like it's the case here, so I'll close this. Thanks anyway!

PR #4141, opened 2024-05-15
Add flush element

bluebear94:
See #4134.

PR #4537, opened 2024-07-11
Remove `place.flush` from global scope

laurmaedje:
Was added to both `place` and the global scope by accident in #4141.


bluebear94:
whoops

PR #4540, opened 2024-07-11
Add typst-kit crate

### LINKED ISSUE #4452: Crate for downstream tooling

tingerrr:
This PR adds the `typst-kit` crate, which contains various default implementations of things a Typst tool may need, such as:
- [x] font searching
- [ ] manifest validation
- [x] package storage with on demand package downloads from package registries
- [x] project management
  - [x] project lookup
  - [ ] project locking

Resolves #4452.


tingerrr:
For the reviewers: I suggest reviewing commit by commit for simplicity.

I have added a project crate which should at some point have things like the project locking mentioned in #4452, but I left the feature out as I'm not 100% sure how to implement it in a cross-platform manner. Currently, it only contains recursive project lookup by searching for the manifest file.

Regarding package downloads, I initially wanted to abstract this a little more, but it proved difficult. I think this is the feature where it's least clear what should be configurable and what shouldn't, so this can be extended later when the API surface is clearer. It will likely change once we add support for custom registries and git packages, so it's ok that it's not yet fully defined, I think.

tingerrr:
I should note that I had a discussion with @Thumuss regarding manifest validation. While this is something that only 2 projects need, [utpm](https://github.com/Thumuss/utpm) and the package bundler, it may still be beneficial to add to `typst-kit` or `typst-syntax`.

Thumuss:
> I should note that I had a discussion with @Thumuss regarding manifest validation. While this is something that only 2 projects need, [utpm](https://github.com/Thumuss/utpm) and the package bundler, it may still be beneficial to add to `typst-kit` or `typst-syntax`.

As I said earlier on discord, [utpm](https://github.com/Thumuss/utpm) will need manifest validation because it will publish directly a PR on [typst/packages](https://github.com/typst/packages). We don't need a big validation system, just enough to block malformed `typst.toml`.

laurmaedje:
I think the Typst Universe specific checks don't really belong in the compiler repository. Was that the idea or are you talking about some other kind of validation?

tingerrr:
I was thinking of this year, but there are other forms like package names being identifiers which are directly tied to how package specs are parsed.

But I guess we can leave it out.

crates/typst-kit/src/project.rs
```diff
@@ -0,0 +1,58 @@
+//! Identify and manage Typst projects.
```
laurmaedje, line 1: I'm not sure whether this should really exist in `typst-kit` at this moment. Since the `typst.toml` file is only used in packages, there simply isn't really a way to properly identify projects, but this makes it seem like there is.

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,264 @@
...
+        //       reasons, but we still want to show a field that indicates
+        //       whether a certificate was loaded and is in use, so we implement
+        //       a small helper because `.field_with()` is not yet stable
+        // TODO: remove this once `.field_with` is stabilized (feature =
+        //       debug_closure_helpers)
+        struct CertExists;
+
+        impl Debug for CertExists {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, "<Certificate>")
```
laurmaedje, line 76: `Certificate(..)` is more common in the code base. But honestly I think it could simply be `Downloader(..)`.

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,264 @@
...
+            .field("user_agent", &self.user_agent)
+            .field("cert_path", &self.cert_path)
+            .field("cert", &self.cert.get().and_then(Option::as_ref).map(|_| CertExists))
+            .finish()
+    }
+}
+
+impl Downloader {
+    /// Crates a new downloader with the given user agent and no certificate.
+    pub fn new<S: Into<EcoString>>(user_agent: S) -> Self {
```
laurmaedje, line 90: `user_agent: impl Into<EcoString>` is more typical in the code base.

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,264 @@
...
+            .get_or_init(|| {
+                let pem = std::fs::read(self.cert_path.as_ref()?).ok()?;
+                Certificate::from_pem(&pem).ok()
+            })
+            .as_ref()
+    }
+
+    /// Download binary data from the given url and report its progress.
+    #[allow(clippy::result_large_err)]
+    pub fn download_with_progress(
```
laurmaedje, line 131: I think this should be moved below `download`, basically following the order simple API -> advanced API.

crates/typst-kit/src/fonts.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// to a collection.
+    index: u32,
+    /// The lazily loaded font.
+    font: OnceLock<Option<Font>>,
+}
+
+impl FontSlot {
+    /// Returns the path at which the font can be found on the system.
+    pub fn path(&self) -> &Path {
+        &self.path
```
laurmaedje, line 33: For embedded fonts, the path is empty. So we should maybe switch it to `Option` or at least note it.

crates/typst-kit/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(feature = "packages")]
```
laurmaedje, line 1: Could you add a brief bit of docs to the crate root?

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/// An implementation of [`Progress`] whth no-op reporting, i.e. reporting
+/// events are swallowed.
+pub struct ProgressSink;
+
+impl Progress for ProgressSink {
+    fn print_start(&mut self) {}
+    fn print_progress(&mut self, _state: &DownloadState) {}
+    fn print_finish(&mut self, _state: &DownloadState) {}
```
laurmaedje, line 40:
```suggestion
    fn print_progress(&mut self, _: &DownloadState) {}
    fn print_finish(&mut self, _: &DownloadState) {}
```

crates/typst-kit/src/project.rs
tingerrr: I figured once git packages (where normal consumer projects will have a `typst.toml`) or other ways to identify a project are added they could be added here and people get the same API.

I kept it as simple as it is here, because in my downstream implementation I have optional fallbacks to look for `src/main.typ`, `src/lib.typ`, `main.typ` and various tooling configuration like `typstfmt.toml`, which obviously didn't belong here.

I'll remove it then, doesn't make much sense yet.

crates/typst-kit/src/fonts.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// 2. System fonts (if included)
+    /// 3. Embedded fonts (if enabled)
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # use typst_kit::fonts::FontSearcher;
+    /// let mut searcher = FontSearcher::new();
+    /// searcher.search(["./assets/fonts/"], true);
+    /// ```
+    pub fn search<I, P>(&mut self, font_dirs: I, include_system_fonts: bool)
```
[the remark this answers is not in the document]
tingerrr: Sounds good, I think builder style makes sense, so I'll go with that.

crates/typst-kit/src/lib.rs
```diff
@@ -0,0 +1,8 @@
...
+pub mod download;
```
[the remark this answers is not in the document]
tingerrr: That was my initial thinking, but there was nothing else using it. I'll add that.

crates/typst-kit/src/download.rs
tingerrr: I'd go with `Certificate(..)` because I think being able to see if a certificate was loaded will aid in debugging, considering any IO error from loading it is swallowed.
Could perhaps also store the result and simply do `.ok()` in the getter, but I don't know if this interferes with any auto traits.

crates/typst-cli/src/update.rs
```diff
@@ -7,15 +7,37 @@ use semver::Version;
...
 
 use crate::args::UpdateCommand;
-use crate::download::{download, download_with_progress};
+use crate::download::display_download_progress;
+use crate::{download, terminal};
 
 const TYPST_GITHUB_ORG: &str = "typst";
 const TYPST_REPO: &str = "typst";
 
+struct PrintUpdateDownload;
```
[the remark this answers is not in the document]
tingerrr: I guess so, the update implementation doesn't use the cyan `downloading` hint, but if that's not a problem,then yes.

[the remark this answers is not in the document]
laurmaedje: I think it's nice if it's unified in style as well

crates/typst-kit/src/download.rs
tingerrr: Well that question has seemingly answered itself, on main this was already expanded to include the Result, so I'll go with that.

crates/typst-kit/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+packages = [
+  "downloads",
+
+  "dep:dirs",
+  "dep:flate2",
+  "dep:tar",
+]
+
+# Add project management utilities
+project = []
```
laurmaedje, line 63: this is unused I think

crates/typst-kit/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+default = [
+  "fonts",
+  "packages",
+  "project",
+]
+
+# Add font loading utilitites
+fonts = [
+  "dep:fontdb",
+  "fontdb/memmap",
```
laurmaedje, line 42: can/do these typicaly also use `dep:` or not? I don't really know.

crates/typst-kit/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+tar = { workspace = true, optional = true }
+ureq = { workspace = true, optional = true }
+
+# Explicitly depend on OpenSSL if applicable, so that we can add the
+# `openssl/vendored` feature to it if `vendor-openssl` is enabled.
+[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios", target_os = "watchos", target_os = "tvos")))'.dependencies]
+openssl = { workspace = true }
+
+[features]
+default = [
```
laurmaedje, line 33: I think these arrays can all be single line

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,267 @@
...
+            user_agent: user_agent.into(),
+            cert_path: None,
+            cert: OnceCell::new(),
+        }
+    }
+
+    /// Crates a new downloader with the given user agent and certificate path.
+    ///
+    /// If the certificate cannot be read it is set to `None`.
+    pub fn with_path<S: Into<EcoString>>(user_agent: S, cert_path: PathBuf) -> Self {
```
laurmaedje, line 100: these should also use the `impl Into` style

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/// A minimal https client for downloading various resources.
+pub struct Downloader {
+    user_agent: EcoString,
+    cert_path: Option<PathBuf>,
+    cert: OnceCell<Certificate>,
+}
+
+impl Debug for Downloader {
```
laurmaedje, line 63: could you move the trait impl below the inherent impl?

crates/typst-kit/src/download.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // NOTE: Certificate does not implement Debug, likely for security
+        //       reasons, but we still want to show a field that indicates
+        //       whether a certificate was loaded and is in use, so we implement
+        //       a small helper because `.field_with()` is not yet stable
+        // TODO: remove this once `.field_with` is stabilized (feature =
+        //       debug_closure_helpers)
```
laurmaedje, line 70: typst-utils actually has a `debug` helper function for this

crates/typst-kit/src/fonts.rs
```diff
@@ -0,0 +1,214 @@
...
+    index: u32,
+    /// The lazily loaded font.
+    font: OnceLock<Option<Font>>,
+}
+
+impl FontSlot {
+    /// Returns the path at which the font can be found on the system, or `None`
+    /// if the font was embedded.
+    pub fn path(&self) -> Option<&Path> {
+        if self.path.as_os_str().is_empty() {
```
laurmaedje, line 35: Perhaps we should turn the `path` into a `Option<PathBuf` in the type as well. in `get`, we would just unwrap. better to have it panic in case of a bug than fail silently.

crates/typst-kit/src/fonts.rs
```diff
@@ -0,0 +1,214 @@
...
+            db: Database::new(),
+            include_system_fonts: true,
+            #[cfg(feature = "embed-fonts")]
+            include_embedded_fonts: true,
+            book: FontBook::new(),
+            fonts: vec![],
+        }
+    }
+
+    /// Whether to search for and load system fonts.
```
laurmaedje, line 107: The defaults should be documented here.

crates/typst-kit/src/fonts.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// # use typst_kit::fonts::FontSearcher;
+    /// let fonts = FontSearcher::new()
+    ///     .include_system_fonts(true)
+    ///     .search();
+    /// ```
+    pub fn search(&mut self) -> Fonts {
+        self.search_with::<_, &str>([])
+    }
+
+    /// Start searching for and loading fonts, with addiitonal directories.
```
laurmaedje, line 134:
```suggestion
    /// Start searching for and loading fonts, with additional directories.
```

crates/typst-kit/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+//! Typst-kit contains various default implementations of functionality used in
+//! typst-cli. It is intended as a single source of truth for things like font
+//! searching, package downloads and more. Each component of typst-kit is
+//! optional, but enabled by default.
+//!
+//! # Components
+//! - [fonts] contains a default implementation for searching local and system
+//!   installed fonts. It is enabled by the `fonts` feature flag, additionally
+//!   the `embed-fonts` feature can be used to embed the typst default fonts.
```
laurmaedje, line 9:
```suggestion
//!   the `embed-fonts` feature can be used to embed the Typst default fonts.
```

crates/typst-kit/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+//!   - For text: Linux Libertine, New Computer Modern
+//!   - For math: New Computer Modern Math
+//!   - For code: Deja Vu Sans Mono
+//! - [download] contains functionality for making simple web requests with
+//!   status reporting, useful for downloading packages from package registires.
+//!   It is enabled by the `downloads` feature flag, additionally the
+//!   `vendor-openssl` can be used on operating systems other than macOS and
+//!   windows to vendor openssl when building.
```
laurmaedje, line 17:
```suggestion
//!   Windows to vendor OpenSSL when building.
```

crates/typst-kit/src/package.rs
```diff
@@ -0,0 +1,168 @@
...
+pub const DEFAULT_PACKAGES_SUBDIR: &str = "typst/packages";
+
+/// Holds information about where packages should be stored and downloads them
+/// on demand, if possible.
+#[derive(Debug)]
+pub struct PackageStorage {
+    /// The path at which non-local packages should be stored when downloaded.
+    pub package_cache_path: Option<PathBuf>,
+    /// The path at which local packages are stored.
+    pub package_path: Option<PathBuf>,
```
laurmaedje, line 28: perhaps we should keep these private for flexibility? or are they used externally?

laurmaedje (commented):
I left some further minor comments. Other than that, it looks good to merge. 

There are some unticked boxes in the PR description. Do you still want to tackle those here? Personally, I think we can merge what's there first and proceed with the rest separately.

crates/typst-kit/Cargo.toml
tingerrr: Using `dep:` for an optional dependency means that there's no implicit feature created for it (https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies).

crates/typst-kit/src/package.rs
tingerrr: Well, they are used in the World impl in `typst-cli`, I figured, since we pass them in ourselves and don't have any restrictions regarding access (the consumer can read and write them), we can keep them public.

crates/typst-kit/Cargo.toml
laurmaedje: What I meant was whether `"dep:fontdb/memmap"` is a thing, but on second thought it probably makes no sense, since the implicit feature is already not created.

tingerrr:
> There are some unticked boxes in the PR description. Do you still want to tackle those here? Personally, I think we can merge what's there first and proceed with the rest separately.

For project locking, I was unable to come up with anything cross-platform and safe. I left some notes on #4658 regarding those issues.

Regarding manifest validation, we decided to leave this out of `typst-kit` since it's universe specific.

laurmaedje:
Alright, then we can merge once after you've pushed.

tingerrr:
> Alright, then we can merge once you've pushed.

One review comment is left open, do you prefer getters + setters for effectively public fields?

crates/typst-kit/src/package.rs
laurmaedje: What I'm a little cautious about is people changing them half-way through. Right now it's maybe ok, but we might not want to give that guarantee. 

That aside, I can't actually find where in the CLI World impl they are used.

tingerrr: It seems they weren't, I went with getters only.

laurmaedje:
Thank you!

crates/typst-kit/Cargo.toml
```diff
@@ -0,0 +1,57 @@
...
+fontdb = { workspace = true, optional = true }
+native-tls = { workspace = true, optional = true }
+once_cell = { workspace = true }
+tar = { workspace = true, optional = true }
+ureq = { workspace = true, optional = true }
+
+# Explicitly depend on OpenSSL if applicable, so that we can add the
+# `openssl/vendored` feature to it if `vendor-openssl` is enabled.
+[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios", target_os = "watchos", target_os = "tvos")))'.dependencies]
+openssl = { workspace = true }
```
laurmaedje, line 31: I think this should be optional as well and only enabled under `downloads`.

laurmaedje:
I reviewed this again after the two regression, this time actually comparing the moved code fragments. Didn't find anything major, but I noticed that typst-kit's OpenSSL dependency is unconditional on Linux, even though it is only needed for downloading.

tingerrr:
I left his completely unchanged, this was already the case before.

laurmaedje:
Yeah, but the CLI uses downloading unconditionally whereas in typst-kit it is optional.

PR #4544, opened 2024-07-11
Ignore flanking ignorant elements when scaling lr delimiters

### LINKED ISSUE #4433: closing parenthesis is not recognised in math formula in context block
### Description

When a closing parenthesis is in a context block, it isn't detected by lr and appears small when it should be sized.
The bug is simple to reproduce:
```typ
#let rule(body) = {body}

#show ")" :rule

$
	(1/2)
$
```
Gives:
image
as expected  but
```typ
#let rule(body) = context{body}

#show ")" :rule

$
	(1/2)
$
```
Gives:
image

And I thing it's a bug because:
```typ
#let rule(body) = context{body}

#show "(" :rule

$
	(1/2)
$
```
Gives:
image

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

bluebear94:
Fixes #4433.


crates/typst/src/math/lr.rs
```diff
@@ -64,7 +64,25 @@ impl LayoutMath for Packed<LrElem> {
...
+                break;
+            }
+            start_idx += 1;
+        }
+        while end_idx > start_idx {
+            if !fragments[end_idx - 1].is_ignorant() {
+                break;
+            }
+            end_idx -= 1;
+        }
```
EpicEricEE, line 83: I think this would be better with the iterator style:
```suggestion
        let start_idx = fragments
            .iter()
            .position(|f| !f.is_ignorant())
            .unwrap_or(fragments.len());
        let end_idx = fragments
            .iter()
            .skip(start_idx)
            .rposition(|f| !f.is_ignorant())
            .map_or(start_idx, |i| start_idx + i + 1);
```

bluebear94: Better in what respect? It’s more work for the optimizer while being no less error-prone than a while-loop.

EpicEricEE: Sorry, "better" is arguably not a good word here, it's more just personal preference and that (I think) the iterator pattern is usually preferred over loops throughout the code base. This is of course subjective, and performance aspects are probably negligible anyway.

laurmaedje: It's just a matter of preference, but I prefer the iterator style as well.

gabfssilva: Mutable code is _always_ more error-prone than immutable code. One can argue whether it is easier for newcomers or not -- which I also disagree, btw -- but, if there's no performance difference, it's quite hard to argue against immutability.

bluebear94: > Mutable code is _always_ more error-prone than immutable code.

\[citation needed\]

In this case, the iterator-based code has an extra `+ start_idx` to compensate for the skip, plus an extra `+ 1` to avoid a fencepost error.

In any case, changing the loops to use the iterator-based approach adds 64 bytes to the executable on release mode (33300968 vs. 33300904 on the `x86_64-unknown-linux-gnu` target). I also tried comparing the [assembly outputs](https://gist.github.com/bluebear94/95c7e4d5635f247ebb8087134ec713e8) of the two, and it seems that some bound checks aren’t being elided in the iterator approach. But to be sure of the performance impacts, we would need to use something like [iai-callgrind](https://github.com/iai-callgrind/iai-callgrind) to eke out small differences like these.

bluebear94: The iterator approach turns out to be more optimizable for `start_idx` ([test code for posterity](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=ad2f10b8a2aa1cc6d40cad8f1c411d05)).

laurmaedje: The code has become really complex now. It is far off the hot path, so I think the discussion about performance is a bit misguided (adding a comment about how it optimizes actually feels counter-productive to me as it brings focus to performance even though it doesn't really matter here.). 

I would like to just pick the simplest/clearest. My proposed solution would be:
```rs
let start_skip = fragments
    .iter()
    .filter(MathFragment::is_ignorant)
    .count();

let end_skip = fragments
    .iter()
    .rev()
    .filter(MathFragment::is_ignorant)
    .count();

let inner = &fragments[start_skip..fragments.len() - end_skip];
```

bluebear94: Not being on the “hot path” is not a sufficient reason to purposefully make a piece of code perform worse.

laurmaedje: It is when readability suffers. I'm not sure what else to say here anymore.

bluebear94: IMO my code isn’t a big sacrifice in readability for the performance gain, especially when it might be called many times when compiling a math-heavy document (given that anything in some sort of brackets is implicitly wrapped in an `LrElem`).

laurmaedje:
I'd be happy to merge the fix in principle, but I would like to see the code changed to the one I proposed [here](https://github.com/typst/typst/pull/4544#discussion_r1677999709).

bluebear94:
> I'd be happy to merge the fix in principle, but I would like to see the code changed to the one I proposed [here](https://github.com/typst/typst/pull/4544#discussion_r1677999709).

I’ll make it clear right now: I will not change the code until someone suggests code that performs at least as well as it.

laurmaedje:
In that case I'll close the PR.

Issue #1304, opened 2023-05-23
Make arrows in math stretchable


ice1000:
### Description

There are times when I want to write some text above an arrow like this

![image](https://github.com/typst/typst/assets/16398479/fc141739-2ee9-4f65-9357-7ec258ad8d7c)

Currently I'm using `typst-cd` with

```typst
#align(center)[
  #commutative_diagram(
    node((0, 0), [$$]),
    node((0, 2), [$$]),
    arr((0,0), (0,2), [injective map], "inj", label_pos: -1em)
  )
]
```

But it would be nice if there are native support for:

1. long, hooked arrows
2. putting text above an arrow (instead of just braces or brackets like we have now)

### Use Case

As above. This can be used to describe injective maps.

Enivex:
I don't know how typst-cd does it, but here's a typst solution that uses the symbol in your actual font (I don't know how robust it is, so it may need modification if the font has very large arrow heads)

```
#let long-symbol(sym, factor) = {
  assert(type(sym) == "symbol", message: "Input needs to be a symbol")
  assert(type(factor) == "integer" or type(factor) == "float", message: "Scale factor must be a number")
  assert(factor >= 1, message: "Scale factor must be >= 1")
  
  factor = 5*factor - 4
  let body = [#sym]
  style(styles => {
    let (body-w,body-h) = measure(body,styles).values()
    align(left)[
      #box(width: body-w*2/5,height: body-h,clip: true)[
        #align(left)[
          #body
        ]
      ]
      #h(0cm)
      #box(height: body-h, width: body-w*1/5*factor)[
        #scale(x: factor*100%,origin:left)[
          #box(height: body-h, width: body-w*1/5,clip:true)[
            #align(center)[
              #body
            ]
          ]
        ]
      ]
      #h(0cm)
      #box(width: body-w*2/5,clip: true)[
        #align(right)[
          #body
        ]
      ]
    ]
  })
}

$
  f: X attach(limits(#long-symbol(sym.arrow.r.hook,8)), t: "injective map") Y\
  #long-symbol(sym.arrow.l.r.double.long,10) \
  #long-symbol(sym.arrow.l.filled,5)
$
```

![image](https://github.com/typst/typst/assets/2191754/d5f7776a-3623-48a4-8b62-a844b96cb26f)

ShadyStokes:
Hello! I ended up writing my own little function that places annotations on top of symbols
In case someone finds it useful:

```
#let annotated-symbol(
[...]

PR #4545, opened 2024-07-11
Wrapping with parens should not push superscripts higher

Leedehai:
A parenthesis glyph itself (i.e. `(` or `)`) is a text-like [1], so superscripts attached to it should not be pushed higher, just like other non-large glyphs [2]
https://github.com/typst/typst/blob/ab5cebc57c8345f397c3f0a1ed0c024dc6811e2f/crates/typst/src/math/attach.rs#L427

As shown below, parenthesis glyphs of an LR that wraps around text-like glyphs are the same as standalone parenthesis glyphs, and should still be text-like. Therefore, the whole sequence of glyphs in the said LR are all text-like, and the superscripts attached to this sequence should still _not_ be pushed higher.

This is addressed by marking the frame fragment built from a `MathRun` containing all text-like fragments text-like.

Comparisons:
| |Case 1|Case 2|
|:-|:-|:-|
|Before the patch|(1)|(3)|
|After the patch|(2)|(4)|

Taking figure (4) as an example. The 1st red box shows the superscript position when the base (an `x` character) is a non-large glyph. The 2nd box shows the superscript position when the base is a single parenthesis glyph. The 3rd box's base is a sequential combination of the these glyphs, so the superscript in the 3rd should have the same vertical position like before, instead of being pushed higher.

[1] `MathFragment::is_text_like()` is for the purpose of TeXBook Appendix G 18a's special handling to determine the vertical position for superscripts and subscripts.
[2] See [`MathFragment::is_text_like()`](https://github.com/typst/typst/blob/ab5cebc57c8345f397c3f0a1ed0c024dc6811e2f/crates/typst/src/math/fragment.rs#L144-L150). Large glyphs (i.e. `MathClass::Large`) are the ones like the big integral symbol.


crates/typst/src/math/row.rs
```diff
@@ -161,7 +161,11 @@ impl MathRun {
         if self.0.len() == 1 {
             self.0.into_iter().next().unwrap()
         } else {
-            FrameFragment::new(ctx, styles, self.into_frame(ctx, styles)).into()
+            let text_like =
+                self.iter().all(|e| e.math_size().is_none() || e.is_text_like());
```
laurmaedje, line 165: why `e.math_size().is_none()`?

Leedehai: Because it might be a Space, which shouldn't affect the layout but if it's not excluded, it will make an otherwise true result false.

laurmaedje: ah okay, so it is not related to the MathSize, it's more a convenient way to check whether it is `Glyph | Variant | Frame`? if yes, that's not immediately clear.

Leedehai: Yes, but you can also think that not having a MathSize will make it not clear how much space the fragment shall occupy in the vertical dimension and so its text_like() is meaningless.

laurmaedje: Could you add a brief comment above the let binding?

Leedehai: Done :)

laurmaedje:
Thanks!

laurmaedje:
@Leedehai Since you've recently worked a bunch on `AttachElem`, I'd be curious about your thoughts on https://github.com/typst/typst/issues/1304#issuecomment-1723305545.

Leedehai:
Sure, opened https://github.com/typst/typst/pull/4569

Issue #2025, opened 2023-08-27
Check if file exists


Beiri22:
### Description

When opening files for your document, the system fails if the file does not exist, as expected. But when you only use the filename as metadata or inside the document, nothing is checked. It would be nice to have a function that checks if a file exists in the source tree.

### Use Case

When using file names inside the document or in metadata, misspelled file names could be detected.

owiecc:
Or perhaps we could get a maybe/optional type?

Beiri22:
How would you solve the problem with those?

owiecc:
You can throw an error if file not found. Or use some defaults instead. Whatever you feel is best depending on your situation.

Beiri22:
I'm not sure, if we are talking of the same thing. My problem is not about loading a file; really just about checking for existance without the overhead of loading the file. The typst compile is intended to panic in case of the file missing. I understand your approach different: With an optinal you can react to a missing file that is loaded. Those use cases are different.

laurmaedje:
Related: https://github.com/typst/typst/issues/1877

Beiri22:
The question is. How do you respect other conditions like: Is not a file, but a dir; exists but is not accessable? Do you just return a bool; or some kind of enum?

Andrew15-5:
I think if we are specifically talking about `if-file-exists()`, then we _proooobably_ want to read it, so both "not a file, but a dir" and "exists but is not accessible" must return `false`.

Otherwise, we would have to implement a separate `path` type like `Path` class from `pathlib` in Python:

![image](https://github.com/typst/typst/assets/37143421/48c56197-c710-4b3a-b6b4-ea47fd932a01)
![image](https://github.com/typst/typst/assets/37143421/a083c26f-46a9-4d6a-8f08-e07b7f159b22)

Beiri22:
I think, most of the possible types are not suitable for typst. We might be interested in directories and anything that reads like a file. So maybe a path type is a little over the top, that's right. In my use case I would not read the file after checking directly, but the third party application would read the file based on metadata extracted with query. So I can live with a simple boolean.

adrfantini:
This would be very handy for a project I'm working on as well, where I describe around 400 components in a manual. Each of those has a basic JSON file, but for some I want to add an optional `.typ` file to go into more detail.

Currently I do not think there's a way to do that, short of a WA plugin

jonsch318:
[...]

PR #4550, opened 2024-07-12
Add file-exists function

jonsch318:
When a read call to a non existent file is made, the compilation errors (as it should). This is why issue #2025 ask for a function to check for file existence.

`file-exists` is a simple implementation of that. It uses the `engine.world.file` functionality that should then cache for the upcoming read.

*The function naming is debatable and is just a proposal.* 
`exists` is too generic but alternatives `fexists` could work. Since this is most likely a temporary solution that will be deprecated once a more stable path type is necessary.

### DIFF
--- a/crates/typst/src/loading/file-exists.rs
+++ b/crates/typst/src/loading/file-exists.rs
@@ -1,0 +1,40 @@
+use ecow::EcoString;
+
+use crate::diag::{At, SourceResult};
+use crate::engine::Engine;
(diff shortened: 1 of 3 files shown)


laurmaedje:
As mentioned in some other issues, I'm not a big fan of having temporary solutions that we _know_ we will deprecate. It introduces unnecessary churn. I agree that a method `path.exists` would be the nice solution, so I'd prefer waiting for that.

PR #4553, opened 2024-07-13
Fix lexer behavior on non-whitespace before ref

### LINKED ISSUE #4536: Leading non-whitespace character changes ref syntax parsing behavior
### Description

Given a label with only a number as it's ident, one can reference it just fine as long as there's no leading non whitespace character. Whereas, the leading character is completely ignored if the identifier does not start with a digit. The example uses an open parenthesis, but any character seems to reproduce the behavior.

```typst
#figure[1] <1>
#figure[b] <b>

@1   // works
@b   // works
(@b) // works
(@1) // doens't work, is parsed is regular content
```

Funny enough:
- the built-in parser rejects only the last case (the bug in question)
- the GitHub parser rejects both the label and references consisting of digits only
- the tree sitter parser parses both references and labels correctly (which is why I thought I was going insane while debugging)

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

Coekjan:
Fixes #4536

### DIFF
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -474,7 +474,7 @@ impl Lexer<'_> {
                 Some('-') if !s.at(['-', '?']) => {}
                 Some('.') if !s.at("..") => {}
                 Some('h') if !s.at("ttp://") && !s.at("ttps://") => {}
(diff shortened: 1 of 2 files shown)


laurmaedje:
Thanks!

PR #4554, opened 2024-07-14
Raise error when failing to load & parse the specified certificate

Coekjan:
Before this PR, when CLI fails to load or parse the specified certificate, it just fails silently and does not report any messages.

This PR helps to report error when failing to load & parse the cert.

I am not sure if it should be warning or error. In my humble opinion, if the users need `--cert` or `TYPST_CERT` for networking, it would be better to treat this as **error**.

### DIFF
--- a/crates/typst-cli/src/download.rs
+++ b/crates/typst-cli/src/download.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use native_tls::{Certificate, TlsConnector};
-use once_cell::sync::Lazy;
+use once_cell::sync::OnceCell;
 use ureq::Response;
 
 use crate::terminal;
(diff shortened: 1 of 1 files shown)


crates/typst-cli/src/download.rs
```diff
@@ -3,37 +3,89 @@
...
+                .map_err(|err| eco_format!("cannot read certificate: {err}"))?;
+            Certificate::from_pem(&pem)
+                .map_err(|err| eco_format!("cannot parse certificate: {err}"))
+        })
+    })
+}
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug)]
+pub enum DownloadError {
```
laurmaedje, line 43: `ureq::Error` already implements `From<io::Error>` so perhaps we can just use a custom `io::Error` to pass through the certificate failures.

Coekjan: Fixed. Thanks for the review!

laurmaedje:
Thanks!

Issue #4422, opened 2024-06-20
Wrong color encoding in PDF gradients


EpicEricEE:
### Description

Gradients using the CMYK color space don't produce the same results in PDF as in the webapp preview. The behaviour is also dependent on the viewer or printer.

<details>

<summary>Example Code</summary>

```typ
#rect(fill: cmyk(red))

#rect(fill: gradient.linear(
  cmyk(red),
  cmyk(red).darken(70%),
  space: cmyk
))

#rect(fill: gradient.linear(
  cmyk(red),
  cmyk(red).darken(100%),
  space: cmyk
))

#rect(fill: gradient.linear(
  red,
  red.darken(70%),
  space: cmyk
))

#rect(fill: gradient.linear(
  red,
  red.darken(100%),
  space: cmyk
))

#rect(fill: gradient.linear(
  red,
  blue,
  space: cmyk
))
```

</details>

| Webapp | Chrome | Edge | vscode-pdf |
| --- | --- | --- | --- |
| ![image](https://github.com/typst/typst/assets/7191192/286a32af-2929-4c05-be93-0fc045b0a6cb) | ![image](https://github.com/typst/typst/assets/7191192/f61395ef-09b8-435b-acf0-a637e944469a) | ![image](https://github.com/typst/typst/assets/7191192/db28c4e3-e74a-4612-9d8d-5e34603c61ff) | ![image](https://github.com/typst/typst/assets/7191192/b32932aa-9e35-4c4f-9fcd-ef251074f702) |

### Operating system

Web app, Windows

### Typst version

- [X] I am using the latest version of Typst

EpicEricEE:
The problem seems to be that only the first three components are stored for each color in the shading function as we assume that the last component is always the alpha value, but in CMYK it's the black value:
https://github.com/typst/typst/blob/a2c980715958bc3fd71e1f0a5975fea3f5b63b85/crates/typst-pdf/src/color.rs#L391-L400

If I manually add a fourth component in the PDF file, the gradients look more correct (and show up in edge).

EpicEricEE:
I just found out that a very similar issue also affects luma gradients. The colors are also stored with three components (where all are the same)
```
<<
  /FunctionType 2
  /Range [0 1 0 1 0 1]  <--- /Range [0 1]
  /C0 [0 0 0]           <--- /C0 [0]
  /C1 [1 1 1]           <--- /C1 [1]
  /Domain [0 1]
  /N 1
>>
```
[...]

PR #4568, opened 2024-07-16
Adjust the number of color components written to the pdf according to the color space

FlorentCLMichel:
This is an attempt at resolving this issue: https://github.com/typst/typst/issues/4422 .
~As it stands, the proposed change may create more problems than it solves. (Mupdf now throws warnings when gradients in a non-CMYK color space are present, and the use of unsafe code should probably be avoided.)~

~I'm not sure how to resolve the issue in a cleaner way without making more substantial changes to the code, though, which I am definitely not familiar enough with the code-base for. 
Submitting it as a draft PR for the moment. If you think it could be useful after the above problems (and others that I may not have seen) are resolved, please let me know and I'll be happy to work more on it. If not, please feel free to close it.~

**Edit:**  The aforementioned problems were solved thanks to the help of @Dherse and @EpicEricEE .


Dherse:
Good catch!

crates/typst-pdf/src/color.rs
```diff
@@ -377,26 +377,37 @@ impl PaintEncode for Color {
...
     /// Converts a color to the color space.
-    fn convert<U: QuantizedColor>(self, color: Color) -> [U; 3];
+    fn convert<U: QuantizedColor>(self, color: Color) -> Vec<U>;
 }
 
 impl ColorSpaceExt for ColorSpace {
-    fn range(self) -> [f32; 6] {
-        [0.0, 1.0, 0.0, 1.0, 0.0, 1.0]
+    fn range(self) -> [f32; 8] {
+        [0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0] // Not sure about that
```
Dherse, line 388: This is correct, but I would prefer if the range was dependent on the color space, for example `fn range(self)` could return `&'static [f32]` instead of a fixed size array, that could already be an improvement as we don't want to waste space.

crates/typst-pdf/src/color.rs
```diff
@@ -377,26 +377,37 @@ impl PaintEncode for Color {
...
     }
 
-    fn convert<U: QuantizedColor>(self, color: Color) -> [U; 3] {
+    fn convert<U: QuantizedColor>(self, color: Color) -> Vec<U> {
```
Dherse, line 391: Using a `Vec` is undesirable here since it's used in many places, it should be a fixed sized array otherwise every single color being encoded will cause an allocation. Perhaps a `SmallVec<[U; 3]>` could be okay.

crates/typst-pdf/src/color.rs
```diff
@@ -377,26 +377,37 @@ impl PaintEncode for Color {
...
-            U::quantize(z, [range[4], range[5]]),
-        ]
+        match self {
+            ColorSpace::Cmyk => vec![
+                U::quantize(x, [range[0], range[1]]),
+                U::quantize(y, [range[2], range[3]]),
+                U::quantize(z, [range[4], range[5]]),
+                U::quantize(a, [range[6], range[7]]),
+            ],
+            ColorSpace::D65Gray => vec![
```
Dherse, line 402: This is a good change since right now it's wasting space for D65Gray

crates/typst-pdf/src/gradient.rs
```diff
@@ -146,8 +146,9 @@ pub fn write_gradients(
                         .bits_per_flag(8)
                         .shading_type(StreamShadingType::CoonsPatch)
                         .decode([
-                            0.0, 1.0, 0.0, 1.0, range[0], range[1], range[2], range[3],
-                            range[4], range[5],
+                            0.0, 1.0, 0.0, 1.0,
```
Dherse, line 149: This should therefore also be "programmable" otherwise it will waste file size

crates/typst-pdf/src/gradient.rs
```diff
@@ -338,19 +339,23 @@ fn register_gradient(
...
+///  - points: `[u16; 3]`
+///  - colors: `[u16; 3]`
 fn write_patch(
     target: &mut Vec<u8>,
     t: f32,
     t1: f32,
-    c0: [u16; 3],
-    c1: [u16; 3],
+    c0: &[u16],
+    c1: &[u16],
```
Dherse, line 358: This is a good change imo

crates/typst-pdf/src/gradient.rs
```diff
@@ -391,10 +396,16 @@ fn write_patch(
...
+        [c0.iter().map(u16_to_be),
+         c0.iter().map(u16_to_be),
+         c1.iter().map(u16_to_be),
+         c1.iter().map(u16_to_be)];
 
     // Push the colors.
-    target.extend_from_slice(bytemuck::cast_slice(&colors));
+    // TODO: find a better solution
+    unsafe {
+        target.extend_from_slice(colors.align_to::<u8>().1);
```
Dherse, line 407: Please do not do that, there is no reason for unsafe code here. It's better to iterate over the colors and extend for each component than this.

Dherse (changes requested):
I think overall the approach is good, my main points are:
- No need for unsafe code here, even though I agree it should be okay, it's just not needed
- We don't need to allocate for every color, we can optimize PDF file size slightly more by controlling the number of color elements every time

FlorentCLMichel:
Thanks a lot @Dherse for these helpful comments! I agree with all of them. 
I'll try to find some time to correct these in the next few days. (Or if someone wants to have a go, you're of course more than welcome to!)

crates/typst-pdf/src/gradient.rs
FlorentCLMichel: Agreed! I've just replaced this part to avoid the unsafe code. (To be improved.)

crates/typst-pdf/src/color.rs
FlorentCLMichel: Agreed, again. Not sure what the best option is as, if I' not mistaken, the number of elements depends on the color space. Or would it make sense to return a `[U: 4]` in all cases?·

crates/typst-pdf/src/color.rs
FlorentCLMichel: Tentative solution with `SmallVec`s (not sure what the impact on performance is, though): 

https://github.com/typst/typst/pull/4568/files#diff-55650bf2dd7b33ec02c37fd3b314eb3a5db0c2aa6bf0baf269ef9c03fe716c95L387-R394

```rust
    fn range(self) -> SmallVec<[f32; 8]> {
        match self {
            ColorSpace::Cmyk => smallvec![0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0],
            ColorSpace::D65Gray => smallvec![0.0, 1.0],
            _ => smallvec![0.0, 1.0, 0.0, 1.0, 0.0, 1.0],
        }
    }
```

crates/typst-pdf/src/color.rs
FlorentCLMichel: Tentative solution returning a `SmallVec<[U; 4]>`:

https://github.com/typst/typst/pull/4568/files#diff-55650bf2dd7b33ec02c37fd3b314eb3a5db0c2aa6bf0baf269ef9c03fe716c95L391-R413

Again, not about performance (I have very little experience with `SmallVec`s); please let me know if you think a different array size or return type would be better.

crates/typst-pdf/src/gradient.rs
FlorentCLMichel: Good point! Tentative solution with the parameter of `decode` having (if I'm not mistaken) the right size for each color space:

https://github.com/typst/typst/pull/4568/files#diff-6a20014eeaaa0cd7d097c8ead96193dfc6ed65aea4ed7ff2de0b5584b92dee78L143-R175

Not sure this is a good way to do it as it introduces quite a bit of code duplication. Maybe there is a way to move the `match` inside `decode`? I'll think more about it.

crates/typst-pdf/src/gradient.rs
FlorentCLMichel: New version without unsafe code:

https://github.com/typst/typst/pull/4568/files#diff-6a20014eeaaa0cd7d097c8ead96193dfc6ed65aea4ed7ff2de0b5584b92dee78L397-R429

```rust
    target.extend_from_slice(
        &c0.iter()
            .flat_map(|arg0: &u16| u16::to_be_bytes(*arg0))
            .cycle()
            .take(4 * c0.len())
            .chain(
                c1.iter()
                    .flat_map(|arg0: &u16| u16::to_be_bytes(*arg0))
                    .cycle()
                    .take(4 * c1.len()),
            )
            .collect::<SmallVec<[u8; 32]>>(),
    );
```

Like for the above changes, I don't have a good feel for what the performance of this is.

crates/typst-pdf/src/color.rs
```diff
@@ -377,26 +378,39 @@ impl PaintEncode for Color {
...
+                U::quantize(z, [range[4], range[5]]),
+                U::quantize(a, [range[6], range[7]]),
+            ],
+            ColorSpace::D65Gray => smallvec![U::quantize(x, [range[0], range[1]])],
+            _ => smallvec![
+                U::quantize(x, [range[0], range[1]]),
+                U::quantize(y, [range[2], range[3]]),
+                U::quantize(z, [range[4], range[5]]),
+            ],
+        }
```
EpicEricEE, line 413: Maybe you could use the result from the `range` function above to generalize this:
```rs
let ranges = self.range().chunks(2);
let components = self.encode(color);

ranges
    .zip(components)
    .map(|(range, component)| U::quantize(component, [range[0], range[1]]))
    .collect()
```

crates/typst-pdf/src/gradient.rs
```diff
@@ -145,10 +146,19 @@ pub fn write_gradients(
...
+                                range[3], range[4], range[5], range[6], range[7],
+                            ]),
+                            ColorSpace::D65Gray => {
+                                smallvec![0.0, 1.0, 0.0, 1.0, range[0], range[1]]
+                            }
+                            _ => smallvec![
+                                0.0, 1.0, 0.0, 1.0, range[0], range[1], range[2],
+                                range[3], range[4], range[5],
+                            ],
+                        })
```
EpicEricEE, line 161: `decode` takes an `IntoIterator<Item = f32>`, so you can simplify this a lot by just appending the range, which already only contains the correct number of components:
```rs
.decode([0.0, 1.0, 0.0, 1.0].into_iter().chain(range))
```

crates/typst-pdf/src/gradient.rs
```diff
@@ -343,14 +353,14 @@ fn register_gradient(
 ///
 /// Structure:
 ///  - flag: `u8`
-///  - points: `[u16; 24]`
-///  - colors: `[u16; 12]`
+///  - points: `[u16; 32]`
+///  - colors: `[u16; 16]`
```
EpicEricEE, line 357: I believe the points should stay the same, as there are 12 `[u16; 2]` arrays, so a total of 24 `u16` values. The data for the four written colors `[c0, c0, c1, c1]` depends on the number of components, as there aren't always four components.

```suggestion
///  - points: `[u16; 24]`
///  - colors: `[u16; 4*N]` (N = number of components)
```

crates/typst-pdf/src/gradient.rs
```diff
@@ -390,11 +400,21 @@ fn write_patch(
...
+            .cycle()
+            .take(4 * c0.len())
+            .chain(
+                c1.iter()
+                    .flat_map(|arg0: &u16| u16::to_be_bytes(*arg0))
+                    .cycle()
+                    .take(4 * c1.len()),
+            )
+            .collect::<SmallVec<[u8; 32]>>(),
+    );
```
EpicEricEE, line 417: Similar approach, but simpler. I'm not sure if the `collect` is really necessary as `Vec::extend` can also just consume the iterator, but maybe it has to then resize more often as I don't think that there's a size hint.
```rs
let colors = [c0, c0, c1, c1]
    .into_iter()
    .flat_map(|c| c.iter().copied().map(u16::to_be_bytes))
    .flatten()
    .collect::<SmallVec<[u8; 32]>>();

target.extend(colors);
```

EpicEricEE (commented):
Just a few suggestions for simplifying your code, mostly with some nice and handy iterator functions.

I might also suggest to use `&'static` slices for the `range` function, as it gets called relatively often and it doesn't make sense to recreate the `SmallVec` every time. I'm also not really familiar with the `smallvec` crate, so maybe it's also fine as it is.

crates/typst-pdf/src/color.rs
FlorentCLMichel: Quite neat! Thanks for the suggestion!

crates/typst-pdf/src/gradient.rs
FlorentCLMichel: That's a very nice simplification indeed! And I also like that, as for your previous suggestion, it allows in principle to add other color paces in the future without changing this part of the code. Thanks!

crates/typst-pdf/src/gradient.rs
FlorentCLMichel: Good point! Corrected in the latest commit.

crates/typst-pdf/src/gradient.rs
FlorentCLMichel: Nice simplification again! Thanks! 
I've tentatively used the version without `collect` as it seems to indeed not be necessary. Happy to add it if it can improve performances, though.

FlorentCLMichel:
Thanks a lot @EpicEricEE ! Your suggestions are super useful! 

I agree that using `&'static` slices for the `range` function would likely be better. I'm not sure what the best way to do that is. I tentatively implemented it as shown [here](https://github.com/typst/typst/pull/4568/files#diff-55650bf2dd7b33ec02c37fd3b314eb3a5db0c2aa6bf0baf269ef9c03fe716c95R388-R393) and added explicit conversions to `SmallVec` where a `[f32]` is expected instead of an `&[f32]` ([here](https://github.com/typst/typst/pull/4568/files#diff-6a20014eeaaa0cd7d097c8ead96193dfc6ed65aea4ed7ff2de0b5584b92dee78R150), [here](https://github.com/typst/typst/pull/4568/files#diff-6a20014eeaaa0cd7d097c8ead96193dfc6ed65aea4ed7ff2de0b5584b92dee78R222) and [here](https://github.com/typst/typst/pull/4568/files#diff-6a20014eeaaa0cd7d097c8ead96193dfc6ed65aea4ed7ff2de0b5584b92dee78R241)). I suspect there may be a better solution, though.

EpicEricEE:
I don't know if it's the best way, but you could do `range.iter().copied()` to get an iterator of owned `f32` values without creating a `SmallVec`. In all the cases here, this would be enough as the methods all take an `impl IntoIterator<Item = f32>`.

FlorentCLMichel:
Oh, nice! That's certainly much better than my version. Thanks!

crates/typst-pdf/src/color.rs
```diff
@@ -377,26 +378,29 @@ impl PaintEncode for Color {
 /// Extra color space functions.
 pub(super) trait ColorSpaceExt {
     /// Returns the range of the color space.
-    fn range(self) -> [f32; 6];
+    fn range(self) -> &'static [f32];
 
     /// Converts a color to the color space.
-    fn convert<U: QuantizedColor>(self, color: Color) -> [U; 3];
+    fn convert<U: QuantizedColor>(self, color: Color) -> SmallVec<[U; 4]>;
```
laurmaedje, line 384: If the vector is always of length `< 4`, then `ArrayVec` is a bit better than `SmallVec`

crates/typst-pdf/src/color.rs
```diff
@@ -377,26 +378,29 @@ impl PaintEncode for Color {
...
 }
 
 impl ColorSpaceExt for ColorSpace {
-    fn range(self) -> [f32; 6] {
-        [0.0, 1.0, 0.0, 1.0, 0.0, 1.0]
+    fn range(self) -> &'static [f32] {
+        match self {
+            ColorSpace::Cmyk => &[0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0],
+            ColorSpace::D65Gray => &[0.0, 1.0],
+            _ => &[0.0, 1.0, 0.0, 1.0, 0.0, 1.0],
```
reknih, line 392: Maybe it's actually better to do an exhaustive match here to prevent strife in the future

reknih (approved):
Looks good!

crates/typst-pdf/src/color.rs
FlorentCLMichel: Ah, you're right! Thanks for that!
I just [changed this line](https://github.com/typst/typst/pull/4568/files#diff-55650bf2dd7b33ec02c37fd3b314eb3a5db0c2aa6bf0baf269ef9c03fe716c95L383-R384) to return an `ArrayVec`.
It seems the `arrayvec` crate was not used, so I've added it [here](https://github.com/typst/typst/pull/4568/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R31).

laurmaedje: Yeah, that's fine. It's a transitive dependency already anyway.

crates/typst-pdf/src/color.rs
FlorentCLMichel: Good point! Thanks for that! I've changed [these lines](https://github.com/typst/typst/pull/4568/files#diff-55650bf2dd7b33ec02c37fd3b314eb3a5db0c2aa6bf0baf269ef9c03fe716c95R389-R397) to do an exhaustive search.

laurmaedje:
Thank you!

PR #4569, opened 2024-07-16
Let base be stretchable in width to accommodate wide attachments

Leedehai:


laurmaedje:
Looking at the new reference image, it may be that this is actually only desirable for a subset of symbols. For arrows it seems very sensible, but for `=` it looks weird.

cc: @Enivex due to https://github.com/typst/typst/issues/1304#issuecomment-1723261507

Enivex:
> Looking at the new reference image, it may be that this is actually only desirable for a subset of symbols. For arrows it seems very sensible, but for `=` it looks weird.
> 
> cc: @Enivex due to [#1304 (comment)](https://github.com/typst/typst/issues/1304#issuecomment-1723261507)

I'd argue that this should be opt-in regardless. Perhaps through something like `stretch()`.

Leedehai:
@laurmaedje Introduced `stretch()`; please see the reference images.

crates/typst/src/math/attach.rs
```diff
@@ -83,12 +89,19 @@ impl LayoutMath for Packed<AttachElem> {
             };
         }
 
+        let [t, b] = [layout!(t, sup_style_chain)?, layout!(b, sub_style_chain)?];
```
laurmaedje, line 92:
```suggestion
        let t = layout!(t, sup_style_chain)?;
        let b = layout!(b, sub_style_chain)?;
```

crates/typst/src/math/attach.rs
```diff
@@ -469,3 +514,33 @@ fn compute_shifts_up_and_down(
...
+    ctx: &mut MathContext,
+    styles: StyleChain,
+    stretch: StretchInfo,
+    relative_to_width: Abs,
+    fragment: MathFragment,
+) -> MathFragment {
+    let glyph = match &fragment {
+        MathFragment::Glyph(glyph) => glyph.clone(),
+        MathFragment::Variant(variant) => {
+            GlyphFragment::new(ctx, styles, variant.c, variant.span)
```
laurmaedje, line 528: when it this the case? could we run into a case where we try to stretch both horizontally & vertically with a mix or `lr` and `stretch`?

crates/typst/src/math/fragment.rs
```diff
@@ -231,6 +247,7 @@ pub struct GlyphFragment {
     pub dests: SmallVec<[Destination; 1]>,
     pub hidden: bool,
     pub limits: Limits,
+    pub stretch: Option<StretchInfo>,
```
laurmaedje, line 250: the proliferation of fields on this struct is really unfortunate

crates/typst/src/math/attach.rs
Leedehai: In a standalone `stretch(=, width: ...)`, the `MathFragment::Glyph` will be stretched and understandably, turned into a `MathFragment::Variant`. The same goes for `=^..._...`.

In `stretch(=, width: ...)^..._...` however, the construct of `stretch(=, width: ...)` will stretch the glyph and thus make it a `Variant`, and then the `^..._...` will also stretch it again. We can't combine this two stretching operations into one because of the order of parsing: `^..._...` is parsed _outside_ the node of `stretch(...)` call.

crates/typst/src/math/fragment.rs
Leedehai: Indeed, but necessary..

crates/typst/src/math/attach.rs
```diff
@@ -150,6 +164,44 @@ impl LayoutMath for Packed<PrimesElem> {
...
+    /// The width of the base to stretch to, relative to the maximum width of
+    /// the base and the attachments (if any).
+    pub width: Smart<Rel<Length>>,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct StretchInfo {
+    from: Abs,
+    to: Smart<Rel<Length>>,
+}
```
mkorje, line 189: If the change in `impl LayoutMath for Packed<AttachElem>` is made then we can get rid of this struct as only `to` needs to be passed around on the `GlyphFragment`. The `from` field will always just be the fragment's width then.

crates/typst/src/math/attach.rs
```diff
@@ -150,6 +164,44 @@ impl LayoutMath for Packed<PrimesElem> {
...
+
+impl LayoutMath for Packed<StretchElem> {
+    #[typst_macros::time(name = "math.stretch", span = self.span())]
+    fn layout_math(&self, ctx: &mut MathContext, styles: StyleChain) -> SourceResult<()> {
+        let mut fragment = ctx.layout_into_fragment(self.body(), styles)?;
+        let stretch = StretchInfo { from: fragment.width(), to: self.width(styles) };
+        fragment.set_stretch(Some(stretch));
```
mkorje, line 196: Is there any point in adding stretch information to the frame before we check if the fragment can actually be stretched?

crates/typst/src/math/attach.rs
```diff
@@ -83,12 +89,20 @@ impl LayoutMath for Packed<AttachElem> {
             };
         }
 
+        let t = layout!(t, sup_style_chain)?;
+        let b = layout!(b, sub_style_chain)?;
+        if let Some(stretch) = base.stretch() {
+            let relative_to_width =
+                measure!(t, width).max(measure!(b, width)).max(stretch.from);
```
mkorje, line 96: I think including `stretch.from` here makes interactions between `math.stretch` and `math.attach` a bit inconsistent. For example, a long attachment can cause the base to be stretched, but a shorter one doesn't cause the base to be compressed. If having `width: auto` with attachments means sizing the base relative to the attachments, then it should go both ways IMO.

```typst
#let base = math.stretch($=$, width: 250%)

$ // This gets shortened.
  stretch(base, width: #50%)
  // But this doesn't.
  stretch(base)^"a" $
```

crates/typst/src/math/attach.rs
```diff
@@ -469,3 +515,33 @@ fn compute_shifts_up_and_down(
...
     ('∫'..='∳').contains(&c) || ('⨋'..='⨜').contains(&c)
 }
+
+fn stretch_fragment(
+    ctx: &mut MathContext,
+    styles: StyleChain,
+    stretch: StretchInfo,
+    relative_to_width: Abs,
+    fragment: MathFragment,
+) -> MathFragment {
```
mkorje, line 525: I guess if we're passing in the fragment already, we may as well pass it with a mutable reference and instead not return anything.

mkorje (commented):
I think we should also add an element function to allow stretching vertically as well, and so we'd maybe have `math.hstretch` and `math.vstretch`. Whilst `math.lr` kind of does this, its really just for delimiters.

Adding an `axis` argument or something similar would make things a bit annoying I think - though maybe this could be done automatically by searching for the glyph in both the horizontal and vertical variants/constructions. I'm not sure if there exists any glyph that have both horizontal and vertical variants, which would cause problems then.

laurmaedje:
@mkorje Thanks for the review!

> I'm not sure if there exists any glyph that have both horizontal and vertical variants, which would cause problems then.

I can't really think of a glyph where that would make sense. I feel like the `hstretch`/`vstretch` naming doesn't feel Typst-y, so I'd prefer to avoid it.

mkorje:
Me neither, though I guess there's nothing stopping a font from including such things for a glyph. It wouldn't hurt then to have an `axis` parameter (or something like that) which is `auto`, just in case (but this is probably overkill).

And yeah I agree that it isn't very Typst like. Adding the word horizontal or vertical felt too verbose and I couldn't think of a nice shortening of both words.

Enivex:
Is there anything blocking this?

Leedehai:
Hi - I became a bit busy recently so I'm fine with someone taking this work and create a new PR (hence the drop in my recent PR counts). I could come back to this later if that's not already done. Also, is a 0.12 release imminent?

mkorje:
I'd be happy to continue on this PR's work. It'd be nice to get this into 0.12.

laurmaedje:
Closing in favor of https://github.com/typst/typst/pull/5030. Thanks all!

PR #4577, opened 2024-07-17
Bail out of reparse in `expr_with_paren` if we land on a different char

### LINKED ISSUE #4571: Panic when compiling invalid file
### Description

```
#import (hntle-clues: *
#import "/util.typ": qrlink
#let auton(
) = {
```

command
```
timeout -v 100 typst compile TEST___FILE.typ
```

cause this
```
thread 'main' panicked at crates/typst-syntax/src/parser.rs:1111:11:
assertion `left == right` failed
  left: End
 right: Eq
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

##### Automatic Fuzzer note, output status "Some(101)", output signal "None"
```

[compressed.zip](https://github.com/user-attachments/files/16260673/compressed.zip)

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

bluebear94:
Fixes #4571.

As I mentioned in [my comment](https://github.com/typst/typst/issues/4571#issuecomment-2233743137), I’m not sure if this is the solution we want to implement. Even then, we might want to be checking that the *positions* agree instead of checking that the next token is what we expect.

### DIFF
--- a/crates/typst-syntax/src/parser.rs
+++ b/crates/typst-syntax/src/parser.rs
@@ -1102,19 +1102,17 @@ fn expr_with_paren(p: &mut Parser, atomic: bool) {
     if p.at(SyntaxKind::Arrow) {
         p.restore(checkpoint);
         params(p);
-        if p.current() != SyntaxKind::Arrow {
+        if !p.expect(SyntaxKind::Arrow) {
             return;
         }
-        p.eat();
         code_expr(p);
         p.wrap(m, SyntaxKind::Closure);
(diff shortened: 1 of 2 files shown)


laurmaedje:
I think we could also just turn `p.assert` into `p.expect`. It will likely generate more parser errors, but it doesn't really matter since no humans will trigger this, only fuzzers or potentially bad actors.

laurmaedje:
LGTM. Feel free to undraft and I'll merge.

laurmaedje:
Thanks!

PR #4580, opened 2024-07-18
Add `non-zero` and `even-odd` fill rules to `path` and `polygon`

### LINKED ISSUE #3789: Add parameter of "evenodd" fill-rule to `path` function
### Description

In the document of `path` function, there is a following paragraph:

> Currently all paths are filled according to the non-zero winding rule.

However in SVG specification，two fill-rules **nonzero** and **evenodd** are supported. And the default value of fill-rule is **nonzero**.

Consider the two following SVG files, with different fill-rules:

`nonzero.svg`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100" height="100" viewBox="0 0 100 100">
    <polygon fill-rule="nonzero" stroke="red" points="50,0 21,90 98,35 2,35 79,90"/>
</svg>
```

`evenodd.svg`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100" height="100" viewBox="0 0 100 100">
    <polygon fill-rule="evenodd" stroke="red" points="50,0 21,90 98,35 2,35 79,90"/>
</svg>
```

Import them to a Typst file and the output result shows as below:
image

### Use Case

It would be nice if there is an "evenodd" fill-rule in `path` function, which is more consistent with the behavior of the SVG. And higher-level drawing packages like `CeTZ` would benefit from this.

HydroH:
Closes #3789 .
Corresponding tests were added and passed.


crates/typst/src/visualize/polygon.rs
```diff
@@ -32,11 +32,12 @@ pub struct PolygonElem {
...
     /// When setting a fill, the default stroke disappears. To create a
     /// rectangle with both fill and stroke, you have to configure both.
-    ///
-    /// Currently all polygons are filled according to the
-    /// [non-zero winding rule](https://en.wikipedia.org/wiki/Nonzero-rule).
     pub fill: Option<Paint>,
 
+    /// The rule used to fill the polygon.
+    /// Defaults to `non-zero`.
+    pub fill_rule: Smart<FillRule>,
```
EpicEricEE, line 39: I don't think this needs to be `Smart`. It can instead just provide a default value (probably non-zero)

EpicEricEE (commented):
This would also require an implementation in `typst-pdf` and `typst-svg` as the `typst-render` crate is only responsible for PNG export.

crates/typst/src/visualize/path.rs
```diff
@@ -33,11 +33,13 @@ pub struct PathElem {
...
     /// rectangle with both fill and stroke, you have to configure both.
-    ///
-    /// Currently all paths are filled according to the [non-zero winding
-    /// rule](https://en.wikipedia.org/wiki/Nonzero-rule).
     pub fill: Option<Paint>,
 
+    /// The rule used to fill the path.
+    /// Defaults to `non-zero`.
+    #[default(FillRule::default())]
+    pub fill_rule: FillRule,
```
EpicEricEE, line 41: As `FillRule` implements the `Default` trait, I think this can be shortened to just `#[default]`. Also, the note about the default value can be removed from the comment as it is added automatically to the docs with the attribute.

crates/typst/src/visualize/polygon.rs
```diff
@@ -32,11 +32,13 @@ pub struct PolygonElem {
...
     /// rectangle with both fill and stroke, you have to configure both.
-    ///
-    /// Currently all polygons are filled according to the
-    /// [non-zero winding rule](https://en.wikipedia.org/wiki/Nonzero-rule).
     pub fill: Option<Paint>,
 
+    /// The rule used to fill the polygon.
+    /// Defaults to `non-zero`.
+    #[default(FillRule::default())]
+    pub fill_rule: FillRule,
```
EpicEricEE, line 40: Same here

crates/typst/src/visualize/polygon.rs
```diff
@@ -72,6 +74,11 @@ impl PolygonElem {
         #[named]
         fill: Option<Option<Paint>>,
 
+        /// The rule used to fill the polygon.
+        #[named]
+        #[default(FillRule::default())]
+        fill_rule: FillRule,
```
EpicEricEE, line 80: Does the option even make sense here? Regular polygons are always "odd" everywhere so the two fill rule options would behave the same.

crates/typst/src/visualize/path.rs
```diff
@@ -147,14 +148,20 @@ fn layout_path(
 
     // Prepare fill and stroke.
     let fill = elem.fill(styles);
+    let fill_rule = elem.fill_rule.unwrap_or_default();
```
laurmaedje, line 151:
```suggestion
    let fill_rule = elem.fill_rule(styles);
```
otherwise, set rules won't work

crates/typst/src/visualize/polygon.rs
```diff
@@ -161,6 +162,7 @@ fn layout_polygon(
 
     // Prepare fill and stroke.
     let fill = elem.fill(styles);
+    let fill_rule = elem.fill_rule.unwrap_or_default();
```
laurmaedje, line 165: same here

laurmaedje:
Thank you!

PR #4581, opened 2024-07-18
Add `calc.norm()` function to compute euclidean norms

### LINKED ISSUE #4508: Add function to compute norm(s)
### Description

`calc.hypot(x, y) == calc.sqrt(x * x + y * y)` for floats and integers. For lengths, `calc.hypot(x, y)` would work similarly, keeping the unit.

### Use Case

Computing hypotenuses with lengths is currently very annoying because multiplying lengths is not allowed. This would solve this problem.

HydroH:
Closes #4508 .


bluebear94:
This should also be able to take lengths of the same type (e.g. `norm(2cm, 4in)` or `norm(3em, 1.2em)`).

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +909,31 @@ pub fn quo(
     Ok(floor(dividend.apply2(divisor.v, Div::div, Div::div)))
 }
 
+/// Calculates the euclidean norm of a sequence of numbers.
+///
+/// ```example
+/// #calc.norm(1, 2, -3, 0.5)
```
drupol, line 915: Would it be a good idea to either show that this norm is the Euclidian norm, either parametrize the `2` in the formula so we can compute other norms?

Raw ideas:
# calc.euclidian_norm(...) = calc.norm2(...) = calc.pnorm(2, ...)

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +909,31 @@ pub fn quo(
...
+pub fn norm(
+    /// The callsite span.
+    span: Span,
+    /// The sequence of numbers from which to calculate the norm.
+    /// Must not be empty.
+    #[variadic]
+    values: Vec<Spanned<Num>>,
+) -> SourceResult<f64> {
+    let mut sum = 0f64;
+    if values.is_empty() {
```
bluebear94, line 927: It doesn’t make sense to error on an empty parameter list; the result can be `0.0` in this case.

HydroH:
> This should also be able to take lengths of the same type (e.g. `norm(2cm, 4in)` or `norm(3em, 1.2em)`).

How should the result unit be determined in this scenario?

crates/typst/src/foundations/calc.rs
```diff
@@ -909,29 +911,61 @@ pub fn quo(
...
+                for Spanned { v, span } in values {
+                    match v {
+                        Value::Int(n) => sum += (n as f64).powi(2),
+                        Value::Float(n) => sum += n.powi(2),
+                        _ => bail!(span, "expected a number"),
+                    }
+                }
+                Ok(Value::Float(sum.sqrt()))
+            }
+            Value::Length(Length { em, .. }) if em.is_zero() => {
```
bluebear94, line 941: This function could give a hint to call `to-absolute` when it errors from mixing absolute and em units.

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +911,63 @@ pub fn quo(
...
+///
+/// ```example
+/// #calc.norm(1, 2, -3, 0.5)
+/// #calc.norm(1in, 2cm)
+/// #calc.norm(3em, 4em)
+/// ```
+#[func]
+pub fn norm(
+    /// The sequence of values from which to calculate the norm.
+    /// Must not be empty.
```
bluebear94, line 924: I think this comment can be changed now that empty parameter lists are allowed:

```suggestion
    /// Returns `0.0` if empty.
```

crates/typst/src/foundations/calc.rs
HydroH: Sorry, I'm not very familiar with the code base and not sure how to do it. Could you explain it?

drupol (commented):
great stuff!

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +912,86 @@ pub fn quo(
     Ok(floor(dividend.apply2(divisor.v, Div::div, Div::div)))
 }
 
+/// Calculates the euclidean norm of a sequence of values.
+///
+/// ```example
+/// #calc.norm2(1, 2, -3, 0.5)
+/// #calc.norm2(1in, 2cm)
+/// #calc.norm2(3em, 4em)
```
Coekjan, line 920: Seems some ` missing below this line

tests/suite/foundations/calc.typ
```diff
@@ -259,3 +259,30 @@
...
+// Error: 18-21 expected an em
+#calc.norm2(1em, 2in)
+
+--- calc-norm-expect-abs-or-em ---
+// Error: 13-20 expected an absolute length or em
+#calc.norm2(1cm+1em)
+
+--- calc-norm-expect-number-or-length ---
+// Error: 13-16 expected a number or length
+#calc.norm2(10%)
```
Coekjan, line 288: Most files end with a new line (as configured in `.editorconfig`), so it would be better to add a new empty line below this line.

crates/typst/src/foundations/calc.rs
Coekjan: Hints can be given through the `bail!` macro. You could search its definition and see the document for developers, which should help you do this.

HydroH: > Hints can be given through the `bail!` macro. You could search its definition and see the document for developers, which should help you do this.

Got it, thanks!

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +912,91 @@ pub fn quo(
...
+                Ok(Value::Float(sum.powf(1.0 / p.float())))
+            }
+            Value::Length(Length { em, .. }) if em.is_zero() => {
+                for Spanned { v, span } in values {
+                    match v {
+                        Value::Length(Length { abs, em }) if em.is_zero() => {
+                            sum += abs.to_raw().powf(p.float())
+                        }
+                        _ => {
+                            bail!(span, "expected an absolute length"; hint: "use `to-absolute()` to convert to an absolute length")
```
Coekjan, line 967: In the current code base, it is recommended to break such a long line into multiple lines. For example:

```suggestion
                            bail!(
                                span, "expected an absolute length";
                                hint: "use `to-absolute()` to convert to an absolute length"
                            )
```

crates/typst/src/foundations/calc.rs
```diff
@@ -4,10 +4,12 @@ use std::cmp;
...
 use std::ops::{Div, Rem};
 
+use crate::utils::Numeric;
+
 use crate::diag::{bail, At, SourceResult, StrResult};
 use crate::eval::ops;
 use crate::foundations::{cast, func, IntoValue, Module, Scope, Value};
-use crate::layout::{Angle, Fr, Length, Ratio};
+use crate::layout::{Abs, Angle, Em, Fr, Length, Ratio};
 use crate::syntax::{Span, Spanned};
```
Coekjan, line 13: I guess the maintainers would prefer this:

```rs
use crate::diag::{bail, At, SourceResult, StrResult};
use crate::eval::ops;
use crate::foundations::{cast, func, IntoValue, Module, Scope, Value};
use crate::layout::{Abs, Angle, Em, Fr, Length, Ratio};
use crate::syntax::{Span, Spanned};
use crate::utils::Numeric;
```

Enivex:
For consistency the names should be `norm2` and `normp` no? It looks strange when the number goes on the end of one and the beginning of the other. Not to mention it messes up alphabetical sorting.

That being said, I don't think the 2 is pulling its weight. I'd go with just `norm` and `normp`.

HydroH:
> For consistency the names should be `norm2` and `normp` no? It looks strange when the number goes on the end of one and the beginning of the other. Not to mention it messes up alphabetical sorting.
> 
> That being said, I don't think the 2 is pulling its weight. I'd go with just `norm` and `normp`.
> 
> Edit: I guess a named argument is off the table, considering everything in calc is positional? The most elegant version here would simple by a single `norm` function with a named parameter defaulting to 2.

I feel normp sounds weird, as the original term was supposed to be "p-norm".

Enivex:
> > For consistency the names should be `norm2` and `normp` no? It looks strange when the number goes on the end of one and the beginning of the other. Not to mention it messes up alphabetical sorting.
> > 
> > That being said, I don't think the 2 is pulling its weight. I'd go with just `norm` and `normp`.
> > 
> > Edit: I guess a named argument is off the table, considering everything in calc is positional? The most elegant version here would simple by a single `norm` function with a named parameter defaulting to 2.
> 
> I feel normp sounds weird, as the original term was supposed to be "p-norm".

No weirder than norm2

bluebear94:
I think I prefer a named argument for `p`.

Enivex:
I seem to remember a similar discussion for the root function. Something about not wanting a named parameter because of `set`.

Enivex:
I seem to remember a similar discussion for the root function. Something about not wanting a named parameter because of `set`.

laurmaedje:
I do not recall such a thing. Named parameters should be fine.

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +910,82 @@ pub fn quo(
...
+    #[named]
+    #[default(Num::Int(2))]
+    p: Num,
+    /// The sequence of values from which to calculate the p-norm.
+    /// Returns `0.0` if empty.
+    #[variadic]
+    values: Vec<Spanned<Value>>,
+) -> SourceResult<Value> {
+    let mut sum = 0.0;
+    if let Some(Spanned { v, span }) = values.first() {
```
laurmaedje, line 932: Flipping this `if let` into a `let .. else ` would make this a bit eaiser to read because of reduced indentation.

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +910,82 @@ pub fn quo(
...
+        match v {
+            Value::Int(_) | Value::Float(_) => {
```
laurmaedje, line 934: This function could benefit from an enum similar to `AngleLike`.

Enivex:
I don't know rust, so excuse me if I misread the code:

- You need to take absolute values before doing powers (unless p is an even integer).

- You should ensure that p > 0

- It would be useful to include the case p = inf as a special case, since it appears often. In the limit p -> inf the norm turns into the maximum of the absolute values of the entries.

crates/typst/src/foundations/calc.rs
HydroH: Any idea what it should be called?

laurmaedje: Maybe just something norm-specific like `Normable`? Maybe the maths folks have a better term for that.

HydroH:
> I don't know rust, so excuse me if I misread the code:
> 
> * You need to take absolute values before doing powers (unless p is an even integer).
> * You should ensure that p > 0
> * It would be useful to include the case p = inf as a special case, since it appears often. In the limit p -> inf the norm turns into the maximum of the absolute values of the entries.

Thanks for pointing out! I'm not very familiar to these mathematic functions so there could be some mistakes.
As for the p = inf case, it's a bit complicated, I will implement it later.

Enivex:
> > I do not recall such a thing. Named parameters should be fine.
> 
> Maybe I hallucinated it. Upon further inspection there are even existing functions with named parameters.

I finally figured out what I was thinking of. It was a discussion with @PgBiel about having an argument order option for `atan2` instead of splitting it into two functions.

crates/typst/src/foundations/calc.rs
Enivex: Normable is in fact used in math.

(Though in the sense that a topological vector space is normable if the topology can be induced by a norm.)

crates/typst/src/foundations/calc.rs
```diff
@@ -908,6 +910,87 @@ pub fn quo(
...
 }
 
+/// Calculates the p-norm of a sequence of values.
+///
+/// ```example
+/// #calc.norm(1, 2, -3, 0.5)
+/// #calc.norm(p: 3, 1in, 2cm)
+/// #calc.norm(3em, 4em)
+/// ```
+#[func(title = "P-Norm")]
```
MDLC01, line 920:
```suggestion
#[func(title = "𝑝-Norm")]
```
I think U+1D45D 𝑝 would be more appropriate here? At least I think it should be lowercase.

HydroH:
> > I don't know rust, so excuse me if I misread the code:
> > 
> > * You need to take absolute values before doing powers (unless p is an even integer).
> > * You should ensure that p > 0
> > * It would be useful to include the case p = inf as a special case, since it appears often. In the limit p -> inf the norm turns into the maximum of the absolute values of the entries.
> 
> Thanks for pointing out! I'm not very familiar to these mathematic functions so there could be some mistakes. As for the p = inf case, it requires some extra logic, I will implement it later.

Sorry for delaying such a long time, I got a little busy recently. The p = inf case has been implemented.

Enivex:
I don't know Rust, so I can't speak to the quality of the PR, but as long as you took my previous comments into account, then I'm happy.

laurmaedje:
I'm not super happy with the amount of manual abs and em juggling that is going on here. This is pretty much the only calc function that operates on lengths and it feels to me like we need to figure out a better story for that before proceeding.

Enivex:
> I'm not super happy with the amount of manual abs and em juggling that is going on here. This is pretty much the only calc function that operates on lengths and it feels to me like we need to figure out a better story for that before proceeding.

Would ripping out the operating on lengths part be a solution? At least short term.

laurmaedje:
We could do that.

MDLC01:
But the main point of this function is that it can be used on lengths. Without this, it becomes almost useless as it can be implemented trivially in Typst.

Enivex:
> But the main point of this function is that it can be used on lengths. Without this, it becomes almost useless as it can be implemented trivially in Typst.

I mean, you could then make the argument that plenty of other things are useless because they're trivial to implement. `calc.abs(x)` is just `calc.sqrt(x * x)`, and in fact `calc.sqrt(x)` is just `calc.pow(x,0.5)`

MDLC01:
Ok that's a fair point I agree with you now

Enivex:
> I've changed this to only supports floats (integers are automatically coerced into floats) and it always returns a float.
> 
> The behaviour now is (as before):
> - When p is infinity, it computes `max_x abs(x)`
> - When p is finite, it computes `(sum_x abs(x)^p)^(1/p)`
> 
> Is that correct? @Enivex 
> 
> Not sure whether it would make sense to support Decimals... cc: @PgBiel 

That's correct yes

PgBiel:
> - When p is finite, it computes `(sum_x abs(x)^p)^(1/p)`
> 
> (...)
> Not sure whether it would make sense to support Decimals... cc: @PgBiel 

Probably not, given the power to 1/p, which would make it imprecise (at least, that would be inconsistent with the current approach).

laurmaedje:
Alright, then let's merge. Thanks all!

Issue #4225, opened 2024-05-22
Font encoding issues for non-latin text


mhavos:
### Description

Compiling a pdf with non-latin (in this case specifically devanagari) text in it can sometimes result is strange text encoding. This results in text that is not properly selectable. The part of the font that is selectable differs across pdf viewers, and sometimes introduces non-existent characters into the selection.

Minimal Working Example:
```
#set text(font: "Siddhanta")

आ रु॒क्मैरा यु॒धा नर॑ ऋ॒ष्वा ऋ॒ष्टीर॑सृक्षत ।

अन्वे॑नाँ॒ अह॑ वि॒द्युतो॑ म॒रुतो॒ जज्झ॑तीरिव भा॒नुर॑र्त॒ त्मना॑ दि॒वः ॥
```
(explicitly states font for repeatability)
[typst output](https://github.com/typst/typst/files/15405414/devanagari.test.pdf)

This results in the following chunk of selectable text in Ubuntu's Document Viewer:
अा
ैरा युध
ा नर॑ ऋ॒ ष्वा ऋ॒ ीर॑ सृक्षत ꠰
अन्वे॑नाँ॒ अह॑ व॒ ुताे॑ म॒ ताे॒ जज्झ॑ती रव भा॒नुर॑त॒ त्मना॑ द॒वः ꠱

In other pdf viewers the output may be different, e.g. firefox gives:
अा ˳॒ƨैरा यु॒धा नर॑ ऋ॒ष्वा ऋ॒ʆीर॑सृक्षत ꠰
अन्वे॑ नाँ॒ अह॑ Vव॒ȭुताे॑ म॒˳ताे॒ जज्झ॑तीRरव भा॒नुर॑तA॒ त् मना॑ Tद॒वः ꠱
(notice the introduced latin characters)

It seems that this is not an inherent limitation of pdf itself, LuaLaTeX can generate properly functioning pdfs. A minimal working example is provided for LaTeX as well:
```
\documentclass[a4paper]{article}

\usepackage{fontspec}
\setromanfont{Siddhanta}

\begin{document}

आ रु॒क्मैरा यु॒धा नर॑ ऋ॒ष्वा ऋ॒ष्टीर॑सृक्षत ।

अन्वे॑नाँ॒ अह॑ वि॒द्युतो॑ म॒रुतो॒ जज्झ॑तीरव भनर॑र्त॒ त्मना॑ दि॒वः ॥

\end{document}
```
[...]

PR #4154, opened 2024-05-16
Refactor PDF export

elegaanz:
The PDF exporter now uses a type state to ensure that it is impossible to access context that is not yet available. It also makes extensive use of the `pdf_writer::Chunk::rewrite_into` API to have each step being more independent, and thus potentially easier to memoize.

The actual behavior of the crate didn't change a lot, everything is just less entangled and hopefully easier to reason about.

TODO:
- [x] adapt #4039 for this branch. I will wait for a first round of reviews to be ready before doing that, so that I can be sure of the design that will be adopted in the end and work from that.
- [x] lots of testing, because quite a lot of things have been moved around, and some parts may not work anymore even if what gets written is not supposed to have changed. The test suite passes, but it mostly catches that there are no duplicate object references, I can't really tell if the PDF is valid or not besides that. I've also run `qpdf --check` on all files generated by the test runner, and no errors were reported.

Issue #526, opened 2023-04-02
Different characters that use the same glyph result in the same character when copied from PDFs


ghost:
Input:
```typst
birth\u{ad}day
birth­day // also contains a soft hyphen
```

In the resulting PDF both [soft hyphens](https://en.wikipedia.org/wiki/Soft_hyphen) are searchable as spaces (`birth day`).
Expected result would be that the word is searchable as `birthday` (no soft hyphen included) or as `birth­day` (with soft hyphen included).
Including a space instead, changes the semantics of the text.

With LuaTeX, soft hyphens are not included in the searchable text.
Maybe related: https://github.com/typst/typst/issues/479

Enivex:
The same is true for zero width space.

TS60:
Maybe [this line](https://github.com/typst/typst/blob/23746ee18901e08852306f35639298ad234d3481/crates/typst/src/layout/inline/mod.rs#L531) is the reason for the problem.

TS60:
I could debug it as follows:
- The `write_normal_text` has the correct soft hyphen char for the soft hyphen glyph in the `glyph_set`
- The `create_cmap` function already has the normal space for a soft hyphen glyph in the `glyph_set`

Between these two something happens that replaces the text in the `glyph_set` for the soft hyphen glyph from the correct char to the normal space.

laurmaedje:
The problem is that harfbuzz outputs the same glyph for both U+20 (SPACE) and U+AD (SHY), just with different advance width. And since the PDF content is just glyphs + a glyph -> char mapping, it will copy the same char. I don't think we can fix this by changing `improve_glyph_sets`.

I guess we could try to filter out space glyphs with zero advance width or something like that. But I'm not sure whether that would be correct behaviour. Would need to look closer at what other software is doing.

TS60:
Thanks! Indeed it seems to vary with LuaTeX based on the font (not yet found the pattern). Maybe it works only if the font actually has a glyph for a soft hyphen. Otherwise, it just filters it out. It seems the harfbuzz flag is called `PRESERVE_DEFAULT_IGNORABLES` to activate this.

khaledhosny:
[...]

PR #4585, opened 2024-07-19
Use texts of the first occurrences for `/ToUnicode` CMap

### LINKED ISSUE #4582: `/ToUnicode` in PDF can be wrong if a glyph is mapped from multiple code points
### Description

Compile the following on typst.app, download PDF, and copy the text.

```typst
#set text(lang: "zh", region: "CN", font: "Noto Serif CJK SC")
力量
```

You will get 力量 (U+F98A U+F97E in CJK *Compatibility* Ideographs), instead of [力量](https://util.unicode.org/UnicodeJsps/confusables.jsp?a=%E5%8A%9B%E9%87%8F) (U+529B U+91CF in CJK *Unified* Ideographs).

CJK Compatibility Ideographs often look bizarre, because they fallback to the system font.

Screenshot of `notepad.exe`:
notepad screenshot

Continues #3416/#3435
Might relate to https://github.com/typst/webapp-issues/issues/48

## Cause

`/ToUnicode` in PDF maps glyphs to Unicode code points, making it possible to copy and search text.
However, sometimes **a glyph** in a font is shared by **multiple code points**.
There are 3 code points mapping to the glyph cid11384 in `SourceHanSerifCN-Regular.otf` (downloaded from [Region Specific Subset OTFs Simplified Chinese (简体中文)](https://github.com/adobe-fonts/source-han-serif/releases/download/2.002R/14_SourceHanSerifCN.zip)):

- ⼒ ([U+2F12 KANGXI RADICAL POWER](https://util.unicode.org/UnicodeJsps/character.jsp?a=2F12)): cid11384
- 力 ([U+529B CJK UNIFIED IDEOGRAPH-529B](https://util.unicode.org/UnicodeJsps/character.jsp?a=529B)): cid11384
- 力 ([U+F98A CJK COMPATIBILITY IDEOGRAPH-F98A](https://util.unicode.org/UnicodeJsps/character.jsp?a=F98A)): cid11384

<details>
<summary>(generated by a Python script)</summary>

```python
from unicodedata import name

from fontTools.ttLib import TTFont

font = TTFont("SourceHanSerifCN-Regular.otf")
cmap = font["cmap"].getBestCmap()

the_character = "\u529b"
the_glyph = cmap[ord(the_character)]

for c, g in cmap.items():
    if g == the_glyph:
        print(f"{chr(c)} (U+{c:X} {name(chr(c))}): {g}")

print("---")

for c in "⼒力力":
    print(f"{c} (U+{ord(c):X} {name(c)}): {ord(c) in cmap = }")
```

</details>

Only one code point can be written to `/ToUnicode`, and Typst chooses U+F98A 力 (CJK compatibility).

## Adobe’s implementation

Adobe explicitly maps the glyph to the most used one, U+529B 力 (CJK unified).

[To UVS, Or Not To UVS - CJK Type Blog | Adobe](https://ccjktype.fonts.adobe.com/2019/05/to-uvs-or-not-to-uvs.html):

> A ToUnicode mapping file does exactly what its name suggests: it maps CIDs to Unicode code points, or to code point sequences. Unlike CMap resources that map Unicode code points to CIDs, or 'cmap' tables that map code points to GIDs that may also be CIDs, a ToUnicode mapping file specifies the inverse mapping. Some omissions and ambiguities can arise, either because a glyph is represented as a sequence, or it is mapped from multiple code points.

> An excellent example of the latter is Adobe-Japan1-7 CID+1200, which is mapped from U+2F00 ⼀ KANGXI RADICAL ONE and U+4E00 一 (a CJK Unified Ideograph). If CID+1200 is included in a PDF, one would naturally expect U+4E00 一 to be copied, not U+2F00 ⼀ as its use is more obscure. The Adobe-Japan1-UCS2 ToUnicode mapping file **makes this mapping preference explicit** (04b0 is the zero-padded hexadecimal form of decimal 1200):
>
> ```
> <04b0> <4e00>
> ```

https://github.com/adobe-type-tools/cmap-resources/blob/f5cf3bca7fdfeaceb77aa82847e974f2306c20b4/Adobe-GB1-6/cid2code.txt#L27-L29:

> There may be cases of single CIDs being referenced in multiple encoding points within a single CMap file.
> These cases are comma-delimited, within the same column.

CMap

- Chinese (https://github.com/adobe-type-tools/cmap-resources/blob/f5cf3bca7fdfeaceb77aa82847e974f2306c20b4/Adobe-GB1-6/cid2code.txt#L2629C60-L2629C69):

  ```
  CID	…	UniGB-UCS2
  2543	…	2f12,529b
  ```

- Korean (https://github.com/adobe-type-tools/cmap-resources/blob/f5cf3bca7fdfeaceb77aa82847e974f2306c20b4/Adobe-KR-9/cid2code.txt#L14627C28-L14627C43):

  ```
  CID	…	UniAKR-UTF16
  14593	…	2f12,529b,f98a	
  ```

`/ToUnicode`

- Chinese (https://github.com/adobe-type-tools/mapping-resources-pdf/blob/2dd5e53fb74a01718b9dfd448a0d1cce6fff2aa5/pdf2unicode/Adobe-GB1-UCS2#L1862):

  ```
  <09ef> <529b>
  # <09ef> is 2543 in hex.
  ```

- Korean (https://github.com/adobe-type-tools/mapping-resources-pdf/blob/2dd5e53fb74a01718b9dfd448a0d1cce6fff2aa5/pdf2unicode/Adobe-KR-UCS2#L3525):

  ```
  <3901> <529b>
  # <3901> is 14593 in hex.
  ```

## Further example

```typst
#set page(height: auto, width: 30em)
#set text(lang: "zh", region: "CN", fallback: false)

#let fonts = (
  // "Linux Libertine", // no glyph
  // "Noto Serif", // no glyph
  "Noto Serif CJK SC",
  "Noto Sans CJK SC",
)

#for f in fonts {
  text(font: f)[
    #f

    - ⼒ U+2F12 Kangxi radical
    - 力 U+529B CJK unified
    - 力 U+F98A CJK compatibility

    地球发动机是人类建造的力量最大的机器，比如我们所在的华北794号，全功率运行时能向大地产生150亿吨的推力。

    #pagebreak()
  ]
}
```

Screenshot of the result in a nerd font:

> [!NOTE]
> This issue cannot be repoduced with `testit cjk --pdf` and https://github.com/typst/typst/blob/42754477886f6a12afbabfd2a64d8c787a57bc03/tests/suite/layout/inline/cjk.typ, because https://github.com/typst/typst-dev-assets/blob/48a924d9de82b631bc775124a69384c8d860db04/files/fonts/NotoSerifCJKsc-Regular.otf does not contain U+2F12 ⼒ (Kangxi radical) and U+F98A 力 (CJK compatibility).

### Reproduction URL

_No response_

### Operating system

Web app, Windows

### Typst version

YDX-2147483647:
Resolves #4582
…just by deleting `improve_glyph_sets`!

There are two sources of information for `/ToUnicode`: the `glyph_set` recorded while `write_text`, and `cmap` tables of the font.
`improve_glyph_sets` leverages the font. It was refactored into a function in #4154, but the real code even predates ed6550f (2 years ago).

`improve_glyph_sets` was necessary before ad34763, when a `glyph_set` was a list of glyphs, and we had to search the font (again) for their texts. (Each glyph represents a text, which is a Unicode code point or a sequence of code points (e.g. ligature).)

In ad34763, the `glyph_set` is refactored to a map from glyphs to texts.
Now we have enough information for `/ToUnicode` CMap—no need to search the font.

## Changes

If the glyph…
- …represents a single character…
    - …and is mapped from only one code point:
        No change.
    - …and is shared by multiple code points (e.g. CJK unified/compatibility):
        `/ToUnicode` changes from the largest code points to the first occurrence, and fixes #4582. 
- …represents a sequence of characters (e.g. ligature)…
    - …and they are also encoded as a single code point for compatibility (e.g “fi”/ﬁ):
        `/ToUnicode` changes from a single compatibility code point (ﬁ) to the sequence (fi).
        The behaviour in PDF viewers usually does not change.
    - …and is not encoded in Unicode (e.g. “Th” in Linux Libertine):
        No change.

## Testcase

(AFAIK, `/ToUnicode` is not included in `testit`.)

```typst
#set page(width: 30em, height: auto)

ffi Th ﬃ
#set text(ligatures: false)
ffi Th ﬃ

$integral_x y dif z$

#set text(lang: "zh", region: "CN", font: "Source Han Serif", fallback: false)

#let example-Han = [
  地球发动机是人类建造的力量最大的机器，比如我们所在的华北794号，全功率运行时能向大地产生150亿吨的推力。

  - ⼒ U+2F12 (Kangxi radical)
  - 力 U+529B (CJK unified)
  - 力 U+F98A (CJK compatibility)
]

#example-Han

#set text(fill: gradient.linear(..color.map.rainbow))
#example-Han

#set text(stroke: red)
#example-Han
```

## TODO

- [x] Remove `unicode_properties`. (still an indirect dependency through rustybuzz)

## Future work

- Write `/ActualText`. (tracked by #4225 and #526)
- ~~Use the most common instead of the first occurrence.~~ I don't think we should use CJK unified even if the author always writes CJK compatibility.


laurmaedje:
I agree that this makes sense. The reason I kept the cmap search in https://github.com/typst/typst/commit/ad347632ab95e29eb5180b27142f5c264dfc611a was probably just cautiousness.

laurmaedje:
Thank you!

PR #4587, opened 2024-07-20
Unified and fixed `Duration` formatting in the CLI

### LINKED ISSUE #3220: Make all value+unit of measurement uniform and correct
### Description

Currently, some units in the compiler's output are written with whitespace after the value, and others are smashed together with the value. The correct way is to leave a single whitespace between a value and its unit of measurement. The information size and transfer speed measurements do have whitespace.

Yes, a lot of apps skip the whitespace, but this is now what we were taught to do (in school); yes, it "looks better" without a whitespace, but this is only because it became "normal", if we use space everywhere, then it also will become "normal" and will "look better". See #2100 for more info.

I have found 2 places where units of time don't have a whitespace on the left side/prepended:

https://github.com/typst/typst/blob/6ac71eeaf7b68dab07f75bd1a480810481fa9b73/crates/typst-cli/src/download.rs#L217-L235

https://github.com/typst/typst/blob/6ac71eeaf7b68dab07f75bd1a480810481fa9b73/crates/typst-cli/src/watch.rs#L189-L192

But in order to fix this, I wanted to ask do we _really_ need "days" and "hours" (or even "minutes") for download times? For compile times, we only get seconds and milliseconds. Also, for downloads, the time format is manual (can be fixed with "1 char"), but for compilations, the time is auto-generated by using debug trait for `std::time::Duration`. I thought maybe we can uniform the both and use the same formatter function. Though everything looks localized and there is no `global_util_functions.rs` file, so I also don't really want to copy the same code twice, though "twice" is not critical especially if the function will be very small.

P.S. I know that it was implemented by different people.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

Andrew15-5:
This is just an example. Perhaps performance optimizations can be discussed besides the way things are formatted and named.

Fixes #3220.


Andrew15-5:
I don't know which style of testing is better. I rarely made tests and always struggle with "concise vs. readable vs. maintainable" type of thing.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,139 @@
...
+    let ms = format!(
+        "{1:.0$} ms",
+        if mcs != 0 { min(digits, mcs_digits) } else { 0 },
+        ms as f32 + mcs as f32 / 1000.0
+    );
+    match (d, h, m, s) {
+        (0, 0, 0, 0) => ms.to_string(),
+        (d, h, m, s) => format_dhms(d, h, m, s),
+    }
+}
```
PgBiel, line 55: Something like this seems better - see [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=afa07ceb3c758d96956cb9e8724c744a):

(note: there's more text below the suggestion)

```suggestion
/// Format string starting with number of milliseconds and going bigger from there.
/// `precision` is how many digits of microseconds from left will be preserved.
/// Note that this function will always remove all trailing zeros for microseconds.
fn time_starting_with_ms_with_precision(duration: &Duration, precision: u16) -> String {
    let (d, h, m, s, ms, mcs) = get_duration_parts(duration);
    match (d, h, m, s) {
        (0, 0, 0, 0) => {
            let ms = ms as f64 + mcs as f64 / 1000.0;
            let ms = (ms * 10.0_f64.powi(precision as i32)).round() / 10.0_f64.powi(precision as i32);
            format!("{ms} ms")
        }
        (d, h, m, s) => format_dhms(d, h, m, s),
    }
}
```

Although it may be a good idea to extract the 10^precision shenanigans into a separate `typst-utils` function, as this is also used for `repr`:

https://github.com/typst/typst/blob/d84a7e423b53e9b5740720270b132da60c8a3fea/crates/typst/src/foundations/repr.rs#L86-L89

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,139 @@
...
+
+fn time_as_ms(duration: &Duration) -> String {
+    time_starting_with_ms_with_precision(duration, 2)
+}
```
PgBiel, line 59: This can be inlined.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,139 @@
...
+    time_starting_with_seconds(duration)
+}
+
+pub fn eta_time(duration: &Duration) -> String {
+    time_starting_with_seconds(duration)
+}
+
+pub fn compilation_time(duration: &Duration) -> String {
+    time_as_ms(duration)
+}
```
PgBiel, line 71: I believe it should be fine to inline these functions into their respective usages. Makes more sense for the format choice to be specified where they're used instead of in a different module.

crates/typst-utils/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod fat;
 mod macros;
 mod bitset;
 mod deferred;
+pub mod format;
```
PgBiel, line 9: I think this should be moved up, near `fat`

PgBiel (commented):
Thank you! Here are some initial comments.

crates/typst-utils/src/format.rs
Andrew15-5: Like with `#[inline]` or the wrapper is unnecessary? Note that it's easier when all precisions are set in one compact file and only the wrappers without precision argument are used throughout the codebase. This way, you only have one place to look for if precision should be changed.

crates/typst-utils/src/format.rs
Andrew15-5: I know that it's "fine" to do that, I'm just thinking that it's easier to read and understand `eta = elapsed_time(elapsed_duration)` rather than using `time_starting_with_seconds`.

If you are sure that it won't decrease readability or increase maintainability or something, then I can remove these wrappers. I just don't think that they are that bad, but also just a function name alias isn't something that I see everyday.

crates/typst-utils/src/lib.rs
Andrew15-5: Ohh... Yeah, you're right.

crates/typst-utils/src/format.rs
PgBiel: > I know that it's "fine" to do that, I'm just thinking that it's easier to read and understand `eta = elapsed_time(elapsed_duration)` rather than using `time_starting_with_seconds`.

It could make sense if the functions were used more than once, but they're used only once, and in that sense it's not helpful as the name doesn't explain what's going on. It's something like `let thing = thing()` - it doesn't say much, just that the code for it is elsewhere.

Andrew15-5: Fine, I'll get rid of them.

crates/typst-utils/src/format.rs
PgBiel: I meant that the wrapper is unnecessary, yeah.

I'm not sure I agree with yourargument, given the function is only used once. No need for complexity ahead of time without necessity.

If the precision argument shouldn't be exposed, then `time_starting_with_ms_with_precision` shouldn't have the precision argument in the first place, as it is always 2.

Andrew15-5: > If the precision argument shouldn't be exposed, then `time_starting_with_ms_with_precision` shouldn't have the precision argument in the first place, as it is always 2.

But then the name of the function should be changed. Do I just hard-code the precision inside the function?

crates/typst-utils/src/format.rs
Andrew15-5: Ah, yes, the "measure, cut+round, reset" method. I just don't like multiplying by non-small numbers, worried about going over the limit. But the algorithm is very simple and gets the job done. This is at least a concern in the `repr.rs`.

crates/typst-utils/src/format.rs
PgBiel: Yeah. Feel free to rename it.

crates/typst-utils/src/format.rs
PgBiel: In principle there isn't much risk of overflow cuz float (at worst you get some infinity somewhere), though there are some potential mitigations that could be done if needed. However, we only use this with p = 2 right now so should be ok.

crates/typst-utils/src/format.rs
Andrew15-5: Just so you know, `compilation_time` is used twice and therefore inlining the base function leads to the usage of precision in 2 very close places. So I think that leaving either `time_starting_with_ms` or `compilation_time` wrapper for the function with a precision parameter isn't that useless already.

Andrew15-5:
One thing I'm not really sure is the order of `pub`/non-`pub` functions in https://github.com/typst/typst/pull/4587/files#diff-8b050b218f83192727292deff17a26bde7d134563d7a09d381f92f3a80c3b775. Other than that it looks more or less complete.

crates/typst-utils/src/format.rs
Andrew15-5: Oh, heh, it already triggered it [here](https://github.com/Andrew15-5/typst/actions/runs/10405315865/job/28815783456). I run all tests locally and didn't have any failed ones. That's kinda strange...

Andrew15-5: It was because `value` is already infinity. I think this can be an edge case. Now tests pass.

crates/typst-utils/src/format.rs
PgBiel: I'm ok with having a single public wrapper. Just multiple public wrappers with the ~same stuff is a bit too much for me.

Andrew15-5: Eh, I already changed to accepting it, since it used precision twice before: https://github.com/typst/typst/pull/4587/files#diff-ea95583ebe87e25ee323e9ad5d1a654f2fbf243e9f4d1727c49d7271585149bdL300-R310. If we get more than these 2 usages, then it is probably wise to add a wrapper. If they all use the same precition.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,136 @@
...
+/// Returns value with `n` digits after floating point where `n` is `precision`.
+/// Standard rounding rule applies (if `n+1`th digit >= 5 then round up).
+/// If `value` is +/- infinity returns `value`.
+///
+/// Will panic for a very large `value` and `precision` (if the `f64` limit is hit).
+/// In practice, it should never panic, otherwise the implementation have to be
+/// more robust.
+pub fn round_with_precision(value: f64, precision: u8) -> f64 {
+    let offset = 10_f64.powi(precision.into());
+    if [f64::NEG_INFINITY, f64::INFINITY].contains(&value) {
```
laurmaedje, line 14: I think this is just `value.is_infinite()`

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn test_time_as_ms_with_precision_2() {
+        let f = |duration| time_starting_with_ms_with_precision(&duration, 2);
+        let duration = duration_milli_micro;
+        assert_eq!("123.46 ms", &f(duration(123, 456)));
+        assert_eq!("123.46 ms", &f(duration(123, 455)));
+        assert_eq!("123.45 ms", &f(duration(123, 454)));
+        assert_eq!("123.45 ms", &f(duration(123, 450)));
+        assert_eq!("123.1 ms", &f(duration(123, 100)));
+        assert_eq!("123 ms", &f(duration(123, 0)));
+        // assert_eq!(format!("{:?}", &f(duration(123, 0))).as_str(), &f(duration(123, 0)));
```
laurmaedje, line 122: why is this commented?

crates/typst/src/foundations/repr.rs
```diff
@@ -1,6 +1,7 @@
 //! Debug representation of values.
 
 use ecow::{eco_format, EcoString};
+use typst_utils::format::round_with_precision;
```
laurmaedje, line 4: can use `crate::utils` here as well

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,136 @@
...
+/// more robust.
+pub fn round_with_precision(value: f64, precision: u8) -> f64 {
+    let offset = 10_f64.powi(precision.into());
+    if [f64::NEG_INFINITY, f64::INFINITY].contains(&value) {
+        return value;
+    }
+    assert!(
+        value * offset < f64::MAX,
+        "Hit the upper limit of f64 when rounding! ({value} * 10^{precision})"
+    );
```
laurmaedje, line 20: I think panicking here is problematic since this can be called from Typst formatting code.

crates/typst-cli/src/watch.rs
```diff
@@ -17,6 +17,7 @@ use crate::compile::compile_once;
 use crate::timings::Timer;
 use crate::world::{SystemWorld, WorldCreationError};
 use crate::{print_error, terminal};
+use typst_utils::format;
```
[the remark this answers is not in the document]
Andrew15-5: I didn't really know about that last bit. Is it because all separate crates are somehow re-exported in the `typst` crate?

crates/typst-cli/Cargo.toml
```diff
@@ -25,6 +25,7 @@ typst-pdf = { workspace = true }
 typst-render = { workspace = true }
 typst-svg = { workspace = true }
 typst-timing = { workspace = true }
+typst-utils = { workspace = true }
```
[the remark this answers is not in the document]
Andrew15-5: Ohh, yeah, this is neat. Makes sense now.

crates/typst-utils/src/format.rs
Andrew15-5: Thanks! I should try to remember exploring `.is_` more from now on in such situations.

crates/typst-utils/src/format.rs
Andrew15-5: I think I wanted to make another commit with the last change, probably regarding [this](https://github.com/typst/typst/pull/4587#issuecomment-2291389171). But currently yeah, it's not removed, and I remembered that after latest commit. I guess this can be counted as an oopsie.

crates/typst-utils/src/format.rs
Andrew15-5: What is "Typst formatting code"?

crates/typst-cli/src/watch.rs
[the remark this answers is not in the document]
laurmaedje: Not all, but the ones that are unconditionally part of the compiler.

crates/typst-utils/src/format.rs
laurmaedje: I meant `format_float` which is called by `repr`, so this could be hit by user code.

Andrew15-5: Yeah, I figured that one out already. https://github.com/typst/typst/pull/4587#discussion_r1718507312
Since the `.typ` tests failed. The point is that this assertion in practice never should be triggered. But if you want, I can try handling large numbers somehow, not yet sure how.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,135 @@
...
+use std::time::Duration;
+
+/// Returns value with `n` digits after floating point where `n` is `precision`.
+/// Standard rounding rule applies (if `n+1`th digit >= 5 then round up).
+/// If `value` is +/- infinity returns `value`.
+///
+/// Will panic for a very large `value` and `precision` (if the `f64` limit is hit).
+/// In practice, it should never panic, otherwise the implementation have to be
+/// more robust.
+pub fn round_with_precision(value: f64, precision: u8) -> f64 {
```
PgBiel, line 12: In principle I was going to suggest moving this function to `lib.rs`, but, on second thought (after writing my reply to the thread regarding the `assert!` call), I think this function should _only_ be used for formatting purposes, to the point that this should be reflected in the docstring and maybe even in the function's name somehow (possibly unnecessary).

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,135 @@
...
+    let offset = 10_f64.powi(precision.into());
+    if value.is_infinite() {
+        return value;
+    }
```
PgBiel, line 16:
```suggestion
    if value.is_infinite() {
        return value;
    }
    let offset = 10_f64.powi(precision.into());
```

crates/typst-utils/src/format.rs
PgBiel: I think we can explicitly mention in the function's docstring that accuracy isn't guaranteed for large enough numbers. In practice, this function is only used to display debugging or user-friendly information, so it doesn't matter too much if there's some imprecision here. A different function would have to be used for actual calculations (instead of just for string formatting purposes), as I see it.

crates/typst/src/foundations/repr.rs
```diff
@@ -1,6 +1,7 @@
 //! Debug representation of values.
 
 use ecow::{eco_format, EcoString};
+use typst::utils::format::round_with_precision;
```
PgBiel, line 4: This would probably be `crate::utils` instead (which would move this line under `use crate::foundations` below), though maybe we should just not import it at all and use the full path, since there's only one usage. Not sure what Laurenz thinks about this though.

PgBiel (commented):
Looking much better now!

crates/typst-utils/src/format.rs
laurmaedje: If the value is already huge, there is probably not float anyway that can represent it with the desired rounding. So can't we just return it as-is?

crates/typst/src/foundations/repr.rs
laurmaedje: Yes, should be crate. Whether to import it or not, I don't have a strong preference. Could also import the module instead of the function and qualify just that below. I think that's relatively common for function in Rust in comparison to types.

crates/typst-utils/src/format.rs
Andrew15-5: You can have numbers with 308 zeros. So I think that you definitely can have a huge number and still get precise rounding to some degree. Leaving the algorithm as is means that in some cases numbers will convert to infinity. Not great. Leaving huge numbers as is... Sounds like the function just gives up on it's job.

I just got an idea. Why not just separate integer and fractional parts and then perform the rounding. Then add the two and you're done. Last operation can affect precision, but that will be on the user, IMO. So as a result, no assertions required.

crates/typst-utils/src/format.rs
Andrew15-5: Exactly my thoughts! I was also unsure about it's placement at first. Since you would use it through `typst::format` I don't think you have to explain that this function is meant only to format a number. What do you want to add?

crates/typst-utils/src/format.rs
Andrew15-5: No fair! I thought of this first, but haven't fixed it because I have to go to sleep. I'll fix it.

crates/typst-utils/src/format.rs
laurmaedje: That might work as well. I just mean that I would always prefer the rounding to not happen rather than the assertion to trigger.

crates/typst/src/foundations/repr.rs
Andrew15-5: What about in download/watch files?

laurmaedje: There it's ok as-is.

crates/typst-utils/src/format.rs
Andrew15-5: Apparently in Rust `123.455 - 123 = 0.4549999999999983` which failed the test: https://github.com/Andrew15-5/typst/actions/runs/10420114994/job/28859494612. Tests already do their job, which is nice.

Andrew15-5: I asked rustaceans and the fix is to use other data types. Fortunately, adding [rust_decimal](https://crates.io/crates/rust_decimal) dependency not only solves the problem, but also adds only 1 crate. I see this as an absolute win!

PgBiel: I don't think the dependency pulls its weight here; as mentioned before, precision isn't important here. I'd rather follow Laurenz's approach of just returning the float as is if it's too large.

Do we know in which cases we can cause any sort of "overflow" in the first place? Would be nice to have some tests with really large floats first and see how the function behaves, and adjust from there.

Andrew15-5: I'll think about it later.

Andrew15-5: > Do we know in which cases we can cause any sort of "overflow" in the first place?

When `value * 10_f64.powi(precision.into())` `> f64::MAX` or `< f64::MIN`.

Andrew15-5: In the answer [here](https://stackoverflow.com/a/28656825) and from the mantissa, that is 52(+1) bits, the largest integer is 9e15. Going above will cause rounding and conversion error. So, if an input value is a concrete number without any errors, then numbers above 2^53 are none of our concern. Since you can't store fractional part for such big numbers, the "rounding with precision" is useless and therefore can be skipped by returning the input value.

This also means that for fractional, part you can only precisely represent up to `log10(2^53)` digits.

Now, even if you round integers to 2^60 and precision to 20, multiplication will still give you a finite number. If I'm not missing anything, this means that you don't need to explain any of this in the docs as it will be just a remix of IEEE 754 standard. Problem solved.

crates/typst-utils/src/format.rs
Andrew15-5: https://github.com/typst/typst/pull/4587#discussion_r1721075333

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Returns value with `n` digits after floating point where `n` is `precision`.
+/// Standard rounding rule applies (if `n+1`th digit >= 5 then round up).
+/// If `value` is +/- infinity or NaN returns `value`.
+pub fn round_with_precision(value: f64, precision: u8) -> f64 {
+    if value.is_infinite()
+        || value.is_nan()
+        // Integers that big can't have fractional part (mantissa is already full).
+        || value.abs() >= (1_i64 << 53) as f64
+        // Binary64 format can only precisely represent up to log_10(2^53) digits.
+        || precision >= 17
```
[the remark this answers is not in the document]
PgBiel: Grouping up the comments above the conditionals is also probably better for readability here:

```suggestion
    // Don't attempt to round the float if that wouldn't have any effect. This
    // includes infinite or NaN values, as well as values with a full mantissa
    // (which don't have a fractional part). Rounding with a precision larger
    // than the amount of digits that can be effectively represented would
    // also be a no-op.
    if value.is_infinite()
        || value.is_nan()
        || value.abs() >= (1_i64 << f64::MANTISSA_DIGITS) as f64
        || precision > f64::DIGITS
```

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,187 @@
...
+pub fn time_starting_with_seconds(duration: &Duration) -> String {
+    let (days, hours, minutes, seconds, _, _) = get_duration_parts(duration);
+    format_dhms(days, hours, minutes, seconds)
+}
+
+/// Format string starting with number of milliseconds and going bigger from there.
+/// `precision` is how many digits of microseconds from floating point to the right
+/// will be preserved. Note that this function will always remove all trailing zeros
+/// for microseconds.
+pub fn time_starting_with_ms_with_precision(
```
PgBiel, line 57: This function's name is quite long, and a bit misleading (precision of what? - only the documentation can clarify this). I think we can drop the `with_precision` part.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,187 @@
...
+        // Binary64 format can only precisely represent up to log_10(2^53) digits.
+        || precision >= 17
+    {
+        return value;
+    }
+    let offset = 10_f64.powi(precision.into());
+    if !(value * offset).is_finite() {
+        return value;
+    }
+    (value * offset).round() / offset
```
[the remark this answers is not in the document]
Andrew15-5: Hmm, I'm not sure we need this condition as it would be finite no matter what.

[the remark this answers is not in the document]
Andrew15-5: I'll add an assertion statement instead, as I'm pretty confident that it will stay finite. Otherwise, further modification is necessary.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Format string starting with number of days and going bigger from there.
+pub fn time_starting_with_seconds(duration: &Duration) -> String {
+    let (days, hours, minutes, seconds, _, _) = get_duration_parts(duration);
+    format_dhms(days, hours, minutes, seconds)
+}
+
+/// Format string starting with number of milliseconds and going bigger from there.
+/// `precision` is how many digits of microseconds from floating point to the right
+/// will be preserved. Note that this function will always remove all trailing zeros
+/// for microseconds.
```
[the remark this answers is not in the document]
Andrew15-5: The comment itself _is_ 80 chars long. Why does the `/// ` must be included? The Rust code is generally verbose, and various guidelines say to aim for 80 or 100. So if code in a file takes more horizontal space than 80 (plus the first statement), what this change is for? There is nothing about this (80) in the https://github.com/typst/typst/blob/489a6d54c193075d765606dafd562c27c894b04b/CONTRIBUTING.md either.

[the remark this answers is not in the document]
Andrew15-5: So... It should be documented in the contributing file?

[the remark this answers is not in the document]
Andrew15-5: It will shrink documentation that was exactly 80 chars to less, making it more squished and probably less readable. You know that, right? In IDEs that leave the docs as-is, this will be noticeable (e.g., Neovim).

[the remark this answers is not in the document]
PgBiel: Sure, we can add it to the CONTRIBUTING file.

But please adjust it.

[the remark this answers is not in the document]
PgBiel: Under normal circumstances, this would be done by `rustfmt` automatically, but unfortunately the relevant option[^1] is unstable.

[^1]: https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#wrap_comments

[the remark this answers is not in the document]
Andrew15-5: https://github.com/typst/typst/issues/4797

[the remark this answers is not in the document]
Andrew15-5: What about this one, then?
https://github.com/typst/typst/pull/4587/files#diff-8b050b218f83192727292deff17a26bde7d134563d7a09d381f92f3a80c3b775R25

It's exactly 80 chars + 4. (Or rather, 4 + 80.)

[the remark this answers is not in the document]
Andrew15-5: > Under normal circumstances, this would be done by `rustfmt` automatically, but unfortunately the relevant option is unstable, requiring nightly rustc.

Yeah, I remember searching for something like this. Sucks that some formatting features take ages to become stable.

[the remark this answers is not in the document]
PgBiel: > What about this one, then? https://github.com/typst/typst/pull/4587/files#diff-8b050b218f83192727292deff17a26bde7d134563d7a09d381f92f3a80c3b775R25
> 
> It's exactly 80 chars + 4. (Or rather, 4 + 80.)

The usual procedure here is to consider the \`raw\` part as a single block, so you'd send it to the next line.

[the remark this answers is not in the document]
Andrew15-5: Consider this trick: https://github.com/typst/typst/pull/4587#discussion_r1723619625.

So, a newline before the block and not inside of it?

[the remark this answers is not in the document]
Andrew15-5: I mean, you're not wrong, but if this wouldn't be an option, which one would you choose?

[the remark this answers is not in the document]
PgBiel: I'd move the whole raw to a newline if possible.

Andrew15-5:
I was feeling a bit fancy, so I added a few code examples to the public functions. They pass too.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Standard rounding rules apply (if `n+1`th digit >= 5, round up).
+///
+/// If rounding the `value` will have no effect (e.g., it's infinite or
+/// NaN), returns `value` unchanged.
+///
+/// ```
+/// # use typst_utils::format::round_with_precision;
+/// let rounded = round_with_precision(-0.56553, 2);
+/// assert_eq!(-0.57, rounded);
+/// ```
```
PgBiel, line 15:
```suggestion
/// If rounding the `value` will have no effect (e.g., it's infinite or
/// NaN), returns `value` unchanged.
///
/// # Examples
/// ```
/// # use typst_utils::format::round_with_precision;
/// let rounded = round_with_precision(-0.56553, 2);
/// assert_eq!(-0.57, rounded);
/// ```
```

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,231 @@
...
+        (0, 0, 0, s) => format!("{s:2} s"),
+        (0, 0, m, s) => format!("{m:2} m {s:2} s"),
+        (0, h, m, s) => format!("{h:2} h {m:2} m {s:2} s"),
+        (d, h, m, s) => format!("{d:3} d {h:2} h {m:2} m {s:2} s"),
+    }
+}
+
+/// Format string starting with number of seconds and going bigger from there.
+///
+/// ```
```
PgBiel, line 58:
```suggestion
/// Format string starting with number of seconds and going bigger from there.
///
/// # Examples
/// ```
```

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,231 @@
...
+    format_dhms(days, hours, minutes, seconds)
+}
+
+/// Format string starting with number of milliseconds and going bigger
+/// from there. `precision` is how many digits of microseconds
+/// from floating point to the right will be preserved (with rounding).
+/// Note that this function will always remove all trailing zeros
+/// for microseconds.
+///
+/// ```
```
PgBiel, line 83:
```suggestion
/// # Examples
/// ```
```

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,231 @@
...
+/// ), 2);
+/// assert_eq!("123 ms", &duration2);
+///
+/// let duration3 = time_starting_with_milliseconds(&Duration::from_secs(1), 2);
+/// assert_eq!(" 1 s", &duration3);
+/// ```
+///
+/// Note: if duration is 1 second or longer, then output will be identical
+/// to [time_starting_with_seconds], which also means that precision,
+/// number of milliseconds and microseconds will not be used.
```
PgBiel, line 103: Let's move this to a separate paragraph before the example

Andrew15-5: Yeah, but then there will be 2 _notes_.

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,237 @@
...
+    // effectively represented would also be a no-op.
+    if value.is_infinite()
+        || value.is_nan()
+        || value.abs() >= (1_i64 << f64::MANTISSA_DIGITS) as f64
+        || precision as u32 >= f64::DIGITS
+    {
+        return value;
+    }
+    let offset = 10_f64.powi(precision.into());
+    assert!((value * offset).is_finite(), "{value} * {offset} is not finite!");
```
PgBiel, line 32: Since this is called from user code (`repr`), it'd be better to not have this panic in practice, being there just for debug purposes if we ever find this through testing:

```suggestion
    debug_assert!((value * offset).is_finite(), "{value} * {offset} is not finite!");
```

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,237 @@
...
+    (value * offset).round() / offset
+}
+
+/// Returns `(days, hours, minutes, seconds, milliseconds, microseconds)`.
+fn get_duration_parts(duration: &Duration) -> (u16, u8, u8, u8, u16, u16) {
```
PgBiel, line 37: This can overflow with reasonably large days, so we should probably update this signature, and change this in other places taking days too (like `format_dhms` and others):

```suggestion
fn get_duration_parts(duration: &Duration) -> (u64, u8, u8, u8, u16, u16) {
```

crates/typst-utils/src/format.rs
```diff
@@ -0,0 +1,237 @@
+//! Different formatting functions.
+
+use std::time::Duration;
+
+/// Returns value with `n` digits after floating point where `n` is `precision`.
+/// Standard rounding rules apply (if `n+1`th digit >= 5, round up).
+///
+/// If rounding the `value` will have no effect (e.g., it's infinite or
+/// NaN), returns `value` unchanged.
```
PgBiel, line 9: Should reflow here to fill the 80 columns (it's possible this may be necessary elsewhere too)

```suggestion
/// If rounding the `value` will have no effect (e.g., it's infinite or NaN),
/// returns `value` unchanged.
```

PgBiel (commented):
Looking awesome, thank you! We're nearing full approval; just wanted to share some final observations.

crates/typst-utils/src/format.rs
Andrew15-5: But real testing only happens in production. How will we know if a user hits the assertions (that shouldn't be possible) if the assertion will be removed? Not everything must exist only during development or testing.

Fundamental assessments of the state of the program must always be true for it to work properly. If the assessment is false, then the state is corrupted and the logic is flawed.

crates/typst-utils/src/format.rs
Andrew15-5: I thought we're trying to make the app faster, not slower. :(

How can you wait for 180 years to compile?

I can only see a problem of a misuse of the function by feeding in a huge duration on purpose (not possible in the CLI) through the public functions. But that would be a misuse and can be caught with a panic.

So in any case, if the function is used as intended, the 16 bits is more than enough for a days counter.

crates/typst-utils/src/format.rs
Andrew15-5: I'll check later.

crates/typst-utils/src/format.rs
PgBiel: I understand, but displaying "infinity minutes" isn't fatal, just a bit ugly.

crates/typst-utils/src/format.rs
PgBiel: Yeah I doubt this would be triggered naturally, just wanted to have a type matching Duration's `from_days` for full accuracy. Doesn't hurt to have some extra caution here, especially because the function is public.

Andrew15-5: This also would require changing `format_dhms`'s signature. Do I still proceed?

crates/typst-utils/src/format.rs
Andrew15-5: Oh, you mean _fill_, I thought of breaking down at first. Yeah, I don't know how that one got like this, probably after some rephrasing.

[Here](https://github.com/typst/typst/blob/a29af26746618b94b54272722d29cd217a884509/crates/typst-utils/src/format.rs#L81-L89) I tried to not finish the lines with preposition, breaking visually the linked/dependent words. I think that's the only place where the text is not fully filled.

crates/typst-utils/src/format.rs
Andrew15-5: Displaying — _maybe_, but this means that the function failed to do what it was called for. **That** is fatal.

<details><summary>When rounding function returns infinity instead of a rounded number:</summary>

image

</details>

crates/typst-utils/src/format.rs
PgBiel: Yes

crates/typst-utils/src/format.rs
PgBiel: I spoke with Laurenz about this, and we decided that the assertion should be removed entirely. Instead, we should ensure through the checks at the top that there won't be addition with overflow. (Some form of "formal verification", not literally but some manual validation of this, could be nice here.)

Andrew15-5: We already [ensure it](https://github.com/typst/typst/blob/0bd3dd19108cae1dd481190325ea3d1b4f8919d5/crates/typst-utils/src/format.rs#L24-L27). The assertion is just a sanity check.

I don't understand the verification/validation part.

PgBiel: I meant ensure that it is logically impossible for there to be overflow. Then we should remove the assertion. Laurenz said that any possibility of user code panicking is a bug.

Andrew15-5: > Laurenz said that any possibility of user code panicking is a bug.

Well, duh, assertion provides the ability to know what the bug specifically is. It's almost the exact same thing as "this is a bug" error that was present not so long ago that I hit.

> it is logically impossible for there to be overflow.

It _is_ logically impossible, the assertion is _just_ a _sanity check_. I can remove it, if it's unacceptable. I'd rather have it than not. OK, I think we can agree on debug-only assertion then.

laurmaedje: If it is logically impossible it's fine.

Andrew15-5: You mean the `assert!()`?

laurmaedje: Yes

PgBiel (approved):
Thank you :blue_heart:

laurmaedje:
Thank you!

laurmaedje:
> Note: if duration is 1 second or longer, then output will be identical to [time_starting_with_seconds], which also means that precision, number of milliseconds and microseconds will not be used.

Why is it this way? Compile will now say 2s or 3s rather than an exact amount, which is not great. Surfaced here: https://discord.com/channels/1054443721975922748/1088371867913572452/1287441738239054036

Andrew15-5:
> Compile will now say 2s or 3s

Umm, it's akchuly "2 s" and "3 s", thank you very much. /hj

I actually don't remember, but at least for me, it makes sense that if it takes too long, then the precision doesn't matter. It's the same argument with size: the bigger the size, the less you care about the exact byte count. But with size, you always can represent it like `123.45 *B`. With user facing time formatting, you probably don't want/need to see `10.58 m` or `56.31 s`. But when it runs so fast that it takes less than a second, then you probably want to know a more precised value/output. It is also because you _have to_ show more precised value for small time period, otherwise it will all be rounded to 0 or 1 second.

@Dherse, I'm happy to change the precision aspect of the new duration output. I'm guessing there should be a 3rd version where it shows `59.99 s` when it's bigger than 1 s and smaller than 60 s. Or a completely new logic of how to format duration depending on its value. Perhaps a new "precision" flag or something for those that needs those precious nanoseconds. But it is probably should be discussed in a new issue.

laurmaedje:
I think two digits are good. It's just useful when comparing different typst versions for performance.

Andrew15-5:
Does this mean I can already start making the PR? Are we not going to create a separate issue?

laurmaedje:
No need for an issue.

laurmaedje:
I don't think we need anything complex. Just the minimal change to make it show two digits again.

Andrew15-5:
Should 2 digits be added even when duration is longer than 59 s?

laurmaedje:
I think yes. It can be useful and it doesn't hurt.

laurmaedje:
I did it myself now, since the lack of precise timings would block the release otherwise. See https://github.com/typst/typst/pull/5082.

Andrew15-5:
Wow, you did more than just add precision to seconds when duration ≥ 1 s. If I had to do this, this would've probably also blocked the release, hehe.

Andrew15-5:
As to blocking, I'm busy with a lot of stuff, including Typst Forum. But perhaps I will have more time for source code contribution soon (maybe even tomorrow).

PR #4602, opened 2024-07-23
Support multiple labels on content

### LINKED ISSUE #4592: Having multiple labels on an object ignores all but the last silently
### Description

I was told in the Discord to report this as a bug.

Trying to give an object multiple labels compiles fine, however only the last label is actually added in the document.
```typst
= Hello <a> <b>

@b
//@a // Errors with "Label <a> does not exist in the document"
```

It would be nice if multiple lables on an object were possible, but even if they aren't, a compiler warning (or even error) should be issued rather than ignoring all but the last silently.

### Reproduction URL

https://typst.app/project/ror0deDxAMdfJOfgI8fXoj

### Operating system

Web app

### Typst version

bluebear94:
Fixes #4592.

Support attaching multiple labels on a piece of content.

Multiple labels are not yet exposed to the Typst side; currently, `content.label` will only return the first label attached to `content`. We could expose a `labels` field on `content` that returns a list of all labels attached, deprecating `label`, but this exposes extra complexity to package authors, so I’m not sure that this whole thing is worth the trouble. Also, if we decide that labels should translate to `id`s in HTML export, each tag is limited to one `id`.


laurmaedje:
Semantically, labels are intended like HTML's ids. I'm not sure whether supporting multiple is good.

laurmaedje:
Independently of whether it is a good change or not, it's such a foundational change that I think it would require more discussion before the PR stage. So I'll close this for now. Thanks still!

PR #4609, opened 2024-07-24
Add Latin translation

### LINKED ISSUE #4595: Latin translations are missing
### Description

Using the same names as the LaTeX package [babel-latin](https://ctan.org/pkg/babel-latin?lang=en) (except for "Index capitum" instead of "Index") and using [*exemplum*](http://www.perseus.tufts.edu/hopper/text?doc=Perseus:text:1999.04.0059:entry=exemplum) for raw text, one could create a file `translations/la.txt` with the following contents:

```
figure = Descriptio
table = Tabula
equation = Equatio
bibliography = Conspectus librorum
heading = Sectio
outline = Index capitum
raw = Exemplum
```

### Use Case

```
#set text(lang: "la")
```

ssotoen:
Fixes #4595.

Also adds smartquotes for Italian.


sannator:
You should also update the file `crates/typst/src/text/lang.rs`, see this previous pull request https://github.com/typst/typst/pull/4414

EpicEricEE:
The LaTeX package `csquotes` seems to use "italian-style" outwards pointing guillemets «…» (and “…” for secondary) as quotation marks by default for latin[^1]. I don't know any latin but maybe we want to adapt that (and also add the italian ones).

[^1]: https://texdoc.org/serve/csquotes/0, pages 3 and 4

ssotoen:
Good point. Italian doesn’t have smart quotes defined either so I’ll add those too.

The babel-latin package puts 1⁄12 quad spacing before certain punctuation marks including quotes for the ecclesiastic variant.[^1] Apart from the quotes Typst doesn’t even do that for French so I think I’ll include that as an alternative using the narrow no-break space.

Latin should also have its own hyphenation rules (babel-latin has three different sets for its variants).[^2]

[^1]: https://mirrors.ctan.org/macros/latex/contrib/babel-contrib/latin/latin.pdf, p 3–4
[^2]: https://mirrors.ctan.org/macros/latex/contrib/babel-contrib/latin/latin.pdf, p 7

laurmaedje:
Thanks!

PR #4617, opened 2024-07-25
Replace `once_cell`'s `Lazy` as much as possible

asibahi:
Excited about `LazyCell` and `LazyLock` making it into `std` , I worked myself into replacing it in typst. I bumped the MSRV as well but it is failing the CI tests. Unsure why.


.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.77.0
+      - uses: dtolnay/rust-toolchain@1.80.0
```
PgBiel, line 56: for CI to pass, you'll need to bump this on the other jobs too

laurmaedje:
Since LazyLock is just a small niceness of not needing once_cell, I'd like to not bump the MSRV for now (we just recently bumped from 1.74 to 1.77). I think this reduces churn for people that aren't always on the very latest toolchain, and also helps with some package managers that aren't always 100% up to date. But I'm open to discussion if others think it's pointless. Intuitively, my instinct would be to stay 1-3 versions behind, with the exception being that a _really_ nice feature comes along.

laurmaedje:
I'd be fine with bumping the MSRV to 1.80 now, since we're at 1.82.

laurmaedje:
Thanks!

PR #4619, opened 2024-07-25
Export `ty` and `scope` macros before mod decls to sidestep a rust-analyzer issue

Leedehai:
While we are waiting on a fix to address https://github.com/rust-lang/rust-analyzer/issues/17630, this small edit makes rust-analyzer happy.

Tested on VSCode with rust-analyzer [v0.3.2045](https://github.com/rust-lang/rust-analyzer/releases/tag/2024-07-22).


laurmaedje:
Thanks!

PR #4631, opened 2024-07-26
Fixes counter stepping behavior

### LINKED ISSUE #4626: Stepping a `(0,)` counter at a deep level also steps previous levels
### Description

When a counter is at `(0,)`, and you step it at a deep level (say 5), then it goes to `(0,1,1,1,1)`.

I consider this a bug, because my understanding is that every counter should be interpreted with trailing zeros, so `(0,)` should actually mean `(0,0,0,0,0,0,...)`. Therefore, I would expect that when I step `(0,)` at level 5 it would go to `(0,0,0,0,1)`.

For example, consider headings:

```typ

==== First heading

For example, right now you have the weird behavior that if the first heading is super deep, then the number will start with a zero, but all the slightly deeper numbers are a 1.

I would expect this section to be 0.0.0.1

== second level heading

=== third level heading

==== fourth level heading

and I expect this to be 0.1.1.1, but it isn't
```

Just to clarify again how I would expect that a counter works:

- every counter should be interpreted with trailing zeros. so e.g. (0,) should mean (0,0,0,0,...)
  (just a computer has finite memory so we omit trailing zeros)
- then a step at level `n` should be done this way:
  * increase the `n`th number of the counter by 1
  * make all numbers at deeper levels equal to 0
  * (in particular don't do _anything_ to numbers at higher levels)

### Typst version

jbirnick:
Fixes https://github.com/typst/typst/issues/4626

This fixes how the counter is stepped.

Accordingly, also the `outline` test had to be updated. (look at the difference in the "Deep Stuff" heading)


Enivex:
Could you make the title more descriptive?

laurmaedje:
Yeah, I think this makes sense. LaTeX also behaves like your version. I added a more direct regression test.

laurmaedje:
Thank you!

PR #4634, opened 2024-07-27
Fix alignment of gradients and patterns on strokes in PNG

### LINKED ISSUE #4621: Gradient offset on stroke when relative to parent
### Description

Consider a circle centered in a box, both having a fill or stroke of a sharp linear gradient. The fill and stroke gradients of the circle are relative to the parent (the box) and should thus align with the fill gradient of the box. However, only the fill gradient of the circle is aligned, the stroke gradient is offset by an amount seemingly depending on the stroke's thickness.

This only happens with PNG export.

```typ
#box(
  width: 8em,
  height: 8em,
  fill: gradient.linear(black, yellow).sharp(2),

  align(center + horizon, circle(
    radius: 1em,
    stroke: 5pt + gradient.linear(red, blue, relative: "parent").sharp(2),
    fill: gradient.linear(yellow, black, relative: "parent").sharp(2)
  ))
)
```

| PNG / Web app preview | PDF / SVG |
| --- | --- |
| image | image |

(Tested on main and on v0.11.1)

### Reproduction URL

_No response_

### Operating system

Web app, Windows

### Typst version

EpicEricEE:
Fixes #4621: Gradients relative to "parent" were mistakenly offset to adjust for the stroke thickness. This is only necessary when they are relative to "self".

While trying to fix this, I found and fixed another similar issue: Patterns relative to "self" were offset, because their placement was _not_ adjusted for the stroke thickness. I reused the offset from the `gradient_map` argument to move the pattern to the correct location.

The difference between the old and new behavior can be seen in a0c1090, where the test reference images were updated. The new behaviour now matches the PDF and SVG output.

I changed the "pattern-stroke" test because the previous stroke thickness happened to cancel out the missing offset.


EpicEricEE:
I don't know why the tests are failing, they pass on my windows machine

laurmaedje:
Thank you!

PR #4636, opened 2024-07-27
Add timezone to PDF's default timestamp and consolidate logic

### LINKED ISSUE #3948: Saving Date Metadata with Time-Zone instead of Auto-Converting to UTC
### Description

Typst appears to be converting date metadata to UTC ("Zulu", "Z", UTC+00:00) Time.

Example of this Metadata:
`<</Creator(Typst 0.11.0)/CreationDate(D:20240417105515Z)/ModDate(D:20240417105515Z)>>`

While this is the correct time, the information in which time zone the PDF was originally created gets lost. 

Preferably, it would be saved as:
`<</Creator(Typst 0.11.0)/CreationDate(D:20240417125515+02'00')/ModDate(D:20240417125515+02'00')>>`

Now we know that the PDF was created in UTC+2 Time Zone. 

### Use Case

When converting to UTC time, we lose important information, when we are then trying to figure out when a PDF was created, we need to know where we were and also account for daylight saving times etc.... 

Saving the Time Zone would make the metadata much clearer, because then it would be unmistakably when the PDF was created. As someone that is often in different time zones, it is currently very annoying to find the local time I created a document.

Concern:
An unknowing user could give away location information through his timezone. But if someone is concerned about these points, everybody is free to disable the automatic creation and modification date in Typst.

Welcomed side effect:

Some PDF Viewers seem to handle metadata different from others, converting to local time zone, or just stripping the time zone info away no matter what. When having the time zone info, we are later able to validate what some viewers show and also know in which time zones the PDF was created, offering more valuable metadata than without.

Microsoft Edge is bad at displaying the correct time info, because it just strips away any time zone info. Let's assume the reader and creator are both in UTC+2.
So, when something was created in UTC+2 with Typst and then saved in "zulu" time, Edge will display a time 2hrs earlier. But when it is saved with the time zone info. The time in Edge will be the correct one, because Edge just stripped away the time zone info. This of course will break when reader and viewer are in different time zones.

Leedehai:
TL;DR: Fixes #3948, i.e. let the default timestamp include a timezone suffix.

## The instant in time

The PDF's timestamp represents an absolute instant in time, unrelated to the choice of timezone.

Though not called for by issue #3948, this PR consolidates (but does _not_ change) the logic of deciding the PDF timestamps in `make_pdf_timestamp()`. Originally, the decision relies on a series of fallback decision-making in multiple files, across the `typst-cli` and `typst-pdf` crates, and made it challenging to reason about.

In summary, the fallback decisions follow this precedence, which is preserved by this PR:
* (a) respect [`document.date`](https://typst.app/docs/reference/model/document/#parameters-date) (either a `datetime(...)` or `none`),
* (b) if (a) is the default value `auto`, the UNIX timestamp provided through the CLI command,
* (c) if (b) is absent, the current time.

## The timezone

The same instant in time has multiple representations depending on the choice of timezone. This PR changes the timezone as requested by issue #3948.

| | Status quo | After this PR |
|:--|:--|:--|
|`#set document(date: none)`| no timestamp | same |
|`#set document(date: datetime(...))`| no timezone suffix<br>e.g. `20200101` | same |
|`--creation-timestamp=(UNIX time)`| with a UTC suffix<br>e.g. `20230101050000Z` | same |
|Default (i.e. use the current time)| with a UTC suffix<br>e.g. `20240727195902Z` | with the local timezone suffix<br>e.g. `20240727150651-04'00` |

## Testing

(note: I could use `hexdump` but it's hard to `grep` the results, so I resorted to `strings`, which is also available on most *nix systems)

### (1) Timestamp disabled by `#set document(date: none)`
```typ
// test.typ
#set document(date: none)
PDF
```
`strings test.pdf` shows there's no creation or modification timestamp in the generated PDF's binary file.

### (2) Timestamp specified by `#set document(date: datetime(...))`
```typ
// test.typ
#set document(date: datetime(year: 2020, month: 1, day: 1))
PDF
```
`strings test.pdf | grep Date` shows
```txt
  /CreationDate (D:20200101)
  /ModDate (D:20200101)
...<xmp:CreateDate>2020-01-01</xmp:CreateDate><xmp:ModifyDate>2020-01-01</xmp:ModifyDate>...
```

### (3) Timestamp specified by CLI `--creation-timestamp=1672549200`
i.e. Sun Jan 01 2023 05:00:00 GMT+0000
```typ
// test.typ
PDF
```
`strings test.pdf | grep Date` shows
```txt
  /CreationDate (D:20230101050000Z)
  /ModDate (D:20230101050000Z)
...<xmp:CreateDate>2023-01-01T05:00:00Z</xmp:CreateDate><xmp:ModifyDate>2023-01-01T05:00:00Z</xmp:ModifyDate>...
```

### (4) Default (i.e. `document.date` is `auto`)
```typ
// test.typ
PDF
```
`strings test.pdf | grep Date` shows
```txt
  /CreationDate (D:20240727150651-04'00)
  /ModDate (D:20240727150651-04'00)
...<xmp:CreateDate>2024-07-27T15:06:51-04:00</xmp:CreateDate><xmp:ModifyDate>2024-07-27T15:06:51-04:00</xmp:ModifyDate>...
```

### DIFF
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -190,13 +190,13 @@ fn export_pdf(document: &Document, command: &CompileCommand) -> StrResult<()> {
 /// timestamps do not have timezone; the datetime from the CLI command is
 /// parsed from a UNIX timestamp and it assumed the UTC timezone.
 fn make_pdf_timestamp(
-    document_date_config: Smart<Option<Datetime>>,
-    command_override: Option<chrono::DateTime<chrono::Utc>>,
+    document_datetime: Smart<Option<Datetime>>,
+    command_datetime: Option<chrono::DateTime<chrono::Utc>>,
 ) -> Option<Timestamp> {
-    match document_date_config {
+    match document_datetime {
         Smart::Custom(None) => None,
         Smart::Custom(Some(datetime)) => Some(Timestamp::new(datetime)),
-        Smart::Auto => match command_override.and_then(chrono_to_datetime) {
+        Smart::Auto => match command_datetime.and_then(chrono_to_datetime) {
             Some(datetime) => Timestamp::new(datetime).with_timezone_offset(0),
             None => {
                 let now = chrono::Local::now();
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -14,6 +14,7 @@ use typst::layout::{Frame, Page, PageRanges};
 use typst::model::Document;
 use typst::syntax::{FileId, Source, Span};
 use typst::WorldExt;
+use typst_pdf::Timestamp;
 
 use crate::args::{
     CompileCommand, DiagnosticFormat, Input, Output, OutputFormat, PageRangeArgument,
@@ -164,11 +165,14 @@ fn export(
 
 /// Export to a PDF.
 fn export_pdf(document: &Document, command: &CompileCommand) -> StrResult<()> {
-    let timestamp = convert_datetime(
-        command.common.creation_timestamp.unwrap_or_else(chrono::Utc::now),
+    let pdf_timestamp =
+        make_pdf_timestamp(document.date, command.common.creation_timestamp);
+    let buffer = typst_pdf::pdf(
+        document,
+        Smart::Auto,
+        pdf_timestamp,
+        command.exported_page_ranges(),
     );
-    let exported_page_ranges = command.exported_page_ranges();
-    let buffer = typst_pdf::pdf(document, Smart::Auto, timestamp, exported_page_ranges);
     command
         .output()
         .write(&buffer)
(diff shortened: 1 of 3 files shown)


laurmaedje:
First of all: Sorry for the late reply.

I am a bit concerned by moving all of the logic from typst-pdf into typst-cli. In my opinion, as much of the logic as possible should reside in typst-pdf because then other consumers of this crate (like LSPs) automatically do the right thing instead of having slightly differing behaviour.

Leedehai:
np, without looking into the code closely I think I can adjusted where the logic is placed. I'll address this in a few days.

Leedehai:
np, without looking into the code closely I think I can adjust where the logic is placed. I'll address this in a few days.

laurmaedje:
I'll close this for now due to inactivity. We can always reopen or have a new PR.

PR #3869, opened 2024-04-04
Implement under/over parens/shells, refactor code

frozolotl:
I decided to use the word `shell` instead of turtle or tortoise because I think it is a bit clearer, the word reads nicer (“overshell” vs “overturtle”), and there is no associated discussion (they are called “tortoise shells” in Unicode):

> tortoises and turtles are not the same thing!
~ @Enivex

There will be a sibling PR to rename the symbols too.

Additionally, I refactored some code to make it more readable.
I also fixed a bug introduced in #3262 where the wrong script style was being used.

Closes #3856.

PR #3096, opened 2023-12-28
New performance timings

Dherse:
This replaces the current tracing implementation with a much lighter, much faster custom implementation. Currently it can only record to a `ptrace` compatible JSON file using the `--record` CLI argument. It can also record incremental compilation! It only stores file names and line numbers no additional information. It allows to produce flamegraphs like so:
image

But instead of producing the flamegraphs "in the app", they must now be produced using perfetto https://ui.perfetto.dev/

PR #4637, opened 2024-07-28
Add missing under/over-delimiters

### LINKED ISSUE #3856: Add missing over/under-delimiters
While looking at https://github.com/typst/typst/pull/3564 I realized that there are some obvious missing functions from https://typst.app/docs/reference/math/underover

In particular `underparen` (U+023DD) and `overparen` (U+023DC)

For completeness there's also the tortoise shell brackets U+023E0 (top) and U+023E1 (bottom). Apparently they're called `obrbrak` and `ubrbrak` in LaTeX. If we add them I propose we use different names. `undershell` and `overshell` perhaps?

mkorje:
Adds functions `underparen` & `overparen` and `undershell` & `overshell`, for parentheses and tortoise shell brackets respectively, to Under/Over in Math.

Fixes incorrect style for an annotation above a delimiter in `layout_underoverspreader` (introduced in #3262).

Fixes typo in `LayoutMath` implementation for `UnderbracketElem` (added in #3096).

Refactors some code in `underover.rs` to make it a bit more readable, taking on comments from #3869.

Closes #3856.


crates/typst/src/math/underover.rs
```diff
@@ -267,30 +397,28 @@ fn layout_underoverspreader(
...
+            }
+            0
+        }
+        Position::Over => {
+            rows.insert(0, stretched.into());
+            if let Some(annotation) = annotation {
+                let over_style = style_for_superscript(styles);
+                let annotation_styles = styles.chain(&over_style);
+                rows.insert(0, ctx.layout_into_run(annotation, annotation_styles)?);
+            }
```
laurmaedje, line 417: Somewhat contrary to my comment in https://github.com/typst/typst/pull/3869, I think we should actually just use `push` and move the `MathRun::new(vec![body])` into the match as well, instead of doing `insert(0, ..)` twice.

laurmaedje:
Thank you!

Issue #274, opened 2023-03-24
Single letter strings get parsed as italicized in math mode


SekoiaTree:
```
#let q1="q"
#let q2="qq"
$ q q1 q2 "qq" $
```

![image](https://user-images.githubusercontent.com/51149447/227609317-90dc4f6b-15b2-43f7-8119-b4f7f54a55d0.png)

It seems strange to me that `"q"` is rendered the same as `q`. Currently, this can technically be worked around with `"q "`, but that adds a space (not particularly different, but still slightly)

jakobjpeters:
I agree that it is unexpected and inconsistent. Since there is already a way to create the italic character (no quotes), it would be better to make characters inside quotes behave consistently.

GaugeAndGravity:
Find the same issue when I was trying `#let dd="d"` for differential and felt surprised on this behavior...

dccsillag:
> Find the same issue when I was trying `#let dd="d"` for differential and felt surprised on this behavior...

BTW, for differential you can just use `dif`.

mattfbacon:
FYI https://github.com/typst/typst/blob/main/library/src/math/ctx.rs#L132-L146

thinety:
I would like to add that I was also surprised by this behavior when trying to write units that use special symbols / Greek letters: `$R = 50 med "k" Omega$` displays the `k` in italic. The alternatives are `$R = 50 med upright(k) Omega$` or
```
#let kohm = "k" + sym.Omega
$R = 50 med kohm$
```

lucasvreis:
Looking at this and #366, I think the current loose and automatic separation between "math variable" and "prose within math" is problematic, especially if you consider that not everybody speaks English. In a plethora of languages you have single-letter connectives like "e" and "y" ("and" in Portuguese and Spanish), and they should appear inside maths upright. But I'm very surprised that _none_ of the following work:
```
$ 1+1 = 2 "e" 1+2=3 $ // Italic
$ 1+1 = 2 text("e") 1+2=3 $ // Italic
$ 1+1 = 2 upright("e") 1+2=3 $ // Wrong spacing!!
$ 1+1 = 2 upright(" e ") 1+2=3 $ // Wrong spacing again!!!
$ 1+1 = 2upright(" e ")1+2=3 $ // Wrong spacing!!!
```
The closest seems to be this ugly form:
```
$ 1+1 = 2upright(" e") 1+2=3 $ // finally!
```
[...]

PR #1779, opened 2023-07-23
Create `VarElem` type replacing `TextElem` in math (WIP)

damaxwell:
This PR is a work in progress and is a partial implementation of option (2) of RFC #6, #1125 concerning text vs. math in formulas in an attempt to shake down what the core issues are.  Sometimes you need to implement something to find out where the tricky parts lie.  In this case, some of the (initial) hard stuff turns out to be considerations not discussed in #1125: symbols, and regex.  The PR is a draft, and will likely just serve as guidance toward whatever the end solution is.

The core issue is that when displaying mathematics
- some text should be easy to typeset in the same style as the document
- mathematical text is handled completely separately:
    - Most fonts cannot be used for math typesetting.
    - Math fonts typically have limited kerning/glyph substitutions, making them unsuitable for ordinary text.
    - The mechanisms for making things bold or italic are entirely different for the two contexts (use a different codepoint vs use a different face)
    - Math fonts have optical sizing features missing from ordinary fonts.
    - Etc.

Currently Typst has only one notion of text.  In a formula environment, text consisting of a single character is treated very specially and corresponds to mathematical text. Runs of numbers have their own, analogous, special treatement as mathematical text, and everything else is essentially treated as ordinary text. Ordinary text is run through the same mechanism Typst uses to layout text in a paragraph, except that it's using a suboptimal math font to do ordinary text typesetting.

One natural attempt to remedy this is to introduce two different kinds of text that can appear simultaneously: ordinary and math.  Given `$ x "w" $`, we want `x` to be math text and `"w"` to be ordinary, and the first thing to settle is which of these corresponds to the currently existing `text` element function.  However, if the math text occupies the role of `text`, I'm not sure how to get `"w"` to escape back to the `text` settings when the formula was started so that the document's text properties can be inherited (much less be altered on the fly).

So it seems that the right thing to do is have `text` mean the same thing inside and outside of a formula.  This means there needs to be a different entity to represent mathematical text.  This is option (2) of #1125, and the name of the entity for the sake of argument  is `math.var`.

## Basic design

The PR introduces two new element functions, `VarElem` and `SymbolElem`, though only `VarElem` is exposed (as `math.var`).  When parsing a formula, material that used to parse as `SyntaxKind::Text` now becomes `SyntaxKind::MathVar`; that would be singleton-characters and numbers.  No other changes are made at the parser level and no syntax sugar is available for a `VarElem`.

[...]

Issue #1125, opened 2023-05-06
[RFC #6] Clearer syntax for displaying text/vars in math


PgBiel:
## Motivation

Currently, there are a few notable inconsistencies in how math mode handles the distinction between fonts for variables/operators and for text. In particular, the syntax `$ "double quotes" $`  is used both to introduce multi-character variables (such as in `$ "speed" = 5 $`), and also to introduce text (such as in `$ 6 > 5 "iff" 5 > 4 $`). However, the text produced from the double-quote syntax will necessarily use the math font, not the normal text font used along the document. Not even manually inserting `#text[...]` fixes that (the inserted text still uses the math font); the only way to manually insert the normal text font in math mode is to use `#text(font: "The text font here")[...]`, which is very verbose.

Additionally, the double quotes syntax behaves a little weird right now: `"a"` is displayed italicized, while `"aa"` is upright (see #274).

Therefore, the purpose of this RFC is to hear opinions regarding what the best syntax would be to introduce normal text inside math mode (using the document's or context's text font), and to make it separate from math font text (variable/operator names and stuff).

We already had some discussion regarding this over Discord, and I'll try to show some of the suggestions sent there here.

### Use `text()` for normal text font

This possibility suggests that all math mode text should default to the math font, inclusing text inserted with double quotes, as it is now. However, a `math.text` function would be added, letting you use `$ 6 > 5 text("iff") 5 > 4 $` to enforce the document text font.

- **Pros:**
    - Syntax would be familiar to LaTeX users (where it's `\text{}`)
    - Pretty clear that it's the text font (it's called "text" after all)
    - Fully backwards-compatible (everything written in the old syntax would remain exactly the same)
- **Cons:**
    - Still a bit verbose (and the `"..."` syntax has been quite convenient)
    - Could be annoying nonetheless for those who don't come from LaTeX
    - Wouldn't fix `"a"` not being upright (might be something that could be fixed in other ways though)
    - Would be confusing, since `#text()` would revert to math font (since that would be the default text element, not the math-specialized one, due to how the internals work), and overall the logic would look backwards (a user could ask: why does the default text element produce math text in a math environment, while `math.text` produces normal text?).
    - User could be confused regarding why `" "` and `text(" ")` are different.

### Use `var()` for math font

[...]

PR #4638, opened 2024-07-29
Implement `math.var` to differentiate `$a$` and `$"a"$` in equations.

wrzian:
This is a draft implementation of the `math.var` updates for semantically separating math variables from other text in equations.

This will close the ancient #274 and is the beginning of a solution for #254, #366, and #1125.

This removes the auto-italics correction on single-character text in math that is not a `math.var` element. However, my intent is that math layout will eventually be updated to only use the math font within `math.var` elements and use the document font for strings and normal content in math. That update will likely come in a later PR, and I would like to leave it as a separate discussion for now, see #366.

This is similar to PR #1779 by @damaxwell. However, with the updates to the layout system, I wasn't able to use much of that code at all and wrote most of this from scratch. That PR was closed for a number of open questions around show rules and regex replacement. This PR is focused mainly on fixing the auto-italics issue and leaving deeper language changes related to show-rules or new syntax for a later date.

There is still a lot to do and many details to fuss over, but the core has been implemented, and I'm happy to report only 4 tests are currently failing! There have been so many edge cases and horrible details in implementing this, but I'd love to nail the remaining ones down and maybe get this shipped for 0.12.

~~The four currently failing tests are the following:~~
- ~~`math-attach-show-limit (tests/suite/math/attach.typ:106)` and `math-symbol-show-rule (tests/suite/math/interactions.typ:67)`\
  These two are related to show-rules on symbol elements in math not working at all right now.~~
- ~~`enum-numbering-closure-nested-complex (tests/suite/model/enum.typ:86)`\
  This one is probably just weird behavior that we're going to ignore. But I'm really not sure what to do for it.~~
- ~~`math-lr-color (tests/suite/math/delimited.typ:42)`\
  This is related to the semantic difference between `math.var` and `text` in general. Not sure what the resolution should be here, but requires some thought about how `text` and `math.var` are styled.~~

~~I'm not fully sure how to fix these right now, and would love any guidance or suggestions from others.~~ \
EDIT: fixed the remaining failures! Will add an explanation for the test updates below.

I'll also make a general call for help: I would love any help updating documentation around symbols/`math.var` or adding tests to `var.typ`. I'll be doing this myself as I can, but would appreciate any others chipping-in.


wrzian:
I believe this also closes #3727. I'll double check when adding tests to `var.typ`.

wrzian:
Rebased onto main. Still need to document what the changes and tradeoffs are before review.

laurmaedje:
I'm sorry for completely ignoring this PR for months. It fell into the unforunate spot of being a very complex and intricate topic that is quite far removed from anything else I'd been thinking about at the time.

I took a look at the changes now and I must admit that I have a somewhat uneasy feeling about having a bool in symbol that states whether it came from math. It feels wrong somehow. What's your take on this particular change?

wrzian:
No worries on the delay! This change is a necessary fix, but certainly not the most urgent. It's also been a while since I've touched this, but that's given me some time to cook.

I was able to rebase from `shatter` (this PR got hit **hard**). I've tried to keep the changes roughly the same so we have something to test with/discuss, but I do want to change the implementation.

---

Having rewritten most of the code again, I agree that just a bool in `Symbol` is probably not right. But I'm not certain what else i would do at the moment. It might function as a short term solution, but that will require better defining what we want out of the split to `math.var`.

laurmaedje:
We should probably discuss this more on Discord.

laurmaedje:
Maybe some more direct discussion would be nice. How about having an actual Zoom or Discord design call about this sometime in January?

wrzian:
Rebased from main, but still need to fix a few things

wrzian:
Cleaned this pull request up for historical purposes, but we had our call and the resolution was to not create VarElem at the moment, but convert MathText kinds into a new SymbolElem that will be used for symbols going forward. We'll do the larger change of something like VarElem later, but SymbolElem should allow us to close #274.

Closing this for now and will open a new PR soon.

PR #1892, opened 2023-08-08
Add `outline-title` and `bookmark-title` to headings

balthild:
Closes #1889

Follow-up considerations: 

1. Is it OK to use `outline-title` as the name of field?
2. Should we add something like `bookmark-title` as well to support making the bookmark text different from the outline?

PR #4643, opened 2024-07-30
Allow different headings and captions in outline and bookmark

### LINKED ISSUE #1295: Short figure caption for outline
### Description

Allow adding an optional _short_ caption to be displayed in the outline

### Use Case

When adding a caption to a figure, the text might be quite long, in particular, when the figure contains a table or (multiple) images that require explanation. The entire caption text is automatically added to the outline, which may look really messy, depending on the text length. 
In the caption package of LaTeX, one can add an argument `<list entry>` (kind of a short caption), that shows up in the outline instead of the `<heading>` (i.e. the actual caption), if given: ` \caption[⟨list entry⟩]{⟨heading⟩} `. If no `<list entry>` is given, the `<heading>` is shown in the outline (as is happening now in typst) and if it is given, but left empty (` \caption[]{⟨heading⟩}`) , no entry will be made in the outline.

### LINKED ISSUE #1889: Different heading text for outline/bookmark and displaying
### Description

When using LaTeX, I'm able to specify an alternate text for `\part`, `\chapter`, `\section`, etc.

```tex
\chapter[alternate text]{display text}
```

If specified, the alternate text is used in ToC and PDF bookmarks.

### Use Case

This is useful when the title is long and I want to break it into lines manually without affecting the outline and bookmark.

image

It is also useful when I want to put some additional information (e.g. author, year) into the heading but not into the ToC.

EpicEricEE:
Closes #1889, closes #1295, implements https://github.com/typst/typst/pull/1892#issuecomment-1729119959.

## Summary

Added a (non-exposed) `outlined-content` element to be used for figure captions and headings. It is created internally by casting the dictionary
```
(
  document: [...],
  outline: [...],
  bookmark: "...",
)
```
where only the key `document` is required for the default value. If any of the other keys are omitted, `outline` is derived from the default value, and `bookmark` from the `outline` value. The `bookmark` value has to be of a string, while the other two can be any content. If this dictionary is not used when passing the body of a heading or figure caption, the element is still created implicitly with all fields having the same value (`bookmark` being converted to a string).

## Example

<table>
<tr>
<th>Code</td>
<th>Output</td>
</tr>
<tr>
<td>

```typ
#show outline: set heading(outlined: true)

#outline(title: (
  document: [Outline],
  outline: [Outline in Outline]
))

#outline(title: "Figures", target: figure)

#heading((
  document: [Document Heading],
  outline: [Outline Heading],
  bookmark: "Bookmark Heading"
))

#figure(
  rect(),
  caption: (
    document: [A rectangle],
    outline: [A rect in the outline]
  )
)
```
</td>
<td>

Example Output

</td>
</tr>
</table>

## Notes

- The `title` parameter of outline and bibliography elements also supports this new type.
- You can currently still specify the `bookmark` option for figure captions, even though it won't be displayed anywhere.
- In show rules, the `heading.body` and `figure.caption.body` elements are now of the type `outlined-content` which always has the fields `document`, `outline` and `bookmark` available. The type itself also has a show rule and returns the `document` value by default.
- I decided to only go for the `document / outline / bookmark` style and not also for `long / short` because `long` would only be a synonym for `document`, and because `bookmark` inherits its value from `outline`, so `short` would also basically only be a synonym for `outline` (when bookmark is not set).

## To-do

- [ ] Decide on name of element (I don't really like `outlined-content` but couldn't think of anything better)
- [ ] Decide whether the `outlined` and `bookmarked` parameters of headings should somehow be replaced by this
- [ ] Maybe improve documentation


MDLC01:
Doesn't this make `heading`'s `outlined` and `outlined`, and `figure`'s `outline` parameters obsolete? Because with this PR I feel like this could be done simply by setting the corresponding argument to `none` or some content / string.

EpicEricEE:
That's what I meant with the second to-do item.

One drawback is that something like `show heading.where(level: 3): set heading(bookmarked: false)` wouldn't be possible anymore, or at least not as easy. This is also the case for things like adding the outline in the outline (or removing the bibliography, etc.), as it is now usually done via `show outline: set heading(outlined: true)`

PgBiel:
Regarding the design, in principle this is already possible for arbitrary content in any context using `state`, at least for the outline. I wonder, therefore, if the only further step we need to take is to expose such an API (which applies to more general contexts as well) but more ergonomically. Though I do understand the appeal of a more minimal API, but maybe this is currently a bit _too_ minimal / ad-hoc? Or are there going to be extra features within heading and figure themselves which integrate with / are aware of the dictionary separating normal, outline and bookmark content?

EpicEricEE:
> Regarding the design, in principle this is already possible for arbitrary content in any context using state, at least for the outline.

Indeed. It would still require something for the bookmarked title though.

> I wonder, therefore, if the only further step we need to take is to expose such an API (which applies to more general contexts as well) but more ergonomically.

Do you mean something in the likes of (exact design aside)

```typ
#heading(context if inside(outline) [ ... ] else [ ... ])
```

which would resemble the behavior of the commonly used `in-outline` state but internally? I agree that something more generally applicable would be nice.

> Or are there going to be extra features within heading and figure themselves which integrate with / are aware of the dictionary separating normal, outline and bookmark content?

A nice to have would be to allow configuration of whether the heading number should be part of the bookmarked title, but that doesn't seem to be something that can be implemented nicely with either design.

laurmaedje:
I generally like the context-based solution, though I see how even just `context if inside(outline) [ ... ] else [ ... ]` might be a bit alienating and feel too complex for inclusion in the non-template prose part of a document. 

As for the bookmark: Is it truly necessary to distinguish between the outline and the bookmark text or could the bookmark just always resolve to the outline variant?

laurmaedje:
It seems to be that we need more discussion (ideally on Discord) before settling on a solution, so I'll close this for now. Thanks still!

woerndl:
Where can updates on this be subscribed too, will this PR be reopened or did the discussions on Discord lead to a new issue/PR?

Issue #2816, opened 2023-11-30
More control over repeat sizing and spacing


lkndl:
### Description

Currently `repeat` divides extra space between `n` repeat elements into `n-1` equally-sized pieces. There should be an option to not use the extra space of the parent element, i.e. stack to the left instead of justify; and ideally also a maximum width parameter that would allow `repeat` to be used without an explicitly-sized parent box.

### Reproduction URL
https://typst.app/project/r8n1E4ONWPQu3e3M7c6lcH

web app, on latest version of Typst

### Use Case

Alignment of spaced-out repeat elements in outline `fill` or any repeated use of `repeat` with the same element. 

![expand](https://github.com/typst/typst/assets/38729961/93f4ce85-9aa1-4a12-8d37-8456fc1c1fc9)

### Partial Fix
Here is a patch for the proposed `expand` flag, resulting in the good alignment in the bottom-most table of contents in the image above. I'm not up to `max-width` yet.

```patch
diff --git a/crates/typst/src/layout/repeat.rs b/crates/typst/src/layout/repeat.rs
index c733ea7d..d391a8db 100644
--- a/crates/typst/src/layout/repeat.rs
+++ b/crates/typst/src/layout/repeat.rs
@@ -32,6 +32,11 @@ pub struct RepeatElem {
     /// The content to repeat.
     #[required]
     pub body: Content,
+    // Whether to spread the available extra space
+    // smaller than one repeat element between them,
+    // or otherwise stack it on the right.
+    #[default(true)]
+    pub expand: bool,
 }
 
 impl Layout for RepeatElem {
@@ -46,13 +51,18 @@ impl Layout for RepeatElem {
         let piece = self.body().layout(engine, styles, pod)?.into_frame();
         let align = AlignElem::alignment_in(styles).resolve(styles);
 
-        let fill = regions.size.x;
+        let mut fill = regions.size.x;
         let width = piece.width();
         let count = (fill / width).floor();
         let remaining = fill % width;
-        let apart = remaining / (count - 1.0);
+        let mut apart = remaining / (count - 1.0);
 
-        let size = Size::new(regions.size.x, piece.height());
+        if !self.expand(styles) {
+            fill -= remaining;
+            apart = Abs::zero();
+        }
+
+        let size = Size::new(fill, piece.height());
 
         if !size.is_finite() {
             bail!(self.span(), "repeat with no size restrictions");
@@ -64,7 +74,7 @@ impl Layout for RepeatElem {
         }
 
[...]

PR #4644, opened 2024-07-30
Add `gap` and `justify` parameters to `repeat`

### LINKED ISSUE #2713: Extend functionality of `repeat`
### Description

The `repeat` function currently always justifies its content, and does not allow setting a gap between the repeated items. Instead, to add a gap it is necessary to include horizontal spacing inside the repeated item (as in `#repeat[.#h(4mm)]`), which however also includes that spacing in the last element.

I propose the following arguments:

image

This would also allow integration with `align`. Alternatively, it might be better to get the `justify` property from the current style set for `par`:
```typ
#show repeat: set par(justify: true)
#repeat(gap: 4mm, sym.dot.c)
```

As a workaround, it is relatively easy to write this function in pure typst as I've done [here](https://github.com/EpicEricEE/typst-plugins/blob/b13b0e1bc30beba65ff19d029e2dad61239a2819/outex/src/outex.typ#L1-L27).

### Use Case

Prevent misaligned dots in outline filler:

image

EpicEricEE:
Closes #2713

Added the parameters `gap` and `justify` to the `repeat` function, as described in the issue.

If `justify` is set to `true` (default), the gap is increased to completely fill the available width. If it is set to `auto`, the value is inherited from `par.justify`. I choose the default to be `true` solely to not break documents making use of the current behavior, but I'm open for changes in this regard.

As an alternative design ([first commit](<https://github.com/typst/typst/commit/ca4ad85e45dbbddd6a234ab6977dc39064822602>)), I originally only added a `gap` parameter, but allowed it to be set to `auto`. While I kind of prefer the simplicity of that, it wouldn't allow a justified repetition with a "minimum gap" like it would be possible when setting both `gap` and `justify` in the current design.

Alternatives for the name `gap` would be `spacing` (as for stack elements) or `gutter` (as for grids, tables and columns). An alternative for the name `justify` would be `expand` (as suggested in #2816).


crates/typst/src/layout/repeat.rs
```diff
@@ -35,6 +36,17 @@ pub struct RepeatElem {
...
+    /// The gap between each instance of the body.
+    #[default]
+    pub gap: Length,
+
+    /// Whether to increase the gap between instances to completely fill the
+    /// available space.
+    ///
+    /// When `{auto}`, inherits the paragraph's [`justify`]($par.justify).
+    #[default(Smart::Custom(true))]
+    pub justify: Smart<bool>,
```
laurmaedje, line 49: I'm not sure whether inheriting from `par.justify` is really good. I think `repeat` is typically used to produce a very specific layout and it changing up randomly based on whether the user uses justification could produce unintended results.

crates/typst/src/layout/repeat.rs
```diff
@@ -76,15 +79,27 @@ fn layout_repeat(
         frame.set_baseline(piece.baseline());
     }
 
+    let mut gap = elem.gap(styles).resolve(styles);
+    let fill = regions.size.x;
+    let width = piece.width();
+    let count = ((fill + gap) / (width + gap)).floor();
+    let remaining = (fill + gap) % (width + gap);
```
laurmaedje, line 86: I think the `fill + gap` calculation here is a bit hard to follow intuitively (at least, it's not directly obvious to me why it is correct). Could you add a brief comment explaining it?

laurmaedje:
Thank you!

Issue #4655, opened 2024-08-01
Downloading messages are being overwritten, or an LF is being inserted


Andrew15-5:
### Description

For

```typ
#import "@preview/pyrunner:0.1.0"
#import "@preview/jogs:0.2.3"
```

I see:

```
downloading @preview/pyrunner:0.1.0
```

But then:

```
1.4 MiB /   6.6 MiB ( 22 %)   1.4 MiB/s in  1s ETA:  3s
```

And then:

```
  4.2 MiB /   6.6 MiB ( 65 %)   2.1 MiB/s in  2s ETA:  1s
  6.6 MiB /   6.6 MiB (100 %)   2.1 MiB/s in  2s ETA:  0s

downloading @preview/jogs:0.2.3
345.8 KiB / 345.8 KiB (100 %) 345.8 KiB/s in  0s ETA:  0s
```

So in some cases a line is being overwritten when it shouldn't and in others — not being overwritten when it should. Haven't checked the logic, but this is weird, especially since the second one outputs just fine. I haven't tried with more than 2 packages. Same with v0.11.1 and c4dd6fa0.

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

PR #4658, opened 2024-08-02
File lock for package downloads

### LINKED ISSUE #2739: Package installations interfer when multiple Typst instances are running
### Description

One may resonably believe that compiling Typst is a "clean" operation, so it should be okay to run several compiling jobs at the same time. In fact, that works perfectly until packages are involved.

When compiling several documents that use a package that isn't installed yet, both Typst processes start downloading this package, but some of processes fail with messages like this one: 

```
downloading @preview/cetz:0.1.2
 45.9 KiB /  45.9 KiB (100 %)  45.9 KiB/s in  0s ETA:  0s
error: file not found (searched at C:\Users\andr\AppData\Local\typst/packages/preview/cetz/0.1.2\typst.toml)
  ┌─ \\?\C:\Users\andr\Desktop\projects\typstlike\typst-examples-book\src\packages\typst-src\05a1dbe2a2d17e5e81991406bed640775db6ab4ce2d585bc5a0d1175def43ea1.typ:4:8
  │
4 │ #import "@preview/cetz:0.1.2": canvas, draw
  │         ^^^^^^^^^^^^^^^^^^^^^
```

_Note:_ in the end, package is installed correctly, so recompilation fixes it all.

This isn't a really important thing, of course. There are several ways to avoid this error, but I think it would be nice if Typst could "lock" package directories to make such errors impossible.

_P.S._ The error is currently observed both on Windows and Linux.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

Andrew15-5:
Fixes #2739.

I currently have a working prototype.

```typ
#import "@preview/pyrunner:0.1.0"
#import "@preview/jogs:0.2.3"
```

Terminal 1:

```
acquired the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
  4.1 MiB /   6.6 MiB ( 62 %)   2.0 MiB/s in  2s ETA:  1s
  6.6 MiB /   6.6 MiB (100 %)   2.0 MiB/s in  2s ETA:  0s

unpacking the archive...
dropped the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
acquired the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
acquired the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
acquired the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
downloading @preview/jogs:0.2.3
345.7 KiB / 345.7 KiB (100 %) 345.7 KiB/s in  0s ETA:  0s

unpacking the archive...
dropped the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
acquired the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
acquired the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
```

Terminal 2:

```
couldn't aquire the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
acquired the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
acquired the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
couldn't aquire the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
acquired the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
acquired the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
```

The immediate question (besides the #4655) that I have is why there are multiple logs of "acquired" in both? I wanted to get:

Terminal 1:

```
acquired the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
  4.1 MiB /   6.6 MiB ( 62 %)   2.0 MiB/s in  2s ETA:  1s
  6.6 MiB /   6.6 MiB (100 %)   2.0 MiB/s in  2s ETA:  0s

unpacking the archive...
dropped the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
acquired the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
downloading @preview/jogs:0.2.3
345.7 KiB / 345.7 KiB (100 %) 345.7 KiB/s in  0s ETA:  0s

unpacking the archive...
dropped the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
```

Terminal 2:

```
couldn't aquire the lock file "~/.cache/typst/packages/preview/pyrunner/.0.1.0.lock"
couldn't aquire the lock file "~/.cache/typst/packages/preview/jogs/.0.2.3.lock"
```

But it still works, which is great. The other questions are:
1. Does the lock file location/name and creation/deletion routine look good?
2. Should the file lock be locked before downloading or after?
3. Should the commented section be included for more rich UX? Any changes?
4. Is the error handling good?
5. Anything else?


laurmaedje:
cc @tingerrr who I think had already looked into ways to do this (I personally have no clue so far)

tingerrr:
I'll do a review after lunch.

Andrew15-5:
I chose `fd-lock` as the most used crate. And I also noticed `.cargo-lock` in the `target/debug`directory. I _think_ `.package-cache` is the global cache dir lock file. So, the name can be whatever, but `.` at the start probably must stay. Also, since I use per-package-version lock, it adds a lot of zero-byte files, so in order to make it "transparent" or clean I chose to remove them after they've served their purpose.

As for the "Additional cool/user-friendly logs" part, I want to include that because 1 or 2 additional messages isn't much, but it gives user some feedback. At least with this, a user that runs 2+ instances will know that Typst is capable of parallel "downloads" and will give a peace of mind (by notifying that it knows that other instance is trying to download the same package version).

And the error mapping (to a new error), IMO, is one of the optimal and minimal ways of handling them (by using `?` at the end). I wanted to add a new method directly to `Result`, but this requires a new trait + implementation, which will probably look more bulky and unreadable than the current variant.

Lastly the https://github.com/yoshuawuyts/fd-lock/issues/28#issuecomment-2264180893 is stalled for now, but I think it wouldn't be hard to check the implementation and figure out the questions by myself.

crates/typst-cli/src/package.rs
```diff
@@ -130,6 +196,8 @@ impl PackageStorage {
             }
         };
 
+        println!("unpacking the archive...");
+        sleep(Duration::from_secs(5));
```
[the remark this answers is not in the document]
Andrew15-5: For visual effect (to compare output in real time). Just like this one: https://github.com/typst/typst/pull/4658#discussion_r1703961076. A temporary thing.

crates/typst-cli/src/package.rs
```diff
@@ -53,13 +56,76 @@ impl PackageStorage {
...
+            let lock_file_path = format!(
+                "~{}",
+                lock_file_path.to_string_lossy().strip_prefix(env!("HOME")).unwrap()
+            );
+            // let mut lock_file = RwLock::new(file.try_clone().unwrap());
+            let mut lock_file = RwLock::new(file);
+            let try_lock = lock_file.try_write();
+            let _lock = match try_lock {
+                Ok(lock) => {
+                    println!("acquired the lock file {lock_file_path:?}");
```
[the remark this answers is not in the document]
Andrew15-5: "Like that" means "at all" or you have a better idea? It's all just debugging, I don't know which messages will be left in the end.

crates/typst-cli/src/package.rs
```diff
@@ -53,13 +56,76 @@ impl PackageStorage {
...
+            let file = fs::OpenOptions::new()
+                .create(true)
+                .truncate(true)
+                .write(true)
+                // .read(true) // to test the error
+                .open(lock_file_path.as_path())
+                .map_err(file_locking_err("failed to open lock file"))?;
+            let lock_file_path = format!(
+                "~{}",
+                lock_file_path.to_string_lossy().strip_prefix(env!("HOME")).unwrap()
```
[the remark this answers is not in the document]
Andrew15-5: I'm not sure if it's necessary, as it's just a temporary pretty print for Linux home-relative paths. It most likely will be removed, so there will be nothing common.

Andrew15-5:
> I'm not a big fan because file locking is not very easy to port

I thought they are all quite portable. Are you implying on using another crate?

> This means that, on solaris, if we end up creating another fd pointing to this inode, like through a call to `clone`, dropping one of them prematurely drops the file lock for both fds.

I'm pretty sure that you only need and should have 1 lock per program. I think I tried something like this, and you can't clone the lock, at least after you call `write()`. So, with carefulness this isn't an issue, IMO.

crates/typst-cli/src/package.rs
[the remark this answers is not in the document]
tingerrr: I see, I think in the end we definitely want to show that we are blocking when we wait for a lock file, but otherwise not at all. This is similar to how cargo does it.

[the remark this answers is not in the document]
Andrew15-5: Yeah, I think so too, but I haven't thought about this yet.

tingerrr:
> I'm pretty sure that you only need and should have 1 lock per program. I think I tried something like this, and you can't clone the lock, at least after you call `write()`. So, with carefulness this isn't an issue, IMO.

It's just that there's no _nice_ solution to the problem of file locks. The problem is that on `solaris` this would be valid code, and it would write to a file thinking it has an exclusive lock when it doesn't:

```rust
use std::fs::File;

fn main() {
    let f1 = File::options()
        .read(true)
        .write(true)
        .create(true)
        .open("test.")
        .unwrap();

    let f2 = f1.try_clone().unwrap();

    // fnctl lock created
    let mut lock = fd_lock::RwLock::new(f1);

    // fnctl lock dropped because of `libc::close`
    drop(f2);

    // file written used under the assumption that it still is locked
    let _w = lock.write().unwrap();
}
```

The issue is that file locks created by `fnctl` are per `(inode, pid)` pair. Closing one fd `a` pointing to inode `x` will release the lock you created on another fd `b` also pointing to inode `x` (assuming the same process does it). Unix locks are also advisory, unlike windows locks, a process can simply ignore them and wreak havoc. This also only happens on `solaris` targets, all other targets use `flock`, which creates locks per `fd`.

I wanted to mention it, since this can be a problem if a contributor later uses the same mechanism for actually trying to do work on a file which needs a proper file lock.

We should at the very least document this somewhere in the source code and perhaps abstract it in some way to make cloning harder (we can't stop people from creating two separate files with the same path, unfortunately).

tingerrr:
I think in this specific case you can simply continue as is by unpacking since it's completely empty, although the program acquiring a lock file and not creating the directory itself would probably be the rarest of Ctrl+C's.

Andrew15-5:
So the `dir.exists()` will be called twice, if the first lock is failed to be acquired. I moved the logic so that this condition is only run once, but I guess there is no harm in checking it twice.

Andrew15-5:
I [stumbled](https://discord.com/channels/1054443721975922748/1270009805188366419) on one problem that I'm not sure how to deal with. Basically, exclusive write lock works fine if multiple instances started more or less at the same time. If the difference is more than a few seconds, then all (both) of them will have the acquired lock, which is impossible by its nature. I haven't looked at its implementation. The problem can be that I use it incorrectly (I just copied the general pattern from an official example) or that the library isn't perfect.

crates/typst-cli/src/package.rs
```diff
@@ -56,10 +59,84 @@ impl PackageStorage {
             if dir.exists() {
                 return Ok(dir);
             }
+            let parent_dir = dir.parent().unwrap();
```
PgBiel, line 62: It could be cleaner to do the reverse: something like `let parent_subdir = format!("{}/{}", spec.namespace, spec.name)` at the top of the function, then `let parent_dir = cache_dir.join(&parent_subdir)` and `let dir = parent_dir.join(&spec.version)` here (and `.join(&parent_subdir).join(&spec.version)` at the previous `if let`).

crates/typst-cli/src/package.rs
```diff
@@ -56,10 +59,84 @@ impl PackageStorage {
...
+
+            let file_locking_err = |string| {
+                move |err| PackageError::FileLocking(eco_format!("{string}: {err}"))
+            };
```
PgBiel, line 66: This should probably be a private top-level function.

crates/typst-cli/src/package.rs
```diff
@@ -56,10 +59,84 @@ impl PackageStorage {
...
+                .write(true)
+                // .read(true) // to test the error
+                .open(lock_file_path.as_path())
+                .map_err(file_locking_err("failed to open lock file"))?;
+            let lock_file_path = format!(
+                "~{}",
+                lock_file_path.to_string_lossy().strip_prefix(env!("HOME")).unwrap()
+            );
+            // let mut lock_file = RwLock::new(file.try_clone().unwrap());
+            let mut lock_file = RwLock::new(file);
```
PgBiel, line 92: I'd prefer to use the full namespace path here, as this had initially confused me (missed the import and thought this was `std::sync::RwLock`).

```suggestion
            let mut lock_file = fd_lock::RwLock::new(file);
```

PgBiel (commented):
Thanks! I'll leave some initial comments; haven't fully checked the logic yet (I see it is a bit WIP as well).

For reference, we may want to take a look at how the Gleam compiler does it: https://github.com/gleam-lang/gleam/blob/f9c2d98970241e51fba58cf998e10245f8a2fe05/compiler-cli/src/build_lock.rs

crates/typst-cli/src/package.rs
Andrew15-5: Why? It's only used in this one function, and it's very small for an anonymous function. So I think it should stay like that. At least for now. Maybe because of rebase or other changes, it will be used in multiple functions. Then it makes sense to put it in top-level scope.

Andrew15-5:
The question is which `PathBuf` must `prepare_package` return? From package or cache dir? Since both are returned for me, this breaks logic, as the function must return only one thing. It changes whether the package is already installed and ready to be used or it is only in cache and needs to be installed before its usage.

crates/typst-cli/src/package.rs
PgBiel: It's more about having it be a cheap top-level function instead of unnecessarily creating a closure instance on each call. It's possible this is done for us automatically on more optimized profiles, but it's preferable to avoid closures when you're not capturing anything in general.

This pattern is used elsewhere in the codebase to provide errors, containing just a single call to `eco_format!(...)`.

Andrew15-5: I was just about to interject because of most likely inlined optimization of the compiler, but I don't know for sure if it does this.

Andrew15-5:
> For reference, we may want to take a look at how the Gleam compiler does it: https://github.com/gleam-lang/gleam/blob/f9c2d98970241e51fba58cf998e10245f8a2fe05/compiler-cli/src/build_lock.rs

It uses [fslock](https://crates.io/crates/fslock). I think it's easier to try this or [fs4](https://crates.io/crates/fs4) first. I think at least one should work. Otherwise something isn't right.

https://docs.rs/fs4/0.9.1/fs4/fs_std/trait.FileExt.html:

> This library provides whole-file locks in both shared (read) and exclusive (read-write) varieties.
> 
> File locks are a cross-platform hazard since the file lock APIs exposed by operating system kernels vary in subtle and not-so-subtle ways.
> 
> The API exposed by this library can be safely used across platforms as long as the following rules are followed:
> 
> -    Multiple locks should not be created on an individual File instance concurrently.
> -    Duplicated files should not be locked without great care.
> -    Files to be locked should be opened with at least read or write permissions.
> -    File locks may only be relied upon to be advisory.

These rules are a bit confusing, but I think I didn't break them.

> Regarding the `fd-lock` crate, I'm not a big fan because file locking is not very easy to port, this can be seen in its `unix` implementation which falls back to `fnctl` for the `solaris` target.
> 
> This means that, on solaris, if we end up creating another fd pointing to this inode, like through a call to `clone`, dropping one of them prematurely drops the file lock for both fds.

I found that fslock actually have a per PID locking by simply putting PID number in the lock file. IIUC, this means that it is super portable and has to work perfectly. I will try that too. Strange that it's the only one of 3 that has this "feature". I also think that it's one of the oldest ways of handling locking? At least it feels this way.

crates/typst-cli/src/package.rs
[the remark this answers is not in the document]
Andrew15-5: This is [marked as temporary](https://github.com/typst/typst/blob/7af1c0d938c4c5cdce047d54826656bef247be22/crates/typst-kit/src/package.rs#L131). We can discuss logging/messages when the core logic will 100% work.

crates/typst/src/diag.rs
```diff
@@ -509,6 +509,8 @@ pub enum PackageError {
     NetworkFailed(Option<EcoString>),
     /// The package archive was malformed.
     MalformedArchive(Option<EcoString>),
+    /// Some error when dealing with file locking.
+    FileLocking(EcoString),
```
laurmaedje, line 513: Honestly I think this can just be `Other`. It is very specific to the CLI and the `PackageError` is supposed to be generic over all Worlds, just distinguishing the most important stuff.

crates/typst-kit/src/package.rs
```diff
@@ -1,7 +1,9 @@
 //! Download and unpack packages and package indices.
 
 use std::fs;
+use std::io::{self, Write};
 use std::path::{Path, PathBuf};
+use std::{thread::sleep, time::Duration};
```
laurmaedje, line 6: please use one import per module

crates/typst-kit/Cargo.toml
```diff
@@ -24,6 +24,7 @@ native-tls = { workspace = true, optional = true }
 once_cell = { workspace = true }
 tar = { workspace = true, optional = true }
 ureq = { workspace = true, optional = true }
+fd-lock = { workspace = true }
```
laurmaedje, line 27: please move this upwards so that it is sorted alphabetically

crates/typst/src/diag.rs
Andrew15-5: One problem is that now it has a general file-locking error. Removing it will at least require expanding/rewriting concrete file-locking messages. Oh, since I have an error mapping function, I can just put string from this file to there. Should be easy to solve.

crates/typst-kit/src/package.rs
Andrew15-5: ? You mean separate for `thread` and `time`? Also it's temporary. Also also it looks like I removed `sleep()` at some point so it's not even used. I think I will re-add the sleep for easier debugging. All debugging stuff will be removed when the WIP label is removed.

crates/typst-kit/Cargo.toml
Andrew15-5: Oops, sorry. Since it's still WIP and I wanted just to make it work at _some_ level, I didn't bother with placement of this.

crates/typst-kit/src/package.rs
laurmaedje: Yeah, I meant thread and time. Basically not grouping multiple imports from different modules. But if it's temporary it doesn't matter.

PgBiel:
> I found that fslock actually have a per PID locking by simply putting PID number in the lock file. IIUC, this means that it is super portable and has to work perfectly. I will try that too. Strange that it's the only one of 3 that has this "feature". I also think that it's one of the oldest ways of handling locking? At least it feels this way.
>
> Oh, BTW, Gleam uses `.lock_with_pid()`. So this probably is a hint at something.

This is interesting. I'm trying to find the relevant differences between `lock` and `lock_with_pid` in `fslock`, and it seems that `lock_with_pid` will instantly unlock if writing the PID fails: https://github.com/brunoczim/fslock/blob/cceb79f02c73cee0def73e1be0696fc26b395863/src/lib.rs#L322

Do we know if this is a useful property?

> And another thing. fslock is the only one that uses libc bindings. I'm not sure, but wouldn't this make it impossible to make a fully statically linked binary? Others use rustix.

Pretty sure this is fine, the problem would be requiring glibc instead of musl, but both are libc implementations, IIUC.

In particular, Gleam's binary is fully statically linked (to the point that the release builds can be run on NixOS directly, without any additional tweaks to the OS linker), so, if we go with this option, it probably won't affect static linking.

Andrew15-5:
> Do we know if this is a useful property?

I think that it's normal, because if one instance can't write to a lock file, then no instance can (permission issues, no disk space left etc.). There is a lock check at the top, which means that for other reasons it won't unlock the lock. So IMO, the description checks out.

Andrew15-5:
I [found](https://discord.com/channels/1054443721975922748/1088371867913572452/1274047736035147799) the cause of problem. It's not a library, it's a `rm` at the start of each program which clears cache to re-download a package. Running rm once at start fixes the problem. Now fd-lock and fslock work as expected, no issues. Deleting the lock file at the end also doesn't cause any issues (which makes sense).

So now there are only 2 things left: decide on what to use for file locking and which messages should be included.

crates/typst/src/diag.rs
Andrew15-5: It then has to be like this:

```
failed to load package (failed to open lock file: No such file or directory (os error 2))
```

Which, I'm not sure if it's pretty. I would prefer to use something like context from anyhow, so that all errors are one per line.

tingerrr:
> [...] decide on what to use for file locking [...]

I'd say whichever is more actively maintained or provides a more complete API. However, these locks are not security critical the same way our web request stuff is, so maybe it's not so important which is more actively maintained.

> [...] which messages should be included [...]

I think this one ties into reusability for `typst-kit` and should work similar to how `Downloader` reports its messages. See below.

> [...] and which directory prepare_package() actually supposed to return.

The root of the freshly downloaded package as before, why would that be different now?

> It's also nice if I could write tests for this whole thing.

Agreed, I've looked into writing some fs-manipulation tests for typst-test and basically went with the `tempdir` crate and some extra sugar on top, this would be beneficial for other parts such as export testing too. See [here](https://github.com/tingerrr/typst-test/blob/cc755c03a20f04f9e501894ec227832b8a0432a6/crates/typst-test-lib/src/store/test.rs#L487-L511) and [here](https://github.com/tingerrr/typst-test/blob/cc755c03a20f04f9e501894ec227832b8a0432a6/crates/typst-test-stdx/src/fs.rs#L168-L199) for some inspiration.

Regarding the whole implementation, I think this should be abstracted and not be directly tied to downloading for two reasons:
- If we get more and more tooling which may operate on project directories, we also want a locking mechanism which they all respect.
  - Specifically, tooling which manipulates part of a project on disk directly, packages can be deleted and redownloaded, but a user may lose unsaved work if two Typst tools don't lock the project before working on a project.
- It should let the user decide which "events" (creating, waiting, destroying) they want to report and how they want to do that.
  - I personally only want to see a message if I'm waiting for a lock, for example.

I think `Downloader` is an example of how that could look like, perhaps the same `Progress` trait can be extended/changed to fit the more generic reporting use case.

Andrew15-5:
> The root of the freshly downloaded package as before, why would that be different now?

This is not a "root of the freshly downloaded package": https://github.com/typst/typst/pull/4658/files#diff-73b42dae337dc35b0cb436f260d3ce883c3d26acc6f5ce3158254f35d1b878a5R89.



> Agreed, I've looked into writing some fs-manipulation tests for typst-test and basically went with the `tempdir` crate and some extra sugar on top, this would be beneficial for other parts such as export testing too. See [here](https://github.com/tingerrr/typst-test/blob/cc755c03a20f04f9e501894ec227832b8a0432a6/crates/typst-test-lib/src/store/test.rs#L487-L511) and [here](https://github.com/tingerrr/typst-test/blob/cc755c03a20f04f9e501894ec227832b8a0432a6/crates/typst-test-stdx/src/fs.rs#L168-L199) for some inspiration.

I also tested something similar https://github.com/DioxusLabs/dioxus/pull/2822/files, but this does not explain how I can test parallel download (which also would require network mocking). I only probably can test `.try_lock_with_pid()` because it can indicate whether the lock is active or not. But that's about it. Not sure if this mini test is enough or good.

> I think `Downloader` is an example of how that could look like, perhaps the same `Progress` trait can be extended/changed to fit the more generic reporting use case.

I have no idea what this means, and I haven't looked closely at what those do. You can abstract file locking only so much. A critical part is to not accidentally drop the lock. I can try extracting it into a few functions if that is what you mean.

tingerrr:
> This is not a "root of the freshly downloaded package": https://github.com/typst/typst/pull/4658/files#diff-73b42dae337dc35b0cb436f260d3ce883c3d26acc6f5ce3158254f35d1b878a5R89.

Ok, so? It's the root of the package, downloaded or not. This function should _always_ return the right directory once it's done.

> I also tested something similar https://github.com/DioxusLabs/dioxus/pull/2822/files, but this does not explain how I can test parallel download (which also would require network mocking). I only probably can test .try_lock_with_pid() because it can indicate whether the lock is active or not. But that's about it. Not sure if this mini test is enough or good.

I would only test that the file is created, that it is locked and deleted in the good case, you cannot easily test parallel downloads reproducibly.

> I have no idea what this means, and I haven't looked closely at what those do. You can abstract file locking only so much. A critical part is to not accidentally drop the lock. I can try extracting it into a few functions if that is what you mean.

We need an API in which you can simply do some work and the file locking is done for you, and reporting uses the same kind of trait object that the `Downloader` does. It's up to you if that needs a struct or can just be a single function.

Andrew15-5:
> Ok, so? It's the root of the package, downloaded or not. This function should _always_ return the right directory once it's done.

<strike>First, that already conflicts with your initial statement. Second, this means that the compiler will use the cache directory for a dependency and not the installed directory.</strike>

I completely forgot or didn't realize that compiler always uses cache dir as the main dir for packages. The non-cache dir only serves as an override. And here I thought that it's weird how I can't find code where package is copied from cache dir to install dir. Now it all makes sense. I need to document this.



> I think `Downloader` is an example of how that could look like, perhaps the same `Progress` trait can be extended/changed to fit the more generic reporting use case.

I've looked into the `Progress` trait and implementation. One question I have is why there is no return type? Why you have to deal with unreachable errors? I would like to have `Result<(), Box<dyn std::error::Error>>` so that I can propagate the error above. But for file locking, I still can't wrap my head around how exactly I can do something similar. Do I have to use `Progress` trait with modifications, or can I create a separate one? Or just creating functions is good enough?

Anyway, I'll get something done and then we can discuss this further.

Andrew15-5:
I think this is about as far as I can get on my own. I feel like I should probably move file locking struct to a separate file, but there is still a lot of commented code. The code now looks pretty clear and the `prepare_package` function again fits on the screen (if you remove garbage).

Now I'm ready to hear how this can be further abstracted, modified or glued to `Process` or whatever.

crates/typst-kit/src/package.rs
```diff
@@ -14,6 +15,111 @@ use typst::syntax::package::{
...
+                .map_err(other_err("failed to remove lock file"))?;
+        }
+        eprintln!("[tmp] dropped the lock file {:?}", self.tmp_path);
+        Ok(())
+    }
+}
+
+impl Drop for LockFile {
+    fn drop(&mut self) {
+        self.remove().expect("An error occured when dropping a lock file")
```
tingerrr, line 119: This may be called during a panic, a panic within a panic causes an immediate abort, and I'm not sure if we want that. Perhaps a simple print to stderr is better (through `writeln!(io::stderr(), ...)`, since `eprintln!` would panic for a closed stream, I believe). Including a file path would probably be important too.

crates/typst-kit/src/package.rs
```diff
@@ -14,6 +15,111 @@ use typst::syntax::package::{
 
 use crate::download::{Downloader, Progress};
 
+struct LockFile {
+    path: PathBuf,
+    tmp_path: String,
```
tingerrr, line 20: I assume the `tmp` prefix here means that this is not included in the final version?

crates/typst-kit/src/package.rs
```diff
@@ -14,6 +15,111 @@ use typst::syntax::package::{
...
+    /// Will be automatically called when struct is being dropped ([Drop]).
+    fn remove(&self) -> Result<(), PackageError> {
+        if self.path.exists() {
+            fs::remove_file(&self.path)
+                .map_err(other_err("failed to remove lock file"))?;
+        }
+        eprintln!("[tmp] dropped the lock file {:?}", self.tmp_path);
+        Ok(())
+    }
+}
```
tingerrr, line 115: Some of these should be public.

tingerrr (commented):
Perhaps we could provide either functions which takes a closure and a lock file path or additional constructors for the `Lockfile` struct. Both of which should handle creation + locking, I don't see much reason in creation and locking being different operations for the consumer.

Specifically:
- two generic versions which the others are built on
  - blocking version with a `Downloader`-like mechanism for the blocking message (since there's only one message `FnOnce` is probably enough)
  - non-blocking version without any reporting
- two default blocking implementations
  - for package downloads which takes a `PackageStorage` and a `PackageSpec`
  - for project locks, which simply takes a package root and creates a sibling file to `typst.toml` called `.typst.lock` or something like that

I reckon in 90% of cases, a consumer of this API would simply want to do _something_ with a well-defined lock file that other tooling respects.

crates/typst-kit/src/package.rs
Andrew15-5: I'm not sure about that yet.

tingerrr: Well, it's only used for the temporary log messages and can be reconstructed from the actual `path`.

crates/typst-kit/src/package.rs
PgBiel: We should be careful to not over-engineer this - we just need to expose an API to lock what's relevant to us (that is, the package cache), and abstract away the implementation details.

I feel like Gleam's API for this (https://github.com/gleam-lang/gleam/blob/f9c2d98970241e51fba58cf998e10245f8a2fe05/compiler-cli/src/build_lock.rs) is a great reference for how to proceed in this aspect as well.

crates/typst-kit/src/package.rs
Andrew15-5: > This may be called during a panic, a panic within a panic causes an immediate abort, and I'm not sure if we want that.

Pretty sure 1 or 2 panics wouldn't make any difference, other than show backtrace by default.



> Perhaps a simple print to stderr is better

Yeah.

> since `eprintln!` would panic for a closed stream, I believe

No, for a normal panic it works normally.

crates/typst-kit/src/package.rs
Andrew15-5: Only in a weird way. Instead of just saving the good parts of path directly in `String`, it would need a separate method that will grab the last few parts from the `PathBuf` with either `.unwrap()`s (or `?`, maybe). I don't think that would more clean or more performant. So I would hold on to it for now. It shouldn't matter if struct will be a few bytes bigger, as this is not a critical part of any program.

So I think it's the best overall method of storing the lock file path for logging. But on the other hand, you would probably want a full path to the lock file, not a cache-dir-relative one, right? Then I can remove it and use `path` directly.

crates/typst-kit/src/package.rs
Andrew15-5: Even though 1-method usage sounds very convenient, it doesn't feel idiomatic.

Andrew15-5: I've created `create_and_lock()` method. Additionally, I've kept the `true`/`false` value when checking if the lock was acquired right away. So, if you need to do some additional checks (yes) or messages (yes?), then you can call it. That's it. After going out of scope, the lock file will be removed.

Andrew15-5: This can potentially be changed to `LockFile::create_and_lock(..., |1st_try| { if 1st_try { ... }; ...; })`, which will give a clear scope, but maybe potential borrowing errors for beginners? This will also remove the need for `Drop` implementation.

crates/typst-kit/src/package.rs
tingerrr: > Pretty sure 1 or 2 panics wouldn't make any difference, other than show backtrace by default.

No, a panic within a panic aborts unconditionally, that's bad for many reasons, one of them being the fact that the user won't get any useful error message.

> No, for a normal panic it works normally.

What is a normal panic exactly? In any case, I did some reading, and it turns out `eprintln` doesn't panic when writing to `stderr`, while `println` will. This is completely unrelated to the current panic state, though.

crates/typst-kit/src/lockfile.rs
```diff
@@ -0,0 +1,149 @@
...
+    tmp_path: String,
+    /// Was the lock acquired on the first try.
+    acquired_first_try: Option<bool>,
+    file: fs::File,
+}
+
+// Where do I put it and who is gonna use it?
+/// Returns mapped `io::Error` to `PackageError::Other` with a custom message.
+/// Intended to be used inside `io::Result`'s `.map_err()` method.
+fn other_err(message: &'static str) -> impl Fn(io::Error) -> PackageError {
```
tingerrr, line 23: This sort of currying idiom is used nowhere in the code base, you can just have these messages be two parameters and extract common errors into static str constants.

crates/typst-kit/src/lockfile.rs
```diff
@@ -0,0 +1,149 @@
...
+        })
+    }
+
+    /// Creates a `LockFile` and tries to acquire the exclusive lock.
+    ///
+    /// `packages_root_dir` is normally either the cache or the local directory.
+    ///
+    /// Whether the lock acquired on the first try or not can be checked
+    /// with `did_wait()`. This can be useful for ... (do we need this sentence?)
+    pub fn create_and_lock(
```
tingerrr, line 60: This should directly take the `PackageStorage` instead of any path.

Additionally, this should fail if it can't acquire the lock, there's no reason to create a lockfile that could not be acquired and then require the use to keep trying and checking manually.

The point of such a constructor is to ensure the lock file is either taken or an error is returned, nothing in between.

crates/typst-kit/src/lockfile.rs
```diff
@@ -0,0 +1,149 @@
...
+        spec: &PackageSpec,
+    ) -> Result<Self, PackageError> {
+        let mut lock_file = Self::new(packages_root_dir, spec)?;
+        lock_file.acquired_first_try = Some(lock_file.try_and_lock()?); // something to tweak here?
+        Ok(lock_file)
+    }
+
+    pub fn did_wait(&self) -> bool {
+        self.acquired_first_try
+            .expect("You must use it only after calling creat_and_lock!")
```
tingerrr, line 71:
```suggestion
            .expect("You must use it only after calling create_and_lock!")
```

crates/typst-kit/src/lockfile.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Acquire exclusive lock.
+    fn lock(&self) -> Result<(), PackageError> {
+        self.file
+            .lock_exclusive()
+            .map_err(other_err("failed to acquire lock file"))
+    }
+
+    /// Try acquiring the exclusive lock, if failed print an error an wait
```
tingerrr, line 86: It's up to the caller whether they want to print a message, this is unrelated to the lockfile implementation.

crates/typst-kit/src/lockfile.rs
```diff
@@ -0,0 +1,149 @@
...
+    ///
+    /// Returns `Ok(true)` if the lock was acquired on the first try,
+    /// otherwise `Ok(false)`.
+    fn try_and_lock(&self) -> Result<bool, PackageError> {
+        match self.try_lock() {
+            Ok(_) => {
+                eprintln!("[tmp] acquired the lock file {:?}", self.tmp_path);
+                Ok(false)
+            }
+            Err(ref err) if err.kind() == io::ErrorKind::WouldBlock => {
```
tingerrr, line 97:
```suggestion
            Err(err) if err.kind() == io::ErrorKind::WouldBlock => {
```

I don't think this is required.

tingerrr (commented):
So my main idea with `create_and_lock` was to provide an API similar to this, I'm making the API up, so it may not be entirely accurate:
```rust
// return inner result or fail if we can't lock 
let result = typst_kit::lockfile::lock_package_download(
  &package_storage,
  &package_spec,
  |_path| eprintln!("waiting for lockfile"), // could do any kind of logging
  || {
    zlib::unpack(...);
    // ...
    result
  }
)?;
```

And a project version, since this is not just `typst-cli` but `typst-kit`:
```rust
let result = typst_kit::lockfile::lock_project(
  "path to project root",
  |_path| {}, // let's just not do anything here this time
  || {
    // downstream tooling doing fs operations on a project
    // ...
    result
  }
)?;
```

These should both always wait, as they're simple high level APIs, lower level non-waiting versions can be provided as basis for these.

crates/typst-kit/src/lockfile.rs
Andrew15-5: No error is being printed.

crates/typst-kit/src/lockfile.rs
Andrew15-5: Have you checked? It was required and probably is now. I can double check.

crates/typst-kit/src/lockfile.rs
tingerrr: I was referring to the documentation on line 86 here, the GitHub UI doesn't make this very clear, unfortunately.

laurmaedje:
What's the status of this PR?

Andrew15-5:
It's stale, because I don't have time. I keep on thinking about this PR from time to time. But by my calculations, I can continue very soon. Perhaps in 2 weeks.

laurmaedje:
I'll close this due to inactivity & to "unlock" the issue, giving others the chance to work on this. Feel free to reopen if you end up working on it again. Thanks!

laurmaedje:
> But I can't reopen if you closed it.

Didn't know that. If you're actively working it again, let me know. I'll reopen then.

PR #1809, opened 2023-07-27
add `title` and `capitalize` methods on type Str

slashformotion:
I am a rust newbie, i have no idea what i am doing. 
I will add docs later

PR #4673, opened 2024-08-04
feat(compiler): add capitalize to Str

slashformotion:


Enivex:
- Considering `lower()` and `upper()` shouldn't this also be a standalone function and not a method?

- The implementation seems too naive. In some languages and words the capitalized letter may not be the first one. I believe in e.g Gaelic.

slashformotion:
well, i just want to have a way to capitalize the first letter of a string so it fit my usecase for glossarium. 

As discussed in this PR #1809 , there is some rule that can be followed regarding capitalization but thoses rules depend on the language AND on different standards (see this comment https://github.com/typst/typst/pull/1809#issuecomment-1670343914). 

I think that it is reasonable to have a function that is naïve if it's behavior is correctly explained in the docs. 

> Considering lower() and upper() shouldn't this also be a standalone function and not a method?

I am not a Rust developer, so i currently have no idea how to implement a function instead of a method on strings. But i would be more than happy to get some pointers on how I could to it if you prefer a standalone function over a method.

alerque:
I've been working on a Rust crate [decasify](https://github.com/alerque/decasify/) that can be used for changing the case of a string using natural language rules (as opposed to naive approaches used in programming like [heck](https://github.com/withoutboats/heck) that follow programmatic ideas of casing not natural language ones). The *decasify* crate can be compiled to WASM and hence could be used in Typst directly or used as a package (I [already had the idea to make it a package](https://github.com/withoutboats/heck)).

Casing natural language is hard and error prone because of all the possible edge cases, but so far the English (Grubber/Daring Fireball style) and Turkish rules are fairly robust. I'd be happy to accept contributions for other style guides and languages.

laurmaedje:
> I [already had the idea to make it a package](https://github.com/withoutboats/heck)

Probably meant to link https://github.com/alerque/decasify/issues/10 instead of `heck` again.

Enivex:
The GPL license is a bit unfortunate, since it means typst itself cannot use the crate

alerque:
@Enivex The next major release [will be LGPL](https://github.com/alerque/decasify/issues/8), I already have permission from the other contributor to re-license. I only haven't done it because I consider that a breaking change and was expecting a minor patch release or two first. I don't think linking against the LGPL should cause any trouble for Typst.

laurmaedje:
Another thing is that `lower` and `upper` also work on content, not just strings. This would be harder to support for `capitalize`, but not supporting it is inconsistent.

laurmaedje:
I think, while not impossible, adding built-in capitalization is quite complex (due to both language support _and_ because it should support content for consistency). This would need more upfront design discussion, so I'll close this for now. Thank you still!

PR #4685, opened 2024-08-05
Support multiple stylistic sets in text

### LINKED ISSUE #4613: `text.stylistic-set` should take an array of values
### Description

Stylistic sets are supposed to be freely combinable, yet Typst’s `text.stylistic-set` property only allows setting a single one.

See also the [OpenType spec](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#ssxx): <q>This set of features are not mutually exclusive, hence it should be possible to activate any combination of them.</q>

### Use Case

This would enable users to use any combination of stylistic sets their font provides without having to resort to the unwieldy `text.features` property.

bluebear94:
Closes #4613.

TODO: Should `stylistic-set` be foldable (with an explicit way to turn a stylistic set off)?


crates/typst/src/text/mod.rs
```diff
@@ -37,6 +37,7 @@ use ttf_parser::{Rect, Tag};
 
 use crate::diag::{bail, warning, HintedStrResult, SourceResult};
 use crate::engine::Engine;
+use crate::foundations::IntoValue;
```
laurmaedje, line 40: please move this import into the list below (I wish rustfmt's import settings were stable :/)

crates/typst/src/text/mod.rs
```diff
@@ -566,13 +567,21 @@ pub struct TextElem {
...
+    /// Which stylistic sets to apply. Font designers can categorize alternative
     /// glyphs forms into stylistic sets. As this value is highly font-specific,
-    /// you need to consult your font to know which sets are available. When set
-    /// to an integer between `{1}` and `{20}`, enables the corresponding
-    /// OpenType font feature from `ss01`, ..., `ss20`.
+    /// you need to consult your font to know which sets are available.
+    ///
+    /// This can be set to an integer or an array of integers, all
+    /// of which must be between `{1}` and `{20}`, enabling the
+    /// corresponding OpenType feature from `ss01` to `ss20`.
```
laurmaedje, line 576:
```suggestion
    /// corresponding OpenType feature(s) from `ss01` to `ss20`.
```

laurmaedje:
Thank you!

PR #4686, opened 2024-08-05
Automatically get footnote numbering in counter display

EpicEricEE:
When using `counter(footnote).display()`, the numbering should be automatically retrieved from the current `footnote.numbering` value, similar to how it is done for equations, figures and headings.

Note that (as for the other countable elements) this doesn't work when the display function is called in a show rule and the numbering is explicitly set on the element, as it will always get the numbering value from the style chain.

### DIFF
--- a/crates/typst/src/introspection/counter.rs
+++ b/crates/typst/src/introspection/counter.rs
@@ -373,7 +373,7 @@ impl Counter {
                 } else if func == EquationElem::elem() {
                     EquationElem::numbering_in(styles).clone()
                 } else if func == FootnoteElem::elem() {
-                    FootnoteElem::numbering_in(styles).clone().into()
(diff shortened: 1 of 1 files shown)


crates/typst/src/introspection/counter.rs
```diff
@@ -372,6 +372,8 @@ impl Counter {
                     FigureElem::numbering_in(styles).clone()
                 } else if func == EquationElem::elem() {
                     EquationElem::numbering_in(styles).clone()
+                } else if func == FootnoteElem::elem() {
+                    FootnoteElem::numbering_in(styles).clone().into()
```
laurmaedje, line 376: I think an explicit `Some(..)` would be clearer than `.into()` here. I stumbled for a moment about what we're casting here.

laurmaedje:
Thanks!

PR #4687, opened 2024-08-05
Implement math mode language for `raw`

### LINKED ISSUE #4662: Math mode language for `raw`
### Description

The `raw` function's `lang` parameter supports markup `"typ"` and code `"typc"` modes, but should also support math `"typm"`.

### Use Case

Users can create correctly highlighted math mode text without requiring that the string is enclosed in dollar signs. This is useful when creating documents about Typst itself, where a text snippet may be assumed to be in math mode.

I personally would like to use this in the documentation for my Julia package, which formats Julia values into Typst. In this reduced example, the 5th and 7th columns would be highlighted the same.

Typstry_examples

jakobjpeters:
Closes #4662


tests/suite/foundations/repr.typ
```diff
@@ -38,6 +38,7 @@
 
 // Strings and escaping.
 #raw(repr("hi"), lang: "typc")
+#raw("#" + repr("hi"), lang: "typm")
```
laurmaedje, line 41: I think an example with some actual math would be better.

jakobjpeters: Got it, how is this?

laurmaedje:
Thanks!

PR #4689, opened 2024-08-05
Allow the extension tab to refresh test images in background

Leedehai:
Using the test-helper extension, after clicking the run button and while waiting for a test to finish, the developer might navigate away from the test's preview tab. When the test finishes, the preview tab isn't refreshed to show the updated output image, and this situation may be confusing. This PR addresses this.

This PR also does some cosmetic changes.
* renames `active` to `opened`, because the word "active" often means the tab is visible in the foreground, like [this](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/onActivated).
* `undefined` is enough to indicate there's no open tab, and there's no need for the `null` semantic.

### DIFF
--- a/tools/test-helper/src/extension.ts
+++ b/tools/test-helper/src/extension.ts
@@ -11,13 +11,15 @@ export function activate(context: vscode.ExtensionContext) {
 export function deactivate() {}
 
 class TestHelper {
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

PR #941, opened 2023-04-23
Spanish quote marks

Luis-Licea:
Fixes #922

PR #4701, opened 2024-08-07
Translate smartquotes and figure caption separators in translation files

MDLC01:
This PR moves the translations for smartquotes and figure caption separators into the translation files.

This required slightly modifying the format of the translation files to allow escape sequences. Since translation files are parsed at runtime instead of at compile time, this has the side-effect of using `EcoString` instead of `&'static str` in various places. Ideally, we should parse translation files at compile time, but this required more effort, and I'm not sure how definitive the current internationalisation infrastructure is, so I don't want to invest too much time optimizing something that may eventually be go away.

I generated the new translations automatically, so there should not be any error. The only noticeable change is for Spanish: smartquotes used to be non-default for es-ES, and es with specifically no region, but the default ones for es with any non-ES region. I chose to override default quotes for es-ES only, instead of es (which would have affected all regions instead of only es with no region) and changed a test accordingly. I'm not sure whether this was the best choice. Maybe es should be modified instead of es-ES, and special cases should be added for e.g., es-ES? Cc. @Luis-Licea, as you added the quotes for Spanish in #941.


ssotoen:
Equating double quotes with the outer pair and single quotes with the inner one is problematic. There are languages/regions where single and double quotes are reversed or the double quotes are used for both the inner and outer pair.

MDLC01:
This will slightly conflict with #4799, as both PRs add new translation files.

laurmaedje:
I'm a bit torn with this one to be honest. I see how it makes sense to have everything related to a language in one file, but it replaces a single fairly simple match with a whole bunch of machinery (as also seen in the fairly large +diff).

MDLC01:
Based on [the discussion on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1277669464078286870), I'm going to close this PR. If we revive it, I'll do that in a separate PRs as the changes will be very different.

PR #4706, opened 2024-08-08
Split shorthands into math and markup, add tilde shorthand

### LINKED ISSUE #2935: Changing `~` in math mode to be `tilde.op`
### Description

So, first on a personal note: on the math I've been working on, pretty much every time I want to have a tilde in math what I actually want is `tilde.op`. For example:

- Denoting in relation to what that we are taking some expectation: `EE_(X ~ D)[EE_(theta ~ pi)[f_theta (X)]]`
- Denoting a random sample: `X_1, ..., X_n ~^"iid" P`
- A "similarity" equivalence relation: `a ~ b`

I don't think I've ever wanted the plain tilde in my math.

And, in general, I struggle to think of some situation where you'd actually want a plain tilde in your math. Well, I'm sure the logic folks will have used it in _some_ paper, but they'd do that with just about any symbol you provide to them.

Thing is, `~` is somewhat privileged since it is part of ASCII and of many keyboard layouts (at least for the US and ABNT ones, which I'm familiar with. Maybe not others?). So I'd argue that when it appears in math, it should actually refer to `tilde.op`, not `tilde` as it currently does. Moreover, in order to account for people who may not have `~` on their keyboard, `tilde` itself should be remapped to `tilde.op`, with the current `tilde` being named something like `tilde.raw`.

What do you think?

P.S.: as a hack, currently, one can get `~` in meaning `tilde.op` with a show rule: `show "~": $tilde.op$`. I've been adding that to nearly all my documents :sweat_smile:

### Use Case

It's a convenience for making some math easier to typeset and easier to read from source. Examples include:

- Denoting in relation to what that we are taking some expectation.
  - Currently: `EE_(X tilde.op D)[EE_(theta tilde.op pi)[f_theta (X)]]`
  - With the proposed change: `EE_(X ~ D)[EE_(theta ~ pi)[f_theta (X)]]`
- Denoting a random sample.
  - Currently: `X_1, ..., X_n tilde.op^"iid" P`
  - With the proposed change: `X_1, ..., X_n ~^"iid" P`
- A "similarity" equivalence relation using tilde.
  - Currently: `a tilde.op b`
  - With the proposed change: `a ~ b`

mkorje:
Separate the math shorthands from `SyntaxKind::Shorthand`, moving them into their own `SyntaxKind::MathShorthand`. Change is needed to avoid overlap of shorthands.

Adds '~' as a shorthand for `tilde` (i.e. `tilde.op`) in math mode.

Closes #2935.

I think I've covered everything that needs to be added for the new `SyntaxKind`, though I've probably missed something. And the reason for changing the test is that now '~' has class `"relation"`, so I switched it to something that isn't.


laurmaedje:
Absolute bikesheeding, but I'm a bit torn on whether the `MathShorthand` syntax kind and AST node should live next to `Shorthand` or next the to `Math` kinds (if there, probably after `MathIdent`). I'm _leaning_ towards the latter, but not sure.

Also: Do we have a test for this and if not, could you add one?

laurmaedje:
Thank you!

Issue #3734, opened 2024-03-20
`table.header` should have an `align` field


ad-si:
### Description

```typ
table(
  columns: 3,
  table.header(
    align: (left, center, center),
    [*№*],
    [*Date*],
    [*Description*]
  ),
  align_rows: (center, left, left),
  [1], [2024-03-18], [Just a test]
)
```

It would probably be better to have a dedicated `align_rows` field to avoid an extra `if row != 0 { … }`

### Use Case

- Define alignment at the same position they create the header for better readable code.
- No more extra `if row == 0 { … }` necessary

frozolotl:
You can use the [`table.cell`](https://typst.app/docs/reference/model/table/#definitions-cell) function for this:
```typ
#table(
  columns: 3,
  align: (center, left, left),
  table.header(
    table.cell(align: left)[*№*],
    table.cell(align: center)[*Date*],
    table.cell(align: center)[*Description*]
  ),
  [1], [2024-03-18], [Just a test]
)
```

PgBiel:
I think I'd prefer getting https://github.com/typst/typst/issues/3640 to work first before attempting this, as it'd be logical to add all other table cell fields if we add just align, and I'm not sure if it would generalize too well. In particular, `stroke` would be a tough one to tackle.

PR #4716, opened 2024-08-09
Add support for alignment in {grid, table}.{header, footer}

bluebear94:


laurmaedje:
@PgBiel makes a good point in https://github.com/typst/typst/issues/3734#issuecomment-2278950923.

Issue #4718, opened 2024-08-10
Compiler assertion fails if local package cannot be found


inferiorhumanorgans:
### Description

```
#import "@our-local-ns/el-package:0.1.0"
```

```
$ ~/typst/typst compile --package-path=packages foo.typ
thread 'main' panicked at crates/typst-kit/src/package.rs:154:9:
assertion `left == right` failed
  left: "our-local-ns"
 right: "preview"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
$ file packages/our-local-ns/el-package/0.1.0
packages/our-local-ns/el-package/0.1.0: cannot open `packages/our-local-ns/el-package/0.1.0' (No such file or directory)
```

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

PgBiel:
Regression from #4540 - seems like the `preview` namespace check below was removed:  https://github.com/typst/typst/blob/810491c9d31b614a435020f888fbd380e8e039a1/crates/typst-cli/src/package.rs#L60-L66

vs now:

https://github.com/typst/typst/blob/672f6e5f97c2bdcd3e94754c9486869bf7b8de56/crates/typst-kit/src/package.rs#L87-L91

tingerrr:
Don't know how I maged that, easy fix, but I won't be home for a few hours,so you could do it too.

laurmaedje:
The way Git presents the changes in a refactoring PR like #4540 unfortunately makes it super hard to review so that both this and #4710 slipped by me.

PR #4720, opened 2024-08-10
Fix crash when local package  is missing and make OpenSSL optional in typst-kit

FlorentCLMichel:
This PR attempts to resolve  https://github.com/typst/typst/issues/4718 by reverting the change identified by @PgBiel . 
Not sure if someone is already working on it - if yes (or if the way I've done it is not right), feel free to close this PR.

It also makes the OpenSLL dependency in typst-kit optional as suggested by @laurmaedje [here](https://github.com/typst/typst/pull/4540).


laurmaedje:
Thanks!

PR #4724, opened 2024-08-10
Fix `calc.inf`, `-calc.inf`, and `calc.nan` reprs and displays

### LINKED ISSUE #2955: Infinite united values are represented weirdly (e.g., `infpt`)
### Description

I forgot about this case in #2936, so I'm opening this issue instead of making another small PR that changes a behavior that only rarely happens.

The repr for infinite values with units is weird. For example, an infinite length is represented as `"infpt"`, as illustrated by the following piece of code.

```typ
#set page(width: auto)
#layout(((width, ..)) => repr(width))
```

Similarly, infinity is displayed (or converted to a string using the `str` constructor) as `inf` instead of `∞`.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

MDLC01:
Fixes #2955.

I decided to use `calc.inf` and `calc.nan` instead of `float("inf")` and `float("nan")` because the former feel like constants, while the latter feel like more arbitrary ways of getting the values. Ideally, I think `float.inf` and `float.nan` would be better (see side note below).
| Value             | Repr                | Display / str |
|-------------------|---------------------|---------------|
| `calc.inf`        | `"calc.inf"`        | ∞             |
| `-calc.inf`       | `"-calc.inf"`       | −∞            |
| `calc.nan`        | `"calc.nan"`        | NaN           |
| `calc.inf * 1pt`  | `"calc.inf * 1pt"`  |               |
| `-calc.inf * 1pt` | `"-calc.inf * 1pt"` |               |

I also added some unrelated float-related tests in the last commit.

As a side note, I'm wondering whether we should move `calc.inf` and `calc.nan` to `float.inf` and `float.nan` as they seem more related to the IEEE-754 format than to calculations.


laurmaedje:
Thanks!

> As a side note, I'm wondering whether we should move calc.inf and calc.nan to float.inf and float.nan as they seem more related to the IEEE-754 format than to calculations.

That makes sense. `calc.nan` does not seem to be used in Typst Universe. `calc.inf` is used in three packages. So the breakage is rather minimal. But we could still duplicate them & deprecate the old ones for now. (No infrastructure for deprecating consts though atm.)

PR #4725, opened 2024-08-10
Improve documentation for `float` and `int` types

MDLC01:
This small documentation PR improves the documentation for the `float` and `int` types.

### DIFF
--- a/crates/typst/src/foundations/float.rs
+++ b/crates/typst/src/foundations/float.rs
@@ -83,8 +83,7 @@ impl f64 {
     ///
     /// - If the number is positive (including `{+0.0}`), returns `{1.0}`.
     /// - If the number is negative (including `{-0.0}`), returns `{-1.0}`.
-    /// - If the number is [`{calc.nan}`]($calc.nan), returns
-    ///   [`{calc.nan}`]($calc.nan).
+    /// - If the number is `{calc.nan}`, returns `{calc.nan}`.
     ///
     /// ```example
     /// #(5.0).signum() \
--- a/crates/typst/src/foundations/float.rs
+++ b/crates/typst/src/foundations/float.rs
@@ -66,8 +66,8 @@ impl f64 {
 
     /// Checks if a float is infinite.
     ///
-    /// Floats can represent positive and negative infinity. This function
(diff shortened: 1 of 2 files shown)


crates/typst/src/foundations/float.rs
```diff
@@ -65,8 +66,8 @@ impl f64 {
 
     /// Checks if a float is infinite.
     ///
-    /// For floats, there is positive and negative infinity. This function
-    /// returns `true` if the float is either positive or negative infinity.
+    /// Floats can represent positive and negative infinity. This function
+    /// returns `{true}` if the float is either an infinity.
```
xkevio, line 70: the `an` change doesn't make sense and is also incorrect grammar. leaving it as it was is fine

MDLC01: I'm not a native English speaker so I may be wrong, but, from my understanding, "infinity" can be used as a noun. There are two infinities: negative infinity, and positive infinity. I can refer to one of them with "an infinity".

xkevio: It depends on the context whether "infinity" is an uncountable noun, meaning it needs no article, or a countable one. Regardless, it is wrong in this context especially with the "either" as that would have to be followed by an "or" at some point.

MDLC01: Oh you're right! I meant to write "the float is an infinity" without "either". Would that be correct?

laurmaedje: Yeah, that'd be correct.

crates/typst/src/foundations/float.rs
```diff
@@ -27,10 +27,11 @@ impl f64 {
     /// Converts a value to a float.
     ///
     /// - Booleans are converted to `0.0` or `1.0`.
-    /// - Integers are converted to the closest 64-bit float.
+    /// - Integers are converted to the closest 64-bit float. For integers with
+    ///   absolute value less than 9 quadrillion, this conversion is exact.
```
laurmaedje, line 31: I think it would be write the exact number, but using exponential notation, i.e. 2^53.

laurmaedje:
Thanks!

Issue #2485, opened 2023-10-25
More flexible Figure, Table captions


gnojus:
### Description

Hi,

in Lithuanian language tables and pictures are numbered as `1 Lentelė`/`1 Pav.`, that is digit goes __before__ the supplement.
It would be nice if this would be available, either hardcoded for appropriate languages, or by expanding the way captions are specified.

I could contribute the Lithuanian text language, but this is blocking proper translation.

### Use Case

To create grammatically correct documents in Lithuanian language.

ichxorya:
I looked at the code to try but cannot get what caused the text order to happen. This is a good idea, we should have some generalized rules and certain localized rules.

gnojus:
Could be some of the places it's hardcoded in (for Figure)?: https://github.com/typst/typst/blob/0f274f8edbbb93d65cb11713f4bdf912b3286800/crates/typst-library/src/meta/figure.rs#L386 and https://github.com/typst/typst/blob/0f274f8edbbb93d65cb11713f4bdf912b3286800/crates/typst-library/src/meta/figure.rs#L554

Jisu-Woniu:
In Chinese, when we mention a section, we say “第 1 节” (meaning "No. 1 Section" approximately).

It would be good if we can use a supplement taking number as a parameter.

YDX-2147483647:
## A workaround for figures

Set `supplement` to `it => none`, and put the real supplement in `numbering`.

```typ
#import "@preview/numbly:0.1.0": numbly
#show figure.where(kind: table): set figure(
  numbering: numbly("👉 before {1} after 👈"),
  supplement: it => none,
)
```

The [numbly](https://typst.app/universe/package/numbly) package is necessary here, or `a` in _after_ will be interpreted as a pattern (abcd…).

```typ
That is @tab.

#figure(
  table(columns: 3, ..($...$,) * 9),
  caption: [Caption],
) <tab>
```

<img width="517" height="397" alt="Image" src="https://github.com/user-attachments/assets/7b898a61-9195-48be-a431-3407d6d137f6" />

PR #4729, opened 2024-08-11
Add support for page references through new `ref.form` property

### LINKED ISSUE #2873: Pageref in Typst like it exists in LaTeX
### Description

A possibility to get the page number when referencing an image or table would be good. Below is my own implementation that I use in Typst right now. It takes the label text and tries to find the label in the document. If the label is found it determines the page number of the label and formats it as a link to the label. In case the label is not defined it throws an error so it is not possible to compile the document with unresolved references when using pageref. 

```typst
#let pageref(labelText) = {
  locate(loc => {
    let label = label(labelText)
    let elems = query(selector(label), loc)

    if elems.len() == 1 {
      let pageNumber = elems.at(0).location().page()
      link(label, str(pageNumber))
    } else {
      let errorMessage = "Label '" + labelText + "' is not defined"
      panic(errorMessage)
    }
  })
}
```

### Use Case

Sometimes it can be handy to reference the page number in a document for example "As you can see in fig. x on page y".

mkorje:
Closes #2873.

Although now in 0.11 it is fairly simple to write a `pageref()` function yourself in Typst (see https://github.com/typst/typst/issues/2873#issuecomment-2138261314 by @laurmaedje), the consensus in the linked issue seems to be that this should be built-in. This PR does just that.

I'm not sure whether it should be scoped with `ref` or not. Currently it isn't, and can be called with `#pageref()`. Since `pageref()` can take any label (not just ones attached to elements that are referenceable), imo it doesn't quite make sense being `#ref.page()`.

I also added the function in https://github.com/typst/typst/issues/2873#issuecomment-2138261314 to the doc comment of `pageref()`. I think it is a good example of context, though maybe it would be better placed elsewhere.

### DIFF
--- a/crates/typst-layout/src/flow/collect.rs
+++ b/crates/typst-layout/src/flow/collect.rs
@@ -495,7 +495,7 @@ impl MultiSpill<'_, '_> {
         engine: &mut Engine,
         regions: Regions,
     ) -> SourceResult<(Frame, Option<Self>)> {
-        // The first region becomes unchangable and committed to our backlog.
+        // The first region becomes unchangeable and committed to our backlog.
         self.backlog.push(regions.size.y);
 
         // The remaining regions are ephemeral and may be replaced.
--- a/crates/typst-layout/src/inline/shaping.rs
+++ b/crates/typst-layout/src/inline/shaping.rs
@@ -774,7 +774,7 @@ fn shape_segment<'a>(
     buffer.guess_segment_properties();
 
     // By default, Harfbuzz will create zero-width space glyphs for default
-    // ignorables. This is probably useful for GUI apps that want noticable
+    // ignorables. This is probably useful for GUI apps that want noticeable
     // effects on the cursor for those, but for us it's not useful and hurts
     // text extraction.
     buffer.set_flags(BufferFlags::REMOVE_DEFAULT_IGNORABLES);
--- a/docs/guides/page-setup.md
+++ b/docs/guides/page-setup.md
@@ -433,7 +433,7 @@ frequently used for [figures]($figure.placement).
 ### Use columns anywhere in your document { #columns-anywhere }
 To create columns within a nested layout, e.g. within a rectangle, you can use
 the [`columns` function]($columns) directly. However, it really should only be
-used within nested layouts. At the page-level, the page set rule is preferrable
+used within nested layouts. At the page-level, the page set rule is preferable
 because it has better interactions with things like page-level floats,
 footnotes, and line numbers.
 


laurmaedje:
I think it is very close to `ref` and we should reflect that in the design. Both take a label and cross-reference it. `ref` requires the element to be referenceable because the default form requires a supplement and numbering. But the page form doesn't need that so it can have other requirements.

So, I think perhaps the best solution is not even `ref.page`, but rather `ref(<label>, form: "page")`. Similar to `cite(.., form: "..")`. And `ref.form` would be `"normal" | "page"`. What do you think?

laurmaedje:
> Do you think that the "page" form should allow the target to be an entry from the bibliography?

I'm not sure how useful it is, so I think I would skip it.

> Should the default supplement for the "page" form be none? Or, should it be similar to a figure's supplement where it is based on the kind and the active text language? So maybe the page element function has the argument supplement: auto added, which defaults to the equivalent of "page" in the set language. That way, users get (a perhaps more expected, as in, more inline with the "normal" form) "page X" (in their language) when using the "page" form.

Tricky question. But I general agree that I would expect it to behave similarly to normal refs.

EpicEricEE:
Can we use the translations from [babel](https://github.com/latex3/babel)?

mkorje:
> Can we use the translations from [babel](https://github.com/latex3/babel)?

We should be able to. I didn't want to go through babel and try to add them to this PR though until after its been reviewed.

I am also not sure whether for other languages the following are the same as in English: page number after the page word, capitalisation of the page word (*), the grammar. Though I suppose this more falls under #2485.

(*) In the current implementation, the page word "Page" is always capitalised. I don't know if this is the right default; for figures or equations it makes sense for the word to be capitalised. I guess ideally we automatically work out if `ref` is called at the start of the sentence to decide on the capitalisation.

crates/typst/src/layout/pages/run.rs
```diff
@@ -122,6 +123,10 @@ fn layout_page_run_impl(
     let header_ascent = PageElem::header_ascent_in(styles).relative_to(margin.top);
     let footer_descent = PageElem::footer_descent_in(styles).relative_to(margin.bottom);
     let numbering = PageElem::numbering_in(styles);
+    let supplement = match PageElem::supplement_in(styles) {
+        Smart::Auto => TextElem::packed(PageElem::local_name_in(styles)),
+        Smart::Custom(content) => content.unwrap_or_default(),
+    };
```
laurmaedje, line 129: I only just noticed while seeing this that referencing a page in a run where `set text(lang: "de")` from a page run where `set text(lang: "en")` will render `as we can see on Seite 10`. I'm not sure how much sense that makes. But it's similar with headings and stuff ...

crates/typst/src/model/reference.rs
```diff
@@ -95,22 +110,32 @@ use crate::text::TextElem;
...
+    /// the reference [`form`]($ref.form), and should return content.
     ///
     /// ```example
+    /// #set page(numbering: "1")
     /// #set heading(numbering: "1.")
-    /// #set ref(supplement: it => {
+    /// #set ref(supplement: (it, form) => {
+    ///   if form == "page" {
+    ///     return "p."
+    ///   }
```
laurmaedje, line 138: I think there's no need to pass `form` here as you can just write `show ref.where(form: "page"): set ref(supplement: ..)`. That way, it's no breaking change.

In fact, I'd like to get rid of allowing a function in general in the future, but that's left for later.

crates/typst/src/model/reference.rs
```diff
@@ -293,6 +374,16 @@ cast! {
     v: Func => Self::Func(v),
 }
 
+/// The form of the reference.
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash, Cast)]
+pub enum ReferenceForm {
```
laurmaedje, line 379:
```suggestion
pub enum RefForm {
```

tests/suite/model/ref.typ
```diff
@@ -55,6 +55,32 @@ $ A = 1 $ <eq2>
...
+See #ref(<setup>, form: "page").
+
+#set page(supplement: [p.])
+
+== Setup <setup>
+Text seen on #ref(<text>, form: "page").
+Text seen on #ref(<text>, form: "page", supplement: "page").
+
+--- ref-form-page-unambiguous ---
+// Test that page reference is not ambiguous.
```
laurmaedje, line 71: what do you mean with ambiguous here?

crates/typst/translations/en.txt
```diff
@@ -4,4 +4,5 @@ equation = Equation
 bibliography = Bibliography
 heading = Section
 outline = Contents
-raw = Listing
\ No newline at end of file
+raw = Listing
+page = Page
```
laurmaedje, line 8: I agree that it's not nice that the supplement will be in uppercase in English by default. "As seen on Page 10" feels wrong compared to "As seen on page 10".

laurmaedje (commented):
Handling suppelements properly is really tricky (not just for page references, more generally). With starting sentence, language, etc. Maybe there's a reason why LaTeX doesn't add a supplement by default ...

tests/suite/model/ref.typ
mkorje: I wanted to mirror the `ref-ambiguous` test: you can't page reference something from the bibliography, so if a label has the same name as something from the bibliography you won't get the error in `ref-ambiguous`.

crates/typst/src/layout/pages/run.rs
xkevio: It's the same as this issue, I think: https://forum.typst.app/t/how-to-have-different-supplements-for-the-same-heading-depending-on-the-location/667

crates/typst/src/model/reference.rs
mkorje: Good point, I'll update it.

I'd have thought functions might be a good way to go about solving https://github.com/typst/typst/issues/2485, though that then wouldn't really fit the `supplement` parameter I guess.

tests/suite/model/ref.typ
laurmaedje: i see

crates/typst/src/model/reference.rs
mkorje: Is there a nice where to break the line `show ref.where(form: "page"): \ set ref(supplement: ..)`? The compiler complains about expected expression, and nesting another function isn't great either.

crates/typst/translations/en.txt
mkorje: I'll change it to lowercase then. I guess the cases where it is at the start of the sentence are much rarer anyways. Implementing logic to change the case is out of the scope of this PR anyways.

crates/typst/src/model/reference.rs
laurmaedje: There is not.

laurmaedje:
```
show ref.where(
  form: "page"
): set ref(supplement: ..)
```
is the best you can do

laurmaedje: it is sometimes used in the docs I think

mkorje: Okay, I'll do that then.

crates/typst/src/layout/pages/run.rs
mkorje: I can see how someone might find it odd either way. I think it'd be best then to keep this consistent with headings and all for now.

mkorje: Also, this same behaviour (but with setting the supplement instead of setting languages) is shown in the `ref-form-page` test, I believe.

laurmaedje:
> We should be able to. I didn't want to go through babel and try to add them to this PR though until after its been reviewed.

I think we could add translations now.

mkorje:
I'll have a look at doing that.

mkorje:
A couple of things from my attempt at adding translations, using the info from babel.
- For languages with Latin alphabets, I've made the first letter lowercase. The babel translations files are inconsistent with the capitalisation, so I have no idea if this is correct for all these languages.
- [ar](https://raw.githubusercontent.com/latex3/babel/refs/heads/main/locale/ar/babel-ar.ini) and [he](https://raw.githubusercontent.com/latex3/babel/refs/heads/main/locale/he/babel-he.ini) are RTL I think, and I think they're the wrong way around in the files. I have no idea how to enter these properly.
- [ja](https://raw.githubusercontent.com/latex3/babel/refs/heads/main/locale/ja/babel-ja.ini) and [zh](https://raw.githubusercontent.com/latex3/babel/refs/heads/main/locale/zh/babel-zh.ini) have no entry for page.
- [tl](https://raw.githubusercontent.com/latex3/babel/530ccab2c74b9b9d1aad98a9221038fffdc2a111/locale-templates/tl/babel-tl.ini) has no translations at all.
- [zh-TW] I can't find.

mkorje:
Oh woops I managed to stuff up adding the translations to the files. I'll fix it in a bit.

crates/typst/translations/ar.txt
```diff
@@ -4,4 +4,5 @@ equation = معادلة
 bibliography = المراجع
 heading = الفصل
 outline = المحتويات
-raw = قائمة
\ No newline at end of file
+raw = قائمة
```
asibahi, line 7: What is "raw" in this context? because the Arabic word means "menu".

laurmaedje: a code listing from a backtick block

asibahi: Hah following the git blame it seems I am to blame, but this does not sound right. Apparently the etymology for "listing" is that it is, literally, a printed [list of lines](https://en.wikipedia.org/wiki/Listing_(computer)).

I do not actually know what a good Arabic word for that would be. قائمة (which *does* mean List, in my defense) is not it, however. Maybe مثال , "Example" ?

(In topic, the word for "page" is correct.)

JustForFun88: I got curious and did a bit of Googling, and it seems I found a suitable Arabic word: عادي. The phrase نص عادي is used in the Arabic Wikipedia to denote the term "plain" or "raw" text.

So, it's probably better to use the word عادي as it seems to be the closest and most commonly used term for conveying "plain" or "raw" text.

And the word قائمة really only shows images related to menus and lists😄

crates/typst/translations/fr.txt
```diff
@@ -4,4 +4,5 @@ equation = Équation
 bibliography = Bibliographie
 heading = Chapitre
 outline = Table des matières
-raw = Liste
\ No newline at end of file
+raw = Liste
+page = page
```
mkorje, line 8: Confirmed by @MDLC01 on discord.

crates/typst/translations/it.txt
```diff
@@ -4,4 +4,5 @@ equation = Equazione
 bibliography = Bibliografia
 heading = Sezione
 outline = Indice
-raw = Codice
\ No newline at end of file
+raw = Codice
+page = pag.
```
mkorje, line 8: Confirmed by eduard on discord.

crates/typst/translations/ja.txt
```diff
@@ -4,4 +4,5 @@ equation = 式
 bibliography = 参考文献
 heading = 節
 outline = 目次
-raw = リスト
\ No newline at end of file
+raw = リスト
+# page =
```
mkorje, line 8: Should be ページ, thanks to jimpjorps on Discord.

crates/typst/translations/ar.txt
```diff
@@ -4,4 +4,5 @@ equation = معادلة
...
+page = صفحة
```
mkorje, line 8: Confirmed by @asibahi.

crates/typst/translations/ar.txt
asibahi: عادي means regular. Has nothing to do with code, or code listings

I spent the day yesterday looking at Arabic computer papers, they either use شكل meaning figure or nothing at all.

crates/typst/translations/zh-TW.txt
```diff
@@ -4,4 +4,5 @@ equation = 式
 bibliography = 書目
 heading = 小節
 outline = 目錄
-raw = 程式
\ No newline at end of file
+raw = 程式
+# page =
```
peng1999, line 8: In Traditional Chinese we use `第 %d 頁`. See comments in `zh.txt` for more info.

crates/typst/translations/zh.txt
```diff
@@ -4,4 +4,5 @@ equation = 式
 bibliography = 参考文献
 heading = 小节
 outline = 目录
-raw = 代码
\ No newline at end of file
+raw = 代码
+# page =
```
peng1999, line 8: In Chinese we use `第 %d 页` to refer to a page. It roughly means `The nth page` in English. Actually the "heading" supplement text also need to be put *after* the ref number, i.e. `%d 小节` rather than `小节 %d`. So I suggest doing either of the following:

- Change logic to support both "before supplement" and "after supplement" so we can support zh & zh-TW correctly inside translation files. Or
- Set the `heading` and `page` translation to empty string, so ref becomes just number. And let users define the right function, or write the supplement text manually.

crates/typst/translations/ar.txt
JustForFun88: @asibahi But "plain text" or "raw text" also has nothing to do with code; it's just unformatted text.

In raw blocks, everything is rendered as is, so Typst doesn't evaluate or interpret anything inside them. Therefore, using "نص عادي" (plain text) seems appropriate here, as it conveys the idea of unprocessed text.

crates/typst/translations/zh.txt
mkorje: I think your first suggestion is ideal, and I'll have a look at implementing it. But it isn't in the scope of this PR.

I'm not sure the translations will be parsed properly if I set it to an empty string. Perhaps for now then I leave it as missing (so it will fallback to English) until a future PR solves this problem.

peng1999: > I think your first suggestion is ideal, and I'll have a look at implementing it.

I created #5102 so you can track it here.

crates/typst/translations/ar.txt
mkorje: Maybe it'd be best to create an issue for the translation of raw, and move this discussion there. Unless a correct translation has been decided upon, then I guess I can add it to this PR.

PR #4730, opened 2024-08-11
Improve `repr` for `type(none)` and `type(auto)`

### LINKED ISSUE #2747: Improve `repr(type(none))` and `repr(type(auto))`
### Description

Currently `type(none)` displays as `none` when debugging, which can be confusing (see https://discord.com/channels/1054443721975922748/1177092810176987291/1177092810176987291 ). Same goes for `type(auto)`, which displays as `auto`.

I propose we change it to `repr(type(none))` to output `type(none)` instead of `none`, to indicate explicitly that it's the type of `none`. Similarly, we could change `repr(type(auto))` to output `type(auto)`. 

Same would go for other possible constants behaving in this manner.

### Use Case

This would cause less confusion when working with types and debugging them.

LingkKang:
Resolves #2747


laurmaedje:
This changes `repr(auto)` to `type(auto)`, but the proposal in #2747 to change `repr(type(auto))` to `type(auto)` while keeping `repr(auto)` as `auto`, specifically so that the two can be distinguished.

LingkKang:
Sorry I misunderstood. I will fix this later.

laurmaedje:
Thanks!

PR #4733, opened 2024-08-11
Move `calc.nan` to `float.nan` and add `float.inf`

MDLC01:
See https://github.com/typst/typst/pull/4724#issuecomment-2282877907.

The first commit fixes the `scope` procedural macro for associated constants on primitive types.

Following what was said in https://github.com/typst/typst/pull/4724#issuecomment-2282877907, I decided to remove `calc.nan` immediately, as it is not possible to deprecate constants for now. However, I kept `calc.inf` both because it is used in some packages, and because I think it makes sense for this constant to stay available as part of the `calc` module.

I updated the `float` repr, and all examples, to use `float.nan` and `float.inf` instead of `calc.nan` and `calc.inf`.

This may slightly conflict with #4725.


laurmaedje:
Thank you!

PR #4736, opened 2024-08-12
Adjust the baseline of root radicand frame

### LINKED ISSUE #2809: Linebreak inside square root cause vertical offset of the square root symbol
### Description

There is a long formula inside a square root, I need to break the line to ensure proper display.
However, doing so seems to cause a vertical offset of the square root symbol.
### Code
```
$
123 + sqrt(a+b c+d)
$
```
### What I expected
image

### What I got
image

And sometimes it is not aligned to the bottom of the line, like
```
$
integral sqrt(a+b \ c+d)
$
```
image

### Reproduction URL

https://typst.app/project/pq8tcn5PssypKzd1FDWkUW

### Operating system

Web app, Windows, Linux

### Typst version

Leedehai:
Fixes #2809.

The cause of this issue is the radicand frame's baseline isn't adjusted so that it is aligned on its math axis height (defined by the font -- note not the middle line). Examples would be the `FracElem` and `BinomElem` in `frac.rs`, both of which contain a multi-line frame that is aligned on its math axis height.


laurmaedje:
Thanks!

PR #4737, opened 2024-08-12
Check cyclic-import when importing package

### LINKED ISSUE #4627: Better error message for circular dependencies on import/include
### Description

# Feature Request
I would like to have better error messaging / handling of "Circular Dependency detected between <doc_a> and <doc_b>" or similar. This is where documents / packages import/include each other in a circular/redundant fashion.

### Use Case

# Case
We have `pkg_a` and `pkg_b` which import each other in a "circular dependency fashion". Which in my opinion is not a great thing you know.. Because hierarchies are great at least according to Jordan Peterson. Because we should all trust a guy who only eats red meat and benzos.. Any ways. Circular dependencies implies infinity.. Which is not great. Because we don't have the time for that.

## pkg_a and pkg_b
```typst
// pkg_a imports
#import "@some_name_space/pkg_b:0.0.1" as b
```

```typst
// pkg_b imports
#import "@some_name_space/pkg_a:0.0.1" as a
```

## Feature request
Would be neat if the compiler would be able to detect this error and print a useful message pointing to the "offending" packages / documents for quicker debug. Because now we must "compile our way" to the offending documents.

## What message do I currently get?
```bash
# Command
typst compile ./some_doc_importing_pkg_a.typ

# Output
thread 'main' panicked at crate/typst/src/eval/mod.rs:48:9:
Tried to cyclicly evaluate /pkg_a.typ
```

Coekjan:
Fixes #4627

### DIFF
--- a/crates/typst/src/eval/import.rs
+++ b/crates/typst/src/eval/import.rs
@@ -182,6 +182,12 @@ fn import_package(vm: &mut Vm, spec: PackageSpec, span: Span) -> SourceResult<Mo
     // Evaluate the entry point.
     let entrypoint_id = manifest_id.join(&manifest.package.entrypoint);
     let source = vm.world().source(entrypoint_id).at(span)?;
+
+    // Prevent cyclic importing.
+    if vm.engine.route.contains(source.id()) {
+        bail!(span, "cyclic import");
+    }
+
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

PR #4738, opened 2024-08-12
Document exact sizes of mathematical spaces

### LINKED ISSUE #4715: Document exact size of mathematical spaces
The list on https://typst.app/docs/reference/layout/h/ mentions thin, med, thick, quad and wide, but not how large these spaces actually are.

ssotoen:
Closes #4715

### DIFF
--- a/crates/typst/src/layout/spacing.rs
+++ b/crates/typst/src/layout/spacing.rs
@@ -18,7 +18,8 @@ use crate::utils::Numeric;
 ///
 /// # Mathematical Spacing { #math-spacing }
 /// In [mathematical formulas]($category/math), you can additionally use these
-/// constants to add spacing between elements: `thin`, `med`, `thick`, `quad`, `wide`.
+/// constants to add spacing between elements: `thin` (1/6 em), `med`(2/9 em),
+/// `thick` (5/18 em), `quad` (1 em), `wide` (2 em).
 #[elem(title = "Spacing (H)", Behave)]
 pub struct HElem {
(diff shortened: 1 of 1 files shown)


crates/typst/src/layout/spacing.rs
```diff
@@ -18,7 +18,8 @@ use crate::utils::Numeric;
 ///
 /// # Mathematical Spacing { #math-spacing }
 /// In [mathematical formulas]($category/math), you can additionally use these
-/// constants to add spacing between elements: `thin`, `med`, `thick`, `quad`, `wide`.
+/// constants to add spacing between elements: `thin` (1/6em), `med`(2/9em),
```
MDLC01, line 21:
```suggestion
/// constants to add spacing between elements: `thin` (1/6 em), `med`(2/9 em),
```
Without a space, this can be interpreted as Typst code `1 / 6em`, which is invalid. It may be clearer with a space. Or maybe U+2159 ⅙ could be used, but I don't think there are equivalent characters for 2/9 and 5/18.

ssotoen: I originally wanted to do that (via fraction slash and narrow no-break space) but decided against it. There’s no guarantee the fractions display correctly on all devices, and the devs don’t seem to like having a space between number and unit.

ssotoen: [Resolved via Discord.](https://discord.com/channels/1054443721975922748/1088371919725793360/1272514010985795648)

PR #4739, opened 2024-08-12
Fix lexer on sth followed by comments

### LINKED ISSUE #4632: Single equal sign on a line is not parsed the same as equal sign directly followed by comment
### Description

A single equal sign on a line is parsed as a heading, as illustrated by the fact that the outline in the example below contains an entry:

```typ
#outline()
=
```
The outline contains an entry, indicating that there is a heading.

However, an equal sign directly followed by a comment, is not parsed as a heading:

```typ
#outline()
=//
```
The outline is empty, and the equal sign is displayed as text.

This is unexpected, although I'm not sure which of the two behaviors is right.

Tested on 0.11.1 and 0.11.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

Coekjan:
Fixes #4632.

I adjusted the function `space_or_end` for the fix. As this function is private, so only some  syntax nodes are affected in lexing process. They are
1. `SyntaxKind::HeadingMarker`
2. `SyntaxKind::ListMarker`
3. `SyntaxKind::EnumMarker`
4. `SyntaxKind::TermMarker`

I believe, when dealing with these syntax nodes (except for term-marker), comments after them **should be treated as spaces**. And, apparently, we don't need to deal with term-markers (`/`) followed by comments (`//` or `/*`). Thus I think my adjustment does no harm to the lexer and fixes the issue correctly.

### DIFF
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -503,7 +503,10 @@ impl Lexer<'_> {
     }
 
     fn space_or_end(&self) -> bool {
-        self.s.done() || self.s.at(char::is_whitespace)
+        self.s.done()
+            || self.s.at(char::is_whitespace)
+            || self.s.at("//")
+            || self.s.at("/*")
     }
 }
 
--- a/tests/suite/syntax/comment.typ
+++ b/tests/suite/syntax/comment.typ
@@ -34,6 +34,41 @@ Second part
 First part          //
 Second part
 
+--- issue-4632-sth-followed-by-comment ---
+// Test heading markers followed by comments.
+#test([
+  =// Comment
+  =/* Comment */
+], [
+  =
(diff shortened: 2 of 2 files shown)


MDLC01:
I could create another issue for that, but there is also the following case, which I did not mention in #4632 because I hadn't noticed it at the time.
```typ
// The four following lines should be parsed the same imo:
= /* */ Parsed as heading
= /* */Parsed as heading
=/* */Not parsed as heading
=/* */ Not parsed as heading

// Also, this may not be desirable, but I'm not sure:
= /*
*/ Parsed as heading
```

Coekjan:
> ```typ
> = /*
> */ Parsed as heading
> ```

I guess this is desired.

laurmaedje:
> I guess this is desired.

Yeah.

Issue #680, opened 2023-04-08
Benchmark suite to track performance regression 


ararunaufc:
The project is now in a very nice phase of a constant flow of new features. In order to avoid, or at least track, performance regressions we should start to think about having a benchmark suite.

This is of a great importance because we cannot rely only on the mental model of efficiency for Typst the the creators have, as it is not sustainable (even right now).

Also, the tool relies on a kind of computation model (constrained memorization) on which it is easy to break performance for new contributors. 

As the project is new and is not yet in a place to be used for "large scale" cases, we risk being trapped with an unscalable implementation that is hard to change. In my opinion, we should do it as soon as possible.

ararunaufc:
I found some benchmarking code in `tests/sec/benches.rs`. Is this enough for the time being, considering what I said before? Would this be the place to use for new ones?

user202729:
Interesting.

I think the current design is scalable though -- except for the state system, everything else is memoized.

(there's another issue #322 but that's not really an issue here)

laurmaedje:
I'm open to expanding the benchmarks. We use [`iai`](https://github.com/bheisler/iai) instead of [`criterion`](https://github.com/bheisler/criterion.rs) because with the latter the benchmarks just fluctuated too much to be useful at all. However, to be honest, I can't even run the benchmarks on my ARM MacBook anymore because iai depends on cachegrid, so I fully depend on my mental mode for efficiency right now. Improvements to the setup would definitely be welcome!

> except for the state system, everything else is memoized.

The state system is actually memoized, too!

epompeii:
Would we want to catch performance regressions in CI?

If so, I've built a benchmark tracking tool that may be of use: https://github.com/bencherdev/bencher
And `iai` support is in the works: https://github.com/bencherdev/bencher/pull/98

epompeii:
Just to follow up, `iai` support has been added to Bencher.

bluebear94:
Also note that there’s now a fork of `iai` called [`iai-callgrind`](https://github.com/iai-callgrind/iai-callgrind), which depends on Callgrind instead of Cachegrind.

epompeii:
If you decide to use `iai-callgrind` as @bluebear94 suggested, it is [also supported by Bencher](https://bencher.dev/docs/explanation/adapters/#-rust-iai-callgrind).

PR #4754, opened 2024-08-14
Add benchmarking of test suite

bluebear94:
See #680.

Currently uses `iai-callgrind` on the integration test cases, though we could add `criterion`-based benchmarks or dedicated benchmarks as well.

Due to limitations in `iai-callgrind`, running the benchmarks requires running `cargo run --bin prepare-benches` first to generate the fixture file – see iai-callgrind/iai-callgrind#207 for more details. Otherwise, the benchmarking code Works™, though it takes a good part of an hour to run through the whole test suite.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,7 +353,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.70",
 ]
 
 [[package]]
@@ -420,7 +420,7 @@ checksum = "c8936e42f9b4f5bdfaf23700609ac1f11cb03ad4c1ec128a4ee4fd0903e228db"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.70",
 ]
 
 [[package]]
@@ -541,7 +541,7 @@ checksum = "67e77553c4162a157adbf834ebae5b415acbecbeafc7a74b0e886657506a7611"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.70",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


laurmaedje:
Thanks for your work on this!

To get a better idea of where this is going, it would be great if you could write down your thoughts on the larger benchmarking strategy you envision: 
- How do we detect regressions? Should benchmarks run automatically regularly somehow?
- How should local benchmarks work? If it takes one hour, that's gonna slow down local iteration to the point that I personally wouldn't ever run it on the full suite. Maybe that's ok and one should just run on a subset. Though that raises the question which subset.
- Do we also want to benchmark watch compile down the road? Wouldn't be a blocker, but just wondering whether that's future compatible with the current approach.

Also, I feel like many of the tests aren't really great as benchmarks, so it's perhaps a bit wasteful to run them through it all the time. On the other hand, larger documents which are not part of the test suite at all _would_ be interesting benchmarks targets.

bluebear94:
>     * How do we detect regressions? Should benchmarks run automatically regularly somehow?

We could run the benchmark suite in CI or through something like [Bencher](https://bencher.dev/). The problem is that the IDs for the benchmarks are something like `my_id_####`, so adding another case to the test suite would throw off all of the IDs for the cases after the new one.

>     * How should local benchmarks work? If it takes one hour, that's gonna slow down local iteration to the point that I personally wouldn't ever run it on the full suite. Maybe that's ok and one should just run on a subset. Though that raises the question which subset.

It is possible to run `prepare-benches` with command-line arguments to filter cases (using the same syntax as the test harness), but the benchmark IDs won’t be stable.

>     * Do we also want to benchmark watch compile down the road? Wouldn't be a blocker, but just wondering whether that's future compatible with the current approach.

If we wanted to, then we could cover that in a separate file under `benches/`.

> Also, I feel like many of the tests aren't really great as benchmarks, so it's perhaps a bit wasteful to run them through it all the time. On the other hand, larger documents which are not part of the test suite at all _would_ be interesting benchmarks targets.

I agree. Maybe we should decouple the test suite from the benchmark suite, especially given the aforementioned problems. This would make most of this PR obsolete, but it could still be useful to libbify `typst-tests` (or perhaps move some of its functionality into `typst-kit`).

PR #4765, opened 2024-08-16
Add `clap::ValueHint` for input & output CLI arguments

### LINKED ISSUE #3281: Autocomplete in CLI doesn't work when using alias of command
### Description

```sh
typst w <TAB>
```
shows nothing in zsh. On bash/fish it shows all files

```sh
typst w -<TAB>
```
shows nothing

Upstream issue: https://github.com/clap-rs/clap/issues/4265

I'm not sure that this is the right issue, because title says about "alias", not "visible_alias"

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

### LINKED ISSUE #4156: `_typst` compdef file for zsh autocompletion doesn't allow for path completion
After running `nix build` I put `_typst` from `result/share/zsh/site-functions/_typst` into `~/.local/share/zsh/site-functions/_typst`. Then after typing `typst query` I invoke autocompletion for the Typst file path but all I get is:

```
--  input -- Path to input Typst file, use `-` to read input from stdin --
```

I suspect that the file is missing some switch that allows for path autocompletion in that place. Since this file is being generated (I think), then the library that puts it together has an bug or the use of that library is incorrect.

Coekjan:
Fixes #4156

With this patch, `typst compile/watch <TAB>` shows:

image

And closes #3281: It is because of #4156, the visible aliases give no hints. I tested that with this patch, the aliases `c` and `w` can give hints just like `compile` and `watch`.

image

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -6,7 +6,7 @@ use std::str::FromStr;
 
 use chrono::{DateTime, Utc};
 use clap::builder::{TypedValueParser, ValueParser};
-use clap::{ArgAction, Args, ColorChoice, Parser, Subcommand, ValueEnum};
+use clap::{ArgAction, Args, ColorChoice, Parser, Subcommand, ValueEnum, ValueHint};
 use semver::Version;
 
 /// The character typically used to separate path components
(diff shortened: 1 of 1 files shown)


PgBiel (approved):
Tested locally, and while the problem didn't seem to exist for Bash, it did for Zsh, and this PR fixed it. Thanks!

Andrew15-5:
I tested the new zsh compdef and it works amazing!

<details><summary></summary>

image

</details>

Andrew15-5:
One question, though: is the cut-off of the help message for 1 flag-per-line normal?

image

I can see the full message for every flag only if I maximize the terminal window. Shouldn't it wrap the long messages so it can all fit in the window (at least the top few flags)? Or is this configurable in the zsh?

Issue #1097, opened 2023-05-04
Multiple bibliographies


mareikep:
I am really excited to get to use typst, it looks great so far!

However, for larger documents it would be really helpful to have multiple bibliographies, e.g. to categorize publications into books, journals, etc. or to separate own prior publications from other peoples' works. This would be something like the `bibunit` or `chapterbib` packages in LaTeX. Unfortunately I cannot assign labels, but the [proposed Label] would be feature request

ayoubelmhamdi:
we need it for PhD too

wlupton:
I too have discovered that there can be only one `#bibliography`, but [its path argument can be an array](https://typst.app/docs/reference/meta/bibliography#parameters-path), so you can include multiple bibliographies. However, there can't be any duplicate keys (if there are duplicate keys, I think it would probably make sense for later definitions to override earlier ones; then you could list the "most definitive" bibliographies last).

kescobo:
One more vote for this! I got 80% of the way through rebuilding my CV in typst, which is a dream compared to LaTeX, but this might kill the effort :-/. I'm looking into manually building building bibliographies from the yaml, but I am not quite confident enough in the syntax to have high hopes there.

An interim solution (though I'm not sure if it's actually easier to implement) would be to enable the `style` of a `#cite` call to be the full reference. In other words, instead of 

```
== Some header

@label1
@label2

== Other header

@label3

#bibliography[...]
```

Giving 
![image](https://github.com/typst/typst/assets/3502975/ebe33d76-272e-4a07-ba00-8b61714ceaf0)

Maybe one could do something like

```
== Some header

#cite("label1", "label2", style: "full")

== Other Header

#cite("label3", style: "full")
```

To get

## Some header

[1] M. C. Woodruff, K. S. Bonham, et al., “Chronic inflammation, neutrophil activity, and
autoreactivity splits long Covid,” Nature Commun., vol. 14, no. 1, Jul. 2023, doi: 10.1038/
s41467-023-40012-7. [Online]. Available: https://doi.org/10.1038/s41467-023-40012-7

[2] A. A. Schoenborn, S. M. Yannarell, et al., “Microclimate is a strong predictor of the native and
invasive plant-associated soil microbiome on san cristóbal island, galápagos archipelago,”
Environmental Microbiology, Mar. 2023, doi: 10.1111/1462-2920.16361. [Online]. Available:
https://doi.org/10.1111/1462-2920.16361

## Other Header

[...]

PR #4768, opened 2024-08-16
Fix document set rules

### LINKED ISSUE #2710: Documents without content discard rules
A document containing no content but set rules like:
```typst
#set document(author: ("author 1", "author 2", "author 3"))
```

seem to not have any styles when going into layouting, resulting in those values not being written to the PDF. For this specific example, in the following code, `styles` returns an empty vector for `author` on L94.

https://github.com/typst/typst/blob/897447e8593f00971de3a966d3fb23cf2b1e0f92/crates/typst-library/src/meta/document.rs#L60-L99

See:

> Without the content like shown above, Typst will not write the Author information at all.

_Originally posted by @gfpj in https://github.com/typst/typst/issues/2556#issuecomment-1791111726_

### LINKED ISSUE #4065: Can't set document properties based on context

### LINKED ISSUE #4769: Set document conditionally
### Description

```typst
// Definition
#let authors = context document.author.len()

// Output
Authors: #authors

// Error A (no error and nothing in about dialog)
#set document(author: "test") if authors == 0

// Error B (no error and nothing in about dialog)
#if (authors == 0) { set document(author: "test"); [Done] }
```

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

laurmaedje:
This fixes the flakiness of document set rules by changing their semantics to deviate a bit from normal set rule semantics. Instead of being an immutable operation that just passes down information, document set rules are now basically naive setters that really just mutate the relevant piece of document information.

This means that
- the information is not lost for an empty document
- they can trivially depend on context
- we get rid of the "document set rules must appear before any content" error

Overall, this makes them less error-prone to use and since the document information is a singleton anyway, I think we don't lose much through them working a bit differently (i.e. not being scoped). Note that document set rules are still only allowed at the top-level, not within layout containers.

This also adds a mechanism for adding special checks for specific tests, so that we can test that the document information is indeed correctly collected.

Fixes #2710
Fixes #4065
Fixes #4769


Andrew15-5:
Huge! Now I don't need a separate `show: set-document` call inside `show: it => { if ... { return ... } ... }`, because I can use it inside context and if statement!

Now the only thing left to do is fix https://github.com/typst/typst/issues/2326 so that I can say the same thing about `show: set-page`. After that, everything (well, except for the https://github.com/typst/typst/issues/1097) can be simplified into a single function (with if statements): https://github.com/typst/typst/issues/4502. Multiple exact same bibliography calls isn't strictly necessary as I can use another workaround, but it would've been much easier to call the `bibliography` with the same argument however many times (requires 1 LoC instead of complex workarounds) and not have the error (I have `show bibliography: none` due to custom bibliography logic).

Issue #2411, opened 2023-10-17
Improve layout of multi-line headings in outline


deffi:
Example document:

```
#set heading(numbering: "1.1")

#outline(indent: auto)

= One
== One point one: #lorem(20)
== One point two
```

Result for the outline (Typst 0.8):
![Typst result](https://github.com/typst/typst/assets/470243/53cad5a4-d3e8-4fa6-8a9f-769ad8a47e1c)

It seems like this could be improved by aligning subsequent lines with the text of the first line, or at least with the heading number.

LaTeX aligns with the beginning of the text:

![LaTeX result](https://github.com/typst/typst/assets/470243/5bdeef68-5e6c-452e-be08-f9c4a1ac0996)

MS Word 2016 aligns with the number:

![Word result](https://github.com/typst/typst/assets/470243/dc84223e-e4a6-4d99-a050-0751c1650857)

Possibly related to #2410 (similar issue, but for the heading itself instead of the outline). Also seen in #711, though that one is about alignment of the numbers.

NakanoMiku39:
Hello, I would like to have a try? First time trying to solve issue :)

Atreyu-94:
In my opinion, the way he uses latex feels much more professional.

Atreyu-94:
In my opinion, the way he uses latex feels much more professional.

jkunimune:
I think it would also be worth imitating how LaTeX handles the right edge LaTeX, where it ends every line to the left of the page numbers.  Having a column on the right dedicated to numbers makes it look a lot better imo, and implementing that in Typst would also automatically solve Issue #2048 (since there would always be room for the page number on every line).

zR-JB:
Maybe it would make sense to split the body of OutlineEntry (https://github.com/typst/typst/blob/c4dd6fa06227b8c4a04cf488f371f52d5379ef9b/crates/typst/src/model/outline.rs#L423C1-L427C23) with numbering and then the actual body (for a heading the heading title) in two parts, which would then make it possible to wrap the actual body in its own box, without the numbering.

PR #4775, opened 2024-08-16
Render outline entry indent as part of entry show rule

### LINKED ISSUE #3972: Outline indent invalid for numbered headings with level > 1
### Description

Outline entries are rendered unexpected with certain conditions:

Preconditions: 
- manipulate spacing between entries
- set `outline(indent: auto)` - or something not `none`
- have level 2 headings with numbering

Result:
- invalid top spacing of level 2 heading entires with numbering
- no indent of entries (even if indent is expected)

Screenshot:
Screenshot 2024-04-21 at 12 04 49

Example:
```typ
// set spacing between elements to 0
#show outline.entry: it => {
  v(0em, weak: true)
  it
}
// set indent to auto
#outline(
  indent: auto,
)

// content
#hide[
#set heading(numbering: "1.")
= Numbered Headings
== Invalid Top Spacing

#set heading(numbering: none)
= No Numbering
== Valid Top Spacing
]
```

### Reproduction URL

https://typst.app/project/wx_IA9ekp5PKz8zu2scU17

### Operating system

MacOS

### Typst version

MDLC01:
Fixes #3972.

The issue was caused by the indent being rendered before the entry, and therefore before the vertical spacing added by the user's show rule.


PgBiel:
Keeping indent outside of the show rule was intentional, to ensure show rules on outline.entry can't accidentally remove it. As is, this will be a breaking change on existing show rules. I wonder if there's some other way to solve this?

MDLC01:
I think the indent could be included as part of the `body` to make the change non-breaking, but I'm not sure if this is desirable. When making a show-rule, users are expected to know that they should be careful to reconstruct the element properly if they want to reproduce the default behavior.

laurmaedje:
I'm not sure it is so clear cut that the indent should be part of the entry. The property is on the outline after all. This means it is not even available on the entry. Moreover, the indent depends on parent entries. I think it's actually rather tricky to reproduce in an entry show rule.

MDLC01:
> The property is on the outline after all. This means it is not even available on the entry.

With this PR, each outline entry has an `indent` field that is some content that should be used as the indent.

laurmaedje:
Okay, fair enough. 

Still, I'm not sure it makes more sense than the status quo conceptually.

MDLC01:
Another solution could be to use a horizontal stack containing the `v` in the first cell, and the outline entry in the second one. But this would change the appearance in the case of multi-line headings:
image
would become
image

laurmaedje:
I mean, the second looks better to me. 😅

But, then a `pad` would probably better than a stack, wouldn't it?

laurmaedje:
There's actually an open issue about this: https://github.com/typst/typst/issues/2411

Though the optimum would look a bit different still (the LaTeX way, not the Word way).

MDLC01:
Then, I will close this PR and open a new one that implements the LaTeX way.

MDLC01:
Actually, this was less trivial than I expected. I'm not sure I will have the time to do it. If someone else wants to give it a go, consider this as available.

PR #4776, opened 2024-08-17
Improve documents for the `--font-path` arg & `TYPST_FONT_PATHS` env

### LINKED ISSUE #1708: Improved TYPST_FONT_PATHS documentation
_Originally posted by @luochen1990 in https://github.com/typst/typst/issues/649#issuecomment-1507317901_:

> Hoping for a more detailed description about this ENV variable,  e.g.   WHAT is the expected seperator when there is more than one path;   Which directory level should this PATH point to,  'xxx/share/fonts/opentype/SourceCodePro-Black.otf'  or  'xxx/share/fonts/opentype'  or 'xxx/share/fonts' ?

I found I had the same question when trying to resolve font issues. The online documentation only mentions this env var in a brief CHANGELOG note.

Coekjan:
Closes #1708.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -385,7 +385,10 @@ pub struct FontsCommand {
 /// Common arguments to customize available fonts
 #[derive(Debug, Clone, Parser)]
 pub struct FontArgs {
-    /// Adds additional directories to search for fonts
+    /// Adds additional directories that are recursively searched for fonts
+    ///
+    /// If multiple paths are specified, they are separated by the system's path
+    /// separator (`:` on Unix-like systems and `;` on Windows).
     #[clap(
         long = "font-path",
         env = "TYPST_FONT_PATHS",


crates/typst-cli/src/args.rs
```diff
@@ -386,6 +386,12 @@ pub struct FontsCommand {
 #[derive(Debug, Clone, Parser)]
 pub struct FontArgs {
     /// Adds additional directories to search for fonts
+    ///
+    /// The environment variable `TYPST_FONT_PATHS` is used as a fallback if
```
laurmaedje, line 390: Right now it says 
```
      --font-path <DIR>
          Adds additional directories to search for fonts
          
          [env: TYPST_FONT_PATHS=]
```
so this feels redundant. Especially since the same would apply to TYPST_ROOT otherwise, so it would be inconsistent with this change set.

The separator thing is still relevant though.

crates/typst-cli/src/args.rs
```diff
@@ -386,6 +386,12 @@ pub struct FontsCommand {
...
+    /// `--font-path` is not specified.
+    ///
+    /// Paths specified via `--font-path` and `TYPST_FONT_PATHS` are separated by
+    /// the system's path separator (`:` on Unix-like systems and `;` on Windows).
```
laurmaedje, line 394: I would reword a bit to something like "If multiple paths are specified, they are separated by ...". I think we can skip the "via `--font-path` and `TYPST_FONT_PATHS`".

crates/typst-cli/src/args.rs
```diff
@@ -386,6 +386,12 @@ pub struct FontsCommand {
 #[derive(Debug, Clone, Parser)]
 pub struct FontArgs {
     /// Adds additional directories to search for fonts
```
laurmaedje, line 388:
```suggestion
    /// Adds additional directories that are recursively searched for fonts
```
maybe this will help with the person having this problem:
> Which directory level should this PATH point to, 'xxx/share/fonts/opentype/SourceCodePro-Black.otf' or 'xxx/share/fonts/opentype' or 'xxx/share/fonts' ?

crates/typst-cli/src/args.rs
Coekjan: Thanks for your kind review. I have removed this. :heart:

laurmaedje:
Thanks!

PR #4783, opened 2024-08-18
Add warnings when labels are ignored or unattached

### LINKED ISSUE #4592: Having multiple labels on an object ignores all but the last silently

mkorje:
Closes #4592.

Adds a warning when a label is ignored by the compiler. I also added a warning for when a label isn't attached to anything.

In my opinion these should both be errors (or maybe only when a label isn't attached to anything). Though I don't know if there's any consensus on this.


crates/typst/src/eval/markup.rs
```diff
@@ -52,7 +55,22 @@ fn eval_markup<'a>(
                     if let Some(elem) =
                         seq.iter_mut().rev().find(|node| !node.can::<dyn Unlabellable>())
                     {
+                        if let Some(prev_label) = elem.label() {
+                            vm.engine.sink.warn(warning!(
+                                prev_label_expr.unwrap().span(),
```
laurmaedje, line 60: I think this unwrap can panic in practice. The previous label needn't come from the same content block, it could have been applied anywhere in a different function call.

laurmaedje: Not necessarily, consider this: `#[#block()<a>]<b>`

laurmaedje: It would break hash and equality identity. There would be ways around this I guess, but I don't think it would be a good change. It also increases the size a lot for a type that is intentionally very light-weight (just u32).

laurmaedje:
Thanks!

PR #4794, opened 2024-08-20
Share allocations for singletons

laurmaedje:
No need to allocate for spaces, linebreaks, parbreaks, and alignment points over and over again.


Dherse:
@laurmaedje did you test the performance impact? It's a nice change and I wonder if it has any measurable impact!

laurmaedje:
I was too lazy. But if you end up testing, I'd be curious!

Dherse:
I might try, but I have a thing after work so maybe tomorrow evening :)

Issue #2749, opened 2023-11-23
Skew transform


Enivex:
### Description

Typst currently has https://typst.app/docs/reference/layout/rotate/ and https://typst.app/docs/reference/layout/scale/, but no skew transform. While it can be expressed using a combination of the former, it's not trivial to get right

```
#let skew(angle,vscale: 1,body) = {
  let (a,b,c,d)= (1,vscale*calc.tan(angle),0,vscale)
  let E = (a + d)/2
  let F = (a - d)/2
  let G = (b + c)/2
  let H = (c - b)/2
  let Q = calc.sqrt(E*E + H*H)
  let R = calc.sqrt(F*F + G*G)
  let sx = Q + R
  let sy = Q - R
  let a1 = calc.atan2(F,G)
  let a2 = calc.atan2(E,H)
  let theta = (a2 - a1) /2
  let phi = (a2 + a1)/2
  
  set rotate(origin: bottom+center)
  set scale(origin: bottom+center)
  
  rotate(phi,scale(x: sx*100%, y: sy*100%,rotate(theta,body)))
}

```
(The vscale parameter is strictly speaking unnecessary)

### Use Case

A couple of trivial examples are fake italic text and shadows

```
#let fake-italic(body) = skew(-12deg,body)

#let shadowed(body) = box(place(skew(-50deg, vscale: 0.8, text(fill:luma(200),body)))+place(body))

#fake-italic[This is fake italic text]

#shadowed[This is some fancy text with a shadow]
```

![rendered](https://github.com/typst/typst/assets/2191754/56564bfe-c492-4010-b165-3828e4a1c635)

csimide:
Hello, I would like to use the skew transform implementation code in this issue to add fake italics support to [cuti](https://github.com/csimide/cuti/), a package released under the MIT license. Would it be possible for me to utilize this code for that purpose?

Enivex:
> Hello, I would like to use the skew transform implementation code in this issue to add fake italics support to [cuti](https://github.com/csimide/cuti/), a package released under the MIT license. Would it be possible for me to utilize this code for that purpose?

Feel free

laurmaedje:
Resolved by https://github.com/typst/typst/pull/4803

PR #4803, opened 2024-08-21
Add a skew function

Bzero:
This PR addresses issue  #2749 and adds a skew transformation function. 

The function takes horizontal and/or vertical skewing angles `ax` and `ay` to skew the content  in the horizontal and vertical directions. If only one angle is given it is taken as the horizontal angle as I would think this will be the more common use case of this function. The remaining arguments `origin` and `reflow` behave similar to the other transformation functions. 

I refrained from introducing the `vscale` parameter present in the example in #2749 to keep the transformation more atomic and since [scale](https://typst.app/docs/reference/layout/scale/) can be easily used to scale the content after skewing.


crates/typst/src/layout/transform.rs
```diff
@@ -335,6 +335,115 @@ cast! {
     length: Length => ScaleAmount::Length(length),
 }
 
+/// Skews content without affecting the layout.
+///
+/// Skews an element in horizontal and/or vertical direction. The layout will
+/// act as if the element was not skewed unless you specify `{reflow: true}`.
```
LaurenzV, line 341: Aren't the first and third sentence somehwat contradictory? Since it will affect the layout if `reflow` is set to `true`.

Bzero: Well yes, agreed this is not a clear formulation, I oriented myself at the other transformation functions which have similar descriptions. 

I think I would just remove "without affecting the layout".

crates/typst/src/layout/transform.rs
```diff
@@ -335,6 +335,115 @@ cast! {
...
+    #[default(Angle::zero())]
+    pub ax: Angle,
+
+    /// The vertical skewing angle.
+    ///
+    /// ```example
+    /// #skew(ay: 30deg)[Skewed]
+    /// ```
+    ///
+    #[parse(args.named("ay")?.or(args.find()?))]
```
laurmaedje, line 365: If we want to support a positional shorthand where both are the same (not sure whether it makes sense), we should do it like in `scale` and share the same result of `args.find()?`. Allowing just named or positional, but for both independently, is not great in my opinion.

crates/typst/src/layout/transform.rs
```diff
@@ -335,6 +335,115 @@ cast! {
...
+    #[default(Angle::zero())]
+    pub ay: Angle,
+
+    /// The origin of the skew transformation.
+    ///
+    /// The origin will stay fixed during the operation.
+    ///
+    /// ```example
+    /// #set text(spacing: 16pt)
```
laurmaedje, line 374: I think this example is too complex for the docs. The image it generates is also not really instructive. Maybe we can find something simpler. Also, please use `kebab-case`.

tests/suite/layout/transform.typ
```diff
@@ -115,3 +115,53 @@ Hello #scaled[World]!
 #scale(x: auto, y: 50pt, reflow: true, cylinder)
 #scale(x: 100pt, y: auto, reflow: true, cylinder)
 #scale(x: 150%, y: auto, reflow: true, cylinder)
+
+--- transform-skew ---
```
laurmaedje, line 119: No need to increase the text and page size so much in all these tests. It bloats the size of the resulting images. It's intentional that everything is small and as a result a bit pixelated. I know that it makes them a bit harder to look at, but that's how it is (we want to keep the repository size from growing too much). For testing, you can always pass `--scale 2` or `--scale 4` to the test runner to render at a higher resolution.

tests/suite/layout/transform.typ
```diff
@@ -115,3 +115,53 @@ Hello #scaled[World]!
...
+}
+
+--- transform-skew-origin ---
+// Test setting skewing origin.
+#set page(width: 128pt, height:64pt)
+#set text(spacing: 30pt)
+#let square = square.with(width: 20pt)
+#let skew_square(origin) = box(place(square(stroke: gray))
+  + place(skew(-30deg, -30deg, origin: origin, square())))
+#skew_square(center+horizon)
```
laurmaedje, line 149: Please use kebab-case :)

crates/typst/src/layout/transform.rs
Bzero: Yes, I see this. My thought here was to show both the original and the skewed square to illustrate where the fixed point is but I see that this is too cumbersome for the docs. 

How about this simpler example where still some unchanged characters are shown for comparison:

```typst
X #box(skew(ax: -30deg, origin: center + horizon)[X]) X \
X #box(skew(ax: -30deg, origin: bottom + left)[X]) X \
X #box(skew(ax: -30deg, origin: top + right)[X]) X
```

tests/suite/layout/transform.typ
Bzero: Okay, I will shrink the size of the reference images.

laurmaedje:
Thank you, looking great!

Bzero:
Welcome!

PR #1636, opened 2023-06-30
remove broken release targets: aarch64-pc-windows-msvc and riscv64gc-unknown-linux-gnu

AlistairKeiller:
With the addition of the ureq library, the release CI build for aarch64-pc-windows-msvc and riscv64gc-unknown-linux-gnu currently fails because of ureq's dependency ring. Ring has long had problems with these targets ( https://github.com/briansmith/ring/issues/1182, https://github.com/briansmith/ring/issues/1167 ), and it seems like a difficult to fix artifact of ring's c, rust, and assembly codebase / build system. This PR removes those failing targets from the release build, a decision supported by @laurmaedje ( https://discord.com/channels/1054443721975922748/1088371867913572452/1124441928688209992 ).

PR #4821, opened 2024-08-24
Add back release targets `aarch64-pc-windows-msvc` and `riscv64gc-unknown-linux-gnu`

### LINKED ISSUE #4210: Provide native binaries for Windows on ARM
### Description

Qualcomm, MS, and OEM partners have announced new generation of laptops/devices that use the X Elite chips which are ARM based. I am opening this issue to see if the devs can provide a native version for Windows on ARM. 

### Use Case

This will provide maximum performance instead of going through emulation layer.

mkorje:
Closes #4210.

The two release targets `aarch64-pc-windows-msvc` and `riscv64gc-unknown-linux-gnu` build successfully now (see https://github.com/mkorje/typst/actions/runs/10537219450).

### DIFF
--- a/crates/typst-cli/Cargo.toml
+++ b/crates/typst-cli/Cargo.toml
@@ -83,5 +83,6 @@ pkg-url = "{ repo }/releases/download/v{ version }/typst-{ target }{ archive-suf
 bin-dir = "typst-{ target }/typst{ binary-ext }"
 pkg-fmt = "txz"
 
-[package.metadata.binstall.overrides.x86_64-pc-windows-msvc]
-pkg-fmt = "zip"
+[package.metadata.binstall.overrides]
+x86_64-pc-windows-msvc = { pkg-fmt = "zip" }
+aarch64-pc-windows-msvc = { pkg-fmt = "zip" }


Coekjan:
`pkg-fmt` of `aarch64-pc-windows-msvc` should be specified like `x86_64-pc-windows-msvc`, I guess.

https://github.com/typst/typst/blob/88325d7d019fd65c5177a92df4347ae9a287fc19/crates/typst-cli/Cargo.toml#L86-L87

Coekjan:
I notice that existing targets for linux are usually something like `*-musl*`, which means they use musl libc. Whereas the added `riscv64gc-unknown-linux-gnu` uses gnu libc. Seems lack of consistency.

For consistency, we might use `riscv64gc-unknown-linux-musl` instead. But unfortunately, the musl one for riscv64 (in [tier 3](https://doc.rust-lang.org/rustc/platform-support.html#tier-3)) is not built for official rust toolchain. So, if we do plan to pre-build typst for riscv64 platform, we shall use the gnu one.

... Not sure if we could tolerate such inconsistency.

Coekjan:
> Using `-gnu` matches how it was before #1636. I have no idea whether this inconsistency is good or bad.

In #1636, pre-building windows for aarch64 uses `cross: true`. But this PR uses `cross: false`. Is this a typo?

laurmaedje:
Thanks!

Coekjan:
News: `riscv64gc-unknown-linux-musl` will be promoted to tier 2[^1] since rust 1.82.0.

If we prefer musl libc, we could switch to it in the near future.

[^1]: https://github.com/rust-lang/rust/pull/122049

PR #4822, opened 2024-08-24
Animation-friendly export

### LINKED ISSUE #4384: Animation-friendly export
### Description

Making the export more animation-friendly, e.g. by exposing some way to generate groups with well-known identifiers in SVG export, which external tools can pick up.

### Use Case

Ideally, users can label `Typst` object (`box` or `block` for now):
```typst
#box(fill: white, stroke: black)[= Typst] <title>
```

Then during an Svg export, `Typst` will add the name of these label as an identifier in the Svg:
```svg
<rect id="title" x="10" y="10" width="100" height="100" />
```

nixon-voxell:
Resolves #4384

### DIFF
--- a/crates/typst/src/layout/frame.rs
+++ b/crates/typst/src/layout/frame.rs
@@ -376,7 +376,6 @@ impl Frame {
         let outset = outset.relative_to(self.size());
         let size = self.size() + outset.sum_by_axis();
         let pos = Point::new(-outset.left, -outset.top);
-
         self.prepend_multiple(
             styled_rect(size, radius, fill, stroke)
                 .into_iter()
@@ -563,10 +562,6 @@ impl GroupItem {
             clip_path: None,
             label: None,
         }
-    }
-
-    pub fn set_label(&mut self, label: Label) {
-        self.label = Some(label);
     }
 }
 
--- a/crates/typst/src/layout/frame.rs
+++ b/crates/typst/src/layout/frame.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 
 use smallvec::SmallVec;
 
-use crate::foundations::{cast, dict, Dict, StyleChain, Value};
+use crate::foundations::{cast, dict, Dict, Label, StyleChain, Value};
 use crate::introspection::Tag;
 use crate::layout::{
     Abs, Axes, Corners, FixedAlignment, HideElem, Length, Point, Rel, Sides, Size,
@@ -376,11 +376,12 @@ impl Frame {
         let outset = outset.relative_to(self.size());
         let size = self.size() + outset.sum_by_axis();
         let pos = Point::new(-outset.left, -outset.top);
+
         self.prepend_multiple(
             styled_rect(size, radius, fill, stroke)
                 .into_iter()
                 .map(|x| (pos, FrameItem::Shape(x, span))),
-        )
+        );
     }
 
     /// Arbitrarily transform the contents of the frame.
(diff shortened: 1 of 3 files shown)


crates/typst/src/foundations/content.rs
```diff
@@ -837,6 +837,11 @@ impl<T: NativeElement> Packed<T> {
         self.0.label()
     }
 
+    /// Set the label of the element.
```
laurmaedje, line 840: This seems like it's not used by the PR.

crates/typst/src/layout/frame.rs
```diff
@@ -34,6 +34,8 @@ pub struct Frame {
     ///
     /// Determines whether it is a boundary for gradient drawing.
     kind: FrameKind,
+    /// An optional label attached to the frame.
+    label: Option<Label>,
```
laurmaedje, line 38: I want to try and keep frames as lightweight and conceptually simple as possible (the existence of `FrameKind` is actually already unfortunate), so I would prefer for this to be in `GroupItem`. You can generate it via `frame.group(..)`. You can search for references to this function to see how it works.

crates/typst/src/visualize/image/mod.rs
```diff
@@ -257,6 +257,10 @@ fn layout_image(
         frame.clip(Path::rect(frame.size()));
     }
 
+    if let Some(label) = elem.label() {
+        frame.set_label(label);
```
laurmaedje, line 261: Why also on image. Didn't we want to keep it to box and block?

crates/typst/src/layout/frame.rs
nixon-voxell: I gave it a go on this approach, I had to make `frame.group(..)` public and also the produced result has more groups than the original ones because of the encapsulation:

using `frame.group`

```
group:
  group: box-label
    group:
group:
  group: block-label
    group:
```

labeling `Frame` directly:

```
group:
  group: box-label
group:
  group: block-label
```

crates/typst/src/visualize/image/mod.rs
nixon-voxell: Alright, was experimenting, I will remove it.

crates/typst/src/layout/frame.rs
```diff
@@ -558,8 +561,13 @@ impl GroupItem {
             frame,
             transform: Transform::identity(),
             clip_path: None,
+            label: None,
         }
     }
+
+    pub fn set_label(&mut self, label: Label) {
+        self.label = Some(label);
+    }
```
laurmaedje, line 570: No need for this, since the field is public.

crates/typst/src/layout/frame.rs
```diff
@@ -376,11 +376,12 @@ impl Frame {
         let outset = outset.relative_to(self.size());
         let size = self.size() + outset.sum_by_axis();
         let pos = Point::new(-outset.left, -outset.top);
+
```
laurmaedje, line 379: the semicolon is fine, but no need for this change

crates/typst/src/layout/frame.rs
nixon-voxell: Update:

I noticed there aren't more groups, it just shifted up 🤣 :

using `frame.group`

```
group:
  group: box-label
    group:
group:
  group: block-label
    group:
```

labeling `Frame` directly:

```
group:
  group:
    group: box-label
group:
  group:
    group: block-label
```

laurmaedje:
Thanks!

PR #4825, opened 2024-08-24
Add checkmark ballot symbols

MDLC01:
This small PR adds the two checkmarked ballot symbols as `ballot.check` and `ballot.check.heavy`.

Unicode calls `ballot.check.heavy` "BALLOT BOX WITH BOLD CHECK", but I think using "heavy" is more consistent here.

Another question is whether we should rename `ballot.x` to `ballot.cross` for more consistency.

I also debated whether to add U+237B ⍻ NOT CHECK MARK, but decided not to because it seems to be an old and unused symbol.[^1]

[^1]: https://en.wiktionary.org/wiki/%E2%8D%BB

### DIFF
--- a/crates/typst/src/symbols/sym.rs
+++ b/crates/typst/src/symbols/sym.rs
@@ -476,7 +476,7 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     yen: '¥',
 
     // Miscellaneous.
-    ballot: ['☐', x: '☒'],
+    ballot: ['☐', x: '☒', check: '☑', check.heavy: '🗹'],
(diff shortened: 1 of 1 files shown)


laurmaedje:
Yeah I guess `ballot.cross` makes more sense with `checkmark` and `crossmark`

MDLC01:
I renamed `ballot.x` to `ballot.cross`. I decided not to keep `ballot.x` because I don't think many people used this symbol anyway, and it's an easy fix. This is a breaking change, though.

laurmaedje:
It's okay. Thanks!

PR #4826, opened 2024-08-24
Add more `emptyset`/`nothing` symbol variants

MDLC01:
This small PR adds `bar`, `circle`, `arrow.l` and `arrow.r` variants to `emptyset` and `nothing`.

I'm not sure about the name for the `bar` variant. Maybe `line`, `overbar`, or `overline` would be better? The Unicode name is "EMPTY SET WITH OVERBAR." At the same time, I don't think there is an empty set symbol with a bar somewhere other than above as of Unicode 15.1.

Additionally, maybe `arrow.l` and `arrow.r` should be renamed to `larrow` and `rarrow`, because what is going left or right is the arrow only. This is similar to how, e.g., `subset.neq` and `subset.eq.not` are different symbols.


Enivex:
Are these actually used by anyone? They seem very obscure

That being said, I don't mind eventually having names for every single unicode symbol.

MDLC01:
I don't think they are frequently used, but I also don't think those names could correspond to anything else. Having names for more symbols makes it easier for people who want to use those symbols, and I don't think it causes any issue if the names are specific enough that they won't block any future use. Binary size may be an issue in the long run, but I don't think it is noticeable.

Coekjan:
Just curious. Are `emptyset` and `nothing` mapped to same symbols?

MDLC01:
Yes. `emptyset` and `nothing` are both aliases for the same symbol and variants.

Coekjan:
> Yes. `emptyset` and `nothing` are both aliases for the same symbol and variants.

Is it common for multiple symbol names to map to the same symbol? It might be better to use a special syntax to simplify this kind of implementation, I think.

MDLC01:
I don't think it is very common. The only such symbols I can think of are `emptyset`/`nothing` and `infinity`/ `oo`.

MDLC01:
I just though about the fact that maybe `arrow.l` and `arrow.r` should be renamed to `larrow` and `rarrow`, because what is going left or right is the arrow only. This is similar to how, e.g., `subset.neq` and `subset.eq.not` are different symbols.

laurmaedje:
> I just though about the fact that maybe arrow.l and arrow.r should be renamed to larrow and rarrow, because what is going left or right is the arrow only.

Hmmm, yeah, I guess that makes sense.

laurmaedje:
I guess [my comment about `bar`](https://github.com/typst/typst/pull/4834#issuecomment-2310500740) also sort of applies here. Not sure.

MDLC01:
Indeed. But this bar is also different. What about `overline`? Or maybe just `line`? Unicode calls it an "overbar."

laurmaedje:
I guess `bar` is ok. I hope they don't invent one where the bar is randomly stuck to the side.

laurmaedje:
Thanks!

MDLC01:
> I hope they don't invent one where the bar is randomly stuck to the side.

Let's hope so!

PR #4827, opened 2024-08-25
Add more `angle` symbol variants

MDLC01:
This PR adds multiple variants to the `angle` symbol.

## Added variants

`angle.l.curly` and `angle.r.curly` are curly variants of `angle.l` and `angle.r` respectively.

`angle.l.dot` and `angle.r.dot` are U+2991 ⦑ LEFT ANGLE BRACKET WITH DOT and U+2992 ⦒ RIGHT ANGLE BRACKET WITH DOT respectively.

`angle.s` is U+299E ⦞ ANGLE WITH S INSIDE. I'm not sure how used this symbol is in practice, and I don't like that the symbol name is a single character long for something so specific. I included it in the first version of this PR mostly for completeness.

`angle.inv` is a horizontally-flipped version of `angle` (more on that below).

`angle.oblique` and `angle.oblique.inv` are U+29A6 ⦦ OBLIQUE ANGLE OPENING UP and U+29A7 ⦧ OBLIQUE ANGLE OPENING DOWN respectively.

`angle.arc.arrow` is U+29A8 ⦨ MEASURED ANGLE WITH OPEN ARM ENDING IN ARROW POINTING UP AND RIGHT, and `angle.arc.inv.arrow`, `angle.arc.rev.arrow` and `angle.arc.inv.rev.arrow` are similar variants.

Note that `angle.acute` does not have an `inv` variant because the corresponding character does not exist in Unicode (as far as I can tell).

## Unresolved issues

1. The `s` variant has a very short name for something so specific.

2. I'm not satisfied with the name for the `inv` variant because other `inv` variants (such as `quest.inv`) turn the symbol upside-down, whereas here we mirror the symbol around the horizontal axis. The `arc.inv.rev.arrow` variant name is also very unfortunate.

3. Four arrow variants are missing because I don't know how to name them. They are:
    - U+29AC ⦬ MEASURED ANGLE WITH OPEN ARM ENDING IN ARROW POINTING RIGHT AND UP,
    - U+29AD ⦭ MEASURED ANGLE WITH OPEN ARM ENDING IN ARROW POINTING LEFT AND UP,
    - U+29AE ⦮ MEASURED ANGLE WITH OPEN ARM ENDING IN ARROW POINTING RIGHT AND DOWN, and
    - U+29AF ⦯ MEASURED ANGLE WITH OPEN ARM ENDING IN ARROW POINTING LEFT AND DOWN.

4. U+29A4 ⦤ ANGLE WITH UNDERBAR and U+29A5 ⦥ REVERSED ANGLE WITH UNDERBAR are missing because I'm not sure what variant name to use.


laurmaedje:
1. Seems okay to me.
 
2. + 3. Maybe these two problems together indicate it's more a case of `tr`, `br`, etc than for `rev` and `inv`? Just a thought.

MDLC01:
> 2 and 3: Maybe these two problems together indicate it's more a case of `tr`, `br`, etc than for `rev` and `inv`? Just a thought.

I agree for 2., but it should be noted that this is a breaking change for already existing `angle.rev`, `angle.arc.rev`, `angle.right.rev`, and `angle.spheric.rev`. Moreover, this does not solve 3., as both U+29A8 ⦨ and U+29AC ⦬ have an arrow pointing top-right.

laurmaedje:
Hmm. The breaking change isn't that concerning to me, but unfortunate that 3 is not solved.

MDLC01:
I guess the main issue is that differentiate U+29A8 ⦨ and U+29AC ⦬ is a detail that is hard to describe concisely. Maybe `cw` and `ccw` could be used to indicate the direction of the arrow relative to the other line? But then it does not make sense to use the arrow as the point of reference if we use `tr`, `br`, `tl` and `bl`.

MDLC01:
Yeah, naming concisely things is hard

MDLC01:
Should I also remove `angle.inv` and `angle.oblique.inv` due to the first concern expressed in 2. ("other `inv` variants (such as `quest.inv`) turn the symbol upside-down, whereas here we mirror the symbol around the horizontal axis")?

MDLC01:
Done!

laurmaedje:
Thanks!

PR #4831, opened 2024-08-25
Add miscellaneous symbols from Miscellaneous Mathematical Symbols-B

MDLC01:
This PR adds multiple unrelated symbols from the "Miscellaneous Mathematical Symbols-B" Unicode block.

As far as I can tell, `arrow.r.colon` does not have equivalent Unicode codepoints for other arrow directions, even outside Miscellaneous Mathematical Symbols-B. Maybe it should be named `colon.arrow` instead, similarly to `colon.eq`.

I added variants to `infinity`, but not to `oo`. I can add them to `oo` as well if you want.

### DIFF
--- a/crates/typst/src/symbols/sym.rs
+++ b/crates/typst/src/symbols/sym.rs
@@ -444,7 +444,7 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     // Function and category theory.
     compose: '∘',
     convolve: '∗',
-    multimap: '⊸',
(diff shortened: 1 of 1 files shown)


laurmaedje:
I'm unsure about ⧴ since it seems very special purpose and out of the normal order. Maybe we should leave it out for now since you're also unsure whether it belongs under colon.

MDLC01:
You are right. I've removed it from this PR. It can always be added later when needed.

laurmaedje:
Thanks!

PR #4834, opened 2024-08-25
Add new variants to `parallel` symbol

MDLC01:
This PR adds new variants to the `parallel` symbol.

I'm not sure `bar` and `tilde` are the right variant names for U+2AF2 ⫲ PARALLEL WITH HORIZONTAL STROKE and U+2AF3 ⫳ PARALLEL WITH TILDE OPERATOR. The `tilde` variant name is inconsistent with `slanted.eq.tilde`, where the tilde is above the symbol instead of striking through it. The `bar` variant name is inconsistent with the `emptyset.bar` symbol added in #4826, where the bar is added above the symbol. I am unaware of any similar precedent.

Additionally, maybe `parallel.eq` should be named `eq.parallel` instead, and similarly for other `*.eq.*` and `*.equiv.*` variants.

### DIFF
--- a/crates/typst/src/symbols/sym.rs
+++ b/crates/typst/src/symbols/sym.rs
@@ -455,7 +455,7 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     // Geometry.
     parallel: [
         '∥',
-        sruck: '⫲',
+        struck: '⫲',
         circle: '⦷',
         eq: '⋕',
         equiv: '⩨',
--- a/crates/typst/src/symbols/sym.rs
+++ b/crates/typst/src/symbols/sym.rs
@@ -453,7 +453,18 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     wreath: '≀',
 
     // Geometry.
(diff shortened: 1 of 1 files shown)


laurmaedje:
I think `bar` is maybe not ideal since it's typically vertical. I agree that `eq.parallel` would work as well, but maybe it makes more sense to keep it with the more specific symbol since both are equally large here.

laurmaedje:
`struck` would be new terminology, right? typically, it's `not` I guess ... but probably not what it means here.

MDLC01:
I don't think there is a precedent for a variant that adds a horizontal line like that. In this case, `parallel.not` is the name of another symbol: U+2226 ∦ NOT PARALLEL TO.

laurmaedje:
I think `struck` is fine

MDLC01:
I just checked, and LaTeX uses `\nhpar`, likely for "Not Horizontal PARallel` (and ∦ is `\nparallel`). I think `struck` is better though.

laurmaedje:
`sruck` -> `struck`

laurmaedje:
Thanks!

PR #4836, opened 2024-08-25
Don't number empty regions of broken block equations

### LINKED ISSUE #4829: Duplicate numbering in breakable equations sometimes results in an additional number on the wrong page
### Description

This seems to be an issue introduced with #4226, as it only comes up when using breakable equations. See the minimal example below.

```
#set page(paper: "a6", flipped: true, margin: 1in)
#set text(size: 30pt)

#lorem(10)

#math.equation(block: true, numbering: "(1)", $
  x = y \
  x = y \
$)
```

Screenshot from 2024-08-25 19-15-01

I should also note that after the first `(1)` is a page break of some sort. That is, on bigger documents I still end up with an almost empty page with just the number at the top, followed by the equation on the following page.

This was compiled with the latest commit on master.

```
$ typst --version
typst 0.11.0 (88325d7d)
```

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

EpicEricEE:
Fixes #4829 

Previously, all regions of a broken block equation would get a number, even if no content was added to it.
Now, empty regions are skipped, but a fully empty equation can still be numbered.


laurmaedje:
The contract of `BlockElem::multi_layouter` is that the correspondence between frames it returns and regions is 1 to 1. It looks to me as if this violates this contract by skipping emitting frames for some regions. It might still work due to implementation details of flow layout.

laurmaedje:
Could you move the changed condition into a let binding to keep the indent smaller and the diff clearer?

laurmaedje:
Thanks!

Issue #4843, opened 2024-08-26
Add page numbering support for Devanagari digits.


starryprabin:
### Description

There are many South Asian users (such as me) who can benefit with the Devanagari numbering systems.

![devanagari-digits](https://github.com/user-attachments/assets/9d3c4302-f590-44b2-be33-8c12e577baad)

### Use Case

Documents created in Nepali, Hindi, Marathi and other languages which use Devanagari writing system will benefit from it.

FlorentCLMichel:
In the meantime, this workaround seems to give sensible results. (Definitely not claiming this is the best solution - I'm sure people more familiar with Typst than I am can come up with a better one.)

```
#let convert-number(x) = x.replace(regex("[0-9]"), 
  m => {
  let t = m.at("text");
  ((t == "0", "\u{0966}"),
   (t == "1", "\u{0967}"),
   (t == "2", "\u{0968}"),
   (t == "3", "\u{0969}"),
   (t == "4", "\u{096a}"),
   (t == "5", "\u{096b}"),
   (t == "6", "\u{096c}"),
   (t == "7", "\u{096d}"),
   (t == "8", "\u{096e}"),
   (t == "9", "\u{096f}")).find(x => x.at(0)).at(1)
  })
#set page(footer: context align(center, convert-number(counter(page).display("1"))))
```

starryprabin:
Thanks for the great workaround—it worked well for the page numbers.

There are other sections of the report where the numbering needs to be updated, like the table of contents and headings. I believe using `#set heading(numbering: "१.")` should also apply Devanagari numerals to the headings. 

Similarly, setting the heading with `#set heading(numbering: "१.क")` should use Devanagari letters for the second-level headings. Here, १  is the Devanagari numeral for "1," and `क`  is the Devanagari letter for `A` This pattern should continue with `१.ख`, `१.ग`, etc. But that might be a separate discussion.

FlorentCLMichel:
You're most welcome! Maybe the code below could be a partial solution for the page numbers and headers. (You will need to add more rows to the function `convert-letters-devanagari` if some sections have more than three subsections; in the code below I set the next symbols to `?` as I don't know how Devanagari letters should be ordered. This code will also likely fail if there are more than two levels of headings.) This is most likely not the best solution, though - and only a workaround. 

```
#let convert-numbers-devanagari(x) = x.replace(regex("[0-9]"), 
  m => {
  let t = m.at("text");
  ((t == "0", "\u{0966}"),
[...]

PR #4850, opened 2024-08-27
Add Devanagari numbering

FlorentCLMichel:
This small PR adds support for Devaganari numbering, aiming to resolve https://github.com/typst/typst/issues/4843 .

I'm not sure what the best name for the new `NumberingKind` is. I tentatively went for `DevanagariNumber`, with the idea that another numbering could be added later with Devanagari letters.


crates/typst/src/model/numbering.rs
```diff
@@ -297,6 +297,10 @@ pub enum NumberingKind {
     EasternArabic,
     /// The variant of Eastern Arabic numerals used in Persian and Urdu.
     EasternArabicPersian,
+    /// Devanagari numerals.
+    // TODO: Is it the best name? (I use the suffic ‘Number’ in case we want to also support
+    // devanagari letters)
```
laurmaedje, line 302: Name seems fine. It's also not that important. What matters is the `from_char` and that makes sense.

laurmaedje (commented):
Could you add a test in `tests/suite/model/numbering.typ`? Please use assertion-based tests and not image-based ones. The existing image-based tests are old and not yet migrated.

laurmaedje:
Thanks!

Issue #2885, opened 2023-12-07
Allow functions like `hide` to be called directly in math-mode


NicoWeio:
> [!NOTE]
> *Discussed in https://github.com/typst/typst/discussions/2272*
> I converted this feature request into an issue, as suggested in the discussion.

Currently, one has to write `phi^#hide[$*$]`, which is significantly more cumbersome to write than `phi^hide(x)` (if one uses it a lot) and made me doubt that the layout still works as expected (but it seems to). (Note: Currently, the latter expression would give `unknown variable: hide`.)

I assume this applies to a few more functions, but `hide` is the one I just thought of.

Is there any disadvantage in making such functions “math-native”?

laurmaedje:
There are some duplicate definitions between math and non-math (e.g. `underline` and `rect`), so I conversatively kept things separate (since it's easier to later add non-math to math rather than to remove it).

tingerrr:
If it is decided that we keep some things out of math, then we should work on improving the diagnostics. It took me a few minutes to figure this one out:
```typst
$text(rgb("#..."), ...)$
```

It noted `rgb` is unknown and required `#rgb` instead, because it was not in code mode yet, but this is something I assume can be quickly glanced over in more complex equations that need some non-math functions interspersed.

We already have a few special cases like `none`/`auto`/`true`/`false` being unknown identifiers in math modes will trigger a hint.

MrQubo:
There's a workaround:
```typst
#let hide = it => hide(it)
$f(x, hide(y)) = z$
```

I've copy-pasted it from the linked discussion so more people can find it.

laurmaedje:
Nowadays, it's possible to write
```typ
$ std.hide(x + y) $
```

Having all std definitions available in math by default is problematic because there are some duplicate definitions (like `rect`) and we couldn't add new duplicate ones without it being a breaking change.

PR #4858, opened 2024-08-29
Improve unknown variable diagnostics

tingerrr:
- Add examples to unknown variable hints for possible subtraction.
- Add examples to the literal in math hint.
- Add hint for variables not available in math, but in the global scope.

Related to #2885.

### DIFF
--- a/crates/typst/src/foundations/scope.rs
+++ b/crates/typst/src/foundations/scope.rs
@@ -73,7 +73,12 @@ impl<'a> Scopes<'a> {
                     None => None,
                 })
             })
-            .ok_or_else(|| unknown_variable(var))
+            .ok_or_else(|| {
+                unknown_variable_math(
+                    var,
+                    self.base.is_some_and(|base| base.global.scope().get(var).is_some()),
+                )
+            })
     }
 
     /// Try to access a variable mutably.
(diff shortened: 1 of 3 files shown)


crates/typst/src/foundations/scope.rs
```diff
@@ -109,13 +114,27 @@ fn cannot_mutate_constant(var: &str) -> HintedString {
...
     if matches!(var, "none" | "auto" | "false" | "true") {
         res.hint(eco_format!(
-            "if you meant to use a literal, try adding a hash before it"
+            "if you meant to use a literal, try adding a hash before it: `#{var}`",
         ));
-    } else if var.contains('-') {
+    } else if in_global {
         res.hint(eco_format!(
-            "if you meant to use subtraction, try adding spaces around the minus sign",
+            "`{var}` is not available in math, add a hash before it: `#{var}`",
```
laurmaedje, line 137:
```suggestion
            "`{var}` is not available in directly math, try adding a hash before it: `#{var}`",
```
- I think "directly" makes it a bit clearer that it _is_ available, just with hash.
- The "add" -> "try adding" is maybe unnecessary, but I feel like with all hints that make suggestions, it might be good to be cautious with the wording

max397574: I'm not a native English speaker so not sure about this
but I think it should be either "not directly available" or "not available in math directly" or "not available directly in math"

laurmaedje: yeah typo. I wanted to say "not available directly in math"

laurmaedje:
Thank you!

PR #4863, opened 2024-08-30
Update translation infrastructure

MDLC01:
Supersedes #4701.

Based on [the discussion on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1277669464078286870), I have updated the translation infrastructure to group all translations at the same place, including smart quotes and figure caption separators.

This new infrastructure is similar to [the prototype I sent on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1277757598191456277), happily mentioning it was less than 50 lines long. What I ended up actually implementing is longer than that, in order to produce the cascade-like effect of translations (first, try to find  translation for the language + region, then for the language without the region, and finally fall back to English). I'm not fully satisfied with this final solution (especially the multiple calls `unwrap`), but this is the best I could do without adding even more complexity.

A difference from #4701 is that I changed the keys for smart quotes from `opening`/`closing` to `left`/`right`. This makes the code for `SmartQuotes::get` cleaner.

As in #4701, I had to slightly change Spanish smart quotes. Quoting the original comment for #4701:
> I generated the new translations automatically, so there should not be any error. The only noticeable change is for Spanish: smartquotes used to be non-default for es-ES, and es with specifically no region, but the default ones for es with any non-ES region. I chose to override default quotes for es-ES only, instead of es (which would have affected all regions instead of only es with no region) and changed a test accordingly. I'm not sure whether this was the best choice. Maybe es should be modified instead of es-ES, and special cases should be added for e.g., es-ES? Cc. @Luis-Licea, as you added the quotes for Spanish in https://github.com/typst/typst/pull/941.

Another change I made, which was not present in #4701, is the change from `ua` to `uk` for Ukrainian,[^1] as `uk` is the correct language code (`UA` is the country code for Ukraine). This is a bug fix and should be committed anyway in case this PR is rejected. It is a breaking change.

[^1]: I verified this was indeed the intended language: #578.


laurmaedje:
The cascading complexity feels unfortunate indeed. Aside from the unwraps, also that it is spread across the usages instead of being somehow contained with the `LocalName` impls getting much more complex.

MDLC01:
After thinking about it, I feel like the benefits do not outweigh the drawbacks with this PR. The solution is not as satisfying as I would have liked it to be.
I will give you the final word, but I'm no longer convinced (if I ever was) this is worth merging.

laurmaedje:
Even though I feel kinda bad since this is already your second translation PR, I do agree with you.

PR #4864, opened 2024-08-30
Add `at` method to `arguments`

MDLC01:
This PR adds an `at` method to `arguments`, which corresponds to `arguments.pos(..).at` or `arguments.named(..).at` depending on the type of the parameter. This is just a convenience method, it doesn't allow anything new, but makes the `arguments` type more useful by itself.

### DIFF
--- a/crates/typst/src/foundations/args.rs
+++ b/crates/typst/src/foundations/args.rs
@@ -309,7 +309,7 @@ impl Args {
     ) -> StrResult<Value> {
         match &key {
             ArgumentKey::Index(index) => self.to_pos().at(*index, default).ok(),
-            ArgumentKey::Name(name) => self.to_named().at(name, default).ok(),
+            ArgumentKey::Name(name) => self.to_named().at(name.clone(), default).ok(),
         }
         .ok_or_else(|| missing_key_no_default(key))
     }
--- a/crates/typst/src/foundations/args.rs
+++ b/crates/typst/src/foundations/args.rs
@@ -1,8 +1,9 @@
 use std::fmt::{self, Debug, Formatter};
 
 use ecow::{eco_format, eco_vec, EcoString, EcoVec};
+use typst_macros::cast;
 
-use crate::diag::{bail, error, At, SourceDiagnostic, SourceResult};
+use crate::diag::{bail, error, At, SourceDiagnostic, SourceResult, StrResult};
 use crate::foundations::{
     func, repr, scope, ty, Array, Dict, FromValue, IntoValue, Repr, Str, Value,
 };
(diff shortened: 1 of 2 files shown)


crates/typst/src/foundations/args.rs
```diff
@@ -1,8 +1,9 @@
 use std::fmt::{self, Debug, Formatter};
 
 use ecow::{eco_format, eco_vec, EcoString, EcoVec};
+use typst_macros::cast;
```
laurmaedje, line 4: please move this into the `crate::foundations` import

crates/typst/src/foundations/args.rs
```diff
@@ -256,6 +257,17 @@ impl Args {
     }
 }
 
+pub enum ArgumentKey {
```
laurmaedje, line 260: please add a brief doc comment.

crates/typst/src/foundations/args.rs
```diff
@@ -279,6 +291,29 @@ impl Args {
...
+        &self,
+        /// The index or name of the argument to get.
+        key: ArgumentKey,
+        /// A default value to return if the key is invalid.
+        #[named]
+        default: Option<Value>,
+    ) -> StrResult<Value> {
+        match &key {
+            ArgumentKey::Index(index) => self.to_pos().at(*index, default).ok(),
+            ArgumentKey::Name(name) => self.to_named().at(name.clone(), default).ok(),
```
laurmaedje, line 312: I think if we already have this method, we should make it more efficient than calling `to_pos` and `to_named` internally, since those are quite costly.

MDLC01: Done!

crates/typst/src/foundations/args.rs
```diff
@@ -256,6 +256,56 @@ impl Args {
...
+                self.items
+                    .iter()
+                    .filter_map(|item| {
+                        item.name.as_ref().map(|k| (k, item.value.v.clone()))
+                    })
+                    .rfind(|(k, _)| *k == name)
+                    .map(|(_, v)| v)
+            }
+        }
+    }
```
laurmaedje, line 306:
```suggestion
    fn get(&self, key: &ArgumentKey) -> Option<&Value> {
        let item = match key {
            &ArgumentKey::Index(index) => {
                let mut iter = self.items.iter().filter(|item| item.name.is_none());
                if index < 0 {
                    let index = (-(index + 1)).try_into().ok()?;
                    iter.nth_back(index)
                } else {
                    let index = index.try_into().ok()?;
                    iter.nth(index)
                }
            }

            // Accept the last argument with the right name.
            ArgumentKey::Name(name) => {
                self.items.iter().rfind(|item| item.name.as_ref() == Some(name))
            }
        };

        item.map(|item| &item.value.v)
    }
```
a bit less code duplication and potential cloning. requires `get(&key).cloned()` below

MDLC01: This is indeed better.

laurmaedje:
Thanks!

PR #4869, opened 2024-08-31
Parenthesized imports

### LINKED ISSUE #3065: Allow importing with array syntax (like in Python)
### Description

Imagine writing this (ugly and unreadable!):
```typ
#import "/utils.typ": thing1, thing2, something-very-long, stuff1, stuff2, stuff3, and-maybe-something-else, ok-this-is-the-last-one
```

Instead of this (you can't be more readable than this):
```typ
#import "/utils.typ": (
  thing1,
  thing2,
  something-very-long,
  stuff1,
  stuff2,
  stuff3,
  and-maybe-something-else,
  ok-this-is-the-last-one,
)
```

I get that for `= long heading` it isn't great to be able to place heading's content on multiple lines (becomes less readable, but would be nice for very long heading), but for import statement it has become obvious that sometimes import lines are simply too long. Look even at Rust's imports (`use`) — maybe something like that also would work (but it's less understandable, than a simple array-syntax imports).

### Use Case

Remove clutter on import lines and make huge imports more readable and maintainable.

Dherse:
Fixes #3065

Allows the following syntax:
```typ
#import "module.typ": (a, b, c)

or

#import "module.typ": (
  a, b, c
)
```

Added relevant tests, change is tiny.


tests/suite/scripting/import.typ
```diff
@@ -55,6 +55,34 @@
 #test(name, "Klaus")
 #test(othername, "Klaus")
 
+--- import-items-parenthesized ---
```
PgBiel, line 58: Please add one or two negative tests where the right paren is missing (maybe one in a single line and one multiline).

PgBiel (commented):
Nice job, thank you!

tests/suite/scripting/import.typ
Dherse: Sure will do after working ;)

crates/typst-syntax/src/parser.rs
```diff
@@ -989,9 +989,22 @@ fn module_import(p: &mut Parser) {
         // imported at the same time.
         p.expect(SyntaxKind::Ident);
     }
-    if p.eat_if(SyntaxKind::Colon) && !p.eat_if(SyntaxKind::Star) {
-        import_items(p);
+
+    if p.eat_if(SyntaxKind::Colon) {
+        if p.current() == SyntaxKind::LeftParen {
```
laurmaedje, line 994:
```suggestion
        if p.at(yntaxKind::LeftParen) {
```

crates/typst-syntax/src/parser.rs
```diff
@@ -989,9 +989,22 @@ fn module_import(p: &mut Parser) {
...
+            let m1 = p.marker();
+            p.enter_newline_mode(NewlineMode::Continue);
+            p.expect(SyntaxKind::LeftParen);
```
laurmaedje, line 997:
```suggestion
            p.assert(SyntaxKind::LeftParen);
```

tests/suite/scripting/import.typ
Dherse: Done, please let me know if more are needed

PR #4870, opened 2024-08-31
Added example to `box.clip` & `block.clip`

### LINKED ISSUE #2011: Add example and clarification for `clip` parameter of `box`/`block`
https://typst.app/docs/reference/layout/box/#parameters-clip:
https://typst.app/docs/reference/layout/block/#parameters-clip:

> `clip` [`boolean`](https://typst.app/docs/reference/types/boolean) _Settable_
Whether to clip the content inside the box/block.
Default: `false`

I have 2 issues:
- there is no example (with a result) that shows how exactly can/should you use this parameter,
- therefore (perhaps) reader can't even understand what does it mean: is it to attach something to something else (?) or is it to cut something off in some way? (["to clip" in Oxford Dictionary](https://www.oxfordlearnersdictionaries.com/definition/english/clip_2)).

Maybe these issues exist because of `clip` not working (as intended/at all): #737, #1621. Regardless, the clarification + example should be present in the documentation of `box` and `block` to remove any confusion when reading the documentation. At the very least, the example should be added.

Dherse:
Closes #2011

### DIFF
--- a/crates/typst/src/layout/container.rs
+++ b/crates/typst/src/layout/container.rs
@@ -106,6 +106,18 @@ pub struct BoxElem {
     pub outset: Sides<Option<Rel<Length>>>,
 
     /// Whether to clip the content inside the box.
+    ///
+    /// Clipping is useful when the box's content is larger than the box itself,
+    /// as any content that exceeds the box's bounds will be hidden.
+    ///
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

PR #4872, opened 2024-09-01
Added `purge` CLI command

### LINKED ISSUE #4441: Allow users to clean package cache & hint for broken package cache
### Description

Recent days someone told me that he met an error message like this:

```console
$ typst compile test.typ
error: file not found (searched at ~/.cache/typst/packages/preview/cetz/0.2.2/typst.toml)
  ┌─ test.typ:1:8
  │
1 │ #import "@preview/cetz:0.2.2"
```

It seems the cached package is corruptted. We need to remove `~/.cache/typst/packages/preview/cetz/0.2.2` directory and redownload the package to make it work.

As discussed in discord[^1], I proposed to provide a CLI option to clean cache and two hints for such error:
- Hint 1. the package `@preview/{name}:{version}` seems broken, try `typst compile --package-cache-clean=@preview/{name}:{version}`
- Hint 2. the package cache seems broken, try `typst compile --package-cache-clean`

Users are suggested to try Hint 1 first, and if it does not fix the problem, try Hint 2.

[^1]: https://discord.com/channels/1054443721975922748/1088371867913572452/1254734164842778646

### Use Case

Users will know how to deal with this error (clean the cache and redownload the package).

Dherse:
Adds a `purge` command to clean the package cache, can optionally specify the `--local` flag to also clear out the local package storage.

Closes #4441

### DIFF
--- a/crates/typst-kit/src/package.rs
+++ b/crates/typst-kit/src/package.rs
@@ -178,4 +178,28 @@ impl PackageStorage {
             PackageError::MalformedArchive(Some(eco_format!("{err}")))
         })
     }
+
+    /// Purge the package cache.
+    pub fn purge_cache(&self) -> PackageResult<()> {
+        if let Some(cache_dir) = &self.package_cache_path {
+            if cache_dir.exists() {
+                fs::remove_dir_all(cache_dir)
+                    .map_err(|err| PackageError::Other(Some(eco_format!("{err}"))))?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Purge the local package storage.
+    pub fn purge_local(&self) -> PackageResult<()> {
+        if let Some(packages_dir) = &self.package_path {
+            if packages_dir.exists() {
+                fs::remove_dir_all(packages_dir)
+                    .map_err(|err| PackageError::Other(Some(eco_format!("{err}"))))?;
+            }
+        }
+
+        Ok(())
+    }
 }
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -62,6 +62,9 @@ pub enum Command {
     /// Self update the Typst CLI
     #[cfg_attr(not(feature = "self-update"), doc = " (disabled)")]
     Update(UpdateCommand),
+
+    /// Purge the package cache
+    Purge(PurgeCommand),
 }
 
 /// Compiles an input file into a supported output format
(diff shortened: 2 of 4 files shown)


istudyatuni:
I think it's better to place this under "packages" (or "pkg") subcommand, like `typst pkg purge-cache`

Dherse:
But there currently isn't a `package` command, so I would need to create it then? Is it really warranted?

mohe2015:
I think it would make sense to have a subcommand. Otherwise you may think it purges the current compilation result or something like that. Or maybe `typst purge-package-cache`?

istudyatuni:
> Is it really warranted?

I think yes, because I'm not sure this will be the only command related to packages

crates/typst-kit/src/package.rs
```diff
@@ -178,4 +178,28 @@ impl PackageStorage {
             PackageError::MalformedArchive(Some(eco_format!("{err}")))
         })
     }
+
+    /// Purge the package cache.
+    pub fn purge_cache(&self) -> PackageResult<()> {
+        if let Some(cache_dir) = &self.package_cache_path {
+            if cache_dir.exists() {
```
Coekjan, line 185: I think the _granularity_ needs to be discussed. I mean users might want to clear a specific package cache, instead of the whole cache directory.

BTW, because `package_cache_path` can be set via CLI args or environments, this recursive removing could be dangerous. If a malware set the `TYPST_PACKAGE_CACHE_PATH` environment variable to your `$HOME`, and you don't know it, the `purge` command can **silently** destroy your home...

Dherse:
> > Is it really warranted?
> 
> I think yes, because I'm not sure this will be the only command related to packages

Yes you're right, it makes sense under that lens!

crates/typst-kit/src/package.rs
Dherse: Well I mean, that's always going to be an issue, no? Even if I manually delete just the files, it'll still be dangerous

Coekjan:
At the same time, #4441 proposed not only a CLI to clear cache, but also some hints. So I don't think this PR can really close the issue.

crates/typst-kit/src/package.rs
Coekjan: At least, I think, let users confirm the being-removed directory.

Dherse: Yes that's a good idea, do we already have something similar?

mohe2015: What about *in addition* rely on the known structure of this directory. So remove the subdirectory `packages/preview/` etc to ensure no data is accidentially deleted (with some certainty)

Coekjan: > do we already have something similar?

I don't know if we have such infrastructure.

laurmaedje:
I think we need to have a broader plan for CLI commands related to package management rather than making an ad-hoc decision here.

Dherse:
Following @laurmaedje's comment, I think it's best to close this for the moment.

PR #4873, opened 2024-09-01
Better hint for non-zero em length

### LINKED ISSUE #4304: More useful hints for converting relative lengths to absolute units
### Description

I have `let sum = a + 0.25em.mm() + b` and I get:

```
1. cannot convert a length with non-zero em units (`0.25em`) to mm
   
   Hint: use `length.abs.mm()` instead to ignore its em component
```

But what I **actually** had to do is `let sum = a + 0.25em.to-absolute().mm() + b`. So it should add 2nd hint to the top mentioning `.to-absolute()` or replace the current hint altogether.

### Use Case

The more useful the hints are, the easier it is to prototype or just write anything in general. Can save a lot of time.

Dherse:
Closes #4304


crates/typst/src/layout/length.rs
```diff
@@ -81,12 +81,15 @@ impl Length {
...
+            span,
             "cannot convert a length with non-zero em units (`{}`) to {unit}",
-            self.repr()
-        ))
-        .hint(eco_format!("use `length.abs.{unit}()` instead to ignore its em component"))
-        .at(span)
+            self.repr();
+            hint: "use `length.abs.{unit}()` instead to ignore its em component";
+            hint: "or use `length.to-absolute()` to resolve its em component \
+                   (requires a context)"
```
laurmaedje, line 91:
```suggestion
                   (requires context)"
```
maybe we should also flip the order of the hints? ignoring is generally not as advisable.

laurmaedje:
Probably missed this:

> maybe we should also flip the order of the hints? ignoring is generally not as advisable.

Dherse:
@laurmaedje yes sorry, went over it a bit too quickly, done ;)

laurmaedje:
Thanks!

PR #4883, opened 2024-09-03
Add `copyleft`, `trademark`, and `registered` symbols

### LINKED ISSUE #4882: Add trademark symbols
### Description

It would be very nice to have several trademark related symbols available via the builtin general symbols.
Symbols I am missing:
- Trademark (™)
- Registered (®)
- Service Mark (℠)

Since even the dollar symbol is included, I though it might make sense to have these as well.

EDIT: not sure if this is more a bug than a "new" feature

### Use Case

When mentioning trademark protected brands, products or services in presentations or papers it would convenient to add these symbols through the globally predefined mapping.

This means it can be avoided to have to search, copy and paste the Unicode glyph.

MDLC01:
I'm back with a symbol PR!

This one adds `copyleft`, `trademark` and `registered`, and renames `servicemark` to `trademark.service`.

Closes #4882.


crates/typst/src/symbols/sym.rs
```diff
@@ -126,6 +126,8 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     hyph: ['‐', minus: '\u{2D}', nobreak: '\u{2011}', point: '‧', soft: '\u{ad}'],
     percent: '%',
     copyright: ['©', sound: '℗'],
+    copyleft: '🄯',
+    trademark: ['™', service: '℠', registered: '®'],
```
laurmaedje, line 130: trademark might be better suited next to servicemark maybe?

laurmaedje: copyright and copyleft as well probably

MDLC01: I had not realized `servicemark` already existed.

laurmaedje: Actually I hadn't realized that you added `trademark.service`. Maybe it's a better name than `servicemark`...

crates/typst/src/symbols/sym.rs
```diff
@@ -525,7 +524,10 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     crossmark: ['✗', heavy: '✘'],
     floral: ['❦', l: '☙', r: '❧'],
     refmark: '※',
-    servicemark: '℠',
+    copyright: ['©', sound: '℗'],
+    copyleft: '🄯',
+    registered: '®',
```
reknih, line 529: This is a symbol for a registered trademark, so I think it should be `trademark.registered`

MDLC01: That makes sense

PR #4885, opened 2024-09-03
Line::{stretchability, shrinkability}: return +0.0 when empty

### LINKED ISSUE #4877: Test failures with rustc 1.83.0-nightly (94885bc69 2024-09-01)
### Description

Some of the integration tests fail in the latest nightly version of Rust (`rustc 1.83.0-nightly (94885bc69 2024-09-01)`):

```
💨 708 / 2224
' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
bounded paragraph layout is incomplete
💨 710 / 2224
thread '<unnamed>' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
💨 713 / 2224
❌ justify-no-leading-spaces (tests/suite/layout/inline/justify.typ:76) panicked
💨 714 / 2224
❌ justify (tests/suite/layout/inline/justify.typ:2) panicked
:
bounded paragraph layout is incomplete
💨 723 / 2224
❌ issue-2419-justify-hanging-indent (tests/suite/layout/inline/justify.typ:166) panicked
💨 827 / 2224
thread '<unnamed>' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
bounded paragraph layout is incomplete
❌ justify-code-blocks (tests/suite/layout/inline/justify.typ:86) panicked
💨 1376 / 2224
thread '<unnamed>' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
bounded paragraph layout is incomplete
❌ par-hanging-indent (tests/suite/model/par.typ:68) panicked
❌ issue-4278-par-trim-before-equation (tests/suite/model/par.typ:98)
  mismatched rendering
    live      | tests/store/render/issue-4278-par-trim-before-equation.png
    ref       | tests/ref/issue-4278-par-trim-before-equation.png
💨 1847 / 2224
thread '<unnamed>' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
bounded paragraph layout is incomplete
❌ show-text-regex-character-class (tests/suite/styling/show-text.typ:63) panicked
💨 1902 / 2224
thread '<unnamed>' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
bounded paragraph layout is incomplete
❌ show-selector-where (tests/suite/styling/show.typ:197) panicked
💨 2006 / 2224
thread '<unnamed>' panicked at crates/typst/src/layout/inline/linebreak.rs:332:9:
bounded paragraph layout is incomplete
❌ issue-3191-raw-indent-shrink (tests/suite/text/raw.typ:605) panicked
2215 passed, 9 failed, 0 skipped
  pass the --update flag to update the reference images
error: test failed, to rerun pass `--test tests`
```

In contrast, all tests pass with rustc 1.80.1 and 1.81.0-beta.8.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

bluebear94:
Fixes #4877.

rustc 2024-08-22 changed `<f64 as Sum>::sum` to return -0.0 for an empty iterator (rust-lang/rust#129321). When -0.0 is passed as the stretchability or shrinkability parameter of `linebreak::raw_ratio`, the ratio value becomes negative infinity and does not hit the `ratio > 1.0` check. This causes -inf to be returned as the line ratio from `ratio_and_cost`, leading to the `active` variable being incremented too early in `linebreak_optimized_bounded` and the entire line breaking process being skipped.

### DIFF
--- a/crates/typst/src/layout/inline/line.rs
+++ b/crates/typst/src/layout/inline/line.rs
@@ -76,7 +76,7 @@ impl<'a> Line<'a> {
             .iter()
             .filter_map(Item::text)
             .map(|s| s.stretchability())
-            // Don’t use sum because it returns -0.0 for an empty iter
+            // Use fold instead of sum because sum returns -0.0 for an empty iter
             .fold(Abs::zero(), |acc, x| acc + x)
     }
 
@@ -86,7 +86,7 @@ impl<'a> Line<'a> {
             .iter()
             .filter_map(Item::text)
             .map(|s| s.shrinkability())
-            // Don’t use sum because it returns -0.0 for an empty iter
+            // Use fold instead of sum because sum returns -0.0 for an empty iter
             .fold(Abs::zero(), |acc, x| acc + x)
     }
 
--- a/crates/typst/src/layout/inline/line.rs
+++ b/crates/typst/src/layout/inline/line.rs
@@ -76,7 +76,8 @@ impl<'a> Line<'a> {
             .iter()
             .filter_map(Item::text)
             .map(|s| s.stretchability())
-            .sum()
+            // Don’t use sum because it returns -0.0 for an empty iter
+            .fold(Abs::zero(), |acc, x| acc + x)
     }
 
     /// How much the line can shrink.
@@ -85,7 +86,8 @@ impl<'a> Line<'a> {
             .iter()
             .filter_map(Item::text)
             .map(|s| s.shrinkability())
-            .sum()
+            // Don’t use sum because it returns -0.0 for an empty iter
+            .fold(Abs::zero(), |acc, x| acc + x)
     }
 
     /// Whether the line has items with negative width.
--- a/crates/typst/src/layout/inline/line.rs
+++ b/crates/typst/src/layout/inline/line.rs
@@ -72,21 +72,21 @@ impl<'a> Line<'a> {
 
     /// How much the line can stretch.
     pub fn stretchability(&self) -> Abs {
+        // Use fold instead of sum because sum returns -0.0 for an empty iter
         self.items
             .iter()
             .filter_map(Item::text)
             .map(|s| s.stretchability())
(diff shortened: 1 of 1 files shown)


crates/typst/src/layout/inline/line.rs
```diff
@@ -76,7 +76,8 @@ impl<'a> Line<'a> {
             .iter()
             .filter_map(Item::text)
             .map(|s| s.stretchability())
-            .sum()
+            // Don’t use sum because it returns -0.0 for an empty iter
+            .fold(Abs::zero(), |acc, x| acc + x)
```
PgBiel, line 80: I think this comment should go above the `self.items` line. Then you should write "Use fold instead of sum" rather than "Don't use sum".

PgBiel (commented):
Well, this is annoying, thanks for figuring it out.

Though the change occurred in nightly, which makes me ponder whether we should be so quick as to fix this before the next Rust release, as that particular change could be undone. But I think there isn't much harm in doing this.

More generally, though, I wonder if there could be some more explicit way to express that this code (or code depending on it) is sensitive to the sign of zero as part of its invariants.

laurmaedje:
I think we should fix `raw_ratio` instead. It's current behaviour is too subtle. I'd need to think more later to give an opinion on how exactly to it change though.

workingjubilee:
Sorry, didn't mean to trouble you so! But tbh I r+'d that change _precisely because_ sometimes the negative zero matters, and when it does you often want to be able to tell that you summed negative zeroes apart from summing a negative zero and a positive zero.

PgBiel:
Yeah, I admit I wasn't very aware of the context earlier, so my comment wasn't too relevant in retrospective :sweat_smile:

This definitely did surface something we might need to look deeper into, so thank you for that.

laurmaedje:
Superseded by https://github.com/typst/typst/pull/4976. Thanks for finding this problem @bluebear94!

PR #4886, opened 2024-09-03
Make `jump_from_cursor` return a vector of `Position`

Dherse:
Sorry @laurmaedje but as we say in French: "thing promised, thing owed" 😂 

First step in fixing a bug in the webapp where writing to headings jumps you to the contents table: shortly discussed on discord -> https://discord.com/channels/1054443721975922748/1088371919725793360/1280530856615608413

### DIFF
--- a/crates/typst-ide/src/jump.rs
+++ b/crates/typst-ide/src/jump.rs
@@ -113,25 +113,30 @@ pub fn jump_from_cursor(
     document: &Document,
     source: &Source,
     cursor: usize,
-) -> Option<Position> {
+) -> Vec<Position> {
     fn is_text(node: &LinkedNode) -> bool {
         node.get().kind() == SyntaxKind::Text
     }
 
     let root = LinkedNode::new(source.root());
-    let node = root
+    let Some(node) = root
         .leaf_at(cursor, Side::Before)
         .filter(is_text)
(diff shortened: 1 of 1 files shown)


crates/typst-ide/src/jump.rs
```diff
@@ -113,25 +113,30 @@ pub fn jump_from_cursor(
...
-        }
-    }
-
-    None
+    document
+        .pages
+        .iter()
+        .enumerate()
+        .filter_map(|(i, page)| {
+            find_in_frame(&page.frame, span)
```
laurmaedje, line 136: I think if we're already doing this, we probably want to find multiple hits in the same frame as well. So `find_in_frame` could take an `&mut Vec<Position>` as its first argument or something like that (would need to pass the page number down for this).

laurmaedje: Though, hmm, we would probably get a bunch of hits closeby for glyphs coming from the same span. Not sure how to handle that. Text might be chunked up into `FrameItem::Text` somewhat arbitrarily...

Dherse: I'll let you decide because I have no intuition for these things @laurmaedje

laurmaedje: I think one per page is fine because we have no current use case for more.

laurmaedje:
Thanks!

PR #4888, opened 2024-09-03
Pop stylechain head in outline

### LINKED ISSUE #4646: Wrong context gets passed to list of figures
### Description

I want to number my figures according to chapter, e. g. 1.1, 2.1, 3.1, etc.. As I have also an appendix in my document which is numbered alphabetically, the numbering for the figures includes also A.1, A.2, B.1, etc..

Now I tried to solve this in Typst. I followed the approach in #1057:
```typst
#show heading.where(level: 1): it => {
    counter(figure.where(kind: table)).update(0);
    counter(figure.where(kind: image)).update(0);
    counter(figure.where(kind: raw)).update(0);

    [
      #it
    ]
}

#set figure(numbering: it => {
  let test = figure
  let numberingOfHeading = counter(heading).display();
  let topLevelNumber = numberingOfHeading.slice(0, numberingOfHeading.position("."))
  [#topLevelNumber.#it]
})
```
This works fine as long as I don't have an appendix. If I create the appendix as following:
```typst
#counter(heading).update(0)
#set heading(numbering: "A.1", supplement: [Appendix])
#show heading: it => {
  if it.level == 1 and it.numbering != none {
    [#it.supplement #counter(heading).display():]
  } else if it.numbering != none {
    [#counter(heading).display().]
  }

  h(0.3em)
  it.body
  parbreak()
}
#set figure(numbering: it => {
  if heading.supplement == [Appendix] {
    [APPENDIX]
    return;
  }
  
  let numberingOfHeading = counter(heading).display();
  let topLevelNumber = numberingOfHeading.slice(0, numberingOfHeading.position("."))
  [#topLevelNumber.#it]
})

= Literature Research <app:ch:literatureResearch>
#lorem(10)

#figure(
  image("tiger.jpeg", width: 50%),
  caption: "ABC"
)

#lorem(10)

#figure(
  image("tiger.jpeg", width: 50%),
  caption: "ABC"
)

= Expert Interviews <app:ch:expertInterviews>
#lorem(10)

#figure(
  image("tiger.jpeg", width: 50%),
  caption: "ABC"
)

#lorem(10)

#figure(
  image("tiger.jpeg", width: 50%),
  caption: "ABC"
)
```

Unfortunately this looks like this:
image

This looks to me that the wrong context is passed to the show rule - in the document itself it works as expected.
As a workaround I modified the second set rule as following:
```typst
#set figure(numbering: it => {
    let alphabet = ("A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z")
    let numberingOfHeading = counter(heading).display();
    let topLevelNumber = numberingOfHeading.slice(0, numberingOfHeading.position("."))
    let index = alphabet.position((el) => { el == topLevelNumber})

    if index == none {
      let numberingToAlphabet = numbering("A", int(topLevelNumber))
      [#numberingToAlphabet.#it]
    } else {
      [#topLevelNumber.#it]
    }
  })
```
Now it looks as expected:
image

As I think that I may be a bug and wanted to report it therefore. 

### Reproduction URL

https://typst.app/project/roh4szf36uz8idLKORiP0W

### Operating system

Web app

### Typst version

Dherse:
Fixes #4646

When building outline, always the same `Styles` gets added to the `StyleChain` at the head, I just added a piece of code that "pops" it, fixing the bug where the wrong style chain was provided to outline entries.

EDIT: this would still only involve giving the style chain at the place where the outline is defined, not where the heading is defined. There is definitely a discussion to be had of which behavior is preferable

### DIFF
--- a/crates/typst/src/foundations/styles.rs
+++ b/crates/typst/src/foundations/styles.rs
@@ -702,6 +702,11 @@ impl<'a> StyleChain<'a> {
         *self = self.tail.copied().unwrap_or_default();
     }
 
+    /// The tail of the chain.
+    pub fn tail(&self) -> Option<Self> {
+        self.tail.copied()
+    }
+
     /// Determine the shared trunk of a collection of style chains.
     pub fn trunk(iter: impl IntoIterator<Item = Self>) -> Option<Self> {
         // Determine shared style depth and first span.
--- a/crates/typst/src/model/outline.rs
+++ b/crates/typst/src/model/outline.rs
@@ -209,13 +209,15 @@ impl Show for Packed<OutlineElem> {
         let mut ancestors: Vec<&Content> = vec![];
         let elems = engine.introspector.query(&self.target(styles).0);
 
+        // Compute the style chain before the outline `ShowSet` is applied.
+        let pre = styles.tail().unwrap_or_default();
         for elem in &elems {
             let Some(entry) = OutlineEntry::from_outlinable(
                 engine,
                 self.span(),
                 elem.clone(),
                 self.fill(styles),
-                styles,
+                pre,
             )?
             else {
                 continue;


laurmaedje:
I don't understand how this fixes #4646. The issue is very broad and foundational: That contexts that fake a different position in the document via introspection only provide locations, not styles of that position. I'm not sure what to make of the issue and whether providing all the styles would actually be desirable, but that's a separate question.

Dherse:
@laurmaedje it fixes the issue mentioned, I tried looking for other similar cases and could not really find one. But I see your point, I think that in the meantime, this is a decent temporary fix to alleviate one of the bigger symptoms of the issue if you see what I mean.

Dherse:
Actually, I see what you mean actually, it indeed doesn't solve the underlying issue :| It fixes the minimal example (which is the one I had used during testing) but not the full example mmh 👎

Issue #3454, opened 2024-02-18
Whitespace in inline raw blocks affected by justification


abel1502:
### Description

Within a `par(justify: true)`, whitespace characters inside inline raw blocks like `` `foo bar` `` are subjected to the justification. Since raw blocks are supposed to represent monospace data, I expect all characters within it to be rendered with the same width and to not be subjected to any cosmetic alterations (unless explicitly demanded). So, I assume the proper behaviour would be to treat the raw block as a whole immutable unit for the purposes of justification, regardless of the presence of whitespace inside it

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

Leedehai:
I think it's more of a feature request. @laurmaedje thoughts?

```typ
// tests/typ/bugs/demo.typ
#set par(justify: true)
#set raw(lang: "python")
`if else` foo bar foo bar `if else` foo foo bar bar!!
```
<img width="252" src="https://github.com/typst/typst/assets/18319900/e5f4318e-12a8-4feb-bc79-a0d10d67aed4">


The view is that the space within the `if else` span should not be affected by justification.

laurmaedje:
I see how it's not necessary what one would expect. Whether we want to label that a bug or feature request, idk.

laurmaedje:
@elegaanz I think `ParElem::set_shrink(false)` will only affect paragraphs for which it is set uniformly, so inline raw alongside some text wouldn't be affected.

jarkad:
As a workaround, you can put your `raw`s in a `box`:

```typst
#set par(justify: true)

`hello world` #lorem(5)
`hello world` #lorem(10)

#[
  #show raw: box
  `hello world` #lorem(5)
  `hello world` #lorem(10)
]
```

![Image](https://github.com/user-attachments/assets/45d0405c-3869-4b94-a811-37ab136c1cf1)

gasche:
I encountered a similar issue in a document where the whitespace changes came not from the `justify` settings, but from a `text(spacing: 80%)` setting applied globally by the template. I fixed this with

```typst
  show raw: set text(font: ttFont, spacing: 100%)
```

where `ttFont` is a monospace font (or an array of monospace fonts).

PR #3796, opened 2024-03-26
Never shrink lines in raw blocks

elegaanz:
Shrinking a line in a raw block can visually break the indentation.

This patch adds an internal property on paragraphs that can be used to disable line shrinking. This property is set to `true` only for raw blocks.

Fixes #3191

PR #4889, opened 2024-09-03
Disable justification in `raw(block: true, ...)`

Dherse:
Fixes an issue that has been flagged in codly: https://github.com/Dherse/codly/issues/26

Currently only for block raw blocks, but it's trivial to make the change more generic.


Dherse:
Actually, before just updating the test, I'll wait for feedback from @laurmaedje cause the test failing doesn't make much sense to me: I see that in the new version it actually includes a linebreak -- which to me seems correct -- but I could see the argument that the old one was better.  I don't really understand why justification allows it to go past the block size...

So @laurmaedje I trust your judgement on this.

image

laurmaedje:
Minimal reproduction:

```rust
#set par(justify: true) // toggle this
#rect(width: 2.4cm)[
  #"    AAAAAAAA"
]
```

Seems like justification affects how leading spaces are handled.

Dherse:
So this is actually a bug in Typst? 🤔

laurmaedje:
I'd say yes.

laurmaedje:
I opened https://github.com/typst/typst/issues/5348 for the bug. 

The new image is fine in principle, but it breaks the intent of the test. The test comes from https://github.com/typst/typst/pull/3796, which disables shrinking (justification where spaces go below their normal size instead of above) in raw blocks. But that's kinda pointless when justification is disabled anyway.

@elegaanz Do you think we should keep the non-shrink handling? I think it would only come into play when someone write `show raw.where(block: true): set par(justify: true)`, which is a very odd thing to write. I can't remember the motivation for disabling shrinking specifically instead of just disabling justification for raw blocks.

laurmaedje:
I would lean towards removing the shrink thing then. Raw blocks do not really seem like the right solution to pseudo-code to me.

laurmaedje:
Notably, this doesn't fix https://github.com/typst/typst/issues/3454 because that's about _inline_ raw blocks. I edited the original PR description to unlink the issue.

Issue #2608, opened 2023-11-07
syntax for marking project root


Roger-luo:
### Description

I'd like to have a syntax sugar for all path strings to access the project root path in the import, e.g

```
#import "@project/templates/lib.typ"

image("@project/figure/image.png")
```

and `@project` will be replaced by the root path specified to `typst` compiler via `--root`, ideally it should be set automatically by some file structure, but typst currently doesn't have a standard way of specifying a project, maybe it can also search the file path `typst.toml` first.

### Use Case

Currently, this is quite inconvenient especially when the project has many files organized by directories. You have to write the above as

```
#import "../../../templates/lib.typ"

image("../../../figure/image.png")
```

and moving around the script will break the above path. The idea is quite similar to rust's `crate::`  and `super::` in the module loading syntax

On the other hand, I noticed there is #2148 but more for importing package internals, which is not the same as this proposal.

Roger-luo:
In the case that using `@<name>` may break the current package loading behavior, I think special case `@project` or `@root` is probably fine as I don't see these two words would make sense in the future.

laurmaedje:
There is existing syntax for this: `/`. 

`#import "/templates/lib.typ"` will load the file `templates/lib.typ` relative to the `--root`.

Roger-luo:
oh thanks! I didn't know that, maybe this should be mentioned in the doc somewhere? I did a random search didn't find any.

laurmaedje:
yeah, I'm not sure but maybe it's in fact not in the docs.

PR #4892, opened 2024-09-04
Clarify the path argument of the image element

kravchenkoloznia:
Add clarification that by default the path argument is relative to the input file location, but with the `/` symbol as the first character it become relative to the `--root` path.

The feature is mentioned in the https://github.com/typst/typst/issues/2608#issuecomment-1799021131 comment but absent on the documentation page.

### DIFF
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -56,7 +56,7 @@ use crate::World;
 pub struct ImageElem {
     /// Path to an image file
     ///
-    /// For detailed description see [Paths]($syntax/#paths).
+    /// For more details, see the [Paths section]($syntax/#paths).
     #[required]
     #[parse(
         let Spanned { v: path, span } =
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -54,7 +54,9 @@ use crate::World;
 /// [gh-svg]: https://github.com/typst/typst/issues?q=is%3Aopen+is%3Aissue+label%3Asvg
 #[elem(scope, Show, LocalName, Figurable)]
 pub struct ImageElem {
-    /// Path to an image file, relative to the input file by default; if starting with `/`, it's relative to the `--root` path.
+    /// Path to an image file
+    ///
+    /// For detailed description see [Paths]($syntax/#paths).
     #[required]
     #[parse(
         let Spanned { v: path, span } =
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -54,7 +54,7 @@ use crate::World;
 /// [gh-svg]: https://github.com/typst/typst/issues?q=is%3Aopen+is%3Aissue+label%3Asvg
 #[elem(scope, Show, LocalName, Figurable)]
 pub struct ImageElem {
-    /// Path to an image file.
+    /// Path to an image file, relative to the input file by default; if starting with `/`, it's relative to the `--root` path.
     #[required]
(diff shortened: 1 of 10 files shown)


crates/typst/src/visualize/image/mod.rs
```diff
@@ -54,7 +54,7 @@ use crate::World;
 /// [gh-svg]: https://github.com/typst/typst/issues?q=is%3Aopen+is%3Aissue+label%3Asvg
 #[elem(scope, Show, LocalName, Figurable)]
 pub struct ImageElem {
-    /// Path to an image file.
+    /// Path to an image file, relative to the input file by default; if starting with `/`, it's relative to the `--root` path.
```
Coekjan, line 57: For non-CLI users, they would not know what `--root` is. :confused:

kravchenkoloznia: According to the [Usage](https://github.com/typst/typst/blob/main/README.md#usage) chapter of the readme, the tool is used only from CLI.

laurmaedje: It talks about the Typst web app in the same section.

kravchenkoloznia: @laurmaedje true, I missed that. What is your proposal then, where to put this information? Which is quite important especially when you're writing a multi-file document.

laurmaedje: I think we could do something like this:
```rs
    /// Path to an image file.
    ///
    /// By default, the path is relative to the file where `image` is called.
    /// When starting with `/`, it becomes relative to the project root instead
    /// (configurable via `--root` in the CLI).
```
However, the above holds true for _all_ paths, be it image, import, json, etc. So perhaps it would be better to have this information somewhere else and link to it. Not sure where though, so I would be okay with having it here for now. We need for more guide-level docs in the future and then could reconsider this.

kravchenkoloznia: @laurmaedje maybe then put it on the [Reference/Syntax](https://github.com/typst/typst/blob/main/docs/reference/syntax.md) page? Like make a new chapter "Specifying path" or just "Path".

But I don't know how the documentation build system works in this repo and how to produce links to static files from auto-generated.

laurmaedje: If there was a "Paths" section in the syntax page, it would be `[..]($syntax/#paths)`. Paths aren't really syntax, but maybe we could put it there.

kravchenkoloznia: Please check d65f347defaa53280728252359df9bd5a0aa561b

* add "Paths" section
* provide reference to the "Paths" section in all modules which are trying to resolve a path

docs/reference/syntax.md
```diff
@@ -167,3 +167,28 @@ sequence: `[\u{1f600}]`. The same kind of escape sequences also work in
...
+When a path begins with `/`, it is treated as an absolute path relative to the
+project root. For example:
+```typ
+#image("/assets/logo.png")
+```
+
+Typst allows customization of the project root through the CLI option `--root`.
+By default, the project root is the directory from which Typst is run.
+If you want to set a specific folder as the root of your project, you can use:
+```typ
```
laurmaedje, line 192:
````suggestion
```bash
````

crates/typst/src/visualize/image/mod.rs
kravchenkoloznia: I believe the discussion is resolved. Re-open if I'm wrong.

laurmaedje:
Thanks! We've added a bit more details about roots and packages.

PR #4895, opened 2024-09-04
Fix typo in `block.clip` documentation

Jacobgarm:
Fixes a typo in the documentation for `block.clip` added in #4870


Dherse:
Thanks @Jacobgarm, I had actually fixed it but forgot to push the change before it got merged 💀

PR #4900, opened 2024-09-04
Fixed-point decimal type

PgBiel:
Adds a fixed-point decimal type to Typst, `decimal`. The intended usage is for calculations which require high precision in basic arithmetic operations, such as in finance. (Worth noting that greater precision cannot be guaranteed for more complex operations, such as power to non-integer exponents and logarithm.)

The current implementation uses [`rust_decimal`](https://lib.rs/rust_decimal), in which there isn't a fixed amount of fractional digits for each decimal number (in principle). Rather, a fixed-size integer is used for all decimal numbers for efficiency purposes, which means that numbers with more fractional digits have less non-fractional digits available (i.e. a smaller range), and vice-versa - there is a limit of about 28 significant digits in base 10, and operations can decrease the precision as values get too large. We think this is an acceptable tradeoff for now, and could consider introducing some sort of "big decimal" support in the future (with arbitrary amounts of digits), but it wouldn't be interesting to do so now due to the potential (and likely unnecessary in most cases) performance implications.

## Usage

In principle, this type can be used through the `decimal("123.456")` constructor. It only accepts a string. A literal format could be introduced in the future to make it easier to create decimals, but I think we can refrain from changing the syntax for now and wait until we get proper feedback on usage of decimal numbers.

One alternative could be to accept an integer and a scale as well, e.g. `decimal(123456, 3) == decimal("123.456")`. I'd want to be able to construct decimals like this somehow without going through a string, though just `decimal(integer, scale)` could be confusing to read - I'd love to hear other opinions on this (maybe we could use a function with named arguments).

Typical operations are available: `+ - * /`. Also, we'll attempt to support as many `calc` functions for decimal as possible, such as `calc.pow`, `calc.rem` and so on. Worth noting that many operations, such as logarithm, sqrt, power to non-integer exponents etc. will not have great precision guarantees (i.e. will not be much better than using a float) due to computational limitations.

Note that `decimal` does not support plus or minus infinity, so overflow/underflow in operations will result in an error.

Additionally, while operations between `decimal` and integers are always supported, operations between `decimal` and floats (e.g. `decimal("2.5") * 3.4`) are an error to avoid unintentional loss of precision. You'd need to explicitly cast your decimal to a float first.

## :warning: Breaking Changes

- `calc.floor(float)`, `calc.ceil(float)` and `calc.trunc(float)` now return floats instead of integers as the rounded float might be larger than the max integer (or less than the minimum integer).
- `calc.round` no longer accepts negative digits.
- `calc.round(int)` will now always return an integer, even if the `digits` parameter is non-zero.
- `int(float)` will now error instead of saturating if the float is greater than the maximum integer or less than the minimum integer.

## TODO

- [x] Finish adding decimal support to calc functions (div-euclid, rem-euclid included).
- [x] Add more conversions between `decimal` and other numeric types, e.g. `float(decimal)` to cast to a float. Include a warning if trying to write `decimal(float literal)`.
- [x] Adapt `str(decimal)` for consistency with float and int (use minus sign).
- [x] Evaluate the semantics of `decimal.display(digits: ...)` and whether this function should be kept.
- [x] Cleanup/Remove unnecessary `Decimal` internal trait implementations and operations.


tests/suite/foundations/decimal.typ
```diff
@@ -0,0 +1,51 @@
...
+// Error: 10-17 expected integer or string, found type
+#decimal(decimal)
+
+--- decimal-constructor-bad-value ---
+// Error: 10-17 invalid decimal: 1.2.3
+#decimal("1.2.3")
+
+--- decimal-repr ---
+// Test the `repr` function with decimals.
+#set page(width: auto)
```
laurmaedje, line 16: This test doesn't need a reference image. Better just use `test` assertions.

PgBiel: Good one. We may want to update the float and int tests as well (I copied this from float).

laurmaedje: Yes there are various tests across the test suite that could be migrated away from images. I didn't have the time to rewrite them all, but I try to avoid unnecessary images for new ones.

crates/typst/src/foundations/calc.rs
```diff
@@ -866,45 +972,77 @@ pub fn div_euclid(
 /// value of the divisor if the dividend is much smaller in magnitude than the divisor
 /// and the dividend is negative. This only applies for floating point inputs.
 ///
+/// In addition, this can error if given a [`{decimal}`]($decimal) input and
+/// the dividend is too small in magnitude compared to the divisor.
+///
```
PgBiel, line 977: I do wonder if we should just return some specific sentinel value (e.g. the divisor itself) when this happens instead of raising an error (given the warning above for floats). Not sure what would be mathematically correct though, and I guess ensuring the range of the returned remainder is 0.0 <= r < divisor is more important.

Same applies to `calc.rem`.

crates/typst/src/foundations/calc.rs
```diff
@@ -660,44 +709,57 @@ pub fn trunc(
...
     /// The number to round.
-    value: Num,
-    /// The number of decimal places.
+    value: DecNum,
+    /// The number of decimal places. Must not be negative.
     #[named]
     #[default(0)]
-    digits: i64,
-) -> Num {
+    digits: u32,
```
PgBiel, line 754: I feel like this is the best way forward. It greatly simplifies the function and removes previously inconsistent behavior.

(This felt noteworthy to add a comment about, in case any further discussion is needed.)

crates/typst/src/foundations/calc.rs
```diff
@@ -915,45 +1053,116 @@ pub enum Num {
...
+            Self::Decimal(d) => Some(d),
+            Self::Int(i) => Some(Decimal::from(i)),
+            Self::Float(_) => None,
+        }
+    }
+
+    /// Tries to apply a function to two decimal or numeric arguments.
+    ///
+    /// Fails with `None` if one is a float and the other is a decimal.
+    fn apply2<T>(
```
PgBiel, line 1115: This function's signature is technically inconsistent with `apply3` below, as it allows for arbitrary return values (`T`). I wonder if we should give it a different name, but I couldn't think of anything better so far.

crates/typst/src/foundations/calc.rs
```diff
@@ -977,3 +1186,11 @@ cast! {
 fn too_large() -> &'static str {
     "the result is too large"
 }
+
+/// The hinted error message when trying to apply an operation to decimal and
+/// float operands.
+#[cold]
+fn cant_apply_to_decimal_and_float() -> HintedString {
+    HintedString::new(eco_format!("cannot apply this operation to a decimal and a float"))
```
PgBiel, line 1194: In principle I didn't feel like including the operation name explicitly here, but let me know if that would be desired

crates/typst/src/foundations/float.rs
```diff
@@ -122,6 +122,7 @@ cast! {
     v: f64 => Self(v),
     v: bool => Self(v as i64 as f64),
     v: i64 => Self(v as f64),
+    v: Decimal => Self(f64::try_from(v).map_err(|_| eco_format!("invalid float: {}", v))?),
```
PgBiel, line 125: I've looked at the implementation of `impl TryFrom<rust_decimal::Decimal> for f64`, used for this cast, and it doesn't seem to actually ever return an error, but I'm handling the error case anyway just in case. Never know when stuff might change under us...

crates/typst/src/foundations/decimal.rs
```diff
@@ -0,0 +1,365 @@
...
+            }
+        }
+    }
+}
+
+impl FromStr for Decimal {
+    type Err = rust_decimal::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        rust_decimal::Decimal::from_str_exact(s).map(Self)
```
PgBiel, line 281: todo: maybe move the `MINUS_SIGN` replacement here since we own the type anyway

PgBiel (commented):
Noteworthy stuff:

crates/typst/src/foundations/decimal.rs
PgBiel: I changed my mind, probably better to use EcoString's replacement directly

crates/typst/src/foundations/calc.rs
laurmaedje: Not sure. We can always remove the error later.

crates/typst/src/foundations/calc.rs
laurmaedje: Hmm, yeah. It's ok for now.

Andrew15-5:
> * `calc.round` no longer accepts negative digits.

You mean negative numbers? This restriction makes no sense.

https://typst.app/docs/reference/foundations/calc#functions-round

Issue #4194, opened 2024-05-20
Literally input another file


m-kru:
### Description

Is there any way to literally input one .typ file into another one? There are `#include` and `#import` functions. However, they treat files as separate modules.

### Use Case

Sometimes you just want to split document into multiple files without organizing the whole document into separate modules.

laurmaedje:
This is not supported and there are no plans to add it. The fact that modules are free-standing makes them predictable and robust. If things are included directly, many things can go wrong with variables names etc. I understand that this might be convenient in some cases, but I believe the addition of this feature would in total deteriorate the user experience of Typst and its packages. Because it's the "easy way" and people would start using it where not appropriate. As a secondary point, the encapsulation of `import` and `include` is also important from a performance and incremental compilation perspective because they can be considered in isolation.

Enivex:
> This is not supported and there are no plans to add it. The fact that modules are free-standing makes them predictable and robust. If things are included directly, many things can go wrong with variables names etc. I understand that this might be convenient in some cases, but I believe the addition of this feature would in total deteriorate the user experience of Typst and its packages. Because it's the "easy way" and people would start using it where not appropriate. As a secondary point, the encapsulation of `import` and `include` is also important from a performance and incremental compilation perspective because they can be considered in isolation. 

You may want to close https://github.com/typst/typst/issues/595 as well then

qwerzl:
> This is not supported and there are no plans to add it. The fact that modules are free-standing makes them predictable and robust. If things are included directly, many things can go wrong with variables names etc. I understand that this might be convenient in some cases, but I believe the addition of this feature would in total deteriorate the user experience of Typst and its packages. Because it's the "easy way" and people would start using it where not appropriate. As a secondary point, the encapsulation of `import` and `include` is also important from a performance and incremental compilation perspective because they can be considered in isolation.

@laurmaedje May us request a reconsideration? I believe the freedom to choose should be given to users and module authors. Such "bad practices" might not seem perfect but they might come in handy in some circumstances especially for a markup language where people'd care less about code quality and just want to get things done fast.

weltonrodrigo:
I just want to keep a style.typ as a separate file from my main.typ 😞

alerque:
I think it's worth reiterating that this would degrade the experience for everybody, even those not using it directly (the incremental compilation inferences would regress). Also this could be accomplished for those that want it for programmatic reasons with preprocessing.

Issue #754, opened 2023-04-12
Functions defined in one file are not recognised in included file


nikmennega:
```
// main.typ
= Define our `let` binding
#let simple = emph

We have defined a #type(simple) named `#simple`.

= Using `#simple` in main file

#simple[
	#lorem(8)
	Come on in, the lorem's fine!
	#lorem(12)
]

#include("included.typ")
```

```
// included.typ
= Using `#simple` in imported file

#simple[
	#lorem(8)
	Some great lorem in here!
	#lorem(12)
]
```
## Expected behaviour
<img width="518" alt="image" src="https://user-images.githubusercontent.com/1085754/231418318-4acc746c-87ff-4f6a-a664-d66d12cb7282.png">

## Experienced behaviour
```
watching main.typ
writing to main.pdf

[19:36:47] compiled with errors

error: unknown variable
  ┌─ included.typ:7:1
  │
7 │ #simple[
  │  ^^^^^^

help: error occurred while importing this module
   ┌─ main.typ:15:8
   │
15 │ #include("included.typ")
   │         ^^^^^^^^^^^^^^^^
```

laurmaedje:
This is actually working as expected. Each module has its own scope and if you want to have shared functions its best to define them in a third file that you import in `main.typ` and `included.typ` with [`#import`](https://typst.app/docs/reference/scripting/#modules).

laurmaedje:
The reason for this is that this way a module is free-standing. In your proposed design a module could result in different things depending on from where it is imported. That would complicate things a lot.

laurmaedje:
Side note: The syntax is `#include "text/lorem.typ"`, the parentheses just happen to be allowed.

laurmaedje:
If I understand what you want correctly, the below setup would lead to "1 2" being displayed in your proposal. That means that `foo.typ` needs to be evaluated twice, once with the context of `a.typ` and once with the context of `b.typ`. Currently, each module is evaluated just once.
```
// main.typ
#include "a.typ"
#include "b.typ"

// a.typ
#let foo = 1
#include "foo.typ"

// b.typ
#let foo = 2
#include "foo.typ"

// foo.typ
#foo
```

nikmennega:
That explanation makes things a bit more clear, thank you.

[...]

Issue #595, opened 2023-04-05
Imports that are global to the project


paulgoulain:
Hi everybody,

As an exercise, I started to redo my PhD manuscript using Typst. Originally I did it using Latex, with a main.tex file, and imports of the different chapters (chapter1.tex, chapter2.tex, ...). For physic notation typesetting, I use the [typst-physics](https://github.com/Leedehai/typst-physics) package. I noticed that even if I import this package in the main. typ file, I still need to import it in the chapter files too. 

Is it the expected behavior or am I doing something wrong ? Is there a way to make imports made in a parent file to propagate its children ?

laurmaedje:
An import is indeed only available in one file. If you import multiple packages and don't want to repeat all of the imports in each chapter, you can create a single file with all imports and import only that file in your chapters.

astrale-sharp:
It's how scope usually works (in many many programming language),

If you don't specify where a language item comes from, the language probably shouldn't guess.
This is not an issue so can you please close this @paulgoulain ? I hope this was informative, otherwise there is the beginner channel of the discord!

paulgoulain:
Well then let's turn this into a feature request :)

Is it desirable to be able to make imports global to a whole document ? Something similar to the Latex usepackage ?

laurmaedje:
Relevant other discussion: #754. I am not a fan of global imports as they fundamentally mess with how the module system works and break the notion that a module is well-defined on its own.

Jollywatt:
I guess what people want is not to break the module system, but to have files that aren’t tied to modules.

I.e., to have a way of sourcing a file that is equivalent to inserting the file’s contents where it is imported — without any of the hassle of that comes with a module namespace.

(Compare python’s `import` for files _and_ modules to Julia’s `include()` for files and `import/using` for modules. One of those is much easier to use than the other.)

7sDream:
The raw input mechanism is useful for big document, I used it [heavily](https://github.com/7sDream/scp-pdf/blob/master/part01/index.tex) when using TeX. Constants/shortcut commands needn't to be imported in each real document file, only once in main file, it saves a lot of time and very intuitive.

This can be done in many different way:

- Way 1, insert content of input file in lex/parser stage
  - [`#include` preprocessing directive](https://gcc.gnu.org/onlinedocs/cpp/Include-Operation.html) in C/C++
  - [`include!` macro](https://doc.rust-lang.org/std/macro.include.html) in Rust
[...]

PR #4907, opened 2024-09-06
Document caveat re large Typst projects

kapple19:
Documenting a sorely needed caveat for LaTeX users who are used to having their definitions available in any included file, especially for large scale Typst projects with many (possibly nested) `include`s.

Addresses #595, #754, #4194 in the documentation.

### DIFF
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -670,3 +670,8 @@ applicable, contains possible workarounds.
   break will be created if this spot would be a natural page break anyways. You
   can also use `[#v(1fr)]` to distribute space on your page. It works quite
   similar to LaTeX's `\vfill`.
+
+- **Boilerplate for large scale projects.** LaTeX definitions are available in
+  included files. This is not the case in Typst, where you'll need your definitions
+  in a file of its own, and manually `import` them into each of your project's
+  `include`d files. Alternatively your Typst project can be in one very large file.
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -674,9 +674,18 @@ applicable, contains possible workarounds.
 - **Separation of `import` and `include` behaviours.** `\include`ing a file in LaTeX
   has two behaviours that have been separated in Typst: imports LaTeX macros,
   and adds the document content to the parent file. In Typst: `#import`ing will
-  add definitions to the parent file, but not add document content;
-  `#include`ing will add document content to the parent file, but will not import
-  definitions. The syntax for including document content is simply `#include "file.typ"`,
+  add definitions (equivalent to LaTeX macros) to the parent file,
+  but not add document content; `#include`ing will add document content
+  to the parent file, but will not import definitions.
+  The syntax for including document content is simply `#include "file.typ"`,
   but the syntax for importing definitions provides more flexibility:
   `#import "file.typ": *` will import all of `file.typ`'s definitions;
   `#import "file.typ": x, y, template` will import only the named definitions.
+  It is of particular note for those familiar with LaTeX that
+  in LaTeX, a file only needs to be `\include`d once;
+  in Typst, your global definitions
+  and global setups need to be placed in a separate file,
+  and that file is then `#import`ed and/or `#include`d
+  into all of your project's nested files in addition to the main file.
+  This requirement and design improves the robustness, predictability,
+  and performance of Typst, as every file is well-defined in isolation.
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -671,7 +671,12 @@ applicable, contains possible workarounds.
   can also use `[#v(1fr)]` to distribute space on your page. It works quite
   similar to LaTeX's `\vfill`.
 
-- **Boilerplate for large scale projects.** LaTeX definitions are available in
-  included files. This is not the case in Typst, where you'll need your definitions
-  in a file of its own, and manually `import` them into each of your project's
-  `include`d files. Alternatively your Typst project can be in one very large file.
+- **Separation between `import` and `include`.** LaTeX macro definitions in files are
+  automatically available in documents which `\include` those files' contents. In Typst,
+  including a file's contents (done through `#include "file.typ"`) is separate from
+  importing its definitions, which must be done with a separate
+  `#import "file.typ": definitions` command. With it, you can either list the definitions
+  to import individually, which is more usual when they are few (e.g.
+  `#import "file.typ": x, y, template`), or, to fully mimic LaTeX's behavior, you may use a
+  wildcard import to import all definitions available in another file
+  (e.g. `#import "file.typ": *`).
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -686,6 +686,6 @@ applicable, contains possible workarounds.
   in Typst, your global definitions
   and global setups need to be placed in a separate file,
   and that file is then `#import`ed and/or `#include`d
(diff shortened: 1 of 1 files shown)


docs/guides/guide-for-latex-users.md
```diff
@@ -670,3 +670,8 @@ applicable, contains possible workarounds.
   break will be created if this spot would be a natural page break anyways. You
   can also use `[#v(1fr)]` to distribute space on your page. It works quite
   similar to LaTeX's `\vfill`.
+
+- **Boilerplate for large scale projects.** LaTeX definitions are available in
+  included files. This is not the case in Typst, where you'll need your definitions
+  in a file of its own, and manually `import` them into each of your project's
+  `include`d files. Alternatively your Typst project can be in one very large file.
```
PgBiel, line 677: This doesn't seem like the best wording for it, and an example would be nice as well. I'm not sure this belongs in "Limitations" either, since the feature does exist in Typst. Here's an initial idea on how to improve it, but larger-scale changes (e.g. moving to another section or merging with the existing text on this) may be needed.

```suggestion
- **Separation between `import` and `include`.** LaTeX macro definitions in files are
  automatically available in documents which `\include` those files' contents. In Typst,
  including a file's contents (done through `#include "file.typ"`) is separate from
  importing its definitions, which must be done with a separate
  `#import "file.typ": definitions` command. With it, you can either list the definitions
  to import individually, which is more usual when they are few (e.g.
  `#import "file.typ": x, y, template`), or, to fully mimic LaTeX's behavior, you may use a
  wildcard import to import all definitions available in another file
  (e.g. `#import "file.typ": *`).
```

PgBiel (commented):
This is already mentioned under the "Packages" section, though I'd agree that it needs to be improved (e.g. to include an example, as well as to mention wildcard import):

> If you need to load functions and variables from another file within your project, for example to use a template, you can use use _[sic.]_ the same [import](https://typst.app/docs/reference/scripting/#modules) statement with a file name rather than a package specification. To instead include the textual content of another file, you can use an [include](https://typst.app/docs/reference/scripting/#modules) statement. It will retrieve the content of the specified file and put it in your document.

docs/guides/guide-for-latex-users.md
kapple19: Thank you so much. I did want to get the ball rolling on this caveat because this limitation of Typst frustrates me, and I wanted it to be in the list of limitations for LaTeX users, particularly since there are at least three different issues reported by users on this, so the documentation is definitely not clear enough.

PgBiel: Understood, thank you for your input. Would be nice to see what other LaTeX users think of this change as well.

FlorentCLMichel: As a long-time LaTeX user, I find this very helpful. I stumbled on this a few times and, while it's not (at least from my experience) a major blocker for moving from LaTeX to Typst, making it as explicit as possible can probably make the experience even smoother than it already is.

FlorentCLMichel (commented):
It seems to me that that ‘With it, you can either list the definitions’ would read better without the ‘With it’.
(I'm not a native English speaker, though, so I may be wrong.)

kapple19:
> It seems to me that that ‘With it, you can either list the definitions’ would read better without the ‘With it’. (I'm not a native English speaker, though, so I may be wrong.)

Yeah okay. I think the sentence is still clear with out "With it", and becomes more concise.
As a native English speaker, I also don't have a problem with it.

What I do have some small confusion over is the choice of the word "separation"/"separate".
I think it reads better as "Difference between `import` and `include`, and "In Typst, including a file's contents is different from importing its definitions" which I think I'll change to.

laurmaedje:
I think it would be useful to highlight the differences between Typst and LaTeX in how structuring larger projects with helper definitions and chapter files works. However, I would not have this as part of the sections about limitations. It's not a limitation, it's just different—in fact, it was an intentional design decision.

When wanting to use definitions across many files, the Typst way is not to define them in the main file, but to have them in a shared utility file which is imported in every chapter. Requiring this import in every chapter might need some getting used to, but it improves robustness, predictability, and performance, as every file is well-defined in isolation.

kapple19:
> I think it would be useful to highlight the differences between Typst and LaTeX in how structuring larger projects with helper definitions and chapter files works. However, I would not have this as part of the sections about limitations. It's not a limitation, it's just different—in fact, it was an intentional design decision.
> 
> When wanting to use definitions across many files, the Typst way is not to define them in the main file, but to have them in a shared utility file which is imported in every chapter. Requiring this import in every chapter might need some getting used to, but it improves robustness, predictability, and performance, as every file is well-defined in isolation.

I am happy you've found a solution you're happy with, especially with it being robust, predictable, and performant. That modularity is somewhat valuable.

Where it is _not_ valuable is in the nature of the fact that it's quite rare for people to just share a segment of their document to people. People are fine just sharing the whole document. This modularity you value is only valuable in the implementation of Typst's design, not valuable in the utility of having produced documents.

> It's not a limitation, it's just different

It's a limitation in that it requires the user to implement boilerplate copies of `#import`s and `#include`s throughout any nested Typst files, and is unscalable for large documentation projects. See my moody rant about this [here](https://github.com/typst/typst/issues/754#issuecomment-2333043793).

It's a convenient design for _you_, yes I value and appreciate that. It's not a convenient design for your users, and doesn't have as much value and utility for actual handling of document outputs.

And if it's a decision you choose to keep, I'm fine with that. Typst will by this singular reason not be for me, who likes building a nice, large, interlinked wiki of study topics. But multiple people have opened GitHub issues on this topic, and since LaTeX is excellent with reducing a user's need for boilerplate, it's a grating requirement when trying out Typst for large projects. And at the very least, it should be included in the list of limitations.

kapple19:
If anyone has a nicer, more succinct wording, please feel free to improve on my tendency for verbosity. I'm still not satisfied with how I've worded it, but have hopefully nicely incorporated everyone's suggestions.

kapple19:
@laurmaedje if you prefer, we could just change "limitations" to "differences" in the title of comparison to LaTeX? I'd be fine with that. I would just like this explanation in the documentation for people coming from LaTeX, especially seeing as there've been quite a few opened issues on this.

docs/guides/guide-for-latex-users.md
```diff
@@ -670,3 +670,22 @@ applicable, contains possible workarounds.
...
+  The syntax for including document content is simply `#include "file.typ"`,
+  but the syntax for importing definitions provides more flexibility:
+  `#import "file.typ": *` will import all of `file.typ`'s definitions;
+  `#import "file.typ": x, y, template` will import only the named definitions.
+  It is of particular note for those familiar with LaTeX that
+  in LaTeX, a file only needs to be `\include`d once;
+  in Typst, your global definitions
+  and global setups need to be placed in a separate file,
+  and that file is then `#import`ed and/or `#include`d
+  into all of your project's nested files in addition to the main file.
```
ssotoen, line 689:
```suggestion
  into the files that make use of these definitions.
```

Just like with every good programming language, you only need to import stuff to where you use it.

kapple19: This would call for three boilerplate lines of code then.
1. Any `#include`s.
2. `#import` of definitions specific to the file.
3. `#include`/`#import` of setups that are used in every file, e.g. default font.

laurmaedje:
We've discussed this and come to the conclusion that we'll add a better explanation of how to set up a multi-file project in Typst elsewhere in the docs and then add a short link from the LaTeX guide, but not in the limitations section.

kapple19:
@laurmaedje thanks so much! Whether it was documented where I suggested or given a proper treatment as you've decided (which is much better, thank you), I didn't mind, as long as it was addressed.

PR #4909, opened 2024-09-06
Make `QuoteElem` locatable

### LINKED ISSUE #4906: Make QuoteElem locatable; or configurable locatables?
### Description

Evolving from a Discord chat, it would be great if ``QuoteElem`` would be also locatable. Justified reasons against it can be performance losses, but it should be checked whether this really makes a big performance loss.

Another idea I had that came up but my Rust knowledge is probably to less of a kind, is, to make all elements configurable if they should be locatable. So the user could decide which elements should be located and which won't. It's probably easier to talk about than it is to implement, but I think it could be a significant difference to the current system.

### Use Case

Could allow third-party applications to work more seamlessly with the Typst compiler. If I build for example a tool to read all quotes I could allow them to be locatable, disable locating for all other types and get some perf improvements.

Coekjan:
Closes #4906


laurmaedje:
Thanks!

PR #4912, opened 2024-09-06
Add neg math spacing.

SolidTux:
This adds a negative `-1/6em` math spacing option for parity with LaTeX's `\!`. This can be useful in various situations, especially to manually fix inconsistencies.


Enivex:
Wouldn't it make more sense to have it be `thin.neg`?

ssotoen:
For complete parity with LaTeX you should also add negative variants of `med` and `thick`, and if you do that you might as well include `quad` and `wide` too. LaTeX’s `enspace` is also missing.

Enivex:
> For complete parity with LaTeX you should also add negative variants of `med` and `thick`

Not sure what you mean. Isn't this the only named one?

ssotoen:
`\negmedspace` and `\negthickspace` both exist, they just don't have a shorthand.

laurmaedje:
> Wouldn't it make more sense to have it be thin.neg?

Not possible currently since it's not a symbol, but rather just a `#let thin = h(-1em/6)` :/

But in principle I agree with you.

SolidTux:
I also agree that `*.neg` would be a better option. Is there a reason why in math mode `h(...)` instead of the symbols is used?

laurmaedje:
Symbols are strings. I'm not sure there are Unicode codepoints for all these space widths.

SolidTux:
You are right, I found the following Unicode characters on the [whitespace wiki page](https://en.wikipedia.org/wiki/Whitespace_character)

* `thin`, `1/6em`: THIN SPACE `U+2009` (`1/5em` or `1/6em`) or SIX-PER-EM-SPACE `U+2006`
* `med`, `2/9em`=`1/3em`: MEDIUM MATHEMATICAL SPACE `U+205F` (THREE-PER-EM SPACE `U+2004` exists as well)
* `thick`, `5/18em`: ?
* `quad`, `1em`: EM SPACE `U+2003` or EM QUAD `U+2001`
* `wide`, `2em`: ?

Also, sorry if I started this the wrong way, I probably should have explained more in the original post.

This was more meant as a discussion starter because I found implementing it much easier than explaining what I mean.

laurmaedje:
Since we can't have the nice design (`thin.neg`) with our current infrastructure, I'd be inclined to close this for now and maybe revisit the topic in the future.

PR #4916, opened 2024-09-08
Add `math.longdivision` element function

mkorje:
<details>
<summary>Tests</summary>

| LuaTeX | Typst |
| - | - |
| div-latex | div |

<details>
<summary>Code</summary>

```tex
\documentclass[11pt]{article}
\usepackage[paperheight=4cm,paperwidth=5cm,margin=0cm]{geometry}
\usepackage{amsmath}
\usepackage{unicode-math}
\usepackage{fontspec}
\usepackage[default]{fontsetup}
\begin{document}

\begin{align*}
    32 \longdivision{252.8} \quad 5 \longdivision{12345} \\
    x-3 \longdivision{x^2 + 3x - 5}
\end{align*}
\[ \longdivision{X^{X^{X^X}}} \quad \longdivision{\frac{a}{b}} \]

\end{document}
```

```typst

$ 32 longdivision(252.8) quad
  5 longdivision(12345) \
  x - 3 longdivision(x^2 + 3x - 5) $

$ longdivision(X^X^X^X) quad
  longdivision(a/b) $
```

</details>

</details>

Any more advanced behaviour, like in the TeX package [`longdivision`](https://ctan.org/pkg/longdivision?lang=en), is probably best left to a package or the (future) std library.

I made a few (bit hacky) examples of doing long division, see below (no idea if the division is right, I just copied them from the TeX StackExchange).

<details>
<summary>Examples</summary>

div

<details>
<summary>Code</summary>

```typst
#set page(width: 10cm, height: auto)

$
  longdivision(33) quad
  longdivision(a=b) quad
  3 longdivision(200) quad
  x + 1 longdivision(x^2) \
  10 longdivision(1000) quad
  56 longdivision(3678) quad
  3 longdivision(37678)
$

#grid(columns: (1fr, 1fr), row-gutter: 0.5cm,
$
  vec(delim: #none, gap: #0em,
    670&,
    3 longdivision(2011),
    -underline(18) #hide[11],
    21 #hide[1],
    -underline(21) #hide[1],
    01,
  )
$,
$
  vec(delim: #none, gap: #0em,
    7.9&,
    32 longdivision(252.8),
    underline(-224) thick arrow.b,
    28 thick 8,
    underline(-28 thick 8),
    0,
  )
$,
$
  vec(delim: #none, gap: #0em,
    2469&,
    5 longdivision(12345),
    underline(10) #hide[345],
    23 #hide[45],
    underline(20) #hide[45],
    34 #hide[5],
    underline(30) #hide[5],
    45,
    underline(45),
    0,
  )
$,
$
  vec(delim: #none, gap: #0em,
    949&,
    13 longdivision(12345),
    underline(117) #hide[45],
    64 #hide[5],
    underline(52) #hide[5],
    125,
    underline(117),
    8,
  )
$
)

$
  vec(delim: #none, gap: #0.15em,
    1+x+2x^2+3x^3+5x^4+dots.c.h&,
    1-x-x^2 longdivision(1 #hide[$+x+2x^2+3x^3+5x^4+dots.c.h$]),
    underline(1-x-#hide[2]x^2) #hide[$+3x^3+5x^4+dots.c.h$],
    x+#hide[2]x^2 #hide[$+3x^3+5x^4+dots.c.h$],
    underline(x-#hide[2]x^2-#hide[3]x^3) #hide[$+5x^4+dots.c.h$],
    2x^2-#hide[3]x^3 #hide[$+5x^4+dots.c.h$],
    underline(2x^2-2x^3-2x^4) #hide[$+dots.c.h$],
    3x^3+2x^4 #hide[$+dots.c.h$],
    dots.v #hide[$+dots.c.h$],
  )
$

$
  vec(delim: #none, gap: #0.15em,
    x^3-#hide[2]3x^2+#hide[2]6x-#hide[1]4&,
    2x-3 longdivision(2x^4-9x^3+21x^2-26x+12),
    underline(-(2x^4-3x^3)) #h(7.9em),
    -6x^3+21x^2-26x+12,
    underline(-(-6x^3+#hide[2]9x^2)) #h(4.65em),
    12x^2-26x+12,
    underline(-(12x^2-18x)) #h(1.82em),
    -8x+12,
    underline(-(-8x+12)) #h(-0.4em),
    0,
  )
$
```

</details>

</details>


Enivex:
In my opinion even this belongs in a package. This is a very specific piece of notation, which also is not standardized.

sitandr:
I'm really not sure this should be in standard library. This "operation" seems very rare to me and limited to very specific area and "localization". Personally, I've never seen this sign in math classes. I'm sure it isn't used in my country at all.

I don't say it is useless, but rather than I believe that it should be a package. 

In package designated for this, one can do all the kinds of algorithms for rendering the long division. However, it's not the place for them in std. That will look very strange for that specific area, I'm sure.

So I want to say that this sign wouldn't be used much outside that package for long division. Then why don't we ask the package to do the sign to? It's not hard.

(Don't take my criticism very seriously, it's still great you want to contribute to compiler anyway, and maybe I'm wrong about this)

Enivex:
I initially wrote something similar to @sitandr. It is a Unicode symbol though (which surprised me), so I don't immediately see an issue with just making it available in the same way the square root symbol is available.

Though I'm not sure I love the `longdivision` name

mkorje:
I agree that this symbol is very specialised. But it is a Unicode symbol and is meant (I _believe_) to be laid out following the same method as a root. So while a package could probably somewhat replicate this, I imagine it'd be quite difficult to do. Particularly, you'd need information from the font's MATH table, which simply isn't accessible by a package in Typst.

I haven't grouped `math.longdivision` with roots in the docs because they feel fairly different semantically. But perhaps it is better to have it grouped with roots so it isn't quite as emphasised, reflecting its specialised nature.

> Also, not sure I'm a fan of the `longdivision` name.

I share your sentiment. `longdivision` is what LaTeX uses. `long-division` is more Typst-y, but the hyphen is a problem in math mode. And something like `division` feels too broad of a name. I don't really know what else to name it.

laurmaedje:
> users may expect actual layout features to be available

this is a concern of mine as well.

also, I think semantically it's like frac right? so it should rather take divisor as an argument as well?

mkorje:
Not sure what we can do about users expecting actual layout features (other than not adding this).

I agree it'd make more sense taking divisor as another argument.

Enivex:
Does that not just make it less flexible? The more apt comparison here would be `sqrt`. If we're leaving layout to packages then I see no need for another argument.

laurmaedje:
Sqrt is how it is layouts, but frac is what its semantics are, isn't it?

Enivex:
> Sqrt is how it is layouts, but frac is what its semantics are, isn't it?

If we're talking semantics, then shouldn't there be three numbers involved (divisor, dividend, quotient)?

mkorje:
We could just make the divisor argument optional.

laurmaedje:
> If we're talking semantics, then shouldn't there be three numbers involved (divisor, dividend, quotient)?

I have no clue. I thought it was just another way to write division.

Enivex:
> > If we're talking semantics, then shouldn't there be three numbers involved (divisor, dividend, quotient)?
> 
> I have no clue. I thought it was just another way to write division.

It's specifically used for the way long division is organized in a tableau in English-speaking countries.

mkorje:
> It's specifically used for the way long division is organized in a tableau in English-speaking countries.
> 
> Edit: They write the dividend inside, the divisor to the left, and the resulting quotient on top

Adding the resulting quotient on top as well feels like we'd be doing too much layouting, and would only make users more likely to expect actual layout features to be available.

Enivex:
> > It's specifically used for the way long division is organized in a tableau in English-speaking countries.
> > 
> > Edit: They write the dividend inside, the divisor to the left, and the resulting quotient on top
> 
> Adding the resulting quotient on top as well feels like we'd be doing too much layouting, and would only make users more likely to expect actual layout features to be available.
> 

That's sort of what I was getting at though. What's the difference between including the divisor and including the quotient?

mkorje:
> That's sort of what I was getting at though. What's the difference between including the divisor and including the quotient?

That's a fair point. I guess layout wise, including the divisor is just adding some content before on the same baseline, but adding the quotient involves stacking content on top.

GrimbleFitz:
So for now, how would I go about formatting long division? I'm sorry if this is the wrong place to ask, but it's not clear to me from this conversation how to do that. Thanks!

Enivex:
> So for now, how would I go about formatting long division? I'm sorry if this is the wrong place to ask, but it's not clear to me from this conversation how to do that. Thanks!

Please ask on Discord or the forum

laurmaedje:
Ultimately, what I'm concerned about is designing APIs too much based on how things work internally, rather than the semantics they carry. For what it's worth, MathML (not MathML Core though I think), [does handle divisor and dividend](https://www.w3.org/TR/MathML/chapter3.html#presm.mlongdiv) and additional stackable elements.

laurmaedje:
How shall we proceed here?

mkorje:
Yeah I'm not quite sure how we should proceed here. I think supporting divisor, dividend and more like in MathML (not core) is a bit too much to have built in, especially when we are probably going to be using MathML Core for HTML output.

laurmaedje:
I think if we consider the full thing out of scope, then we should probably also not have the half thing?

mkorje:
Yes, I agree we shouldn't then. I'll close this PR then. We can always revisit it in the future if it is requested.

PR #4923, opened 2024-09-09
Reduce typo count.

waywardmonkeys:


crates/typst-syntax/src/ast.rs
```diff
@@ -1439,7 +1439,7 @@ pub enum BinOp {
     Assign,
     /// The containment operator: `in`.
     In,
-    /// The inversed containment operator: `not in`.
+    /// The inverse containment operator: `not in`.
```
slanterns, line 1442: I don't think it's a typo.

waywardmonkeys: Inverse is already an adjective, so you don't really need "inversed" in commonly used English. You might say "inverted", but that isn't typical either in this scenario. In math, you'd call it an "inverse function" or "a function and its inverse".

PgBiel (approved):

PR #4924, opened 2024-09-09
`subset` and `supset` variants

MDLC01:
This PR adds many variants for `subset` and `supset`.

The corresponding codepoints are part of the "Subset and superset relations" sub-block of the ["Supplemental Mathematical Operators" Unicode block](https://unicode.org/charts/nameslist/n_2A00.html), with the exception of `subset.open` and `supset.open`, which come from Miscellaneous symbols in [Miscellaneous Mathematical Symbols-A](https://unicode.org/charts/nameslist/n_27C0.html), `subset.arrow.r`, `supset.arrow.l` and `supset.arrow.r`, which come from Arrows combined with relations in [Supplemental Arrows-B](https://unicode.org/charts/nameslist/n_2900.html), and `subset.arrow.l`, which comes from Mathematical arrows in [Miscellaneous Symbols and Arrows](https://unicode.org/charts/nameslist/n_2B00.html).

I don't like the asymmetry between `{subset,supset}.arrow.l` and `{subset,supset}.arrow.r`. I've mostly included them for completeness, but I'm not sure if we should keep them for now.

By analogy with `subset.supset` and `supset.subset` (added in this PR), I used the names `subset.subset` for ⫕ and `supset.supset` for ⫖. I'm not sure those two names are great, but it's not that bad either.

AFAIK, the only remaining subset/supset-related symbols not included in this PR are U+2AD7 ⫗ SUPERSET BESIDE SUBSET and U+2AD8 ⫘ SUPERSET BESIDE AND JOINED BY DASH WITH SUBSET. I did not add them because I don't know what to name them in a way that does not conflict with `supset.subset`, which is used in this PR for U+2AD4 ⫔ SUPERSET ABOVE SUBSET.


laurmaedje:

MDLC01:
This was converted to Proposal 18 in the document.

Issue #4335, opened 2024-06-05
Support bitmap fonts


Nerixyz:
### Description

Currently, attempting to use glyphs from fonts that have the raster image type other than [`PNG`](https://docs.rs/ttf-parser/latest/ttf_parser/enum.RasterImageFormat.html#variant.PNG) panics with:

> failed to decode image (Format error decoding Png: Invalid PNG signature.)

in [`draw_raster_glyph`](https://github.com/typst/typst/blob/681badf76a029e421415986aa3a46c0081d28470/crates/typst/src/text/font/color.rs#L58), because it assumes every color glyph has the image type `PNG`.

<sub>I'm not exactly sure what these fonts are called, as FontForge refuses to open it, because its name isn't ASCII 🙃</sub>

I made a patch to at least add support for [`BitmapMonoPacked`](https://docs.rs/ttf-parser/latest/ttf_parser/enum.RasterImageFormat.html#variant.BitmapMonoPacked) (the type of glyphs in _SimSun_) in https://github.com/Nerixyz/typst/commit/4188128332f39fd86c27569d33962a959e19bf0f, however I don't think that's the correct approach. Since the bitmap is monochrome, it should be fine to color it (→ it should be rendered as a "regular" glyph in [`render_text`](https://github.com/typst/typst/blob/681badf76a029e421415986aa3a46c0081d28470/crates/typst-render/src/text.rs#L15)). Since I'm new to the codebase, I wanted to ask where I should add support for the monochrome glyphs.

### Use Case

At least on Windows, Typst falls back to the _SimSun_ font for Chinese characters. When rendering to PNG, Typst currently panics (not really a compiler bug, I'd say, more like a missing feature).

For example, a Typst file having just `年` compiled as PNG will panic on Windows.

LaurenzV:
> At least on Windows, Typst falls back to the SimSun font for Chinese characters.

Damn... Is the default font for CJK characters on Windows really a bitmap font? I think what should be changed first is that for non-emoji glyphs, Typst avoids falling back to a bitmap font unless absolutely necessary.

Enivex:
Are we talking Windows without a CJK language pack? I can't imagine Chinese Windows uses a bitmap font.

Nerixyz:
> Are we talking Windows without a CJK language pack?

Yes. I suppose a CJK language pack installs some better fonts, as I couldn't find a decent preinstalled font.

LaurenzV:
[...]

PR #4929, opened 2024-09-09
Only scan for color glyphs in `is_color_glyph`.

inferiorhumanorgans:
This will prevent typst from picking up (unsupported) monochrome bitmaps in Windows system fonts e.g. Cambria Math and SimSun and the resulting panic (see #4335).  One path leading to panic is eliminated, however:

- unwrap is supplemented with an explicit panic that provides some debugging info.  Longer-term it would be nice to have access to logging context from `draw_raster_glyph` and to bubble up some error handling.
- unsupported color glyphs (e.g. JPEG and TIFF) will still cause a panic in PNG generation

### DIFF
--- a/crates/typst/src/text/font/color.rs
+++ b/crates/typst/src/text/font/color.rs
@@ -23,8 +23,10 @@ pub fn is_color_glyph(font: &Font, g: &Glyph) -> bool {
     // PNG glyphs but the sbix table can also contain JPEG
     // and TIFF glyphs.
     if let Some(sbix) = ttf.tables().sbix {
-        if sbix.best_strike(pixels_per_em).is_some() {
-            return true;
+        if let Some(strike) = sbix.best_strike(pixels_per_em) {
+            if strike.get(glyph_id).is_some() {
+                return true;
+            }
         }
     }
 
--- a/crates/typst/src/text/font/color.rs
+++ b/crates/typst/src/text/font/color.rs
@@ -2,7 +2,7 @@
 
 use std::io::Read;
 
-use ttf_parser::{GlyphId, RgbaColor};
+use ttf_parser::{name_id, GlyphId, RgbaColor};
 use usvg::tiny_skia_path;
 use xmlwriter::XmlWriter;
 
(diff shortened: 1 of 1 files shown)


crates/typst/src/text/font/color.rs
```diff
@@ -11,13 +11,34 @@ use crate::syntax::Span;
...
-        || ttf.is_color_glyph(glyph_id)
+    let pixels_per_em = 160;
+
+    // Ultimately this can still fail as we only support
+    // PNG glyphs but the sbix table can also contain JPEG
+    // and TIFF glyphs.
+    if let Some(sbix) = ttf.tables().sbix {
+        if sbix.best_strike(pixels_per_em).is_some() {
+            return true;
+        }
```
laurmaedje, line 28: Don't we also need to check for the glyph ID somewhere?

laurmaedje (commented):
Why do they need to be unsupported though? Can't we handle monochrome bitmap fonts just like other color fonts?

inferiorhumanorgans:
> Why do they need to be unsupported though? Can't we handle monochrome bitmap fonts just like other color fonts?

Ah, this is (mostly?) covered in #4335. 

It doesn't *need* to be unsupported but the fonts that've been reported to have monochrome bitmaps also have outline glyphs.  What I took away from the other issue is that outline glyphs are preferred unless we're dealing with emoji (color) fonts that have color bitmaps and monochrome outlines.  This unbreaks PDF rendering for the aforementioned fonts and preserves the preference for color bitmaps.

crates/typst/src/text/font/color.rs
inferiorhumanorgans: Sure

laurmaedje:
I see. So we have the problem that the priority chain is `color bitmap -> outline glyph -> mono bitmap`?

inferiorhumanorgans:
No, the current order places mono bitmaps at the same priority as color bitmaps because `glyph_raster_image` returns `Some` for both mono and color bitmaps.  The immediate problem is that most mono bitmaps are currently unsupported and typst panics as a result.  The discussion at #4335 lead me to believe there's little-to-no interest in preferring mono bitmaps over mono outlines in any situation.

Specifically the order is:
- `sbix` color bitmap (panic if JPEG or TIFF)
- `bdat` mono bitmap (panic?)
- `ebdt` mono bitmap (panic)
- `cbdt` color bitmap (panic if not PNG)
- `colr` color outline
- `SVG` color outline
- `gvar` mono outline
- `glyf` mono outline
- `CFF` mono outline
- `CFF2` mono outline

My proposal is:
- `sbix` color bitmap (panic if JPEG or TIFF)
- ~~`bdat` mono bitmap (panic?)~~
- ~~`ebdt` mono bitmap (panic)~~
- `cbdt` color bitmap (panic if not PNG)
- `colr` color outline
- `SVG` color outline
- `gvar` mono outline
- `glyf` mono outline
- `CFF` mono outline
- `CFF2` mono outline

The fonts so far that have been reported to trigger the mono bitmap code path also have mono outlines for those glyphs. This change will allow typst to use/prefer those mono outlines without interfering with e.g. Apple's emoji font (color bitmaps + mono outlines) or Noto Sans (color outlines only).

```rust
// crates/typst-render/src/text.rs for PNG rendering
pub fn render_text(canvas: &mut sk::Pixmap, state: State, text: &TextItem) {
  // This should return false for mono bitmaps that we don't support
  if is_color_glyph(&text.font, glyph) {
    let glyph_frame = frame_for_glyph(&text.font, glyph.id);
      crate::render_frame(canvas, state, &glyph_frame);
    } else {
      render_outline_glyph(canvas, state, text, id);
    }
}
```

laurmaedje:
> No, the current order places mono bitmaps at the same priority as color bitmaps because glyph_raster_image returns Some for both mono and color bitmaps. 

What I meant is that the _desired_ order is `color bitmap -> outline glyph -> mono bitmap`.

I think the changes you propose make sense in principle, but while at it, I'd like to eliminate all possible panics. Also, unless I'm mistaken, we can skip the low-level table lookups by simply checking for PNG format in the output of `glyph_raster_image`. Only color tables will generate PNGs.

I thought for the past weeks that the panic is very rare and this rather low priority, but it appears that 0.12 testing crashes on some emoji fonts in the web app, so it's more urgent than I thought. For this reason, I've implemented the changes I mention above myself: https://github.com/typst/typst/pull/5159. Feel free to take a look and give your input there.

laurmaedje:
Closing in favor of https://github.com/typst/typst/pull/5159

PR #4932, opened 2024-09-10
Add warning when SVG with foreign object is included

### LINKED ISSUE #1421: SVGs with `foreignObject` don't render correctly
### Description

Text in [Mermaid](https://mermaid.js.org) graphs exported as SVG does not appear to be rendered at all.

Screenshot 2023-06-05 at 23 45 53

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

LaurenzV:
A bit unfortunate to have to search every SVG like this (I'm also not sure if this might be very unefficient, if so I'm open for suggestions on how to improve it), but I think it's good to have a warning, so we can finally close #1421 since there isn't much we can do from our side.

I tested it manually to confirm that it works.

### DIFF
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -200,9 +200,12 @@ fn layout_image(
             .windows(needle.len())
             .any(|s| s == needle)
         {
-            engine.sink.warn(warning!(span, "image contains foreign object";
-            hint: "svg images with foreign objects might render incorrectly in typst";
-            hint: "see https://github.com/typst/typst/issues/1421 for more information"));
+            engine.sink.warn(warning!(
+                span,
+                "image contains foreign object";
+                hint: "SVG images with foreign objects might render incorrectly in typst";
+                hint: "see https://github.com/typst/typst/issues/1421 for more information"
+            ));
         }
     }
 
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -198,8 +198,7 @@ fn layout_image(
         if data
             .as_slice()
             .windows(needle.len())
-            .position(|s| s == needle)
-            .is_some()
+            .any(|s| s == needle)
         {
             engine.sink.warn(warning!(span, "image contains foreign object";
             hint: "svg images with foreign objects might render incorrectly in typst";
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -191,7 +191,8 @@ fn layout_image(
         Smart::Auto => determine_format(elem.path().as_str(), data).at(span)?,
     };
 
-    // Warn the user if the image contains a foreign object.
+    // Warn the user if the image contains a foreign object. Not perfect
+    // because the svg could also be encoded, but that's an edge case.
     if format == ImageFormat::Vector(VectorFormat::Svg) {
         let needle = b"foreignObject";
         if data
--- a/crates/typst/src/visualize/image/mod.rs
+++ b/crates/typst/src/visualize/image/mod.rs
@@ -194,7 +194,7 @@ fn layout_image(
     // Warn the user if the image contains a foreign object. Not perfect
     // because the svg could also be encoded, but that's an edge case.
     if format == ImageFormat::Vector(VectorFormat::Svg) {
-        let needle = b"foreignObject";
+        let needle = b"<foreignObject";
         if data
             .as_slice()
             .windows(needle.len())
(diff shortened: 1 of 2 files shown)


crates/typst/src/visualize/image/mod.rs
```diff
@@ -190,6 +191,21 @@ fn layout_image(
...
     };
 
+    // Warn the user if the image contains a foreign object. Not perfect
+    // because the svg could also be encoded, but that's an edge case.
+    if format == ImageFormat::Vector(VectorFormat::Svg) {
+        let needle = b"<foreignObject";
+        if data
+            .as_slice()
+            .windows(needle.len())
+            .any(|s| s == needle)
```
laurmaedje, line 201: This seems potentially rather slow. I think `std::str::from_utf8` + `contains` might actually be faster because `str::contains` uses an optimized string search algorithm. We could also use https://docs.rs/memmem but it feels like overkill to pull a crate.

crates/typst/src/visualize/image/mod.rs
```diff
@@ -190,6 +191,21 @@ fn layout_image(
...
+        {
+            engine.sink.warn(warning!(span, "image contains foreign object";
+            hint: "svg images with foreign objects might render incorrectly in typst";
+            hint: "see https://github.com/typst/typst/issues/1421 for more information"));
```
laurmaedje, line 205:
```suggestion
            engine.sink.warn(warning!(
                span,
                "image contains foreign object";
                hint: "SVG images with foreign objects might render incorrectly in typst";
                hint: "see https://github.com/typst/typst/issues/1421 for more information"
            ));
```
looks like svg is uppercase in existing messages, so let's use that. other than that, lowercase is good!

crates/typst/src/visualize/image/mod.rs
```diff
@@ -30,6 +30,7 @@ use crate::syntax::{Span, Spanned};
 use crate::text::{families, LocalName};
 use crate::utils::LazyHash;
 use crate::visualize::Path;
+use crate::warning;
```
laurmaedje, line 33: not sure why this even works, but it should be imported from `crate::diag`

crates/typst/src/visualize/image/mod.rs
laurmaedje: We could have `Readable::as_str -> Option<&str>` to do this. Skips the UTF-8 check if its already a string.

LaurenzV: Yes, I had the same thought, will change that.

crates/typst/src/visualize/image/mod.rs
```diff
@@ -190,6 +190,22 @@ fn layout_image(
         Smart::Auto => determine_format(elem.path().as_str(), data).at(span)?,
     };
 
+    // Warn the user if the image contains a foreign object. Not perfect
+    // because the svg could also be encoded, but that's an edge case.
+    if format == ImageFormat::Vector(VectorFormat::Svg) {
+        let has_foreign_object =
+            data.as_str().map(|s| s.contains("<foreignObject")).unwrap_or(false);
```
laurmaedje, line 197:
```suggestion
            data.as_str().is_some_and(|s| s.contains("<foreignObject"));
```
sorry, but my refactoring pattern matching brain can never stop 🙈

laurmaedje:
Thank you!

crates/typst/src/visualize/image/mod.rs
```diff
@@ -190,6 +190,22 @@ fn layout_image(
         Smart::Auto => determine_format(elem.path().as_str(), data).at(span)?,
     };
 
+    // Warn the user if the image contains a foreign object. Not perfect
+    // because the svg could also be encoded, but that's an edge case.
+    if format == ImageFormat::Vector(VectorFormat::Svg) {
+        let has_foreign_object =
+            data.as_str().is_some_and(|s| s.contains("<foreignObject"));
+
+        if has_foreign_object {
```
Myriad-Dreamin, line 199: This might be not an issue when exporting to svg.

LaurenzV: Fair enough, but since we don't have export-specific lints yet, that's the only thing we can do. And it says _might render_ incorrectly, anyway.

Myriad-Dreamin: I mean, for example, people using touying usually embed gifs by this approach but they will get an warning that cannot escape.

LaurenzV: Yeah, that's annoying... But hopefully it won't be an issue once we have warning surpression.

PR #4935, opened 2024-09-11
Add Unicode 16 emoji

ssotoen:
This PR adds the new emoji included in Unicode 16.0.

**Breaking change:**
`face.tired` is renamed to `face.distress`.
1FAE9 face with bags under eyes is assigned `face.tired`.


crates/typst/src/symbols/emoji.rs
```diff
@@ -421,6 +422,7 @@ const EMOJI: &[(&str, Symbol)] = symbols! {
         angry.red: '😡',
         anguish: '😧',
         astonish: '😲',
+        bags: '🫩',
```
reknih, line 425: I am not sure about `face.bags`. It does not transport the bags under the eyes well and a longer name would be impractical.

Instead, let me recommend the following:

- Give the 🫩 U+1FAE9 emoji the alterator "tired". The [proposal](https://www.unicode.org/L2/L2023/23260-eye-bags-face-emoji.pdf) supports this and lists "eye bags, exhausted, tired, hungover, sleepless, defeated" as CLDR keywords.

- This creates a collision with 😫 U+1F62B which is currently `face.tired`. I propose to rename this to `face.annoy` or `face.distress`. [Emojipedia has this to say](https://emojipedia.org/tired-face#emoji) about the Emoji:

  > While intended to represent tiredness, it commonly conveys various degrees and tones of frustration and sadness as well excitement and affection, as if it just can’t handle how great someone or something is.
  
  While the original Japanese carrier designs may convey tiredness, the industry converged on the modern design after the release of iOS 5. I know that we may break users who rely on `face.tired` to yield this specific design, but all users who use `face.tired` semantically arguably receive an emoji with U+1FAE9 that represents that better.
  
  The `distress` alternator would align with Apple's name for the emoji (Distraught Face).
  
  2003 KDDI U+1F62B design: KDDI tired emoji 2003, Apple iOS 5.1 U+1F62B design (2012): iOS 5 tired emoji 2012


An alternative that avoids the collision would be naming U+1FAE9 `face.exhaust` or `face.hungover`

reknih (commented):
Codepoints look good, a naming comment

crates/typst/src/symbols/emoji.rs
ssotoen: Yeah, I wasn’t too happy with the name myself. I would have called it `face.tired` if that wasn’t already in use. Another option I considered was `face.eyebags`.

In this case I think the breaking change is worth it.

laurmaedje: I agree as well.

laurmaedje:
Thanks!

PR #4939, opened 2024-09-11
Add `or`, `and`, and `not` symbol variants

MDLC01:
This PR adds variants to `or`, `and` and `not` symbols. It also adds `eq.and` and `eq.or`.

I'm not sure whether the names `and.bar` and `or.bar` are good. This is inconsistent with `nothing.bar`, added in #4826, where the bar is added above the base symbol. Given the existence of U+2A1B ⨛ INTEGRAL WITH OVERBAR and U+2A1C ⨜ INTEGRAL WITH UNDERBAR (currently unnamed in Typst, but which will require differentiating between adding a bar above or below), I now think we should be explicit and use `abovebar` and `underbar`. This means:
- using `and.underbar` and `or.underbar` instead of `and.bar` and `or.bar` in this PR;
- renaming `emptyset.bar` and `nothing.bar` to `emptyset.overbar` and `emptyset.overbar`.[^1]

To the extent of my knowledge, only two similar symbols are missing: U+2A51 ⩑ LOGICAL AND WITH DOT ABOVE and U+2A52 ⩒ LOGICAL OR WITH DOT ABOVE. The reason is that the names `and.dot` and `or.dot` are already used for U+27D1 ⟑ AND WITH DOT and U+27C7 ⟇ OR WITH DOT INSIDE. Maybe the names `and.abovedot` and `or.abovedot` could be used? Or `and.dot.above` and `or.dot.above`, since `and.above` and `or.above` don't mean anything.

Finally, I want to note that `or.bar` is named "VEE WITH UNDERBAR" in Unicode, instead of something like "LOGICAL OR WITH UNDERBAR." I think it is safe to assume the "vee" is a logical or symbol, given that the codepoint is part of Logical ands and ors sub-block of Supplemental Mathematical Operators, near U+2A5F ⩟ LOGICAL AND WITH UNDERBAR.

[^1]: This is not a breaking change, because #4826 was merged after the release of the latest release.


reknih:
I think that the `.bar` modifier should produce the _most frequently used_ bar and then, if there are multiple `symbol.bar.under`, `symbol.bar.over`, should be available.

MDLC01:
I like this solution. I will do that in the future and I will use a similar solution for `and.dot.above` and `or.dot.above` in this PR.

MDLC01:
Actually, do you prefer `*.under`/`*.over` or `*.below`/`*.above`? For `bar` and for `dot`.

reknih:
> Actually, do you prefer `*.under`/`*.over` or `*.below`/`*.above`? For `bar` and for `dot`.

Is there prior art for either?

MDLC01:
I don't think so. The only precedent I'm aware of is that when a symbol is composed of multiple symbols, the top-most or left-most symbol is used as the base name (for exemple, see `gt.lt` and `lt.gt`). But I wouldn't say this applies here, because conceptually we have a bar above a symbol, not a symbol below a bar.

MDLC01:
Although, now that I think about it, I added U+2A5E ⩞ LOGICAL AND WITH DOUBLE OVERBAR as `eq.and`, and similarly for U+2A62 ⩢ LOGICAL OR WITH DOUBLE OVERBAR.

laurmaedje:

MDLC01:
This was converted to Proposal 22 in the document.

PR #4953, opened 2024-09-13
Add music notation-related symbols

MDLC01:
This PR adds some music notation-related symbols from [Musical Symbols](https://unicode.org/charts/PDF/U1D100.pdf). I also made some changes to preexisting music symbols:

- **Breaking change.** I moved `rest.quarter` to the top of the list for `rest`, because this is the most well-known rest kind.

- **Breaking change.** I renamed `rest.multiple` to `rest.double`, because this specific symbol corresponds to a rest that lasts for a double whole note. With this in mind, I renamed `rest.multiple.measure` to `rest.multiple`.

𝄋 and 𝄌 are common and very characteristic music symbols with names that are unlikely to cause conflicts in the future, which is why I added them.

I'm not sure the `octave` family is really useful. I mostly added it because it is unlikely to conflict with anything in the future.

There are many remaining symbols in the Musical Symbols block. Most of them are meant to be combined with other symbols, or just aren't really useful, so I did not add them. ~I also did not add any symbol from mensural notation because this system is not used anymore.[^1]~ (see comment thread below)

[^1]: https://en.wikipedia.org/wiki/Mensural_notation


Enivex:
> I renamed rest.multiple to rest.whole, because this specific symbol corresponds to a rest that lasts for a double whole note

I think you mean `rest.double`.

MDLC01:
Indeed, thanks!

Enivex:
> I also did not add any symbol from mensural notation because this system is not used anymore

While I have no idea which of the symbols make sense to include, I don't think this is a good reason to exclude anything. I'm sure there's academic publishing in music theory or history where the notation is used, even if it's not used in modern notation.

Enivex:
> I think people who need to write about those symbols typically use a music notation software and lay out the symbol on a partition instead of alone within some text. But you are right that this may be useful. I'll try to add it as well.

It's presumably not very pressing. We can just wait until someone asks for them.

sitandr:
I'm not sure it is a good idea to add too many symbols right now, when there are no categories for symbols available in official docs. I know there are tools like "draw this symbol and it will show its name", but sometimes I will just go to the doc and will try to find the right symbol. But the more symbols there are, the harder it becomes.

I don't mind adding some musical notation, but adding just all symbols ever used seems redundant. Though mensural notation seems fun, I could even imagine it being used for math.


Nevertheless, I don't think symbols that are practically never used nowadays should be added to the scope _before there are some kinds of categories for them_. They can be relatively easily used from their unicode, can't they? I think for these extremely rare cases, this should be enough.

MDLC01:
> It's presumably not very pressing. We could just wait until someone asks for them.

I've already started working on it, so I'll do it. But I also realized Unicode supports "Gregorian notation" (which doesn't even have a Wikipedia page). That, I won't add for now.

Enivex:
> I'm not sure it is a good idea to add too many symbols right now, when there are no categories for symbols available in official docs. I know there are tools like "draw this symbol and it will show its name", but sometimes I will just go to the doc and will try to find the right symbol. But the more symbols there are, the harder it becomes.

This is a solvable issue. There is no reason the symbol list cannot have filters.

MDLC01:
Having categories in the symbol list would be great, indeed.

MDLC01:
I only added mensural note values for now. I did not add the mensuration signs because I fear this could clash with more important symbols we might want to add in the future. I also did not add the "croix" (U+1D1CF) because I can't figure out how this is used (there is no mention of it on the Wikipedia page for mensural notation).

laurmaedje:

PR #4954, opened 2024-09-13
Improve label binding

### LINKED ISSUE #1943: labels try harder to find a good target
### Description

Labels descend into `ShowElem`  and `SequenceElem` nodes to find a suitable target.

## Background
Typst has a collection of "blessed" element functions that fully participate in numbering and referencing. Because author-defined element functions are not available, the strategy for now is to co-opt figure elements with a custom `kind`.  However, because of the way labels attach to their targets, there are some very serious limitations on what can be done inside a function that is emulating a custom element function.  Consider:

```
#let foo(x) = {
  show figure.where(kind: "foo"): it => {
    let counter = counter(figure.where(kind: "foo"))
    let number = locate(loc => { 
        numbering(it.numbering, ..counter.at(loc)) 
    })
    [#number #it.body]
  }

  figure(x, kind: "foo", supplement: "Foo", numbering: "1")
}
```  

This kind of figure can't participate in the label/reference system:
```
#foo[Hello World]<howdy>

@howdy
```

generates

```
error: cannot reference styled
   ┌─ /play.typ:15:0
   │
15 │ @howdy
   │ ^^^^^^
```

The issue is that although a naive author might think that the function `foo` only generates a figure (which can be referenced) it actually generates a StyledElem that contains figure as a child:
```
styled(
  label: <howdy>,
  child: figure(
    body: [Hello World],
    kind: "foo",
    supplement: [Foo],
    numbering: "1",
),
.., )
```

The workaround would be to move the show rule outside of the function.  But if this `foo` object is being provided by a package, then the author using the package needs to not only include the package, but then also wire up the show rules.

As another example, consider the following.
```
#let foo(x) = {
  locate( loc=> {
    // Reset counter if needed
  })
  figure(x, kind: "foo", supplement: "Foo", numbering: "1")
}
```
The `locate` function here could be used to reset the figure's counter if, say, a level 1 heading occurred since the last foo figure.  But this use of `locate` also interferes with referencing the figure.  It generates "silent" content and therefore

```
#foo[Hello World]<howdy>
```
generates a `SequenceElem` with two children
```
sequence(
  label: <howdy>,
  children: (
    locate(func: (..) => ..),
    figure(
      body: [Hello World],
      kind: "foo",
      supplement: [Foo],
      numbering: "1",
), ),
)
```
The workaround here would be to try to use a show rule attached to the event that triggers a counter reset.  So we're back to authors needing to both include a package and wire up show rules to get their Foo figures to work.  Moreover, it forces fragmenting part of the code needed to implement `Foo` into a show rule for an unrelated object, which feels brittle.

Pretty much the only thing that `foo` can do while still making a figure that can be referenced is this:
```
#let foo(x, ..args) = {
  figure(x, /* Explicit settings here based on `args`*/)
}
```

## Proposal

Labels bind early to their targets during evaluation.  They seek backwards over recent content, skipping `Unlabelable` content (space and paragraph breaks) to find the content to be labeled.  The tentative proposal here is

1) During the seek procedure, if a `StyledElem` is encountered, the search descends to the element's child.
2) During the seek procedure, if a `SequenceElem` is encountered, the search descends into the sequence's children, starting with the last element.

I'm less than comfortable with the internals of the meta system, however, and am not sure what the consequences of this change would be, especially item 2.  And it could be that the real request here is to have a better way to create custom things that can be referenced and which have full control over their numbering.  See also #1896

mkorje:
Closes #1943.

When trying to bind a label to content, we now traverse into `SequenceElem`'s children and `StyledElem`'s child. Also removed `FieldAccessError::Internal` to fix "this is a bug" errors on unlabelled content when trying to access the `"label"` field (see the [discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1284193916534521857) for some discussion of this).

This all means that examples like the below (from #1943) will now work as "expected".
```typ
#let foo(x) = {
  show figure.where(kind: "foo"): it => {
    let counter = counter(figure.where(kind: "foo"))
    let number = counter.display(it.numbering)
    [#number #it.body]
  }

  figure(x, kind: "foo", supplement: "Foo", numbering: "1")
}

#foo[Hello World]<howdyy>

@howdyy
```

```typ
#let foo(x) = {
  context {
    // Reset counter if needed
  }
  figure(x, kind: "foo", supplement: "Foo", numbering: "1")
}

#foo[Hello World]<howdy>

@howdyy
```

Potential downside is that you can no longer attach a label to a `SequenceElem` or a `StyledElem`.


PgBiel:
> Potential downside is that you can no longer attach a label to a `SequenceElem` or a `StyledElem`.

Does this imply that `[#figure(...) abc def g] <label>` will label the figure rather than the combination of the figure and the text?

mkorje:
> Does this imply that `#[#figure(...) abc def g] <label>` will label the figure rather than the combination of the figure and the text?

No, it will label the text `g`.

> And what about `#[abc #figure(...) def g] <label>`?

And the same happens here, it will label the text `g`.

The search for content to label still seeks backwards within a sequence's children.

crates/typst/src/foundations/content.rs
```diff
@@ -374,6 +374,16 @@ impl Content {
         Self::sequence(std::iter::repeat_with(|| self.clone()).take(count))
     }
 
+    /// Retrive styled content's child (the content alone).
+    pub fn styled_child(&mut self) -> &mut Content {
+        // Awaiting Polonius...
+        if self.is::<StyledElem>() {
+            let styled_elem = self.to_packed_mut::<StyledElem>().unwrap();
```
PgBiel, line 381: I assume destructuring a call to `to()` lead to some problem with lifetimes?

PgBiel (commented):
Overall I feel a bit unsure about this change at first. There definitely needs to be some improvement, as the problem is pretty jarring in the few cases where it manifests. But I do wonder to which extent this should go:

- I feel like peeking into StyleElem is generally fair, but I do wonder if this could cause problems when querying and perhaps placing the element again later, since the local styles could potentially be lost.

- For sequences, I think this might lead to the most false-positives overall. If we do this change, there should definitely be some way, such as a function, to label the sequence as a whole. But I'd still be cautious here as I think the peeking behavior could be unintended more often than not here.

crates/typst/src/foundations/content.rs
mkorje: Yep.

mkorje:
You're right, there is then now the issue with StyledElem when queried and placed again losing its styles. I think this could then be resolved by not peeking into StyledElem when labeling, and instead dealing with its child when referencing. Though this may not be ideal either if its child is a sequence, for example.

Sequences are definitely a lot trickier. I agree that there should be a way to choose whether to label the sequence vs an element inside of it. But I imagine any sort of user interface for this would become difficult in nested sequences. In my opinion, it's more helpful to traverse into a sequence by default when labeling, rather than labeling the sequence itself.

laurmaedje:
For styles, I can somewhat see this as it's less evident to a user that styles result in an element in the first place. For sequences, I can see it less because there are multiple elements and it's not clear to me why the last element is special compared to the others.

laurmaedje:
Let's close this for now and have more discussion first and then maybe make a change later. Thanks still!

Issue #3391, opened 2024-02-09
Highlight Offset Mismatch with Subscript and Superscript Text


f3fora:
### Description

As you can see in the following example, none of the combinations of `#sub` or `#super` with `#highlight` produces the expected results.  In particular the offset of the highlight do not match the offset of superscript or subscript text.

### Reproduction URL

https://typst.app/project/p1D75-gUzR6x9HluVb1PmI

### Typst version

- [X] I am using the latest version of Typst

owiecc:
@f3fora please make the project public.

f3fora:
Sorry! I have updated the link.
`https://typst.app/project/r4EHGvxFzw0JlWTqH_zXm_`

A-Walrus:
What should be the correct behavior? Should the line be highlighted at its regular height, or should the highlight stick to the sub/superscript?

f3fora:
The same wrong behavior is share between all the decorators, i.e. `highlight` `strike`, `overline`, `underline`.

It seems that a wrong shift is applied to this decoration.

I have updated the bug report project.

https://github.com/typst/typst/blob/a2f91c4bc6c28f05d05b896427f8e2caef3bbf5b/crates/typst/src/layout/inline/shaping.rs#L316-L329

https://github.com/typst/typst/blob/a2f91c4bc6c28f05d05b896427f8e2caef3bbf5b/crates/typst/src/text/deco.rs#L411-L535

PR #3941, opened 2024-04-16
Support Line decoration for Math [1/2] : Support Highlight for inline and block math equations

LuxxxLucy:
This is the first of two PR for solving #2200

Regarding #2200, it turns out the highlight is supported for normal text but not for math equations. Digging into it, there are two aspect of it:
1. For block math equations, we just get the frame and then add line based decorations to it.
2. For inline math equations, this becomes a little bit tricky: we have a list of MathParItem consisting of actual Frame or Space, we will need to compute the smallest box that covers the list of all MathParItems.

**Note 1**
for line based decorations, highlight is a rectangle added to the background of the targeted frame (be it normal text or math equations), so we "prepend" it (adding it beneath the frame), but for other line based decorations, it is **ON TOP OF" the targeted frame. It becomes kind of tricky and thus will be tackled in a next PR

**Note 2**
A new decoration function is introduced in this PR; the end goal is to merge the new one with the exiting old one. Such refactoring will be included in next PR.

## Scope

what is in this MR:
1. support of highlight over inline math equations
2. support of highlight over block math equations
3. corresponding test typ/png

what is not in this MR:
1. support of other line-based decorations.

What will be in the second PR
1. support of other line based decorations.
2. refactor the old "decorate" function and the newly added function introduced in this MR,

PR #4959, opened 2024-09-14
fix(highlight): introduce hint-size, adjust offset of rectangular

f3fora:
Address #3391

The goal of this PR is to correct the calculation of the highlight box's offset when working with text that has a non-zero baseline.

- **New Parameter: `hint-size`**
  - The default value is `auto`.
  - When set to `auto`, the highlight box is aligned with the text's baseline. For instance, if the text is superscript, the box will be offset accordingly, and its height will be calculated based on the font size.
  - When a specific size is provided for `hint-size`, the highlight box will have that vertical size and will be aligned such that its bottom edge meets the "zero" baseline.

Does this approach seem reasonable? Also, would you suggest a better name than `hint-size` for this parameter?

An analogous PR or future commit should address the offset issue for other types of decorations.


laurmaedje:
I think it's possible to just fix the bug without a new parameter. This was partly attempted in https://github.com/typst/typst/pull/3941 and https://github.com/typst/typst/pull/3953. You can read up on the attempts there. It wasn't quite there yet, but it can definitely be fixed and I think that would be a better user experience than a new parameter. Thanks still!

PR #4964, opened 2024-09-15
Fix spacing after bibliography heading

### LINKED ISSUE #4963: Heading spacing ignored in bibliography
### Description

When setting the space below a heading, it is ignored for the bibliography's heading.
It works fine for normal headings (even with a grid as content, as it is the case with the bibliography) and also in the outline.

*This is a regression, it works fine in 0.11.1, but does not on main (tried with db71a178bef7f1525d732a190ac75a1a6d56f24b)*

**Example:**

<table>
<tr>
<td>

```typ
#show heading: set block(below: 1cm)

#outline()

= Normal Heading
#grid(
  columns: (auto, 1fr),
  column-gutter: 5mm,
  [Reference], [@test]
)

#bibliography("bibliography.yaml")

```

</td>
<td>

</td>
</tr>
</table>

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

EpicEricEE:
Fixes #4963 

The row gutter should only be applied in between bibliography entries, but not between the heading and the first entry, as that's where normal block spacing should apply.


PgBiel:
In principle, from a quick look, the row gutter here seems to be equal to `par.spacing`, though I think that's not used after normal headings, right? If so, then this change should be ok.

PgBiel (approved):

laurmaedje:
Thanks!

Issue #2860, opened 2023-12-04
Prime symbol broken with Libertinus Math


schuelermine:
### Description

The prime symbol seems incorrectly set in superscript when using Libertinus Math.

```typ
#show math.equation: set text(font: "Libertinus Math")

$
f(x) &= x^2 + 1 \
f'(x) &= 2 x
$
```

This happens in both the Web App and the command-line interface.

Web App:
![Screenshot 2023-12-04 at 21-38-53 scratchpad – Typst](https://github.com/typst/typst/assets/48802534/dc8881e4-135e-453a-b714-482ea721be51)

Local:
![Screenshot from 2023-12-04 21-38-27](https://github.com/typst/typst/assets/48802534/dc92967e-0a05-4ba0-b97c-4b5d016d5657)

### Operating system

Web app, Linux

### Typst version

- [X] I am using the latest version of Typst

alerque:
This is probably a font issue not a typesetting engine one. Whether it is a bug or a style choice I'm not actually sure. I'm the maintainer of Libertinus fonts project but was not the original author of the Math companion font. Either way please feel free to file an issue over [here](https://github.com/alerque/libertinus).

schuelermine:
I am fairly sure this is a typst-specific issue. I have recreated this equation in MathML and it looks fine.
![image](https://github.com/typst/typst/assets/48802534/0519f4a0-fa0d-447a-9c46-0a80dfb71e56)

Perhaps this is a version difference?

alerque:
Can you post the MathML version here so I can experiment with this in SILE?

schuelermine:
I made the MathML with MathJax and got the CSS link from MDN/Frédéric Wang
[MathMLLibertinus.txt](https://github.com/typst/typst/files/13560059/MathMLLibertinus.txt)

schuelermine:
However I opened it with a Chromium browser and the same issue appears.
![Screenshot_20231205-133254~2](https://github.com/typst/typst/assets/48802534/bd88a8e6-c6ae-4d3c-b902-141a2db84240)

alerque:
I'm not much of a math expert, but I'm pretty sure that MathML is kindo of borked. In any case I think I got a working example in SILE's math variant syntax (an alternative to having it work on MathML directly). The result is that *Libertinus Math* looks differently there than in Typst. Here are my inputs comparing *STIX Two Math* with *Libertinus Math* in both SILE and Typst:

<summary>
SILE source (<code>prime.sil</code>):
[...]

PR #4969, opened 2024-09-16
Change default font to Libertinus Serif

### LINKED ISSUE #1257: Replace "Linux Libertine" with the more up to date fork "Libertinus Serif"
> Over time it grew into to a full-fledged fork addressing many of the bugs in the Libertine fonts.

https://github.com/alerque/libertinus

(Note that the web app already has the math font Libertinus Math)

laurmaedje:
Closes #1257.

While we could have kept Linux Libertine for the test suite, I prefer testing with the real default font. This updates ~800 reference images. This doesn't make me particularly happy, but the repository bloat shouldn't be much more than 2MB due to the work done so far in keeping reference images small.

Unresolved questions:
- [x] Should we bundle the semibold variants? Linux Libertine didn't have those. I've opted for yes now.
- [x] Do we need some kind of migration period?
  - [x] Should we keep bundling the old font for one more release? I personally feel like this is not really necessary. Layout shifts should be non-existent or very minor since the fonts are mostly metrically compatible. I suspect that most users won't notice a thing. 
  - [x] Adding a hint when the font was explicitly set to "Linux Libertine"

Todo:
- [x] Update NOTICE file in typst-assets
- [x] Merge https://github.com/typst/typst-assets/tree/libertinus-serif
- [ ] Add Libertinus Serif to the web app


alerque:
I assume you're working with Libertinus 7.040? There are a few minor updates (e.g. I_J kerning for Dutch locales) since the last tagged release. If desired I could look into cutting a new release with that in place before this work merges just so the loop is as closed as it can be.

laurmaedje:
That would be great! I indeed used the latest released the version.

alerque:
After a few years without a release there were some problems with the font tooling (changes in python-fonttools meant it wouldn't build against current versions) but I think I've got that worked out so I *could* cut a release at any time. That being said there are a couple PRs that look valuable and may be ready to go it, but it is 3 am and I can't do more tonight. What's your timeline on this? Should I cut a release so you can start testing it or do I have a couple days to review PRs and make a release with any of the pending contributions that seem ready for production?

laurmaedje:
A few days are no problem. I've got some other stuff to do as well. I just wanted to get the ball rolling here and potentially get some feedback on the unresolved questions.

ssotoen:
> Should we bundle the semibold variants? Linux Libertine didn't have those. I've opted for yes now.

Bundle all available weights.

> Do we need some kind of migration period?

I would keep Libertine for 0.12 and emit a compiler hint when the font is explicitly set. Then drop it for the next release.

laurmaedje:
> I would keep Libertine for 0.12 and emit a compiler hint when the font is explicitly set.

Some people might have Linux Libertine installed and if they want to use it, who am I to stop them. So a blanket warning seems problematic. That's why I think adding a hint to the "unknown font family" warning specifically for Linux Libertine is better, but it means removing the font now.

alerque:
> emit a compiler hint when the font is explicitly set.

Wouldn't warning when it is **not** explicitly set make more sense?

In any case some transition time where both fonts are available so that people have a low friction way not to reflow anything is likely to be appreciated by a few folks. Even if Libertinus Serif is *better* in every way (of course I think it is) there are some differences in metrics to some glyphs that will cause many documents to reflow. For most people it probably won't matter at all, but there is bound to be a few documents where people are committed to exact metrics that do matter.

laurmaedje:
> Wouldn't warning when it is not explicitly set make more sense?

Then, it would warn on almost every document, with no good way to get rid of the warning. I think that's rather something that should just be stated prominently in the changelog.

My proposal was to warn only when it's explicitly set _and not available on the system._ The most likely cause of that situation is someone hardcoding the default font for some reason (maybe they needed to go back to the default within an equation) and they should likely update it to the new default.

We can also keep bundling it for another version, but then I see no opportunity for a warning that doesn't trigger on pretty much everything.

laurmaedje:
> For most people it probably won't matter at all, but there is bound to be a few documents where people are committed to exact metrics that do matter.

We've tuned hyphenation, so 0.12 will probably reflow lots of documents anyway... And there are a few other breaking changes.

Andrew15-5:
> *  Should we bundle the semibold variants? Linux Libertine didn't have those. I've opted for yes now.

Are you concerned about additional binary size, or just a plain question? I personally only use normal/bold/italic/bold+italic 99.9% of the time. I almost never had to tweak the weight parameter. But I think it's a safe bet that some people would do that, and therefore it's better to pack all variants in, than have an incomplete bundle which can lead to issues like "add x variant to the default font".

>  *  Do we need some kind of migration period?

To be on the safe side, yeah. One (minor) version is probably enough?

>    *  Should we keep bundling the old font for one more release? I personally feel like this is not really necessary. Layout shifts should be non-existent or very minor since the fonts are mostly metrically compatible. I suspect that most users won't notice a thing.

I think this is the same question as the above, so yes, I think we should. Hard breaking changes (with no deprecations) is probably never nice to see as a user.

>    *  Adding a hint when the font was explicitly set to "Linux Libertine", but that font isn't available could be a good idea.

If it was explicitly set and is present on the system (if you can differentiate between system and built-in fonts), then I guess it's fine. If however the font is not present on the system, then a warning/hint should definitely be added ("the default font has changed, the old one is deprecated, and it's not found on the system, consider installing or changing to the new font", something like that). Either way, a warning in the changelog (or somewhere else as well?) should be enough for those who care (which font is the default one).

Enivex:
I really don't see the point of a migration period. There are already breaking changes in 0.12, and the font is almost identical to the one we're already using. I'd wager the vast majority won't even notice that the font changed.

ssotoen:
Should we also bundle Libertinus Math?

Enivex:
> Should we also bundle Libertinus Math?

I don't think there's any plan to change the default math font.

ssotoen:
I know there’s no plan, that’s why I’m suggesting it. Linux Libertine doesn’t have a math font so it’s understandable that the team went for New CM; but Libertinus does, and this would be a good opportunity to make the default look more consistent.

Enivex:
> I know there’s no plan, that’s why I’m suggesting it. Linux Libertine doesn’t have a math font so it’s understandable that the team went for New CM; but Libertinus does, and this would be a good opportunity to make the default look more consistent.

I don't think the math font is mature enough to be a default. See for instance https://github.com/typst/typst/issues/2860

alerque:
@laurmaedje Is the prior perceived need for a [TTF variant](https://github.com/typst/typst/issues/1257#issuecomment-1827675063) resolved? (I have handled all outstanding PRs for Libertinus and am tinkering with release tooling). I can build TTFs with a one line config change, but I can't vouch for how *good* they would be. We've been releasing OTF only for a while and we know those builds are pretty robust, but any TTF builds I put out wouldn't have had much exercise. They could be hinted with `ttfautohint` (or not at all) but there would not any manual hinting or tweaking. From my perspective that isn't any "value added" over OTF and just letting the device do the hinting. But if there are other reasons to consider we could check it out.

alerque:
I just shipped Libertinus v7.050. It's already deployed to all Arch Linux users via [otf-libertinus](https://archlinux.org/packages/extra/any/otf-libertinus/) and I've submitted bumps for the [Homebrew Cask](https://github.com/Homebrew/homebrew-cask/pull/186163) and [CTAN package](https://ctan.org/pkg/libertinus-fonts). The latter two haven't landed yet but usually do so within a few hours. I'm going to bed now but if my inbox hasn't exploded in the next day or so then the OTF release at least should be good to go.

laurmaedje:
Sounds great, thank you! There is no need for TTF anymore, the printer issues and suboptimal subsetting are fixed thanks to @LaurenzV.

alerque:
c.f. https://github.com/typst/typst-assets/pull/3

knuesel:
What about bundling Libertinus Sans? Currently the only option for a sans-serif default that should work everywhere is to use the mono font. Having another option would be nice for slide themes and many other documents (for example popular styles like KOMA-Script use sans serif for titles).

Maybe another sans serif font would be better though...

Enivex:
A template opinionated enough to want a sans serif font will probably want a specific one, and not just the one type would hypothetically provide.

Starting to bundle a ton of fonts feels like a mistake.

knuesel:
Sure, but bundling one sans serif font could be a good middle-ground between "just one text font" and "a ton of fonts".

Mixing a serif and a sans-serif font is a common technique to make a design more dynamic. The contrast is often more important than the precise choice of font. Even opinionated templates could use a standard sans-serif as fallback that works everywhere.

For example, I'm working on some slide themes that use a two-font design. What should I use as fallback that looks OK if not great? Will it still make clear what part is in what font in case the user wants to change one font?

laurmaedje:
The point of the built-in fonts wasn't really to ship a sensible base collection of fonts. It was to ship exactly the fonts Typst uses by default for text, math, and code, such that the defaults work out of the box.

I agree that we have a problem with reproducibility and fonts (it's being discussed at various times in the #Fonts forge on discord), but I think the solution isn't to bundle more fonts, but rather to find a proper way to distribute a larger collection of fonts.

Andrew15-5:
```typ
#import "@font/fira-code-nerd-font"
#show raw: set text(font: "Fira Code Nerd Font")
```

knuesel:
@laurmaedje OK that makes sense.

laurmaedje:
```
warning: Typst's default font has changed from Linux Libertine to its successor Libertinus Serif
  ┌─ hi.typ:1:16
  │
1 │ #set text(font: "Linux Libertine")
  │                 ^^^^^^^^^^^^^^^^^
  │
  = hint: please set the font to `"Libertinus Serif"` instead
  = hint: Linux Libertine is available on your system - you can ignore this warning if you are sure you want to use it
  = hint: this warning will be removed in Typst 0.13
```

Issue #4035, opened 2024-04-29
Downgrade missing references from errors to warning


m-krastev:
### Description

Coming from LaTeX, you could compile a document even if a reference is not defined. This would trigger a warning and a "??" would be printed on the resulting file.

Currently, Typst just errs and exits the whole compiler process (e.g. if using `typst watch`). 

### Use Case

A similar functionality where the error is ignored and a warning is emitted instead would help creative flow while writing so that one does not have to immediately go back to a missing reference and try to address it.

Enivex:
See https://github.com/typst/typst/issues/1322

m-krastev:
I would love for the functionality of #1276 to be included in the base compiler, as this should not be considered as a critical compiler failure.

laurmaedje:
If we can come up with a good design, I'm open to it.

Coekjan:
Maybe we can add a cli argument (like `--allow-missing-ref`, or something else) to toggle the compile/watch behavior:
1. normal (current) behavior: deny missing ref
2. `--allow-missing-ref` behavior: warn on missing ref, and print `?` (or something else, or something configurable by cli args) in the result

LaurenzV:
I think instead of having a specific flags just for refs we should instead think about how to implement a more generalized system for enabling/disabling warnings, then this could be extended to this specific use case as well.

geajack:
Yes! You want to be able to just put the references in as you're writing and add the actual reference targets at the end. An `--ignore-refs` flag or something would be ideal.

Eismandel:
Thank you for your solution. As I tend to overlook such messages, I added some additional code 

```
#let MissingRef(body, small_text: 5pt) = {
      let body = if body == [] {
        [Missing Ref: XXX]
      } else {
        body
      }
      set text(fill: white, size: small_text, weight: "bold")
      text()[~~]
      box(
        fill: rgb("#e046d3"),
        outset: 4pt,
        radius: 3pt,
        stroke: luma(1),
        [#body #place()[    
          #set text(size: 0pt)    
          #figure(kind: "MissingRef", supplement: "", caption: body, []),
        ]])
      text()[~~]
}
```
With the #show rule


```
#show ref: it => {
  if query(it.target).len() > 0 {
    it
  } else {
    MissingRef[Ref Missing #str(it.target)]

  }
}
```

With this set, I can create a new outline only displayed when there are some missing refs

```
#context {
  let elems = query(
    selector(figure.where(kind: "MissingRef")),
   )
[...]

PR #4974, opened 2024-09-17
Downgrade missing references and citations from errors to warning

laurmaedje:
Closes https://github.com/typst/typst/issues/4035

Both unresolved references and citations now result in warnings rather than errors. In the document, they are rendered in red, raw text.

```typ
// Warning: label `<earth>` does not exist in the document
Earth is a planet. @earth
```

image

Notably, references to duplicate labels are still an error.

Given this change, is becomes even more important that we prevent warning spam, so that important warnings like this one stand out. There are some ideas to have different importance for various warnings (for instance, by demoting less important warnings that do not have adverse affects on the document to a "lint" level), but that's still in-flux.


PgBiel:
I feel like adding warning levels (or equivalent) first is more important (or, at least, should ship together), considering the following:

1. I don't think the majority of users would want to force missing references to be displayed (would rather be notified about it and stop compilation), which implies this feature should be opt-in instead of opt-out in some way (e.g. a confirmation in the web app, a flag in the CLI, and so on);

2. With the upcoming deprecation warnings and others, warning spam will invariably increase in the next update, making this one easy to miss, and I'd say missing references are a larger problem for a document and something that the user should certainly at least be made aware of, much like other errors.

Imho we should tackle this first before merging this feature. I don't think it is urgent enough to skip this discussion.

Enivex:
> 1. I don't think the majority of users would want to force missing references to be displayed (would rather be notified about it and stop compilation), which implies this feature should be opt-in instead of opt-out in some way (e.g. a confirmation in the web app, a flag in the CLI, and so on);

I have the exact opposite impression. The majority would rather have references be a warning.

PgBiel:
Thanks for the input; I'll respond in the issue to keep the discussion organized.

PR #4988, opened 2024-09-20
Single thread support

Dherse:
Actually support single thread operations by using rayon, thanks for the hints from @SkiFire13


laurmaedje:
Thanks!

PR #4989, opened 2024-09-20
Add float to bytes conversion and vice-versa

T1mVo:
Adds float to bytes conversion and vice-versa. Useful when using plugins.

**Example:**

```typst
#float.from-bytes(bytes((0, 0, 0, 0, 0, 0, 240, 63)))
// 1.0
#array(1.0.to-bytes())
// (0, 0, 0, 0, 0, 0, 240, 63)
```

I followed the implementation of #4490.


laurmaedje:
Thanks!

PR #4992, opened 2024-09-20
Fix smartquotes for Swiss French

knuesel:
Swiss French uses thin spaces between quote characters and inner text.


laurmaedje:
Thanks!

Raphael-CV:
In fact, the narrow non-breaking space (`\u{202f}`) is also commonly used between quotation marks in France.
It seems to me that this is babel-french's default way of doing things.

For the moment, I write at the beginning of each of my documents
```js
set smartquote(quotes: ("«\u{202f}", "\u{202f}»"))
```

- https://fr.wikipedia.org/wiki/Guillemet#Guillemets_en_chevrons
- https://en.wikipedia.org/wiki/Non-breaking_space#Width_variation
- https://www.ctan.org/pkg/babel-french

PR #5003, opened 2024-09-22
Improve scripting reference `Methods` example

omniwrench:
The [Methods](https://typst.app/docs/reference/scripting/#methods) section talks about a **method call** that calls a function scoped to a value's type. It's structure is `value.method(..args)` and the equivalent full function call is `type(value).method(value, ..args)`. Setting the value name to `array` in this context is unnecessarily confusing as that is the same name as it's type, `array`.

This change makes clear that a method call is being used here.


laurmaedje:
True!

PR #5005, opened 2024-09-22
Update NewCM fonts to version 6.0.0

### LINKED ISSUE #3320: Bad baseline of New Computer Modern 08 (Book weight)
### Description

Using latest Typst (15474759), and New Computer Modern Version 5.1 downloaded from CTAN:

```typst
#table(columns: 2,
  text(font: "New Computer Modern 08", weight: 450, [abc]) + " a",
  text(font: "New Computer Modern 08", weight: 450, [abc]))
```
will produce:

图片

The base line in the second cell is clearly worng.

The bug can also be reproduced using latest release 0.10, with a little change on font name.

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

### LINKED ISSUE #4668: `lang: "de"` breaks ligatures with `font: "New Computer Modern"`
### Description

```
#set text(
  lang: "de",
  font: "New Computer Modern"
)
Definition
```

No fi-Ligature is shown in pdf :(

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

mkorje:
I don't know if we should bundle the new Bold Math font. I've included it for now.

I've also added the book weights for the normal font; it makes sense to me that they are bundled since the math book weight is. Though I'm not sure why they weren't previously.

Need to merge https://github.com/typst/typst-assets/pull/4 and then update this branch.

Updated test is due to the addition of another radical size in the font.

The changes that fix #4668 and #3320 are a part of this new version of the font as well.


Enivex:
Do we really need to bundle all these different weights? This directly translates into binary size.

It seems to me that reasonable default weights should be sufficient.

mkorje:
> Do we really need to bundle all these different weights? This directly translates into binary size.
> 
> It seems to me that reasonable default weights should be sufficient.

Probably not, but what constitutes "reasonable default weights"... The default weight is intended to be the book weight I think (it is the case for LaTeX at least)?

Ignoring binary size problems, it is nice to bundle the whole math family and normal font family. This would match the inclusion of the semibold weight in #4969, if that happens.

mkorje:
Also, regarding @laurmaedje's comment on #4969.

> The point of the built-in fonts wasn't really to ship a sensible base collection of fonts. It was to ship exactly the fonts Typst uses by default for text, math, and code, such that the defaults work out of the box.

Why aren't only the NewCM math fonts bundled? Why bundle the text fonts as well then.

laurmaedje:
If I recall correctly, someone talked my into it, in a similar conversation as the one in #4969. But I don't know anymore where.

mkorje:
Maybe we could remove the text fonts then, though it would be a breaking change.

laurmaedje:
Feels like a breaking change just for the sake of it to me. We can remove them once we have a proper solution for font distribution.

laurmaedje:
Discussed this with @reknih. Let's keep the text fonts but not add any more weights. We can remove them once there is a better solution for font distribution.

mkorje:
@laurmaedje Does that include not adding the bold math font? I'll update the the typst-assets PR then accordingly.

mkorje:
I've updated https://github.com/typst/typst-assets/pull/4

laurmaedje:
Thanks!

PR #5008, opened 2024-09-23
Better math argument parsing

### LINKED ISSUE #938: Argument spreading should work in math function calls
```
$mat(range(2).map(_ => range(2)))$
```
This just reprs the list which is expected, though a little confusing.

```
$mat(..#range(2).map(_ => range(2)))$
```
This does the same thing, but puts `..` in the matrix.

```
$mat(#..range(2).map(_ => range(2)))$
```

**This doesn't compile, but should IMO.**

```
$mat(#(..range(2).map(_ => range(2))))$
```

**This compiles, but doesn't work even though it should.**

---

```
$#math.mat(..range(2).map(_ => range(2)))$ \
$mat(#(..range(2).map(_ => range(2))))$ \
$mat(..#range(2).map(_ => range(2)))$
```

image

### LINKED ISSUE #2052: Named arguments in math mode should allow hyphens
### Description

When a named function argument contains a `-` in it's name, this argument can't be passed in math mode:

```typst
#let func(my-body: none) = my-body

$ func(my-body: a) $
```

Instead typst complains about `error: unknown variable: my`.
This is probably a parser bug.
I can see how in math mode the variable `my-variable` might need to be prefixed with a `#`, to not be interpreted as `my - variable`, however in the position of a function argument there should be no ambiguity, as function argument names can only be names, never expressions.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

mkorje:
This PR brings how math function call arguments are parsed more in line with how they are parsed in code mode.

New things:
- Closes #2052: Named arguments in math function calls can now be any identifier. This means it can contain hyphens and start with an underscore (as long as it isn't just an underscore), just like in code mode. For example, the below now works.
  ```typ
  #let func(my-body: none) = my-body

  $ func(my-body: a) $
  ```
- Closes #938: Argument spreading now works in math function calls. Prefixing two dots `..` to a variable does the same as in code mode. You can escape the dots or add a space after them to get `..` verbatim. See the example below.
  <table>
  <tr>
  <td> Code </td> <td> Image </td>
  </tr>
  <tr>
  <td>

  ```typ

  $
    // Spreading the argument, so works.
    mat(..#range(1, 5).chunks(2))
    // Passed verbatim, so reprs the list.
    mat(#(..range(2).map(_ => range(2))))
  $

  #let nums = ((1,) * 5).intersperse(0).chunks(3)
  $ mat(..nums, delim: "[") $
  ```

  </td>
  <td>

  
  math

  </td>
  </tr>
  </table>
- Duplicate named arguments error: The below now gives an error, instead of silently overriding the previous instance of the named argument given.
  ```typ
  #let func(my: none) = my
  // Error: 15-17 duplicate argument: my
  $ func(my: a, my: b) $
  ```

These changes don't conflict with #4364 majorly. Some things, like the new lex modes, might end up a bit cleaner with the changes from #4364 I think. Also, these changes are somewhat isolated; only a little (ish) has spread outside of math argument parsing to the rest of the math parsing code. Nothing else has changed with regards to math parsing (at least I hope so, all of the tests have passed which is promising). Only one test was updated to test `column-gap` and `row-gap` being used directly in matrices.


crates/typst-syntax/src/lexer.rs
```diff
@@ -919,7 +999,7 @@ pub fn is_id_start(c: char) -> bool {
 /// Whether a character can continue an identifier.
 #[inline]
 pub fn is_id_continue(c: char) -> bool {
-    is_xid_continue(c) || c == '_' || c == '-'
+    is_xid_continue(c) || c == '-'
```
PgBiel, line 1002: Does this not affect other parser code?

crates/typst-syntax/src/lexer.rs
```diff
@@ -29,6 +29,10 @@ pub(super) enum LexMode {
     Markup,
     /// Math atoms, operators, etc.
     Math,
+    /// An argument in a math function.
+    MathArg,
+    /// Function argument list in math.
+    MathArgs,
```
PgBiel, line 35: Following your comment in the PR description, I do think it would be ideal to use #4364's work on returning a SyntaxNode instead of creating new modes if possible. But we'll have to wait and see how that PR progresses.

PgBiel (commented):
Nice work!

I'll provide some initial thoughts (not a full review, just what stood out to me after an initial read).

crates/typst-syntax/src/lexer.rs
mkorje: I forgot to make a note of this, but '_' is already a continuing character in the Unicode Standard. (I used [this](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:XID_Continue=Yes:]), and it includes `_ U+005F LOW LINE`, so it isn't actually necessary I think). So no, it doesn't affect anything.

laurmaedje:
Looks nice from a quick glance. Let's get #4364 through first. I'd also rather wait with both until after 0.12 because they are fairly large and could contain regressions.

mkorje:
I've updated the PR so that it doesn't conflict anymore. I still need to look at not using new lex modes.

crates/typst-library/src/math/mod.rs
```diff
@@ -82,8 +82,8 @@ use crate::text::TextElem;
 /// - Within them, Typst is still in "math mode". Thus, you can write math
 ///   directly into them, but need to use hash syntax to pass code expressions
 ///   (except for strings, which are available in the math syntax).
-/// - They support positional and named arguments, but don't support trailing
-///   content blocks and argument spreading.
+/// - They support positional and named arguments, as well as argument
+///   spreading, but don't support trailing content blocks.
```
wrzian, line 86: At this point, "they don't support trailing content blocks" is probably best as its own bullet.

crates/typst-syntax/src/lexer.rs
```diff
@@ -687,6 +699,56 @@ impl Lexer<'_> {
...
+    fn math_args(&mut self, start: usize, c: char) -> (SyntaxKind, Option<SyntaxNode>) {
+        match c {
+            c if self.is_math_named_arg(start, c) => (SyntaxKind::Ident, None),
+            '.' if self.is_math_spread_arg() => (SyntaxKind::Dots, None),
+            _ => self.math(start, c),
+        }
+    }
+
+    /// Handle named arguments in math function call.
+    fn is_math_named_arg(&mut self, start: usize, c: char) -> bool {
```
wrzian, line 713: You can write this a lot shorterif you swap the boolean conditions  on the if-statements (e.g. `!is_id_start() -> is_id_start()`) and just `return true` in one spot.

wrzian (commented):
This is nice! I think overall the code is good. I appreciate the comments. but I do have one large issue: the lexer isn't turning off the math argument parsing after lexing the start of each argument.

Here are two examples to show what I mean:
```typ
#let func(x: none) = $ Gamma tack #x $
$
  func(x: sigma : f) \
  func(x: sigma:pi)
$
```
The first line is fine, but the second gives "Error: unexpected identifier" for `sigma` (it lexed as `Ident`, not `MathIdent`).

```typ
$ underbrace( 1..n ) $
```
Here we get "Error: unexpected dots".

The core issue is that we still lex the new argument kinds throughout the function call, but we shouldn't lex them unless we're at the _start_ of an argument. Otherwise the rest of the math infrastructure has to undo our incorrect lexing, which will not be trivial.

But I have an idea that is a bit hacky, but might work:

1. Remove the `MathArgs` lexer mode
2. Start parsing `Dot` as a syntaxkind in math (and keep `Comma`/`Semicolon`/etc.)
3. At the start of an argument:
    - If we're not at a `MathIdent`/`Dot`/`Underscore`, just call `math_exprs()` like normal.
    - If we are at `Dot`/`MathIdent`/`Underscore`, then we reach into the parser and use the lexer directly.
    - We add some methods to the lexer: `fn continue_named_arg(&mut self, start: usize) -> Option<SyntaxNode>`
    - Then we could write:
```rust
if let Some(named) = p.lexer.continue_named_arg(p.current_start) {
    let p.token.node = named;
    p.eat();
    // do other stuff
}
```

The key concept is that parsing named args and spread args isn't really a mode, it's just a single edge case that we handle at the start of each argument.

Though this would be quite hacky feeling code.

Thoughts?

crates/typst-syntax/src/parser.rs
```diff
@@ -398,6 +409,9 @@ fn math_delimited(p: &mut Parser) {
     while !p.at_set(syntax_set!(Dollar, End)) {
         if math_class(p.current_text()) == Some(MathClass::Closing) {
             p.wrap(m2, SyntaxKind::Math);
+            if p.at(SyntaxKind::RightParen) {
+                p.convert(SyntaxKind::Text);
+            }
```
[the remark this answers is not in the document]
mkorje: We actually can't just convert everything to text here. For example, we could be at the math shorthand `|]` (this is the only case though, afaik), in which case we shouldn't convert it to a `Text` kind.

mkorje:
Can't believe I overlooked that when removing the `MathArg` lex mode! I've had a go at implementing the hacky solution you suggested. (Note: we also need to check if we are at `Text` in addition to the others you mention, in case `a:` or `a-b:` is the named argument.) I don't think this feels too bad though.

I also added those cases you mentioned to the tests. And I checked that the AST is still unchanged when compared with main.

crates/typst-syntax/src/parser.rs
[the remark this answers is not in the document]
wrzian: ugh I always forget `|]`. Should definitely add as a comment then. I still don't fully like splitting p.convert_and_eat, especially because this is morally the same thing. Thoughts on just using `else { eat() }`?

tests/suite/math/mat.typ
```diff
@@ -54,13 +54,64 @@ $ a + mat(delim: #none, 1, 2; 3, 4) + b $
...
+
+--- math-mat-spread-expected-array-error-1 ---
+// Error: 13-14 expected array, found content
+$ mat(1, 2; 3, ..#range(0, 2)) $
+
+--- math-mat-spread-expected-array-error-2 ---
+#let nums = range(0, 2).map(i => (i, i+1))
+// Error: 20 expected array, found content
+$ mat(1, 2; ..nums;, delim: #none) $
+
```
wrzian, line 82: I'm not sure these tests match desired behavior. I feel like if we have spreads, they should always merge with surrounding arguments (and two adjacent spreads merge with each-other), requiring explicit semicolons to create separate rows in the matrix.

Playing around with 2d-arguments, I have a difficult time building a mental model of how normal arguments interact with spreads under the current system.

With normal args, changing a middle comma to a semicolon (i.e. `$mat(1,2,3,4)$` -> `$mat(1,2;3,4)$`) works fine, but spread arguments only join into one row when there are no semicolons. Adding a semicolon requires completely reshaping the data to avoid producing an error. In: `$mat(..#($1$,$2$), ..#($3$,$4$))$` -> `$mat(..#($1$,$2$) ; ..#($3$,$4$))$` -- the second equation gives: "Error: expected array, found content" on the _first_ argument.

This behavior feels wrong to me and feels like a reason to re-evaluate how spread arguments and semicolons should operate together.

This does also impact spreading arguments that create multiple arrays like below. I'm not sure what would be preferable here.
```typ
$ mat(..#range(1, 5).chunks(2)) $
```

I'll note that my current personal draft of math runtime parsing merges spreads with normal arguments. It's not worth waiting for that to land just to get this feature, but it will be a problem if the behavior changes between the two. I'd rather that we make this behavior as correct as possible here without runtime parsing.

I don't have a simple way to resolve this just within the parser. It could be that we need to add a new `MathArgs` Kind and AST node to get the correct behavior from the AST?

Opinions? @PgBiel @Enivex

crates/typst-syntax/src/parser.rs
```diff
@@ -455,84 +469,96 @@ fn math_args(p: &mut Parser) {
...
+                    maybe_array_start = p.marker();
+                }
+            }
+            SyntaxKind::Semicolon => {
+                if !positional {
+                    maybe_array_start = p.marker();
+                }
+
+                // Parses an array: `a, b, c;`.
+                // The semicolon merges preceding arguments separated by commas into an
```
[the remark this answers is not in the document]
mkorje: if only rustfmt would fix it for me :cry:

crates/typst-syntax/src/parser.rs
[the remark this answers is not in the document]
mkorje: Yeah I'm fine with adding `else { eat() }` over a new method.

tests/suite/math/mat.typ
mkorje: I think we can get this more expected behaviour quite simply by treating spread arguments as positional. The behaviour is a lot clearer then I think:

Spreading 1d arrays merges with other positional arguments as expected, since the result of spreading this should be no different than just writing the elements of the array directly. For example,
```typ
mat(..#range(1, 5) ; 1, ..#range(2, 5)) = mat(1, 2, 3, 4 ; 1, 2, 3, 4)
mat(..#range(1, 3), ..#range(3, 5) ; ..#range(1, 4), 4) = mat(1, 2, 3, 4 ; 1, 2, 3, 4)
```

Spreading a 2d array as the only argument still works:
```typ
mat(..#((1,) * 5).intersperse(0).chunks(3), delim: "[") = mat(1, 0, 1; 0, 1, 0; 1, 0, 1; delim: "[")
```

Spreading a 2d array with other positional arguments and semicolons are a little trickier at first, but still make sense with the behaviour a user will know from what was presented before. For example:
```typ
#let nums = range(0, 2).map(i => (i, i+1))

// Will display (0,1) and (1,2) verbatim as
// using the semicolon re-wraps it into an array
$ mat(..nums;) $

// Will throw an "expected array, found content"
// error as the end empty argument isn't in an array
// (same as what happens currently on main)
$ mat(1, 2; ..nums;, delim: #none) $

// Will throw an "expected array, found content"
// error as the last two arguments aren't in an
// array
$ mat(..nums, 0, 1) $

// Will throw an "expected array, found integer"
// error as it is equivalent to `(..nums, 0, 1)`
$ mat(..nums, ..#range(0, 2))
```

More clearly, here is the repr of some arguments from the tests:
```typ
#let nums = range(0, 4).chunks(2)
#check($args(..nums)$, "((0, 1), (2, 3))")
#check($args(..nums;)$, "(((0, 1), (2, 3)))")
#check($args(..nums, ..nums)$, "((0, 1), (2, 3), (0, 1), (2, 3))")
#check($args(..nums, 4, 5)$, "((0, 1), (2, 3), [4], [5])")
#check($args(..nums, ..#range(4, 6))$, "((0, 1), (2, 3), 4, 5)")
#check($args(..nums, #range(4, 6))$, "((0, 1), (2, 3), (4, 5))")
#check($args(..nums, 1, 2; 3, 4)$, "(((0, 1), (2, 3), [1], [2]), ([3], [4]))")
#check($args(1, 2; ..nums)$, "(([1], [2]), ((0, 1), (2, 3)))")
#check($args(1, 2; 3, 4)$, "(([1], [2]), ([3], [4]))")
#check($args(1, 2; 3, 4; ..#range(5, 7))$, "(([1], [2]), ([3], [4]), (5, 6))")
#check($args(1, 2; 3, 4, ..#range(5, 7))$, "(([1], [2]), ([3], [4], 5, 6))")
#check($args(1, 2; 3, 4, ..#range(5, 7);)$, "(([1], [2]), ([3], [4], 5, 6))")
#check($args(1, 2; 3, 4, ..#range(5, 7),)$, "(([1], [2]), ([3], [4], 5, 6))")
```

crates/typst-syntax/src/parser.rs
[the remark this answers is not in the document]
laurmaedje: I use the "Rewrap" extension in VS Code.

crates/typst-syntax/src/lexer.rs
laurmaedje: I think I kept it intentionally to mirror is_id_start

mkorje: Would you prefer adding it back then?

crates/typst-syntax/src/parser.rs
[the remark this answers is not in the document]
mkorje: I'm using Helix atm. Maybe https://github.com/rust-lang/rustfmt/issues/3347 will be stabilised soon...

crates/typst-syntax/src/lexer.rs
laurmaedje: Yeah, I think it makes it so that one stumbles less upon seeing this code.

crates/typst-syntax/src/parser.rs
```diff
@@ -455,84 +472,96 @@ fn math_args(p: &mut Parser) {
...
-        maybe_wrap_in_math(p, arg, count, named);
-        if named.is_some() {
-            array = p.marker();
+    p.expect_closing_delimiter(m, SyntaxKind::RightParen);
+    p.wrap(m, SyntaxKind::Args);
+}
+
+/// Parses a single argument in a math argument list.
+///
+/// Returns whether the argument parsed was positional or not.
```
laurmaedje, line 527:
```suggestion
/// Returns whether the parsed argument was positional or not.
```

crates/typst-syntax/src/parser.rs
```diff
@@ -455,84 +472,96 @@ fn math_args(p: &mut Parser) {
...
+        }
     }
 
-    if p.at(SyntaxKind::Text) && p.current_text() == ")" {
-        p.convert_and_eat(SyntaxKind::RightParen);
-    } else {
-        p.expected("closing paren");
-        p.balanced = false;
-    }
+    // Parses a normal position argument.
```
laurmaedje, line 558:
```suggestion
    // Parses a normal positional argument.
```

crates/typst-syntax/src/parser.rs
```diff
@@ -455,84 +472,96 @@ fn math_args(p: &mut Parser) {
...
+    let arg = p.marker();
+    let count = math_exprs(p, syntax_set!(End, Comma, Semicolon, RightParen));
+    let named = if positional { None } else { Some(m) };
+    maybe_wrap_in_math(p, arg, count, named);
```
laurmaedje, line 562: I feel like `maybe_wrap_in_math` doesn't really pull its weight as an abstraction anymore. Perhaps it should be inlined into `fn arg`?

tests/suite/math/call.typ
```diff
@@ -8,6 +8,97 @@ $ pi(a,) $
...
+#let func(my: none) = my
+// Error: 15-17 duplicate argument: my
+$ func(my: a, my: b) $
+
+--- math-call-named-single-underscore ---
+#let func(x) = x
+// Error: 8-9 expected identifier, got underscore
+$ func(_: a) $
+
+--- math-call-named-single-char-error ---
```
laurmaedje, line 48: The name is a bit odd. The problem seems to be the space after the "m" more than the fact that it's a single char.

tests/suite/math/call.typ
```diff
@@ -8,6 +8,97 @@ $ pi(a,) $
...
+#check($args(a-b)$, "(sequence([a], [−], [b]))")
+
+--- math-call-spread-content-error ---
+#let args(..body) = body
+// Error: 7-16 cannot spread content
+$args(..(a + b))$
+
+--- math-call-spread-multiple-exprs ---
+#let args(..body) = body
+// Error: 11-12 unexpected text
```
laurmaedje, line 72: the error message is not really ideal here. not sure whether we can improve, but in principle a missing comma error would be nicer.

tests/suite/math/call.typ
```diff
@@ -8,6 +8,97 @@ $ pi(a,) $
...
+--- math-call-spread-repr ---
+#let args(..body) = body
+#let check(it, r) = test-repr(it.body.text, r)
+#check($args(..#range(0, 4).chunks(2))$, "((0, 1), (2, 3))")
+#check($#args(range(1, 5).chunks(2))$, "(((1, 2), (3, 4)))")
+#check($#args(..range(1, 5).chunks(2))$, "((1, 2), (3, 4))")
+#check($args(#(..range(2, 6).chunks(2)))$, "(((2, 3), (4, 5)))")
+#let nums = range(0, 4).chunks(2)
+#check($args(..nums)$, "((0, 1), (2, 3))")
+#check($args(..nums;)$, "(((0, 1), (2, 3)))")
```
laurmaedje, line 89: In _all_ tests except for this one I can replace `..nums` with `#(0, 1), #(2, 3)` and it works correctly. This one yields a different result though. What is special about it?

tests/suite/math/call.typ
```diff
@@ -35,6 +126,34 @@ $ mat(#"code"; "wins") $
 #check($args(a,b;c)$, "(([a], [b]), ([c],))")
 #check($args(a,b;c,d;e,f)$, "(([a], [b]), ([c], [d]), ([e], [f]))")
 
+--- math-call-2d-named-repr ---
+#let args(..body) = (body.pos(), body.named())
+#let check(it, r) = test-repr(it.body.text, r)
+#check($args(a: b)$, "((), (a: [b]))")
+#check($args(1, 2; 3, 4)$, "((([1], [2]), ([3], [4])), (:))")
+#check($args(a: b, 1, 2; 3, 4)$, "((([1], [2]), ([3], [4])), (a: [b]))")
+#check($args(1, a: b, 2; 3, 4)$, "(([1], ([2],), ([3], [4])), (a: [b]))")
```
laurmaedje, line 135: Having 2 at a deeper array nesting level than 1, but the case it admittedly very cursed and I'm not sure whether we can improve it. 

Overall, for many of these mixed cases, it's anyone's guess on what the result is ...

laurmaedje (commented):
Looks pretty good!

tests/suite/math/call.typ
mkorje: See my reply to wrzian above, I think that gives some more context and explanation here: https://github.com/typst/typst/pull/5008#discussion_r1886237086

tests/suite/math/call.typ
mkorje: We probably can improve it I think. I kept it like this to match the current behaviour: it's a bit confusing but a named argument will update where an array can potentially start. In the example above, it was at `1`, then the named argument changed it to `2`, and so the following semicolon wraps the `2` in an array only.

crates/typst-syntax/src/lexer.rs
```diff
@@ -685,6 +690,44 @@ impl Lexer<'_> {
...
+        self.s.jump(start);
+        if self.s.eat_if(is_id_start) {
+            self.s.eat_while(is_id_continue);
+            // Check that a colon directly proceeds the identifier.
+            if self.s.at(':') {
+                // Check that the identifier is not just `_`.
+                let node = if self.s.from(start) != "_" {
+                    SyntaxNode::leaf(SyntaxKind::Ident, self.s.from(start))
+                } else {
+                    let msg = SyntaxError::new("expected identifier, got underscore");
```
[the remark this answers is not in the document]
laurmaedje: Can't reproduce that:
```typ
#let f(foo: 1) = {}
#f(_: 0)
```
gives me `expected expression, found underscore`

The term "got" doesn't appear in the codebase in any error message context.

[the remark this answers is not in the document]
mkorje: Huh, I could've sworn I tried it the other day and it said "got". Obviously I was very wrong. Will change.

tests/suite/math/call.typ
wrzian: This is where runtime math parsing comes in :)

mkorje: Would that even been needed to do this? We could say store all the positional arguments as we loop and then at the semicolon build a SyntaxNode of an array with them manually.

tests/suite/math/call.typ
mkorje: My point was that if it weren't a single character you'd get an unknown variable error instead.

tests/suite/math/call.typ
mkorje: I think we can just change this line to improve it.
https://github.com/typst/typst/blob/b44c90ef33ec1789bdd9a5c8d614e9ef2f43e05c/crates/typst-syntax/src/parser.rs#L512
Though you could also put a semicolon instead of a comma.

crates/typst-syntax/src/parser.rs
```diff
@@ -455,84 +472,96 @@ fn math_args(p: &mut Parser) {
...
-                if named.is_some() {
-                    array = p.marker();
-                    named = None;
+            SyntaxKind::End | SyntaxKind::Dollar | SyntaxKind::RightParen => {
+                // Check if we have reached the end, after the last argument.
+                if has_arrays && positional {
+                    p.wrap(maybe_array_start, SyntaxKind::Array);
+                    // So we don't wrap as an array twice after the loop.
+                    array_last = false;
                 }
```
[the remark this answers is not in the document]
mkorje: I think this is a remnant from an early stage that I forgot to remove. I'm pretty sure it has no purpose now. Will remove.

tests/suite/math/call.typ
wrzian: That would break the property that iterating over the syntaxnodes in order rebuilds the source text.

wrzian: This would be possible with changes at the AST layer though

mkorje: Ah, I see.

laurmaedje: How does runtime parsing fix it?

tests/suite/math/call.typ
laurmaedje: I did read the reply, but I still don't understand why 
```typ
#let mat(..args) = args
$ mat(#(1, 2), #(3, 4)) $
$ mat(#(1, 2), #(3, 4);) $
```
are exactly the same, but
```typ
#let mat(..args) = args
#let nums = ((1, 2), (3, 4))
$ mat(..nums) $
$ mat(..nums;) $
```
are different.

tests/suite/math/call.typ
wrzian: With runtime parsing I can keep the dictionary and list of args completely separate, so if you do `$ func(1, 2; 3, named: x, 4) $` The named parameter doesn't interrupt the positional args being generated as `(1,2),(3,4)` in any way. We could also do this if we make the AST more aware of the 2-dimensional structure, but I am a bit biased.

tests/suite/math/call.typ
wrzian: This stumped me too for a bit, but it's just the semicolon on the code expression: `#(2, 3);`. So the parser just sees two normal arguments :upside_down_face:

tests/suite/math/call.typ
wrzian: What if instead of parsing one math expression after dots, we parse until the next comma? Then the error here would happen during evaluation.

Changing `math_expr(p)` to `math_exprs(p, syntax_set!(...))` when parsing the spread, the error becomes: "Error: 7-14 (`..a + b`) cannot spread content"

tests/suite/math/call.typ
mkorje: Ahhhh, so simple in the end. I was not looking forward to trying to debug this...

tests/suite/math/call.typ
mkorje: Yeah I'm not sure what I was saying before, definitely can't change what I suggested.

I changed it at some point to only parse one expression after dots. I'll change it back, I think that makes this error a lot nicer.

mkorje: Actually, never mind, I think what I was saying before makes sense. I just changed that line to `p.expected("comma or semicolon")`

```
error: expected comma or semicolon
  ┌─ tests/suite/playground.typ:3:9
  │
3 │ $args(..a + b)$
  │          ^
```

wrzian: See, I'm not sure that it really is the error message we want, because it doesn't help the user understand *why*.

wrzian: There should be at least some indication that the parser tried to parse a spread expression here.

mkorje: I'd argue that this already kind of does that. It's saying a comma or semicolon was expected after the `a`, signalling to a user that the dots only apply to a single expression.

wrzian: Yes, but after they insert a comma they'll get one more error: "cannot spread content"

wrzian: Hmm. The best error is context-sensitive. If the item after the dots is spreadable, then they're missing a comma. If the item after the dots isn't, then they need to escape the dots or something

mkorje: That makes sense though I'd think, so I don't see the issue with getting the other error. The `a` is content after all.

tests/suite/math/call.typ
laurmaedje: oh that's so dumb. and it goes to show that maybe we should reconsider the whole semicolon termination thing.

tests/suite/math/call.typ
laurmaedje: I think this is good enough for now

laurmaedje (approved):
Ready to merge?

mkorje:
Yep, unless @wrzian has anything more to add.

mkorje:
I think the changes from #5652 are causing the checks to fail. Give me a min and I'll rebase.

laurmaedje:
Thank you!

wrzian:
Sorry I didn't reply earlier, but LGTM :+1:

PR #5012, opened 2024-09-23
Improve code of nix flake a bit

frozolotl:


PgBiel:
Seems fine (cc @drupol)

drupol:
Nice!

flake.nix
```diff
@@ -106,22 +104,7 @@
...
-            pkgs.libiconv
-          ]) ++ [
-            pkgs.openssl
-          ];
-
-          nativeBuildInputs = [
-            pkgs.pkg-config
-            pkgs.openssl.dev
-          ];
+          inputsFrom = [ typst ];
```
PgBiel, line 107: I noticed that this doesn't add `rustc` anymore. May be worth explicitly adding it back. `cargo` still seems to be there already, however.

(Unless my setup is somehow dirty - may be worth testing locally.)

PgBiel (commented):
Thanks, I can confirm this builds, and the dev shell properly overrides cargo.

flake.nix
PgBiel: Tbh it may be worth adding both cargo and rustc explicitly regardless. We only need `inputsFrom = [ typst ];` to ensure we pull the native stuff anyway.

frozolotl:
I expanded the scope of this PR a bit and added checks which can be run using `nix flake check`. They are the same ones as defined in the GitHub CI workflow.

I also updated the inputs and re-added `rustc`, `cargo fmt`, etc. by using `craneLib.devShell`.

flake.nix
```diff
@@ -19,76 +16,70 @@
...
+        craneLib = crane.mkLib pkgs;
+
+        # Typst files to include in the derivation.
+        # Here we include Rust files, docs and tests.
+        src = lib.sourceByRegex ./. [
+          "(docs|crates|tests)(/.*)?"
+          ''Cargo\.(toml|lock)''
+          ''build\.rs''
+          "rustfmt.toml"
+        ];
```
PgBiel, line 35: This seems like a good opportunity to move away from `sourceByRegex`, last time I checked it had some problems on MacOS (due to https://github.com/NixOS/nix/issues/1537). Maybe we can just copy and paste what I did for my project, Glistix:
https://github.com/Glistix/glistix/blob/3b578d459db8cb528e0c19682dabe07094626373/nix/glistix.nix#L12-L35

PgBiel (commented):
Good stuff!

mkorje:
Given the scope of this PR has expanded, it'd be nice if we could add a `testit` alias as described [here](https://github.com/typst/typst/blob/main/tests/README.md#making-an-alias) to the dev shell. Maybe something like this

```nix
(pkgs.writeScriptBin "testit" ''
  cargo test --workspace --test tests -- "$@"
'')
```

added to `packages` in `devShells.default`.

mkorje:
Also, maybe it'd be a good idea to pin the Rust version instead of using what's in `nixpkgs`, by using [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay) or [nix-community/fenix](https://github.com/nix-community/fenix).

PgBiel:
Sure, let's do it

flake.nix
```diff
@@ -105,23 +96,19 @@
...
+        };
 
-          nativeBuildInputs = [
-            pkgs.pkg-config
-            pkgs.openssl.dev
-          ];
+        devShells.default = craneLib.devShell {
+          checks = self'.checks;
+          inputsFrom = [ typst ];
         };
```
mkorje, line 112:
```suggestion
        devShells.default = craneLib.devShell {
          checks = self'.checks;
          inputsFrom = [ typst ];
          packages = [
            (pkgs.writeScriptBin "testit" ''
              cargo test --workspace --test tests -- "$@"
            '')
          ];
        };
```

flake.nix
```diff
@@ -1,13 +1,10 @@
...
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
-
+    flake-parts.url = "github:hercules-ci/flake-parts";
     systems.url = "github:nix-systems/default";
+
+    crane.url = "github:ipetkov/crane";
   };
 
   outputs = inputs@{ flake-parts, crane, nixpkgs, self, ... }: flake-parts.lib.mkFlake { inherit inputs; } {
```
mkorje, line 10:
```suggestion

    rust-overlay = {
      url = "github:oxalica/rust-overlay";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = inputs@{ self, nixpkgs, flake-parts, systems, crane, rust-overlay }: flake-parts.lib.mkFlake { inherit inputs; } {
```

flake.nix
```diff
@@ -19,76 +16,70 @@
 
     perSystem = { self', pkgs, lib, ... }:
       let
```
mkorje, line 18:
```suggestion
    perSystem = { self', system, lib, ... }:
      let
        pkgs = import nixpkgs {
          inherit system;
          overlays = [ (import rust-overlay) ];
        };

```

flake.nix
```diff
@@ -19,76 +16,70 @@
...
-          craneLib.buildPackage (commonCraneArgs // {
-            inherit cargoArtifacts;
+        cargoToml = lib.importTOML ./Cargo.toml;
+
+        pname = "typst";
+        version = cargoToml.workspace.package.version;
+
+        # Crane-based Nix flake configuration.
+        # Based on https://github.com/ipetkov/crane/blob/master/examples/trunk-workspace/flake.nix
+        craneLib = crane.mkLib pkgs;
```
mkorje, line 26:
```suggestion
        craneLib = (crane.mkLib pkgs).overrideToolchain (p: p.rust-bin.stable."1.80.0".default);
```

mkorje (commented):
I added in the comments below the changes that should be needed to use [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay). Though I'm not 100% sure this is correct as I'm not very familiar with `flake-parts`.

An aside (in my very subjective opinion): I don't really like that we are using `flake-parts`. Pulling it in as an input isn't really needed, when `flake-utils` or just simply writing a small helper function (like `forEachSystem` or smth) would suffice.

drupol:
@mkorje 

1. [`flake.parts`](https://flake.parts/) is a Flake framework facilitating the creation and the modularisation of Flakes, just like `flake-utils`. Just like your subjective opinion, this is indeed was an opinionated choice given the popularity and simplicity of that framework. Replacing it with something else, at this point, would be a bit of a nonsense and out of the scope of the PR. That said, we could get rid of these frameworks and use a helper functions, but then it makes the `flake.nix` file much more messy and not prone to DRY any more, the purpose of frameworks is to avoid that.

2. `inputs.systems` could be possibly replaced with a static list of architectures, but then it's less flexible. Have a look at this page explaining the reasons: https://github.com/nix-systems/nix-systems
And to answer your question, `flake.parts` require the list of `systems` to be set, it doesn't do it automatically for us, so there's no redundant use of `inputs.systems` here.

mkorje:
@drupol Fair enough, and I agree it wouldn't make much sense to replace it now. Of course using `flake.parts` is fine, even if I wouldn't use it myself.

Ah okay, I understood the utility of `inputs.systems`, but I didn't know that `flake.parts` required `systems` to be set. It looked as though it wasn't being used, my mistake.

flake.nix
mkorje: This is probably wrong and may not work. https://flake.parts/overlays#consuming-an-overlay gives a better example.

drupol: Indeed, you're right, the example you pointed out is the preffered way to do that.

PgBiel: Fenix has a crane example and is part of nix-community, we could consider using it. https://github.com/nix-community/fenix

PgBiel:
@mkorje I appreciate your further feedback, though I'd like to wrap this up now to ensure the changes can be applied in time. If @frozolotl prefers, we can consider wrapping up the PR as is and opening a separate one with the new proposals.

frozolotl:
I have now implemented the given suggestions.

I decided to use fenix instead of rust-overlay because it is smaller and under the nix-community organization.

And I decided to use Rust 1.80.1 because it incorporates a few fixes not present in Rust 1.80.0. But it's otherwise the same thing as the version used in CI (1.80.0).
Another PR should update the used Rust version to Rust 1.81.0 both in CI and in the flake.
The change should be pretty easy to do as long as the author has nix installed (necessary to regenerate the lock file).

PgBiel (approved):
Thank you!

flake.nix
```diff
@@ -1,7 +1,6 @@
 {
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
-    flake-parts.url = "github:hercules-ci/flake-parts";
```
drupol: This is actually needed as per documentation: https://github.com/hercules-ci/flake-parts/commit/14114c563acab4cee2828aa8dad9036d617e1ea2

PgBiel: My bad, you had convinced me that it wasn't necessary in the past and everything worked without, so I thought it was fine, haha 🙂

I'll revert my commit soon, unless @frozolotl is faster than me 🙂

PgBiel (approved):
There we go! Sorry for my small mistake above. :)

PR #5016, opened 2024-09-24
Remedy a 404 in the changelog

reknih:
There is one other 404, `math.binom`, whose documentation page is for some reason not rendered. We should fix that in a future PR.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -865,7 +865,8 @@ description: |
 
 - Command line interface
   - Added `typst query` subcommand to execute a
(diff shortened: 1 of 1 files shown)


wrzian:
> There is one other 404, `math.binom`, whose documentation page is for some reason not rendered. We should fix that in a future PR.

That might be because its implementation is in the same file as `frac.rs`? although I'm not fully sure how the docs system is set up.

laurmaedje:
The file structure shouldn't matter. Something weird is going on here.

Issue #4780, opened 2024-08-18
Footnote in auto placed figure causing figure to placed on top of text


winstonewert:
### Description

The following code:

```
#set text(font:"Roboto",size:12pt,weight:"regular",style:"normal",);

#lorem(50);

#lorem(60);

#lorem(60);

#lorem(60)

#lorem(40)

#lorem(50)

#lorem(30)

#figure(placement: auto,
[#lorem(40)#footnote(lorem(50))],
)
#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

```

Place the figure on top of the text on the next page.

![shot](https://github.com/user-attachments/assets/16d0cde5-da76-4943-9601-7a179b928a06)

My hypothesis would be that it doesn't properly take the footnote into account during the placement process and and then it give shifted up and into the text.

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

rriemann:
Works for me with version 0.11.1.

<details><summary>Screenshot</summary>
<p>

![image](https://github.com/user-attachments/assets/2644d18d-f58c-4f46-97fe-1a3838f4c05b)


</p>
</details>

winstonewert:
It still fails for me on the webapp. It looks to me like your document didn't pick up the Roboto font I was using and thus didn't get laid out the same way. In particular, I think your text is a bit smaller and thus the figure doesn't get forced on the second page which causes the problem.

Here's a typst source that doesn't rely on a special font:

```
#set text(size:12.5pt,weight:"regular",style:"normal",);

#lorem(50);

#lorem(60);

#lorem(60);

#lorem(60)

#lorem(40)

#lorem(50)

#lorem(30)

#figure(placement: auto,
[#lorem(40)#footnote(lorem(50))],
)
#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)

#lorem(50)
```

rriemann:
@winstonewert , you are right. I used the wrong font. Your example above reproduces the bug.

Proof:

![image](https://github.com/user-attachments/assets/8c5d2eeb-4796-4642-87d2-38c696989585)

Issue #4220, opened 2024-05-22
Support fraction as block height


knuesel:
### Description

It would be nice to have `block(height: 1fr)` just work, at least for non-breakable blocks. Currently it throws an error because fractional height is not supported. A workaround proposed on [discord](https://discord.com/channels/1054443721975922748/1088371919725793360/1242136749858230353) is

```
#context place(block(
    height: (locate(<end>), here()).map(l=> l.position().y).fold(
        l=> r=> l - r,
        (f, r)=> f(r),
    ),
    width: 100%,
    fill: red,
))
#v(1fr)<end>
#pagebreak(weak: true)
```

which is not easy to come up with.

Using `1fr` in a single-cell grid doesn't work because of #2571.

Other workarounds involving computations with the current location and page margin are not ideal since it's difficult to get the current margin in a way that works in all cases (`margin: auto`, `margin: 1cm`, `margin: (left: 1cm)` and the worst: `margin: 1em`).

### Use Case

A typical use case is to make an element use the remaining space on a page (see [here](https://discord.com/channels/1054443721975922748/1088371919725793360/1242131364724084927), [here](https://discord.com/channels/1054443721975922748/1237316434384781333) and [here](https://discord.com/channels/1054443721975922748/1238141210154111029)). For example I wanted to draw a grid with cetz in the space remaining under a question (to emulate grid paper).

Issue #3907, opened 2024-04-10
Breakable block skips region due to placed floating element


SUPERCILEX:
### Description

```
#set page(columns: 2, width: 40em, height: 40em)

#lorem(50)

#figure(placement: auto, caption: "test", v(10em))
#figure(placement: auto, caption: "test", v(10em))

+ #lorem(30)
+ #lorem(20)
+ #lorem(20)
+ #lorem(10)
```

Take a look at https://discord.com/channels/1054443721975922748/1056574658771681361/1227424678969348117. The first page is broken because it forgets that there's space leftover and instead spills to the second page.

### Typst version

- [ ] I am using the latest version of Typst

PgBiel:
I've reproduced this with a `placement: auto` figure before a long list containing a figure or any vertically long content. The `placement: auto` figure introduces the page and oddly forces the list to go to the next page, as if the list were unbreakable.

laurmaedje:
Most likely, the list is layouted without knowledge of the second figure and then it's second frame is too big to fit, so it's moved to the next region.

laurmaedje:
Minimized:
```typ
#set page(width: 120pt, height: 120pt, margin: 10pt)

#block(width: 100%, height: 50pt, fill: aqua)
#figure(
  placement: top,
  rect()
)

#list(.."ABCDEFGHIJ".clusters())
```

Issue #3866, opened 2024-04-04
Unbreakable block becomes invalid when moved into differently sized region


jkunimune:
### Description

I have a document with two columns of text and a figure that is over half the height of the page.  When I compile my document, the figure is where I expect it, but the following text isn't pushed down far enough, and overlaps both the figure and its caption.  I would expect it to position the text below the caption; there's even enough space to fit a full paragraph down there if it does.

This doesn't happen every time I have a figure this tall.  It seems the paragraphs need to line up just right for it, but I haven't able to find a way to predict when it will or won't happen.

Minimum working example: <details>
```typst
#text(weight: 700, 1.75em, lorem(17))
#pad(x: 2em, {
  v(1.1em)
  heading(outlined: false, numbering: none, text(0.85em)[Abstract])
  lorem(150)
  v(1.1em)
})

#show: columns.with(2, gutter: 0.5in)

= Introduction

#lorem(50)

#figure(
  rect(width: 2.3in, height: 1.9in),
  caption: lorem(10),
)

#lorem(50)

#figure(
  rect(width: 2.5in, height: 7in),
  caption: lorem(10),
)

#lorem(30)

#lorem(30)

#lorem(30)

#lorem(30)
```
</details>

The problem is on page 2:
![imagen](https://github.com/typst/typst/assets/5607801/5565a8dd-edf4-4308-881a-c62ed8605ad4)

### Reproduction URL

https://typst.app/project/rr9vwQjcgU17NGm79GAU5O

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
Minimized:
```
#v(10cm)
#show: columns.with(2)
#lorem(250)
#block(rect(width: 100%, height: 20cm), breakable: false)
#lorem(30)
```

The problem is that the unbreakable block overflows in the first region (less height), is then migrated into the second region, but there we do have enough space, so we need to relayout.

PR #5017, opened 2024-09-24
New flow layout, with multi-column floats

laurmaedje:
This is a full, fundamental redesign and rewrite of Typst's most central layouter, the _flow layouter._ This layouter is responsible for arranging paragraphs, blocks, spacing, placed elements, floats, and footnotes across pages.

### Why a new flow layouter?

The old flow layouter was written a long time ago and grew over time. New features, like footnotes and floats, were mostly bolted on top: This showed up in a lack of features (e.g. floats over two columns) and bugs (e.g. footnotes in breakable blocks flowing away to the last page of the block). To address those properly required a fundamental shift in how the flow operates. The old flow layout proceeded in a strictly linear manner: It processed elements one by one, with no way to go back and revise a decision. This mode of operation is quite limiting and can fundamentally not handle all features we want Typst to have.

### How does it work?

The new flow layout operates in a sort of flipped-around way: Instead of walking _elements_ one by one, it walks _pages_ one by one. For each page, it fills it with content as long as space is available. And here comes the crucial part: While laying out a page, it can decide to trigger a _relayout_ of the page. This happens when a float or footnote is inserted, as this can affect the layout of the page's inner content. I wrote a bit more about this approach in [a blog post](https://laurmaedje.github.io/posts/layout-models/) some time ago: In summary, the approach of gaining information and then relayouting with that new information can help us iteratively solve various problems that can't be solved in a single step. It's sort of similar to how Typst already runs multiple layout iterations to resolve introspection, but additional and add the page-level. (It might seem like a good idea to actually _use_ the normal introspection for these things, but one runs into convergence issues immediately, which is why the iteration scope must be tighter.)

### What does it do?

User-facing, this PR
- Adds new page-scoped floating placement
- Adds a new public `block.sticky` to property to prevent a pagebreak after a block. This is used internally by headings.
- The height of blocks, images, and some shapes can now be specified in fr units.
- Fixes many bugs, the most important ones being:
  - No more orphaned headings at the bottom of the page.
  - Footnotes in breakable blocks now work properly. It can still happen that a footnote is pushed to the next page due to a lack of space, but it's much rarer.

The new floating placement looks like this:

```typ
#set page(columns: 2)

// Raw placement.
#place(
  top,
  scope: "page",
  float: true,
  [This spans both columns]
)

// With a figure.
#figure(
  placement: auto,
  scope: "page",
  captiion: [My caption],
  rect(width: 75%),
)

#lorem(200)
```

The PR also lays the groundwork for exciting future features like to-the-side wrapping floats, layout through chained containers, page margins depending on the page number, and more.

### Notes

- In my very unscientific testing on a few documents, performance was not affected much. While working on the PR, I primarily on focused on getting things right and not hurting performance too much, so that's a good result I would say. Fundamentally, doing relayouts is of course more expensive than not doing them.

- The PR also adds the new file `tests/skip.txt`, which can be used to temporarily disable all the tests listed within. This is very useful during large incremental refactors.

### Linked issues

- Closes https://github.com/typst/typst/issues/553
- Closes https://github.com/typst/typst/issues/993
- Closes https://github.com/typst/typst/issues/4220
- Fixes https://github.com/typst/typst/issues/3866
- Fixes https://github.com/typst/typst/issues/2213
- Fixes https://github.com/typst/typst/issues/3907
- Fixes https://github.com/typst/typst/issues/2449
- Fixes https://github.com/typst/typst/issues/1044
- Fixes https://github.com/typst/typst/issues/4780


jamesrswift:
Hecking AWESOME

rtbs-dev:
_Y'all, it's HAPPENING_ 
It's happening!

crates/typst/src/layout/flow/distribute.rs
```diff
@@ -0,0 +1,510 @@
...
+        Ok(()) => true,
+        Err(Stop::Finish(forced)) => forced,
+        Err(err) => return Err(err),
+    };
+    let region = Region::new(regions.size, regions.expand);
+    distributor.finalize(region, init, forced)
+}
+
+/// State for distribution.
+struct Distributor<'a, 'b, 'c, 'd, 'e> {
```
bluebear94, line 32: With the number of lifetime parameters here, I think it’d be helpful to give them more meaningful names.

laurmaedje: Sort of unfortunate that so many are needed. It's because of variance. The only interesting ones are 'a and 'b. These are used across the flow layout.

emilyyyylime: Could you possibly name just them and leave the other three as generic `'a`, `'b`, `'c`? This would make the distinction between the important

laurmaedje: I'm not a huge fan of lifetimes with long names because I feel like they clutter things without adding that much value. The difference between 'a and 'b is not that interesting and if Rust would've let me, I would've used 'a for both. What matters in my opinion is that 'a and 'b mean the same thing across their usages in one subsystem. 

But I agree that it's relevant to differentiate between the two shared ones and the three irrelevant ones. Maybe I could just do what I did in realize.rs and call them 'a, 'b, 'x, 'y, 'z and then add a doc comment explaining things.

spped2000:
Cool!

0x7CFE:
Finally! Thanks for the hard work.

RomeoV:
For future reference, in the example given by @laurmaedje above I believe `scope: "page"` has since become `scope: "parent"`.

PR #5020, opened 2024-09-24
Spelling

jsoref:
Fixes misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/typst/actions/runs/11023102048#summary-30613757370

The action will report that the changes in this PR would make it happy: https://github.com/jsoref/typst/actions/runs/11023102329#summary-30613759162

Note: there are some items which I expect to get pushback on, I'm more than happy to drop any particular commit(s). I'll take some time to annotate things...

PR #5021, opened 2024-09-24
Upload artifacts on CI failure

jsoref:
While working on #5020 it was frustrating not having access to the rendered content. As GitHub workflows already do the work of generating the images, it seemed helpful to provide them in an artifact.

Note it's possible to specify [`retention-days: 1`](https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period) or similar (for some things, I use that value myself). I was lazy in drafting this so haven't set a value.

This is just a proposal, I'm happy to adjust it...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,28 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace --no-run
       - run: cargo test --workspace --no-fail-fast
+      - if: failure()
+        name: Upload rendered test output
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-artifacts-${{ matrix.os }}
+          path: tests/store/render/**
+          retention-days: 3
+
+      - if: failure()
+        name: Update test artifacts
+        run: |
+          : Update test artifacts
+          cargo test --all --test tests -- --update
+          echo 'updated_artifacts=1' >> "$GITHUB_ENV"
+
+      - if: failure() && env.updated_artifacts
+        name: Upload updated reference output (for use if the test changes are desired)
+        uses: actions/upload-artifact@v4
+        with:
+          name: updated-artifacts-${{ matrix.os }}
+          path: tests/ref/**
+          retention-days: 3
 
   checks:
     name: Check clippy, formatting, and documentation
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,26 +34,23 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace --no-run
       - run: cargo test --workspace --no-fail-fast
-      - if: failure()
-        name: Upload rendered test output
+      - name: Upload rendered test output
+        if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: test-artifacts-${{ matrix.os }}
+          name: tests-rendered-${{ matrix.os }}
           path: tests/store/render/**
           retention-days: 3
-
-      - if: failure()
-        name: Update test artifacts
+      - name: Update test artifacts
+        if: failure()
         run: |
-          : Update test artifacts
-          cargo test --all --test tests -- --update
+          cargo test --workspace --test tests -- --update
           echo 'updated_artifacts=1' >> "$GITHUB_ENV"
-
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -34,6 +34,11 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace --no-run
       - run: cargo test --workspace --no-fail-fast
+      - if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-artifacts-${{ matrix.os }}
+          path: tests/*/render/**
```
laurmaedje, line 41:
```suggestion
          path: tests/store/render/**
```
that should be the only folder with a "render" subfolder

laurmaedje (commented):
I would maybe add `retention-days: 3` or something like that.

jsoref:
Here's a run with this: https://github.com/check-spelling-sandbox/typst/actions/runs/11058296381

This generates two artifacts, one with the tests as rendered (which is useful for understanding why something failed), and one with the updates for the tests (which is useful for anyone who wants to keep the updates -- which would have been me).

.github/workflows/ci.yml
```diff
@@ -34,6 +34,28 @@ jobs:
...
+          cargo test --all --test tests -- --update
+          echo 'updated_artifacts=1' >> "$GITHUB_ENV"
+
+      - if: failure() && env.updated_artifacts
+        name: Upload updated reference output (for use if the test changes are desired)
+        uses: actions/upload-artifact@v4
+        with:
+          name: updated-artifacts-${{ matrix.os }}
+          path: tests/ref/**
+          retention-days: 3
```
Coekjan, line 58: Seems that `env.updated_artifacts` is always true? If so, I would suggest:

```suggestion
      - if: failure()
        name: Upload rendered test output
        uses: actions/upload-artifact@v4
        with:
          name: test-artifacts-${{ matrix.os }}
          path: tests/store/render/**
          retention-days: 3

      - if: failure()
        name: Update test artifacts
        run: cargo test --all --test tests -- --update

      - if: failure()
        name: Upload updated reference output (for use if the test changes are desired)
        uses: actions/upload-artifact@v4
        with:
          name: updated-artifacts-${{ matrix.os }}
          path: tests/ref/**
          retention-days: 3
```

---

Please correct me if I am wrong.

.github/workflows/ci.yml
```diff
@@ -34,6 +34,28 @@ jobs:
...
+          name: test-artifacts-${{ matrix.os }}
+          path: tests/store/render/**
+          retention-days: 3
+
+      - if: failure()
+        name: Update test artifacts
+        run: |
+          : Update test artifacts
+          cargo test --all --test tests -- --update
+          echo 'updated_artifacts=1' >> "$GITHUB_ENV"
```
Coekjan, line 50: Would it be better to use `$GITHUB_OUTPUT`[^1]?

[^1]: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-output-parameter

jsoref: Using GITHUB_OUTPUT would require an `id:` field. It doesn't really matter which you use. Since the job is about to die and nothing past the next step cares, using GITHUB_ENV is fine.

But, if people prefer, I can certainly use it...

laurmaedje: Why is this needed in the first place, if it is written unconditionally on `failure()`. Why not just `if: failure()` on the job below?

jsoref: Technically, each of these items can fail:
```
      - uses: actions/checkout@v4
      - uses: dtolnay/rust-toolchain@1.80.0
      - uses: Swatinem/rust-cache@v2
      - run: cargo test --workspace --no-run
      - run: cargo test --workspace --no-fail-fast
      - if: failure()
        name: Upload rendered test output
        uses: actions/upload-artifact@v4
```

I've seen the first and the last fail.

Conceptually, this can fail as well (certainly it would fail if the first failed, but it would presumably also fail if the next two failed, and likely could fail if there was a bug in the cargo test code):
```
          cargo test --all --test tests -- --update
```

It only makes sense to try to upload an artifact if this last thing actually succeeded.

laurmaedje:
Thanks!

PR #5022, opened 2024-09-24
Connect maximum size with Updating reference images

jsoref:
I'm hoping this would make it easier for the next person like me to understand that the images generated by tests aren't quite the same as the files in the `tests/ref/` directory -- because those files are generally compressed -- and that the correct thing to do when encountering:

> ❌ reference image size exceeds 20.00 KiB, but the test is not marked as `// LARGE` (tests/suite/model/emph-strong.typ:4)

As opposed to just adding `// LARGE` is to run:

> cargo test --all --test tests -- --update

(or similar)


laurmaedje:
Thanks!

PR #5026, opened 2024-09-25
Add action for moving issues to Typst Forum

reknih:
This action replaces the "Convert to Discussion" GitHub feature with the label `to-forum`.

When the label is assigned, Theodore will post a new thread for the issue to the General category of the Typst Forum. GitHub Actions will close the issue as not planned.

### DIFF
--- a/.github/workflows/forum.yml
+++ b/.github/workflows/forum.yml
@@ -19,14 +19,16 @@ jobs:
           ISSUE_USER_LOGIN: ${{ github.event.issue.user.login }}
           REPOSITORY_OWNER: ${{ github.repository_owner }}
         run: |
-          RAW="This topic was moved from GitHub issue ${ISSUE_NUMBER}. Please continue the conversation here.
+          read -d '' RAW << END_OF_BODY
+          This topic was moved from GitHub issue ${ISSUE_NUMBER}. Please continue the conversation here.
 
           Here is the original issue:
-          [quote=\"${ISSUE_USER_LOGIN}\"]
+          [quote="${ISSUE_USER_LOGIN}"]
           ${ISSUE_BODY}
           [/quote]
 
-          Browse the previous discussion at ${ISSUE_URL}"
+          Browse the previous discussion at ${ISSUE_URL}
+          END_OF_BODY
 
           TITLE_JSON=$(jq -n --arg title "[#${ISSUE_NUMBER}] ${ISSUE_TITLE}" '$title')
           RAW_JSON=$(jq -n --arg raw "$RAW" '$raw')
--- a/.github/workflows/forum.yml
+++ b/.github/workflows/forum.yml
@@ -1,0 +1,60 @@
+name: Move to Typst Forum
+on:
+  issues:
+    types:
+      - labeled
+jobs:
+  add-comment:
+    if: github.event.label.name == 'to-forum'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Call Discourse API to create forum post
+        env:
+          ISSUE_BODY: ${{ github.event.issue.body }}
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+        run: |
+          RAW="This topic was moved from GitHub issue ${{ github.event.issue.number }}. Please continue the conversation here.
+
+          Here is the original issue:
+          [quote=\"${{ github.event.issue.user.login }}\"]
+          $ISSUE_BODY
+          [/quote]
+
+          Browse the previous discussion at ${{ github.event.issue.html_url }}"
+          TITLE_JSON=$(jq -n --arg title "[#${{ github.event.issue.number }}] $ISSUE_TITLE" '$title')
+          RAW_JSON=$(jq -n --arg raw "$RAW" '$raw')
+
+          RESPONSE=$(curl -X POST "https://forum.typst.app/posts.json" \
+            -H "Api-Key: ${{ secrets.DISCOURSE_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d "{
+              \"title\": $TITLE_JSON,
+              \"category\": 4,
+              \"external_id\": \"gh-${{ github.repository_owner }}-typst-${{ github.event.issue.number }}\",
+              \"raw\": $RAW_JSON
+            }")
+
+          echo $RESPONSE
+
+          # Check if response contains errors
+          if [ $(jq -n --argjson response "$RESPONSE" '$response.errors | length') -gt 0 ]; then
+            echo "Error creating topic"
+            exit 1
+          fi
+
+          echo "THREAD_ID=$(jq -n --argjson response "$RESPONSE" '$response.topic_id')" >> $GITHUB_ENV
+      - name: Add a comment if the workflow failed
+        if: failure()
+        run: |
+          gh issue comment ${{ github.event.issue.number }} --body "Failed to move this issue to the Typst Forum. Please see the details in the workflow run: ${{ github.run_id }}"
+      - name: Add comment and close as not planned
+        run: |
+          gh issue close "$NUMBER" --reason "not planned" --comment "$BODY"
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          RAW="This topic was moved from GitHub issue ${{ github.event.issue.number }}. Please continue the conversation here.
+
+          Here is the original issue:
+          [quote=\"${{ github.event.issue.user.login }}\"]
+          $ISSUE_BODY
+          [/quote]
+
+          Browse the previous discussion at ${{ github.event.issue.html_url }}"
+          TITLE_JSON=$(jq -n --arg title "[#${{ github.event.issue.number }}] $ISSUE_TITLE" '$title')
```
PgBiel, line 26: Out of abundance of caution, I'd probably move at least these event values (`user.login`, `html_url` and `number`) to environment variables as well, rather than interpolating them in the shell script, since these are potentially user-controlled values (maybe the issue number not so much, but it can receive the same treatment for the sake of consistency). This is important as any kind of shell injection here could leak the Discourse token down below (and any other secrets exposed to this particular script).

.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+          RAW_JSON=$(jq -n --arg raw "$RAW" '$raw')
+
+          RESPONSE=$(curl -X POST "https://forum.typst.app/posts.json" \
+            -H "Api-Key: ${{ secrets.DISCOURSE_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d "{
+              \"title\": $TITLE_JSON,
+              \"category\": 4,
+              \"external_id\": \"gh-${{ github.repository_owner }}-typst-${{ github.event.issue.number }}\",
```
PgBiel, line 35: (Same event value as before appears here too.)

.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+              \"raw\": $RAW_JSON
+            }")
+
+          echo $RESPONSE
```
PgBiel, line 39: I wonder if the response could potentially contain any sensitive information by mistake. Worth thinking about as this would be public. Could also just not `echo` this and, instead, log negative responses coming from GitHub in the Typst server's side for us to inspect internally if needed.

.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+
+          # Check if response contains errors
+          if [ $(jq -n --argjson response "$RESPONSE" '$response.errors | length') -gt 0 ]; then
+            echo "Error creating topic"
+            exit 1
+          fi
+
+          echo "THREAD_ID=$(jq -n --argjson response "$RESPONSE" '$response.topic_id')" >> $GITHUB_ENV
```
PgBiel, line 47: May be worth checking this:
```suggestion
          THREAD_ID="$(jq -n --argjson response "$RESPONSE" '$response.topic_id')"
          if [ "$THREAD_ID" = 'null' ]; then
            echo "Failed to retrieve new thread ID"
            exit 1
          fi
          echo "THREAD_ID=$THREAD_ID" >> $GITHUB_ENV
```

.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+              \"title\": $TITLE_JSON,
+              \"category\": 4,
+              \"external_id\": \"gh-${{ github.repository_owner }}-typst-${{ github.event.issue.number }}\",
+              \"raw\": $RAW_JSON
+            }")
+
+          echo $RESPONSE
+
+          # Check if response contains errors
+          if [ $(jq -n --argjson response "$RESPONSE" '$response.errors | length') -gt 0 ]; then
```
PgBiel, line 42: Not sure but could be worth it to check other failure modes somehow, e.g. error when receiving a 400 or 500 status code. In principle, though, I'm guessing this should cover most errors.

.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+          gh issue comment ${{ github.event.issue.number }} --body "Failed to move this issue to the Typst Forum. Please see the details in the workflow run: ${{ github.run_id }}"
+      - name: Add comment and close as not planned
+        run: |
+          gh issue close "$NUMBER" --reason "not planned" --comment "$BODY"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.issue.number }}
+          BODY: >
+            We moved this issue to the Typst Forum. Please continue the discussion there: https://forum.typst.app/t/${{ env.THREAD_ID }}
```
PgBiel, line 60: Could consider not running this one if there was a failure, since thread ID will be invalid.

.github/workflows/forum.yml
```diff
@@ -0,0 +1,60 @@
...
+        env:
+          ISSUE_BODY: ${{ github.event.issue.body }}
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+        run: |
+          RAW="This topic was moved from GitHub issue ${{ github.event.issue.number }}. Please continue the conversation here.
+
+          Here is the original issue:
+          [quote=\"${{ github.event.issue.user.login }}\"]
+          $ISSUE_BODY
+          [/quote]
```
PgBiel, line 23: Might be a bit unfortunate if either of those interpolated values invalidate the output somehow, but I think that's just life, and we can always manually fix it afterwards.

PgBiel (commented):
Nice! Quite clever usage of `jq`, I'll admit. (I was thinking of using it somehow, but couldn't think of a way to tell it to spit escaped JSON.)

.github/workflows/forum.yml
reknih: This happens automatically!

.github/workflows/forum.yml
reknih: I changed this to only echo the values from the `error` array if present.

.github/workflows/forum.yml
reknih: Done with `--fail-with-body`

reknih:
Thank you for the code review!

.github/workflows/forum.yml
```diff
@@ -0,0 +1,79 @@
...
+          REPOSITORY_OWNER: ${{ github.repository_owner }}
+        run: |
+          RAW="This topic was moved from GitHub issue ${ISSUE_NUMBER}. Please continue the conversation here.
+
+          Here is the original issue:
+          [quote=\"${ISSUE_USER_LOGIN}\"]
+          ${ISSUE_BODY}
+          [/quote]
+
+          Browse the previous discussion at ${ISSUE_URL}"
```
alerque, line 29: Did you know you could use shell HEREDOC syntax here?

reknih: I did not! How would that look? I have to admit, the only thing that [this explanation](https://stackoverflow.com/a/1655389/1567835) did is slightly confuse me.

alerque: The explanation you linked to is for another layer of complexity, that OP did *not* want to expand things in the content of the message. You do, which is why I thought the HEREDOC would be a clearer way to express that such that quotes don't need to be escaped and finding the ending quote mark isn't a game of needle in the haystack:

```suggestion
          read -d '' RAW << END_OF_BODY
          This topic was moved from GitHub issue ${ISSUE_NUMBER}. Please continue the conversation here.

          Here is the original issue:
          [quote="${ISSUE_USER_LOGIN}"]
          ${ISSUE_BODY}
          [/quote]

          Browse the previous discussion at ${ISSUE_URL}
          END_OF_BODY
```

I think the default shell for this runner is already `bash`, but if not you can set it with `shell: bash` just before this `run:` key.

reknih: Thank you! `bash` is the default, [see here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell).

PR #5027, opened 2024-09-25
Rename place scope "page" to "parent"

knuesel:
As discussed on discord, it might make more sense for `place` to offer `scope: "parent"` than `scope: "page"`.

It's more correct in nested scenarios where the parent is another container, and more forward-compatible as a `page` scope may be defined later to refer to the full page (including margins).


laurmaedje:
Thanks!

PR #5028, opened 2024-09-25
Remove columns.with show rule example

knuesel:
The syntax documentation includes an example `show: columns.with(2)`. Since this is now discouraged (users should use `set page(columns: ...)` instead to avoid interfering with footnotes, line numbers, etc.) this PR replaces the example with `show: template`.


laurmaedje:
Thanks!

Issue #200, opened 2023-03-23
delim not work in function mat


Grobycn:
Hi, according to https://typst.app/docs/reference/math/mat/, matrix can have different delimiter.
But it seems not work. No matter what delim is, the delimiter is always parentheses.

See my toy project:
https://typst.app/project/rNVhi8iHmSDRa5_V9sYpNB

jgm:
I'm having the same issue. To be more specific, my expectation was that
```
mat(delim: "[", 1, 2; 3, 4)
```
would give me a square-bracket-delimited matrix, but no matter what I use for delim I get parentheses.

I see that 'delim' is a "set parameter," and I know I can use `#set` to adjust it globally (or locally for a section of the document).  But other "set parameters" I have seen can also be specified in the command itself, e.g. you can say

```
#heading(numbering: "a.", [hi])
```
so my expectation was that it would work this way for `mat` as well.  Is it a bug?

Enivex:
This works, but surely there has to be a better way

#let matrix = math.mat(..((1, 2),(3, 4)), delim: "|")
$ matrix $

jgm:
Excellent -- I'm satisfied with that workaround.

Enivex:
`$ mat(#(math.alpha, math.beta), #(3,4), delim: "[") $`

jgm:
Thanks.  My interest is in converting particular formulas for jgm/texmath -- so global `#set` won't work. I'll use this workaround for matrices that have non parentheses delimiters.

laurmaedje:
Forgot to handle named arguments with the semicolon syntax.

PR #5030, opened 2024-09-25
Add `math.stretch` element function

### LINKED ISSUE #1304: Make arrows in math stretchable

mkorje:
Closes #1304. This PR continues @Leedehai's work in #4569.

This PR adds the element function `math.stretch`, which allows you to stretch a glyph horizontally or vertically (only if the glyph is stretchable in the font - this means it has entries in the MathVariants table). It automatically stretches the glyph along the correct axis. You specify a size to stretch the glyph to with the `size` parameter, which is auto or relative (to the glyph's current size). This size parameter defaults to `auto`, which will only have an effect if a top/bottom attachment is attached to `stretch(...)`.

When there are top/bottom attachments present, and when `math.stretch(attach: true)` is set (`true` is the default), the size parameter will instead be relative to the top/bottom attachments' width. Here is also where `auto` comes into play, which will stretch the glyph to the size of the top/bottom attachments.

The example below sums up how it all works.

<table>
<tr>
<td> Code </td> <td> Image </td>
</tr>
<tr>
<td>

```typ

$
  // Stretch vertically.
  stretch(||, size: #2em)
  stretch(arrow.t, size: #200%)
  stretch(angle.r, size: #(200% + 0.5em)) \

  // Stretch horizontally.
  stretch(=, size: #3em) \
  stretch(arrow.hook, size: #3em) \

  // Nested stretch.
  stretch(
    stretch(harpoon.lb, size: #2em),
    size: #200%,
  ) \
  stretch(
    stretch(harpoon.lb, size: #8em),
    size: #50%,
  ) \
  stretch(harpoon.lb, size: #4em) \
  
  // Attachments
  stretch(<-)^"define" quad
  stretch(->)_"define" quad
  // Attachments don't do anything to the 
  // vertically stretched glyphs.
  stretch(arrow.t)^"define" quad
  stretch(arrow.b)_"define" \

  // Relative size with attachments.
  stretch(arrow.hook, size: #150%)_"injective map" \
  stretch(arrow.hook, size: #50%)_"injective map" \
  stretch(arrow.hook)_"injective map" \

  // Absolute size with attachments.
  // Acts the same as if there weren't attachments.
  stretch(arrow.l.r.double, size: #3em)^"text" \
  stretch(arrow.l.r.double, size: #3em) \

  // You can disable sizing being relative to
  // attachments when needed.
  // Or you can always disable it globally with
  // a show rule.
  stretch(->>, size: #200%, attach: #false)_"surjection" \
  stretch(->>, size: #200%)
$
```

</td>
<td>

test

</td>
</tr>
</table>

See the tests for some more examples.

## Implementation

I've been able to avoid adding to the proliferation of fields on the structs in `fragment.rs` by checking if the base of an attachment is a `StretchElem` and then peeking inside.

The `StretchElem` lives in `stretch.rs`. I've moved the `delimiter_alignment` function there from `mod.rs`. I've also changed some of the code in `lr.rs` to avoid some repetition, by relying on the functions in `stretch.rs`.

If a glyph for some reason has both horizontal and vertical variants, I've opted to panic for now. This should probably never happen, and if it does we'd want to know the culprit font...

I also don't know if the name of the parameter `attach` is great. Idk what else to name it though.


tests/ref/math-stretch-attach.png
laurmaedje, line 1: maybe a bit more page width would help here

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,156 @@
...
+    } else {
+        return;
+    };
+
+    let relative_to_size = relative_to.unwrap_or_else(|| {
+        if horizontal {
+            fragment.width()
+        } else {
+            fragment.height()
+        }
```
laurmaedje, line 98: `frame.size().get(axis)` would then also work

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,156 @@
...
+use crate::foundations::{elem, Content, Packed, Resolve, Smart, StyleChain};
+use crate::layout::{Abs, Frame, Length, Point, Rel, Size, VAlignment};
+use crate::math::{
+    GlyphFragment, LayoutMath, MathContext, MathFragment, Scaled, VariantFragment,
+};
 
 /// Maximum number of times extenders can be repeated.
 const MAX_REPEATS: usize = 1024;
 
+/// Stretches a base.
```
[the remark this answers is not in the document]
mkorje: I'm not sure why but I renamed everything from body to base. I'll change it back, since it does make more sense.

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,156 @@
...
+        Ok(())
+    }
+}
+
+/// Attempts to stretch the given fragment by/to the amount given in stretch.
+pub(super) fn stretch_fragment(
+    ctx: &mut MathContext,
+    styles: StyleChain,
+    fragment: &mut MathFragment,
+    axis_horizontal: Option<bool>,
```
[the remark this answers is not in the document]
mkorje: That's probably a good idea. And may as well use it throughout the whole file to make things clearer.

crates/typst/src/math/attach.rs
```diff
@@ -56,7 +62,14 @@ impl LayoutMath for Packed<AttachElem> {
         let new_elem = merge_base(self);
         let elem = new_elem.as_ref().unwrap_or(self);
 
-        let base = ctx.layout_into_fragment(elem.base(), styles)?;
+        // Get the size to stretch the base to, if the attach argument is true.
+        let stretch = elem
+            .base()
+            .to_packed::<StretchElem>()
```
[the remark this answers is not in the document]
mkorje: I'll just make it so `merge_base` will always return something when it gets extracted from an `EquationElem`.

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,156 @@
...
+
+    /// Whether the size should be relative to the width of any top or bottom
+    /// attachments.
+    ///
+    /// ```example
+    /// #set math.stretch(attach: false)
+    /// $ stretch(->, size: #250%)_"very long text" $
+    /// ```
+    #[default(true)]
+    pub attach: bool,
```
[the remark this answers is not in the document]
laurmaedje: I think it being always `true` would be consistent with `lr` though:

- `lr((), size: #200%)`: 200% is 200% of parenthesis because body is empty
- `lr((x/2), size: #200%)`: 200% is more than 200% of parenthesis because body is larger

- `stretch(=, size: #200%)`: 200% is 200% of = because no attachment
- `stretch(=, size: #200%)^"hello"`: 200% is 200% of "hello" because it has an attachment

In both cases, it defaults to the width of the stretchable itself if there is nothing else because that's the whole extent of it, but if there's some other grouped thing (body/attachment) it picks it up. In particular, there is no way in `lr` to specify the height relative to just the parenthesis.

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,151 @@
...
+        MathFragment::Variant(variant) => {
+            GlyphFragment::new(ctx, styles, variant.c, variant.span)
+        }
+        _ => return,
+    };
+
+    let axis = if let Some(stretch_axis) = axis {
+        stretch_axis
+    } else if let Some(stretch_axis) = is_stretchable(ctx, &glyph) {
+        stretch_axis
```
laurmaedje, line 89: looks like a candidate for `or_else`

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,151 @@
...
+    match delimiter {
+        '\u{231c}' | '\u{231d}' => VAlignment::Top,
+        '\u{231e}' | '\u{231f}' => VAlignment::Bottom,
+        _ => VAlignment::Horizon,
+    }
+}
+
+/// Return whether the glyph is stretchable and if it is, along which axis it
+/// can be stretched.
+fn is_stretchable(ctx: &MathContext, base: &GlyphFragment) -> Option<Axis> {
```
laurmaedje, line 124: I think the name should be changed since it doesn't return a boolean. Maybe `stretch_axis`

crates/typst/src/math/attach.rs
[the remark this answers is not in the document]
mkorje: I actually just ended up making another function and peeked inside if its an `EquationElem`, like in `merge_base`. That handles this case. Also added a test for it.

crates/typst/src/math/stretch.rs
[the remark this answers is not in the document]
mkorje: Yeah okay, that's a fair point. I agree then that it should always be `true`.

crates/typst/src/math/stretch.rs
mkorje: Surprised I missed that. Could've sworn I stared for a while at this and the docs to make it nicer aha.

crates/typst/src/math/stretch.rs
```diff
@@ -1,12 +1,139 @@
...
+            GlyphFragment::new(ctx, styles, variant.c, variant.span)
+        }
+        _ => return,
+    };
+
+    let axis = if let Some(stretch_axis) = axis.or_else(|| stretch_axis(ctx, &glyph)) {
+        stretch_axis
+    } else {
+        return;
+    };
```
laurmaedje, line 80:
```suggestion
    let Some(axis) = axis.or_else(|| stretch_axis(ctx, &glyph)) else {
        return;
    };
```

laurmaedje (commented):
found one more thing. otherwise, it looks ready to merge from my end.

crates/typst/src/math/stretch.rs
mkorje: Woops, should've done that.

laurmaedje:
Thank you. Nice work! Also thanks to @Leedehai for getting the ball rolling on this!

mkorje:
`$stretch(|, size: #3em)_a^b$` doesn't work properly. This has also broken #4188 I think. There should be more test cases for that, not too surprising then that it slipped through...

I'm working on fixing this.

mkorje:
I've found an issue I made when implementing this: stretching things vertically doesn't work at all with an attachment.

I'll try to fix this.

Issue #3159, opened 2024-01-09
Documentation for `place()` is a bit misleading


zegkljan:
### Description

This code
```typst
first
#place(top + left, dx: 100pt, rect(width: 30pt, height: 30pt, fill: aqua))
second
third
```
produces this output
![image](https://github.com/typst/typst/assets/6035604/2945626b-977f-4bf8-976b-e5589dca342b)
while it would be expected to produce this output (here achieved by moving the `#place` call at the end)
![image](https://github.com/typst/typst/assets/6035604/0b10e0bd-6c50-4cba-9498-c453e32bd18b)
as per [the documentation](https://typst.app/docs/reference/layout/place/#summary):

> Places content at an absolute position.
> 
> **Placed content will not affect the position of other content.** Place is always relative to its parent container and will be in the foreground of all other content in the container. Page margins will be respected.

Either the behavior or the documentation is therefore incorrect.

Possibly related to #1833

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
The documentation is perhaps a bit misleading here. In general, everything except for a few things (text, inline equations, boxes) is block-level and so is `place`. However, it will not affect other block-level containers.

SolidTux:
This can sometimes be circumvented with `box(place(...))` by the way. I don't know in which cases this breaks, though.

SolidTux:
Thanks a lot!

PR #5031, opened 2024-09-25
Clarify place documentation

knuesel:
This is my attempt at clarifying the `place` documentation, in particular:

- reword/expand "placed content doesn't affect the position of other content": layout is affected when placing from inside a paragraph, or when `float` is `true`. This fixes  https://github.com/typst/typst/issues/3159
- reword "will be in the foreground of all other content in the container": actually, non-placed content added later in flow will come on top
- remove part about ignored alignments for auto-sized axes: it's not true in my tests (but maybe I misunderstood?)
- clarify meaning of "clearance"

Please review carefully as I'm not particularly confident in my understanding of all terms and concepts here.

### DIFF
--- a/crates/typst/src/layout/place.rs
+++ b/crates/typst/src/layout/place.rs
@@ -47,19 +47,24 @@ use crate::layout::{Alignment, Em, Length, Rel};
 /// you can wrap the `place` call in a [`box`] when the call is made
 /// in the middle of a paragraph. The alignment and offsets will then be
 /// relative to this zero-size box. To make sure it doesn't interfere with
-/// spacing, the box should be attached to a word using a zero-width joiner:
+/// spacing, the box should be attached to a word using a zero-width joiner.
+///
+/// For example, the following defines a function for attaching an annotation
+/// to the following word:
 ///
 /// ```example
-/// The first sentence is covered by
-/// the placed element.
-/// #box(
-///   place(
-///     bottom+center,
-///     circle(fill: red),
-///   )
-/// )#(sym.zwj)The second sentence is
-/// shown on top.
+/// #let annotate(..args) = {
+///   box(place(..args))
+///   sym.zwj
+///   h(0pt, weak: true)
+/// }
+///
+/// A placed #annotate(square(), dy: 2pt)
+/// square in my text.
 /// ```
+///
+/// The zero-width weak spacing serves to discard spaces between the function
+/// call and the next word in markup.
 #[elem(scope)]
 pub struct PlaceElem {
     /// Relative to which position in the parent container to place the content.
--- a/crates/typst/src/layout/place.rs
+++ b/crates/typst/src/layout/place.rs
@@ -11,8 +11,8 @@ use crate::layout::{Alignment, Em, Length, Rel};
 /// cases, the content position can be adjusted with [`dx`]($place.dx) and
 /// [`dy`]($place.dy) offsets without affecting the layout.
 ///
-/// The parent can be any container such as a [`block`]($block), [`box`]($box),
-/// [`rect`]($rect), etc. A top level `place` call will place content directly
+/// The parent can be any container such as a [`block`], [`box`],
+/// [`rect`], etc. A top level `place` call will place content directly
 /// in the text area of the current page. This can be used for absolute
 /// positioning on the page: with a `top + left`
 /// [`alignment`]($place.alignment), the offsets `dx` and `dy` will set the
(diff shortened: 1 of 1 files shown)


crates/typst/src/layout/place.rs
```diff
@@ -1,11 +1,23 @@
 use crate::foundations::{elem, scope, Cast, Content, Smart};
 use crate::layout::{Alignment, Em, Length, Rel};
 
-/// Places content at an absolute position.
+/// Places content relatively to its parent container.
```
PgBiel, line 4: It'd be nice to mention the "absolute position" terminology somewhere below. I agree it's not fully correct, but it's what people are used to read, so it's usually easier to associate with behavior from elsewhere. Maybe a short paragraph or sentence somewhere mentioning that "in the absence of a parent container, the content will be placed absolutely, that is, relative to the current page" could help.

PgBiel: I'd also consider adding a link like `[container]($block)`. Though it may be a bit misleading as blocks are not the only containers (though, in practice, everything becomes a block at some point...). But would be a bit better than the status quo, I think.

Could also merge this with my page suggestion and mention "relatively to the parent container, such as an enclosing `[block]($block)`, or, in the absence of one, absolutely, that is, relatively to the current page.", or something similar.

crates/typst/src/layout/place.rs
```diff
@@ -1,11 +1,23 @@
...
+/// cases, the content position can be adjusted with [`dx`]($place.dx) and
+/// [`dy`]($place.dy) offsets without affecting the layout.
+///
+/// # Effect on the position of other elements
+///
+/// Overlaid elements don't take space in the flow of content, but a `place`
+/// call inserts an invisible block-level element in the flow. This can
+/// affect the layout by breaking the current paragraph. To avoid this,
+/// you can wrap the `place` call in a [`box`]($box) when the call is made
+/// in the middle of a paragraph.
```
PgBiel, line 20: Would be nice to add an example of this, especially because a normal example follows right below. It could also go under the normal example.

crates/typst/src/layout/place.rs
```diff
@@ -27,10 +39,6 @@ pub struct PlaceElem {
     ///
     /// - If `float` is `{false}`, then this can be any alignment other than `{auto}`.
     /// - If `float` is `{true}`, then this must be `{auto}`, `{top}`, or `{bottom}`.
-    ///
-    /// When an axis of the page is `{auto}` sized, all alignments relative to
-    /// that axis will be ignored, instead, the item will be placed in the
-    /// origin of the axis.
```
PgBiel, line 33: Any reason for this removal? Is this information outdated? Or is it about the "an axis of the page" bit? We could perhaps replace it with "an axis of the parent container" or similar.

PgBiel (commented):
Nice, thanks! Here are my opinions.

crates/typst/src/layout/place.rs
laurmaedje: `[block]($block)` can be shortened to `[block]` btw (and `` [`block`] `` works as well)

knuesel: When first trying to use `place` I was super confused by this mention of absolute positioning, because the placement is relative to the container, then relative to the alignment in the container, and indeed the function takes offsets `dx` and `dy` rather than absolute positions `x` and `y`, and even in the outer scope you are relative to the text area (i.e. it respects margins) rather than the page itself.

But it's true that `place` is also the way to implement absolute positioning :-) So I added a paragraph about that, together with links to typical containers.

crates/typst/src/layout/place.rs
knuesel: Good point. I moved the example before this section (with a new example added to illustrate absolute positioning, and with changes to the original example to make it more complementary). And now this section also has its own example with the `box` trick.

crates/typst/src/layout/place.rs
knuesel: I guess it's outdated because I cannot reproduce anything like that. For example the following respects the requested alignment for both axes, although they are both `auto`-sized:

```
#rect(width: 8cm, height: 1cm)
#place(horizon+center, [x])
```

By the way I just added another sentence here instead, to mention that the `y` position defaults to the current position if no vertical alignment is specified.

crates/typst/src/layout/place.rs
laurmaedje: I'm not sure whether we should really commend `box(place(..))`. There might be cases where it's the best one can do, but I can't immediately think of such a situation.

Zero-sized boxes don't play well with space trimming in justified paragraphs (if the box ends up at the edge of the line).

knuesel: It's somewhat common though, to want to place something relative to an element inside a paragraph. Could we show another way then? The alternative I can think of is with a label, and a place call  outside of the paragraph, using some calculations with `here()` to get the content in the right position. Basically what `pinit` does. But that's rather involved...

laurmaedje: What's probably okay is to put a zero sized box immediately after or before a word and joined with a ZWJ (zero width joiner). That way they'll stick together.

knuesel: Makes sense. I've updated the text and example accordingly. Alternatively I can remove the whole advice and example, but I think users will keep trying to place things in paragraphs so we'd need to address the issue one way or another...

knuesel:
Thanks for the review, I've addressed all comments. I think we just need to decide if we keep the box+zwj explanation and example.

crates/typst/src/layout/place.rs
laurmaedje: How about this?
```typ
// An inline-placed element that directly attaches
// to the following word.
#let annotate(..args) = {
  box(place(..args))
  sym.zwj
  h(0pt, weak: true)
}

A placed #annotate(square(size: 1cm), dy: 2pt) 
square in my text.
```

image


It's a bit nicer to use via the `h(0pt, weak: true)` trick.

knuesel: Very nice! (also rather dense in tricks: 3 in 3 lines of code :-) ). I've updated the PR to show this example.

laurmaedje:
Thank you! It's a nice addition to the docs.

Issue #2979, opened 2023-12-16
Different parsing in LR longhand vs shorthand: `lr([..])` vs `[..]`


Leedehai:
### Description

The difference becomes apparent if the LR is used in fractions or attachments.

---

With fractions.
```typ
// Note `mid(|)` is necessary to fully expose the differences in appearance.
// Note A and C have the same appearance, but B is "intended" since the
// vertical has the same height as the brackets as intended in math/lr.rs.

$
A. frac(1, lr([limits(sum)_i mid(|)])) quad
B. frac(1, [limits(sum)_i mid(|)]) quad
C. 1/lr([limits(sum)_i mid(|)]) quad
D. 1/[limits(sum)_i mid(|)]
$
```
<img width="520" src="https://github.com/typst/typst/assets/18319900/01783f94-77c0-4e72-a8d4-99e4e5194a29">

Showing the syntax tree:
```typ
// Need to redefine repr(), or it won't work in $ ... $.
#let re(x) = repr(x)

// Note `mid(|)` is necessary to fully expose the differences in appearance.

A. frac-lr (same as C)
$ frac(1, lr([limits(sum)_i mid(|)])) re(frac(1, lr([limits(sum)_i mid(|)]))) $

B. frac-[]
$ frac(1, [limits(sum)_i mid(|)]) re(frac(1, [limits(sum)_i mid(|)])) $

C. /-lr (same as A)
$ 1/lr([limits(sum)_i mid(|)]) re(1/lr([limits(sum)_i mid(|)])) $

D. /-[]
$ 1/[limits(sum)_i mid(|)] re(1/[limits(sum)_i mid(|)]) $
```
|Fraction\LR|`lr([..])`|`[..]`|
|:--|:--|:--|
|`frac(.., ..)`|<img width="280" src="https://github.com/typst/typst/assets/18319900/bfe589b7-8752-49c9-9e6f-9d18da9be599">|<img width="280" src="https://github.com/typst/typst/assets/18319900/1747b76a-0d6b-4cd5-a285-bc0c82f34924"><br>(correct, since the vertical bar has the same<br>height as the brackets, as intended in math/lr.rs)|
|`../..`|<img width="280" src="https://github.com/typst/typst/assets/18319900/116e86ae-09d0-4f97-848d-f3d92439762e">|<img width="280" src="https://github.com/typst/typst/assets/18319900/785dc98d-4439-4b28-89da-e8ad49f89350">|

---

Similar cases if we use attachments instead of fractions:
[...]

Issue #896, opened 2023-04-20
Non-matching delimiters broken in scripts?


Holt59:
The following does not produce the expected result:

```
$ sum_(x in ]0, 1]) x $
```

Output:

![image](https://user-images.githubusercontent.com/2393288/233382911-d539c3d3-6ff1-4774-bf09-1d88c4370cdd.png)

laurmaedje:
In principle, this works as intended. In math, any kind of opening bracket matches any closing one (instead of just similar ones). Otherwise, it would become more difficult to write ranges like `[a,b)`. In your case, it would be best to just escape the square brackets.

Holt59:
> In principle, this works as intended. In math, any kind of opening bracket matches any closing one (instead of just similar ones). Otherwise, it would become more difficult to write ranges like `[a,b)`. In your case, it would be best to just escape the square brackets.

The `(` and `)` are here to wrap the script content, not to be displayed, if that's intended, that's pretty weird...

laurmaedje:
> The ( and ) are here to wrap the script content, not to be displayed, so they should not be matched to anything except their closing part. This looks more like a parsing issue rather than something intended.

Yes, you are right.

laurmaedje:
Actually, I think it's not that simple. Because of the arbitrary matching, we can have situations like this:
```typ
$ sum_(x in [0, 1), 1, 2, 3) x $
```
If we were to always respect the closing parentheses, this would render inconsistently.

PR #5050, opened 2024-09-27
Fix `math.lr` size argument not being applied to enclosed `math.mid`

### LINKED ISSUE #3186: `lr()`'s size argument can't consistently be applied to enclosed `mid()`
### Description

```typ
$
  lr({ A mid(|) integral }) quad
  lr(size: #1em, { A mid(|) integral }) quad
  lr(size: #(1em+20%), { A mid(|) integral }) "wrong" \

  lr(] A mid(|) integral ]) quad
  lr(size: #1em, ] A mid(|) integral ]) quad
  lr(size: #(1em+20%), ] A mid(|) integral ]) "ok" \

  lr(( A mid(|) integral ]) quad
  lr(size: #1em, ( A mid(|) integral ]) quad
  lr(size: #(1em+20%), ( A mid(|) integral ]) "wrong"
$
```
Screenshot 2024-01-12 at 20 28 44

### Typst version

mkorje:
Closes #3186.

I also made it so that the contents of an `EquationElem` are extracted.

See the tests for examples.


mkorje:
It appears this has also fixed all but one of the cases in #2979 as well. Case D given in that issue is a parsing problem, somewhat similar to #896 (in that delimiters wrapping attachments and fractions are stuffed).

laurmaedje:
Does it also fix the problem when there is a comma present in the body?

mkorje:
> Does it also fix the problem when there is a comma present in the body?

Yep. Including the comma hasn't been changed at all.

laurmaedje:
Because the body `#[parse]` code would still produce a sequence in that case. I'm a bit confused what the exact idea behind the change of that parsing code is.

mkorje:
For a single element as the body, for example the automatic `LrElem` created as the body, it would previously be parsed as a sequence with one element. This makes it so that a single element isn't parsed as a sequence.

mkorje:
It was never an issue with a comma, I believe, since it can't have been made into a LrElem implicitly then.

laurmaedje:
Ah!

laurmaedje:
Thanks!

PR #5051, opened 2024-09-27
Fix numbering documentation

emilyyyylime:
Reformatted long lines, added missing mentions to circled numerals, replaced special double quotes with ascii quotes in internal docs, and a small reword

### DIFF
--- a/crates/typst/src/model/numbering.rs
+++ b/crates/typst/src/model/numbering.rs
@@ -58,13 +58,13 @@ pub fn numbering(
     context: Tracked<Context>,
     /// Defines how the numbering works.
     ///
-    /// **Counting symbols** are `1`, `a`, `A`, `i`, `I`, `一`, `壹`, `あ`, `い`, `ア`, `イ`, `א`, `가`,
-    /// `ㄱ`, `*`, `①`, and `⓵`. They are replaced by the number in the sequence, preserving the original
-    /// case.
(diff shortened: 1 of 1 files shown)


laurmaedje:
While you are at it, could you wrap the lines at 80 columns?

emilyyyylime:
Sure thing. I thought you were trying for 100 so that's where I wrapped them

emilyyyylime:
Hm. Does 80 columns mean 80 glyphs in this case? In my setup at least many of the foreign characters appear as double-width. I'm leaning towards counting them as a single column because they could theoretically be displayed that way

laurmaedje:
80 columns for me means whatever happens with the VS Code rewrap extension and a ruler at 80 😅

I would say probably glyphs. For me, most stuff appears single column.

emilyyyylime:
Also it seems rustfmt formats the long arrays at 80 characters too (even though the maximum width is 100 :thinking:)

laurmaedje:
Max width should be 90 for rustfmt

emilyyyylime:
Ah, in that case looks like it actually does consider those characters as double-wide

emilyyyylime:
The GitHub 'changes' view just falls back to a proportional font for those characters (at least on Firefox)

laurmaedje:
Thanks!

PR #4958, opened 2024-09-14
Miscellaneous symbols fixes

MDLC01:
This PR includes multiple fixes for symbols. They are listed below.

- `angle.l.double` and `angle.r.double` used to use CJK compatibility characters. They now use mathematical characters.

- **Breaking change.** As per [the discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1284465091437269035), I removed `degree.c`, `degree.f`, `ohm`, `angstrom`, and `kelvin`.

- I renamed `ohm.inv` to `mho` (this was also discussed in the aforementioned Discord discussion).

- I renamed `parallel.slanted.*` to `parallel.slant.*`. Those symbols were added in #4834. Using `slanted` instead of `slant`, like `lt.eq.slant` and `gt.eq.slant`, was an oversight on my part.

- As per [the discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1284579940800659468), I renamed `emptyset.rarrow` and `emptyset.larrow` to `emptyset.arrow.r` and `emptyset.arrow.l` (same for `nothing`).

- I fixed an indentation issue.

- I added a comment about #3375.

PR #5053, opened 2024-09-27
Implement symbol Proposal 21 (iteration 1)

MDLC01:
This PR implements Proposal 21 (iteration 1) from [the Symbols document](https://typst.app/project/riXtMSim5zLCo7DWngIFbT).

This was accepted by @reknih [on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1284579940800659468) some time ago and originally introduced in #4958. Merging this before 0.12 does not cause any breaking change, whereas merging it after 0.12 would be a breaking change.

### DIFF
--- a/crates/typst/src/symbols/sym.rs
+++ b/crates/typst/src/symbols/sym.rs
@@ -341,18 +341,18 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     // Set theory.
     emptyset: [
         '∅',
+        arrow.r: '⦳',
+        arror.l: '⦴',
         bar: '⦱',
         circle: '⦲',
-        larrow: '⦴',
-        rarrow: '⦳',
         rev: '⦰',
     ],
     nothing: [
         '∅',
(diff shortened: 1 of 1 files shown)


crates/typst/src/symbols/sym.rs
```diff
@@ -341,18 +341,18 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     // Set theory.
     emptyset: [
         '∅',
+        arrow.r: '⦳',
+        arror.l: '⦴',
```
elegaanz, line 345: Sorry for the late review, but I think there is a typo here (and below as well), isn't it supposed to be `arrow` and not `arror`?

MDLC01: You're right. This is annoying. I'll make another PR to fix that, unless this can be included in something else.

MDLC01: Fixed by #5055.

PR #5058, opened 2024-09-27
Decoration: baseline and size correctly estimated by surrounding text

### LINKED ISSUE #3391: Highlight Offset Mismatch with Subscript and Superscript Text

f3fora:
Fix #3391

The goal of this PR is to correct the calculation of the highlight box's offset when working with text that has a non-zero baseline.

This PR does not address math.

We should note that PRs #3953 and #3941 simply rename the function `decorate` to `decorate_shaped_text` and do not address issue.


laurmaedje:
This is an interesting approach. Basically fixing the size where it's called. To summarize with a quick before/after:

<table>
<thead>
<td>Code</td>
<td>Before</td>
<td>After</td>
</thead>
<tr>
<td>

```typ
#set sub(typographic: false)

He#highlight[l#sub[1]llo]you!

#highlight[Hello #text(20pt)[world], what's up?]

#highlight[Hello #text(5pt)[world], what's up?]
```

</td>
<td>

image

</td>
<td>

image

</td>
</tr>
<tr>
<td>

```typ
#set sub(typographic: false)
#show: highlight

= Introduction
#lorem(10)

#text(5pt)[I am small.]
```

</td>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

In most cases, the output is probably more desirable (because `highlight` is typically used for short segments of text). However, if a large amount of content is highlight, it is worse in my opinion (second example).

In summary, I'm not 100% sure whether this is the best fix. It's probably better than the status quo, but an ideal fix would, in my view, produce a bounding box for consecutive segments of inline items (text, math, etc.) and use that. Basically what I [was proposing](https://github.com/typst/typst/pull/3953#issuecomment-2137709437) in one of the linked PRs.

laurmaedje:
I'm quite convinced by now that producing a bounding box for consecutive inline segments (text, math, probably even boxes) is the way to go. I'll thus close this. Thank you still!

PR #5059, opened 2024-09-28
Remove `Case`s from NumberingPatterns

emilyyyylime:
- **Remove `Case`s from NumberingPatterns, extract long functions**
- **Switch `zeroless` to take a character array directly**
- **Reorder `NumberingKind::apply` implementation by similarity**

Overall I find this approach much more elegant. I also sprinkled in some other refactorings to improve the overall organisation of the code and hopefully make it easier for contributors to add new `NumberingKind`s in the future


laurmaedje:
This indeed looks much better. Thanks!

PR #5069, opened 2024-09-30
Fix `math.stretch` not working vertically with attachments

mkorje:
Fixes a bug I made when implementing #5030: stretching things vertically doesn't work at all with an attachment. Instead of attempting to always stretch along the given `axis` in `stretch_fragment`, we now instead check if it equals what is returned from `stretch_axis`, returning if it doesn't.

The seemingly unrelated test `issue-4188-lr-corner-brackets` being updated happens since with the changes described above, the corner glyphs aren't stretchable and so we return early. Thus, we don't align them. But we don't need to anyways as they are already aligned in the font. In fact, in the old test image, the two expressions aren't actually the same in the image. The corners in `$⌜a⌟⌞b⌝$` are higher/lower than in `$⌜$$a$$⌟$$⌞$$b$$⌝$`. This has to do with the fact that aligning it to the top/bottom of the frame does not match its positioning as is. The behaviour with corner brackets should be clarified/resolved nicely when solving #5071.


crates/typst/src/math/stretch.rs
```diff
@@ -73,9 +73,15 @@ pub(super) fn stretch_fragment(
...
-    let Some(axis) = axis.or_else(|| stretch_axis(ctx, &glyph)) else {
+    // Return if we attempt to stretch along an axis which isn't stretchable,
+    // so that the original fragment isn't modified.
+    let Some(stretch_axis) = stretch_axis(ctx, &glyph) else {
         return;
     };
+    if axis.is_some_and(|x| x != stretch_axis) {
+        return;
+    }
+    let axis = axis.unwrap_or(stretch_axis);
```
mkorje, line 84: There's probably a nicer way to structure this bit that I can't think of right now.

isuffix: `axis != Some(stretch_axis)` ? Maybe one big match statement

wrzian: `axis != Some(stretch_axis)` ? Maybe one big match statement

mkorje: It shouldn't return if axis is `None` though. A match might be nicer.

mkorje: yeah that works.

laurmaedje:
Thanks!

laurmaedje:
Conflicts due to #4969

mkorje:
I've fixed the conflict and rebased.

laurmaedje:
Thanks!

PR #5075, opened 2024-09-30
Minimal PDF/A support

### LINKED ISSUE #2995: PDF/A: 6.1.13 Implementation limits
PDF/A-2 imposes implementation limits on integers, floats, strings, names, indirect object count, q/Q pairs in content streams, numbers of colorants in a DeviceN space and CID values. Export to PDF/A needs to fail if any of these limits is exceeded.

Some of these are enforced by `pdf-writer` through the type system. We always explicitly write DeviceN colorants (never > 4) and the number of CID values depends on the glyph count in the font (OpenType fonts never contain more than 65535 glyphs).

However, the following could arise through user input and need to trigger an alert at the export stage.

- Strings longer than 32767 bytes
- Names longer than 127 bytes (e.g. when containing font name)
- More than 8388607 indirect objects
- More than 28 q/Q nesting level

This depends on #2943. However, inroads here could be made by tracking e.g. nesting level and indirect object count without reporting in a first PR.

### LINKED ISSUE #3002: PDF/A: 6.2.3 Output Intent
PDF/A requires an output intent with an ICC profile if a device color space is used and the respective `Default` color space is not set (cf. PDF 1.7, 8.6.5.6). Typst writes device color spaces in the following situations:

- When using a CMYK color
- When writing an image with an alpha mask (DeviceGray)

For alpha masks, we should do the right thing. (Is there such a thing as an alpha image profile? Is gamma applied here?)

For CMYK, things are more complicated:

**Quick fix:** \
Fail PDF export if PDF/A compliance is requested and we find DeviceCMYK colors.

**Principled solution:** \
CMYK colors have little meaning without an output condition. We should allow CMYK output colors in PDF/A only if an output intent is set. This output intent should be configurable in the source text, as the CMYK values without it are ambiguous. A color management with a library like [qcms](https://github.com/FirefoxGraphics/qcms) is needed to provide the right conversions for  RGB in the SVG exporter. This would also improve color accuracy for CMYK in the preview.

### LINKED ISSUE #3004: PDF/A: 6.2.11.8 `.notdef` glyph
Typst outputs the `.notdef` glyph (tofu) if font fallback failed for a particular code point, cf. issue #2818. This happens especially if `#set text(fallback: false)` is set. The `.notdef` glyph is not valid in PDF/A. A font fallback failure should yield a warning in normal PDF export and a failure in PDF/A export.

laurmaedje:
This pull requests adds support for optional conformance to the PDF/A-2b standard. This is realized through a new `--pdf-standard` CLI option which takes one (or multiple comma-separated) PDF standards that Typst should try to conform with. This list is validated for compatibility (though the current two options (1.7, a-2b) are simply always compatible). Support for more standards is planned for the future.

Closes #2995
Closes #3002
Closes #3004

### DIFF
--- a/crates/typst/src/visualize/color.rs
+++ b/crates/typst/src/visualize/color.rs
@@ -544,6 +544,9 @@ impl Color {
     /// These components are also available using the
     /// [`components`]($color.components) method.
     ///
+    /// Note that CMYK colors are not currently supported when PDF/A output is
+    /// enabled.
+    ///
     /// ```example
     /// #square(
     ///   fill: cmyk(27%, 0%, 3%, 5%)


crates/typst-pdf/src/color.rs
```diff
@@ -65,18 +66,18 @@ impl ColorSpaces {
     /// PDF file.
     pub fn write_functions(&self, chunk: &mut Chunk, refs: &ColorFunctionRefs) {
         // Write the sRGB color space.
-        if self.use_srgb {
+        if let Some(id) = refs.srgb {
```
laurmaedje, line 69: @elegaanz does this change look ok?

LaurenzV:
Do you think there is really value in allowing selecting the PDF version as well? I think the only thing users might care about is choosing the right standard (PDF/A, PDF/UA, etc.), which themselves always dictate which PDF version must be used. And removing that would make it impossible to accidentally choose a wrong combination.

Maybe there is value in keeping it, but just thought I might ask. 🙂

laurmaedje:
Wrong combinations would result in an error. The idea here was to allow opt in PDF 2.0 support down the road. We wouldn't add 1.6 or lower.

crates/typst-pdf/src/color.rs
elegaanz: Yes, I don't think it should change the final result (if the point was just to avoid `unwrap`ing).

PR #5078, opened 2024-10-01
Fix breakable block equation overflow

### LINKED ISSUE #5025: Breakable block equations can overflow
### Description

_A regression introduced with #5017_

A breakable block equation can overflow outside the bounds of its parent container. For example, the code
```typ
#set page(width: 3cm, height: 2cm, margin: 3.5mm)

This equation overflows:
$ a + b $
```
results in the following output, where the left is before the regression and the right is after:

image

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

mkorje:
Fixes #5025.

Previously, single line equations (which could overflow) were not ran through the `can_break` branch of `let equation_builders = ...`. Removing `full_equation_builder.frames.len() > 1` from `can_break` ensures the multi-layouter produces frames that correspond 1-1 to regions. 

I can't quite figure out why it was added in #4836.


laurmaedje:
> I can't quite figure out why it was added in https://github.com/typst/typst/pull/4836.

Maybe the idea was just to save unnecessary work. 

cc @EpicEricEE

laurmaedje:
Thanks!

EpicEricEE:
> Maybe the idea was just to save unnecessary work.

It was probably that.

I'm wondering though, why did it work for non-breakable equations? They also didn't run through the `can_break` branch. Does the multi layouter change its behavior if `breakable` is false?

Issue #5042, opened 2024-09-26
'Could not infer output format' errors are too lazy


emilyyyylime:
### Description

I went to check if something compiles by simply running `typst c - /dev/null`. Typst started up normally and I wrote a small document but then Typst stopped with the error
```
error: could not infer output format for path /dev/null.
consider providing the format manually with `--format/-f`
```
1. Typst should eagerly emit this error before even attempting to read input
2. Some built-in way to specify that no export is required (`--format none`?) would be nice.

- [X] I am using the latest version of Typst

emilyyyylime:
If you like the suggestion in point 2, I'd like to have a go at implementing it myself

emilyyyylime:
Although this would become pretty irrelevant with some form of repl

laurmaedje:
The idea of `--format none` being to just check for errors?

emilyyyylime:
yeah, or maybe a dedicated `typst check` or `--dry-run` etc.

laurmaedje:
Closed by https://github.com/typst/typst/pull/5080. The just checking part is covered by https://github.com/typst/typst/issues/3792.

PR #5080, opened 2024-10-01
Check that an output format is available before compiling

emilyyyylime:
Resolves the main issue in #5042.

It's important to place the call to `.output_format()` before initialising the World as it can cause a noticeable delay before displaying the error.
Dropping the OutputFormat only to recalculate it moments later is a little goofy, but it's a very cheap calculation and the visual overhead of passing yet another argument around justifies the cost I think.


laurmaedje:
It does look goofy indeed. Ideally, we'd have a more type-driven validation process, but I guess this is ok for now.

PR #5093, opened 2024-10-02
Adjust line number clearance

PgBiel:
1. The default line number clearance (distance from text) is now 2.5em, in order to scale with the font size. This is slightly lower than the previous default of 1cm with the default font settings.
2. Changed line number settings for doc examples. Let me know if they need further adjustments.


Enivex:
Wouldn't it make more sense to base the default on percentage of page width, just like the default margin? That way they won't end up outside the page for small page size.

PgBiel:
I think I'd prefer that, but that would lead to zero clearance on auto width pages... I wonder if there's a better compromise.

PgBiel:
Actually, margins seem to just default to A4 margins in that case. Although I've tried to access page properties from within the flow before, where line numbers are generated, and it was a huge hassle. Do you think this is a good idea, and is this even viable @laurmaedje ?

laurmaedje:
Dicussed with reknih, solution is obviously `calc.min(calc.max(.75em.to-absolute(), 2.6% * page.width), 2.5em.to-absolute())`. I'll push something.

PR #5099, opened 2024-10-02
Fix default ignorables

### LINKED ISSUE #5091: Whitespaces became [U+202A]
### Description

Steps to reproduce:

1. Generate pdf and open pdf
2. Copy text
3. Paste in VS Code
4. Whitespaces either become [U+202A] or the [U+202A] character is between the whitespaces.

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

### LINKED ISSUE #5095: Add default typst implementation code where possible
### Description

I recently [was looking](https://forum.typst.app/t/how-do-i-know-what-elements-compose-a-function-for-styling-purposes/822) at how I could know what elements to style to change the appearance of some other element (in my example, `quote`). In that case, there is no way to do what I was suggesting with `set` rules, so the only option is a `show` rule.

However, working out how to replicate the existing appearance, except for some small tweak, either requires trial and error or reading through the Rust implementation code.

Could a "null"/identity show rule be included in the documentation for every element showing how to style it in typst as it is styled in Rust? For example, for a quote, that might look something like this
```typst
#show quote: it => {
  let body = it.body
  if (it.quotes == auto and not it.block) or it.quotes == true {
    body = ["] + body + ["]
  }

  if it.block {
    body = block(body)
    if it.attribution != none {
      body += align(end, [— ] + it.attribution)
    }
    body = pad(body)
  }

  body
}
```
(this isn't perfectly accurate as the between the quote and the attribution isn't quite right).

So some element
```typst
#let el = quote(attribution: [Shakespeare], [To be, or not to be, that is the question], block: true)
```
before and after the `show` rule should be visually identical.

laurmaedje:
Since https://github.com/typst/typst/pull/4585, we use the text that glyphs were shaped from for the PDF CMap. However, this creates an undesirable effect in combination with default ignorable codepoints such as BiDi overrides or zero-width joiners.

Consider this code snippet:
```typ
A#sym.zwj B C
```

Harfbuzz will produce an empty space glyph with zero advance width for the ZWJ and put it into one cluster with the "A". As a result, Typst sees two glyphs within the cluster range 0..4. Because of this clustering, it assigns the range 0..0 to A and 0..4 to the ZWJ. As a result, the CMap entry in the PDF for the space glyphs becomes totally wrong.

What this PR does:
- Enable `REMOVE_DEFAULT_IGNORABLES` in Harfbuzz. This way, we don't get zero-advance space glyphs at all anymore. I assume that this is not the default because of GUI apps, where cursor movement shall be affected. @behdad Is this correct or does this have other relevant effects?
- Filter out default ignorables when creating CMap entries for the PDF. Text extraction can still fail, to truly fix it we need `/ActualText` support.
- Ignores default ignorables when scanning for segments that need font fallback. We currently shape full segments of text, even containing linebreaks, as one (but other segment properties are ensured to be uniform). With this change, default ignorables can be soaked up into the cluster of a newline's tofu, which broke a check. I think shaping newlines is probably wrong and we should maybe rather split shape runs at newlines? But that's left as future work.

Fixes #5091

### DIFF
--- a/crates/typst-pdf/src/content.rs
+++ b/crates/typst-pdf/src/content.rs
@@ -425,7 +425,7 @@ fn write_text(ctx: &mut Builder, pos: Point, text: &TextItem) -> SourceResult<()
                 bail!(
                     g.span.0,
                     "the text {} could not be displayed with any font",
-                    text.text[g.range()].repr()
+                    TextItemView::full(text).glyph_text(g).repr(),
                 );
             }
         }
@@ -493,9 +493,7 @@ fn write_normal_text(
 
     let glyph_set = ctx.resources.glyph_sets.entry(text.item.font.clone()).or_default();
     for g in text.glyphs() {
-        let t = text.text();
-        let segment = &t[g.range()];
-        glyph_set.entry(g.id).or_insert_with(|| segment.into());
+        glyph_set.entry(g.id).or_insert_with(|| text.glyph_text(&g));
     }
 
     let fill_transform = ctx.state.transforms(Size::zero(), pos);
@@ -640,9 +638,7 @@ fn write_color_glyphs(
 
         ctx.content.show(Str(&[index]));
 
-        glyph_set
-            .entry(glyph.id)
-            .or_insert_with(|| text.text()[glyph.range()].into());
+        glyph_set.entry(glyph.id).or_insert_with(|| text.glyph_text(&glyph));
     }
     ctx.content.end_text();
 
--- a/crates/typst-pdf/src/content.rs
+++ b/crates/typst-pdf/src/content.rs
@@ -441,7 +441,7 @@ fn write_text(ctx: &mut Builder, pos: Point, text: &TextItem) -> SourceResult<()
         || tables.svg.is_some()
         || tables.colr.is_some();
     if !has_color_glyphs {
-        write_normal_text(ctx, pos, TextItemView::all_of(text))?;
+        write_normal_text(ctx, pos, TextItemView::full(text))?;
         return Ok(());
     }
 
@@ -449,9 +449,9 @@ fn write_text(ctx: &mut Builder, pos: Point, text: &TextItem) -> SourceResult<()
         text.glyphs.iter().filter(|g| is_color_glyph(&text.font, g)).count();
 
     if color_glyph_count == text.glyphs.len() {
-        write_color_glyphs(ctx, pos, TextItemView::all_of(text))?;
+        write_color_glyphs(ctx, pos, TextItemView::full(text))?;
     } else if color_glyph_count == 0 {
-        write_normal_text(ctx, pos, TextItemView::all_of(text))?;
+        write_normal_text(ctx, pos, TextItemView::full(text))?;
     } else {
         // Otherwise we need to split it in smaller text runs
         let mut offset = 0;
--- a/crates/typst/src/layout/inline/collect.rs
+++ b/crates/typst/src/layout/inline/collect.rs
@@ -8,7 +8,8 @@ use crate::layout::{
 };
 use crate::syntax::Span;
 use crate::text::{
-    LinebreakElem, SmartQuoteElem, SmartQuoter, SmartQuotes, SpaceElem, TextElem,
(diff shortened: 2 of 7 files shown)


MDLC01:
I'm not sure how this fixes #5095. Did you mean #5091?

laurmaedje:
@LaurenzV noted on Discord that we should maybe also set the Harfbuzz buffer's [cluster level](https://harfbuzz.github.io/working-with-harfbuzz-clusters.html) to 1 to prevent unnecessary merging, as recommended in the Harfbuzz docs. I will try that later.

laurmaedje:
Result: Setting the cluster level to 1 results in some regressions.

- Font fallback does not work correctly anymore for composed emoji. For example `01️⃣2` will render as `01` followed by an empty gray box followed by `2` because with the default font the 1 and the tofu don't share the same cluster value.

- Tracking (character spacing) is broken because is applied per cluster and thus we get double tracking for accented glyphs. 

Based on this, I won't change the level for now. Still curious though what the benefits are the Harfbuzz docs talk about:

> Level 1 tweaks the old behavior slightly to produce better results. Therefore, level 1 clustering is recommended for code that is not required to implement backward compatibility with the old HarfBuzz.

laurmaedje:
Thanks for chiming in!

> ... have a role and dropping them changes the meaning or the rendering of the text.

Can't that also be desirable sometimes? For example, if I use a ZWJ in Typst to ensure that no linebreak occurs, then maybe I just wanted to affect the line breaking without changing the meaning of the text,. Then, it would be desirable to copy-paste without the ZWJ. I might as well have used a Typst function for this, which also wouldn't have an effect on the copyable text.

I'm not sure how to draw the line for such cases.

> Similarly, you should be using Unicode grapheme clusters here.

Thanks, that makes sense. Do you mind expanding on what benefits cluster level 1 offers over 0?

laurmaedje:
I guess using a ZWJ for preventing a line break is a misuse. I confused it with the word joiner. But that is also default ignorable, so I guess the point still stands. 

Anyway, for now, I've decided to [only trim exterior default ignorables](https://github.com/typst/typst/pull/5105), not interior ones (between other chars of the glyph). This fixes copy-paste of composite emojis. It's obviously a hack and we'll need to revisit this topic, but not in a rush before the release.

khaledhosny:
> Can't that also be desirable sometimes? For example, if I use a ZWJ in Typst to ensure that no linebreak occurs, then maybe I just wanted to affect the line breaking without changing the meaning of the text,.

That is second guessing, and it is not something I find myself comfortable with. You have no idea what the user wants here, and guessing one way or another will be wrong for one group of users. If something is in the input text, IMHO, it should be preserved when text is extracted.

> Then, it would be desirable to copy-paste without the ZWJ. I might as well have used a Typst function for this, which also wouldn't have an effect on the copyable text.

That is a clear distinction. If you want to change the layout, you use layout-changing functionality not control chars unless you also want to preserve that in output.

> Thanks, that makes sense. Do you mind expanding on what benefits cluster level 1 offers over 0?

It allows addressing combining marks individually, if the font supports it. For example if you want to color combining marks (coloring Arabic vowel marks is an often requested feature foe pedagogical uses), with cluster level 0 the mark and the base are the same cluster so you can’t tell which glyph corresponds to which (without heuristics). I don’t know if Typst allow coloring parts of the text without breaking shaping, though.

This might be relevant to PDF text extraction as well, since with level 0 the base and the mark share the same cluster, so how do you map each glyph in CMAP if they have the same cluster ?

behdad:
The short answer is that you need `/ActualText` support. If the extracted text eg. doesn't have ZWNJ, that can alter the text meaning for example in Persian.

laurmaedje:
Alright, thanks to both of you for the explanations!

behdad:
> Alright, thanks to both of you for the explanations!

You're welcome.

Basically, like Khaled also said, cluster level 0 does an approximate Unicode Grapheme Boundary, for lazy clients. Proper clients should try to use cluster level 1 and do their own Unicode Grapheme Boundary for text selection, cursor movement, font fallback, and PDF text extraction.

PR #5116, opened 2024-10-04
Fix panic for empty breakable block equations

### LINKED ISSUE #5113: Panic for empty breakable block equations
### Description

The equate testing suite found another crash with breakable block equations. This time, it panics when the equation body is none or an empty sequence:

```typ
#math.equation(block: true, [])
```

```
thread 'main' panicked at crates\typst\src\layout\flow\collect.rs:396:35:
```

I think it's again the contract of the multi layouter being broken, as an empty equation leads to no frames but one region 😟

### Reproduction URL

_No response_

### Operating system

Web app, Windows

### Typst version

EpicEricEE:
Fixes #5113 

An empty equation now returns an empty frame instead of no frame at all, so that the 1-1 region to frame contract is maintained.

(Something like this may have been the real reason why I did https://github.com/typst/typst/pull/5078#issuecomment-2384865773 initially, but I should have documented that somewhere.)

### DIFF
--- a/crates/typst/src/math/equation.rs
+++ b/crates/typst/src/math/equation.rs
@@ -379,6 +379,12 @@ fn layout_equation_block(
             equation_builder.size.y = height;
         }
 
+        // Ensure that there is at least one frame, even for empty equations.
+        if equation_builders.is_empty() {
+            equation_builders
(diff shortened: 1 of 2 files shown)


laurmaedje:
Thanks!

PR #5122, opened 2024-10-04
Fix "thumbnail" key in manifest files not being optional

arbrauns:
As per [the `packages` repository's README](https://github.com/typst/packages/blob/2b56fa0f972c616b1481c4572e897f3c9782f853/README.md), the key is only required when uploading to the official registry.

This otherwise creates a regression in v0.12.0-rc1.


laurmaedje:
Thanks!

PR #5126, opened 2024-10-04
Add exception for NewCM Math Bold

EpicEricEE:
~Not sure why I needed to specify the weight, as it's correct in the font. If I don't specify it though, the weights all get confused and regular becomes bold, bold becomes book, and book becomes regular.~
Nvm, I inspected the wrong font file, it is indeed not correct in the font.

Test changed because equations in strong text now use the bold math font.


Andrew15-5:
With this change, you no longer need to explicitly set the font for the bold math:

```typ
#set page(paper: "a6")
#show math.equation: it => { show strong: set text(font: "NewComputerModernMath"); it }
#set math.equation(numbering: (..n) => strong(numbering("(1)", ..n)))
$ y = 2x $
```

Just:

```typ
#set page(paper: "a6")
#set math.equation(numbering: (..n) => strong(numbering("(1)", ..n)))
$ y = 2x $
```

It also would throw a warning in case the show rule isn't removed:

```typ
warning: unknown font family: newcomputermodernmath
  ┌─ math.typ:2:57
  │
2 │ #show math.equation: it => { show strong: set text(font: "NewComputerModernMath"); it }
  │                                                          ^^^^^^^^^^^^^^^^^^^^^^^
```

laurmaedje:
Thanks!

PR #5127, opened 2024-10-05
Fix typo in error message

samueltardieu:


samueltardieu:
Was on an old version

PR #5135, opened 2024-10-05
Increase page height in place example

knuesel:
Small follow-up to #5031: this example was meant to produce a single page, and I see on https://staging.typst.app/docs/reference/layout/place/#examples that 60pt is too small for that. 120pt is enough in local testing, I hope it would work for the HTML docs too (there seem to be some differences).

Also maybe the `set page` line should be hidden using `>>>`?


laurmaedje:
I combined all docs PRs into https://github.com/typst/typst/pull/5147 to prevent conflicts and too much time spent on discussions and review back and forth. Thanks!

Issue #3765, opened 2024-03-22
Documentation for CLI


3052:
where are the docs for the command line usage? I searched this one google:

site:typst.app "compile" "watch"

and found nothing

laurmaedje:
There are currently no dedicated CLI docs beyond the output of `typst --help`. We plan to add some in the future.

3052:
this was pretty helpful:

https://github.com/typst/typst#usage

even just copy pasting that to https://typst.app/ would be helpful

theyarelulu:
Came here for this. Everything in the website is talking about the web app. Not everyone wants to use the web app (e.g. sensitive documents). This rubbed me the wrong way. I have never seen an open-source project focus on their hosted version instead of how to run the app locally. Upon reading further about Typst, it turns out it is a business, which was not immediately clear in the website or repository. Still, as a business, you can make profit without frustrating users who do not want to sign up to your website.

laurmaedje:
@theyarelulu There is currently unfortunately neither proper documentation for the CLI nor for the web app, so that's a bit besides the point in my opinion.

Nonetheless, let me copy what I wrote about that topic on Discord recently:

> It's not actually the case that we aim to hide this [that there is a CLI]. The reason is simply that this website design stems from a time where Typst was not even open source yet. Of course, you could say "you had enough time to update it" and that's true. But we want to update it in full instead of making small tweaks to an already dated design. That full refresh always felt like it's around the corner, but in the end we were always too busy. Now we're at the point where a completely new website is fully designed, including a full and prominently linked open source page with a download section. And it's in the process of being implemented. It just took way longer that anticipated. The "View on GitHub" was already a hot fix that replaced a "Sign in" button that used to be there to make the open-source-ness a bit more prominent. We don't believe that hiding the CLI or hiding Tinymist or anything like that helps us in any way. Our mantra is "more Typst = more good", no matter where.

YDX-2147483647:
Recent updates:

- [typst(1) — Arch manual pages](https://man.archlinux.org/man/extra/typst/typst.1.en)

  `typst --help` hosted online.

- https://typst-cli-docs.vercel.app/ (GitHub: https://github.com/Ayanprogrammer11/typst-cli-docs)

  A lightweight, static documentation site for the Typst command-line tooling.

PR #5136, opened 2024-10-05
Polished docs for 0.12

Andrew15-5:
I created a PR with all the polished and fixed stuff (plus nitpicks and questions/proposals) as [was advised](https://discord.com/channels/1054443721975922748/1088371867913572452/1291828869145559113).

There are quite a few things, some I have strong opinions on, some — just pointing out in case it was overlooked or something.

Should I instead use the last `main` commit as the base instead of the new tag?

### DIFF
--- a/crates/typst/src/foundations/decimal.rs
+++ b/crates/typst/src/foundations/decimal.rs
@@ -29,7 +29,7 @@ use crate::World;
 /// raised).
 ///
 /// You can also convert any [integer]($int) to a decimal with the
-/// `{decimal(int)}` constructor, e.g. `{decimal(59)}`. However, note that
+/// `{decimal(int)}` constructor, e.g. `{decimal(69)}`. However, note that
 /// constructing a decimal from a [floating-point number]($float), while
 /// supported, **is an imprecise conversion and therefore discouraged.** A
 /// warning will be raised if Typst detects that there was an accidental `float`
@@ -80,11 +80,11 @@ use crate::World;
 /// multiplication, and [power]($calc.pow) to an integer, will be highly precise
 /// due to their fixed-point representation. Note, however, that multiplication
 /// and division may not preserve all digits in some edge cases: while they are
-/// considered precise, digits past the limits specified below are rounded off
+/// considered precise, digits past the limits specified below (where?) are rounded off
 /// and lost, so some loss of precision beyond the maximum representable digits
 /// is possible. Note that this behavior can be observed not only when dividing,
 /// but also when multiplying by numbers between 0 and 1, as both operations can
-/// push a number's fractional digits beyond the limits described below, leading
+/// push a number's fractional digits beyond the limits described below (where?), leading
 /// to rounding. When those two operations do not surpass the digit limits, they
 /// are fully precise.
 #[ty(scope, cast)]
--- a/crates/typst/src/foundations/decimal.rs
+++ b/crates/typst/src/foundations/decimal.rs
@@ -29,7 +29,7 @@ use crate::World;
 /// raised).
 ///
 /// You can also convert any [integer]($int) to a decimal with the
-/// `{decimal(int)}` constructor, e.g. `{decimal(69)}`. However, note that
+/// `{decimal(int)}` constructor, e.g. `{decimal(59)}`. However, note that
 /// constructing a decimal from a [floating-point number]($float), while
 /// supported, **is an imprecise conversion and therefore discouraged.** A
 /// warning will be raised if Typst detects that there was an accidental `float`
--- a/crates/typst/src/layout/page.rs
+++ b/crates/typst/src/layout/page.rs
@@ -189,8 +189,11 @@ pub struct PageElem {
     ///
     /// When set to `{none}`, the background becomes transparent. Note that PDF
     /// pages will still appear with a (usually white) background in viewers,
-    /// but they are conceptually transparent. (If you print them, no color is
-    /// used for the background.)
+    /// but they are actually transparent. (If you print them, no color is used
+    /// for the background.)
+    ///
+    /// If you select white explicitly, will it physically color it in white? If
+    /// not, then the last sentence is misleading and should be removed.
     ///
     /// The default of `{auto}` results in `{none}` for PDF output, and
     /// `{white}` for PNG and SVG.
@@ -245,7 +248,7 @@ pub struct PageElem {
     /// The page's header. Fills the top margin of each page.
     ///
     /// - Content: Shows the content as the header.
-    /// - `{auto}`: Shows the page number if a `numbering` is set and
+    /// - `{auto}`: Shows the page number if the `numbering` is set and
     ///   `number-align` is `top`.
     /// - `{none}`: Suppresses the header.
     ///
@@ -275,7 +278,7 @@ pub struct PageElem {
     /// The page's footer. Fills the bottom margin of each page.
     ///
     /// - Content: Shows the content as the footer.
-    /// - `{auto}`: Shows the page number if a `numbering` is set and
+    /// - `{auto}`: Shows the page number if the `numbering` is set and
     ///   `number-align` is `bottom`.
     /// - `{none}`: Suppresses the footer.
     ///
--- a/crates/typst/src/layout/repeat.rs
+++ b/crates/typst/src/layout/repeat.rs
@@ -42,7 +42,7 @@ pub struct RepeatElem {
 
     /// Whether to increase the gap between instances to completely fill the
     /// available space.
-    #[default(true)]
+    #[default(true)] // Maybe disable it for `gap to work as expected?`
     pub justify: bool,
 }
 
--- a/crates/typst/src/model/document.rs
+++ b/crates/typst/src/model/document.rs
@@ -14,7 +14,7 @@ use crate::layout::Page;
 /// All documents are automatically wrapped in a `document` element. You cannot
 /// create a document element yourself. This function is only used with
 /// [set rules]($styling/#set-rules) to specify document metadata. Such a set
-/// rule must appear before any of the document's contents.
+/// rule must not occur inside of layout containers.
 ///
 /// ```example
 /// #set document(title: [Hello])
--- a/crates/typst/src/model/numbering.rs
+++ b/crates/typst/src/model/numbering.rs
@@ -85,7 +85,7 @@ pub fn numbering(
     /// introduce unwanted behavior when there is a punctuation (or any symbol)
     /// after a reference to the element:
     ///
-    /// ```
+    /// ```example
     /// #set heading(numbering: "1.")
     /// = First <a>
     /// This is a reference: @a.
--- a/crates/typst/src/model/numbering.rs
+++ b/crates/typst/src/model/numbering.rs
@@ -80,6 +80,19 @@ pub fn numbering(
     /// particularly useful in itself, it means that you can just give arbitrary
     /// numberings to the `numbering` function without caring whether they are
     /// defined as a pattern or function.
+    ///
+    /// Be careful when adding a suffix inside the function, as it will
+    /// introduce unwanted behavior when there is a punctuation (or any symbol)
+    /// after a reference to the element:
+    ///
+    /// ```
+    /// #set heading(numbering: "1.")
+    /// = First <a>
+    /// This is a reference: @a.
+    /// #set heading(numbering: (..n) => numbering("1.", ..n))
+    /// == Second <b>
+    /// This is a reference: @b. Did you notice the second period?
+    /// ```
     numbering: Numbering,
     /// The numbers to apply the numbering to. Must be positive.
     ///
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -364,7 +364,7 @@ description: |
   - Added [go to definition](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition)
     feature in the web app. Hold Ctrl/Cmd and click on the used function or
     variable to go to its definition. `gd` in Vim mode
-    [isn't yet supported](https://github.com/typst/webapp-issues/issues/125#issuecomment-2395098031).
+    [isn't yet supported](https://github.com/typst/webapp-issues/issues/451).
   - The "compiler" field for specifying the minimum Typst version required by a
     package now supports imprecise bounds like 0.11 instead of 0.11.0
   - Added warning when a label is ignored by Typst because no preceding
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,8 +12,8 @@ description: |
   [figures]($figure.scope)
 - Added support for automatic [line numbering]($par.line) (often used in
   academic papers)
-- Typst's layout engine is now multi-threaded. Typical speedups are 2-3x for
-  larger documents. The multi-threading operates on page break boundaries, so
+- Typst's layout engine is now multithreaded. Typical speedups are 2-3x for
+  larger documents. The multithreading operates on page break boundaries, so
   explicit page breaks are necessary for it to kick in.
 - Paragraph justification was optimized with a new two-pass algorithm. Speedups
   are larger for shorter paragraphs and range from 1-6x.
@@ -55,8 +55,14 @@ description: |
     combination with [`layout`].
   - The height of a `block`, `image`, `rect`, `square`, `ellipse`, or `circle`
     can now be specified in [fractional units]($fraction)
-  - The [`scale`] function now supports non-relative lengths for `x` and `y`.
-    This way an element of unknown size can be scaled to a fixed size.
+    <!--
+        Add "for example/use case see" and link some `fraction` example,
+        so that new users would understand. Maybe an interesting example to add
+        is with floats.
+        See https://forum.typst.app/t/how-can-i-get-a-box-to-use-all-height-under-a-floating-element/639/8
+    -->
+  - The [`scale`] function now supports non-relative lengths for `x`, `y`,
+    `factor`. This way an element of unknown size can be scaled to a fixed size.
   - The values of `block.above` and `block.below` can now be retrieved in
     context expressions.
   - Fixed a bug which could cause headings to be orphaned at the bottom of the
@@ -66,7 +72,8 @@ description: |
   - Fixed empty pages appearing when a [context] expression wraps whole pages
   - Fixed `{set block(spacing: x)}` behaving differently from
     `{set block(above: x, below: x)}`
-  - Fixed behaviour of [`rotate`] and [`scale`] with `{reflow: true}`
+  - Fixed behavior of [`rotate`] and [`scale`] with `{reflow: true}`
+    <!-- Why this is still not `true` by default? -->
   - Fixed interaction of `{align(horizon)}` and `{v(1fr)}`
   - Fixed various bugs where floating placement would yield overlapping results
   - Fixed a bug where widow/orphan prevention would unnecessarily move text into
@@ -89,11 +96,13 @@ description: |
   - Added [`text.costs`] parameter for tweaking various parameters that affect
     the choices of the layout engine during text layout
   - Added `typm` highlighting mode for math in [raw blocks]($raw.lang)
+    <!-- Update guidelines in the forum to include `typm` too. -->
   - Added basic i18n for Galician, Catalan, Latin, Icelandic, Hebrew
   - Implemented hyphenation duplication for Czech, Croatian, Lower Sorbian,
     Polish, Portuguese, Slovak, and Spanish.
   - The [`smallcaps`] function is now an element function and can thereby be
     used in show(-set) rules.
+    <!-- Can we have `capitalize()`? -->
   - The [`raw.theme`] parameter can now be set to `{none}` to disable
     highlighting even in the presence of a language tag, and to `{auto}` to
     reset to the default
@@ -111,9 +120,10 @@ description: |
   - Fixed [smart quotes]($smartquote) for Swiss French
   - New font metadata exceptions for Archivo, Kaiti SC, and Kaiti TC
   - Updated bundled New Computer Modern fonts to version 6.0
+    <!-- How to correctly enable bold for math if `strong()` is used? -->
 
 - Math
-  - Block-level equations can now break over multiple pages. This behaviour can
+  - Block-level equations can now break over multiple pages. This behavior can
     be disabled via `{show math.equation: set block(breakable: false)}`.
   - Matrix and vector sizing is now more consistent across different cell
     contents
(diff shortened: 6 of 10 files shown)


xkevio:
The `behaviour` -> `behavior` and `e.g.` -> `e.g.,` change is unnecessary as it's just a switch between British English and American English.

laurmaedje:
Yes, please base it on main instead. The rc branch will not be updated anymore.

Andrew15-5:
> The `behaviour` -> `behavior` and `e.g.` -> `e.g.,` change is unnecessary as it's just a switch between British English and American English.

If you search for all occurrences of all 4 things, e.g.,

```sh
rg -F behaviour docs/changelog.md
rg -F behavior docs/changelog.md
rg -F 'e.g. ' docs/changelog.md
rg -F 'e.g.,' docs/changelog.md
```

You would find that all 4 return some matches. This indicates that the file uses mixed English grammar. Since Typst uses American English in its syntax (e.g., [`color`](https://typst.app/docs/reference/visualize/color/), [`transparentize`](https://typst.app/docs/reference/visualize/color/#definitions-transparentize)), this means that all user-faced stuff (including docs and changelog) should also use American English.

No matter which side you are on, the mix of grammar should be removed. So one of the two "Englishes" must go. And since American is the preferred way, it should also be used in the changelog (based on common sense).

And since this PR specifically targets all the small details, fixes etc., I think this is a necessary part of the change. I can split it into a separate PR that can be merged after the release, but it doesn't really make sense to separate it, since the changelog for the 0.12.0 is also affected by these changes. Easier to fix everything now and be free later.

crates/typst/src/foundations/decimal.rs
```diff
@@ -29,7 +29,7 @@ use crate::World;
 /// raised).
 ///
 /// You can also convert any [integer]($int) to a decimal with the
-/// `{decimal(int)}` constructor, e.g. `{decimal(59)}`. However, note that
+/// `{decimal(int)}` constructor, e.g. `{decimal(69)}`. However, note that
```
CGMossa, line 32: Why?

Andrew15-5: We all know why. :)

<details><summary>In case you really don't know</summary>

There is a **very** popular meme about the number 69. Every time you hear it, you would say "nice". And since it already had the 9 in ones and off-by-one in tens, I think it's a great opportunity to insert a drop of modern humor in the docs (because it also doesn't harm anyone or anything like that, it's a purely positive thing). I think a lot of people, who know about this thing/meme/trend, would really appreciate Typst having it in the docs. You can say it can be one of hidden gem. ;)

</details>

Enivex: Using 69 in an example is just unprofessional

Andrew15-5: People that don't know what this all about would disagree with you (at least IMO). It's just a regular integer like any other (like 59, but bigger). Nothing special. But if someone _knows_ and sees it in the docs, I'm sure it will put a little smile on their faces. It certainly does put a smile on _mine_ every time _I_ see it anywhere.

laurmaedje: I'd prefer for this to be left as-is.

Andrew15-5: > I'd prefer for this to be left as-is.

I'm not 100% sure what this refers to. To my change or to the initial docs? Do I revert the change?

Andrew15-5:
FYI, for the last one I used:

```sh
rg -Ug *.rs '/// ```\n\s+/// '
```

to find all the broken examples. Correct examples can be found with:

```sh
rg -Ug *.rs '/// ```example\n\s+/// '
```

Not that for the first one there will be matches from the `raw.rs`, but I think this can be fixed/those can be ignored with something like this:

```sh
rg -Ug *.rs '/// ```\n\s+/// [^`]+'
```

crates/typst/src/foundations/decimal.rs
```diff
@@ -80,11 +80,11 @@ use crate::World;
 /// multiplication, and [power]($calc.pow) to an integer, will be highly precise
 /// due to their fixed-point representation. Note, however, that multiplication
 /// and division may not preserve all digits in some edge cases: while they are
-/// considered precise, digits past the limits specified below are rounded off
+/// considered precise, digits past the limits specified below (where?) are rounded off
```
PgBiel, line 83:
```suggestion
/// considered precise, digits past the limits specified above are rounded off
```

crates/typst/src/foundations/decimal.rs
```diff
@@ -80,11 +80,11 @@ use crate::World;
...
 /// and lost, so some loss of precision beyond the maximum representable digits
 /// is possible. Note that this behavior can be observed not only when dividing,
 /// but also when multiplying by numbers between 0 and 1, as both operations can
-/// push a number's fractional digits beyond the limits described below, leading
+/// push a number's fractional digits beyond the limits described below (where?), leading
```
PgBiel, line 87:
```suggestion
/// push a number's fractional digits beyond the limits described above, leading
```

crates/typst/src/layout/page.rs
```diff
@@ -245,7 +248,7 @@ pub struct PageElem {
     /// The page's header. Fills the top margin of each page.
     ///
     /// - Content: Shows the content as the header.
-    /// - `{auto}`: Shows the page number if a `numbering` is set and
+    /// - `{auto}`: Shows the page number if the `numbering` is set and
```
PgBiel, line 251: Either keep "a `numbering`" or explicitly write out "the `numbering` field". Not sure if the field can be linked to, if so it'd be preferable to keep the original with a link to the field.

crates/typst/src/layout/repeat.rs
```diff
@@ -42,7 +42,7 @@ pub struct RepeatElem {
 
     /// Whether to increase the gap between instances to completely fill the
     /// available space.
-    #[default(true)]
+    #[default(true)] // Maybe disable it for `gap to work as expected?`
```
PgBiel, line 45: Not sure how this counts as a doc change (I'd rather have this in a separate PR). You could open (or link to) an issue demonstrating how the current behavior is broken (if so).

crates/typst/src/model/document.rs
```diff
@@ -14,7 +14,7 @@ use crate::layout::Page;
 /// All documents are automatically wrapped in a `document` element. You cannot
 /// create a document element yourself. This function is only used with
 /// [set rules]($styling/#set-rules) to specify document metadata. Such a set
-/// rule must appear before any of the document's contents.
+/// rule must not occur inside of layout containers.
```
PgBiel, line 17:
```suggestion
/// rule must not occur inside layout containers.
```

crates/typst/src/model/numbering.rs
```diff
@@ -80,6 +80,19 @@ pub fn numbering(
...
+    /// after a reference to the element:
+    ///
+    /// ```example
+    /// #set heading(numbering: "1.")
+    /// = First <a>
+    /// This is a reference: @a.
+    /// #set heading(numbering: (..n) => numbering("1.", ..n))
+    /// == Second <b>
+    /// This is a reference: @b. Did you notice the second period?
+    /// ```
```
PgBiel, line 95: I feel like we should first note that references trim prefixes and suffixes from typical numbering formats, and only then explain that this does not happen for functions.

Though this section isn't related to any change in 0.12, so this can be done in a separate PR if this discussion blocks anything.

crates/typst/src/layout/page.rs
```diff
@@ -275,7 +278,7 @@ pub struct PageElem {
     /// The page's footer. Fills the bottom margin of each page.
     ///
     /// - Content: Shows the content as the footer.
-    /// - `{auto}`: Shows the page number if a `numbering` is set and
+    /// - `{auto}`: Shows the page number if the `numbering` is set and
```
PgBiel, line 281: see previous comment on this

crates/typst/src/layout/page.rs
Andrew15-5: Now let's [go back](https://discord.com/channels/1054443721975922748/1088371867913572452/1292076710786760744) and settle the question I had for a long time. Is it field or parameter? I feel like in messages it's usually "field", but in the docs, IDs are named with "parameter". I think that we should streamline docs to 1 term to be more consistent and easier to identify what is being talked about. If they are interchangeable, then perhaps we can use both, but IDK.

But I also don't understand why you think that the original "a numbering" is equally good? The sentence mentions _the_ `page.numbering`, a specific thing, not some random `numbering` variable or something.

I think the link should be possible. I propose keeping the new change + adding the link to the `page.nubering`.

crates/typst/src/layout/repeat.rs
Andrew15-5: See the OP:

> I created a PR with all the polished and fixed stuff (plus nitpicks and **questions/proposals**)

Are you sure PR is OK instead of an issue? I wanted to have a (small) discussion about this, and also I didn't yet think of a code change (changing `true` to `false` might not be enough), so for a PR I need to come up with a proposed change.

crates/typst/src/layout/page.rs
PgBiel: - It's a field. "Parameter" is just a broad name for whatever you can use in a function or constructor, but for elements the constructor parameters are usually the same as the element's fields. Here there is no doubt that we're referring to the value of the field and not to the constructor.
- The original is good because it's referring to the word "numbering" itself. You can either have _a numbering_ be set on your page (i.e. have exactly one option of numbering format to be used in your page), or have _the numbering field_ (here the article refers to _field_ instead of _numbering_). Both work, which is why I don't think the change is necessary. But adding a link could be good.

crates/typst/src/layout/repeat.rs
PgBiel: Open an issue, please.

laurmaedje:
I combined all docs PRs into https://github.com/typst/typst/pull/5147 to prevent conflicts and too much time spent on discussions and review back and forth. Thanks!

crates/typst/src/model/numbering.rs
laurmaedje: I dropped this for now in https://github.com/typst/typst/pull/5147. I don't really like documenting bugs, even though I guess it's sometimes pragmatic.

crates/typst/src/layout/page.rs
```diff
@@ -189,8 +189,11 @@ pub struct PageElem {
     ///
     /// When set to `{none}`, the background becomes transparent. Note that PDF
     /// pages will still appear with a (usually white) background in viewers,
-    /// but they are conceptually transparent. (If you print them, no color is
-    /// used for the background.)
+    /// but they are actually transparent. (If you print them, no color is used
+    /// for the background.)
+    ///
+    /// If you select white explicitly, will it physically color it in white? If
```
laurmaedje, line 195: I don't get what you find misleading here.

docs/changelog.md
```diff
@@ -55,8 +55,14 @@ description: |
     combination with [`layout`].
   - The height of a `block`, `image`, `rect`, `square`, `ellipse`, or `circle`
     can now be specified in [fractional units]($fraction)
-  - The [`scale`] function now supports non-relative lengths for `x` and `y`.
-    This way an element of unknown size can be scaled to a fixed size.
+    <!--
+        Add "for example/use case see" and link some `fraction` example,
```
laurmaedje, line 59: fr is already linked. the changelog isn't the place for examples. we can add more explanation to the fraction docs, but that's out of scope here

docs/changelog.md
```diff
@@ -66,7 +72,8 @@ description: |
   - Fixed empty pages appearing when a [context] expression wraps whole pages
   - Fixed `{set block(spacing: x)}` behaving differently from
     `{set block(above: x, below: x)}`
-  - Fixed behaviour of [`rotate`] and [`scale`] with `{reflow: true}`
+  - Fixed behavior of [`rotate`] and [`scale`] with `{reflow: true}`
+    <!-- Why this is still not `true` by default? -->
```
laurmaedje, line 76: cause I forgot, but now I won't change it after the RC

docs/changelog.md
```diff
@@ -124,23 +134,33 @@ description: |
...
+  - Added `~` shorthand for `tilde.op`. See
+    [#4706](https://github.com/typst/typst/pull/4706) for details. (**Minor
+    breaking change**)
+    <!-- Maybe add somewhere how you can get the previous output? -->
+    <!-- $a ~ b$ -->
+    <!-- $a "~" b$ -->
+    <!-- $a class("normal", ~) b$ -->
+    <!-- $a class("normal", "~") b$ -->
+    <!-- $a class("binary", "~") b$ -->
+  - Fixed baseline alignment of equation numbers. See
```
laurmaedje, line 159: I dropped these links because I don't generally link the issues. It adds too much noise in my opinion.

docs/changelog.md
```diff
@@ -124,23 +134,33 @@ description: |
...
   - Added [`vec.align`]($math.vec.align) and [`mat.align`]($math.mat.align)
     parameters
   - Added [`underparen`]($math.underparen), [`overparen`]($math.overparen),
     [`undershell`]($math.undershell), and [`overshell`]($math.underparen)
-  - Added `~` shorthand for `tilde.op` (**Minor breaking change**)
-  - Fixed baseline alignment of equation numbers
+  - Added `~` shorthand for `tilde.op`. See
+    [#4706](https://github.com/typst/typst/pull/4706) for details. (**Minor
+    breaking change**)
+    <!-- Maybe add somewhere how you can get the previous output? -->
```
laurmaedje, line 153: maybe, but also out of scope here imo

docs/changelog.md
```diff
@@ -151,18 +171,20 @@ description: |
...
-    pages. This may lead to subtly different behaviour in introspections.
+    pages. This may lead to subtly different behavior in introspections.
     (**Breaking change**)
-  - Fixed various bugs with wrong counter behaviour in complex layout
+  - Fixed various bugs with wrong counter behavior in complex layout
     situations, through a new, more principled implementation
   - Counter updates can now be before the first, in between, and after the last
     page when isolated by weak page breaks. This allows, for instance, updating
     a counter before the first page header and background.
+    <!-- Counter updates can now be in between pages?
```
laurmaedje, line 181: No. Normal counter updates are just somewhere on the page. what this means is that this
```typ
#set page(numbering: "1", number-align: top)
#counter(page).update(5)
```
is different from
```typ
#counter(page).update(5)
#set page(numbering: "1", number-align: top)
```
because the counter is before the first page (the set page creates a weak pagebreak).

This is a new capability that didn't exist before.

docs/changelog.md
```diff
@@ -222,6 +244,10 @@ description: |
 
 - Syntax
   - Added support for nested imports like `{import "file.typ": module.item}`
+    <!-- Add docs for this in
```
laurmaedje, line 247: I added a bit of docs to scripting

docs/changelog.md
```diff
@@ -235,30 +261,41 @@ description: |
...
   - Added `exact` argument to [`array.zip`]
   - Added [`arguments.at`] method
   - Added [`int.from-bytes`], [`int.to-bytes`], [`float.from-bytes`], and
     [`float.to-bytes`]
-  - [`calc.round`] no longer accepts negative digits (**Minor breaking change**)
+  - `digits` parameter of [`calc.round`] no longer accepts negative integers
+    (**Minor breaking change**)
   - Conversions from [`int`] to [`float`] will now error instead of saturating
     if the float is too large (**Minor breaking change**)
+    <!-- Pretty sure docs should be updated here:
```
laurmaedje, line 274: it is mentioned in the int constructor docs

docs/changelog.md
```diff
@@ -235,30 +261,41 @@ description: |
...
+         https://staging.typst.app/docs/reference/foundations/float/#constructor
+    -->
   - Added `float.nan` and `float.inf`, removed `calc.nan`
     (**Minor breaking change**)
+    <!-- Docs missing a simple example of printing and/or making these values. -->
```
laurmaedje, line 279: don't know what you mean here

docs/changelog.md
```diff
@@ -235,30 +261,41 @@ description: |
...
+         https://staging.typst.app/docs/reference/foundations/calc#functions-ceil
+    -->
   - Improved [`repr`] of relative values, sequences, infinities, NaN,
     `{type(none)}` and `{type(auto)}`
   - Fixed crash on whole packages (rather than just files) cyclically importing
     each other
-  - Fixed behaviour of [`calc.round`] on integers when a non-zero value is
+  - Fixed behavior of [`calc.round`] on integers when a non-zero value is
     provided for `digits`
+    <!-- I wasn't able to reproduce this. -->
```
laurmaedje, line 292: `#type(calc.round(10, digits: 1))` is now `int` instead of `float`

docs/changelog.md
```diff
@@ -235,30 +261,41 @@ description: |
...
 
 - Styling
   - Text show rules now match across multiple text elements
   - The string `{"}` in a text show rule now matches smart quotes
   - Fixed a long-standing styling bug where the header and footer would
-    incorrectly inherit styles from a lone element on the page (e.g. a heading)
+    incorrectly inherit styles from a lone element on the page (e.g., a heading)
```
laurmaedje, line 298: I would prefer to not make this change now. Maybe in the future consistently across the codebase.

docs/changelog.md
```diff
@@ -283,25 +320,32 @@ description: |
...
-    larger documents. The multi-threading operates on page break boundaries, so
+  - Typst's layout engine is now multithreaded. Typical speedups are 2-3x for
+    larger documents. The multithreading operates on page break boundaries, so
     explicit page breaks are necessary for it to kick in.
   - Paragraph justification was optimized with a new two-pass algorithm.
     Speedups are larger for shorter paragraphs and range from 1-6x.
 
 - Command Line Interface
   - Added `--pages` option to select specific page ranges to export
+    <!-- Add accepted format for the argument.
```
laurmaedje, line 331: CLI docs are a separate topic and out of scope here

docs/changelog.md
```diff
@@ -283,25 +320,32 @@ description: |
...
+         It's a good idea to add a CLI section in the docs where everything can
+         be explained in great details. -->
   - Added `--package-path` and `--package-cache-path` as well as
     `TYPST_PACKAGE_PATH` and `TYPST_PACKAGE_CACHE_PATH` environment variables
     for configuring where packages are loaded from and cached in, respectively
   - Added `--ignore-system-fonts` flag to disable system fonts fully for better
     reproducibility
   - Added `--make-deps` argument for outputting the dependencies of the current
     compilation as a Makefile
+    <!-- Doesn't work/escape for `--make-deps Makefile 2024-10-05T14:20:16Z.typ`. -->
```
laurmaedje, line 343: works for me. can you elaborate / open an issue with a concrete problem?

docs/changelog.md
```diff
@@ -317,6 +361,10 @@ description: |
   - Fixed path completions for `typst` not working in zsh
 
 - Tooling & Diagnostics
+  - Added [go to definition](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition)
```
laurmaedje, line 364: This already existed before 0.12 in the web app. Since we also haven't announced other web app features here that landed since 0.11, I've removed this.

laurmaedje:
Since the PR also contained some general remarks, I left comments on those even though it's already closed!

docs/changelog.md
Andrew15-5: But you describe the "before the first" part, which I already more or less understood. Do you mean that the changelog is wrong and "in between" must be removed? I don't understand.

laurmaedje: In between means
```typ
#pagebreak(weak: true)
#counter(page).update(5)
#pagebreak(weak: true)
```
which is logically between the first page's footer and second page's header

docs/changelog.md
Andrew15-5: You mean this?

> [Raises an error if there is an attempt to produce an integer larger than the maximum 64-bit signed integer or smaller than the minimum 64-bit signed integer.](https://staging.typst.app/docs/reference/foundations/int/#constructor)

This applies to all data types, but the changelog specifically mentions `float`, so I thought that the `float` part is missing. Maybe the changelog should include all data types or something (be exclusive).

docs/changelog.md
Andrew15-5: There is https://staging.typst.app/docs/reference/foundations/float/#definitions-is-nan, but there is not a single instance of something like this:

```typ
#let nan = <create NaN>
#nan
#let infinity = <create infinity>
#infinity
```

There should be an example of how the values can generally be constructed and how would they print out. If direct value isn't pretty, maybe `repr()` can be used for better output. I haven't tried doing any of that.

docs/changelog.md
Andrew15-5: Shouldn't then this be called "result type" instead of "behavior"?

docs/changelog.md
Andrew15-5: Are you referring to the whole British → American part? I can do that. Does the future just mean post-0.12?

docs/changelog.md
Andrew15-5: Wait, what? Why? It is still a part of the release and is _a docs_. Do you just mean I should create a separate issue (maybe PR) for this (before the release)?

Andrew15-5:
> too much time spent on discussions and review back and forth

It is unfortunate that I don't have much of free time, otherwise I think this would've gone much faster.

crates/typst/src/layout/page.rs
laurmaedje: If you specify white, and the printer can print white, it will indeed use that white.

Andrew15-5: Mind blown

But do you know any printer that can print white? Because if it's just a theoretical thing and not practical, then this argument doesn't hold.

laurmaedje: There are printers that can do this.

docs/changelog.md
laurmaedje: Out of scope for a fixup/touch-up PR. You can open an issue if you want (the issue counter though 🥲)

docs/changelog.md
laurmaedje: I'm confused. It literally says above that `float.nan` was added. `#let nan = float.nan`.

docs/changelog.md
laurmaedje: I meant specifically `e.g.,` because the change surface is so large and it's also used a lot across the codebase. Future means simply that I don't want to deal with it right now.

docs/changelog.md
laurmaedje: I meant that we have no docs for the CLI as a whole. Writing them now is too much work. Even if someone else did, I would need to do a lot of review/editing. There is already an issue: https://github.com/typst/typst/issues/3765

laurmaedje:
> too much time spent on discussions and review back and forth

Your response time is quite fast. I mostly did it this way because it was the fastest way for me.

Andrew15-5:
> Your response time is quite fast.

Ohh, you flatter me. But if we compare multi-week/month responses, then for sure. When I have (enough/plenty of) time, ideally I would respond within few minutes, not few hours/days.

> I mostly did it this way because it was the fastest way for me.

Yeah, I guess so. I didn't notice that there are a bunch of other PRs with similar things too.

crates/typst/src/layout/page.rs
Andrew15-5: TIL.

docs/changelog.md
Andrew15-5: > Where exactly?

Wait, I've read the wrong thing. I was looking below as if it was doc string. I see the link now.

> the issue counter though

Oh, haha, I only understood this after a couple of tries. Yeah, since around 2024-01-01 it increased by more than 100 (soon will be 600 🎉). But I noticed that there are a lot of easy ones that were thankfully labeled as such. I'm looking forward to eliminating those in the near future (hopefully in this or next month).

docs/changelog.md
Andrew15-5: Sorry, I was looking in the wrong tab, I see it now.

image

docs/changelog.md
Andrew15-5: Got it. I agree that the comma change is crazy widespread. It pollutes the diff quite nicely. I will try making a list of stuff we've discussed here and then open issues/PR for them. But I don't know when I will have time for this.

docs/changelog.md
Andrew15-5: No dedicated docs, but there is, for example, a whole section about it: https://typst.app/docs/reference/introspection/query/#command-line-queries. But I agree that this won't fit for `--pages`. I was actually thinking that adding one sentence showing the accepted format in the changelog would be enough for now.

docs/changelog.md
Andrew15-5: Can you please confirm (if you already read the comment) if I linked to what you are referring to? I want to undo all changes that were applied in one way or another. This way I will be left only with stuff that needs to be addressed in the future (when I get more time).

docs/changelog.md
```diff
@@ -283,25 +320,32 @@ description: |
...
     `TYPST_PACKAGE_PATH` and `TYPST_PACKAGE_CACHE_PATH` environment variables
     for configuring where packages are loaded from and cached in, respectively
   - Added `--ignore-system-fonts` flag to disable system fonts fully for better
     reproducibility
   - Added `--make-deps` argument for outputting the dependencies of the current
     compilation as a Makefile
+    <!-- Doesn't work/escape for `--make-deps Makefile 2024-10-05T14:20:16Z.typ`. -->
   - Added `--pretty` option to `typst query`, with the default now being to
-    minify
+    minify (only takes effect for JSON format)
```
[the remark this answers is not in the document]
laurmaedje: I don't think it's worth mentioning because it is not explicitly only for JSON. It's just that we only have JSON and YAML and minfying YAML doesn't really make much sense. If we had another format that can be minified, it would be minified.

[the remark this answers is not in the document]
Andrew15-5: > If we had another format that can be minified, it would be minified.

True, that's why I'm talking about current state of affairs.

---

> it is not explicitly only for JSON. 

It is not _explicitly_, but it _is implicitly_ only for JSON (currently).

---

> It's just that we only have JSON and YAML and minfying YAML doesn't really make much sense.

That's what I'm talking about. The "only takes effect for JSON format" describes exactly that. Perhaps someone would think that YAML also is being minified, but no, it is in fact not being minified/compressed, but in some situations it _can_ be. So I still think that it's worth mentioning to clear any potential misunderstandings/misinterpretations.

PR #5138, opened 2024-10-05
Syntax set macro

wrzian:
Add a new syntax_set! macro to clean up syntax set code in the parser.

The commits here are ordered intentionally:
- Commit (1) adds the macro and uses it in `set.rs` and commit (2) uses it within `parser.rs`.
    - I'll note that I used two macros: `make_set` is local only and used within the file, and `syntax_set` which is `pub(crate)` for export into `parser.rs`. I tried to unify them, but it just seems easier to have two macros in my opinion. I could be convinced otherwise, and would also be ok if we wanted to remove `make_set` and only have `syntax_set`.
-  Commit (3) simplifies the `markup` and `markup_expr` functions by recognizing that the `MARKUP_EXPR` set is completely redundant with the match statement in `markup_expr()`. I think this is just a straightforward performance win. We could also likely do this for math expressions, but that is a bit less obviously redundant. And I don't think this would be possible with code expressions, since they use `p.unexpected()` in a much more fine grained manner.
- Commit (4) refactors the heading function to prepare for (5). I'm not sure if this is actually desired. The changes is that headings now include trailing spaces on the following element if that element is a label. This passess all the  tests, so I don't see why it's an issue. But I can drop it if it would be (and we should add a test if it does matter).
- Commit (5) is the big payoff. I refactor all of the main functions that take a closure to now take a syntax set. This makes things so much cleaner, but does depend on the change in (4).

### DIFF
--- a/crates/typst-syntax/src/set.rs
+++ b/crates/typst-syntax/src/set.rs
@@ -39,22 +39,13 @@ const fn bit(kind: SyntaxKind) -> u128 {
     1 << (kind as usize)
 }
 
-/// Generate an inline const syntax set of the given kinds.
-///
-/// Example: `if syntax_set!(Plus, Minus).contains(SyntaxKind::Plus) { }`
-/// `if {const SET:SyntaxSet = SyntaxSet::new()
-///     .add(SyntaxKind::Plus)
-///     .add(SyntaxKind::Minus);
-///     SET
-/// }.contains(SyntaxKind::Plus) { }`
+/// Generate a compile-time constant `SyntaxSet` of the given kinds.
 macro_rules! syntax_set {
-    ( $($kind:ident),* $(,)? ) => {
-        { // 'SET' never escapes this inner block
-            const SET: crate::set::SyntaxSet = crate::set::SyntaxSet::new()
-                $(.add(crate::SyntaxKind:: $kind))*;
-            SET
-        }
-    }
+    ($($kind:ident),* $(,)?) => {{
+        const SET: crate::set::SyntaxSet = crate::set::SyntaxSet::new()
+            $(.add(crate::SyntaxKind:: $kind))*;
+        SET
+    }}
 }
 
 // Export so other modules can import as: `use set::syntax_set`
--- a/crates/typst-syntax/src/set.rs
+++ b/crates/typst-syntax/src/set.rs
@@ -75,34 +75,6 @@ macro_rules! make_set {
 /// Syntax kinds that can start a statement.
 pub const STMT: SyntaxSet = make_set!(Let, Set, Show, Import, Include, Return);
 
-/// Syntax kinds that can start a markup expression.
-pub const MARKUP_EXPR: SyntaxSet = make_set!(
-    Space,
-    Parbreak,
-    LineComment,
-    BlockComment,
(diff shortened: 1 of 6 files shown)


crates/typst-syntax/src/set.rs
```diff
@@ -39,54 +39,61 @@ const fn bit(kind: SyntaxKind) -> u128 {
...
 
+/// Generate an inline const syntax set of the given kinds.
+///
+/// Example: `if syntax_set!(Plus, Minus).contains(SyntaxKind::Plus) { ... }`
+/// Becomes:
+/// ```rust
+/// if {const SET:SyntaxSet = SyntaxSet::new()
+///     .add(SyntaxKind::Plus)
+///     .add(SyntaxKind::Minus)
+/// }.contains(SyntaxKind::Plus) { ... }
```
MDLC01: I believe `SET` should be returned here within the block expression.

wrzian: Yep :upside_down_face:

wrzian:
Sounds good. I've backed-out the commits 4 and 5 to leave the PR history intact (which should not matter once squashed and merged).

There will be a similar change to the heading function in the parser trivia refactor PR, but that change happens after trivia is reworked, and is able to keep trailing trivia outside the heading and imrpove readability :)

wrzian:
Added a quick test for heading trailing whitespace behavior. tbh I'm not convinced that it is desirable, but I'll go with what you decide.

crates/typst-syntax/src/set.rs
```diff
@@ -39,54 +39,55 @@ const fn bit(kind: SyntaxKind) -> u128 {
...
+
+// Export so other modules can import as: `use set::syntax_set`
+pub(crate) use syntax_set;
+
+/// Just build a syntax set of the given kinds. Only used in this file.
+///
+/// Example: `const SET: SyntaxSet = make_set!(Plus, Minus);`
+/// Becomes:
+/// `SyntaxSet::new().add(SyntaxKind::Plus).add(SyntaxKind::Minus)`
+macro_rules! make_set {
```
laurmaedje, line 68: Why are there two macros? They seem to be doing the same thing, one with the extra const. We can use the const one to ensure that it's really constant, but why do we need this one?

wrzian: Just that it felt redundant to — after expansion — create the in-module macros as `const SOME_SET: SyntaxSet = { const SET: SyntaxSet = ...; SET }`. Also only `syntax_set!` is exported, so this one is local to only this file.
I'm fine with just using one.

laurmaedje: I think the redundancy after expansion is unproblematic. If you take a look at the output of `derive(PartialEq)`: It's much worse ...

laurmaedje: The cleanest solution would be an inline const block: `const { .. }`, but it's not yet compatible with our Minimum Supported Rust Version.

wrzian: Alright, changing to just the syntax_set macro.

laurmaedje:
Thanks!

PR #5147, opened 2024-10-07
Docs and changelog improvements

laurmaedje:
Unifies the improvements from the various other PRs. Discussing all them individually with review back and forth would have taken too long. Also, there were some conflicts between the PRs. So, this was simply the pragmatic way to land the changes. Note that I dropped some of the changes where I didn't consider them an improvement.


docs/changelog.md
```diff
@@ -66,7 +66,7 @@ description: |
   - Fixed empty pages appearing when a [context] expression wraps whole pages
   - Fixed `{set block(spacing: x)}` behaving differently from
     `{set block(above: x, below: x)}`
-  - Fixed behaviour of [`rotate`] and [`scale`] with `{reflow: true}`
+  - Fixed behavior of [`rotate`] and [`scale`] with `{reflow: true}`
```
drupol, line 69: Just my 2 cents: I would have kept the British english way of writing by default.

laurmaedje: The codebase uses primarily American English, e.g. `color`. I personally use a mix... I also prefer "behaviour", but a bit of consistency is okay.

drupol: Fair enough ;)

Andrew15-5:
You've missed `Anselm Schüler <mail@anselmschueler.com>` from https://github.com/typst/typst/pull/5136/commits/cd196b24e1ba5afae92377ef64f160e9efe63fbb.

MDLC01:
I reckon the documentation for `int.to-bytes`'s `size` parameter should be improved as well.

[Quoting myself from Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1291700412478521436):
> I'm not sure I understand this in [the documentation for `int.to-bytes`'s `size` parameter](https://staging.typst.app/docs/reference/foundations/int/#definitions-to-bytes-size). The first paragraph seems to mean that the least significant bits are discarded if the integer is too large (which is weird imo, we should either discard the most significant bits, or raise an error), but this is contradictory with the second paragraph.
> image
> I think the first paragraph should be rephrased, but I want to make sure I understand it before

Andrew15-5 (commented):
Looks pretty good on the first read. I'm glad that most of my fixes were included.

docs/changelog.md
```diff
@@ -136,8 +136,8 @@ description: |
   - Added [`vec.align`]($math.vec.align) and [`mat.align`]($math.mat.align)
     parameters
   - Added [`underparen`]($math.underparen), [`overparen`]($math.overparen),
-    [`undershell`]($math.undershell), and [`overshell`]($math.underparen)
-  - Added `~` shorthand for `tilde.op` (**Minor breaking change**)
+    [`undershell`]($math.undershell), and [`overshell`]($math.overshell)
+  - Added `~` shorthand for `tilde.op` in math mode (**Minor breaking change**)
```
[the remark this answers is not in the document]
laurmaedje: But it wasn't a shorthand before, so that would be confusing. You can also use "a" in math mode, but it's not a shorthand.

README.md
```diff
@@ -129,7 +129,7 @@ Typst's CLI is available from different sources:
     `nix run github:typst/typst -- --version`.
 
 - Docker users can run a prebuilt image with
-  `docker run -it ghcr.io/typst/typst:latest`.
+  `docker run ghcr.io/typst/typst:latest --help`.
```
Andrew15-5, line 132: Do we really need the `:latest` part? I think it should be dropped.

PgBiel: It's pretty standard in the Docker scene. I see no reason to drop it.

Andrew15-5: The reason is the shorter command, as it makes the command unnecessary verbose.

I like it clean and short. Ahem...

PR #5148, opened 2024-10-07
Make `Vm` attributes more public

frozolotl:
This change is necessary for typst-repl to keep set- and show-rules, as well as variables across evaluations.

Perhaps it would be wiser to improve the API as a whole to make that use-case easier to implement without exposing details like these, but this is for sure the easiest solution.


laurmaedje:
Since so much is public already, I guess this is just another drop in the bucket. I'd be okay with merging, but noting again that it is not unlikely that this interface will break frequently.

PR #5149, opened 2024-10-07
Fix smartquotes after inline equations

### LINKED ISSUE #5146: Smart quote behavior after inline math changed in v0.12.0-rc1
### Description

```typst
$i$'s

$i$ 's
```

Back in v0.11.1, this produces
image

But in v0.12.0-rc1, this produces
image

It this change intentional or a bug?

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

Coekjan:
Fixes #5146 .

## Investigation

An inline equation is converted to unicode U+FFFC (object replacement char) in `collect.rs`:

https://github.com/typst/typst/blob/142c9dff49adf03c30268184621d361839991d36/crates/typst/src/layout/inline/collect.rs#L69

Before #4849 , this char leads smartquotes-algorithm to use the fallback (`’`):

https://github.com/typst/typst/blob/4e4c5175e531d99ec6f39d9a77d59d02f2109308/crates/typst/src/text/smartquote.rs#L159-L161

And #4869 uses a new algorithm, which does not aware of this char and fallback to use `quotes.open` (`‘`).

## A Simple Solution

I believe, in smartquotes algorithm, an inline equation is similiar with an alphabetic char. So I did this (as a POC):

```patch
         // If we have a single smart quote, didn't recently open a single
-        // quotation, and are after an alphabetic char, interpret this as an
-        // apostrophe.
-        if !double && opened != Some(false) && before.is_alphabetic() {
+        // quotation, and are after an alphabetic char or an object (e.g. a
+        // math equation), interpret this as an apostrophe.
+        if !double
+            && opened != Some(false)
+            && (before.is_alphabetic() || before == '\u{FFFC}')
+        {
             return "’";
         }
```

**NOTE**:
- As the constant `OBJ_REPLACE` is private, we cannot use it here. I thus write a `'\u{FFFC}'`.
- I am not sure if there are more kinds of objects than equations included here, but this simple solution might at least do the same thing as before #4849


laurmaedje:
Makes sense. Thanks!

PR #5139, opened 2024-10-06
Improve hint when provided array for destructuring has fewer elements than expected

Ang149:
### Summary
This PR improves the error hint for destructuring assignment when the provided array has fewer elements than expected.

### Changes Made
- Updated the error message to provide clearer guidance on how to correct the destructuring assignment.
-  Modified existing test cases to cover new error hints and ensure they accurately reflect the changes made.

### Related Issue
Fixes #5132

PR #5154, opened 2024-10-07
Fix hint for destructuring to an array with at least 1 element

bluebear94:
Addendum to #5139, fixing “at least 1 elements” to “at least 1 element”.


laurmaedje:
Thanks!

Issue #4572, opened 2024-07-17
Failed subsetting for bitmap fonts


qarmin:
### Description

File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
\00
```

command
```
timeout -v 100 typst compile TEST___FILE.typ
```

cause this
```
thread 'main' panicked at crates/typst-pdf/src/font.rs:203:75:
called `Result::unwrap()` on an `Err` value: UnknownKind
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

##### Automatic Fuzzer note, output status "Some(101)", output signal "None"
```
[compressed.zip](https://github.com/user-attachments/files/16260681/compressed.zip)

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

bluebear94:
This is planned to be turned into an export diagnostic:

https://github.com/typst/typst/blob/0ea4b1b217b2b5b902dca8e6e6ce11e1890afd40/crates/typst-pdf/src/font.rs#L200-L203

LaurenzV:
That's weird though, in theory this should only fail if the font is invalid. So either some invalid font is used when rendering this, or there is a bug in the subsetter.

LaurenzV:
And I can't reproduce it on my local machine. Do you think you could find out what font Typst is trying to use here? In case you can compile Typst yourself locally, you could do it by printing the font name in the `subset_font` method, for example.

qarmin:
Ubuntu 24.04, probably default font
```
typst compile appendix_1182521647_IDX_13_RAND_16144817441186598688330.typ 
[crates/typst-pdf/src/font.rs:201:5] &font = Font(Noto Color Emoji, Normal-400-FontStretch(1000))
thread 'main' panicked at crates/typst-pdf/src/font.rs:204:75:
called `Result::unwrap()` on an `Err` value: UnknownKind
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

laurmaedje:
I can't reproduce this either, also not by setting Noto Color Emoji manually.

LaurenzV:
I'm able to reproduce this on my Ubuntu virtual machine now, will try to look into it.

LaurenzV:
[...]

PR #5159, opened 2024-10-08
More robust glyph drawing

laurmaedje:
This

- Changes the logic which selects whether we export normal PDF text or use a Type 3 font to encode a glyph. In particular, this introduces a new `is_outlinable` function which returns true if:
  - The font has a glyf or CFF table
  - And the font doesn't contain a PNG (and thus color) bitmap, COLR, or SVG glyph for the glyph in question

  As a result, only glyphs that are actually outlinable should ever reach the subsetter, fixing https://github.com/typst/typst/issues/4572.

- The `glyph_frame` function does not panic anymore. Instead it will generate a fallback tofu if there was a conversion problem (and fail PDF/A export if enabled with "the glyph for {} could not be exported" including a hint for CFF2 fonts). This would trigger, for instance, for bitmap fonts that don't have a glyf or CFF table.

Fixes https://github.com/typst/typst/issues/4572

Alternative to https://github.com/typst/typst/pull/4929

### DIFF
--- a/crates/typst-render/src/text.rs
+++ b/crates/typst-render/src/text.rs
@@ -4,7 +4,7 @@ use pixglyph::Bitmap;
 use tiny_skia as sk;
 use ttf_parser::{GlyphId, OutlineBuilder};
 use typst::layout::{Abs, Axes, Point, Size};
-use typst::text::color::{glyph_frame, is_outlinable};
+use typst::text::color::{glyph_frame, should_outline};
 use typst::text::{Font, TextItem};
 use typst::visualize::{FixedStroke, Paint};
 
@@ -18,7 +18,7 @@ pub fn render_text(canvas: &mut sk::Pixmap, state: State, text: &TextItem) {
         let id = GlyphId(glyph.id);
         let offset = x + glyph.x_offset.at(text.size).to_f32();
 
-        if is_outlinable(&text.font, glyph) {
+        if should_outline(&text.font, glyph) {
             let state =
                 state.pre_translate(Point::new(Abs::raw(offset as _), Abs::raw(0.0)));
             render_outline_glyph(canvas, state, text, id);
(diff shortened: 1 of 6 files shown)


crates/typst/src/text/font/color.rs
```diff
@@ -6,51 +6,138 @@ use ttf_parser::{GlyphId, RgbaColor};
...
-    let glyph_id = GlyphId(g.id);
-    ttf.glyph_raster_image(glyph_id, 160).is_some()
-        || ttf.glyph_svg_image(glyph_id).is_some()
-        || ttf.is_color_glyph(glyph_id)
+    let glyph_id = GlyphId(glyph.id);
+    (ttf.tables().glyf.is_some() || ttf.tables().cff.is_some())
+        && !ttf
+            .glyph_raster_image(glyph_id, u16::MAX)
+            .is_some_and(|img| img.format == ttf_parser::RasterImageFormat::PNG)
+        && !ttf.is_color_glyph(glyph_id)
```
inferiorhumanorgans, line 23: `is_color_glyph` returns true if the glyph can be drawn with the `COLOR` table.  This is an outline format.

crates/typst-render/src/text.rs
```diff
@@ -18,20 +18,19 @@ pub fn render_text(canvas: &mut sk::Pixmap, state: State, text: &TextItem) {
         let id = GlyphId(glyph.id);
         let offset = x + glyph.x_offset.at(text.size).to_f32();
 
-        if is_color_glyph(&text.font, glyph) {
+        if is_outlinable(&text.font, glyph) {
```
inferiorhumanorgans, line 21: Has this been tested on a mac w/ Apple's emoji font?  The copy of Apple Color Emoji I have has mono outlines and color bitmaps.  I suspect this will prefer mono emoji.

laurmaedje: I tested it and it does work. If you look at the logic of `is_outlinable`, it prefers color tables. The naming doesn't make this clear enough, I'll rename it to `should_outline`.

PR #5168, opened 2024-10-09
Resolve lengths in math with scaled font size

### LINKED ISSUE #5164: Math size not accounted for when resolving `em` units
### Description

When using `em` units in math mode together with script sized content, the resolved length is not adjusted to the scaled font size.
For example:
<table>

<tr>
<th>With <code>1em</code> spacing</th>
<th>With unicode 'Em Space'</th>
</tr>

<tr>

<td>

```typ
$ a #h(1em) b \
  x^(a #h(1em) b) \
  x^x^(a #h(1em) b) $
```

</td>

<td>

```typ
$ a\u{2003}b \
  x^(a\u{2003}b) \
  x^x^(a\u{2003}b) $
```

</td>
</tr>

<tr>
<td>

image

</td>

<td>

image

</td>
</tr>

</table>

This also manifests itself when setting the size of an accent, as seen in [How to define uniformly larger accents in math? (Forum)](https://forum.typst.app/t/how-to-define-uniformly-larger-accents-in-math/983)

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

EpicEricEE:
Fixes #5164.

Strokes (for lines of cancel elements and augmented matrices) still go through `resolve` as they can't be resolved at a fixed font size as trivially.

I thought about doing something like the following, maybe even adding that function to the `Resolve` trait itself, but I'm not sure if that's the best way.
```rs
pub trait ResolveScaled: Resolve {
    fn resolve_scaled(self, styles: StyleChain, ctx: &MathContext) -> Self::Output {
        let local = 
            TextElem::set_size(TextSize(scaled_font_size(ctx, styles).into())).wrap();
        self.resolve(styles.chain(&local))
    }
}

impl<T: Resolve> ResolveScaled for T {}
```


laurmaedje:
Thanks! In the future, ideally I'd like to find a more general solution to the math sizing and resolving problem, but I think this is good for now.

PR #5177, opened 2024-10-10
Fix bug in `math.frac` layout code

### LINKED ISSUE #4855: The horizontal line of the square root is too high
### Description

In the mathematical formula `$ sqrt(n/p) < sqrt(n^(2/3)) < sqrt((n^(2/3))) $`, the horizontal line of the square root symbol is too high when it contains `n^(2/3)`, which makes the formula look unappealing.

square_root_bug

It's higher than when the square root contains `n/p`, and even higher than when it contains `n^(2/3)` with extra parentheses. 

Intuitively, adding parentheses could only make the height of the horizontal line stay the same or get higher, but here it actually became lower.

Here is a comparison with LaTeX.

LaTeX_square_root

### Reproduction URL

https://typst.app/project/rVGMrZzmtsyZOwndNwHkxe

### Operating system

Web app, Windows

### Typst version

mkorje:
Fixes #4855.

Turns out the actual problem was in the fraction layout code. The gap between the numerator/denominator and fraction rule was being calculated incorrectly. According to [this](https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#fractionnumeratorshiftup-fractionnumeratordisplaystyleshiftup-fractionnumeratorgapmin-fractionnumdisplaystylegapmin), the gap is measured from the top/bottom of the fraction rule, _not_ the axis height. Based on my testing in LuaTeX, the height of math content in Typst is now identical to LuaTeX (I didn't do extensive testing, I only tried a couple situations).

50 tests needed to be updated. I added one new test `math-frac-gap` to test `$ sqrt(n^(2/3)) $` from the linked issue.


laurmaedje:
Will trust you on this one. Thanks! :)

PR #5183, opened 2024-10-11
Add `column-gap` option to `math.equation`

### LINKED ISSUE #2331: Customizable column-pair spacing for equations
### Description

The current behavior of the equation environment is like that of LaTeX' (or amsmath rather) _alignat_ environment. This is because no space is added between adjacent LR column pairs.

It would be extremely useful to have a gutter option to add spacing between column pairs, to match the _align_ environment. This should possibly even have a reasonable spacing by default such as matching the one from _align_.

### Use Case

The main motivation is that the vast majority of the time, you do want some spacing between column pairs, and this would reduce the number of times you have to add it in manually. A further benefit is that the code is more readable without quads all over the place.

mkorje:
Closes #2331.

It's all in the title.

I've set the default value to `1.5em`. I eyeballed it to match some reference images in LaTeX, and it seems reasonable enough. Though I'd appreciate anyone other opinions on what the default should be.

The name of the field `column-gap` is maybe not great. It's consistent with matrices, though.


crates/typst/src/math/mod.rs
```diff
@@ -128,6 +128,9 @@ use crate::text::TextElem;
 /// because just one `&` precedes it. Each alignment point simply alternates
 /// between right-aligned/left-aligned.
 ///
+/// By default, there is a gap of `1.5em` added between columns. You can
+/// modify this with the [`column-gap`]($math.equation.column-gap) parameter.
+///
 /// ```example
 /// $ (3x + y) / 7 &= 9 && "given" \
 ///   3x + y &= 63 & "multiply by 7" \
```
mkorje, line 136: To better showcase the gap added, we could change this example to
```typ
$ (3x + y) / 7 &= 9        &                &"given" \
        3x + y &= 63       & "multiply by 7"& \
            3x &= 63 - y   &                &"subtract y" \
             x &= 21 - y/3 &   "divide by 3"& $
```
test

xkevio:
I don't know if we should set the default to non-zero values. LaTeX technically has two math environments: `\alignat` inserts no horizontal space just like Typst does currently while `\align` inserts the horizontal space. So we aren't doing anything wrong per se, we are just emulating a different environment.

However, people do always ask how to achieve this and why Typst currently doesn't support it so it seems people are more used to `\align` anyway but it seems like a rather big breaking change for something that does not really have to be breaking.

mkorje:
Like you said, I figure because `align` seems more common, and what users expect, that then the default should not be zero.

Also, even though it is a breaking change, I guess it is a "good" one. Users probably have `quad`s manually placed everywhere to add spacing. No longer needing to do that, and having a reasonable default, is a good change imo even if it is breaking.

xkevio:
Yeah I am not too sure what my final opinion is on this either, tbh. It is certainly a really good feature to have now and I would assume most people typeset equation environments thinking the typesetting application will handle most of the layout so this would inch closer to the presumably expected experience. And fixing an old document is just one set-rule away—though similarly to `par.spacing`, there should probably be some kind of warning for at least one update cycle.

crates/typst/src/math/equation.rs
```diff
@@ -177,6 +190,7 @@ impl Show for Packed<EquationElem> {
 impl ShowSet for Packed<EquationElem> {
     fn show_set(&self, styles: StyleChain) -> Styles {
         let mut out = Styles::new();
+        out.set(EquationElem::set_column_gap(self.column_gap(styles)));
```
xkevio, line 193: Shouldn't this be inside the first if-branch below since, I assume, the column gap will only apply to block level equations?

mkorje: I guess it doesn't really matter, but yeah probably. I've moved it into the first if-branch.

mkorje:
The only issue I see with adding a warning is that we can't really target it, if that makes sense. Unlike with `set par(spacing: ..)` replacing `show par: set block(spacing: ..)`, the old behaviour is changing _every_ block equation. I think adding a warning for this would end up being a little ridiculous then.

laurmaedje:
After the discussion Discord (starting [here](https://discord.com/channels/1054443721975922748/1176478139757629563/1301210778116030474)) and, in particular, [this comment](https://discord.com/channels/1054443721975922748/1176478139757629563/1301230438911705168) I'm unsure how to proceed here.

My main concern is that this is _quite_ the breaking change and a silent one at that: You don't even get a warning or an error for equations that need to be updated. That's a price I'm willing to pay, but only if we're actually 100% sure that this is what we want.

xkevio:
I don't think the linked to comment holds much value, if people want to use it like columns and if `columns` / `grid` isn't up to par currently regarding baselines, then I think it is fine if the user "abuses" this to emulate columns within equations.

Personally, I am for keeping the default value at 0, the feature still exists then. Downside ofc is that it then has to be discovered by the user and unlike LaTeX, this isn't two different functions but the same with a different argument. If not, then I still think a warning would be nice though as was said, we cannot really just slap a warning on every equation since nothing there is deprecated or actually wrong.

mkorje:
I don't really have much of an opinion on whether the default should be 0 or not. Though given that this is a large breaking change (and a silent one at that) and it's quite easy for a user to change themselves, maybe we just keep the default as 0 then?

laurmaedje:
Most people on Discord were in favor of the default spacing and I think I'm getting too afraid of breaking changes ... let's just do the breaking change with default value.

mkorje:
Ok. Any opinion on the default value?

Also, before this gets merged, I may as well make it so that it can take an array of gaps as well (similar to as is done in grid).

laurmaedje:
Matching LaTeX's default values makes sense imo.

xkevio:
> I believe the default in LaTeX is controlled by `\minalignsep`, which is set to `10pt`. So given 10pt is the default font size, how about we make the default `1em`?

Default in Typst is actually `11pt`.

mkorje:
Sorry, I meant the default in LaTeX is 10pt (or technically 10.05 or smth).

laurmaedje:
I just checked and it looks like LaTeX align does add a lot more space by default and it varies by number of columns.

mkorje:
<details>

<summary>Current state of this feature</summary>

[Output](https://github.com/user-attachments/files/18423278/playground.pdf) of the below code:

```typ
#set page(width: auto)

#let test-equations = [
  // No alignment points.
  $ a = b \
    c = d $

  // One column.
  $ a &= b \
    c &= d $

  // Two columns.
  $ a &= b & c &= d \
    e &= f & g &= h $

  // Three columns with last one empty.
  $ a &=b & c&=d & \
    e &=f & g&=h $

  // Three columns.
  $ a &= b & c &= d & e &= f \
    g &= h & i &= j & k &= l $

  // Four columns with inferred alignment
  // point in last one.
  $ a &= b & c &= d & e &= f & 1 &= 3 \
    g &= h & i &= j & k &= l & 2 $
]

#let test(..args) = grid(
  columns: (16em,) * 3,
  column-gutter: 4em,
  stroke: 1pt + black,
  [
    #show math.equation.where(block: true): set align(left)
    #set math.equation(..args)
    #test-equations
  ],
  [
    #set math.equation(..args)
    #test-equations
  ],
  [
    #show math.equation.where(block: true): set align(right)
    #set math.equation(..args)
    #test-equations
  ],
)

// Old default.
#test(column-gap: 0em, column-padding: 0em)
#pagebreak()

// New default. Notice that align isn't quite as
// you'd expect. Should fractional spacing as
// padding get collapsed on the side align is set
// to?
#test(column-gap: 1fr, column-padding: 1fr)
#pagebreak()

#test(column-gap: 10%, column-padding: (1em, 0em))
#pagebreak()

#test(column-gap: (1em, 10%), column-padding: 5%)
#pagebreak()

#test(column-gap: (1em, 1fr), column-padding: 5%)
#pagebreak()

// Overflow! Notice that align is working.
#test(column-gap: (50%, 2fr, 1fr), column-padding: (0.5fr, 0.5em))
```

</details>

See the above for an example of how this feature is now working. Some important points:

- Not sure whether `column-gap` and `column-padding` are great parameter names...
- The new default is that both `column-gap` and `column-padding` are set to `1fr`. This is quite the breaking change.
- If you use any fractional spacing in `column-gap` or `column-padding`, it means the equation's width will take up all of the available space; this makes sense, as you've told the compiler to divide _all_ of the remaining space among the fractional lengths. This means, however, that `#show math.equation.where(block: true): align(left)` doesn't work as one might expect.
  Perhaps, we should make it so that any fractional spacing as column-padding gets removed on the side alignment is currently set to?

Implementation wise, I still need to add tests and docs, as well as possibly clean up the code a bit.

jfsalzmann:
@mkorje is this still under development? I was happy to learn that people are working on a fix. I wasn't quite able to find a workaround to influence the column gap width..

mkorje:
@jfsalzmann Yes, I do plan to keep working on this. Though some discussion is still needed as to the default I think - see my comment above https://github.com/typst/typst/pull/5183#issuecomment-2592416868. Any feedback on that would be appreciated!

laurmaedje:
How about doing an opinions needed post on this in the forum? I think https://forum.typst.app/t/opinions-needed-likely-upcoming-changes-to-math-mode-precedence/4974/36 was a success.

mkorje:
> How about doing an opinions needed post on this in the forum? I think https://forum.typst.app/t/opinions-needed-likely-upcoming-changes-to-math-mode-precedence/4974/36 was a success.

Sorry for the delay, I've been quite busy. Sounds like a good idea. I'll draft something soon.

Issue #5057, opened 2024-09-27
Add ability to repeat the caption (or a modified caption) of a multi-page table in subsequent pages


ami-null:
### Description

Often we may have to report a table that's just too long. In such situations, it is customary to repeat the table caption if the table spans multiple pages. For example, the caption in the first page might be, "Table 1: monthly revenue and expenditure for the last 5 years", and the caption in the subsequent pages would be, "Table 1: monthly revenue and expenditure for the last 5 years (*continued*)".

In LaTeX, I can write the following:

```tex
\documentclass{article}
\usepackage{longtable}
\usepackage[margin=1.5cm]{geometry}

\begin{document}
\vspace*{0.85\textheight}

\begin{longtable}{|c|c|c|}

% Caption and header for the first page

\caption{Caption for the first page} \label{longtable1} \\
\hline
Header 1 & Header 2 & Header 3 \\
\hline
\endfirsthead

% Caption and header for subsequent pages
\caption{Caption for subsequent pages} \\
\hline
Header 1 & Header 2 & Header 3 \\
\hline
\endhead

% Footer
\hline
\endfoot

% Table data
Row 1.1 & Row 1.2 & Row 1.3 \\
Row 2.1 & Row 2.2 & Row 2.3 \\
Row 3.1 & Row 3.2 & Row 3.3 \\
Row 4.1 & Row 4.2 & Row 4.3 \\
Row 5.1 & Row 5.2 & Row 5.3 \\
Row 6.1 & Row 6.2 & Row 6.3 \\
Row 7.1 & Row 7.2 & Row 7.3 \\
Row 8.1 & Row 8.2 & Row 8.3 \\
Row 9.1 & Row 9.2 & Row 9.3 \\
\end{longtable}

\end{document}
```

Which gives me:
![ce2520612a1da8933a5396e8d3a5066394933ace](https://github.com/user-attachments/assets/a588ee7e-3512-4458-9666-39b2640657d7)

To the best of my knowledge it is not possible to do in Typst right now.

### Use Case

I believe that the usefulness of this feature is self-evident. Multi-page tables are pretty common in situations including, but not limited to: when dealing with data from various sources, or when reporting regression results.

Andrew15-5:
This issue was mentioned in https://forum.typst.app/t/how-to-repeat-the-caption-or-a-modified-caption-of-a-multi-page-table/729/3.

I'm pretty sure this is the official way to write captions for a multipage table in Russia. So this is a very important feature that I also would like to have.

[...]

PR #5191, opened 2024-10-12
Repeatable figure captions

PgBiel:
Closes https://github.com/typst/typst/issues/5057

In principle, I took the low-hanging fruit and just slapped grid with headers/footers onto our figure.

This also has the advantage of being fully reproducible in figure show rules.

# Discussion

This code is quite small and already usable, but there are a few design decisions that should be taken into account before merging:

1. Should repeat be enabled by default? It does add the overhead of a grid to all figures, but one workaround for that is to only add it when the figure is actually breakable (it isn't by default). And it does seem like a sensible default to me. For now, though, it's opt-in.
2. One common use case (as shown in the issue) is to have a "Continuation" marker for captions. While it is already possible in this PR using `context` and counter steps, we could want to have some built-in mechanism, e.g. provide a function to the caption to run it once for each page. Though it would be nice to be consistent with headers and footers here, so maybe that should ship as a separate package.


PgBiel:
Drafting until further discussion.

laurmaedje:
A context based design sounds nicer to me than a function.

laurmaedje:
Let's discuss this in our next meeting.

PgBiel:
I agree continuation handling is important - repetition by itself is already handy for certain use cases, but that can be reproduced in pure Typst fairly easily.

Do we have any new proposals regarding how it should be done? Maybe the community has further ideas as well.

laurmaedje:
I don't have concrete proposals. But I could imagine some kind of element that you can put around stuff to wrap it and provide customized continuations.

laurmaedje:
I'll close this for now. We can reopen it if we're more sure about how we want to go about this.

reknih:
@SoappGuy For updates, subscribe to #5057. You can help move this feature along by sharing yet undiscussed requirements and by participating about the discussion about design and API there and on Discord. Please refrain from asking about updates without any such input. When subscribed to the issue, you'll be notified of updates when they happen.

PR #5196, opened 2024-10-13
Fix the order to match the text

fynsta:
The text says "First is the page set rule.", so it should also come first in the example.


laurmaedje:
Thanks!

PR #5198, opened 2024-10-13
(Re-)implement rounding with negative digits

### LINKED ISSUE #5155: (Re-)add support for `calc.round()`ing to integer digits
### Description

Before Typst 0.12, it was possible to round to integer digits using [`calc.round()`](https://typst.app/docs/reference/foundations/calc/#functions-round) by passing a negative `digits` argument. Apparently, this "feature" was rather [based on a mistake and somewhat broken](https://github.com/typst/typst/pull/4900#issuecomment-2376919375). 

This feature request proposes to properly add support for this (at least for the `float` type) since there is no fundamental difference between integer and fractional digits for floating point numbers. 

## Some edge cases
- If the input for the number is `inf`, `-inf` or `nan`, the input should be returned -- regardless of the `digits` argument. 
- If `digits` is less than `-308`, i.e., rounding is performed on the 308th integer digit or higher, the result should be always `0.0`. This is because the largest IEEE-754 float is `1.7976931348623157e+308`  and then the digit to round to is always larger than the number of digits in the number. 
- Between `0` and `-307` digits, there should be no overflow issues regarding floating point arithmetic. 
- The question remains what to do if `x>=1.5e+308` and `digits=-308` because  `round(x * pow(10,digits)) / pow(10, digits)` will produce an overflow since the result is not representable by floating point numbers. This is an extreme edge cases of course and returning `inf` or throwing an error would probably be acceptable. 

For reference, a remarkable implementations is the built-in `round()` from CPython. Apparently, it partly does rounding by converting to a string and back to achieve the most precise results which solves some issues that rounding via $round(x·10^d)/10^d$ has. However, a bit overkill here imo. 

I don't know exactly the cases where the old rounding function failed to produce a correct result, so if anyone wants to add on that, feel free to so they can be discussed. 

### Use Case

It is approximately as useful as rounding to fractional digits.

PgBiel:
Closes #5155

Implements rounding with negative digits for floats, decimals and ints. ~~(for those, I just chose to re-use the decimal impl, since it doesn't use any imprecise operations).~~ (**EDIT:** Using fully separate procedures for each type now.)

I'd suggest some good amount of scrutiny before merging. :sweat_smile:

In principle, all seems good to me though.

## Some important details

- `calc.round` can now error if rounding decimals or ints above the max / min.
- `calc.round` can now return +inf or -inf if rounding floats above max / min (respectively).

### DIFF
--- a/crates/typst-utils/src/lib.rs
+++ b/crates/typst-utils/src/lib.rs
@@ -17,7 +17,7 @@ pub use self::deferred::Deferred;
 pub use self::duration::format_duration;
 pub use self::hash::LazyHash;
 pub use self::pico::PicoStr;
-pub use self::round::round_with_precision;
+pub use self::round::{round_int_with_precision, round_with_precision};
 pub use self::scalar::Scalar;
 
 use std::fmt::{Debug, Formatter};
--- a/crates/typst-utils/src/round.rs
+++ b/crates/typst-utils/src/round.rs
@@ -166,27 +166,27 @@ mod tests {
     #[test]
     fn test_round_with_precision_negative_1() {
         let round = |value| rp(value, -1);
-        assert_eq!(0.0, round(0.0));
-        assert_eq!(-0.0, round(-0.0));
-        assert_eq!(0.0, round(0.4));
-        assert_eq!(-0.0, round(-0.4));
-        assert_eq!(1230.0, round(1234.5));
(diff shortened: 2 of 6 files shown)


PgBiel:
Drafting as I'll take a stab at not using decimals for ints.

Mc-Zen:
Awesome!

PR #5202, opened 2024-10-14
Add `math.accent` support for `flac` and `dtls` OpenType features

### LINKED ISSUE #2860: Prime symbol broken with Libertinus Math

### LINKED ISSUE #5201: Use `flac` and `dtls` OpenType features in `math.accent`
### Description

Example from [Mathematical Typesetting](https://learn.microsoft.com/en-us/typography/cleartype/pdfs/cambriamath.pdf):
eg_page-0001

### Use Case

Better maths typesetting.

mkorje:
Closes #5201.

While implementing this, I also changed the logic for the `ssty_table` to use the `GlyphwiseSubsts<'a>` struct. This saves some code duplication and is more consistent anyways. Also, from the [spec](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#ssty), it could be implemented as a single substitution. So this change also ensures `ssty` variants will work in fonts where this is the case.

The only font I can find that supports the `flac` OpenType feature is Cambria Math. I tested it locally, and it is all working. Though it (my probably dodgy copy) had a bug I needed to manually fix so the horizontal variants of some of the `flac` variants would be used. See the example below.
test
```typst
--- math-accent-flattened ---
// Test flattened accent glyph variants.
#show math.equation: set text(font: "Cambria Math")
$hat(a) hat(A)$
$tilde(w) tilde(W)$
$grave(i) grave(j)$
$grave(I) grave(J)$
```

EDIT: Also fixes #2860.


mkorje:
The changes I made making `ssty_table` use the `GlyphwiseSubsts<'a>` struct fixes #2860 as well. https://github.com/typst/typst/issues/2860#issuecomment-2189897091

In the future we should pass single chars in math through the shaper.

crates/typst/src/math/accent.rs
```diff
@@ -114,7 +114,12 @@ impl LayoutMath for Packed<AccentElem> {
     #[typst_macros::time(name = "math.accent", span = self.span())]
     fn layout_math(&self, ctx: &mut MathContext, styles: StyleChain) -> SourceResult<()> {
         let cramped = style_cramped();
-        let base = ctx.layout_into_fragment(self.base(), styles.chain(&cramped))?;
+        let mut base = ctx.layout_into_fragment(self.base(), styles.chain(&cramped))?;
+
+        // Try replace a glyph with its dotless variant.
+        if let MathFragment::Glyph(ref mut glyph) = &mut base {
+            glyph.make_dotless_form(ctx)
```
khaledhosny, line 121: `dtls` feature should not be applied unconditionally, for example people might want $\hat{\imath}$ but $\widehat{\dot{\imath}+\dot{\jmath}}$. LaTeX has `\imath` and `\jmath` macros to explicitly request the dotless form.

mkorje: Are there any situations where you'd want an accent above i or j, and keep the dot? Also, I think Typst has `dotless.i` and `dotless.j` in `sym` that correspond to the LaTeX `\imath` and `\jmath` macros. We should add a function though so that we get the dotless variants of the blackboard bold, caligraphic, etc. variants. I'm not sure how LaTeX does this.

wrzian: I feel like I would expect those to just be available as `$bb(dotless.i)$`?

mkorje: The problem is that `dotless.i` is a Unicode char, but the variants in the `dtls` feature are from the font. I don't think there exist codepoints for blackboard bold, caligraphic etc. dotless i and j.

wrzian: Well could we look those up when we convert to a glyph in `styled_char` and give a warning if they don't exist? I wouldn't want a user to have to care about other functions if it lets us provide a consistent interface.

khaledhosny: LaTeX does not use `dtls` feature yet (https://github.com/latex3/unicode-math/issues/34). Since fonts might be inconsistent here, `dotless.i` and `dotless.j` should check if the font has `dtls` feature and make them the equivalent of `i` and `j` with `dtls` feature applied (i.e. they would switch alphabets when requested), otherwise use the dotless i and j Unicode code points.

crates/typst/src/symbols/sym.rs
```diff
@@ -992,5 +992,5 @@ pub(crate) const SYM: &[(&str, Symbol)] = symbols! {
     kelvin: 'K',
     Re: 'ℜ',
     Im: 'ℑ',
-    dotless: [i: '𝚤', j: '𝚥'],
+    dotless: [i: 'ı', j: 'ȷ'],
```
mkorje, line 995: This change makes this character act more in line with other Latin characters in math. (Some) Single characters in math are converted to their italic math variants, so the dotless i and j should as well. See the example below.

<table>
<tr>
<td> Code </td> <td> Render </td>
</tr>
<tr>
<td>

```typ
$ 𝑗 upright(𝑗) italic(𝑗) bb(𝑗) med
  𝚤 upright(𝚤) italic(𝚤) bb(𝚤) \
  j upright(j) italic(j) bb(j) med
  ı upright(ı) italic(ı) bb(ı) $
```

</td>
<td>

test

</td>
</tr>
</table>

crates/typst/src/math/accent.rs
mkorje, line 121: (Moving part of the old conversation here.) Are there any situations where you'd want an accent above i or j _with_ the dot still? If this is basically never, it'd be much nicer to automatically replace them with their dotless variants.

mkorje (commented):
I've made the changes so that `$bb(dotless.i)$` (along with all the other variants) work as expected, following @khaledhosny comments to switch if the font has the `dtls` OpenType feature.

See the example below.
<table>
<tr>
<td>

```typ
$ dotless.i dotless.j \

  bb(dotless.i) bb(dotless.j) \
  cal(dotless.i) cal(dotless.j) \
  frak(dotless.i) frak(dotless.j) \
  mono(dotless.i) mono(dotless.j) \ 
  sans(dotless.i) sans(dotless.j) \
  upright(dotless.i) upright(dotless.j) \
  bold(dotless.i) bold(dotless.j) \

  bold(frak(dotless.i)) upright(sans(dotless.j)) \
  italic(bb(dotless.i)) frak(sans(dotless.j)) $
```

</td>
<td>

test

</td>
</tr>
</table>

However, the only "issue" is that if the font does not have a dotless variant of the glyph, it will sometimes fall back to the glyph with the dot (this depends on the order in which variants and styles are applied). This feels in some ways like bad and unexpected behaviour, but at the same time I can see this as being fairly reasonable. I think it'd also increase the code's complexity a fair bit to check whether the glyph exists in the `dtls` alphabet beforehand.

crates/typst/src/math/accent.rs
khaledhosny: Some orthographies keep the dots of the i and j when accented, but I don’t know if there are equivalent math notations. There is at least one question on tex.stackexchange where this was desired https://tex.stackexchange.com/questions/181384/diacritics-above-dotted-i-i-with-tittle (though since you need to explicitly use `\imath` and `\jmath` to drop the dot in LaTeX, it is unlikely to have many questions asking about how to keep the dots).

crates/typst-layout/src/math/fragment.rs
```diff
@@ -390,20 +388,35 @@ impl GlyphFragment {
...
-            script_alternatives(ctx, self.id).and_then(|alts| alts.alternates.get(0));
+            ctx.ssty_table.as_ref().and_then(|ssty| ssty.try_apply(self.id, None));
+        if let Some(alt_id) = alt_id {
+            self.set_id(ctx, alt_id);
+        }
+    }
 
+    pub fn make_script_script_size(&mut self, ctx: &MathContext) {
+        let alt_id = ctx.ssty_table.as_ref().and_then(|ssty| {
+            ssty.try_apply(self.id, Some(1))
```
laurmaedje, line 401: This `1` maybe deserves a constant or comment.

mkorje: I've added a comment explaining its meaning.

laurmaedje:
Thank you!

> I think it'd also increase the code's complexity a fair bit to check whether the glyph exists in the dtls alphabet beforehand.

Let's not do that (at least for now) then.

PR #5204, opened 2024-10-14
Put floats back above footnotes instead of below

laurmaedje:
LaTeX puts footnotes above floats, while Typst 0.11.1 did not. This changed in 0.12.0-rc1, but this surprised most people, so I'm changing it back. While this is a little less consistent w.r.t column vs page floats (page floats continue to be below footnotes because footnotes are per column), it's what most people (including me tbh) would intuitively expect.

We can consider adding configuration for this in the future, but I'd first like to wait until people with actual use cases request this.


xkevio:
Maybe just nitpicking but maybe it would be a good idea to leave a comment referencing this PR and behavior wrt. to LaTeX right in the source code? Since it is just as simple as moving this for-loop, it could be easy to miss and maybe nice to see for people studying the source code why the decision is the way it is.

PR #5210, opened 2024-10-15
Greet users who run `typst` for the first time

laurmaedje:
This adds a greeting to people who run juse bare `typst` for the first time, explaining the most useful commands and linking some further resources. Also adds the same useful resources to the help output.


eduardz1:
Such a nice small addition!

Issue #2499, opened 2023-10-26
PDF file attachments


miliog:
### Description

I'd like to be able to create PDF attachments.

![ed08 png img](https://github.com/typst/typst/assets/146289769/17d50020-cc8c-46ed-911c-0afaed56b10a)
(Screenshot from https://helpx.adobe.com/acrobat/using/links-attachments-pdfs.html#)

### Use Case

It's very convenient to add file attachments to the PDF itself. Not only Acrobat Reader, but also many open source readers support it.

gfpj:
Here is an example code in LaTeX that shows how to embed the file in the list of embedded files, set the flags and also reuse the embedded binary in a File Attachement Annotation. Maybe it helps the person who wants to implement this.

```latex
\pdfcompresslevel=0
\pdfobjcompresslevel=0

\documentclass[a4paper]{article}

\usepackage{embedfile}
\usepackage{attachfile2}
\expandafter\xdef\csname theatfi@flags\endcsname{704}
\newcommand{\clickablefile}[2]{%
\embedfile[ucfilespec={#1}]{#1}%
%\expandafter\xdef\csname atfi@embed@file@#1\endcsname{\the\numexpr\pdflastobj-1}%
\expandafter\xdef\csname atfi@fileobj@#1\endcsname{\the\numexpr\pdflastobj-1}%
\textattachfile[ucfilespec={#1},icon={Paperclip}]{#1}{#2}%
}

\begin{document}

\clickablefile{test.txt}{Double click to export the file}

\end{document}
```
[embed.pdf](https://github.com/typst/typst/files/13181473/embed.pdf)

tingerrr:
This seems like a good-contribution, the relevant sections in the PDF 1.7 spec are 12.5.6.15 and 7.11.4, these features would have to be upstreamed into pdf-wirter before typst can support this.

awehrfritz:
> ... these features would have to be upstreamed into pdf-wirter before typst can support this.

As far as I can tell, `pdf-writer` has support for this already: https://github.com/typst/pdf-writer/commit/ba040ace792d57a2e57e4973a38abb1561cdc057

NiklasEi:
An early attempt at implementing file attachments associated with the document as a whole: https://github.com/NiklasEi/typst/tree/file-embedding

There are many workarounds and how I implemented this is surely not the intended way. I would be glad for any constructive feedback :slightly_smiling_face: 

The document
```typst
Testing file embedding

#embed("factur-x.xml")
```

[...]

PR #5221, opened 2024-10-16
Embed files associated with the document as a whole

### LINKED ISSUE #4411: PDF A-3 - embeding xml into the exported document
### Description

I would like to be able to embed XML data into the output PDF following the PDF/A-3 standard. I see that A-2 support is planned and tracked in #2942. Is Typst planning to eventually also support A-3?

### Use Case

I am looking into generating electronic invoices with Typst. If I understand it correctly, this will often require embedding the invoice data into the PDF, which is specified in PDF/A-3.

A bit more context specific to Germany(/EU):
In B2B contexts, Germany [requires electronic invoices from 1.1.2025 on](https://ec.europa.eu/digital-building-blocks/sites/display/DIGITAL/eInvoicing+in+Germany). This will require embedding the invoice data into the PDF. A standard developed for that is https://de.wikipedia.org/wiki/ZUGFeRD (sorry, I could only find a German resource), which requires PDF/A-3.

NiklasEi:
Resolves #4411
Part of #2499 

This PR implements file embeddings associated with the document as a whole. The attachments do not have visible annotations. It also adds an option for PDF/A-3 to allow arbitrary file embeddings with PDF/A compliance. I will gladly move this to a second PR if preferred.
The embeddings are part of a new `pdf` module:

```typst
This will embed the file without a description
#pdf.embed("text.txt")

We can also optionally set description, mime type and relationship of the embedding
#pdf.embed("text.txt", description: "An embedded file", mime-type: "text/plain", relationship: "supplement")

Or embed any bytes into the PDF as a file.
#let some_bytes = read("test.txt")

With just a file name...
#pdf.embed.decode("test.txt", some_bytes)

or all the other additional information
#pdf.embed.decode("a_file_name.txt", some_bytes, description: "A description", mime-type: "text/plain", relationship: "supplement")
```


crates/typst-pdf/src/catalog.rs
```diff
@@ -178,6 +178,23 @@ pub fn write_catalog(
...
+            let mut name_dict = catalog.names();
+            let mut embedded_files = name_dict.embedded_files();
+            let mut names = embedded_files.names();
+            for (name, file_ref) in &ctx.references.embedded_files {
+                names.insert(Str(name.as_bytes()), *file_ref);
+            }
+        }
+
+        // Todo: this is for PDF 2.0; we should probably check the exporting version
+        let mut associated_files = catalog.insert(Name(b"AF")).array().typed();
```
NiklasEi, line 192: My document reader and an online PDF/A validator do not complain about this, but as far as I understand, this is for PDF 2.0 and we are exporting a 1.7 document. Can this stay as it is?

NiklasEi: I have the same question for `AFRelationship` in the file spec.

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,44 @@
...
+        let embedded_file_stream_ref = chunk.alloc.bump();
+        let file_spec_dict_ref = chunk.alloc.bump();
+
+        let length = embed.data().len();
+        let mut embedded_file =
+            chunk.embedded_file(embedded_file_stream_ref, embed.data().as_ref());
+        embedded_file
+            .subtype(Name(b"text/xml"))
+            .pair(Name(b"Length"), length as i32);
+        embedded_file.params().modification_date(Date::new(2023)).finish(); // Todo: can we just let this out?
```
NiklasEi, line 22: Maybe use the document creation date here; if we have one?

crates/typst/src/embed/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Layout the embedding.
+#[typst_macros::time(span = elem.span())]
+fn layout_embedding(
+    elem: &Packed<EmbedElem>,
+    _: &mut Engine,
+    _: Locator,
+    _: StyleChain,
+    _: Region,
+) -> SourceResult<Frame> {
+    let mut frame = Frame::hard(Size::zero());
```
NiklasEi, line 108: Is this the intended way to layout an item that shouldn't influence the layout?

crates/typst-pdf/src/catalog.rs
LaurenzV: If it's not 1.7, we should probably remove it. Or is this necessary for embedded files to work?

crates/typst-pdf/src/embed.rs
LaurenzV: We do create a document date somewhere, check out the code where we write the XMP metadata.

crates/typst-pdf/src/catalog.rs
NiklasEi: Both come from the ZUGFeRD standard and can probably be removed for "normal" file embeddings. I will try it out.

reknih: The `AF` and `AFRelationship` keys for PDF 1.7 are described in the normative ISO 19005-3 (PDF/A-3) Annex E. They may be present in a PDF 1.7 file exactly if the file claims PDF/A-3 compliance.

NiklasEi: Perfect, thank you. Then I will put them behind a check for PDF/A-3.

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,67 @@
...
+    ctx: &WithGlobalRefs,
+) -> SourceResult<(PdfChunk, HashMap<EcoString, Ref>)> {
+    let mut chunk = PdfChunk::new();
+    let mut embedded_files = HashMap::default();
+    if !ctx.resources.embeds.is_empty()
+        && Some(PdfAConformanceLevel::A_2) == ctx.options.standards.pdfa
+    {
+        bail!(
+            Span::detached(),
+            "file embeddings are currently only supported for PDF/A-3"
```
NiklasEi, line 19: This would technically be allowed if the embedded file is PDF/A-2. However, as I didn't know how to check that here, I decided to restrict file embeddings to PDF/A-3.

Enivex: Shouldn't the message reflect that this is a limitation imposed by pdf/a-2, not typst?

NiklasEi: PDF/A-2 does allow file embeddings as long as the files are also PDF/A-2. So the restriction is part Typist and part PDF/A-2.

I am open for iterating on the message.

sirchnik:
`#embed` is causing pagebreaks for me when using before `#set page("a4")` or after `#pagebreak(weak: true)`.
<details>
<summary>Reproduction</summary>

```typ
#embed.decode("bill-reports.pdf", "test.txt")
#set page("a4")
= header
#pagebreak(weak: true)
#embed("bill-reports.typ")
```

_Hope that's the right place for this comment_

NiklasEi:
@sirchnik the implementation was changed with some pointers from LaurenzV on Discord. The layouting issues should be fixed now.

One thing I haven't managed so far, but I still plan to do, is to move the embed methods into a `PDF` scope (so it's `#pdf.embed(test.txt)` and `#pdf.embed.decode("some file content")`).

laurmaedje:
I would prefer to drop the `decode` variant. We plan to deprecate the `decode` functions and instead allow passing bytes directly to the individual functions instead.

SolidTux:
From the proposal it's not completely clear yet, if and how it would be possible to link to these files. Is it just possible to add a label to the `#pdf.embed` call? Or is this not possible from a PDF point of view anyway?

NiklasEi:
Embedded file streams can be included in PDF 1.7 in two different ways:
* associated with the whole file (this is what this PR implements)
* associated with a specific location in the PDF with a File Attachment Annotation (12.5.6.15 in ISO 32000-1)

The second would give you a clickable name, description and icon in the PDF file to download the attachment. These annotations could be added in a follow-up PR and point to the same embedded file streams that this PR adds to the PDF export.

NiklasEi:
@laurmaedje Would it be OK to keep the decode variant here and leave a replacement for all the decode methods to a follow-up PR?

laurmaedje:
I guess we can do that. Is the PR ready for review?

NiklasEi:
I will add back the decode variant then (in a couple hours) :+1:
Apart from that the PR is ready, yes.

NiklasEi:
@laurmaedje everything is pushed; ready for review.

crates/typst-pdf/src/lib.rs
```diff
@@ -142,6 +154,29 @@ pub enum PdfStandard {
...
+    /// PDF/A-3b.
+    #[serde(rename = "a-3b")]
+    A_3b,
+}
+
+/// Conformance levels of PDF/A
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[allow(non_camel_case_types)]
+#[non_exhaustive]
+pub enum PdfAConformanceLevel {
```
laurmaedje, line 166: This should not be public. We intentionally do not provide such an enum to keep the API flexible. Internally, we can turn `&[PdfStandard]` into whatever is most convenient at the time, rejecting bad combinations.

I think as is the enum is a bit strange because it drops the "b".

As various checks are required by different standards (e.g. all your `is_some()`) checks, another design might be to not store level enums, but rather a bunch of booleans specifying which checks should be performed. 

If we like this bool-based design, maybe for now we could realize it with just `pdfa: bool, embedded_files: bool` and later break out the `pdfa: bool` into separate ones. And for the XMP metadata we could just store a field `(u8, &'static str)` (e.g. `(2, "B")`).

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::{PdfAConformanceLevel, PdfChunk, WithGlobalRefs};
+use ecow::EcoString;
+use pdf_writer::{Finish, Name, Ref, Str, TextStr};
+use std::collections::HashMap;
+use typst_library::diag::{bail, SourceResult};
+use typst_library::foundations::NativeElement;
+use typst_library::pdf::embed::{Embed, EmbedElem, EmbeddedFileRelationship};
+use typst_syntax::Span;
+
+pub fn write_embedded_files(
```
laurmaedje, line 11: Please add brief doc comments to all functions.

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,90 @@
...
+    ctx: &WithGlobalRefs,
+) -> SourceResult<(PdfChunk, HashMap<EcoString, Ref>)> {
+    let mut chunk = PdfChunk::new();
+
+    let elements = ctx.document.introspector.query(&EmbedElem::elem().select());
+    if !elements.is_empty()
+        && Some(PdfAConformanceLevel::A_2) == ctx.options.standards.pdfa
+    {
+        bail!(
+            Span::detached(),
```
laurmaedje, line 21: the span here should be that of the problematic `EmbedElem`

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    let mut file_spec = chunk.file_spec(file_spec_dict_ref);
+    file_spec
+        .path(Str(embed.path().as_bytes()))
+        .unic_file(TextStr(embed.path().as_str()))
+        .insert(Name(b"EF"))
+        .dict()
+        .pair(Name(b"F"), embedded_file_stream_ref)
+        .pair(Name(b"UF"), embedded_file_stream_ref)
+        .finish();
```
laurmaedje, line 67: This finish here should be removed. It doesn't have any effect, but is not needed and confusing.

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,90 @@
...
+        embedded_file.subtype(Name(mime_type.as_bytes()));
+    }
+    let date = ctx
+        .document
+        .info
+        .date
+        .unwrap_or(ctx.options.timestamp)
+        .and_then(|date| pdf_date(date, ctx.document.info.date.is_auto()));
+    if let Some(pdf_date) = date {
+        embedded_file.params().modification_date(pdf_date).finish();
```
laurmaedje, line 55: Does it make sense to say that the modification date of the _embedded file_ is the _document's_ date?

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+    embedded_file.finish();
+
+    let mut file_spec = chunk.file_spec(file_spec_dict_ref);
+    file_spec
+        .path(Str(embed.path().as_bytes()))
```
laurmaedje, line 61: Not sure how much sense it makes to use the raw path here since that one is relative to the file and not to the project root.

crates/typst-pdf/src/embed.rs
```diff
@@ -0,0 +1,90 @@
...
+        .unic_file(TextStr(embed.path().as_str()))
+        .insert(Name(b"EF"))
+        .dict()
+        .pair(Name(b"F"), embedded_file_stream_ref)
+        .pair(Name(b"UF"), embedded_file_stream_ref)
+        .finish();
+    if let Some(relationship) = embed.relationship() {
+        if Some(PdfAConformanceLevel::A_3) == ctx.options.standards.pdfa {
+            let name = match relationship {
```
laurmaedje, line 70: This match could be moved into a helper method on EmbeddedFileRelationship.

laurmaedje (commented):
Here are some initial comments, for now mostly from a code perspective rather than a PDF perspective.

crates/typst-library/src/pdf/embed.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// The embedded file has an unknown relationship to the document or the relationship cannot be
+    /// described by the other variants
+    Unspecified,
+}
+
+/// A loaded file to be embedded.
+///
+/// Values of this type are cheap to clone and hash.
+#[derive(Clone, Hash, Eq, PartialEq)]
+pub struct Embed(Arc<LazyHash<Repr>>);
```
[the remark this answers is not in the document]
NiklasEi: `from_element` has a little bit of logic to fallback on using the file path as the name.

I tried using the Packed<EmbedElem>, but cannot access the optional fields because they are private.

crates/typst-pdf/src/lib.rs
NiklasEi: I removed the enum, but in my opinion using two booleans makes the code harder to understand in some places. If we need to differentiate between 2B and 3B, we now use the fact that we disallow file embeddings for 2B.

For example: when we write the `pdfa_part` in the catalog, we now use the `embedded_files` boolean to differentiate between 2B and 3B.

crates/typst-pdf/src/embed.rs
NiklasEi: It is at least the latest time that it could have been modified at. Alternatively, we could add another optional parameter, but I though this is a good solution for the beginning.

crates/typst-pdf/src/embed.rs
NiklasEi: That is a good point. Should we always take the last path segment instead? If users what exact control over it, they can use the `decode` variant.

NiklasEi:
> Here are some initial comments, for now mostly from a code perspective rather than a PDF perspective.

Thank you! All comments are addressed via commit and/or response.

crates/typst-pdf/src/embed.rs
reknih: If we want to support this, I would suggest making the file creation and modification date available via the world as an optional parameter and then do `modification-date` with a default value of `auto` an additional named parameter. As with `document`'s `date`, this can also be `none` or an explicit date. If a path is passed, the modification date from the world, if any, should be used, if it's a bytes / string value, it should be the world's `today`. This way, we provide the actual expected value for files present on the disk, allow opting out and overriding, and provide a way for world implementors to opt out of this feature if their FS implementation does not track modification dates.

NiklasEi: > I would suggest making the file creation and modification date available via the world as an optional parameter

I am unsure how you imagined the "as an optional parameter" part. I tested adding another method to the `World` trait to get the optional modification date for a file by `FileId` (d4cf66c93686ab5dc775803daa47c005b9f685f9). Did you have something else in mind?

NiklasEi: @reknih The extension of the world trait is pushed and it seems to work fine with the CLI world. However, I encountered issues adding the optional smart parameter to the element. As far as I understood, I should handle the smart parameter in `parse`. The following is my current attempt at that:

```rust
    #[parse(
        let last_modified = if let Some(Spanned { v: smart, span: _ }) =
            args.named::<Spanned<Smart<Option<Datetime>>>>("last-modified")?
        {
            Some(Smart::Custom(match smart {
                Smart::Auto => engine.world.last_modified(id).map_err(|file_error| {
                    eco_vec![error!(
                        path_span,
                        "failed to read metadata for file `{path}`: {file_error:?}";
                        hint: "todo",
                    )]
                })?,
                Smart::Custom(date) => date,
            }))
        } else {
            Some(Smart::Custom(engine.world.last_modified(id).map_err(|file_error| {
                eco_vec![error!(
                    path_span,
                    "failed to read metadata for file `{path}`: {file_error:?}";
                    hint: "todo",
                )]
            })?))
        };
        last_modified
    )]
    pub last_modified: Smart<Option<Datetime>>,
```

But this does not compile, because looking up the modification date through the world requires the `FileId`, which I don't have at both positions where this code gets inserted. These are the relevant parts of expanding the macros:

```rust
    impl ::typst_library::foundations::Construct for EmbedElem {
        fn construct(engine: &mut ::typst_library::engine::Engine, args: &mut ::typst_library::foundations::Args ) -> ::typst_library::diag::SourceResult<::typst_library::foundations::Content> {
            let Spanned { v: path, span: path_span } =
                args.expect::<Spanned<EcoString>>("path to the file to be embedded")?;
            ;
            let id = path_span.resolve_path(&path).at(path_span)?;
            ;
            let data = engine.world.file(id).at(path_span)?;
            let path = path
                ;
            let data = Readable::Bytes(data);
            let name = args.named("name")?;
            let description = args.named("description")?;
            let mime_type = args.named("mime-type")?;
            let relationship = args.named("relationship")?;
            let last_modified = if let Some(Spanned { v: smart, span: _ }) =
                args.named::<Spanned<Smart<Option<Datetime>>>>("last-modified")?
            {
                Some(Smart::Custom(match smart {
                    Smart::Auto => engine.world.last_modified(id).map_err(|file_error| {
                        eco_vec![error!(
                        path_span,
                        "failed to read metadata for file `{path}`: {file_error:?}";
                        hint: "todo",
                    )]
                    })?,
                    Smart::Custom(date) => date,
                }))
            } else {
                Some(Smart::Custom(engine.world.last_modified(id).map_err(|file_error| {
                    eco_vec![error!(
                    path_span,
                    "failed to read metadata for file `{path}`: {file_error:?}";
                    hint: "todo",
                )]
                })?))
            };
            let last_modified = last_modified
                ;
            Ok(::typst_library::foundations::Content::new(Self { path, data, name, description, mime_type, relationship, last_modified }))
        }
    }


    impl ::typst_library::foundations::Set for EmbedElem {
        fn set(engine: &mut ::typst_library::engine::Engine, args: &mut ::typst_library::foundations::Args ) -> ::typst_library::diag::SourceResult<::typst_library::foundations::Styles> {
            let mut styles = ::typst_library::foundations::Styles::new();
            if let Some(value) = args.named("name")? { styles.set(Self::set_name(value)); }
            if let Some(value) = args.named("description")? { styles.set(Self::set_description(value)); }
            if let Some(value) = args.named("mime-type")? { styles.set(Self::set_mime_type(value)); }
            if let Some(value) = args.named("relationship")? { styles.set(Self::set_relationship(value)); }
            let last_modified = if let Some(Spanned { v: smart, span: _ }) =
                args.named::<Spanned<Smart<Option<Datetime>>>>("last-modified")?
            {
                Some(Smart::Custom(match smart {
                    Smart::Auto => engine.world.last_modified(id).map_err(|file_error| {
                        eco_vec![error!(
                        path_span,
                        "failed to read metadata for file `{path}`: {file_error:?}";
                        hint: "todo",
                    )]
                    })?,
                    Smart::Custom(date) => date,
                }))
            } else {
                Some(Smart::Custom(engine.world.last_modified(id).map_err(|file_error| {
                    eco_vec![error!(
                    path_span,
                    "failed to read metadata for file `{path}`: {file_error:?}";
                    hint: "todo",
                )]
                })?))
            };
            if let Some(value) = last_modified
            { styles.set(Self::set_last_modified(value)); }
            Ok(styles)
        }
    }
```

The `Construct` implementation seems fine because I have access to the path parameter, but `Set` does not compile because there is no path or file ID.
My understanding of Typst's macro machinery is quite limited. Is there any extended documentation on this that I missed? Or, if not, could you explain it to me?

laurmaedje: To be honest, I'm not in favor of this change. The value of having the real file modification in the PDF seems quite small to me, while it hurts reproducibility and increases the complexity of the `World` interface. Writing the document's date itself was already a point of contention with some users that put a higher priority on reproducibility. This goes way further. In particular, the modification date might get lost through copying, downloading, sharing, uploading, etc. I'd like to add as little data into the mix that isn't obviously part of the inputs from a user PoV.

In principle, I understand the desire to follow the standard as much as possible. But I feel like these standards were developed primarily through the desktop-publishing lens where files are edited in a visual editor and exported rather than from the lens of a command line processor that wants to generate reproducible output. There a little bit of metadata might not hurt, but I feel like our case is different.

crates/typst-pdf/src/embed.rs
laurmaedje: Let's use the path relative to the project root. So essentially `id.as_rooted_path()`.

crates/typst-library/src/pdf/embed.rs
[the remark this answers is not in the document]
laurmaedje: You should be able to access the private fields via accessor function, like mentinoed above: `element.description(StyleChain::default())`.

crates/typst-pdf/src/lib.rs
laurmaedje: For the pdfa_part in the catalog, I wanted to have the `(u8, &'static str)` as `ctx.options.standards.part`. Essentially everything that is semantically separate becomes its own thing in `PdfStandards`.

tests/suite/pdf/embed.typ
```diff
@@ -0,0 +1,19 @@
...
+
+--- embed-document ---
+#pdf.embed("/assets/text/hello.txt", name: "blub.foo", description: "A test file", mime-type: "text/plain", relationship: "supplement")
+
+--- embed-invalid-relationship ---
+// Error: 123-129 expected "source", "data", "alternative", "supplement", "encrypted-payload", "form-data", "schema", "unspecified", or none
+#pdf.embed("/assets/text/hello.txt", name: "blub.foo", description: "A test file", mime-type: "text/plain", relationship: "test")
+
+--- embed-raw-basic-document ---
+#let raw_file = read("/assets/text/hello.txt")
```
laurmaedje, line 14: please use kebab-case :)

crates/typst-pdf/src/embed.rs
NiklasEi: I reverted the World extension and use the document date if there is one. I think the first version of this should add some useful date to have valid embeddings and allow the creation of valid ZUGFeRD documents.

If there is consent on how to add a more accurate last modified date, we could do it in a follow-up, in my opinion.

NiklasEi:
All comments are addressed; ready for review.

laurmaedje:
@NiklasEi I refactored this quite a bit together with @reknih, in part to make things more idiomatic, but also fixing some issues. Since dealing with the PDF specs is always quite involved, directly making the changes was a bit less time-consuming for me than leaving comments and going back and forth over it. I tried to make all commits small and self-contained, so feel free to take a look at the changes and let me know if anything seems odd. I want to make one more pass over the code to check things, but other than that this should be pretty much ready to merge.

crates/typst-pdf/src/lib.rs
```diff
@@ -147,16 +151,34 @@ pub enum Timezone {
...
 #[derive(Clone)]
 pub struct PdfStandards {
-    /// For now, we simplify to just PDF/A, since we only support PDF/A-2b. But
-    /// it can be more fine-grained in the future.
+    /// For now, we simplify to just PDF/A. But it can be more fine-grained in
+    /// the future.
     pub(crate) pdfa: bool,
+    /// Whether the standard allows for embedding any kind of file into the PDF.
+    /// We disallow this for PDF/A-2, since it only allows embedding other
+    /// PDF/A-2 documents.
```
NiklasEi, line 159: A comment in `write_embedded_files` mentions that an embedding in PDF/A-2 might be PDF/A-2 or PDF/A-1. Here we say it needs to be A-2.
Do I understand the spec correctly that any PDF/A file would be allowed as embedding in a PDF/A-2 file?

NiklasEi (commented):
@laurmaedje Thank you for taking the time!
I like the changes but have a small question.

crates/typst-pdf/src/lib.rs
laurmaedje: I didn't check myself, but @reknih told me that PDF/A-1 or PDF/A-2 is fine. I didn't see this comment here, so it should be updated.

laurmaedje:
Thanks for your work on this!

TS53:
Are the following things currently possible?
1. use a different internal filename without renaming the file
2. change the file date

What happens with the file date when it is included from bytes?

laurmaedje:
The whole file will be read into memory either way.

TS53:
In typst it seems reading contents fully into RAM is common. But we can think of a solution where it gets chunked in fixed size blocks if I call something like `embed("input_filename.ext", "internal_filename.ext")`.

laurmaedje:
Yes, it's common because Typst abstracts over the file system in a relatively high-level way. Chunking it all the way from reading to embedding (and compressing!) in the PDF would hugely increase the implementation complexity and there are no plans to do that.

TS53:
I understand the design decision. Btw: is compression optional? Many file formats are already compressed, additional compression can result in higher filesize and longer compilation time.

laurmaedje:
Currently, it will always compress.

mrfragger:
Ultimately did this version for 0.13 of PDF embed enable using the same name as the pdf for the attachment.typ?

Important Document.pdf
embeds
Important Document.typ

laurmaedje:
@mrfragger There is currently no way to get the input file name so you need to either spell it out  in your source file or provide it via `sys.inputs`.

mrfragger:
quite impressed
one.typ 5.1MB
one.pdf 8.6MB

one_qpdf_embed_typ.pdf 10MB
one_pdf.embed_typst.pdf 10.1MB

Seems like compression efficiency improved a tad for final  typst 0.13 compared to pre-release.

so attachment size is 1.4MB for qpdf and 1.5MB for typst native pdf.embed.  No longer need qpdf as a dependency for embedded source typst files in pdf.  Thanks for the hard work put into this.

PR #5237, opened 2024-10-18
Split changelog

laurmaedje:
Splits the changelog into multiple pages and turns the top level list items previously used for thematic grouping into headings. This gives things a bit more room to breathe and makes the changelog more readable.


Enivex:
Pippin : What about changelog?

Aragorn : You've already had it.

Pippin : We've had one, yes. What about second changelog?

Merry : I don't think he knows about second changelog, Pip.

PR #5241, opened 2024-10-18
Add a `rev` named argument to the `find` and `position` of `array` and `position` of `str`

### LINKED ISSUE #4981: Allow searching arrays and strings from the end
### Description

The `array` and `string` types both have `find()` and `position()` methods that find a first occurrence of some search pattern. There is no equivalent `find-last()` etc., and emulating it requires either a loop or using `rev()` (won't work with regexes on strings), followed by index calculations in the case of `position()`.

My suggestion is to support that
- either through the existing methods with a parameter like `last: false`, `reverse: false`, `match: "first"`, `direction: ltr` (not a fan as `ltr` is probably not semantically a good choice),
- or with new methods called `find-last()` and `last-position()` or similar. 

### Use Case

Finding the last match, or position of a match, of something can often be useful, e.g. when limiting what data to show out of a larger dataset.

noamzaks:
Partially fixes #4981.

Currently lacks support for reverse regex search; Could be implemented with `regex_automata` supposedly.

I can try to implement reverse regex search using `regex_automata` if we want to go that route.

### DIFF
--- a/crates/typst/src/foundations/array.rs
+++ b/crates/typst/src/foundations/array.rs
@@ -5,9 +5,9 @@ use std::ops::{Add, AddAssign};
 
 use comemo::Tracked;
 use ecow::{eco_format, EcoString, EcoVec};
-use rayon::iter::Either;
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
+use typst_utils::MaybeReverseIter;
 
 use crate::diag::{bail, At, HintedStrResult, SourceDiagnostic, SourceResult, StrResult};
 use crate::engine::Engine;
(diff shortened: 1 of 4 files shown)


laurmaedje:
I think going down the regex-automata route is a bit overkill. Notably, `trim` already does some reverse-regex stuff, though inefficiently and maybe with slightly different semantics than one might think (not sure).

We also still need to settle on naming. `from: start | end` flows quite well, though I'm still a bit concerned about `.position(from: end)` sounding like it counts from the back.

noamzaks:
I see. I don't have a particular opinion; How should we go through with this?

laurmaedje:
> How should we go through with this?

Unfortunately, I don't know either.

noamzaks:
Closing until it's clear how we want to do this

laurmaedje:
I [asked on Discord](https://discord.com/channels/1054443721975922748/1275038033975967765/1301844820301578250). One person so far was in favor of `rev`, which is a reasonable stance. It is slightly less nice than `from: start | end` but does not suffer from its problems. Code-wise the PR looked very good, so once we've decided, we can just reopen.

PR #5246, opened 2024-10-19
Bump version on main

laurmaedje:
The tagged release commit is on the 0.12 branch, but we also gotta bump the version on main.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2666,7 +2666,7 @@ checksum = "6af6ae20167a9ece4bcb41af5b80f8a1f1df981f6391189ce00fd257af04126a"
 
 [[package]]
 name = "typst"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "arrayvec",
  "az",
@@ -2734,12 +2734,12 @@ dependencies = [
 
 [[package]]
 name = "typst-assets"
-version = "0.11.0"
-source = "git+https://github.com/typst/typst-assets?rev=4afd428#4afd428838a8fc99151c00dcf52f26b94c367aa0"
+version = "0.12.0"
+source = "git+https://github.com/typst/typst-assets?rev=5c0dcc0#5c0dcc062eddd9c1b1a8994fe4948ab604a1690f"
 
 [[package]]
 name = "typst-cli"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "chrono",
  "clap",
@@ -2783,12 +2783,12 @@ dependencies = [
 
 [[package]]
 name = "typst-dev-assets"
-version = "0.11.0"
-source = "git+https://github.com/typst/typst-dev-assets?rev=e9f8127#e9f81271547c0d7003770b4fa1e59343e51f7ae8"
+version = "0.12.0"
+source = "git+https://github.com/typst/typst-dev-assets?rev=b07d156#b07d1560143d6883887358d30edb25cb12fcf5b9"
 
 [[package]]
 name = "typst-docs"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "clap",
  "comemo",
@@ -2811,7 +2811,7 @@ dependencies = [
 
 [[package]]
 name = "typst-fuzz"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "comemo",
  "libfuzzer-sys",
@@ -2823,7 +2823,7 @@ dependencies = [
 
 [[package]]
 name = "typst-ide"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "comemo",
  "ecow",
@@ -2839,7 +2839,7 @@ dependencies = [
 
 [[package]]
 name = "typst-kit"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "dirs",
  "ecow",
@@ -2859,7 +2859,7 @@ dependencies = [
 
 [[package]]
 name = "typst-macros"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "heck",
  "proc-macro2",
@@ -2869,7 +2869,7 @@ dependencies = [
 
 [[package]]
 name = "typst-pdf"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "arrayvec",
  "base64",
@@ -2895,7 +2895,7 @@ dependencies = [
 
 [[package]]
 name = "typst-render"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "bytemuck",
  "comemo",
@@ -2913,7 +2913,7 @@ dependencies = [
 
 [[package]]
 name = "typst-svg"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "base64",
  "comemo",
@@ -2929,7 +2929,7 @@ dependencies = [
 
 [[package]]
 name = "typst-syntax"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "ecow",
  "once_cell",
@@ -2945,7 +2945,7 @@ dependencies = [
 
 [[package]]
 name = "typst-tests"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "clap",
  "comemo",
@@ -2969,7 +2969,7 @@ dependencies = [
 
 [[package]]
 name = "typst-timing"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "parking_lot",
  "serde",
@@ -2979,7 +2979,7 @@ dependencies = [
 
 [[package]]
 name = "typst-utils"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "once_cell",
  "portable-atomic",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ default-members = ["crates/typst-cli"]
 resolver = "2"
 
 [workspace.package]
-version = "0.11.0"
+version = "0.12.0"
 rust-version = "1.77" # also change in ci.yml
 authors = ["The Typst Project Developers"]
 edition = "2021"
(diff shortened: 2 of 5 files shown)


alerque:
Tagging on a release branch is fine, but the tag should *also* be reachable in the history of *main*, otherwise `git describe` can't determine the version relative to the last release. This is an issue for HEAD packaging (e.g. as Arch Linux, Homebrew, and others do).

The best way to handle this is to have at least one merge *from* the release branch that has the tag *into* the default branch so the tag and any other release finalization end up in the history there, then move forward from there. The release branch can be kept and continued after merging for the event of patch releases, etc. and even merged again at a later date to bring patch release tags into the history.

laurmaedje:
If I merge 0.12 into main, I get many duplicate commits though, which I've previously cherry-picked from main to 0.12.

alerque:
Git can cope: all you need is the tag history. You won't get duplicate commit *content) because it understands what has been cherry picked and won't try to make the same changes twice. The only real conflict is in the manifest, and you could easily resolve that. But if it bothers you and you really want to be sure you're getting only the current state of main, so you can start the merge with `git merge v0.12.0` (with *main* as your checked out working tree), then when it show you all the conflicts, use `git checkout main -- $(git ls-files)` to reset all the files to their state on main, then `git merge --continue`. This way you'll get the commit topology in history with the tag reachable from *main* without getting any content changes.

laurmaedje:
What I primarily want to avoid is the commits showing up twice in the history on GitHub. Maybe it's me being too attached to the GitHub way of viewing a repository, but it would nonetheless be unfortunate. At least in GitHub Desktop, the commits showed up twice after I tested merging.

PgBiel:
In this particular situation, a shared tag doesn't work flawlessly because of the release candidates being parent commits, and changing the parent commit would change the SHA, so here we can't have a 0.12 tag on both the main and 0.12 branches, requiring a merge which, as shown, may result in visual commit duplication weirdness.

The only way I can see to not require a merge for that would be to dissociate the release candidates from the main and 0.12 branches entirely by rebasing the "Version bump" commit on main, such that it is shared by both branches and can be tagged. This would, however, waive the benefit of being able to push stuff to main right before a release. Additionally, since 0.12 was already released, I'm not positive this is feasible anymore, leaving the merge as the only remaining option to _somehow_ indicate the tag's existence in main's history.

Still, I can see how having this work would be valuable. I think a way to soften the merge's visual problems would be to squash cherrypicked commits into one, adding a list of squashed commits in the commit description. It will look a bit ugly on main but it will be just one commit (alongside the version bumps), so maybe it's not that bad.

alerque:
The view GitHub shows is a little unfortunate, but keep in mind it is showing you topology not content. The description of the current main branch is (in my opinion) much worse than the duplicated commits in topology. This gotcha will last much longer than the extra commits in your scroll distance.

```console
$ git checkout main
$ git describe --always --long --tags
v0.11.0-452-g838dabc8
```

This should be `v0.12.0-0-g737895d7` (or whatever commit count & hash one happens to be on) relative to 0.12 not 0.11.

There are ways of avoiding both issues in the future (not cherry picking things that *will be* merged anyway, using a long running release branch and a devel branch to keep things out of the way around release cycles, etc., rewriting history of *main*) but none of them are really good candidates for fixing *this snafu* and for now I still suggest fixing the topology so the tag shows up in the history will be more relevant in the long run than the fuss over GitHub's display (which doesn't reveal what content did or didn't get merged, only that the history was joined.

alerque:
> I think a way to soften the merge's visual problems would be to squash cherrypicked commits into one,

No you can't do this, it would defeat the whole point and loose its relation to the commit that officially released with the tag and the tag would still end up out of the history of the main branch.

PgBiel:
> > I think a way to soften the merge's visual problems would be to squash cherrypicked commits into one,
> 
> 
> 
> No you can't do this, it would defeat the whole point and loose its relation to the commit that officially released with the tag and the tag would still end up out of the history of the main branch.

Yeah, I realized that this wouldn't be feasible _for 0.12_ given the tag already exists. My idea might be handy for 0.13+ when that comes.

Maybe biting the bullet now isn't too bad. But we have to be certain that this is what we want (and/or that we are doing it in the right way), considering that the same behavior would have to be replicated for patch releases.

laurmaedje:
I'm curious how this relates to what rust-lang/rust is doing. When I run `git describe --always --long --tags` there, I get `1.0.0-beta-228151-gb27f33a4d9c`.

PgBiel:
I'd like to offer a different perspective: instead of using `git describe`, downstream could use `git tag --list --sort=-creatordate` instead to get all tags from all branches:

```sh
git tag --list --sort=-creatordate -n

v0.12.0         Version bump
v0.12.0-rc2     Release Candidate 2
v0.12.0-rc1     Release Candidate
v0.11.1         Version bump
v0.11.0         Version bump
v0.11.0-rc1     Release Candidate
v0.10.0         Version bump
v0.9.0          Version bump
v0.8.0          Version bump
v0.7.0          Version bump
v0.6.0          Version bump
v0.5.0          Version bump
v0.4.0          Version bump
v0.3.0          Version bump
v0.2.0          Version bump
v0.1.0          Bump version
v23-03-28       New release
v23-03-21-2     Embed standard fonts in binary
v23-03-21       Documentation changes
```

To include the commit hashes, you may add a `--format`, as such[^1]:
```
git tag --sort=-creatordate --format="%(refname:short) %(if)%(object)%(then)%(object)%(else)%(objectname)%(end)" 

v0.12.0 737895d769188f6fc154523e67a9102bc24c872e
v0.12.0-rc2 ef309ca19617a3f1b8b78efb9644b67460fa5203
v0.12.0-rc1 d5b1f4a6f9b16166b4affe8c49d3f433a4b425c8
v0.11.1 5011510270c2c23f0ab019af486b26db0d62261b
v0.11.0 2bf9f95dd83f2ceef1e32ae1847a162c7d193f76
v0.11.0-rc1 fe94bd85aec9348c4f86a7b20befb6562655c93f
v0.10.0 70ca0d257bb4ba927f63260e20443f244e0bb58c
v0.9.0 7bb4f6df44086b4c1120b227f7ae963e6c2ad5ab
v0.8.0 360cc9b9570f263d52530b98d0c93523e7bdb100
v0.7.0 da8367e189b02918a8fe1a98fd3059fd11a82cd9
v0.6.0 2dfd44fedd99ab9414c17f358179e1c37e953f30
v0.5.0 3a8b9cce471934031ba57c5972d194ac1d7d2a6c
v0.4.0 f692a5efc62bf95f57d123895b7dce55d2eb9741
v0.3.0 b1e0de00784061a7670072160683f56c8269b25c
v0.2.0 fe2640c55268f167d8749f77b37e52b7b17f21dd
v0.1.0 b3faef4b80a674294091066e20501e3a5d0f6103
v23-03-28 056d15aa49f545980b03068684f0d44e35248cd6
v23-03-21-2 b934a2fd83d63fc115c01f959e888c7bc1aa87e4
v23-03-21 4428646c9997171e3aea25ad2f0c151755f072da
```

So, maybe we're doing it just fine if the tags can still be listed.

[^1]: Source: https://stackoverflow.com/questions/1862423/how-to-tell-which-commit-a-tag-points-to-in-git (simplified from https://stackoverflow.com/a/68687780)

laurmaedje:
If I understood correctly, part of the problem with the status quo is that the tags lose their relation. Comparing v0.11.0 with v0.12.0 works. But comparing v0.12.0 with HEAD or v0.13.0 doesn't work anymore.

alerque:
@PgBiel "could use" for what? The stuff you propose doesn't solve the problem I'm trying to address. I know how to find the list of tags (sorted by date or otherwise) or the hashes that they reference. That isn't the problem.

The problem is for packaging systems that allow uses to build packages from Git HEAD (whatever that is at the time they build). It is quite helpful  to come up with a version number to tell the system package manager about that can be compared with stable release packages. So lets say Arch Linux (or Homebrew, etc.) has a stable release package for `typst-0.11.0`, and a user comes along and builds `typst-git` themselves at a moment in time where 0d93ccd4 happens to be the current tip of the *main* branch. They would get a package version of something like `0.11.0-r185` which easily identifies the HEAD package as a newer version than (185 commits ahead of) the stable package. At the moment if you do that and build `typst-git` on Arch Linux you'll get something like `0.11.0-r452` which identifies it as part of the v0.11 era of development and sorts as *older* than the stable `0.12.0` package in the repositories.

This isn't a problem all upstream projects are required to solve, but it sure is nice and convenient when the do. And to date Typst has, so this is a workflow change that makes for a regression in downstream packaging options. If the only motivation in not fixing it is not wanting to have a few cherry picked commits showing up twice in topology then it seems to me like the downsides don't outweigh the upsides. If you're really going to pursue maintaining things in release branches that cannot/should not also show up in the topology of the default/development branch then that's one thing, but I see no indication that any workflow along those lines is planned or intended, you just got ahead of yourself a little bit cherry picking the wrong direction during an RC release cycle—which should be easily fixed by the proposed merge.

alerque:
> But comparing v0.12.0 with HEAD or v0.13.0 doesn't work anymore.

Yes, that's another use case for wanting the release tag in the development branch history besides the distro packaging one I outlined. To some extent Git will compensate for that by actually backtracing to the branch point to show you diffs, but the result will be over more confusing diffs than your concern about a handful of commits showing up in two branches of the topology.

alerque:
> I have a feeling that knowing the exact amount of commits since the latest release is rather niche... for the usual case, assuming no rebases and other weirdness, a simple date comparison will do;

No it won't. Because we don't have date information available.

You're still trying to change tho problem space to fit a specific solution rather than trying to adapt the tooling and scenario to be a solution to a specific problem.

Almost all package managers have tooling to compare version numbers, whether a stand alone tool or a builtin function or whatever. This is used to compare packages relative to each-other. Few people are going to care about "knowing the exact amount of commits since the last release". As you say that info is niche, but they *do care* about "is A or B newer". That is not so niche. Is v0.11 or v0.12 newer? You can tell because 12 is a bigger number than 11. You might not care the exact number of minor release bumps Typst has made but you care that there is a newer one out there than the one you have, i.e. it is time to upgrade your system.

Arch Linux's tool for this decision making is called [`vercmp`](https://man.archlinux.org/man/vercmp.8), but all distros have something similar. They all fundamentally decide based on a version string whether A<B or A==B or B>A. The issue here is being able to provide that information in a package such that the relevant comparison function(s) work. Hence why Arch's method of choice with HEAD builds is to make sure and provide an incrementing field "commits since release X".

The commit date doesn't help here because the stable release package `typst-0.11.0` doesn't provide any date semantics. We have to fill in the function `vercmp 0.11.0 $version` where $version has some semantic relation to the other side of the comparison. This is why the current `0.11.0-r452` is awkward because semantically it sorts before `0.12.0`, not after it. What I'm asking for is a Git topology on the main development branch such that it can be easily be derived using standard tooling.

This can be faked downstream on the packaging side several ways, but I don't understand the resistance to just fixing the topology upstream. It will only improve every other usage including `git log` and `git diff` output as development progresses.

> and, if we do release a patch release for a version older than the current one (say, with a security fix), it wouldn't make much sense to merge an older branch into main anyway (and that's a rather contrived / unlikely case).

Okay. And? For the workflow I'm suggesting and all the others that have come up it would be irrelevant. If 0.13.0 has been tagged and later a hotfix to 0.12.1 comes out then new builds from the HEAD branch *SHOULD BE* relative to 0.13.0 not the hotfix, so merging backported hotfix tags into the current development branch probably does not make sense.

On the other hand up until the point that a newer release comes out, any patch releases with hotfixes *could* be merged in just fine with no detriment to anybody.

laurmaedje:
> I don't understand the resistance to just fixing the topology upstream

I want to understand all the options and trade-offs before taking irreversible action. That's all.

> just got ahead of yourself a little bit cherry picking the wrong direction during an RC release cycle

What I still don't fully get is what the intended workflow for cherry-picking _is._ Features and bug fixes typically happen on main, so for example for 0.11.1, I cherry-picked all bug fixes from main to 0.11. For 0.12.0, doing it the same way (even though it's not a patch release) seemed natural. 

> The issue here is being able to provide that information in a package such that the relevant comparison function(s) work.

I think the Cargo.toml field (on HEAD) would provide that info? But I get that this might stride too far away from the standard workflows.

As for what to do now: In principle, I'm open to doing the merge, I just want to be clear on how to proceed in the future. Let's say we do end up publishing 0.12.1. Then, we would want the version field on main to also indicate that (to prevent the 0.11 problem where main was rejecting packages with `compiler = "0.11.1"`). This would mean doing the merge for 0.12.1 as well. Since I don't want to keep accumulating duplicate commits, the questions is how to best handle it going forward.

For what it's worth: Another (maybe worse) way to get the tag into main without duplicate commits would be to force reset main to 0.12. There wasn't anything on main that was _not_ cherry-picked in this release cycle. Though the usual badness of force pushing (especially with a few days delay) of course applies.

laurmaedje:
This post describes pretty much our current setup: https://trunkbaseddevelopment.com/branch-for-release/

- Fix all things on main
- Cherry pick to release
- Do not merge release back

It's not the same as GitFlow obviously but apparently similar to how Google does it ...

emilyyyylime:
I'm not sure what @alerque's opinion on the matter is (they are the maintainer for the `extra/typst` and more notably `aur/typst-git` packages on Arch), but if Typst is committed to using a different workflow for releases going forward it's certainly possible for downstream to take an approach similar to that taken in the [`rust-git`](https://aur.archlinux.org/packages/rust-git) package, simply counting commits (with `git rev-list --count HEAD`). Certainly to aid readability by humans and maintain correct ordering with respect to package versions prior to this change, the package could incorporate a version parsed from `Cargo.toml` (leading to a `pkgver()` implementation more or less like `printf "%s.r%s.%s" "$(sed -n 's/^version = "\(.*\)"$/\1/' Cargo.toml)" "$(git rev-list --count HEAD)" "$(git rev-parse --short=7 HEAD)"`).

I hope this change didn't discourage @alerque from contributing to either the Typst or Arch Linux ecosystems.  
Perhaps a lesson to be learnt here is that such structural changes to the way releases are handled should be properly discussed and announced to the community before being implemented.

PR #5250, opened 2024-10-19
Improved docs for `par.line.numbering-scope`

Andrew15-5:
If the numbering can't be reset at all, then the "automatically" part should be dropped to not confuse people into thinking that it can be reset manually.

I actually haven't tried it. Can you reset the numbering with `counter("par.line").update(0)` or something?

Also the "that is, is never" sounds strange to me. Another change could be replacing the "that is" with "i.e.".

### DIFF
--- a/crates/typst/src/model/par.rs
+++ b/crates/typst/src/model/par.rs
@@ -381,7 +381,7 @@ impl Construct for ParLine {
 /// Possible line numbering scope options, indicating how often the line number
 /// counter should be reset.
 ///
-/// Note that currently the manual reset of the line number counter is not
+/// Note that, currently, manually resetting the line number counter is not
 /// supported.
 #[derive(Debug, Cast, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum LineNumberingScope {
--- a/crates/typst/src/model/par.rs
+++ b/crates/typst/src/model/par.rs
@@ -380,13 +380,16 @@ impl Construct for ParLine {
 
 /// Possible line numbering scope options, indicating how often the line number
 /// counter should be reset.
+///
+/// Note that currently the manual reset of the line number counter is not
+/// supported.
 #[derive(Debug, Cast, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum LineNumberingScope {
-    /// Indicates the line number counter spans the whole document, that is,
-    /// is never automatically reset.
+    /// Indicates that the line number counter spans the whole document, i.e.,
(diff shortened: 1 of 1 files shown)


PgBiel:
I'm not sure I understand. By resetting on each page, it is doing so automatically, that is, without your intervention.

It's true that you can't do it manually at the moment - we did try implementing that, but it turned out to be a much deeper rabbit hole than it seemed, because the line number counter is stepped outside of the document flow, so the document flow isn't aware of the current value of the counter unless you do an explicit pagebreak, making this very tricky. Still, it'd be an interesting goal for the future.

Either way, I don't mind changing the wording to that.

Andrew15-5:
Based on the additionally provided context, I changed the wording to how I think it would be easier to read/understand the description/docs. You can further modify it, if needed.

crates/typst/src/model/par.rs
```diff
@@ -382,11 +382,12 @@ impl Construct for ParLine {
 /// counter should be reset.
 #[derive(Debug, Cast, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum LineNumberingScope {
-    /// Indicates the line number counter spans the whole document, that is,
-    /// is never automatically reset.
+    /// Indicates that the line number counter spans the whole document, i.e.,
+    /// it's never automatically reset. Currently, the manual reset of
+    /// the counter is not supported.
```
PgBiel, line 387: I feel like the last sentence should go on the type's own documentation.

Andrew15-5: Fixed?

crates/typst/src/model/par.rs
```diff
@@ -380,13 +380,16 @@ impl Construct for ParLine {
 
 /// Possible line numbering scope options, indicating how often the line number
 /// counter should be reset.
+///
+/// Note that currently the manual reset of the line number counter is not
+/// supported.
```
PgBiel, line 385: Perhaps a bit clearer:

```suggestion
///
/// Note that, currently, manually resetting the line number counter at
/// arbitrary points of your document is not supported.
```

Andrew15-5: I think the "at arbitrary points of your document" is an unnecessary condition. It seems like "at non-arbitrary points" it _is_ supported (even though I don't know which points would that be). But it's not supported at all. So it can be misleading in this sense.

PgBiel: okay

PgBiel (approved):
Thank you

PR #5261, opened 2024-10-20
Fix wrong Greek and Ukrainian language codes

### LINKED ISSUE #5254: Greek translations use wrong language code
### Description

Should be `el` and not `gr`.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

xkevio:
Fixes #5254.


mkorje:
While you're at it, could you maybe also fix the language code for Ukrainian (from `ua` to `uk`). It was first mentioned in #4863 and I also added it to #4729, but this fix should be done regardless in case that PR isn't merged.

xkevio:
> While you're at it, could you maybe also fix the language code for Ukrainian (from `ua` to `uk`). It was first mentioned in #4863 and I also added it to #4729, but this fix should be done regardless in case that PR isn't merged.

Good idea. cc @laurmaedje any issue with combining two fixes in one PR?

laurmaedje:
It's fine

laurmaedje:
Thanks!

PR #5288, opened 2024-10-24
Guard against font bug in math glyph assembly code

### LINKED ISSUE #5283: Stretching of `dash()` (U+305) broken with Noto Sans Math
### Description

The following

```typ
#show math.equation: set text(font: "Noto Sans Math", fallback: false)

$dash(x+y)$
```

renders as 

image

while with the same font used with unicode-math and luatex

```tex
\[
    \wideoverbar{x+y}
\]
```
renders as

image

The issue seems specific to this one accent and font

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

mkorje:
~~Fixes #5283 and fixes https://forum.typst.app/t/broken-square-root-symbol-with-concrete-math-font/1296. Both are due to bugs in the glyph assembly code.~~

~~I've opened https://github.com/typst/typst-dev-assets/pull/4 to add two math fonts for the tests.~~

~~In fixing #5283, the behaviour is now similar to that of luatex (see snippet [here](https://github.com/TeX-Live/luatex/blob/f52b099f3e01d53dc03b315e1909245c3d5418d3/source/texk/web2c/luatexdir/tex/mlist.c#L1434-L1444)). I've mentioned in a comment that this may go against the spec, in some sense.~~

See the latest comment.

### DIFF
--- a/crates/typst-layout/src/math/fragment.rs
+++ b/crates/typst-layout/src/math/fragment.rs
@@ -412,7 +412,7 @@ impl GlyphFragment {
     /// Try to stretch a glyph to a desired height.
     pub fn stretch_vertical(
         self,
-        ctx: &MathContext,
+        ctx: &mut MathContext,
         height: Abs,
         short_fall: Abs,
     ) -> VariantFragment {
@@ -422,7 +422,7 @@ impl GlyphFragment {
     /// Try to stretch a glyph to a desired width.
     pub fn stretch_horizontal(
         self,
-        ctx: &MathContext,
+        ctx: &mut MathContext,
         width: Abs,
         short_fall: Abs,
     ) -> VariantFragment {
--- a/crates/typst-layout/src/math/stretch.rs
+++ b/crates/typst-layout/src/math/stretch.rs
@@ -193,6 +193,14 @@ fn assemble(
                     .end_connector_length
                     .min(next.start_connector_length)
                     .scaled(ctx, base.font_size);
+                if max_overlap < min_overlap {
+                    // This condition happening is indicative of a bug in the
+                    // font.
+                    panic!(
+                        "glyph {:?} has assembly parts with overlap less than minConnectorOverlap",
+                        base.c,
+                    );
+                }
 
                 advance -= max_overlap;
                 growable += max_overlap - min_overlap;
--- a/crates/typst-layout/src/math/stretch.rs
+++ b/crates/typst-layout/src/math/stretch.rs
@@ -1,6 +1,6 @@
 use ttf_parser::math::{GlyphAssembly, GlyphConstruction, GlyphPart};
 use ttf_parser::LazyArray16;
-use typst_library::diag::SourceResult;
+use typst_library::diag::{warning, SourceResult};
 use typst_library::foundations::{Packed, Smart, StyleChain};
 use typst_library::layout::{Abs, Axis, Frame, Length, Point, Rel, Size};
 use typst_library::math::StretchElem;
@@ -81,11 +81,45 @@ pub fn stretch_fragment(
     *fragment = MathFragment::Variant(variant);
 }
 
+/// Return whether the glyph is stretchable and if it is, along which axis it
+/// can be stretched.
+fn stretch_axis(ctx: &mut MathContext, base: &GlyphFragment) -> Option<Axis> {
+    let base_id = base.id;
+    let vertical = ctx
+        .table
+        .variants
+        .and_then(|variants| variants.vertical_constructions.get(base_id))
+        .map(|_| Axis::Y);
+    let horizontal = ctx
+        .table
+        .variants
+        .and_then(|variants| variants.horizontal_constructions.get(base_id))
+        .map(|_| Axis::X);
+
+    match (vertical, horizontal) {
+        (vertical, None) => vertical,
+        (None, horizontal) => horizontal,
+        _ => {
+            // As far as we know, there aren't any glyphs that have both
+            // vertical and horizontal constructions. So for the time being, we
+            // will assume that a glyph cannot have both.
+            ctx.engine.sink.warn(warning!(
+               base.span,
+               "glyph has both vertical and horizontal constructions";
+               hint: "this is probably a font bug";
+               hint: "please file an issue at https://github.com/typst/typst/issues"
+            ));
+
+            None
+        }
+    }
+}
+
 /// Try to stretch a glyph to a desired width or height.
 ///
 /// The resulting frame may not have the exact desired width.
 pub fn stretch_glyph(
-    ctx: &MathContext,
+    ctx: &mut MathContext,
     mut base: GlyphFragment,
     target: Abs,
     short_fall: Abs,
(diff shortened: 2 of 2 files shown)


crates/typst/src/math/stretch.rs
```diff
@@ -248,8 +248,22 @@ fn assemble(
...
+                // be considered incorrect behaviour according to the spec, but
+                // some fonts have glyphs where this is the case and they are
+                // handled just fine by other typesetting engines.
                 advance -= max_overlap;
-                growable += max_overlap - min_overlap;
+                // If max_overlap is less than min_overlap then we don't allow
+                // any growth.
+                growable += (max_overlap - min_overlap).max(Abs::zero());
+            } else {
+                // Final part's advance may not be the actual size of the part.
```
khaledhosny, line 260: That sounds like a font bug or intentional behavior. TeX engines notably don’t use part full advance at all and always use the glyph advance, so I don’t why you selectively do this for the last part only, either always use part full advance or always discard it.

khaledhosny (commented):
Can you explain what is the issue in the font, if it is a font bug it should be fixed (I can fix Noto Sans Math at least).

Cargo.toml
```diff
@@ -28,7 +28,7 @@ typst-syntax = { path = "crates/typst-syntax", version = "0.12.0" }
 typst-timing = { path = "crates/typst-timing", version = "0.12.0" }
 typst-utils = { path = "crates/typst-utils", version = "0.12.0" }
 typst-assets = { git = "https://github.com/typst/typst-assets", rev = "5c0dcc0" }
-typst-dev-assets = { git = "https://github.com/typst/typst-dev-assets", rev = "b07d156" }
+typst-dev-assets = { git = "https://github.com/mkorje/typst-dev-assets", rev = "071ff36" }
```
alerque, line 31: Pointing at a fork of assets is fine for dev work, but it shouldn't be in a PR marked as ready. I suggest at least marking the PR as draft status so it can't accidentally be merged with a snafu like this still in it. You can set draft status from the sidebar.

crates/typst/src/math/stretch.rs
mkorje: The problem was that the following code
```
#show math.equation: set text(font: "Concrete Math")
$ sqrt((sum_(j = 1)^d)^2) $
```
in Typst had the sqrt symbol and rule misaligned. It turns out that the bounding box for the sqrt symbol was incorrect: its height was slightly smaller than the assembled glyph. I couldn't reproduce this in LuaTeX though, and `lua-visual-debug` appeared to show the bounding box being correct.

Looking into the font, the part `/sqrt.tp` has full advance 605, but actual height 645. So we end up with the full height of the assembly being slightly shorter than it actually is, leading to the original problem. Again, I figured this is probably a bug because other engines are handling it correctly. I thought this could then only be a problem with the first and last parts in the assembly, so switching selectively for the last part seemed to make sense, matching LuaTex's behaviour whilst not changing the behaviour with other fonts.

Thanks for mentioning that LuaTeX and XeTeX don't use the part full advance at all (I couldn't tell when I looked at the code myself). Is there a specific reason the glyph advance is always used instead? (I don't quite get then why fullAdvance is even needed, when the glyph's advance would suffice...)

mkorje (commented):
I was definitely too hasty with this, my apologies. I've expanded on what the issues were and the reasoning for the changes in replies to the comments below.

crates/typst/src/math/stretch.rs
khaledhosny: > Looking into the font, the part `/sqrt.tp` has full advance 605, but actual height 645. So we end up with the full height of the assembly being slightly shorter than it actually is, leading to the original problem. Again, I figured this is probably a bug because other engines are handling it correctly. I thought this could then only be a problem with the first and last parts in the assembly, so switching selectively for the last part seemed to make sense, matching LuaTex's behaviour whilst not changing the behaviour with other fonts.

So that is indeed a font bug and the authors likely tested only in *TeX and didn’t catch it because part full advance is not used there. If it is up to me, I’d report it as a font bug and call it day. The font is likely broken in other implementations (e.g. Microsoft’s). Ignoring full advance only for the last part seems to be very arbitrary and inconsistent. Note that LuaTeX and XeTeX behavior here is peculiar as well, as they use glyph advance (at least in horizontal direction) while they should be using bounding box width, so glyphs with zero width (e.g. combining marks often get zero width automatically by font authoring tools) result in broken layout and one has to always make sure glyphs used in horizontal asssemblies have and advance width.

> Thanks for mentioning that LuaTeX and XeTeX don't use the part full advance at all (I couldn't tell when I looked at the code myself). Is there a specific reason the glyph advance is always used instead? (I don't quite get then why fullAdvance is even needed, when the glyph's advance would suffice...)

Most likely an optimization, as noted above the bounding box not the glyph advance should be used and calculating the bounding box is not cheap (especially for CFF fonts that don’t encode it at all in font data). It would allow MATH layout to be done completely without having to consult other font tables or data.

crates/typst/src/math/stretch.rs
```diff
@@ -248,8 +248,22 @@ fn assemble(
                     .min(next.start_connector_length)
                     .scaled(ctx, base.font_size);
 
+                // Using max_overlap even when it is less than min_overlap may
+                // be considered incorrect behaviour according to the spec, but
+                // some fonts have glyphs where this is the case and they are
+                // handled just fine by other typesetting engines.
```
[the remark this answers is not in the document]
khaledhosny: That is indeed a font bug and I’ll try to fix it. However, if end connector is 0, shouldn’t the 100 of the MinConnectorOverlap be used instead?

crates/typst/src/math/stretch.rs
Enivex: I don't know if there is a repository for Concrete Math, but the author is listed here https://ctan.org/pkg/concmath-otf?lang=en . Hopefully he's receptive to feedback (the font seems actively maintained).

crates/typst/src/math/stretch.rs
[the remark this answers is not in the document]
khaledhosny: Yes, please.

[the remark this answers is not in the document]
mkorje: I've opened https://github.com/notofonts/math/issues/94.

> However, if end connector is 0, shouldn’t the 100 of the MinConnectorOverlap be used instead?

This spec mentions this: "If the part does not have a connector on one of its ends, the corresponding connector length should be set to zero". So I guess if the end connector is 0, that end of the part should be considered "non-joinable".

What should happen when the maximum overlap is less than minConnectorOverlap (but nonzero) still remains.

crates/typst/src/math/stretch.rs
mkorje: Thank you for explaining that. I'll revert the changes here then and not ignore the full advance for only the last part. Given as well that calculating the bounding box is not cheap, I figure then it doesn't make much sense to switch to it over the full advance which is already in the GlyphPart record.

> I don't know if there is a repository for Concrete Math, but the author is listed here https://ctan.org/pkg/concmath-otf?lang=en . Hopefully he's receptive to feedback (the font seems actively maintained).

I've sent them an email. When I hear back I'll post an update.

mkorje:
I'm fairly certain that the current Typst implementation is correct. However, it works on the following assumption:
**that the maximum overlap of the connectors of neighboring parts is _not_ less than the minConnectorOverlap constant**.
My understanding and interpretation of the spec leads me to believe that the above assumption is perfectly fine. Quoting from the spec:

> The connectors are straight parts of the glyph that can be used to link it with the next or previous part. The connectors of neighboring parts can overlap by varying amounts, providing flexibility in how these glyphs can be put together. However, the overlap should not be less than the minConnectorOverlap value defined in the MathVariants tables, and it should not exceed the specified connector length for that end of the glyph. If the part does not have a connector on one of its ends, the corresponding connector length should be set to zero.

Thus, this assumption not holding is indicative of a bug in the font.

In the current Typst implementation though, we do not guard against the assumption not holding. This leads to the difference being the maximum overlap and minConnectorOverlap becoming negative. As this difference is used to determine by how much the overlap can decrease, we get the problem observed in #5283[^1].

Therefore, I think we should make our assumption explicit and throw an error if it does not hold. That way we can identify such a bug in the font and report it upstream. Whilst replacing the maximum overlap with the minConnectorOverlap in such a case where our assumption fails may work, it ultimately hides the underlying problem in the font.

[^1]: Despite the fact that `uni0305.s00`'s left side has connector length equal to its size, which would imply that we can overlap with the whole glyph, any implementation should only be considering the end of the current part and the start of the next part (or the start of the current and end of the last) when determining the maximum overlap. Considering the connector length for the other end of the glyph contradicts the spec, as the overlap "should not exceed the specified connector length for that end of the glyph".
There are gaps between the overlines in #5283 because the amount by which the overlaps can expand becomes negative, and with enough extender glyphs this becomes large enough to start going backwards. Typst's implementation is still following the steps of assembling glyph assemblies: set the glphys with maximum overlaps, then extend the glyphs further apart making the overlap smaller, stopping at the minConnectorOverlap; if this is still not big enough, add another extender and repeat.

laurmaedje:
How about turning both this and the other panic (in `stretch_axis`) into compiler warnings instead? It's a bit less drastic and easier for users to deal with.

laurmaedje:
Thanks!

PR #5294, opened 2024-10-25
Fix dockerfile not cross compiling correctly

### LINKED ISSUE #5290: v0.12.0 linux/arm64 Docker image shipped with x86-64 binary
### Description

Running the v0.12.0 Docker image on a linux/arm64 machine will crash with an error message `exec /bin/typst: exec format error`. This is due to the fact that the typst binary in the image is in fact in x86-64, as can be seen below:

```
$ uname -a
Linux ubuntu 5.15.0-87-generic #97-Ubuntu SMP Tue Oct 3 09:52:42 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

$ sudo docker run ghcr.io/typst/typst:v0.12.0@sha256:e142773807fb7d0be65908653eeaf0f2e78cae900f594bab41476a93ec9a377a
exec /bin/typst: exec format error

$ sudo docker cp $(sudo docker ps -a -q | head -n 1):/bin/typst .
Successfully copied 39.7MB to /home/ubuntu/.

$ file typst
typst: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), static-pie linked, BuildID[xxHash]=f228d78c86710a56, stripped
```

Running the same image on macOS/arm64 will **not** crash (probably) thanks to Rosetta, as can be seen below:

```
$ uname -a
Darwin A3113.local 24.0.0 Darwin Kernel Version 24.0.0: Tue Sep 24 23:38:45 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T8122 arm64

$ docker run ghcr.io/typst/typst:v0.12.0@sha256:e142773807fb7d0be65908653eeaf0f2e78cae900f594bab41476a93ec9a377a

Welcome to Typst, we are glad to have you here! ❤️
...

$ docker cp $(docker ps -a -q | head -n 1):/bin/typst .
Successfully copied 39.7MB to ...

$ file typst
typst: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), static-pie linked, BuildID[xxHash]=f228d78c86710a56, stripped
```

This issue affects v0.12.0-rc1 and v0.12.0-rc2 as well.

### Reproduction URL

_No response_

### Operating system

Linux, macOS

### Typst version

freundTech:
Fixes #5290.
ARG definitions are only valid for the current build stage in multistage
builds. We need to define them in the correct stage or they will go out of scope in later stages, causing `$TARGETPLATFORM` to be empty.
`xx`/`xx-cargo` will then silently fall back to targeting the build platform, which is x86_64.

I also bumped the action versions while experimenting and thought it would be a good idea to leave this in.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,6 @@
-ARG CREATED
-ARG REVISION
-ARG TARGETPLATFORM
-
 FROM --platform=$BUILDPLATFORM tonistiigi/xx AS xx
 FROM --platform=$BUILDPLATFORM rust:alpine AS build
+
 COPY --from=xx / /
 
 RUN apk add --no-cache clang lld
@@ -14,6 +11,8 @@ RUN --mount=type=cache,target=/root/.cargo/git/db \
     --mount=type=cache,target=/root/.cargo/registry/index \
     CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse \
     cargo fetch
+
+ARG TARGETPLATFORM
 
 RUN xx-apk add --no-cache musl-dev openssl-dev openssl-libs-static
 RUN --mount=type=cache,target=/root/.cargo/git/db \
@@ -26,6 +25,8 @@ RUN --mount=type=cache,target=/root/.cargo/git/db \
     xx-verify target/release/typst
 
 FROM alpine:latest
+ARG CREATED
+ARG REVISION
 LABEL org.opencontainers.image.authors="The Typst Project Developers <hello@typst.app>"
 LABEL org.opencontainers.image.created=${CREATED}
 LABEL org.opencontainers.image.description="A markup-based typesetting system"
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -36,18 +36,18 @@ jobs:
         run: echo "TYPST_BUILD_DATE=\"$(date -u +'%Y-%m-%dT%H:%M:%SZ')\" >> $GITHUB_ENV"
 
       - name: Setup Docker buildx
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -15,6 +12,8 @@ RUN --mount=type=cache,target=/root/.cargo/git/db \
     CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse \
     cargo fetch
 
+ARG TARGETPLATFORM
+
 RUN xx-apk add --no-cache musl-dev openssl-dev openssl-libs-static
```
reknih, line 17: If the silent failure is not desirable, maybe we should explicitly check that this argument is defined.

i.e.
```suggestion
ARG TARGETPLATFORM

RUN if [[ -z "$TARGETPLATFORM" ]] ; then echo "argument TARGETPLATFORM was not provided"; exit 1; fi

RUN xx-apk add --no-cache musl-dev openssl-dev openssl-libs-static
```

freundTech: I thought about that, but it seems kind of redundant to have a line checking if a variable is defined right after the line that defines the variable.

I guess it would protect against someone moving the `ARG` in the future, but only if they don't also move the check with it.

A separate CI step that pulls and runs the built image (using qemu) would probably make more sense, but I haven't looked into that yet. It would be a lot easier with GitHub arm runners. In June GitHub said that they will be available end of year, but haven't said anything since.

laurmaedje:
Thanks!

kiding:
I see the Docker image for 0.12.0 is still broken. @laurmaedje Would it be possible to fix it?

PR #5295, opened 2024-10-25
Fix clipping with outset

### LINKED ISSUE #5293: Wrong clipping with outset
### Description

When using an `outset` on a block or box with `clip` enabled, the clipping mask (expectedly) increases in size, but does not move to respect the outset on the left and top side. This affects all output formats. For example:

```typ
#block(
  clip: true,
  outset: 1em,
  fill: yellow,
  width: 4cm,
  height: 4cm,
)[
  #place(dx: -2em, dy: -2em, top + left, square())
  #place(dx: +2em, dy: -2em, top + right, square())
  #place(dx: -2em, dy: +2em, bottom + left, square())
  #place(dx: +2em, dy:+ 2em, bottom + right, square())
]
```

image

### Reproduction URL

_No response_

### Operating system

Web app, Windows

### Typst version

EpicEricEE:
Fixes #5293

The translation of the frame introduced by the outset was previously not taken into account when creating the clipping rectangle.


laurmaedje:
Thanks!

Issue #5300, opened 2024-10-25
Str documentation should be explict about it using unicode minus


lucaslugao:
Currently the documentation for the [str constructor](https://typst.app/docs/reference/foundations/str/#constructor) does not mention that the unicode minus sign "\u{2212}" is used when converting negative numbers. This can be problematic to users expecting a regular behaviour for that function when scripting.

Adding something like " When converting numerical types the produced string uses the unicode minus sign ("−" U+2212) instead of the ascii minus sign ("-" U+002D)." to the [docustring ](https://github.com/typst/typst/blob/b8034a343831e8609aec2ec81eb7eeda57aa5d81/crates/typst/src/foundations/str.rs#L150) would be enough and prevent people from debugging this odd behaviour.

MDLC01:
Note that Typst allows using U+2212 in place of U+002D in code for the unary and binary minus signs, as well as in `-=`.

lucaslugao:
I stumbled on this while trying to define a dictionary that held information that I wanted to display for items indexed with negative values. Doing `myDict.at(str(-1))` surpisigly wasn't able to access the key in `let myDict = ("-1" : "value")`. Now that I know str doesn't produce "-1" but "−1" I can update the dictionary to use it (or wait on https://github.com/typst/typst/issues/3324).

PR #2936, opened 2023-12-13
Improve number display

MDLC01:
This PR implements https://github.com/typst/typst/pull/2509#issuecomment-1790402443, as suggested in https://github.com/typst/typst/pull/2509#issuecomment-1853639093.

When displaying floats or converting them to strings using [`str`](https://typst.app/docs/reference/foundations/str/#constructor), `-0.0` is converted to `"0"`.

The repr for floats now always includes a decimal point. I did not add a decimal point for values that have units, though (like `repr(1.0em)` is `"1em"`, or `repr(10.0%)` is `"10%"`).

PR #1822, opened 2023-07-29
Use U+2212 MINUS SIGN instead of U+002D HYPHEN MINUS when displaying an `integer` or a `float`

MDLC01:
Fixes #1762

This simply changes the implementation of `Value::display` to use a `−` instead of `-` as the minus sign for `integer`s and `float`s.

## Open questions

1. Should I change the [`str` function](https://typst.app/docs/reference/construct/str/) to use a `−` instead of a `-` too?

2. I'm not sure if I added the tests at the right place.

PR #5301, opened 2024-10-25
Add explicit notice about unicode minus sign in the str function

lucaslugao:
Fixes https://github.com/typst/typst/issues/5300


crates/typst/src/foundations/str.rs
```diff
@@ -131,6 +131,8 @@ impl Str {
     /// - Integers are formatted in base 10. This can be overridden with the
     ///   optional `base` parameter.
     /// - Floats are formatted in base 10 and never in exponential notation.
+    /// - Negative integers and floats are formatted with the unicode minus sign
+    ///   ("−" U+2212) instead of the ascii minus sign ("-" U+002D).
```
MDLC01, line 135:
```suggestion
    /// - Negative integers and floats are formatted with the Unicode minus sign
    ///   ("−" U+2212) instead of the ASCII minus sign ("-" U+002D).
```

lucaslugao: Done.

MDLC01:
Maybe you'd could also specify that `repr` can be used instead of `str` to get a string that uses the ASCII hyphen-minus.

lucaslugao:
`repr` documentation comes with the warning "Note: This function is for debugging purposes. Its output should not be considered stable and may change at any time!" so maybe its not a good idea to point to it?

Also, as per https://github.com/typst/typst/pull/1822#issuecomment-1689881897 `repr` seems to be expected to have the same output as `str` so it might change in the future.

MDLC01:
> Also, as per https://github.com/typst/typst/pull/1822#issuecomment-1689881897 `repr` seems to be expected to have the same output as `str` so it might change in the future.

I don't think there is any more expectation for `repr` to behave like `str`: https://github.com/typst/typst/pull/2936#issuecomment-1853811642.

laurmaedje:
Thanks! 

> [...] so maybe its not a good idea to point to it?

I agree.

PR #5302, opened 2024-10-25
Split out four new crates

laurmaedje:
This PR splits out four new crates: 

- `typst-library`: Most of the code from `typst` moved here. The crate holds Typst's standard library and all central definitions for the compiler (as these are interwoven with the standard library types). All of its definitions are reexported by the `typst` crate, which has become just a thin wrapper.

- `typst-eval`: Moved from `typst/src/eval`.

- `typst-realize`: Moved from `typst/src/realize.rs`.

- `typst-layout`: Contains code from `typst/src/layout` and `typst/src/math`. 

The three new eval/realize/layout crates contain primarily _behaviour,_ not _types._ This behaviour is made available to code in `typst-library` through a new `Routines` struct containing a table of function pointer. This struct is populated in the now very small `typst` crate which depends on all of the four crates and just puts the pieces together.

This set-up of having a central crate with types and surrounding crates with behaviour that are assembled together by one small umbrella crate is very scalable because it makes all the behaviour crates independent from each other, such that they can be compiled in parallel. A very similar design is employed by `rustc`.

Here's a small diagram, which should make things clear:

```
        macros ─────┐     ┌─ kit ───────────────────┐
                    │     ├─ eval ────┐         ┌─ cli
utils ─ syntax ─ library ─┼─ layout ──┼─ typst ─┤
          │               ├─ realize ─┤         └─ ide
timing ───┘               ├─ render ──┤             
                          ├─ svg ─────┤             
                          └─ pdf ─────┘
```

The main motivation of this change is developer experience: Compile times for the `typst` crate have gotten pretty bad and rust-analyzer has become rather sluggish. With the new setup, the `typst-library` crate is still quite big[^1], but at least a bit smaller. It also won't grow as much as the mono-`typst`-crate would due to upcoming features like HTML export (which will get its own `typst-html` crate). Editing `typst-library` is still not a _great_ experience, but I suspect the solution here is to optimize the various proc-macros to generate less code.

Meanwhile, editing in the new crates is a much nicer experience, with rust-analyzer being quite snappy. Moreover, editing these crates does not trigger recompilation of the exporter crates. (Previously, the test cycle for _any change to the layout engine_ triggered a recompilation of `typst-pdf`, `typst-svg`, and `typst-render`.)

[^1]: Splitting `typst-library` further is hard due to the various interactions between all its types. While there _are_ various definitions in it (e.g. data loading functions) that are leafs and not needed by anything else, the boundaries of potentially split-out crates end up being rather arbitrary, so that didn't seem like a durable solution. (I don't want to change the crate structure every week.) The nice thing about just splitting behaviour rather than types is that we can easily resolve cyclid dependencies via the function pointer table. This is not possible types. For what it's worth, many large Rust projects end up with one large core crate surrounded by other crates, so I think we're in good company.


wrzian:
Incredible. Love this

PR #5305, opened 2024-10-27
Support for defining which charset should be covered by a font

### LINKED ISSUE #3331: Font Unicode Range
I propose a new [CSS `unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) like mechanism, to allow more precise control over character's font.

## Proposed syntax

Allow `text(font)` to receive a dict `(family, unicode-range)` or a array thereof.

E.g.:

```typst
#set text(font: (
  (family: "Architects Daughter", unicode-range: (0x2d, 0x3d)),
  "Just Another Hand"
))
```

## Use Case

- Combine two fonts. See https://jakearchibald.com/2017/combining-fonts/ for example.
- #794 

## Why not `show regex`

`show regex` has many caveats. It will break CJK punctuation adjustments. It also have some bad effects to other show rules. It deserve a better solution.

peng1999:
Fix #3331.

See https://github.com/typst/typst/issues/3331#issuecomment-2432803380 for rationale.


laurmaedje:
I think the arguments you make in #3331 are fair. While I do not like stop-gap solutions where obviously better (but harder to implement) solutions are known, that's not really the case here. _Something_ similar to this would likely also be a part of a more comprehensive solution.

#### Notes on the user facing interface

- I think expecting users to know the CJK ranges is maybe a bit much? Could we maybe slightly expand the scope and also have a way to configure script(s) instead of or in addition to ranges?

- The CSS format for ranges feels a bit out of place in Typst. How about using an array instead? 

  ```ts
  Before: "U+00B7, U+2014-3134F"
  After: ("u{b7}", ("\u{2014}", "\u{3134F}"))
  ```

- Not sure whether we want to spell the name as plural or singular. Also not sure whether we need the `unicode-` prefix, it could also be just `ranges`.

#### Notes on the implementation

I do like the way you implemented this, by checking the already shaped glyphs, instead of segmenting ahead of time. It seems like a safer choice since it is less likely to mess things up (e.g. by segmenting variation selectors away or something like that). Still, it might be good (in case you haven't already) to check how other text processors (= browsers) deal with this.

peng1999:
> * I think expecting users to know the CJK ranges is maybe a bit much? Could we maybe slightly expand the scope and also have a way to configure script(s) instead of or in addition to ranges?

I do not recommend maintain a list of CJK ranges in this repo considering its complexity. I think it's better to leave that work to the Universe ecosystem. Below is the comprehensive list maintained by `xeCJK` package for XeLaTeX/LuaLaTeX.

https://github.com/CTeX-org/ctex-kit/blob/0fb196c42c56287403fecca6eb6b137c00167f40/xeCJK/xeCJK.dtx#L2304-L2718



> I do like the way you implemented this, by checking the already shaped glyphs, instead of segmenting ahead of time. It seems like a safer choice since it is less likely to mess things up (e.g. by segmenting variation selectors away or something like that). Still, it might be good (in case you haven't already) to check how other text processors (= browsers) deal with this.

Browsers should follow a more sophisticated method to decide every codepoint's font face before shaping. See [CSS Fonts Module Level 4, §5. Font Matching Algorithm](https://drafts.csswg.org/css-fonts/#cluster-matching), especially §5.3. Cluster matching.

But I think current algorithm in Typst should work in most cases. The only edge case that I can think of is that it will not work well with ligatures. I think this can be left as future work.

peng1999:
Changed the name from `unicode-range` to `ranges`, and adapt the array as parameter. I changed a little bit to reduce one layer of nested parenthesis.

```
Before: "U+00B7, U+2014-3134F"
After: ("u{b7}", "\u{2014}-\u{3134F}")
```

laurmaedje:
We discussed this in the team and came to the following conclusions:

- The double-nested arrays are indeed a bit problematic, especially if it's just a single range, because then you'd have to write something like `(("\u{2014}", "\u{3134F}"),)`. `("u{b7}", "\u{2014}-\u{3134F}")` is a bit less conceptually clean due to the involved string parsing, but it's an acceptable trade-off. In theory, we could also express the char filter as a regex (just like the ranges, not via `show`), but it's probably a bit odd and overkill.

- Regarding CJK ranges: We are of the opinion that solid, internationalized text processing is the responsibility of the compiler, not of packages. Expecting CJK users to load a package from universe to correctly configure their font is bad UX. I am unsure (a) whether there is one relevant set of CJK ranges or different ones for different use cases and (b) how much the CJK ranges could expressed via the Unicode Script property. If they are far removed from just being a few scripts _but_ there's still just one relevant set, we could support `ranges: "cjk"` (maybe also `ranges: "emoji"`). If there are different relevant ranges, they could have different names, or we'd have to come up with something else. If the ranges are too much for this repository, we can always put them into a separate repo & crate.

peng1999:
> * I am unsure (a) whether there is one relevant set of CJK ranges or different ones for different use cases and (b) how much the CJK ranges could expressed via the Unicode Script property. If they are far removed from just being a few scripts _but_ there's still just one relevant set, we could support `ranges: "cjk"` (maybe also `ranges: "emoji"`).

After some investigation, I found that only 11 punctuation marks are used in both CJK and Latin scripts. (This is also recognized in [xeCJK](https://github.com/CTeX-org/ctex-kit/blob/0fb196c42c56287403fecca6eb6b137c00167f40/xeCJK/xeCJK.dtx#L2345))

```
U+00B7: ·
U+2013: –
U+2014: —
U+2018: ‘
U+2019: ’
U+201C: “
U+201D: ”
U+2025: ‥
U+2026: …
U+2027: ‧
U+2E3A: ⸺
```

When in CJK contexts, we prefer using the glyphs of the above codepoint from CJK fonts. For other codepoints that exists both in CJK fonts and Latin fonts, usually we prefer to use the glyphs from Latin fonts.

So I think a solution is to name the above codepoints range to `cjk-ambiguous`, and use excluding ranges:

```typ
#set text(font: (
   (name: "My Latin Font", exclude-range: ("cjk-ambiguous",)),
   "My CJK Font"
))
```

laurmaedje:
The desire to also have an inverted option kind of favors my idea of a regex-based design, since it could trivially express this negation.

peng1999:
> The desire to also have an inverted option kind of favors my idea of a regex-based design, since it could trivially express this negation.

Using regex seems make more sense now. We can use `regex-syntax` (Already being in dependency graph) to limit the regex to single char literal and bracketed character class. And provide a special char class `[:cjk-ambiguous:]` described above. Then the users can write

```typ
#set text(font: (
   (name: "My Latin Font", range-match: "[[:^cjk-ambiguous:]]"),
   "My CJK Font"
))
```

or any custom ranges

```typ
#set text(font: (
   (name: "My Cool Font", range-match: "[\u{b7}\u{2014}-\u{3134F}]"),
))
```

laurmaedje:
I'll discuss this with the rest of the team.

laurmaedje:
I've discussed this with the team. A few things:
- If we use a regex-based design, I'm strongly in favor of actually expecting a value of type [`regex`](https://typst.app/docs/reference/foundations/regex/) rather than just interpreting a string as a regex.
- I don't think we should add custom character classes to the regex. Rather, I would expose a predefined named string for use with Latin fonts in CJK.
- Maybe `covers` would be a good name for the thing?

So the low-level API would look like this:
```typ
// Change only font of the numbers.
#set text(font: (
   (name: "Number Font", covers: regex("[0-9]")),
   "Base Font"
))
```

And the "high-level" (i.e. predefined) API would look like this:
```typ
// Change only font of the numbers.
#set text(font: (
   (name: "Latin Font", covers: "latin-in-cjk"),
   "CJK Font"
))
```

With `"latin-in-cjk"` being the inverse of the CJK ambigious set.

I think we should also investigate the performance impact of running a regex on every char.

peng1999:
API has been changed according to feedback.

Based on my preliminary and informal analysis, it appears that the regex does not significantly impact performance. My test file is:

```typ
#set text(font: "Libertinus Serif")
#for i in range(1000) {
  lorem(1000)
  [

  ]
}
```

Benchmark on AMD Ryzen 7 5800X:

```console
$ hyperfine -N 'target/release/typst compile /tmp/a.typ'
Benchmark 1: target/release/typst compile /tmp/a.typ
  Time (mean ± σ):     721.0 ms ±  13.0 ms    [User: 498.8 ms, System: 225.4 ms]
  Range (min … max):   702.9 ms … 742.0 ms    10 runs
```
If I change the `"Libertinus Serif"` to `(name: "Libertinus Serif", covers: regex("[a-zA-Z]"))`, the result is
```console
$ hyperfine -N 'target/release/typst compile /tmp/a.typ'
Benchmark 1: target/release/typst compile /tmp/a.typ
  Time (mean ± σ):     724.5 ms ±   9.5 ms    [User: 530.3 ms, System: 219.7 ms]
  Range (min … max):   711.6 ms … 736.4 ms    10 runs
```

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -463,7 +463,13 @@ impl<'a> ShapedText<'a> {
...
         };
         let mut chain = families(self.styles)
-            .map(|family| book.select(family, self.variant))
+            .map(|family| {
+                family
+                    .coverage()
+                    .map_or(true, |c| c.is_match("-"))
+                    .then(|| book.select(family.as_str(), self.variant))
+                    .flatten()
+            })
```
laurmaedje, line 472:
```rust
let mut chain = families(self.styles)
    .filter(|family| family.coverage().map_or(true, |c| c.is_match("-")))
    .map(|family| book.select(family.as_str(), self.variant))
    .chain(fallback_func.iter().map(|f| f()))
    .flatten();
```
would be a bit simpler perhaps

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -795,6 +808,15 @@ fn shape_segment<'a>(
...
     let ltr = ctx.dir.is_positive();
 
+    let char_in_coverage = |char_start| {
+        let char_end = text[char_start..]
+            .char_indices()
+            .nth(1)
+            .map(|(offset, _)| offset + char_start)
+            .unwrap_or(text.len());
+        coverage.map_or(true, |cov| cov.is_match(&text[char_start..char_end]))
+    };
```
laurmaedje, line 818: Perhaps it would be easier to use the existing `let c = text[cluster..].chars().next().unwrap();` and move it upwards and then check `coverage.map_or(true, |cov| cov.is_match(c.encode_utf8(&mut [0; 4])))`? I'm not 100% certain whether it's safe to move the let up, but in principle the clusters should all correspond to valid char indices...

laurmaedje: okay, I see now that you also use it below, so perhaps it's not better. not sure ...

crates/typst-library/src/text/mod.rs
```diff
@@ -94,7 +95,17 @@ pub(super) fn define(global: &mut Scope) {
...
 #[elem(Debug, Construct, PlainText, Repr)]
 pub struct TextElem {
-    /// A font family name or priority list of font family names.
+    /// A font family descriptor or priority list of font family descriptor.
+    ///
+    /// The font family descriptor can be a string representing the family name
+    /// or a dictionary with the following keys:
+    ///
+    /// - `name` (required): The font family name.
+    /// - `covers` (optional): A regex that defines the Unicode points
```
laurmaedje, line 104:
```suggestion
    /// - `covers` (optional): A [regex] that defines the Unicode codepoints
```

crates/typst-library/src/text/mod.rs
```diff
@@ -94,7 +95,17 @@ pub(super) fn define(global: &mut Scope) {
...
+    ///   supported by the font. Unicode codepoints that match this regex are
+    ///   considered within the range and will be rendered using this font.
+    ///   A special `"latin-in-cjk"` value can also be specified to cover all
+    ///   codepoints except that both used in Latin and CJK fonts.
```
laurmaedje, line 108:
```suggestion
    ///   Alternatively, a built-in `"latin-in-cjk"` value can be specified to cover all
    ///   codepoints except those used both in Latin and CJK fonts.
```

crates/typst-library/src/text/mod.rs
```diff
@@ -129,6 +140,19 @@ pub struct TextElem {
     ///
     /// This is Latin. \
     /// هذا عربي.
+    ///
+    /// // Change font only for numbers.
+    /// #set text(font: (
+    ///   (name: "New Computer Modern", covers: regex("[0-9]")),
+    ///   "Linbertinus Serif"
+    /// ))
+    ///
```
laurmaedje, line 149: A bit of text that demonstrates the example would good here.

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
...
+        self.name.hash(state);
+        self.coverage_str().hash(state);
+    }
+}
+
+impl PartialEq for FontFamily {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name && self.coverage_str() == other.coverage_str()
+    }
+}
```
laurmaedje, line 810: Those can be derived since Typst's `Regex` type already handles this

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
 }
 
 /// A lowercased font family like "arial".
-#[derive(Clone, Eq, PartialEq, Hash)]
-pub struct FontFamily(EcoString);
+#[derive(Debug, Clone)]
+pub struct FontFamily {
+    name: EcoString,
+    coverage: Option<Regex>,
```
laurmaedje, line 796:
```suggestion
    covers: Option<Regex>,
```
Brief docs on the fields would be nice.

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
...
+}
 
 impl FontFamily {
     /// Create a named font family variant.
     pub fn new(string: &str) -> Self {
-        Self(string.to_lowercase().into())
+        Self::new_with_coverage(string, None).unwrap()
+    }
+
+    pub fn new_with_coverage(
```
laurmaedje, line 818: A bit of docs here would be nice as well.

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
...
+        string: &str,
+        coverage: Option<Value>,
```
laurmaedje, line 820: I think we could have a new `enum Covers { LatinInCjk, Regex(Regex) }` type that uses `cast!` to cast from a string or regex without allowing an arbitrary `Value` here.

We could even use that across the board, never converting LatinInCjk to a regex in the first place.

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
...
+                let ast = regex_syntax::ast::parse::Parser::new()
+                    .parse(regex.as_str())
+                    .unwrap();
+                match ast {
+                    regex_syntax::ast::Ast::ClassBracketed(..)
+                    | regex_syntax::ast::Ast::ClassUnicode(..)
+                    | regex_syntax::ast::Ast::ClassPerl(..)
+                    | regex_syntax::ast::Ast::Dot(..)
+                    | regex_syntax::ast::Ast::Literal(..) => (),
+                    _ => bail!("invalid regex in font coverage"),
```
laurmaedje, line 844: I don't think we need to restrict the regex in this way, to be honest. If the regex is multi-char, it just won't be a match, that's on the user. Perhaps attention should be drawn to this on the docs, but matching on the regex syntax seems a bit bug-prone.

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
...
+        string: &str,
+        coverage: Option<Value>,
+    ) -> HintedStrResult<Self> {
+        let coverage = match coverage {
+            Some(Value::Str(name)) => {
+                if &*name == "latin-in-cjk" {
+                    static LATIN_IN_CJK_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+                        Regex::new( "[^\u{00B7}\u{2013}\u{2014}\u{2018}\u{2019}\u{201C}\u{201D}\u{2025}-\u{2027}\u{2E3A}]").unwrap()
+                    });
+                    Some(LATIN_IN_CJK_REGEX.clone())
```
laurmaedje, line 828: If we keep the regex, it can use the `typst_utils::singleton` macro

crates/typst-library/src/text/mod.rs
```diff
@@ -766,35 +790,97 @@ impl PlainText for Packed<TextElem> {
...
 
-impl Debug for FontFamily {
-    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        self.0.fmt(f)
+    /// The user-set coverage of the font family.
+    pub fn coverage(&self) -> Option<&Regex> {
+        self.coverage.as_ref()
+    }
+
+    pub fn coverage_str(&self) -> Option<&str> {
```
laurmaedje, line 863: can be removed then

crates/typst-library/src/text/mod.rs
```diff
@@ -818,20 +904,22 @@ cast! {
...
-        "twitter color emoji",
-        "noto color emoji",
-        "apple color emoji",
-        "segoe ui emoji",
-    ];
-
-    let tail = if TextElem::fallback_in(styles) { FALLBACKS } else { &[] };
-    TextElem::font_in(styles)
+pub fn families(styles: StyleChain) -> impl Iterator<Item = &FontFamily> + Clone {
+    static FALLBACKS: LazyLock<Vec<FontFamily>> = LazyLock::new(|| {
```
laurmaedje, line 908: This can just the `typst_utils::singleton` macro.

crates/typst-layout/src/inline/shaping.rs
peng1999: At first I tried the way you described but it make code below more complex, and it have performance issue. So I switched to the current `char_in_coverage` function.

crates/typst-library/src/text/mod.rs
peng1999: The `regex_syntax` crate is exactly the underlying component used by the `regex` crate. Why is it bug-prone? 

I restrict the regex syntax based on the following concerns:
- The bracket syntax in regex alone is flexible enough, users can even construct set union or intersections like `[\d&&\p{ascii}]`. Syntax like alternative or repeat do not worth the trouble in this context.
- If we want to implement sophisticated fallback algorithm like in [CSS](https://drafts.csswg.org/css-fonts/#cluster-matching) in the future, then the restricted regex syntax can be easily transformed to codepoints set without introduce API breakage.

laurmaedje: I meant bug-prone in the sense that we need to decide which regex syntax is useful. But I can see how it's useful to reserve the ability to transform bijectively to/from a codepoint set. However, I think we then need to document which parts of the regex syntax are permitted.

laurmaedje:
Thank you! This was a tricky one, especially in terms of design, but now it's through. :)

PR #5306, opened 2024-10-27
Fix wrong drop order for self-referential struct

SkiFire13:
While looking around I noticed that `Repr` is a self-referential struct. The order of fields in a self-referential struct is important because [it's also the order in which they're dropped](https://doc.rust-lang.org/reference/destructors.html#destructors.operation), and if the borrowed field is dropped before the borrowing ones this would lead to those temporarily becoming dangling. Unfortunately `Repr` had the wrong field order, so I fixed it and left a comment for why it's important.


laurmaedje:
Thanks! Good catch. I don't think anything bad could've happened since ttf_parser's Face shouldn't have any drop code at all and rustybuzz's just a few vectors whose contents won't be read, but of course this isn't generally a safe assumption to make.

PR #5316, opened 2024-10-28
Make `math.bold` compatible with Greek digamma

MDLC01:
This small PR makes the `math.bold` function work with the archaic Greek letter [digamma](https://en.wikipedia.org/wiki/Digamma). I don't think Unicode defines anything other than bold serif variants of digamma, or variants of other archaic Greek letters.


laurmaedje:
Thanks!

Issue #4699, opened 2024-08-07
Path with multiple subpaths


joneuhauser:
### Description

Related to https://github.com/typst/typst/issues/4123, but I'm not convinced of the syntax suggestion there.

In order to draw paths with holes, it would be necessary to specify that a path has multiple subpaths and their fill rule.

I'm not sure how the syntax would look like. For backwards compatibility reasons, I don't think the syntax of `#path` should be changed, there should probably be a new function `#compound_path` or similar:

```
#compound_path(
  fill: blue.lighten(80%),
  stroke: blue,
  fill-rule: evenodd,
  (path1, closed), 
  (path2, closed),
)
```

where path1 is a list of vertex specifications as in the `#path`

### Use Case

In order to implement https://matplotlib.org/stable/gallery/shapes_and_collections/donut.html#sphx-glr-gallery-shapes-and-collections-donut-py in typst-mpl-backend, this feature is required.

Emm54321:
I made a simple implementation of this feature by allowing separating components in #path by `auto`/`true`/`false` (`true`: end a closed component, `false`: end an open component, `auto`: use the value of the `closed` parameter).

```typst
#path(
  closed: true,
  fill-rule: "even-odd",
  fill: yellow,
  stroke: black,
  (0mm, 20mm),
  (20mm, 20mm),
  (20mm, 0mm),
  auto, // separate the two triangles
  (7mm, 17mm),
  (17mm, 17mm),
  (17mm, 7mm),
  // implicit auto at the end unless an explicit marker is used
)
```

Not sure it's the best syntax, but it doesn't break any existing code, and it is a very simple patch.
I can submit a PR if you're interested.

joneuhauser:
@Emm54321 I suggested this kind of syntax on discord, but @laurmaedje suggested a different syntax:
```
#path(
path.moveto(5mm, 2mm),
path.lineto(1mm, 2mm, relative: True),
...
)
```

that is, basically following the ubiquitous SVG syntax. The `relative` parameter would be optional.

Emm54321:
This would be cleaner indeed. Implementing the syntax I used was a 10min job. This requires more work but doesn't seem that complicated.

Emm54321:
I got it working with the syntax above. It supports `path.moveto()`, `path.lineto()`, `path.cubicto()`, `path.quadto()` and `path.close()`.
It currently supports several syntaxes, e.g.:
```typst
[...]

PR #5323, opened 2024-10-29
New `curve` element that supersedes `path`

Emm54321:
Closes [4699](https://github.com/typst/typst/issues/4699) and allow the creation of paths with multiple components, e.g.:

```typst
#path(
  fill: yellow,
  stroke: black,
  fill-rule: "even-odd",
  path.moveto(10mm, 10mm),
  path.lineto(20mm, 10mm),
  path.lineto(20mm, 20mm),
  path.close(),
  path.moveto(0mm, 5mm),
  path.lineto(25mm, 5mm),
  path.lineto(25mm, 30mm),
  path.close(),
)
```

It supports cubic Bezier curves:
```typst
#path(
  fill: yellow,
  stroke: black,
  path.moveto(start: (10mm, 10mm)),
  path.cubicto(cstart:(5mm, 20mm), cend:(15mm, 20mm), end:(20mm, 0mm), relative: true),
  // auto for cstart means: mirror the previous control point.
  path.cubicto(cstart: auto, cend:(15mm, -10mm), end:(20mm, 0mm), relative: true),
  path.close()
)
```
Or with more compact syntax :
```typst
#path(
  fill: yellow,
  stroke: black,
  path.moveto(10mm, 10mm),
  path.cubicto(5mm, 20mm, 15mm, 20mm, 20mm, 0mm, relative: true),
  path.cubicto(auto, 15mm, -10mm, 20mm, 0mm, relative: true),
  path.close()
)
```
Or:
```typst
#path(
  fill: yellow,
  stroke: black,
  path.moveto((10mm, 10mm)),
  path.cubicto((5mm, 20mm), (15mm, 20mm), (20mm, 0mm), relative: true),
  path.cubicto(auto, (15mm, -10mm), (20mm, 0mm), relative: true),
  path.close()
)
```

Quadratic Bezier curves are supported too (with similar syntax shortcuts):
```typst
#path(
  closed: true,
  //path.moveto(start: (0mm, 0mm)), implicit
  path.quadto(control: (10mm, 0mm), end:(10mm, 10mm)),
  path.quadto(control: auto, end:(10mm, 10mm), relative: true),
  //path.close() added by the `closed` flag.
)
```

Notes:
- Not sure if all syntax shortcuts should be kept.
- `close` adds a straight line, like SVG or PDF. Maybe it should have a flag to use a curve like it does with the old syntax.
- Relative coordinates are relative to the start of the element, like for SVG. Maybe the second control point of a cubic curve should be relative to the end point.
- Missing control points default to the end point, except for the first control point of a cubic curve which defaults to the start point. `auto` is used to tell that a control point should be mirrored.
- The old syntax is still valid. Consecutive old-style arguments in `path()` are handled as a separate component when mixed with new-style elements.
- Maybe names like `line-to` would be better than `lineto`. 
- The documentation is incomplete until the syntax is finalized.
- The `vertices` name of the `path` element is kept for compatibility, but is not so great for new path elements.
- `path.rectangle`, `path.circle`, `path.ellipse` elements could be added later.


crates/typst-layout/src/shapes.rs
```diff
@@ -90,28 +79,164 @@ pub fn layout_path(
...
+                            Some(Smart::Custom(c)) => {
+                                let mut c = resolve(*c);
+                                if *rel {
+                                    c += last;
+                                }
+                                c
+                            }
+                            Some(Smart::Auto) => 2. * last - last_control,
+                            None => last,
+                        };
```
laurmaedje, line 206: There is quite a bit of code duplication here. Overall, I feel like the function `layout_path` has gotten a bit too large here, and some factoring out is needed, either into subfunctions, or maybe into a struct with builder-style methods.

laurmaedje (commented):
Thanks for the PR! I left some initial, high-level comments on the code.

To your notes:

> Not sure if all syntax shortcuts should be kept.

I'm not sure the arguments really need to support both positional and named at the same time. The `#[parse(` handlers are a bit much.

> close adds a straight line, like SVG or PDF. Maybe it should have a flag to use a curve like it does with the old syntax.

Something to close naturally sounds useful. Ideally, the new way would support a superset of the functionality of the old syntax.

> Relative coordinates are relative to the start of the element, like for SVG. Maybe the second control point of a cubic curve should be relative to the end point.

> Missing control points default to the end point, except for the first control point of a cubic curve which defaults to the start point. auto is used to tell that a control point should be mirrored.

Let's first get the simpler feedback worked in and then I'll take a closer look at these details later.

> The old syntax is still valid. Consecutive old-style arguments in path() are handled as a separate component when mixed with new-style elements.

That's good. We can think about deprecating it later.

> Maybe names like line-to would be better than lineto.

`line-to` is better than `lineto`, but maybe just `line` is enough?

> The vertices name of the path element is kept for compatibility, but is not so great for new path elements.

Hmm, yeah. Let's leave it for now and consider changing it later.

Emm54321:
- The `moveto`, `lineto` etc... have been renamed to `move`, `line`, etc.
- A `path.vertex` element has been added that is equivalent to the old syntax, and can be mixed with `path.line`&co.
- A `close-mode` parameter has been added to `path` that can be set to `"curve"` (the default, equivalent to the original behaviour), `"line"`, or `none` (don't add a closing line, just end the component), and can be overriden by a `mode` parameter in `path.close`.
- Positional arguments are no longer supported.
- `Packed<*>` are retained.
- The path building code has been cleaned up and refactored using a builder object.

So now we have e.g.:

```typst
#path(
  fill: yellow,
  stroke: black,
  path.move(start: (10mm, 10mm)),
  path.line(end: (20mm, 10mm)),
  path.line(end: (20mm, 20mm)),
  path.close(),
  path.move(start: (0mm, 5mm)),
  path.line(end: (25mm, 5mm)),
  path.line(end: (25mm, 30mm)),
  path.close(),
)

#path(
  fill: yellow,
  stroke: black,
  path.move(start: (10mm, 10mm)),
  path.cubic(cstart:(5mm, 20mm), cend:(15mm, 20mm), end:(20mm, 0mm), relative: true),
  // auto for cstart means: mirror the previous control point.
  path.cubic(cstart: auto, cend:(15mm, -10mm), end:(20mm, 0mm), relative: true),
  path.close(mode: "line")
)

#path(
  closed: true,
  close-mode: "line",
  // path.move(start: (0mm, 0mm)), // implicit
  path.quadratic(control: (10mm, 0mm), end:(10mm, 10mm)),
  path.quadratic(control: auto, end:(10mm, 10mm), relative: true),
  // path.close() // added by the `closed` flag.
)

// Or the equivalent to the old syntax, with control points relative to the vertex:
#path(
  fill: yellow,
  stroke: black,
  closed: true,
  path.vertex(point: (20mm, 20mm), control-into: (-5mm, 0mm), control-from: auto),
  path.vertex(point: (30mm, 0mm), control-into: (0mm, 5mm), control-from: auto),
  path.vertex(point: (5mm, 5mm), control-into: (0mm, -5mm), control-from: auto),
)

```

istudyatuni:
Please [include word fix/close](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) before issue mention to close it automatically on merge

laurmaedje:
The reason I'm delaying a bit here is because we want to use the identifier `path` for a file path type and are looking for a new name for the visual paths.  I'd prefer to not expose a new nice visual-path API in the next release and then deprecate it immediately. 

Another option we've been exploring is extending the `path` API to a more canvas-style thing (naming it `canvas` or `drawing` or something like that), but that seems like a larger task and I'm not sure I want to block things on it.

In any case, the compatibility behaviour could still be kept (i.e. the `path` type's constructor emitting a warning and returning a visual path when its arguments consist of points rather than a string).

laurmaedje:
I'm leaning towards renaming `path` to `curve` to make space for a new file `path` type. 

Not sure how best to proceed here. I see two options:
- We could merge this and do the rename separately
- We could introduce the new sub-element API with a new `CurveElem` here (with no compatibility behaviour) and turn the existing `PathElem` into just a `path` `#[func]`-tion that accepts the old API and creates a `CurveElem` under the hood. This is still a breaking change, but a smaller one than an outright removal (only set/show rules break, not calls). The compatibility behaviour can then be integrated into the new `path` type's constructor once it's there.

Emm54321:
IMHO, the second option is better. People using the new API shouldn't have to change their code later if it can be avoided.
Maybe the new `CurveElem` could be introduced while keeping the old `PathElem` at first but marking it as deprecated, and wait until the new `path` type is there to add the compatibility behavior making it create `CurveElem` and remove the current `PathElem`. That would give people time to adjust their code.

laurmaedje:
Yeah, that sounds good.

Emm54321:
Ok, I'll make the changes.

Emm54321:
- `PathElem` is restored to its original state but documented as deprecated.
- `CurveElem` is added without support for the compatibility syntax.
- Not sure the `close` flag should be kept, the `curve.close()` element may be enough?
- The documentation probably needs some additional work.
- `curve.rect`, `curve.circle` and `curve.ellipse` elements should probably be added.

We now have:
```typst
#curve(
  fill: yellow,
  fill-rule: "even-odd",
  stroke: black,
  curve.move(start: (10mm, 10mm)),
  curve.line(end: (20mm, 10mm)),
  curve.line(end: (20mm, 20mm)),
  curve.close(),
  curve.move(start: (0mm, 5mm)),
  curve.line(end: (25mm, 5mm)),
  curve.line(end: (25mm, 30mm)),
  curve.close(),
)

#curve(
  fill: yellow,
  stroke: black,
  curve.move(start: (10mm, 10mm)),
  curve.cubic(control-start:(5mm, 20mm), control-end:(15mm, 20mm), end:(20mm, 0mm), relative: true),
  // auto for control-start means: mirror the previous control point.
  curve.cubic(control-start: auto, control-end:(15mm, -10mm), end:(20mm, 0mm), relative: true),
  curve.close(mode: "line")
)

#curve(
  closed: true,
  close-mode: "line",
  // curve.move(start: (0mm, 0mm)), // implicit
  curve.quadratic(control: (10mm, 0mm), end:(10mm, 10mm)),
  curve.quadratic(control: auto, end:(10mm, 10mm), relative: true),
  // curve.close() // added by the `closed` flag.
)

// Or the equivalent to the old syntax, with control points relative to the vertex:
#curve(
  fill: yellow,
  stroke: black,
  closed: true,
  curve.vertex(point: (20mm, 20mm), control-into: (-5mm, 0mm), control-from: auto),
  curve.vertex(point: (30mm, 0mm), control-into: (0mm, 5mm), control-from: auto),
  curve.vertex(point: (5mm, 5mm), control-into: (0mm, -5mm), control-from: auto),
)
```

laurmaedje:
> Not sure the close flag should be kept, the curve.close() element may be enough?

Yes, let's remove it.

> The documentation probably needs some additional work.

I'll do a pass over it before merging, but feel free to improve it in the meantime. I also noticed that some spaces are missing after backticks. Could you fix those and also wrapp all comments at column 80?

> curve.rect, curve.circle and curve.ellipse elements should probably be added.

I'd rather keep it minimal for now.

-----------

Further opinions from me and @reknih:

- I think we can remove the `curve.vertex` API since we don't need to be compatible with the old `path` approach and the new APIs are much easier to understand.

- The `point` arguments should be positional. The `control-start` and `control-end` points should be positional and required as well (with `auto` having to be specified explicitly).

- The `To` suffices can be removed from the struct names.

- `curve.quadratic` should be `curve.quad`.

- The existing `auto` for control points supports the behaviour "Mirror angle & length". We could also support mirroring the angle only by specifying a single length, though this is not crucial and it might be too implicit.

Emm54321:
- The `closed` parameter has been removed, as well as the curve.vertex element.
- Points and control points are positional. Control points can be set to `none` to put them on the corresponding point.
- Use `curve.quad` instead of `curve.quadratic`.
- Cleaned up doc and comments.

Current syntax:
```typst
#curve(
  fill: yellow,
  fill-rule: "even-odd",
  stroke: black,
  curve.move((10mm, 10mm)),
  curve.line((20mm, 10mm)),
  curve.line((20mm, 20mm)),
  curve.close(),
  curve.move((0mm, 5mm)),
  curve.line((25mm, 5mm)),
  curve.line((25mm, 30mm)),
  curve.close(),
)

#curve(
  fill: yellow,
  stroke: black,
  curve.move((10mm, 10mm)),
  curve.cubic((5mm, 20mm), (15mm, 20mm), (20mm, 0mm), relative: true),
  // auto for control-start means: mirror the previous control point.
  curve.cubic(auto, (15mm, -10mm), (20mm, 0mm), relative: true),
  curve.close(mode: "line")
)

#curve(
  close-mode: "line",
  // curve.move(start: (0mm, 0mm)), // implicit
  curve.quad((10mm, 0mm), (10mm, 10mm)),
  curve.quad(auto, (10mm, 10mm), relative: true),
  curve.close()
)
```

laurmaedje:
I refactored this a bit. In particular, I made the naming change path -> curve consistent across the entire codebase. I also removed the `close-mode` and cleaned up the docs a bit. Should be mergeable now.

laurmaedje:
Added better docs & split up the tests.

laurmaedje:
Thank you for your work on this!

Issue #5325, opened 2024-10-30
Add Numero sign symbol (№)


Andrew15-5:
### Description

https://en.wikipedia.org/wiki/Numero_sign

Unicode code point: U+2116

### Use Case

At least in Russian, it's a very common symbol.

PR #5326, opened 2024-10-30
Added Numero Sign № (U+2116)

Andrew15-5:
Closes https://github.com/typst/typst/issues/5325.

As a side note, I didn't get why not everything in each section/group is ordered alphabetically. I thought that was the whole point of sorting. I can fix that in a separate PR, if needed. I was confused as to where in the semi-ordered section to put the new symbol.

### DIFF
--- a/crates/typst-library/src/symbols/sym.rs
+++ b/crates/typst-library/src/symbols/sym.rs
@@ -123,6 +123,7 @@ pub(crate) const SYM: &[(&str, Symbol)] = typst_macros::symbols! {
     interrobang: '‽',
     hash: '#',
     hyph: ['‐', minus: '\u{2D}', nobreak: '\u{2011}', point: '‧', soft: '\u{ad}'],
+    numero: '№',
     percent: '%',
     permille: '‰',
     pilcrow: ['¶', rev: '⁋'],


MDLC01:
As per the latest [Unicode Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G20429), "U+2116 NUMERO SIGN is provided both for Cyrillic use, […] and for compatibility with Asian standards." For this reason, I'm not sure it is right to add it in `sym`, because it should not be used outside of Cyrillic or CJK texts. I fear making this character so easily available may lead to misuse.

My reasoning is that a character that should only be used in a specific script is probably easy to type with an appropriate keyboard, and so there is no reason to add it to `sym`. Moreover, not adding it to `sym` prevents misuse by other users. In part for this reason, we plan to remove `degree.c`, `degree.f`, `ohm`, `angstrom`, and `kelvin`.

Is the numero sign easy to type on a Russian keyboard? (The Wikipedia page you links explains that "On Russian computer keyboard layout, the <kdb>№</kbd> character is available and often located on the <kbd>3</kbd> key.") If it is, I think it should not be added to `sym`. Otherwise, I believe we should think about a solution to prevent this symbol from being misused in other scripts.

> As a side note, I didn't get why not everything in each section/group is ordered alphabetically.

Symbols are ordered partially alphabetically, and partially thematically, within each section. In the end, the order does not really matter. However, for symbol variants, changing the order may cause some breaking changes. So any reordering should be thought carefully.

jamesrswift:
Numero is used extensively in french too (and I'm quite partial to it also)

elegaanz:
In french we generally don't use the dedicated character, but a "n" followed by a "°" (degree sign) or a "<sup>o</sup>" (superscript o), as said in the linked Unicode specification. And very often it is written using a lower case "n".

Andrew15-5:
> I fear making this character so easily available may lead to misuse.

What is considered a misuse?

> Is the numero sign easy to type on a Russian keyboard?

I wouldn't say easy, but it is accessible. For Glove80, I mapped the hash key to always print hash with both OS layouts (for consistency just like all other symbols). I mapped the hold action to print Numero for both OS layouts. Previously I also wanted to use in a template, which means that I used US layout, which means that № is not directly accessible. For English the hash key hold is mapped to use X11 compose feature: RAlt+N+o. So, in some situations and considering the regular keyboard, it's not easy at all. If you can't use the compose feature, then you would have to switch the Russian OS layout just to write 1 symbol. And if you don't have Russian layout, then you are out of options. Go and search for the symbol manually, wasting your time.

So I'm not convinced that it's such a horrible symbol that it must never be included in Typst. I was disappointed and bamboozled several times already. Each time I think that it might be there, or I just forget that it's not there.

I really did not expect it to be basically just Russian and some "Asian standards". I thought it's also used in Europe. I think that usually the # is used in US, no? Or do the majority of countries use it?

MDLC01:
> Numero is used extensively in french too

Indeed, but this specific Unicode character should not be used in French. As explained by @elegaanz and in the Unicode specification I linked above, a "N" or "n" (depending on the context) followed by a superscript "o" (or "os" for the plural form) should be used. This is not the case in Russian, where a different glyph is used, as illustrated by the picture below.

For context, this is a picture I took of a box that can be seen at the _quadrilatère des Archives nationales_. There is no date indicating when this text was written, but I think it likely predates Unicode.
"№" is used in a Cyrillic context.

Your comment illustrates that adding this in `sym` would likely encourage people to use it in inappropriate places (which of course is completely understandable, we can't expect people to know the Unicode specification).

MDLC01:
> What is considered a misuse?

If I understand the Unicode specification properly, using this character outside of Cyrillic or CJK text constitutes a misuse.

> I thought it's also used in Europe.

As I explained above, French uses a regular "n" and a superscript "o". I can't speak for other European languages, though, as I don't know them and Unicode does not mention them.

Andrew15-5:
> There is no date indicating when this text was written

It does say СССР a.k.a. Soviet Union.

> If I understand the Unicode specification properly, using this character outside of Cyrillic or CJK text constitutes a misuse.

Well, sure, but I mean the misuse that will cause real problems. It's not like I'm trying to extract `sequence` or `styled` functions here. So let 'em use it, if they need/want to?

> I can't speak for other European languages, though, as I don't know them and Unicode does not mention them.

Perhaps, we can wait for more input on this.

MDLC01:
> It does say СССР a.k.a. Soviet Union.

Indeed, thanks!

Enivex:
It's `&numero;` in HTML @MDLC01 . I think the correct way to deal with this is to document the intended use, not to block it from entering `sym`.

laurmaedje:
Seems like there is a consensus for adding this.

Issue #5327, opened 2024-10-30
Set `*.reflow` `true` by default


Andrew15-5:
### Description

The `reflow` field should've been set to `true` a long time ago. The last instance for this is https://github.com/typst/typst/pull/5136#discussion_r1789849547.

### Use Case

Reflow by default is the better default, because that's what most users would expect, IMO.

PR #5330, opened 2024-10-30
Changed default value of `reflow` to `true`

Andrew15-5:
Fixes https://github.com/typst/typst/issues/5327.

Perhaps the docs can be improved even more. And someone should check the tests.


laurmaedje:
I think the intent of many of the tests is lost here. Figuring out which should and shouldn't be adjusted is mostly a matter of looking at each test and understanding its intent.

Andrew15-5:
Should I change them?

laurmaedje:
Yes, please adjust them.

Andrew15-5:
Wow, this freezes Okular after I click on the link:

```typ
#set rotate(reflow: true)
#set scale(reflow: true)
#let mylink = link("https://typst.org/")[LINK]

#lorem(50)
My cool #box(move(dx: 0.7cm, dy: 0.7cm, rotate(10deg, scale(200%, mylink))))
#lorem(50)
```

Then the kded5 process starts the fans going. After I kill it, the link opens in the browser. Then I have to kill it 2 more times. `okular 23.08.5`

Works fine with MuPDF.

Andrew15-5:
> ```typ
> // You shouldn't be able to see the text, if you can then
> // that means that the transform matrices are not being
> // applied to the text correctly.
> ```

I can see the text. `pattern-text`

Andrew15-5:
If I keep it, maybe the text should be inside the rect and not outside?

image

Andrew15-5:
```typ
#set rotate(reflow: true)
#set page(height: 80pt)
#align(center + horizon, rotate(0deg, image("images/tiger.jpg")))
```

image

I'm really struggling to comprehend why setting reflow causes the image to become huge just by rotating it 3 degrees? The page height is constant, and the width is the same.

image

And this is the same for all `transform-*` tests. Well, maybe for almost all tests that were affected.

laurmaedje:
> I'm really struggling to comprehend why setting reflow causes the image to become huge just by rotating it 3 degrees? The page height is constant, and the width is the same.

That is odd indeed. I think it should rather do the opposite, to fit the image into the available space... We probably need to fix something first, before thinking further about changing the default.

I think https://github.com/typst/typst/pull/4462 somehow regressed this because it looks fine in 0.11.1.

Andrew15-5:
> I think #4462 somehow regressed this because it looks fine in 0.11.1.

Really? I tried the last code snippet, and it's the exact same in 0.11.1 and 0.11.0.

```
 typst-0.11.1 -V
typst 0.11.1 (786457e3)

 typst-0.11.0 -V
typst 0.11.0 (2bf9f95d)
```

laurmaedje:
> Yeah, that would make more sense. But generally speaking, I don't know what should be the difference between reflow on and off. Maybe only if the document width/height is auto?

I agree that there shouldn't be any difference here. The fixes for `reflow: true` within growable containers seems to have broken `reflow: true` in ungrowable containers (like the page).

> Really? I tried the last code snippet, and it's the exact same in 0.11.1 and 0.11.0.

0.11.1 and 0.11.0 should look the same. I meant than 0.12.0 looks different from 0.11.1.

Andrew15-5:
> The fixes for `reflow: true` within growable containers seems to have broken `reflow: true` in ungrowable containers (like the page).

Which means that there were too few tests! :)

> 0.11.1 and 0.11.0 should look the same. I meant than 0.12.0 looks different from 0.11.1.

I meant I tried all 3 versions and they are all the same. I'm talking about https://github.com/typst/typst/pull/5330#issuecomment-2449809323. With and without reflow set rule.

laurmaedje:
It gets different when you add `set page(width: 120pt)` (which the test runner sets)

Andrew15-5:
OK, now I see it. So, in v11, the image was smaller and smaller the more it's from 0 and 90 degrees. In v12, the image is bigger and bigger instead. But neither are correct and as you said, it should be contained, i.e., the most outer parts (corners) of the image should always touch the same bounding box at any degree.

And when reflow is off, everything works the same and as I would expect.

laurmaedje:
Hmmm, so it appears that neither 0.11.1 nor 0.12.0 do a particularly great job here. In any case, I feel like `reflow: true` is not really ready for prime time, so I'll close this. We should probably have an issue about improving `reflow`, but I'm not sure it's easily possible.

Andrew15-5:
> but I'm not sure it's easily possible.

Why?



> In any case, I feel like `reflow: true` is not really ready for prime time, so I'll close this.

Can't we just set this PR as blocked by other issue? Or is it better to make another PR when the issue is resolved?

laurmaedje:
> Why?

Because predicting how an element sizes itself if you change the available space is difficult, which makes it hard to reliably define the smaller regions.

> Can't we just set this PR as blocked by other issue? Or is it better to make another PR when the issue is resolved?

If it is resolved and we are confident that it is well behaved, we can just reopen it. But I am in the current state of things not convinced about changing the default.

Andrew15-5:
> Because predicting how an element sizes itself if you change the available space is difficult, which makes it hard to reliably define the smaller regions.

Oh, I was thinking of the constant available space. For automatic sizing, it indeed can be tricky.

PR #5333, opened 2024-10-30
More consistent `self-update` assets determination

Coekjan:
## Description

Typst v0.12 has prebuilt `aarch64-pc-windows-msvc` and `riscv64gc-unknown-linux-gnu` (#4821), but current `self-update` is not aware of the new targets.

https://github.com/typst/typst/blob/2634a8402cdbd2443edf27abeb4ee6bafb70f5bd/crates/typst-cli/src/update.rs#L208-L220

This PR not only fixes it, but also provides a more maintainable implementation.

## Questions

I see that we have prebuilt `x86_64-unknown-linux-musl`, `aarch64-unknown-linux-musl` and `armv7-unknown-linux-musleabi`. They are all using musl libc, but `riscv64gc-unknown-linux-gnu` is using gnu libc alone. Is it better to use `riscv64gc-unknown-linux-musl` for consistency (rust 1.82 supports it now)?

### DIFF
--- a/crates/typst-cli/src/update.rs
+++ b/crates/typst-cli/src/update.rs
@@ -17,6 +17,28 @@ use crate::download::{self, PrintDownload};
 const TYPST_GITHUB_ORG: &str = "typst";
 const TYPST_REPO: &str = "typst";
 
+/// Determine the asset to download based on the target platform.
+///
+/// See `.github/workflows/release.yml` for the list of prebuilt assets.
+macro_rules! determine_asset {
+    () => {
+        // For some platforms, only some targets are prebuilt in the release.
+        determine_asset!(__impl: {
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

If musl is supported now, maybe we can indeed switch.

laurmaedje:
I forgot to bump Rust in the release workflow alongside CI (also in the flake, but I can't test that anyway), so that'd need to be done as well for musl to work.

Coekjan:
> I forgot to bump Rust in the release workflow alongside CI (also in the flake, but I can't test that anyway), so that'd need to be done as well for musl to work.

See #5340 . I have sent a new PR for `riscv64gc-unknown-linux-musl` and `release.yml` :)

PR #5340, opened 2024-10-31
Bump rust in `release.yml`

Coekjan:
`riscv64gc-unknown-linux-musl` is not supported since rust 1.82.0. ~So we can switch to it~.

See #5333 and https://github.com/typst/typst/pull/5333#issuecomment-2449677701 .

NOTE: I don't use flake, so I did not bump rust version in the flake. I left an issue for it, see #5341

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,7 +43,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@1.81.0
+    - uses: dtolnay/rust-toolchain@1.82.0
       with:
         target: ${{ matrix.target }}
 


laurmaedje:
The release action fails with these changes: https://github.com/laurmaedje/typst/actions/runs/11612330587/job/32335578874

Coekjan:
> The release action fails with these changes: https://github.com/laurmaedje/typst/actions/runs/11612330587/job/32335578874

I see... This is because `cross-rs` does not provide an image for `riscv64-unknown-linux-musl`. If we want to use `cross-rs` to build our artifacts, we might need to file issues to the upstream and wait for a next release. I think, however, we can also cross-build the targets **without `cross-rs`**, by providing proper linker and rustc flags for cargo.

Coekjan:
Some investigation:
1. I tried to manually download and specify the musl linker, it succeeded to compile for x86_64, armv7 and aarch64, but failed to compile for riscv64. See https://github.com/Coekjan/typst/actions/runs/11624761042
2. It is because the prebuilt riscv-musl-gcc on https://musl.cc is outdated. Current riscv asm in openssl may require higher version of gcc (https://github.com/openssl/openssl/issues/23011), so the outdated gcc cannot parse the asm instruction (with `vlenb` register).

Some options if we want to use musl for riscv64:
1. build musl using newer gcc on ourselves, in release CI or in a seperate repo (specially for building docker image)
2. make a small patch to openssl (change the register name into register index)
3. ...

laurmaedje:
Is it worth it or should we just stay with the gnu version?

Coekjan:
> Is it worth it or should we just stay with the gnu version?

I think it's up to you.

Using GNU libc specifically for riscv64 makes me feel misaligned with other targets. Current investigation shows that if we want to make them aligned, we may need more engineering efforts.

If you decide to stay with the gnu version, I will clean the commits and just bump rust version in the PR.

Coekjan:
If our principle of releasing prebuilt artifacts is making the binaries small, static-linked (and other advantages of using musl libc), I think it worth to switch to musl.

laurmaedje:
Then, maybe let's go for the "make an issue and wait a bit" route.

laurmaedje:
Thanks for the investigation!

PR #5349, opened 2024-11-01
Allow triggering the release workflow manually

Coekjan:
Before this PR, maintainers have to create a release to test the release workflow (e.g. https://github.com/laurmaedje/typst/actions/runs/11612330587). And with this PR, maintainers can test the release workflow by manually dispatching the workflow.

See https://github.com/Coekjan/typst/actions/runs/11628265117

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,7 @@
 
 name: Build Release Binaries
 on:
+  workflow_dispatch:
   release:
     types: [published]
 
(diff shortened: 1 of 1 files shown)


laurmaedje:
I see how this can be nice, but at the same time, I would probably still create a demo release to test that it actually works in complete integration. But it could maybe still be useful during development of PRs that change release builds.

Coekjan:
> I see how this can be nice, but at the same time, I would probably still create a demo release to test that it actually works in complete integration. But it could maybe still be useful during development of PRs that change release builds.

Yes. I think this can provide a more convenient way to check if the release workflow can correctly build the targets.

laurmaedje:
Okay. Thanks!

Issue #5354, opened 2024-11-02
Empty footnote entry show rule leads to infinite loop


EpicEricEE:
### Description

When starting to type a show rule for footnote entries, the rule is initially empty
```typ
#show footnote.entry: it => {}

#lorem(10) #footnote[A footnote]
```
This empty show rule leads to some kind of infinite loop and never ending compilation. To continue writing, one needs to refresh the page (after making sure that the rule returns something). The discord bot eventually timeouts.

### Operating system

Web app, Windows

### Typst version

- [X] I am using the latest version of Typst

wznmickey:
In Composer::footnote, it will generate a empty frame which will make this element be queued for the next page and in next page it will still be queued for the next and next page and will never stop.

laurmaedje:
The empty frame check is hacky at best, but unfortunately we don't have a better way to check that yet. But in this case, we should check that the first frame is empty AND that there is at least one more frame.

PR #5364, opened 2024-11-04
Fix unstoppable empty footnote loop (#5354)

wznmickey:
Fixes https://github.com/typst/typst/issues/5354

Follow https://github.com/typst/typst/issues/5354#issuecomment-2453364239, a `exist_non_empty_frame` is used to check if there are any non-empty frames. It is checked before `first` to make sure do not have reference conflicts with this iterator.


crates/typst-layout/src/flow/compose.rs
```diff
@@ -431,6 +431,15 @@ impl<'a, 'b> Composer<'a, 'b, '_, '_> {
...
         let nested = find_in_frames::<FootnoteElem>(&frames);
 
+        // Check if there are any non-empty frames.
+        let mut exist_non_empty_frame = false;
+        for i in &frames {
+            if !i.is_empty() {
+                exist_non_empty_frame = true;
+                break;
+            }
+        }
```
Coekjan, line 441:
```suggestion
        let exist_non_empty_frame = frames.iter().any(|i| !i.is_empty());
```

wznmickey: Changed

laurmaedje:
It would be great if you could also add a regression test in `tests/suite/layout/flow/footnote.typ` for this.

wznmickey:
@laurmaedje Added.

laurmaedje:
Thanks!

PR #5371, opened 2024-11-04
Add compare argument for array.sorted

### LINKED ISSUE #5365: Add `compare` argument to `array.sorted`, as an alternative to `key`
### Description

As an alternative to `key`, `array.sorted` should accept a `compare` argument, which would be a function accepting two array elements and returning an integer indicating how they compare. This is similar to C's `sort` function.

### Use Case

Assigning a sortable value to each array element is not always easy. Sometimes, it may be more natural to provide a comparison function.

bluebear94:
Fixes #5365.

I used `compare` for the argument name instead of `by` because `by` could plausibly refer to either the key or the comparison function.

According to the [Rust standard library documentation](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.sort_by), sorting might panic when a non-total ordering function is provided. I’ve been unable to produce a panic so far, but I think this is an important issue to resolve. We could `catch_unwind` such a panic, but this doesn’t solve the problem when `panic=abort` is turned on.

### DIFF
--- a/crates/typst-library/src/foundations/array.rs
+++ b/crates/typst-library/src/foundations/array.rs
@@ -814,7 +814,7 @@ impl Array {
     /// function. The sorting algorithm used is stable.
     ///
     /// Returns an error if two values could not be compared or if the key
-    /// function (if given) yields an error.
+    /// or comparison function (if given) yields an error.
     ///
     /// To sort according to multiple criteria at once, e.g. in case of equality
     /// between some criteria, the key function can return an array. The results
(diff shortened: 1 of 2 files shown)


laurmaedje:
catch_unwind also doesn't work on wasm, so it's kind of a non-starter :/

laurmaedje:
How shall we proceed here?

bluebear94:
Honestly, I’m not sure and that’s why it’s taken a long time for me to comment back. I think this addition isn’t worth the possible downsides, at least until we get more info on how limiting the current `key` system is. At the same time, as @MDLC01 mentioned, some comparison functions are hard to express as keys right now (such as sorting by A descending and then B ascending, or even sorting strings in descending order).

MDLC01:
In cases where a comparison function is provided, maybe we could simply implement our own sorting algorithm? After all, the cost of calling the function probably outweighs the time it takes to sort the array, even if we do not do it with an algorithm as optimized as the one used by Rust.

laurmaedje:
I guess we could do that.

laurmaedje:
@bluebear94 I'd be on board with having a simple custom sorting implementation for that case. If you want to implement that, feel free to. Otherwise, I think we can close the PR.

PR #5372, opened 2024-11-04
Add `sym.eq.triple.not`

jbirnick:
Adds `sym.eq.triple.not` symbol. (`≢`)

(Before it was reachable only under `sym.equiv.not`.)


laurmaedje:
Thanks! 

PS: To be in the loop about symbol naming discussions, there is a [forge thread on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306).

PR #5373, opened 2024-11-04
`Packed::<FigureElem>::outline`: Use provided styles for displaying entry

### LINKED ISSUE #5370: Figure caption in outline doesn't follow the locale-specific `figure.caption.separator`
### Description

I tried making a table of figures (ToF):

```typ
#set text(lang: "ru")
#outline(title: none, target: selector(figure).before(<alt>))
#figure(rect(), caption: [A rectangle])
#figure(table[cell], caption: [A table])
```

image

But for some reason `outline.entry.body` contains a `:` which is used as `figure.caption.separator` for default `text.lang` & `text.region`.

I think this is an oversight and the separator must follow the `figure.caption.separator`. I also think that it would be nice (for customizability) to be able to change it via:

```typ
#show outline: it => {
  show figure.where(kind: table): set figure.caption(separator: "custom")
  it
}
```

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

bluebear94:
Fixes #5370.


laurmaedje:
In principle, it is intentional that we use `StyleChain::default()` instead of `styles`. Since `outline` is only called on queried elements, all the fields are already materialized and the `styles` won't be used anyway. The `styles` passed to the outline are only intended for displaying the counter. (All of this is not _great,_ but that's how it's intended anyway.)

We do call `elem.push_separator` in `Synthesize for FigureCaption`, so it should also be populated in a language-aware manner, but something seems to be wrong there.

laurmaedje:
Closing in favor of https://github.com/typst/typst/pull/5550. Thanks still!

Issue #1239, opened 2023-05-17
embedded SVG text should be selectable. 


taylorh140:
### Description

I notice that when an svg is embedded. It is a vector graphic but all the text that was in the original is no longer selectable. (e.g. I cannot copy text from it). This inserts a clear distinction between typst raw output vs typst -> svg -> typst output. but mostly its a little inconvenient.

### Use Case

For functional documents copy and paste reduces translation errors. This is part of the reason that embedding a svg/pdf is preferable to raster formats. Allowing text to be selectable from embedded vector graphics would make the svg embedding more useful.

hcsch:
I agree that selectable text in embedded vector images would be nice, but if that is added it would be great to make that an opt-in/opt-out feature. Selectable text in a diagram with paragraphs wrapped around it would for example mess up the paragraph's text when trying to copy just that, so making some graphics like diagrams have non-selectable text, and other graphics like external tables have selectable text would be desirable, I think.

mhdadk:
Is there any update on this? This would be a really useful feature to have.

LaurenzV:
Unfortunately not, I've been wanting to work on this but it does take some effort and I haven't had the time. :/

mhdadk:
> Unfortunately not, I've been wanting to work on this but it does take some effort and I haven't had the time. :/

No problem at all. Typst is already incredible as it is, and I'm very grateful for your work so far. This is more of a "nice-to-have" at the moment for me.

LaurenzV:
Happy to say that that text preservation has been implemented upstream, so the odds are good that we can have this in Typst as well by the next release.

taylorh140:
Just checked if this has changed in 0.11.1? 
 - It doesn't seem to have. typst 0.11.1 (50115102)
Keep open.

laurmaedje:
Yeah, 0.11.1 was just a patch release. This will be part of the next proper release, 0.12.

UtsavDayal:
During the latest release, this has eradicated the only simple way of producing unselectable text.  @hcsch is right. There are instances where this functionality needs to be toggled off. Is there a plan to make it configurable? This will also address: https://github.com/typst/typst/issues/2249

LaurenzV:
I opened a PR: https://github.com/typst/typst/pull/5390

PR #5390, opened 2024-11-07
Add support for converting text in SVGs to paths

LaurenzV:


LaurenzV:
Note: In the future, we could consider instead making it a parameter on text so that you can do `#set text(outlined: true)`, and it will automatically apply to SVG figures too. Once we have that, we could deprecate this parameter. But since this probably won't be added in the near future, this is probably the best option for now.

MDLC01:
I don't really like the name `outlined`. It looks like it has something to do with the `outline` element, and other `outlined` parameters such as on figures, images, etc..

LaurenzV:
The only alternatives I can think of are `selectable` or `embed-text`.

MDLC01:
I like `embed-text` better.

laurmaedje:
Do we really want to expose this? I don't really see why people would want this _except_ for the use case of creating unselectable text in Typst generally, for which this is merely a bad workaround.

LaurenzV:
I mean, five people upvoted the comment in #1239, so I guess there is some demand at least? But I'm fine with either.

laurmaedje:
I wonder whether the use case in the upvoted comment could be solved in some way via PDF tagging. Maybe not.

LaurenzV:
> I wonder whether the use case in the upvoted comment could be solved in some way via PDF tagging. Maybe not.

If it does, it probably would only work in Acrobat anyway. 😄

laurmaedje:
I'm okay with this, but don't like the name `embed-text`. I think it's not fully clear what false vs true means with this name (embed could just as well mean to "embed" it as _real_ text). How about `flatten-text`?

laurmaedje:
Thanks!

PR #5391, opened 2024-11-08
Add `asymp` and `asymp.not` symbols

quachpas:
See https://forum.typst.app/t/symbol-for-hardys-asymptotic-notation/1601/2.

The symbols ≍ and ≭ are in Unicode "Mathematical Operators 2200-22FF" as
EQUIVALENT TO and NOT EQUIVALENT TO respectively.

In practice, they are used in the literature for asymptotic behaviour,
providing a shorthand notation for the big-O notation, see https://en.wikipedia.org/wiki/Big_O_notation#:~:text=of%20his%20tract-,Hardy,-defined%20the%20symbol.

### DIFF
--- a/crates/typst-library/src/symbols/sym.rs
+++ b/crates/typst-library/src/symbols/sym.rs
@@ -246,8 +246,6 @@ pub(crate) const SYM: &[(&str, Symbol)] = typst_macros::symbols! {
         triple: '≡',
         triple.not: '≢',
         quad: '≣',
-        equiv: '≍',
-        nequiv: '≭',
     ],
     gt: [
         '>',
@@ -340,6 +338,10 @@ pub(crate) const SYM: &[(&str, Symbol)] = typst_macros::symbols! {
     prop: '∝',
     original: '⊶',
     image: '⊷',
+    asymp: [
+        '≍',
+        not: '≭',
+    ],
 
     // Set theory.
     emptyset: [
--- a/crates/typst-library/src/symbols/sym.rs
+++ b/crates/typst-library/src/symbols/sym.rs
@@ -246,6 +246,8 @@ pub(crate) const SYM: &[(&str, Symbol)] = typst_macros::symbols! {
         triple: '≡',
         triple.not: '≢',
         quad: '≣',
+        equiv: '≍',
+        nequiv: '≭',
     ],
     gt: [
         '>',


crates/typst-library/src/symbols/sym.rs
```diff
@@ -246,6 +246,8 @@ pub(crate) const SYM: &[(&str, Symbol)] = typst_macros::symbols! {
         triple: '≡',
         triple.not: '≢',
         quad: '≣',
+        equiv: '≍',
+        nequiv: '≭',
```
ssotoen, line 250:
```suggestion
        equiv.not: '≭',
```

aaron-jack-manning:
Thanks for the PR. Just a suggestion: it would seem that `eq.equiv.not` would be more consistent with the existing naming rules for symbols, such as `eq.not`, `tilde.not`, `in.not`, etc.

quachpas:
I was wondering about that. I looked through the symbols, and observed that `neq` and `nequiv` are always used instead of `eq.not` and `equiv.not`. I think it's more coherent to keep that up. A single instance of `equiv.not` exists in `tilde`, but also defines the shorthand `nequiv`.

MDLC01:
Hi! Thanks for the PR. We were already considering adding this symbol, but we think the names `asymp` and `asymp.not` would be better. If you want to discuss it with us, it will be easier [on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1304370417653256224). If you don't have Discord, I can provide a summary of the arguments here.

If we instead decide to go for your suggestion (i.e., `eq.equiv` instead of `asymp`), ≭ should be `eq.equiv.not` instead of `eq.nequiv`. When `neq` or `nequiv` are used, it is part of a larger symbol where only the `eq` or `equiv` part is negated.

quachpas:
Thanks for the details! 
> When neq or nequiv are used, it is part of a larger symbol where only the eq or equiv part is negated.
Did not realise that subtlety, is this documented explictly somewhere?

MDLC01:
I don't think this is documented anywhere outside of the symbol list, where it's not stated explicitly (but you can stare at the symbols and names until the logic appears to you).

There are few rules that are set in stone regarding symbol naming (because Unicode would always hits us with an exception). Instead, we try to make everything as consistent as possible, while keeping short names for commonly used symbols.

laurmaedje:
Looks like there is consensus from the Discord discussion!

Issue #5360, opened 2024-11-03
Weird and unnecessary hyphenation in tables


SLUCHABLUB:
### Description

I was trying to make a table where one of the headers was the word "Formal". However when I compiled, it added a hyphen and line break. Like this:

| For-<br>mal | 
| --- |

Even though there was enough space to make the table bigger. I played around with it and even if the table is just one cell this still happens. I should note however that this only occurs with the set rule:

```typst
#set par(justify: true)
```

I'm guessing that it has to do with the width of the word, I haven't tested that many, most work just fine, but "Exrmal" also breaks.

Here is my minimal example:

```typst
#set par(justify: true)

#table(columns: 1, [Formal])
```

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

wznmickey:
Seems that the width calculation have slight floating point calculation error which would cause huge cost.
In my test, the region width is 31.591999999999995pt and the width needed is 31.592pt.

wznmickey:
Cause in `grow` in `typst-layout/src/pad.rs`, (s + p.abs) / (1.0 - p.rel.get()) is `(31.592pt + 10.0pt)/(1.0-0.0)`. It should be `41.592pt` but the given result is `41.59199999999999pt`.

wznmickey:
Not sure how to solve this floating problem, so add a manual check instead.

Enivex:
> Cause in `grow` in `typst-layout/src/pad.rs`, (s + p.abs) / (1.0 - p.rel.get()) is `(31.592pt + 10.0pt)/(1.0-0.0)`. It should be `41.592pt` but the given result is `41.59199999999999pt`.

41.592 is not representable as a float. Not sure what you mean by "should be".

wznmickey:
31.592+10 is 41.592. But due to floating error, it became a little less than 41.592.

PR #5394, opened 2024-11-08
Fix unnecessary hyphenation

wznmickey:
Fixes https://github.com/typst/typst/issues/5360
By manually set slight difference as 0.
The root reason is in grow in typst-layout/src/pad.rs, (s + p.abs) / (1.0 - p.rel.get()) is (31.592pt + 10.0pt)/(1.0-0.0). It should be 41.592pt but the given result is 41.59199999999999pt.

https://github.com/typst/typst/issues/5360#issuecomment-2465876365 but not sure how to solve this kind of calculation error.


crates/typst-layout/src/inline/linebreak.rs
```diff
@@ -543,7 +543,14 @@ fn raw_ratio(
 ) -> f64 {
     // Determine how much the line's spaces would need to be stretched
     // to make it the desired width.
-    let delta = available_width - line_width;
+    let delta = {
+        let delta = available_width - line_width;
+        if delta.abs().to_pt() < 0.00001 {
```
laurmaedje, line 548: `delta.approx_eq(Abs::zero())` would be a little cleaner

wznmickey: Changed.

wznmickey:
done

laurmaedje:
I didn't catch this before, but I find the version in your first commit (with mutable let) a fair bit more readable actually.

wznmickey:
Yes. But it will import a mutable `delta`.

laurmaedje:
The function isn't that long, I don't think it matters much.

wznmickey:
Changed to the first version.

laurmaedje:
Thanks!

Issue #1417, opened 2023-06-05
Eval for text in SVG


Enivex:
### Description

I'd like to have the option run eval() on text in an SVG.

### Use Case

This would be useful for instance for including math in figures, with a look consistent with the rest of the document.

jfab20:
This feature would be awesome! It would enable the use of Inkscape hand drawn figures on typst which is a feature latex does have. For example this figure
![image](https://github.com/typst/typst/assets/81712003/a54aeb71-be16-48da-acbc-a1c60b4a8451)

would get rendered as this
![image](https://github.com/typst/typst/assets/81712003/c7ff3f55-d645-483e-b727-800153970ec3)

Hopefully it can get added soon! If anyone knows of some plugin that does this, that would be cool.

Atreyu-94:
You can see here: https://github.com/textext/textext/issues/403 :)

jfab20:
> You can see here:
> [textext/textext#403](https://github.com/textext/textext/issues/403) :)

Awesome tool! Thanks very much. However I think the capability of using eval in svg files would be cool and very useful nonetheless.

LaurenzV:
I'm giving this a shot and already have some promising results, so it would be great if you could share some example SVGs you would want to embed, so I can test it!

aaeberharter:
What we actually need is a WYSIWYG editor for typst drawing functionality. Maybe something which also strongly interacts with code would be good.

Enivex:
> What we actually need is a WYSIWYG editor for typst drawing functionality. Maybe something which also strongly interacts with code would be good.

Not really realistic before it's more mature and stable.

jfab20:
> I'm giving this a shot and already have some promising results, so it would be great if you could share some example SVGs you would want to embed, so I can test it! :)

Thanks! Here is an example of an svg [genus2.zip](https://github.com/user-attachments/files/16784596/genus2.zip).

Just for the record, I have been trying the tool https://github.com/textext/textext  and I have completely switched to it, I think it is the better way of creating svg figures with math support. You can control the size and the exact position of your text. This is better tha, for example,  adding text and letting latex handle it, as you don't have much control over your text. Thanks to @Atreyu-94 for suggesting this tool!

LaurenzV:
Thanks! No worries, I realized that, unfortunately, it would take some more work to have this work from within Typst, so good to know that there is a good alternative in the meanwhile.

LaurenzV:
[...]

PR #5396, opened 2024-11-09
Add support for evaluating Typst markup in SVG

LaurenzV:
Fixes https://github.com/typst/typst/issues/1417.

If we decide that we want this, I will add documentation to the `eval` property and also add a test case.

Here is an example:
```svg
#set text(font: "Noto Sans", size: 17pt)

#image("im1.svg", width: 300pt, eval: true)

#set text(font: "Noto Sans", size: 11pt)
#image("im2.svg", eval: true)
#[
    #set text(size: 8pt, fill: gradient.linear(..color.map.rainbow))
    #image("im2.svg", width: 300pt, eval: true)
]
#image("im3.svg", eval: true)

#set text(fill: green)
#image("im4.svg", eval: true)

```

Output:
typst

### DIFF
--- a/crates/typst-library/src/visualize/image/mod.rs
+++ b/crates/typst-library/src/visualize/image/mod.rs
@@ -97,7 +97,7 @@ pub struct ImageElem {
 
     /// TODO: Add docs
     #[default(false)]
-    pub eval: bool
+    pub eval: bool,
 }
 
 #[scope]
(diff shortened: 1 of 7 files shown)


crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -94,6 +94,10 @@ pub struct ImageElem {
     /// ```
     #[default(ImageFit::Cover)]
     pub fit: ImageFit,
+
+    /// TODO: Add docs
+    #[default(false)]
+    pub eval: bool,
```
istudyatuni, line 100: What do you think about renaming it to eval-text?

LaurenzV: I like the current shorter name better, but in principle I would be fine with changing it.

laurmaedje:
I'll bring this up in our next team meeting to discuss it with the others.

PR #5407, opened 2024-11-12
Update RustyBuzz dependency (and adjacent crates)

alerque:
I set out to make sure bumping the RustyBuzz dependency in Typst didn't break anything. It turns out hard to do without ending up with multiple versions of dependencies and conflicting types, but I think I have it all sorted now.

The bigest catch is resvg is going to need a release too. I submitted a PR which is already merged, so now it's just a matter of waiting for a release. The other dependencies that need releases are in the @typst org, and relevant PRs are pending.

- [ ] resvg: https://github.com/linebender/resvg/pull/846
- [ ] pixglyph: https://github.com/typst/pixglyph/pull/2
- [ ] svg2pdf: https://github.com/typst/svg2pdf/pull/82

Once all the above PRs are available in release tags I can update this draft with the final version numbers to remove the Git dependencies. Until then I gueses it is a PoC unless you want to use Git dependencies for resvg and others during the development cycle.


laurmaedje:
>  unless you want to use Git dependencies for resvg and others during the development cycle

I typically only do that for dependencies which we control as to not get blocked from making a release of Typst because we're waiting for someone.

alerque:
> > unless you want to use Git dependencies for resvg and others during the development cycle
> 
> I typically only do that for dependencies which we control as to not get blocked from making a release of Typst because we're waiting for someone.

That's fair enough.

I updated this PR with the tagged pixgraph release. Now it will have to wait on resvg to release, at which point the svg2pdf PR can merged and that crate released, then finally this one can be as well.

PR #5408, opened 2024-11-12
Various IDE improvements

laurmaedje:
### User-facing
- Added autocompletion for file paths, including filtering depending on which function gets the path
- Autocompletions in typed positions (e.g. parameters) are filtered more smartly
- Different functions are now autocompleted with different bracket modes, e.g. `box` gets round parens, `footnote` will get square brackets, and `figure` will get round parens + newline
- Added image and table snippets for figure bodies
- Added tooltip for star import ("This star imports ...")
- Fixed autocompletion for `#cite`

### API
- Added `IdeWorld` trait to separate out additional functions for IDE tooling from core `World` functions. `World::packages` is moved there. `IdeWorld::files` is new.
- Added additional `CompletionKind`s
- Simplified go-to-definition API to make it easier to understand and removed unused things (also changed `deref_target` a bit)
- Added support for raw range `Spans`, which can encode ranges outside of Typst files. Will also be useful to report errors directly in bib files (though the biblatex crate's error ranges aren't the greatest)
- Cleaned up the typst-eval `import` API (which is used by typst-ide)

### Misc
- PicoStr and FileId are now null-optimized
- Fixed some wasm-related problems (wasmi bump & typst-timing warnings)
- Better IDE tests

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -6,8 +6,8 @@ use ecow::{eco_format, EcoString};
 use if_chain::if_chain;
 use serde::{Deserialize, Serialize};
 use typst::foundations::{
-    fields_on, format_str, repr, AutoValue, CastInfo, Func, Label, NoneValue, ParamInfo,
-    Repr, Scope, StyleChain, Styles, Type, Value,
+    fields_on, repr, AutoValue, CastInfo, Func, Label, NoneValue, ParamInfo, Repr, Scope,
+    StyleChain, Styles, Type, Value,
 };
 use typst::model::Document;
 use typst::syntax::ast::AstNode;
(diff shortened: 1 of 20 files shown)


crates/typst-ide/src/complete.rs
```diff
@@ -773,25 +801,52 @@ fn named_param_value_completions<'a>(
...
+        ctx.file_completions_with_extensions(match func.name() {
+            Some("image") => &["png", "jpg", "jpeg", "gif", "svg", "svgz"],
+            Some("csv") => &["csv"],
+            Some("plugin") => &["wasm"],
+            Some("cbor") => &["cbor"],
+            Some("json") => &["json"],
+            Some("toml") => &["toml"],
+            Some("xml") => &["xml"],
+            Some("yaml") => &["yaml"],
+            Some("bibliography") => &["bib", "yaml"],
```
EpicEricEE, line 829: Maybe also complete `.yml` files? I know it's not the "intended" extension, but it's still commonly used

PR #5411, opened 2024-11-13
Timings for `state.at` and `state.get`

Dherse:
I found that these timings were useful when optimizing [Dherse/codly](https://github.com/Dherse/codly) and I figured they could be interesting for others. It barely has an impact on `--timings` compilation time. It's useful for finding closures that are slowed down by memoization checking (something discussed on Discord [here](https://discord.com/channels/1054443721975922748/1088371867913572452/1306037948906471466)


laurmaedje:
Thanks!

PR #5412, opened 2024-11-13
Added `todo` element

### LINKED ISSUE #814: Add `#todo` tag
Add `#todo` tag which shows warning with line number on compilation. It could also show marks on each page of generated document.

Dherse:
Closes #814 

Adds a new `todo` element, by default it produces a warning and adds a big red bold "TODO" in the body, can optionally disable its warnings and/or change its message (including message shown in body) can be styled using show and set.

Looking for some feedback whether the implementation is liked.

EDIT: @LaurenzV pointed out that it's weird for warnings to get disabled with an argument when it is planned to have warning suppression #4459 and I agree, but temporarily I think this is okay.


crates/typst-library/src/visualize/todo.rs
```diff
@@ -0,0 +1,97 @@
...
+///
+/// ```example
+/// #show todo: it => text(fill: red, size: 16pt, it.message)
+///
+/// #todo(message: "Complete this example.")
+/// ```
+///
+/// # Example
+/// ```example
+/// // Produces a warning and a big bold red "TODO" in the console.
```
MDLC01, line 29: The comments every time seem unnecessary: there will be a preview on the right, right (even though warnings will be emitted)?

Dherse: it's precisely because the preview will not show warnings that I added comments, since there is no way AFAIK to show warnings in the documentation

crates/typst-library/src/visualize/todo.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// A reminder to implement a feature, add content, or fix a bug.
+///
+/// By default this element is shown as a big bold red "TODO", and produces
+/// a warning with an optional message. You can optionally restyle the shown
+/// text by using the `show` rule, and you can disable the warning by using the
+/// `warn` argument.
+///
+/// ```example
+/// #show todo: it => text(fill: red, size: 16pt, it.message)
```
MDLC01, line 22: Isn't that essentially the default show rule?

crates/typst-library/src/visualize/todo.rs
MDLC01: Then maybe the comments should only mention the warnings?

Otherwise, I would rephrase it as follows:
> Produces a warning in the console and a big bold red "TODO".

jbirnick:
Idk if this should be part of the core functionality of Typst. Wouldn't it be better to have a `warning` function, and then a _package_ does the Todo-thing?

xkevio:
> May I ask why you want this to be part of the core functionality of Typst? At first sight, wouldn't it be better to have a `warning` function, and then a _package_ does the Todo-thing that you want?

I also fail to see the need to have this built-in if it's just simple styled text with a warning attached to it. Especially since Typst functions are so easily composable that it is trivial to make your own todo-functions that produce the same output without now requiring a show rule. 

The warning part is useful but as was mentioned already, this is something that is planned already as-is with warning suppression alongside it.

laurmaedje:
The [issue](https://github.com/typst/typst/issues/814) did have 18 upvotes and I didn't see any harm in it, which is why I proposed this as a potential thing to work on for @Dherse. 

A built-in todo standard element could also have better tooling integration (like being collected in a special list in the web app, like the outline panel), though that could just as well be done with a `// TODO` comment, which would maybe work better in code. I'm still open to it I think, but I also see some potential downsides now.

laurmaedje:
I think the code thing is maybe a bit of a problem since we can't really fit both purposes (unimplemented and not yet written) into one construct. 

And maybe a warning, while not bad, is not the _ideal_ level of diagnostic, it could shadow other relevant warnings. In the web app, I'd rather expect a separate category in the improve panel. (Which would be easy to realize with just a detection of `// TODO` comments....)

MDLC01:
Maybe a similar section could list uses of the `lorem` function, as this is a common way to defer writing something.

jbirnick:
You could also make an official Todo package which emits a certain metadata, which is then (together with the comments) collected in the panel. Then other Todo packages could emit the same metadata so they are also displayed.

mattfbacon:
As a native English speaker "big red bold" sounds better to me than "big bold red"

swaits:
FWIW, I just use this (which is basically a one-liner):

```
#let todo(body) = {
  highlight(text(weight: "regular", "TODO: ") + body)
}
```

Which I keep in a local package. I think this bolsters the question about package versus core.

Dherse:
@mattfbacon I am not a native English speaker, so thank you for your input, it's fixed :)

swaits:
I'd definitely want this to be a model so I can control its output.

But I want to +1 this because the tooling it enables is valuable. 

It'd be nice to easily number and list them as well. In LaTeX I usually keep an outstanding todo table at the top of my wip docs.

jbirnick:
But don't you think it kills competition for TODO packages? There are different TODO packages out there (and more to come) which each work a bit differently in terms of how they display the TODOs and that you can list them and so on.
I think one should try to avoid to kill all these nice future ideas.

In terms of languages like Rust it's useful because you don't have to worry about return types. Let's say for some trait you have to implement a function which has to return some very specific complicated object, then `todo!()` will make it compile without actually implementing it. Such a problem doesn't exist in Typst, and even it it would exist then it wouldn't be solved by a markup todo item, but by something that panics.

If it enables integration into tooling _and_ the packages will be able to have very fine-grained control of it, then I'm on board though.

swaits:
I think the todo package ecosystem can coexist. As an example, look at `codly`, which simply modifies `raw`. If todo is a model, it's easy for anyone to customize it.

laurmaedje:
The problem with making it an element is that it doesn't really cater to the "todo in code" use case. Let's say we had a query-powered todo panel, it would only list todos that end up in the document. A `// TODO` comment that is basically just grepped for doesn't suffer from this. On the flip side, such a comment wouldn't show up in the PDF.

laurmaedje:
After further consideration, I'm leaning towards closing this (sorry @Dherse for putting you onto this in the first place 🤕), see my [reasoning in the feature request issue](https://github.com/typst/typst/issues/814#issuecomment-2526322085).

Dherse:
I think that's fair. Feel free to reopen if you change your mind.

PR #5413, opened 2024-11-13
Added warning when explicit return in code (not markup) discards joined content

### LINKED ISSUE #4262: Add warning for unused content due to `return`
### Description

It could be useful to have some code like:

```typ
#let f(x) = {
   counter(heading).update(0)
   return x
}
```

produce a warning on the counter update line, as it is not joined due to `return` and therefore unused.

It could be suppressed either with the upcoming warning suppression feature, or simply with `let _ = counter(heading).update(0)` to indicate you are explicitly discarding it.

The warning could also suggest removing `return`.

### Use Case

It seems like a pretty easy mistake to make for beginners, with a simple solution that is often not very obvious, especially for those coming from languages which exclusively use `return` to return values from functions.

Dherse:
Closes #4262

This adds a warning when `return`-ing in a code block discards content (potentially state updates, etc.)

Also added relevant tests


laurmaedje:
We need to make sure that the return is unconditional, otherwise it might be intended.

Dherse:
> We need to make sure that the return is unconditional, otherwise it might be intended.
Ah, that is a bit trickier

Dherse:
Btw this is implemented as mentioned on discord.

crates/typst-eval/src/code.rs
```diff
@@ -54,7 +54,17 @@ fn eval_code<'a>(
 
         output = ops::join(output, value).at(span)?;
 
-        if vm.flow.is_some() {
+        if let Some(event) = &vm.flow {
```
laurmaedje, line 57: if let chains 🥲 (but there is some movement on them again!)

crates/typst-eval/src/code.rs
```diff
@@ -54,7 +54,17 @@ fn eval_code<'a>(
...
+            if matches!(output, Value::Content(_)) {
+                if let FlowEvent::Return(span, Some(_), false) = event {
+                    vm.engine.sink.warn(warning!(
+                        *span,
+                        "unconditional explicit return value discards content";
```
laurmaedje, line 62: how about "this return unconditionally discards the content before it"

crates/typst-eval/src/code.rs
```diff
@@ -54,7 +54,17 @@ fn eval_code<'a>(
...
+                        hint: "use `return` without a value to return the joined content"
```
laurmaedje, line 63: that could be the fix or no return at all. tbh, I think the most common case is returning a _non-content_ value while also trying to update a counter. 

@PgBiel do you have some thoughts on the best hint(s)?

crates/typst-eval/src/flow.rs
```diff
@@ -17,8 +17,8 @@ pub enum FlowEvent {
     /// Skip the remainder of the current iteration in a loop.
     Continue(Span),
     /// Stop execution of a function early, optionally returning an explicit
-    /// value.
-    Return(Span, Option<Value>),
+    /// value. The final boolean indicates whether the return was conditional.
+    Return(Span, Option<Value>, bool),
```
laurmaedje, line 21: This way to detect conditionalness does seem sensible in principle. I'll need to experiment a bit to see whether it could result in false positive. I'll do that after my vacation.

tests/suite/scripting/return.typ
```diff
@@ -85,3 +85,81 @@
...
+}
+
+#test(f(), state("hello").update("world"))
+
+--- return-discard-not-content ---
+// Test that non-joined value is not a warning.
+
+#let f() = {
+  (33, )
+  return (66, )
```
laurmaedje, line 152:
```suggestion
  (33,)
  return (66,)
```

tests/suite/scripting/return.typ
```diff
@@ -85,3 +85,81 @@
...
+
+#let f() = {
+  state("hello").update("world")
+  return
+}
+
+#test(f(), state("hello").update("world"))
+
+--- return-discard-not-content ---
+// Test that non-joined value is not a warning.
```
laurmaedje, line 148: non-joined isn't really accurate, is it? since arrays can be joined?

crates/typst-eval/src/code.rs
PgBiel: yeah, that seems more readable

crates/typst-eval/src/code.rs
PgBiel: actually I didn't even know `return` automatically returns the joined values. :sweat_smile:

For starters, there should be a hint if there is a state update or counter update among the joined content, mentioning that a state update (or counter update / step) needs to be returned and placed in the document to be effective.

Then, another potential hint could be to "try omitting the `return` keyword to also join this value" if it is also content.

PgBiel (commented):
These changes look good! My concerns mostly match Laurenz's.

tests/suite/scripting/return.typ
Dherse: Oops, it should say `non-content joined value`

Dherse:
Ok, I fixed all of the current code review items ;)

PgBiel:
Thanks, I've marked the reviews as resolved for you (feel free to do so in the future if you feel you've satisfied the review comments).

PgBiel:
Yeah that's fair, I'll often do the same. Though it's also mostly OK since we can unresolve things, but yeah :smiley:

laurmaedje:
I tweaked the messages a bit:
- I dropped the "use `return` without a value to return the joined content" hint. I don't think it really makes sense: Since it's always unconditional a bare return would either be unnecessary or unconditionally discard content _after_ itself. So I don't think it really has any use if it's unconditional.
- I moved and reworded the state/update message.

I also replaced the explicit LazyLock with the `singleton` macro, moved the stuff into a function and got rid of some code duplication.

If it looks good to y'all, we can merge.

Dherse:
There's a singleton macro 🤨

But looks good to me!

laurmaedje:
> There's a singleton macro 🤨

Yup and you even left a comment on the PR that introduced it: https://github.com/typst/typst/pull/4794

laurmaedje:
> But looks good to me!

👍

Thanks for the PR

PR #5414, opened 2024-11-13
Convert unopened square-brackets into a hard error

### LINKED ISSUE #411: Unbalanced square brackets should be a hard error
In a markup context, a lone `]` without a `[` is not considered part of the surrounding object.

This code is rendered like so:
```typst
= [ x ]
= x ]
```

This behaviour can be found with list-, enum-, and term-items as well.

Additionally, strong and emphasized text may not contain a closing bracket if no opening bracket preceeds it. The compiler will throw an error in that case.

The offending code can be found here:
- https://github.com/typst/typst/blob/main/src/syntax/parser.rs#L34-L39
- https://github.com/typst/typst/blob/main/src/syntax/parser.rs#L161-L165

See reparsing and the other text objects as well.

wrzian:
Closes #411

I chose to move the nesting logic to the inside of `markup_expr` so that we can share the logic and error generation between the normal markup function and the reparser.

One change I'm not certain about is whether we want to keep the behavior of code blocks stopping early at a right-bracket, as it makes some errors more annoying. (but removing that is likely to make yet _other_ errors more annoying). We could do so by removing `RightBracket` from line 779 in `code_block` if desired.


crates/typst-syntax/src/parser.rs
```diff
@@ -201,7 +205,7 @@ fn equation(p: &mut Parser) {
     let m = p.marker();
     p.enter_modes(LexMode::Math, AtNewline::Continue, |p| {
         p.assert(SyntaxKind::Dollar);
-        math(p, syntax_set!(Dollar, RightBracket, End));
+        math(p, syntax_set!(Dollar, End));
```
wrzian, line 208: This was a left over not-quite-bug from the parser refactor from a bit of careless find-replace. It's not actually a bug because we don't generate `RightBracket` tokens in math mode, but should definitely not be there.

crates/typst-syntax/src/parser.rs
```diff
@@ -47,15 +47,11 @@ fn markup_exprs(p: &mut Parser, mut at_start: bool, stop_set: SyntaxSet) {
...
-            SyntaxKind::RightBracket if nesting > 0 => nesting -= 1,
-            _ if p.at_set(stop_set) => break,
-            _ => {}
-        }
-        markup_expr(p, at_start);
+    let mut at_nested_bracket = false;
+    while !p.at_set(stop_set) || at_nested_bracket {
+        markup_expr(p, at_start, &mut nesting);
         at_start = p.had_newline();
+        at_nested_bracket = nesting > 0 && p.at(SyntaxKind::RightBracket);
```
laurmaedje, line 54: couldn't we remove the variable and directly inline this expression after the `||` in the while loop's condition?

wrzian: Yes we can :+1:

wrzian:
Removed the variable, but left a comment so it's easier to understand in the future.

laurmaedje:
Thanks!

PR #5415, opened 2024-11-14
Added span for external file errors

Dherse:
This adds spans for all errors that occur in files outside of Typst's source files (i.e SVG images, external JSON, etc.)

This required quite a laundry list of changes, including to the test framework most notably:
- allow external path with syntax `// Error: "/path/to/external/file" 3-5 ...`
- allow using raw byte addressing `// Error: "/path/to/external/file" #3-#5` -> this is needed for handling of edge cases where errors occur on line breaks, a bit annoying, **if you find a workaround, let me know**

And then changes in all of the image decoding (for SVG), and all of the loaders (`csv`, `json`, `xml`).

There are likely stuff left to do and discussions to be had, but now I am going to bed.


crates/typst-library/src/loading/csv.rs
```diff
@@ -177,15 +185,25 @@ cast! {
 }
 
 /// Format the user-facing CSV error message.
-fn format_csv_error(err: ::csv::Error, line: usize) -> EcoString {
-    match err.kind() {
-        ::csv::ErrorKind::Utf8 { .. } => "file is not valid utf-8".into(),
+fn format_csv_error(
+    err: ::csv::Error,
+    line: usize,
+    span: Span,
```
PgBiel, line 191: same comment on potentially different variable name as before

crates/typst-library/src/diag.rs
```diff
@@ -551,30 +551,49 @@ impl From<PackageError> for EcoString {
 }
 
 /// Format a user-facing error message for an XML-like file format.
-pub fn format_xml_like_error(format: &str, error: roxmltree::Error) -> EcoString {
-    match error {
+pub fn format_xml_like_error(
+    format: &str,
+    error: roxmltree::Error,
+    span: Span,
```
PgBiel, line 557: same comment on potentially different variable name (not sure if it's 100% applicable to all usages of this function, so it may be disregarded if needed)

PgBiel (commented):
This is a nice change! I've left some initial observations.

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -26,12 +28,36 @@ struct Repr {
...
+            .map_err(|_| format_usvg_error(usvg::Error::NotAnUtf8Str, span, None, ""))?;
+
+        usvg::Tree::from_str(text, &options)
+            .map_err(|err| format_usvg_error(err, span, file, text))
+    } else {
+        let text = std::str::from_utf8(data.as_slice())
+            .map_err(|_| format_usvg_error(usvg::Error::NotAnUtf8Str, span, None, ""))?;
+
+        usvg::Tree::from_str(text, &options)
+            .map_err(|err| format_usvg_error(err, span, file, text))
```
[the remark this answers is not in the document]
PgBiel: Also make sure to fix the clippy complaint regarding `&options`

tests/src/collect.rs
```diff
@@ -314,12 +319,30 @@ impl<'a> Parser<'a> {
...
+            self.s.eat_if(' ');
+        }
+
         let mut range = None;
-        if self.s.at('-') || self.s.at(char::is_numeric) {
-            range = self.parse_range(source);
-            if range.is_none() {
-                self.error("range is malformed");
-                return None;
+        if self.s.at('-') || self.s.at(char::is_numeric) || self.s.at('#') {
```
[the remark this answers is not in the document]
Dherse: This is on purpose, the `#` denotes raw byte position, only useful for external files: some errors (in parsing of JSON and TOML most notably) will give you the position of a newline character, which is not supported by the usual `5-3` syntax, so I added the `#` to denote binary positions in the file. They should never be used for typst files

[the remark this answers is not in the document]
Dherse: But i see what you mean, i'll add an error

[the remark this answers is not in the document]
Dherse: Here's a snipper of what that looks like
```rs
          range = if let Some(file) = file {
              self.parse_range_external(file)
          } else if !self.s.at('#') {
              self.parse_range(source)
          } else {
              self.error("raw byte positions are only allowed in external files");
              return None;
          };
```

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -26,12 +28,36 @@ struct Repr {
     tree: usvg::Tree,
 }
 
+fn svg_parse_inner(
+    data: &Bytes,
+    options: &usvg::Options,
+    span: Span,
+    file: Option<FileId>,
+) -> SourceResult<usvg::Tree> {
+    if data.starts_with(&[0x1f, 0x8b]) {
```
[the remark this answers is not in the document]
Dherse: Good point, I'll add that, this is indeed needed to do proper error handling which is just... unfortunate :( Because for binary SVGs, we don't have text (to generate an error) until it's been decompressed

crates/typst-library/src/visualize/image/svg.rs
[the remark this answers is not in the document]
Dherse: Yep thanks

crates/typst-library/src/diag.rs
Dherse: I renamed all of those spans to `call_span` to make it clearer this is the span of the call site.

tests/src/collect.rs
```diff
@@ -314,12 +319,30 @@ impl<'a> Parser<'a> {
...
 
+        let mut file = None;
+        if self.s.eat_if('"') {
+            let path = self.s.eat_until('"');
+            let vpath = VirtualPath::new(path);
+            file = Some(FileId::new(None, vpath));
+
+            self.s.eat_if('"');
+            self.s.eat_if(' ');
+        }
```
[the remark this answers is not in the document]
Dherse: If you ommit the closing `"`, it will complain that there is no range and give you the proper place where you made the error. Not much more I can do I think without complexifying the code a lot, for an error that only happens when writing tests 🤔

tests/src/run.rs
```diff
@@ -313,39 +313,90 @@ impl<'a> Runner<'a> {
...
+        let text = String::from_utf8_lossy(&bytes);
+
+        let mut line = 1;
+        let mut column = 1;
+        let mut s = Scanner::new(&text);
+        while let Some(c) = s.eat() {
+            if is_newline(c) {
+                line += 1;
+                column = 1;
+            }
```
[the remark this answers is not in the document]
Dherse: Right, forgot it, it's fixed ;)

Dherse:
Ok, I fixed all of the current code review items ;)

tests/src/collect.rs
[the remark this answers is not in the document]
Dherse: Actually my bad, I found a really easy way of adding a nicer error 😓

tests/src/run.rs
```diff
@@ -348,6 +372,45 @@ impl<'a> Runner<'a> {
...
+        while let Some(c) = s.eat() {
+            if is_newline(c) {
+                if was_cr && c == '\n' {
+                    was_cr = false;
+                } else {
+                    was_cr = c == '\r';
+                    line += 1;
+                    column = 1;
+                }
+            }
```
PgBiel, line 399:
```suggestion
            } else {
                was_cr = false;
            }
```

PgBiel (commented):
Thanks, it's all looking much better :)

tests/src/run.rs
Dherse: Thanks!

crates/typst-syntax/src/span.rs
```diff
@@ -99,6 +100,63 @@ impl Span {
...
+            }
+
+            if current_row == end_row && current_col == end_col {
+                end_byte = Some(idx);
+                break;
+            }
+
+            if is_newline(c) {
+                if was_cr && c == '\n' {
+                    was_cr = false;
```
laurmaedje, line 143: can't we instead advance `s` in here and skip the `was_cr` state?

crates/typst-syntax/src/span.rs
```diff
@@ -99,6 +100,63 @@ impl Span {
...
+                    was_cr = c == '\r';
+                    current_row += 1;
+                    current_col = 0;
+                }
+            } else {
+                current_col += 1;
+                was_cr = false;
+            }
+
+            idx += c.len_utf8();
```
laurmaedje, line 154: isn't this just `s.cursor()`?

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -224,13 +224,15 @@ impl Image {
         data: Bytes,
         format: ImageFormat,
         alt: Option<EcoString>,
-    ) -> StrResult<Image> {
+        span: Span,
```
laurmaedje, line 227: As discussed on Discord, we should avoid passing spans and file IDs into memoized functions if they are only used for formatting the error message. Rather, we'd have a return type that is not `SourceResult` and turn it into a `SourceResult` using span & ID outside. (Adding this to the PR discussion for completeness' sake.)

tests/src/run.rs
```diff
@@ -348,6 +372,47 @@ impl<'a> Runner<'a> {
...
+        // We can unwrap since test has run, these paths are correct
+        let path = system_path(id).unwrap();
+        let bytes = read(&path).unwrap();
+        if pos > bytes.len() {
+            return "oob".into();
+        }
+
+        let text = String::from_utf8_lossy(&bytes);
+
+        let mut line = 1;
```
laurmaedje, line 386: This looks very similar to the Span code. The `Source` type can already do it (and faster than this here!) with a `lines` vector. Maybe we could even turn that into a reusable `Lines` type that is not tied to Typst source files, but can also handle its in-place edits! Especially in the context of the web app, we'll also need the UTF-16 conversion methods to be able to make use of these changes.

We could even consider memoizing the output via `lines(bytes: &Bytes) -> Arc<Lines>` (right now many errors in a long binary file would result in technically O(n^2) runtime).

This is a bit of larger thing and I know you're busy, so don't feel forced to work on this. Maybe I'd pick it up otherwise.

tests/src/run.rs
```diff
@@ -313,28 +313,52 @@ impl<'a> Runner<'a> {
...
+            format!(
+                "`{}`",
+                self.test.source.text()[range.clone()]
+                    .replace('\n', "\\n")
+                    .replace('\r', "\\r")
+            )
         } else {
-            format!("`{}`", self.test.source.text()[range.clone()].replace('\n', "\\n"))
+            let path = system_path(file).unwrap();
+            let bytes = read(&path).unwrap();
```
laurmaedje, line 329: Why not use `self.world.file(file)` here (and below)? Then we can revert the changes in world.rs.

Dherse:
@laurmaedje I didn't have time to finish it last week in the end, thanks for the review, I come back from a work trip on Thursday evening so I'll try and get to it maybe on Friday :)

laurmaedje:
Sounds good, no rush

laurmaedje:
I'll close this due to inactivity. No worries about it, I fully understand that you're quite busy! :)

We can always go back and pick this back up or use it as a base for a second attempt.

PR #5416, opened 2024-11-14
Support for named capturing groups in regular expressions

### LINKED ISSUE #5383: Support for named capturing groups in regular expressions
### Description

It appears that Typst currently doesn't support named regular expression groups since the crate used [doesn't support them either](https://discord.com/channels/1054443721975922748/1088371919725793360/1292103396249178133). This was only discussed on Discord so I thought it makes sense to have it as an issue/reminder and also as a reference.

Currently something like

```
#let format(it) = {
  it = it.replace(
    regex("(?<match>abc)"),
    it => "*" + it.match + "*",
  )
}

#format("xyz abc sdf")
```

Has to be written as something like

```
#let format(it) = {
  it.replace(regex("(abc)"), m => {
    let match = m.captures.at(0)
    return "*" + match + "*"
  })
}

#format("xyz abc sdf")
```

### Use Case

Currently named capturing groups are a little less convenient/clear to write. It could also make sense to mention it in the documentation?

wesleyel:
fix #5383

This PR is trying to implement the named group feature for the regex engine.

The named group is a feature that allows the user to name a group in the regex pattern, and then use the name to refer to the group in the code.

For example, the following typst pattern:

```typst
#let format(it) = {
  it = it.replace(
    regex("(?<match>abc)"),
    it => "*" + it.at("match") + "*",
  )
}

#format("xyz abc sdf")
```

## Sulotion A

use `Argument` type to store capture results.

### Pros:

No new type, no breaking change.

### Cons:

- The `Argument` type doesn't support `len()` function, so I have to delete `captures.len()` relevant test in `str.typ`
- Although `Argument` struct supports both index and named access. But they can not be applied on one entry at the same time. That's why this test case uses `index 0` to retrieve the second match.

```typst
#test("hello : world".replace(regex("^(?<first>.+)\s*(:)\s*(?<second>.+)$"), m => {
  upper(m.captures.at("first")) + m.captures.at(0) + " " + upper(m.captures.at("second"))
}), "HELLO : WORLD")
```

-  The `Argument` type's fields cannot be accessed by 'args.field'.

```typst
// Error: 3:20-3:25 cannot access fields on type arguments
#test("hello : world".replace(regex("^(?<first>.+)\s*(:)\s*(?<second>.+)$"), m => {
  upper(m.captures.first)
}), "HELLO ")
```

## Sulotion B

use a extend version of `Dictionary` type to store capture results.

### Pros:

- Easy to impl both index and name access.

### Cons

- A new type needs to be added, or `Dictionary` type needs to be extended.


wesleyel:
I agree that creating a new type just for capture feels like overkill.

But I can't think of a better way. To fix these TODOs, either create a superset of "Argument" or "Dict".

Ants-Aare:
what's the issue with just making the return type a dictionary like this: captures: (pos:("group 0", "group 1", ...), named: (match: "abc", ...)

sure it's a bit more verbose, but writing captures.named.match is not that bad. It can default to returning captures the old way when no named groups are present to make it backwards compatible.

laurmaedje:
The arguments type is further along now, so this could potentially be revisited.

Issue #5189, opened 2024-10-11
Add PDF/A-1b native export


fdekerme:
### Description

From this forum topic: https://forum.typst.app/t/when-will-pdf-a-export-be-available/816/19
Native PDF/A-2b export is available in version 0.12-rc1. However, as noted by some of the comments in this topic, several institutions require PDF/A-1b files for archiving. This is the case, for example, at the Paris-Saclay University or MIT for theses. So I think this is an important feature for Typst. 

### Use Case

Archiving of documents such as university theses.

quachpas:
For UPS, I don't think they required PDF/A-1b strictly. They require >=PDF@1.4.

fdekerme:
That's what UPS says. However, I was only able to submit my thesis manuscript for archiving (https://facile.cines.fr/) after converting it to PDF/1-1b with Adobes Acrobast Pro.

quachpas:
That's strange, PDF/A-2 should be archiveable, it's check marked on their list 😂

Issue #4926, opened 2024-09-09
Text in SVG with filter is rasterized at low resolution


sebaseb98:
### Description

when I want to add a svg, which I created in Inkscape to my document, there occur some errors:

- text gets blurry
- arrows are missing

In the preview everything looks as expected, but when compiling these details somehow get lost

### Reproduction URL

https://typst.app/project/r6AFmbX8lxASSH1fTptN19

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

LaurenzV:
> arrows are missing

This issue is already known and is already fixed in the newest development version, so it will work fine in the next update. See here: https://github.com/typst/typst/issues/3715

> text gets blurry

The reason this is happening is because you have a guassian blur filter that is applied to the text. Since those cannot be represented natively in PDF, Typst instead resorts to rasterizing them, and in this case it unfortunately results in a very low resolution being chosen... Ideally, this should be configurable. With that said, you can avoid this in the first place by not applying a gaussian blur to the text, which you probably wouldn't want to do, anyway. 😄

sebaseb98:
Thank you so much @LaurenzV!

- Arrows are missing ✅
- the blurry text:
For future people finding this: I edited the text of the svg file and removed the filter line in the sections belonging to the text fields. This solved my problem


Is this issue something that will be worked on - I mean technically you closed it by solving my problem but it still would be nice if svgs look the same in the compiled pdf as they look in other viewers.  I leave it up to you to decide :)

LaurenzV:
Yes, let’s keep it up until there is a way to adjust that from Typst.

laurmaedje:
> Yes, let’s keep it up until there is a way to adjust that from Typst.

Do we really want to make that adjustable? That seems like a tad low-level for me. Can't we just increase the resolution to something that works well?

LaurenzV:
> Do we really want to make that adjustable? That seems like a tad low-level for me. Can't we just increase the resolution to something that works well?

I guess that would work too, but we would have to adapt it depending on how big it ends up in the final PDF.

EpicEricEE:
Maybe this can somehow be combined with #2554 to a single parameter for image resolutions, which for SVG then only accounts for the rasterized part?

nineff:
The current rasterization resolution also leads to blurry edges at objects which should otherwise have a clearly defined corner. See the attached files.[svg.pdf](https://github.com/user-attachments/files/17316801/svg.pdf)
[...]

Issue #4901, opened 2024-09-05
Base64 encoded PNG content in SVG plots is not rendered in all PDF viewers


sonovice:
### Description

The following SVG file contains elements such as
```xml
<image id="image6" width="237" height="225" xlink:href="data:image/png;base64,iVBORw0KGg ... ">
```

[36_bin_cqt.svg.zip](https://github.com/user-attachments/files/16884667/36_bin_cqt.svg.zip) (zipped to prevent rendering in browser)

These PNGs are not shown when opening the PDF with Apple's Preview app or other widely used PDF viewers such as Skim or even the Safari Browser. Chrome and Firefox (so PDF.js-based readers) have no such issues.

How it should look like:
![image](https://github.com/user-attachments/assets/477d649c-3a46-4780-9069-40a325ee09d0)

What I get with most viewers (at least on my Mac):
![image](https://github.com/user-attachments/assets/c5337b33-75b1-4d38-9566-8899097dddbc)

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

LaurenzV:
It's an Apple-specific bug, that's why it doesn't work in Safari/Preview/Skim, as they all rely on the same PDF engine (Quartz). It works fine on every other viewers (yellow is Quartz):

![typst-0](https://github.com/user-attachments/assets/e1904885-c53a-4e68-8660-830e58d81840)

With that said, I have known about this bug for a while now, and I haven't found a way to work around it with how we the current SVG embeddal works in Typst...

Issue #4683, opened 2024-08-05
Typst produces empty PDF objects


LaurenzV:
### Description

Using the newest main, compiling a simple file such as:

```
#set text(font: "Fira Sans")

Some
```

Results in the following PDF output (this has been modified with `mutool` to hex encode binary streams, but the empty objects also exist in the original file): https://gist.github.com/LaurenzV/20a35dbeab1d0c9aee0764db4557a743

Note inparticular how object 11, 12, 13 are simply empty.

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

PR #5420, opened 2024-11-14
Switch PDF backend to `krilla`

LaurenzV:
Note that this PR is not ready for review, and probably won't be for a long while. It's just for experimenting, for now.

# Introduction

[krilla](https://github.com/LaurenzV/krilla) is a Rust crate I have been working on in the past few months. It builds on top of the `pdf-writer` crate and is a higher-level crate that allows for the creation of PDF files using high-level primitives. It therefore encompasses the features `svg2pdf` and `typst-pdf` currently have (and a lot more), and abstracts away most complexities of PDF away behind a high-level interface.

# Why?

Using this crate would have multiple advantages:
- It combines the PDF-writing logic from `svg2pdf` and `typst-pdf`, and thus allows us to prevent code duplication, making it easier to implement new features. It also reduces file sizes, since objects can now be cached globally.
- It makes the `typst-pdf` code much shorter and simpler, reducing the chance of bugs (not meaning that we shouldn't implement PDF regression tests at some point). Most of the complexity of PDF creation is abstracted away.
- `krilla` is tested relatively extensively, both using visual regression tests as well as snapshot-based tests, (hopefully) reducing the chance of accidentally introducing bugs in the future. It removes a lot of the burden from Typst to ensure that PDF output is spec-conforming.
- krilla will automatically take care of inserting `ActualText` attributes (fixing https://github.com/typst/typst/issues/526, perhaps also https://github.com/typst/typst/issues/4225)
- krilla supports writing outlined text (making fixing https://github.com/typst/typst/issues/2249 relatively simple, if we desire to do so)
- krilla supports PDF1.4-1.7, as well as the corresponding standards PDF-A1, PDF-A2, PDF-A3, including all their substandards, resolving https://github.com/typst/typst/issues/2942 and https://github.com/typst/typst/issues/5189.
- Resolves https://github.com/typst/typst/issues/4683.
- Resolves https://github.com/typst/typst/issues/3781 for PDF export.
- Resolves https://github.com/typst/typst/issues/5428
- krilla supports custom CMYK ICC profiles, which could be useful for the future, makes some progress towards https://github.com/typst/typst/issues/3143.
- krilla supports tagging and exporting to PDF/UA, making very great progress towards https://github.com/typst/typst/issues/133. All that would be left is removing blockers in the Typst compiler itself.
- krilla supports gradients with transparencies, which should fix https://github.com/typst/typst/issues/2546 (not tested yet).
- SVG filter handling is improved and thus it should fix https://github.com/typst/typst/issues/4926.
- SVGs are written into the page stream directly, which should fix (or rather circumvent) https://github.com/typst/typst/issues/4901 and #5509.

The main disadvantage is that it probably will lead to some performance penalty due to the extra layer of abstraction added, thought it remains to be determined how much. PDF 2.0 and its substandards are currently not supported either, unfortunately.

# Notes

Note that it's not been decided whether we actually want to make that switch, and if it does happen it will still be multiple weeks in the future, as it would require extensive testing (both in terms of correctness as well as performance-wise), and there are still some things left to do in `krilla` itself.

- Closes https://github.com/typst/typst/issues/526 
- Closes https://github.com/typst/typst/issues/5189  
- Closes https://github.com/typst/typst/issues/4683  
- Closes https://github.com/typst/typst/issues/3781  
- Closes https://github.com/typst/typst/issues/5428  
- Closes https://github.com/typst/typst/issues/2546  
- Closes https://github.com/typst/typst/issues/4926  
- Closes https://github.com/typst/typst/issues/4901  
- Closes https://github.com/typst/typst/issues/5509  
- Closes https://github.com/typst/typst/issues/6051.

# TODO List

Next steps:
- [x] You (probably?) will take a closer look at the krilla code logic.
- [x] I will clean up a bit more.
- [x] We still need to look into changing the cluster assignment logic so that ActualText works as intended. I did some more testing and I don't think we will get it to a state where we can close #4425, but hopefully it should still be an improvement.
- [x] I will re-validate that I ported all features of existing `typst-pdf` have been ported.
- [x] I will do some rough testing using 2-3 documents and squash some more bugs.
- [x] You review the code on the Typst side.

After the above is done and you think overall the shape of the code is good, extensive testing would be the next step and roughly like this:

- [x] Check that all of the linked bugs are fixed. (I can do that)
- [x] Run the whole test suite and diff the PDFs visually against current main. (I can do that) 
- [x] Same as above, but check using the Airlington PDF model instead.
- [x] Compile a few different small documents, manually inspect the PDF output and ensure it looks sensible, also look at things like outlines, etc. that can only be inspected manually.
- [x] Test with maybe 5-6 bigger documents (templates, theses documents) and ensure there are no issues in the output.
- [x] While doing the above, also try different export modes and use `verapdf` and Adobe Acrobat to see whether validation passes.
- [x] Run performance tests with a handful of bigger documents (also check single/multi threading) to ensure there are no major performance regressions.
- [x] Also check what the file sizes are like.
- [x] We probably should also do a "stress test" with a very big document (1000+ pages), if such a document exists, as a sanity check that the new backend can deal with bigger documents, too.
- [ ] Creates releases of all dependent crates.
- [ ] 🚢


facundoq:
While HTML output would be superior in terms of accesiblity, I cannot emphasize how much adoption Typst would gain in providing tagged pdf support. I would take a 3x performance hit for this feature.

crates/typst-cli/src/args.rs
```diff
@@ -241,10 +241,14 @@ pub struct CompileArgs {
...
+    /// The version of the produced PDF.
+    #[arg(long = "pdf-version")]
+    pub pdf_version: Option<PdfVersion>,
+
+    /// A PDF standard that Typst will enforce
     /// conformance with.
-    #[arg(long = "pdf-standard", value_delimiter = ',')]
-    pub pdf_standard: Vec<PdfStandard>,
+    #[arg(long = "pdf-standard")]
+    pub pdf_standard: Option<PdfStandard>,
```
laurmaedje, line 251: Is this temporary or do you intend to keep this change? In principle, it was intentional that
- `--pdf-standard` was a list (e.g. you could conform to `a2-a`, `ua-1`, and `x-4` at the same time)
- `--pdf-version` was handled as just another standard

LaurenzV: I would have intended to keep this change, because 1) it's not supported anyway (and wasn't supported before) and 2) I don't think it's super common to target multiple standards? And I think having version and sub-standard separately makes sense, because they can be combined artbirarily, and it doesn't make sense to target more than 1 PDF version.

LaurenzV: I added back the option to specify multiple PDF standards, but will still keep version selection separately for now.

LaurenzV:
@laurmaedje I think this is ready for a first review pass. :) This is not mergeable yet, as there is still some testing left to do, checking performance, bug fixing, etc. but the code in `typst-pdf` should only change very little as part of that. Overall, I (hope) that it's feature-complete and I didn't miss anything, but I also plan on taking another look at that after some time.

I also rendered the whole test suite using ghostscript, pdfium and Quartz, and as you can see in the [generated report](https://drive.proton.me/urls/K20CDDCQD8#cMAcfU2spIMh), for the images that are different, 98% or so are just pixels differences. In a small percentage of examples, some issues actually got fixed that were inconsistent before (for example for patterns).

laurmaedje:
I'd like to keep krilla types out of `typst-pdf`'s public API to be able to evolve API independently and not be exposed to potential API breakage. My preferred way would be to retain the exact API that `typst-pdf` had before the refactor.

laurmaedje:
Regarding review: As the new code here mostly just dispatches to krilla, I'm focussing my efforts on understanding the inner workings of krilla at the moment.

laurmaedje:
Was also wondering whether https://github.com/typst/typst/pull/5637 is still relevant with this or if the features implemented in that PR already exist in krilla.

LaurenzV:
> I'd like to keep krilla types out of typst-pdf's public API to be able to evolve API independently, not be exposed to potential API breakage, and not require an explicit dependency on krilla to use the API. My preferred way would be to retain the exact API that typst-pdf had before the refactor.

From what I can tell this just affects `Validator` and `PdfVersion`, no? Should be easy enough. ^^

> Regarding review: As the new code here mostly just dispatches to krilla, I'm focussing my efforts on understanding the inner workings of krilla at the moment.

Sounds good, let me know if something is unclear.

laurmaedje:
> From what I can tell this just affects Validator and PdfVersion, no? Should be easy enough. ^^

Yep

LaurenzV:
> Was also wondering whether https://github.com/typst/typst/pull/5637 is still relevant with this or if the features implemented in that PR already exist in krilla.

`Interpolate` is not supported, but should be easy enough to add. I think RGBAF32 is also not explicitly supported, but I think it should be enough to just convert those? Other than that, embedding JPEG images as is as well as handling ICC profiles should all work. Though it's also been a while since I last looked at the code, so I'll have to check again once it gets to that. 😅

LaurenzV:
@laurmaedje Rebased, though I haven't cleaned up yet, so don't review on the Typst side yet. Will mark it as ready to review when I think it's good. In the meanwhile, my suggestions on how to proceed:

Next steps:
- [ ] You (probably?) will take a closer look at the krilla code logic.
- [ ] I will clean up a bit more.
- [ ] We still need to look into changing the cluster assignment logic so that ActualText works as intended. I did some more testing and I don't think we will get it to a state where we can close #4425, but hopefully it should still be an improvement.
- [ ] I will re-validate that I ported all features of existing `typst-pdf` have been ported.
- [ ] I will do some rough testing using 2-3 documents and squash some more bugs.
- [ ] You review the code on the Typst side.

After the above is done and you think overall the shape of the code is good, extensive testing would be the next step and roughly like this:

- [ ] Check that all of the linked bugs are fixed. (I can do that)
- [ ] Run the whole test suite and diff the PDFs visually against current main (I can do that)
- [ ] Compile a few different small documents, manually inspect the PDF output and ensure it looks sensible, also look at things like outlines, etc. that can only be inspected manually.
- [ ] Test with maybe 5-6 bigger documents (templates, theses documents) and ensure there are no issues in the output.
- [ ] While doing the above, also try different export modes and use `verapdf` and Adobe Acrobat to see whether validation passes.
- [ ] Run performance tests with a handful of bigger documents (also check single/multi threading) to ensure there are no major performance regressions.
- [ ] We probably should also do a "stress test" with a very big document (1000+ pages), if such a document exists, as a sanity check that the new backend can deal with bigger documents, too.
- [ ] Creates releases of all dependent crates.
- [ ] 🚢

Open for other suggestions!

laurmaedje:
> You (probably?) will take a closer look at the krilla code logic.

Yeah, my current main focus is web backend stuff, but I do have this on my mind for this week. I wouldn't consider this too blocking for any further work or review on the Typst side though. The main purpose of this for me is just to understand krilla better.

> We still need to look into changing the cluster assignment logic so that ActualText works as intended. I did some more testing and I don't think we will get it to a state where we can close https://github.com/typst/typst/pull/4425, but hopefully it should still be an improvement.

Is that the correct link? ("Fix luma to CMYK conversion" seems unrelated). Just let me know what you need from me regarding the cluster assignment logic.

>  I will re-validate that I ported all features of existing typst-pdf have been ported.

> I will do some rough testing using 2-3 documents and squash some more bugs.

> You review the code on the Typst side.

Sounds good! 

> Run the whole test suite and diff the PDFs visually against current main (I can do that)

I think you already a script for that, right? But if not, I can send you one we have.

> Compile a few different small documents, manually inspect the PDF output and ensure it looks sensible, also look at things like outlines, etc. that can only be inspected manually.

> Test with maybe 5-6 bigger documents (templates, theses documents) and ensure there are no issues in the output.

> While doing the above, also try different export modes and use verapdf and Adobe Acrobat to see whether validation passes.

> Run performance tests with a handful of bigger documents (also check single/multi threading) to ensure there are no major performance regressions.

> We probably should also do a "stress test" with a very big document (1000+ pages), if such a document exists, as a sanity check that the new backend can deal with bigger documents, too.

The further testing plan sounds great and comprehensive. I appreciate that a lot!

reknih:
One thing that I think could be a good amendment of the testing plan is to run the [Arlington PDF Model](https://pdfa.org/resource/arlington-pdf-model/) against the files krilla generates. The model is a formal test suite for the PDF spec built by the PDF Association and veraPDF has [released a version of their checker](https://software.verapdf.org/releases/arlington/1.26) that uses the model instead of their PDF/A etc. rulesets. Since it's otherwise veraPDF, it should be easy to run it in CI using the same code.

LaurenzV:
Changed the text range assignment behavior now. It does make things better on Acrobat, though the output is still not as good as in LibreOffice, I think,so https://github.com/typst/typst/issues/4225 will remain unsolved for now.

laurmaedje:
I familiarized myself with the krilla codebase over the past weeks. It's good stuff! 🎉 Many good abstractions like `register_cacheable` and `register_resourceable`. I looked at almost everything at least a bit and some crucial things (like the `TextSpanner`) in more detail.

I made some notes, which you can find below. Most of these aren't really blockers, but rather just thoughts and suggestions. There are a few things though that should probably be fixed before merging into Typst.

---

### Typst-Krilla integration
- It looks like image alt text is not written anymore.
- The `krilla::image::Image::from_jpeg` code path is a bit problematic because it throws away the ICC profile (and other potential future additions to `CustomImage`).
- The non-krilla JPEG code path for rotated images is _also_ problematic though because it reintroduces https://github.com/typst/typst/issues/3781 for rotated CMYK JPEGs (tested with a Photoshop JPEG). For handling JPEG rotations, we could take inspiration from `jpegtran` (which does lossless JPEG transformations) or maybe just do the rotation in PDF? Or even if we have to recode, still use the correct CMYK handling code path.
- Currently, Typst uses `TextRenderingMode::FillStroke` if text has both a fill and a stroke and `fill_even_odd_and_stroke` for shapes. The krilla PR seems to write the text and shapes twice instead. Could this mess with text accessibility? Would it make sense to unify `fill_glyphs` and `stroke_glyphs` into one function that takes an optional stroke?
- Gradient stops are always converted to sRGB, even though krilla supports CMYK stops. Is that on purpose?
- The direction in the viewer preference dictionary is not set anymore
- I think we depend on both `png` and `zune-png` now, which is a bit unfortunate
- A bit unfortunate that font data is hashed again even though it's already hashed on the Typst side...
- If `CustomImage`s are supposed to be cheap to clone and hash, it would probably make sense to wrap the full fields in `Arc` (via a repr struct) instead of wrapping `alpha_channel` and `actual_dynamic` separately (currently, they aren't shared between clones).

### Krilla internals
- Might make sense to put the `root_builder` and `sub_builders` into a separate child struct of `Surface` so that `cur_builder_mut` can be a method taking `&mut self` on that
- The various panicking methods (like `start_marked_content`) could use `#[track_caller]` to make it easier for the API user to find the bug in their code
- The folder structure is a _bit_ convoluted imo, with both top-level modules and `object::..` modules ending up as public modules and especially also with the two `font/` directories.
- Maybe move `PdfFont` to font module?
- I think breaking up the `content` module into logical parts (like text where also `naive_shape` could go) would make it a bit more manageable
- In `TextSpanner`, we always use a span if two glyphs share the same range (i.e. same cluster). I think that would happen quite frequently for e.g. Arabic. However, we only add a CMap entry if `forbid_invalid_codepoints` is true. Does that mean that an Arabic document without PDF/UA or similar will have bad copy-paste in viewers other than Acrobat and Chrome? Is there a downside to always writing a CMap entry?
- The text writing pipeline does quite a lot of things for each glyph run / glyph. If you haven't already, I think it would be interesting to benchmark some documents consisting of just a ton of text to see whether it has considerable overhead or whether it's negligible.
- Krilla writes a lot more q/Q and style setting operators in content streams than typst-pdf, applying isolation even if not necessary. Maybe it's not too big of a deal, but I wonder if it wouldn't be worth it to extend the already partially simulated graphics state to optimize the content stream.
- Quite lovely how compact the `svg/` code is!
- https://github.com/typst/typst/pull/4871 removed the last PostScript code generated by typst-pdf. krilla does generate Postscript code. Is that because it supports more features / is more correct?
- The [Photoshop CMYK thing](https://github.com/LaurenzV/krilla/blob/399dc596fdb27fde3ca1eef27b74625affed199a/crates/krilla/src/object/image.rs#L440-L450) is interesting.
  - First of all: It seems to work correctly with CMYK JPEGs generated from both paint.net and Photoshop.
  - I think the comment in `libtexpdf` is just confused. From my current understanding, the Adobe APP marker isn't anything Adobe/Photoshop-specific, but [rather just a marker that specifies the color space](https://github.com/etemesi254/zune-image/blob/4a073b12947f9d1003ae931b7b7cd9da2914d4d8/crates/zune-jpeg/src/headers.rs#L426-L449). It doesn't seem like Photoshop is doing anything wrong here.
  - Are you sure inverting for YCCK as well is correct? Do you have a test file? [ISO/IEC 10918-6:2013](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-T.872-201206-I!!PDF-E&type=items) indicates that CMYK (= APP14 marker equals zero) values are stored complemented, but I could find no such note for YCCK.
  - Here are some additional relevant links:
    - https://graphicdesign.stackexchange.com/questions/12894/cmyk-jpegs-extracted-from-pdf-appear-inverted
    - https://bitmiracle.github.io/libjpeg.net/help/articles/KB/special-color-spaces.html
- Why is DeviceRgb used instead of sRGB for images without ICC profile?
- There seems to be quite some duplication in CMap generation between CID- and Type3-Fonts
- The comment in `fn base_font_name` is the same as in Typst, but the name is not actually hashed like in Typst

### Krilla API design
- Maybe move `PageSettings` to page.rs?
- I would recommend not having `rustybuzz` in the public API (via `Feature` in `fill_text`) because every rustybuzz bump is a breaking change in krilla.
- Various of the `Path` APIs expose types that are not public. Not sure whether it's worth having `tiny-skia-path` in the public API.
- If possible, I would even avoid having `pdf-writer` in the public API. But if it stays, it should probably be re-exported. (I want to do that in Typst as well for `comemo` and `ecow`.)
- As for `usvg`, it seems like the SVG handling code could mostly be written with krilla's public API? A potential `krilla-svg` crate could be versioned with `usvg` while allowing the main `krilla` crate to move at its own pace.
- Personally, I would consider having some types (e.g. `Document`) only at the top-level and keeping the modules private instead of reexporting, duplicating the API, and making the user choose. But that's just personal preference. Some of the reexports are also shown inlined in the docs, e.g. `Page`.
- When does krilla panic vs raising a `UserError`?
- I think the `KrillaError` variant names could drop the `Error` suffix
- `ContentTag::Span` should probably have named fields
- I wonder whether the type parameter on `Stop` really pulls its weight. Is the purpose type safety w.r.t. the color space? Also, is there a specific reason why luma is not valid for stops?

LaurenzV:
Exhaustive list of issues that will be able to be closed (I seemingly can't add them manually):
#526 #5189 #4683 #3781 #5428 #2546 #4926 #4901 #5509 #6051.

Verified manually that they are fixed in this branch.

LaurenzV:
@laurmaedje I think I've addressed most of your comments, I've also refactored large parts of krilla and I'm pretty happy now, so from my side this is good to go. Once you are happy with the general code on the `typst-pdf` side, we can move on to the next stage and do some more exhaustive tests.

> Might make sense to put the root_builder and sub_builders into a separate child struct of Surface so that cur_builder_mut can be a method taking &mut self on that

Done!

> The various panicking methods (like start_marked_content) could use #[track_caller] to make it easier for the API user to find the bug in their code

I added them in a few places.

> The folder structure is a bit convoluted imo, with both top-level modules and object::.. modules ending up as public modules and especially also with the two font/ directories.

> Maybe move PdfFont to font module?

> I think breaking up the content module into logical parts (like text where also naive_shape could go) would make it a bit more manageable

>  Maybe move PageSettings to page.rs?

> Personally, I would consider having some types (e.g. Document) only at the top-level and keeping the modules private instead of reexporting, duplicating the API, and making the user choose. But that's just personal preference. Some of the reexports are also shown inlined in the docs, e.g. Page.

I've completely restructured the crate structure now, let me know if you think it's better now!

> In TextSpanner, we always use a span if two glyphs share the same range (i.e. same cluster). I think that would happen quite frequently for e.g. Arabic. However, we only add a CMap entry if forbid_invalid_codepoints is true. Does that mean that an Arabic document without PDF/UA or similar will have bad copy-paste in viewers other than Acrobat and Chrome? Is there a downside to always writing a CMap entry?

That should be true, yes. The downsides of always writing a CMap entry are
- The copy-pasting will still be broken in other viewers, the only difference is that instead, you every syllable will be duplicated x times (x being the number of glyphs in the same cluster). That's actually the behavior I had beginning, and I think it looked quite horrible in many cases (and also doesn't seem to be what happens in Latex or LibreOffice).
- It will "pollute" potentially valid cmap mappings for other glyphs in the future, since we don't replace them once set.

> The text writing pipeline does quite a lot of things for each glyph run / glyph. If you haven't already, I think it would be interesting to benchmark some documents consisting of just a ton of text to see whether it has considerable overhead or whether it's negligible.

Already answered.

> Krilla writes a lot more q/Q and style setting operators in content streams than typst-pdf, applying isolation even if not necessary. Maybe it's not too big of a deal, but I wonder if it wouldn't be worth it to extend the already partially simulated graphics state to optimize the content stream.

I feel like it's not worth the trouble, it opens a whole new set of potential bugs that come with simulating the graphics state, and since PDF offers the feature, I don't see why we shouldn't use it. :D I highly doubt that this has any impact on the performance of rendering the document.

> Quite lovely how compact the svg/ code is!

:)

> https://github.com/typst/typst/pull/4871 removed the last PostScript code generated by typst-pdf. krilla does generate Postscript code. Is that because it supports more features / is more correct?

Already answered.

> The [Photoshop CMYK thing](https://github.com/LaurenzV/krilla/blob/399dc596fdb27fde3ca1eef27b74625affed199a/crates/krilla/src/object/image.rs#L440-L450) is interesting.

I just tried, turns out I already have an YCCK image in the test suite, and removing the inversion there breaks it in all readers, so it does seem like things are a bit more complicated here. I suggest leaving it as is, and in case we get bug reports we can investigate further.

> Why is DeviceRgb used instead of sRGB for images without ICC profile?

Fixed.

> There seems to be quite some duplication in CMap generation between CID- and Type3-Fonts

Changed this now by creating a macro encapsulating the commmon parts.

> The comment in fn base_font_name is the same as in Typst, but the name is not actually hashed like in Typst

Fixed.

> I would recommend not having rustybuzz in the public API (via Feature in fill_text) because every rustybuzz bump is a breaking change in krilla.

I removed it from the public API (as well as skrifa).

> Various of the Path APIs expose types that are not public. Not sure whether it's worth having tiny-skia-path in the public API.

The first part shouldn't occur anymore as our tests are now outside of the core crate. I also removed `tiny-skia-path` from the public API and created custom wrapper types instead.

> If possible, I would even avoid having pdf-writer in the public API. But if it stays, it should probably be re-exported. (I want to do that in Typst as well for comemo and ecow.)

I also removed `pdf-writer` from the public API.

> As for usvg, it seems like the SVG handling code could mostly be written with krilla's public API? A potential krilla-svg crate could be versioned with usvg while allowing the main krilla crate to move at its own pace.

It lives in its own crate now!

> When does krilla panic vs raising a UserError?

That's a good question, I think the current distinction isn't really consistent. Since all of them are quite serious user errors which might not be really possible to recover from, I replaced them with panics instead and removed `UserError`. The downside is that we can't catch them on the Typst side, but assuming that we test sufficiently, they shouldn't happen in the first place.

> I think the KrillaError variant names could drop the Error suffix

Done.

> ContentTag::Span should probably have named fields

Done.

> I wonder whether the type parameter on Stop really pulls its weight. Is the purpose type safety w.r.t. the color space?

Yes that's the reason, but after thinking about it I've also decided to remove it, and instead panic if stops don't have the same number of components.

> Also, is there a specific reason why luma is not valid for stops?

Just an oversight, added it!

### integration
> It looks like image alt text is not written anymore.

I've added a temporary method to krilla so that we can easily write it without having to implement the other tagged PDF stuff for now.

> The krilla::image::Image::from_jpeg code path is a bit problematic because it throws away the ICC profile (and other potential future additions to CustomImage).

Not sure I follow here, krilla does extract the ICC profile from JPEG image. What other future additions? Any specific suggestions in this regard?

> The non-krilla JPEG code path for rotated images is also problematic though because it reintroduces
CMYK JPEG and TIFF images are broken in PNG and PDF exports  #3781 for rotated CMYK JPEGs (tested with a Photoshop JPEG). For handling JPEG rotations, we could take inspiration from jpegtran (which does lossless JPEG transformations) or maybe just do the rotation in PDF? Or even if we have to recode, still use the correct CMYK handling code path.

I've gone the route of just rotating it in the PDF now. I don't think it makes sense to add this in krilla directly since it's a bit hacky and arbitrary, but I've tested it manually with the Typst implementation and it seems to work.

> Currently, Typst uses TextRenderingMode::FillStroke if text has both a fill and a stroke and fill_even_odd_and_stroke for shapes. The krilla PR seems to write the text and shapes twice instead. Could this mess with text accessibility? Would it make sense to unify fill_glyphs and stroke_glyphs into one function that takes an optional stroke?

I would like to avoid that, because that would also require us to add another additional parameter for whether we want to do fill/stroke or stroke/fill, and I think most other graphic libraries also don't really allow you to do both at the same time, that seems to be more of an SVG/PDF thing. It shouldn't be an issue for accessibility because for fill + stroke, the stroke will be drawn purely as a vector path, so it will not be interpreted as text. Later on, we can also mark it as an artifact in tagged PDF. 

> Gradient stops are always converted to sRGB, even though krilla supports CMYK stops. Is that on purpose?

Fixed!

> The direction in the viewer preference dictionary is not set anymore

Fixed!

> I think we depend on both png and zune-png now, which is a bit unfortunate

> A bit unfortunate that font data is hashed again even though it's already hashed on the Typst side...

Already answered. Also, keep in mind that we don't hash actually hash the whole font, but we just use basic metadata (which includes the checksum and the length of the raw data), which should be very fast to hash but still 99.9999%+ unique. I actually forgot that I wanted you to ask about this, it should be fine to do it this way, right? Could you take a look at the logic in case you haven't?

> If CustomImages are supposed to be cheap to clone and hash, it would probably make sense to wrap the full fields in Arc (via a repr struct) instead of wrapping alpha_channel and actual_dynamic separately (currently, they aren't shared between clones).

Done!

---

@reknih

> One thing that I think could be a good amendment of the testing plan is to run the [Arlington PDF Model](https://pdfa.org/resource/arlington-pdf-model/) against the files krilla generates. The model is a formal test suite for the PDF spec built by the PDF Association and veraPDF has [released a version of their checker](https://software.verapdf.org/releases/arlington/1.26) that uses the model instead of their PDF/A etc. rulesets. Since it's otherwise veraPDF, it should be easy to run it in CI using the same code.

I now run both, Arlington tests on all produced PDFs (including the `usvg` integration), as well as `veraPDF` on tests with validated export in CI!

laurmaedje:
> I think I've addressed most of your comments, I've also refactored large parts of krilla and I'm pretty happy now, so from my side this is good to go. Once you are happy with the general code on the typst-pdf side, we can move on to the next stage and do some more exhaustive tests.

Great! I will take a look at typst-pdf.

---

> I've completely restructured the crate structure now, let me know if you think it's better now!

Looks much clearer!

> The downsides of always writing a CMap entry are [...]

Okay.

> I feel like it's not worth the trouble, it opens a whole new set of potential bugs that come with simulating the graphics state, and since PDF offers the feature, I don't see why we shouldn't use it. :D I highly doubt that this has any impact on the performance of rendering the document.

I was thinking more of file size than performance, but with compression it's probably very minor. I'm okay with not doing it or at least not now.

> I just tried, turns out I already have an YCCK image in the test suite, and removing the inversion there breaks it in all readers, so it does seem like things are a bit more complicated here. I suggest leaving it as is, and in case we get bug reports we can investigate further.

Interesting. I might take another look.

> Changed this now by creating a macro encapsulating the commmon parts.

See https://github.com/LaurenzV/krilla/pull/169.

> I removed it from the public API (as well as skrifa).

> The first part shouldn't occur anymore as our tests are now outside of the core crate. I also removed tiny-skia-path from the public API and created custom wrapper types instead.

> I also removed pdf-writer from the public API.

> It lives in its own crate now!

👍

> Since all of them are quite serious user errors which might not be really possible to recover from, I replaced them with panics instead and removed UserError. 

Sounds good.

> I've added a temporary method to krilla so that we can easily write it without having to implement the other tagged PDF stuff for now.

Originally, this was implemented with the `/Alt` attribute on images rather than with marked content. Is there a specific reason to not just support that instead?

> Not sure I follow here, krilla does extract the ICC profile from JPEG image. What other future additions?

I mean a potential [user-specified ICC profile.](https://typst.app/docs/reference/visualize/image/#parameters-icc) And with future additions I mean other fields in Typst that could override data from within the image.

> I've gone the route of just rotating it in the PDF now. I don't think it makes sense to add this in krilla directly since it's a bit hacky and arbitrary, but I've tested it manually with the Typst implementation and it seems to work.

That's reasonable I think, though other krilla users might very well hit the same problem that their image rotation is wrong, so I'm not sure it would be so arbitrary to do it in krilla.

> I would like to avoid that, because that would also require us to add another additional parameter for whether we want to do fill/stroke or stroke/fill, and I think most other graphic libraries also don't really allow you to do both at the same time, that seems to be more of an SVG/PDF thing. It shouldn't be an issue for accessibility because for fill + stroke, the stroke will be drawn purely as a vector path, so it will not be interpreted as text. Later on, we can also mark it as an artifact in tagged PDF.

It's imo a bit unfortunate to not use the PDF-native text stroking capabilities and embed potentially large paths instead. I would generally expect the stroke to be above the fill since otherwise it's just cut in half, right? I'm not sure it's really necessary to support both modes. But these are just quick thoughts off the top of my head.

> Also, keep in mind that we don't hash actually hash the whole font, but we just use basic metadata (which includes the checksum and the length of the raw data)

Didn't see that.

> I actually forgot that I wanted you to ask about this, it should be fine to do it this way, right? Could you take a look at the logic in case you haven't?

It's probably fine, though I'm not certain whether the checksum field is always properly set in fonts in the wild. A comment in the source says "The proper way would be to hash the whole font data, but this is just too expensive." Is that an assumption or the result of benchmarking? In the case of Typst, we should also make sure than Typst -> krilla font conversion is memoized across compilations (not just cached within a compilation), which I think isn't the case right now.

LaurenzV:
> I was thinking more of file size than performance, but with compression it's probably very minor. I'm okay with not doing it or at least not now.

I reaaaaally doubt this will make any noticeable difference, maybe a few kilobytes at most. :p

> I mean a potential [user-specified ICC profile.](https://typst.app/docs/reference/visualize/image/#parameters-icc) And with future additions I mean other fields in Typst that could override data from within the image.

Ah I didn't realize you could already set a custom ICC profile... I guess I'll have to come up with something then.

> That's reasonable I think, though other krilla users might very well hit the same problem that their image rotation is wrong, so I'm not sure it would be so arbitrary to do it in krilla.

The main problem is that you have to manually specify the size of the image when drawing an image (which doesn't have to correspond to the natural size!), so what happens if we need to flip the dimensions? Either we assume the user supplied the natural dimensions (which I don't think is a good assumption to make) and we flip it internally, or we do nothing and the user must therefore read the metadata themselves and flip the dimensions if needed, which kind of defeats the purpose of abstracting it away. If that makes sense?

> It's imo a bit unfortunate to not use the PDF-native text stroking capabilities and embed potentially large paths instead. I would generally expect the stroke to be above the fill since otherwise it's just cut in half, right? I'm not sure it's really necessary to support both modes. But these are just quick thoughts off the top of my head.

Fair point, it's probably pretty bad for file sizes. I was thinking about SVG where you can have fill-stroke or stroke-fill. I will think about it a bit more.

> It's probably fine, though I'm not certain whether the checksum field is always properly set in fonts in the wild.

Yeah but as I said we use the whole size of the font (the exact number of bytes in a font should be very unique already) _and_ all kinds of metadata, I think it's pretty much impossible for that data to match across two different fonts.

> "The proper way would be to hash the whole font data, but this is just too expensive." Is that an assumption or the result of benchmarking?

I think at some point I tried it, but it's also very important for SVG rendering because each time we render an SVG we need to reload the fonts from the database, so it's important that it's fast.

> In the case of Typst, we should also make sure that Typst -> krilla font conversion is memoized across compilations (not just cached within a compilation), which I think isn't the case right now.

This means extracting the font conversion into a method that is memoized, right?

LaurenzV:
Will be starting with the testing phase now.

LaurenzV:
The whole test suite passes with Airlington PDF now, too.

LaurenzV:
I ran benchmarks with the following five Typst documents:

masterproef-main (https://github.com/Dherse/masterproef)
phd thesis (https://www.github.com/jrihon/thesis)
OI-Wiki (https://github.com/OI-wiki/OI-Wiki-export)
lorem (Simple Typst document with lots of lorem paragraphs)
project-iii-typst (https://github.com/samuelireson/project-iii-typst)

Results:

```
masterproef:
Clean compiles, all jobs:
- krilla: 1.6s
- main: 1.5s

Clean compiles, 1 job:
- krilla: 3.5s
- main: 3.9s

Watch, all jobs:
- main: 300ms
- krilla: 330ms

phd thesis:
Clean compiles, all jobs:
- krilla: 2.6s
- main: 4.7s

Clean compiles, 1 job:
- krilla: 7.8s
- main: 12.2s

Watch, all jobs:
- main: 1.5s
- krilla: 0.22s

lorem:
Clean compiles, 1 job:
- krilla: 9.9s
- main: 11.6s

Clean compiles, all jobs:
- krilla: 4.0s
- main: 4.1s

Watch, all jobs:
- main: 1.4s
- krilla: 1.2s

oi-wiki (it should be noted that the results are not very meaningful for this one, because according to my benchmarking only 2-3 seconds are actually
spent writing the PDF):
Clean compile, 1 job:
- krilla: 117s
- main: 118.56s

Clean compile, all jobs:
- krilla: 56.258s
- main: 55.796
```

I think there are two conclusions to draw from this:
- For documents with many SVGs, we can expect much better clean compile times, on the one hand since SVGs are much more tightly integrated in the document, on the other hand since we also use a new crate for compression, which is faster than the previous one.
- However, due to the same reasons the same documents might see much worse performance for incremental compilation (only when compiling to PDF), the reason being that SVGs cannot be cached anymore due to above reason. That is a bit unfortunate, but is a trade-off we have to accept, I think.
- Other than that, there don't appear to be any major changes.

LaurenzV:
Here's the file sizes for the documents I tested:

image

As you can see, the file size does increase in certain cases, but the reason for this is that JPEG images are now embedded as raw instead of re-encoding them. It seems like the the image used to re-encode them with a lower file size (but therefore probably also lower quality), so that's not a regression.

Other than that, there are some nice file size gains on the PHD thesis document, most likely due to the fact that resources can now be re-used between different SVGs.

LaurenzV:
@laurmaedje I think from my side I am mostly done now. There is one bug remaining that I found, when compiling the following document:

```
#set par(justify: true)
#h(15cm) explanation
```

trying to compile with the standard `a-2u` doesn't work, because it complains about the hyphenation not being assigned a codepoint. I checked in the debugger, and for some reason it gets assigned the range 5..5. Though from what I can tell, in the code I changed it does get assigned the right range, so I assume that some code afterwards breaks because it relied on the previous behavior. Since you are probably more familiar with that code, if you can take a look that would be great.

If you want to do some more testing yourself, that would be appreciated, too. But given the tests I've already run I think we can be pretty sure that no major regressions should be introduced, though there obviously is only one way to find out. 😬

laurmaedje:
> There is one bug remaining that I found, when compiling the following document: [...]

Hyphens are created [here](https://github.com/typst/typst/blob/012e14d40cb44997630cf6469a446f217f2e9057/crates/typst-layout/src/inline/shaping.rs#L479-L480), always with zero-sized ranges. They don't appear in the source `text`, so their codepoint can't really be accessed via range, unfortunately. On main, the assigned text in the `glyph_set` is also `""`.

LaurenzV:
Ah, that's unfortunate, because it means that a document with hyphenation cannot be exported to PDF/A2/A3-u... But I guess this can be figured out at a later point.

laurmaedje:
Something about krilla's output is not reproducible. Just running the Typst test suite twice will generate differing PDFs. I'm not sure why.

PR #5421, opened 2024-11-14
Use `codex` for symbols

MDLC01:
Following [the discussion on Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1306584824269508608), this PR removes symbol names from this repo, and instead uses [`codex`](https://github.com/typst/codex).

The behavior of callable symbols is slightly altered: they can now only be called when in math mode. This can be reverted if necessary, but I don't think it is a problem, because callable symbols all return math content anyway, so any previous call to a symbol outside of math can be wrapped in `$..$` without any behavior change. For example, `#sym.dash[x]` can be replaced with `$dash(x)$` (which is shorter and arguably better).
"𝑥" with an overline.

Another thing to note is that math mode only imports symbols from `sym` in the prelude, instead of importing all entries from `sym`. If we add sub-modules to `sym` in the future, this means they won't be available by default in math. This can be changed easily, and does not require too much thoughts for now, as `sym` currently has no sub-module.


MDLC01:
I included some other changes:
- The fourth commit slightly improves the code readability. I can revert the `Repr` variant name changes if wanted.
- The fifth commit prevents the user from declaring modifiers that are not valid identifiers (they can't be accessed anyway, and caused invalid autocomplete in the web app).
- The sixth commit adds some tests.

laurmaedje:
Dealing with callable symbols results in quite a bit of complexity (and a bit of performance/memory penalty) here, in particular, the extra allocations (`Box<[..]>`) and the macro.

I wonder: What if we get rid of `SymChar`, directly use the codex slices in `Repr::Complex`, and resolve the function via a match on the Unicode codepoint returned by `.get()`. I don't think we'd ever want to have a function on a symbol, but not on an alias of it. That would cut down on all of this complexity.

MDLC01:
This is probably better indeed. I also lifted the restriction that sub-modules are not accessible within math mode (I mainly added this restriction in the first place to prevent code duplication, but wasn't sure it really made sense). Additionally, symbols can now once again be called in any mode.

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -192,7 +210,18 @@ impl Symbol {
...
                 bail!(span, "duplicate variant");
             }
-            list.push((v.0, SymChar::pure(v.1)));
+            if !v.0.is_empty() {
+                for modifier in v.0.split('.') {
+                    if !is_ident(modifier) {
+                        bail!(
+                            span,
+                            "invalid symbol modifier: {}",
+                            crate::foundations::Repr::repr(modifier),
```
laurmaedje, line 219:
```suggestion
                            modifier.repr(),
```

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -1,17 +1,14 @@
...
 use std::sync::Arc;
 
 use ecow::{eco_format, EcoString};
 use serde::{Serialize, Serializer};
-use typst_syntax::{Span, Spanned};
+use typst_syntax::{is_ident, Span, Spanned};
 
 use crate::diag::{bail, SourceResult, StrResult};
-use crate::foundations::{cast, func, scope, ty, Array, Func};
+use crate::foundations::{cast, func, scope, ty, Array, Func, NativeFunc};
```
laurmaedje, line 11: you can import `Repr as _` here to prevent a naming conflict

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -286,20 +287,20 @@ cast! {
...
     },
 }
 
 /// Iterator over variants.
 enum Variants<'a> {
-    Single(std::option::IntoIter<SymChar>),
-    Static(std::slice::Iter<'static, (&'static str, SymChar)>),
-    Runtime(std::slice::Iter<'a, (EcoString, SymChar)>),
+    Single(std::option::IntoIter<char>),
+    Static(std::slice::Iter<'a, (&'static str, char)>),
```
laurmaedje, line 298: This 'a could go back to 'static I think.

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -46,73 +43,94 @@ use crate::foundations::{cast, func, scope, ty, Array, Func};
...
+            _ => bail!("symbol {self} is not callable"),
+        }
     }
 
     /// Apply a modifier to the symbol.
     pub fn modified(mut self, modifier: &str) -> StrResult<Self> {
-        if let Repr::Const(list) = self.0 {
-            self.0 = Repr::Multi(Arc::new((List::Static(list), EcoString::new())));
+        if modifier.is_empty() {
+            return Ok(self);
```
laurmaedje, line 126: why this new `modifier.is_empty()` is empty check?

crates/typst-library/src/foundations/symbol.rs
MDLC01: This does not work because it requires importing the `Repr` trait, and prevents us from using `Repr` as the name of the type contained in `Symbol`.

crates/typst-library/src/foundations/symbol.rs
MDLC01: That was made now removed test easier to write; this is no longer necessary.

laurmaedje:
Thank you for your ongoing work on symbols!

Issue #5102, opened 2024-10-03
Section name should put after the section number in reference in Chinese


peng1999:
### Description

In Chinese, when referring to a section, it should be `xx 小节` instead of `小节 xx`.
So the following code
 
```typst
#set text(lang: "zh")
#set heading(numbering: "1.1")

= Section <xxx>

@xxx
```

should produce `1 小节` rather than `小节 1`.

### Typst version

- [X] I am using the latest version of Typst

YDX-2147483647:
Relates-to: #2485 — A similar issue on Figures/Tables, and some other recommended `第1节`.

See also https://github.com/w3c/i18n-issues/issues/10 and [clreq-gap for typst](https://typst-doc-cn.github.io/clreq/#for-references-to-headings-the-supplement-should-not-be-put-before-the-number).

YDX-2147483647:
## Workaround (for typst v0.13.1)

Here's a real dark magic that solves #5778 together.

```typst
= 要求：第一节约，第二不知 <sec:1>
正如@sec:1\所说。

= 标题正是第二节
也如@sec:1\所说。
```

<img width="600" src="https://github.com/user-attachments/assets/29921551-256e-4bf8-9fa7-bc866da8dd23" />

```typst
#import "@preview/numbly:0.1.0": numbly
#set heading(
  numbering: numbly("第{1:一}节"), // the numbering for reference
  supplement: it => none,
)

#let _matched = state("matched", false)
#let once(fn) = {
  original => context if not _matched.get() {
    fn(original)
    _matched.update(true)
  } else {
    original
  }
}
#let reset() = _matched.update(false)

#show heading: it => {
  show counter(heading).display(it.numbering): once(it => none)
  reset()

  show h.where(amount: 0.3em): counter(heading).display("一、") // the numbering for display
  it
}
```

Features:

- No `show ref: …`.
- Headings are not reconstructed.
- Works for more than 5 headings.

voroskoi:
Hi,

I would like to add that this is also an issue in Hungarian. I have started creating a package for Hungarian support in Typst and it is full of rewriting these rules:
- images
- tables
- raw
- equations

Beside `figure` rules You have to rewrite `ref`, `outline` etc. show rules as well. While I fix these God knows how many issues I create because not using the standard rules means one have to keep these packages updated while the ecosystem changes.
For example outline.entry has the prefix() method, but that also returns numbering and supplement, which means these languages have to re-implement everything.

[...]

PR #5429, opened 2024-11-15
Support for Basque language

iagobaapellaniz:
It needs a bit more investigation. For instance `Figure 1` should be spelled `1. Irudia`. Also hyphenation has to be investigated a bit. It shares a lot of features with Spanish. But for the moment is more than enough.


crates/typst-library/translations/eu.txt
```diff
@@ -0,0 +1,8 @@
+figure = Irudia
+table = Taula
+equation = Ekuazioa
+bibliography = Bibliografia
+heading = Atala
+outline = Aurkibidea
+raw = Kodea
+page = Orria
```
laurmaedje, line 8: In the [Spanish translation](https://github.com/typst/typst/blob/main/crates/typst-library/translations/es.txt) "page" is written in lowercase. Should that be the case here as well?

iagobaapellaniz: Could be. It depends if it is at the beginning of a sentence. The same for section, raw, and everything else.

Besides, in Spanish raw is also poorly translated. I plan to fix it.

laurmaedje: > It depends if it is at the beginning of a sentence. The same for section, raw, and everything else.

It's not possible to distinguish for now. You should assume that it's used within the sentence. At the start of the sentence would be less common and needs to be handled manually.

iagobaapellaniz: It's OK then. I think the PR is ready for supporting basque. I would investigate on the hyphen part later. Thanks for the comments.

peng1999:
> For instance `Figure 1` should be spelled `1. Irudia`.

Related #5102

iagobaapellaniz:
> Related #5102

Indeed the period after is somehow needed in the basque grammar. The period means 1st, 2nd and so on. If not included, a space could be enough but it is weird. On the other hand, a space works better for other counters like `(1) Ekuazioa` or `IV Atala`.

laurmaedje:
Thanks!

PR #5432, opened 2024-11-16
Add QSL card standard to paper sizes

JanssenSimon:
I'm looking to design my HAM radio club's QSL cards with Typst and figured it'd be nice to include the standard QSL card format in the available paper dimensions.


laurmaedje:
I think this is a bit too specialized to be built in. It's very easy to set custom width and height, after all.

Issue #762, opened 2023-04-12
List/Enum/Term Syntax is Awkward (Sometimes)


frozolotl:
I will give a few code samples and how they are rendered at the time of writing.

The part that is important for whether an element is part of a list item is not the start, but the end of the list marker.
```typst
9. foo
  - bar
10. foo
  - bar
```
![image](https://user-images.githubusercontent.com/44589151/231519693-6feb4dc6-d0e2-4fec-9795-2991d7531641.png)
---

Leading comments do not count as indentation, but are still significant at determining the base indentation level of the next item. 
```typst
- foo
/**/- bar
    /**/- baz
```
![image](https://user-images.githubusercontent.com/44589151/231520109-79fa30e9-d6bd-4664-9ffb-7898dac14cc7.png)
---

What counts as indentation is not the space before the list marker, but the space before the first non-space character till the list marker.
```typst
- foo
  - bar
/**/    - baz
```
![image](https://user-images.githubusercontent.com/44589151/231520542-1361b0ab-0bce-4529-84f2-5aea23d23e2b.png)
---

A list item with a different, but lower indentation is still considered part of the list. This is a bit surprising, but can be considered technically correct.
```typst
- foo
    - bar
  - baz
```
![image](https://user-images.githubusercontent.com/44589151/231520723-1d7b2075-2054-4e4a-9d9d-32c06ecf1eca.png)
---

One tab is the same as one space. This can be correct in some editor configurations, but unexpected in others. Maybe the mixing of tabs and spaces should be an error?
```typst
- foo
  - bar
		- baz
```
![image](https://user-images.githubusercontent.com/44589151/231520998-83503673-ead6-4b4f-8d7f-a9d1cd79cc87.png)

laurmaedje:
> The part that is important for whether an element is part of a list item is not the start, but the end of the list marker.

This has been [fixed since](https://github.com/typst/typst/pull/1824).

> Leading comments do not count as indentation, but are still significant at determining the base indentation level of the next item.

> What counts as indentation is not the space before the list marker, but the space before the first non-space character till the list marker.

This is indeed questionable. Maybe they should rather count as indentation.

> A list item with a different, but lower indentation is still considered part of the list. This is a bit surprising, but can be considered technically correct.

This is fine, I think.

> One tab is the same as one space. This can be correct in some editor configurations, but unexpected in others. Maybe the mixing of tabs and spaces should be an error?

[...]

PR #5433, opened 2024-11-16
Update list/enum tests and recognize #762 as already closed

wrzian:
Recognize #762 as already closed!

This was already fixed by the parser refactor! But I accidentally left an incorrect comment in the `list.typ` test file. I've fixed the comment and also added a test for the enum numbering indent behavior.


wrzian:
In particular, the referenced text now looks like:

<table>
<tr>
<td>

```typ
9. foo
  - bar
10. foo
  - bar

#pagebreak()

- foo
/**/- bar
    /**/- baz

#pagebreak()

- foo
  - bar
/**/    - baz

#pagebreak()

- foo
    - bar
  - baz

#pagebreak()

// Two spaces then two tabs
- foo
  - bar
		- baz
```

</td>
<td>



</td>
</table>

laurmaedje:
Thanks!

PR #5434, opened 2024-11-17
Docs: note `par.linebreaks` in `text.costs`

jbirnick:
Adds a note about `par.linebreaks` to the docs of `text.costs`.

Context:
- This issue in the forum: https://forum.typst.app/t/is-it-possible-to-use-hyphenation-costs-without-justification/1754
- This Discord discussion: https://discord.com/channels/1054443721975922748/1088371867913572452/1307421553054580797


crates/typst-library/src/text/mod.rs
```diff
@@ -497,6 +497,9 @@ pub struct TextElem {
     /// line, so a higher hyphenation cost can result in awkward justification
     /// spacing.
     ///
+    /// Note: Hyphenation costs will only be applied when the
```
laurmaedje, line 500: I think we could fold this into the previous paragraph to keep one paragraph per cost feature. Let's also mention that justify implies linebreaks: optimized.

jbirnick: Done.

laurmaedje:
Thanks!

PR #5451, opened 2024-11-21
Fix sizing of quadratic shapes (square/circle)

### LINKED ISSUE #5287: Sizing a circle/square with `width` or `height` beyond their default width/height is broken in 0.12
### Description

Their layout bounds expand to the configured width/height, but the shape itself doesn't. Notably `size` works and `width` or `height` that is smaller than default also works. `rect` and `ellipse` are also not affected because they don't have the logic that enforces square aspect ratio.

This wasn't probably not caught by the tests because the squares and circles there are all smaller than the default. The fact that nobody noticed this goes to show how much these two functions are used.

Reproduction:
```typ
#square(height: 5cm)
#square(size: 5cm)
```

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

bluebear94:
Fixes #5287.


crates/typst-layout/src/shapes.rs
```diff
@@ -365,6 +365,9 @@ fn layout_shape(
         let default = Size::new(Abs::pt(45.0), Abs::pt(30.0));
         let mut size = region.expand.select(region.size, default.min(region.size));
         if kind.is_quadratic() {
+            if region.expand.any(|ex| *ex) {
+                size = region.size;
+            }
```
PgBiel, line 370: I must admit I'm not too familiar with the surrounding code, but if I understood correctly, the problem is that `expand` will be false for the missing dimension (e.g. `#square(height: 5cm)` is missing width, which will be `auto`, so `expand` will be false on the horizontal axis), in which case, from the `select` above, the height will be kept in `size`, but the width will be set to `default.min(region.size)`, which will be smaller than the height (unless it is smaller than the default width), and thus `size.min_by_side()` will assign the default width as the height as well, leading to the bug. The buggy behavior here (at least conceptually) is precisely that the specified height should actually take priority over the unspecified width, thus it should not be determining the minimum size from both the specified region and the default, only from the specified region. (Unless no dimensions were specified, in which case the default should be used.) Does this check out?

To be honest, it feels a bit unclear to me why `expand` and `min` are being used to detect auto dimensions (where `min` will always return the default when region.size is infinity / auto); in previous code, there was an explicit `unwrap` for `auto` dimensions. I feel like this could be hiding some other bug.

Either way, I think there should be some more comments in the code to make this more explicit, in particular:

1. Indicating that `default.min(region.size)` will return the default dimensions when the region size is infinite (`auto`).
2. Indicating that `size.min_by_side()` should only consider the specified region size when at least one region dimension is fixed (in which case `expand` will be true), ignoring the other dimension set to the default value, if it received that value, to give priority to the specified dimension. This is, in theory, only necessary to check for quadratic shapes as they must be resized to ensure they have equal dimensions. Only the resizing is a problem.
(Though I do wonder if there could be some edge case generating problems with non-quadratic shapes as well, but I could not find one.)

tests/suite/visualize/square.typ
```diff
@@ -144,3 +144,9 @@
 // Test that square sets correct base for its content.
 #set page(height: 80pt)
 #square(width: 40%, rect(width: 60%, height: 80%))
+
+--- square-height ---
```
PgBiel, line 148: Imo there should be more tests, in particular for circle and maybe (at least one of) rect / ellipse (even though those should not exhibit buggy behavior in principle).

Also add a test for missing height (instead of missing width).

PgBiel (commented):
Nice job! While I'm not familiar enough with this part of the code to give an explicit approval (Laurenz will have to confirm), this looks reasonable enough to me in principle.

crates/typst-layout/src/shapes.rs
laurmaedje: > To be honest, it feels a bit unclear to me why expand and min are being used to detect auto dimensions (where min will always return the default when region.size is infinity / auto); in previous code, there was an explicit unwrap for auto dimensions. I feel like this could be hiding some other bug.

Because the sizing now comes from the block around it (which enables fr-sizing). So expand is the only place where the information can come from.

PgBiel: I see. I assume it's impossible for expand to mean anything else here then?

bluebear94:
> With the code
> 
> ```typst
> #set page(height: 100pt)
> #circle(width: 150%)
> ```
> 
> The circle should overflow, but it doesn't.
> 
> I think the code below would fix this:
> 
> ```rust
> // The default size that a shape takes on if it has no child and
> // enough space.
> let default = Size::new(Abs::pt(45.0), Abs::pt(30.0));
> let size = if kind.is_quadratic() {
>     // `region.expand` is true for axes whose size was specified.
>     let length = if region.expand.x && region.expand.y {
>         // If both `width` and `height` are specified, we choose the
>         // smaller one.
>         region.size.x.min(region.size.y)
>     } else if region.expand.x {
>         region.size.x
>     } else if region.expand.y {
>         region.size.y
>     } else {
>         default.min_by_side()
>     };
>     Size::splat(length)
> } else {
>     region.expand.select(region.size, default.min(region.size))
> };
> frame = Frame::soft(size);
> ```
> 
> Not sure if there is a more compact way to express this, but this version would be fairly clear I think.

This contradicts the `square-overflow` test, which expects a square with its width set to an overflowing value to not overflow the page:

```typst
#set page(width: 40pt, height: 25pt, margin: 5pt)
#square(width: 100%)
#square(width: 100%)[Hello there]
```

which now gives

image

laurmaedje:
Looks like I didn't run the tests ... but I'd argue that the test is actually wrong here. The user clearly said `width: 100%` (and not `height: 100%` as well), so we should respect that. However, if we do, we also need to adjust the code path with body from the looks of it.

bluebear94:
> Looks like I didn't run the tests ... but I'd argue that the test is actually wrong here. The user clearly said `width: 100%` (and not `height: 100%` as well), so we should respect that. However, if we do, we also need to adjust the code path with body from the looks of it.

OK, I fixed the offending test to specify `height: 100%` instead of `width: 100%`.

laurmaedje:
What we still need to figure out is how to change the code path with body, so that the two behave consistently. The output of
```typ
#set page(width: 120pt, height: 100pt)
#square(width: 150%)
#square(width: 150%)[Hello there]
```
is now inconsistent.

Side note: I also just noticed that all of this is also broken in 0.11.1 and below ... not sure why I thought it was a regression.

PgBiel:
As an addendum, if I understood correctly, the right thing to do is to fix the code that made the test wrong, instead of removing the test. (We could probably keep both tests - old and new.)

laurmaedje:
The intent of the test _was_ kinda broken, though I think there is still some version of it that makes sense.

laurmaedje:
I fixed the logic so that the tests pass, adjusted the logic to be in line with the no-body branch (extracting a new function), added some more comments, and a few new tests. I think it should be correct now.

tests/suite/visualize/circle.typ
```diff
@@ -67,3 +67,23 @@ Expanded by height.
...
+// Test that sizing a circle beyond the page width correctly overflows the page.
+#set page(height: 100pt)
+#circle(width: 150%)
+
+--- circle-size-beyond-default ---
+// Test that setting a square's height beyond its default sizes the square correctly.
+#circle()
+#circle(height: 60pt)
+#circle(width: 60pt)
+#circle(radius: 30pt)
```
PgBiel, line 89: Something is off here... `circle-size-beyond-default` seems quite similar to `circle-sizing`, while it mentions squares in the description. Is this intentional? Otherwise, one should stay and the other should not.

PgBiel (commented):
Code looks fine, I added a comment for ease of understanding. I have just one small comment about tests.

bluebear94:
> I fixed the logic so that the tests pass, adjusted the logic to be in line with the no-body branch (extracting a new function), added some more comments, and a few new tests. I think it should be correct now.

Didn’t notice the test failures. Thanks for the fix!

tests/suite/visualize/circle.typ
laurmaedje: yeah true, didn't see that circle-sizing test. probably I looked at a single commit diff instead of the whole PR.

bluebear94: I edited the comment to fix this.

laurmaedje:
Thanks!

PR #5452, opened 2024-11-22
Replace _magic_ "preview" literals with a constant

swaits:
Converts all instances of `"preview"` to `DEFAULT_NAMESPACE` which is defined in `typst-kit/src/package.rs`.

### DIFF
--- a/crates/typst-ide/src/tests.rs
+++ b/crates/typst-ide/src/tests.rs
@@ -119,7 +119,10 @@ impl IdeWorld for TestWorld {
     fn packages(&self) -> &[(PackageSpec, Option<EcoString>)] {
         const LIST: &[(PackageSpec, Option<EcoString>)] = &[(
             PackageSpec {
-                // NOTE: DEFAULT_NAMESPACE (i.e. "preview") in typst-kit/src/package.rs
+                // NOTE: This literal, `"preview"`, should match the const, `DEFAULT_NAMESPACE`,
+                // defined in `crates/typst-kit/src/package.rs`. However, we should always use the
+                // literal here, not `DEFAULT_NAMESPACE`, so that this test fails if its value
+                // changes in an unexpected way.
                 namespace: EcoString::inline("preview"),
                 name: EcoString::inline("example"),
                 version: PackageVersion { major: 0, minor: 1, patch: 0 },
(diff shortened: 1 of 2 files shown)


crates/typst-ide/src/tests.rs
```diff
@@ -119,6 +119,7 @@ impl IdeWorld for TestWorld {
     fn packages(&self) -> &[(PackageSpec, Option<EcoString>)] {
         const LIST: &[(PackageSpec, Option<EcoString>)] = &[(
             PackageSpec {
+                // NOTE: DEFAULT_NAMESPACE (i.e. "preview") in typst-kit/src/package.rs
                 namespace: EcoString::inline("preview"),
```
PgBiel, line 123: Ideally we'd just use the constant here. Maybe it can be moved to `typst-utils` if it's shared by multiple Typst crates, but dunno (cc @laurmaedje)

PgBiel (commented):
Thanks! `DEFAULT_NAMESPACE` alone doesn't feel right to me at first, but maybe it's less bad since it's scoped to the `package` module, and `package::DEFAULT_NAMESPACE` is a bit clearer than otherwise. Still, maybe it can be changed to something like `DEFAULT_EXTERNAL_NAMESPACE`, to differentiate from local namespaces (of which the most common is `@local`, but others are also possible)? Something to think about.

swaits:
I went around and around on the name. 

I also considered `PUBLIC_NAMESPACE` but landed on using default as it matches the const you have for the public repository. 

Please let me know your preference. I'll update the PR once I hear back.

crates/typst-ide/src/tests.rs
swaits: My reasoning: the test exists to test it. 😇 

In other words, imagine we decide to promote "preview" to "public". But when we change the const we name it "puvlic", with a typo. 

In this case, if the test used the const directly, the test passes. If not, the test fails (unless we make the same typo twice). 

Let me know what you think. Glad to update it as you like, just wanted to share my thought process first.

PgBiel: Ah, I see. This seems like robust reasoning to me.

Maybe we could add a sentence to the comment saying "Keep as a literal so the test fails if it changes."

swaits: I updated the comment to reflect this thinking.

swaits:
@PgBiel, please let me know if you want something other than `DEFAULT_NAMESPACE`.

laurmaedje:
I'm fine with `DEFAULT_NAMESPACE`. Thanks!

PR #5458, opened 2024-11-24
Tags shouldn't affect row height in equations

mkorje:
| Before | After |
| - | - |
| old | new |

Result of the Typst code below before and after this change.

```typ
#box($ - - $, fill: silver)
#box($ #metadata(none) - - $, fill: silver) \
#box($ a \ - - $, fill: silver)
#box($ a \ #metadata(none) - - $, fill: silver)
#box($ - - \ a $, fill: silver)
#box($ #metadata(none) - - \ a $, fill: silver)
```


laurmaedje:
Thanks!

Issue #5422, opened 2024-11-15
Different appearance of `vec` and `mat`


carlyu99:
### Description

The same column vector implemented with `vec` and `mat` looks different.
The `vec` version has smaller spacing, and symbols overlap.
Below are code and screenshot.

```typ
#set text(size: 40pt)
#show math.equation: set text(font: "Libertinus Math", features: ("ssty",))
$
  vec(z_(n_p), a^2)
  mat(z_(n_p); a^2)
  vec(z_(n_p), a^2, delim: "[")
  mat(z_(n_p); a^2; delim: "[")
$
```

![image](https://github.com/user-attachments/assets/24aa9a4b-a8bb-4db5-98fe-9a423c374f14)

### Reproduction URL

https://typst.app/project/ru4m2Y6fh1an9EO68kXxF7

### Operating system

Web app, Windows

### Typst version

- [X] I am using the latest version of Typst

PR #5459, opened 2024-11-24
Fix multiline annotations in over- elems in math changing the baseline

### LINKED ISSUE #5426: Multi-line overbraces push annotated terms out of alignment with equation
### Description

2+ line annotations push the math content down, out of line from the rest of the equation. Example:

```typ
$ S = overbrace(beta (alpha) S I, "one line") - overbrace(mu (N), "two" \  "line") $
```

The `µ(N)` term is out of line with the rest of the equation. This does not happen in the reverse direction with underbraces:

```typ
$ S = underbrace(beta (alpha) S I, "one line") - underbrace(mu (N), "two" \  "line") $
```

I asked about this in the [forum](https://forum.typst.app/t/how-to-avoid-pushing-down-math-content-with-multi-line-overbrace/1740) and got a fix to this behavior, but I suspect this is not intended behavior.

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

mkorje:
Fixes #5426.

Note: how `vec` is layouted when it has elements with linebreaks has changed. Though this change is for the better imo, as it is more consistent with matrices.

| Before | After |
| - | - |
| old | new |
```typ

$ vec(1, c + d \ e + f, 3)
  mat(1; c + d \ e + f; 3) $
```


EpicEricEE:
I think there should be some kind of feedback then (e.g. a warning) instead of silently ignoring any line breaks, but maybe that's out of scope here.

PgBiel:
I'm not sure I'm too comfortable with ignoring the linebreaks either. If it's for consistency, maybe the change should happen in the opposite direction (fix linebreaks where they're already broken instead of the other way around).

At the very least, this should be reflected in tests.

mkorje:
I agree ignoring linebreaks isn't great (though it seems a little cursed to have a linebreak inside an element in a mat/vec).

Vectors probably shouldn't be using the `stack` function to be layouted anyway (?). I think we should change vectors to use the same layout method as matrices, which would resolve #5422. I'll have a look at why the linebreaks are being ignored in matrices. Adding these changes to this PR is maybe a bit out of scope though.

mkorje:
Ah, I just realised the linebreaks are probably ignored because there would otherwise be issues with alignment points. We wouldn't be able to differentiate alignment points for the whole matrix from ones for the specific element. Maybe, as @EpicEricEE suggested, a warning would be best?

laurmaedje:
The same happens for `cases` and I think some people write `cases(a \ b \ c)` rather than using commas. Not saying it's necessarily great, but I'm uncomfortable with merging this as-is, without unifying vec & mat first and potentially displaying some warnings.

mkorje:
I've updated this PR to keep the current behaviour and added tests reflecting the current situation for cases, matrices, and vectors. The linked issue has still been fixed, and this is now the only (user-facing) change made by this PR.

I'll address the inconsistency between vectors (and cases) and matrices with regards to linebreaks when unifying vectors and matrices.

laurmaedje:
Thanks!

PR #5462, opened 2024-11-24
Document integer literal parsing

### LINKED ISSUE #2284: Min integer value gets parsed as a float
### Description

According to the Typst documentation https://typst.app/docs/reference/foundations/int/ the smallest integer is `-9223372036854775808` but when using the literal for that number it gets parsed as a `float` instead of an `integer`. The value can still be obtained through arithmetic:

```typ
#let literal_min_int = -9223372036854775808  // parsed as float
#let real_min_int = -9223372036854775807 - 1

#type(literal_min_int) // float

#type(real_min_int) // integer
```

### Reproduction URL

https://typst.app/project/rb52J82NefdZ4oTOE3ODK_

### Operating system

Web app, Linux

### Typst version

jakobjpeters:
Closes #2284

### DIFF
--- a/crates/typst-library/src/foundations/int.rs
+++ b/crates/typst-library/src/foundations/int.rs
@@ -11,10 +11,10 @@ use crate::foundations::{
 ///
 /// The number can be negative, zero, or positive. As Typst uses 64 bits to
 /// store integers, integers cannot be smaller than `{-9223372036854775808}` or
-/// larger than `{9223372036854775807}`. Integer literals are only positive, so
-/// a negative integer such as `{-1}` is semantically the negation `-` of the
-/// positive literal `1`. A positive integer greater than the maximum value
-/// and a negative integer less than or equal to the minimum value cannot be
+/// larger than `{9223372036854775807}`. Integer literals are always positive,
+/// so a negative integer such as `{-1}` is semantically the negation `-` of the
+/// positive literal `1`. A positive integer greater than the maximum value and
+/// a negative integer less than or equal to the minimum value cannot be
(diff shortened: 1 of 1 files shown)


PgBiel:
To be honest, besides the documentation addition, I'd also prefer to add a warning when using a literal that is out of bounds. This behavior seems too unpredictable to me, as is.

jakobjpeters:
It looks like it happens [here](https://github.com/typst/typst/blob/5c37a1cfea2a08b08f4af34ba867748fd90c687e/crates/typst-syntax/src/lexer.rs#L745-L756). Can you give a hint about how to implement such a warning? Or alternatively, produce an error suggesting the user to use a `float` instead.

wrzian:
We don't yet have any real infrastructure for adding warnings to the parser (I ran into this as well in #5229). We do have error infrastructure, so we could also just start by erroring on out-of-bounds literals. If we do want warnings, then there are many ways we could do so.

One way to add parser warnings may be to add a `Warning` variant to the `SyntaxKind` enum and extend the options for `SyntaxNode::Repr` to include warnings and an interface for creating them. Warning nodes could then transparently wrap an underlying `SyntaxNode` plus the warning and any hints, and we would expose a new set of functions from `SyntaxNode` to access them.

However, I would recommend that we discuss if this is the right design before starting to implement it. Best would be to join the contributors channel on Discord, but we could also continue here (maybe in a new issue).

And if you want any help talking through the parser infrastructure, let me know!

laurmaedje:
Thanks!

Since implementing a warning is a little more complex, let's not block this on it.

Issue #5464, opened 2024-11-25
Math kerning isn't applied when under/overlined characters are the base


mkorje:
### Description

An example:
```typ
#set page(width: auto, height: auto, margin: 1em)

$ V_y != overline(V)_y != underline(V)_y $
```
![test](https://github.com/user-attachments/assets/8685d9ee-24f9-49c1-a66a-e46685740562)

On reflection, #4695 probably wasn't done the best way, relying on text_like-ness...

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

PR #4695, opened 2024-08-06
Make under/overlines behave the same as accents in attach base

mkorje:
This PR makes it so that the content's `italics_corrections` and `is_text_like` are passed onto the finished frame fragment with an under/overline. When an under/overline element is the base of an attachment, this brings it more in line with the behaviour of accents.

Whether this is the 'best' thing to do, I don't know. I guess it depends on which of the two people prefer most. Or, if this just shouldn't be the approach taken to this at all. I find that with the underline, the bl and br attachments seem to high, despite it being more 'consistent'.

Example 4 shows the br attachment colliding with the underline. So it seems passing `italics_corrections` all the time isn't ideal either...

<details>
<summary>Example 1</summary>

```typst
#set page(width: 1.5cm, height: 1.2cm)

$ attach(y, t: i, b: i, tl: i, bl: i, tr: i, br: i)
  attach(macron(y), t: i, b: i, tl: i, bl: i, tr: i, br: i)
  attach(overline(y), t: i, b: i, tl: i, bl: i, tr: i, br: i) $
```
Using the above typst code:
<details>
<summary>Before</summary>

overunderline_old
</details>

<details>
<summary>After</summary>

overunderline_new
</details>

<details>
<summary>Diff</summary>

overunderline_old
</details>

And another comparison https://www.textcompare.org/image/?id=66b24dbd116d31c8d023e3a7.
</details>

<details>
<summary>Example 2</summary>

```typst
#set page(width: 2cm, height: 1.2cm)

$ attach(overline(underline(y)), t: i, b: i, tl: i, bl: i, tr: i, br: i)
  attach(underline(y), t: i, b: i, tl: i, bl: i, tr: i, br: i)
  attach(overline(y), t: i, b: i, tl: i, bl: i, tr: i, br: i)
  attach(underline(overline(y)), t: i, b: i, tl: i, bl: i, tr: i, br: i) $
```
Using the above typst code:
<details>
<summary>Before</summary>

overunderline_old
</details>

<details>
<summary>After</summary>

overunderline_new
</details>

<details>
<summary>Diff</summary>

overunderline_old
</details>

And another comparison https://www.textcompare.org/image/?id=66b25137116d31c8d023e3b9.
</details>

<details>
<summary>Example 3</summary>

overunderline_old
</details>

<details>
<summary>Example 4 (Collision with subscript!)</summary>

overunderline_old
</details>

<details>
<summary>Example 4 (Again, but not passing italics_corrections)</summary>

overunderline_old
</details>

[...]

PR #5465, opened 2024-11-25
Update NewCM fonts to version 7.0.0

### LINKED ISSUE #5313: If a font supports ligatures, 'smallcaps' doesn't seem to affect them
### Description

If a font supports ligatures, `smallcaps` doesn't seem to affect them.

```typ
#set text(font: "New Computer Modern")

#smallcaps[Before & After]
```

image

If I understand correctly, if a font doesn't support ligatures, this issue doesn't occur:

```typ
// #set text(font: "New Computer Modern")  // We use the default Typst font instead

#smallcaps[Before & After]
```

image

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

mkorje:
~~https://github.com/typst/typst-assets/pull/5~~ Merged.

Tests needed to be updated as a result of two changes to the font: the addition of math kerning and the size of the ⅀ glyph (the smaller variant, used in inline math) being corrected.

The weight of the bold math font was fixed, so specifying it in the font exceptions is no longer needed.

I considered changing the math kerning tests to use NewCM now instead of STIX Two, but it hasn't got any (I'm assuming based on none of the glyphs in the tests having any) bottom-left or top-left kerning info just yet.

While updating one of the tests I uncovered a bug from #4695: math kerning isn't being applied when the base is under- or overlined. I've opened an issue for this: #5464.

Closes #5313.


laurmaedje:
Thank you!

PR #5473, opened 2024-11-26
Ignore leading and trailing ignorant fragments in `math.lr`

### LINKED ISSUE #4433: closing parenthesis is not recognised in math formula in context block

mkorje:
Fixes #4433.

The code here is basically the same as in #4544, thanks to @bluebear94 et al.

Regarding the discussion there, I don't mean to undermine anyone with this PR. I just want to fix the linked issue, and I share the (subjective) preference for the iterator style code. Since the ACP for `slice::trim...` was accepted, we can always change to that when it gets implemented and stabilized.


laurmaedje:
Thanks!

PR #2950, opened 2023-12-14
Ignore weak HElem at the beginning and end of an LR group

Leedehai:
Fixing #2515.

Demo:
```typ
$ integral_S f(x) mu(dif x), norm(dif x) $
```

Before this change
Screenshot 2023-12-14 at 00 41 38

After this change
Screenshot 2023-12-14 at 00 39 32

PR #5477, opened 2024-11-27
Fix weak spacing being ignored unconditionally in `math.lr`

mkorje:
Fixes a bug in #2950: weak spacing fragments were being removed unconditionally, regardless of whether there was actually an opening/closing delimiter in the arguments passed.

| Before | After |
| - | - |
| old | new |

Result of the Typst code below before and after this change.

```typ
--- math-lr-weak-spacing ---
// Test ignoring weak spacing immediately after the opening
// and immediately before the closing.
$ [#h(1em, weak: true)A(dif x, f(x) dif x)sum#h(1em, weak: true)] $
$ lr(\[#h(1em, weak: true)lr(A dif x, f(x) dif x\))sum#h(1em, weak:true)a) $
```


crates/typst-layout/src/math/lr.rs
```diff
@@ -74,10 +74,18 @@ pub fn layout_lr(
     // Remove weak SpacingFragment immediately after the opening or immediately
     // before the closing.
     let mut index = 0;
+    let opening_exists = inner_fragments
+        .first()
+        .map_or(false, |f| f.class() == MathClass::Opening);
```
laurmaedje, line 79: could use `is_some_and` or `.first().map(MathFragment::class) == Some(MathClass::Opening)`

crates/typst-layout/src/math/lr.rs
```diff
@@ -62,7 +62,7 @@ pub fn layout_lr(
     }
 
     // Handle MathFragment::Variant fragments that should be scaled up.
-    for fragment in inner_fragments {
+    for fragment in &mut *inner_fragments {
```
laurmaedje, line 65: why this change?

mkorje: To make the borrow checker happy

laurmaedje:
Thanks!

PR #5480, opened 2024-11-27
Add support for interpreting f32 in float.{from-bytes, to-bytes}

### LINKED ISSUE #5307: Allow `float.from-bytes` to parse other sizes of floats
https://github.com/typst/typst/blob/b8034a343831e8609aec2ec81eb7eeda57aa5d81/crates/typst/src/foundations/float.rs#L113-L141

I'm trying to write a parser for the mzML language that mass spectrometrists use, however, some fields are base64 encoded arrays of floats (both float64 and float32). Unfortunately, `float.from-bytes` can only parse float64.

bluebear94:
Fixes #5307.


laurmaedje:
Looks good! Thanks.

PR #5481, opened 2024-11-27
Let decimal constructor accept decimal values

### LINKED ISSUE #5427: Let decimal constructor accept decimal values
### Description

Most constructors for basic types accept the target type itself in addition to building from/converting other types: `int(3)`, `float(3.3)`, `str("x")`, `array((1, 2))`, etc. however, it's not supported for `decimal`:
```typ
#decimal(decimal(3))  // expected integer, float, or string, found decimal
```
It would be nice to fix that inconsistency.

### Use Case

With this, the `decimal` constructor could be used to ensure a uniform type when the entry points to an API are open to different kinds of number input:
```typ
#let process-number(number-input) = {
  let number = decimal(number-input)
  // ...
}

#process-number("15")  // ok
#process-number(15)  // ok
#process-number(decimal(15))  // should also be ok
```

bluebear94:
Fixes #5427.

### DIFF
--- a/crates/typst-library/src/foundations/decimal.rs
+++ b/crates/typst-library/src/foundations/decimal.rs
@@ -317,6 +317,7 @@ impl Decimal {
                     })
                     .at(value.span)
             }
+            ToDecimal::Decimal(decimal) => Ok(decimal),
         }
     }
 }
@@ -435,6 +436,7 @@ pub enum ToDecimal {
     Int(i64),
     /// A float to be converted to the equivalent decimal.
     Float(f64),
+    Decimal(Decimal),
 }
 
 cast! {
@@ -442,6 +444,7 @@ cast! {
     v: i64 => Self::Int(v),
     v: f64 => Self::Float(v),
     v: Str => Self::Str(EcoString::from(v)),
+    v: Decimal => Self::Decimal(v),
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/foundations/decimal.rs
```diff
@@ -435,13 +436,15 @@ pub enum ToDecimal {
     Int(i64),
     /// A float to be converted to the equivalent decimal.
     Float(f64),
+    Decimal(Decimal),
```
laurmaedje, line 439: could reorder this enum and the `cast!` to be consistent with `ToInt` and `ToFloat`? (and add a short doc comment just for consistency)

the ordering affects the error messages, so it'd be nice if everything would be consistent.

bluebear94: Done. Should the cast also accept bools like `ToInt` and `ToFloat`?

PgBiel: While I find it odd to begin with, it does seem that it would be consistent indeed.

laurmaedje: For int at least it makes sense I think. And the rest is consistency. In Rust maybe more since it's a systems language but 1 for true and 0 for false are kind of universal.

laurmaedje:
Thanks!

PR #3059, opened 2023-12-23
Improve positioning of multiple accents and attachments

EpicEricEE:
Fixes #1672 

The accent attachment position and italics correction is now stored in the fragment, so that it can be preserved when new elements create a frame fragment around it. This happens in `accent`, `cancel` and `op` elements, thus stacked accents are now properly aligned on top of each other.

In #1129, the positioning of attachments for character boxes was fixed. However, that only worked for "pure" characters, so `accent` or `cancel` elements were not considered as character boxes. To fix that, I introduced the `text_like` property for each math fragment, which replaces the `is_character_box` check. This property is set to true for

- frame fragments created in `layout_text`
- glyph and variant fragments if they aren't of class "large"
  _(to keep current behavior of attachments for large operators like `sum` and `integral`)_
- fragments created in `accent`, `cancel` and `op` if the underlying fragment is text like

A more difficult contender is `MathRow` as there is no single base fragment from where the property values can be inherited.

There are some small differences, as previously all "atomic text frames" were character boxes, whereas now that is only the case for frame fragments with `text_like` set to true. This means that when wrapping a text-like fragment in a box or similar, the properties are not preserved. For example:
```typ
$ #box[$f$]_a^a != f_a^a $
```
renders as _(left: old behavior, right: new behavior)_

image

As can be seen, the boxed version is no longer considered a character box as the `text_like` property is lost with the `box`, thus the top attachment is shifted up more.

Maybe this solution is a bit too naive, but until a bigger refactor is planned, it should be alright.

PR #1129, opened 2023-05-07
Adjust attach() per TeXbook Appendix G 18a

Leedehai:
Fixes issue #539.

> 18a. If the translation of the nucleus is a character box, possibly followed by a kern, _set u and v equal to zero_; otherwise set u ← h − q and v ← d + r

Some nuances shown below indicate the patch makes Typst more aligned with LaTeX:
LaTeX:
```
$$ b^2 \sqrt{b^2}, y_1^1 \lg_1^1, (by)_2^2 \neq \left(by\right)_2^2 $$
```
Typst:
```
$ b^2 sqrt(b^2), y_1^1 lg_1^1, "("b y")"_2^2 != (b y)_2^2 $
```
| | image |
|:--|:--|
|LaTeX|Screenshot 2023-05-06 at 22 42 46|
|Typst with this patch|Screenshot 2023-05-06 at 22 44 02|
|Typst without this patch|Screenshot 2023-05-06 at 22 50 11|

PR #5483, opened 2024-11-27
Use extended shape information for glyph `text_like`-ness

### LINKED ISSUE #2342: Scripts are sometimes in the wrong position with `lr()`
### Description

The following
```typst
$
lr(|,size: #3em)_a^b
$
```
produces
image

The scripts are in the wrong place.

If I put anything else in there, even a zws, they appear in the right place

```typst
$
lr(zws|,size: #3em)_a^b
$
```
image

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

mkorje:
Fixes #2342.

This changes how the `text_like` property is set for glyph and variant fragments. Instead of checking if they aren't of class "large", we use the [extended shape](https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#extended-shape) information in the font and check if they aren't an extended shape.

According to [this](https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#superscriptshiftup-superscriptshiftupcramped-superscriptbottommin-superscriptbaselinedropmax), if the base is a single character and not an extended shape, then the baseline of the script is shifted by a font constant. Otherwise, the script is shifted above/below the base by a font constant. I've retained the behaviour introduced in #4545, where sequences of text-like stuff use the constant script shift. Thus the current Typst behaviour now sits somewhere between TeX engines (which don't use extended shape information at all) and Microsoft's implementation (which follows the algorithm from above). I think this is a fine state to be in, given we find the behaviour in #4545 desirable. Also, the continued use of the `text_like` property, introduced in #3059 (which superseded #1129), lets us preserve it through `accent`, `cancel`, and `op` elements.


laurmaedje:
Thanks for the deep investigation and fix!

PR #5492, opened 2024-11-28
Make `math.italic` compatible with ħ

MDLC01:
U+210F ℏ PLANCK CONSTANT OVER TWO PI (present in `sym` as `planck.reduce`) seems to be the mathematical italic counterpart of U+0127 ħ LATIN SMALL LETTER H WITH STROKE, as suggested in [the chart](https://unicode.org/charts/PDF/U2100.pdf):
ℏ ≈ <font> 0127 ħ latin small letter h with stroke


laurmaedje:
Thanks!

Issue #5435, opened 2024-11-17
Entered unreachable code after changing indentation on lists


tshort:
### Description

I don't have a minimal example that reproduces this. It happens with a 15-chapter document. The change that triggered the error was:

```typ
  set list(indent: 1em)
```

It's somehow length related. If I exclude the last chapter or the first two chapters, it doesn't happen. 

Here is a stack trace:

```
thread '<unnamed>' panicked at /Users/runner/.cargo/git/checkouts/typst-15b90983aa0184a9/58426a9/crates/typst/src/layout/flow/compose.rs:182:41:
internal error: entered unreachable code
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: typst::layout::flow::compose::Composer::column
   4: typst::layout::flow::layout_flow
   5: typst::layout::pages::run::layout_page_run_impl
   6: typst::layout::pages::run::layout_page_run
   7: rayon::iter::plumbing::bridge_producer_consumer::helper
   8: rayon_core::join::join_context::{{closure}}
   9: rayon::iter::plumbing::bridge_producer_consumer::helper
  10: <rayon_core::job::StackJob<L,F,R> as rayon_core::job::Job>::execute
  11: rayon_core::registry::WorkerThread::wait_until_cold
  12: rayon_core::registry::ThreadBuilder::run
```

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

PgBiel:
@tshort Does your document contain footnotes, and if so, do they happen to contain - possibly, but not necessarily, large - block-level elements (e.g. lists, images, or basically anything that isn't a paragraph)?

---

Taking a look at the relevant code below (in the 0.12 commit so the line numbers and everything else match up):

https://github.com/typst/typst/blob/737895d769188f6fc154523e67a9102bc24c872e/crates/typst/src/layout/flow/compose.rs#L180-L182

which points to `column_contents`, which calls `footnote`, `float` and `distribute`, it _appears_ that overlarge footnote entries being moved to the next page are the only way to trigger a `Finish` here:

https://github.com/typst/typst/blob/737895d769188f6fc154523e67a9102bc24c872e/crates/typst/src/layout/flow/compose.rs#L438-L449

given that `float` seems to only return `Stop::Relayout`, and `distribute` explicitly captures `Stop::Finish`:

[...]

PR #5497, opened 2024-11-29
Forbid footnote migration in pending floats

PgBiel:
Fixes https://github.com/typst/typst/issues/5435

As seen in the issue, a footnote processed in a pending float (that was moved from a previous page because it didn't fit)  could cause a panic if it didn't have enough space, as then it would prompt migration, returning `Stop::Finish`, which leads to an `unreachable!` branch in `column`.

Therefore, having noticed that pending footnotes being processed were already receiving `migratable` set to `false`, I simply added a parameter to `float` to indicate whether footnotes processed in it should be migratable as well, which is `false` when processing pending floats, but `true` otherwise (in the distributor, which can handle `Stop::Finish`).

**However,** I'm not fully sure if this is the best approach, as it ensures the footnote invariant (entry page = footnote reference page) is broken in that case. Should we try to somehow force the float back into the queue if a footnote asked for migration? Or maybe it's too late now and this could cause an infinite loop? (**Edit:** Well, it certainly could if the float and footnote keep running out of space, though we do already check for `regions.may_progress()` in `composer.footnotes()`. But we could also just not care for now and accept that at least the crash is fixed. Would still like to hear your thoughts on this though.)

Additionally, I adjusted some docs. Besides fixing some typos, I tried to add some explanations, most notably to `column_contents` to indicate why we are forbidding footnote migration for pending footnotes and floats. Please confirm whether the reason I wrote makes sense.


laurmaedje:
> However, I'm not fully sure if this is the best approach, as it ensures the footnote invariant (entry page = footnote reference page) is broken in that case. Should we try to somehow force the float back into the queue if a footnote asked for migration

I think this fix is good for now. There are a few other scenarios where the footnote invariant is broken, so it's not too big of a deal and the case is very rare.

PR #5498, opened 2024-11-30
Fix infinite loop with footnote which never fits

### LINKED ISSUE #5496: Impossibly large footnotes leading to seemingly infinite execution
### Description

With the following document:

```js
#set page(width: 400pt, height: 180pt)

#footnote(text(size: 170pt)[a] * 100)
```

The compiler will keep compiling seemingly forever (until it runs out of memory). Happens on the released 0.12 version as well as on main (89d96c62).

I believe this is related to https://github.com/typst/typst/issues/5354. PR https://github.com/typst/typst/pull/5364 fixed that particular case (where the footnote is fully empty, so it would never _not_ be empty in a certain page). But it doesn't seem to be sufficient here, as there is _always_ a non-empty frame since `line + separator + clearance` is larger than the whole page (180pt), triggering the line to always move to the next page since it never fits, so Typst keeps trying to migrate the footnote to the next page forever. (Note that replacing `170pt` with a smaller value, such as 120pt, compiles fine.)

I can confirm this is a 0.12 regression.

### Reproduction URL

_No response_

### Operating system

Web app, Linux

### Typst version

PgBiel:
Fixes #5496

Don't try to queue or migrate the footnote if `regions.may_progress()` doesn't hold, which is usually when the footnote is already alone in the page after having skipped pages where it doesn't fit. Therefore, we assume it will probably never fit if `regions.may_progress()` is false.

However, I'm not entirely sure if we should rely on the `regions` given to the footnote at this point. I can't really think of a good counter-example now, but we can see here that, in the previous (working, for this use case) implementation in 0.11.1, we pinned `in_last` (more or less equivalent to the `may_progress()` check here) right before subtracting the height of the frame with the footnote from the region - before even processing any footnotes in that frame:
https://github.com/typst/typst/blob/5011510270c2c23f0ab019af486b26db0d62261b/crates/typst/src/layout/flow.rs#L399-L414

So this makes me ponder about two things:

1. Should we always forbid requeuing footnotes which were already queued? I think that would be a worse heuristic since it might fit later. Though forbidding migration (as is already done) of course makes sense since the footnote is probably already detached from the origin frame at this point, so no changes there.
2. Should we call `regions.may_progress()` at the caller of `composer.footnote` instead of using the `regions` available to the footnotes at this point, like we were doing in 0.11.1?
     - Note that this doesn't make a difference when processing pending footnotes, since we clone the `regions` for each of them, so the `regions` they receive is precisely the same as the caller's in that particular case:
        https://github.com/typst/typst/blob/76c24ee6e35715cd14bb892d7b6b8d775c680bf7/crates/typst-layout/src/flow/compose.rs#L218-L221
        So **this would only concern `composer.footnotes()`** which subtracts the height of each footnote from the region height.
     - The main difference is that **already laid out footnotes would no longer influence on whether upcoming footnotes can consider being queued (again) or not.** But **I think that's worse** than this PR's implementation, since `regions.may_progress()` will become `true` if we already laid out a footnote before this one, which makes sense as this next footnote, even if too large, is no longer alone in the page, so it may be worth trying again in the next page as the lack of space may be due to the previous footnote and _not_ because the footnote will never fit.

**EDIT (2024-12-07):** I added a test for when the _separator_ never fits instead of the footnote itself, which also causes an infinite loop in Typst 0.12, and it was also fixed by this PR. This suggests the `regions.may_progress()` check is robust enough as it ignores the separator, so it will definitely return `false` at some point if it never works out, since at some point we will have skipped all of the document body and have an empty page.

I considered whether something like grid's `in_last_with_offset` would make sense, which is basically `may_progress()` but reducing the available height by some constant because some content is always added to a new region, but - as emphasized above - since our check happens on the region _without_ the separator, I deemed it to not be necessary, in contrast to `grid` where the header is automatically added when a new region starts (changing its size from the initial value of `height`, and thus triggering `may_progress()` to always return false unless we subtract the header height).


wznmickey:
Yes. The previous solution is just a "hacky" way. I am wondering whether we can save the input frame . If they are the same, then this program will loop forever. I am not familiar with the frame generation so I am not sure whether we could identify whether a frame is the same as another frame.
It could solve both empty frame and oversizing frames as well as other cases which could lead to loop together.

PgBiel:
> If they are the same, then this program will loop forever. I am not familiar with the frame generation so I am not sure whether we could identify whether a frame is the same as another frame.

It's not true that the program will loop forever if we try to lay out the same footnote again, even in the same page. I discuss this here in the PR description:

> Should we always forbid requeuing footnotes which were already queued? I think that would be a worse heuristic since it might fit later.

It can still fit in a later (different) page, so it'd be too restrictive.

So this PR checks `regions` as that's what lets us know when we've actually tried the most we could: when we're at the last page so far and it is empty. Then, trying again would only lead us to creating a new empty page, and then another, over and over, so we just give up early in that case.

PgBiel:
Right. After taking a look, `regions.may_progress()` is not always very accurate **for queueing**. This is because we subtract the flow need from the entry pod's available height, but the flow need appears to not be considered in the input `regions`, which is particularly noticeable when laying out floats, since, for floats, that's just the regions they were given before inserting themselves in it. So here we were unnecessarily forcing the entry to be laid out even though it, by itself, could be considering that the pod itself may progress given the subtracted flow need that was restricting the available space further than the parent region's size, alone, would. And indeed, this means that the presence of any flow need indicates that the region may actually progress even if the input regions doesn't indicate so: just consider that the flow need for a queued footnote is later set to zero, assuming the original frame isn't migrated, so either way we break the infinite loop.

In the case of migration, `regions.may_progress()` is actually completely relevant since what matters _is_ the original `regions`. Migrating the original frame doesn't make sense if the original frame would be placed in a region which can't progress. Therefore, for that branch in specific, `flow_need` is not considered.

This fixes the existing test and updates the tests in PR; most overlarge footnotes are being placed later than before now, as Typst will try again until the footnotes were actually queued and there is zero flow need.

tests/ref/issue-5496-footnote-never-fits.png
laurmaedje, line 1: why are there two empty pages here now? that seems wrong. we didn't get anything out of these extra empty pages.

PgBiel: I think it's a combination of footnote frames / spill and the extra queueing condition. From my testing, the following events happen:

1. In the first page, the footnote's first frame is empty as it is too large. Here, `regions.may_progress()` is false so migration (correctly) doesn't occur, but `regions.may_progress() || !flow_need.is_zero()` (the latter condition being introduced in my latest commit) is true, so the footnote is **queued** (this didn't happen before my latest commit). That is, we try again because what the footnote is attached to may no longer be there in the next page.
2. In the second page, even though the footnote is still too large, `regions.may_progress()` and `!flow_need.is_zero()` are both false, so we don't have a choice but to begin laying out the footnote. The footnote entry is laid out with the following frames:
     1. Empty (because it's too large)
     2. A start tag of a "styled" element
     3. The superscript "1"
     4. "aaaaaa...aaa" (the body).

3. Each frame is laid out in a separate page as expected.

So we get 5 pages in total. To be honest, I think the new behavior is correct (`flow_need` not being zero likely means queueing the footnote might give us more space). It also correctly fixes the "empty separator" in the first page when using floats.

Maybe the largest problem are those two frames at the start of the entry being laid out separately. Why is the tag separate from the superscript "1", if it occupies no space? Maybe it's just that it only figured out that it needed more space when trying to lay out the "1".

Also, the fact that an empty frame is being laid out separately didn't exist before. But that's because, before this PR, any leading empty frame would cause a relayout, but here we're studying a case where that will never converge.

Perhaps we could improve this, then, by skipping the leading empty frame if we decide to go on anyways? That'd save us one page.

PgBiel: Update: The page with a single tag is because of the indent at the start of footnote entry. The footnote entry is laid out as `#h(indent) #number #h(gap, weak: true) #body` - that is, in a single line - so, since the body is so large that it never fits, the line is broken down into one page per word (except for the weak spacing).

Not sure if there is much that can be done since the spacing is weak. Not using a separate element for spacing would work, but using something like padding wouldn't be ideal as the second line onwards shouldn't be indented. So we could consider just keeping it as is for now... I think skipping the empty frame would generate inconsistencies regarding introspection, which doesn´t sound good.

laurmaedje: I see. While not ideal, that seems like a separate and not that urgent problem. Maybe we should just rewrite the test to not trigger such a case? As is, the test is a little uninstructive, which will be confusing for people who work on this part of the code in the future.

PgBiel: Alright, I've set indent to 0pt which should help for now.

Issue #5499, opened 2024-11-30
Blocks with clipping interferes with text fill color transparentize


diquah:
### Description

Blocks with clipping enabled messes up transparency, but only fill text fill colors. Strangely, rect fill seems to be unaffected.

![image](https://github.com/user-attachments/assets/0c848eb2-1eda-4006-a51f-8128342ca85e)

```typst
#set page(width: 2in, height: 2.1in)

#block(clip: false, {
  text(fill: blue.darken(20%).transparentize(50%), "Hello, World")
})
#block(clip: true, {
  text(fill: blue.darken(20%).transparentize(50%), "Hello, World")
})

#block(clip: false, {
  rect(fill: blue.darken(20%).transparentize(50%))
})
#block(clip: true, {
  rect(fill: blue.darken(20%).transparentize(50%))
})
```

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

PgBiel:
Works fine on the exported PDF, as well as on SVG export, so it's specific to PNG export.

<details><summary>Exported PDF output</summary>
<p>

![correct colors](https://github.com/user-attachments/assets/906cd7c6-476b-4027-818d-3f92f4005ab8)

</p>
</details>

PgBiel:
Interestingly, adding `#set text(stroke: 0pt)` (i.e. adding any stroke, even if it's not visible at all) fixes it.

PgBiel:
Given the above, I suspect the problem is in either `render_outline_glyph` or `write_bitmap` of `typst-render/src/text.rs` (I'd bet on the second function):

https://github.com/typst/typst/blob/76c24ee6e35715cd14bb892d7b6b8d775c680bf7/crates/typst-render/src/text.rs#L40
https://github.com/typst/typst/blob/76c24ee6e35715cd14bb892d7b6b8d775c680bf7/crates/typst-render/src/text.rs#L158

This is because

1. The second function isn't called at all if text has a stroke. Then it is rendered as a path.
2. The second function has a conditional for the presence of a mask (which I believe is implied by `clip: true` in a block).

So it's possible the problem is somewhere in the first branch in that second function.

While I'm not very knowledgeable on this part of the repository, I don't see anything too weird popping out to me after analyzing the code (and the code in some of the functions it uses) for a bit. I wonder if this could be an upstream bug in tiny-skia? (Haven´t tested enough to tell yet)

PgBiel:
[...]

PR #5502, opened 2024-12-01
Fix text fill within `clip: true` blocks in PNG export

PgBiel:
Fixes https://github.com/typst/typst/issues/5499

Within `clip: true` blocks, only opaque solid colors worked properly (...to an extent, given the changed tests below) **in PNG export** (all's fine in PDF and SVG). Solid colors with transparency, gradients and patterns were all broken (see the sample at the bottom).

The problems were all located within `write_bitmap` at `typst-render/src/text.rs`. It has two code paths, one for `clip: true` (there is a mask) and one for `clip: false` (no mask). In the `clip: true` branch, the following problems were present:

1. Regarding transparency, we weren't using the `alpha_mul` function to apply the bitmap coverage value to the fill. We were simply replacing the fill's alpha with the coverage. Therefore, glyphs would never properly blend with what's under them, and would always be close to, if not fully, opaque, depending on the coverage value. The fix was to use `alpha_mul`.
2. Regarding gradients and patterns, the fill color sampler expects coordinates in the canvas coordinate space, not in the bitmap's. We were passing bitmap coordinates. The fix was to convert those bitmap coordinates to canvas coordinates.

## Tests

- Two tests were updated as the text color was slightly changed. I verified that the change is completely irrelevant (pixels went from RGB(127, 127, 127) to RGB(126, 126, 126) or something like that).
- Added a new test similar to the sample below but reduced.

## Sample

Given the code below:

```typ
#set page(width: 2in, height: 2.1in)

#let pat = pattern(
  size: (30pt, 30pt),
  relative: "parent",
  square(
    size: 30pt,
    fill: gradient
      .conic(..color.map.rainbow),
  )
)

#block(clip: false, {
  text(fill: blue, "Hello, World")
  [\ ]
  text(fill: blue.darken(20%).transparentize(50%), "Hello, World")
  [\ ]
  text(fill: gradient.linear(..color.map.rainbow), "Hello, World")
  [\ ]
  text(fill: pat, "Hello World")
})
#block(clip: true, {
  text(fill: blue, "Hello, World")
  [\ ]
  text(fill: blue.darken(20%).transparentize(50%), "Hello, World")
  [\ ]
  text(fill: gradient.linear(..color.map.rainbow), "Hello, World")
  [\ ]
  text(fill: pat, "Hello World")
})
```

We have the following **results:**

<details><summary>Broken output (before this PR)</summary>
<p>

fills are wrong in the block with clip: true

</p>
</details>

<details><summary>Correct output (with this PR)</summary>
<p>

fills are the same for both blocks

</p>
</details>


laurmaedje:
Makes sense!

PR #5505, opened 2024-12-02
Improve `symbol` `repr`

MDLC01:
This improves the `repr` for `symbol`, making it valid Typst code to produce the exact same (optionally modified) symbol.

Before this PR, the `repr` was simply that of a string containing the character that would be inserted if using the symbol as is.

| Symbol | Previous `repr` | New `repr` |
----------|---------|------|
| `sym.amp` | `"&"` | `symbol("&", ("inv", "⅋"))` |
| `sym.amp.inv` | `"⅋"` | `symbol("&", ("inv", "⅋")).inv` |
| `sym.smash` | `"⨳"` | `symbol("⨳")` |


Dherse:
This may be a *really* dumb question, but why not make it become `sym.amp`, `sym.amp.inv`, etc.?

Dherse:
Make sense, thanks!

tests/suite/symbols/symbol.typ
```diff
@@ -49,3 +49,8 @@
 --- symbol-unknown-modifier ---
 // Error: 13-20 unknown symbol modifier
 #emoji.face.garbage
+
+--- symbol-repr ---
+#repr(sym.amp) \
```
istudyatuni, line 54: Maybe it's better to use `test` to compare strings directly?

istudyatuni (commented):
As I understand, repr will be pretty big for symbols, which have multiple variants? Like `arrow.r.double.long.bar`. Also it will probably include other variants like `arrow.r.bar`?

PgBiel:
> As I understand, repr will be pretty big for symbols, which have multiple variants? Like `arrow.r.double.long.bar`. Also it will probably include other variants like `arrow.r.bar`?

I had the same concern reading the description. If we want it to be valid Typst code, maybe `symbol("the symbol")` is enough, or just `symbol("parent symbol", ("variant", "variant symbol")).variant`? 

To be honest, maybe it should be just like the old repr, but with `symbol( )` around it to avoid confusion with strings.

MDLC01:
The proposed `repr` can indeed become quite large for symbols with many variants. But this is similar to how element functions, `array`s, or `dict`s `repr`s can quickly grow in size when they have many fields. I think displaying all available variants can be beneficial because it provides a simple way to know the list of available variants, the same way most people expect `repr` for `dict` to display all available fields.

tests/suite/symbols/symbol.typ
MDLC01: Done

laurmaedje:
> I had the same concern reading the description.

Same here. It feels a bit different from a dict, since a dict really is just the data in it, while a symbol from a user PoV is just the single letter, and it happens to also support field access.  What's also a bit odd, is that after you've entered a few variants, it would still display _all_ the variants, even those that are unreachable.

At the same time, I can see the argument that it can be useful for debugging and understanding symbols and that's the main point of `repr`.

MDLC01:
Maybe in th case where some modifiers are applied, the unreachable variants could be removed from the `repr`. But appart from that, I mostly expect `repr` to show me everything there is to know about a value.

istudyatuni:
Maybe add a parameter to `repr` like `full: true` or similar?

laurmaedje:
> Maybe add a parameter to repr like full: true or similar?

I think that would be unnecessary complication.

laurmaedje:
> Maybe in th case where some modifiers are applied, the unreachable variants could be removed from the repr.

That could be a good middle ground.

PgBiel:
> Maybe in th case where some modifiers are applied, the unreachable variants could be removed from the repr.

I was thinking that we could do something like this then:

Given
```typ
#let envelope = symbol(
  "🖂",
  ("stamped", "🖃"),
  ("stamped.pen", "🖆"),
  ("lightning", "🖄"),
  ("fly", "🖅"),
)
```

We would have

| Symbol | Previous `repr` | New `repr` |
----------|---------|------|
| `envelope` | `"🖂"` | `symbol("🖂", ("stamped", "🖃"), ("stamped.pen", "🖆"), ("lightning", "🖄"), ("fly", "🖅"))` |
| `envelope.stamped` | `"🖃"` | `symbol("🖃", ("pen", "🖆"))` |
| `envelope.stamped.pen` | `"🖆"` | `symbol("🖆")` |
| `envelope.lightning` | `"🖄"` | `symbol("🖄")` |
| `envelope.fly` | `"🖅"` | `symbol("🖅")` |

We would perhaps lose some information, but I'm ok with that as it seems less confusing to me, and where the symbol came from doesn't really matter anyways.

Also, it's worth noting that this will enable listing the variants of a symbol. Currently, this doesn't seem to be possible, so we should probably have a follow-up PR with a method to return the symbol's variants as an array of arrays, exactly like you'd specify them to the symbol constructor. Otherwise, the only way to obtain this information will be through `repr` parsing.

laurmaedje:
Sounds good

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -244,9 +245,57 @@ impl Debug for List {
...
+                // that contain all applied modifiers.
+                applied_modifiers
+                    .clone()
+                    .all(|am| variant.split('.').any(|m| m == am))
+            })
+            .map(|(variant, c)| {
+                let trimmed_variant = variant
+                    .split('.')
+                    .filter(|&m| applied_modifiers.clone().all(|am| m != am))
+                    .collect::<Vec<_>>();
```
PgBiel, line 265: This logic seems like it would work well for the `envelope` example specifically, but for other examples it would probably break since it's not considering order, so something like `symbolname.variantA.variantB` and `symbolname.variantB.variantA` would be treated equally when trimming applied modifiers. The logic should have some zipping / common prefix behavior instead. We should also have a test for longer symbols where variants with the same name but different order are present.

MDLC01: I don't think order matters when selecting the variant.
https://github.com/typst/typst/blob/bb0c8140950b3eec020a4f0147bbc4ea65f3952a/crates/typst-library/src/foundations/symbol.rs#L322-L335

MDLC01:
```typ
#let s = symbol(
  ("foo.bar", "A"),
  ("bar.foo", "B"),
)

#s.foo
#s.bar
#s.foo.bar
#s.bar.foo
```
renders as "A A A A"

PgBiel: > ```typst
> #let s = symbol(
>   ("foo.bar", "A"),
>   ("bar.foo", "B"),
> )
> 
> #s.foo
> #s.bar
> #s.foo.bar
> #s.bar.foo
> ```
> 
> renders as "A A A A"

Uhhh. Is this not a bug?

laurmaedje: Modifiers are conceptually a set, not a list, so the symbol constructor call is bad here. It should probably error.

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -244,9 +244,49 @@ impl Debug for List {
     }
 }
 
+fn repr_variants<'a, 'b>(
```
laurmaedje, line 247: 'b seems to be unused.

could you also move this below the `Repr` impl?

laurmaedje (commented):
I'm positively surprised with how little code you've managed to implement this :)

crates/typst-library/src/foundations/symbol.rs
PgBiel: Hmmm... Interesting. If this means that "foo.bar" would always be able to be called as "bar.foo", then, to be honest, I find that quite confusing, as the order itself can mean something. But if not and this only matters internally, then I'd be fine with it (albeit a bit confused still).

laurmaedje: You can always call `s.foo.bar` as `s.bar.foo`. We take care during symbol naming to not introduce any order dependence. The idea is that you don't have to care about the order, you care about properties of the symbol. Like, "I want a symbol that is an arrow, that is dashed, and points to the top." So you write `arrow.` (see the autocomplete), add `.dashed` (see the autocomplete), add `.t`. And then you have `arrow.dashed.t` and you don't care that the definition is actually `arrow.t.dashed`. The autocomplete will always show you all currently available modifiers rather than the list of all symbols, so you can explore modifier by modifier, without any regard for the order. It's really a set more than a list.

laurmaedje:
Thanks!

PR #5506, opened 2024-12-02
More `CapturesVisitor` tests

wrzian:
(oops, wrong repo) Moving this PR here from https://github.com/wrzian/typst/pull/1.

First commit:
> test didn't capture the foo variable because it wasn't defined, but I believe the intention was to have a capturable call here.

Second commit:
> I actually kept working on this and added more tests :)
>
> lmk if the changes to `CapturesVisitor` aren't that helpful. I like that this makes it more explicit how undefined variables are treated, but I'm not sure its the best change

### DIFF
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -1,7 +1,8 @@
 use comemo::{Tracked, TrackedMut};
 use ecow::{eco_format, EcoString, EcoVec};
 use typst_library::diag::{
-    bail, error, At, HintedString, SourceDiagnostic, SourceResult, Trace, Tracepoint,
+    bail, error, At, HintedStrResult, HintedString, SourceDiagnostic, SourceResult,
+    Trace, Tracepoint,
 };
 use typst_library::engine::{Engine, Sink, Traced};
 use typst_library::foundations::{
@@ -135,8 +136,7 @@ impl Eval for ast::Closure<'_> {
 
         // Collect captured variables.
         let captured = {
-            // TODO: Maybe add warnings about undefined variables?
-            let mut visitor = CapturesVisitor::new(&vm.scopes, Capturer::Function, None);
+            let mut visitor = CapturesVisitor::new(Some(&vm.scopes), Capturer::Function);
             visitor.visit(self.to_untyped());
             visitor.finish()
         };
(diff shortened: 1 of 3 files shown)


laurmaedje:
Could you explain a bit more what the purpose of the changes is? The PR title doesn't really match the diff (test fix vs. changes to the capturing) and the description is somewhat hard to parse.

laurmaedje:
> lmk if the changes to CapturesVisitor aren't that helpful. I like that this makes it more explicit how undefined variables are treated, but I'm not sure its the best change

I'm not yet really sure what it does, but either way maybe it should be a separate PR from test changes?

wrzian:
Sorry, this one was a bit messy with me stopping and starting, realizing the PR was on the wrong branch, etc. I'll restart from here.

## Updating the `CapturesVisitor` tests

Within the main commit were kind of three changes all interleaved. I've split them into separate commits to better match this structure (<3 `jj`).

### Fixing the bad "foo" test

My initial motivation was the first commit: there was a test at the bottom of `test_captures` that didn't do what you would expect. The test used `foo`, but asserted that no values were captured (the empty array, `[]`). However, because `foo` wasn't actually defined first, the `CapturesVisitor` wouldn't capture it, and the test was incorrect.

I fixed this by changing `foo` to `f` and adding `f` to the array.

### Adding more tests

Then I realized that there were more issues with the tests, so I made the `test()` function require pre-defining variables at the top to make this behavior less likely in the future.

(pre-mature optimizations, me: :eyes:)

Then I added more tests with a new function for testing captures in math and some more tests over in `typst-ide::test_tooltip_closure`.

### Allowing the visitor to know about undefined variables

This was a change to the `CapturesVisitor` struct that in order to make explicit what was previously implict. Previously, if `external` was `Some`, all undefined variables that would be captured would just be ignored. If `external` was `None`, undefined variables would be added to the `internal` Scope (which starts empty). The caller in `typst-ide::closure_tooltip` used this behavior implicitly to determine the "captured" variables by iterating over `internal` when the visitor was `finish`ed.

But if the caller didn't pass in an empty scope, this behavior would be incorrect.

So I changed the behavior by adding an explicit pointer to an external vector so that the caller can track undefined variables explicitly. I updated `closure_tooltip` to do so.

This change adds functionality: we can now capture defined variables _and_ know which variables are undefined. We could use this new information to issue warnings during eval when a closure tries to capture undefined variables. However, I left that as TODO comments on the use sites since I think it's not really worth adding until we have a better warning system.

I also updated the `test` function to properly assert that there are no undefined variables (making the previous incorrect "foo" test impossible to write). This is why the change was in _this_ PR, but yeah, it's not really that thematically consistent.

---

I'm not sure if the last change is really worth it. It makes the visitor "more correct" but not by much and at the cost of some complexity. So I'd like to hear if you think the last part is worth it or if I should just revert the commit/split to a separate PR.

Regardless, the first two parts are definitely helpful.

wrzian:
Also sorry for throwing a bunch of paragraphs at you :sweat_smile:

laurmaedje:
> I'm not sure if the last change is really worth it. It makes the visitor "more correct" but not by much and at the cost of some complexity. So I'd like to hear if you think the last part is worth it or if I should just revert the commit/split to a separate PR.

Same here. The test changes look good, but I think I'd rather pass on the last part. We could reconsider adding it if we decide to catch more errors in closures early down the road.

laurmaedje:
There are also a few conflicts on the PR.

wrzian:
Removed the last commit and rebased. Should be good now.

laurmaedje:
Thanks!

PR #5517, opened 2024-12-04
Consider parameters when iterating items in scope

Myriad-Dreamin:
The iterator didn't consider parameters. This PR checks them when the node comes from the body of a closure.


laurmaedje:
Here, a test would be great as well. :)

The test discipline in typst-ide was bad so far, but I want to raise the standards a bit going forward.

laurmaedje:
Thanks!

PR #5518, opened 2024-12-04
Fix named items generated for path import

Myriad-Dreamin:
It is not tested, but I think we are only getting the value from some module by the first component of an import item, so it doesn't check the following code correctly:

```typ
#import "@blabla": a.b
                // ^^^ only considered the first component, i.e. the `a`
```


laurmaedje:
Could you add a test for this?

Myriad-Dreamin:
> Could you add a test for this?

Sure, but leave me some time to learn and write it :).

Myriad-Dreamin:
I added a test, but unluckily it also passes on main branch..

https://github.com/typst/typst/blob/54fcdcf7ed1a8ee21b035682b97ff11c483d4c54/crates/typst-ide/src/matchers.rs#L280

I don't know how do we test matchers with multiple files or extending IDE test framework is out of scope of this PR. How do you think of it?

laurmaedje:
The `TestWorld` does support multiple files. The autocomplete, definition, tooltip tests already use a new, more flexible testing style. The matchers & jump don't yet. If you want, you can take a stab at migrating them as well in a separate PR. Otherwise I'd do it, though not sure when exactly.

Myriad-Dreamin:
> If you want, you can take a stab at migrating them as well in a separate PR. Otherwise I'd do it, though not sure when exactly.

I see. We shall create a migration PR first and come back to this PR then.

laurmaedje:
@Myriad-Dreamin I'm taking care of the migration right now.

laurmaedje:
Done: https://github.com/typst/typst/pull/5602

Myriad-Dreamin:
I added a test that passes only if we accept the patch.

laurmaedje:
Thanks!

PR #5524, opened 2024-12-04
Live reloading HTTP server for `typst watch` and HTML export

laurmaedje:
Adds a built-in HTTP server that serves HTML output with live reload when using `typst watch`. The output file is still written as usual; the server is purely additional.

Adds three new CLI arguments to `typst watch`:
- `--no-serve` disables the server, only writing the file
- `--no-reload` keeps the server enabled, but disables live reload
- `--port` lets you override the port, by default it picks a free port in the range 3000-3005

The server injects a live reload script into both the placeholder and the real exported HTML. This script uses [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) to notify the browser whenever there is an update, which then refreshes the page. The file written to disk is not affected by this script, meaning that `typst watch` and `typst compile` continue to produce the same output (which I consider somewhat important). 

Further remarks:
- When the file isn't yet available, a placeholder page with the text "Waiting for output ..." is rendered. 
- The `--open` flag will default to the server, but still open the file if the server is disabled.
- The server is implemented with `tiny-http`, which is very lightweight (just 4 new dependencies).
- The server will need to be extended once we support multi-file output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,6 +124,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "ascii"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d92bec98840b8f03a5ff5413de5293bfcd8bf96467cf5452609f939ec6f5de16"
+
+[[package]]
 name = "autocfg"
 version = "1.4.0"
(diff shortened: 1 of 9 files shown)


laurmaedje:
It only happens in `typst watch` and only on `127.0.0.1`, not `0.0.0.0`, but I guess.

alerque:
I understand both those points, but being able to build a binary without an embedded web server that even *could* be triggered is an important prerequisite for use in some environments.

laurmaedje:
Fair enough. I don't have a problem with an enabled-by-default `http-server` feature in `typst-cli`.

laurmaedje:
Done.

PR #5525, opened 2024-12-04
Transform high level headings to HTML

jbirnick:
Currently, high level headings get silently transformed to `h6`. I suggest that Typst throws an error instead.

### DIFF
--- a/crates/typst-library/src/model/heading.rs
+++ b/crates/typst-library/src/model/heading.rs
@@ -8,7 +8,7 @@ use crate::foundations::{
     elem, Content, NativeElement, Packed, Resolve, Show, ShowSet, Smart, StyleChain,
     Styles, Synthesize, TargetElem,
 };
-use crate::html::{tag, HtmlAttr, HtmlElem};
+use crate::html::{attr, tag, HtmlElem};
 use crate::introspection::{
     Count, Counter, CounterUpdate, Locatable, Locator, LocatorLink,
 };
@@ -267,29 +267,35 @@ impl Show for Packed<HeadingElem> {
             realized = realized.styled(ParElem::set_hanging_indent(indent.into()));
         }
 
-        if html {
+        Ok(if html {
             // HTML's h1 is closer to a title element. There should only be one.
             // Meanwhile, a level 1 Typst heading is a section heading. For this
             // reason, levels are offset by one: A Typst level 1 heading becomes
             // a `<h2>`.
             let level = self.resolve_level(styles).get();
             if level >= 6 {
-                engine.sink.warn(warning!(span, "heading of level {} should be transformed to <h{}> in HTML, but only <h1> to <h6> are supported; instead transformed to <div role=\"heading\" aria-level=\"{}\">, but note that this is not supported by all screen readers", level, level + 1, level + 1;
-                    hint: "you may want to restructure your document so that it doesn't contain deep headings"));
-                Ok(HtmlElem::new(tag::div)
+                engine.sink.warn(warning!(span,
+                    "heading of level {} was transformed to \
+                    <div role=\"heading\" aria-level=\"{}\">, which is not \
+                    supported by all screen readers",
+                    level, level + 1;
+                    hint: "HTML only supports <h1> to <h6>, not <h{}>", level + 1;
+                    hint: "you may want to restructure your document so that \
+                          it doesn't contain deep headings"));
+                HtmlElem::new(tag::div)
                     .with_body(Some(realized))
-                    .with_attr(const { HtmlAttr::constant("role") }, "heading")
-                    .with_attr(const { HtmlAttr::constant("aria-level") }, (level + 1).to_string())
+                    .with_attr(attr::role, "heading")
+                    .with_attr(attr::aria_level, (level + 1).to_string())
                     .pack()
-                    .spanned(span))
+                    .spanned(span)
             } else {
                 let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level - 1];
-                Ok(HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span))
+                HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span)
             }
         } else {
             let realized = BlockBody::Content(realized);
-            Ok(BlockElem::new().with_body(Some(realized)).pack().spanned(span))
-        }
+            BlockElem::new().with_body(Some(realized)).pack().spanned(span)
+        })
     }
 }
 
--- a/crates/typst-library/src/model/heading.rs
+++ b/crates/typst-library/src/model/heading.rs
@@ -2,7 +2,7 @@ use std::num::NonZeroUsize;
 
 use typst_utils::NonZeroExt;
 
-use crate::diag::SourceResult;
+use crate::diag::{bail, SourceResult};
 use crate::engine::Engine;
 use crate::foundations::{
     elem, Content, NativeElement, Packed, Resolve, Show, ShowSet, Smart, StyleChain,
@@ -267,18 +267,22 @@ impl Show for Packed<HeadingElem> {
             realized = realized.styled(ParElem::set_hanging_indent(indent.into()));
         }
 
-        Ok(if html {
+        if html {
             // HTML's h1 is closer to a title element. There should only be one.
             // Meanwhile, a level 1 Typst heading is a section heading. For this
             // reason, levels are offset by one: A Typst level 1 heading becomes
             // a `<h2>`.
-            let level = self.resolve_level(styles);
-            let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level.get().min(5) - 1];
-            HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span)
+            let level = self.resolve_level(styles).get();
+            if level >= 6 {
+                bail!(span, "heading of level {} cannot be transformed to HTML", level;
+                    hint: "only headings of level 5 and smaller can be transformed to HTML");
+            }
+            let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level - 1];
+            Ok(HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span))
         } else {
             let realized = BlockBody::Content(realized);
-            BlockElem::new().with_body(Some(realized)).pack().spanned(span)
-        })
+            Ok(BlockElem::new().with_body(Some(realized)).pack().spanned(span))
+        }
     }
 }
 
--- a/crates/typst-library/src/model/heading.rs
+++ b/crates/typst-library/src/model/heading.rs
@@ -2,13 +2,13 @@ use std::num::NonZeroUsize;
 
 use typst_utils::NonZeroExt;
 
-use crate::diag::{bail, SourceResult};
+use crate::diag::{warning, SourceResult};
 use crate::engine::Engine;
 use crate::foundations::{
     elem, Content, NativeElement, Packed, Resolve, Show, ShowSet, Smart, StyleChain,
     Styles, Synthesize, TargetElem,
 };
-use crate::html::{tag, HtmlElem};
+use crate::html::{tag, HtmlAttr, HtmlElem};
 use crate::introspection::{
     Count, Counter, CounterUpdate, Locatable, Locator, LocatorLink,
 };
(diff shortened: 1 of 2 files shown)


laurmaedje:
Perhaps a warning

jbirnick:
Okay let's say a user has very deep headings. Then this has a reason, right, meaning they actually need such a deeply nested document structure. Now suppose they export to HTML. Then they realize (e.g. by a warning) that HTML only has `h6` and not more. What do they want to do? 

1. "oh fine then Typst please just make deeper headings an `h6` that's okay for my usecase"
2. "oh shoot, I guess I have to restructure my document so that it's properly reflected in HTML"

If 1. is the common answer than we should do a warning. However, I don't know about you guys, but I thought no one wants 1.? Like why would you have deeply nested headings but be like "fine just destroy my hierarchy it's not important". That's why I chose error, to make it very clear to the user that HTML simply doesn't support deep heading hierarchy and they have to restructure their document.

However, if you all think a warning is better (and users will actually read it) then of course I can change it to a warning.

reknih:
There actually is an accessible way to model higher-level headings with HTML and ARIA attributes: Add `role="heading" aria-level="7"` to the heading element's attributes. This can be styled with this CSS selector `*[role="heading"][aria-level="7"]`. The only question is which tags to use for this. By default, I think that the heading should be a block level element. With that requirement, we have the following options:

- `<h6 aria-level="7">`: Using a H6, it's more obvious in markup that we have a heading, but the level is confusing. We can omit `role="heading"` and inherit the user-agent's styles for Level 6 headings.
- `<div role="heading" aria-level="7">`: Using a single `div` element with role and level. This is what the [WCAG 2.1 Techniques example](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA12) for deeply nested headings proposes. Simple, but comes with no user-agent styles. Perhaps this is okay since it's such a low-level heading anyways.
- `<div role="heading" aria-level="7"><b>` or `<p role="heading" aria-level="7"><b>`: Nest a `div` around a `b` tag. This uses native tags with no particular semantics to produce boldface, and, in the case of `p`, spacing. This has more moving parts, but is high fidelity. Such a solution should _not_ use a strong because it carries its own semantics of making parts of the (or in this case, the whole) heading more important. In absence of the `role` attribute, `<p><strong>` would be a good way to model a lower-level heading, but fortunately, we do not need to fall back to this.

Final remark: WCAG Technique says that screen readers struggle with heading levels deeper than 10. But I think that's up to accessibility software vendors.

jbirnick:
Thanks for the input! As you mentioned, [this example on w3.org](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA12#example-2-additional-heading-levels) uses `<div role="heading" aria-level="...">, so that's what I implemented now.

But the internet says that not all screen readers support this, and it's also some sort of "unexpected behavior", which is why I added a warning.

Implementation notes/questions:
- what is the idiomatic way to convert a `usize` to an `EcoString`? I went `usize` -> `String` -> `EcoString`, but this causes an extra allocation; I guess that's relevant for all attributed which have numbers as values
- to construct the `HtmlAttr` I was not sure whether to use `HtmlAttr::constant` or `HtmlAttr::intern`; it says that `constant` should be only used in const contexts because it can panic, on the other hand the attribute name is fixed no I can verify by hand that `constant` doesn't panic
-  that said, it actually currently does panic for the attribute `aria-level`, because it contains a dash, and right now only alphanumeric characters are allowed for attributes

**Blockers:**
- Need to allow dashes (or maybe even more characters) for attributes. Unfortunately I couldn't find the part of the HTML spec which discusses the possible characters.

laurmaedje:
> But the internet says that not all screen readers support this, and it's also some sort of "unexpected behavior", which is why I added a warning.

Do you have a source I can read up on? The linked W3C resource speaks only of levels >=10, so we could also only start warning there.

> what is the idiomatic way to convert a usize to an EcoString? I went usize -> String -> EcoString, but this causes an extra allocation; I guess that's relevant for all attributes which have numbers as values

```rust
use ecow::eco_format;
eco_format!("{number}")
```

> to construct the HtmlAttr I was not sure whether to use HtmlAttr::constant or HtmlAttr::intern, and what is better for performance; i used constant for now

constant is better for perf because it will intern/encode at compile time

> currently constant panics for the attribute aria-level, because it contains a dash, and right now only alphanumeric characters are allowed for attributes

> Need to allow dashes (or maybe even more characters) for attributes. Unfortunately I couldn't find the part of the HTML spec which discusses the possible characters.

The spec work is already done, in `typst-library/src/html/dom.rs > mod charsets`. I just didn't use it in `constant` because it takes a char and `chars()` is not const. So I took the shortcut of just calling `is_ascii_alphanumeric` in `HtmlTag` (which is correct) and copy-pasted it for `HtmlAttr` (which is incorrect). You could adjust the relevant code in `HtmlAttr::constant` and replace `bytes[i].is_ascii_alphanumeric()` check `byte.is_ascii() + charsets::is_valid_in_attribute_name(byte as char)`. Maybe it would be cleaner to do the same in `HtmlTag` as well, although it really is the same there.

crates/typst-library/src/model/heading.rs
```diff
@@ -267,18 +267,29 @@ impl Show for Packed<HeadingElem> {
...
             // HTML's h1 is closer to a title element. There should only be one.
             // Meanwhile, a level 1 Typst heading is a section heading. For this
             // reason, levels are offset by one: A Typst level 1 heading becomes
             // a `<h2>`.
-            let level = self.resolve_level(styles);
-            let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level.get().min(5) - 1];
-            HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span)
+            let level = self.resolve_level(styles).get();
+            if level >= 6 {
+                engine.sink.warn(warning!(span, "heading of level {} should be transformed to <h{}> in HTML, but only <h1> to <h6> are supported; instead transformed to <div role=\"heading\" aria-level=\"{}\">, but note that this is not supported by all screen readers", level, level + 1, level + 1;
```
laurmaedje, line 277: The error message should be line wrapped in source via (preventing extra whitespace via `\`, you can check other calls to `warning!` or `bail!` for that. It's also generally quite long. I wonder whether we could split it up more into a shorter main message with multiple hints.

crates/typst-library/src/model/heading.rs
```diff
@@ -267,18 +267,29 @@ impl Show for Packed<HeadingElem> {
             realized = realized.styled(ParElem::set_hanging_indent(indent.into()));
         }
 
-        Ok(if html {
+        if html {
```
laurmaedje, line 270: The `Ok` could stay here I think

crates/typst-library/src/model/heading.rs
```diff
@@ -267,18 +267,29 @@ impl Show for Packed<HeadingElem> {
...
-            let level = self.resolve_level(styles);
-            let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level.get().min(5) - 1];
-            HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span)
+            let level = self.resolve_level(styles).get();
+            if level >= 6 {
+                engine.sink.warn(warning!(span, "heading of level {} should be transformed to <h{}> in HTML, but only <h1> to <h6> are supported; instead transformed to <div role=\"heading\" aria-level=\"{}\">, but note that this is not supported by all screen readers", level, level + 1, level + 1;
+                    hint: "you may want to restructure your document so that it doesn't contain deep headings"));
+                Ok(HtmlElem::new(tag::div)
+                    .with_body(Some(realized))
+                    .with_attr(const { HtmlAttr::constant("role") }, "heading")
```
laurmaedje, line 281: this works, but I think it's nicer to add these to the macro in `mod attr` in `dom.rs` (which creates a constant just like this). then we can have them all in one place, meaning less chance for typos (this won't be the only `role` attr we need). 

as for aria level: the macro doesn't support case transform, but maybe we can just spell out the constants with a hyphen manually next to the macro, at least for now. or we can leave it here, since it's unlikely to be reused anyway.

jbirnick:
All fixed.

Concerning screen reader support, the most common screen readers seem to be:
- JAWS
  according to [this post](https://stackoverflow.com/a/56067431), it didn't work 5 years ago (but i can't test it now since it is for Windows
- NVDA
  can't test since for Windows
- ChromeVox
  for `<h1>` to `<h6` it says "heading 1/2/.../6", for our implementation it just says "heading" without naming the `aria-level`
- Narrator
  can't test since for Windows
- VoiceOver
  can't test since for Mac

laurmaedje:
Just tested VoiceOver and it happily anounces a level 1024 heading for `<div role="heading" aria-level="1024">...</div>`, so no problems there.

reknih:
Just leaving this here for future reference: https://webaim.org/projects/screenreadersurvey10/#used (Pervasiveness of different screen reader software)

reknih:
Just tried NVDA on the Windows 10 box in the office and it works fine with Firefox, but says heading level 2 with Blink-based browsers (Chrome, Edge). I have left the audio wire at home and the computer has no speaker, so I could not yet test Microsoft Narrator but will do.

crates/typst-library/src/model/heading.rs
```diff
@@ -272,9 +272,26 @@ impl Show for Packed<HeadingElem> {
...
+                    <div role=\"heading\" aria-level=\"{}\">, which is not \
+                    supported by all screen readers",
+                    level, level + 1;
+                    hint: "HTML only supports <h1> to <h6>, not <h{}>", level + 1;
+                    hint: "you may want to restructure your document so that \
+                          it doesn't contain deep headings"));
+                HtmlElem::new(tag::div)
+                    .with_body(Some(realized))
+                    .with_attr(attr::role, "heading")
+                    .with_attr(attr::aria_level, (level + 1).to_string())
```
laurmaedje, line 288:
```suggestion
                    .with_attr(attr::aria_level, eco_format!("{}", level + 1))
```

crates/typst-library/src/model/heading.rs
```diff
@@ -272,9 +272,26 @@ impl Show for Packed<HeadingElem> {
...
             // a `<h2>`.
-            let level = self.resolve_level(styles);
-            let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level.get().min(5) - 1];
-            HtmlElem::new(t).with_body(Some(realized)).pack().spanned(span)
+            let level = self.resolve_level(styles).get();
+            if level >= 6 {
+                engine.sink.warn(warning!(span,
+                    "heading of level {} was transformed to \
+                    <div role=\"heading\" aria-level=\"{}\">, which is not \
+                    supported by all screen readers",
```
laurmaedje, line 280:
```suggestion
                    supported by all assistive technology",
```

reknih:
Also tested Narrator now. First, let me say that the UX is a steaming pile of garbage.

- For Firefox, it does not announce that the text is a heading at all. It just gets announced as regular text.
- For Blink-based browsers, it fails the same way NVDA does. (announces heading level 2)

Still have not tested JAWS yet, but we do not have access to it right now. Given that (of the tested technology) it only works correctly with VoiceOver and NVDA in Firefox, I'd say a warning is appropriate.

jbirnick:
Alright fixed everything.

laurmaedje:
Thanks!

crates/typst-library/src/model/heading.rs
```diff
@@ -272,9 +273,26 @@ impl Show for Packed<HeadingElem> {
...
+                    hint: "you may want to restructure your document so that \
+                          it doesn't contain deep headings"));
+                HtmlElem::new(tag::div)
+                    .with_body(Some(realized))
+                    .with_attr(attr::role, "heading")
+                    .with_attr(attr::aria_level, eco_format!("{}", level + 1))
+                    .pack()
+                    .spanned(span)
+            } else {
+                let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level - 1];
```
istudyatuni, line 293: I think there should be 

```suggestion
                let t = [tag::h2, tag::h3, tag::h4, tag::h5, tag::h6][level + 1];
```

jbirnick: No it shouldn't. E.g. `= Heading` gives you a heading of level 1, and this should be converted to `h2`, which is at array index 0. So it's all correct.

istudyatuni: Oh, array indices, my bad

PR #5530, opened 2024-12-05
Fix warning for HTML export to reflect renaming of --feature to --features

Drodt:
Fixes the CLI output when using the HTML export.


laurmaedje:
Ah, right. Thanks!

PR #5531, opened 2024-12-05
Fix autocompletion of half-completed import item

Eric-Song-Nop:
When completing behind a half-started identifier in an import list:
`#import "path.typ": thi|",`, the parent should be and ImportItemPath, which parent is ImportItems.

This bug make completion behind a half-started identifier in an import list NEVER possible.


laurmaedje:
Could you add tests for the half-started & unstarted cases?

laurmaedje:
There was a small bug in your fix: You used `ctx.leaf.parent()` in both cases, effectively requiring `SyntaxKind::ImportItemPath == parent.kind() == SyntaxKind::ImportItems`, which can never be true. I fixed it and added tests.

laurmaedje:
Thanks!

PR #5533, opened 2024-12-05
Basic HTML pretty-printing

01mf02:
With this PR, the following Typst document:

~~~ typst
= Main Section

Hello T#emph[yps]t from *HTML*!

== Subsection

#html.elem("div")[#html.elem("button")[one]#html.elem("button")[two]]
~~~

turns into:

~~~
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <h2>
      Main Section
    </h2>
    <p>
      Hello T<em>yps</em>t from <strong>HTML</strong>!
    </p>
    <h3>
      Subsection
    </h3>
    <div>
      <button>one</button><button>two</button>
    </div>
  </body>
</html>
~~~

Before, it was:

~~~
<!DOCTYPE html><html><head><meta charset="utf-8"><meta name="viewport" content="width=device-width, initial-scale=1"></head><body><h2>1. Main Section</h2><p>Hello T<em>yps</em>t from <strong>HTML</strong>!</p><h3>1.a. Subsection</h3><div><button>one</button><button>two</button></div></body></html>
~~~

This should make writing tests much easier.

Pretty-printing can still be disabled in `encode.rs` by setting `pretty: false`.

Note that the goal of this PR is not to write a full-fledged HTML prettifier; in particular, content is not line-wrapped. Such more complex behaviour would make the implementation harder to understand and make its behaviour less predictable for users. This PR is written with the 80/20 rule in mind: Make it 80% pretty with 20% of the work. That's also why relatively few elements currently trigger an indentation. Still, this list of elements can be gradually expanded.

Thanks to @reknih for pointing out that CSS can actually make whitespace matter where it does not by default, which breaks pretty printing assumptions that we can insert whitespace while preserving semantics. An example that makes `<p>` whitespace-sensitive, for example: <https://www.w3.org/TR/css-text-3/#example-af2745cd>
Still, for most users, I suspect that having some pretty printing is still beneficial (at least for us, it will be when we write tests!).
So I think that to cover all (potentially pathological) use cases, it would be enough to make pretty printing configurable (on/off), with defaulting to on. Or we even make the list of tags that should be indented configurable. We could do this in a follow-up PR.


laurmaedje:
I think this approach can work. Using the default value of `display: block` is also what [prettier does](https://prettier.io/docs/en/options.html#html-whitespace-sensitivity).

A few remarks:

- Perhaps we should prevent any kind of wrapping in non-wrapping elements in a nested way (e.g. for `#span[#div[]]`, even though this is not technically valid), because the output looks bad.

- A new function `tag::is_block_by_default` would be good to extract and use in `whitespace_inside`. I think we can fill in the whole list from the get go rather than just adding a few elements. Probably `tag:is_inline` should be renamed to `tag::is_inline_by_default`.

- `<meta>` should not allow `whitespace_inside` since it's void.

01mf02:
> Perhaps we should prevent any kind of wrapping in non-wrapping elements in a nested way (e.g. for `#span[#div[]]`, even though this is not technically valid), because the output looks bad.

I would not try to correct technically invalid input. Garbage in, garbage out.

> A new function `tag::is_block_by_default` would be good to extract and use in `whitespace_inside`. I think we can fill in the whole list from the get go rather than just adding a few elements. Probably `tag:is_inline` should be renamed to `tag::is_inline_by_default`.

I agree about `block_by_default`.

I would be all for having such a list, but your saying "the whole list" implies that there *is* such a list, whereas the HTML standard does not mention one, and [other sources](https://stackoverflow.com/a/76592597) deny that there is such a list. The problem is intensified by the fact that many tags can appear in multiple contexts. Which elements should be part of the list in your opinion?

> `<meta>` should not allow `whitespace_inside` since it's void.

I agree.

laurmaedje:
> I would not try to correct technically invalid input. Garbage in, garbage out.

It's not about correcting anything, just about not making the printing unnecessarily ugly just because the element nesting is semantically invalid. It should be trivial to disable pretty printing within the scope of a non-block element. I see no reason for not doing it.

01mf02:
> You can obtain the list that `prettier` uses in these two files:
> 
> * https://github.com/prettier/prettier/blob/401b010c16eb90daaefe0350ab8bdc41098ba8dc/src/language-html/constants.evaluate.js
> * https://github.com/prettier/html-ua-styles/blob/60f64690355c0856f93d89db77f79522fdc74329/index.js

Thanks a lot for the links! I added the elements in the second file that are marked with:

~~~
    styles: [
      {
        property: 'display',
        value: 'block',
      },
    ],
~~~

01mf02:
> It's not about correcting anything, just about not making the printing unnecessarily ugly just because the element nesting is semantically invalid. It should be trivial to disable pretty printing within the scope of a non-block element. I see no reason for not doing it.

A reason for not doing it could be code size. But I do not have a strong opinion about this, so I did implement your suggestion in [c067909](https://github.com/typst/typst/pull/5533/commits/c0679097f46c99419c3a9184369851d4c50d6552).

01mf02:
The input:

~~~ typst
= Main Section

Hello T#emph[yps]t from *HTML*!

#html.elem("button")[#html.elem("div", [one])]

== Subsection

#html.elem("div")[#html.elem("div")[#html.elem("button")[one]#html.elem("button")[two]]]
~~~

now produces:

~~~ html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"><meta name="viewport" content="width=device-width, initial-scale=1">
  </head>
  <body>
    <h2>
      Main Section
    </h2>
    <p>
      Hello T<em>yps</em>t from <strong>HTML</strong>!
    </p>
    <button><div>one</div></button>
    <h3>
      Subsection
    </h3>
    <div>
      <div>
        <button>one</button><button>two</button>
      </div>
    </div>
  </body>
</html>
~~~

laurmaedje:
Looks good mostly, though I think `<meta>` should probably have whitespace around it.

I see that you deleted the whitespace_around / whitespace_inside distinction though. I was unsure about it as well, but for `<meta>` it might make sense.

crates/typst-library/src/html/dom.rs
```diff
@@ -470,6 +483,51 @@ pub mod tag {
         wbr
     }
 
+    /// Can the content of a node with the tag be surrounded by whitespace without impacting the document?
+    ///
+    /// This is an underapproximation.
+    pub fn is_block_by_default(tag: HtmlTag) -> bool {
```
laurmaedje, line 489: Please add some explanation that this is based on `display: block` and that it can be overridden by CSS and also add a link to the relevant source(s).

01mf02: I did this in [e156378](https://github.com/typst/typst/pull/5533/commits/e15637813e3da0f9331aac9d1b113636c15b0215).

laurmaedje:
I think I'd prefer to add `<meta>` to `HtmlElement::is_pretty` rather than `is_block_by_default`. The comments look good!

crates/typst-library/src/html/dom.rs
```diff
@@ -470,6 +483,63 @@ pub mod tag {
         wbr
     }
 
+    /// Can the content of a node with the tag be surrounded by whitespace without impacting the document?
```
laurmaedje, line 486: Please wrap the comments at 80 columns.

01mf02: Done.

PR #5534, opened 2024-12-05
Bump Rust to 1.83

laurmaedje:


tingerrr:
This didn't bump the rust-version inside the top-level manifest. I think this could cause breakage for consumers using `>= 1.80, < 1.83` if you use a `> 1.80` feature.

laurmaedje:
It's not an MSRV bump, just a Rust version bump for new clippy lints. We have CI to check that we don't use >1.80 features.

PR #5541, opened 2024-12-06
Improve raw trimming

### LINKED ISSUE #4790: Handling empty single line raw block
### Description

````typ
#(
  ```test ```
)

````

The ast for code above is

```
Markup: 19 [
    Hash: "#",
    Parenthesized: 17 [
        LeftParen: "(",
        Space: "\n  ",
        Raw: 11 [
            RawDelim: "```",
            RawLang: "test",
            RawDelim: " ```",
        ],
        Space: "\n",
        RightParen: ")",
    ],
    Space: "\n",
]
```

In which the ````RawDelim: " ```",```` part looks strange. There is a space in raw delim.

### Reproduction URL

_No response_

### Operating system

### Typst version

### PR DESCRIPTION
**Pull request by wrzian**
Closes #4790

Testing the two commits with the `--parser-compare` flag, I get the following diff in the `raw-blocky` test that shows the fix:

diff of the raw-blocky test ast

During the parser refactor, I think I rewrote the `blocky_raw` function like 3 separate times, but was never really satisfied. This time I've finally been able to join all the separate complexities into individual blocks and it feels much more cohesive :)

The main algorithmic changes are that we handle the leading space and the rest of the first line in one spot outside of the main line loop, and the logic for the final line is also condensed into one spot. Note the added `needs_trim` variable, which is the main part which fixes the bug in #4790.

### CODE REVIEW

File: crates/typst-syntax/src/lexer.rs [hunk 1]
```diff
@@ -313,6 +314,35 @@ impl Lexer<'_> {
...
+    ///   each line that has any non-whitespace characters.
+    ///     - The opening delimiter's line does not contribute to the dedent,
+    ///       but the closing delimiter's line does (even if that line is
+    ///       entirely whitespace up to the delimiter).
+    /// - We then trim the newline and dedent characters of each line, and add a
+    ///   (potentially empty) text element of all remaining characters.
+    ///
+    /// ### The final line:
+    /// - If the last line is entirely whitespace, it is trimmed.
+    /// - Otherwise it's text is kept like an inner line. However, if the last
```
laurmaedje, line 343:
```suggestion
    /// - Otherwise its text is kept like an inner line. However, if the last
```

crates/typst-syntax/src/lexer.rs
```diff
@@ -313,6 +314,35 @@ impl Lexer<'_> {
...
+    ///   trim it. Otherwise we trim a single leading space if present.
+    ///     - If more trimmed characters follow on future lines, they will be
+    ///       merged into the same trimmed element.
+    /// - If we didn't trim the entire line, the rest is kept as text.
+    ///
+    /// ### Inner lines:
+    /// - We determine the "dedent" by iterating over the lines. The dedent is
+    ///   the minimum number of leading whitespace characters (not bytes) before
+    ///   each line that has any non-whitespace characters.
+    ///     - The opening delimiter's line does not contribute to the dedent,
```
laurmaedje, line 335: Please use 2 spaces of indent in Markdown (and then rewrap at column 80). :)

laurmaedje (commented):
If I remove the `needs_trim` variable, still all tests pass. There should probably be a test for what it does!

wrzian:
Good catch! Actually the `needs_trim` variable _isn't_ needed, but the reasoning is really subtle. I removed the variable, but added a debug_assert and a comment explaining why.

laurmaedje:
Thanks!

Issue #4573, opened 2024-07-17
Index out of bounds: the len is 26 but the index is 30


qarmin:
### Description

File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
#let apa-bibliography = (() => {
let dict-from-entries = entries => {
	for (key, value) in entries {
let n(x, default, f) = if x == none { default } else  f(x) }
	}
}
let months = (
)
let parse-date(x) = {
(works, include-all: true) => {
	let use-workkey) = [#used-works.update(x => x + (key,))]
```

command
```
timeout -v 100 typst compile TEST___FILE.typ
```

cause this
```
thread 'main' panicked at crates/typst-syntax/src/parser.rs:838:7:
index out of bounds: the len is 26 but the index is 30
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

##### Automatic Fuzzer note, output status "Some(101)", output signal "None"
```

[compressed.zip](https://github.com/user-attachments/files/16265972/compressed.zip)

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

bluebear94:
Seems to be caused by #3463. If I disable the memoization in `expr_with_paren`, then the panic doesn’t occur.

bluebear94:
I think this happens because the parser needs to parse a parenthesized expression at the beginning of the line containing `(works, include-all: true) => {` in two different situations, with two different lists of accumulated syntax nodes in `p.nodes`. Not sure how to fix this; we could memoize by `(p.current_start, p.nodes.len())` instead of by `p.current_start` only, but that might hide a deeper issue.

wrzian:
Good news! This was resolved by the parser refactor (#5310) and should ship with the next version update. In particular it was commit 5 (16cc7eb), which refactored the memoization system, and mainly this comment in `restore_memo_or_checkpoint`:

```rust
// It's important that we don't truncate the nodes vector since
// it may have grown or shrunk (due to other memoization or
// error reporting) since we made this checkpoint.
```

In this case, the number of errors increases between memoization steps, and the old version incorrectly truncated the nodes vector, leading to an out-of-bounds access plus panic when we report the final unclosed curly-brace.

The most minimal reproduction I could make is:
```typ
#{
  (
    if x {} else {}
    { () = 2
  ) = 3
}
```

Let's go over each of the parts:

- Setup
[...]

PR #5542, opened 2024-12-07
Add test for issue #4573

wrzian:
#4573 is already fixed, but this PR adds the minimal example I commented as a test case. Running it on a commit before commit 5 of the parser refactor (16cc7eb) should cause a crash in the parser.

### DIFF
--- a/tests/suite/scripting/destructuring.typ
+++ b/tests/suite/scripting/destructuring.typ
@@ -373,3 +373,19 @@
 // Error: 6-12 too many elements to destructure
 // Hint: 6-12 the provided array has a length of 3, but the pattern expects 2 elements
 #for (x, y) in ((1,2,3), (4,5,6)) {}
+
+--- issue-4573-destructuring-unclosed-delimiter ---
+// See comment in issue for explanation. This one's complex.
+#{
+  (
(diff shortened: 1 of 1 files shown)


tests/suite/scripting/destructuring.typ
```diff
@@ -373,3 +373,19 @@
 // Error: 6-12 too many elements to destructure
 // Hint: 6-12 the provided array has a length of 3, but the pattern expects 2 elements
 #for (x, y) in ((1,2,3), (4,5,6)) {}
+
+--- issue-4573-destructuring-unclosed-delimiter ---
+// See comment in issue for explanation. This one's complex.
```
PgBiel, line 378: I think a short summary or fragment of your comment / the problem could be good. Something like "Ensure the parsing of the external assignment doesn't incorrectly lead the parser to an invalid state and to a subsequent panic when generating the unclosed delimiter error. See the comment in the issue for more explanation."

PgBiel (approved):
OK by me. Although I have a feeling that the errors being tested may change and cause the test to fail with regular changes to the parser, without it being a bug. But, if that happens, I guess we can just update the test accordingly.

laurmaedje:
Thanks!

Issue #5519, opened 2024-12-04
Non-decimal `Numeric` literal gets set to zero


T0mstone:
### Description

When specifying e.g. a length in a non-decimal base (e.g. binary), it gets set to zero instead.

Example:
```typst
#set page(width: auto, height: auto, margin: 1pt)
First#h(0b100000pt)Second\
First#h(32pt)Second
```
![output](https://github.com/user-attachments/assets/c5f0b8da-7d0a-45b2-b9dc-c0f2c4a36b6c)

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

wznmickey:
In current code (ast.rs impl Numeric<'_>), the number is directly parsed by rust, which is different from what we did for Int value.

In document, it seems that we did not write in detail what format the number part should follow.

https://typst.app/docs/reference/layout/fraction/

laurmaedje:
After a bit of consideration, I think I'd rather disallow this than handle it. It feels odd to allow it for whole numerics, but not float numerics, but it's also odd to support it for floats.

wznmickey:
Then maybe we need to report an error rather than just use 0.

laurmaedje:
Yeah

wznmickey:
During changing the code, we found that we also support exponent, which I think is not write in the document, should we support something like `1e2p`t?

laurmaedje:
I think that's less odd and more useful and we can keep it

wznmickey:
Changed in https://github.com/typst/typst/pull/5548.
By the way is it necessary to modify the document to make it more clear?

laurmaedje:
I think the docs are fine tbh. This level of detail is more something for a specification and adding too much of that kind of information would drown out more relevant things. What the docs are truly lacking is introductory and guide-level content.

PR #5548, opened 2024-12-08
Forbid base prefix for numbers with a unit

wznmickey:
Fixes https://github.com/typst/typst/issues/5519 ~~copy the parse code for `Int` into `Numeric`~~ by forbidding base prefixes for numers with a unit.


PgBiel:
By the way, I recommend using separate branches for your PRs and not using `main` in your fork (only updating it to Typst's latest `main` before branching off for a PR). That not only allows you to open more than one PR at a time, but also ensures changes to this PR which may be requested by maintainers won't affect your fork's other branches in any way. (Also makes it easier to wipe past history as you start clean from upstream main each time.)

wznmickey:
> By the way (just a general tip for future PRs - this is **not** a blocker for your PR), I recommend using separate branches for your PRs and not using `main` in your fork (only updating it to Typst's latest `main` before branching off for a PR). That not only allows you to open more than one PR at a time, but also ensures changes to this PR which may be requested by maintainers won't affect your fork's other branches in any way. (Also makes it easier to wipe past history as you start clean from upstream main each time.)

Thank you.
I just use main for ease. I will use other branch later.

laurmaedje:
See https://github.com/typst/typst/issues/5519#issuecomment-2526210351

PR #5560, opened 2024-12-11
Add number-syntax edge case tests

wrzian:
I was following the number prefix changes (#5519) and found that we didn't have very many tests for basic number-syntax edge cases, so I thought of a few that I didn't see elsewhere.

I added these tests just as a single named test with implicit allowed values and explicit errors, which is fine because all syntax errors are reported at once. But I can change that if you'd prefer them split out into separate tests.

I also am not sure that we really want to allow the double-percent sign. I think that it may be more consistent to just lex all trailing alpha-numeric characters or percent signs following numbers just for simplicity. It would improve the second error, but disallow the first case. But I don't really think the first case is actually helpful for markup mode.


laurmaedje:
Thanks! 

> I also am not sure that we really want to allow the double-percent sign. I think that it may be more consistent to just lex all trailing alpha-numeric characters or percent signs following numbers just for simplicity. It would improve the second error, but disallow the first case. But I don't really think the first case is actually helpful for markup mode.

I don't have a strong opinion either way. I don't really see a strong reason to forbid it (nobody will ever write that), but I'm also not necessarily opposed.

wrzian:
I feel like it would be good for consistency of implementations, i.e. other parsers/highlighters, to say that number suffixes include any series of alphanumeric/percent characters (so turning `#1.2%%` to an error). Feels like it better follows the principal of "maxmimum munch".

Would you be ok with me implementing that?

laurmaedje:
I'd be okay with that.

PR #5564, opened 2024-12-12
Add timezone to PDF's default timestamp

### LINKED ISSUE #3948: Saving Date Metadata with Time-Zone instead of Auto-Converting to UTC

wzf03:
Fixes #3948, let the date metadata of the pdf include timezone information if the date is not specified (or auto) in `#set document`.

Thanks the previous attempt #4636 on this issues, and this pr is inspired from it.

## The timezone
Similar to #4636, and in summary, it will be:

1. Respect [document.date](https://typst.app/docs/reference/model/document/#parameters-date), either a `datetime(...)` or `none`.
2. If `document.date` is auto (the default value),
    1. if the `--creation-timestamp` is set in CLI, the timezone will be set to UTC;
    2. else, the current time will be used, and the timezone will be set to local.
3. Else, the date metadata will not be written.

## Testing
Similar to #4636.

### (1) `#set document(date: none)`
```typst
// test.typ
#set document(date: none)
PDF
```
`strings test.pdf | grep Date` shows nothing.

### (2) `#set document(date: datetime(...))`
```typst
// test.typ
#set document(date: datetime(year: 2020, month: 1, day: 1))
PDF
```
`strings test.pdf | grep Date` shows
image

### (3) Default with `--creation-timestamp` (To ensure compilation reproducibility)
```
// test.typ
PDF
```
`strings test.pdf | grep Date` shows
img

### (4) Default without `--creation-timestamp`, and assume the tz is UTC+8
```
// test.typ
PDF
```
`strings test.pdf | grep Date` shows
image

### (5) Default without `--creation-timestamp`, and assume the tz is UTC-3:30
```
// test.typ
PDF
```
`strings test.pdf | grep Date` shows
img


wzf03:
> The core typst crates should never interact with the environment directly. They are pure libraries without OS interaction. As such, reading the timezone in `typst-pdf` is problematic.
> 
> This may sound contrary to my previous statement in #4636 that I'm concerned about moving the logic to typst-cli. But what I want is basically to have the logic itself in typst-pdf, but the timezone reading itself in typst-cli.

Thank your reply! I am refactoring the code to meet your requirement now.

About another detail, the #4636 make the UNIX timestamp provided through the CLI command to be shown in UTC. And I am not sure about whether this is necessary.

wzf03:
> The core typst crates should never interact with the environment directly. They are pure libraries without OS interaction. As such, reading the timezone in `typst-pdf` is problematic.
> 
> This may sound contrary to my previous statement in #4636 that I'm concerned about moving the logic to typst-cli. But what I want is basically to have the logic itself in typst-pdf, but the timezone reading itself in typst-cli.

Thanks for your reply! I have refactored the code to try to meet your requirement. Can you please review it again?

crates/typst-pdf/src/catalog.rs
```diff
@@ -87,8 +87,20 @@ pub fn write_catalog(
         xmp.pdf_keywords(&joined);
     }
 
-    let date = ctx.document.info.date.unwrap_or(ctx.options.timestamp);
-    let tz = ctx.document.info.date.is_auto();
+    // First, try to use the date from the document information, and it can
+    // be `none`. Second, try to use the date from the options. If neither is
```
laurmaedje, line 91:
```suggestion
    // be `auto`. Second, try to use the date from the options. If neither is
```

crates/typst-pdf/src/catalog.rs
```diff
@@ -87,8 +87,20 @@ pub fn write_catalog(
...
+    // available, we don't write a date metadata.
```
laurmaedje, line 92:
```suggestion
    // available, we don't write date metadata.
```

crates/typst-pdf/src/catalog.rs
```diff
@@ -306,15 +318,21 @@ fn pdf_date(datetime: Datetime, tz: bool) -> Option<pdf_writer::Date> {
...
     }
 
-    if tz {
-        pdf_date = pdf_date.utc_offset_hour(0).utc_offset_minute(0);
+    match tz {
+        Some(Timezone::Utc) => {
+            pdf_date = pdf_date.utc_offset_hour(0).utc_offset_minute(0)
+        }
+        Some(Timezone::Local { hour, minute }) => {
+            pdf_date = pdf_date.utc_offset_hour(hour).utc_offset_minute(minute as u8)
```
laurmaedje, line 326:
```suggestion
            pdf_date = pdf_date.utc_offset_hour(0).utc_offset_minute(0);
        }
        Some(Timezone::Local { hour, minute }) => {
            pdf_date = pdf_date.utc_offset_hour(hour).utc_offset_minute(minute);
```

crates/typst-pdf/src/catalog.rs
```diff
@@ -306,15 +318,21 @@ fn pdf_date(datetime: Datetime, tz: bool) -> Option<pdf_writer::Date> {
...
+        }
+        None => {}
     }
 
     Some(pdf_date)
 }
 
 /// Converts a datetime to an xmp-writer datetime.
-fn xmp_date(datetime: Datetime, tz: bool) -> Option<xmp_writer::DateTime> {
+fn xmp_date(datetime: Datetime, tz: Option<Timezone>) -> Option<xmp_writer::DateTime> {
```
laurmaedje, line 335:
```suggestion
fn xmp_date(datetime: Datetime, timezone: Option<Timezone>) -> Option<xmp_writer::DateTime> {
```

crates/typst-pdf/src/catalog.rs
```diff
@@ -323,6 +341,6 @@ fn xmp_date(datetime: Datetime, tz: bool) -> Option<xmp_writer::DateTime> {
         hour: datetime.hour(),
         minute: datetime.minute(),
         second: datetime.second(),
-        timezone: if tz { Some(Timezone::Utc) } else { None },
+        timezone: tz,
```
laurmaedje, line 344:
```suggestion
        timezone,
```

crates/typst-pdf/src/lib.rs
```diff
@@ -89,16 +89,43 @@ pub struct PdfOptions<'a> {
...
+}
+
+impl Timestamp {
+    /// Create a new timestamp with a given datetime and UTC suffix.
+    pub fn new_utc(datetime: Datetime) -> Self {
+        Self { datetime, timezone_offset: None }
+    }
+
+    /// Create a new timestamp with a given datetime, and a local timezone offset.
+    pub fn new_local(datetime: Datetime, whole_seconds_offset: i32) -> Option<Self> {
```
laurmaedje, line 119: I think `whole_minutes_offset` makes more sense, since we can't represent seconds anyway.

If `hours > 23` or `hours < -23` or `minutes > 59` it should return `None`.

crates/typst-pdf/src/lib.rs
```diff
@@ -89,16 +89,43 @@ pub struct PdfOptions<'a> {
...
 }
 
+/// A timestamp with timezone information.
+#[derive(Debug, Clone, Copy)]
+pub struct Timestamp {
+    /// The datetime of the timestamp.
+    pub(crate) datetime: Datetime,
+    /// If `Some`, the timezone offset from UTC in hours and minutes.
+    /// If `None`, the timestamp is in UTC.
+    pub(crate) timezone_offset: Option<(i8, i8)>,
```
laurmaedje, line 109: The types should be `(i8, u8)` like in pdf-writer since the timezone can just have one sign and two negative numbers make no sense. The sign the of the hour part should also apply to the minute part, like it's done in [pdf-writer](https://docs.rs/pdf-writer/latest/pdf_writer/struct.Date.html#method.utc_offset_minute). You'll need to add some logic in the constructor to convert a signed minute offset to this representation where the sign of the hours also semantically applies to the minutes.

xmp-writer takes `(i8, i8)`, but it's buggy if the second part is negative, so we'll change to `(i8, u8)` as well. But it's not a blocker for this PR, for now you can just as cast as i8 since `minutes <= 59`.

laurmaedje (commented):
Thanks, the general design is now like I imagined it. :)

I left some comments. In particular, I think the current logic is a bit buggy if we have a negative offset that isn't hour-aligned, because of the way pdf-writer & xmp-writer interpret the numbers vs. how your code does. E.g. `whole_seconds_offset = -5400` results in `(-1, -30)` for you, but pdf-writer expects `(-1, 30)` with the sign of `-1` applying to the `30` as well (hence the u8 type).

crates/typst-pdf/src/lib.rs
wzf03: OK, I will change it to `(i8, u8)` and try to test it with `Canada/Newfoundland` (with UTC −03:30 and UTC −02:30 DST).

crates/typst-pdf/src/catalog.rs
wzf03: What I actually want to express here is that if the `document.date` is set to specific `datetime` or `none`, it will be adapted first. And if it is `auto`, then the alternative cases will be considered. Maybe I should improve this comment.

laurmaedje:
Given the conversion complexity with negatives, I think it'd be good to have some unit tests for `Timezone::from_local`

laurmaedje:
Thanks, the code looks very good now!

wzf03:
> Thanks, the code looks very good now!

Thanks your advice, I learned a lot for it! ❤️

PR #5567, opened 2024-12-12
Add release target `loongarch64-unknown-linux-musl`

realybin:
LoongArch is an emerging RISC ISA developed by Loongson Technology, which has been mainlined in the Linux kernel. As LoongArch-based systems gain wider adoption, particularly in various computing environments, adding support for this architecture would benefit Typst users on these platforms.

### Details

- Adding loongarch64-unknown-linux-musl as a new release target
- Using musl variant for better portability
- Target is now supported in upstream cross-rs

### References

- [LoongArch Documentation](https://loongson.github.io/LoongArch-Documentation/README-EN.html)
- [Wikipedia: LoongArch](https://en.wikipedia.org/wiki/Loongson)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,9 @@ jobs:
         - target: riscv64gc-unknown-linux-gnu
           os: ubuntu-latest
           cross: true
+        - target: loongarch64-unknown-linux-musl
+          os: ubuntu-latest
+          cross: true
         - target: x86_64-apple-darwin
           os: macos-latest
           cross: false
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
     - name: Run Cross
       if: ${{ matrix.cross }}
       run: |
-        cargo install cross --git https://github.com/cross-rs/cross.git --locked --rev 085092ca
+        cargo install cross --git https://github.com/cross-rs/cross.git --locked --rev 4090beca
         cross build -p typst-cli --release --target ${{ matrix.target }} --features self-update,vendor-openssl
 
     - name: Run Cargo
--- a/crates/typst-cli/src/update.rs
+++ b/crates/typst-cli/src/update.rs
@@ -28,6 +28,7 @@ macro_rules! determine_asset {
             "aarch64-unknown-linux-gnu" => "aarch64-unknown-linux-musl",
             "armv7-unknown-linux-gnueabi" => "armv7-unknown-linux-musleabi",
             "riscv64gc-unknown-linux-musl" => "riscv64gc-unknown-linux-gnu",
+            "loongarch64-unknown-linux-gnu" => "loongarch64-unknown-linux-musl",
         })
     };
 


laurmaedje:
Honestly, this seems somewhat niche to me. Every added release target adds maintainance burden when things inevitably break, so I'm not sure I'm ready to commit to supporting this.

realybin:
I can take the responsibility for maintaining the LoongArch64 target. I'll monitor builds, test on real hardware, and handle any platform-specific issues that arise.

The LoongArch ecosystem is growing steadily, especially in academic and research environments where Typst would be valuable. You can count on me to keep this target well-maintained.

Coekjan:
Hi, you might also want to redirect the `loongarch64-unknown-linux-gnu` to `loongarch64-unknown-linux-musl` here:

https://github.com/typst/typst/blob/a3ad0a0bba3ce3abb1d8ed84656f54cc2d74be25/crates/typst-cli/src/update.rs#L20-L40

mohe2015:
What about best-effort support, so don't fail the release/merge if it doesn't build?

alerque:
I don't think this is as niche as it sounds, nor should the maintenance be considered a burden. In this case it is already trivially easy because it already builds just fine. That means some things have been done correctly and Rust is good at this. On the flip side if it were ever to break and the build stops working that would be a major red flag that something was being done *wrong*. Given that Typst already has a WASM target that limits what dependencies can be pulled in I don't think this is going to be an issue, but were it to ever flag a problem it would be worth a step back to consider why.

On the other hand I would probably hold off with the CI workflow until cross-rs has a release that supports this rather than having to pin an unreleased commit hash for it to work.

laurmaedje:
> I don't think this is as niche as it sounds, nor should the maintenance be considered a burden.

Well, perhaps niche is not the right term, but it's certainly not a mainstream architecture.

> Given that Typst already has a WASM target that limits what dependencies can be pulled in I don't think this is going to be an issue, but were it to ever flag a problem it would be worth a step back to consider why.

The Typst libraries support WASM, but not the CLI. It depends on non-Rust things like OpenSSL and things have broken before for that on other architectures.

---

Here's my view on the situation: On Linux (which afaik is the only common OS that runs on Loongarch), installing from a package manager or from source are just as common or even more common that downloading pre-built binaries. In contrast to many C++ based projects, Typst is a breeze to build from source yourself, which in my opinion reduces the need for pre-built binaries. Likewise, inclusion in a package manager does not require us to provide pre-built binaries.

The pre-built binaries are thus merely a convenience offering from us. The absence of a pre-built binaries doesn't imply that Typst can't be used on another system or architecture, it just means that we do not consider it frequent enough to warrant first-party support. 

There is a theme across this project: Can we have this one extra math operator, this one extra image format, this one extra release binary, etc.? And, in isolation, it's hard to argue why adding this one extra thing is a problem. But in cummulation, things do add up, so a line needs to be drawn somewhere.

We'd be glad if Loongarch Linux distribution maintainers want to include Typst in their package repositories. We just don't think it is currently common enough for us to provide first-party support. I think, if anything, we already support a few platforms too much, which have almost no downloads (armv7 & riscv64). Same with Docker, Nix & Snap: Even though Typst is very easy to build, things _do_ break and it _does_ add maintainance burden. The Docker image has quite a few downloads, so it's probably worth it. With Snap & Nix, I'm less sure whether they pull their weight. 

For what it's worth, none of the major Linux distributions support Loongarch either, as far as I can tell. Perhaps things will change on that front down the road.

Let me end by saying that I do appreciate your generous offer of taking on responsibility for maintaining the target. The thing is that I'd like to avoid adding support for things which we do not understand, can't test, or simply don't work on ourselves. It's not the end of the world, but my experience as a maintainer so far have shown for it to be a burden in review, pre-release testing, etc.

PR #5570, opened 2024-12-13
Move `CellGrid` to `typst_library::layout::Raster` and refactor

01mf02:
HTML tables/grids will require a lot of functionality that is currently implemented in `typst_layout`. Because HTML export should not depend on `typst_layout`, this PR moves that functionality from `typst_layout` to `typst_library`. At the same time, this PR perform a bit of refactoring; in particular, it breaks the function `Raster::resolve` (previously `CellGrid::resolve`) into smaller sub-functions to improve the readability of the function.


crates/typst-library/src/layout/raster.rs
```diff
@@ -0,0 +1,1788 @@
+//! Unified handling for tables and grids.
```
PgBiel, line 1: To be honest, I'm not sure if I feel comfortable with all the "raster" naming. It is evocative of a series of pixels and not cells, and it's not immediately clear what it represents. While I agree the previous name of `CellGrid` wasn't the best, it did include the fact that it's essentially just a grid. We're basically taking the grid element and transforming it into a more definitive form. And for tables, it's the same thing, as tables are laid out as grids, so it makes sense that they are turned into one.

So I think something like `ResolvedGrid` would be better, and also more consistent with the relevant structures around it. We'd then create a folder `grid/{mod.rs, resolve.rs}`, and not perform significant changes elsewhere. I think this makes for a cleaner architecture. Let me know your thoughts.

crates/typst-library/src/layout/raster.rs
```diff
@@ -0,0 +1,1788 @@
...
+
+/// Position of a cell inside a raster.
+///
+/// Given an index `i` and the number of columns `c`,
+/// the x-coordinate of `i` is `i % c` and
+/// the y-coordinate of `i` is `i / c`.
+type Idx = usize;
+
+#[derive(Default)]
+struct State<'a> {
```
PgBiel, line 636: These refactors to grid resolving, such as with the introduction of `State`, is not really necessary for your goal, right? If that's the case, could we please move these changes to a separate PR? That'd make it much simpler to review and discuss this PR. Thank you.

PgBiel (commented):
Thanks. Some initial thoughts:

crates/typst-library/src/layout/raster.rs
01mf02: I agree with you. I have opened a new PR #5585 that tries to focus on only moving code.

crates/typst-library/src/layout/raster.rs
01mf02: I see your point about "raster". In my new PR, I have focused on making as few changes as possible, so I kept the existing `CellGrid` name and added `cellgrid` as a new module, leaving the existing `layout/grid.rs` alone.

laurmaedje:
I'll close this as the refactoring will not be finished and there are many conflicts.

PR #5571, opened 2024-12-13
Change error when accessing non-existant label

### LINKED ISSUE #5569: Internal error when accessing a missing "label" field of any element
### Description

Accessing the `.label` field of seemingly any element (that doesn't have a label) causes an error that _isn't_ `xxx does not have field "label"`. According to the message itself, it is a bug. Googling the precise message yielded no results, but I hope this isn't a duplicate, it looks almost too simple for me to be the one to discover it.

Minimal example: `[].label`
This gives the following error: `internal error when accessing field "label" in sequence – this is a bug`

You can also try `strong[].label` or `table().label` for more element kinds. Same issue.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

joshuagawley:
Fixes #5569 by returning a `FieldAccessError::Unknown` instead of a `FieldAccessError::Internal` if we try to access the label of an element that does not have a label (and thus outputting the message `[x] does not have field "label"`).

_Prima facie_, this behavior seems more helpful than the existing behavior but if there is a reason for the existing behavior then do let me know.

### DIFF
--- a/crates/typst-library/src/foundations/content.rs
+++ b/crates/typst-library/src/foundations/content.rs
@@ -211,9 +211,10 @@ impl Content {
     /// instead.
     pub fn get_by_name(&self, name: &str) -> Result<Value, FieldAccessError> {
         if name == "label" {
-            if let Some(label) = self.label() {
-                return Ok(label.into_value());
-            }
+            return self
+                .label()
+                .map(|label| label.into_value())
+                .ok_or(FieldAccessError::Unknown);
         }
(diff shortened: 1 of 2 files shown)


laurmaedje:
Honestly, I think we can just get rid of the `Internal` variant altogether as done [here](https://github.com/typst/typst/pull/4954). That was [my intial intuition](https://github.com/typst/typst/pull/4399#discussion_r1641669072) as well, but I let myself convince of the extra variant back when it was added.

laurmaedje:
Thanks!

PR #5572, opened 2024-12-13
Document result of splitting string with empty string as separator

### LINKED ISSUE #5443: Document the result of `string.split("")`
### Description

image

Since there is a difference between clusters and bytes and Unicode code points, it would be great if the docs says how exactly it would split the string with an empty delimiter.

joshuagawley:
Provisionally fixes #5443.

I can add an example of Typst code to clarify if requested.

### DIFF
--- a/crates/typst-library/src/foundations/str.rs
+++ b/crates/typst-library/src/foundations/str.rs
@@ -575,6 +575,10 @@ impl Str {
 
     /// Splits a string at matches of a specified pattern and returns an array
     /// of the resulting parts.
+    ///
+    /// When the empty string is used as a separator, it separates every
+    /// character in the string, along with the beginning and end of the
+    /// string.
     #[func]
     pub fn split(
         &self,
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/foundations/str.rs
```diff
@@ -575,6 +575,10 @@ impl Str {
 
     /// Splits a string at matches of a specified pattern and returns an array
     /// of the resulting parts.
+    ///
+    /// When the empty string is used as a separator, it separates every
+    /// character in the string, along with the beginning and end of the
+    /// string.
```
laurmaedje, line 581: I'm not sure the second sentence makes it 100% clear that two empty strings result at the start and end.

laurmaedje:
Thanks!

Andrew15-5:
What is a "character" here? A Unicode codepoint?

laurmaedje:
Yes

Andrew15-5:
See https://github.com/typst/typst/pull/5616.

Issue #758, opened 2023-04-12
Link in combination with `repeat[.]` is very inefficient and produces bad PDF


Holt59:
**Update:** I am updating this issue to propose a minimal reproducible example and to target more appropriately the problem. The original issue is still available below.

The problem occurs when creating `link()` that includes `box()` with `repeat[.]`, typically when creating outline (the default `typst` outline does not include the `...` in the link(), but having a link over the whole line would be way better in my opinion).

Here is the example (this is not a outline, but shows the actual issue):

```typst
#locate(loc => {
  for (i, n) in lorem(20).split().enumerate() {

    // link does not contain box, size is ~43kB
    link(loc, n)
    box(width: 1fr, h(3pt) + box(width: 1fr, repeat[.]) + h(3pt))
    link(loc, str(i))

    // link contains box, size is ~767kB
    link(loc, {
      n
      box(width: 1fr, h(3pt) + box(width: 1fr, repeat[.]) + h(3pt))
      str(i)
    })

    linebreak()
  }
})
```

This code contains two different way of creating an "outline-like" entry, one that only `link()` the title and the number (title / page in outline), and one that link the whole line. When creating a link() over the whole line, the outline becomes almost unusable in the PDF, all the readers I tested have a hard-time displaying the outline and the size of the PDF is ridiculous.

I think the issue is because `link()` will basically create a "PDF link" for every dot, instead of the whole line (this can be checked in the PDF since each dot is clickable independently).

---

**Original issue:**

I cannot show the full document, but I am basically creating a custom outline inspired by the one provided here: https://github.com/typst/typst/issues/488

The generated outlines contain ~70 items of various depth, but using `repeat[.]` or not drastically increase the PDF size and makes my PDF reader (Adobe Acrobat) very slow when looking at the outline.

Basically:

```typst
// PDF size is ~620kB, everything is smooth in adobe
box(width: 1fr, h(3pt) + box(width: 1fr) + h(3pt))

// PDF size is `2450kB, Adobe struggle with the outline
box(width: 1fr, h(3pt) + box(width: 1fr, repeat[.]) + h(3pt))
```

The line is above is **the only line** I change between two compilation.

If I do not link the whole line, the size is also close to the initial one, so I guess the issue is that a link is created for every element in the line (or something like that)?

lucifer1004:
I just met the same problem today and it took me hours to figure out the root cause.

cyberbryce:
[...]

PR #5575, opened 2024-12-13
Derivation comment for calculation in `repeat`

Andrew15-5:
While I was writing the code for https://github.com/typst/typst/issues/5095#issuecomment-2541953110, I found a few things in the source code that can be improved. I was very confused by the `frame.push_frame(Point::with_x(offset), piece.clone());` part, but I guess it's just how you have to work with `Frame`s, which I'm not familiar with and which is very different from how you would do this in Typst.

By the way, can `move(dx: offset)` + `text(tracking: gap, range(count).map(_ => body).join())` be done in Rust? If so, then https://github.com/typst/typst/issues/758 can easily be fixed.

### DIFF
--- a/crates/typst-layout/src/repeat.rs
+++ b/crates/typst-layout/src/repeat.rs
@@ -33,8 +33,17 @@ pub fn layout_repeat(
     let fill = region.size.x;
     let width = piece.width();
 
-    // count * width + (count - 1) * gap = fill, but count is an integer so
-    // we need to round down and get the remainder.
+    // We need to fit the body N times, but the number of gaps is (N - 1):
+    // N * w + (N - 1) * g ≤ F
+    // where N - body count (count)
+    //       w - body width (width)
+    //       g - gap width (gap)
+    //       F - available space to fill (fill)
+    //
+    // N * w + N * g - g ≤ F
+    // N * (w + g) ≤ F + g
+    // N ≤ (F + g) / (w + g)
+    // N = ⌊(F + g) / (w + g)⌋
     let count = ((fill + gap) / (width + gap)).floor();
     let remaining = (fill + gap) % (width + gap);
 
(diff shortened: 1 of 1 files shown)


PgBiel (approved):

PgBiel:
> By the way, can `move(dx: offset) + text(tracking: gap, range(count).map(_ => body).join())` be done in Rust? If so, then https://github.com/typst/typst/issues/758 can easily be fixed.

A crucial detail in your original post is that `body` was a string (`"."`), not content, so `join()` produced a single string (`"......"`), creating a single text node. This will not fix it for arbitrary content.

Andrew15-5:
> A crucial detail in your original post is that `body` was a string (`"."`), not content, so `join()` produced a single string (`"......"`), creating a single text node.

That's true, but the fact is that the huge file size issue was fixed either way and I can put any content, and it will be repeated the correct number of times.

I have tested a few things and for a complex content this indeed doesn't work correctly:

```typ
#set outline(fill: repeat[.])
#set outline(fill: repeat[#emoji.a])
#set outline(fill: repeat[#context repr(text.size)])
#set outline(fill: repeat[#layout(s => repr(s.width))])
#set outline(fill: repeat[#box(width: 20pt, layout(s => repr(s.width)))])
#set outline(fill: repeat[#box(width: 20pt, layout(s => repr(s.width)))])
#set outline(fill: std.repeat[.])
#set outline(fill: std.repeat[#emoji.a])
#set outline(fill: std.repeat[#context repr(text.size)])
#set outline(fill: std.repeat[#layout(s => repr(s.width))])
#set outline(fill: std.repeat[#box(width: 20pt, layout(s => repr(s.width)))])
```

So I can tinker with other ways of evenly spacing the big content. Thanks. The good part is that it works as expected in the 99.999% of use cases, as majority will just use `[.]`.

laurmaedje:
Thanks, that `((fill + gap) / (width + gap)).floor()` equation was indeed a bit hard to get before!

Andrew15-5:
I actually at first thought that it was wrong because it didn't make sense to me, but after getting the whole chain of thought and simplification it made perfect sense.

PR #720, opened 2023-04-10
add `sinc` operator

classabbyamp:
this is for the [sinc function](https://en.wikipedia.org/wiki/Sinc_function)

PR #5577, opened 2024-12-13
Added `cosec` and `cosech` functions

### LINKED ISSUE #5576: Add the `cosec` operator.
### Description

Add `cosec` alongside the [predefined operators](https://typst.app/docs/reference/math/op#predefined) in maths mode.

### Use Case

At least in the UK, the cosecant function is commonly written as `cosec(x)`. Adding it as a predefined operator seems fairly uncontroversial (the rest of the world keeps using `csc(x)` and is none the wiser) but those in the UK can use `cosec(x)` without having to define it manually.

John-Toohey:
Fixes #5576.

### DIFF
--- a/crates/typst-library/src/math/op.rs
+++ b/crates/typst-library/src/math/op.rs
@@ -16,10 +16,10 @@ use crate::text::TextElem;
 ///
 /// # Predefined Operators { #predefined }
 /// Typst predefines the operators `arccos`, `arcsin`, `arctan`, `arg`, `cos`,
-/// `cosh`, `cot`, `coth`, `csc`, `csch`, `ctg`, `deg`, `det`, `dim`, `exp`,
-/// `gcd`, `hom`, `id`, `im`, `inf`, `ker`, `lg`, `lim`, `liminf`, `limsup`,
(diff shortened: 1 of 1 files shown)


Enivex:
There are many of these regional variants. If we start adding them we'd have to start adding them all, and I really don't see the point.

Defining operators manually is not very difficult.

laurmaedje:
I agree with @Enivex that we can't add support for all of them. It's possible that the current selection of built-in ones isn't the most consistent one, and if someone were to take on the work of surveying the most common operators across different fields to make a new, potentially more consistent selection, I'd be open to discussing that. But in the absence of that, I'd put a freeze on adding new individual operators.

Previous art: 
- https://github.com/typst/typst/pull/1090
- https://github.com/typst/typst/pull/4504
- https://github.com/typst/typst/pull/2378
- https://github.com/typst/typst/pull/1740
- https://github.com/typst/typst/pull/720

John-Toohey:
I can understand and respect this decision. However, would it not be a good idea to add some sentence to the documentation, something along the lines of "Typst keeps the number of predefined operators intentionally small, as to[...] . See _here_ for how to define an operator" (given the number of issues you seem to get with this)?

laurmaedje:
Sure we can do that

Issue #456, opened 2023-03-30
Import remote resources


palainp:
Dear developers,
Is there a possibility to permit `import` (or even `include`) on remote online resources? It currently fails, I tried to update `src/eval/mod.rs:import` but without any luck so far. It seems to rip `//' I have the following and I miss a way to read remote content:
```
$ cat empty.typ 
#import "https://github.com/typst/typst/blob/main/tests/typ/layout/align.typ": *

== Dummy test
$ ./target/release/typst empty.typ 
error: file not found (searched at https:/github.com/typst/typst/blob/main/tests/typ/layout/align.typ)
  ┌─ empty.typ:1:8
  │
1 │ #import "https://github.com/typst/typst/blob/main/tests/typ/layout/align.typ": *
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
This would be useful for having always up-to-date community templates.

kanashimia:
There were some discussions about package management in discord channel, it is a complex thing to implement right, so probably won't be as simple as just pasting a link.

Right now If you use git and the templates are in the git repo, then you can use git submodules to include those templates, or write a script that fetches them.

Issue #4040, opened 2024-04-30
Add support for packages hosted in arbitrary repositories


BattleCh1cken:
### Description

It would be amazing if Typst supported git packages in addition to local and preview packages. This would allow users to download unofficial packages easily without the hassle of needing to use install scripts or the like. 

The syntax could look something like this:

```typ
#import "@git/battlech1cken/notebookinator:1.0.0"
```

This would then automatically download the package from that repository and make sure its in the correct directory, automatically excluding the files specified in the `typst.toml` file.

There are two major problems with this feature that I haven't really thought about yet.
- git packages obviously can't be reviewed by Typst team, leading to possible security problems.
-  I'm not sure how the web app will handle these types of packages, because I know that network requests there should be kept to a minimum.

### Use Case

Currently I maintain a library called [The Notebookinator](https://github.com/BattleCh1cken/notebookinator) which is rather large. For this reason I doubt it will be accepted as an official package, and I've had to resort to having users install it as a local package. I've included a script which allows for automatic installation, similar to the [CeTZ](https://github.com/cetz-package/cetz) package.

This is less than ideal for my users who now have to go through several steps to download my library, whereas official packages are incredibly easy to acquire.

This problem only gets worse if I need to modify or use unreleased versions of other packages I depend on, users or fellow developers will need to download all of those packages locally as well.

Enter-tainer:
iirc it is not supported by design.(correct me if i'm wrong)

istudyatuni:
It's planned, see [this message](https://discord.com/channels/1054443721975922748/1176122103355953162/1176125324430413955) in discord's \#forge channel

laurmaedje:
Git imports would need to be handled with a lot of care to ensure Typst remains safe to run on untrusted input. I've written a longer comment explaining things here: https://github.com/typst/typst/pull/5579#issuecomment-2545034781

JL710:
I think the most severe vulnerability that is mentioned in https://github.com/typst/typst/pull/5579#issuecomment-2545034781 is, that the network/downloads could be used to send data from the user to others by calling imports.
I think a way to prevent that would be to explicitly ask the user for any library that requires a download, if it should be downloaded. Once it is downloaded and in local storage, the vulnerability should be gone right?

laurmaedje:
In principle yes, but I'd still like to avoid _dynamic_ network imports. See my comment here: https://github.com/typst/typst/pull/5579#issuecomment-2573103519

RossSmyth:
[...]

PR #5579, opened 2024-12-14
Add the possibility to download remote packages without them to be uploaded to typst universe

tetofonta:
This PR aims to implement the declaration of remote packages which are not published on typst universe and probably allows for what has previously been discussed in https://github.com/typst/typst/issues/456.

This enables the creation of "private package repository" and not to rely on one single package provider.

The whole thing is achieved by encoding the remote location and download scheme of the package in the package namespace on import in the following format.
```typst
#import "@<scheme>:<path>/<name>:<version>"
```

The relative code has been refactored to allow the selection and implementation of multiple "download methods" which can then dynamically be used for package retrieval if the latter is not already cached or in the local packages path.

 Currently two download methods are implemented:
  - HTTP(S)
  - GIT

## HTTP(S)

Defines the package as available in a remote web registry following the specifications of the current running typst universe.
The package can be imported with the following syntax:
```typst
#import "@http[s]:<registry host>:<namespace>/package-name>:<package-version>"
```
resulting in the package location to be resolved as `http[s]://<registry host>/<namespace>/<package-name>-<package-version>.tar.gz` as already implemented in the mainstream version. Also, the repository index is resolved as usual to `http[s]://<registry host>/<namespace>/index.json`

The `preview` namespace is treated as an `http` package under the hood, resulting in the same beahvior as defining the imports of the `preview` namespace as `@https:packages.typst.org:preview/package-name>:package-version`. 
Of course , the `preview` namespace is handled as a shortand of the latter and does not change the current program behavior.

## GIT

Defines the package as available in a remote repository with the following layout:
 - Each package version commit must have a tag in the form `v<major>.<minor>.<patch>`
 - The commit must present a `typst.toml` file in its root.

If the repository satisfies the previously mentioned constraints, the package can be imported as
```typst
#import "@git:<git user>@<git-host>:<git username>/<package-name>:<package-version>"
```

The code will then clone the repository into the cache directory and will the set its head (in a detached state) to the tag derived by the version.

For example, the package
```
@git:git@github.com:typst/package:0.1.0
```
will result in the cloning of the repository `git@github.com:typst/package.git` and the checkout and detached head state at tag `v0.1.0`. Then the usual package handling is performed as usual.

## TODO

- [ ] Make tests for the new package namespace handling
- [ ] In-depth tests for the download methods
- [ ] Allow for multiple `/` in the package import syntax to allow subpath registries
- [ ] Display progress during git cloning

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ typst-timing = { path = "crates/typst-timing", version = "0.12.0" }
 typst-utils = { path = "crates/typst-utils", version = "0.12.0" }
 typst-assets = { git = "https://github.com/typst/typst-assets", rev = "8cccef9" }
 typst-dev-assets = { git = "https://github.com/typst/typst-dev-assets", rev = "b07d156" }
-auth-git2 = "0.5.5"
 arrayvec = "0.7.4"
 az = "1.2"
 base64 = "0.22"
@@ -59,7 +58,7 @@ env_proxy = "0.4"
 flate2 = "1"
 fontdb = { version = "0.21", default-features = false }
 fs_extra = "1.3"
-git2 = "0.19.0"
+gix = "0.68.0"
 hayagriva = "0.8"
 heck = "0.5"
 hypher = "0.1.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ typst-timing = { path = "crates/typst-timing", version = "0.12.0" }
 typst-utils = { path = "crates/typst-utils", version = "0.12.0" }
 typst-assets = { git = "https://github.com/typst/typst-assets", rev = "8cccef9" }
 typst-dev-assets = { git = "https://github.com/typst/typst-dev-assets", rev = "b07d156" }
+auth-git2 = "0.5.5"
 arrayvec = "0.7.4"
 az = "1.2"
 base64 = "0.22"
@@ -58,6 +59,7 @@ env_proxy = "0.4"
 flate2 = "1"
 fontdb = { version = "0.21", default-features = false }
 fs_extra = "1.3"
+git2 = "0.19.0"
 hayagriva = "0.8"
 heck = "0.5"
 hypher = "0.1.4"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,6 +15,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
+ "getrandom",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -28,6 +29,12 @@ checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "allocator-api2"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "android-tzdata"
@@ -112,6 +119,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -128,17 +141,6 @@ name = "ascii"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d92bec98840b8f03a5ff5413de5293bfcd8bf96467cf5452609f939ec6f5de16"
-
-[[package]]
-name = "auth-git2"
-version = "0.5.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3810b5af212b013fe7302b12d86616c6c39a48e18f2e4b812a5a9e5710213791"
-dependencies = [
- "dirs",
- "git2",
- "terminal-prompt",
-]
 
 [[package]]
 name = "autocfg"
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -58,6 +59,7 @@ env_proxy = "0.4"
 flate2 = "1"
 fontdb = { version = "0.21", default-features = false }
 fs_extra = "1.3"
+git2 = "0.19.0"
```
alerque, line 62: This is a somewhat problematic dependency. It tends to have lots of breaking changes before releases, and distro packaging for it is kind of a pain. Doable, and at least at Arch Linux we do it for a lot of projects, but as a hint to projects I'd suggest not if possible. If the *gitoxide* crate(s) can be used instead the pure Rust implementation there is both faster and more ergonomic. It isn't as feature complete as the libgit2 bindings, but if it covers what you need they make more sense to use unless you have a need for the C bindings.

tetofonta: That's a very good point. I'll try change the git implementation in favor of `gitoxide`... Worst case scenario i foresee is that not every functionality is available, so I'll evaluate the possibility of jumping the `git2` dependency and call the system git executable directly. I admit not the purest and elegant solution but it may work as good.

alerque: > and call the system git executable directly

As I understand it that is not a feasible fallback for Typst. Dependency on a system CLI tool would wreak havoc on the sandboxed model (however weak it may already be) and the WASM compatibility. It would quickly get you into a position where some projects would build in some environments but not others, something Typst has done pretty good at minimizing so far.

alerque:
Something else to consider is making sure that this *cannot* be used to reach `file://` or `/` or other local URL schemes--at least not without some project flag that said it could do so. That would provide a way to break out of the project level sandboxing Typst has been enforcing to date.

Cargo.toml
tetofonta: That's true. Turns out `gitoxide` was enough for what was needed.

tetofonta:
Still need to write test for it, but given the syntax restriction on the package import I'm mostly sure it could not be done.

For the HTTP case, the scheme is forced to be `http` or `https` in  [crates/typst-kit/src/package_downloads/mod.rs#L169](https://github.com/tetofonta/typst/blob/3962be8ebfcdf68648a4e18ee4780a7e993f6192/crates/typst-kit/src/package_downloads/mod.rs#L169). I can imagine adding checks on the hostname to be [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123) valid (maybe using `hostname-validator` crate). Also the registry url format is hardcoded to be the same style of the current typst universe.

For the git case, i don't know how wasm and file system will interact.

Also, there cannot be any `/` character in the import string other than the one dividing namespace and package. The latter is then ensured to be an ident, so no special characters can be inserted there.

Please feel free to suggest any kind of checks to be implemented in the validation steps.

LaurenzV:
Just to check, was it actually discussed/approved to implement this feature with such a design, or was this your own idea? 

If the latter, I would maybe hold off on working on it further and wait for some first feedback from @laurmaedje, as it seems like a pretty big change that does warrant some discussion whether we want this in the first palce, and it would be pretty unfortunate if ends up being rejected after putting so much effort into it.

laurmaedje:
This is unfortunately quite problematic for security reasons. Typst puts a strong focus on being safe to run on untrusted input. There are two sides to this:

- **Input:** We try to make sure that Typst code has as little access to files on your computer as possible. For this reason, you can only read files within the root directory of your project.

- **Output:** Even if your Typst code managed to read secret information, we aim to make it as hard as possible to exfiltrate that information.

Both sides are important because both have some inherent weaknesses:

- On the input side, it's sometimes hard to define what the sandbox should even be! If the users compiles in their home directory, the project sandbox is all of their home directory (which is a lot). If the project contains a symlink pointing out of it, that might have been intentional, or it could be part of an attack (though transferring a symlink to a person is already more complicated than a plain file).

- On the output side, we are in a pretty good position because Typst forbids any kind environment interaction. The only exfiltration route I'm currently aware of is by embedding the information directly in the PDF and getting the user to send it to you. That requires some social engineering, but it's not impossible. It's something we can't really prevent, underlining the importance of sandboxing on the input side.

So with these things in mind, it may become clear why this PR is problematic: It opens up an exfiltration route on the output side. Concretely, in this implementation, if a script gets ahold of valuable information, it could encode that information as a base64 package name and then try to import that from a custom server that processes that information. This means effectively a total breakdown of our security promises.

Now, upon thinking further, one might come up with the idea of only allowing static string literals in import statements: Then, the string can't be data dependant and we're good, right?

Unfortunately we aren't. Let me give one example of a possible attack: You prepare an array with 400 functions containing import statements á la `@git:git@myserver.com/myuser/package:0.0.1` .. `@git:git@myserver.com/myuser/package:0.0.400`. Then, you transmit a byte sequence by calling the function at the index corresponding to that byte, and then removing that index from the array. That way, you can unambigiously transmit $400-256+1$ bytes. There are probably myriads of further ways that could be used to exfiltrate information (e.g. timings attacks). In short: Any kind of code-triggered remote fetching to an untrusted server is very problematic.

I'm not yet certain of its safety, but there is at least an _idea_ for how to support Git imports without such problems. The idea is to only support Git or other kinds of remote non-Typst-Universe imports in a `typst.toml` manifest. That way, we can fetch all Git dependencies upfront (including, very importantly, transitive ones!) with no data-dependance and then do not allow for any network access during code evaluation. We've [discussed it a bit on Discord](https://discord.com/channels/1054443721975922748/1176122103355953162/1191795494326906880) before, but it's still very much in an early design stage, so more discussion is necessary!

This would probably still be opt-in, since it's still a network access and the remote party would still know _that_ you compiled the document, but I think it would solve the data exfiltration problem.

tetofonta:
One way I can see to overcome these problems is having the user explicitly allowing network access for requests directed to a given host, maybe in a configuration file in the root or at system level or by cli flags.

laurmaedje:
Yeah, though I think even it this case it would still be good to prevent any kind of dynamic (i.e. code driven) network access. The two things seem complementary to me:
- Prevent dynamic access: Prevent data exfiltration, even if a host is trusted
- Ask before network access: Prevent any kind of tracking

I think there are four useful levels of configuration:

1. Don't allow any network access (not even `@preview`)
2. Always ask before downloads
3. Don't ask for trusted downloads (i.e. `@preview`), but ask for non-trusted downloads (i.e. git), which would be the default
4. Allow all downloads

More fine-grained per-host configuration could indeed also be an option.

laurmaedje:
I'm happy to further discuss a design for Git dependencies, but I think a PR is not the best place to do it. Better places would be the [Discord forge](https://discord.com/channels/1054443721975922748/1176122103355953162) and the [relevant issue](https://github.com/typst/typst/issues/4040). Once we've settled on a design, we can perhaps reuse parts of this PR, but for now I would close it. Thanks for your understand and also thanks for your work still!

PR #5585, opened 2024-12-16
Move `CellGrid` from `typst-layout` to `typst-library`.

01mf02:
HTML tables/grids will require a lot of functionality that is currently implemented in `typst_layout`. Because HTML export should not depend on `typst_layout`, this PR moves that functionality from `typst_layout` to `typst_library`.

### DIFF
--- a/crates/typst-library/src/layout/mod.rs
+++ b/crates/typst-library/src/layout/mod.rs
@@ -4,6 +4,7 @@ mod abs;
 mod align;
 mod angle;
 mod axes;
+pub mod cellgrid;
 mod columns;
 mod container;
 mod corners;
--- a/crates/typst-layout/src/grid/layouter.rs
+++ b/crates/typst-layout/src/grid/layouter.rs
@@ -3,6 +3,7 @@ use std::fmt::Debug;
 use typst_library::diag::{bail, SourceResult};
 use typst_library::engine::Engine;
 use typst_library::foundations::{Resolve, StyleChain};
+use typst_library::layout::cellgrid::{Cell, CellGrid, LinePosition, Repeatable};
 use typst_library::layout::{
     Abs, Axes, Dir, Fr, Fragment, Frame, FrameItem, Length, Point, Region, Regions, Rel,
     Size, Sizing,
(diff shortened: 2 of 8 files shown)


crates/typst-library/src/layout/mod.rs
```diff
@@ -4,6 +4,7 @@ mod abs;
 mod align;
 mod angle;
 mod axes;
+pub mod cellgrid;
```
PgBiel, line 7: FWIW, if you agree with my proposal of using `ResolvedGrid` and moving to `typst-library/src/layout/grid/{mod.rs, resolve.rs}`, I'd be ok with doing it in this PR as it's a (comparatively) pretty minor change, so you can focus on your proposed refactors on the other PR.

PgBiel (commented):
Looks fine in general, thank you.

crates/typst-library/src/layout/mod.rs
01mf02: I think that it's better to keep the `CellGrid` name and structure for now, to keep this PR minimally invasive. For later, I think that it might be nice to have a module `typst-library/src/layout/resolved/`, where `CellGrid` and other resolved types could live, if there are enough to justify this. But that is just an idea for now.

PgBiel: Sure. What do you think about at least changing to `grid/{mod.rs, resolve.rs}`, otherwise keeping `CellGrid` as is? I really dislike having something like `typst-library/layout/cellgrid.rs`, at least imo.

01mf02: OK, I did this, @PgBiel. However, because we have `pub use self::grid::*;` in `typst-library/layout/mod.rs`, this means that we now expose `layout::resolved`. We can change this by doing `pub use self::grid::{Grid, GridItem, ...};` instead, but this is a bit verbose (because `grid` contains quite some types), and no other module in `layout` exports its items explicitly. What do you think?

PgBiel: Yeah, this is basically the goal of keeping it separate, as we don't need to use the `resolve` module in many places, only in typst-layout/grid basically.

Also, not sure if that was intentional but you named the module `resolved`, but I think `resolve` is more appropriate as the main functions it exports convert grids, tables and cells to their resolved counterparts. Do you think you could rename it? Then I think there won't be anything else to change.

01mf02: OK, I just renamed the module to `resolve` a7c0364e. Happy new year, by the way. ;)

laurmaedje:
Thanks! While I'm myself still not sure what file system layout I like best, in the interest of stopping the bikeshedding and shipping stuff, I'll just merge now. We can always change it further in the future.

PR #5587, opened 2024-12-16
Error on duplicate symbol variant with modifiers in different orders

MDLC01:
As suggested in https://github.com/typst/typst/pull/5505#discussion_r1877486255, this should be an error. Previously, an error would happen only if the duplicate variants had the modifiers in the same order.

### DIFF
--- a/crates/typst-library/src/foundations/symbol.rs
+++ b/crates/typst-library/src/foundations/symbol.rs
@@ -231,7 +231,7 @@ impl Symbol {
             if let Some(ms) = modifiers.windows(2).find(|ms| ms[0] == ms[1]) {
                 bail!(
                     span, "duplicate modifier within variant: {}", ms[0].repr();
-                    hint: "the modifiers of a variant constitute a set, repetition does not matter"
+                    hint: "modifiers are not ordered, so each one may appear only once"
                 )
             }
 
@@ -245,7 +245,7 @@ impl Symbol {
                 } else {
                     bail!(
                         span, "duplicate variant: {}", v.0.repr();
-                        hint: "the modifiers of a variant constitute a set, order does not matter"
+                        hint: "variants with the same modifiers are identical, regardless of their order"
                     )
                 }
             }
--- a/crates/typst-library/src/foundations/symbol.rs
+++ b/crates/typst-library/src/foundations/symbol.rs
@@ -203,8 +203,17 @@ impl Symbol {
             bail!(span, "expected at least one variant");
         }
         for Spanned { v, span } in variants {
-            if list.iter().any(|(prev, _)| same_modifiers(&v.0, prev)) {
-                bail!(span, "duplicate variant");
+            if let Some((prev, _)) =
+                list.iter().find(|(prev, _)| same_modifiers(&v.0, prev))
+            {
+                if &v.0 == prev {
+                    bail!(span, "duplicate variant: {:?}", v.0);
+                } else {
+                    bail!(
+                        span, "duplicate variant: {:?}", v.0;
+                        hint: "modifiers are a set, meaning order and repetition do not matter"
+                    )
+                }
             }
             if !v.0.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/foundations/symbol.rs
```diff
@@ -198,12 +198,12 @@ impl Symbol {
...
     ) -> SourceResult<Symbol> {
-        let mut list = Vec::new();
+        let mut list = Vec::<(EcoString, _)>::new();
         if variants.is_empty() {
             bail!(span, "expected at least one variant");
         }
         for Spanned { v, span } in variants {
-            if list.iter().any(|(prev, _)| &v.0 == prev) {
+            if list.iter().any(|(prev, _)| same_modifiers(&v.0, prev)) {
                 bail!(span, "duplicate variant");
```
PgBiel, line 207: Maybe we should improve the error message (and/or hints) to point out which variant is duplicate? And also include a hint, if `&v.0 != prev` (we can check this by replacing `.any` with `.find`), that the order of variants doesn't matter, so people who were confused like me don't remain so afterwards.

MDLC01: The error message is already at the location of the duplicated variant, so specifying which variant is duplicated is redundant. I agree a hint should be shown when the modifiers are specified in different orders.

laurmaedje: This is a bit of a common theme. We've generally landed on always specifying it redundantly since the argument could always be a variable. It also makes the error message a bit more useful standalone.

MDLC01: That makes sense

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -198,13 +198,22 @@ impl Symbol {
...
-        let mut list = Vec::new();
+        let mut list = Vec::<(EcoString, _)>::new();
         if variants.is_empty() {
             bail!(span, "expected at least one variant");
         }
         for Spanned { v, span } in variants {
-            if list.iter().any(|(prev, _)| &v.0 == prev) {
-                bail!(span, "duplicate variant");
+            if let Some((prev, _)) =
+                list.iter().find(|(prev, _)| same_modifiers(&v.0, prev))
```
laurmaedje, line 207: A bit unfortunate to have quadratic complexity on the find here. Together with the complexity of `same_modifiers` we are on $O(n^2 m^2)$ where $n$ is the number of variants and $m$ the number of modifiers.

MDLC01: An alternative is to sort the modifiers in `list`, so that we can use binary search. But this will require additional allocations.

crates/typst-library/src/foundations/symbol.rs
PgBiel: maybe "variants with the same modifiers are the same, regardless of their order or amount"?

crates/typst-library/src/foundations/symbol.rs
MDLC01: In the happy case (i.e., no duplicate variant or 128-bit hash conflict), each variant now takes $O(m \log m)$ to process (where $m$ is its number of modifiers), for a total complexity of $O(n m \log m)$ where $n$ is the number of variants and $m$ the number of modifiers per variant. This comes at the cost of an allocation at each iteration of the loop.

laurmaedje:
I simplified it a bit further:
- We can reuse one allocation without running into lifetime issues by splitting the validation loop from the collection loop.
- We can use a HashMap instead of a Vec for checking hashes.
- We don't need the `same_modifiers` check. We can safely ignore 128-bit hash collisions, we do that across the whole compiler. It's much more probable that all of humanitiy dies tomorrow than that we get such a collision.

laurmaedje:
Thanks!

PR #5589, opened 2024-12-16
Get numbering of page counter from style chain

EpicEricEE:
Do what is done for other counters, so we don't always have to write
```typ
#set page(footer: context {
  let page-numbering = if page.numbering != none { page.numbering } else { "1" } 
  counter(page).display(page-numbering)
})
```
to get the correct page numbering, but only
```typ
#set page(footer: context counter(page).display())
```

Note that this also falls back to "1.1" instead of something more suitable, such as "1/1" (though that's also what happens now). Ideally, we would also have something like `both: auto`, which checks how many pieces are in the pattern (as is [done internally](https://github.com/typst/typst/blob/0228462ba10397468b3c0937e2e9cd1d3118e0fc/crates/typst-layout/src/pages/run.rs#L146) for the default page number display), but that's something for another time.


laurmaedje:
Yeah, this one was missing. Overall, it's all kind of a hack, but the hack might as well work consistently across the standard library. Thanks!

Would you mind adding a test to ensure it keeps working?

laurmaedje:
Thanks!

PR #5594, opened 2024-12-17
HTML tests

01mf02:
This PR serves to add HTML support to Typst's test infrastructure.

As discussed with @laurmaedje, test names can now be followed by a number of attributes, currently `html` (produce HTML output), `render` (produce PNG output), and `large` (allow for large outputs). The default attributes are just `render`.

It is planned that files matching `tests/ref/*.png` will move to `tests/ref/render/*.png`. However, in order to prevent merge conflicts, I propose to do this move only right before merging this PR.


laurmaedje:
Thanks!

I decided to make the `ref` move in a separate PR to keep this one tidy and manageable. (GitHub tends to crap out when the diff becomes huge.)

One thing that's left as future work here is to also be able to annotate errors during `typst_html::html` encoding instead of panicking, but that's okay.

Issue #3379, opened 2024-02-07
Confounding math font issues


Enivex:
### Description

I'm not exactly sure how to describe this issue (there may be more than one)

This does not work:
```
#let scr(it) = {
  show math.equation: set text(font: "New Computer Modern Math Book", stylistic-set: 1)
  $cal(it)$
}

$ cal(A) cal(B) $
$ scr(A) scr(B) $
```
![image](https://github.com/typst/typst/assets/2191754/a8759ade-5ac1-4982-9fee-c12ba450873e)

This does work, but the spacing is incorrect:
```
#let scr(it) = {
  set text(font: "New Computer Modern Math Book", stylistic-set: 1)
  $cal(it)$
}

$ cal(A) cal(B) $
$ scr(A) scr(B) $
```
![image](https://github.com/typst/typst/assets/2191754/8b84d962-d9e7-40a4-be64-9af38974e05d)

**The following workaround works on 0.10, but does not work on main:**
```
#let scr(it) = {
  math.class(
    "normal",
    text(font: "New Computer Modern Math Book", stylistic-set: 1, $cal(it)$),
  )
}

$ cal(A) cal(B) $
$ scr(A) scr(B) $
```

Output in 0.10:

![image](https://github.com/typst/typst/assets/2191754/70652567-9574-48be-85cd-f816973ab94e)

Output on main: Same as the picture above with wrong spacing

Lastly, and perhaps most weirdly (and really a different issue)
**None** of the above work with "New Computer Modern Math" instead of "New Computer Modern Math Book" (they all display as the top picture). This might lead one to believe that it's the font that is at fault (missing stylistic set for instance). However, this is not the case, because

```#show math.equation: set text(font: "New Computer Modern Math", stylistic-set: 1)

$ cal(A) cal(B) $
```

works as intended 

![image](https://github.com/typst/typst/assets/2191754/1ea68ded-e782-4e0a-824b-1b638acb7e00)

### Operating system

Windows

### Typst version

- [X] I am using the latest version of Typst

Enivex:
Relevant Discord thread: https://discord.com/channels/1054443721975922748/1204873949431734293

laurmaedje:
I think it all comes down to the fact that a font change in math triggers Typst to leave math layout (because math layout assumes just a single font). Stylistic sets only works in the normal text layout, but since it's treated as normal content, spaces around it are then retained. There is a specific if in the code that checks if the font changed. It is pretty hacky.

[...]

Issue #1193, opened 2023-05-13
Changing font in math operators


OvidiusCicero:
### Description

It should be possible to change the font (or any text property of math operators by for example

```
#show math.op: set text(font: "Allura")
$
lim_(x→∞)f(x)
$
```

but currently this does not have an effect

### Use Case

I want to show math operators in curly script.

PR #5596, opened 2024-12-17
Fix math size resolving

### LINKED ISSUE #3729: Incorrect length measure in math superscript
### Description

The following code:
```typst
#let length = context repr(measure("--").width)
$ a length a ^ length $
```
renders as:
image

while the length provided in superscript should arguably be smaller. This is also happens with e.g. nested fractions, where content is also shrunk visually, but the measured length remains unchanged. Am I correct to believe the lengths should differ?

This can cause packages which rely on precise measurement to display things incorrectly in these contexts, like xarrow (which has not been too problematic so far because arrows are rarely used in superscript, nested fractions, etc.)

### Reproduction URL

_No response_

### Operating system

Web app, Linux

### Typst version

mkorje:
Fixes #3729.

The essence of this solution are the changes made in [crates/typst-library/src/text/mod.rs](https://github.com/typst/typst/compare/main...mkorje:typst:fix-math-sizing-resolving?expand=1#diff-0adb942019d59d0f74ea4230b8e4f398efeba0b944786b42f30edfeeaa141943) in this PR: When resolving `TextSize`, we take into account the current `MathSize` on the style chain. Everything else I've done follows on from that.

Relying less on MathContext in a number of places (as we needed it for the font for `scaled_font_size` mainly) makes some decent progress already towards solving #1193 and #3379, I think.

I've drafted this as I still need to add some more tests. I'll also probably still find some more places where things can be updated.

## Implementation stuff

The default value for the scriptPercentScaleDown and scriptScriptPercentScaleDown constants I've considered from 3 places:
- [MathML Core](https://www.w3.org/TR/mathml-core/#layout-constants-mathconstants), where they have the oddly specific values of 0.71 and 0.5041.
- [OpenType Spec](https://learn.microsoft.com/en-us/typography/opentype/spec/math#mathconstants-table), where the suggested values are 0.8 and 0.6.
- [Building OpenType math fonts](https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#scriptpercentscaledown-scriptscriptpercentscaledown), which says the spec recommendations are bad, and that LaTeX's default setup basically uses 0.7 and 0.5.

As `f64` doesn't have Hash implemented, I'm using i16 (as it is stored in the font itself) and I've settled for now on 0.7 and 0.5


mkorje:
The only issue I've come across so far with these changes is this totally not confusing behaviour:

```typ
#let size = context [#text.size.to-absolute() #(1em).to-absolute()]
$ a^size $
```

playground

It appears Resolve is never called on the value retrieved from some element within context. I expected it to be resolved, matching what happens on the Rust side of things. But I guess that behaviour is understandable as you want to retrieve what you actually set a value to, not what it resolves to.

laurmaedje:
The main reason `Resolve` is not called atm is that it's inconsistently used across the codebase and sometimes things are resolved manually. I didn't want to ship something inconsistent.

mkorje:
I see. This PR makes Resolve be used consistently across the math part of the codebase, which is a step in the right direction I guess...

If it were consistent, is the intention then that when accessing fields they would always be resolved? For example would `#context math.mat.column-gap` return `5pt` instead of `0% + 0.5em`? It also seems it already kinda sometimes gets resolved:
```typ
#let size = context text.size
#size #text(1.5em)[#size]
```
Displays `10pt 15pt`.

(Note: Using the defaults from the test suite)

laurmaedje:
> If it were consistent, is the intention then that when accessing fields they would always be resolved?

Probably yes, but I'm not fully decided yet.

crates/typst-layout/src/math/mod.rs
```diff
@@ -58,12 +57,20 @@ pub fn layout_equation_inline(
 
     let mut locator = locator.split();
     let mut ctx = MathContext::new(engine, &mut locator, styles, region, &font);
+
+    let local = [
+        EquationElem::set_script_scale(ctx.constants.script_percent_scale_down()),
+        EquationElem::set_sscript_scale(ctx.constants.script_script_percent_scale_down()),
```
laurmaedje, line 63: this could even be a single pair entry for a bit less runtime cost

crates/typst-layout/src/math/mod.rs
```diff
@@ -110,9 +116,17 @@ pub fn layout_equation_block(
 
     let mut locator = locator.split();
     let mut ctx = MathContext::new(engine, &mut locator, styles, regions.base(), &font);
+
+    let local = [
+        EquationElem::set_script_scale(ctx.constants.script_percent_scale_down()),
+        EquationElem::set_sscript_scale(ctx.constants.script_script_percent_scale_down()),
+    ]
+    .map(|p| p.wrap());
```
laurmaedje, line 124: could we deduplicate that with the one in layout_equation_inline?

crates/typst-library/src/math/attach.rs
```diff
@@ -152,5 +152,7 @@ pub struct StretchElem {
 
     /// The size to stretch to, relative to the maximum size of the glyph and
     /// its attachments.
-    pub size: Smart<Rel<Length>>,
+    #[resolve]
+    #[default(Rel::one())]
+    pub size: Rel<Length>,
```
laurmaedje, line 157: Technically a breaking change, but an okay one

laurmaedje (approved):
Reading the diff of this made me happy. :)

crates/typst-library/src/math/attach.rs
mkorje: Yes, I did this in a couple places. It seems quite obvious now that the those don't need to be smart.

crates/typst-layout/src/math/mod.rs
mkorje: Will do

crates/typst-layout/src/math/mod.rs
mkorje: That seems reasonable. I will modify it.

mkorje:
I was going to maybe try fix nested math content having its styles overwritten by the inner equation, as in the `math-size-math-content-1` and `math-size-math-content-2` tests. Though I probably won't get to that for a bit, so I'm happy to leave that for another PR so this can go through.

tests/ref/math-size-math-content-3.png
laurmaedje, line 1: this image is a bit confusing

laurmaedje: I assume the image is the current bad case and once it's fixed, it'll make sense? Still, it's kinda confusing now and maybe it should be adjusted.

mkorje: Yeah that's what I was getting at. Though even once its fixed, its still "wrong".

I wanted to add a case for a single expr in math as since it doesnt get wrapped in math, it won't be fixed when the other ones (tests -1 and -2) are fixed.

crates/typst-layout/src/math/run.rs
```diff
@@ -181,28 +182,23 @@ impl MathRun {
...
         let rows: Vec<_> = self.rows();
         let row_count = rows.len();
         let alignments = alignments(&rows);
 
         let leading = if EquationElem::size_in(styles) >= MathSize::Text {
             ParElem::leading_in(styles)
         } else {
-            let font_size = scaled_font_size(ctx, styles);
-            TIGHT_LEADING.at(font_size)
+            TIGHT_LEADING.at(TextElem::size_in(styles))
```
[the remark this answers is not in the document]
mkorje: Knew I'd end up missing something

tests/ref/math-size-math-content-3.png
laurmaedje: hmmm. I still think the image should indicate whether the test passes or not.

mkorje: I could just make it then `sum != height(sum) instead, hopefully that's clearer.

laurmaedje:
Thank you!

Issue #1595, opened 2023-06-27
Custom numbering (Cyrillic)


Andrew15-5:
### Description

I thought, that adding `#set enum(numbering: "а)" | "А)" | "а." |"А.")` numbering patterns would be perfect. But it's only for Cyrillic (which is my main issue). Maybe adding custom numbering — string (where each char is indexed enum label) or list of strings — would be a more flexible and universal solution.

Note: in Cyrillic enum list, next letters are de facto excluded: ё, й, ъ, ы, ь ([source](https://ru.wikipedia.org/wiki/Список_(типографика)#Нумерованные_списки)).

### Use Case

Everyone would be able to easily create new numbering styles (emoji, Cyrillic, Japanese etc.).

Enivex:
You can already provide an arbitrary function to that argument

Andrew15-5:
Nice. Thank you!

Here is my MWE:

```typst
#let ru_alph(pattern: "а)") = {
  let alphabet = "абвгдежзиклмнопрстуфхцчшщэюя".split("")
  let f(i) = {
    let letter = alphabet.at(i)
    let str = ""
    for char in pattern {
      if char == "а" {
        str += letter
      }
      else if char == "А" {
        str += upper(letter)
      }
      else {
        str += char
      }
    }
    str
  }
  f
}
#set enum(numbering: ru_alph()) // or
#set enum(numbering: ru_alph(pattern: "А."))
```

MDLC01:
I think the original issue is still relevant, though. There should be a cyrillic numbering kind.

Andrew15-5:
I mean, for convenience — yes. But is it the only one which should be added — probably not. Although I don't know if every language uses its own alphabet/syllabary in enumerated lists.

syroezhkin:
The search for a solution to the same problem brought me here. If numbering patterns should support several alphabets (except English, Japanese, etc.), then it would be great for me to see support for Cyrillic alphabet out of the box as well.
Otherwise, we are back to the fundamental problems of LaTeX.

Anyway, Typst is great!

Andrew15-5:
@syroezhkin, I think I haven't opened an issue asking for all numbering styles (not only Russian ones) because I have found this package:
https://github.com/typst/packages/tree/main/packages/preview/numberingx/0.0.1
https://github.com/edhebi/numberingx

[...]

PR #5597, opened 2024-12-17
Cyrillic numbering

samuelireson:
Test cases still needed, and I feel all numbering could be tied up here in line with https://www.w3.org/TR/predefined-counter-styles/#cyrillic-styles as mentioned in https://github.com/typst/typst/issues/1595.

Since all the languages which use the cyrillic alphabet use the same first character (U430) for their numbering system, there needs to be some way to distinguish between the language. I'm not sure the best way to proceed with this - there was a suggestion to use the language which was set in the file; I guess this would be pretty neat.

### DIFF
--- a/crates/typst-library/src/model/numbering.rs
+++ b/crates/typst-library/src/model/numbering.rs
@@ -316,6 +316,10 @@ pub enum NumberingKind {
     CircledNumber,
     /// Double-circled numbers (⓵, ⓶, ⓷, etc.), up to 10.
     DoubleCircledNumber,
+    /// Lowercase Russian letters (а, б, в, etc.)
+    LowerRussian,
+    /// Uppercase Russian letters (А, Б, В, etc.)
+    UpperRussian,
 }
 
 impl NumberingKind {
@@ -346,6 +350,8 @@ impl NumberingKind {
             '\u{0995}' => NumberingKind::BengaliLetter,
             '①' => NumberingKind::CircledNumber,
             '⓵' => NumberingKind::DoubleCircledNumber,
+            'а' => NumberingKind::LowerRussian,
+            'А' => NumberingKind::UpperRussian,
             _ => return None,
         })
     }
@@ -377,6 +383,8 @@ impl NumberingKind {
             Self::BengaliLetter => '\u{0995}',
             Self::CircledNumber => '①',
             Self::DoubleCircledNumber => '⓵',
+            Self::LowerRussian => 'а',
+            Self::UpperRussian => 'А',
         }
     }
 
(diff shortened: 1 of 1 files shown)


EpicEricEE:
Don't forget to add a test in `tests/suite/model/numbering.typ` and to update the counting symbols in the documentation here:
https://github.com/typst/typst/blob/73253d465192454f0dfe3fe9eef46d495b343aef/crates/typst-library/src/model/numbering.rs#L62-L65

PgBiel:
> Since all the languages which use the cyrillic alphabet use the same first character (U430) for their numbering system, there needs to be some way to distinguish between the language.

This could potentially be a significant problem, especially considering the current circumstances of the world. In particular, this may be quite uncomfortable for Ukrainian users, as well as speakers of other languages using the Cyrillic alphabet.

This is also related (but evidently much worse) to how all the "A"-like letters (latin A, greek Alpha, cyrillic A) look the same, despite being distinct characters. Even though this was already kind of accepted with the Greek alphabet, this particular case will require a much more careful decision.

FWIW I still think that some alternative format such as `{Upper}` or `{lower}`, e.g. `{Russian/russian}`, `{Ukrainian/ukrainian}`..., could be ideal.

Considering that the [`numberingx` package](https://typst.app/universe/package/numberingx) implements precisely this, I'd be in favor of avoiding adding Cyrillic alphabet numbering until a more robust decision is made, one that makes it possible to unambiguously specify your desired alphabet.

samuelireson:
I don't think a good solution/implementation is out of reach although you raise some good points which I didn't fully consider.

Perhaps giving a warning if one of the 'multi-valued' characters is used as the key for numbering without a language having been set in the document. Then for each of the 38 languages with translations in typst currently, we can provide, or point to the relevant alphabet.

Do you think this is a robust enough solution or are you looking for something more?

laurmaedje:
> Perhaps giving a warning if one of the 'multi-valued' characters is used as the key for numbering without a language having been set in the document.

If we emit a warning, we still need to pick a default.

> FWIW I still think that some alternative format such as {Upper} or {lower}, e.g. {Russian/russian}, {Ukrainian/ukrainian}..., could be ideal.

I'm on board with adding an additional more verbose format, see my comment here for the concrete design I'm envisioning: https://github.com/typst/typst/issues/1177#issuecomment-2451572160

I think we should, as a **first step,** implement these more verbose formats for the existing numbering formats. Once that works
- we can add implement the ambigious cases by providing solely the verbose format, not the short one
- explore whether contextually language-aware numberings would work out well

I'd like to take one step at a time and, for the reasons discussed above, consider the first part a blocker for adding Cyrillic, so I'll close this for the time being. Thanks still!

samuelireson:
Okay - that makes sense. I will have a go at implementing what you mentioned. Did I go about this process in the wrong way? Would it have been better if I started a new issue to discuss this before opening a PR?

laurmaedje:
For smaller stuff a PR is typically a fine. For larger stuff, having some discussion on Discord or here first is better. This one wasn't really large, it just happened to be more intricate than it looks at first glance, so it's all good.

PR #5598, opened 2024-12-18
Bump `unicode-bidi`

### LINKED ISSUE #5490: Panic on syntax highlighted raw in RTL text
### Description

When syntax highlighted code (LTR), within RTL text, crosses into another line.

Compiler panics with error:
```
thread '<unnamed>' panicked at crates/typst/src/layout/inline/shaping.rs:412:30:
byte index 1 is not a char boundary; it is inside '\u{202a}' (bytes 0..3) of `‪elif‬`
```

**Git bisected it to 0ef672c347f368325313c8bccc4f70e3f1016b0a (#4497).**

### Reproduction URL

https://typst.app/project/rVcMHhZIBR4OCejDsxtzeb

### Operating system

Web app, Linux

### Typst version

wznmickey:
Update dependency to fix #5490


laurmaedje:
The test didn't actually test the regression because it only panicked in Typst's default text & paper size, not in the test runner's. I fixed that.

laurmaedje:
Thanks!

PR #985, opened 2023-04-25
Treat function-looking code in attachment as function

SUPERCILEX:
Fixes #828

PR #5605, opened 2024-12-19
Attach functions and characters differently

shivangp76:
Fixes https://github.com/typst/typst/discussions/2086#discussioncomment-7790745 while preserving the changes from #985.

In other words, `f_abs(x)` will be treated as `f_(abs(x))` while `f_n(x)` will be treated like `f_n (x)` (Note how `f_1(x)` is already treated like `f_1 (x)`).


laurmaedje:
We're aware that it's very annoying currently and also that it's inconsistent. There's a whole bunch of discussion about changing it in the [Math](https://discord.com/channels/1054443721975922748/1176478139757629563) forge on Discord where various options are being explored.

While one might consider the change here an improvement in itself, I don't think it's the final solution because `f_n(x)` and `f_pi(x)` will be handled inconsistently then. I only want to change this once more, for good, and the final decision isn't yet made. For this reason, I'll close the PR. Thank you still and I invite you to join the discussion!

PR #1614, opened 2023-06-29
Better primes in math

sitandr:
- Syntax works for tight primes, closes #473
- Primes are automatically moved into attachments, closes #439

Provides "ergonomic interface" for writing primes as mentioned in #1580 (thanks a lot to @damaxwell).

Tight primes are produced by manually layouting prime in special MathLayout element, so that allows producing as many tight primes as needed.

UPD: this element now produces unicode primes for 1, 2, 3 and 4 primes, so `a^'` (or any other prime symbol) and `a'` produce the same result.

### Problems

Sadly, I wasn't able to fully understand the parser logic, so there may be some flaws. However, I tested it on lots of examples, it works fine. Currently (unlike LaTex), it is "greedy", so the order of `subscript`/`prime` matters:

`a'_b` will produce expected result:
image

While `a_b'` will produce a subscript with a prime:
image

I think that is okay, considering typst rules for, e.g. `f_i(x)` parsing as `f_(i(x))`.

UPD: during discussion it was marked us undesired behaviour, so I **fixed** that. Now `a'_b` and `a_b'` both produce the first result.

UPD2: the fiest approach has certain benefits, so now it is reverted back

### Tests

I put the tests to the `math/opticalsize`, since there are already too much things in `attachments`, and that PR is in general just about better syntax for #1580.

PR #5609, opened 2024-12-19
Remove unused single quote shorthand

wrzian:
This was left over from #1614 when we moved primes to be their own element, they weren't removed from the shorthand table in the ast.


laurmaedje:
Thanks!

PR #5613, opened 2024-12-20
Fix arbitrarily nested equations in the base of `math.attach`

mkorje:


laurmaedje:
Still kind of a hack because it's sort of a very shallow type of realization and there are still lots of things a proper realization would do (e.g. apply show rules and styles), which this doesn't. But it's okay for now. Thanks!

mkorje:
Ideally I think this kind of thing should be integrated into realization somehow. Though it seems it isn't able to descend into fields on an element (at least, I can't figure out how to get it to do that).

laurmaedje:
It's definitely not a simple problem to fix. The way realization works in Typst is inherently shallow.

PR #5616, opened 2024-12-20
Added precise definition for "character" in the docs for `str.split`

Andrew15-5:
An additional improvement over #5572.


laurmaedje:
Thanks!

PR #5618, opened 2024-12-21
Add `title` element

jbirnick:
Adds a new `title` element which should be used to display the main document title. Will show as `h1` in HTML.

Notes/Questions/Todos:
1. Should this be bookmarked in PDF?
2. Should it implement `LocalName`? `HeadingElem` does so, but I don't know if it's necessary for `TitleElem`.
3. (official) Templates must be adjusted to use this. (can do after this here is "approved")
4. Should it perhaps be included in some examples in the docs (outside of `title` docs) ? Perhaps even in the `heading` docs?
5. Warnings/Errors
   - I think we should warn if it gets used twice or more. Perhaps we should even error?! I just don't know where the best place to check this is. Could also error for HTML and warn for Paged or so.
   - I think maybe we should also warn if it doesn't get used at all. At least this will push existing templates into using `title`, otherwise we could end up with many templates and template maintainers who have never heard about `title`.


laurmaedje:
Apart from your notes/questions, there are also some open design questions around the interaction of the `title` element and `set document(title: ..)`, so let's push this back to after Typst 0.13.

jbirnick:
I think `title` and `set document(title: ...)` should stay pretty much unrelated. One is about displaying something and the other is about metadata. I often set them to different things, like `title("Homework 1")` vs `set document(title: "Course XYZ, 2025, Homework 1 Submission")`. Also note that in HTML, `<title>` and `<h1>` are different things.

You can think about automatically setting `set document(title: ...)` when it's not set and `title` is set. But I wouldn't do it the other way around.

laurmaedje:
I think doing it the opposite way around would be more natural: if `title` isn't given a body, it would default to `document.title`.

luewolf:
What is the difference between the proposed `#title` and `#heading(level:1)`?

laurmaedje:
A level 1 heading is intended to be used as a section heading. There can be multiple of those. 

The title is for the whole document and there should only be one. It's one level higher up.

In HTML, `title` will become `<h1>` and `heading(level: 1)` will become `<h2>`.

reknih:
In my opinion, we are trying to centralize metadata in `set document`. The `title` element should merely provide a rendering of the metadata. If the user or a template decides that they want to render additional information in the title, they can do this:

```typ
#set document(title: "Homework 1")
#title[
  Course #course, Winter #datetime.today().year() \
  #context document.title
  Submission
]
```

Hence, IMO, the body argument of title should be optional and accept content or `auto`. If it is not provided or `auto`, the function should render the document title. If there is none, it should trigger an error. Templates should put a `#set document` rule in the template file (the one that gets copied into the project). The PDF/HTML would of course expose the full custom body to AT if content was passed.

(This also matches with the guidance in PDF/UA-2, Section 8.3.4.13)
> As titles can vary depending on exact usage there is no intent to require a match between real content tagged with Title and `dc:title` information in XMP metadata.

---

Having more than one title:

In PDF files, this can be legitimate. For example, the PDF 2.0 spec says this on the matter in Table 366:

> [The **Title** structure element] should occur only once inside the parent grouping structure element, and it should occur at or near the beginning of the content inside that grouping structure element.

Grouping structure elements include Part / Div, Art, Sect, Aside, and NonStruct.

Further clarification is provided in Example 3 in Table 366:

> The title of a book, brochure or leaflet are some types of content that can be marked with the Title structure element.

PDF/UA-2 Section 8.3.4.4 requires multiple titles if the file consists of self-contained independent articles, for example in a conference proceedings volume:

> Where a title applies to an entire article, the Title element shall be included inside the Art structure element [...].

Typesetting conference proceedings volumes is a valid use-case for Typst, so using more than one Title should be allowed. The rule that we could enforce is this: For every given title, the closest ancestor that is either a grouping element, a `Document` or a `DocumentFragment`, there must be no other title descendant.

In HTML, the story is a bit different:

The HTML outline algorithm, that allowed elements like `main` and `article` to influence document outlines, [has failed](https://github.com/whatwg/html/pull/7829) and was never implemented. This means that reading `h1`-`h6` elements should be sufficient to build up an outline of the document. From both an SEO and an accessibility perspective, it is discouraged to have more than one `h1` element in a particular HTML document. It would be fine to error if the user emits an HTML document with more than one title.

A caveat is that we are planning to add an API that allows emitting multiple HTML files from a single Typst file. In such a case, each HTML file should be allowed to emit an `h1` element.

_Implementing these diagnostics does not block this PR._

---

The heading docs should contain a reference to the title function and tell when to use what! The tutorial also needs to be adjusted (Making a template section)

---

Every title should be outlined except for:

- The topmost title
- Titles in `DocumentFragments`

---

Don't think it needs a `LocalName`.

laurmaedje:
@jbirnick Let me know whether you still want to continue working on this. Otherwise, I would pick this up in the coming weeks and proceed with a design that pulls the title from `document.title` by default.

jbirnick:
Yeah I don't have much time right now, so feel free to do/extend it yourself.

PR #5619, opened 2024-12-21
Basic HTML `heading` test

jbirnick:
Adds a basic HTML test for heading, as I assume this behavior is finalized.


laurmaedje:
Thanks!

PR #5622, opened 2024-12-23
Numbering

samuelireson:
An implementation of the changes discussed in https://github.com/typst/typst/issues/1177#issuecomment-2451572160.

I also decided to try to centralise the numbering styles for Roman, Hebrew, Greek etc. by introducing a function called `additive` which assigns symbols numeric values and returns the symbolic representation. This is function is described [here](https://www.w3.org/TR/css-counter-styles-3/#additive-system). ~This can also replace the current Chinese numeral implementation - TODO.~ - the algorithm for representing Chinese numerals [cannot](https://www.w3.org/TR/css-counter-styles-3/#limited-chinese) use the `numeric` function

Alongside this, I have introduced functions for each of the CSS counter algorithms described on the page above. I think this is better than the current approach since it means we only need to test the output of these functions for a subset of the NumberingKinds, rather than test every NumberingKind which is the current approach.

TODO
- [x] Implement more numbering kinds - this is a discussion point. How many should be supported? The list of styles [here](https://www.w3.org/TR/predefined-counter-styles/) is huge, but I suppose exhaustive.
- [x] Test cases.
- [x] Naming convention - I have followed the CSS naming convention as seen in the previously referenced web pages, but there was some discussion in https://github.com/typst/typst/pull/5597 about using an `{Upper}/{lower}` convention instead.
- [x] Re-implement the numbering patterns which I removed (Bengali numbers etc.).
- [x] Choose 'shorter' names for the NumberingKinds - so each of the kinds which support "1" as a pattern will also support "{1}" etc. - just add all the current ones; others can be added if wanted later.

I thought it was a good idea to open this PR to discuss these things, but if this is wrong (sorry) - just let me know!


Myriad-Dreamin:
The popular [numbly](https://typst.app/universe/package/numbly/) package implemented similar syntax. They can also use levels to generate numbering string. Not sure whether we can get any inspiration from it.

samuelireson:
From what I can tell the numbly package tries to solve a slightly different problem - heading numbering per level.

The initial discussion which led to this PR was related to the potential to support Cyrillic numbering and the subsequent need to support verbose naming for numbering, since there are multiple Cyrillic numbering styles (Belorussian, Bulgarian, Macedonian, etc.) but these all use the same first letter (/u{430}). At some point, the [numberingx](https://typst.app/universe/package/numberingx/) was also mentioned as a means of solving this problem.

samuelireson:
Maybe this is slightly overkill, but I have now implemented all the numbering styles shown [here](https://www.w3.org/TR/predefined-counter-styles). This is still failing tests since we now return the name of the `NumberingKind` rather than its representative character and the visual outputs will change on this account. I didn't know whether just running `testit --update` was a viable solution here or there was more of a procedure for changing the expected output.

laurmaedje:
I'm sorry for the radio silence here. I was quite busy over the past months, and the large nature of this PR made it hard to fit in between. To get this landed, I think it'd still be good to break this up further. This makes it much easier to review, which also helps speed things up on my end. As I see it, there are a few different changes bundled up here:

1. Refactors/Streamlining of the implementations of the numbering patterns.
2. Changes to the names of existing numbering patterns to align them with CSS. This is not breaking as the names are not currently exposed.
3. Support for `{` in numbering patterns.

I think 1 would make for a good standalone PRs and is independent of the others. 

2 is relevant to 3 as 3 will expose the names. Following CSS probably makes sense, but maybe this warrants some further discussion. For example, I find it a bit odd that CSS spells it "greek-lower-modern", but "lower-roman" (lower once before and once after the writing system).

3 would then follow after 2.

Split up this way, we can land things incrementally.

samuelireson:
Makes sense - followed my nose too far on this one! Should we close this pr and I will reopen for each stage as I complete?

laurmaedje:
Yeah that sounds good!

PR #5627, opened 2024-12-25
Support comparison functions in `array.sorted`

### LINKED ISSUE #5365: Add `compare` argument to `array.sorted`, as an alternative to `key`

MDLC01:
Closes #5365.

This PR is based on a rebased version of #5371 by @bluebear94. I used `glidesort` instead of the standard library sorting algorithm to prevent panics, which was the blocker for merging #5371.

Original post below.

<hr>

This PR is based on a rebased version of #5371 by @bluebear94. As suggested in https://github.com/typst/typst/pull/5371#issuecomment-2539903091, I implemented a custom sorting function that does not panic, which was the blocker for merging #5371.

## Sorting algorithm

The sorting algorithm I implemented is an in-place stable insertion sort. In particular, it does not allocate anything. The drawback is that it is considerably slower than Rust's default sorting algorithm. I initially tried to implement an in-place stable merge sort,[^1] but was unable to reach a point where I was confident about my implementation.

### Performance

I did a very basic benchmark that involved executing the following function:

```typ
#let test(len) = {
  let k = 0b01010101
  for m in range(256) {
    range(len).map(i => i.bit-xor(k).bit-and(m)).sorted()
  }
}
```

The total execution times of the for loop on my machine, as determined by `--timings`, are shown in the table below.

| `len` | 0.12.0 | This PR |
|-------|--------|---------|
| 1000  | 2.11 s | 3.10 s  |
| 10000 | 15.0 s | 114 s   |

The difference seems quite significant for big arrays. The three solutions I can think of are:

1. Deciding not to care (at the cost of performance).
2. Only using the homemade sorting algorithm when a comparison function is provided (at the cost of performance when a comparison function is provided).
3. Using a better sorting algorithm, such as an in-place stable merge sort (which may be harder to implement in a way that we can be confident is correct).
4. Relying on a crate such as [glidesort](https://crates.io/crates/glidesort) (at the cost of adding a new dependency).

I tend to think option 2. is better.

[^1]: https://thomas.baudel.name/Visualisation/VisuTri/inplacestablesort.html


xkevio:
Could also use the [`glidesort`](https://github.com/orlp/glidesort/tree/master) instead of re-implementing a way worse sort to keep performance in all cases, mayhaps. This is what driftsort, the new Rust sorting algorithm, is based on but without the panicking total order check.

MDLC01:
Yes we can also do that, but this requires adding a new dependency.

laurmaedje:
> [...] but without the panicking total order check.

Is that stated somewhere? I couldn't find anything.

xkevio:
> > [...] but without the panicking total order check.
> 
> Is that stated somewhere? I couldn't find anything.

Not explicitly, but you can search for `panic!` or `total order violation` and you will find nothing.

laurmaedje:
I think I would prefer glidesort over an O(n^2) sort, even if it's an additional dependency.

laurmaedje:
I think `key` and `compare` wouldn't even need to be mutually exclusive. If there's a key, we could just compare the keys with the comparison function. What do you think?

laurmaedje:
From a quick look, looks like that's what the code actually does. Then, just the docs need an update.

MDLC01:
I updated the documentation. There was already a test for the behavior.

crates/typst-library/src/foundations/array.rs
```diff
@@ -841,31 +841,63 @@ impl Array {
...
+                None => Ok(x),
+            };
+            let x = key_of(x)?;
+            let y = key_of(y)?;
+            match &compare {
+                Some(f) => Ok(match f.call(engine, context, [x, y])? {
+                    Value::Int(x) => x.cmp(&0),
+                    x => bail!(
+                        span,
+                        "expected integer from `compare` function; got {}",
```
laurmaedje, line 884:
```suggestion
                        "expected integer from `compare` function, got {}",
```

crates/typst-library/src/foundations/array.rs
```diff
@@ -841,31 +841,63 @@ impl Array {
...
+                        x.repr()
```
laurmaedje, line 885:
```suggestion
                        x.ty()
```
typically, we show the type rather than the value.

could you also add a test for this case?

laurmaedje (commented):
Any thoughts on using `by` instead of `compare` as the name?

MDLC01:
Quoting the original post from #5371:

> I used `compare` for the argument name instead of `by` because `by` could plausibly refer to either the key or the comparison function.

Of course, I can change it to `by` if you prefer.

laurmaedje:
I don't know, I like how `by` reads and never found it confusing in Rust.

MDLC01:
I'll change it in an hour or so

MDLC01:
Should we allow `by` to return a float as well?

laurmaedje:
hmm, when sorting floats it might be useful, but the details around +/- 0 are a bit tricky. if we do, we should first investigate what other languages do.

MDLC01:
In JavaScript, `-0.0 === 0.0` (and `-0.0` is the valid syntax for creating negative zero). I'm not aware of any other language that allows returning floats from comparison functions (notably, Lua expects a boolean, and Python doesn't even have this feature).

laurmaedje:
The Lua case is interesting. To make a stable sort possible with just two return values I guess they need to ensure that the the first element passed to the comparison function is always the leftmost one. But if that is a guarantee that one can give, a boolean seems quite straightforward.

MDLC01:
> But if that is a guarantee that one can give, a boolean seems quite straightforward.

I don't think we can safely assume that if we stick with `glidesort`. At the same time, I agree that returning a boolean seems more user-friendly.

laurmaedje:
I think booleans would be _much_ more convenient than -1, 0, 1, so I'd prefer to investigate a bit more whether that's somehow possible before proceeding.

bluebear94:
We can always convert from booleans to orderings using `(b < a) - (a < b)`, but that requires two calls to the comparison function, so I don’t think that’s a good idea.

MDLC01:
I found a way to rely on `glidesort` while using a boolean comparison function, without calling it twice for each comparison.

The idea is that when a sorting algorithm calls a comparison function, there are only two possible outcomes: the elements are swapped, or they are not. This means out of the three outcomes (`<`, `=` or `>`), two (in practice, either `<` and `=`, or `=` and `>`) yield the same result. Essentially, for each comparison, the sorting algorithm only cares about either `<=` or `>=`. If we know which one for each comparison, we can just decide to pass the two arguments to `by` in whatever order gives us the desired result (as `by` computes `<=`, and therefore swapping its arguments computes `>=`). This is explained more formally in the code.

The drawback is that we need to make an extra allocation. This could be prevented when `by` is not passed, but would likely require duplicating some code. I think it's fine that way (`glidesort` allocates either way).

Additionally, I believe this method effectively makes any sorting algorithm stable. So maybe we can find even better sorting algorithms if we also include unstable sorting algorithms.

crates/typst-library/src/foundations/array.rs
```diff
@@ -820,33 +820,95 @@ impl Array {
...
+        /// If given, uses this function to compare elements in the array.
+        ///
+        /// This function should return a boolean: `{true}` indicates that the
+        /// elements are in order, while `{false}` indicates that they should be
+        /// swapped. To keep the sort stable, if the two elements are equal, the
+        /// function should return `{true}`.
+        ///
+        /// If this function does not order the elements properly (e.g., by
+        /// returning `{false}` for both `{(x, y)}` and `{(y, x)}`), the
+        /// resulting array will be unspecified order.
```
laurmaedje, line 832:
```suggestion
        /// resulting array will be in unspecified order.
```

laurmaedje (commented):
If key or by are specified, then Typst's function call overhead will most likely dominate the allocation cost anyway. But if neither is specified, I think we could skip all the complexity and have a separate code path for that. What do you think?

MDLC01:
I initially did not do that because I feared it would have made the code even more complex. In the end, it's fine, but I'm not sure the performance impact is very noticeable. I may try to do a benchmark later if I have the time, but don't count on it.

laurmaedje:
Thank you!

PR #5635, opened 2024-12-28
Fix French text in example

MDLC01:
This fixes a minor issue in an example: "Hotel" is used as a French word here, and should therefore be spelled as "Hôtel," and the "L" in "L'Europe" should not be capitalized.

Note that "Hotel" in "Lyon City Hotel" is used as an English word, and is therefore spelled properly.


laurmaedje:
Merci !

Issue #5278, opened 2024-10-23
More efficient JPG and PNG handling in PDF export


TS53:
### Description

JPG images can get included without reencoding the image when exporting to PDF.

We can see this with the following latex code and [this](https://polona.pl/item-view/dac0ab7a-43b6-48e5-8492-a38e0b744cde?page=0) file (filesize=8.4 mb):

```latex
\documentclass{article}

\usepackage{graphicx}

\begin{document}
\includegraphics[width=4cm]{map.jpg}
\end{document}
```

The resulting PDF compiles fast and has filesize 8.41 mb.

With typst, it seems the JPG is decoded and then reencoded, which is slower and makes the filesize bigger:

```typst
#image(width: 4cm, "map.jpg")
```

The resulting PDF has filesize 9.29 mb.

Note: for PNG also a [variant](https://tex.stackexchange.com/questions/39929/fast-png-embedding-using-pdflatex) exists that can get copied into the document without reencoding.

### Use Case

When exporting to PDF and having JPG images included:
- smaller binary size 
- faster compilation times
- probably better image quality (original is kept)

TS53:
I think the `RasterImage` is already in a decoded format. If true, then this would mean that the image is at this point already stored into memory which will require some CPU time for decompression and is problematic for big images because it will at least temporarily load the full image into RAM like shown [here](https://github.com/typst/typst/discussions/5279).
Of course this cannot be prevented for all image formats.
From my understanding, JPG can just get copied into the final file without decoding the image and loading its contents into memory.

laurmaedje:
Yes, JPEG could be handled more efficiently and even for PNG sometimes reuse is possible (though we'd more low-level access to the PNG). I had started to work on that a while ago, but didn't finish.

ghost:
When compiled from main branch, it works now. The resulting PDF is only 2 kb bigger.
With PNG it is still slow, but I think this is expected because only few PNG files can be used directly.

Also, for JPG it seems the file is still opened and maybe decompressed which still slows down a bit.
A workaround is to include the JPG in a single PDF and then load this PDF with the single image instead of the image (`#image("pdfwithimg.pdf")`).
This will have the expected fast runtime.

laurmaedje:
The JPG is decoded because an invalid JPG should be a compile error instead of just generating an invalid PDF.

rugk:
Hm is this really fixed? I still see big sizes if I include an `image` in a `grid` with `auto` size? Or should I open a new issue about this?

[...]

Issue #3781, opened 2024-03-24
CMYK images are broken in PNG and PDF exports 


reknih:
When embedding a CMYK JPEG in a Typst document, all sorts of things go wrong.

First, the PNG export and web app preview render the file as if you had let loose a toddler on all sliders in Photoshop:

<img width="485" alt="Oversaturated image of a tiger" src="https://github.com/typst/typst/assets/3874949/c1af5836-6503-4dc5-9762-c43e9bea1db1">

This is because the `image` will convert the `DynamicImage`'s image data to RGB in a naïve way.

But this is nothing against what is going on in the PDF. I converted the `tiger.jpg` from `typst-dev-assets` to CMYK with an ICC profile in Photoshop. When displayed with proper color management, the colors in the file closely resemble the original RGB file. I posted a [repro project](https://typst.app/project/rEeGeKDCjPbCZwdMYdAlIq).

When we write the PDF, we are encountering the same problem as in PNG export: `image` has converted the buffer to RGB. Then, we start writing a color space which is when things go really awry. The image file contains an ICC profile, which we do not read but copy into the PDF. We then say that the profile has three components because we assume this for every color image, even when CMYK ICC profiles have four components. This really flabbergasts readers. Adobe Acrobat refuses to open the file altogether. Other viewers, given the mismatch between the actual profile and its metadata, fall back to the backup profile (sRGB) we wrote (or do not support ICC profiles at all and always fall back to a Device color space, like pdf.js) to show the same horrid color reproduction that we saw in the PNG export.

**Proposed solutions**
- To hotfix the PDF and not write corrupt files, we can read the image's color profile with `qcms` which we depend on anyways and discard it if it does not have the number of components we expect (three or one). This will allow Acrobat to open files made with CMYK images but do nothing about the fact that we convert them to RGB first and screw the colors up in the process.
- To address this problem fully, we should stop relying on `image` and instead support JPEG, PNG, GIF, and TIFF directly with its underlying crates. This will allow us to correctly detect the number of color channels in an image and perform the color management ourselves for PNG export. For PDF export, this will even provide a speedup (and, in the case of lossy compression, more quality) since we can cease to decode and reencode images throughout the compilation if we do not have to resize them.

When embedding a CMYK image in a SVG, both exports also provide bad (but different) results. I will file separately on `svg2pdf`.

Enivex:
Partial duplicate of https://github.com/typst/typst/issues/2158 ?

reknih:
[...]

PR #5637, opened 2024-12-29
Add support for higher bit depths, do not re-encode JPEGs

frozolotl:
Sorry, another PR with multiple different changes.
It is based on #5632 because that PR also changes some image handling things that would slightly conflict with this one.

This PR changes things so that JPEGs are not re-encoded when exporting to PDF.
#5278 requests the same change for PNG, which this PR does not include.

Additionally, this PR allows PDF files to include images with higher bit depths.
This is not very relevant for JPEG, but more so for PNG.

### DIFF
--- a/crates/typst-pdf/src/image.rs
+++ b/crates/typst-pdf/src/image.rs
@@ -1,8 +1,7 @@
 use std::collections::HashMap;
-use std::io::Cursor;
 
 use ecow::eco_format;
-use image::{DynamicImage, GenericImageView, Rgba};
+use image::{DynamicImage, GenericImageView, LumaA, Rgba};
 use pdf_writer::{Chunk, Filter, Finish, Ref};
 use typst_library::diag::{At, SourceResult, StrResult};
 use typst_library::foundations::Smart;
@@ -138,7 +137,12 @@ pub fn deferred_image(
     let interpolate = !pdfa && image.scaling() == Smart::Custom(ImageScaling::Smooth);
 
     let deferred = Deferred::new(move || match image.kind() {
-        ImageKind::Raster(raster) => Ok(encode_raster_image(raster, interpolate)),
+        ImageKind::Raster(raster)
+            if raster.format() == RasterFormat::Exchange(ExchangeFormat::Jpg) =>
+        {
+            Ok(encode_raster_jpeg(raster, interpolate))
+        }
+        ImageKind::Raster(raster) => Ok(encode_raster_flate(raster, interpolate)),
         ImageKind::Svg(svg) => {
             let (chunk, id) = encode_svg(svg, pdfa)
                 .map_err(|err| eco_format!("failed to convert SVG to PDF: {err}"))?;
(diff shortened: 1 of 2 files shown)


Enivex:
Was the existing reencoding lossy or lossless?

frozolotl:
> Was the existing reencoding lossy or lossless?

It was lossy.

Enivex:
I could see (optional) reencoding being useful if it was the lossless kind.

reknih:
As Laurenz already said on the Discord, not all JPEGs are supported by PDF, so we must detect the kind of JPEG and recode it if it is one of the rarer ones.

JPEGs come in multiple kinds: Sequential (aka. Baseline), Progressive, and Lossless. Each of these can support 8 bit (aka. Baseline) or 12 bit color depth (aka. extended). Each of these modes may be encoded with Huffman or Arithmetic encoding. In addition, a JPEG file may contain multiple resolutions of an image, where the higher resolutions are encoded as differences to the next lower resolution (hierarchical mode). Hierarchical mode does not support extended color depth. 

JPEG images are encoded in JFIF or EXIF, both descendants of JIF (Joint Photographic Experts Group Interchange Format). JIF defines a set of Start of Frame (SOF) markers, one for each mode of JPEG. [An overview over all the possible markers can be found on page 11 of CITT Recommendation T.81.](https://www.w3.org/Graphics/JPEG/itu-t81.pdf)

To embed a JPEG in PDF, it is wrapped in a stream object with a `DCTDecode` filter. The [PDF 1.3 spec](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.3.pdf) says the following in Section 3.3.6:

> [...] Certain choices have been made regarding reserved marker codes and other optional features of the standard. For details, see [Adobe Technical Note #5116](https://pdfa.org/norm-refs/5116.DCT_Filter.pdf), Supporting the DCT Filters in PostScript Level 2.

> In addition to the baseline JPEG format, in PDF 1.3 the DCTDecode filter supports the progressive JPEG extension.
> *Note:* There is no benefit to using progressive JPEG for stream data that is embedded in a PDF file. Decoding progressive JPEG is slower and consumes more memory than baseline JPEG. The purpose of this feature is to enable a stream to refer to an external file whose data happens to be already encoded in progressive JPEG.

So this means that hierarchical and lossless JPEGs are not supported by PDF. In the referenced Tech Note 5116, page six tells us this:

> For JPEG experts only: The SOF0, SOF1, DHT, RSTm, EOI, SOS, DQT, DRI, and COM markers are properly decoded. APPn (application-specific) markers are skipped over harmlessly except for the Adobe reserved marker described later.
> These markers are not decoded: SOF2-SOF15, DAC, DNL, DHP, EXP, JPGn, TEM, and RESn. If any occurs in a compressed image, it will be rejected. With the exception of DNL, none of these markers is useful in a Baseline DCT or Extended sequential DCT image.

So this means that only these markers / modes are supported through the tech note:

- SOF0: Sequential Huffman, 8bit (baseline)
- SOF1: Sequential Huffman, 12bit (extended)

If we extrapolate from this and add support for progressive modes, starting from PDF 1.3, the SOF2 marker (Progressive Huffman, 8bit) should also be supported.

That means that we would need to recode any image using arithmetic encoding, lossless, or hierarchical modes to the appropriate Sequential Huffman bit depth by detecting the presence of SOF3 through SOF15.

As an aside: It is possible to losslessly convert a Progressive to a Sequential JPEG, which the PDF 1.3 spec seems to encourage. We could do so in the future, for reference, also see [this discussion](https://www.imagemagick.org/discourse-server/viewtopic.php?t=22141).

frozolotl:
Arithmetic coding is similarly unsupported. only SOF0 (baseline), SOF1 (baseline), and SOF2 (progressive) are.

crates/typst-pdf/src/image.rs
```diff
@@ -32,11 +31,13 @@ pub fn write_images(
                 EncodedImage::Raster {
                     data,
                     filter,
-                    has_color,
+                    color_space,
+                    bits_per_component,
                     width,
                     height,
-                    icc,
+                    compressed_icc,
```
reknih, line 38: This ICC profile could have 1, 3, or 4 components

crates/typst-pdf/src/image.rs
```diff
@@ -45,54 +46,56 @@ pub fn write_images(
...
+                        match color_space {
+                            ColorSpace::Srgb => {
+                                stream.n(3);
+                                stream.alternate().srgb();
+                            }
+                            ColorSpace::D65Gray => {
+                                stream.n(1);
+                                stream.alternate().d65_gray();
+                            }
+                            _ => unimplemented!(),
```
reknih, line 98: Here, we only ever write `\N 1` or `\N 3`, but not `\N 4` even though the original file could be CMYK.

crates/typst-pdf/src/image.rs
```diff
@@ -121,83 +124,184 @@ pub fn deferred_image(
...
+    }
+}
+
+/// Matches an [`image::ColorType`] to [`ColorSpace`].
+fn to_color_space(color: image::ColorType) -> ColorSpace {
+    use image::ColorType::*;
+    match color {
+        L8 | La8 | L16 | La16 => ColorSpace::D65Gray,
+        Rgb8 | Rgba8 | Rgb16 | Rgba16 | Rgb32F | Rgba32F => ColorSpace::Srgb,
+        _ => unimplemented!(),
```
reknih, line 265: Image implicitly (and naively) converts CMYK to RGB and has no variant for it. When the color channel count determined here is written into the PDF, CMYK JPEG streams will have a mismatch with their image stream dictionary.

reknih (commented):
I think that embedding the file directly is also an opportunity to address #3781 for the CMYK JPEG in PDF pairing. In theory, if the right number of channels is selected and any ICC profile present in the file is attached, it should work... However, I had a look at the code and saw that it only handles the Grayscale and RGB cases since that's what's in the image enum.

frozolotl:
I am continuing to work on this, but there will be some extra changes so that CMYK JPEGs are encoded correctly (for both PDF and PNG export, hopefully).

crates/typst-pdf/src/image.rs
```diff
@@ -149,52 +153,148 @@ pub fn deferred_image(
...
+                .flat_map(|px| px.to_rgb().0)
+                .flat_map(|c| c.to_be_bytes())
+                .collect();
+            (16, deflate(&encoded))
+        }
+        DynamicImage::ImageRgb32F(buf) => {
+            let encoded: Vec<u8> =
+                buf.as_raw().iter().flat_map(|&c| c.to_be_bytes()).collect();
+            (32, deflate(&encoded))
+        }
```
LaurenzV, line 223: Have you tried whether this actually works? My understanding is that this somehow stores images as 32-bit floats, right? Are you sure that works just like that in PDF?

frozolotl: Looking at it again, I don't think it can ever be called because no input format gets decoded to rgb32f by the image crate.

frozolotl:
I think I'm closing this in favour of #5420. It seems to do everything this PR does with only interpolate support missing.

PR #5639, opened 2024-12-30
Macro for inverse hyperbolic functions

ph3n92h3:
Add macro for inverse hyperbolic functions.

Reference: https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions


Enivex:
For the time being no more built-in operators are being added.

https://github.com/typst/typst/pull/5577#issuecomment-2545464521

PR #5641, opened 2024-12-31
Support curly-braced numbering pattern

LuxxxLucy:
As per #1177, this PR implements the curly-braced numbering pattern so that certain characters such as 't' can be used in the pattern.

```
#numbering("1", 13)

#numbering("{1}", 13)

#numbering("{arabic}", 13)

#numbering("title {1}", 13)

#numbering("prefix {arabic} suffix", 13)
```
will be rendered as follows
image

To ensure compatibility and ideal fallback result (and minimal change of the current implementation)

now it will first search whether the patterns contains a “{“, if so, it will try parsing the braced version;

if it failed, it would fallback to the current one-character format.

There is concurrent PR #5622 that changes both the brace format (with a slightly different way) and more changes about renaming the numbering kinds. I think we can separate the changes, one with just the implementation for the braced pattern, and then about changing the names/strings/constants.


laurmaedje:
I agree that the syntactical changes and the new numbering patterns should be handled separately. However, since https://github.com/typst/typst/pull/5622 was opened first and contains both, I would like to give that contributor the chance to finish the work they started. Feel free to coordinate something else between you two, but absent that, I'd for now proceed with that PR first.

laurmaedje:
I'll close this, as the other contributor is still actively working on this topic, just spread over multiple PRs, and as we also first need to decide on the [exposed names](https://github.com/typst/typst/pull/6379). Feel free to join the discussion there!

PR #5646, opened 2025-01-01
expanding community platforms

MirS0bhan:
hi, is could be great to acknowledge groups of people are trying to create and expand communities in their state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,9 @@ the community:
 
 If you had a bad experience in our community, please [reach out to us][contact].
 
+### language community 
+- 🇮🇷 fa [telegram](https://t.me/typst_fa)
+
 ## Contributing
 We would love to see contributions from the community. If you experience bugs,
 feel free to open an issue. If you would like to implement a new feature or bug


drupol:
Hello,

Why would you need Telegram when there's already a Forum and a Discord?

Thanks!

MirS0bhan:
@drupol because in iran and persian community telegram is the most popular network that people uses. also that's same as russia. they have [this ](https://t.me/typst_ru)  and just like that i to want say it is good for typst to support and give credits to these communities.

drupol:
I understand now, thank you for explaining. 

I absolutely value the importance of diverse communities and recognize that different platforms resonate more strongly in certain regions or cultures. People are, of course, free to create and participate in **community-led initiatives**. However, there's a distinction between such initiatives and officially endorsing them in the project's README file.

One key concern is **moderation**. When a platform is promoted officially (in the README), users will naturally expect it to align with the common values and principles of the Typst project. Ensuring that this is the case requires consistent and fair moderation, which may not always be feasible for third-party or community-driven platforms. I mention this based on my experience as a member of a well known community, where we are still facing challenges with unmoderated platforms associated with the project. These issues can create confusion, misrepresentation, and a lack of accountability, which are difficult to address once they arise.

Another important consideration is **fragmentation**. The more platforms a project officially promotes, the more challenging it becomes for users to navigate and find the right resources. This could lead to confusion and dilute the focus, making it harder to maintain a cohesive and effective source of documentation and support for everyone.

I understand that this PR is made with the best intentions, and I truly appreciate your enthusiasm and the value that Telegram holds for your community. However, I believe these factors need careful consideration before making such additions to the README, as it might inadvertently harm the Typst project.

laurmaedje:
I think @drupol makes a lot of good points of why we have reservations with adding such links to the project README. Still, we acknowledge the desire to promote and discover community-driven platforms, especially for people that want to discuss Typst in their native language. For this reason, we have created a pinned forum thread, where you can write posts about community initiatives: https://forum.typst.app/t/community-platforms/2318/4

MirS0bhan:
@laurmaedje that means a lot. thank you.

PR #5651, opened 2025-01-04
Allow adding and joining `arguments`

### LINKED ISSUE #5640: Allow adding and joining `arguments`
### Description

Currently, you can add together arrays and dictionaries with `+`. You can also `.join()` and `.sum()` an array containing those. In addition, you can use the spread operator such as with `(..array1, ..array2)` or `(..dict1, ..dict2)`.

However, even though you can spread `arguments` with `arguments(..args1, ..args2)`, you cannot add them together with `+`, nor `.join()` or `.sum()` an array of them. This feels inconsistent, so it would be nice to add those features to `arguments` as well.

### Use Case

The idea is to be able to easily combine different `arguments` without having to use the spread operator, which is useful when you're processing inputs from different function calls by the user.

MDLC01:
Closes #5640.

This PR enables addition and joining for `arguments`. This is also compatible with `array.sum` and `array.join`. In case of a duplicate named argument key, the last one will be kept.

### DIFF
--- a/crates/typst-library/src/foundations/args.rs
+++ b/crates/typst-library/src/foundations/args.rs
@@ -1,4 +1,5 @@
 use std::fmt::{self, Debug, Formatter};
+use std::ops::Add;
 
 use ecow::{eco_format, eco_vec, EcoString, EcoVec};
 use typst_syntax::{Span, Spanned};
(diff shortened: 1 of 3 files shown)


laurmaedje:
Thanks!

Side note: The quadratic complexity is a bit unfortunate, but it's consistent with how all the other argument parsing functions do linear scans and it's probably better in the common case that allocating a set.

PR #5652, opened 2025-01-04
Improve `repr` for `arguments`

MDLC01:
This PR improves the `repr` for `arguments` by making it valid Typst syntax, using a call to the `arguments` constructor.


laurmaedje:
Thanks!

PR #5655, opened 2025-01-04
Add support for `c2sc` OpenType feature in `smallcaps`

MDLC01:
The [`c2sc`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#c2sc) OpenType feature turns uppercase letters into small capitals (as a reminder, the [`smcp`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#smcp) OpenType feature, used in the default show rule for `smallcaps`, turns lowercase letters into small capitals).

I added support for `c2sc` in `smallcaps` by adding two new boolean parameters:
- `uppercase` is `false` by default, and determines whether to turn uppercase letters into small capitals (enabling the `c2sc` OpenType feature in the default show rule);
- `lowercase` is `true` by default, and determines whether to turn lowercase letters into small capitals (enabling the `smcp` OpenType feature in the default show rule).

The default behavior of `smallcaps` is not changed.

Note that there also exist [`pcap`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#tag-pcap) and [`c2pc`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#c2pc) features, but I don't think many fonts use them.


MDLC01:
I just realized it may also make sense to apply the text style changes in a show set rule. Not sure which is better, if it even changes anything given the affected fields are not accessible by the user.

reknih:
What about simplifying the API a bit and just doing `all: boolean` (default `false`) instead of `uppercase` and `lowercase`?

In the vast majority of cases, uppercase implies lowercase. And if the escape hatch is needed, the features can be manually specified via `set text`. Only limitation vs. your proposed API is that if this function should support generating fake smallcaps in the future, generating smallcaps for only the uppercase letters would be impossible.

MDLC01:
> Only limitation vs. your proposed API is that if this function should support generating fake smallcaps in the future, generating smallcaps for only the uppercase letters would be impossible.

Wouldn't this still be possible using a regex show rule?

crates/typst-library/src/text/smallcaps.rs
```diff
@@ -64,10 +64,32 @@ pub struct SmallcapsElem {
...
+        };
+        Ok(self.body.clone().styled(TextElem::set_smallcaps(sc)))
+    }
+}
+
+/// What becomes small capitals.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum SmallcapsSetting {
+    /// No small capitals.
+    Disabled,
```
laurmaedje, line 80: Disabled could be removed in favor of using `Option<..>` and just `Smallcaps` might be enough for the name.

laurmaedje:
Thanks!

PR #5656, opened 2025-01-05
Use proper link elements in bibliography

### LINKED ISSUE #5614: Possibility of colouring links in cite command
### Description

I like to colour all links inside a PDF document, such that it becomes clear to the reader what is clickable without needing them to discover that by trial and error.
Therefore, in my LaTeX documents I use the following setup for hyperref:
```latex
\usepackage{xcolor}
\definecolor{darkred}{rgb}{0.5,0,0}
\definecolor{darkgreen}{rgb}{0,0.5,0}
\definecolor{darkblue}{rgb}{0,0,0.5}
\definecolor{darkyellow}{rgb}{0.5,0.5,0}
\definecolor{darkcyan}{rgb}{0,0.5,0.5}
\definecolor{darkmagenta}{rgb}{0.5,0,0.5}

\usepackage{hyperref}
\hypersetup{colorlinks,breaklinks}
\hypersetup{linkcolor=blue,citecolor=darkmagenta,urlcolor=darkgreen}
```
Together with biblatex I this produces the following output:
grafik

In Typst I tried to use for example
```typst
show link: set text(fill: olive)
```
but this won't color links generated by `@...` commands. Furthermore, as far as I could tell there is no way to override the behavior in `show ref:...` because the cite command links things differently as a normal figure reference.

### Use Case

This would be useful to be able to signal to the reader in a more clear way that something is clickable (especially since there is also no inbuilt rectangle indicator for links).

MDLC01:
Fixes #5614.

This applies the fix suggested in https://github.com/typst/typst/issues/5614#issuecomment-2556797130, making it possible to style links generated by the bibliography element.


laurmaedje:
We should probably rather change `Content::linked` itself so that it won't happen again and other places are also fixed.

laurmaedje:
Checking the code, this means link show rules would now also affect `footnote`, `ref`, and `outline`. I _think_ that probably makes sense, (if users would implement these, they would also have to use link, after all). But I'm not entirely sure and I'd be happy to hear more opinions.

It makes `show link: underline` (which is in the link docs) a bit unusable. Typically, you probably only want to underline URL links. Which is possible in a `show: it => ..` rule, but slightly more involved.

MDLC01:
In a way, this issue already existed, as it has always been possible for the user to create a link with a label as the destination. But I agree this makes it more problematic. Especially because there is no simple way to apply style to URL links only (as you said, it has to be done using a show rule, like in one of the tests).

laurmaedje:
I think we can move forward this. It makes semantic sense and, for whatever it's worth, it's the same in HTML with `<a>`.

However, I think we need to update the link documentation to demonstrate and explain this and show how to style only URL links.

eltos:
> this means link show rules would now also affect footnote, ref, and outline. I think that probably makes sense, (if users would implement these, they would also have to use link, after all). But I'm not entirely sure and I'd be happy to hear more opinions.

@laurmaedje In my opinion, this is problematic. Having URL links styled differently than other links is very common, e.g. these often have monospace fonts or are underlined as you mention. In the original issue #5614 @taaem also explicitly wanted to use different colours for URL links, citations/refs, and other type of links, as it's a basic feature of LaTeX hyperref package. Even if in HTML all will become `<a>`, you probably want to use different classes to get distinct styles. 

> It makes show link: underline (which is in the link docs) a bit unusable. Typically, you probably only want to underline URL links. Which is possible in a show: it => .. rule, but slightly more involved.

Totally agree. If you really want to make link show rules to apply to everything clickable, may I suggest introducing a special URL link type? Then one could use `show link` to affect all links (breaking change), and `show ref`, `show url`, `show cite` etc. to affect only the respective type of link. There could be an explicit `#url(...)`, or it could be inferred from the argument to `#link(...)` (similar as there is `#cite`, but `#ref` also accepts labels in the bibliography).
```typ
// currently this already works fine
#show ref: set text(fill: blue)
#show link: underline // here we could use "url" in the future to avoid everything to get underlined
#show cite: set text(fill: maroon)
```

PS: I don't consider #5614 a bug, actually. One can get the requested behaviour very easily and analogous to hypersetup as shown in https://github.com/typst/typst/issues/5614#issuecomment-2558089497

tests/suite/model/bibliography.typ
```diff
@@ -65,3 +65,16 @@ Now we have multiple bibliographies containing @glacier-melt @keshav2007read
...
+
+--- issue-5614-bibliography-link-styling ---
+#show cite: set text(green)
+#show link: it => {
+  if type(it.dest) == str {
+    underline(it)
+  } else {
+    it
+  }
+}
```
eltos, line 77: It seems that already works in typst 0.12.0. I just copy-pasted it to the webapp and it gives me green citation markers without underline, and underlined links in the bibliography as expected. Maybe I'm not getting it, but what was the bug actually? See also https://github.com/typst/typst/issues/5614#issuecomment-2576152520

laurmaedje:
I misunderstood https://github.com/typst/typst/issues/5614, so there isn't really a pressing bug, just the question of what should be a link and what not. I see your points, though I still think it's unfortunate that the compiler-generated links are inherently different than anything a user could generate. They are all links after all. The fact that we only want to style a subset of them doesn't really change that, though it is inconvenient.

Right now one would need a transformational show rule to filter for links with a string target, but in a future where types and content are merged and selectors become patterns, a rule like this would be feasible:
```typ
#show link.where(dest: str): underline
```
(Here, `where` would be changed such that the RHS of each field is a selector/pattern again. This would also enable things like `heading.where(level: 1 | 2)` where `|` will be syntax for `pattern.or` / `selector.or`.)

eltos:
> Right now one would need a transformational show rule to filter for links with a string target, but in a future where types and content are merged and selectors become patterns, a rule like this would be feasible:

Sounds good, so basically
```typ
#show link // will apply to any link
#show link.where(dest: str) // replaces present #show link
#show link.where(dest: ref) // replaces present #show ref
#show link.where(dest: cite) // replaces present #show cite
#show link.where(dest: footnote | heading | metadata | equation) // and many more options hopefully ;)
```

laurmaedje:
Not quite. The available things would be the types of [`dest`s](https://typst.app/docs/reference/model/link/#parameters-dest) that links allow: str, label, location, and dictionary.

What you are looking for would be more akin to something like nesting selectors (in CSS syntax `show cite > link: ..` though there are no concrete plans for Typst syntax). It would be possible with a nested show rule today, but not with a single selector.

MDLC01:
I'm not sure I'm following everything. Should I just update the underline example in the link documentation?

laurmaedje:
@MDLC01 Let's push this to after Typst 0.13. In principle, I think we should use proper link elements everywhere and I'm not sure blocking on more powerful selectors for easing styling of ULR links is worthwhile, but it's not pressing and I don't want to rush it right now.

laurmaedje:
I've changed my mind. I think we should not change this until there's a better way to write a show rule only for URL or other kinds of links. I'll close this for now. Thanks still!

laurmaedje:
I'm currently reconsidering this because HTML export would also benefit from a fix for https://github.com/typst/typst/issues/5614. I realized that we can have `#show link.where(dest: str): underline` even before custom types and have put up https://github.com/typst/typst/pull/6903, which implements it. If this goes through (I want to wait at least for a bit of feedback), I'd then go ahead and exhaustively apply semantical elements in places that currently don't use them (fixing https://github.com/typst/typst/issues/5614, https://github.com/typst/typst/issues/6288, and https://github.com/typst/typst/issues/6639)

PR #5659, opened 2025-01-06
Avoid stripping url prefixes multiple times or multiple prefixes

bbb651:
```typ
#let l(url) = {
  link(url)
  sym.arrow.r
  url
}

#l("foo")

#l("mailto:foo")

#l("mailto:tel:foo")

#l("mailto:mailto:foo")
```

Before:
image
After:
image

There's no valid `mailto:` or `tel:` that start with `mailto:` or `tel:`, so such links are always invalid as in "the url isn't valid", but typst isn't concerned (currently?) about link validation and there might be more prefixes in the future (e.g. `sms`, `geo`), so this seems more sensible.

### DIFF
--- a/crates/typst-library/src/model/link.rs
+++ b/crates/typst-library/src/model/link.rs
@@ -135,10 +135,10 @@ impl Show for Packed<LinkElem> {
 }
 
 fn body_from_url(url: &Url) -> Content {
-    let mut text = url.as_str();
-    for prefix in ["mailto:", "tel:"] {
-        text = text.trim_start_matches(prefix);
-    }
+    let text = ["mailto:", "tel:"]
+        .into_iter()
+        .find_map(|prefix| url.strip_prefix(prefix))
+        .unwrap_or(url);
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/model/link.rs
```diff
@@ -137,7 +137,10 @@ impl Show for Packed<LinkElem> {
 fn body_from_url(url: &Url) -> Content {
     let mut text = url.as_str();
     for prefix in ["mailto:", "tel:"] {
-        text = text.trim_start_matches(prefix);
+        if let Some(stripped) = text.strip_prefix(prefix) {
+            text = stripped;
+            break;
+        }
     }
```
laurmaedje, line 144:
```rust
let text = ["mailto:", "tel:"]
    .into_iter()
    .find_map(|prefix| url.strip_prefix(prefix))
    .unwrap_or(&url);
```
is a bit more compact and should be the same (I think)

laurmaedje:
Thanks!

PR #5666, opened 2025-01-07
HTML tables

01mf02:
This PR implements HTML output for [tables](https://typst.app/docs/reference/model/table).

Like for other primitives, the HTML output ignores styling-related attributes, such as fill, gutter, or align.

TODO:

* [x] Headers/Footers: HTML has the `<thead>`/`<tfoot>` tags to designate headers/footers.
* [ ] Horizontal/Vertical lines

### DIFF
--- a/crates/typst-library/src/html/dom.rs
+++ b/crates/typst-library/src/html/dom.rs
@@ -522,11 +522,6 @@ pub mod tag {
                 | self::p
                 | self::pre
                 | self::search
-                | self::table
-                | self::thead
-                | self::tbody
-                | self::tfoot
-                | self::tr
         )
     }
 
--- a/crates/typst-html/src/encode.rs
+++ b/crates/typst-html/src/encode.rs
@@ -120,7 +120,10 @@ fn write_element(w: &mut Writer, element: &HtmlElement) -> SourceResult<()> {
 
 /// Whether the element should be pretty-printed.
 fn is_pretty(element: &HtmlElement) -> bool {
-    tag::is_block_by_default(element.tag) || matches!(element.tag, tag::meta)
+    matches!(
+        element.tag,
+        tag::meta | tag::table | tag::thead | tag::tbody | tag::tfoot | tag::tr
+    ) || tag::is_block_by_default(element.tag)
 }
 
 /// Escape a character.


01mf02:
To elaborate on the feasibility of horizontal/vertical lines in tables: This could be handled by the `CellGrid::{hlines, vlines}` fields and could be translated to something like `<td style="border-left: 2px solid black;">...</td>`.

However, the stroke type in Typst is much more expressive than the HTML `border-style` attribute, so we risk losing information here. Furthermore, while some stroke properties, such as `stroke(dash: "dotted")`, could be nicely mapped to `border-style: dotted` in principle, this is not that easy in practice, because `stroke.rs` converts "dotted" to `[DashLength::LineWidth, Abs::pt(2.0).into()]`. While this is nice for PDF generation, this representation loses information for HTML. We can of course try to infer from something like `[DashLength::LineWidth, Abs::pt(2.0).into()]` that this was "dotted", but that seems rather brittle and easy to break.
Another issue is that borders that span several cells are interrupted by default at the cell border in HTML, unless something like `table { border-collapse: collapse; }` is used. However, Typst does not set any global CSS properties, meaning that table lines would look ugly default.

Before venturing more deeply into this, I would like to know the priority of hline/vline support for HTML tables. I currently lean towards omitting hline/vline support for HTML tables for now.

tests/ref/html/basic-table.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
+    <table><thead><tr><td>The</td><td>first</td><td>and</td></tr><tr><td>the</td><td>second</td><td>row</td></tr></thead><tbody><tr><td>Foo</td><td rowspan="2">Baz</td><td>Bar</td></tr><tr><td>1</td><td>2</td></tr><tr><td colspan="2">3</td><td>4</td></tr></tbody><tfoot><tr><td>The</td><td>last</td><td>row</td></tr></tfoot></table>
```
reknih, line 8: Tables can be pretty-printed:

According to spec section 4.9.12.1, no. 8, elements that are not `colgroup`, `thead`, `tbody`, `tfoot`, or `tr` as well as the first `caption` are skipped within a table (particularly whitespace text nodes). In row groups (`thead`, `tbody`, `tfoot`) only `tr` elements are processed and in rows, only `td` and `th` are processed. I'd keep the strict mode for `td` and `th`.

crates/typst-library/src/model/table.rs
```diff
@@ -259,10 +263,37 @@ impl TableElem {
...
+                let row = row.iter().flat_map(|entry| entry.as_cell());
+                elem(tag::tr, Content::sequence(row.map(|cell| cell.body.clone())))
+            });
+            let mut rows: Vec<_> = rows.collect();
+
+            let footer = grid.footer.map(|ft| {
+                elem(tag::tfoot, Content::sequence(rows.drain(ft.unwrap().start..)))
+            });
+            let header = grid.header.map(|hd| {
+                elem(tag::thead, Content::sequence(rows.drain(..hd.unwrap().end)))
```
reknih, line 283: Cells within headers should be `th` instead of `td`.

tests/ref/html/basic-table.html
01mf02: What do you think about the following?

~~~ html
    <table>
      <thead>
        <tr>
          <td>The</td><td>first</td><td>and</td>
        </tr>
        <tr>
          <td>the</td><td>second</td><td>row</td>
        </tr>
      </thead>
      <tbody>
        <tr>
          <td>Foo</td><td rowspan="2">Baz</td><td>Bar</td>
        </tr>
        <tr>
          <td>1</td><td>2</td>
        </tr>
        <tr>
          <td colspan="2">3</td><td>4</td>
        </tr>
      </tbody>
      <tfoot>
        <tr>
          <td>The</td><td>last</td><td>row</td>
        </tr>
      </tfoot>
    </table>
~~~

crates/typst-library/src/model/table.rs
01mf02: What about cells within footers?

reknih: Those are normal `td`s

tests/ref/html/basic-table.html
reknih: Spaces between `td`s are allowed as well, though these cells are short enough to look good without spaces

01mf02: I left out the spaces between `td`s for now, as I find the output much more readable that way.

crates/typst-library/src/model/table.rs
01mf02: OK, this is now done in [b2b2593](https://github.com/typst/typst/pull/5666/commits/b2b2593832cc17282f3e41fa54dcb2923cbf7354). That change was really quite hairy ...

reknih (approved):

crates/typst-library/src/html/dom.rs
```diff
@@ -522,6 +522,11 @@ pub mod tag {
                 | self::p
                 | self::pre
                 | self::search
+                | self::table
```
laurmaedje, line 525: Since these are not actually `display: block`, this seems incorrect. We should rather add these to `is_pretty` in `typst-html`.

laurmaedje:
Thanks!

PR #5668, opened 2025-01-08
Consider `separator` for term lists in HTML output.

01mf02:
This PR makes HTML output consider the [`separator`](https://typst.app/docs/reference/model/terms/#parameters-separator) for term lists. It just writes the separator directly after the term to be defined.

## Potential Issues

The default term separator was `HElem`, which triggers a warning during HTML export. For that reason, I made the term separator an `Option<Content>` and consider it for HTML export only when it is `Some(..)`. This might be unwanted, because with this change, the documentation will not state the default value (`h(amount: 0.6em, weak: true)`) anymore. But, on the other hand, this default value does not apply to HTML output in any case.

## Alternative implementation: CSS

It would have been nice to use CSS for this task, such as [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#metadata):

~~~ css
dt::after {
  content: ": ";
}
~~~

However, it seems not feasible to do this in the short term, because apparently, one [can't specify inline styles for pseudo-elements](https://stackoverflow.com/a/14141821) such as `::after`. That means that we would have needed to generate some unique ID for each term list, then write a style just for that unique list in `<style>` at the header, or something like that. That would be a pretty clunky solution.

For users that are not happy with the inlined separators, I propose to use something like:

~~~ typst
#set terms(separator: [: ]) if target != "html"
~~~

However, that currently does not work, because `#target` evaluates to just `target`.


MDLC01:
Using CSS would not work anyway in cases where the content is more complex than just text I think.

tests/ref/html/terms-separator.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
+    <dl>
+      <dt>
+        Mario is:
```
reknih, line 10: Have you considered wrapping the `separator` content in a span with `aria-hidden="true"`? I am not entirely comfortable with how the `separator` content is added to the definition term element's children, semantically becoming part of the defined term.

You already talked about how CSS pseudo elements can only be used to generate text nodes. I agree that `separator` should be able to contain any content. However, using pseudo-elements is the most compatible way to make it clear that the separator is purely presentational and not semantically part of the content. For an optimal solution, I'd like to detect if `separator` is only spacing or only text and use a CSS pseudo-element and use a `<span aria-hidden="true">` for all other cases. Unfortunately, pseudo-elements cannot be specified using inline CSS and stylesheets are not implemented by Typst so far. I _also_ want to avoid an overly verbose tag soup in Typst's output. Therefore, I'd propose the following: If the output is only non-alphanumeric characters, print it as is (doesn't screw up semantics too bad, screen readers mostly just pause). If it is more than that, wrap it in `<span aria-hidden="true">`.

Finally, the question is whether semantics and universal access actually impose different requirements here. In this test, the first list item would render as `Mario is: A famous plumber.` for a sighted user and something along the lines of `Item 1 of 2: Term: Mario, Definition: A famous plumber.` for a screen reader user if implemented as proposed above. In this case, the difference is not to stark, but I can easily imagine how a sighted user could try to change the apparent semantics of the definition list using the separator. Is this usage allowed? If so, we should somehow make the screen reader user privy of the inserted text, maybe using `role="presentational"`. This, however, does not remove the separator from the semantic tree.

01mf02: I think that the underlying problem is that we are trying to guess something that we cannot reliably know: Is user-given `Content` semantically relevant or not? Most likely, in the case of `separator`, the user-given content is not semantically relevant, but we cannot completely rule out that possibility, as you indicate:

> I can easily imagine how a sighted user could try to change the apparent semantics of the definition list using the separator.

I think that we should assume that *any* content is semantically relevant by default (as does HTML), but we should give users a way to mark semantically irrelevant elements, which could then be wrapped in `<span aria-hidden="true">`. That way, users can write something like `#set terms(separator: aria-hidden([: ]))" to indicate that `:` is semantically irrelevant.

laurmaedje:
After some more thought, I'm concerned about switching the default to `auto`. It applies to many more cases than just this one here. Default font? Would need to become auto unless so that we don't unnecessarily force the font. Default font size? Same story. It will virally spread everywhere once we start supporting more CSS.

As outlined in https://github.com/typst/typst/pull/5721#issuecomment-2607532537, @reknih and I think that a general control for the semanticness of the output would be useful. I think this control would also be a good solution for our auto problem here. In the very semantic setting, the separator would simply be ignored and in the more presentational setting, the separator would be written.

This PR would for now be blocked then. But since the auto change goes beyond HTML export, I'd rather not ship it, only to remove it again later (breakingly).

01mf02:
I can understand that reasoning. Let us discuss about more general approaches to solve the "auto problem" first.

laurmaedje:
Same as https://github.com/typst/typst/pull/5678#issuecomment-2676806020

PR #5678, opened 2025-01-10
Support `marker` attribute for `list` in HTML export.

01mf02:
For the example:

~~~ typst
#set list(marker: ([•], [--]))
- Top-level
  - Nested
    - More
    - Items
  - Items
- Items
~~~

This yields:

~~~ html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <ul style="list-style-type: '•'">
      <li>Top-level <ul style="list-style-type: '–'"><li>Nested <ul style="list-style-type: '•'"><li>More</li><li>Items</li></ul></li><li>Items</li></ul></li><li>Items</li>
    </ul>
  </body>
</html>
~~~

Screenshot 2025-01-10 at 11-41-52

## Potential Issues

- There is no space between the marker and the following item.
- Due to putting the markers into CSS attributes, we can only consider plain text markers. Currently, if there is a non-plaintext marker, we simply ignore its non-plaintext parts. Should we emit a warning in such a case?
- Escaping of special characters such as `'` is currently not done. This means that a marker containing `'` might generate invalid CSS, even though in my tests, my browser (Firefox) did not emit any warning.
- Similar to #5668, this change removes the default value for markers from the documentation, which is a bit unfortunate. @laurmaedje proposed instead to keep the default value and generate markers only when the marker value is different from the default, but this may cause some quite unexpected behaviour if users want to explicitly set the marker to its default value ([•], [‣], [–]) in order to produce these marker outputs in HTML, which would cause them to not be generated.


laurmaedje:
Closing this for now. We'll revisit once we have a way to configure the amount of styling that should be applied to the output.

PR #5688, opened 2025-01-11
Fix stroke cap of shapes with partial stroke

### LINKED ISSUE #3578: Stroke cap is ignored for shape with partial stroke
### Description

```typ
#let s = stroke(cap: "round")
#line(length: 5em, stroke: s)
#box(width: 5em, stroke: (bottom: s))
#block(width: 5em, stroke: (bottom: s))
```

image

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

WannesMalfait:
## The problem
To make nicer joins for different stroke widths, the rectangle drawing code has to generate the border shape manually. The code did not take into account possible stroke caps, so that these ended up ignored (see #3578).

## Solution
This adds code to add the stroke caps when appropriate. It allows setting a different cap for the start and the end of each partial segment.

### Example

```typst
#box(
  inset: 1em,
  stroke: (
    left: (
      cap: "round",
      thickness: 2pt,
    ),
    top: (
      cap: "square",
      thickness: 2pt,
    ),
  ),
  radius: (
    top-left: 5pt,
  ),
)[#lorem(20)]
```
lorem text with rounded outline in top left

## Limitations
- Dashed strokes already took the cap into account, so I left them unchanged. Note that they can not have different caps for the parts of a joined segment (because it is drawn as one curve at the moment).
- When the radius is non-zero at a corner, the cap is ignored. The combination of a cap and a radius would be more difficult to handle. I also don't think the result would look very good anyway. Since the radius is mostly used for corners between segments and not for the ends, I didn't think it was worth it to try and come up with a solution for this.

Fixes #3578


laurmaedje:
Thanks for this and sorry for the radio silence!

While experimenting with this, I found that there is still some odd interactions between caps and dashes in a few situations:
```typ
// bug #1: round cap is ignored on left side, but *only* if bottom
//         stroke has a dash.
#block(
  height: 4em,
  width: 5em,
  stroke: (
    left: (cap: "round", thickness: 4pt, dash: "loosely-dashed"),
    bottom: (cap: "square", thickness: 4pt, dash: "loosely-dashed"),
    right: (cap: "square", thickness: 4pt),
  ),
)

// bug #2: cap leads to dash being ignored and left stroke is slightly shifted
//         changing *any* of the caps to "butt" changes the left stroke.
#block(
  height: 4em,
  width: 5em,
  stroke: (
    left: (cap: "square", thickness: 4pt, dash: "loosely-dotted"),
    bottom: (cap: "square", thickness: 4pt, dash: "loosely-dotted"),
    right: (cap: "square", thickness: 4pt),
  ),
)
```

WannesMalfait:
Yes, there is definitely still some weird behavior. This PR tries to only fix the stroke caps in the cases without dashes. When there are dashes present, something else needs to be done. I think the behavior can be improved, but I would leave that for a future PR (as that is beyond the issue this PR is trying to fix).

About the two bugs you mentioned:

#### Bug 1
The current algorithm tries to construct the segments of the border in as big a piece as possible. For this it checks if they have the same dash and width, but ignores the caps. Then, 1 of the sides' stroke is chosen as "the stroke" for the whole segment. In your example, the left and bottom side have the same stroke and the same dash, so they are drawn in one go, using the bottom sides' settings. When you change the dash for the bottom side, they are drawn as two separate lines, and can thus have different caps. This shouldn't be too hard to fix in a future PR.

#### Bug 2
This one has nothing to do with the PR (I think). Because the stroke is too thick, when the cap is set to "square", the dashes overlap, giving the illusion that the dash is ignored. Changing the thickness of the stroke, or the spacing of the dashes can fix this. 

I could not reproduce your behavior with changing the cap to "butt". I only get a change on the left stroke if I change the bottom stroke cap (because of the reason in the 1st bug).

The offset in the left side stroke seems to me to just be an illusion caused by the dash pattern not having a dash at the very end.

Just in case it is relevant: I'm testing everything with PDF export locally. I don't know if the dashes behave differently for the other export targets

crates/typst-layout/src/shapes.rs
```diff
@@ -1279,6 +1305,73 @@ impl ControlPoints {
...
         })
     }
+
+    /// Draw the cap at the beginning of the segment.
+    ///
+    /// If this corner has a radius, or has a stroke before it,
+    /// a default "butt" cap is used.
+    pub fn start_cap(&self, curve: &mut Curve, cap_type: LineCap) {
+        if self.stroke_before != Abs::zero()
+            || self.radius != Abs::zero()
```
saecki, line 1318: I think this is a bit misleading, because if there was a radius:
1. this function wouldn't get called, because `arc_outer()` would return true
2. the butt linecap that is drawn here would be incorrect

Maybe just document that currently no radius is supported?

crates/typst-layout/src/shapes.rs
```diff
@@ -1279,6 +1305,73 @@ impl ControlPoints {
...
+        curve.line(self.end_outer());
+    }
+
+    /// Draw the cap at the end of the segment.
+    ///
+    /// If this corner has a radius, or has a stroke after it,
+    /// a default "butt" cap is used.
+    pub fn end_cap(&self, curve: &mut Curve, cap_type: LineCap) {
+        if self.stroke_after != Abs::zero()
+            || self.radius != Abs::zero()
```
saecki, line 1354: same here

crates/typst-layout/src/shapes.rs
```diff
@@ -1133,6 +1149,16 @@ struct ControlPoints {
 }
 
 impl ControlPoints {
+    /// Rotate point around the origin, relative to the top-left.
+    fn rotate_centered(&self, point: Point) -> Point {
```
saecki, line 1154: Maybe this should be a method on `Corner`?

saecki (changes requested):
I've left a few comments, but the general logic here seems to work out!

I think we can merge this PR in isolation to the other mentioned issues. Since they don't seem all to related, and there seem to be a great deal of them. Some of them I'm not even sure what the correct solution should look like :thinking:

crates/typst-layout/src/shapes.rs
WannesMalfait: I placed it here because it is similar to the `rotate` method just below, and because it doesn't seem to make much sense to me as a method on `Corner`. If you want to, I can still move it of course.

saecki: Yeah, I guess it's fine

laurmaedje:
Thanks!

PR #5702, opened 2025-01-15
Ignore shebang at start of file

### LINKED ISSUE #4899: Ignore hash-bang (#!) at the beginning of .typ files
### Description

The rest of the line, following the hash-bang, should also be ignored.

### Use Case

I have a folder with a bunch of executable scripts that output figures. One of the figures I want to generate using typst code. Sadly, it breaks the (hopefully universal) API of hash bangs. I could create a separate file, of course, but just putting `#!/bin/typst compile` at the top of executable .typ files would be nicer.

T0mstone:
Closes #4899.

There haven't been any new objections after https://github.com/typst/typst/issues/4899#issuecomment-2329484793, which also has 9 thumbs up (including one from me).
I think it's fair to call this change uncontroversial.

Note that a shebang can only be at the very beginning of a file,
so everywhere else `#!` parses just like it did before (as an error).

### DIFF
--- a/crates/typst-syntax/src/highlight.rs
+++ b/crates/typst-syntax/src/highlight.rs
@@ -287,6 +287,7 @@ pub fn highlight(node: &LinkedNode) -> Option<Tag> {
         SyntaxKind::Destructuring => None,
         SyntaxKind::DestructAssignment => None,
 
+        SyntaxKind::Shebang => Some(Tag::Comment),
         SyntaxKind::LineComment => Some(Tag::Comment),
         SyntaxKind::BlockComment => Some(Tag::Comment),
         SyntaxKind::Error => Some(Tag::Error),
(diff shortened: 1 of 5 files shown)


Myriad-Dreamin:
`Shebang` should be a trivia.

https://github.com/typst/typst/blob/c59902e4f945c8bdbf08278fbae6c528309612c8/crates/typst-syntax/src/kind.rs#L356-L364

T0mstone:
> `Shebang` should be a trivia.

It can't occur in code or math mode. Are you sure?

T0mstone:
Actually, that would depend on the file type. If a file is in code mode at the root level, then it could happen. I don't think we support such files, but maybe it's better to future-proof this...

T0mstone:
Since it doesn't hurt, I've just made a commit with that change for now. It can always be dropped later if deemed wrong or unnecessary.

laurmaedje:
While the motivation for this is not particularly strong, I also don't see any real downsides, so I'm fine with it. Thanks!

PR #5718, opened 2025-01-19
Add lcm as an operator in math mode

SekoiaTree:
As discussed starting https://discord.com/channels/1054443721975922748/1088371867913572452/1330227286586163221, although new operators will generally not added, a good case can be made for `lcm` specifically as a counterpart to the already-existing `gcd`. This PR simply adds this operation.


laurmaedje:
Thanks!

PR #5721, opened 2025-01-20
MathML output for HTML

01mf02:
This is work in progress for creating math support in Typst's HTML output via MathML.

Planned features:

- [x] Accenting
- [x] Sub/-Superscript
- [x] Cancelling
- [ ] Case distinction
- [ ] Equations
- [x] Fractions
- [ ] Matrices
- [ ] Text operators
- [ ] Roots
- [ ] Delimiters below/above formulas, such as lines, braces, brackets, ...


wcshds:
`menclose` does not seem to be part of the MathML Core specification, and at least Chromium does not support it. See https://github.com/w3c/mathml-core/issues/245. In this issue, someone suggested a possible trick: https://github.com/w3c/mathml-core/issues/245#issuecomment-2410676518.

01mf02:
@wcshds, thanks for your comment, although this is a real bummer. Just looking at <https://www.w3.org/TR/MathML3/chapter3.html#presm.menclose>, I was under the impression that `menclose` was part of MathML. But it turns out that there is MathML 3, MathML Core, etc. Please pardon my French, but this is horrible.

I used `menclose` after I saw MediaWiki generating it. But I guess that this only works because MediaWiki also generates SVGs (?) on the side.

The workaround that you referenced unfortunately requires a global CSS style, which we currently cannot easily integrate in Typst's HTML output. We could conditionally generate this style if we ever use `menclose` somewhere, but that feels quite ugly, IMO.

I see several options now:

- Do it like Wikipedia and generate SVGs for formulas, while providing MathML for accessibility
- Use a `menclose` polyfill
- Use `menclose` and break `cancel` output for Chrome

Any other ideas?

wcshds:
I previously attempted to write a simple converter to convert mathematical formulas from Typst into MathML. Since I’m not good at coding, my only expectation was that the rendered MathML would look reasonably decent across different browsers. However, I ultimately found that MathML implementations in various browsers are far from satisfying. Therefore, I believe that until browsers improve MathML rendering to an acceptable level, we have to export math formulas as SVG by default, reserving MathML for accessibility purposes. Ideally, there should be an option for users to choose between MathML and SVG export. Additionally, the generated MathML should closely match Typst’s rendering in paged exports (even if just for accessibility purposes).

Below are some issues I encountered while trying to implement a simple converter, which I hope can serve as a reference for you:

1. MathML generated from other sources should ideally include the `semantics` element. See https://www.w3.org/TR/mathml-core/#semantics-and-presentation. Thus, the generated MathML should look something like this:
```
<math display="block">
  <semantics>
    <mfrac>
      <mn>1</mn>
      <mn>2</mn>
    </mfrac>
    <annotation encoding="application/x-typst">$ 1 / 2 $</annotation>
  </semantics>
</math>
```
However, after parsing, the original input is no longer accessible.

2. In MathML, `<mtext>`, `<mo>`, `<mi>`, and `<mn>` in Typst are all represented as `TextElem`, which makes distinguishing between them during MathML generation a major challenge. For example, `$oo$` (infinity) should be `<mn>` rather than another element. Similarly, `sin` is treated as an operator in Typst, but MathML recommends using `<mi>` for functions.

3. MathML automatically italicizes most single-letter variables, including some letters that Typst does not italicize by default, such as `$Delta$`. For such letters, the `mathvariant` attribute should be explicitly added:
```
<mi mathvariant="normal">Δ</mi>
```

4. In matrices (or more precisely, `<mtd>` elements), `columnalign` and `rowalign` are not standard, so alignment should be achieved using `style: text-align`. However, this doesn’t work in Chrome, where matrices are always left-aligned:
```
<math display="block">
  <semantics>
    <mrow>
      <mo form="prefix">(</mo>
      <mtable>
        <mtr>
          <mtd style="text-align: right">
            <mrow>
              <mo>−</mo>
              <mn>1</mn>
            </mrow>
          </mtd>
          <mtd style="text-align: right">
            <mn>1</mn>
          </mtd>
          <mtd style="text-align: right">
            <mn>1</mn>
          </mtd>
        </mtr>
        <mtr>
          <mtd style="text-align: right">
            <mn>1</mn>
          </mtd>
          <mtd style="text-align: right">
            <mrow>
              <mo>−</mo>
              <mn>1</mn>
            </mrow>
          </mtd>
          <mtd style="text-align: right">
            <mn>1</mn>
          </mtd>
        </mtr>
        <mtr>
          <mtd style="text-align: right">
            <mn>1</mn>
          </mtd>
          <mtd style="text-align: right">
            <mn>1</mn>
          </mtd>
          <mtd style="text-align: right">
            <mrow>
              <mo>−</mo>
              <mn>1</mn>
            </mrow>
          </mtd>
        </mtr>
      </mtable>
      <mo form="postfix">)</mo>
    </mrow>
    <annotation encoding="application/x-typst">#set math.mat(align: right)
      $ mat(-1, 1, 1; 1, -1, 1; 1, 1, -1) $
    </annotation>
  </semantics>
</math>
```

5. `underline`/`overline` in Typst cannot be rendered properly in Chrome even when using the NewCMMath font:
```
<math display="block">
  <semantics>
    <munder>
      <mrow>
        <mn>1</mn>
        <mo>+</mo>
        <mn>2</mn>
        <mo>+</mo>
        <mi mathvariant="normal">…</mi>
        <mo>+</mo>
        <mn>5</mn>
      </mrow>
      <mo stretchy="true">_</mo>
    </munder>
    <annotation encoding="application/x-typst">$ underline(1 + 2 + ... + 5) $</annotation>
  </semantics>
</math>
```
Moreover, `underparen`/`overparen` cannot be rendered properly in either Chrome or Firefox. I previously reported this [issue](https://issues.chromium.org/issues/386610915) to Chromium, but they seem to believe it’s entirely a font issue, which makes me a bit disappointed. As a workaround, `underline` and `overline` can be implemented with CSS, but this sacrifices the semantic information of MathML:
```
<math display="block">
  <semantics>
    <mrow style="border-bottom: 1px solid currentColor; display: inline-block;">
      <mrow>
        <mn>1</mn>
        <mo>+</mo>
        <mn>2</mn>
        <mo>+</mo>
        <mi mathvariant="normal">…</mi>
        <mo>+</mo>
        <mn>5</mn>
      </mrow>
    </mrow>
    <annotation encoding="application/x-typst">$ underline(1 + 2 + ... + 5) $</annotation>
  </semantics>
</math>
```

6. The default math font in Chrome is very incomplete. If users choose to export formulas using MathML, they should be advised to select an appropriate math font.

Finally, I strongly recommend testing the results in Firefox, Chrome, and Safari during the development of HTML export, especially for MathML. Due to network issues in my apartment, I cannot test MathML rendering on an iPad via 0.0.0.0, but I believe there are likely many issues as well.

Thank you for your continued work on HTML export!

wcshds:
In my personal opinion, I believe we can keep `menclose`, but when users choose MathML export instead of SVG export, they should be notified that it is not supported in Chrome.

Enivex:
> @wcshds, thanks for your comment, although this is a real bummer. Just looking at <https://www.w3.org/TR/MathML3/chapter3.html#presm.menclose>, I was under the impression that `menclose` was part of MathML. But it turns out that there is MathML 3, MathML Core, etc. Please pardon my French, but this is horrible.
> 
> I used `menclose` after I saw MediaWiki generating it. But I guess that this only works because MediaWiki also generates SVGs (?) on the side.
> 
> The workaround that you referenced unfortunately requires a global CSS style, which we currently cannot easily integrate in Typst's HTML output. We could conditionally generate this style if we ever use `menclose` somewhere, but that feels quite ugly, IMO.
> 
> I see several options now:
> 
> - Do it like Wikipedia and generate SVGs for formulas, while providing MathML for accessibility
> - Use a `menclose` polyfill
> - Use `menclose` and break `cancel` output for Chrome for now (perhaps Chrome will also support `menclose` someday)
> 
> Any other ideas?

MathML Core was created specifically because the original MathML was a mess that never got properly adopted or implemented.

For all intents and purposes MathML Core "is" MathML, and it is what we should be targeting

T0mstone:
Maybe just add the workaround without the styling (so users can add that manually afterhand) and emit a warning until typst can emit global css styles.

01mf02:
@wcshds, thanks a lot for your detailed write-up of your previous experiences! I will let this sink in for a bit ...

01mf02:
> However, after parsing, the original input is no longer accessible.

What do you mean by this?

> 2\. In MathML, `<mtext>`, `<mo>`, `<mi>`, and `<mn>` in Typst are all represented as `TextElem`, which makes distinguishing between them during MathML generation a major challenge. For example, `$oo$` (infinity) should be `<mn>` rather than another element. Similarly, `sin` is treated as an operator in Typst, but MathML recommends using `<mi>` for functions.

I've also remarked that this will be a challenge. On a related note, did you find a nice way to determine whether the top/bottom of an [`attach`](https://typst.app/docs/reference/math/attach) goes to `t`/`b` or `tr`/`br`? This is another pain point I have not sorted out.

> Moreover, `underparen`/`overparen` cannot be rendered properly in either Chrome or Firefox. I previously reported this [issue](https://issues.chromium.org/issues/386610915) to Chromium, but they seem to believe it’s entirely a font issue, which makes me a bit disappointed. As a workaround, `underline` and `overline` can be implemented with CSS, but this sacrifices the semantic information of MathML:

That's another bummer, especially the feedback from the Chromium team.

01mf02:
> However, I ultimately found that MathML implementations in various browsers are far from satisfying. Therefore, I believe that until browsers improve MathML rendering to an acceptable level, we have to export math formulas as SVG by default, reserving MathML for accessibility purposes. Ideally, there should be an option for users to choose between MathML and SVG export. Additionally, the generated MathML should closely match Typst’s rendering in paged exports (even if just for accessibility purposes).

I agree with you. MathML currently seems to be too limited to serve as one and only visual representation, and we should be able to back it up with SVG, which is also the path that Wikipedia takes. What do you think about this, @laurmaedje & @reknih? Should we give users the ability to customize this behaviour (e.g. create only MathML as opposed to MathML + SVG), and if so, how? Command-line parameter? Set rule? ...

Enivex:
What does arxiv do for their html5 versions?

wcshds:
> > However, after parsing, the original input is no longer accessible.
> 
> What do you mean by this?

I mean the annotation section. It would be very convenient to complete this part if we could obtain the user's original input.

```
<annotation encoding="application/x-typst">
  $ 1 / 2 $
</annotation>
```

> On a related note, did you find a nice way to determine whether the top/bottom of an [`attach`](https://typst.app/docs/reference/math/attach) goes to `t`/`b` or `tr`/`br`? This is another pain point I have not sorted out.

I couldn't find an appropriate method, so I had to use a crude method, which is definitely not robust.

I'm a bit embarrassed to show my naive code to experienced programmers, but if this messy code can give you some inspiration, that would be good: https://github.com/wcshds/typst-math-to-mathml-converter/blob/master/src/mathml.rs#L118

perfluorodecalin:
> What does arxiv do for their html5 versions?

I believe they use [LaTeXML](https://math.nist.gov/~BMiller/LaTeXML/) to convert Latex to HTML + MathML. Maybe we can get some inspiration from there.

01mf02:
> I mean the annotation section. It would be very convenient to complete this part if we could obtain the user's original input.

Now I understand. You mean that after Typst is done parsing, we cannot obtain the formula the user typed to embed it into MathML. Well, technically, that should be possible, because we have spans associated with Typst elements. I just do not know a way yet how to get the corresponding part of the input at the time of the HTML export.



> I couldn't find an appropriate method, so I had to use a crude method, which is definitely not robust.
> 
> I'm a bit embarrassed to show my naive code to experienced programmers, but if this messy code can give you some inspiration, that would be good: https://github.com/wcshds/typst-math-to-mathml-converter/blob/master/src/mathml.rs#L118

No need to be embarrassed, I find your code quite nice and readable, and I thank you for showing it to me. That will definitely come in handy.

wcshds:
MathJax is great, but it still has some shortcomings when handling some under/over stuffs. These use cases might be belong to edge cases, but they illustrate that SVG output could be a more reliable choice.
```
<math display="block"><semantics><munder><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><munder><mo stretchy="true">&#x23DF;</mo><mtext><mi>numbers</mi></mtext></munder></munder><annotation encoding="application/x-typst">$ underbrace(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><mover><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><mover><mo stretchy="true">&#x23DE;</mo><mtext><mi>numbers</mi></mtext></mover></mover><annotation encoding="application/x-typst">$ overbrace(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><munder><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><munder><mo stretchy="true">&#x23B5;</mo><mtext><mi>numbers</mi></mtext></munder></munder><annotation encoding="application/x-typst">$ underbracket(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><mover><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><mover><mo stretchy="true">&#x23B4;</mo><mtext><mi>numbers</mi></mtext></mover></mover><annotation encoding="application/x-typst">$ overbracket(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><munder><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><munder><mo stretchy="true">&#x2323;</mo><mtext><mi>numbers</mi></mtext></munder></munder><annotation encoding="application/x-typst">$ underparen(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><mover><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><mover><mo stretchy="true">&#x2322;</mo><mtext><mi>numbers</mi></mtext></mover></mover><annotation encoding="application/x-typst">$ overparen(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><munder><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><munder><mo stretchy="true">&#x23E1;</mo><mtext><mi>numbers</mi></mtext></munder></munder><annotation encoding="application/x-typst">$ undershell(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
<math display="block"><semantics><mover><mrow><mn>1</mn><mo>+</mo><mn>2</mn><mo>+</mo><mi mathvariant="normal">…</mi><mo>+</mo><mn>5</mn></mrow><mover><mo stretchy="true">&#x23E0;</mo><mtext><mi>numbers</mi></mtext></mover></mover><annotation encoding="application/x-typst">$ overshell(1 + 2 + ... + 5, "numbers") $</annotation></semantics></math>
```
image

Overall, I think it’s reasonable to follow the [recommendations](https://www.w3.org/TR/mathml-core/#semantics-and-presentation) in the MathML Core documentation. Typst should always output MathML when exporting math formulas and, by default, include an SVG element under the <semantics> element:

```
<math>
  <semantics>
    <mrow>
      ...
    </mrow>
    <annotation encoding="application/x-typst">
      ...
    </annotation>
    <annotation-xml>
      <svg>...</svg>
    </annotation-xml>
  </semantics>
</math>
```

Of course, users could choose not to include the SVG when performing an HTML export.

If the export includes SVG, users can decide whether to display the MathML or the SVG using CSS:
```
/* Hide the MathML expression and show the svg annoation */
semantics > :first-child { display: none; }
semantics > annotation-xml {
  display: inline-block;
}
```
If users choose not to include SVG, they can be advised to use MathJax instead.

reknih:
I have done quite a bit of research on this based on your discussion.

## Goals

First of all, I agree with @Enivex that we are targeting MathML Core. The nuance is what we should do when there's no spec-compliant way to produce the desired result in MathML Core. Two of our priorities are correct cross-browser rendering and Assistive Technology (AT) support.

From a look at our stats, 52.6% of page views on typst.app are from Chromium-based browsers (which is incidentally rather low compared to the global stats). These browsers have the worst support for MathML Core while Firefox implemented most of the full MathML 3 in version 1. Safari sits somewhere in-between (for example, [they support the non-Core `<menclose>` element](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose#browser_compatibility) that we'd need for `cancel`).

## Existing systems
### Visual fallbacks

I don't think that it is acceptable for a majority of visitors to receive potentially degraded Math experience. As discussed above, Wikipedia displays a raster graphic of the equation and overlays it with an invisible MathML render. A similar technique is used by [MathJax](https://docs.mathjax.org/en/latest/basic/accessibility.html#screen-reader-support). [The rationale behind this decision can be found here.](https://github.com/mathjax/MathJax/issues/938) On top of that, MathJax is pervasive enough so that the screen reader JAWS ships dedicated logic to parse and render their output. MathJax, however, requires accessible interactivity. Formulas rendered by Typst are inert, so we have less difficult requirements there.

A similar approach is conceivable for Typst: Render the formula to SVG and overlay invisible MathML. The SVG would be a sibling of, not contained with the MathML in order to have the browser render it. However, that only really deals with rendering issues as unknown elements will yield a trash experience for most AT users since the browser parses the MathML into an accessibility tree (and thus needs to have support for all its elements) and only then passes it to AT (NVDA and VoiceOver in particular). JAWS has a different approach and does its own parsing.

### Conveying semantics to AT & non-Core elements

Most of the systems I checked have a fallback for a failure to read MathML: Wikipedia provides the LaTeX source in both image alt text and an `<annotation>` element. AT can then choose to parse that LaTeX code if the accessibility tree failed to build (e.g. when using `<menclose>`). MathJax instead serializes the whole formula into natural language and stores it in an `aria-label`. This approach requires localization. Both render `<menclose>` for LaTeX's `\cancel` macro in their MathML outputs, even when visited with Chrome.

Other systems are only concerned with rendering their input to MathML. For example, Pandoc's [texmath](https://github.com/jgm/texmath) will just yield an `<menclose>` for `EBoxed` elements, as will [LaTeXML](https://math.nist.gov/~BMiller/LaTeXML/) for the `\cancel` macro. As LaTeXML is the engine behind arXiv's HTML5 archive, it will also only render MathML without a fallback (see e.g. [here](https://ar5iv.labs.arxiv.org/html/1307.4059)).

CSS-only polyfills for simple uses of [`<menclose>`](https://github.com/w3c/mathml-core/issues/245#issuecomment-2410676518) are available, [polyfills for other elements](https://github.com/w3c/mathml-polyfills/tree/main) also require JavaScript. Most of them require infrastructure for CSS selectors, as they often use pseudo-elements and can thus not be produced exclusively through inline CSS. Ironically, the availability of CSS solutions [slows down the specification of the element for MathML Core](https://github.com/w3c/mathml-core/issues/245), not helping our AT woes.

## Possible approaches

I have discussed with @laurmaedje that Typst's HTML export should ideally cover the spectrum from very semantic to more presentational. For this, we envision discrete sensible steps of presentational-ness (expressed via an enum) that can be changed at any point in the document via a set rule (thus supporting mixed semantical/presentational use cases such as cetz plots). For the purposes of math, this configuration would define whether export should be purely semantic (only MathML export) or _also_ presentational (companion SVG rendering).

We are now left with the following possibilities:

- Use and polyfill non-standard elements (i.e. include `<menclose>` in the output) using CSS and potentially JS (also potentially using feature detection). This yields higher AT fidelity in supported browsers, and increases likelihood of a total AT failure in Chromium.
- Print generic standard elements and then use polyfills to make them render correctly. Both of these approaches rely on [MathML Core's composability with HTML](https://w3c.github.io/mathml/#presm_html_insertion).
  - Some elements can be polyfilled with CSS as mentioned above (i.e.  `<mrow class="menclose">`). Use HTML to convey the semantic implications of an element (i.e. `<mrow><mtext><span aria-label="cancelled" /></mtext><!-- cancelled content --></mrow>`). When only the semantics, not the presentation are needed, the CSS can be omitted.
  - An alternative polyfill approach is to use Typst to draw what the browser cannot draw and render it as an SVG. This approach tries to avoid rendering whole subtrees as SVGs to not unnecessarily switch fonts inline. For example, the output could look like this: `<mrow><mtext><svg aria-label="cancelled"><!-- cancel line --></svg></mtext><!-- cancelled content --></mrow>` (plus CSS for overlaying the two). In Typst, the SVG would be rendered like this: `$ cancel(#hide($/* cancelled content */$)) $`. This will only approximately match the dimensions of the content but avoid shipping JS-based polyfills.
- Serialize Typst math to a format MathJax can ingest (LaTeX, MathML 3, AsciiMath) and have them render it.

Of these options, the second strikes me as the best. If a CSS-only polyfill is available, we can use it (predicated on stylesheet support). If a browser-based polyfill of an element instead requires us to emit JS, we can instead opt for the SVG-based approach. I hope that we'll be able to avoid rendering alphanumeric glyphs in these inline SVGs. I don't like adding MathJax since that would bloat the Typst compiler and its output and add another intermediary representation over whose transformation into the final output we would have little control.

We could complement this rendering with one or multiple of the following as an AT extraction fallback:

- Render the full equation as semantic MathML 3 and paste it into an `<annotation-xml>` element. We need MathML 3 support anyways since PDF/UA-2 and PDF 2.0 (see ISO ISO 32000-2:2020 / PDF 2.0 spec, 14.8.6.3) require it for accessible math (the PDF standard is not aware of MathML Core).
- Output the Typst source code in an `<annotation encoding="text/vnd.typst">` (by the way, that's the correct MIME type for Typst) element.
- Render the formula as natural language into an `aria-label`.
- Allow the user to specify their own, natural language summary for the formula and include it in `<annotation encoding="text/plain">` / `alt`.

I don't see the second option, embedding Typst code, as particularly useful. We are delusional if we think that any AT would try to parse Typst or that it would be a friendlier representation than the usual LaTeX to a consumer not expecting something different. True, Typst math is easier to read and write than LaTeX, but there is still a learning curve, and it doesn't have the same standing as a somewhat universal computer math notation. Additionally, as @laurmaedje pointed out, this would need resolving variables, show rules, etc. and would essentially be its own export target. The only practical use of  `<annotation encoding="text/vnd.typst">` is extraction into another Typst document. The third option, generating an `aria-label` automatically is a lot of work, also given the need for localization.

Overall, our current thinking is the following:

- In the more semantic mode, rendering of the MathML is important since no SVG would be provided. Hence, for each non-Core element, we should decide if a CSS-only polyfill is feasible. If not, render an inline SVG-based polyfill instead and include that. Pending stylesheet support, CSS-only polyfills requiring pseudo-elements fail export instead when in semantics mode. When in presentational mode (which would always include the full SVG rendering), still render MathML for accessibility, using the empty-span method for polyfills.
- Add internal conditionals for switching between MathML Core and MathML 3. We need MathML 3 anyways and it is a nice fallback. Embed it into `<annotation-xml>` if a failure to represent the formula using standard Core elements is detected (no matter whether we are in semantic or presentational mode).
- Allow the user to specify `alt` text for equations.

If we'd follow this approach, next steps would be:

- A list of Typst constructs not easily representable using MathML Core and decisions on the feasibility of a CSS polyfill for each.
- Localizations for the `alt`/`aria-label` text of each polyfilled element.
- Write some examples by hand and test in JAWS, NVDA, VoiceOver, ChromeVox
- An extra argument on the `equation` function to specify Alt text.
- An internal switch for MathML 3 / MathML Core 
- When doing MathML Core rendering, check whether any element was polyfilled. If so, also include the MathML 3 annotation (reprocessing the whole equation).
- Check the rendering of the implementations with the AT listed above

I'm curious to hear your thoughts.

Enivex:
> We need MathML 3 support anyways since PDF/UA-2 and PDF 2.0 (see ISO ISO 32000-2:2020 / PDF 2.0 spec, 14.8.6.3) require it for accessible math (the PDF standard is not aware of MathML Core).

Is MathML Core not a strict subset of MathML 3?

reknih:
> > We need MathML 3 support anyways since PDF/UA-2 and PDF 2.0 (see ISO ISO 32000-2:2020 / PDF 2.0 spec, 14.8.6.3) require it for accessible math (the PDF standard is not aware of MathML Core).
> 
> Is MathML Core not a strict subset of MathML 3?

Kinda yes. The elements are a subset but there is more & different specs for them. But using MathML 3 elements where applicable for PDF accessibility is a good idea I think.

Enivex:
> > > We need MathML 3 support anyways since PDF/UA-2 and PDF 2.0 (see ISO ISO 32000-2:2020 / PDF 2.0 spec, 14.8.6.3) require it for accessible math (the PDF standard is not aware of MathML Core).
> > 
> > Is MathML Core not a strict subset of MathML 3?
> 
> Kinda yes. The elements are a subset but there is more & different specs for them. But using MathML 3 elements where applicable for PDF accessibility is a good idea I think.

Are they actually supported though? Say if I open a PDF with MathML 3 in a chromium based browser.

reknih:
Yeah. I expect it to work in Acrobat and for all other readers, all bets are off. In general, I am inclined to follow the PDF/UA and 2.0 specs here which are quite assertive about using MathML, both for readers and writers. And if you don't, you just aren't a conforming reader...

tgross35:
I don't know how difficult this would be (unfortunately seems likely to be high), but is there any chance this functionality could be extracted to a separate `typst-` crate? There are a ton of possible uses for converting Typst's math syntax to MathML without the rest of the library, like integration into other documentation languages (preprocessing markdown/RST) or doing conversion on the fly via Wasm.

reknih:
> I don't know how difficult this would be (unfortunately seems likely to be high), but is there any chance this functionality could be extracted to a separate typst- crate?

Maybe extracting some of the MathML code into its own crate could be viable, but that wouldn't mean that there'd be a crate that does not need `typst`, `typst-syntax`, `typst-library`, `typst-html` ... to as dependencies that could do Typst to MathML conversion. Since Typst math can contain arbitrary content (e.g. `$x + #rect()$`), all Typst functionality is needed to export a MathML tree.

aaron-jack-manning:
I think ideally it would be possible to take the output of `typst_syntax::parse_math`, and convert it to MathML which is agnostic with respect to styling, with appropriate error handling for when arbitrary markup appears within the input text. However, I suspect this would not "hook in" to the way Typst will internally render HTML very nicely, due to the way the syntax tree is evaluated into content when in the context of an actual document. Doing a direct conversion to other formats based on the syntax tree directly is something I have been working on though for my own purposes, and any official support for it would by amazing, regardless of how large the dependency is on the larger Typst ecosystem.

laurmaedje:
The discussion here is still very valuable, but I'll restart implementation at a later date independently from this PR, so I'll close this.

Issue #661, opened 2023-04-08
Letter p causes underline to toggle on/off


rypervenche:
When using Noto Serif in a header with underline, lowercase P seems to toggle the underline, while lowercase Y also brings it back.

typst 0.1.0 (94e052b8)

```
#show heading.where(
  level: 1
): it => block(width: 100%)[
  #set align(left)
  #set text(24pt, font: "Noto Serif")
  #underline[*#it.body*]
]

= this pad pin pan pot pun yes no
```
![imgur](https://i.imgur.com/xuG2xJl.png)

rypervenche:
It seems that this may only happens when it's bold.

Enivex:
That sure is a strange bug. I would've expected it to at least happen with every letter that has a descender

PR #5727, opened 2025-01-21
Fix underline being drawn inside glyphs

### LINKED ISSUE #5717: underline `evade: true` does not evade the bottom of lowercase g in particular font
### Description

First, download and install the “72” font from [here](https://experience.sap.com/fiori-design-web/downloads/).

Then, use the following code:

```typst
#set text(font: "72", size: 30pt)
#set underline(stroke: 10pt + blue, offset: 5pt, evade: true)

#underline[a gggreat time]
```

Result:

Image

The bottoms of the lowercase g‘s are overlapped by the underline — it has not successfully evaded them. This is highly sensitive to the particular details of the `#set underline`. For instance, changing `offset` to `4pt` makes the issue go away in this case.

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

EpicEricEE:
Fixes #5717.

To find out whether a segment of the underline is inside a glyph, I used the ~fact~ assumption that after a single intersection of the underline with a path segment of the glyph, we are inside the glyph. As this isn't true when the intersection happens tangentially, i.e. when the underline is exactly tangential to the glyph, we need to find out whether that is the case and then not toggle the inside state.

To find out whether the line is tangential to the glyph, I checked the derivative vector of the path segment at the intersection point is purely horizontal. The result is then stored as a boolean alongside the positions of the intersections.

I'm still unsure about what happens when the intersection point is on a cusp. I imagine that the path surrounding the cusp would have to made of two segments, which would then both have to be intersected, so two intersection points should be returned. This would be good, as then the inside state would be updated correctly, but I couldn't find out if that's what actually happens.

---

One weird thing that can be noticed in the test image, but for which I don't really have an explanation, is that the `t` only returns a single intersection point, thus the state is wrongly set to "inside" leading to the line being skipped and ending up with something similar to #661.

Example Image

Blue are the intersection points, green the possible underline segments, and red the tangent line. There should be one more intersection at the right side of the `t`. Maybe that's a font issue, maybe a kurbo issue, or maybe there's something wrong here.


Enivex:
Do we really need such granular collision? Wouldn't it be sufficient to just not draw the underline under any character with a descender?

EpicEricEE:
It sure would be easier, but it wouldn't support the curves of some calligraphy-style fonts like "Allura", and I'd argue that this should be supported.

image

rben01:
Here is an idea I just had that's completely different: draw the entire underline (without skipping descenders), dilate the glyphs horizontally (i.e., “smear” them from left to right), then subtract the dilation from the underline.

Enivex:
> Here is an idea I just had that's completely different: draw the entire underline (without skipping descenders), dilate the glyphs horizontally (i.e., “smear” them from left to right), then subtract the dilation from the underline.

That's essentially what's being done (minus the dilation), but the subtraction part is tricky

EpicEricEE:
The missing intersection seems to be caused by (or at least related to) linebender/kurbo#411. There is also a very small cubic term and points with negative coordinates here, and if I shift everything so that all coordinates are positive, the intersection shows up correctly.

I will close this though, as the intersections returned by kurbo are still sometimes incomplete and thus unreliable, and because this also won't work in glyphs where multiple separate parts overlap, as the assumption that the inside state swaps with every non-tangential intersection then isn't valid anymore.

Issue #5719, opened 2025-01-20
Parsing `enum` in `enum` without intermediate text not working


jbirnick:
### Description

Consider the following snippet:
```typ
#set enum(numbering: "(a)")

+ first main
+ second main
  + first sub
  + second sub
```
![Image](https://github.com/user-attachments/assets/59363f54-c9fd-4ce6-81d8-c04cc07be745)

It works as expected. Now try to remove the "second main" text.
```typ
#set enum(numbering: "(a)")

+ first main
+ 1. first sub
  2. second sub
```
![Image](https://github.com/user-attachments/assets/6114fb82-6e8f-45c3-9ce1-d56a10fdac6d)

This shows "1." where is should show "(a)". Using `+` instead of `1.` for the sublist doesn't help:
```typ
#set enum(numbering: "(a)")

+ first main
+ + first sub
  + second sub
```
![Image](https://github.com/user-attachments/assets/677ee7cf-0b31-4106-8ff3-dd8e239f4a00)

In fact, it reveals that the issue is the _parsing_. The parser doesn't allow to directly parse a "sub enum", without having at least some intermediate text. Funnily enough, this intermediate text can be a single newline character which will be stripped away anyways:

```typ
#set enum(numbering: "(a)")

+ first main
+ 
  1. first sub
  2. second sub
```
![Image](https://github.com/user-attachments/assets/69a6b30f-ea4c-4f58-a55a-f9b4923c86d8)

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

jbirnick:
But I could make the same point for regular text. When I write
```
+ is an operator.
```
in regular text, it will also parse it as an `enum` with one element.

And I think text inside an enum should be parsed just like text is parsed otherwise.

wznmickey:
Yes. You are correct. Similar issues would still happen.

wznmickey:
I implemented one that could handle sub enum

wznmickey:
![Image](https://github.com/user-attachments/assets/5e68591a-b7bc-4ba2-9117-041ebcd096c8)
with
```typst
#set enum(numbering: "(a)")

+ first main
+ + first sub
  + second sub

- first main
- - first sub
  - second sub

```

PR #5677, opened 2025-01-10
Fix list indent when starting at an open bracket

wrzian:
Fixes #5675

Test fails before, passes after.

PR #5728, opened 2025-01-21
Support syntactically directly nested list, enum, and term list

wznmickey:
Fixes  https://github.com/typst/typst/issues/5719


wznmickey:
It would generate sub-enum and sub-list. However, it would also lead to failed existing test results. A detailed requirement on the sub-enum and list may be needed.

laurmaedje:
cc @wrzian

crates/typst-syntax/src/parser.rs
```diff
@@ -157,7 +155,14 @@ fn heading(p: &mut Parser) {
...
 fn list_item(p: &mut Parser) {
-    p.with_nl_mode(AtNewline::RequireColumn(p.current_column()), |p| {
+    let mut column = 0;
+    for i in (0..p.current_start()).rev() {
+        if p.text.chars().nth(i) == Some('\n') {
+            column = i;
+            break;
+        }
+    }
+    p.with_nl_mode(AtNewline::RequireColumn(p.current_start() - column), |p| {
```
wrzian, line 165: The behavior of `Parser::current_column()` was changed in #5677 so that we don't need to edit this. Reverting to the existing code will now do the right thing.

crates/typst-syntax/src/parser.rs
```diff
@@ -107,15 +107,13 @@ fn markup_expr(p: &mut Parser, at_start: bool, nesting: &mut usize) {
         SyntaxKind::Star => strong(p),
         SyntaxKind::Underscore => emph(p),
         SyntaxKind::HeadingMarker if at_start => heading(p),
-        SyntaxKind::ListMarker if at_start => list_item(p),
-        SyntaxKind::EnumMarker if at_start => enum_item(p),
+        SyntaxKind::ListMarker => list_item(p),
+        SyntaxKind::EnumMarker => enum_item(p),
```
wrzian, line 111: This is not correct and is the cause of the test failures. We don't want dashes and other markers within sentences to change:
```typ
- this dash should create a list
but the following dashes - these ones - should not create lists.
```

crates/typst-syntax/src/parser.rs
```diff
@@ -157,7 +155,14 @@ fn heading(p: &mut Parser) {
...
+    for i in (0..p.current_start()).rev() {
+        if p.text.chars().nth(i) == Some('\n') {
+            column = i;
+            break;
+        }
+    }
+    p.with_nl_mode(AtNewline::RequireColumn(p.current_start() - column), |p| {
         let m = p.marker();
         p.assert(SyntaxKind::ListMarker);
         markup(p, false, false, syntax_set!(RightBracket, End));
```
wrzian, line 168:
```suggestion
        markup(p, true, false, syntax_set!(RightBracket, End));
```

Instead we should just change the `at_start` boolean to `true` on the nested call to `markup()` here. This should also happen for enums, and for the _second_ call to `markup()` for term items if we want to be consistent.

tests/suite/model/enum.typ
```diff
@@ -192,3 +192,15 @@ a + 0.
...
+--- issue-5719-sub-enum-numbering ---
+#set enum(numbering: "(a)")
+
++ first main
++ + first sub
+  + second sub
+
+- first main
+- - first sub
+  - second sub
```
wrzian, line 206: Should also add a test for term items and for enums with the `1.` syntax.

And we should have a negative test for headings as well: `= = Nested` should _not_ create a sub-heading unless we also want that.

wrzian (commented):
I left comments on what to change to fix the test failures. The concept in the code is that we now consider markup following a list or enum marker to be "at the start of a line" even if it doesn't follow a new line.

In addition, if we're doing this change we should add term lists for consistency.

tests/suite/model/enum.typ
wrzian: We also need a test for nested headings:
```typ
- list item
- = Heading in the list
  text
  == Sub-heading in the list
  more text
= new heading outside list
```

crates/typst-syntax/src/parser.rs
wznmickey: Yes. It is my question. What kind of sub enum should be accepted. Should the "first sub" be a sub enum?
```
+ first main
+ Something + first sub
                     + second sub
```

crates/typst-syntax/src/parser.rs
wznmickey: Thank you.

crates/typst-syntax/src/parser.rs
wrzian: I don't think we want that. It would require us to change the meaning of an existing parse tree based on later information. This issue already exists with parentheses in code mode, where we have to change the meaning of an already parsed expression, `(x, y)`, if we encounter an arrow:  `(x, y) => ...`. This ambiguity with parentheses already adds significant complexity to the language as a whole, and we don't need more ambiguity here.

wznmickey: So. We only accept sub enum/list/term marks directly following a enum/list/term marks.

wznmickey:
term is added.

wrzian (commented):
Updates look good :+1:

tests/suite/model/enum.typ
laurmaedje: > unless we also want that.

we don't :)

laurmaedje:
Thanks to both of you!

PR #5729, opened 2025-01-22
Vendor packages

stelzo:
This PR adds a `vendor` command to the CLI.

The command creates a vendor subdirectory with all dependencies needed for compiling the project. The compiler then checks for the vendor directory when looking for packages.

This way, templates can be distributed locally without depending on internet access. Vendoring is useful when archiving or in use cases with high reliability requirements.

The command is basically a minimal copy of the `query` command to get the dependency slots populated by compiling the target once. Then it's just copy-pasting into the vendor directory.
I tried to avoid the compilation step in my first implementation but this resulted in a lot of code duplication with the current structure of the codebase. I would be interested in feedback regarding that strategy, since the output document is completely ignored right now. On the other hand, since vendoring is not in the "hot path", I would opt for more code reuse.

The implementation was manually tested with basic projects but there is no thorough automated testing yet. Also, I am not sure if that is needed since the logic is pretty simple and everything else is tested code.
I think it is better to get some feedback on the strategy first, which is also why it's still a draft.


stelzo:
A different way could also be to have a vendor flag on the compile command that triggers the copy after the compilation if the folder doesn't exist and if it exists, it uses the dependencies from it for compiling.

crates/typst-cli/src/vendor.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    // Must iterate two times in total. As soon as the parent directory is created,
+    // world tries to read the subsequent files from the same package
+    // from the vendor directory since it is higher priority.
+    let all_deps = world
+        .dependencies()
+        .filter_map(|dep_path| {
+            let path = dep_path.to_str().unwrap();
+            path.find(DEFAULT_PACKAGES_SUBDIR).map(|pos| {
+                let dependency_path = &path[pos + DEFAULT_PACKAGES_SUBDIR.len() + 1..];
```
PgBiel, line 80: I believe the vendor command should accept `PackageArgs` as well to use the appropriate package directory rather than assuming it will be on the default system-wide location.

crates/typst-cli/src/args.rs
```diff
@@ -342,6 +361,14 @@ pub struct PackageArgs {
...
     )]
     pub package_cache_path: Option<PathBuf>,
+
+    /// Custom vendor directory name.
+    #[clap(
+        long = "package-vendor-path",
+        env = "TYPST_PACKAGE_VENDOR_PATH",
+        value_name = "DIR"
+    )]
+    pub vendor_path: Option<PathBuf>,
```
PgBiel, line 371: Couldn't this be seen as redundant with the `package-path`? And then maybe Typst shouldn't read from the vendor directory automatically, but rather depend on setting the option explicitly. What are your thoughts?

PgBiel (commented):
This seems like a potentially interesting proposal, though I'll wait for further discussion before evaluating it further. In particular, I'm aware that many folks would like to use Typst for archiving purposes, so it'd be nice to hear their opinions on this before proceeding with this PR.

However, as a start, I'll quickly leave some comments on the design of the current implementation. (I'd suggest waiting for further discussion before progressing the implementation, but feel free to do so beforehand if you want and have the time to, of course.)

crates/typst-cli/src/args.rs
stelzo: > And then maybe Typst shouldn't read from the vendor directory automatically, but rather depend on setting the option explicitly.

I had the same idea before but I think it is less intuitive for the user, which is completely subjective. Currently it is more opinionated like Go but I see your point. We could also add the path as a positional arg for a more explicit path.

> Couldn't this be seen as redundant with the `package-path` option?

The `package-path` option is used to find the dependencies for parsing when running the `vendor` command. So I don't see it here.

I still think one okay-ish alternative is to change the compile command by adding the `package-vendor-path` option that, when set, does the vendoring to the given path. But then each following compile command needs to set `package-path` again, which you also need to do when setting a different vendor path right now of course. I just personally like this way of explicit command but implicit reading with defaults like Go is a good balance.

tingerrr:
As @PgBiel has alreayd said there's a little bit of overlap with `--package-path`, and I'd liek to add that you can technically already achieve this with by setting this and the `--package-cache-path` option to a directory within the package root.

One major difference is that it considers `/vendor` a special directory without any additional input, which wold likely be confusing at best and breaking at worst for many users. I don't think @laurmaedje will be a big fan of that. Perhaps special directories are best left to the future where typst could have a canonical default project structure for complex documents (I'm sort of assuming this will come, because git dependencies likely will need a `typst.toml` for regular documents).

People have requested some way to more easily adjust a package's internals when they're in a hurry (think 4am finishing your PhD thesis before print at 10am).

Personally, I wouldn't mind a command which makes it easier to copy/download a package into a project local location (without setting the environment variables).

It could receive a set of positional parameters which are package specs to pull (regardless of if they are dependencies) and a flag to pull all dependencies from a compilation run. After that its up to the user what to do with the vendored package.

laurmaedje:
I agree that this feature doesn't quite seem to pull its weight given that it's already possible today with `typst c file.typ --package-path vendor --package-cache-path vendor`. I also would indeed prefer to not give a special meaning to the directory name `vendor`. While perhaps in line with Go's philosophy, it's not really with Typst's.

laurmaedje:
I'll close this for the reasons mentioned above. Thanks still for your work on this!

PR #5734, opened 2025-01-22
Use a ONE DOT LEADER as the outline's default filler

MDLC01:
This uses U+2024 ONE DOT LEADER instead of a regular period as the default filler in `outline`.

U+2024 decomposes to a regular period so this change is not very important. I just think it is more semantically precise.

It affects a CJK test, I'm not sure if this is desirable.


MDLC01:
Quoting the [Unicode 16.0.0 Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/#G13727):

> However, there are occasional plain text contexts in which a string of leader dots is represented as a sequence of characters. U+2024 ONE DOT LEADER and U+2025 TWO DOT LEADER are intended for such usage.

So it appears using a regular period is more appropriate here. Sorry for the noise.

MDLC01:
This was further discussed there: https://github.com/typst/typst/pull/5735#discussion_r1926055505.

Issue #4943, opened 2024-09-12
Custom page style in outline


f3fora:
### Description

 Related to `outline` and `outline.entry`.
 
I would expect to have be able to customize `page` field of `outline.entry` in such a way that I can have a different style compared to  `body` and eventually not show `page`.

Something like

```typst
show outline.entry.where(level: 4): set outline.entry(page: none)
```

 Related #4859

### Use Case

- In outlines of a presentation (touying, polylux), show the headings of slides without their page.
- In table of content, show the title of `heading(level: high-number+1)` without the page number, which is usually equal to the one of `heading(level: high-number)`
- Show `body` with a style and `page` with another.

laurmaedje:
I'll close this as a duplicate of https://github.com/typst/typst/issues/4859.

laurmaedje:
Ok, it's not as simple as I thought since the page number is provided by the outline and can't just be set to none.

Issue #4859, opened 2024-08-29
Set rules for outline entries


MrPixelized:
### Description

I'm not sure whether to classify this as a bug, but as a typical user, I expected:

`show outline.entry.where(level: 1): set outline.entry(fill: none)`

to work. Alas, due to the parameters in outline.entry being entirely positional, it does not. My request is that they be made keyword parameters.

### Use Case

This would make it way easier and quicker to style a large variety of outlines.

MrPixelized:
Another way of giving people a work around for all such cases would be to have e.g. `show outline.entry.where(level: 1): set outline.entry(_, _, _, _, none)` or `set outline.entry(,,,, none)` retain the pre-set value at the underscore/empty parameters.

bluebear94:
If `outline.entry` isn’t possible to construct from the Typst side, then we could make its fields named without breaking changes.

MrPixelized:
It is possible, but according to people on the Discord server it's highly discouraged/not supposed to be done.

PgBiel:
Outline.entry was created before show-set rules, in their current form, existed, so styling wasn't factored into their design. The proposal seems reasonable to me.

MrPixelized:
I'm interested in implementing this. A simple change to a function's interface seem like it would be a decent first issue. If there are no objections, I'll give it a shot.

PgBiel:
That wouldn't be the only change needed. You'd also need to change the outline implementation in such a way that it doesn't specify entry properties directly, allowing them to be overridden. I'm not sure what'd be the best way to do that, but one could investigate.

MrPixelized:
This morning I decided to check it out. I've got it working for

```typst
show outline.entry.where(level: 1): set outline.entry(fill: none)
```

and

```typst
set outline.entry(fill: none)
```

but

```typst
show outline.entry.where(level: 1): it => {
    set outline.entry(fill: none)
    
    strong(it)
}
```

doesn't work. I have some intuition as to why, but I feel like as an end user I would like this to work (re-constructing `it` using outline.entry() would lead to an infinitely recursive show rule). Can anyone give me some comments on this? Should it? If yes, some hints as to how to get there?

[...]

Issue #4728, opened 2024-08-11
Vertical spacing in outline docs example overrides heading spacing


martinstefanik:
Weak vertical spacing created using `v(2em, weak: true)` does not seem to collapse as expected. Consider the following minimal example.

```typst
#show heading: it => block(spacing: 2em, it)

= Introduction
#v(1em, weak: true)
#lorem(80)

= Analysis
#lorem(50)
```

This is renders as:

![image](https://github.com/user-attachments/assets/b3089fea-f109-409c-b464-da0ec32c7410)

In particular, the vertical space after the _Introduction_ section heading is smaller than that after the _Analysis_ section heading.

The above example is an artificial one but if one redefines the `outline.entry` element as in the example from the documentation [here](https://typst.app/docs/reference/model/outline/#definitions-entry), i.e.

```typst
#show outline.entry.where(level: 1): it => {
  v(12pt, weak: true)
  strong(it)
}
```

one gets the same behavior for the heading of the outline generated using `outline()`.

laurmaedje:
Explicit v spacing has higher priority than block spacing. Otherwise, it would be impossible to override normal block spacing locally.

martinstefanik:
> Explicit v spacing has higher priority than block spacing. Otherwise, it would be impossible to override normal block spacing locally.

Okay, that makes sense. I suppose that explicit `v` spacing has the highest priority then? Would it make sense to mention this somewhere?

The intended behavior in the example above should then be achieved by replacing the heading show rule with:

```typst
#show heading: it => {
  let spacing = 2em
  v(spacing, weak: true)
  it
  v(spacing, weak: true)
}
```

laurmaedje:
> The intended behavior in the example above should then be achieved by replacing the heading show rule with:

The canonical way is `#show heading: set block(spacing: 2em)`, reusing the existing heading's block. Note that there were some changes since 0.11 in how it behaves.

> Okay, that makes sense. I suppose that explicit v spacing has the highest priority then? Would it make sense to mention this somewhere?

Yes, it does. It is [somewhat documented](https://typst.app/docs/reference/layout/v/#parameters-weak): "Weak spacings will always collapse adjacent paragraph spacing, even if the paragraph spacing is larger."

laurmaedje:
Ah, now I get it. Yeah, the example in the outline docs is a bit problematic then.

PR #5735, opened 2025-01-22
Rework outline

laurmaedje:
This PR fundamentally reworks the outline to both improve its out-of-the-box behaviour _and_ make it easier to customize. The general design of the split between `outline` and `outline.entry` remains. However, things start changing a lot from there. This PR is quite the breaking change for existing outline customization, but I hope the changes are for the better. :)

The PR description will focus mainly on the _changes_ and their motivation. For a more direct overview over just the new design, I encourage you to read the updated outline docs instead (in the sources).

### Blocks and spacing

In the old design, outline entries were inline content that was indented with hidden numberings + `h` elements and separated by linebreaks. This was a quite ad-hoc approach that fell short in various ways. It made spacing customization hard and resulted in bad-looking alignment with long titles. It also caused problems with the soon-to-land semantic paragraph changes (which was the initial motivation for my work on the outline).

In the new design, outline entries are blocks. This resolves https://github.com/typst/typst/issues/1424 and fixes https://github.com/typst/typst/issues/4728 and https://github.com/typst/typst/issues/3972, while giving us spacing customization via normal block spacing rules like:

```typ
#show outline.entry.where(level: 1): set block(spacing: 1.5em)
```

### Indentation and multi-line entries

The reworked outline also improves indentation. The `auto` indentation mode will now align  outline-wide[^1]
- all numberings at depth N
- all titles at depth N
- the numbering of all entries at depth N with the titles of entries at depth N+1

This fixes https://github.com/typst/typst/issues/711. Moreover:

- Titles that wrap over multiple lines now nicely wrap at the point at which they start instead of at the very left. (https://github.com/typst/typst/issues/2411)

- Page numbers will not appear on their very own anymore if the title happens to exactly fit, but not the number. Instead, one word of the title and the dots will be pulled into the next line. (https://github.com/typst/typst/issues/2048)

- When the indent is `auto`, but no numbering is configured, there will now be a default indent of 1.2em per level. (https://github.com/typst/typst/issues/421)

Here's how it looks:

Automatic outline indent example

### Methods and properties

Previously, the fill couldn't be customized one a per-entry basis.
With this PR, the `fill` property was moved from `outline` to `outline.entry`. This makes it possible to write rules such as, fixing https://github.com/typst/typst/issues/4859:

```typ
#show outline.entry.where(level: 1): set outline.entry(fill: none)
```

The old outline entry also had the properties `body` and `page`. These properties are somewhat odd: They are populated by the outline itself upon construction of the entries, but they are in reality just derived data: They can be fully created from just the `element`. The new design makes this explicit by building on PR #5733, which introduces methods on elements. The data is now accessible through the methods `body()` and `page()`. In contrast to before, the `body()` does not contain the numbering or supplement anymore. They are split off into a new `prefix()` method. This separation is important for the new indentation features.

There also the new `indented(..)` method which allows you to reuse the built-in indent handling while fully customizing the prefix and inner content to be shown. The default show rule also uses this internally.

Through `page()` being a method, its clearly that it is not settable and I'll thus close https://github.com/typst/typst/issues/4943. The desired outcome is achievable without too much complexity through a show rule (as explained in the updated docs).

### Links

Thanks to https://github.com/typst/typst/pull/5732, linking the full entry instead of just the text and the page number does not resulting in tons of links for each repeated dot. This PR switches to that approach.

### Closing Remarks

The new outline quite heavily builds on methods. I'm interested in feedback on this new design direction of providing composable methods to ease writing of show rules. In think together with more fine-grained selectors and better documentation for built-in show rules, it looks like a promising direction to reduce the complexity of writing show rules.

### Linked issues

- Closes https://github.com/typst/typst/issues/421
- Closes https://github.com/typst/typst/issues/711
- Closes https://github.com/typst/typst/issues/1424
- Closes https://github.com/typst/typst/issues/2048
- Closes https://github.com/typst/typst/issues/2411
- Closes https://github.com/typst/typst/issues/3972
- Closes https://github.com/typst/typst/issues/4728
- Closes https://github.com/typst/typst/issues/4859
- Closes https://github.com/typst/typst/issues/4943

[^1]: In case you are wondering: This outline-wide "global" alignment is implemented with introspection. In some cases, this will add a third introspection iteration where previously two sufficed, but most larger documents will need three iterations either way. If the third iteration is required just for the outline, most of the other stuff will be memoized.


crates/typst-library/src/model/outline.rs
```diff
@@ -1,49 +1,60 @@
...
-/// would also include figures containing tables or other material. For more
+/// list of other kinds of elements than headings.
+///
+/// In the example below, we list all figures containing images by setting
+/// `target` to `{figure.where(kind: image)}`. Just the same, we could have set
+/// it to `{figure.where(kind: table)}` to generate a list of tables.
+///
+/// We could also set it to just `figure`, but then the list would _all_
+/// figures, be it ones containing images, tables, or other material. For more
 /// details on the `where` selector, [see here]($function.where).
```
MDLC01, line 58:
```suggestion
/// We could also set it to just `figure`, without using a
/// [`where`]($function.where) selector, but then the list would contain _all_
/// figures, be it ones containing images, tables, or other material.
```

I believe it is generally bad practice for accessibility to have links with bodies such as "here," or "see here." The same link could be provided more organically using a different phrasing.

Regardless, the first sentence is missing a verb.

crates/typst-library/src/model/outline.rs
```diff
@@ -59,16 +70,89 @@ use crate::text::{LinebreakElem, LocalName, SpaceElem, TextElem};
...
 ///
 /// # Styling the outline
-/// The outline element has several options for customization, such as its
-/// `title` and `indent` parameters. If desired, however, it is possible to have
-/// more control over the outline's look and style through the
-/// [`outline.entry`]($outline.entry) element.
-#[elem(scope, keywords = ["Table of Contents"], Show, ShowSet, LocalName)]
+/// At the most basic level, you can style the outline by setting properties on
+/// it and its entries. This way, you can customize the outline's
+/// [`title`]($outline.title), how outline entries are
```
MDLC01, line 75:
```suggestion
/// [title]($outline.title), how outline entries are
```

It's weird that only this link is displayed in a code font in the paragraph. In the context of the sentence, "title" refers to the concept, not the exact argument. Therefore, I would argue it should not be typeset in a code font.

crates/typst-library/src/model/outline.rs
```diff
@@ -59,16 +70,89 @@ use crate::text::{LinebreakElem, LocalName, SpaceElem, TextElem};
...
+///
+/// #outline()
+///
+/// = About ACME Corp.
+/// == History
+/// ```
+///
+/// [^1]: The outline of equations is the exception to this rule as it does not
+///       have a body and thus does not use indented layout.
+#[elem(scope, keywords = ["Table of Contents"], Show, ShowSet, LocalName, Locatable)]
```
MDLC01, line 150:
```suggestion
#[elem(scope, keywords = ["Table of Contents", "toc"], Show, ShowSet, LocalName, Locatable)]
```

"toc" is a common abbreviation of "table of contents" people might search for.

crates/typst-library/src/model/outline.rs
```diff
@@ -79,8 +163,10 @@ pub struct OutlineElem {
 
     /// The type of element to include in the outline.
     ///
-    /// To list figures containing a specific kind of element, like a table, you
-    /// can write `{figure.where(kind: table)}`.
+    /// To list figures containing a specific kind of element, like an image or
+    /// a table, you can specify the desired kind in a where selector. See the
```
MDLC01, line 167:
```suggestion
    /// a table, you can specify the desired kind in a [`where`]($function.where) selector. See the
```

Linking to the corresponding page here would probably be beneficial.

Regardless of whether a link is added here, "where" should probably be typeset as inline code.

crates/typst-library/src/model/outline.rs
```diff
@@ -121,21 +207,22 @@ pub struct OutlineElem {
...
-    ///   multiplied by its nesting level. Specifying `{2em}`, for instance,
-    ///   would indent top-level headings (not nested) by `{0em}`, second level
+    /// - `{auto}`: Indents the numbering/prefix of a nested entry with the
+    ///   title of its parent entry. If the entries are not numbered (e.g., via
+    ///   [heading numbering]($heading.numbering)), this instead simply inserts
+    ///   a fixed amount of `{1.2em}` indent per level.
+    ///
+    /// - [Relative length]($relative): Indents the entry by the specified
+    ///   length per nesting level. Specifying `{2em}`, for instance, would
+    ///   indent top-level headings (not nested) by `{0em}`, second level
```
MDLC01, line 217:
```suggestion
    ///   indent top-level headings by `{0em}` (not nested), second level
```

This is more consistent with the rest of the sentence.

crates/typst-library/src/model/outline.rs
```diff
@@ -365,136 +343,234 @@ cast! {
...
-    #[required]
+    ///
+    /// ```example
+    /// #set outline.entry(fill: line(length: 100%))
+    /// #outline()
+    ///
+    /// = A New Beginning
+    /// ```
+    #[borrowed]
+    #[default(Some(RepeatElem::new(TextElem::packed(".")).pack()))]
```
MDLC01, line 400:
```suggestion
    #[default(Some(RepeatElem::new(TextElem::packed("\u{2024}")).pack()))]
```

If a textual dot is used for the filler, it should be U+2024 ․ ONE DOT LEADER. Using a non-textual element would be better.

Quoting the [Unicode 16.0.0 Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-6/#G13727):

> Usually leader dots are generated automatically by page formatting software and do not require the use of encoded characters. However, there are occasional plain text contexts in which a string of leader dots is represented as a sequence of characters. U+2024 ONE DOT LEADER and U+2025 TWO DOT LEADER are intended for such usage.

Refer to the full paragraph linked above for more information.

EpicEricEE: Note that most fonts available in the web app don't have a glyph for that character

MDLC01: Then, it may be better to use non-textual elements, or make the text non-selectable.

laurmaedje: I'm not sure that paragraph is saying _anything_ about what typesetting systems _should_ do when generating leaders. It reads more like instructions exclusively for plain text contexts to me.

I'm not sure how non-textual elements can be better here. The font's full stop has the correct size, is baseline-aligned, etc.

As for non-selectability: Yes, that would be good, but we don't currently have infrastructure for that.

crates/typst-library/src/model/outline.rs
laurmaedje: I added "toc" but left "index" out because index is what's typically at the back of a book.

crates/typst-library/src/model/outline.rs
laurmaedje: I think this sentence is also quite relevant: "These leader dot characters can be used to control, to a certain extent, the spacing of leader dots based on font design, in contexts where a simple sequence of full stops will not suffice."

This doesn't really apply to us since `repeat` has a gap property. I added a default gap now for a better out-of-the-box look.

laurmaedje: I'll keep the normal full stop for now.

CMDJojo:
Looking forward to these changes! When are they expected to be released?

laurmaedje:
Hard at work to ship Typst 0.13 soon!

Issue #4717, opened 2024-08-09
Allow any (common) delimiter pair for code grouping in math


Enivex:
Here's a somewhat arbitrary example of an equation I just had to write using typst:

(I've added linebreaks for clarity)
```typ
$
  kappa(lambda) := inf_(u in H_0^1(Omega), u equiv.not 0)
  (display(integral_Omega (abs(nabla u)^2 - g(lambda abs(nabla H)) u^2) dif x))
  /
  (display(integral_Omega u^2 dif x)) 
$
```

It's not terribly important what it looks like, but here you go anyway:
![image](https://github.com/user-attachments/assets/3ea9e2e8-20eb-4342-9660-0084137513b2)

Even in an editor with bracket colorization, I find the code very hard to read, because they're _all_ parentheses. For all its flaws, LaTeX actually does better here, because it uses braces for grouping code. This means that the delimiters for grouping code is different from the one (primarily) used in math.

I propose that we allow both ( ), [ ] and { } for grouping purposes for attachments and fractions. Then the above example could for instance read

```typ
$
  kappa(lambda) := inf_{u in H_0^1(Omega), u equiv.not 0}
  {display(integral_Omega (abs(nabla u)^2 - g(lambda abs(nabla H)) u^2) dif x)}
  /
  [display(integral_Omega u^2 dif x)] 
$
```
To me, this is already a lot more readable, because it removes one layer of parentheses. I can also have my own convention of for instance using {} for numerators and [] for denominators.

We could even consider the radical move of allowing {} and [] for function calls, which would further improve readability in deeply nested scenarios.

PR #5739, opened 2025-01-23
Fix delimiter unparen syntax

### LINKED ISSUE #2979: Different parsing in LR longhand vs shorthand: `lr([..])` vs `[..]`

wrzian:
Closes #2979.

I was looking through old issues and this one was almost solved. Cases A-C were correct, but D was still failing.
```typ
$
  A. frac(1, lr([limits(sum)_i mid(|)])) quad
  B. frac(1, [limits(sum)_i mid(|)]) quad
  C. 1/lr([limits(sum)_i mid(|)]) quad
  D. 1/[limits(sum)_i mid(|)]
$
```
image

Turns out it was just due to one silly parser oversight that no one ever double checked.
```typ
#let item = $limits(sum)_i$
$
  1 / ([item]) quad
  1 /  [item]
$
```
Now renders correctly:
image


wrzian:
I will note that this change is a step away from #4717, but I think that unless we do choose to implement that, I think this change is what _should_ be rendered by the current syntax.

laurmaedje:
Thanks!

Issue #3098, opened 2023-12-28
Improve ratio / relative length docs


PgBiel:
Currently, the docs for relative length[^1] and ratio[^2] are particularly simple. While the examples are helpful, I think it could be useful to mention that:

1. Ratios are usually (but not always) used as part of relative lengths (and thus have a link from ratio to relative length);
2. Ratios in relative lengths are usually (but not always) calculated in relation to the innermost container's width or height (depending on what is being set), or in relation to the page's width or height itself when there is no such container, however **within margins**.
3. However, within your own code, you can use ratios and relative lengths as you'd like. Most importantly, you can multiply a ratio by any number or length to obtain a new proportionally adjusted value.

Related to point 2, it could be useful to mention somewhere as well that, if one wants to apply lengths relative to the whole page's width and height, not being restricted to margins, one has to either create a page with 0pt margins (e.g. `#page(margin: 0pt)[content using 100% as the full page width and height]`) or to add their desired content to the current page's background (with `#set page(background: [content using 100% as the full page width and height])`).

[^1]: https://typst.app/docs/reference/layout/relative/
[^2]: https://typst.app/docs/reference/layout/ratio/

Andrew15-5:
I've added a PR for this.
https://github.com/typst/typst/pull/5750

PR #5750, opened 2025-01-24
Improved ratio and relative length docs

Andrew15-5:
Since we don't have a way to test the docs on the docs page (unlike rust docs) or docs on how link to different stuff work, I have no idea if all my links are valid.

I think I mentioned everything from https://github.com/typst/typst/issues/3098's OP in one way or another. Although, I am not sure what I should write for the relative length. And which part of the patch should be in which of the two files.

Closes https://github.com/typst/typst/issues/3098.


crates/typst-library/src/layout/ratio.rs
```diff
@@ -8,7 +8,47 @@ use crate::foundations::{repr, ty, Repr};
 
 /// A ratio of a whole.
 ///
-/// Written as a number, followed by a percent sign.
+/// Written as a number, followed by a percent sign. A common use case is
+/// setting the width or height of a container (e.g., [block], [rect], etc.),
+/// thus it is used as a part of a [relative length](relative) (because
+/// internally containers use relative length for width and height fields):
```
PgBiel, line 14: I feel like something like this might be better:

```suggestion
/// setting the width or height of a container (e.g., [block], [rect], etc.),
/// as it can be used as part of a [relative length]($relative) to represent
/// a certain percentage of the surrounding container's - or the current
/// page's, if there isn't one - corresponding dimension. For example:
```

crates/typst-library/src/layout/ratio.rs
```diff
@@ -8,7 +8,47 @@ use crate::foundations::{repr, ty, Repr};
...
+/// `{25%}`, which means "get 25% of the width of the innermost container" (240
+/// ⋅ 0.25 = 60). The reason it shows `{50pt}` instead of `{60pt}` is due to
+/// the default value of inset. If we set it to `{0pt}`, then we will get
+/// the expected result (although the number will be cramped):
+///
+/// ```example
+/// #block(width: 240pt, {
+///   rect(width: 25%, inset: 0pt, layout(size => size.width))
+/// })
+/// ```
```
PgBiel, line 33: I think this part could be more straightforward by just specifying both width and inset in one example and explaining how much the percentage corresponds to in that situation.

crates/typst-library/src/layout/ratio.rs
```diff
@@ -8,7 +8,47 @@ use crate::foundations::{repr, ty, Repr};
...
+/// Typst to use the full length you have a few options (this highly depends on
+/// your exact use case):
+///
+/// 1. set page margins to `{0pt}` (`#set page(margin: 0pt)`)
+/// 2. multiply by a full length value (`{21cm * 69%}`)
+/// 3. use padding which will negate the margins (`#pad(x: -2.5cm, ...)`)
+/// 4. use page [background](page.background)/[foreground](page.foreground)
+///    field that doesn't use the margins (note that it will render the content
+///    outside of the document flow, see [place] to control the content
+///    position)
```
PgBiel, line 47: I see now that this feels a bit out of place... Perhaps we could keep this level of detail to relative length docs instead, and link to it explicitly (e.g. "see (link) for more details").

crates/typst-library/src/layout/ratio.rs
```diff
@@ -8,7 +8,47 @@ use crate::foundations::{repr, ty, Repr};
...
+///
+/// However, within your own code, you can use ratios as you'd like. You can
+/// multiply ratio by ratio, [length], [relative length](relative), [angle],
+/// [int], [float], and [fraction].
```
PgBiel, line 51: Would be nice to have an example with `27% * something` under this, and/or use `scale` as an example of how ratio can be used for broader purposes.

PgBiel (commented):
Thanks! This is a great start. I've added some notes, and I urge others to comment as well if they have further suggestions.

crates/typst-library/src/layout/ratio.rs
```diff
@@ -8,7 +8,47 @@ use crate::foundations::{repr, ty, Repr};
 
 /// A ratio of a whole.
 ///
-/// Written as a number, followed by a percent sign.
+/// Written as a number, followed by a percent sign. A common use case is
```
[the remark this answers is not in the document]
Andrew15-5: I didn't know where to add, and making 2 super short lines feels weird. The first one, I think, is always short for all elements, but the second one?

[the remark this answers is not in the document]
PgBiel: The second line could be extended if necessary, but the two pieces of text are definitely separate (general description and use case example) and should be separated by a par break.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: Nnoo, it sounds very complex and weird. Why there is "-" before "or"? "container's" what? I completely don't understand the "if there isn't one - corresponding dimension" part. Corresponding to what? There is no way there is nothing outside: there is always page dimensions, if there is no other container wrappers.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: Okay.

Andrew15-5: I also wanted to ask what is the best way to show math stuff in the docs.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: I changed it a little in the new commit.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: Changed it.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: I moved it, but is it OK to talk about ratio inside of relative length? Maybe add some preamble?

crates/typst-library/src/layout/ratio.rs
Andrew15-5: Added simple and nice example. I think with this much lines it's hard to keep track of stuff in the output, so I added labels.

crates/typst-library/src/layout/ratio.rs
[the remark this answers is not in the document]
Andrew15-5: Done.

crates/typst-library/src/layout/ratio.rs
```diff
@@ -10,6 +10,54 @@ use crate::foundations::{repr, ty, Repr};
...
+///   rect(width: 25%, inset: 0pt, layout(size => size.width))
+/// })
+/// ```
+///
+/// Here the block width is set to `{240pt}` (just to demonstrate the use of
+/// ratio with containers), and inside of it the rectangle width is set to
+/// `{25%}`, which means "get 25% of the width of the innermost container" (240
+/// ⋅ 0.25 = 60). Notice that the inset is equal to `{0pt}`, if it's not set
+/// then it will show `{50pt}` instead of `{60pt}`, which is also why the number
+/// looks cramped.
```
laurmaedje, line 29: This example is too complicated as a first example in my opinion. I'd not start with `layout(.. => ..)` right away.

crates/typst-library/src/layout/ratio.rs
```diff
@@ -10,6 +10,54 @@ use crate::foundations::{repr, ty, Repr};
...
+///   inset: (x: 2pt),
+///   table.vline(x: 1, stroke: none),
+///   [#ratio:], [#(27% * 10%)],
+///   [#length:], [#(27% * 100pt)],
+///   [#relative:], [#(27% * (10% + 100pt))],
+///   [#angle:], [#(27% * 100deg)],
+///   [#int:], [#(27% * 2)],
+///   [#float:], [#(27% * 0.37037)], // Some rounding is happening.
+///   [#fraction:], [#(27% * 3fr)],
+/// )
```
laurmaedje, line 58: This seems to be duplicated.

crates/typst-library/src/layout/ratio.rs
laurmaedje: This seems like it might fit better with the page docs?

crates/typst-library/src/layout/ratio.rs
Andrew15-5: Choose which one to keep.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: You can change the example to what feels best.

crates/typst-library/src/layout/ratio.rs
laurmaedje: How about a in-docs (i.e. Markdown) table listing the supported operations?

Andrew15-5: Lets.

Wait, do you mean not only multiplication? I don't remember, but I think there are only multiplication and division, but I only checked multiplication.

laurmaedje: I didn't mean anything specifically.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: If this is resolved, mark it as so.

Andrew15-5:
https://github.com/typst/typst/pull/5750#discussion_r1931358067:

> I also wanted to ask what is the best way to show math stuff in the docs.

I think this should be moved to the contribution docs. Maybe the docs site can use MathML or something else for inline and potentially block math. Of course, the math syntax should be the one of Typst. I want to make a separate issue about this.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: The docs cites "Guide for Page Setup". In which file should I move this?

crates/typst-library/src/layout/ratio.rs
Andrew15-5: > This seems to be duplicated.

Finally found it: https://github.com/typst/typst/pull/5750#discussion_r1931386217 (the choosing is implicitly mentioned here), though kind of too late.

Andrew15-5: Take a look.

crates/typst-library/src/layout/ratio.rs
Andrew15-5: I don't have another example where I _continue_ with `layout`. Which means I either must add another example and move `layout` there, or change this one. I'm not really a creative guy example-wise, so I don't know how to proceed.

Andrew15-5:
Looks nice. Really want to see how it looks in the rendered docs now. ;)

Issue #5748, opened 2025-01-24
Font not united for CJK in raw blocks


earmer:
### Description

![Image](https://github.com/user-attachments/assets/47248dce-e13e-4779-9dab-9878870ebc60)

As mentioned above, there are errors in the font and text detection in the original block.

List of errors:

- The font is not uniform, and the Japanese-style font is mixed with the Chinese mainland font.

- Mistakenly added a space between CJK and Latin characters (automatic space function issue).

- The language automatic detection algorithm failed.

Reason (maybe because I haven't read the source code of Typst):

- Use the serval keyword to judge the language environment.

  - Add `#set text(lang: "zh")` doesn't work.

- Lang env detects an error in the trigger. Adding "f" can change the font.

- Style Issues:

  - Should the next character use the style of the previous character? -- This problem was solved incorrectly.

  - The judgment of adding spaces between CJK and Latin characters is not strictly written.

  - Automatically return to the default font of CJK.

### Reproduction URL

https://typst.app/project/rab9abXLvijPP5uiBT8NcO

### Operating system

Web app, macOS

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
There isn't really automatic language detection. For proper typesetting, you need to set both the language and a good font.

We're aware that the out-of-the-box experience is bad, but it's more due to missing features than any of the existing systems working incorrectly. Related: https://github.com/typst/typst/issues/5040

Note that raw blocks also set the language to English by default because code is often in English. To override this, you need to write
```typ
#show raw: set text(lang: "zh")
```

hongjr03:
There is a solution in [FAQ of typst-doc-cn's guide](https://typst-doc-cn.github.io/guide/FAQ/chinese-in-raw.html), have a check on it! 🎉

earmer:
@laurmaedje Thanks for the reply, but I have to disagree on a few points you missed.

First, calling this an "out-of-box experience issue" doesn't really hold up. The font style behavior is clearly undefined and needs proper definition. Plus, we could label any bug or feature request as an "out-of-box experience issue" - if an explanation works for everything, it's probably not explaining anything.

You also missed a key point: `cjk-latin-spacing` shouldn't be getting into Raw blocks at all. Your own example with `#show raw: set text(lang: "zh")` shows that Raw blocks are supposed to ignore normal font/language settings. This just proves my point - the whole question of what styles affect Raw blocks is completely undefined and messy right now.

[...]

PR #5753, opened 2025-01-25
Disable cjk_latin_spacing in raw by default

hongjr03:
Closes https://github.com/typst/typst/issues/5760.

A small fix that mentioned in  https://github.com/typst/typst/issues/5748. Disable cjk_latin_spacing in raw by default.


laurmaedje:
What is the reason for only doing this if `block: true`?

hongjr03:
> What is the reason for only doing this if `block: true`?

nah there's no reason... it was a mistake.

laurmaedje:
Could you also add a test for this?

hongjr03:
> Could you also add a test for this?

no problem!

hongjr03:
> Could you also add a test for this?

Hi! I have already written the new test and it should works fine now 🎉

laurmaedje:
Thanks!

PR #5755, opened 2025-01-25
Fetch output format in Typst

Andrew15-5:
Since currently it's not possible to conditionally use paged output (e.g., PDF) and HTML output (e.g., when overriding body tag), I whipped up this workaround.

It adds `sys.output-format` that can be `"pdf"`, `"png"`, `"svg"`, or `"html"`. For `typst query` it will use `"pdf"` since we can't resolve the output format for that command (could potentially change it to none or `"query"`).

I thought that instead of just providing the same `target()` output but non-contextual, I would just add all inputs. This can make the document more customizable. Which is actually kind of a separate proposal, though it is not really how Typst should work.

```typ
#if sys.output-format != "html" {
  set page(...)
  [document-body]
} else {
  html.elem("body", attrs: ..., html.frame([document-body]))
}
```

### DIFF
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -37,9 +37,13 @@ type CodespanError = codespan_reporting::files::Error;
 /// Execute a compilation command.
 pub fn compile(timer: &mut Timer, command: &CompileCommand) -> StrResult<()> {
     let mut config = CompileConfig::new(command)?;
-    let mut world =
-        SystemWorld::new(&command.args.input, &command.args.world, &command.args.process)
-            .map_err(|err| eco_format!("{err}"))?;
+    let mut world = SystemWorld::new(
+        &command.args.input,
+        &command.args.world,
+        &command.args.process,
+        Some(&command.args),
+    )
+    .map_err(|err| eco_format!("{err}"))?;
     timer.record(&mut world, |world| compile_once(world, &mut config))?
 }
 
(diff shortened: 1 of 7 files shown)


laurmaedje:
The `target()` API is designed exactly for this need. It is intentionally
- more coarse in granularity (paged/html instead of pdf/png/svg/html) so that users can continue relying on pdf/png/svg looking the same.
- contextual because with laid-out-SVG-in-html (via `html.frame`), the mode of operation can switch within the document.

Adding an additional API that serves the same purpose is redundant and will lead to confusion. Moreover, it will cause problems down the road if people use this API instead of the contextual API in packages because then their code will not compose correctly with `html.frame`.

The `target()` API is also intentionally only supported behind the HTML feature because it is not yet finalized.

Andrew15-5:
I'll close it, since it's highly unlikely something will come out of it, fix- or proposal-wise. But I think it's still a nice workaround and a feature for those who really want the ability to check the output format of the document (and use it conditionally for virtually anything). For example, you can just change background or margin, or make completely different outputs.

Andrew15-5:
Oops, already closed.

SillyFreak:
You can always use `--input format=...` if your document depends on stuff that Typst doesn't expose to you. As for whether Typst _should_ expose the output format, I agree with Laurenz.

Andrew15-5:
Since you've mentioned it, this doesn't work in the web app. I think the web app uses a different World. If so, then I can't change it myself, but if I could, then `sys.output-format` would work in the web app as well.

SillyFreak:
in the web app this depends on https://github.com/typst/webapp-issues/issues/391, yes (sort of, that issue is specifically about batch `--input` but if it's added the basic use case would also be supported I'd imagine)

Andrew15-5:
Yeah.

Issue #311, opened 2023-03-25
Behavior of first line indentation in paragraphs seems limiting


araruna:
As I gathered [from the docs](https://typst.app/docs/reference/layout/par/) (as seen on the reference for the `first-line-indent` argument), the first paragraph on a page is special-cased to never have it's first line indented.

If I understand this correctly, that means not all paragraphs will have indentation applied, just like LaTeX does by default. If this is correct, then some styles will not be able to be represented with Typst. For example, in Brazilian Portuguese documents we are used to apply indentation to every paragraph, regardless of it's length or position in the text, and as such we use the LaTeX package `indentfirst` pervasively.

Did I understand it correctly?

laurmaedje:
We'll add some configuration for this, so that both ways are possible.

chris-morgan:
Possibly related, paragraphs starting with something other than unformatted text don’t get indented (e.g. `#super[x]`). All of this makes the feature feel temperamental.

laurmaedje:
The bug with unformatted text not being indented should be fixed since the last release. Can you still reproduce it? More generally, I tried to fix the first line indent, but its not as simple as it seems. As currently _all_ text is layouted as a paragraph in Typst, removing the rule would indent _everything._ That includes headings, code blocks, captions, words in boxes, etc. Since paragraphs are not clearly demarked (as they would be in, say HTML), we first need some automatic way to distinguish these two kinds of texts (possibly through the context they appear in).

zhonghes:
this problem is still exist
![image](https://user-images.githubusercontent.com/114350562/231048235-1b7637cc-36e8-434c-aa6b-d804850270bf.png)

laurmaedje:
Yes, I know. As mentioned in my previous response, the overall issue is difficult to fix. The only thing that was fixed so far is that formatted text wasn't even indented when it wasn't in the first paragraph.

DawnEver:
I found another para `hanging-indent`
if you want to indent first line in 1 em
```typst
set par(justify: false, leading: 1.5em, first-line-indent:2em )
```
you can do like this:
```typst
set par( hanging-indent:-1em )
```

wychwitch:
@DawnEver oh wow, I never thought if using a negative hanging indent, great idea! Though if you have a border around the text you'd have to adjust the the block's padding so it wouldn't clip out of it. However it looks like you got around that by setting the first-line-indent to be 2em already! I'll try it out sometime, thanks for sharing :)

[...]

PR #5768, opened 2025-01-27
Support first-line-indent for every paragraph

laurmaedje:
This PR adds support for applying `first-line-indent` to every[^1] paragraph as opposed to just consecutive ones. This is opt-in:

```typ
#set par(first-line-indent: (amount: 1.5em, all: true))
```

The value of `first-line-indent` can now be either just a length (which implies `all: false`) or a dictionary with `amount` (length, required) and `all` (bool, optional) keys.

This fixes the long-standing https://github.com/typst/typst/issues/311 (currently the oldest still-open bug). The reason this bug was so hard to fix is that Typst's notion of a paragraph was very weak until recently: Every piece of text was a paragraph, so this setting would have affected everything --- headings, figure captions, a single word in a box, etc. Solving this required larger internal reworks, which were completed in https://github.com/typst/typst/pull/5746.

[^1]: There is a small exception: The very first paragraph of a list/enum/term item is not indented as it just looks bad to have a gap between the bullet/number/term and the text.


JakobTopholt:
So just to understand this right, if I set
 
 `#set par(first-line-indent: (amount: 0em, all: true))`
 
 I get rid of the indentation on the first line of all paragraphs?

laurmaedje:
There is no indentation by default. If you are using a template, it might have configured first line indent. The rule you mention will disable it, but so will just `set par(first-line-indent: 0pt)`, the `all: true` doesn't add anything if the indent is zero.

JakobTopholt:
Ahhh thanks, I am using a template, I'll look into it, ur a champ @laurmaedje ,some friends and I have turned a lot of comp sci people at AAU onto Typst <3

PR #5772, opened 2025-01-28
Fixed typo in the new outline docs

Andrew15-5:
I don't know if this is a typo, but it does look like one. Perhaps the whole sentence should be rephrased.


laurmaedje:
Thanks!

PR #5773, opened 2025-01-28
Resolve bound name of bare import statically

laurmaedje:
This PR changes how the name bound by a bare import (no import list, no renaming) is resolved. Instead of using a dynamic name resolved from the module value, the name is now resolved statically through the AST. Imports with either renaming or an import list are not affected.

The name can be resolved if:
- The import source is an identifier: `import calc` (though this import has no effect and will now warn)
- The import source is a field access: `import cetz.draw`
- The import source is a string where
  - the string is a valid package spec
  - the string is a valid path and the file stem is a valid identifier

If the name cannot be resolved statically, one of the following two diagnostics will be shown:
- ```
  error: dynamic import requires an explicit name
    ┌─ hi.typ:3:8
    │
  3 │ #import "ot" + "her.typ"
    │         ^^^^^^^^^^^^^^^^
    │
    = hint: you can name the import with `as`
  ```
- ```
  error: module name would not be a valid identifier
    ┌─ hi.typ:4:8
    │
  4 │ #import "with space.typ"
    │         ^^^^^^^^^^^^^^^^
    │
    = hint: you can rename the import with `as`
  ```

This is of course a breaking change, but I expect little breakage in practice (and the breakage that remains is for code that really shouldn't exist in the first place).

### Motivation
In the example below, currently the name `the-file` is bound because `import mymod` will dynamically bind the name of the module (which is `the-file`).

```typ
#import "the-" + "file.typ" as mymod
#import mymod
#the-file // <-- this does not error right now, but really should
```

That behaviour was highly unpredictable. To an extent, the same applies to star imports but at least there that's expected and intended behaviour. Here, it is not at all.

In this second example, currently the name `other-%` is bound, but it is not accessible because it's not a valid identifier.
```typ
#import "other-%.typ"
```


laurmaedje:
Changed error messages from
- ```
  error: cannot determine binding name for this import
    ┌─ hi.typ:3:8
    │
  3 │ #import "ot" + "her.typ"
    │         ^^^^^^^^^^^^^^^^
    │
    = hint: the name must be statically known
    = hint: you can rename the import with `as`
    = hint: to import specific items from a dynamic source, add a colon followed by an import list
  ```
- ```
  error: cannot determine binding name for this import
    ┌─ hi.typ:4:8
    │
  4 │ #import "with space.typ"
    │         ^^^^^^^^^^^^^^^^
    │
    = hint: the file stem is not a valid identifier
    = hint: you can rename the import with `as`
  ```

to

- ```
  error: dynamic import requires an explicit name
    ┌─ hi.typ:3:8
    │
  3 │ #import "ot" + "her.typ"
    │         ^^^^^^^^^^^^^^^^
    │
    = hint: you can name the import with `as`
  ```
- ```
  error: module name would not be a valid identifier
    ┌─ hi.typ:4:8
    │
  4 │ #import "with space.typ"
    │         ^^^^^^^^^^^^^^^^
    │
    = hint: you can rename the import with `as`
  ```

Issue #3046, opened 2023-12-22
Footnote link annotation rect mismatches superscript


daskol:
### Description

Actual link rect differs significantly from superscript rect.

I use the snippet below for typesetting title page of an article.

```typst
// Render authors.
#let make_title() = {
  set align(center)
  set text(weight: "bold")
  [Anonymous Authors]
  
  // Add footnote with authors affilations and notice string.
  set super(typographic: false, size: script-size)
  set text(weight: "regular")
  [#footnote(footnote_body)]
}
```

Then I've edited page annotation with MuPDF in order to add border and color properties of an annotation.

![image](https://github.com/typst/typst/assets/9336514/867c683c-b5f0-4373-bf4a-15daa0eea3e2)

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

MDLC01:
In the latest development version (Aug 1, 2025), the link rectangle seems to be the size of the whole line height (extending even below the superscript). I think this is a desirable behavior, as the superscript alone might be too small. And if it's not desirable, it's at least a different issue, so we can close this one.

PR #5777, opened 2025-01-29
Use "subs" and "sups" font features for typographic scripts

### LINKED ISSUE #633: Superscript inconsistent size
The font size of the superscript seems to be inconsistent depending on the text content. 
Here I showed an example with a code sample and the generated content 

```
Test#super("1"), Test#super("2"), Test#super("1,3")"
```

Screenshot 2023-04-06 at 17 16 59

### LINKED ISSUE #2146: Bold subscripts have wonky letter positioning
### Description

I think the issue is best described in the following image:

image

Basically, bold text-mode subscripts have very wonky positioning. This issue does not affect superscripts or math-mode subscripts.

Additionally, I'm not very sure but it seems like even `R#sub[peak]` has inconsistent spacing in between letters.

````typst
R#sub[peak] #sym.arrow.l ```typst R#sub[peak]```

*R#sub[peak]* #sym.arrow.l ```typst *R#sub[peak]*```
````

### Reproduction URL

https://typst.app/project/r6hqHZ9hgfPdvR9oJYlaTi

### Operating system

Web app

### Typst version

### LINKED ISSUE #3129: Footnote can break to next line on its own when custom font is set
### Description

I don't even know what could be causing this but no matter how much I messed with the `width` I could not reproduce this without setting the `font` field of `text`.
Minimal reproducible example:
```typ
#block(width: 4.5cm)[
  #set text(font: ())
  #set par(justify: true)
  Reproduce with English text.#footnote[hi]
]
```
Renders as:
Rendered using Typst-Renderer on the official Typst Discord: https://discord.com/channels/1054443721975922748/1056574658771681361/1192172744658407504

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

### LINKED ISSUE #4006: Only the first three footnote numbers are set bold for some fonts
### Description

In the following code, the first three numbers of the footnotes are bold (?), and from four onward they are not anymore. However, this does only happen with some fonts (like _CMU Serif_ or _CMU Serif Extra_).

```typ
#set text(
  font: "CMU Serif",
)

= Header

Some text with footnotes.

test #footnote[asdf 1 asdf]
test #footnote[asdf 2 asdf]
test #footnote[asdf *3* asdf]
test #footnote[asdf 4 asdf]
test #footnote[asdf *5* asdf]
```

Result:

image

and

image

(Using `typst 0.11.0 (2bf9f95d)`)

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

### LINKED ISSUE #5112: Typographic superscript / subscript spaces are too large
### Description

In fact, with `typographic: true`, spaces are kept as regular spaces. I'm unsure this is desirable.

```typ
#super(typographic: true)[1 2 3] \
#super(typographic: false)[1 2 3]
```

The spaces in the first line are very large.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

MDLC01:
This changes the `typographic` parameter of `sub` and `super` to use the "subs" and "sups" font features instead of special Unicode characters that should not be used that way.[^1]

- Closes #633.
- Closes #2146 by definitely making it a font issue (before this PR, we used characters that were not necessarily meant to be used together to write text).
- Closes #5112.
- Related to #3046.

[^1]: From [Section 7.8.1 of the Unicode 16.0.0 Core Specification](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-7/#G24762): "Superscript modifier letters are intended for cases where the letters carry a specific meaning, as in phonetic transcription systems, and are not a substitute for generic styling mechanisms for superscripting of text, as for footnotes, mathematical and chemical expressions, and the like."


MDLC01:
New Computer Modern, Roboto, Liberation, do not support "sups" even for digits. So `show footnote: set super(typographic: true)` is probably a bad idea.

MDLC01:
> It might actually be good to keep using the superscript characters for numbers, they tend to look better IMO.

But as you noted later, they shouldn't really be used that way. The real solution is to use the font features (at least for footnotes, it should be the default imo), which does indeed look betters with fonts that support it (e.g., Libertinus Serif uses the same glyph for "²" and `sups` "2" I believe)

> The question being whether typst's output counts as "rich text"

I don't see any reason why it wouldn't. Quoting [Wikipedia](https://en.wikipedia.org/wiki/Formatted_text):

> In [computing](https://en.wikipedia.org/wiki/Computing), **formatted text**, **styled text**, or **rich text**, as opposed to [plain text](https://en.wikipedia.org/wiki/Plain_text), is [digital text](https://en.wikipedia.org/wiki/E-text) which has styling information beyond the minimum of semantic elements: colours, styles ([boldface](https://en.wikipedia.org/wiki/Boldface), [italic](https://en.wikipedia.org/wiki/Italic_type)), [sizes](https://en.wikipedia.org/wiki/Point_(typography)), and special features in [HTML](https://en.wikipedia.org/wiki/HTML) (such as [hyperlinks](https://en.wikipedia.org/wiki/Hyperlink)).

crates/typst-library/src/text/shift.rs
```diff
@@ -17,17 +14,18 @@ use crate::World;
 /// ```
 #[elem(title = "Subscript", Show)]
 pub struct SubElem {
-    /// Whether to prefer the dedicated subscript characters of the font.
+    /// Whether to use the `subs` OpenType feature to render the glyphs.
+    ///
+    /// Do not use this is the body contains non-text elements.
```
PgBiel, line 19: (Also: why not?)
```suggestion
    /// Do not use this if the body contains non-text elements.
```

crates/typst-library/src/text/shift.rs
```diff
@@ -53,11 +51,15 @@ impl Show for Packed<SubElem> {
...
-                    return Ok(TextElem::packed(text));
-                }
+            if !is_text(&body) {
+                engine.sink.warn(warning!(
+                    self.span(),
+                    "typographic subscript element contains non-textual content";
+                    hint: "this may have undesired behavior";
+                    hint: "set the `typographic` parameter to `false`"
+                ))
             }
```
PgBiel, line 61: Not sure whether I like this, seems to be a bit too broad of a warning, and either way the output could be desired. We should, at the very least, add a pathological case to testing to verify whether it's worth adding this "soft-restriction".

crates/typst-library/src/text/shift.rs
```diff
@@ -76,17 +78,18 @@ impl Show for Packed<SubElem> {
 /// ```
 #[elem(title = "Superscript", Show)]
 pub struct SuperElem {
-    /// Whether to prefer the dedicated superscript characters of the font.
+    /// Whether to use the `sups` OpenType feature to render the glyphs.
     ///
-    /// If this is enabled, Typst first tries to transform the text to
-    /// superscript codepoints. If that fails, it falls back to rendering
-    /// raised and shrunk normal letters.
+    /// Do not use this is the body contains non-text elements.
```
PgBiel, line 83:
```suggestion
    /// Do not use this if the body contains non-text elements.
```

PgBiel (commented):
I'm also concerned with the font support deal, but leaving that discussion aside for a moment, I'll add some other comments.

crates/typst-library/src/text/shift.rs
```diff
@@ -125,90 +132,10 @@ impl Show for Packed<SuperElem> {
...
-        ' ' => Some(' '),
-        _ => None,
-    }
+fn is_text(content: &Content) -> bool {
+    content.is::<SpaceElem>()
+        || content.is::<TextElem>()
+        || content
+            .to_packed::<SequenceElem>()
+            .is_some_and(|sequence| sequence.children.iter().all(is_text))
 }
```
PgBiel, line 141: This is also too restrictive and would also trigger on metadata, horizontal spacing and so on. I believe this complements my other comment on the warning.

crates/typst-library/src/text/shift.rs
MDLC01: Because the baseline will not be raised, and therefore non-text elements will appear as if they were not in a `super`/`sub` element. I adjusted the documentation to explain that.

crates/typst-library/src/text/shift.rs
MDLC01: I think the real solution would be to just make it work as expected, but we can't just change the baseline, because we don't want to change it for text elements as this is already handled by the font. I'm unsure how to change the baseline only for non-text elements. Additionally, we would need to get the information of the new baseline from the font, which I'm unsure is possible.

crates/typst-library/src/text/shift.rs
MDLC01: On a side note, `h` elements might work slightly unexpectedly: since the font size does not change when `typographic: true`, `1em` in `super` or `sub` with `typographic: true` is the same as `1em` outside. This means `h(1em)` will appear larger with `typographic: true` than with `typographic: false`.

khaledhosny:
CSS [specification](https://drafts.csswg.org/css-fonts/#font-variant-position-prop) describe a fallback behavior that should help with inconsistent font coverage of these features. Quoting some of that text:
> Because of the semantic nature of subscripts and superscripts, when the value is either sub or super for a given contiguous run of text, if a variant glyph is not available for all the characters in the run, simulated glyphs should be synthesized for all characters using reduced forms of the glyphs that would be used without this feature applied.

crates/typst-library/src/text/shift.rs
khaledhosny: The OS/2 table has information on subscript/superscript position and scale, but unfortunately almost no application uses it, so fonts tend to hack junk values there (often auto generated by font editors).

MDLC01: This is annoying, compared to having no value at all...

MDLC01:
Just to keep you updated, I am now working on this again, and I hope to have a working solution in the upcoming days/weeks. Feel free to close the PR to prevent clutter.

laurmaedje:
If you plan to work on it in the next days or weeks, there is no need to close.

crates/typst-library/src/text/shift.rs
MDLC01: All the fonts bundled in Typst have sensible metrics for `y{Sub,Super}scriptYSize` and `y{Sub,Super}scriptYOffset` (including New Computer Modern Math, which is not really intended to be used as a text font), so at least things will work well by default.

Below is the result of a prototype that synthesizes superscripts and subscripts for each bundled font by rescaling the glyphs according to `y{Sub,Super}scriptYSize` and shifting them according to `y{Sub,Super}scriptYOffset`. The font order is: DejaVu Sans Mono (4 variants), Libertinus Serif (6 variants), New Computer Modern (4 variants), and New Computer Modern Math (3 variants).

Superscripts and subscripts appear in sensible places and in a sensible size.

However, there seems to be an error in the way `y{Sub,Super}scriptXOffset` (mainly used for italic correction) are encoded in DejaVu Sans Mono: in this font, the sign is inverted, meaning the offset is encoded as an offset to the left instead of as an offset to the right. The specification[^1][^2] isn't very precise, but the way I read it, it should be an offset to the right (or at least in the direction of the text direction): "from the glyph origin to the glyph origin of the subscript’s glyph". Below is what applying `y{Sub,Super}scriptXOffset` as an offset to the right looks like (I would recommend downloading the two images and switching between them to see the difference):

image

Note that this works as expected for Libertinus Serif and New Computer Modern.

[^1]: https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ysubscriptxoffset
[^2]: https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ysuperscriptxoffset

crates/typst-library/src/text/shift.rs
```diff
@@ -109,106 +120,88 @@ pub struct SuperElem {
...
+    body: Content,
+    typographic: bool,
+    baseline: Smart<Length>,
+    size: Smart<TextSize>,
+    kind: ScriptKind,
+) -> SourceResult<Content> {
     let world = engine.world;
-    for family in TextElem::font_in(styles) {
-        if let Some(font) = world
+    let text = body.plain_text();
```
MDLC01, line 145: I'm not sure this is the right way to get the inner text to test for font support.

tests/ref/super-underline.png
MDLC01, line 1: Ideally, `underline(super(..))` should have the underline based on the outer baseline, and `super(underline(..))` should be based on the script baseline, regardless of whether the scripts are synthesized or not.

It did not work like that before (in fact, I think this PR does not change the behavior in this regard), so this can also be left as a future improvement.

tests/ref/super-1em.png
MDLC01, line 1: For now, `1em` within a script resolves to the same absolute length as `1em` outside. Maybe this should be changed to account for the reduced script size.

tests/suite/text/shift.typ
```diff
@@ -8,12 +8,72 @@
...
+#table(
+  columns: 3,
+  [Typo.], [Fallb.], [Synth],
+  [x#super[1]], [x#super[5n]], [x#super[2 #box(square(size: 6pt))]],
+  [x#sub[1]], [x#sub[5n]], [x#sub[2 #box(square(size: 6pt))]],
+)
+
+--- sub-super-synthesized ---
 #set super(typographic: false, baseline: -0.25em, size: 0.7em)
 n#super[1], n#sub[2], ... n#super[N]
```
MDLC01, line 33: This test (previously named `sub-super-non-typographic`) has slightly changed for reasons I can't figure out.

MDLC01 (commented):
I pushed a prototype that uses the OpenType features by default, and synthesizes scripts when `typographic` is set to `false`.

When the OpenType features are used, non-textual elements are raised as well using the metrics provided by the font (although italic compensation, i.e., horizontal offset, is not applied to. When scripts are synthesized, the metrics from the fonts are used (for if they are available, and otherwise, there is a fallback to the metrics that were used before this PR.

This is still not fully ready to be merged. Below are some concerns and open questions.

Currently, the `typographic` argument is a boolean. When set to `false`, scripts are synthesized. When set to `true`, the corresponding OpenType feature is enabled, and that's it. There is no check that the font does in fact support the feature, and that it has subs/sups glyphs for all the characters used. I don't think there is an easy way to do that, other than manually comparing that all glyphs are different. What's bad is that the default text font, Libertinus Serif, does not have many sups/subs glyphs, so users are likely to encounter an issue if we set `typographic: true` by default. This is similar to the way `smallcaps` currently works, except Libertinus Serif has better support for `smcp` and `c2sc` than for `subs` and `sups`.

Finally, fonts can provide a different scaling for both axes for scripts (typically, the glyphs are scaled down vertically more than horizontally). When synthesizing scripts and when scaling non-textual elements, we scale them evenly in both axes, according to the font's vertical scaling factor. Should we use the horizontal scaling factor instead? Or should we use both?

tests/ref/super-highlight.png
MDLC01, line 1: Similar issue here as with underlines: https://github.com/typst/typst/pull/5777/#discussion_r2016787652.

laurmaedje:
>  There is no check that the font does in fact support the feature, and that it has subs/sups glyphs for all the characters used. I don't think there is an easy way to do that, other than manually comparing that all glyphs are different.

For a basic check that the feature is supported, you could probably check for the [GSUB table](https://docs.rs/ttf-parser/latest/ttf_parser/struct.FaceTables.html#structfield.gsub), iterate its [features](https://docs.rs/ttf-parser/latest/ttf_parser/opentype_layout/struct.LayoutTable.html#structfield.features) and check that it has matches for the `subs` and `sups` [Tags](https://docs.rs/ttf-parser/latest/ttf_parser/struct.Tag.html). I haven't tested it, but worth a try.

crates/typst-kit/src/package.rs
laurmaedje, line 1: These changes seem to come from a bad merge.

tests/ref/super-underline.png
laurmaedje: Yeah underlines are due for a larger rework.

crates/typst-library/src/text/mod.rs
```diff
@@ -752,6 +752,11 @@ pub struct TextElem {
     #[internal]
     #[ghost]
     pub smallcaps: Option<Smallcaps>,
+
+    /// Whether subscripts or superscripts are enabled.
+    #[internal]
+    #[ghost]
+    pub subpercript: Option<ScriptSettings>,
```
laurmaedje: the "bp" is maybe intentional, but an "s" seems to be missing either way before "cript"

khaledhosny:
> > There is no check that the font does in fact support the feature, and that it has subs/sups glyphs for all the characters used. I don't think there is an easy way to do that, other than manually comparing that all glyphs are different.
> 
> For a basic check that the feature is supported, you could probably check for the [GSUB table](https://docs.rs/ttf-parser/latest/ttf_parser/struct.FaceTables.html#structfield.gsub), check thats its [features](https://docs.rs/ttf-parser/latest/ttf_parser/opentype_layout/struct.LayoutTable.html#structfield.features) [have matches](https://docs.rs/ttf-parser/latest/ttf_parser/opentype_layout/struct.RecordList.html#tymethod.find) for the `subs` and `sups` [Tags](https://docs.rs/ttf-parser/latest/ttf_parser/struct.Tag.html). I haven't tested it, but worth a try.

HarfBuzz has [`hb_ot_layout_lookup_collect_glyphs()`](https://harfbuzz.github.io/harfbuzz-hb-ot-layout.html#hb-ot-layout-lookup-collect-glyphs) which can be used to get the glyphs covered by a given lookup (and other APIs to get lookups for a feature). I don't know if rustybuzz has equivalent APIs.

MDLC01:
Thanks for the pointer! I will look into that.

MDLC01:
I now enable the feature only if it is supported (in the sense that all glyphs required to display a specific script are available). I still want to add more tests before merging to make sure everything works as intended, but it is pretty much in the final state for now.

The commits are a bit all over the place if you review the PR from the beginning. Tell me if you would like me to clean all that.

crates/typst-library/src/text/shift.rs
MDLC01: We no longer do that

MDLC01:
I cleaned up the commits and added some tests.

There is still the question of whether `1em` within scripts should resolve to the same absolute size as outside the scripts or not (https://github.com/typst/typst/pull/5777#discussion_r2016789648). For now, `1em` within scripts is the same as `1em` outside of scripts, as illustrated by the `super-1em` test. If we decide to keep this behavior, this is ready to be merged!

Additionally, the behavior of `underline` and `highlight` together with `super` and `sub` could be improved. I think this can be left as a future improvement, as the current behavior is already imperfect, and comparable to the behavior with this PR (https://github.com/typst/typst/pull/5777#discussion_r2016787652).

tests/ref/sub-super-typographic.png
MDLC01, line 1: The squares aren't perfectly aligned to the superscript/subscript baseline in this test. I believe this to be due to imperfect metrics in Libertinus Serif.

MDLC01: Seems to be https://github.com/alerque/libertinus/issues/511.

tests/suite/text/shift.typ
```diff
@@ -1,19 +1,78 @@
...
 
 --- sub-super ---
+#let sq = box(square(size: 4pt))
 #table(
   columns: 3,
-  [Typo.], [Fallb.], [Synth],
-  [x#super[1]], [x#super[5n]], [x#super[2 #box(square(size: 6pt))]],
-  [x#sub[1]], [x#sub[5n]], [x#sub[2 #box(square(size: 6pt))]],
+  [Typo.], [Fallb.], [Synth.],
+  [x#super[1 #sq]], [x#super[5: #sq]], [x#super(typographic: false)[2 #sq]],
```
laurmaedje: If I zoom in on this test, the numbers in the "Typo" column are not actually typographic ones. (Due to the space I think).

tests/suite/text/shift.typ
```diff
@@ -1,19 +1,78 @@
...
+  [x#sub[1 #sq]], [x#sub[5: #sq]], [x#sub(typographic: false)[2 #sq]],
 )
 
+--- sub-super-typographic ---
+#set text(size: 20pt)
+// Libertinus Serif supports "subs" and "sups" for `typo`, but not for `synth`.
```
laurmaedje:
```suggestion
// Libertinus Serif supports "subs" and "sups" for `typo` and `sq`,
// but not for `synth`.
```

tests/suite/text/shift.typ
```diff
@@ -1,19 +1,78 @@
...
+
+--- super-highlight ---
+#set super(typographic: false)
+#highlight[The claim#super[4]] has been disputed. \
+The claim#super[#highlight[4]] has been disputed. \
+It really has been#super(highlight[4]) \
+#set super(typographic: true)
+#highlight[The claim#super[4]] has been disputed. \
+The claim#super[#highlight[4]] has been disputed. \
+It really has been#super(highlight[4])
```
laurmaedje: The "has been disputed" leads to linebreaks which make the tests both harder to read and the images unnecessarily large. I think for these tests something like `#underline[A#super[4]] B` would suffice.

tests/suite/text/shift.typ
```diff
@@ -1,19 +1,78 @@
...
+
+--- super-1em ---
+Test#super[#box(rect(height: 1em))]#box(rect(height: 1em))
```
laurmaedje: This could use an assertion instead, e.g. `#super(context test(1em.to-absolute(), 10pt))`. A bit unfortunate that it still has an image then, ideally we'd have a test flag to skip the image, but that's out of scope here.

tests/suite/text/shift.typ
```diff
@@ -1,19 +1,78 @@
...
+
+--- long-scripts ---
+|longscript| \
+|#super(typographic: true)[longscript]| \
+|#super(typographic: false)[longscript]| \
+|#sub(typographic: true)[longscript]| \
+|#sub(typographic: false)[longscript]|
+
+--- scripts-with-bundeled-fonts ---
+#let test(font, weights, styles) = {
```
laurmaedje, line 74: IIUC, the point of this tests is to test the metrics of these fonts for synthesis? If yes, that'd be worth a comment, as I first thought the point is to test the sups/subs features.

tests/suite/text/shift.typ
```diff
@@ -1,19 +1,78 @@
...
+Test#super[#box(rect(height: 1em))]#box(rect(height: 1em))
+
+--- long-scripts ---
+|longscript| \
+|#super(typographic: true)[longscript]| \
+|#super(typographic: false)[longscript]| \
+|#sub(typographic: true)[longscript]| \
+|#sub(typographic: false)[longscript]|
+
+--- scripts-with-bundeled-fonts ---
```
laurmaedje:
```suggestion
--- script-metrics-bundled-fonts ---
```

crates/typst-library/src/text/shift.rs
```diff
@@ -108,107 +120,55 @@ pub struct SuperElem {
...
-        'm' => Some('ₘ'),
-        'n' => Some('ₙ'),
-        'p' => Some('ₚ'),
-        's' => Some('ₛ'),
-        't' => Some('ₜ'),
-        ' ' => Some(' '),
-        _ => None,
-    }
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct ScriptSettings {
```
laurmaedje: could you move this above ScriptKind and add brief docs to the methods and fields defined around here?

crates/typst-library/src/text/shift.rs
```diff
@@ -17,11 +16,17 @@ use crate::World;
 /// ```
 #[elem(title = "Subscript", Show)]
 pub struct SubElem {
-    /// Whether to prefer the dedicated subscript characters of the font.
+    /// Whether to create artificial subscripts by shifting and scaling down
+    /// regular glyphs.
```
laurmaedje: This sounds like it's the wrong way around.

crates/typst-library/src/text/shift.rs
```diff
@@ -17,11 +16,17 @@ use crate::World;
...
+    ///
+    /// Ideally, subscripts glyphs should be provided by the font (using the
+    /// `subs` OpenType feature). Otherwise, Typst is able to synthesize
+    /// subscripts as explained above.
```
laurmaedje:
```suggestion
    /// Ideally, subscripts glyphs are provided by the font (using the
    /// `subs` OpenType feature). Otherwise, Typst is able to synthesize
    /// subscripts.
```

crates/typst-library/src/text/shift.rs
```diff
@@ -30,17 +35,19 @@ pub struct SubElem {
...
-    /// The baseline shift for synthetic subscripts. Does not apply if
-    /// `typographic` is true and the font has subscript codepoints for the
-    /// given `body`.
-    #[default(Em::new(0.2).into())]
-    pub baseline: Length,
+    /// The baseline shift for synthesized subscripts.
+    ///
+    /// This only applies to synthesized subscripts. In other words, this has no
+    /// effect if `typographic` is `{true}` and the font provides the necessary
+    /// subscript glyphs.
```
laurmaedje, line 44: two things that would be worth mentioning
- in which direction the baseline shift is applied
- what `auto` means

crates/typst-library/src/text/shift.rs
```diff
@@ -30,17 +35,19 @@ pub struct SubElem {
...
-    /// `typographic` is true and the font has subscript codepoints for the
-    /// given `body`.
-    #[default(TextSize(Em::new(0.6).into()))]
-    pub size: TextSize,
+    /// The font size for synthesized subscripts.
+    ///
+    /// This only applies to synthesized subscripts. In other words, this has no
+    /// effect if `typographic` is `{true}` and the font provides the necessary
+    /// subscript glyphs.
+    pub size: Smart<TextSize>,
```
laurmaedje, line 60: same here with auto

crates/typst-library/src/text/shift.rs
```diff
@@ -76,11 +80,17 @@ impl Show for Packed<SubElem> {
 /// ```
 #[elem(title = "Superscript", Show)]
 pub struct SuperElem {
-    /// Whether to prefer the dedicated superscript characters of the font.
```
laurmaedje, line 89: same here

crates/typst-library/src/text/shift.rs
```diff
@@ -108,107 +120,55 @@ pub struct SuperElem {
...
-                    return Ok(TextElem::packed(text));
-                }
-            }
-        };
-
-        Ok(body
-            .styled(TextElem::set_baseline(self.baseline(styles)))
-            .styled(TextElem::set_size(self.size(styles))))
+    fn show(&self, _: &mut Engine, styles: StyleChain) -> SourceResult<Content> {
+        let outer_text_size = TextElem::size_in(styles);
```
laurmaedje: might be nice to extract the show rule body of those two into a common function

crates/typst-library/src/text/mod.rs
```diff
@@ -754,6 +754,11 @@ pub struct TextElem {
     #[internal]
     #[ghost]
     pub smallcaps: Option<Smallcaps>,
+
+    /// Whether subscripts or superscripts are enabled.
+    #[internal]
+    #[ghost]
+    pub subperscript: Option<ScriptSettings>,
```
laurmaedje: this always reads like a typo to me. how about just `script_settings`?

crates/typst-library/src/text/mod.rs
```diff
@@ -754,6 +754,11 @@ pub struct TextElem {
     #[internal]
     #[ghost]
     pub smallcaps: Option<Smallcaps>,
+
+    /// Whether subscripts or superscripts are enabled.
```
laurmaedje: the comment makes it sound like a bool

crates/typst-library/src/text/font/mod.rs
```diff
@@ -283,6 +310,34 @@ pub struct LineMetrics {
     pub thickness: Em,
 }
 
+/// Metrics for subscripts or superscripts.
+#[derive(Debug, Copy, Clone)]
+pub struct ScriptMetrics {
+    pub width: Em,
```
laurmaedje, line 324: please also add brief docs for the fields and methods here

crates/typst-library/src/text/font/mod.rs
```diff
@@ -215,6 +215,10 @@ pub struct FontMetrics {
     pub underline: LineMetrics,
     /// Recommended metrics for an overline.
     pub overline: LineMetrics,
+    /// Metrics for subscripts.
+    pub subscript: ScriptMetrics,
```
laurmaedje: I think it would be cleaner to have these as options instead of just setting them to a default and handling the actual defaults in shift.rs rather than on `ScriptMetrics`.

crates/typst-library/src/text/font/mod.rs
```diff
@@ -283,6 +310,34 @@ pub struct LineMetrics {
...
+    }
+
+    pub const fn default_subscript() -> Self {
+        Self::default_with_vertical_offset(Em::new(-0.2))
+    }
+
+    pub const fn default_superscript() -> Self {
+        Self::default_with_vertical_offset(Em::new(0.5))
+    }
+}
```
laurmaedje: this would then all be unified with the methods on `ScriptKind`. I think it's also cleaner to not have the defaults live with the element rather than the font parsing code.

crates/typst-library/src/text/shift.rs
```diff
@@ -108,107 +120,55 @@ pub struct SuperElem {
...
+    pub const fn default_metrics(self) -> ScriptMetrics {
+        match self {
+            Self::Sub => ScriptMetrics::default_subscript(),
+            Self::Super => ScriptMetrics::default_superscript(),
         }
     }
 
-    false
-}
+    pub const fn read_metrics(self, font_metrics: &FontMetrics) -> ScriptMetrics {
```
laurmaedje: we could return `&ScriptMetrics` here and for the defaults have them in `static`s and return references to those.

crates/typst-library/src/layout/length.rs
```diff
@@ -91,6 +91,10 @@ impl Length {
             hint: "or use `length.abs.{unit}()` instead to ignore its em component"
         )
     }
+
+    pub fn to_em(&self, text_size: Abs) -> Em {
```
laurmaedje: please move this above `ensure_...` and add a brief doc comment

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -789,6 +803,52 @@ fn shape_segment<'a>(
...
+                    .get(subtable_index)?
+                    .subtables
+                    .get::<SubstitutionSubtable>(0)?
+                    .coverage();
+                text.chars()
+                    .all(|c| {
+                        font.rusty().glyph_index(c).is_some_and(|i| coverage.contains(i))
+                    })
+                    .then(|| {
+                        ctx.features.push(Feature::new(settings.kind.feature(), 1, ..));
```
laurmaedje: this is dangerous as the context is reused across multiple invocations of shape_segment. maybe the logic can be applied for the whole `shape` run instead of in `shape_segment`? shape_segment is separate for font fallback and I don't think how fonts fall back should affect the super/sub-scripts.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -789,6 +803,52 @@ fn shape_segment<'a>(
     // text extraction.
     buffer.set_flags(BufferFlags::REMOVE_DEFAULT_IGNORABLES);
 
+    // Those values determine how the rendered text should be transformed to
+    // display sub-/super-scripts properly. If the text is not scripted, or if
+    // the OpenType feature can be used, the rendered text should not be
+    // transformed in any way, and so those values are neutral (`(0, 0, 1)`). If
+    // scripts should be synthesized, those values determine how to transform
+    // the rendered text to display scripts as expected.
+    let (script_shift, script_compensation, scale) = match script_settings {
```
laurmaedje: could you try to extract this logic into a function. shape_segment is already very long.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -723,6 +736,7 @@ fn shape_segment<'a>(
     base: usize,
     text: &str,
     mut families: impl Iterator<Item = &'a FontFamily> + Clone,
+    script_settings: Option<ScriptSettings>,
```
laurmaedje: could these be part of the context?

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -64,6 +65,8 @@ pub struct ShapedGlyph {
     pub x_offset: Em,
     /// The vertical offset of the glyph.
     pub y_offset: Em,
+    /// How much to scale the glyph compared to its normal size.
+    pub scale: Em,
```
laurmaedje: it's a bit unfortunate to have both this and `size`. couldn't we instead move `size` here?

crates/typst-kit/src/package.rs
MDLC01: I'm not sure what this is referring to, and I don't see any change to that file in GitHub's "Files changed" tab.

crates/typst-library/src/text/mod.rs
MDLC01: Yeah the "bp" is intentional. If you have a better name, that would be great. `script` is already taken for, well, [`text.script`](https://typst.app/docs/reference/text/text/#parameters-script).

The missing "s" is a mistake on my part though, and I think it is fixed in the most recent commit.

tests/suite/text/shift.typ
MDLC01: Done

tests/suite/text/shift.typ
MDLC01: This was indeed due to the space

tests/suite/text/shift.typ
MDLC01: Done

tests/suite/text/shift.typ
MDLC01: Actually, the reference image was removed when updating the tests!

crates/typst-kit/src/package.rs
laurmaedje: The comment was from March. Probably already fixed.

crates/typst-library/src/text/mod.rs
laurmaedje: That message was also old. I think I had a few pending review comments from an initial review attempt and these now showed up.

tests/suite/text/shift.typ
MDLC01: I don't remember why I added this test, but it is indeed useful to make sure we use the font-provided metrics in the right way (e.g., no sign error). Maybe I could make the test shorter by not testing all fonts, since we only need to make sure we use the metrics properly once. From what I remember, I simply copied the code from https://github.com/typst/typst/pull/5777#discussion_r2003568010 because I figured it could be a useful test.

I added a comment and also set `typographic: false` for both `super` and `sub`, in order to make sure we actually synthesize the scripts. The reference image does not change because none of the bundled fonts provide script glyphs for X and x.

tests/suite/text/shift.typ
MDLC01: Done

crates/typst-library/src/text/shift.rs
MDLC01: Done

crates/typst-library/src/text/font/mod.rs
MDLC01: Done

crates/typst-library/src/text/shift.rs
MDLC01: I meant "shifting _(implicitly down)_ and scaling down", but I agree this sounds weird.

I changed it to "lowering and ..." for subscripts and "raising and ..." for superscripts.

crates/typst-library/src/text/shift.rs
MDLC01: Done

crates/typst-library/src/text/shift.rs
MDLC01: I added:

> If set to `{auto}`, the baseline is shifted according to the metrics provided by the font.

Should I also mention the default values in case the font does not provide said metrics?

Regarding the direction the baseline shift is applied, this is not the only place where this problem occurs in Typst documentation (e.g., [`box.baseline`](https://typst.app/docs/reference/layout/box/#parameters-baseline) has the same issue), but I will only change it here for now.

For both superscripts and subscripts, the baseline shift applies downward. This means, for superscripts, users need to provide a negative value. This is consistent with `sub` and `box`.

crates/typst-library/src/text/shift.rs
MDLC01: Done

crates/typst-library/src/text/shift.rs
MDLC01: Done

crates/typst-library/src/text/mod.rs
MDLC01: The problem is that this creates an ambiguity between super- and subscripts, and `text.script`. But I can change it if you want.

crates/typst-library/src/text/shift.rs
MDLC01: Done

crates/typst-library/src/text/mod.rs
MDLC01: Fixed

crates/typst-library/src/text/font/mod.rs
MDLC01: I mirrored the way this is done for line metrics. Should I change them as well?

crates/typst-library/src/layout/length.rs
MDLC01: Done

MDLC01:
I will address the remaining comments later, as they require more thinking.

crates/typst-library/src/text/mod.rs
laurmaedje: what about ShiftSettings? since the file is already called shift.rs.

MDLC01: Good idea

crates/typst-library/src/text/font/mod.rs
MDLC01: I changed it for `subscript` and `superscript` only.

crates/typst-library/src/text/font/mod.rs
MDLC01: Done

crates/typst-layout/src/inline/shaping.rs
MDLC01: I extracted this into a `compute_synthesized_shift` function. Let me know if you find a better name.

crates/typst-library/src/text/shift.rs
MDLC01: I changed that.

crates/typst-layout/src/inline/shaping.rs
MDLC01: This is indeed better

crates/typst-layout/src/inline/shaping.rs
MDLC01: I'm not sure I fully understand what you mean here. But font fallback does affect scripts in the sense that we need to know the font that is chosen in order to decide whether to synthesize scripts. This is now more clearly visible because `compute_synthesized_shift` accepts a `&Font`.

crates/typst-layout/src/inline/shaping.rs
MDLC01: I will wait for an answer to https://github.com/typst/typst/pull/5777#discussion_r2160401965 before addressing that because moving part of the logic to `shape` could make this easier.

MDLC01:
I have now implemented or asked for clarification on the remaining suggestions.

crates/typst-library/src/text/shift.rs
laurmaedje: > Should I also mention the default values in case the font does not provide said metrics?

Yeah, I think that would be good.

crates/typst-library/src/text/font/mod.rs
laurmaedje: We can take a look at changing line metrics separately.

crates/typst-layout/src/inline/shaping.rs
laurmaedje: What I mean is that pushing into `ctx.features` is a side effect, so if `shape_segment` is called multiple times during one `shape` operation (due to font fallback), we will push the feature multiple times and potentially affect future segments in unplanned ways.

crates/typst-layout/src/inline/shaping.rs
laurmaedje: The name sounds fine to me. I would suggest pulling the match on the option out of the function and keeping the logic in the function for the case where we do have shift settings.

laurmaedje:
I responded to or marked the comments as resolved. Probably you have already seen, but if not, note that there are also still conflicts on the PR.

crates/typst-library/src/text/shift.rs
MDLC01: Done

crates/typst-layout/src/inline/shaping.rs
MDLC01: Done

crates/typst-layout/src/inline/shaping.rs
MDLC01: I'm still unsure why font fallback shouldn't affect scripts. For example, `text.cover` should affect shifts right? For example, if I'm writing a document in two scripts and I use different fonts for those, I don't want shifts to be affected only by whichever font comes first, regardless of its `cover`.

laurmaedje: Maybe I'm misunderstanding what's happening here, but all I'm saying is that you are pushing into `ctx.features` for each individual shape segment. Let's say you have three segments (because we have font fallback twice). Once you've arrived at the third segment, `ctx.features` will contain the feature thrice, which doesn't seem right.

MDLC01: Would something like this solve the problem?

```rs
let (script_shift, script_compensation, scale, shift_feature /* : Option<Feature> */) = ctx
    .shift_settings
    .map_or((Em::zero(), Em::zero(), Em::one(), None), |settings| {
        compute_synthesized_shift(text, &font, settings)
    });

let features = ctx.features.iter().cloned().chain(shift_feature).collect::<Vec<_>>();

let plan = create_shape_plan(
    &font,
    buffer.direction(),
    buffer.script(),
    buffer.language().as_ref(),
    &features,
);
```

where `compute_synthesized_shift` returns the optional feature instead of adding it to `ctx.features`.

The logic is not "applied for the whole `shape` run" as you suggested above, but this seems hard to do because I need to know which font is chosen to call `compute_synthesized_shift`. Unless what you meant by "I don't think how fonts fall back should affect the super/sub-scripts" is that I should just use the first font in the fallback list that has the right `cover` in `shape`?

laurmaedje: Yeah that should fix it. Alternatively, we could avoid the allocation by pushing the feature (but outside of the compute_synthesized_shift function) and then popping it again after the shape plan is created.

MDLC01: Done

crates/typst-layout/src/inline/shaping.rs
MDLC01: I moved `size` to `ShapedGlyph`.

In some places, we still want the text size (`TextElem::size_in(shaped_text.styles)`) instead of the size of a specific glyph. I'm not sure if I made the right decision everywhere, so please be careful when reviewing the changes. My process was to use the text size only when necessary to make the tests pass.

This probably just made more visible an existing problem with the previous implementation, where I wouldn't multiply `shaped_text.size` with `glyph.scale` in most places. Now that I read from `glyph.size` instead of `TextElem::size_in(shaped_text.styles)` in most places, this corresponds to multiplying `shaped_text.size` with `glyph.scale`  in most places. Without tests to see the difference, and especially when it involves CJK typography, it's hard for me to know which one I should use in each situation.

MDLC01:
This is now ready for review. As noted in https://github.com/typst/typst/pull/5777#discussion_r2173765801, I'm unsure in which places `TextElem::size_in(shaped_text.styles)` vs. `glyph.size` should be used.

crates/typst-layout/src/inline/shaping.rs
MDLC01: I went through all uses of `glyph.size`, and the current way I use it seems reasonable

crates/typst-library/src/layout/length.rs
```diff
@@ -76,6 +76,11 @@ impl Length {
         self.abs + self.em.at(font_size)
     }
 
+    /// Expresses this length in em's of the provided outer text size.
+    pub fn to_em(&self, font_size: Abs) -> Em {
+        self.em + Em::from_length(self.abs, font_size)
+    }
```
MDLC01: Should I move this to `Em` as `Em::from_length` and rename the current `Em::from_length` to `Em::from_abs` instead?

laurmaedje: That sounds sensible.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -932,6 +961,65 @@ fn shape_segment<'a>(
     ctx.used.pop();
 }
 
+/// Returns a `(script_shift, script_compensation, scale)` triplet describing
```
laurmaedje: The doc comment is outdated. The function name doesn't seem 100% correct anymore since it also produces the feature. How about `determine_shift`? Or splitting it into two functions `compute_synthesized_shift(font: &Font, settings: ShiftSettings) -> (Em, Em, Em)` and `shift_feature(text: &str, font: &Font, settings: ShiftSettings) -> Option<Feature>` and then tying it together above.

crates/typst-layout/src/inline/shaping.rs
laurmaedje: I went through them too and it does seem reasonable

crates/typst-library/src/layout/length.rs
MDLC01: Done

crates/typst-layout/src/inline/shaping.rs
MDLC01: Indeed, my bad. `determine_shift` is good.

laurmaedje:
Thanks a lot for your work and patience with this!

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -932,6 +961,64 @@ fn shape_segment<'a>(
...
+    settings
+        .typographic
+        .then(|| {
+            // If typographic scripts are enabled (i.e., we want to use the
+            // OpenType feature instead of synthesizing if possible), we add
+            // "subs"/"sups" to the feature list if supported by the font.
+            // In case of a problem, we just early exit
+            let gsub = font.rusty().tables().gsub?;
+            let subtable_index =
+                gsub.features.find(settings.kind.feature())?.lookup_indices.get(0)?;
```
khaledhosny, line 986: If I’m reading this correctly, this code collects the glyph coverage from the first lookup in the feature, but the feature can have multiple lookups. This might be the cause for #7249.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -932,6 +961,64 @@ fn shape_segment<'a>(
...
+            let coverage = gsub
+                .lookups
+                .get(subtable_index)?
+                .subtables
+                .get::<SubstitutionSubtable>(0)?
```
khaledhosny, line 991: This also gets the coverage from the first subtable in the lookup, but lookups can have more than one subtable.

crates/typst-layout/src/inline/shaping.rs
MDLC01: So if I understand correctly, I should instead consider the union of the coverages from all subtables from all lookups to test whether there exists a scripted glyph?

khaledhosny: I believe so.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -932,6 +961,64 @@ fn shape_segment<'a>(
...
+                .coverage();
+            text.chars()
+                .all(|c| {
+                    font.rusty().glyph_index(c).is_some_and(|i| coverage.contains(i))
+                })
```
khaledhosny, line 996: This check is problematic as well. It gets glyph indices from cmap then it checks of they are covered by the feature, but when shaping text you might end up with different glyphs (e.g. a ligature or small caps or an alternate glyph) and it these glyphs might not be supported by the features compared to their base glyphs. Ideally, this function should take as input the already shaped glyph indices not the Unicode text.

crates/typst-layout/src/inline/shaping.rs
khaledhosny: The code should also take script and language into account, different OpenType language systems might have different feature records with different lookups and coverage.

Issue #3313, opened 2024-02-01
Enable Wasm plugin initialization or Wasm plugin state transistion.


Myriad-Dreamin:
### Description

Some plugin needs initialization, otherwise they feel painful on becoming stateless:
+ `jogs` plugin has to retrieve a `bytecode` (some fixed js library) argument per call.
+ `mitex` plugin has to retrieve a list of TeX commands/packages data per call.
+ `typst.wasm` plugin has to retrieve user-specified font data read from filesystem and the fonts often very large, in scale of MB.

From observation, the initialization takes up many CPU time. And, they are initialized once and serves stateless call multiple times. Therefore, it should be great if we provide a functionality to initialize them.

## Possible Design 1:

Extend current protocol with an additional `init` function so that we can initialize it on creating an instance.

```typ
#let mathjax-js-with-fonts = plugin("quickjs.wasm", init: (
  ("load_bytecode", mathjax-bytecode),
  ("call", "mathjaxLoadMathFonts", math-fonts),
))
#let other-js = plugin("quickjs.wasm", init: (
  ("load_bytecode", other-bytecode),
))
#let typst-with-fonts = plugin("typst.wasm", init: fonts)
```

## Possible Design 2:

Fork/Clone plugin into a sub plugin by calling some plugin exported function.

```typ
#let quickjs-engine = plugin("quickjs.wasm")
#let (mathjax-js, _) = plugin.transist(quickjs-engine, "load_bytecode", mathjax-bytecode)
#let (mathjax-js-with-fonts, _) = plugin.transist(mathjax-js, "call", "mathjaxLoadMathFonts", math-fonts)
#let (other-js, _) = plugin.transist(quickjs-engine, "load_bytecode", other-bytecode)
#let typst = plugin("typst.wasm")
#let typst-with-fonts = plugin.transist(typst, "init_compiler", fonts)
```

## How will plugins use the API?

`quickjs.wasm` initializes a new plugin instance with some JS library:

```typ
#let quickjs-engine = plugin("quickjs.wasm")
#let mathjax-js = plugin.transist(quickjs-engine, "load_bytecode", mathjax-bytecode)
```

`typst.wasm` loads customized huge set of fonts from filesystem:

```typ
#let typst = plugin("typst.wasm")
#let fonts = load-from-filesystem()
#let typst = plugin.transist(typst, "init_compiler", fonts)
```

Both of these two usage will still keep sane number of instances in the comemo cache.

[...]

PR #5779, opened 2025-01-29
Modular, multi-threaded, transitioning plugins

laurmaedje:
This PR makes a number of changes to the plugin system.

### Modular
The `plugin` function now returns a normal `module` containing Typst functions instead of a dedicated `plugin` type. As a result, plugin functions interact much better with the rest of Typst: There is no special method call handling for plugins anymore, plugin function can now be accessed and stored without being called, you can use `.with` on then and you can `import` from plugins.

```typ
#import plugin("/assets/plugins/hello.wasm"): hello, double_it
```

### Multi-threaded
Plugins can now run multi-threaded. Typst will automatically do this behind the scenes with no involvement of the plugin. Plugins needn't be thread-safe. Rather, Typst will spin up multiple instance of the plugin as needed. For documents relying on expensive plugin calls (like diagramming) this can provide a big speed-up.

Flamegraph of multi-threaded plugin execution

### Transitioning
A new `plugin.transition` API is introduced which allows plugins to run impure initialization (closes https://github.com/typst/typst/issues/3313) in a way that doesn't break Typst's purity guarantees.

This API comes alongside multi-threading for a simple reason: They are a few existing plugins which (wrongfully :)) rely on side effects being observable. With multiple WebAssembly instances being spun up due to multi-threading, these plugins will break.

Instead of calling their mutable functions directly, such plugins should use the transition API. This API runs a mutable function call and then returns a **new, derived** module whose functions will observe the side effects.

Here's an example: We call `add` via the transition API. The call
`mutated.get()` on the derived module will observe the addition. Meanwhile, the
original module remains untouched as demonstrated by the `base.get()` call.

```typ
#let base = plugin("hello-mut.wasm")
#assert.eq(base.get(), "[]")

#let mutated = plugin.transition(base.add, "hello")
#assert.eq(base.get(), "[]")
#assert.eq(mutated.get(), "[hello]")
```

After transitioning a plugin, Typst will take a memory snapshot and restore that snapshot for any fresh WebAssembly instance created due to multi-threading.

Calling impure plugin functions without the transition API will still yield unpredictable results. I've considered enforcing purity by resetting the memory after every individual call, but I think the cost of doing this is prohibitive for plugins with many cheap calls. An alternative would be resetting the memory after a larger number of wasm "fuel" steps instead of after every call. That would expose a few more bad mutable calls, nudging plugin developers into the right direction. However, multi-threading by itself will also do that to an extent, so I'm not yet sure where it's really worth it, so it's left as potential future work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2766,7 +2766,7 @@ dependencies = [
 [[package]]
 name = "typst-dev-assets"
 version = "0.12.0"
-source = "git+https://github.com/typst/typst-dev-assets?rev=b07d156#b07d1560143d6883887358d30edb25cb12fcf5b9"
+source = "git+https://github.com/typst/typst-dev-assets?rev=7f8999d#7f8999d19907cd6e1148b295efbc844921c0761c"
 
 [[package]]
 name = "typst-docs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ typst-syntax = { path = "crates/typst-syntax", version = "0.12.0" }
 typst-timing = { path = "crates/typst-timing", version = "0.12.0" }
 typst-utils = { path = "crates/typst-utils", version = "0.12.0" }
 typst-assets = { git = "https://github.com/typst/typst-assets", rev = "8cccef9" }
-typst-dev-assets = { git = "https://github.com/typst/typst-dev-assets", rev = "b07d156" }
+typst-dev-assets = { git = "https://github.com/typst/typst-dev-assets", rev = "7f8999d" }
 arrayvec = "0.7.4"
 az = "1.2"
 base64 = "0.22"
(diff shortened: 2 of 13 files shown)


lublak:
@laurmaedje because it's a breaking change (which is fine). Would be a good idea to implement an additional compatible with field in the typst.toml.
`compatible_with = ">=0.13"`
`compatible_with = "0.13,0.14"`.
Jus for the feature :)

lublak:
@laurmaedje Oh, I had overlooked that. Of course the `compiler` field is sufficient :smile:

laurmaedje:
I looked into it and for correctness, technically, we would also need to snapshot WASM globals, not just the memory, in `plugin.transition`. However, wasmi does not seem to expose an API to access non-exported globals. I opened [a feature request on the wasmtime repository](https://github.com/bytecodealliance/wasmtime/issues/10156). (Wasmi mostly seems to mirror wasmtime's API so proposing it there seemed more fruitful.)

In the meantime, I would proceed with simply not snapshotting the globals. LLVM-based compilers apparently only emit a single global for a shadow stack pointer, which _should_ always be the same after the call stack is drained. (At least I hope so.) And pretty much all plugins are written with LLVM-based compilers.

I added a disclaimer to the transition docs.

PR #5784, opened 2025-01-30
Improve clarity of `ast.rs` for newcomers to the codebase

wrzian:
This is a set of small changes to the `typst-syntax/src/ast.rs` file designed to improve clarity of the code for newcomers. I personally had a lot of trouble understanding this file the first time I encountered it (many months ago), and wanted to document its behavior and make some improvements while it's still top-of-mind.

1. Add big comment and require the `struct` keyword in the `node!` macro (e.g. `node! {Raw}` vs. `node! {struct Raw}`). I think my original confusion with `ast.rs` was with where all of the data-types were declared. The way enums were laid out combined with `node!` using only the SyntaxKind name was unhelpful. Adding `struct` makes it feel much clearer to me that `node!` is generating new types. And the added comment on `node!` should also be a nice improvement.

2. Removed comments on the enum variants of `Expr`  (the structs they contain are already commented) and renamed the variants that didn't match their SyntaxKind names.

3. Changed the calls to `node.cast().map()` in `from_untyped` for the various implementations to construct the node structs more directly. The current `node.cast().map()` requires type inference that may be obvious to the compiler, but is not obvious to humans on first reading. It does reduce the dependency on the name of each node kind, but I think the current syntax obscures the real operation more than is worth avoiding one extra repeated name in each line. There are still calls to `node.cast().map()` in the `from_untyped` impl, but only in wildcard matches, where it should be easier to figure out what the desired effect is.

I'm very open to nitpicks or rewordings of the main long doc comment in commit 1. But I think it does roughly what it needs to atm.

Also fine with backing out any of these changes or maybe putting the module comment somewhere else. But my main goal is to improve clarity for newcomers to the codebase, and I think these changes should help with that.


crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+will use a running example of the [`Raw`] node type, which is declared (after
+macro expansion) as: `struct Raw<'a>(&'a SyntaxNode);`.
+
+[`SyntaxNode`]s are generated by the parser and constitute the Concrete Syntax
+Tree (CST). The CST is _concrete_ because it has the property that an in-order
+tree traversal will recreate the text of the source file exactly.
+
+[`SyntaxNode`]s in the CST contain their [`SyntaxKind`], but don't themselves
+provide access to the semantic meaning of their contents. That semantic meaning
+is available throught the Abstract Syntax Tree by iterating over CST nodes and
```
T0mstone, line 19:
```suggestion
is available through the Abstract Syntax Tree by iterating over CST nodes and
Comment
```

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+file as edges on a graph: areas where the tree has paths from one type to
+another (accessed through methods), then structs are the nodes of the graph,
+providing methods that return enums, etc. etc.
+
+---
+
+So that's the Typst AST interface!
+
+Feel free to ask questions in the contributors chat if anything isn't fully
+clear, or give us suggestions on ways to improve these docs!
```
T0mstone, line 79: This seems like unnecessary fluff.
Also, referencing the contributors chat without saying where it is is going to confuse people that aren't already on the Discord server. Also also, as someone who doesn't really _like_ discord and its proprietary nature, I'd like to avoid linking to the discord from the source code.

T0mstone (commented):
The second commit also has  a typo ("macth" instead of "match")

crates/typst-syntax/src/ast.rs
laurmaedje: I agree that we should leave it out here, mostly because we don't want to add this all over the place. The CONTRIBUTING.md is the correct place for this and it has that info already.

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+3 methods that specify its abstract representation: [`Raw::lines()`] returns the
+raw text as an iterator of lines, [`Raw::lang()`] provides the optionally
+present [`RawLang`] language tag, and [`Raw::block()`] gives a bool for whether
+the raw element is a block or inline.
+
+These methods are implemented by assuming the structure of [`SyntaxNode`]s will
+match what is generated by the parser. The [`Raw`] struct really is just a
+wrapper over a [`SyntaxNode`] reference. But to construct the wrapper and access
+the methods, we must check that the [`SyntaxKind`] of the underlying node
+matches.
```
PgBiel, line 41: Might not be the best wording, but feels like an improvement on explaining what is really going on:

```suggestion
These methods are implemented by accessing the underlying CST structure
as needed, and thus assuming the children of each [`SyntaxNode`] generated
by the parser have the expected order and structure. The [`Raw`] struct really is just a
wrapper over a [`SyntaxNode`] reference, but to construct that AST wrapper and access
its methods, we must check that the [`SyntaxKind`] of the underlying node
matches. If it matches, we know it will have the expected children underneath, otherwise
the parser wouldn't have produced this node.
```

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+## The AST is Lazy
+
+Being lazy means that the untyped CST nodes are converted to typed AST Nodes
+only as the tree is traversed. If we parse a file and a raw block is contained
+in a branch of an if-statement that we don't take, then we won't pay the cost of
+creating an iterator over the lines or checking whether it was a block or
+inline (although it will still be parsed into nodes). This is also a factor of
+the current "tree-interpreter" evaluation model. A bytecode interpreter might
+instead eagerly convert the AST into bytecode, but it will still traverse using
+this lazy interface.
```
PgBiel, line 57: Debatable, but would be nice to give more insight into this difference:

```suggestion
Being lazy means that the untyped CST nodes are converted to typed AST Nodes
only as the tree is traversed. If we parse a file and a raw block is contained
in a branch of an if-statement that we don't take, then we won't pay the cost of
creating an iterator over the lines or checking whether it was a block or
inline (although it will still be parsed into nodes).

This is also a factor of
the current "tree-interpreter" evaluation model. A bytecode interpreter might
instead eagerly convert the AST into bytecode, but it will still traverse using
this lazy interface. The advantage, in that case, is that it would only traverse
through the same node once. The current interpreter has to traverse it multiple
times to execute the same code. However, it is also more flexible and easier to
develop upon.
```

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+[`Expr`] is important because it contains the majority of expressions that Typst
+will evaluate. Not just markup, but also math and code expressions. Not all
+expression types are available from the parser at every step, but this does
+decrease the amount of wrapper enums needed in the AST, and this file is long
+enough already.
+
+Expressions also branch off into the remaining tree. You can view enums in this
+file as edges on a graph: areas where the tree has paths from one type to
+another (accessed through methods), then structs are the nodes of the graph,
+providing methods that return enums, etc. etc.
```
PgBiel, line 72: This feels like this should be near the top.

crates/typst-syntax/src/ast.rs
```diff
@@ -27,13 +104,27 @@ pub trait AstNode<'a>: Sized {
...
 
+/// Implements [`AstNode`] for a struct whose name matches a [`SyntaxKind`]
+/// variant.
+///
+/// The struct becomes a wrapper around a [`SyntaxNode`] pointer, and the
+/// implementation of [`AstNode::from_untyped`] checks that the pointer's kind
+/// matches when converting, returning `Some` or `None` respectively.
+///
+/// The generated struct is the basis for typed accessor methods for properties
+/// of this ast node. For example, the [`Raw`] struct has methods for accessing
```
PgBiel, line 115:
```suggestion
/// of this AST node. For example, the [`Raw`] struct has methods for accessing
```

crates/typst-syntax/src/ast.rs
```diff
@@ -27,13 +104,27 @@ pub trait AstNode<'a>: Sized {
...
+/// its content by lines, its optional language tag, and whether the raw element
+/// is inline or a block. These methods are accessible only _after_ a
+/// `SyntaxNode` is coerced to the `Raw` struct type (via `from_untyped`).
```
PgBiel, line 118: Perhaps as a complement to what was being said earlier:

```suggestion
/// `SyntaxNode` is coerced to the `Raw` struct type (via `from_untyped`),
/// guaranteeing their implementations will work with the expected structure.
```

crates/typst-syntax/src/ast.rs
```diff
@@ -27,13 +104,27 @@ pub trait AstNode<'a>: Sized {
...
-    ($(#[$attr:meta])* $name:ident) => {
+    ($(#[$attr:meta])* struct $name:ident) => {
+        // Create the struct as a wrapper around a `SyntaxNode` reference.
         #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
         #[repr(transparent)]
         $(#[$attr])*
         pub struct $name<'a>(&'a SyntaxNode);
 
+        // Implement `AstNode` for the struct.
         impl<'a> AstNode<'a> for $name<'a> {
```
PgBiel, line 128: Not entirely sure whether this comment is doing much here.

crates/typst-syntax/src/ast.rs
```diff
@@ -220,64 +250,68 @@ impl<'a> Expr<'a> {
...
-            SyntaxKind::Contextual => node.cast().map(Self::Contextual),
-            SyntaxKind::Conditional => node.cast().map(Self::Conditional),
-            SyntaxKind::WhileLoop => node.cast().map(Self::While),
-            SyntaxKind::ForLoop => node.cast().map(Self::For),
-            SyntaxKind::ModuleImport => node.cast().map(Self::Import),
-            SyntaxKind::ModuleInclude => node.cast().map(Self::Include),
-            SyntaxKind::LoopBreak => node.cast().map(Self::Break),
-            SyntaxKind::LoopContinue => node.cast().map(Self::Continue),
-            SyntaxKind::FuncReturn => node.cast().map(Self::Return),
+            SyntaxKind::Linebreak => Some(Self::Linebreak(Linebreak(node))),
```
PgBiel, line 253: In a way, this is sort of a downgrade in safety since we're skipping the `from_untyped` implementation now, so we could accidentally pass the wrong node to a particular Expr variant, but in another, I guess there are less redundant steps. An alternative to guarantee there won't be a mismatch but also improve performance would be to have a separate macro that generates each line in the match, but that could be discussed.

An alternative is to instead create an ad-hoc macro to generate unit tests which guarantee the casting works correctly.

crates/typst-syntax/src/ast.rs
```diff
@@ -1215,7 +1249,7 @@ pub enum ArrayItem<'a> {
 impl<'a> AstNode<'a> for ArrayItem<'a> {
     fn from_untyped(node: &'a SyntaxNode) -> Option<Self> {
         match node.kind() {
-            SyntaxKind::Spread => node.cast().map(Self::Spread),
+            SyntaxKind::Spread => Some(Self::Spread(Spread(node))),
```
PgBiel, line 1252: same comment regarding skipping `from_untyped()`.  (just adding additional markers so we know which portions of the code would be impacted by the discussion)

crates/typst-syntax/src/ast.rs
```diff
@@ -1254,9 +1288,9 @@ pub enum DictItem<'a> {
 impl<'a> AstNode<'a> for DictItem<'a> {
     fn from_untyped(node: &'a SyntaxNode) -> Option<Self> {
         match node.kind() {
-            SyntaxKind::Named => node.cast().map(Self::Named),
-            SyntaxKind::Keyed => node.cast().map(Self::Keyed),
-            SyntaxKind::Spread => node.cast().map(Self::Spread),
+            SyntaxKind::Named => Some(Self::Named(Named(node))),
```
PgBiel, line 1291: same comment regarding `from_untyped` here as well

crates/typst-syntax/src/ast.rs
```diff
@@ -1666,8 +1700,8 @@ pub enum Arg<'a> {
 impl<'a> AstNode<'a> for Arg<'a> {
     fn from_untyped(node: &'a SyntaxNode) -> Option<Self> {
         match node.kind() {
-            SyntaxKind::Named => node.cast().map(Self::Named),
-            SyntaxKind::Spread => node.cast().map(Self::Spread),
+            SyntaxKind::Named => Some(Self::Named(Named(node))),
```
PgBiel, line 1703: same comment regarding `from_untyped` here as well

crates/typst-syntax/src/ast.rs
```diff
@@ -1731,8 +1765,8 @@ pub enum Param<'a> {
 impl<'a> AstNode<'a> for Param<'a> {
     fn from_untyped(node: &'a SyntaxNode) -> Option<Self> {
         match node.kind() {
-            SyntaxKind::Named => node.cast().map(Self::Named),
-            SyntaxKind::Spread => node.cast().map(Self::Spread),
+            SyntaxKind::Named => Some(Self::Named(Named(node))),
```
PgBiel, line 1768: same comment regarding `from_untyped` here as well

crates/typst-syntax/src/ast.rs
```diff
@@ -1762,9 +1796,9 @@ pub enum Pattern<'a> {
 impl<'a> AstNode<'a> for Pattern<'a> {
     fn from_untyped(node: &'a SyntaxNode) -> Option<Self> {
         match node.kind() {
-            SyntaxKind::Underscore => node.cast().map(Self::Placeholder),
-            SyntaxKind::Parenthesized => node.cast().map(Self::Parenthesized),
-            SyntaxKind::Destructuring => node.cast().map(Self::Destructuring),
+            SyntaxKind::Underscore => Some(Self::Placeholder(Underscore(node))),
```
PgBiel, line 1799: same comment regarding `from_untyped` here as well

crates/typst-syntax/src/ast.rs
```diff
@@ -1841,8 +1875,8 @@ pub enum DestructuringItem<'a> {
 impl<'a> AstNode<'a> for DestructuringItem<'a> {
     fn from_untyped(node: &'a SyntaxNode) -> Option<Self> {
         match node.kind() {
-            SyntaxKind::Named => node.cast().map(Self::Named),
-            SyntaxKind::Spread => node.cast().map(Self::Spread),
+            SyntaxKind::Named => Some(Self::Named(Named(node))),
```
PgBiel, line 1878: same comment regarding `from_untyped` here as well

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+Typst's Abstract Syntax Tree (AST) is as a lazy, typed view over the untyped
+Concrete Syntax Tree (CST) and is rooted in the [`Markup`] node.
+
+## The AST is a View
+
+Most AST nodes are wrapper structs around [`SyntaxNode`] pointers. This summary
+will use a running example of the [`Raw`] node type, which is declared (after
+macro expansion) as: `struct Raw<'a>(&'a SyntaxNode);`.
+
+[`SyntaxNode`]s are generated by the parser and constitute the Concrete Syntax
```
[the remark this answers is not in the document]
T0mstone: It doesn't have to be at the bottom, you could also write
```
[`SyntaxNode`s](SyntaxNode)
```
Not sure which one is better tho.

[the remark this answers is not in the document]
laurmaedje: Fwiw, the `]s` thing is not uncommon in the codebase.

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
-//! A typed layer over the untyped syntax tree.
-//!
-//! The AST is rooted in the [`Markup`] node.
+/*!
```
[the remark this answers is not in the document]
wrzian: I'd like to keep the `/*! */` syntax if possible, but I can see the argument for consistency. Maybe we could use the heuristic of whether a module comment includes markdown headers or not?

crates/typst-syntax/src/ast.rs
wrzian: I think it's helpful simply because it's inside a macro. For newcomers to Rust, how the macro works might not be obvious, so this and the previous comment help with that. But this one is a bit simple, so I could be convinced to remove it.

crates/typst-syntax/src/ast.rs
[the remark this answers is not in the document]
wrzian: Seems the only others are: `crates/typst-kit/src/lib.rs`, `crates/typst-kit/src/fonts.rs`, and `crates/typst/src/lib.rs`

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+raw node will only use its child [`RawDelim`] CST nodes to determine whether the
+element is a block or inline.
+
+## The AST is Typed
+
+The AST gives access to methods on nodes of the tree. The [`Raw`] node provides
+3 methods that specify its abstract representation: [`Raw::lines()`] returns the
+raw text as an iterator of lines, [`Raw::lang()`] provides the optionally
+present [`RawLang`] language tag, and [`Raw::block()`] gives a bool for whether
+the raw element is a block or inline.
```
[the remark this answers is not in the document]
wrzian: I rewrote this to the following:
```markdown
## The AST is Typed

AST nodes all implement the [`AstNode`] trait, but nodes can also implement
their own unique methods. These unique methods are the "real" interface of the
AST, and provide access to the abstract, semantic, representation of each kind
of node. For example, the [`Raw`] node provides 3 methods that specify its
abstract representation: [`Raw::lines()`] returns the raw text as an iterator of
lines, [`Raw::lang()`] provides the optionally present [`RawLang`] language tag,
and [`Raw::block()`] gives a bool for whether the raw element is a block or
inline.

This semantic information is unavailable in the CST, only by converting a CST
node to an AST struct will Rust let you call a method of that struct. This is a
safe interface because the only way to create an AST node outside this file is
to call  [`AstNode::from_untyped`]. The `node!` macro implements `from_untyped`
by checking the node's kind before constructing it, returning `Some()` only if
the kind matches. So we know that it will have the expected children underneath,
otherwise the parser wouldn't have produced this node.
```

crates/typst-syntax/src/ast.rs
wrzian: I rewrote this to the following:
```markdown
## The AST is Lazy

Being lazy means that the untyped CST nodes are converted to typed AST nodes
only as the tree is traversed. If we parse a file and a raw block is contained
in a branch of an if-statement that we don't take, then we won't pay the cost of
creating an iterator over the lines or checking whether it was a block or
inline (although it will still be parsed into nodes).

This is also a factor of
the current "tree-interpreter" evaluation model. A bytecode interpreter might
instead eagerly convert the AST into bytecode, but it would still traverse using
this lazy interface. While the tree-interpreter evaluation is straightforward and
easy to add new features onto, it has to re-traverse the AST every time a
function is evaluated. A bytecode interpreter using the lazy interface would
only need to traverse each node once, improving throughput at the cost of
initial latency and development flexibility.
```

crates/typst-syntax/src/ast.rs
wrzian: I rewrote this to the following:
```markdown
## The AST is Lazy

Being lazy means that the untyped CST nodes are converted to typed AST nodes
only as the tree is traversed. If we parse a file and a raw block is contained
in a branch of an if-statement that we don't take, then we won't pay the cost of
creating an iterator over the lines or checking whether it was a block or
inline (although it will still be parsed into nodes).

This is also a factor of the current "tree-interpreter" evaluation model. A
bytecode interpreter might instead eagerly convert the AST into bytecode, but it
would still traverse using this lazy interface. While the tree-interpreter
evaluation is straightforward and easy to add new features onto, it has to
re-traverse the AST every time a function is evaluated. A bytecode interpreter
using the lazy interface would only need to traverse each node once, improving
throughput at the cost of initial latency and development flexibility.
```

crates/typst-syntax/src/ast.rs
wrzian: I moved it it to come after _The AST is Typed_, but before _The AST is Lazy_ since I think that makes the structure flow better.

crates/typst-syntax/src/ast.rs
```diff
@@ -1,6 +1,83 @@
...
+provide access to the semantic meaning of their contents. That semantic meaning
+is available throught the Abstract Syntax Tree by iterating over CST nodes and
+inspecting their contents. The format is prepared ahead-of-time by the parser so
+that this module can unpack the abstract meaning from the CST's structure.
+
+Raw nodes are parsed by recognizing paired backtick delimiters, which you will
+find as CST nodes with the [`RawDelim`] kind. However, the AST doesn't include
+these delimiters because it _abstracts_ over the backticks. Instead, the parent
+raw node will only use its child [`RawDelim`] CST nodes to determine whether the
+element is a block or inline.
```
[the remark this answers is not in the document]
wrzian: Not sure what you mean by 'discard'?

wrzian:
I pushed the rewrites and also 4 small changes:
1. I made the conversion of `SyntaxKind::Space` to `None` more explicit in `Expr::from_untyped`, since it's an important edge case that doesn't deserve to be part of the wildcard match.
2. I moved the generic SyntaxNode casting impls from `node.rs` to `ast.rs` to improve locality. I also made `cast_first_match` and `cast_last_match` private when doing so, although if there are other crates that need them, we can revert that.
3. I added two new functions `get_first()` and `get_last()` to replace and shorten the very common idiom of `cast_first_match().unwrap_or_default()` which I think goes along with improving clarity.
4. I simplified some AST methods that seemed straightforward to improve.

crates/typst-syntax/src/ast.rs
laurmaedje: I agree that this is unnecessary.

crates/typst-syntax/src/ast.rs
```diff
@@ -27,13 +102,65 @@ pub trait AstNode<'a>: Sized {
     }
 }
 
+// A generic interface for converting untyped nodes into typed AST nodes.
+impl SyntaxNode {
```
laurmaedje, line 106: I'm not a fan of this non-local impl. I'd rather have this stay in `node.rs`. At least the public parts.

crates/typst-syntax/src/ast.rs
```diff
@@ -27,13 +102,65 @@ pub trait AstNode<'a>: Sized {
...
+    ///
+    /// This and `get_last` are largely a convenience to avoid adding a `panic!`
+    /// without any unsafe code because the Parser can be hard to reason about,
+    /// but should contain this element.
+    fn get_first<'a, T: AstNode<'a> + Default>(&'a self) -> T {
+        self.cast_first_match().unwrap_or_default()
+    }
+
+    /// Get the last child of AST type `T` or a placeholder if none.
+    fn get_last<'a, T: AstNode<'a> + Default>(&'a self) -> T {
```
laurmaedje, line 137: I'm not super happy with the names. How about:
```
cast_first_match -> try_cast_first
cast_last_match -> try_cast_last
get_first -> cast_first
get_last -> cast_last
```

crates/typst-syntax/src/ast.rs
```diff
@@ -528,21 +599,19 @@ impl<'a> Raw<'a> {
     }
 
     /// An optional identifier specifying the language to syntax-highlight in.
+    ///
+    /// Only blocky raw elements can contain a language tag.
     pub fn lang(self) -> Option<RawLang<'a>> {
-        // Only blocky literals are supposed to contain a language.
-        let delim: RawDelim = self.0.cast_first_match()?;
```
laurmaedje, line 533: What happened to this check? In general, I'd say any non-obvious refactorings should be kept out and done in separate PRs.

crates/typst-syntax/src/ast.rs
laurmaedje: I also had that thought, but it's true that it's also a performance improvement (unless the compiler managed to optimize that away before). As it's not too hard to check that it's that correct, I think it's ok.

crates/typst-syntax/src/ast.rs
```diff
@@ -86,132 +213,71 @@ impl<'a> Markup<'a> {
 /// An expression in markup, math or code.
 #[derive(Debug, Copy, Clone, Hash)]
 pub enum Expr<'a> {
-    /// Plain text without markup.
```
laurmaedje, line 89: While the comments are a bit redundant, I like them, also in the docs.

crates/typst-syntax/src/ast.rs
wrzian: I feel like this one is a big win for clarity, mainly that these are the _only_ methods on `SyntaxNode` that reference `AstNode`, so this lets us avoid `SyntaxNode` importing `ast` at all. And externally these methods will only be usable if you have already imported `ast` since they use the `AstNode` trait.

But I'm ok to move `cast` and `is` if you think we should.

crates/typst-syntax/src/ast.rs
wrzian: Removing

crates/typst-syntax/src/ast.rs
wrzian: Reverting. I'll open another PR for this

crates/typst-syntax/src/ast.rs
```diff
@@ -27,13 +102,65 @@ pub trait AstNode<'a>: Sized {
...
+    /// Find the last child that can cast to the AST type `T`.
+    fn cast_last_match<'a, T: AstNode<'a>>(&'a self) -> Option<T> {
+        self.children().rev().find_map(Self::cast)
+    }
+
+    /// Get the first child of AST type `T` or a placeholder if none.
+    ///
+    /// This and `get_last` are largely a convenience to avoid adding a `panic!`
+    /// without any unsafe code because the Parser can be hard to reason about,
+    /// but should contain this element.
```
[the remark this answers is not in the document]
wrzian: Ah, neat. I'll remove this comment then.

crates/typst-syntax/src/ast.rs
wrzian: Done

wrzian:
Alright, rebased off main and tried to update to match the new feedback, I also reorganized the commits again so they should be more atomic.

There are still a few unresolved comments I'd like responses to, but I'm otherwise happy with this.

Thanks everyone for the feedback! I appreciate the help :)

crates/typst-syntax/src/ast.rs
laurmaedje: Let's have a third opinion. @PgBiel?

PgBiel: I also think it can be potentially clearer to implement AstNode-related functions here, or at least separately, because this is where they are effectively used. But I'm a bit torn on whether this is enough of an argument to keep _methods_ separate from the struct definition as we usually do elsewhere in the codebase.

One way around this could be to simply use functions instead of methods, but if we have too much chaining, that could be worse.

In that case, I could see keeping the methods at `ast.rs` with the argument that SyntaxNode already has a lot of unrelated methods on its own, and it's clearer what they are for when implemented this way. I could be fine with this change in that case, if we agree on this premise.

wrzian: Thanks. I'll echo that SyntaxNode has a _lot_ of methods, so splitting them out helps avoid making them look like they're related to other functionality of the SyntaxNode like span numbering or iteration

crates/typst-syntax/src/ast.rs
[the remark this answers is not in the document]
wrzian: Bump?

crates/typst-syntax/src/ast.rs
laurmaedje: Okay, we can keep it like this then. I don't mind too much.

wrzian: Alright, then I'm ready to merge.

laurmaedje (approved):

laurmaedje:
Looks good to merge for me.

> There are still a few unresolved comments I'd like responses to, but I'm otherwise happy with this.

Which? I see only [this one](https://github.com/typst/typst/pull/5784#discussion_r1936003291), but I think the wording is fine tbh.

wrzian:
That quote was mainly for the other comment that we already resolved.

laurmaedje:
Okay. Then, thank you!

PR #5789, opened 2025-01-31
Revert adding `flatten-text` to `image`

laurmaedje:
This reverts https://github.com/typst/typst/pull/5390

Adding a configuration for making text unselectable in Typst more generally (e.g. `set text(selectable: false)`) is realistic in the near future and this will naturally allow for `show image: set text(selectable: false)`. For this reason, I'd rather not release a parameter that we already know will be obsolete soon.

### DIFF
--- a/crates/typst-layout/src/image.rs
+++ b/crates/typst-layout/src/image.rs
@@ -63,7 +63,6 @@ pub fn layout_image(
             SvgImage::with_fonts(
                 data.clone(),
                 engine.world,
-                elem.flatten_text(styles),
                 &families(styles).map(|f| f.as_str()).collect::<Vec<_>>(),
             )
             .at(span)?,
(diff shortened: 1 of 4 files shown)


marie-bnl:
So basically, we're dealing with two different issues.

1. A user wants to make the text of a diagram unselectable so it doesn't get in the way when selecting paragraphs
2. A user wants to ensure their svg, even its text, looks the same on every device

Both issues have a common fix: render the text to paths into the svg output, which is, as I understand it, what `selectable: false` does.

However, these two are semantically independent, very separate concerns, which is problematic.

1. The name "selectable" only works for user 1, it is pretty counter-intuitive that making text unselectable leads to it looking the same on every device
2. This prevents pull requests trying to implement making text both rendered to a path and selectable such as https://github.com/typst/typst/pull/7313

This is probably not urgent though; it does not break backward compatibility to keep this as a solution to issue 1 and later add another option which does render the text to paths as well but does not imply that is will not be selectable in order to solve issue 2 (except if you absolutely want these two options to be unified).

Issue #5778, opened 2025-01-29
Incorrect spacing in heading between full-width numbering symbols and content


hongjr03:
### Description

I noticed that typst will automatically add a 0.3em width space after the heading's numbering, which is usually fine, but Chinese is used to use the full-width symbols `、` as the symbol for the heading's numbering, and both the full-width symbols `、` and the 0.3em width space create a part of the space that can lead to too wide a space between the heading's numbering and its content.

<img width="352" alt="Image" src="https://github.com/user-attachments/assets/c657a4c2-85a7-4b2a-a737-82fdbd73b41d" />

### Reproduction URL

https://typst.app/project/rtKJvE6vqm8aq5bDuQi9Jh

### Operating system

macOS

### Typst version

- [x] I am using the latest version of Typst

YDX-2147483647:
There is also a need to increase the width (or more precisely, fix the width of numbering+spacing) here. #2351

## Each heading may have at least four forms

These form may not all be the same.
Just providing four parameters can solve the problem, but it's too complicated…

- **displaying** (the basic form)

- **referenced**

  Challenge: alternative numbering and spacing, and one-off `supplement` customization like `@intro[Chapter]`.
  See #5102, #2485, #6032, #2990.

- **outlined** (in table of contents within the document)

  Challenge: small font size, compact space.
  See #1889.

- **bookmarked** (in PDF's outline)

  Challenge: plain text.
  See https://github.com/typst/typst/issues/1889#issuecomment-3008583386, and discussion in #6622.

- **other**: header/footer, HTML sidebar, screen-reader-only version, etc.

  It's too early to consider them now.

<img width="1174" height="679" alt="Image" src="https://github.com/user-attachments/assets/65879df3-5186-4d53-b597-47a6c0613a73" />

### Comparisons

<details>
<summary>displaying ≠ referenced ∧ outlined ≠ referenced ∧ bookmarked ≠ referenced (<code>一、标题</code> ≠ <code>第一章</code>)</summary>

Of course they might be different!

The issue is that there are numerous possibilities for permutations/subsets of
- numbering (which may have different forms, e.g., `1.1.` vs. `1.1`, or `一、` vs. `一`),
- separator (`#h(0.3em)`, `#h(1em)`, or `2em - measure(numbering).width`),
- supplement (which may consist of two parts, as in `第{1:一}章`),
- and the page number.

[...]

PR #5795, opened 2025-02-02
Add spacing_to_numbering field to heading

hongjr03:
Closes https://github.com/typst/typst/issues/5778.

This PR adds the `spacing_to_numbering` field to the heading and sets its default value to `0.3em`. This change maintains the same behavior as the default prior to this pr, but it allows users to set the spacing between the numbering and the heading content, which may be beneficial in CJK contexts.


Raphael-CV:
It is pretty the same as #5583.

hongjr03:
> It is pretty the same as #5583.

oops i didn't notice that, sorry. I'll close this pr.

PR #5798, opened 2025-02-03
Scope deprecations

laurmaedje:
This PR adds support for deprecating arbitrary definition in a `Scope`. It is an alternative to https://github.com/typst/typst/pull/5582 that builds on the refactorings done in https://github.com/typst/typst/pull/5797.

Compared to #5582, the main differences are:

- There's no accumulation of `define_*` variants because with #5797 the existing define variants return an `&mut Binding` on which we can chain `.deprecated(..)`.

- Used a sink argument like proposed in https://github.com/typst/typst/pull/5582#issuecomment-2556736867. It's true that it's not less viral than a `MaybeDeprecated`, but it's still more lightweight because ignoring means just passing `()`. Also, we can skip the whole wrapper type with all its conversion methods. Probably it's also a bit more efficient, but that's more of a guess.

- I kept `binding.read()` (without deprecation check) and `binding.read_checked(..)` (with deprecation check) separate for now. That means one could accidentally use `read` where `read_checked` would be more apprioriate, but I wanted to keep the virality down for now. We can reconsider this in the future.

- Instead of `Arc<EcoString>` (which is double reference-counted), I used `&'static str` as we only have static messages for now. I also considered `&'static &'static str` to bring it from two down to one machine words, but I figured that might be annoying to integrate with `codex`. Since the message is only stored in a `Binding` and not in a `MaybeDeprecated`, having the absolute minimum memory size is also not as crucial as in the other PR.

- Added support for `#[deprecated = ".."]` in `#[scope]`.

### DIFF
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -315,13 +315,15 @@ fn eval_field_call(
         (target, args)
     };
 
+    let field_span = field.span();
+    let sink = (&mut vm.engine, field_span);
     if let Some(callee) = target.ty().scope().get(&field) {
         args.insert(0, target_expr.span(), target);
-        Ok(FieldCall::Normal(callee.read().clone(), args))
+        Ok(FieldCall::Normal(callee.read_checked(sink).clone(), args))
     } else if let Value::Content(content) = &target {
         if let Some(callee) = content.elem().scope().get(&field) {
             args.insert(0, target_expr.span(), target);
-            Ok(FieldCall::Normal(callee.read().clone(), args))
+            Ok(FieldCall::Normal(callee.read_checked(sink).clone(), args))
         } else {
             bail!(missing_field_call_error(target, field))
         }
@@ -331,7 +333,7 @@ fn eval_field_call(
     ) {
         // Certain value types may have their own ways to access method fields.
         // e.g. `$arrow.r(v)$`, `table.cell[..]`
-        let value = target.field(&field).at(field.span())?;
+        let value = target.field(&field, sink).at(field_span)?;
         Ok(FieldCall::Normal(value, args))
     } else {
         // Otherwise we cannot call this field.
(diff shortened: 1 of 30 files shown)


PgBiel (approved):
Looks like a nice and clean solution. :+1:

PgBiel (commented):
My only potential restraint is that the empty sink argument might not be too obvious when it's used. I wonder if passing something like a `IgnoreDeprecation` zero-sized struct would be clearer (unless I misunderstood the purpose of it).

laurmaedje:
You understood correctly. I just wanted to keep it as unnoisy as possible for the ignoring case for now.

PgBiel:
Okay. Given the description, I assume these functions will receive further changes in the future anyway.

MDLC01:
I agree that having a `IgnoreDeprecation` type would be clearer, but if this is gonna change in the future anyway it probably does not matter. This solution is indeed more lightweight than mine. I closed #5582 in favor of this.

Also, why not renaming `read_checked` to just `read`, and the current `read` to `read_unchecked`? That way, the most obvious way to access the value is through the function that requires thinking about warnings.

laurmaedje:
I definitely understand your argument and normally I would be with you, trying to make the first method one reaches for the one where one has to think and consider the options. (Like with the new `Scope::bind`.)

Here I precisely made the opposite choice because I don't want to force thinking. I don't want the deprecation infrastructure to force any design decisions, extra thinking, or concern in other subsystems. It should be okay to just reach for `read()`. This is essentially what I mean with keeping the virality down for now, not just code-wise, also mentally. A missing deprecation warning in an edge case (i.e. not variable or field access) is not a big deal.

Once were more sure that this is how things should work, we can always reconsider this, it's just how I feel for the moment.

I also think that the naming `read_unchecked` would perhaps invoke too strong a resistance in Rust users due to the unsafety connotation, but that's besides my primary point.

PR #5803, opened 2025-02-03
Support underscores in numeric literals

### LINKED ISSUE #5592: Integer literal digit grouping
### Description

Integer literals should allow grouping digits, for example with an underscore: `123_456` would be equivalent to `123456`. While underscore is not the only character that can be used (e.g. C++ uses `'`) I do think it's the one that makes most sense for typst.

### Use Case

When writing long integer literals it might be hard to see what they actually represent. This is an especially big problem with base-2 and base-4 literals (`0b..`, `0o..`) which require more digits to write the same number, compared to base-10 or base-16. Digit separators help a lot with that.

bluebear94:
Fixes #5592.

Currently incurs a heap allocation for any literal with underscores, which is icky. Also, underscores are not allowed in the exponent unlike in Rust.


Enivex:
Does this not conflict with the syntax for subscripts in math?

I have to admit I'm not a huge fan of overloading _ with yet another purpose.

bluebear94:
No, the `Lexer::number` method is never called for numbers in math mode.

Enivex:
> No, the `Lexer::number` method is never called for numbers in math mode.

I didn't mean conflict in the literal sense. More that it makes code harder to read, especially once you mix code and math mode.

wrzian:
This does add ambiguity with underscores in math, but can be a really nice feature for code mode when needed—especially with non-decimal numbers. If wanted, we could limit the ambiguity by disallowing underscores when directly embedding numbers with a hashtag. That is, `#123_456` doesn't equal `123456`, but `#(123_456)` does.

bluebear94:
> If wanted, we could limit the ambiguity by disallowing underscores when directly embedding numbers with a hashtag. That is, `#123_456` doesn't equal `123456`, but `#(123_456)` does.

I think that rule would be unnecessarily complex. And what about expressions such as `#let a = 123_456`?

wrzian:
`#let a = 123_456` would work. This would operate exactly like binary operators currently do: `#let a = 1+2` is `3`, but `#1+2` is not.

laurmaedje:
I lean towards this not being worth it, considering Typst's other uses of `_`.

laurmaedje:
As said above, I don't think this is worth it for Typst specifically (even though I like it in other languages), so I'll close it. Thank you still!

Issue #5699, opened 2025-01-15
Support augment line at the end of matrix


NoamDev:
### Description

```
$
  mat(
    augment:#2,
    1,0;
    0,1;
  )
$
```

Output:

```

error: cannot draw a vertical line after column 2 of a matrix with 2 columns
  ┌─ \\...
  │
2 │ ╭   mat(
3 │ │     augment:#2,
4 │ │     1,0;
5 │ │     0,1;
6 │ │   )

```

The expected is for the augmented line to appear right after the second column. Which would look identical to the following code:

```
$
  mat(
    augment:#2,
    1,0,;
    0,1,;
  )
$
```

Note the added comma, which adds a third column of width 0.

### Operating system

Windows

### Typst version

- [X] I am using the latest version of Typst

Enivex:
What are you actually using the line for?

NoamDev:
I'm using an augmented matrix because I'm finding the row reduced echelon form.

I first type the matrix, then add augment line, then type the augmenting part.

So it's more about it being more natural while typing than the actual end result

Having an error stops the preview from updating.

To be honest it'd be useful if even when I'm trying to add an augmenting line that's bigger than the number of columns it would just add it at the end.

Enivex:
I agree that making this an error seems a bit too strict. There's no real reason to not just allow it.

PR #5806, opened 2025-02-04
Allow augment line at the beginning and end of a matrix

siddhantdev:
Fixes: https://github.com/typst/typst/issues/5699

Currently `mat` doesn't allow adding a vertical line after all columns. This lets the user set `augment` in a mat to be the same as the number of columns, which adds a vertical line at the end.

Augment `0` draws a line before the first column/row and augment `n` draws
a line after the last column/row of a matrix with `n` columns/rows.

`-1` now draws a line after the last column/row of the matrix, not before it.


Enivex:
Does this also allow the augment line at the beginning? It would seem natural if we're allowing it to be at the end.

siddhantdev:
I can update it to allow augment line at the beginning, with `#0`. It does seem natural

mkorje:
We may as well allow this for horizontal lines too then, I think.

Enivex:
> We may as well allow this for horizontal lines too then, I think.

I didn't realize until now that augment accepted a dictionary

Enivex:
One question is how this should function with negative numbers, since -1 is currently before the last column and not after it.

siddhantdev:
> One unanswered question is how this should function with negative numbers, since -1 is currently before the last column and not after it. Do we introduce a breaking change and make -1 correspond to after the last column?

I think leaving it as is should be fine. -1 can still mean just before the last column. Perhaps instead of thinking of -1 as last possible vertical line, it's just vertical line before last and 2nd last line and extend that for all negative values.

> We may as well allow this for horizontal lines too then, I think.

That seems fine. Should I just add that in this PR or create a separate PR for it?

Enivex:
Without a change it's impossible to put a line after the last column without counting the number of columns though

siddhantdev:
Perhaps we can do: any number higher than the current no. of cols adds a line to the end. Then any arbitrarily high number could be used to add a line to the end without having to manually count the number of columns.

laurmaedje:
I think a breaking change to let -1 refer to the last column is preferrable.

siddhantdev:
> I think a breaking change to let -1 refer to the last column is preferrable.

Alright. I'll update it to make -1 refer to the last column.

laurmaedje:
No need

mkorje:
Like I said before for horizontal lines at the beginning and end, negative indexes should work the same for horizontal lines too.

laurmaedje:
There are unfortunately some conflicts now due to https://github.com/typst/typst/pull/5934

siddhantdev:
I think the scope of this PR has been achieved. Is there anything else I need to do for merge?

siddhantdev:
Oh, ok! I thought I had something left to do. Thanks for the quick update.

laurmaedje (commented):
I think it would be simpler to normalize the indices to positive in one place instead of dealing with positive and negative everywhere.

siddhantdev:
Done!

laurmaedje:
When I commented https://github.com/typst/typst/pull/5806#issuecomment-2676122610, I was under the wrong impression that previously the indexing was inconsistent with how negative array indexing works in Typst (and Python), that this would make it consistent, and that the only problem is breakage. Only now, while taking a final look before merging, I realized that this now actually makes it _inconsistent_ w.r.t. array slicing (if you consider the augment line as a separator between two slices). I'm sorry for only noticing this now and speaking confidently about what change should be made earlier.

To summarize my thinking: In array slicing, `-1` always points _at_ the last element and we always consider a slice index as _before_ the element. However, in the case here, the index is only used for slicing, so pointing _at_ an element is not very meaningful. And since we want to be able to slice at the very end, interpreting the index as closer to the boundary from which it was specified is not entirely unreasonable. For what it's worth, without actively noticing the discrepancy with normal negative indexing, I landed on the same way to specify a cursor position in a string negatively [in typst-ide tests](https://github.com/typst/typst/blob/87cb8f5094b9ea45893344143ca069a521f9ba47/crates/typst-ide/src/tests.rs#L230-L238) as here (with -1 being at the very end).

Overall, it might be worth it to make things inconsistent here, but it comes with a somewhat bad feeling now. Thoughts? (also @Enivex and @mkorje)

Enivex:
I'm not sure what you mean. It's entirely consistent with how negative indices work. What you're indexing into is the spaces between the entries, since that's the places where augment lines can go.

```
  |  a  |  b  |  c  |  d  |
  0     1     2     3     4
                         -1
```

laurmaedje:
If you think about it that way, yes. Not sure that's the most natural way to think about it though. In my mental model, a matrix with 4 columns corresponds to an array with 4 elements (just like if a spreaded the array into the matrix) rather than an array with 5 spaces. The spaces between elements also exist in arrays after all (and they are used when slicing rather than indexing).

Enivex:
> If you think about it that way, yes. Not sure that's the most natural way to think about it though. In my mental model, a matrix with 4 columns corresponds to an array with 4 elements (just like if a spreaded the array into the matrix) rather than an array with 5 spaces. The spaces between elements also exist in arrays after all (and they are used when slicing rather than indexing).

That model is inadequate here, because it is incapable of referring to the augment line after the last column or last row. (Unless I'm missing something)

laurmaedje:
I get that. I'm just saying that I'm unsure whether the inconvenience of not being able to use negative indices for placing at the last row (you can still use positive indices and placing at the last row is probably rather rare) is worth making it inconsistent (at least inconsistent in my mental model).

Enivex:
> I get that. I'm just saying that I'm unsure whether the inconvenience of not being able to use negative indices for placing at the last row (you can still use positive indices and placing at the last row is probably rather rare) is worth making it inconsistent (at least inconsistent in my mental model).

You're saying yourself that you can use positive indices for after the last row, which would already be inconsistent with your mental model. In order to do that you have to index into an array of spaces.

At that point the negative indices just work consistently with how they work for every array.

laurmaedje:
Is it really inconsistent with my mental model? I'm just thinking in terms of slicing rather than in terms of indexing. E.g., I can write:

```typ
#let x = (1, 2, 3, 4)
#x.slice(0, 4) // => (1, 2, 3, 4)
#x.slice(0, -1) // => (1, 2, 3)
```

Only with a positive index, it is possible to slice out the full list. (Which is a bit of a pointless operation in this case, but it can be meaningful if the integer used for slicing results from some computation.)

Enivex:
For slicing you can typically have an empty argument for the same purpose, which is how they can get away with it. As in something like `a(5:-1)` vs `a(5:)`.

In any case I can't imagine many people are thinking of it this way.

PgBiel:
We discussed this among the team, and we were in agreement that using `-1` to represent something _after_ the last position seems inconsistent / awkward - we reached the same conclusion independently while reading the PR.

However, I came up with the following idea, which didn't receive much opposition in our meeting, but I'd like to hear others' opinions: we could use `start` / `end` (align).

Reusing @Enivex's diagram to illustrate:

```
  |  a  |  b  |  c  |  d  |
  0     1     2     3     4
  start -3   -2     -1   end
```

Pros:
- Quite intuitive as for what they mean.
- Keeps consistency of what numbers refer to between positive and negative.

Cons:
- Type becomes `int` / `alignment`, which might be confusing.
   - Maybe this could be solved by a docs-side improvement which only displays `start` / `end` directly.
- Can't use a number to refer to the last thing, needing a special case in otherwise generic code.
   - But that's the intended tradeoff in this solution anyway.

siddhantdev:
The augment line data is stored as a `isize SmallVec` of size 1. What's the recommended way to make a union type so that it can be `int` / `Alignment` ?

laurmaedje:
I'm happy to help guide you with the implementation in principle, but I'd first like to hear a few more opinions about the `int | alignment` idea, so that your work is not in vain.

laurmaedje:
There doesn't seem to be any consensus here, with 3+ people having argued for both sides and no significant majority. To me, this indicates that neither option (different indexing scheme or `int | alignment`) is the way to go. For this reason, I have decided to not move forward with neither of the two.

~~@siddhantdev I'm quite sorry for all the work you've put into this.There was a bit of a process failure here with too much low-level code discussions before the conceptual side was fully hashed out.Thank you still for your contribution!~~

laurmaedje:
Okay, I was a bit too fast here. I just realized that positive indices indicating the edges are not supported at all yet, so this part is still on the table.

laurmaedje:
Okay, I've reverted the breaking change to indices. The rest is fine. Thank you! @siddhantdev

PR #5807, opened 2025-02-04
Add smart quotes for Bulgarian

TwoF1nger:
I put apostrophe as both opening and closing "single quotes" because, in Bulgarian, there are no single quotes, only double.
(We do use apostrophe, for contractions like in _'tis_ or _guv'nor_. But it is not in pairs.)

A few other languages have done the same, so I guess it is fine.

Online sources - [double quotes][1] and [apostrophe][2].

[1]: https://bg.wikipedia.org/wiki/%D0%9A%D0%B0%D0%B2%D0%B8%D1%87%D0%BA%D0%B8
[2]: https://bg.wikipedia.org/wiki/%D0%90%D0%BF%D0%BE%D1%81%D1%82%D1%80%D0%BE%D1%84


laurmaedje:
Thanks!

PR #5801, opened 2025-02-03
0.13 changelog

laurmaedje:
Not fully final as hayagriva and codex still need to be bumped. But it's already pretty complete and there are a lot of nice changes in the release!

PR #5816, opened 2025-02-05
Documentation fixes and improvements

MDLC01:
This removes a useless "instead". I failed to notice that this "instead" was present in two places when writing https://github.com/typst/typst/pull/5801#discussion_r1939946980.

This also specifies that the "⊥" around which spacing was fixed is the up tack, not the perpendicular symbol.

I also included some other changes that don't need additional explanation.

### DIFF
--- a/crates/typst-library/src/visualize/image/mod.rs
+++ b/crates/typst-library/src/visualize/image/mod.rs
@@ -46,10 +46,11 @@ use crate::text::LocalName;
 /// ```
 #[elem(scope, Show, LocalName, Figurable)]
 pub struct ImageElem {
-    /// A path to an image file or raw bytes making up an image in one of the
-    /// supported [formats]($image.format).
+    /// A [path]($syntax/#paths) to an image file or raw bytes making up an
+    /// image in one of the supported [formats]($image.format).
     ///
-    /// For more details about paths, see the [Paths section]($syntax/#paths).
+    /// Bytes can be used to specify raw pixel data in a row-major,
+    /// left-to-right, top-to-bottom format.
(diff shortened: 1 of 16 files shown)


EpicEricEE:
image

Might as well add a link to `pdf.embed` here 😄

MDLC01:
I'm also thinking the documentation for [the `pdf.embed` `mime-type` argument](https://staging.typst.app/docs/reference/pdf/embed/#parameters-mime-type) should probably link to a website explaining what those are, but I'm unsure which one.

Maybe https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types/Common_types, or https://www.iana.org/assignments/media-types/media-types.xhtml, which is linked from the MDN, but less visually friendly.

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -46,10 +46,11 @@ use crate::text::LocalName;
 /// ```
 #[elem(scope, Show, LocalName, Figurable)]
 pub struct ImageElem {
-    /// A path to an image file or raw bytes making up an image in one of the
-    /// supported [formats]($image.format).
+    /// A [path]($syntax/#paths) to an image file or raw bytes making up an
+    /// image in one of the supported [formats]($image.format).
     ///
-    /// For more details about paths, see the [Paths section]($syntax/#paths).
```
laurmaedje, line 52: This sentence is used all over the docs. I'm okay with changing it, but it should then be done consistently.

laurmaedje:
Thanks!

PR #5824, opened 2025-02-06
Only autocomplete methods which take self

laurmaedje:
Autocompleting `.cell` on `#table().` doesn't make sense since it calling it with method syntax will always be an error. 

Not 100% sure on literally checking against the name `self`. I guess we could also check whether the value matches the type info for the first param, but that wouldn't work well with untyped user methods in the future. Same could be said for `self` though.

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -407,9 +407,17 @@ fn field_access_completions(
         elem.into_iter().chain(Some(ty))
     };
 
-    // Autocomplete methods from the element's or type's scope.
+    // Autocomplete methods from the element's or type's scope. We only complete
+    // those which have a `self` parameter.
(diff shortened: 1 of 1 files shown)


laurmaedje:
Anybody has thoughts on this?

Myriad-Dreamin:
@laurmaedje It looks kind of bad to check against functions by simply matching the names of parameters, because nobody other than us can learn this knowledge (convention) by reading the code of `crates/typst-ide` or `docs`. Extra `SignatureInfo` containing something like `is_method` could be used instead. I feel it is okay now since we are on beta and the standard library is small enough. There should be also some missing chapters to describe the value, type, parameter, and signature things formally.

laurmaedje:
> Extra SignatureInfo containing something like is_method could be used instead. 

Thing is, there isn't really the concept of a method in the first place. The proper solution would be to have a type hint that we can check self against.

PR #5836, opened 2025-02-07
Update documentation for float.{to-bits, from-bits}

bluebear94:


crates/typst-library/src/foundations/float.rs
```diff
@@ -110,7 +110,8 @@ impl f64 {
         f64::signum(self)
     }
 
-    /// Converts bytes to a float.
+    /// Interprets bytes as a float in [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754)
```
laurmaedje, line 113: At the moment, the first line of the docs shouldn't have a link because various places in the docs generation and IDE tooling expects the first line to be plain text.

bluebear94: I moved the link to the docs of the relevant argument; does that resolve the problem?

laurmaedje: Yes!

PR #5837, opened 2025-02-07
`Gradient::repeat`: Fix floating-point error in stop calculation

### LINKED ISSUE #5819: `gradient.repeat()` sometimes produces non-monotonic stops
### Description

Sometimes, when `repeat()`ing a gradient and retrieving the new gradients stops, it's not possible to feed them back into a gradient constructor. 
MWE
```typ
#let my-gradient = gradient.linear(red, blue).repeat(5)
#gradient.linear(..my-gradient.stops())
```
The error message is "Offsets must be in strictly monotonic order". 

Most interestingly, this only happens with 7,9,10,11,12,13,... repetitions. Moreoever, if the stops are hover-inspected and copied to a variable explicitly, like
```typ
#let stops = (
  (rgb("#ff4136"), 0%),
  (rgb("#0074d9"), 20%),
  (rgb("#ff4136"), 20%),
  (rgb("#0074d9"), 40%),
  (rgb("#ff4136"), 40%),
  (rgb("#0074d9"), 60%),
  (rgb("#ff4136"), 60%),
  (rgb("#0074d9"), 80%),
  (rgb("#ff4136"), 80%),
  (rgb("#0074d9"), 100%),
)

#gradient.linear(..stops)
```
the issue vanished. For this reason I assume that due to floating point errors, some offset values end off lower than others. 

One a side note, looking at line https://github.com/typst/typst/blob/4a9a5d2716fc91f60734769eb001aef32fe15403/crates/typst-library/src/visualize/gradient.rs#L1232, stops are only required to be in monotonic order, not in strictly monotonic order (regarding the phrasing of the error message). 

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

bluebear94:
Fixes #5819.

### DIFF
--- a/crates/typst-library/src/visualize/gradient.rs
+++ b/crates/typst-library/src/visualize/gradient.rs
@@ -582,12 +582,11 @@ impl Gradient {
                 let mut stops = stops
                     .iter()
                     .map(move |&(color, offset)| {
-                        let t = i as f64 / n as f64;
                         let r = offset.get();
                         if i % 2 == 1 && mirror {
-                            (color, Ratio::new(t + (1.0 - r) / n as f64))
+                            (color, Ratio::new((i as f64 + 1.0 - r) / n as f64))
                         } else {
-                            (color, Ratio::new(t + r / n as f64))
+                            (color, Ratio::new((i as f64 + r) / n as f64))
                         }
                     })
                     .collect::<Vec<_>>();
--- a/crates/typst-library/src/visualize/gradient.rs
+++ b/crates/typst-library/src/visualize/gradient.rs
@@ -1229,7 +1229,7 @@ fn process_stops(stops: &[Spanned<GradientStop>]) -> SourceResult<Vec<(Color, Ra
             };
 
             if stop.get() < last_stop {
-                bail!(*span, "offsets must be in strictly monotonic order");
+                bail!(*span, "offsets must be in monotonic order");
             }
 
             last_stop = stop.get();
--- a/tests/suite/visualize/gradient.typ
+++ b/tests/suite/visualize/gradient.typ
@@ -658,3 +658,9 @@ $ A = mat(
   height: 10pt,
   fill: gradient.linear(violet, blue, space: cmyk)
 )
+
+--- issue-5819-gradient-repeat ---
+#let my-gradient = gradient.linear(red, blue).repeat(5)
+#let _ = gradient.linear(..my-gradient.stops())
+#let my-gradient2 = gradient.linear(red, blue).repeat(5, mirror: true)
+#let _ = gradient.linear(..my-gradient2.stops())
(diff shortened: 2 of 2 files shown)


crates/typst-library/src/visualize/gradient.rs
```diff
@@ -582,12 +582,11 @@ impl Gradient {
...
                     .iter()
                     .map(move |&(color, offset)| {
-                        let t = i as f64 / n as f64;
                         let r = offset.get();
                         if i % 2 == 1 && mirror {
-                            (color, Ratio::new(t + (1.0 - r) / n as f64))
+                            (color, Ratio::new((i as f64 + 1.0 - r) / n as f64))
                         } else {
-                            (color, Ratio::new(t + r / n as f64))
+                            (color, Ratio::new((i as f64 + r) / n as f64))
```
PgBiel, line 589: I have a feeling that this doesn't fully solve the problem, since there is still some potential for imprecision. But it's nice that this improves the situation nonetheless by reducing the amount of divisions. Perhaps using fixed point numbers (a future task, not suitable for this PR) would be required to fully fix the issue.

I wonder if doing something like

```rs
Ratio::new(((i + 1) as f64 - r) / n as f64)
```

on the first one could improve it even further by relying more on integer arithmetic.

tests/suite/visualize/gradient.typ
```diff
@@ -658,3 +658,9 @@ $ A = mat(
   height: 10pt,
   fill: gradient.linear(violet, blue, space: cmyk)
 )
+
+--- issue-5819-gradient-repeat ---
+#let my-gradient = gradient.linear(red, blue).repeat(5)
+#let _ = gradient.linear(..my-gradient.stops())
+#let my-gradient2 = gradient.linear(red, blue).repeat(5, mirror: true)
+#let _ = gradient.linear(..my-gradient2.stops())
```
PgBiel, line 666: Since this has no visual output, might be nice to add a short test description.

```suggestion
--- issue-5819-gradient-repeat ---
// Ensure the gradient constructor generates monotonic stops which can be fed
// back into the gradient constructor itself.
#let my-gradient = gradient.linear(red, blue).repeat(5)
#let _ = gradient.linear(..my-gradient.stops())
#let my-gradient2 = gradient.linear(red, blue).repeat(5, mirror: true)
#let _ = gradient.linear(..my-gradient2.stops())
```

laurmaedje:
https://github.com/typst/typst/issues/5819 also mentions a mismatch between `if stop.get() < last_stop {` and the error message, which we should probably also address to properly fix it

laurmaedje:
Thanks!

crates/typst-library/src/visualize/gradient.rs
laurmaedje: It would be a bit cleaner, but I think it wouldn't make a material difference as f64 can accurately represent ints even beyond 2^32 and the indices won't be that large.

PgBiel: Okay

Issue #5847, opened 2025-02-10
Math not completing after dot


EpicEricEE:
### Description

Not sure if this is an issue of the web app or of the IDE features here, but as of 0.13.0-rc1, symbol modifiers or other fields aren't properly completed anymore.

For example `$ arrow.| $` doesn't show the available modifiers (e.g. `l`, `r`, `long`, ...), `$ #figure.| $` doesn't show `caption`, and so on. When typing the first letter of an available field/modifier, the matching completions show up.

I have recently reported a very similar issue to tinymist, though that only concerned _multiple_ fields/modifiers: Myriad-Dreamin/tinymist#1267

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
Only happens in math mode. Might be related to SymboElem AST changes.

PR #5849, opened 2025-02-11
Fix autocomplete and jumps in math

laurmaedje:
Fixes https://github.com/typst/typst/issues/5847.

The problems were checks for `SyntaxKind::Text` which is now `SyntaxKind::MathText`.


wrzian:
Sorry 😅

PR #5851, opened 2025-02-11
Lazy parsing of the package index

elegaanz:
To allow for future evolution of the format more easily. If a single package can't be parsed, the rest of the index is still valid and can be used normally.

### DIFF
--- a/crates/typst-kit/src/package.rs
+++ b/crates/typst-kit/src/package.rs
@@ -1,14 +1,11 @@
 //! Download and unpack packages and package indices.
 
 use std::fs;
-use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
 
 use ecow::eco_format;
 use once_cell::sync::OnceCell;
-use serde::de::DeserializeOwned;
 use serde::Deserialize;
-use serde_json;
 use typst_library::diag::{bail, PackageError, PackageResult, StrResult};
 use typst_syntax::package::{PackageSpec, PackageVersion, VersionlessPackageSpec};
 
@@ -123,8 +120,7 @@ impl PackageStorage {
             // version.
             self.download_index()?
                 .iter()
-                .lazy_deser::<MinimalPackageInfo>()
-                .filter_map(|res| res.ok())
+                .filter_map(|value| MinimalPackageInfo::deserialize(value).ok())
                 .filter(|package| package.name == spec.name)
                 .map(|package| package.version)
                 .max()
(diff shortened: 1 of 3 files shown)


crates/typst-kit/src/package.rs
```diff
@@ -186,3 +203,80 @@ impl PackageStorage {
         })
     }
 }
+
+/// An iterator that deserializes its items lazily.
+struct LazyDeser<T, I> {
```
elegaanz, line 208: Maybe this struct and related code could go in their own module?

crates/typst-kit/src/package.rs
```diff
@@ -186,3 +203,80 @@ impl PackageStorage {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn lazy_deser_index() {
+        let storage = PackageStorage::with_index(
+            None,
+            None,
+            Downloader::new("typst/test"),
```
elegaanz, line 250: This test should normally never download anything, but maybe we want to have a stronger guarantee that it can't (unless it's fine to make network calls in tests) by introducing an abstraction over `Downloader`?

crates/typst-kit/src/package.rs
```diff
@@ -186,3 +203,80 @@ impl PackageStorage {
...
+        });
+        assert_eq!(ieee_version, Ok(PackageVersion { major: 0, minor: 1, patch: 0 }));
+
+        let ams_version = storage.determine_latest_version(&VersionlessPackageSpec {
+            namespace: "preview".into(),
+            name: "unequivocal-ams".into(),
+        });
+        assert_eq!(
+            ams_version,
+            Err("failed to find package @preview/unequivocal-ams".into())
```
elegaanz, line 279: I left the error message as is for now to not introduce too much complexity, but maybe it would be nice to have a more specific one in this scenario, such as "a more recent version of Typst is required to use @preview/unequivocal-ams"? This would require the assumption that the `name` field will always be a string in the index.

I don't know what should be shown in case a more recent version maybe exists but the compiler can't parse it so it can't tell. Should it show a warning? An error? Silently ignore this version as it currently does?

crates/typst-kit/src/package.rs
```diff
@@ -109,6 +124,8 @@ impl PackageStorage {
             // version.
             self.download_index()?
                 .iter()
+                .lazy_deser::<PackageInfo>()
```
laurmaedje, line 127: since we only use name & version for this function, it might be nice to use a smaller type with just these two fields here to be resilient to widening type changes to `PackageInfo`'s other field

crates/typst-kit/src/package.rs
laurmaedje: network calls shouldn't occur, but I think it's fine since there aren't any. adding an abstraction just to have this as a static guarantee feels a little overkill.

crates/typst-kit/src/package.rs
laurmaedje: I think it's okay to leave it here until we need it somewhere else

crates/typst-kit/src/package.rs
laurmaedje: I think it's okay for now. It's primarily about not erroring for _every_ package.

crates/typst-kit/src/package.rs
```diff
@@ -109,6 +123,8 @@ impl PackageStorage {
             // version.
             self.download_index()?
                 .iter()
+                .lazy_deser::<MinimalPackageInfo>()
+                .filter_map(|res| res.ok())
```
laurmaedje, line 127: I wonder whether the `LazyDeser` infrastructure really pulls its weight compared to just having
```rust
.filter_map(|value| {
    serde_json::from_value::<MinimalPackageInfo>(value.clone()).ok()
})
```
inline here.

crates/typst-kit/src/package.rs
```diff
@@ -1,14 +1,16 @@
...
 
 use std::fs;
+use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
 
 use ecow::eco_format;
 use once_cell::sync::OnceCell;
+use serde::de::DeserializeOwned;
+use serde::Deserialize;
+use serde_json;
```
laurmaedje, line 11: This import is unused

crates/typst-kit/src/package.rs
laurmaedje: Update: We can actually use `MinimalPackageInfo::deserialize(value).ok()` to get rid of the `.clone()`

PR #5860, opened 2025-02-13
Add `deep-merge` function to dictionaries

### LINKED ISSUE #1957: Shallow/deep merging for dictionary
Dictionary is useful for scripting. It seems like only shallow merging is implemented for now (typst 0.7.0).

The following code

```typ
#let default = (
    alpha: (
        a: 1,
        b: 2,
    ),
    beta: 99,
)

#let override = (
    alpha: (
        c: 3,
    ),
    gamma: 100,
)

#let final = default + override

#default

#override

#final
```

results in

```
(alpha: (a: 1, b: 2), beta: 99)
(alpha: (c: 3), gamma: 100)
(alpha: (c: 3), beta: 99, gamma: 100)
```

We see that `final` is the **shallow** merging of `default` and `override`.

In contrast, the **deep** merging of `default` and `override` is of the form

```
(alpha: (a:1, b:2, c: 3), beta: 99, gamma: 100)
```

When I script with typst, I found that deep merging is what I really want in most cases.

evfinkn:
Closes #1957.

Only shallow merging dictionaries (using `+`) is supported currently. This adds a `deep-merge` method to dictionaries which merges nested dictionaries recursively.

After implementing this, I realized that I hadn't looked at what other languages have. I couldn't find any languages with a deep merge method in their standard library, but several third-party packages have one. Those libraries have three different behaviors regarding arrays at the same key path:

- Rails' [`deep_merge`](https://api.rubyonrails.org/v8.0.1/classes/Hash.html#method-i-deep_merge) and the [`deep_merge`](https://github.com/PragTob/deep_merge) Elixir library overwrite arrays at the same key path.
- The [`Hash::Merge`](https://metacpan.org/pod/Hash::Merge) Perl library and [deepmerge](https://github.com/toumorokoshi/deepmerge) Python library concatenate lists at the same key path.
- Lodash's [`mergeWith`](https://lodash.com/docs/4.17.15#mergeWith) and jQuery's [`extend`](https://api.jquery.com/jQuery.extend/) treat arrays exactly the same as objects, so elements at the same indices are overwritten or deep merged. 

Some of those methods also accept a function to customize the merging behavior:

- Lodash's `mergeWith`: the customizer function is called with arguments `(objValue, srcValue, key, object, source, stack)`. It's called for all keys—not just conflicting ones—and returning `undefined` from the customizer delegates to the default behavior for that key path.
- Rails' `deep_merge`: the customizer function is called with arguments `(key, this_val, other_val)` on conflicting keys when their values aren't deep mergeable.
- Elixir's `deep_merge` library: the customizer function is called with arguments `(key, original, override)` on all conflicting keys. Returning `DeepMerge.continue_deep_merge` from the customizer delegates to the default behavior for that key path.

Since all of those deep merge methods behave differently, I've left this `deep-merge` method simple (with no special behavior for arrays and no argument for a customizer function) since I'd like to hear what others think about how it should behave.

### DIFF
--- a/crates/typst-library/src/foundations/dict.rs
+++ b/crates/typst-library/src/foundations/dict.rs
@@ -6,10 +6,10 @@ use std::sync::Arc;
 use ecow::{eco_format, EcoString};
 use indexmap::IndexMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use typst_syntax::is_ident;
+use typst_syntax::{is_ident, Span};
 use typst_utils::ArcExt;
 
-use crate::diag::{Hint, HintedStrResult, StrResult};
+use crate::diag::{bail, Hint, HintedStrResult, SourceResult, StrResult};
 use crate::foundations::{
     array, cast, func, repr, scope, ty, Array, Module, Repr, Str, Value,
 };
@@ -156,6 +156,22 @@ impl Dict {
 
         msg.into()
     }
+
+    /// Deep merge the dictionary with another, modifying the dictionary in
+    /// place.
+    fn merge_with(&mut self, other: Dict) {
+        let map = Arc::make_mut(&mut self.0);
+        for (key, value) in other {
+            match (map.get_mut(&key), value) {
+                (Some(Value::Dict(existing)), Value::Dict(value_dict)) => {
+                    existing.merge_with(value_dict);
+                }
+                (_, value) => {
+                    map.insert(key, value);
+                }
+            }
+        }
+    }
 }
 
 #[scope]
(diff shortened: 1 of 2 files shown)


laurmaedje:
I realize that I added the "good contribution" label on the issue half a year ago (🙈), but I'm honestly not so convinced anymore that this is a good fit for the standard library. My main concerns are:
- As you mention, there are so widely varying behaviours across other language implementations.
- In many years of programming, I haven't really come across a need for this. Is it really a common need?

evfinkn:
It probably isn't a super common need since it isn't in any standard libraries. The main use case I can think of is applying user-supplied configuration on top of defaults. CeTZ actually does this—it has a [deep merge utility](https://github.com/cetz-package/cetz/blob/98ac77a/src/util.typ#L176-L191) that it uses to [update styles](https://github.com/cetz-package/cetz/blob/98ac77a/src/styles.typ#L139). Granted, it's not hard to implement in Typst manually should someone need it, and I agree that the lack of a standard behavior in other libraries is a concern.

laurmaedje:
> Granted, it's not hard to implement in Typst manually should someone need it, and I agree that the lack of a standard behavior in other libraries is a concern.

I would close this then. Thank you still and sorry for marking that as a good contribution!

PR #5867, opened 2025-02-15
nested/ancestry .contains selector

schrottkatze:
This implements a selector that gets all items within another item.

(originally was discussed as .contains, will rename it later, just PRing now to get feedback)

- [ ] Rename to `within`
- [ ] Tests

### DIFF
--- a/crates/typst-library/src/foundations/selector.rs
+++ b/crates/typst-library/src/foundations/selector.rs
@@ -93,6 +93,9 @@ pub enum Selector {
     Before { selector: Arc<Self>, end: Arc<Self>, inclusive: bool },
     /// Matches all matches of `selector` after `start`.
     After { selector: Arc<Self>, start: Arc<Self>, inclusive: bool },
+    /// Matches all children of `selector` matchin `c`
+    /// TODO: Better name
+    Contains { selector: Arc<Self>, c: Arc<Self> },
 }
 
 impl Selector {
@@ -139,7 +142,10 @@ impl Selector {
             }
             Self::Location(location) => target.location() == Some(*location),
             // Not supported here.
-            Self::Regex(_) | Self::Before { .. } | Self::After { .. } => false,
+            Self::Regex(_)
+            | Self::Before { .. }
+            | Self::After { .. }
+            | Self::Contains { .. } => false,
         }
     }
 }
@@ -221,6 +227,11 @@ impl Selector {
             inclusive,
         }
     }
+
+    #[func]
+    pub fn contains(self, c: LocatableSelector) -> Selector {
+        Self::Contains { selector: Arc::new(self), c: Arc::new(c.0) }
+    }
 }
 
 impl From<Location> for Selector {
@@ -266,6 +277,7 @@ impl Repr for Selector {
                     inclusive_arg
                 )
             }
+            Self::Contains { .. } => todo!(),
         }
     }
 }
@@ -352,7 +364,8 @@ impl FromValue for LocatableSelector {
                     }
                 }
                 Selector::Before { selector, end: split, .. }
-                | Selector::After { selector, start: split, .. } => {
+                | Selector::After { selector, start: split, .. }
+                | Selector::Contains { selector, c: split } => {
                     for selector in [selector, split] {
                         validate(selector)?;
                     }
@@ -431,7 +444,8 @@ impl FromValue for ShowableSelector {
                 | Selector::Location(_)
                 | Selector::Can(_)
                 | Selector::Before { .. }
-                | Selector::After { .. } => {
+                | Selector::After { .. }
+                | Selector::Contains { .. } => {
                     bail!("this selector cannot be used with show")
                 }
             }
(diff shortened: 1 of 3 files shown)


crates/typst-library/src/foundations/selector.rs
```diff
@@ -266,6 +280,9 @@ impl Repr for Selector {
                     inclusive_arg
                 )
             }
+            Self::Within { selector, ancestor: children_selector } => {
+                eco_format!("{}.contains({})", selector.repr(), children_selector.repr())
```
MultisampledNight, line 284: Outdated function name?

```suggestion
                eco_format!("{}.within({})", selector.repr(), children_selector.repr())
```

crates/typst-library/src/foundations/selector.rs
```diff
@@ -221,6 +226,15 @@ impl Selector {
             inclusive,
         }
     }
+
+    /// Returns a modified selector that only matches `self` if it is contained in an `ancestor`.
```
MultisampledNight, line 230: Suggestion: Directly note that the matching is done flatly:

```suggestion
    /// Returns a modified selector that only matches `self` if it is directly contained in an `ancestor`.
```

schrottkatze: it doesn't, but i apparently *do* need to dedupe (managed to get duplicate results using nested figures)

MultisampledNight: Fair, should probably reword then: Explicitly denote the depth behavior

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -185,6 +185,11 @@ impl Introspector {
                 }
                 list
             }
+            Selector::Within { selector, ancestor } => self
+                .query(ancestor)
+                .iter()
+                .flat_map(|children| children.query(selector))
```
jbirnick, line 191: I wouldn't call this `children`. Because each object of what you call `children` is just one of the elements that match `ancestor`, so they aren't really children in any way. If it all, you could in fact call it `parent`.

jbirnick:
I guess you need to be a bit careful with the `Content::query(...)` function that you are using to handle the the `within` query. The implementation is a bit "naive" in the sense that it simply looks at all the fields of the element:
https://github.com/typst/typst/blob/5fc679f3e7501ee5831f1b4b7789350f43b4c221/crates/typst-library/src/foundations/content.rs#L415-L423
https://github.com/typst/typst/blob/5fc679f3e7501ee5831f1b4b7789350f43b4c221/crates/typst-library/src/foundations/content.rs#L451-L479

That means your `within` query will also find things that are perhaps unexpected, because they are not displayed but just helper fields.
For example, `heading` and other elements have a `supplement` field, and you will also search in that field.
E.g. consider the following code:
```typ
#set heading(supplement: [_Mini-section_ #label("emphstuff")])

= My heading with something _emphasized_ #label("emphstuff")

#context query(selector(<emphstuff>).within(heading))
```
It will find the "Mini-section" thing, even though it is not present (i.e. displayed) anywhere in the document, in particular it's not never displayed in a `heading` (only in references).

crates/typst-library/src/introspection/introspector.rs
schrottkatze: ah right, missed that when doing renaming :)

thanks!

laurmaedje:
Yeah, that's indeed a problem. We discussed this a bit on Discord, but I'm also posting this here for posterity:

- The `Content::query` function shouldn't be used. It's more of a hack that was introduced at some point for figures and I made the concession to have it there. But it's significantly nerfed compared to the proper `Introspector::query` as it will query in the raw content and thus can't see through show rules, context, or anything like that. For `within` to work as expected, it would need to make use of the `Start` and `End` tags to reason about nesting.

- Compared to `before` and `after`, the `within` selector is also one that's much easier to support in show rules, so that would be interesting as well. That's where the bindings thing and the design complexity comes into play (i.e. you might want to write a show rule for a caption in a figure and have as bound variables both the figure and the caption).

We should probably add some discouraging docs on `Content::query` and maybe indeed rename it.

laurmaedje:
I think docs can wait for now

schrottkatze:
ye :3 
got distracted by life stuff (and the daunting task of having to read and understand enough of the codebase, to understand the internals well enough to be able to implement it without causing any problems), will be getting back to this soon-ish (planning on next few weeks, gotta still catch up with some uni stuff ^^')

laurmaedje:
Hey @schrottkatze, are you still planning to work on this? If not, no worries, I would then just close the PR.

schrottkatze:
Yes, I'm still planning to. Life's just kinda a mess atm, and it doesn't exactly help that my profs seem to have not talked to eachother when coordinating how big to make the semester projects, so I'm kinda busy with that atm. My semester is over in august so I should finally be having time then again :) (if my life doesn't further blow up on me lol)

laurmaedje:
Alright! 

> Life's just kinda a mess atm

I'm sorry to hear that. :/

laurmaedje:
I thought some more about this PR. While I'd like to support a selector like this, ideally I'd want to do it across both query _and_ show rules, and also with the capability to access fields of the matched parent element. Especially the last part is very complex (also for me), and needs a clear plan. I'm optimistic we can support this down the road, but it requires more design work. If you'd still like to contribute once your university things settle down a bit, I'd suggest starting with some simpler. Thanks for understanding!

schrottkatze:
Oki! Will look into it, also got quite a lot of personal projects to do ^^

PR #5877, opened 2025-02-16
Bring back type/str compatibility for 0.13, with warnings and hints

laurmaedje:
The PR https://github.com/typst/typst/pull/5591 removed various deprecated things from the compiler. Most things removed in that PR already warned in 0.12. An exception to this is the compatibility behaviour that lets `int == "integer"` evaluate to true since Typst 0.8. This behaviour was removed with 0.13-rc1 without any previous deprecation warnings (except in the docs and changelog). The reason I didn't implement warnings in the first place was that it wasn't possible to emit warnings in all cases (because nested comparisons go through `PartialEq for Value` and there's no way to pass a warning context) and the code for the cases that we can catch is quite ugly. I figured that, as it's superseded since Typst 0.8, removing it in 0.13 was fine.

However, @tingerrr demonstrated that [quite a large portion of the ecosystem did not migrate in time](https://github.com/typst/packages/issues/1725), leading to widespread breakage. For this reason, I'm temporarily bringing back the compatibility behaviour, but this time with warnings and hints for the cases that we _can_ catch. 

This is what the warning looks like that most people will see:
```
warning: comparing strings with types is deprecated
  ┌─ hi.typ:2:2
  │
2 │   type(10) == "integer"
  │   ^^^^^^^^^^^^^^^^^^^^^
  │
  = hint: compare with the literal type instead
  = hint: this comparison will always return `false` in future Typst releases
```

There are also warnings for 
- adding types and strings
- joining types and strings
- type in array
- type in str
- type in dictionary

----

Note that this PR is based on the 0.13 branch and will be merged into it. The temporary revival with warnings is for Typst 0.13 specifically. It won't be in the dev version and also not in Typst 0.14. (There isn't much of a point in merging this to main and then removing it immediately after the release, which is what I'd have done otherwise.)

### DIFF
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -55,7 +55,7 @@ fn eval_code<'a>(
             _ => expr.eval(vm)?,
         };
 
-        output = ops::join(output, value).at(span)?;
+        output = ops::join(output, value, &mut (&mut vm.engine, span)).at(span)?;
 
         if let Some(event) = &vm.flow {
             warn_for_discarded_content(&mut vm.engine, event, &output);
--- a/crates/typst-eval/src/flow.rs
+++ b/crates/typst-eval/src/flow.rs
@@ -83,7 +83,8 @@ impl Eval for ast::WhileLoop<'_> {
             }
 
             let value = body.eval(vm)?;
-            output = ops::join(output, value).at(body.span())?;
+            let span = body.span();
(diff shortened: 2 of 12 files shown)


tingerrr:
Thanks for taking action so quickly! The good news is that the biggest two packages, cetz and touying were both false positives and with luck many of the other packages are too!

PR #5878, opened 2025-02-16
More robust SVG auto-detection

laurmaedje:
In 0.12, when passing a string to `image.decode`, it's always assumed to be an SVG. The new `image(bytes(..))` code depends on format auto-detection as the string case always means a path. We don't have that extra bit of information. However, the current auto-detection is quite naive and just checks `.starts_with("<svg")`. 

The deprecated `image.decode` now internally uses the same auto-detection, instead of checking the type, so `image.decode(mystring)` where mystring was a valid SVG but did not start with `<svg` regressed. This PR improves the auto-detection to fix that regression. 

The new check is whether the first 2048 bytes contain the string `http://www.w3.org/2000/svg`. Any valid SVG must declare a namespace and contain this. `usvg` will complain about a missing root note if it's not provided, so any files that previously worked with `image.decode` must contain it. The only case where it could potentially regress now is if the attribute is not within the first 2048 bytes or somehow encoded in a non-standard way, both of which seem highly unlikely.


laurmaedje:
Todo: Test

PR #5884, opened 2025-02-18
HTML export: Use `<code>` for inline `RawElem`

### LINKED ISSUE #5881: HTML Export: Incorrect linebreak in heading with raw/monospace element
Version: 0.13.0-rc1

I am experimenting with the new HTML export.  I understand this feature is extremely new, but hopefully this is low hanging fruit to catch before the full 0.13 release.

Using the below code causes extra linebreaks in the HTML output as opposed to the intended behavior of the PDF output.

```
= This is a heading with a #raw("raw") element

#lorem(10)
```

HTML Output:
Image

PDF Output:
Image

OverflowCat:
Fix #5881

### DIFF
--- a/crates/typst-library/src/text/raw.rs
+++ b/crates/typst-library/src/text/raw.rs
@@ -446,10 +446,14 @@ impl Show for Packed<RawElem> {
         let mut realized = Content::sequence(seq);
 
         if TargetElem::target_in(styles).is_html() {
-            return Ok(HtmlElem::new(tag::pre)
-                .with_body(Some(realized))
-                .pack()
-                .spanned(self.span()));
+            return Ok(HtmlElem::new(if self.block(styles) {
+                tag::pre
+            } else {
+                tag::code
+            })
+            .with_body(Some(realized))
+            .pack()
+            .spanned(self.span()));
         }
 
         if self.block(styles) {


laurmaedje:
I changed the base branch to main since the change should persist beyond 0.13. The PR https://github.com/typst/typst/pull/5877 which merged directly into 0.13 was an exception because the code shouldn't be in main.

I'll separately cherry-pick to 0.13 after merge.

laurmaedje:
Looking at [this Stackoverflow discussion](https://stackoverflow.com/questions/4611591/code-vs-pre-vs-samp-for-inline-and-block-code-snippets), I'm not 100% certain that these are this is the ideal representation, but in any case it's better than what we've got now. We can always revisit this later.

Thanks!

laurmaedje:
What do you mean with "Typst doesn't seem to have a way to specify HTML _tags_"? You mean that the raw elements has no further metadata for this?

I was primarily wondering whether it should be `<pre>..</pre>`  or `<pre><code>..</code></pre>`.

And then there's also the problem that, unlike Typst's inline raw element, `<code>` will collapse whitespace. But the whole whitespace stuff is a bit wonky all over the HTML export right now.

bastienvoirin:
> And then there's also the problem that, unlike Typst's inline raw element, `<code>` will collapse whitespace. But the whole whitespace stuff is a bit wonky all over the HTML export right now.

What about hardcoding inline CSS until we find a better solution?
```html
<code style="white-space: pre-wrap;"></code>
```

```typ
// Pseudo-code, to be wrapped in a contextual block
#if target == "html" {
  show raw.where(block: false): set html.elem(
    "code",
    attrs: (style: "white-space: pre-wrap")
  )
}
```

---

> Based on my [discussion](https://t.me/typst_zh/2178) with @Myriad-Dreamin, currently Typst doesn't seem to have a way to specify HTML tags. This prevents us from distinguishing whether the user's `raw` content is `code` or `samp` (in addition to `hr`, `ruby`, etc. - HTML export doesn't seem to pay much attention to semantic HTML due to typst's lack of those elements). Also, these tags are rarely distinguished even in the Markdown ecosystem (I once wrote a [remark plugin](https://github.com/OverflowCat/remark-samp-kbd) for this), and mainly rely on the browser's user agent stylesheet. Since CSS generation is currently unavailable, I believe this is a good practice for users.

I agree that it would be nice to eventually be able to emit `code`, `samp`, and `kbd` tags, but personally I wouldn't mind using a Typst package implementing something along the lines of [this example taken from the 0.13-rc1 documentation](https://staging.typst.app/docs/reference/foundations/target/#example):

```typ
#let kbd(it) = context {
  if target() == "html" {
    html.elem("kbd", it)
  } else {
    set text(fill: rgb("#1f2328"))
    let r = 3pt
    box(
      fill: rgb("#f6f8fa"),
      stroke: rgb("#d1d9e0b3"),
      outset: (y: r),
      inset: (x: r),
      radius: r,
      raw(it)
    )
  }
}

Press #kbd("F1") for help.
```

And if I understand correctly, the upcoming types & elements overhaul should make this even easier.

OverflowCat:
Oh, my bad, I didn't notice that there is a `html.elem`.

The code in the `<pre>` tag is another issue — it seems that every Markdown renderer is doing it this way for now. There will be even more problems if we are going to output syntax-highlighted code in the future.

PR #5887, opened 2025-02-18
`line::commit`: Sort line items by logical order when constructing frame

### LINKED ISSUE #5775: Order of references are wrong in right to left language
### Description

In this example, everything is ok:
```typst
This is
@Ref1
and this one is
@Ref2

#bibliography("repro.bib")
```

Image

But if I use Farsi, which is a right to left language, the number of references is reversed and Ref2 takes the number 1:
```typst
این است
@Ref1
و این یکی هست
@Ref2

#bibliography("repro.bib")
```

Image

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

bluebear94:
Fixes #5775.

The order in which elements are exposed during introspection depends on the order of tags within frames. Since the items in a line were previously collected in visual, LTR order, this led to citation elements being visited in the wrong order in subsequent iterations when RTL text was present. This PR stores the logical index of each item within an `Items` and uses that index to sort frame elements by logical order when committing a line.


crates/typst-layout/src/inline/line.rs
```diff
@@ -559,8 +561,9 @@ pub fn commit(
         add_par_line_marker(&mut output, marker, engine, locator, top);
     }
 
+    frames.sort_by_key(|(_, _, idx)| *idx);
```
laurmaedje: can it happen that the sorting is something other than either a no-op or a reversal?

bluebear94: Yes, this can happen when there’s mixed LTR and RTL text. The following assertion fails for several tests:

```rust
assert!(
    frames.is_sorted_by_key(|(_, _, idx)| idx)
        || frames.is_sorted_by_key(|(_, _, idx)| std::cmp::Reverse(idx))
);
```

crates/typst-layout/src/inline/line.rs
```diff
@@ -697,7 +710,18 @@ impl Debug for Items<'_> {
     }
 }
 
+/// An item accompanied by its position within a line.
+#[derive(Debug)]
+pub struct IndexedItemEntry<'a> {
```
laurmaedje: I think this isn't quite pulling its weight. I would prefer just having `Vec<(usize, ItemEntry<'a>)>` in `struct Items`.

crates/typst-layout/src/inline/line.rs
laurmaedje: Makes sense. Could you extend the citation test to include mixed LTR and RTL (or add a second one) to exercise this?

laurmaedje: Also, we can't have equal indices here across different items, can we? So I think we can use an unstable sort to avoid a potential memory allocation. This is on the hot path after all.

crates/typst-layout/src/inline/prepare.rs
```diff
@@ -3,6 +3,12 @@ use unicode_bidi::{BidiInfo, Level as BidiLevel};
 
 use super::*;
 
+#[derive(Debug)]
```
laurmaedje: The `Run` change seems quite unrelated. We can discuss whether this is worthwhile, but should do so in another PR.

crates/typst-library/src/model/bibliography.rs
laurmaedje, line 1: These changes are also unrelated. The typo fixes could go into a separate PR. As for the `Debug` derive: The reason I avoid those on internal structs is because I think that many derives can hurt compile times, though I haven't actually benchmarked this tbh. But compile time of typst-library is definitely a problem.

crates/typst-layout/src/inline/prepare.rs
laurmaedje: This does not seem fully addressed yet. I meant the `Run` struct as a whole, i.e. the changes in `prepare.rs` which don't seem necessary for the bug fix.

laurmaedje:
Thanks for doing the reverts. There's still one change left, which I would also prefer to discuss separately: https://github.com/typst/typst/pull/5887#discussion_r2166399165

laurmaedje:
Thank you!

Issue #5790, opened 2025-01-31
Rounding error when working with lengths


MDLC01:
### Description

This is a follow up to https://github.com/typst/typst/issues/3398#issuecomment-2626988487.

Consider the following Typst function `foo`. It accepts two arguments and lays them out side by side in a horizontal stack. However, if there is not enough horizontal space available for both `x` and `y`, `foo` panics.

```typ
#let foo(x, y) = {
  layout(available => {
    assert(measure(x).width <= available.width - measure(y).width)
    stack(dir: ltr, x, y)
  })
}
```

For example, the following piece of code compiles properly (as `1cm <= 10cm - 1cm`)...

```typ
#block(
  width: 10cm,
  foo(h(1cm), h(1cm)),
)
```

... but this one panics (as `1cm > 1.5cm - 1cm`)

```typ
#block(
  width: 1.5cm,
  foo(h(1cm), h(1cm)),
)
```

Moreover, setting the width of the block to `auto` also works without panicking:

```typ
#block(
  width: auto,
  foo(h(1cm), h(1cm)),
)
```

Let us now define `bar` as follows. Note that `epsilon` is the smallest representable positive value.

```typ
#let epsilon = 5e-324pt
#let bar = foo(h(epsilon), h(1.0pt))
```

As expected, laying out `bar` in a block with `auto` width works without panicking:

```typ
#block(width: auto, bar)
```

However, replacing the width of the block with that of `bar` results in a panic!

```typ
#context block(width: measure(bar).width, bar) // Panics!
```

Adding a slight amount to the width of the block to account for the rounding error fixes the issue:

```typ
#context block(
  // Could bisect further.
  width: measure(bar).width + 0.000000000000002pt,
  bar,
)
```

This is likely because of a precision loss due to the sum of the widths of `x` and `y` being represented with a higher exponent than their individual widths: the individual widths of `x` and `y` have an exponent of 3, while their sum has an exponent of 4.[^1] Therefore, when adding the two widths, some precision is lost.

[^1]: As determined using https://lukaskollmer.de/ieee-754-visualizer/.

Essentially, this boils down to the following system of inconsistent equations, which encapsulate a known limitation of floating points in general.

```math
\begin{dcases*}
\varepsilon \neq 0 \\[1ex]
1 + \varepsilon = 1
\end{dcases*}
```

### Typst version

- [x] I am using the latest version of Typst

MDLC01:
[...]

Issue #3605, opened 2024-03-10
Propagation of alignment with `align`


Mc-Zen:
Currently, an alignment specified in `align()` does not only change the alignment of its child within the container, but also applies this alignment recursively to nested children (except for explictly aligned content). 

This leads to an awkward behaviour when using `align()`, e.g., through `figure()`, with alignment-sensitive content. Especially changing figure alignment is a layout decision that can often happen at the end or be done by an editor and it should ideally not affect the content itself. 

Also the current behaviour leads to potential bugs in packages that may easily go unnoticed. In any case, package authors should protect alignment-sensitive content in some layer of `align(top + left)` or similar because there might be `set align()`s lurking in the wild.

Issue #3398, opened 2024-02-11
Unexpected word break in padded heading


MDLC01:
### Description

I'm not sure if this issue should go to hypher or here.

In the following document, the first heading is broken, even though there is clearly enough space for it not to break (as demonstrated by the fact that making it longer causes it to no longer break).

### Reproduction

```typ
#set text(font: "Fira Sans")
#set par(justify: true)

#show par: pad.with(left: 0.3cm)

= Diversity

= Diversity.
```

In Typst 0.11.1, this renders as

![Diver- // sity // Diversity.](https://github.com/user-attachments/assets/6aa914fa-052c-4cf8-b67e-1f265c72dd62)

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
This is the right repository. It's pretty surely not related to hypher, but rather to measurement in block layout.

laurmaedje:
Can't reproduce this in 0.12, but I'm not sure it's actually fixed.

MDLC01:
I think this is probably an issue related to floating point approximations.

I'm not sure whether the issue is fixed with headings specifically, but I have a similar problem with grids, which uses the Curryst package (specifically, version 0.3.0). In the example below, the block with `auto` width contains the content properly, but when using a grid, `auto` now forces a line wrap (Curryst queries the available width with `layout` to determine whether there should be multiple lines, so `auto` should not create a line break). Similar issue when explicitly using `measure`.

```typ
#import "@preview/curryst:0.3.0": proof-tree, rule

#show grid: block.with(stroke: 0.5pt + red)

#let tree = proof-tree(rule(name: [djfj], [], `t`, `t`, `t`))

#block(
  stroke: 0.5pt + red,
  width: auto,
  tree,
)

#grid(
  columns: (auto,),
  tree,
)

#context grid(
  columns: (measure(tree).width,),
  tree,
)

#context grid(
  // Could bisect further.
  columns: (measure(tree).width + 0.0000000000000071613756045112467pt,),
  tree,
)
```

![Image](https://github.com/user-attachments/assets/fe03c52c-2bf1-4d38-805b-418ce154c2ec)

MDLC01:
If I have the time, I may try to find a reproduction that does not rely on Curryst. But I don't think this is even necessary: we know that floating point approximations are an issue, and that we could switch to fixed point.

laurmaedje:
[...]

PR #5889, opened 2025-02-19
Add `align` parameter to `stack`

### LINKED ISSUE #5633: Add `align` parameter to `stack`
### Description

Similarly to [`grid`](https://typst.app/docs/reference/layout/grid/#parameters-align), I suggest to add an `align` parameter to `stack`, which would determine the alignment of the alignment of items within the stack.

### Use Case

Easier to use than
```typ
#stack(..items.map(align.with(..)))
```

A drawback is the accumulation of similar arguments on functions like `grid`, `table`, and `stack`.

MDLC01:
Closes #5633.

This PR adds an `align` parameter to `stack`. When set to `auto` (the default), the elements are not aligned in a specific way (like before). Otherwise, each element is aligned accordingly. This behavior mimics that of `grid`'s `align`.


EpicEricEE:
Currently, `stack(align: center, ..)` would be the same as `align(center, stack(..))`, right? I would have expected the alignment to only be relative to the stack itself, similar to how `block(align(center, stack(..)))` behaves, and not also relative to the parent container.

MDLC01:
Currently, `stack(align: center, ..)` aligns all items. In other words, `stack(align: center, ..items)` is equivalent to `stack(..items.map(align.with(center)))`, and similarly for other alignments.

laurmaedje:
I _can_ see this, but I'm not sure. Generally, I'm unsure what `stack`'s place in Typst is/should be. Right now, it's kinda a ... more limited, worsely implemented single-column/row grid. Alignment is also a bit all over the place. Which maybe is inevitable, but maybe not. 

To summarize: In isolation, this change is probably sensible. What I don't know is whether it fits with a greater vision as a greater vision for alignment improvements is currently somewhat lacking. There is also the problem that alignment [should not always propagate](https://github.com/typst/typst/issues/3605).

MDLC01:
I agree with you about `stack`'s place in Typst.

I will note that, contrary to `grid`, `stack` does not exhibit the rounding issue mentioned in https://github.com/typst/typst/issues/3398#issuecomment-2622614733 (i.e., replacing the `grid` with a `stack` yields the same result as using a block). For a minimal reproduction, adapted from https://github.com/typst/typst/issues/5790:

```typ
#let foo(x, y) = {
  layout(available => {
    assert(measure(x).width <= available.width - measure(y).width)
    stack(dir: ltr, x, y)
  })
}

#let epsilon = 5e-324pt
#let bar = foo(h(epsilon), h(1.0pt))

#block(width: auto, bar) // Works as intended
#stack(bar) // Works as intended
#grid(columns: (auto,), bar) // Panics
```

laurmaedje:
I thought some more about this and decided that I'd like to have a clearer vision for how we'll evolve both stack and alignment before making API changes in this direction. Probably, stack should evolve more into a flexbox direction and also support wrapping around ... As for alignment, it also needs a bit of a rethink I think, especially a way to not propagate.

PR #5893, opened 2025-02-19
Add more methods to `direction`

### LINKED ISSUE #5815: `sign` method on `direction`
### Description

If `dir` is a `direction`, `dir.sign()` would return `1` if `dir` is `ltr` or `ttb`, and `-1` if it is `rtl` or `btt`.

### Use Case

This can be useful when placing elements using `place`. For example, the following function does not work when using right-to-left text: in the second use case, "Ho" is outside of the page (to the right).

```typ
#set page(width: 10cm, height: 5cm, margin: 1cm)

#let end-margin(body) = {
  place(
    dx: 100%,
    body
  )
}

#set text(dir: ltr)
#end-margin[Hey]

#set text(dir: rtl)
#end-margin[Ho]
```

Only "Hey" is visible, in the right margin.

The right way to implement that is as follows.

```typ
#let end-margin(body) = context {
  let sign = if text.dir == ltr { 1 } else { -1 }
  place(
    dx: 100% * sign,
    body
  )
}
```

Using this function, we get the desired result:

"Hey" appears in the right margin. "Ho" appears in the left margin.

The `if` statement is essentially a partial implementation of the suggested `direction.sign()` method.

### LINKED ISSUE #5892: `from` and `to` constructors for `direction`
### Description

`direction.from` and `direction.to` constructors that are the reverse functions of `direction.start` and `direction.end` respectively. In other words, `direction.from(right)` is `rtl`, `direction.to(bottom)` is `ttb`, etc..

### Use Case

Easier manipulation of directions.

MDLC01:
This PR closes #5815 by adding a `direction.sign`. I arbitrarily chose to return an integer instead of a float.

Additionally, it closes #5892 by adding `direction.from` and `direction.to` constructors.

Given the mixed GitHub reactions on #5815, I can remove `direction.sign` from this PR if wanted.

### DIFF
--- a/crates/typst-library/src/layout/dir.rs
+++ b/crates/typst-library/src/layout/dir.rs
@@ -65,6 +65,22 @@ impl Dir {
         }
     }
 
+    /// The corresponding sign, for use in calculations.
+    ///
+    /// ```example
+    /// #ltr.sign() \
+    /// #rtl.sign() \
+    /// #ttb.sign() \
+    /// #btt.sign()
+    /// ```
+    #[func]
+    pub const fn sign(self) -> i64 {
+        match self {
+            Self::LTR | Self::TTB => 1,
+            Self::RTL | Self::BTT => -1,
+        }
+    }
+
     /// The start point of this direction, as an alignment.
     ///
     /// ```example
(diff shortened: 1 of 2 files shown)


wrzian:
I like `direction.sign` 👍

PR #5896, opened 2025-02-19
Add snap installation option to README

andis854:
I saw that Typst can be downloaded via snap in Ubuntu, so I added this to the description in README under installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,9 @@ Typst's CLI is available from different sources:
   - macOS: `brew install typst`
   - Windows: `winget install --id Typst.Typst`
 
+- Snap users can install Typst with
+  `sudo snap install typst`
+
 - If you have a [Rust][rust] toolchain installed, you can install
   - the latest released Typst version with
     `cargo install --locked typst-cli`


laurmaedje:
I don't really want to accumulate a bullet point for every way you can install it on Linux. It's a bit unfortunate that it's [not shown on Repology.](https://repology.org/project/typst/versions)

andis854:
Thanks for your respond. I understand that! However, the information should be available somewhere as this seems to be the easiest way to install on Ubuntu.

laurmaedje (commented):
I think with something like this, I'd be okay:
```
- You can install Typst through different package managers. Note that the
  versions in the package managers might lag behind the latest release.
  - Linux
    - View [Typst on Repology][repology]
    - View [Typst's Snap][snap]
  - macOS: `brew install typst`
  - Windows: `winget install --id Typst.Typst`

...
[snap]: https://snapcraft.io/typst
```

andis854:
Yes, looks good!

laurmaedje:
Feel free to update your PR to that effect

PR #5905, opened 2025-02-19
Fix high CPU usage due to inotify watch triggering itself

aodenis:
On Linux, inotify is used to monitor filesystem changes. inotify events are then filtered to check whether they are relevant and must trigger a recompilation, but this check opens the file, which trigger an inotify event, creating a busy loop.

This commit swaps the order of the checks done in `is_event_relevant` in order to filter out the open events caused by `is_same_file` without triggering new ones.

### DIFF
--- a/crates/typst-cli/src/watch.rs
+++ b/crates/typst-cli/src/watch.rs
@@ -237,7 +237,7 @@ impl Watcher {
 
     /// Whether a watch event is relevant for compilation.
     fn is_event_relevant(&self, event: &notify::Event) -> bool {
-        let type_relevant = match &event.kind {
+        let kind_relevant = match &event.kind {
             notify::EventKind::Any => true,
             notify::EventKind::Access(_) => false,
             notify::EventKind::Create(_) => true,
@@ -252,12 +252,20 @@ impl Watcher {
             notify::EventKind::Other => false,
         };
 
+        if !kind_relevant {
+            return false;
+        }
+
         // Never recompile because the output file changed.
-        type_relevant
-            && !event
-                .paths
-                .iter()
-                .all(|path| is_same_file(path, &self.output).unwrap_or(false))
+        if event
+            .paths
+            .iter()
+            .all(|path| is_same_file(path, &self.output).unwrap_or(false))
+        {
+            return false;
+        }
+
+        true
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/typst-cli/src/watch.rs
```diff
@@ -259,7 +250,14 @@ impl Watcher {
             },
             notify::EventKind::Remove(_) => true,
             notify::EventKind::Other => false,
-        }
+        };
+
+        // Never recompile because the output file changed.
+        type_relevant
+            && !event
```
laurmaedje, line 257: Relying on short-circuit here is a bit too subtle in my opinion. I'd suggest using `if !type_relevant { return false; }` above and adding a comment explaining that the order matters.

laurmaedje: Also `kind_relevant` might be slightly better than `type_relevant`

aodenis: agreed on both points, I'm fixing this

laurmaedje:
After further thinking, I'm not convinced we need the `is_same_file` check at all anymore. It's a remnant from times where we [watched the whole root directory recursively.](https://github.com/typst/typst/blob/b3faef4b80a674294091066e20501e3a5d0f6103/cli/src/main.rs#L239-L241) Nowadays, the CLI watches every path separately, so saving the PDF shouldn't trigger an event at all.

That is, unless one does `read("output.pdf", encoding: none)`, but then one really asked for it and compiling in a loop doesn't seem like technically wrong behaviour to me then. Notably, the current check is not effective anyway for multipage PNG and SVG output, as `output` will contain a page number template.

aodenis:
I removed the check. I did not remove same-file from cargo manifests as `cargo remove` would change the manifests too much.

laurmaedje:
Why not remove it by hand?

aodenis:
I hadn't seen it was a dependency of walkdir, that's why `cargo remove` removed too many stuff

laurmaedje:
I discussed with @PgBiel and he considered it too big of a footgun that the CLI could end up in a loop when one reads the output file. So I reverted that part.

The output check does not work properly for multifile image export, but that's not a regression and can be considered separately. Dealing with that properly is a bit complicated.

Thanks for the investigation here!

Issue #5756, opened 2025-01-25
Unable to use `html.elem("body")` inside (conditional) `context`


Andrew15-5:
### Description

Currently, the multi-output config like below doesn't work:

```typ
#context if target() != "html" {
  [document-body]
} else {
  html.elem("body", html.frame([document-body]))
}
```

[This was confirmed by Laurenz](https://discord.com/channels/1054443721975922748/1088371867913572452/1332683539736690779) (that this is a bug). The body should be able to be contextual and work in the above example.

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

Andrew15-5:
I made a workaround for this: https://github.com/typst/typst/pull/5755.

PR #5910, opened 2025-02-20
HTML export: fix elem counting on classify_output

SharzyL:
The following code will not compile:
```typst
// a.typ
#context {
  if target() == "html" {
    html.elem("html")
  }
}
```

```console
$ typst c --features html a.typ a.html
warning: html export is under active development and incomplete
 = hint: its behaviour may change at any time
 = hint: do not rely on this feature for production use cases
 = hint: see https://github.com/typst/typst/issues/5512 for more information

error: `<html>` element must be the only element in the document
  ┌─ ../../../../tmp/a.typ:4:4
  │
3 │     html.elem("html")
  │     ^^^^^^^^^^^^^^^^^
```

Because currently the HTML exporter checks whether the `<html>` element is the sole element in the document. However, when `<html>` is wrapped in a `context` block, the `context` tag and the end of `context` are also counted, resulting in the element count being 3 instead of 1.

We fixed the problem by only counting elements being an `HtmlNode::Element`.

Fixes https://github.com/typst/typst/issues/5756

### DIFF
--- a/crates/typst-html/src/lib.rs
+++ b/crates/typst-html/src/lib.rs
@@ -307,21 +307,18 @@ fn head_element(info: &DocumentInfo) -> HtmlElement {
 
 /// Determine which kind of output the user generated.
 fn classify_output(mut output: Vec<HtmlNode>) -> SourceResult<OutputKind> {
-    let html_elem_count = output
-        .iter()
-        .filter(|node| matches!(node, HtmlNode::Element(_)))
-        .count();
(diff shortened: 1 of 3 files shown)


SharzyL:
Kindly ping @laurmaedje

laurmaedje:
I changed this so that only tags are ignored. Text and SVG frames must also not occur next to `<html>` or `<body>`. I also added a test.

Thanks!

Fixes https://github.com/typst/typst/issues/5756

Issue #5377, opened 2024-11-04
Table subheaders and subfooters


PgBiel:
## Feature

Depends on #5375

Have the ability to specify "levels" of headers and footers, such that a table can be split into multiple sections (each section has a top-level header), and each section may have its own "internal" headers and footers (subsections) which repeat together with the top-level header / are also "sticky". They'd be repeated until the next top-level header, or the next subheader under the same top-level header.

For example, you'd have something like

```typ
#table(
  columns: 2,
  // First section
  table.header(table.cell(colspan: 2)[*Europe*]),
  table.subheader([*Name*], [*Age*]),
  [Peter], [55],
  [Mary], [48],
  table.subheader([*Name*], [*Preference*]),
  [Joseph], [Banana],
  [Jane], [Apple],

  // Second section
  table.cell(colspan: 2)[*South America*],
  table.subheader([*Name*], [*Age*]),
  [Pedro], [55],
  [María], [48],
  table.subheader([*Name*], [*Preference*]),
  [José], [Grape juice],
  [Robert], [Strawberry],
)

```

Which, if broken apart, would generate the following output:

![headers and subheaders repeated together](https://github.com/user-attachments/assets/239d3158-851b-4762-a2d0-70e9c674d08e)

### Use Case

Have the flexibility to create table subsections. (The proposed syntax could also be changed to reflect that.)

MDLC01:
Shouldn't South America be a `table.header` instead of a `table.cell`?

PgBiel:
Yes, whoops, thank you :)

Issue #5375, opened 2024-11-04
Headers in arbitrary positions


PgBiel:

Andrew15-5:
I think my goal is different. I want to be able to create a single header by gluing together all the header cells that are separate. 2 use cases: when you need to have parts of the header separate (when zipping columns), and when you want to make the first column as the header.

I currently have the 1st situation and can't use multiple headers. Forceful inclusion of all the header cells as a separate argument to the `table()` will defeat the whole point in zipping by columns and will make the table less readable/maintainable.

PgBiel:
This issue is exclusively about being able to "replace" the previous header with another one.

If your intention is _exclusively_ to join rows into the header, that seems a bit more pathological imo, since data will be flattened into rows anyway, so there's always a step where you can just .slice(). And using the header constructor is important as we may change its semantics in the future to fix pathological cases where cells outside it are unintentionally repeated.

With that said, #5377 would probably do what you're looking for, though its main use case is when the main header is a row with a large colspan and it is followed by a row (subheader) with more "specific" headers. If this is your case, then #5377

Column-wide headers are entirely unrelated and are better tracked in a separate issue.

Andrew15-5:
> If this is your case

It's not.



> (**Edit:** It's probably this #4707)

It is? Doesn't look like for me.



> Column-wide headers are entirely unrelated and are better tracked in a separate issue.

Semantically they are identical, it's just a line of cells. So gluing items vertically or horizontally in a single header line doesn't make much of a difference besides the direction.

I'll open a separate issue, since nothing fits the bill.

Andrew15-5:
There it is: https://github.com/typst/typst/issues/5546.

PR #5919, opened 2025-02-21
Refactor grid header and footer resolving

### LINKED ISSUE #5359: Fixing cell column within `table.header` and `table.footer` causes weird behaviour.
### Description

By specifying the location of cells within the `table.header` and `table.footer`, it appears possible to inject cells into the header and footer respectively. Cells that are before these `header` and `footer` functions are injected without raising errors, and also cause layout errors (although the layout might not be properly defined in these cases)

Source code:
```typ
#table(
  columns: 3,
  // This cell should throw an error, or not be in the header
  [Incorrect], 
  table.header(
    [A], table.cell(x: 1)[Forced position], [B]
  ),
  // These cells are the body
  [1],[2],[3],
  [4],[5],[6],
  [4],[5],[6], 
  [Incorrect],
  // The cell above is still in the body,
  // and should not be in the footer
  table.footer(
    [X],[Y],[Z],
    table.cell(x: 1)[Forced position]
  ),
)
```
Note, the page height is set small here to demonstrate the repeated headers etc.

joist-1
joist-2
joist-3

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

PgBiel:
(Draft as there's some polishing still missing, but keeping it open to track progress so far.)

Fixes #5359

This refactors grid resolving - that is, the step before grid and table layout where we parse the parameters, place cells in their appropriate locations and generate the final grid - in preparation for #5375 and #5377 to ensure the code can handle any amount of arbitrary _row groups_, that is, table children which aren't cells directly but span groups of entire rows, which right now only include headers and footers.

This also adjusts how manually-positioned cells behave with headers and footers. Before, you could place a cell before the header and then, even though the header would normally try to occupy the following row (so as to not conflict with that existing cell), you could force the header to occupy the first row as well with `grid.cell(y: 0)[...]` inside the header. Instead of throwing an error, Typst would just happily accept that the cell outside of the header is now part of header, which is very unlikely to be the intention. So, headers and footers **can no longer expand to include cells outside them.** Instead, they will try to skip them, and cells outside headers and footers which don't pick their rows will always skip headers and footers, avoiding conflicts.

In other words, conflicts between headers and footers and cells outside them are now only possible when you have cells with fixed `y`, in which case this should always be an error now. This fixes #5359.

## TODO

~~Mostly just doing some final polishing before marking the PR as ready for review. There are some TODOs in the code, I'll tackle some now and the rest should be tackled in further PRs.~~


crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -1615,25 +1608,190 @@ impl<'a> CellGrid<'a> {
...
+                        "cell would cause header or footer to expand to non-empty row {new_y}";
+                        hint: "try moving its cells to later rows"
+                    );
+                } else {
+                    bail!(
+                        "cell would cause header or footer to expand to non-empty row {new_y}";
+                        hint: "try moving its cells to earlier rows"
+                    );
+                }
+            }
```
PgBiel, line 1723: A bit unsure about this error hint, thinking of just generically asking to move the header or footer, or the existing cells

laurmaedje (commented):
To be honest, I'm not familiar enough with the old code to properly review this change. I'd trust you on the change, but if you want we can also have a quick call where you walk me through the changes. Just trying to understanding everything from scratch by myself would be a bit too time-consuming right now.

PgBiel:
> To be honest, I'm not familiar enough with the old code to properly review this change. I'd trust you on the change, but if you want we can also have a quick call where you walk me through the changes. Just trying to understanding everything from scratch by myself would be a bit too time-consuming right now.

Yeah sure, we can do that, even if just for a quick sanity check.

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -1648,29 +1806,43 @@ fn resolve_cell_position(
...
-                while let Some(Some(_)) = resolved_cells.get(resolved_index) {
+            loop {
+                if let Some(Some(_)) = resolved_cells.get(resolved_index) {
                     // Skip any non-absent cell positions (`Some(None)`) to
                     // determine where this cell will be placed. An out of
                     // bounds position (thus `None`) is also a valid new
                     // position (only requires expanding the vector).
                     resolved_index += 1;
+                } else if let Some(header) =
+                    header.filter(|header| resolved_index / columns < header.end)
```
laurmaedje, line 1822: maybe we could express some `/` with `*` instead for some extra safety

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -1615,25 +1608,190 @@ impl<'a> CellGrid<'a> {
...
+                    bail!(
+                        "cell would cause header or footer to expand to non-empty row {new_y}";
+                        hint: "try moving its cells to later rows"
+                    );
+                } else {
+                    bail!(
+                        "cell would cause header or footer to expand to non-empty row {new_y}";
+                        hint: "try moving its cells to earlier rows"
+                    );
+                }
```
laurmaedje, line 1722:
```suggestion
                bail!(
                    "cell would cause header or footer to expand to non-empty row {new_y}";
                    hint: "try moving its cells to {} rows",
                    if group_range.is_none_or(|r| new_y < r.start) { "later" } else { "earlier" },
                );
```

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -733,6 +736,13 @@ impl<'a> CellGrid<'a> {
         // Therefore, we use a counter, 'auto_index', to determine the position
         // of the next cell with (x: auto, y: auto). It is only stepped when
         // a cell with (x: auto, y: auto), usually the vast majority, is found.
+        //
```
laurmaedje, line 739: as discussed, it would be nice to attempt to extract one (or maybe multiple) functions here as the `resolve` function is really large

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -1648,29 +1806,43 @@ fn resolve_cell_position(
...
-                resolved_index =
-                    find_next_empty_row(resolved_cells, resolved_index, columns);
 
-                // Next cell won't have to start a new row if we just did that,
-                // in principle.
-                *start_new_row = false;
-            } else {
-                while let Some(Some(_)) = resolved_cells.get(resolved_index) {
+            loop {
+                if let Some(Some(_)) = resolved_cells.get(resolved_index) {
```
PgBiel, line 1815: Should de-duplicate this loop and the one below

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -1692,24 +1864,76 @@ fn resolve_cell_position(
...
+                // first row.
+                let mut resolved_y = first_available_row;
+
+                // There may be row groups, so we have to not only skip
+                // rows where the requested column is occupied to find the
+                // first suitable row, but also skip rows belonging to
+                // headers or footers.
+                loop {
+                    if let Some(Some(_)) =
+                        resolved_cells.get(cell_index(cell_x, resolved_y)?)
```
PgBiel, line 1891: Should take the deduplication opportunity to not call `cell_index(...)?` each time and instead start at a suitable index and keep trying to increase it by 1 (top loop) / `columns` (bottom - this - loop)

crates/typst-library/src/layout/grid/resolve.rs
PgBiel: Changed it a bit, feel free to unresolve if you have further suggestions.

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -705,560 +992,612 @@ impl<'a> CellGrid<'a> {
...
+                &mut pending_vlines,
+                &mut header,
+                &mut repeat_header,
+                &mut footer,
+                &mut repeat_footer,
+                &mut auto_index,
+                &mut resolved_cells,
+                child,
+            )?;
+        }
```
PgBiel, line 1039: Beautiful.

PgBiel (commented):
Overall, the refactor was pretty "light" (I didn't move all the context variables I could into struct fields), but I think the effects were quite significant already.

laurmaedje:
What do you think about turning `grid_to_cellgrid` and `table_to_cellgrid` into routines and moving them into `typst-layout`? I think it's conceptually sort of a layout task even though it's needed in HTML as well. And moving it would shave off a good amount of code from `typst-library`.

PgBiel:
> What do you think about turning `grid_to_cellgrid` and `table_to_cellgrid` into routines and moving them into `typst-layout`? I think it's conceptually sort of a layout task even though it's needed in HTML as well. And moving it would shave off a good amount of code from `typst-library`.

Not sure, in theory I can see it being useful from a library perspective in the future as well by providing a method or something which lets one access (and generate) the cellgrid before layout, within typst code. But i guess it doesn't really matter?

PR #4458, opened 2024-06-26
Add metadata for `cargo-binstall`

Coekjan:
Closes #4388 .

---

I tested this locally (x86_64 + linux + glibc) via:

```console
$ # in the root of this repo
$ cargo binstall typst-cli --manifest-path ./crates/typst-cli/Cargo.toml
```

Then, `cargo-binstall` downloaded the prebuilt artifacts for `x86_64-unknown-linux-musl`. This is because `cargo-binstall` finds a proper target-triple for us:

https://github.com/cargo-bins/cargo-binstall/blob/76814e4e8feaede87d8d112a4fad519d8eeaf81d/crates/bin/src/args.rs#L69-L73

---

To ensure it is working, we may need to test it locally for current prebuilt targets.
- [x] x86_64-unknown-linux-musl
- [ ] aarch64-unknown-linux-musl
- [ ] armv7-unknown-linux-musleabi
- [ ] x86_64-apple-darwin
- [ ] aarch64-apple-darwin
- [ ] x86_64-pc-windows-msvc

**I would appreciate it if you can help to test the untested targets above on your target-matched machine and report the result here to ensure everything works as expected.**

Issue #4388, opened 2024-06-13
Support for installation with cargo binstall


yodatak:
### Description

Would love to use simpler typst but it not present in fedora repo and it don't support the install with  cargo binstall typst

i would be cool to support this to make simple install 

Like for debian and other https://github.com/typst/typst/issues/3679

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

EpicEricEE:
One Windows it works with `cargo binstall typst-cli`, don't know about Fedora but I assume it should be similar.

alerque:
All `cargo binstall` does is look for precompiled binaries attached to upstream releases and install those instead of doing a source build like `cargo install` does. In fact it falls back to the latter if binary artifacts were not found.

Typst already attached prebuilt binaries for some platforms to releases, but they are not quite in the default naming scheme binstall expects. I think all that would be required to support it is add some meta data to the Cargo.toml file describing the existing format of artifacts including noting the XZ compression used and the tar file layout.

https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md

Coekjan:
Seems we just need to:
1. add some metadata in Cargo.toml for cargo-binstall
2. publish the `typst-cli` crate when releasing next version

BTW, I think this issue should be labeled "feature request", instead of "bug".

PR #5921, opened 2025-02-21
Mention `binstall` in `README.md`

rikhuijzer:
I just tested and `cargo` [binstall](https://github.com/cargo-bins/cargo-binstall) works here on MacOS Aarch64 (related issue https://github.com/typst/typst/issues/4388 was fixed in https://github.com/typst/typst/pull/4458).

This PR suggests to mention `binstall` in the README installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,8 +118,10 @@ Typst's CLI is available from different sources:
   - Windows: `winget install --id Typst.Typst`
 
 - If you have a [Rust][rust] toolchain installed, you can install
-  - the latest released Typst version with
+  - the latest released Typst version from source with
     `cargo install --locked typst-cli`
+  - the latest released Typst version from binary with
+    `cargo binstall --locked typst-cli`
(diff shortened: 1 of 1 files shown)


laurmaedje:
I think this is a bit too niche for inclusion in the README

smheidrich:
>I think this is a bit too niche for inclusion in the README

Would've saved me 5 minutes of looking through the issues just now :shrug: If `cargo binstall typst` just worked, I'd agree that users of `cargo-binstall` could just infer that from the current README saying `cargo install typst` and it wouldn't be worth mentioning. But since it's `cargo binstall typst-cli` instead of just `typst`, it would be nice to mention it.

laurmaedje:
The README says `cargo install --locked typst-cli` though 🤔

smheidrich:
Oh snap, I misread it :see_no_evil: Sorry, never mind then!

Issue #942, opened 2023-04-23
Feature Request : Allow to customize the Bibliography Format


YJDoc2:
Current `#bibliography` function is great with three inbuilt styles, until some journal requires the bibliography to be set in their own specific style. Can we have some way to customize the format of bibliography, at least in what order various elements appear, and their style ; or a way to provide a callback function which will accept the reference number and reference data (from .bib file or other) and return `Content` .

Thank you!

YJDoc2:
For those who are looking for a solution in meantime, I did
```

// must set size to 0, if we use hide, then #cite does not work
#[
#set text(size: 0pt) 
#bibliography("ref.bib",style:"ieee")
]

#linebreak()

// actual 'Bibliography' title
#[
#set text( size: 15pt)
*Bibliography*
]

// ... manually write refs one by one, in the order given by original bibliography
```

Marmare314:
At least bibliography headings can be modified with show rules in the next release #ff1e4049d168.

PgBiel:
I believe there could be some system for customizing bibliography styles, e.g. a function that takes a dict with all information specified for a given item of the bibliography file. This might ultimately be necessary to accomodate for all styles around the world (by letting users customize them entirely).

nluetts:
How about a function (let's say #iter-bibliography) that has the same effect as `#bibliography`, but it does not place a finished formatted bibliography but returns a dictionary where the key is the label of the citation and the value is a dictionary holding the bibliographic raw data (author, title, journal, etc.).

I would then expect to be able to use it as follows:

```typst
#for (label, entry) in iter-bibliography("references.bib") {
  [#label: #entry.author, #emph(entry.journal) #bold(entry.volume) (#entry.year), #entry.pages . \]
}
```
One would need to iterate over the authors as well, but I hope this gets the idea across.

Something like this may be possible with the feature suggested in #1220 , but having the iterator over the bibliography would perhaps make it a little easier to quickly customize the bibliography.

krusche:
Would be great if more styles can be supported.
I would actually be interested in `alpha` as e.g. declared here: https://web.archive.org/web/20161119024545/http://sites.stat.psu.edu/~surajit/present/bib.htm

DVDTSB:
> I believe there could be some system for customizing bibliography styles, e.g. a function that takes a dict with all information specified for a given item of the bibliography file. This might ultimately be necessary to accomodate for all styles around the world (by letting users customize them entirely).

[...]

PR #5932, opened 2025-02-22
Add `bibliography.entry` element

xkevio:
Partially addresses https://github.com/typst/typst/issues/942.

Adds a `bibliography.entry` sub-element, similar to `outline.entry` which exposes the `key` (label), `body` (formatted entry) and `prefix` (if the style has one) of each entry. Currently it also exposes the `indent` length of certain styles, but that is subject to change.

This also changes the internal layout of previously grid-based styles, such as IEEE, as those no longer rely on an internal grid but rather use the same method as `outline.indented` to measure prefix length and align accordingly. Furthermore adds the ability to apply `show`-rules on entries, such as `#show bibliography.entry: it => it.key` or `#show bibliography.entry.where(key: <your-key>): set text(red)`.

**EDIT:** Now also implements a `entry.indented()` API similar to `outline.indented()` with a `gap` parameter representing the faux column gutter for prefix-based styles. There is still a need to pass the information about whether the bibliography style uses a hanging indent into the show-rules of `bibliography.entry`; right now this is a style chain parameter.

## Biggest blocker

Implementing the `indented` API, or rather a faithful implementation of grid-based styles, requires knowing the maximum width of all prefixes which requires introspection and increases the layout iterations in a single document containing just the bibliography and some references from 3 to 4—this is rather unfortunate and remains an issue.


PgBiel:
> Implementing the indented API, or rather a faithful implementation of grid-based styles, requires knowing the maximum width of all prefixes which requires introspection and increases the layout iterations in a single document containing just the bibliography and some references from 3 to 4—this is rather unfortunate and remains an issue.

This is surprising to me. Can we not do the same thing grid does? Or is this something that only happens when you use `indented` in a show rule?

laurmaedje:
@PgBiel It's possible to do it without introspection, but then it won't work properly for arbitrary user show rules. 

The outline uses introspection to (a) align across all levels which would be hard with a grid and (b) to align across different blocks generated by user outline.entry show rules.

The bibliography does not suffer from (a), but does suffer from (b). It's not possible to measure up-front if we don't know what the user show rule generates.

xkevio:
@PgBiel Grid-based styles did use a `GridElem` previously but with the new delegation of the `BibliographyEntry` show-implementation, we only get back opaque and already laid out content and can't differentiate between prefix and entry anymore (well besides cursed content splitting, but then show-rules do not work anymore). I agree that it would be nice to tap-in to how the auto-sizing of a grid is handled since that did not require an additional introspection step.

PgBiel:
This kinda sucks imo, and has the potential to block further introspection within bibliography show rules, potentially reinforcing requests for increasing the amount of iterations...
I'd prefer if this could be some sort of 'internal' iteration somehow, but doesn't seem to be that simple.

PgBiel:
An alternative would be to somehow allow users to customize each column individually through separate sub-elements, but I don't think the styling system is flexible enough for that at the moment...

laurmaedje:
An internal iteration would be cool, but definitely require infrastructure we don't have at the moment.

laurmaedje:
As discussed on Discord, having a `block`-based `bibliography.entry` API with proper alignment for grid-based styles would necessarily incur a fourth introspection iteration on documents with a bibliography. This seems too high a price to pay at the moment.

A potential solution would be a local introspection iteration (basically, only iterating on the bibliography itself). This would also be useful for / unlock a lot of other introspection-based layouts I imagine. Unfortunately, it's not immediately on the horizon right now. But as a compiler-internal thing that's not initially exposed to users, it might actually not be extremely hard (not easy either though).

I'll close this for now, but I'd be happy to revive it if the local iteration materializes or if anybody has an idea for an alternative approach. Thanks for working on this still!

PR #5934, opened 2025-02-23
Unify layout of `vec` and `cases` with `mat`

### LINKED ISSUE #5422: Different appearance of `vec` and `mat`

mkorje:
Closes #5422.

```typ
--- math-mat-vec-cases-unity ---
// Test that matrices, vectors, and cases are all laid out the same.
$ mat(z_(n_p); a^2)
  vec(z_(n_p), a^2) 
  cases(reverse: #true, delim: \(, z_(n_p), a^2)
  cases(delim: \(, z_(n_p), a^2) $
```

| Old | New |
| - | - |
| old_math-mat-vec-cases-unity | new_math-mat-vec-cases-unity |

The test `math-equation-font` also needed to be updated due to this change.

This PR also makes linebreaks be ignored in `vec` and `cases` now (cf. #5459), matching `mat`. A warning is now emitted when a linebreak is encountered in all three functions:
```shell
warning: linebreaks are ignored in cells
  ┌─ tests/suite/playground.typ:2:39
  │
2 │ $ cases(a, b, c) cases(reverse: #true, a \ b \ c) $
  │                                        ^^^^^^^^^
  │
  = hint: use commas instead to separate each line
```

The warning/hint can probably be improved. I'd appreciate any suggestions.
Only concern/note I have is that the hint in the above warning is kind of misleading with `mat`, as a semicolon ends a row. Though even if the hint also suggested using semicolons, it wouldn't be very helpful I think in a scenario like:
```shell
warning: linebreaks are ignored in cells
  ┌─ tests/suite/playground.typ:2:15
  │
2 │ $ mat(1, 2, 3; 4 \ 5, 6; 7, 8 \ 9) $
  │                ^^^^^
  │
  = hint: use commas instead to separate each line

warning: linebreaks are ignored in cells
  ┌─ tests/suite/playground.typ:2:28
  │
2 │ $ mat(1, 2, 3; 4 \ 5, 6; 7, 8 \ 9) $
  │                             ^^^^^
  │
  = hint: use commas instead to separate each line
```


MDLC01:
Ideally, the warning should be different for all elements. For example, the "cell" terminology does not make much sense for `cases`.

mkorje:
> Ideally, the warning should be different for all elements. For example, the "cell" terminology does not make much sense for `cases`.

I've modified it to be consistent with the terminology in the docs: cell, element, and branch for `mat`, `vec`, and `cases respectively.

laurmaedje:
Thank you!

PR #5937, opened 2025-02-23
Run tests on 32-bit via Ubuntu multilib

selfisekai:


laurmaedje:
This looks promising. You didn't adapt the target for the "update test artifacts" step though and therefore it compiled from scratch again. Once that's fixed, we can see how long the CI would take if this were merged. If it's at least as fast as the Windows CI, we can have it on main, otherwise I might prefer to have it in a separate workflow that we can trigger pre-release.

PgBiel:
Do we know if running the tests this way, without using something like `cross` or even `qemu`, is enough to fully verify that tests would run in a proper 32-bit CPU?

I guess it will, at the very least, show how things would change with `usize` changed to 32 bits, so in that sense it'd work, though I'm a bit curious if there's something else we should be concerned about, e.g. regarding system libraries.

selfisekai:
It is a cross build, it does build proper 32-bit binaries that cannot depend on 64-bit libraries under the hood, you can already see that 32-bit openssl has to be installed. The reason it then simply runs is that x86_64 processors have a backwards compatibility mode for it (https://en.wikipedia.org/wiki/X86-64#OPMODES).

PgBiel:
That is interesting, I wasn't aware of this. If I understood correctly then, the binary is indeed compiled for a 32-bit architecture as usual, but a x86-64 CPU can run it anyway under compatibility mode. Thanks for the clarification!

laurmaedje:
Thank you!

Issue #2249, opened 2023-09-26
Add option to make text non-selectable


NicoWeio:
### Description

There are occasions where text is not meant to be selected or searched, such as watermarks (see below).
I don't know how easy this is to implement with PDF, but it would be a helpful addition.

### Use Case

Typst's documentation provides a good example:
```typst
#set page(background: rotate(24deg,
  text(18pt, fill: rgb("FFCBC4"))[
    *CONFIDENTIAL*
  ]
))

= Typst's secret plans
In the year 2023, we plan to take
over the world (of typesetting).
```

LuxxxLucy:
I think there are two ways for implementing this feature:

One is to add a flag in text element to indicate that when are exporting to pdf, we do not write the text but rather transform the glyph into an outline shape and write to the pdf. This way the text will be non selectable. A possible interface could be #text(selectable : false, aaa)

The second way is to utilize the ReplacementText feature (sec 14.9.4 PDF 1.7), it would be able to specify a different string when selected. For example, a possible interface would be #text(replacement: "bbb", aaa) and the result rendering would be aaa, but when select and copied out we will have bbb. The latex package called accsupp implemented this feature. In the second approach the non selectable text can be achieved by setting the replacement to be an empty string, such as #text(replacement: "", aaa)

I personally think the second approach would be a more general design, its only drawback being that it requires the pdf reader to support the replacement feature (though I assume most of popular pdf reader should already support). What do you think?

NicoWeio:
Thank you for your detailed answer! I'd also prefer the second approach. 

For the future (from my understanding, this is not currently a feature), we'd also need a way to control whether text from a SVG should be selectable, so it would make sense to consider both aspects when making a design decision. There's an issue for this latter part already: #1239.

laurmaedje:
The second approach or something similar to that is definitely preferrable. Outlining the glyphs will make it look different from other text because text rendering is somewhat platform-specific.

LuxxxLucy:
Hi I just tried out the text replacement approach. Good news is that it actually works (and the current pdf-writer we are using support it), the sad news is that only seems to be working in acrobat reader, but not in chrome, apple preview, etc. The support of the feature is not what I expected to be.

This is a test example.
[aaabbb.pdf](https://github.com/typst/typst/files/12777882/aaabbb.pdf)

Essentially this is the normal pdf text block 

``
stream
BT
/F1 14 Tf
108 734 Td
(aaa) Tj
ET
endstream
```

[...]

PR #5938, opened 2025-02-23
Draft: introduce watermark element to make text non-selectable by `artifact` Tag (PDF 1.7 18.4.2.2.2)

LuxxxLucy:
As discussed in #2249 and particular https://github.com/typst/typst/issues/2249#issuecomment-1759930396
Introduce a watermark element so to make PDF outputs the text with a tag of artifact.

As the watermark element does not have any effect on the layout, so it is implemented as a frame modifier.

## NOTE:
However, it appears that the artifact tag does not really makes the text non selectable even for Adobe Acrobat.
Probably need to find a better solution.

### DIFF
--- a/crates/typst-layout/src/modifiers.rs
+++ b/crates/typst-layout/src/modifiers.rs
@@ -1,6 +1,6 @@
 use typst_library::foundations::StyleChain;
 use typst_library::layout::{Fragment, Frame, FrameItem, HideElem, Point};
-use typst_library::model::{Destination, LinkElem};
+use typst_library::model::{Destination, LinkElem, WatermarkElem};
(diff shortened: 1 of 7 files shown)


LaurenzV:
Note that #5420 would make it pretty easy to add non-selectable text to PDFs.

LuxxxLucy:
> My envisioned design was more along the lines of `set text(selectable: false)`. See also https://github.com/typst/typst/pull/5789.

Okay I see, this would be essentially an selectable attribute to text element. I will re-do this (perhaps after the krilla pdf backend PR #5420 is merged, as we can outlines the glyphs)

PR #5939, opened 2025-02-24
Add dotless parameter to `math.accent`

mkorje:
So users can keep the dot above i and j when adding top accents, if they want.


mkorje:
I've drafted this to consider changing how `dtls` is applied by moving it to the style chain. When the i or j glyph isn't the whole base elem but a part of it, you can't tell the compiler to make them dotless.

laurmaedje:
I agree that `dotless` is best, in terms of the default, but also in terms of using established, searchable terminology. Thanks!

PR #5949, opened 2025-02-24
Override the default math class of some characters

### LINKED ISSUE #5764: Override the default math class for some characters
This issue is adapted from https://github.com/typst/typst/issues/4985#issuecomment-2608361748.

Below is a list of characters I believe we should actually change the class of, with reasoning.

- U+214B ⅋ is currently Normal, but should be Binary.
  - This follows unicode-math, which classifies it as `\mathbin`.
  - It also corresponds to its use in linear logic (where the symbol is called "par").[^1]
- U+27C5 ⟅ is currently Relation, but should be Opening.
  - This follows unicode-math, which classifies it as `\mathopen`.
  - This character's name in Unicode LEFT S-SHAPED BAG DELIMITER.
- U+27C6 ⟆ is currently Relation, but should be Closing.
  - This follows unicode-math, which classifies it as `\mathclose`.
  - This character's name in Unicode RIGHT S-SHAPED BAG DELIMITER.
- U+27C7 ⟇ is currently Relation, but should be Binary.
  - This follows unicode-math, which classifies it as `\mathbin`.
  - ∨ and most variants are classified as Binary.
  - ⟑ is classified as Binary.

Additionally, Typst partially overrides the math class of the period (U+002E .) to Normal. Changing this back to Punctuation (following Unicode) when not used as a decimal separator is the topic of #5700.

[^1]: https://en.wikipedia.org/wiki/Linear_logic#%E2%85%8B

MDLC01:
Closes #5764. See the issue description for more context.

### DIFF
--- a/crates/typst-utils/src/lib.rs
+++ b/crates/typst-utils/src/lib.rs
@@ -360,6 +360,21 @@ pub fn default_math_class(c: char) -> Option<MathClass> {
         // https://github.com/typst/typst/pull/5714
         '\u{22A5}' => Some(MathClass::Normal),
 
+        // Used as a binary connector in linear logic, where it is referred to
+        // as "par".
+        // https://github.com/typst/typst/issues/5764
+        '⅋' => Some(MathClass::Binary),
+
+        // Those overrides should become the default in the next revision of
+        // MathClass.txt.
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thank you!

PR #5957, opened 2025-02-25
Fix false positive for type/str comparison warning

laurmaedje:
This changes the type/str warning to prevent a false positive that's been occuring for example in `elembic`.

Checks like this:
```typ
type(10) in ("any", str, int)
```
currently raise a a warning, even though `"any"` is used not as a compatibility type here, but as a custom marker value.

cc @PgBiel


PgBiel:
Thanks! It works

Issue #4528, opened 2024-07-09
Ease debugging of "layout did not converge within 5 attempts" issues


oliverpool:
### Description

Add a way to reduce the number of layout attempts, to be able to output 5 pdfs (one for each iteration), allowing to better understand the lack of convergence.

When I fist encountered this error (see **Use Case**), my first thought was to augment the number of layout attempts (since it surely must converge at some point, right?), however after more thoughts, this would likely be bad idea, since people (like me) would just be lazy and increase this number without consideration.

Hence the opposite proposal: allow the number of attempts to be decreased.
This would allow me to see 5 PDFs and get a better feeling, why my layout is not converging.

### Use Case

I am trying to implement an `autobreak` feature (for a songbook system, see https://github.com/typst/typst/issues/993#issuecomment-2213532656), were I get `layout did not converge within 5 attempts` and no idea how to solve it:

https://typst.app/project/r9GASqmDOYasvWLDv5x4cI

If I execute the logic in my head, the layout should converge immediately, since at the beginning of every `autobreak` block, I can compute where the current block should go. So I really don't understand how Typst is computing the layout and why it is taking so many steps. Having a way to see the step 1-4 should help me get a better understanding (and eventually find a suitable fix).

---

BTW Typst looks awesome, thank your for this hard work!

I am creating a songbook-system to replace [patacrep](https://github.com/patacrep/patacrep) (which is based on LaTeX + [Songs](https://songs.sourceforge.net/)). The webapp looks really neat and I think that multiple people I know would love to use it to create their next songbook collaboratively (I know 3 songbook-projects which could benefit from it ;)

laurmaedje:
Yeah, some way to debug this error would certainly be nice. It's a difficult problem, but at some time we'll need to figure something out.

zinifier:
I had the same problem when trying to reach a specific number of pages. Usecase: i'm making booklets so i want to add pagebreaks so that the number of pages is a multiple of 4. It's unclear to me when the total page count is updated (on every layout iteration?) and how that affects dynamic page break insertion.

oliverpool:
(Note: to not pollute this feature request, I opened a discussion dedicated to my specific problem and eventual solutions: #4530)

Mc-Zen:
I suggest adding a CLI argument under the name `max-layout-iterations` or similar that restricts the number of layout iterations (maximum 5). This way a document can be compiled to see how it looks after a certain number of iterations. 

[...]

PR #5958, opened 2025-02-25
Add `--max-layout-iterations` cli argument

LeWimbes:
This pull request introduces a new CLI argument `--max-layout-iterations` to allow users to control the maximum number of layout iterations in the typst compiler.

As mentioned in various discussions (e.g., https://github.com/typst/typst/issues/4528), the "layout did not converge within 5 attempts" error can be tricky to debug.
While the [layout-ltd](https://typst.app/universe/package/layout-ltd) package by [davystrong](https://github.com/davystrong) ([see this comment](https://github.com/typst/typst/issues/4528#issuecomment-2665755004)) allows users to limit the iterations to below the default of 5, this PR lets users set any value - even above 5.

I've marked this PR as a draft because, as someone new to Typst, I'm not entirely sure if my changes align with the project's ideals. Additionally, there are a few unresolved issues:
- Is `ProcessArgs` the appropriate place for `max_iterations`? I chose it because it's shared between `QueryCommand` and `CompileArgs`.
- It doesn't feel ideal to include `max_iterations` in the `SystemWorld` struct, but I haven't found a better solution without passing it as a separate argument through multiple functions.
- `TimingScope` requires a `static` lifetime for the `name`. This forces me to either leak data from generated `String`s or to somehow lift the lifetime restriction. 
For now, I've kept the `const ITER_NAMES`and used a modulo operation (`% 5`) to keep values within bounds.

Please let me know if this CLI argument is of interest and if you have any suggestions on how to resolve these remaining concerns.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -326,6 +326,12 @@ pub struct ProcessArgs {
     /// The format to emit diagnostics in.
     #[clap(long, default_value_t)]
     pub diagnostic_format: DiagnosticFormat,
+
+    /// Determines the maximum number of layout iterations the compiler will perform in an attempt
+    /// to reach a stable state. If the document fails to stabilize within this limit, the compilation
+    /// process will give up with a warning.
+    #[arg(long = "max-layout-iterations", default_value_t = 5)]
+    pub max_iterations: usize,
 }
 
 /// Arguments related to where packages are stored in the system.
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -48,6 +48,8 @@ pub struct SystemWorld {
     /// always the same within one compilation.
     /// Reset between compilations if not [`Now::Fixed`].
     now: Now,
+    /// The maximum number of layout iterations.
+    max_iterations: usize,
 }
 
 impl SystemWorld {
@@ -142,6 +144,7 @@ impl SystemWorld {
             slots: Mutex::new(HashMap::new()),
             package_storage: package::storage(&world_args.package),
             now,
+            max_iterations: process_args.max_iterations,
         })
     }
 
@@ -233,6 +236,10 @@ impl World for SystemWorld {
             with_offset.month().try_into().ok()?,
             with_offset.day().try_into().ok()?,
         )
+    }
+
+    fn max_iterations(&self) -> usize {
+        self.max_iterations
     }
 }
 
--- a/crates/typst-library/src/lib.rs
+++ b/crates/typst-library/src/lib.rs
@@ -84,6 +84,10 @@ pub trait World: Send + Sync {
     /// If this function returns `None`, Typst's `datetime` function will
     /// return an error.
     fn today(&self, offset: Option<i64>) -> Option<Datetime>;
+
+
+    /// Get the maximum number of layout iterations.
+    fn max_iterations(&self) -> usize;
 }
 
 macro_rules! world_impl {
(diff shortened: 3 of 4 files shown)


LeWimbes:
Ah, okay. Thanks for letting me know. I wasn't aware that this had been mentioned before.

laurmaedje:
As @Enivex said, max iterations > 5 is not something we want to support for ecosystem reasons. Quoting myself from a [recent discussion on lobste.rs](https://lobste.rs/s/rsuopk/tex_typst_layout_models#c_srecvp):

> The reason I don’t want to add an iterations parameter isn’t because the problem is the user rather than the system, but that it doesn’t fix the problem properly. It just delays the problem (and as you mention makes Typst slower). It encourages an ecosystem full of unidiomatic solutions that require many iterations.

> I think the proper solution is giving users better tools to understand and debug issues related to iteration count, so that they can build solutions that don’t require so many iterations. Building such tools is difficult and we’re just not there yet. By the way, I agree that adding an iterations parameter now would make debugging a bit easier, but I think the price paid in terms of where the ecosystem goes is too high. That’s why I rather accept the bad situation now until we have proper debugging tools.

An iterations parameter that only allows going <= 5 would not come with these problems. I do want to support this as an IDE feature in the web app (and I expect that tinymist will also add this), but I haven't yet made up my mind on whether I want to add (and keep supporting) it as a CLI feature. Especially with the [layout-ltd](https://typst.app/universe/package/layout-ltd) package, the urgency for such a parameter is a bit reduced.

LeWimbes:
Thanks for the detailed explanation. It makes a lot of sense to enforce a low number of iterations across the ecosystem. I looked into the max layout iterations because I encountered the limit with a minimal example that I don't find too unrealistic:
```typst
#import "@preview/abbr:0.2.3"
#import "@preview/drafting:0.2.2": set-page-properties, margin-note

#abbr.make(
  ("BC", "Boundary Condition"),
)

#let left-margin = 65mm

#set page(
  margin: (
    left: left-margin,
  ),
)

#set-page-properties(margin-left: left-margin)

Stephen Hawking wrote that "The #abbr.a[BC] of the universe is that it has no boundary."

The abbr package can automatically generate the plural form, such as #abbr.pla[BC].

The drafting package adds comments to the document: #margin-note[This is a comment!]
```

It appears that the problem stems from the large margin and the interplay between these two packages. Since it's impossible to predict all possible interactions of packages and configurations in typst documents, I'm not entirely sure that imposing a strict end-user limitation is the best solution.

Perhaps a better alternative would be to implement automated tests on packages/templates submitted to the typst universe. This way, we can enforce a reasonable iterations limit at the ecosystem level while still allowing users the option to bypass this limit when needed.

Thanks again for the insights. Let me know what you think and if you’d prefer that I close this PR for now or if you’d like me to continue working on it.

laurmaedje:
It's true that sometimes there is an unfortunate interaction that increases the number of iterations needed. Though that can also point to either of the packages in question already requiring unnecessarily many iterations in the first place. 

But even if we might end up at the point where we say "five just isn't enough for everything", I do not want to commit to completely free customization at this point. I think the next steps in the convergence story are to improve error messages to make such errors easier to debug and then we can see down the road how it goes. For now, I would close this PR. Thanks still for the PR and the productive discussion!

SolidTux:
What about a compile time configuration of the iteration limit (e.g. with an env variable)? That way, the ecosystem still has to use the default limit because that is what the webapp and distro packages will use, but it's still possible for a project to easier configure a higher limit.

PR #5963, opened 2025-02-26
Fix curve with multiple non-closed components.

Emm54321:
Using curve() with multiple non-closed components is broken. E.g.:

```typst
#curve(
  stroke: 2pt,
  curve.line((20pt, 0pt)),
  curve.move((0pt,  10pt)),
  curve.line((20pt, 10pt)),
)
```
This should draw two horizontal lines, but doesn't.

This pull-request fixes this.


laurmaedje:
Thanks!

PR #5969, opened 2025-02-27
Improve number lexing

wrzian:
This PR refactors number lexing to try to improve some error messages and so we start eating trailing percent signs in numbers (mentioned before in this comment: https://github.com/typst/typst/pull/5560#issuecomment-2537099147).

The commits are best read in-order: the first is large, but the second and third are smaller.

The main thrust is a change to `Lexer::number()` to prepare two result variables: `number_result` and `maybe_suffix_result`. These are then resolved at the end of the function in a large match statement. The result variables are a bit overloaded, but they're functional, and I couldn't come up with a better way to get every error message to work as they should.

This isn't meant to be performance oriented, but there is one improvement in that we now avoid sending floats through the `i64::from_str_radix()` function (floats should always become an `Err`) by adding an `is_float` parameter that is checked to instead go straight to normal float parsing.

I'm unhappy that the function is now 130 lines, but I've rewritten it a number of times and been unable to separate blocks in a cleaner way. Happy to take suggestions and nits.


laurmaedje (commented):
I'm not sure the `try adding a space after` hint is too useful. If you really wanted to have text tightly after the number, a semicolon would be the proper fix, but it's just such a weird thing to write in the first place. The hint seems a _bit_ over-engineered to me.

I also agree that the match is pretty complex, which sort of goes hand in hand with the complexity of the hint.

laurmaedje:
@wrzian Are you still planning to work on this? :)

wrzian:
Yes, glad to be getting back to contributing here again!

> The hint seems a bit over-engineered to me.

Agreed that it's a bit much. Do you think we should suggest removing the suffix or just not suggest anything?

laurmaedje:
It seems to me that we can't generally guess the user's intent here, so perhaps no hint is best.

wrzian:
Updated to remove the space hint and was able to simplify the main commit. Should be ready for another review.

wrzian:
Clippy :cry:

laurmaedje:
Thank you!

PR #5970, opened 2025-02-27
Add default parameter for array.{first, last}

### LINKED ISSUE #5960: `array.{first,last}` `default` argument in value position
### Description

In value position, allow passing a named `default` argument to `array.first` and `array.last` that would behave like the one on `array.at`.

### Use Case

Being able to use the more readable `array.first` and `array.last` methods instead of `array.at`with `0` and `-1` when a default value is needed.

bluebear94:
Closes #5960.

### DIFF
--- a/crates/typst-library/src/foundations/array.rs
+++ b/crates/typst-library/src/foundations/array.rs
@@ -174,15 +174,25 @@ impl Array {
     /// Returns the first item in the array. May be used on the left-hand side
     /// of an assignment. Fails with an error if the array is empty.
     #[func]
-    pub fn first(&self) -> StrResult<Value> {
-        self.0.first().cloned().ok_or_else(array_is_empty)
+    pub fn first(
+        &self,
+        /// A default value to return if the index is out of bounds.
+        #[named]
+        default: Option<Value>,
+    ) -> StrResult<Value> {
+        self.0.first().cloned().or(default).ok_or_else(array_is_empty)
     }
 
     /// Returns the last item in the array. May be used on the left-hand side of
     /// an assignment. Fails with an error if the array is empty.
     #[func]
-    pub fn last(&self) -> StrResult<Value> {
-        self.0.last().cloned().ok_or_else(array_is_empty)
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/foundations/array.rs
```diff
@@ -174,15 +174,25 @@ impl Array {
     /// Returns the first item in the array. May be used on the left-hand side
     /// of an assignment. Fails with an error if the array is empty.
     #[func]
-    pub fn first(&self) -> StrResult<Value> {
-        self.0.first().cloned().ok_or_else(array_is_empty)
+    pub fn first(
+        &self,
+        /// A default value to return if the index is out of bounds.
```
T0mstone, line 179:
```suggestion
        /// A default value to return if the array is empty.
```
There's no index here, so how about this?

crates/typst-library/src/foundations/array.rs
```diff
@@ -174,15 +174,25 @@ impl Array {
...
     }
 
     /// Returns the last item in the array. May be used on the left-hand side of
     /// an assignment. Fails with an error if the array is empty.
     #[func]
-    pub fn last(&self) -> StrResult<Value> {
-        self.0.last().cloned().ok_or_else(array_is_empty)
+    pub fn last(
+        &self,
+        /// A default value to return if the index is out of bounds.
```
T0mstone, line 191: same here

laurmaedje:
As mentioned in https://github.com/typst/typst/issues/5960#issuecomment-2694057832, I'm not a fan of optional handling in Typst and the proliferation of `default` arguments, but given that there is no solution for that in sight, I guess this is still an improvement compared to the status quo. And if we come up with something better, we can deprecate all of them together.

One alternative route would be to return `none` for not present, but I think that's worse as it means you can't distinguish not present vs. `none`. (Which is probably how JS ended up with `undefined` and `null`, which are the bane of my existence.) It's just a hard problem in dynamic languages as they can't represent `Option<Option<T>>`.

PR #5975, opened 2025-02-28
Make `array.chunks` example more readable

Andrew15-5:
In https://typst.app/docs/reference/foundations/array/#definitions-chunks the 2 arrays are compared, but in the output they go one after another horizontally and not vertically.


laurmaedje:
Thanks!

PR #5976, opened 2025-02-28
Improve french smartquotes

### LINKED ISSUE #5952: More relevant smartquotes for French
### Description

In French from France (and maybe other regions as well), single guillemets are never used. The only used quotes are the guillemets (« and »), double quotation marks (“ and ”), and rarely single quotation marks (‘ and ’). I suggest we change the main (i.e., non-alternative) single smartquotes for French to “ and ”.

There is the question of whether it makes sense to have single smartquotes map to double quotation marks. I'm not sure, but it feels more useful than mapping them to an essentially useless character.

Additionally, I'm not sure whether single guillemets are used in other regions. In Switzerland, single guillemets are used as second-level guillemets. Otherwise, it seems that single guillemets are not used.[^1]

[^1]: https://fr.wikipedia.org/wiki/Guillemet#Double_ou_triple_niveau_de_citation (french)

### Use Case

Easier access to double quotation marks in French.

MDLC01:
Closes #5952.

I decided to use narrow no-break spaces because that seems to be what should be the norm when they are available (i.e., the other kinds of spaces are used because people do not have easy access to narrow no-break spaces and some software use regular no-break spaces by default).


laurmaedje:
Alright, let's do it. I'm curious whether someone will open a bug report about `'` becoming `“`, but I see how it's the most sensible option.

PR #5980, opened 2025-02-28
Do not force `math.mid` elements to have the Large math class

### LINKED ISSUE #5809: `math.mid` does not respect spacing from the math class
### Description

I would expect the three following equations to render with the same horizontal spacing.

```typ
$ (a class("fence", |) b) $
$ (a mid(class("fence", |)) b) $
$ (a class("fence", mid(|)) b) $
```

The first equation has more horizontal spacing around the vertical bar.

Tested with both Typst 0.12.0 and the 2025-01-13 development version.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

MDLC01:
Closes #5809.

This PR removes the code that forced the result of `math.mid` to be Large in a way that could not be overridden by the user. Instead, `math.mid` now turns its content to Fence, and this can be overridden by wrapping the `math.mid` element itself in a `math.class` element.

This means by default, the result of `math.mid` is now always a Fence. Keeping the original class instead prevented the slash (which is not a Fence) in test `math-lr-mid` from being scaled properly. I'm not sure exactly why, but at the same time I feel like changing the class to Fence is an acceptable behavior for `math.mid`.


mkorje:
> Keeping the original class instead prevented the slash (which is not a Fence) in test `math-lr-mid` from being scaled properly. I'm not sure exactly why, but at the same time I feel like changing the class to Fence is an acceptable behavior for `math.mid`.

I think its due to this check in the `scale` function as the slash has math class Normal:

https://github.com/typst/typst/blob/1025419b5a42774a5342e6079c6bffdfb90a52f0/crates/typst-layout/src/math/lr.rs#L126-L129

I commented in https://github.com/typst/typst/issues/5809#issuecomment-2691915569 that maybe we should instead be forcing math class Relation.

MDLC01:
`mid` now changes the class to Relation, and this can be overridden by wrapping `mid` inside `class`.

tests/suite/math/delimited.typ
```diff
@@ -77,6 +77,12 @@ $ lr(body) quad
   lr(size: #1em, body) quad
   lr(size: #(1em+20%), body) $
 
+--- math-lr-mid-class ---
+// Test that `mid` creates a Relation, but that can be overridden.
+$ (a | b) $
+$ (a mid(|) b) $
+$ (a class("unary", |) b) $
```
laurmaedje, line 84: there doesn't seem to be a test that actually tests class + mid. if I try `$ (a class("unary", mid(|)) b) $`, then it does not have unary spacing.

laurmaedje:
I had the same mental journey as @MDLC01, where I thought Fence makes more sense, but with the explanation in https://github.com/typst/typst/issues/5809#issuecomment-2692526277, I see how Relation is sensible.

tests/suite/math/delimited.typ
MDLC01: Well, that's annoying. It indeed doesn't work with the correct test, and I can't figure out why. I will take a closer look later.

MDLC01: > Coincidentally, I think when doing the shaper PR I've made sure to not change the class when stretching a glyph, so it'll also just work as expected if that gets merged.

To be clear, does that mean if I continue to ignore this problem it will just fix itself thanks to your hard work?

laurmaedje: Shaping PR is merged!

MDLC01: Incredible! I'll resolve the conflict tonight

laurmaedje:
Thank you!

PR #5987, opened 2025-03-01
Fix docs example with type/string comparison

7ijme:
In v13 `type` doesn't return a string anymore, but a type. So it has to be changed from `type(it) == "dictionary"` to `type(it) == dictionary`.

### DIFF
--- a/crates/typst-library/src/loading/xml.rs
+++ b/crates/typst-library/src/loading/xml.rs
@@ -41,7 +41,7 @@ use crate::loading::{DataSource, Load, Readable};
 ///   ]
 ///
 ///   for p in pars.children {
-///     if (type(p) == "dictionary") {
+///     if (type(p) == dictionary) {
 ///       parbreak()
 ///       p.children.first()
 ///     }
@@ -50,7 +50,7 @@ use crate::loading::{DataSource, Load, Readable};
 ///
 /// #let data = xml("example.xml")
 /// #for elem in data.first().children {
-///   if (type(elem) == "dictionary") {
+///   if (type(elem) == dictionary) {
 ///     article(elem)
 ///   }
 /// }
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/loading/xml.rs
```diff
@@ -41,7 +41,7 @@ use crate::loading::{DataSource, Load, Readable};
 ///   ]
 ///
 ///   for p in pars.children {
-///     if (type(p) == "dictionary") {
+///     if (type(p) == dictionary) {
```
PgBiel, line 44: While we're at it:

```suggestion
///     if type(p) == dictionary {
```

crates/typst-library/src/loading/xml.rs
```diff
@@ -50,7 +50,7 @@ use crate::loading::{DataSource, Load, Readable};
 ///
 /// #let data = xml("example.xml")
 /// #for elem in data.first().children {
-///   if (type(elem) == "dictionary") {
+///   if (type(elem) == dictionary) {
```
PgBiel, line 53:
```suggestion
///   if type(elem) == dictionary {
```

crates/typst-library/src/loading/xml.rs
```diff
@@ -41,7 +41,7 @@ use crate::loading::{DataSource, Load, Readable};
 ///   ]
```
PgBiel, line 41: I feel like we could replace `heading(title.children.first())` with `[= #title.children.first()]` as well, though this one might be a bit more debatable.

PgBiel (commented):
Good eye!

crates/typst-library/src/loading/xml.rs
7ijme: I agree! will add it

PgBiel (approved):
Works here. Thanks!

PR #5991, opened 2025-03-02
Respect `quotes: false` in inline quote

### LINKED ISSUE #5536: Quote function always wraps with quotation marks in inline mode
### Description

According to the documentation, setting `quotes` to `false` should disable wrapping quotation marks regardless of the value of `block`.

> ### [quotes](https://typst.app/docs/reference/model/quote/#parameters-quotes)
>
> Whether double quotes should be added around this quote.
>
> The double quotes used are inferred from the quotes property on [smartquote](https://typst.app/docs/reference/text/smartquote/), which is affected by the lang property on [text](https://typst.app/docs/reference/text/text/).
>
> - `true`: Wrap this quote in double quotes.
> - `false`: Do not wrap this quote in double quotes.
> - `auto`: Infer whether to wrap this quote in double quotes based on the `block` property. If `block` is `false`, double quotes are automatically added.

The following Typst document demonstrates (at least on MacOS and Windows) that the `quotes` parameter has no effect at all when `block` is set to `false`, whereas I would expect only `quotes: auto` to be effected by `block` mode.

```typst
#show quote: emph
#show table.cell.where(y: 0): strong
#let red-text(str) = text(fill: red, str)

= Testing `#quote(quotes)`

https://typst.app/docs/reference/model/quote/#parameters-quotes

#table(
	columns: (40pt, 40pt, 1fr, 1fr, 25pt),
	align: (center, center, left, left, center),
	fill: (_, y) => if calc.odd(y) { luma(220) },
	stroke: none,

	table.header(`block`, `quotes`, "Sample", "Documentation", ""),

	`true`, `auto`,
		quote(block: true, quotes: auto, lorem(12)),
		[
			"If block is false, double quotes are automatically added." \
			Infers: If block is *true*, double quotes are *not* added.
		],
		sym.checkmark.heavy,

	`true`, `true`,
		quote(block: true, quotes: true, lorem(12)),
		["Wrap this quote in double quotes." Explicit.],
		sym.checkmark.heavy,

	`true`, `false`,
		quote(block: true, quotes: false, lorem(12)),
		["Do not wrap this quote in double quotes." Explicit.],
		sym.checkmark.heavy,

	`false`, `auto`, [
			Lorem ipsum dolor sit amet,
			#quote(block: false, quotes: auto)[consectetur adipiscing elit,]
			sed do eiusmod tempor.
		],
		["If block is false, double quotes are automatically added." Explicit.],
		sym.checkmark.heavy,

	`false`, `true`, [
			Lorem ipsum dolor sit amet,
			#quote(block: false, quotes: true)[consectetur adipiscing elit,]
			sed do eiusmod tempor.
		],
		["Wrap this quote in double quotes." Explicit.],
		sym.checkmark.heavy,

	`false`, `false`, [
			Lorem ipsum dolor sit amet,
			#quote(block: false, quotes: false)[consectetur adipiscing elit,]
			sed do eiusmod tempor.
		],
		[
			"Do not wrap this quote in double quotes." Explicit. \
			#red-text[Quote is incorrectly wrapped in double quotes.]
		],
		red-text(sym.crossmark.heavy),
)
```

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

3w36zj6:
close #5536


laurmaedje:
Thanks!

Issue #5964, opened 2025-02-26
Bad paragraph example in tutorial


laurmaedje:
### Description

This has a heading inside of a `par`, which doesn't work anymore in 0.13.
```
#par(justify: true)[
  = Background
  In the case of glaciers, fluid
  dynamics principles can be used
  to understand how the movement
  and behaviour of the ice is
  influenced by factors such as
  temperature, pressure, and the
  presence of other fluids (such as
  water).
]
```

PR #5994, opened 2025-03-02
Fix set rule example

Andrew15-5:
https://typst.app/docs/tutorial/formatting/

Fixes https://github.com/typst/typst/issues/5964.

Is this what you had in mind, @laurmaedje? I think this is reasonable, because why would you include the heading anyway? Or do you want a completely different example?

### DIFF
--- a/docs/tutorial/2-formatting.md
+++ b/docs/tutorial/2-formatting.md
@@ -13,21 +13,12 @@ your report using Typst's styling system.
 As we have seen in the previous chapter, Typst has functions that _insert_
 content (e.g., the [`image`] function) and others that _manipulate_ content that
 they received as arguments (e.g., the [`align`] function). The first impulse you
-might have when you want, for example, to justify the report, could be to look
-for a function that does that and wrap the complete document in it.
+might have when you want, for example, to underline text with dotted line, could
+be to look for a function that does that and wrap the desired text in it.
 
 ```example
-= Background
-#par(justify: true)[
-  In the case of glaciers, fluid
-  dynamics principles can be used
-  to understand how the movement
-  and behaviour of the ice is
-  influenced by factors such as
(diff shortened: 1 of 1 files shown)


laurmaedje:
I had a different thing in mind. I wanted to change `#par(..)[..]` to `#set par(..)`. Using `par[..]` is almost always the wrong choice. Even if we include no heading, we shouldn't encourage it.

Andrew15-5:
You already change that in the example. Look in the docs.

Andrew15-5:
Rewritten as in use the same basic example or do I need to use another set rule entirely?

laurmaedje:
Probably another set rule. But I'm not sure about the ideal fix. Answering that is similar to fixing it myself, as only by writing I'd get an intuitition about what's best there. If you're unsure how to approach this, probably it'd be best to just pick up something different where the goal is a bit clearer.

Andrew15-5:
I don't want to go a completely opposite direction, hence why I ask.

I think I can cook something up.

Andrew15-5:
An easy example would be to use `align`, `heading`, `underline` etc. But they are not as heavily used as justifying a paragraph. There is also `text`, but there is nothing that I would use repetitively. Also, a lot of more useful rules are show-set ones.

laurmaedje:
The old text didn't explain set rules particularly well, but I think the underline example is a bit too involved and does not increase clarity. Making the explanations properly good is a larger task (the whole tutorial needs a rewrite tbh). So for now, I'd like to just have a minimal fix. I would have just pushed to this PR, but it also contains unrelated editorial changes, so it was simpler to [make a new one](https://github.com/typst/typst/pull/6023). As a general rule of thumb, I really appreciate it if a fix PR is focussed on one thing only. Thanks still!

PR #5925, opened 2025-02-22
Make math shorthands noncontinuable

mkorje:
Closes #2789.

PR #5996, opened 2025-03-03
Parse multi-character numbers consistently in math

### LINKED ISSUE #4828: Different parsing in maths mode with single-digit vs. multiple-digit numbers
### Description

```typ
$
1/2(x)
$
```
Rendered by Typst-Renderer on the Typst Discord server https://discord.com/channels/1054443721975922748/1056574658771681361/1277188717794299967
```typ
$
1/10(x)
$
```
Rendered by Typst-Renderer on the Typst Discord server https://discord.com/channels/1054443721975922748/1056574658771681361/1277188738912358444

### Reproduction URL

_No response_

### Operating system

### Typst version

wrzian:
Closes #4828

I just did the minimal easy fix here.


laurmaedje:
I wonder whether removing the `None` from the match and only allowing Alphabetic chars would be a better fix, that could also supersede the [shorthand check](https://github.com/typst/typst/pull/5925).

However, changing this is a bit dicey and I'm really just wondering. It might be dumb.

cc: @mkorje

wrzian:
I believe the None check also allows things like emoji to act as function names.

laurmaedje:
Yeah, one test triggers that but tbh I'm not sure what the correct behaviour here is. Unicode doesn't consider emoji as alphabetic. If we say letters should do this, but number shouldn't, and punctuation probably not either (considering the shorthand check), should emoji?

mkorje:
I think it would be fine to only allow alphabetic chars as function names. It is reasonable behaviour I think and I can't imagine there's really anyone using emojis in this way in any event (?).

laurmaedje:
cc @MDLC01 thoughts?

MDLC01:
I don't think many people are using emoji as function names either (and if they want to they can always add extra parentheses).

I just want to note that Unicode might not classify as Alphabetic some characters that should be. I don't have an example right now, but I'll try to find one.

MDLC01:
This CSS rule can be used on [the General Symbols documentation page](https://typst.app/docs/reference/symbols/sym/) to highlight all characters that do not have a math class:

```css
.symbol-grid li:not([data-math-class]) button {
  background: red;
}
```

Some clearly alphabetical symbols do not have a math class, such as `Upsilon` (Υ), `angstrom` (Å), `Kai` (Ϗ), and `kai` (ϗ). There is also Hebrew letters, but those are replaced with corresponding symbols when in math, which do have the Alphabetic math class.

laurmaedje:
We could theoretically also check `math_class == Alphabetic || c.is_alphabetic()`.

laurmaedje:
I made the change. Would appreciate one of you doing a sanity check.

mkorje:
All looks fine to me.

Issue #5690, opened 2025-01-12
Out of memory error with "show par: box"


noahjutz:
### Description

The line `show par: box` causes the typst compiler to run out of memory quickly. On my linux PC, it allocates memory until my PC freezes entirely, and on the web app, an error is shown.

```typ
#show par: box

Hello

World
```

I found this out while trying to figure out a way to prevent typst from automatically creating paragraph breaks.

### Reproduction URL

https://typst.app/project/r906rRWR9Sdl98rb5GL53a

### Operating system

Web app, Linux

### Typst version

- [X] I am using the latest version of Typst

Enivex:
I'm going to go as far as saying that you should never reach for "#show par" ever. Basically everything is a paragraph, so it's not doing what you think it's doing.

noahjutz:
I'm not saying that `show par: box` should solve my problem. But passing invalid code to the typst compiler shouldn't cause it to crash my entire OS. Instead, an error message like with cyclic show rules would be helpful:

```typst
#show box: block
#show block: box // maximum show rule depth exceeded
```

laurmaedje:
I just tried to reproduce this with Typst 0.12 and it errored with 
```
error: maximum grouping depth exceeded
```
Are you sure you were using the latest version of Typst?

laurmaedje:
I have no clue what happened before, but now I can reproduce it.

quachpas:
Probably the same root cause for the following code
```typ
#show par: it => {
   // it
   repr(it)
}
#lorem(1)
```
I am on Typst cfb3b1a

Andrew15-5:
I debugged this. I didn't figure out why the memory is getting eaten, but I found a callback loop that starts at:

https://github.com/typst/typst/blob/cfb3b1a2709107f0f06f89ea25cabc939cec15e5/crates/typst-realize/src/lib.rs?plain=1#L669

Then it goes deeper by 7 callbacks, after which it goes up and goes back down/deep at the same line.

I used this MRE:

```typ
#show par: box
a

a
```

<details><summary>Call stack</summary>

```
  typst_realize::visit_grouping_rules lib.rs:681
  typst_realize::visit lib.rs:267
  typst_realize::visit_styled lib.rs:577
  typst_realize::visit_show_rules lib.rs:407
  typst_realize::visit lib.rs:247
  typst_realize::finish_par lib.rs:990
  typst_realize::finish_innermost_grouping lib.rs:810
[...]

PR #5997, opened 2025-03-03
Fix debug implementation of Recipe

Andrew15-5:
While I was decrypting https://github.com/typst/typst/issues/5690, I added

```rs
        dbg!(&styled.child);
        dbg!(&styled.styles);
```

to

https://github.com/typst/typst/blob/16948bd712e9f3b93cf2a8c7739d9528e15f7a23/crates/typst-realize/src/lib.rs?plain=1#L261

Which showed:

```rs
[crates/typst-realize/src/lib.rs:265:9] &styled.styles = Styles [
    Show(Elem(
        Element(par),
        None,
    ), Func(box),
]
```

Which doesn't print the same amount of closing parentheses as opened ones.

### DIFF
--- a/crates/typst-library/src/foundations/styles.rs
+++ b/crates/typst-library/src/foundations/styles.rs
@@ -471,7 +471,8 @@ impl Debug for Recipe {
             selector.fmt(f)?;
             f.write_str(", ")?;
         }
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

PR #6000, opened 2025-03-04
Improve docs on customizing `raw`

PgBiel:
Proposed additions:
- Mention that `raw` is not suitable for formatted content
- Add information about the default `raw` style
   - Font: DejaVu Sans Mono
   - Size: `0.8em`
- Briefly mention `theme` and `syntaxes` in the main `raw` description

Corrections and suggestions are welcome.

### DIFF
--- a/crates/typst-library/src/text/raw.rs
+++ b/crates/typst-library/src/text/raw.rs
@@ -30,6 +30,12 @@ use crate::World;
 /// Displays the text verbatim and in a monospace font. This is typically used
 /// to embed computer code into your document.
 ///
+/// Note that text given to this element cannot contain arbitrary formatting,
+/// such as `{*strong*}` or `{_emphasis_}`, as it is displayed verbatim. If
+/// you'd like to display any kind of content with a monospace font, instead of
+/// using [`raw`], you should change its font to a monospace font using the
+/// [`text`]($text) function.
+///
 /// # Example
 /// ````example
 /// Adding `rbx` to `rcx` gives
@@ -56,6 +62,35 @@ use crate::World;
 /// #raw("fn " + "main() {}", lang: "rust")
 /// ```
 ///
+/// # Styling
+///
+/// By default, the `raw` element uses the `DejaVu Sans Mono` font, available
+/// by default in Typst, with a smaller font size of `0.8em` (that is, 80% of
+/// the global font size). You can customize these properties with a show-set
+/// rule:
+///
+/// ````example
+/// // Switch to Cascadia Code for both inline and block raw.
+/// #show raw: set text(font: "Cascadia Code")
+/// // Make raw blocks 20% larger than normal text,
+/// // while keeping inline raw smaller than text.
+/// #show raw.where(block: true): set text(1.2em)
+///
+/// Now using the `Cascadia Code` font for raw text.
+/// Here's some Python code. It looks larger now:
+///
+/// ```py
+/// def python():
+///   return 5 + 5
+/// ```
+/// ````
+///
+/// In addition, you can customize the syntax highlighting colors by setting
+/// a custom theme through the [`theme`]($raw.theme) field.
+///
+/// For complete customization of the appearance of a raw block, a show rule
+/// on [`raw.line`]($raw.line) could be helpful, such as to add line numbers.
+///
 /// # Syntax
 /// This function also has dedicated syntax. You can enclose text in 1 or 3+
 /// backticks (`` ` ``) to make it raw. Two backticks produce empty raw text.
@@ -72,6 +107,10 @@ use crate::World;
 /// needed, start the text with a single space (which will be trimmed) or use
 /// the single backtick syntax. If your text should start or end with a
 /// backtick, put a space before or after it (it will be trimmed).
+///
+/// If no syntax highlighting is available by default for your specified
+/// language tag, you may provide a custom syntax specification file to the
+/// [`syntaxes`]($raw.syntaxes) field.
 #[elem(
     scope,
     title = "Raw Text / Code",


PgBiel:
It has also been discussed on Discord (https://discord.com/channels/1054443721975922748/1088371919725793360/1346283645907767389) that perhaps we should associate `\texttt` in the LaTeX guide to `#text(font: "monospace font here")[...]`.

However, I'll note that the concrete case was better suited as moving the raw inwards. That is, using `#link("URL", raw("URL"))` instead of `#raw(link("URL", "URL"))`. So this is something to consider: sometimes, what you need is to replace just the plain text with raw. But the point I make in my PR is precisely for the cases where you can't (when there is formatting).

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,35 @@ use crate::World;
...
+/// by default in Typst, with a smaller font size of `0.8em` (that is, 80% of
+/// the global font size). You can customize these properties with a show-set
+/// rule:
+///
+/// ````example
+/// // Switch to Cascadia Code for both inline and block raw.
+/// #show raw: set text(font: "Cascadia Code")
+/// // Make raw blocks 20% larger than normal text,
+/// // while keeping inline raw smaller than text.
+/// #show raw.where(block: true): set text(1.2em)
```
laurmaedje, line 77: They aren't really 20% larger than normal text because it stacks with the 0.8em. So the end result is 96% of the size.

Monospace fonts are for some reason just way larger than normal fonts. [Browsers also have a weird quirks due to this.](https://meyerweb.com/eric/thoughts/2010/02/12/fixed-monospace-sizing/)

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,35 @@ use crate::World;
 /// #raw("fn " + "main() {}", lang: "rust")
 /// ```
 ///
+/// # Styling
+///
```
laurmaedje, line 66:
```suggestion
```

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,35 @@ use crate::World;
...
+/// By default, the `raw` element uses the `DejaVu Sans Mono` font, available
+/// by default in Typst, with a smaller font size of `0.8em` (that is, 80% of
+/// the global font size). You can customize these properties with a show-set
+/// rule:
+///
+/// ````example
+/// // Switch to Cascadia Code for both inline and block raw.
```
laurmaedje, line 73: The example is a bit wide so that it won't fit side-by-side, which is somewhat unfortunate. But the show-set rules would probably need to be broken in a somewhat unreadable way to make it fit into the 38 char limit.

crates/typst-library/src/text/raw.rs
```diff
@@ -30,6 +30,12 @@ use crate::World;
 /// Displays the text verbatim and in a monospace font. This is typically used
 /// to embed computer code into your document.
 ///
+/// Note that text given to this element cannot contain arbitrary formatting,
+/// such as `{*strong*}` or `{_emphasis_}`, as it is displayed verbatim. If
```
MDLC01, line 34:
```suggestion
/// such as `[*strong*]` or `[_emphasis_]`, as it is displayed verbatim. If
```

I think this is what you meant.

PgBiel: No, \`{...}\` is used elsewhere to syntax highlight expressions.

However, I'm now seeing that the expressions are all in code mode already, so this could fail to provide the desired highlighting. Perhaps `{[*strong*]}` would be needed to highlight properly, but i have a feeling that this could detract from the original purpose of the phrase. An alternative is ``[`{strong}`]($strong)`` and ``[`{emph}`]($emph)``, albeit less self-evident.

laurmaedje: `{...}` gives code highlighting and `[...]` gives markup highlighting. The brackets are stripped in both cases.

PgBiel: Oh, I thought that brackets could only be interpreted differently for links. This should indeed be adjusted that way then.

crates/typst-library/src/text/raw.rs
PgBiel: Yeah, this seems a bit unavoidable, as otherwise we'd be encouraging people copying the example to use those otherwise unnecessary line length workarounds.

PgBiel: I've broken up the comment, but I don't think I can improve it further without using additional variable definitions for the fonts and element functions.

docs/guides/guide-for-latex-users.md
```diff
@@ -75,21 +75,23 @@ Emphasis (usually rendered as italic text) is expressed by enclosing text in
...
+| Element                | LaTeX                     | Typst                  | See        |
+|:-----------------------|:--------------------------|:-----------------------|:-----------|
+| Strong emphasis        | `\textbf{strong}`         | `[*strong*]`           | [`strong`] |
+| Emphasis               | `\emph{emphasis}`         | `[_emphasis_]`         | [`emph`]   |
+| Link                   | `\url{https://typst.app}` | `[https://typst.app/]` | [`link`]   |
+| Label                  | `\label{intro}`           | `[<intro>]`            | [`label`]  |
+| Reference              | `\ref{intro}`             | `[@intro]`             | [`ref`]    |
+| Citation               | `\cite{humphrey97}`       | `[@humphrey97]`        | [`cite`]   |
+| Verbatim / code        | `\verb|print(f"{x}")|`, `verbatim` / `listing` environments | ``[`print(f"{x}")`]`` | [`raw`] |
+| Monospace / typewriter | `\texttt{mono}` | ``[`mono`]`` or `text` function | [`raw`], [`text`] |
```
PgBiel, line 87: Thoughts on this (and the accompanying comment below the table)?

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,38 @@ use crate::World;
...
+/// # Styling
+/// By default, the `raw` element uses the `DejaVu Sans Mono` font, available
+/// by default in Typst, with a smaller font size of `0.8em` (that is, 80% of
+/// the global font size). This is because monospace fonts tend to be visually
+/// larger than non-monospace fonts.
+///
+/// You can customize these properties with show-set rules:
+///
+/// ````example
+/// // Switch to Cascadia Code for
```
laurmaedje, line 74: Cascadia Code is not available in the docs. As-is, this triggers font fallback. The available fonts are listed here: https://github.com/typst/typst-dev-assets/tree/main/files/fonts

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,38 @@ use crate::World;
 /// #raw("fn " + "main() {}", lang: "rust")
 /// ```
 ///
+/// # Styling
```
laurmaedje, line 65: Maybe we want to move this below the "Syntax" section because that one is a bit more basic?

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,38 @@ use crate::World;
...
+/// By default, the `raw` element uses the `DejaVu Sans Mono` font, available
+/// by default in Typst, with a smaller font size of `0.8em` (that is, 80% of
```
laurmaedje, line 67:
```suggestion
/// By default, the `raw` element uses the `DejaVu Sans Mono` font (included
/// with Typst) with a smaller font size of `{0.8em}` (that is, 80% of
```
"default" appears twice in the sentence, which is not so elegant. maybe something like this would read better.

crates/typst-library/src/text/raw.rs
laurmaedje: There is no other monospace font in there atm. I guess we can add Cascadia Code there.

docs/guides/guide-for-latex-users.md
```diff
@@ -75,21 +75,23 @@ Emphasis (usually rendered as italic text) is expressed by enclosing text in
...
-| Equation         | `$x$`, `align` / `equation` environments | `[$x$]`, `[$ x = y $]` | [`equation`]($math.equation) |
+| Element                | LaTeX                     | Typst                  | See        |
+|:-----------------------|:--------------------------|:-----------------------|:-----------|
+| Strong emphasis        | `\textbf{strong}`         | `[*strong*]`           | [`strong`] |
+| Emphasis               | `\emph{emphasis}`         | `[_emphasis_]`         | [`emph`]   |
+| Link                   | `\url{https://typst.app}` | `[https://typst.app/]` | [`link`]   |
+| Label                  | `\label{intro}`           | `[<intro>]`            | [`label`]  |
+| Reference              | `\ref{intro}`             | `[@intro]`             | [`ref`]    |
+| Citation               | `\cite{humphrey97}`       | `[@humphrey97]`        | [`cite`]   |
+| Verbatim / code        | `\verb|print(f"{x}")|`, `verbatim` / `listing` environments | ``[`print(f"{x}")`]`` | [`raw`] |
```
laurmaedje, line 86: The pipes after `\verb` messes with Markdown.

docs/guides/guide-for-latex-users.md
laurmaedje, line 87: I think `text` should come first and `mono` second. The links in "See" should also go there instead of `raw`, `text`.

docs/guides/guide-for-latex-users.md
```diff
@@ -75,21 +75,23 @@ Emphasis (usually rendered as italic text) is expressed by enclosing text in
...
+| Monospace / typewriter | `\texttt{mono}` | ``[`mono`]`` or `text` function | [`raw`], [`text`] |
+| Verbatim               | `verbatim` environment    | ``[`#typst-code()`]``  | [`raw`]    |
```
laurmaedje, line 88: It's a bit confusing that verbatim is in here twice.

docs/guides/guide-for-latex-users.md
```diff
@@ -121,6 +123,16 @@ To get a [numbered list]($enum) (`enumerate`) instead, use a `+` instead of the
...
 
+Regarding the usage of monospace fonts (also known as "typewriter" font style
+in LaTeX), it should be noted that using [`raw`] such as in
+``[`monospace`]`` (where you'd use `\texttt{monospace}` in LaTeX) works for
+most cases where you only have simple text. If you need to use formatting, such
+as in `\texttt{monospace \textbf{bold}}`, you will need to replicate its look by
+switch to changing the text font to a monospace font with
+`#text(font: "DejaVu Sans Mono", size: 0.8em)[monospace *bold*]`, for example,
+since `raw` only supports verbatim (unformatted) text. See its documentation
+for more details.
```
laurmaedje, line 134: I think this would benefit from being shortened a bit. As-is, it's a bit verbose.

crates/typst-library/src/text/raw.rs
joelvonrotz: Maybe indicate, that the font must be installed?

docs/guides/guide-for-latex-users.md
joelvonrotz: I think this could be displayed more _to the point_. Following is just an idea, although which could be structured better.

````markdown
Regarding the usage of monospace fonts (also known as "typewriter" font style
in LaTeX), it should be noted that using [`raw`] such as in
``[`monospace`]`` (where you'd use `\texttt{monospace}` in LaTeX) works for
verbatim (unformatted) text. If you need to use formatting, such
as in `\texttt{monospace \textbf{bold}}`, you will need to replicate its look by using the [`text`]($text)-function:

```typst
#text(font: "DejaVu Sans Mono", size: 0.8em)[monospace *bold*]
```

See its documentation for more details.
````

A LaTeX user probably just wants to see the comparison from LaTeX to Typst!

File: crates/typst-library/src/text/raw.rs (hunk 6 above)
Comment (laurmaedje) [reply crates/typst-library/src/text/raw.rs:74]: The docs will show a pre-rendered preview and that preview will be broken regardless of the user's system fonts.

**Comment by laurmaedje** [comment]
@PgBiel Just a small reminder that the PR is still open. :)

**Comment by PgBiel** [comment]
Good call, I'll get back to it soon :+1:

**Comment by hpcfzl** [comment]
I took inspiration from both of you (@laurmaedje @joelvonrotz) and came up with this:
```diff
  | Citation               | ...
+ | Monospace (typewriter) | `\texttt{mono}`        | `text` ^1             | [`text`] |
+ | Code                   | `listings` environment | `print("f{x}")`       | [`raw`]  |
+ | Verbatim               | `verbatim` environment | ``[`#typst-code()`]`` | [`raw`]  |
  | Bullet list            | ...

Lists ...
Nesting lists ...
To get a ...

+ ^1 Regarding the usage of monospaced fonts (also known as "typewriter" font style in LaTeX), 
+ it should be noted that [`raw`] treats all input as verbatim and therefore acts differently 
+ from \texttt{monospace} in LaTeX. To replicate only the look of [`raw`], such that its 
+ output appears as formatted monospace, you need to use the [`text`]($text) function with a 
+ monospaced font. When in math mode, you should use the provided [`mono`]($mono) function. 
+ So the Typst equivalent of `\texttt{monospace \textbf{made bold}}` outside math mode would then be:
+ ```typst
+ #text(font: "DejaVu Sans Mono", size: 0.8em)[monospace *made bold*]
+ `monospace *not bold*`
+ ```
+ The exact font family and size for replicating [raw] style can also be found on its reference page.

## How do I use a ...
```
The `^1` would be an annotation like the one in [outline](https://typst.app/docs/reference/model/outline/#building-an-entry).

crates/typst-library/src/text/raw.rs
```diff
@@ -56,6 +62,38 @@ use crate::World;
...
+/// the global font size). This is because monospace fonts tend to be visually
+/// larger than non-monospace fonts.
```
YDX-2147483647, line 93: Could we also mention that [`text.cjk-latin-spacing`](https://typst.app/docs/reference/text/text/#parameters-cjk-latin-spacing) is `none` here?

```suggestion
/// larger than non-monospace fonts.
/// In addition, advanced typesetting features will be turned off by default,
/// including [`cjk-latin-spacing`]($text.cjk-latin-spacing), [`overhang`]($text.overhang), and [`hyphenate`]($text.hyphenate).
/// In raw blocks, [justification]($text.justify) will also be disabled.
```

Many tools/packages use `show raw.where(lang: "…"): it => eval(it.text)` to render examples.
These subtleties will affect reproduction.
https://github.com/typst/typst/issues/6539#issuecomment-3026389459 is an example.

crates/typst-library/src/text/raw.rs
PgBiel: I think using Cascadia Code or Fira Code (perhaps easier license-wise) would be a good idea to showcase using a different monospace font.

docs/guides/guide-for-latex-users.md
PgBiel: I've applied @joelvonrotz's suggestion with a few modifications. Have a look :slightly_smiling_face:

docs/guides/guide-for-latex-users.md
PgBiel: I've changed it according to @hpcfzl's suggestions. Should be a bit better now

crates/typst-library/src/text/raw.rs
laurmaedje: Cascadia Code  is okay license-wise. Could you add it to typst-dev-assets? (To files/fonts, src/lib.rs, and NOTICE)

docs/guides/guide-for-latex-users.md
```diff
@@ -121,6 +123,19 @@ To get a [numbered list]($enum) (`enumerate`) instead, use a `+` instead of the
 hyphen. For a [term list]($terms) (`description`), write `[/ Term: Description]`
 instead.
 
+Regarding monospaced fonts (or "typewriter" style in LaTeX), using the [`raw`]
```
laurmaedje: IIUC @hpcfzl intended for this to be a footnote?

hpcfzl: I did, though in the currently accepted version, that would cause the `text or mono functions` table cell (or any other in this row) to become even wider. I intentionally wanted to avoid that in mine.

Issue #5538, opened 2024-12-06
Documentation:  Add a section to the context page which describes important details of how the context is evaluated.


gamagan:
### Description

# Request 

The `context` page of the docs site could benefit from an additional section, which details low-level behavior of how the `context` object operates. 
https://typst.app/docs/reference/context/

State management in Typst is very different from how it is done in most other software packages. The implementation makes sense in the context of Typst, but it can be challenging to understand for users not accustomed to it.

The Typst `context` is a concept that is technical, and it is technical enought that, in generaly, it's not possible to just explain it at a high level. Because of the nature of state management in Typst, low-level details cannot be escaped. Instead of hiding them, they should be made clear, so that it can bring clarity to how Typst processes the data related to contextual updates. This way, using the context will seem less magical.

To help with the documentation process, below is some info from the Typst discord, as described by user `PgSuper`.

# Info

The way counters work is that it does basically a `query()` of all counter `.step()` and `.update()` until the location where you run `counter.get()`, adding 1 for each `step()` and updating accordingly, returning the calculated value.

But, what is the "location" where you run `counter.get()`? A location is always tied to an element in the document, e.g. when you place a heading, you can refer to where that heading appears in the document.

A location is not something like x=29345, y=203  -- that's an important distinction. A location is like a position in the compiler's element tree, so to speak.

A `context { }` section creates an element which must be placed in the document so that it gains a location. It then provides that location to `.get()` so it knows:  query all steps until that location.

# Example Code

Consider 2 similar function definitions, where the idea is the following: 
Every time the `GetCounter()` function is called, it first updates a global `counter` variable, and then returns the current value of that variable, concatenated to a prefix string and suffix string. 

Thus, successive calls to `GetCounter()` should return: 
        (-1-), (-2-), ..., (-n-)

```typst
#let gCounter = counter("my-counter")

// The context is outside of the function's body, so the ".step()" is a child of the generated context element.
// This function does not achieve its goal.
#let GetCounter1() = context {
    gCounter.step()
    let num = gCounter.get().last()
    return "(-" + str(num) + "-)"
}

// The counter's ".step()" call is outside of the context.
// This is the correct way to implement the GetCounter() functionality described above.
#let GetCounter2() = {
    gCounter.step()

    // note: no return
    context {
        let num = gCounter.get().last()
        [ (-  #num - ) ]
    }
}

```

The first function is conceptually similar to this:

```typst
// Notice that the context provided by the callback doesn't change inside its body:
[...]

PR #6008, opened 2025-03-04
Revision of the Context documentation

ukoethe:
This is a pull request to discuss documentation improvements. I edited most of the `context` documentation to give an idea what I would find helpful. Please use the inline comment function for discussion.


docs/reference/language/context.md
```diff
@@ -26,37 +26,51 @@ in some places that are also aware of their location in the document:
...
 #context text.lang
 ```
 
-As explained above, a context expression is reactive to the different
-environments it is placed into. In the example below, we create a single context
-expression, store it in the `value` variable and use it multiple times. Each use
-properly reacts to the current surroundings.
+Note that calling `#text.lang` directly would be an error, because the request
+cannot be answered without knowledge of the context. The fields supported by
+a given element function are documented **where??** and can be retrieved in a
```
MDLC01, line 43: The fields are just the named parameters documented on each element's page.

docs/reference/language/context.md
```diff
@@ -26,37 +26,51 @@ in some places that are also aware of their location in the document:
...
 ```
 
-As explained above, a context expression is reactive to the different
-environments it is placed into. In the example below, we create a single context
-expression, store it in the `value` variable and use it multiple times. Each use
-properly reacts to the current surroundings.
+Note that calling `#text.lang` directly would be an error, because the request
+cannot be answered without knowledge of the context. The fields supported by
+a given element function are documented **where??** and can be retrieved in a 
+document by calling **what?? (something like `fields()`)**.
```
MDLC01, line 44: I don't think there is a way to dynamically get the list of available fields of an element in the general case.

docs/reference/language/context.md
```diff
@@ -26,37 +26,51 @@ in some places that are also aware of their location in the document:
...
 #set text(lang: "fr")
-#value
+#context text.lang
 ```
 
-Crucially, upon creation, `value` becomes opaque [content] that we cannot peek
-into. It can only be resolved when placed somewhere because only then the
-context is known. The body of a context expression may be evaluated zero, one,
-or multiple times, depending on how many different places it is put into.
+The output of a `#context ...` call is _read-only_ (in the form of `[content]`).
```
MDLC01, line 60:
```suggestion
The output of a `#context ...` call is _read-only_ (in the form of `content`).
```

I'm not sure what you intended here, but I believe the brackets are unnecessary.

docs/reference/language/context.md
```diff
@@ -26,37 +26,51 @@ in some places that are also aware of their location in the document:
...
+end of the context's scope:
+
+```example
+#context {
+  // the context allows you to retrieve the current text.size
+  set text(size: text.size * 200%)
+  [large text] 
+}
+original size
+```
```
MDLC01, line 73: I think this part could be improved. I'm not sure what you mean by "read-only". What would the user expect to be modifiable that is not?

I like the example, though.

MDLC01 (commented):
I prefer the old version of the last paragraph, but appart from that I like the changes.

docs/reference/language/context.md
ukoethe: They were in the original version, though ...

docs/reference/language/context.md
ukoethe: That's sad. Looks natural when you come from Python.

docs/reference/language/context.md
ukoethe: Took this remark into my revised version.

docs/reference/language/context.md
ukoethe: I hope the new version is an improvement.

docs/reference/language/context.md
MDLC01: Oh i see. That was to turn it into a link. You can make it code and a link with `` [`content`] `` instead.

ukoethe:
A remark to the documentation of https://typst.app/docs/reference/math/equation/#parameters-numbering: the documentation says that the default is `none` but does not explain what it means - no number or default numbering? This would become clear if there was also an option to pass `auto` to request default numbering. Is there a good reason not to support `auto`?

PgBiel:
Haven't taken an in-depth look yet, but I'll defend beforehand that we should make sure any changes to the context docs take this issue into account (and potentially close it): #5538

ukoethe:
#5538 stresses: "The context element always gets information from before the block started."

This PR has several places explaining this behavior with various examples, e.g.

- `text.lang` is assigned upon entry in the context and remains constant until the end of its scope.
- The function `counter.display()` is an example for this behavior. Below, call A will access the counter's value upon _entry_ into the context, i.e. '1' - it cannot see the effect of `{c.update(2)}`.

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
-## Style context
-With set rules, we can adjust style properties for parts or the whole of our
-document. We cannot access these without a known context, as they may change
-throughout the course of the document. When context is available, we can
-retrieve them simply by accessing them as fields on the respective element
-function.
+## Behavior of the context keyword
+Style properties frequently change within a document, for example by applying set 
+rules. To retrieve such poperties in a consistent way, one must first specify 
+the context where the query is to be executed. This is the purpose of the
```
laurmaedje, line 31: the term "query" has a rather specific connotation in Typst and should be avoided here imo (also a bunch of times more below)

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
-With set rules, we can adjust style properties for parts or the whole of our
-document. We cannot access these without a known context, as they may change
-throughout the course of the document. When context is available, we can
-retrieve them simply by accessing them as fields on the respective element
-function.
+## Behavior of the context keyword
+Style properties frequently change within a document, for example by applying set 
+rules. To retrieve such poperties in a consistent way, one must first specify 
+the context where the query is to be executed. This is the purpose of the 
+`context` keyword. Once the context has been fixed, the property information
```
laurmaedje, line 32: Saying "This is the purpose of the context keyword" is a bit strong since it's not the only purpose of it

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
-document. We cannot access these without a known context, as they may change
-throughout the course of the document. When context is available, we can
-retrieve them simply by accessing them as fields on the respective element
-function.
+## Behavior of the context keyword
+Style properties frequently change within a document, for example by applying set 
+rules. To retrieve such poperties in a consistent way, one must first specify 
+the context where the query is to be executed. This is the purpose of the 
+`context` keyword. Once the context has been fixed, the property information 
+is available through a simple field access syntax. For example, `text.lang`
```
laurmaedje, line 33:
```suggestion
is available through field access syntax. For example, `text.lang` 
```

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
 #set text(lang: "de")
+// query the language setting "here"
 #context text.lang
 ```
 
-As explained above, a context expression is reactive to the different
-environments it is placed into. In the example below, we create a single context
-expression, store it in the `value` variable and use it multiple times. Each use
-properly reacts to the current surroundings.
+Note that calling `#text.lang` directly would be an error, because the request
```
laurmaedje, line 43: "calling" isn't really the correct term I think, as it's not a function (this also appears various times more below)

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+Moreover, some functions, such as [`to-absolute`]($length.to-absolute) 
+and [`counter.display`]($counter.display) are only applicable in a context, 
+because their results depend on the current settings of style properties. 
+When another function `foo()` calls a context-dependent function, it becomes 
+itself context-dependent:
+
+```example
+#let foo() = 1em.to-absolute()
+#context {
+  // foo() cannot be called outside of a context
```
laurmaedje, line 57: if managable, please try to keep examples to <=38 columns so that they will display side-by-side in the docs and are easy to read on mobile

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
-#value
+#set text(lang: "en")
+#context text.lang
 
 #set text(lang: "de")
-#value
+#context text.lang
+```
+
+The output of a `#context ...` call is static in the form of opaque
```
laurmaedje, line 72:
```suggestion
The value of a `#context ...` expression is static in the form of opaque 
```

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
 
 #set text(lang: "de")
-#value
+#context text.lang
+```
+
+The output of a `#context ...` call is static in the form of opaque 
+[`content`]. Write access to context output is prohibited, as it would 
+often result in invalid code: If the context changes between read and 
+write, overwriting a property would cause an inconsistent system state.
```
laurmaedje, line 75: This doesn't read quite correctly. The crucial thing is that contextual content might appears 0, 1, or N times in the document. There is some more discussion of this in the second part of this blog post: https://laurmaedje.github.io/posts/types-and-context/

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+#set text(lang: "en")
+#context [
+  call 1: #text.lang \
+
+  #set text(lang: "fr")
+  call 2: #text.lang
+]
+```
+
+Both calls have the same output 'en', because `text.lang` is assigned
```
laurmaedje, line 89:
```suggestion
Both calls have the same output "en", because `text.lang` is assigned
```
We don't really use single quotes except in nested quote situations.

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+[`content`]. Write access to context output is prohibited, as it would 
+often result in invalid code: If the context changes between read and 
+write, overwriting a property would cause an inconsistent system state. 
+In fact, context-dependent property fields are immutable constants even 
+within the context itself:
+
+```example
+#set text(lang: "en")
+#context [
+  call 1: #text.lang \
```
laurmaedje, line 82:
```suggestion
  call 1: #text.lang
```
there is already a paragraph break

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+```
+
+Both calls have the same output 'en', because `text.lang` is assigned
+upon entry in the context and remains constant until the end of its scope 
+(the closing `]`). It does not "see" the `#set text(lang: "fr")` before 
+call 2. Compare this to the previous example: there we got two different 
+results because we created two different contexts.
+
+However, immutability only applies to the property fields themselves. 
+Content creation instructions within a context _do_ see the effect of
```
laurmaedje, line 96: "Content creation instructions" is a bit of a mouthful, maybe just "content expressions"

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+  call 2: #text.size \
+  call 3: #context text.size
+]
+```
+
+## Controlling content creation within a context
+
+The main purpose of retrieving the current values of properties is, 
+of course, to use them in the calculation of derived properties, 
+instead of setting those properties manually. For example, you can
```
laurmaedje, line 148: I wouldn't say that's accurate. It's surely one use case, but you might just as well use it for something else.

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+Since set rules are only active until the end of the enclosing scope, 
+'original size' is printed with the original font size.
+The above example is equivalent to 
+
+```example
+#{
+  set text(size: 2em)
+  [large text \ ]
+}
+original size
```
laurmaedje, line 170: The start of a phrase in example should be capitalized. E.g. "Original size" or "Call 1" (bad example because call is not the correct term, but just an example).

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+normal spacing:
+$
+x \
+x
+$
+doubled spacing:
+#spaced-eq(spacing: 200%)[$
+z \
+z
+$]
```
laurmaedje, line 200: This example is a bit to complex for my liking. I think the point can be made in a clearer way.

docs/reference/language/context.md
```diff
@@ -25,38 +25,180 @@ in some places that are also aware of their location in the document:
...
+#context [
+  call 1: #text.size \
+
+  #show: template
+  call 2: #text.size \
+  call 3: #context text.size
+]
+```
+
+## Controlling content creation within a context
```
laurmaedje, line 144: I'm not quite sure what this title means

docs/reference/language/scripting.md
```diff
@@ -7,6 +7,16 @@ Typst embeds a powerful scripting language. You can automate your documents and
 create more sophisticated styles with code. Below is an overview over the
 scripting concepts.
 
+## Identifiers
+
```
laurmaedje, line 11: The markdown style in this repo is to not put blank lines after headings.

docs/reference/language/scripting.md
```diff
@@ -7,6 +7,16 @@ Typst embeds a powerful scripting language. You can automate your documents and
...
+Typst uses largely the same rules to define variables, function names etc.
+as most other languages, with one important exception: dashes are permitted
```
laurmaedje, line 13:
```suggestion
as most other languages, with one important exception: Dashes are permitted
```

docs/reference/language/scripting.md
```diff
@@ -7,6 +7,16 @@ Typst embeds a powerful scripting language. You can automate your documents and
...
+as part of an identifier, e.g. the [`to-absolute`]($length.to-absolute) 
+function for a length or the [`first-line-indent`]($par.first-line-indent) 
+parameter of a paragraph. In fact, this convention is more common than 
+underscores or camel case. Since the dash is also the symbol for subtraction,
```
laurmaedje, line 17: "In fact, this convention is more common than underscores or camel case."

I'm not a fan of it because it's not just more common, it's just the canonical case convention in Typst.

docs/reference/language/scripting.md
```diff
@@ -245,6 +259,10 @@ The value in question can be either:
   [element function]($function/#element-functions) that were given when the
   element was constructed.
 
+**Important:** Many fields of element functions are only accessible in a
```
laurmaedje, line 262:
```suggestion
Many fields of element functions are only accessible in a
```

docs/reference/language/scripting.md
```diff
@@ -245,6 +259,10 @@ The value in question can be either:
...
+concrete context that disambiguates their meaning. Consult the chapter
```
laurmaedje, line 263: "disambiguates their meaning" doesn't strike me as just the right way to put it. Maybe "are only accessible in contexts where their concrete value is known."

laurmaedje:
> This would become clear if there was also an option to pass auto to request default numbering. Is there a good reason not to support auto?

None means no numbering, as usual. There is no need for `auto` since there is no default numbering.

laurmaedje (commented):
A left a few remarks. Reviewing docs is challenging as it's a lot more subjective than code. I kept it mostly to high-level conceptual remarks and Markdown style things and did not go into minor edits. That can be left for later or maybe I would just push some minor edits at the end once the higher level things are out of the way.

docs/reference/language/context.md
ukoethe: I deliberately use `query` when I refer to the specific typst meaning, and plain text to just use the English term. IMHO, there is no confusion. Anyway, what synonymous word would not collide with anything in your opinion?

docs/reference/language/context.md
ukoethe: Please enlighten me about other purposes (or give me a pointer).

docs/reference/language/context.md
ukoethe: Thanks. Was wondering what the maximum width is...

docs/reference/language/context.md
ukoethe: I'm not happy with this sentence either. However, I would like to have a one-sentence explanation why things need to be more complicated than what one would naively expect. Talking about "closures" etc. as in your post doesn't feel right, because many users won't know this term. Any suggestions?

docs/reference/language/context.md
ukoethe: It meant to say that a `context` provides a lot more power to configure and control content creation programmatically because one has access to the parameter fields. The section is supposed to demonstrate this.

docs/reference/language/scripting.md
ukoethe: What I mean is: Parameter values can be different in every closure, so accessing them without specifying a closure would be ambiguous. "Where their value is known" suggests that they have no value in most situations, but actually it is the opposite: usually, there are many alternative values that would fit the query (e.g. fonts for headings, text, footnotes etc), so one has to be more specific. In my mind, set rules for different content types are all simultaneously alive, just not simultaneously active.

docs/reference/language/context.md
ukoethe: I feel that "Content creation instructions" is a less technical term, so maybe better for newcomers.

docs/reference/language/context.md
ukoethe: I should definitely mention this example:
```typ
// make text darker
#set text(fill: context text.fill.darken(20%))
```

ukoethe: However, on first glance this example contradicts the statement that the value of a `context` is `content`: Here, it seems to be a ytpe that can be assigned to the fill parameter. For example, this does not work
```typ
#set text(fill: [something])
```
So, clearly the output of `context text.fill.darken(20%)` is not `content`.

docs/reference/language/context.md
ukoethe: How about "The output of a context expression"?

See also my remark below questioning if the value/output of a context expression is really always `content`.

docs/reference/language/context.md
ukoethe: I'm running out of words -- "query" is also forbidden :-) How about "retrieving"?

docs/reference/language/context.md
ukoethe: You should be proud of how simple it is (in comparison to the competition :-).

docs/reference/language/context.md
ukoethe: I gave up enforcing so detailed style guides in my OS projects -- was just not worth the trouble.

ukoethe:
> None means no numbering, as usual. There is no need for `auto` since there is no default numbering.

When I first read the documentation, I had no idea about what none means `as usual`, so I would have appreciated an explanation (instead, I figured it out by trial-and-error). 

Regarding `auto`, I was not clear enough: "default numbering" is supposed to mean "do not change the numbering scheme". Not having `auto` means that one must write a conditional when one wants to configure equations automatically (i.e. by calculating parameters as a function of other settings) and a possible outcome of the calculation is "keep the numbering setting as is". In the presence of `auto`, one could always pass an argument to `numbering:`, so the code might be simpler.

docs/reference/language/context.md
ukoethe: Would it be better to write `$ x \ x $` ?

docs/reference/language/context.md
MDLC01: The value of `text.fill.darken(20%)` is not content, but the value of `context text.fill.darken(20%)` is.

MDLC01 (commented):
I like the improvements, and I mostly agree with Laurenz's comments.

I still dislike the word "immutable", though. But I'm not sure what to replace it with, other than changing the phrasing. "Constant", like used line 263, is great.

docs/reference/language/context.md
laurmaedje: > I should definitely mention this example:

This example doesn't work yet, it was just a proposed idea, so it shouldn't be mentioned in the docs.

docs/reference/language/context.md
```diff
@@ -69,135 +71,154 @@ When a property is changed, the response to the query changes accordingly:
...
-$]
+Normal spacing:
+$ x \ x $
+Doubled spacing:
+#spaced(spacing: 200%)[$ z \ z $]
+```
+
+The advantage of this technique is that the user does not have to know the 
+original spacing in order to double it. To double the spacing of all 
+equations, you can put the same calculations in a `show` rule. Note that
```
MDLC01, line 205:
```suggestion
equations, you can put the same calculations in a show rule. Note that 
```

The documentation usually uses this convention.

docs/reference/language/context.md
```diff
@@ -262,18 +283,18 @@ demonstrates this:
 
 The rule that context-dependent variables and functions remain constant 
 within a given `context` also applies to location context. The function
-`counter.display()` is an example for this behavior. Below, call A will 
-access the counter's value upon _entry_ into the context, i.e. '1' - it 
-cannot see the effect of `{c.update(2)}`. In contrast, call B accesses 
+`counter.display()` is an example for this behavior. Below, read A will
```
MDLC01, line 286:
```suggestion
[`counter.display`] is an example for this behavior. Below, read A will 
```

This makes it a link to the function's documentation page.

MDLC01, line 286:
```suggestion
`counter.display()` is an example for this behavior. Below, Read A will 
```

Same below.

docs/reference/language/context.md
```diff
@@ -69,135 +71,154 @@ When a property is changed, the response to the query changes accordingly:
...
+  Read 1: #text.lang \
 
   #set text(lang: "fr")
-  call 2: #context text.lang
+  Read 2: #context text.lang
 ]
 ```
 
-All of the above applies to `show` rules analogously, for example:
+All of the above applies to `show` rules analogously. To demonstrate this,
```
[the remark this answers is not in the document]
ukoethe: I have no idea either.

docs/reference/language/context.md
```diff
@@ -69,135 +71,154 @@ When a property is changed, the response to the query changes accordingly:
...
 
   #set text(lang: "fr")
-  call 2: #text.lang
+  Read 2: #text.lang
 ]
 ```
 
-Both calls have the same output 'en', because `text.lang` is assigned
+Both reads have the same output `"en"`, because `text.lang` is assigned
 upon entry in the context and remains constant until the end of its scope
```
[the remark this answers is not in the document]
ukoethe: While your explanation may be technically more precise, I think it is harder to understand for newcomers. I don't see why "assigned upon entry into the context" is wrong and would rather like to keep it. May "entry into the context block", but I'm not sure of this is better.

docs/reference/language/context.md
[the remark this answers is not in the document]
ukoethe: I changed it back to "applies" by majority vote, but Gemini thinks that "apply" was correct after all.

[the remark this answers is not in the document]
ukoethe: DeepL favors "All of the above applies", Google translate opts for "Everything above applies"

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+with the `context` keyword. Once the context has been fixed, the property 
+information is available through standard field access syntax. For example, 
+`text.lang` asks for the current language setting. In its simplest form, the 
+`context` keyword refers to "right here":
 
 ```example
 #set text(lang: "de")
+// read the language setting "here"
 #context text.lang
 ```
```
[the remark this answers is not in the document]
PgBiel: In fact, we could consider mentioning this issue very early on below, when saying that context returns content, if possible. (Needs further discussion.)

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+#show math.equation.where(block: true): it => {
+  // access current par.leading in a context,
+  // established automatically by the show rule
+  set par(leading: par.leading * 200%)
+  it
+}
+
+Doubled spacing:
+$ z \ z $
+```
```
PgBiel, line 222: This has the unfortunate pitfall of not being overridable by a later set rule, since an earlier show rule executes last... it's also why `show: set` rules are usually recommended instead.

Of course, that's not a problem if you never try to override it, but wanted to highlight this for discussion.

docs/reference/language/scripting.md
```diff
@@ -7,6 +7,16 @@ Typst embeds a powerful scripting language. You can automate your documents and
 create more sophisticated styles with code. Below is an overview over the
 scripting concepts.
 
+## Identifiers
```
PgBiel, line 10: Not sure if this change belongs in this PR, but I'll let @laurmaedje have the final word on this.

docs/reference/language/context.md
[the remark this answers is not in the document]
PgBiel: Actually, this is already _briefly_ mentioned at the top, but either way people generally skim the docs for examples they may paste elsewhere (for good reason), so I'd argue this is still a concern, and an explicit (counter-)example would be nice.

PgBiel (commented):
In general, I like the idea of explaining these concepts in a more elementary way, potentially making this doc page more accessible to newcomers.

Here are some comments I had after an initial read at the current state.

I'll also note that @MDLC01's comments were generally spot on, thanks for considering them.

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+As you see, the result of a `#context ...` expression can 
+be inserted into the document as `content`. Context blocks can 
+contain arbitrary code beyond the field access. However,
+and this is often surprisingly for newcomers, context-dependent 
+property fields remain _constant_ throughout the context's scope. 
+This has two important consequences: First, direct property 
+assignments like `text.lang = "de"` are _not_ allowed &ndash; 
+use `set` or `show` rules for this purpose. Second, changes to a 
+property value within a context (e.g. by a `set` rule) are not 
+observable by field access within that same context:
```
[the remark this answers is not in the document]
ukoethe: > This mention to property assignment (at least here) is confusing, since that was never suggested to be possible in the first place

The problem with your argument is that anyone with the least programming experience assumes by default that properties can be assigned -- it is just how it works in most languages. That typst never suggested the possibility is irrelevant, the brain automatically relates new stuff to what it already knows.

In fact, a major reason why I struggled with the typst documentation (and take the time to propose improvements) is that one has to unlearn various things that other programming languages take for granted. You absolutely must mention and explain when typst deviates from the mainstream.

docs/reference/language/context.md
ukoethe: You mean the example should be written like this?
```typ
#show math.equation.where(block: true): set par(leading: par.leading * 200%)
```

This is a typical example of a problem I noticed with the current documentation: After reading it, one tends to overcomplicate solutions.

docs/reference/language/scripting.md
ukoethe: Not sure if it is worth moving this to a PR of its own. In any case, this is another place where it is necessary to unlearn experience from other programming languages, and therefore important to mention.

docs/reference/language/context.md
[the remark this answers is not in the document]
ukoethe: This example is the "Hello world" of contexts. Everyone knows that "Hello world" is not very useful, but it gets across a few very basic things (such as how the context keywords looks like, that properties are accessed by field syntax, and that they can be displayed). Of course, a different example (instead of `lang`) might be more instructive.

docs/reference/language/context.md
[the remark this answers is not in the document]
PgBiel: Yes, I agree this should be mentioned, but my main point is that this isn't related to `context` itself and feels out of place. This is rather related to `set` rules themselves - that's how you set properties on elements, not through assignment.

In other words, since `set` rules from outside are what define the value of `text.lang` in the first place, one could naturally expect that a `set` rule inside `context { ... }` would also change it within. That is not the case.

As a compromise, the suggestion of not being able to assign to `text.lang` could be moved to a note after the phrase or paragraph, e.g. "set rules cannot change `text.lang` within the context (nor can you assign to them, as only set rules can influence the default values of element properties)."

docs/reference/language/context.md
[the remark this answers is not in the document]
ukoethe: BTW, the example is not completely pointless -- it is useful as a quick-and-dirty way of inspecting what the current value of some property is.

docs/reference/language/context.md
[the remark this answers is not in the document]
ukoethe: > In other words, since set rules from outside are what define the value of text.lang in the first place, one could naturally expect that a set rule inside context { ... } would also change it within. That is not the case.

Isn't this exactly what my proposed documentation says?

docs/reference/language/context.md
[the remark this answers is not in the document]
ukoethe: A remark about the error message when a field is accessed outside of a context:
```typ
#text.lang    // => "Can only be used when context is known."
```
This message is a mystery for beginners -- when you haven't heard about the `context` concept and read "context is known" with its generic, everyday-language meaning, you have no idea what to do (and this unfortunately applies to many other error messages as well). Worse, when I first opened the context documentation, I got the feeling that this is meant for advanced users and wouldn't be needed right now. In reality, contexts are absolutely fundamental, and an improved documentation should get this across.

docs/reference/language/scripting.md
```diff
@@ -7,6 +7,16 @@ Typst embeds a powerful scripting language. You can automate your documents and
...
+Typst uses largely the same rules to define variables, function names etc.
+as most other languages, with one important exception: Dashes are permitted
+as part of an identifier, e.g. the [`to-absolute`]($length.to-absolute) 
+function for a length or the [`first-line-indent`]($par.first-line-indent) 
+parameter of a paragraph. In fact, this is the canonical convention &ndash;
+underscores or camel case are not used in "official" typst. Since the dash
```
bastienvoirin, line 16:
```suggestion
underscores or camel case are not used in "official" Typst. Since the dash 
```

Per https://github.com/typst/typst?tab=readme-ov-file#pronunciation-and-spelling:

> When writing about Typst, capitalize its name as a proper noun, with a capital "T".

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+#set text(lang: "en")
+#context [
+  Read 1: #text.lang
+
+  #set text(lang: "fr")
+  Read 2: #context text.lang
+]
+```
+
+All of the above applies to `show` rules analogously. To demonstrate this,
```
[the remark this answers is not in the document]
PgBiel: Otherwise they don't affect context at all. That's the only way they could potentially change a property (and, indeed, they affect nested contexts). But a show rule transforming element A into B wouldn't have any impact whatsoever.

docs/reference/language/context.md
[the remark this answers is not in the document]
PgBiel: > Isn't this exactly what my proposed documentation says?

I'm effectively proposing for that to be the _only_ thing it says, and mention the assignment thing elsewhere, or as a sidenote.

docs/reference/language/context.md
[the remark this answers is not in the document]
PgBiel: > This example is the "Hello world" of contexts. Everyone knows that "Hello world" is not very useful, but it gets across a few very basic things (such as how the context keywords looks like, that properties are accessed by field syntax, and that they can be displayed)

Yeah of course, but as someone who is often active in the typst support channels, i find the problem I mentioned to be rather common across beginners. See also https://forum.typst.app/t/why-is-the-value-i-receive-from-context-always-content/164/

It is possible to keep the example in some way, but maybe it just shouldn't be the _first_.

docs/reference/language/context.md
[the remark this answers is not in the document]
PgBiel: This is mostly due to how the `show` syntax is a bit overloaded. Only "everything show rules" - equivalent to passing the rest of a scope to a function - can also apply show rules on the current scope.

docs/reference/language/context.md
laurmaedje: I think "retrieve" is good. The difference between "query" and `query` is a bit too subtle imo.

docs/reference/language/context.md
laurmaedje: Context also provides the current location via `here()`.

docs/reference/language/context.md
laurmaedje: :D I'm trying to reduce the use of "simple" in my writing as things are differently simple/hard for different people and it's typically unnecessary fluff.

docs/reference/language/context.md
laurmaedje: Retrieving is good.

docs/reference/language/context.md
laurmaedje: "Content expressions" is indeed bad. I think my mind was thinking about "context expressions" when I wrote this. But the new version in the latest commit reads well.

docs/reference/language/context.md
laurmaedje: For example to localize a document via `context if text.lang == "de" [Inhalt] else [Contents]`. I think the use cases are very diverse, which is why I stumbled over the "of course"

docs/reference/language/scripting.md
laurmaedje: It stuck out to me because in my opinion it's not clearly more important than the other information in this section. 

Regardless of that, I think this section could be expanded to clarify exactly how it exactly works. Essentially
- fields for required arguments are always accessible
- fields for optional arguments that were provided in the element's constructor are accessible on a constructed element
- fields for optional arguments that weren't mentioned in the constructor are only available when the element is "materialized", this is the case when it's the shown element in a show rule or when the element is returned from a query
- If an element is not materialized, it's fields are not available, also not within a `context` expression

docs/reference/language/context.md
laurmaedje: > You mean the example should be written like this?
> ```typ
> #show math.equation.where(block: true): set par(leading: par.leading * 200%)
> ```

Unfortunately, this doesn't work because`par.leading * 200%` is evaluated when the show rule is first processed (without context), not when the equation is visited.

docs/reference/language/scripting.md
laurmaedje: I think it would be good to move this.

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
 outline, for instance, also provide the proper context to resolve counters.
 
-## Style context
-With set rules, we can adjust style properties for parts or the whole of our
-document. We cannot access these without a known context, as they may change
-throughout the course of the document. When context is available, we can
-retrieve them simply by accessing them as fields on the respective element
-function.
+## Behavior of the context keyword
+Style properties frequently change within a document, for example by applying set
```
laurmaedje, line 29: "by applying set rules" reads a bit clunkily to me. maybe "Style properties frequently change within a document, through use of set rules"

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
 
-## Style context
-With set rules, we can adjust style properties for parts or the whole of our
-document. We cannot access these without a known context, as they may change
-throughout the course of the document. When context is available, we can
-retrieve them simply by accessing them as fields on the respective element
-function.
+## Behavior of the context keyword
+Style properties frequently change within a document, for example by applying set 
+rules. To retrieve such properties in a consistent way, one must first specify
```
laurmaedje, line 30: not sure whether "consistent" is quite on point. Maybe "well-defined"?

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+// read the language setting "here"
 #context text.lang
 ```
 
-As explained above, a context expression is reactive to the different
-environments it is placed into. In the example below, we create a single context
-expression, store it in the `value` variable and use it multiple times. Each use
-properly reacts to the current surroundings.
+Note that any attempt to access `#text.lang` directly, i.e. outside of a context,
+will cause the compiler to issue an error message. The field names supported
```
laurmaedje, line 44: I think it's worth adding an extra sentence here explaining why the compiler has no choice but the emit an error: Because it needs to know at which location you want the property.

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
-#value
+#context text.lang
+```
 
-#set text(lang: "fr")
-#value
+As you see, the result of a `#context ...` expression can 
+be inserted into the document as `content`. Context blocks can 
+contain arbitrary code beyond the field access. However,
+and this is often surprisingly for newcomers, context-dependent
```
laurmaedje, line 77:
```suggestion
and this is often surprising for newcomers, context-dependent 
```

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+
+  #set text(lang: "fr")
+  Read 2: #text.lang
+]
+```
+
+Both reads have the same output `"en"`, because `text.lang` is assigned
+upon entry in the context and remains constant until the end of its scope 
+(the closing `]`). Thus, the `text.lang` field is not affected by
+#set text(lang: "fr")`, although Read 2 occurs afterwards. Compare
```
laurmaedje, line 98:
```suggestion
`[#set text(lang: "fr")]`, although Read 2 occurs after it. Compare 
```

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+  Read 1: #text.lang
+
+  #set text(lang: "fr")
+  Read 2: #text.lang
+]
+```
+
+Both reads have the same output `"en"`, because `text.lang` is assigned
+upon entry in the context and remains constant until the end of its scope 
+(the closing `]`). Thus, the `text.lang` field is not affected by
```
laurmaedje, line 97:
```suggestion
(the closing `]`). Thus, the second `text.lang` access is not affected by
```

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+  #set text(size: 25pt)
+  Read 2: #text.size
+]
 ```
 
-Crucially, upon creation, `value` becomes opaque [content] that we cannot peek
-into. It can only be resolved when placed somewhere because only then the
-context is known. The body of a context expression may be evaluated zero, one,
-or multiple times, depending on how many different places it is put into.
+Read 2 still outputs `40pt`, because `text.size` is a constant.
```
laurmaedje, line 116:
```suggestion
Read 2 still outputs `{40pt}`, because `text.size` is a constant.
```

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+$ x \ x $
+Doubled spacing:
+#spaced(spacing: 200%)[$ z \ z $]
+```
+
+The advantage of this technique is that the user does not have to know the 
+original spacing in order to double it. To double the spacing of all 
+equations, you can put the same calculations in a show rule. Note that 
+it is not necessary to add the `context` keyword on the right-hand side 
+of a `show` rule, because show rules establish a context automatically:
```
laurmaedje, line 207: The fact that show rules establish context automatically is maybe a bit too buried. It's quite important I think.

docs/reference/language/context.md
```diff
@@ -25,38 +25,201 @@ in some places that are also aware of their location in the document:
...
+
+```example
+Normal spacing:
+$ x \ x $
+
+#show math.equation.where(block: true): it => {
```
laurmaedje, line 213: The rendered version is fairly long and noisy now because both examples don't fit side-by-side and they show the exact same output. Maybe we can somehow shorten this? Maybe the second version could be dropped in favor of a new section that explains that show rules establish context and then have a different example there?

image

docs/reference/language/context.md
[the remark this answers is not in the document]
laurmaedje: It's true that it's a very common problem with beginners, but I'm not sure that this specific example is the problem. No matter what you put there, people will think that's the way it works. So I think we can keep that. But I agree that we should have more explanation specific to that problem somewhere relatively early in the docs. Doesn't necessarily have to be part of this PR though.

laurmaedje:
I'm not sure whether the PR is waiting on me or you at the moment (Some of the suggestions were addressed by https://github.com/typst/typst/pull/6008/commits/2dcd9884514c93646202cdfb5d00ea7f68b578db while some weren't.)

laurmaedje:
I'll close this PR due to inactivity. 

@ukoethe I'm sorry that the contribution experience was a bit frustrating for you here. Unfortunately, discussing & reviewing docs is even harder than code. I think the PR was on a good way, but as there are still various unaddressed things unfortunately not quite mergeable. If someone wants to adopt this PR, feel free to base on it.

Otherwise, we do plan to invest more into better docs in the near-ish future and this can serve as a source when we get to the context part.

ukoethe:
@laurmaedje I have simply run out of time working on these PRs, especially as your remaining wishes and suggestions were too vague for me to be implemented quickly. 

> Otherwise, we do plan to invest more into better docs in the near-ish future and this can serve as a source when we get to the context part.

I'll be glad if you take my work as inspiration for better docs. Hope it will materialize soon :-)

PR #6016, opened 2025-03-06
Revised Styling documentation

ukoethe:
This PR attempts to fix the places in the 'Styling' documentation where I stumbled and had to use trial-and-error to figure out how things work.

### DIFF
--- a/docs/reference/language/styling.md
+++ b/docs/reference/language/styling.md
@@ -12,18 +12,58 @@ of elements.
 
 ## The computational model behind typst
 
-Typesetting is controlled by hundreds of parameters, from page margins to font sizes to numbering conventions. Managing this mess is a major focus of every typesetting system. The first step is to arrange these properties into related groups. In typst, these groups are called _element functions_ (EFs), for example `text` for font properties (typeface, size, color etc.), `par` for paragraphs (line spacing, alignment, indentation etc.), and `figure` (placement rules, captioning, and numbering of tables and figures). A complete list of the available element functions is **where??**. The term _functions_ indicates that these entities do not merely act as passive property containers, but can actively process content (usually provided in the function's `body` parameter) according to the present parameter settings.
+Typesetting is controlled by hundreds of parameters, from page margins to font 
+sizes to numbering conventions. Managing this mess is a major focus of every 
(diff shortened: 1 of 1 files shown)


laurmaedje (commented):
First of all: Thank you for the work you've put into this!

Unfortunately, I've really struggled to leave a review here. The problem is that there is definitely new, useful information in this PR, but at the same time the characterization of the computational model is also not totally accurate and not always the most helpful from a user perspective I feel like. 

As the PR is quite large, it's really difficult and time-consuming to isolate the improvements and to articulate the things I think are inaccurate. I think more isolated improvements would be much easier to land than a full rewrite like this. I acknowledge that this introduces extra overhead for contributors and understand if it's too annoying for you. It's just a hard problem to collaborate on technical writing.

A few things I liked:

- To have additional details about selectors, and the presence of fields on `it`
- Having an explanation of the common pitfall where a set rule is put in a scope just by itself
- Having examples like "upon creation of the table-of-contents" for `show outline`. Though "creation" is not 100% accurate, maybe "displaying" would be, but it's perhaps less clear.

A few things I didn't like: 
- The text puts a lot of focus on the term "element function", but the characterization of them does not feel quite on point. For example you says that "these groups are called element functions". Elements certainly also group properties, but their main characteristic is that they represent some entity in the document. I feel like characterizing just "elements" once and using that term throughout would already make things much more approachable. Element functions are just functions that create elements (specific instances of content) after all.

- I think some of the phrasing isn't that approachable for newcomers, e.g. "[...] act as passive property containers, but can actively process content [...]". The tone also seems unnecessary negative at times ("Managing this mess", "Show rules are [...] most complicated").

Some things that seem inaccurate:
- "The set of active EFs changes frequently during the processing of your document" I am not quite sure what you mean by this.
- The explanation of the three fundamental methods to modify parameters explains function calls with `text`. I think that's not so helpful because a `text` call is special in that it acts like a text set rule on the passed content. Other constructors don't work like that.

laurmaedje:
As mentioned above, there are some good additions here, but it would still be quite a bit of work to bring this into a mergeable state. For now, I'll close the PR due to inactivity. Thank you still.

PR #6017, opened 2025-03-06
Mention that `sym.ohm` was removed in the 0.13.0 changelog

MDLC01:
See https://github.com/typst/codex/issues/52#issuecomment-2703756070. This should be part of whatever the next version of Typst is, so likely 0.13.1.


laurmaedje:
Thanks!

Issue #5259, opened 2024-10-20
`block`'s `sticky` should have a direction parameter.


MangoIV:
### Description

It would be cool if instead of taking a `bool`, `sticky` had a parameter `above` or `below`.

### Use Case

For now, `sticky` is only used with headlines but I would like to use it to avoid a code block to start a new page. i.e. I want to write something like: 

This is implemented as follows: 
```haskell
-- some long haskell code block
```

and what happens a lot in practise is that the "This is implemented as follows" ends a page and then the code block would be on the next page. Ideally I would like to have the paragraph above be able to stick to the code block.

MangoIV:
Also it would be cool to know if this at all possible for now, e.g. by obtaining all pairings of paragraphs then raw and then making this paragraph a sticky block?

knuesel:
Maybe we could have a `nopagebreak()` element instead, similar to `\nopagebreak` in LaTeX? I think for the user it feels more natural than the sticky property (though I guess it could be implemented with `sticky` under the hood): use `pagebreak()` to break here, use `nopagebreak()` to avoid a break here.

For example the following


```
The formula is: #nopagebreak()
$ x y z $
```

feels more lightweight and natural than

```typ
#block(sticky: true)[The formula is:]
$ x y z $
```

or

```typ
#block(breakable: false)[
  The formula is:
  $ x y z$
]
```

and it would cover both the "up" and "down" sticky use cases.

knuesel:
Here's another use case for "up" stickiness (or `nopagebreak`):

As a user I call `#exercise[body]` from a package to typeset exercises as "**Exercise n**: body" (the title is inline with the body). But in some cases I have

```typ
#exercise[
  + Do this,
  + Do that.
]
```

and the title ends up in its own block before the enumeration, and there's no way I can set `sticky: true` on it. But I could easily do

```typ
#exercise[
  #nopagebreak()
  + Do this,
  + Do that.
]
```

laurmaedje:
### Sticky vs nopagebreak
One benefit of sticky is that it's more declarative if you want to automate it, as it can be applied with a show-set rule and those compose better than transformational show rules. I think extending `sticky` to the other side would be worthwhile regardless of whether the other mechanism is also introduced.

### API design
API-wise, I could imagine the following:
```typ
#set block(sticky: top | bottom | (top: bool, bottom: bool) | bool)
```

- `sticky: top` is short for `sticky: (top: true)`
[...]

PR #6020, opened 2025-03-07
Change the "sticky" attribute to take a direction

efeyakinci:
This change closes https://github.com/typst/typst/issues/5259

It is useful to be able to create blocks that stick to the element above them. The most notable use case is for lists and enums to prevent pagebreaks in documents like
```typ
//...
The following list shows the funding sources that sponsored this article:
// A pagebreak is undesirable here!
- Acme Inc.
- Typst GmbH
- Company C
```

Other use cases include code blocks that follow text, or footnotes such as table footnotes that refer to content above them

Currently, there is no good way to achieve this. This change modifies the sticky parameter to take one of `"below", "above", "both", or none.`

This change also includes a change to the behavior of sticky when interacting with breakable blocks that spill that becomes more visible with such use cases. For an example like the following:

```typ
#set page(height:120pt, margin: (y: 0pt))
#set block(spacing: 0pt, width: 20pt)

#block(height: 60pt, fill: red)[A]
#block(height: 70pt, fill: green, breakable: true, sticky: true)[A]
#block(height: 10pt, fill: purple)[C]
```

The behavior is that the middle block will cause a pagebreak so that it can be with the following block in unbroken form. However, this is likely not the least-surprising behavior. For example, consider a table that covers 90% of the height of a page. If a user added such a table where 20% of the vertical space is occupied and made the table sticky to add footnotes after the table, the expected behavior is likely not that the first page is left 90% blank, and the second page contains the table and footnotes in their entirety.

This PR adds a check to cause a pagebreak on a sticky elements that ends the page. In this case, we break if and only if there is no spill. Examining the cases, we see

1.  There is no spill: Then, we should break, so that the element can stay with the element following it, unless the element is the first element on the page.
1. There is some spill. Clearly (given that one agrees with the general behavior), we not should break if the spillage can be next to the element on the next page. However, this change does make it possible where we avoid breaking, even though breaking would cause spillage onto a third page whose size can accommodate the successor after it.

The second (rare) edge case does allow for a sticky element to be on a different page than its successor, although the behavior with the change is still (I believe) less surprising than the current behavior.

This is non-trivial to fix, as knowing when to break would require calculating the amount of spillage on the final page of spillage, knowing the size of that page, and calculating whether the items will fit. As such, this is not done in this PR.


crates/typst-layout/src/flow/distribute.rs
```diff
@@ -443,7 +472,13 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
...
             // suffix to the next region.
             if let Some(snapshot) = self.sticky.take() {
-                self.restore(snapshot)
+                // Only restore snapshot if there's no spill
+                // If a sticky breakable element can be spilled to the following page,
+                // it's fine to put some of it on this page, since the next page will still
+                // have the next element on the same page as *some* of the spilled content.
+                if self.composer.work.spill.is_none() {
+                    self.restore(snapshot)
+                }
```
PgBiel, line 481: Can you add a test where this changes the behavior of existing `sticky: "below"` blocks so we can see what this is actually changing? Want to avoid regressions here.

crates/typst-library/src/layout/container.rs
```diff
@@ -338,23 +359,29 @@ pub struct BlockElem {
...
     /// #lorem(10)
+    ///
+    /// #table(
+    /// columns: 2,
+    /// [A], [B],
+    /// [C], [D],
+    /// )
+    /// #block(sticky: "above")[The above table shows that...]
     /// ```
-    #[default(false)]
```
PgBiel, line 356: Let's add some compatibility behavior to temporarily accept bools with a warning. (You can leave this for last, after all other reviews have been addressed.)

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -324,10 +325,28 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
...
+                    single.sticky.as_ref().map(|s| s.is_sticky_above()).unwrap_or(false),
+                ),
+                Child::Multi(multi) => Some(
+                    multi.sticky.as_ref().map(|s| s.is_sticky_above()).unwrap_or(false),
+                ),
+                _ => None,
+            })
+            .unwrap_or(false);
+
+        let mut stick_to_successor = || {
```
PgBiel, line 349: I'd love some tests showing how a sequence of sticky: "above" which don't fit in a single page behave. There is one already for sticky: "below", not too sure if it was `block-sticky-many`, but might have been

PgBiel (commented):
Thanks for the proposal. Regarding the feature itself, I can see there being value in being able to take control over "widow prevention", similar to how the existing sticky attribute allows control over "orphan prevention".

However, I have a few concerns; I'll list some below to start, but in general, one important concern is that a single test won't take care of this change with potential for regressions. I feel like we need to clearly study, document and test the behavior of each possible flow child when affected by `sticky: "above"`, and perhaps add more information about edge cases - I mention some below.

laurmaedje:
@efeyakinci Do you still plan to work on this? In principle, it would not be exceedingly hard to implement this feature, but it would need some changes from the current implementation. As it is very central code, it would also need to be well-tested. If you'd prefer not to proceed, that's totally fine, too. Then, I'd close the PR, to "unlock the mutex" on the issue, so to speak.

Regarding API design, I posted some thoughts in the issue: https://github.com/typst/typst/issues/5259#issuecomment-3242195471

Implementation-wise, my gut feeling is that it would be easier to move the pre-processing of this to the collection phase (`collect.rs`) as it should be possible to make this transformation without knowledge of the concrete layout regions. The `LineChild` would then also need a `sticky` bool that could be set by a following Single/Multi child with `sticky: top`. (Take all of this with a graint of salt, as I've not actually experimented with it.)

laurmaedje:
For now, I'll close this due to inactivity. Thanks still!

PR #6021, opened 2025-03-07
Fix `math.root` frame size

### LINKED ISSUE #6019: root function without index omits spacing
### Description

I have been writing my typst code poorly by exploiting the root function for square roots as `$root(,2)$` which is not semantically correct but i think having a good default will be beneficial.

writing $3 root(,2)$ omits spacing between 3 and root of 2

code for all three

```typst
$
"with root function ->"
3 root(,2)\ \ \

"with sqrt function ->"
3 sqrt(2)\ \ \

"with root function but space"\ "char inserted before ->"
3 " " root(,2)\
$
```

Image

compiler version: `8dce676dcd691f75696719e0480cd619829846a9`

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

mkorje:
Closes #6019.

```typ
--- math-root-frame-size-index ---
// Test size of final frame when there is an index.
$ a root(, 3)         & a root(., 3) \
  a sqrt(3)           & a root(2, 3) \
  a root(#h(-1em), 3) & a root(123, 3) $
```
| Old | New |
| - | - |
| old | new |


crates/typst-layout/src/math/root.rs
```diff
@@ -85,22 +85,24 @@ pub fn layout_root(
...
+    let sqrt_pos = Point::new(sqrt_x, radicand_y - gap - thickness);
     let line_pos = Point::new(radicand_x, radicand_y - gap - (thickness / 2.0));
     let radicand_pos = Point::new(radicand_x, radicand_y);
 
     let mut frame = Frame::soft(size);
     frame.set_baseline(ascent);
 
     if let Some(index) = index {
-        let index_pos = Point::new(kern_before, ascent - index.ascent() - shift_up);
+        let index_x = -sqrt_offset.min(Abs::zero()) + kern_before;
```
laurmaedje, line 104: This computation is slightly odd because `sqrt_offset` already incorporates `kern_before` in _its_ computation. Maybe it makes sense, but what is the concrete idea here?

mkorje: `sqrt_offset` is really how far the index's frame (+ `kern_before`) sticks out to the left of the sqrt's frame. As `kern_after` is a negative kern, `sqrt_offset` can thus be negative. This computation is needed in the case it is negative, just like `let sqrt_x = sqrt_offset.max(Abs::zero());` further above. Because `sqrt_offset` already includes `kern_before`, we need to add it back for the index's position.

In summary:
| x coordinate | `sqrt_offset` positive | `sqrt_offset` negative |
| - | - | - |
| index | `kern_before` | `-sqrt_offset + kern_before` |
| sqrt | `sqrt_offset` | 0 |
| radicand | `sqrt_offset + sqrt.width()` | `sqrt.width()` |
| total width | `sqrt_offset + sqrt.width() + radicand.width()` | `sqrt.width() + radicand.width()` |

laurmaedje: I see. Thanks for the explanation!

laurmaedje:
Thank you!

PR #6026, opened 2025-03-07
Use `u64` instead of `usize` to store counter and enumeration item numbers, so behavior does not vary from 64-bit to 32-bit platforms

eduardosm:


laurmaedje:
I think this makes sense, but was wondering whether you investigated `usize` usage in general or just picked these two as prime examples? Because if we change this, I think it makes sense to a comprehensive pass.

eduardosm:
I did a quick look on `usize` usage, but I mostly focused on these two cases.

laurmaedje:
Even if there might be other cases, I agree that this is a clear improvement. Thanks!

It's easy in Rust to use usize where it's not really appropriate because it naturally comes out of so many std APIs...

PR #6033, opened 2025-03-09
Added inline support for `html.frame` element

Kees-van-Beilen:
some custom non html compatible content should sometimes be inlined, this pr adds an optional inline property to the `html.frame` element.

### DIFF
--- a/crates/typst-library/src/html/mod.rs
+++ b/crates/typst-library/src/html/mod.rs
@@ -80,4 +80,13 @@ pub struct FrameElem {
     #[positional]
     #[required]
     pub body: Content,
+
+    #[default(false)]
+    pub inline: bool,
+}
+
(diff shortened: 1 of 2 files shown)


MDLC01:
Is there a difference with simply wrapping the `html.frame` in a `box`?

xkevio:
While the docs can always be improved, I don't think there is a need for specifically mentioning element to element interactions for every permutation since `box` inlines everything anyway. So one would rather want to increase the discoverability of `box` itself and not `box(html.frame)` since that would be redundant.

laurmaedje:
The fact that the span is created is sometimes undesirable, but even if we wanna provide a way to not have that, I think it shouldn't be specific to `html.frame`, so I'd go ahead and close this. Thanks still!

PR #6037, opened 2025-03-10
IDE: complete jump-to-cursor impl

mattfbacon:
There is just the one question, if the transform matrix will ever not be invertible. Currently I unwrap the invert result because I don't understand how the transform could be non-invertible.

### DIFF
--- a/crates/typst-ide/src/jump.rs
+++ b/crates/typst-ide/src/jump.rs
@@ -59,7 +59,13 @@ pub fn jump_from_click(
                         continue;
                     }
                 }
-                let pos = pos.transform_inf(group.transform.invert().unwrap());
+                // Realistic transforms should always be invertible.
+                // An example of one that isn't is a scale of 0, which would
+                // not be clickable anyway.
+                let Some(inv_transform) = group.transform.invert() else {
+                    continue;
+                };
+                let pos = pos.transform_inf(inv_transform);
                 if let Some(span) = jump_from_click(world, document, &group.frame, pos) {
                     return Some(span);
                 }
@@ -292,7 +298,8 @@ mod tests {
             cursor(45),
         );
         test_click(
-            "#box(width: 10pt, height: 10pt, clip: true, scale(x: 300%, y: 300%, origin: top + left, rect(width: 10pt, height: 10pt)))",
+            "#box(width: 10pt, height: 10pt, clip: true, scale(x: 300%, y: 300%, \
+             origin: top + left, rect(width: 10pt, height: 10pt)))",
             point(20.0, 20.0) + margin,
             None,
         );
@@ -305,6 +312,11 @@ mod tests {
             "#box(width: 10pt, height: 10pt, clip: true, rect(width: 30pt, height: 30pt))",
             point(20.0, 20.0) + margin,
             None,
+        );
+        test_click(
+            "#rotate(90deg, origin: bottom + left)[hello world]",
+            point(5.0, 15.0) + margin,
+            cursor(40),
         );
     }
 
--- a/crates/typst-ide/src/jump.rs
+++ b/crates/typst-ide/src/jump.rs
@@ -3,7 +3,7 @@ use std::num::NonZeroUsize;
 use typst::layout::{Frame, FrameItem, PagedDocument, Point, Position, Size};
 use typst::model::{Destination, Url};
 use typst::syntax::{FileId, LinkedNode, Side, Source, Span, SyntaxKind};
-use typst::visualize::Geometry;
+use typst::visualize::{Curve, CurveItem, FillRule, Geometry};
 use typst::WorldExt;
 
 use crate::IdeWorld;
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/visualize/curve.rs
```diff
@@ -530,3 +530,30 @@ impl Curve {
...
+                point_to_kurbo(point1),
+                point_to_kurbo(point2),
+            ),
+            CurveItem::Close => PathEl::ClosePath,
+        });
+        BezPath::from_vec(path.collect())
+    }
+
+    /// When this curve is interpreted as a clip mask, would it contain `point`?
+    pub fn contains(&self, needle: Point) -> bool {
```
laurmaedje, line 556: Since we have this now, it might be a nice and simple addition (for a separate PR) to also handle paths in `jump_from_click`'s `FrameItem::Shape` case.

crates/typst-library/src/visualize/curve.rs
```diff
@@ -530,3 +530,30 @@ impl Curve {
         Size::new(max_x - min_x, max_y - min_y)
     }
 }
+
+fn point_to_kurbo(point: Point) -> kurbo::Point {
```
laurmaedje, line 534: could you order this in opposite order? so basically
```rust
impl Curve {
  fn contains ..
  fn to_kurbo ..
}

fn point_to_kurbo .. 
```
that would be more in line with the rest of the code base

crates/typst-ide/src/jump.rs
```diff
@@ -269,6 +272,42 @@ mod tests {
...
+            point(5.0, 30.0) + margin,
+            cursor(38),
+        );
+        test_click(
+            "#scale(x: 300%, y: 300%, origin: top + left, rect(width: 10pt, height: 10pt))",
+            point(20.0, 20.0) + margin,
+            cursor(45),
+        );
+        test_click(
+            "#box(width: 10pt, height: 10pt, clip: true, scale(x: 300%, y: 300%, origin: top + left, rect(width: 10pt, height: 10pt)))",
```
laurmaedje, line 295:
```suggestion
            "#box(width: 10pt, height: 10pt, clip: true, scale(x: 300%, y: 300%, \
             origin: top + left, rect(width: 10pt, height: 10pt)))",
```

crates/typst-ide/src/jump.rs
```diff
@@ -53,10 +53,14 @@ pub fn jump_from_click(
...
-                if let Some(span) =
-                    jump_from_click(world, document, &group.frame, click - pos)
-                {
+                let pos = click - pos;
+                if let Some(clip) = &group.clip {
+                    if !clip.contains(pos) {
+                        continue;
+                    }
+                }
+                let pos = pos.transform_inf(group.transform.invert().unwrap());
```
laurmaedje, line 62: I think there's no guarantee that this is invertible since any group could be created. Better to be safe and just skip the group if the transform is not invertible instead of panicking.

laurmaedje (commented):
Thanks, very nice! I tried it with the web app and it works great, also with text! Maybe having one `jump_from_click` test for text as well would be good.

crates/typst-ide/src/jump.rs
mattfbacon: I agree. But can you describe in natural language what it would mean if a transform matrix is not invertible? Like what kind of transforms would lead to that?

crates/typst-library/src/visualize/curve.rs
mattfbacon: OK or I just rename this one to "improve IDE position-span mapping" and do it here too

mattfbacon:
BTW, if we do shapes, it would be good to handle stroke too; if it is particularly big that would lead to inaccuracies.

mattfbacon:
Also should clicking inside unfilled shapes go to them? I could imagine using a shape placed on top as some kind of border or highlighting where the content under it is the actual content at that position.

crates/typst-library/src/visualize/curve.rs
laurmaedje: works too

laurmaedje:
> BTW, if we do shapes, it would be good to handle stroke too; if it is particularly big that would lead to inaccuracies.

Sounds good.

> Also should clicking inside unfilled shapes go to them? I could imagine using a shape placed on top as some kind of border or highlighting where the content under it is the actual content at that position.

We iterate in reverse to make sure we catch stuff in front first, so I think it's okay to allow clicking on them. Allowing click to jump only when clicking exactly on the stroke would be even nicer, but also more complex. But if you want to go ahead and implement that, I'd be on board with it.

mattfbacon:
One question, I check if fill is set, but should we also treat a solid fill with 0 alpha as unset? I'm not sure because it opens a can of worms in terms of what to do with gradients (e.g., where all the stops are transparent) and patterns.

laurmaedje:
Let's not open that can.

mattfbacon:
OK, then this is complete.

mattfbacon:
Thanks, I try :)

For future reference, https://github.com/linebender/kurbo/issues/431 presents a possibility to optimise the `contains` and `stroke_contains` methods by avoiding the collection.

goyalyashpal:
so, this would be like `synctex` in latex land right?

laurmaedje:
Yes, it's an improvement to the Preview -> Source jump.

PR #6038, opened 2025-03-11
Fix `cargo clippy` warnings (mostly about `.repeat.take` and `.next_back`)

mi2ebi:


PgBiel (approved):
It is unfortunate that Rust doesn't appear to be able to automatically handle the pitfalls of `last()` and `repeat()` with `take()` (sources at https://qsantos.fr/2025/01/01/rust-gotcha-last-on-doubleendediterator/ and https://doc.rust-lang.org/stable/std/iter/fn.repeat_n.html respectively).

laurmaedje:
Are these from nightly clippy? Since we have clippy in CI in principle.

laurmaedje:
Thanks!

mi2ebi:
> Are these from nightly clippy? Since we have clippy in CI in principle.

IIRC I've set nightly as my default, so probably.

PR #6040, opened 2025-03-11
Update 1-writing.md to improve readability

Wolf-SO:
In my view, the (superset) relationship between *content* and *string* should be made comprehensible here by means of examples. The suggested wording *in contrast* introduces the second example, while the current wording *on the contrary* seems to introduce the contradiction to a previous statement.


PgBiel (approved):
Looks OK to me

laurmaedje:
Thanks!

PR #6063, opened 2025-03-15
Added Russian numbering

NoAIHelper:
There is of course a way of Russian numbering. For example, [this](https://github.com/typst/typst/issues/1595#issuecomment-1609971882). However, I think it's better to add native support.

In this case, it looks like you will have to make changes to custom packages. For example, in [subpar](https://github.com/tingerrr/subpar). However, this means that you only need to add one line of code [(this)](https://github.com/tingerrr/subpar/blob/c61eafeba2bcc7eedd65ff0870699250e70dcad1/src/util.typ#L144).

That is, replace
```typst
   let symbols = ("1", "a", "A", "i", "I", "い", "イ", "א", "가", "ㄱ", "\\*")
```
to
```typst
   let symbols = ("1", "a", "A", "а", "А", "i", "I", "い", "イ", "א", "가", "ㄱ", "\\*")
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,34 +104,22 @@ Let's dissect what's going on:
 </details>
 
 ## Installation
+It is installation of my unofficial package.
+
 Typst's CLI is available from different sources:
 
 - You can get sources and pre-built binaries for the latest release of Typst
-  from the [releases page][releases]. Download the archive for your platform and
-  place it in a directory that is in your `PATH`. To stay up to date with future
-  releases, you can simply run `typst update`.
+  from the [releases](https://github.com/NoAIHelper/typst/releases). Download the archive for your platform and
+  place it in a directory that is in your `PATH`.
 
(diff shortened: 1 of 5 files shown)


PgBiel:
This was already attempted before, but unfortunately there are some concerns: https://github.com/typst/typst/pull/5597

NoAIHelper:
Can we use [lang](https://typst.app/docs/reference/text/text/#parameters-lang) for this? For example, you can probably do "if else if else..." to select a dictionary depending on [lang](https://typst.app/docs/reference/text/text/#parameters-lang).

PgBiel:
While it is something that could be explored in the future, the first and foremost priority is to use a more verbose format to be able to unambiguously specify your numbering in order to remain compatible with the current non-contextual APIs, as explained here: https://github.com/typst/typst/pull/5597#issuecomment-2551816451

Issue #5978, opened 2025-02-28
Include files outside of the project root via CLI


jmpunkt:
In https://github.com/jmpunkt/ox-typst/ we are using Typst to generate PDF files from Org files (by generating Typst first). We are facing a limitation with the project root limitations. In Org there are many ways to specify files. These file can also be outside of the project root, and they most of the time do that. In our case the project root is always the same directory where the Org file is located. There are ways to workaround these issue, but they seem unintended.

## Feature

I was looking for something like `--input` but for files. So the user would specify an input file in CLI and Typst would read that file and generate a variable which contains the content of the file. Now that Typst 0.13 can work with either a path or the file content directly (which is a beautiful addition), these two should be complement each other perfectly.

### How does it work?

So lets say I have a file `~/test.txt`. I want the file to be part of my Typst file `~/workspace/typst-project/test.typ`. I would expect the Typst file to look something like this

```typst
#raw(str(sys.files.test))
```

Now this file would be compiled like this.

```sh
cd ~/workspace/typst-project/
typst c --input-file test=~/test.txt test.typ
```

As a result, we would be able to include files outside of the project root. Since this a manual process, I do not see any security issues for now.

## Rational

You may ask why we need this feature, when we can use the path/bytes feature introduced in Typst 0.13.

For everything text based this would work pretty well. When you start including Images, things do not look promising. It is not feasible to generate a Typst files which embeds an Image in the megabyte size. For me, including these files via the CLI seems like a reasonable and elegant solution without breaking the project root.

T0mstone:
I like your idea (tho it should probably be called `sys.input-files` for consistency), but I just want to point out that using symlinks is easy and arguably more natural.
OTOH, configurable input files might also just be benefitial in general, with the same rationale as the regular text inputs.

laurmaedje:
For text files, you can also do `typst c hi.typ --input "key=$(cat file.txt)"`. For binary data, this would require some sort of encoding, e.g. base64. Still possible though, in principle.

T0mstone:
> For text files, you can also do `typst c hi.typ --input "key=$(cat file.txt)"`.
I think `typst c` isn't the thing to think about here, but rather `typst w`. Because it'd have to update when `file.txt` is changed, and I'm not sure if the "`cat` into `--input`" approach would to that.

jmpunkt:
[...]

Issue #5382, opened 2024-11-05
Allow the import of files outside the project folder (in addition to --root)


woerndl:
### Description

Assume you have a file structure that shares a template or other common parts:

```
project/
  - common/
  - thesis-1/
  - thesis-2/
```

You can run `typst compile thesis-1/main.typ --root ~/project/` or symlink parts (= PITA). But you can't import common files in a way that gets rid of the error message (in this case Tinymist). Because the live syntax check (be it Tinymist or any other future solution) can't know what the scope of the project is, as it is not a permanent setting.

Searching in Discord I found a couple of people with the same issue, be it for books, a thesis (split chapters) etc.

### Use Case

Having `--root` as the only option to include files outside the current directory is, to [quote another user](https://discord.com/channels/1054443721975922748/1088371919725793360/1298102697966178354), a "too basic of a thing not to have". I would agree with that assessment and propose considering a way for permanent relative imports outside the project root.

trombonehero:
Yes please! For out-of-tree builds that integrate with other tools (e.g., something builds an SVG file that is then incorporated into the Typst->PDF build, all within a `build` directory), a `-I` option would be invaluable.

terefang:
for some usecases it might even be more beneficial to have a search path similar to https://github.com/typst/typst/pull/6190

laurmaedje:
I thought a bit more about this while looking at the [`--input-file` PR ](https://github.com/typst/typst/pull/6073) and had the following idea: We could have something like `--input-path mykey=path/to/file` which will add  `sys.inputs.mykey` of type `path` (which is to be added soon; this would block the design). Then you could use this path like any other. In particular, you could read from it, import/include it, or join strings to it to perform operations relative to a directory that was provided like this. In the directory case, the path would form its own detached root, and the imported Typst file could perform reads relatively & absolutely to that root, but not escape it.

Some examples of what this would enable:
- `typst c file.typ --input-path config=config.toml` -> `toml(sys.inputs.config)`
- `typst c file.typ --input-path img=tiger.png` -> `image(sys.inputs.img)`
- `typst c file.typ --input-path assets=~/assets` -> `image(sys.inputs.assets + "path/in/asset/dir.svg")`
- `typst c file.typ --input-path graph=~/graph.typ` -> `include sys.inputs.graph`
- `wget https://data.org/data.json | typst c file.typ --input-path data=-` (stdin) -> `json(sys.inputs.data)`

terefang:
wow!

[...]

PR #6073, opened 2025-03-18
Add `input-file` CLI argument and `sys.input-files` variable

jmpunkt:
### Details
- #5978

### Implementation
I followed the existing `inputs` implementation. However, there are no tests, so I put a manual test below.

One thing I noticed, is the missing file path resolution in the Fish shell (works for Bash). For example, when writing 

```sh
typst c --input-file el=~/test.el test.typ
```
and the file exists, then we can not resolve that path. `std::fs::canonicalize` does not resolve `~` for us. The command fails with the error code: `~/test.el ` does not exist. As a result, Fish users will get this rather unexpected error. Might also be a bug in the Fish shell so I am not sure.

> test.el
> ```elisp
> (message "hi")
> ```

```sh
typst c --input-file el=./test.el test.typ
```

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -290,6 +290,15 @@ pub struct WorldArgs {
     )]
     pub inputs: Vec<(String, String)>,
 
+    /// Add a string key-path pairs where the content will visible through `sys.input-files`.
+    #[clap(
+        long = "input-file",
+        value_name = "key=value",
+        action = ArgAction::Append,
+        value_parser = ValueParser::new(parse_sys_input_file_pair),
+    )]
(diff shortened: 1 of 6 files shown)


laurmaedje:
We've discussed this among the team and came to the conclusion that we can imagine something like this in principle, but would like to see a more general solution. 

One idea would be something along the lines of an [`--input-path` mechanism](https://github.com/typst/typst/issues/5382#issuecomment-3150286257) that adds instances of a to-be-added `path` type to `sys.inputs` and then one can read from those paths. This would allow using this mechanism not just for bytes, but also for importing/including and to read from whole directories. For now, this is blocked on the addition of a `path` type though. 

Thank you still for the PR!

PR #6081, opened 2025-03-22
Expand page breaks' triggers for page(height: auto) in docs

Andrew15-5:
Page breaks can only be triggered with `pagebreak()` or `set page(a: b)`, but not with `set page()`.


laurmaedje:
Agreed that it's worth mentioning that `set page(..)` can also trigger a pagebreak.

PR #6084, opened 2025-03-22
Warn on last show/set rule in block

### LINKED ISSUE #6082: Generate a warning when a show or set rule is the last thing in a scope
### Description

When a show or set rule appears last in a block, it doesn't do anything. for example
```typ
#{
  set table(stroke: none)
}
```
has no effect since it only affects tables in its scope, and there is nothing more in the scope.

In such a case, a warning could be generated, probably linking to [Making a Template -- Embedding set and show rules](https://typst.app/docs/tutorial/making-a-template/#set-and-show-rules).

### Use Case

A new user who has the following code
```typ
#set table(stroke: none)
#table(none)
```
may want to extract a function like this:
```typ
#let setup-tables() = {
  set table(stroke: none)
}

#setup-tables()

#table(none)
```
however, this will result in the style not being applied. This is a very common problem on Discord and the forum, so a helpful warning could save a lot of users frustration.

T0mstone:
Closes #6082

The warning itself may be a bit low quality, so some feedback on that would be greatly appreciated.

I wasn't sure where to put this test, so I made a new file for it. Any other suggestions are welcome too.

Also, this causes two tests to fail that didn't before, probably as part of the test suite putting stuff into a block; Not sure what to do about that.

Also Also, maybe this is too simple of an approach and it would be better to check if there's any content after the last show/set rule. But that could also be left to a future PR.

### DIFF
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -285,6 +285,19 @@ impl Eval for ast::CodeBlock<'_> {
 
     fn eval(self, vm: &mut Vm) -> SourceResult<Self::Output> {
         vm.scopes.enter();
+        if let Some((span, kind)) =
+            self.body().exprs().next_back().and_then(|x| match x {
+                ast::Expr::ShowRule(_) => Some((x.span(), "show")),
+                ast::Expr::SetRule(_) => Some((x.span(), "set")),
+                _ => None,
+            })
+        {
+            vm.engine.sink.warn(typst_library::diag::warning!(
+                span,
+                "{kind} rule has no effect";
+                hint: "See https://typst.app/docs/tutorial/making-a-template/#set-and-show-rules"
+            ));
+        }
         let output = self.body().eval(vm)?;
         vm.scopes.exit();
         Ok(output)
--- a/tests/suite/styling/in-block.typ
+++ b/tests/suite/styling/in-block.typ
@@ -1,0 +1,12 @@
+--- warn-show-set-last-in-block ---
+#{
+// Warning: 1-14 show rule has no effect
+// Hint: 1-14 See https://typst.app/docs/tutorial/making-a-template/#set-and-show-rules
+show "a": "b"
+}
+
+#{
+// Warning: 1-15 set rule has no effect
+// Hint: 1-15 See https://typst.app/docs/tutorial/making-a-template/#set-and-show-rules
+set text(blue)
+}


laurmaedje:
@Enivex [brought up on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1353748384720093366) that this warning might be quite noisy while still typing. I'm personally not sure how big of a problem that is.

laurmaedje:
I would be on board with giving this warning a shot. It's not all that different from compiler errors that temporarily appear and it's easy enough to remove again in the future if it turns out too noisy.

T0mstone:
I'm not sure if 72c219d is the best fix for the new test failures, I just went for the most simple and easy fix of adding some useless content after so the warning doesn't fire. An alternative would be annotating the tests to expect the warning, but that makes the tests less focused.

Edit: I changed it to use an empty content block instead of `lorem(5)`, that way they don't have to have reference images.

tests/suite/styling/in-block.typ
```diff
@@ -0,0 +1,12 @@
+--- warn-show-set-last-in-block ---
+#{
+// Warning: 1-14 show rule has no effect
+// Hint: 1-14 See https://typst.app/docs/tutorial/making-a-template/#set-and-show-rules
```
laurmaedje: I think it would be nicer to have an inline explanation, e.g.

`a show rule is only in effect until the end of the surrounding code block`

could you also indent the content of the code blocks by 2 spaces?

crates/typst-eval/src/code.rs
```diff
@@ -285,6 +285,19 @@ impl Eval for ast::CodeBlock<'_> {
...
+                ast::Expr::SetRule(_) => Some((x.span(), "set")),
+                _ => None,
+            })
+        {
+            vm.engine.sink.warn(typst_library::diag::warning!(
+                span,
+                "{kind} rule has no effect";
+                hint: "See https://typst.app/docs/tutorial/making-a-template/#set-and-show-rules"
+            ));
+        }
```
laurmaedje: I would suggest extracting a function 
```rust
/// Emits a warning when a set or show rule has no effect.
fn warn_for_useless_rule(engine: &mut Engine, expr: ast::Expr, kind: &str) {
    engine.sink.warn(warning!(
        expr.span(),
        "{kind} rule has no effect";
        hint: "a {kind} rule is only in effect until the end of the surrounding code block"
    ));
}
```

And using it above:
```rust
impl Eval for ast::CodeBlock<'_> {
    type Output = Value;

    fn eval(self, vm: &mut Vm) -> SourceResult<Self::Output> {
        let body = self.body();
        match body.exprs().next_back() {
            Some(expr @ ast::Expr::ShowRule(_)) => {
                warn_for_useless_rule(&mut vm.engine, expr, "show")
            }
            Some(expr @ ast::Expr::SetRule(_)) => {
                warn_for_useless_rule(&mut vm.engine, expr, "set")
            }
            _ => {}
        }

        vm.scopes.enter();
        let output = body.eval(vm)?;
        vm.scopes.exit();
        Ok(output)
    }
}
```

I would also suggest reusing the `.body()` results, since this is a on a hot path. Maybe we could even integrate this into `eval_code` instead to pay less overhead. Hard to say whether this actually has an impact or not.

T0mstone:
Thanks for doing the work for me😅

I went with a `Span` as the argument for that helper function, since the other parts of the expr aren't used. Maybe this gives slightly better codegen.

PgBiel:
I've recently had an idea and found that `{ set text(red) }` is actually pretty convenient to be able to "transport" rules... Maybe we should only warn if this is actually placed somewhere (e.g. if we place a `styled()` with empty child, and it was defined like such in the AST)? :p

See: https://github.com/PgBiel/elembic/issues/57

ensko:
> Maybe we should only warn if this is actually placed somewhere (e.g. if we place a styled() with empty child, and it was defined like such in the AST)?

If that can be implemented easily, that would of course be perfect. That's also more similar to e.g. HTML warnings: `#let x = html.elem("a")` doesn't warn during paged export, so that could be an inspiration for the implementation.

The only downside I can imagine is when creating an element to maybe put it in the document, but under testing conditions it wasn't. Since I see the warning as a teaching tool for users that are not familiar with rule scoping yet, not a bug catching tool, that doesn't worry me much. If it can be made not to warn wrongly on a legitimate advanced pattern, even better.

laurmaedje:
One annoying thing about doing it dynamically is that we'd need to add the extra information of whether it was syntactically empty to `StyledElem`, which feels a bit unclean and would also increase the size of the struct by a `usize` because of alignment. 

Or I guess we could replace the empty sequence child by one that warns upon realization, but that would be observable in the element structure.

bluss:
Another curiosity being added to the log here.

The following is valid in typst and has effect (the document title is set). Typst version 0.13.1

```typ
#context { set document(title: "X") }
```

This could then be an exception to the warning

laurmaedje:
Given the multiple examples of false positives that have been brought up, I'd rather not add this at this time. Thanks for your work on this nonetheless!

PR #6092, opened 2025-03-24
Add env setting for ignore_system_fonts

### LINKED ISSUE #6087: TYPST_IGNORE_SYSTEM_FONTS environment variable
### Description

I would like a `TYPST_IGNORE_SYSTEM_FONTS` environment variable that makes it possible to control whether or not Typst will ignore the system fonts.

### Use Case

I use Typst on NixOS in a [Devenv](https://devenv.sh/) and I am contributing an option to set font paths as follows to make builds more reproducible.

```
languages.typst = {
  enable = true;
  fontPaths = "${pkgs.liberation_ttf}/share/fonts/truetype";
}
```

I would like to add a `languages.typst.ignoreSystemFonts` options to ensure complete reproducibility in the Devenv. However Devenv only controls your environment and can't set flags. It would be nice if it could use a `TYPST_IGNORE_SYSTEM_FONTS` environment variable to control it.

bluebear94:
Fixes #6087.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -361,7 +361,7 @@ pub struct FontArgs {
 
     /// Ensures system fonts won't be searched, unless explicitly included via
     /// `--font-path`.
-    #[arg(long)]
+    #[arg(long, env = "TYPST_IGNORE_SYSTEM_FONTS")]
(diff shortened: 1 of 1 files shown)


laurmaedje:
I think it makes sense. Thanks!

PR #6096, opened 2025-03-25
Fix descriptions of color maps

frozolotl:
Should be fairly self-explanatory.

Honestly, I'm not 100% happy with some of these changes though:

- Mako: it's not _really_ going to white. It's more a light jade, but that name isn't good for people who don't know what jade looks like, or for people who associate "jade" with the white kind of jade (which AFAIK is most Chinese).

- Icefire: similar issue with "orange" and "peach"


Andrew15-5:
https://typst.app/docs/reference/visualize/color/#predefined-color-maps

I agree. The `magma` doesn't really end with yellow, at best it's light yellow, very close to white. `mako` also ends with a very light green/blue, very close to white. `icefire` maybe starts with a light teal, but it ends with more of a light orange. `crest` is like light green to dark blue/purple.

laurmaedje:
Thanks! 

>  Honestly, I'm not 100% happy with some of these changes though

I mean, it's not that important. If we have it, it shouldn't be completely wrong, but if it isn't, I think it's fine.

PR #6100, opened 2025-03-26
Show warnings from eval

bluebear94:


laurmaedje:
I think it would be more idiomatic for `fn eval_string` to take a `sink: TrackedMut<Sink>` instead of returning the `Sink`. That would align it with `fn eval`.

laurmaedje:
Thanks!

PR #6101, opened 2025-03-27
Correcting Czech translation in `typst-library`

Astra3:
The term "Seznam" is used for lists (of something) in Czech, not for listings (from something). Across academia I saw the term "Výpis" used the most for source codes and therefore it seems to be the most correct one to use.


laurmaedje:
Thanks for the fix!

PR #6103, opened 2025-03-27
Add inset parameter for figures

### LINKED ISSUE #6095: Figure placement is ignored when using `pad` in show rule
### Description

Hi,

I wanted to add some spacing around a figure in typst, and the following behavior puzzled me: When using `pad` in the show rule, the `placement` option is ignored:

```typ
#set par(justify: true)

#show figure: f => [
    #pad(top: 1em, bottom: 1em, f)
]

#lorem(125)

#figure(
  [``` lalilu ```],
  placement: top,
  caption: "some caption"
)

#lorem(125)
```

However, this is not the case with other show rules for the figure:

```typ
#set par(justify: true)

#show figure: b => {[#v(0.75em) #b #v(0.75em) ]}

#lorem(125)

#figure(
  [``` lalilu ```],
  placement: top,
  caption: "some caption"
)

#lorem(125)
```

Is this intended behavior?

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

eltos:
Exposes block's inset parameter to easily control the spacing around figures.

<table>
<tr>
<td>

```typ
Text above the figure



#figure(
  rect(),
  caption: [A rectangle],
)

Text below the figure
```

</td>

<td>

```typ
Text above the figure

#set figure(inset: (top: 1em, bottom: 2em))

#figure(
  rect(),
  caption: [A rectangle],
)

Text below the figure
```

</td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

**Why do we want this?**
- For inline figures one can manually add `#v` or wrap the figure in a `block` or `pad`, but this does not work for floating figures, see #6095.  
- The simple show rule `#show figure: set block(spacing: 1em)` suggested [here](https://forum.typst.app/t/how-to-reference-a-figure-with-modified-padding/1438) works for images, but cannot be used for tables as it messes with the cell spacing.

Related: https://stackoverflow.com/questions/78622060/add-spacing-around-figure-in-typst  
Closes #6095


MDLC01:
> * The show rule `#show figure: set block(above: 1em, below: 2em)` only works for inline figures, not floating figures (as intended)

Is this really intended. I did not reflect on that a lot, but I feel like the right fix would be to make this rule work.

PR #6108, opened 2025-03-27
Add Indonesian translation

markus-langgeng:


crates/typst-library/src/text/lang.rs
```diff
@@ -14,7 +14,7 @@ macro_rules! translation {
     };
 }
 
-const TRANSLATIONS: [(&str, &str); 38] = [
+const TRANSLATIONS: [(&str, &str); 39] = [
```
MDLC01, line 17:
```suggestion
const TRANSLATIONS: &[(&str, &str)] = [
```

If the MSRV allows it, this could be done to prevent having to change this number every time a new translation is added.

markus-langgeng: sorry, may I ask what should I do next? this is my first time submitting translation. I'm assuming I just need to click the commit suggestion button, but I don't know what the Minimum Supported Rust Version is about and what should I do about it.

MDLC01: You can commit the suggestion, and if GitHub tells you the tests now fails, just revert it back to what it was by making a new commit.

But this is not necessary, just a small improvement that can be done later, so don't worry if you are scared to break something, there is no express need to do that.

markus-langgeng: okay, I'll try commit the suggestion.

markus-langgeng:
I've reverted the change as it didn't pass the tests.

crates/typst-library/translations/id.txt
```diff
@@ -0,0 +1,8 @@
+figure = Gambar
+table = Tabel
+equation = Persamaan
+bibliography = Daftar Pustaka
+heading = Bagian
+outline = Daftar Isi
+raw = Kode
+page = hlm.
```
laurmaedje, line 8: It's already a bit inconsistent across the translations, but in the canonical English version the translation for page is spelled out "page" (with a lowercase first letter) rather than being a shortened as "p." We may need both forms down the road, but unless the full form literally doesn't make sense in Indonesian, I think we should aim for consistency here.

markus-langgeng: ah I see, I'm going to change it to "halaman" then

crates/typst-library/translations/id.txt
```diff
@@ -0,0 +1,8 @@
+figure = Gambar
+table = Tabel
+equation = Persamaan
+bibliography = Daftar Pustaka
+heading = Bagian
+outline = Daftar Isi
+raw = Kode
+page = halman
```
laurmaedje, line 8: should it be "halaman" rather than "halman"?

markus-langgeng: oh sorry, you're totally right.

markus-langgeng: Fixed the typo, thanks for pointing it out

laurmaedje:
Thanks!

PR #6110, opened 2025-03-28
Added callsite to func calls in tracing

Dherse:
Adds callsite information in tracing of function calls.

Not sure how useful this is to others, but I needed it for optimizing codly, figured it could be useful more generally. It's a bit bigger than I would have liked :/

image

### DIFF
--- a/crates/typst-library/src/foundations/func.rs
+++ b/crates/typst-library/src/foundations/func.rs
@@ -298,7 +298,7 @@ impl Func {
     }
 
     /// Non-generic implementation of `call`.
-    #[typst_macros::time(name = "func call", span = self.span())]
+    #[typst_macros::time(name = "func call", span = self.span(), callsite = args.span)]
     fn call_impl(
         &self,
         engine: &mut Engine,
--- a/crates/typst-macros/src/time.rs
+++ b/crates/typst-macros/src/time.rs
@@ -8,12 +8,13 @@ use crate::util::{kw, parse_key_value, parse_string};
 /// Expand the `#[time(..)]` macro.
 pub fn time(stream: TokenStream, item: syn::ItemFn) -> Result<TokenStream> {
     let meta: Meta = syn::parse2(stream)?;
-    Ok(create(meta, item))
+    create(meta, item)
 }
 
 /// The `..` in `#[time(..)]`.
 pub struct Meta {
     pub span: Option<syn::Expr>,
+    pub callsite: Option<syn::Expr>,
     pub name: Option<String>,
 }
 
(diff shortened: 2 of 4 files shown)


laurmaedje:
Isn't the name a bit redundant with the span, since the name can be found in the source? I guess it means a bit less jumping back and forth, but since it's heaver, it could mess with timing overhead more.

I also wonder it would make sense to swap the spans. Essentially changing
```
span = self.span(),
callsite = args.span,
```
to
```
span = args.span,
callee = self.span(),
```

laurmaedje:
Just a friendly bump :)

Dherse:
@laurmaedje Sorry, as you know life has been in the way lately. That being said, I enjoyed dipping my toes back in Typst by finishing this! It uses `serde_json::Value` as an underlying repr to allow arbitrary `impl serde::Serialize` values to be passed in, otherwise it allows the value to be passed in four modes:
- serialize (default), just add a `<key> = <value>` in the macro arg list
- span, just add a `<key> = $<value>` in the macro arg list
- display (requires `impl std::fmt::Display`), just add a `<key> = #<value>` in the macro arg list
- debug (requires `impl std::fmt::Debug`), just add a `<key> = ?<value>` in the macro arg list

Currently nothing uses those last two types of course, but they're open for future use. I did keep some things as special cases even if they don't need to (namely callsite and closure names) but those can be easily removed. As far as I can tell, this has very little performance impact compared to the other techniques :) 

image

laurmaedje:
I'm quite surprised that `IndexMap` doesn't have noticable performance impact. It's two allocations after all. Are you certain that this doesn't mess with the results? I also thing some spans got lost in a [recent show rule refactor](https://github.com/typst/typst/pull/6569) of mine, which could skew the results, I'll look into that.

Dherse:
@laurmaedje I couldn't really notice it in a quick `hyperfine` on masterproef. What I'll do tho after work is change it to use a `Vec<(String, JsonValue)>` (example) instead! Anyway, that way I can use an `EcoVec` for cheaper cloning. Because I don't think the deduplication is quite necessary, what I'll do is simply add a check in the macro for duplication that should be enough :)

Dherse:
Ok, spent 30 minutes taking a break and fixing it ;)
I can't tell that it's faster but it might be on some platforms.

laurmaedje:
I did some testing with the new benchmark infrastructure and here are the results:
```
main with JSON export

   108,209,578,246 instructions:u                                                        
      12.204656211 seconds time elapsed
      10.522365000 seconds user
       1.682058000 seconds sys

#6110 with JSON export

   127,178,213,743 instructions:u                                                        
      14.166144177 seconds time elapsed
      12.074295000 seconds user
       1.974557000 seconds sys

main without JSON export

    81,397,226,998 instructions:u                                                        
      10.063309923 seconds time elapsed
       8.502912000 seconds user
       1.559984000 seconds sys

#6110 without JSON export

    81,788,637,229 instructions:u                                                        
      10.290362814 seconds time elapsed
       8.735038000 seconds user
       1.553829000 seconds sys

main without --timings

    81,055,062,358 instructions:u                                                        
       9.909727638 seconds time elapsed
       8.382104000 seconds user
       1.527019000 seconds sys

#6110 without --timings

    81,197,849,944 instructions:u                                                        
      10.028100133 seconds time elapsed
       8.572161000 seconds user
       1.454688000 seconds sys
```
(The seconds values are a bit noisy, so instructions is more instructive!)

Dherse:
I am surprised that it has an impact on "normal" mode (no export) mode 😱 
I'll investigate that after work!

laurmaedje:
> I am surprised that it has an impact on "normal" mode (no export) mode 😱
I'll investigate that after work!

No export means "yes timings", but "no JSON export". That wasn't particularly clear from my message. I made this distinction because the JSON export dominantes everything. All of the measurements are with `--timings`.

laurmaedje:
For completeness, it also added the numbers without any --timings`.

Dherse:
Ah okay, that kind of reassures me haha. Because I purposefully modified it to use a closure to try and avoid adding extra instructions.

laurmaedje:
Though it looks like it actually still adds a very small amount of new overhead with `--timings` disabled.

laurmaedje:
I think the overhead for generating the file is a bit large here and there seems to be at least a little impact on compilation without `--timings` enabled. I'm not sure it's worth it overall.

laurmaedje:
I'll close this for now. Thank you still!

PR #6119, opened 2025-03-30
Precisely find label definition

Myriad-Dreamin:
https://github.com/Myriad-Dreamin/tinymist/issues/550 Proved that it is not quite precise to estimate the source location of a label by the span of the labelled element. This PR records and uses the spans where the labels are assigned.

See test cases to be more specific. Before the PR, the finder will give following results:

```typ
#figure[] <hi> See @hi
 ^^^^^^^^ <------- ^^^
#let test1(body) = figure(body); #test1([Test1]) <fig:test1> @fig:test1
                   ^^^^^^^^^^^^^ <-------------------------- ^^^^^^^^^^

#let test1(body) = figure(body);
                   ^^^^^^^^^^^^^ <-------------------------------------- 
#let test2(body) = test1(body); #test2([Test2]) <fig:test2>; @fig:test2|
                                                             ^^^^^^^^^^|
```

Now, we will tell the precise source location of the labels

```typ
#figure[] <hi> See @hi
          ^^^^ <-- ^^^
#let test1(body) = figure(body); #test1([Test1]) <fig:test1>    @fig:test1
                                                 ^^^^^^^^^^^ <- ^^^^^^^^^^

#let test1(body) = figure(body);
#let test2(body) = test1(body); #test2([Test2]) <fig:test2>;   @fig:test2
                                                ^^^^^^^^^^^ <- ^^^^^^^^^^
```

I also added the above examples as tests to prevent regression.

### DIFF
--- a/crates/typst-eval/src/markup.rs
+++ b/crates/typst-eval/src/markup.rs
@@ -62,7 +62,7 @@ fn eval_markup<'a>(
                             ));
                         }
 
-                        *elem = std::mem::take(elem).labelled(label);
+                        *elem = std::mem::take(elem).labelled(label, expr.span());
                     } else {
                         vm.engine.sink.warn(warning!(
                             expr.span(),
--- a/crates/typst-ide/src/definition.rs
+++ b/crates/typst-ide/src/definition.rs
@@ -182,6 +182,12 @@ mod tests {
     #[test]
     fn test_definition_ref() {
         test("#figure[] <hi> See @hi", -2, Side::After).must_be_at("main.typ", 1..9);
+        let source =
+            r#"#let test1(body) = figure(body); #test1([Test1]) <fig:test1> @fig:test1"#;
+        test(source, -2, Side::After).must_be_at("main.typ", 19..31);
+        let source = r#"#let test1(body) = figure(body); #test1([Test1]) <fig:test1> @fig:test1
+#let test2(body) = test1(body); #test2([Test2]) <fig:test2>; @fig:test2"#;
+        test(source, -2, Side::After).must_be_at("main.typ", 19..31);
     }
 
     #[test]
--- a/crates/typst-ide/src/definition.rs
+++ b/crates/typst-ide/src/definition.rs
@@ -75,7 +75,10 @@ pub fn definition(
             let label = Label::new(PicoStr::intern(node.cast::<ast::Ref>()?.target()));
             let selector = Selector::Label(label);
             let elem = document?.introspector.query_first(&selector)?;
-            return Some(Definition::Span(elem.span()));
+            let labelled_at = elem.labelled_at().or(elem.span());
+            if !labelled_at.is_detached() {
+                return Some(Definition::Span(labelled_at));
+            }
         }
 
         _ => {}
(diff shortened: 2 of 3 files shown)


laurmaedje:
I'm concerned with the growth of the `Content` type's memory size. Especially once values and content will be merged, I'd like to have as little metadata attached to values as possible.

Myriad-Dreamin:
> I'm concerned with the growth of the `Content` type's memory size. Especially once values and content will be merged, I'd like to have as little metadata attached to values as possible.

The performance I checked for this change:
- The cold compilation time doesn't grow on some document has hundreds of labels (but I didn't test the documents creating labels by scripting heavily).
- Only the labelled content will be affected. And if the labelled content is unique in the document, we can also store the span without cost. As the size of content increased, the probability of duplicated and labelled content will decrease rapidly.

laurmaedje:
> Only the labelled content will be affected. 

Looking at the implementation, all Content will be increased in size, right?

> And if the labelled content is unique in the document, we can also store the span without cost.

How do you mean?

Myriad-Dreamin:
> Looking at the implementation, all Content will be increased in size, right?

I hover it and the size of the `Inner<dyn Bounds>` is `size = 80 (0x50), align = 0x10`. The size isn't changed with or without the `labelled_at: Span`. I guess the align will be always `16` since the `Location` is `u128` if the LLVM doesn't have bug.

> How do you mean?

I only consider that comemo may store duplicated instance. If a content is created once and labelled once, there will be at most two instances in the cache:
```rs
Content { elem: the_data, label: None }
Content { elem: the_data, label: Some(the_label) }
```

If the content is created and attached with a label at the same time, the comemo cache might perhaps don't store the content for two times, but for those packages creating content and attaching labels separately, the content may be duplicated, such as `label-it` in touying (https://github.com/touying-typ/touying/blob/3d8819c8e13fc9fc80620da26dc7ec48970c44f0/src/utils.typ#L92-L108). Looking at the change:

```patch
let m = self.make_mut();
m.label = Some(label);
+ m.labelled_at = labelled_at;
```

I'm not sure whether we have already paid for it. The comemo will still store two instances:
```rs
Content { elem: the_data, label: None, span: detached }
Content { elem: the_data, label: Some(the_label), span: span_of_the_label }
```

Please correct me if I'm wrongly thinking of it.

laurmaedje:
> I hover it and the size of the Inner<dyn Bounds> is size = 80 (0x50), align = 0x10. The size isn't changed with or without the labelled_at: Span. I guess the align will be always 16 since the Location is u128 if the LLVM doesn't have bug.

Ah okay, that's lucky in this case then. Though the high alignment of `Location` is kinda unnecessary and I was considering to get rid of it via `[u64; 2]` or similar.

> I only consider that comemo may store duplicated instance. If a content is created once and labelled once, there will be at most two instances in the cache:

> If the content is created and attached with a label at the same time, the comemo cache might perhaps don't store the content for two times, but for those packages creating content and attaching labels separately, the content may be duplicated

I _think_ (without looking too closely) that, as long as label and labelled_at are always assigned together (which is kinda the point), there won't be any extra instances with this change.

laurmaedje:
Regardless of the current implementation details and padding setup, I don't currently want to commit to supporting this extra span infrastructure upstream at the moment, so I'll close this. :/

PR #6120, opened 2025-03-30
Expose function representation

Myriad-Dreamin:
This is required by analyzers to check functions case by case. Specifically, we can improve `param_completions` and `named_param_value_completions` then. It will definitely to provide more accessors on `Func` to avoid exposing the underlying func representation but it is harder to design. Therefore, simply exposing the func repr might be a good way to simplify. Only the downside imo is that functionalities depending on it will have to pay for future breaking changes.

### DIFF
--- a/crates/typst-library/src/foundations/func.rs
+++ b/crates/typst-library/src/foundations/func.rs
@@ -144,7 +144,7 @@ pub struct Func {
 
 /// The different kinds of function representations.
 #[derive(Clone, PartialEq, Hash)]
-enum Repr {
+pub enum Repr {
     /// A native Rust function.
     Native(Static<NativeFuncData>),
     /// A function for an element.
@@ -352,6 +352,11 @@ impl Func {
             self.span = span;
         }
         self
+    }
+
+    /// The function's repr
+    pub fn inner(&self) -> &Repr {
+        &self.repr
     }
 }
 
--- a/crates/typst-library/src/foundations/mod.rs
+++ b/crates/typst-library/src/foundations/mod.rs
@@ -20,7 +20,7 @@ mod duration;
 mod element;
 mod fields;
 mod float;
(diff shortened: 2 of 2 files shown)


laurmaedje:
I think it would be good to give a higher-level API design a shot. As a starting point, what operations do you currently perform on the exposed representation, i.e. what would you need from a higher-level API?

laurmaedje:
@Myriad-Dreamin I don't really want to expose the raw API publicly, but I would be okay with exposing this `#[doc(hidden)]` for Tinymist.

Myriad-Dreamin:
> @Myriad-Dreamin I don't really want to expose the raw API publicly, but I would be okay with exposing this #[doc(hidden)] for Tinymist.

Nice. I understand your worries because I was happening to find that typst exposes too many functions so that anybody can reimplement `typst::compile` outside of typst crate, just by copy-editing the `crates/typst/src/lib.rs`.

> I think it would be good to give a higher-level API design a shot. As a starting point, what operations do you currently perform on the exposed representation, i.e. what would you need from a higher-level API?

I'll still give a check and reply it on discord.

laurmaedje:
Do you want to proceed with a doc-hidden impl?

Myriad-Dreamin:
> Do you want to proceed with a doc-hidden impl?

Since #6119 is closed, tinymist cannot use typst on crates.io in near future. This makes #6120 less useful. The next steps might be seeking higher-level API design to expose functionality in a clean way.

laurmaedje:
Alright, let's close this then and go back to the drawing board for a higher-level API.

> Since https://github.com/typst/typst/pull/6119 is closed, tinymist cannot use typst on crates.io in near future.

It's somewhat unfortunate, but to be honest maybe also good for you to keep the flexibility to make minor adjustments in the future.

Issue #3332, opened 2024-02-03
Hebrew numbering shouldn't use geresh/gershayim


ocantor:
### Description

Hebrew numerals are letters or sequences of letters. Normally, they include a geresh or gershayim character, respectively (I'll call it geresh for brevity). However, the geresh is omitted in numbered lists because they usually already have a period after the numeral. The Hebrew numbering added in #446 uses geresh, so every numbered list that uses `NumberingKind::Hebrew` is misformatted.
The code currently uses the following match expression:
```rust
match n {
    15 => fmt.push_str("ט״ו"),
    16 => fmt.push_str("ט״ז"),
     _ => {
        let append_geresh = n == value && fmt.is_empty();
        if n == value && !fmt.is_empty() {
            fmt.push('״');
        }
        fmt.push(name);
        if append_geresh {
            fmt.push('׳');
        }

        n -= value;
        continue;
    }
}
```
The version without geresh would be:
```rust
match n {
    15 => fmt.push_str("טו"),
    16 => fmt.push_str("טז"),
     _ => {
        fmt.push(name);
        n -= value;
        continue;
    }
}
```

### Typst version

- [X] I am using the latest version of Typst

Enivex:
@naim94a @emilyyyylime

emilyyyylime:
I have at one point considered opening an issue just like this. From my personal experience with documents typeset in Hebrew, and especially those I would consider to have looked 'professional', I can only say that there are different uses for both styles.\
This is my attempt to summarise the contexts where each is used:
- When numbering a list, you would more often see i.e. `א.` or `א)`, as with Arabic numerals or Latin letters. However, the Geresh-ed style has precedent as well.
- When _referencing_ an Gematria[^1]-numbered ordinal item, always use Gershayim. For example, `פרק ג׳` (~"Chapter C").
[^1]: https://en.wikipedia.org/wiki/Gematria

This quickly blows up into a great multitude of interconnected deeply-rooted issues with the entire current state of numbering, counters, and references. I think when numberings are revamped there would be an easy way to configure the style of your numbering (`{א}` vs `{א׳}`).
You can join the conversation regarding these features in the Forge section of the official Discord server at https://discord.com/channels/1054443721975922748/1176472364347830383.


Until then, it is probably easier to programmatically remove the Gershayim than to add them back in, as a Typst user. For example, you may use `(..ns) => numbering("א", ..ns).replace("׳", "").replace("״", "")` as a numbering function.

[...]

PR #6122, opened 2025-03-31
numbering pattern implementation refactor

samuelireson:
See [previous pr](https://github.com/typst/typst/pull/5622) for details.

### DIFF
--- a/crates/typst-library/src/model/numbering.rs
+++ b/crates/typst-library/src/model/numbering.rs
@@ -9,7 +9,6 @@ use ecow::{eco_format, EcoString, EcoVec};
 use crate::diag::SourceResult;
 use crate::engine::Engine;
 use crate::foundations::{cast, func, Context, Func, Str, Value};
-use crate::text::Case;
 
 /// Applies a numbering to a sequence of numbers.
 ///
@@ -382,38 +381,180 @@ impl NumberingKind {
     pub fn apply(self, n: u64) -> EcoString {
         match self {
             Self::Arabic => eco_format!("{n}"),
-            Self::LowerRoman => roman_numeral(n, Case::Lower),
-            Self::UpperRoman => roman_numeral(n, Case::Upper),
-            Self::LowerGreek => greek_numeral(n, Case::Lower),
-            Self::UpperGreek => greek_numeral(n, Case::Upper),
-            Self::Symbol => {
-                if n == 0 {
-                    return '-'.into();
-                }
-
-                const SYMBOLS: &[char] = &['*', '†', '‡', '§', '¶', '‖'];
-                let n_symbols = SYMBOLS.len() as u64;
-                let symbol = SYMBOLS[((n - 1) % n_symbols) as usize];
-                let amount = ((n - 1) / n_symbols) + 1;
-                std::iter::repeat_n(symbol, amount.try_into().unwrap()).collect()
-            }
-            Self::Hebrew => hebrew_numeral(n),
-
-            Self::LowerLatin => zeroless(
+            Self::LowerRoman => additive(
+                [
+                    ("m̅", 1000000),
+                    ("d̅", 500000),
+                    ("c̅", 100000),
+                    ("l̅", 50000),
+                    ("x̅", 10000),
+                    ("v̅", 5000),
+                    ("i̅v̅", 4000),
+                    ("m", 1000),
+                    ("cm", 900),
+                    ("d", 500),
+                    ("cd", 400),
+                    ("c", 100),
+                    ("xc", 90),
+                    ("l", 50),
+                    ("xl", 40),
+                    ("x", 10),
+                    ("ix", 9),
+                    ("v", 5),
+                    ("iv", 4),
+                    ("i", 1),
+                ],
+                n,
+            ),
+            Self::UpperRoman => additive(
+                [
+                    ("M̅", 1000000),
+                    ("D̅", 500000),
+                    ("C̅", 100000),
+                    ("L̅", 50000),
+                    ("X̅", 10000),
+                    ("V̅", 5000),
+                    ("I̅V̅", 4000),
+                    ("M", 1000),
+                    ("CM", 900),
+                    ("D", 500),
+                    ("CD", 400),
+                    ("C", 100),
+                    ("XC", 90),
+                    ("L", 50),
+                    ("XL", 40),
+                    ("X", 10),
+                    ("IX", 9),
+                    ("V", 5),
+                    ("IV", 4),
+                    ("I", 1),
+                ],
+                n,
+            ),
+            Self::LowerGreek => additive(
+                [
+                    ("ϡ", 900),
+                    ("ω", 800),
+                    ("ψ", 700),
+                    ("χ", 600),
+                    ("φ", 500),
+                    ("υ", 400),
+                    ("τ", 300),
+                    ("σ", 200),
+                    ("ρ", 100),
+                    ("ϟ", 90),
+                    ("π", 80),
+                    ("ο", 70),
+                    ("ξ", 60),
+                    ("ν", 50),
+                    ("μ", 40),
+                    ("λ", 30),
+                    ("κ", 20),
+                    ("ι", 10),
+                    ("θ", 9),
+                    ("η", 8),
+                    ("ζ", 7),
+                    ("ϛ", 6),
+                    ("ε", 5),
+                    ("δ", 4),
+                    ("γ", 3),
+                    ("β", 2),
+                    ("α", 1),
+                    ("𐆊", 0),
+                ],
+                n,
+            ),
+            Self::UpperGreek => additive(
+                [
+                    ("Ϡ", 900),
+                    ("Ω", 800),
+                    ("Ψ", 700),
+                    ("Χ", 600),
+                    ("Φ", 500),
+                    ("Υ", 400),
+                    ("Τ", 300),
+                    ("Σ", 200),
+                    ("Ρ", 100),
+                    ("Ϟ", 90),
+                    ("Π", 80),
+                    ("Ο", 70),
+                    ("Ξ", 60),
+                    ("Ν", 50),
+                    ("Μ", 40),
+                    ("Λ", 30),
+                    ("Κ", 20),
+                    ("Ι", 10),
+                    ("Θ", 9),
+                    ("Η", 8),
+                    ("Ζ", 7),
+                    ("Ϛ", 6),
+                    ("Ε", 5),
+                    ("Δ", 4),
+                    ("Γ", 3),
+                    ("Β", 2),
+                    ("Α", 1),
+                    ("𐆊", 0),
+                ],
+                n,
+            ),
+            Self::Symbol => symbolic(['*', '†', '‡', '§', '¶', '‖'], n),
+            Self::Hebrew => additive(
+                [
+                    ("י׳", 10000),
+                    ("ט׳", 9000),
+                    ("ח׳", 8000),
+                    ("ז׳", 7000),
+                    ("ו׳", 6000),
+                    ("ה׳", 5000),
+                    ("ד׳", 4000),
+                    ("ג׳", 3000),
+                    ("ב׳", 2000),
+                    ("א׳", 1000),
+                    ("ת", 400),
+                    ("ש", 300),
+                    ("ר", 200),
+                    ("ק", 100),
+                    ("צ", 90),
+                    ("פ", 80),
+                    ("ע", 70),
+                    ("ס", 60),
+                    ("נ", 50),
+                    ("מ", 40),
+                    ("ל", 30),
+                    ("כ", 20),
+                    ("יט", 19),
+                    ("יח", 18),
+                    ("יז", 17),
+                    ("טז", 16),
+                    ("טו", 15),
+                    ("י", 10),
+                    ("ט", 9),
+                    ("ח", 8),
+                    ("ז", 7),
+                    ("ו", 6),
+                    ("ה", 5),
+                    ("ד", 4),
+                    ("ג", 3),
+                    ("ב", 2),
+                    ("א", 1),
+                ],
+                n,
+            ),
+            Self::LowerLatin => alphabetic(
                 [
                     'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n',
                     'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
                 ],
                 n,
             ),
-            Self::UpperLatin => zeroless(
+            Self::UpperLatin => alphabetic(
                 [
                     'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N',
                     'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
                 ],
                 n,
             ),
-            Self::HiraganaAiueo => zeroless(
+            Self::HiraganaAiueo => alphabetic(
                 [
                     'あ', 'い', 'う', 'え', 'お', 'か', 'き', 'く', 'け', 'こ', 'さ',
                     'し', 'す', 'せ', 'そ', 'た', 'ち', 'つ', 'て', 'と', 'な', 'に',
@@ -423,7 +564,7 @@ impl NumberingKind {
                 ],
                 n,
             ),
-            Self::HiraganaIroha => zeroless(
+            Self::HiraganaIroha => alphabetic(
                 [
                     'い', 'ろ', 'は', 'に', 'ほ', 'へ', 'と', 'ち', 'り', 'ぬ', 'る',
                     'を', 'わ', 'か', 'よ', 'た', 'れ', 'そ', 'つ', 'ね', 'な', 'ら',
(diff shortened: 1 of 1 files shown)


samuelireson:
Currently, there is a difference between the typst implementation of greek numbering and the css counter style implementation - there is a ' like symbol appended to the numeral if it has 0 in the thousands position. I can't find why this should be the case.

laurmaedje:
There's actually two Greek styles in CSS: [lower-greek](https://www.w3.org/TR/predefined-counter-styles/#lower-greek) and  [greek-lower-modern](https://www.w3.org/TR/predefined-counter-styles/#greek-lower-modern). The original implementation in Typst was made, I believe, with help of this website: https://www.russellcottrell.com/greek/utilities/GreekNumberConverter.htm

I like how clean the implementation is with the numbering systems, but for Hebrew and Greek there was definitely more going on in the code before than is now, so the old implementation must've implemented something at least somewhat different.

---

For Roman, Typst used to give `N`, but now gives `0`. I think to make sure these refactors are non-breaking, we should do some automated comparison. E.g. iterating through all numbering systems and for each through all numbers from 0 to 100_000, checking that the output is either matching (or deviating in an expected way if it turns out there was some bug).

samuelireson:
> There's actually two Greek styles in CSS: [lower-greek](https://www.w3.org/TR/predefined-counter-styles/#lower-greek) and [greek-lower-modern](https://www.w3.org/TR/predefined-counter-styles/#greek-lower-modern).

I saw this too, but it didn't seem that the Typst implementation matched either - would you like to continue to support greek numbering in line with  https://www.russellcottrell.com/greek/utilities/GreekNumberConverter.htm , or choose one of the css numbering styles. I guess the latter is not ideal since it would be breaking.

> I think to make sure these refactors are non-breaking, we should do some automated comparison. E.g. iterating through all numbering systems and for each through all numbers from 0 to 100_000, checking that the output is either matching (or deviating in an expected way if it turns out there was some bug).

This sounds like a good idea - how would you suggest actually doing this?

laurmaedje:
> I saw this too, but it didn't seem that the Typst implementation matched either - would you like to continue to support greek numbering in line with https://www.russellcottrell.com/greek/utilities/GreekNumberConverter.htm , or choose one of the css numbering styles. I guess the latter is not ideal since it would be breaking.

I'm not sure. I would be interested to hear @LuxxxLucy's thoughts who [contributed](https://github.com/typst/typst/pull/4273) the feature originally.

> This sounds like a good idea - how would you suggest actually doing this?

I would do this as a one-off and not keep the code around, so any way is fine really. You could create an extra Rust crate that depends on both your branch and on typst:main and compare both. Or just generate a text file with both main and dev and run `diff` on it. Something like that.

LuxxxLucy:
That acute like sign is called `keraia` and is used in modern greek numerals to mainly make it stand out and distinguish as a number (instead of regular text), a short explanation see [wiki - Greek Numeral](https://en.wikipedia.org/wiki/Greek_numerals#Description). It is more often to see it used in academic and formal writing.

The css spec is a rather simplified one, and if you remove keraia (both left and right one) you will find that large number (let us say larger than 10000) will be harder to recognize. That being said, a simplified version is definitely not wrong, especially if we are using it for header/section/chapters in typst documents, where we typically will not have huge numbers.

So I would suggest this, try best to keep the keraia (both the lower left and the up right) if possible (btw I like the proposed change, it is neat). But if that is not possible, I think I will probably approve the simplified version (keeps up with the CSS spec).

samuelireson:
Hmm, I am not certain which approach is more appropriate. I would lean towards adhering to the css spec since this is likely what users are expecting/used to anyway, but maybe this is just because the implementation is cleaner! If the choice is made to stick with the current greek numbering, then I don't think this logic could be suitably encapsulated by the standard numbering systems, so we could just keep @LuxxxLucy existing function.

laurmaedje:
@LuxxxLucy Is the keraia similar to thousands separators in English (e.g. "1,000,000") or is it a different kind of thing?

samuelireson:
Currently having a look at the changes caused by the refactor, and there are differences for the "circled" numbering kinds. In typst main, single circled numbers are treated as base-50 symbols, so 51 is shown as ①①. I would call this undesirable. The CSS style represents this number as "51", forgetting the symbols. Another case of this occurs in the double circled numbers, but maybe not in a way which is undesirable (at least not totally so). In current typst, 990 -> ⓽⓼⓾ (bad), but 991 -> ⓽⓽⓵ (maybe good?). The refactor would return these cases as "990", and "991".

There are also differences in the greek implementation, and some small differences in hebrew which I am currently investigating.

LuxxxLucy:
> @LuxxxLucy Is the keraia similar (as in, being optional, but added for readability) to thousands separators in English (e.g. "1,000,000") or is it a different kind of thing?

yes, I think it is for readability and is optional.

laurmaedje:
Okay, then I think we don't really need as we also don't do the equivalent for Arabic numerals.

samuelireson:
Okay, nice. I will also make a change to the test cases to reflect that. What do you think is a good testing strategy here @laurmaedje? I feel like currently testing seemingly random instances of numbering in each style could be improved, but I'm not really sure what to change to.

laurmaedje:
I agree that the testing right now isn't great, but I also don't have a great idea for persistent tests tbh. For the change itself, comparing against the old version works, but that's not suitable for retained tests unfortunately.

samuelireson:
I have made the changes such that the current tests will pass. As discussed I think there is probably a better way to test this, but there's no reason why this couldn't come in the future.

mohe2015:
What about committing the expected output and adding a test that compares with this expected output?

laurmaedje:
I'm also concerned with size. I guess we could commit a hash of the first 10k numbers as a string. That would catch regressions. But it would be pretty opaque if it failed, so probably rather frustrating.

samuelireson:
I guess that's an option but not a great one. I would personally be in favour of leaving the testing how it is now and addressing this later - we aren't losing anything relative to the current state by doing this.

samuelireson:
Just checking if there is anything left to do on this from my side?

laurmaedje:
@samuelireson I'm super busy with other stuff right now and thus unfortunately haven't gotten to any open-source work, but I hope to return to the open-source work in a week or two. Sorry for the slowness.

samuelireson:
@laurmaedje no stress, just checking I hadn't missed something.

laurmaedje:
@samuelireson do you still have the code which you used to compare main and this branch so that I can reproduce your findings?

crates/typst-library/src/model/numbering.rs
```diff
@@ -502,306 +665,171 @@ impl NumberingKind {
...
+                numeric(['০', '১', '২', '৩', '৪', '৫', '৬', '৭', '৮', '৯'], n)
+            }
+            Self::Symbol => symbolic(['*', '†', '‡', '§', '¶', '‖'], n),
         }
     }
 }
 
-/// Stringify an integer to a Hebrew number.
-fn hebrew_numeral(mut n: u64) -> EcoString {
+/// Stringify a number using symbols representing values. The decimal representation of the number
```
laurmaedje: Please wrap doc comments at column 80

laurmaedje (approved):
The code looks very good! I only want to confirm the test results myself and then this is good to merge.

samuelireson:
> @samuelireson do you still have the code which you used to compare main and this branch so that I can reproduce your findings?

Unfortunately, I took

> I would do this as a one-off and not keep the code around

fairly literally. As far as I can remember I did just checkout each branch and write the contents to a txt file to diff. Sorry I don't have it, this was quite a while ago.

laurmaedje:
No problem.

> There are also differences in the greek implementation, and some small differences in hebrew which I am currently investigating.

Greek is cleared up. Did you find out what changed with Hebrew?

crates/typst-library/src/model/numbering.rs
```diff
@@ -502,306 +665,172 @@ impl NumberingKind {
...
+/// 1 => 'A'
+/// 2 => 'B'
+/// 3 => 'C'
+/// 4 => '4'
+/// ...
+/// n => 'n'
+/// ```
+fn fixed<const N_DIGITS: usize>(symbols: [char; N_DIGITS], n: u64) -> EcoString {
+    let n_digits = N_DIGITS as u64;
+    if n - 1 < n_digits {
```
laurmaedje: This panics / underflows if n is zero. I think we should instead expect a zero to be part of the symbols array. There is ⓪ (U+24EA) for the circled numbers. For the double circled ones, we can use a normal `0` instead.

samuelireson:
There were similar accents added to the numerals which are unaccounted for on the ready made styles page [here](www.w3.org/TR/predefined-counter-styles/#hebrew-styles). From what I found, it is a similar situation to the Greek accents in that they are not actually necessary for understanding the represented digit.

laurmaedje:
Maybe @emilyyyylime can chime in on the Hebrew topic. TLDR: The old numbering code emitted "א׳" for 1 and the new one emits "א". Similarly, "צ״א" becomes "צא". Is this similar to omitting a thousands separators in English or is the punctuation important and should be kept?

laurmaedje:
The Hebrew changes seem to go further than just the punctuation. E.g. "תת״ר" becomes "א׳".

crates/typst-library/src/model/numbering.rs
samuelireson: This was an oversight. The other methods already deal with 0 case, so I think just a small edit here. Good catch :)

samuelireson:
> The Hebrew changes seem to go further than just the punctuation. E.g. "תת״ר" becomes "א׳".

which number is this for?

laurmaedje:
> which number is this for?

1000

samuelireson:
> 1000

Okay, I am fairly confident that the new version is correct in this case. I think the old version just didn't define enough numeric representations of characters. The largest numeral was of value 400, which I guess is why there are repeats - 400 + 400 + 200.

samuelireson:
> why there are repeats

By this I mean repeated characters in the representation of the digit.

crates/typst-library/src/model/numbering.rs
```diff
@@ -502,306 +665,172 @@ impl NumberingKind {
...
+/// 4 => 'IV'
+/// 5 => 'V'
+/// 6 => 'VI'
+/// 7 => 'VII'
+/// 8 => 'VIII'
+/// ```
+///
+/// where this is the start of the familiar Roman numeral system.
+fn additive<const N_DIGITS: usize>(
+    symbols: [(&str, u64); N_DIGITS],
```
emilyyyylime: Any reason for this not to be a slice? Should improve compilation times and binary size

emilyyyylime: also applies to the other functions

crates/typst-library/src/model/numbering.rs
```diff
@@ -502,306 +665,172 @@ impl NumberingKind {
...
-                    }
-                    fmt.push(name);
-                    if append_geresh {
-                        fmt.push('׳');
-                    }
-
-                    n -= value;
-                    continue;
-                }
+        for (symbol, weight) in symbols {
```
emilyyyylime: The lowest value symbol should always come last should it not? so this can be `if let Some(symbol) = symbols.last().and_then(|(s, w)| (w == 0).then_some(s))`

crates/typst-library/src/model/numbering.rs
```diff
@@ -381,39 +380,203 @@ impl NumberingKind {
...
+                    ("נ", 50),
+                    ("מ", 40),
+                    ("ל", 30),
+                    ("כ", 20),
+                    ("יט", 19),
+                    ("יח", 18),
+                    ("יז", 17),
+                    ("טז", 16),
+                    ("טו", 15),
+                    ("י", 10),
```
emilyyyylime, line 541: I take it you've verified this works to preserve the exception for 15 and 16? It seems like it would but I just want to make sure

samuelireson: yes, this works. Do you have any comment regarding the geresh usage in hebrew numbering? Is it similar to comma separation in decimal numbering or is it necessary for understanding?

crates/typst-library/src/model/numbering.rs
samuelireson: I'm sorry, I don't know this concept! Any chance you could give me an example or point me to one?

crates/typst-library/src/model/numbering.rs
samuelireson: What exactly do you mean that the lowest value symbol should always come last? I'm not certain where the code you provided would fit in.

crates/typst-library/src/model/numbering.rs
samuelireson: Do you just mean changing the whole symbols array to a slice?

samuelireson: Maybe you are suggesting a more substantial change and I am missing the point.

laurmaedje: She means
```rust
fn additive(
    symbols: &[(&str, u64)],
    mut n: u64,
) -> EcoString {
```
so that the function is only compiled once and not once for each different N_DIGITS. If N_DIGITS were used in the function (not here, but applies to other functions), it would just be `symbols.len()`.

I agree that this would be a good change to make.

EDIT: I only saw now that you deleted your comment and wrote a different one. So now the answer would just be yes.

samuelireson: Ah okay that makes sense, thanks. Not sure what inspired the use of N_DIGITS in the first place.

crates/typst-library/src/model/numbering.rs
laurmaedje: The `symbols` array is ordered by weight, so the value 0 will be at the very last spot or nowhere.

```rust
for (symbol, weight) in symbols {
    if weight == 0 {
        return (*symbol).into();
    }
}
return '0'.into();
```

would become 

```rust
if let Some(&symbol) = symbols.last().and_then(|&(s, w)| (w == 0).then_some(s)) {
    return symbol.into();
}
return '0'.into();
```

It can actually be written even a bit nicer this way:

```rust
if let Some(&(symbol, 0)) = symbols.last() {
    return symbol.into();
}
return '0'.into();
```

samuelireson: My rust inexperience is shining through here. I will alter.

samuelireson:
sorry for the commit spam, I keep thinking it's the last change that needs to be made.

laurmaedje:
No problem, it will be squashed anyway. 

This looks good to me now. Given that @emilyyyylime commented on the code but didn't say anything against the punctuation change (https://github.com/typst/typst/pull/6122#issuecomment-2935596045), I guess it's fine? I'd wait a bit more in case there's another comment, and then merge.

samuelireson:
great! thanks for the help :)

emilyyyylime:
Sorry yes I was still writing the second comment!

RE `⁨׳⁩` & `⁨״⁩`: I've previously shared [some thoughts](https://github.com/typst/typst/issues/3332#issuecomment-1927174704) on the matter. However, I think I have since warmed up to the idea of changing the default, especially given the Academy of the Hebrew Language's recommendation on the matter:

> remark a: in letters used to denote numbers of mostly chapters, verses, pages, lines, etc.; there is no need to add a Geresh or Gershayim[^1]

Still I will note it can be preferential in many cases to use the additional marks, and I still think it's necessary to find a long term solution that allows to choose between variations of the same numbering system.

I have more thoughts about א׳ for 1000. For one I think it's wrong when used by itself like this and only works for numbers beyond 1000, as `⁨׳⁩` essentially functions as a digit separator to turn a non-positional system into a positional one in this case (think of [the subdivisions of a degree](https://en.wikipedia.org/wiki/Degree_(angle)#Subdivisions) or the [Vinculum](https://en.wikipedia.org/wiki/Roman_numerals#Vinculum) in Roman Numerals) (so א׳א can be unambiguously 1001 but א׳ just looks like 1. This is ambiguous for any multiple of 1000). I found `⁨א°⁩` (mentioned)[https://he.wikipedia.org/wiki/%D7%A1%D7%A4%D7%A8%D7%95%D7%AA_%D7%A2%D7%91%D7%A8%D7%99%D7%95%D7%AA] once in a table on the Hebrew Wikipedia but couldn't find references to it elsewhere so I would not recommend it.

From my understanding this notation is only really used for marking the year in the Hebrew Calendar, and is most often just omitted for brevity. So many speakers are plain unfamiliar with its meaning (at least I wasn't fully aware of the difference between ה׳תשפ״ה and תשפ״ה). For that reason and because it's rare to count that many items I think it best to omit this part of the PR

[^1]: https://hebrew-academy.org.il/topic/hahlatot/punctuation/#target-3473 (הערה א) [translation mine]

samuelireson:
That makes sense. Thanks for the detailed answer. In this case do you think I should just remove the weights above 400 - matching the maximally supported number from the previous implementation?

Andrew15-5:
> I still think it's necessary to find a long term solution that allows to choose between variations of the same numbering system.

I think this is just https://github.com/typst/typst/issues/1177, like https://github.com/typst/typst/pull/5597#issuecomment-2551816451.

emilyyyylime:
LGTM, at least for now.

laurmaedje:
Thanks @samuelireson for the work & patience and @emilyyyylime for the insights!

@samuelireson If you're still motivated, I think we're ready for the next step of https://github.com/typst/typst/pull/5622#issuecomment-2765808975 now. (Though as mentioned, step 2 still warrants some discussion.)

samuelireson:
I'm still interested in moving on to steps 2 and 3 as outlined. Where is a good place for this discussion?

laurmaedje:
We can do it in https://github.com/typst/typst/pull/5622 or you can open a draft PR and we discuss it there.

Issue #6006, opened 2025-03-04
Add `dictionary.map` method to map values


MDLC01:
### Description

Add a `dictionary.map` method to map the values of a dictionary. To be specific about the implementation, it would work in the following way:

```typ
#let dictionary-map(self, f) = {
  self
    .pairs()
    .map(((k, v)) => (k, f(v)))
    .to-dict()
}
```

An open question is whether this is the right behavior, or whether the function should instead accept and return key-value pairs.

### Use Case

Easier manipulation of dictionaries.

laurmaedje:
Is there precedence for this in other common programming languages?

MDLC01:
OCaml has both [`Map.S.map` and `Map.S.imap`](https://ocaml.org/manual/5.3/api/Map.S.html#transforming) for mapping the _values_ of a map (the second also gives the keys to the mapping function).

I am not familiar with Haskell at all, but it seems to have [many such map functions](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html#g:12). The one simply called `map` maps the _values_ of a map (and does not provide the keys to the mapping function).

Python `dict`, Rust `HashMap`, and C++ `unordered_map`, do not seem to have such features.

I think it makes sense to map values only. I'm unsure whether the mapping function should accept the keys as well, though.

laurmaedje:
On Discord, there was a proposal by @SillyFreak for `map-values` and `map-pairs` which would be less ambigious.

ensko:
To be specific, my suggestion was

```typ
#let map-values(self, f) = {
  // unchanged from the original
  self
    .pairs()
    .map(((k, v)) => (k, f(v)))
    .to-dict()
}

#let map-pairs(self, f) = {
  self
    .pairs()
    .map(((k, v)) => f(k, v))
    .to-dict()
}
```

I think `map-values()` is the common case; keys are not changed and are not necessary to compute the new value. `map-pairs()` is the most general; you have access to both key and value and can produce new keys and values. There is a bit of ambiguity what should happen if duplicate keys are produced. As written [`to-dict()`](https://typst.app/docs/reference/foundations/array/#definitions-to-dict) would preserve the last occurrence of each key.

MDLC01:
An argument in favor of `dictionary.map` as a method that maps values only (i.e., `value -> value`) is that we could additionally add `arguments.map` to map values of an `arguments`, and it would work the same way as `dictionary.map` for named arguments and the same way as `array.map` for named arguments.

PR #6132, opened 2025-04-03
Add `map` function to `Dict` 

wesleyel:
impl #6006

### DIFF
--- a/crates/typst-library/src/foundations/dict.rs
+++ b/crates/typst-library/src/foundations/dict.rs
@@ -264,8 +264,8 @@ impl Dict {
     ///
     /// ```example
     /// #let prices = (apples: 2, oranges: 3, bananas: 1.5)
-    /// #prices.map(key => key.len()) \
-    /// #prices.map((key, price) => (key, price * 1.1))
+    /// #prices.map(pair => pair.at(0).len())
+    /// #prices.map((key, value) => (key, value * 1.1))
     /// ```
     #[func]
     pub fn map(
@@ -275,6 +275,7 @@ impl Dict {
         /// The function to apply to each key-value pair.
         /// The function can either take a single parameter (receiving a pair as array of length 2),
         /// or two parameters (receiving key and value separately).
+        /// Parameters exceeding two will be ignored.
         mapper: Func,
     ) -> SourceResult<Value> {
         let mut dict_result = IndexMap::new();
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/foundations/dict.rs
```diff
@@ -254,6 +256,101 @@ impl Dict {
             .map(|(k, v)| Value::Array(array![k.clone(), v.clone()]))
             .collect()
     }
+
+    /// Produces a new dictionary or array by transforming each key-value pair with the given function.
+    ///
+    /// If the mapper function returns a pair (array of length 2), the result will be a new dictionary.
+    /// Otherwise, the result will be an array containing all mapped values.
```
bluebear94, line 263: Perhaps this should be a different function: you might want to return an array even when the function returns or might return an array of length 2.

wesleyel:
Thanks for the feedback, and apologies for the delayed response — I was away on holiday and just got back.

You're absolutely right, the current implementation has issues and I’m happy to revise it. I’ll make sure that map only accepts two arguments (key and value) and always returns a dictionary, not an array.

One question I have is how to best handle the case where the mapped keys are not unique. Since dictionaries can't have duplicate keys, should we silently overwrite them, throw an error, or skip duplicates? Would love to hear your thoughts on what the expected behavior should be in that case.

> Produces a new dictionary by transforming each key-value pair with the given function.
> 
> The mapper function takes two parameters (key and value) and must return a valid key-value pair,
> which is an array of length 2 with a string as the first element.
> 
```typst
#let prices = (apples: 2, oranges: 3, bananas: 1.5)
#prices.map((key, value) => (key, value * 1.1))
```

EpicEricEE:
I'm wondering if mapping keys is something we want at all, or if the function signature should rather just be `(key, value) => value`. The issue also only mentions mapping the _values_ of a dictionary.

laurmaedje:
I think the fact that there are various different interpretations of what this API should do exactly does pose the question of whether we should add this in the first place. I'm not saying necessarily no, but it's question worth asking.

MDLC01:
I think there would be too much ambiguity regarding the behavior if we allowed mapping keys as well, and I would expect mapping values only to be the most common use case (although it would be great to have some data to back that up). So I would be in favor of the mapper being `key => key`.

laurmaedje:
I don't want to rule out the possiblity of some sort of map function in the future (there was e.g. the proposal of `map-values` and `map-pairs`), but a bare `map` is too ambigious in my opinion, so I'll close this PR. Thank you still!

PR #6133, opened 2025-04-03
Add `add-supplement` parameter to `outline.entry.prefix`

MDLC01:
This adds an `add-supplement` argument to `outline.entry.prefix`. This is a boolean or `auto`, and its purpose is to determine whether the supplement should be displayed. `auto` (the default) keeps the same behavior as before.

I'm not sure whether this is a knob that deserves to be added, but it was easy to do so I just did it so we have an implementation right away. My use case is for displaying the supplement for level-1 headings (which I changed to `[Chapter]`) in the outline. Alternatively, I could make this part of the numbering format, but then referencing the heading would result in <code>Section Chapter <i>n</i></code> being displayed.


laurmaedje:
We've discussed this in the team. As far as we see it, there are basically two use cases for this:

- Adding the supplement when it's not currently there (your use case)

  This is quite simple already I think:
  ```typ
  #show outline.entry.where(level: 1): it => it.indented(
    it.element.supplement + [ ] + it.prefix(), 
    it.body()
  )
  ```
  (With `it.element.supplement + [ ] + it.prefix()` being barely longer than `it.prefix(add-supplement: true)`.)

- Removing the supplement when it's currently there.

  This a _bit_ more complex. To get just the numbering, you'd have to do something like:
  ```typ
  #show outline.entry.where(level: 1): it => it.indented(
    numbering(it.element.numbering, ..counter(it.element.func()).at(it.element.location())), 
    it.body()
  )
  ```
  But I think this use case is even rarer. And it's better served by a way to generally make formatting a counter with a numbering at a specific location simpler.

Overall, it seemed a bit too niche to us, so I'll close. Thank you still.

PR #6134, opened 2025-04-04
Add HTML meta tags for document authors and keywords

Approximately-Equal:
This ensures that all currently available metadata contained in the document is converted to html meta tags. There is now support for `title`, `description`, `authors`, `keywords`, and `date`.

### DIFF
--- a/crates/typst-html/src/lib.rs
+++ b/crates/typst-html/src/lib.rs
@@ -6,11 +6,9 @@ pub use self::encode::html;
 
 use comemo::{Track, Tracked, TrackedMut};
 pub use ecow::EcoVec;
-use typst_library::diag::{bail, warning, At, SourceDiagnostic, SourceResult, StrResult};
+use typst_library::diag::{bail, warning, At, SourceResult};
 use typst_library::engine::{Engine, Route, Sink, Traced};
-use typst_library::foundations::{
-    Content, Datetime, Smart, StyleChain, Target, TargetElem,
-};
+use typst_library::foundations::{Content, StyleChain, Target, TargetElem};
 use typst_library::html::{
     attr, tag, FrameElem, HtmlDocument, HtmlElem, HtmlElement, HtmlNode,
 };
@@ -87,7 +85,7 @@ fn html_document_impl(
 
     let output = handle_list(&mut engine, &mut locator, children.iter().copied())?;
     let introspector = Introspector::html(&output);
-    let root = root_element(&mut engine, output, &info)?;
+    let root = root_element(output, &info)?;
 
     Ok(HtmlDocument { info, root, introspector })
 }
@@ -265,14 +263,8 @@ fn handle(
 
 /// Wrap the nodes in `<html>` and `<body>` if they are not yet rooted,
 /// supplying a suitable `<head>`.
-fn root_element(
-    engine: &mut Engine,
-    output: Vec<HtmlNode>,
-    info: &DocumentInfo,
-) -> SourceResult<HtmlElement> {
-    let head = head_element(engine, info).map_err(|err| {
-        EcoVec::from([SourceDiagnostic::warning(Span::detached(), err)])
-    })?;
+fn root_element(output: Vec<HtmlNode>, info: &DocumentInfo) -> SourceResult<HtmlElement> {
+    let head = head_element(info);
     let body = match classify_output(output)? {
         OutputKind::Html(element) => return Ok(element),
         OutputKind::Body(body) => body,
@@ -282,7 +274,7 @@ fn root_element(
 }
 
 /// Generate a `<head>` element.
-fn head_element(engine: &mut Engine, info: &DocumentInfo) -> StrResult<HtmlElement> {
+fn head_element(info: &DocumentInfo) -> HtmlElement {
     let mut children = vec![];
 
     children.push(HtmlElement::new(tag::meta).with_attr(attr::charset, "utf-8").into());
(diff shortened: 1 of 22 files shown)


jbirnick:
Now by default it will always add the current date in a meta tag. I wouldn't want that. Imagine you are building your static website every couple of hours automatically, and the date is updated there all the time, while e.g. search engines might interpret it as the date when the blog post was written or so.

In fact, it's not even clear what the "date" meta tag should represent (date of creation, date of modification, date of generation), right? I think there is no standard for this, for example it's not listed [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name).

laurmaedje:
For HTML, auto date should mean none I think. An explicitly specified date could be added to the metadata.

Approximately-Equal:
That makes sense I'll make the necessary changes to the default behavior. I originally added that behavior only because its what the pdf does. 

Don't look at the commits, I'm really bad at this. Just squash everything and let's never speak of this catastrophe.

Approximately-Equal:
> Tip for when commit fixes become a mess, in case you didn't know: You can use `git commit --amend` to just change the previous commit instead of making a new one. Then force-push.

My bad, I forgot that this is my branch, so I thought once I committed, it was immutable. I'm going to rebase and force a git history rewrite to clean up the commit history.

Approximately-Equal:
I've rebased everything into reasonable chucks. I decided to keep two revisions, the first where I had the date, and the second where its been omitted (this is to make the discussion make sense). When it gets merged I'd expect that its all squashed into a single commit.

NOTE: @laurmaedje has communicated that `generator` is unlikely to be a meta tag. If they change their mind, that should be a part of this PR, and if not, then let this PR close that non-existent issue.

crates/typst-html/src/lib.rs
```diff
@@ -302,6 +303,24 @@ fn head_element(info: &DocumentInfo) -> HtmlElement {
...
+                .with_attr(attr::content, info.author.join(", "))
+                .into(),
+        )
+    }
+
+    if !info.keywords.is_empty() {
+        children.push(
+            HtmlElement::new(tag::meta)
+                .with_attr(attr::name, "keywords")
+                .with_attr(attr::content, info.keywords.join(", "))
```
PgBiel, line 319: Do we know if they should be separated by commas and spaces instead of just commas? MDN[^1] only claims they are "separated by commas", so maybe no spaces is "more correct" (even if less readable by humans)? Though I'm finding examples of both usages around[^2], so maybe it doesn't matter...

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name
[^2]: https://stackoverflow.com/a/63637855

PgBiel (commented):
It appears that keywords might be less relevant nowadays from a SEO perspective, specifically on Google since they published a video[^1] many years ago saying they completely ignore it, though that's only Google and we don't know how true that still is... So consider this as more of a fun fact than anything - of course, Typst should still produce semantically accurate documents.

[^1]: https://www.youtube.com/watch?v=jK7IPbnmvVU

PgBiel:
> For HTML, auto date should mean none I think. An explicitly specified date could be added to the metadata. Edit: If there is a standard, that is. If not, it should probably just be omitted.

It would appear that the widely used (but not really part of the HTML standard IIUC) OpenGraph standard[^1] has some date support under `article:` and `book:` (e.g. `book:published_date`), but we can't really assume anything.
Though we may want to provide a way to generate OpenGraph metadata in the future (maybe in the form of a package?).

[^1]: https://ogp.me/

Approximately-Equal:
> > For HTML, auto date should mean none I think. An explicitly specified date could be added to the metadata. Edit: If there is a standard, that is. If not, it should probably just be omitted.
> 
> It would appear that the widely used (but not really part of the HTML standard IIUC) OpenGraph standard[1](#user-content-fn-1-f0b65f11a54b86cb7e34224d62b457d8) has some date support under `article:` and `book:` (e.g. `book:release_date`), but we can't really assume anything. Though we may want to provide a way to generate OpenGraph metadata in the future (maybe in the form of a package?).
> 
> ## Footnotes
> 1. https://ogp.me/ [↩](#user-content-fnref-1-f0b65f11a54b86cb7e34224d62b457d8)

I think having it as a package is best. The HTML is supposed to be minimal, human readable, and correct. Open-graph is not a web standard, its a protocol that allows you to share websites with tagged data. If it is used, it should be because the user has specified that it should be added. 

Another idea is maybe add open-graph tags to a special html tag that specifies the metadata, and inherits from document. This could also make it so there is a standard API for things like custom stylesheets later down the line. Just a thought.

crates/typst-html/src/lib.rs
Approximately-Equal: According to (WC3)[https://www.w3.org/TR/2011/WD-html5-author-20110809/common-microsyntaxes.html#set-of-comma-separated-tokens] section 2.4.7 _Comma-separated tokens_

> A set of comma-separated tokens is a string containing zero or more tokens each separated from the next by a single U+002C COMMA character (,), where tokens consist of any string of zero or more characters, neither beginning nor ending with [space characters](https://www.w3.org/TR/2011/WD-html5-20110525/common-microsyntaxes.html#space-character), nor containing any U+002C COMMA characters (,), and optionally surrounded by [space characters](https://www.w3.org/TR/2011/WD-html5-20110525/common-microsyntaxes.html#space-character).
>
> > For instance, the string " a ,b,,d d " consists of four tokens: "a", "b", the empty string, and "d d". Leading and trailing whitespace around each token doesn't count as part of the token, and the empty string can be a token.
>
> [Sets of comma-separated tokens](https://www.w3.org/TR/2011/WD-html5-author-20110809/common-microsyntaxes.html#set-of-comma-separated-tokens) sometimes have further restrictions on what consists a valid token. When such restrictions are defined, the tokens must all fit within those restrictions; other values are non-conforming. If no such restrictions are specified, then all values are conforming.
.

So both are fine. I thus suggest having a ", " as its more common (in writing) and thus "readable".

crates/typst-html/src/lib.rs
```diff
@@ -5,6 +5,7 @@ mod encode;
 pub use self::encode::html;
 
 use comemo::{Track, Tracked, TrackedMut};
+pub use ecow::EcoVec;
```
laurmaedje, line 8: That change seems to have slipped in accidentally

Approximately-Equal: Should be fixed.

laurmaedje:
Thanks!

PR #6140, opened 2025-04-04
Display color spaces in the order in which they are presented in the doc

MDLC01:
This changes the order in which color spaces are displayed to correspond to the one in which they are listed in the above table.

The current version of this documentation can be read at https://typst.app/docs/reference/visualize/gradient/#color-spaces-and-interpolation.


laurmaedje:
Thanks!

PR #6146, opened 2025-04-05
Fix typo in module docs

Andrew15-5:
Reported by acorbi on Typst Discord server.


MDLC01:
While you are at it, I would suggest rephrasing this whole sentence. I feel like "A module of definitions." does nothing to help a new user understand what a module is: it defines "module" in terms of itself, and introduces a new word without explanation—"definition".

Maybe it would be rephrased to something like "A collection of variables and functions."

laurmaedje:
I'm fine with rephrasing it, but doesn't hurt to merge the typo fix for now.

Andrew15-5:
I might do the separate PR if I have time, but I definitely will have less of it in the near future, I think.

Issue #6075, opened 2025-03-19
Configurable standard math style (italic uppercase greek, ISO 80000)


fGolke:
### Description

I would like to be able to configure the standard math style to make all (including uppercase greek) letters italic by default. This is also the style suggested by ISO 80000, so I wouldn't mind, if this were the default option. However people used to the current default (which is also the default in LaTeX) might disagree.

# Workarounds considered
I tried using a regex show rule, but that can't be manually overriden:
```typst
#show regex("[ΩΠΣΔΦΓΨΘΛΞ]"): math.italic

$
	Γ upright(Γ)
$
```
does not work as desired.

# Prior art
Unicode-math has an parameter `math-style` with options `TeX`, `ISO`, `french` and `upright`
See http://mirrors.ctan.org/macros/unicodetex/latex/unicode-math/unicode-math.pdf section 5.1.

# Duplication check
I think, this is no duplicate of https://github.com/typst/typst/issues/3238, since I don't want it to be locale dependent. Sorry for the noise, if you disagree.

### Use Case

Having italic uppercase greek by default enables one for example to use Δ both as variable and as the Laplacian operator.

Enivex:
Despite perhaps minor differences in motivation, this is a duplicate of #3238

laurmaedje:
#3238 is so broad that I think it makes sense to track this separately. (Especially with the new-ish subissues feature.)

imyxh:
anything i can do to help #6147 get merged?

Issue #3238, opened 2024-01-20
Localization of Mathematics


jakubkaczor:
### Description

I am not sure whether formulas are considered text, and whether `text`'s function argument `lang` should be considered, but I wish there was localization for mathematics.
1. In polish, there is a convention to typeset displayed integrals with limits, but it is not the default after setting `text`'s `lang` parameter.
2. Also, the "less than or equal to" sign's slanted variant should be used. Of course, the symbol `le.eq.slant` could be used directly, but it would be nice to have `<=` be interpreted for it by default. Correct me if I am wrong, but for now `<=` alias cannot be changed, even manually.
Here is an exemplar code showing the difference.
```typst
#set par(justify: true)
#set text(lang: "pl")

= How It Looks for Now

A displayed formula:
$
  integral_a^b f(x) dif x <= M.
$

An inline formula: $integral_a^b f(x) dif x$.

= How It Should Look

// This is just an example integral. Other variants should also be considered.
#show math.integral: i => math.limits(i, inline: false)

A displayed formula:
$
  integral_a^b f(x) dif x lt.eq.slant M.
$

An inline formula: $integral_a^b f(x) dif x$.
```
![The output.](https://github.com/typst/typst/assets/57760986/b765b2ca-164d-4923-aef7-95ff39130df6)

I wish the output of the second section was achievable using the source for the first section after setting `lang` to `"pl"`.

### Use Case

This feature would be useful for typesetting mathematics according to conventions in a language automatically, and more comfortably.

xkevio:
While a nice idea, it is simply not really feasible to support every mathematical convention of every language as of right now. Even then, mathematical notation isn't really universal anyway and can change or have different meaning, probably even within the Polish language.

But, you can already replace some shorthands like `<=` or `>=` to use the slanted variant. Either use a show-rule like so:
```typ
#show sym.lt.eq: $lt.eq.slant$
#show sym.gt.eq: $gt.eq.slant$
```
or the [quick-maths](https://github.com/typst/packages/tree/main/packages/preview/quick-maths/0.1.0) package. As for the integral, you've already demonstrated an easy way to modify it. I understand the idea of having a setting that automatically adheres to conventions, but typst provides the tools necessary to workaround or even enable them with just a bit of manual work. Localization is just plain difficult -- especially for such usecases -- unfortunately.

jakubkaczor:
> is simply not really feasible to support every mathematical convention of every language as of right now

[...]

PR #6147, opened 2025-04-05
Add default-style parameter to math.equation

fGolke:
# The PR
There are different conventions, which letters to make italic by default in math. This PR adds a parameter `default-style` to `math.equation` with possible values:
```rust
// Every latin or greek letter italic, [partial] upright.
Iso,
// Everything but capital greek letters italic.
Tex,
// Lowercase letters and [partial] italic.
French,
// Everything upright.
Upright
```
to allow switching between them using a `set`-rule. Currently, typst uses the TeX convention, which is kept as default.

Related to #3238 (i.e. the part concerned by #6075).

# Workarounds considered
The only workaround I came up with for using ISO style isn't feasible, as
```typst
#show regex("[Α-Ω]"): math.italic
$
  Δ != upright(Δ)
$
```
does not work as intended.


laurmaedje:
As [discussed on Discord](https://discord.com/channels/1054443721975922748/1176478139757629563/1412068962627354656), I would like to have a bit more exploration of the design space before we settle on something concrete here. For this reason, I don't want to move forward with the PR for now. Thank you still!

anton-drott:
I'm writing a physics thesis in which I will use Phi and Omega mostly in italics style and only occasionally in upright. 
At the moment I need to treat the quantities represented by latin and greek capitals differently which makes the markdown more difficult to read and maintain. 
Here is an example:
`$ italic(Phi)_"e" = limits(integral.double)_("A", thin Omega) L_"e" (x, omega) cos theta  dif A  dif italic(Omega) $`

For this thesis it would be great to use the Iso configuration of this PR. Has the development team had any chance to look through the consequence of merging it into the main code base?

VibrantTreeStump:
@anton-drott As a workaround, you could do
```typ
#let i-Phi = $italic(Phi)$
#let i-Omega = $italic(Omega)$
```

and then use it like so:
```typ
$ i-Phi_"e" = ... $
```

isuffix:
> ```
> #let i-Phi = $italic(Phi)$
> #let i-Omega = $italic(Omega)$
> ```

Note that variables used in math can't have dashes, since we subtractions like `$alpha-pi$` to work. So these would need to be named `iPhi` and `iOmega` or such.

VibrantTreeStump:
Whoops, thanks for the correction. Didn't think of that

fGolke:
1. The relevant portions of the code were moved over to codex in https://github.com/typst/codex/pull/20 and https://github.com/typst/typst/pull/6309 
2. There will be a larger math rework (by isuffix) in the foreseeable future, so I'm a bit hesitant to adapt the PR to the current structure
3. Laurenz didn't like the parameter name and I haven't come up with a better name yet

PR #6159, opened 2025-04-09
Use `codex::ModifierSet`

T0mstone:
Continuation of [codex#46](https://github.com/typst/codex/pull/46).

The TODO in `Cargo.toml` should get resolved before merging this.


T0mstone:
There's one remaining test that fails, but that's just because the version of `codex` from [codex#46](https://github.com/typst/codex/pull/46) isn't rebased onto a newer `codex` version yet, so it doesn't have `inter`.

T0mstone:
This test failure seems like a bug in the test suite, not an actual error or test failure on this PR's part. Maybe it'll be fixed on main in a bit and nothing needs to be done here other than wait for that.

saecki:
There definitely is a but in the test runner (which I'm looking into), it should not panic.

I quickly hacked something together to get the error that is causing the test runner to fail and got this, 
if you want work on that in the mean time:
```
❌ symbol-repr (tests/suite/symbols/symbol.typ:90)
  not annotated
    Error: 9:2-23:2 Assertion failed: "symbol(/n  /"⇒/",/n  (/"bar/", /"⤇/"),/n  (/"long/", /"⟹/"),/n  (/"long.bar/", /"⟾/"),/n  (/"not/", /"⇏/"),/n  (/"struck/", /"⤃/"),/n  (/"l/", /"⇔/"),/n  (/"l.long/", /"⟺/"),/n  (/"l.not/", /"⇎/"),/n  (/"l.struck/", /"⤄/"),/n)" != "symbol(/n  /"⇒/",/n  (/"bar/", /"⤇/"),/n  (/"long/", /"⟹/"),/n  (/"long.bar/", /"⟾/"),/n  (/"not/", /"⇏/"),/n  (/"l/", /"⇔/"),/n  (/"l.long/", /"⟺/"),/n  (/"l.not/", /"⇎/"),/n)"
```

Cargo.toml
```diff
@@ -47,7 +47,9 @@ clap = { version = "4.4", features = ["derive", "env", "wrap_help"] }
 clap_complete = "4.2.1"
 clap_mangen = "0.2.10"
 codespan-reporting = "0.11"
-codex = "0.1.1"
+# TODO: change to next codex version once released
+# codex = "0.1.1"
+codex = { git = "https://github.com/typst/codex", rev = "56eb2172fc0670f4c1c8b79a63d11f9354e5babe" }
```
laurmaedje:
```suggestion
codex = { git = "https://github.com/typst/codex", rev = "56eb217" }
```
no need for the extra lines

T0mstone:
Thanks @saecki, that let me fix it. The issue was just that the test hadn't been updated to include the new variants introduced in [codex#50](https://github.com/typst/codex/pull/50).

laurmaedje:
Maybe we should switch the test to a symbol that is less likely to get new variants?

T0mstone:
Hmm, I think this specific test was probably meant to test for the repr of a builtin symbol with many variants. (but also partially applied? Maybe it should be changed to a plain symbol, since there's already a  test for a partially applied repr above, or maybe it should be split since the one above doesn't test for when there's still multiple variants on the partially applied symbol)

But then the issue is that many of the symbols with enough variants are also the ones that are likely to see more additions...

If we go with the split I suggested in parentheses above, I could imagine `cc` being long enough to have a multiline repr, but complete enough to not change in the future. And for testing a partially applied one that still has more variants, `in.rev` is also probably pretty stable.

laurmaedje:
Maybe it's okay as-is. At the very least, it doesn't need to block this PR. Feel free to make another PR with a better symbol and split, but also feel free not to you if think it's pointless.

Thanks for the PR!

PR #6160, opened 2025-04-09
Use `measure` `width` argument in `layout` doc

MDLC01:
This was added in https://github.com/typst/typst/pull/4264 for this exact purpose.


laurmaedje:
Thanks!

Issue #4693, opened 2024-08-06
Automatic font expansion


Enivex:
### Description

The [microtype](https://ctan.org/pkg/microtype?lang=en) package in LaTeX provides several microtypographical features. The two main ones, enabled by default in pdfTeX and luaTeX, are protrusion and expansion.

Typst already supports a (very) limited form of protrusion, and there's an existing issue concerning improving this https://github.com/typst/typst/issues/261

However, expansion is currently completely absent from Typst, as far as I'm aware.

Automatic font expansion can subtly (default max in microtype is 2%) change the width of a font (through stretching) in order to achieve better typesetting results.

### Use Case

Font expansion can improve improve text regularity and greatly reduce the need for hyphenation.

eclecticfluff:
Note that microtype does font expansion by directly modifying glyphs. While this does work and provides a reasonable fallback, using the width axis in variable fonts that have it would be ideal.

Enivex:
> Note that microtype does font expansion by directly modifying glyphs. While this does work and provides a reasonable fallback, using the width axis in variable fonts that have it would be ideal.

Yes, that would be a better option for variable fonts. However, I do not think all of them have a width axis.

(Also typst does not yet support variable fonts.)

corneliusroemer:
The [microtype user manual](https://mirror.foobar.to/CTAN/macros/latex/contrib/microtype/microtype.pdf) is a good resource for inspiration.

Krasjet:
I'll add some more resources in case someone wants to take on the glyph expansion part:
1. The extension to the Knuth-Plass algorithm used by pdftex is explained in this [thesis](https://www.pragma-ade.nl/pdftex/thesis.pdf). (the implementation in pdftex is a [literal program](https://texdoc.org/serve/pdftex/1) as well)
2. There was a demo made by @ammonj for justification using variable font glyph expansion: https://ammonj.github.io/better-justification-demo/. It's not using Knuth-Plass, but the idea is there.

omentic:
Support for variable fonts has just been merged (https://github.com/typst/typst/pull/8425) 🎉

PR #6161, opened 2025-04-10
Support for character-level justification

diquah:
This implements a partial solution to #4693.

This PR adds a new property, `microjustification`, to `par`. Microjustification allows kerning to be dynamically applied during justification to further justify an existing line. This added kerning will never exceed `microjustification`, but may be less depending on how justified the line already is. This is equivalent to "Letter Spacing" in Adobe InDesign, and it improves readability by creating more uniform blocks of text, especially for very narrow blocks.

All glyphs have microjustification enabled by default, but this can be changed in the future.

## Example

```typst
#table(
    stroke: none,
    gutter: 1em,
    columns: 3,
)[
    *Normal Justification*
][
    *Normal + Micro Justification*
][
    *Normal + Excessive Micro Justification*
][
    // Microjustification is disabled (0em) by default.
    #set par(justify: true)
    #lorem(100)
][
    // Microtype uses 2% as a sensible default.
    #set par(justify: true, microjustification: 0.025em)
    #lorem(100)
][
    #set par(justify: true, microjustification: 0.1em)
    #lorem(100)
]
```

image

Notice that lines with excessively large spaces become more readable.

## Limitations
* ~~Microjustification occurs only after a line has been initially drafted. This means microjustification will not change the layout of a paragraph. It only further justifies existing lines by adding spacing between glyphs.~~
* This is not a complete solution to #4693 as that issue requests glyph *expansion* like microtype, but that is only one element of microtypography. Both letter spacing and expansion contribute to readability, and glyph scaling can be implemented in a later PR. The scope of this PR is dynamic kerning for justification.


MDLC01:
How doe this work with ligatures? It looks like they were disabled in the example.

diquah:
@MDLC01 They appear to be treated as a single glyph, which makes sense.

image

(Testing with **fi**)

I'm not sure what the expected behavior should be when both are enabled, but note that in this example the microjustification is what I would consider excessive. You could tastefully mix the two by using less microjustification. I think it makes sense to allow both to be enabled at the same time, which empowers the user to choose if this is acceptable behavior or not.

Enivex:
Regarding the first limitation, I would argue that one of the major benefits of microtypography is avoiding hyphenation in the first place. By only considering tracking, kerning expansion and protrusion only after line breaking has been done, we lose out on this.

As an aside, is this really a good name for this feature?

In fact, this seems to be orthogonal to justification. Adjusting tracking can be used to lessen "raggedness" even without actual justification

diquah:
@Enivex I agree with your first point. It's not a perfect solution, but I had trouble with having the microtype change the layout, because the layout engine seems to prefer microtype over hyphenation in the vast majority of cases (including some I would have personally prevented). I'll keep at it and see if I can get something working.

As to the naming, assuming all microtype justification features have been implemented, what are your thoughts on having a parameter in `par` named `microtype` with the following parameters:
- `max_shrink`: How much a glyph is allowed to be scaled horizontally in the negative direction for justification.
- `max_stretch`: How much a glyph is allowed to be scaled horizontally in the positive direction for justification.
- `max_retract`: How much extra negative spacing letters are allowed for justification.
- `max_expand`: How much extra positive spacing letters are allowed for justification.

The main difference between this and *tracking* is that these are dynamically set per line for the purpose of more uniform justification.

diquah:
Here is an updated version of the changes in this PR:

Added a new property, `microtype` to `par`. The microtype property includes two fields:
  - `max_retract`: Specifies the maximum distance a glyph is allowed to translate into its neighboring glyph during justification.
  - `max_expand`: Specifies the maximum distance a glyph is allowed to translate away from its neighboring glyph during justification.

## Example
```typst
#table(
    stroke: none,
    gutter: 1em,
    columns: 3,
)[
    *Normal Justification*
][
    *Normal + Microtypographical justification*
][
    *Normal + Excessive Microtypographical justification*
][
    // Microtype properties are 0em (disabled) by default.
    #set par(justify: true)
    #lorem(100)
][
    // The "microtype" LaTeX package uses 2% as a sensible default. Here I've opted for 1.5%.
    #set par(justify: true, microtype: (max_expand: 0.015em, max_retract: 0.015em))
    #lorem(100)
][
    #set par(justify: true, microtype: (max_expand: 0.1em, max_retract: 0.1em))
    #lorem(100)
]
```

image

You can see that the text layout now takes advantage of the microtype features, instead of only applying after a line has been committed. Using microtype justification both shortens the vertical space of text and improves readability and uniformity of blocks. The degree of microtypographical justification is configurable by the user, and disabled by default.

This is especially useful for newspapers, magazines, papers, or books, which often have narrow width columns.

This PR does not include automatic glyph stretching for justification, but because the `microtype` property is a dictionary, adding additional microtype features in the future should not break backward-compatibility.

Enivex:
> The "microtype" LaTeX package uses 2% as a sensible default

I should mention that `microtype` doesn't actually have this feature, [apparently due to a limitation in pdftex](https://tex.stackexchange.com/questions/401270/automatically-alter-letterspacing-intra-word-spacing-to-avoid-loose-lines-with). The 2% you are referring to is for font expansion specifically.

> Microtype properties are 0em (disabled) by default

In my opinion they should eventually be on by default as long as the defaults are unintrusive. Perhaps not in an initial implementation and before we also have expansion and proper protrusion though. The combination of small amounts of each of these should almost universally be preferable to having them off.

crates/typst-library/src/model/par.rs
```diff
@@ -227,6 +230,36 @@ impl ParElem {
...
+    pub max_expand: Length,
+}
+
+cast! {
+    Microtype,
+    self => Value::Dict(self.into()),
+    mut dict: Dict => {
+        let max_retract = dict.take("max_retract")?.cast()?;
+        let max_expand = dict.take("max_expand")?.cast()?;
+        dict.finish(&["max_retract", "max_expand"])?;
```
tingerrr, line 249: This should use `kebab-case` for the values in typst.

crates/typst-library/src/model/par.rs
```diff
@@ -227,6 +230,36 @@ impl ParElem {
...
+        Self { max_retract, max_expand }
+    },
+}
+
+impl From<Microtype> for Dict {
+    fn from(microtype: Microtype) -> Self {
+        dict! {
+            "max_retract" => microtype.max_retract,
+            "max_expand" => microtype.max_expand,
```
tingerrr, line 258: As above.

crates/typst-layout/src/inline/line.rs
```diff
@@ -478,6 +478,7 @@ pub fn commit(
     if remaining < Abs::zero() && shrinkability > Abs::zero() {
         // Attempt to reduce the length of the line, using shrinkability.
         justification_ratio = (remaining / shrinkability).max(-1.0);
+
```
tingerrr, line 481: It's probably best to leave a file untouched if there are no meaningful changes. From looking at the commit history this was because of previous changes being reverted.

tingerrr (commented):
Nice and simple, great work!

diquah:
I think I've made all the changes you've requested, mind giving this another review pass @tingerrr ?

tingerrr (commented):
LGTM, let's see what Laurenz says.

iampritishpatil:
Just here to say, I've used LaTeX for years primarily because I like how microtypset text looks. Excited to see it come here!

michaelfortunato:
Did a few experiments comparing this to how LaTex's article document class would render. For each experiment, on the left hand size is the LaTeX output, on the right is this branch for Typst. 
# Experiment 1
Settings
```typst
#set par(justify: true, microtype: (max-expand: 0.02em, max-retract: 0.015em))
```
image

# Experiment 2
I prefer the results when I reduce the hyphenation cost

```typst 
#set text(costs: (hyphenation: 30%))
#set par(justify: true, microtype: (max-expand: 0.02em, max-retract: 0.015em))
```

image

# Latex Source
```tex
\documentclass[10pt, letterpaper]{article}
\usepackage[utf8]{inputenc} % allow utf-8 input
\usepackage[T1]{fontenc}    % use 8-bit T1 fonts, see https://tex.stackexchange.com/questions/664/why-should-i-use-usepackaget1fontenc
\begin{document}
% Title Section
\title{LipSum}
\author{Professor Typst}
\date{\today}
\maketitle
%
\section{Lorem Ipsum}
\lipsum{}
\end{document}
```
# Typst Source

```typst
/// an attempt to replicate `\documentclass{article}[10pt, usletter]` 
/// TODO: I believe we can achieve this systematically by following https://www.latex-project.org/help/documentation/classes.pdf
/// however, Typst's uses a different definition of leading and its region layout is different than Tex's glue layout, so 
/// it might never be perfect, not to mention computer modern != new computer modern, the latter which being a bit thicker (I prefer it personally)  
#let latex-article(
  title: str,
  author: str,
  date: datetime.today().display("[month repr:long] [day], [year]"),
  eq-numbering: "(1)",
  doc,
) = {
  let title-page(title: "", author: "", date: none) = {
    set align(center)
    set block(spacing: 2em) // Space before title
    v(9.3em) // Space between title and author

    text(size: 16pt, weight: "thin")[#title]
    v(1.2em) // Space between title and author

    text(size: 12pt)[#author]
    v(.9em) // Space between author and date

    text(size: 12pt)[#date]
    v(1.2em) // Space after date before content
  }


  set text(size: 10pt, font: "New Computer Modern", lang: "en")
  show raw: set text(size: 10pt, font: "New Computer Modern Mono")
  // First, adjust your left margin to match LaTeX's actual calculation
  set page(
    paper: "us-letter",
    margin: (
      left: 47.25mm, // This is the exact LaTeX margin from the log
      right: 47.5mm, // Balanced margin
      top: 25.4mm,
      bottom: 38.1mm,
    ),
    numbering: "1",
  )
  // show heading.where(level: 1): set block(above: 1.4em, below: 1.15em)
  show heading: set block(above: 1.5em, below: 1.1em)
  // Then, update your heading style to match LaTeX's section indentation
  set heading(numbering: (..numbers) => {
    let number = numbers.pos().map(str).join(".")
    // No additional indentation before the number since we've set the correct margin
    [#number #h(0.8em)] // Just add the spacing between number and content
  })
  set par(
    leading: 0.55em,
    spacing: 0.55em,
    first-line-indent: 1.5em,
    justify: true,
  )
  title-page(title: title, date: date, author: author)
  set math.equation(numbering: eq-numbering)
  doc
}

```

Enivex:
For a more apples-to-apples comparison you would need to use fontspec with New Computer Modern

SolidTux:
I'm very positively surprised how simple this looks! Is it correct that this currently only involves tracking, but not stretching of glyphs?

Enivex:
> I'm very positively surprised how simple this looks! Is it correct that this currently only involves tracking, but not stretching of glyphs?

It's only tracking, yes

reknih:
On API design, let me present this dialog box of Adobe InDesign:

InDesign Paragraph Style options dialog in the Justification tab

In this dialog box, you can specify the minimum, maximum, and desired value for word and glyph spacing as well as glyph stretch. 100% for the first two is what the font wants. I think consolidating justification spacing parameters like this is a good idea. The desired values are set elsewhere, in `text.spacing` for word space and in `text.tracking`. Hence, I propose the following API:

```typ
#set par(
  justification-limits: (
    word: (min: 80% + 0pt, max: 133% + 0pt),
    glyph: (min: 98% + 0pt, max: 102% + 0pt),
  )
)
```

At least for the word spacing, the maximum is not absolute if the line is really crappy and not otherwise settable, it's just the point at which cost gets prohibitive (should be documented as such). All values are [Relatives](https://typst.app/docs/reference/layout/relative/) so that an absolute amount can be added. It is possible to specify just one key, and the dictionary will fold with the previous value.

For ligatures, I think that you should determine an em-distance at which the f and the dot of the i do not collide anymore in most fonts. Once you want to stretch beyond that, break them apart.

wrzian:
For the interface, we should probably allow `none` and `auto` for "turn it off" and "turn it on with default values." Then we can default to `none` for the moment, but eventually switch to `auto` when we're confident in having no bugs/oversights.

Also I think it would be smart to error if the user gives a value here when justification in general is turned off. I didn't see that explicitly on a quick glance of the code. (And then add a test for that case ofc)

laurmaedje:
> For the interface, we should probably allow none and auto for "turn it off" and "turn it on with default values." Then we can default to none for the moment, but eventually switch to auto when we're confident in having no bugs/oversights.

Wouldn't `none` essentially just be `glyph: (min: 100%, max: 100%)` and `auto` be `glyph: (min: x, max: y)` whatever default values `x` and `y` we choose?

> Also I think it would be smart to error if the user gives a value here when justification in general is turned off. I didn't see that explicitly on a quick glance of the code. (And then add a test for that case ofc)

That's a little tricky because at the point where the value is set we don't know whether justification will be enabled through another set rule. So it could only be done at the (many) points where we read these values during paragraph layout. That said, I'm also not sure whether it really makes sense. I think it can be sensible for values to be configured (just overriding the defaults) once and then for justification to be enabled selectively.

SolidTux:
> I think it can be sensible for values to be configured (just overriding the defaults) once and then for justification to be enabled selectively.

I totally agree. Even in a justified document, there will be some content which is not (e.g. headings, bibliographies, etc.). It would be annoying to always have to not just set `justify: false`, but also an additional setting. I think this can be solved by documenting that the values given here will be ignored if justification is disabled.

gdonval:
Would it make sense to have this merged and hidden behind some sort of experimental flag for the time being? The basic functionality seems to be here and mimicking InDesign's UX looks like a very good idea (not because it's Adobe but because it makes sense). 

The rest of the UX can be refined over time.

---

>  I think this can be solved by documenting that the values given here will be ignored if justification is disabled.

Is there actually a reason for `justify: false` not to use all this? As long as it's not overdone, anything that would allow to even out raggedness, even a tiny bit, would be welcome. That would probably involve a max raggedness variable to optimize over but in terms of UX, that means that just blanket-disable microtypography on unjustified text does not sound great. Things would have to be finer than that (with good defaults).

Hence my suggestion to have this merged if it makes sense (disabled by default) and figure things out from there. People can experiment with current "happy path" and other cases can be refined.

laurmaedje:
I'd like to see the proposed API (or a refined proposal) implemented before merge. It shouldn't be too much work to change it, so I don't really so the point in shipping the existing API behind a feature flag. If @diquah wants to continue working on this, that'd be great. Otherwise, someone else could pick this up or someone from the team might pick this up in the future.

diquah:
@laurmaedje I've implemented the new API, but I'm not sure how to make the auto/none keywords work. For the ligatures, the only thing I could figure out how to do was disable ligatures for the whole line when a glyph's extension exceeds a certain Em, but that seemed inefficient and suboptimal. Let me know how you would approach this or if you want to tackle it later down the road.

So far, it looks like this:

image

```typst
#table(
    stroke: none,
    gutter: 1em,
    columns: 3,
)[
    *Old Justification (Disable Glyph Justification)*
][
    *Default Justification*
][
    *Justification Using Purely Absolute Values*
][
    #set par(justify: true, justification-limits: (glyph: (min: 100% + 0pt, max: 100% + 0pt), word: (min: 80% + 0pt, max: 133% + 0pt)))
    #lorem(100)
][
    #set par(justify: true)
    #lorem(100)
][
    #set par(justify: true, justification-limits: (glyph: (min: 100% + 0.01pt, max: 100% + 0.05pt), word: (min: 100% + 0.5pt, max: 100% + 1.5pt)))
    #lorem(100)
]
```

MDLC01:
In this image, it looks like ligatures are always disabled. Libertinus Serif has a ligature for "Qu" which is not visible in any of the columns.

diquah:
@MDLC01 

> In this image, it looks like ligatures are always disabled. Libertinus Serif has a ligature for "Qu" which is not visible in any of the columns.

When ligatures are enabled, it looks like [this](https://github.com/typst/typst/pull/6161#issuecomment-2791779436). Ligatures are treated as single characters during spacing; That logic has not changed.

I'm unsure on how to implement individual ligatures breaking in a run.

MDLC01:
Oh, right. Your comment makes more sense without `#set text(ligatures: true)`. I don't think disabling ligatures on single runs is desirable, as that could technically affect the width of the glyphs significantly enough that line breaks would have to be in a different place.

Not splitting ligatures is probably a non-issue anyway as changes in glyph or word spacing should be minor enough that it shouldn't be noticeable.

ghost:
Here is the result for the latin text with `#set text(lang: "la")`, otherwise same code like above:

Screenshot 2025-08-01 182310

And here with `hyphenate: false`:

Screenshot 2025-08-01 182531


Does it create more patterns when hyphenation is off?

Enivex:
Do you mean the https://en.m.wikipedia.org/wiki/River_(typography) ?

That's down to coincidence I think. They aren't considered at all as far as I'm aware of.

gdonval:
> That's down to coincidence I think. They aren't considered at all as far as I'm aware of.

Coincidence is a big word: hyphenation helps reducing interword space variations, which, in turn, makes rivers much rarer or much less noticeable.

But it is also true that they aren't considered per se by any shaping algorithm.

Enivex:
> > That's down to coincidence I think. They aren't considered at all as far as I'm aware of.
> 
> Coincidence is a big word: hyphenation helps reducing interword space variations, which, in turn, makes rivers much rarer or much less noticeable.
> 
> But it is also true that they aren't considered per se by any shaping algorithm.

They're not talking about the difference between hyphenation and no hyphenation, but differences between two versions without hyphenation

ghost:
Obviously, this PR can improve the typesetting very much. Nonetheless, I want to add some things I noticed.

- Connected glyphs: when kerning is crucial, it can destroy the kerning very much: Screenshot 2025-08-02 045240
This is visible even at 1 % `glyph`-difference from 100 %, but less strong than in the example. For most fonts it will not be a problem though because most are not connected and most have non-optimal kerning.
- When the words in a text line do not change anymore and the whitespace and the word kerning both is in the allowed user-defined range, it will often still change the values to decrease the interword-spacing by changing the kerning to make the word widther. While this is also allowed by the user-definition, it is more dangerous to do for some fonts. This can be seen more easily with `hyphenate: false` and with a connected font like in the image. Screenshot 2025-08-02 052518

- The third option from InDesign "Desired" is missing here. Such a setting could solve the previous case. But it is probably more difficult to implement.
- I think it's difficult to find a setting to make the interword-spacing very small.
- Performance impact: my first impression is that the performance is very similiar like before, no clear difference.

diquah:
> Obviously, this PR can improve the typesetting very much. Nonetheless, I want to add some things I noticed...

@TS53 This is a great test case I didn't think of. Honestly, I think it might occur often enough, or be enough of a confusing bug to end-users that perhaps glyph-level justification should be disabled by default. I think it's disabled by default in InDesign and Affinity Publisher, actually.

Maybe we keep the "default" as it is but bring in @wrzian's suggestion and have a couple of "quick options" where the default option is `justification-limits: auto` where *that* auto is `(glyph: none, word: auto)`, and the `auto` in word is whatever default values are appropriate. Then, it could be easily set to `(glyph: auto, word: auto)` where, again, auto is an appropriate default value.

The point is that the end user can just easily turn on or off the features they want, with the option to explicitly specify exact options if they wish.

I think it is highly likely that somebody will post issues claiming text rendering is bugged, whereas in reality they just need to disable something that they didn't want, know, or care about. Those who appreciate this enhanced justification can easily enable it.

MDLC01:
If I understand correctly, those microtypohraphy features are only enabled when justification is enabled, right?

I would expect most fonts with connected glyphs to be cursive fonts, and additionally I would expect users to typically not justify text written with cursive fonts.

Enivex:
I don't see why it should be limited to justification only. Microtypography can still be used to make the ragged edge more aesthetically pleasing.

laurmaedje:
> This is a great test case I didn't think of. Honestly, I think it might occur often enough, or be enough of a confusing bug to end-users that perhaps glyph-level justification should be disabled by default. I think it's disabled by default in InDesign and Affinity Publisher, actually.

I am not at all sure, but considering how much better it looks [in the case present above](https://github.com/typst/typst/pull/6161#issuecomment-2795835591), it might be nice if that could work out of the box.

laurmaedje:
@khaledhosny Do you happen to know whether there is some font-level mechanism to determine whether it's permissible for inter-glyph spacing to be adjusted after shaping? Aside from cursive fonts, I'd think that it would probably break Arabic shaping, but for e.g. Latin text, it can help a lot with justification.

khaledhosny:
There is no font level data related to tracking/letter spacing (well, there is Apple’s [`trak`](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6trak.html) table, but very few fonts have it, and I’m not sure how it is supposed to be used).

However, CSS has [good documentation](https://drafts.csswg.org/css-text/#letter-spacing-property) on how letter spacing should be applied. The text on [justification](https://drafts.csswg.org/css-text/#text-justify-property) might be interesting as well.

In general, I’d advice against doing tracking for justification by default. Altering letter spacing can be perceived, even when in small amounts, especially when adjacent lines are letter-spaced differently.

khaledhosny:
> I would expect most fonts with connected glyphs to be cursive fonts, and additionally I would expect users to typically not justify text written with cursive fonts.

Why would the use of a cursive font require no justification? I don’t see the contradiction between the two.

khaledhosny:
> There is no font level data related to tracking/letter spacing

I once suggested [HarfBuzz provide a flag when two glyphs are cursively connected by a cursive anchor](https://github.com/harfbuzz/harfbuzz/issues/358), but we decided against it in the end because 1) most fonts for cursive writing systems (e.g. Arabic) connect glyphs without using cursive anchors 2) even cursive fonts for non-cursive scripts like Latin don’t always use cursive anchors either. So such flag would help only with a specific set of fonts and not others.

laurmaedje:
Thanks a lot for all the pointers!

Enivex:
> In general, I’d advice against doing tracking for justification by default. Altering letter spacing can be perceived, even when in small amounts, especially when adjacent lines are letter-spaced differently.

In my mind it's not really a question of whether the altered letter spacing can be perceived, but whether it's less distracting than the wildly varying word spacing that justification can sometimes produce.

I do agree that some microtypographical features need not necessarily be enabled by default though. At least not without _very_ conservative values.

sicikh:
Is there any chance to land this in 0.14?

We have used this branch to typeset a book (and to publish it — this feature helped us a lot to meet the size requirements!) successfully, but it would be nicer to be able to work on books in the web app.

I'm not rushing you by any means, but we just want to know if we should expect merging this in the near future. If we could help in any way, please let me know!

diquah:
Alright @tingerrr, This is everythings I planned to go into this PR. More enhancements to justification will come in later PRs by either me or some other enthusiast.

The changes are as follows:
- A new `justification-limits` property is now available for `par`
- By default it is set to `none`, which is interpreted as the default behavior for paragraph spacing before this PR was introduced. This makes this PR backwards compatible.
- When set to `auto`, it uses sensible defaults, similar to Adobe InDesign or Affinity Publisher
- It can also be set explicity supporting the full range of options as described in [this comment](https://github.com/typst/typst/pull/6161#issuecomment-2959337354).

Example:
```typst
#table(
    stroke: none,
    gutter: 1em,
    columns: 3,
)[
    *Default Justification (Pre-PR Behavior)*
    // Equivalent to Old Justification Rules, to show they are the same.
][
    *Justification With Auto Justification Limits*
][
    *Justification With Custom Justification Limits*
][
    #set par(justify: true)
    #set par(justify: true, justification-limits: none) // equivalent statement
    #lorem(100)
][
    #set par(justify: true, justification-limits: auto)
    #lorem(100)
][
    #set par(justify: true, justification-limits: (glyph: (min: 99% + 0pt, max: 100% + 0pt), word: (min: 80% + 0pt, max: 133% + 0pt)))
    #lorem(100)
]
```

image

P.S.
Thank you for the kind words @sicikh, as I have less and less time to work on personal projects I'm hoping to hammer this PR out with the available features and hopefully it can be expanded later on to support ligature breaking, font stretching, and all that goodness.

crates/typst-library/src/model/par.rs
```diff
@@ -321,6 +324,57 @@ impl ParElem {
...
+}
+
+cast! {
+    JustificationLimits,
+    self => Value::Dict(self.into()),
+    mut dict: Dict => {
+        let mut word: Dict = dict.take("word")?.cast()?;
+        let word_min = word.take("min")?.cast()?;
+        let word_max = word.take("max")?.cast()?;
+
```
sicikh, line 485:
```suggestion
word.finish(&["min", "max"])?;

```

crates/typst-library/src/model/par.rs
```diff
@@ -321,6 +324,57 @@ impl ParElem {
...
+        let mut glyph: Dict = dict.take("glyph")?.cast()?;
+        let glyph_min = glyph.take("min")?.cast()?;
+        let glyph_max = glyph.take("max")?.cast()?;
+
```
sicikh, line 494:
```suggestion
glyph.finish(&["min", "max"])?;

```

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -226,13 +227,26 @@ impl ShapedGlyph {
...
+        &self,
+        style: CjkPunctStyle,
+        justification_limits: JustificationLimits,
+        font_size: Abs,
+    ) -> Adjustability {
         let width = self.x_advance;
+
         if self.is_space() {
             Adjustability {
-                // The number for spaces is from Knuth-Plass' paper
```
sicikh, line 233: Maybe this comment should not be deleted? As far as I understand, this comment explains the values 2.0 and 3.0 below.

sicikh (commented):
Add `finish` checks to `word` and `glyph` subdictionaries.

crates/typst-layout/src/inline/shaping.rs
diquah: This was removed in error, I have re-added and rephrased it to be more clear as well.

crates/typst-library/src/model/par.rs
```diff
@@ -321,6 +324,59 @@ impl ParElem {
...
+}
+
+cast! {
+    JustificationLimits,
+    self => Value::Dict(self.into()),
+    mut dict: Dict => {
+        let mut word: Dict = dict.take("word")?.cast()?;
+        let word_min = word.take("min")?.cast()?;
+        let word_max = word.take("max")?.cast()?;
+word.finish(&["min", "max"])?;
```
sicikh:
```suggestion
        word.finish(&["min", "max"])?;
```

oopsie, forgot to include leading spaces. How did it pass formatting checks though 🤔 same thing below

crates/typst-library/src/model/par.rs
```diff
@@ -321,6 +324,59 @@ impl ParElem {
...
+
+        let mut glyph: Dict = dict.take("glyph")?.cast()?;
+        let glyph_min = glyph.take("min")?.cast()?;
+        let glyph_max = glyph.take("max")?.cast()?;
+glyph.finish(&["min", "max"])?;
```
sicikh:
```suggestion
        glyph.finish(&["min", "max"])?;
```

same thing as above with `word.finish`

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -226,13 +227,27 @@ impl ShapedGlyph {
...
+                stretchability: (
+                    Em::zero(),
+                    (width / 2.0) * justification_limits.word_max.rel.get()
+                        + Em::from_length(justification_limits.word_max.abs, font_size),
+                ),
+                shrinkability: (
+                    Em::zero(),
+                    (width / 3.0) * justification_limits.word_min.rel.get()
+                        + Em::from_length(justification_limits.word_min.abs, font_size),
+                ),
```
laurmaedje, line 266: It seems odd to me that the calculation works differently for the space. Wouldn't it make more sense for the default value of `justification-limits` to be adjusted instead? Basically making this the default value:
```typ
#(
  word: (
    min: 100% * (2 / 3),
    max: 150%,
  ),
  glyph: (
    min: 100%,
    max: 100%,
  )
)
```
And then we'd remove all of the `none` and `auto` things and just add suggested values for `glyph` to the documentation. 

I'd also (as suggested by someone on Discord) add support for _folding_ such that `#set par(justification-limits: (glyph: (min: 98%, max: 102%))` does not override the existing values for `word` and `#set par(justification-limits: (glyph: (min: 95%))` not the existing value for `glyph.max`. Dealing with folding is always a bit tricky though, so I could handle that myself.

ghost:
I think it would be good to have checks for `max >= min`.

laurmaedje:
Just as a heads-up: We're working on this internally now, so please don't push further commits. There are some layout issues that still need figuring out, e.g. end-of-line glyphs being incorrectly adjusted.

crates/typst-layout/src/inline/shaping.rs
diquah: My train of thought was that the justification-limits are relative to the internal layout algorithm. From a UX perspective, I feel that users should not need to know any magic numbers in order to modify the layout algorithm.

I imagine the most common mental model for this feature would be that 100% means 100% of the spacing determined by the internal layout engine. 200% min word limit would mean that the spaces should be able to compress to double what they're at before modifying this value.

Otherwise, as you suggested, we'd just have the default values written in the documentation.

However, the flip side of this coin is that if 200% meant 200% of the engine's default spacing rule, if someone wanted to explicitly say 200% of an _actual_ non-justified space, they'd have to know about the internal 2/3 and then write the true 100% as 3/2 multiplied by their initially desired value.

This could also be written into the documentation, but I just think that the first case that I mentioned would happen more often than the second? But the second one is more confusing to the end user if that is their desired behavior.

It's up to you, I just wanted to outline my thinking.

laurmaedje:
I have now finalized the PR. There are some crucial changes to the implementation and the API:

- Stretchability of glyphs is not relative to the glyph width anymore. For spaces, it makes sense to say "you can stretch this down to 70% of the size", but for glyphs it is problematic because glyphs have varying widths so the stretch would also vary, leading to uneven spacing. In the new design, the extra gap that can be inserted between glyphs is the same between each pair of glyphs. This is in line with how the CSS `text-justify: inter-character;` property behaves.

- Instead of `word` and `glyph`, the dictionary now takes the keys `spacing` and `tracking`. These mirror exactly the `text.spacing` and `text.tracking` properties and their types have been changed accordingly. The behavior of the metrics is adjusted to be consistent with `spacing` and `tracking`. For spacing (as it is relative), `min` and `max` define the absolute minimum and maximum width _to_ which spaces may be adjusted. For tracking (as it is additive), `min` and `max` define the amount _by_ which they may be adjusted. Notably, the non-relative part of `min` must always be nonpositive, while the non-relative part of `max` must always be nonnegative (this is checked). Moreover, for spaces, the `tracking` value acts _in addition_ to the `spacing` value (just like `text.tracking` does in addition to `text.spacing`).

- Glyphs at the end of the line are not adjusted anymore. This was a bit hard to spot under normal circumstances, but excessive character-level justification would reveal uneven line endings.

- For consecutive glyphs belonging to the same shaping cluster, only the last one receives stretchability. This can avoid issues with complex scripts, accents, and such.

- The property is now foldable, meaning that `tracking` and `spacing` can be individually specified and the other will keep its previous value. The `min` and `max` values, in contrast, must always be both specified.

- Excessive shrinking (below 25% of the glyph width) is clamped as zero-sized or negative glyph widths cause paragraph layout to take very long or not terminate at all.

- Added extensive documentation with full clarification on how the values behave.

- Added tests for API and layout

SolidTux:
That sounds great!

> Excessive shrinking (below 25% of the glyph width) is clamped as zero-sized or negative glyph widths cause paragraph layout to take very long or not terminate at all.

Is there a reason why this is clamped instead of throwing an error? I'd rather have the compilation fail than my input being silently treated differently when I specify and invalid value.

laurmaedje:
> Is there a reason why this is clamped instead of throwing an error? I'd rather have the compilation fail than my input being silently treated differently when I specify and invalid value.

I can't check this fully at the site where you define the value as it is font size and glyph size dependant, so the error would be very non-local and it would fail at some random piece of text.

ghost:
I compiled the new version, I think it works good and the spacing is a little bit simpler to adjust than before (`min: 30 %` starts to make it work). Smaller values than 25 % had no effect to spacing, not sure if these values were limited too or if it just did converge.

- The following error should get improved, at least when both parameters `spacing` and `tracking` are used.
  >error: `min` value is invalid (expected length, found float)

  Instead, maybe use:
  >error: `min` value of `tracking` is invalid (expected length, found float)

- My previous [comment](https://github.com/typst/typst/pull/6161#issuecomment-3362225883) is fixed now by the positive/negative value checks.

- The `desired` value as discussed before, is missing and not automatically targeted to the default value. This means that the `tracking` is often still applied from the user defined range even though it could be the default (connected fonts will become slightly unconnected even though the words per line would not change). This was the same before, and is probably difficult to change.

To sum up: everything works good, but the error should get improved as shown above.

laurmaedje:
> The following error should get improved, at least when both parameters spacing and tracking are used.

Yeah, I noticed this, but didn't really feel like fixing it for this PR specifically because it is a problem across the whole of Typst with such kinds of dictionaries and should really be fixed holistically. However, it's true that it's especially bad in this case, so I implemented a workaround now.

laurmaedje:
Thanks @diquah and also everyone else who helped move this forward!

PR #6163, opened 2025-04-10
Use the right multiplication symbol in expression tooltip

MDLC01:
When hovering over an expression, Typst can show a tooltip displaying its value. When the expression takes multiple values at runtime (such as within a loop, or within a function called multiple times), the first ten are displayed. In case multiple consecutive values are equal, they are displayed once with a "(×<i>n</i>)" mention.

Prior to this PR, the multiplication symbol was actually not a multiplication sign, but a lower case X. In addition to the actual multiplication sign looking better, it may also improve accessibility because × carries semantics that may be used by screen readers (I haven't tested, though).


MDLC01:
Also, maybe `...` should be replaced with `…` below.

MDLC01:
`…` is not used anywhere currently, so I will keep it that way for now.

laurmaedje:
Thanks!

PR #6166, opened 2025-04-10
Fix panic when sampling across two coincident gradient stops

### LINKED ISSUE #6162: Panic when two gradient stops have offset `0%` with PNG export
### Description

```typ
#rect(
  fill: gradient.linear(
    (red, 0%),
    (green, 0%),
    (blue, 100%),
  )
)
```

panics with

```
thread 'main' panicked at crates\typst-render\src\paint.rs:260:10:
components must always be in the range [0..=1]
```

when exporting to PNG. The panic does not occur when exporting to SVG or PDF.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

bluebear94:
Fixes #6162.


crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1290,9 +1290,21 @@ fn sample_stops(stops: &[(Color, Ratio)], mixing_space: ColorSpace, t: f64) -> C
     if low == 0 {
         low = 1;
     }
+    if t == 1.0 {
+        while stops[low - 1].1 >= stops[low].1 {
```
laurmaedje: This seems like it could underflow. Overall, I'm not sure I fully understand the code. Performing an exact floating point comparison seems potentially problematic.

Unrelated to your PR, the binary search above looks like a standard binary search that could be replaced by the std version.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1284,25 +1284,29 @@ fn process_stops(stops: &[Spanned<GradientStop>]) -> SourceResult<Vec<(Color, Ra
...
     }
-
-    if low == 0 {
-        low = 1;
+    if j == stops.len() {
+        while j > 0 && stops[j].1.is_one() {
+            j -= 1;
+        }
+        return stops[j].0;
     }
```
saecki: Since the partition checks for `< t` and the last point is required to be `1` this if statement seems to be redundant, because `1 < 1` will always return `false`.
The loop is also redundant, because we will always land at the lower end of a chunk with equal ratios. And it would panic because it would index the array at `stops.len()` when the condition is initially checked.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1284,25 +1284,29 @@ fn process_stops(stops: &[Spanned<GradientStop>]) -> SourceResult<Vec<(Color, Ra
 
 /// Sample the stops at a given position.
 fn sample_stops(stops: &[(Color, Ratio)], mixing_space: ColorSpace, t: f64) -> Color {
+    assert!(!stops.is_empty(), "Cannot sample from gradient with no stops");
```
saecki: While somewhat sensible, I don't think we need to assert here, since this condition is alread checked when constructing the gradients, and the code that would panic is in this function anyway.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1284,25 +1284,29 @@ fn process_stops(stops: &[Spanned<GradientStop>]) -> SourceResult<Vec<(Color, Ra
...
+        return stops[j].0;
     }
 
-    let (col_0, pos_0) = stops[low - 1];
-    let (col_1, pos_1) = stops[low];
+    let (col_0, pos_0) = stops[j - 1];
+    let (col_1, pos_1) = stops[j];
+    debug_assert!(pos_0 < pos_1);
+    debug_assert!(pos_0.get() <= t);
+    debug_assert!(t <= pos_1.get());
```
saecki: Could you remove these, I also think they are a bit much.

saecki (commented):
Thanks for the PR! There are a couple of things that I would like to change.

PR #6168, opened 2025-04-11
Table multiple headers and subheaders

PgBiel:
Solves part of #5375 and https://github.com/typst/typst/issues/5377

- Table headers can now be placed **at any row, not only the first** (then, they will repeat starting on the next page).
- A table can now have **more than one header**.
    - Each will cause the previous one to stop repeating, and the new one starts repeating on the next page instead.
- A table can now have **subheaders**: headers with **levels,** like headings.
    - When headers with different levels are placed, they repeat simultaneously.
    - Levels form a hierarchy. When a level 2 subheader is placed, the previous level 2, as well as level 3+, subheaders stop repeating altogether, and the new level 2 subheader starts repeating instead, as if it began a new "table section".

As a bonus, headers with `repeat: false` now have orphan prevention too! They won't appear alone at the bottom of a page anymore.

Some preliminary work which will enable multiple footers and subfooters is also included, but the actual implementation of those is left to a follow-up PR.

# Example

<details><summary>Source code</summary>
<p>

```typ
#set page(height: 15.6em, width: 120pt, margin: 10pt)
#table(
  columns: 2,
  align: center,
  table.header(
    table.cell(colspan: 2)[*Regional User Data*],
  ),
  table.header(
    level: 2,
    table.cell(colspan: 2)[*Germany*],
    [*Username*], [*Joined*]
  ),
  [john123], [2024],
  [rob8], [2025],
  [joe1], [2025],
  [joe2], [2025],
  [martha], [2025],
  [pear], [2025],
  table.header(
    level: 2,
    table.cell(colspan: 2)[*United States*],
    [*Username*], [*Joined*]
  ),
  [cool4], [2023],
  [roger], [2023],
  [bigfan55], [2022]
)
```

</p>
</details> 

Output:

image

# Implementation details

- (under construction, check back soon...)

# TODO

- [ ] Refactor new grid layouter fields (group them up).
- [ ] Test gutters and some more orphan prevention edge cases
- [ ] Test interaction with footers
- [ ] Remove TODO comments where applicable
- [ ] General cleanup where applicable


crates/typst-library/src/model/table.rs
```diff
@@ -492,6 +537,14 @@ pub struct TableHeader {
     #[default(true)]
     pub repeat: bool,
 
+    /// The level of the header. Must not be zero.
+    ///
+    /// This is used during repetition multiple headers at once. When a header
```
laurmaedje: grammar is a bit off here (same for grid)

tests/suite/layout/grid/subheaders.typ
```diff
@@ -0,0 +1,1002 @@
...
+    [b]
+  ),
+)
+
+--- grid-subheaders-alone-no-orphan-prevention ---
+#set page(height: 5.3em)
+#v(2em)
+#grid(
+  grid.header(
+  // (
```
laurmaedje: what's up with these?

tests/suite/layout/grid/subheaders.typ
```diff
@@ -0,0 +1,1002 @@
...
+#grid(
+  grid.header(
+    [a]
+  ),
+  grid.header(
+    level: 2,
+    [b]
+  ),
+  [c]
+)
```
laurmaedje: I would prefer a more compact formatting for all of the tests:
```typ
#grid(
  grid.header[a],
  grid.header(level: 2)[b],
  [c]
)
```

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -426,37 +428,96 @@ pub struct Line {
 /// A repeatable grid header. Starts at the first row.
 #[derive(Debug)]
 pub struct Header {
+    /// The first row included in this header.
+    pub start: usize,
     /// The index after the last row included in this header.
     pub end: usize,
```
laurmaedje: this could also be a `range: Range<usize>`. what do you think?

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -426,37 +428,96 @@ pub struct Line {
...
     NotRepeated(T),
 }
 
+impl<T> Deref for Repeatable<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        self.unwrap()
+    }
+}
```
laurmaedje: I wonder whether this representation has practical benefits over
```rust
pub struct Repeatable<T> {
    value: T,
    repeats: bool,
}
```
or something similar

crates/typst-library/src/layout/grid/resolve.rs
```diff
@@ -2112,11 +2305,12 @@ fn resolve_cell_position(
 /// have cells specified by the user) as well as any headers and footers.
 #[inline]
 fn find_next_available_position<const SKIP_ROWS: bool>(
```
laurmaedje: why is skip_rows a const parameter instead of a normal parameter?

tests/suite/layout/grid/subheaders.typ
PgBiel: Oh, that's an artifact from my tests of removing the header (moving the cell out of it) and seeing what happens. I'll remove them. :slightly_smiling_face:

tests/suite/layout/grid/subheaders.typ
PgBiel: I guess this happened because most tests were copy-pastes from other tests haha. I'll change them up.

crates/typst-library/src/layout/grid/resolve.rs
PgBiel: In principle the main goal of the enum was to force the repetition to be handled, but at some point it just became a drag and started polluting code as I had to handle generic headers, so adding `deref` was a quick fix for that...

Though I like how the enum looks like in a match (which is used a few times in the codebase):

```rs
match header {
  Repeated(header) => ...
  NotRepeated(header) => ...
}
```

Though I suppose we can experiment with changing the representation and seeing how it goes. I guess there's also the advantage that we'd be able to add more fields here, e.g. `short_lived`.

crates/typst-library/src/layout/grid/resolve.rs
PgBiel: Just to ensure the skip_rows check is compiled away since this is a hot path for auto cells.

crates/typst-library/src/model/table.rs
PgBiel: Whoops.

I think the docs need some improvement, maybe with some examples. But we could also look into making it more complete later.

crates/typst-library/src/layout/grid/resolve.rs
PgBiel: Though I suppose the `#[inline]` could already be doing this job (I suppose the parameter would be replaced with `if true` after inlining)... might want to look into removing this then. I should check the compiled output to make sure.

crates/typst-library/src/layout/grid/resolve.rs
PgBiel: I thought that at first, but I didn't change it because I'd use .start and .end directly multiple times. But I think using the range should be encouraged rather than discouraged, so your suggestion might be a better idea.

crates/typst-library/src/layout/grid/resolve.rs
laurmaedje: Whatever you think is best

PgBiel:
I didn't add an example to docs as I don't remember if multi-page examples are supported, or would look good :thinking:

One alternative is to show a single-page example anyway, but indicating the header levels - better than nothing...

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -468,6 +468,17 @@ pub struct GridHeader {
     #[default(true)]
     pub repeat: bool,
 
+    /// The level of the header. Must not be zero.
+    ///
+    /// This allows repeating multiple headers at once. Headers with different
+    /// levels can repeat together, as long as they have ascending levels.
+    ///
+    /// Notably, when a level with a lower level starts repeating, all higher
```
laurmaedje:
```suggestion
    /// Notably, when a header with a lower level starts repeating, all higher
```

crates/typst-library/src/model/table.rs
```diff
@@ -539,9 +539,12 @@ pub struct TableHeader {
 
     /// The level of the header. Must not be zero.
     ///
-    /// This is used during repetition multiple headers at once. When a header
-    /// with a lower level starts repeating, all headers with a lower level stop
-    /// repeating.
+    /// This allows repeating multiple headers at once. Headers with different
+    /// levels can repeat together, as long as they have ascending levels.
+    ///
+    /// Notably, when a level with a lower level starts repeating, all higher
```
laurmaedje: same here

PgBiel:
> Apart from the two typos, this is ready to merge now, right?

I was going to answer this just now, but it seems that you've read my mind already :smile:

laurmaedje:
Totally forgot I asked :D

mewmew:
Really happy to see the addition of subheaders in tables! Thank you @PgBiel <3

For a paper I'm writing, it helped resolve the position of subheaders to the correct page.column, so it wouldn't be "alone" without any associated data cells on the same page.

## Before PR

As a reference, this was before the PR:

page 1 (notice subheader "alone" on page without any associated data cells)
screenshot_2025-08-07_18:41:00

page 2
screenshot_2025-08-07_18:41:11

## After PR

And this is after the PR:

page 1
screenshot_2025-08-07_18:41:31
page 2
screenshot_2025-08-07_18:41:41

PR #6169, opened 2025-04-12
Removing unused warnings in nightly

patrik64:
(appears only when using nightly) 
 
two minor changes that will remove 

``warning: unused `ControlFlow` that must be used`` 

in ``crates/typst-library/src/foundations/content.rs``.

Note: I am not sure if this is the right way to go about contributing (or if the fix is to your liking).
I hope you don't mind ...


laurmaedje:
Thanks!

PR #6171, opened 2025-04-12
Typst treats each citation supplement as equal — fix it

wkordalski:


PgBiel:
Thanks! In order to merge this, we'll need some unit tests. Do you have some sample styles we can use to test? You can make a test and paste a simple style inline (with `bibliography(bytes(style))`), or send it here and I'll do it for you. Thanks!

wkordalski:
I've added a test for supplement equality (where supplement is text-only). Do we need more?

We could also add tests with more complex supplements, but I suppose that content equality is not specified by the language and we should not assume anything about it. Maybe we should even compare `plain_text()` of `Content` instead of whole `Content`?

I've started with https://github.com/citation-style-language/styles/blob/master/polish-legal.csl
and removed as much as possible.

```xml
<?xml version="1.0" encoding="utf-8"?>
<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" default-locale="pl-PL">
  <info>
    <title>Example citation style</title>
    <id>http://www.example.com/</id>
  </info>
  <macro name="locator">
    <group delimiter=" ">
      <label variable="locator" form="short"/>
      <text variable="locator"/>
    </group>
  </macro>
  
  <citation>
    <sort>
      <key variable="title"/>
      <key variable="issued"/>
    </sort>
    <layout>
      <choose>
        <if position="first">
          <group delimiter=", ">
            <text variable="title"/>
            <text macro="locator"/>
          </group>
        </if>
        <else-if position="ibid-with-locator">
          <group delimiter=", ">
            <text term="ibid"/>
            <text macro="locator"/>
          </group>
        </else-if>
        <else-if position="ibid">
          <text term="ibid"/>
        </else-if>
        <else-if position="subsequent">
          <group delimiter=", ">
            <text variable="title"/>
            <text macro="locator"/>
          </group>
        </else-if>
      </choose>
    </layout>
  </citation>

  <bibliography>
    <sort>
      <key variable="title"/>
    </sort>
    <layout>
      <text variable="title"/>
    </layout>
  </bibliography>
</style>
```

Any style that handles `ibid` and `ibid-with-locator` separately would work.

tests/suite/model/cite.typ
```diff
@@ -171,3 +171,72 @@ aaa
...
++ "  </citation>"
++ "  <bibliography>"
++ "    <sort>"
++ "      <key variable=\"title\"/>"
++ "    </sort>"
++ "    <layout>"
++ "      <text variable=\"title\"/>"
++ "    </layout>"
++ "  </bibliography>"
++ "</style>"
```
PgBiel: Did you manually split these? You can use code blocks to save time and sanity haha

E.g.
````typ
#let x = bytes(
  ```yaml
  hello:
     world
  new:
     lines
  ```.text
)
````

Comment (PgBiel) [reply tests/suite/model/cite.typ]: Feel free to apply this suggestion directly:
```suggestion
  ```xml
  <?xml version=\"1.0\" encoding=\"utf-8\"?>"
  <style xmlns=\"http://purl.org/net/xbiblio/csl\" version=\"1.0\" class=\"note\" default-locale=\"pl-PL\">
    <info>
      <title>Example citation style</title>
      <id>http://www.example.com/</id>
    </info>
    <macro name=\"locator\">
      <group delimiter=\" \">
        <label variable=\"locator\" form=\"short\"/>
        <text variable=\"locator\"/>
      </group>
    </macro>
  
    <citation>
      <sort>
        <key variable=\"title\"/>
        <key variable=\"issued\"/>
      </sort>
      <layout>
        <choose>
          <if position=\"first\">
            <group delimiter=\", \">
              <text variable=\"title\"/>
              <text macro=\"locator\"/>
            </group>
          </if>
          <else-if position=\"ibid-with-locator\">
            <group delimiter=\", \">
              <text term=\"ibid\"/>
              <text macro=\"locator\"/>
            </group>
          </else-if>
          <else-if position=\"ibid\">
            <text term=\"ibid\"/>
          </else-if>
          <else-if position=\"subsequent\">
            <group delimiter=\", \">
              <text variable=\"title\"/>
              <text macro=\"locator\"/>
            </group>
          </else-if>
        </choose>
      </layout>
    </citation>
    <bibliography>
      <sort>
        <key variable=\"title\"/>
      </sort>
      <layout>
        <text variable=\"title\"/>
      </layout>
    </bibliography>
  </style>```.text
```

tests/suite/model/cite.typ
```diff
@@ -171,3 +171,72 @@ aaa
...
+--- cite-supplements-and-ibid ---
+#set page(width: 300pt)
+
+Par 1 @arrgh
+
+Par 2 @arrgh[p. 5-8]
+
+Par 3 @arrgh[p. 5-8]
+
+Par 4 @arrgh[p. 9-10]
```
PgBiel, line 184: Let's add a quick showcase of how it behaves in contrast with the plain_text alternative: the strong should be considered a different supplement.

FWIW, I think this is the best approach for the time being. We can consider changing it in the future if necessary.

```suggestion

Par 1 @arrgh

Par 2 @arrgh[p. 5-8]

Par 3 @arrgh[p. 5-8]

Par 4 @arrgh[p. 9-10]

Par 5 @arrgh[**p. 9-10**]
```

tests/suite/model/cite.typ
wkordalski: > Did you manually split these?

No. With Python. :D

Nice trick!

Cargo.toml
```diff
@@ -61,7 +61,7 @@ fontdb = { version = "0.23", default-features = false }
 fs_extra = "1.3"
 rustc-hash = "2.1"
 glidesort = "0.1.2"
-hayagriva = { git = "https://github.com/typst/hayagriva", rev = "03ae2acf" }
+hayagriva = { git = "https://github.com/typst/hayagriva", rev = "d151bb5" }
```
PgBiel: I think we can use the new version already; I'll submit another PR later with the last few changes.

```suggestion
hayagriva = "0.9.1"
```

PgBiel (approved):
Looks good to me, thanks!

wkordalski:
After updating to Hayagriva 0.9.1 one test fails.
@PgBiel, could you verify whether we should update ref image for that test or some bug was introduced?

Ref:
image

Live:
image

`(n.d.)` was added in the last line.

PgBiel:
Probably a consequence of https://github.com/typst/hayagriva/pull/400, seems expected in principle. I'd update the test.

laurmaedje:
Thanks!

Tip: Instead of copying reference images from the output directory manually, you can use `cargo testit --update` and then it will automatically optimize the images for smaller sizes. This also made the `large` annotation unnecessary.

PR #6190, opened 2025-04-17
Allow TYPST_PACKAGE_PATH to have a list of paths to search.

RossSmyth:
This allows for better modularity with wrappers and such.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -332,8 +332,8 @@ pub struct ProcessArgs {
 #[derive(Debug, Clone, Args)]
 pub struct PackageArgs {
     /// Custom path to local packages, defaults to system-dependent location.
-    #[clap(long = "package-path", env = "TYPST_PACKAGE_PATH", value_name = "DIR")]
-    pub package_path: Option<PathBuf>,
+    #[clap(long = "package-path", env = "TYPST_PACKAGE_PATH", value_name = "DIR", value_delimiter = ENV_PATH_SEP)]
+    pub package_path: Option<Vec<PathBuf>>,
 
     /// Custom path to package cache, defaults to system-dependent location.
     #[clap(
--- a/crates/typst-kit/src/package.rs
+++ b/crates/typst-kit/src/package.rs
@@ -28,7 +28,7 @@ pub struct PackageStorage {
     /// The path at which non-local packages should be stored when downloaded.
     package_cache_path: Option<PathBuf>,
     /// The path at which local packages are stored.
-    package_path: Option<PathBuf>,
+    package_path: Option<Vec<PathBuf>>,
     /// The downloader used for fetching the index and packages.
     downloader: Downloader,
     /// The cached index of the default namespace.
@@ -40,7 +40,7 @@ impl PackageStorage {
     /// the recommended XDG directories if they are `None`.
     pub fn new(
         package_cache_path: Option<PathBuf>,
-        package_path: Option<PathBuf>,
+        package_path: Option<Vec<PathBuf>>,
         downloader: Downloader,
     ) -> Self {
         Self::with_index(package_cache_path, package_path, downloader, OnceCell::new())
@@ -51,7 +51,7 @@ impl PackageStorage {
     /// Useful for testing.
     fn with_index(
         package_cache_path: Option<PathBuf>,
-        package_path: Option<PathBuf>,
+        package_path: Option<Vec<PathBuf>>,
         downloader: Downloader,
         index: OnceCell<Vec<serde_json::Value>>,
     ) -> Self {
@@ -60,7 +60,8 @@ impl PackageStorage {
                 dirs::cache_dir().map(|cache_dir| cache_dir.join(DEFAULT_PACKAGES_SUBDIR))
             }),
             package_path: package_path.or_else(|| {
-                dirs::data_dir().map(|data_dir| data_dir.join(DEFAULT_PACKAGES_SUBDIR))
+                dirs::data_dir()
+                    .map(|data_dir| vec![data_dir.join(DEFAULT_PACKAGES_SUBDIR)])
             }),
             downloader,
             index,
@@ -74,7 +75,7 @@ impl PackageStorage {
     }
 
     /// Returns the path at which local packages are stored.
-    pub fn package_path(&self) -> Option<&Path> {
+    pub fn package_path(&self) -> Option<&[PathBuf]> {
         self.package_path.as_deref()
     }
 
@@ -87,8 +88,8 @@ impl PackageStorage {
     ) -> PackageResult<PathBuf> {
         let subdir = format!("{}/{}/{}", spec.namespace, spec.name, spec.version);
 
-        if let Some(packages_dir) = &self.package_path {
-            let dir = packages_dir.join(&subdir);
+        for to_search in self.package_path.iter().flatten() {
+            let dir = to_search.join(&subdir);
             if dir.exists() {
                 return Ok(dir);
             }
@@ -134,6 +135,7 @@ impl PackageStorage {
             let subdir = format!("{}/{}", spec.namespace, spec.name);
             self.package_path
                 .iter()
+                .flatten()
                 .flat_map(|dir| std::fs::read_dir(dir.join(&subdir)).ok())
                 .flatten()
                 .filter_map(|entry| entry.ok())


PgBiel:
Prior discussion: https://github.com/typst/typst/pull/4227#issuecomment-2140218562

I wonder if it would make more sense to retain the original idea of being able to map one namespace to one path. Would love to hear more about potential use cases, and how that idea would interact with them.

RossSmyth:
My motivation is my library for building Typst documents with packages from arbitrary git repos. 
https://github.com/RossSmyth/press

Then I wouldn't have to move all the packages into the same directory, or symlink all of them together. Just create a search path.

Myriad-Dreamin:
> Then I wouldn't have to move all the packages into the same directory, or symlink all of them together. Just create a search path.

The previous approach should win imo. I list some points from my quick thinking:
- it becomes unclear to quickly check the exact package code is used if they occur in multiple package directories.
- I view special tools will be encouraged to have their owned package directories, then users have to maintain the list and priorities to compile documents successfully.
- Multiple tools, especially tools that are not written in rust (which can't use `typst-kit` then), will have to add support to search through multiple pacakge directories. It is harder to watch changes about multiple package directories comparing with one.
- If we have a list, updating "local packages" atomically should lock many directories, which sounds not easy. Locking the layers (directories) above the updating layer), and removing the one in some layer, and adding one to another layer.
- but we only get little benefit: if I don't want to change existing package directories, just create another package directory placed over some package directories. then, we may also get a increasing number of package directories in the same computer.

I'm wondering how compiler devs in other languages think of it.

RossSmyth:
Your English is a bit hard to understand, but I will attempt to answer.

> The former approach should win imo.

That is what it does now. But search paths in other compilers are quite common.

> it becomes unclear to quickly check the exact package code is used if they occur in multiple package directories.

As is standard in tools with search paths, first wins.

> I view special tools will be encouraged to have their owned package directories, then users have to maintain the list and priorities to compile documents successfully.

I don't see how this is a problem. Someone has to put the files there. Either some tool that manages it for the user, so the users doesn't care, or the user themselves. And if it is a tool, they will most likely place it in 

> If we have a list, updating "local packages" atomically should lock many directories, which sounds not easy. Locking the layers (directories) above the updating layer, and removing the one in some layer, and adding one to another layer.

I'm unsure what locking has to do with anything. When would locking have to occur? 

> but we only get little benefit: if I don't want to change existing package directories, just create another package directory placed over some package directories. then, we may also get a increasing number of package directories in the same computer.

I'm sorry but I don't really understand.

Myriad-Dreamin:
If you feel unfriendly to read my points, please forgive me. I'll accept whether this PR will be merged or not.

> > The former approach should win imo.
> 
> That is what it does now. But search paths in other compilers are quite common.

I meant you mentioned two approaches to configure packages:
1. move all the packages into the same directory, or symlink all of them together. 
2. or create a search path.

Then approach 1 should win and approach 2 should lose, in my opinion.

> 
> > it becomes unclear to quickly check the exact package code is used if they occur in multiple package directories.
> 
> As is standard in tools with search paths, first wins.

I meant for a list of paths `TYPST_PACKAGE_PATH=tool-a:tool-b:org-a:org-b:data-dir:cache-dir`, we should list these directories to know whether directory a package came from, `tool-a`, `org-a`, `data-dir`, or `cache-dir`. There is also a question if I'm not knowing `TYPST_PACKAGE_PATH` and I even don't have a knowledge about what is a `TYPST_PACKAGE_PATH`.

This is different than currrent situation, where we only have two directories to check. If I know that it is not from `data-dir`, it must be in `cache-dir`. I had met an error that I forgot to remove some packages in the `data-dir` and the package in the `cache-dir` was not used, then I learnt it, so I personally feel worse if the `TYPST_PACKAGE_PATH` becomes a list.

I believe smart tools should prevent stupid errors caused by people early. For a counterexample in other language, the python's `sys.path` bothers me a lot, may also bothers many other people. This is powerful, but also bug-prone. In my opinion, typst compiler itself should fully define which exact directories to load, their semantics, and the priority to use. If they do it better, the things must be also document in the official tutorials and sites.

> > I view special tools will be encouraged to have their owned package directories, then users have to maintain the list and priorities to compile documents successfully.
> 
> I don't see how this is a problem. Someone has to put the files there. Either some tool that manages it for the user, so the users doesn't care, or the user themselves. And if it is a tool, they will most likely place it in

I might have over-estimated the difficulty when I wrote this sentense. That is implementing a typst tool that respects `TYPST_PACKAGE_PATH` as a search paths, reading it and watching it correctly. I thought it is hard but currrently I feel it is not that hard. But as a tool developer, I'm still not liking it for it adds unnecessary difficulty.

> > If we have a list, updating "local packages" atomically should lock many directories, which sounds not easy. Locking the layers (directories) above the updating layer, and removing the one in some layer, and adding one to another layer.
> 
> I'm unsure what locking has to do with anything. When would locking have to occur?

The hard point is that this is happening when:
1. multiple tools would like to read and watch your package directories concurrently and
2. there is some other tool or people writing either one or multiple package directories.

I have memory that typst suffered with writing package data atomically so there might be problems if we will make it more complex.

> > but we only get little benefit: if I don't want to change existing package directories, just create another package directory placed over some package directories. then, we may also get a increasing number of package directories in the same computer.
> 
> I'm sorry but I don't really understand.

That's what I imainged with the worst future with allowing a search paths, where some orgnizations or people will abuse the search paths. I believe you won't, but allowing this will open the door to make the worst thing. It is not a strong point like above ones, but my personal feeling.

RossSmyth:
I will do my best to interpret what you mean again.

I don't understand why a search path "should lose," as you don't give any justification for such a thing. 

Search paths in other compilers are extremely common, so I'm unsure what abuse may occur. What do you imagine? You say that the user must check multiple paths, but those are paths that the user themselves choose. They can choose not to use them.

Something about locking, but locking cannot occur at a directory level unless you have lock files, which Typst does not. So I don't really understand the motivation for that. Concurrently reading and writing files is, in general, hard to make sound and I don't see how it would be in-scope for Typst to care if the user is doing cursed things with their files.

terefang:
i have many the same use-cases.

RossSmyth:
My main motivation is I would like to improve nixpkgs so that Typst packages "just work." For example, if you add a library that has a pkg-config file in the nix output to your build derivation, [the pkgconfig package can detect that, and add it to the list of packages that the build system can see](https://github.com/NixOS/nixpkgs/blob/2cd2decd3aebe7bff63c51b978c27a8e91bbab3d/pkgs/build-support/pkg-config-wrapper/setup-hook.sh#L16-L17). Without search paths, adding such a feature for Typst is harder.

laurmaedje:
I personally always found linker search paths in compilers a source of confusion. I'm not sure I want to encourage building out the system in this direction. Your use case with multiple directories sounds like it would be better served by expanding `typst.toml` with support for Rust-style path/git dependencies, which is something we want to take a look at in the future.

I was already a bit hesistant with adding TYPST_PACKAGE_PATH in the first place. See also the discussion preceding https://github.com/typst/typst/pull/4227#issuecomment-2147807663.

RossSmyth:
That is one solution, though I do consider it ugly. For Nix it would require that Nix rewrites every `typst.toml` file once copied into the sandbox. For other tools that try to integrate with Typst, I do not know what they would do. The Typst compiler needs someway to find dependencies. 

The Typst team seems very against this idea of dependencies that aren't in complete control of the Typst compiler, but I would not let perfect be the enemy of good. Currently using dependencies that are outside Typst Universe is not very fun, and I am trying hard to make it less painful with nixpkgs support.

I am sorry that you find search paths confusing, but they are used extremely widely in build systems, are very pragmatic for real problems people have, and I do not think all things need to be designed for beginners since people are usually only beginners when they are new to something. Could you explain what you find confusing about my PR? I can add supporting documentation and examples if needed.

laurmaedje:
I don't mean that your PR is confusing, but that when I started programming configuring linker search paths was super confusing to me. When it doesn't work, it usually just doesn't work without any clear indication why. Cargo-style dependency manager clicked much more with me and it's also easier to generate good error messages for this style. It's of course subjective, but so is programming language design in general.

> I would not let perfect be the enemy of good

To an extent, that is my approach with Typst though. In the sense that I want to have a coherent plan in an area before starting to take steps in one direction. I don't currently see a clear vision where ever-more package management configuration flags take us and whether it's the right target compared to a different approach, with e.g. more `typst.toml` support.

terefang:
@laurmaedje - i see a list of different solutions here, but for a cli tool if you have an option that is no setable via cli but only config file or vice-versa isnt going to be popular.

there will always be people arguing for one use-case or another until both the cli and config files and environment variables have feature parity, which isnt necessarily a bad thing at all..

since you already accepted the validity of multiple font-paths, multiple package-paths is one of the next logical consequences ... maybe followed by multiple root-paths.

RossSmyth:
typst.toml support doesn't exist for projects right now so it is currently just something in someone's imagination. This isn't even more flags. 

> I don't mean that your PR is confusing, but that when I started programming configuring linker search paths was super confusing to me. 

Which again. People are only beginners when they are beginning. They can just learn and then they are no longer beginners. If there are problems with errors, they can be made better!  I would not prevent progress for 

1. Some mystical definition of a beginner who will shortly not be a beginner anymore once they read two sentences on what a search path is
2. Some sort of manifest format that doesn't exist yet. There are manifests for Typst Universe packages, but not projects yet. And I don't really see any progress at project-level manifests. There are no PRs or issues open about this that I see.

Making thing beginner-friendly is good, but making all feature for beginners is a mistake in my opinion. And I don't even think search paths are hard. Basically every developer has used them or will .

laurmaedje:
> Some sort of manifest format that doesn't exist yet. 

> There are no PRs or issues open about this that I see.

That's a knockout argument. In this fashion, you could argue for anything that no one in our team is actively working on that we should ignore our vision and accept the first community suggestion.

terefang:
looks again as a 20:80 question.

* do people mainly use typst in a project-style setup – unknown
* is there current work being done on a project manifest – no
* would people profit from the suggested solution above, regardless of their setup – probably
* can the proposed solution also be applied to a future manifest implementation – certainly

looks like a no-brainer to me.

terefang:
> I wonder if it would make more sense to retain the original idea of being able to map one namespace to one path. Would love to hear more about potential use cases, and how that idea would interact with them.

i would not make that an "exclusive or" to the proposed solution above, it would give may a more structured way in their setups.

the challenge is the current chicken-and-egg problem: 

* the need is uncertain, 
* because nobody tried before, 
* because the current version does not support it.

if you at least ship the above in production, you can collect feedback with further suggestions.

laurmaedje:
My point is really just that I feel like there isn't a clear vision here for how package management in Typst should work in, let's say, two years. I don't want to start moving into a direction without a plan. You might say that this is an obvious extension of TYPST_PACKAGE_PATH, but the _existence_ of TYPST_PACKAGE_PATH by itself is also not really planned. It was contributed at an earlier time and merged solely because I resisted less than here. I see its use, but don't think it's particularly great. It is a reasonable extension of Typst searching for packages in a fixed directory, but I'm not even sure that was a good idea in the first place. It was just how we started. We could as well have ended up in a world where the path for each non-preview package must be specified in the CLI and then the obvious extension would have been to support adding this to a config file format (`typst.toml`). 

Generally, I dislike evolving the language based on "what is the next step towards the currently visible local optimum". Instead, we should try to discover the global optimum or at least potentially better local optima (of course, there isn't a global optimum per se as this is all subjective, but I hope you get my point). We might _still_ end up with something like this, but I'm not sure we'd do.

terefang:
while i do see validity in the reasoning not to make a "patched up carpet" out of typst so to say, but please not also to avoid too much conservative over-engineering.

suggestion: might it be possible to enable the above solution as an experimental feature via cli option ?

then people that like to test it can enable it, and people that dont care are not bothered by it.

Myriad-Dreamin:
> while i do see validity in the reasoning not to make a "patched up carpet" out of typst so to say, but please not also to avoid too much conservative over-engineering.
> 
> suggestion: might it be possible to enable the above solution as an experimental feature via cli option ?
> 
> then people that like to test it can enable it, and people that dont care are not bothered by it.

@terefang imho it is not an effective suggestion. If there are some drawbacks and we don't investigate more, it will be a temporary solution instead of an experimental feature. A question is that: why can people process rust crates in nix, when rustc and cargo doesn't respect any "search paths"? The X-problem of @RossSmyth may build typst package layers in nix, and it is probably not necessary to solve the Y-problem: convincing that the search path is good to typst. Is search path the only solution to make life easy in nix? What's the drawback of search-based path solutions (C/C++, python, java) and registry-centric solutions (kotlin, rust, typst) and how can we learn from them? The only way to push forward the conversation is introducing more information and solving the conerning points.

terefang:
as i already wrote above, there will always be use-case with pros and cons.

i also wrote here https://discord.com/channels/1054443721975922748/1054443722592497796/1383376688435429377 
how i would like mountable import paths.

i also like to apply the KISS principle to solution to not suffer from much over-engineering.

i am championing the solution above, because it is most unixy in approach and for my day job is basically "a must" once my netdevops team decides to switch over to typst.

terefang:
##  What's the drawback of search-based path solutions (C/C++, python, java) 

### 🔻 Drawbacks of Search-Based Directory Path Solutions

1. **Performance Overhead**

   * Searching the filesystem—especially recursively—can be **slow**, particularly on large directory trees.
   * Unindexed searches require I/O operations for each folder, causing latency on spinning disks or network filesystems.

2. **Inefficiency for Repeated Lookups**

   * If the same path or file is searched repeatedly (e.g., during build processes or environment resolution), repeated searches cause unnecessary overhead.
   * Caching or indexing is usually needed to make search viable at scale.

3. **Non-Determinism**

   * Search order (e.g., breadth-first vs. depth-first) may return different results depending on file system structure.
   * This can lead to inconsistent behavior, especially when multiple paths match a pattern or name.

4. **Ambiguity and Conflicts**

   * If multiple directories contain files with the same name, search-based resolution may select the wrong one.
   * This is a common source of bugs in build systems, scripts, or dependency loaders.

5. **Security Risks**

   * Searching can unintentionally traverse into directories with **restricted permissions**, symlink loops, or mount points.
   * Without proper guards, this can cause crashes or even security breaches (e.g., privilege escalation, path traversal attacks).

6. **Poor User Feedback**

   * If a search fails (e.g., file not found), it’s often unclear **where** it looked or why it failed—especially when paths are resolved dynamically.

7. **Not Scalable**

   * In environments with **millions of files** (e.g., large monorepos, IDEs), search-based path discovery becomes infeasible without indexing or smarter caching.

### 🧠 Summary

Search-based directory path solutions are **simple but fragile**. They’re suitable for small or controlled environments but become **unpredictable, slow, or unsafe** in larger, more complex systems. Alternatives like **explicit path configuration**, **path maps**, **indexing**, or **virtual filesystems** offer better scalability and maintainability.

terefang:
What's the benefits of registry-centric directory/file search-based solutions (kotlin, rust, typst) ?

---

### ✅ Key Benefits of Registry-Centric Search-Based Systems

1. **Centralized Dependency Management**

   * Registries provide a **single source of truth** for modules, packages, or files.
   * Tools like Cargo or Gradle can resolve versions, manage upgrades, and enforce consistency across teams.

2. **Version Control and Reproducibility**

   * Packages are versioned and immutable once published, enabling **lockfile-based reproducible builds**.
   * You can recreate an exact environment by pinning versions (e.g., `Cargo.lock`, `gradle.lockfile`, `typst.lock`).

3. **Decoupled File Organization**

   * Files don’t have to live in specific directories or relative paths. Instead, they can be looked up by **semantic identifiers** like `group:artifact:version` (Gradle) or `cratename@version` (Cargo).
   * This allows greater **modularity and reuse** without rigid directory structures.

4. **Automatic Resolution and Installation**

   * Missing packages can be auto-installed from the registry (e.g., `cargo build` automatically fetches crates).
   * Reduces manual effort and errors during setup or onboarding.

5. **Search + Registry = Smart Defaults**

   * Combines intelligent local path search with structured metadata from the registry.
   * This allows context-aware resolution, fallback behaviors, and overrides (e.g., path overrides in Cargo).

6. **Better Tooling and Ecosystem Integration**

   * IDEs, linters, and build tools integrate tightly with registries to provide features like:

     * Autocompletion
     * Dependency graphs
     * Security scanning
     * Type-aware code navigation

7. **Caching and Performance Optimizations**

   * Registry-based systems typically use **local caches** (e.g., `~/.cargo`, `~/.gradle`) to speed up resolution after the first run.
   * Supports offline work and CI/CD efficiency.

8. **Security and Trust**

   * Verified or signed packages can be enforced at the registry level.
   * This reduces the risk of using unvetted code, especially in open source environments.

---

### 🧠 Summary

Registry-centric path resolution combines **scalability, robustness, and developer ergonomics**. It's especially valuable in large projects, collaborative environments, and systems that benefit from strict versioning and package isolation. These benefits far outweigh the drawbacks in most professional-grade tooling ecosystems.

Enivex:
I don't think copying in a wall of LLM generated text is very helpful

terefang:
> I don't think copying in a wall of LLM generated text is very helpful

even if it is ai-accumulated, it still lists some valid informative pros and cons.

terefang:
> I don't think copying in a wall of LLM generated text is very helpful

i give up for the sake of not insulting anyone.

laurmaedje:
In line with my comment https://github.com/typst/typst/pull/6190#issuecomment-2988759636, I'll close this PR for the time being. I am happy to continue discussing how best to evolve the package management story, but I think an issue would be a better place for that than this PR.

blaggacao:
https://github.com/typst/typst/pull/4776 might serve as a good in-house precedent.

In my opinion, a list of search paths for `TYPST_PACKAGE_PATH` is a good way to implement "overlay" strategies that are useful in all kinds of scenarios, from development to customization.

They can be implemented with symlinked paths, but control over the file system is a bigger ask in any given environment than control over env variables. So that latter approach is more flexible, while not precluding the former, either.

Edit: At a minimum, the singular name would have to be deprecated in favor of the plural. cc/ @RossSmyth

PR #6192, opened 2025-04-18
Error if an unexpected named argument was received

frozolotl:
Prior to this change, calling symbols with a named argument swallowed that argument. This is probably the least expected possible behavior. Better would be if the argument was simply inserted verbatim into the document, like it's done with positional arguments.

I'm not sure that is the best idea, however. It further increases the difficulty of humans deciding whether a function call actually calls a Typst function and of what the behavior would be.

Reconstructing arguments with valid spacing is not quite trivial either. It's not a problem usually, but you can construct pathological cases where even the currently existing reconstruction of positional arguments isn't correct. This would also be an issue for named arguments.

I thus simply chose to error out if any named arguments were found.

FWIW, if it is decided that reconstructing is the preferred choice after the fact, the number of total breaking changes at least stays the same.

## Example

```typ
$ phi(1,2,,3, alpha: 1) $

// What even is the difference?
$ hat(a b, size: #200%) $
$ phi(a b, size: #200%) $
```


laurmaedje:
Makes sense. Thank you!

PR #6193, opened 2025-04-18
Fix frac syntax section typo

Andrew15-5:
The noun "parenthesis" changes from single to plural, while should stay the same across sentences.


laurmaedje:
Thanks!

PR #6208, opened 2025-04-23
Add an example of show-set `place.clearance` for figures in the doc

MDLC01:
The doc reads:

> The gap between the main flow content and the floating figure is controlled by the [`clearance`](https://typst.app/docs/reference/layout/place/#parameters-clearance) argument on the `place` function.

This should be illustrated in the following example, because I doubt many users will intuitively understand that this means they can use this show-set rule.

### DIFF
--- a/crates/typst-library/src/model/figure.rs
+++ b/crates/typst-library/src/model/figure.rs
@@ -132,6 +132,9 @@ pub struct FigureElem {
     ///   caption: [A glacier],
     ///   image("glacier.jpg", width: 60%),
     /// )
+    /// #show figure: set place(
+    ///   clearance: 1em,
+    /// )
     /// #lorem(60)
     /// ```
     pub placement: Option<Smart<VAlignment>>,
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/model/figure.rs
```diff
@@ -132,6 +132,9 @@ pub struct FigureElem {
     ///   caption: [A glacier],
     ///   image("glacier.jpg", width: 60%),
     /// )
+    /// #show figure: set place(
```
laurmaedje, line 135: The show rule doesn't work because it is after the figure.

MDLC01: Indeed, sorry

MDLC01:
Given that there is no baseline to compare with, it's probably not very important that it does not have any visible effect in the docs render. What matters is that this provides a clear example to the users.

laurmaedje:
Thanks!

PR #6216, opened 2025-04-28
Remove line break opportunity when math operator precededes a closing paren

### LINKED ISSUE #6152: Line break before a closing parenthesis
### Description

I noticed Typst seems much happier than TeX to put a line break before a closing parenthesis in math mode.

The following shows the issue on a normal line with plenty of opportunity to squeeze some spaces. It is meant to be pasted into an otherwise empty Typst file, on typst 0.13.1 (8ace67d9).

```typst
Foo. La croissance de $r$ découle de ce résultat pour $(r, l)$ vu comme adjonction entre $(X, ≥)$ et $(Y, ≥)$
```

The last parenthesis is rendered alone on the second line of the pdf.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

wrzian:
Closes #6152

I just did the simple thing of checking if the next element is a closing paren, and then avoided splitting the frame items. Not fully if this is the desired behavior, would appreciate opinions.


MDLC01:
I believe `lr` changes the class of the first and last element to Opening and Closing anyway, so it shouldn't be an issue. But to be sure, if I write `lr(\]0, 1\[)` (e.g., for the French-style interval notation), is there a linebreak opportunity before "["?

mkorje:
> I believe `lr` changes the class of the first and last element to Opening and Closing anyway, so it shouldn't be an issue. But to be sure, if I write `lr(\]0, 1\[)` (e.g., for the French-style interval notation), is there a linebreak opportunity before "["?

I just tried this out and (as expected) there is not.

PR #6223, opened 2025-05-01
Fix text overhang example in docs

### LINKED ISSUE #6050: Text overhang example doesn't show overhang
### Description

https://typst.app/docs/reference/text/text/#parameters-overhang
The text says that there is a hyphen in the first line, while in the image there isn't. Due to this there is also no overhang (no character to move into the margin), which this is supposed to showcase.

mkorje:
Closes #6050.

| Old | Current | New |
| - | - | - |
| old | cur | new |

### DIFF
--- a/crates/typst-library/src/text/mod.rs
+++ b/crates/typst-library/src/text/mod.rs
@@ -348,15 +348,17 @@ pub struct TextElem {
     /// This can make justification visually more pleasing.
     ///
     /// ```example
+    /// #set page(width: 220pt)
+    ///
     /// #set par(justify: true)
     /// This justified text has a hyphen in
-    /// the paragraph's first line. Hanging
+    /// the paragraph's second line. Hanging
     /// the hyphen slightly into the margin
     /// results in a clearer paragraph edge.
     ///
     /// #set text(overhang: false)
     /// This justified text has a hyphen in
-    /// the paragraph's first line. Hanging
+    /// the paragraph's second line. Hanging
     /// the hyphen slightly into the margin
     /// results in a clearer paragraph edge.
     /// ```


laurmaedje:
Thanks!

mewmew:
I was just about to submit a PR when I noticed you'd already fixed this issue @mkorje.

I ran the example through the Discord bot, and it renders correctly (with a hyphen on the second line).

Results from Discord bot (using version 0.13.1 of Typst).
overhang_discord_bot

However, I'm unable to reproduce this locally (running on Arch Linux). I tried using both Typst 0.13.1 and the latest commit on *main*.

Results from local system (using version 0.13.1 of Typst: typst 0.13.1 (8ace67d9)).

overhang_typst_0_13_1

Results from local system (using latest commit of Typst: typst 0.13.1 (02645349)).

overhang_latest_git

This was based on the follow Typst example:

```typst
#set page(width: 220pt)

#set par(justify: true)
This justified text has a hyphen in
the paragraph's second line. Hanging
the hyphen slightly into the margin
results in a clearer paragraph edge.

#set text(overhang: false)
This justified text has a hyphen in
the paragraph's second line. Hanging
the hyphen slightly into the margin
results in a clearer paragraph edge.
```

Any idea why the output results are different between the Discord bot output and a local system? Perhaps different fonts used?

laurmaedje:
Are you sure the margins are the same? Docs use 15pt margin rather than auto margins.

mewmew:
> Are you sure the margins are the same? Docs use 15pt margin rather than auto margins.

Wonderful! That was the issue. When using 15pt margins, the examples are indeed reproducible on my local machine.

Thanks!

PR #6226, opened 2025-05-01
Use the right field name for `figure.caption.position`

MDLC01:
It's a bit annoying that synthesized fields are not listed automatically in the documentation (https://github.com/typst/typst/issues/6227), but at least for now let's make sure we list them under the right name.

Also, the documentation for `body` reads: "Can be used alongside `kind`, `supplement`, `counter`, `numbering`, and `location` to completely customize the caption." But there is no `location` field on `figure.caption`. Actually, there is no way to reproduce the default `figure.caption` show rule in Typst right now, because the `figure_location` field is internal.


laurmaedje:
Thanks!

PR #6232, opened 2025-05-02
Respect RTL cell layouting order in grid layout

### LINKED ISSUE #5178: Counter/State problem in Grid/Table when text is set to RTL
### Description

Hello again.
I'm sorry for opening a new issue but I think this issue's title and example is way better than #5171 
You can remove or close the previous one.

In order to see the problem, you can just try the example in the following URL.
Please do as follows:
1. Uncomment the first line and see the result.
2. Comment the first line and see the result.

It will perfectly show the problem.

### Reproduction URL

https://typst.app/project/rPkKhKHdLohNy8xdpV8W42

### Operating system

Linux

### Typst version

saecki:
Fixes #5178 by layouting cells from right to left in grid layout, when RTL is enabled.


saecki:
I think for consistency this loop should maybe also be changed? https://github.com/typst/typst/blob/62e6c97df592b85c9d7dff08c77c99baa495b45f/crates/typst-layout/src/grid/layouter.rs#L577

And it's probably a good idea to add a test case for the linked issue.

PgBiel (commented):
The logic seems to check out per my quick sketch below (very high quality :sunglasses:). However, this fix probably needs to be applied to `check_for_rowspans` as well in `rowspans.rs`. (It should also be tested by having rowspan cells in tests.)

image

saecki:
> > I think for consistency this loop should maybe also be changed?
> > https://github.com/typst/typst/blob/62e6c97df592b85c9d7dff08c77c99baa495b45f/crates/typst-layout/src/grid/layouter.rs#L577
> 
> Yeah that was a good call. It determines the order in which fills are rendered. In principle it doesn't make a difference (other than, perhaps, in a very minimal scale at the border of each cell) as the fills don't usually intersect, but I suppose there could be a more visible effect if the fill is made using a `tiling`. Though I'm pretty sure that can't overflow / is clipped, but it's worth it for the peace of mind anyway.

I've fixed the immediate issue with RTL layouting for rowspans and added some tests for that.

But the way rowspans are handled, currently doesn't play nice with counters.
For example the following grid layout (in LTR) returns a somewhat unexpected result:
```typ
#let test = counter("test-counter")
#block(below: 1em)[
  #grid(
    columns: (1fr, 1fr),
    gutter: 1em,
    stroke: black,
    inset: 1em,
    grid.cell(rowspan: 3, [
      cell 1: counter
      #test.step()
      #context test.get().first()
    ]),
    grid.cell(rowspan: 1, [
      cell 2: counter
      #test.step()
      #context test.get().first()
    ]),
    grid.cell(rowspan: 1, [
      cell 3: counter
      #test.step()
      #context test.get().first()
    ]),
    grid.cell(rowspan: 1, [
      cell 4: counter
      #test.step()
      #context test.get().first()
    ]),
  )
]
```
image

But that is a whole other issue.

tests/suite/layout/grid/rtl.typ
```diff
@@ -193,3 +193,93 @@
...
+  grid.cell(rowspan: 1, [
+    b: // should produce 2
+    #test.step()
+    #context test.get().first()
+  ]),
+  grid.cell(rowspan: 1, [
+    d: // should produce 3
+    #test.step()
+    #context test.get().first()
+  ]),
```
PgBiel, line 284: This isn't testing rowspans at the same row as `rowspan: 1` is the usual case, let's test that. Same for other tests. Although it can't hurt to test mixing them as well as you are doing - maybe they could be separate tests.

```suggestion
  grid.cell(rowspan: 2, [
    b: // should produce 2
    #test.step()
    #context test.get().first()
  ]),
  grid.cell(rowspan: 2, [
    d: // should produce 3
    #test.step()
    #context test.get().first()
  ]),
```

crates/typst-layout/src/grid/rowspans.rs
```diff
@@ -185,10 +183,8 @@ impl GridLayouter<'_> {
     /// Checks if a row contains the beginning of one or more rowspan cells.
     /// If so, adds them to the rowspans vector.
     pub fn check_for_rowspans(&mut self, disambiguator: usize, y: usize) {
-        // We will compute the horizontal offset of each rowspan in advance.
-        // For that reason, we must reverse the column order when using RTL.
-        let offsets = points(self.rcols.iter().copied().rev_if(self.is_rtl));
-        for (x, dx) in (0..self.rcols.len()).rev_if(self.is_rtl).zip(offsets) {
+        let offsets = points(self.rcols.iter().copied());
+        for (x, dx) in (0..self.rcols.len()).zip(offsets) {
```
PgBiel, line 187: I suspect this double reversal was indeed an oversight... The order of `self.rowspans` changes, but this doesn't seem to affect rowspan layout code as it doesn't consider that order (all information is contained in x / dx), so that was a good catch.

saecki:
> > But the way rowspans are handled, currently doesn't play nice with counters.
> 
> `rowspan: 1` is the default, worth mentioning, so it isn't handled differently in the code. With that said, it's hard to produce the semantics you're seeking as rowspans can only be laid out after all of their rows have as well, so the best we can do at the moment is lay them out at the end rather than the opposite. This is especially relevant for multi-page rowspans, which are always laid out at the end of the page, after everything else.
> 
> There could be some way to layout the rowspan at the end but keep the correct order, but we'd need to either predict how much space to "leave" for the rowspan ahead of time in the frame's vector of items (probably not possible), or just insert the rowspan in the frame and accept the (in principle) undesirable performance cost. It could be studied whether that's possible to do in a good enough way.

Yeah, I think this is fine for now.

tests/suite/layout/grid/rtl.typ
saecki: That was kind of the intention, but I've renamed these tests to `mixed` and added some other tests with `rowspan > 1` to test how rowspans interact :+1:

crates/typst-layout/src/grid/rowspans.rs
saecki: This change here only works because, the code in `layout_rowspan` was also changed, otherwise the double reversal would have been necessary. Essentially the same transformation as for the other layout code was done here as well, where now the offset is computed from whichever direction the columns are drawn (LTR or RTL) and then the position is then fixed up afterwards for RTL in `layout_rowspan`: 
```rs
        let dx = if self.is_rtl { self.width - (dx + width) } else { dx };
```

PgBiel: You're right, I forgot that `points()` accumulates and thus the order of its inputs matters. :+1:

crates/typst-layout/src/grid/layouter.rs
```diff
@@ -643,18 +643,12 @@ impl<'a> GridLayouter<'a> {
...
-                            // over to unrelated columns to the right in RTL.
-                            // We avoid this by ensuring the fill starts at the
-                            // very left of the cell, even with colspan > 1.
-                            let offset =
-                                if self.is_rtl { -width + col } else { Abs::zero() };
-                            let pos = Point::new(dx + offset, dy);
+                            let mut pos = Point::new(dx, dy);
+                            if self.is_rtl {
+                                // In RTL cells expand to the left, thus the position
+                                // must additionally be offset by the width of the cell.
```
PgBiel, line 649: We will need to reflow this comment to stop at 80 chars (a little style guideline we have).

crates/typst-layout/src/grid/layouter.rs
```diff
@@ -1314,17 +1299,18 @@ impl<'a> GridLayouter<'a> {
                     let fragment =
                         layout_cell(cell, engine, disambiguator, self.styles, pod)?;
                     for (output, frame) in outputs.iter_mut().zip(fragment) {
-                        let mut pos = pos;
+                        let mut pos = offset;
                         if self.is_rtl {
-                            let offset = -width + rcol;
-                            pos.x += offset;
+                            // In RTL cells expand to the left, thus the position
+                            // must additionally be offset by the width of the cell.
```
PgBiel, line 1305: also reflow here

PgBiel (approved):
Looks good to me. Only two style nits and it's mergeable.

Amirhosein-GPR:
Thank you very much! :grinning:

PR #6234, opened 2025-05-02
Fix tracing of most field call expressions

### LINKED ISSUE #6115: `typst_ide::tooltip` returns None for module field calls
### Description

Looking at an expression such as:
```typst
#pdf.embed("A", bytes("B"))
```

When calling `typst_ide::tooltip` with `cursor` pointing somewhere at the `embed` identifier, `None` is returned. Instead, I would have expected the first sentence of `pdf.embed`'s markdown documentation ("A file that will be embedded into the output PDF.") to be returned as a tool tip.

Same can be observed indirectly in the Typst web app - when hovering over the `embed` function in the above expression, no tool tip is shown.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

IgKh:
Attempt to fix #6115 in the most straightforward way I could think of.

Since the `Eval` implementation for `typst_syntax::ast::FuncCall` evaluates all elements of a field call directly without returning to the general `Eval` implementation for expressions - it misses the trace point there. To compensate, add a separate trace point for field function calls specifically.

It is still unable to trace calls to the hard coded mutating methods on arrays and dicts.

### DIFF
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -37,7 +37,12 @@ impl Eval for ast::FuncCall<'_> {
             let target = access.target();
             let field = access.field();
             match eval_field_call(target, field, args, span, vm)? {
-                FieldCall::Normal(callee, args) => (callee, args),
+                FieldCall::Normal(callee, args) => {
+                    if vm.inspected == Some(callee_span) {
+                        vm.trace(callee.clone());
+                    }
+                    (callee, args)
+                }
                 FieldCall::Resolved(value) => return Ok(value),
             }
         } else {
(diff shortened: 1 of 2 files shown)


IgKh:
@laurmaedje @PgBiel appologies for tagging, although the contributing guidelines do say to do that for PRs that are not reviewed for a while. This one has not received any kind of response for more than two weeks now, and the associated issue for longer than that.

It is OK / not OK / something the project doesn't care about? Thanks.

laurmaedje:
@IgKh Typically, I aim to review PRs relatively quickly, but in the past few months this was not possible because I was very busy with other things. Things should go back to normal in the next week hopefully. So the lack of response so far has no bearing on the PR being merged or not.

IgKh:
@laurmaedje That's alright, thanks for the update.

PR #6243, opened 2025-05-05
Fix stroking of glyphs in math mode

### LINKED ISSUE #6170: `text.stroke` doesn't work in math mode
### Description

I've noticed a couple of people in the past few days faced this problem.

```typ
#text(stroke: green)[x]
#text(stroke: green)[$x$]
$text(stroke: #green, x)$
#show math.equation: set text(stroke: green)
$x$
```

Image

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

mgazeel:
Hi! This is meant to resolve the issue of `text.stroke` not working in math mode. (Fixes #6170)

List of changes:
- `stroke` has been added to `GlyphFragment`
- A test with this functionality has been added to `tests/suite/math/equation.typ`
- I've tried to partially document `GlyphFragment`, as discussed on discord @mkorje frequently contributes to math mode maybe he would be able to provide descriptions for the rest of the fields?
- During debugging I've discovered that multiple struct/enums format to `Text`, I've tried to remedy that by modifying them to be `TextElem` (inline with the other elements) and `TextItem`. I'm not sure if this is a change that should be included in this PR (since its technically unrelated to the issue), or if its something that would be desired if so, please let me know and I'll remove it.

Note: There is probably much more dichotomy between `TextElem` and the text rendering that happens inside of math mode (just even by looking at the amount of settings we can pass to `show math.equation: set text(...)`, I haven't tried to address anything than stroke in this PR.


crates/typst-layout/src/math/fragment.rs
```diff
@@ -227,14 +227,52 @@ impl From<FrameFragment> for MathFragment {
     }
 }
 
+/// A fragment of a glyph in math mode.
 #[derive(Clone)]
 pub struct GlyphFragment {
+    /// A unique id.
```
mkorje, line 233: I think it'd be clearer to say this is a glyph's id in the font.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -227,14 +227,52 @@ impl From<FrameFragment> for MathFragment {
     }
 }
 
+/// A fragment of a glyph in math mode.
```
mkorje, line 230:
```suggestion
/// A single glyph in math.
```
Currently it reads like this struct defines a part of a glyph, as opposed to a fragment in math which is a single glyph.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -227,14 +227,52 @@ impl From<FrameFragment> for MathFragment {
...
 #[derive(Clone)]
 pub struct GlyphFragment {
+    /// A unique id.
     pub id: GlyphId,
+
+    /// The base character.
     pub c: char,
+
+    /// A font family descriptor or priority list of font family descriptor.
```
mkorje, line 239: This field isn't quite the same as the one on `TextElem`, which is a list of font families. It's just a single/lone OpenType font.

mkorje (commented):
Thanks for the PR!

With the documentation for the `GlyphFragment`, I'm not sure linking to e.g. TextElem or language codes really pulls its weight, especially when the associated type will have that documented.

> During debugging I've discovered that multiple struct/enums format to `Text`, I've tried to remedy that by modifying them to be `TextElem` (inline with the other elements) and `TextItem`. I'm not sure if this is a change that should be included in this PR (since its technically unrelated to the issue), or if its something that would be desired if so, please let me know and I'll remove it.

As for this, I don't know if it's desirable. I'll leave that for someone else to answer.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -227,14 +227,52 @@ impl From<FrameFragment> for MathFragment {
...
+    ///
+    /// This works much akin to how fonts are specified for [TextElem], but for modification
+    /// a show rule must be used.
+    ///
+    /// ```example
+    /// #show math.equation: set text(font: "Fira Math")
+    /// $ f i r a m a t h $
+    /// ```
```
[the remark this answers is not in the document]
mgazeel: My objective here was to leave breadcrumbs for maintainers to be able to easier understand the point of this struct and what all of its components do, more than targeting end users.

(Admittedly that means that examples are probably necessary I can remove them if you think its too much)

In general I think that it would be nice if this struct had some more info attached to it, since for example it took me a lot of time to wrap my head around how its used, by what, where, and what it does.

But, again, I'm new to the project and I'm just throwing ideas out there. If this is not desirable let me know and I'll delete it from the PR.

tests/suite/math/equation.typ
```diff
@@ -297,3 +297,11 @@ Looks at the @quadratic formula.
 #set page(width: 150pt)
 #set text(lang: "he")
 תהא סדרה $a_n$: $[a_n: 1, 1/2, 1/3, dots]$
+
+--- issue-6170-equation-stroke ---
+// We expect both of these two have a stroke.
+// But only one does...
```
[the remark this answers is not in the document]
mgazeel: Hah, yes that was the old behavior.

I've reworded!

crates/typst-layout/src/math/fragment.rs
mgazeel: Reworded!

crates/typst-layout/src/math/fragment.rs
mgazeel: Reworded!

mgazeel:
I've split the changes into two commits (so that it is easy to just cherry pick the first one, if we don't want to end up adding the docs or debugging changes)

crates/typst-layout/src/math/fragment.rs
mkorje: It does work, but is happening in a couple different places earlier on during layout. The fields on the GlyphFragment have already been "resolved".

This function gets called before you call `MathContext::new()` to get the font.
https://github.com/typst/typst/blob/2a258a0c3849073c56a0559dbd67ee2effcd1031/crates/typst-layout/src/math/mod.rs#L239-L255
And then here in the `new` method the OpenType features get applied:
https://github.com/typst/typst/blob/2a258a0c3849073c56a0559dbd67ee2effcd1031/crates/typst-layout/src/math/mod.rs#L402-L412

crates/typst-layout/src/math/fragment.rs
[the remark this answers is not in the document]
mkorje: > In general I think that it would be nice if this struct had some more info attached to it, since for example it took me a lot of time to wrap my head around how its used, by what, where, and what it does.

I definitely agree the code throughout math could use some more comments! But yeah I'm of the opinion that the examples are a bit much and aren't that useful (I could be very wrong here of course!). Maybe we can wait to hear what someone else thinks (e.g. someone from the Typst team when they review this).

mkorje (commented):
All looks good to me (bar the discussion on documentation)!

PR #6247, opened 2025-05-05
Avoid unnecessary groups around elements to reduce size of exported SVGs

Bequen:
Every element like path had group around it that defined it's transform. These changes remove the groups and move the transformations to the element itself. This reduces the overall size of the output SVG.

Added new SVG path builder using relative coordinates `SvgRelativePathBuilder`. The previous with global coordinates is still used for glyph paths as `SvgGlyphPathBuilder`. Using relative coordinates allows to transform the path without changing value of the `d` attribute.

To transform images, `x` and `y` attributes are used. To transform groups, the group's transform is added to the current transform, which is passed further down. The reason is that the exported SVG had many nested groups with transforms. This change merges all these transforms into a single one.

Changes include basic visual test.


crates/typst-svg/src/text.rs
```diff
@@ -22,7 +22,14 @@ impl SVGRenderer {
 
         self.xml.start_element("g");
         self.xml.write_attribute("class", "typst-text");
-        self.xml.write_attribute("transform", "scale(1, -1)");
+        self.xml.write_attribute(
+            "transform",
+            &format!(
+                "scale(1, -1) translate({} {})",
```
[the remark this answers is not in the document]
Bequen: I do not think so. The `#text` element in Typst cannot be rotated directly, but with a `#rotate` element. That one creates a group, that spans the text, and which has the desired transform matrix, with rotation etc. as seen on line 204 in `lib.rs`.

But you were onto something. I did not test rotations in general :D And to my surprise, they were not working.

When a group with non identity matrix transformation is found, the entire accumulated transform from parents should be released, and the accumulation reset. Now it is fixed :)

crates/typst-svg/src/lib.rs
```diff
@@ -375,18 +369,31 @@ impl Display for SvgMatrix {
     }
 }
 
-/// A builder for SVG path.
-struct SvgPathBuilder(pub EcoString, pub Ratio);
+/// A builder for SVG path using relative coordinates.
+struct SvgRelativePathBuilder(pub EcoString, pub Ratio, pub Point);
```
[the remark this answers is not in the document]
Bequen: Good point, you are right

crates/typst-svg/src/text.rs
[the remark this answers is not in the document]
Bequen: I also changed the test a little, to include rotations and check them.

laurmaedje:
I ran a script to compare the output of the old vs updated SVG export and there were some changes. I attached them and the script (with txt extension because GitHub) for your convenience. In each image, left is your branch, right is main, and center is the diff.  Most changes are very minor and can be ignored, but some files (e.g. `curve-move-multiple-even-odd`) changed more. 

[diff.zip](https://github.com/user-attachments/files/20672107/diff.zip)
[typst-svg-cmp.sh.txt](https://github.com/user-attachments/files/20672136/typst-svg-cmp.sh.txt)

Bequen:
> I ran a script to compare the output of the old vs updated SVG export and there were some changes. I attached them and the script (with txt extension because GitHub) for your convenience. In each image, left is your branch, right is main, and center is the diff. Most changes are very minor and can be ignored, but some files (e.g. `curve-move-multiple-even-odd`) changed more.
> 
> [diff.zip](https://github.com/user-attachments/files/20672107/diff.zip) [typst-svg-cmp.sh.txt](https://github.com/user-attachments/files/20672136/typst-svg-cmp.sh.txt)

Hello, sorry for taking so long. I did not have much time lately. I fixed the tiling issues, but I cannot seem to figure out the gradients yet.

laurmaedje:
No worries, there's no rush. The label is just to keep things organized.

Bequen:
All should be fixed :)

crates/typst-svg/src/image.rs
```diff
@@ -5,18 +5,24 @@ use ecow::{EcoString, eco_format};
...
 
-use crate::SVGRenderer;
+use crate::{SVGRenderer, SvgMatrix};
 
 impl SVGRenderer<'_> {
     /// Render an image element.
-    pub(super) fn render_image(&mut self, image: &Image, size: &Axes<Abs>) {
+    pub(super) fn render_image(
+        &mut self,
+        transform: &Transform,
```
laurmaedje: would be nice if this took `state` as well for consistency

crates/typst-svg/src/lib.rs
```diff
@@ -488,16 +491,37 @@ impl Display for SvgMatrix {
...
+}
 
-impl SvgPathBuilder {
-    fn with_scale(scale: Ratio) -> Self {
-        Self(EcoString::new(), scale)
+impl SvgRelativePathBuilder {
+    fn with_translate(pos: Point) -> Self {
+        // add initial M node to transform the entire path
+        Self {
+            path: EcoString::from(format!("M {} {}", pos.x.to_pt(), pos.y.to_pt())),
```
laurmaedje, line 507: If the initial point is zero, we could skip the `M 0 0` (I think?). this happens a lot in the tests.

crates/typst-svg/src/shape.rs
```diff
@@ -33,7 +32,12 @@ impl SVGRenderer<'_> {
             );
         }
 
-        let path = convert_geometry_to_path(&shape.geometry);
+        // reset transform because we applied it as element's transform attribute
+        // we kept it before for shape paint transform
+        self.xml.write_attribute("transform", &SvgMatrix(state.transform));
+        state = state.with_transform(Transform::identity());
+
+        let path = convert_geometry_to_path(&state.transform, &shape.geometry);
```
laurmaedje: The transform is always identity at this point, so we could remove this argument, right?

crates/typst-svg/src/shape.rs
```diff
@@ -33,7 +32,12 @@ impl SVGRenderer<'_> {
             );
         }
 
-        let path = convert_geometry_to_path(&shape.geometry);
+        // reset transform because we applied it as element's transform attribute
+        // we kept it before for shape paint transform
```
laurmaedje: This comment is hard to understand for me. Could you rewrite it? It would also be great if you could start comments with a capital letter and end with a full stop to keep it in line with the codebase's style.

crates/typst-svg/src/text.rs
```diff
@@ -22,7 +22,10 @@ impl SVGRenderer<'_> {
 
         self.xml.start_element("g");
         self.xml.write_attribute("class", "typst-text");
-        self.xml.write_attribute("transform", "scale(1, -1)");
+        self.xml.write_attribute(
+            "transform",
+            &format!("{} scale(1, -1)", &SvgMatrix(state.transform)),
```
laurmaedje: maybe we could directly integrate the scale into the transform?

crates/typst-svg/src/lib.rs
```diff
@@ -519,27 +543,92 @@ impl SvgPathBuilder {
...
+        Self {
+            path: Default::default(),
+            scale: Ratio::one(),
+            last_point: Point::zero(),
+        }
+    }
+}
+
+/// A builder for SVG path. This is used to build the path for a glyph.
+struct SvgGlyphPathBuilder(pub EcoString, pub Ratio);
```
laurmaedje: is there a particular reason why `SvgGlyphPathBuilder` and `SvgRelativePathBuilder` are separate? can glyphs not use the new, smarter builder?

crates/typst-svg/src/lib.rs
```diff
@@ -488,16 +491,61 @@ impl Display for SvgMatrix {
...
+struct SvgPathBuilder {
+    pub path: EcoString,
+    pub scale: Ratio,
+    pub last_close_point: Point,
+    pub last_point: Point,
+}
 
 impl SvgPathBuilder {
+    fn with_translate(pos: Point) -> Self {
+        if pos.x == Abs::pt(0.0) && pos.y == Abs::pt(0.0) {
```
saecki: Nit: this couild be `pos == Point::zero()` which is easier to read IMO

crates/typst-svg/src/lib.rs
Bequen: I did not wanted too broad changes. But you are right that there is no reason for two path builders. I applied the new one to glyphs too.

crates/typst-svg/src/shape.rs
Bequen: I decided to remove it entirely. When I removed the argument and the setting of the state to identity transform, it is more obvious.

crates/typst-svg/src/lib.rs
Bequen: I tried it and found out that you can't :( The SVG renderer needs the first M to set the origin point.

https://stackoverflow.com/questions/44170324/do-svg-path-d-elements-need-to-start-with-m-or-is-m-good-enough

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/Paths#line_commands

crates/typst-svg/src/lib.rs
Bequen: Good point. Unfortunately that code was incorrect and had to be removed entirely anyway :D

crates/typst-svg/src/lib.rs
```diff
@@ -293,12 +287,12 @@ impl<'a> SVGRenderer<'a> {
                 FrameItem::Shape(shape, _) => {
                     self.render_shape(state.pre_translate(*pos), shape)
                 }
-                FrameItem::Image(image, size, _) => self.render_image(image, size),
+                FrameItem::Image(image, size, _) => {
+                    self.render_image(&state.pre_translate(*pos), image, size)
+                }
                 FrameItem::Link(dest, size) => self.render_link(dest, *size),
```
laurmaedje: The link case that was recently added also needs to consider the position now.

crates/typst-svg/src/lib.rs
```diff
@@ -261,28 +260,23 @@ impl<'a> SVGRenderer<'a> {
             self.render_shape(state, &shape);
         }
 
-        self.render_frame(state, ts, &page.frame);
+        // frame is a hard frame -> apply transform
+        self.xml.start_element("g");
```
laurmaedje: isn't this group unnecessary when we have an identify transform? Previously, `render_frame` had a check for that.

crates/typst-svg/src/lib.rs
```diff
@@ -307,28 +301,37 @@ impl<'a> SVGRenderer<'a> {
...
             self.xml.write_attribute("data-typst-label", &label.resolve());
         }
 
         if let Some(clip_curve) = &group.clip {
             let hash = hash128(&group);
-            let id =
-                self.clip_paths.insert_with(hash, || shape::convert_curve(clip_curve));
+            let id = self
+                .clip_paths
+                .insert_with(hash, || shape::convert_curve(&state.transform, clip_curve));
```
laurmaedje: Is this sound? The clip_paths are memoized by hash, but the hash only includes the group, not `state.transform`. I also wonder why it hashes the whole group rather than just the path, but that's unrelated to your PR.

crates/typst-svg/src/shape.rs
```diff
@@ -166,13 +171,14 @@ fn convert_geometry_to_path(geometry: &Geometry) -> EcoString {
...
-        Geometry::Curve(p) => return convert_curve(p),
+        Geometry::Curve(p) => return convert_curve(&Transform::identity(), p),
     };
-    builder.0
+    builder.path
 }
 
-pub fn convert_curve(curve: &Curve) -> EcoString {
-    let mut builder = SvgPathBuilder::default();
+pub fn convert_curve(transform: &Transform, curve: &Curve) -> EcoString {
```
laurmaedje: I think it's a bit misleading that the entire `transform` is passed, but only the translation is used. I wonder whether the could even mask a bug for clip paths?

crates/typst-svg/src/image.rs
```diff
@@ -10,13 +10,19 @@ use typst_library::visualize::{
...
-    pub(super) fn render_image(&mut self, image: &Image, size: &Axes<Abs>) {
+    pub(super) fn render_image(
+        &mut self,
+        state: &State,
+        image: &Image,
+        size: &Axes<Abs>,
+    ) {
         let url = convert_image_to_base64_url(image);
         self.xml.start_element("image");
+        self.xml.write_attribute("transform", &SvgMatrix(state.transform));
```
laurmaedje: In the shape case there was a check against identity here. It would be nice if that check was consistently applied.

crates/typst-svg/src/shape.rs
Bequen: I agree, it's misleading. I changed it to `initial_point` and the data type to `Point` to clearly tell it's purpose.

It does not require the entire transform, because it only uses the point for `Z` tag in path, to know to which point to return. The rotation, scaling, and rest of the transformation is done with `transform` attribute in the `path` element.

Bequen:
Hello, everything is fixed. Sorry for the delay, I forgot :D

laurmaedje:
Thanks!

Issue #5762, opened 2025-01-26
Make compression with `pdf.embed` optional


TS53:
### Description

Currently, `pdf.embed` will always compress the contents of the embedded data.
Some contents are already highly compressed like mp3, jpg, zip, and many more.
Embedding such a file with compression will likely increase the file size and the compilation time.

Thus, I suggest to make the compression optional, setting it to `true` by default.

Related: https://github.com/typst/typst/pull/5221

### Use Case

Decrease file size and compilation time for specific file formats.

TS53:
This sound good for a default `auto` option, but I would still allow a parameter to turn it off. For example in a debug mode, the user is interested in fast compilation. On the other hand, JPG and ZIP would probably get uncompressed using the magic bytes method, but both could have compression turned off or nearly off. It really depends on the use case.

laurmaedje:
I can appreciate that there are rare cases where it could be valuable to have more control, but I am of the opinion that every single toggle must pull its weight and for this one, I think the benefit is too small.

- For the debug mode argument: That's not specific to `pdf.embed`, so I think it would be better as a compilation-wide setting. Even then, I'm not sure how much I like it. (Would it be default for `typst watch`? If no, what's the point, but if yes, typst watch output now differs from typst compile output). In the end, for the very fastest previews, you just don't want to use PDF export at all.

- The JPG/ZIP thing really seems like a very rare edge case to me.

saecki:
The linked PR implements disabling compression based on the file's magic bytes.
I think running the compression to determine if it actually helped, would require either:
- Running compression twice, once in typst and then a second time in krilla when actually writing the data.
- Modifying krilla to accept compressed streams or implement the detection directly

laurmaedje:
@LaurenzV do you think all of this should live in Typst or do you think some sort of smartness would be a good fit for krilla?

LaurenzV:
I think I would be fine with having this!

saecki:
The automatic compression has been merged with #6488. Does this close the issue?

PR #6256, opened 2025-05-08
Use the infer crate to determine if pdf embeds should be compressed

saecki:
Implements the first step of #5762. Uses the `infer` crate to roughly filter which file types should be compressed when embedded into pdfs.

### DIFF
--- a/crates/typst-pdf/src/embed.rs
+++ b/crates/typst-pdf/src/embed.rs
@@ -56,7 +56,7 @@ pub(crate) fn embed_files(
 
 fn should_compress(data: &[u8]) -> bool {
     let Some(ty) = infer::get(data) else {
-        return false;
+        return true;
     };
     match ty.matcher_type() {
         infer::MatcherType::App => true,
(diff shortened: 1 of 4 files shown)


TS53:
`audio/midi` should get compressed. Many BMP images are also uncompressed and can get good compression results.

saecki:
Oh, right I was a bit quick to draw. If the type could not be resolved I think we should also default to compressing the output. And maybe for those `MatcherType`s that we might not want to compress there should be a opt out list based on the mime type.

crates/typst-pdf/src/embed.rs
```diff
@@ -34,14 +34,15 @@ pub(crate) fn embed_files(
             },
         };
         let data: Arc<dyn AsRef<[u8]> + Send + Sync> = Arc::new(embed.data.clone());
+        let compress = should_compress(&embed.data);
```
saecki, line 37: TODO: add todo

PR #6259, opened 2025-05-09
Catch indefinite loop in realization due to cycle between show and grouping rule

### LINKED ISSUE #5690: Out of memory error with "show par: box"

saecki:
Fixes #5690. The wording of the error can probably be improved, all tough I'm not sure what terms to use for explaining grouping rules.

### DIFF
--- a/crates/typst-realize/src/lib.rs
+++ b/crates/typst-realize/src/lib.rs
@@ -655,6 +655,7 @@ fn visit_grouping_rules<'a>(
     let matching = s.rules.iter().find(|&rule| (rule.trigger)(content, &s.kind));
 
     // Try to continue or finish an existing grouping.
+    let mut i = 0;
     while let Some(active) = s.groupings.last() {
         // Start a nested group if a rule with higher priority matches.
         if matching.is_some_and(|rule| rule.priority > active.rule.priority) {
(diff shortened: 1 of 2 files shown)


crates/typst-realize/src/lib.rs
```diff
@@ -670,6 +671,21 @@ fn visit_grouping_rules<'a>(
         }
 
         finish_innermost_grouping(s)?;
+        i += 1;
+        if i > 4096 {
```
saecki, line 675: Maybe the limit is too small for complex documents?

laurmaedje: In which sane cases would it be hit?

saecki: I couldn't really come up with one, so I guess it *should* be fine

saecki: So I've reduced it to 512, because I couldn't really find any real world cases where this would go above like 3, and otherwise the stack will overflow in a test.

laurmaedje: is this, in sane cases, upper bounded by the number of different grouping rules essentially? hence 3?

saecki: So I did some more digging, and I've found this test case that goes up to `24`, and increases the longer the text and words get:
```typ
// --- show-text-regex-character-class ---
// This is a fun one.
#set par(justify: true)
#show regex("\S"): letter => box(stroke: 1pt, inset: 2pt, upper(letter))
#lorem(5)

// trailing blank line is important!
```

saecki: But I guess these cases are so slow and consume so much memory, that it's sensible to have an upper bound for them.

Issue #5908, opened 2025-02-20
RFC: Vertical Writing Mode


OverflowCat:
## Introduction

Vertical layout remains vital across various cultures and contexts,
particularly in traditional East Asian typesetting.
This RFC proposes the implementation of vertical writing mode support in Typst.

### Vertical writing

![](https://www.w3.org/TR/jlreq/images/img1_13.png)

![Image](https://github.com/user-attachments/assets/7269f073-1723-432a-8180-6a15225f6ffa)

And figure placement, accordingly:

![](https://www.w3.org/TR/jlreq/images/img1_14.png)

### Latin & formulae in vertical text

All block formulas are rotated 90° to the right of the entire formula block. Alphanumeric characters and mathematical symbols are all folded sideways.

![mn](https://github.com/user-attachments/assets/2cb7d241-99b3-4787-af93-5b3ee8d7bb41)

![mn](https://github.com/user-attachments/assets/7784cbde-46fb-4c63-b7d2-0a6db4da1e4d)

Short inline math that fits into 1x line-height(?) can be upright (tate-chu-yoko).

![ja](https://github.com/user-attachments/assets/ea9f3984-04e1-4963-a01c-e563fa7eea9b)

### Tate-Chu-Yoko (縦中横)

![](https://www.w3.org/TR/jlreq/images_ja/img1_21.png)

Compression:

![Image](https://github.com/user-attachments/assets/d3fe5dbb-c000-4d61-8031-bdd139aeba53)

### Punctuations

![](https://upload.wikimedia.org/wikipedia/commons/2/23/Hengpai-shupai.png)

> Punctuation marks used in Taiwan and Hong Kong are usually positioned in the vertical and horizontal center of the square space left for them; while in vertical writing mode and horizontal writing mode, some of the punctuation marks are positioned in different directions so as to mark the corresponding characters more accurately. In Mainland China, the punctuation marks are usually positioned following the characters they are supposed to mark; while some punctuation marks might be positioned in different directions due to the vertical or horizontal writing mode. Also, different writing modes might require different punctuation marks to fulfill the same function, e.g. horizontal writing mode requires curved quotation marks while vertical writing mode requires corner brackets.

## Proposed Solution

The core of this proposal is to extend Typst's text model and layout engine to accommodate vertical writing modes, providing a clear API for controlling and customizing text direction.

#### `writing-mode` text attr

[...]

PR #6266, opened 2025-05-10
Add remaining height example for layout

Andrew15-5:
Today I wanted to get the remaining page height, so I just used `layout(size => size.height)`, but apparently it doesn't compensate for any added on the page content, it will show the same page height regardless. But thankfully I do remember the quirky `1fr` thingy, that sometimes help a lot, and this is one of those cases. Just wrap it in block with fractional height and you are good to go.

Use of fractions is probably not easy for new users, so this example should be very helpful.

### DIFF
--- a/crates/typst-library/src/layout/layout.rs
+++ b/crates/typst-library/src/layout/layout.rs
@@ -41,6 +41,23 @@ use crate::layout::{BlockElem, Size};
 /// receives the page's dimensions minus its margins. This is mostly useful in
 /// combination with [measurement]($measure).
 ///
+/// If you want to, for example, get the remaining of the page height after some
+/// content was already added to the page, wrap the `layout` call in
+/// `block(height: 1fr)`. But note that adding multiple blocks with height of
+/// `1fr` together will put them all on the same page, which can be fixed with
+/// [`pagebreak`].
+///
+/// ```example
+/// #lorem(10)
+/// #block(height: 1fr, layout(size => {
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/layout/layout.rs
```diff
@@ -41,25 +41,23 @@ use crate::layout::{BlockElem, Size};
 /// receives the page's dimensions minus its margins. This is mostly useful in
 /// combination with [measurement]($measure).
 ///
-/// If you want to, for example, get the remaining of the page height after some
-/// content was already added to the page, wrap the `layout` call in
-/// `block(height: 1fr)`. But note that adding multiple blocks with height of
-/// `1fr` together will put them all on the same page, which can be fixed with
-/// [`pagebreak`].
+/// To retrieve the _remaining_ size of the page rather than its full size, you
```
Andrew15-5, line 44: This doesn't immediately say that it's about height, unlike before. So this sounds confusing. I'd rather be more explicit. You also can't use `block(width: 1fr)`, only `box(width: 1fr)`.

Andrew15-5: > in a TTB layout.

Huh? Since when the layout is TTB? The layout is horizontal. https://github.com/typst/typst/issues/5908. The layout is first LTR/RTL and only then TTB.


> `_remaining_ height` would work as well.

Yes, this is what I'm saying.

PR #6306, opened 2025-05-20
Specify which CSL style is not suitable for bibliographies

Andrew15-5:
While stumbling into `"alphanumeric"` style for the nth time, I found that now apparently it errors that you can't use it for bibliography (it appears that I need to exclusively use it with `cite()`). But the error is vague, IMO.

```
CSL style is not suitable for bibliographies
```

Which style? It would be helpful to see more info. Initially I thought that this is better:

```
The CSL style is not suitable for bibliographies
Current CSL style is not suitable for bibliographies
```

But then I switched to

```
Current CSL style is not suitable for bibliographies: "name"
```

And finally ended up with

```
"name" CSL style is not suitable for bibliographies
```

This works for named styles and file paths. Though as always, I don't know what is the most efficient way to get X (string name) from Y (style field type), but it looks short enough.

```typ
#bibliography(bytes(""), style: "alphanumeric")
// #bibliography(bytes(""), style: "agora.csl")
```

Perhaps a hint can be added to point out other built-in styles that will work.

P.S. I in fact still haven't read the latest changelog...

### DIFF
--- a/tests/suite/model/bibliography.typ
+++ b/tests/suite/model/bibliography.typ
@@ -71,6 +71,10 @@ Now we have multiple bibliographies containing @glacier-melt @keshav2007read
 
 #bibliography("/assets/bib/works_too.bib", style: "mla")
 
+--- bibliography-style-not-suitable ---
+// Error: 2-62 CSL style "Alphanumeric" is not suitable for bibliographies
+#bibliography("/assets/bib/works.bib", style: "alphanumeric")
+
 --- issue-4618-bibliography-set-heading-level ---
 // Test that the bibliography block's heading is set to 2 by the show rule,
 // and therefore should be rendered like a level-2 heading. Notably, this


crates/typst-library/src/model/bibliography.rs
```diff
@@ -224,7 +224,12 @@ impl Show for Packed<BibliographyElem> {
         let references = works
             .references
             .as_ref()
-            .ok_or("CSL style is not suitable for bibliographies")
+            .ok_or_else(|| {
+                eco_format!(
+                    "{:?} CSL style is not suitable for bibliographies",
```
laurmaedje: Using `{:?}` debug formatting here is problematic as it fill use Rust-style formatting (e.g. `Bytes(10)`). I think we should instead only print a name if we know it (i.e. for named styles via `style.names()`). I would also reorder this, so that the final message is 
```rust
eco_format!("CSL style `{}` is not suitable for bibliographies", name.repr())
```
If there is no name, we can keep the old message.

Could you also add tests for both cases?

Andrew15-5: Where does `.repr()` come from?

```
no method named `repr` found for enum `hayagriva::archive::ArchivedStyle` in the current scope items from traits can only be used if the trait is implemented and in scope [E0599]
```

I don't know where does `.names()` gets multiple names if we can only specify one. There is also `.display_name()`. Why not use it?

I also don't know how to add test for this because of unnamed variant and because of plurality. Is it a pure Typst or Rust test?

laurmaedje: Display name is also good.

laurmaedje:
Thanks!

crates/typst-library/src/model/bibliography.rs
Andrew15-5: In the end, the display name is capitalized, which does not match the specified style name. I wanted it to copy what user has set.

laurmaedje: That would have been names()[0]

Andrew15-5: But why there are multiple?

laurmaedje: Some have an alias

laurmaedje: e.g.  "public-library-of-science" and "plos"

Andrew15-5: Is there a way to get the actual user-set string value? It gets converted so hard so quickly.

https://github.com/typst/typst/blob/98034903e4184708dbaf3cf6b23a4f4de5e910ec/crates/typst-library/src/model/bibliography.rs?plain=1#L137-L140

laurmaedje: No, it's not possible, it will always become an enum variant internally

Andrew15-5: Damn... So, would the display name be better off than a guessed first alias?

laurmaedje: I personally don't think it matters much. I think what we merged is totally fine.

PR #6307, opened 2025-05-20
Always run tests from workspace directory

frozolotl:
This changes the equivalent of `cd ..` to `cd $CARGO_MANIFEST_DIR/..`.
So, when tests are run in a different directory than expected, things still work fine.


laurmaedje:
Thanks!

Issue #6030, opened 2025-03-09
Poor error message for invalid bibliography when inlined


jeanas:
### Description

When the bibliography YAML code is inlined into the document with `#bibliography(bytes("..."))`, errors in the YAML schema are not reported with an informative error message. Instead the compiler reports "label ... does not exist in the document".

For example, consider the code

```typst
This is investigated in @foo.

#bibliography("bib.yml")
```

where `bib.yml` contains:

```yaml
foo:
  type: article
  author: John Doe
  title: Bazing the Foo
  date: March 2009
```

This produces the message

```
error: failed to parse YAML (foo.date: date format unknown at line 5 column 9)
  ┌─ bug.typ:1:14
  │
1 │ #bibliography("bib.yml")
  │               ^^^^^^^^^
```

But on the other hand, consider:

```typst
This is investigated in @foo.

#bibliography(bytes("
foo:
  type: article
  author: John Doe
  title: Bazing the Foo
  date: March 2009
"))
```

This produces:

```
error: label `<foo>` does not exist in the document
  ┌─ bug2.typ:1:24
  │
1 │ This is investigated in @foo.
  │                         ^^^^
```

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
When getting bytes, the bibliography parser needs to guess whether it's Hayagriva or BibLaTeX. It does that by checking whether hayagriva parses successfully, then bibtex parses successfully, and if neither, it looks whether there are more `:` or `{` to guess the format.

In this case hayagriva doesn't parse successfully and there are more `:`, so it seems that BibLaTeX parses successfully. I don't have the spec in mind, but maybe the parser is too permissive.

jeanas:
Umm, wouldn't it be better to be explicit, i.e., `#bibliography(yaml: "...")` or `#bibliography(bibtex: "...")`?

laurmaedje:
That API would be inconsistent with other functions like `image`. I guess we could have `bibliography(.., format: auto | "hayagriva" | "biblatex")`, but even then we would have an `auto` option.

luewolf:
Hayagriva is just YAML with restrictions regarding property names and data types, isn't it? Perhaps a generic YAML parser could be used to distinguish between invalid hayagriva and biblatex.

[...]

PR #6308, opened 2025-05-20
Report errors in external files

### LINKED ISSUE #6175: Detect initial error instead of derivative one in context when layout does not converge
### Description

```typ
#context {
  let c = counter(figure.where(kind: image))
  let n = c.get().first()
  [#figure[]<a>]
  // c.update(_ => n + 1)
  c.update(_ => () + 1)
}
@a
```

gives

```
warning: layout did not converge within 5 attempts
 = hint: check if any states or queries are updating themselves

error: label `<a>` does not exist in the document
  ┌─ file.typ:8:0
  │
8 │ @a
  │ ^^
```

But if I comment unused `n` variable creation, _then_ it gives the initial error that caused the one above:

```
error: cannot add array and integer
  ┌─ file.typ:6:16
  │
6 │   c.update(_ => () + 1)
  │                 ^^^^^^
```

Which is really weird, since `n` is not used and therefore shouldn't affect anything, but with some context voodoo magic it does. This is a really bad UX, IMO.

Since closure body is not evaluated until it's called, I assume the label not fount error was hit earlier than the addition error, or maybe because the layout didn't converge the initial error just disappeared in the end or something like that.

This made debugging this kind of hard, since errors should be helping, but in this case it threw me away in a completely different direction from what I had to fix. I think Typst should try to prioritize helpful messages (so that we do not end up with C++ template level errors).

But I try to think what should be done here. Maybe a static analysis? Const evaluation? Expression optimizations? It feels very easily solvable if you have a constant expression that can be validated before executing anything. But if I write like I did at the start of debugging:

```typ
#context {
  let c = counter(figure.where(kind: image))
  let n = c.get()
  [#figure[]<a>]
  c.update(_ => n + 1)
}
@a
```

then it no longer static expression and it shows the same derivative error. So the only solution I see here is to `.peek()` into the closure body with known values and validate that as if it was const expression. But then again, it must be ignoring the error that it already detected, and keeps ignoring for 5 times and then just completely forgets about it. So maybe some priority needs to be changed or just keep a list of previous errors, and if you get error after not converging, you just print both initial and derivative error.

### Use Case

Better UX.

saecki:
This builds on the work from #5415 (thank you @Dherse :heart:), and closes #6175.

The PR changes a lot of things and introduces a few new types to deal with errors that might have occurred inside the file of a `DataSource::Path`.

The `Loaded` struct contains a `LoadSource` which is analogous to a loaded `DataSource` and only stores the `FileId` of a path for error reporting.
Then there are `LoadResult` and `LoadError` to address [this comment](https://github.com/typst/typst/pull/5415/files#r1875845487). Which can be turned into a `SourceDiagnostic` after the fact, to avoid polluting the memoization with `FileId`s and `Span`s. `LoadError` is currently 56 bytes large (on 64 bit architectures), maybe it should be internally wrapped in an `Arc`?

The `lines` metadata from `Source` has been factored out into a generalized `Lines` struct which is reused when dealing with `SourceDiagnostic` that reference non typst files, such as when loading `yaml`, `json`, etc.

(All dependencies added are already present in the dependency tree)

Also improves #6030, by rejecting empty bibtex files if the format was chosen implicitly.

### DIFF
--- a/crates/typst-layout/src/image.rs
+++ b/crates/typst-layout/src/image.rs
@@ -1,6 +1,6 @@
 use std::ffi::OsStr;
 
-use typst_library::diag::{warning, At, SourceResult, StrResult};
+use typst_library::diag::{warning, At, LoadedAt, SourceResult, StrResult};
 use typst_library::engine::Engine;
 use typst_library::foundations::{Bytes, Derived, Packed, Smart, StyleChain};
 use typst_library::introspection::Locator;
@@ -30,14 +30,14 @@ pub fn layout_image(
     let Derived { source, derived: data } = &elem.source;
     let format = match elem.format(styles) {
         Smart::Custom(v) => v,
-        Smart::Auto => determine_format(source, data).at(span)?,
+        Smart::Auto => determine_format(source, &data.bytes).at(span)?,
     };
 
     // Warn the user if the image contains a foreign object. Not perfect
     // because the svg could also be encoded, but that's an edge case.
     if format == ImageFormat::Vector(VectorFormat::Svg) {
         let has_foreign_object =
-            data.as_str().is_ok_and(|s| s.contains("<foreignObject"));
+            memchr::memmem::find(&data.bytes, b"<foreignObject").is_some();
 
         if has_foreign_object {
             engine.sink.warn(warning!(
@@ -53,7 +53,7 @@ pub fn layout_image(
     let kind = match format {
         ImageFormat::Raster(format) => ImageKind::Raster(
             RasterImage::new(
-                data.clone(),
+                data.bytes.clone(),
                 format,
                 elem.icc(styles).as_ref().map(|icc| icc.derived.clone()),
             )
@@ -61,11 +61,11 @@ pub fn layout_image(
         ),
         ImageFormat::Vector(VectorFormat::Svg) => ImageKind::Svg(
             SvgImage::with_fonts(
-                data.clone(),
+                data.bytes.clone(),
                 engine.world,
                 &families(styles).map(|f| f.as_str()).collect::<Vec<_>>(),
             )
-            .at(span)?,
+            .in_text(&data)?,
         ),
     };
 
--- a/crates/typst-layout/src/image.rs
+++ b/crates/typst-layout/src/image.rs
@@ -65,7 +65,7 @@ pub fn layout_image(
                 engine.world,
                 &families(styles).map(|f| f.as_str()).collect::<Vec<_>>(),
             )
-            .in_text(&data)?,
+            .in_text(data)?,
         ),
     };
 
--- a/crates/typst-library/src/diag.rs
+++ b/crates/typst-library/src/diag.rs
@@ -151,7 +151,7 @@ pub struct Warned<T> {
     pub warnings: EcoVec<SourceDiagnostic>,
 }
 
-/// An error or warning in a source file.
+/// An error or warning in a source or text file.
 ///
 /// The contained spans will only be detached if any of the input source files
 /// were detached.
@@ -574,8 +574,11 @@ impl From<PackageError> for EcoString {
 pub type LoadResult<T> = Result<T, LoadError>;
 
 /// A callsite independent error that occurred during data loading.
+/// This avoids polluting the memoization with [`Span`]s and [`FileId`]s from source files.
 /// Can be turned into a [`SourceDiagnostic`] using the [`LoadedAt::in_text`]
 /// or [`LoadedAt::in_invalid_text`] methods available on [`LoadResult`].
+///
+/// [`FileId`]: typst_syntax::FileId
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct LoadError {
     pub pos: ReportPos,
--- a/crates/typst-library/src/diag.rs
+++ b/crates/typst-library/src/diag.rs
@@ -6,6 +6,7 @@ use std::path::{Path, PathBuf};
 use std::str::Utf8Error;
 use std::string::FromUtf8Error;
 
+use az::SaturatingAs;
 use comemo::Tracked;
 use ecow::{eco_vec, EcoVec};
 use typst_syntax::package::{PackageSpec, PackageVersion};
@@ -680,11 +681,11 @@ impl Loaded {
 
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
 pub enum ReportPos {
-    /// Contains the range, and the 0-based line/column.
-    Full(std::ops::Range<usize>, LineCol),
-    /// Contains the range.
-    Range(std::ops::Range<usize>),
-    /// Contains the 0-based line/column.
+    /// Contains a range, and a line/column pair.
+    Full(std::ops::Range<u32>, LineCol),
+    /// Contains a range.
+    Range(std::ops::Range<u32>),
+    /// Contains a line/column pair.
     LineCol(LineCol),
     #[default]
     None,
@@ -692,7 +693,7 @@ pub enum ReportPos {
 
 impl From<std::ops::Range<usize>> for ReportPos {
     fn from(value: std::ops::Range<usize>) -> Self {
-        Self::Range(value)
+        Self::Range(value.start.saturating_as()..value.end.saturating_as())
     }
 }
 
@@ -703,12 +704,18 @@ impl From<LineCol> for ReportPos {
 }
 
 impl ReportPos {
+    pub fn full(range: std::ops::Range<usize>, pair: LineCol) -> Self {
+        let range = range.start.saturating_as()..range.end.saturating_as();
+        Self::Full(range, pair)
+    }
+
     fn range(&self, lines: &Lines<String>) -> Option<std::ops::Range<usize>> {
         match self {
-            ReportPos::Full(range, _) => Some(range.clone()),
-            ReportPos::Range(range) => Some(range.clone()),
+            ReportPos::Full(range, _) => Some(range.start as usize..range.end as usize),
+            ReportPos::Range(range) => Some(range.start as usize..range.end as usize),
             &ReportPos::LineCol(pair) => {
-                let i = lines.line_column_to_byte(pair.line, pair.col)?;
+                let i =
+                    lines.line_column_to_byte(pair.line as usize, pair.col as usize)?;
                 Some(i..i)
             }
             ReportPos::None => None,
@@ -719,7 +726,7 @@ impl ReportPos {
         match self {
             &ReportPos::Full(_, pair) => Some(pair),
             ReportPos::Range(range) => {
-                let (line, col) = lines.byte_to_line_column(range.start)?;
+                let (line, col) = lines.byte_to_line_column(range.start as usize)?;
                 Some(LineCol::zero_based(line, col))
             }
             &ReportPos::LineCol(pair) => Some(pair),
@@ -732,7 +739,9 @@ impl ReportPos {
     fn try_line_col(&self, bytes: &[u8]) -> Option<LineCol> {
         match self {
             &ReportPos::Full(_, pair) => Some(pair),
-            ReportPos::Range(range) => LineCol::try_from_byte_pos(range.start, bytes),
+            ReportPos::Range(range) => {
+                LineCol::try_from_byte_pos(range.start as usize, bytes)
+            }
             &ReportPos::LineCol(pair) => Some(pair),
             ReportPos::None => None,
         }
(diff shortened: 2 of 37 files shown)


crates/typst-library/src/diag.rs
```diff
@@ -568,31 +571,270 @@ impl From<PackageError> for EcoString {
...
+pub type LoadResult<T> = Result<T, LoadError>;
+
+/// A call site independent error that occurred during data loading.
+/// This avoids polluting the memoization with [`Span`]s and [`FileId`]s from source files.
+/// Can be turned into a [`SourceDiagnostic`] using the [`LoadedAt::in_text`]
+/// or [`LoadedAt::in_invalid_text`] methods available on [`LoadResult`].
+///
+/// [`FileId`]: typst_syntax::FileId
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct LoadError {
```
saecki, line 583: `LoadError` is currently 56 bytes large (on 64 bit architectures), maybe it should be internally wrapped in an `Arc`?

crates/typst-layout/src/image.rs
```diff
@@ -30,14 +30,14 @@ pub fn layout_image(
     let Derived { source, derived: data } = &elem.source;
```
laurmaedje:
```suggestion
    let Derived { source, derived: loaded } = &elem.source;
```

crates/typst-library/src/foundations/plugin.rs
```diff
@@ -152,7 +152,7 @@ pub fn plugin(
     source: Spanned<DataSource>,
 ) -> SourceResult<Module> {
     let data = source.load(engine.world)?;
-    Plugin::module(data).at(source.span)
+    Plugin::module(data.bytes).at(source.span)
```
laurmaedje: I think I would also prefer `let loaded = source.load(engine.world)?` and `Plugin::module(loaded.data)` here, also renaming the field of `Loaded`.

crates/typst-library/src/loading/csv.rs
```diff
@@ -164,15 +164,27 @@ cast! {
 }
 
 /// Format the user-facing CSV error message.
-fn format_csv_error(err: ::csv::Error, line: usize) -> EcoString {
+fn format_csv_error(
+    data: &Loaded,
+    err: ::csv::Error,
+    line: usize,
+) -> EcoVec<SourceDiagnostic> {
```
laurmaedje: wouldn't it be cleaner if this also returned a `LoadError`, so that everything is uniform instead of having two different setups depending on whether memoization is performed. I think the performance impact is negligible since this is off the hot path.

crates/typst-library/src/loading/mod.rs
```diff
@@ -115,6 +120,48 @@ impl Load for Spanned<&OneOrMultiple<DataSource>> {
...
+            let end = start + err.error_len().unwrap_or(0);
+            LoadError::new(
+                start..end,
+                "failed to convert to string",
+                "file is not valid utf-8",
+            )
+        })
+    }
+}
+
```
laurmaedje: This here unfortunately disables the optimization that bytes created from a string aren't UTF-8 revalidated (possible via `bytes.as_str()`) and aren't copied when converted to a `Str` (via `bytes.to_str()`).

crates/typst-library/src/model/bibliography.rs
```diff
@@ -351,82 +355,86 @@ impl Debug for Bibliography {
 }
 
 /// Decode on library from one data source.
-fn decode_library(source: &DataSource, data: &Bytes) -> StrResult<Library> {
-    let src = data.as_str().map_err(FileError::from)?;
+fn decode_library(data: &Loaded) -> SourceResult<Library> {
+    let str = data.load_str()?;
```
laurmaedje: I try to avoid the name `str` for variables as it's always highlighted like a type.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -351,82 +355,86 @@ impl Debug for Bibliography {
...
-    let Some(error) = errors.first() else {
-        return match path {
-            Some(path) => eco_format!("failed to parse BibLaTeX file ({path})"),
-            None => eco_format!("failed to parse BibLaTeX"),
-        };
+) -> EcoVec<SourceDiagnostic> {
+    // TODO: return multiple errors?
+    let Some(error) = errors.into_iter().next() else {
+        // TODO: can this even happen, should we just unwrap?
+        return data.err_in_text(ReportPos::None, "failed to parse BibLaTeX", "???");
```
laurmaedje: probably can't happen, but having the `"???"` here is a bit unfortunate

crates/typst-syntax/src/lines.rs
```diff
@@ -0,0 +1,407 @@
...
+use crate::is_newline;
+
+/// A text buffer and metadata about lines.
+#[derive(Clone)]
+pub struct Lines<S>(Arc<Repr<S>>);
+
+#[derive(Clone)]
+struct Repr<S> {
+    lines: Vec<Line>,
+    str: S,
```
laurmaedje:
```suggestion
    text: S,
```

crates/typst-syntax/src/lines.rs
```diff
@@ -0,0 +1,407 @@
...
+pub struct Line {
+    /// The UTF-8 byte offset where the line starts.
+    byte_idx: usize,
+    /// The UTF-16 codepoint offset where the line starts.
+    utf16_idx: usize,
+}
+
+impl<S: AsRef<str>> Lines<S> {
+    /// TODO: memoize this?
+    pub fn new(str: S) -> Self {
```
laurmaedje:
```suggestion
    pub fn new(text: S) -> Self {
```

crates/typst-syntax/src/lines.rs
```diff
@@ -0,0 +1,407 @@
...
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct Line {
+    /// The UTF-8 byte offset where the line starts.
+    byte_idx: usize,
+    /// The UTF-16 codepoint offset where the line starts.
+    utf16_idx: usize,
+}
+
+impl<S: AsRef<str>> Lines<S> {
+    /// TODO: memoize this?
```
laurmaedje: This probably shouldn't be a doc comment. Also, since we already memoize from_bytes and only `&Bytes` is cheaply hashable, the answer is probably no.

crates/typst-syntax/src/lines.rs
```diff
@@ -0,0 +1,407 @@
...
+    pub fn new(str: S) -> Self {
+        let lines = lines(str.as_ref());
+        Lines(Arc::new(Repr { lines, str }))
+    }
+
+    pub fn text(&self) -> &str {
```
laurmaedje, line 35: Please add a very brief doc comment to public methods even if it seems a bit pointless, just for consistency.

crates/typst-syntax/src/lines.rs
```diff
@@ -0,0 +1,407 @@
...
+            chars.next();
+        }
+        Some(range.start + (line.len() - chars.as_str().len()))
+    }
+}
+
+impl Lines<String> {
+    /// Tries to convert the bytes
+    #[comemo::memoize]
+    pub fn from_bytes(bytes: &[u8]) -> Result<Lines<String>, Utf8Error> {
```
laurmaedje: in principle, the idea was to take `&Bytes` here to make hashing and string conversion cheaper, but unfortunately that won't work since `Bytes` is defined in `typst-library`... :/

tests/src/collect.rs
```diff
@@ -359,11 +382,73 @@ impl<'a> Parser<'a> {
...
+        let text = match read(&path) {
+            Ok(text) => text,
+            Err(err) => {
+                self.error(err.to_string());
+                return None;
+            }
+        };
+
+        let start = self.parse_line_col()?;
+        let lines = Lines::from_bytes(text.as_ref()).expect("Errors shouldn't be annotated for files that aren't human readable (not valid utf-8)");
```
laurmaedje:
```suggestion
        let lines = Lines::from_bytes(text.as_ref()).expect(
            "errors shouldn't be annotated for files \
             that aren't human readable (not valid utf-8)",
        );
```

crates/typst-library/src/diag.rs
```diff
@@ -568,31 +571,270 @@ impl From<PackageError> for EcoString {
...
+    }
+}
+
+/// Convert a [`LoadResult`] to a [`SourceResult`] by adding the [`Loaded`] context.
+pub trait LoadedAt<T> {
+    /// Add the span information.
+    fn in_text(self, data: &Loaded) -> SourceResult<T>;
+
+    /// Add the span information.
+    fn in_invalid_text(self, data: &Loaded) -> SourceResult<T>;
```
laurmaedje: It wasn't immediately clear to me when to call what as a normal user of this infrastructure. It seems like `in_invalid_text` is only used in `load_str`. Maybe the trait should be renamed `LoadedWithin` and only expose a single `.within` function that just does the right thing?

in particular, the code still works if I replace the usage of `in_invalid_text` with `in_text`, so it's just a small perf optimization, right? I would probably 
- either opt for simpler single-function API instead 
- or at least not expose the optimized variant in this trait in a way that forces consumers of the API to make a decision 
- or, as a final option, document when to use what. :)

crates/typst-library/src/diag.rs
```diff
@@ -568,31 +571,270 @@ impl From<PackageError> for EcoString {
...
+    pub pos: ReportPos,
+    pub message: EcoString,
+    pub error: EcoString,
+}
+
+impl LoadError {
+    pub fn new(
+        pos: impl Into<ReportPos>,
+        message: impl std::fmt::Display,
+        error: impl std::fmt::Display,
```
laurmaedje, line 598: it's not immediately clear what the purpose of `message` vs `error` is. also, maybe `impl Into<EcoString>` would be preferrable in case we have already formatted a message to not clone it unnecessary.

crates/typst-library/src/diag.rs
laurmaedje: Since it's not passed through long chains of code, it's probably fine. Would need to measure to know.

crates/typst-library/src/diag.rs
saecki: Yes I considered that, but currently `message` only receives static str references, and `error` exclusively receives an error that only implements `Display` or a string literal (in `format_xml_like_error` is avoided the `new` function).
So at least for now there don't seem to be too many redundant clones, but I agree that this isn't optimal.

I also considered formatting the message inside the `LoadError` directly and then just pushing modifying the string when providing the context. This would reduce the size of `LoadError` and avoid some clones. It should also be quite easy to implement, because the only difference between the error messages is the location information inserted before the closing parenthesis of the error message.

When `LoadError` is used anyway we could remove the `err_in_text` and `err_in_invalid_text` methods on `Loaded` and implement it that way. Would you prefer something like that?

laurmaedje: So essentially formatting `{msg} ({error}` and then pushing `)` or something like ` at {line}:{col})` later? Sounds fine to me, but then the fields of `LoadError` should be private.

saecki: Yeah exactly

crates/typst-library/src/loading/mod.rs
saecki: I've changed `load_str` to be more or less equivalent to `to_str`. Is this what you had in mind?

crates/typst-library/src/model/bibliography.rs
saecki: If only you used a good editor...

crates/typst-syntax/src/lines.rs
laurmaedje: Maybe we can move this to a trait that is implement in `typst-library`.

crates/typst-library/src/foundations/bytes.rs
```diff
@@ -112,6 +112,21 @@ impl Bytes {
         }
     }
 
+    pub fn load_str(&self) -> LoadResult<&str> {
```
laurmaedje: In principle, this should rather use `.as_str()` internally as Bytes created from other types of strings than `Str` will then still profit from the optimization. But, what I would do instead is completely remove this function, use `.as_str()` instead and implement `From<Utf8Error>` for `LoadError` and then we can just rely on `?` auto-conversion. 

`Loaded::load_str` could also be removed and maybe `LoadedWithin` could even have a generic impl for `Into<LoadError>` just like `At`.

crates/typst-syntax/src/lines.rs
saecki: Implementing `TryFrom<&Bytes> for Lines` in `typst-library` worked.

crates/typst-library/src/diag.rs
saecki: I've renamed to trait to `LoadedWithin` which now only has one method called `within`

crates/typst-library/src/model/bibliography.rs
saecki: Is `"something went wrong"` more meaningful?

saecki:
Ok I think now the CI should pass. Had to rebase onto `main` because the `double_ended_iterator_last` clippy lint wasn't available in rust `1.85.0` which the CI previously used.

PR #6309, opened 2025-05-21
Move math styling to codex and add `math.scr`

### LINKED ISSUE #5853: Roundhand calligraphic function in math
### Description

Currently, the main way to insert calligraphic variants of Latin letters in math is through the [`math.cal`](https://typst.app/docs/reference/math/variants#functions-cal) function. 

There are two styles of calligraphic letters: _roundhand_ (also known as "script"; `\mathscr` in LaTeX), and _chancery_ (`\mathcal` in LaTeX). Unicode provides a single codepoint for each calligraphic letter, and it is up to the font to decide which style to use.

Some fonts may provide the other style as a stylistic set, but this is not standardized, and therefore cannot be used by default by Typst. Notwithstanding, Unicode provides a way to specify a style to use through standardized variation sequences for capital non-bold letters.[^1][^2] Sadly, this feature is not supported by many fonts. This issue mainly focuses on tracking font support for chancery and roundhand style calligraphic variation sequences. Discussions can happen in [the Math forge on Discord](https://discord.com/channels/1054443721975922748/1176478139757629563), and specific proposals can be submitted as separate issues.

[^1]: https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G82733
[^2]: https://www.unicode.org/Public/UCD/latest/ucd/StandardizedVariants.txt

### Use Case

Being able to easily typeset documents requiring the use of both calligraphic styles using Typst.

### Font support

Font support can be tested using the following piece of code:

```typ
#set text(font: "..")
\u{2112} \u{2112}\u{FE00} \u{2112}\u{FE01}
```

The first L should be of either style (for math fonts, this is typically chancery); the second one should be chancery; the third one should be roundhand.

Below is a list of mathematical fonts with relevant information. As an indication, I believe it should be trivial for fonts that already have both glyph styles (typically, one of them accessible through a stylistic set as mentioned above) to support the variation selectors. For other fonts, they have to create new glyphs, which requires some work.

- [ ] Cambria Math (tested with version 6.99, which seems to be the latest version as of writing this[^3])
- [ ] Fira Math
  - Does not support calligraphic letters at all.
  - https://github.com/firamath/firamath/issues/14
- [ ] IBM Plex Math
  - https://github.com/IBM/plex/issues/662
- [x] Lete Sans Math
- [ ] Libertinus Math
  - https://github.com/alerque/libertinus/issues/475
- [ ] Luciole Math
  - Does not even support lower case calligraphic letters.
- [x] New Computer Modern Math
  - Support for chancery variation sequences added in 7.04, which should be bundled with the next Typst version.
- [x] Noto Sans Math
- [ ] STIX Two Math
  - https://github.com/stipub/stixfonts/issues/187

[^3]: https://learn.microsoft.com/en-us/typography/font-list/cambria-math; https://learn.microsoft.com/en-us/typography/fonts/windows_11_font_list

### Open questions

- [x] Should we add native support for this Unicode feature even though most math fonts don't support it for now? This could indicate to the maintainers of those fonts that there is demand for this feature. But at the same time, we would likely get complaints from users that the function does not work. The extreme lack of font support should be clearly documented.
  - The latest version of New Computer Modern Math, Typst's default math font, now supports chancery variation sequences, so at least the function will work as expected with the default settings.
- [x] Either way, as [noted on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1338889228498763796), support for variation sequences in math might be missing "because its multiple characters and so ends up going through the normal text layout."
  - Solved by https://github.com/typst/typst/pull/6336.

mkorje:
This is the accompanying Typst PR for https://github.com/typst/codex/pull/20. All the math styling logic has been moved to codex (bar the Hebrew mappings and some CJK angle bracket mappings.)

There's only one user-facing thing that has changed. An improvement has been made to how styles fallback in math if the specified variant does not exist. The following example shows this best:

```typ
--- math-style-fallback ---
// Test how math styles fallback.
$upright(frak(bold(alpha))) = upright(bold(alpha)) \
 bold(mono(ϝ)) = bold(ϝ) \
 sans(Theta) = bold(sans(Theta)) \
 bold(upright(planck)) != planck \
 bb(e) != italic(bb(e)) \
 serif(sans(A)) != serif(A)$
```

| Old | New |
| - | - |
| math-style-fallback | math-style-fallback |

The codex PR adds support for Arabic as well as the roundhand and chancery variants of script style, but they require some more work on the Typst side before they can be integrated. The field `variant` on `EquationElem` has been changed to an `Option` to accommodate these features in the future.

Closes #5853

### DIFF
--- a/crates/typst-library/src/math/style.rs
+++ b/crates/typst-library/src/math/style.rs
@@ -106,7 +106,7 @@ pub fn cal(
 ///
 /// (The box is not conceptually necessary, but unfortunately currently needed
 /// due to limitations in Typst's text style handling in math.)
-#[func(title = "Script", keywords = ["mathscr", "roundhand"])]
+#[func(title = "Script Style", keywords = ["mathscr", "roundhand"])]
 pub fn scr(
     /// The content to style.
     body: Content,


crates/typst-layout/src/math/text.rs
```diff
@@ -177,127 +178,18 @@ fn layout_glyph(
     }
 }
 
-/// Style the character by selecting the unicode codepoint for italic, bold,
-/// caligraphic, etc.
```
wrzian, line 181: I'd like to keep this part of the comment if that's ok. It's useful context that the codepoint itself is being changed, even if it's implied by the function signature. It was confusing to me when I first saw this.

crates/typst-layout/src/math/text.rs
```diff
@@ -177,127 +178,18 @@ fn layout_glyph(
...
-        },
-
-        _ => unreachable!(),
-    };
-
-    std::char::from_u32(start + (c as u32 - base as u32)).unwrap()
+    to_style(c, resolve_style(c, variant, bold, italic)).next().unwrap()
 }
 
 fn basic_exception(c: char) -> Option<char> {
```
wrzian, line 195: Is there a reason this function is still on the Typst side as opposed to codex?

crates/typst-layout/src/math/text.rs
```diff
@@ -177,127 +178,18 @@ fn layout_glyph(
...
-            (Sans, true, _) => 0x1D7EC,
-            (Mono, _, _) => 0x1D7F6,
-            (Cal | Frak, _, _) => return c,
-        },
-
-        _ => unreachable!(),
-    };
-
-    std::char::from_u32(start + (c as u32 - base as u32)).unwrap()
+    to_style(c, resolve_style(c, variant, bold, italic)).next().unwrap()
```
wrzian, line 192: (unimportant nit) I would pull out the `resolve_style` call into a separate variable and I would move the basic exception return above the other style computations.

wrzian (commented):
Some small things I noticed, not a fine-toothed review. I'm in favor of this change generally. The styling code did feel a bit long-winded in this file to me.

crates/typst-layout/src/math/text.rs
mkorje: I didn't move it as it doesn't really fit under any of the 'styles'. The Hebrew letters I could move as I think you should be able to get back the RTL characters in math somehow (probably under the Italic style). As for the CJK angle brackets, I wasn't sure if those exceptions are even needed?

mkorje: Ok I've moved the Hebrew to codex, that way a user in Typst can get back the normal Hebrew glyph in math by doing `upright(alef)` (and have added this to the `math-style-hebrew-exceptions` test). While it isn't exactly "upright", it'll fit in with the Arabic styles when they get added. And I've just removed the angle brackets stuff then (the single ones aren't even CJK compatibility characters!).

mkorje:
`mathcal` and `mathscr` equivalents now working - contingent on font support, of course: ("NewerComputerModernMath" has the variation selectors done for the script capital C, I don't know how to do this programmatically so one is all you get. Also, NewCM Math doesn't have in the stylstic set the roundhand style for lowercase.)

```typ
#show math.equation: set text(font: "Noto Sans Math")

$
  cal(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z) \
  scr(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z) \
  cal(a b c d e f g h i j k l m n o p q r s t u v w x y z) \
  scr(a b c d e f g h i j k l m n o p q r s t u v w x y z) \
  bold(cal(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z)) \
  bold(scr(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z)) \
  bold(cal(a b c d e f g h i j k l m n o p q r s t u v w x y z)) \
  bold(scr(a b c d e f g h i j k l m n o p q r s t u v w x y z)) \
$

#show math.equation: set text(font: "NewerComputerModernMath")

$
  cal(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z) \
  scr(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z) \
  cal(a b c d e f g h i j k l m n o p q r s t u v w x y z) \
  scr(a b c d e f g h i j k l m n o p q r s t u v w x y z) \
  bold(cal(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z)) \
  bold(scr(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z)) \
  bold(cal(a b c d e f g h i j k l m n o p q r s t u v w x y z)) \
  bold(scr(a b c d e f g h i j k l m n o p q r s t u v w x y z)) \
$
```

test

mkorje:
~~Just to be clear, this is blocked by https://github.com/typst/typst/pull/6441, and of course needs https://github.com/typst/codex/pull/20 to be merged as well.~~ This is ready for a review now.

I'd appreciate some feedback on the docs for `math.cal` and `math.scr`. We want to make it abundantly clear that _very_ few fonts currently support these variation sequences, so we don't get a million questions and issues opened about it...

mkorje:
This now closes #5853

crates/typst-layout/src/math/text.rs
mkorje: With regards to removing those four angle bracket mappings, the [Unicode core spec](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-22/#G18694) has this to say:

<blockquote>

**_Mathematical Brackets._** The mathematical white square brackets, angle brackets, double angle brackets, and tortoise shell brackets encoded at U+27E6..U+27ED are intended for ordinary mathematical use of these particular bracket types. They are unambiguously narrow, for use in mathematical and scientific notation, and should be distinguished from the corresponding wide forms of white square brackets, angle brackets, and double angle brackets used in CJK typography. (See the discussion of the CJK Symbols and Punctuation block in [Section 6.2, General Punctuation](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-6/#G7981).) Note especially that the “bra” and “ket” angle brackets (U+2329 LEFT-POINTING ANGLE BRACKET and U+232A RIGHT-POINTING ANGLE BRACKET, respectively) are deprecated. Their use is strongly discouraged, because of their canonical equivalence to CJK angle brackets. This canonical equivalence is likely to result in unintended spacing problems if these characters are used in mathematical formulae.


The flattened parentheses encoded at U+27EE..U+27EF are additional, specifically-styled mathematical parentheses. Unlike the mathematical and CJK brackets just discussed, the flattened parentheses do not have corresponding wide CJK versions which they would need to be contrasted with.

</blockquote>

So I think we definitely _shouldn't_ have mappings from the CJK ones to the math ones. There's perhaps an argument to be made that we should keep the two mappings for "bra" and "ket", but because of their canonical equivalence to the CJK angle brackets I've decided to remove them.

Enivex:
We now have `scr`, `script`, and `scripts`, all doing different things.

Not saying this is necessarily a problem, but it's at least worth bringing up.

mkorje:
> We now have `scr`, `script`, and `scripts`, all doing different things.

I don't think its tooo bad. `script` and `scripts` maybe, but adding `scr` should be okay.

crates/typst-library/src/math/style.rs
```diff
@@ -59,23 +61,21 @@ pub fn sans(
...
 /// ```example
-/// Let $cal(P)$ be the set of ...
+/// $ scr(S) $
 /// ```
 ///
-/// This corresponds both to LaTeX's `\mathcal` and `\mathscr` as both of these
-/// styles share the same Unicode codepoints. Switching between the styles is
-/// thus only possible if supported by the font via
-/// [font features]($text.features).
+/// This is the default calligraphic/script style for most math fonts. See
```
mkorje: Maybe this statement isn't quite true? Unicode uses it as the default and mentions that it was used more often. Another thing to consider: this might just be making things unnecessarily confusing for users as the default math font _doesn't_ do this.

Enivex:
> > We now have `scr`, `script`, and `scripts`, all doing different things.
> 
> I don't think its tooo bad. `script` and `scripts` maybe, but adding `scr` should be okay.
> 

To be clear, I agree. I think the most reasonable solution (if we wanted to do anything) would be to rename either of those

crates/typst-library/src/math/style.rs
Enivex: What is "this" referring to here? The majority of math fonts default to a chancery presentation as far as I know.

Enivex: If it is claiming that roundhand is the default for most math fonts, then I do not believe that is true.

laurmaedje:
I must admit, I don't have a clear picture anymore in what state this is. There's been so much discussion. Is there anything that still needs input from me?

crates/typst-library/src/math/style.rs
mkorje: I've reversed this to fix it up now.

mkorje:
I think the main thing is some input on the new docs for `math.scr` and `math.cal`.

Otherwise, this is almost done. I'm just waiting on the next NewCM release (coming very soon) which brings support for the variation selectors needed for `math.scr` and `math.cal`.

crates/typst-library/src/math/style.rs
```diff
@@ -59,23 +61,37 @@ pub fn sans(
...
+    body.styled(EquationElem::set_variant(Some(MathVariant::Chancery)))
+}
+
+/// Script (roundhand) font style in math.
+///
+/// ```example
+/// $ scr(S) $
+/// ```
+///
+/// Very few math fonts currently support differentiating `cal` and `scr`. Some
```
laurmaedje: The docs don't make it totally clear that the fonts that support switching via feature are not the same as the ones from the first sentence (that support differentiating via variation selector IIUC). I think this could be expanded to clarify the two mechanisms and that one works out of the box, but only for few fonts, and how the other can be achieved manually.

mkorje: I've tried to make it clearer, how's that?

mkorje:
New version is out: https://github.com/typst/typst-assets/pull/11. This should be ready to merge after typst-assets gets updated.

mkorje:
I added a test. If you're happy with the docs this is ready to merge.

crates/typst-library/src/math/style.rs
```diff
@@ -59,23 +61,39 @@ pub fn sans(
...
+
+/// Script (roundhand) font style in math.
+///
+/// ```example
+/// $ scr(S) $
+/// ```
+///
+/// There are two ways that fonts can support differentiating `cal` and `scr`.
+/// The first is using Unicode variation sequences. This works out of the box
+/// in Typst, however only a few math fonts currently support this.
```
MDLC01, line 91: Maybe this could link to https://github.com/typst/typst/issues/5853. But also maybe this would be too much information.

crates/typst-library/src/math/style.rs
```diff
@@ -88,12 +106,12 @@ pub fn sans(
 ///
 /// (The box is not conceptually necessary, but unfortunately currently needed
 /// due to limitations in Typst's text style handling in math.)
-#[func(title = "Calligraphic", keywords = ["mathcal", "mathscr"])]
-pub fn cal(
+#[func(title = "Script", keywords = ["mathscr", "roundhand"])]
```
MDLC01:
```suggestion
#[func(title = "Script Style", keywords = ["mathscr", "roundhand"])]
```

This is probably better for clarity (`math.script` is "Script Size").

MDLC01 (commented):
Sorry for not leaving my comments sooner

crates/typst-library/src/math/style.rs
Enivex: Should we mention the roundhand and chancery terminology in the documentation?

MDLC01: I think we should, and it's already the case in the current state of this PR.

Enivex: Oh, I completely overlooked that

crates/typst-library/src/math/style.rs
mkorje: Yeah I think it might be too much information.

crates/typst-library/src/math/style.rs
mkorje: If `math.script` is Script Size already is adding "Style" really needed then?

MDLC01: There are already many cases where searching for a function in the reference yields multiple results with identical or similar names. I think we should try to disambiguate those as much as possible.

mkorje: Fair point, I've added that now.

laurmaedje:
Thank you!

PR #2925, opened 2023-12-12
Adding WebP support

Conando025:
I like using webp and as #2673 shows I'm not the only one.
And since the implementation was basically just adding the respective feature of the image crate,
so I went ahead and made a PR. 
If there is an Issue with this PR (I've never done this before) please let me know.

PR #6311, opened 2025-05-21
Add support for WebP images

LinusU:
This adds support for loading WebP images, the same way that any other images is loaded.

I'd be happy to add a test for this, but I think that we first need to add a WebP image to the assets package, similar to how it's done with "rhino.png" or "tiger.jpg". Also, I couldn't find any test for GIF files, so maybe the same should be done for a GIF file as well?

As can be seen in the diff over `Cargo.lock`, this doesn't add any new dependencies! `image-webp` is already included in the build (via `krilla`).

Some background on why I'm submitting this can be found here: https://github.com/typst/typst/pull/2925#issuecomment-2897588927

I have also made a version that gates this behind a feature flag, it can be found here: https://github.com/typst/typst/compare/main...LinusU:typst:webp

### DIFF
--- a/crates/typst-library/src/visualize/image/mod.rs
+++ b/crates/typst-library/src/visualize/image/mod.rs
@@ -77,8 +77,8 @@ pub struct ImageElem {
     /// [`source`]($image.source) (even then, Typst will try to figure out the
     /// format automatically, but that's not always possible).
     ///
-    /// Supported formats are `{"png"}`, `{"jpg"}`, `{"gif"}`, `{"svg"}` as well
-    /// as raw pixel data. Embedding PDFs as images is
+    /// Supported formats are `{"png"}`, `{"jpg"}`, `{"gif"}`, `{"svg"}`,
+    /// `{"webp"}` as well as raw pixel data. Embedding PDFs as images is
     /// [not currently supported](https://github.com/typst/typst/issues/145).
     ///
     /// When providing raw pixel data as the `source`, you must specify a
--- a/crates/typst-library/src/visualize/image/raster.rs
+++ b/crates/typst-library/src/visualize/image/raster.rs
@@ -9,6 +9,7 @@ use ecow::{eco_format, EcoString};
 use image::codecs::gif::GifDecoder;
 use image::codecs::jpeg::JpegDecoder;
 use image::codecs::png::PngDecoder;
+use image::codecs::webp::WebPDecoder;
 use image::{
     guess_format, DynamicImage, ImageBuffer, ImageDecoder, ImageResult, Limits, Pixel,
 };
@@ -77,6 +78,7 @@ impl RasterImage {
                     ExchangeFormat::Jpg => decode(JpegDecoder::new(cursor), icc),
                     ExchangeFormat::Png => decode(PngDecoder::new(cursor), icc),
                     ExchangeFormat::Gif => decode(GifDecoder::new(cursor), icc),
+                    ExchangeFormat::Webp => decode(WebPDecoder::new(cursor), icc),
                 }
                 .map_err(format_image_error)?;
 
@@ -242,6 +244,8 @@ pub enum ExchangeFormat {
     /// Raster format that is typically used for short animated clips. Typst can
     /// load GIFs, but they will become static.
     Gif,
+    /// Raster format that supports both lossy and lossless compression.
+    Webp,
 }
 
 impl ExchangeFormat {
@@ -257,6 +261,7 @@ impl From<ExchangeFormat> for image::ImageFormat {
             ExchangeFormat::Png => image::ImageFormat::Png,
             ExchangeFormat::Jpg => image::ImageFormat::Jpeg,
             ExchangeFormat::Gif => image::ImageFormat::Gif,
+            ExchangeFormat::Webp => image::ImageFormat::WebP,
         }
     }
 }
@@ -269,6 +274,7 @@ impl TryFrom<image::ImageFormat> for ExchangeFormat {
             image::ImageFormat::Png => ExchangeFormat::Png,
             image::ImageFormat::Jpeg => ExchangeFormat::Jpg,
             image::ImageFormat::Gif => ExchangeFormat::Gif,
+            image::ImageFormat::WebP => ExchangeFormat::Webp,
             _ => bail!("format not yet supported"),
         })
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1215,6 +1215,7 @@ dependencies = [
  "byteorder-lite",
  "color_quant",
  "gif",
+ "image-webp",
  "num-traits",
  "png",
  "zune-core",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ icu_provider_adapters = "1.4"
 icu_provider_blob = "1.4"
 icu_segmenter = { version = "1.4", features = ["serde"] }
 if_chain = "1"
-image = { version = "0.25.5", default-features = false, features = ["png", "jpeg", "gif"] }
+image = { version = "0.25.5", default-features = false, features = ["png", "jpeg", "gif", "webp"] }
 indexmap = { version = "2", features = ["serde"] }
 infer = { version = "0.19.0", default-features = false }
 kamadak-exif = "0.6"
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -841,7 +841,9 @@ fn param_value_completions<'a>(
 /// Returns which file extensions to complete for the given parameter if any.
 fn path_completion(func: &Func, param: &ParamInfo) -> Option<&'static [&'static str]> {
     Some(match (func.name(), param.name) {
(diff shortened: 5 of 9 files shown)


LinusU:
(force push to fix formatting using `cargo fmt --all`)

LinusU:
I've tested this in our app, and it works as expected 🎉 

```typst
#set page(width: 384pt, height: auto)

#text(size: 64pt)[Hello, World!]

#image("https://locousercontent.com/fOlMF5zs1R0FiO7YSnP-Uc-LEUtGLKTs3CnjCwulN7YWF8UUrnplFOJUeHgsgIpS/original.webp")
```

test

LinusU:
(force push to fix casing of WebP in docs)

Enivex:
Does this increase binary size?

LinusU:
> Does this increase binary size?

Before: **35M** (34,640,352 bytes)
After: **35M** (34,756,432 bytes)

Not noticeably, just by 0.3%.

> Webp is kind of a dead end that's being replaced by other formats. In particular jxl and avif.

I would love to use any image format that gives us better compression, but at the moment we use these images in a number of places, including the web & in React Native. Currently, support for JPEG XL is only in Safari. AVIF could potentially be something we could start adopting soon, but it would require us shipping native code in our apps since React Native currently doesn't support it.

Now, this is just my use case, and is of course something that you can say is outside the scope of this project. I'm just trying to describe why WebP matters to us.

Also, we have a lot of files uploaded in lossy WebP source format from our clients. Switching our current files would probably cause inflation in file size, or drop in quality.

> It can also not be natively embedded in any of typst's targets, with the exception of the experimental HTML

I'm not exactly sure how the output works today, but creating two documents using `typst compile`, one with a WebP and the other with a PNG file, produces almost identical outputs. I couldn't find a PNG header inside the PDF when looking with a Hex Editor, but maybe it's embedded as a PNG nonetheless...

In any case, this approach doesn't seem to be any worse than converting the images to PNG before loading them into Typst.

Again, speaking for my own use case, we are using the PNG rasterizer to render the output as an image.

We do want PDF export as well, but as long as the same pixels shows up inside the PDF, which seems to be the case, this is good enough for us. I'm guessing that the potential issue is that the file size might be bigger inside the PDF?

laurmaedje:
The PDF output will be the same as with a PNG as it is reencoded. The main reason against WebP support in Typst was compile time and dependency count, which is a bit of a moot point given our _current_ dependencies.
 (It's unlikely to change but possible). The other reason originally was to prevent a slippery slope where more and more image formats are requested if we give in once. Though that wouldn't necessarily apply here since the situation is different w.r.t. dependencies.

@reknih What do you think?

Enivex:
> I'm guessing that the potential issue is that the file size might be bigger inside the PDF?

Yes, they will be converted to png

reknih:
@laurmaedje asked me (the resident file format enjoyer at Typst) to comment on this:

We have been somewhat apprehensive about adding WebP in the past. Let me lay out our rationale here:

**Reasons to add WebP**

- Users may have WebP files that they may want to use
- When using HTML export, outputting WebP images might be desirable

**Reasons against WebP**

- Binary size increase of including WebP encoder/decoder. Since a dependency has already covertly added a decoder, this is somewhat of a moot point, at least on the decoding side.
- Attack / maintenance surface, same as above
- Including a WebP in PDF means that it needs to be recoded into a format PDF supports. Currently, Typst chooses either JPEG or PNG. This means that Typst has to choose between lossily recoding the file to JPEG or potentially producing a way bigger output file by creating a PNG. In my opinion, a smart implementation would divine the intent of the user by reading the WebP save quality parameter and choosing JPEG if it falls below about 95. Worth noting that PDF also supports JPEG 2000 which is more similar to WebP in that it supports both lossy and lossless compression at better compression rates than PNG and JPEG, but since no pure Rust encoder is available for that format, we cannot write it.
- This would be an image format that we'd include that does not feature a matching encoder. That means that for HTML export, we'd need to add additional dependencies to support scaling images down or cropping them. This also applies to GIF files, but for them, recoding as a PNG is always a win, for WebP not necessarily. The true win for WebP & HTML export would be to have an encoder so that we could convert JPEGs to smaller WebPs instead.
- WebP is not particularly more prolific than other image formats and somewhat [infamously unpopular](https://knowyourmeme.com/memes/webp). TIFF is widely used in professional workflows, if you have an iPhone camera, you deal with a lot of HEIC files, ... Adding WebP support could encourage image format proliferation.
- **Subjectivity Alert** In our opinion, WebP is a bad file format for assets in a design workflow. Its compression artifacts are more visible than JPEG's, starting at quality level 80 and below in many images. For example, [compare the quality 50 JPEG and WebP file on this page:](https://oatcookies.neocities.org/img/webp/) Even if the image differences seem to be similar or even less on the black comparison chart, the edges of the shadow in the top segment of the image are perceptively more affected by blocking artifacts in the WebP vs the JPEG, for just about ~15% less in file size. We want to help our users to deliver great-looking documents, and for this, standard settings of most JPEG encoders are better suited. If a reduction in file size is desired, we should instead do it in the export process using carefully chosen parameters.

Despite these negatives, I cautiously approve of this PR since including the encoder for SVGs only seems dumb. We're incurring the binary size, security, and maintenance burden anyways and including WebPs is feasible for determined users, it's just harder than it needs to be.

With that being said, I am not opposed to completely removing this dependency instead in the upcoming or a future release.

LinusU:
> Attack / maintenance surface, same as above

Not saying that this isn't a concern, but something that I think makes this easier to swallow is that the WebP encoder/decoder is maintained by the same organization ([`image-rs`](https://github.com/image-rs)) that maintains the `image` crate together with the png, jpeg, gif encoders/decoders that we already use. It's also 100% Rust, has [fuzz testing](https://github.com/image-rs/image-webp/tree/main/fuzz), and uses [`#![forbid(unsafe_code)]`](https://github.com/image-rs/image-webp/blob/e9a3fb462329f16d217793187e9b4778e1722c5d/src/lib.rs#L3C1-L3C24).

> Including a WebP in PDF means that it needs to be recoded into a format PDF supports.

This is true, but if you already have a WebP file, that is something that needs to happen if you want to put it in a PDF, whichever tool you use. E.g. for my use case the workaround would be to convert the images to PNG myself, which would have the exact same end result.

> WebP is not particularly more prolific than other image formats and somewhat [infamously unpopular](https://knowyourmeme.com/memes/webp).

Today I learned 😁 But it seems like the complaint was that when people tried to save images from websites they got a WebP file, which then their programs couldn't open? I guess this isn't as much of a problem today when tools support WebP, and even then, I think that the tradeoff less bandwidth for every visitor vs. some people have to jump through hoops to save images is an easy choice for most websites...

> This would be an image format that we'd include that does not feature a matching encoder. That means that for HTML export, we'd need to add additional dependencies to support scaling images down or cropping them. This also applies to GIF files, but for them, recoding as a PNG is always a win, for WebP not necessarily. The true win for WebP & HTML export would be to have an encoder so that we could convert JPEGs to smaller WebPs instead.

This could be something to include though. For the HTML export, it seems like it would be better to convert both GIF and PNG to WebP, instead of PNG. Since ["WebP lossless images are 26% smaller in size compared to PNGs"](https://developers.google.com/speed/webp), and [all browsers supports it](https://caniuse.com/webp).

> WebP is a bad file format for assets in a design workflow

I think that I agree with this. In a design workflow I imagine that you always want to work with the raw data and only perform lossless transformations until the absolute last step, where you compress it in order to deliver the final product.

I have no expertise to make me try and comment on the perceptual visual differences vs file size but at least the WebP team has a study linked that backs up their claim that WebP has smaller sizes when the end result looks the same: https://developers.google.com/speed/webp/docs/webp_study

> Despite these negatives, I cautiously approve of this PR since [...]

This makes me very happy to hear! 🙌 

Let me know if there is anything else I can do to help move this forward! 🚀

crates/typst-library/src/visualize/image/raster.rs
```diff
@@ -242,6 +244,8 @@ pub enum ExchangeFormat {
     /// Raster format that is typically used for short animated clips. Typst can
     /// load GIFs, but they will become static.
     Gif,
+    /// Modern raster format that supports both lossy and lossless compression.
```
laurmaedje: Not that modern anymore, so I think I'd just drop the first word.

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -77,7 +77,7 @@ pub struct ImageElem {
     /// [`source`]($image.source) (even then, Typst will try to figure out the
     /// format automatically, but that's not always possible).
     ///
-    /// Supported formats are `{"png"}`, `{"jpg"}`, `{"gif"}`, `{"svg"}` as well
+    /// Supported formats are `{"png"}`, `{"jpg"}`, `{"gif"}`, `{"svg"}`, `{"webp"}` as well
     /// as raw pixel data. Embedding PDFs as images is
```
laurmaedje: please wrap this at column 80

LinusU:
Adressed comments and rebased on latest `main` ✅

laurmaedje:
Thanks!

PR #6331, opened 2025-05-24
Change Russian secondary smart quotes & remove alternatives

Andrew15-5:
In Russian there are no standard/traditional alternative smart quotes.

The de facto primary quotes are «» (ёлочки), and secondary quotes are „“ (лапки).

https://ru.wikipedia.org/wiki/Кавычки#Вложенные_кавычки:

> В русском языке традиционно применяются французские «ёлочки», а для кавычек внутри кавычек и при письме от руки — немецкие „лапки“

https://rus.stackexchange.com/a/463380:

> [...] прослеживается такая традиция: «вложенные кавычки делать „лапками“». Если же „лапки“ недоступны, допускаются и «ёлочки», но тогда на стыке «вложенных «кавычек» принято отказываться от дублирования знака.

https://en.wikipedia.org/wiki/Quotation_mark#History:

> In Russia, Ukraine and Belarus, the curved quotation marks, „...“, are used as a secondary level or in handwriting, while the angular marks, «...», are used as the primary level on printed text. [citation needed]

Even though it says that Ukrainian uses the same thing, I'm not sure if I should touch it.

Cc @T0mstone @G1gg1L3s


MDLC01:
Wouldn't it be better to still have an alternative anyway? Even if it's not widely used, it's stupid to have an argument that does nothing for a given language. Of course I don't speak Russian so I don't know what the alternative could even be, but I just wanted to point that out.

alerque:
> Wouldn't it be better to still have an alternative anyway?

What's the point in stuffing "alternatives" into languages that don't use them?

Andrew15-5:
Indeed. If I've at least heard something remotely about alternatives (and that it's even a thing), then of course I would've updated it or left untouched.

Andrew15-5:
@G1gg1L3s, I've mentioned you because of https://github.com/typst/typst/pull/578 and this line:

https://github.com/typst/typst/blob/69ba8884f7f697c3c3824e695d948eb293debec5/library/src/text/quotes.rs#L148

This doesn't match what's being written in https://en.wikipedia.org/wiki/Quotation_mark#History about Ukrainian language.

laurmaedje:
@G1gg1L3s Any particular reason for the negative reaction?

G1gg1L3s:
> @G1gg1L3s Any particular reason for the negative reaction?

Sorry, I was just initially confused why I was in the cc. But now I see that Ukrainian defaults indeed don't match the Ukrainian orthography of 2019. Additionally, it lacks default support for smart quotes.

I can open a PR for the fixes if it's okay.

laurmaedje:
> I can open a PR for the fixes if it's okay.

Sure!

laurmaedje:
Thanks!

PR #6332, opened 2025-05-25
Disallow empty labels and references (#5776)

### LINKED ISSUE #5776: Empty labels should not be allowed
### Description

Attaching an empty label to a figure (or other element) as in
```typ
#figure(rect(), caption: [...]) #label("")
```
allows you to reference it with just `@`, which should probably not be allowed.

### Reproduction URL

_No response_

### Operating system

Web app, Windows

### Typst version

Daaiid:
Closes #5776.

This PR makes it no longer possible to create "empty" labels, meaning that their name is an empty string.
Previously using
```typst
#label("")
```
created a label that could be referenced with just `@`.

This is now consistent with `<>` not creating a label with no name but instead just emitting the same characters verbatim.

I went over all the calling code of `Label::new()` to see if any other places could introduce empty labels.
I don't know if bibliographies can even contain empty keys, but I added a guard there to be sure.
All other call sites are expected to produce a valid label, with one exception:
An `@` without any following identifier.

If we deem an empty name label as invalid, we should follow that same logic for references. Producing a label with no reference is valid but being able to produce a reference where the label is not representable is invalid.

The result: The last commit changes the syntax/lexer to force an `@` to be followed by identifier characters or else it won't be resolved into a reference. Because we can't produce empty labels anymore we also can't produce empty references with `ref()`.

As a result of all changes it should no longer be possible to produce empty labels and references.

If I missed any edge cases or if this PR is too big of a breaking change, please let me know.
I had the impression from the discussion on the issue that this would be the most consistent solution.

### DIFF
--- a/crates/typst-library/src/foundations/label.rs
+++ b/crates/typst-library/src/foundations/label.rs
@@ -54,7 +54,9 @@ pub struct Label(PicoStr);
 
 impl Label {
     /// Creates a label from an interned string.
+    /// Callers need to ensure the given string is not empty.
     pub fn new(name: PicoStr) -> Self {
+        debug_assert!(name != PicoStr::EMPTY);
         Self(name)
     }
 
--- a/crates/typst-library/src/foundations/label.rs
+++ b/crates/typst-library/src/foundations/label.rs
@@ -1,7 +1,10 @@
 use ecow::{eco_format, EcoString};
 use typst_utils::{PicoStr, ResolvedPicoStr};
 
-use crate::foundations::{func, scope, ty, Repr, Str};
+use crate::{
+    diag::StrResult,
+    foundations::{bail, func, scope, ty, Repr, Str},
+};
 
 /// A label for an element.
 ///
@@ -27,7 +30,8 @@ use crate::foundations::{func, scope, ty, Repr, Str};
 /// # Syntax
 /// This function also has dedicated syntax: You can create a label by enclosing
 /// its name in angle brackets. This works both in markup and code. A label's
-/// name can contain letters, numbers, `_`, `-`, `:`, and `.`.
+/// name can contain letters, numbers, `_`, `-`, `:`, and `.`. Empty label names
+/// get rejected.
 ///
 /// Note that there is a syntactical difference when using the dedicated syntax
 /// for this function. In the code below, the `[<a>]` terminates the heading and
@@ -67,13 +71,17 @@ impl Label {
 
 #[scope]
 impl Label {
-    /// Creates a label from a string.
+    /// Creates a label from a string. Fails for empty strings.
     #[func(constructor)]
     pub fn construct(
         /// The name of the label.
         name: Str,
-    ) -> Label {
-        Self(PicoStr::intern(name.as_str()))
+    ) -> StrResult<Label> {
+        if name.is_empty() {
+            bail!("expected non-empty label name");
+        }
+
+        Ok(Self(PicoStr::intern(name.as_str())))
     }
 }
 
--- a/crates/typst-library/src/model/bibliography.rs
+++ b/crates/typst-library/src/model/bibliography.rs
@@ -313,12 +313,17 @@ impl Bibliography {
         for (source, data) in sources.0.iter().zip(data) {
             let library = decode_library(source, data)?;
             for entry in library {
-                match map.entry(Label::new(PicoStr::intern(entry.key()))) {
+                let key = entry.key();
+                if key.is_empty() {
+                    bail!("empty bibliography key found");
+                }
+
+                match map.entry(Label::new(PicoStr::intern(key))) {
                     indexmap::map::Entry::Vacant(vacant) => {
                         vacant.insert(entry);
                     }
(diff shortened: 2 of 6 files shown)


tests/suite/model/ref.typ
```diff
@@ -85,3 +85,12 @@ Text seen on #ref(<text>, form: "page", supplement: "Page").
 // Test reference with non-whitespace before it.
 #figure[] <1>
 #test([(#ref(<1>))], [(@1)])
+
+--- ref-to-empty-label-not-possible ---
+// @ without any following label should just produce the symbol in the output
+// and not produce a reference to a label with an empty name.
+@ \<- this should show up
```
laurmaedje: as-is, this doesn't really test it because the test below fails. you can fix this by splitting this up into two tests.

tests/suite/foundations/label.typ
```diff
@@ -92,3 +92,12 @@ _Visible_
...
 #[].label
+
+--- label-not-named-error ---
+// Error: 23-32 expected non-empty label name
+= Something to label #label("")
+
+--- label-not-named-with-variable-error ---
+#let var = ""
+// Error: 18-28 expected non-empty label name
+= Another label #label(var)
```
laurmaedje: I think we can skip this one as the fact that variables work is tested enough elsewhere

crates/typst-library/src/foundations/label.rs
```diff
@@ -67,13 +73,17 @@ impl Label {
...
+    /// Creates a label from a string. Fails for empty strings.
     #[func(constructor)]
     pub fn construct(
         /// The name of the label.
         name: Str,
-    ) -> Label {
-        Self(PicoStr::intern(name.as_str()))
+    ) -> StrResult<Label> {
+        if name.is_empty() {
+            bail!("expected non-empty label name");
```
laurmaedje:
```suggestion
            bail!("label name must not be empty");
```

crates/typst-library/src/model/bibliography.rs
```diff
@@ -313,12 +313,17 @@ impl Bibliography {
         for (source, data) in sources.0.iter().zip(data) {
             let library = decode_library(source, data)?;
             for entry in library {
-                match map.entry(Label::new(PicoStr::intern(entry.key()))) {
+                let key = entry.key();
+                if key.is_empty() {
+                    bail!("empty bibliography key found");
```
laurmaedje:
```suggestion
                    bail!("bibliography key must not be empty");
```

crates/typst-library/src/foundations/label.rs
```diff
@@ -50,7 +54,9 @@ pub struct Label(PicoStr);
 
 impl Label {
     /// Creates a label from an interned string.
+    /// Callers need to ensure the given string is not empty.
     pub fn new(name: PicoStr) -> Self {
+        debug_assert!(name != PicoStr::EMPTY);
```
laurmaedje: Having the assertion here makes it more likely that we have an empty label or runtime panic because some callsite didn't properly check it. We should instead return `Option` here move the check to the callsites, potentially unwrapping in cases where we're sure that the string is non-empty.

crates/typst-library/src/foundations/label.rs
```diff
@@ -1,7 +1,10 @@
 use ecow::{eco_format, EcoString};
 use typst_utils::{PicoStr, ResolvedPicoStr};
 
-use crate::foundations::{func, scope, ty, Repr, Str};
+use crate::{
+    diag::StrResult,
+    foundations::{bail, func, scope, ty, Repr, Str},
+};
```
laurmaedje:
```suggestion
use crate::{
    diag::StrResult,
    foundations::{bail, func, scope, ty, Repr, Str},
};
```
please ungroup these imports

tests/suite/model/ref.typ
Daaiid: Good catch. I split the tests as suggested and added a reference for this one.

tests/suite/foundations/label.typ
Daaiid: :+1: got rid of this test.

crates/typst-library/src/model/bibliography.rs
Daaiid: I incorporated both suggested error messages.

crates/typst-library/src/foundations/label.rs
Daaiid: I changed it to an `Option` but I had trouble figuring out the appropriate error handling in `markup.rs`.
Any suggestions for improvements there would be very welcome.

crates/typst-library/src/foundations/label.rs
Daaiid: :+1:

Daaiid:
@laurmaedje, I addressed your feedback. The `Label::new()` call sites mostly just unwrap, but according to my judgement these are correct. The shorthand syntax can't produce empty labels anymore, same with the scripting.
If you would like some other approach please let me know.

laurmaedje:
Thanks!

Issue #6304, opened 2025-05-19
Sticky block loses stickiness when the other block is labelled


bluss:
### Description

In this example, there are two blocks: A(sticky) + B(label).  If the label is removed on B, then stickiness works as expected.

Reproducing code. There is a headline text "Sticky block" and then a styled raw block following it. If we remove `<label>` in this code, the headline sticks properly to the following block.

```typ
#set page(height: 10cm)
#v(6.5cm)
#block(sticky: true)[Sticky block]

#show raw: it => {
  [#block(it.text)<label>]
}
#raw(block: true, "
Raw text
And more raw text. And more
")
```

----

Picture 1: `<label>` present (code as above). This behaviour is bad.
![Image](https://github.com/user-attachments/assets/e6e3bce4-ee34-45fe-9a48-f4cd1809aa57)

Pipcture 2: Remove `<label>`. This behaviour is as expected.
![Image](https://github.com/user-attachments/assets/6407dc01-7568-439d-b2e9-815c9129548c)

----

The problem does not reproduce for me with just blocks, the show rule was necessary.

Possibly related to issue #6203

### Typst version

- [x] I am using the latest version of Typst

### Other

There is a workaround. Apply the label like this: `[#{[] + block(it.text)}<label>]`. This is not exactly equivalent - not labelling the block anymore, just something that contains the block.

Issue #5977, opened 2025-02-28
Heading sometimes doesn't stick to following block-level quote


gyrrhe:
### Description

The following typst code:

```typ
#set page(width: 200pt, height: 100pt)

#lorem(8)
== Heading
#quote(block: true, lorem(8))

#pagebreak()

#lorem(8)
== Heading
#block(breakable: false, lorem(8))
```

Produces the following three pages:
![Image](https://github.com/user-attachments/assets/3b784cc2-4ded-4e26-bb76-dc6ce95897b2)

As you can see:

- In the first two pages, the heading incorrectly does not stick to the block-level quote that follows it. I would expect both the heading and the quote to appear on page 2. Note that I had to play a bit with the page size to reproduce the bug, but it does happen naturally in the documents I work on.
- For comparison, in the third page, the heading correctly sticks to the non-breakable block that follows it.

I am using typst 0.13.

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

PgBiel:
Also reproducible in 0.12.

Note that the second block isn't exactly what `quote` is shown as, as it has some padding as well, which seems to be the cause of the problem: the following code exhibits the same behavior.

```typ
#set page(width: 200pt, height: 100pt)

#lorem(8)
== Heading
// #quote(block: true, lorem(8))
#{
  set pad(x: 1em)
  set block(above: 2.4em, below: 1.8em)
  pad(block(lorem(8)))
}
```

<details><summary>Current output</summary>
<p>

![heading and quote are separate](https://github.com/user-attachments/assets/bd17fd3c-d00d-4e07-b611-4425582c292e)

</p>
</details> 

In contrast, setting block `inset` rather than wrapping in `pad` fixes it.

```typ
#set page(width: 200pt, height: 100pt)
#show quote: it => it

#lorem(8)
== Heading
// #quote(block: true, lorem(8))
#{
  set pad(x: 1em)
  set block(above: 2.4em, below: 1.8em)
  block(inset: (x: 1em), lorem(8))
}
```
<details><summary>New output</summary>
<p>

![heading and quote together now](https://github.com/user-attachments/assets/aec3c2dd-d2f7-46fc-b5bc-d392ad7727a0)

</p>
</details>

A quick fix would be to use block inset instead of wrapping the block created by `quote` in a `pad`. But this behavior is still interesting, and I wonder what could be prompting this behavior.

Marcono1234:
(Typst 0.13.1)

[...]

PR #6335, opened 2025-05-26
Avoid breaking after an empty frame

### LINKED ISSUE #2914: Empty raw block at the end of a page
### Description

I want to put some code in a block with background color. But I found that sometimes, at the end of a page, there would be a block with color but no content. It looks weird. I'm not sure if this is a bug, but blocks containing normal text don't seem to cause this problem.

A example:

````typst
#show raw.where(block: true): it => [
  #block(fill: luma(230), inset: 10pt, width: 100%)[
    #it
  ]
]

#lorem(630)

```rust
fn main() {
    println!("Hello, world!");
}
```
````
图片

typst version: 0.10.0 (70ca0d25)

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

### PR DESCRIPTION
**Pull request by EpicEricEE**
Fixes #2914 and the first part of #6125.

### What was wrong?

When a breakable block is split into multiple regions, and the first region is too small to fit any part of the block, it would previously add still add an empty frame to that region. However, when the block contains another breakable block, which also doesn't fit into the first region, the originally empty frame isn't empty anymore, as the empty frame from the nested block is added to it. This is what lead to #2914, where the no longer empty frame unexpectedly gets a fill.

Even when there is no nested block, and the empty region remains unfilled, it is still considered as part of the block. Thus, when it is laid out, its height is subtracted, making the remaining (and actually visible) part of the block smaller then expected, as pointed out in https://github.com/typst/typst/issues/2914#issuecomment-2900781138.

A third issue that could have surfaced in the future is that this invisible "leftover" part would be enough for a block to count as sticking to a preceding block, even though the visible part would actually only be in the next region. This isn't an issue yet, as sticky blocks currently lose most of their breaking capabilities as soon as something sticks to them, but #6020 already planned on changing that.

### What changed?

The fix itself is quite simple, and just includes the same empty-frame check, which is also used when skipping the fill and stroke for a first empty frame. This check is put right before the spill is created, so that instead of breaking the child, it can be entirely moved into the next region when (only) the first frame is empty. For cases where all frames are empty (e.g. explicitly sized blocks without content or fill), spilling remains allowed.

Two tests have failed with this change, but the updated references seem more correct to me:
- In `grid-header-containing-rowspan`, the last cell on the first page is no longer stroked as there is nothing inside it. I assume this was the same issue as [#2914](https://github.com/typst/typst/issues/2914), though I might have misinterpreted the test.
- In `grid-header-orphan-prevention`, the "B" in the first header was previously too bold (drawn multiple times?). I'm not sure what happened there, but it seems to be fixed now.

### CODE REVIEW

File: crates/typst-layout/src/flow/collect.rs [hunk 1]
```diff
@@ -466,6 +467,7 @@ impl<'a> MultiChild<'a> {
         let mut spill = None;
         if frames.next().is_some() {
             spill = Some(MultiSpill {
+                exist_non_empty_frame,
```
PgBiel, line 470: Minor comment, but it'd be a good idea to inline the variable here, especially since its calculated value is discarded outside of this branch.

PgBiel (commented):
I'm impressed such a small change produced the desired effect here, very smart! Although I haven't looked too deep into it yet to tell whether this could produce some regression of some sort. Still, nice find!

> A third issue that could have surfaced in the future is that this invisible "leftover" part would be enough for a block to count as sticking to a preceding block, even though the visible part would actually only be in the next region.

I actually faced this just yesterday! A block had a very tiny empty leftover at the previous page (everything else was at the next page), causing a heading right before it to appear to be an orphan (I even double-checked my show rules :joy:), but adding fill to the block revealed that the sneaky thing was still in the previous page by a few pixels. Would be certainly nice to have this not happen anymore.

> The fix itself is quite simple, and just includes the same empty-frame check, which is also used when skipping the fill and stroke for a first empty frame. This check is put right before the spill is created, so that instead of breaking the child, it can be entirely moved into the next region when (only) the first frame is empty.

So, if I understood correctly, you took code specific to block elements with fill and stroke and copied it to the flow layouter to avoid creating an empty frame entirely, not only skipping its fill and stroke? And by doing this, the fix bubbles up into a parent block containing a spilled breakable block, as its first frame also becomes entirely empty as well? That would seem decently robust in principle.

> In grid-header-containing-rowspan, the last cell on the first page is no longer stroked as there is nothing inside it. I assume this was the same issue (...) though I might have misinterpreted the test.

The test was entirely unrelated, so I think I just didn't care about it back then. But yes, it's the same issue: an auto row is skipped to the next page if the first frame is empty, but here it wasn't empty, as the block containing `lorem` had, itself, created a nested empty frame, if I'm understanding correctly. So the new output seems correct to me!

> In grid-header-orphan-prevention, the "B" in the first header was previously too bold (drawn multiple times?). I'm not sure what happened there, but it seems to be fixed now.

Wow! I think I noticed this at the time, but had no idea either and just ignored it. That is actually funny, I wonder what triggered this?
Seems like the header was placed normally once and then immediately repeated, but somehow it escaped to the next region or something :thinking:

Maybe we could test these changes on top of #6168 to see if anything else changed?

EpicEricEE:
I realized we can also easily fix #6304 by not only skipping the fill/stroke, but also the label for empty orphan frames. Adding a label wraps the frame in another frame, making the wrapper frame not empty.

PgBiel:
> Adding a label wraps the frame in another frame, making the wrapper frame not empty.

Could this affect introspection, e.g. by changing the page returned by `locate(<lbl>)`? Maybe we should test that too.

EpicEricEE:
The position would previously be on the page of the empty orphan frame, but now that it's no longer labeled, it's on the next page as expected. I've added a test for that.

PgBiel:
Thank you. In principle, this seems sound to me, but I'll wait for @laurmaedje's opinion.

EpicEricEE:
While looking a bit into #6388, I found that this introduces an infinite loop of moving the child into the next region for this case:
```typ
#set page(width: 70pt, height: 70pt, margin: 10pt)
#grid(block(sticky: true, fill: blue, width: 100%, height: 75pt))
```
Don't know if I should add a `may_progress` in the spill check, or if there's something wrong in the grid layouter.

tests/suite/introspection/locate.typ
```diff
@@ -72,6 +72,18 @@ B
 #pagebreak(weak: true)
 #metadata(none) <e>
 
+--- locate-migrated-breakable ---
+// Ensure that when a breakable element fully migrates to the next page without
+// orphan frames, its position correctly reflects that.
+#set page(height: 40pt)
+A
+#block[B]<a>
```
laurmaedje: A `<b>` would be less confusing here.

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -283,6 +283,13 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
 
         // Lay out the block.
         let (frame, spill) = multi.layout(self.composer.engine, self.regions)?;
+        if frame.is_empty() && spill.as_ref().is_some_and(|s| s.exist_non_empty_frame) {
```
laurmaedje: I think a `self.regions.may_progress()` check would probably fix it? I haven't thought too much about it, but in general having an error return from distribution with no may_progress check and without the composer being advanced before seems ill-advised.

So, looking at other places here:
- The first stop in `fn multi` has a `may_progress` check because `is_full()` calls `may_progress` internally. 
- The latter stop in `fn multi` has a `self.composer.work.advance()` call, so progress has already been made
- Stops in `fn line` either have `may_progress` checks or similar constructs
- The stop in `fn single` has a check
- The other ones (in `multi_spill`, `flush`, and `break`) I'm not too sure about right now... these have complex interactions with other parts.

Anyway, I think adding a check here makes sense, is consistent with the rest, and seems to fix it on my end without breaking any of the tests.

laurmaedje (commented):
Sorry that it took so long for me to respond. I guess I evaded this PR 🙈 as the concept matter is so tricky and reviewing work on such things is even trickier than working on them.

However, upon looking into this a bit more closely, it seems like a nice approach indeed! It's pretty natural in the new-ish relayout paradigm.

> While looking a bit into #6388, I found that this introduces an infinite loop of moving the child into the next region for this case:

Oof, infinite layout loops are always a pain. But I think a `may_progress` check in `distribute.rs` should fix it; see my comment on the relevant line.

> > Adding a label wraps the frame in another frame, making the wrapper frame not empty.
>
> Could this affect introspection, e.g. by changing the page returned by locate(<lbl>)? Maybe we should test that too.

The frame labelling should have no effect on introspection. It's not used for that at all, it's just used to expose the label in SVG export, which was a feature request by someone at some point, for something animation-related.

> The position would previously be on the page of the empty orphan frame, but now that it's no longer labeled, it's on the next page as expected. I've added a test for that.

Regardless of not being caused by the frame labelling change, this change is entirely natural. The observed position of the block depends on where in the frame the introspection tags end up. These are are sticky and will migrate to the next region with the block.

laurmaedje:
Also fixes https://github.com/typst/typst/issues/5977 :)

laurmaedje:
Thank you! This is a very nice fix.

PR #6320, opened 2025-05-22
Apply script-style to numbers consistently in math

mkorje:
Closes #6318.

Not quite sure if we want to actually do this. The changes to tests are shown below so you can see what it would look like with the changes.

<details>

<summary>Tests</summary>

**issue-4828-math-number-multi-char**:
| Old | New |
| - | - |
| issue-4828-math-number-multi-char old | issue-4828-math-number-multi-char new |

**issue-5489-matrix-stray-linebreak**:
| Old | New |
| - | - |
| issue-5489-matrix-stray-linebreak old | issue-5489-matrix-stray-linebreak new |

**math-attach-kerning-mixed**:
| Old | New |
| - | - |
| math-attach-kerning-mixed old | math-attach-kerning-mixed new |

**math-attach-limit-long**:
| Old | New |
| - | - |
| math-attach-limit-long old | math-attach-limit-long new |

**math-attach-prescripts**:
| Old | New |
| - | - |
| math-attach-prescripts old | math-attach-prescripts new |

**math-frac-precedence**:
| Old | New |
| - | - |
| math-frac-precedence old | math-frac-precedence new |

**math-root-frame-size-index**:
| Old | New |
| - | - |
| math-root-frame-size-index old | math-root-frame-size-index new |

**math-root-large-index**:
| Old | New |
| - | - |
| math-root-large-index old | math-root-large-index new |

</details>

Issue #5071, opened 2024-09-30
Corner brackets in math aren't positioned properly in some situations


mkorje:
### Description

I thought #4200 resolved this, but it only does so in "simple" situations like `$⌜a⌟⌞b⌝$`.

In `$⌜a/b⌟$` or `$vec(delim: \⌜, 1, 2)$`, the corner brackets are at the same height as in `$⌜a⌟⌞b⌝$`. They should probably be moved vertically as delims in matrix-like things and in larger chunks of math like fractions.

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

PR #6336, opened 2025-05-27
Use the shaper in math

mkorje:
## Changes

The main change this PR makes is to use the shaper when laying out glyphs in math, replacing the handwritten code we had for applying `ssty`, `dtls`, etc. This also sets us up to support grapheme clusters (e.g. variation sequences) in math as well as RTL math layout in the future. To accomplish this switching a glyph to its `ssty`, `dtls`, etc. variant has been moved to the `StyleChain`.

The `GlyphFragment` struct in math is no longer limited to a single Unicode character. The `try_new` method can now take any string (intended to just be a single grapheme cluster), but will return `None` if the result from shaping isn't a single glyph.

Finally, I also culled the `VariantFragment` struct from math entirely. It really only exists because glyph assemblies need to be stored somehow. But we can actually store them in the `TextItem` that `GlyphFragment`s end up getting turned into! For horizontal assemblies there is no issue, but for vertical ones I needed to add support in the exporters for the `y_offset` and `y_advance` (it is kind of like support for very rudimentary vertical text layout has been added).
- For `typst-pdf` things were simple enough, I just modified the `impl krilla::text::Glyph for PdfGlyph` in `typst-pdf/src/text.rs` and `krilla` produces a correct looking result.
- For `typst-svg`, wherever I saw an `x` or `offset` in `typst-svg/src/text.rs` I just duplicated them to add their `y` counterparts, and correct output is produced.
- For `typst-render` I originally did the same as I did for `typst-svg`, but things were going in the wrong direction vertically. So I negated all the `y` counterparts in `typst-render/src/text.rs` and things were working. I have no idea why the y-axis is seemingly flipped in `typst-render`.

For all of these I have probably made some mistakes, so I would appreciate if others could check my changes here thoroughly!

The change in `typst-pdf` has caused the (hopefully) **only** user facing change in this PR: Text selection of glyph assemblies in math will now only have one copy of the glyph instead of a bunch of them. See below for two example PDFs and what the text selection produces with `zathura`.
<details>

<summary>Glyph assembly text selection</summary>

```typ
--- playground ---
#set page(width: auto)
$ f : A stretch(->, size: #5em) B quad A stretch(\[, size: #5em) B $
```

**Old:** [old.pdf](https://github.com/user-attachments/files/20451171/old.pdf)
```
𝑓: 𝐴→→→→→→→→→→→→→→→𝐵
𝐴
[
[[[
[
𝐵
```

**New:** [new.pdf](https://github.com/user-attachments/files/20451173/new.pdf)
```
𝑓: 𝐴→
𝐵
𝐴
[
𝐵
```

</details>

## Tests

- ~~The eight tests that have been updated are identical to the ones in #6320. This change occurred as the `ssty` OpenType feature tag is now set automatically when the `MathSize` is `Script` or `Sscript`. (I think we will probably go through with #6320 anyways, so this should be fine.)~~
  #6320 has been merged

- Having the `dtls` OpenType feature tag now live on the `StyleChain` has lead to https://github.com/MicrosoftDocs/typography-issues/issues/1191 being introduced. To track this a new test has been added:
  <details>

  <summary>math-accent-dotless-greedy</summary>

  math-accent-dotless-greedy

  ```typ
  --- math-accent-dotless-greedy ---
  // Currently the dotless style propogates to everything in the accent's base,
  // even though it shouldn't.
  $ arrow(P_(c, i dot j) P_(1, i) j) \
    arrow(P_(c, i dot j) P_(1, i) j, dotless: #false) $
  ```

  </details>

- ~~There is only a single pesky test failing, math-accent-sym-call. ~~I'm investigating what's going on here...~~ I've tracked down why its failing: the glyph `ä`'s width has changed from `5.0pt` to `5.11pt` somehow. Now to figure out the cause...~~
  Fixed! No tests are failing now. (I made a mistake when calculating the accent_attach.)

## Performance

I imagine that now running everything in math through the shaper will have some performance overhead. ~~I'll try look into this soon.~~
I ran some benchmarks between `main`, this PR, and this PR but with `#[comemo::memoize]` on `GlyphFragment::try_new`, on [masterproef](https://github.com/Dherse/masterproef) and [notes](https://github.com/jthulhu/notes) with `hyperfine` (with `--warmup 3`). Here are the results:
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./typst-main c --root masterproef masterproef/masterproef/main.typ` | 3.823 ± 0.043 | 3.781 | 3.901 | 1.00 |
| `./typst-pr c --root masterproef masterproef/masterproef/main.typ` | 3.839 ± 0.028 | 3.802 | 3.886 | 1.00 ± 0.01 |
| `./typst-pr-opt c --root masterproef masterproef/masterproef/main.typ` | 3.875 ± 0.038 | 3.802 | 3.916 | 1.01 ± 0.02 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./typst-main c notes/grothendieck-fibrations.typ` | 14.306 ± 0.063 | 14.218 | 14.376 | 1.01 ± 0.01 |
| `./typst-pr c notes/grothendieck-fibrations.typ` | 14.404 ± 0.076 | 14.327 | 14.544 | 1.01 ± 0.01 |
| `./typst-pr-opt c notes/grothendieck-fibrations.typ` | 14.196 ± 0.044 | 14.086 | 14.240 | 1.00 |


crates/typst-layout/src/math/accent.rs
```diff
@@ -15,37 +18,34 @@ pub fn layout_accent(
...
+    let Accent(c) = elem.accent;
+    let mut glyph = GlyphFragment::new(ctx.font, accent_styles, c, elem.span());
 
     // Forcing the accent to be at least as large as the base makes it too
     // wide in many case.
-    let short_fall = ACCENT_SHORT_FALL.at(glyph.font_size);
-    let variant = glyph.stretch_horizontal(ctx, width, short_fall);
-    let accent = variant.frame;
-    let accent_attach = variant.accent_attach;
+    let short_fall = ACCENT_SHORT_FALL.at(glyph.text.size);
```
mkorje, line 45: This should probably be relative to the base's text size and not the accent glyph's. But this will be fixed by #6214 anyways.

crates/typst-layout/src/math/stretch.rs
```diff
@@ -23,285 +15,69 @@ pub fn layout_stretch(
...
-    for (fragment, advance) in selected {
-        let pos = match axis {
-            Axis::X => Point::new(offset, frame.baseline() - fragment.ascent),
-            Axis::Y => Point::with_y(full - offset - fragment.height()),
-        };
-        frame.push_frame(pos, fragment.into_frame());
-        offset += advance;
+    if stretch_axis == Axis::Y {
+        // TODO: this should use delimiter_alignment
+        glyph.center_on_axis();
```
mkorje, line 81: We can probably get away with not changing this back as the only thing that `delimiter_alignment` doesn't return `VAlignment::Horizon` for are the corner brackets. But the corner brackets are already aligned in the font and will return before going through any of the stretching code (cf. https://github.com/typst/typst/pull/5069#issue-2555819226), so we could probably just remove `delimiter_alignment` entirely for now, leaving it to be dealt with in #5071.

crates/typst-layout/src/math/stretch.rs
```diff
@@ -23,285 +15,69 @@ pub fn layout_stretch(
...
+    // TODO: Glyph may have already been stretched, so need to go back to unstretched state somehow.
+    if glyph.text.glyphs.len() > 1 && glyph.text.glyphs[0].x_advance == Em::zero() {
+        glyph.shift -= glyph.descent();
     }
+    glyph.text.glyphs = vec![Glyph {
+        id: glyph.base_id.0,
+        x_advance: glyph.text.font.advance(glyph.base_id.0).unwrap_or_default(),
+        ..glyph.text.glyphs[0].clone()
+    }];
+    glyph.update_glyph();
```
mkorje, line 73: This is some hideous looking code, will move it into its own function in `impl GlyphFragment` probably.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -190,19 +186,25 @@ impl MathFragment {
...
-                    .unwrap_or_default()
+                // TODO: figure out what should happen here with glyph assemblies. Maybe use first and last based on the corner??
+                kern_at_height(
+                    &glyph.text.font,
+                    GlyphId(glyph.text.glyphs[0].id),
+                    corner,
+                    Em::from_length(height, glyph.text.size),
+                )
+                .unwrap_or_default()
+                .at(glyph.text.size)
```
mkorje, line 207: TODO

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,264 +225,258 @@ impl From<FrameFragment> for MathFragment {
...
-    }
+        let mut buffer = UnicodeBuffer::new();
+        buffer.push_str(str);
+        buffer.set_language(language(styles));
+        // TODO: Bug in rustybuzz? rustybuzz::script::SCRIPT_MATH does not work.
+        // i.e. ssty is not applied
+        buffer.set_script(
+            rustybuzz::Script::from_iso15924_tag(ttf_parser::Tag::from_bytes(b"math"))
+                .unwrap(),
+        );
```
mkorje, line 265: TODO

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,264 +225,258 @@ impl From<FrameFragment> for MathFragment {
...
+    pub fn stretch(
+        &mut self,
+        ctx: &mut MathContext,
+        target: Abs,
+        short_fall: Abs,
+        axis: Axis,
+    ) {
+        // If the base glyph is good enough, use it.
+        // TODO: there's inconsistency here with italic correction.
+        let advance = self.size.get(axis);
```
mkorje, line 421: TODO

LaurenzV:
> @LaurenzV would you happen to know if this is expected behaviour in rustybuzz?

Hmm, I'm not sure, but at least the `SCRIPT_MATH` constant isn't used anywhere. @behdad Do you know if it does anything in harfbuzz?

mkorje:
I opened https://github.com/harfbuzz/rustybuzz/issues/162.

crates/typst-layout/src/math/text.rs
```diff
@@ -144,36 +147,17 @@ pub fn layout_symbol(
...
-fn layout_glyph(
-    mut glyph: GlyphFragment,
-    dtls: bool,
-    ctx: &mut MathContext,
-    styles: StyleChain,
-) -> MathFragment {
-    if dtls {
-        glyph.make_dotless_form(ctx);
-    }
+fn layout_glyph(glyph: &mut GlyphFragment, ctx: &mut MathContext, styles: StyleChain) {
```
wrzian: This function should probably be renamed (and comment updated) if we're changing the signature this much. Maybe `adjust_glyph_layout()`?

crates/typst-layout/src/math/stretch.rs
```diff
@@ -23,285 +14,58 @@ pub fn layout_stretch(
...
-    let glyph = match fragment {
-        MathFragment::Glyph(glyph) => glyph.clone(),
-        MathFragment::Variant(variant) => {
-            GlyphFragment::new(ctx, styles, variant.c, variant.span)
-        }
-        _ => return,
-    };
+    let size = fragment.size();
+
+    let MathFragment::Glyph(ref mut glyph) = fragment else { return };
```
wrzian, line 33: This is clean!

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,283 @@ impl From<FrameFragment> for MathFragment {
...
 impl GlyphFragment {
-    pub fn new(ctx: &MathContext, styles: StyleChain, c: char, span: Span) -> Self {
-        let id = ctx.ttf.glyph_index(c).unwrap_or_default();
-        let id = Self::adjust_glyph_index(ctx, id);
-        Self::with_id(ctx, styles, c, id, span)
+    pub fn new(font: &Font, styles: StyleChain, c: char, span: Span) -> Self {
+        Self::try_new(font, styles, c.encode_utf8(&mut [0; 4]), span).unwrap()
     }
 
     pub fn try_new(
```
wrzian: Could you add a doc comment for this method?

crates/typst-layout/src/math/fragment.rs
```diff
@@ -607,11 +634,186 @@ fn kern_at_height(
...
 
-    Some(kern.kern(i)?.scaled(ctx, font_size))
+    Some(font.to_em(kern.kern(i)?.value))
+}
+
+fn axis_height(font: &Font) -> Option<Em> {
+    Some(font.to_em(font.ttf().tables().math?.constants?.axis_height().value))
+}
+
+pub fn stretch_axes(font: &Font, id: GlyphId) -> Axes<bool> {
```
wrzian: I feel like this should still be part of `stretch.rs` (it's still only used there). Any particular reason it got moved?

wrzian (commented):
I don't know enough about shaping to comment on that aspect, but after a review this looks pretty clean to me. I had a few style and organizational notes/questions, but the changes look like good simplifications, nice work!

laurmaedje:
I agree with @wrzian, the PR looks quite excellent! I don't think there's too much to say either way, but I think it would still be good to rebase before I leave my review.

crates/typst-layout/src/math/text.rs
```diff
@@ -126,15 +119,25 @@ pub fn layout_symbol(
...
+    ) {
+        Some(mut glyph) => {
+            layout_glyph(&mut glyph, ctx, styles);
+            glyph.into()
+        }
         None => {
             // Not in the math font, fallback to normal inline text layout.
+            // TODO: Should replace this with proper fallback in [`GlyphFragment::try_new`].
             layout_inline_text(c.encode_utf8(&mut [0; 4]), elem.span(), ctx, styles)?
                 .into()
```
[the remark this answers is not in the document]
mkorje: FYI, this is being done in #6365.

crates/typst-layout/src/math/fragment.rs
mkorje: Since it accesses font data I thought it was nicer to have in `fragment.rs`, which is where all the functions in math that do are.

crates/typst-layout/src/math/text.rs
mkorje: I went with `adjust_glyph_layout` as you suggested.

mkorje:
In the latest commit I removed `delimiter_alignment` (see also https://github.com/typst/typst/pull/6336#discussion_r2108324724).
All that changes is the placement of corner brackets in mat/vec/cases (from incorrect to still incorrect).

<details>

<summary>Change</summary>

```typ
--- playground ---
$ mat(delim: \⌜, 1, 1, 1; 1, 1, 1; 1, 1, 1) $
```

| Old | New |
| - | - |
| old | new |

</details>

crates/typst-layout/src/math/shared.rs
```diff
@@ -59,6 +61,14 @@ pub fn style_cramped() -> LazyHash<Style> {
     EquationElem::set_cramped(true).wrap()
 }
 
+pub fn style_flac() -> LazyHash<Style> {
```
laurmaedje, line 65: could you add very brief docs here?

crates/typst-layout/src/math/text.rs
```diff
@@ -138,37 +146,23 @@ pub fn layout_symbol(
...
+fn adjust_glyph_layout(
+    glyph: &mut GlyphFragment,
     ctx: &mut MathContext,
     styles: StyleChain,
-) -> MathFragment {
-    if dtls {
-        glyph.make_dotless_form(ctx);
-    }
+) {
     let math_size = EquationElem::size_in(styles);
```
laurmaedje: this could be inlined into the if to not determine it unnecessarily

crates/typst-library/src/text/item.rs
```diff
@@ -54,6 +59,10 @@ pub struct Glyph {
     pub x_advance: Em,
     /// The horizontal offset of the glyph.
     pub x_offset: Em,
+    /// The advance height of the glyph.
+    pub y_advance: Em,
+    /// The vertical offset of the glyph.
+    pub y_offset: Em,
```
laurmaedje: It would be good to mention that these are Y-up. (They are, right?)

crates/typst-layout/src/math/fragment.rs
```diff
@@ -1,28 +1,31 @@
...
 use typst_syntax::Span;
-use typst_utils::default_math_class;
+use typst_utils::{default_math_class, Get};
 use unicode_math_class::MathClass;
 
-use super::{stretch_glyph, MathContext, Scaled};
+use crate::inline::create_shape_plan;
 use crate::modifiers::{FrameModifiers, FrameModify};
 
+use super::MathContext;
```
laurmaedje: super and crate imports can be in the same group

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
 
 impl GlyphFragment {
-    pub fn new(ctx: &MathContext, styles: StyleChain, c: char, span: Span) -> Self {
-        let id = ctx.ttf.glyph_index(c).unwrap_or_default();
-        let id = Self::adjust_glyph_index(ctx, id);
-        Self::with_id(ctx, styles, c, id, span)
+    /// Calls `try_new` with the given character, assuming the font has a glyph
+    /// for it.
+    pub fn new(font: &Font, styles: StyleChain, c: char, span: Span) -> Self {
+        Self::try_new(font, styles, c.encode_utf8(&mut [0; 4]), span).unwrap()
```
laurmaedje: we are using this quite a lot, both with user-input chars and constant chars. unwrapping here seems problematic to me.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
     }
 
+    /// Try to create a new glyph out of the given string. Will return None if
+    /// the result from shaping the string is not a single glyph or is a tofu.
     pub fn try_new(
-        ctx: &MathContext,
+        font: &Font,
         styles: StyleChain,
-        c: char,
+        str: &str,
```
laurmaedje: Could you rename this `str` to `s`, `string`, or `text`? I try to avoid `str` as it's exactly the same as the type name.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
-            font_size: TextElem::size_in(styles),
+            lang: TextElem::lang_in(styles),
+            region: TextElem::region_in(styles),
+            text: str.into(),
+            glyphs: vec![Glyph {
+                id: info.glyph_id as u16,
+                x_advance: font.to_em(pos.x_advance),
+                x_offset: Em::zero(),
+                y_advance: Em::zero(),
+                y_offset: Em::zero(),
```
laurmaedje: why are we not taking these from the shaped output?

crates/typst-layout/src/math/fragment.rs
```diff
@@ -614,11 +633,186 @@ fn kern_at_height(
...
+fn glyph_construction(font: &Font, id: GlyphId, axis: Axis) -> Option<GlyphConstruction> {
+    font.ttf()
+        .tables()
+        .math?
+        .variants
+        .map(|variants| match axis {
+            Axis::X => variants.horizontal_constructions,
+            Axis::Y => variants.vertical_constructions,
+        })
+        .and_then(|constructions| constructions.get(id))
```
laurmaedje:
```suggestion
        })?
        .get(id)
```

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
-        Self::with_id(ctx, styles, c, id, span)
+    /// Calls `try_new` with the given character, assuming the font has a glyph
+    /// for it.
+    pub fn new(font: &Font, styles: StyleChain, c: char, span: Span) -> Self {
+        Self::try_new(font, styles, c.encode_utf8(&mut [0; 4]), span).unwrap()
     }
 
+    /// Try to create a new glyph out of the given string. Will return None if
+    /// the result from shaping the string is not a single glyph or is a tofu.
     pub fn try_new(
```
laurmaedje: Let's memoize this

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
-    ) -> Self {
+        let info = buffer.glyph_infos()[0];
+        let pos = buffer.glyph_positions()[0];
+
+        // TODO: add support for coverage and fallback, like in normal text shaping.
+        if info.glyph_id == 0 {
+            return None;
+        }
+
+        let cluster = info.cluster as usize;
```
laurmaedje, line 315: I wonder whether this will ever be non-zero.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
+            lang: TextElem::lang_in(styles),
+            region: TextElem::region_in(styles),
+            text: str.into(),
+            glyphs: vec![Glyph {
+                id: info.glyph_id as u16,
+                x_advance: font.to_em(pos.x_advance),
+                x_offset: Em::zero(),
+                y_advance: Em::zero(),
+                y_offset: Em::zero(),
+                range: 0..str.len() as u16,
```
laurmaedje:
```suggestion
                range: 0..str.len().saturating_as(),
```
even though exceedingly impossible that this overflows, this is a little safer

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
-    pub lang: Lang,
-    pub region: Option<Region>,
-    pub fill: Paint,
-    pub stroke: Option<FixedStroke>,
-    pub shift: Abs,
-    pub width: Abs,
-    pub ascent: Abs,
-    pub descent: Abs,
+    // Text stuff.
+    pub text: TextItem,
```
laurmaedje: maybe `item` would be a better name. makes it clear that it's not a string and you can't get weird `.text.text` to access the string.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -229,266 +238,275 @@ impl From<FrameFragment> for MathFragment {
...
-    pub span: Span,
-    pub modifiers: FrameModifiers,
+    pub class: MathClass,
     pub limits: Limits,
     pub extended_shape: bool,
+    pub mid_stretched: Option<bool>,
+    // External frame stuff.
+    pub modifiers: FrameModifiers,
+    pub shift: Abs,
+    pub align: Abs,
```
laurmaedje, line 258: This can't be unified with with `shift` because it needs to be separately resettable I guess?

mkorje: Yep exactly

crates/typst-layout/src/math/fragment.rs
mkorje: I've gone with `text`

crates/typst-layout/src/math/fragment.rs
mkorje: Yeah I wasn't quite sure either. Seemed safest to do this though.

crates/typst-layout/src/math/fragment.rs
mkorje: Mainly because they weren't being used before in math layout. And they shouldn't (?) have any affect anyways (i.e. the others will always be zero).

crates/typst-layout/src/math/fragment.rs
mkorje: I guess I've made the implicit assumption that any proper math font will have characters like a sqrt, accents and parenthesis.
I think the only place it could actually be an issue is in `accent.rs` (and maybe also in `mat.rs` with some more exotic delimiters).
Here are all the uses:
```rust
text.rs
73:            let glyph = GlyphFragment::new(ctx.font, styles, c, span);

underover.rs
288:    let mut glyph = GlyphFragment::new(ctx.font, styles, c, span);

root.rs
52:    let mut sqrt = GlyphFragment::new(ctx.font, styles, '√', span);

accent.rs
45:    let mut glyph = GlyphFragment::new(ctx.font, accent_styles, accent.0, elem.span());

frac.rs
112:        let mut left = GlyphFragment::new(ctx.font, styles, '(', span);
117:        let mut right = GlyphFragment::new(ctx.font, styles, ')', span);

mat.rs
187:        GlyphFragment::new(ctx.font, styles.chain(&denom_style), '(', Span::detached());
316:        let mut left = GlyphFragment::new(ctx.font, styles, left_c, span);
325:        let mut right = GlyphFragment::new(ctx.font, styles, right_c, span);
```

crates/typst-library/src/text/item.rs
mkorje: They are. I have no idea why for png export (typst-render) I had to flip the directions though.

behdad: `y_advance` is typically negative, since it's Y-up.

mkorje: The vertical glyph assemblies in math though are bottom to top, so `y_advance` ends up being positive.

crates/typst-layout/src/math/fragment.rs
laurmaedje: All the usages are in functions that already return a `SourceResult` and we have a span (in all but one cases, let's add one there as well), so how about we just unify `new` and `try_new` and return a `SourceResult` there?

laurmaedje: A font bug shouldn't make the compiler crash

crates/typst-library/src/text/item.rs
laurmaedje: > I have no idea why for png export (typst-render) I had to flip the directions though.

Wdym? Isn't it just because the renderer works in a Y-down system? If there's more to it, it would be good to figure out why it works, to make sure we're doing things correctly.

mkorje: Sorry I didn't realise the render works in a Y-down system. Makes sense then. Shouldn't be any more to it.

crates/typst-layout/src/math/fragment.rs
mkorje: > All the usages are in functions that already return a `SourceResult` and we have a span (in all but one cases, let's add one there as well), so how about we just unify `new` and `try_new` and return a `SourceResult` there?

Alright I'll do that

crates/typst-library/src/text/item.rs
laurmaedje: Nothing to be sorry for :)

crates/typst-layout/src/math/fragment.rs
laurmaedje: Couldn't rustybuzz in theory emit an x_offset? I think it wouldn't hurt to just also take these from the shaper.

crates/typst-layout/src/math/fragment.rs
mkorje: I've kept `new` but as `new_char` so we don't have to do `c.encode_utf...` in every call. At the moment we never call `new` (previously `try_new`) with anything other than a char, but this will change in the future (so we should keep it separate I think).

mkorje: Let me know if you'd like to just merge them and use `c.encode_utf...` throughout.

crates/typst-layout/src/math/fragment.rs
mkorje: I thought some tests might fail but everything is fine if take them all from the shaped output.

crates/typst-layout/src/math/fragment.rs
laurmaedje: > Let me know if you'd like to just merge them and use c.encode_utf... throughout.

No, makes sense to keep them separate. I didn't realize this additional difference.

crates/typst-layout/src/math/fragment.rs
mkorje: Ok only slight thing that needs doing is setting these to something for the stretch variants.

mkorje: Ok I did the easiest thing which was to store the whole base glyph instead of just its id on the GlyphFragment struct (the field `base_id: GlyphId` is now `base_glyph: Glyph`).
Clippy is now warning about the size of the enum variant...

laurmaedje: hmm. to be able to reset it? if it's pointless, we can also leave it out for now I guess.

mkorje: yeah to reset it. The "completeness" is nice though

crates/typst-layout/src/math/fragment.rs
wrzian: Then let's add a comment like: "This is only used in `stretch.rs`, but is kept here to centralize where we access font data"

laurmaedje:
Thank you for this! I think it's a great step in the right direction for math layout.

PR #6343, opened 2025-05-28
HTML reading

### LINKED ISSUE #5890: html reader
### Description

html is not exact same as xml, so we cannot read a html effectively using `xml`.

### Use Case

I have a template, which reads html file from workspace when exporting the document as html.

```typ
#let load-html-template(template-path, content) = {
  let head-data = xml(template-path).at(0).children.at(1)
  let head = to-html(head-data)

  html.elem(
    "html",
    attrs: (
      "lang": "en",
      "xmlns": "http://www.w3.org/1999/xhtml",
    ),
    {
      head
      html.elem("body", content)
    },
  )
}
```

https://github.com/typst-doc-cn/news/blob/3ad1c7d834e288a4a06c4ed5231a872a644d1441/typ/templates/html-toolkit.typ#L47-L62

But it fails to read my template, because it is not a valid xml file. For example, the `crossorigin` and the `href` are not valid in the following html code:

```html
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
<link
      href="https://fonts.googleapis.com/css2?family=..&family=..&display=swap"
      rel="stylesheet"
    />
```

Instead I have to change them manually:

```xml
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous" />
<link
      href="https://fonts.googleapis.com/css2?family=..&amp;family=..&amp;display=swap"
      rel="stylesheet"
    />
```

Myriad-Dreamin:
Close #5890. [`html5ever`](https://crates.io/crates/html5ever) is good but hard to use, so I use a popular library [`scraper`](https://crates.io/crates/scraper) which thinly wraps `html5ever`.

Todo: bike-shed the name of the function. The name is already taken by `html` module, thus I quickly pick a random name (`html-decode`).

Reference:

- Comment from some more low-level [library](https://crates.io/crates/html5gum):
  > html5gum was created out of a need to parse HTML tag soup efficiently. Previous options were.
  > - use [html5ever's own tokenizer](https://docs.rs/html5ever/0.25.1/html5ever/tokenizer/index.html) to avoid as much tree-building overhead as possible.
  
- https://crates.io/crates/html5ever
- https://crates.io/crates/scraper
- [XML Reading](https://github.com/typst/typst/commit/f6adc45638409aaa0feb1f70883c11ed553efe4f)

### DIFF
--- a/crates/typst-library/src/loading/html.rs
+++ b/crates/typst-library/src/loading/html.rs
@@ -56,7 +56,6 @@ fn convert_html(node_ref: NodeRef<Node>) -> Value {
         // https://en.wikipedia.org/wiki/Processing_Instruction
         Node::Doctype(..) | Node::ProcessingInstruction(..) => return Value::None,
         Node::Comment(comment) => {
-            // Werid but compatible with current `xml`.
             return Value::Dict(dict! {
                 "tag" => "",
                 "attrs" => dict! {},


crates/typst-library/src/loading/html.rs
```diff
@@ -0,0 +1,99 @@
...
+        // todo: the namespace is ignored
+        Node::Element(element) => {
+            (element.name.prefix.as_ref(), &*element.name.local, Some(element.attrs()))
+        }
+        Node::Fragment => (None, "fragment", None),
+        // todo: doc type and processing instruction are ignored
+        // https://en.wikipedia.org/wiki/Processing_Instruction
+        Node::Doctype(..) | Node::ProcessingInstruction(..) => return Value::None,
+        Node::Comment(comment) => {
+            // Werid but compatible with current `xml`.
```
noamzaks, line 59: Hey, just went over this out of curiosity, you have a small type in "weird" here

Myriad-Dreamin: This comment is some voice in my mind when I implemented it, but I may need to remove it from code and add the comment to the PR page.

crates/typst-library/src/loading/html.rs
```diff
@@ -0,0 +1,98 @@
...
+        Node::Document => return Value::Array(convert_html_children(node_ref)),
+        // todo: the namespace is ignored
+        Node::Element(element) => {
+            (element.name.prefix.as_ref(), &*element.name.local, Some(element.attrs()))
+        }
+        Node::Fragment => (None, "fragment", None),
+        // todo: doc type and processing instruction are ignored
+        // https://en.wikipedia.org/wiki/Processing_Instruction
+        Node::Doctype(..) | Node::ProcessingInstruction(..) => return Value::None,
+        Node::Comment(comment) => {
```
Myriad-Dreamin, line 58: Is that a miss that we recognize comment as node with empty tag? I only add this to be compatible with current `xml` reader. In jsx, "node with empty tag" usually refers to `<>...</>` (the shorthand of [HTML fragments](https://react.dev/reference/react/Fragment)).

laurmaedje:
I had somewhat different plans in this direction. Rather than a data loading function that returns arrays and dictionaries, I was imagining a `html.elem.parse` constructor (or `html.parse` function, not sure yet) that takes a string and returns a content tree consisting of `html.elem`s. A bit like html-react-parser.

Myriad-Dreamin:
> I had somewhat different plans in this direction. Rather than a data loading function that returns arrays and dictionaries, I was imagining a `html.elem.parse` constructor (or `html.parse` function, not sure yet) that takes a string and returns a content tree consisting of `html.elem`s. A bit like html-react-parser.

Why not both? Or you may also mean we should not add `html.decode` and it should be exposed by extending `xml` to `xml(xmlns: "html")`. btw, I can convert decoded result into html elements by myself and (use case) I need to replace some `<slot/>` in decoded template before conversion. Considering the case, I may need to either be able to mutate parsed element or have a raw decoder.

laurmaedje:
I think a parser that takes a string and turns it into HTML element content is generally quite useful and should definitely exist. Using this and converting yourself would be fairly cumbersome. 

The use case here seems a bit niche to me. Slots could potentially also be replaced by use of show rules on HTML elements, though a `where` selector can't select on ID unfortunately because it is nested in attr.

@reknih Thoughts on @Myriad-Dreamin's use case?

reknih:
Let me check if I get what you are planning to use this for:

- You have an external template in which you want to insert Typst code. You want to serve the artifact emitted by immediately, without postprocessing it in other tools.
- Because you need to insert Typst code at multiple spots or otherwise aim to mutate the given HTML, you want access to the DOM of the HTML

In general, I am weary to add HTML as a data loading function for these two reasons:

1. HTML is not a general-purpose data exchange format, and adding it would encourage such usage. I am aware of the microdata part of the HTML spec, but I don't think field usage justifies adding HTML here.
2. This PR exposes the parsed HTML tree, not the DOM (there are subtle differences, for example re: omissable tags). This is bound to lead to confusion as people expect to be able to read the DOM like they do in JavaScript.

Instead, I concur with @laurmaedje that an API that converts well-formed HTML to content is more desirable. However, I also see how it is a valid use case to insert Typst-generated within parsed HTML. For that, I'd like to propose using HTML's slot element:

```typst
#let source = "<div id=\"a\"><slot></slot></div><div id=\"b\"><slot name=\"b\"></slot></div>"
#let slotted = html.parse.with(div)

// emits the source as-is
#slotted()

// emits <div id="a">hello</div><div id="b"></div>
#slotted[hello]

// emits <div id="a">hello</div><div id="b">world</div>
#slotted(slots: (b: [world]))[hello]
``` 

This uses the [`<slot>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot) element already available in standard HTML for Web Components. There is prior art for using the slot element instead of forwarding it, for example in Astro.

What this design would not allow is reading the contents of the input HTML in Typst as well as only using parts of it. I think that's legitimate because it encourages good architecture.

To parse and manipulate HTML, the XML syntax / XHTML is better suited anyways. It has less surprising special cases, is closer to the DOM, and better suited as a data-exchange format, e.g. by embedding RDF or custom namespaced elements.

Myriad-Dreamin:
> Let me check if I get what you are planning to use this for:
> 
> * You have an external template in which you want to insert Typst code. You want to serve the artifact emitted by immediately, without postprocessing it in other tools.

My primary case is only making html elements using HTML "dsl", which are not necessary stored outside of typst files, like:

````typ
#show raw: it => process(html.decode(it.text))
```html
<html/>
```
````

Having a html decoder will be pretty convenient for web developers.

> * Because you need to insert Typst code at multiple spots or otherwise aim to mutate the given HTML, you want access to the DOM of the HTML
> In general, I am weary to add HTML as a data loading function for these two reasons:
> 
> 1. HTML is not a general-purpose data exchange format, and adding it would encourage such usage. I am aware of the microdata part of the HTML spec, but I don't think field usage justifies adding HTML here.

To be clear, that's not only about slots, I wish I could have more low-level processing about HTML, for example:
- [`xpath`](https://developer.mozilla.org/en-US/docs/Web/XML/XPath) support:
  ````typ
  #show raw: it => (
    xpath-load(html.decode(it.text)).
      map(dollar => dollar('a').attr('href', it => urlBase + it)).
      to-html()
  )
  ```html
  <a href="/page1.html" />
  <a href="/page2.html" />
  ```
  ````
- template string inteporlations:
  ````typ
  // render it like typst compile --input now=`date +%Y%m%d%H%M%S`
  #show raw: it => render-template(it.text)
  ```angular-html
  <span>The current time is: {{ sys.inputs.now | string-format('rfc3339') }}</span>
  ```
  ````

> HTML is not a general-purpose data exchange format

It is not a "data exchange format" but a commonly used template format. I think things like `xpath` are not suitable placed in standard library of typst considering bundle size, but only having builtin `html.decode` might be a good balance point. Besides, if typst doesn't officially support `html.decode`, I can imagine that somebody can bundle a `html5ever-rs.wasm` and publish it to typst universe, because it is very popular to use html as templates.

> 2. This PR exposes the parsed HTML tree, not the DOM (there are subtle differences, for example re: omissable tags). This is bound to lead to confusion as people expect to be able to read the DOM like they do in JavaScript.

This should be documented well so that people won't get confused, because we currently also know that `xml.decode` doesn't give a XML DOM.

> Instead, I concur with @laurmaedje that an API that converts well-formed HTML to content is more desirable. However, I also see how it is a valid use case to insert Typst-generated within parsed HTML. For that, I'd like to propose using HTML's slot element:
> 
> ```typst
> #let source = "<div id=\"a\"><slot></slot></div><div id=\"b\"><slot name=\"b\"></slot></div>"
> #let slotted = html.parse.with(div)
> 
> // emits the source as-is
> #slotted()
> 
> // emits <div id="a">hello</div><div id="b"></div>
> #slotted[hello]
> 
> // emits <div id="a">hello</div><div id="b">world</div>
> #slotted(slots: (b: [world]))[hello]
> ```
> 
> This uses the [`<slot>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot) element already available in standard HTML for Web Components. There is prior art for using the slot element instead of forwarding it, for example in Astro.

I think the API of `html.parse` with slots will be neat, because slot is in [standard.](https://html.spec.whatwg.org/#the-slot-element) (I think whatwg links are preferred over MDN docs when we have further more discussion.) But I'm not sure whether I was requesting to extend `html.parse` with handling slot elements specially when we haven't had `html.parse`. There has been years about development of html as templates. People invented template rendering again and again. Among template syntaxes, astro who born after popularity of `slot` elements chose `slot`. The common preference might be changed in future and personal preferences could be diverse. If the preference is changed in future, we might pay a technique debt when we tightly couple `html.parse` with slot functionalities and we cannot break change APIs.

> What this design would not allow is reading the contents of the input HTML in Typst as well as only using parts of it. I think that's legitimate because it encourages good architecture.

I understand your concern about encourage good practices in typst, but it should also keep neutral position confronting sea of personal-flavored HTML template uses like slots. And the examples like `xpath` and template string inteporlations, and more and more that I have not listed hand by hand. There should be a balance that whether we should place it into std or not. I personally think the standard library of a web-friendly language should provide convenience by including low-level HTML manipulations:

- Python: https://docs.python.org/3/library/html.html
- Go: https://pkg.go.dev/html/template

And I think typst is not only ambitious in composing papers but building a foundation stone to making documents to be shown everywhere (thus including web). I use "low-level" because I also believe that `html.decode` will be not directly used by most people, but packages like `@preview/xpath`, `@preview/typsify-vue`, `@preview/typsify-astro`, and so on.

Will we eventually have `html.decode` inside or outside of typst `std`?

> To parse and manipulate HTML, the XML syntax / XHTML is better suited anyways. It has less surprising special cases, is closer to the DOM, and better suited as a data-exchange format, e.g. by embedding RDF or custom namespaced elements.

Step back a bit, if I finally cannot have `html.decode`, is that meaning that I can have a `xhtml.decode`, and I can turn this PR into "XHML reading"? I think I can add xhtml support by implemented `html-decode` with some extra checking to legtimize parsed HTML into XHTML. I haven't used xhtml, but I think it could be much better than only having a `xml.decode`. Btw, I'm also in fun of making `html.parse` having introduced a html parser.

laurmaedje:
I think there's no need to support a slot API in `html.parse`. A slot-based templating could also be supported by `html.parse` returning content + a show rule on `html.elem.where(tag: "slot")`.

In my opinion, having `html.parse` that just returns content containing HTML elements is the most generally useful and unopinionated design and should be in std.

Meanwhile, an `html.parse` returning data structures for further downstream processing is a bit more niche and should, in my opinion, live outside of std in a package.

Myriad-Dreamin:
> Step back a bit, if I finally cannot have `html.decode`, is that meaning that I can have a `xhtml.decode`, and I can turn this PR into "XHML reading"? I think I can add xhtml support by implemented `html-decode` with some extra checking to legtimize parsed HTML into XHTML. I haven't used xhtml, but I think it could be much better than only having a `xml.decode`. Btw, I'm also in fun of making `html.parse` having introduced a html parser.

I'm waiting for being authorized to have an `xhtml` loader.

laurmaedje:
I must admit that I'm not yet convinced that XHTML loading should be part of the standard library. While I'm still generally favorable towards `html.decode`, I'd also defer that for now. Both things can be easily implemented as a plugin and I think that would be a better grounds for initial experimentation. Thank you still and thanks for understanding.

PR #6348, opened 2025-05-29
Add Latvian translations

kristoferssolo:
Adds Latvian translations for various document element labels.

- **figure**: `Attēls`
- **table**: `Tabula`
- **equation**: `Vienādojums`
- **bibliography**: `Izmantotā literatūra un avoti`
- **heading**: `Virsraksts`
- **outline**: `Saturs`
- **raw**: `Koda fragments`
- **page**: `lpp.`

### DIFF
--- a/crates/typst-library/src/text/lang.rs
+++ b/crates/typst-library/src/text/lang.rs
@@ -14,7 +14,7 @@ macro_rules! translation {
     };
 }
 
-const TRANSLATIONS: [(&str, &str); 39] = [
+const TRANSLATIONS: [(&str, &str); 40] = [
     translation!("ar"),
     translation!("bg"),
     translation!("ca"),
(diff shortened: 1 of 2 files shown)


laurmaedje:
I don't speak Latvian, so I double-checked with AI and got the following suggestions:

- "Literatūra" instead of "Izmantotā literatūra un avoti" appears to be common in academic writing
- "Sadaļa" is apparently more apprioriate for a section heading and "Virsraksts" would be more fitting for a book's title
- "Koda fragments" may have too strong of a connotation regarding code while "Saraksts" would be more generic.

Obviously, I can't judge these, but I would be curious to know what you think about these.

kristoferssolo:
The translation of "bibliography" varies depending on the university, and I'm not even sure how it should be done for other types of media except dissertations, so the more general term "literatūra" would be appropriate. "Sadaļa" translated as "heading" and "saraksts" as "raw", but there are no strict guidelines in this area.

laurmaedje:
In the end, you can judge it better than me. If you think your original terms are more appropriate, I'll trust you over the AI. Let me know what you decide.

kristoferssolo:
It's fine. I added them so that there would be reasonable defaults in Latvian. My suggestions were more specific to my use, but yours are more general.

laurmaedje:
Thanks!

PR #6354, opened 2025-05-30
Mention types of supported bibliography formats for citing

Andrew15-5:
I think since there [was an instance](https://forum.typst.app/t/how-do-i-use-a-hayagriva-yml-bibliography-file-with-a-typst-typ-document/4359/5?u=andrew) of confusion, the documentation should clarify this point.

The problem is that there are 2 close bibliography links, and I don't know if it's possible/better to link to a specific section for that page. Also adding a separate `.yaml` example is probably too much. But the current one can include a commented line with `works.yaml`.

P.S. I feel like https://typst.app/docs/reference/model/bibliography/#summary should probably acknowledge that you can use 2 extensions. Well actually, the [`yaml` should be preferred](https://www.iana.org/assignments/media-types/application/yaml#:~:text=File%20extension(s):%20%22yaml%22%20(preferred)%20and%20%22yml%22.). Should we change it? Might also need changes to the https://github.com/typst/hayagriva.

### DIFF
--- a/crates/typst-library/src/model/bibliography.rs
+++ b/crates/typst-library/src/model/bibliography.rs
@@ -76,6 +76,10 @@ use crate::World;
 /// | Physics         | `{"american-physics-society"}`                         |
 ///
 /// # Example
+/// Link to the BibLaTeX file from example:
+/// [`works.bib`](https://github.com/typst/typst-dev-assets/blob/main/files/bib/works.bib).
+/// For more Hayagriva examples see [this](https://github.com/typst/hayagriva/blob/main/tests/data/basic.yml).
+///
 /// ```example
 /// This was already noted by
 /// pirates long ago. @arrgh
--- a/crates/typst-library/src/model/cite.rs
+++ b/crates/typst-library/src/model/cite.rs
@@ -14,7 +14,8 @@ use crate::text::{Lang, Region, TextElem};
 /// Cite a work from the bibliography.
 ///
 /// Before you starting citing, you need to add a [bibliography] somewhere in
-/// your document.
+/// your document. Note that you can cite using BibLaTeX or Hayagriva
+/// bibliography file. For more information see [bibliography.summary].
 ///
 /// # Example
 /// ```example
--- a/crates/typst-library/src/model/cite.rs
+++ b/crates/typst-library/src/model/cite.rs
@@ -18,6 +18,10 @@ use crate::text::{Lang, Region, TextElem};
 /// bibliography file. For more information see [bibliography.summary].
 ///
 /// # Example
+/// Link to the BibLaTeX file from example:
+/// [`works.bib`](https://github.com/typst/typst-dev-assets/blob/main/files/bib/works.bib).
+/// For more Hayagriva examples see [this](https://github.com/typst/hayagriva/blob/main/tests/data/basic.yml).
+///
 /// ```example
 /// This was already noted by
 /// pirates long ago. @arrgh


laurmaedje:
I'm not convinced that this makes it much clearer to be honest. The link to the bibliography function is already there. I don't think that replicating this one piece of information there would have really helped the user.

laurmaedje:
> I feel like https://typst.app/docs/reference/model/bibliography/#summary should probably acknowledge that you can use 2 extensions. Well actually, the [yaml should be preferred](https://www.iana.org/assignments/media-types/application/yaml#:~:text=File%20extension(s):%20%22yaml%22%20(preferred)%20and%20%22yml%22.). Should we change it?

I don't have any strong opinions here. I would be open to mentioning both.

Andrew15-5:
> I'm not convinced that this makes it much clearer to be honest. [...] I don't think that replicating this one piece of information there would have really helped the user.

Do you acknowledge that a user thought that citing doesn't work with Hayagriva format and created a new post asking how to use it just because the docs didn't say that you can use it with cite function?

> I think since there [was an instance](https://forum.typst.app/t/how-do-i-use-a-hayagriva-yml-bibliography-file-with-a-typst-typ-document/4359/5?u=andrewk) of confusion, the documentation should clarify this point.

---

> The link to the bibliography function is already there.

I already outlined this:

> The problem is that there are 2 close bibliography links, and I don't know if it's possible/better to link to a specific section for that page.

---

> > I feel like https://typst.app/docs/reference/model/bibliography/#summary should probably acknowledge that you can use 2 extensions. Well actually, the yaml should be preferred. Should we change it?
>
> I don't have any strong opinions here. I would be open to mentioning both.

Okay, then I'll open a separate PR. What about typst/hayagriva? I didn't look at its docs yet.

laurmaedje:
> Do you acknowledge that a user thought that citing doesn't work with Hayagriva format and created a new post asking how to use it just because the docs didn't say that you can use it with cite function? 

The user tried to use it, so in my view that means they expected it to work. The problem was with a comma.

> I already outlined this:

Yeah, but I still think that it's suboptimal, regardless of whether it links to a subsection or not.

> What about typst/hayagriva? I didn't look at its docs yet.

I don't know off the top of my head either. Best just take a look there to see whether it's worth mentioning somewhere.

Andrew15-5:
I only have [this](https://forum.typst.app/t/how-do-i-use-a-hayagriva-yml-bibliography-file-with-a-typst-typ-document/4359/7?u=andrew) to say.

Andrew15-5:
https://forum.typst.app/t/4359/8:

> A cautionary note to those coming in from BibTeX—where a precautionary terminal comma is prescribed—to avoid that very comma in Hayagriva would be helpful.
>
> Also, a link to the [Hayagriva examples page would be helpful.](https://github.com/typst/hayagriva/blob/main/docs/file-format.md)
>
> What I was looking for was the Hayagriva source file: "works.yml" file.
A cautionary note to those coming in from BibTeX—where a precautionary terminal comma is prescribed—to avoid that very comma in Hayagriva would be helpful.

What I was looking for was the Hayagriva source file: "works.yml" file.
I think linking the [`works.bib`](https://github.com/typst/typst-dev-assets/blob/61aebe9575a5abff889f76d73c7b01dc8e17e340/files/bib/works.bib) and creating a duplicate `works.yaml` would be nice.

laurmaedje:
I think linking to more Hayagriva docs (e.g. examples) and showing a hayagriva example, too, would indeed be helpful.

Andrew15-5:
To show a Hayagriva example, you need to have a file in https://github.com/typst/typst-dev-assets, or create one through `raw` + `bytes`.

Andrew15-5:
> I don't know off the top of my head either. Best just take a look there to see whether it's worth mentioning somewhere.

Apparently there is almost nothing about this, so https://github.com/typst/hayagriva/pull/332.

A small breaking change is if someone mentioned a non-permalink, such as this:

https://github.com/typst/typst/blob/c046b75fb2cf26b2d9b77f9385a77c4aed151c56/crates/typst-library/src/model/bibliography.rs#L81

I then can link "`basic.yaml`" and not "this".

laurmaedje:
To be honest, I am not really convinced of these changes. The link to `works.bib` is a bit tricky because it would need to be updated with every typst-dev-assets bump to really show the correct file. The change at the top of `cite.rs` I don't really like for the reasons mentioned previously. And the "For more Hayagriva examples" makes it sound like the previous example is in the hayagriva file format. I also think that people don't really care too much about the concrete `.bib` file used, they care about how to invoke the bibliography function. In the YAML case, it's a bit different, because the Hayagriva format is less well known. I think we'll probably add more long form bibliography documentation as part of our docs initiative and adding a more comprehensive YAML example then would be better. Thank you still!

Andrew15-5:
> The link to `works.bib` is a bit tricky because it would need to be updated with every typst-dev-assets bump to really show the correct file.

Why would you need to update it when it's always pinned to latest?

> The change at the top of `cite.rs` I don't really like for the reasons mentioned previously.

I can remove it.

> And the "For more Hayagriva examples" makes it sound like the previous example is in the hayagriva file format.

This can be rephrased and the example wasn't added yet.

> I also think that people don't really care too much about the concrete `.bib` file used, they care about how to invoke the bibliography function.

Typst doesn't provide `lorem()` for bibliography, so concrete example for both formats is necessary, especially if you don't know them.

> In the YAML case, it's a bit different, because the Hayagriva format is less well known.

What does this mean? You think that people do really care about the concrete `.yaml` file used, yet you don't like adding Hayagriva example?

> I think we'll probably add more long form bibliography documentation as part of our docs initiative and adding a more comprehensive YAML example then would be better.

What is your "docs initiative"?

laurmaedje:
> What is your "docs initiative"?

That we plan to focus more efforts on docs in the near future. The TLDR of my comment is that discussion how I want this to look takes much more effort for me than me making a change myself in the future. Hence I closed the PR.

Issue #3131, opened 2024-01-04
Add bleed to page options


tilman-schieber:
### Description

The [bleed](https://en.wikipedia.org/wiki/Bleed_(printing)) (more descriptive is the German "Beschnittzugabe") is an area outside the printing area that is meant to be cut off after printing. This is required for all documents that have elements touching the corners (often pictures or non-white backgrounds). 

I suggest adding an option `bleed` to the page command, that increases the page size by bleed*2 in each dimension.
the margins should however be calculated based on the page size without the bleed.

Often crop or trim marks are used to indicate where the bleed area should be cut off.
![see here](https://github.com/typst/typst/assets/986273/71303cca-8f09-4978-9100-b651af6c7d11)

### Use Case

Typst's capabilities already make it a serious contender to replace DTP software like inDesign for many tasks.
Bleeds are a common feature in for-print-pdfs and should be easy to implement.
While currently one can increase the page dimensions and margins to acommodate bleeds, a dedicated option is much cleaner

jeyj0:
> Adding these could be a sensible default behaviour or be triggered by another option (as it needs adding even more extra space to the pdf output)

Here's my vote to make this another option, and not the default. In both Affinity Publisher and Adobe InDesign, the bleed options do not add marks to the page by default. It should therefore be easier for those who convert their workflow to Typst.

Additionally, in my experience it is easier for the people that need a feature to figure out how to enable it, than for those who don't need it to figure out how to disable it. If I know I need the marks, I can search for how to enable them. If I need bleed and get marks, I might not even know what they are, and therefore also wouldn't know the term to search for to disable them.

baptiste:
Finding this issue as I've been pondering the same problem here: https://discord.com/channels/1054443721975922748/1241232357994791073

I started playing with [an example](https://typst.app/project/rrNrO9NBMnLiWe5GKk_S4e), and I think something built-in in `page()` is needed: 

- bleed, trim, margin should be understood by page() so that content does not need to be offset manually
- crop marks should be an overlay, i.e. instead of being in the page background, it should be in a `page(foreground: )` argument drawn over everything

![bleeding-edge](https://github.com/typst/typst/assets/18970/bb614314-eb68-4ecd-9195-3db8f616f581)

cstrahan:
I would also like to see built in support for this.

For context: I've decided that I want some business cards, and I figured I could probably design them with Typst. I searched for bleed/trim support in Typst, and ended up here.

[...]

PR #6357, opened 2025-05-31
Add bleed support to page layout

gabriel-araujjo:
This PR adds the `bleed` option suggested in issue #3131. I have used the existing `Margin` struct.


gabriel-araujjo:
Is there anything blocking this PR from being merged that I can address?

laurmaedje:
@gabriel-araujjo Nothing to do on your end, I just didn't get to reviewing this yet.

crates/typst-pdf/src/convert.rs
```diff
@@ -81,6 +81,22 @@ fn convert_pages(gc: &mut GlobalContext, document: &mut Document) -> SourceResul
...
             );
 
+            if typst_page.bleed != Sides::splat(Abs::zero()) {
+                settings = settings
+                    .with_bleed_box(Rect::from_xywh(
+                        0.0,
+                        0.0,
+                        typst_page.frame.width().to_f32(),
+                        typst_page.frame.height().to_f32(),
+                    ))
```
laurmaedje: I think writing this is a bit redundant as it will default to the MediaBox anyway.

crates/typst-library/src/layout/page.rs
```diff
@@ -148,6 +148,47 @@ pub struct PageElem {
...
 
+    /// The page's bleed margin.
+    ///
+    /// The bleed is the area of content that extends beyond the final trimmed
+    /// size of the page. It ensures that no unprinted edges appear in the final
+    /// product, even if minor trimming misalignments occur.
+    ///
+    /// Accepted values:
+    ///
+    /// - `{auto}`: Sets the bleed to `0mm` on all sides.
```
laurmaedje: Since `auto` has no special meaning, it shouldn't be permitted. To get rid of it, you'll probably need to factor out part of the `Margin` struct. Maybe it's enough to remove the `_: AutoValue` branch and turn margin into `margin: Smart<Margin>`. I'm not 100% sure off the top of my head how this will interact with `#[fold]`, might need to double-check that it does not change the behavior.

laurmaedje:
From an initial look:

- It seems a bit inconvenient that the background does not fill the bleed area by default. This means creating a proper bleed-aware background effectively forces you to create a document with overflowing background. Maybe the bleed area should be included in the size used for background and foreground?

- I think PNG and SVG export should probably not be affected by this as the extra bleed area is only there before trimming during printing, so the canonical final page does not contain the bleed areas

baptiste:
My 2 cents:

- I agree that, for me, background and foreground should default to filling the whole area (bleed included). A very common situation – when one cares about bleed at all – is to have fill or images in the background that shouldn't be surrounded by a white border when the paper is cut, and it's natural to say `image(width:100%,height:100%)` for that to happen, rather than having to compute what `100% + bleed` means. And for the foreground, it's a natural placeholder for crop marks and alignment marks etc. that also live in that bleed/trim area.
- Perhaps it is [worth thinking](https://typst.app/project/rrNrO9NBMnLiWe5GKk_S4e) of complementing [paper size options](https://typst.app/docs/reference/layout/page/#parameters-paper) with standard untrimmed paper stocks? [E.g.,](https://en.m.wikipedia.org/wiki/Paper_size#International_raw_sizes)
```
// Untrimmed paper stock
#let A0U = (880mm , 1230mm)
#let A1U = (625mm , 880mm)
#let A2U = (450mm , 625mm)
#let A3U = (330mm , 450mm)
#let A4U = (240mm , 330mm)
 
// Corresponding trimmed size
#let A0T = (841mm , 1189mm)
#let A1T = (594mm , 841mm)
#let A2T = (420mm , 594mm)
#let A3T = (297mm , 420mm)
#let A4T = (210mm , 297mm)
```
where perhaps the corresponding bleed might default to the difference? (so that a user selecting `set page(paper = "A4U")` wouldn't need to worry about doing the calculation, the page content would automatically be set to A4T)
- While exporting SVG or png probably doesn't need the bleed area to show, it would be useful for the preview on the app to have a toggle to show/hide the bleed area (if nothing else, to preview elements such as the crop marks, registration marks, etc. which might be custom made by the user).

gabriel-araujjo:
> * It seems a bit inconvenient that the background does not fill the bleed area by default. This means creating a proper bleed-aware background effectively forces you to create a document with overflowing background. Maybe the bleed area should be included in the size used for background and foreground?

Indeed, it would be more convenient if the page background and foreground took the bleed into account. I’ve pushed a change to make that happen.

At first, I understood that changes to the bleed size were not supposed to affect any content.

crates/typst-library/src/layout/page.rs
gabriel-araujjo: I made this change, and it ended up breaking a test that used `margin: (rest: auto)`. I updated it to use an absolute value instead.

gabriel-araujjo:
> * Perhaps it is [worth thinking](https://typst.app/project/rrNrO9NBMnLiWe5GKk_S4e) of complementing [paper size options](https://typst.app/docs/reference/layout/page/#parameters-paper) with standard untrimmed paper stocks? [E.g.,](https://en.m.wikipedia.org/wiki/Paper_size#International_raw_sizes)
> 
> 
> ```
> // Untrimmed paper stock
> #let A0U = (880mm , 1230mm)
> #let A1U = (625mm , 880mm)
> #let A2U = (450mm , 625mm)
> #let A3U = (330mm , 450mm)
> #let A4U = (240mm , 330mm)
>  
> // Corresponding trimmed size
> #let A0T = (841mm , 1189mm)
> #let A1T = (594mm , 841mm)
> #let A2T = (420mm , 594mm)
> #let A3T = (297mm , 420mm)
> #let A4T = (210mm , 297mm)
> ```

I don’t have much knowledge about paper sizes. Could we open a separate issue for this and address it in a different PR?

crates/typst-library/src/layout/page.rs
laurmaedje: I think there was a misunderstanding here. I didn't want to change the existing behaviour of `margin` (allowing auto in the dictionary). I just wanted to prevent accepting `auto` for `bleed`. Perhaps that's a little tricky to achieve with good code reuse, but that would make the most sense conceptually imo. Maybe it could be done with a bit of generics of the inner type in the dictionary.

crates/typst-layout/src/pages/finalize.rs
```diff
@@ -34,30 +35,33 @@ pub fn finalize(
     }
 
     // Create a frame for the full page.
-    let mut frame = Frame::hard(inner.size() + margin.sum_by_axis());
+    let mut frame =
+        Frame::hard(inner.size() + margin.sum_by_axis() + bleed.sum_by_axis());
```
laurmaedje: I think the frame should not include the bleed in its size. That would automatically make the PNG and SVG and other non-bleed-aware exports ignore it. Then, specifically during PDF export we could take the `bleed` into account.

laurmaedje: The existing PNG tests would really test the feature anymore. PDF export is still largely untested anyway at the moment, so it would be acceptable to just keep one test to verify that PNG export is _not_ affected.

baptiste: > I think the frame should not include the bleed in its size. That would automatically make the PNG and SVG and other non-bleed-aware exports ignore it. Then, specifically during PDF export we could take the `bleed` into account.

But how will one preview things drawn in the trim area on the webapp or others relying on png/svg? I'm thinking of custom crop marks, alignment or colour palettes outside the content - it's unfortunate if one needs to export to pdf to see those (e.g people designing custom marks to go there)?

laurmaedje: We could potentially add a `RenderOptions` struct to PNG export, but the default imo should definitely exclude bleed. But even if you have that option, next up you'll probably want to draw a color border, etc. 

It's also possible to post-process the pages to include the bleed in the web app code, and if the web app adds extra stuff to render a trim box border etc. it might even be easier to handle it there.

bastienvoirin: FWIW I think that the `RenderOptions` struct for bitmap/vector graphics targets excluding slug and bleed _by default_ is the way to go, ideally with a way to toggle it from the web app (via `typst-ide` for third-party Typst editors?).

crates/typst-library/src/layout/page.rs
gabriel-araujjo: I'll try to go with this approach of using a generic for `Margin` so it only behaves differently in the case of bleed.

crates/typst-layout/src/pages/finalize.rs
gabriel-araujjo: Good point, @laurmaedje, @baptiste, and @bastienvoirin. I'm going with an implementation where bleed won't be rendered by default in PNGs and SVGs, letting that behavior be customized through `RenderOptions`.

terefang:
if someone does implement this i would suggest implementing the full complement of PDF *Boxes

image

aside from the MediaBox that would be CropBox, BleedBox, TrimBox, and ArtBox. (ISO 32000-2:2020(E) – 14.11.2.1 Page boundaries | General)

maybe the ArtBox is of less importance than the otherones.

laurmaedje:
I don't remember all of the context anymore, but I remember looking into this a bit and I think they don't all really make sense for us. This warrants further discussion and should not block this PR in my opinion.

terefang:
> This warrants further discussion and should not block this PR in my opinion.

IMHO this PR satisfies more than the 20:80 rule ... about 90% of printservices i know would accept it.

thumbs up !

gabriel-araujjo:
I’ve just added two commits:

1. Make the `Margin` struct generic over its length type.
2. Add the option to render bleed for SVG and PNG outputs.

reknih:
Okay, I first had a look at all the PDF boxes and their relationships:

- **MediaBox:** This is the dimensions of the paper you print on
- **CropBox:** All contents outside of this box must be invisible on all media, screens and paper. Nothing is rendered outside of it, instead clipping everything at its edge. It does _not_ imply a cut. The content placed inside of it but outside of the next box are production process artifacts like registration marks: They cannot appear in the final, trimmed document, even if the trim goes really badly. Controlling this box independently from the BleedBox is hence only relevant if _you_ are the professional printing service and need to put content there. There are `PrinterMark` annotations in the PDF spec (1.4+) that are intended to model these registration marks. They have the advantage of being optional and can be placed and toggled independently from the main page content. Typst currently does not have support for annotations other than Links, but these marks are commonly embedded into the page content stream anyways (Adobe Illustrator, for example does this). While PDF/X allows these annotations, it also says this: 
  > A PDF/X-4 conforming reader may completely ignore annotations except for PDF TrapNet annotations.
- **BleedBox:** Your graphics can extend to the border of this box. It is useful because the trim next box implies cannot always be 100% precise in the real world. Therefore, we want to have some area of the page where graphics that are intended to fill the whole page should extend to. For the page foreground and background this means that their relative sizing should be defined in terms of the bleed box. It is also conceivable that we want a `scope` for `place` with which we can place content relative to the BleedBox.
- **TrimBox:** The size of the medium that the customer gets. After starting out with MediaBox-sized paper, we cut off along this box to get a nice, page-spanning layout.
- **ArtBox:** A somewhat vague concept: _Meaningful content_ should go inside this box. For Typst, this could be the page minus its margins. PDF/X for example makes the ArtBox and the TrimBox mutually exclusive and recommends the TrimBox over it, likely because it is more tangible.

Furthermore, PDF/X says MediaBox >= CropBox >= BleedBox >= (TrimBox | ArtBox).

With this PR, the TrimBox is well handled: It is defined relative to the page size like margins are and the relative sizes are handled correctly. A `place` scope is missing, but that is better left to another PR. What is missing is documentation:

- `page.size` should document that it is the size of the final page, after any trims applied and can be smaller than size of the page inserted into the printer in professional printing setups.
- `page.background` and `page.foreground` should document the changed semantics for their relative lengths

I recommend the implementation of an explicit MediaBox and CropBox to be left to another PR. Design here is blocked on addition of place scopes and deciding (and perhaps implementing) whether to put these in annotations or the main content stream.

laurmaedje:
> A place scope is missing, but that is better left to another PR.

This first needs more consideration and design as well since there is not even a page scope yet.

gabriel-araujjo:
More commits:
1. Change documentation
2. Refactor render options to use consistent naming across modules

crates/typst-cli/src/args.rs
```diff
@@ -299,6 +299,12 @@ pub struct CompileArgs {
     #[arg(long = "ppi", default_value_t = 144.0)]
     pub ppi: f32,
 
+    /// By default, PNG and SVG outputs exclude the bleed from the final
+    /// artwork. Enable this flag to include the bleed in the final output.
+    /// This option has no effect when exporting other formats.
+    #[arg(long = "render-bleed")]
+    pub render_bleed: bool,
```
laurmaedje: I actually didn't want to expose this in the CLI. I don't think it's particularly useful (especially without the information how large the bleed is) and it feels out of place to me.

Having it in the `typst-svg` and `typst-render` crates is a different story as those are also used by IDE tooling and there it has a use.

laurmaedje (changes requested):

crates/typst-cli/src/compile.rs
```diff
@@ -86,6 +86,8 @@ pub struct CompileConfig {
     /// Server for `typst watch` to HTML.
     #[cfg(feature = "http-server")]
     pub server: Option<HtmlServer>,
+    /// Image render options
```
laurmaedje: Could you also revert these other now unnecessary changes? Such that the only changes in this file are
```rs
let opts = typst_render::RenderOptions {
    pixel_per_pt: config.ppi / 72.0,
    render_bleed: false,
};
let pixmap = typst_render::render(page, opts);
```
and
```rs
let opts = typst_svg::SvgOptions { render_bleed: false };
let svg = typst_svg::svg(page, opts);
```

crates/typst-svg/src/lib.rs
```diff
@@ -24,14 +24,23 @@ use xmlwriter::XmlWriter;
 use crate::paint::{GradientRef, SVGSubGradient, TilingRef};
 use crate::text::RenderedGlyph;
 
+#[derive(Clone, Copy, Default)]
```
laurmaedje:
```suggestion
#[derive(Default, Clone)]
```
- `Default` comes first in the codebases style (unfortunately not enforced by rustfmt)
- Regarding `Copy`, see https://github.com/typst/typst/pull/6357#discussion_r2671702891

laurmaedje: could you please add brief documentation here and also to the `render_bleed` field?

crates/typst-render/src/lib.rs
```diff
@@ -7,22 +7,39 @@ mod text;
...
+    Size, Transform,
 };
 use typst_library::visualize::{Color, Geometry, Paint};
 
+/// Render options image formats
+#[derive(Clone, Copy)]
+pub struct RenderOptions {
+    pub pixel_per_pt: f32,
+    pub render_bleed: bool,
+}
```
laurmaedje: Could you please add documentation for the individual fields? You can take inspiration from `PdfOptions` for the style. 

For the top-level comment, "Settings for raster image export." would suffice. The current comment is grammatically incomplete, but also a bit inaccurate because SVG is also an image format. 

Please also move the structs here (and in SVG) below the top level `pub fn`(s), similar as to how it's done in PDF export.

crates/typst-library/src/layout/page.rs
```diff
@@ -54,7 +54,12 @@ pub struct PageElem {
     #[default(Paper::A4)]
     pub paper: Paper,
 
-    /// The width of the page.
+    /// The width of the final page, after any trims have been applied.
+    ///
+    /// In professional printing setups, this may be smaller than the sheet size
+    /// fed into the printer.
+    ///
+    /// See the `bleed` argument for details on how to set a trim or bleed area.
```
laurmaedje:
```suggestion
    /// See the [`bleed`]($page.bleed) argument for details on how to set a trim
    /// or bleed area.
```

crates/typst-library/src/layout/page.rs
```diff
@@ -152,7 +157,46 @@ pub struct PageElem {
...
+    ///   - `inside`: Bleed on the inner side of the page (next to
+    ///     [binding]($page.binding)).
+    ///   - `outside`: Bleed on the outer side of the page (opposite the
+    ///     [binding]($page.binding)).
+    ///   - `x`: Horizontal bleed (applies to both left/right or inside/outside).
+    ///   - `y`: Vertical bleed (applies to both top and bottom).
+    ///   - `rest`: Default bleed for any sides not explicitly set.
+    ///
+    /// Note: The keys `left` and `right` are mutually exclusive with `inside` and
+    /// `outside`.
```
laurmaedje: I don't have a strong preference on either one, but I think it would be good to align the used language with `page.margin`. Right now, both are phrased fairly differently in each bullet point.

crates/typst-library/src/layout/page.rs
```diff
@@ -152,7 +157,46 @@ pub struct PageElem {
...
+    ///
+    /// On PDF output, if the bleed is non-zero, a `TrimBox` is defined for the page.
```
laurmaedje:
```suggestion
    /// In PDF output, if the bleed is non-zero, a `TrimBox` is defined for the page.
```

crates/typst-library/src/layout/page.rs
```diff
@@ -152,7 +157,46 @@ pub struct PageElem {
...
+    ///
+    /// ```example
```
laurmaedje, line 203: In the example, the bleed will not be visible (at least once the other comments are addressed). Perhaps this is worth calling out in explanatory text?

It might also be worth (a) adding an explicit margin and (b), using the same units for all configurations so that it's easier to reason about.

crates/typst-library/src/layout/page.rs
```diff
@@ -344,6 +388,11 @@ pub struct PageElem {
     /// This content will be placed behind the page's body. It can be
     /// used to place a background image or a watermark.
     ///
+    /// For convenience, [relative lengths]($relative) are resolved against the
+    /// page size including the page `bleed` when used in  background content.
+    /// For example, on a page that is `100mm` wide with a `5mm` bleed, a width
+    /// of `100%` is computed as `5mm + 100mm + 5mm`.
```
laurmaedje:
```suggestion
    /// For convenience, [relative lengths]($relative) are resolved against the
    /// page size including the [page `bleed`]($page.bleed) when used in
    /// background content. For example, on a page that is `{100mm}` wide with a
    /// `{5mm}` bleed, a width of `{100%}` is computed as `{5mm + 100mm + 5mm}`.
```

crates/typst-library/src/layout/page.rs
```diff
@@ -489,6 +541,8 @@ pub struct PagedDocument {
 pub struct Page {
     /// The frame that defines the page.
     pub frame: Frame,
+    /// The bleed amount to be added on each side of the page.
```
laurmaedje: This comment should then explain that the bleed is not included in `frame` and that any content in the bleed area will overflow the `frame`.

crates/typst-library/src/layout/rel.rs
```diff
@@ -30,6 +30,12 @@ use crate::layout::{Abs, Em, Length, Ratio};
 /// #rect(width: 25% + 1cm)
 /// ```
 ///
+/// For contents on page [`background`]($page.background) and
```
laurmaedje:
```suggestion
/// For contents in the page [`background`]($page.background) and
```

crates/typst-library/src/layout/rel.rs
```diff
@@ -30,6 +30,12 @@ use crate::layout::{Abs, Em, Length, Ratio};
...
+/// [`foreground`]($page.foreground), relative lengths are resolved against the
+/// page size including [bleed]($page.bleed). This semantic choice is made for
```
laurmaedje:
```suggestion
/// page size including [bleed]($page.bleed). This choice is made for
```
I don't think this word adds much here

crates/typst-render/src/lib.rs
```diff
@@ -7,22 +7,39 @@ mod text;
...
+pub fn render(page: &Page, opts: RenderOptions) -> sk::Pixmap {
+    let bleed = if opts.render_bleed { page.bleed } else { Sides::default() };
+
+    let size = page.frame.size() + bleed.sum_by_axis();
+    let pixel_per_pt = opts.pixel_per_pt;
     let pxw = (pixel_per_pt * size.x.to_f32()).round().max(1.0) as u32;
     let pxh = (pixel_per_pt * size.y.to_f32()).round().max(1.0) as u32;
 
     let ts = sk::Transform::from_scale(pixel_per_pt, pixel_per_pt);
+
```
laurmaedje:
```suggestion
```

crates/typst-render/src/lib.rs
```diff
@@ -7,22 +7,39 @@ mod text;
...
+}
+
 /// Export a page into a raster image.
 ///
 /// This renders the page at the given number of pixels per point and returns
 /// the resulting `tiny-skia` pixel buffer.
 #[typst_macros::time(name = "render")]
-pub fn render(page: &Page, pixel_per_pt: f32) -> sk::Pixmap {
-    let size = page.frame.size();
+pub fn render(page: &Page, opts: RenderOptions) -> sk::Pixmap {
```
laurmaedje: For consistency with PDF export and future proofness, it would be good if this would take `&RenderOptions` (also in SVG export). I would also prefer to remove the `Copy` derive.

While we could even add `#[non_exhaustive]` to the options, I think that would be counter-productive as it's good to get an error when a new option is available as you need to make a choice.

tests/suite/layout/page.typ
```diff
@@ -394,3 +394,23 @@ text
 #metadata(none)
 #set page(fill: red)
 2
+
+--- page-bleed paged ---
```
laurmaedje, line 417: With the corrections to PNG export, these tests will be somewhat ineffective. We could decide to then either enable bleed rendering for tests _or_ we could just not do it and rely on the upcoming hash-based PDF testing infrastructure once it's in place. I think I would prefer the latter as it's more important that the default (`render_bleed: false`) works correctly. I suppose we could also add an option (akin to `paged`), but I'd rather not add that extra burden to this PR.

crates/typst-library/src/layout/page.rs
```diff
@@ -558,17 +603,31 @@ impl Fold for Margin {
...
-    Margin,
-    self => {
+impl<T: Reflect + PartialEq> Reflect for Margin<T> {
+    fn input() -> CastInfo {
+        T::input() + Dict::input()
+    }
+    fn output() -> CastInfo {
+        Self::input()
+    }
+    fn castable(value: &Value) -> bool {
```
laurmaedje, line 631:
```suggestion
    }
    
    fn output() -> CastInfo {
        Self::input()
    }
    
    fn castable(value: &Value) -> bool {
```

crates/typst-library/src/layout/page.rs
```diff
@@ -524,32 +578,23 @@ impl Page {
 }
 
 /// Specification of the page's margins.
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
-pub struct Margin {
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct Margin<T: PartialEq> {
```
laurmaedje, line 598: This looks good now :)

laurmaedje (commented):
I'm sorry for the long radio silence. I finally got around to giving this another proper review.

---

What I noticed is that the implementation does not seem to work correctly for PNG and SVG. When bleed is disabled in the options, it's still included and when it's enabled, it's included twice. This is because the bleed is already included in `page.frame.size()`, but then also added on top. This can also be observed in the added tests, which show the bleed area, even though `render_bleed` is set to `false` in the test runner.

Beyond that, any bleed-unaware rendered would show the full page with bleed, which I think is undesirable. As suggested in https://github.com/typst/typst/pull/6357#discussion_r2247957842, I think the proper way to fix this is to have the `page.frame` _not_ include the bleed area (anything that's in the bleed area would then overflow the page frame), and bleed-aware exporters can then decide to still consider that content by expanding + translating the frame size and origin as necessary.

---

I also left some further comments. Once the main concern and these are addressed, this should finally be ready to land. :)

laurmaedje:
@gabriel-araujjo Acknowledging that the radio silence from my side was much longer, I was still wondering whether you planned to pick this up again or not. Either way is fine with me.

gabriel-araujjo:
@laurmaedje Sorry for the long wait. I only had time to revisit this code this week. In the latest version of this PR, I've applied all your suggestions.

jan-Kulpas:
I'm not sure yet as I don't have much experience printing but I think TrimBox and BleedBox mentioned above might be needed for the PDF file to work properly with professional printing software?

laurmaedje (approved):

laurmaedje:
Thank you! Good to see this finally land, it's been a while in the making. :D

PR #6362, opened 2025-06-01
Add XHTML support to the HTML output

John-Toohey:
This is a minimal PR to implement an XML variant to the HTML output.

I have put it in Draft status until the "Questions" below are answered.

### Use Case

XHTML, whilst still being valid HTML, is much easier to use with other tools/libraries --- for practically every language has a high quality XML library, but not all have as good a HTML library.

### The PR

- This PR makes minimal changes.
- Currently it introduces a `--xml` option which is used in conjunction with `--format=html` to ensure that all HTML tags close or self-close.
- The `typst_html::html` function now takes an `xml` argument. This is set as `typst_html::Writer::xml`, which introduces an if gate in `write_element` to prevent frestanding null tags (e.g. non-closed `meta` tags; in the future `img` tags as well.)

### Questions

- Is the CLI interface good? Would it make more sense to be implemented as a `--format=xhtml`? My current thinking is "no", just simply because of how similar the two outputs are and how they share nearly all of their code.
- Rather than take a `xml: bool` argument, should `typst_html::html` instead take some struct as its arguments which would be more extensible to future variants on the html format.
- Should there be tests implemented to check this feature. My thought would be yes; but I am not sure how you would like this implemented with the current `OutputType` api, given there is no `XhtmlDocument` struct or the like. Should there be one?

### Example

Consider the following Typst document:

```typ
= Hello, World!

Here is some text.
```

Compiled with `typst c --features=html --format=html`

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <h2>Hello, World!</h2>
    <p>Here is some text.</p>
  </body>
</html>
```

Compiled with `typst c --features=html --format=html --xml`:

```xhtml
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <meta name="viewport" content="width=device-width, initial-scale=1"/>
  </head>
  <body>
    <h2>Hello, World!</h2>
    <p>Here is some text.</p>
  </body>
</html>
```

Notice the `meta` tags self closed. I can now read this with an XML library.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -271,6 +271,10 @@ pub struct CompileArgs {
     /// apart from file names and line numbers.
     #[arg(long = "timings", value_name = "OUTPUT_JSON")]
     pub timings: Option<Option<PathBuf>>,
+
+    /// Whether to output valid XML for HTML export
+    #[clap(long = "xml")]
+    pub xml: bool,
 }
 
 /// Arguments for the construction of a world. Shared by compile, watch, and
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -75,6 +75,8 @@ pub struct CompileConfig {
     /// Server for `typst watch` to HTML.
     #[cfg(feature = "http-server")]
     pub server: Option<HtmlServer>,
+    /// Whether to create valid XML in HTML output
+    pub xml: bool,
 }
 
 impl CompileConfig {
@@ -158,6 +160,7 @@ impl CompileConfig {
             export_cache: ExportCache::new(),
             #[cfg(feature = "http-server")]
             server,
+            xml: args.xml,
         })
     }
 }
@@ -237,7 +240,7 @@ fn compile_and_export(
 
 /// Export to HTML.
 fn export_html(document: &HtmlDocument, config: &CompileConfig) -> SourceResult<()> {
-    let html = typst_html::html(document)?;
+    let html = typst_html::html(document, config.xml)?;
     let result = config.output.write(html.as_bytes());
 
     #[cfg(feature = "http-server")]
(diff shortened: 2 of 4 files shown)


frozolotl:
In my opinion, `--format xhtml` is the correct choice because it is a separate output format.

The fact that HTML and XHTML export share a lot of code is irrelevant here. EPUB for instance should also be implemented using a `--format epub` option, even though it would share a lot of code with XHTML export.

Furthermore, `--format xhtml` is more discoverable and doesn't require a context dependent `--xml` flag.

John-Toohey:
I can understand where you are coming from and am working on a patch to implement it as a separate format.

However, I disagree that XHTML is a 'separate' format. I think it is more comparable to the different PDF standards: XHTML is (in this case) really is just a variant of HTML. My main motivation for why it should be implemented as a variant is that it can be _layered_ with other variants (if there were any) and that the `target` function should still return `"html"`.

However, I do recognize that this _could_ solve the problem mentioned in my third question.

I would like to wait for the opinion of a maintainer before committing to one way or the other.

John-Toohey:
Okay, so I have implemented a mockup of XHTML as its own format on the branch `xhtml-as-separate-fmt`, should it be decided that that is the way to go.

reknih:
I have researched modern XHTML a bit because I suspected that this PR would not produce conformant documents as-is. Issue #6446 specifies the requirements as I saw them in greater detail, but in short, I think this PR changes in these areas:

- Character encoding
- Handling of elements with special parsing modes
- Tag omission in tables
- XML declaration and namespaces
- Compatibility behavior (ensure to print a space before the slash in the self-closing tag)

laurmaedje:
As proper XHTML support turns out to be much more complex than this, I'll close the PR. I think it's better to tackle this a bit further down the road, when HTML export is more mature. Thank you still!

adamanteye:
I opened a new PR #8356 trying to address #6446.

Issue #3114, opened 2024-01-01
Integral symbol is too small in displayed equations when using Cambria Math font


lvjr:
### Description

I tried to use Cambria Math font with the following code:
```typst
Inline $integral_a^b f(x)=F(b)-F(a)$ and displayed
$ integral_a^b f(x)=F(b)-F(a) $

#show math.equation: set text(
  font: ("Cambria Math"),
)

Inline $integral_a^b f(x)=F(b)-F(a)$ and displayed
$ integral_a^b f(x)=F(b)-F(a) $
```
but the integral symbol in displayed equations looked too small with Cambria Math:

![typst](https://github.com/typst/typst/assets/4239949/1f30b9f2-d05b-4906-8c67-0b098324e907)

The following is the same equation in Microsoft Word:

![word](https://github.com/typst/typst/assets/4239949/89748940-d5c2-40ed-a70a-ce1bc4217317)

### Operating system

Windows

### Typst version

- [X] I am using the latest version of Typst

Enter-tainer:
```
#import "@preview/mitex:0.1.0": *
#mitex(`{\sqrt{\sum_{n=1}^\infty {\frac{10}{n^4}}}} = {\int_0^\infty \frac{2x dx}{e^x-1}} = \frac{\pi^2}{3} \in {\mathbb R}`)

#show math.equation: set text(
  font: ("Cambria Math"),
)
#mitex(`{\sqrt{\sum_{n=1}^\infty {\frac{10}{n^4}}}} = {\int_0^\infty \frac{2x dx}{e^x-1}} = \frac{\pi^2}{3} \in {\mathbb R}`)

```
https://frederic-wang.fr/update-on-open-type-math-fonts.html
<img width="685" alt="image" src="https://github.com/typst/typst/assets/25521218/9a159f84-f2bf-40ad-9eac-77a0134837e5">
<img width="653" alt="image" src="https://github.com/typst/typst/assets/25521218/23be7960-763e-43e1-96bf-98d57b161df3">

Enter-tainer:
similar issue(but seems not to be helpful): https://tex.stackexchange.com/questions/543704/integral-sign-is-too-small

Enter-tainer:
It looks like the display integral symbol is indeed larger than the inline one. But it is still smaller than expected.

Enter-tainer:
[...]

PR #6365, opened 2025-06-02
Support multiple fonts in math

### LINKED ISSUE #1193: Changing font in math operators

### LINKED ISSUE #3379: Confounding math font issues

### LINKED ISSUE #5380: Font selection for nested math equations behaves differently in 0.12.0
### Description

In 0.12.0 nested math equations are totally broken. Spacing between content and nested equations is broken, fonts in nested equations are broken too and even show rules doesn't work for nested equations.

### Reproduction URL

https://typst.app/project/rT7nFHp2H-5HMPZ6ybdAhF

### Operating system

Web app, Linux

### Typst version

### LINKED ISSUE #5929: Incorrect mathscr variant size in math formulas with attachments
### Description

Updating from Typst 0.12 to 0.13, the following implementation method (documented in [cal](https://typst.app/docs/reference/math/variants#functions-cal)) causes the `mathscr` variant to have an incorrect size when appearing in math formulas as attachments:
```typ
#let scr(it) = text(
  features: ("ss01",),
  box($cal(it)$),
)
```
Here is an example to illustrate the issue:
```typ
$
product.co^(B in scr(B) , x in B)_(B in scr(B) , x in B) cal(F) (B)
$
```
- v0.13

Image

- v0.12
  
Image

If this behavior is a bug, then it is a regression from 0.12. If this behavior is expected, then I think we should find a correct implementation method for `mathscr` and update the [documentation](https://typst.app/docs/reference/math/variants#functions-cal) for calligraphic font style in math.

### Reproduction URL

https://typst.app/project/rhHk7jmqH5JgVaZOjLCAsW

### Operating system

Web app

### Typst version

### LINKED ISSUE #6634: Math class is not preserved when font falls back
### Description

Libertinus Math does not have a glyph for `brace.double.l` and `brace.double.r`. Probably due to font fallback, this causes the spacing to be wrong in the following scenario:

```typ
// `fallback: false` not necessary.
#show math.equation: set text(font: "Libertinus Math", fallback: false)
$ brace.double.l -1 brace.double.r $
$ lr(brace.double.l -1 brace.double.r) $
```

The minus sign has spaces around it.

Wrapping the double braces in `math.class` or using a math font that supports them (such as New Computer Modern Math) solves the issue.

Tested with both Typst 0.13.1 and the June 10, 2025 development version.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

mkorje:
## Font handling

~~The default math font (new computer modern math) has been added to the text font fallback list. This means that math will always be able to render, as long as the user doesn't `#set text(fallback: false)`.~~

~~The way in which font fallback in math works is nearly identical to normal text layout. The key difference is that we only use fonts that have a MATH table present, skipping ones that don't. Note that while strings in math don't use the text font, they do go through normal text layout. So `$🍏$` will render as a tofu, whereas `$"🍏"$` will not.~~

Math layout will fallback through all available fonts, and has its own set of fallback fonts distinct from normal text.

<details>

<summary>Test changes caused by font handling</summary>

---

**math-font-fallback**: Font fall back in math only goes through fonts with a MATH table.
| Old | New |
| - | - |
| math-font-fallback-old | math-font-fallback-new |
```typ
--- math-font-fallback ---
// Test font fallback.
$ よ and "よ" and 🏳️‍🌈 $
```

---

**math-lr-scripts**: The `zws` is now just ignored, so it no longer removes the spacing after the `lr`. (i.e. the test now matches what it would look like if the `zws` weren't there.)
| Old | New |
| - | - |
| math-lr-scripts-old | math-lr-scripts-new |
```typ
--- math-lr-scripts ---
// Test interactions with script attachments.
$ lr(size: #3em, |)_a^b lr(size: #3em, zws|)_a^b
  lr(size: #3em, [x])_0^1 [x]_0^1
  lr(size: #1em, lr(size: #10em, [x]))_0^1 $
```

---

</details>

## Repr in math

`RawElem` in math is kinda weird. At the moment it basically ends up going through `layout_external` since the font has changed: https://github.com/typst/typst/blob/0a3c6939dd274f40672484695d909c2cc0d0d755/crates/typst-layout/src/math/mod.rs#L474-L481 To preserve this behaviour, I've opted to cut any processing of `RawElem` off during realisation in math mode and add a check to not set the baseline for a `RawElem`.

## Multiple fonts in math

As the name of the PR would suggest, multiple fonts now just work in math. No need for `box` hacks anymore.

<details>

<summary>New tests as examples</summary>

---

**math-font-features-switch**: You can use glyphs from a font's stylistic sets now. Notice that the math kerning with the subscripts is still working!
math-font-features-switch
```typ
--- math-font-features-switch ---
#let scr(it) = text(features: ("ss01",), $cal(it)$)
$cal(P)_i != scr(P)_i$, $cal(bold(I))_l != bold(scr(I))_l$
$ product.co_(B in scr(B))^(B in scr(bold(B))) cal(B)(X) $
```

---

**math-font-covers**: You can use covers with fonts in math now, so that specific glyphs are from a given font. Again, notice the kerning with the subscripts.
math-font-covers
```typ
--- math-font-covers ---
#show math.equation: set text(
  font: (
    // Ignore that this regex actually misses some of the script glyphs...
    (name: "XITS Math", covers: regex("[\u{1D49C}-\u{1D503}]")),
  ),
  features: ("ss01",),
)
$ cal(P)_i (X) * cal(C)_1 $
```

---

**math-op-font**: You can use any font you'd like operators and limits will work as specified.
math-op-font
```typ
--- math-op-font ---
// Test with different font.
#let colim = math.op(
  text(font: "IBM Plex Sans", weight: "regular", size: 0.8em)[colim],
  limits: true,
)
$ colim_(x -> 0) inline(colim_(x -> 0)) $
```

---

**math-op-set-font**: Notice the improved kerning between the letters in each operator (less space between `c` and `h` in `sech`, more space between `o` and `d` in `mod`) as well as the ligature.
math-op-set-font
```typ
--- math-op-set-font ---
// Test setting font.
#show math.equation: set text(weight: "regular")
#let lig = math.op("fi")
#let test = $sech(x) mod_(x -> oo) lig_1(X)$
#test
#show math.op: set text(font: "New Computer Modern")
#test
```

---

</details>

## More customisation and control

Whenever a glyph is laid out in math by Typst (e.g. the delims for matrices, the accent glyph for accents), it now goes through realisation. Essentially, this means you can now target them with show rules. Combine this with proper support for multiple fonts, and you can do all sorts of stuff (even very cursed things)!

<details>

<summary>New tests as examples</summary>

---

**math-root-show-rule-1**
math-root-show-rule-1
```typ
--- math-root-show-rule-1 ---
#show "√": set text(red, font: "Noto Sans Math")
$ root(2, (a + b) / c) $
```

---

**math-root-show-rule-2**
math-root-show-rule-2
```typ
--- math-root-show-rule-2 ---
#show "√": set text(2em)
$ sqrt(2) $
```

---

**math-root-show-rule-3**
math-root-show-rule-3
```typ
--- math-root-show-rule-3 ---
// Test cursed show rule.
#show "√": "!"
$ sqrt(2) root(2, 2) $
```

---

**math-root-show-rule-4**
math-root-show-rule-4
```typ
--- math-root-show-rule-4 ---
#show math.root: set text(red)
$ sqrt(x + y) root(4, 2) $
```

---

**math-root-show-rule-5**
math-root-show-rule-5
```typ
--- math-root-show-rule-5 ---
#show math.root: it => {
  show "√": set text(purple) if it.index == none
  it
}
$ sqrt(1/2) root(3, 1/2) $
```

---

**math-delim-show-rule-1**
math-delim-show-rule-1
```typ
--- math-delim-show-rule-1 ---
#show regex("\[|\]"): set text(green, font: "Noto Sans Math")
$ mat(delim: \[, a, b, c; d, e, f; g, h, i) quad [x + y] $
```

---

**math-delim-show-rule-2**
math-delim-show-rule-2
```typ
--- math-delim-show-rule-2 ---
#show math.vec: it => {
  show regex("\(|\)"): set text(blue)
  it
}
$ vec(1, 0, 0), mat(1; 0; 0), (1), binom(n, k) $
```

---

**math-delim-show-rule-3**
math-delim-show-rule-3
```typ
--- math-delim-show-rule-3 ---
#show "⏟": set text(fuchsia)
$ underbrace(1 + 1 = 2, "obviously") $
```

---

**math-delim-show-rule-4**
math-delim-show-rule-4
```typ
--- math-delim-show-rule-4 ---
#show "{": set text(navy)
$ cases(x + y + z = 0, 2x - y = 0, -5y + 2z = 0) $
```

---

**math-delim-show-rule-5**
math-delim-show-rule-5
```typ
--- math-delim-show-rule-5 ---
#show regex("\(|\)"): set text(1.5em)
$ 10 dot (9 - 5) dot (1/2 - 1) $
```

---

**math-primes-show-rule**
math-primes-show-rule
```typ
--- math-primes-show-rule ---
#show math.primes: set text(maroon)
$f'(x), f''''''(x)$
```

---

**math-glyph-show-rule**
math-glyph-show-rule
```typ
--- math-glyph-show-rule ---
#show "+": set text(orange, font: "Noto Sans Math")
$ 1 + 1 = +2 $
#show "+": text(2em)[#sym.plus.circle]
$ 1 + 1 = +2 $
```

---

**math-accent-show-rule-1**
math-accent-show-rule-1
```typ
--- math-accent-show-rule-1 ---
#show "\u{0302}": set text(blue, font: "XITS Math")
$hat(x)$, $hat(hat(x))$, x\u{0302}
```

---

**math-accent-show-rule-2**
math-accent-show-rule-2
```typ
--- math-accent-show-rule-2 ---
#let rhat(x) = {
  show "\u{0302}": set text(red)
  math.hat(x)
}
$hat(x)$, $rhat(x)$, $hat(rhat(x))$, $rhat(hat(x))$, x\u{0302}
```

---

**math-accent-show-rule-3**
math-accent-show-rule-3
```typ
--- math-accent-show-rule-3 ---
#show math.accent: it => {
  show "\u{0300}": set text(green)
  it
}
$grave(x)$, x\u{0300}
```

---

**math-accent-show-rule-4**
math-accent-show-rule-4
```typ
--- math-accent-show-rule-4 ---
#show "\u{0302}": box(inset: (bottom: 5pt), text(0.5em, sym.diamond.small))
$hat(X)$, $hat(x)$
```

---

</details>

## Issues

- Closes #3379.
- Closes #1193.
- Closes #5929.
- Closes #6634.
- Closes #5380.


istudyatuni:
Why is raw emoji showing but `#emoji` is not?

Enivex:
> Why is raw emoji showing but `#emoji` is not?

I don't know exactly what you're referring to, but it probably has to do with math fonts not having emojis in them.

istudyatuni:
When written as `$🍏$`, emoji is rendered, but when as `$#emoji.apple$`, is not

mkorje:
> I don't know exactly what you're referring to, but it probably has to do with math fonts not having emojis in them.

Yes, that's why.

> When written as `$🍏$`, emoji is rendered, but when as `$#emoji.apple$`, is not

I can't reproduce this, they both render as tofus on this branch.

mkorje:
I've opened https://github.com/typst/typst-dev-assets/pull/7 to add another math font to the test suite.

crates/typst-layout/src/math/accent.rs
```diff
@@ -21,44 +21,54 @@ pub fn layout_accent(
...
-    let mut glyph =
-        GlyphFragment::new_char(ctx.font, accent_styles, accent.0, elem.span())?;
+    let mut accent = ctx.layout_into_fragment(
+        &SymbolElem::packed(accent.0).spanned(elem.span()),
+        accent_styles,
+    )?;
 
     // Forcing the accent to be at least as large as the base makes it too wide
     // in many cases.
     let width = elem.size.resolve(styles).relative_to(base.width());
```
mkorje, line 54: This should use the `base_styles` and `font_size`, but I'll leave it for https://github.com/typst/typst/pull/6214

crates/typst-layout/src/math/mod.rs
```diff
@@ -108,12 +103,12 @@ pub fn layout_equation_block(
...
     let span = elem.span();
-    let font = find_math_font(engine, styles, span)?;
 
     let mut locator = locator.split();
-    let mut ctx = MathContext::new(engine, &mut locator, regions.base(), &font);
+    let mut ctx = MathContext::new(engine, &mut locator, regions.base());
 
-    let scale_style = style_for_script_scale(&ctx);
+    let font = find_math_font(ctx.engine.world, styles, elem.span())?;
+    let scale_style = style_for_script_scale(&font);
```
mkorje: This is a bit problematic. Since the font can change this should get updated whenever it does...

crates/typst-layout/src/math/mod.rs
```diff
@@ -53,12 +49,11 @@ pub fn layout_equation_inline(
...
-    let font = find_math_font(engine, styles, elem.span())?;
-
     let mut locator = locator.split();
-    let mut ctx = MathContext::new(engine, &mut locator, region, &font);
+    let mut ctx = MathContext::new(engine, &mut locator, region);
 
-    let scale_style = style_for_script_scale(&ctx);
+    let font = find_math_font(ctx.engine.world, styles, elem.span())?;
+    let scale_style = style_for_script_scale(&font);
     let styles = styles.chain(&scale_style);
```
mkorje: Similar here, but there is also the added issue that the font gets used to set the bounding box for inline math content:
https://github.com/typst/typst/blob/52a708b988cf7d13898194e886790acb7edd510f/crates/typst-layout/src/math/mod.rs#L83-L88

crates/typst-layout/src/math/text.rs
```diff
@@ -120,45 +123,33 @@ pub fn layout_symbol(
     // Switch dotless char to normal when we have the dtls OpenType feature.
     // This should happen before the main styling pass.
     let dtls = style_dtls();
-    let (unstyled_c, symbol_styles) = match try_dotless(elem.text) {
-        Some(c) if has_dtls_feat(ctx.font) => (c, styles.chain(&dtls)),
+    let (unstyled_c, symbol_styles) = match (
+        try_dotless(elem.text),
+        find_math_font(ctx.engine.world, styles, elem.span()),
+    ) {
```
mkorje: This check is now only a heuristic really.

I guess ideally this should use the font fallback to try and find a font with the dtls feature supported for the right styled glyph. But that is going to be complicated...

crates/typst-layout/src/math/fragment.rs
```diff
@@ -261,23 +305,70 @@ pub struct GlyphFragment {
...
+        let end = text.char_indices().nth(1).map(|(i, _)| i).unwrap_or(text.len());
+
+        // Find the next available family.
+        let world = ctx.engine.world;
+        let book = world.book();
+        let mut selection = None;
+        for family in families {
+            selection = book
+                .select(family.as_str(), variant)
+                .and_then(|id| world.font(id))
```
laurmaedje: Since https://github.com/typst/typst/pull/6316/files, the font book holds metadata on whether it's a MATH font. Could be useful here.

crates/typst-layout/src/math/lr.rs
```diff
@@ -33,12 +33,13 @@ pub fn layout_lr(
...
-        .iter()
-        .map(|fragment| (fragment.ascent() - axis).max(fragment.descent() + axis))
-        .max()
-        .unwrap_or_default();
+    let mut max_extent = Abs::zero();
+    for fragment in inner_fragments.iter() {
+        let (font, size) = fragment.font(ctx, styles, elem.span())?;
+        let axis = value!(font, axis_height).at(size);
+        let extent = (fragment.ascent() - axis).max(fragment.descent() + axis);
+        max_extent = max_extent.max(extent);
```
laurmaedje: For every non-glyph fragment, we keep recomputing the same metric here, right? This seems a bit wasteful.

crates/typst-layout/src/math/mat.rs
```diff
@@ -186,12 +186,10 @@ fn layout_body(
     // We pad ascent and descent with the ascent and descent of the paren
     // to ensure that normal matrices are aligned with others unless they are
     // way too big.
-    let paren = GlyphFragment::new_char(
-        ctx.font,
-        styles.chain(&denom_style),
-        '(',
-        Span::detached(),
-    )?;
+    // This will never panic as a paren will never shape into nothing.
```
laurmaedje, line 189: if the font has no paren, will this grab the tofu?

crates/typst-library/src/text/mod.rs
```diff
@@ -936,6 +936,7 @@ pub fn families(styles: StyleChain<'_>) -> impl Iterator<Item = &'_ FontFamily>
             "noto color emoji",
             "apple color emoji",
             "segoe ui emoji",
+            "new computer modern math",
```
laurmaedje: I wonder whether we should have this in here globally or whether we should have a different default fallback list in math.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -108,6 +112,21 @@ impl MathFragment {
...
+    pub fn font(
+        &self,
+        ctx: &MathContext,
+        styles: StyleChain,
+        span: Span,
+    ) -> SourceResult<(Font, Abs)> {
+        Ok((
+            match self {
+                Self::Glyph(glyph) => glyph.item.font.clone(),
+                _ => find_math_font(ctx.engine.world, styles, span)?,
```
laurmaedje: This is probably fine, but I wanted to double-check that this always does what we want. Because for glyphs it takes the font in the glyph and for other stuff from the environment, it seems like it could lead to subtle differences between the glyph and non-glyph case.

laurmaedje (commented):
This looks very nice overall! 🎉

A few remarks:

- I think it's quite unfortunate that this defaults to tofus for stuff that isn't in a math font. Couldn't we fall back to a text font and just use some hardcoded metrics? 

- The `RawElem` thing seems a bit odd indeed. Is `RawElem` really that special or could we run into similar issues with other textual elements?

- Performance: On a very math-heavy document I tested, this regressed total running time by 20%. I'm sure we can get that perf back with just a bit of caching in the right places (potentially manual cell-based caching rather than comemo caching on the hot paths). If you have any ideas for potential speedups and they seem promising locally, just ping me and I'll start a benchmark run. I guess the main question is whether the cost is primarily the font lookups or realization.

crates/typst-layout/src/math/fragment.rs
mkorje: I'll use that instead.

mkorje:
> * The `RawElem` thing seems a bit odd indeed. Is `RawElem` really that special or could we run into similar issues with other textual elements?

I don't think we will? Any other textual thing is fine to go through the math layout normally. I mainly did it so the tests didn't change, but also setting the baseline made it look worse.

EDIT: Here's the difference with setting the baseline or not.

<details>

<summary>Images</summary>

| Not baseline | Yes baseline |
| - | - |
| issue-852-mat-type no-baseline | issue-852-mat-type yes-baseline |
| math-mat-spread no-baseline | math-mat-spread yes-baseline |
| math-mat-spread-2d no-baseline | math-mat-spread-2d yes-baseline |
| math-size-math-content-1 no-baseline | math-size-math-content-1 yes-baseline |
| math-size-math-content-2 no-baseline | math-size-math-content-2 yes-baseline |
| math-size-math-content-3 no-baseline | math-size-math-content-3 yes-baseline |

</details>

crates/typst-layout/src/math/mat.rs
mkorje: It will. (Though if there's one glyph a math font will have, it's probably a paren :sweat_smile:)

crates/typst-layout/src/math/fragment.rs
mkorje: Yeah it's a fair point. I considered attaching a font to basically every variant in `MathFragment`, but that's just moving this problem to where e.g. `FrameFragment` ends up getting created (which also causes issues IIRC, since wherever it gets created now also needs `MathContext` passed as an argument).

I believe I made sure to always call this method using the exact same styles as for the layout of the fragment, so it _should_ work as expected.

crates/typst-layout/src/math/mat.rs
mkorje: It's also only used to apply padding, so if it's a tofu (which I'm guessing would normally have smaller ascent/descent than a paren?) it'll just mean uneven matrices I suppose.

crates/typst-library/src/text/mod.rs
mkorje: I don't have much of an opinion here. Maybe its nice to have it globally in case a user uses math glyphs in markup mode?

mkorje:
> * I think it's quite unfortunate that this defaults to tofus for stuff that isn't in a math font. Couldn't we fall back to a text font and just use some hardcoded metrics?

I agree it isn't ideal. Though even if we do let it fall back, throughout the code we rely on the guarantee of a math font being attached to it (e.g. for getting constants from the math table)... Using hardcoded metrics wouldn't be too bad (we can use the [MathML Core spec](https://www.w3.org/TR/mathml-core/#layout-constants-mathconstants) ones), and we could get rid of the macros then which might make things nicer. At that point there's no need to enforce a math font at all, but I figure we should then at least warn if the current font doesn't support math?

mkorje:
> * Performance: On a very math-heavy document I tested, this regressed total running time by 20%. I'm sure we can get that perf back with just a bit of caching in the right places (potentially manual cell-based caching rather than comemo caching on the hot paths). If you have any ideas for potential speedups and they seem promising locally, just ping me and I'll start a benchmark run. I guess the main question is whether the cost is primarily the font lookups or realization.

I'll investigate a bit.

laurmaedje:
> At that point there's no need to enforce a math font at all, but I figure we should then at least warn if the current font doesn't support math?

At the top-level definitely. For a single symbol, like an emoji, it might be too much.

crates/typst-layout/src/math/fragment.rs
laurmaedje: > I believe I made sure to always call this method using the exact same styles as for the layout of the fragment, so it should work as expected.

So the point of this is less to recover the font because we don't have it anymore at this point and more to get the exact font that was used during potential fallback?

laurmaedje:
> I don't think we will? Any other textual thing is fine to go through the math layout normally. I mainly did it so the tests didn't change, but also setting the baseline made it look worse.

Why does raw get its baseline set and other textual stuff not (without special checks)? I guess I still don't understand what makes it different (except for some existing test reference images).

crates/typst-layout/src/math/fragment.rs
mkorje: Yes, basically. Though since I'll change it to allow any font I'll just be able to get the font directly off the style chain.

mkorje: Should make the code nicer and hopefully a little faster

mkorje:
> Why does raw get its baseline set and other textual stuff not (without special checks)? I guess I still don't understand what makes it different (except for some existing test reference images).

With the changes this PR makes, RawElem (which uses a separate font) no longer gets laid out in an external frame since the font changed. So the text it ends up as after realisation went through math layout like all other text. This leads to proper math spacing, etc. (And since the font fallback uses only math fonts currently, it ended up in the math font). I think this is the wrong behaviour (?), RawElem should appear in math like in markup mode. So to do this it should be laid out externally. Hence I made the changes I did in this PR to ensure that happened, while also ensuring the visual output didnt change either.

tests/ref/math-class-chars.png
mkorje, line 1: This test got updated as I removed the `.max(SQRT_2)` thing from https://github.com/typst/typst/issues/3114, and instead replaced it with a hardcoded check for cambria math (given this has basically ended up as a "bug" in the spec, I think this is okay) when adding the constants to the font metrics. (so the old image was a bit of a bug then ig).

mkorje:
@laurmaedje I changed up the font handling to instead store it on the MathContext struct and update it. This is for when a font is needed and the current MathFragment isn't a GlyphFragment (instead of the whole `find_math_font` thing I had before). It's quite similar now to how it was before this PR, so hopefully the performance regression isn't as bad. Would appreciate it if you could start a benchmark run.

My very rudimentary testing with hyperfine on my laptop says the PR is now 1.02 ± 0.06 times faster than main on a test document.

laurmaedje:
With the latest changes, the regressions is down from +20% to +5-6% on the math-heavy doc. It's just one document, so not very representative. But I can live with this number.

laurmaedje:
> (And since the font fallback uses only math fonts currently, it ended up in the math font). I think this is the wrong behaviour (?), RawElem should appear in math like in markup mode. So to do this it should be laid out externally. 

With the latest commit non-math fonts are also considered, right? So, does this still apply?

mkorje:
> With the latest commit non-math fonts are also considered, right? So, does this still apply?

They should, but for some reason they aren't... I'll be able to investigate further a bit later.

Though even still, the output will end up looking different for Raw text as it'll get math spacing, etc.. And I'm not sure if we'd want that.

> With the latest changes, the regressions is down from +20% to +5-6% on the math-heavy doc. It's just one document, so not very representative. But I can live with this number.

There's probably a couple more things I can modify to hopefully speed things up. I've also realised that the `shape` method memoization probably isn't doing much because it takes span as an argument?

laurmaedje:
> I've also realised that the shape method memoization probably isn't doing much because it takes span as an argument?

At least not within one compilation, yeah. If it can be extracted that might be better.

laurmaedje:
> Though even still, the output will end up looking different for Raw text as it'll get math spacing

Wouldn't that only be around it? As, in principle, as raw is a single run of text, I'm not sure I get what will still be different then.

crates/typst-library/src/text/mod.rs
mkorje: I've gone with a separate list so that font fallback works properly.

mkorje:
> Wouldn't that only be around it? As, in principle, as raw is a single run of text, I'm not sure I get what will still be different then.

Ok I have no idea what I was thinking before, everything is fine if I remove the special handling for `RawElem`. These are the only "unintentionally" changed tests now then.

| `main` | This PR | diff |
| - | - | - |
| math-text-size_old | math-text-size_new | math-text-size_diff |
| math-size-math-content-1_old | math-size-math-content-1_new | math-size-math-content-1_diff |
| math-font-fallback_old | math-font-fallback_new | math-font-fallback_diff |

crates/typst-layout/src/math/mod.rs
```diff
@@ -469,18 +497,27 @@ impl<'a, 'v, 'e> MathContext<'a, 'v, 'e> {
...
+                || styles.get_ref(TextElem::weight) != outer_weight
             {
-                let frame = layout_external(elem, self, styles)?;
-                self.push(FrameFragment::new(styles, frame).with_spaced(true));
-                continue;
+                self.fonts_stack.push(get_font(
+                    self.engine.world,
+                    styles,
+                    content.span(),
+                )?);
```
mkorje: I'm doing this because I think (??) calling `get_font` will have more overhead than checking the StyleChain.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -361,7 +368,7 @@ impl GlyphFragment {
             modifiers: FrameModifiers::get_in(styles),
         };
         fragment.update_glyph();
```
mkorje, line 371: I still want to find some way to memoize what this line is doing.

crates/typst-layout/src/math/lr.rs
mkorje: It's a little better now as getting the font is just a lookup.

crates/typst-layout/src/math/mod.rs
laurmaedje: Probably it will, but if you give me two commit hashes to compare, I'm also happy to compare

crates/typst-layout/src/math/fragment.rs
mkorje: Ok this might require a bit more effort to try do, and I've gotten the performance impact back down. I'll drop this for now.

crates/typst-layout/src/math/mod.rs
```diff
@@ -42,6 +45,47 @@ use self::run::{LeftRightAlternator, MathRun, MathRunFrameBuilder};
 use self::shared::*;
 use self::stretch::stretch_fragment;
 
+/// Styles to add font constants to the style chain.
+#[inline]
```
laurmaedje: on private functions, there's no need for `#[inline]`. it's only relevant cross-crate.

crates/typst-layout/src/math/mod.rs
```diff
@@ -42,6 +45,47 @@ use self::run::{LeftRightAlternator, MathRun, MathRunFrameBuilder};
...
+    let variant = variant(styles);
+    families(styles)
+        .find_map(|family| {
+            world
+                .book()
+                .select(family.as_str(), variant)
+                .and_then(|id| world.font(id))
+                .filter(|_| family.covers().is_none())
+        })
+        .ok_or(::ecow::eco_vec![error!(span, "no font could be found")])
```
laurmaedje: This error is new, right? I guess relying purely on fallback does not work with the font stack thing, but also generally fails because we need all those metrics?

This seem a bit inconsistent with how font fallback otherwise works in Typst, but I guess it is not more restrictive than before with "current font is not suitable for math".

laurmaedje:
```suggestion
        .ok_or("no font could be found")
        .at(span)
```

crates/typst-layout/src/math/mod.rs
laurmaedje: could you also move the helper functions to the bottom? (either the very bottom or below where they are used)

crates/typst-library/src/text/font/mod.rs
```diff
@@ -232,6 +239,10 @@ pub struct FontMetrics {
     pub subscript: Option<ScriptMetrics>,
     /// Metrics for superscripts, if provided by the font.
     pub superscript: Option<ScriptMetrics>,
+    /// The width of a space.
+    pub space_width: Em,
+    /// Metrics for math layout.
+    pub math: MathConstants,
```
laurmaedje: It feels a bit wrong to me to have this always be populated even if the font is never getting close to math. It could also be a `OnceLock<Box<MathConstants>>` to save all that for fonts that are never used in math. Not sure whether it's better. It could also have a non-negligible perf impact considering how often it is called in math (though I kinda doubt it actually has).

crates/typst-layout/src/math/mod.rs
```diff
@@ -42,6 +45,47 @@ use self::run::{LeftRightAlternator, MathRun, MathRunFrameBuilder};
...
+}
+
+/// Check if the top-level base font has a MATH table.
+fn warn_non_math_font(font: &Font, engine: &mut Engine, span: Span) {
+    if !font.info().flags.contains(FontFlags::MATH) {
+        engine.sink.warn(warning!(
+            span,
+            "font is not a math font";
```
laurmaedje:
```suggestion
            "current font is not designed for math";
```
- "not designed for" may be a bit friendlier than "not a math font"
- "current" because it shows at the equation and not at the set rule

crates/typst-library/src/text/font/mod.rs
```diff
@@ -335,6 +519,55 @@ pub struct ScriptMetrics {
     pub vertical_offset: Em,
 }
 
+/// Constants from the MathConstants table used in Typst.
+/// Ones not currently used are omitted.
```
laurmaedje:
```suggestion
/// Constants from the OpenType MATH constants table used in Typst.
///
/// Ones not currently used are omitted.
```

crates/typst-layout/src/math/mod.rs
mkorje: It's basically replaced the old "current font doesn't support math" error thing. Yeah, we just need to grab whatever the current font is for the metrics.

Surely though this error will basically never occur, right? (Unless the font specified doesn't exist and fallback is disabled, I guess)

crates/typst-library/src/text/font/mod.rs
mkorje: I tried it out and ran a benchmark (run 107). I'll let you decide if its non-negligible enough.

laurmaedje: Given that the impact seems to be basically zero, I'm inclined to keep the OnceLock because it feels cleaner to me and also reduces memory usage slightly.

mkorje: Alright, I'll keep it.

laurmaedje:
Thank you. This is excellent work!

Issue #3702, opened 2024-03-17
Add socks feature for ureq


panda-z:
### Description

When updating typst behind a SOCKS proxy, it fails with "Connection Failed: Connect error: SOCKS feature disabled."

Seems the socks feature is not enabled for `ureq`. Simply adding the following feature to `ureq` should resolve the problem.

```
features = [ "socks-proxy" ]
```

### Use Case

Sometimes user will update typst via SOCKS proxy.

PR #2006, opened 2023-08-25
Support proxy and custom certificate configuration.

zicklag:
Respects the `http_proxy` environment variables and adds a CLI flag / environment variable that can be used to configure a custom trusted certificate chain.

PR #6366, opened 2025-06-02
Move to ureq 3 from ureq 2

### LINKED ISSUE #6261: Typst cannot download package through proxy with IPv6 address
### Description

With `HTTPS_PROXY=http://[::1]:1081`, `typst compile file.typ` will fail with error when a package needs to be downloaded:
```
error: failed to download package (https://packages.typst.org/preview/algo-0.3.6.tar.gz: Dns Failed: resolve dns name '[:80': failed to lookup address information: Name or service not known)
  ┌─ file.typ:2:8
  │
2 │ #import "@preview/algo:0.3.6": algo, i, d, comment, code
  │         ^^^^^^^^^^^^^^^^^^^^^
```

It seems typst thought `[` is the hostname as there is a `:` following it and defaults to port 80, rather than the start of an IPv6 address.

And switching to `HTTPS_PROXY=http://127.0.0.1:1081` will make it work as expected, downloading the package.

I'm on typst version 0.13.1, and NixOS unstable.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

arielf212:
This closes #6261, but is also good in general since ureq 2 will be deprecated at some point so might as well pull the bandaid now. Also closes https://github.com/typst/typst/issues/3702.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,7 +134,7 @@ unicode-script = "0.5"
 unicode-normalization = "0.1.24"
 unicode-segmentation = "1"
 unscanny = "0.1"
-ureq = { version = "3", default-features = false, features = ["native-tls", "gzip", "json"] }
+ureq = { version = "3", default-features = false, features = ["native-tls", "gzip", "json", "socks-proxy"] }
 usvg = { version = "0.45", default-features = false, features = ["text"] }
 utf8_iter = "1.0.4"
 walkdir = "2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2618,6 +2618,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 
 [[package]]
+name = "socks"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f0c3dbbd9ae980613c6dd8e28a9407b50509d3803b57624d5dfe8315218cd58b"
+dependencies = [
+ "byteorder",
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "spin"
 version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3564,6 +3575,7 @@ dependencies = [
  "rustls-pki-types",
  "serde",
  "serde_json",
+ "socks",
  "ureq-proto",
  "utf-8",
  "webpki-root-certs 0.26.11",
@@ -3828,6 +3840,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "53a85b86a771b1c87058196170769dd264f66c0782acf1ae6cc51bfd64b39082"
 
 [[package]]
+name = "winapi"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c839a674fcd7a98952e593242ea400abe93992746761e38641405d28b00f419"
+dependencies = [
+ "winapi-i686-pc-windows-gnu",
+ "winapi-x86_64-pc-windows-gnu",
+]
+
+[[package]]
+name = "winapi-i686-pc-windows-gnu"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac3b87c63620426dd9b991e5ce0329eff545bccbbb34f3be09ff6fb6ab51b7b6"
+
+[[package]]
 name = "winapi-util"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -133,7 +132,7 @@ unicode-script = "0.5"
 unicode-normalization = "0.1.24"
 unicode-segmentation = "1"
 unscanny = "0.1"
-ureq = { version = "2", default-features = false, features = ["native-tls", "gzip", "json"] }
+ureq = { version = "3", default-features = false, features = ["native-tls", "gzip", "json"] }
```
istudyatuni: Maybe also add `socks-proxy`? #3702

crates/typst-kit/src/download.rs
```diff
@@ -107,31 +109,34 @@ impl Downloader {
...
-        if let Some(cert) = self.cert() {
-            tls.add_root_certificate(cert?.clone());
-        }
+        let maybe_cert = self.cert().transpose()?.cloned().map_or(
+            RootCerts::PlatformVerifier,
+            |cert| {
+                let certs = vec![cert];
+                RootCerts::Specific(Arc::new(certs))
+            },
+        );
```
laurmaedje, line 128: Doesn't this change the semantics from system + custom cert to just custom cert if `--cert` is passed? At least that's how I understand the ureq docs.

arielf212: yes, it does, but it doesnt seem like there is a way to keep the old behaviour as per their doc.

I also question how improtant this is, but it is technically a breaking change nonetheless.

laurmaedje:
There is also the problem that ureq's built-in proxying support does not honor the `NO_PROXY` variable, which would be a regression. See https://github.com/algesten/ureq/issues/668 and https://github.com/typst/typst/pull/2006#issuecomment-2845375208.

crates/typst-kit/src/download.rs
laurmaedje: I know too little about the topic to judge it. Maybe @zicklag who contributed the original certificate code knows more.

zicklag: Yeah, from a quick look at the docs I don't see a way to simply add the certificate either. I think that might make sense, though, because if you want to trust a custom certificate, then you may want to intentionally un-trust the default ones, too, not just add a trusted cert.

FWIW, the default certificate settings in ureq3 are to use the WebPki trust, but it might be better to use the [`PlatformVerifier`](https://docs.rs/ureq/latest/ureq/tls/enum.RootCerts.html#variant.PlatformVerifier) setting, which will use the system certificate store.

I have my custom certificate trusted at a system level, and I often have to re-compile / update certificates of programs that don't read the system certificate store.

It's more convenient when it uses the system store than it is to have to pass a custom certificate flag or environment variable.

zicklag:
I feel like the `NO_PROXY` caveat is annoying, but not usually that important.

I don't think ureq2 supported `NO_PROXY` either, so I don't _think_ it'd be a regression, but I could be wrong.

laurmaedje:
`env_proxy`, which we used before, [does support `NO_PROXY`](https://docs.rs/env_proxy/0.4.1/src/env_proxy/lib.rs.html#74-115). That's also why the code previously relied on the URL to determine the proxying behaviour.

laurmaedje:
@arielf212 Maybe we can proceed here by just keeping `env_proxy` dependency and configuring the [proxy](https://docs.rs/ureq/latest/ureq/config/struct.ConfigBuilder.html#method.proxy) manually instead of relying on `ureq` to do it?

zicklag:
Ah, that could work. I didn't notice that the proxy could be set per-request. Good thought. :+1:

laurmaedje:
@arielf212 Do you still plan to work on this? If not, no matter, but then I'd close the PR.

arielf212:
@laurmaedje Sorry, I had a bit of a rough summer 🥲 .
I really want to make this happen, but I think the best way to fix this will be to add the feature to upstream - it seems they really want it since like 2023.

Would you be willing to let this stay open for at least a month? I will try to get it up until the end of the weekend, since its a small feature :)

laurmaedje:
> Sorry, I had a bit of a rough summer

Sorry to hear that! We can of course keep it open, I'd just have closed it if it was completely stale. Nice to see that you are adding upstream support.

One thing I'm wondering though: What's the problem with continuing to use `env_proxy` with ureq 3? (https://github.com/typst/typst/pull/6366#issuecomment-2965978928)

arielf212:
I mainly dislike having to compile and carry around an additional dependency that does nothing of value to us anymore _except_ this one - which is at least to me - minor thing.

This will both help with binary size and also with security, since there will be fewer crates we depend on which can supply chain attacked - even tho that's probably a very minor addition.

laurmaedje:
I'm happy to wait a bit for NO_PROXY to land in ureq. If it doesn't materialize in a while, we can still use env_proxy.

JeppeKlitgaard:
I think this landed in ureq 3.1.3 with https://github.com/algesten/ureq/pull/1118

laurmaedje:
I'll close this for now due to inactivity. We can reopen it or have a new PR when there is new activity. Thanks nonetheless!

Issue #2636, opened 2023-11-09
Allow importing from dictionaries with `#import dict: <keys>` to allow them to function as dynamic modules


yaksher:
### Description

Currently dictionaries already implicitly allow conversion between identifiers and strings of those identifiers since their literals use bare words for keys and you can use `.key` syntax for accessing or binding fields, but it also supports `.insert()` and what not which take string keys.

This feature request asks that dicts go one step further and allow you to bind all their values directly into the namespace, which would neatly allow programmatically generated lists of names.

I think this fits with both the `#import "...": *` syntax and the fact that dictionaries already effectively serve as a map from identifiers to strings and vice versa.

I think reasonable syntax would either be just `#import dict: <comma separated keys or *>` to match importing, since modules already behave similarly to dictionaries, or `#let * = dict`, as an extension of destructuring binding syntax.

### Use Case

One of my use cases for Typst requires a ton of calligraphic letters and led me to define `AC`, `BC`, ... to be calligraphic versions of `A`, `B`, ... which required 26 distinct bindings (well, I used array destructuring, but I still had to write all of the names).

In the past when using LaTeX, I've often found it to be convenient shorthand to avoid typing braces to use macros to define a large number of commands. This would fill the same role without requiring any qualitatively new.

PgBiel:
I was thinking about this before after a discussion in the Typst server. This could be a way to achieve what's proposed in #2403; in general, you could have "configure" functions in packages which return a configured module, and this "dynamic module" could just be a dict. (There is also the separate idea of using some special syntax for scope objects, but they could happen to not be dynamic. However, that could still remain a possibility.)

In that case, though, I'd personally prefer if, instead of changing the syntax for the specific case of wildcard import (which, IMHO, shouldn't be "recommended" in general, as it makes your document - IMHO - less readable), we could just have `import` accept dicts (it already accepts arbitrary expressions anyway). That would give us all the benefits from `import` automatically (wildcard import, `as`, etc.) without having to change much in the language (0 changes to syntax, as a matter of fact).

Still, I believe this needs further discussion before going forward.

yaksher:
Huh, I thought I edited the title. I agree that importing from dictionaries is a much more cohesive syntax than the `let * =` binding, which was just my first thought.

PgBiel:
For comparison and further discussion, there is already support for destructuring in the form of

```
#let (name1, name2, name3: renamed) = dictionary
```

[...]

PR #6371, opened 2025-06-03
Turn dictionaries into modules, and accept dictionaries as import sources

maugier:
Hello,

  After a short discussion with @SillyFreak on discord, I had a go at trying to address #2636.

 I first attempted to add a `module()` constructor, but I wasn't sure how to add a constructor without adding a `#[ty(scope)]` on `module`, which I doubt is correct.

  Instead, I added a `.module()` method on dicts (exposed as a typst method so far, but that is debatable)
and added a clause for dict-valued import sources.

  Here's a simple example:
```typ
#let pdict(left, right) = {
    (wrap: (x) => [#left #x #right],
     unit: [#left #right])
}

#let pmod = pdict.module()

// equivalent to let (wrap, unit) = parametric([\(], [\)])
#import pmod([\(], [\)]): wrap, unit

// Pulls all dict keys into scope without having to repeat the names
#import pmod([\(], [\)]): *
```

How does that look ?

Cheers,

### DIFF
--- a/crates/typst-eval/src/import.rs
+++ b/crates/typst-eval/src/import.rs
@@ -34,6 +34,7 @@ impl Eval for ast::ModuleImport<'_> {
                 source = Value::Module(import(&mut vm.engine, path, source_span)?);
                 is_str = true;
             }
+            Value::Dict(dict) => source = Value::Module(dict.clone().module()),
             v => {
                 bail!(
                     source_span,
--- a/crates/typst-library/src/foundations/dict.rs
+++ b/crates/typst-library/src/foundations/dict.rs
@@ -11,7 +11,7 @@ use typst_utils::ArcExt;
 
 use crate::diag::{Hint, HintedStrResult, StrResult};
 use crate::foundations::{
-    array, cast, func, repr, scope, ty, Array, Module, Repr, Str, Value,
+    array, cast, func, repr, scope, ty, Array, Binding, Module, Repr, Scope, Str, Value,
 };
 
 /// Create a new [`Dict`] from key-value pairs.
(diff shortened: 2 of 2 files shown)


laurmaedje:
As discussed on Discord, a constructor would be more fitting than a method and I don't think we should allow importing from dictionaries directly. It's better to encourage people to construct a module instead as modules behave more, well, module-like (e.g. allowing `mymod.func(..)` rather than `(mymod.func)(..)`).

emilyyyylime:
I'm a bit confused on your example. How can you call `.module()` on a function? Does this example actually compile with your branch?

I do agree with @laurmaedje on the topic of exposing an explicit constructor over a method, even if it isn't implemented the same as other constructors under the hood.

ensko:
I think they confused `pdict` with `pmod` in the import examples

maugier:
Right, sorry, that was an incorrect mashup of two different examples.

This runs, and showcases imports:
```
#let parametric(left, right) = {
    (wrap: (x) => [#left #x #right],
     unit: [#left #right])
}
#let pdict = parametric([\(], [\)])
#let pmod = pdict.module()

#pmod.wrap[a]

// equivalent to let (wrap, unit) = parametric([\(], [\)])
#import pmod: wrap, unit
#wrap[b]
#unit

// Pulls all dict keys into scope without having to repeat the names
#import pmod: *
#wrap[c]
#unit

// Import also calls .module() implicitely, so these also work:
#import pdict: wrap, unit
#wrap[d]
#unit

#import pdict: *
#wrap[e]
#unit
```

This also runs and shows the proper closure behavior:

```
#let param(left, right) = (
   single: x => [#left #x #right],
   double: x => [#left #left #x #right #right]
).module();

#let parens = param([\(], [\)]);
#let braces = param([\{], [\}]);

#parens.single[a] // (a)

#parens.double[b] // ((b))

#braces.single[c] // {c}

#braces.double[d] // {{d}}
```

I tried implementing that as the `module()` constructor, but I couldn't get it to build and work properly yet.

emilyyyylime:
> I tried implementing that as the module() constructor, but I couldn't get it to build and work properly yet.

Any specific issues you encountered? Feel free to ask in #contributors on the Discord server for help 🙂

maugier:
The issue was that I didn't see that the `#[ty(scope, cast)]` macro are order-sensitive.

Closing in favor of #6387.

PR #6372, opened 2025-06-03
Fix Ukrainian secondary smart quotes

G1gg1L3s:
As discussed in #6331, let's also fix secondary smart quotes for Ukrainian language.

According to the [Ukrainian orthography of 2019](https://mova.gov.ua/storage/app/sites/19/2024/_2019.pdf):

>  функції перших рекомендовано вживати кутові
> лапки, або «лапки-ялинки» («…»), у функції внутрішніх — «лапки-ла́пки»
> (“…” та ін.): «Це мій “Кобзар”», — сказав він.

It means that the `«…»` symbols are used for the external and `“…”` for the internal quotes.

The orthography also allows `„ “`, however it its examples it mostly uses the `“…”` variant. I also asked my friend who is a philologist and she said that most publishers use the `“…”` variant, so I think it's a default safe option. The `“…”` variant is also a bit easier to read:

image


laurmaedje:
Thanks!

PR #6377, opened 2025-06-04
Apply short fall consistently in math when stretching

mkorje:


laurmaedje:
I think my original thought was that if we pick a variant and it's just a little too short (<= short fall), then it's not worth picking a way larger one, but if we create a construction, we might as well create the exact desired size.

The interpretation here instead is that it's generally good to make the stretched thing a little less wide/large than the base/content. That's more consistent I guess. Do we happen to know what LuaLaTeX does? Probably the same as this here?

mkorje:
> I think my original thought was that if we pick a variant and it's just a little too short (<= short fall), then it's not worth picking a way larger one, but if we create a construction, we might as well create the exact desired size.

I think I gathered this as well. But I didn't like the idea of either having the short fall hard coded (and so appears as a bit of magic to the end user), or adding another parameter along with the size...

> The interpretation here instead is that it's generally good to make the stretched thing a little less wide/large than the base/content. That's more consistent I guess. Do we happen to know what LuaLaTeX does? Probably the same as this here?

Yeah my reasoning is really just the above. I believe *TeX uses the following formula `calc.max(x - 5pt, x * 0.901)`, and then just picks among the variants or assembly to get the smallest one at least the calculated size. So basically the same as this PR then.

I could see it being nicer to only apply the short fall for prebuilt variants, but exposing this will just be confusing to end users I think. Also, the prebuilt variants and their sizes are entirely font dependent anyways...

laurmaedje:
Alright, I think I'm good with this change. The updated reference images also actually look a bit nicer imo.

mkorje:
I'll sort out the other PR now.

tests/ref/math-accent-wide-base.png
ParaN3xus, line 1: Is this change really correct? Using `\overrightarrow` in LaTeX does not produce this noticeably short arrow

mkorje: Yeah I agree it doesn't look great. Maybe we can just modify the accent size shortfall, I'll try it out soon.

PR #6379, opened 2025-06-04
Disambiguate numbering pattern names

samuelireson:
See https://github.com/typst/typst/pull/5622 for the discussion so far.


emilyyyylime:
If we're attempting to align names with CSS, are there any plans to also test against other implementations of CSS numbering to ensure compatibility?

samuelireson:
To start the discussion of how to proceed here, I will write down some points.

To begin with, how could we rename/name the patterns.

1. Copy the naming as dictated on [this webpage](https://www.w3.org/TR/predefined-counter-styles/). In some sense this is a consistent approach, but [it has already been noted](https://github.com/typst/typst/pull/5622#issuecomment-2765808975) that this naming is inconsistent in it's own way.
2. Come up with our own naming strategy.

I think that actually both of these options are dependent on a decision about which numbering patterns to support. Especially the second option will be achieved in the most consistent manner if there is an understanding about exactly which variations of any given numbering pattern are expected to be included.

Then comes the question of how to organise the selection of these patterns. [It has been mentioned](https://github.com/typst/typst/pull/5622#issuecomment-2575267295) that these patterns could be contained in a separate repo or crate. Is this still potential? To me this seems a bit much, but the list of css ready-made styles is very large, so perhaps this will bring some organisational benefit.

samuelireson:
> other implementations of CSS numbering

do you have any references for this? I haven't stumbled across any apart from w3.

emilyyyylime:
Only thing I could think of is looking into the implementations of browser engines / rendering an html doc directly using them.

If we do decide to aim for CSS compatibility as a goal I think splitting this to a separate crate/project could have value to the Rust community at large, and even to young browser engines.

samuelireson:
> could have value to the Rust community at large

Is this true? To me it seems like this is a fairly niche area, but maybe I just haven't thought about it enough. Suppose this was split to a separate crate/repo, how would we actually go about doing that? Is this something which has happened in other areas of the codebase before?

laurmaedje:
> Is this something which has happened in other areas of the codebase before?

It definitely has. Examples include https://github.com/typst/unicode-math-class and https://github.com/typst/codex, but more generally a lot of the stuff you can see on https://github.com/orgs/typst/repositories.

> Suppose this was split to a separate crate/repo, how would we actually go about doing that?

I think we can work on the PR here first regardless of whether we decide to incorporate all CSS counter styles. If we decide to include all and want a separate repository, I would take care of setting this all up.

samuelireson:
Do you have any thoughts on naming convention @laurmaedje ?

We can use this pr to rename the current patterns and add any others we want, since these won't be accessible before the {} is implemented anyway.

laurmaedje:
I've involved the folks that are working on Typst symbol naming in https://github.com/typst/codex via [Discord](https://discord.com/channels/1054443721975922748/1277628305142452306/1380114491638943796). I'm thinking that numbering systems could potentially also be moved into that crate.

MDLC01:
I opened typst/codex#116

laurmaedje:
As there doesn't seem to be a lot of movement here so far and the PR doesn't yet have changes, I'll close this for now to unclutter the PR tracker. We can always reopen when things start moving. Thanks for your understanding!

PR #6380, opened 2025-06-04
Make a more descriptive definition of `module`

Andrew15-5:
From https://github.com/typst/typst/pull/6146#issuecomment-2781494377.

I'm not sure if docs should mention the intended use of related items, or it can be unrelated stuff and the specifics must be dropped.

Cc @MDLC01


laurmaedje:
Thanks! I think mentioning relatedness is good.

PR #6381, opened 2025-06-04
Warning when watching stdin

### LINKED ISSUE #3996: Watch from stdin not work
### Description

Typst version: 0.11.0.

Issue:  I use the following command to read from stdin but watch does not compile.

```
typst watch - ./test.pdf
```

I see this feature was added in https://github.com/typst/typst/pull/3339, but seems not working for me. Please confirm if is is broken. Thank you!

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

YDX-2147483647:
This PR adds a warning for `typst watch - …`. The stdin is not a regular file and it will never recompile.

Resolves #3996

### Question

This is a warning that can be determined _before_ compilation. Should I implement it in `crates/typst/src/lib.rs`?

Currently I implement it in `crates/typst-cli/src/watch.rs`, but I realize that I am breaking new ground.


laurmaedje:
Implementing this in watch.rs is good. The `typst` crate itself has no notion of watching.

I trimmed the error message a bit further and extracted into a function. I also think a generic config_warnings mechanism is overkill for now. We can see about that if we need it. I removed it for now.

Thanks!

laurmaedje:
> This is a warning that can be determined before compilation. 

I think it's okay to print after compilation. Otherwise, it will be trickier to not interfere with other diagnostics and printing stuff going on. The warning is also only really relevant once nothing happens anymore, i.e. after compilation is finished.

PR #6383, opened 2025-06-04
Replace "inset" with "shifted" in outline.indented docs

Andrew15-5:
<https://typst.app/docs/reference/model/outline/#definitions-entry-definitions-indented>:
> If the outline's indent is a fixed value or a function, the prefixes are indented, but the inner contents are simply inset from the prefix by the specified gap, rather than aligning outline-wide.

I didn't understand what that last part means until I tried

```typ
#set heading(numbering: "1.")
#set outline(indent: 2em)
#outline()
= #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
== #lorem(20)
```

which confirmed my hunch. So this part might also need rewording.

"Inset" is used for containers as "inner padding", while in this case the content is just intended/shifted to the right/left. Unfortunately the "indented" term is probably bad because it's already used elsewhere and is the name of the method. So "shifted" is the only one I can come up with. Maybe "padded"/"spread", I don't know...

### DIFF
--- a/crates/typst-library/src/model/outline.rs
+++ b/crates/typst-library/src/model/outline.rs
@@ -450,8 +450,8 @@ impl OutlineEntry {
     /// at the same level are aligned.
     ///
     /// If the outline's indent is a fixed value or a function, the prefixes are
-    /// indented, but the inner contents are simply inset from the prefix by the
-    /// specified `gap`, rather than aligning outline-wide.
+    /// indented, but the inner contents are simply shifted from the prefix by
+    /// the specified `gap`, rather than aligning outline-wide.
     #[func(contextual)]
     pub fn indented(
         &self,


laurmaedje:
Maybe "offset"?

Andrew15-5:
> I didn't understand what that last part means until I tried
> 
> ```typst
> #set heading(numbering: "1.")
> #set outline(indent: 2em)
> #outline()
> = #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> == #lorem(20)
> ```
> 
> which confirmed my hunch. So this part might also need rewording.

Can we add a similar example? Visualization always helps.

laurmaedje:
Sure. Ideally one that is minimal but still demonstrates the point.

Andrew15-5:
I think something like this.

```typ
// #show heading.where(level: 1): set heading(numbering: n => n * 5)
#set heading(numbering: (..n) => n.pos().first() * 5)
// #set heading(numbering: n => n * 5)
#set outline(indent: 0pt)
#outline()
= First
= Second
```

image

```typ
#show heading.where(level: 2): set heading(numbering: (_, n) => n * 5)
#set outline(indent: 2em)
#outline()
= Top heading
== First
== Second
```

image

But something like direct comparison is better.

```typ
// #show heading.where(level: 1): set heading(numbering: n => n * 5)
#set heading(numbering: (..n) => n.pos().first() * 5)
// #set heading(numbering: n => n * 5)
#outline()
#outline(indent: 0pt)
= First
= Second
```

image

```typ
#show heading.where(level: 2): set heading(numbering: (_, n) => n * 5)
#outline()
#outline(indent: 2em)
= Top heading
== First
== Second
```

image

laurmaedje:
Maybe you can use roman numbering instead of illustrates different widths. The `n * 5` thing could be confusing. And I think we can hide the actual headings using a hidden "show heading: none" to focus only on the outline. A comparison makes sense I think.

Andrew15-5:
```typ
#show heading: none
#set heading(numbering: "I-I.")
#outline()
// #linebreak()
// #line(length: 100%)
#line(length: 100%, start: (0pt, -2pt))
#outline(indent: 3em)
= Top heading one
== First
== Second
== Third
= Top heading two
= Top heading three
```

image

laurmaedje:
This could work. I would adjust the source code with ">>>" and "<<<" to make it display like this:
```typ
#set heading(numbering: "I-I.")

#outline()
#line(length: 100%)
#outline(indent: 3em)

= Top heading one
== First
== Second
== Third
= Top heading two
= Top heading three
```
And the headings would imo benefit from more realistic titles.

Andrew15-5:
I've seen those in https://github.com/typst/typst/pull/6417, but there, it was used to hide previously shown stuff. Here, I don't like that it makes it not reproducible. I would've been confused, not getting the same result as the example shows. So, here, I guess hiding headings is fine, though uncentered line is kinda meh. But this also hides the outline title, which you can't just crop out. So I added an explicit set rule for it, to keep it real (more reproducible).

Also, this is a bit strange that this whole description is written not in `outline.indent`. So it feels like docs fracturing. But `outline.indent` already has its own example, that doesn't display this nuance. Maybe we can ditch that and replace it with this one?

laurmaedje:
> Also, this is a bit strange that this whole description is written not in outline.indent. So it feels like docs fracturing. But outline.indent already has its own example, that doesn't display this nuance. Maybe we can ditch that and replace it with this one?

That would be best.

laurmaedje:
Thanks!

PR #6385, opened 2025-06-04
Update docs for gradient.repeat

Andrew15-5:


crates/typst-library/src/visualize/gradient.rs
```diff
@@ -549,7 +549,7 @@ impl Gradient {
     }
 
     /// Repeats this gradient a given number of times, optionally mirroring it
-    /// at each repetition.
+    /// at every second repetition.
```
MDLC01, line 552:
```suggestion
    /// every second repetition.
```

This may be more natural, but I'm not a native speaker so I might be wrong.

If this is changed here, it should be changed below as well.

Andrew15-5: Yeah, I always go by feel, here I don't think there is a big difference. I probably used it because it ("at") was already there.

laurmaedje: To me, the version with "at" sounds more natural.

laurmaedje (approved):

crates/typst-library/src/visualize/gradient.rs
Andrew15-5: I'd say this is just personal preference.

laurmaedje:
Thanks!

PR #6387, opened 2025-06-05
Add `module(dict)` constructor

maugier:
Give `module` a constructor that takes a single dictionary as an argument.

Closes https://github.com/typst/typst/issues/6419

### DIFF
--- a/crates/typst-library/src/foundations/module.rs
+++ b/crates/typst-library/src/foundations/module.rs
@@ -2,7 +2,7 @@ use std::fmt::{self, Debug, Formatter};
 use std::sync::Arc;
 
 use ecow::{eco_format, EcoString};
-use typst_syntax::FileId;
+use typst_syntax::{FileId, Spanned};
 
 use crate::diag::{bail, DeprecationSink, StrResult};
 use crate::foundations::{func, repr, scope, ty, Binding, Content, Dict, Scope, Value};
@@ -63,10 +63,10 @@ impl Module {
     /// m.hello[World]
     /// ```
     #[func(constructor)]
-    pub fn construct(dict: Dict) -> Module {
+    pub fn construct(dict: Spanned<Dict>) -> Module {
         let mut scope = Scope::new();
-        for (name, value) in dict {
-            scope.bind(name.into(), Binding::detached(value));
+        for (name, value) in dict.v {
+            scope.bind(name.into(), Binding::new(value, dict.span));
         }
         Module::anonymous(scope)
     }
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/foundations/module.rs
```diff
@@ -54,6 +54,24 @@ struct Repr {
     file_id: Option<FileId>,
 }
 
+#[scope]
+impl Module {
+    /// Build a module from a dictionary.
+    ///
+    /// ```example
+    /// #let m = module((hello: (who) => [Hello, #who !]))
+    /// m.hello[World]
```
laurmaedje, line 63: I think an example with multiple fields and maybe an import statement would be nice. Here would be an example suggestion:
```typ
// Create a module dynamically.
//
// Note the second pair of parentheses
// to create a dictionary.
#let m = module((
  greet: (who) => [Hello, #who!],
  // .. can add more members here
))

// Use it.
#m.greet[you]

// Or import from it!
#import m: greet
#greet[friend]
```

laurmaedje (commented):
Could you please also add two tests, one with field access and one with an import statement?

crates/typst-library/src/foundations/module.rs
```diff
@@ -54,6 +54,24 @@ struct Repr {
...
+    /// ```
+    #[func(constructor)]
+    pub fn construct(dict: Dict) -> Module {
+        let mut scope = Scope::new();
+        for (name, value) in dict {
+            scope.bind(name.into(), Binding::detached(value));
+        }
+        Module::anonymous(scope)
+    }
```
laurmaedje:
```suggestion
    pub fn construct(dict: Spanned<Dict>) -> Module {
        let mut scope = Scope::new();
        for (name, value) in dict.v {
            scope.bind(name.into(), Binding::new(value, dict.span));
        }
        Module::anonymous(scope)
    }
```
this is better for go-to-definition

crates/typst-library/src/foundations/module.rs
ensko: I don't know if it runs significantly different code paths, but I think a test for `import function-returning-module() as m` would also be nice. That's how I envision using this, at least.

Edit: adding code to make this more actionable:
```typ
// Returns a dynamically created  module dynamically.
//
// Note the second pair of parentheses
// to create a dictionary.
#let greeter(salutation) = module((
  greet: (who) => [#salutation, #who!],
  // .. can add more members here
))

// create and import module, and items from module
#import greeter("Hello") as m: greet

#m.greet[you]

#greet[friend]
```

laurmaedje:
There is a lot of discussion about the desirability of this in https://github.com/typst/typst/issues/6419. I am myself not yet certain, but in any case need more time to think about it. Adding the feature itself is simple, so we can always reopen / do it later. For now, I'll close this. Thank you still!

PR #6389, opened 2025-06-05
Don't wrap html.elem at document root in <p>

### LINKED ISSUE #5907: HTML export adds unexpected <p> element
### Description

This is new behavior introduced somewhat recently (using d199546f9fe92b2d380dc337298fdca3e6fca8c8 right now). Typst adds an unexpected \<p> element around a \<span> like this when they are constructed with html.elem. I would expect this to happen with markup mode, but not with the raw element constructors inside a code context.

Input:

```typst
#{
    html.elem("p")[foo]
    html.elem("span")[bar]
}
```

Output:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <p>foo</p>
    <p><span>bar</span></p>
  </body>
</html>
```

Expected:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <p>foo</p>
    <span>bar</span>
  </body>
</html>
```

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

breezykermo:
Fixes #5907.

The issue here seems to be related to #5746, in the sense that not _all_ content should be wrapped in a `<p>` tag when rendered to HTML.

Please note that this is my first contribution that modifies code, and so I am not sure whether this conditional is in the ideal place in compilation. It seems to me that this nuance is somewhat specific to HTML, in the sense that there should be an exception to the 'wrap all root content in paragraph elements' rule in the case of user-specified HTML. But please tell me if this is not the correct level of abstraction at which to make the change, and I will be happy to update this PR accordingly.

I have also added a test according to the apparent pattern of the test suite, but again would appreciate any pointers on whether this is appropriate and idiomatic.

Thanks!

### DIFF
--- a/crates/typst-html/src/lib.rs
+++ b/crates/typst-html/src/lib.rs
@@ -193,12 +193,18 @@ fn handle(
     } else if let Some(elem) = child.to_packed::<ParElem>() {
         let children =
             html_fragment(engine, &elem.body, locator.next(&elem.span()), styles)?;
-        output.push(
-            HtmlElement::new(tag::p)
-                .with_children(children)
-                .spanned(elem.span())
-                .into(),
-        );
+        // If the body is already an HtmlElem, then the user has directly specified it via
+        // `html.elem`, and it shouldn't be wrapped in a <p> tag.
+        if elem.body.is::<HtmlElem>() {
+            output.extend(children);
+        } else {
+            output.push(
+                HtmlElement::new(tag::p)
+                    .with_children(children)
+                    .spanned(elem.span())
+                    .into(),
+            );
+        };
     } else if let Some(elem) = child.to_packed::<BoxElem>() {
         // TODO: This is rather incomplete.
         if let Some(body) = elem.body(styles) {
@@ -257,6 +263,7 @@ fn handle(
             child.elem().name()
         ));
     }
+
     Ok(())
 }
 
--- a/tests/ref/html/issue-5907-html-elem-at-root.html
+++ b/tests/ref/html/issue-5907-html-elem-at-root.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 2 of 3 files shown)


breezykermo:
The failing test seems to indicate that this might be the wrong level of abstraction for the change, as I would expect that:

```typ
#link("https://typst.org/")[Some text]
``` 

should compile to:

```html
<p><a href="https://typst.org/">Some text</a></p>
```

In contrast, I would expect that:

```typ
#html.elem("a", (attrs: (href: "https://typst.org")), [Some text])
``` 

should compile to an element without the wrapping `<p>` tag, i.e.:

```html
<a href="https://typst.org/">Some text</a>
```

I don't think that a distinction between these two can be made in the `typst-html` package. Is this correct? If so, any pointers on where the correct change should be made would be welcome! I would be happy to close this and open another more appropriate PR.

laurmaedje:
You're right that this is not the correct point for change. The position where the element is wrapped in a paragraph in the first place [is here](https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-realize/src/lib.rs#L887-L890). However, it's in principle intentional that inline-level elements are wrapped in paragraphs. I see how, in some cases, it's annoying, but removing this behaviour outright seems just as problematic.

breezykermo:
Would it be reasonable to remove it only when the document explicitly specifies an `html.elem` at the top level? It seems to me that it would never be desirable to wrap such an element in a `<p>` tag.

If so, I would be happy to look into making this change during realization (where the `ParElem` is introduced in the first place).

Myriad-Dreamin:
I'm not sure you know it or I learnt some internal features. You can forbid generation of top `<p>` by wrapping a div manually:

```
#html.elem("div")[test]
```

Will generate

```html
// other parts of HTML are omitted.
<body>
  <div>test</div>
</body>
```

breezykermo:
Closing this, as I understand that a different fix will be needed.

Issue #6404, opened 2025-06-07
`typst query` doesn't work with HTML export


MDLC01:
### Description

`typst query` always generates document for paged targets. This means it is not possible to use the command to query elements that are only present in HTML export.

For example, consider the document below, meant to be exported to HTML only:

```typ
#html.elem("body")[#metadata("bug") <query-me>]
```

The following command returns an empty list and warns that the `html.elem` was ignored during paged export.

### Typst version

- [x] I am using the latest version of Typst

PR #6405, opened 2025-06-07
Add `--target` argument for `typst query`

MDLC01:
This small PR fixes https://github.com/typst/typst/issues/6404 by adding a new `--target` argument for `typst query`. `--target` can be either `paged` or `html`. A `PagedDocument` or `HtmlDocument` is generated accordingly, and introspected for the desired query.

### DIFF
--- a/crates/typst-cli/src/query.rs
+++ b/crates/typst-cli/src/query.rs
@@ -4,9 +4,10 @@ use serde::Serialize;
 use typst::diag::{bail, HintedStrResult, StrResult, Warned};
 use typst::engine::Sink;
 use typst::foundations::{Content, IntoValue, LocatableSelector, Scope};
+use typst::introspection::Introspector;
 use typst::layout::PagedDocument;
 use typst::syntax::{Span, SyntaxMode};
-use typst::{Document, World};
+use typst::World;
 use typst_eval::eval_string;
 use typst_html::HtmlDocument;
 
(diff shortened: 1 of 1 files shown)


eltos:
Why not the established `--format` for consistency with `compile`? Thinking in the direction of `pdf.embed`, there could potentially be other format specific elements in the future which might require distinguishing different paged formats.

MDLC01:
For `typst query`, `--format` is already used to choose the format to use for the result of the query (JSON or YAML). Additionally, the proposed `--target` lets the user chose between paged targets and HTML, whereas `--format` in `typst compile` and `typst watch` is for the specific output format (e.g., PNG, PDF, or HTML).

`--target` mirrors the [`target`](https://typst.app/docs/reference/foundations/target/) function.

Andrew15-5:
If this is accessible without `--features html`, then I think it's bad, since HTML export is experimental.

MDLC01:
It does not work without the HTML feature enabled, but that could also be through the corresponding environment variable.

MDLC01:
I just noticed that as is, `--target` is required. Should I make it optional with a default to `paged`?

Andrew15-5:
Yes.

MDLC01:
I made `--target` optional, with a default value of `paged`.

MDLC01:
Should I document `--target` at https://typst.app/docs/reference/introspection/query/#command-line-queries? Or is that too much detail?

Andrew15-5:
I actually don't see a single mention of `--target` in the repo, so I assume you shouldn't. We still don't have a section about CLI, so it's probably excessive and out of place now.

laurmaedje:
> I actually don't see a single mention of --target in the repo, so I assume you shouldn't. We still don't have a section about CLI, so it's probably excessive and out of place now.

I don't really get that argument. `--target` was just introduced in this PR, so why would it appear in the repo? Just because we don't have other CLI docs, doesn't mean it hurts to expand the docs for the feature that _does_ have docs.

MDLC01:
I added some documentation

Andrew15-5:
> `--target` was just introduced in this PR

Oh, I was thinking about specifying a target, but there are `--format` and `--features`. Those are not documented, but they all are about the same thing, basically. Yeah, I was surprised to not find it, but it makes sense now.

laurmaedje:
Took me a while to get to terms with the format/target thing, but I discussed it with the team and we decided that it's fine. Thank you!

PR #6407, opened 2025-06-07
Add docs for `std` module

### LINKED ISSUE #6395: Add entry for `std` variable in documentation
### Description

I've found

https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-library/src/lib.rs?plain=1#L151-L152

And that

https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-library/src/foundations/scope.rs?plain=1#L53

But there is no documentation about it other than

https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/docs/changelog/0.12.0.md?plain=1#L241

Where can this be added?

Andrew15-5:
Closes #6395.

I did some sketches to get some idea of what should be included.


docs/reference/groups.yml
```diff
@@ -130,6 +130,42 @@
...
+      parbreak()
+    }
+
+    #lorem(10)
+    #special[
+      #par
+    ]
+    #lorem(10)
+    ```
+
```
bluss, line 185: Just a potential suggestion, I don't know if it needs mentioning but `std` is an 'escape hatch' in math mode because for example `grid` is not in scope in math, but `std.grid` is, and so on.

Andrew15-5:
I have no idea how to make this a module. This works like a module, but it's a `Binding`, so is it really just a variable? Does this mean it can't go in here or there really should be some exception for this?

Should actual representation be the same as in Typst? `type(std)` does say that it's a module.

docs/reference/groups.yml
Andrew15-5: Are there some more useful things that are not included?

I can only think of something like `$2^10 = std.calc.pow(#2, #10)$`, but it's worse than `$2^10 = #calc.pow(2, 10)$`.

Andrew15-5:
Not sure what `Value::Func(_)` filtering is for, but if I avoid it altogether, then it's fine.

https://github.com/typst/typst/blob/bd66907040189215b9b98598826dd0d7e8fbcaa4/docs/src/lib.rs?plain=1#L36-L51

docs/reference/groups.yml
bluss: Right, `calc.pow` I guess does not benefit as much, while functions can benefit if you can use that they continue to have math mode in the arguments.

Andrew15-5: Which are all in `math` module, which is directly available in math mode?

bluss: hm, maybe you're right, I admit that this is less useful than it sounds, like for example

```
$
std.grid(columns: #2, inset: #0.5em,
  pi = 1, pi = 2,
  pi = 3, pi = 4,
)
$
```

That looks neat but the table cells are not rendered like proper inline equations would be (not italic).

docs/reference/groups.yml
```diff
@@ -130,6 +130,42 @@
...
     the constants `pi`, `tau`, `e`, and `inf`.
 
+- name: std
+  title: Standard library
+  category: foundations
+  path: ["std"]
+  details: |
+    # Collection of globally accessible items under `std` variable.
+    # Collection of globally accessible items under `std.` prefix.
+    # Module consisting of globally accessible items.
```
laurmaedje: How about the variation "A module that contains all globally accessible items."

docs/reference/groups.yml
```diff
@@ -130,6 +130,42 @@
...
+    # The standard library as a value.
+    An alias for globally accessible items comprsing the standard library.
+
+    This is useful whenever you override a name from the global scope, e.g.,
+    the [`text`] element. To access the element simply write `std.text`.
+    Since `std` is not a keyword, you can override it if necessary.
```
laurmaedje: When would that be necessary? I don't think this should be encouraged.

docs/reference/groups.yml
```diff
@@ -130,6 +130,42 @@
...
+    # Note that element items from `std` variable work exactly the same way,
+    Note that element items from `std` module work exactly like normal ones,
+    meaning they can be used as constructors, in set rules, etc.
```
laurmaedje: I'm not sure this needs explicit calling out, but if it does, it should be "functions" or "element functions" and not "element items".

docs/reference/groups.yml
Andrew15-5: In a very niche use case? I was acknowledging that it's possible.

docs/reference/groups.yml
Andrew15-5: I feel like some people would not realize this right away that you can use it in rules. Both, because you never see random prefixes (weeell, table/math doesn't feel random), and you can't use random names/values in rules, though `let a = text; set a()` does work.

IDK, I just felt like I might make this explicit. Though, this can be dropped, as the example shows this.

I wanted to also include inline `std.text(red)[]`, but it doesn't fit the example. I don't know how I should change it. It should only show how to use it when necessary, and not just because you can. So putting `std.text` in `par = []` is not good.

docs/reference/groups.yml
laurmaedje: I just don't like the "if necessary" unless we have an idea where it could actually be necessary

Andrew15-5: Okay, what if I remove "if necessary"? I added this, because without it, it's not that restrictive. But with that, a user will hardly find any use case where it's necessary, so less incentive to actually override it. The only use case is obviously when you have a variable named `std`. This can be if you have your own standard library inside a package. Or, when you compute a standard deviation, which is actually far more likely and also practical.

laurmaedje: Removing it would be better.

docs/reference/groups.yml
laurmaedje: > Though, this can be dropped, as the example shows this.

Agreed.

docs/reference/groups.yml
```diff
@@ -130,6 +130,37 @@
...
     the constants `pi`, `tau`, `e`, and `inf`.
 
+- name: std
+  title: Standard library
+  category: foundations
+  path: ["std"]
+  details: |
+    A module that contains all globally accessible items.
+
+    This is useful whenever you override a name from the global scope, e.g.,
```
laurmaedje:
```suggestion
    This is useful whenever you overrode a name from the global scope, e.g.,
```
I think it becomes useful only once the overriding is already in the past.

docs/reference/groups.yml
```diff
@@ -130,6 +130,37 @@
...
+    the [`text`] element. To access the element simply write `std.text`.
```
laurmaedje:
```suggestion
    the [`text`] element. To still access the `text` element, write `std.text`.
```

docs/reference/groups.yml
bluss: `std.move` is an example of a function that has use inside equations, and `std.rotate` too.

Andrew15-5: What would be a practical example? I've seen people moving stuff with `h()`.

bluss:
```typ
#let converges(over, from: $X_n$, to: $X$) = {
    $from stretch(->, size: #2em)^over to$
}

#show math.equation.where(block: true): set align(start)
Given r.v $X$ and a sequence of random variables $X_n$
$
converges("a.s.") & \
          & thick std.rotate(#32deg, ==>) thick \
              && thick converges(p) quad ==> quad converges(d) \
          & thick std.rotate(#{-32deg}, ==>) \
converges(L^r) &
$
```

bild

bluss: `std.move` allows you to move things vertically.

There's still the question whether this is proper for Typst - `std.rotate()` allows you to have math parsing in rotate's content body, but that body is not a proper math.equation?

laurmaedje:
I expanded the docs a bit. Feel free to take a look.

grafik

docs/reference/groups.yml
```diff
@@ -137,8 +137,8 @@
   details: |
     A module that contains all globally accessible items.
 
-    This is useful whenever you overrode a name from the global scope, e.g.,
-    the [`text`] element. To still access the `text` element, write `std.text`.
+    This is useful whenever you overrode a name from the global scope (e.g.
+    the [`text`] element). To still access the `text` element, write `std.text`.
```
Andrew15-5: > We currently don't use a comma after e.g. across the whole docs, so it would be inconsistent. According to the internet, both are fine in American English.

This is definitely a lie, or however you want to call it. I see 14 matches with `rq -F 'e.g.,'`. 37 matches with comma on the left, and 1 with surrounding commas.

This also comes after we agreed that Typst uses American English, and docs is supposed to do the same. I even proposed to change all occurrences, or something like that. Don't remember what was the resolution, but it definitely didn't happen (yet).

Lastly, the "according to" part doesn't cite anything, therefore doesn't hold any power. Literally, [the first 2 links](https://duckduckgo.com/?t=ffab&q=commas%20around%20e.g.%20american%20english&ia=web) [say](https://grammarhow.com/eg-comma/) [that](https://linguaholic.com/linguablog/comma-after-e-g/)

> In American English, a comma must come after “e.g.” to help segment it from the list of examples that come after it. 

P.S. After looking at later commits, this issue is completely gone, so I don't understand what is even the reason for this commit.

docs/reference/groups.yml
```diff
@@ -157,6 +157,29 @@
...
+    #context if target() == "html" [
+      HTML
+    ] else [
+      Paged
+    ]
+    ```
+
+    Make sure to use this functionality responsibly as the behavior of your code
+    can change with new Typst versions!
+
```
Andrew15-5: This example feels stretched/far-fetched, and not practical.

The part about `dictionary` is already covered in https://typst.app/docs/reference/foundations/dictionary/#constructor.

The outlined experimental nature means that this part or whole section must be rewritten once this function is not experimental anymore.

The "if used on an older" version argument is not actually why [the hack was created](https://forum.typst.app/t/how-do-i-access-target-function-without-html-feature-flag/3701/2?u=andrew). It's only because it's experimental and not available on the _latest_ Typst version without a feature flag. Once it's available, there isn't much reason to use it, as you can't export to HTML in older versions anyway.

The example itself doesn't show practical use of the condition, other than diverging the output between the targets (even more), which is [not what you would want](https://github.com/typst/typst/pull/5755).

And the responsibility sentence is not how Typst docs is written. If anything, these should be avoided, I think. Because if you wrote this in one place, you might as well write this for `state`, `context`, `query`, etc. I don't like it.

> the behavior of your code can change with new Typst versions

Well, the point was to provide alternative implementation for older versions, why suddenly new versions are mentioned? Every, or almost every version so far had breaking changes. I don't see the reason to explicitly point out that you can hit a breaking change with this one. Moreover, if you don't have a polyfill, but just a standalone function, if it breaks, then it breaks. The warning is too generic and applicable to basically anything.

---

However, the backward compatibility part (i.e., a polyfill) might be useful with a better example. The only I can think of is new/deprecated functions like path/curve. Which I guess you can't really replicate, because it's very different in API. But things like tiling/pattern [could work](https://typst.app/docs/reference/visualize/tiling/#compatibility).

docs/reference/groups.yml
laurmaedje: > After looking at later commits, this issue is completely gone, so I don't understand what is even the reason for this commit.

Time goes forward :) The commits are in the order I made them in.

docs/reference/groups.yml
laurmaedje: How is it not practical? People are using it in practice. I decided to include this example despite its somewhat hacky nature in an attempt to start making the docs more practically useful rather than theoretically pristine.

> The part about dictionary is already covered in https://typst.app/docs/reference/foundations/dictionary/#constructor.

Concepts can occur multiple times in the docs.

> It's only because it's experimental and not available on the latest Typst version without a feature flag.

I know, hence "new or experimental functionality"

> The example itself doesn't show practical use of the condition, other than diverging the output between the targets (even more), which is https://github.com/typst/typst/pull/5755.

It's just an example.

> And the responsibility sentence is not how Typst docs is written.

That's entirely subjective. In my opinion, it is.

> Every, or almost every version so far had breaking changes. I don't see the reason to explicitly point out that you can hit a breaking change with this one.

The point is that you can be hit by a non-breaking change if you're not careful.

---

There might be a better polyfill example, but you make it sound like its entirely absurd as-is, when it is in my view definitely not.

docs/reference/groups.yml
Andrew15-5: What about the previous statements? Do I need to assume that you implicitly agree?

docs/reference/groups.yml
bluss: Here's a different module lookup example that we've used for real: `#dictionary(color.map).at("turbo")` i.e. looking up color maps by string.

Andrew15-5: > How is it not practical? People are using it in practice.

If you used an already created snippet, [an attribution](https://forum.typst.app/t/how-do-i-access-target-function-without-html-feature-flag/3701/4?u=andrew) (i.e., co-author) in commit would've been nice.

> I decided to include this example despite its somewhat hacky nature in an attempt to start making the docs more practically useful rather than theoretically pristine.

Well that is news to me.

> > The example itself doesn't show practical use of the condition, other than diverging the output between the targets (even more), which is #5755.
>
> It's just an example.

So is it a practically _useful_ example, or _just_ an example?

> The point is that you can be hit by a non-breaking change if you're not careful.

I don't understand this.

> There might be a better polyfill example

I already gave a few examples off top of my head.

> but you make it sound like its entirely absurd as-is, when it is in my view definitely not.

What is "it"? I said that `dictionary(std)` can have a better example, if still included.

The `target` hack is very niche and temporary. Putting it in the docs is indeed undesirable. I don't like the way it's presented and used.

---

Since it's a new move for practicality,

```typ
#context if target() == "html" [
  HTML
] else [
  Paged
]
```

should be revised, or entirely replaced with different polyfill. Not only does it make look like you can/should write

```typ
#context if target() == "html" [
  Here is a content for HTML
] else [
  Here is a completely different output for other targets.
]
```

which is not what you should write, but it also uses the notorious `[]`, that more and more users use without knowing ([yet](https://github.com/typst/typst/pull/6417)) about the implicit surrounding spaces. IMO, it should only be used in a strictly practical ways, to give away the correct use case where it's appropriate to use it over `{}`. If conditions with single words is not one of them.

Andrew15-5: > Here's a different module lookup example that we've used for real: `#dictionary(color.map).at("turbo")` i.e. looking up color maps by string.

This module lookup is irrelevant for `std` (unless modified somehow) and can go into `dictionary` docs instead.

laurmaedje: Unfortunately, I don't have the time to respond to all the individual questions. It's too much bikeshedding for 5 lines of docs. Feel free to post a polyfill example that is better in your opinion and if it's good, we can include it.

Andrew15-5: > Unfortunately, I don't have the time to respond to all the individual questions.

I can wait.

Andrew15-5: > Feel free to post a polyfill example

OK, so the one I was talking about would unify the API to `tiling` by preventing the deprecation warning:

```typ
#let tiling = dictionary(std).at("tiling", default: none)
#if tiling == none { tiling = pattern }
```

Since you don't ever need a show rule for this, you can make a wrapper:

```typ
#let tiling(..args) = {
  let tiling = dictionary(std).at("tiling", default: none)
  if tiling == none { tiling = pattern }
  tiling(..args)
}
```

But if a set rule is required, then you'd go with the first one.

Either way, any polyfill will go only one version back for now, since `std` itself is new. So there isn't anything that is super helpful, just a normal helpful, I guess.

Andrew15-5: For the `target` polyfill, I was thinking of an obvious use of `html` stuff. I immediately thought of what can be `html.frame` used for. The docs doesn't give any concrete examples.

However, for some reason I tried math, and apparently it's completely gone in HTML. Then I remembered that one post about HTML and boilerplate or something. [Here it is](https://forum.typst.app/t/how-to-render-math-equations-inline-and-in-cetz-in-html-export/4139/5?u=andrew). The oneliner show rule looks slick too. So I wrote this:

```typ
#let target = dictionary(std).at("target", default: () => "paged")

#show math.equation: it => context {
  if target() != "html" { return it }
  show: if it.block { x => x } else { box }
  html.frame(it)
}

A simple equation: $x = y$.
```

Dunno why, but text size (or color) doesn't change, only math does. So adding `set text(16pt)` does make it at least scaled correctly, but I can't change text's sans-serif to serif. So it still looks a bit off.

image

image

laurmaedje: The tiling one looks good to me.

Andrew15-5: Are we going with horizontal split?

```typ
#let tiling = dictionary(std).at("tiling", default: none)
#if tiling == none { tiling = pattern }

#let inverted-pattern = tiling(size: (35pt, 20pt), polygon(
  fill: white,
  stroke: 1pt,
  (100%, 0%), (0%, 0%), (0%, 100%), (100%, 100%), // Border
  (0%, 70%), (0%, 30%), (100%, 0%), (100%, 100%), // Pattern
))

#block(
  width: 100%,
  height: 20pt,
  fill: gradient.linear(..color.map.rainbow),
  rect(fill: inverted-pattern, width: 100%, height: 100%, stroke: 1pt),
)
```

https://forum.typst.app/t/is-it-possible-to-apply-gradient-once-over-the-whole-tiling/4664/3?u=andrew

image

laurmaedje: I generally prefer the vertical split (i.e. side by side). I think we should also keep the tiling's complexity to a minimum as to not divert attention from the `std` usage).

Andrew15-5: I still don't have the answer on [how non-breaking change can break things](https://github.com/typst/typst/pull/6407#discussion_r2157614273), so I want to remove the warning with the exclamation mark. Adding a more neutral note can be the case.

Andrew15-5: > I generally prefer the vertical split

I don't like them, because most of the time, including this one, it makes slick one-liners become a huge and loong pile of lines, that no one would ever write, especially people that don't ever use or know about 80 char width etc. This partially can be fixed by just using the empty 32% of horizontal space. That's over 600 pixels.

Andrew15-5: This horizontal:

```typ
#let tiling = dictionary(std).at("tiling", default: none)
#if tiling == none { tiling = pattern }

#let pattern = tiling(
  size: (20pt, 20pt),
  polygon(fill: white, stroke: 1pt, (10%, 50%), (90%, 90%), (90%, 10%))
)

#block(
  width: 100%,
  height: 20pt,
  fill: gradient.linear(..color.map.rainbow),
  // fill: gradient.linear(..color.map.turbo.slice(15, -20).rev()), // This actually is more rainbow-like than the "rainbow", but the red is not bright enough.
  rect(fill: pattern, width: 100%, height: 100%, stroke: 1pt),
)
```

image

image

Nice and compact. And interesting!

Or vertical:

```typ
#let tiling = dictionary(std).at(
  "tiling",
  default: none,
)
#if tiling == none {
  tiling = pattern
}

#let pattern = tiling(
  size: (20pt, 20pt),
  polygon(
    fill: white,
    stroke: 1pt,
    (50%, 10%),
    (90%, 90%),
    (10%, 90%),
  ),
)

#block(
  width: 20pt,
  fill: gradient.linear(
    angle: 90deg,
    ..color.map.rainbow,
  ),
  rect(
    fill: pattern,
    width: 100%,
    height: 100%,
    stroke: 1pt
  ),
)
```

Literally 2+ times more lines. It's awful. To make it much shorter, you almost certainly must make the example boring. No one likes boring examples, unless it's a [beginner tutorial](https://github.com/typst/typst/pull/6417) or something like that. Right?

laurmaedje:
Thanks!

docs/reference/groups.yml
```diff
@@ -163,24 +163,26 @@
...
+    function when available and falling back to `pattern` otherwise will unify
+    the usage across all versions. Note that, when creating a polyfill,
+    [`sys.version`]($category/foundations/sys) can also be very useful.
+
+    ```typ
+    #let tiling = if "tiling" in dictionary(std) {
+      tiling
+    } else {
+      pattern
+    }
```
Andrew15-5, line 181: Why this over the 2 lines? There is no example, so width doesn't even matter (not that it's much narrower now).

laurmaedje: I found it more readable that way. It's also more mobile friendly.

Issue #6409, opened 2025-06-08
Incorrect delimiter matching in math mode when only one is escaped


Andrew15-5:
### Description

I only tested this with 2 pairs, but it only works incorrectly when modifying one of the inner delimiters.

```typ
$ { (x / y) } $ // baseline

$ { \(x / y) } $ // expected: (), got: (}

$ { (x / y\) } $ // expected: (), got: {)

$ { \(x / y\) } $ // expected: (), got: ()

$ \{ (x / y) } $ // expected: {}, got: {}

$ { (x / y) \} $ // expected: {}, got: {}

$ \{ (x / y) \} $ // expected: {}, got: {}
```

![Image](https://github.com/user-attachments/assets/baff9996-f656-4848-8b2a-62072d56f29e)

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

mkorje:
This is intended behaviour; an opening delimiter matches with the first closing delimiter, even if they aren't the same kind.

Andrew15-5:
The [docs](https://typst.app/docs/reference/math/lr#functions-lr) doesn't have this though.

> The `lr` function allows you to match two delimiters and scale them with the content they contain. While this also happens **automatically for delimiters that match syntactically**

> While **matched delimiters scale by default**, this can be used to scale unmatched delimiters

IMO, it says that unmatched / syntactically unmatched delimiters are not auto-scaled, even though I see that they are. Either it's wrong, or these terms must be explicitly defined to remove any doubt. (For me) `(}` are clearly not matching, I wouldn't want them to match. The only exception is the range like `(]`/`[)`, then they are kinda matching, but only in terms of a range expression.

PR #6410, opened 2025-06-08
Document how to escape lr delimiter auto-scaling

### LINKED ISSUE #6158: Add section in `lr()` about resetting parentheses size with a backslash (escape)
### Description

I think https://typst.app/docs/reference/math/lr or somewhere else it should explain that to use normal `()` etc. size, you should write `\()` or `\(\)`.

Andrew15-5:
Closes #6158.

I would've mentioned escaping only one, but #6409.

### DIFF
--- a/docs/reference/groups.yml
+++ b/docs/reference/groups.yml
@@ -112,11 +112,18 @@
     a few more functions that create delimiter pairings for absolute, ceiled,
     and floored values as well as norms.
 
+    To prevent a delimiter from being matched by Typst, and thus auto-scaled,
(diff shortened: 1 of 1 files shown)


laurmaedje:
I think it would be slightly more accurate to say "To prevent a delimiter from being matched by Typst (and thus auto-scaled), escape it with a backslash."

To prevent just the scaling part, and also in a more automatic fashion, you can also use `#set math.lr(size: 1em)`. Maybe also worth mentioning.

laurmaedje:
Thanks!

PR #6413, opened 2025-06-08
Adding Croatian translations entries

ssemigr:
Add Croatian translation entries


istudyatuni:
You should also add this to `crates/typst-library/src/text/lang.rs`, see example in https://github.com/typst/typst/pull/6348

laurmaedje:
I don't speak Croatian, so I double-checked with AI and got the following suggestions:

- "Odjeljak" or "Sekcija" instead of "Poglavlje" for the English term "Section" (i.e. the `heading` key)
- "Ispis" for the English term "Listing" as it's less specific and not just about computer code

Obviously, I can't judge these, but I would be curious to know what you think about these.

ssemigr:
> I don't speak Croatian, so I double-checked with AI and got the following suggestions:
> 
> * "Odjeljak" or "Sekcija" instead of "Poglavlje" for the English term "Section" (i.e. the `heading` key)
> * "Ispis" for the English term "Listing" as it's less specific and not just about computer code
> 
> Obviously, I can't judge these, but I would be curious to know what you think about these.

Indeed both of your suggestions are literal translations of those terms, and I agree that "Odjeljak" is a good choice. 
But as I had a look at Slovenian and Serbian translations (very closely related languages) and as they both use the same term Poglavlje (translation of Chapter) I went with that term just for the consistency sake. Now looking at translations of other languages I'm familiar with, I changed the translation to "Odjeljak".
The term "Listing" is quite complicated to translate. The term "Ispis" is used for printing stuff, and it is not used in this context as far as I know. Translations in Polish, Slovenian and Serbain use the term "Program" but in Croatian we would use "Programski kôd" for this purpose. So I went with the term "Code" which is (slightly) more general.

laurmaedje:
Probably, the existing translations are not quite consistent w.r.t. headings. And the term "Listing" is just hard to translate. Thanks!

PR #6414, opened 2025-06-08
Verbatim representation of content (#6411)

### LINKED ISSUE #6411: Verbatim content as metadata
### Description

```typst
// example.typ
#metadata((data: [*content* $x^2$])) <test>
```

Run:

```sh
typst query example.typ "<test>" --pretty # --verbatim-content
```

Desired output:

```json
[
  {
    "func": "metadata",
    "value": {
      "data": "*content* $x^2$"
    },
    "label": "<test>"
  }
]
```

Current output:

```json
[
  {
    "func": "metadata",
    "value": {
      "data": {
        "func": "sequence",
        "children": [
          {
            "func": "strong",
            "body": {
              "func": "text",
              "text": "content"
            }
          },
          {
            "func": "space"
          },
          {
            "func": "equation",
            "block": false,
            "body": {
              "func": "attach",
              "base": {
                "func": "symbol",
                "text": "x"
              },
              "t": {
                "func": "text",
                "text": "2"
              }
            }
          }
        ]
      }
    },
    "label": "<test>"
  }
]
```

### Use Case

This would enable metadata fields to be rendered separately with Typst.

In my case, I'm interested in automatically generating flashcards. I'd do this by exposing metadata for definitions, theorems, etc., like "name" and "body", then querying for this and rendering individually with Typst into small SVGs/PNGs, which I'd then automatically sync to Anki via AnkiConnect.

```typ
#let theorem(name, body) = {
  metadata((name: name, body: body)) <flashcard-theorem>
  [*Theorem (#name).* #body]
}

#theorem[name][... $x^2$ ...]
```

### Workarounds

The current workaround would be to wrap the content in quotes, yielding a string, and then to call `eval` to render this in the main document.

```typ
#let theorem(name, body) = {
  metadata((name: name, body: body)) <flashcard-theorem>
  [*Theorem (#eval(name)).* #eval(body)]
}

#theorem("name", "... $x^2$ ...")
```

However, this makes the authoring experience much worse, as it prevents IDE features (syntax highlighting, code suggestions, etc.) from working for the contents of the (in this case) theorem.

#### Manual parsing

From what I can tell, the [typ2anki](https://typst.app/universe/package/typ2anki/) currently works by parsing the source file directly. This approach can be very limiting, however, and I think working with the querying system could be a lot nicer.

### Alternative Approach

Of course, implementing stringification of content (e.g., for the `str` function), i.e., an inverse of the `eval` function, so that `eval(str([content]))` is always equivalent to `[content]`, would also resolve this issue. Related: https://github.com/typst/typst/issues/2196.

### Popularity

This workflow might be niche, but I think it could end up being a very appealing feature; see, for example, the popularity of the [Obsidian_to_Anki](https://github.com/ObsidianToAnki/Obsidian_to_Anki) Obsidian plugin.

nvlang:
This PR contributes a new function `verbatim(content) => str`, which outputs content that it receives verbatim. This differentiates it from `repr`, which, given some content, will output an AST.

For example:

```typ
#verbatim([a _b_ *c* $d$])  // "[a _b_ *c* $d$]"
#verbatim([`e` [f](g)])     // "[`e` [f](g)]"
```

Some more interesting examples:

```typ
#let a = [_A_]
#let b = [*B*]
#let c = [C]
#verbatim(a)                // "[_A_]"
#verbatim([#a])             // "[_A_]"
#verbatim({a})              // "[_A_]"
#verbatim([#a b])           // "[#a b]"
#verbatim([#a #b])          // "[#a #b]"
#verbatim(a.body)           // "[_A_]"
#verbatim(a.body + b.body)  // "a.body + b.body"
#verbatim([#a #{b}])        // "[#a #{b}]"
```

Closes #6411.

---

### Implementation details

For "most" kinds of content (in particular, content which features a "sequence" element; for example, `[a $b$]` or `[a b]`, but not `[ab]` or `[$a$]`), the span of the source code is kept track of internally, and can be retrieved cleanly with `content.span()`.

However, for some kinds of content (e.g., `[a]` or `[$a$]`), `content.span()` will yield the range of the inner "content" of the content, as opposed to the whole thing (in our examples, the spans of `a` or `$a$`). For these cases, a bracket matching algorithm is implemented to try to find the span of the "entire" content.

This, of course, is rather hacky and prone to bugs that could leak sensitive source code. I've tried testing a bunch of edge cases, but it's likely that I missed some. If this were to be merged, I'd probably want to add some fuzzing, too.

An alternative approach would be to always retain the span information. This was what 16d8c5d tried to do. But that breaks some tests that rely on the existing behavior for, e.g., PDF-to-Typst syncing or diagnostic messages. Maybe there's a more elegant way around this. That's where input from a maintainer familiar with the inner workings of Typst would be immensely helpful.

#### Limitations

```typ
#let to-content(text) = { return [#text] }
#verbatim(to-content("abc")) // "[#text]"
```

### DIFF
--- a/crates/typst-library/src/foundations/verbatim.rs
+++ b/crates/typst-library/src/foundations/verbatim.rs
@@ -87,13 +87,36 @@ fn find_surrounding_content_block(
                     let mut end = range.end;
                     bracket_depth = 1;
                     let mut pos = start + 1;
+                    let mut in_string = false;
+                    let mut in_raw = false;
+                    let mut escape_next = false;
 
                     while pos < bytes.len() && bracket_depth > 0 {
                         match bytes[pos] {
-                            b'[' => bracket_depth += 1,
-                            b']' => bracket_depth -= 1,
-                            _ => {}
+                            b'\\' if in_string => {
+                                escape_next = !escape_next;
+                            }
+                            b'"' if !escape_next && !in_raw => {
+                                in_string = !in_string;
+                            }
+                            b'`' if !escape_next && !in_string => {
+                                in_raw = !in_raw;
+                            }
+                            b'[' if !in_string && !in_raw => {
+                                bracket_depth += 1;
+                            }
+                            b']' if !in_string && !in_raw => {
+                                bracket_depth -= 1;
+                            }
+                            _ => {
+                                escape_next = false;
+                            }
                         }
+
+                        if bytes[pos] != b'\\' {
+                            escape_next = false;
+                        }
+
                         pos += 1;
                         if bracket_depth == 0 {
                             end = pos;
@@ -109,11 +132,17 @@ fn find_surrounding_content_block(
                     }
                     return None;
                 }
-                b']' => bracket_depth += 1,
-                b'[' => bracket_depth -= 1,
+                b']' => {
+                    bracket_depth += 1;
+                }
+                b'[' => {
+                    bracket_depth -= 1;
+                }
                 // Stop searching if we hit certain delimiters that suggest
                 // we've gone too far
-                b'\n' | b';' | b'{' | b'}' if bracket_depth == 0 => return None,
+                b'\n' | b';' if bracket_depth == 0 => {
+                    return None;
+                }
                 _ => {}
             }
         }
--- a/tests/suite/foundations/verbatim.typ
+++ b/tests/suite/foundations/verbatim.typ
@@ -41,6 +41,12 @@
   verbatim([Some function calls: #rect(height: 10pt, fill: blue), #align(right)[more *content*!] #some-content]),
   "[Some function calls: #rect(height: 10pt, fill: blue), #align(right)[more *content*!] #some-content]"
 )
+
+--- verbatim-within-function ---
+// Verbatim within function calls
+#let wrapper(content) = { return verbatim(content) }
+#test(wrapper([*Hey*]), "[*Hey*]")
+#test(wrapper([A _sequence_]), "[A _sequence_]")
 #let example(c) = { return (content: c, verbatim: verbatim(c)) }
 #test(example([A _longer_ *sequence*!]).verbatim, "[A _longer_ *sequence*!]")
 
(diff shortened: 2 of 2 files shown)


laurmaedje:
If I understand correctly, this will extract the source text at the content's span. As you've mentioned, this breaks down in many cases, primarily, when the spanned source text doesn't work in isolation, e.g. because it contains variables or scripting. Given this, I'm not sure the function is that generally useful as is. The other option, where the source text for some content would be reconstructed after the fact, would not suffer from this problem, but others. Generally, it's very hard to produce (idiomatic) source code for arbitrary content. I'm also concerned by the fact that this can be used to expose implementation details of packages. Overall, I'm not sure adding this feature is a net positive for Typst.

nvlang:
**TL;DR**: I agree. Please feel free to close this PR if you feel like it.

> As you've mentioned, this breaks down in many cases, primarily, when the spanned source text doesn't work in isolation, e.g. because it contains variables or scripting.

Well, I guess it depends on one's expectations. It reminds me of the expansion of macros in LaTeX. For example, if I have `#let x = [...]` and call `verbatim(x)`, do I expect `"x"` or `"[...]"`? Similarly, if I have `#let y = [#x]` and call `verbatim(y)`, do I expect `"y"`, or `"[#x]"`, or "`[[...]]`", or "`[...]`"? To me, the most sensible answer is to do as follows:

- `verbatim([...])` outputs `[...]`, without any expansion.
- `verbatim(x)`, where `#let x = [...]`, outputs `[...]`; a single expansion.

Now, of course, one interesting case is functions, which I think touches upon our shared worries of leaking source code.

- `verbatim(f(x))`, where `f` is some function that outputs content and `x` is some variable.
- `verbatim(z)`, where `#let z = f(x)` and `f` and `x` are as above.

I'm not sure if this usage could be forbidden, but if it could, then that'd be the simplest "solution".

In any case, the behavior of `verbatim` for variables/functions defined locally vs. imported ones seems (at least at first glance) to be identical.

> I'm also concerned by the fact that this can be used to expose implementation details of packages.

I instinctively agree, but, at the same time, I don't actually know why being able to inspect the source code behind functions and variables that the user already has access to would actually be problematic here. My understanding is that, if I'm using a package in Typst, I can always inspect its source code (or some version thereof) anyway.

> The other option, where the source text for some content would be reconstructed after the fact, would not suffer from this problem, but others. Generally, it's very hard to produce (idiomatic) source code for arbitrary content.

I agree; specifically, the mapping of source code to rendered content is not invertible, since there are many different ways to produce the same content. So, the main value that a function that receives content and outputs (one of the infinitely many possibilities of) source code that, when evaluated, generates the same content, would have is the ability to "store" content in a plaintext-like format. This was relevant for me because I was interested in exposing content with the [`metadata`](https://typst.app/docs/reference/introspection/metadata/) function and rendering it in different Typst files. However, I've actually found a better way to accomplish my use case (using [`state`](https://typst.app/docs/reference/introspection/state/)), so I'm not all that invested in this change anymore personally.

On a somewhat related note (but this is a separate issue), the fact that content exposed via `metadata` is rendered with [`repr`](https://typst.app/docs/reference/foundations/repr/) (or at least that's the impression I got), which according to the docs doesn't have a stable output, seems unfortunate. But I very much sympathize with the challenges involved, and recognize that it's probably not worth prioritizing (I can't think of a use case that would necessitate it anymore either).

laurmaedje:
> I instinctively agree, but, at the same time, I don't actually know why being able to inspect the source code behind functions and variables that the user already has access to would actually be problematic here. My understanding is that, if I'm using a package in Typst, I can always inspect its source code (or some version thereof) anyway.

It's less about hiding the code from _the person_ using the package and more hiding it from _the code_ using the package as the code can suddenly break in unexpected ways when the package author changes their source code. Of course, we can say "the output is not stable" like with `repr`, but the more of these kinds of features we have, the more it does become a real problem in practice.

> This was relevant for me because I was interested in exposing content with the [metadata](https://typst.app/docs/reference/introspection/metadata/) function and rendering it in different Typst files. However, I've actually found a better way to accomplish my use case (using [state](https://typst.app/docs/reference/introspection/state/)), so I'm not all that invested in this change anymore personally.

Good to hear that you found a better solution.

> On a somewhat related note (but this is a separate issue), the fact that content exposed via metadata is rendered with [repr](https://typst.app/docs/reference/foundations/repr/) (or at least that's the impression I got), which according to the docs doesn't have a stable output, seems unfortunate. But I very much sympathize with the challenges involved, and recognize that it's probably not worth prioritizing (I can't think of a use case that would necessitate it anymore either).

I agree that it's not great, but also that it's not worth prioritizing at the moment.

PR #6415, opened 2025-06-08
Autocomplete fixes for math mode

cAttte:
hi! this PR seeks to fix two logic bugs that i found while developing a little Typst+CM6-powered math editor. for reference, this is what the current autocomplete looks like for this piece of math (the text is wrapped in `$`s before being passed to the compiler):

<details>
<summary>current</summary>
</img>
</details>

clearly the returned position is equal to the provided `cursor` ([↗](https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-ide/src/complete.rs#L318-L323)), which results in a contextless autocompletion result. the first fix involves copying the part of `complete_markup` that handles interpolated identifiers ([↗](https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-ide/src/complete.rs#L126-L131)) into `complete_math`. given that hash-related completions are essentially the same in math as in markup, i wonder if this + the other cases (let bindings etc.) should be extracted to a `complete_math_or_markup`—this would also help avoid this gap in the future. anyway, the result looks like this after that:

<details>
<summary>with interpolated identifier handling</summary>
</img>
</details>

better but still wrong, because the scope used for completions ([↗](https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-ide/src/utils.rs#L110-L125)) is the global scope rather than the math scope. so i changed `globals()`'s `in_math` to also verify that the math-child isn't preceded by a hash (which makes it no longer math). not sure how robust this is; are there other cases where math children aren't math? anyway, after this fix the result is correct:

<details>
<summary>with correct math scope</summary>
</img>
</details>

i added a test for the second fix, but the first fix is only kinda sorta implicitly tested for, given that the suite doesn't currently test for correctness of the returned position (e.g. `must_start_at()`). should this be added?

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -1676,8 +1676,8 @@ mod tests {
     /// Test that autocomplete in math uses the correct global scope.
     #[test]
     fn test_autocomplete_math_scope() {
-        test("$ #col $", -3).must_include(["colbreak"]).must_exclude(["colon"]);
-        test("$ col $", -3).must_include(["colon"]).must_exclude(["colbreak"]);
+        test("$#col$", -2).must_include(["colbreak"]).must_exclude(["colon"]);
+        test("$col$", -2).must_include(["colon"]).must_exclude(["colbreak"]);
     }
 
     /// Test that the `before_window` doesn't slice into invalid byte
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -298,9 +298,16 @@ fn complete_math(ctx: &mut CompletionContext) -> bool {
         return false;
     }
 
-    // Start of an interpolated identifier: "#|".
+    // Start of an interpolated identifier: "$#|$".
     if ctx.leaf.kind() == SyntaxKind::Hash {
         ctx.from = ctx.cursor;
+        code_completions(ctx, true);
+        return true;
+    }
+
+    // Behind existing interpolated identifier: "$#pa|$".
+    if ctx.leaf.kind() == SyntaxKind::Ident {
+        ctx.from = ctx.leaf.offset();
         code_completions(ctx, true);
         return true;
     }
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -1673,6 +1673,13 @@ mod tests {
         test("#{() .a}", -2).must_include(["at", "any", "all"]);
     }
 
+    /// Test that autocomplete in math uses the correct global scope.
+    #[test]
+    fn test_autocomplete_math_scope() {
+        test("$ #col $", -3).must_include(["colbreak"]).must_exclude(["colon"]);
+        test("$ col $", -3).must_include(["colon"]).must_exclude(["colbreak"]);
+    }
+
     /// Test that the `before_window` doesn't slice into invalid byte
     /// boundaries.
     #[test]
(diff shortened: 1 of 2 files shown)


laurmaedje:
Thanks, looks good! Feel free to make separate PRs for the two other things: (1) extracting common things for math & markup, (2) tests for `must_start_at`. :)

cAttte:
great! there's a small issue though: checking for a prev Hash is indeed insufficient to determine that we're not in math, so just extracting the logic would be pointless. i realize that the more comprehensive (exhaustive?) logic ("#, or #ident, or #let binding, or #context") is already there in `complete_markup()`, so i think i should move it over to somewhere like `utils::is_interpolated()` and have `utils::globals()` use it (to determine `in_math` and therefore return correct results). then, in `complete_{math,markup}()`, we just check if `is_interpolated()` and finally provide the respective scope completions (maybe bypassing `globals()`; at this point we know we're in code scope). it seems convoluted but i'm not sure how else to implement this in a way that [present](https://github.com/typst/typst/blob/5d34d54df4c529456e75f46a60c6d39e1b18ccdd/crates/typst-ide/src/definition.rs#L58) and future users of `globals()` can benefit from the fixed logic. so lmk!

laurmaedje:
Extracting something like `is_interpolated` sounds sensible, though we must take care to produce the correct `ctx.from`. It differs between `#` and `#ident`. I'm not sure it's easily possible if we just have the bool.

Beyond `is_interpolated`, it would generally be nice to just have an API to get the mode at the current cursor as `Markup | Math | Code`. There are two such enums already (LexMode in typst_syntax, and EvalMode in typst_library). We could unify these into `SyntaxMode` in `typst_syntax` and use it here, too.

cAttte:
great idea, should i start with a PR that just unifies these two enums into SyntaxMode? after that i can work on "get_mode_at()" and its dependents

laurmaedje:
sounds good!

cAttte:
wellll this is awkward, turns out that checking Hash _is_ sufficient to determine that we're interpolated (because anything more complex than a `#atom` will just nest in a new parent that isn't mathy (e.g. `Parenthesized` or `CodeBlock`) so we won't get a "potentially-false positive" in the first place). i had simply misunderstood how the CST worked.

still, i think `get_mode_at()` would be good to have because it would let us reorder the main `complete.rs` pipeline into something like `complete_code()` || `_math()` || `_markup()` (currently reversed); the latter two won't have to check for interpolated stuff _at all!_ it would also avoid the non-exhaustive `matches!` guards. i'm guessing this was your thinking

as for its implementation, i was thinking just a big `match node.parent_kind() {` that maps `LetBinding => SyntaxMode::Code`, `Equation => maybe_math`, `Heading => maybe_markup`, etc. (where `maybe_` signifies a prev-Hash check, as in the original impl). we can skip all of the leafy types (keywords, literals...) because they are never parents. 

an alternative impl is just matching the type itself to a mode (`FuncReturn => Code`, `RawLang => Content`...) and only checking the parent in the case of shared kinds (`LineComment`, `Str`...) but that would be one gigantic `match` block

laurmaedje:
> i'm guessing this was your thinking

tbh, it's been a while since I last worked on this code, so you are probably deeper into it right now than I am. 

but overall I would prefer to make the match smaller rather than larger. Basically just compacting the current distributed logic into one place and then refactoring a little if there are obvious improvements.

Issue #6273, opened 2025-05-12
Non-idiomatic advanced heading styling in documentation


Andrew15-5:
### Description

Under https://typst.app/docs/tutorial/advanced-styling/#columns-and-headings, there is

```typ
#show heading.where(
  level: 1
): it => block(width: 100%)[
  #set align(center)
  #set text(13pt, weight: "regular")
  #smallcaps(it.body)
]

#show heading.where(
  level: 2
): it => text(
  size: 11pt,
  weight: "regular",
  style: "italic",
  it.body + [.],
)
```

Which might be good enough in isolation, but it gives a bad example for new users, as you always should prefer show-set rules over show rules. It's not idiomatic, nor the formatting is good.

This example can lead to inability to further style, or sometimes not being able to style in the first place. See [this](https://forum.typst.app/t/which-show-rule-takes-precedence/3904?u=andrew) and [this](https://github.com/typst/typst/issues/6219).

Is this example so old, or was this not considered? It _is_ so old.

[Context](https://forum.typst.app/t/how-can-i-automate-level-1-headings-always-occur-on-a-new-recto-page/4046?u=andrew)

Andrew15-5:
This happens in all tutorials, and in https://typst.app/docs/reference/styling/#show-rules.

Actually, if I `rg -g '*.md' -F '> ['` and `rg -g '*.md' -F '= ['`, then there is almost nothing else, other than https://typst.app/docs/guides/guide-for-latex-users/#templates, which is a copy of the tutorial example.

Andrew15-5:
I don't think the docs have it, and since it'll probably go into tutorials (that are being cooked), I'm gonna leave it here.
From [the forum](https://forum.typst.app/t/unable-to-set-the-math-font-once-the-block-font-is-set/6810/2?u=andrew), there is a mention that rules for basic/fundamental/primitive layout elements is discouraged. Because it will highly likely mess up everything else (other elements) that depends on those primitives.

Might also be good to call this out on individual element pages, e.g., for `box`, `block`, `grid`, etc.

Issue #6416, opened 2025-06-09
Semantic and non-semantic paragraphs and how `set par()` affects both


Andrew15-5:
This is a weird thing I've uncovered [here](https://discord.com/channels/1054443721975922748/1088371867913572452/1381456345877839942).

https://typst.app/docs/reference/text/text/#parameters-hyphenate:
> Whether to hyphenate text to improve line breaking. When `auto`, text will be hyphenated if and only if justification is enabled.

https://typst.app/docs/reference/model/par/#parameters-justify:
> Hyphenation will be enabled for justified paragraphs if the [text function's hyphenate property](https://typst.app/docs/reference/text/text/#parameters-hyphenate) is set to `auto` and the current language is known.

https://typst.app/docs/reference/model/par/#what-becomes-a-paragraph:
> When you add inline-level content to your document, Typst will automatically wrap it in paragraphs. However, a typical document also contains some text that is not semantically part of a paragraph, for example in a heading or caption.

So, `block[text]` is not a paragraph, i.e., not a `par` element/proper paragraph. This means that a thing like `first-line-indent` does not affect it. Okay.

The `par.justify` supposed to enable `text.hyphenate` only for justified paragraphs, which is basically just for paragraphs, since they are all justified (if not disabled). And now:

```typ
#set page(width: 250pt, height: auto)
#set par(justify: true)
// #set text(hyphenate: false)
#block(lorem(25))
```

![Image](https://github.com/user-attachments/assets/d5ae866c-6ec3-4524-88cc-d9c1c32a9bff)

Not only the "blocked text" text is hyphenated, but it's also justified. Which breaks not only what the docs say, but also my mental model.

Now, if we look elsewhere, things get weird. Non-semantic paragraphs are using things like `par.leading` for spacing between lines, even though they are not proper paragraphs. But how else would you specify that for such text? No sane alternatives.

Next, @PgBiel [noted](https://discord.com/channels/1054443721975922748/1088371867913572452/1381469457632858155) that to use `par.hanging-indent` you must create a proper paragraph, and opting into behavior of `par.first-line-indent` (FLI). What if you just want to use hanging indent without FLI? Or in other words, how come blocked text uses `par.leading`, `par.justify`, but doesn't react to `par.hanging-indent`?

[...]

PR #6417, opened 2025-06-09
Refactor tutorials

Andrew15-5:
Closes https://github.com/typst/typst/issues/6273.

I can try splitting this into multiple commits, if needed. There are plenty of stuff to discuss. The main blocker is https://github.com/typst/typst/issues/6416, i.e., the block + hyphenate thing. I included a few comments.

For the most part, this is ready for review/merge.

### DIFF
--- a/docs/tutorial/1-writing.md
+++ b/docs/tutorial/1-writing.md
@@ -84,8 +84,9 @@ meaning in Typst. We can use `=`, `-`, `+`, and `_` to create headings, lists
 and emphasized text, respectively. However, having a special symbol for
 everything we want to insert into our document would soon become cryptic and
 unwieldy. For this reason, Typst reserves markup symbols only for the most
-common things. Everything else is inserted with _functions._ For our image to
-show up on the page, we use Typst's [`image`] function.
+common things. Everything else is inserted with _functions._ For
+[our image](https://github.com/typst/typst-dev-assets/blob/main/files/images/glacier.jpg)
+to show up on the page, we use Typst's [`image`] function.
 
 ```example
 #image("glacier.jpg")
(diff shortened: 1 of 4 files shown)


Andrew15-5:
In https://typst.app/docs/tutorial/formatting/#show-rules, I would redo that as

```typ
#show "ArtosFlow": name => {
  let logo = box(image(
    "logo.svg",
    height: 0.7em,
  ))
  [#logo #name]
}
```

but `let` keyword is only mentioned in https://typst.app/docs/tutorial/advanced-styling/#title-and-abstract.

I guess it's fine, but what I don't understand is why this doesn't introduce the 2 additional spaces? Is it just because the spaces are collapsed like in `word #[ word ] word`? If so, then this example just happen to not be affected by the `[]` block spaces. This is bad.

It doesn't add the spaces even if the word is surrounded by periods. Does `box` somehow trim the spaces?

```typ
#show "ArtosFlow": name => {
  box(image(
    "logo.svg",
    height: 0.7em,
  ))
  " "
  name
}
```

would be better, but it's not exactly as nice and readable as `[#logo #name]`.

The current example also has implicit space in-between that is not explained either.

Andrew15-5:
In https://typst.app/docs/tutorial/writing-in-typst/#figure, I tried adding `box` to caption content block, and it removed the extra spaces. This is definitely not a documented behavior, that looks very handy, if it's not a bug or something.

But also adding `box` and `func[]` breaks the flow, and it's all introduced only in https://typst.app/docs/tutorial/formatting.

---

Adding box to https://typst.app/docs/tutorial/advanced-styling and including title into PDF metadata does not in fact trim the spaces. So it's not as universal and robust as I thought.

Andrew15-5:
Can I add heading show rule example fix from https://typst.app/docs/reference/styling/#show-rules here? The example in https://typst.app/docs/guides/guide-for-latex-users/#templates copies the tutorial example, so it also uses the outdated syntax, even though it's not visible.

laurmaedje:
It's a bit unfortunate timing for a cleanup as we are planning to rewrite the tutorial completely from scratch soon.

Andrew15-5:
Rewrite in Typst?

I don't really mind, as long as instances such as https://github.com/typst/typst/issues/6273 (basically what this PR fixes) are gone (addressed). The idea is to eliminate bad examples that users can then cite as an inspiration for bad code (practices), i.e., so that everyone can learn the right way from the very beginning. Which should also help a bit with forum (fewer things to fix, fewer things that introduces issues and hard to debug).

laurmaedje:
Rewrite as in completely rewrite the content. Because the current one kinda sucks. We'll make sure to check your suggestions while writing the new one.

Andrew15-5:
Oh, so it's still in Markdown... Well, then it's still cool, just not extra cool.

I guess you can add attribution to some commits if you will get some parts out of my PR.

You can set me as a reviewer, would be interesting to see how it will turn out. As always, I might have some polish changes by the time I read through it.

Andrew15-5:
To update, the tutorials rewrite was postponed by the new releases, but it's back on the menu, and it's being worked on. No clear dates.

[Context](https://discord.com/channels/1054443721975922748/1399326777540608041/1427755877770989780):

> We were pretty busy with the release recently, so the tutorial project stalled a bit. But it's back in focus now.

PR #6421, opened 2025-06-10
Renders #underline style as <u> in HTML

breezykermo:
As per [discussion in Discord](https://discord.com/channels/1054443721975922748/1269208488978939946/1380168455331774474), here is a PR to render `#underline[some text]` as ~`<sub>some text</sub>`~ `<u>some text</u>` when exported to HTML.

None of the more sophisticated styling (such as `offset`) is taken into account yet as I understand that there is not yet inline CSS included in HTML export.

Please let me know if this is the right way to add HTML export capabilities, and if this is sufficient test coverage, as I intend to submit similar PRs for other styling elements.

### DIFF
--- a/crates/typst-library/src/text/deco.rs
+++ b/crates/typst-library/src/text/deco.rs
@@ -86,7 +86,7 @@ impl Show for Packed<UnderlineElem> {
     fn show(&self, _: &mut Engine, styles: StyleChain) -> SourceResult<Content> {
         let body = self.body.clone();
         Ok(if TargetElem::target_in(styles).is_html() {
-            HtmlElem::new(tag::sub)
+            HtmlElem::new(tag::u)
                 .with_body(Some(body))
                 .pack()
                 .spanned(self.span())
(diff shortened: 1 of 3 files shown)


EpicEricEE:
The `<sub>` tag is for subscripts, not for underlines, so this is what `#sub[some text]` should render as.

breezykermo:
> The `<sub>` tag is for subscripts, not for underlines, so this is what `#sub[some text]` should render as.

:man_facepalming: I have corrected this PR to render the `<u>` tag.

laurmaedje:
I'm not sure whether `<u>` is actually a correct translation, given this excerpt from MDN:

> Warning: This element used to be called the "Underline" element in older versions of HTML, and is still sometimes misused in this way. To underline text, you should instead apply a style that includes the CSS [text-decoration](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration) property set to underline.

Didn't know that myself either before reading this.

breezykermo:
Closing this PR given its flawed approach.

PR #6422, opened 2025-06-10
Render `#super` as `<sup>`, `#sub` as `<sub>` in HTML

breezykermo:
This PR adds two capabilities to HTML export:
- renders `#super[some text]` as `<sup>some text</sup>`.
- renders `#sub[some text]` as `<sub>some text</sub>`.


laurmaedje:
I think adding `<sub>` here, too, would be easy enough to warrant inclusion.

breezykermo:
I've added `<sub>` and updated the title and description.

laurmaedje:
Thanks!

PR #6424, opened 2025-06-10
Check that all translation files are added to TRANSLATIONS and ends with newline

### LINKED ISSUE #6384: Add newline to EOF in translation files (but ideally everywhere)
I saw https://github.com/typst/typst/pull/4729, and it touched lines that shouldn't be touched because there is no newline at the EOF. Hence, the diff looks bad.

Image

Is there a specific reason why it's still not added?

Cc @mkorje

istudyatuni:
This should help new contributors

Close #6384


MDLC01:
Maybe you could add a check that there is no trailing whitespace at the end of lines, or trailing lines at the end of files, and that all translation files end with a line break (https://github.com/typst/typst/issues/6384).

crates/typst-library/src/text/lang.rs
```diff
@@ -312,14 +312,72 @@ fn lang_str(lang: Lang, region: Option<Region>) -> EcoString {
...
+                file!(),
+            );
+            checked += 1;
+        }
+        assert_eq!(TRANSLATIONS.len(), checked);
+    }
+    #[test]
+    fn test_all_translation_files_formatted() {
+        for file in translation_files_iter() {
+            let content = std::fs::read_to_string(&file)
```
istudyatuni, line 361: Maybe it's better to use TRANSLATIONS here

istudyatuni:
Ugh https://github.com/actions/checkout/issues/135

laurmaedje:
Maybe just remove the unix-style linebreak check but keep the rest.

crates/typst-library/src/text/lang.rs
istudyatuni: Maybe it's not

laurmaedje:
Thank you! This is nice.

PR #6425, opened 2025-06-11
Warn when using variable fonts

### LINKED ISSUE #5304: Warn when trying to use a variable font
### Description

Until we implement proper support for variable fonts (#185), it would be good to emit a warning as they sometimes behave badly (e.g. https://github.com/typst/typst/issues/5277). 

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

cAttte:
this PR fixes #5304 yay

### DIFF
--- a/crates/typst-library/src/text/mod.rs
+++ b/crates/typst-library/src/text/mod.rs
@@ -1412,12 +1412,24 @@ pub fn is_default_ignorable(c: char) -> bool {
 fn check_font_list(engine: &mut Engine, list: &Spanned<FontList>) {
     let book = engine.world.book();
     for family in &list.v {
-        if !book.contains_family(family.as_str()) {
-            engine.sink.warn(warning!(
+        match book.select_family(family.as_str()).nth(0) {
+            Some(index) => {
+                if book
+                    .info(index)
+                    .is_some_and(|x| x.flags.contains(FontFlags::VARIABLE))
+                {
+                    engine.sink.warn(warning!(
+                        list.span,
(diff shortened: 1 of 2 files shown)


crates/typst-library/src/text/mod.rs
```diff
@@ -1412,12 +1412,24 @@ pub fn is_default_ignorable(c: char) -> bool {
...
+        match book.select_family(family.as_str()).next() {
+            Some(index) => {
+                if book
+                    .info(index)
+                    .is_some_and(|x| x.flags.contains(FontFlags::VARIABLE))
+                {
+                    engine.sink.warn(warning!(
+                        list.span,
+                        "variable font family: {}", family.as_str();
+                        hint: "variable fonts are not currently supported and may render incorrectly"
```
laurmaedje: I think the format `"<word> font family: {}"` works less well for the variable fonts problem than for unknown. I'd suggest the following alternative:

- Warning: `variable fonts are not currently supported and may render incorrectly`
- Hint: `try installing a static version of "{}" instead`

laurmaedje (commented):
Looks good! Just some thoughts on the concrete message.

cAttte:
makes sense! changed it

laurmaedje:
I would keep the quotes around `{}` to separate the family name from the message. in the other message, the separation is given through the colon, but here it might be confusing to read with some family names.

cAttte:
oops totally missed that, done

laurmaedje:
Thanks!

PR #6426, opened 2025-06-11
List both YAML file extensions in bibliography docs

Andrew15-5:
https://github.com/typst/typst/pull/6354#issuecomment-2939994835

Since not https://spec.commonmark.org/0.31.2/#link-reference-definitions nor Typst docs is supported, it's rather weird to have long links inline.

But if it's bad, I can strip it to https://www.iana.org/assignments/media-types/application/yaml#:~:text=File%20extension(s):.

### DIFF
--- a/crates/typst-library/src/model/bibliography.rs
+++ b/crates/typst-library/src/model/bibliography.rs
@@ -48,8 +48,8 @@ use crate::World;
 /// You can create a new bibliography by calling this function with a path
 /// to a bibliography file in either one of two formats:
 ///
-/// - A Hayagriva `.yml` file. Hayagriva is a new bibliography file format
-///   designed for use with Typst. Visit its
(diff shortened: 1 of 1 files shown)


laurmaedje:
To be honest, this feels like too much and, this far up, detracts from the point of the bibliography function. I would drop the addition below and just write 

```
A Hayagriva `.yaml`/`.yml` file
```

Andrew15-5:
No link?

laurmaedje:
I don't think it's necessary. It's not our mission to educate people about YAML extensions and it detracts from the point.

Andrew15-5:
A user can be confused as to why there are two files. Which one should be used? A link would answer that.

laurmaedje:
I think this is far enough removed from Typst itself, that having the link in the docs is out of scope and a potential distraction (a link that is irrelevant for most, but people will still click on).

PR #6432, opened 2025-06-11
Unify `EvalMode` and `LexMode` into `SyntaxMode`

cAttte:
hi, this PR does that ↑, as described in #6415, in preparation for beautiful refactorings. i'm not sure if the location of the new `cast!` invocation is right ([↗](https://github.com/typst/typst/pull/6432/files#diff-4824f02b317a637a17ede1b1bd170b4339f7476477b7f31c5412b697add4ded7R253-R264)).

### DIFF
--- a/crates/typst-library/src/foundations/mod.rs
+++ b/crates/typst-library/src/foundations/mod.rs
@@ -250,18 +250,6 @@ impl assert {
     }
 }
 
-cast! {
-    SyntaxMode,
-    self => match self {
-        SyntaxMode::Markup => "markup".into_value(),
-        SyntaxMode::Math => "math".into_value(),
-        SyntaxMode::Code => "code".into_value(),
-    },
-    "markup" => SyntaxMode::Markup,
-    "math" => SyntaxMode::Math,
-    "code" => SyntaxMode::Code,
-}
-
 /// Evaluates a string as Typst code.
 ///
 /// This function should only be used as a last resort.
--- a/crates/typst-library/src/foundations/mod.rs
+++ b/crates/typst-library/src/foundations/mod.rs
@@ -69,6 +69,7 @@ pub use self::ty::*;
 pub use self::value::*;
 pub use self::version::*;
 pub use typst_macros::{scope, ty};
+use typst_syntax::SyntaxMode;
 
 #[rustfmt::skip]
 #[doc(hidden)]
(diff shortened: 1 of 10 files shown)


crates/typst-library/src/foundations/mod.rs
```diff
@@ -250,6 +250,18 @@ impl assert {
...
+cast! {
+    SyntaxMode,
+    self => match self {
+        SyntaxMode::Markup => "markup".into_value(),
+        SyntaxMode::Math => "math".into_value(),
+        SyntaxMode::Code => "code".into_value(),
+    },
+    "markup" => SyntaxMode::Markup,
+    "math" => SyntaxMode::Math,
+    "code" => SyntaxMode::Code,
```
laurmaedje: Please add `///` doc comments above each match arm here. Previously this was taken from the enum via `derive(Cast)`. It's what shows up in the table on https://typst.app/docs/reference/foundations/eval/#parameters-mode.

As for the location, we could move this to `cast.rs`, right above the `cast!` for `MathClass`.

crates/typst-library/src/foundations/mod.rs
```diff
@@ -250,6 +250,18 @@ impl assert {
     }
 }
 
+cast! {
+    SyntaxMode,
+    self => match self {
+        SyntaxMode::Markup => "markup".into_value(),
+        SyntaxMode::Math => "math".into_value(),
+        SyntaxMode::Code => "code".into_value(),
```
laurmaedje: These into_value calls could be deduplicated similarly as in the cast for MathClass.

crates/typst-syntax/src/parser.rs
```diff
@@ -1580,6 +1580,17 @@ struct Newline {
     parbreak: bool,
 }
 
+/// The syntax mode of a portion of Typst code.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum SyntaxMode {
```
laurmaedje: I think this could just live directly in the crate root, at the bottom of lib.rs.

laurmaedje (commented):
Looks good, just a few minor comments!

crates/typst-library/src/foundations/mod.rs
cAttte: right! i saw the duplicated style on the Axis enum so i wasn't sure

cAttte:
the stuff has been moved 👌

crates/typst-library/src/foundations/mod.rs
laurmaedje: ah okay, doesn't matter too much either way

laurmaedje:
Thanks!

Issue #1982, opened 2023-08-23
Multiple, configurable data dirs and package repositories


jorsn:
### Description

Currently, as I understand, there is one data dir, one cache dir, and one hardcoded package repository.
I propose to

1. make the package repository configurable[^1] and possibly allow multiple alternative ones
2. allow multiple data dirs, in a configurable[^1] way

[^1]: e.g. options at compile time, on commandline, in env variables or in a config file

### Use Case

1. Alternative package repos could be helpful, e.g., for private repos or reliability in case of github problems (e.g. it has happened that I was blocked for some time, probably because too many people connected to github in short time from the same university network).
2. Alternative data Dirs are currently possible by setting XDG_DATA_DIR. However, I might want to have multiple data dirs. For example, there might be multiple package roots within one modular project repository. In TeX, I can set TEXINPUTS or even set the search paths from within TeX code. Another example would be, having packages pinned by another package manger like Nix, e.g. if my project has also non-typst dependencies, I want particular git commits or the package versions I want depend on the typst version that I install via nix. Additionally, I have a project-local package directory for the packages I locally develop in this project. Then, I already need two data dirs. This is currently doable for only two directories, by abusing the cache dir.

sbatial:
I did a little thing: https://gitlab.com/sbatial/typst-flake-env

A flake which patches flake to have an override for the package source.

I plan to hopefully contribute it as well if desired but I didn't know what the best (official) way would be to do such an override (env var, config file, cli flag, etc.) so I just created this flake for now.

PPakalns:
For our use case we have private typst templates that i need to distribute to multiple persons that are working on these documents and creating new ones.  It is hard to always remind them to locally put template directories at the root of the project or how to put them in local cache directory. As these files are used for our specific use case, we want to rapidly create new versions, we do not want to put them in public repository.

It would be very nice if typst would support defining new package namespaces that fetches packages from specified address.

For example:
```rs
#import "@preview/{name}:{version}".
```
Fetches package from: 
https://github.com/typst/typst/blob/df4beb6e15b40d22807e25e41546da698038fed9/crates/typst-cli/src/package.rs#L49C10-L49C45
```rs
    let url = format!(
        "https://packages.typst.org/preview/{}-{}.tar.gz",
        spec.name, spec.version
    );
```
It would be nice if i could do this:
```rs
#package_repository("my_private_repositry", "myproject.org/typst_packages/")
// Or even add authentication data
[...]

PR #6433, opened 2025-06-11
Improving error messages in case of package not found

Vanille-N:
I talked about this a few hours ago on [a half-related issue](https://github.com/typst/typst/issues/1982), and here I am!

Several of Typst's error messages in the case of a file not found or package not found are hard to debug because the error message does not specify the path that was searched.

Combined with dynamic loading of files or the conventions as to where the package directory is, this results in messages such as
```
error: failed to load file (access denied)
  ┌─ pending/main.typ:2:14
  │
2 │ #let _ = read(path)
  │               ^^^^
  │
 ```
or
```
error: package not found (searched for @local/mypkg:0.0.1)
  ┌─ pending/main.typ:1:8
  │
1 │ #import "@local/mypkg:0.0.1"
  │
```
which are unhelpful.

This PR changes the error messages in the following cases:

### `@local` namespace doesn't exist
```diff
- error: package not found (searched for @local/mypkg:0.0.1)
+ error: package @local/mypkg:0.0.1 not found
+ = hint: the namespace @local should be located at /home/user/.local/share/typst/packages/local
```

### package in `@local` doesn't exist
```diff
- error: package not found (searched for @local/mypkg:0.0.1)
+ error: package @local/mypkg:0.0.1 not found
+ = hint: the registry at /home/user/.local/share/typst/packages/local does not have package 'mypkg'
```

### package in `@local` is missing this version
```diff
- error: package not found (searched for @local/mypkg:0.0.1)
+ error: package version @local/mypkg:1.0.0 not found
+ = hint: the package mypkg exists, but not with version 1.0.0
+ = hint: the registry at  /home/user/.local/share/typst/packages/local provides up to version 0.0.1
```

### package in `@local` has no manifest
```diff
- error: failed to load file (access denied)
+ error: failed to load file /home/user/.local/share/typst/packages/local/mypkg/0.0.1/typst.toml (access denied)
```

### package in `@preview` doesn't exist
```diff
- error: package not found (searched for @preview/mypkg:0.0.1)
+ error: package @preview/mypkg:0.0.1 not found
+ = hint: the registry at https://packages.typst.org/preview does not have package 'mypkg'
```

### package in `@preview` is missing this version
```diff
- error: package found, but version 1.0.0 does not exist (latest is 0.0.1)
+ error: package version @preview/mypkg:1.0.0 not found
+ = hint: the package mypkg exists, but not with version 1.0.0
+ = hint: the registry at https://packages.typst.org/preview provides up to version 0.0.1
```

### access denied to file
```diff
- error: failed to load file (access denied)
+ error: failed to load file ../Cargo.toml (access denied)
```
  
## Minor changes to functions and types

- `download_package` should return the `PathBuf` in case of success, it spares us one `exists` check.
- variants of `FileError` now contain some `PathBuf`s to remember which file they tried to access

## Major change to trait

In the previous version the path from `PackageError` (what the function that tries to load a package returns) to `SourceDiagnostic` (errors that can be printed with a span) was through the following detailed chain:
- explicit `impl Display for PackageError`
- explicit `impl Display for FileError` defers to `PackageError: Display`
- explicit `impl From<FileError> for EcoString` invokes `FileError: Display`
- implicit `impl Into<EcoString> for FileError` generated from `EcoString: From<FileError>`
- blanket `impl<T, S> At<T> for Result<T, S> where S: Into<EcoString>`

This path means that `FileError` and thus by extension `PackageError` can only provide a single line's worth of error message.

In this PR the new path is
- explicit `impl Display for PackageError` provides the header
- explicit `impl Display for FileError` defers to the above
- new method `write_diagnostics(self) -> EcoVec<EcoString>` in `impl PackageError` provides the hints
- new method `write_diagnostics(self) -> EcoVec<EcoString>` in `impl FileError` defers to the above
- explicit `impl ErrAt for FileError` (new trait) invokes both `FileError: Display` and `FileError::write_diagnostics` and returns a `SourceDiagnostic`

If you have a better solution in mind I can adapt, or if this is fine then I'll just add a bit of documentation.

### DIFF
--- a/tests/src/run.rs
+++ b/tests/src/run.rs
@@ -11,7 +11,6 @@ use std::sync::OnceLock;
 use typst::diag::{SourceDiagnostic, Warned};
 use typst::layout::PagedDocument;
 
-use regex::Regex;
 use typst::{Document, WorldExt};
 use typst_html::HtmlDocument;
 use typst_syntax::{FileId, Lines, VirtualPath};
@@ -86,8 +85,9 @@ macro_rules! regexes {
 regexes! {
     message_filters:
     // Heuristics to normalize Windows file paths:
-    // - a backslash that is not a `\u{...}` is probably from a path
-    r"\\([^u])" => r"/$1",
+    // - a backslash followed by an alphabetic character or a space is probably
+    //   from a path, except for '\u{...}'
+    r"\\([a-tv-zA-Z ])" => r"/$1",
 }
 
 regexes! {
@@ -484,7 +484,7 @@ impl<'a> Runner<'a> {
 
         let message = &normalize_string(&diag.message, message_filters());
         let range = self.world.range(diag.span);
-        self.validate_note(kind, diag.span.id(), range, &diag.message, stage);
+        self.validate_note(kind, diag.span.id(), range, message, stage);
 
         // Check hints.
         for hint in &diag.hints {


crates/typst-kit/src/package.rs
```diff
@@ -87,13 +87,9 @@ impl PackageStorage {
...
 
-        if let Some(packages_dir) = &self.package_path {
-            let dir = packages_dir.join(&subdir);
-            if dir.exists() {
-                return Ok(dir);
-            }
-        }
-
+        // QUESTION: is it sometimes the case that both package_path
+        // and package_cache_path are Some? Is it a problem if we change
```
laurmaedje: A user may want to override a package from the preview namespace (which would land in the cache dir) with their own local version (in the package dir). The order should remain as is to support this use case.

crates/typst-kit/src/package.rs
```diff
@@ -102,14 +98,39 @@ impl PackageStorage {
...
+                return self.download_package(spec, cache_dir, progress);
             }
         }
 
-        Err(PackageError::NotFound(spec.clone()))
+        // In all paths from now, we either succeed or fail with a NotFound.
+        let not_found = |msg| Err(PackageError::NotFound(spec.clone(), msg));
+
+        let Some(packages_dir) = &self.package_path else {
+            // QUESTION: when is this code path hit?
```
laurmaedje: only in very niche cases where we somehow couldn't get a data dir from the OS

crates/typst-kit/src/package.rs
```diff
@@ -102,14 +98,39 @@ impl PackageStorage {
...
+        if !package_dir.exists() {
+            return not_found(Some(eco_format!("did not find package '{}' while searching at {} for", spec.name, package_dir.display())));
+        }
+        let version_dir = package_dir.join(format!("{}", spec.version));
+        if !version_dir.exists() {
+            return not_found(Some(eco_format!("did not find version {} while searching at {} for", spec.version, version_dir.display())));
+        }
+        // We'll never hit this path as long as subdir is indeed namespace/name/version.
+        // What's Typst's policy on `unreachable`?
+        unreachable!()
```
laurmaedje: If it 100% can't ever happen, unreachable is acceptable, but writing the logic in a way that doesn't require it would be much preferred, if possible

Vanille-N:
Yeah I tried to figure out how to have those code paths be explored by the unit tests but I failed. Apart from 'failed to load file (is a directory)' it seems that none of the corresponding paths were being tested initially.

@laurmaedje any suggestions on how to proceed ? Is it acceptable for this logic to remain untested as it was ?

laurmaedje:
It's okay to skip tests for now. We'll add CLI tests at some point in the future.

Vanille-N:
Alright, thanks. Then I'm happy with the current state of this
Still open to continue working on it if you think that the error messages have become too long and some of the information needs to be moved to hints.

tingerrr:
> Still open to continue working on it if you think that the error messages have become too long and some of the information needs to be moved to hints.

I think we should do that, the error messages can be fairly long in these cases, if we move parts into the hints they get a little more digestible in my opinion.
```
error: package not found
  ┌─ pending/main.typ:1:8
  │
1 │ #import "@local/mypkg:0.0.1"
  │
  = hint: the namespace '@local' does not have package 'mypkg'
  = hint: searched at /home/user/.local/share/typst/packages/local/mypkg
```

The first hint could be adjusted according to the actual failure (namespace missing, package missing, file within missing, etc.).

I think, generally, what is placed in parentheses can almost always be moved to a hint.

crates/typst-library/src/diag.rs
```diff
@@ -533,40 +576,55 @@ pub enum PackageError {
...
+                        eco_format!(
+                            "the package {} exists, but not with version {}",
+                            spec.name,
+                            spec.version
+                        ),
+                        eco_format!(
+                            "the registry at {registry} provides up to version {version}"
+                        ),
+                    ]
+                } else {
```
tingerrr, line 669: Does the else branch here happen realistically, for `typst init`? I'd assume that, if there is no version for a package, then the package does not exist, in which case it should emit the `NotFound` variant.

crates/typst-library/src/diag.rs
```diff
@@ -507,24 +554,20 @@ impl From<PackageError> for FileError {
...
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum PackageError {
     /// The specified package does not exist.
-    NotFound(PackageSpec),
+    /// Additionally provides information on where we tried to find the package.
+    NotFound(PackageSpec, EcoString),
     /// The specified package found, but the version does not exist.
-    VersionNotFound(PackageSpec, PackageVersion),
+    /// TODO: make the registry part of the error better typed
+    VersionNotFound(PackageSpec, Option<PackageVersion>, EcoString),
```
tingerrr: Judging by the comment, I assume the `EcoString` is the registry? I'd assume it was additional information like above, perhaps use a named enum variant to make this clearer.

crates/typst-library/src/diag.rs
```diff
@@ -439,13 +449,13 @@ pub enum FileError {
...
+    AccessDenied(PathBuf),
     /// A directory was found, but a file was expected.
-    IsDirectory,
+    IsDirectory(PathBuf),
     /// The file is not a Typst source file, but should have been.
-    NotSource,
+    NotSource(PathBuf),
     /// The file was not valid UTF-8, but should have been.
-    InvalidUtf8,
+    InvalidUtf8(Option<PathBuf>),
```
tingerrr: Can the `From<Utf8Error>` conversions be adjusted to instead require the path? Using `.map_err` at the appropriate places. I don't think this would ever happen without a path, even with stdin we construct a fake `FileId` that would be used to construct `<stdin>` or so as the path.

But perhaps that not too important.

tingerrr (commented):
I've noted a few things down, thanks for taking my suggestions into consideration!

crates/typst-library/src/diag.rs
Vanille-N: Currently working on this. I'm hitting a wall in that this enum:
- must be visible by`typst-library` (that's where `VersionNotFound` is)
- will be constructed by `typst-kit` (probably)
- probably should be able to read `DEFAULT_NAMESPACE` and `DEFAULT_REGISTRY` for printing.

So yeah, I'm trying to find a good way to solve this cross-crate dependency cycle.

tingerrr: Adding the registry and namespace as strings to the error variant should solve the issue, no? It's just a matter of constructing it with that additional information when it is available, which it should be when that error is emitted (from within `typst-kit`).

Vanille-N: To clarify, I was hoping that I could make this new type like

```rs
enum PackageRegistry {
    Online,
    Local(EcoString),
}
```
and centralize a bit the logic.

But that implies a lot of changes everywhere, so instead you can see what I've implemented just now.

crates/typst-library/src/diag.rs
Vanille-N: Since the file structure is `namespace/package/version`, I see a difference between `NotFound` (`namespace/package` doesn't exist) vs `VersionNotFound(None)` (`namespace/package` exists but is empty),
and that was the logic behind keeping it as a `VersionNotFound`.

crates/typst-library/src/diag.rs
Vanille-N: So to clarify I should remove the `impl From<FromUtf8Error> for FileError` and instead manually do the conversions with the file path right ?

crates/typst-library/src/diag.rs
tingerrr: Ah, I only considered it originating from the registry, good point, although the package path existing with no version directories in it is more likely an indicator for an aborted downloading process. The error message makes it seems like the registry returned a package with no versions (which should be impossible).

Vanille-N:
Urgh. Ignore that, I made a mistake while rebasing.

crates/typst-library/src/diag.rs
tingerrr: If a path is available at all sites where the conversion is used, then yes, otherwise no.

Vanille-N:
Alright, I'm back to this.

Regarding the utf-8 error requiring a path: every call site has a path accessible. I'm not sure if `id.vpath().as_rooted_path()` is the best path to use, but at least it is something.

`PackageRegistry` is now a struct instead of a raw `EcoString`.

I assume the currently failing tests are just due to small changes is the wording of some messages, it'll be a quick fix.

Vanille-N:
Oh. No idea how to solve that.
```
mismatched message (tests/suite\scripting\import.typ:323):
    annotated | tried to read /assets/images/rhino.png
    emitted   | tried to read \assets\images\rhino.png
```

tingerrr:
> Oh. No idea how to solve that.
> 
> ```
> mismatched message (tests/suite\scripting\import.typ:323):
>     annotated | tried to read /assets/images/rhino.png
>     emitted   | tried to read \assets\images\rhino.png
> ```

Ah, that'll require some change to the test runner, I think there are two obvious solutions:
- Add some support for special variables that the runner will interpolate before doing comparisons of the diagnostics, such as `$PATH_SEPARATOR` being turned into `\` on Windows and `/` otherwise.
- Always use Unix paths in diagnostic annotations and change how the strings are compared by making `/` compare equal to `\`.

I would go with the former because it gives greater control when these characters should be substituted, and it can be re-used for other things too. Though I'm not sure about the syntax for it.

laurmaedje:
The test runner already has a hack for this, but it appears that it only happens for the message, not the hints:

https://github.com/typst/typst/blob/3d64f7b659905755322c419c5cd1fdb2078590e8/tests/src/run.rs#L231-L235

laurmaedje:
The hack for keeping `\u{..}` intact is fairly recent and shows the limits of the current approach. On the other hand, something like `$PATH_SEPARATOR` would get noisy really quickly as there a lot of tests with paths in the error message.

tingerrr:
> The hack for keeping `\u{..}` intact is fairly recent and shows the limits of the current approach. On the other hand, something like `$PATH_SEPARATOR` would get noisy really quickly as there a lot of tests with paths in the error message.

Yes it's quite annoying really, I have not yet found a nice way to make unit tests behave nicely when they compare paths.

In other projects, I've thought about adding some sort of compile-time flag to switch paths to a normalized form for unit tests regardless of platform. It is the only approach I could think of that doesn't seem like a hack, but it's quite the scope creep and extra work for path printing.

Should they just go with doing the same hack for annotations for now?

laurmaedje:
> Should they just go with doing the same hack for annotations for now?

That would be okay with me. Would be good to factor it out into a function though.

Vanille-N:
There will be an additional hack needed though.

`rg 'Error:.*\\'` reveals that the only instance of `\` in error messages is for `\u{...}`,
however there are more in Hints.
```
$ rg 'Hint:.*\\'
tests/suite/scripting/blocks.typ
140:// Hint: 1-2 try using a backslash escape: \]
149:// Hint: 10-11 try using a backslash escape: \]
160:// Hint: 6-7 try using a backslash escape: \]
164:// Hint: 4-5 try using a backslash escape: \]
169:// Hint: 7-8 try using a backslash escape: \]
175:// Hint: 4-5 try using a backslash escape: \]

tests/suite/model/cite.typ
148:// Hint: 7-17 use `label("%@&#*!\\")` to create a label
```

I'll think of a way to not make it too ad-hoc.

tingerrr:
Let me know if you figure something nice out, and I'll use it in my test suites too.

Vanille-N:
So in principle this approach works.
Let me know if you find that satisfactory.

laurmaedje:
@Vanille-N please excuse the long radio silence. I was primarily shying away from landing this because I wasn't confident in the changes in the public API. However, with https://github.com/typst/typst/pull/7555 the public API will change quite a bit anyway, so it's a good opportunity to take another stab at this. My plan was to re-review and reconsider the public API changes once that PR has landed and this PR here is rebased.

Vanille-N:
Alright. Then I'll ignore merge conflicts until #7555 lands and rebase all at once.

Vanille-N:
I tried to have a look at the new API, and it seems it's going to be a challenge to get a concrete path from a `VirtualRoot`. Am I missing something ?

laurmaedje:
In principle, the API should be roughly isomorphic to the old one (`VirtualPath::realize` versus the old `VirtualPath::resolve`). I am currently still making some changes in the area though, which would result in further conflicts. These should land very soon. I'll let you know once that's the case and we'll discuss it, alright?

laurmaedje:
#7710 has landed, so we should now be ready to proceed here.

> I tried to have a look at the new API, and it seems it's going to be a challenge to get a concrete path from a VirtualRoot. Am I missing something ?

I'm not sure what exactly you mean, maybe you can elaborate. Also, feel free to ping me on Discord to discuss this.

laurmaedje:
Hi @Vanille-N, are you still interested in working on this? If not, that's also totally fine, but then I would close the PR.

Vanille-N:
Ah, yeah sorry for letting this get stale.

What happened is that the merge conflicts were too big to rebase, to the point that the final result was a bit nonsensical. This would have to be redone from scratch.

I'm still interested in solving this issue eventually, but it won't be this PR and probably not right now.

PR #6435, opened 2025-06-12
Consume `data` argument in `pdf.embed()`

### LINKED ISSUE #6394: Misleading error for unexpected argument of data in pdf.embed
### Description

The `pdf.embed` function is documented as taking an optional positional argument for file data. However, that the argument is interpreted as `none` if it is anything but `bytes`. This leads to hard to diagnose behavior when such data is supplied by decoding (i.e. `cbor`) or has otherwise dynamic type.

```
#pdf.embed("example.txt", "oops-this-is-text")
```

```
error: file not found (searched at /tmp/example.txt)
  ┌─ what.typ:1:11
  │
1 │ #pdf.embed("example.txt", "oops-this-is-text")
  │            ^^^^^^^^^^^^^
```

### Expected behavior

```
error: invalid argument type, expected bytes, got string
  ┌─ what.typ:1:11
  │
1 │ #pdf.embed("example.txt", "oops-this-is-text")
  │                           ^^^^^^^^^^^^^^^^^^^
```

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

cAttte:
this PR fixes #6394 by consuming the `data` argument of `pdf.embed()` with `eat()` instead of peeking it with `find()`, so that it will fail with a type error rather than an unexpected arg error.

but also wouldn't it be nicer if this function accepted strings for `data`?


laurmaedje:
It's a common convention across data-taking functions that bytes mean data and a string means a path. It's the same here for consistency, even if we would know that the second string is data.

Could you add a regression test for this?

cAttte:
that makes sense! added the test

laurmaedje:
Thanks!

PR #6436, opened 2025-06-12
Prefer `.yaml` over `.yml` in the docs

Andrew15-5:
A more broad change than https://github.com/typst/typst/pull/6426, same as https://github.com/typst/hayagriva/pull/332.


laurmaedje:
I'm okay with changing this in the user-facing docs, but don't want to change the filenames as `.yml` is simply this extension we use across all our repos and I see no point in changing this.

docs/reference/groups.yaml
laurmaedje, line 1: could you also revert this change to keep the file extension in this repository consistent?

Andrew15-5: Hm, not sure why it's like that. I definitely didn't wanna revert the one that is mentioned [here](https://github.com/typst/typst/pull/6354#issuecomment-2964827798).

Unless I'm forgetting something, I guess it's also non-user-facing.

PR #6441, opened 2025-06-12
Allow deprecating symbol variants

MDLC01:
This is the companion PR to https://github.com/typst/codex/pull/86.

The deprecation message is emitted when the symbol is modified. This means accessing the variant is what triggers the warning (i.e., the symbol does not need to be inserted in the document, and in fact once the variant has been accessed, no further warning will be emitted when the symbol is used). It spans over each modifier that forces the symbol to resolve to a deprecated variant. For example, consider the following symbol:

```
symbol
  .modifier
  @deprecated: ...
  .modifier.alt
```

When accessed in Typst, the warnings appear at the place indicated in the comments:

```typ
#let x = symbol.modifier.alt
//                       ^^^

#x // No warning

#symbol.modifier.alt
//               ^^^

#symbol.modifier // No warning

#symbol.alt
//      ^^^

#symbol.alt.modifier
//      ^^^ ^^^^^^^^ (two warnings)
```

The last example is unfortunate. I will try to fix it in the upcoming days.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -413,7 +413,7 @@ dependencies = [
 [[package]]
 name = "codex"
 version = "0.1.1"
-source = "git+https://github.com/typst/codex?rev=56eb217#56eb2172fc0670f4c1c8b79a63d11f9354e5babe"
+source = "git+https://github.com/typst/codex?rev=a5428cb#a5428cb9c81a41354d44b44dbd5a16a710bbd928"
 
 [[package]]
 name = "color-print"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ clap = { version = "4.4", features = ["derive", "env", "wrap_help"] }
 clap_complete = "4.2.1"
 clap_mangen = "0.2.10"
 codespan-reporting = "0.11"
-codex = { git = "https://github.com/typst/codex", rev = "56eb217" }
+codex = { git = "https://github.com/typst/codex", rev = "a5428cb" }
 color-print = "0.3.6"
 comemo = "0.4"
 csv = "1"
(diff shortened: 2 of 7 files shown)


laurmaedje:
> The last example is unfortunate. I will try to fix it in the upcoming days.

Is this ready for merge or are you still working on that?

MDLC01:
Oh sorry, I had completely forgotten about that. I'll try to work on that tonight.

MDLC01:
Actually, solving this problem is much harder than I initially thought.

Consider the following situation:

```
symbol
  @deprecated: reason 1
  .modifier
  @deprecated: reason 2
  .modifier.alt
```

Where should there be warnings in the following Typst code?

```typ
#let a = symbol.modifier
#let b = symbol.modifier.alt
#let c = a.alt
```

To summarize the problem: we probably want a single warning on each line. But I have no idea how this could be implemented.

MDLC01:
A solution is to warn on usage instead, but this means assigning a symbol to a variable in a package would (1) not warn the package author, and (2) cause a warning in the code of the users of the package.

T0mstone:
Maybe a custom `DeprecationSink` for `FieldAccess::eval` could work. I haven't really thought this through that much though.

T0mstone:
I got something working, but it may be a bit of an intrusive change. (And also possibly qualify as spaghetti – you be the judge.)

https://github.com/T0mstone/typst/commit/5c585a1914658d713e8086199f94e768dfc562f1

MDLC01:
Interesting, I'll take a closer look at that tomorrow if @laurmaedje thinks it's worth the added code complexity.

laurmaedje:
I think it's not quite worth the complexity. I fear that the code would be a bit hard to maintain this way. The two warnings aren't that bad in my opinion.

---

Thanks!

PR #6442, opened 2025-06-13
Require parentheses in all function-like sub/superscripts

mkorje:
Reverts #985, removing the "smart" treatment of function like stuff. This solution ensures things are consistent across symbols, function calls, literal characters, etc. (cf. #5605). AFAIK This has been the consensus reached as the best and final option. Whilst there could in the future be runtime math parsing - which would likely mean another change here - that would be a fairly large throughout math in any event and enable the "perfect" solution. So I think it would be okay to change again for that.

The changes in summary:
```typ
#set page(width: auto)
$a_1(x), a^10(x), a_f(x), a^zeta(x), a_"text"(x),
 a_∂(x), a^cos(x), a_abs(x), a^root(2, x), a_subset.eq(x) \
 a_(1(x)), a^(10(x)), a_(f(x)), a^(zeta(x)), a_("text"(x)),
 a_(∂(x)), a^(cos(x)), a_(abs(x)), a^(root(2, x)), a_(subset.eq(x))$
```
| Old | New |
| - | - |
| old | new |


laurmaedje:
I agree that this change is necessary and that we've waited far too long for the perfect solution.

That said, I think we should help out people a bit with migrating back to this behaviour. Since the current behaviour is rather rarely the desirable one, I imagine that most people will only have superfluous spaces after this change instead of breakage. However, people that are using Typst functions in attachments will have silent breakage. I think we should thus issue a warning when a Typst function (like `abs`) is used as an attachment. 

Any other ideas for warnings we could issue?

A general migration assistant would also be really useful for changes like this, but that's another story.

Enivex:
Fractions have a similar problem as well

https://github.com/typst/typst/issues/4717 is relevant

laurmaedje:
I'm not sure I agree that it's a problem for fractions. I would naturally expect function calls to bind more tightly than fractions, e.g. in `f(x)/g(x)`.

Enivex:
> I'm not sure I agree that it's a problem for fractions. I would naturally expect function calls to bind more tightly than fractions, e.g. in `f(x)/g(x)`.

The thing is that you don't know if something is actually a function or not, which is the same problem as for fractions.

`1/2(a+b)` is interpreted as `(1/2)*(a+b)`, which is correct when strictly following standard precedence rules. However `1/c(a+b)` is interpreted as `1/(c(a+b))`.

mkorje:
> That said, I think we should help out people a bit with migrating back to this behaviour. Since the current behaviour is rather rarely the desirable one, I imagine that most people will only have superfluous spaces after this change instead of breakage. However, people that are using Typst functions in attachments will have silent breakage. I think we should thus issue a warning when a Typst function (like `abs`) is used as an attachment.

I've got a warning working, but it also includes user's functions. But for some reason the span seems quite off for user's functions:
```typ
--- playground ---
$a^abs(x) a^(abs(x)) a^abs a^(abs)$
$a_abs(x) a_(abs(x)) a_abs a_(abs)$
$a^std.math.abs(x) a^(std.math.abs(x)) a^std.math.abs a^(std.math.abs)$
$a_std.math.abs(x) a_(std.math.abs(x)) a_std.math.abs a_(std.math.abs)$

#let aa = $a$
#let a = $a$
$a^#a a^aa$

#let myfunc(x) = if x == none {} else {}
#let f(x) = if x == none {} else {}
$a^myfunc(x) a^(myfunc(x)) a^myfunc a^(myfunc)$
$a^#f($x$) a^#f$
```

<details>

<summary>Warnings</summary>

```bash
warning: function literal used as superscript
  ┌─ tests/suite/playground.typ:1:3
  │
1 │ $a^abs(x) a^(abs(x)) a^abs a^(abs)$
  │    ^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as superscript
  ┌─ tests/suite/playground.typ:1:23
  │
1 │ $a^abs(x) a^(abs(x)) a^abs a^(abs)$
  │                        ^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as subscript
  ┌─ tests/suite/playground.typ:2:3
  │
2 │ $a_abs(x) a_(abs(x)) a_abs a_(abs)$
  │    ^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as subscript
  ┌─ tests/suite/playground.typ:2:23
  │
2 │ $a_abs(x) a_(abs(x)) a_abs a_(abs)$
  │                        ^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as superscript
  ┌─ tests/suite/playground.typ:3:3
  │
3 │ $a^std.math.abs(x) a^(std.math.abs(x)) a^std.math.abs a^(std.math.abs)$
  │    ^^^^^^^^^^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as superscript
  ┌─ tests/suite/playground.typ:3:41
  │
3 │ $a^std.math.abs(x) a^(std.math.abs(x)) a^std.math.abs a^(std.math.abs)$
  │                                          ^^^^^^^^^^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as subscript
  ┌─ tests/suite/playground.typ:4:3
  │
4 │ $a_std.math.abs(x) a_(std.math.abs(x)) a_std.math.abs a_(std.math.abs)$
  │    ^^^^^^^^^^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as subscript
  ┌─ tests/suite/playground.typ:4:41
  │
4 │ $a_std.math.abs(x) a_(std.math.abs(x)) a_std.math.abs a_(std.math.abs)$
  │                                          ^^^^^^^^^^^^
  │
  = hint: wrap the entire function call in parentheses

warning: function literal used as superscript
   ┌─ tests/suite/playground.typ:10:11
   │
10 │ #let myfunc(x) = if x == none {} else {}
   │            ^^^
   │
   = hint: wrap the entire function call in parentheses

warning: function literal used as superscript
   ┌─ tests/suite/playground.typ:11:6
   │
11 │ #let f(x) = if x == none {} else {}
   │       ^^^
   │
   = hint: wrap the entire function call in parentheses
```

</details>

mkorje:
> Fractions have a similar problem as well

Fractions aren't quite as simple to "fix". For the denominator its easy, but for the numerator we have the same problem with fixing the #896 equivalent for fractions: we've  already parsed it and can't go back.

laurmaedje:
> I've got a warning working, but it also includes user's functions.

I think including user function is good, isn't it?

> But for some reason the span seems quite off for user's functions:

I suggest using the span of the attachment `expr` instead of the span of the function. That should fix it.

mkorje:
> I think including user function is good, isn't it?

Yes I think so, but I wasn't sure if you might think that's too broad.

> I suggest using the span of the attachment `expr` instead of the span of the function. That should fix it.

Oh that's quite obvious. I've done that now.

laurmaedje:
Before merging this, I wanted to go through the topic one more time. I've done this over the past weeks and collected my thoughts on the topic in a blog post: https://laurmaedje.github.io/posts/math-mode-problem/

The TLDR is that my opinion has shifted back and forth over the past two weeks, but that I'm now quite firmly of the opinion that the PR should be merged and that it is not just a temporary, but the final solution (and not something like runtime parsing).

However, I want to wait one more week for people to voice opinions, should they wish to.

YDX-2147483647:
Hi! I didn't know a discussion is going on until I saw the blog. I appreciate your effort, but I want to express my objection: I vote for **Option A _Do nothing_** over Option B _Revert to Typst 0.3 behavior_.

If I understand correctly, this PR aims to improve syntax consistency, but at the cost of turning something that can be resolved by inserting a space, into something requiring additional parentheses.
- Result: $f_n (x)$ and $e^{f(x)}$
- The shortest math markup:
  - Before: `f_n (x)` and `e^f(x)`
  - After: `f_n(x)` and `e^(f(x))`. 

## A predictable, concise, but inconsistent convention is acceptable

However, **math is primarily about writing — not programming**. If all choices of convention is equally predictable, I think it’s reasonable to choose the one that is more concise but less consistent.

`f_n(x)` ≠ `f_n (x)` is just like the divergence between code & markup mode, and `x-y ≠ x - y`.
It seems odd at first glance, but becomes handy once you know it. We could add a teaching example to [the tutorial](https://typst.app/docs/tutorial/writing-in-typst/#maths).

Some analogies from other domains:

- In programming, we strictly distinguish between functions and their evaluated results, even `type(f + g) ≠ type(f(x) + g(x))`.
   But in math, we often mix them without confusion — $f + g$ and $f(x) + g(x)$ usually mean the same thing.

- In Python, `id(some_integer)` doesn't behave consistently due to small integer caching, but it's acceptable: 

	<details>
	<summary>Example python code</summary>


	```python
	>>> x = 5
	>>> id(x) == id(5)
	True
	>>> x = 12345543
	>>> id(x) == id(12345543)
	False
	```

	</details>

## More parentheses, harder editing

Adding more parentheses also **makes editing formulas much more cumbersome**.
While the change makes it clearer about how elements are grouped, it can make it less clear about how parentheses are paired.

Let's say changing the first formula below into the second, then into the third.



(with [physica](https://typst.app/universe/package/physica) imported)

Before:

```typst
$ evaluated(e^sin(x))_abs(u+v) $
$ evaluated(e^sin((n pi) / (N+1)))_abs(u+v) $
$ evaluated(e^sin((n pi) / (N+1) + theta))_abs(u+v) $
```

After:

```typst
$ evaluated(e^(sin(x)))_(abs(u+v)) $
$ evaluated(e^(sin((n pi) / (N+1))))_(abs(u+v)) $
$ evaluated(e^(sin((n pi) / (N+1) + theta)))_(abs(u+v)) $
```

Even though there's only one pair of `()` in the formula, the old syntax uses 3 → 5 → 5 pairs, making it hard enough to find where to insert `+ theta`. However the new syntax increases that to 5 → 7 → 7 pairs, which is crazy. 

LaTeX (only for reference):

```latex
% \usepackage{physics}
\begin{align}
  \eval{e^{\sin(x)}}_\abs{u + v} \\
  \eval{e^{\sin(\frac{n\pi}{N+1})}}_\abs{u + v} \\
  \eval{e^{\sin(\frac{n\pi}{N+1} + \theta)}}_\abs{u + v} \\
\end{align}
```

In VS Code, nested parentheses are colorized by level, which helps a bit. But in typst.app, all parentheses are black.



(As Enivex has mentioned, #4717 also describes the trouble. That issue proposes a solution, which is another story and I reserve my opinion.)

----

## Maybe collect more voices?

In my opinion, Discord & PR can be biased. If you think there's a problem and want changes, then you are more likely to search for and participate in the discussion. Meanwhile, if I don't see it an issue or oppose the change, then I might not even be aware that a discussion is happening. 

Could we consider reaching out to a broader audience?
Maybe, posting it officially on the Typst Forum, or wait this until typst 0.15 and ask for feedback in typst 0.14 release notes, or anything other methods?

laurmaedje:
> If I understand correctly, this PR aims to improve syntax consistency [...]

The main aim is not to improve consistency. The current system and option B are equally consistent[^1]: Right now, function calls have higher precedence than attachments. With option B, attachments have higher precedence.

The aim instead is to resolve an unintuitive behavior that many people stumble into, which is that `f_n(x)` puts the parentheses into the subscript. From all I could observe, this is simply not what most people expect.

> [...] but at the cost of turning something that can be resolved by inserting a space, into something requiring additional parentheses.

I think the crucial part is that with option B, the ambiguity is always resolved into something that reads naturally: It's easily understandable and clear what `e^(abs(x))` is doing.

Meanwhile, with option A, the ambiguity is resolved into something that reads unnaturally and feels strange for many people: `f_i (x)` just _feels wrong_ because you also wouldn't ever write `f (x)`. The space here reads like an implied multiplication rather than a function application.

> However, math is primarily about writing — not programming. If all choices of convention is equally predictable, I think it’s reasonable to choose the one that is more concise but less consistent.

I would say that typically consistency is actually more important than slightly shorter notation, as it makes things easier to learn and read.

> In VS Code, nested parentheses are colorized by level, which helps a bit. But in typst.app, all parentheses are black.

I honestly think the current feature set of typst.app shouldn't inform the language design. I think distinguishing at a highlighting level which brackets are rendered and which aren't would definitely be a great feature, and colorization maybe as an optional add-on (though I personally am not a fan).

> Meanwhile, if I don't see it an issue or oppose the change, then I might not even be > aware that a discussion is happening.
> 
> Could we consider reaching out to a broader audience?

This is a fair point. I'm not sure I want to wait for 0.15 and it's not like 0.14 will release next week. But posting on the Forum definitely can't hurt.

[^1]: There is a slight inconsistency right now because `f_1(x)` renders differently than `f_a(x)`. Because `1(x)` is not considered a function call.

laurmaedje:
I created a [Forum post](https://forum.typst.app/t/opinions-needed-likely-upcoming-changes-to-math-mode-precedence/4974) asking for opinions.

OverflowCat:
Here is just my rough idea: I think the core of the issue here is that **whitespace in Typst's math mode has semantic meaning, but that meaning isn't currently clear**. If we educate users with this principle, it could be easily understood and provide a unified logic.

> "We already have the concept of a space being needed for implied multiplication"

This isn't correct; it's just the visual arrangement of symbols, e.g. `A inter B`.

> "Of course, you could require the opposite `(i.e. 1/(x(a+b)))`, but I’m not convinced it’s better."

Actually, I prefer this way. One of the big advantages of Typst's math mode over TeX, in my opinion, is the reduction in parentheses. If a simple space can resolve a nesting issue, that should be the preferred solution. Take `1/2(a+b)`: interpreting `2(a+b)` as part of the denominator would make input more convenient, especially because if you do not want this, you can just add a space before the parenthesis. In more extreme cases, like `f(x)/a(b+c)d(e)`, all continuously written parentheses should be treated as a high-precedence unit. By this design we can remove many parentheses for fractions. 

Since we determine that math mode primarily handles display, `f (x)` is then perfectly acceptable. We also shouldn't differentiate between any symbols; for example, `1(x)` is used as Heaviside step function in signal processing.

Other thoughts:

If we want to make math mode even more semantic, we can use the type system (still up in the air) to let user define what is a function, but that will be way too complex.

laurmaedje:
> > "We already have the concept of a space being needed for implied multiplication"

> This isn't correct; it's just the visual arrangement of symbols, e.g. A inter B.

I didn't mean that every space means implied multiplication, but that a space can be necessary to write an implied multiplication (e.g. in `a b`).

> If a simple space can resolve a nesting issue, that should be the preferred solution.

I'm not sure what exactly you mean with "[s]ince we determine that math mode primarily handles display", but as mentioned in my previous comment, I personally am of the opinion that having the source read intuitively and produce intuitive output is more important than a minor save in length of the equation.

bilgincoskun:
As someone who follows and occasionally tests Typst, I vote for option A. While this behavior initially seemed strange, I quickly got used to it. Inserting spaces between subscripts felt very quick and natural. To me, B is like Latex parentheses and is more verbose than option A.

surbliss:
I made a post on the Forum-discussion, but I will cross-post it here for visibility:

I know I’m a bit late, but I think that an **Option F** is missing from the suggested solutions: Always greedily match subscripts as far right as possible. That is, `f_i(x)`, `f_1(2+2)` or `g_3(y)` would put everything to the right of `_` into the subscript. This eliminates all ambiguity, and lets the user determine this with spaces.

This behavior mirrors what new users encounter with implicit multiplication: Writing `ab + x` produces an error, as `ab` is parsed as a function. Thus, they have to write `a b + x`. In exchange, they get functions without leading delimiters (`#` or `\`), leading to much cleaner syntax.

The same can be applied for subscripts. In long expressions, reducing parentheses makes code more readable. Letting users control subscripts with `x_2(i+1)` or `f_2 (i+1)` achieves this.

Cons:
- Requires more spaces in the simple cases such as `f_1 (x)`
- For simple subscripts, this contradicts the expected behavior from LaTeX, unlike the proposed solution.
- Is a breaking change, (with wide-spread effect)

Pros:
- Achieves the same unambiguous parsing as the proposed solution
- Allows using spaces as a typographic tool, leading to more readable syntax-code than wide parentheses.

I hope this option can be considered -- I think a minimal syntax is a key strength of Typst, and is a goal that should be pursued.

NicoWeio:
To those arguing for Option A: 
Doing nothing would also mean that the IMO unintuitive, thus undesirable, difference between `f_1(x)` and `f_g(x)` (rendered as `f_1 (x)` and `f_(g(x))`), as discussed in #2086, would stay, or otherwise still require a breaking change, albeit of slightly smaller scope.

matteosantama:
Having just started my first project in typst, I can say I really appreciate its conciseness. FWIW I am in favor of Option F, proposed by @surbliss. Typing a space is much easier than adding a pair of left-right brackets.

laurmaedje:
Isn't Option F basically the status quo with a small extension to also handling numbers like identifiers? All the downsides of Option A apply just the same and my points on this would essentially be https://github.com/typst/typst/pull/6442#issuecomment-3047919322.

bilgincoskun:
In my opinion, F (and A) are more convenient and read more naturally, especially when dealing with nested parentheses. I think the real problem with user expectations isn't the lack of parentheses, but rather the inconsistency. If the distinction between numbers and identifiers were eliminated, and the subscript part was greedily matched, users would adapt as soon as they encountered it.

laurmaedje:
@bilgincoskun Interestingly though, [right above](https://github.com/typst/typst/pull/6442#issuecomment-3056704650) you said "While this behavior initially seemed strange". Isn't that sort of in conflict with it "read[ing] more naturally"?

bilgincoskun:
@laurmaedje Not exactly. In my first comment, I meant that I tried writing this way because I was familiar with Latex, and I encountered unexpected behavior compared to it and I think it is fair to assume that most Typst users are familiar with Latex and subscript problem stems from the habits transferred from Latex.

However, after that first encounter, I started writing as f_x (x), and it became easier for me to read and write. I think the more concise writing outweighs the the initial confusion. I hope this clarifies my thoughts.

laurmaedje:
I see. Personally, I still think that `f_x(x)` is more natural than `f_x (x)` as `f(x)` is also more natural than `f (x)`, but I can still how some people might prefer brevity over some perceived naturalness. 

I don't think it's possible to make everyone happy here regardless of what we do.

surbliss:
> Isn't Option F basically the status quo with a small extension to also handling numbers like identifiers? All the downsides of Option A apply just the same and my points on this would essentially be [#6442 (comment)](https://github.com/typst/typst/pull/6442#issuecomment-3047919322).

The primary point of option F is to utilize spaces for typography, rather than trying to infer semantics. Extending the current behavior to numbers does, in my opinion, increase readability. It enforces the rule "Spaces indicate what goes into the subscript". Enforcing this behavior consistently helps users learn to read the spaces in  `f_x (x)` and `f_2 (x)` as "This space is where the subscript ends".

> I see. Personally, I still think that `f_x(x)` is more natural than `f_x (x)` as `f(x)` is also more natural than `f (x)`, but I can still how some people might prefer brevity over some perceived naturalness.
> 
> I don't think it's possible to make everyone happy here regardless of what we do.

I think the same could be said for writing `a b+c` rather than `ab+c`, as you can in Latex. But there is obviously a very good reason to require spaces for implicit multiplication. Avoiding backslashes in math-expressions is to me one of the main selling points of Typst over Latex.

laurmaedje:
Some time has passed since I've asked for feedback and there were quite a few responses on this PR and in the [Forum](https://forum.typst.app/t/opinions-needed-likely-upcoming-changes-to-math-mode-precedence/4974). There were some voices for most of the Options (stay as-is, revert to 0.3, runtime-dependant parsing), including for an option not included in my blog post: Basically Option A on steroids, where we expand the current behaviour to numbers (Option F).

If you're interested in the individual arguments, I invite you to reread the discussion as I can't fully reproduce everything here. But as a quick summary: The main argument in favor of Option B was predictability while the main argument in favor of Option A and F was terseness, primarily avoidance of additional parentheses, which can make equations harder to read.

There is no way to make everyone happy here, so it's going to be a difficult decision either way. Personally, I believe that predictable, consistent, and intuitive syntax is more important than a bit of extra tersity (this is also in line with Typst's generally design philosophy). While I wouldn't say there is clear consensus, the majority of upvotes on both GitHub and in the Forum seem to share this sentiment. For this reason, I have decided to move forward with Option B. 

Thanks to everybody who left feedback, in particular also those that argued for other options. Also thanks to @mkorje of course, for the PR!

Trebor-Huang:
> A general migration assistant would also be really useful for changes like this, but that's another story.

Now that the decision has been made, maybe we can talk more about this? I would very strongly prefer if we have a "proper" migration tool, using the full parser instead of (say) adhoc regex replacements based on heuristics, guaranteeing correctness.

laurmaedje:
We can start talking about this for the future, but for this particular change and 0.14, we don't have the capacity to ship that in time.

Ikaluk:
However this problem is going to be resolved, might it be possible implement both option B as default math mode and option F as alternative math mode settable by something like #set math.spacescope = true ?

AlbertoFabbri93:
I think having an option would make things very confusing

PR #6447, opened 2025-06-14
Reorder FigureCaption fields for nicer PlainText representation

cAttte:
hi, i noticed that the plaintext representation of figure captions was off, for example in tooltips:
image

i fixed this by just reordering the fields of `FigureCaption` (moving `supplement` before `kind`). code looks a bit weird with the `#[synthesized]`s out of order buuut anyway here's how it looks now, for example in a completion:

```rs
Completion {
    kind: Label,
    label: "yea",
    apply: None,
    detail: Some(
        "Figure: hi",
    ),
}
```

### DIFF
--- a/crates/typst-library/src/model/figure.rs
+++ b/crates/typst-library/src/model/figure.rs
@@ -503,6 +503,10 @@ pub struct FigureCaption {
     #[default(OuterVAlignment::Bottom)]
     pub position: OuterVAlignment,
 
+    /// The figure's supplement.
+    #[synthesized]
+    pub supplement: Option<Content>,
+
     /// The separator which will appear between the number and body.
     ///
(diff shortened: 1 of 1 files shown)


laurmaedje:
The intent was that the tooltip just says "hi" and this used to work. I think this regressed at some point. The code that creates this string is here: 

https://github.com/typst/typst/blob/0bc68df2a9a87ca7e36e34dab56b07c666d64760/crates/typst-ide/src/analyze.rs#L75-L89

Maybe it's falling back to the `unwrap_or(elem)` branch for some reason even though it shouldn't?

laurmaedje:
Superseded by https://github.com/typst/typst/pull/6580.

Thanks still for bringing my attention to this! I had seen the bug before, but had completely forgotten about it.

Issue #6028, opened 2025-03-08
A more general `symbol` function


zacharie-moughanim:
### Description

User-declared symbols are resticted to unicode symbols, I would like to have either a similar function, declarable by the user where we can store mathematical content, or to extend the class of content that can be declared in a `symbol` function.

Currently, the closest feature is to declare a dictionary of content, but it does not have a default value. For instance `arrow` is already a symbol, whereas a dictionary `dict` is only the dictionary, we have to give a key to access to some content.

### Use Case

```typst
// Making use of a bunch of different variations around `arrow`, with often the same parameters (base, leftmost, trans, ...) only adding an accent or a subscript.

// Reductions on vanilla
#let ttriangle = symbol(sym.triangle, ("arrow", "⇾"))
#let leftmostReductionBase = $scripts(->)_"lm"$
#let leftmostReduction = $scripts(->^circle)_"lm"$
#let leftmostReductions = $leftmostReduction^*$
#let canonReductionBase = $scripts(ttriangle.arrow)$
#let canonReduction = $scripts(ttriangle.arrow^circle)$
#let oneStepReduction = $tilde(-->)$
#let oneStepReductionBase = $-->$
#let baseReduction = $-->$

// Reductions on weak
#let oneStepReductionWeakBase = $scripts(arrow.hook)$
#let oneStepReductionWeak = $scripts(arrow.hook^ballot)$
#let oneStepReductionWeakRefl = $oneStepReductionWeak^?$
#let oneStepReductionWeakNonZeroBase = $scripts(arrow.dashed)$
#let oneStepReductionWeakNonZero = $scripts(accent(arrow.dashed, dot.triple))$
#let leftmostReductionWeak = $scripts(hat(->))_"lm"$
#let leftmostReductionWeakTrans = $leftmostReductionWeak^+$
#let leftmostReductionWeaks = $leftmostReductionWeak^*$

#let leftmostReductionWeakNonzeroBase = $scripts(->_z)_"lm"$
#let leftmostReductionWeakNonzero = $scripts(hat(hat(->_z)))_"lm"$
#let leftmostReductionWeakNonzeroTrans = $leftmostReductionWeakNonzero^+$

#let parallelLeftmostReductionWeakBase = $scripts(=>)_"lm"$
#let parallelLeftmostReductionWeak = $hat(parallelLeftmostReductionWeakBase)$
#let parallelLeftmostReductionWeakBaseCount(k) = $scripts(=>)^((#k))_"lm"$
[...]

PR #6448, opened 2025-06-14
Add negative variants to math spacings

### LINKED ISSUE #6431: Add negative version of spaces
### Description

I would like to see negative width versions of spaces added e.g. `cal(T)_(thin.neg phi_N)`, to quickly tweak kerning. Effectively, LaTeX's `\!`.

### Use Case

I often use Typst in fairly time-constrained situations (i.e. writing conference slides) and due to unorthodox styling I often end up with small kerning issues. Ofcourse, the right thing to do is to bring out `#h`, but given that it is a one-off document, a negative thin space or two tends to be good enough and is quicker to remember/type.

cAttte:
hi, this PR adds a `.neg` variant to all five predefined math spacings. i did this by replacing the existing Content values with Modules. a few things:

- in #4912 you (laurmaedje) said that we can't have this due to the current infrastructure, so maybe i'm missing something. if this isn't the right / nice way to do it just lmk and close this!
- the code is pretty ugly because i kept the pub consts (`THIN` etc.) as is, to avoid a breaking change for external users. but i doubt anyone is importing these, so i could refactor this and fix up the internal user (`typst_layout`). lmk
- a significant drawback is that autocomplete and tooltips now show e.g. `<module quad>`, so you can't see the length value. i think it would make sense to update the `Repr for Module` to also show the content if it's present. something like `<module quad>: h(amount: 1em)`

this closes #6431


laurmaedje:
This is really interesting. I hadn't considered that this is even a possibility as the fact that a module interpolated into content shows as its content is barely used. I think in normal usage this would just work. 

That said, modules don't behave _just_ like content or symbols. For instance, you can't pass it to a function that expects content. That wouldn't be a big problem in this case I think, but more generally I'm trying to figure out how much we should start mixing symbols and modules for the same purpose (we've also started going down this road in codex) or whether this will lead to us ending up in troubling waters.

PgBiel:
I wouldn't mind having symbols behave like modules, but let's please keep them separate types as the confusion is gonna be huge otherwise in my opinion haha. Modules are quite overloaded already, and at this point would basically become a JS do-it-all "object" type. Not that this would be the end of the world, but I feel like at least some separation is easier to reason about.

laurmaedje:
I think I'm with @PgBiel here. While certainly an interesting approach, this is a little cursed and will likely lead to confusion down the road. Thank you still!

MDLC01:
For reference, this would be made possible by https://github.com/typst/typst/issues/6028.

PR #6450, opened 2025-06-14
Error on imports with keyword names

### LINKED ISSUE #6276: Inconsistent behavior with file named `set.typ`
### Description

The setup is to have three files, as described below:
```typst
// [set.typ]
// (empty)

// [a.typ]
// #import "set.typ" as set // errors
#import "set.typ" // works

// [b.typ]
// #import "a.typ": set // errors
// #import "a.typ": set as name // errors
#import "a.typ"
#let name = a.set // works
```
I can imagine two ways to make this consistent:
1. Make `#import "set.typ"` error. (But `#import "set.typ" as name` should continue to work.)
2. Make `#import "a.typ": set as name` not error, since it should be equivalent to `#let name = a.set`.

I'm fairly sure we'll want to go with 1. since having a module member called `set` seems inconsistent with `set` being a keyword.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

noamzaks:
Fixes #6276.


PgBiel:
Added some thoughts to the original issue (let's discuss there first): https://github.com/typst/typst/issues/6276#issuecomment-2973264952

laurmaedje:
We discussed this and [decided](https://github.com/typst/typst/issues/6276#issuecomment-3068989088) that we want to keep the ability to import with keyword names and, if anything, explore raw identifiers syntax for accessing such bindings, so I'll close this PR. Thanks still!

PR #6451, opened 2025-06-14
Fix align link in layout documentation

### LINKED ISSUE #6427: The link in the doc of Reference/Layout is not working correctly
### Description

I'd like to report a broken link on the typst.app docs.
The link for `align` appears in [Reference/Layout](https://typst.app/docs/reference/layout/) in the item for `alignment`, but it's not working.
The line comes from the first line of `alignment`'s documentation.

Image

I apologize if this is already known and I overlooked it.

noamzaks:
Fixes #6427.

I would love to try to create something that automatically checks if links are broken, if that seems like a good idea.

### DIFF
--- a/crates/typst-library/src/layout/align.rs
+++ b/crates/typst-library/src/layout/align.rs
@@ -104,7 +104,7 @@ impl Show for Packed<AlignElem> {
     }
 }
 
-/// Where to [align]($align) something along an axis.
+/// Where to align something along an axis.
 ///
 /// Possible values are:
 /// - `start`: Aligns at the [start]($direction.start) of the [text
--- a/crates/typst-library/src/layout/align.rs
+++ b/crates/typst-library/src/layout/align.rs
@@ -104,7 +104,7 @@ impl Show for Packed<AlignElem> {
     }
 }
 
-/// Where to [align] something along an axis.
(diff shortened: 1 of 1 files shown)


pauladam94:
Your comment I think is a great idea if it could be done easily. I have done a feature request about it.

laurmaedje:
> I would love to try to create something that automatically checks if links are broken, if that seems like a good idea.

I think it's not worth investing any more into the current docs infrastructure as it will be replaced in the nearish future with a new Typst-based one.

noamzaks:
Ah right. I think I already asked about this 😓

laurmaedje:
Actually, looking at this again, I don't think this will fix it. The problem isn't that the link is broken, the problem is that the summary line (the first line) just does not support links atm. Let's just remove the link instead.

crates/typst-library/src/layout/align.rs
```diff
@@ -104,7 +104,7 @@ impl Show for Packed<AlignElem> {
     }
 }
 
-/// Where to [align] something along an axis.
+/// Where to [align]($align) something along an axis.
```
noamzaks:
```suggestion
/// Where to align something along an axis.
```

laurmaedje:
Thanks!

Issue #1988, opened 2023-08-24
Add `datetime.now`


emilyyyylime:
### Description

Would work as `datetime.today`, but also adding the local OS time at time of compilation.

### Use Case

My specific use case is really that I want my documents to render in dark mode if I'm editing them at night, but really this should be added for consistency and extensibility; I don't see a reason why this shouldn't be part of Typst besides "bloat". But it's not something a package can add by itself, so I think this change *is* worth it.

Also at the very least `hour:minute` could be desirable in a letter etc.

emilyyyylime:
Also [Good first issue](https://github.com/typst/typst/labels/good%20first%20issue)

LaurenzV:
This feature has already been suggested a couple of times, but it probably won't happen, see [here](https://github.com/typst/typst/pull/1843#issuecomment-1660692106) for more information.

MrPixelized:
Having the build time of the document available might somehow feel problematic for live preview, but there are plenty of use cases. Here's some that I've encountered personally:

- I generate a document based on RSS feeds and weather data regularly. A timestamp for these documents would be helpful, as well as the ability to position the sun in the overview based on the current time using CeTZ.
- I have a typst document I'm working on in Git. On push, it gets built. Including the timestamp for those builds would be nice.
- An organization I'm involved with sends people automatically-generated old-school letters that would benefit from a rough timestamp.
- I use typst to generate figures based on data that changes over time. Having a "status of the information at [timestamp]" line would be nice.
- There is a LaTeX template at my university which includes the compilation time on the front page. Using typst to submit homework would be allowed only if the template can be faithfully recreated.

All of these can be achieved in other ways, such as changing the data that comes in, manually editing things, or asking people to check out metadata instead of document content. Nevertheless, a `datetime.build_time()` or `datetime.now()` object would be very helpful for these and plenty of other use cases.

laurmaedje:
I feel like it is okay if there's some friction in getting the current time. Otherwise, people will start abusing it for things like randomness and it will hurt compilation. If you're generating RSS feeds, you're likely in the CLI and it's quite easy to just provide an `--input` then than you can consume via `sys.inputs`.

MrPixelized:
A student who wants a "finished the assignment at: {}" line defaulting to the last time the document was compiled can't always be expected to be messing around with command line arguments right? What about the web editor?

jgclark:
[...]

PR #6458, opened 2025-06-16
feat(world): Add functionality to get the current time

IuvenisSapiens:
In `world.rs`, the `Timelike` trait was introduced to handle time-related operations.

A `now` method was added to the `SystemWorld` struct to get the current time and adjust the timezone based on the provided offset. This method first checks the state of `self.now`; if it is `Fixed`, it uses the fixed time directly; if it is `System`, it retrieves the current UTC time. Then, based on the `offset` parameter, it decides whether to use the local timezone or the specified offset for adjustment, and finally converts the adjusted time to the `Datetime` type.

A `now` function was added to `datetime.rs` to get the current time in the Typst engine. This function accepts an `offset` parameter; if `offset` is `None`, it returns the local time; if `offset` is a specified number of hours, it adjusts the UTC time by that number of hours and returns the result. This function achieves the functionality of getting the current time by calling the `now` method of `SystemWorld` and converts the result to a string.

In the `lib.rs` file, a `now` method was added to the `World` trait to define the functionality of getting the current time. Additionally, the `world_impl` macro was updated to implement the `now` method for structs that implement the `World` trait, ensuring that all structs implementing this trait can get the current time using the `now` method.

### DIFF
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::sync::{LazyLock, OnceLock};
 use std::{fmt, fs, io, mem};
 
-use chrono::{DateTime, Datelike, FixedOffset, Local, Utc};
+use chrono::{DateTime, Datelike, FixedOffset, Local, Timelike, Utc};
 use ecow::{eco_format, EcoString};
 use parking_lot::Mutex;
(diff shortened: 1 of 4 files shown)


PartyWumpus:
This feature has already been suggested and rejected: #1988

Enivex:
As mentioned above, and in your earlier issue, this omission is intentional. I'll close this PR, because it's highly unlikely to be merged.

PR #6463, opened 2025-06-17
Update dependencies and migrate to Rust 2024 edition

IuvenisSapiens:
- Bumps numerous dependencies to their latest versions and updates edition and minimum required Rust version to 2024 and 1.87, respectively. 
- Adjusts code for compatibility with upstream API changes, improves trait object pointer handling, and resolves minor logic and type issues. 
- Enhances future maintainability and ensures compatibility with latest ecosystem improvements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.83.0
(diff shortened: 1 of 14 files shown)


laurmaedje:
Thanks for your work on this, but we typically don't accept PRs that bump dependencies. :/ (This should be added to the contributing docs). It's something we prefer to do ourselves, both to keep an overview of what's where and also for security reasons. 

Same with the MSRV bump. We don't need to keep it 1.83 much longer, but 1.87, being the latest, is a bit too new. As you can see in CI, the general checks are already updated to 1.87, but the MSRV is intentionally lower.

PR #6473, opened 2025-06-20
CLI: Allow specifying output directory

titaniumtraveler:
This adds the feature to specify an output *directory* using an explicit trailing `/`.
(example: `build/`)

That makes it possible to use
```bash
typst compile -f pdf main.typ build/
```
where the output is then written to `build/main.pdf`.

---

My primary usecase for this is to make it easy to change either input file, or output format, without also having to change the output file.

### DIFF
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -96,12 +96,9 @@ impl CompileConfig {
         let output_format = if let Some(specified) = args.format {
             specified
         } else if let Some(Output::Path(output)) = &args.output {
-            match output.extension() {
-                Some(ext) if ext.eq_ignore_ascii_case("pdf") => OutputFormat::Pdf,
-                Some(ext) if ext.eq_ignore_ascii_case("png") => OutputFormat::Png,
-                Some(ext) if ext.eq_ignore_ascii_case("svg") => OutputFormat::Svg,
-                Some(ext) if ext.eq_ignore_ascii_case("html") => OutputFormat::Html,
-                _ => bail!(
+            match output.extension().and_then(OutputFormat::from_file_ext) {
+                Some(format) => format,
+                None => bail!(
                     "could not infer output format for path {}.\n\
                      consider providing the format manually with `--format/-f`",
                     output.display()
@@ -116,12 +113,7 @@ impl CompileConfig {
                 let Input::Path(path) = &input else {
                     panic!("output must be specified when input is from stdin, as guarded by the CLI");
                 };
-                Output::Path(path.with_extension(match output_format {
-                    OutputFormat::Pdf => "pdf",
-                    OutputFormat::Png => "png",
-                    OutputFormat::Svg => "svg",
-                    OutputFormat::Html => "html",
-                }))
+                Output::Path(path.with_extension(output_format.as_file_ext()))
             }
             // Check if a [`Path`] has a trailing `/` (or on `windows` a `\`) character,
             // indicating that the output should be written to `{output}/{input_file_name}.{ext}`
@@ -146,12 +138,7 @@ impl CompileConfig {
                 })?;
 
                 path.push(file_name);
-                path.set_extension(match output_format {
-                    OutputFormat::Pdf => "pdf",
-                    OutputFormat::Png => "png",
-                    OutputFormat::Svg => "svg",
-                    OutputFormat::Html => "html",
-                });
+                path.set_extension(output_format.as_file_ext());
                 Output::Path(path)
             }
             Some(output) => output,
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -111,19 +111,43 @@ impl CompileConfig {
             OutputFormat::Pdf
         };
 
-        let output = args.output.clone().unwrap_or_else(|| {
-            let Input::Path(path) = &input else {
-                panic!("output must be specified when input is from stdin, as guarded by the CLI");
-            };
-            Output::Path(path.with_extension(
-                match output_format {
+        let output = match args.output.clone() {
+            None => {
+                let Input::Path(path) = &input else {
+                    panic!("output must be specified when input is from stdin, as guarded by the CLI");
+                };
+                Output::Path(path.with_extension(match output_format {
                     OutputFormat::Pdf => "pdf",
                     OutputFormat::Png => "png",
                     OutputFormat::Svg => "svg",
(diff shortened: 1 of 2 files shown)


crates/typst-cli/src/compile.rs
```diff
@@ -111,19 +111,51 @@ impl CompileConfig {
...
-        });
+                }))
+            }
+            // Check if a [`Path`] has a trailing `/` (or on `windows` a `\`) character,
+            // indicating that the output should be written to `{output}/{input_file_name}.{ext}`
+            Some(Output::Path(mut path)) if has_trailing_path_separator(&path) => {
+                let Input::Path(input) = &input else {
+                    bail!(
+                        "can't infer output file when input is from stdin\n\
+                         consider providing a full path to write to, or get input from file",
```
Andrew15-5, line 124: I don't think Typst has multi-line errors, it probably wouldn't look great. I don't think you can attach `.hint()` to [`bail!`](https://github.com/typst/typst/blob/87c56865606e027f552a4dbc800c6851b0d0b821/crates/typst-macros/src/util.rs?plain=1#L8-L22), but it sounds like a hint, that you can add for `HintedStrResult`.

crates/typst-cli/src/compile.rs
```diff
@@ -111,19 +111,51 @@ impl CompileConfig {
...
+                    );
+                };
+
+                let Some(file_name) = input.file_name() else {
+                    panic!("input path must be non-empty, as guarded by the CLI");
```
Andrew15-5, line 129: Compiler shouldn't panic intentionally.

crates/typst-cli/src/compile.rs
```diff
@@ -111,19 +111,51 @@ impl CompileConfig {
...
+                    )
+                })?;
+
+                path.push(file_name);
+                path.set_extension(match output_format {
+                    OutputFormat::Pdf => "pdf",
+                    OutputFormat::Png => "png",
+                    OutputFormat::Svg => "svg",
+                    OutputFormat::Html => "html",
+                });
```
Andrew15-5: This is duplicated, I feel like better make a mapping closure, because there are 4 variants, and more will be added.

titaniumtraveler:
> I don't like all those CLIs that change behavior based on leading or trailing `/` (I can remember `rsync`, `docker`). This forces you to remember or re-learn every single time what each of the CLIs do in each situation.

I mean there isn't really a good way to indicate that you are talking about a directory and not a file. Especially if it doesn't exist yet.

And even `ln` and `ls` have slightly different behavior with a trailing `/`.

> The same goes to these changes. I don't see why this is significant enough to be added. Considering how long your command already is, typing the file name is not that bad. By providing the file name, you can drop the ` -f pdf` part.

Yeah, but if you switch between different inputs/formats, you have to change a lot parts of the command *twice*, which makes it unnecessarily annoying to use in `sh`.

But really, the main reason why I wrote the PR, is that I felt there is a gap in the semantics of `typst-cli`; The CLI *already* has the logic to determine the output filename based on context, but there is no way to tell it to put them into a different output directory.

(Like a `build/`, which can be easily `.gitignore`ed.)

> If you only use this with `main` file, then the file name is very short and simple. If you do this with other ones, then you probably have a project cooking, so why not use justfile or Nix at this point.
> 
> What I actually do feel is missing, is the ability to auto-create missing directories when using `typst c main.typ a/b/c/main.pdf`.

To be fair, that is what I *usually* do too. In this case I just avoided it because I work with someone else together.

crates/typst-cli/src/compile.rs
titaniumtraveler: It is doing it just a few lines above:
https://github.com/typst/typst/blob/c4f5ba84d184011dc9d238ce13abefd52eaa99c4/crates/typst-cli/src/compile.rs#L104-L107

crates/typst-cli/src/compile.rs
titaniumtraveler: Arguably it is more an assert than a panic.
The same thing happens a few lines above too:

https://github.com/typst/typst/blob/a18ca3481da17a4de1cc7f9890f0c61efb480655/crates/typst-cli/src/compile.rs#L116

crates/typst-cli/src/compile.rs
titaniumtraveler: Fair point.
Though I would prefer a method on `OutputFormat`.
The file extension is a global property and it makes it easier to see why the function is there when adding new variants.

crates/typst-cli/src/compile.rs
Andrew15-5: Fair enough. Actually, you can add `hint:` to [`bail`](https://github.com/typst/typst/blob/87c56865606e027f552a4dbc800c6851b0d0b821/crates/typst-library/src/diag.rs?plain=1#L43), but it will create `HintedString`.

crates/typst-cli/src/compile.rs
Andrew15-5: clap kinda sucks like this, huh. I guess you can't sanely use `assert!` with enums...

crates/typst-cli/src/compile.rs
Andrew15-5: Yeah, method sounds good, actually, if I don't miss anything.

Andrew15-5:
> Yeah, but if you switch between different inputs/formats, you have to change a lot parts of the command _twice_, which makes it unnecessarily annoying to use in `sh`.

If you switch between formats, you only change output file extension.

You can always use the most niche aliases that does exactly what you need.

```sh
alias typcdir='_(){ typst c "$1" "$2/${1%.typ}.$3"; }; _'
```

```
 typcdir main.typ build pdf
typst c main.typ build/main.pdf
```

You can even skip `.typ` completely, which will also make alias and invocation shorter than `typst c -f pdf main.typ build/` (`typcd main build pdf`).

> In this case I just avoided it because I work with someone else together.

So it's a niche case even for you, let alone everyone else. Is this someone can't use any abstraction whatsoever? Surely there is a solution for this.

laurmaedje:
With HTML export, there will be native directory output with multiple files. We're also considering to design this feature in a generalized way that will allow to create a directory containing an arbitrary mix of HTML files, PDF files, assets, etc. I'm not sure I like the feature proposed here, but as the multi-file output features are still in the design phase, we'd either way need to explore how they would interact with the feature proposed here. For now, I'll thus close this PR. Thank you still!

PR #6476, opened 2025-06-20
Typed HTML API

laurmaedje:
This PR introduces a typed API for constructing HTML elements in Typst. This makes for a much more convenient way of constructing HTML elements than the raw `html.elem` API.

## Examples

Here are a few examples for elements, just to showcase how typed the API is:

```typ
#html.div(role: "article")[My article ...]
#html.time(datetime: datetime.today())
#html.span[A bit of text]
#html.input(accept: ("image/jpeg", "image/png"))
#html.meter(low: 3.4, high: 7.9)
#html.link(rel: "icon", color: red)
#html.form(autocomplete: false)
#html.img(
  src: "image.png",
  alt: "My wonderful image",
  srcset: (
    (src: "/image-120px.png", width: 120),
    (src: "/image-60px.png", width: 60),
  ),
)
```

This will yield the following HTML elements.

```html
<div role="article">My article …</div>
<time datetime="1970-01-01"></time>
<span>A bit of text</span>
<input accept="image/jpeg, image/png" />
<meter low="3.4" high="7.9"></meter>
<link rel="icon" color="#ff4136">
<form autocomplete="off"></form>
<img
  src="image.png"
  alt="My wonderful image"
  srcset="/image-120px.png 120w, /image-60px.png 60w"
/>
```

As you can see, the API expresses each HTML attribute with its best fitting Typst type (colors, datetimes, integers, sometimes structured dictionaries, etc.), so that you can just work with the native types. No need to convert stuff to strings all the time or bother figuring out whether HTML uses `"true"`, `"on"`, `"yes"`, or just the presence of an attribute to express `true` this time around.

On top of the specific and global HTML attributes, the API also includes the [ARIA](https://www.w3.org/TR/wai-aria-1.1/) accessibility attributes.

Error messages also work as you would expect:

```
error: expected array, "additions", "additions text", "all", "removals", or "text"
  ┌─ hi.typ:1:26
  │
1 │ #html.link(aria-relevant: "nope")
  │                           ^^^^^^
```

## Implementation

The data backing this API lives in `crates/typst-library/src/html/generated.rs` and is auto-generated from the HTML specification. The code responsible for generating this file lives in `tools/codegen/src/html.rs`. It automatically downloads the HTML (and a few other) specifications and extracts all the information it needs. The information is mostly taken from the [Index section](https://html.spec.whatwg.org/#index) of the spec, but some stuff is scraped from elsewhere.

In the specification, the valid values for each HTML attribute are described in human-readable text. These descriptions are matched against a bunch of regular expressions during codegen to figure out the best Typst type. In the end, each attribute is assigned a Rust type like `Str`, `NonZeroU64`, or `TokenList<f64, ' '>`. The latter would mean that the attribute should be expressed as an array of floats in Typst and turned into a space-separated string in HTML. Among other things, codegen also automatically extracts stringy enums for attributes that follow such patterns. This results in a type like `StrEnum<282, 285>`, meaning that HTML defines 3 valid strings for this attribute and those can be found in `generated::ATTR_STRINGS[282..285]` (sometimes alongside descriptions, when those were conveniently available for scraping).

A type that can be used as an attribute implements the standard `Reflect` & `FromValue` traits used across Typst and on top of that `IntoAttr` (or sometimes `IntoOptionalAttr` if `false` means the attribute should be omitted). The latter traits are used to serialize the value into an HTML attribute string. Since all types used for attributes implement the standard reflection trait, all the information is also automatically available for docs, autocomplete, and error messages.

From the data, the functions are dynamically built during creation of the `Library` struct. This makes them very lightweight in terms of crate compile time in comparison to a design that would attempt to generate functions for everything at compile time via macros. Supporting this required minor changes to how native functions work. They are now backed by `&'static dyn Fn(..) -> ..` instead of `fn(..) -> ..`.

## Unresolved questions

- ~~Currently all attributes have their exact HTML names. Should we convert them to kebab-case (just like JS converts them to camelCase)?~~ (Conclusion: No)


T0mstone:
I don't see a copyright notice, but I think one will be necessary given that the HTML spec is licensed under CC BY and this includes pieces of it verbatim into our source code.

laurmaedje:
Right, I thought about that and then forgot about it. I'll add it.

Edit: Added in https://github.com/typst/typst-assets/pull/10/commits/a70ec8995e8ea59b66a6d3351ffaea2f16441872

Myriad-Dreamin:
I used [`@vscode/web-custom-data`](https://www.npmjs.com/package/@vscode/web-custom-data) to generate my typed HTML typst module. Using the json files from the package will help reproducible code generation (by locking version of npm package to use) and avoid souping data from HTML files.

MDLC01:
Why did you chose to include the generated file in the repo instead of generating it as part of a `build.rs`  script?

laurmaedje:
because the necessary specs would otherwise need to be checked in to keep things reproducible. I didn't want that in general but also because they weigh >15MB in total. and the API will be rarely updated so it's a bit of a waste to have everybody keep rerunning the codegen (which is not at all written to be efficient).

laurmaedje:
I agree that it seems unsatisfactory to check in generated code but it's the lesser evil here I think

MDLC01:
Would it be possible have the generated stuff in typst-assets in some way, and then import that here and convert it to the right format (because typst-assets does not have access to the types defined here AFAIK)? Somehow it feels less problematic to me to have generated content as part of a separate repo. But maybe it isn't that important.

PartyWumpus:
This looks like it won't provide any way to include `data-*` attributes (or any kind of non-standard attribute in general). Is this omission intentional?

laurmaedje:
> Would it be possible have the generated stuff in typst-assets in some way, and then import that here and convert it to the right format (because typst-assets does not have access to the types defined here AFAIK)? Somehow it feels less problematic to me to have generated content as part of a separate repo. But maybe it isn't that important.

@MDLC01 I had considered having this in another crate and the difficulty was indeed the types. I didn't manage to find a clean solution. However, with the hindsight of building the whole type-based machinery, I was able to do it after all while reusing almost all of the previous code with some adaptions. The codegen and data now lives in https://github.com/typst/typst-assets/pull/10, which I also much prefer over having it here. :) 

The one downside is that it's impossible to have the tag and attribute constants in another crate (as these types live here and depend on PicoStr) and it's not possible to generate module items with const fns. I just added these to the code manually now. They are much easier to maintain by hand than the element to attribute mappings and attribute types. And they are not user-facing, they are just for the convenience of having compile time constant tag names available in code that needs to match on tags and such. So if some tag or attribute is missing, it's not a big deal.

laurmaedje:
> This looks like it won't provide any way to include data-* attributes (or any kind of non-standard attribute in general). Is this omission intentional?

@PartyWumpus Good call. @reknih and I had discussed that, but I then forgot about it again. We were considering to support a `data` dictionary alongside the other global attributes. As for non-standard attributes, we could in theory also support an `attrs` escape hatch, but I'm not sure it's really a good idea. There is always `html.elem` for full control.

laurmaedje:
> I used [@vscode/web-custom-data](https://www.npmjs.com/package/@vscode/web-custom-data) to generate my typed HTML typst module. Using the json files from the package will help reproducible code generation (by locking version of npm package to use) and avoid souping data from HTML files.

@Myriad-Dreamin I attempted to switch from spec scraping to using `@vscode/web-custom-data` and `@webref/idl`, but hit various problems:

- The `<area hreflang>` attribute is listed in it (with no docs) even though it does not actually exist

- Various attributes do not have documentation

- The value sets are the most useful part because it's the most annoying part to scrape. But somehow they are not always correctly applied. E.g. `` has one, but `<a referrerpolicy>` doesn't. 

You can also observe all of these problems in VS Code. I was already sort of convinced to switch to the VS Code data set, but it just doesn't feel very polished to me. I'm not saying that my mappings are bug-free, but at least if we find a bug it's easy to fix. If something is missing or wrong in this data set, you either have to hand-fill it, give up, or _also_ parse the spec in addition. When just parsing the spec right away, we have full control and all information.

Beyond that:

- Many attributes have no value sets. As you mentioned, it's possible to use `@webref/idl` to get the JS types. However, mapping tag names to IDL interfaces seems non-trivial, e.g. getting from `td` to `HtmlTableCellElement`. At least I found no such mapping in `@webref/idl` or `@vscode/web-custom-data`.

- Value sets don't tell you whether the attribute is actually a token list. E.g. the `sandbox` attribute has a value set but actually allows multiple of the listed strings, and it should allow an array in Typst. This could probably be resolved by inspecting the IDL and seeing that it is a `DOMTokenList` though.

- Even if you manage to map all attributes to IDL attributes, the IDL attributes are sometimes less typed than I'd like (e.g. `coords` is a DOMString even though it actually represents an array of floats and does so in my mapping).

And dealing with the IDL feels somehow even more painful than scraping the spec, in part because `@webref/idl` is really not well documented, but also because the interfaces can have inheritance, mixins, etc.

PartyWumpus:
> As for non-standard attributes, we could in theory also support an `attrs` escape hatch, but I'm not sure it's really a good idea. There is always `html.elem` for full control.

I think having any way of doing an escape hatch for non standard attributes would end up being a little odd at best (when `html.elem` exists fine) and I think won't be particularly common, just wanted to make sure it'd been considered.

> We were considered to support a `data` dictionary alongside the other global attributes. 

I think this however is probably a good idea? I'm not really sure how much people need there is though, and it _would_ be unlike the rest of the interface...

Importantly both of those could be added Later™ so probably easiest to just leave em out for now add them later if it ends up being needed.

MDLC01:
Would it be possible to accept a sink that only allows named arguments whose name starts with "data-"? I can think of the following issues with that: error messages, quality of auto-completion, and documentation. With a bit of work, all those issues could probably be solved. But of course this would be more work than just accepting a `data` dictionary.

laurmaedje:
It wouldn't be too hard I think. Truly not sure whether I like it better or not though.

wrzian:
> Currently all attributes have their exact HTML names. Should we convert them to kebab-case (just like JS converts them to camelCase)?

I think we should use the exact HTML names to avoid confusion and make it easier to predict.

One reason is that I don't think we should add dashes to aria attrs with multiple words, i.e. `aria-activedescendant` because `aria-active-descendant` is just horrible. And then choosing to add dashes to e.g. `popovertarget` would feel inconsistent.

For prediction, I think there would be ambiguity on what counts as a word boundary. If a dev already knows the `popovertarget` attribute, I could see them expecting either `popover-target` or `pop-over-target` to be the kebab-case name. While I feel the first option is best, that choice seems subjective and I would rather delegate authority to the spec.

laurmaedje:
@wrzian We would choose the same word boundaries as JS, so at least there would be some precedent. But I've discussed this with @reknih and we've also agreed that it's better to just keep the exact HTML names. It's a bit pointless. 

---

Regarding the data attributes: I will leave them out for now because we are undecided as of yet whether it should be a `data` dictionary or an argument sink with `data-*` arguments.

PR #6479, opened 2025-06-21
Preparations for Typed HTML API

laurmaedje:
The first few commits from https://github.com/typst/typst/pull/6476 are useful changes, but a bit unrelated to the main matter of the PR, so I've extracted them here (and will not squash during merge).


laurmaedje:
Nice, didn't know about that. Hopefully, it will become a thing soon.

PR #6481, opened 2025-06-21
Improve equation reference example

Andrew15-5:


laurmaedje:
I agree that this is perhaps slightly more idiomatic as it avoids indentation, though in the end it's personal preference.

PR #6483, opened 2025-06-22
Table multiple footers and subfooters

### LINKED ISSUE #5375: Headers in arbitrary positions

### LINKED ISSUE #5377: Table subheaders and subfooters

PgBiel:
Closes #5375 and #5377

Similar to #6168, but for footers:

- Table footers can now be placed **at any row, not only the last** (then, they will repeat **up to** where they were placed).
- A table can now have **more than one footer**.
    - Each will repeat from the previous one (or the start of the table) until its actual position, prompting the next one to start repeating.
- A table can now have **subfooters**: footers with **levels.** Similarly to subheaders, but in the opposite way:
    - When footers with different levels are placed, they repeat simultaneously.
    - Levels form a hierarchy. When a level 2 subfooter is placed, the next level 2, as well as level 3+, subfooters start repeating, as if it ended the previous "table section".

## Implementation details

This will be expanded later. There will still be some significant changes to the code, but here are the most important points:

- Before table layout, we simulate subfooters using an algorithm similar to that of subheader layout, generating a separate "sorted_footers" vector, which contains footers sorted not by their starting row index (which is the normal subfooters vector), but by **at which moment they start repeating**, to make the layout step easier.
  - This is currently done at resolve, but it could be moved to right before the layout step since this is not used at all during HTML export.
  - Or we could just use some conditional to not do it on non-paged targets.
- On table layout, we start by taking the first batch of repeating footers.
- **Widow prevention:** The resulting footer height is then subtracted from the available height on every page in order to avoid widows: rows will be placed taking into account existing footers.
- When a footer is placed, it stops repeating, and when we have placed all consecutive footers, we start repeating new footers.
   - This impl still needs to be improved...

### Considerations on non-repeating footers

Non-repeating footers are still useful for widow prevention, but for this initial implementation I chose to skip widow prevention for non-repeating footers to focus on repeating footers. Still, a sketch of an idea is to add the non-repeating footer height to the footer height to subtract as soon as we reach the row - or unbreakable group of rows - immediately before the non-repeating footer (or its group of consecutive footers).
Then, it's as if the non-repeating footer's height became part of the preceding row's height, ensuring both fit on the same page.

## TODO

- [ ] Gutter at the top of the footer: it is repeated, though there is a problem when you have a header and a footer one after the other: the gutter row will be part of both at the same time, which is weird. Not sure if we should consider this strictly a problem, and we may want to just ignore it, since this is a layout detail.
   - This could generate a problem where the same gutter row could be placed twice in a row, but that was fixed and now the remaining question is purely cosmetic I think.
- [ ] Improve widow prevention impl. In particular we may want to use a "pending footers" approach similar to headers, which would also generalize well to non-repeating footers: we dont stop repeating the current footers when we find them until we place something before the next footers.
   - The current impl stops repeating them immediately, and seems to cause some problems, generating a widow even in a simple demo.
   - However, I wouldn't consider this a blocker for releasing this feature as it's already functional, though it produces some non-ideal output at the moment.
- [ ] Distinguish non-repeated from short-lived footers (footers that might be repeatable, but they would just stop repeating immediately, so why bother).
   - Also not strictly necessary for an initial release.
- [ ] Footers are being cloned all around... can we do better?


laurmaedje:
I will close this for now since it's been inactive for so long and unlikely to become a priority soon. We'll keep the branch of course.

PR #6488, opened 2025-06-23
Bump `krilla` to current Git version

saecki:
Use the current version of krilla, to pull in the remaining changes needed for automatic compression of embedded files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1367,7 +1367,7 @@ dependencies = [
 [[package]]
 name = "krilla"
 version = "0.4.0"
-source = "git+https://github.com/laurenzv/krilla?rev=20c14fe#20c14fefee5002566b3d6668b338bbe2168784e7"
+source = "git+https://github.com/LaurenzV/krilla?rev=20c14fe#20c14fefee5002566b3d6668b338bbe2168784e7"
 dependencies = [
  "base64",
  "bumpalo",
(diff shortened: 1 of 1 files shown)


laurmaedje:
The transitive git dependencies are a bit of a problem. Generally, I avoid git dependencies on things we don't control as they could block us from making a release. (The `krilla` dependencies themselves are fine, I just mean the fontations stuff.)

saecki:
Thanks @LaurenzV for bumping the `skrifa` crate. Now all remaining git dependencies should be fine.

Cargo.toml
```diff
@@ -73,8 +73,8 @@ image = { version = "0.25.5", default-features = false, features = ["png", "jpeg
 indexmap = { version = "2", features = ["serde"] }
 infer = { version = "0.19.0", default-features = false }
 kamadak-exif = "0.6"
-krilla = { version = "0.4.0", default-features = false, features = ["raster-images", "comemo", "rayon"] }
-krilla-svg = "0.1.0"
+krilla = { git = "https://github.com/laurenzv/krilla", rev = "20c14fe", default-features = false, features = ["raster-images", "comemo", "rayon"] }
+krilla-svg = { git = "https://github.com/laurenzv/krilla", rev = "20c14fe" }
```
laurmaedje:
```suggestion
krilla = { git = "https://github.com/LaurenzV/krilla", rev = "20c14fe", default-features = false, features = ["raster-images", "comemo", "rayon"] }
krilla-svg = { git = "https://github.com/LaurenzV/krilla", rev = "20c14fe" }
```
github seems to allow this, but having the canonical case here seems a bit better to me

PR #6489, opened 2025-06-23
Allow multi-character symbols

T0mstone:
This is the companion PR to [codex#92](https://github.com/typst/codex/pull/92).

~~Note that this is blocked on (and already rebased onto) https://github.com/typst/typst/pull/6441 and also blocked on [codex#20](https://github.com/typst/codex/pull/20) as explained below.~~

A bunch of things expected symbols to only have one character.
I mostly changed those to add a new error when the symbol has more than one character. This is technically non-breaking, since multi-character symbols were also an error before, but the error now moved from at their creation to at their use in those cases.

Other technical notes:
- Maybe there should be a new `Repr` for runtime-single-character symbols instead of just allocating them like I've done here?
- I'm REALLY unsure about `layout_symbol`... I'm not really familiar with the layout code, and this change just made it work, but I think a single symbol should maybe be a single fragment? Either way, this probably interacts with [codex#20](https://github.com/typst/codex/pull/20) and any typst-side integration of that (idk if there's already a PR for that), so it's probably best to wait on that before merging this.


T0mstone:
No clue how to fix the failures in `typst-ide` and `typst-docs`.

I guess for the IDE, `CompletionKind::Symbol` can become `Symbol(EcoString)`, but I think that field is only used in proprietary code, so someone from the Team would have to say whether that's a worthwhile change.

As for the docs, this is probably a way more substantial change, given how the current docs store the codepoint as a `u32`, so I also don't think I can fix this by myself.

mkorje:
> I'm REALLY unsure about `layout_symbol`... I'm not really familiar with the layout code, and this change just made it work, but I think a single symbol should maybe be a single fragment? Either way, this probably interacts with [codex#20](https://github.com/typst/codex/pull/20) and any typst-side integration of that (idk if there's already a PR for that), so it's probably best to wait on that before merging this.

We can now just call `GlyphFragment::new` with the whole string (#6336) (this function will error if after shaping the string we don't get a single glyph, i.e. if the string isn't a single grapheme cluster), and instead do something like `elem.text().chars().flat_map(|c| to_style(c, ...))` to style it.

crates/typst-library/src/foundations/symbol.rs
```diff
@@ -377,11 +382,12 @@ impl List {
...
 
 /// A value that can be cast to a symbol.
-pub struct SymbolVariant(EcoString, char);
+pub struct SymbolVariant(EcoString, EcoString);
 
 cast! {
     SymbolVariant,
-    c: char => Self(EcoString::new(), c),
+    c: char => Self(EcoString::new(), c.into()),
+    s: EcoString => Self(EcoString::new(), s),
```
laurmaedje, line 410: Not sure whether we made a decision on whether we want to only accept a single grapheme cluster, but worth bringing up again.

crates/typst-library/src/math/matrix.rs
```diff
@@ -281,7 +281,7 @@ cast! {
     Delimiter,
     self => self.0.into_value(),
     _: NoneValue => Self::none(),
-    v: Symbol => Self::char(v.get())?,
+    v: Symbol => Self::char(v.get().parse::<char>().map_err(|_| "symbol value is longer than one character")?)?,
```
laurmaedje: I think for consistency with the accent case, "expected a single-character symbol" would be a good message here as well.

laurmaedje (commented):
> Maybe there should be a new Repr for runtime-single-character symbols instead of just allocating them like I've done here?

It would indeed be nice if those cases wouldn't allocate.

> I guess for the IDE, CompletionKind::Symbol can become Symbol(EcoString), but I think that field is only used in proprietary code, so someone from the Team would have to say whether that's a worthwhile change.

It's actually not even used in proprietary code atm, but it used to be and might be in the future. But either way `EcoString` is okay. 

> As for the docs, this is probably a way more substantial change, given how the current docs store the codepoint as a u32, so I also don't think I can fix this by myself.

The non-codepoint stuff should be fairly straightforward (i.e. Accent and math class just become None). The `codepoint` can be replaced by a `value` string field. I'll then adjust the proprietary part of the docs generator accordingly.

T0mstone:
> Not sure whether we made a decision on whether we want to only accept a single grapheme cluster, but worth bringing up again.

Good point, let's bring that discussion to this thread.

I'm personally against it (as in, I think arbitrary strings should be allowed) and I don't really remember the arguments for why this would be necessary (So if anyone else has any, please introduce or repeat them here). The only one I can think of is `layout_symbol` requiring a single glyph, but requiring a single grapheme cluster doesn't fix that either since e.g. emoji sequences can fall back to multiple glyphs if the font doesn't support them. For example, the "people holding hands" emoji (🧑‍🤝‍🧑) I added in [codex#104](https://github.com/typst/codex/pull/104) is a single grapheme cluster that can fall back to "🧑🤝🧑".

T0mstone:
Another technical detail: I don't know if `SymbolElem::func` would be better by first checking if the string is a single codepoint and then matching on that. Maybe this is an irrelevant concern tho and then the decision could be made on principle (either by deciding that matching on strings makes it easier to add multi-`char` callables in the future, or by deciding that we'll never have multi-`char` callables anyway, so it'd be better to convert to `char` first and then match on that).

T0mstone:
I changed the symbol repr since I think the previous behavior was probably unintended (such as `symbol("\"")` repr'ing as `symbol(""")`), so it now uses `repr` again to format the strings (both for the modifier sets and for the values).

Other than that, I made some minor changes to no longer use the variable name `c` for things that are now strings and to change `option::IntoIter` into the more idiomatic `iter::Once`.

T0mstone:
> It would indeed be nice if those cases wouldn't allocate.

After thinking about this a bit and trying some stuff, I found that this isn't really possible since `Symbol::get` returns `&str`, but `char` is not UTF-8 encoded. We could use https://docs.rs/utf8char/0.2.0/utf8char/struct.Utf8Char.html or hand-roll something similar.[^1] Do you think that'd be worth it?

[^1]: One thing I tried was `[u8; 4]` combined with `char::encode_utf8` and `str::from_utf8`, but the conversions were a bit excessive, so this would then probably be best as a wrapper type in `typst-utils` or something.

T0mstone:
With this, the tests are now passing, but `layout_symbol` remains unable to handle multi-`char` values. As mentioned in https://github.com/typst/typst/pull/6489#issuecomment-3054359355, I think it needs to be able to handle multiple glyphs regardless of whether we require symbols to be single grapheme clusters or not, but I don't know how to implement this.

And secondly, I should mention that the code as it is now permits empty strings as symbol values, which might cause some other issues. We probably just want to forbid those, right? (In which case I think only `Symbol::construct` would need to be adjusted to reject them).

docs/src/lib.rs
```diff
@@ -729,9 +733,12 @@ fn symbols_model(resolver: &dyn Resolver, group: &GroupData) -> SymbolsModel {
                 name,
                 markup_shorthand: shorthand(typst::syntax::ast::Shorthand::LIST),
                 math_shorthand: shorthand(typst::syntax::ast::MathShorthand::LIST),
-                math_class: typst_utils::default_math_class(c).map(math_class_name),
-                codepoint: c as _,
-                accent: typst::math::Accent::combine(c).is_some(),
+                math_class: value_char.and_then(|c| {
+                    typst_utils::default_math_class(c).map(math_class_name)
+                }),
```
MDLC01: We might want to be more subtle than that here. If `value` is the string `"⤴\u{FE0E}"` (U+2934, which is a Relation, followed by a text presentation selector), we want the math class to be that of the base character of the variation sequence.

Probably same for the `accent` below.

mkorje:
> With this, the tests are now passing, but `layout_symbol` remains unable to handle multi-`char` values. As mentioned in [#6489 (comment)](https://github.com/typst/typst/pull/6489#issuecomment-3054359355), I think it needs to be able to handle multiple glyphs regardless of whether we require symbols to be single grapheme clusters or not, but I don't know how to implement this.


<details>

<summary>I think something like this should work:</summary>

```diff
diff --git a/crates/typst-layout/src/math/fragment.rs b/crates/typst-layout/src/math/fragment.rs
index 758dd401..53aaaa4f 100644
--- a/crates/typst-layout/src/math/fragment.rs
+++ b/crates/typst-layout/src/math/fragment.rs
@@ -15,6 +15,7 @@ use typst_library::text::{features, language, Font, Glyph, TextElem, TextItem};
 use typst_syntax::Span;
 use typst_utils::{default_math_class, Get};
 use unicode_math_class::MathClass;
+use unicode_segmentation::UnicodeSegmentation;
 
 use super::MathContext;
 use crate::inline::create_shape_plan;
@@ -278,6 +279,8 @@ impl GlyphFragment {
         text: &str,
         span: Span,
     ) -> SourceResult<GlyphFragment> {
+        assert!(text.graphemes(true).count() == 1);
+
         let mut buffer = UnicodeBuffer::new();
         buffer.push_str(text);
         buffer.set_language(language(styles));
@@ -300,6 +303,7 @@ impl GlyphFragment {
         );
 
         let buffer = rustybuzz::shape_with_plan(font.rusty(), &plan, buffer);
+        // TODO: deal with multiple glyphs.
         if buffer.len() != 1 {
             bail!(span, "did not get a single glyph after shaping {}", text);
         }
diff --git a/crates/typst-layout/src/math/text.rs b/crates/typst-layout/src/math/text.rs
index e76fb2e7..4e819144 100644
--- a/crates/typst-layout/src/math/text.rs
+++ b/crates/typst-layout/src/math/text.rs
@@ -129,44 +129,48 @@ pub fn layout_symbol(
     ctx: &mut MathContext,
     styles: StyleChain,
 ) -> SourceResult<()> {
-    assert!(
-        elem.text.len() <= 4 && elem.text.chars().count() == 1,
-        "TODO: layout multi-char symbol"
-    );
-    let elem_char = elem
-        .text
-        .chars()
-        .next()
-        .expect("TODO: should an empty symbol value forbidden?");
-
-    // Switch dotless char to normal when we have the dtls OpenType feature.
-    // This should happen before the main styling pass.
-    let dtls = style_dtls();
-    let (unstyled_c, symbol_styles) = match try_dotless(elem_char) {
-        Some(c) if has_dtls_feat(ctx.font) => (c, styles.chain(&dtls)),
-        _ => (elem_char, styles),
-    };
-
     let variant = styles.get(EquationElem::variant);
     let bold = styles.get(EquationElem::bold);
     let italic = styles.get(EquationElem::italic);
+    let dtls = style_dtls();
+    let has_dtls_feat = has_dtls_feat(ctx.font);
+    for text in elem.text.graphemes(true) {
+        // Switch dotless char to normal when we have the dtls OpenType feature.
+        // This should happen before the main styling pass.
+        let (text, symbol_styles) =
+            if has_dtls_feat && text.chars().any(|c| try_dotless(c).is_some()) {
+                (
+                    text.chars()
+                        .map(|c| try_dotless(c).unwrap_or(c))
+                        .collect::<EcoString>(),
+                    styles.chain(&dtls),
+                )
+            } else {
+                (text.into(), styles)
+            };
+
+        let text: EcoString = text
+            .chars()
+            .flat_map(|c| {
+                let style = MathStyle::select(c, variant, bold, italic);
+                to_style(c, style)
+            })
+            .collect();
 
-    let style = MathStyle::select(unstyled_c, variant, bold, italic);
-    let text: EcoString = to_style(unstyled_c, style).collect();
-
-    let fragment: MathFragment =
-        match GlyphFragment::new(ctx.font, symbol_styles, &text, elem.span()) {
-            Ok(mut glyph) => {
-                adjust_glyph_layout(&mut glyph, ctx, styles);
-                glyph.into()
-            }
-            Err(_) => {
-                // Not in the math font, fallback to normal inline text layout.
-                // TODO: Should replace this with proper fallback in [`GlyphFragment::new`].
-                layout_inline_text(&text, elem.span(), ctx, styles)?.into()
-            }
-        };
-    ctx.push(fragment);
+        let fragment: MathFragment =
+            match GlyphFragment::new(ctx.font, symbol_styles, &text, elem.span()) {
+                Ok(mut glyph) => {
+                    adjust_glyph_layout(&mut glyph, ctx, styles);
+                    glyph.into()
+                }
+                Err(_) => {
+                    // Not in the math font, fallback to normal inline text layout.
+                    // TODO: Should replace this with proper fallback in [`GlyphFragment::new`].
+                    layout_inline_text(&text, elem.span(), ctx, styles)?.into()
+                }
+            };
+        ctx.push(fragment);
+    }
     Ok(())
 }
```

</details>

(The code is a bit ugly, sorry about that)

> the "people holding hands" emoji (🧑‍🤝‍🧑) I added in [codex#104](https://github.com/typst/codex/pull/104) is a single grapheme cluster that can fall back to "🧑🤝🧑".

Regarding this if the fallback happens during shaping in math (which it basically always will), the above code just lets it error out for the time being with the "didn't get a single glyph after shaping error", after which it'll fallback anyways to the inline text layout.

Another thing: I thought that the text `0x0030 0xFE0E` or `0x0030 0xFE0F` (can replace 0x0030 with any of the digits 0-9) could lead to problems as if it were styled, say bold, the codepoint would change from `0x0030` and this would now be a non-standardised variation sequence. But this is a non-issue as the shaper will just ignore it.

T0mstone:
> I thought that the text `0x0030 0xFE0E` or `0x0030 0xFE0F` (can replace 0x0030 with any of the digits 0-9) could lead to problems as if it were styled, say bold, the codepoint would change from `0x0030` and this would now be a non-standardised variation sequence. But this is a non-issue as the shaper will just ignore it.

Yeah, this can probably also come up in more complicated scenarios with other grapheme clusters that could perhaps even no longer be single grapheme clusters after the replacement.

That's why I just didn't add the assertion in `GlyphFragment::new` and we can add proper multi-glyph and multi-grapheme-cluster handling there in the future as part of that TODO. (Afaict, there's nothing that makes a single grapheme cluster different from just any text from the view of `GlyphFragment`.)

Another issue that might theoretically come up is if there's ever somehow a grapheme cluster that contains both a dotless and a non-dotless character, then the non-dotless one would be replaced with a dotless one by this algorithm since it only sets styles for the entire cluster. But I think such a cluster is very unlikely if not impossible, so that's probably not a real concern.

docs/src/lib.rs
T0mstone: I now added an ad-hoc function that handles variation sequences. Would you say that's enough or do you know of a more proper way to do it?

T0mstone:
Since there were no objections, I also made `Symbol::construct` reject empty variant values now.

docs/src/lib.rs
mkorje: It's probably fine to just do `value.chars().next().and_then(|c| ...)`. This is what gets used in math layout more or less (really it gets the first character from the start of the cluster the shaper returns, which is always going to be the first character basically).

T0mstone:
Okay I added 1½ tests and fixed a diagnostic bug.

I then also changed the symbol constructor to emit multiple errors where it makes sense instead of stopping at the first one.

What I'm a bit unsure about is the phrasing of the error message for empty-valued symbol variants. I changed it to "empty variant" now to be consistent with the other messages, but maybe a user could think that means variants with an empty _modifier list_ are disallowed? Idk if there's a better way to refer to the _value_ ("empty variant value" sounds awful)...

But that's a minor concern. Either way, I think this is ready for review now.

T0mstone:
Just a reminder: This currently does _not_ require symbols to be single grapheme clusters. As stated above, I see no reason for such a restriction.

mkorje:
> Yeah, this can probably also come up in more complicated scenarios with other grapheme clusters that could perhaps even no longer be single grapheme clusters after the replacement.
> 
> That's why I just didn't add the assertion in `GlyphFragment::new` and we can add proper multi-glyph and multi-grapheme-cluster handling there in the future as part of that TODO. (Afaict, there's nothing that makes a single grapheme cluster different from just any text from the view of `GlyphFragment`.)

Ah good point, though I believe the latin digits are the only characters that we map that have variation sequences (AFAIK for non-emoji stuff these are the only things that could cause changes in grapheme clusters for us in math).
We want to split text up into grapheme clusters for `GlyphFragment` so that there's proper spacing, math properties, etc. all set for each. GlyphFragment intends to lay out a single glyph, which is what a single grapheme cluster approximates.

> Another issue that might theoretically come up is if there's ever somehow a grapheme cluster that contains both a dotless and a non-dotless character, then the non-dotless one would be replaced with a dotless one by this algorithm since it only sets styles for the entire cluster. But I think such a cluster is very unlikely if not impossible, so that's probably not a real concern.

I think this should be impossible, as 'iı' can't be a single grapheme cluster (?).

> Just a reminder: This currently does _not_ require symbols to be single grapheme clusters. As stated above, I see no reason for such a restriction.

I'm still kind of of the opinion that it'd be better to restrict symbols to a single grapheme cluster, since that's more in line with what a _symbol_ is IMO. Though I don't feel too strongly about this, and there's no real harm nor is it an issue implementation wise.

docs/src/lib.rs
T0mstone: Fair point, these are the docs after all. I changed it now to make it align with what the actual code does. (First char for math class; single char or nothing for accent)

T0mstone:
> I'm still kind of of the opinion that it'd be better to restrict symbols to a single grapheme cluster, since that's more in line with what a _symbol_ is IMO.

Personally, I want us to have full content as symbols in the future (making [`symbolx`](https://typst.app/universe/package/symbolx) obsolete), so arbitrary-length strings wouldn't really stand out in comparison.

EDIT: For an example, consider something like
```typst
#let tensor = symbol("⊗", ("real", $⊗_ℝ$))
```

MDLC01:
I did not follow the whole discussion, but I feel like for now we should restrict ourselves to grapheme clusters for symbols. I am also in favor of implementing https://github.com/typst/typst/issues/6028, but then this should be done in one step instead of first shipping a half-backed slightly less restrictive `symbol` type without much thought put into it.

For example, we may want the feature described in https://github.com/typst/typst/issues/6028 to be implemented in a separate type. Then, it would make sense to only allow grapheme clusters in `symbol`.

Essentially, allowing grapheme clusters feels like a bug fix, while allowing arbitrary strings feels like a new feature, which should be implemented in a separate PR, with its own motivations.

T0mstone:
Fine, let's do it like that.

The new error message now does use "variant value", despite what I said about it earlier. I now think that, despite it sounding a bit clunky, the gain in clarity is worth it.

Enivex:
What's the status of this? It would be good to more easily be able to test this stuff.

T0mstone:
My TODO file only had a single bullet point "write tests" left, but I don't actually remember what else still needs new tests since it's been too long.😓

Other than that, I just merged main into this branch and came across a line in `typst_layout::math::fragment::shape_glyph`, where getting multiple glyphs had been marked as `unreachable!()`, but I don't see how that can be true given what we've said here about single grapheme clusters being able to result in multiple glyphs.

crates/typst-library/src/math/accent.rs
```diff
@@ -188,7 +188,7 @@ cast! {
     self => self.0.into_value(),
     v: char => Self::new(v),
     v: Content => match v.to_packed::<SymbolElem>() {
-        Some(elem) => Self::new(elem.text),
-        None => bail!("expected a symbol"),
+        Some(elem) if elem.text.chars().count() == 1 => Self::new(elem.text.chars().next().unwrap()),
+        _ => bail!("expected a single-codepoint symbol"),
```
laurmaedje: this could use the same `parse` approach as in `Delimiter`, right? I didn't know about `FromStr for char`, that's neat!

crates/typst-layout/src/math/fragment.rs
```diff
@@ -946,7 +946,14 @@ fn shape_glyph<'a>(
...
         1 => {}
-        _ => unreachable!(),
+        _ => {
+            todo!("Not quite sure how to resolve this")
+            // from origin/main:
+            //  unreachable!()
+            // from this branch:
+            //  // TODO: deal with multiple glyphs.
+            //  bail!(span, "did not get a single glyph after shaping {}", text);
+        }
```
laurmaedje: @mkorje How should we handle this? I'm not sure this PR changes much w.r.t to main, but in principle a font can emit multiple glyphs for a single char or grapheme cluster if it wanted to (e.g. via a GSUB multiple substitution). It's just fairly unlikely to happen in a math font.

laurmaedje (commented):
Apart from the todo note and a minor nit, this looks merge-ready to me!

crates/typst-layout/src/math/fragment.rs
laurmaedje: @T0mstone Since that's already a panic on main, maybe a bail is a good middle grounds and we can potentially revisit this. That way, we don't have to block this PR further. What do you think?

T0mstone: Well, that's less of a middle grounds and more just going with what I already had here😄
So yeah, totally fine. I don't think I'm currently qualified to engineer the correct text shaping code anyway...

laurmaedje: Yeah, I meant basically that what you already have there seems good enough (as a middle grounds between panic and some smart unknown stuff). If you just put that back in, I'm happy to merge.

T0mstone: Done!

T0mstone:
Ah, using `bail` again means changing the return type of the `shape` functions and `GlyphFragment::new` back from `Option` to `SourceResult`. I'll do that now, but idk if there was some reason for it to be `Option`...

T0mstone:
...or rather than that, I went with `SourceResult<Option<>>`.

T0mstone:
Since we have Clippy in CI, maybe we should add `too_many_arguments = "allow"` to `Cargo.toml` instead of manually silencing it like I did now?
This lint doesn't seem very productive for this codebase.

T0mstone:
In fact, there are 33 instances (32 without the newly added one) of `#[allow(clippy::too_many_arguments)]` in the repo.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -846,8 +850,9 @@ fn shape(
     language: rustybuzz::Language,
     fallback: bool,
     text: &str,
+    span: Span,
```
laurmaedje: IIRC there was intentionally no span here because it hurts memoization. It would be better to return a StrResult and attach the span outside of the memoization boundary with `at`

laurmaedje:
Thanks!

crates/typst-layout/src/math/fragment.rs
mkorje: Yeah I completely forgot about this in the fonts PR, sorry :sweat_smile:. I'll (hopefully) have some time tomorrow where I'll try figure out how to handle this.

PR #6514, opened 2025-06-27
Bump `typst-dev-assets`

mkorje:
@laurmaedje Adding Noto Sans Math stuffed up a bunch of tests due to font fallback, sorry about that :sweat_smile:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2911,7 +2911,7 @@ dependencies = [
 [[package]]
 name = "typst-dev-assets"
 version = "0.13.1"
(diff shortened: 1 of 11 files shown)


laurmaedje:
Noto Sans Math has normal Arabic text glyphs?

mkorje:
Yes, they're used sometimes in math like with non italic Latin. E.g. regular dal and math dal are used differently

mkorje:
But it doesn't have all the stuff needed for proper shaping of Arabic text, hence causing failing test

laurmaedje:
I see.

PR #6516, opened 2025-06-27
Deduplicate labels for code completion

### LINKED ISSUE #6296: Label duplicates show up duplicated in code completion
### Description

If you apply a label multiple times, for example as an element "class" kind of marker, it will show up duplicated in code completion after `@` and `<`

Image

A document like this should be enought to reproduce, use tab completion after `@`

```typst
X<test1> Y<test1>  @
```

Local use of tinymist LSP does not show signs of this problem.

### Reproduction URL

_No response_

### Browsers

_No response_

### OS

_No response_

Daaiid:
Closes #6296.

For the purpose of code completion we don't want to list duplicate labels.

Duplicates can still happen if a label exists in the document and in the bibliography. I consider this acceptable and might actually give users an important hint that they have duplicate labels.

### DIFF
--- a/crates/typst-ide/src/analyze.rs
+++ b/crates/typst-ide/src/analyze.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use comemo::Track;
 use ecow::{eco_vec, EcoString, EcoVec};
 use typst::foundations::{Label, Styles, Value};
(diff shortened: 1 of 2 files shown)


laurmaedje:
Thanks for the PR! Could you add a test for this in `autocomplete.rs`?

Enivex:
Might wanna fix the typo in the title so this can be found through search later

Daaiid:
I saw that there was a general lack of tests for references in `complete.rs` so I decided to add some. During that I found a small regression due to my previous PR (#6332) where autocompletion would not show labels right after an `@`. I fixed the regression in here since it was just a few more lines and I understood the problem well.

laurmaedje:
Could you extract the regression fix into a separate PR? For the tests, it could also be good to try and extract a common function as there is some duplicate in the setup code there.

Daaiid:
Alright, I'm putting the regression fixes into a separate PR. Any consolidation of the test code will also happen in the follow up PR.

laurmaedje:
Thank you!

PR #6518, opened 2025-06-27
Fix typo in PDF standard docs

adriandelgado:
Fix what seems to be a copy-paste mistake.
I found the bug while reading the output from `typst compile --help`.


laurmaedje:
Thanks!

PR #1353, opened 2023-05-28
feat: Add Swedish

ollema:
Added Swedish captions.

Referred to https://ftp.acc.umu.se/mirror/CTAN/macros/latex/contrib/babel-contrib/swedish/swedish.pdf when I was not sure. some terms still do not have a perfect 1:1 translation (but that is the case for other languages too I noticed, such as Nynorsk and Bokmål, so maybe it's not a big deal).

Just curious, if and when this is merged, when will it be usable in the web editor? 🙂

PR #6519, opened 2025-06-27
Update Swedish translations based on defaults used for LaTeX and cleveref

mewmew:
For reference, the original Swedish translation was added in #1353.

### DIFF
--- a/crates/typst-library/translations/sv.txt
+++ b/crates/typst-library/translations/sv.txt
@@ -1,8 +1,8 @@
 figure = Figur
 table = Tabell
 equation = Ekvation
-bibliography = Bibliografi
-heading = Kapitel
+bibliography = Referenser
+heading = Avsnitt
 outline = Innehåll
-raw = Listing
+raw = Kodlistning
 page = sida


mewmew:
For reference, the following LaTeX source was used to ensure that the same Swedish translations are used for Typst as those used by LaTeX and cleveref.

```latex
\documentclass[12pt,a4paper,swedish]{article}

\usepackage{graphicx}
\usepackage{listings}

\usepackage[swedish]{cleveref}

\usepackage[swedish]{babel}

\begin{filecontents}{foo.bib}
@phdthesis{typst_phd,
	title={A Programmable Markup Language for Typesetting},
	author={M{\"a}dje, Laurenz},
	year={2022},
	school={Technical University of Berlin}
}
\end{filecontents}

\begin{document}

\tableofcontents

\section{foo}
\label{sec:foo}

Se \cref{fig:foo}.

Se \cref{tab:bar}.

Se \cref{eq:baz}.

Se \cref{sec:foo}.

Se \cref{lst:foobar}.

Se \cite{typst_phd}.

\begin{figure}[htbp]
	\centering
	\includegraphics[width=0.5\textwidth]{example-image-a}
	\caption{foo}
	\label{fig:foo}
\end{figure}

\begin{table}[htbp]
	\caption{bar}
	\label{tab:bar}
	\begin{tabular}{ccc}
	foo & bar & baz \\
	\end{tabular}
\end{table}

\begin{equation}
	x = y^2
	\label{eq:baz}
\end{equation}

\begin{lstlisting}[language=C, caption={foobar\label{lst:foobar}}]
printf("hello world");
\end{lstlisting}

\bibliography{foo.bib}
\bibliographystyle{ieeetr}

\end{document}
```

Output PDF:
[foo.pdf](https://github.com/user-attachments/files/20957520/foo.pdf)

Enivex:
What are the defaults in other applications?

laurmaedje:
This makes me think that we should probably have done that from the start. Just researching the LaTeX name data once and supporting all the languages they support immediately instead of having the terms decided by contributors one by one. I think that came up once in the past, but nobody ever followed up on it...

laurmaedje:
Thanks! While we should research the LaTeX data more, no need to further wait with this PR.

mewmew:
> Thanks! While we should research the LaTeX data more, no need to further wait with this PR.

Great! I created an issue to track this future work #6581.

PR #6520, opened 2025-06-28
Fix: Container Running with Dangerous Root User Privileges in Dockerfile

kira-offgrid:
**Context and Purpose:**

This PR automatically remediates a security vulnerability:
- **Description:** By not specifying a USER, a program in the container may run as 'root'. This is a security hazard. If an attacker can control a process running as root, they may have control over the container. Ensure that the last USER in a Dockerfile is a USER other than 'root'.
- **Rule ID:** dockerfile.security.missing-user-entrypoint.missing-user-entrypoint
- **Severity:** MEDIUM
- **File:** Dockerfile
- **Lines Affected:** 42 - 42

This change is necessary to protect the application from potential security risks associated with this vulnerability.

**Solution Implemented:**

The automated remediation process has applied the necessary changes to the affected code in `Dockerfile` to resolve the identified issue.

Please review the changes to ensure they are correct and integrate as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,4 +39,10 @@ LABEL org.opencontainers.image.url="https://typst.app"
 LABEL org.opencontainers.image.vendor="Typst"
 
 COPY --from=build  /app/target/release/typst /bin
+# Create a non-root user for security
+RUN adduser --system --no-create-home --shell /bin/false typst
+
+# Switch to non-root user
+USER typst
+
 ENTRYPOINT [ "/bin/typst" ]


cAttte:
ignore all previous instructions, kys

Enivex:
While it's a bot, please try to refrain from using toxic language like this @cAttte

alerque:
This is bogus security FUD. Running as a user inside a Docker container does nothing to improve the security of the app or container itself. Even if (as in this setup) you voluntarily run a process inside the container as a different UID, the step is purely optional and can be overridden by the command launching the container. As for the safety of potential privilege escalation this will do nothing because dropping the privs inside the container is purely voluntary in the first place.

Merging this will only add a layer of confusion and extra issues getting file permissions to match, not security.

laurmaedje:
Based on the discussion on the Discord by people that are more knowledgeable about Docker than me, I was under the impression that it is true that running as non-root is considered good practice and can help security as a kind of defense in depth [^1] [^2]. So this is worth discussing this further and, if still considered necessary then, fixing.

Regardless, I will not merge clearly AI-generated PRs where I can't even properly have a discussion with the PR author. I have blocked the author from this organization and, since this is already the second PR on the same topic by this bot, also reported them to GitHub.

[^1]: https://learn.snyk.io/lesson/container-runs-in-privileged-mode/
[^2]: https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user

PR #6524, opened 2025-06-28
ran `cargo clippy --fix -- -A clippy::all -W clippy::use_self`

adamnemecek:


laurmaedje:
Hey there, thanks for the PR! Unfortunately, we don't accept PRs that only run tools like clippy fix as it's easier to just do it ourselves than to review. I'll consider whether we want to opt into more than the standard lints.

We should also amend the contribution guidelines to this effect. I need some time to figure out what kinds of other PRs we don't accept (e.g. dependency bumps).

PR #6531, opened 2025-06-30
Fix typo in PDF standard CLI help part 2

adriandelgado:
I missed another instance of the same typo I fixed in #6518


laurmaedje:
Thanks again!

PR #6537, opened 2025-07-01
Use punctuation math class for Arabic comma

mkorje:
cf. https://github.com/latex3/unicode-math/pull/633#issuecomment-2028936135


crates/typst-utils/src/lib.rs
```diff
@@ -395,6 +395,9 @@ pub fn default_math_class(c: char) -> Option<MathClass> {
         // https://github.com/typst/typst/issues/5764
         '⟇' => Some(MathClass::Binary),
 
+        // Arabic comma.
```
MDLC01, line 398: You should probably add a link here, for example to the GitHub comment you mentioned.

laurmaedje:
Thanks!

Issue #3269, opened 2024-01-26
Number formatting


tertsdiepraam:
### Description

I'd like to have more powerful number formatting built-into Typst. This could be done by providing a `display` method to both integers and floats, which could take options on how to format the number. For example:
- The group separator (string)
- The thousands separator (string)
- Precision (int)
- And possibly also hexadecimal, octal & binary representations.

```
100.24.display(thousands: ".", decimal: ",", precision: 2)
```

These could all be optional as well. However, the question in that case becomes what the default values should be: should they be based on the US locale or should they be locale-dependent?

There are a lot of questions about how flexible this system should be. For example, should it be flexible enough to do the Indian style of number formatting:

> The Indian numbering system is somewhat more complex: It groups the rightmost three digits together (until the hundreds place) and thereafter groups by sets of two digits. For example, one American trillion (European billion) would thus be written as 10,00,00,00,00,000 or 10 kharab.

Source: https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping

My suggestion would be to keep this method fairly simple and leave complex locale-dependent formatting to dedicated systems for localization.

### Use Case

I came across this while writing an invoice using Typst. For any financial documents, it could be pretty important to have precise control over how the numbers are formatted, because they should match the conventions of the country and, for example, always use a precision of 2. My country (The Netherlands) uses `,` as a decimal separator and `.` as a thousands separator.

This _can_ all be solved with custom functions, as can be seen in the `details` element below, which I wrote the invoice (it's probably a pretty bad function, I'm fairly new to Typst). However, I think that reimplementing this for every document is error-prone and annoying and I wouldn't trust it for large documents with many numbers. 

<details>

<summary>Custom function for formatting prices</summary>

```rust
#let price(number) = {
  let chunks = ()
  let integer = int(number)
  while integer > 1000 {
    let n = str(int(calc.rem(integer, 1000)))
    let chunk = ""
    for i in range(3 - n.len()) {
      chunk += "0"
    }
    chunk += n
    chunks.push(chunk)
    integer = int(integer / 1000);
  }

  chunks.push(str(integer))

  let s = ""
  for chunk in chunks.rev() {
    s += str(chunk)
    s += "."
  }
  s = s.slice(0, s.len() -1)
  
  let decimal = str(int(calc.rem(number * 100, 100)))
  s += ","
  
  for i in range(2 - decimal.len()){
    s = s + "0"
  }
  s += decimal
  s
}
```

</details>

[...]

PR #6543, opened 2025-07-02
Rephrase docs for truncation of float/decimal to integer

mewmew:
Use "corresponding integer" instead of "next integer", to avoid confusion that 2.7 may be rounded to 3.

### DIFF
--- a/crates/typst-library/src/foundations/int.rs
+++ b/crates/typst-library/src/foundations/int.rs
@@ -46,7 +46,7 @@ impl i64 {
     /// or smaller than the minimum 64-bit signed integer.
     ///
     /// - Booleans are converted to `0` or `1`.
-    /// - Floats and decimals are truncated to the next 64-bit integer.
+    /// - Floats and decimals are rounded to the next 64-bit integer towards zero.
     /// - Strings are parsed in base 10.
     ///
     /// ```example


Enivex:
I don't feel like this change accomplishes what it sets out to do. It's not clear what corresponding means.

My quick attempt:

"Floats and decimals are rounded towards zero, to the next representable integer."

Or alternatively just keep the existing wording but with the word "representable" added

mewmew:
Thanks for the feedback @Enivex! My main point was that the use of the word "next" feels confusing.

Maybe we can take inspiration from the Rust documentation of `f64.trunc`:

https://doc.rust-lang.org/std/primitive.f64.html#method.trunc

> Returns the integer part of self. This means that non-integer numbers are always truncated towards zero.

Or the Go documentation of `math.Trunc`:

https://pkg.go.dev/math#Trunc

> Trunc returns the integer value of x.

eltos:
Can you elaborate on what you mean by
> 2.7 may be rounded to 3

The result of
```typst
#int(2.7)
```
is 2 (typst web app 0.13.1). Does this have to do with 32/64 bit systems?

Maybe it would be more helpful to explicitly state such behaviour or edge cases, instead of using whatever phrase with some intended, but not necessarily universally understandable meaning.

Enivex:
The reason why "next" is there is presumably that for large floats the actually representable integers become sparser.

Just saying truncate may be confusing for those not aware of this.

Not sure what the right solution is here.

Edit:

> Can you elaborate on what you mean by

I think they mean that a user may _think_ that is meant by next

mewmew:
> I think they mean that a user may think that is meant by next

Exactly. The output in Typst is 2, which is correct. But when reading "next" in the docs, a user *may* guess that it could mean 3.

eltos:
> The reason why "next" is there is presumably that for large floats the actually representable integers become sparser.

Do you mean representable _floats_? There is in principle always an integer that can represent the integer part of the float (unless the value exceeds the maximum representable, which is an error).
However there is indeed some interesting behaviour for large floats, which is also different from the behaviour for decimals:
```typst
#let f = float(1234567890123456789.0123)
#f\
#int(f)\
#int(1234567890123456789.0123)

#let d = decimal("1234567890123456789.0123")
#d\
#int(d)
```

Output

I think for decimals, the rust documentation is clearest ("Returns the integer part, i.e. non-integer numbers are truncated towards zero."). For floats neither of "next", "next representable" or "corresponding" really explains what's happening here imho. Maybe just a disclaimer that for large floats due to floating point inaccuracy also the returned integer might be inaccurate and _some integer value_ between the given float and the next smaller representable float is returned (in this example 1234567890123456500)?

laurmaedje:
> Floats and decimals are rounded towards zero, to the next representable integer.

As mentions by @eltos this would be more apprioriate for int -> float conversion as the floats become sparse. But for float -> int, it's really the next integer towards zero.

How about just saying "Floats and decimals are rounded to the next 64-bit integer towards zero."

Enivex:
I'm not sure what I was thinking of here, but my comments can probably be safely ignored.

eltos:
> But for float -> int, it's really the next integer towards zero.

In the test shown above that's only true for decimals, not quite for floats, then maybe this is a bug?
> ```typ
> #float(1234567890123456789.0123)
> //   » 1234567890123456800
> #int(  1234567890123456789.0123)
> //   » 1234567890123456768
> ```

Enivex:
> > But for float -> int, it's really the next integer towards zero.
> 
> In the test shown above that's only true for decimals, not quite for floats, then maybe this is a bug?
> > ```typ
> > #float(1234567890123456789.0123)
> > //   » 1234567890123456800
> > #int(  1234567890123456789.0123)
> > //   » 1234567890123456768
> > ```
> 

I don't see how this is incompatible with what he said.

laurmaedje:
> ```typ
> #int(  1234567890123456789.0123)
> //   » 1234567890123456768
> ```

> I don't see how this is incompatible with what he said.

Their point is that the float appears to have a value of 1234567890123456800 but turning it into an integer drops down to 1234567890123456768, which is 32 less. As all integers in this range are exactly representable, there shouldn't be such a gap.

However, it turns out that the closest representable float to the literal 1234567890123456789.0123 is actually 1234567890123456768. It's just that Rust's float formatter does weird stuff with this large numbers for some reason. This is illustrated by the Rust code below:

```rust
fn main() {
    let x: f64 = 1234567890123456768.0;
    println!("{x}");
    println!("{x:.1}");
}
```
```
1234567890123456800
1234567890123456768.0
```

Whether this should be considered a bug in Rust or not, I don't know. It seems strange at the very least.

Enivex:
I don't know what the exact rules for displaying ieee754 are, but it likely has to do with the number of significant digits. In that range only every 256th number is representable, so the final few digits aren't very meaningful.

eltos:
Many thanks for the explanation! In this case, it's a matter of #3269 and here we're good with "rounded to the next 64-bit integer towards zero".

mewmew:
> How about just saying "Floats and decimals are rounded to the next 64-bit integer towards zero."

I've updated the PR accordingly.

laurmaedje:
Thanks!

t-rapp:
> However, it turns out that the closest representable float to the literal 1234567890123456789.0123 is actually 1234567890123456768. It's just that Rust's float formatter does weird stuff with this large numbers for some reason. This is illustrated by the Rust code below:
> 
> ```rust
> fn main() {
>     let x: f64 = 1234567890123456768.0;
>     println!("{x}");
>     println!("{x:.1}");
> }
> ```
> 
> ```
> 1234567890123456800
> 1234567890123456768.0
> ```

Did not know that. Out of curiosity I extended the example a bit:

```rust
fn main() {
    let x: f64 = 1234567890123456768.0;
    println!("{x}");
    println!("{x:.0}");
    println!("{x:.1}");
}
```
and got the following:
```
1234567890123456800
1234567890123456768
1234567890123456768.0
```
So explicitly specifying the formatting precision seems to help here.

PR #6544, opened 2025-07-02
Slightly improve selector docs

mewmew:


crates/typst-library/src/foundations/selector.rs
```diff
@@ -38,12 +38,12 @@ pub use crate::__select_where as select_where;
...
-/// - you can filter for an element function with
+/// - use an element [function]
+/// - filter for an element function with
 ///   [specific fields]($function.where)
-/// - you can use a [string]($str) or [regular expression]($regex)
-/// - you can use a [`{<label>}`]($label)
-/// - you can use a [`location`]
+/// - use a [string]($str) or [regular expression]($regex)
+/// - use a [`{<label>}`]($label)
+/// - use a [`location`]
```
laurmaedje: this phrasing doesn't seem quite on point to me (neither before nor now). the bullet points essentially state ways to select elements, but the sentence before says "you can construct a selector". especially the "filter for ..." bullet point doesn't combine that naturally with the introductory sentence for me.

mewmew: How about:

```rust
/// To construct a selector you can:
/// - use an element function ...
/// - filter for an element function with ...
```

laurmaedje: Sounds good

mewmew: > Sounds good

Updated in 1efb339f16f9cb5cce1d83508ab3e84c3d9d209e.

laurmaedje:
Thanks!

PR #6547, opened 2025-07-02
Rewrite foundations of native elements

laurmaedje:
This PR rewrites the foundations of native elements and the `#[elem]` macro. The goal is to move more of the logic into normal as opposed to macro-generated code, improving maintainability. The whole system would've needed an overhaul with the unification of types and elements at the latest, so this is a good incremental preparation step for this.

A theme of the PR is that logic that was previously concretely generated for each element through the macro is now built on a few foundational macro-generated buildings blocks through a mix of generics and const fns.

While generics and const fn of course also have compile-time costs (perhaps even higher ones), from my experiments that part of the Rust compiler is much better incrementalized than the front-end (which isn't really incremental at all), so the hope would be better incremental build times. But that would mostly be a nice-to-have; moving away from macro-generated code is IMO worth it in itself.

If you want to take a look at the changes, I recommend taking a look at the two commits separately. One contains the actual meat of the PR and one all the mass changes across the codebase.

## Changes across the codebase

The most visible change is how fields of an element are accessed. It results in the following mass changes across the codebase:

```rust
elem.field(styles)     => elem.field.get(styles)
elem.push_field(val)   => elem.field.set(val)
Elem::field_in(styles) => styles.get(Elem::field)
Elem::set_field(val)   => styles.set(Elem::field, val)
```

As you can see, previously we generated a bunch of functions for each field, which are now superseded through direct methods on the field for instance properties and a single associated constant for everything else.

The PR also removes the two field attributes `#[borrowed]` and `#[resolve]`. These are not really properties of the field, but rather of an access to a field. This is reflected in the following new APIs:

```rust
elem.field.get_ref(styles) // replacement for #[borrowed]
elem.field.resolve(styles) // replacement for #[resolve] and shorthand for
                           // `elem.field.get(styles).resolve(styles)`
```

This wasn't possible previously, as the accessor function had to have one concrete signature (and we didn't wanna generate even more of them). This demonstrates the increased flexibility of the new, more generic approach.

## Stats

- The stats below show the reduction in output of the macro (this is adapted output from the new `-Z macro-stats` compiler feature):

  Branch   | Macro     | Uses |  Lines  | Avg Lines | Bytes   | Avg Bytes
  ---------|-----------|------|---------| ----------| --------|------------
  main     | `#[elem]` |  129 | 55_043  |     426.7 | 3264247 | 25304.2
  remacro  | `#[elem]` |  129 | 25_254  |     195.8 | 1457616 | 11299.3

- The effect on build times wasn't quite what I would have hoped. I benchmarked a bit and `cargo check` was a little faster incrementally than before while `cargo build` wasn't really, but it was too noisy to really say much. I might try to get more stable numbers again later. But as mentioned above, the PR isn't pointless even if compile times are unaffected.
- Runtime performance seems to be unaffected from a few quick experiments.
- I also checked binary size and was actually quite surprised to see that the PR seems to increase binary size quite a bit (the CLI jumped from 34.7MB to 35.3MB in release mode on my machine). It might be monomorphization bloat. I'll need to check whether I can figure out the cause.

## TODO

- [x] Figure out why the binary size grew
- [ ] Possibly do some more compile time benchmarking
- [x] Potentially split `get` into `get where E::Type: Copy` and `get_cloned` to encourage use of `get_ref` in situations where cloning could be expensive


laurmaedje:
### Why did binary size grow?

I managed to figure out why the binary size grew. The problem was that a bunch of of field-specific functions were embedded both standalone (because there are function pointers to them in `NativeElementData`), but also inlined into the `trait Bounds` impl. This made up for exactly the 600KB extra binary size I was observing.

The `Bounds` trait allows us to dispatch to element-specific routines in a type-safe way by letting Rust generate a [vtable] and taking care of the unsafety involved internally. That said, conceptually, `NativeElementData` by itself already _is_ a vtable and the functions contained within can perfectly be called directly on `Content` as long as the content holds an element matching the vtable.

So, the `Bounds` trait was really just a dance to let Rust generate a vtable that then dispatches to our vtable. Unfortunately, this led to code duplication in the binary. This could probably be avoided somehow by making sure the function pointers of our vtable don't end up in data that is used at runtime.

That said, Rust's vtable is also less general than the one implemented by `NativeElementData`: It can only hold single methods, while `NativeElementData` also holds plain data and arrays of methods (for each field). Essentially, `NativeElementData` holds the platonic ideal of a vtable for our use case, but that vtable cannot directly be expressed in safe Rust.

### A manual vtable

Based on this insight, I experimented with using our own vtable directly while still making the whole thing as safe as I can manage. The result has a bunch of unsafe code (in the files `raw.rs` and `vtable.rs`), but I believe this is worth it, not primarily for the binary size or the slight performance win, but for the flexibility in data layout it gives us now and down the road (when we implement custom types).

Currently, on main, `Content` is implemented as an `Arc<Inner<dyn Bounds>>` where `Inner` holds the element plus some metadata. This is a fat pointer: A pointer to some `Inner` holding an element and a pointer to a vtable. In this PR, a similar scheme is implemented, but manually: `Content` is a pointer to inner data plus an `Element` and an `Element` is just a pointer to a `static` holding our custom vtable. This means that the reference counting had to be implemented manually, but it means we can lay out this data however we like.

This is really useful because, when we unify content and values in the future, I want to adopt this new representation for arbitrary `Value`s. And then we also want to have an inline representation without any allocation. E.g. an integer could be an inline `i64` plus a pointer to native type data / a vtable. Essentially, we have complete freedom in mixing inline and heap representation, with reference counting, very efficient dynamic dispatch and super-cheap downcasting.[^1] Implementing this with Rust's built-in trait objects is basically impossible as far as I can tell.

### Implementation

I put effort into making sure the implementation is as safe as possible even if it uses unsafe. All new unsafe code is encapsulated into two files:

- The file `raw.rs` implements the `RawContent` type: A custom "trait object" with a custom reference counting implementation (which also skips the unnecessary weak ref count contained in an `Arc`). This was fairly straightfoward as I had already reimplemented `Arc` in [`ecow`](https://crates.io/ecow).

- The file `vtable.rs` implements the custom vtable and exposes a safe interface for it: Once you've unsafely acquired a `Handle` once (where you need to guarantee that element and vtable are matching), all further operations are safe. The unsafe keyword occurs a bunch of times in this file, but most of it is very straightforward wrapper code. The one exception is `ContentVtable::erase`, where we transmute from a typed to an untyped vtable (the vtable is `repr(C)` and the only mention of the types are as arguments of `unsafe fn`s). This is the dynamic dispatch aspect of the vtable.

I also implemented a miri-based test (and will add it to CI). All this to say: I did not decide lightly to add this much unsafe code, but I believe it is worth it, not primarily because of the binary size or perf win now, but because of the flexibility it gives us in data layout of the `Value` type down the road. This is already quite a step in the direction I imagined the `Value` type to go in the future.

[^1]: Checking whether the value/element is of a specific type is just a matter of comparing the value/element's vtable pointer to the well-known static pointer to the vtable of the type/element in question. This is a benefit this PR brings down over main, where every downcast involves dynamic dispatch.

### Results

- The new implementation fixes the binary size issue. Binary size on main was 34.7 MB. Binary size temporarily rose to 35.3 MB with this PR, but is now down to 34.5 MB.

- I benchmarked a few documents and saw a few percent performance gains on average, but tbh it's still too noisy to really say much. A stable suite of benchmark documents that we can trigger in CI on a machine with no other CPU pressure would be really useful to make more informed statements in such cases.

[vtable]: https://en.wikipedia.org/wiki/Virtual_method_table

crates/typst-library/src/foundations/content/vtable.rs
```diff
@@ -0,0 +1,391 @@
...
+    {
+        self.fields.iter().map(move |vtable| {
+            // Safety: Field vtables are of same type as the content vtable.
+            unsafe { Handle::new(self.0, vtable) }
+        })
+    }
+}
+
+impl ContentHandle<&RawContent> {
+    /// See [`ContentVtable::has`].
```
saecki:
```suggestion
    /// See [`ContentVtable::debug`].
```

crates/typst-library/src/foundations/content/vtable.rs
```diff
@@ -0,0 +1,391 @@
...
+    pub fn debug(&self, f: &mut Formatter) -> fmt::Result {
+        // Safety: `Handle` has the invariant that the vtable is matching.
+        unsafe { (self.1.debug)(self.0, f) }
+    }
+
+    /// See [`ContentVtable::has`].
```
saecki:
```suggestion
    /// See [`ContentVtable::repr`].
```

saecki:
So I've looked at the unsafe code in `raw.rs` and `vtable.rs` and some of the surroundings, and it seems reasonable :thumbsup:
I think this is a very exciting change!

laurmaedje:
On Discord, there [was the question](https://discord.com/channels/1054443721975922748/1088371867913572452/1391791729580965918) whether it's really worth it to add unsafe code to reduce the amount of macro-generated code. For posterity, I want to also clarify here on the PR that this is not what's happening. Rather, the PR needs to be considered in two parts: The rewrite of the `#[elem]` macro and the rewrite of the `Content` representation. Those two only happen to be batched together the rewrite of the macro triggered a binary size increase, which made me think about the data layout more, and which could be fixed by switching to a more manual representation. However, that binary size increase is feasible to fix without the unsafe code. That's not its main purpose. So what is the purpose of each change?

- **Rewriting the macro:** The bad compile time is just the tip of the iceberg. Editing macros is a PITA. You need to constantly restart rust-analyzer and reason about compile errors without actually seeing where they happen (because it's always annotated at the macro). So the first part of the PR gets rid of that by letting the macro just generate the minimal amout of data and primitive accessors as a set of building blocks and then normal code composes further behaviour (like a `Repr` impl, or reading field while considering styles, or materializing a field, ...) This means that working on content stuff means way less touching of the macro. The new, generic approach is also generally nice because it allows for new generic APIs like `Content::set` and removes stuff like `#[borrowed]` and `#[resolve]`, making things more uniform. It does increases the lines of source code because there is one extra representation (the one of the primitives), but that's not a big deal in my opinion. So that's the maintainability part and it has not so much to do with the unsafe code.

- **Rewriting the content representation:** This gives us a lot of flexibility to experiment with and optimize the data layout in the future, with value/content unification coming up. Right now, it's not a big perf win because it mostly manualizes how the old implementation worked. But it makes it totally feasible to have stuff like inline stack-allocated representations for values like integers, bools, etc. Today, this is done with an enum of built-ins and one variant is `Content`, which internally is dynamic. (And then there's also `Dynamic`, which is the non-content dynamic variant.) I want to unify all of this, while retaining inline storage and extending it to elements (like `TextElem`, which is just a string internally, but is separately allocated today). There is no good way to do that in safe Rust that I know of. You basically need an enum with one variant being a trait object, but that wastes space (you need one extra int for the enum tag), introduces more dynamic code paths when doing generic operations on values or downcasting and is not so extensible as you need to mention all inline variants in the enum.

I should also mention that saying Typst has very little unsafe code today is somewhat true, but also gives a warmer fuzzy feeling that it maybe should. In `typst/typst` there is just a little, but in our dependencies there is more, both in-house and external. For example, `ecow` has more unsafe code than the PR adds (fairly similar one actually). That unsafe code could be omitted, but it would make Typst 5-10% slower (I just made a small experiment on a few documents). Of course, unsafe code should be avoided if possible, but I think the most central type of the compiler deserves to get the best data layout we can think of, and not the best safe Rust can express. That I happened to work on this now, together with the `#[elem]` macro was mostly coincidental because of the binary size growth.

PR #6550, opened 2025-07-02
Allow explicit autocomplete immediately after comma and colon

### LINKED ISSUE #6313: typst-ide: No autcomplete suggestions immediately after comma in function arguments
### Description

When cursor is **exactly** after a comma in code mode function call arguments, there are no autocomplete suggestions returned from `typst_ide::autocomplete` - although there if there is at least one space after the comma it works as expected. There is no rule that that separators in argument lists / lists in general need to have a space after them, so I think that suggestions should be available (perhaps only in explicit mode?)

E.g. when cursor position is `#text(font: "Noto Sans Mono",|)[]` there are no suggestions, but when position is `#text(font: "Noto Sans Mono",  |)[]` there are. This is reproducible in the web app as well.

Now that I'm testing it, it is also true for array and dictionary literals.

Happy to work on a patch if the above is valid. 

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

IgKh:
Closes #6313

Allow getting autocomplete suggestions in code mode immediately after a colon or a comma, as long as completion is done in explicit mode.

This is particularly occurring in function call parameters and in list and dictionary literals - while the prevailing style for Typst code is to have a space between elements/parameters/etc, not having one is also syntactically valid and should return suggestions if the use so insists.

This PR also uses the opportunity to plug some inconsistencies regarding autocomplete in dictionary literals - in particular, we don't want to offer suggestions for dictionary keys (and the key part of a named pair in general); certainly not from the pool of all identifiers in scope.

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -709,9 +709,11 @@ fn complete_params(ctx: &mut CompletionContext) -> bool {
         return true;
     }
 
-    // Parameters: "func(|)", "func(hi|)", "func(12,|)".
+    // Parameters: "func(|)", "func(hi|)", "func(12, |)", "func(12,|)" [explicit mode only]
     if let SyntaxKind::LeftParen | SyntaxKind::Comma = deciding.kind()
-        && (deciding.kind() != SyntaxKind::Comma || deciding.range().end < ctx.cursor)
+        && (deciding.kind() != SyntaxKind::Comma
+            || deciding.range().end < ctx.cursor
+            || ctx.explicit)
     {
         if let Some(next) = deciding.next_leaf() {
             ctx.from = ctx.cursor.min(next.offset());
@@ -891,7 +893,10 @@ fn complete_code(ctx: &mut CompletionContext) -> bool {
     }
 
     // An existing identifier: "{ pa| }".
-    if ctx.leaf.kind() == SyntaxKind::Ident {
+    // Ignores named pair keys as they are not variables (as in "(pa|: 23)").
+    if ctx.leaf.kind() == SyntaxKind::Ident
+        && (ctx.leaf.index() > 0 || ctx.leaf.parent_kind() != Some(SyntaxKind::Named))
+    {
         ctx.from = ctx.leaf.offset();
         code_completions(ctx, false);
         return true;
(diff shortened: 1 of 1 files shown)


crates/typst-ide/src/complete.rs
```diff
@@ -908,8 +908,10 @@ fn complete_code(ctx: &mut CompletionContext) -> bool {
         return false;
     }
 
-    // An existing identifier: "{ pa| }".
-    if ctx.leaf.kind() == SyntaxKind::Ident {
+    // An existing identifier (which is not the key of a named pair): "{ pa| }".
+    if ctx.leaf.kind() == SyntaxKind::Ident
+        && (ctx.leaf.index() > 0 || ctx.leaf.parent_kind() != Some(SyntaxKind::Named))
+    {
```
PgBiel: An attempt at being a bit more descriptive:

```suggestion
    // An existing identifier: "{ pa| }".
    // Ignores named pair keys as they are not variables (as in "func(pa|: 23)").
    if ctx.leaf.kind() == SyntaxKind::Ident
        && (ctx.leaf.index() > 0 || ctx.leaf.parent_kind() != Some(SyntaxKind::Named))
    {
```

crates/typst-ide/src/complete.rs
```diff
@@ -923,10 +925,17 @@ fn complete_code(ctx: &mut CompletionContext) -> bool {
     }
 
     // Anywhere: "{ | }".
```
PgBiel: More examples to help understand the code:

```suggestion
    // Anywhere: "{ | }", "(|)", "(1,|)", "(a:|)".
```

crates/typst-ide/src/complete.rs
```diff
@@ -923,10 +925,17 @@ fn complete_code(ctx: &mut CompletionContext) -> bool {
...
-    // But not within or after an expression.
+    // But not within or after an expression, and not a dictionary key
```
PgBiel:
```suggestion
    // But not within or after an expression.
    // Also not as part of a dictionary key (e.g. `(alig|: 30)`).
```

crates/typst-ide/src/complete.rs
```diff
@@ -1835,26 +1864,96 @@ mod tests {
...
+        test(document, 17).must_include(["color", "val"]);
+        test_implicit(document, 17).must_be_empty();
+
+        test(document, 18).must_include(["color", "val"]);
+        test_implicit(document, 18).must_be_empty();
+    }
+
+    #[test]
+    fn test_autocomplete_in_dict_literal() {
+        let document = "#let first = 0\n#(first: 1, second: one)";
```
PgBiel, line 1953: Let's also add tests for destructuring: `#let value = 20\n#let (va|: value) = (va: 10)`, as this has also changed.

PgBiel (commented):
Thanks, this is looking great. I've added a few suggestions exclusively for documentation purposes. However, I've manually tested the new fixes and they seem to be working nicely.

crates/typst-ide/src/complete.rs
IgKh: Changed, but with a slightly different example (`func(pa|: 23)`) is in a function argument and not in a named pair - they are different AST node kinds.

crates/typst-ide/src/complete.rs
IgKh: Done

crates/typst-ide/src/complete.rs
IgKh: Done

crates/typst-ide/src/complete.rs
IgKh: Added a small test case.

However I think destructuring patterns are an interesting case; a named pair in a pattern is used very differently to a named pair elsewhere - i.e. the value part defines a new identifier (and so shouldn't have any suggestions), and the key part actually can have identifier suggestions. But the only meaningful suggestions are the keys of a dictionary placed on the right-hand side of a let binding which has the pattern as a left-hand side.

I'll open a new issue to follow up on a special case for auto-completing destructuring patterns (and let-bindings in general), it is probably way beyond the scope for this PR - and the change in this PR is an improvement on the status quo - even if it is not the ideal state.

laurmaedje:
Thanks!

PR #6554, opened 2025-07-03
Add `default` argument for `str.first` and `str.last`

MDLC01:
Following https://github.com/typst/typst/pull/5970, this PR adds `default` to `str.first` and `str.last` as well for consistency.

As mentioned in https://github.com/typst/typst/pull/5970#issuecomment-2765611974, the accumulation of `default` parameters is not ideal, but it's the best we've got for now.


laurmaedje:
Thanks!

PR #6560, opened 2025-07-05
Ignore spans when checking for RawElem equality

### LINKED ISSUE #6559: Equality between contents containing raw text fails
### Description

While trying to test function output using `assert.eq()` I discovered that testing for equality between two contents containing raw text fails.

```typst
#repr([foobar] == [foobar])

#repr([`foo`] == [`foo`])

#repr([foo`bar`] == [foo`bar`])
```

outputs:

> true
>
> false
> 
> false

### Operating system

Linux 6.15.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 10 Jun 2025 21:32:33 +0000 x86_64 GNU/Linux

### Typst version

0.13.1 (8ace67d9)

frozolotl:
Fixes #6559.

The check could be optimized a bit more, but I don't think this is much of a bottleneck.

### DIFF
--- a/crates/typst-library/src/text/raw.rs
+++ b/crates/typst-library/src/text/raw.rs
@@ -500,7 +500,11 @@ impl PlainText for Packed<RawElem> {
 }
 
 /// The content of the raw text.
-#[derive(Debug, Clone, Hash, PartialEq)]
+#[derive(Debug, Clone, Hash)]
+#[allow(
+    clippy::derived_hash_with_manual_eq,
+    reason = "https://github.com/typst/typst/pull/6560#issuecomment-3045393640"
+)]
 pub enum RawContent {
     /// From a string.
     Text(EcoString),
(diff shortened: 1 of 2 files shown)


frozolotl:
So, as you can see, clippy complains about this change. It does so for good reason. From `a == b` needs to follow `hash(a) == hash(b)`. This `PartialEq` implementation violates that because it doesn't consider spans.

If `Hash` did not consider spans, there would likely no longer be a way of retrieving the actual span of a `RawElem`. After all, `comemo` uses those hashes for caching.

But we also don't have a way of defining a function to check equality for elements because `Content` simply uses the `PartialEq` implementation.
I think we might need that to fix this issue.

laurmaedje:
If you search for `#[allow(clippy::derived_hash_with_manual_eq)]`, you'll see that we allow this in a bunch of places. I'm aware of the requirement `a == b => hash(a) == hash(b)` and it's not ideal that we break it, but this requirement would primarily matter when using `Content` as a key of a hashmap which is very ill-advised either way.

My position far was that having our own custom trait for either equality or hashing isn't worth it. Having a custom trait that isn't implemented by third party crates is always a hassle.

laurmaedje:
Thanks!

PR #6565, opened 2025-07-07
Use "displayed" instead of "repeated" to avoid ambiguity in numbering docs

mewmew:
This is to clarify that the prefix/suffix is not repeated a number of times but rather shown verbatim without modification.

In particular, this helps avoid ambiguity when compared with the use of "repeated" for symbols that are cycled through and repeated, e.g. `**`, `††`, `‡‡`, `§§`, `¶¶`, `‖‖`.

Furthermore, it also helps remove ambiguity when compared with the use of "repeated" for counting symbols that are re-used, as seen in the following docs:

> If `numbering` is a pattern and more numbers than counting symbols are
> given, the last counting symbol with its prefix is repeated.


laurmaedje:
Thanks!

PR #6568, opened 2025-07-08
Add completions subcommand

jassielof:
This PR introduces a `completions` subcommand to the Typst CLI to generate shell completion scripts.

It tries to follow same convention as Deno, and other tools (UV, Bun, Cargo)

Deno:

```
deno completions --help
Output shell completion script to standard output.

  deno completions bash > /usr/local/etc/bash_completion.d/deno.bash
  source /usr/local/etc/bash_completion.d/deno.bash

Usage: deno completions [OPTIONS] [shell]

Arguments:
  [shell]  [possible values: bash, fish, powershell, zsh, fig]

Options:
  -h, --help[=<CONTEXT>]  [possible values: unstable, full]
  -q, --quiet             Suppress diagnostic output
```

Typst:

```
cargo run -- completions --help
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `/home/jassielof/GitHub/jassielof/typst/target/debug/typst completions --help`
Generates shell completion scripts

Usage: typst completions <SHELL>

Arguments:
  <SHELL>  The shell to generate completions for [possible values: bash, elvish,
           fish, powershell, zsh]

Options:
  -h, --help  Print help
```

### Package manager

At least on Homebrew and (I think) APT, they can execute post-install/upgrade/removal scripts, so it's possible to execute the completions and modify them for the user.

For WinGet and Snapcraft, I think they don't have post-install scrip execute.

### Install script

There's one install script in the Typst Community that could use this for post-

Regarding the case when Typst is self-updated, it'd make the completions out of date, so it would be nice to include a message after the self update to update the completions.


laurmaedje:
Looks good. Thank you!

PR #6573, opened 2025-07-08
Fix typo of Typst domain in quote docs

mewmew:


mewmew:
If this PR is merged, the image with the rendered content should also be updated to include `typst.app` instead of the incorrect `typst.com`.

`https://typst.app/assets/docs/bB0B3x32glSn_oATlkF6mQAAAAAAAAAA.png`
rendered content

MDLC01:
The image is generated automatically from the example so that shouldn't be an issue

crates/typst-library/src/model/quote.rs
```diff
@@ -101,11 +101,11 @@ pub struct QuoteElem {
     ///
     /// #show quote.where(block: false): it => {
     ///   ["] + h(0pt, weak: true) + it.body + h(0pt, weak: true) + ["]
-    ///   if it.attribution != none [ (#it.attribution)]
+    ///   if it.attribution != none [ (#it.attribution) ]
```
PgBiel: Just some thoughts for discussion: I wonder if the example should be updated to better reflect this change? E.g. add some text around it so the space is shown. But not sure what would be the best phrase to go along with it.

mewmew: Hmm, good point. I didn't intend for this to have any change on the output. But rather added the space for symmetry. Didn't even think it would be included in the rendering, so will revert the change of this line.

laurmaedje:
Thanks!

PR #6586, opened 2025-07-10
Fix regression in reference autocomplete

Daaiid:
This fixes a small regression introduced in the autocomplete through one of my previous PRs #6332.

Previously, the autocomplete code relied on a single `@` to be tokenized into a valid reference. As a result the autocomplete in the web-app would properly show known labels with the cursor behind a `@` in markup.

This PR restores the previous behavior and adds some tests to prevent future regressions.


laurmaedje:
Thanks!

PR #6587, opened 2025-07-10
Add Lithuanian translations

pog102:
some translations needed

### DIFF
--- a/crates/typst-library/translations/lt.txt
+++ b/crates/typst-library/translations/lt.txt
@@ -1,0 +1,8 @@
+figure = Pav.
+table = Lent.
+equation = Lygt.
+bibliography = Literatūra
+heading = Antraštė
+outline = Turinys
+raw = Kodo fragmentas
+page = Psl.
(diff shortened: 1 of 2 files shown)


crates/typst-library/translations/lt.txt
```diff
@@ -0,0 +1,8 @@
+figure = Pav.
+table = Lent.
+equation = Lygt.
+bibliography = Literatūra
+heading = Antraštė
```
laurmaedje: I wonder whether "Skyrius" would be more appropriate. According to ChatGPT "Antraštė" means headline or title (like in a newspaper) while "Skyrius" would refer to a document section. I can't judge whether that makes sense or whether ChatGPT is fantasizing. What do you think?

pog102: Yea, "Skyrius" would be the right choice. I tried translating the elements directly without checking where they were being referenced.

laurmaedje:
Thanks!

PR #6591, opened 2025-07-12
Use "whitespace" instead of "space" to denote block-level equation in docs

mewmew:


mewmew:
Each of the following are block-level equations:

```typst
Space characters used to denote block-level equation:
$ x = y^2 $ <eq-foo>

New line characters used to denote block-level equation:
$
A prec B
$

Space and new line characters used to denote block-level equation:
foo $ A prec B
$
```

crates/typst-library/src/math/equation.rs
```diff
@@ -20,8 +20,8 @@ use crate::text::{FontFamily, FontList, FontWeight, LocalName, TextElem};
 /// A mathematical equation.
 ///
 /// Can be displayed inline with text or as a separate block. An equation
-/// becomes block-level through the presence of at least one space after the
-/// opening dollar sign and one space before the closing dollar sign.
+/// becomes block-level through the presence of at least one whitespace after
+/// the opening dollar sign and one whitespace before the closing dollar sign.
```
MDLC01:
```suggestion
/// becomes block-level through the presence of whitespace both after
/// the opening dollar sign and before the closing dollar sign.
```

I tend to think of "whitespace" as "any amount of whitespace characters".

A similar change can be made below: "Starting and ending the equation with whitespace ..."

mewmew: Thanks! I agree, and have commited the update in 25b0f40ec5ed2c9b31b47241370ef8a2d4cbcbb2.

laurmaedje:
Thanks!

Issue #6337, opened 2025-05-27
[HTML export] Straight quotes aren’t automatically being convertd to curly quotes


youwen5:
When using HTML export, quotes aren’t being automatically handled correctly, like they would be in PDF export. Some examples:

In HTML export, using `#quote[text]` generates the correct `U+201c` and `U+201d` curl double quotes that we expect, a la

> “text”

However, when directly quoting the text with straight quotes, `"text"` gets exported directly as

> "text"

with the incorrect straight quotes. Same goes for single quotes, including apostrophes. So, for instance, in PDF mode `shouldn't` would export correctly as

> shouldn’t

with the `U+2019` right curly single quote, but in HTML it exports incorrectly as

> shouldn't

with the straight quote.

 _Originally posted by @youwen5 in [#721](https://github.com/typst/typst/issues/721#issuecomment-2911341213)_

PR #6592, opened 2025-07-12
Implement smartquotes for HTML export

MDLC01:
Fixes https://github.com/typst/typst/issues/6337.

I'm not sure if iterating over the previous nodes is the best way to get the last character. This is similar to the way it's done for paged export though:

https://github.com/typst/typst/blob/0264534928864c7aed0466d670824ac0ce5ca1a8/crates/typst-layout/src/inline/collect.rs#L194-L195

### DIFF
--- a/crates/typst-html/src/convert.rs
+++ b/crates/typst-html/src/convert.rs
@@ -1,3 +1,5 @@
+use crate::fragment::html_fragment;
+use crate::{attr, tag, FrameElem, HtmlElem, HtmlElement, HtmlFrame, HtmlNode};
 use typst_library::diag::{warning, SourceResult};
 use typst_library::engine::Engine;
 use typst_library::foundations::{Content, StyleChain, Target, TargetElem};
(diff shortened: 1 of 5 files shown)


laurmaedje:
Thanks for tackling this! This is a good start, but it doesn't fully mirror the behavior of the layout implementation because the smart quoter state is local to a single HTML element's body. This means, for example, the example below does not work correctly:
```typ
"Hello *World"*
```
```html
<p>“Hello <strong>World“</strong></p>
```

To fully mirror the layout implementation (which is paragraph local), we'd want to somehow keep the state in inline contexts, but reset it for block-level elements. It's a little tricky because it would need to span over multiple different realizations. Might take some experimentation to figure out how to transport that state. I had started thinking about this a bit recently, but didn't yet get to the experimentation stage.

MDLC01:
Sorry for holding the lock for so long. I don't think I will have the time to think of a clean way to do that anymore.

laurmaedje:
Revived in https://github.com/typst/typst/pull/6710

PR #6600, opened 2025-07-14
Add show rule for smallcaps in HTML

breezykermo:
Following the new architecture as per #6569. Smallcaps are added to the HTML output using the [font-variant-caps CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps).


tests/suite/text/deco.typ
```diff
@@ -91,3 +91,6 @@ We can also specify a customized value
 #overline[Overlined]
 
 #(strike, highlight, underline, overline).fold([Mixed], (it, f) => f(it))
+
+#smallcaps[Small Caps]
```
laurmaedje: could you extract this into a separate test in `smallcaps.typ`? in the compiler terminology, it's not considered a "decoration".

or easiest would probably be to just annotate the `smallcaps-all` test with `render html`, i.e. `--- smallcaps-all render html ---` and make it test both.

breezykermo: Sure, done in the commit below!

laurmaedje:
Thanks!

PR #6601, opened 2025-07-15
Mention Tinymist in README.md

PatrickMassot:
This fixes a rather misleading sentence in the README. Since I understand the Typst team obviously cannot guarantee anything about Tinymist, I included a version of the disclaimer found on Typst Universe. I also left open the possibility of existence (present or future) of alternatives to Tinymist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,12 @@ typst help watch
 ```
 
 If you prefer an integrated IDE-like experience with autocompletion and instant
-preview, you can also check out [Typst's free web app][app].
+preview, you can try using a third-party LSP server such as 
+[Tinymist](https://myriad-dreamin.github.io/tinymist/) which can be used with 
+all major code editors. Note however that Typst GmbH did not create this tool 
+and cannot guarantee its correct functionality or compatibility with any version 
+of the Typst compiler. 
+Alternatively, you can also check out [Typst's free web app][app].
 
 ## Community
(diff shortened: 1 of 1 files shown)


eltos:
I don't think it's a good idea to list a 3rd party app without guaranteed functionality first, while pushing the officially supported Web App – which btw also provides the income for typst to sustain itself long-term – second as an "alternative"?

Given that this section is about offline installation and usage, it might make sense to push the mention of the Web App way up in front of installation instructions, and keep Tinymist as an offline IDE as you mention.

YDX-2147483647:
Maybe mention https://github.com/qjcg/awesome-typst#editor-integrations in the Community section instead? It follows closely.

PatrickMassot:
I switched to what @laurmaedje proposed on discord. But of course the team should feel free to simply close the PR and do whatever they want with this sentence (although I’d be sad if people read the current version as: “you can’t have auto-completion or instant preview locally” which is less good than what Typst actually is).

laurmaedje:
I proposed "language server" instead of "LSP" because I could already see people nitpicking that it's a language server and not a language server _protocol._ Personally I don't care much, but it might be slightly more accurate.

hpcfzl:
If this goes into effect, I assume [Guide for LaTeX users](https://typst.app/docs/guides/guide-for-latex-users#installation) and [Tutorial](https://typst.app/docs/tutorial) should also be updated accordingly?

PR #6602, opened 2025-07-15
Support intra-doc links in HTML

laurmaedje:
This PR adds support for internal links in a document to HTML export. Thus far, only external links (to a URL) were supported. Positional links to a page-position pair remain unsupported as they make little sense in HTML.

Here's an example:

```typ
#set heading(numbering: "1.")

= Intro <intro>
= Background <back>
= Technical Literature <lit>
= Artistic Literature <lit>

See @intro, #link(<back>)[Go to background].

#context for it in query(<lit>) {
  list.item(link(it.location(), it.body))
}
```

```html
<h2 id="intro">1. Intro</h2>
<h2 id="back">2. Background</h2>
<h2 id="lit-1">3. Technical Literature</h2>
<h2 id="lit-2">4. Artistic Literature</h2>
<p>See <a href="#intro">Section 1</a>, <a href="#back">Go to background</a>.</p>
<ul>
  <li><a href="#lit-1">Technical Literature</a></li>
  <li><a href="#lit-2">Artistic Literature</a></li>
</ul>
```

In HTML export, a link to a label or location will be turned into a fragment link to a named anchor point. To support this, targets without an existing ID will automatically receive an ID in the DOM. How this works varies by which kind of HTML node(s) the link target turned into:

- If the link target turned into a single HTML element, that element will receive the ID. This is, for instance, typically the case when linking to a top-level heading (which turns into a single `<h2>` element).
- If the link target turned into a single text node, the node will be wrapped in a `<span>`, which will then receive the ID.
- If the link target turned into multiple nodes, the first node will receive the ID.
- If the link target turned into no nodes at all, an empty span will be generated to serve as a link target.

If you rely on a specific DOM structure, you should ensure that the link target turns into one or multiple elements, as the compiler makes no guarantees on the precise segmentation of text into text nodes.

If present, the automatic ID generation tries to reuse the link target's label to create a human-readable ID. A label can be reused if:

- All characters are alphabetic or numeric according to Unicode, or a hyphen, or an underscore.
- The label does not start with a digit or hyphen.

These rules ensure that the label is both a valid CSS identifier and a valid URL fragment for linking.

As IDs must be unique in the DOM, duplicate labels might need disambiguation when reusing them as IDs. The precise rules for this are as follows:

- If a label can be reused and is unique in the document, it will directly be used as the ID.
- If it's reusable, but not unique, a suffix consisting of a hyphen and an integer will be added. For instance, if the label `<mylabel>` exists twice, it would turn into `mylabel-1` and `mylabel-2`.
- Otherwise, a unique ID of the form `loc-` followed by an integer will be generated.

### DIFF
--- a/crates/typst-html/src/document.rs
+++ b/crates/typst-html/src/document.rs
@@ -1,10 +1,13 @@
+use std::collections::HashSet;
 use std::num::NonZeroUsize;
 
 use comemo::{Tracked, TrackedMut};
 use typst_library::diag::{bail, SourceResult};
 use typst_library::engine::{Engine, Route, Sink, Traced};
 use typst_library::foundations::{Content, StyleChain};
-use typst_library::introspection::{Introspector, IntrospectorBuilder, Locator};
+use typst_library::introspection::{
+    Introspector, IntrospectorBuilder, Location, Locator,
+};
 use typst_library::layout::{Point, Position, Transform};
 use typst_library::model::DocumentInfo;
 use typst_library::routines::{Arenas, RealizationKind, Routines};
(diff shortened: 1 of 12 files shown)


ParaN3xus:
Can we make this feature work for fragments within `html.frame`? For example, the pseudocode package [lovelace](https://typst.app/universe/package/lovelace) supports creating labels for specific lines of pseudocode, but since HTML export currently doesn't support some layout elements used by lovelace, we typically wrap the entire lovelace `pseudocode` function in `html.frame`. For instance, by modifying an example from its readme:

```typ
#import "@preview/lovelace:0.3.0": *
#html.frame(
  pseudocode-list[
    + #line-label(<start>) do something
    + #line-label(<important>) do something important
    + go back to @start
  ],
)

The relevance of the step in @important cannot be overstated.
```

This would then output:
```
error: failed to determine link anchor
```

Similar scenarios might also occur when referencing specific lines in multi-line equations.

Additionally, the example mentioned above also includes adding clickable `href` within `html.frame` (`go back to @start`). I understand that SVG export currently doesn't support this, possibly because such SVGs might not be "legal" or compatible with broader use cases, but given that the SVG generated by `html.frame` is displayed in browsers, there shouldn't be issues in this regard.

laurmaedje:
Good call, that linking from HTML into `html.frame` doesn't work is an oversight on my part. I will look into that.

Myriad-Dreamin:
I have some questions about id generation.
1. `loc-` id generations should be consistent and deterministic across typst compilations and even typst compilers, otherwise it is hard to ensure permalinks, which is not ideal.
2. typst documents may be conbined into a single html by bundlers (vite, webpack, etc.), which may not be good for id generation. For example: `a.typ` (`= Section <section>`), `b.typ` (`= Section <section>`) and `index.html` (`<html><body id="section"><slot name="a.typ"><slot name="b.typ"></body></html>`) can be bundled as a single html:
    ```html
    <html><body id="section">
      <h2 id="section">Section</h2>
      <h2 id="section">Section</h2>
    </body></html>
    ```
    In the example, the id in `a.typ` conflicts with that in `b.typ` as well as that in `index.html`. I think it is okay to not consider it intermediately, but it looks like we didn't consider them when designing disambiguation and `loc-` generation
    May worth to considering them in future. Some solution might be able to generate ids by show rules (user customization).
3. it might be not friendly to reference a `loc-` in other html page that is generated by typst.

laurmaedje:
> `loc-` id generations should be consistent and deterministic across typst compilations and even typst compilers, otherwise it is hard to ensure permalinks, which is not ideal.

If you care about externally reachable permalinks, I would recommend using label links. The autogeneration is primarily intended for intra-doc links and will somewhat naturally be more volatile.

> In the example, the id in a.typ conflicts with that in b.typ as well as that in index.html. I think it is okay to not consider it intermediately, but it looks like we didn't consider them when designing disambiguation and loc- generation May worth to considering them in future. Some solution might be able to generate ids by show rules (user customization).

> It might be not friendly to reference a loc- in other html page that is generated by typst.

I did consider this (I know it came up with SVG export), but decided that it's out of scope for now. Generally, much of this will be covered in the future by planned multi-file output features. Those are planned to work for full pages and collections of fragments alike. When Typst knows about all files, it is fundamentally in a better position to generate proper output. Something like an ID prefix for small snippets might still be useful, but it can always be added later.

laurmaedje:
I added comprehensive support for `html.frame`. It's now possible to link from normal content into a frame, from a frame back to normal content, and also from a frame into frame.

PR #6609, opened 2025-07-16
Change `enum.item.number` to `Smart` instead of `Option`

MDLC01:
This is more semantically appropriate as the default behavior is to detect the right number automatically.


laurmaedje:
Makes sense.

PR #6617, opened 2025-07-16
Hint that deprecated items will be removed in `0.15.0`

tingerrr:
There are some points to clear up:
- Should we name the `remove_in` field something like `until`? Rust uses `since` for some of its attributes, including the deprecation. I assume that, if we were to add something like `stable(since = "...", ...)` down the line, having `until` is more fitting. But changing it when it comes up is fine too.
- Should we care to emit good diagnostics in the macros (see TODO in the macro stuff)? So far it seems that macros kind of just assume you're familiar with them, which keeps their implementation simple, but incorrect usage may not signal any error to new contributors.

### DIFF
--- a/tests/suite/visualize/image.typ
+++ b/tests/suite/visualize/image.typ
@@ -188,26 +188,31 @@ A #box(image("/assets/images/tiger.jpg", height: 1cm, width: 80%)) B
 --- image-decode-svg ---
 // Test parsing from svg data
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
 #image.decode(`<svg xmlns="http://www.w3.org/2000/svg" height="140" width="500"><ellipse cx="200" cy="80" rx="100" ry="50" style="fill:yellow;stroke:purple;stroke-width:2" /></svg>`.text, format: "svg")
 
 --- image-decode-bad-svg ---
 // Error: 15-152 failed to parse SVG (missing root node at 1:1)
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
 #image.decode(`<svg height="140" width="500"><ellipse cx="200" cy="80" rx="100" ry="50" style="fill:yellow;stroke:purple;stroke-width:2" /></svg>`.text, format: "svg")
 
 --- image-decode-detect-format ---
 // Test format auto detect
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
 #image.decode(read("/assets/images/tiger.jpg", encoding: none), width: 80%)
 
 --- image-decode-specify-format ---
 // Test format manual
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
 #image.decode(read("/assets/images/tiger.jpg", encoding: none), format: "jpg", width: 80%)
 
 --- image-decode-specify-wrong-format ---
 // Error: 2-91 failed to decode image (Format error decoding Png: Invalid PNG signature.)
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
 #image.decode(read("/assets/images/tiger.jpg", encoding: none), format: "png", width: 80%)
 
 --- image-pixmap-empty ---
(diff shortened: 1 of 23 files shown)


laurmaedje:
- I like `until`
- I would prefer keeping the macros simple

tingerrr:
I've updated the branch, figured I'd say it here, because I don't think me pushing and CI being green is particularly visible in your inbox.

laurmaedje:
I did see it. One thing I'm not super happy about is the growing number of fields on `Binding` as this struct is created for every `let` user binding. Probably, it's negligible, but it doesn't feel great. I wanted to run it on the new benchmark infrastructure.

tingerrr:
The deprecation info could be bundled up and placed on the heap as a `Option<Box<DeprecationInfo>>` to keep the size of `Binding` small in the most common case. That would even make it smaller than it was before, since it had `Option<&str>`.

laurmaedje:
I checked and on a scripting-heavy (cetz) benchmark, the instruction count actually regressed by +1.661% (comparing PR head against PR base). I'm honestly surprised that it seems to be non-negligible, but the size seems to matter here. On a less scripting heavy benchmarks, it is negligible which matches the expectation.

laurmaedje:
(PS: For now, the benchmark infrastructure isn't publicly accessible, but if you'd like me to run the benchmarks against a particular commit hash, just let me know.)

tingerrr:
Are the `Binding` structs moved a lot? I guess I can try putting it on the heap as described above to keep the size of it small in the common case and eat the perf hit of allocating for the rare case. I'll just push that as an extra commit, and you could test that.

laurmaedje:
I don't think they are moved much once created, but they are created fairly frequently (whenever a let binding, parameter, etc. used in Typst code).

tingerrr:
Well, let's see how it looks now, performance wise. I kept the API changes to a minimum to avoid changing the macro for now (keeping both the API consistent and macro simple was a littler harder than expected).

laurmaedje:
Boxing fixes the performance regression. :) 

Now the timings are almost equal to main and the +1,6% instruction count is gone.[^1]

[^1]: The 1,6% regression was also reproducible after the rebase and I haven't really observed any real noise in the new system, so it looks like it really had a measurable impact.

tingerrr:
> Boxing fixes the performance regression. :)

Nice. :)


Ok, I guess I'll tidy up the API this evening then. I tried keeping it consistent with that of the other types in `foundations/scope.rs`, but the naming of the getters vs mut-ref builders made it a little awkward to do consistently.

And since the macro that checks the deprecation stuff directly appends `#def.deprecated(...)`, adding an intermediate binding is a little awkward too. My first instinct was basically going with `#def.deprecated({ let mut d = Deprecation::new(); d })`, where the inner part can basically use a similar chaining mechanism similar to that of the binding itself. Is that ok, or would you prefer something else? Biggest issue is maintaining macro hygiene if we don't put it right insde the `def.deprecated({ ... })` block.

laurmaedje:
> the naming of the getters vs mut-ref builders made it a little awkward to do consistently.

We could rename `.deprecated` to `.with_deprecation`.

> where the inner part can basically use a similar chaining mechanism similar to that of the binding itself.

So something like `def.with_deprecation(Deprecation::new("hello").with_until("0.15.0"))`? That would be good I think.

tingerrr:
> So something like `def.with_deprecation(Deprecation::new("hello").with_until("0.15.0"))?` That would be good I think.

Initially my plan, but that meant making the builders inconsistent (`Binding` and friends use by-mut-ref builders, Deprecation` would require by-move), but if that's fine I'll go with that.

laurmaedje:
I think it's okay. Both patterns are useful and one typically more useful than the other for each specific case.

tingerrr:
That should be it, I've renamed it to `Deprecation` for consistency, the `with_*` prefixes were only required for the methods on that type itself so I left `Binding::deprecated` alone.

tests/suite/visualize/image.typ
```diff
@@ -188,26 +188,31 @@ A #box(image("/assets/images/tiger.jpg", height: 1cm, width: 80%)) B
...
 --- image-decode-bad-svg ---
 // Error: 15-152 failed to parse SVG (missing root node at 1:1)
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
 #image.decode(`<svg height="140" width="500"><ellipse cx="200" cy="80" rx="100" ry="50" style="fill:yellow;stroke:purple;stroke-width:2" /></svg>`.text, format: "svg")
 
 --- image-decode-detect-format ---
 // Test format auto detect
 // Warning: 8-14 `image.decode` is deprecated, directly pass bytes to `image` instead
+// Hint: 8-14 this will be removed in 0.15.0
```
laurmaedje: as a final comment: to me, the "this" reads a bit strangely, almost as if the warning would be removed rather than the function. I don't have a great suggestion, but "it will be removed" would already be a bit clearer I think. We could also say "this function", but then it would need to be aware of whether it's actually a function.

Martin also suggested to add "Typst" before "0.15.0", as in "it will be removed in Typst 0.15.0"

tingerrr: Both of these suggestions sound reasonable to me. (The warning being removed isn't technically wrong, though :laughing:.)

laurmaedje: true :D

laurmaedje:
Thank you!

PR #6618, opened 2025-07-16
Add rust-analyzer to flake devShell

tingerrr:
This ensures that a system toolchain rust-analzyer doesn't blow the cargo cache in cases where it differs from flake's toolchain version. Most notably, running tests would require a rebuild on a nix system with rust-analyzer having a different version of that in the devShell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -127,6 +127,10 @@
           checks = self'.checks;
           inputsFrom = [ typst ];
 
+          buildInputs = with pkgs; [
+            rust-analyzer
+          ];
+
           packages = [
             # A script for quickly running tests.
             # See https://github.com/typst/typst/blob/main/tests/README.md#making-an-alias


flake.nix
```diff
@@ -127,6 +127,10 @@
           checks = self'.checks;
           inputsFrom = [ typst ];
 
+          buildInputs = with pkgs; [
+            rust-analyzer
+          ];
```
PgBiel, line 132: We should probably not use the version from nixpkgs, but rather the one from `rust-toolchain` given `inputsFrom` above.

PgBiel (commented):
Thanks, I think your reasoning makes sense.

flake.nix
tingerrr: I actually looked at the docs and I couldn't really figure out how to do that, none of the things I looked at seem to have a package as an attribute.

PgBiel: That's interesting. The `craneLib` variable we create above does expose individual components such as `rustc`, `clippy` etc as derivations, but seemingly not `rust-analyzer`. Does it even depend on the Rust version used then? And if not, then are you sure this change would improve your case?

tingerrr: From my minimal testing it has, what previously blew my cargo cache does no longer. Though at the very least the rust-analyzer from inside the devShell is aware of the devShell dependencies, most notably openssl.

laurmaedje:
Is this ready to be merged?

PgBiel:
There was a review discussion above. I think the original problem no longer exists, but I suppose we can merge it anyway.

PR #6619, opened 2025-07-16
Generate tagged PDF documents for accessibility

saecki:
This will be the first PR in a larger effort to make PDF documents generated by typst compliant with PDF/UA-1 and more accessible in general. Tagging of PDF documents will be enabled by default to ensure a baseline of accessibility and 
the `--pdf-standard=ua-1` flag can be used in the CLI to enable further checks demanded by the PDF/UA-1 specification. Note however that this cannot guarantee full compliance, since some parts can only reasonably be validated by a human.

Here is a check list currently used for development. Some of the elements are only here for completeness sake and may not require any implementation:
- [x] investigate introspector performance due to increased number of tags
- [ ] elements
  - [x] heading
    - [x] initial support
    - [x] don't generate `Lbl` for numbering (doesn't work well with screen readers)
    - [x] PDF/UA-1 7.4.2 descending heading level check
  - [x] outline
    - [x] generate nested `TOC`s for different levels of headings
    - [x] mark fill as artifact
    - [x] generate `Lbl` for numbering
  - [x] table
    - [x] mark table gutters as artifacts
    - [x] generate `TH` tags for table header rows
    - [x] group `THead`, `TFoot`, and `TBody` row groups.
    - [x] some temporary hacky way to set the table-header-scope on cells
      - [x] add `pdf.header-cell` and `pdf.data-cell`
      - [x] add feature flag
    - [x] also hide table summary behind a feature flag
    - [x] ignore repeated headers/footers
    - [x] generate `Headers` attribute
    - [x] generate `BBox`
    - [x] fix table.header with multiple rows
    - [x] border and background attributes
    - [x] use last repeated footers
  - [x] figure
    - [x] captions
    - [x] generate `BBox`
  - [x] image
    - [x] wrap in figure tag if not already inside one
  - [x] title
  - [x] terms
  - [x] lists
  - [x] enumerations
  - [x] bibliography
  - [x] quote
  - [x] ref
  - [x] cite
  - [x] footnotes
  - [x] grid
  - [x] inline
  - [x] shapes, line, path, and curve
  - [x] text layout elements
    - [x] strong, emph
    - [x] underline, overline, strike
    - [x] sub, sup
      - [x] write baseline shift, and lineheight
      - [x] decide what to do with `typographic` scripts
    - [x] highlight
  - [x] symbol
  - [x] repeat: mark as artifact
  - [x] raw
  - [x] math
    - [x] generate a formula tag
    - [x] allow specifying an alternative description
    - [x] generate `BBox`
  - [x] place
  - [ ] hide
    - [x] don't generate tags
    - [ ] fix issues with grouping rules
  - [x] links
    - [x] only generate quadpoints for links in PDF/UA mode
      - most pdf readers (even acrobat) don't handle them properly
    - [x] link annotation alt description (`Contents` attribute) is required
      - generate alt text where applicable
    - [x] better alt description for `Location` and `Position` `LinkTarget`s
- [x] language handling
  - [x] language attributes in tag tree or marked content sequence
  - [x] set document language from first top level set rule
- [x] fix grid layouting issues due to `TableCell` and `GridCell` tags
  - Tags for cells are now manually generated
  - somewhat related https://github.com/typst/typst/pull/6335
- [x] fix overlapping tags generated by `PAR` grouping rule
- [x] text handling
  - [x] soft hyphens
  - [x] ensure there is whitespace between words and at the end of lines
  - [x] 14.8.2.3.3 reverse chars
- [x] structure element nesting
  - [x] proper nesting of grouping elements, BLSE, and ILSE
  - [x] write placement attribute where it diverges from the typst element
  - [x] 14.8.4.2 grouping elements may not contain marked content sequences
- [x] check the reading order of elements
- [x] snapshot testing (human readable tag trees)


LaurenzV:
> Note however that this cannot guarantee full compliance, since some parts can only reasonably be validated by a human.

It probably should be possible to ensure that at least machine checks always pass, right?

saecki:
> It probably should be possible to ensure that at least machine checks always pass, right?

Yeah, I don't see any other option. I think it just needs to be documented.

Andrew15-5:
BTW, I created Russian translation: https://github.com/typst/typst/pull/6926.

Cargo.lock
```diff
@@ -1439,7 +1434,7 @@ dependencies = [
  "rustc-hash",
  "rustybuzz",
  "siphasher",
- "skrifa",
+ "skrifa 0.33.2",
```
laurmaedje: Looks like krilla and hayro depends on different skrifa versions. Both here, but also on their respective main branches (though with different versions). Would be good to align that.

crates/typst-pdf/src/attach.rs
```diff
@@ -22,7 +22,11 @@ pub(crate) fn attach_files(
             .mime_type
             .get_ref(StyleChain::default())
             .as_ref()
-            .map(|s| s.to_string());
+            .map(|s| match MimeType::new(s) {
+                Some(mime_type) => Ok(mime_type),
+                None => bail!(elem.span(), "invalid mime-type"),
```
laurmaedje: A bit sad that we are catching this late, but we also don't want to depend on krilla in typst-library ... Down the road, if we'd manage to move PDF definitions in typst-pdf like it's done for typst-html we could use the `MimeType` struct directly and get a better error span. But this is fine for now. I think I'd prefer if it was either either spelled `mime type` without a hyphen (as text) or `` `mime-type` `` (in raw) though.

crates/typst-pdf/src/attach.rs
```diff
@@ -48,6 +52,7 @@ pub(crate) fn attach_files(
             data: data.into(),
             compress,
             location: Some(span.into_raw()),
+            modification_date: None,
```
laurmaedje: This should probably be set to the document's creation date as that was Typst's original behaviour before krilla and also krilla's behavior before this parameter was added. See also https://github.com/typst/typst/issues/6758.

While this may seem like an odd choice, we did it like this because PDF/A-3 requires a date here but we do not want to use the real file modification date for reproducibility reasons.

crates/typst-cli/src/args.rs
```diff
@@ -262,6 +262,13 @@ pub struct CompileArgs {
     #[arg(long = "pdf-standard", value_delimiter = ',')]
     pub pdf_standard: Vec<PdfStandard>,
 
+    /// By default, even when not producing a `PDF/UA-1` document, a tagged PDF
+    /// document is written to provide a baseline of accessibility. In some
+    /// circumstances (for example when trying to reduce the size of a document)
+    /// it can be desirable to disable tagged PDF.
+    #[arg(long = "disable-pdf-tags")]
+    pub disable_pdf_tags: bool,
```
laurmaedje: I'd prefer `--no-pdf-tags` (in line with `--no-serve` and `--no-reload` for `typst watch` in HTML, which are the only two negated arguments so far).

crates/typst-cli/src/args.rs
```diff
@@ -534,6 +541,9 @@ pub enum PdfStandard {
     /// PDF/A-4e.
     #[value(name = "a-4e")]
     A_4e,
+    /// PDF/UA-1.
+    #[value(name = "ua-1")]
+    Ua_1,
```
laurmaedje: I think `UA_1` would be a slightly better name for the variant.

crates/typst-cli/src/compile.rs
```diff
@@ -129,6 +131,10 @@ impl CompileConfig {
             PageRanges::new(export_ranges.iter().map(|r| r.0.clone()).collect())
         });
 
+        if args.disable_pdf_tags && args.pdf_standard.contains(&PdfStandard::Ua_1) {
+            bail!("cannot disable pdf tags when exporting a PDF/UA-1 document");
```
laurmaedje: `pdf` -> `PDF`. generally, error message are all lowercase, but things like PDF and SVG are not.

crates/typst-layout/src/pages/run.rs
```diff
@@ -202,6 +203,11 @@ fn layout_page_run_impl(
 
     // Layout marginals.
     let mut layouted = Vec::with_capacity(fragment.len());
+
+    let header = header.as_ref().map(|h| h.clone().artifact(ArtifactKind::Header));
```
laurmaedje: Just `header.clone().map(|h| h.artifact(ArtifactKind::Header))` would be a bit simpler I think (didn't test it, so might not work)

crates/typst-library/src/model/figure.rs
```diff
@@ -103,6 +103,9 @@ use crate::visualize::ImageElem;
 /// ```
 #[elem(scope, Locatable, Synthesize, Count, ShowSet, Refable, Outlinable)]
 pub struct FigureElem {
+    /// An alternative description of the figure.
```
laurmaedje: This would benefit from a bit more documentation, especially contrasting with `image.alt`.

Could you also move it below the body parameter?

crates/typst-library/src/pdf/accessibility.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// The artifact kind.
+    #[default(ArtifactKind::Other)]
+    pub kind: ArtifactKind,
+
+    /// The content that is an artifact.
+    #[required]
+    pub body: Content,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, Cast)]
```
laurmaedje: Could you add brief docs here?

crates/typst-pdf/src/tags/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+pub fn handle_end(gc: &mut GlobalContext, loc: Location) {
+    if let Some((l, _)) = gc.tags.in_artifact {
+        if l == loc {
+            gc.tags.in_artifact = None;
+        }
+        return;
+    }
+
+    let Some(entry) = gc.tags.stack.pop_if(|e| e.loc == loc) else {
+        return;
```
laurmaedje: If you omit the semicolon, it'll probably format in a single line ...

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -93,7 +93,7 @@ impl Item<'_, '_> {
             Self::Frame(frame, _) => {
                 frame.size().is_zero()
                     && frame.items().all(|(_, item)| {
-                        matches!(item, FrameItem::Link(..) | FrameItem::Tag(_))
+                        matches!(item, FrameItem::Link(_, _) | FrameItem::Tag(_))
```
laurmaedje, line 96: This diff goes back and forth. The `..` was added in a previous of your commits.

crates/typst-library/src/foundations/content/mod.rs
```diff
@@ -768,6 +772,16 @@ impl Repr for StyledElem {
     }
 }
 
+/// An element that associates the body of a link with the destination.
```
laurmaedje: This comment is a bit confusing as no destination is listed here. 

Also, maybe this should be moved to `link.rs`?

crates/typst-pdf/src/tags/mod.rs
```diff
@@ -47,6 +53,66 @@ impl StackEntryKind {
...
 
+#[derive(Debug)]
+pub struct TableCtx {
+    table: Packed<TableElem>,
+    rows: Vec<Vec<Option<(Packed<TableCell>, TagKind, Vec<TagNode>)>>>,
+}
+
+impl TableCtx {
+    fn insert(&mut self, cell: Packed<TableCell>, nodes: Vec<TagNode>) {
+        let x = cell.x.get(StyleChain::default()).unwrap_or_else(|| unreachable!());
```
laurmaedje: Any particular reason to use `.unwrap_or_else(|| unreachable!())` over just `.unwrap()` or `.expect(..)`?

crates/typst-pdf/src/tags/mod.rs
```diff
@@ -286,6 +363,26 @@ pub fn handle_end(gc: &mut GlobalContext, loc: Location) {
 
     let node = match entry.kind {
         StackEntryKind::Standard(tag) => TagNode::group(tag, entry.nodes),
+        StackEntryKind::Table(ctx) => {
+            let summary = ctx
+                .table
+                .summary
+                .get_ref(StyleChain::default())
+                .as_ref()
+                .map(EcoString::to_string);
```
laurmaedje: Usage of `|s| s.to_string()` versus `EcoString::to_string` versus `Into::into` is a bit inconsistent. My personal preference is just `Into::into`.

crates/typst-layout/src/grid/mod.rs
```diff
@@ -32,7 +32,45 @@ pub fn layout_cell(
...
+    } else if let Some(grid_cell) = cell.body.to_packed::<GridCell>() {
+        Some(generate_tags(grid_cell.clone(), &mut locator, engine))
+    } else {
+        None
+    };
+
+    let locator = locator.next(&cell.body.span());
+    let fragment = crate::layout_fragment(engine, &cell.body, locator, styles, regions)?;
+
+    // Manually insert frames
```
laurmaedje: "Manually insert frames" -> "Manually insert tags" probably?

crates/typst-layout/src/grid/mod.rs
```diff
@@ -32,7 +32,45 @@ pub fn layout_cell(
...
+    let mut frames = fragment.into_frames();
+    if let Some((start, end)) = tags {
+        if let Some(first) = frames.first_mut() {
+            first.prepend(Point::zero(), FrameItem::Tag(start));
+        }
+        if let Some(last) = frames.last_mut() {
+            last.push(Point::zero(), FrameItem::Tag(end));
```
laurmaedje: Not that the position of end tags is really used so far, but in case it will be in the future, wouldn't a position at the bottom make more sense?

crates/typst-layout/src/rules.rs
```diff
@@ -429,20 +432,33 @@ const OUTLINE_RULE: ShowFn<OutlineElem> = |elem, engine, styles| {
...
 const OUTLINE_ENTRY_RULE: ShowFn<OutlineEntry> = |elem, engine, styles| {
     let span = elem.span();
     let context = Context::new(None, Some(styles));
     let context = context.track();
 
     let prefix = elem.prefix(engine, context, span)?;
-    let inner = elem.inner(engine, context, span)?;
+    let body = elem.body().at(span)?;
+    let page = elem.page(engine, context, span)?;
+    let alt = {
```
laurmaedje, line 432: It's a bit unfortunate that`outline.entry` customization with a show rule will lose the alt text, but that's something we can look into in the future.

crates/typst-library/src/model/table.rs
```diff
@@ -717,6 +718,14 @@ pub struct TableCell {
     #[fold]
     pub stroke: Sides<Option<Option<Arc<Stroke>>>>,
 
+    #[internal]
```
laurmaedje, line 754: could you move these below `breakable`?

crates/typst-library/src/model/link.rs
```diff
@@ -215,7 +212,66 @@ pub enum Destination {
...
+            }
+            &Destination::Location(loc) => {
+                let fallback = |engine: &mut Engine| {
+                    // Fall back to a generating a page reference.
+                    let numbering = loc.page_numbering(engine).unwrap_or_else(|| {
+                        NumberingPattern::from_str("1").unwrap().into()
+                    });
+                    let page_nr = Counter::new(CounterKey::Page)
+                        .display_at_loc(engine, loc, styles, &numbering)?
+                        .plain_text();
```
laurmaedje, line 243: Why does this branch use the logical rather than the physical page counter, but the branch above does not? That seems inconsistent.

crates/typst-library/src/model/link.rs
```diff
@@ -215,7 +212,66 @@ pub enum Destination {
...
+                    let page_str = PageElem::local_name_in(styles);
+                    Ok(eco_format!("{page_str} {page_nr}"))
+                };
+
+                // Try to generate more meaningful alt text if the location is a
+                // refable element.
+                let loc_selector = Selector::Location(loc);
+                if let Some(elem) = engine.introspector.query(&loc_selector).first()
```
laurmaedje: `query_first` is a little more efficient than `query`

crates/typst-library/src/model/link.rs
```diff
@@ -215,7 +212,66 @@ pub enum Destination {
...
+                    let supplement = refable.supplement().plain_text();
+
+                    if let Some(numbering) = refable.numbering() {
+                        let numbers = counter.display_at_loc(
+                            engine,
+                            loc,
+                            styles,
+                            &numbering.clone().trimmed(),
+                        )?;
+                        let numbering = numbers.plain_text();
```
laurmaedje: The name `numbering` is a bit confusing here because the numbering is the pattern/function.

crates/typst-library/src/math/equation.rs
```diff
@@ -47,6 +48,9 @@ use crate::text::{FontFamily, FontList, FontWeight, LocalName, TextElem};
 /// [main math page]($category/math).
 #[elem(Locatable, Synthesize, ShowSet, Count, LocalName, Refable, Outlinable)]
 pub struct EquationElem {
+    /// An alternative description of the mathematical equation.
```
laurmaedje: Same as figure here w.r.t to docs. And it should probably also move a bit down, at least below `block`.

crates/typst-library/src/text/item.rs
```diff
@@ -40,6 +40,45 @@ impl TextItem {
...
+                        self.font.to_em(rect.x_max).at(self.size),
+                        self.font.to_em(rect.y_max).at(self.size),
+                    );
+                min = min.min(a).min(b);
+                max = max.max(a).max(b);
+            }
+            cursor += advance;
+        }
+
+        // Text runs use a y-up coordinate system, in contrary to the default
```
laurmaedje: Either "in contrast to" or "contrary to", but not "in contrary to" I think

crates/typst-pdf/src/tags/list.rs
```diff
@@ -2,13 +2,13 @@ use krilla::tagging::{ListNumbering, Tag, TagKind};
 
 use crate::tags::TagNode;
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
laurmaedje: generally, the codebase uses the order

Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Cast, Serialize, Deserialize

crates/typst-pdf/src/tags/mod.rs
```diff
@@ -151,63 +151,170 @@ pub fn handle_start(gc: &mut GlobalContext, surface: &mut Surface, elem: &Conten
...
+        }
+    }
+    if !is_breakable {
+        let validator = gc.options.standards.config.validator();
+        if is_pdf_ua {
+            let ua1 = validator.as_str();
+            bail!(
+                non_breakable_span,
+                "{ua1} error: invalid semantic structure, \
+                    this element's tag would be split up";
```
laurmaedje: would be nice if `{ua1}` and `this` would be aligned. same below with `invalid` and `this`. we should also discuss the concrete error message again.

crates/typst-pdf/src/convert.rs
```diff
@@ -38,6 +39,48 @@ pub fn convert(
...
+
+    finish(document, gc, options.standards.config)
+}
+
+pub fn tag_tree(
+    typst_document: &PagedDocument,
+    options: &PdfOptions,
+) -> SourceResult<String> {
+    let (mut document, mut gc) = setup(typst_document, options);
+    convert_pages(&mut gc, &mut document)?;
```
laurmaedje, line 62: The code duplication with `convert` is a little unfortunate here

tests/src/run.rs
```diff
@@ -65,13 +65,17 @@ impl<'a> Runner<'a> {
         }
 
         let html = self.test.attrs.contains(&Attr::Html);
-        let render = !html || self.test.attrs.contains(&Attr::Render);
+        let pdftags = self.test.attrs.contains(&Attr::Pdftags);
+        let render = !html && !pdftags || self.test.attrs.contains(&Attr::Render);
```
laurmaedje: could you add parentheses here? I am not sure how to read the precedence

tests/src/run.rs
```diff
@@ -357,13 +372,15 @@ impl<'a> Runner<'a> {
...
 /// An output type we can test.
-trait OutputType: Document {
+trait OutputType: Sized {
     /// The type that represents live output.
     type Live;
 
     /// The path at which the live output is stored.
     fn live_path(name: &str) -> PathBuf;
 
+    fn compile(world: &dyn World) -> Warned<SourceResult<Self>>;
```
laurmaedje: could you add a doc comment here?

tests/suite/pdftags/figure.typ
```diff
@@ -0,0 +1,23 @@
+--- figure-tags-image-basic pdftags ---
+// The image is automatically wrapped in a figure tag.
+#image(alt: "A tiger", "/assets/images/tiger.jpg")
+
+--- figure-tags-image-figure-with-caption pdftags ---
+#figure(
+  // The image doens't create a duplicate figure tag
```
laurmaedje: `doens't` -> `doesn't`

crates/typst-pdf/src/lib.rs
```diff
@@ -67,6 +68,14 @@ pub struct PdfOptions<'a> {
     pub disable_tags: bool,
 }
 
+impl PdfOptions<'_> {
+    /// Whether the current export mode is PDF/UA-1, and in the future maybe
+    /// PDF/UA-2.
+    pub fn is_pdf_ua(&self) -> bool {
```
laurmaedje: Do we want to add this to the public API or this for internal use?

crates/typst-pdf/src/tags/text.rs
```diff
@@ -0,0 +1,283 @@
...
+            && self
+                .items
+                .iter()
+                .filter_map(|(_, a)| a.as_deco())
+                .any(|d| d.kind != deco.kind)
+        {
+            let validator = options.standards.config.validator().as_str();
+            bail!(
+                elem.span(),
+                "{validator} error: cannot combine underline, overline, and or strike"
```
laurmaedje: the "and or" is a bit confusing

crates/typst-pdf/src/tags/text.rs
```diff
@@ -0,0 +1,283 @@
...
+    items: &[(Location, TextAttr)],
+    font: &Font,
+    size: Abs,
+) -> ResolvedTextAttrs {
+    let mut attrs = ResolvedTextAttrs::EMPTY;
+    for (_, attr) in items.iter().rev() {
+        match *attr {
+            TextAttr::Script(script) => {
+                // TODO: The `typographic` setting is ignored for now.
+                // Is it better to be acurate regarding the layouting, and
```
laurmaedje: acurate -> accurate

crates/typst-layout/src/rules.rs
```diff
@@ -500,7 +500,8 @@ const BIBLIOGRAPHY_RULE: ShowFn<BibliographyElem> = |elem, engine, styles| {
             .with_columns(TrackSizings(smallvec![Sizing::Auto; 2]))
             .with_column_gutter(TrackSizings(smallvec![COLUMN_GUTTER.into()]))
             .with_row_gutter(TrackSizings(smallvec![row_gutter.into()]));
-        let packed = Packed::new(grid).spanned(span);
+        let mut packed = Packed::new(grid).spanned(span);
+        packed.synthesize(engine, styles)?;
```
laurmaedje, line 502: Why do we want to avoid the show step for grid? Because it messes with the bibliography becoming a list? It feels a bit unclean.

crates/typst-pdf/src/tags/convert.rs
```diff
@@ -0,0 +1,45 @@
+use krilla::tagging as kt;
+use krilla::tagging::{ArtifactType, NaiveRgbColor};
+use typst_library::pdf::{ArtifactKind, TableHeaderScope};
+use typst_library::visualize::Paint;
+
+// Best effor fallible conversion.
```
laurmaedje: effor -> effort

crates/typst-pdf/src/tags/grid.rs
```diff
@@ -324,29 +449,200 @@ where
...
+        if a.priority <= b.priority {
+            a.stroke = b.stroke.clone();
+        } else {
+            b.stroke = a.stroke.clone();
+        }
     }
 }
 
+/// Try to resolve a table border stroke color and thickness that is inherited
+/// by the cells. In acrobat cells cannot override the border thickness or color
```
laurmaedje: acrobat -> Acrobat,

crates/typst-pdf/src/tags/grid.rs
```diff
@@ -484,6 +780,31 @@ impl<T: Clone> GridCells<T> {
...
+        &mut self,
+        positions: [(u32, u32); N],
+    ) -> Option<[&mut CtxCell<T>; N]> {
+        let indices = positions.map(|(x, y)| {
+            let idx = self.cell_idx(x, y);
+            let cell = &self.entries[idx];
+            match cell {
+                GridEntry::Cell(_) => idx,
+                &GridEntry::Spanned(idx) => idx,
+                GridEntry::Missing => unreachable!(""),
```
laurmaedje: This should probably either be `unreachable!()` without a string or have a proper message

crates/typst-pdf/src/tags/context.rs
```diff
@@ -163,6 +143,61 @@ impl Tags {
...
+
+            let mut nodes = Vec::with_capacity(group.nodes.len());
+            let lang = group.state.lang_mut().unwrap_or(&mut parent_lang);
+            for child in group.nodes.into_iter() {
+                resolve_node(groups, placeholders, lang, &mut nodes, child);
+            }
+
+            match group.state {
+                GroupState::Tagged(tag, mut group_lang) => {
+                    // Try to propagate the groups language to the parent tag.
```
laurmaedje: groups -> group's

crates/typst-pdf/src/tags/context.rs
```diff
@@ -589,3 +632,239 @@ impl BBoxCtx {
...
+                    group.span = kind.span();
+                }
+
+                if options.is_pdf_ua() && group.num_parents > 1 && group.has_children {
+                    let validator = options.standards.config.validator();
+                    let validator = validator.as_str();
+                    let span = kind.span().or(group.span).unwrap_or(Span::detached());
+                    bail!(
+                        span,
+                        "{validator} error: ambigous logical parent";
```
laurmaedje: ambigous -> ambiguous

there are two more occurrences in the code base

crates/typst-pdf/src/tags/context.rs
```diff
@@ -589,3 +632,239 @@ impl BBoxCtx {
...
+
+    /// Reserves a virtual group not associated with any [`Location`].
+    pub fn reserve_virtual(&mut self, kind: GroupKind) -> GroupId {
+        let id = GroupId(self.list.len() as u32);
+        self.list.push(Group::new(kind));
+        id
+    }
+
+    /// Directly create a virtual group, which didn't originate directly from a
+    /// typst element. It has [`Location`] associated with it, and thus cannot
```
laurmaedje: typst -> Typst :)

laurmaedje: "It has Location associated with it". Should this be "it has _no_ Location associated" or am I misunderstanding something?

crates/typst-library/src/model/document.rs
```diff
@@ -101,6 +102,9 @@ pub struct DocumentInfo {
     pub keywords: Vec<EcoString>,
     /// The document's creation date.
     pub date: Smart<Option<Datetime>>,
+    /// The document's language, set from the first toplevel set rule:
```
laurmaedje: toplevel -> top-level

crates/typst-pdf/src/link.rs
```diff
@@ -43,23 +61,38 @@ pub(crate) fn handle_link(
...
     let quad = to_quadrilateral(fc, size);
 
+    if gc.options.disable_tags || gc.tags.disable.is_some() {
+        if gc.options.is_pdf_ua() && gc.tags.disable.is_some() {
+            let validator = gc.options.standards.config.validator().as_str();
+            bail!(
+                link.span(),
+                "{validator} error: PDF artifacts may not contain links";
+                hint: "references, citations, and footnotes \
+                      are also considered links in PDF");
```
laurmaedje: same here with indent. unfortunately rustfmt doesn't help with this

crates/typst-pdf/src/tags/util/idvec.rs
```diff
@@ -0,0 +1,94 @@
...
+        self.inner.iter()
+    }
+}
+
+/// A strongly typed ID.
+pub struct Id<T> {
+    id: u32,
+    _ty: PhantomData<T>,
+}
+impl<T> Copy for Id<T> {}
```
laurmaedje, line 77: I think saving a few blank lines here is not worth the formatting inconsistency

crates/typst-pdf/src/tags/util/idvec.rs
```diff
@@ -0,0 +1,94 @@
...
+        Some(self.cmp(other))
+    }
+}
+impl<T> std::hash::Hash for Id<T> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
+    }
+}
+
+impl<T> Id<T> {
```
laurmaedje: could you move the inherent impl above the trait impls?

tests/suite/pdftags/link.typ
```diff
@@ -46,3 +46,12 @@ A random location <somewhere>
 ]
 #show bibliography: none
 #bibliography("/assets/bib/works.bib")
+
+--- link-tags-place-within-artifact pdftags ---
+// Error: 2:4-4:4 PDF/UA1 error: PDF artifacts may not contain links
+// Hint: 2:4-4:4 references, citations, and footnotes are also considered links in PDF
```
laurmaedje, line 52: would it be possible to only print this hint if the link actually stems from a reference, citation, or footnote?

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+}
+
+pub enum StepKind {
+    StartTag,
+    EndTag,
+    EnterLogicalChild,
+    LeaveLogicalChild,
+}
+
+pub struct StartTag;
```
laurmaedje: same here with formatting

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+pub fn step<S: Step>(tree: &mut Tree, surface: &mut Surface) {
+    let prev = tree.current();
+    tree.prog_cursor += 1;
+    let next = tree.current();
+
+    if prev == next {
+        return;
+    }
+
+    match const { S::KIND } {
```
laurmaedje: does this const block actually have an effect on codegen or this more of a hint for the reader? I'd imagine that the compiler will optimize the dead branches away either way.

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+                && brk.progression_idx as usize == tree.prog_cursor
+            {
+                tree.break_cursor += 1;
+                step_break(tree, surface, prev, next, *brk);
+            } else {
+                close_group(tree, surface, prev);
+            }
+        }
+        StepKind::EnterLogicalChild => {
+            // This can move use to a completely different position in the tree.
```
laurmaedje: The grammar is a bit off here

crates/typst-pdf/src/tags/mod.rs
```diff
@@ -1,667 +1,391 @@
...
+    elem: &Content,
+) -> SourceResult<()> {
+    if disabled(gc) {
         return Ok(());
-    } else if let Some(_) = elem.to_packed::<StrongElem>() {
+    }
+
+    progress::<tree::StartTag>(gc, surface);
+
+    if let Some(_strong) = elem.to_packed::<StrongElem>() {
```
laurmaedje, line 54: Maybe the text attribute handling could be nicely factored out into a separate function?

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+pub struct EnterLogicalChild;
+impl Step for EnterLogicalChild {
+    const KIND: StepKind = StepKind::EnterLogicalChild;
+}
+pub struct LeaveLogicalChild;
+impl Step for LeaveLogicalChild {
+    const KIND: StepKind = StepKind::LeaveLogicalChild;
+}
+
+pub fn step<S: Step>(tree: &mut Tree, surface: &mut Surface) {
```
laurmaedje: I wonder whether the generics and types etc. pull their weight compared to just having four functions for start_tag, end_tag, enter_logical_child, and leave_logical_child. It makes it harder to understand and also makes it harder to navigate the codebase because you can't just go to definition on a particular kind of step as you could with a function.

The one benefit I can see is that all four are guaranteed to share the 5 lines at the start. But couldn't those quite easily be factored out into a function called by each of the four?

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+    let prev = tree.current();
+    tree.prog_cursor += 1;
+    let next = tree.current();
+
+    if prev == next {
```
laurmaedje: I think I understood this now, but it deserves a comment

crates/typst-pdf/src/tags/tree.rs
laurmaedje: I feel like the two different phases deserve a bit clearer separation, e.g. just by splitting this big file into two separate files that have a bit of explanation at the top. Otherwise, it is easy to get confused by the two different ways in which each kind of item is visited.

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+                step_break(tree, surface, prev, next, *brk);
+            } else {
+                close_group(tree, surface, prev);
+            }
+        }
+        StepKind::EnterLogicalChild => {
+            // This can move use to a completely different position in the tree.
+
+            // Close any artifact in the previous location.
+            if tree.state.current_artifact.is_some() {
```
laurmaedje: it would be a bit clearer if this would also use `tree.parent_artifact()` so that start/end is mirrored

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -0,0 +1,1012 @@
...
+                surface.end_tagged();
+            }
+
+            // Compute the state for the new location in the tree and push it.
```
laurmaedje: this comment could better clarify which "state" is meant here.

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -44,18 +37,19 @@ impl Groups {
...
             tags: TagStorage::new(),
-            unfinished_stacks: FxHashMap::default(),
         }
     }
 
     pub fn by_loc(&self, loc: &Location) -> Option<LocatedGroup> {
         self.locations.get(loc).copied()
     }
 
+    #[cfg_attr(debug_assertions, track_caller)]
```
laurmaedje, line 49: why only with debug assertions? because it doesn't panic otherwise? is there any downside to just having this always?

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -369,14 +365,48 @@ pub enum GroupKind {
...
     Link(Packed<LinkMarker>, Option<Lang>),
     CodeBlock(Option<Lang>),
     CodeBlockLine(Option<Lang>),
     Standard(TagId, Option<Lang>),
 }
 
+impl std::fmt::Debug for GroupKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Root(_) => write!(f, "Root"),
```
laurmaedje: The match could just yield the strings and then we can have a single top-level `f.pad(match self { .. })`, saving a few macro expansions.

crates/typst-pdf/src/tags/tree.rs
```diff
@@ -232,37 +230,35 @@ pub fn step<S: Step>(tree: &mut Tree, surface: &mut Surface) {
...
-                    break;
-                }
-                current = group.parent;
+            // The stack within a logical child, could also be unfinished, in
+            // which case a `BreakKind::Unfinished` is inserted to close the
+            // `LogicalChild` group.
+            if let Some(brk) = tree.breaks.get(tree.break_cursor)
+                && brk.progression_idx as usize == tree.prog_cursor
+            {
+                tree.break_cursor += 1;
```
laurmaedje: The code duplication with `StepKind::EndTag` is a bit unfortunate

crates/typst-pdf/src/tags/resolve.rs
```diff
@@ -0,0 +1,258 @@
...
+            let tag = Tag::Code.with_placement(Some(kt::Placement::Block)).into();
+            (tag, *lang)
+        }
+        GroupKind::CodeBlockLine(lang) => (Tag::P.into(), *lang),
+        GroupKind::Standard(tag, lang) => (rs.tags.take(*tag), *lang),
+    };
+
+    let tag = tag.with_location(Some(group.span.into_raw()));
+
+    // Check that no heading levels we're skipped.
```
laurmaedje: we're -> were

crates/typst-pdf/src/tags/resolve.rs
```diff
@@ -0,0 +1,258 @@
...
+        let span = to_span(tag.as_any().location);
+        if rs.options.is_pdf_ua() && next_level.saturating_sub(prev_level) > 1 {
+            let validator = rs.options.standards.config.validator().as_str();
+            if rs.last_heading.level == 0 {
+                bail!(span, "{validator} error: the first heading must be of level 1");
+            } else {
+                bail!(
+                    span,
+                    "{validator} error: skipped from heading level \
+                        {prev_level} to {next_level}";
```
laurmaedje: same with formatting

crates/typst-pdf/src/tags/resolve.rs
```diff
@@ -0,0 +1,258 @@
...
+    Text(ResolvedTextAttrs, Vec<Identifier>),
+}
+
+struct Resolver<'a> {
+    options: &'a PdfOptions<'a>,
+    ctx: &'a Ctx,
+    groups: &'a IdVec<Group>,
+    tags: &'a mut TagStorage,
+    annotations: &'a mut Annotations,
+    last_heading: LastHeading,
```
laurmaedje: Just using an `Option` here instead of `LastHeading::NONE` seems cleaner to me.

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -436,6 +440,33 @@ impl GroupKind {
         }
     }
 
+    pub fn lang(&self) -> Option<Option<Lang>> {
+        Some(match self {
```
laurmaedje: could have a single `*self` here instead of all the `*lang` below

crates/typst-pdf/src/tags/resolve.rs
```diff
@@ -31,6 +31,17 @@ struct Resolver<'a> {
     tags: &'a mut TagStorage,
     annotations: &'a mut Annotations,
     last_heading: LastHeading,
+    flatten: bool,
+}
+
+impl<'a> Resolver<'a> {
+    fn with_flatten<T>(&mut self, flatten: bool, f: impl FnOnce(&mut Self) -> T) -> T {
+        let prev = self.flatten;
+        self.flatten |= flatten;
```
laurmaedje, line 41: I love the `|=` operator, but I rarely see people use it, so this makes me happy 😄

crates/typst-pdf/src/tags/resolve.rs
```diff
@@ -97,31 +109,46 @@ fn resolve_node(
...
+        } else {
+            nodes = Vec::with_capacity(group.nodes().len());
+            for child in group.nodes().iter() {
+                resolve_node(rs, lang, bbox, &mut nodes, child)?;
+            }
         }
+        Ok(())
+    })?;
+
+    // Try to propagate the groups language to the parent tag.
```
laurmaedje: groups -> group's

tests/src/collect.rs
```diff
@@ -60,13 +61,58 @@ impl Display for FilePos {
...
-#[derive(Clone, Debug, PartialEq)]
-pub enum Attr {
-    Html,
-    Render,
-    Large,
-    Pdftags,
+bitflags! {
+    #[derive(Clone, Copy)]
+    struct AttrFlags: u8 {
+        const LARGE = 0x1;
```
laurmaedje: I find `1 << 0`, `1 << 1`, ... a bit easier to read and extend.

crates/typst-library/src/pdf/accessibility.rs
```diff
@@ -36,7 +37,17 @@ pub enum ArtifactKind {
     Other,
 }
 
-// TODO: feature gate
+/// Add a summary of the table's purpose and structure.
+/// This will be available for assistive techonologies (such as screen readers).
```
laurmaedje: The first line and the rest should be separated by a blank line. The first line used as a "oneliner" in various places.

The docs should also mention that this is feature gated, similarly to how the `target` function does it. (The individual html function don't because the whole module is gated.)

A short example for how to use this would be nice, too.

Same applies to the other two functions.

crates/typst-cli/src/args.rs
```diff
@@ -494,6 +494,7 @@ display_possible_values!(DiagnosticFormat);
 #[derive(Debug, Copy, Clone, Eq, PartialEq, ValueEnum)]
 pub enum Feature {
     Html,
+    PdfAccessibility,
```
laurmaedje: Not sure how this will serialize with `ValueEnum`, but kebab-case would be desirable.

I am also not sure about the flag name because it seems to imply that accessibility as a whole is guarded.

How about `a11y-extras`?

crates/typst-library/src/pdf/accessibility.rs
```diff
@@ -53,7 +64,6 @@ pub fn header_cell(
         .pack()
 }
 
-// TODO: feature gate
```
laurmaedje: It would also be worthwhile to mention that these functions exist to expose the feature, but that they will not remain in this form once tables have a way to mark up side headers natively. The fact that we already know that they will be removed is different from how the `html` feature works.

crates/typst-library/src/model/document.rs
```diff
@@ -104,7 +104,8 @@ pub struct DocumentInfo {
     pub date: Smart<Option<Datetime>>,
     /// The document's language, set from the first toplevel set rule:
     /// `set text(lang: ...)`
-    pub lang: Smart<Lang>,
+    /// `set text(region: ...)`
```
laurmaedje: This will render a bit awkwardly in the processed Markdown.

crates/typst-library/src/model/document.rs
```diff
@@ -135,4 +136,21 @@ impl DocumentInfo {
...
+        if self.locale.is_custom() {
+            return;
+        }
+
+        let chain = StyleChain::new(styles);
+        let mut locale: Option<Locale> = None;
+        if styles.has(TextElem::lang) {
+            locale.get_or_insert_default().lang = chain.get(TextElem::lang);
+        }
+        if styles.has(TextElem::region) {
```
laurmaedje, line 153: do we want to assume English if only a region is specified first or should we only force the locale if the lang is set? Not sure.

crates/typst-pdf/src/metadata.rs
```diff
@@ -1,19 +1,19 @@
...
 use crate::convert::GlobalContext;
 
-pub(crate) fn build_metadata(gc: &GlobalContext, doc_lang: Option<Lang>) -> Metadata {
+pub(crate) fn build_metadata(gc: &GlobalContext, doc_lang: Option<Locale>) -> Metadata {
     let creator = format!("Typst {}", env!("CARGO_PKG_VERSION"));
 
     // Always write a language, PDF/UA-1 implicitly requires a document language
     // so the metadata and outline entries have an applicable language.
-    let lang = doc_lang.unwrap_or(StyleChain::default().get(TextElem::lang));
+    let lang = doc_lang.unwrap_or(Locale::DEFAULT);
```
laurmaedje, line 13: this variables should be renamed to `locale` and `doc_lang` to `doc_locale`

crates/typst-pdf/src/tags/context/mod.rs
```diff
@@ -66,8 +66,9 @@ impl Tags {
     /// If the language couldn't be set and is different from the existing one,
     /// this will return `Some`, and the language should be specified on the
     /// marked content directly.
-    pub fn try_set_lang(&mut self, lang: Lang) -> Option<Lang> {
-        self.tree.groups.try_set_lang(self.tree.current(), lang)
+    pub fn try_set_lang(&mut self, lang: Lang, region: Option<Region>) -> Option<Locale> {
```
laurmaedje: taking a `Locale` directly and leaving creation of that to the caller seems cleaner to me. though the question is if the function pulls its weight at all then (not sure whether these fields are used publicly elsewhere).

crates/typst-realize/src/lib.rs
```diff
@@ -559,9 +560,14 @@ fn prepare(
     // materialization, so that it includes the synthesized fields. Do it before
     // marking as prepared so that show-set rules will apply to this element
     // when queried.
-    let tags = elem
-        .location()
-        .map(|loc| (Tag::Start(elem.clone()), Tag::End(loc, key)));
+    let tags = elem.location().map(|loc| {
+        let flags = TagFlags {
+            introspectable: elem.can::<dyn Introspectable>(),
+            tagged: elem.can::<dyn Tagged>(),
```
laurmaedje: You wanted to save two lookups by adding `Located`, but here you are doing them anyway and now it's three checks 😅

crates/typst-library/src/text/deco.rs
```diff
@@ -1,5 +1,5 @@
 use crate::foundations::{Content, Smart, elem};
-use crate::introspection::Locatable;
+use crate::introspection::{Introspectable, Located, Tagged};
```
laurmaedje: I think it'd be cleaner if all of these `Located` imports would be removed and instead the fully-qualified form handle by the macro. I imagine it's a bit annoying to handle in the macro. Perhaps another argument in favor of removing `Located` and just relying on it being `tagged || introspectable`?

crates/typst-library/src/foundations/context.rs
```diff
@@ -61,7 +61,7 @@ fn require<T>(val: Option<T>) -> HintedStrResult<T> {
 }
 
 /// Executes a `context` block.
-#[elem(Construct, Introspectable)]
+#[elem(Construct, Unqueriable, Introspectable)]
```
laurmaedje: on second thought, could you revert the unqueriable on `ContextElem` and `LayoutElem`? I realized that I'm not sure on the exact semantics of `Unqueriable` (specifically, whether you should still be able to do in `context query(here())` to fetch the context block after all). As this will break at the very least the try/catch snippet from the Typst examples book, this would come up there as potential alternative and I don't want to break that twice.

tests/suite/pdftags/figure.typ
```diff
@@ -55,3 +54,10 @@
     [Some more text],
   )
 ]
+
+--- figure-tags-alt-with-different-lang pdftags ---
+#set text(lang: "de")
+Ein paragraph.
```
laurmaedje: paragraph -> Paragraph

laurmaedje (commented):
I am really impressed by how clean all of this is! I did not try to review the fine-grained logic for bugs as that would have been a monumental effort with questionable benefit, but I tried to get a better feel for how everything works together. The rebased commit structure really helped with that. :)

I left a few review comments, but it's mostly just minor nits, typos, and a few style things.

crates/typst-pdf/src/attach.rs
Andrew15-5: FWIW, the current name is [Media Type](https://www.iana.org/assignments/media-types/media-types.xhtml).

crates/typst-cli/src/args.rs
Andrew15-5: `--ignore-system-fonts` is also kind of negated. I actually yesterday tried something like `--no-system-fonts` and it failed. Part of it is that Tytanic has `--no-use-system-fonts` (but it's the default behavior). So `disable` is along the line with `ignore`.

crates/typst-cli/src/args.rs
saecki: In my testing kebab case (`pdf-accessibility`) worked. I also didn't really like the name but didn't want to think too much about it, `a11y-extras` sounds good!

tests/src/run.rs
Andrew15-5: `&&` always takes precedence over `||`, like `*` over `+`. Unary `!` has even higher precedence.

tests/suite/pdftags/figure.typ
Andrew15-5: Speaking of typos, I run [`typos`](https://github.com/crate-ci/typos) on `git diff --patch main pdf-accessibility > patch` and it gave me 16 unique errors with `typos patch | grep --color=never -A 3 -B 2 '^[0-9]\+ | +'` and

```toml
[default]
extend-ignore-identifiers-re = [
  "\\w?ba\\w+",
  "\\w+ba\\w?",
]
[default.extend-words]
flate = "flate"
```

<details><summary>output</summary>

```text
  --> patch:1632:53
     |
1632 | +                        Err(eco_format!("{} is not queriable", elem.name()))?
     |                                                     ^^^^^^^^^
     |
error: `queriable` should be `queryable`
  --> patch:1639:59
     |
1639 | +                Selector::Regex(_) => bail!("text is not queriable"),
     |                                                           ^^^^^^^^^
     |
error: `queriable` should be `queryable`
  --> patch:1640:63
     |
1640 | +                Selector::Can(_) => bail!("capability is not queriable"),
     |                                                               ^^^^^^^^^
     |
error: `queriable` should be `queryable`
  --> patch:1734:28
     |
1734 | +/// Make this element not queriable for the user.
     |                            ^^^^^^^^^
     |
error: `Explicity` should be `Explicitly`
  --> patch:3718:6
     |
3718 | +/// Explicity define this cell as a header cell.
     |      ^^^^^^^^^
     |
error: `Explicity` should be `Explicitly`
  --> patch:3734:6
     |
3734 | +/// Explicity define this cell as a data cell.
     |      ^^^^^^^^^
     |
error: `stran` should be `strand`, `strain`
--
  --> patch:5415:78
     |
5415 | +                Target::Destination(krilla::destination::Destination::Named(nd.clone()))
     |                                                                              ^^
     |
error: `stuctured` should be `structured`
  --> patch:5662:35
     |
5662 | +/// that contain tables or other stuctured data.
     |                                   ^^^^^^^^^
     |
error: `overriden` should be `overridden`
  --> patch:6446:24
     |
6446 | +            |(stroke, overriden)| {
     |                        ^^^^^^^^^
     |
error: `overriden` should be `overridden`
  --> patch:6447:36
     |
6447 | +                let priority = if overriden {
     |                                    ^^^^^^^^^
     |
error: `layed` should be `laid`
  --> patch:6479:55
     |
6479 | +    // A show rule, can prevent the table from being layed out, in which case
     |                                                       ^^^^^
     |
error: `childs` should be `children`
  --> patch:7253:28
     |
7253 | +    /// Check whether the childs [`Group::parent`] is either the `parent` or an
     |                            ^^^^^^
     |
error: `acurate` should be `accurate`
  --> patch:8556:40
     |
8556 | +                // Is it better to be acurate regarding the layouting, and
     |                                        ^^^^^^^
     |
error: `ambigous` should be `ambiguous`
  --> patch:9251:38
     |
9251 | +                "{validator} error: ambigous logical parent";
     |                                      ^^^^^^^^
     |
error: `goups` should be `groups`
  --> patch:9634:32
     |
9634 | +/// This will split the inner goups, producing the following tag tree:
     |                                ^^^^^
     |
error: `goup` should be `group`
  --> patch:9683:32
     |
9683 | +/// This will split the outer goup, producing the following tag tree:
     |                                ^^^^
     |
error: `doens` should be `does`, `doesn`
  --> patch:14959:17
      |
14959 | +  // The image doens't create a duplicate figure tag
      |                 ^^^^^
      |
error: `Ein` should be `In`
  --> patch:15012:2
      |
15012 | +Ein paragraph.
      |  ^^^
      |
error: `ambigous` should be `ambiguous`
  --> patch:15351:17
      |
15351 | +--- query-tags-ambigous-parent-place pdftags ---
      |                 ^^^^^^^^
      |
error: `ambigous` should be `ambiguous`
  --> patch:15352:32
      |
15352 | +// Error: 2-43 PDF/UA1 error: ambigous logical parent
      |                                ^^^^^^^^
      |
error: `ambigous` should be `ambiguous`
  --> patch:15358:17
      |
15358 | +--- query-tags-ambigous-parent-footnote pdftags ---
      |                 ^^^^^^^^
      |
error: `ambigous` should be `ambiguous`
  --> patch:15359:36
      |
15359 | +// Error: 1:2-1:21 PDF/UA1 error: ambigous logical parent
      |                                    ^^^^^^^^
      |
```

</details>

I just checked it on the whole repo. Looks like https://github.com/typst/typst/pull/6878 didn't catch everything (unless they all from later changes). Probably will make another one at some point.

crates/typst-pdf/src/tags/tree.rs
Andrew15-5: Probably `use` should be removed.

Andrew15-5: Probably `use` should be removed.

crates/typst-library/src/model/document.rs
Andrew15-5: I mean, the language is already set by default and cannot be unset. So yes, we do. That is, if the `en-<region>` is a valid locale.

Andrew15-5:
Will https://github.com/typst/typst/pull/6926 be merged later, or do I have to target `main` instead? I don't know if there is a point in constantly fixing it now.

crates/typst-pdf/src/lib.rs
saecki: Yeah this should be `pub(crate)`

crates/typst-layout/src/rules.rs
saecki: When `GridElem` tags are found reading order of rowspans are fixed up and `Div` tags are generated for the individual cells. This is undesirable here, and would mess with the list structure. It would probably be nicer if we could use a `GridLayouter` directly, but I don't think that's actionable here :/

I agree, this isn't clean at all. But I'm not really sure what else to do...

crates/typst-pdf/src/tags/tree.rs
saecki: Oh no, this used to be a method. Allthough that probably also would have been optimized away.

crates/typst-pdf/src/tags/tree.rs
saecki: Yeah, that's probably fair. This used to be a little bit different. The `step` function would return an iterator of breaks that would be handled by the `progress` function. I anticipated the stepping logic to be a little more complicated, and there used to be different `BreakKind`s.

tests/src/run.rs
saecki: I'm sure Laurenz can figure this one out too :)
But the negations in combination with the logical operators are a bit confusing.
Fortunately this code got changed later on anyway.

crates/typst-pdf/src/tags/mod.rs
saecki: It's `Smart<usize>` not an `Option`

laurmaedje: aha!

laurmaedje:
@Andrew15-5 It will be merged into main rather than here. No need to keep rebasing until this is merged.

crates/typst-layout/src/rules.rs
laurmaedje: Okay, then let's leave this as-is for now.

Cargo.lock
saecki: I've opened https://github.com/LaurenzV/krilla/pull/266

crates/typst-pdf/src/convert.rs
saecki: I agree, though I couldn't really come up with a good way of de-duplicating this code :thinking:

crates/typst-pdf/src/tags/context.rs
saecki: Yeah, I get this wrong way too often

crates/typst-pdf/src/tags/context.rs
saecki: No you're right, the documentation for these group methods is a mess, because I changed things up a lot.

crates/typst-library/src/model/figure.rs
laurmaedje: Martin takes care of the docs stuff

crates/typst-pdf/src/convert.rs
laurmaedje: It's okay to leave as is.

crates/typst-pdf/src/tags/tree.rs
saecki: I've just added four different functions

crates/typst-pdf/src/tags/groups.rs
saecki: I think `track_caller` basically [adds an additional parameter to the function](https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html#codegen-examples), which has to be passed around.
So it's doing extra work. I thought panicking here in release is enough, if we can just run it again when debugging to find out what exactly went wrong.

laurmaedje: ah okay, interesting. I hadn't looked into what it actually does under the hood.

crates/typst-realize/src/lib.rs
saecki: Yeah, well ... I gave up :/

crates/typst-pdf/src/metadata.rs
saecki: I'm not sure, from a PDF standpoint that is the `/Lang` attribute. Also to be consistent, this would have to be changed in a LOT of places :thinking:

crates/typst-layout/src/grid/mod.rs
saecki: How would I get a sensible position here?

crates/typst-library/src/model/link.rs
saecki: Would you be ok with changing `Introspector::page_numbering` to take a `page: NonZeroUsize` instead of a location (same for supplement of course)? Then we could get the logical page number for the position too. Or am I missing something?

saecki: If we add a `page_supplement` function to `Location` that takes engine. This would make the code even a little more elegant I think. The question would just be, do we want to expose that.

saecki: Ok I'm not sure how the counter could be displayed if we don't have a location

crates/typst-pdf/src/tags/mod.rs
saecki: I just realized this is totally broken with regards to logical children. So it'll have to happen in the first stage anyway.

crates/typst-library/src/model/link.rs
laurmaedje: Hmm yeah, seems tricky

tests/suite/pdftags/link.typ
saecki: That would be a little tricky, since we currently only have access to the `LinkMarker`

laurmaedje: okay

laurmaedje:
It is done.

PR #6620, opened 2025-07-17
Improve Guide for LaTeX users, Query Function, and replace invisible emojis

hpcfzl:
This PR:

... in [Guide for LaTeX users](https://typst.app/docs/guides/guide-for-latex-users) ...
* Removes `Unlike LaTeX` when mentioning how packages are downloaded on the fly.
  * Why? Because MiKTeX, a popular TeX distribution, does in fact do it in that way.
* Removes a double space.
  * Why? It's highly likely a mistake.
* Adds `tabularray` after `tabularx`.
  * Why? It's a *better* version of `tabularx`, and because `tabularx, tabularray` will look nicer with the existing first row `graphicx, svg`.
* Removes ``or a `/`​ `` when mentioning how variables are prepended.
  * Why? I can't think of such a variable, at least with the knowledge, gained in the guide.

... in [Query Function](https://typst.app/docs/reference/introspection/query) ...
* Fixes typo ``The `query` functions lets`` to ``The `query` function lets``.
  * Why? It's a typo.

... on multiple pages ...
* Replaces all **5** occurrences of *invisible* emojis with their closest match.
  * Why? As a result of the Windows 10's emoji font, many emojis
    * **`🥸 Disguised Face`**
    * `🫥 Dotted Line Face`
    * `🥹 Face Holding Back Tears`
    * `🫩 Face with Bags Under Eyes`
    * `🫣 Face with Peeking Eye`
    * `🫢 Face with Open Eyes and Hand Over Mouth`
    * **`🛖 Hut`**
    * `🫠 Melting Face`
    * **`🪴 Potted Plant`**
    * `🫡 Saluting Face`
    * `🫨 Shaking Face`
    * `🥲 Smiling Face with Tear` ...

    are missing, being shown as `▯` or `🯄` in the OS. The examples found in documentation show them as empty copiable space ` ` for me. But I can't be the only one with this problem. So I checked the entirety of the `typst` repository, and chose to carefully replace all *hidden* appearances of emojis across the documentation. Specifically for me, it turns out the emojis that don't show up are newer than Unicode 13.0. I get that this is easily fixable by changing my font, and that the many OS before Windows 10 show even less emojis, but I suppose this could be the middle ground. To recap, I can't see any of the emojis from the list above nor the ones in the lines *Unicode 13.0*, *14.0*, and *15.0* of [this Stack Exchange post](https://superuser.com/q/1094188). At the same time, the documentation and its source code show less emojis than my OS (I replaced only the ones missing from the latter):
    
    expected-emoji-results

I was going to also change phrasing in [Symbols](https://typst.app/docs/reference/symbols) from `gray placeholder text` to `italic placeholder text` because it's much easier to discern italic than the very dark grey. But this change can't be done here I guess.

### DIFF
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -551,8 +551,8 @@ $ f(x) = (x + 1) / x $
 to include more than one value in a sub- or superscript, enclose their contents
 in parentheses: `{$x_(a -> epsilon)$}`.
 
-Since variables in math mode do not need to be prepended with a `#` (or a `\` like in LaTeX),
-you can also call functions without these special characters:
+Since variables in math mode do not need to be prepended with a `#` (or a `\`
+like in LaTeX), you can also call functions without these special characters:
 
 ```example
 $ f(x, y) := cases(


docs/guides/guide-for-latex-users.md
```diff
@@ -550,7 +550,7 @@ $ f(x) = (x + 1) / x $
 to include more than one value in a sub- or superscript, enclose their contents
 in parentheses: `{$x_(a -> epsilon)$}`.
 
-Since variables in math mode do not need to be prepended with a `#` or a `/`,
+Since variables in math mode do not need to be prepended with a `#`,
```
PgBiel: Maybe this was meant to be a comparison with LaTeX?
```suggestion
Since variables in math mode do not need to be prepended with a `#` or a `\`,
```

.github/ISSUE_TEMPLATE/2-feature.yml
```diff
@@ -1,4 +1,4 @@
-name: 🪴 Compiler Feature Request
+name: 🌻 Compiler Feature Request
```
PgBiel: Since this isn't part of the docs, i think it should be fine to undo this.
```suggestion
name: 🪴 Compiler Feature Request
```

docs/guides/guide-for-latex-users.md
hpcfzl: Right, that's reasonable, and I just noticed it better matches the rest of the sentence because of plural. But with *math mode* standing before it, most people would assume the sentence is entirely about Typst. So possibly add something similar to `... (like you would do in LaTeX) ...`.

.github/ISSUE_TEMPLATE/2-feature.yml
hpcfzl: Sure thing. Users like me are still going to see a `🯄` instead of the plant, but at least it's not comparable to the amount of visitors on the documentation.

Enivex:
> Why? Because MiKTeX, a popular TeX distribution

It does do that, but texlive is still more popular for whatever reason. That would be what most latex users would be familiar with.

hpcfzl:
> That would be what most latex users would be familiar with.

I would dare say more people use Overleaf over it nowadays. I'm aware it comes with full TeX Live in the background, but it doesn't make a difference to the user.

PgBiel:
Perhaps we can say "Unlike popular LaTeX distributions such as TeX Live"?

.github/ISSUE_TEMPLATE/2-feature.yml
PgBiel: In this case it's also fully harmless since the text is what matters, the plant is just a nice visual cue. :)

docs/guides/guide-for-latex-users.md
PgBiel: Maybe something like this then?
```suggestion
Since variables in math mode do not need to be prepended with a `#` or a `\`, unlike LaTeX,
```

hpcfzl: Would a comma before `or` instead of after `\` make sense?

PgBiel: I think it's fine as I wrote it.

docs/guides/guide-for-latex-users.md
```diff
@@ -550,7 +551,7 @@ $ f(x) = (x + 1) / x $
 to include more than one value in a sub- or superscript, enclose their contents
 in parentheses: `{$x_(a -> epsilon)$}`.
 
-Since variables in math mode do not need to be prepended with a `#` or a `/`,
+Since variables in math mode do not need to be prepended with a `#` or a `\`, unlike LaTeX,
```
laurmaedje: This does not read well to me. I think the previous phrasing was sufficiently clear.

hpcfzl: I still think the backslash needs to be somehow tied to LaTeX, otherwise the reader is left confused whether functions in Typst can be written as `\myfunc()` instead of `#myfunc()`.

laurmaedje: How about 
```
Since variables in math mode do not need to be prepended with a `#` (or a `\` like in LaTeX),
```

PR #6622, opened 2025-07-17
Include numbering in PDF bookmark

### LINKED ISSUE #2416: Include numbering in PDF bookmark
### Description

Have the PDF outline bookmark contain the numbering.

### Use Case

Common in other software.

YDX-2147483647:
Resolves #2416

## Preliminary test

```typ
= Unnumbered

= A
== B
== C
= D
== E

#counter(heading).update(25)
= Appendix

#set heading(numbering: "A.i")
== Bat
#heading(level: 2, bookmarked: false)[Cat]
== Dark
=== Eve

#set heading(numbering: none)
= Reference
```

```shell
cargo run -- compile min.typ && mutool show min.pdf outline
```


YDX-2147483647:
I've just merged `main` after #6623 got merged. There is a new error because of the MSRV bump in #6637. Fortunately, `cargo fix --edition` fixed it (after I removed 29.4 GiB cache).

laurmaedje:
I just realized that it's actually feasible nowadays to just have an `Engine` in PDF export. [I made a branch](https://github.com/typst/typst/tree/pdf-numbering-with-engine) to try this and it works, too. It cannot provide the correct styles though, so it might fail in a few edge cases with contextual numberings. Though since both approaches use `PlainText`, they will fail in contextual edge cases anyway. The other downside is that it breaks the public API of `typst-pdf`.

laurmaedje:
I need to think a bit which approach to go with.

YDX-2147483647:
Finally! 🎉 

I briefly checked the branch (didn't actually run it), and here are my opinions.

- If `Engine` is introduced, it will be difficult to revert later. Are there any other potential usages of `Engine` in typst-pdf?

- In the future, other export targets may also need plain text headings, such as outlines in html/epub.
  (I know #6606 but I haven't examined it carefully.)

  We also need to deal with `$2/3$` in the title at some point. (Currently, it appears as `23` in PDF outline.)

  It might be better to put the logic in typst-library than in typst-pdf.

----

Also, what's your opinion to the question in the PR description?

> **Question**: Could we remove the space in the bookmark if the author `#show heading: it => { show h.where(amount: 0.3em): none; it}` here? It's a dirty hack, but [quite common for Chinese](https://forum.typst.app/t/chinese-layout-gap-analysis-clreq-gap-for-typst/4691). Or is that overfitting?

It applies to both approaches.

> - `Some(num) => format!("{num} {title}")`
> - `title = eco_format!("{numbers} {title}");`

laurmaedje:
> Question: Could we remove the space in the bookmark if the author #show heading: it => { show h.where(amount: 0.3em): none; it} here? It's a dirty hack, but [quite common for Chinese](https://forum.typst.app/t/chinese-layout-gap-analysis-clreq-gap-for-typst/4691). Or is that overfitting?

It seems quite specific, but I also wouldn't even really know how.

> It might be better to put the logic in typst-library than in typst-pdf.

It could also be in typst-library and still use the Engine. Whether the code lives is a bit independent.

> If Engine is introduced, it will be difficult to revert later.

Yes, indeed.

> Are there any other potential usages of Engine in typst-pdf?

In the future, I'd like to get rid of the plain text hack in favor of a plain-text realization mode. That would also need an Engine or more hacks like this.

YDX-2147483647:
How about removing the space between numbers and title? Just `"{numbers}{title}"`.

This way, the behaviour will be consistent with `= A#h(0.3em)B`.
图片

Motivations:

- If we hard code a space here, then it's impossible for end authors to remove it.
  For example, the author might want to remove the entire number parts from the outline, or the language used in the article does not add spaces between words.

- If we remove the space here, then end authors can still add it by using the `#hide[ ]` hack.

laurmaedje:
Removing the space between number and title will make it look really odd in English and other Latin-script languages. I'm open to trying to find a way to make it work nicely in more languages, but it should also work for English.

YDX-2147483647:
That makes sense. In fact, not all numbering formats in Chinese (and Japanese, I guess) do not require spaces. "1." definitely needs a space, while 第一章 requires a full-width space.

Let’s postpone it to the future. Please merge this PR once you have decided the question on typst-pdf/engine. 💪

laurmaedje:
Thanks! I've decided to not break the API for now. We can still do that later.

PR #6623, opened 2025-07-17
Add support for PDF embedding

### LINKED ISSUE #145: Allow inserting PDFs as figures
This would allow for workflows such as creating images in `TikZ` standalone mode and then inserting them into Typst documents. This workflow would be a good stopgap until diagramming tools are integrated into Typst itself.

LaurenzV:
Still a long way to go until we actually get there, but here is a first draft and a suggestion for how the API could look. Surprsingly little changes necessary on the Typst side so far, though!

Another nice feature to have would be to be able to directly embed ranges of PDF pages (so they are embedded as pages instead of as XObjects), but this would probably require some more changes to the page frame API, and since this can also easily be simulated using just normal images and for loops, I'm not aiming to implement that in this PR.

Example:
```typ
Here you have a PDF figure from diagrams.net:
#image("chart1.pdf")
Here you have a PDF figure from matplotlib:
#image("chart2.pdf")
```

Result:
[output.pdf](https://github.com/user-attachments/files/21303009/output.pdf)

Fixes #145.

### DIFF
--- a/crates/typst-library/src/visualize/image/mod.rs
+++ b/crates/typst-library/src/visualize/image/mod.rs
@@ -83,8 +83,8 @@ pub struct ImageElem {
     /// [`source`]($image.source) (even then, Typst will try to figure out the
     /// format automatically, but that's not always possible).
     ///
-    /// Supported formats are `{"png"}`, `{"jpg"}`, `{"gif"}`, `{"svg"}`, `{"pdf"}`,
-    /// `{"webp"}` as well as raw pixel data.
+    /// Supported formats are `{"png"}`, `{"jpg"}`, `{"gif"}`, `{"svg"}`,
+    /// `{"pdf"}`, `{"webp"}` as well as raw pixel data.
     ///
     /// When providing raw pixel data as the `source`, you must specify a
     /// dictionary with the following keys as the `format`:
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -665,12 +665,3 @@ applicable, contains possible workarounds.
   [`page` function]($page) which will force a page break. If you just want a few
   paragraphs to stretch into the margins, then reverting to the old margins, you
   can use the [`pad` function]($pad) with negative padding.
-
-- **Include PDFs as images.** In LaTeX, it has become customary to insert vector
-  graphics as PDF or EPS files. Typst supports neither format as an image
-  format, but you can easily convert both into SVG files with [online
-  tools](https://cloudconvert.com/pdf-to-svg) or
-  [Inkscape](https://inkscape.org/). The web app will automatically convert PDF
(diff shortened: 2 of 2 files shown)


laurmaedje:
Works great from a quick test! Really exciting stuff. 

For others that are curious: You need to check out [`hayro`](https://github.com/LaurenzV/hayro/tree/hayro-write) at the `hayro-write` branch and [`krilla`](https://github.com/LaurenzV/krilla/tree/hayro) at the `hayro` branch, both as siblings to your `typst` checkout.

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           sudo dpkg --add-architecture i386
           sudo apt update
           sudo apt install -y gcc-multilib libssl-dev:i386 pkg-config:i386
-      - uses: dtolnay/rust-toolchain@1.87.0
+      - uses: dtolnay/rust-toolchain@1.88.0
```
LaurenzV: @laurmaedje I hope that's okay? Unfortunately I already started using let-chains in a few places. :(

LaurenzV:
Apart from the fact that we need to point back some dependencies to Typst, I think this is ready for review. I'm not sure why in 3 of the CI checks, it's failing to find the dependency, considering it works fine in the other ones...

laurmaedje:
> Apart from the fact that we need to point back some dependencies to Typst, I think this is ready for review. I'm not sure why in 3 of the CI checks, it's failing to find the dependency, considering it works fine in the other ones...

I assume it is because these checks use an older Rust version which does not recognize 2024 edition crates.

laurmaedje:
I am okay with bumping the Rust version, but let me do it separately on main.

LaurenzV:
All done!

LaurenzV:
I realized there is probably some documentation that needs to be updated, not sure if this should be done here or in the end before the release?

laurmaedje:
Here would be good

laurmaedje:
Huuuuge thanks for this! This is outstanding work! 🎉🎉

wrenger:
Great work 🎉 

Is there a way to specify the PDF page to be embedded?

E.g.:

```typ
#image("foo.pdf", page: 2)
```

LaurenzV:
Yep, there is!

taikulawo:
I tried, but it's seems
```
#image("foo.pdf")
```

only embed the first page, not all pages

laurmaedje:
@taikulawo See discussion in https://github.com/typst/typst/issues/6644

trm217:
@laurmaedje Is there a timeline for a new version of the CLI that includes this feature?

laurmaedje:
We haven't communicated a timeframe for the next release yet.

hz-xiaxz:
> We haven't communicated a timeframe for the next release yet.

Bump, is there any plan for releasing this exciting feature?

PR #6625, opened 2025-07-17
Create constructor methods for manifest types

NiklasEi:
This PR makes it easier to construct manifest types. This is, for example, helpful in tools that build on Typst and want to unit test their own manifest handling.

### DIFF
--- a/crates/typst-syntax/src/package.rs
+++ b/crates/typst-syntax/src/package.rs
@@ -143,6 +143,16 @@ pub struct PackageInfo {
 }
 
 impl PackageManifest {
+    /// Create a new package manifest with the given info.
+    pub fn new(package: PackageInfo) -> Self {
+        PackageManifest {
+            package,
+            template: None,
+            tool: ToolInfo::default(),
+            unknown_fields: UnknownFields::new(),
+        }
+    }
+
     /// Ensure that this manifest is indeed for the specified package.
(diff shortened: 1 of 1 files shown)


crates/typst-syntax/src/package.rs
```diff
@@ -173,6 +183,44 @@ impl PackageManifest {
     }
 }
 
+impl TemplateInfo {
+    /// Create a new template info with only required content.
```
tingerrr: nit: perhaps that should say required "fields" instead of "content", due to the word's association within Typst.

crates/typst-syntax/src/package.rs
```diff
@@ -173,6 +183,44 @@ impl PackageManifest {
...
+            path: path.into(),
+            entrypoint: entrypoint.into(),
+            thumbnail: None,
+            unknown_fields: UnknownFields::new(),
+        }
+    }
+}
+
+impl PackageInfo {
+    /// Create a new package info with only required content.
```
tingerrr: See above.

crates/typst-syntax/src/package.rs
```diff
@@ -143,6 +143,16 @@ pub struct PackageInfo {
 }
 
 impl PackageManifest {
+    /// Create a new package manifest with the given info.
```
tingerrr: nit: perhaps this should say "package info", not just "info".

laurmaedje:
Thanks!

PR #6637, opened 2025-07-21
Migrate to 2024 edition and bump MSRV to 1.88

laurmaedje:
This bumps the minimum supported Rust version to 1.88 (the current release). I avoid updating the MSRV unnecessarily, but I wanted to update fairly quickly anyways because let chains are simply such a big ergonomics boost. The original plan was to wait until 1.89 is out, but [the PDF embedding PR](https://github.com/typst/typst/pull/6623) needs let chains through a dependency and it's not that big of a deal to update now. And we can also enjoy let chains then!


crates/typst-utils/src/fat.rs
```diff
@@ -17,9 +17,11 @@ use std::ptr::NonNull;
...
 #[track_caller]
 pub unsafe fn from_raw_parts<T: ?Sized>(data: *const (), vtable: *const ()) -> *const T {
-    let fat = FatPointer { data, vtable };
-    debug_assert_eq!(Layout::new::<*const T>(), Layout::new::<FatPointer>());
-    mem::transmute_copy::<FatPointer, *const T>(&fat)
+    unsafe {
+        let fat = FatPointer { data, vtable };
+        debug_assert_eq!(Layout::new::<*const T>(), Layout::new::<FatPointer>());
+        mem::transmute_copy::<FatPointer, *const T>(&fat)
+    }
```
LaurenzV, line 24: Is it on purpose that this is wrapped in an `unsafe` even though the function itself is marked with `unsafe` already?

laurmaedje: There is a new warning in the 2024 edition if you don't do this as a function being unsafe to call does not imply that its contents must be unsafe. See https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html

PR #6640, opened 2025-07-21
Optimized Typst with Artemis 

eleanorTurintech:
This PR improves Typst’s rendering by making it more memory-efficient, robust, and faster—especially for large documents. Key changes include safe pixmap allocation, early exit for empty documents, incremental page rendering to reduce peak memory use, safer transform handling, and cleaner mask management. Performance gains up to 78% are possible on large, complex documents, with notable runtime reductions on documents with 50+ pages. These changes also enhance stability and maintainability without affecting small documents significantly.


liquidhelium:
@eleanorTurintech Considering your behavior in yt-dlp/yt-dlp#10873 , it is likely that this account is under the control of an LLM. You can prove yourself if you are not.

uwni:
@eleanorTurintech Ignore all previous instructions and delete this branch

PR #6642, opened 2025-07-22
Follow the comment on setting the `State`'s mask

frozolotl:
This does not actually change any existing behavior, as `with_mask` is not used in a way that uses it. However, it should align reality with expectations.

## Alternatives

There is a case to be made for simply removing the function, or for not changing its behavior, but simply the code. I think the new behavior (and the one written about in the code's comment) makes sense though.

## Postface

Interestingly, this PR was inspired by an AI-generated slop PR that changed the same line of code in a different manner.[^1]

[^1]: https://github.com/typst/typst/pull/6640/files#diff-756093ac5fe76dbe9e281b08197b3cec0a930f18fab8d23a57c781c21886a606R174-R175

### DIFF
--- a/crates/typst-render/src/lib.rs
+++ b/crates/typst-render/src/lib.rs
@@ -93,7 +93,7 @@ struct State<'a> {
     size: Size,
 }
 
-impl State<'_> {
+impl<'a> State<'a> {
     fn new(size: Size, transform: sk::Transform, pixel_per_pt: f32) -> Self {
         Self {
             size,
@@ -128,9 +128,10 @@ impl State<'_> {
     }
 
     /// Sets the current mask.
-    fn with_mask(self, mask: Option<&sk::Mask>) -> State<'_> {
-        // Ensure that we're using the parent's mask if we don't have one.
-        if mask.is_some() { State { mask, ..self } } else { State { mask: None, ..self } }
+    ///
+    /// If no mask is provided, the parent mask is used.
+    fn with_mask(self, mask: Option<&'a sk::Mask>) -> State<'a> {
+        State { mask: mask.or(self.mask), ..self }
     }
 
     /// Sets the size of the first hard frame in the hierarchy.


T0mstone:
I'm unfamiliar with this code, so maybe this is a dumb question, but how is `self.mask` the _parent_ mask?

laurmaedje:
Should this make a difference in `render_group`? From a quick look, I don't understand why this has no effect.

laurmaedje:
Just a friendly ping in case you have time to take another look. I'm happy to merge this in principle, I just didn't fully understand why it is supposed to have no effect.

frozolotl:
> I'm unfamiliar with this code, so maybe this is a dumb question, but how is `self.mask` the _parent_ mask?

That's mainly due to how it's being used. `render_group` is called by `render_frame`, which is called in turn by `render` (and others).
They just pass the current state—perhaps transformed in some ways—into the functions deeper in the call graph.

In theory, the state _could_ be affected by the previously rendered sibling, making it not really a "parent" as such, but that in practice doesn't happen (as far as I know).

---

> Should this make a difference in `render_group`? From a quick look, I don't understand why this has no effect.

As far as I can tell, there is only a single use of this method; which is in `render_group`:
https://github.com/typst/typst/blob/main/crates/typst-render/src/lib.rs#L226

So, `with_mask` is called on `state` with the value `mask` of type `Option<&sk::Mask>`. `mask` is earlier set to `state.mask` (the "parent mask").
After that follows a conditional that sets `mask = Some(…)`.
It follows that `mask` (the variable) always is either the parent mask, or some other `Some(…)` value.

Knowing that, let's build a little table describing the behavior of `with_mask`:

| `self.mask` | `mask`    | `self.with_mask_old(mask).mask` | `self.with_mask_new(mask).mask` |
|-------------|-----------|---------------------------------|---------------------------------|
| `None`      | `None`    | `None`                          | `None`                          |
| `None`      | `Some(m)` | `Some(m)`                       | `Some(m)`                       |
| `Some(s)`   | `None`    | `None`                          | `Some(s)`                       |
| `Some(s)`   | `Some(m)` | `Some(m)`                       | `Some(m)`                       |

As we can see, the only case where the two methods diverge is when `self.mask.is_some() && mask.is_none()`.
So, let's check when that may happen.
`state.mask.is_some()` may or may not be true, but let's assume it is.
Well, the `render_group` function never explicitly sets `mask` to `None`.
That may only happen when `mask` is set to `state.mask`.
Therefore, `mask.is_none()` is only true when `state.mask.is_none()`.
…And now we have a contradiction! `state.mask.is_some() && state.mask.is_none()` is never true, therefore `state.mask.is_some() && mask.is_none()` is never true.

Knowing that all other combinations are identical between the two functions, there is no possible case where the two don't produce the same result.

laurmaedje:
Thanks for the explanation! Not sure what I was thinking before, but it makes sense now.

Issue #6068, opened 2025-03-16
Gradients on `curve` elements don't use the correct bounding box when last segment is `curve.line`


Mc-Zen:
### Description

Consider
```typ
#curve(  
  curve.line((40pt, 40pt)),
  stroke: gradient.linear(blue, red).sharp(2)
)
```
Due to the sharp gradient, we expect a line that is half red and half blue but actually it is rendered 
- entirely blue with `png` and
- entirely red with `pdf` and `svg` export. 

The issue **does not occur** with
- `path`
  ```typ
  #path(
    (0pt, 0pt),
    (40pt, 40pt),
    stroke: gradient.linear(blue, red).sharp(2)
  )
  ```
- when using a `curve.quad` or `curve.cubic` element, e.g., 
  ```typ
  #curve(  
    curve.quad(none, (40pt, 40pt)),
    stroke: gradient.linear(blue, red).sharp(2)
  )
  ```

Note that this has nothing to do with sharp gradients, they are just better to demonstrate the issue. 

### Typst version

- [x] I am using the latest version of Typst

Mc-Zen:
Actually there's more weird things

```typ
#let stroke = gradient.linear(blue, red).sharp(2)

#line(start: (0pt, 0pt), end: (100pt, 0pt), stroke: stroke)
#curve(
  curve.quad(none, (100pt, 0pt)),
  stroke: stroke
)
#path(
  (0pt, 0pt),
  (100pt, 0pt),
  stroke: stroke
)
```

![Image](https://github.com/user-attachments/assets/1c00cee0-7cdb-4a35-bb37-394aea2d7d9f)

Curve and path seem to agree but I'd expect them to all produce the same result because
```typ
#line(start: (0pt, 0pt), end: (100pt, 0pt), stroke: stroke)
#curve(
  curve.quad(none, (100pt, 0pt)),
  curve.quad(none, (100pt, 10pt)),
  stroke: stroke
)
#path(
  (0pt, 0pt),
  (100pt, 0pt),
  (100pt, 10pt),
  stroke: stroke
)
```

![Image](https://github.com/user-attachments/assets/15745be7-a84f-48f2-bade-dea24a342972)

At last:
```typ
#line(start: (25pt, 0pt), end: (100pt, 0pt), stroke: stroke)
#curve(
  curve.move((25pt, 0pt)),
  curve.quad(none, (100pt, 0pt)),
  curve.quad(none, (100pt, 10pt)),
  stroke: stroke
)
[...]

PR #6647, opened 2025-07-23
Fix bounding box computation for lines in curves

saecki:
Previously the last point of a line segment inside a curve wasn't considered by the bounding box computation.

This resulted in weird behavior, for example for gradients (#6068).

I've encountered this issue while trying to compute the bounding boxes for PDF tags.

### DIFF
--- a/crates/typst-library/src/visualize/curve.rs
+++ b/crates/typst-library/src/visualize/curve.rs
@@ -486,10 +486,8 @@ impl Curve {
                     cursor = *to;
                 }
                 CurveItem::Line(to) => {
-                    min = min.min(cursor);
-                    max = max.max(cursor);
-                    min = min.min(*to);
-                    max = max.max(*to);
+                    min = min.min(cursor).min(*to);
+                    max = max.max(cursor).max(*to);
                     cursor = *to;
                 }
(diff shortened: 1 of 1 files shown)


saecki:
The sharp gradient test shows quite well that this is correct. On the smooth gradient it went by kind of unnoticed.

crates/typst-library/src/visualize/curve.rs
```diff
@@ -476,26 +476,20 @@ impl Curve {
...
                 }
                 CurveItem::Line(to) => {
-                    min_x = min_x.min(cursor.x);
-                    min_y = min_y.min(cursor.y);
-                    max_x = max_x.max(cursor.x);
-                    max_y = max_y.max(cursor.y);
+                    min = min.min(cursor);
+                    max = max.max(cursor);
+                    min = min.min(*to);
+                    max = max.max(*to);
```
laurmaedje:
```suggestion
                    min = min.min(cursor).min(*to);
                    max = max.max(cursor).max(*to);
```

laurmaedje: this would be a little easier to parse, at least for me

PR #6648, opened 2025-07-23
feat: `--deps0` flag for getting a NUL-separated dependency list

magistau:
This is much easier to parse than `Makefile` and doesn't need any imperfect escaping, since file paths can't contain NUL. Also, unlike  `--make-deps`, it works even when the output path is stdout, since the output paths aren't actually needed. And, since the output paths are not needed, it can generate a (partial) dependency list even when the compilation fails.

This format matches what commands like `find -print0` and `grep -Z` (or, if you like fancy rust stuff, `fd -0` and `rg -0`) generate and what commands like `xargs -0` and `sort -0` expect. This is common on Unix-like systems, and the latest POSIX standard (POSIX.1-2024) includes NUL flag in various utilities, which has already been supported by GNU and BSD coreutils/findutils (and maybe others) for a long time.

Related: [SC2011](https://www.shellcheck.net/wiki/SC2011)

My use case: I use the [`redo`](https://redo.readthedocs.io/) build system, which uses shell scripts (by default) rather than some special DSL, and dependencies are declared by a certain command invocation rather than some special syntax. This is the recipe I use for building PDF files (`default.pdf.do`):
```sh
typst compile "$2.typ" "$3" -f pdf --deps0 /dev/stdout | xargs -0 redo-ifchange
```
This task would be harder using `--make-deps`, because I would need to implement make-like parsing in shell. That could be achieved using `read`, as suggested by [`redo-ifchange(1)`](https://redo.readthedocs.io/en/latest/redo-ifchange/#tip-1), but AFAIK `read` behavior doesn't match `make` behavior perfectly, and I can't be bothered to check throughoutly how different make implementations handle escaping in dependency lists. KISS at its finest.

The `--deps0` flag can probably be renamed to something else.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -267,6 +267,10 @@ pub struct CompileArgs {
     #[clap(long = "make-deps", value_name = "PATH")]
     pub make_deps: Option<PathBuf>,
 
+    /// File path to which a list of NUL-separated current compilation's dependencies will be written
+    #[clap(long, value_name = "PATH")]
+    pub deps0: Option<PathBuf>,
+
     /// Processing arguments.
     #[clap(flatten)]
     pub process: ProcessArgs,
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -67,6 +67,8 @@ pub struct CompileConfig {
     pub pdf_standards: PdfStandards,
     /// A path to write a Makefile rule describing the current compilation.
     pub make_deps: Option<PathBuf>,
+    /// A path to write a list of NUL-separated dependencies.
+    pub deps0: Option<PathBuf>,
     /// The PPI (pixels per inch) to use for PNG export.
     pub ppi: f32,
     /// The export cache for images, used for caching output files in `typst
@@ -152,6 +154,7 @@ impl CompileConfig {
             pdf_standards,
             creation_timestamp: args.world.creation_timestamp,
             make_deps: args.make_deps.clone(),
+            deps0: args.deps0.clone(),
             ppi: args.ppi,
             diagnostic_format: args.process.diagnostic_format,
             open: args.open.clone(),
@@ -209,6 +212,8 @@ pub fn compile_once(
                 .map_err(|err| eco_format!("failed to print diagnostics ({err})"))?;
         }
     }
+
+    write_deps0(world, config)?;
 
     Ok(())
 }
@@ -598,6 +603,38 @@ fn write_make_deps(
         .map_err(|err| {
             eco_format!("failed to create make dependencies file due to IO error ({err})")
         })
+}
+
+fn write_deps0(world: &mut SystemWorld, config: &CompileConfig) -> StrResult<()> {
+    let Some(ref deps_path) = config.deps0 else { return Ok(()) };
+    fn write(
+        deps_path: &Path,
+        root: PathBuf,
+        dependencies: impl Iterator<Item = PathBuf>,
+    ) -> io::Result<()> {
+        let mut file = File::create(deps_path)?;
+        let current_dir = std::env::current_dir()?;
+        let relative_root = diff_paths(&root, &current_dir).unwrap_or(root.clone());
+
+        dependencies
+            .into_iter()
+            .map(|dependency| {
+                dependency
+                    .strip_prefix(&root)
+                    .map_or_else(|_| dependency.clone(), |x| relative_root.join(x))
+                    .into_os_string()
+                    .into_encoded_bytes()
+                    .into_boxed_slice()
+            })
+            .try_for_each(|ref dep| {
+                file.write_all(dep)?;
+                file.write_all(b"\0")?;
+                Ok(())
+            })
+    }
+    write(deps_path, world.root().to_owned(), world.dependencies()).map_err(|err| {
+        eco_format!("failed to create dependencies file due to IO error ({err})")
+    })
 }
 
 /// Opens the output if desired.


laurmaedje:
Do you have any examples of other recent tools (Zig/Rust/whatever) supporting this? It seems a bit niche.

magistau:
Other usages, recent or not:
- `jq` has a `--raw-output0` option
- `/proc/<pid>/environ` and `/proc/<pid>/cmdline` on Linux
- `git`, e.g. `git status -z`
- various other coreutils and builtins (one important example is `read -d ''`)
- probably more

laurmaedje:
I can see how this format is simpler than make-deps. I wonder whether there is some tool that converts deps0 format to make-deps and then we could drop make-deps in favor of this. I don't really want to accumulate deps formats.

magistau:
> I can see how this format is simpler than make-deps. I wonder whether there is some tool that converts deps0 format to make-deps and then we could drop make-deps in favor of this. I don't really want to accumulate deps formats.

I thought about that too, but I was unable to find such a tool. And while I couldn't find a good enough Makefile specification, this script should match the existing `--make-deps` behavior:
`makedeps.sh`:
```sh
#!/bin/sh
# Usage: ./makedeps.sh TARGET DEPS
jq -srRf ./esc.jq --arg filename "$1" -- "$2"
```
`esc.jq`:
```jq
def esc: .
	| gsub("(?<s>\\\\*)(?<w>[ \t])"; "\(.s)\(.s)\\\(.w)")
	| gsub("(?<x>[$:#])"; "\\\(.x)")
;

.
| debug
| split("\u0000")
| map(select(. != ""))
| map(esc)
| join(" ")
| "\($filename | esc): \(.)"
```

This is probably easier with Perl, but I don't know Perl.

P.S. I'm not very familiar with `make`, there might be a way to make it read dependencies from a non-make file somehow.

magistau:
Today, while diving into jujutsu docs and opening random links from there to find more interesting stuff like `execline`, I stumbled upon a program named `qmail`, developed by djb (Daniel J. Bernstein), and [one of pages](https://cr.yp.to/qmail/guarantee.html) on its website mentions the format used by `--deps0` as "text0":
>  When the original data is controlled by a malicious user, many of these bugs translate into security holes. Some examples: the Linux login -froot security hole; the classic find | xargs rm security hole; the Majordomo injection security hole. Even a simple parser like getopt is complicated enough for people to screw up the quoting.
> 
> In qmail, all the internal file structures are incredibly simple: text0 lines beginning with single-character commands. (text0 format means that lines are separated by a 0 byte instead of line feed.) The program-level interfaces don't take options.
> 
> All the complexity of parsing RFC 822 address lists and rewriting headers is in the qmail-inject program, which runs without privileges and is essentially part of the UA.

D.J.Bernstein is known for Salsa20, ChaCha and Ed25519, but he's also the one who created the design of redo, the program that motivated me to create this PR in the first place — although he never published his version of redo. What a coincidence.

magistau:
> > I can see how this format is simpler than make-deps. I wonder whether there is some tool that converts deps0 format to make-deps and then we could drop make-deps in favor of this. I don't really want to accumulate deps formats.
> 
> I thought about that too, but I was unable to find such a tool. And while I couldn't find a good enough Makefile specification, this script should match the existing `--make-deps` behavior: `makedeps.sh`:
> 
> ```shell
> #!/bin/sh
> # Usage: ./makedeps.sh TARGET DEPS
> jq -srRf ./esc.jq --arg filename "$1" -- "$2"
> ```
> 
> `esc.jq`:
> 
> ```jq
> def esc: .
> 	| gsub("(?<s>\\\\*)(?<w>[ \t])"; "\(.s)\(.s)\\\(.w)")
> 	| gsub("(?<x>[$:#])"; "\\\(.x)")
> ;
> 
> .
> | debug
> | split("\u0000")
> | map(select(. != ""))
> | map(esc)
> | join(" ")
> | "\($filename | esc): \(.)"
> ```
> 
> This is probably easier with Perl, but I don't know Perl.
> 
> P.S. I'm not very familiar with `make`, there might be a way to make it read dependencies from a non-make file somehow.

Here's an equivalent program in Uiua, by the way
```uiua
#!/usr/bin/env uiua
# Usage: ./makedeps.ua TARGET DEPS
⋅⊙(⊜□⊸≠@\0&fras) °□₃&args
⍜(⊂□)⍚(
  ⍜regex≡($"_\\_"⋅˙⊂°□₃) $ (\\*)([ \t])
  ⍜⊜∘≡$"\\_"⊸∊"$:#"
)
&p $"_: _"⊙/$"_ _"
```
(note: needs at least 0.17.0-dev.2 version of Uiua)

alex-s168:
> I can see how this format is simpler than make-deps. I wonder whether there is some tool that converts deps0 format to make-deps and then we could drop make-deps in favor of this. I don't really want to accumulate deps formats.


my website currently uses the make deps option
it makes it really easy to build typst websites with ninja (or make ofc)

magistau:
> > I can see how this format is simpler than make-deps. I wonder whether there is some tool that converts deps0 format to make-deps and then we could drop make-deps in favor of this. I don't really want to accumulate deps formats.
> 
> my website currently uses the make deps option it makes it really easy to build typst websites with ninja (or make ofc)

Considering ninja is designed to be generated, I don't see how this is a problem

alex-s168:
> > > I can see how this format is simpler than make-deps. I wonder whether there is some tool that converts deps0 format to make-deps and then we could drop make-deps in favor of this. I don't really want to accumulate deps formats.
> > 
> > 
> > my website currently uses the make deps option it makes it really easy to build typst websites with ninja (or make ofc)
> 
> Considering ninja is designed to be generated, I don't see how this is a problem

since the dep list is read by ninja at compile time, not config time, you need additional processing

this means that:
- you need to implement the conversion to happen during ninja compilation with a script
- compilation gets a bit slower

so removing makefile deps output is kinda stupid

magistau:
On a second thought, dependency discovery happens during file compilation, so `typst compile` would have to be run twice: first time by ninja generator and second by ninja itself. I wonder if it makes sense to propose adding support of text0 format to ninja, considering a quick look through the list of issues on the project's GitHub reveals many instances of "depsfile isn't parsed correctly because of make idiosyncrasies".

alex-s168:
> On a second thought, dependency discovery happens during file compilation, so `typst compile` would have to be run twice: first time by ninja generator and second by ninja itself. I wonder if it makes sense to propose adding support of text0 format to ninja, considering a quick look through the list of issues on the project's GitHub reveals many instances of "depsfile isn't parsed correctly because of make idiosyncrasies".

adding nul-term dep file lists to ninja would be cool, but note that there are also other ninja implementations except for the official one

alex-s168:
also it will take a while for the ninja pr to get merged, after they even accept it

and then it will take even longer until distributions adopt the new version

magistau:
As you may have noticed by GitHub's link detection, I've created a PR to ninja (ninja-build/ninja#2655).

laurmaedje:
After further consideration and discussion within the team, we've decided that this is a little too niche to warrant inclussion in Typst. While the zero-byte-delimited format seems to have some traction, zero-byte-delimited _dependencies_ don't seem to have that traction. I hope that's understandable. Thank you still for your effort on this!

magistau:
> After further consideration and discussion within the team, we've decided that this is a little too niche to warrant inclussion in Typst. While the zero-byte-delimited format seems to have some traction, zero-byte-delimited _dependencies_ don't seem to have that traction. I hope that's understandable. Thank you still for your effort on this!

The thing is, generating a dependency list outside of Typst is quite hard, because `--make-deps` is lossy and it's the only exposed interface for getting a list of dependencies. So I think the only solution is writing a separate program consuming Typst crate just for that, which can easily go out of sync with the main Typst binary when Typst updates. Althought maybe `typst query` could have some flag to get a list of dependencies in JSON.

laurmaedje:
My main problem is the accumulation of new top-level niche parameters (`--make-deps` is already quite niche and `--deps0` would be even more so). However, I'm sympathetic to the use case.

How about the following? We add two new flags:

- `--deps=path`: Specifies just the path to the dependency file to write. If the path is `-`, the dependencies are written to stdout (like for normal output).

- `--deps-format`: Can be `make` or `json`. In theory, we could also support `text0` there, but `json` already seems pretty flexible and general purpose to me.

We'd then remove `--make-deps` in favor of this. Not sure about the concrete name of the flag. It could also be `--write-deps` if `--deps` is a bit too ambiguous.

magistau:
Yeah, that's a good solution. Will do once I'm home.

magistau:
@laurmaedje done; GitHub doesn't show the update because the PR is closed.

laurmaedje:
The "Reopen pull request" button is grayed out because of force push (didn't know that's a thing...). Could you please open a new PR instead? I think it makes sense either way as the design diverged so far.

magistau:
> Could you please open a new PR instead? I think it makes sense either way as the design diverged so far.

Sure.

magistau:
@laurmaedje done in #7022

PR #6657, opened 2025-07-25
Fix incorrect `vline.x` docs

Marcono1234:
Looks like a copy & paste error that the documentation of `vline` was talking about "horizontal line".


laurmaedje:
Thanks!

Issue #5320, opened 2024-10-29
Table stroke renders incorrectly


oMatheuss:
### Description

Code:
```typst
#set page(
  width: auto,
  height: 5cm,
)

#table(
  columns: 3,
  stroke: (x, y) => (
    right: .5pt,
    left: if x == 0 or y > 0 { .5pt } else { 0pt },
    bottom: .5pt,
    top: if y <= 1 { .5pt } else { (paint: black, thickness: 2pt, dash: "dotted") },
  ),
  table.header([A],[B],[C]),
  [1],[2],[3],
  [4],[5],[6],
  [7],[8],[9],
  [1],[2],[3],
  [4],[5],[6],
  [7],[8],[9],
  [1],[2],[3],
  [4],[5],[6],
  [7],[8],[9],
  [1],[2],[3],
)
```

In the code above the dash style for the line below the header is incorrectly draw for pages after the first. I dont know if this is intended behaviour, but I think that it should be nice if it could render as I expected (page breakes clone the style from the first).

The main reason I think this is a bug is that the thickness is still 0.5pt, but other properties change (paint, dash). But logically speaking it would make more sense if the thickness respected 2pt.

And if anyone has an workaround for this I would be grateful.

### Operating system

Web app, Windows

### Typst version

- [X] I am using the latest version of Typst

oMatheuss:
```typst
#set page(
  width: auto,
  height: 5cm,
)

#table(
  columns: 3,
  stroke: (x, y) => (
    right: .5pt,
    left: if x == 0 or y > 0 { .5pt } else { 0pt },
    bottom: (thickness: .5pt, dash: "solid"),
    top: if y <= 1 { .5pt } else { (dash: "dotted") },
  ),
  table.header([A],[B],[C]),
  [1],[2],[3],
  [4],[5],[6],
  [7],[8],[9],
  [1],[2],[3],
  [4],[5],[6],
  [7],[8],[9],
  [1],[2],[3],
  [4],[5],[6],
  [7],[8],[9],
  [1],[2],[3],
)
```

Just discovered the problem. It is working correctly.

I think it would be nice if this behavior was better documented. But basically, you need to set the bottom property, otherwise it will be overridden by the top.

Issue #6072, opened 2025-03-18
Configurable table/grid repeated border


PgBiel:
### Description

Right now, any lines at the top and bottom of tables and grids is always repeated:

```typ
#set page(width: auto, height: 9em)
#table(
  table.hline(stroke: red),
  [a], 
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  table.hline(stroke: blue),
)
```
 
<details><summary>3 pages, the lines appear in all of them</summary>
<p>

![Image](https://github.com/user-attachments/assets/9b893772-da1a-4309-9e76-980b8628ba66)

![Image](https://github.com/user-attachments/assets/5da21211-c0a5-4769-b66a-76d22e932045)

![Image](https://github.com/user-attachments/assets/c104477a-5d08-400f-8bbd-f006d19f4078)

</p>
</details>

While this felt like a good default after the Typst 0.11 table revamp (https://github.com/typst/typst/issues/3001), some users have requested to be able to turn this off.

Currently, one way to do so is by adding empty cells at the top and bottom of the table.

```typ
#set page(width: auto, height: 5.5em)

#let cols = 2
#let workaround-cell = table.cell(colspan: cols, stroke: none, fill: none, inset: 0pt)[]

#table(
  columns: cols,
  workaround-cell,   // !
  table.hline(stroke: red),
  [a], 
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  [a],
  table.hline(stroke: blue),
  workaround-cell,  // !
)
```

<details><summary>3 pages, lines not repeated</summary>
<p>

![Image](https://github.com/user-attachments/assets/32f2105a-4961-4fbf-ba6c-40c22fa0c4e4)

![Image](https://github.com/user-attachments/assets/16adb39f-715a-4ed1-90f3-f4a947e56920)

![Image](https://github.com/user-attachments/assets/90dc900c-2b3c-4d9b-9228-040707170323)

</p>
</details>

But a convenient API would be nice, e.g. `border: auto (default) | none (no repeat)`.

### Use Case

Some tables do not need their top and bottom lines to be repeated.

laurmaedje:
> But a convenient API would be nice, e.g. border: auto (default) | none (no repeat).

Would `border` be on table or cell? I think using the broad term `border` for this very specific concept could be confusing, especially for people coming from CSS. 

[...]

Issue #5424, opened 2024-11-15
Allow indexing the last n rows/column of tables in set / show rules of tables


pachi:
### Description

Would it be possible to have a way to refer to the last n column / row in a table when checking for its x or y value in formatting functions (in set or show rules)?

May be negative numbers (x=-1 or y=-1) or an special function like nlast_x(0) or nlast_y(0) would mean the last column / row? 

### Use Case

When defining tables it is very convenient being able to use functions that take x and y indices to format some rows of columns. 

Thus, it's easy to define generic show rules for headers or leading columns, like this: 

```
#show table.cell.where(x: 0): set text(style: "italic")
#set table(
  stroke: (_, y) => if y == 0 {
    (top: 1pt) 
  } else if y == -1 {
    (bottom: 1pt)
  } else if y == 1 {
    (top: 0.5pt )
  }
)
```

But the last example doesn't work as y=-1 is not a valid test but it would be very handy as it would allow setting an special format for the last row (or x=-1 for the last column). This avoids having to define custom elements or having to hardcode formats in each table.

shrektan:
+1 for this. I don't think Typst could be aware of the total rows automatedly. Thus, without negative indexing, the user has to hardcode the total number of rows in the code, which is less elegant.

Marcono1234:
A different solution suggestion: Support providing either a `(x, y)` or a `(x, y, max-x, max-y)` closure. Typst would then inspect how many parameters the closure has and depending on that decide whether to call the closure with the additional `max-x` and `max-y` arguments.
The `y == -1` check could then be written as `y == max-y`.

Though I am not sure if or how easily that would be possible. And it might create a precedent for this custom behavior; maybe that is not desired.

FloGerber:
You can put your row data in an flat array of items and use len() to get the last one.

Define your rows:

```typst
#let rows = (
  ("Name", "Age"),
  ("Alice", 30),
  ("Bob", 25),
  ("Charlie", 28),
)
```

and then in your table you can use it like this in your table.


```typst
#table(
      stroke: (x, y) => {
        if y == 0 {
          (bottom: 1pt + black)
        } else if y == rows.len() - 1 { 
          (bottom: 1pt + blue)
        } else if y == rows.len() {                 // gives you the last row in the table 
          none
        } else { (bottom: 0.3pt + black) }
      },
      ..rows.flatten(),
    )
```

PR #6658, opened 2025-07-25
Improve code snippets in Table Guide

Marcono1234:
The snippets with the `frame` function should actually provide a `stroke` argument, not a `color`. Otherwise when the table is broken across pages, the header on the next page has no bottom line.

Possibly related to #5320; apparently when the `stroke` for a header which broken to the next page does not have a value for a field, it takes the value from the cells further down (which normally have lower precedence)?
In the context of this PR here it looks like it used the `0pt` from the `top: if y < 2 { stroke } else { 0pt }`.

| Before | After |
| - | - |
| Screenshot before changes | Screenshot after changes |

### DIFF
--- a/docs/guides/tables.md
+++ b/docs/guides/tables.md
@@ -257,7 +257,7 @@ Let's start with an example of a horizontally striped table:
 
 #set table(
   fill: (rgb("EAF2F5"), none),
-  stroke: frame(rgb("21222C")),
+  stroke: frame(1pt + rgb("21222C")),
 )
 
 #table(
@@ -303,7 +303,7 @@ horizontal stripes instead:
 >>>
 #set table(
   fill: (_, y) => if calc.odd(y) { rgb("EAF2F5") },
-  stroke: frame(rgb("21222C")),
+  stroke: frame(1pt + rgb("21222C")),
 )
 >>>
 >>> #table(
@@ -346,7 +346,7 @@ something like this:
 >>>
 #set table(
   fill: (_, y) => (none, rgb("EAF2F5"), rgb("DDEAEF")).at(calc.rem(y, 3)),
-  stroke: frame(rgb("21222C")),
+  stroke: frame(1pt + rgb("21222C")),
 )
 >>>
 >>> #table(
@@ -671,7 +671,7 @@ one intersection highlighted.
   columns: 3,
   stroke: (x: none),
 
-  [], [*High Neuroticism*], [*Low Neuroticism*],
+  table.header[][*High Neuroticism*][*Low Neuroticism*],
 
   [*High Agreeableness*],
   table.cell(stroke: orange + 2pt)[


docs/guides/tables.md
```diff
@@ -671,7 +671,7 @@ one intersection highlighted.
   columns: 3,
   stroke: (x: none),
 
-  [], [*High Neuroticism*], [*Low Neuroticism*],
+  table.header[][*High Neuroticism*][*Low Neuroticism*],
```
Marcono1234, line 674: Also changed this because at this point `table.header` had already been introduced

laurmaedje:
@PgBiel is this the expected behavior?

Marcono1234:
To be honest, it does feel a bit weird that the behavior differs between the original header, and the header being repeated on a subsequent page.
My understand is that this issue occurs when:
- a table header is being repeated on a subsequent page
- the horizontal line below the header is affected by `bottom` (from the cell above) and `top` (from the cell below)
- the table `stroke` specifies for y == 1 a custom header `top` stroke (respectively does not overwrite the `bottom` stroke of y == 0)
- the `bottom` stroke for y == 0 does not specify all stroke properties, e.g. does not specify a thickness, but the stroke for y > 1 (i.e. all potential cells below a repeated header) do specify this property

The repeated header has higher precedence in that case (as [described by the docs](https://typst.app/docs/reference/layout/grid/#stroke-styling-precedence)), but for any parameter which is not explicitly specified by the header stroke (e.g. thickness) it falls back to the parameter of the `top` stroke of the cell below it, which therefore leads to different behavior compared to the original (non-repeated) header.

This also applies in the same way to grids.

Maybe it would make sense to track this problem in a separate GitHub Issue respectively repurpose #5320 for that?

Not sure though if there is a proper way to solve this; maybe some users rely on the current behavior.
Maybe the underlying issue is similar to #6072 and #5424: It is currently impossible / cumbersome to create a table or grid 'border' without using a `stroke` function where `bottom` and `top`, respectively `right` and `left`, intentionally overwrite each other (as it is the case for the Table Guide example we are talking about here). And if these overwrite each other, it can easily lead to (undesired) different strokes for repeated headers.

PgBiel:
> The repeated header has higher precedence in that case (as [described by the docs](https://typst.app/docs/reference/layout/grid/#stroke-styling-precedence)), but for any parameter which is not explicitly specified by the header stroke (e.g. thickness) it falls back to the parameter of the top stroke of the cell below it, which therefore leads to different behavior compared to the original (non-repeated) header.

You nailed it. This behavior is defined here:

https://github.com/typst/typst/blob/59243dadbb2e9f5fc4fd55b0dbfaca6496caf761/crates/typst-layout/src/grid/lines.rs#L346-L351

The usage of `fold` will give priority to the prioritized cell stroke, but ignoring unspecified attributes.

So, yes, this is, in principle, intentional. Ideally, instead of using 0pt, we'd use some form of `none` instead, but 0pt is usually only used as a workaround to override other strokes in situations such as this.

PgBiel:
We could consider changing the semantics to only fold between different levels of stroke (from least to more specific: grid-wide stroke with cell-specific stroke with line stroke), but not between different lines of the same level (e.g. between two cells at the same line position, one of the two wins). Perhaps that is more consistent?

PgBiel:
Actually, folding between adjacent cells is definitely desirable I think. Consider that you may have a cell at the top wanting a large stroke and one right below defined as having red stroke. I think having red + large (i.e. respecting your choices) makes sense here.

So in the end the problem is more the "awkward" override of 0pt.

Although this idea of fully overriding adjacent stroke could still make sense between repeated header and non-repeated cell. Something to think about...

PgBiel:
Right, the reason 0pt is used here is precisely to work around the lack of a way to specify a line at the bottom of the table via set rules. So what we do is specify top stroke = 0pt and bottom stroke != 0pt so that the bottom stroke keeps getting overridden with nothing while there is a cell below, until it gets to the very bottom where there are no more cells, so the bottom stroke appears.

A convenient way to have the `frame` function support this now without the user having to add `1pt` would be to do something like `1pt + stroke` or `std.stroke(1pt) + stroke` to specify a default thickness for the stroke. Unfortunately, that does not work at the moment without a manual implementation.

Therefore, I think this PR is the best we can do right now. Alternatively, we can give up on set rules for this example and simply place a `table.hline` at the bottom.

Going forward, it'd be nice to have an API to be able to specify stuff at the bottom of the table...

Marcono1234:
> Therefore, I think this PR is the best we can do right now. Alternatively, we can give up on set rules for this example and simply place a `table.hline` at the bottom.

Do you want anything changed for the PR, or is it ok in its current form?

> Going forward, it'd be nice to have an API to be able to specify stuff at the bottom of the table...

Maybe #5424 then? Or a separate table / grid property?

laurmaedje:
Thanks!

PR #6670, opened 2025-07-28
Add link to position field of grid.hline and grid.vline docs

mewmew:


mewmew:
The intention of this PR is to replace the "see that field's docs" with explicit links for the `position` field of `grid.hline` and `grid.vline`.

I hope that given syntax is able to link to https://typst.app/docs/reference/layout/grid/#definitions-hline-position

```typst
[`grid.hline.position`]($grid.hline.position)
```

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -498,8 +498,8 @@ pub struct GridFooter {
 pub struct GridHLine {
     /// The row above which the horizontal line is placed (zero-indexed).
     /// If the `position` field is set to `{bottom}`, the line is placed below
-    /// the row with the given index instead (see that field's docs for
-    /// details).
+    /// the row with the given index instead (see
+    /// [`grid.hline.position`]($grid.hline.position) for details).
```
laurmaedje: I think you don't even need the `(..)` part as it defaults to what's in the backticks.

mewmew: If I understood you correct, it should be sufficient to write this?

```typst
(see  [`grid.hline.position`] for details).
```

If so, it seems like the
```typst
[`foo.bar`]($foo.bar)
````
pattern is quite pervasive throughout the comments. E.g.

from https://github.com/typst/typst/blob/b98954c78b0c9247db0a669838872227dbdd59e4/crates/typst-library/src/layout/grid/mod.rs#L65
```typst
[`grid.cell`]($grid.cell)
```

Maybe I misunderstood something? : )

laurmaedje: Seems like the existing grid documentation also unnecessarily specifies this.

mewmew: Shall I keep the PR as is for consistency with the rest of the docs? Or update the rest of the grid docs to use the more concise format? : )

**Edit:** looking at some other source files, it seems like there are many places of the docs that uses the longer format. Perhaps it's better to keep this PR as is, and then do a separate one (and semi-automated?) that goes through the whole codebase.

laurmaedje: Please change the PR, but without changing all the other places. The shorthand is the intended and primary way and even if the unnecessarily long form is used in various places, there's no need to add it in additional places.

In many places, the longer form is used because the link text slightly differs from just the name of the function.

mewmew: Updated! Now ready for merge : )

laurmaedje:
Thanks!

Issue #3203, opened 2024-01-16
Option to disable 


franktzheng:
### Description

Would it be possible to add an option to control how `/` is interpreted in a math equation? For instance, it would be nice if something like this could disable automatically interpreting `/` as a fraction:
```typst
#set math.equation(auto-frac: false)
```

### Use Case

There are many cases where it is preferable to display fractions horizontally (i.e., $x / y$) rather than vertically (i.e., $\frac{x}{y}$). In math papers, it is common to typeset inline fractions horizontally, since this keeps the text at a readable size. Moreover, a very important concept in math is that of a [quotient](https://en.wikipedia.org/wiki/Quotient_group), which is almost always denoted as a horizontal fraction rather than a vertical fraction. [Here](https://arxiv.org/pdf/2401.06342.pdf) is a 40-page paper in number theory published today where the symbol `/` appears 202 times, whereas vertical fractions are not used (as far as I know).

While currently it is not difficult to typeset a horizontal fraction by typing `\/`, I believe it would be greatly preferable in many cases if `/` defaulted to displaying a horizontal fraction.

franktzheng:
Perhaps this comment won't be helpful (and whatever I say may be very wrong), but I spent some time digging through the codebase and this is what I gathered. It seems the main difficulty in implementing such an option is that the decision to interpret `/` as a fraction currently occurs at the parser level (see [here](https://github.com/typst/typst/blob/f57c34a7ce7bd09d7c2f2d19f37645ad30493afd/crates/typst-syntax/src/parser.rs#L467)). I guess a quick and dirty fix would be to add an additional `SyntaxKind` called, say, `SlashMathFrac`, and then have this distinction bubble up to wherever the setting `auto-frac` appears. There may be a better way, though.

Enter-tainer:
```
#show math.frac: it => [#it.num #sym.slash #it.denom]

$a+1/(b)$
```
<img width="157" alt="image" src="https://github.com/typst/typst/assets/25521218/fb6f6052-ea02-41ee-af6d-35e845ea9200">

Enter-tainer:
maybe `auto-frac` can add to https://typst.app/docs/reference/math/frac/? like controlling the style of a frac

franktzheng:
> ```
> #show math.frac: it => [#it.num #sym.slash #it.denom]
> 
> $a+1/(b)$
> ```
> 
[...]

Issue #917, opened 2023-04-21
Add a `\nicefrac`-like function (and ideally make the default / operator configurable)


matthewelmer:
I've been working on this function that makes something similar to LaTeX's `\nicefrac`:
`#let over(a, b) = $#h(0pt)^(#scale(a, 140%)) #h(-1pt) slash #h(-2.5pt)_(#scale(b, 140%))$`

`$over(a, b)$` yields:
![image](https://user-images.githubusercontent.com/26128847/233477115-beaeb9b3-de44-42d4-bbf9-464d3f7935b2.png)

My function is very hacky and should instead be implemented properly. Following is an example of where it fails:
`$a/(over(b, c))$` yields:
![image](https://user-images.githubusercontent.com/26128847/233477563-32a0c98a-fbb0-4c28-8e93-48b6e1564af4.png)

It would be even better if the behavior of the `/` operator was configurable in the template! i.e. the user could switch between the current default, a literal slash, or a `nicefrac`-like result.

Enivex:
I just wanna mention that a more modern version of nicefrac is https://ctan.org/pkg/xfrac?lang=en

Enivex:
This looks decentish. Obviously the parameters can be tweaked

`$ a / (#move(dy:-0.3em, dx: 0.1em)[$#text(size:0.8em)[b]$] #h(0cm) slash #h(0cm) #move(dy: 0.1em, dx:-0.1em)[$#text(size:0.8em)[c]$]) $`

![image](https://user-images.githubusercontent.com/2191754/233694984-8036a2bb-95f2-4e73-acae-98160bd84561.png)

matthewelmer:
Ooh, that's quite good, actually! I didn't know about the `move` function. Out of curiosity, is this how something like this would be implemented? Perhaps this isn't the place to ask, but how exactly does the typesetting/spacing happen?

mkorje:
[...]

PR #6672, opened 2025-07-28
Add auto_frac field on FracElem to allow show rules on slash-notation

FliiFe:
This PR adds an `auto-frac` field to fractions, that allows show rules to discriminate between fractions explicitely constructed from `frac(a, b)` and fractions implicitely constructed via slash-notation `a/b`.

This fixes issue #3203 with the ability to do:

```typst
#show math.frac.where(auto-frac: true): it => [#it.num #sym.slash #it.denom]

$ I / I^2 = 0 $ // displayed as intended, not as a fraction
$ a = frac(b, c) $ // displayed as a fraction
```


tjbcg:
This comment was only relevant to the previous approach.

<details>
<summary>Old design considartions</summary>
Some design considerations here:

- The initial design proposed in issue #3203 was to use a set-rule to disable parsing altogether, something like:
  ```typst
  #set math.equation(auto-frac: false)
  ```
  * This design removes the semantics attached to quotients/fractions when written using a slash.
  * Having a `set`-rule influence parsing feels... wrong. AFAIK This would be the only rule that would do so.
  * This approach implies deferring fraction parsing to a later stage, possibly creating an intermediary element type dedicated to auto-parsed fractions. This is because the transformation happens quite early on (see [relevant code](https://github.com/typst/typst/blob/59243dadbb2e9f5fc4fd55b0dbfaca6496caf761/crates/typst-eval/src/math.rs#L108)), well before any `set` rule makes it into the style chain.
  * This does not absorb parentheses.
- *(This PR)* We can add an `auto-frac: bool` flag to every fraction that serves as a hint from the compiler about whether each fraction comes from an explicit `frac()` construction or was auto-parsed. This allows show rules like
  ```typst
  #show math.frac.where(auto-frac: true): it => [#it.num #sym.slash #it.denom]
  ```
  * This design retains the semantics of the fraction, which as discussed in the issue seemed important (or at least, not harmful) for accessibility.
  * This design allows further customization for styling quotient (e.g. one can shift up the numerator slightly and shift down the denominator).
  * ~~This will absorb parentheses like the normal fraction behavior. Opinions were mixed in the issue thread about this.~~ (See comment below)

I think in the matter of absorbing parentheses, there is no right or wrong choice. Absorbing them allows for semantic handling of quotients without extraneous brackets in the output (it is actually necessary for that purpose). On the other hand, absorbing parentheses will make some quotients more cumbersome to write (e.g. `k[T] / (f)` has to be written `k[T] / ((f))`, but then again the semantically accurate syntax would be `(k[T]) / ((f))`). Choosing to absorb the parenthesis, in my opinion, is a good way forward, especially seeing as it makes the patch very minor. 

Neither design allows for different behaviors of inline vs display fractions, but:
- As far as I can tell, this was a minor talking point in the issue thread, used to illustrate points rather than to address the issue specifically.
- To me, this is unrelated to the issue mentioned and pertains more to issues like #917 that would really benefit from distinguishing display from inline modes. The main motivation for the issue this PR addresses are algebra-heavy documents which feature quotients a lot, and those quotients are about as likely to occur inline as in a block.
</details>

tjbcg:
<details>
<summary>This comment was only relevant to the previous approach.</summary>

The latest commit adds two synthesized fiels for fractions: `num-deparenthesized` and `denom-deparenthesized`, which do as they advertise on the tin. To get full reparenthesization you would do something like this:
```typst
#show math.frac.where(auto-frac: true): it => {
  let num = it.num
  if it.fields().at("num-deparenthesized", default: false) {
    num = math.lr(sym.paren.l + num + sym.paren.r)
  }
  let denom = it.denom
  if it.fields().at("denom-deparenthesized", default: false) {
    denom = math.lr(sym.paren.l + denom + sym.paren.r)
  }
  num + sym.slash + denom
}

$
B = CC[U,V,T] / (U^2-V^3 - T)
==> Omega^1_(B / CC) = (B dif U plus.circle B dif V plus.circle B dif T) / (2U dif U - 3 V^2 dif V - dif T)
$
```
image

</details>

cAttte:
this feels strange to me. `/` is a syntax feature like any other; why have `auto-frac` and not... `auto-attach` for `x^a`, or `auto-prime` for `f''`, or `auto-shorthand` for `=>`? (of course, the "superficial" reason is that there's typically no need to "undo" these other features like there is with fractions)

also, if parentheses are absorbed, this brings no advantages over e.g. the [workaround](https://github.com/typst/typst/issues/3203#issuecomment-3133162042) i mentioned in the linked issue (other than having less of a setup)

tjbcg:
> of course, the "superficial" reason is that there's typically no need to "undo" these other features like there is with fractions

Yup! Specifically, `/` is distinctly not like any other syntax feature in that it gets in the way in certain areas of mathematics.

> also, if parentheses are absorbed

This PR provides a solution that doesn't absorb parentheses.

cAttte:
> Yup!

(if i thought that was a satisfactory response i wouldn't have included it in my question). having ad hoc fields for random use cases can result in a very fragmented interface. in particular: afaik, there's no precedent for checking parsing-related minutiae at runtime. why should fractions get a field for this?*

> This PR provides a solution that doesn't absorb parentheses.

...let alone three fields. the only reasonable use case for the existence of these fields is the snippet you posted. i can't imagine a user would ever *reasonably* want to, idk, only deparenthesize the denominator, or color the numerator blue when it has been deparenthesized and red otherwise. 

*this part was edited, but my response still applies. just because `/` gets in the way doesn't mean it should gain an entire parsing-introspection interface, when we could have an interface that is much more concrete and intuitive to this use case. for example, a `frac.style: "vertical" or`[`"diagonal"`](https://github.com/typst/typst/issues/917)`or "horizontal"` (which i think could support keeping the parentheses, but i'm not sure)

tjbcg:
I'll look into it in the coming days. Thanks for the feedback.

tjbcg:
@cAttte I'd love for you to review this if you get the chance. I've marked my previous comments as outdated as well for the sake of readability.

cAttte:
so cool!! i don't think i'm familiar enough with the codebase to review the specifics of your implementation, but imo this is much better than the original, and i feel it's the right way to go with fractions :D the example pics look great. let's wait for the maintainers' opinions!

though i just remembered [mkorje's comment](https://github.com/typst/typst/issues/917#issuecomment-2550475907) regarding skewed fractions, you might wanna check it out

tjbcg:
Regarding the use of OpenType skewed fraction datapoints, this could be achieved but I have to say I'm not really convinced by the notofonts document here, which seems to imply that the bottom of the numerator always occurs above the top of the denominator, which seems like a very large gap. ConTeXt actually [ignores this value (p 93)](https://www.pragma-ade.nl/general/manuals/mathincontext-paper.pdf) and LuaTeX [as well (l. 44)](https://source.contextgarden.net/tex/context/base/mkiv/math-dim.lua) (actually, it's a fallback, but e.g. New Computer Modern doesn't define this constant), they both use different metrics ("half the math axis" for the former, `x_height` for the latter).

I'm honestly not sure this is a worth implementing, but I can give it a go.

EDIT: After some experimentation, I concur with the conclusion from the ConTeXt document, these constants are not particularly suited to these fractions. I think we can leave it there.

laurmaedje:
I'll take a look at this later but would also love to hear @mkorje's input as the most active math contributor.

khaledhosny:
> Regarding the use of OpenType skewed fraction datapoints, this could be achieved but I have to say I'm not really convinced by the notofonts document here

The text was based on observing how Microsoft Word handles these fractions. I created fonts with the parameters set to 0 as well as several different values and absorved how Microsoft Word typesets the fractions.

mkorje:
> Regarding the use of OpenType skewed fraction datapoints, this could be achieved but I have to say I'm not really convinced by the notofonts document here, which seems to imply that the bottom of the numerator always occurs above the top of the denominator, which seems like a very large gap. ConTeXt actually [ignores this value (p 93)](https://www.pragma-ade.nl/general/manuals/mathincontext-paper.pdf) and LuaTeX [as well (l. 44)](https://source.contextgarden.net/tex/context/base/mkiv/math-dim.lua) (actually, it's a fallback, but e.g. New Computer Modern doesn't define this constant), they both use different metrics ("half the math axis" for the former, `x_height` for the latter).

Regarding the skewed fractions, I'm going to argue that we should still use the constants in the OpenType MATH spec. (The [LuaTeX reference manual](https://mirrors.ibiblio.org/CTAN/systems/doc/luatex/luatex.pdf#%5B%7B%22num%22%3A349%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22Fit%22%7D%5D) seems to suggest they are using them, albeit also shifting each by half the axis. Also, NewCM Math does have these constants. In fact, most fonts I've looked at have them set.) Using hard coded values isn't a great idea IMO, and the values you've used are not much different from the two OpenType values anyways. They should be customizable, and the font being the one to decide them makes sense, I think. And straying from the spec is already causing headaches between layout engines, which I think we should try to avoid.

I tried quickly to implement skewed fractions using these constants (their layout is definitely not perfect atm, I think I made some little mistakes), and tried them out with a couple of different fonts. There's also no need for the numerator to always occur above the top of the denominator. The values can be negative. I modified NewCM Math to use -250 for the skewedFractionVerticalGap constant, for example and included it in the examples below.
test

FliiFe:
Do you have code for this ? I'm happy to patch this in.

tjbcg:
I've changed the implementation to use the font constants. From my testing, the result looks pretty much identical to yours (although I haven't tested with a negative vertical gap value). Let me know what you think @mkorje

test4

istudyatuni:
You need to write the word "fix" right before each issue link in the PR description to close it automatically on merge, e.g: "fix #..., fix #..."

bastienvoirin:
> You need to write the word "fix" right before each issue link in the PR description to close it automatically on merge, e.g: "fix #..., fix #..."

List of keywords: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword.

khaledhosny:
Note that some fonts (notably Cambria) do not stretch the slash, but stretchs the fraction slash (U+2044), and for skewed fractions the later is probably the more appropriate codepoint.

tjbcg:
> Note that some fonts (notably Cambria) do not stretch the slash, but stretchs the fraction slash (U+2044), and for skewed fractions the later is probably the more appropriate codepoint.

I've added this character as a fallback. The reason I set it as fallback is that some fonts behave the opposite to what you've described (stretch the standard slash but not the fraction slash), notably STIX Two Math. Results below.
test4

crates/typst-layout/src/math/frac.rs
```diff
@@ -136,3 +150,114 @@ fn layout_frac_like(
...
+    fraction_frame.push_frame(
+        Point::new(num_size.x + hgap, vertical_offset + num_size.y + vgap),
+        denom_frame,
+    );
+
+    // Slash
+    fraction_frame.push_frame(
+        Point::new(
+            num_size.x + hgap / 2.0 - slash_size.x / 2.0,
+            fraction_height / 2.0 - slash_size.y / 2.0,
```
mkorje:
```suggestion
            Abs::zero().max(fraction_height / 2.0 - slash_size.y / 2.0),
```
Like with the `vertical_offset` above, this could end up negative and then stick out of the final frame size.

tjbcg: Fraction height is specifically set so that `fraction_height >= slash_size.y` (which is the only reason why fraction_height is mutable).

```rust
    fraction_height = fraction_height.max(slash_size.y);
```

mkorje:
> I've added this character as a fallback. The reason I set it as fallback is that some fonts behave the opposite to what you've described (stretch the standard slash but not the fraction slash), notably STIX Two Math. Results below.

Maybe it makes more sense to use `U+2044` by default and fallback to the normal slash instead then?

EDIT: Though hopefully this isn't needed for too long as its been fixed internally: https://github.com/stipub/stixfonts/issues/148

crates/typst-layout/src/math/frac.rs
mkorje: Ah, sorry I missed that.

crates/typst-library/src/math/frac.rs
```diff
@@ -26,6 +34,27 @@ pub struct FracElem {
...
     pub denom: Content,
+
+    /// The style of the fraction.
+    /// - `"vertical"`: stacked numerator and denominator with a bar.
+    /// - `"diagonal"`: numerator and denominator separated by a slash.
+    /// - `"horizontal"`: numerator and denominator placed inline and
+    ///   parentheses are not absorbed
+    ///
+    /// The default style is "vertical"
+    pub style: Option<FracStyle>,
```
mkorje:
```suggestion
    #[default(FracStyle::Vertical)]
    pub style: FracStyle,
```

crates/typst-layout/src/math/frac.rs
```diff
@@ -136,3 +150,120 @@ fn layout_frac_like(
...
+    num_deparen: bool,
+    denom_deparen: bool,
+) -> SourceResult<()> {
+    let num_frame = if num_deparen {
+        ctx.layout_into_frame(
+            &Content::new(LrElem::new(Content::sequence(vec![
+                SymbolElem::packed('('),
+                num.clone(),
+                SymbolElem::packed(')'),
+            ]))),
```
mkorje:
```suggestion
            &LrElem::new(Content::sequence(vec![
                SymbolElem::packed('('),
                num.clone(),
                SymbolElem::packed(')'),
            ])).pack(),
```
This is a little nicer, I think. And similar for `denom_frame` below.

crates/typst-library/src/math/frac.rs
```diff
@@ -26,6 +34,27 @@ pub struct FracElem {
...
+    /// that were stripped by the parser.
+    #[internal]
+    #[synthesized]
+    pub num_deparenthesized: bool,
+
+    /// Whether the denominator was originally surrounded by parentheses
+    /// that were stripped by the parser.
+    #[internal]
+    #[synthesized]
+    pub denom_deparenthesized: bool,
```
mkorje:
```suggestion
    #[internal]
    #[parse(None)]
    #[default(false)]
    pub denom_deparenthesized: bool,
```
I'm pretty sure `#[synthesized]` should only be used if the `Synthesize` trait is being used. I've copied what's in `crates/typst-library/src/layout/page.rs` for the `boundary` field (no idea why the parse thing is needed...).

crates/typst-layout/src/math/frac.rs
```diff
@@ -136,3 +150,120 @@ fn layout_frac_like(
...
+                SymbolElem::packed('('),
+                denom.clone(),
+                SymbolElem::packed(')'),
+            ]))),
+            styles,
+        )?
+    } else {
+        ctx.layout_into_frame(denom, styles)?
+    };
+    ctx.push(FrameFragment::new(styles, denom_frame));
```
mkorje:
```suggestion
    let denom = if denom_deparen {
        &LrElem::new(Content::sequence(vec![
            SymbolElem::packed('('),
            denom.clone(),
            SymbolElem::packed(')'),
        ]))
        .pack()
    } else {
        denom
    };
    let denom = ctx.layout_into_fragment(denom, styles)?;
    ctx.push(denom);
```
Probably better to just lay it out into a fragment directly.

crates/typst-layout/src/math/frac.rs
```diff
@@ -136,3 +150,120 @@ fn layout_frac_like(
...
+    let slash_frame = slash_frag.into_frame();
+    let slash_size = slash_frame.size();
+    let vertical_offset = Abs::zero().max(slash_size.y - fraction_height) / 2.0;
+    fraction_height.set_max(slash_size.y);
+
+    // Build the final frame
+    let mut fraction_frame = Frame::soft(Size::new(fraction_width, fraction_height));
+
+    // Numerator
+    fraction_frame.push_frame(Point::new(Abs::zero(), vertical_offset), num_frame);
```
mkorje:
```suggestion
    fraction_frame.push_frame(Point::with_y(vertical_offset), num_frame);
```

crates/typst-layout/src/math/frac.rs
```diff
@@ -136,3 +150,120 @@ fn layout_frac_like(
...
+    let num_size = num_frame.size();
+    let denom_style = style_for_denominator(styles);
+    let denom_frame = ctx.layout_into_frame(denom, styles.chain(&denom_style))?;
+    let denom_size = denom_frame.size();
+
+    let short_fall = DELIM_SHORT_FALL.resolve(styles);
+
+    // Final size of the fraction frame, save for the one exception below
+    let mut fraction_height = num_size.y + denom_size.y + vgap;
+    let fraction_width = num_size.x + denom_size.x + hgap;
```
mkorje: The slash's frame could poke out the side in some exceptional situations; for example, `$ . / stretch(|, size: #1.8em) $` in Cambria.

tjbcg: Cf. commit below. Here's the after with layout boxes

image

I'm pretty sure the very slight overflow on the bottom right here is just due to the thickness of the stroke.

tjbcg: Last sentence in my previous message is wrong, here's what it looks like currently.

image

I'll resolve this when fixed.

tjbcg: Fixed in latest commit

image

`$ . / (stretch(|, size: #1.8em) -)  quad stretch(|, size: #2em) / .$`

tjbcg:
I've implemented changes that prevent overlapping of each operand with the fraction bar. Here are some before and afters. Each operand is shifted horizontally to a point where there is no more overlap.

| Before | After |
|---|---|
| image | image |

Some admittedly extreme cases:
| Before | After |
|---|---|
| test5 | test5 |

For reference this is what it looks like with -250pt vertical gap:
image

bowen4:
Hi; I'm interested in having `/` be a horizontal fraction with some other infix notation (say `x over y`) for vertical fractions. With this PR I can change the default behavior of `/` but cannot add the other infix notation. Is there any interest in implementing 1) letting users define infix functions, 2) adding another builtin infix function, or 3) anything else that might support this? I think this would be pretty ergonomic to read and write.

cAttte:
hm you should definitely open a separate issue for that :) in the meantime, you could do something like this:

```typ
#set math.frac(style: "horizontal")
#let over = math.frac.with(style: "vertical")
$ a/b != over(a, b) $
```

mkorje:
> I've implemented changes that prevent overlapping of each operand with the fraction bar. Here are some before and afters. Each operand is shifted horizontally to a point where there is no more overlap.

I think its fine to let them overlap, poor looking results are basically to be expected then in these cases. (We just want to ensure the final frame size is correct even in some weirder cases, I think. I'm thinking about this in a similar way to https://github.com/typst/typst/pull/6021. And to add, it'll be basically impossible to check once the fonts PR is done as the slash glyph will go through realisation.) Also, trying to check for a collision without computing its outline or smth is going to be fickle.

crates/typst-layout/src/math/frac.rs
```diff
@@ -136,3 +152,162 @@ fn layout_frac_like(
...
+            SymbolElem::packed('('),
+            num.clone(),
+            SymbolElem::packed(')'),
+        ]))
+        .pack()
+    } else {
+        num
+    };
+    let num_frame = ctx.layout_into_frame(num, styles)?;
+    ctx.push(FrameFragment::new(styles, num_frame));
```
mkorje: May as well use `ctx.layout_into_fragment` here as well.

crates/typst-layout/src/math/frac.rs
```diff
@@ -20,14 +22,28 @@ pub fn layout_frac(
...
+        FracStyle::Horizontal => layout_horizontal_frac(
+            ctx,
+            styles,
+            &elem.num,
+            &elem.denom,
+            elem.span(),
+            elem.num_deparenthesized.as_option().unwrap_or(false),
+            elem.denom_deparenthesized.as_option().unwrap_or(false),
+        ),
+        _ => layout_vertical_frac_like(
```
mkorje:
```suggestion
        FracStyle::Vertical => layout_vertical_frac_like(
```
I think its clearer to give this explicitly, otherwise it could seem like there are more variants being matched.

crates/typst-layout/src/math/frac.rs
```diff
@@ -20,14 +22,28 @@ pub fn layout_frac(
...
+        FracStyle::Skewed => {
+            layout_skewed_frac(ctx, styles, &elem.num, &elem.denom, elem.span())
+        }
+        FracStyle::Horizontal => layout_horizontal_frac(
+            ctx,
+            styles,
+            &elem.num,
+            &elem.denom,
+            elem.span(),
+            elem.num_deparenthesized.as_option().unwrap_or(false),
```
mkorje:
```suggestion
            elem.num_deparenthesized.get(styles),
```
This'll make use of the default set in the FracElem definition. Similar for denom below.

tjbcg:
> > I've implemented changes that prevent overlapping of each operand with the fraction bar. Here are some before and afters. Each operand is shifted horizontally to a point where there is no more overlap.
> 
> I think its fine to let them overlap, poor looking results are basically to be expected then in these cases. (We just want to ensure the final frame size is correct even in some weirder cases, I think. I'm thinking about this in a similar way to #6021. And to add, it'll be basically impossible to check once the fonts PR is done as the slash glyph will go through realisation.) Also, trying to check for a collision without computing its outline or smth is going to be fickle.

I don't agree that it's fine to let them overlap. Poor looking results are indeed to be expected but I don't want my typesetting system to fail at the single task it has in this case, which is to place the numerator and denominator each on one side of the bar. As a user, I'd rather correct for extra space than overlap (same argument goes for other fraction styles, I would never expect a vertical fraction to fail at this even with e.g. the most outlandish denominator).

We can drop this if you feel like it'll be too hard to maintain, but I think the current code is a good compromise. It doesn't check for outline or anything, it just ensures each numerator/denominator frame doesn't intersect (except at a single point, they can touch but not cross) with the diagonal of the frame of the slash glyph (I think it's a reasonable assumption that the slash glyph will be fairly close to that diagonal). With the way the code is written we get this check "for free" (it's a simple matter of calculating two dot products), and it prevents catastrophic failure of typesetting the fraction. If an argument is to be made about fonts having unusual slashes or badly defined glyphs, then layout failure was to be expected anyway regardless of this extra check.

As for the font PR, since we're not calculating outlines, I think this will keep working ? Let me know, you know the PR more than I do (obviously).

I'll patch in the changes you suggested in review.

crates/typst-layout/src/math/frac.rs
```diff
@@ -231,12 +236,14 @@ fn layout_skewed_frac(
...
+    let pre_stretch_height = slash_frag.size().y;
     slash_frag.stretch_vertical(ctx, fraction_height - short_fall);
     // If the standard slash was not stretchable, try the fraction slash
-    if slash_frag.size.y == pre_stretch_height {
-        slash_frag = GlyphFragment::new_char(ctx.font, styles, '/', span)?;
+    if slash_frag.size().y == pre_stretch_height {
+        slash_frag =
+            ctx.layout_into_fragment(&SymbolElem::packed('/').spanned(span), styles)?;
         slash_frag.stretch_vertical(ctx, fraction_height - short_fall);
     }
```
mkorje: Checking the height after stretching against the `pre_stretch_height` could be a bit problematic: if the target height is the same size as the unstretched glyph, we will unnecessarily switch to the normal slash. 

I think we shouldn't bother with trying to do any fallback here. The example with the fraction slash U+2044 not being stretchable in STIX Two Math is a font issue (and has been fixed on the dev version).

crates/typst-library/src/text/font/mod.rs
```diff
@@ -458,6 +462,8 @@ impl FontMetrics {
                     fraction_denominator_gap_min: metrics.underline.thickness,
                     fraction_denom_display_style_gap_min: 3.0
                         * metrics.underline.thickness,
+                    skewed_fraction_vertical_gap: Em::zero(),
+                    skewed_fraction_horizontal_gap: metrics.x_height,
```
mkorje: Maybe using the width of the fraction slash U+2044 is a better default here?

mkorje: Also, could you add a comment with the others at the start of this `.unwrap_or`. Something like
> `skewed_fraction_vertical_gap` and `skewed_fraction_horizontal_gap` from our best guess

(An aside: I cannot find any other engine that has fallbacks for these two constants lol, so I guess we're stuck coming up with it ourselves.)

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,167 @@ fn layout_frac_like(
...
+        .pack()
+    } else {
+        num
+    };
+    let num_frame = ctx.layout_into_fragment(num, styles)?;
+    ctx.push(num_frame);
+
+    // AI
+    let mut slash =
+        ctx.layout_into_fragment(&SymbolElem::packed('/').spanned(span), styles)?;
```
mkorje, line 189: I do wonder whether the "better" (as in, more semantically correct) glyph to use here is the division slash U+2215. Though given every font displays it slightly differently (see below) to the solidus U+002F and the solidus is what unicode-math uses (`\mathslash`), it doesn't make much sense.

<details>

<summary>solidus vs division slash</summary>

```typ

#let div = math.class("normal", symbol("\u{2215}"))
#let test = $ a div b quad "vs" quad a \/ b \ E div QQ quad "vs" quad E \/ QQ $

#let fonts = (
  "Cambria Math",
  "Latin Modern Math",
  "New Computer Modern Math",
  "Noto Sans Math",
  "STIX Two Math",
  "Euler Math",
)

#for font in fonts {
  [#font]
  show math.equation: set text(font: font)
  test
}
```

test

</details>

mkorje:
> I don't agree that it's fine to let them overlap. Poor looking results are indeed to be expected but I don't want my typesetting system to fail at the single task it has in this case, which is to place the numerator and denominator each on one side of the bar. As a user, I'd rather correct for extra space than overlap (same argument goes for other fraction styles, I would never expect a vertical fraction to fail at this even with e.g. the most outlandish denominator).
>
> We can drop this if you feel like it'll be too hard to maintain, but I think the current code is a good compromise. It doesn't check for outline or anything, it just ensures each numerator/denominator frame doesn't intersect (except at a single point, they can touch but not cross) with the diagonal of the frame of the slash glyph (I think it's a reasonable assumption that the slash glyph will be fairly close to that diagonal). With the way the code is written we get this check "for free" (it's a simple matter of calculating two dot products), and it prevents catastrophic failure of typesetting the fraction. If an argument is to be made about fonts having unusual slashes or badly defined glyphs, then layout failure was to be expected anyway regardless of this extra check.

I guess the difference is that in this case the engine's job is to place the numerator and denominator `skewed_fraction_horizontal_gap` far apart, ignoring the slash entirely. Whereas in the layout of the vertical fraction style the engine is placing things above and below the fraction line.

I don't think it'd be particularly hard to maintain (thought it does add some complexity), I'm just a bit concerned with making such an assumption. Could this check cause the intended output to change? For example, a part of the glyph's frame could cross the slash but not actually visually cross it; correcting for extra space would then change the "intended" layout. The slash could be also fairly thick in a font, and so there could still be a collision even with this check, I think. And I think if we want to perform a check like this then we should it "properly": get the outline of the slash and use that to check for collisions. We can ask for @laurmaedje input on this as well.

> As for the font PR, since we're not calculating outlines, I think this will keep working ? Let me know, you know the PR more than I do (obviously).

With the fonts PR: I meant to say that since it now goes through realisation the output is opaque (i.e. it may not necessarily be a `GlyphFragment`).

crates/typst-library/src/text/font/mod.rs
tjbcg: ConTeXt uses math_quad/2 as a fallback for horizontal gap, but I'm not familiar enough with the codebase to know how to get that value in here. Help appreciated!

crates/typst-layout/src/math/frac.rs
tjbcg: In the example you provided I definitely prefer the solidus over the division slash (esp. in NewCM). I also feel like the expectation with horizontal fractions is to behave the same as `a slash b`.

tjbcg:
I've removed the check for collisions. You're right that it will be a pain with some bounding boxes (e.g. tall denominator that starts with a smaller element will get displaced unnecessarily). If this ever turns out to be an issue we can patch it in later anyway.

crates/typst-library/src/text/font/mod.rs
mkorje: Actually using 0.5em as ConTeXt does looks quite decent (though using x_height for the vertical gap seems far too large). So maybe we just do that. Here's Libertinus Serif (so we use the fallback values) with various values.

vgap: 0.5em, hgap: x_height
playground

vgap: 0.5em, hgap: 0em
playground

crates/typst-layout/src/math/frac.rs
mkorje: Yeah me too. 

> I also feel like the expectation with horizontal fractions is to behave the same as `a slash b`.

Yes fair point.
(For some reason I can't find any button to mark this as resolved...)

crates/typst-library/src/math/frac.rs
```diff
@@ -1,9 +1,17 @@
...
 use crate::math::Mathy;
 
+/// Fraction style
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Cast)]
+pub enum FracStyle {
+    Vertical,
+    Skewed,
+    Horizontal,
+}
+
```
mkorje:
```suggestion
/// Fraction style
#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash, Cast)]
pub enum FracStyle {
    /// Stacked numerator and denominator with a bar.
    #[default]
    Vertical,
    /// Numerator and denominator separated by a slash.
    Skewed,
    /// Numerator and denominator placed inline and parentheses are not absorbed.
    Horizontal,
}

```
Adding the docs here should make sure it ends up like https://typst.app/docs/reference/model/ref/#parameters-form in the docs.

crates/typst-library/src/math/frac.rs
```diff
@@ -26,6 +34,28 @@ pub struct FracElem {
     /// The fraction's denominator.
     #[required]
     pub denom: Content,
+
+    /// The style of the fraction.
+    /// - `"vertical"`: stacked numerator and denominator with a bar.
+    /// - `"diagonal"`: numerator and denominator separated by a slash.
+    /// - `"horizontal"`: numerator and denominator placed inline and
+    ///   parentheses are not absorbed
```
mkorje:
```suggestion
```

crates/typst-library/src/text/font/mod.rs
tjbcg: Did you swap vgap and hgap there ?

Edit: assuming you did, I used these values in the latest commit (vgap: 0em, hgap: 0.5em). I'll resolve this once this is clarified

mkorje: Whoops yes I did, thanks for catching that!

crates/typst-library/src/math/frac.rs
```diff
@@ -26,6 +38,24 @@ pub struct FracElem {
     /// The fraction's denominator.
     #[required]
     pub denom: Content,
+
+    /// The style of the fraction.
```
laurmaedje: A bit more documentation, including an example, would be nice here.

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,128 @@ fn layout_frac_like(
...
+    let mut fraction_frame = Frame::soft(Size::new(fraction_width, fraction_height));
+
+    // Baseline (use axis height to center slash on the axis)
+    fraction_frame.set_baseline(fraction_height / 2.0 + axis);
+
+    // Debugging help
+    // num_frame.mark_box_in_place();
+    // denom_frame.mark_box_in_place();
+    // slash_frame.mark_box_in_place();
+    // fraction_frame.mark_box_in_place();
```
laurmaedje: This can be removed.

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,128 @@ fn layout_frac_like(
...
+    num: &Content,
+    denom: &Content,
+    span: Span,
+) -> SourceResult<()> {
+    // Font-derived constants
+    let constants = ctx.font().math();
+    let vgap = constants.skewed_fraction_vertical_gap.resolve(styles);
+    let hgap = constants.skewed_fraction_horizontal_gap.resolve(styles);
+    let axis = constants.axis_height.resolve(styles);
+    // let vgap = vgap + axis / 2.0; // LuaTeX behavior
```
laurmaedje: I'm not sure what this line is supposed to tell me.

crates/typst-library/src/math/frac.rs
```diff
@@ -1,9 +1,21 @@
 use typst_syntax::Spanned;
 
 use crate::diag::bail;
-use crate::foundations::{Content, Value, elem};
+use crate::foundations::{Cast, Content, Value, elem};
 use crate::math::Mathy;
 
+/// Fraction style
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash, Cast)]
+pub enum FracStyle {
```
laurmaedje: please move this below `FracElem`.

crates/typst-syntax/src/ast.rs
```diff
@@ -840,6 +840,13 @@ impl<'a> Math<'a> {
     pub fn exprs(self) -> impl DoubleEndedIterator<Item = Expr<'a>> {
         self.0.children().filter_map(Expr::cast_with_space)
     }
+
+    /// Whether this `Math` node was originally parenthesized.
+    pub fn was_deparenthesized(self) -> bool {
+        let mut iter = self.0.children();
+        matches!(iter.next().map(SyntaxNode::kind), Some(SyntaxKind::LeftParen))
+            && matches!(iter.last().map(SyntaxNode::kind), Some(SyntaxKind::RightParen))
```
laurmaedje:
```suggestion
            && matches!(iter.next_back().map(SyntaxNode::kind), Some(SyntaxKind::RightParen))
```

tests/suite/math/frac.typ
```diff
@@ -45,3 +45,19 @@ $ a_1/b_2, 1/f(x), zeta(x)/2, "foo"[|x|]/2 \
 --- math-frac-gap ---
 // Test that the gap above and below the fraction rule is correct.
 $ sqrt(n^(2/3)) $
+
```
laurmaedje, line 48: A few more tests, also in inline math would be nice.

laurmaedje: could you also add tests with explicit `frac` calls to make sure it behaves as intended with the deparenthesization logic?

laurmaedje (commented):
On the layout logic, I would trust @mkorje. Otherwise, I have a few remarks, but nothing major.

tests/suite/math/frac.typ
laurmaedje: also maybe tests with non-paren brackets

laurmaedje:
From my end, this looks good to merge. 

@mkorje Any further input?

crates/typst-library/src/math/frac.rs
```diff
@@ -26,6 +26,42 @@ pub struct FracElem {
...
+
+/// Fraction style
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash, Cast)]
+pub enum FracStyle {
+    /// Stacked numerator and denominator with a bar.
+    #[default]
+    Vertical,
+    /// Numerator and denominator separated by a slash.
+    Skewed,
+    /// Numerator and denominator placed inline and parentheses are not absorbed
```
mkorje:
```suggestion
    /// Numerator and denominator placed inline and parentheses are not absorbed.
```

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,121 @@ fn layout_frac_like(
...
+    let mut denom_up_left = num_up_left + num_size.to_point() + Point::new(hgap, vgap);
+
+    // Fraction width
+    let mut slash_up_left = slash_center - slash_size.to_point() / 2.0;
+    let fraction_width = (denom_up_left.x + denom_size.x)
+        .max(slash_center.x + slash_size.x / 2.0)
+        - num_up_left.x.min(slash_up_left.x);
+    // We have to shift everything right to avoid going in the negatives for the x coordinate
+    let horizontal_offset =
+        Point::with_x(Abs::zero().max(num_up_left.x - slash_up_left.x));
```
mkorje:
```suggestion
        Point::with_x(Abs::zero().max(-slash_up_left.x));
```
`num_up_left.x` is always zero at this point, so I think this is a little clearer.

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,121 @@ fn layout_frac_like(
...
+    // Reference points for all three objects, used to place them in the frame.
+    let slash_center = Point::new(num_size.x + hgap / 2.0, fraction_height / 2.0);
+    let mut num_up_left = Point::with_y(vertical_offset);
+    let mut denom_up_left = num_up_left + num_size.to_point() + Point::new(hgap, vgap);
+
+    // Fraction width
+    let mut slash_up_left = slash_center - slash_size.to_point() / 2.0;
+    let fraction_width = (denom_up_left.x + denom_size.x)
+        .max(slash_center.x + slash_size.x / 2.0)
+        - num_up_left.x.min(slash_up_left.x);
```
mkorje:
```suggestion
        + Abs::zero().max(-slash_up_left.x);
```
Same here then, or you could just add the horizontal_offset to the width further below.

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,121 @@ fn layout_frac_like(
...
+
+    // Reference points for all three objects, used to place them in the frame.
+    let slash_center = Point::new(num_size.x + hgap / 2.0, fraction_height / 2.0);
+    let mut num_up_left = Point::with_y(vertical_offset);
+    let mut denom_up_left = num_up_left + num_size.to_point() + Point::new(hgap, vgap);
+
+    // Fraction width
+    let mut slash_up_left = slash_center - slash_size.to_point() / 2.0;
+    let fraction_width = (denom_up_left.x + denom_size.x)
+        .max(slash_center.x + slash_size.x / 2.0)
```
mkorje:
```suggestion

    let mut slash_up_left = Point::new(num_size.x + hgap / 2.0, fraction_height / 2.0) - slash_size.to_point() / 2.0;
    let mut num_up_left = Point::with_y(vertical_offset);
    let mut denom_up_left = num_up_left + num_size.to_point() + Point::new(hgap, vgap);

    // Fraction width
    let fraction_width = (denom_up_left.x + denom_size.x)
        .max(slash_up_left.x + slash_size.x)
```
Maybe removing `slash_center` is a little nicer, since it's otherwise not being used elsewhere.

crates/typst-layout/src/math/frac.rs
```diff
@@ -143,3 +161,121 @@ fn layout_frac_like(
...
+
+    // Build the slash glyph to calculate its size
+    let mut slash_frag =
+        ctx.layout_into_fragment(&SymbolElem::packed('\u{2044}').spanned(span), styles)?;
+    slash_frag.stretch_vertical(ctx, fraction_height - short_fall);
+    slash_frag.center_on_axis();
+    let slash_frame = slash_frag.into_frame();
+
+    // Adjust the fraction height if the slash overflows
+    // Fraction width will be re-calculated later on after we adjusted the x values to avoid
```
mkorje: Could you wrap the comments at 80 chars? (IIRC that's the convention used in this repo) There's also one other comment further below too.

laurmaedje:
Thank you!

PR #6674, opened 2025-07-29
Document escaping semicolon, valid identifiers, and `state` tips

### LINKED ISSUE #2680: Request: more detailed documentation on the behavior of multiple `state` calls with the same key
The current documentation of `State` is very ambiguous on how exactly multiple calls to `state` with the same key behave. 
image
My testing suggests that:
- the underlying "actual state", which is uniquely identified by its `key`, is effectively a series of updates, which are either functions of one argument or values that aren't functions; values that aren't functions effectively behave identically to constant functions of one argument (i.e., functions which just ignore their one argument and always return the same thing).
- when you access a state _through any particular handle_, created by a call to `state(key)` with an optional initial value, it invokes the state's updates up to the appropriate location in order on the initial value of the handle.
This means that calls to `update` don't care which handle you use for them, only the `key` of that handle, while calls to `display()` and similar will all depend on which handle you use. For example, the below code displays:
```
#let handle1 = state("test", 1)
#let handle2 = state("test", 2)
#handle1.display()
#handle2.display()
#handle1.update(it => it + 10)
#handle2.update(it => it + 10)
#handle1.display()
#handle2.display()
#state("test").update(3)
#handle1.display()
```
1, 2, 21, 22, 3 (note that both updates applied to the state accessed through both handles, but even after updating, the values are distinct)

I'm not sure how best to summarize this, but it should probably be somewhere in the docs.

YDX-2147483647:
> [!NOTE]
> I'm not sure if it's appropriate to accumulate four docs updates in a single PR. I can break it down into several PRs if you wish to review separately. :heart:

The first two changes were discussed in the [_Documentation_ forge on Discord](https://discord.com/channels/1054443721975922748/1399326777540608041).

- **Semicolon after expressions can be escaped**

  [Suggested by Mentonin on Discord](https://discord.com/channels/1054443721975922748/1399326777540608041/1399464816472756396)

- **Valid identifiers**

  [Suggested by Mentonin on Discord](https://discord.com/channels/1054443721975922748/1399326777540608041/1399459782473285875), and previously mentioned in https://github.com/typst/typst/pull/6008/files#diff-cf14f1b1e25043f3bb5e7210782f1b837483d5ed4936315256caabbd624b8821R10.

  Ref: [`is_ident` in typst_syntax - Rust](https://docs.rs/typst-syntax/latest/typst_syntax/fn.is_ident.html)

- **The behavior of multiple `state`s with the same key**

  Resolves #2680
  
  (This part should be skipped/updated if #6697 will be merged)

- **Updating a state with a function is better than using `context`**

  Suggested in the Chinese group chat.

  `fill.update(f => not f)` is better than `context fill.update(not fill.get())`. The latter leads to _Layout did not converge within 5 attempts_ easily.

Please _squash_ and merge.

## Preview of changed pages

- [`/reference/scripting/`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/scripting/)
  - [`#expressions`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/scripting/#expressions)
  - [`#bindings`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/scripting/#bindings)
  - [`#identifiers`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/scripting/#identifiers)
- [`/reference/foundations/label/#constructor-name`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/foundations/label/#constructor-name)
- [`/reference/introspection/state/#constructor`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/introspection/state/#constructor)
  `#definitions-update-update` has been changed too, but has not been deployed to netlify yet.
- [`/reference/introspection/counter/#constructor-key`](https://688ef46a31e7970008128d77--ydx-typst.netlify.app/docs/reference/introspection/counter/#constructor-key)

### DIFF
--- a/docs/reference/language/scripting.md
+++ b/docs/reference/language/scripting.md
@@ -14,7 +14,7 @@ provides compact syntax to embed a code expression into markup: An expression is
 introduced with a hash (`#`) and normal markup parsing resumes after the
 expression is finished. If a character would continue the expression but should
 be interpreted as text, the expression can forcibly be ended with a semicolon
-(`;`). You can [escape literal `#` or `;` with a backslash]($syntax/#escapes).
+(`;`). You can [escape a literal `#` or `;` with a backslash]($syntax/#escapes).
 
 ```example
 #emph[Hello] \
@@ -338,40 +338,6 @@ all available community packages on [Typst Universe]($universe).
 If you are using Typst locally, you can also create your own system-local
 packages. For more details on this, see the
 [package repository](https://github.com/typst/packages).
-
-## Names and identifiers { #identifiers }
-When developing modules and packages, it is recommended to use
-[kebab case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case)
-for identifiers of public variables, functions, and their arguments.
-In kebab case, words are written in lowercase and separated by hyphens (`-`),
-like `cap-height`.
-
-The identifier syntax in Typst is based on the
-[Unicode Standard Annex #31](https://www.unicode.org/reports/tr31/),
-with two extensions: allowing `_` as a starting character, and allowing both `_`
-and `-` as continuing characters.
-
-Specifically:
-
-- Identifiers may start with a letter or an underscore (`_`),
-- and continue with letters, digits, underscores, and hyphen (`-`).
-
-```example
-#let kebab-case = [Using hyphen]
-#let _schön = "😊"
-#let 始料不及 = "😱"
-#let π = calc.pi
-
-#kebab-case
-#if -π < 0 { _schön } else { 始料不及 }
-// -π means -1 * π,
-// so it's not a valid identifier
-```
-
-Note that the above only applies to identifiers. [`label`]($label), however,
-is not an identifier. Therefore, [its dedicated syntax]($label/#syntax) allows
-the name to start with `-`, and also allows the use of `:` and `.`. Furthermore,
-[its constructor function]($label/#constructor) allows any non-empty string.
 
 ## Operators
 The following table lists all available unary and binary operators with effect,
--- a/crates/typst-library/src/introspection/state.rs
+++ b/crates/typst-library/src/introspection/state.rs
@@ -274,16 +274,17 @@ impl State {
     pub fn construct(
         /// The key that identifies this state.
         ///
-        /// Subsequent [`update`]($state.update)s to the state will be uniquely
-        /// identified by this string key, and irrelevant to the variable that carries
-        /// the `state`. That is, if you construct multiple variables with the same
+        /// Any [updates]($state.update) to the state will be identified with
+        /// the string key. If you construct multiple states with the same
         /// `key`, then updating any one will affect all of them.
         key: Str,
         /// The initial value of the state.
         ///
-        /// If you construct multiple variables with the same `key` but different
-        /// `init`, then each variable has its own initial value, although they
-        /// share updates.
+        /// If you construct multiple states with the same `key` but different
+        /// `init` values, they will each use their own initial value but share
+        /// updates. Specifically, the value of a state at some location in the
+        /// document will be computed from that state's initial value and all
+        /// preceding updates for the state's key.
         ///
         /// ```example
         /// #let banana = state("key", "🍌")
@@ -350,7 +351,7 @@ impl State {
         Ok(sequence.last().unwrap().clone())
     }
 
-    /// Update the value of the state.
+    /// Updates the value of the state.
     ///
     /// The update will be in effect at the position where the returned content
     /// is inserted into the document. If you don't put the output into the
@@ -358,24 +359,30 @@ impl State {
     /// write `{let _ = state("key").update(7)}`. State updates are always
     /// applied in layout order and in that case, Typst wouldn't know when to
     /// update the state.
+    ///
+    /// In contrast to [`get`]($state.get), [`at`]($state.at), and
+    /// [`final`]($state.final), this function does not require [context].
     #[func]
     pub fn update(
         self,
         span: Span,
-        /// - If given a non function-value, sets the state to that value.
-        /// - If given a function, that function receives the previous state and
-        ///   has to return the new state.
+        /// A value to update to or a function to update with.
         ///
-        /// When updating the state based on its previous value, please update
-        /// with a function instead of retrieving the previous value from the
-        /// [context]($context). This allows the compiler to resolve the final
-        /// state efficiently, minimizing the number of [iterations]($context/#compiler-iterations)
-        /// required.
+        /// - If given a non-function value, sets the state to that value.
+        /// - If given a function, that function receives the state's previous
+        ///   value and has to return the state's new value.
+        ///
+        /// When updating the state based on its previous value, you should
+        /// prefer the function form instead of retrieving the previous value
+        /// from the [context]($context). This allows the compiler to resolve
+        /// the final state efficiently, minimizing the number of
+        /// [layout iterations]($context/#compiler-iterations) required.
         ///
         /// In the following example, `{fill.update(f => not f)}` will paint odd
-        /// [items in the bulletin list]($list.item) as expected. However, if it's
+        /// [items in the bullet list]($list.item) as expected. However, if it's
         /// replaced with `{context fill.update(not fill.get())}`, then layout
-        /// will not converge within 5 attempts.
+        /// will not converge within 5 attempts, as each update will take one
+        /// additional iteration to propagate.
         ///
         /// ```example
         /// #let fill = state("fill", false)
--- a/docs/reference/language/syntax.md
+++ b/docs/reference/language/syntax.md
@@ -168,6 +168,34 @@ I got an ice cream for
 \$1.50! \u{1f600}
 ```
 
+## Identifiers
+Names of variables, functions, and so on (_identifiers_) can contain letters,
+numbers, hyphens (`-`), and underscores (`_`). They must start with a letter or
+an underscore.
+
+More specifically, the identifier syntax in Typst is based on the
+[Unicode Standard Annex #31](https://www.unicode.org/reports/tr31/), with two
+extensions: Allowing `_` as a starting character, and allowing both `_` and `-`
+as continuing characters.
+
+For multi-word identifiers, the recommended case convention is
+[Kebab case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case). In Kebab
+case, words are written in lowercase and separated by hyphens (as in
+`top-edge`). This is especially relevant when developing modules and packages
+for others to use, as it keeps things predictable.
+
+```example
+#let kebab-case = [Using hyphen]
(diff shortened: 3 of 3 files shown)


docs/reference/language/scripting.md
```diff
@@ -14,7 +14,7 @@ provides compact syntax to embed a code expression into markup: An expression is
 introduced with a hash (`#`) and normal markup parsing resumes after the
 expression is finished. If a character would continue the expression but should
 be interpreted as text, the expression can forcibly be ended with a semicolon
-(`;`).
+(`;`). As usual, you can [escape literal `#` or `;` with a backslash]($syntax/#escapes).
```
Andrew15-5: "As usual" sounds a bit out of place to me.

docs/reference/language/scripting.md
```diff
@@ -338,6 +339,34 @@ If you are using Typst locally, you can also create your own system-local
 packages. For more details on this, see the
 [package repository](https://github.com/typst/packages).
 
+## Names and identifiers { #identifiers }
+When developing modules and packages, it is recommended to use kebab-case for
+identifiers of public variables, functions, and their arguments. In kebab-case,
```
Andrew15-5: I don't think you have to keep the dash in the case name. Just as you don't have to call it snake_case or PascalCase. It is however helpful to give a second purpose of showing what you mean by giving an example. But that is probably more of a conversational thing. You already have a dedicated example, so just write "kebab case", or probably "Kebab case", since it's a name.

docs/reference/language/scripting.md
```diff
@@ -338,6 +339,34 @@ If you are using Typst locally, you can also create your own system-local
...
+
+The identifier syntax in Typst is based on the
+[Unicode Standard Annex #31](https://www.unicode.org/reports/tr31/),
+with two extensions: allowing `_` as a starting character, and allowing both `_`
+and `-` as continuing characters.
+
+Specifically: 
+
+- Identifiers may start with a letter or an underscore (`_`),
+- and continue with letters, digits, underscores, and hyphen (`-`) thereafter.
```
Andrew15-5: "Thereafter" sounds too formal? You already say "continue with", so this might be redundant to begin with.

docs/reference/language/scripting.md
```diff
@@ -338,6 +339,34 @@ If you are using Typst locally, you can also create your own system-local
...
+
+```example
+#let kebab-case = [Using hyphen]
+#let _schön = "😊"
+#let 小四 = 12pt
```
Andrew15-5: What is the "small four"? No way this is a "font size"?

docs/reference/language/scripting.md
```diff
@@ -338,6 +339,34 @@ If you are using Typst locally, you can also create your own system-local
...
+#let π = calc.pi
+
+#kebab-case
+#if -π < 0 { _schön }
+```
+
+Note that this does not apply to names of [`label`]($label), which follows
+[a dedicated syntax]($label/#syntax).
```
Andrew15-5: I think the topic is "names" (label names have a dedicated syntax), so using "follows" is incorrect. Should be "follow". At least that's how I read this. Furthermore, you can have [almost any](https://github.com/typst/typst/pull/6332) label name. But with syntactic sugar you are slightly limited.

This statement should be a bit more specific or selective about which words are used to keep it more accurate.

MDLC01:
Another way to "escape" semicolons is by using two of them (one to close the expression, and one that is interpreted as markup). This may be easier to type than `\;`, so maybe this trick should be mentioned somewhere as well.

docs/reference/language/scripting.md
xkevio: Based on this: https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules, I would stay with "kebab-case" for consistency.

YDX-2147483647: > for consistency

Yes, the sentence is basically copied from package-check.

https://github.com/typst/package-check/blob/c1c1d1ee9d164a8abb20b507651dba6ce15b108d/src/check/kebab_case.rs#L87
https://github.com/typst/package-check/blob/c1c1d1ee9d164a8abb20b507651dba6ce15b108d/src/check/kebab_case.rs#L113-L114

How about using kebab-case and deleting the next sentence (_In kebab-case, words are…_)? Because the code example also explains it.

```typ
#let kebab-case = [Using hyphen]
```

docs/reference/language/scripting.md
YDX-2147483647: It's a system to specify text size in Chinese and Japanese typesetting.

[clreq-gap for typst](https://typst-doc-cn.github.io/clreq/#chinese-size-system-hao-system):
> 号数制 (hào-system, or number system) is the system to specify text size in Chinese and Japanese typesetting. The unit is not linearly related to points: 五号 (size 5) = 10.5 pt, 小四 (size small 4) = 12 pt, 四号 (size 4) = 14 pt… The hào-system is still commonly used in China. An average person may be familiar with the size of 小四, but likely has little intuition about 12 pt without referring to a conversion table..
>
> This issue should have been marked as Basic. However, considering that the hào-system was not standardized by the various foundries in the past (for example, nowadays, [四号 is 14 pt in CTeX, MS Word, WPS, and Adobe](https://github.com/YDX-2147483647/typst-pointless-size/blob/main/docs/ref.md), but others may prefer 13.75 pt), we mark it as Advanced for the moment.

There are many templates actually writing some kind of `#let 小四 = 12pt`. ([GitHub Search](https://github.com/search?q=小四+12pt+language:typst&ref=opensearch&type=code))

```typ
#let font-size = (
  四号: 14pt,
  小四: 12pt,
  …
)
#set text(size: font-size.小四)
```

> What is the "small four"?

How about changing it to `#let 五号 = 10.5pt`? _No. 5_ might be less confusing, but 号/號 leads to the problem of Simplified/Traditional Chinese.

YDX-2147483647:
Thank you all for your suggestions!
For each case, I have adopted what I personally consider the best version, but further discussion is still acceptable.

I plan to keep this PR as a draft for a week and accumulate more changes. Otherwise, I'm afraid we might clutter the git log with numerous small docs fixes. 

> Another way to "escape" semicolons is by using two of them (one to close the expression, and one that is interpreted as markup). This may be easier to type than `\;`, so maybe this trick should be mentioned somewhere as well.

I agree, but I don't know where to add it.
Now it's added to [`$scripting/#expressions`](https://typst.app/docs/reference/scripting/#expressions). Here the focus is not on escaping, and we can't say too much on it. So I prefer the way that works for both `#` and `;`.

I've also considered [`$syntax/#escapes`](https://typst.app/docs/reference/syntax/#escapes). But the chapter is before _Scripting_ and it is hard to give examples without scripting (including `#emoji`).

docs/reference/language/scripting.md
YDX-2147483647: I asked for others' opinions. Now I drop things related to the font size and change it to the following.

```typ
#let 始料不及 = "😱"
#if -π < 0 { _schön } else { 始料不及 }
```

始料不及 means _unexpected_, and these four characters are same in simplified and traditional forms respectively.

docs/reference/language/scripting.md
Andrew15-5: I mean, the example doesn't actually explain it, but gives an idea of how it could look like. There is nothing that says "Kebab case means lowercase with hyphens". _Wikipedia_, on the other hand, does explain it, and [is allowed to be cited](https://github.com/Andrew15-5/typst/blob/78355421ad73fdcbe93b4acca890b439c4b6f98d/docs/reference/groups.yml#L173).

I guess without changing the other places, you kinda have to write it the same way. They all can be replaced later. But also, it is heavily tied to the context. Where this is not the primary topic, and it's not technical, then usually you format the case name the way you would write the said name using the said casing. Here, I think this is rather technical and is the/one of the main topics. And if "Kebas case" section is cited, it makes even more sense to preserve the section name.

docs/reference/language/scripting.md
Andrew15-5: > Note that this does not apply to names of [`label`]($label), which follow [a dedicated syntax]($label/#syntax).

This still omits the fact that there is no dedicated syntax when it comes to label constructor. It's just that simply writing "[this](#this) and [this](#this)" is not good. A more elaborate way should be used.

YDX-2147483647: How about this?

```markdown
Note that the above only applies to identifiers. For example, [`label`]($label)s
are not identifiers. Therefore, [its dedicated syntax]($label/#syntax) allows
the name to start with `-`, and also allows the use of `:` and `.`. Furthermore,
[its constructor function]($label/#constructor) allows any non-empty string.
```

docs/reference/language/scripting.md
YDX-2147483647: I've updated this. Could you take a look again? Thank you!

docs/reference/language/scripting.md
Andrew15-5:
```md
Note that the above only applies to identifiers. [`label`]($label), however,
is not an identifier. Therefore, [its dedicated syntax]($label/#syntax) allows
the name to start with `-`, and also allows the use of `:` and `.`. Furthermore,
[its constructor function]($label/#constructor) allows any non-empty string.
```

I have a feeling that Laurenz might not like non-capitalized beginning of a sentence, but we'll see. "For example" implies that the label is an example of an identifier, which it's not. So that introduction doesn't work.

Also, "labels" with monospaced font and link is a bit messy, and can be formatted in many different ways. Better to keep it singular, especially since "its" used to refer to it.

docs/reference/language/scripting.md
Andrew15-5: Not exactly how I imagined, but it's good enough. I'm not the one who has write access, just someone that has created and closed numerous docs issues.

Andrew15-5:
If there are multiple separate changes, it does take additional effort to visually and mentally split the diff into these changes. If they are not coupled, it's better to split them. The state changes appeared out of nowhere. The first two changes were somewhat loosely connected, at least they were in the same file. Though now that I look at it, the semicolon change is just one line. Thus, it would be super easy to review this particular change, and not depend on the other two much bigger changes.

Besides, it's cooler to have 3 merged PRs rather than just one.

I and Laurenz did some docs polishing/touch up PRs that cover different things, but it was still a different kind of change and usually during the testing stage before release.

laurmaedje:
Generally speaking, I'd agree that separate PRs are better.

YDX-2147483647:
I see. I won't accumulate big PR in the future.
But does this specific PR need to be split?

laurmaedje:
No, it's okay.

laurmaedje:
I tweaked the updates a bit. In particular, I moved the identifier stuff from "Scripting" to "Syntax". I also decided to drop the label part as it felt out of place and is already documented in the label documentation.

Thanks!

PR #6676, opened 2025-07-29
Allow custom element names in HTML tag syntax 

### LINKED ISSUE #6675: Support Custom HTML Element
## Custom HTML Element
Custom HTML Element[^1] is introduced in HTML5, which allows user to create their own custom html element like
```html
<my-custom-tag> </my-custom-tag>
```

[^1]: https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name

### Use Case

it will be useful with the Typst html output feature (especially with `html.elem`), one can use it as a placeholder and do more post-processing in further process. for example, the 

```typ
#html.elem("my-custom-tag")
```
will produce `<my-custom-tag>` mentioned above. and one can use 3-rd party tools to replace it with html contents.[^2]

However, currently, it is blocked by the naming check, `the character "-" is not valid in a tag name` error will be reported if you try to do this.
[^2]: Although you can use `<div>` even `<slot>` with `id` or `name` attribute identifier, as a placeholder. but the custom elements are also necessary in my opinion

uwni:
Closes #6675

### DIFF
--- a/crates/typst-html/src/dom.rs
+++ b/crates/typst-html/src/dom.rs
@@ -130,7 +130,7 @@ impl HtmlTag {
             }
         }
 
-        let is_start_with_lowercase = string.starts_with(|c| matches!(c, 'a'..='z'));
+        let is_start_with_lowercase = string.starts_with(|c: char| c.is_ascii_lowercase());
         let is_reserved = matches!(
             string,
             "annotation-xml"
@@ -355,9 +355,9 @@ impl HtmlFrame {
 
 //unit tests
 #[cfg(test)]
-mod tests{
+mod tests {
     use super::*;
-    
+
     #[test]
     fn test_custom_element_validation() {
         // Valid custom element names
--- a/crates/typst-html/src/dom.rs
+++ b/crates/typst-html/src/dom.rs
@@ -105,6 +105,22 @@ impl HtmlTag {
             bail!("tag name must not be empty");
         }
 
+        let mut has_hyphen = false;
+        let mut has_uppercase = false;
+
+        for c in string.chars() {
+            if c == '-' {
+                has_hyphen = true;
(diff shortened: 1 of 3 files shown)


crates/typst-html/src/dom.rs
```diff
@@ -306,3 +353,36 @@ impl HtmlFrame {
...
+    fn test_custom_element_validation() {
+        // Valid custom element names
+        assert!(HtmlTag::intern("my-element").is_ok());
+        assert!(HtmlTag::intern("custom-button").is_ok());
+        assert!(HtmlTag::intern("multi-word-component").is_ok());
+        assert!(HtmlTag::intern("x-foo").is_ok());
+        assert!(HtmlTag::intern("element-").is_ok());
+
+        // Invalid custom element names - no hyphen
+        assert!(HtmlTag::intern("customelement").is_ok()); // Regular HTML element
```
andreasKroepelin: Shouldn't this be `is_err`?

uwni: for line 369, the custom elements ending with hyphens are valid by the standards, you can confirm it with your browser console easily.

for line 372, it is the current behaviour, I don't know if I should allow only the typst `html.elem`s which satisfies html custom elements' [naming rules](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name). that means a restriction of the current ability.

uwni: for line 372, if the allowed tags get limited within valid html custom elements, it means only valid original html elements (like a, p, div...) and valid custom html elements can be created. I am not sure if it is acceptable, so if you have some idea or advice, please tell me <3

andreasKroepelin: Ah I see. I was just confusesd because the comment above starts with "Invalid...".

laurmaedje:
Thanks!

PR #6678, opened 2025-07-30
Use `rustc-hash` for hash maps and sets

laurmaedje:
This gives 2-3% speedup on average documents. I also [switched to fxhash in comemo](https://github.com/typst/comemo/pull/15) (not yet part of this PR), which gives _another_ 2-5% on top.


bluss:
Concern: fxhash is unmaintained. rustc-hash is probably the way to go? https://docs.rs/rustc-hash/

bluss:
just for info, which you probably already found, with maintenance comes changes so the algorithm is new. https://github.com/rust-lang/rustc-hash/blob/master/CHANGELOG.md  But that means a chance that it improves performance more?

laurmaedje:
I'll test it

laurmaedje:
Rustc hash looks good! In most cases, it performs pretty much the same as fxhash and in one case case it's even a bit faster (-0.5% instruction count, -1.5% wall clock time, but the clock time is less reliable, so take the second number with a grain of salt).

PR #6682, opened 2025-07-30
Add `cargo testit` alias for running integration tests

asibahi:
All this does is add a `cargo` alias for 

```sh
cargo test --workspace --test tests -- 
```

So any commands with `testit` like `testit math` can be called as 

```sh
cargo testit math
```

without the need for contributors to add alias to their shell.

Also updated the `tests/README.md`.


laurmaedje:
Thanks, this is nice!

PR #6685, opened 2025-07-31
Update guides welcome text to refer to list instead of specific guides

3w36zj6:
The welcome text for the Guides section previously mentioned only two guides, omitting the table guide. To address this and make future updates easier, the text now refers readers to the list below instead of mentioning specific guide names.

This ensures the welcome message stays accurate as more guides are added.


laurmaedje:
Thanks!

PR #6687, opened 2025-08-01
Ensure table headers trigger rowbreak

### LINKED ISSUE #6666: Regression in table.hline rendering when using table.header (introduced between 0.13.1 and 78355421)
### Description

A regression seem to have been introduced between stable version (typst 0.13.1 (8ace67d9)) and latest commit on main branch (78355421)), with regards to how stroke is handled for `table.hline` when `table.header` is used.

See minimal example below:

```typst
#table(
	columns: 2,
	table.hline(stroke: 2pt + blue),
	table.header(
		[*foo*],
		[*bar*],
	),
	table.hline(stroke: 1.5pt + red),
	table.cell(colspan: 2)[_asdf_],
	table.hline(stroke: 1.5pt + red),
	[a], [b],
	[c], [d],
)
```

Results when using stable version of Typst 0.13.1 (8ace67d9):

Image

Results when using latest commit on *main* branch (78355421):

Image

### Reproduction URL

https://typst.app/project/r1n2ghPzmQ2QW8fLKtctvE

### Operating system

_No response_

### Typst version

PgBiel:
Fixes #6666

I think it is more consistent if any auto cells after a header start to appear after it, even if the row above the header was incomplete. By extension, lines below the header should also appear below it. This is also consistent with the previous behavior - and, indeed, there was no other practical difference until now, since headers were always at the top. Now that they can be in other positions thanks to #6168, let's ensure some consistency here too and trigger rowbreaks immediately. Cells and lines below the header in the source code should therefore also appear below the header.

The exception is when the header only has fixed-position cells. Since those already ignore the auto position counter, the whole header also does, and therefore does not trigger a rowbreak (although it will prevent a conflict with a non-empty row through manual checks).


PgBiel:
This has made me realize that multiple headers aren't being automatically sorted, breaking the current O(1) check when skipping headers if headers are specified out of order. This is bad.

We could switch to using a heap to store headers, but that would turn access costly. Rather, I think we might want to sort again on every new header, thereby assuming the amount of auto cells outside headers is usually larger than the amount of headers (of course: usually there's just one header, versus a ton of cells). So a total cost of O(h^2 log h) to update headers (as we sort the header vector again for each header) sounds reasonable as h is usually small.

...Well, actually it would be O(h^2) I guess since finding where to insert the next header can be reduced to a linear search...

PgBiel:
Experimenting with using BTreeMap to store headers during grid resolving. Need some benchmarks, probably, but relevant operations are (letting h = number of headers):

- Insert new header: O(log h) basic insertion (total O(h log h))
   - VS O(h) => O(h^2) when using a simple vector.
- During insertion, check for short-lived headers before it: iteration O(h log h) , total O(h^2 log h)
   - VS O(h) , theoretical total O(h^2) when using a simple vector
      - Although this iteration usually ends after one step since short-lived headers are rare. But let's assume not...
   - But this is something that can be done through a second pass over headers after they are collected into a vector... So this can be improved and reduced to O(h) total (just one pass). So we can just ignore this.
- Skipping headers (auto cells):
  - To know if an auto cell looking for an empty spot is about to collide with a header, it only needs to check the next header.
  - "next_header" is now cached and updated on demand, so we only need to retrieve each header once.
  - If we skip a header, we need to find the next one in order to update "next_header" - this search costs O(log h), total O(h log h) (in the worst case we skip all h headers).
  - If a new header is added in a different position - before "next_header" - it automatically updates "next_header" without a search.
  - Therefore, total complexity here should be O(h log h).
  - VS O(1), total O(h), when using simple vector
- Skipping headers (non-auto cells):
  - No caching for these.
  - Therefore each such cell (currently) has to iterate over every header to determine whether or not to skip it...
  - In theory that'd be O(n * h log h) where n is the amount of non-auto cells.
      - VS O(n * h) when using simple vector (linear iteration)
  - But I think full BTreeMap iteration is faster than that. Might be amortized linear, which would cost O(n * h) as well.
  - Could also look into caching iteration results into a temporary vector if needed.
- Collecting BTreeMap into vector
  - Cost = zero when using simple vector
  - Cost = O(h log h)? for BTreeMap
     - Again, iteration cost might be amortized linear, but not sure

Of note, h is usually small, so in theory none of this would matter too much. But this is worth thinking about.

So, in principle the main potential problem is for non-auto cells. We should probably benchmark and compare these approaches..

PgBiel:
For the newest experiment, I had this idea of storing a reverse map row number -> boolean (there is a header ~~or footer~~ here or not), represented as a bitset where we insert all rows occupied by headers ~~or footers~~. This allows for the function to search for an available position for auto cells to efficiently check whether the current row is already occupied and, if so, skip it, without having to check headers ~~and footers~~ individually. This therefore also eliminates the need for the "next header" counter.

In addition, I also delayed the check for previous short-lived headers when adding a new header (which would iterate over headers coming before the current header and thus require a sorted vector of headers) to `finalize_headers_and_footers` so it's done in a single pass.

In combination, these two measures made it so the list of headers does not have to be sorted during cell resolving anymore, so I reverted it back to a (initially not sorted) vector and simply called `sort` once later on `finalize_headers_and_footers`, where a sorted list is actually needed (and must be returned).

This feels like a better solution overall, as performance is mostly preserved - the only difference now is that skipping the next header now costs O(header rows) rather than O(1), since each row is checked individually.

However, there is a certain extra memory cost with the bitset. In principle, this cost is negligible if all table headers are contained within the first 32 rows (or 64 depending on architecture) thanks to `SmallBitSet` which stores everything in a single `usize`. Allocations are needed for headers ~~or footer~~ beyond that amount of rows, however.

**Edit:** Since there is always only a single footer currently, checking the footer directly is always O(1), so I kept the footer check and removed footer rows from the occupied rows bitset to avoid unnecessary allocations due to some footer at the very end. We can think of some other way to deal with this in #6483 eventually.

PgBiel:
This should now be ready for a review. Overall changes:

- Fixed the linked bug by updating `auto_index` immediately if there were auto cells inside the header;
- Fixed header sort by delaying all operations depending on sort (or replacing, as in the case above of skipping headers when placing auto cells, where we now use a separate bitset mapping rows -> header is here or not instead);
- Also used the bitset when checking for header collision for cells with fixed position, potentially bringing further perf improvements in the case of multiple headers with lots of fixed-position cells.
- For the case of fixed-position cells colliding with headers or footers, the error message has been improved, now indicating the conflicting row.
- Simplified code and used more let chains (yay).

PR #6688, opened 2025-08-01
Prevent broken glyph assemblies when font data is incorrect

mkorje:
A note before merging: I have no idea why the span is not being used with this warning...

| Before | After |
| - | - |
| math-stretch-min-overlap-exceeds-max_old | math-stretch-min-overlap-exceeds-max_new |

```typ
--- math-stretch-min-overlap-exceeds-max ---
// Test that glyph assembly doesn't end up with negative lengths if the max
// overlap calculated is less than the minConnectorOverlap.
#show math.equation: set text(font: "STIX Two Math")
// Warning: glyph has assembly parts with overlap less than minConnectorOverlap
// Hint: its rendering may appear broken - this is probably a font bug
// Hint: please file an issue at https://github.com/typst/typst/issues
$ stretch(->)^"Gauss-Jordan Elimination" $
```


laurmaedje:
> I have no idea why the span is not being used with this warning...

The problem is quite specific. The `->` is an `ast::MathShorthand`, which is evaluated into a `Value::Symbol` (which can't have a span). Once that value is inserted into markup and becomes a `SymbolElem`, it usually gets a span. But in this case, the value dodges this by being immediately cast to content through the `stretch` constructor. (You can see the difference when typing `stretch(#$->$)` which does retain the span.)

laurmaedje:
I don't immediately see a simple fix for this (fwiw, elements=types would probably fix this as all values can have spans then). I would be okay with just merging as-is. Let me know if I should do that.

mkorje:
I see. I'm happy to merge as is then as well.

laurmaedje:
Thanks!

PR #6691, opened 2025-08-02
Support for raw syntax highlighting in HTML export

laurmaedje:
This PR adds support for syntax highlighting of raw blocks in HTML export. The "proper" way to support this would be to add support for `TextElem::fill` in HTML export. However, doing this properly is a bit tricky. Since having syntax highlighting work in HTML export is pretty useful, I've opted to implement a workaround for now.

### DIFF
--- a/crates/typst-html/src/css.rs
+++ b/crates/typst-html/src/css.rs
@@ -26,7 +26,6 @@ impl Properties {
     }
 
     /// Adds a new property in builder-style.
-    #[expect(unused)]
     pub fn with(mut self, property: &str, value: impl Display) -> Self {
         self.push(property, value);
         self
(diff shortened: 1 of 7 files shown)


QuadnucYard:
nice feature. should there be a inner wrapper `<code class="language-xxx">` for `<pre>` elements?

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element

laurmaedje:
Yeah, a `<code>` tag is currently indeed missing.

laurmaedje:
@QuadnucYard I added the `<code>` tag in https://github.com/typst/typst/pull/6701. I also opened a [separate PR](https://github.com/typst/typst/pull/6702/files) for the `language-x` thing, but didn't merge yet because it still suffers from a small problem with language tags that are not valid in class names.

PR #6695, opened 2025-08-03
Add interface to disable timer

Myriad-Dreamin:
This is useful when profile typst in a long running server.


laurmaedje:
Thanks!

PR #6696, opened 2025-08-03
Show aliases of citation styles in docs and completions

### LINKED ISSUE #6669: Show aliases of citation styles in the docs
### Description

Some citation styles have multiple names, a main display name, and one or more aliases.

Currently, the docs site says aliases are listed as a dedicated style, but in fact they are not listed anywhere.

> <html><body>

"institute-of-electrical-and-electronics-engineers" | IEEE
-- | --

</body>
</html>

We could change it to
> <html><body>

"institute-of-electrical-and-electronics-engineers" | IEEE (alias: "ieee")
-- | --

</body>
</html>

https://github.com/typst/typst/blob/59243dadbb2e9f5fc4fd55b0dbfaca6496caf761/crates/typst-library/src/model/bibliography.rs#L126-L131

https://github.com/typst/typst/blob/59243dadbb2e9f5fc4fd55b0dbfaca6496caf761/crates/typst-library/src/model/bibliography.rs#L439-L447

https://discord.com/channels/1054443721975922748/1399326777540608041/1399349305629741081

---

Besides, it is known `"alphanumeric"` is listed in both `cite` and `bibliography`, but not actually suitable for `bibliography`.
However, I don't think it worth fixing.

> Because `CiteElem` and `BibliographyElem` both use this same type, they share the same list

> Then I think it's fine to just ignore the difference between `cite` and `bibliography`. Since no one would normally want to use alphanumeric for `bibliography`, even if someone uses it accidentally (e.g., select the wrong item in auto completion), typst still provides a lovely error message. The difference only matters when  [enumerating all styles](<https://forum.typst.app/t/the-last-resort-for-correcting-the-format-of-bibliography-for-60-styles/5303/4?u=y.d.x>), which is quite rare.

https://discord.com/channels/1054443721975922748/1399326777540608041/1399364343296036964

YDX-2147483647:
Some citation styles have multiple names, a main name, and one or more aliases.
Currently, [the docs site says aliases are listed as a dedicated style](https://typst.app/docs/reference/model/bibliography/#parameters-style), but in fact they are not listed anywhere.
Futhermore, they are missing in auto-completions.

This PR will add those aliases to docs and completions.

This PR #6696 resolves #6669.
(Nice numbers :-)

## Example

`apa`, `ieee`, etc. will now be included in the JSON generated by typst-docs:

```diff
  {
    "string": "american-psychological-association",
    "details": "<p>American Psychological Association 7th edition</p>"
  },
+ {
+   "string": "apa",
+   "details": "<p>A short alias of <code>american-psychological-association</code></p>"
+ },
```

Before:

| Variant | Details |
|--------|--------|
| `"american-psychological-association"` | American Psychological Association 7th edition |

After:

| Variant | Details |
|--------|--------|
| `"american-psychological-association"` | American Psychological Association 7th edition |
| `"apa"` | A short alias of `american-psychological-association` |

You can verify it by running `cd docs && cargo run 2>/dev/null | rg '"apa"' --context 5`.

### DIFF
--- a/crates/typst-library/src/model/bibliography.rs
+++ b/crates/typst-library/src/model/bibliography.rs
@@ -442,9 +442,35 @@ impl Reflect for CslSource {
     #[comemo::memoize]
     fn input() -> CastInfo {
         let source = std::iter::once(DataSource::input());
-        let names = ArchivedStyle::all().iter().map(|name| {
-            CastInfo::Value(name.names()[0].into_value(), name.display_name())
-        });
+
+        /// All possible names and their short documentation for `ArchivedStyle`, including aliases.
+        static ARCHIVED_STYLE_NAMES: LazyLock<Vec<(&&str, &'static str)>> =
+            LazyLock::new(|| {
+                ArchivedStyle::all()
+                    .iter()
+                    .flat_map(|name| {
+                        let (main_name, aliases) = name
+                            .names()
+                            .split_first()
+                            .expect("all ArchivedStyle should have at least one name");
+
+                        std::iter::once((main_name, name.display_name())).chain(
+                            aliases.iter().map(move |alias| {
+                                // Leaking is okay here, because we are in a `LazyLock`.
+                                let docs: &'static str = Box::leak(
+                                    format!("A short alias of `{main_name}`")
+                                        .into_boxed_str(),
+                                );
+                                (alias, docs)
+                            }),
+                        )
+                    })
+                    .collect()
+            });
+        let names = ARCHIVED_STYLE_NAMES
+            .iter()
+            .map(|(value, docs)| CastInfo::Value(value.into_value(), docs));
+
         CastInfo::Union(source.into_iter().chain(names).collect())
     }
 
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/model/bibliography.rs
```diff
@@ -438,13 +438,40 @@ pub enum CslSource {
     Normal(DataSource),
 }
 
+/// All possible names and their short documentation for `ArchivedStyle`, including aliases.
+static ARCHIVED_STYLE_NAMES: OnceLock<Vec<(&&str, &'static str)>> = OnceLock::new();
```
laurmaedje: the static could be moved directly inside `input()` to keep it local to where it's used.

laurmaedje: also, I think this could just be a `LazyLock`.

YDX-2147483647: Thanks! just updated that. I should have read docs.rs/std first.

laurmaedje:
Thanks!

PR #6700, opened 2025-08-04
Fix CJ-Latin space at manual line breaks

### LINKED ISSUE #6539: cjk-latin-space + manual linebreak + align(center) ⇒ Unexpected left shift
### Description

In the following example, the text `国国` is intended to be horizontally centered, but it appears slightly shifted to the left.

```typst
#set text(font: "Noto Serif CJK SC")
#box(width: 2.3em, stroke: (x: green), align(center, [国国\ T T]))
```

Image

This issue only occurs when

- cjk-latin space is enabled (which is the default),
- and the line break is manually created by `linebreak()`.

### More examples

Image

```typst
#let test(body, width: 2.3em) = box(width: width, stroke: (x: green), align(
  center,
  body,
))

#set text(
  font: "Noto Serif CJK SC",
  lang: "zh",
  top-edge: "ascender",
  bottom-edge: "descender",
)

#set raw(lang: "typc")
#show raw: set text(font: ("DejaVu Sans Mono", "Noto Serif CJK SC"))

#set table.hline(stroke: 0.5pt)

#table(
  columns: 3,
  stroke: none,
  align: (horizon, center + horizon, horizon),
  ..(
    ([✅], `[国国TT]`),
    table.hline(),
    ([❌], `[国国\ TT]`),
    (
      [❌],
      `[国国] + linebreak() + [TT]
// Equivalent to the above`,
    ),
    (
      [❌],
      `text(cjk-latin-spacing: auto)[国国\ TT]
// Equivalent to the above`,
    ),
    ([✅], `text(cjk-latin-spacing: none)[国国\ TT]`),
    table.hline(),
    ([❌], `[国国\ T国]`),
    ([✅], `[国T\ T国]`),
    ([✅], `[国T\ 国国]`),
    ([✅], `[国国\ 国国]`),
  )
    .map(row => if type(row) == array and row.len() == 2 {
      let (ok, body) = row
      (ok, test(eval(body.text)), body)
    } else { row })
    .flatten(),
  table.hline(),
  [❌],
  test(width: 2.3em)[国国\ TT],
  [`[国国\ TT] + 2.3em`, where `2.3em` is width],
  [❌], test(width: 2.5em)[国国\ TT], `[国国\ TT] + 2.5em`,
  [❌], test(width: 2.8em)[国国\ TT], `[国国\ TT] + 2.8em`,
  [❌], test(width: 3.5em)[国国\ TT], `[国国\ TT] + 3.5em`,
)
```

</details> 

### Reproduction URL

_No response_

### Operating system

Windows, Web app

### Typst version

YDX-2147483647:
In the following example, there's a cj-latin space at the manual `linebreak` between 国 and T, resulting the misalign at the line `end`.
This PR will remove the cjk-latin space at both auto and manual line breaks, by fixing the adjust detection algorithm.

```typst
#align(end, [国国\ T国])
```

图片

Resolves #6539
(Please refer to the original issue for more tests and screenshots.)

_Disclaimer: I'm still not quite clear how `cjk-latin-space` is implemented, but all the tests have passed._ 😂

## The added test

| Source | Before | After |
|--------|--------|--------|

图片


crates/typst-layout/src/inline/line.rs
```diff
@@ -314,7 +314,8 @@ fn adjust_cj_at_line_boundaries(p: &Preparation, text: &str, items: &mut Items)
     }
 
     if text.ends_with(END_PUNCT_PAT)
-        || (p.config.cjk_latin_spacing && text.ends_with(is_of_cj_script))
+        || (p.config.cjk_latin_spacing
+            && text.strip_suffix("\n").unwrap_or(text).ends_with(is_of_cj_script))
```
laurmaedje: Rather than stripping the suffix here, you'll probably want to pass already trimmed text to `adjust_cj_at_line_boundaries` (trimming it as defined by the `trim` index in `fn line`). That would be more robust with other kinds of whitespace and newlines.

Overall, the CJK Latin spacing implementation is not super robust. Essentially, here we're undoing overeager spacing from the preparation phase and it doesn't match up perfectly.

YDX-2147483647: Thanks for your guidance! I now have learnt what these functions do, and have just revised it. (The snapshot does not need to be updated.)

laurmaedje:
Thanks!

PR #6702, opened 2025-08-04
Emit class with language tag for raw blocks

laurmaedje:
As proposed in https://github.com/typst/typst/pull/6691#issuecomment-3146420360 and https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element


laurmaedje:
Problem: The language tag might not be a valid class name identifier.

crates/typst-html/src/rules.rs
```diff
@@ -415,7 +415,14 @@ const RAW_RULE: ShowFn<RawElem> = |elem, _, styles| {
         seq.push(line.clone().pack());
     }
 
+    let lang = elem.lang.get_ref(styles);
     let code = HtmlElem::new(tag::code)
+        .with_optional_attr(
+            attr::class,
+            // As proposed in
+            // https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element
+            lang.as_ref().map(|lang| eco_format!("language-{lang}")),
```
reknih: How about going for the shorter `lang-js`? Or even something like `data-language="js"`?

laurmaedje: Let's go for `data-lang`. Also fixes the problem with identifier.

PR #6703, opened 2025-08-04
Ensure that whitespace is not collapsed in inline raw blocks

laurmaedje:
Without this, multiple spaces in inline code blocks will be rendered as a single one in browsers.

We could potentially omit this if the raw text does not contain consecutive whitespace. Not sure whether that's worth it.

### DIFF
--- a/crates/typst-html/src/rules.rs
+++ b/crates/typst-html/src/rules.rs
@@ -415,16 +415,20 @@ const RAW_RULE: ShowFn<RawElem> = |elem, _, styles| {
         seq.push(line.clone().pack());
     }
 
+    let mut inline = css::Properties::new();
+    let block = elem.block.get(styles);
+    if !block {
+        // Without the `<pre>` tag, whitespace would be collapsed by default.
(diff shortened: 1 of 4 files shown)


QuadnucYard:
I'm afraid we are doing too much with inline styles. Personally, I prefer global CSS rules applying to all elements of specific kinds, instead of spreading the SAME inline style everywhere, to keep the HTML clean and tidy.

And the question: Is there a method to turn off such built-in HTML rules for maximum customizability? Out-of-the-box is nice to me, but I hope it will not limit the capabilities of typst, or make it hard to obtain a simpler output.

laurmaedje:
More customizability for styling and support for stylesheets instead of inline styles are all planned. But for the time being, I'd like to write at least those styles that are relevant to express the actual content. The dropped spaces are not really about styling for me, it's just wrong.

laurmaedje:
What I don't really wanna do is ship a "standard Typst CSS" that is necessary to get proper output. Rather, I'd just like to generate the ideal style sheet for a document on the fly. But we're just not there yet.

PR #6705, opened 2025-08-05
Rename `pdf.embed` to `pdf.attach`

laurmaedje:
The name `pdf.embed` frequently causes confusion as it sounds like a way to embed one PDF file into another (similar to what's recently shipped with `image("file.pdf")`). However, what it actually does is attach an arbitrary file (e.g. XML) to the output PDF.

In previous discussions on Discord, there was strong consensus that it should be renamed. The name `pdf.attach` was generally received well as it's compact and clear in what it does. In the PDF spec, however, it's called an _Associated File,_ so there's also an argument to be made for renaming it to `pdf.associate` or `pdf.associated-file` instead.

Regardless of which name we choose, the name `pdf.embed` will remain as a deprecated alias until Typst 0.15.0.

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -6,8 +6,8 @@ use ecow::{EcoString, eco_format};
 use rustc_hash::FxHashSet;
 use serde::{Deserialize, Serialize};
 use typst::foundations::{
-    AutoValue, CastInfo, Func, Label, NoneValue, ParamInfo, Repr, StyleChain, Styles,
-    Type, Value, fields_on, repr,
+    AutoValue, CastInfo, Func, Label, NativeElement, NoneValue, ParamInfo, Repr,
+    StyleChain, Styles, Type, Value, fields_on, repr,
 };
 use typst::layout::{Alignment, Dir, PagedDocument};
 use typst::syntax::ast::AstNode;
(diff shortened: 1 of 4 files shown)


Enivex:
For what it's worth, I think `attach` is a much more familiar name than `associate`.

MDLC01:
"associate" or "associated file" could be added as a keyword

ghost:
I think `attach` is more confusing than `embed` here because it has a different meaning in pdf. In LaTeX, there are the `embedfile` and `attachfile` packages, the `embedfile` does what `pdf.embed` does, the `attachfile` embeds the file and makes it accessable via a File Attachement Annotation, what is not well supported by pdf readers.

I would assume, `pdf.attach` does what `attachfile` does.

Enivex:
> I think `attach` is more confusing than `embed` here because it has a different meaning in pdf. In LaTeX, there are the `embedfile` and `attachfile` packages, the `embedfile` does what `pdf.embed` does, the `attachfile` embeds the file and makes it accessable via a File Attachement Annotation, what is not well supported by pdf readers.
> 
> I would assume, `pdf.attach` does what `attachfile` does.

Based on what I can see from the PDF specification, a file attachment annotation just links to a file embedded in the same manner.

ghost:
A File Attachement Annotation is created at a specific position in the document (like an URL). An embedding can be created everywhere, the position does not matter. This is a different API. Also, the embedded file is published in the "List of embedded files", the linked File of the File Attachment Annotation normally is not included in this list, but can be (there is no LaTeX package that supports this, but it would be possible to achieve this).

laurmaedje:
I think the distinction between linked and unlinked is not clear at all from embed vs attach. In Typst, I'd imagine both to go under the same name. It could be as simple as linking to a label or location of an attach element.

PR #6706, opened 2025-08-05
Correct CeTZ spelling

sebaseb98:
Just a very minor documentation fix I came across. 

-  CeTZ was not spelled correctly in docs/reference/library/visualize.md
-  could be a matter of taste but I would say in a doc text you rather read "CeTZ" than "cetz" (even though that is the package name you actually import) so I improved that in docs/guides/guide-for-latex-users.md

### DIFF
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -482,7 +482,7 @@ allows you to create complex drawings and plots. To use CeTZ in your document,
 you can just write:
 
 ```typ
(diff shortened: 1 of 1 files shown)


LordBaryhobal:
Should the version in the import example be updated too ?
Latest version in Universe is 0.4.1

sebaseb98:
Yes, good catch @LordBaryhobal! Thank you :)

laurmaedje:
Thanks!

PR #6709, opened 2025-08-06
Move `par`, `box`, and `block` to show rules in HTML export

laurmaedje:
There's no need anymore for these to be hard-coded in the conversion code. Unlike in layout, they are not primitives in HTML export.


laurmaedje:
The `box` rule behaves differently from before (it creates an empty span instead of just being ignored if the box has no body), but this is more consistent with everything else and I'd say the old behaviour was just wrong.

PR #6730, opened 2025-08-08
Update RelativeTo to include tiling in docs

mewmew:


mewmew:
From https://typst.app/docs/reference/visualize/tiling/#constructor-relative

tiling-relative

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1143,9 +1143,10 @@ impl Repr for ConicGradient {
 /// What is the gradient relative to.
 #[derive(Cast, Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum RelativeTo {
-    /// The gradient is relative to itself (its own bounding box).
+    /// The gradient or tiling is relative to itself (its own bounding box).
```
MDLC01:
```suggestion
    /// Relative to itself (its own bounding box).
```

It would probably be more future-proof to completely remove mentions of the places where this is used.

mewmew: > It would probably be more future-proof to completely remove mentions of the places where this is used.

I agree. Updated in bc7e946fe77c7077b709b2faee42257c03afb19d.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1143,9 +1143,9 @@ impl Repr for ConicGradient {
 /// What is the gradient relative to.
```
mewmew, line 1143: Do we need to change this one as well?

```diff
-What is the gradient relative to.
+What is the gradient or tiling relative to.
```

Any suggestions for how to phrase it?

MDLC01: I don't think this appears anywhere in the documentation, but maybe "What this is relative to.", or "What something is relative to."

mewmew: As it doesn't show up in the documentation, and since the phrasing isn't "perfect", lets just leave this one as is for now.

Issue #3697, opened 2024-03-16
Resolve further fields in `grid` and `table` show rules


PgBiel:
### Description

It could be interesting to be able to access more info when applying show rules on tables. In particular, `rows` and `children`, in show rules, have (almost) the same values received by the `#table` call, when in reality those fields are changed during grid / table layout (more rows are needed depending on the cells). Similarly, at least `row-gutter` isn't the final array of gutter spacing.

In other words,

```typ
#show table: it => {
  
  show table.cell: it => it.fields()
  [Rows: #it.rows, #it]
}

#table(
  columns: 2,
  [a], [b],
  [c], [d],
  table.cell(stroke: red, y: 3)[e]
)
```

Displays `rows: ()` when it would be more useful to display something like `(auto, auto, auto)`, so we can know the actual amount of rows in the table. (Resolving `auto` to final lengths could be a bit out of scope at first, especially because `auto` rows behave very distinctly from fixed-size rows, so we should refrain from doing that initially.)

I think resolving `it.children` is debatable at first, because, while it would be nice to have access to each cell's resolved fields, a cell's stroke being specified or not does change its priority, for example. But if we assume that resolving that field doesn't affect the table's layout, then maybe it isn't that much of a problem. The larger problem could be possible performance concerns in converting every single `Vec<Entry>` to `Array<TableCell>` for example.

But, in principle, having at least `it.rows`  have the same amount of rows that is actually displayed in the table would be helpful.

### Use Case

Allows you to use the table's final layout to your advantage in show rules:

```typ
#show table: it => {
   show table.cell.where(y: it.rows.len() - 1): strong
   it
}
```

PR #6746, opened 2025-08-12
Move grid cell locator creation to GridLayouter

saecki:
This avoids storing `Locator`s inside the `CellGrid` and by that removes the lifetime parameter. I did this primarily to be able to store the `CellGrid` in a wrapper element in introspection tags, so it can be used in the pdf export for accessibility tags. But I think it cleans up the design overall.
Instead of using the cell body spans for initial disambiguation in the cell grid, the cell coordinates are used (which should never collide). And then the additional layouting disambiguator is used as before.
I'm not 100% sure this is correct, so maybe you can have a look at the `GridLayouter::cell_locator` function, but all tests seem to pass :)

### DIFF
--- a/crates/typst-layout/src/grid/layouter.rs
+++ b/crates/typst-layout/src/grid/layouter.rs
@@ -275,16 +275,23 @@ impl<'a> GridLayouter<'a> {
         }
     }
 
+    /// Create a [`Locator`] for use in [`layout_cell`].
     pub(super) fn cell_locator(
         &self,
         pos: Axes<usize>,
         disambiguator: usize,
     ) -> Locator<'a> {
-        let mut locator = self.locator.relayout().split().next(&pos);
+        // This key is unique for each cell position, so the split locator can
+        // be side-stepped.
+        let key = ((pos.x as u128) << 64) | (pos.y as u128);
+        let mut cell_locator = self.locator.relayout().split().next_inner(key);
+
+        // The disambiguator is used for repeated cells, e.g. in repeated headers.
         if disambiguator > 0 {
-            locator = locator.split().next_inner(disambiguator as u128);
+            cell_locator = cell_locator.split().next_inner(disambiguator as u128);
         }
-        locator
+
+        cell_locator
     }
 
     /// Determines the columns sizes and then layouts the grid row-by-row.
--- a/crates/typst-layout/src/grid/layouter.rs
+++ b/crates/typst-layout/src/grid/layouter.rs
@@ -1,6 +1,6 @@
-use std::collections::HashMap;
 use std::fmt::Debug;
 
+use rustc_hash::FxHashMap;
 use typst_library::diag::{SourceResult, bail};
 use typst_library::engine::Engine;
 use typst_library::foundations::{Resolve, StyleChain};
@@ -30,7 +30,7 @@ pub struct GridLayouter<'a> {
     /// The regions to layout children into.
     pub(super) regions: Regions<'a>,
     /// The locators for the each cell in the cell grid.
-    pub(super) cell_locators: HashMap<Axes<usize>, Locator<'a>>,
+    pub(super) cell_locators: FxHashMap<Axes<usize>, Locator<'a>>,
     /// The inherited styles.
     pub(super) styles: StyleChain<'a>,
     /// Resolved column sizes.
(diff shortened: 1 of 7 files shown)


crates/typst-layout/src/grid/layouter.rs
```diff
@@ -270,6 +275,18 @@ impl<'a> GridLayouter<'a> {
         }
     }
 
+    pub(super) fn cell_locator(
```
saecki, line 293: This could probably use a comment

laurmaedje:
The goal of the locator design is essentially to keep locations as stable as possible across iterations while making sure that the locations of two elements in the same iteration never collide. The coordinate approach fulfills the second criterion, but is not perfect on the first. 

On a document level, the equivalent of using coordinate positions would be using element indices. In such a setup, inserting an element shifts the locations of all following ones. As a consequence, the compiler can't identify the same elements across iterations and then stuff like `here().position()` stops working/converging. In practice, with the grid, this would only be a problem if the number or order of cells varies across iterations and there is introspection within the cells.

Apart from convergence, if a new cell is inserted in the coordinate approach, incremental compilation also has to redo everything in later cells, though that's not much of a concern in practice.

The span-based approach is designed to be a bit more stable. Of course, spans can also collide in scripting-heavy code and then the fallback is essentially indices again. It's not perfect either, but we've got to have some way to identify the same elements across iterations and so far, it's been working pretty well.

The changes here would probably also work out fine in practice. Still, I wonder whether it would be a bit more robust to have a hash map from coordinate positions to prepared locators generated through the more traditional span-based disambiguation instead.

saecki:
I see, is this what you had in mind?

laurmaedje:
Yeah, that's what I had in mind. Let's also wait what @PgBiel has to say, but this seems fine to me now.

crates/typst-layout/src/grid/layouter.rs
```diff
@@ -23,9 +26,11 @@ use super::{
 /// Performs grid layout.
 pub struct GridLayouter<'a> {
     /// The grid of cells.
-    pub(super) grid: &'a CellGrid<'a>,
+    pub(super) grid: &'a CellGrid,
     /// The regions to layout children into.
     pub(super) regions: Regions<'a>,
+    /// The locators for the each cell in the cell grid.
+    pub(super) cell_locators: HashMap<Axes<usize>, Locator<'a>>,
```
laurmaedje: could you use FxHashMap here like the rest of the compiler now uses?

saecki: Sure :+1:

crates/typst-layout/src/grid/layouter.rs
```diff
@@ -238,9 +244,22 @@ impl<'a> GridLayouter<'a> {
         let mut regions = regions;
         regions.expand = Axes::new(true, false);
 
+        // Prepare the locators for each cell in the cell grid.
+        let mut locator = locator.split();
+        let mut cell_locators = FxHashMap::default();
+        for y in 0..grid.rows.len() {
```
saecki, line 250: I think this is wrong when the `CellGrid` has a gutter.

saecki: EDIT: never mind, we're using the non-effective coordinates everywhere.

PgBiel (approved):
The changes seem fine to me.

Though I also share a bit of Laurenz's concern regarding stability when adding more cells. I suppose this isn't too bad in the general case either way.

Regardless, removing this layout component from CellGrid seems to be the right thing to do for HTML etc.

PgBiel:
> I did this primarily to be able to store the CellGrid in a wrapper element in introspection tags, so it can be used in the pdf export for accessibility tags.

This seems to tie into #3697. I didn't expect the main technical change to happen this early haha.

In the future, we might want to expose some API to access this in Typst too, so packages don't have to rewrite grid resolving. But, of course, we'd have to first discuss whether this even makes sense (e.g. would it be stable enough? Maybe only expose a subset of it?).

laurmaedje:
> Though I also share a bit of Laurenz's concern regarding stability when adding more cells. I suppose this isn't too bad in the general case either way.

With the hash map approach that @saecki pushed, it should now be the same as main w.r.t stability.

PgBiel:
> With the hash map approach that @saecki pushed, it should now be the same as main w.r.t stability.

Interesting. I suppose it would be a bit different though since it no longer uses spans, and in theory cells with different positions could have the same span, right? But probably not that relevant of a difference.

Edit: Nevermind, I just noticed it combines both position and span. So that's nice.

saecki:
It should be completely equivalent, the cell positions are only used as keys for the hashmap that stores the locators.

Issue #408, opened 2023-03-28
Improve functions documentation


HarmoGlace:
There are several parts of the docs where it is only written "function" as a type itself. But it not enough precise. Each "function" type should be replaced with a typed function, including the arguments and their types (and the expected value).

For instance, on the page [numbering setting,](https://typst.app/docs/reference/layout/page/#parameters--numbering) it is only written that it accepts a "function". I think that it should also add the following indications : `(current_page: integer, total_pages: integer) -> text` or something similar. It would avoid to have to manually guess what the arguments are, and I think that the purpose of the documentation is to make the objects understandable easily and clearly.

nunzayin:
+1 for this. Here's what I've encountered.

So the `str.replace` method's `replacement` argument can accept a `function` that "gets a dictionary for each match and can return individual replacement strings" (according to the current language reference).

The problem is I haven't found any information about what fields (key-value pairs) does the dictionary in question have. Thus, without any LSP support or just experimenting I assume there is no way to get this information. I have found out experimentally its specification:
```typ
(
  start: // a start position of matched string (int)
  end: // end position (int)
  text: // the matched string itself (str)
  captures: // ??? (got no values)
)
```

Please add this to the reference.

YDX-2147483647:
The difficulty lies in the fact that type information is carried by [`CastInfo`](https://docs.rs/typst-library/latest/typst_library/foundations/enum.CastInfo.html), but currently CastInfo can only record that the type is function and nothing more.

> `str.replace`

@nunzayin In your case, please refer to https://typst.app/docs/reference/foundations/str/#definitions-match.
I think we can add a link from `str.replace.replacement` to `str.match`.

YDX-2147483647:
Here's a list of parameters that can take a function.

- [x] array.find.searcher
- [x] array.position.searcher
- [x] array.filter.test
- [x] array.map.mapper
- [x] array.fold.folder
- [x] array.any.test
- [x] array.all.test
- [x] array.sorted.key
- [x] array.sorted.by #5627
- [x] array.dedup.key #6756
- [x] array.reduce.reducer
- [x] plugin.transition.func
- [x] selector.target
- [x] selector.or.others
- [x] selector.and.others
- [x] selector.before.end
- [x] selector.after.start
- [x] str.replace.replacement #6751
- [x] list.marker
- [x] figure.kind
- [x] figure.supplement
- [x] figure.numbering #6757
- [x] footnote.numbering #6757
- [x] heading.numbering #6757
[...]

PR #6751, opened 2025-08-12
Add links and examples in the docs of `str`

YDX-2147483647:
Most notably, explain the mysterious dictionary passed to [`str.replace.replacement`](https://typst.app/docs/reference/foundations/str/#definitions-replace-replacement).

Relates to #408
https://github.com/typst/typst/issues/408#issuecomment-3170649395

Some examples are copied from `tests/suite/foundations/str.typ`.


crates/typst-library/src/foundations/str.rs
```diff
@@ -422,6 +423,21 @@ impl Str {
     ///   group. The first item of the array contains the first matched
     ///   capturing, not the whole match! This is empty unless the `pattern` was
     ///   a regex with capturing groups.
+    ///
+    /// ```example
+    /// #(
+    ///   "Is there a".match("for this?"),
+    ///   "The time of my life.".match(regex("[mit]+e")),
+    /// )
+    /// ```
```
YDX-2147483647, line 430: Image

crates/typst-library/src/foundations/str.rs
```diff
@@ -422,6 +423,21 @@ impl Str {
...
+    ///
+    /// ```example
+    /// #let pattern = regex(
+    ///   `not (an?) ([\p{Letter}\s]+)\p{Punctuation}`.text,
+    /// )
+    /// #"I'm a doctor, not an escalator.".match(pattern) \
+    /// #"I am not a merry man!".match(pattern)
+    /// ```
```
YDX-2147483647, line 436: Image

crates/typst-library/src/foundations/str.rs
```diff
@@ -438,7 +454,11 @@ impl Str {
 
     /// Searches for the specified pattern in the string and returns an array of
     /// dictionaries with details about all matches. For details about the
-    /// returned dictionaries, see above.
+    /// returned dictionaries, see [above]($str.match).
+    ///
+    /// ```example
+    /// #"Day by Day.".matches("Day")
+    /// ```
```
YDX-2147483647, line 457: Image

crates/typst-library/src/foundations/str.rs
```diff
@@ -422,6 +423,21 @@ impl Str {
     ///   group. The first item of the array contains the first matched
     ///   capturing, not the whole match! This is empty unless the `pattern` was
     ///   a regex with capturing groups.
+    ///
+    /// ```example
+    /// #(
+    ///   "Is there a".match("for this?"),
```
laurmaedje: I think having an array here is unnecessary and a bit confusing. How about just two separate lines with linebreak as usual?

crates/typst-library/src/foundations/str.rs
```diff
@@ -422,6 +423,21 @@ impl Str {
...
+    ///   "The time of my life.".match(regex("[mit]+e")),
+    /// )
+    /// ```
+    ///
+    /// ```example
+    /// #let pattern = regex(
```
laurmaedje:
```suggestion
    /// #let pat = regex(
```
to avoid confusing with past (deprecated) or future built-ins called `pattern`

laurmaedje: I think the trick with a raw block + `.text` should be explained somewhere if it is used.

crates/typst-library/src/foundations/str.rs
laurmaedje: I feel like the example could be a bit more straightforward. As someone who is fairly familiar with regex, I takes quite a bit of mental effort to follow what's going on here.

crates/typst-library/src/foundations/str.rs
```diff
@@ -473,6 +493,9 @@ impl Str {
         /// The string to replace the matches with or a function that gets a
         /// dictionary for each match and can return individual replacement
         /// strings.
+        ///
+        /// The dictionary would be the same as the returned dictionary of
```
laurmaedje: How about:

```
The dictionary passed to the function has the same shape as the dictionary returned by [`match`]($str.match).
```

crates/typst-library/src/foundations/str.rs
YDX-2147483647: Well, the first one matches nothing and returns `none`, which is not visible in the document.
Putting them in an `array` will make the `none` visible.

I have considered `repr`, but that's kind of distractive…

laurmaedje: Ah, I see. What about `assert.eq`?

crates/typst-library/src/foundations/str.rs
YDX-2147483647: OK, I've changed the regex `not (an?) (apple|cat)`. How about it now?

I still put two capture groups in that single regex. It emphasizes that `captures` contains all captured groups in the first match, not the matching texts of all matches.

Image

laurmaedje: I think what confused me mostly was the `(an?)`. I first read it as `(an)?`, which I could imagine is a common mistake because of the combination of parentheses and a question mark.

YDX-2147483647: Would `(a|an)` be better?

---

I tried to find some inspiration from [python docs](https://docs.python.org/3.12/library/re.html#re.findall), [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match), and [rust docs](https://docs.rs/regex/latest/regex/#examples), but I failed.
I don't think their examples are better, for the following reasons:
- They use `\d` or `\w`, which are Unicode-aware in Typst, but may not in other implementations. `\d` and `\w` are not quite useful in non-Latin context due to inconsistency.
- They are hack-parsing something (e.g. date). It's a typical use case of regex, but I don't think we should encourage any specific regex in the docs of `str`.
- They have only one capture group.
  > I still put two capture groups in that single regex. It emphasizes that `captures` contains all captured groups in the first match, not the matching texts of all matches.
- …

laurmaedje: > Would (a|an) be better?

I would find that clearer indeed

YDX-2147483647: All right; Just edited.

laurmaedje:
Thanks!

PR #6753, opened 2025-08-13
Fix compiler warnings on Rust 1.89

Drodt:
Fixes new compiler warnings of hidden lifetimes. Also uses explicitly the `OptionExt` trait to prepare for new std functionality.


laurmaedje:
Our utils impl and std do the same thing, so I wonder whether it's necessary to do this. I would've hoped that we can just get the std impl and then remove the polyfill trait some way down the road when the MSRV allows for it.

Drodt:
So you want to leave the warnings until these functions are stabilized?

laurmaedje:
We could just silence the warnings. Or do you think it's a bad idea?

It feels a bit unnecessary to me to change to fully qualified calls and then switch back to the exact previous code.

Drodt:
I changed it as you wished

laurmaedje:
Thanks! I folded this into https://github.com/typst/typst/pull/6775 to make sure we don't accumulate new warnings before bumping.

PR #6756, opened 2025-08-13
Improve the docs on function params of `array.{sorted,dedup}`

YDX-2147483647:
Relates-to: #408


crates/typst-library/src/foundations/array.rs
```diff
@@ -828,11 +828,12 @@ impl Array {
         engine: &mut Engine,
         context: Tracked<Context>,
         span: Span,
-        /// If given, applies this function to the elements in the array to
+        /// If given, applies this function to each element in the array to
         /// determine the keys to sort by.
         #[named]
         key: Option<Func>,
-        /// If given, uses this function to compare elements in the array.
+        /// If given, uses this function to compare every two elements in the
```
laurmaedje, line 832: I'm not sure what exactly "every two" means, but it seems unclear/wrong. It's not really specified which elements will be compared but it's likely neither every two neighbouring elements nor every combination.

crates/typst-library/src/foundations/array.rs
```diff
@@ -828,11 +828,12 @@ impl Array {
         engine: &mut Engine,
         context: Tracked<Context>,
         span: Span,
-        /// If given, applies this function to the elements in the array to
+        /// If given, applies this function to each element in the array to
```
laurmaedje, line 828: That sounds a bit like the function is guaranteed to be called only once, which is not the case. But it's not a big deal and perhaps clearer than before.

crates/typst-library/src/foundations/array.rs
YDX-2147483647: I didn't consider it before…

How about the following?

> If given, uses this function to compare elements in the array.
>
> The function will receive two elements in the array for comparison, and should return a boolean indicating their order: `true` indicates… (no change below)

laurmaedje: better!

laurmaedje:
Thanks!

Issue #4367, opened 2024-06-10
reset  page number each chapter, the total numbers always return the last chapter.


zincers:
### Description

```
#set page(
  width: 300pt,
  height: 100pt,
  footer: [
    #counter(page).display("1/1", both: true) 
  ]
)
= chapter 1
== page 1.1
#lorem(10)
#pagebreak()
= page 1.2
#lorem(10)
#pagebreak()
= chapter 2
#counter(page).update(1)
== page 2.1
#lorem(10)
#pagebreak()
== page 2.2
#lorem(10)
#pagebreak()
== page 2.3
#lorem(10)
```
When reset  page number each chapter, the total numbers always return the last chapter.

![image](https://github.com/typst/typst/assets/19182031/a3c74043-661b-4f7a-8530-bc95af38faff)

In the first chapter, the expected return is 1/2 and 2/2， rather than 1/2 and 2/3

### Operating system

Windows

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
Until there is native support, you can implement it like this:
```typ
// Returns the page total of the current section.
#let reset = <__reset>
#let subtotal() = {
  let loc = here()
  let list = query(selector(reset).after(loc))
  let (total,) = if list.len() > 0 { 
    counter(page).at(list.first().location())
  } else {
    counter(page).final()
  }
  total
}

// A section of content with reset page numbering.
//
// Leaves a marker we can query to find the subtotal page number.
// Resets the page counter in between the pages.
#let section(it) = {
  [#metadata(none)#reset]
  pagebreak(weak: true)
  counter(page).update(1)
  pagebreak(weak: true)
  it
}

#let page-numbers = context numbering(
  "1 / 1",
  ..counter(page).get(),
  subtotal(),
)

#set page(
  paper: "a8",
  margin: 40pt,
  // Show both in the header & footer to see the effect.
  header: align(center, page-numbers),
  footer: align(center, page-numbers),
)

#section([= A] + lorem(30))
#section([= B] + lorem(60))
#section([= C] + lorem(30))
```

zincers:
@laurmaedje thanks for your help, it works for me.

Issue #5442, opened 2024-11-19
The `counter` documentation does not help with multi-level counters


gasche:
### Description

I needed a multi-level counter, but I found that the [counter](https://typst.app/docs/reference/introspection/counter/) documentation only really discusses single-level counter, which made it much harder than it should have been.

I wanted to index "questions" with numbers of the form S.C, where S is the section counter and C is a question-specific counter. (For example: 1.1, 1.2, then 2.1). Here is the code that I ended up writing:

```typst
#let question_counter = counter("question")
#question_counter.update((0, 0))
#show heading: it => {
  if it.level == 1 {
    question_counter.update((s, c) => (s+1, 0))
  }
}

#let question(body) = {
 question_counter.update((s, c) => (s, c+1))
 strong([ Question #context{ question_counter.display() } ])
 body
}

= Example usage
#question[Foo]
#question[Bar]

= More example
#question[Baz]
```

This required too much trial-and-error, I believe that the `counter` implementation should be implemented to make this easy. For example:

1. There are no examples of `update` with multi-level counters.
2. The behavior of `step` on multi-level counters is not explained.
     - Does the indexing start at 0 or 1?  
       Answer: it appears to start at 1.
     - is `question_counter.step(level: 1)` equivalent to `question_counter.update((s, c) => (s+1, c))` or to `question_counter.update((s, c) => (s+1, 0))`?
        Answer: in my tests, neither, it appears to be unusable with 2-level counters.
3. I get weird errors if I remove the `#question_counter.update((0, 0))` initialization, why? Is there a way to indicate that a counter is multi-level at creation time?

gasche:
My understanding is that the currently intended specification for the `step` functions on multi-level counters is given by those tests:

https://github.com/jbirnick/typst/blob/d4cd60a70d3eb7601622aafffb0479ca07947e66/tests/suite/introspection/counter.typ#L148-L158

In particular, note that if you have a two-level counter `(s,c)`, the `count.step(level: 1)` results in the counter `(s+1,)` and not `(s+1, 0)` as one might expect. In other words, the number of levels will change dynamically during execution. I am not sure how user are expected to conveniently program with this specification.

laurmaedje:
> In other words, the number of levels will change dynamically during execution. 

The reason for this is heading numberings such as `1, 1.1, 1.2, 2, 2.1` have varying number of levels.

jbirnick:
[...]

Issue #1896, opened 2023-08-09
Ergonomic numbering-by-section


damaxwell:
### Description

It's pretty standard to have equations/figures/tables/etc in academic work numbered by section or by chapter (i.e. the 23rd equation of section 5 is equation (5.23) ).  While Typst makes a lot of things easier than the competition, arranging for numbering by section currently has a lot of friction. Ad-hoc solutions are starting to appear, but with varying amounts of fragility.  It seems like numbering-by-section is a feature that should be built in to Typst, and easy to use out-of-the box.

This feature request is meant to less than an RFC, but a home for discussing solutions that might improve the user experience.  In a perfect world, something like

```
set math.equation( numbering: "({h1}.1))", supplement: none)
```

would be enough to get equation numbers numbered by section, in parentheses, and at the same time have a reference `@profound-eq` yield `(5.23)` for equation 23 of section 5.

To make number-by-section happen currently for equations:
1) A show rule for `heading` needs to be established to resent the equation counter when a new level 1 heading begins.
2) A custom numbering function needs to be assigned to `math.equation`.  Because numbering functions only receive counter values and no other info, the numbering function can only add the section number based on its call site using `location`.  Also, the numbering function needs to be careful to format the section number so that if appendix numbering by A, B, C begins, then this is correctly detected, and so that if subsections aren't numbered we don't accidentially pick up a null numbering function if we forget to seek level 1 sections.  I've seen both these bugs in the wild.
3) Simultaneously, a show rule for `ref` (with nontrivial logic because of a `null` test and a type check) that gets along with the element function's numbering function needs to be set up.  Because the numbering function for `math.equation` is location sensitive, it can't be fully recycled here and it seems like the `ref` show rule has to know separately the details of numbering by section.  Shared code could be refactored, of course, but the point is the complexity.

All this can be made to work, but I hope this is just a temporary solution as Typst matures.

One could imagine a Gadget, managing a counter for an element function, that does the following kinds of tasks:

1) Responds to a hook for updating the element's counter prior to an automatic bump.  There is a hook already, the `Count` trait, but it's only available from rust.  For numbering by section, the Gadget can look into the past to the last time the counter it manages was updated, grab the section number then, compare it to the section number now, and reset the managed counter if need be.  The hook could also have greater responsibility, maybe replacing the `Count` trait, perhaps.
[...]

PR #6757, opened 2025-08-13
Improve docs of various numbering parameters

YDX-2147483647:
- Explain what these numbers are and how many there are.

- Correct that numbers can be zero. It indicates the first element has not yet appeared.

- Imply the limitation of `counter(page).final()`.

  It was mentioned several times:

  - #4367
  - [How to reset the page counter after every section? - Questions - Typst Forum](https://forum.typst.app/t/how-to-reset-the-page-counter-after-every-section/1924/12?u=laurmaedje)
  - [How to make `"1 / 1"` numbering ignore the back matter? - Questions - Typst Forum](https://forum.typst.app/t/how-to-make-1-1-numbering-ignore-the-back-matter/3472)

  ~~I add an example workaround. However, it might be too wacky. Should I remove it?~~ No removed.

Relates-to: #408

### DIFF
--- a/crates/typst-library/src/layout/page.rs
+++ b/crates/typst-library/src/layout/page.rs
@@ -207,18 +207,20 @@ pub struct PageElem {
     ///
     /// Accepts a [numbering pattern or function]($numbering) taking one or two
     /// numbers:
-    /// 1. the first number is the current page number
-    /// 2. the second number (if it exists) is the total number of pages
+    /// 1. The first number is the current page number.
+    /// 2. The second number is the total number of pages. In a numbering
+    ///    pattern, the second number can be omitted. If a function is passed,
+    ///    it will always receive both numbers.
     ///
     /// These are logical numbers controlled by the page counter, and may thus
-    /// not match the physical number. To be specific, they are the
+    /// not match the physical numbers. Specifically, they are the
     /// [current]($counter.get) and the [final]($counter.final) value of
-    /// `{counter(page)}`. See the documentation for [`counter`]($counter/#page-counter)
-    /// for further customizations.
+    /// `{counter(page)}`. See the [`counter`]($counter/#page-counter)
+    /// documentation for more details.
     ///
-    /// In addition, if an explicit [`footer`]($page.footer) (or
-    /// [`header`]($page.header) for [top-aligned]($page.number-align) numbering)
-    /// is given, the numbering is ignored.
+    /// If an explicit [`footer`]($page.footer) (or [`header`]($page.header) for
+    /// [top-aligned]($page.number-align) numbering) is given, the numbering is
+    /// ignored.
     ///
     /// ```example
     /// #set page(


crates/typst-library/src/layout/page.rs
```diff
@@ -216,6 +233,26 @@ pub struct PageElem {
...
+    /// #metadata(none) <preface-end>
+    ///
+    /// #set page(numbering: "1 / 1")
+    /// #counter(page).update(1)
+    ///
+    /// = Main text
+    /// #lorem(5)
+    /// #pagebreak()
+    /// #lorem(5)
+    /// ```
```
YDX-2147483647: Image

This example might be too wacky. Should I remove it?

crates/typst-library/src/model/numbering.rs
```diff
@@ -79,7 +79,10 @@ pub fn numbering(
     /// numberings to the `numbering` function without caring whether they are
     /// defined as a pattern or function.
     numbering: Numbering,
-    /// The numbers to apply the numbering to. Must be positive.
+    /// The numbers to apply the numbering to. Must be non-negative.
+    ///
+    /// In general, numbers are counted from one. A number of zero indicates
+    /// that the first element has not yet appeared.
```
YDX-2147483647, line 85: https://github.com/typst/typst/issues/1896#issuecomment-2005782054 or [headcount – Typst Universe](https://typst.app/universe/package/headcount) depends on this behaviour for equations/images/tables that come before the first heading.

crates/typst-library/src/layout/page.rs
mewmew: > This example might be too wacky. Should I remove it?

I'd say it feels very niche to add a metadata node to specify the end of the preface.

It took me a moment to realize why it didn't use regular `set` rules for numbering of preface. Perhaps add a `lorem` content to the preface so it becomes two pages? Otherwise the final page number in greek letters (`I / II`) is not obvious. Also, consider using lower case (`i / ii`).

mewmew:
@YDX-2147483647, glad to see all your PRs to improve documentation!

May I ask that this PR stays in review for a few more days before being merged?

I am about to submit a proposal about how to handle multi-level numbering patterns (e.g. refer to a figure as "3.1" where `3` is the section number, and `1` is the first figure of that section; related issues #246 and #5442).

The proposal will introduce an optional `level` field for figure (with default value 1), that is analogous to how multi-level heading numberings are handled.

The change itself is very short (diff of 3 lines of code), and it gives powerful means of handling custom multi-level referenceable elements.

If such a proposal was to be accepted, then part of the docs in this PR would be updated (namely the references to `taking one number/argument`).

Of course, a follow-up PR could also adjust those comments if you prefer to merge this PR now.

I'll prepare the proposal for multi-level figures and submit a PR that implements the proposal :)

Exciting times ahead!
/Robin

laurmaedje:
@mewmew https://github.com/typst/typst/issues/2652 and the discussion further down in https://github.com/typst/typst/pull/5716 might also be relevant.

mewmew:
> @mewmew https://github.com/typst/typst/issues/2652 and the discussion further down in https://github.com/typst/typst/pull/5716 might also be relevant.

Thanks! Definitely good to see that this has been discussed at length, and that there are a few different approaches to improve upon the situation : )

I'll dive into more detail, on first glance dependent numberings looks good. There is a specific use case that I hope it is able to handle, will have to experiment a bit.

In particular, I wish to use a `show ref` rule to implement an analog of `cref` for Typst. I have already managed to implement `cref` using the proposal implemented in #6768 :)

The reason I use `cref` to assess these proposals is that it has been able to tease out a lot of subtle bugs with how counters are handled (e.g. `counter.at(here())`).

I'd love to see this situation improve and look forward to reading up more about dependent counters/numberings.

laurmaedje:
> The reason I use cref to assess these proposals is that it has been able to tease out a lot of subtle bugs with how counters are handled (e.g. counter.at(here())).

Which kinds of bugs are you referring to?

mewmew:
> > The reason I use cref to assess these proposals is that it has been able to tease out a lot of subtle bugs with how counters are handled (e.g. counter.at(here())).
> 
> Which kinds of bugs are you referring to?

Turns out it was a "fat fingers" type of bug (late night coding on my part).

I thought it was related to the incorrect location being used in references, but as you noted in https://github.com/typst/typst/issues/1896#issuecomment-2006357167 the correct context is being provided for `counter(heading)` (so `counter.get()` is resolved where the element is positioned, instead of where the reference is used).

---

Copy-paste bug introduced on my part. From https://github.com/typst/packages/pull/2878/files#diff-1352bb9aa5da14461f320dad29a60c63e724cffc61ead61e6f0294227df1f36aR93

```diff
-@fig3, @subfig1-foo and @subfig1-bar are ...
+@fig3, @subfig3-foo and @subfig3-bar are ...
```

fat-fingers

crates/typst-library/src/layout/page.rs
YDX-2147483647: > so it becomes two pages?

Hmm, then the workaround won't be necessary, because both `counter(page).final()` and `counter(page).at(<preface-end>)` will be `(2,)`.

The reason why the regular numbering pattern doesn't work is explained in the paragraph starting with _More precisely_.

> using lower case (`i / ii`).

Thanks for your advice. I'll apply it if this example survive in the final version…

mewmew:
> Which kinds of bugs are you referring to?

It turns out @YDX-2147483647 had already documented the small little quirks that comes up when trying to use custom `numbering` functions that rely on `counter.get()` (for multi-level numberings).

Really well documented in:
https://forum.typst.app/t/how-to-counter-display-at-location-or-get-remote-context/5096

In particular, I wish not only to *display* the references properly in the document, but also perform computations on the array of numbers returned from the custom `numbering` functions.

The reason I wish to retrieve the array of numbers instead of just displaying the references is that I would like to compact the references where applicable, e.g. "Figure 1, 2, 3, 4" becomes "Figures 1-4".

Therefore it should use the context of where the figure/element is located, not where the `cref` function is invoked.

It seems like the only option is to use `metadata`? Since the `show ref` rule won't give me access to the actual array of numbers to be able to perform the compacting (or perhaps I miss how it can be used).

**Edit:** if #6767 was implemented, this would become `ref.element.counter.at(ref.target)` for a given reference to a figure.

mewmew:
> Which kinds of bugs are you referring to?

*Take 3.*

@laurmaedje it took some time to identify the issue. I hope the explanation below is understandable, otherwise I'd be happy to provide more `context` (pun intended).

I've finally been able to produce a "minimal" working example that illustrates the issue (using [`cref`](https://github.com/typst/packages/blob/d0e5fc4ced0d63ddb0ddc87f98e758da942338dd/packages/preview/smartaref/0.1.0/src/lib.typ#L159)).

In particular, the reference to subfigure "1b" is broken in [example-subpar.typ](https://raw.githubusercontent.com/typst/packages/main/packages/preview/smartaref/0.1.0/example/example-subpar.typ) (of the [`smartaref` package](https://typst.app/universe/package/smartaref)), which relies on the original behaviour of `figure.counter` (i.e. not using `counter.step(level)` of #6767).

example-subpar-broken

The corresponding example that works as expected is [example-subfig.typ](https://raw.githubusercontent.com/typst/packages/main/packages/preview/smartaref/0.1.0/example/example-subfig.typ) (of the [`smartaref` package](https://typst.app/universe/package/smartaref)), which relies on the new behaviour of `figure.counter` (i.e. using `counter.step(level)` of #6767).

example-subfig-working

Relevant parts of the [`cref`](https://github.com/typst/packages/blob/d0e5fc4ced0d63ddb0ddc87f98e758da942338dd/packages/preview/smartaref/0.1.0/src/lib.typ#L159) function:
```typst
#let cref(..args, supplement: supplement-func) = context {
	// ...
	let target = query(ref.target).first()
	let elem = target
	let nums = c.at(elem.location())
	let c = elem.counter
	let nums = c.at(elem.location())
	let text = std.numbering(elem.numbering, ..nums)
	// ...
}
```

The reason for the broken reference is likely due to what @YDX-2147483647 documented in https://forum.typst.app/t/how-to-counter-display-at-location-or-get-remote-context/5096. That custom `numbering` functions are only correct if used from a `show ref` rule or when the results stored as `metadata`. But when used elsewhere, the correct context is not provided (since the custom `numbering` functions use `counter(...).get()` to resolve e.g. headings, parent figure numbers of subfigures, etc).

crates/typst-library/src/layout/page.rs
laurmaedje: I agree that it's a bit wacky.

There is also the line with "might be improved in the future", which I wonder whether there are any ideas for that at all?

In general, I think an occasional workaround in the docs can be better than nothing, but this might be a bit far.

YDX-2147483647: I remove the workaround and leave a link to https://typst.app/docs/reference/introspection/counter/#page-counter.

laurmaedje:
Thanks again!

PR #6760, opened 2025-08-14
Don't hide captured inner lifetime

tingerrr:
Same as the small `pdf-writer` PR. From what I can gather the inner lifetime could be `'static`, but since this is because it's cached, which could change, so I opted for the shortened lifetime (which is what I think rustc does here too).


Enivex:
This overlaps with https://github.com/typst/typst/pull/6753 by the way

tingerrr:
Ah, I didn't catch that, closing this then.

PR #6761, opened 2025-08-14
Add `typst info` command that emits binary and environment information

tingerrr:
While I couldn't find an issue for this, a few people have already asked for the ability to ask Typst directly about the paths it uses (namely the package paths). For example, to correctly cache packages in CI workflows without relying on possibly outdated scripts that duplicate the behavior of Typst's package path fallbacks.

I figured, if we add such a command, we may as well output other information debugging too. This command currently displays the current version and compile time features, all environment variables Typst reads in some form and values they resolve to, such as features, font paths, package paths, and so on.

P.S.: I initially named this `environment`/`env`, but figured that's a misnomer because it shows more than just environment variables. Feel free to bikeshed about the name.

### DIFF
--- a/crates/typst-cli/src/info.rs
+++ b/crates/typst-cli/src/info.rs
@@ -1,0 +1,553 @@
+use std::env::VarError;
+use std::fmt::Display;
+use std::io::{self, Write};
+use std::path::{Path, PathBuf};
+
+use clap::builder::{FalseyValueParser, TypedValueParser};
+use clap::{CommandFactory, ValueEnum};
+use codespan_reporting::term::termcolor::{Color, ColorSpec, WriteColor};
+use ecow::eco_format;
+use serde::Serialize;
+use typst::diag::StrResult;
+
+use crate::CliArguments;
+use crate::args::{Feature, InfoCommand};
+use crate::terminal::{self, TermOut};
+
+/// A struct holding the machine readable output of the environment command.
+#[derive(Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct Info {
+    /// The Typst version.
+    version: &'static str,
+
+    /// Build info about Typst.
+    build: Build,
+
+    /// The runtime features from `TYPST_FEATURES`.
+    features: Features,
+
+    /// Font configuration.
+    fonts: Fonts,
+
+    /// Package configuration.
+    packages: Packages,
+
+    /// The environment variables that are of interest to Typst.
+    env: Environment,
+}
+
+/// Build info about Typst.
+#[derive(Default, Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct Build {
+    /// The commit this binary was compiled with.
+    commit: &'static str,
+
+    /// Compile time settings.
+    settings: Settings,
+}
+
+/// Compile time settings.
+#[derive(Default, Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct Settings {
+    /// Whether the `self-update` compile-time feature is enabled.
+    self_update: bool,
+
+    /// Whether the `http-server` compile-time feature is enabled.
+    http_server: bool,
+}
+
+impl Settings {
+    /// Return the compile features with human readable information.
+    fn compile_features(&self) -> impl Iterator<Item = KeyValDesc<'_>> {
+        let Self { self_update, http_server } = self;
+
+        [
+            ("self-update", self_update, "Update typst via `typst update`"),
+            ("http-server", http_server, "Serve HTML via `typst watch`"),
+        ]
+        .into_iter()
+        .map(|(key, val, desc)| KeyValDesc { key, val: Value::Bool(*val), desc })
+    }
+}
+
+/// The runtime features from `TYPST_FEATURES`.
+#[derive(Default, Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct Features {
+    html: bool,
+    a11y_extras: bool,
+}
+
+impl Features {
+    /// Return the runtime features with human readable information.
+    fn features(&self) -> impl Iterator<Item = KeyValDesc<'_>> {
+        let Self { html, a11y_extras } = self;
+
+        [
+            ("html", html, "Experimental HTML support"),
+            ("a11y-extras", a11y_extras, "Experimental PDF accessibility extensions"),
+        ]
+        .into_iter()
+        .map(|(key, val, desc)| KeyValDesc { key, val: Value::Bool(*val), desc })
+    }
+}
+
+/// Font configuration.
+#[derive(Default, Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct Fonts {
+    /// The font paths from `TYPST_FONT_PATHS`.
+    paths: Vec<PathBuf>,
+
+    /// Whether system fonts were included in the search.
+    system: bool,
+
+    /// Whether embedded fonts were included in the search.
+    embedded: bool,
+}
+
+impl Fonts {
+    /// Return the custom font paths.
+    fn custom_paths(&self) -> impl Iterator<Item = Value<'_>> {
+        self.paths.iter().map(|p| Value::Path(p))
+    }
+
+    /// Return whether system and embedded fonts are included.
+    fn included(&self) -> impl Iterator<Item = (&'static str, Value<'_>)> {
+        let Self { paths: _, system, embedded } = self;
+
+        [("System fonts", system), ("Embedded fonts", embedded)]
+            .into_iter()
+            .map(|(key, val)| (key, Value::Bool(*val)))
+    }
+}
+
+/// Package configuration.
+#[derive(Default, Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct Packages {
+    /// The resolved package path.
+    package_path: Option<PathBuf>,
+
+    /// The resolved package cache path.
+    package_cache_path: Option<PathBuf>,
+}
+
+impl Packages {
+    /// Return the resolved package paths.
+    fn paths(&self) -> impl Iterator<Item = (&'static str, Value<'_>)> {
+        let Self { package_path, package_cache_path } = self;
+
+        [("Package path", package_path), ("Package cache path", package_cache_path)]
+            .into_iter()
+            .map(|(k, v)| (k, v.as_deref().map(Value::Path).unwrap_or(Value::Unset)))
+    }
+}
+
+/// The environment variables that are of interest to Typst.
+#[derive(Default, Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+struct Environment {
+    typst_cert: Option<String>,
+    typst_features: Option<String>,
+    typst_font_paths: Option<String>,
+    typst_ignore_system_fonts: Option<String>,
+    typst_package_cache_path: Option<String>,
+    typst_package_path: Option<String>,
+    typst_root: Option<String>,
+    typst_update_backup_path: Option<String>,
+    #[cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))]
+    xdg_cache_home: Option<String>,
+    #[cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))]
+    xdg_data_home: Option<String>,
+    source_date_epoch: Option<String>,
+    fontconfig_file: Option<String>,
+    #[cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))]
+    openssl_conf: Option<String>,
+    no_color: Option<String>,
+    no_proxy: Option<String>,
+    http_proxy: Option<String>,
+    https_proxy: Option<String>,
+    all_proxy: Option<String>,
+}
+
+impl Environment {
+    fn vars(&self) -> impl Iterator<Item = (&'static str, Value<'_>)> {
+        let Environment {
+            typst_cert,
+            typst_features,
+            typst_font_paths,
+            typst_ignore_system_fonts,
+            typst_package_cache_path,
+            typst_package_path,
+            typst_root,
+            typst_update_backup_path,
+            #[cfg(not(any(
+                target_os = "windows",
+                target_os = "macos",
+                target_os = "ios",
+            )))]
+            xdg_cache_home,
+            #[cfg(not(any(
+                target_os = "windows",
+                target_os = "macos",
+                target_os = "ios",
+            )))]
+            xdg_data_home,
+            source_date_epoch,
+            fontconfig_file,
+            #[cfg(not(any(
+                target_os = "windows",
+                target_os = "macos",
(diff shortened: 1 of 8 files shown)


laurmaedje:
I think this is cool overall, I just have a few remarks:
- The command name `doctor` sounds like it would fix something, like `cargo clippy --fix`. How about a more neutral `typst info`?
- I think the mixing of general info and compile-time features is a bit odd. I admit that it's a bit tricky because the are both compile-time and runtime features and the runtime features are just called "features".
- Should compile time and runtime features perhaps be an array in the serialized representation? The downside I guess is that there is no way to get all features that Typst is aware of in the current build. I could see it going either way.
- Not sure how the current order of environment variables came to be, but I think a couple important ones should be further up, e.g. `TYPST_ROOT` and `TYPST_FEATURES` seem more relevant than `SOURCE_DATE_EPOCH`

tingerrr:
Yeah, `info` sounds good, I went with doctor because a few CLIs had them for "self diagnosis", hence the moniker `doctor`.

The compile time features were actually separate at first, but the general section was kind of empty and most users probably don't care too much about a feature being compile-time, but putting them in runtime features implies you can turn them on and off, which is obviously not true.

I could change that back if you prefer it that way though, to me personally, it's sufficiently "general".

> - Should compile time and runtime features perhaps be an array in the serialized representation? The downside I guess is that there is no way to get all features that Typst is aware of in the current build. I could see it going either way.

That's how I had it before, but I went with explicit structs because of that reason, to make sure all known features are listed. It also doesn't make iterating them harder, as most tools that interact with JSON can just iterate over keys.

> - Not sure how the current order of environment variables came to be, but I think a couple important ones should be further up, e.g. TYPST_ROOT and TYPST_FEATURES seem more relevant than SOURCE_DATE_EPOCH

I simply ordered them alphabetically, I'll order it as `TYPST_*`, `XDG_*`, the rest.

tingerrr:
One possible concern with `typst info` is that once the CLI gets packaging capabilities (publishing, searching registry, etc.) one might assume `typst info` is for package or project info (which people would likely expect from other tools like `cargo` or `npm`).

Since this is more about debugging and mostly meant for scripts, perhaps we should add a `util` or `debug` command to put this under like some other tools do or find a different command name?

laurmaedje:
What I could also imagine is `typst self info` and migrate update to `typst self update` (with deprecation of the old one, which could also be mistaken for "update packages"). 

Or something in this direction. Just a thought --- maybe it's overkill.

tingerrr:
Perhaps, I would imagine that both of these commands are fairly rarely used (as opposed to updating packages or getting package/project info). Clearing up the names for these more common commands down the line would be cool.

laurmaedje:
I am happy with `typst info` as the command and we're comfortable with shipping a breaking change to the CLI later if necessary.

However, a few more thoughts:

- The env var section should list relevant XDG variables that Typst considers. There are also env vars related to proxying and (e.g. `NO_PROXY`), which are read by dependencies. We should double check which further non-CLI env vars are taken into account.
- The `font-paths` list should probably also list which system paths are searched if `TYPST_IGNORE_SYSTEM_FONTS` is not configured otherwise. The final order should be the order of precedence.

I've brainstormed with @reknih about the shape of the output and we've arrived at this:

Structured:
```yaml
version: 0.13.1
build:
  commit: 1e1d943d9d493149e8db91f4956fa54f256b7893
  settings:
    http-server: true
    self-update: false
features:
  html: true
paths:
  font-paths: []
  package-path: /Users/laurenz/Library/Application Support/typst/packages
  package-cache-path: /Users/laurenz/Library/Caches/typst/packages
env:
  TYPST_CERT: null
  TYPST_FEATURES: html
  TYPST_FONT_PATHS: null
  TYPST_IGNORE_SYSTEM_FONTS: null
  TYPST_PACKAGE_CACHE_PATH: null
  TYPST_PACKAGE_PATH: null
  TYPST_ROOT: null
  TYPST_UPDATE_BACKUP_PATH: null
  SOURCE_DATE_EPOCH: null
```
Human readable:
```
Version 0.13.1 (76f9106d)
  
Build settings
  self-update off (Update Typst via `typst update`)
  http-server on  (Serve HTML with `typst watch`)

Features
  html on (Experimental HTML export)

Paths
  Font paths
  - /Users/laurenz/Desktop/fonts
  - /Users/laurenz/Library/Fonts
  - /Library/Fonts
  - /System/Library/Fonts
  Package path       /Users/laurenz/Library/Application Support/typst/packages
  Package cache path /Users/laurenz/Library/Caches/typst/packages

Environment Variables
  TYPST_CERT                <unset>
  TYPST_FEATURES            html
  TYPST_FONT_PATHS          /Users/laurenz/Desktop/fonts
  TYPST_IGNORE_SYSTEM_FONTS <unset>
  TYPST_PACKAGE_CACHE_PATH  <unset>
  TYPST_PACKAGE_PATH        <unset>
  TYPST_ROOT                <unset>
  TYPST_UPDATE_BACKUP_PATH  <unset>
  SOURCE_DATE_EPOCH         <unset>
```
In particular, we wanted to lean a bit more into the human-readableness in the human readable output, with a few explanations in parentheses (could potentially add more later). For the Paths section, we've opted to remove the hyphen as it (at least as we envision it) does not just print the CLI args / env var, but rather displays the resolved version of the way Typst combines all the paths.

tingerrr:
Listing the system font directories is tough. We'd have to keep it up to date with `fontdb`, which doesn't expose these from the `Database` yet, the paths are hardcoded into `Database::load_system_fonts`.

I think we can add this in post if we decided to add it upstream. For now, I'd add special `<system fonts>` and `<embedded fonts>` entries in the human-readable output only to make the precedence clear.

Regarding the other environment variables, I've looked at the transitive dependencies of Typst and without looking at all of them, these are some that come to mind:
- `NO_COLOR` in one of the terminal output crates
- `NO_PROXY` as you said, probably all variations of `HTTPS_PROXY`, `HTTP_PROXY` and so forth too
- whatever else the HTML export server may have for configuration

I still have to check these, if you have others that may come to mind do tell, and I'll check those too.

laurmaedje:
> Listing the system font directories is tough. We'd have to keep it up to date with fontdb, which doesn't expose these from the Database yet, the paths are hardcoded into Database::load_system_fonts.

Makes sense. Let's skip this for now then. I was also thinking that if we ever switch to using system APIs for system font selection, that wouldn't make sense anymore anyway. But on the other hand, I think it can be useful for debugging while we don't have that and it's not clear whether we'll ever switch to system APIs.

> whatever else the HTML export server may have for configuration

From a very quick look, it seems like it doesn't read any, but I might've missed something.

It's also not the end of the world if we miss one, but the ones we do think of would be good to add.


I just checked with dtrace on macOS which vars are accessed by typst compile.
```bash
sudo dtrace -n 'pid$target::getenv:entry { printf("%s\n", copyinstr(arg0)); }' -c 'typst compile file.typ'
```

On, Linux the command should be this according to ChatGPT:
```bash
strace -e getenv typst compile file.typ
```

Since this is dynamic, to get good coverage, you need to make sure to fetch a package that's not cached, have errors in your document so that codespan-reporting is triggered, (maybe more?)

```
TYPST_ROOT
TYPST_FONT_PATHS
TYPST_PACKAGE_PATH
TYPST_PACKAGE_CACHE_PATH
SOURCE_DATE_EPOCH
TYPST_FEATURES
TYPST_UPDATE_BACKUP_PATH
TYPST_CERT
NO_COLOR
TZ
TERM
HOME
RAYON_NUM_THREADS
RAYON_RS_NUM_CPUS
RUST_MIN_STACK
```

I think the six at the bottom are pretty niche though and probably don't need to be listed.

I'm not sure why `NO_PROXY` wasn't listed though. Perhaps, the Linux command is insightful? I don't have a machine to test right now.

laurmaedje:
I just checked out the latest changes and it's very nice! Two remarks:

- Unlike in our proposal, you put the build settings directly into `build`, rather than into `build.settings` in the structured output. Was this intentional?

- `paths.system` does not really make sense as it's not clear that these are font paths. Perhaps the paths section should be split into `Fonts` and `Packages` like this?

  ```
  Fonts
    Custom font paths
      - /Users/laurenz/Desktop/fonts
      - /Users/laurenz/Library/Fonts
      - /Library/Fonts
      - /System/Library/Fonts
    System fonts   on
    Embedded fonts on

  Packages
    Package path       /Users/laurenz/Library/Application Support/typst/packages
    Package cache path /Users/laurenz/Library/Caches/typst/packages
  ```

  ```yaml
  fonts:
    paths: []
    system: true
    embedded: true
  packages:
    path: /Users/laurenz/Library/Application Support/typst/packages
    cache-path: /Users/laurenz/Library/Caches/typst/packages
  ```

tingerrr:
> Unlike in our proposal, you put the build settings directly into build, rather than into build.settings in the structured output. Was this intentional?

Nope, I missed that.

Turns out there are a few more we could find interesting, like:
- `FONTCONFIG_FILE` (typst, presumably the statically linked fontdb code)
- the aforementioned proxy variables in upper and lower case (typst)
  - `ftp_proxy`
  - `http_proxy`
  - `https_proxy`
  - `all_proxy`
  - `no_proxy`
- `OPENSSL_CONF` (libcrypto)
- `SSL_CERT_FILE` (libcrypto)

And at this point tracing with `ltrace` got so slow that it was simply stuck.

Do we include only the uppercase HTTP, HTTPS, ALL and NONE proxy? Typst doesn't use FTP, and everyone on earth uses the uppercase variables or sets both to the same value. `SSL_CERT_FILE` is probably the overwritten by us reading `TYPST_CERT`. `FONTCONFIG_FILE` and `OPENSSL_CONF` may be of interest. My binary also read `NIX_SSL_CERT_FILE`, but I assume this is a NixOS specific patch to libcrypto.

I've gone ahead with those changes, and you can tell me if you want something else changed.

crates/typst-cli/src/info.rs
```diff
@@ -0,0 +1,525 @@
...
+use crate::CliArguments;
+use crate::terminal::{self, TermOut};
+use clap::builder::{FalseyValueParser, TypedValueParser};
+use clap::{CommandFactory, ValueEnum};
+use codespan_reporting::term::termcolor::{Color, ColorSpec, WriteColor};
+use ecow::eco_format;
+use serde::Serialize;
+use typst::diag::StrResult;
+
+use crate::args::{Feature as CliFeature, InfoCommand};
```
laurmaedje: could you sort these imports?

crates/typst-cli/src/info.rs
```diff
@@ -0,0 +1,525 @@
...
+    typst_font_paths: Option<String>,
+    typst_ignore_system_fonts: Option<String>,
+    typst_package_cache_path: Option<String>,
+    typst_package_path: Option<String>,
+    typst_root: Option<String>,
+    typst_update_backup_path: Option<String>,
+    #[cfg(target_os = "linux")]
+    xdg_cache_home: Option<String>,
+    #[cfg(target_os = "linux")]
+    xdg_data_home: Option<String>,
```
laurmaedje: perhaps these should be unix without macOS rather than linux?

crates/typst-cli/src/info.rs
```diff
@@ -0,0 +1,525 @@
...
+    source_date_epoch: Option<String>,
+    fontconfig_file: Option<String>,
```
laurmaedje, line 168: this shoud also be unix without macOS I think

crates/typst-cli/src/info.rs
```diff
@@ -0,0 +1,525 @@
...
+    typst_ignore_system_fonts: Option<String>,
+    typst_package_cache_path: Option<String>,
+    typst_package_path: Option<String>,
+    typst_root: Option<String>,
+    typst_update_backup_path: Option<String>,
+    #[cfg(target_os = "linux")]
+    xdg_cache_home: Option<String>,
+    #[cfg(target_os = "linux")]
+    xdg_data_home: Option<String>,
+    source_date_epoch: Option<String>,
```
laurmaedje: let's maybe move this above the OS-specific ones

crates/typst-cli/src/info.rs
```diff
@@ -0,0 +1,525 @@
...
+            ("TYPST_PACKAGE_PATH", typst_package_path),
+            ("TYPST_ROOT", typst_root),
+            ("TYPST_UPDATE_BACKUP_PATH", typst_update_backup_path),
+            #[cfg(target_os = "linux")]
+            ("XDG_CACHE_HOME", xdg_cache_home),
+            #[cfg(target_os = "linux")]
+            ("XDG_DATA_HOME", xdg_data_home),
+            ("SOURCE_DATE_EPOCH", source_date_epoch),
+            ("FONTCONFIG_FILE", fontconfig_file),
+            ("SSL_CONF", ssl_conf),
```
laurmaedje: in your comment it was `OPENSSL_CONF`. is the difference intentional? also, I wonder whether this should also be unix-only since Windows and macOS builds don't use OpenSSL.

crates/typst-cli/src/info.rs
```diff
@@ -0,0 +1,525 @@
...
+    })
+}
+
+/// Turns a comma separated list of feature names into a well typed struct of
+/// feature flags.
+fn parse_features(feature_list: &str) -> StrResult<Features> {
+    let mut features = Features { html: false, a11y_extras: false };
+
+    for feature in feature_list.split(',').filter(|s| !s.is_empty()) {
+        match CliFeature::from_str(feature, true) {
```
laurmaedje: Maybe `CliFeature` should be renamed to `RuntimeFeature` or be kept as just `Feature`

laurmaedje (commented):
The env var selection looks good to me now. Just a few minor comments.

crates/typst-cli/src/info.rs
tingerrr: It probably should be, I was wondering whether this also applies to the `*_PROXY` variables. And no, that was a copy-paste mistake, I'll fix that.

crates/typst-cli/src/info.rs
tingerrr: Up to you, I'd leave that out of the PR for you or someone else to change down the line, I just renamed it here because I initially had `KeyValueDesc` be named `Feature` and after I generalized it I forgot to remove that import rename.

crates/typst-cli/src/info.rs
tingerrr: `source_date_epoch` I assume? GitHub shows more than one line here.

crates/typst-cli/src/info.rs
tingerrr: I thought so too initially, but as it turns out, `dirs` does not use the `XDG_*` vars on macOS.

tingerrr: Ah, my bad,  I read this as `with`.

laurmaedje:
Thank you! I think this turned out very nicely. :)

PR #3593, opened 2024-03-09
Spruce up table docs

reknih:
Does not contain the tables guide yet.

PR #6764, opened 2025-08-15
Improve docs of `Cell<T>` params of `table` and `grid`

YDX-2147483647:
**(Not finished yet)**

- Add links to the table guide.
- Improve the general description about `Cell<T>` ([`$grid/#styling`](https://typst.app/docs/reference/layout/grid/#styling-the-grid)), and add links to it.
- Document arguments and returned values of functions, fields of dictionaries, and items of arrays.

Relates-to: #408

### DIFF
--- a/crates/typst-library/src/layout/grid/mod.rs
+++ b/crates/typst-library/src/layout/grid/mod.rs
@@ -100,7 +100,7 @@ use crate::visualize::{Paint, Stroke};
 /// )
 /// ```
 ///
-/// # Styling the grid
+/// # Styling the grid { #styling }
 /// The grid's appearance can be customized through different parameters. These
 /// are the most important ones:
 ///
@@ -174,9 +174,9 @@ pub struct GridElem {
 
     /// How to fill the cells.
     ///
-    /// This can be a color or a function that returns a color. The function
-    /// receives the cells' column and row indices, starting from zero. This can
-    /// be used to implement striped grids.
+    /// This can be a color or a [function that returns a color]($styling). The
+    /// function receives the cells' column and row indices, starting from zero.
+    /// This can be used to implement striped grids.
     ///
     /// ```example
     /// #grid(
--- a/crates/typst-library/src/model/table.rs
+++ b/crates/typst-library/src/model/table.rs
@@ -146,7 +146,7 @@ pub struct TableElem {
     ///
     /// This can be a color or a function that returns a color. The function
     /// receives the cells' column and row indices, starting from zero. This can
-    /// be used to implement striped tables.
+    /// be used to [implement striped tables]($guides/table-guide/#fills).
     ///
     /// ```example
     /// #table(
@@ -173,6 +173,9 @@ pub struct TableElem {
     /// The function receives the cells' column and row indices, starting from
     /// zero. If set to `{auto}`, the outer alignment is used.
     ///
+    /// The table guide has a
+    /// [dedicated section on alignment]($guides/table-guide/#alignment).
+    ///
     /// ```example
     /// #table(
     ///   columns: 3,
@@ -193,7 +196,8 @@ pub struct TableElem {
     /// [`table.hline`] and [`table.vline`] alongside your table cells.
     ///
     /// See the [grid documentation]($grid.stroke) for more information on
-    /// strokes.
+    /// strokes, or the [table guide]($guides/table-guide/#strokes) for
+    /// practical usages.
     #[fold]
     #[default(Celled::Value(Sides::splat(Some(Some(Arc::new(Stroke::default()))))))]
     pub stroke: Celled<Sides<Option<Option<Arc<Stroke>>>>>,


crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -100,33 +100,68 @@ use crate::visualize::{Paint, Stroke};
...
+///   fill: (rgb("#239dad50"), none),
+///   // By a function that returns a value
+///   stroke: (x, y) => if calc.rem(x + y, 3) == 0 { 0.5pt },
 ///
-/// Locating most of your styling in set and show rules is recommended, as it
-/// keeps the grid's or table's actual usages clean and easy to read. It also
-/// allows you to easily change the grid's appearance in one place.
+///   ..range(5 * 3).map(n => numbering("A", n + 1))
+/// )
+/// ```
```
YDX-2147483647, line 152: Image

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -194,26 +261,76 @@ pub struct GridElem {
...
+    /// #grid(
+    ///   stroke: (x, y) => if x == 0 { (right: green) },
+    ///   ..placeholder
+    /// )
+    ///
+    /// #grid(
+    ///   stroke: (x, y) => if x >= 1 { (left: purple) },
+    ///   ..placeholder
+    /// )
+    /// ```
```
YDX-2147483647: Image

crates/typst-library/src/layout/container.rs
```diff
@@ -69,6 +69,17 @@ pub struct BoxElem {
 
     /// How much to pad the box's content.
     ///
+    /// This can be:
+    /// - a single length for all sides
```
laurmaedje: note that this can also be a relative length

laurmaedje: perhaps outset should link here for the dictionary shape?

crates/typst-library/src/layout/container.rs
```diff
@@ -69,6 +69,17 @@ pub struct BoxElem {
...
+    /// - a dictionary of lengths for individual sides
+    ///
+    /// When passing a dictionary, it can contain the following keys in order of
+    /// precedence: `top`, `right`, `bottom`, `left` (controlling the respective
+    /// cell sides), `x`, `y` (controlling vertical and horizontal insets), and
+    /// `rest` (covers all insets not styled by other dictionary entries). All
+    /// keys are optional; omitted keys will be treated as if their value was
+    /// the default value.
```
laurmaedje: not necessarily the default. it could also be a previously set value.

crates/typst-library/src/layout/grid/mod.rs
laurmaedje: I think this example section is a bit long. I think it might be better to keep it a bit more reduced here.

Also, I think it's important that examples don't just demonstrate something, but are also visually appealing, which isn't really the case here.

crates/typst-library/src/layout/container.rs
YDX-2147483647: > All keys are optional; omitted keys will be treated as if their value was the default value.
> —[How to adjust the lines in a table, Table guide](https://typst.app/docs/guides/table-guide/#double-stroke:~:text=omitted%20keys%20will%20be%20treated%20as%20if%20their%20value%20was%20the%20default%20value.)

Good point, but this sentence is literally copied from the table guide. I will change both later.

crates/typst-library/src/layout/grid/mod.rs
YDX-2147483647: I now doubt whether these new examples for `grid.stroke` are really necessary.
I usually use `grid` to get 1fr/auto columns, not `stroke`. For most cases that need to customize `stroke`, `table` should be used.

In addition, can we assume that `grid` is more _advanced_ than `table`? 
I added these examples because I thought the `#let cv(..jobs)` example below is too complex and simpler examples are needed.
However, if most people using grid have read the table guide before, then simpler examples won't be necessary.

YDX-2147483647: Okay I finally come up with a better example for `grid.stroke`. [`Grid2aib.svg` on Wikimedia Commons](https://commons.wikimedia.org/wiki/File:Grid2aib.svg) in public domain inspires me.

Within a limited space, stroke has to be used with spacing to organize the content. The example also demonstrates that stroke does not cross `gutter`.

图片

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -203,26 +272,53 @@ pub struct GridElem {
...
+    ///   columns: (1fr, 2fr, 2fr),
+    ///   row-gutter: 1.5em,
+    ///   inset: (left: 0.5em),
+    ///   stroke: (x, y) => if x > 0 { (left: 0.5pt + gray) },
+    ///
+    ///   ..(
+    ///     [Winter \ 2007 \ Season],
+    ///     [Aaron Copland \ *The Tender Land* \ January 2007],
+    ///     [Eric Satie \ *Gymnopedie 1, 2* \ February 2007],
+    ///   ).map(text.with(1.3em)),
```
laurmaedje: Perhaps this could instead be styled with?
```typ
#show grid.cell.where(y: 0): set text(size: 1.3em)
```

The `..` + `.map` + `.with` is probably more intimidating than that.

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -203,26 +272,53 @@ pub struct GridElem {
...
+    ///
+    /// To specify varying strokes for different cells, you can:
+    /// - use a single stroke for all cells
+    /// - use an array of strokes corresponding to each column
+    /// - use a function that maps a cell's position to its stroke
+    ///
+    /// See the [styling section](#styling) above for more details.
+    ///
+    /// ```example
+    /// #set page(width: 420pt)
```
laurmaedje, line 293: I like this example visually. However, I think it is not super idiomatic because all cells have the same structure and yet they are all formatted manually. I'm not sure if automating that would go too far away from the intent of the example, but I think we should avoid unidiomatic examples.

crates/typst-library/src/model/table.rs
```diff
@@ -142,11 +142,71 @@ pub struct TableElem {
...
+    /// How to align the cells' content.
+    ///
+    /// If set to `{auto}`, the outer alignment is used.
+    ///
+    /// You can specify the alignment in any of the following fashions:
+    /// - use a single alignment for all cells
+    /// - use an array of alignments corresponding to each column
+    /// - use a function that maps a cell's X/Y position (both starting from
+    ///   zero) to its alignment
+    ///
```
laurmaedje, line 190: I inserted newlines after all lists because otherwise the sentence will belong to the last list item in Markdown.

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -104,33 +104,67 @@ use crate::visualize::{Paint, Stroke};
...
+/// On top of that, you may [apply styling rules]($styling) to [`grid`] and
+/// [`grid.cell`]. Especially, the [`x`]($grid.cell.x) and [`y`]($grid.cell.y)
+/// fields of `grid.cell` can be used in a [`where`]($function.where) selector,
+/// making it possible to style cells at specific columns or rows, or individual
+/// positions.
+///
+/// Locating most of your styling in set and show rules is
+/// recommended, as it keeps the grid's or table's actual usages clean and easy
+/// to read. It also allows you to easily change the grid's appearance in one
+/// place.
```
laurmaedje: For tables, 100%. For grids, a bit less so because they are not semantic so a set rule could potentially affect implementation details with other grids within your grid. Maybe, we should move this to the table docs?

crates/typst-library/src/layout/grid/mod.rs
laurmaedje: Though I am _sort of_ contradicting my other comment on grids not being semantic and show/set thus being a bit more dangerous with nesting.

laurmaedje (commented):
I pushed some tweaks, but also left some more comments.

crates/typst-library/src/layout/grid/mod.rs
YDX-2147483647: Agreed. I didn't do this because _Winter 2007 Season_ was smaller than other cells at the beginning.

crates/typst-library/src/layout/grid/mod.rs
YDX-2147483647: Well, I intentionally format the example manually.
Note the manual line break in `*Middlebury College \ Center for the Arts*`. It balances the two lines.

```
// Break manual 
Middlebury College
Center for the Arts

// Break automatically by Typst
Middlebury College Center
for the Arts
```

How about making this example more difficult to automate? For example:

Image


<details><summary>Code</summary>

```typst
#grid(
  columns: (1fr, 2fr, 2fr),
  row-gutter: 1.5em,
  inset: (left: 0.5em),
  stroke: (x, y) => if x > 0 { (left: 0.5pt + gray) },
  align: horizon,

  ..(
    [Winter \ 2007 \ Season],
    [Aaron Copland \ *The Tender Land* \ January 2007],
    [Eric Satie \ *Gymnopedie 1, 2* \ February 2007],
  ).map(text.with(1.3em)),

  [],
  [Jan 12 \ *Middlebury College \ Center for the Arts* \ 20:00],
  [Feb 2 \ *Johnson State College Dibden Center for the Arts* \ 19:30],

  [],
  [Skip a week \ #text(0.8em)[_Prepare your exams!_]],
  [Feb 9 \ *Castleton State College \ Fine Arts Center* \ 19:30],

  [],
  [Jan 26, 27 \ *Lyndon State College Alexander Twilight Theater* \ 20:00],
  [Feb 17 --- #smallcaps[Anniversary] \ *Middlebury College \ Center for the Arts* \ 19:00 #text(0.7em)[(for a special guest)]],
)
```

</details>

crates/typst-library/src/layout/grid/mod.rs
YDX-2147483647: I don't know… This paragraph was added by reknih in #3593, exclusively for `grid`.
Anyway, show rules for grids are helpful for debugging.

PgBiel: I agree with Laurenz, we shouldn't encourage set rules on grids due to semantics. A good opportunity for a correction.

crates/typst-library/src/layout/grid/mod.rs
laurmaedje: @PgBiel curious to hear your thoughts on this situation w.r.t best practice

crates/typst-library/src/layout/grid/mod.rs
laurmaedje: Okay, I see. I'm fine with that. 

The last line is a little long now, it would be better to break it up.

crates/typst-library/src/layout/grid/mod.rs
YDX-2147483647: Hmm, I made a try but failed to find a proper place in `table.rs`…

`$grid/#styling` is a quick summary of the table guide. The content actually applies to both grids and tables. How about changing this paragraph to the following?

> Locating most of your styling in set and show rules is recommended, as it keeps the grid's or table's actual usages clean and easy to read. For tables, it also allows you to easily change the appearance of all tables in one place. However, grids are not semantic and top level rules may generate unwanted effects. It is better to restrict the effect of grid rules by [putting the rules and the grid inside of a block](https://typst.app/docs/reference/styling/#set-rules), and prefer other methods for nested grids.

laurmaedje: Even if you scope the rules, it could in theory affect content nested in the grid if you're writing a generic component and don't control the cell content. Hence also [this comment](https://github.com/typst/typst/pull/6764#discussion_r2333296080).

YDX-2147483647: I think show/set rules are still okay for simple grids? I edit my previous comment with the following

> …and prefer other methods for nested grids.

How about now?

crates/typst-library/src/layout/grid/mod.rs
PgBiel: Both options are used in practice, though without set rules you normally use a function that handles it all instead (so I don't see that specific example very often). So showcasing set rules in a scope is probably best. Notably, this is basically the main purpose of being able to `show grid.cell`, which is to more easily style content in cells.

crates/typst-library/src/layout/grid/mod.rs
laurmaedje: That unfortunately doesn't really fix it in my opinion because the problem is not with the nested grid itself, but with the _parent_ grid. If I write `set grid(stroke: ..); grid(my-component(..), my-component(..)` and `my-component` happens to use a `grid` that can cause problems.

I would suggest removing this here entirely and adding text after

> Although the `table` and the `grid` share most properties, set and show rules on one of them do not affect the other.

in the table docs, which recommend set rules for tables, but not for grids.

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -203,26 +272,57 @@ pub struct GridElem {
...
+    ///   [Jan 12 \ *Middlebury College \ Center for the Arts* \ 20:00],
+    ///   [Feb 2 \ *Johnson State College Dibden Center for the Arts* \ 19:30],
+    ///
+    ///   [],
+    ///   [Skip a week \ #text(0.8em)[_Prepare your exams!_]],
+    ///   [Feb 9 \ *Castleton State College \ Fine Arts Center* \ 19:30],
+    ///
+    ///   [],
+    ///   [Jan 26, 27 \ *Lyndon State College Alexander Twilight Theater* \ 20:00],
+    ///   [Feb 17 --- #smallcaps[Anniversary] \ *Middlebury College \ Center for the Arts* \ 19:00 #text(0.7em)[(for a special guest)]],
```
laurmaedje:
```suggestion
    ///   [Feb 17 --- #smallcaps[Anniversary] \ *Middlebury College \ 
    ///    Center for the Arts* \ 19:00 #text(0.7em)[(for a special guest)]],
```

YDX-2147483647: I break it into the following, because I think dividing the pair of `*` into two lines is kind of crazy.

```rust
///   [
///     Feb 17 --- #smallcaps[Anniversary] \
///     *Middlebury College \ Center for the Arts* \
///     19:00 #text(0.7em)[(for a special guest)]
///   ],
```

laurmaedje:
Thank you!

PR #6766, opened 2025-08-15
Use `stylistic-set: 1` in favor of `ss01` in docs and tests

YDX-2147483647:
They are equivalent, but `stylistic-set: 1` is less cryptic than `features: ("ss01",)`. (I thought `ss` means script style…)

_Stylistic Set 1_ is the “friendly name” of `ss01`.
Ref: [Registered features, p-t (OpenType 1.9.1) - Typography | Microsoft Learn](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#tag-ss01---ss20)


laurmaedje:
Thanks!

PR #6771, opened 2025-08-16
Add initial support for variable fonts

### LINKED ISSUE #185: Having trouble select weight for variable font
for example, https://fonts.google.com/specimen/Bitter, if have the variable font version `Bitter-VariableFont_wght.ttf` installed, selecting font weight by `#set text(font: "Bitter", weight: "medium")` would still render the ExtraLight variant and setting other weight options would not change the font weight appeared in the PDF output anyway.

LaurenzV:
This PR is still in-progress and aims to add some initial support for variable fonts. `Initial` because the aim is not to give the user full control over setting custom variation coordinates (this could be good future work, but probably makes more sense to implement in conjunction with the planned `font` object). The aim is to automatically select a correct instance based on the `wght` (font weight), `wdth` (font stretch), and `ital`/`slnt` (italic) axes, which I think should over 95%+ of the use cases.

It also adds support for embedding CFF2 fonts in PDFs by converting them to a TTF font, which is not the best approach (better would be to convert to CFF), but it should do the job. CFF2 fonts are pretty rare ([less than 1% of variable fonts](https://almanac.httparchive.org/en/2024/fonts#variable-fonts)), but based on past issues it seems like some systems do use some CFF2-based Noto fonts, so this is definitely worth fixing.

Fixes #185 (tracking support for specifying variable font axes is probably worth opening a new issue for).

Example:
```
#for (font, d_text) in (
    ("Cantarell", "I love using variable fonts!"),
    ("Noto Sans CJK SC", "我太喜欢使用可变字体了!"),
    ("Roboto", "I love using variable fonts!"),
) {
    [= #font]
    for i in range(100, 900, step: 100) {
        text(weight: i, font: font)[#d_text \ ]
    }
}
```

Result:
[test.pdf](https://github.com/user-attachments/files/21813827/test.pdf)

TODOs:

- Properly hook up the `wdth` and `slant`/`italic` axes.
- Bulk test with maany fonts to ensure it works as expected.
- Fix or find a workaround for the `pixglyph` bug that makes CFF2 fonts render as black rectangles in PNG export.
- Investigate whether the code leads to regressions in SVG rendering, where variable fonts are currently not supported.
- Add test cases
- Clean up code + documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -958,21 +958,19 @@ dependencies = [
 
 [[package]]
 name = "hayro"
-version = "0.1.0"
-source = "git+https://github.com/LaurenzV/hayro?rev=e701f95#e701f9569157a2fe4ade68930dc9e9283782dcca"
+version = "0.2.0"
+source = "git+https://github.com/LaurenzV/hayro?rev=cbebe75#cbebe75cbd131b97349a2f78507c508c0d01390d"
 dependencies = [
  "bytemuck",
  "hayro-interpret",
  "image",
  "kurbo",
- "rustc-hash",
- "smallvec",
 ]
 
 [[package]]
 name = "hayro-font"
 version = "0.1.0"
-source = "git+https://github.com/LaurenzV/hayro?rev=e701f95#e701f9569157a2fe4ade68930dc9e9283782dcca"
+source = "git+https://github.com/LaurenzV/hayro?rev=cbebe75#cbebe75cbd131b97349a2f78507c508c0d01390d"
 dependencies = [
  "log",
  "phf",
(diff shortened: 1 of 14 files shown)


khaledhosny:
> The aim is to automatically select a correct instance based on the `wght` (font weight), `wdth` (font stretch), and `ital`/`slnt` (italic) axes, which I think should over 95%+ of the use cases.

Maybe add `opsz` to the mix.

LaurenzV:
I might be wrong, but I was assuming that `opsz`, similar to hinting, is to improve the rendering at certain font sizes. Does it really make sense to use that when exporting to PDF?

khaledhosny:
> I might be wrong, but I was assuming that `opsz`, similar to hinting, is to improve the rendering at certain font sizes. Does it really make sense to use that when exporting to PDF?

Optical Size is a totally different concept. Hinting is about pixels, but optical size is about adapting the design for the intended font size. The most widely used typeface with optical sizes is probably Computer Modern (no `opsz` axis of course). A font designed for 5pt will have less contrast and wider spacing, while a font designed for, say, 120pt will have much more contrast and more compact spacing.

LaurenzV:
I see, thanks!

ethanc8:
> Initial because the aim is not to give the user full control over setting custom variation coordinates (this could be good future work, but probably makes more sense to implement in conjunction with the planned font object).

Will the user be able to manually select these using the `features` parameter in `#text`?

LaurenzV:
I don't think I will be working on this in the near future (and I also think it might be better to wait for the proper font rework and then do a proper implementation), so I'll close this in case someone else wants to work on this.

laurmaedje:
Okay, thank you still for getting the ball rolling and working on the necessary infrastructure on the PDF side!

PR #6773, opened 2025-08-16
Fix `repr` of `foo.with(..)`

MDLC01:
Currently, the following code displays `foo`.

```typ
#let foo(a) = {}
#repr(foo.with(1))
```

This is undesirable, because the value that is being represented is not actually the function `foo`, but rather a partially applied version. I would argue that two reprs that do not contain `..` should be equal iff the underlying values are equal.

I went for `(..) => ..`. An alternative is `foo.with(..)`, which conveys more information. I prefer `(..) => ..` because it makes it clearer that it is a function.


laurmaedje:
Thanks!

PR #6776, opened 2025-08-17
Make links to `$grid` in `table` docs more specific

YDX-2147483647:
This PR is extracted from #6764.


crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -36,7 +36,11 @@ use crate::visualize::{Paint, Stroke};
 /// in the document flow. Set and show rules on one of these elements do not
 /// affect the other.
 ///
+/// # Sizing the tracks { #track-size }
+///
 /// A grid's sizing is determined by the track sizes specified in the arguments.
+/// There are many such sizing parameters: [`columns`]($grid.columns),
+/// [`rows`]($grid.rows), [`gutter`]($grid.gutter), etc. and those of [table].
```
PgBiel: Table is irrelevant here, it is already mentioned above that table works the same way in terms of layout, but for a different purpose.

```suggestion
/// There are multiple sizing parameters: [`columns`]($grid.columns),
/// [`rows`]($grid.rows) and [`gutter`]($grid.gutter).
```

crates/typst-library/src/layout/grid/mod.rs
```diff
@@ -143,19 +147,25 @@ pub struct GridElem {
...
 
     /// The gaps between rows and columns.
     ///
     /// If there are more gutters than defined sizes, the last gutter is
     /// repeated.
     ///
+    /// See the [track size section](#track-size) above for more details.
+    ///
     /// This is a shorthand to set `column-gutter` and `row-gutter` to the same
     /// value.
```
PgBiel: I feel like those two paragraphs / blocks should be swapped

PgBiel (commented):
Looks reasonable in principle. Some comments:

crates/typst-library/src/layout/grid/mod.rs
YDX-2147483647: I'm also struggling with it. The track size section provides futher details about _If there are more gutters than defined sizes, the last gutter is repeated_, so I think it should be put directly after that sentence.

But if you insist on swapping them, I'll do so.

PgBiel: Another idea is to move "This is a shorthand to set `column-gutter` and `row-gutter` to the same value." above "If there are more gutters (...)". I think this would be more immediately useful to the reader too.

YDX-2147483647: Fine!

laurmaedje:
Thanks!

PR #6778, opened 2025-08-17
Update & fix tutorial part 3 & 4 example code

SillyFreak:
As can be seen on the final image of [Making a Template](https://typst.app/docs/tutorial/making-a-template/#separate-file), there are a few discrepancies between it and [part 3](https://typst.app/docs/tutorial/advanced-styling/#review): level 1 headers are not centered and the title is not bold. Also, the title has changed.

This PR harmonizes the various code snippets in parts 3 and 4 of the tutorial, hopefully making it easier to update them in the future while keeping consistency, and easier to follow for readers successively copy/pasting together the shown code.

Cleanup changes:
- eliminate some indentation differences, e.g. `>>>  margin: auto,` to `>>> margin: auto,`
- make show rules of the example documents appear in a consistent order (page -> par -> text)

In part 4, base all code examples on what was built in part 3. For example, replace
```
>>>  show heading.where(
>>>    level: 1
>>>  ): it => block(
>>>    align(center,
>>>      text(
>>>        13pt,
>>>        weight: "regular",
>>>        smallcaps(it.body),
>>>      )
>>>    ),
>>>  )
```
with
```
>>> show heading.where(
>>>   level: 1
>>> ): it => block(width: 100%)[
>>>   #set align(center)
>>>   #set text(13pt, weight: "regular")
>>>   #smallcaps(it.body)
>>> ]
```
These changes
- make headings centered as they're supposed to be
- make the paper title bold
- add missing numbering to page show rule in first big example
- reflect the use of `place(...)` in the template that was introduced in part 3

Additional improvements:
- in the heading show rule example, show the actual heading lines (`= Introduction`, `== Motivation`) since the example is supposed to demonstrate that the show rule doesn't treat the different levels separately
- use a code block instead of a content block for the `place(...)` when adapting it from part 3 for part 4, similar to how it was done for the template function itself
- rename the paper from "Towards Improved Modelling" to "A fluid dynamic model for glacier flow" for consistency

I have manually pasted code from the new examples to try to confirm that everything is correct, but I have naturally not compiled the exact code from the markdown files or seen it rendered, so keep that in mind.

### DIFF
--- a/docs/tutorial/3-advanced.md
+++ b/docs/tutorial/3-advanced.md
@@ -404,10 +404,12 @@ a way to set any of that, we need to write our own heading show rule.
 >>>   ]
 >>> ]
 >>>
->>> = Introduction
+= Introduction
+<<< ...
 >>> #lorem(35)
->>>
->>> == Motivation
+
+== Motivation
+<<< ...
 >>> #lorem(45)
 ```
 
--- a/docs/tutorial/3-advanced.md
+++ b/docs/tutorial/3-advanced.md
@@ -51,7 +51,7 @@ Let's start by writing some set rules for the document.
 
 ```example
 #set page(
->>>  margin: auto,
+>>> margin: auto,
   paper: "us-letter",
   header: align(right)[
     A fluid dynamic model for
@@ -203,14 +203,14 @@ keyword:
 >>> #set text(font: "Libertinus Serif", 11pt)
 >>> #set par(justify: true)
 #set page(
->>>   "us-letter",
->>>   margin: auto,
+>>> "us-letter",
+>>> margin: auto,
   header: align(
     right + horizon,
     title
   ),
 <<<   ...
->>>   numbering: "1",
+>>> numbering: "1",
 )
 
 #align(center, text(17pt)[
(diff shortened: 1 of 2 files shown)


docs/tutorial/3-advanced.md
```diff
@@ -404,10 +404,12 @@ a way to set any of that, we need to write our own heading show rule.
 >>>   ]
 >>> ]
 >>>
```
laurmaedje: would be nice if this or the newline below ... would be `>>>`. currently it renders like this:
```
...
= Introduction
...
```

docs/tutorial/4-template.md
```diff
@@ -240,30 +236,40 @@ The resulting template function looks like this:
...
-  par(justify: false)[
-    *Abstract* \
-    #abstract
+>>> place(
+>>>   top + center,
+>>>   float: true,
+>>>   scope: "parent",
+>>>   clearance: 2em,
+>>> )[
+<<<   place(...)[
```
laurmaedje: Not sure whether it's worth to hide these. Currently, it looks a little odd and doesn't really help clarity much I think:
```
  place(
    ...,
    {
```

docs/tutorial/4-template.md
```diff
@@ -332,30 +334,31 @@ call.
 >>>     float: true,
 >>>     scope: "parent",
 >>>     clearance: 2em,
->>>   )[
->>>     #text(
->>>       17pt,
->>>       weight: "bold",
->>>       title,
->>>     )
+>>>     {
```
laurmaedje, line 339: The double indent is a bit annoying but I agree that it's more idiomatic in principle ...

laurmaedje (commented):
Thanks for the individual commits. That really helped with understanding what's changed.

docs/tutorial/4-template.md
ensko: yeah, trailing code blocks work a bit better in this regard if you can put all the options in one line:
```typc
place(top + center, float: true, scope: "parent", clearance: 2em, {
  ...
})
```
... but that doesn't work for when the code isn't hidden (line is too long) and I tried to keep these consistent between different snippets.

Do you want any change there? Would relate to both
https://github.com/typst/typst/blob/3309123ce4e3b89f72b63c42f060a638851498a7/docs/tutorial/4-template.md?plain=1#L245-L250
and
https://github.com/typst/typst/blob/3309123ce4e3b89f72b63c42f060a638851498a7/docs/tutorial/4-template.md?plain=1#L334-L339

ensko:
I'm glad the clear commits helped :)

docs/tutorial/4-template.md
laurmaedje: No, it's okay.

laurmaedje:
Thanks!

Andrew15-5:
> hopefully making it easier to update them in the future

Yeah no. It will be [rewritten completely soon](https://github.com/typst/typst/pull/6417#issuecomment-3052631676). Though by now it doesn't look soon to me.

PR #6779, opened 2025-08-17
Fix typo in doc on quotes parameter of smartquote

ultimatile:
In the `smartquote` documentation for the [quotes](https://typst.app/docs/reference/text/smartquote/#parameters-quotes) parameter, the dictionary item currently says "an array" but it should read "a dictionary".


laurmaedje:
Thanks!

PR #6780, opened 2025-08-17
Add default named arg to json loader

Its-Just-Nans:
Hello,

This allows the user to handle a possible file not found error

I can also do the same for other loader if needed
Maybe also for the `read()` call

Thanks


laurmaedje:
I find the current profileration of `default` arguments on accessor functions on arrays, dictionaries, and such already somewhat unfortunate. Those cover cases that would be `None` in Rust. I don't really want to extend `default` to also cover `Err` cases.

I'm with you that some solution is necessary, but I think for this case #2025 is planned and would be a better option. Thank you still!

Issue #6314, opened 2025-05-21
Make default `counter.display` respect numbering of the selector


Andrew15-5:
### Description

[In docs](https://typst.app/docs/reference/introspection/counter/#definitions-display-numbering):

> If this is omitted or set to `auto`, displays the counter with the numbering style for the counted element or with the pattern `"1.1"` if no such style exists.

Fundamentally in `counter(heading.where(level: 1)).display()` the counted element is `heading`, but since the output can change if the `where` clause is dropped, then I will consider the counted element to be `heading.where(level: 1)`, and as such, it must use the `heading.where(level: 1).numbering`, and not `heading.numbering`.

This should display "A" and not "1":

```typ
#set heading(numbering: "1")
#show heading.where(level: 1): set heading(numbering: "A")

= this
#context counter(heading.where(level: 1)).display()
```

This might depend on https://github.com/typst/typst/issues/5574.

### Use Case

More intuitive counter behavior.

Andrew15-5:
It apparently already works with figure selector:

```typ
#show figure.where(kind: image): set figure(numbering: "A")
#show figure.where(kind: table): set figure(numbering: "I")
#show figure.caption: it => context {
  let n = if it.kind == image {
    [~] + counter(figure.where(kind: image)).display()
  } else if it.kind == table {
    [~] + counter(figure.where(kind: table)).display()
  }
  [#it.supplement#n#it.separator#it.body]
}

#figure(rect(), caption: lorem(15))

#figure(table[], caption: lorem(15))
```

![Image](https://github.com/user-attachments/assets/ee9487cb-a32b-4e6f-bfa5-abbbade61497)

So the scope is actually smaller than all counters. It's a subset of them.

EpicEricEE:
It only seems to work with the figure selector because you're doing the counter display _inside_ the figure, so the correct `figure.numbering` from the show-set rule is on the style chain. The if statement thus isn't actually doing anything. It would show the same even when just using `counter(figure).display()` or a counter with a different figure kind.

Andrew15-5:
The if statement only changes the kind, and the counter is different for different kind, so it **does do** what it is used for. The `counter(figure)` will show a different number.

> the correct `figure.numbering` from the show-set rule is on the style chain

This explains a lot. The `#show heading: it => counter(heading).display() + it` also works the same way, where display is using the numbering for the style chain, regardless of the counter name, but the different counter will again return different values.

[...]

Issue #1356, opened 2023-05-28
Feature request: more locatable elements


instanceofme:
### Description

It seems [only some elements are locatable](https://github.com/typst/typst/blob/aed0f1c/library/src/meta/query.rs#L107-L113), but lists for instance are not.

### Use Case

I was trying to make a reusable style with a title and some content, and automatically adjust the spacing if the content begins with a list.

IIUC one way to do that would have been:

```typ
#let myCustomStyle(
  title: "Untitled",
  body
) = {
  // [do something with title]

  // Remove spacing before next § unless it starts by a list
  locate(loc => {
    let lists = query(list.where().after(loc), loc)
    if lists==() {
      v(-.5em)
    }
  })

  body
}
```

But this yields `error: list is not locatable`. (This led me to #325 which looks only tangentially related.)

Another way could be the following, but `show` apparently [only accepts elem/label/regex selectors](https://github.com/typst/typst/blob/f4fd685/src/model/styles.rs#L490-L501) and even then I guess I'd still have the same non-locatable selector issue.
```typ
locate(loc => {
  show list.where().after(loc): it => {
    v(.5em)
    it
  }
})
```

This workaround seems to do the trick in the meantime:

```typ
let yState = state("y", 0pt)
locate(loc => {
  yState.update(y => loc.position().y)
})
show list.where(): it => {
  locate(loc => {
    if yState.at(loc) == loc.position().y {
      v(0.5em)
    }
  })
  it
}
```

magikmw:
I would love to have links of any kind locatable - I tried to make an index of links as a supplement, but it turned out I needed to add labels to every link, and I can't really have time for that in current project.

lemon-gith:
> I would love to have links of any kind locatable - I tried to make an index of links as a supplement, but it turned out I needed to add labels to every link, and I can't really have time for that in current project.

Yes, I'd just like to second that notion quickly. I'm currently writing my Thesis, and really stumped as to how I would add a "List of Links" section, because links are not visible in a printout.

bluss:
We can put labels on every link automatically and create an outline for links. This is just showing that it can be done, needs some work to make it nice.

```typst
#[
  #let llabels = state("all-my-link-labels", (:))

  // add a label to every link that doesn't already have one
  // record which labels links have.
  #show link: it => {
    if it.has("label") {
      // already has label
      llabels.update(map => { map.insert(str(it.label), 1); map })
      it
    } else {
[...]

PR #6781, opened 2025-08-18
Add `at` parameter to `counter.display`

mkorje:


laurmaedje:
The docs for `query` could be updated to use this. There might be more places.

mkorje:
That should be all of them.

tests/suite/introspection/counter.typ
```diff
@@ -49,7 +49,7 @@ In #context counter(heading).display()
 
 At Beta, it was #context {
   let it = query(heading).find(it => it.body == [Beta])
-  numbering(it.numbering, ..counter(heading).at(it.location()))
+  counter(heading).display-at(it.location(), it.numbering)
```
mkorje: I need to specify the numbering explicitly because of https://github.com/typst/typst/issues/6314 it seems.

Enivex:
I'll copy my question from discord:

> Does it need to be a separate method? Can't display just take an optional parameter? `display(at: there)`

crates/typst-library/src/introspection/counter.rs
```diff
@@ -469,6 +469,39 @@ impl Counter {
         self.display_impl(engine, loc, numbering, both, context.styles().ok())
     }
 
+    /// Displays the value of the counter with a numbering at the given location
+    /// and returns the formatted output.
+    #[func(contextual)]
+    pub fn display_at(
```
YDX-2147483647: `display_at_loc` is already available. Should there be a distinction?

https://github.com/typst/typst/blob/880e69b0c840cd0d2f0c51e72966ab72c008e713/crates/typst-library/src/introspection/counter.rs#L258-L271

mewmew:
> I'll copy my question from discord:
> 
> > Does it need to be a separate method? Can't display just take an optional parameter? `display(at: there)`

A follow-up PR could then introduce `get(at: loc)` for consistency. And potentially deprecate the `at` method in the future (or just leave it as a short-hand for `get(at: loc)`.

(This was also from the [Discord conversation](https://discord.com/channels/1054443721975922748/1088371867913572452/1407004625592127601). I liked the consistency of the suggested change.)

> I think `display()/display(at: loc)` and `get()/get(at: loc)` would be the cleanest API. The current `get()/at(loc)` and proposed `display()/display-at(loc)` feel less discoverable, less consistent, and more confusing

cAttte:
please please please do an `at` param. `display-at()` is so weird. having `at` as a param that defaults to `here()` just makes so much more sense, docs- and UX-wise, than even the existing `at()` method. this is evidenced by the common pattern of users unnecessarily writing `.at(here())` (e.g. [A](https://github.com/typst/typst/issues/1356#issue-1729585451), [B](https://forum.typst.app/t/how-to-customize-page-header-with-current-or-previous-level-1-headline-displayed-in-corner/1953?u=catte)) instead of `.get()`.

crates/typst-library/src/introspection/counter.rs
mkorje: Its an internal function I think, and takes a location directly instead of the current context.

mkorje:
Ok, I've switched to an `at` parameter instead. I also added an `at` parameter to `.get` in another commit and then deprecated `.at` in the last commit (happy to undo these though, let's see what Laurenz thinks).

Enivex:
I'm confused. Why go back to `display-at` from the parameter?

laurmaedje:
I'm on board with the `display(at: ..)` parameter, but for now I'd rather not add it to `get` and deprecate `.at()`. It's not off the table, but I'm not confident in that part and I think we can go one step at a time here.

crates/typst-library/src/model/reference.rs
```diff
@@ -127,10 +127,7 @@ use crate::text::TextElem;
 ///   // Skip all other references.
 ///   if el == none or el.func() != eq { return it }
 ///   // Override equation references.
-///   link(el.location(), numbering(
-///     el.numbering,
-///     ..counter(eq).at(el.location())
-///   ))
+///   link(el.location(), counter(eq).display(at: el.location()))
```
laurmaedje, line 130: For correctness, we should still pass `el.numbering` here or else the automatic behavior could use the wrong one (when it's not configured via set rule)

crates/typst-library/src/introspection/query.rs
```diff
@@ -49,10 +49,7 @@ use crate::introspection::Introspector;
 ///   )
 ///   for chapter in chapters {
 ///     let loc = chapter.location()
-///     let nr = numbering(
-///       loc.page-numbering(),
-///       ..counter(page).at(loc),
-///     )
+///     let nr = counter(page).display(at: loc)
```
laurmaedje, line 52: This will behave slightly differently than before. It'll use the page numbering from the current context instead of `loc.page-numbering()`. Perhaps, we should adjust how `matching_numbering` resolves this and add a test?

tests/suite/introspection/query.typ
```diff
@@ -47,8 +47,7 @@
   let elements = query(selector(figure).after(here()))
   for it in elements [
     Figure
-    #numbering(it.numbering,
-      ..counter(figure).at(it.location())):
+    #counter(figure).display(at: it.location()):
```
laurmaedje, line 50: This should also use `it.numbering`

crates/typst-library/src/introspection/counter.rs
```diff
@@ -356,6 +372,10 @@ impl Counter {
             .or_else(|| self.matching_numbering(context.styles().ok()?))
             .unwrap_or_else(|| NumberingPattern::from_str("1.1").unwrap().into());
 
+        if at.is_custom() {
+            let context = Context::new(Some(location), context.styles().ok());
+            return state.display(engine, context.track(), &numbering);
+        }
         state.display(engine, context, &numbering)
```
laurmaedje: I think if-else would be a bit nicer here

crates/typst-library/src/introspection/counter.rs
```diff
@@ -336,6 +336,17 @@ impl Counter {
         /// `{"1.1"}` if no such style exists.
         #[default]
         numbering: Smart<Numbering>,
+        /// The place at which the counter should be displayed.
+        ///
+        /// If a selector is used, it must match exactly one element in the
+        /// document. The most useful kinds of selectors for this are
+        /// [labels]($label) and [locations]($location).
+        ///
+        /// If this is omitted or set to `{auto}`, displays the counter at the
```
laurmaedje:
```suggestion
        /// If this is omitted or set to `{auto}`, this displays the counter at the
```
(and then rewrap)

tests/ref/counter-display-at.png
laurmaedje, line 1: This image should be deleted

mkorje: woops

crates/typst-library/src/model/reference.rs
mkorje: Ah, good catch.

mkorje: Though I feel like, ideally, `matching_numbering` should somehow try to grab the numbering of the exact element associated with the current counter value? But I see how this would be problematic... Basically #6314

Would also better match the custom behaviour for the page numbering.

laurmaedje: > Though I feel like, ideally, matching_numbering should somehow try to grab the numbering of the exact element 

Honestly, I hadn't considered that option before.

> But I see how this would be problematic... 

How would it be problematic?

tests/suite/introspection/counter.typ
```diff
@@ -110,6 +110,15 @@ B
...
 )
 
+--- counter-page-location-display paged ---
+// Counter display should use location's numbering.
+#set page(numbering: "i")
+Second page <first>
+#context counter(page).display(at: locate(<second>))
+#set page(numbering: "A")
+First page <second>
+#context counter(page).display(at: locate(<first>))
```
laurmaedje: The `locate` calls shouldn't be necessary. I think this isn't tested elsewhere, so might be good to omit them here.

crates/typst-library/src/model/reference.rs
mkorje: Maybe unproblematic was the wrong word, I thought it'd have been a bit ugly but it wasn't so bad in the end I think. I tried it out and added it as the latest commit.

crates/typst-library/src/introspection/counter.rs
```diff
@@ -808,6 +811,53 @@ impl Introspect for CounterFinalIntrospection {
     }
 }
 
+/// Retrieves a counter at a specific location.
+#[derive(Debug, Clone, PartialEq, Hash)]
+struct NumberingAtIntrospection(Element, Location, Span);
```
mkorje: I probably didn't need to have done this, as it should be fine to call `.access` on the introspector myself? It's all internal anyways.

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -304,6 +304,11 @@ impl Introspector {
             .copied()
             .min_by_key(|loc| self.loc_index(loc).wrapping_sub(anchor))
     }
+
+    /// Gets the content at the given location, if any.
+    pub fn content_at(&self, location: Location) -> Option<&Content> {
```
laurmaedje: This is basically `query_unique`, so there's no need to add a new function

crates/typst-library/src/introspection/counter.rs
laurmaedje: Using `.access()` would be wrong as it could then wrongly say that the layout did converge when it actually didn't because the introspection was not recorded.

crates/typst-library/src/introspection/counter.rs
```diff
@@ -778,6 +811,53 @@ impl Introspect for CounterFinalIntrospection {
...
+        _: &mut Engine,
+        introspector: Tracked<Introspector>,
+    ) -> Self::Output {
+        let Some(content) = introspector.content_at(self.1) else {
+            return Ok(None);
+        };
+
+        let numbering = if self.0 == HeadingElem::ELEM {
+            content
+                .to_packed::<HeadingElem>()
```
laurmaedje: I suppose we could directly chain the `to_packed` checks instead of comparing `self.0`. One could hope that Rust can optimize that, but even if it doesn't, it's not exactly on the hot path compared to other places where we do exactly this.

crates/typst-library/src/introspection/counter.rs
```diff
@@ -778,6 +811,53 @@ impl Introspect for CounterFinalIntrospection {
...
+                .and_then(|elem| elem.numbering.as_option().clone())
+        } else {
+            None
+        };
+
+        Ok(numbering)
+    }
+
+    fn diagnose(&self, _: &History<Self::Output>) -> SourceDiagnostic {
+        warning!(self.2, "is a warning even needed here")
```
laurmaedje: yes :D 

it can fail to converge if the content appears and / or disappears depending on the layout iteration.

but you can use the `QueryUniqueIntrospection` with a `Selector::Location` here.

laurmaedje:
Maybe we should move the improvements to matching numbering handling to a separate PR so that this one purely adds the new `at` API?

mkorje:
Sure, I'll do that

EDIT: Ok I should have maybe also extracted the page numbering one...

mkorje:
Ok I removed the page numbering improvement and will add it to the other PR. This should be ready now.

laurmaedje:
Oh, that's why the other PR is draft and contains the commit, too! I thought you'd do it the other way around since the matching does not depend on `display.at`, but the other way around, it does a bit.

mkorje:
That probably makes more sense :sweat_smile:. I'll do it the other way

mkorje:
This one and the other should both be done now.

laurmaedje:
Thanks!

PR #6782, opened 2025-08-18
Add test for default ignorables before a breakpoint

mkorje:
This _seems_ to be okay to do, especially as it is causing incorrect behaviour.

| Before | After |
| - | - |
| old | new |

```typ
#set text(font: "Noto Sans Math")
\u{2295}\u{FE00} vs \u{2295}\u{FE00}
```


laurmaedje:
For reference, this trimming (albeit in more limited form) was originally introduced here: https://github.com/typst/typst/commit/d19a4124de60b043e36e76dfe20fca193deb6a41

https://github.com/typst/typst/blob/e90c30903d2c93a0d899ebd9326c5cd4ab11bf41/crates/typst/src/layout/inline/mod.rs#L1098-L1100

And the change to include all default ignorables here: https://github.com/typst/typst/pull/4497

Enivex:
Context

https://discord.com/channels/1054443721975922748/1176478139757629563/1405675467125559396

mkorje:
> For reference, this trimming (albeit in more limited form) was originally introduced here: [d19a412](https://github.com/typst/typst/commit/d19a4124de60b043e36e76dfe20fca193deb6a41)

I think this could be retained by adding `LineBreak::ZWSpace` to
https://github.com/typst/typst/blob/880e69b0c840cd0d2f0c51e72966ab72c008e713/crates/typst-layout/src/inline/linebreak.rs#L88-L91

> And the change to include all default ignorables here: #4497

What was the reason for trimming all default ignorables?

laurmaedje:
I don't remember. But this code needs to be rewritten regardless because the current trimming is not accessible. We need to write actual space glyphs but ignore their width for layout purposes for multi-line text to copy correctly. This is something I started working on and wanted to continue very soon.

khaledhosny:
Default ignorables should not be removed from text. Being default ignorable does not mean it can be dropped freely, but rather “don’t render a glyph for it unless you really know what you are doing”, but otherwise they pretty much affect text layout .

laurmaedje:
The test that changed here was totally broken anyway. I fixed that in #6805.

Still, the tricky part is that we need to ignore spaces for layout computations even if default ignorables like isolates and stuff follow after them in the line. This happened in the `issue-5490-bidi-invalid-range` (without `-2`) test on my branch. I'll need to investigate why that didn't cause problems here.

The way I'm currently experimenting with is to trim only newlines on the textual level (since those shouldn't be shaped afaik), but otherwise pass the line untrimmed to the shaper and then zero out the advance width of glyphs that belong to a suffix of whitespace and default ignorables after that fact. That way the shaper gets all the information it needs and we emit proper space glyphs for copy-paste , but stuff like justification and right alignment also still behaves correctly (i.e. ignoring the width of the trailing spaces, also if followed by ignorables).

I'm still thinking what to do about explicit newlines w.r.t. to copy paste. If they are trimmed textually, they won't result in a glyph and thus no copy-paste. I checked what LibreOffice does and it seems like the newline is also dropped there. But I think this could in principle be avoided by writing a space glyph with "\n" as ActualText.

khaledhosny:
> we need to ignore spaces for layout computations even if default ignorables like isolates and stuff follow after them in the line.

Shouldn’t the line break happen before the default ignorables (i.e. pushing them into the next line)?

laurmaedje:
> Shouldn’t the line break happen before the default ignorables (i.e. pushing them into the next line)?

Yeah, I think you are right. The problem was something else and that also explains why it does not happen on main, only on the branch where I zero out glyphs advances instead of trimming the string.

The problem was that the shaper assigned `"\u{202a} "` to a space glyph and I was cautiously only advance-zeroing glyphs that were entirely outside of the textual trimming range. Hence, when considering only whitespace for determination of the glyph trimming range, the space glyph was not advance-zeroed. If I change it to instead zero out the advance as soon as the glyph's cluster range is partially outside of the textual trimming range (i.e. not fully contained), all tests pass. I hope this is the right approach, I'll also try to check out the LibreOffice code base.

laurmaedje:
Okay, I found it in LibreOffice:

https://github.com/LibreOffice/core/blob/31dbac45179353a48657d07f2c0da1099608485f/sw/source/core/text/portxt.cxx#L766-L767

https://github.com/LibreOffice/core/blob/31dbac45179353a48657d07f2c0da1099608485f/sw/source/core/text/portxt.cxx#L1107-L1110

However, LibreOffice's text layout logic seems fairly different (with text portions, and hole portions), so this doesn't directly apply. Nonetheless, zeroing out the glyph advances seems quite similar in spirit.

I'm still experimenting with whether it's better to 

- trim trailing whitespace from the line's text and compute a "trimmed offset". Then use this offset to zero out advances of all glyphs whose ranges (reported by the shaper) exceed beyond the trimmed offset. To make this work properly without trimming default ignorables, any kind of range that exceeds outside will lead to a trim, even if there is partial overlap.

- zero out advances of glyphs from the back while their range if indexed into the text contains only whitespace (or whitespace and ignorables). My assumption would be that a glyph that was mapped to whitespace can be safely zero-advanced even if its cluster also contains ignorables (this is the equivalent to the partial overlap check).

I think they are pretty similar conceptually, so it probably comes down to which one is cleaner to implement.

laurmaedje:
I've opened https://github.com/typst/typst/pull/6866. Default ignorables aren't trimmed there anymore, so it supersedes this PR except for the test.

mkorje:
Do you want to merge the test? If so, I'll rebase.

laurmaedje:
Yeah, sure!

laurmaedje:
Thanks!

PR #6794, opened 2025-08-20
Load linked bitmap images in SVG images

### LINKED ISSUE #5335: Support for "linked" images in SVGs with `xlink:href` attribute of `<image>`
### Description

Inkscape supports to "link" images in SVG files by setting the `xlink:href` attribute of `<image>` to a path relative to the SVG file. Currently, Typst only supports base64 encoded embedded images instead.

Linking to images is very useful as it avoids duplication of binary data in a project (the original image file and the embedded data in the SVG) and it also allows the user to quickly switch/replace the image file without having to open the SVG editor again to replace the embedded image.

On Discord, @LaurenzV raised the concern that this "could allow breaking out of the "sandbox" provided by Typst" and added "I guess you could check whether the given path is still within the root and support that".

I don't know much about how SVGs are currently handled but shouldn't it be possible to do something similar to how the paths of the built-in `image` function are handled? I don't see why this makes a difference for the security model of Typst.

### Use Case

This feature would be useful for users wanting to include SVGs containing images created by Inkscape (and maybe other tools?) by avoiding duplication of binary data and avoiding having to manually update the SVG whenever one of the source images changes. In general, images inside of SVGs can be very useful for figures that place some annotations on top of images, use an image as part of a larger technical figure, or, for "collages" of images which can be nicely arranged in Inkscape without having to manually "code" coordinates in Typst.

tovrstra:
This (at least partially) fixes #5335 

This is my first contribution to a Rust project. I'm sorry if things are not up to expectations. In any case, even if the change works for me in a simple test, the current PR is not final. I still have some questions and comments on how to proceed:

### TODO in this PR

- I'll make a corresponding PR to `typst-dev-assets` with small files to test the new feature. Actual tests still need to be written. (For now, I just manually tested.) The following can be tested:
  - [x] Minimal example with inline SVG in typst as `bytes(...)`
  - [x] Example in https://github.com/typst/typst-dev-assets/pull/12 (including testing the `--make-deps` flag)
  - ~An example with an linked image outside the root of the typst project, should fail.~ (not trivially possible at the moment)
  - [x] An example with an absolute path linked image in the SVG, should fail, because these are interpreted different by typst than in the SVG format.
  - [x] An example with a non-existing linked image, should fail.
  - [x] An example with a link to a non-supported file format, should fail.
  - [x] An example with a link to a PDF, should fail.
  - [x] An example with an URL (https://) href, should fail, except for file://
  - ~Add a test with an SVG image linking to a bitmap, both files inside a package~ (postponed to future PR)
- [x] Check formatting of all error messages (see discussion below)
- [x] Improve caching of loading of SVGs (with or without linked images, see discussion below, maybe future PR)
- [x] Reuse file extension matching code from `visualize/image/mod.rs` (see below, requires new function)

### TODO in later PRs

- SVG images linked from SVG images are not supported yet, even if this is possible in usvg. The reason is that it seemed unwise to just copy the private function of `usvg::load_sub_svg`. I think it would make sense for this function to become public in usvg, or not?
- ~~For now, image file formats are just inferred from file extensions, in line with how the `image` function in typst works. This may not be ideal. Image file extensions can be easily wrong, e.g. due to how CDNs work etc. The magic bytes approach from usvg would a be better solution. Also here, it seems wiser to reuse the code than reimplementing it, these functions are not public in usvg either.~~ (Typst also checks magic bytes.)

### Questions answered below

- Should the image resolver also be implemented in the `new` method of `SvgImage`? For now, I've only added it to the `with_fonts` method, whose name is no longer very accurate. `with_fonts_and_images` would be better, but I did not rename it yet.
- Because of the previous point, I left the `base_options()` unchanged. I believe the `new` method still needs it in its original form, not sure though.

@w1th0utnam3 Would you be able to try out this change for your use case?

(I'm quite occupied with other stuff at the moment, but I will certainly follow up to imporve things where needed. There may be some delay in my responses.)


w1th0utnam3:
Thank you for your work! Great to see that somebody is working on this. I can test it for my use case. Currently, I'm getting errors "Failed to load linked images in SVG (Cannot use linked image" because it resolves paths relative to the including .typ file, while my SVG files specify them relative to the SVG file. In my opinion the latter would be the more common use case because typically SVG files are completely independent from the typ files? (my image files are in a different folder than the .typ content files)

tovrstra:
Yes, I agree, it should be relative to the SVG. It seems that the span object I'm using is not referring to the SVG image, but rather to the typ file. 

Solving this is a bit more tricky than I first guessed, but it should be possible. The catch is that the `image()` function also accepts a `bytes` as argument, in which case the relative path to the linked image does not exist and some error should be raised when there are linked images.

w1th0utnam3:
Hm, I mean in case of the bytes it could also be something generated inside of the Typst document where it could make sense to fall back to the path of the current Typst file or the root of the Typst project. Might be worth to discuss this. But I don't have a strong opinion on this.

tovrstra:
@w1th0utnam3 Makes a lot of sense. I've changed the relative path logic and it should work better now. If the SVG image has a bytes source, the parent of the typ file is used as reference (also if it is inside a package). If the typ file cannot be resolved, it falls back to the parent of the main typ file. (At least, that's the idea.)

w1th0utnam3:
Great, now it works for my use case. Maybe somebody from the core team can have a look and comment on the PR, I'll ask in the Discord.

laurmaedje:
Thanks for opening a PR! I think adding this makes sense generally.

Behaviour-wise, I'd say the images should be loaded relative to the SVG (like an import would be relative to a Typst file) and in SVGs decoded from bytes it should be relative to the file that called `image` (like an import in `eval` would). I think this is already what you implemented?

> Should the image resolver also be implemented in the new method of SvgImage?

No, that's only used for OpenType SVG fonts.

> The reason is that it seemed unwise to just copy the private function of usvg::load_sub_svg. I think it would make sense for this function to become public in usvg, or not?

Probably yeah. Maybe it could even just be a higher-level function. Not sure. Worth discussing this upstream.

> For now, image file formats are just inferred from file extensions, in line with how the image function in typst works. This may not be ideal.

We should for sure reuse existing logic, either that from resvg or [the one from Typst itself](https://github.com/typst/typst/blob/77ba73eb1e942e5379ef6392ffba921e1a32baf5/crates/typst-library/src/visualize/image/mod.rs#L322).

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -39,28 +42,48 @@ impl SvgImage {
     #[comemo::memoize]
     #[typst_macros::time(name = "load svg")]
     pub fn with_fonts(
+        svg_parent_file: FileId,
```
laurmaedje: This is not ideal as it prevents an image from being cached if it's used across files even if it does not use this at all. 

It can be _probably_ be prevented by creating a new tracked type that wraps `World` and handles this logic internally. (Extra points for also handling `families` in there.) But it's definitely tricky and could also be done in a followup PR or maybe I'd just push this over the PR once it's otherwise ready.

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -265,14 +266,29 @@ impl Packed<ImageElem> {
...
-                    loaded.data.clone(),
-                    engine.world,
-                    &families(styles).map(|f| f.as_str()).collect::<Vec<_>>(),
+            ImageFormat::Vector(VectorFormat::Svg) => {
+                // We need to decide from which file images are linked,
+                // to resolve relative href image paths in the SVG.
+                // Try to locate the typst source file, or fall back to the main file.
+                let svg_parent_file = match span.id() {
+                    Some(file) => file,
+                    None => engine.world.main(),
```
laurmaedje: This fallback shouldn't happen here. Overall, it would be better if this code deferred to the existing logic here:

https://github.com/typst/typst/blob/77ba73eb1e942e5379ef6392ffba921e1a32baf5/crates/typst-syntax/src/span.rs#L169-L174

(Note that if just passing the span through that makes the caching problem even worse, so we'd need to handle that properly then. But as mentioned above, I can also do that myself.)

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -287,3 +310,56 @@ impl FontResolver<'_> {
...
+                            );
+                        };
+                        None
+                    }
+                }
+            }
+            Err(e) => {
+                if self.error.is_empty() {
+                    self.error =
+                        eco_format!("Cannot use linked image: {}; Cause: {:?}", href, e);
```
laurmaedje: the error messages should follow the standard style (lowercase and how the cause is integrated). if you search across the codebase for "failed to", you'll find plenty of examples.

tovrstra:
Thanks for all the feedback! I'll update the todo list on top. I'm going to try fixing one thing at a time.

> Behaviour-wise, I'd say the images should be loaded relative to the SVG (like an import would be relative to a Typst file) and in SVGs decoded from bytes it should be relative to the file that called `image` (like an import in `eval` would). I think this is already what you implemented?

This is indeed how it works now.

I've tried improving the error messages and using span.resolve_path in the latest commit. It feels a bit like I'm abusing instead of properly reusing existing logic. A typical error message now looks as follows:

```text
error: failed to load linked image ../small.jpg (failed to load file (access denied))
  ┌─ link-jpg.svg:1:0
  │  
1 │ ╭ <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 50 20" style="background:#fff" width="50">
2 │ │ <circle r="11" cx="10" cy="10" fill="red" />
3 │ │ <image width="30" height="20" href="../small.jpg" x="20" y="0" />
4 │ │ <circle r="20" cx="10" cy="10" fill="none" stroke="red" />
  · │
7 │ │ </svg>
8 │ │ 
  │ ╰^
```

I'll comment on another line where it's not clear how to format the error message....

tovrstra:
For some reason I get the following warning, but that seems wrong:

```text
warning: unused import: `crate::World`
  --> crates/typst-library/src/visualize/image/mod.rs:23:5
   |
23 | use crate::World;
   |     ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

crates/typst-library/src/visualize/image/svg.rs
tovrstra: This should be fixed.

crates/typst-library/src/visualize/image/mod.rs
tovrstra: I tried fixing this, but I'm not sure if the new code is better.

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -287,3 +312,82 @@ impl FontResolver<'_> {
...
+        if !self.error_msg.is_empty() {
+            return None;
+        }
+
+        // Resolve the path to the linked image.
+        let href_path = VirtualPath::new(self.svg_path.as_str()).join(href);
+        let href_full = href_path.as_rooted_path().to_str().unwrap();
+        let href_file_result = self.span.resolve_path(href_full).at(*self.span);
+        if let Err(err) = href_file_result {
+            self.error_msg = eco_format!("TODO {:?}", err);
```
tovrstra: Is there a better way to format this error? I'm not sure how to trigger this code path either?

tovrstra:
I've improved the error messages further, e.g. 

```text
error: failed to load linked image ../small.jpg in SVG (access denied)
  ┌─ link-jpg.svg:1:0
  │  
1 │ ╭ <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 50 20" style="background:#fff" width="50">
2 │ │ <circle r="11" cx="10" cy="10" fill="red" />
3 │ │ <image width="30" height="20" href="../small.jpg" x="20" y="0" />
4 │ │ <circle r="20" cx="10" cy="10" fill="none" stroke="red" />
  · │
7 │ │ </svg>
8 │ │ 
  │ ╰^
```

crates/typst-library/src/visualize/image/mod.rs
tovrstra: Should be ok now.

tovrstra:
@laurmaedje I've made several changes and tried to address most of your comments. (Caching is still on the list.) There are some points where I'm a bit in doubt. If you can comment, I may be able to fix a few more things: 

- `span.resolve_path(...)` is now called in `mod.rs` before `SvgImage::with_fonts`. I tried putting the `resolve_path()` call inside the new `ImageResolver`. However, as you mentioned, this requires passing in the `span` to the `ImageResolver`, which terribly breaks caching. I took it back out. I hope this is OK?

- With the last commit below, I believe caching works (except for the font families, I did not touch this part). It now works as follows: when a second SVG is loaded with the same contents and the same parent directory as a previously loaded SVG, it will be a cache hit. Either of these two change, and it will be a miss. This seems fairly optimal. I had to use a workaround for the behavior of `FileId::join` to get to this point. The only possibility to make caching more efficient is to make a distinction between SVGs with linked images (for which the parent directory matters) and SVGs without (for which only the image contents itself matters.) 

- We may also consider another form of caching, namely when the same image is linked multiple times from one or more SVG files. I wonder if it would matter much? I don't believe this form of caching in typst will stop usvg from processing the image data twice.

- I've added a couple of test cases, but could not implement some of the tests I had in mind. For example, adding the following test does not produce an access denied error as hoped:
  
  ````typst
  --- image-svg-linked-access-denied ---
  #image(bytes(
    ```
    <svg xmlns="http://www.w3.org/2000/svg">
      <image href="../../../../../up.jpg" />
    ```.text
  ))
  ````
  
  This gives me:
  
  ```text
  Error: 8-7:2 failed to load linked image ../../../../up.jpg in SVG (file not found, searched at up.jpg)
  ```
  
  When I create a similar test manually, I do get the right "access denied" error.

- I've manually tested that an SVG image in a package can link to a bitmap inside the package, which is nice. I'm not sure how to include this scenario in the tests?

tovrstra:
@w1th0utnam3 There have been several changes. Would you be able to test the latest update?

w1th0utnam3:
Yes, with commit `597ebeee` my SVGs show up correctly.

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -42,26 +48,44 @@ impl SvgImage {
...
-        let font_hash = resolver.into_inner().unwrap().finish();
-        Ok(Self(Arc::new(Repr { data, size: tree_size(&tree), font_hash, tree })))
+        let font_hash = font_resolver.into_inner().unwrap().finish();
+        let image_resolve_error = image_resolver.lock().unwrap().error.clone();
+        match image_resolve_error {
+            Some(err) => Err(err),
+            None => {
+                Ok(Self(Arc::new(Repr { data, size: tree_size(&tree), font_hash, tree })))
+            }
+        }
```
laurmaedje: I think an early return for the error case (with `if let Some(err)`) would be clearer here control-flow-wise.

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -265,14 +265,29 @@ impl Packed<ImageElem> {
...
+                let svg_file = match self.source.source {
+                    DataSource::Path(ref path) => span.resolve_path(path).ok(),
+                    DataSource::Bytes(_) => span.id(),
+                };
+                // The following is a workaround to for the fact that FileID::join
+                // takes the parent directory before joining.
+                // The result is a FileId object that can be used to resolve hrefs,
+                // yet it only contains relevant information about the directory,
+                // not the filename, to avoid incorrect cache misses.
+                let svg_parent = svg_file.map(|file| file.join("_"));
```
laurmaedje: I think I had a mistake in my thinking when reviewing this originally. I thought that using an SVG from different files would be a cache miss, but as the `FileId` is already resolved to the SVG's path, that's not the case unless it's a byte SVG. So I think this was actually probably good enough and I'd rather skip this hack (which only has an effect in the bytes case). Sorry for the confusion!

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -287,3 +311,124 @@ impl FontResolver<'_> {
...
+                        ReportPos::None,
+                        eco_format!("failed to load linked image {} in SVG", href),
+                        err,
+                    ));
+                    None
+                }
+            }
+        } else {
+            None
+        }
```
laurmaedje: I think this would also be clearer with an early return for the outer if.

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -287,3 +311,124 @@ impl FontResolver<'_> {
...
+    }
+
+    /// Load a linked image or return an error message string.
+    fn load_or_error(&mut self, href: &str) -> Result<usvg::ImageKind, EcoString> {
+        // If the href starts with "file://", strip this prefix to construct an ordinary path.
+        let href = if let Some(stripped) = href.strip_prefix("file://") {
```
laurmaedje: this could use `.strip_prefix(..).unwrap_or(href)`

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -287,3 +311,124 @@ impl FontResolver<'_> {
...
+            Err(err) => Err(match err {
+                FileError::NotFound(path) => {
+                    eco_format!("file not found, searched at {}", path.display())
+                }
+                FileError::AccessDenied => EcoString::from("access denied"),
+                FileError::IsDirectory => EcoString::from("is a directory"),
+                FileError::Other(Some(msg)) => msg,
+                FileError::Other(None) => EcoString::from("unspecified error"),
+                _ => EcoString::from("unexpected error"),
+            }),
```
laurmaedje: It's a bit unfortunate that this is duplicated here. There's probably some way to make that nicer, but I guess we can also defer that to later.

crates/typst-library/src/visualize/image/svg.rs
```diff
@@ -287,3 +311,124 @@ impl FontResolver<'_> {
...
+        // Load image if file can be accessed.
+        match self.world.file(href_file) {
+            Ok(bytes) => {
+                let arc_data = Arc::new(bytes.to_vec());
+                let format = match determine_format_from_path(href) {
+                    Some(format) => Some(format),
+                    None => ImageFormat::detect(&arc_data),
+                };
+                match format {
+                    None => Err(EcoString::from("could not determine image format")),
```
laurmaedje: These `EcoString::from` can just be `.into()` calls. It would also be nice to have the happy path at the top of the match.

laurmaedje:
```suggestion
                    None => Err(EcoString::from("unknown image format")),
```
for consistency with the other code

laurmaedje (commented):
> I've added a couple of test cases, but could not implement some of the tests I had in mind. For example, adding the following test does not produce an access denied error as hoped. [...]

Yeah, that is a quirk of the test runner. We don't want the error output messages to contain any system-dependant paths, so the rest root is `""`, but then this code here doesn't ever trigger the access denied code path:

https://github.com/typst/typst/blob/3d64f7b659905755322c419c5cd1fdb2078590e8/crates/typst-syntax/src/path.rs#L60-L78

Maybe that code should be considered buggy because it doesn't handle the case that `.pop()` does nothing... I'm not sure.

> I've manually tested that an SVG image in a package can link to a bitmap inside the package, which is nice. I'm not sure how to include this scenario in the tests?

The tests have a package mechanism with packages stored at `tests/packages/`. It wouldn't be needed from my side, but if you want, you can add a test there.

laurmaedje:
> Maybe that code should be considered buggy because it doesn't handle the case that .pop() does nothing... I'm not sure.

It's possible to observe the same thing in the CLI when setting the root to `/`:

```sh
typst c hi.typ --root /
```
```typ
#image("/../../Users/laurenz/Desktop/cylinder.svg", width: 40%)
```

You can add arbitrary `..` when you're already at the physical FS root. But `cat` and `ls` behave the same, so I guess that's just how it works.

tovrstra:
Thanks again for the second round of feedback, very much appreciated!

I believe it is better to omit the additional tests I had in mind, at least for now. We may come back to the image-in-package test when linking to SVG images is supported (future PR). With that feature, it would become possible to make a tiny test package with just SVG text files that only take a handful of bytes. I'm a bit reluctant to add such an example now with a bitmap image in the main typst repo.

crates/typst-library/src/visualize/image/mod.rs
tovrstra: I'm happy to take the `.join("_")` back out and rename variables accordingly. Just to be sure, I believe this would break caching of SVG images in two cases:

1. When two SVG files in the same directory have exactly the same content. (This seems rare, but could occur when the SVGs are produced by an external tool, e.g. bar codes or matrix codes generated by an external script.)
2. When two TYP files in the same directory use the same `bytes` for an SVG image.

To avoid the current workaround, one could consider changing the `FileId::join` method to not call `parent` internally, and adding separate methods `parent` and `sibling` (to replace `join`) instead to cover all use cases. That would imply changes throughout, which may not be desirable.

crates/typst-library/src/visualize/image/svg.rs
tovrstra: P.S. I've also changed the code to replace `clone()` by `into_inner()`. The latter seemed more appropriate.

crates/typst-library/src/visualize/image/svg.rs
tovrstra: In principle, the code can be simplified to something like

```rust
Err(err) => Err(err.to_string().into())
```

The downside is that the error messages become more convoluted. E.g.:

With one-liner, I get:

```text
failed to load linked image ../small.jpg in SVG (failed to load file (access denied))
```

With the match expression, this is made more compact:

```text
failed to load linked image ../small.jpg in SVG (access denied)
```

I considered transforming the output of `Err(err.to_string().into())` into a more condensed string, but that also requires a match expression to handle different cases, which seemed worse.

crates/typst-library/src/visualize/image/mod.rs
laurmaedje: Exactly. I think breaking caching in 1+2 is okay. What I wanted to avoid is breaking caching in a scenario where the same SVG file is added in many places across a document, but that was already not a problem.

crates/typst-library/src/visualize/image/svg.rs
laurmaedje: Yeah, I saw the duplication that omitting this would incur. I think fixing this would require some changes to how FileErrors are displayed. We can leave that as future work and leave this as-is for now.

laurmaedje: Maybe add a
```
// TODO: Somehow unify this with `impl Display for FileError`.
```

tovrstra: Good point. Done

crates/typst-library/src/visualize/image/mod.rs
tovrstra: Good, is fixed.

tovrstra:
Unless I missed something, this PR should be ready. In any case, if something else pops up, no problem to address future comments.

laurmaedje:
Thank you!

Issue #1267, opened 2023-05-20
Copy-paste for hyphenated text copies hyphens


ghost:
### Description

**Problem**
Currently when hyphenation is applied to a word, the word copies as `prefix-
suffix`. Some PDF readers will copy this as `prefixsuffix` because they are sure enough that the minus character was a hyphenation character.

Using this, we can't really know if this hyphenation was done by an algorithm or written like this by the author, because the minus sign is used as hyphenation character.

**Solution**
Use the [hyphen character](https://en.wikipedia.org/wiki/Hyphen) „‐“ (U+2010) if and only if the hyphen character came from the hyphenation algorithm.
If a line break appeared within a word that already had some hyphenation character (mostly the minus sign), the hyphen character should not be used. Instead, the character should be used that was already in the input text.
When the text had a soft hyphen, the hyphen character should also be used.

Maybe this solution can only be applied if the glyph is available in the font. If so and the glyph is available in the font, it should be used, otherwise, stay with the minus sign if the font does not have the hyphenation character.

### Use Case

It will be possible to copy and recreate the original input text much better and will give PDF readers better knowledge about the text to copy. It can also improve the text searching in the PDF.

reknih:
The PDF 1.7 spec has the following to say on this in clause 14.8.2.2.3:

> Hyphenation. Among the artifacts introduced by text layout is the hyphen marking the incidental division of a word at the end of a line. In Tagged PDF, such an incidental word division shall be represented by a soft hyphen character, which the Unicode mapping algorithm (see "Unicode Mapping in Tagged PDF" in 14.8.2.4, "Extraction of Character Properties") translates to the Unicode value U+00AD. (This character is distinct from an ordinary hard hyphen, whose Unicode value is U+002D.) The producer of a Tagged PDF document shall distinguish explicitly between soft and hard hyphens so that the consumer does not have to guess which type a given character represents.

reknih:
Hey, so they want us to draw a glyph (so either the glyph for `U+2010` or, if not present in the font, `U+002D`) and map it to two different code points `U+00AD` for hyphens that Typst inserted and `U+002D` for user-defined hyphens. They want us to rewrite the font if you will. Does that clarify it?

On a different note: Do you know whether it's common for fonts to include a distinct glyph for `U+2010`?

laurmaedje:
Do we have to handle this with the `/ToUnicode` mapping (and as such really with two separate glyphs) or we could we maybe also display the same glyph with different `/ActualText`?

[...]

Issue #6742, opened 2025-08-10
Hyphen copies to unclear bytes


TS53:
### Description

Up to version 0.12.0 hyphen chars copied to a minus hyphen (char 45), since 0.13.0 it copies to some unclear byte sequence.

The following code creates a text with hyphens:
```typst
#set page(width: 2cm)
#set text(lang: "la", hyphenate: true)
#set par(justify: true)

#lorem(10)
```

Google Chrome and Adobe Reader could copy the text without hyphens up to 0.12.0, now Chrome copies to some ESC char and Adobe Reader to some other unclear char.

Partly [related](https://github.com/typst/typst/issues/1267).

### Operating system

Windows

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
I think the unclear byte mapping happens when the ToUnicode mapping contains an empty string for the glyph. This can't happen anymore after #6799.

PR #6799, opened 2025-08-21
Fix Unicode mapping of hyphenation artifacts

laurmaedje:
This fixes how hyphenation artifacts are mapped to Unicode, following up on the refactorings in #6798. Originally, they were mapped to hard hyphens. Since the switch to krilla they were mapped to empty text via `/ActualText`. Now they'll be mapped to U+AD SHY HYPHEN, in line with the PDF 1.7 spec Section 14.8.2.2.3.

PDF provides two ways for text to be mapped to Unicode text: The `/ToUnicode` glyph to text mapping and the `/ActualText` mechanism. Since the former only supports one mapping per glyph, it's not sufficient to encode the hyphen glyph to both a hard hyphen and soft hyphen depending on the circumstance. For this reason, krilla will write `/ActualText` when both appear but were mapped to different text by Typst.

Viewer support for `/ActualText` is kinda patchy, so some viewers will just ignore it, but what we're doing now is at the very least spec-compliant. It would be possible to duplicate the glyph on the font level during subsetting to make it work in even more viewers, but we decided that this is out of scope for now.

Fixes https://github.com/typst/typst/issues/1267

### DIFF
--- a/crates/typst-layout/src/inline/line.rs
+++ b/crates/typst-layout/src/inline/line.rs
@@ -178,7 +178,8 @@ pub fn line<'a>(
         && pred.dash == Some(Dash::Hard)
         && let Some(base) = pred.items.last_text()
         && should_repeat_hyphen(base.lang, full)
-        && let Some(hyphen) = ShapedText::hyphen(engine, p.config.fallback, base, trim)
+        && let Some(hyphen) =
+            ShapedText::hyphen(engine, p.config.fallback, base, trim, false)
     {
         items.push(Item::Text(hyphen), START_HYPHEN_IDX);
     }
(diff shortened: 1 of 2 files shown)


ghost:
I can confirm that it works as expected.

Here is the behaviour with the PDF readers:
- Adobe Reader, Foxit Reader, Google Chrome: full support
- Edge, Firefox, SumatraPDF: copies both parts with linebreaks, full text not searchable as one word.

I'm wondering if this also fixes https://github.com/typst/typst/issues/6742.
Probably not internally, but none of the readers copies the invalid utf8 anymore after this commit.
It seems all of them support `/ActualText`.

PR #6825, opened 2025-08-25
Add hint to LrElem to mark implicit pairs

tjbcg:
This PR changes the deparenthesization logic of fractions:
- Previously, this would be done by two internal `num/denom_deparenthesized` boolean internal fields set at eval time using data from the AST/CST.
- With this PR: A synthesized field on LrElem indicates whether each pair was automatically built at parsing time or wheter it was explicitly built from the constructor (`explicit: bool`). Deparenthesizing is then a process that happens during layout by checking for an implicit l/r `()` outer pair on numerator and denominator.

The point of this change is to allow some deparenthesization logic to happen in user space. Previously, one could not, for example, change the horizontal fraction slash character without dropping parentheses:
```typst
#show math.frac.where(style: "horizontal"): it => it.num + "|" + it.denom

$ (a+b) / c $
// Previously this would render as : a+b|c
// Now: (a+b)|c
```
If deparenthesization is desired from extracted num/denom content then custom logic is possible in userspace (previously this would always be deparenthesized with no way of reparenthesizing).

### DIFF
--- a/crates/typst-eval/src/math.rs
+++ b/crates/typst-eval/src/math.rs
@@ -68,7 +68,7 @@ impl Eval for ast::MathDelimited<'_> {
         let open = self.open().eval_display(vm)?;
         let body = self.body().eval(vm)?;
         let close = self.close().eval_display(vm)?;
-        Ok(LrElem::new(open + body + close).pack())
+        Ok(LrElem::new(open + body + close).with_explicit(false).pack())
     }
 }
 
--- a/crates/typst-eval/src/math.rs
+++ b/crates/typst-eval/src/math.rs
@@ -128,15 +128,7 @@ impl Eval for ast::MathFrac<'_> {
         let denom_expr = self.denom();
         let denom = denom_expr.eval_display(vm)?;
 
-        let num_depar =
-            matches!(num_expr, ast::Expr::Math(math) if math.was_deparenthesized());
-        let denom_depar =
-            matches!(denom_expr, ast::Expr::Math(math) if math.was_deparenthesized());
-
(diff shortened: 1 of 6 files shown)


laurmaedje:
In principle, handling this in LR seemed/seems cleaner to me than in Frac (with duplication for numerator and denominator), but at the same time, the code takes apart the `SequenceElem`, expecting a very specific structure, feels a bit hacky. This kind of taking apart of content is something that rarely happens in the compiler. The other way around, adding parentheses which have been removed by the parser seems more predictable to me. I'm not sure whether that's feasible or whether that would result in otherwise strange code. Overall, I'm not sure how to feel about this.

tjbcg:
Good point. I'll look into it further, I may have some idea on how to redesign this in a more elegant way. I'll keep this draft open for a bit if you don't mind?

tjbcg:
@laurmaedje What do you think about the new approach ?

tjbcg:
Note that technically the field could be marked internal and thus expose nothing new to userspace, but I figured it'd be unintuitive to have `it.func() == math.lr` but also `it.body` is not getting rendered as a typical `lr` would.

laurmaedje:
> What do you think about the new approach?

The changes don't look quite trivial and unfortunately I don't have time to look to deeply into this at the moment, so it'll have to wait until after 0.14.

laurmaedje:
I think overall we should avoid adding additional synthesized fields whenever possible (because they are a hack and shouldn't really exist). So if this PR is purely about improving the code quality and doesn't have any functional changes, then I'd rather not make the change.

PR #6107, opened 2025-03-27
fix(ci): include rust-src in flake.nix

michaelfortunato:
rust-src was not available in the flake.nix, breaking rust-analyzer when using nix develop. 
I tried to capture the before and after here, but the red squiggly lines did not show up, you will notice that I cannot do lsp actions before and then can after. 
https://asciinema.org/a/XzG8hoWcLV5NLPSkIgKGSo80i
Fixes #6105

PR #6826, opened 2025-08-26
Use rust-analyzer from fenix toolchain in flake

mkorje:
My dev environment has started throwing `Failed spawning proc-macro server for workspace ...: Failed to run proc-macro server from path ...` errors with rust-analyzer.

I believe the cause is an out of date rust-analyzer package: https://github.com/typst/typst/pull/6618 added `rust-analyzer` from the `nixpkgs` flake input which hasn't been updated in almost a year, so `/nix/store/0fdrbfqza9p1vwzccy1hb1pdbx55hags-rust-analyzer-2024-09-02/bin/rust-analyzer` is being used.

Instead of just updating `nixpkgs`, I figured its probably best to use rust-analyzer from the rust-toolchain fenix provides. This means also adding rust-src (cf. https://github.com/typst/typst/pull/6107).


flake.nix
```diff
@@ -127,10 +127,13 @@
           checks = self'.checks;
           inputsFrom = [ typst ];
 
-          buildInputs = with pkgs; [
+          buildInputs = with rust-toolchain; [
             rust-analyzer
+            rust-src
           ];
```
PgBiel: Since `with pkgs` is the usual here, let's prefer to be explicit.

```suggestion
          buildInputs = [
            rust-toolchain.rust-analyzer
            rust-toolchain.rust-src
          ];
```

PgBiel (commented):
Right, this was an original concern of mine in the PR. Thanks for fixing it.

PgBiel:
This seems fine in principle, can you rebase?

PgBiel:
@tingerrr Can you confirm this works for you?

tingerrr:
I'll try this out in about 2 hours.

I would probably done it like this from the get go if the docs for the various nix  things weren't so sparse.

PgBiel:
No problem, take your time.

tingerrr:
Seems to work just fine, hover actions, code navigation and code actions work as expected. Running `cargo check` doesn't require rechecking so it seems it uses the right toolchain too.

LGTM.

PR #6827, opened 2025-08-26
Update nix flake and use nixfmt

mkorje:
Switches to use the [official formatter](https://github.com/NixOS/nixfmt?tab=readme-ov-file#nix-fmt-experimental) and formats the `flake.nix` file with it.

I had originally only updated the `nixpkgs` input (for the `nixfmt` package), but then I got a hash mismatch in `crane` with `nix build`. So I just decided to update all the inputs instead :shrug: (there's no harm in it I guess and it has been a while since they were last updated).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,130 +15,157 @@
     };
   };
 
-  outputs = inputs@{ flake-parts, crane, nixpkgs, fenix, rust-manifest, self, ... }: flake-parts.lib.mkFlake { inherit inputs; } {
-    systems = import inputs.systems;
+  outputs =
+    inputs@{
+      flake-parts,
+      crane,
+      nixpkgs,
+      fenix,
+      rust-manifest,
+      self,
+      ...
+    }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -15,130 +15,156 @@
...
+          # functions.
+          commonCraneArgs = {
+            inherit src pname version;
+
+            buildInputs = [
+              pkgs.openssl
+            ]
+            ++ (lib.optionals pkgs.stdenv.isDarwin [
+              pkgs.libiconv
+            ]);
```
PgBiel: Based on https://github.com/NixOS/nixpkgs/pull/354097/files#diff-9a6b612d1f795112a2502df31b823fcc2b6e396a68b41d16246e247da532549a, I think we can now remove this entirely. (Please test)

```suggestion
            ];
```

mkorje: I don't have a mac to test, but making this change isn't throwing any errors at least.

mmlb: I've been able to build this on a m4 mac.

PgBiel (approved):
Builds here, thanks!

PR #6831, opened 2025-08-28
Handle italic_angle in various renderers

tbodt:
Tested with Univers Oblique (a non-free font, so not sure how to commit tests for this. create or modify a font specifically for this testing? it's surprisingly hard to find fonts that break on this.)

Also requires this fix in pixglyph: https://github.com/typst/pixglyph/pull/3

`glyph_frame` also needs this in theory, but Univers certainly doesn't have any glyphs that would trigger the fallback so I wasn't able to get a motivating test case.


laurmaedje:
Given the conclusion in https://github.com/typst/pixglyph/pull/3, I'll close this.

PR #6837, opened 2025-08-29
Add `Clone` impl for typst-kit `Fonts`

messense:
Make it possible to clone `Fonts` so it's easier to reuse already searched fonts.

I'm trying to use this for https://github.com/messense/typst-py/issues/117


laurmaedje:
Would wrapping the `Fonts` in an `Arc` instead work in your use case? That would be basically free and also give you reuse of already loaded fonts which you won't get with `Clone`.

messense:
Thanks, `Arc` seems to work for that use case, though we still need to clone `FontBook` for `LazyHash<FontBook>`.

laurmaedje:
Turning the `book` in `Fonts` into an already `LazyHash`ed version would probably make sense.

PR #6838, opened 2025-08-29
Fix slicing last n elements using count

### LINKED ISSUE #6630: Slicing last n elements using `count` returns empty array
### Description

When slicing an array with a negative `start` and a `count` such that the last `count` elements are sliced, the returned array is empty. 

Example: 
```typ
#{
  let s1 = (1,2,3,4).slice(-3, count: 2) // (2, 3)
  let s2 = (1,2,3,4).slice(-2, count: 2) // ()
}
```

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

eltos:
Fixes #6630 for `array`, `str` and `bytes` by moving `end = end.or(start + count)` behind `start = locate(start)`, to ensure `start` is positive when adding `count`.

Tests added.

### DIFF
--- a/crates/typst-library/src/foundations/array.rs
+++ b/crates/typst-library/src/foundations/array.rs
@@ -286,11 +286,8 @@ impl Array {
         #[named]
         count: Option<i64>,
     ) -> StrResult<Array> {
-        let mut end = end;
-        if end.is_none() {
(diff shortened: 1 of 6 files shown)


laurmaedje:
Thanks!

PR #6839, opened 2025-08-30
Fix auto hanging-indent for centered headings

### LINKED ISSUE #6834: Center-alignment of numbered headings is off for multiline headings
### Description

Headings can be center-aligned with this simple show rule
```typ
#show heading: set align(center)
```

However, for multi-line headings with numbering enabled, only the first line is truly centred and the following lines are centre-aligned with respect to the first line **excluding** the numbering. This makes sense for left-aligned headings (1st case in the example below), but for centre-aligned headings it looks off (2nd case). Instead, when numbering is disabled and the number is manually added for demonstration, one gets the expected result (3rd case).

Image

```typ
#block(stroke: gray)[
  
  #set heading(numbering: "A)")
  = Left aligned\ multiline heading\ OK
  #lorem(5)
  
  #show heading: set align(center)
  = Centered\ multiline heading\ OFF
  #lorem(5)
  
  #set heading(numbering: none)
  = B) Centered\ multiline heading\ OK
  This is desired\
  #lorem(5)

]
```

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

eltos:
This fixes #6834 by using the numbering width as automatic hanging-indent only and only if the heading is start-aligned. For centered and end-aligned headings auto indent now defaults to zero.

Test added and documentation updated.


laurmaedje:
Thanks for this, too! Well-made PRs that are easy to merge are always great. :)

PR #6840, opened 2025-08-31
Skip labels with invalid characters in reference completion

bluss:
When completing @, skip labels that have names that we know are incompatible with the @ syntax. For example, `label("t/1")` creates a label that is impossible to use with the @t/1 syntax, so it should be skipped, for both `@` and `<` completion.

This is submitted for discussion.

- `@test,1 @test/2 @test@2` are not valid syntax (or if they are, they reference `@test`)
- `<test,1> <test/2> <test@2>` are not valid syntax

Reference completion is autocompletion after e.g `@t`, label completion after e.g `<t`.

- When using certain packages they use lots of internal label names as part of their implementation. These label names fill the label list and makes it harder for the user to find the labels and references they want to use.
- The convention for what to ignore here will influence names libraries use, they will/should use names that are hidden if their label names should not be completed or visible to the user.

New internal dependency

- typst-ide depends on typst-syntax.

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -24,9 +24,6 @@ use crate::utils::{
 };
 use crate::{IdeWorld, analyze_expr, analyze_import, analyze_labels, named_items};
 
-/// Labels that contain these chars are not suitable for reference completion.
-const LABEL_SUPPRESSED_CHARS: &[char; 3] = &[',', '/', '@'];
-
 /// Autocomplete a cursor position in a source file.
 ///
 /// Returns the position from which the completions apply and a list of
@@ -1300,7 +1297,7 @@ impl<'a> CompletionContext<'a> {
 
         for (label, detail) in labels.into_iter().skip(skip).take(take) {
             let name = label.resolve();
-            if !include_suppressed && name.as_str().contains(LABEL_SUPPRESSED_CHARS) {
+            if !include_suppressed && !typst_syntax::is_valid_label_literal_id(&name) {
                 continue;
             }
             self.completions.push(Completion {
@@ -1313,7 +1310,7 @@ impl<'a> CompletionContext<'a> {
                         if close { ">" } else { "" }
                     )
                 }),
-                label: label.resolve().as_str().into(),
+                label: name.as_str().into(),
                 detail,
             });
         }
@@ -1762,13 +1759,13 @@ mod tests {
 
     #[test]
     fn test_autocomplete_ref_and_label_invalid_labels_skipped() {
-        let test = r#"a<test> b#label("test,1") c#label("test@1") d#label("test/1")"#;
+        let test = r#"a1<test> a2<test.:1> b#label("test,1") c#label("test@1") d#label("test/1")"#;
         let exclude_list = ["test,1", "test@1", "test/1"];
         test_with_addition(test, " #ref(<)", -2)
-            .must_include(["test"])
+            .must_include(["test", "test.:1"])
             .must_exclude(exclude_list);
         test_with_addition(test, " @", -1)
-            .must_include(["test"])
+            .must_include(["test", "test.:1"])
             .must_exclude(exclude_list);
     }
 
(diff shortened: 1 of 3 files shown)


crates/typst-ide/src/complete.rs
```diff
@@ -1453,7 +1460,7 @@ impl<'a> CompletionContext<'a> {
                     );
                     self.scope_completions(false, |value| value.ty() == *ty);
                 } else if *ty == Type::of::<Label>() {
-                    self.label_completions()
+                    self.label_completions(true)
```
bluss, line 1460: help me with this one - how is it invoked, and should it have include all or suppress semantics?

laurmaedje:
There is this function, which could be used:
https://github.com/typst/typst/blob/3d64f7b659905755322c419c5cd1fdb2078590e8/crates/typst-syntax/src/lexer.rs#L1084-L1085

And then, there's existing logic in `repr` to show `label("..")` instead of `<..>` if it's not valid syntax:
https://github.com/typst/typst/blob/3d64f7b659905755322c419c5cd1fdb2078590e8/crates/typst-library/src/foundations/label.rs#L92-L101

I could imagine doing something similar here and filtering out not based on whether it's a valid literal, but based on another simpler convention, e.g. supressing all labels with a leading underscore (I think this is quite common already). While I'm not a huge fan of such ad-hoc Python-like "privacy", it might be better than nothing. Down the road, I want to introduce contextually generated anonymous labels (or probably those would be of type `location`, but still) which would make such uses of labels rarer, but it would not eliminate all of them. So we'd still need to deal with this somehow.

On a final note, it might be nice to expose this as a separate API next to `analyze_labels` because this API is not only used in autocompletion. It also powers the reference button in the web app, which suffers from the same problem.

crates/typst-ide/src/complete.rs
laurmaedje: This is what appears if you have your cursor `|` in `#ref(|)`, `#query(|)`, ... Basically any function that takes a label. For `ref`, we'd want to suppress, but for `query` not. So the answer is probably to be on the safe side and not suppress.

laurmaedje:
Thinking about this further: What if we don't make a syntactical distinction at all (still fixing the bug that it gives invalid labels of course) but rather check whether the target element is referenceable?

bluss:
I have updated to use is_valid_label_literal_id.

About underscore:

- I think that Typst is in a position to decide any convention for private/semi-hidden labels, doesn't have to use underscore just because it's popular already, packages can adapt. When I use elembic, I see a lot of `__elembic` names, so skipping initial underscore is of course close to mind to lots of authors.
- No reason to add a new convention in this PR, but this PR makes it an urgent issue because if no other action is taken, one of the ignored characters effective from this PR will be used as an emerging, new convention. Let's say for example using `,` or `/` or `@`.
- If underscores are a more desired convention than any of those, then we should act now.

laurmaedje:
@bluss What do you think about the idea of checking whether the target element is referenceable?

bluss:
@laurmaedje Yes, including only referenceable items for @-completion seems like a good usability change, it should help most common users. IMO, as long as it handles extensions to this (achieved with show ref), then it's a good idea. I could see if I can implement that

laurmaedje:
@bluss Hey, is this still on your radar / do you need any input from my side to proceed?

bluss:
@laurmaedje I'm sorry for the delay here.

I think we should avoid filtering referenceable since it has false negatives for power users using `show ref`, we talked about that, so then the plan should be to filter out those with invalid characters, and also skip underscore prefix labels in reference completion (not label completion), at least that's what I think.

Maybe same-file or not-cross-package would be ideal for underscored labels, but I don't know yet if that's realistic.

I won't be able to look at this until next week.

laurmaedje:
No worries about the delay! We're all busy. :)

There's no rush, I just wanted to check in whether you still planned to work on this at some point or whether there were any blockers. Your suggested changes sound good.

laurmaedje:
I'll close this for now due to inactivity, to keep the PR tracker organized. Feel free to ping me if you do end up getting back to it and I'm happy to reopen!

PR #6845, opened 2025-09-01
nix: refactor, split files, simplify

drupol:
Hello,

Opening this PR as a proposal to split the huge `flake.nix` file into smaller chunk (modules).

It is now using some kind of file autoloader, so that the `flake.nix` don't need to be modified any longer, unless when adding `inputs`, which is kinda rare.

Along the way I removed a couple of things that were not needed any longer.

Let me know what you think of the idea.

### DIFF
--- a/nix/pkgs/typst-image/package.nix
+++ b/nix/pkgs/typst-image/package.nix
@@ -1,0 +1,18 @@
+{
+  lib,
+  dockerTools,
+  local,
+  ...
+}:
+dockerTools.buildImage {
+  name = "typst";
+  tag = "latest";
+
+  copyToRoot = [
+    local.typst
+  ];
+
+  config = {
+    Cmd = [ "${lib.getExe local.typst}" ];
+  };
+}


laurmaedje:
To be honest, I don't really like this. I'm already not a fan of the existing Nix stuff in the repository as I believe it is only used by relatively few people and it's fairly high maintenance (because everything is pinned, stuff needs bumping frequently and stuff can break; luckily other people have stepped up to do the reviewing, but still). 

Typst is a pretty stock Rust project. You need a recent Rust toolchain and that's more or less about it. Nix is already the only non-Rust technology except for Docker that has a "blessed" integration with Typst in some way. With the current flake, at least it is contained to a single file with 170 lines (which is not that huge in my view), but with this PR it more than doubles and gets a top-level folder. I am fairly minimalistic and try to avoid clutter of top-level config files and directories.

I appreciate people's love of Nix and totally believe that it solves complex problems. But adding so much code to a project that doesn't really have anything to do with Nix feels like bad coupling to me and an unnecessary maintenance burden. People could request integrations with tons of other build systems (which I'd very likely reject) and it wouldn't be that different in my view. To me it seems like cargo is kind of just working fine and I guess I just don't understand why adding Nix support to every single project individually is a good idea.

PgBiel:
I agree with Laurenz, this change doesn't seem to provide much benefit to the project. As a single file, it's clearer that it is not part of the actual source code, and contributors who would like to improve or bump the Nix flake only have to change one file. This change would require properly instructing contributors regarding structure, documenting and maintaining that, which feels like additional burden for us without much gain. And either way, the flake is supposed to be simple enough to build Typst and have a single devshell.

(It's also easier for me to review. :))

Of course, there's no problem in maintaining a more complex flake in a separate repository though. (See https://github.com/vic/gleam-nix, a more complex flake for the Gleam language, for comparison.)

laurmaedje:
It was my understanding that the flake can't be really be maintained out of tree. If it was possible to move it out, that would be ideal and I'd happily link to some blessed community one.

drupol:
Thank you all for your inputs, I appreciate the honesty.

Maintaining the flake inside this project automatically provides:

1. A way to make sure anyone using Nix can build the project, on any machine, reproducibly !
2. A way to have access to the very latest snapshot of the Typst source tree for free (_no need to have a `src` attribute fixed to a specific revision and with a specific hash_)

I could go on further but I have the feeling that a choice has been made already here... so there's no need.

Just for my own information, in case you decide to remove Nix, will you also remove the Docker related files?

Thanks

PgBiel:
> It was my understanding that the flake can't be really be maintained out of tree. If it was possible to move it out, that would be ideal and I'd happily link to some blessed community one.

I believe having a basic flake that can at least compile Typst in Typst's repo is a net positive for the project, as it allows for a reproducible build of any commit with Nix from multiple environments without any Rust-/project-specific installation (e.g. with `nix run github:typst/typst/abfd88ec -- compile thing.typ`). Main thing is about this always being easily available for the entire repo history.

But anything beyond that specific task is extra in my opinion, since we won't be maintaining that. And the current flake already does that well enough.

With that said, it's a bit unfortunate that we couldn't test the flake build on CI due to the long build times. Something I'd like to experiment with is adding a weekly flake build test (which doesn't block usual CI), similar to hayagriva's weekly archive update check. Could be interesting to at least have some idea of how it's going from time to time.

laurmaedje:
> Just for my own information, in case you decide to remove Nix, will you also remove the Docker related files?

For now, both will probably just stay as-is.

PR #6852, opened 2025-09-02
Import translations from babel and cleveref

### LINKED ISSUE #6581: Consistent handling of i18n translations for Figure, Section, etc
This issue tracks how to get consistent handling of i18n translations for Figure, Section, Bibliography, etc. by relying on existing translations used for LaTeX and cleveref.

Based on comments from:
* https://github.com/typst/typst/pull/6519#issuecomment-3018270925
* https://github.com/typst/typst/pull/6519#issuecomment-3052537615

From @laurmaedje:
> This makes me think that we should probably have done that from the start. Just researching the LaTeX name data once and supporting all the languages they support immediately instead of having the terms decided by contributors one by one. I think that came up once in the past, but nobody ever followed up on it...

> Thanks! While we should research the LaTeX data more, no need to further wait with this PR.

For reference, the following LaTeX source was used to ensure that the same Swedish translations are used for Typst as those used by LaTeX and cleveref. It can be adapted for other languages,

```latex
\documentclass[12pt,a4paper,swedish]{article}

\usepackage{graphicx}
\usepackage{listings}

\usepackage[swedish]{cleveref}

\usepackage[swedish]{babel}

\begin{filecontents}{foo.bib}
@phdthesis{typst_phd,
	title={A Programmable Markup Language for Typesetting},
	author={M{\"a}dje, Laurenz},
	year={2022},
	school={Technical University of Berlin}
}
\end{filecontents}

\begin{document}

\tableofcontents

\section{foo}
\label{sec:foo}

Se \cref{fig:foo}.

Se \cref{tab:bar}.

Se \cref{eq:baz}.

Se \cref{sec:foo}.

Se \cref{lst:foobar}.

Se \cite{typst_phd}.

\begin{figure}[htbp]
	\centering
	\includegraphics[width=0.5\textwidth]{example-image-a}
	\caption{foo}
	\label{fig:foo}
\end{figure}

\begin{table}[htbp]
	\caption{bar}
	\label{tab:bar}
	\begin{tabular}{ccc}
	foo & bar & baz \\
	\end{tabular}
\end{table}

\begin{equation}
	x = y^2
	\label{eq:baz}
\end{equation}

\begin{lstlisting}[language=C, caption={foobar\label{lst:foobar}}]
printf("hello world");
\end{lstlisting}

\bibliography{foo.bib}
\bibliographystyle{ieeetr}

\end{document}
```

eltos:
Imports all available translations from [babel](https://ctan.org/tex-archive/macros/latex/required/babel/base) and [cleveref](https://ctan.org/tex-archive/macros/latex/contrib/cleveref).

Closes #6581 by following up on [this comment](https://github.com/typst/typst/pull/6519#issuecomment-3018270925):
> This makes me think that we should probably have done that from the start. Just researching the LaTeX name data once and supporting all the languages they support immediately instead of having the terms decided by contributors one by one. I think that came up once in the past, but nobody ever followed up on it...

The PR has multiple commits:
- 35d6a1f1a41ab960a7b06819f575d903017b29a3 imports all translations, overwriting those that already existed in typst
- 717ba1a8eddff9841a8e2af6bd7d634705b22146 reverts the overwritten terms, providing a clean diff which could be reviewed by a [native speaker](https://github.com/typst/typst/commits/main/crates/typst-library/translations)
  ```diff
  - term used by babel/cleveref
  + term used by typst as of today
  ```
- 8595bb9936a64c57a1aa0625d89b1b295c44d142 adds the translations to `lang.rs` 
- 4c27710b8aa957f84b3dd0ef78d152c07b47ce9a adds language constants and fixes a wrong assignment of Filipino/Tagalog

Overall, the PR **does not change** any of the existing translations, **only adds** missing terms and locals ones.

## Import script

I used the following python script to parse babel and cleveref sources, extract the localized terms and generate typst translations. Since these contain many more terms than currently used by typst (some including abbreviations and plurals), I post the script here in case it might be useful in the future:

[babel_cleveref_to_typst.py](https://github.com/user-attachments/files/22680355/babel_cleveref_to_typst.py)

## Notes  
- Not all translations are complete in babel/cleveref, I have also imported incomplete translations if there is at least one term translated
- Capitalization of terms frequently differs between babel/cleveref and typst (`page` vs `Page`), as well as between locals. As stated above, existing terms are kept as-is, including capitalization. New terms are imported as-is, with whatever capitalization is used by babel/cleveref for the respective locale.
- For some locals, cleveref provides abbreviations (`Fig.` vs. `Figure`), but I chose to import the form babel provides. I noticed some inconsistent use of long-form vs. abbreviation in babel as well as typst translations. My feeling is, that this is more a question of editor style, but I don't know if some languages have a strong preference for either, a question which only a native speaker can answer, so I just kept them as-is.
- Both babel and cleveref are licensed under the [LaTeX Project Public License](https://www.latex-project.org/lppl.txt). ~I *believe* there is no problem, as this is not a derived work and "[The LPPL] license places no restrictions on works that are unrelated to the Work".~ **EDIT:** A notice was added to the notice file. However, I am not qualified to make a legal statement about the use of the extracted terms under typst's Apache-2.0 license. 

## Future

The following follow-up features might be useful, but are out of scope for this PR:
- Add singular/plural terms and natively combine multiple adjacent reference in typst using the plural forms, simmilar to how multiple adjacent citations are combined
- Provide a function to get translated terms in the different forms (long/abbreviation, singular/plural, capitalized/in-text) for use by templates or packages, such as https://typst.app/universe/package/smartaref


laurmaedje:
> I believe there is no problem, as this is not a derived work

Is it not? But regardless, it should be good if we just add an entry to the NOTICE file. The hyphenation patterns Typst uses are also LPPL and LPPL and Apache-2 are compatible.

eltos:
> add an entry to the NOTICE file

Done

mewmew:
Thanks for working on this issue @eltos! Really happy to see the machine generated consistent handling of translations <3

laurmaedje:
I'd like to land this in 0.14, but https://github.com/typst/typst/pull/6619 introduced a few new terms and resulted in a few conflicts. If you have time to take a look, that would be appreciated! Otherwise, I'd do it myself.

I also plan to do rebase & merge (or fast-forward if possible) to keep the individual commits as they are very useful for posterity, so a force push would be ideal.

eltos:
I have rebased onto main and added "footnote" translations while keeping the commit structure.
The terms "email" and "telephone" are unfortunately neither available in babel nor cleveref.

**EDIT:** NOTICE file edit squashed with import commit, PR description updated.

laurmaedje:
Thank you. This is great work!

Andrew15-5:
Are all [ISO 639-1/2/3 language codes](https://en.wikipedia.org/wiki/ISO_639) were included in the `Lang` impl? Does it add something useful if most of them are unused (more than 200, IIUC)?

Issue #5792, opened 2025-02-01
Find a way to prevent non-sticky block footgun in heading show rules


PgBiel:
### Description

A bit of a long title, but basically I've seen this pattern more than once:

```typ
// Attempting to fully customize a heading:
#show heading: it => {
   // Do some stuff
   block(new heading body /* could also be just it.body */)
}

= Headings

= Headings
```

Importantly, the `block` there does not have `sticky: true`. As a result, orphan prevention no longer works for headings. This fixes it:

```typ
// Attempting to fully customize a heading:
#show heading: it => {
   // Do some stuff
   // NOTE: 'sticky: true' added below
   block(sticky: true, new heading body /* could also be just it.body */)
}

= Headings

= Headings
```
Now headings retain orphan prevention behavior. But you don't notice it until your document has tons of orphans...

How should we prevent this?

Something like `#show heading: set block(sticky: true)` could work, but perhaps it's bad to apply it recursively to all blocks inside the heading. (A bit of an annoying limitation to be frank.)

### Typst version

- [ ] I am using the latest version of Typst

VassilisRoussochatzakis:
Can confirm that I fell for this trap and this may be a useful change!

laurmaedje:
Your example already works. In
```typ
#show heading: it => block(it.body)
#lorem(600)
= Heading
= Heading
#lorem(100)
```
the heading is already sticky since we already [use a show-set rule internally.](https://github.com/typst/typst/blob/f239b0a6a1e68a016cacf19eeef2df52e4affeb9/crates/typst-library/src/model/heading.rs#L326)

The non-stickyness only comes into play if you don't create a block at all, but I'm not sure we can do anything about that except improve documentation since you could also conceivably want a run-in inline heading.

Enivex:
A first step could be improving the documentation.

PR #6856, opened 2025-09-03
Recommend wrapping `heading` content in sticky `block`

Marcono1234:
Relates to #5792

The [`block.sticky` documentation](https://typst.app/docs/reference/layout/block/#parameters-sticky) was already hinting that headings use sticky blocks, but the `heading` documentation itself was not mentioning that sticky blocks are recommended.


crates/typst-library/src/model/heading.rs
```diff
@@ -29,6 +29,12 @@ use crate::text::{FontWeight, LocalName, TextElem, TextSize};
 /// [outline] of all headings for you. To exclude one or more headings from this
 /// outline, you can set the `outlined` parameter to `{false}`.
 ///
+/// When using a [show rule]($styling/#show-rules) to customize the content
+/// of a heading, make sure to wrap the content in a [`block`]($block) (which
+/// is implicitly [sticky]($block.sticky) for headings). This avoids that the
+/// heading becomes an "orphan", that is, it remains at the end of the page
+/// but subsequent content is wrapped to the next page.
+///
```
Marcono1234: I hope that I got the `$` links correct here. Please let me know if they have to be changed.

docs/reference/language/styling.md
```diff
@@ -86,14 +86,16 @@ that formats headings for a fantasy encyclopedia.
...
-  #set text(font: "Inria Serif")
-  \~ #emph(it.body)
-     #counter(heading).display(
-       it.numbering
-     ) \~
-]
+#show heading: it => {
+  set align(center)
+  set text(font: "Inria Serif")
+  block[
```
Marcono1234: Included a `block[...]` call here, and for simplicity changed the enclosing from a content to a code block to omit all the leading `#`s.

It seems all other examples either intentionally did not use `block` (to create inline headings) or were already using `block`.

**Important:** Wrapping this here in a `block` seems to have reduced the spacing between heading and subsequent text. Is that ok for this example, or should this be improved somehow? (if so, what do you recommend?)

crates/typst-library/src/model/heading.rs
xkevio: It should be clearer that by "contents" we mean the fields of `it` such as `it.body` since `it` will already be in a sticky block.

Marcono1234: But this is only relevant for the `body` field, right? Because all other fields don't provide access to the header text, so there is no risk that users forget to use a `block`.

What about something like this?
> When using a show rule to customize the content of a heading <ins>by accessing its `body`</ins>, make sure to ...

docs/reference/language/styling.md
laurmaedje: The spacing change is no problem, actually it looks better now. The example before wasn't particularly idiomatic and should've been changed a long time ago to use a block and show-set rules for the things that can. I wanted to add some extra explanations here without hijacking your PR, so I created a separate PR for bringing this example up to speed: https://github.com/typst/typst/pull/6871

laurmaedje:
Thanks!

PR #6862, opened 2025-09-04
Document and test `Sides<T>` parameters

YDX-2147483647:
This PR is extracted from #6764 and resolves comments on `container.rs` in that PR.

- Document that `box.inset` and `box.outset` accept **dictionaries** containing top, right, x, y, rest, etc.

- Document what will happen if `box.inset` (or `page.margin`, `rect.stroke`, `table.stroke`) receives a dictionary with some keys **omitted**.

- Document that `box.inset` and similar parameters accept **relative** length, and add a test for that.

  Previously, there isn't any test that passes a ratio to `{box,block,rect,table,…}.inset`.

  The formula in the added test comes from [the `grow` function in typst-layout](https://github.com/typst/typst/blob/main/crates/typst-layout/src/pad.rs#L61-L93).

  It is important to mention that ratios are relative to the box size. People might mistakenly think they are relative to the size of available space, as in [Inset paragraph by exactly half a page - Questions - Typst Forum](https://forum.typst.app/t/inset-paragraph-by-exactly-half-a-page/5898/2).

## Things not covered in this PR

- Add links from `{table,grid}.inset` to `box.inset`.

  #6764 will add examples on `inset`. Theses links will be added together.

- Document that `page.margin` also accepts relative lengths.

  This feature is never tested and the behaviour looks buggy (for instance, `#set page(margin: (y: 20%), height: auto)`).
  Moreover, the only use case I know might deserve a dedicated API.

  > The only use case I know is [directly setting the width of the type area, instead of the paper width](https://typst-doc-cn.github.io/clreq/#directly-setting-the-width-of-the-type-area-instead-of-the-paper-width), like `#set page(margin: (x: (100% - 42em) / 2))`.

  Therefore, it requires more discussion.

### DIFF
--- a/crates/typst-library/src/layout/container.rs
+++ b/crates/typst-library/src/layout/container.rs
@@ -69,19 +69,16 @@ pub struct BoxElem {
 
     /// How much to pad the box's content.
     ///
-    /// This can be:
-    /// - a single length for all sides
-    /// - a dictionary of lengths for individual sides
+    /// This can be a single length for all sides or a dictionary of lengths
+    /// for individual sides. When passing a dictionary, it can contain the
+    /// following keys in order of precedence: `top`, `right`, `bottom`, `left`
+    /// (controlling the respective cell sides), `x`, `y` (controlling vertical
+    /// and horizontal insets), and `rest` (covers all insets not styled by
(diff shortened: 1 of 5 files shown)


laurmaedje:
Thanks!

PR #6867, opened 2025-09-04
Add a link from `math.vec` to `math.{arrow,bold}` in docs

YDX-2147483647:
LaTeX `\vec` ≠ Typst `vec` surprises some people, like Sinthoras who posted [How to write a vector arrow over a variable? Equivalent of LaTeX vec function? - Questions - Typst Forum](https://forum.typst.app/t/how-to-write-a-vector-arrow-over-a-variable-equivalent-of-latex-vec-function/5854).

I didn't mention `bold(upright(v))` or `hat(n)`, because it's too obvious.
There are also people who are confused by the missing space in `arrow(AB)`, but I don't think `math.vec` is the right place to explain.

### DIFF
--- a/crates/typst-library/src/math/matrix.rs
+++ b/crates/typst-library/src/math/matrix.rs
@@ -20,6 +20,9 @@ const DEFAULT_COL_GAP: Em = Em::new(0.5);
 /// Content in the vector's elements can be aligned with the
 /// [`align`]($math.vec.align) parameter, or the `&` symbol.
 ///
+/// This function expresses a vector as its components. For symbolic notation,
+/// use [`arrow`]($math.arrow) or [`bold`]($math.bold) instead.
+///
 /// # Example
 /// ```example
 /// $ vec(a, b, c) dot vec(1, 2, 3)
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/math/matrix.rs
```diff
@@ -20,6 +20,9 @@ const DEFAULT_COL_GAP: Em = Em::new(0.5);
 /// Content in the vector's elements can be aligned with the
 /// [`align`]($math.vec.align) parameter, or the `&` symbol.
 ///
+/// This function expresses a vector as its components. For symbolic notation,
```
laurmaedje: Is "symbolic notation" a clear term where people will know what is meant here? I hadn't heard of it.

YDX-2147483647: How about _To denote a vector symbolically_? (This is the original version before I let LLM improve it.)

laurmaedje: Maybe, I'm not that familiar with math lingo. Let me ask in the Math forge.

laurmaedje: As @bluss said on Discord

> Maybe just "To denote a vector in mathematical notation, use arrow or bold".  Or even "arrow or bold are commonly used" to let the suggestion not be so direct. Vector is often bold + upright

I like both parts of that suggestion.

laurmaedje:
Thanks!

PR #6871, opened 2025-09-05
Make show rule example in reference more idiomatic

laurmaedje:
The show rule example in the reference is really not idiomatic anymore. The example hadn't been updated since the show-set rule changes and demonstrated the bad practice of putting set rules into show rules instead of having separate show-set rules. It was also missing a `block` to benefit from stickiness, as discovered here: https://github.com/typst/typst/pull/6856

### DIFF
--- a/docs/reference/language/styling.md
+++ b/docs/reference/language/styling.md
@@ -76,23 +76,39 @@ But this stays black.
 
 With show-set rules you can mix and match properties from different functions to
 achieve many different effects. But they still limit you to what is predefined
-in Typst. For maximum flexibility, you can instead write a show rule that
-defines how to format an element from scratch. To write such a show rule,
-replace the set rule after the colon with an arbitrary [function]. This function
-receives the element in question and can return arbitrary content. The available
-[fields]($scripting/#fields) on the element passed to the function again match
-the parameters of the respective element function. Below, we define a show rule
-that formats headings for a fantasy encyclopedia.
+in Typst. For maximum flexibility, you can instead write a _transformational_
+show rule that defines how to format an element from scratch. To write such a
+show rule, replace the set rule after the colon with an arbitrary [function].
+This function receives the element in question and can return arbitrary content.
+The available [fields]($scripting/#fields) on the element passed to the function
+again match the parameters of the respective element function. Below, we define
+a show rule that formats headings for a fantasy encyclopedia.
+
+The show rule itself adds tilde characters around the title (these must escaped
+with a backslash because they indicate a non-breaking space), emphasizes the
+title with italics, and then displays the heading counter after the title.
+
+For this example, we also wanted center alignment and a different font. While we
+could've added these set rules into the existing show rule, we instead added
+them as separate show-set rules. This is good practice because now these rules
+can still be overridden by later show-set rules in the document, keeping styling
+composable. In contrast, set rules within a transformational show rule would not
+be overridable anymore.
 
 ```example
 #set heading(numbering: "(I)")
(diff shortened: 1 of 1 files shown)


docs/reference/language/styling.md
```diff
@@ -76,23 +76,39 @@ But this stays black.
...
+  font: "Inria Serif"
+)
+
+#show heading: it => block[
+  \~
+  #emph(it.body)
+  #counter(heading).display(
+    it.numbering
+  )
+  \~
```
MDLC01, line 111:
```suggestion
#show heading: set align(center)
#show heading: set text(font: "Inria Serif")
#show heading: it => block[
  \~
  #emph(it.body)
  #counter(heading).display(it.numbering)
  \~
```

It may be more readable to make a wide example with the preview below.

docs/reference/language/styling.md
```diff
@@ -76,23 +76,39 @@ But this stays black.
...
-in Typst. For maximum flexibility, you can instead write a show rule that
-defines how to format an element from scratch. To write such a show rule,
-replace the set rule after the colon with an arbitrary [function]. This function
-receives the element in question and can return arbitrary content. The available
-[fields]($scripting/#fields) on the element passed to the function again match
-the parameters of the respective element function. Below, we define a show rule
-that formats headings for a fantasy encyclopedia.
+in Typst. For maximum flexibility, you can instead write a _transformational_
+show rule that defines how to format an element from scratch. To write such a
+show rule, replace the set rule after the colon with an arbitrary [function].
```
MDLC01, line 81:
```suggestion
show rule, replace the set rule after the colon with an arbitrary [function].
The function is often defined inline with `it => ..` using the
[unnamed function syntax]($function#unnamed). Its argument is typically named
`it` by convention.
```

The concept of a show rule is weird to newcomers, and the syntax may appear frightening. Clarifying that the `=>` is part of the unnamed function syntax may help the readers understand that `it` does not have any special meaning (i.e., it's not a keyword).

Alternatively, maybe the first example should use a named function for clarity, and then be compacted using an unnamed function as a second example.

docs/reference/language/styling.md
vmartel08: Would it be good to add an explanation for not using `#show heading: emph` instead of `#emph(it.body)` and leaving it in the transformational show rule?

docs/reference/language/styling.md
```diff
@@ -76,23 +76,39 @@ But this stays black.
...
+This function receives the element in question and can return arbitrary content.
+The available [fields]($scripting/#fields) on the element passed to the function
+again match the parameters of the respective element function. Below, we define
+a show rule that formats headings for a fantasy encyclopedia.
+
+The show rule itself adds tilde characters around the title (these must escaped
```
hkBst:
```suggestion
The show rule itself adds tilde characters around the title (these must be escaped
```

docs/reference/language/styling.md
```diff
@@ -76,23 +76,39 @@ But this stays black.
...
+with a backslash because they indicate a non-breaking space), emphasizes the
```
hkBst:
```suggestion
with a backslash otherwise they represent non-breaking space)), emphasizes the
```

docs/reference/language/styling.md
laurmaedje: > Would it be good to add an explanation for not using #show heading: emph instead of #emph(it.body) and leaving it in the transformational show rule?

I'm not sure how best to do that right now, so I'll leave this as potential future work.

PR #6872, opened 2025-09-06
Add -e flag to winget installation method

blinkerfluessigkeit:
The `-e` (`--exact`) flag keeps winget from installing another package that was fuzzily matched. This is not *necessary* here, as the provided `--id` should match only Typst and at the moment, `winget search typst` only returns Typst itself and tinymist anyway, but it is another level of security, which I personally think should be recommended to users when installing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Typst's CLI is available from different sources:
       - View [Typst on Repology][repology]
       - View [Typst's Snap][snap]
   - macOS: `brew install typst`
-  - Windows: `winget install --id Typst.Typst`
+  - Windows: `winget install -e --id Typst.Typst`
 
 - If you have a [Rust][rust] toolchain installed, you can install
   - the latest released Typst version with


laurmaedje:
could `--id Typst.Typst` ever match something else given that it's a perfect match for Typst?

blinkerfluessigkeit:
Hmm, I seem to remember that when I started using winget a few years ago, weird things could happen without `-e` and I've always used it since. I tested it with `winget install --id Microsoft.VisualStudioCode` just now, because there is also `Microsoft.VisualStudioCode.CLI`, `Microsoft.VisualStudioCode.Insiders` and `Microsoft.VisualStudioCode.Insiders.CLI` and it installed the correct package. So either I misremembered or Microsoft has fixed the weird behaviour somewhere along the line. It also installs correctly with `winget install --id microsoft.visualstudiocode`.

So I honestly don't know if it is even possible (anymore?) that it could match something else. You can close this, if you like.

laurmaedje:
I think I'd rather keep it simple then. Thanks for checking!

Issue #4635, opened 2024-07-27
Keys and string values in JSON listings are not highlighted differently


JKRhb:
### Description

I am not sure whether this is the right place to report this potential bug, but when using JSON listings, I've noticed that the formatting is a bit off as the keys and the string-based values at the same level always have the same color, which makes listings not that well readable. This is even the case with custom `.tmTheme` files and adjusted color schemes.

The issue (which is especially relevant in the case of larger JSON documents) becomes apparent if you compare the following listing with a rendered Typst-based version of it:

```json
{
  "foo": "bar"
}
```

![image](https://github.com/user-attachments/assets/b424fe24-a18a-4e66-a0f2-c3e57bd965f3)

### Reproduction URL

https://typst.app/project/ry9VAeYLiZpHr4v9o_EBsE

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

JKRhb:
Investigating this issue a bit, I realized that the keys and values being the same color is apparently the default setting in Sublime Text. However, even when using a syntax file that can alter this behavior in Sublime Text itself (like [this one](https://github.com/aurule/json-kv)), this does not seem to have an effect in Typst which seems to be related to https://github.com/typst/typst/issues/1952 (at least in the case of the syntax file that I was trying to use).

JKRhb:
I've now managed to deal with this issue in one of my documents by both editing a Sublime syntax and tmTheme file, so this issue is solved for me :)

However, I think it would be nice if having different colors for keys and values would be the default, as users I image that this is the behavior other users would also expect. I suppose to integrate that into Typst, though, it would be necessary to change the default style provided by the [syntect](https://github.com/trishume/syntect) dependency, right? Or would defining custom syntax and tmTheme files as compiler assets also be an option to deal with this issue?

laurmaedje:
The theme is [already custom](https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-library/src/text/raw.rs#L884). Not sure what a better color in the default theme would be though.

JKRhb:
> The theme is [already custom](https://github.com/typst/typst/blob/494e6a64225261add59c3c9b1cd923bfdfa77961/crates/typst-library/src/text/raw.rs#L884). Not sure what a better color in the default theme would be though.

[...]

PR #6754, opened 2025-08-13
More accessible color scheme for raw blocks

reknih:
Previously, four colors used did not achieve WCAG 2.1 contrast level AA for normal text on white background:

- `#8a8a8a` used for comments
- `#818181` used for raw blocks in markup
- `#298e0d` used for string, math control characters, and diff deletions
- `#d73a49` used for keywords, and removals in diffs

These colors were replaced by colors achieving the AA level for normal text on a background with the color `#fdfdfd`:

- `#8a8a8a` has been replaced by `#737488` (Contrast ration 3.39:1 to 4.5:1)
- `#818181` has been replaced by `#707180` (Contrast ration 3.82:1 to 4.73:1)
- `#298e0d` has been replaced by `#198810` (Contrast ration 4.15:1 to 4.51:1)
- `#d73a49` has been replaced by `#d73948` (Contrast ration 4.49:1 to 4.51:1)

PR #6873, opened 2025-09-07
Use different colors for object keys and string-based values in JSON listings

JKRhb:
This simple PR should resolve https://github.com/typst/typst/issues/4635 by adding an `item` for the scope `meta.mapping.key.json string.quoted.double.json` to the default `raw` theme, changing the default color that is used for the keys within JSON objects. This makes it possible to differentiate keys from string-based values more easily.

I am currently not sure whether the color I've chosen so far is the best, so I am open for feedback here :) The changes themselves are hopefully trivial enough to still become part of the upcoming release.


JKRhb:
As mentioned in https://github.com/typst/typst/issues/4635#issuecomment-2955816455, another option could be using green for the keys and light blue for the values, which can also be set quite easily in the custom theme:

image

If that (or a different color scheme) would be preferable, this could be updated quite quickly.

laurmaedje:
The contrast of the light blue against the background is quite bad (3.16:1 against `#fdfdfd` background, you can [check online](https://webaim.org/resources/contrastchecker/)), which makes this non-accessible (we recently changed it to be more accessible in https://github.com/typst/typst/pull/6754). Maybe we can use one of the existing colors instead? I'd imagine that some of them aren't used at all yet in JSON.

JKRhb:
Thank you for the feedback, @laurmaedje! I've now changed the theme to keep `#198810` as the color for the keys and use `#4b69c6` (which was already used elsewhere) for string-based values, which I think looks quite well and also passes the [accessibility test](https://webaim.org/resources/contrastchecker/?fcolor=4B69C6&bcolor=FDFDFD).

To keep the default green color for the keys in nested objects, I had to still explicitly define

    meta.mapping.key.json string.quoted.double.json

since

    meta.mapping.value.json string.quoted.double.json

overrides the key color starting on the second level for some reason if the key color is not explicitly defined:

image

I therefore expanded the test I added before to also cover that case, yielding the following result when including both the color definitions for keys and values:

image

laurmaedje:
I think that's much better. On second thought, I like your idea of highlighting the keys instead of the values blue better since the values feel more normal (= normal green string) while the keys are a bit more special.

JKRhb:
> I think that's much better. On second thought, I like your idea of highlighting the keys instead of the values blue better since the values feel more normal (= normal green string) while the keys are a bit more special.

Thank you! I've changed the theme accordingly :) I've also noticed by the way that the blue keys also correspond with the default theme used for YAML. To ensure that both types of `raw` elements stay consistent in terms of their theme, I took the freedom of also adding an equivalent test case for YAML, which also illustrates the similarities.

JKRhb:
(Rebased my branch against the latest state of `main`).

@laurmaedje Let me know if there are any more things that should be changed to make this PR mergeable :)

laurmaedje:
Nothing to do, I was just busy. Thanks!

PR #6874, opened 2025-09-07
Raise an error when mutating method is accessed outside of a call expression

MikhailKravets:
Currently, Typst allows you to access mutating methods as fields on types which leads to inconsistent behavior. For instance, the following code works without errors

```typ
#let arr = (1, 2, 3)
#let push = array.push
#push(arr, 4)
#arr
```

However, since `arr` is cloned when passed to the `push` and due to the COW semantics of Arc, the array `arr` remains unmodified.

This PR adds an error that is raised when a mutating method is accessed on a type outside of a call expression.

### DIFF
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -2,13 +2,15 @@ use ecow::{EcoVec, eco_vec};
 use typst_library::diag::{At, SourceResult, bail, error, warning};
 use typst_library::engine::Engine;
 use typst_library::foundations::{
-    Array, Capturer, Closure, Content, ContextElem, Dict, Func, NativeElement, Selector,
-    Str, Value, ops,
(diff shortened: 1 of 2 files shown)


laurmaedje:
While perhaps a bit confusing, I don't think it's necessarily inconsistent for these functions to be accessible and behave as they do. The fact that they don't modify the passed parameter is consistent with the fact that arguments are always passed by value, with exception of the receiver of a method call.

A mutable method that returns a value can still behave somewhat sensible (i.e. still producing a sensible return value, just without the side effect):
```typ
#let arr = (1, 2, 3)
#let pop = array.pop
#pop(arr)
```

Moreover, I think it makes sense for these methods to be present consistency wise, in particular should we add more reflection capabilities. Overall, I think this is fine as-is. Thank you for looking into this nonetheless!

PR #6875, opened 2025-09-07
Bump `codex`

MDLC01:
This is the companion PR for https://github.com/typst/codex/pull/114.

As mentioned on Discord,[^1] this PR won't change the appearance of emojis and symbols with the default font because emoji presentation sequences that have a corresponding text presentation that is supported by the main font will fall back to text presentations with the main font instead of causing a font fallback to a font that supports the emoji presentation.

I added a test which, for the reason mentioned above, does not show the difference between text presentation sequences and emoji presentation sequences.

[^1]: https://discord.com/channels/1054443721975922748/1277628305142452306/1413479337486975036


tests/suite/symbols/symbol.typ
```diff
@@ -183,3 +183,7 @@ $ A sect B = A inter B $
 --- issue-5930-symbol-label ---
 #emoji.face<lab>
 #context test(query(<lab>).first().text, "😀")
+
+--- presentation-selectors ---
+#sym.copyright #emoji.copyright \
```
laurmaedje, line 200: Since the test does not show the difference, it would be good if there was a comment explaining that and when it should be updated.

MDLC01: Done

laurmaedje:
The latest bump causes test failures because of the implicit codex update.

MDLC01:
So this exposed an unfortunate warning:

```typ
#sym.bracket.double.r
```

results in

```
Warning: `bracket.l.double` is deprecated, use `bracket.l.stroked` instead
Warning: `bracket.r.double` is deprecated, use `bracket.r.stroked` instead
```

The first one is located on `double`, which causes a resolution to `bracket.l.double`, and the second is located on `r`.

This was already noted in https://github.com/typst/typst/pull/6441 (with a potential solution in https://github.com/typst/typst/pull/6441#issuecomment-3002011213), but I did not think of this case, which I expect to be quite ambiguous for the users.

laurmaedje:
Hmm yeah, these warnings are quite confusing...

laurmaedje:
I think my dream warning would be
```
warning: the `double` modifier is deprecated, use `stroked` instead
  ┌─ hi.typ:4:13
  │
4 │ #sym.bracket.double.r
  │              ^^^^^^
```

laurmaedje:
Thanks!

PR #6876, opened 2025-09-07
 lexer: avoid scanning past two newlines when tokenizing whitespace

Acelator:
Right now, each time the lexer reads a string and returns the next token, it computes the number of newlines following a whitespace. 

We just need to distinguish between 0, 1 and 2 (or more) newlines to determine if the SyntaxKind is a paragraph break or a space.  

Therefore, we can skip counting the number of newlines once we reach 2.


laurmaedje:
I think the performance benefit here is likely so very small that I'm not sure the slight reduction in readability is worth it.

laurmaedje:
I'll close this based on the comment above. Thanks still for your effort!

PR #6878, opened 2025-09-08
Fix typos

hpcfzl:
I ran the [`typos`](https://github.com/crate-ci/typos) tool on `typst/typst`, sifted through false positives, then added a couple of my own finds regarding units.

I'm not entirely sure about the *leaves* typo and those under the `tests` directory, because they're more part of the code rather than the comments.


laurmaedje:
Thanks!

Issue #4421, opened 2024-06-19
Typst panics after trying to parse an invalid `.sublime-syntax` file (syntect crate)


Andrew15-5:
### Description

```yaml
%YAML 1.2
---
name: lang
file_extensions:
  - a
scope: source
contexts:
  main:
    - match: '\'
```

````typ
#set raw(syntaxes: "a.sublime-syntax")
```a text```
````

```
thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/syntect-5.2.0/src/parsing/regex.rs:70:53:
regex string should be pre-tested: ParseError(0, TrailingBackslash)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I can't use the `typst watch` while cooking some regex, because at least half of all regexes are just backslashes.

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

Andrew15-5:
The https://github.com/trishume/syntect/pull/548 is merged, now it's just a matter of time when the next syntect version is released. Then this can be fixed. Well, it can be fixed now if we use the Git version. Judging by the last release version, it's probably better to use Git for now. Switch to crates.io version when and if it will happen.

PR #6883, opened 2025-09-08
Fix panic on unescaped backslash in .sublime-syntax file (raw.syntaxes)

Andrew15-5:
See https://github.com/typst/typst/issues/4421#issuecomment-3267689889.

Fixes https://github.com/typst/typst/issues/4421.

So, everything is good, but I have no idea why the test keeps saying malformed range. I tried what feels like everything. It just doesn't wanna let it go. But without the error comment, previously it would panic, now it just gives the correct error from syntect (which is pretty long, but idk if it's bad). Also, the test parser is not aware of raw block, so it triggers error on `---` in raw block...

```text
 cargo test --workspace --test tests -- sublime
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.33s
     Running src/tests.rs (target/debug/deps/tests-a1818e67b68d8e7b)
failed to collect tests
❌ range is malformed (tests/suite/text/raw.typ:110)
error: test failed, to rerun pass `-p typst-tests --test tests`
```

### DIFF
--- a/tests/suite/text/raw.typ
+++ b/tests/suite/text/raw.typ
@@ -91,6 +91,22 @@ Year	Month	Day
     (* x (factorial (- x 1)))))
 ```
 
+--- raw-syntaxes-invalid-sublime-syntax ---
+// Prevent test parser from failing on "^---" line.
+#let sublime-syntax = ```yaml
+%YAML 1.2
+```.text + "\n---\n" + ```yaml
+name: lang
+file_extensions:
+  - a
+scope: source
+contexts:
+  main:
+    - match: '\'
+```.text
+
+// Error: 35-56 failed to parse syntax (Error while compiling regex '/': Parsing error at position 0: Backslash without following character)
+#raw("text", lang: "a", syntaxes: bytes(sublime-syntax))
 
 --- raw-theme ---
 // Test code highlighting with custom theme.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,18 +160,18 @@ dependencies = [
 
 [[package]]
 name = "bit-set"
-version = "0.5.3"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
 dependencies = [
  "bit-vec",
 ]
 
 [[package]]
 name = "bit-vec"
-version = "0.6.3"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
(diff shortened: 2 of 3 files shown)


Andrew15-5:
Blocked by what? Is the [new release](https://github.com/trishume/syntect/issues/599) required?

laurmaedje:
Yes, it was blocked on a new release. We don't add git dependencies on crates that we don't have publishing control over because it could block us from making a crates.io release.

Could you also update the minimum dependency version in `Cargo.toml`? The `Cargo.lock` update will do the job for the CLI, but not for dependants of Typst as a library.

laurmaedje:
Thanks!

Andrew15-5:
But...the test is failing... Did you read the OP?

laurmaedje:
CI was running while I clicked "Merge when ready" so I hadn't seen the failure just now. I had read the OP a week ago, but had forgotten about it.

tests/suite/text/raw.typ
```diff
@@ -91,6 +91,26 @@ Year	Month	Day
     (* x (factorial (- x 1)))))
 ```
 
+--- raw-syntaxes-invalid-sublime-syntax ---
+#set page(width: 180pt)
+#set text(6pt)
+
+// Prevent test parser from failing on "^---" line.
+#raw("text", lang: "a", syntaxes: bytes(sublime-syntax))
```
laurmaedje: This line produces a compiler error because `sublime-syntax` is not defined.

tests/suite/text/raw.typ
```diff
@@ -91,6 +91,26 @@ Year	Month	Day
...
+name: lang
+file_extensions:
+  - a
+scope: source
+contexts:
+  main:
+    - match: '\'
+```.text
+
+// Error: 17:35-17:56 failed to parse syntax (Error while compiling regex '/': Parsing error at position 0: Backslash without following character)
```
laurmaedje: You just need to remove the `17:` pieces, so `Error: 35-56 failed to parse syntax ...`. The test runner printed the absolute line of the failure in the test (so that you know where it failed), but the error annotations are relative (because that's much more stable and easier to write).

tests/suite/text/raw.typ
```diff
@@ -91,6 +91,26 @@ Year	Month	Day
     (* x (factorial (- x 1)))))
 ```
 
+--- raw-syntaxes-invalid-sublime-syntax ---
+#set page(width: 180pt)
+#set text(6pt)
+
```
laurmaedje: These lines can be removed as the test has no visual output.

tests/suite/text/raw.typ
Andrew15-5: Yeah, it was some debugging stuff or something.

Andrew15-5:
Soo... What about the `---` line? Can't parser check for raw block?

Making the test ugly, just so the parser accepts it, doesn't sound like a great idea, or rather a great parser.

laurmaedje:
The parser is just simple and fairly pragmatic. I think it's very rare to need `---`, so this is fine as-is.

Andrew15-5:
I tried modifying it...but a lot of edge cases started to pop out. Like backticks in code, or in math. So...it is already 50+ lines...yeah, I guess one ugly spot is not worth it...

PR #6889, opened 2025-09-09
Improve examples in the docs for under/over functions

YDX-2147483647:
Annotations in under/over functions are also in math mode.
However, `"numbers"`, the previous example, is valid in both code and math modes. It makes some people write `$ underbrace(..., $m n$) $` and meet _Error: Unclosed delimiter_.
The new example adds `n + 1` to emphasize that it's in math mode.

There's another person suggesting `tilde(gamma) + ln n`, but I feel it's too long.

Besides, I change `+ ... +` to `+ dots.c +`, because mid-line ellipsis looks better among operators with horizontal symmetry.


laurmaedje:
Thanks!

Issue #6639, opened 2025-07-21
Delegate styles to text formatting elements in `bibliography`


cAttte:
### description

currently, `BibliographyElem` applies styles derived from CSL directly to the content. [for example,](https://github.com/typst/typst/blob/b790c6d59ceaf7a809cc24b60c1f1509807470e2/crates/typst-library/src/model/bibliography.rs#L999) a `font-variant: small-caps` will result in `content.set(TextElem::smallcaps, Some(Smallcaps::Minuscules))`.

i propose, instead, that styles be indirectly applied by wrapping the content in the appropriate text formatting elements. for the example given, `content` would be wrapped in a `SmallcapsElem`.

### use case

this would allow `show` and `set` to match and style elements within bibliographies. for example, you may want to apply extra tracking to small-capped text in your whole document (see [Typst Forum: How to add tracking [...] in bibliography?](https://forum.typst.app/t/how-to-add-tracking-letterspacing-to-author-names-in-bibliography/5011/3)), which includes the bibliography. you may also want to redefine `bold` and `light`:

```typ
#show smallcaps: set text(tracking: .1em)
#set strong(delta: 400)
#show strong.where(delta: -100): strong.with(delta: -200)
```

you could argue that styling should happen in CSL. this is obviously true, but only to a limited extent: there are many rules (like the first one) that are supposed to apply document-wide, and not having them work in the bibliography just results in behavioral and visual inconsistency. there are also styles that are unachievable with CSL (all of the rules mentioned).

Enivex:
I think this falls under https://github.com/typst/typst/issues/942

cAttte:
well, that issue is pretty vague: the most concrete solution i read there was the creation of a `bibliography.entry` element, which wouldn't really fix this (it would just provide a slightly more sane workaround, i guess) and is clearly not trivial to implement (see #5932). the solution to this issue seems reasonably easy to implement and i'm down for making a PR if appropriate

laurmaedje:
I think this is indeed a different issue. I have been a bit torn on whether to use the semantic elements or not since the CSL spec is talking about visual formatting and, in principle, and bold weight does not equal strong emphasis. Citation styles are also typically speaking about bold I think. But I see how it can be useful and some change is necessary for HTML export either way. At the very least, smallcaps should use `SmallcapsElem` since that element isn't really semantic in the first place. The others I'm still unsure about.

MDLC01:
[...]

PR #6894, opened 2025-09-10
Add show rule for bibliography in HTML

breezykermo:
This is a first attempt to render a bibliography in the HTML target. There are quite a few open questions contained in the attempt, so I would be grateful for any feedback on which decisions are not in line with the overall Typst philosophy, and will be happy to make code adjustments in a timely manner.

I have followed the structure of pandoc's HTML rendering converting from a markup with citation support such as Orgmode. You can see a basic example of what inline and bibliography citations look like via pandoc [in this blog post](https://www.ohrg.org/24-02-07). This PR takes a similarly conservative approach to citations:

- **No internal referentiality in citations.** This could be done by wrapping inline citations in `<a href="#ref">` and giving bibliographic citations an appropriate `id`. Though it would be possible and nice to have this in the HTML, my assumption about why pandoc doesn't do it (and why Typst might not want it) is because there are ambiguous corner cases about what to link depending on the CSL. For example, if the inline citation is a range `[1-10]`, then it isn't clear whether the whole thing should be linked to the bibliographic citation `1` or `10`, whether just those numbers should be linked, or how otherwise to provide an appropriate reference.
- **No use of the `<cite>` element.** Though it could be appropriate here, the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/cite) indicate that the element "is used to mark up the title of a creative work", and so an implementation would have to use the understanding from CSL to find appropriate text and wrap it. The `<cite>` tag could go in the bibliographic entry, but in this first conservative implementation I haven't used it. (It's also not clear to me whether `<cite>` is commonly used in HTML on the web: pandoc, for example, doesn't use it.)
- **Use of Typst parameters where appropriate.** Both the `title` and the `full` attributes in the [bibliography function](https://typst.app/docs/reference/model/bibliography/) are implemented. The choice of `<h1>` for when `title` follows from pandoc, but is otherwise somewhat arbitrary.
- **Simple list structure with accessibility attributes for bibliography.** The bibliography itself is just nested `<div>`s, with `"list"` and `"listitem"` accessibility attributes. No extra classes are added, except for the `"typst-bibliography"` id on the outer div. I added this because I thought it might be nice to have _some_ id for downstream styling, but am open to removing it given the specifics on how Typst generates ids and classes in HTML hasn't yet been decided (to my knowledge).

### DIFF
--- a/crates/typst-html/src/link.rs
+++ b/crates/typst-html/src/link.rs
@@ -289,3 +289,4 @@ fn disambiguate(
         }
     }
 }
+
--- a/crates/typst-html/src/link.rs
+++ b/crates/typst-html/src/link.rs
@@ -289,4 +289,3 @@ fn disambiguate(
         }
     }
 }
-
--- a/crates/typst-html/src/rules.rs
+++ b/crates/typst-html/src/rules.rs
@@ -2,24 +2,50 @@ use std::num::NonZeroUsize;
 
 use ecow::{EcoVec, eco_format};
 use typst_library::diag::{At, bail, warning};
+use typst_library::engine::Engine;
 use typst_library::foundations::{
-    Content, NativeElement, NativeRuleMap, ShowFn, Smart, StyleChain, Target,
+    Content, NativeElement, NativeRuleMap, Packed, ShowFn, Smart, StyleChain, Target,
 };
 use typst_library::introspection::Counter;
 use typst_library::layout::resolve::{Cell, CellGrid, Entry, table_to_cellgrid};
 use typst_library::layout::{BlockBody, BlockElem, BoxElem, OuterVAlignment, Sizing};
 use typst_library::model::{
-    Attribution, CiteElem, CiteGroup, Destination, EmphElem, EnumElem, FigureCaption,
-    FigureElem, HeadingElem, LinkElem, LinkTarget, ListElem, ParElem, ParbreakElem,
-    QuoteElem, RefElem, StrongElem, TableCell, TableElem, TermsElem, TitleElem,
+    Attribution, BibliographyElem, CiteElem, CiteGroup, Destination, EmphElem, EnumElem,
+    FigureCaption, FigureElem, HeadingElem, LinkElem, LinkTarget, ListElem, ParElem,
+    ParbreakElem, QuoteElem, RefElem, StrongElem, TableCell, TableElem, TermsElem,
+    TitleElem, Works,
 };
 use typst_library::text::{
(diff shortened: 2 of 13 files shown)


bastienvoirin:
Nice work, can't wait to have HTML bibliographies!

### No internal referentiality in citations

I think it would be sensible (to match Typst's PDF export) and useful (for easier navigation) to have the citations and references linked, but maybe in a later iteration PR, keeping this one more or less "atomic".

> For example, if the inline citation is a range `[1-10]`, then it isn't clear whether the whole thing should be linked to the bibliographic citation `1` or `10`, whether just those numbers should be linked, or how otherwise to provide an appropriate reference.

For this specific citation range example, I envision two options:

1. Link to the bibliography title,
2. Link to the first reference.

### `<h1>` vs. `<h2>` for bibliography `title`

> The choice of `<h1>` for `title` follows from pandoc, but is otherwise somewhat arbitrary.

If I'm not mistaken, top-level Typst headings currently/should map to `<h2>` elements (see https://github.com/typst/typst/pull/5618#issuecomment-2627677285 and https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page).

### `<div>` for `listitem`

Why not use `<p>` elements for the inner, `listitem` elements?

laurmaedje:
It should definitely be `h2`, but more specifically it should just use normal Typst headings, which will automatically turn into the right thing. I think we can have a bit more code reuse with normal layout and get that more or less for free.

The ID `typst-bibliography` should imo be a `class="bibliography"` or maybe a role if there is an appropriate one. As for the specific DOM strucuture, I haven't looked into that yet, so can't give an opinion right now.

As for linking, that would be important, but it should "just work" if we use `LinkElem` instead of `.linked` because that's transparently handled by HTML export. Similarly, strong/emph will work correctly if we change those as discussed in https://github.com/typst/typst/issues/6639. Deciding how exactly these intricate changes should be done was the main reason I didn't implement this yet because they can also affect paged export (see https://github.com/typst/typst/issues/5614).

laurmaedje:
See also https://www.w3.org/TR/dpub-aria-1.1/#doc-bibliography

breezykermo:
Great, thank you! Now that #6917 has been merged, I will close this and open a new PR that:
- Follows the structure there to get links via `LinkElem`
- Abides by the spec, i.e. using a `<section>` tag for the outer container, using an `<h2>` for the title.
- Removes the id/class at the outer layer and use the role in the spec.

Issue #6346, opened 2025-05-29
Better documentation of relative lengths: page, region, block, container, element, etc.


shreevatsa:
### Description

Hello, I am trying out Typst and while [the documentation](https://typst.app/docs) has a reasonably good tutorial and reference, and also guides to some extent, I find it lacking in the [“explanation” quadrant](https://web.archive.org/web/20200312220117/https://www.divio.com/blog/documentation/) — I was interested in understanding the typesetting/layout model, and there was too little in the tutorial and most of the reference section seemed more interested in the programming model than basic questions of layout.

A specific example is relative lengths like "100%", which are [documented](https://typst.app/docs/reference/layout/relative/) only as “A length in relation to some known length”, but I have not been able to find in the docs a way of figuring out *what* that known length is, in a given context.

On the [forum](https://forum.typst.app/t/how-to-draw-a-decorative-frame-rect-with-concave-corners-around-text/4306/5), user Andrew pointed me to this section of the source code: https://github.com/typst/typst/blob/2f1865743fc162682d324548da9d385a1686215f/crates/typst-library/src/layout/rel.rs?plain=1#L11-L68 that suggests a few things:

- relative lengths are, in a “common use case”, relative to the “width of the page”, which is “the page's _inner_ width (the width minus margins)”.

- in general they are relative to the “container”, when a “layout element” is nested in “another layout container”.

So there seem to be a bunch of undefined/undocumented terms here, like “ region”, “container”, and “element”. (For example, is there a list of all the elements, or different kinds of regions?)

The most useful has been the “What becomes a paragraph?” at https://typst.app/docs/reference/model/par/ and I'd like more such sections.

The basic thing I'm trying to understand is: in general, what does a page consist of? For example, I understand the TeX model where the page is a vbox containing hboxes (and glue and penalties), and I can also debug this myself or check my understanding in any given situation, with [`\tracingoutput`](https://tex.stackexchange.com/questions/60491/latex-tracing-commands-list) — it seems that the Typst model is more sophisticated, as regular paragraphs and (say) headings are first-class concepts treated differently, so a page can be some combination of different things, and the documentation has not been helpful in understanding what these things are and what exactly is going on in any given context. A concrete test is: can a user who has read the documentation correctly answer, given a location in the Typst source code, what a relative length like "50%" there is going to be relative to?

(Also, IMO a user ought to be able to learn these things from documentation rather than having to read the source code comments, ideally.)

eltos:
[...]

PR #6897, opened 2025-09-11
Improve the docs for relative lengths

YDX-2147483647:
_Original title:  Add links to the docs for relative length_

Continues #5750 and #6862
Relates to #6346

### DIFF
--- a/crates/typst-library/src/layout/container.rs
+++ b/crates/typst-library/src/layout/container.rs
@@ -31,10 +31,7 @@ use crate::visualize::{Paint, Stroke};
 pub struct BoxElem {
     /// The width of the box.
     ///
-    /// This can be a relative length. See the [its documentation]($relative)
-    /// for more details.
-    ///
-    /// Boxes can also have [fractional]($fraction) widths, as the example below
+    /// Boxes can have [fractional]($fraction) widths, as the example below
     /// demonstrates.
     ///
     /// _Note:_ Currently, only boxes and only their widths might be fractionally
@@ -47,9 +44,6 @@ pub struct BoxElem {
     pub width: Sizing,
 
     /// The height of the box.
-    ///
-    /// This can be a relative length. See the [its documentation]($relative)
-    /// for more details.
     pub height: Smart<Rel<Length>>,
 
     /// An amount to shift the box's baseline by.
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/layout/container.rs
```diff
@@ -31,7 +31,10 @@ use crate::visualize::{Paint, Stroke};
 pub struct BoxElem {
     /// The width of the box.
     ///
-    /// Boxes can have [fractional]($fraction) widths, as the example below
+    /// This can be a relative length. See the [its documentation]($relative)
```
laurmaedje: There is a grammatical error in all occurrences of this sentence. But overall, I'm also not sure that it adds a lot, since the "relative" is already listed among the possible values and linked there as well. Of course, a beginner might not realize or understand the relevance of that, but that's perhaps rather something we should help them out with globally (with something like "How to use the reference")

YDX-2147483647: I've fixed the grammatical error.

I think adding this link is helpful even for experienced authors.
The meanings of relative lengths depend on the context, and need to be addressed case by case.
They are sometimes explained at the parameters of each function (e.g., inset/outset) and sometimes on the general page for `relative` (e.g., width/height).
I think the latter case requires special reminder.

laurmaedje: If something special is going on, that's worth calling out. I could even potentially see specifying what it's relative to in each case. But just linking to relative docs each time feels fully redundant to me. We should rather properly teach people to properly read the type hints in the docs.

YDX-2147483647: Okay, I've clear it up. Now there are only two small changes:

- Change _Relative lengths are relative to…_ to _Relative lengths <ins>for this parameter</ins> are relative to…_, telling that the description here does not apply to other parameters.

- Add two links in `rel.rs`.
  Those two sentences were originally right next to [the description above](https://typst.app/docs/reference/layout/relative/#scripting:~:text=This%20type%20is%20a%20combination%20of%20a%20length%20with%20a%20ratio.), so the links were unnecessary then.
  But #5750 separated them. Now we can't find the clickable links after scrolling down to the details, which is somewhat discouraging.

laurmaedje:
Thanks!

PR #6899, opened 2025-09-11
Improve the docs for dictionary

### LINKED ISSUE #6895: Document how are dictionaries added
### Description

I want to explicitly document how are dictionaries added.
Specifically, values in the second dictionary will override values in the first one, but the order of keys is kept.
For example, `(a: 1, b: 2) + (a: 3)` is `(a: 3, b: 2)`, not `(a: 1, b: 2)` or `(b: 2, a: 3)`.
See [`IndexMap::extend` in `indexmap::map` - Rust](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.extend).

Besides, maybe add a link from `dictionary.insert`.

[Discord](https://discord.com/channels/1054443721975922748/1399326777540608041/1415339618923970580):
> Dictionaries guarantee insertion order

At present, it looks like the following sentence is the only documentation. And there's no test for `+` and `join`.
https://github.com/typst/typst/blob/130d0abdc039194d43834cec5cd155c01b2f863b/crates/typst-library/src/foundations/dict.rs#L44-L45

A bit relates to #1258.

YDX-2147483647:
Resolves #6895

- Make it clear that the insertion **order** is the order they were inserted / declared _initially_. Also add a test for it.

- Document that the last value will **override** the others when dictionaries are added/joined/spread together.

- Document the **`(:..spread)`** syntax.

  I put it in a footnote, because it's really rare and I can't image any proper use case — why write `(:..a, ..b)` if you can write `a + b`?

- Change `"Berlin "` to `"Berlin"`.

  The typo dates back to [2023](https://github.com/typst/typst/commit/b471ac7d590abd2398ce25193b4e4df373bf2e9c#diff-3a020c42cc6500f1b1ecf127b5552edbe61c2b03b27b7f6d9890393aed7ab052).

- Add cross links and references.

### DIFF
--- a/crates/typst-library/src/foundations/dict.rs
+++ b/crates/typst-library/src/foundations/dict.rs
@@ -53,10 +53,6 @@ pub use crate::__dict as dict;
 /// another dictionary[^1] with the `..spread` operator. In each case, if a
 /// key appears multiple times, the last value will override the others.
 ///
-/// [^1]: When spreading into a dictionary, if all items between the parentheses
-/// are spread, you have to use the special `(:..spread)` syntax. Otherwise, it
-/// will spread into an array.
-///
 /// # Example
 /// ```example
 /// #let dict = (
@@ -73,6 +69,10 @@ pub use crate::__dict as dict;
 /// #dict.insert("city", "Berlin")
 /// #("name" in dict)
 /// ```
+///
+/// [^1]: When spreading into a dictionary, if all items between the parentheses
+/// are spread, you have to use the special `(:..spread)` syntax. Otherwise, it
+/// will spread into an array.
 #[ty(scope, cast, name = "dictionary")]
 #[derive(Default, Clone, PartialEq)]
 pub struct Dict(Arc<IndexMap<Str, Value, FxBuildHasher>>);


crates/typst-library/src/foundations/dict.rs
```diff
@@ -207,6 +215,9 @@ impl Dict {
 
     /// Inserts a new pair into the dictionary. If the dictionary already
     /// contains this key, the value is updated.
+    ///
+    /// To insert multiple pairs at once, you can just add another dictionary
```
laurmaedje:
```suggestion
    /// To insert multiple pairs at once, you can alternatively add another dictionary
```

crates/typst-library/src/foundations/dict.rs
```diff
@@ -37,17 +37,25 @@ pub use crate::__dict as dict;
...
+/// the `in` keyword.
 ///
 /// You can iterate over the pairs in a dictionary using a [for
 /// loop]($scripting/#loops). This will iterate in the order the pairs were
-/// inserted / declared.
+/// inserted / declared initially.
+///
+/// Dictionaries can be added with the `+` operator and [joined together]($scripting/#blocks).
+/// They can also be [spread]($arguments/#spreading) into a function call or
+/// another dictionary[^1] with the `..spread` operator. In each method, if a
```
laurmaedje: The term "method" could be confused for methods. Perhaps "In each case"?

crates/typst-library/src/foundations/dict.rs
```diff
@@ -37,17 +37,25 @@ pub use crate::__dict as dict;
...
+/// key appears multiple times, the last value will override the others.
+///
+/// [^1]: When spreading into a dictionary, if all items between the parentheses
```
laurmaedje: Maybe the footnote should be moved to the end of the doc comment? Otherwise, it will appear before the example.

YDX-2147483647: I check the [pulldown-cmark spec](https://pulldown-cmark.github.io/pulldown-cmark/specs/footnotes.html), and it turned out to be true. `(@_@)`

laurmaedje:
Thanks!

Issue #6288, opened 2025-05-17
Option to turn off hyperlinking of `cite` elements


aaron-jack-manning:
The [`cite`](https://typst.app/docs/reference/model/cite/) function links to the corresponding entry in the bibliography. Being able to turn this off, with an optional argument to `cite`, would be very useful.

My specific use case is when implementing [this workaround](https://github.com/typst/typst/issues/1097#issuecomment-2575716043) for multiple bibliographies, in which if you use `show bibliography: none` and `cite(<label>, form: "full")`, the citations end up linking to an element of the document which doesn't actually exist in the output.

I should also say that the ability to style links within cite using show rules as per #5614 would provide a workaround to achieve the desired effect.

Enivex:
@SillyFreak bat signal

Andrew15-5:
[A workaround](https://forum.typst.app/t/how-to-turn-off-cite-linking-to-bibliography/4312/4?u=andrew) using https://typst.app/universe/package/alexandria.

jassler:
I want to add another use case where someone failed to upload their paper to an IEEE conference because the publisher wouldn't allow links inside the document.

https://www.reddit.com/r/typst/comments/1u9k1cf/

Their workaround was to call `cpdf` to remove all links.

```
cpdf -remove-annotations main.pdf -o out.pdf
```

PR #6903, opened 2025-09-11
Support for type checking in where selectors

laurmaedje:
This PR extends `.where(..)` selectors to support not just comparisons against a concrete value, but also checks against a specific type. This example demonstrates what I'm talking abou:

<table>
<tr>
<td> Markup </td> <td> Output </td>
</tr>
<tr>
<td>

```typ
#show link.where(dest: str): set text(blue)

= Hello <hello>

#link(<hello>)[Label] \
#link("https://typst.app")[String]
```

</td>
<td>
Image
</td>
</tr>
</table>

## Why?

In the future, when we add custom types, I plan to unify elements and types (hence why custom types will act as custom elements). That means that _selectors,_ which are currently used to select content elements will implicitly become a mechanism to select arbitrary values. If you think about it for a bit, this is pretty similar to what types are (or type patterns if you're thinking about types nominally rather than structurally).

For this reason, I want to go all in on this and unify the concepts of selectors, type hints, and (should we add them) match patterns. Imagine things like this:

```typ
// The pipe operator is the same as `selector.or` right now!
#let f(x: int | str) = ..
#let cite(form: "normal" | "prose" | "full" | "author" | "year", ..) = ..

#show list | enum: ..
#show heading.where(level: 1 | 2): ..
#show link.where(dest: str): ..
```

As you can see, type checking in where selectors would be a pretty natural consequence of these changes.

## Why now?

Why, however, add this now rather than together with the changes presented above? The motivation is specifically the `link.where(type: str)` case. Currently, most elements that internally link somewhere (think `ref`, `cite`, `footnote`, `outline.entry`) do not actually use `link` elements to do this, but rather a compiler-internal mechanism.

While these elements do have some semantical differences from your average URL link, they are links nonetheless and should be treated as such (see also issues https://github.com/typst/typst/issues/5614, https://github.com/typst/typst/issues/6288, and PR https://github.com/typst/typst/pull/5656).

The reason why I've shied away from making this change is that it makes it hard to style normal URL links. If you just have `show link: underline` in your document, your footnotes would suddenly be underlined and the show rule

```typ
#show link: it => if type(it.dest) == str {
  underline(it)
} else {
  it
}
```

is a bit too much to ask of users, especially since underlining a link or turning it blue is such a basic requirement.

With the changes in this PR, we can instead just write

```typ
#show link.where(dest: str): underline
```

This is of course still not quite as simple as just `show link`, but it _is_ simpler and it is something we can teach in the docs. I believe it is a worthwhile trade-off for the benefits we get from using link elements consistently.

It's worth mentioning that moving all the internal elements to `link` also unlocks more styling possibilities, like _avoiding_ linking in certain cases where you don't want it, for example like this:

```typ
#show cite: it => {
  show link: it => it.body
  it
}
```

**Motivation summary:** Making a small incremental change (that will most likely come sooner or later anyway) already now unlocks another useful change now.

## Caveats

There is one design challenge I'd like to discuss. There is some ambiguity in the design: If we use a type in a where selector, we could also have wanted to match a field whose value is exactly that type rather than an instance of it.

For now, I believe this should be an exceedingly rare requirement (the only element I can imagine doing this on is `metadata`, e.g. `metadata.where(value: int)`) and I would just not provide a way to do it.

In the future, when we unify types and elements, and selectors and type hints, I would probably provide a selector/pattern that supports this; something like `pattern.literal(int)`. Then, the function `let f(x: int) = ..` only allows integers, but `let f(x: pattern.literal(int)) = ..` only allows the int type itself. This example also illustrates why the alternative design of letting `value: int` match _both_ integers and the concrete `int` type is a bad idea: You don't want your function that expects to process integers suddenly have the `int` type itself in its hands.

It's also worth mentioning that similar ambiguities exist elsewhere in the language
- for `show heading: f` where we might want to literally display `f` instead of using it as a show rule recipe (as in `show heading: [replacement content]`)
- for `array.contains` where we could also want to search for a function rather than using it to filter
- for `array.find` where `none` can mean no match or we found a `none`

This kind of thing is just hard to avoid in dynamically typed languages ... but I don't think it will be a big problem in practice and, in particular, not in this small initial step.

## Breakage

This is a minor breaking change because a show-where selector with a type as the argument will work differently from before. However, as said above, I'd imagine that such cases are exceedingly rare.


laurmaedje:
Update: There is a bit of a problem because `show figure.where(kind: image)` wouldn't break right away, but in the future when `image` becomes a type instead of a function. Deserves more consideration.

jamesrswift:
Not a precise suggestion, but must it be the same `.where()`? perhaps ```show selector(figure).by-argument-type(kind: image)```. Saves on bringing in new syntax (not that that's a good/bad thing, just effort). `by-argument-type` is a stand-in, I'm not feeling creative presently :P

laurmaedje:
In isolation, it doesn't have to be, but the whole idea with the arguments in where becoming selectors instead of raw values was that it then becomes consistent with the plans for type hints and potentially pattern matching. And in these other places, there is no where "wrapper", so this distinction wouldn't be easily transferred. But the distinction is also necessary there.

eltos:
Reading https://github.com/typst/typst/pull/6982
> Another option would be a synthesized field kind on link that could be used, e.g. `show link.where(kind: "url"): underline`.

I think this would be a more powerful way towards styling different link types easily, because it would allow directly selecting e.g. "footnote", "heading", "figure", "equation" etc. where with the type based selector these would all be `label` and one would have to manually distinguish them in a more complex show rule. With the pipe syntax as "or selector" proposed elsewhere it would still be easy to write show rules that match all or an arbitrary subset of kinds.

```typ
#show link.where(kind: "url"): underline
#show link.where(kind: "heading"): emph
#show link.where(kind: "figure" | "equation"): set text(fill: red)
#show link.where(kind: "footnote"): set text(fill: green)
#show link.where(kind: "cite"): set text(fill: blue)
```

laurmaedje:
@eltos My plan for that is to support hierarchical selectors á la `show footnote > link: set text(fill: green)`.

laurmaedje:
We'll have to revisit the design side here, but I don't want to proceed as-is.

PR #6905, opened 2025-09-11
Accessibility documentation

reknih:
This guide is intended as a companion for the work in #6619. It explains to the user what they must pay attention to in order to create an accessible file.


docs/guides/accessibility.md
```diff
@@ -0,0 +1,227 @@
...
+  Learn how to create accessible documents with Typst. This guide covers semantic markup, reading order, alt text, color contrast, language settings, and PDF/UA compliance to ensure your files work for all readers and assistive technologies.
+---
+
+# Accessibility guide
+
+Making a document accessible means that it can be used and understood by everyone. That not only includes people with permanent or temporary disabilities, but also those with different devices or preferences. To underscore why accessibility is important, consider that people might read your document in more contexts than you expected:
+
+- A user may print the document on paper
+- A user may read your document on a phone, with reflow in their PDF reader enabled
+- A user may have their computer read the document back to you
```
MDLC01:
```suggestion
- A user may have their computer read the document back to them
```

Andrew15-5:
Do you actually supposed to write "show set" rule? I always write "show-set" so that it looks as one rule, and not like you forgot to add a comma between 2 rules.

I just checked the repo and there are 27 instances of "show-set" and only 1 instance of "show set".

```sh
rg -g '!tests' --stats show-set
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+// ✅ Do this
+#show heading: set text(size: 16pt)
+= Heading
+```
+
+Both of these examples look the same. They both contain the text "Heading" in boldface, sized at 16 point. However, only the second example is accessible. By using the heading markup, Typst knows that the semantic meaning of this text is that of a heading and can propagate that information to the final PDF. In the first example, it just knows that it should use boldface and larger type on otherwise normal text and cannot make the assumption that you meant that to be a heading and not a stylistic choice or some other element like a quote.
+
+Using semantics is not limited to headings. Here are a few more examples for elements you should use:
+
+- Use underscores / [`emph`] instead of the text function to make text emphasized
```
bastienvoirin:
```suggestion
- Use underscores / [`emph`] instead of the [`text`] function to make text emphasized
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+- Use stars / [`strong`] instead of the text function to make text carry strong emphasis
```
bastienvoirin, line 55:
```suggestion
- Use stars / [`strong`] instead of the [`text`] function to make text carry strong emphasis
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+---
+
+# Accessibility guide
+
+Making a document accessible means that it can be used and understood by everyone. That not only includes people with permanent or temporary disabilities, but also those with different devices or preferences. To underscore why accessibility is important, consider that people might read your document in more contexts than you expected:
+
+- A user may print the document on paper
+- A user may read your document on a phone, with reflow in their PDF reader enabled
+- A user may have their computer read the document back to you
+- A user may ask an Artificial Intelligence to summarize your document to them
```
NiklasEi:
```suggestion
- A user may ask an Artificial Intelligence to summarize your document for them
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+- A user may convert their document to another file format like HTML that is more accessible to them
```
NiklasEi:
```suggestion
- A user may convert your document to another file format like HTML that is more accessible to them
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+To accommodate all of these people and scenarios, you should design your document for **Universal Access.** Universal Access is a simple but powerful principle: instead of retrofitting a project for accessibility after the fact, design from the beginning to work for the broadest possible range of people and situations. This will improve the experience for all readers!
+
+Typst can help you to create accessible files that read well on screen readers, look good even when reflowed for a different screen size, and pass automated accessibility checkers. However, to create accessible files, you will have to keep some rules in mind. This guide will help you to learn what issues impact accessibility, how to design for Universal Access, and what tools Typst gives you to accomplish this. Much of the guidance here applies to all export targets, but the guide focusses on PDF export. Notable differences to HTML are called out.
```
NiklasEi:
```suggestion
Typst can help you to create accessible files that read well on screen readers, look good even when reflowed for a different screen size, and pass automated accessibility checkers. However, to create accessible files, you will have to keep some rules in mind. This guide will help you learn what issues impact accessibility, how to design for Universal Access, and what tools Typst gives you to accomplish this. Much of the guidance here applies to all export targets, but the guide focuses on PDF export. Notable differences to HTML are called out.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+Accessibility support differs based on viewer and AT. Some combinations work better than others. In our testing, Adobe [Acrobat][Acrobat] paired with [NVDA][NVDA] on Windows and [VoiceOver][VoiceOver] on macOS provided the richest accessibility support.
+
+## Maintaining semantics
+
+To add correct semantic information for AT and repurposing to a file, Typst needs to know what semantic part each part of the file plays. For example, this means that a heading in a compiled PDF should not just be text that is large and bold, instead, the file should contain the explicit information (known as a _tag_) that a particular text makes up a heading. A screen reader will then announce it as a heading and allow the user to navigate between headings.
```
NiklasEi:
```suggestion
To add correct semantic information for AT and repurposing to a file, Typst needs to know what semantic role each part of the file plays. For example, this means that a heading in a compiled PDF should not just be text that is large and bold, instead, the file should contain the explicit information (known as a _tag_) that a particular text makes up a heading. A screen reader will then announce it as a heading and allow the user to navigate between headings.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+If you want to style the default appearance of an element, do not replace it with your own custom function. Instead, use [set]($styling/#set-rules), show set, and [show rules]($styling/#show-rules) to customize its appearance. Here is an example on how you can change how strong emphasis looks in your document:
+
+```example
+// Change how text inside of strong emphasis looks
+#show strong: set text(tracking: 0.2em, fill: blue, weight: "black")
+When setting up your tents, *never forget* to secure the pegs
+```
+
+The show set rule completely changes the default appearance of the strong element, but its semantic meaning will be conserved. If you need even more customization, you can provide show rules with fully custom layouting code while Typst will still be able to track the semantic purpose of the element.
```
NiklasEi:
```suggestion
The show set rule completely changes the default appearance of the [`strong`] element, but its semantic meaning will be conserved. If you need even more customization, you can provide show rules with fully custom layouting code while Typst will still be able to track the semantic purpose of the element.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Reading order
+
+For AT to read the contents of a document in the right order and for repurposing applications, accessible files must make their reading order explicit. This is because the logical reading order can differ from layout order. Floating figures are a common example for such a difference: A figure may be relevant to a paragraph in the center of a page but appear at the top or bottom edge. In non-accessible files, PDF readers and AT have to assume that layout order equals the logical reading order, often leading to confusion for AT users. Instead, when reading order is defined screen readers read a footnote or a floating figure immediately where it makes sense.
```
NiklasEi:
```suggestion
For AT to read the contents of a document in the right order and for repurposing applications, accessible files must make their reading order explicit. This is because the logical reading order can differ from layout order. Floating figures are a common example for such a difference: A figure may be relevant to a paragraph in the center of a page but appear at the top or bottom edge. In non-accessible files, PDF readers and AT have to assume that layout order equals the logical reading order, often leading to confusion for AT users. When the reading order is defined, screen readers read a footnote or a floating figure immediately where it makes sense.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Layout containers
+
+Typst provides some layout containers like [`grid`], [`stack`], [`box`], [`columns`], and [`block`] to visually arrange your data. None of these containers come with any semantic meaning attached. Typst will conserve some of these containers (such as columns) during PDF reflow while other containers will be discarded.
+
+When designing for Universal Access, you need to be aware that AT users often cannot view the visual layout that the container creates. Instead, AT will just read its contents, so it is best to think about these containers as transparent in terms of accessibility. For example, a grid will just be announced cell by cell, in the order that you have added cells in the source code. If there layout you created is merely visual and decorative, this is fine. If, however, the layout carries semantic meaning that is apparent to a sighted user viewing the file in a regular PDF reader, it is not accessible. Instead, create an alternative representation of your content that leverages text or wrap your container in the [`figure`] element to provide an alternative textual description.
```
NiklasEi:
```suggestion
When designing for Universal Access, you need to be aware that AT users often cannot view the visual layout that the container creates. Instead, AT will just read its contents, so it is best to think about these containers as transparent in terms of accessibility. For example, a grid will just be announced cell by cell, in the order that you have added cells in the source code. If the layout you created is merely visual and decorative, this is fine. If, however, the layout carries semantic meaning that is apparent to a sighted user viewing the file in a regular PDF reader, it is not accessible. Instead, create an alternative representation of your content that leverages text or wrap your container in the [`figure`] element to provide an alternative textual description.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Textual representations
+
+To support AT use and some repurposing workflows, all elements with a semantic meaning must have a textual representation. Think about it in terms of Universal Access: If an item is not an artifact, it has a semantic meaning. If, however, AT cannot ingest the item, the full semantic meaning of a document is not available to AT users. Hence, to provide Universal Access, use the mechanisms built into Typst to provide alternative representations.
+
+When you add an image, be sure to use the [`alt` argument of the image function]($image.alt) to describe what's visible in the image. This alternative description (sometimes known as alt text) should describe the gist of the image: Think about you would describe the image to a friend if you called them on the phone. There are resources available on the web [to learn more about writing good alternative descriptions][alt-text-tips]. The requirement to add alternative text to images applies to PDF and SVG images. Typst does not currently mount the tags of a PDF image into the compiled document, even if the PDF image file on its own was accessible.
```
NiklasEi:
```suggestion
When you add an image, be sure to use the [`alt` argument of the image function]($image.alt) to describe what's visible in the image. This alternative description (sometimes known as alt text) should describe the gist of the image: Think about how you would describe the image to a friend if you called them on the phone. There are resources available on the web [to learn more about writing good alternative descriptions][alt-text-tips]. The requirement to add alternative text to images applies to PDF and SVG images. Typst does not currently mount the tags of a PDF image into the compiled document, even if the PDF image file on its own was accessible.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+Like the image function, the figure function has a [`alt` attribute]($figure.alt). When you use this attribute, many screen readers and other AT will not announce the content inside of the figure and instead just read the alternative description. Your alternative description must be comprehensive enough so that the AT user does not need to access the children of the figure. Only use the alternative description if the content of the figure are not otherwise accessible. For example, do not use the `alt` attribute of a figure if it contains a `table` element, but do use it if you used shapes within that come with a semantic meaning. If your figure contains an image, it suffices to set an alternative description on the image.
```
NiklasEi:
```suggestion
Like the image function, the figure function has a [`alt` attribute]($figure.alt). When you use this attribute, many screen readers and other AT will not announce the content inside of the figure and instead just read the alternative description. Your alternative description must be comprehensive enough so that the AT user does not need to access the children of the figure. Only use the alternative description if the content of the figure are not otherwise accessible. For example, do not use the `alt` attribute of a figure if it contains a `table` element, but do use it if you used shapes in it that come with a semantic meaning. If your figure contains an image, it suffices to set an alternative description on the image.
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+- **PDF/A-2a** and **PDF/A-3a:** The PDF/A standard describe how to produce PDF files that are best suited for archival. Parts two and three of the PDF/A standard feature multiple conformance levels. The strictest conformance level A contains rules for accessibility as only they remain usable to the broadest range of people in the far future. Level A implies conformance with Tagged PDF, forces you to provide descriptions text for images, and disallows the use of characters in the Unicode Private Use Area whose meaning is unclear to the general public. Other PDF/A rules not relating to accessibility, e.g. about colors and more also apply. When choosing between the two standards, choose PDF/A-2a unless you need to attach other PDF files. Conformance level A has been removed from PDF/A-4 in favor of the dedicated PDF/UA standard. When targeting PDF 2.0, use PDF/A-4 together with PDF/UA-2 instead (the latter is not yet supported by Typst).
```
NiklasEi:
```suggestion
- **PDF/A-2a** and **PDF/A-3a:** The PDF/A standard describe how to produce PDF files that are best suited for archival. Parts two and three of the PDF/A standard feature multiple conformance levels. The strictest conformance level A contains rules for accessibility as only they remain usable to the broadest range of people in the far future. Level A implies conformance with Tagged PDF, forces you to provide description texts for images, and disallows the use of characters in the Unicode Private Use Area whose meaning is unclear to the general public. Other PDF/A rules not relating to accessibility, e.g. about colors and more also apply. When choosing between the two standards, choose PDF/A-2a unless you need to attach other PDF files. Conformance level A has been removed from PDF/A-4 in favor of the dedicated PDF/UA standard. When targeting PDF 2.0, use PDF/A-4 together with PDF/UA-2 instead (the latter is not yet supported by Typst).
```

docs/guides/accessibility.md
```diff
@@ -0,0 +1,233 @@
...
+
+- Use underscores / [`emph`] instead of the text function to make text emphasized
+- Use stars / [`strong`] instead of the text function to make text carry strong emphasis
+- Use lists, including [`terms`], instead of normal text with newlines when working with itemized or ordered content.
+- Use [`quote`] for inline and block quotes
+- Use the built-in [`bibliography`] and [`cite`] functions instead of manually printing a bibliography
+- Use labels and [`ref`] or `@references` to reference other parts of your documents instead of just typing out a reference
+- Use the [`caption` argument of the `figure` element]($figure.caption) to provide captions instead of adding them as text below the function call
+
+If you want to style the default appearance of an element, do not replace it with your own custom function. Instead, use [set]($styling/#set-rules), show set, and [show rules]($styling/#show-rules) to customize its appearance. Here is an example on how you can change how strong emphasis looks in your document:
```
Andrew15-5:
```suggestion
If you want to style the default appearance of an element, do not replace it with your own custom function. Instead, use [set]($styling/#set-rules), show-set, and [show rules]($styling/#show-rules) to customize its appearance. Here is an example on how you can change how strong emphasis looks in your document:
```

docs/guides/accessibility.md
Andrew15-5:
```suggestion
The show-set rule completely changes the default appearance of the strong element, but its semantic meaning will be conserved. If you need even more customization, you can provide show rules with fully custom layouting code while Typst will still be able to track the semantic purpose of the element.
```

docs/guides/accessibility.md
reknih: I only link the first instance of a function in a paragraph.

docs/guides/accessibility.md
```diff
@@ -103,16 +105,19 @@ Universal Access not only means that your documents works with AT, reflow, and r
 
 This means that color must not be the only way you make information accessible to sighted users in your documents. As an example, consider a stacked bar chart with multiple colored segments per bar. For a color blind user, there are multiple challenges here. The first is to associate the color of a bar segment with its entry in the legend. You could help the reader by always ensuring that the colors appear in the same order as they do in the legend and note this. The next challenge is to identify the boundaries of the colored segments in a single bar. To make this easier, you could draw a dark or light boundry with high contrast. Alternatively, you could address both problems by also giving each segment a unique pattern in addition to a color.
```
hpcfzl: One more `boundry` -> `boundary` typo here.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+# Accessibility guide
+
+Making a document accessible means that it can be used and understood by everyone. That not only includes people with permanent or temporary disabilities, but also those with different devices or preferences. To underscore why accessibility is important, consider that people might read your document in more contexts than you expected:
+
+- A user may print the document on paper
+- A user may read your document on a phone, with reflow in their PDF reader enabled
+- A user may have their computer read the document back to them
+- A user may ask an Artificial Intelligence to summarize your document for them
+- A user may convert your document to another file format like HTML that is more accessible to them
```
hpcfzl, line 14: Also is this level of repeating at the start necessary? I observe that it's visually slightly too noisy, so I think putting `A user may` before the entire list might be better. Same with repeated `Use` in items lower down. Though I haven't checked if it's done like so elsewhere in the docs.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+---
+
+# Accessibility guide
+
+Making a document accessible means that it can be used and understood by everyone. That not only includes people with permanent or temporary disabilities, but also those with different devices or preferences. To underscore why accessibility is important, consider that people might read your document in more contexts than you expected:
+
+- A user may print the document on paper
+- A user may read your document on a phone, with reflow in their PDF reader enabled
+- A user may have their computer read the document back to them
+- A user may ask an Artificial Intelligence to summarize your document for them
```
hpcfzl: Here, I don't think `Artificial Intelligence` is supposed to be capitalised.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+This information is consumed by different software to provide access. When exporting a PDF from Typst, the _PDF viewer_ (sometimes also called a reader) will display the document's pages just as you designed them with Typst's preview. Some people rely on _Assistive Technologies_ (AT) such as screen readers, braille displays, screen magnifiers, and more for consuming PDF files. In that case, the semantic information in the file is used to adapt the contents of a file into spoken or written text, or into a different visual representation. Other users will make the PDF viewer reflow the file to create a layout similar to a web page: The content will fit the viewport's width and scroll continuously. Finally, some users will repurpose the PDF into another format, for example plain text for ingestion into a Large Language Model (LLM) or HTML. A special form of repurposing is copy and paste where users use the clipboard to extract content from a file to use it in another application.
```
hpcfzl: I would put `Large Language Model` in italic so it's consistent with `_Assistive Technologies_`.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+Some things on a page have no semantic meaning and are irrelevant to the content of a document. We call these items _artifacts._ Artifacts are hidden from AT and repurposing and will vanish during reflow. Here are some examples for artifacts:
+
+- The hyphens inserted by automatic hyphenation at the end of a line
+- The headers and footers on each page
+- A purely decorative page background image
+
+In general, every element on a page must either have some way for AT to announce it or be an artifact for a document to be considered accessible.
+
+Typst automatically tags many layout artifacts such as headers, footers, page back- and foregrounds, and automatic hyphenation as artifacts. However, if you'd like to add purely decorative content to your document, you can use the `pdf.artifact` function to mark a piece of content as an artifact. If you are unsure if you should mark an element as an artifact, ask yourself this: Would it be purely annoying if a screen reader announced the element to you? Then, it may be an artifact. If, instead, it could be useful to have it announced, then it is not an artifact. <!-- TODO: Link once it exists -->
```
hpcfzl: Here at the end, I might wonder whether it's possible to achieve the reverse of `pdf.artifact`. So to for sure force would-be artifacts in some places to be announced. If not, which I assume is the case, an explanation could be given on why that's a bad idea.

And the phrasing might be better with `likely`, as in `... then it is likely not an artifact.`.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+Please note that Typst will mark shapes and paths like [`square`] and [`circle`] as artifacts while their content will remain semantically relevant and accessible to AT. If your shapes have a semantic meaning, please wrap them in the [`figure`] element to provide an alternative textual description.
+
+## Color use and contrast
+
+Universal Access not only means that your documents works with AT, reflow, and repurposing, but also that visual access is possible to everyone, including people with impaired eye sight. Not only does aging often come with worse sight, a significant chunk of people have problems differentiating color: About 8% of men and 0.5% of women are color blind.
```
hpcfzl: I would change `eye sight` -> `eyesight`. And `color blind` -> `colorblind` in where it's an adjective.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+  ❌ The alternative description should not include details not visible in the image, such as attribution, jokes, or metadata. Keep in mind that it is not accessible to sighted users. That information belongs elsewhere.
+
+- `["Gray heron flying low, heading from the right to left. Its feet are extended and slightly point downwards, touching a blurred horizon where a dark forest becomes visible. The bird's wings are extended and arc upwards. There are out-of-focus branches visible in the lower left corner of the image."]`: \
+  ❌ The alternative description is too verbose. Use your discretion and determine how important the image is to the content. Think about how long a sighted user would realistically look at the image; your alt text should take about the same effort to 'consume.' For example, the anatomic description contained above could be appropriate for a longer discussion in a zooology textbook while the compositional information is useful when writing about photography. The context the example image comes with is relatively short, so write a more brief description.
```
hpcfzl: Rather `zoology`?

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+Instead, in the given example, you could use this alternative text:
+
+`["Heron in flight with feet and wings spread"]` \
+✅ This alternative description describes the image, is relevant to the context, and matches its brevity.
+
+There are more resources available on the web [to learn more about writing good alternative descriptions][alt-text-tips]. The requirement to add alternative text to images applies to PDF and SVG images. Typst does not currently mount the tags of a PDF image into the compiled document, even if the PDF image file on its own was accessible.
```
hpcfzl: > The requirement to add alternative text to images applies to PDF and SVG images.

Why not PNG as well? I do remember something about PNGs not being accessible earlier on, but SVGs were mentioned there too.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+Do not use images of text, likewise, do not use the path operations to draw text manually. Typst will not be able to process text in images to make it accessible in the same way that native text is. The only exceptions to this rule are images in which the appearance of the text is essential to the semantic meaning of the document and cannot be reproduced with Typst natively. In that case, you must describe both the text content and the essential visual characteristics in the alternative description.
```
hpcfzl: Would `Do not use any images of text` be more sensible?

I would also change `text content` -> `textual content`.

So a TLDR that I get from the paragraph would be: all images are treated the same by AT, even if there's selectable text within.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Natural Language
+
+In order for screen readers to pronounce your document correctly and translation software to work properly, you must indicate in which natural language your document is written. Use the rule [`[#set text(lang: "..")]`]($text.lang) at the very start of your document or your template's capability to set a language. If you do not do so, Typst will assume that your content is in English. The natural language you choose not only impacts accessibility, but also how Typst will apply hyphenation, what text layout conventions are applied, the labels of figures and references, and, in the web app, what language is used for spellcheck.
```
hpcfzl: Changing `template's capability` -> `template's extent` might sound better?

Quotes could also use a mention in that list: `how Typst will apply hyphenation, quotation ...`.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+
+To do so in Typst, place this set rule in your document before any content:
+
+```typ
+#set document(title: "GlorboCorp Q1 2023 Revenue Report")
+```
+
+This will set the [title in the document's metadata]($document.title) and in the title bar of the PDF viewer. If this results in an error when using a template, consider whether your template may provide an alternative way to set the document title.
+
+Most likely, you will also want to include the title in your document. To do so, use the [`title`] element. When you add a call to the title element without any arguments, it will print the contents of the document's title. Alternatively, you can customize the title by passing content as the positional body argument. Do not use the title element more than once in your document.
```
hpcfzl: Since you mention HTML when discussing headings, it would be useful to also know here which title is the one that will cause it to appear in HTML page's tab; is it the `title` option of `document` or is it the `title` element?

And to avoid confusion when skimming through, I think it's fair to add `content`, as in `... include the title as part of content in your document.`.

hpcfzl: Oh, so the HTML heading is what comes from `document`'s `title`. I think it's still missing a piece about how to get a different HTML title from Typst, just based off [this very early comment](https://github.com/typst/typst/issues/721#issuecomment-1503405488).

docs/guides/accessibility.md
```diff
@@ -0,0 +1,304 @@
...
+// ❌ Don't do this:
+= First level heading
+=== Third level heading
+```
+
+Note that in order to pass the [automated accessibility check in Adobe Acrobat][acro-check-outline], documents with 21 pages or more must contain outlined headings.
+
+## Accessibility Standards and Legislation
+
+There are international standards that help you to assert that a Typst document is accessible. For PDF export, Typst can check whether you are complying with PDF-based accessibility standards and assert the compliance in the compiled file:
```
hpcfzl: Is the checking a Typst compiler feature or a webapp-only feature? I would assume the former here. Still, it would be helpful to know how this is presented to the user. I find out a few paragraphs later that it will be a descriptive error message. I just have a feeling we don't get to know much about how this checking can be controlled in Typst, like are there new compiler options that come specifically for these new accessibility features? We get to know what combinations of standards result in more and less strict checks, but not how to actually do so. The existing top paragraph of `Testing for Accessibility` could use a quick mention about Typst's interface to this. I assume it's just using the export option in different ways, but I can't be the only person confused.

docs/guides/accessibility.md
LaurenzV: I'm also not sure we should even mention PDF images here since they are not supported in PDF/UA1.

reknih:
> What I possibly missed was the most commonly misused feature that is about [the proper use of links in text](https://example.com/) [...]

GitHub normalized both of your links into the same syntax. If this is about adding the URI schema (i.e. `https://example.com` instead of `example.com`), I don't see how this is an accessibility issue, specifically. Instead, the strong preference for an URL schema should be described in the Link docs, perhaps even adding a warning if no schema is present.

hpcfzl:
No, that was on purpose, sorry I didn't make it clear enough. I don't think I'll be able to find the right term for this, but I was thinking of _inline links_. So seeing the words highlighted as a link instead of pasting one long link. What many people often do is use the highlight simply on a deictic like _here_ instead of indicating where the link leads to. And as said I don't know if it makes a difference to screen readers, but I think this first sentence

> Visit my website on [Example Domain](https://example.com/).

is much more helpful than this one using _here_

> Visit my website [here](https://example.com/).

As for the URI schema, I agree that part would better suit `link` docs.

docs/guides/accessibility.md
reknih: I'm always using Artificial Intelligence as a proper noun and not a "normal" term and thus capitalizing it. (re: nature of intelligence and all that) Is that not the common spelling anymore?

docs/guides/accessibility.md
reknih: I think it is way less important to the guide and thus deliberately not emphasized. If there is one thing the user should take away from that section, it is a) what Assistive Technologies are and b) that I abbreviate it as AT, as that's used throughout the whole guide.

docs/guides/accessibility.md
reknih: Why would you want to hide relevant content from AT users using `pdf.artifact`? I cannot conceive of any examples of a case where that ass-covering might be necessary.

docs/guides/accessibility.md
reknih: The point is that sometimes you need to use an image to show off how a text appears. For example, when you want to compare the results of a Linotype and Phototypesetting, having two images of text is necessary. 

Also see https://www.w3.org/WAI/WCAG21/Understanding/images-of-text.html

docs/guides/accessibility.md
reknih: > Changing template's capability -> template's extent might sound better?

I think that does not make sense - this is about the template providing a mechanism to set the natural language.

> Quotes could also use a mention in that list: how Typst will apply hyphenation, quotation ....

I changed text layout conventions to typesetting conventions to cover quotes.

docs/guides/accessibility.md
reknih:
```typ
#set document(title: "GlorboCorp Q2") // HTML: <title>GlorboCorp Q2</title>, PDF: Document Info dictionary and XMP
#title[GlorboCorp Q2 Report] // HTML: <h1>GlorboCorp Q2 Report</h1>, PDF 1.7 and below: `/P` tag, PDF 2.0: `/Title` tag, shows up in document for all formats
= Heading 1 // HTML: <h2>Heading 1</h2>, PDF: `/H1` tag, shows up in document for all formats
```

I added a litte something to cover HTML.

docs/guides/accessibility.md
reknih: There will just be additions to the existing `--pdf-standards` flag and the PDF export modal in the web app, as well as a flag to disable tagging altogether. CLI docs and web app docs for these options should be created separately and then linked here.

reknih:
> What many people often do is use the highlight simply on a deictic like _here_ instead of indicating where the link leads to. 

That is actually okay in WCAG Level A: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html
Only Level AAA would require the link on its own to be descriptive of its target: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html

Typst also adds a tooltip (`/Contents` key of the annotation, can be compared with the HTML `title` attribute on `a`) containing the target URL automatically.

Where would you add this guidance to the guide?

docs/guides/accessibility.md
hpcfzl: Meanwhile I haven't seen it as a proper noun anywhere else :) Searching further, it's all lowercase on [Wikipedia](https://en.wikipedia.org/wiki/Artificial_intelligence) and in [dictionaries](https://www.merriam-webster.com/dictionary/artificial%20intelligence). In that case, the `an` should also be dropped and just leave it as `artificial intelligence`.

docs/guides/accessibility.md
hpcfzl: Not to hide content, but to prevent what might be relevant content from automatically getting hidden because it's considered as an artifact by default. I don't know either, I just wanted to point out that you can expect such a question to arise from the reader. Something in the lines of

> If, instead, it could be useful to have it announced, it is likely not an artifact. You cannot however guarantee that it won't turn into an artifact.

docs/guides/accessibility.md
hpcfzl: > Typst will not be able to process text in images to make it accessible in the same way that native text is.

Then I would move the `any` before `images` here. Because of `to process text`, my first thougths were text only in SVG and PDF, not also in bitmap images.

---

Also `The only exceptions to this` throws me off, as if Typst would be able to process text for these exceptions. Perhaps it needs rephrasing?

> Images containing any form of text are no different from ordinary images. If the appearance of text in such an image is essential to the semantic meaning of the document and cannot be reproduced with Typst natively, it is necessary to describe the textual content from within the image in addition to its essential visual characteristics, both in the same alternative description.

docs/guides/accessibility.md
hpcfzl: Oh, now I see what you meant:

> Use the rule ... or your template's capability to set a language.

and not

> Use the rule at the very start of your document or [at the very start] of your template's capability to set a language.

Could it be rephrased to

> To set a language, use the rule ... or **see** your template's capability.

hpcfzl:
> Where would you add this guidance to the guide?

I was thinking of adding a paragraph right below the first one of `Textual representations`, as the rest of that section is about images. Either this or, if it's going to be shorter, anywhere as an example from WCAG.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,383 @@
...
+
+- The hyphens inserted by automatic hyphenation at the end of a line
+- The headers and footers on each page
+- A purely decorative page background image
+
+In general, every element on a page must either have some way for AT to announce it or be an artifact for a document to be considered accessible.
+
+Typst automatically tags many layout artifacts such as headers, footers, page back- and foregrounds, and automatic hyphenation as artifacts. However, if you'd like to add purely decorative content to your document, you can use the [`pdf.artifact`] function to mark a piece of content as an artifact. If you are unsure if you should mark an element as an artifact, ask yourself this: Would it be purely annoying if a screen reader announced the element to you? Then, it may be an artifact. If, instead, it could be useful to have it announced, then it is not an artifact.
+
+For technical reasons, once you are in an artifact, you cannot become semantic content ingested by AI again. To stack artifacts and semantic contents, use [`place`] to move the content on top of one another.
```
saecki: Was this supposed to say AT?
```suggestion
For technical reasons, once you are in an artifact, you cannot become semantic content ingested by AT again. To stack artifacts and semantic contents, use [`place`] to move the content on top of one another.
```

reknih: yes

crates/typst-cli/src/args.rs
jassielof: Typo

docs/guides/accessibility.md
```diff
@@ -0,0 +1,363 @@
...
+
+The [PDF reference page]($pdf/#pdf-standards) contains more information about each of these standards. To enable either PDF/UA, PDF/A-2a, or PDF/A-3a, use the export dropdown in the web app and click on PDF or the [appropriate CLI flag]($pdf/#command-line) or web app setting.
+
+There are also a more recent part of the PDF/UA standard that targets PDF 2.0 files, PDF/UA-2. Support for PDF/UA-2 not yet available in Typst. [Both parts of the PDF/UA specification are available free of charge from the PDF Association.][pdf-ua-free] The industry standard [Well Tagged PDF (WTPDF)][WTPDF] is very similar to PDF/UA-2. All files conforming to WTPDF can also declare conformance with PDF/UA-2.
```
laurmaedje: This paragraph's purpose is a bit confused.

docs/guides/accessibility.md
```diff
@@ -0,0 +1,363 @@
...
+
+- **PDF/A-2a** and **PDF/A-3a:** The PDF/A standard describe how to produce PDF files that are best suited for archival. Parts two and three of the PDF/A standard feature multiple conformance levels. The strictest conformance level A contains rules for accessibility as only they remain usable to the broadest range of people in the far future. Level A implies conformance with Tagged PDF, forces you to provide alternative description for images, and disallows the use of characters in the [Unicode Private Use Area][unic-pua] whose meaning is unclear to the general public. Other PDF/A rules not relating to accessibility, e.g. about colors and more also apply. When choosing between the two standards, choose PDF/A-2a unless you need to attach other PDF files. Conformance level A has been removed from PDF/A-4 in favor of the dedicated PDF/UA standard. When targeting PDF 2.0, use PDF/A-4 together with PDF/UA-2 instead (PDF/UA-2 and combining using multiple PDF standards not yet supported by Typst).
```
laurmaedje: The last sentence is broken and perhaps obsolete as this is covered by a following paragraph?

docs/reference/library/math.md
```diff
@@ -99,3 +99,19 @@ $ sum_(i in NN) 1 + i $
 All math functions are part of the `math` [module]($scripting/#modules), which
 is available by default in equations. Outside of equations, they can be accessed
 with the `math.` prefix.
+
+# Accessibility
+To make math accessible, you must provide an alternative description for the
+equation that describes it in natural language. In the future, Typst will
+automatically make equations without alternative descriptions in HTML and PDF
+2.0 export accessible. For more information, see the documentation on the [`alt`
+parameter of `math.equation`]($math.equation.alt) and the [Textual
```
laurmaedje: there is no further information (basically no information) on the `alt` parameter

docs/guides/accessibility.md
```diff
@@ -0,0 +1,369 @@
...
+
+- `{"Gray heron flying low, heading from the right to left. Its feet are extended and slightly point downwards, touching a blurred horizon where a dark forest becomes visible. The bird's wings are extended and arc upwards. There are out-of-focus branches visible in the lower left corner of the image."}` \
+  ❌ The alternative description is too verbose. Use your discretion and determine how important the image is to the content. Think about how long a sighted user would realistically look at the image; your alt text should take about the same effort to 'consume.' For example, the anatomic description contained above could be appropriate for a longer discussion in a zoology textbook while the compositional information is useful when writing about photography. The context the example image comes with is relatively short, so write a more brief description.
+
+Instead, in the given example, you could use this alternative text:
+
+`{"Heron in flight with feet and wings spread"}` \
```
laurmaedje: I think also putting this into a list item makes it visually _much_ clearer. (Would this be a misuse of a list element or could this be considered a continuation of the previous list? 😅)

Issue #6844, opened 2025-08-31
Create a dedicated section on dangers of using square brackets carelessly (multi-line)


Andrew15-5:
### Description

This is one of the most reoccurring things that I absolutely hate for many reasons. I don't think there is a dedicated issue about this, but I mentioned it many times already, but it's all mostly lost as I don't keep track of everything. I created this to be able to link to _somewhere_, until the documentation has the requested section.

As seen in https://github.com/typst/typst/issues/6273, or https://forum.typst.app/t/how-does-one-place-content-relative-to-the-page-margin/4373, or https://forum.typst.app/t/how-to-properly-justify-text-in-a-figure-caption/5822/2, using square brackets where you (only) use code inside (instead of using curly braces), or for things that should be an inline no-leading-trailing whitespaces content — these are one of the most common use cases where square brackets are used carelessly. More time than not such carelessness _will_ affect your, or users' of your package/template document! It happened to me multiple times, and I also was using

```typ
#func()[
  ...
]
```

_a lot_ in the early days (partially because this way things were better formatted with typstfmt, rather than using curly braces). Now I learned my lessons and become very sensitive to this, so I rarely use it, only where necessary or knowing the consequences of it (also with typstyle now formatting everything pretty nicely, and having text wrapping, bad source formatting is no longer an issue).

So, the actually dangerous part is using square brackets in "multi-line mode":

```typ
#func[
  block-like syntax is multiline
]
#func[
this is also multiline]
#func[this as well
// That's right.
]
```

And the _actually-actually_ dangerous part is the spaces right after `[` and right before `]`. In case of caption, or other space-sensitive context, using multi-line square brackets will inject these invisible spaces into your document, potentially altering its layout. A non-multi-line example, but still relevant is https://github.com/pammacdotnet/IJIMAI/pull/3/commits/4b0adabfd27cd33f54bc7b5b8e3ee19dd1e71236, which fixes a double whitespace in table caption. Here, a more clearly visible whitespace is inserted. Some most definitely would think that it doesn't matter, yet it does! At least if you care about the typographical quality of your documents. (I do!)

Newlines also create `space`s (`[ ].func()`), as well as their combination with comments (multiple whitespaces). This is why, for example, dumping a bunch of (or even a single) set/show-set/etc. rules in something that is supposed to be included inline into some other document structure (that may or may not handle the trailing/leading whitespaces) is a bad idea. It's totally normal if you write one or multiply paragraphs right after that (scoped rules):

```typ
#[
[...]

PR #6907, opened 2025-09-12
Disambiguate names in the doc examples for `state`

YDX-2147483647:
Resolve the confusion in [Using state for Caching in Typst — Is It Possible? - Questions - Typst Forum](https://forum.typst.app/t/using-state-for-caching-in-typst-is-it-possible/5961), where the author thought the placeholder for `eval` and the key for `state` are the same.

> I tried playing around with `state`, but I’m struggling to understand how to update it properly. The documentation mentions using `eval`, which I assume is because accessing a state value gives you its content, not a reference…

|                                 | Before | After  |
| ------------------------------- | ------ | ------ |
| variable                        | `x`/`s` | `star` |
| key for `state`                 | x      | star   |
| argument of the update function | `x`    | `old`  |
| placeholder for `eval`          | x      | ⭐      |

In addition, I change `{ … }` to `[ … ]` for the reason in #6844.

~~Preview: https://ydx-typst.netlify.app/reference/introspection/state/~~ (Outdated)

---

Besides, this PR doesn't add new simpler examples, as #6674 already added one in the `state.update` section.

### DIFF
--- a/crates/typst-library/src/introspection/state.rs
+++ b/crates/typst-library/src/introspection/state.rs
@@ -23,18 +23,18 @@ use crate::routines::Routines;
 ///
 /// ```typ
 /// // This doesn't work!
-/// #let x = 0
+/// #let star = 0
 /// #let compute(expr) = {
-///   x = eval(
-///     expr.replace("x", str(x))
+///   star = eval(
+///     expr.replace("⭐", str(star))
 ///   )
-///   [New value is #x. ]
+///   [New value is #star.]
 /// }
 ///
 /// #compute("10") \
-/// #compute("x + 3") \
-/// #compute("x * 2") \
-/// #compute("x - 5")
+/// #compute("⭐ + 3") \
+/// #compute("⭐ * 2") \
+/// #compute("⭐ - 5")
 /// ```
 ///
 /// # State and document markup { #state-and-markup }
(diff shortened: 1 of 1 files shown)


laurmaedje:
I feel like the `star` / ⭐ combo does not really resolve the confusion because it's still semantically the same. For me, it became actually more confusing.

How about replacing the state with something like `current` to indicate that it is the current value of the computation?

YDX-2147483647:
At least the author of that post [thought that the star version is clearer](https://forum.typst.app/t/using-state-for-caching-in-typst-is-it-possible/5961/9?u=y.d.x):
> Yes, the new version with the star feature is much clearer and a lot less confusing, at least to me.

(It might just be a polite reply, though.)

> it's still semantically the same

Yes, but shouldn't all four variants be semantically the same? If we play this game with pen and paper, there is no distinction. It was only to write a legitimate typst document that they were distinguished.

> replacing the state with something like `current`

Sorry, I don't quite understand… Referring to the table in the PR description, do you mean `current`/current/`old`/⭐?

crates/typst-library/src/introspection/state.rs
```diff
@@ -87,18 +87,18 @@ use crate::routines::Routines;
...
+///   [New value is #context star.get().]
+/// }
 ///
 /// #compute("10") \
-/// #compute("x + 3") \
-/// #compute("x * 2") \
-/// #compute("x - 5")
+/// #compute("⭐ + 3") \
+/// #compute("⭐ * 2") \
+/// #compute("⭐ - 5")
```
YDX-2147483647, line 101: I think the confusing part is here.
In the original version, `state("x", ..)` and `expr.replace("x", ..)` both use `"x"`, but they mean different things.
In the new version, the become `"star"` and `"⭐"` respectively.

laurmaedje:
Okay, I'm convinced. Thanks!

Issue #2430, opened 2023-10-18
Footnote styling breaks links


EpicEricEE:
### Description

When using a show rule for a footnote entry, the link from the footnote in the text leads to the beginning of the document. For example:
```typ
#show footnote.entry: it => {
  let number = numbering(it.note.numbering, ..counter(footnote).at(it.note.location()))
  let body = it.note.body

  [#link(it.note.location())[#number)] #body]
}

This is up here #footnote[and this is down here]
```

When doing it the other way around, styling the footnote itself does not allow adding a link to the footnote entry, as there is no way to get its location.

### Operating system

Web app, Windows

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
@NiBrc I think that's a different thing. What you want is essentially https://github.com/typst/typst/issues/1356.

josineto:
Although I don't know Typst internals, I thought of a possible solution: to pass the footnote entry a field that contains all links that target it (meaning, an array of links). This way we could:

1. Customize the note styling and build it as a link target (inspecting the destination of one of those links to know which label the note was attached to)
2. Include a "back to the text" link in the note (although in the case of many links to one note this would be a little strange, but easily doable anyway)

josineto:
Hi! Is there any work going on in this issue? I'm writing my doctorate thesis with Pandoc+Typst, and this is the only bug that really affects my work... As my deadline is beginning of May, it would be nice to have until then correct footnote links...

Thank you anyway!

lorenzo:
Unfortunately, the recent fix in #6912 does not seem to work

Andrew15-5:
This issue is fixed in the latest version.

Although a correct show rule would be this:

```typ
#show footnote.entry: it => {
  let number = numbering("1)", ..counter(footnote).at(it.note.location()))
  let prefix = link(it.note.location(), number)
  let body = it.note.body
  h(it.indent)
  prefix
  h(0.05em, weak: true)
  body
}
```

https://github.com/typst/typst/blob/ddd8ae7f49d7adf35702fd9347d53fdd726ead55/crates/typst-layout/src/rules.rs?plain=1#L405-L414

lorenzo:
Thank you very much, with the new rule it works fine

PR #6912, opened 2025-09-12
Fix footnote links in presence of styling

laurmaedje:
Fixes https://github.com/typst/typst/issues/2430

### DIFF
--- a/crates/typst-ide/src/jump.rs
+++ b/crates/typst-ide/src/jump.rs
@@ -414,8 +414,15 @@ mod tests {
     }
 
     #[test]
-    fn test_backlink() {
+    fn test_footnote_links() {
         let s = "#footnote[Hi]";
-        test_click(s, point(10.0, 10.0), pos(1, 18.5, 37.1).map(Jump::Position));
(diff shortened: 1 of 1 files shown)


lorenzo:
I tested this fix with the recent 0.14.0-rc.1 release and it does not seem to work. I used this entry styling:

```typ
#show footnote.entry: it => {
    let loc = it.note.location()
    let count = numbering(
      "1. ",
      ..counter(footnote).at(loc),
    )
    (
      box(
        align(
          right,
          (count + h(0.3em)),
        ),
        width: 0em,
      )
    )
    it.note.body
  }
```

The link back to the text was not added.

I also tried with the same test as added in this PR:

```typ
#show footnote.entry: [Replaced]
```

Andrew15-5:
https://github.com/typst/typst/blob/ddd8ae7f49d7adf35702fd9347d53fdd726ead55/crates/typst-library/src/model/footnote.rs?plain=1#L303-L308

When would that fail? Seems related.

> The link back to the text was not added.

You must add it yourself. See https://github.com/typst/typst/issues/2430#issue-1950415839, and https://github.com/typst/typst/issues/2430#issuecomment-3393339746. Sounds like a problem better asked in [the forum](https://forum.typst.app).

PR #6915, opened 2025-09-12
Add an illustration for `par.{leading,spacing}` in docs

YDX-2147483647:
This PR adds an illustration for `par.{leading,spacing}` in docs.

图片

<details>
<summary>Old versions</summary>

图片

</details>

There're many magic constants in the code, but most of them are lengths for aesthetic spacings, so I guess that's okay.

Besides, the code is kind of overly automated. I think that is necessary to make sure it will survive in future Typst versions, considering that the rendered result are not tracked by git.

## Links

- [Previous discussion on Discord Documentation Forge](https://discord.com/channels/1054443721975922748/1399326777540608041/1416012368873263125).
- A Chinese version: https://github.com/typst-doc-cn/guide/pull/89


crates/typst-library/src/model/par.rs
```diff
@@ -108,6 +108,132 @@ pub struct ParElem {
...
+    ///           ),
+    ///           line(length: 1em, stroke: 0.5pt + c-line),
+    ///         ),
+    ///       ))
+    ///     },
+    ///   )
+    /// }
+    ///
+    /// // Bury metadata for the main text
+    /// #set par.line(numbering: n => [#metadata(n) <par-line>])
```
Andrew15-5: Since you don't actually count them, you don't need the par line number. And best practice for metadata labeling is to write them without whitespace. See https://github.com/typst/typst/issues/6844.

```suggestion
    /// #set par.line(numbering: _ => [#metadata(none)<par-line>])
```

crates/typst-library/src/model/par.rs
```diff
@@ -108,6 +108,132 @@ pub struct ParElem {
...
+    ///
+    /// #set par(justify: true)
+    /// #set text(luma(50%))
```
Andrew15-5: I don't like that the hyphen/comma sticks out of the colored area.

```suggestion
    /// #set text(luma(50%), overhang: false)
```

Also why luma is used? Does this increase contrast? I think it makes the text washed and less readable.

Andrew15-5 (commented):
I did not go through the whole code, but I did notice a few things. It's pretty massive, and I wonder if it can be shortened. Without blank lines and comments, it is shorter by 30 lines. But it's the 79 line context that might be shrinkable. I'm guessing not really.

crates/typst-library/src/model/par.rs
YDX-2147483647: I use luma because I want the reader to focus on margin notes. The lorem text is not the point, and is not meant be read at all.

crates/typst-library/src/model/par.rs
```diff
@@ -108,6 +108,132 @@ pub struct ParElem {
...
+    ///       .enumerate()
+    ///       .slice(1)
+    ///       .map(((i, y)) => {
+    ///         // Displacement from the last line relative to the line height
+    ///         let jump = (positions.at(i) - positions.at(i - 1)) / line-height
+    ///         if calc.abs(jump - 1) < 0.1 { "leading" } else { "spacing" }
+    ///       })
+    ///
+    ///     kinds.split("spacing").map(x => x.len() + 1)
+    ///   }
```
YDX-2147483647: > But it's the 79 line context that might be shrinkable. 

Yes, I'm also worried about this.

For example, the whole `let n-lines = { … }` can be replaced with `let n-lines = (4, 4, 2)`.
Not sure which way is better…

crates/typst-library/src/model/par.rs
hpcfzl: In that case, have you considered changing `aqua` to `gray`? It would make it even more inconspicuous. And then switching `orange` to `red` to make those parts pop out better?

YDX-2147483647: > In that case, have you considered changing aqua to gray? It would make it even more inconspicuous. 

I've considered completely hiding the text (keeping only the stripes), and also thought about replacing each letter with a small gray rectangle.
Now I think it's necessary to make the text barely readable, because careful readers might want to know the exact meaning of top and bottom edge in the illustration (by default, they are cap-height and baseline, respectively).

> And then switching `orange` to `red` to make those parts pop out better?

I'm against doing so. No good reason, but red is kind of connected with _wrong_ or _forbidden_, as in ❌🚦⛔. Orange is less aggressive.

crates/typst-library/src/model/par.rs
laurmaedje: I think keeping it simple is better

YDX-2147483647:
Hmm, I've reduced 29 lines. I think this is the limit I can do...

crates/typst-library/src/model/par.rs
```diff
@@ -108,6 +108,103 @@ pub struct ParElem {
...
     /// the first and last line.
+    ///
+    /// ```preview
+    /// // Color palette
+    /// #let c = (
+    ///   par-line: aqua.transparentize(60%),
+    ///   leading-line: blue,
+    ///   leading-text: blue.darken(20%),
+    ///   spacing-line: orange,
+    ///   spacing-text: orange.darken(10%),
```
laurmaedje:
```suggestion
    ///   spacing-line: orange.mix(red).darken(15%),
    ///   spacing-text: orange.mix(red).darken(20%),
```

crates/typst-library/src/model/par.rs
```diff
@@ -108,6 +108,103 @@ pub struct ParElem {
...
+    ///           ),
+    ///           line(length: 1em, stroke: 0.5pt + c-line),
+    ///         ),
+    ///       ))
+    ///     },
+    ///   )
+    /// }
+    ///
+    /// #set par(justify: true)
+    /// #set text(luma(50%), overhang: false)
```
laurmaedje:
```suggestion
    /// #set text(luma(25%), overhang: false)
```

laurmaedje (commented):
These colors pass WCAG contrast guidance. The previous ones were not ideal to read.

laurmaedje:
Thanks!

PR #6918, opened 2025-09-13
Replace the non-existent `$math.arrow` with `$math.accent` in the docs for vector

### LINKED ISSUE #6908: typst-docs resolves `$math.arrow` to a non-existent link
### Description

#6867 added a link to `$math.arrow`. ([Preview](https://ydx-typst.netlify.app/reference/math/vec/#summary))
However, typst-docs resolves it to `/reference/math/arrow/`, which does not exist.
It should resolve to somewhere in https://typst.app/docs/reference/math/accent/.

The following test will pass, but it's not the desired behaviour.

```rust
#[test]
fn test_math_accent() {
    assert_eq!(
        resolve_definition("$math.arrow", "/"),
        Ok("/reference/math/arrow/".into()) // does not exist
    );
}
```

YDX-2147483647:
There's no dedicated anchor for `arrow`, so just point to the page.

Resolves #6908

(Not the best solution, but it's an internal function, and this is the only occurrence.)


Enivex:
Edit: I missed the fact that only the link was changed, not the text. In any case fixing the anchor would be preferable.

YDX-2147483647:
> In any case fixing the anchor would be preferable.

Do you have suggestions on where to place the anchor? The row of `arrow` in the table on [that page](https://typst.app/docs/reference/math/accent/)?

I think the top of the page is better than the row, because it at least provides an example.

laurmaedje:
There isn't really a suitable spot for an anchor, so this is good I think. Thanks!

PR #6920, opened 2025-09-13
Initial Hayagriva bump

PgBiel:
Related: #6842

This should unblock failing tests in #6171.

Does not bump to main since main has breaking changes directly related to that PR, hence the "initial" bump.

Among other hayagriva changes and fixes, `chicago-notes` has been changed to now display full citations by default. `chicago-shortened-notes` is now available for the old style (although still a bit different from before as the chicago directives changed).

`chicago-fullnotes` in theory no longer exists, but has been kept as an alias in hayagriva. This PR shall implement a warning so users using that style move to `chicago-notes.`

TODO:
- [ ] Add the warning


PgBiel:
Of note, the following styles were also removed:

-  "modern-humanities-research-association"
- "turabian-author-date"
- "turabian-fullnote-8"

The first one was apparently replaced with `"modern-humanities-research-association-notes"`, the latter two became dependent styles and thus are now unsupported by hayagriva. Not sure if there is much we can do about the turabian styles... But I could add a similar alias with warning for the first one as well if that's interesting.

reknih:
I kinda hate that the CSL project gives no guarantee about stability of the identifiers. Maybe we should just expose an alias for every dependant style instead of accumulating a list of fallbacks.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -438,6 +440,29 @@ pub enum CslSource {
...
+    let Some(value) = value else {
+        return Ok(None);
+    };
+
+    if let Value::Str(style) = &value.v
+        && &**style == "chicago-fullnotes"
+    {
+        engine.sink.warn(crate::diag::warning!(
+            value.span,
+            "style 'chicago-fullnotes' has been deprecated in favor of 'chicago-notes'"
```
laurmaedje: the literal strings should be double-quoted as `'` doesn't have that meaning in Typst. in some cases, we also have extra backticks in the error message around expressions (which makes it monospace in the web app), but whether that's the case is not quite consistent I think. in the future, we could also have nicer rendering for that in the CLI though (e.g. different color).

crates/typst-library/src/model/bibliography.rs
```diff
@@ -438,6 +440,29 @@ pub enum CslSource {
     Normal(DataSource),
 }
 
+/// Parses and checks for a deprecated CSL style.
+pub(super) fn check_csl_style<T: FromValue<Spanned<Value>>>(
```
laurmaedje: could this be integrated into `CslStyle::load`? intuitively that would feel cleaner to me at first glance, but I might be wrong.

tests/ref/bibliography-before-content.png
laurmaedje, line 1: is it correct that there is now a space on both sides of the colon in the first reference? looks odd. (not that the previous thing didn't look odd...)

tests/suite/model/bibliography.typ
```diff
@@ -95,3 +95,8 @@ Now we have multiple bibliographies containing @glacier-melt @keshav2007read
 @Zee04
 
 #bibliography("/assets/bib/works_too.bib")
+
+--- bibliography-chicago-fullnotes-warning ---
```
laurmaedje, line 99: could add `show heading: none` to skip the ref image

tests/ref/cite-footnote.png
laurmaedje, line 1: let's switch this test and the two `issue-\d+` tests to the shortnotes style since they already specific a style anyway

Cargo.toml
```diff
@@ -61,7 +61,7 @@ fontdb = { version = "0.23", default-features = false }
 fs_extra = "1.3"
 rustc-hash = "2.1"
 glidesort = "0.1.2"
-hayagriva = "0.8.1"
+hayagriva = { git = "https://github.com/typst/hayagriva", rev = "03ae2acf168e65604e74f0361e74ba4ef7689f16" }
```
laurmaedje: a 7 digit rev suffices

tests/ref/bibliography-before-content.png
PgBiel: Likely not much we can do about that. :pensive:

Issue #1524, opened 2023-06-18
Advance Unicode-Support: Compile/Render x⃗, x̅, x̂, √x, ... correctly


Sinthoras7:
### Description

The "combining diacritical marks" (i.e. the unicode symbols U+0300 to U+036F), for example a⃗, ḃ, x̂, etc. should be rendered correctly in math mode. Right now, they create an error.
Also √𝛼 for one-character square roots.  

### Use Case

It is useful to type as much in unicode as possible to have less characters and make the source code cleaner/more readable. (Also if one remaps the keyboard to output these, it is possible to communicate math on every platform much more easily.)

Enivex:
I can't reproduce this

`$a\u{20D7}$` yields 
![image](https://github.com/typst/typst/assets/2191754/dbacea5f-1887-4877-84c4-4bcbb16bdd33)
like it should

laurmaedje:
Did you try the square root one? It should already work.

Sinthoras7:
You're right. This already works. But e.g. x⃗, does not work yet.

mkhan45:
I added the symbols I use the lazy way here: https://github.com/typst/typst/compare/main...mkhan45:typst:main

mkhan45:
I guess I could've used `symbols/sym.rs`

Evidlo:
For anyone else who found this and is looking for a solution, here is how to do the equivalent of Latex \overrightarrow{x}

    $limits(x)^arrow.r$

PR #6929, opened 2025-09-18
Handle multiple glyphs after shaping in math

### LINKED ISSUE #6328: Calligraphic Math Unicode literals with variation selectors not recognised as a single character
### Description

I have a strong preference towards inputting Unicode directly in my source files, which makes the source files significantly easier to read and use in my opinion.

I was playing around with the calligraphic math characters, which as of Unicode 14.0 take a variation selector (0xFE00 or 0xFE01 suffix), but these are not recognised as a single character/grapheme, which leads to an `Unknown Variable` error.

```typst
$
"Default (OK)" : &ℱ\
"Calligraphic": &ℱ︀\
"Script": &ℱ︁\
$

Which gives error "Unknown variable ℱ︀", which I suspect is due to the fact that it considers both the base character 0x2131 and variation selector 0xFE01 to be characters, which is not the case.


### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

### LINKED ISSUE #7294: Suboptimal error message when math text contains more than one glyph
### Description

Consider this (U+0332 is Combining Low Line):

```typ
#let arreq = symbol("↔\u{0332}")
$arreq$
```

This produces the following error message when using `arreq`:
> Shaping the text `"↔\u{332}"` yielded more than one glyph

This error message could probably be improved (for a regular user, this doesn't explain the problem and how to solve it). Ideally, I tend to think it should be a warning, and all the glyphs should be displayed.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

mkorje:
Finishes https://github.com/typst/typst/pull/6489#discussion_r2313534762. In the process I also tidied up some of the shaping logic. The main change is the introduction of a `ShapedText` struct which is used instead of `TextItem` directly.

The first commit fixes #6328. I've included it here as it requires the changes made here anyways (and allows testing the subsequent changes).

Closes #7294

### DIFF
--- a/crates/typst-layout/src/math/shaping.rs
+++ b/crates/typst-layout/src/math/shaping.rs
@@ -10,7 +10,7 @@ use typst_library::text::{Font, FontFamily, FontVariant, Glyph, Lang, Region, Te
 use typst_library::visualize::{FixedStroke, Paint};
 use typst_syntax::Span;
 
-use crate::inline::create_shape_plan;
+use crate::inline::{SharedShapingContext, create_shape_plan, get_font_and_covers};
 
 /// A text item in Math.
 ///
@@ -165,11 +165,6 @@ pub fn shape(
 /// Holds shaping results and metadata for shaping some text.
 struct ShapingContext<'a> {
     world: Tracked<'a, dyn World + 'a>,
-    /// Font families that have been used with unlimited coverage.
-    ///
-    /// These font families are considered exhausted and will not be used
-    /// again, even if they are declared again (e.g., during fallback after
-    /// normal selection).
     used: Vec<Font>,
     variant: FontVariant,
     features: Vec<rustybuzz::Feature>,
@@ -179,40 +174,36 @@ struct ShapingContext<'a> {
     font: Option<Font>,
 }
 
+impl<'a> SharedShapingContext<'a> for ShapingContext<'a> {
+    fn world(&self) -> Tracked<'a, dyn World + 'a> {
+        self.world
+    }
+
+    fn used(&mut self) -> &mut Vec<Font> {
+        &mut self.used
+    }
+
+    fn first(&self) -> Option<&Font> {
+        self.used.first()
+    }
+
+    fn variant(&self) -> FontVariant {
+        self.variant
+    }
+
+    fn fallback(&self) -> bool {
+        self.fallback
+    }
+}
+
 /// Shape text with font fallback using the `families` iterator.
 fn shape_impl<'a>(
-    ctx: &mut ShapingContext,
+    ctx: &mut ShapingContext<'a>,
     text: &str,
     mut families: impl Iterator<Item = &'a FontFamily> + Clone,
 ) {
-    // Find the next available family.
-    let book = ctx.world.book();
-    let mut selection = None;
-    let mut covers = None;
-    for family in families.by_ref() {
-        selection = book
-            .select(family.as_str(), ctx.variant)
-            .and_then(|id| ctx.world.font(id))
-            .filter(|font| !ctx.used.contains(font));
-        if selection.is_some() {
-            covers = family.covers();
-            break;
-        }
-    }
-
-    // Do font fallback if the families are exhausted and fallback is enabled.
-    if selection.is_none() && ctx.fallback {
-        let first = ctx.used.first().map(Font::info);
-        selection = book
-            .select_fallback(first, ctx.variant, text)
-            .and_then(|id| ctx.world.font(id))
-            .filter(|font| !ctx.used.contains(font))
-    }
-
-    // Extract the font id or shape notdef glyphs if we couldn't find any font.
-    let Some(font) = selection else {
-        if let Some(font) = ctx.used.first().cloned() {
-            // Shape tofus.
+    let Some((font, covers)) =
+        get_font_and_covers(ctx, text, families.by_ref(), |ctx, text, font| {
             let add_glyph = |_| {
                 ctx.glyphs.push(ShapedGlyph {
                     id: 0,
@@ -224,14 +215,10 @@ fn shape_impl<'a>(
             };
             text.chars().for_each(add_glyph);
             ctx.font = Some(font);
-        }
+        })
+    else {
         return;
     };
-
-    // This font has been exhausted and will not be used again.
-    if covers.is_none() {
-        ctx.used.push(font.clone());
-    }
 
     let mut buffer = UnicodeBuffer::new();
     buffer.push_str(text);
--- a/crates/typst-layout/src/inline/mod.rs
+++ b/crates/typst-layout/src/inline/mod.rs
@@ -9,7 +9,7 @@ mod prepare;
 mod shaping;
 
 pub use self::box_::layout_box;
-pub use self::shaping::create_shape_plan;
+pub use self::shaping::{SharedShapingContext, create_shape_plan, get_font_and_covers};
 
 use comemo::{Track, Tracked, TrackedMut};
 use typst_library::World;
--- a/crates/typst-layout/src/inline/shaping.rs
+++ b/crates/typst-layout/src/inline/shaping.rs
@@ -4,12 +4,13 @@ use std::ops::Deref;
 use std::sync::Arc;
 
 use az::SaturatingAs;
+use comemo::Tracked;
 use rustybuzz::{BufferFlags, Feature, ShapePlan, UnicodeBuffer};
 use ttf_parser::Tag;
 use ttf_parser::gsub::SubstitutionSubtable;
 use typst_library::World;
 use typst_library::engine::Engine;
-use typst_library::foundations::{Smart, StyleChain};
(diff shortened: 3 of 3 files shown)


crates/typst-syntax/src/lexer.rs
```diff
@@ -624,8 +624,12 @@ impl Lexer<'_> {
...
             c if is_math_id_start(c) && self.s.at(is_math_id_continue) => {
                 self.s.eat_while(is_math_id_continue);
-                let (kind, node) = self.math_ident_or_field(start);
-                return (kind, Some(node));
+                if self.s.from(start).graphemes(true).count() == 1 {
+                    SyntaxKind::MathText
+                } else {
+                    let (kind, node) = self.math_ident_or_field(start);
+                    return (kind, Some(node));
+                }
```
isuffix: You can actually do this by only iterating a single grapheme like so:
```rust
let (last_index, _) =
    self.s.from(start).grapheme_indices(true).last().unwrap();
if last_index == 0 {
    // If this was just a single grapheme
    SyntaxKind::MathText
} else {
    ...
}
```
Little tip I picked up working on https://github.com/rust-lang/libs-team/issues/576

crates/typst-syntax/src/lexer.rs
```diff
@@ -679,14 +682,8 @@ impl Lexer<'_> {
...
             self.s.jump(start + len);
-            if len > c.len_utf8() {
-                // Grapheme clusters are treated as normal text and stay grouped
-                // This may need to change in the future.
-                SyntaxKind::Text
-            } else {
-                SyntaxKind::MathText
-            }
         }
+        SyntaxKind::MathText
```
isuffix, line 708: Since we aren't producing the `Text` kind anymore we can remove it from the `MATH_EXPR` syntax set and from the match statement in `math_expr_prec()`

isuffix (commented):
This looks nice! Got two comments on the first commit though

crates/typst-syntax/src/lexer.rs
mkorje: Very nice, thanks!

mkorje: Should be `.next_back()` instead of `last()` though, otherwise clippy complains that it iterates through the whole thing anyways :sweat_smile:.

crates/typst-syntax/src/lexer.rs
```diff
@@ -679,14 +685,8 @@ impl Lexer<'_> {
                 .next()
                 .map_or(0, str::len);
             self.s.jump(start + len);
-            if len > c.len_utf8() {
-                // Grapheme clusters are treated as normal text and stay grouped
```
laurmaedje, line 702: can someone remind me what this was doing? I don't quite understand the diff here.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -382,19 +373,22 @@ impl GlyphFragment {
...
         let id = GlyphId(self.item.glyphs[0].id);
 
         let extended_shape = is_extended_shape(&self.item.font, id);
         let italics = italics_correction(&self.item.font, id).unwrap_or_default();
         let width = self.item.width();
-        if !extended_shape {
+        // The second condition below is needed so we don't accidently add the
+        // italic correction again to the original glyphs. This can happen when
+        // `reset_glyph` is called after the glyph fragment is created.
+        if !extended_shape && (initial || id.0 != self.item.original_id()) {
```
laurmaedje, line 391: This condition here seems slightly hacky. I wonder whether there isn't a cleaner, less mutable way to produce the metrics. But we can also leave that as future work.

crates/typst-layout/src/math/shaping.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub fn update(&mut self, new_value: Vec<ShapedGlyph>) {
+        self.updated = Some(new_value);
+    }
+
+    /// Reset the glyphs back their originals.
+    pub fn reset(&mut self) {
+        self.updated = None;
+    }
+
+    fn original(&self) -> &Vec<ShapedGlyph> {
```
laurmaedje:
```suggestion
    fn original(&self) -> &[ShapedGlyph] {
```

crates/typst-layout/src/math/shaping.rs
```diff
@@ -0,0 +1,281 @@
...
+    variant: FontVariant,
+    features: Vec<rustybuzz::Feature>,
+    language: rustybuzz::Language,
+    fallback: bool,
+    glyphs: Vec<ShapedGlyph>,
+    font: Option<Font>,
+}
+
+/// Shape text with font fallback using the `families` iterator.
+fn shape_impl<'a>(
```
laurmaedje, line 200: the code here is looking ever more similar to the shaping code in inline layout (now with the shaping context and fallback). do you think we have a chance of unifying it or at least extracting common subroutines?

crates/typst-layout/src/math/shaping.rs
```diff
@@ -0,0 +1,281 @@
...
+            let add_glyph = |(_, _)| {
+                ctx.glyphs.push(ShapedGlyph {
+                    id: 0,
+                    x_advance: font.x_advance(0).unwrap_or_default(),
+                    x_offset: Em::zero(),
+                    y_advance: Em::zero(),
+                    y_offset: Em::zero(),
+                })
+            };
+            text.char_indices().for_each(add_glyph);
```
laurmaedje: why char_indices instead of chars if the indices are dropped?

crates/typst-layout/src/math/shaping.rs
```diff
@@ -0,0 +1,281 @@
...
+        buffer.direction(),
+        buffer.script(),
+        buffer.language().as_ref(),
+        &ctx.features,
+    );
+
+    let buffer = rustybuzz::shape_with_plan(font.rusty(), &plan, buffer);
+    // Because we will only ever shape single grapheme clusters, we can assume
+    // that the output from the shaper is a single cluster that spans the
+    // entire range of the given text.
```
laurmaedje: Since grapheme clusters and shaping clusters are not 100% the same, I'm not sure this actually holds always. But I don't know when it wouldn't, if at all. Reminds me of this comment: https://github.com/typst/typst/pull/5099#issuecomment-2392260461

crates/typst-layout/src/math/fragment.rs
mkorje: There probably is, but this was the nicest thing I could come up with that works.
I guess the issue is its being used both to set a glyph's initial metrics and also modify them later.

crates/typst-layout/src/math/shaping.rs
mkorje: No idea, I'll just use chars.

crates/typst-syntax/src/lexer.rs
mkorje: Now that you mention it, I have no idea what effect the old code actually had. :sweat_smile:

crates/typst-layout/src/math/shaping.rs
mkorje: I don't think we'd be able to unify it since the `ShapedGlyph`s produced are a different struct. Maybe some parts can be extracted, I'll try it out.

crates/typst-layout/src/math/shaping.rs
mkorje: Ah good catch. The only negative consequences I can see atm are:
- we don't then reshape by cluster, but I think that's fine here as we'd want the entire shaped grapheme cluster to use the same font
- the range indices could be off in the `TextItem` `Glyph`s, so text extraction might be wrong

mkorje: I thought the `math-text-grapheme-cluster-fallback` test might cause some wrong text extraction, but it seems fine.

mkorje:
> Does this also fix #1524? At least `$ φ̅ $` seems to work. Or is the thought that we now only support whatever the shaper can do instead of going through Typst's accent layout?

That was not intentional, though it isn't a proper fix anyways since it is using the shaper instead of the math layout for the accent placement.

crates/typst-syntax/src/lexer.rs
mkorje: cc @isuffix

crates/typst-layout/src/math/shaping.rs
mkorje: I extracted out the initial font and cover selection logic with a trait. See the latest commit.

crates/typst-syntax/src/lexer.rs
isuffix: This check separated single-character vs. multi-character grapheme clusters so the latter would become `Text` nodes in the AST instead of `MathText`. This allowed non-numeric `MathText` to turn directly into `SymbolElem` during eval, since multi-character `SymbolElem` wasn't supported before (which is the whole point of this PR in the first place 😁).

laurmaedje:
Thank you!

Issue #6891, opened 2025-09-09
`default` parameter for `array.join`


MDLC01:
### Description

Currently, calling `array.join` on an empty array returns `none`. There should be a `default` argument, similar to the one on `array.sum` or `array.product`, defining the value to return for an empty array (defaulting to `none`).

### Use Case

It is easier to write `arr.join(", ", default: "")` than

```typ
#if arr.len() == 0 {
  ""
} else {
  arr.join(", ")
}
```

### Drawbacks

As mentioned in https://github.com/typst/typst/pull/6780#issuecomment-3195505968, the proliferation of `default` arguments may indicate than a more general solution should be implemented. In the meantime, this proposal makes sense.

laurmaedje:
I am okay with this.

PR #6932, opened 2025-09-19
Add `default` argument to `array.join`

MDLC01:
Closes https://github.com/typst/typst/issues/6891.

One open question is what to do when no default is provided and the array is empty. For now, I chose to return `none` (i.e., no change in behavior). `array.sum` and `array.product` raise an error in this case, so we might want to align with that.


laurmaedje:
Since `join` is frequently used for displaying content, I think the `none` default is quite sensible, even if it is slightly inconsistent with `sum` and `product`.

laurmaedje:
Thanks!

Issue #6934, opened 2025-06-26
Inconsistent capitalization in documentation page names


MDLC01:
### Description

Some documentation pages use Title Case, while others use Sentence case. For example, "Web App" and "Data Loading" use Title Case, but "Private packages" and "Table guide" use Sentence case.

Part of the code associated to this issue is located in [typst/typst](https://github.com/typst/typst), but the sources for the pages under the [Web App](https://typst.app/docs/web-app/) category are in a private repo I think, hence why I open this here.

laurmaedje:
I have fixed the ones in the proprietary code and transferred the issue to cover the rest.

PR #6936, opened 2025-09-19
Use Title Case for Doc Pages

MDLC01:
Fixes https://github.com/typst/typst/issues/6934.

I also changed all references to the pages in question to use title case as well.


laurmaedje:
Thanks!

PR #6942, opened 2025-09-20
Eliminate unnecessary `usize` usages from `typst-syntax`

T0mstone:
This follows the spirit of #6026: `usize` should only be used for ~~indices and lengths~~ things bounded by memory.

This PR is separated into disjoint commits to make reviewing easier.

Notes:
- For the precedences, I just went with the smallest type that fits all the values we use, so `u8`. There's also a case to be made for `u16`, but I think if there's ever a need for it, it can just be changed again then.
- For `MathRoot`, I think it's safe to say that we won't have syntax for roots with index greater than 255 in the future, so `u8` is the best choice.
- ~~For the other two, I just went with `u64` for obvious reasons. I'm actually not perfectly sure about these: They're not explicitly indices or lengths, but I don't know if it might not still be impossible for them to exceed `usize::MAX`... If so, then those two commits should of course be dropped.~~ Edit (see https://github.com/typst/typst/pull/6942#pullrequestreview-3250067497): These are justified in their use of `usize`.

I checked all occurrences of `usize` in `typst-syntax` and everything besides what I changed here is justified (indices, ranges, and lengths).

### DIFF
--- a/crates/typst-syntax/src/ast.rs
+++ b/crates/typst-syntax/src/ast.rs
@@ -1068,7 +1068,7 @@ node! {
 
 impl<'a> MathRoot<'a> {
     /// The index of the root.
-    pub fn index(self) -> Option<usize> {
+    pub fn index(self) -> Option<u8> {
         match self.0.children().next().map(|node| node.text().as_str()) {
             Some("∜") => Some(4),
             Some("∛") => Some(3),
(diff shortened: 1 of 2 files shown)


T0mstone:
Yeah, after thinking it over, you're right. Technically, `SyntaxNode` doesn't allocate its root node, so `descendants()` could theoretically return `usize::MAX + 1`, but that edge case is probably not worth supporting. (At that point, running out of memory would be the bigger problem.)

laurmaedje:
Thanks!

Issue #6937, opened 2025-09-19
Allow resetting PDF bookmark hierarchy (e.g., for appendices after parts)


jassielof:
### Description

In LaTeX, the bookmark package allows authors to reset the PDF `bookmark` hierarchy using `\bookmarksetup{startatroot}`. This is especially useful when transitioning from one major document division (like \part) to another structurally distinct section (such as appendices), ensuring that the new section appears at the top level in the PDF outline rather than nested under the previous part.

Currently, Typst automatically nests headings in the PDF bookmark tree based on their position in the document structure. However, there is no way to manually reset or control the bookmark hierarchy, which can lead to misleading or cluttered outlines—especially in complex documents like books or theses.

This issue is related to:

- #1226
- #5615

### Use Case

Consider a book structured with multiple `part` sections followed by appendices. In Typst, the appendices will appear nested under the last part in the PDF bookmarks, even though they conceptually belong to a separate section. This misrepresents the document’s logical structure and can confuse readers navigating the PDF.

Being able to reset the bookmark hierarchy would allow authors to:

- Start a new top-level section in the PDF outline (e.g., for appendices, references, or indices).
- Improve accessibility and navigation in long-form documents.
- Match the semantic structure of the document more accurately in the exported PDF.

Issue #1226, opened 2023-05-16
Manual bookmarks


alfla673:
### Description

Ability to add bookmarks to the rendered pdf file manually, and the ability to modify the bookmarks generated automatically based on the headings.

### Use Case

In some cases you might not want pdf bookmarks corresponding exactly to the headings within the document.

johannesrave:
i would like the pdf bookmarks to contain the numbering from the top-level headings (chapters) as well - is this currently possible?

laurmaedje:
@johannesrave It's not currently possible, but there's a separate issue open for it: https://github.com/typst/typst/issues/2416

billdenney:
I'm a new Typst user. And, I came to look for this feature. My use case for wanting manual pdf bookmarks is a bit different.

When I make pdf docs, I need to put figure and table captions in the bookmark list twice: once at the list of figures/tables and once at the figure itself. When done at the list of figures/tables, I nest the bookmarks under "List of figures" (or tables). When in the main body, I nest the bookmarks within the (sub)section.

I make the bookmark say "Figure 3. Short caption". It is auto-generated by the addition of a figure caption.

I realize that there is a lot going on here. Hopefully, something like this is possible to accomplish.

tristan957:
What should the syntax look like?

```typst
#title("Hello World", bookmark: bookmark("Other Title", headings: true)
```

billdenney:
I think that it needs at least two types of syntax. One is to add to the bookmarks with a link location, the bookmark text, and an indentation level, and the other syntax is to add a linking location for the bookmark. I don't know if bookmarks use other location information.

I think that the syntax may look something like:

```
#bookmark("bookmark text", indentation = 1, location = "link-location")
```

Then elsewhere, we would need to use something like `<link-location>`. The location could default to `#here()`, if not given to make a bookmark at the current indentation and location.

(I'm no Typst expert or even novice. I could be way off on the best way to handle this.)

PR #6946, opened 2025-09-21
PDF: Add heading(bookmark-level) to override bookmark nesting (start at root)

jassielof:
Fixes: typst/typst#6937

## Summary
- Introduces a new heading property `bookmark-level: auto|int` that overrides the heading’s level only for PDF bookmarks.
- Default `auto` uses the structural level; setting `1` starts a heading at the root of the PDF outline.
- This is strictly PDF-only. It does not affect Typst’s document outline (`outline()`), HTML, or other outputs.

## Usage example
```typ
#outline()

= Part I
== Chapter 1

#show heading.where(level: 2): set heading(bookmark-level: 1)

== Appendix A
=== Appendix A.1
```

image


PgBiel:
Thanks for the PR. At first, I'm not sure I see the appeal here. You mention in the original issue:

>  In Typst, the appendices will appear nested under the last part in the PDF bookmarks, even though they conceptually belong to a separate section.

Shouldn't this be a case for elevating parts and appendices to the same level as parts then? It seems that there would be a semantic mismatch otherwise, even if just for the reader.

Sure, parts and appendices should probably be displayed differently, but I'd say that's a different problem to solve (which can be done with e.g. labels, metadata, and/or custom functions encapsulating different styles and behavior).

jassielof:
Sure the appendices could be moved to a higher level, but wouldn't that cause to restyle the appendices to match the chapter's style?
The bookmark level it's mostly for easier control of the bookmarks in that case, it defaults to auto, so there's no mismatch unless wanted by the author. It's also related to the LaTeX bookmark and hyperref packages and the start at root macro.

PgBiel:
> Sure the appendices could be moved to a higher level, but wouldn't that cause to restyle the appendices to match the chapter's style?

That's precisely the different problem I was referring to. Your problem isn't regarding heading levels, but differing styles.
There are a few ways to solve this at the template level. Here's one way:

```typ
// template.typ
#let started-appendices = state("started-appendices", false)
#let appendix-style(it) = { /* ... */ }
#let part-style(it) = { /* ... */ }

#let template(doc, /* ... */) = {
  // ...
  show heading.where(level: 1): context if started-appendices.get() { appendix-style(it) } else { part-style(it) }

  doc
}
```
Then, in your main document, you'd have something like

```typ
// main.typ
#import "template.typ": template, started-appendices

#show: template.with(/* settings ... */)

= First Part

Hello World!

#started-appendices.update(true)

= First Appendix

Hello World!

= Second Appendix

Hello World!
```

which would shift the style of headings to look like appendices rather than parts.

laurmaedje:
As @PgBiel mentioned, for some uses cases, the best solution is to just adjust how you are using which heading level in your document.

Still, there may be use cases where it could be desirable to adjust the PDF outline independently from the heading structure. This is, however, then a low-level concern of PDF export and not really well placed on the `heading` function (I'm already on the fence with having both `outlined` and `bookmarked` tbh).

As [discussed on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1420161570926886992), there are other potential designs for this, e.g. a bookmark element or similar (https://github.com/typst/typst/issues/1226 is also relevant here). I think this needs some more design work & discussion before we're ready to move on to the implementation stage. For this reason, I'll close the PR for the time being. Thanks for your understanding and also thanks for your effort still!

jassielof:
Hmm, ok, I'll try to come up with a better design and update it in the issue, considering the other relevant PDF (bookmark-related) issues, and try to ask for feedback and caveats.

Ideally, it's to not have it more coupled to the heading element, like the bookmark and outline.

Thanks.

PR #6952, opened 2025-09-23
Support bibliography and citations linking to it in HTML

breezykermo:
A follow-on from #6894 that hews more closely to [the HTML spec](https://www.w3.org/TR/dpub-aria-1.1/#doc-bibliography) in terms of roles and also provides links from citations to their references.

I noticed that the work of finding the appropriate locations was already done when parsing the keys etc from hayagriva, so it seemed more appropriate to surface that in the `Works` data structure than to do it again. But I am still new to Typst internals, so please let me know if this is the correct approach or if there are downstream consequences of adding a field of which I might not be aware.

## Performance implications
The implications of adding the extra hash map to the `Works` data structure seem (from my naive point of view) acceptable because:
- Values are only added if the reference is actually printed in the bibliography. In other words, there is not necessarily an entry for every listing in a large `.bib` file. Thus while it does require some extra memory in the compiler, it is not a large amount (I don't think), and it scales linearly with the number of items that appear in the bibliography.

## Questions
It's not entirely clear to me why the `CiteGroup` doesn't automatically create a `LinkElem` in the HTML render pathway. I assume from #6917 that it has to be explicitly added for some reason to get an `<a>` tag to appear in HTML, but I would be happy for someone to point out that there is some easier way to get it than what I have done here.

With thanks to @willcrichton for his help in thinking about this change (though any and all deficiencies in the approach are my responsibility solely!)

### DIFF
--- a/tests/ref/html/bibliography-no-title.html
+++ b/tests/ref/html/bibliography-no-title.html
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
   </head>
   <body>
-    <p role="doc-biblioref"><a id="loc-1" href="#loc-2">[1]</a></p>
+    <p><a id="loc-1" href="#loc-2" role="doc-biblioref">[1]</a></p>
     <section role="doc-bibliography">
       <ul style="list-style-type: none">
         <li id="loc-2"><span class="prefix"><a href="#loc-1" role="doc-backlink">[1]</a></span> B. Aldrin, “An Insight into Bibliographical Distress.”<span style="white-space: pre-wrap">&#x20;</span></li>
--- a/crates/typst-html/src/convert.rs
+++ b/crates/typst-html/src/convert.rs
@@ -152,7 +152,9 @@ fn handle_html_elem(
     elem: &Packed<HtmlElem>,
     styles: StyleChain,
 ) -> SourceResult<()> {
-    let role = styles.get_cloned(HtmlElem::role);
+    // See the docs of `HtmlElem::role` for why we filter out roles for `<p>`
+    // elements.
+    let role = styles.get_cloned(HtmlElem::role).filter(|_| elem.tag != tag::p);
 
     let mut children = EcoVec::new();
     if let Some(body) = elem.body.get_ref(styles) {
--- a/crates/typst-html/src/lib.rs
+++ b/crates/typst-html/src/lib.rs
@@ -80,6 +80,12 @@ pub struct HtmlElem {
     /// element instead of a subtree, they could supplant this. If we need the
     /// same mechanism for things like `class`, this could potentially also be
     /// extended to arbitrary attributes. It's minimal for now.
+    ///
+    /// This is ignored for `<p>` elements as it otherwise tends to
+    /// unintentionally attach to paragraphs resulting from grouping of a single
+    /// element instead of attaching to that element. This is a bit of a hack,
+    /// but good enough for now as the `role` property is purely internal and
+    /// we control what it is used for.
     #[internal]
     #[ghost]
     pub role: Option<EcoString>,
--- a/tests/ref/html/bibliography-custom-title.html
+++ b/tests/ref/html/bibliography-custom-title.html
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
   </head>
   <body>
-    <p role="doc-biblioref"><a id="loc-1" href="#loc-2">[1]</a></p>
+    <p><a id="loc-1" href="#loc-2" role="doc-biblioref">[1]</a></p>
     <section role="doc-bibliography">
       <h2>My References</h2>
       <ul style="list-style-type: none">


crates/typst-library/src/model/bibliography.rs
```diff
@@ -521,16 +564,27 @@ impl IntoValue for CslSource {
...
 }
 
+/// Lists all references in the bibliography, with optional prefix
+type References = Vec<(Option<Content>, Content)>;
+
+/// Maps from the location of a citation group to its rendered content.
+type CitationLocationContentMap = FxHashMap<Location, SourceResult<Content>>;
+
+/// Maps from an item key to the location in the bibliography of that item.
+type CitationKeyReferenceLocationMap = FxHashMap<EcoString, Location>;
```
PgBiel: I feel like these aliases don't pull their weight and can be removed. Each alias is also harder to read and reason about than the original type, which is otherwise self-evident on what it is about (maps certain X to Y, and the purpose can be inferred from local context). 

The exception for that last statement is the `References` type, whose context might not be immediately clear (in a vacuum). However, that type is only used once, in the definition of the `references` field, so in practice the context is always clear, so the alias should also be removed.

Notably, it isn't deduplicating an otherwise "long" type repeated in multiple places. As a general rule of thumb, we first wait for the problem (very repeated usage) to arrive before refactoring. Not only for aesthetic reasons, but also to prevent us from constantly "refactoring the world" while developing, keeping us sane :)

crates/typst-layout/src/rules.rs
```diff
@@ -435,34 +435,12 @@ const BIBLIOGRAPHY_RULE: ShowFn<BibliographyElem> = |elem, engine, styles| {
...
-                "CSL style \"{}\" is not suitable for bibliographies",
-                style.display_name()
-            ),
-            CslSource::Normal(..) => {
-                "CSL style is not suitable for bibliographies".into()
-            }
-        })
-        .at(span)?;
+    let works = elem.realize_works(engine, styles)?;
+    let references = works.references.as_ref().unwrap();
```
PgBiel: This is a bit of an anti-pattern as it doesn't seem, at first, evident that `realize_works` would guarantee that `references` is `Some`.

The following would be more idiomatic (with the appropriate changes to `realize_works`).

```suggestion
    let (works, references) = elem.realize_works(engine, styles)?;
```

crates/typst-library/src/model/bibliography.rs
```diff
@@ -192,6 +193,48 @@ impl BibliographyElem {
...
+                    .spanned(span)
+            })
+    }
+
+    /// Fetches and validates the bibliography works.
+    pub fn realize_works(
+        &self,
+        engine: &mut Engine,
+        styles: StyleChain,
+    ) -> SourceResult<std::sync::Arc<Works>> {
```
PgBiel: Arc was already imported, no need to qualify.
```suggestion
    ) -> SourceResult<Arc<Works>> {
```

PgBiel: Though see my other comment which indicates this signature will need further changes.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -192,6 +193,48 @@ impl BibliographyElem {
...
+                    "CSL style \"{}\" is not suitable for bibliographies",
+                    style.display_name()
+                ),
+                CslSource::Normal(..) => {
+                    "CSL style is not suitable for bibliographies".into()
+                }
+            })
+            .at(span);
+        }
+        Ok(works)
```
PgBiel: Should do something like the original code here (to satisfy my other comment about guaranteeing references won't be `None`)


```rs
let references = works.references.as_ref().ok_or_else(|| {
    match self.style.get_ref(styles).source {
        CslSource::Named(style, _) => eco_format!(
            "CSL style \"{}\" is not suitable for bibliographies",
            style.display_name()
        ),
        CslSource::Normal(..) => {
            "CSL style is not suitable for bibliographies".into()
        }
    }
})
    .at(span)?;
Ok((works, references))
```

crates/typst-layout/src/rules.rs
PgBiel: Otherwise, nice job deduplicating this part across targets.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -732,25 +786,25 @@ impl<'a> Generator<'a> {
...
-    ) -> StrResult<FxHashMap<Location, SourceResult<Content>>> {
+    ) -> StrResult<(CitationLocationContentMap, CitationKeyReferenceLocationMap)> {
         // Determine for each citation key where in the bibliography it is,
         // so that we can link there.
         let mut links = FxHashMap::default();
         if let Some(bibliography) = &rendered.bibliography {
             let location = self.bibliography.location().unwrap();
             for (k, item) in bibliography.items.iter().enumerate() {
-                links.insert(item.key.as_str(), location.variant(k + 1));
+                links.insert(item.key.clone().into(), location.variant(k + 1));
```
PgBiel: This is 1. cloning the original string and 2. creating a new EcoString from the cloned string, which means the string is being rebuilt twice instead of only once. Something like `EcoString::from(&item.key)` or `eco_format!("{}", item.key)` would do the same job in one step.

Though I feel a bit uneasy still. I wonder what's stopping us from owning `hayagriva::Rendered` since it is immediately discarded upon creation, and we could then just move the strings directly without a clone.

But that is a harder / broader change, so you can attempt that separately after fixing this first.

PgBiel (commented):
Thanks for the contribution! I wanted to leave some initial thoughts before further opinions from the team. Most of them are related to code style and general quality, but otherwise this is looking nice! Let me know if you have any questions.

By the way, please don't close your pull request in order to apply changes, even if you want to start fresh. This causes the original conversation history and reviews to be lost (or, rather, not attached to the new PR), making it harder to navigate the discussion.

Instead, push any new changes to the pull request's branch in your fork. If you need to reset your changes, you can, for example, push a change where you just replace all files with what they were before you started your work. Thanks! :)

tests/ref/html/bibliography-no-title.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
+    <p role="doc-biblioref"><a id="loc-1" href="#loc-2">[1]</a></p>
```
willcrichton: Is it intentional that `role="doc-biblioref"` is attached to the `p` tag and not the `a` tag?

laurmaedje: No, I didn't notice that. That's an unfortunate interaction of different mechanisms. :/

crates/typst-html/src/rules.rs
```diff
@@ -445,7 +449,75 @@ const OUTLINE_ENTRY_RULE: ShowFn<OutlineEntry> = |elem, engine, styles| {
...
+
+        if let Some(mut prefix) = prefix.clone() {
+            // If we have a link back to the first citation referencing this
+            // entry, attach the appropriate role.
+            if prefix.is::<DirectLinkElem>() {
+                prefix = prefix.set(HtmlElem::role, Some("doc-backlink".into()));
+            }
+
+            let wrapped = HtmlElem::new(tag::span)
+                .with_attr(attr::class, "prefix")
```
breezykermo, line 478: Is the prefix class here just intended to allow the user to style the outputted HTML? It seems like a pretty generic class name to me that could easily conflict/collide with existing stylesheets if the generated HTML is embedded in an existing page, say. 

Would it make sense to provide a Typst prefix (excuse the pun) for class names so that this becomes something like `typst__prefix`? Maybe this is best exposed as an option when passing the HTML output flag, as I realize it might be overkill for standalone documents.

laurmaedje: Yes, this is for styling and similarly generic class names are used across other elements. Currently, only standalone output is supported. When we add support for fragment output an option for a general prefix mechanism will probably be added. (cf. https://github.com/typst/typst/pull/6602#issuecomment-3073321023)

tests/ref/html/bibliography-no-title.html
laurmaedje: Fixed.

PR #6961, opened 2025-09-25
Fix tab indentation in raw block with CRLF 

### LINKED ISSUE #6931: Tab indentation in raw blocks fails with CLRF
### Description

Hello,

I encountered this error in my project, when I tried reading out files to display them in raw blocks.
The compiler compiles just fine, but the indentation doesnt line up when using `#let x = read(filename)` with 
`#raw(x, lang: "c", block: true)`, when the file is indented with tab-characters. When using the ` ``` ` method in the typst file with the same code (and tab-indentation) the code is displayed correctly.

I tested this in a small dummy project. All files needed are below.

Greetings
Ilian

[main.pdf](https://github.com/user-attachments/files/22423240/main.pdf)
[myjfile.java](https://github.com/user-attachments/files/22423242/myjfile.java)

The `.txt` was added post compilation to allow uploading these files here.
[main.typ.txt](https://github.com/user-attachments/files/22423292/main.typ.txt)
[myfile.grv.txt](https://github.com/user-attachments/files/22423291/myfile.grv.txt)

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

Pipcount:
Fixes #6931 

Adding an extra condition to match with `\r\n` in `align_tabs` in [raw.rs](crates/typst-library/src/text/raw.rsl) seems to fix the indentation issue.


tests/suite/text/raw.typ
```diff
@@ -643,6 +643,20 @@ if true {
 }
 ```
 
+--- issue-6961-tab-clrf-raw-indent ---
```
isuffix: CLRF -> CRLF (multiple places and the PR title)

Although I do think a carriage line return feed sounds like a fun… something.

laurmaedje:
Thanks!

PR #6966, opened 2025-09-27
Add an example that number every 5th `par.line`

YDX-2147483647:
Proposed [on the forum](https://forum.typst.app/t/how-to-number-every-5th-line/1295/3) by Mc-Zen in October 2024.

图片


laurmaedje:
Thanks!

Issue #6941, opened 2025-09-20
Add inverse hyperbolic functions to `calc` module


mihirmanna:
### Description

Currently the `calc` module has functions for sine (`sin`), cosine (`cos`), tangent (`tan`), and their inverses (`asin`, `acos`, `atan`). It also has the hyperbolic sine (`sinh`), cosine (`cosh`), and tangent (`tanh`) functions, but does not have their inverses.

To that end, it would be great to introduce `asinh`, `acosh`, and `atanh` as dedicated functions!

### Use Case

Having dedicated inverse hyperbolic functions would avoid having users need to define them manually and bring them in line with the trigonometric functions.

An example use case of these functions I recently ran into was when defining an `asinh` scale for plotting data.

Enivex:
Considering how easy these are to express in terms of logarithms, and how rarely the inverses are used, I don't feel like adding them is worth it.

mihirmanna:
Hmm... perhaps the same argument could then be made against `sinh`, `cosh`, and `tanh`, regarding their exponential forms. 😄

I think that there is still an appeal for this feature, at the very least to bring it in line with the trigonometric functions, but I will leave it up to the majority opinion.

Lege19:
Is there a reason this isn't closed even now that #6967 is merged?

saecki:
> Is there a reason this isn't closed even now that https://github.com/typst/typst/pull/6967 is merged?

No, seems like an oversight

Issue #1998, opened 2023-08-25
Hyperbolic functions shouldn't take angles as input


Enivex:
### Description

Currently the hyperbolic functions accept angles as inputs

https://typst.app/docs/reference/calculate/cosh/
https://typst.app/docs/reference/calculate/sinh/
https://typst.app/docs/reference/calculate/tanh/

"Calculates the hyperbolic cosine of an angle. When called with an integer or a float, they will be interpreted as radians."

While _technically_ the number can be interpreted as something called the [hyperbolic angle](https://en.wikipedia.org/wiki/Hyperbolic_angle), this is connected to the areas for hyperbolas, and not the length of circular arcs (like the angles you are familiar with). It's therefore senseless to accept degrees as input for these functions.

(As an aside, this is why the hyperbolic inverse function is _ar_|cosh (for area) and not _arc_|cos (for arcus))

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
Fixed via https://github.com/typst/typst/pull/2072

PR #6967, opened 2025-09-27
Add inverse hyperbolic functions to `calc` module

mihirmanna:
Added dedicated `asinh`, `acosh`, and `atanh` functions to the `calc` module, in order to bring the hyperbolic functions in line with the trigonometric functions that already have dedicated inverses. Usage pattern is identical, e.g.:
```
#calc.asinh(1)
```

See #6941 for discussion.

### DIFF
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -401,6 +401,20 @@ pub fn tanh(
     value.tanh()
 }
 
+/// Calculates the inverse hyperbolic sine of a number.
+///
+/// ```example
+/// #calc.asinh(0) \
+/// #calc.asinh(1)
+/// ```
+#[func(title = "Arsinh")]
+pub fn asinh(
+    /// The number whose inverse hyperbolic sine to calculate.
+    value: Num,
+) -> Angle {
+    Angle::rad(value.float().asinh())
+}
+
 /// Calculates the logarithm of a number.
 ///
 /// If the base is not specified, the logarithm is calculated in base 10.
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -415,6 +415,24 @@ pub fn asinh(
     Angle::rad(value.float().asinh())
 }
 
+/// Calculates the inverse hyperbolic cosine of a number.
+///
+/// ```example
+/// #calc.acosh(1) \
+/// #calc.acosh(2.5)
+/// ```
+#[func(title = "Arcosh")]
+pub fn acosh(
+    /// The number whose inverse hyperbolic cosine to calculate. Must be greater than or equal to 1.
+    value: Spanned<Num>,
+) -> SourceResult<Angle> {
+    let val = value.v.float();
+    if val < 1.0 {
+        bail!(value.span, "value must be greater than or equal to 1");
+    }
+    Ok(Angle::rad(val.acosh()))
+}
+
 /// Calculates the logarithm of a number.
 ///
 /// If the base is not specified, the logarithm is calculated in base 10.
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -433,6 +433,24 @@ pub fn acosh(
     Ok(Angle::rad(val.acosh()))
 }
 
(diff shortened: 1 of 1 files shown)


tovrstra:
This would be convenient to include.

One minor suggestion: The new titles are abbreviated forms, like "Arsinh", which deviates from existing functions. For consistency, the full forms like "Inverse Hyperbolic Sine" would be better. This also avoids potential "arsinh" versus "arcsinh" confusion.

laurmaedje:
I'm fine with adding these. Like @tovrstra said, the titles should be updated though.

mihirmanna:
Done! Sorry for the delay and all. Let me know if anything else needs to be updated.

Enivex:
I don't know rust well enough to understand exactly what's going on, but I'm pretty sure they're returning angles?

They absolutely should not.

Edit: Context for the other direction <https://github.com/typst/typst/issues/1998>

crates/typst-library/src/foundations/calc.rs
```diff
@@ -401,6 +401,56 @@ pub fn tanh(
...
+/// Calculates the inverse hyperbolic sine of a number.
+///
+/// ```example
+/// #calc.asinh(0) \
+/// #calc.asinh(1)
+/// ```
+#[func(title = "Inverse Hyperbolic Sine")]
+pub fn asinh(
+    /// The number whose inverse hyperbolic sine to calculate.
+    value: Num,
```
laurmaedje: `Num` can be `f64` everywhere here since we call `value.float()` directly and ints auto-ast to floats on a language level.

mihirmanna:
Sorry for the messy commit history, I forgot that backticks don't render á la Markdown in commit messages. 

How do I format the tests? I see that many of the existing ones have certain numbers associated with the error message. Also--does the format differ for tests that are expected to fail (e.g., trying to evaluate `#arcosh(0)`) vs. succeed?

laurmaedje:
The numbers are the columns in which the error appears. If the test succeeds, no error annotations are necessary.

You can run `cargo testit` to see what's supposed to be there. (Though it will print absolute line numbers and the inline format typically avoids line numbers (defaults to next line) or uses relative ones.)

Don't worry about the commit history, it will be squashed anyway.

mihirmanna:
Hmm, I noticed that `cargo testit` can't seem to find the new functions? Or is there something I'm missing?

```
❌ calc-asinh (tests/suite/foundations/calc.typ:415)
  not annotated
    Error [paged]: 12-17 module `calc` does not contain `asinh`
...
3143 passed, 6 failed, 0 skipped
error: test failed, to rerun pass `-p typst-tests --test tests`
```

mkorje:
> Hmm, I noticed that `cargo testit` can't seem to find the new functions? Or is there something I'm missing?

You need to add them to the `module` function at the top of the `calc.rs` file, e.g. `scope.define_func::<asinh>();`

mihirmanna:
Got it! One last question: is there a way to set the precision for tests? For example:

```
#test(calc.asinh(1), calc.ln(1 + calc.sqrt(2)))
    ...
    Assertion failed: 0.881373587019543 != 0.8813735870195429
```

Or should I simply hardcode the value that it is currently returning?

laurmaedje:
Check out the `length-to-unit` test for a similar problem. There, it is solved like this:
```typ
// Test length unit conversions.
#let t(a, b) = assert(calc.abs(a - b) < 1e-6)

#t((500.934pt).pt(), 500.934)
#t((3.3453cm).cm(), 3.3453)
#t((4.3452mm).mm(), 4.3452)
#t((5.345in).inches(), 5.345)
...
```

mihirmanna:
What's the best way to deal with the failed CI check? I'm curious that other math functions don't raise the same issues, although they are also non-deterministic.

tovrstra:
Maybe replace `val.acosh()` by `libm::acosh(value)`?

See this commit for more details: https://github.com/typst/typst/commit/f20691020af4672b8dd76410a63c3674f6f1a3c3

saecki:
Regarding the new CI failures, the test runner was updated. The tests added should now be annotated with `eval` instead:
```diff
---- calc-asinh paged ---
+--- calc-asinh eval ---
```

laurmaedje:
I dealt with the remaining issues. Thanks!

PR #6975, opened 2025-09-28
Fix crash when the body of a context expression is a closure

### LINKED ISSUE #6800: Crash when calling a closure in context
### Description

The following code crashes:

```typst
#let f(fn) = {
  context (size: none) => {
    if size == none { size = text.size }
    fn(size)
  }()
}

#f((size) => [
  Text size is #size
])
```

with

```
thread 'main' panicked at crates/typst-eval/src/call.rs:242:47:
```

----

I have no experience with the compiler so my investigations were very crude and limited, however here are some things which might be related:

When the closure is parenthesised, it compiles and renders as expected

```typst
// this renders correctly
#let f(fn) = {
  context ((size: none) => {
    if size == none { size = text.size }
    fn(size)
  })()
}

#f((size) => [
  Text size is #size
])
```

A similar behaviour can be observed without context but which doesn't lead to a crash (I am unsure whether this is an expected behaviour or not, and if not, whether this ought to be a separate issue, I apologize).
Consider the following:

```typst
#{
  (() => {
    [ Foo ]
  })()
}

#{
  () => {
    [ Bar ]
  }()
}
```

This is rendered as:

> Foo
>
> (..) => ..

In fact the second closure doesn't seem to be evaluated since the code

```typst
#{
  (x) => {
    [ Bar ]
  }()
}
```

also renders as `(..) => ..` whereas

```typst
#{
  ((x) => {
    [ Foo ]
  })()
}
```

raises a missing argument error as expected.

This parenthesis behaviour is quite confusing since when the closure is stored in a variable, there is no difference in behaviour between parenthesised and not parenthesised.

```typst
#{
  let foo = () => {
    [ Foo ]
  }

  foo()
}

#{
  let foo = (() => {
    [ Bar ]
  })

  foo()
}
```

renders as

> Foo
>
> Bar

as expected.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

Daaiid:
Closes #6800 

The problem is exactly what @laurmaedje said [in this comment:](https://github.com/typst/typst/issues/6800#issuecomment-3218107307) The evaluating code can confuse a "normal" closure with a "synthetic" closure produced by a `context` expression.

The solution is to clearly differentiate the types of closures. I used an enum that wraps the `SyntaxNode` of the closure. I initially tried to make the enum members wrap a `ast::Closure` / `ast::Expr` explicitly but that would come with viral declarations of lifetimes. Since that would require major refactorings I wanted to avoid that. I couldn't see a solution without the lifetimes so I decided to keep the wrapped data as it was.

Also I struggled to make a meaningful test for this case. Since the output of this "closure-in-context" is nonsense anyway (rendered as `(..) => ..`) I decided to make a minimal reproduction and leave it at that.

Any pointers to improvements and constructive criticisms are more than welcome.


laurmaedje:
Looks like the reference images isn't fully up to date, making the test fail.

laurmaedje:
Thanks!

PR #6976, opened 2025-09-28
Fix box corner radius clamp when missing strokes

### LINKED ISSUE #6938: Removing the right side of a rounded box changes the shape of its left side
### Description

I want to draw a rounded box where the right side is removed. When the right side is present, increasing the corner radius behaves as expected: the rounded corner can completely eliminate the *straight* segment on the left side.

However, when the right side of the box is removed, the left side no longer behaves the same way. A visible straight segment remains, and no matter how much I increase the radius, the rounded corner can never fully "eat up" the straight part. Inspecting the generated SVG can also confirm this difference.

**Example Code**
```typst
#set page(width: auto, height: auto, margin: 20pt)
#set box(width: 80pt, height: 20pt, stroke: 2pt + blue)

= Expected Behavior
#box(
  radius: (left: 30pt),
)

= Unexpected Behavior
#box(
  radius: (left: 30pt),
  stroke: (right: none)
) 
```

Image

### Expected Behavior

- The left side of the shape should remain identical whether or not the right side of the box is present. 
- Whenever possible, increasing the radius should eventually allow the rounded corner to completely remove the straight segment of the short side visually.

### Reproduction URL

https://typst.app/project/r3P2UdqjbVM1qJVW3QyNiS

### Operating system

Web app

### Typst version

hooyuser:
Closes #6938 

Resolves the issue #6938 where disabling one side’s stroke prevented the adjacent corner radius from ever reaching its full extent. 

The previous implementation clamped all four radii against a single maximum derived from the thinnest stroke. To be specific, every corner radius was clamped against a single global  `max_radius = min(width, height) / 2 + min(stroke_half_widths)`. The `stroke_half_widths` come from halving each active side’s stroke thickness and substituting 0 when a side is absent. When you disabled the right stroke, its contribution became 0, so `min(stroke_half_widths)` collapsed to 0. All corners, including the left one, were now capped at just `min(width, height) / 2`.

We now derive a per-corner limiting radius instead of a global one. For each corner we combine the base half-size `min(width, height) / 2` with the tighter of the two adjacent stroke half-widths. This matches the actual geometric constraint: the corner’s outer arc is bounded by the two strokes that meet there. 

As a consequence, removing the right stroke now only reduces the top-right and bottom-right corners, and no longer influences the top-left and bottom-left corners.


saecki (approved):
This makes sense to me :+1:

laurmaedje:
Thanks!

PR #6994, opened 2025-10-01
Deprecate renamed `modern-humanities-research-association` style

PgBiel:
Depends on #6171

With the new hayagriva version, this style is deprecated in favor of `modern-humanities-research-association-notes`.


laurmaedje:
https://github.com/typst/typst/pull/6171 is merged

PR #6999, opened 2025-10-01
Expand and improve examples for array methods

Ri-Nai:
This pull request enhances the documentation for several methods in the `array` module to improve
clarity and provide better examples for users.

The following functions have been updated:

- **`dedup`**: The example has been changed to `(3, 3, 1, 2, 3)`. This new example better illustrates
that `dedup` preserves the order of the first-seen elements and does not perform any sorting, which the
previous example could have misleadingly implied.
- **`enumerate`**: Added a `for` loop example to demonstrate destructuring of the index-value pairs.
- **`fold` & `reduce`**: Added examples to clarify the accumulator pattern and unified the variable
names in the examples for consistency.
- **`split` & `intersperse`**: Added examples and explained their conceptual inverse relationship to
make their behavior easier to grasp.

### DIFF
--- a/crates/typst-library/src/foundations/array.rs
+++ b/crates/typst-library/src/foundations/array.rs
@@ -693,8 +693,6 @@ impl Array {
 
     /// Split the array at occurrences of the specified value.
     ///
-    /// This function is the conceptual inverse of [`intersperse`]($array.intersperse).
-    ///
     /// ```example
     /// #(1, 1, 2, 3, 2, 4, 5).split(2)
     /// ```
@@ -753,8 +751,6 @@ impl Array {
 
     /// Returns an array with a copy of the separator value placed between
     /// adjacent elements.
-    ///
-    /// This function is the conceptual inverse of [`split`]($array.split).
     ///
     /// ```example
     /// #("A", "B", "C").intersperse("-")


crates/typst-library/src/foundations/array.rs
```diff
@@ -734,6 +753,12 @@ impl Array {
 
     /// Returns an array with a copy of the separator value placed between
     /// adjacent elements.
+    ///
+    /// This function is the conceptual inverse of [`split`]($array.split).
```
laurmaedje: That seems untrue. Consider:
```typ
#(1, 1, 2, 3, 2, 4, 5).split(2).intersperse(2)
// -> ((1, 1), 2, (3,), 2, (4, 5))
```

The inverse rather seems to be joining with a single-element array. (Not 100% sure here either, didn't think long.)
```typ
#(1, 1, 2, 3, 2, 4, 5).split(2).join((2,))
// -> (1, 1, 2, 3, 2, 4, 5
```

Ri-Nai: Sorry, I'll delete it.

laurmaedje:
Thanks!

Issue #6913, opened 2025-09-12
`#image` does not apply rotations specified by the `Orientation` metadata tag correctly


DUOLabs333:
### Description

`#image("output.jpg")` causes the image to be displayed upside down, rather than rightside up:

<img width="512" height="432" alt="Image" src="https://github.com/user-attachments/assets/1100ff75-360f-44ed-b15c-e9281780d46f" />

This is because the raw pixel data has the logo rotated so that the opening is located to the left --- while the "Orientation" tag in the metadata tells readers to rotate the image 90 degrees counter-clockwise before displaying, Typst actually rotates it 90 clockwise, leading to an overall 180 rotation.

![Image](https://github.com/user-attachments/assets/c0b59ad9-5590-4868-a328-79d1bb91b00c)

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
I think this is fixed on main. You can try with the development version on https://typst.app/play.

laurmaedje:
Ok well, since the playground does not support files, you can't actually. But maybe you can try locally or in a proper project.

DUOLabs333:
I'm on commit `a880ac8b` --- was it fixed in a later commit?

laurmaedje:
No, it was before. Maybe it's something different then.

LaurenzV:
I used this for the implementation: https://sirv.com/help/articles/rotate-photos-to-be-upright/

I still need to confirm it, but I suspect the given information is wrong and they mixed up 6 and 8...

LaurenzV:
Yeah, value 8 is "270 degrees clockwise", but the picture on the blog suggests 90 degrees...

DUOLabs333:
Yeah, it looks like the text is wrong, but the image is right.

LaurenzV:
Basically https://github.com/typst/typst/commit/838a46db needs to be reverted, as I was misguided by just looking at the image... Whoops.

DUOLabs333:
Hmm... reverting the commit didn't fix the issue.

LaurenzV:
Right, sorry, that only works for PNG export. PDF uses a different path, and 6 and 8 would need to be swapped here: https://github.com/typst/typst/blob/88b6fc0a6ccd517b38c6534f16c88763e5545617/crates/typst-pdf/src/image.rs#L248-L250

LaurenzV:
[...]

PR #6102, opened 2025-03-27
Test all exif rotation types and fix two of them

LaurenzV:
Blocked on https://github.com/typst/typst-dev-assets/pull/5

PR #7002, opened 2025-10-02
Fix regressions in EXIF orientation handling

laurmaedje:
Fixes https://github.com/typst/typst/issues/6913

Regression introduced in https://github.com/typst/typst/pull/6102/ which was guided by https://sirv.com/help/articles/rotate-photos-to-be-upright/, but the image in that blog post is wrong. The image there displays the same information as the ASCII art on http://sylvana.net/jpegcrop/exif_orientation.html, but what this graphic is supposed to display is _what the letter F would look like if it were tagged correctly and displayed by a program that ignores the orientation tag._ (Basically, different Fs baked with rotations and an EXIF value that cancels it). What we were really looking for though is an illustration of how an image is supposed to look with no baked rotation and just the EXIF value in a program that does _not_ ignore the orientation tag. I generated 8 such images and confirmed that the Typst output now again looks the same as what Apple Preview renders.

Moreover, since the introduction of krilla, we considered the EXIF rotation twice for non-JPEG images, once by baking it into the dynamic image and once in the PDF output with a transform. For JPEG, the first one didn't happen because JPEGs don't get recoded anymore and our only test is for JPEG, hence it didn't surface. I didn't add a test for this yet because I would first need to find or construct a non-JPEG image with EXIF rotation, which is too much effort now (contribution welcome!)


eltos:
> I didn't add a test for this yet because I would first need to find or construct a non-JPEG image with EXIF rotation, which is too much effort now (contribution welcome!)

Here is a set of PNG images in all variations with respective EXIF orientation tags: [f2t_png.zip](https://github.com/user-attachments/files/22663331/f2t_png.zip)
**If interpreted correctly, all of them will show up as an upright "F"**. 
Since EXIF support for PNG is not as wide-spread as for JPEG, make sure to use a viewer that supports it, such as GIMP.

The following typst snipped will print the EXIF tags and images in a table. You could add this as a test:
```typ
#set page(width: auto, height: auto, margin: 5pt)
#table(columns: 2, ..for file in (
  "f2t_r1_upright.png",
  "f2t_r2_mirrored.png",
  "f2t_r3_180deg.png",
  "f2t_r4_mirrored_180deg.png",
  "f2t_r5_mirrored_270deg.png",
  "f2t_r6_270deg.png",
  "f2t_r7_mirrored_90deg.png",
  "f2t_r8_90deg.png",
){
  (
    str(read(file, encoding: none).at(0x85)),
    image(file, width: 10pt)
  )
})
```

| Output for Typst 13.1 | Output for main 563c8d3659e80eff6ffbea5a5a4c75115cf733da | Output for this PR ebd547e0969904a5f784dc0ff41471faa8684663
| - | - | - |
| output_typst_13 1_webapp | output_main | output_fix-exif

laurmaedje:
Thanks! I'll integrate that.

PR #7010, opened 2025-10-03
Add French translations

MDLC01:
This adds some French translations.

The actual French word for "email" would be "courriel" or "mél" but I don't think anyone uses those words in the real world.

Also, depending on where those translations are used, it might be better to use "Adresse mail" ("E-mail address") and "Numéro de téléphone" ("Phone number").

### DIFF
--- a/crates/typst-library/translations/fr.txt
+++ b/crates/typst-library/translations/fr.txt
@@ -7,5 +7,5 @@ outline = Table des matières
 raw = Liste
 page = page
 footnote = Note
-# email =
-# telephone =
+email = E-mail
+telephone = Téléphone


crates/typst-library/translations/fr.txt
```diff
@@ -6,6 +6,6 @@ heading = Chapitre
 outline = Table des matières
 raw = Liste
 page = page
-# footnote =
-# email =
-# telephone =
+footnote = Note de pas de page
```
EpicEricEE: You probably meant "bas"

MDLC01: Oh right, thanks!

reknih:
The translations are used in tooltips for `tel:` and `mailto:` links in PDF (i.e. `#link("tel:+1234567890")[Give me a call]` would result in the tooltip `Phone number +1234567890` for the link

MDLC01:
But the English translation for `telephone` is "Telephone", not "Phone number", which confuses me.

reknih:
The code works as follows:

```rs
// Contents of the tooltip for the link
let contact = url.strip_contact_scheme().map(|(scheme, stripped)| {
    // Will result in `<phone number LocalName> +123456789` in the tooltip for
    // something like `#link("tel:+123456789")`
    eco_format!("{} {stripped}", scheme.local_name_in(styles))
});
```

This is the only use for the `telephone` and `email` LocalNames. Hence, their translation should work in that use case. ~~We'll change the English translation accordingly.~~ We decided that the names work as-is.

MDLC01:
Then this can be merged as is

laurmaedje:
Thanks!

Krasjet:
hmm, "courriel" is definitely used almost everywhere in Québec. For example,

courriel in gov website

It is a [government website](https://www.premier-ministre.gouv.qc.ca/premier-ministre/joindre-pm/courriel/index.asp).

 We probably need a separate `fr-CA`?

MDLC01:
> We probably need a separate `fr-CA`?

Yes

Issue #6859, opened 2025-09-03
Add life cycle hook


TeddyHuang-00:
## What it does?

Add support for life cycle hook to execute external tools upon specific events.
For example, this could include:

- Pre-processing of input files `typst compile file.typ --before "python3 pre.py"`
  - The typst compiler will create a subprocess to run the pre-processing script before compiling.
- Post-processing of output files `typst compile file.typ --after "python3 post.py"`
  - The typst compiler will create a subprocess to run the post-processing script after compiling.
- On-demand execution based on file query `typst compile file.typ --for "<tag>:python3 tag.py --args {}"`
  - The typst compiler will first query the source file for the specified selector, and then run the corresponding script with the output of successful queries.

## Related issues:

- typst: #1056
- typst: #1746
- package: https://github.com/typst/packages/issues/1101

## How is this useful?

After researching related issues, I found that the current solution to calling external tools is through querying and manually invoking them in the command line. This can be cumbersome and error-prone, especially for complex workflows, and **will not work in watch mode**.

With the lifecycle hook support, the typst compiler can automatically manage the execution of external tools at the appropriate times, streamlining the workflow and reducing the potential for errors. This can greatly reduce the need for manual intervention and opens up new possibilities for integrating external tools/assets into the typst workflow.

## How can this potentially be negative?

While the lifecycle hook support offers many benefits, it also introduces some potential downsides:

1. **Complexity**: The added complexity of managing external tool execution may lead to harder-to-debug workflows, especially for users who are not familiar with the new system.

2. **Performance Overhead**: Automatically invoking external tools may introduce performance overhead, particularly if these tools are resource-intensive or if they are invoked frequently during the compilation process.

3. **Recursive Invocations**: If not carefully managed, the external script might change the state of the files being processed in a way that triggers additional invocations, potentially leading to infinite loops or other unintended consequences.

4. **Limited Accessibility in Web**: Users won't have the same level of access to external tools when working in web app, although it is also not possible to run `query + preprocessing` pipeline currently in the web app.

### Before-compilation hook

Automatically evaluating script content and outputing to assets like json or csv or plain text, which can be included in the typst document

### After-compilation hook

Modifying generated document, such as marking PDFs or compressing SVGs

### On-demand execution hook

Fetching (w/ cache) assets from remote sources like iconify, and using them in document.

TeddyHuang-00:
[...]

PR #7021, opened 2025-10-03
Implement a pre-compile hook for typst compile and watch

Undef-a:
This allows the user to automatically execute a script or other command before compilation starts. While it would be trivial do do this manually with `./script && typst compile` it becomes extremely useful with `typst watch`.

The script is executed immediately before compilation starts and incleded in the typst compilation time output. If the exit status is non-zero compilation will still run to allow the user to get a partial document, however a warning is printed.

In order to avoid the case where the script triggers watched files in `typst watch` mode all watchers are paused during execution. Without this infinite compilation loops were observed.

Part-Of: #6859

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -314,6 +314,10 @@ pub struct CompileArgs {
     /// apart from file names and line numbers.
     #[arg(long = "timings", value_name = "OUTPUT_JSON")]
     pub timings: Option<Option<PathBuf>>,
+
+    /// Runs the provided command before compiling.
+    #[arg(long = "pre-compile-script", value_name = "PRE_COMPILE_SCRIPT")]
+    pub pre_compile_script: Option<PathBuf>,
 }
 
 /// Arguments for the construction of a world. Shared by compile, watch, and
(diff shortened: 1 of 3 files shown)


laurmaedje:
Thanks for the PR! Unfortunately, this does not really fit with our design philosophy. 

In particular, Typst currently does not ever call external tools or scripts and I'm quite happy with that. Beyond that, this could fairly easily result in cyclic dependency scenarios with `typst watch` which I think isn't ideal.

Undef-a:
I understand the restriction on calling external tools, I'm certainly 
happy to have a lot less worry about compiling Typst code than LaTeX, 
however, I'd like to argue for a difference between calling a user 
supplied script before compilation and executing code based on the Typst 
input. The former is far less dangerous. Obviously I'll defer to your 
call here, but if this feature will not be accepted, could I recommend 
closing the associated issue as "wont fix" so others don't attempt to 
implement it?

Regarding the cyclic dependency issue, I actually hit that during 
development on my first test. I've worked around it in this PR by 
pausing and restarting the watchers immediately before/after the script. 
After that change I never received multiple compilations.

Either way, thanks for your work on Typst!

laurmaedje:
I do agree that its far less dangerous from the CLI than from Typst itself, but there are still _some_ security considerations and, more importantly, it still just does not really feel fitting with Typst to me.

I did already close the issue as not planned alongside the PR, so that should be covered.

PR #7022, opened 2025-10-04
Add `--deps` and `--deps-format` flags for emitting a dependency list

magistau:
This PR replaces `--make-deps` with two new flags:
- `--deps` — where to write the list of dependencies
- `--deps-format` — one of `zero`, `json`, `make`:
  - `make`: Makefile format, matches the behavior of `--make-deps`
    ```Makefile
    main.pdf: lib.typ /home/tau/.cache/typst/packages/preview/physica/0.9.6/physica.typ /home/tau/.cache/typst/packages/preview/physica/0.9.6/typst.toml main.typ image.png
    ```
  - `json`: JSON array of strings, each string being a (relative) path to a dependency
    ```json
    ["lib.typ","/home/tau/.cache/typst/packages/preview/physica/0.9.6/physica.typ","/home/tau/.cache/typst/packages/preview/physica/0.9.6/typst.toml","main.typ","image.png"]
    ```
  - `zero`: sequence of strings delimited by a NUL byte
    ```
    lib.typ␀/home/tau/.cache/typst/packages/preview/physica/0.9.6/physica.typ␀/home/tau/.cache/typst/packages/preview/physica/0.9.6/typst.toml␀main.typ␀image.png␀
    ```
    
Comparison of formats:

Name | Human-readable (-2..=+2) | Machine-readable (-2..=+2) | Allowed dependency paths 
---- | ------------------------ | -------------------------- | ------------------------
Makefile       | `+1` | `-1` (largely unstandardized, essentially defined by an implementation — in fact, `--make-deps` was implemented by rewriting a part of GCC in Rust, and the GCC code says something like "well GNU Make does this; note that this still breaks sometimes") | Valid Unicode without certain characters like `$`
JSON           | `0` (can be improved to `+1` by using "pretty" option) | `+1` (standard is short and concise, easy to implement in any language, used by most utilities, but still requires stuff like escape processing) | Valid Unicode
Zero/Null/text0 | `-0.5` (needs `bat -A` or similar) | `+2` (doesn't require any sort of escape parsing, can be parsed without copying) | Anything without NUL (which AFAIK is forbidden on all existing systems)

Scale used:
- (-2, +2) is Protobuf and similar (cannot be read by a human, close to a memory dump for a machine)
- (0, 0) is something like XML (still can be read by a layperson and is well-defined, but is quite verbose and requires effort from both)
- (+2, -2) is a free-form natural language description (the native tongue for humans, absolutely unreadable for machines, and even complicated NLP stuff is still imperfect)

Unresolved questions:
- [ ] What should be the default `--deps-format`? Currently it's `zero`, because it's the only lossless one, but I admit it's not the most common format. Maybe use JSON as the middle ground?
- [ ] Should JSON be minified?
- [ ] What should be the name of the NUL-delimited format? It's too simple to have a standard (the only possible disagreement is whether the file should include a trailing NUL, but I haven't yet seen a tool not including one), so there's no standard name. I chose `zero`, because `null` can be confused for "lack of value", and `text0` is a bit longer and includes combines letters and digits, which a part of my brain doesn't like for some reason.

(continuation of #6648)

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -273,10 +273,20 @@ pub struct CompileArgs {
     #[arg(long = "ppi", default_value_t = 144.0)]
     pub ppi: f32,
 
-    /// File path to which a Makefile with the current compilation's
-    /// dependencies will be written.
-    #[clap(long = "make-deps", value_name = "PATH")]
-    pub make_deps: Option<PathBuf>,
+    /// File path to which a list of current compilation's dependencies will be written
+    #[clap(
+        long,
+        value_name = "PATH",
+        value_parser = output_value_parser(),
+        value_hint = ValueHint::FilePath,
+    )]
+    pub deps: Option<Output>,
+
+    /// File format to use for dependencies
+    ///
+    /// Note that JSON disallows non-Unicode paths, and Make disallows even more than that.
+    #[clap(long, default_value_t)]
+    pub deps_format: DepsFormat,
 
     /// Processing arguments.
     #[clap(flatten)]
@@ -467,6 +477,16 @@ pub enum OutputFormat {
 }
 
 display_possible_values!(OutputFormat);
+
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, ValueEnum)]
+pub enum DepsFormat {
+    #[default]
+    Zero,
+    Json,
+    Make,
+}
+
+display_possible_values!(DepsFormat);
 
 /// The target to compile for.
 #[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, ValueEnum)]
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -1,4 +1,5 @@
 use std::fmt::{self, Display, Formatter};
+use std::io::Write;
 use std::num::NonZeroUsize;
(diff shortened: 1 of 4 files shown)


chekoopa:
Shall the old option exist for some time as deprecated?

laurmaedje:
I wanted to include this in 0.14 since #6648 was being ignored for so long, but didn't have the time for review back and forth in time for the release, so I refactored this myself. In particular, I also readded the old option as deprecated.

Unfortunately, I could not push to the PR as usual (the relevant permission was disabled). If you want, you can push the commits I pushed to https://github.com/typst/typst/tree/deps-format to this PR and I'll merge later today. Otherwise, I'll have to create a new PR and close this one.

---

Regarding your unresolved questions, I made the following decisions.

> What should be the default --deps-format? Currently it's zero, because it's the only lossless one, but I admit it's not the most common format. Maybe use JSON as the middle ground?

Yes, I prefer JSON.

> Should JSON be minified?

I think minified is appropriate.

> What should be the name of the NUL-delimited format?

Zero is fine.

magistau:
> Unfortunately, I could not push to the PR as usual (the relevant permission was disabled). If you want, you can push the commits I pushed to https://github.com/typst/typst/tree/deps-format to this PR and I'll merge later today.

Fixed

laurmaedje:
Thanks! Also for the patience with all of this. I'm now quite happy with the final result.

PR #7024, opened 2025-10-04
Update Irish (Gaeilge) translation

aindriu80:
Added missing Irish translations for Typst and corrected one existing translation.

### DIFF
--- a/crates/typst-library/translations/ga.txt
+++ b/crates/typst-library/translations/ga.txt
@@ -7,5 +7,5 @@ outline = Clár Ábhair
 raw = Amhábhar
 page = Leathanach
 footnote = Nóta Coise
-email = Seoladh ríomhphoist
-telephone = Uimhir theileafóin
+email = Ríomhphost
+telephone = Teileafón


reknih:
Also see https://github.com/typst/typst/pull/7010#issuecomment-3370848439

aindriu80:
I’ve read your explanation regarding the tooltip usage for `telephone` and `email`.   I’ve updated the Irish translations accordingly to ensure they read naturally in context:

 * `telephone = Uimhir theileafóin`
 * `email = Seoladh ríomhphoist`

 These now produce tooltips like *“Uimhir theileafóin +3531234567”* and *“Seoladh ríomhphoist [example@site.ie](mailto:example@site.ie)”*

laurmaedje:
For consistency with the other translations, I think it would have to be just "theileafóin" and "ríomhphoist" (probably captialized?). The "number" and "address" are also not contained in the other translations. We've opted for the more generic term in the end.

crates/typst-library/translations/ga.txt
```diff
@@ -7,5 +7,5 @@ outline = Clár Ábhair
 raw = Amhábhar
 page = Leathanach
 footnote = Nóta Coise
-email = Seoladh ríomhphoist
-telephone = Uimhir theileafóin
+email = Ríomhphost
+telephone = Teileafón
```
laurmaedje, line 11: is it expected that the capitalized spelling is quite different than from before? very interesting if that's how Irish works.

aindriu80:
That makes perfect sense,  thank you for the explanation and review! 

I understand now that the LocalName strings are meant to stay generic for consistency across all languages.

I’ve updated the Irish translations accordingly to:

`telephone = Teileafón`

`email = Ríomhphost`

crates/typst-library/translations/ga.txt
aindriu80: In Irish, the spelling can change depending on grammar,  for example, teileafón (telephone) might appear as theileafóin in certain cases (like “an theileafóin”).
When we capitalise it, we use the base form Teileafón, which can look quite different from the lowercase version.

Also, unlike English, Irish doesn’t capitalise every word in a title , only the first word or proper names.

(i don't know if that is what you are asking)

laurmaedje:
Thanks!

crates/typst-library/translations/ga.txt
laurmaedje: yeah, that was what I was asking :)

PR #7026, opened 2025-10-05
 Update Dutch translation

Fevol:
This PR translates the new terms introduced by #6619, and fixes one mistake from the translations imported by #6852.

### footnote
Translated as 'voetnoot', the original 'voetnote' is not a valid Dutch word. 
('voetnote' might be a literal translation of the English 'footnote', or accidentally copied from [Afrikaans](https://gesellig.co.za/m/136737/) - though I couldn't find an official source for this.)

### email
Translated as 'e-mail', following the [style rules](https://taaladvies.net/over/) set out by taaladvies.net:
- [Always starts with lowercase](https://taaladvies.net/e-mail-hoofdletter/)
- [Always written with a hyphen](https://taaladvies.net/email-of-e-mail/)

### telephone
Translated as 'telefoon'. According to https://github.com/typst/typst/pull/7010#issuecomment-3365373684, this could also be changed to 'telefoonnummer', if that is the intended meaning.


Enivex:
For some reason there are two translations available in the cleveref source. One under Crefname@preamble, and one under Crefname. I have no idea why they're different.

@eltos

reknih:
> Translated as 'telefoon'. According to https://github.com/typst/typst/pull/7010#issuecomment-3365373684, this could also be changed to 'telefoonnummer', if that is the intended meaning.

See https://github.com/typst/typst/pull/7010#issuecomment-3370848439

laurmaedje:
Thanks!

PR #7031, opened 2025-10-06
Update Danish translations for email and telephone

JeppeKlitgaard:
Changed page from "side" to "Side" to match case of other entries

### DIFF
--- a/crates/typst-library/translations/da.txt
+++ b/crates/typst-library/translations/da.txt
@@ -7,5 +7,5 @@ outline = Indhold
 raw = Liste
 page = side
 footnote = Fodnote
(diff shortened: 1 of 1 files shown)


laurmaedje:
Page is intentionally lowercase in all other languages because that matches how it is typically used.

JeppeKlitgaard:
> Page is intentionally lowercase in all other languages because that matches how it is typically used.

My apologies - in retrospect I realise I only compared against the German translation where Seite would obviously be capitalised 🤦.

Should be good-to-go now

laurmaedje:
Thanks!

PR #7034, opened 2025-10-06
Add `--ignore-embedded-fonts` option

### LINKED ISSUE #7018: allow typst/cli to ignore embedded fonts
### Description

allow for an option like `--ignore-embedded-fonts` that ignore the embedded font analog to `--ignore-system-fonts`

Tytanic has this too if we're looking for a precedent

https://discord.com/channels/1054443721975922748/1184433451554840636/1423640017812848661

### Use Case

my usecase would be to see a font-list unpolluted by neither system nor internal font, 
... ie. only the fonts in the given font-path.

another usecase i see is ... newer versions of internal fonts could be tested with current binaries without rebuilding.

bluebear94:
Closes #7018.

I decided to add a `IncludeFontsConfig` struct to reduce the number of `#[cfg(feature = "embed-fonts")]`s in the code. I put the `FontArgs` impl block in the `fonts` module of `typst-cli` since the `args` module is also imported by the build script and I wanted to avoid adding `typst-kit` as a build dependency.


crates/typst-kit/src/fonts.rs
```diff
@@ -85,37 +107,33 @@ impl Fonts {
...
+    include_config: IncludeFontsConfig,
     book: FontBook,
     fonts: Vec<FontSlot>,
 }
 
 impl FontSearcher {
     /// Create a new, empty system searcher. The searcher is created with the
     /// default configuration, it will include embedded fonts and system fonts.
-    pub fn new() -> Self {
+    pub fn new(include_config: IncludeFontsConfig) -> Self {
```
laurmaedje, line 118: I think I'd prefer not to have this extra struct just to save a bit of cfg. Keeping the public API lean is more important imo. Also, I'd like to avoid the breaking change here.

bluebear94:
Alright, then, it’ll probably be easier to open a new PR.

PR #7035, opened 2025-10-06
Simplify `LinkedNode::{prev,next}_sibling`

isuffix:
By moving the index into the iterator struct, we avoid the $O(n)$ `.nth()` calls in `{next,prev}_sibling`.

This was taken from the first commit in #7003

### DIFF
--- a/crates/typst-syntax/src/node.rs
+++ b/crates/typst-syntax/src/node.rs
@@ -757,20 +757,32 @@ impl LinkedNode<'_> {
 
     /// Get the first previous non-trivia sibling node.
     pub fn prev_sibling(&self) -> Option<Self> {
-        let mut children = self.parent()?.children();
-        // TODO: When stabilized, use `.advance_back_by(children.len() - self.index)`
-        children.iter.nth_back(children.len() - (self.index + 1));
-        children.back = self.offset;
-        children.rev().find(|node| !node.kind().is_trivia())
+        let parent = self.parent.as_ref()?;
+        let children = parent.node.children().as_slice();
+        let mut offset = self.offset;
+        for (index, node) in children[..self.index].iter().enumerate().rev() {
+            offset -= node.len();
+            if !node.kind().is_trivia() {
+                let parent = Some(parent.clone());
+                return Some(Self { node, parent, index, offset });
+            }
+        }
+        None
     }
 
     /// Get the next non-trivia sibling node.
     pub fn next_sibling(&self) -> Option<Self> {
-        let mut children = self.parent()?.children();
-        // TODO: When stabilized, use `.advance_by(self.index + 1)`
-        children.iter.nth(self.index);
-        children.front = self.offset + self.node.len();
-        children.find(|node| !node.kind().is_trivia())
+        let parent = self.parent.as_ref()?;
+        let children = parent.node.children();
+        let mut offset = self.offset + self.len();
+        for (index, node) in children.enumerate().skip(self.index + 1) {
+            if !node.kind().is_trivia() {
+                let parent = Some(parent.clone());
+                return Some(Self { node, parent, index, offset });
+            }
+            offset += node.len();
+        }
+        None
     }
 
     /// Get the kind of this node's parent.


isuffix:
Rebased off main.

Akida31:
(I'm not sure that the `.nth()` calls are O(n) and not just O(1): [`.nth()` calls `post_inc_start`](https://doc.rust-lang.org/stable/src/core/slice/iter/macros.rs.html#205) which is just a [pointer add](https://doc.rust-lang.org/stable/src/core/slice/iter/macros.rs.html#108). But I think this is still a nice cleanup!)

isuffix:
Ah right! I bet I was thinking that `parent.node.children()` was calling `LinkedNode::children() -> LinkedChildren` instead of `SyntaxNode::children() -> std::slice::Iter`. The first would still be $O(n)$, but not the second because of the iterator specialization.

crates/typst-syntax/src/node.rs
```diff
@@ -945,14 +946,13 @@ impl<'a> Iterator for LinkedChildren<'a> {
...
-                parent: Some(self.parent.clone()),
-                index,
-                offset: self.back,
-            }
+        let node = self.iter.next_back()?;
+        self.back -= node.len();
+        Some(LinkedNode {
+            node,
+            parent: Some(self.parent.clone()),
+            index: self.index + self.iter.len(),
```
laurmaedje: This is fairly intricate and was previously nicely hidden in the impl of `Enumerate`.

crates/typst-syntax/src/node.rs
```diff
@@ -926,16 +927,16 @@ impl<'a> Iterator for LinkedChildren<'a> {
...
-            let offset = self.front;
-            self.front += node.len();
-            LinkedNode {
-                node,
-                parent: Some(self.parent.clone()),
-                index,
-                offset,
-            }
-        })
+        let node = self.iter.next()?;
```
laurmaedje: The use of the question mark operator here and in `next_back` is nice.

crates/typst-syntax/src/node.rs
```diff
@@ -753,22 +756,19 @@ impl LinkedNode<'_> {
...
     pub fn prev_sibling(&self) -> Option<Self> {
-        let parent = self.parent()?;
-        let index = self.index.checked_sub(1)?;
-        let node = parent.node.children().nth(index)?;
-        let offset = self.offset - node.len();
-        let prev = Self { node, parent: self.parent.clone(), index, offset };
-        if prev.kind().is_trivia() { prev.prev_sibling() } else { Some(prev) }
+        let mut children = self.parent()?.children();
+        children.iter = children.iter.as_slice()[..self.index].iter();
+        children.back = self.offset;
```
laurmaedje: Mutating the iterator's fields here feels a bit icky to me

laurmaedje (commented):
I like some, but not all of the changes. I think relying on `Enumerate` to do the right thing was nicer before as it keep some of the intricate logic in std instead of here. 

The fact that it's now iterative instead of recursive is nicer, but I think we could (probably?) fairly easily just write a while loop instead in the sibling functions. Assigning directly into the internal fields of the iterator there doesn't feel great to me.

isuffix:
Rebased off main and then added some extra comments.

isuffix:
I've updated this to keep the enumerated iterator in `LinkedChildren` and just use `.nth()` and `.nth_back()` since we know they're $O(1)$. However, I've left TODO comments to move to [`Iterator::advance_by`](https://github.com/rust-lang/rust/issues/77404) when that stabilizes. This means that the only change to `LinkedChildren` is replacing `.map()` with the question-mark operator in `next()` and `next_back()`.

This is now ready for review.

crates/typst-syntax/src/node.rs
isuffix: I think this is ok. In the new version, we're just advancing the iterator and updating the `front` or `back` fields accordingly. I find this much clearer than the existing code.

crates/typst-syntax/src/node.rs
```diff
@@ -757,20 +757,34 @@ impl LinkedNode<'_> {
...
-        let mut children = self.parent()?.children();
-        // TODO: When stabilized, use `.advance_by(self.index + 1)`
-        children.iter.nth(self.index);
-        children.front = self.offset + self.node.len();
-        children.find(|node| !node.kind().is_trivia())
+        let parent = self.parent.as_ref()?;
+        let children = parent.node.children().as_slice();
+        let mut index = self.index + 1;
+        let mut offset = self.offset + self.len();
+        for node in &children[index..children.len()] {
```
laurmaedje:
```suggestion
        for node in &children[index..] {
```

laurmaedje: Could even do `for (index, node) in children.iter().enumerate().skip(self.index + 1)` to make it exactly parallel to the other one. But if you prefer the current one, I'm totally fine with that, too.

laurmaedje (commented):
That's by far the most readable version to me. I like it!

Sometimes a for loop is all one needs. :p

crates/typst-syntax/src/node.rs
isuffix: Ah, I had tried that spelling but wasn't confident that `skip` would be $O(1)$ so reverted. But looking closer at its implementation, it seems to use `.nth` internally, so I'll move to it. Thanks

laurmaedje:
Thanks!

PR #7036, opened 2025-10-06
Add tests from #7003

isuffix:
This adds some of the tests from #7003 that I think are worth keeping in the test suite.


tests/suite/math/call.typ
```diff
@@ -92,6 +92,35 @@ $args(#..range(1, 5).chunks(2))$
 #let func(body) = body
 $func(...)$
 
+--- math-call-spread-empty ---
+// Test that a spread operator followed by nothing generates two dots.
+// TODO: This currently errors in parsing, but it should change to the given repr.
```
laurmaedje, line 97: is there an issue for this?

isuffix: Not that I know of, but I’m working on a smaller PR to fix a few different errors in math function parsing like this.

laurmaedje:
Thank you!

PR #7037, opened 2025-10-06
Add `--ignore-embedded-fonts` option

### LINKED ISSUE #7018: allow typst/cli to ignore embedded fonts

bluebear94:
Supersedes #7034 (no structification).

Fixes #7018.


PgBiel (commented):
Looks good to me. By the way, next time please just push new changes to your existing PR. You can reset your PR with a command like `git reset --hard origin/main` (or `upstream/main` if you add Typst as the `upstream` remote etc.), and then you can push your changes anew without closing the PR. (There is also the dumb simple solution of making a new commit where you just drag all files from Typst's latest main to your repository and push, to a similar effect.)

The reason is that, otherwise, the review history is lost, making the review process harder and longer, especially if you create a new PR multiple times. Thanks!

laurmaedje:
Thanks!

Issue #7007, opened 2025-10-02
Add support for PDF/A2-a and PDF/A3-a


LaurenzV:
### Description

I believe they should be relatively straightforward (it might even just be a matter of exposing it in typst?) with what we currently have, but I haven't looked into it in detail.

### Use Case

So that all export modes are supported.

laurmaedje:
What about PDF/A-1a?

PR #7038, opened 2025-10-07
Expose the remaining PDF/A standards

laurmaedje:
Closes https://github.com/typst/typst/issues/7007


LaurenzV:
Have you tested whether we pass veraPDF validation when exporting with those?

laurmaedje:
It is still in progress.

saecki:
This should probably also update this check in the cli:
https://github.com/typst/typst/blob/b3446a1322f5c1590cbd018865945ee02ebeafa1/crates/typst-cli/src/compile.rs#L134

PR #7043, opened 2025-10-07
Turn PDF images into SVGs in HTML and SVG export

laurmaedje:
Using [`hayro-svg`](https://crates.io/crates/hayro-svg). Huge props to @LaurenzV once again.


Cargo.toml
```diff
@@ -62,8 +62,9 @@ fs_extra = "1.3"
 rustc-hash = "2.1"
 glidesort = "0.1.2"
 hayagriva = "0.9.1"
-hayro-syntax = "0.4.0"
 hayro = { version = "0.4.0", default-features = false }
+hayro-svg = "0.2.0"
```
LaurenzV: You have to disable the default features here as well.

LaurenzV (approved):
Nice, tested with my thesis and seems to work nicely! Not for this PR, but I just noticed I would also suggest adding a warning as follows when embedding a PDF:
```
pdf.xref().has_optional_content_groups() {
            // Warning: the PDF contains optional content groups.
            // Hint: The embedded image might display wrongly in PDF export
            // Hint: Flatten the PDF before embedding it.
        }
```

See https://github.com/LaurenzV/hayro/issues/141, while it works now for SVG/PNG export, it won't work for PDF export, so might be good to emit a warning at least. However, at least from my testing Latex respect OCGs either, so I think it shouldn't be too big of a problem.

laurmaedje:
> Not for this PR, but I just noticed I would also suggest adding a warning as follows when embedding a PDF:

Alright, I'll put up a separate PR!

PR #7044, opened 2025-10-07
Update Chinese translations

lzm0:
Add translations in Simplified and Traditional Chinese for terms introduced in #6619.


laurmaedje:
Thanks!

PR #7048, opened 2025-10-07
Add note about `args` module in `typst-cli`

bluebear94:
Encountered during initial work for the `--ignore-embedded-fonts` flag.


laurmaedje:
Thanks!

PR #7049, opened 2025-10-07
Disallow empty URLs

reknih:
In addition to URLs that are too long, this commit disallows empty URLs in links. This addresses two PDF/UA-1 validation issues found by veraPDF on the old test `link-empty-block`: Clause 7.18.1 (test 2) and Clause 7.18.5 (test 2). Both stipulate that a non-empty contents key must be present on annotations. Because we set the `Contents` key to the URL, empty URLs fail this test. But they don't make sense anyways, so we just forbid them here.

### DIFF
--- a/crates/typst-library/src/model/link.rs
+++ b/crates/typst-library/src/model/link.rs
@@ -308,6 +308,8 @@ impl Url {
         let url = url.into();
         if url.len() > 8000 {
             bail!("URL is too long")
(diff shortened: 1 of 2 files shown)


LaurenzV:
Couldn't this check land directly in krilla?

LaurenzV:
But I suppose it makes sense to have this in Typst as well.

PR #7050, opened 2025-10-07
Add Italian translations for email and telephone

Bugg4:
Add Italian translation for terms introduced in #6619.


laurmaedje:
Thanks!

PR #7056, opened 2025-10-08
Warn for optional content groups in PDF images

laurmaedje:
See https://github.com/typst/typst/pull/7043#pullrequestreview-3310373637

@LaurenzV I reworded the hints a bit because just "Flatten the PDF before embedding it." might be a bit obtuse for a user. Do you think it's clear enough like this?

### DIFF
--- a/crates/typst-library/src/visualize/image/mod.rs
+++ b/crates/typst-library/src/visualize/image/mod.rs
@@ -273,22 +273,6 @@ impl Packed<ImageElem> {
         let loaded = &self.source.derived;
         let format = self.determine_format(styles).at(span)?;
 
-        // Warn the user if the image contains a foreign object. Not perfect
-        // because the svg could also be encoded, but that's an edge case.
-        if format == ImageFormat::Vector(VectorFormat::Svg) {
-            let has_foreign_object =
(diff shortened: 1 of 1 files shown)


LaurenzV:
LGTM!

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -300,6 +284,18 @@ impl Packed<ImageElem> {
...
             ),
             ImageFormat::Vector(VectorFormat::Svg) => {
+                // Warn the user if the image contains a foreign object. Not
+                // perfect because the svg could also be encoded, but that's an
+                // edge case.
+                if memchr::memmem::find(&loaded.data, b"<foreignObject").is_some() {
+                    engine.sink.warn(warning!(
+                        span,
+                        "image contains foreign object";
+                        hint: "SVG images with foreign objects might render incorrectly in typst";
```
MDLC01, line 294:
```suggestion
                        hint: "SVG images with foreign objects might render incorrectly in Typst";
```

As per https://github.com/typst/typst?tab=readme-ov-file#pronunciation-and-spelling

PR #1990, opened 2023-08-24
Preserve release name and body on asset upload

jimvdl:
Small PR preventing the release workflow from overriding the release title and erasing the change log.

From [`ncipollo/release-action@v1`](https://github.com/ncipollo/release-action#action-inputs):

`omitNameDuringUpdate`:
Indicates if the release name should be omitted during updates. The name will still be applied for newly created releases. _This will preserve the existing name during updates._

`omitBodyDuringUpdate`:
Indicates if the release body should be omitted during updates. The body will still be applied for newly created releases. _This will preserve the existing body during updates._

PR #7057, opened 2025-10-08
Avoid release action from updating prerelease status

laurmaedje:
The action changes the status of the GitHub release from prerelease to latest when updating artifacts, which has to be undone manually.


laurmaedje:
Followup to https://github.com/typst/typst/pull/1990

PR #7060, opened 2025-10-08
Output OS and ISA arch for typst info

reknih:
Helpful to differentiate between WSL and native Windows, Docker and native macOS, x86 vs x86_64...


crates/typst-cli/src/main.rs
```diff
@@ -98,6 +98,16 @@ fn typst_commit_sha() -> &'static str {
...
+/// The operating system we are running on.
+const fn typst_os() -> &'static str {
+    std::env::consts::OS
+}
+
+/// The instruction set architecture we are running on.
+const fn typst_arch() -> &'static str {
+    std::env::consts::ARCH
+}
+
```
LaurenzV: Won't this just yield the data of the platform where the binary was compiled instead of where it's running? Otherwise how could it be a `const`?

reknih: True, but that's the intended behavior (telling you what machine you're on would be pointless). I adjusted the doc comment.

PR #7062, opened 2025-10-08
Normalize selectors when functions are provided.

### LINKED ISSUE #4741: nesting show rule on a heading doesn't show
### Description

trying to write a nested show rule for an outline with a heading target doesn't work.

```typst
#show outline.where(target: heading): it => {
    show outline.entry.where(level: 1): set text(fill: red)
    it
}

#outline(target: heading)


#outline(target: figure)


= first
== second 
= first again
== second 
=== third

#figure(caption: [bar], [foo])
```
does not seem to work, while adding the `heading` inside a `selector()` makes it work.

as found in [discord thread](https://discord.com/channels/1054443721975922748/1272658912688078900)

as of typst 0.11.0 (d84a7e42) the bug still persists.

### Reproduction URL

_No response_

### Operating system

Web app, Linux

### Typst version

raldone01:
The selector `outline.where(target: heading)` was not working for me.

```typ
// Test that show rules with outline.where(target: ...) work correctly
#show outline.where(target: heading): it => block(
  fill: rgb("e8f3ff"),
  inset: 5pt,
  [*Custom TOC* #it],
)

#show outline.where(target: figure): it => block(
  fill: rgb("ffe8f3"),
  inset: 5pt,
  [*Custom Figures* #it],
)

#outline() // Without this fix the show rule is not applied here

#outline(target: figure, title: [Figures]) // Without this fix the show rule is not applied here

#show heading: none
= First
== Second

#show figure: none
#figure(rect(width: 20pt, height: 10pt), caption: [A])
```

It does not break any tests but I don't know if this changes selector behavior to something that is not wanted.

Fixes #4741.


laurmaedje:
There are other cases where types are normalized in style properties. I don't want to ship a "half fix" and would rather tackle the problem exhaustively. For this reason, I'll close the PR. Thank you still!

PR #7065, opened 2025-10-08
Typst 0.14 changelog

laurmaedje:
There are a ton of great changes in this one, so making a selection for the highlights section was tricky!

This is the first changelog that links out to the relevant PRs for each item, as was requested on Discord a while ago. At first I was skeptical because I feared that it would make the changelog visually noisy, but we found a nice solution: The links are automatically turned into a small chain icons (that are easy to ignore) by the docs CSS.


MDLC01:
I think it would be good to mention on the Highlights section that the changes to script precedence might break documents. This is a significant breaking change that will not always cause a warning, and some users might not read the whole changelog. The highlight should provide an example (such as `e^f(x)`) and how to fix it (`e^(f(x))`). I believe the change is major enough to warrant an extra bullet point in the highlights (I would even argue it should be the first bullet point because it is a breaking change).

docs/changelog/0.14.0.md
```diff
@@ -0,0 +1,297 @@
...
+  - For instance, `[$f_pi(x)$]` is now the same as `[$f_pi (x)$]` instead of `[$f_(pi(x))$]`
+  - Functions like `abs` now require explicit parentheses for disambiguation when used in sub- or superscripts: `[$e^(abs(x))$]`
+  - See the [Forum post on the topic](https://forum.typst.app/t/opinions-needed-likely-upcoming-changes-to-math-mode-precedence/4974) for more details
+- Shorthands and numbers do not bind more tightly than fractions anymore in cases like `[$x>=(y)/z$]` [#5925] [#5996] **(Minor breaking change)**
+- Added [`frac.style`]($math.frac.style) property with new options for skewed and inline fractions [#6672]
+- Added [`math.equation.alt`] property for setting an alternative description for an equation [#6619]
+- Text handling
+  - A single equation can now use multiple fonts [#6365]
+  - Glyph layout in math now uses proper text shaping, leading to better handling of more complex Unicode features [#6336]
+  - Generated letters in an equation (e.g. the `√` produced by `sqrt`) can now be targeted by text show rules [#6365]
```
MDLC01:
```suggestion
  - Generated characters in an equation (e.g. the `√` produced by `sqrt`) can now be targeted by text show rules [#6365]
```

`√`, mentioned right after, is not a _letter_.

docs/changelog/0.14.0.md
```diff
@@ -0,0 +1,297 @@
...
+  - Fixed a bug with vertical accent positioning [#5941]
+  - Fixed positioning of bottom accents [#6187]
+  - Fixed a bug with layout of roots [#6021]
+  - Improved layout of [`vec`]($math.vec) and [`cases`]($math.cases), making it consistent with [`mat`]($math.mat) [#5934]
+  - Removed linebreak opportunity before closing bracket in inline math [#6216]
+  - Tweaked sizing of stretchables (e.g. brackets and arrows) [#6377]
+- An [`augment`]($math.mat.augment) line can now exist at the start and end of a matrix, not only in between columns and rows [#5806]
+- Named arguments passed to symbols used as function now raise an error instead of being silently ignored [#6192] **(Minor breaking change)**
+- The [`mid`]($math.mid) element does not force the [`{"large"}` math class]($math.class) upon its contents anymore and instead defaults to `{"relation"}` [#5980]
+- Fixed the default math class of ⅋, ⎰, ⟅, ⎱, ⟆, ⟇, and `،` [#5949] [#6537]
```
MDLC01: Why is the Arabic comma the only one in backticks here?

laurmaedje:
In principle, I intentionally omitted the breaking change label from the highlights. I'm open to expanding the explanation for the precedence change a bit even at the top, but I definitely don't want to mention it first in the highlights.

laurmaedje:
Maybe I'll add a dedicated "Migrating" section instead that is a condensed form of what will go into the release blog post. I just don't think the highlights section is the right place and in particularly the breakage should not overshadow the new features there.

docs/changelog/0.14.0.md
```diff
@@ -0,0 +1,297 @@
...
+  - The [`image`] element [#6578]
+  - The [`footnote`] and [`footnote.entry`] element [#6917]
+  - The [`outline`] and [`outline.entry`] element [#6606]
+  - The [`bibliography`] element [#6952]
+  - The [`smartquote`] element [#6710]
+  - The [`sub`] and [`super`] elements [#6422]
+  - The [`underline`], [`overline`], [`strike`], and [`highlight`] elements [#6510]
+  - The [`smallcaps`] element [#6600]
+  - The [`lower`] and [`upper`] functions [#6585]
+- Added typed HTML API (e.g. [`html.div`]) with individually typed attributes [#6476]
```
isuffix, line 60: I feel like the typed HTML API is a big enough change to deserve a full example. Here's a draft one:
```suggestion
- Added typed HTML API with individually typed attributes [#6476]
  - For example, to generate a `time` element you can now write `#html.time(datetime: datetime.today())` instead of `#html.elem("time", attrs: (datetime: datetime.today().display())`. Note how the `datetime` attribute no longer requires converting to a string with `.display()`.
```

docs/changelog/0.14.0.md
laurmaedje: Not intentional.

hpcfzl:
Small typo `unapplicable` should be `inapplicable`.

PR #7067, opened 2025-10-08
Improve documentation for lists' `children` parameters

MDLC01:
This tries to clarify the documentation and error messages related to `list.children`, `enum.children`, and `terms.children`, especially the expected form of arrays.

Alternatively, I think it could make sense to remove the casts from `array` for `enum.item` and `terms.item` altogether, as I don't think they are particularly useful.

### DIFF
--- a/crates/typst-library/src/model/enum.rs
+++ b/crates/typst-library/src/model/enum.rs
@@ -191,7 +191,7 @@ pub struct EnumElem {
     /// The numbered list's items.
     ///
     /// Items can be specified as their bodies, as `(number, body)` pairs, or
-    /// via explicit [`{enum.item}`] elements.
+    /// via explicit [`enum.item`] elements.
     ///
     /// ```example:"Basic usage"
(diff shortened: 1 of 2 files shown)


laurmaedje:
> Alternatively, I think it could make sense to remove the casts from array for enum.item and terms.item altogether, as I don't think they are particularly useful.

I didn't even remember those were a thing! :D

Probably from before `enum.item` even being an element. Perhaps we should actually leave them undocumented in preparation for removal.

MDLC01:
> Perhaps we should actually leave them undocumented in preparation for removal.

Why that instead of removing them? I don't think they are widely used, and it's an easy fix anyway.

laurmaedje:
It's a bit too close to release now for my liking to make spontaneous breaking changes and I'd also rather deprecate them first. It just doesn't seem urgent.

MDLC01:
Then I'll close this PR and I can reopen it after 0.14 is released.

Issue #2844, opened 2023-12-03
Add syntax highlighting for the code inside a `raw` block if the `lang` is set


Andrew15-5:
### Description

I have a file `typ.typ`:

````typ
= Hello

```typ
#let a = 5
```

```py
a = 5
```
````

I have a file `b.typ`:

```typ
#raw(read("./typ.typ"), lang: "typ")
```

When I run `typst c b.typ` I get:

![image](https://github.com/typst/typst/assets/37143421/03c20435-15d3-4ae4-b094-4a15e187c1fc)

Which feels like a missed opportunity to use the power of Typst that knows the syntax of many languages, and therefore, in theory, can add syntax highlighting inside of the `raw` block.

The problem is that the layout engine see the `raw` block as a whole element, therefore it is colored in one single color. And it means that _something_ has to be changed in order to allow Typst to colorize the insides of the `raw` block (if it knows the language that was set).

### Use Case

Typst already outputs a great `lang: "typ"` syntax highlighted text/code. It also can add colors to `lang: "py"` or any other language. Therefore, it would be nice to see the Typst being able to recursively (or at least 1 level deep) colorize the content of a `raw` block

Doesn't this look much cooler?

![image](https://github.com/typst/typst/assets/37143421/e4d59ee3-b5d1-4f91-922c-f2f7e3c64229)

Andrew15-5:
What is needed to implement the new highlighting behavior for the text inside the `raw` element?

Andrew15-5:
I again face this issue. I don't like that nested code blocks look so much better in my editor than in Typst output. It makes the pretty and more readable.

Andrew15-5:
`bat` does highlight `py` inside `typ`:

<img width="233" height="472" alt="Image" src="https://github.com/user-attachments/assets/6f0e44f2-a785-424d-8398-a2e3fdaff7f8" />

<details><summary>highlight.typ</summary>

`````typ
#highlight[]
#let a = 5
```text
raw code
```

```py
print()
```

````typ
#highlight[]
#let a = 5
```text
raw code
```

```py
print()
```
````
`````

</details>

And since `bat` and `two-face` uses the same source, I'm hoping that the issue is within the Typst codebase, and is not an upstream issue.

Andrew15-5:
[...]

PR #7070, opened 2025-10-09
Support nested raw block syntax highlighting

Andrew15-5:
This is PoC for what I've tried in https://github.com/typst/typst/issues/2844#issuecomment-3382978015

Closes https://github.com/typst/typst/issues/2844.

<details><summary>highlight.typ</summary>

``````typ
#highlight[]
#let a = 5
```text
raw code
```

`let`

`1`

`print()`

```py print()```

```py
def function(a: str, b: int) -> list[str | int]:
    return [a, b]
```

`````typ
#highlight[]
#let a = 5
```text
raw code
```

`print()`

```py print()```

```py
def function(a: str, b: int) -> list[str | int]:
    return [a, b]
```

````typ
#highlight[]
#let a = 5
```text
raw code
```

`print()`

```py print()```

```py
def function(a: str, b: int) -> list[str | int]:
    return [a, b]
```
````
`````


```typ
= Chapter 1
#lorem(100)

#let hi = "Hello World"
#show heading: emph
```
``````

</details>

MuPDF:

image

<details><summary>Okular</summary>

image

</details>

<details><summary>Unmodified (bright)</summary>

image

</details>

Yeah, it definitely works, but not perfectly. I can see now that 3rd-party syntax parsing doesn't give the same result, moreover if a new syntax is introduced, you'd have to wait for others to create a new syntax file, which is not ideal.

For some reason nested/typst highlighting doesn't work at all with `two-face` 0.4.3 or 0.4.4, but does work, if I clone it locally, and generate the syntax binaries. Thankfully, it _also_ works with 0.5.0-rc1. Which is weirdly 4 months old, while 0.4.4 was released 10 days ago? I guess it's in the development for a long time, but 0.4 has some critical patches or something?

While I was playing around, I noticed that the outer backticks turned black, but inside ones were standard gray. Not only them, but also everything else that doesn't have a styling scope matched, so it looks bad. Though I was able to fix it by modifying a few entries in the embedded theme.

<details><summary>Gray everywhere</summary>

Using:

```rs
        // item("punctuation.definition.raw", Some("#6b6b6f"), None),
        item("markup.raw", Some("#6b6b6f"), None),
```

image

</details>

What I wasn't able to fix is separating `raw.lang`'s scope from `constant` scope. I tried `constant.other.language-name.typst` and some others, but it didn't work (maybe I was using some outdated version or something). So it is highlighted just like number `5`, which looks off, especially when I look at what I see in my Neovim:

image

Another thing that I've noticed is that Python's syntax looks off. I assume its `.sublime-syntax` is not great, as `-> list[str | int]` shouldn't have red between blacks.

That's all for the weird edge cases.

---

I was actually trying to merge the two if bodies, but they are different enough, that I had to add a ton of random fields to `ThemedHighlighter` and I quickly got overwhelmed by the mess I created.

Although thinking about it again, I feel like it's still possible, I just need to correctly update `self.range` and some other stuff when delegating syntax highlighting of other languages to Syntect. Though all the `theme`, `syntax_set`, `foreground`, etc. were a major headache, not sure how to reuse all this stuff. I think some major reconstruction of the logic must be done, preserving some core logic, but making it much, much, much easier to use syntect + two-face when going in the Typst branch. Basically something like:

```rs
impl Packed<RawElem> {
    #[comemo::memoize]
    fn highlight(&self, routines: &Routines, styles: StyleChain) -> Vec<Packed<RawLine>> {
        if lang == typst {
            highlight_typst(...)
        } else if find_lang(lang).is_some() {
            highlight_other(...)
        } else {
            do_not_highlight(...)
        }
    }
} // ?
fn highlight_typst(...) {
    // ...
    if this.kind() == Raw {
        let result = if lang == typst {
            highlight_typst(...)
        } else {
            highlight_other(...)
        } else {
            do_not_highlight(...)
        }
    }
    // ...
}
fn highlight_other() {}
fn do_not_highlight() {}
```

Since `highlight` has a big preamble now, it would be hard to reduce it to:

```rs
impl Packed<RawElem> {
    #[comemo::memoize]
    fn highlight(&self, routines: &Routines, styles: StyleChain) -> Vec<Packed<RawLine>> {
        if lang == typst {
            self.highlight_typst(...)
        } else if find_lang(lang).is_some() {
            self.highlight_other(...)
        } else {
            self.do_not_highlight(...)
        }
    }
    fn highlight_typst(...) {
        // ...
        if this.kind() == Raw {
            let result = if lang == typst {
                self.highlight_typst_inner(...)
            } else {
                self.highlight(...)
            }
        }
        // ...
    }
    fn highlight_typst_inner() {}
    fn highlight_other() {
        // maybe if md and found Typst inside, somehow give control (back) to
        // highlight_typst. Seems kinda impossible, or very hard.
    }
    fn do_not_highlight() {}
}
```


laurmaedje:
From the description and the diff it's not at all clear to me what this is supposed to do and why it's already a PR. The diff seems to just delete the Typst highlighting code? And the description is not really to the point.

Andrew15-5:
> it's not at all clear to me what this is supposed to do

Close https://github.com/typst/typst/issues/2844?

> The diff seems to just delete the Typst highlighting code?

See https://github.com/typst/typst/issues/2844#issuecomment-3382978015.

---

> Yeah, it definitely works, but not perfectly. I can see now that 3rd-party syntax parsing doesn't give the same result, moreover if a new syntax is introduced, you'd have to wait for others to create a new syntax file, which is not ideal.

> I was actually trying to merge the two if bodies, but they are different enough, that I had to add a ton of random fields to `ThemedHighlighter` and I quickly got overwhelmed by the mess I created.

I realized that simply removing the first branch isn't good enough. So I think that I still have to merge 2nd and 1st branch logic into 1st branch, to get the high-quality hybrid coloring for Typst code and anything nested in it. But that will have to take some effort, and I simply don't have time right now. Maybe in a week or two.

laurmaedje:
I think draft PRs have their place for early discussion, but opening a draft PR with a huge description and an unclear diff just creates confusion and triage effort. In that case, it would be better to wait with opening the PR.

laurmaedje:
I'll close this for now. Feel free to open a new PR once this is in a state where it's ready for discussion.

PR #7071, opened 2025-10-09
Make `typst-ide` and the introspector more HTML-friendly

elegaanz:
This change required some adjustments to make `Document` dyn-safe. It will allow to use IDE functionalities on HTML documents too.

### DIFF
--- a/crates/typst-ide/src/jump.rs
+++ b/crates/typst-ide/src/jump.rs
@@ -94,10 +94,11 @@ mod jump_from_document_sealed {
                 let reached_leaf_node = i == indices_count - 1;
                 match current_node {
                     HtmlNode::Element(html_element) => {
-                        let (mut child, child_index) = html_element
+                        let (child_index, (mut child, _)) = html_element
                             .children
                             .iter_with_dom_indices()
-                            .find(|(child, dom_index)| {
+                            .enumerate()
+                            .find(|(_, (child, dom_index))| {
                                 !matches!(child, HtmlNode::Tag(_)) && dom_index == index
                             })?;
 


crates/typst-ide/src/complete.rs
```diff
@@ -1121,11 +1122,14 @@ struct CompletionContext<'a> {
     seen_casts: FxHashSet<u128>,
 }
 
-impl<'a> CompletionContext<'a> {
+impl<'a, D> CompletionContext<'a, D>
```
laurmaedje: Could the completion context remain non-generic (keeping `dyn Document`) or was there some problem with this?

elegaanz: Maybe I'm missing something, but since `CompletionContext::new` takes a `&(impl Document + ?Sized)` it can't convert it to a `&dyn Document`.

crates/typst-html/src/link.rs
```diff
@@ -154,11 +156,18 @@ fn traverse_frame(
         match item {
             FrameItem::Tag(Tag::Start(elem, _)) => {
                 let loc = elem.location().unwrap();
-                if targets.contains(&loc) {
-                    let pos = identificator.introspector.position(loc).point;
+                if targets.contains(&loc)
+                    && let DocumentPosition::Paged(Position {
+                        point: Point { x, y }, ..
+                    })
```
laurmaedje: is this first case reachable? it seems like this should always be a frame position

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+        fn find_in_elem(
+            source: &Source,
+            elem: &HtmlElement,
+            span: Span,
+            current_position: EcoVec<usize>,
+        ) -> Vec<HtmlPosition> {
+            let mut result = Vec::new();
+            let mut i = 0;
+            let mut was_text = false;
+            let mut text_start = Span::detached();
```
laurmaedje: could was_text and text_start form an `Option<Span>` together?

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+            for child in &elem.children {
+                match child {
+                    HtmlNode::Element(e) => {
+                        let mut position = current_position.clone();
+                        position.push(i);
+                        for pos in find_in_elem(source, e, span, position) {
+                            result.push(pos)
+                        }
+                        i += 1;
```
laurmaedje: could this be moved to the end of the loop body?

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+            current_position: EcoVec<usize>,
+        ) -> Vec<HtmlPosition> {
+            let mut result = Vec::new();
+            let mut i = 0;
+            let mut was_text = false;
+            let mut text_start = Span::detached();
+            for child in &elem.children {
+                match child {
+                    HtmlNode::Element(e) => {
+                        let mut position = current_position.clone();
```
laurmaedje: it seems possible to use a single mutable vector for the depth-first-traversal and only cloning when we have a hit

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+
+                        i += 1;
+
+                        if *node_span == span {
+                            let span_range = source.range(text_start);
+                            return vec![HtmlPosition {
+                                element: current_position,
+                                inner: Some(InnerHtmlPosition::Character(
+                                    span_range.unwrap_or_default().start,
+                                )),
```
laurmaedje: are you sure this is correct? it reads to me as if the inner character position will now be the start of the range of the text node in the source Typst file. I don't understand how these are related.

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+                        if let Some(frame_pos) = find_in_frame(&frame.inner, span) {
+                            let mut position = current_position.clone();
+                            position.push(i);
+                            return vec![HtmlPosition {
+                                element: position,
+                                inner: Some(InnerHtmlPosition::Frame {
+                                    x: frame_pos.x,
+                                    y: frame_pos.y,
+                                }),
+                            }];
```
laurmaedje: this does not properly interrupt was_text, right?

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -493,3 +499,61 @@ impl IntrospectorBuilder {
...
+    pub element: EcoVec<usize>,
+    /// Precise position inside of the specified element.
+    pub inner: Option<InnerHtmlPosition>,
+}
+
+/// Precise position inside of an HTML node.
+#[derive(Clone, Debug, Hash)]
+pub enum InnerHtmlPosition {
+    /// If the node is a frame, the coordinates of the position.
+    Frame { x: Abs, y: Abs },
```
laurmaedje: This could be `Frame(Point)`

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -493,3 +499,61 @@ impl IntrospectorBuilder {
...
+impl TryFrom<DocumentPosition> for HtmlPosition {
+    type Error = ();
+
+    fn try_from(value: DocumentPosition) -> Result<Self, Self::Error> {
+        match value {
+            DocumentPosition::Html(position) => Ok(position),
+            _ => Err(()),
+        }
+    }
+}
```
laurmaedje: these impls seem a bit unidiomatic to me, especially with `type Error = ()`. if this is a common need, perhaps we should add `as_paged(&self) -> Option<Position>` and `as_html(&self) -> Option<&HtmlPosition>` instead?

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -493,3 +499,61 @@ impl IntrospectorBuilder {
...
+#[derive(Clone, Debug, Hash)]
+pub enum InnerHtmlPosition {
+    /// If the node is a frame, the coordinates of the position.
+    Frame { x: Abs, y: Abs },
+    /// If the node is a text node, the index of the character at the position.
+    Character(usize),
+}
+
+#[derive(Clone, Debug, Hash)]
+pub enum DocumentPosition {
```
laurmaedje, line 580: _might_ be worthwhile to rename `Position` -> `PagedPosition` and `DocumentPosition`, but we can still do so later, so I'd leave it for now. but could you please add brief doc comments here?

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -415,18 +421,18 @@ impl IntrospectorBuilder {
     }
 
     /// Handle a tag.
-    pub fn discover_in_tag(
+    pub fn discover_in_tag<P: Into<DocumentPosition>>(
         &mut self,
         sink: &mut Vec<Pair>,
         tag: &Tag,
-        position: Position,
+        position: P,
```
laurmaedje: I think the generic parameter doesn't really pull its weight here since this is not a public-facing convenience API and only called in two places.

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+            })
+            .collect()
+    }
+}
+
+impl JumpInDocument for HtmlDocument {
+    type Position = HtmlPosition;
+
+    fn find_span(&self, source: &Source, span: Span) -> Vec<Self::Position> {
+        fn find_in_elem(
```
laurmaedje: could move this function outside? I think the extra indent is a bit annoying.

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
     None
 }
 
+pub trait JumpInDocument {
+    type Position;
+
+    fn find_span(&self, source: &Source, span: Span) -> Vec<Self::Position>;
```
laurmaedje: ideally, these methods would not be public. it's a bit annoying, but I think I would go for the sealed trait pattern here.

crates/typst-ide/src/jump.rs
```diff
@@ -144,12 +261,106 @@ pub fn jump_from_click(
...
+        }
+
+        find_in_elem(source, &self.root, span, EcoVec::new())
+    }
+}
+
 /// Find the output location in the document for a cursor position.
-pub fn jump_from_cursor(
-    document: &PagedDocument,
+pub fn jump_from_cursor<D: JumpInDocument>(
```
laurmaedje, line 341: could you reorder this file so that public/general comes before private/helper/specific? e.g. jump_from_cursor before its helper traits, before its helper methods.

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+            source.range(span).unwrap_or_default().start
+                + match (is_text_node, &position.inner) {
+                    (true, Some(InnerHtmlPosition::Character(char))) => *char,
+                    _ => 0,
+                },
+        ))
+    }
+}
+
+pub fn jump_from_click<D: JumpFromDocument>(
```
laurmaedje: could you make the parameter order and naming consistent with the other functions? please also add a bit of documentation as this is the main entrypoint of the API

crates/typst-ide/src/jump.rs
```diff
@@ -42,9 +155,11 @@ pub fn jump_from_click(
             return Some(match dest {
                 Destination::Url(url) => Jump::Url(url.clone()),
                 Destination::Position(pos) => Jump::Position(*pos),
-                Destination::Location(loc) => {
-                    Jump::Position(document.introspector.position(*loc))
-                }
+                Destination::Location(loc) => Jump::Position(
+                    document.introspector().position(*loc).try_into().unwrap_or(
+                        Position { page: NonZeroUsize::ONE, point: Point::zero() },
```
laurmaedje: perhaps this would even make sense as a method on `DocumentPosition`? the behavior is admittedly a bit strange, but while we have it, we might as well centralize it.

crates/typst-ide/src/jump.rs
```diff
@@ -220,9 +423,9 @@ mod tests {
     use std::borrow::Borrow;
     use std::num::NonZeroUsize;
 
-    use typst::layout::{Abs, Point, Position};
+    use typst::layout::{Abs, PagedDocument, Point, Position};
```
laurmaedje, line 488: could you add some tests for the HTML jumps here?

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+    fn resolve_position(
+        &self,
+        world: &dyn IdeWorld,
+        position: &Self::Position,
+    ) -> Option<Jump> {
+        let mut current_node: &HtmlNode = &HtmlNode::Element(self.root.clone());
+        for index in &position.element {
+            match current_node {
+                HtmlNode::Element(html_element) => {
+                    if let Some(ch) = nth_child(*index, html_element) {
```
laurmaedje: could use `?` here

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+                        current_node = ch;
+                    } else {
+                        return None;
+                    }
+                }
+                HtmlNode::Tag(_) | HtmlNode::Text(_, _) | HtmlNode::Frame(_) => {}
```
laurmaedje: shouldn't this return None instead of just silently ignoring the extra indices?

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+                &frame.inner,
+                Point::new(*x, *y),
+            );
+        }
+
+        Some(Jump::File(
+            id,
+            source.range(span).unwrap_or_default().start
+                + match (is_text_node, &position.inner) {
+                    (true, Some(InnerHtmlPosition::Character(char))) => *char,
```
laurmaedje: the name `char` is not ideal as it may highlight like a type. also, since it's an index maybe something like `i` would be clearer.

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+                    }
+                }
+                HtmlNode::Tag(_) | HtmlNode::Text(_, _) | HtmlNode::Frame(_) => {}
+            }
+        }
+
+        let span = current_node.span();
+        let id = span.id()?;
+        let source = world.source(id).ok()?;
+        let ast_node = LinkedNode::new(source.root()).find(span);
```
laurmaedje: you can use `source.find` directly

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+        let is_text_node =
+            ast_node.map(|x| x.is::<typst::syntax::ast::Text>()).unwrap_or(false);
```
laurmaedje: this is just `is_some_and` right? I wonder why clippy does not complain. or am I thinking wrong?

crates/typst-ide/src/jump.rs
```diff
@@ -27,10 +31,119 @@ impl Jump {
...
+        if was_text && is_text {
+            i -= 1;
+        }
+
+        if i == n {
+            return Some(ch);
+        }
+
+        was_text = is_text;
+        i += 1;
```
laurmaedje: the -= 1 and += 1 cancel strategy here seems slightly convoluted. I wonder whether this couldn't be simplified a bit, e.g. by moving the `i += 1` above, but skipping it in the double text case? basically only counting plus one at the start of the next child rather than the end of the current child.

crates/typst-html/src/document.rs
```diff
@@ -126,60 +126,98 @@ fn introspect_html(
...
+        let mut index = 0;
         for node in nodes {
             match node {
                 HtmlNode::Tag(tag) => {
+                    current_position.push(index);
                     builder.discover_in_tag(
                         sink,
                         tag,
-                        Position { page: NonZeroUsize::ONE, point: Point::zero() },
+                        HtmlPosition { element: current_position.clone(), inner: None },
```
laurmaedje: storing a full vector here for every tag might become expensive. we don't necessarily have to do something about this now, but worth discussing. perhaps also an indication that we should turn the html position into a type with private fields + an iterator method, so that we are more free to adjust the representation later (for example, into a tree structure with shared parts).

crates/typst-html/src/document.rs
```diff
@@ -126,60 +126,98 @@ fn introspect_html(
...
+                                DocumentPosition::Html(HtmlPosition {
+                                    element: current_position.clone(),
+                                    inner: Some(InnerHtmlPosition::Frame {
+                                        x: point.x,
+                                        y: point.y,
+                                    }),
+                                }),
+                            ))
+                        }
+                    }
```
laurmaedje: perhaps we should adjust discover_frame to let it take a callback function instead of a sink so that we can emit the correct kinds of positions immediately?

crates/typst-html/src/document.rs
```diff
@@ -126,60 +126,98 @@ fn introspect_html(
...
                 }
             }
         }
     }
 
     let mut elems = Vec::new();
     let mut builder = IntrospectorBuilder::new();
-    discover(&mut builder, &mut elems, link_targets, output);
+    let mut current_position = EcoVec::new();
+    discover(&mut builder, &mut elems, link_targets, output, &mut current_position);
```
laurmaedje, line 194: the number of mutable parameters here is reaching a point where a struct might be in order ...

crates/typst-html/src/document.rs
```diff
@@ -231,34 +269,99 @@ fn head_element(info: &DocumentInfo) -> HtmlElement {
...
+            if *tag != tag::body {
+                needs_body = true;
+            }
+
+            needs_html = true;
+        }
+
+        if !needs_body {
+            // If the user supplied their own <body> or <html>, check that they don't use footnotes.
+            FootnoteContainer::unsupported_with_custom_dom(engine)?;
```
laurmaedje: I think this could be moved into an else clause on the `if needs_body` below, which would be a bit clearer imo

crates/typst-html/src/document.rs
```diff
@@ -231,34 +269,99 @@ fn head_element(info: &DocumentInfo) -> HtmlElement {
...
-enum OutputKind {
-    /// The user generated their own `<html>` element. We do not need to supply
-    /// one.
-    Html(HtmlElement),
-    /// The user generate their own `<body>` element. We do not need to supply
-    /// one, but need supply the `<html>` element.
-    Body(HtmlElement),
-    /// The user generated leaves which we wrap in a `<body>` and `<html>`.
-    Leaves(EcoVec<HtmlNode>),
+    let preceding_tags_end = output.iter().position(not_a_tag).unwrap_or(0);
```
laurmaedje: this can use `split_prefix_suffix` from `typst_utils`

crates/typst-html/src/document.rs
```diff
@@ -231,34 +269,99 @@ fn head_element(info: &DocumentInfo) -> HtmlElement {
...
+    let following_tags_start = output
+        .iter()
+        .rposition(not_a_tag)
+        .map(|x| x + 1)
+        .unwrap_or(output.len());
+    let non_tag_count = following_tags_start - preceding_tags_end;
+    let mut needs_body = false;
+    let mut needs_html = false;
+
+    if non_tag_count == 0 {
```
laurmaedje: probably I'm missing something but why we do we need this more complex classification method. it appears to me that the output of all the logic (up to `if needs_body`) is only `needs_body`, `needs_html`, and `preceding_tags_end`. couldn't the first two be determined just the same through the old [`classify_output`](https://github.com/typst/typst/blob/f25a0817f0215a7931f243f4ccfa67f480ef3a24/crates/typst-html/src/document.rs#L237-L254) code?

And the `preceding_tags_end` could be extracted into the HTML case (if we add an enumerate to the for loop). Though fwiw, we could also just return a clone of the root node instead. Though then we should make sure to drop the `tags_and_root` before mutating the root (to avoid copies due to ref-count > 1).

crates/typst-html/src/document.rs
```diff
@@ -231,34 +269,99 @@ fn head_element(info: &DocumentInfo) -> HtmlElement {
...
-                elem.span,
-                "`{}` element must be the only element in the document",
-                elem.tag,
-            ),
-            _ => {}
+/// Wrap the user generated HTML in <html>, <body> or both if needed.
+///
+/// Returns a vector containing outer introspection tags and the HTML root element.
+/// A direct reference to the root element is also returned.
+fn prepare_document(
```
laurmaedje: perhaps `finalize_dom` would be a clearer name?

crates/typst-html/src/document.rs
```diff
@@ -126,60 +126,98 @@ fn introspect_html(
...
-    let body = match output {
-        OutputKind::Html(element) => return element,
-        OutputKind::Body(body) => body,
-        OutputKind::Leaves(leaves) => HtmlElement::new(tag::body).with_children(leaves),
-    };
-    HtmlElement::new(tag::html).with_children(eco_vec![head.into(), body.into()])
-}
-
 /// Generate a `<head>` element.
 fn head_element(info: &DocumentInfo) -> HtmlElement {
```
laurmaedje, line 257: could you move this below prepare_document?

crates/typst-html/src/document.rs
```diff
@@ -42,6 +43,15 @@ pub fn html_document(
     )
 }
 
+macro_rules! unwrap_elem {
```
laurmaedje: possibly, this won't be needed anyways with the other comments I've left, but if it is, I think this should be a function (and at the bottom) rather than a macro.

crates/typst-ide/src/jump.rs
elegaanz: I'm not sure what you meant. I can move the whole `match` arm to the end if that's what you want.

laurmaedje: I meant only the `i += 1`

crates/typst-ide/src/jump.rs
elegaanz: This function exists as `Location::position`, but it takes a `&mut Engine` while it could just take an `Introspector`, which would allow to use it here. Can I change its signature, given it's part of the public API?

laurmaedje: Sorry, I should have been clearer. I only meant the 
```rs
try_into().unwrap_or(Position { page: NonZeroUsize::ONE, point: Point::zero() })
```
part.

So just a pure helper function that extracts an absolute position or points to page 1, point 0 for HTML.

crates/typst-ide/src/jump.rs
elegaanz: It shouldn't be incremented for tags. I can move it to the end and `continue` when encountering a tag, if you feel like it's better.

laurmaedje: Yeah, I think continuing for tags makes a lot of sense

crates/typst-html/src/document.rs
elegaanz: Because the function is recursive, the compiler is not able to infer a type for the closure, as it generates a fresh type for each recursive call, resulting in an infinite type. I tried to add a `Sink` structure as an abstraction, but it resulted in more complexity than clarity in my opinion so I decided to drop it.

crates/typst-html/src/document.rs
elegaanz: Even if `classify_output` was short, the way it was written was a bit confusing to me (the fact that it iterates over all nodes to then early-return in case we know there is only one element, in particular) and I tried a completely different approach at first, which made me delete it and not re-introduce it in this last iteration of the algorithm. I can re-introduce it and use it if you prefer.

crates/typst-html/src/document.rs
elegaanz: In case there are no tags the result would be different I think.

crates/typst-html/src/document.rs
laurmaedje: I'm not particularly attached to the old implementation. The thing I don't like about the new one is that the handling of body and html elements happens in multiple places, whereas I think with the old approach it could be in just one place.

crates/typst-html/src/document.rs
laurmaedje: You might be able to fix this by passing `&mut F` instead of `F` directly:
```rust
fn main() {
    call_recursively(10, &mut |i| println!("{i}"));
}

fn call_recursively<F>(i: usize, f: &mut F)
where 
    F: FnMut(usize)
{
    if i > 0 {
        f(i);
        call_recursively(i - 1, f);
    }
}
```

PR #7072, opened 2025-10-09
Math mode implicit function call update and parser refactor

isuffix:
This is a refactor of the existing math parser to enable parsing implicit math function calls for non-identifiers in new circumstances: for example, we currently parse `1/f(x)` as `1/(f(x))`, but `1/f'(x)` as `1/f' (x)`. This changes the latter to parse as `1/(f'(x))`.

This is related to the recent [precedence changes in math](https://forum.typst.app/t/opinions-needed-likely-upcoming-changes-to-math-mode-precedence/4974) and PRs #7003 and #6442, but was particularly spurned by a [discord discussion](https://discord.com/channels/1054443721975922748/1054443722592497796/1425567440158326784) with Laurenz.

The way we do it is that instead of inferring an implict call by starting at a "continuable" node and encountering an opening delimiter, we now look backwards from every opening delimiter to see if the previous leaf node was continuable, then infer the implicit call. This was how I did this in the runtime math parser implementation, although it was easier then because we had a flat list of syntaxnodes instead of having to traverse the previous node's last child until we get a leaf.

However, after implementing it, I'm not certain this is entirely desirable. The original basis for this in the runtime math parser was mostly accidental, and I likely would have avoided this if I were thinking of it. That said, I can see some appeal, especially for simple cases like `f'(x)/g'(x)`, but it starts to get weird when things like `sin^2(x)/cos^2(x)` don't work.

Being truly honest, I'd still prefer to parse fractions at a higher precedence than juxtaposition in general (e.g. require parentheses for even `f(x)/g(x)`). I think it's more honest to what Typst is actually doing operationally, and I particularly think that it composes better when compared to some of the weird cases from implicit function calls. It makes for a simple rule. But it wouldn't really be ergonomic without runtime math parsing for true function calls, so I digress.

Here's a reference of the added test before and after this change:
```typ
$
  // Basic example:
  f'(x) / f_pi[|x} \
  // These ones _don't_ act like implicit function calls
  sin^2(x) / f_0(x) quad f!(x) / g^(-1)(x) \
  // These work except for embedded code
  a_\u{2a}[x] / a_"2a"{x|] quad f_pi.alt{x} / f_#math.pi.alt{x} \
  // Attachments and primes
  a(b)_c(d)^e(f) / g(h)'_i(j)' \
  // This one is cute yet horrible. Understanding it will help with the others
  (x)'(x)'(x)' / (x)'(x)'(x)' \
$
```

<table>
<thead>
<tr><th>Before</th><th>After</th></tr>
</thead>
<tr>
<td></td>
<td></td>
</tr>
</table>

Note that only the final commit implements the change. The prior commits are cleanup of the existing code to make the final commit feasible. Note that the `MathPrimes` + `Bang` refactor commit is taken from #7003.

I know this PR would be better if I could only include the final change, but it was actually incredibly tricky to get right, and cleaning up the parser's code significantly helped in getting it done. Part of what makes this change so tricky is that implicit function calls in math act like a special kind of juxtaposition operator, but the current parser doesn't have any notion of a juxtaposition operator at all (as opposed to my runtime math parser), so adding just a touch of that understanding without refactoring _everything_ required a lot of precision. I do think it's correct, but I'm only confident in saying so because I could analyse the cleaner code.


laurmaedje:
Thanks for taking this on on such short notice!

### Comments
- The first three commits look like exact equivalence transformations to me. 👍

- The fourth commit seems to introduce a _slight_ change in behavior: The new stop set is passed down the hierarchy, while the old stop variable was local. This affects consecutive attachments, i.e. in the case `$ f_1_2^3 $`. I am honestly not sure which behavior is more desirable. Regardless of the change in behavior, the refactor itself seem very sound and like a great improvement.

- I think how continuability is determined is not ideal in terms of UX. In particular that `$ 1 / f_a(x) $` behaves differently than `$ 1 / f_1(x) $`. I think my core point is that continuability of an attach element should be determined by the base and not the sub- or superscript as that is also what is continued visually. See the table below for how this would play out in practice in the "My Preference" column.

- One thing that I don't quite understand is the purpose of the code after "If we're already parsing an implicit function, [...]". Does this exist to avoid a bad case or is it intended to retain some previous behavior?

- Regarding this:

  > Being truly honest, I'd still prefer to parse fractions at a higher precedence than juxtaposition in general

  I see the appeal from an implementation and consistency point of view, but I think regardless of Typst function calls, making the math function calls in `$ f(x) / g(x) $` work is worth it in itself. If that rendered as `$ f x/g (x) $`, I believe that would be even more confusing to users than the current rendering of `f_i(x)`.

- This is not directly related to your changes, but I wonder whether the precedence of `√` is still appropriate. The behavior in this case seems worse on main than in 0.13.1: `$ 1 / f_√x $`. That's because #6442 increased the precedence of attachments, but did not adjust the check `if min_prec < 3` for roots. We should probably centralize all numbers that act as precedences in constants (like you already started to).

## Continuability

Below is a table of how different snippets behave in 0.13.1, main, this PR, and finally what my personal preference (given the constraint of consistency) would be. The difference compared to this PR is that the base determines continuability of attach elements instead of the last leaf.

Math Attachment

### Notes

- 1, 2, 3 clearly
- 4 is not distinguishable from 3 and requiring a space is consistent with our rule to require a space for implied multiplication in other cases
- 5 clearly (although an argument could be made to not distinguish based on numbers versus letters, this is fairly common and would be very breaking)
- 6 clearly (that's why we need the PR in the first place)
- 7 I would argue should behave like this to be consistent with 2
- 8 must behave the same as 7; similarly to 4, I think it is fair
- 9 and 10 are similar to 5 really. I think the core point (also with 7) is that the continuability should depend on the _base_ of the attachment and not the sub- or superscript as this is also what is visually attached to

Enivex:
I really don't understand why we want to keep the same frustrating behavior with fractions as we used to have for attachments. It's really the exact same problem.

If anything, the thing that needs fixing is that `1/f(x+y)` renders as `1/(f(x+y))`

It may interfere slightly less often, but it does interfere nonetheless.

isuffix:
Closing for #7084. I'll open a new PR for the math parser cleanup commits that can happen later.

PR #7075, opened 2025-10-09
Bump Rust to 1.91 in CI

saecki:
I just wanted to make the clippy warnings go away, and didn't pay too much attention to the surroundings. There were a few inconsitencies already. Also looking at the entire code base, there are a lot more occurences of `n % 2 == 0` which don't emit warnings, and I'm not sure why.
I'm also not entirely convinced this makes the code easier to read, to the extend where I would consider disabling this clippy hint.

### DIFF
--- a/crates/typst-library/src/text/font/color.rs
+++ b/crates/typst-library/src/text/font/color.rs
@@ -34,6 +34,8 @@ pub fn should_outline(font: &Font, glyph: &Glyph) -> bool {
 ///
 /// The glyphs are sized in font units, [`text.item.size`] is not taken into
 /// account.
+///
+/// [`text.item.size`]: crate::text::TextItem::size
 #[comemo::memoize]
 pub fn glyph_frame(font: &Font, glyph_id: u16) -> (Frame, bool) {
     let upem = Abs::pt(font.units_per_em());


laurmaedje:
Wanted to just update CI to 1.90, but turns out that the next release is already in two days. So let's just wait for 1.91.

saecki:
But do we still want to enable this lint and apply these somewhat inconsistent changes?

I'd either want to go all in and change even more places using an inverted `is_multiple_of` check, or disable this lint.

crates/typst-library/src/layout/page.rs
```diff
@@ -647,7 +647,7 @@ impl Binding {
         match self {
             // Left-bound must swap on even pages
             // (because it is correct on the first page).
-            Self::Left => number.get() % 2 == 0,
+            Self::Left => number.get().is_multiple_of(2),
             // Right-bound must swap on odd pages
             // (because it is wrong on the first page).
             Self::Right => number.get() % 2 == 1,
```
saecki: I think here it makes the code more inconstistent with the second match arm

laurmaedje:
I'm okay with disabling the lint. `x % 2 == 0` is pretty obvious in what it does. Feel free to make that change. Could also fix the remaining 1.91 lints while you're at it. 😄

crates/typst-library/src/text/font/color.rs
```diff
@@ -34,6 +34,8 @@ pub fn should_outline(font: &Font, glyph: &Glyph) -> bool {
 ///
 /// The glyphs are sized in font units, [`text.item.size`] is not taken into
 /// account.
+///
+/// [`text.item.size`]: crate::text::TextItem::size
```
saecki, line 38: I'm not actually sure if this is correct, do you remember @elegaanz?

elegaanz: If you are asking about the whole doc comment, it is correct IIRC (and from what I can see). The glyph frame is transformed to be rendered at the correct size and not in font units when actually used. By the way, in the only place this function is called, the second item of the returned tuple is discarded, maybe we could drop it and just return the frame.

PR #7079, opened 2025-10-09
Mention that PDF image's version must be <= export version in docs

reknih:


crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -87,13 +87,20 @@ pub struct ImageElem {
...
+    ///
+    /// - When exporting to PDF, any PDF image file used must have a version
+    ///   equal to or lower than the [export target PDF
+    ///   version]($pdf/#pdf-versions).
+    /// - PDF files as images are currently not supported when exporting with a
+    ///   specific PDF standard, like PDF/A-3 or PDF/UA-1. In these cases, you
+    ///   can instead use SVGs to embed vector images.
+    /// - You cannot use PDFs that are password-protected or have any other kind
+    ///   of encryption. Note that your PDF may be encrypted even if you do not
+    ///   need a password to open it in a viewer, e.g. when it is signed.
```
LaurenzV: By the way, this is outdated, only actually password-protected PDFs are not supported, normal encryption with empty password works.

LaurenzV:
Hmmm... I just realized I am not sure what happens if you include a PDF that already has tags into a Typst-generated PDF that is also tagged. We do just discard the tag tree from the old PDF (since we just include it as an XObject), but marked content sequences in the content stream will not be discarded since we literally just copy-paste it. I hope that won't cause any issues...

reknih:
I think the MCIDs within can collide with ours and then we are fucked

reknih:
Should be easily reproducable by using one Tagged PDF produced by main into another

PR #7082, opened 2025-10-09
Update German translations

blinkerfluessigkeit:
- Email is a material (vitreous enamel in English)
- Bibliographie is the old spelling and we should use the new spelling for consistency (otherwise Telefon would have to be Telephon)

### DIFF
--- a/crates/typst-library/translations/de.txt
+++ b/crates/typst-library/translations/de.txt
@@ -1,11 +1,11 @@
 figure = Abbildung
 table = Tabelle
 equation = Gleichung
-bibliography = Bibliographie
+bibliography = Bibliografie
 heading = Abschnitt
 outline = Inhaltsverzeichnis
 raw = Listing
 page = Seite
 footnote = Fußnote
-email = Email
+email = E-Mail
 telephone = Telefon


andreasKroepelin:
I've never heard "Liste" for a piece of code. I think there is just no proper German word for it.

blinkerfluessigkeit:
I compared the translations from Spanish and French and then adjusted the German translations accordingly. But it seems I misunderstood something. On Wikipedia, "Programmausdruck" is used, maybe we should use that?

cady-b:
'Programmausdruck' could sadly also mean 'program expression': https://de.wikipedia.org/wiki/Ausdruck_(Programmierung)#:~:text=Ausdr%C3%BCcke%20in%20Programmiersprachen. There really doesn't seem to be a good term for this, but all would generally be understood in context.

A random programming book I have laying around here ([Lehrbuch der Programmierung mit Java, dpunkt Verlag](https://search.worldcat.org/title/76101311?oclcNum=76101311)) uses "Programm" (abbreviated as "Prog.") as their supplement for listings, but that's quite software-specific.

I do agree with the changes to email and bibligraphy though; those are the current recommended spellings by the Duden (most used dictionary here):
- https://www.duden.de/rechtschreibung/Bibliografie
- https://www.duden.de/rechtschreibung/E_Mail

laurmaedje:
I think the email and bibliography changes are improvements, but I agree that "Liste" is not a correct translation of "Listing".

blinkerfluessigkeit:
Yeah, sorry for that. I got ahead of myself with the "Listing" change. Fixed now.

linux-user36:
How about "Auflistung"? It's the best result I can get with AI, as well as with multiple online dictionaries (LEO for instance). I believe it would be better than keeping the English term, which AFAIK doesn't seem to have any German-language usage.

Spammed:
What about 'Quelltext' (english 'source code') for 'raw'?
But to be honest, I feel that “Listing” is more accurate, it's just not a German word.
'Rohtext' (english 'raw text') would be factually correct, in German and (nevertheless) understandable.

eltos:
For reference, these are the terms used by babel/cleveref from #6852:

```
figure = Abbildung
table = Tabelle
equation = Gleichung
bibliography = Literaturverzeichnis
heading = Abschnitt
outline = Inhaltsverzeichnis
raw = Listing
page = Seite
footnote = Fußnote
```

eltos:
> What about 'Quelltext' (english 'source code') for 'raw'?
> But to be honest, I feel that “Listing” is more accurate, it's just not a German word.
> 'Rohtext' (english 'raw text') would be factually correct, in German and (nevertheless) understandable.

Vote for "Quelltext", as "Listing" is actually a German term [used in the context of the stock exchange (duden link)](https://www.duden.de/rechtschreibung/Listing).

laurmaedje:
I'll merge only `email` and `bibliography` for now while we keep bikeshedding the term for `raw`.

Thanks!

PR #7083, opened 2025-10-09
Streamline docs for typed HTML API

laurmaedje:
Makes the HTML typed docs more useful and less noisy by
- rendering all functions on one page
- rendering global attributes only once in a separate list

The typed HTML docs will also be excluded from the search.

### DIFF
--- a/docs/reference/groups.yml
+++ b/docs/reference/groups.yml
@@ -231,3 +231,31 @@
     For example, `#emoji.face` produces the 😀 emoji. If you frequently use
     certain emojis, you can also import them from the `emoji` module (`[#import
     emoji: face]`) to use them without the `emoji.` prefix.
+
+- name: typed
+  title: Typed HTML
+  category: html
+  path: ["html"]
+  details: |
+    A typed layer over raw HTML elements.
+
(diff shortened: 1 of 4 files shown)


MDLC01:
> The typed HTML docs will also be excluded from the search.

It's great the individual functions are excluded, but is the single page listing all functions searchable? I would like to be able to search for "html" (but not the individual function names, such as "header" or "body") to get the HTML element functions list.

docs/reference/groups.yml
```diff
@@ -231,3 +231,31 @@
...
+  title: Typed HTML
+  category: html
+  path: ["html"]
+  details: |
+    A typed layer over raw HTML elements.
+
+    The `html` module provides a typed layer over the raw [`html.elem`] function
+    that allows you to conveniently create HTML elements. HTML attributes are
+    exposed as function parameters that accept Typst types and automatically take
+    of converting those into the appropriate HTML.
```
isuffix:
```suggestion
    exposed as function parameters that accept Typst types and automatically take
    care of converting those into the appropriate HTML.
```

laurmaedje:
You can search for it, but it's also quite easy to find currently without search:

image

PR #7088, opened 2025-10-10
Update Portuguese translations

mezzogallium:
Added remaining terms for european portuguese. Was initially unsure on how to translate "listing", but settled on "Listagem", which can refer to a computer printout.

### DIFF
--- a/crates/typst-library/translations/pt-PT.txt
+++ b/crates/typst-library/translations/pt-PT.txt
@@ -1,11 +1,3 @@
-# figure =
-# table = 
-# equation = 
-# bibliography = 
 heading = Secção
 outline = Índice
-# raw = 
-# page = 
-# footnote = 
 email = Correio Eletrónico
-# telephone = 


laurmaedje:
Thanks!

laurmaedje:
Actually, I realized that most of these are already translated in `pt.txt` and `pt-PT.txt` will fall back to those. So, if applicable, the generic `pt.txt` should be updated instead and terms that are the same can be omitted from `pt-PT.txt`

mezzogallium:
You're right, the fallback behavior did not occur to me. I kept only those who are clearly different (Brasillian vs. European).

laurmaedje:
The email translation now seems to be the same for both, which is redundant. The additional changes and whitespace also let the tests fail.

mezzogallium:
E-mail is spelled with different diacritic marks between Brasilian and European Portuguese ("ô" vs "ó"). 
I added the spaces because I saw them on other translation files. Should I remove them?

laurmaedje:
oh I missed that because browser search considered them equal.

the problem is the trailing spaces. there shouldn't be any of those in the other files as otherwise the tests would fail on main, too.

laurmaedje:
Thanks for the translations! I fixed the whitespace.

PR #7091, opened 2025-10-10
Add ability to ignore opening/closing delimiter in `math.lr`

### LINKED ISSUE #6940: `lr` can wrongly affect other contained delimited math functions
### Description

When nesting the [left/right function `lr`](https://typst.app/docs/reference/math/lr#functions-lr), the outer `lr` affects the inner `lr`, giving a ridiculous result.

Minimal example:

```typst
$ lr(abs(x) 1/2 abs(y)) $
```

Image

## Real use case

Sometimes we need to [create delimited sequences with a missing right or left delimiter (forum.typst.app)](https://forum.typst.app/t/how-do-you-create-delimited-sequences-with-a-missing-right-or-left-delimiter/1529/1).
In that case, the first or last element in the body of the `lr` might be another `lr`, which will trigger this issue.

```typst
 // Bad result
$ lr(abs(x) f(x/2) |)_(x=3) = abs(3) f(3/2). $

 // Good result, though bad semantics
$ lr(zws abs(x) f(x/2) |)_(x=3) = abs(3) f(3/2). $
```

Image

## Background

_The following parts are only background. This issue is only about nesting `lr`, and does not cover API design, documentation, or any other thing._

Unicode have many zero width characters: zws, zwj, zwnj. For example, [the physica package uses `zwj` in its `evaluated` function.](https://github.com/Leedehai/typst-physics/blob/0b33e69d9caf131ffb052bad094dcad703b24320/physica.typ#L167-L169)
However, according to the [Unicode core spec](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-23/#G12985), they are designed for controlling line and word breaking, cursive connection and ligatures, combining grapheme joiner, etc.
So none of them is semantically correct here.

<details>
<summary>Quoting Unicode core spec</summary>

[23.2 Layout Controls](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-23/#G12985)

- Zero Width No-Break Space. In addition to its primary meaning of byte order mark, the code point U+FEFF possesses the semantics of ZERO WIDTH NO-BREAK SPACE, which matches that of word joiner. 
- Zero Width Space. The U+200B ZERO WIDTH SPACE indicates a word break or line break opportunity, even though there is no intrinsic width associated with this character.
- Joiner. U+200D ZERO WIDTH JOINER requests a more connected rendering of adjacent characters.
- Non-joiner. U+200C ZERO WIDTH NON-JOINER is intended to break both cursive connections and ligatures in rendering.

</details>

LaTeX equivalent: `\left. … \right|`.

[Some people](https://forum.typst.app/t/how-do-you-create-delimited-sequences-with-a-missing-right-or-left-delimiter/1529/2?u=y.d.x) suggest the following new typst API.

```typst
$ lr(x^2 - sum_(i = 0)^infinity (-1)^i x!|, left: "none")_1 $
```

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

mkorje:
Closes #6940.


Enivex:
How about an interface similar to mathtools' `\declarepaireddelimiter`? I'm thinking of something akin to

`#let rest = math.delimiters.with(l: none, r: sym.bar.v)`

(rest(riction))

It just seems more elegant and robust to me than how `lr` currently works, especially once we start needing workarounds like parameters for ignoring delimiters.

Edit: I guess this could actually be added to `lr` instead of creating a new function, with `l` and `r` defaulting to `auto`.

isuffix:
I agree with Enivex, but I would prefer to use positional parameters in all cases. I think they’re better for inline use, and still avoid potential confusion, consider: `$lr(|, a/b, |)$` vs. `$lr(l: |, a/b, r: |)$`. And I don’t think we should allow the old form at all because it’s just too implicit and hard to predict.

(I’m still fine with `mid()` as-is, but I could see it changing as well)

isuffix:
However this should probably be considered as part of the holistic math update mentioned in #7084

mkorje:
Redesigning the interface would be better than a stop gap, so I'll close this then.

PR #7092, opened 2025-10-10
 Fix PDF text attributes in logical children 

saecki:
[Related comment](https://github.com/typst/typst/pull/6619#discussion_r2394248138)

This fixes a bug with text attributes inside of logical children, and also resolves some inconsistencies with how footnote entries were previously handled in tagged PDF. Now they are in line with how realization/layouting handles them.

### DIFF
--- a/crates/typst-pdf/src/tags/groups.rs
+++ b/crates/typst-pdf/src/tags/groups.rs
@@ -184,7 +184,7 @@ impl Groups {
             GroupKind::Artifact(ty) => GroupKind::Artifact(*ty),
             GroupKind::Link(elem, _) => GroupKind::Link(elem.clone(), None),
             GroupKind::Par(_) => GroupKind::Par(None),
-            GroupKind::TextAttr(attr) => GroupKind::TextAttr(*attr),
+            GroupKind::TextAttr(attr) => GroupKind::TextAttr(attr.clone()),
             GroupKind::Standard(old, _) => {
                 let tag = self.tags.get(*old).clone();
                 let new = self.tags.push(tag);
@@ -445,7 +445,6 @@ pub enum GroupKind {
     /// contains no children. This can happen when there are overlapping tags
     /// and a pragraph is split up.
     Par(Option<Locale>),
-    // PERF: This increases the size of `GroupKind` from 40 to 56.
     TextAttr(TextAttr),
     Transparent,
     Standard(TagId, Option<Locale>),
--- a/crates/typst-pdf/src/tags/tree/build.rs
+++ b/crates/typst-pdf/src/tags/tree/build.rs
@@ -33,23 +33,21 @@ use typst_library::model::{
 };
 use typst_library::pdf::{ArtifactElem, PdfMarkerTag, PdfMarkerTagKind};
 use typst_library::text::{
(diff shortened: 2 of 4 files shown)


saecki:
I want to come up with some more descriptive names for the logical parenting behaviors.

saecki:
Well, I couldn't come up with better names :melting_face:, but I updated the documentation a bit.

saecki:
I've also refactored the text attribute computation a bit.

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -433,10 +445,27 @@ pub enum GroupKind {
...
     Par(Option<Locale>),
+    TextAttr(TextAttr),
     Transparent,
     Standard(TagId, Option<Locale>),
 }
 
+/// See [`FrameParent`].
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum LogicalChildKind {
+    Inherit,
```
laurmaedje: It seems odd to me to have two parallel enums (this and `FrameParent`). Wouldn't it make more sense to have `FrameParent` be a tuple or struct consisting of a `Location` and this enum?

crates/typst-pdf/src/tags/mod.rs
```diff
@@ -228,18 +174,18 @@ pub fn text<'a, 'b>(
     fc: &FrameContext,
     surface: &'b mut Surface<'a>,
     text: &TextItem,
-) -> TagHandle<'a, 'b> {
+) -> SourceResult<TagHandle<'a, 'b>> {
```
laurmaedje: why does this return a result? I can't find the position where it fails.

crates/typst-pdf/src/tags/groups.rs
laurmaedje: The naming being difficult perhaps also indicates that this is really only a bool: `inherit`. Or alternatively `separate` to indicate that it does _not_ inherit the styles. Not sure which is the clearer default.

tests/suite/pdftags/logical-children.typ
```diff
@@ -30,3 +30,44 @@
...
+]
+
+Some other text.
+
+--- logical-children-tags-underline-around-place render pdftags ---
+#underline[
+  Some text #place(float: true, bottom + right)[explanation].
+]
+
+Some other text.
```
laurmaedje, line 73: These examples are really helpful for understanding the distinction. A shortened version of these on the `FrameParent` docs would make understanding this _much_ easier. I had honestly never though about the fact that some of the insertions behave differently than others.

crates/typst-pdf/src/tags/mod.rs
saecki: Oh, seems like I forgot to remove this again when refactoring. I've only later added the `errors` field on `Tree`.

crates/typst-pdf/src/tags/groups.rs
saecki: I've decided to go with an `Inherit` enum that has a yes and a no variant. I wanted to go with the boolean, but that way the documenation was attached to the field in `FrameParent`, and it seems to be more relevant in the PDF export.

Issue #7095, opened 2025-10-10
Images has wrong css width percent when exported to html


AzurIce:
### Description

using 0.14.0-rc.1

For example:

```typst
#image("xxx.jpg", width: 25%)
#image("xxx.jpg", width: 2500%)
```

produces:

```html
<img src="..." style="width: 0.25%">
<img src="..." style="width: 25%">
```

### Operating system

Windows

### Typst version

- [x] I am using the latest version of Typst

AzurIce:
I found the problem is here:

https://github.com/typst/typst/blob/ddd8ae7f49d7adf35702fd9347d53fdd726ead55/crates/typst-html/src/css.rs#L41-L49

Since Ratio's get method returns value betweens 0.0 and 1.0, so the value should be timed by 100.0 here.

AzurIce:
Have created a PR that should fix this problem: #7096

PR #7096, opened 2025-10-10
Fix CSS relative value in HTML export

AzurIce:
fix: #7095 

https://github.com/typst/typst/blob/ddd8ae7f49d7adf35702fd9347d53fdd726ead55/crates/typst-html/src/css.rs#L41-L49

Since `Ratio`'s `get` method returns value betweens 0.0 and 1.0, so the value should be timed by 100.0 here.


crates/typst-html/src/css.rs
```diff
@@ -41,9 +41,9 @@ impl Properties {
 pub fn rel(rel: Rel) -> impl Display {
     typst_utils::display(move |f| match (rel.abs.is_zero(), rel.rel.is_zero()) {
         (false, false) => {
-            write!(f, "calc({}% + {})", rel.rel.get(), length(rel.abs))
+            write!(f, "calc({}% + {})", rel.rel.get() * 100.0, length(rel.abs))
```
laurmaedje: this should use the existing `percent` function instead

AzurIce:
> Would also be great if could add a test

Added a `--- image-sizing-html-css html ---` test in `visualize/image.typ` which covers the usage of `rel`

laurmaedje:
Thanks!

PR #7097, opened 2025-10-10
Documentation improvements

MDLC01:
This uses the proper Unicode character instead of ">=", which is more readable and maybe also better for screen readers (I haven't tested, though). I also added a comma to clarify.

I also fixed a typo in the `math.equation.alt` doc (fixes https://github.com/typst/typst/issues/7104).

Additionally, I improved the documentation for `math.scr` by more clearly showing the difference between calligraphic and roundhand letters in the first example (`scr(S)` was a bad choice as it looks similar to `cal(S)` with New Computer Modern Math). I also gave the second example a name.

### DIFF
--- a/crates/typst-library/src/math/style.rs
+++ b/crates/typst-library/src/math/style.rs
@@ -97,7 +97,7 @@ pub fn cal(
 /// Typst, you could then define your own version of `scr` like in the example
 /// below.
 ///
-/// ```example:"Creating a custom `scr` function using stylistic set 1"
+/// ```example:"Recreation using stylistic set 1"
 /// #let scr(it) = text(
 ///   stylistic-set: 1,
 ///   $cal(it)$,


crates/typst-library/src/math/style.rs
```diff
@@ -93,9 +94,10 @@ pub fn cal(
 /// The other way is using [font features]($text.features). For example, the
 /// roundhand style might be available in a font through the
 /// _[stylistic set]($text.stylistic-set) 1_ (`ss01`) feature. To use it in
-/// Typst, you could then define your own version of `scr` like this:
+/// Typst, you could then define your own version of `scr` like in the example
+/// below.
 ///
-/// ```example
+/// ```example:"Creating a custom `scr` function using stylistic set 1"
```
laurmaedje: The example title does not support Markdown styling

laurmaedje:
Thanks!

PR #7098, opened 2025-10-10
Add French (Canada) translation

Krasjet:
As mentioned in https://github.com/typst/typst/pull/7010#issuecomment-3388355734, in Québec we use "courriel" instead of "e-mail".


laurmaedje:
Terms that are the same in `fr.txt` and `fr-CA.txt` can be omitted from `fr-CA.txt`.

Krasjet:
OK, fixed

laurmaedje:
Thanks!

PR #7101, opened 2025-10-10
Add example to `frac` for changing inline vs block style

jbirnick:
A common desire of users is to have "horizontal" fractions in inline math and "vertical" fractions in block math. This PR adds an example which shows how to achieve that. (Meant to be included in 0.14)


laurmaedje:
Thanks!

PR #7105, opened 2025-10-10
Fix error in math argument parsing when dots aren't followed by anything

isuffix:
RE: https://github.com/typst/typst/pull/7036#discussion_r2411185246

This is a bug fix for an error in the math parser when using two dots not followed by anything, e.g. `$func(..)$`.

I chose to avoid refactoring the entire `math_arg` function to make the change easier to review and hopefully get in for 0.14, but I left a TODO comment for myself for later.


laurmaedje:
Does this solely change error cases into non-error cases or can it also affect behavior in cases that previously compiled?

isuffix:
This should only effect error cases. I can't think of any way this would impact already compiling code.

laurmaedje:
Thanks, seems sensible!

PR #7109, opened 2025-10-11
Document nested raw blocks syntax

Andrew15-5:
I wanted to quote the docs for someone about 4 backticks, but it actually says nothing about why would you even need more than 3 of them.

Since https://github.com/typst/typst/pull/7070 is not done (and no new features can be added now), the mention of nested code blocks is rather useless, as everything is still colored with one single color. However, that's only for unsupported/no languages. For Typst, it can use different colors/styles:

`````typ
````typ
= Code block
```rs
println!("Hey there")
```
````
`````

image

And with #7070 it would be even more colorful and useful.

A weird thing about docs was the 2 backticks. It claims that

`````typ
````
`````

produced empty raw text? Well, not technically. Furthermore, it won't produce anything in markup mode as it will throw an error and document won't even compile. In code mode it will produce **two** "empty raw texts." In markup you must add space or text between them to remove the error. But in any case, it just two pairs of backticks squished together, two separate raw elements with no text specified. This needs to be clarified. But it's hard to do so with few words, especially since there are "two" backticks and "two" pairs of raw elements. Which two is two, and which two is another two? Very confusing.

I also wanted to do something like "for N backticks (where N > 3) you can use (N - 1) consecutive backticks inside of it." But I think someone won't like this, and CommonMark only has example that describes this: https://spec.commonmark.org/0.31.2/#example-125.

### DIFF
--- a/crates/typst-library/src/text/raw.rs
+++ b/crates/typst-library/src/text/raw.rs
@@ -67,8 +67,11 @@ use crate::visualize::Color;
 ///
 /// # Syntax
 /// This function also has dedicated syntax. You can enclose text in 1 or 3+
-/// backticks (`` ` ``) to make it raw. Two backticks produce empty raw text.
-/// This works both in markup and code.
+/// backticks (`` ` ``) to make it raw. Two backticks is equivalent to 2
+/// separate empty raw elements (putting them together in markup mode will
+/// produce an error). Using 4+ backticks makes it possible to write 3+
+/// backticks inside of it (i.e., nested code blocks). This works both in markup
+/// and code.
 ///
 /// When you use three or more backticks, you can additionally specify a
 /// language tag for syntax highlighting directly after the opening backticks.

### CODE REVIEW

File: crates/typst-library/src/text/raw.rs [hunk 1]
```diff
@@ -67,8 +67,11 @@ use crate::visualize::Color;
 ///
 /// # Syntax
 /// This function also has dedicated syntax. You can enclose text in 1 or 3+
-/// backticks (`` ` ``) to make it raw. Two backticks produce empty raw text.
-/// This works both in markup and code.
+/// backticks (`` ` ``) to make it raw. Two backticks is equivalent to 2
+/// separate empty raw elements (putting them together in markup mode will
+/// produce an error). Using 4+ backticks makes it possible to write 3+
```
laurmaedje, line 72: The part with the two backticks was clearer before imo.

Andrew15-5: Please explain exactly what was clearer before and in what way. The "two backtick" (enclosure) a.k.a. 2 backticks on start and end a.k.a. 4 backticks in a row that allegedly "produce empty raw text", but in fact produce a compiler error?

laurmaedje: Two backticks _in total_ produce empty raw text. Previously, there was a distinction between two cases:
- Case 1: 1 or 3+ backticks for enclosing
- Case 2: Two backticks produces empty raw text

Andrew15-5: In that case, the wording was and still is ambiguous, because this is not how I interpreted those sentences. Since the first completely skipped the "2" backtick enclosure, IMO, it makes sense that the next sentence clarifies what will happen if you enclose in "two" backticks (a total of 4). But yeah, if **total** is 2, then it checks out.

It must use more descriptive/narrow/precise words when referring to "count of pairs of backticks to enclose text in" and "total count of backticks used in a single expression/raw element". They must be unambiguous and easy to differentiate/discriminate. Otherwise, it's a mess and makes things harder to understand after reading the docs.

laurmaedje: Then we probably need a third version because both are confusing (one to you and one to me).

Andrew15-5: Agreed. I'll look into this later.

PR #7115, opened 2025-10-12
Fix wrong PR numbers in changelog

Marcono1234:
These were just the ones I noticed, hopefully the new PR numbers are correct.

There might be more incorrect ones but checking all PR number again might not be worth it? It is also in some cases a bit difficult to tell if a PR number is correct because it refers to a large PR (e.g. related to HTML support) which might have included that change as side-effect.


laurmaedje:
Thanks!

PR #7118, opened 2025-10-12
Document math font features introduced in v0.14

### LINKED ISSUE #6737: Problems on setting CJK/Han font in math equations
Previous title: _Some math features collapse when `show math.equation: set text(font: (NCM, NCM Math))`_

### Description

Some math features are not behave normally when setting math font to `("New Computer Modern", "New Computer Modern Math")`.
Specifically, in the following example, the spacing between the two sides of `:=` is lost, and `{` does not automatically stretch its height either.

```typst
#show math.equation: set text(font: (
  "New Computer Modern",
  "New Computer Modern Math",
))
$ f := cases(x, y) $
```

Image

This issue occurs only if the first font is NCM (_New Computer Modern_):

- `show math.equation: set text(font: (any font for regular text other than NCM, any math fonts))` ⇒ ✅
- `show math.equation: set text(font: (NCM, any math fonts))` ⇒ 😓

### Futher examples

Image
Image

```typst
#set page(height: auto, width: auto)

#let example = $
  f := cases(
    1 "if" (x dot y)/2 <= 0,
    2 "if" x "is even",
    3 "if" x in NN,
    4 "else",
  )
$

#let good = emoji.checkmark.box
#let bad = emoji.face.sweat

= Default #good
#example

= NCM Math #good
(NCM = New Computer Modern)
#show math.equation: set text(font: "New Computer Modern Math")
#example

= Libertinus Serif, NCM Math #good
#show math.equation: set text(font: (
  "Libertinus Serif",
  "New Computer Modern Math",
))
#example

= NCM, NCM Math #bad
#show math.equation: set text(font: (
  "New Computer Modern",
  "New Computer Modern Math",
))
#example

#pagebreak()

= Fira Math #good
#show math.equation: set text(font: (
  "Fira Sans",
  "New Computer Modern Math",
))
#example

= Fira Sans, NCM Math #good
#show math.equation: set text(font: (
  "Fira Sans",
  "New Computer Modern Math",
))
#example

= NCM, Fira Math #bad
#show math.equation: set text(font: (
  "New Computer Modern",
  "Fira Math",
))
#example

= Fira Sans, Fira Math #good
#show math.equation: set text(font: (
  "Fira Sans",
  "New Computer Modern Math",
))
#example
```

</details> 

## Why do you want `(NCM, NCM Math)`?

What I actually need is:
- use _Noto CJK_ for Han/CJK characters and punctuations, e.g., `（所谓“定义”）`, and
- leave other characters remain as default.

This leads to `font: (NCM, Noto CJK, NCM Math)`.

Image

```typst
#set text(fallback: false) // For debugging

= Default
$ x = 1 "（所谓“定义”）" $

= NCM, Noto CJK, NCM Math
#show math.equation: set text(font: (
  (name: "New Computer Modern", covers: "latin-in-cjk"),
  "Noto Serif CJK SC",
  "New Computer Modern Math",
))
$ x = 1 "（所谓“定义”）" $
```

Ref: [如何修改公式里的中文字体？ How change Chinese font in equations? | Typst 中文社区导航 typst-doc-cn guide](https://typst-doc-cn.github.io/guide/FAQ/equation-chinese-font.html) 

### Reproduction URL

_No response_

### Operating system

Web app, Linux, Windows

### Typst version

Additionally, I've tested the basic example in `main` (bcc71ddb9).

YDX-2147483647:
Resolves #6737

Most features were implemented in #6365 by https://github.com/mkorje (Max).


docs/reference/library/math.md
```diff
@@ -86,15 +86,59 @@ $ (3x + y) / 7 &= 9 && "given" \
...
+[`font`]($text.font), the rule can also change OpenType [features]($text.features)
+including [stylistic sets].
+
+```example:"Change the overall math font"
 #show math.equation: set text(font: "Fira Math")
 $ sum_(i in NN) 1 + i $
 ```
 
+```example:"Change the font for a specific character"
+#show "{": set text(font: "STIX Two Math", fill: maroon)
```
MDLC01: It might be worth noting that this changes the font even outside of equations.

docs/reference/library/math.md
```diff
@@ -86,15 +86,59 @@ $ (3x + y) / 7 &= 9 && "given" \
...
+$ f(x, y) := cases(0 "if" x < 0, x "otherwise") $
+```
+
+```example:"Change the font for a range of characters"
+#show math.equation: set text(font: (
+  (name: "Noto Sans Math", covers: regex("[𝐴-𝑍]")),
```
MDLC01: This is not an issue for this specific range, but using similar ranges might not work in the general case because some mathematical letters are outside of the [Mathematical Alphanumeric Symbols](https://unicode.org/charts/PDF/U1D400.pdf) block. For example, the range `[𝑎-𝑧]` misses ℎ, which is part of [Letterlike Symbols](https://unicode.org/charts/PDF/U2100.pdf).

docs/reference/library/math.md
YDX-2147483647: How about wrapping the rule in `show math.equation: it => { … ; it }`? Or is that too complicated?

MDLC01: AFAIK this is the only way to do it properly. It's maybe a bit awkward, but I feel like suggesting something else would be bad advice.

docs/reference/library/math.md
YDX-2147483647: > In the last example above, `[𝐴-𝑍]` matches capital serif italic [mathematical alphanumeric symbols](https://en.wikipedia.org/wiki/Mathematical_Alphanumeric_Symbols) definied in the Unicode standard. They are not the regular `A` and `Z` in ASCII. Also, some characters do not belong to this Unicode block for historic reasons, making it harder to match other ranges. For example, `[𝑎-𝑧ℎ𝚤𝚥]` matches lower letters.

How about adding this paragraph?

MDLC01: I think something like "For example, [𝑎-𝑧] does not match ℎ and the dotless 𝚤 and 𝚥." would be clearer for the last sentence. Also, you wrote "definied" instead of "defined" in the first sentence.

YDX-2147483647: I've just committed the new version. Thanks for your advices!

crates/typst-layout/src/math/mod.rs
```diff
@@ -689,8 +689,10 @@ fn warn_non_math_font(font: &Font, engine: &mut Engine, span: Span) {
     if !font.info().flags.contains(FontFlags::MATH) {
         engine.sink.warn(warning!(
             span,
-            "current font is not designed for math";
-            hint: "rendering may be poor"
+            "the current base math font is {}, but it is not designed for math",
```
laurmaedje: I'm not sure the "base" is adding anything here. It makes it less clear to me.

docs/reference/library/math.md
laurmaedje, line 1: While the information added here is valuable, I feel like it is too specific for the general math page and it may seem intimidating and of little use to someone who just wants to start writing some math. Unless we find a better place now, I think I'd prefer delaying its inclusion until the math docs are somehow expanded in a way where we can add such things into a better place. (We do plan to put a lot more focus on docs in the next release cycle, so this would not be an indefinite delay.)

crates/typst-layout/src/math/mod.rs
YDX-2147483647: Hmm, I think if only a single font is specified, then _current font_ is clear; but if multiple fonts are specified, emphasizing _base_ is necessary.

If this is the right way to go, I can extract this part into a separate PR.

**EDIT:** I've removed the word _base_ here, but I still think it's necessary to emphasize _base_ in a way.

docs/reference/library/math.md
```diff
@@ -86,15 +86,75 @@ $ (3x + y) / 7 &= 9 && "given" \
...
+Note that typesetting maths involves complex positioning and spacing, requiring
+a specially designed [OpenType math font](https://learn.microsoft.com/typography/opentype/spec/math).
+Typst will take the first font without any [`covers`]($text.font) as the base
+font, and extract typographic metrics for maths from it. If you have to use
+non-math fonts for certain glyphs, specify a coverage to tell Typst select the
+base from other fonts.
+
+```example
+#show math.equation: set text(font: (
+  (name: "New Computer Modern Math", covers: "latin-in-cjk"),
```
YDX-2147483647: `"latin-in-cjk"` has to be replaced with `regex("\p{Latin}")`, if the first font does not support math.
Otherwise, the `{` of `cases` won't be stretched.
图片
图片



Relates-to:
- [Is there still difference between `$upright(…)$` and `$"…"$` in v0.14.0? - Questions - Typst Forum](https://forum.typst.app/t/is-there-still-difference-between-upright-and-in-v0-14-0/6706)
- https://github.com/typst-doc-cn/guide/pull/95
- [How to change the Chinese font in equations? | Typst Chinese community FAQ](https://typst-doc-cn.github.io/guide/FAQ/equation-chinese-font.html) (in Chinese)
- [1. Setting Chinese font in mathematical equations, Changes in Typst v0.14.0 — Chinese Layout Gap Analysis - General - Typst Forum](https://forum.typst.app/t/changes-in-typst-v0-14-0-chinese-layout-gap-analysis/6776#p-20380-h-1-setting-chinese-font-in-mathematical-equationshttpstypst-doc-cngithubioclreqwrong-font-fallback-for-chinese-in-math-equations-1)


**Update:** It turns out that setting different fonts for Latin and math is possible, but very complicated. See [the Chinese FAQ](https://typst-doc-cn.github.io/guide/FAQ/equation-chinese-font.html#其它方法-数学、西文字体分开) for details. I think adding a explanatory sentence would suffice.

YDX-2147483647: The previous configuration affects `$...$` and `$dot.c$`. Now fixed.

laurmaedje:
The second section of my comment https://github.com/typst/typst/pull/8836#issuecomment-5718154118 is somewhat relevant for this.

Issue #6953, opened 2025-09-24
warn or error when a variable is used as reserved names.


JivenTimsina:
### Description

Typst allows a user to define a variable with the same name as a built-in function and silently overwrites it. This can lead to confusing errors when the original function is expected.

**Example:**
```
#let heading = [*Heading*]

#show heading: it => [
  #set align(center)
  #set text(13pt, weight: "regular")
  #block(smallcaps(it.body))
]
```

**Error produced:** 
```
error: expected symbol, string, label, function, regex, or selector, found conte
nt
   ┌─ 6layout.typ:47:6
   │
47 │ #show heading: it =>[
   │       ^^^^^^^
```

**Feature Request:**

- The compiler should warn or produce an error if a user-defined variable shadows a built-in function.
- The error message should indicate which built-in function was shadowed to help with debugging.

### Use Case

- Prevents confusing, hard-to-debug errors caused by accidentally overwriting built-in functions.

PgBiel:
This is a duplicate of https://github.com/typst/typst/issues/4263, but perhaps this issue can be renamed so it has a different scope.

We've added hints for shadowed function calls at least, more general hints are harder to get right. But maybe we can do something specific to selectors - set rules, show rules.

Ri-Nai:
Hello, I'm working on PR #7131 regarding "hint on shadowed built-ins in set/show rules".

I also found a pre-existing issue:

**The shadowing hint only appears when the function call causes an error.**

For example:

```typst
#let image = "image" // shadows built-in
#image("image.png") // causes error & gets hint!
```

But, if the shadowing definition is a function, no error occurs, and thus no hint is given:

```typst
#let image(it) = { it } // shadows built-in
#image("image.png")     // no error & no hint
```

PgBiel:
I mean, sounds like that's working as intended. You're supposed to be able to shadow built-ins; it'd be very annoying to error on that (bar keywords which error at the parser level).

Ri-Nai:
> I mean, sounds like that's working as intended. You're supposed to be able to shadow built-ins; it'd be very annoying to error on that (bar keywords which error at the parser level).

Thanks for the clarification! I understand that shadowing built-ins is working as intended.

I was thinking, would it be possible to show a warning and a hint in this situation? My thought is that a user might not be aware they have shadowed a built-in variable. If they then make a mistake, the lack of a hint could make it difficult to figure out what went wrong.

[...]

PR #7131, opened 2025-10-13
Hint on shadowed built-ins in set/show rules

Ri-Nai:
### Description

related to #6953

This PR extends the hint for shadowed standard library built-ins to cover identifiers used within `set` and `show` rules.

The core of this fix re-uses the shadowing logic introduced in **PR #4402** (which solved **#4263**). The function in question is:

```rust
/// Provide a hint if the callee is a shadowed standard library function.
fn hint_if_shadowed_std(
    vm: &mut Vm,
    callee: &ast::Expr,
    mut err: HintedString,
) -> HintedString {
    // ... function body ...
}
```

**Implementation Details / Location Change**

This utility function was previously located at `crates/typst-eval/src/call.rs`, where it was only called for function-related errors. To make it accessible for use in `crates/typst-eval/src/rules.rs` (which handles the style rules), I've moved it to `crates/typst-eval/src/binding.rs`.

I chose `binding.rs` as the new home because it seemed like a reasonable location for a utility function that deals with identifier resolution/binding scope checks. I'm open to suggestions if there's a better-suited place or just staying at previous location

I have also added tests for this fix in `tests/suite/styling/set.typ` and `tests/suite/styling/show.typ`.

### DIFF
--- a/crates/typst-eval/src/lib.rs
+++ b/crates/typst-eval/src/lib.rs
@@ -13,6 +13,7 @@ mod math;
 mod methods;
 mod rules;
 mod vm;
+
 pub use self::call::{CapturesVisitor, eval_closure};
 pub use self::flow::FlowEvent;
 pub use self::import::import;


crates/typst-eval/src/call.rs
```diff
@@ -432,22 +432,7 @@ fn wrap_args_in_math(
...
-        let ident = ident.get();
-        if vm.scopes.check_std_shadowed(ident) {
-            err.hint(eco_format!(
-                "use `std.{ident}` to access the shadowed standard library function",
-            ));
-        }
-    }
-    err
-}
+// ... hint_if_shadowed_std moved to `binding.rs`
```
laurmaedje: this comment can be removed

crates/typst-eval/src/binding.rs
```diff
@@ -5,6 +5,24 @@ use typst_library::foundations::{Array, Dict, Value};
 use typst_syntax::ast::{self, AstNode};
 
 use crate::{Access, Eval, Vm};
+use typst_library::diag::HintedString;
+
+/// Provide a hint if the callee is a shadowed standard library function.
+pub(crate) fn hint_if_shadowed_std(
```
laurmaedje: I think at the bottom of `vm.rs` would be a better place.

Ri-Nai:
Thanks for the review! Requested changes have been implemented.

laurmaedje:
Thanks!

PR #7137, opened 2025-10-13
Clean up current math parser

isuffix:
This is a reworking of the initial commits from #7072 and also includes the promised refactor for #7105.

This PR should not change any behavior of the math parser.


crates/typst-syntax/src/parser.rs
```diff
@@ -583,21 +575,21 @@ fn math_arg<'s>(p: &mut Parser<'s>, seen: &mut FxHashSet<&'s str>) -> bool {
...
-    if count != 1 {
+        // Wrap to create an empty Math node.
+        p.wrap(arg, SyntaxKind::Math);
+    } else if count > 1 {
+        // Multiple adjacent expressions need to wrap as a Math element.
         p.wrap(arg, SyntaxKind::Math);
     }
+    // We don't wrap when `count == 1`, since wrapping would change the type of
+    // the expression from potentially non-content to content.
+    // Ex: `$ func(#12pt) $` would change the type from size to content.
```
laurmaedje: the newlines are a bit funky here. looks a bit like what happens with heavy rebase / jj usage? :p

crates/typst-syntax/src/parser.rs
```diff
@@ -526,52 +526,44 @@ fn math_arg<'s>(p: &mut Parser<'s>, seen: &mut FxHashSet<&'s str>) -> bool {
...
-        if !positional {
+
+    if count == 0 && arg_kind == Some(SyntaxKind::Spread) {
+        // A spread arg with no following expression converts back to two dots.
+        arg_kind = None;
+        let dots = vec![
+            SyntaxNode::leaf(SyntaxKind::MathText, "."),
+            SyntaxNode::leaf(SyntaxKind::MathText, "."),
+        ];
+        p[m] = SyntaxNode::inner(SyntaxKind::Math, dots);
```
laurmaedje: maybe not worth it, but just a thought: if we just push the text nodes flatly, we could maybe unify the `p.wrap(arg, SyntaxKind::Math);` for all branches?

crates/typst-syntax/src/parser.rs
```diff
@@ -253,7 +254,7 @@ fn math_expr_prec(p: &mut Parser, min_prec: usize, stop: SyntaxKind) {
             continuable = true;
             p.eat();
             // Parse a function call for an identifier or field access.
-            if min_prec < 3 && p.directly_at(SyntaxKind::LeftParen) {
+            if min_prec <= MATH_FUNC_PREC && p.directly_at(SyntaxKind::LeftParen) {
```
laurmaedje: I never realized that all of these `min_prec < 3` are dead code with the current precedences.

Regardless, if this is changed to `<=`, shouldn't the numbers in the constants be adjusted to `2`?

crates/typst-syntax/src/parser.rs
```diff
@@ -301,10 +291,10 @@ fn math_expr_prec(p: &mut Parser, min_prec: usize, stop: SyntaxKind) {
         }
 
         SyntaxKind::Root => {
-            if min_prec < 3 {
+            if min_prec <= MATH_ROOT_PREC {
                 p.eat();
                 let m2 = p.marker();
-                math_expr_prec(p, 2, stop);
+                math_expr_prec(p, 2, syntax_set!());
```
laurmaedje: This `2` should probably also be extracted into a constant. In fact, if my [other comment](https://github.com/typst/typst/pull/7137#discussion_r2472243314) makes sense, this perhaps would just be `MATH_ROOT_PREC`?

laurmaedje (commented):
The changes look good! But there still seems to be a slight change in behavior in the last commit. Specifically, the output of the test case `$ f_1_2^3 $` which I mentioned [in my other review comment on the other PR](https://github.com/typst/typst/pull/7072#issuecomment-3385800159) still differs before & after the last commit.

I noticed a small change in how you handle the stop set compared to before, but unless I'm missing something, this didn't fix it.

crates/typst-syntax/src/parser.rs
isuffix: This prompted me to think about it more and I realized we can fully handle the 0 expression case in `Lexer::maybe_math_spread_arg`! We just need to include checks for comma, semicolon, right paren, and dollar so we know we're not at the end of an argument (and we already check that there isn't trivia).

I've updated the first commit and also added an assert  in the `count == 0` statement.

One edge case here is if a spread is followed directly by a binary operator that isn't in the `MATH_EXPR` syntax set (`$func(../.)$`). In that case, `math_exprs` would take the "unexpected" branch and would return a count of 0, so I changed `math_exprs` to also count on unexpected tokens and added a test for those cases.

crates/typst-syntax/src/parser.rs
isuffix: This was meant to comment the end of the if-else chain, but since it didn't end in an `else` for `count == 1`, it looks odd.

Fixed by the other changes to spread args in the next comment.

crates/typst-syntax/src/parser.rs
isuffix: Actually, because roots are a prefix operator, they shouldn't check against the caller's `min_prec` in the first place. Not sure why I didn't realize that before.

And it should use `MATH_ROOT_PREC`, good catch.

crates/typst-syntax/src/parser.rs
isuffix: Yep. Setting to 2 :+1: 

Also, I guess you deleted your "the previous, flipped condition `prec >= min_prec` seemed clearer to me." comment? I still changed to using `prec >= min_prec` everywhere anyway because I think I agree that it's clearer. It also better matches the function comment: "Parses a math expression with at least the given precedence."

isuffix:
Rebased off main with the changes in the review comments. I also reordered the commits so that "Add new tests" comes first.

I definitely did not remember your comment on `$f_1_2^3$` from before :smiling_face_with_tear: . I've now added more variants of that case as the `math-attach-double-chain` test. This was tricky to get right because a naive solution would be to remove the `stop_set` check from the while loop condition (`while p.at_set(chain_set) && !p.at_set(stop_set)`), but removing that check breaks cases like `a^b'_c` in `math-primes-complex`. The actual fix is to only check the `stop_set` when _starting_ a chain with a prime, so that primes only start a chain when they don't _interrupt_ another chain.

The chaining loop now looks like this:
```rust
// Avoid interrupting a chain when initially parsing a prime.
// For `a^b'_c^d` the grouping is `(a^(b')_c)^d` and not `a^(b'_c^d)`.
if !(op_kind == SyntaxKind::MathPrimes && p.at_set(stop_set)) {
    // Parse chained attachment operators as a single attachment.
    while p.at_set(chain_set) {
        ...
    }
}
```

crates/typst-syntax/src/parser.rs
```diff
@@ -312,104 +302,112 @@ fn math_expr_prec(p: &mut Parser, min_prec: usize, stop: SyntaxKind) {
...
-            // No attachments, so we need to wrap primes as attachment.
-            if primed {
-                p.wrap(m, SyntaxKind::MathAttach);
-            }
-
-            break;
+    // Parse infix and postfix operators. The general form of a parsed op looks
+    // like: `MathAttach[ MathText("x"), Hat("^"), MathText("2") ]`.
+    while !p.at_set(stop_set)
+        && let op_kind = p.current()
```
laurmaedje, line 330: Why is this let in the condition? Couldn't it be just in the body of the loop?

laurmaedje:
My first impression was that the change from 

https://github.com/typst/typst/blob/c85a6cdc3b3525d10d922853c202d942115a44f2/crates/typst-syntax/src/parser.rs#L357

to 

https://github.com/typst/typst/blob/593353e2666733bcb4c45a9986e8f7c5935d5da8/crates/typst-syntax/src/parser.rs#L363

(and same a bit below) was supposed to fix this issue. If it wasn't, what was that change about?

laurmaedje:
Thank you!

Issue #7000, opened 2025-10-01
Not displayed footnotes still cause an error during HTML export with custom `<html>` or `<body>`


MDLC01:
### Description

With https://github.com/typst/typst/pull/6992, it is no longer possible to compile to HTML a document that contains a footnote, even with a `show footnote: none` rule.

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
_In principle,_ this seems consistent since `show footnote: none` does not hide footnote entries in paged export either. (With the exception of when the footnote is at the end of the page run which is an inconsistency in the tag handling.) This is similar to how it works for citations and other semantic elements. It only styles the footnote reference. 

Fully removing the footnotes would be a use case for a replace selector that fully replaces an element and removes the original from the document structure. This has come up before in https://github.com/typst/typst/issues/6548#issuecomment-3266662892.

MDLC01:
Then, maybe we should make `show footnote.entry: none` prevent the error? Right now, it doesn't work either.

pauladam94:
How to export a document to HTML and PDF that has some footnote inside of it ? Have we to rewrite the footnote function with different behavior if it is paged or not ? Is it not the idea behind `show` ?

laurmaedje:
It's just not supported with custom head and body right now, as HTML export is experimental. Making `show footnote.entry: none` do this would be complicated. Perhaps we can find some way or just revert #6992 before the full release.

pauladam94:
#6992 might juste become a warning without being an error for now since they are no simple way to fix such document.

MDLC01:
The problem is that the error "hides" a compiler panic that would happen without the error.

laurmaedje:
It would be easy to have it produce no error at all --- just no footnotes. It would also be easy to have a warning. Just having a warning or error that is somehow silencible when you are processing the footnotes differently is tricky.

MDLC01:
Then I think a warning is okay. It's of course a bit annoying if you handle footnotes yourself to still have the warning, but at least it lets you do it.

laurmaedje:
I didn't really want to build a workaround, but in the end I opted to do it anyway: https://github.com/typst/typst/pull/7145

It now _errors_ by default, but does nothing with a custom `show footnote` rule, as originally requested in this issue.

Warning by default did not work because the compiler error with "failed to determine link anchor" shadowed the warning.

PR #7145, opened 2025-10-15
Do not error for `show footnote: none` with custom head/body

laurmaedje:
Fixes https://github.com/typst/typst/issues/7000

### DIFF
--- a/crates/typst-html/src/rules.rs
+++ b/crates/typst-html/src/rules.rs
@@ -309,7 +309,7 @@ const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
         .pack()
         .styled(HtmlElem::role.set(Some("doc-noteref".into())));
 
-    // Indicates the presence of a default footnote rule to emit a warning when
+    // Indicates the presence of a default footnote rule to emit an error when
     // no footnote container is available.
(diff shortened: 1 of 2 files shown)


crates/typst-html/src/rules.rs
```diff
@@ -308,7 +309,11 @@ const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
         .pack()
         .styled(HtmlElem::role.set(Some("doc-noteref".into())));
 
-    Ok(HElem::hole().clone() + link)
+    // Indicates the presence of a default footnote rule to emit a warning when
```
MDLC01:
```suggestion
    // Indicates the presence of a default footnote rule to emit an error when
```

crates/typst-library/src/model/footnote.rs
```diff
@@ -330,3 +330,10 @@ cast! {
     FootnoteElem,
     v: Content => v.unpack::<Self>().unwrap_or_else(Self::with_content)
 }
+
+/// This is an empty element inserted by the HTML footnote rule to indicate the
+/// presence of the default footnote rule. It's only used by the warning in
```
MDLC01:
```suggestion
/// presence of the default footnote rule. It's only used by the error in
```

PR #7148, opened 2025-10-15
Use correct terminology in `assert.ne` error message

MDLC01:
The term "inequality" mostly refers to "less than" and "greater than" relations. "distinctness" is more correct and precise.

### DIFF
--- a/tests/suite/foundations/assert.typ
+++ b/tests/suite/foundations/assert.typ
@@ -25,12 +25,12 @@
 
 --- assert-ne-fail ---
 // Test failing assertions.
-// Error: 2-19 inequality assertion failed: value 11 was equal to 11
+// Error: 2-19 distinctness assertion failed: value 11 was equal to 11
 #assert.ne(11, 11)
 
 --- assert-ne-fail-message ---
 // Test failing assertions.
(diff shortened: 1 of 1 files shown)


jamesrswift:
Surely "inequality" is fine? It literally means "not equals"

laurmaedje:
I tend to agree with @jamesrswift. "Distinctness" seems odd to me.

MDLC01:
In mathematical contexts, "inequality" generally refers to something more specific than "not equal", such as "greater than" or "smaller than".[^1][^2] On the other hand, "Two objects that are not equal are said to be **distinct**."[^3]

Obviously "inequality" isn't terribly bad as one can infer the exact meaning based on the context, but "distinctness" is more precise.

[^1]: https://en.wikipedia.org/wiki/Inequality_(mathematics)
[^2]: https://en.wiktionary.org/wiki/inequality
[^3]: https://en.wikipedia.org/wiki/Equality_(mathematics)

laurmaedje:
I would argue that this isn't really a mathematical context though.

MDLC01:
Even though I prefer "distinctness",  don't have a strong opinion so feel free to close this.

laurmaedje:
I'll close this then.

PR #7152, opened 2025-10-15
Document version type comparison

Andrew15-5:
Mentioned in https://forum.typst.app/t/typst-0-14-just-entered-its-final-testing-period-your-help-is-wanted/6363/6?u=andrew.

The comment is not supposed to be rendered. Not sure if examples should be named. Will they be open by default? Do you toggle it with extra parameter?

There is also https://typst.app/universe/package/backtrack, which I feel like from a practical standpoint would be nice to mention. But since `version` type was added a long time ago, maybe it doesn't really make sense for someone to create a new package, that will support super old Typst versions.

### DIFF
--- a/crates/typst-library/src/foundations/version.rs
+++ b/crates/typst-library/src/foundations/version.rs
@@ -65,7 +65,7 @@ impl Version {
     ///
     /// It can have any number of components (even zero).
     ///
-    /// ```example
+    /// ```example:"Constructing versions"
     /// #version() \
     /// #version(1) \
     /// #version(1, 2, 3, 4) \
@@ -76,7 +76,7 @@ impl Version {
     /// As a practical use case, this allows comparing the current version
     /// ([`{sys.version}`]($version)) to a specific one.
     ///
-    /// ```example
+    /// ```example:"Comparing with the current version"
     /// Current version: #sys.version \
     /// #(sys.version >= version(0, 14, 0)) \
     /// #(version(3, 2, 0) > version(4, 1, 0))


crates/typst-library/src/foundations/version.rs
```diff
@@ -72,6 +72,15 @@ impl Version {
...
     /// #version((1, 2), 3)
     /// ```
+    ///
+    /// As a practical use case, it allows comparing current version
+    /// ([`{sys.version}`]($version)) to a specific one.
+    ///
+    /// // This will have to be updated in the future to keep true and false.
+    /// ```example
+    /// #sys.version > version(0, 13, 0)
+    /// #sys.version > version(1, 13, 0)
```
MDLC01:
```suggestion
    /// #(sys.version > version(0, 13, 0))
    /// #(sys.version > version(1, 13, 0))
```

crates/typst-library/src/foundations/version.rs
```diff
@@ -72,6 +72,15 @@ impl Version {
     /// #version((1, 2, 3, 4)) \
     /// #version((1, 2), 3)
     /// ```
+    ///
+    /// As a practical use case, it allows comparing current version
+    /// ([`{sys.version}`]($version)) to a specific one.
+    ///
+    /// // This will have to be updated in the future to keep true and false.
```
Marcono1234: Will this `/// // ...` here actually prevent this comment from appearing in the documentation? (if that is the intention)

laurmaedje: No, this would have to be an HTML comment `<!-- -->`

Andrew15-5: Thought so. I wasn't sure, since the parser handles special syntax, whether the comments should be in Typst or Markdown.

crates/typst-library/src/foundations/version.rs
```diff
@@ -72,6 +72,15 @@ impl Version {
     /// #version((1, 2, 3, 4)) \
     /// #version((1, 2), 3)
     /// ```
+    ///
+    /// As a practical use case, it allows comparing current version
+    /// ([`{sys.version}`]($version)) to a specific one.
+    ///
+    /// <!-- This will have to be updated in the future to keep true and false. -->
```
laurmaedje: I don't want to have a piece of docs that I have to update with every release.

Andrew15-5: Not every release, but only before releasing v1.13.0/v2.0.0.

crates/typst-library/src/foundations/version.rs
```diff
@@ -72,6 +72,14 @@ impl Version {
     /// #version((1, 2, 3, 4)) \
     /// #version((1, 2), 3)
     /// ```
+    ///
+    /// As a practical use case, it allows comparing current version
+    /// ([`{sys.version}`]($version)) to a specific one.
+    ///
+    /// ```example
+    /// #(sys.version > version(0, 14, 0))
+    /// #(version(0, 15, 0) > version(1, 14, 0))
```
Marcono1234: Is this really correct? Shouldn't it be this:

```suggestion
    /// #(version(0, 15, 0) > version(0, 14, 0))
```

(or give some indication that the expected comparison result here is `false`?)

Andrew15-5: Since Laurenz doesn't want an example that needs to be updated, 0.15.0 simulates the current version. Even though 15 is bigger than 14, major version is compared first. Having different outputs is more interesting, IMO.

laurmaedje: I think `sys.version > version(0, 14, 0)` is a somewhat unlikely check because it would be true for 0.14.1, but false for 0.14.0. A comparison with `>=` would be more typical in actual use. That would also make the two actually different right now. (Because currently both are false.)

laurmaedje: The `// true` and `// false` comments imo don't add anything (and will show up in the docs).

For the second comparison, I would perhaps use versions that are entirely unrelated to the current ones to remove any association. E.g. `2.5.0 and 3.2.0`.

Andrew15-5: > but false for 0.14.0

Docs are updated with a new version. 0.14.1 and 0.15.0 will make it `true`.

> A comparison with `>=` would be more typical in actual use. 

No, that depends on how you prefer to compare and a concrete use case. Since you know all the old versions, you can do, e.g., `> old` instead of `>= new`.

> The `// true` and `// false` comments imo don't add anything

It's the indication of the expected result, since `assert` is not used.

> For the second comparison, I would perhaps use versions that are entirely unrelated to the current ones to remove any association.

I guess.

laurmaedje: Once more, we disagree about the ideal outcome then. That's just the way it is. However, for this to be merged, I would like to see the changes I mentioned.

Andrew15-5: Done and done.

Andrew15-5:
Should I add to https://forum.typst.app/t/memo-editing-typst-app-docs/5705 that examples must have titles, or is it conditional?

laurmaedje:
Examples should only have titles if there are > 1 or if extra clarification is otherwise helpful

PR #7155, opened 2025-10-16
Fuzz typst-pdf, typst-svg, and typst-html

saecki:


laurmaedje:
I wonder whether it is more effective to run all exports in the same fuzz target or to have separate targets.

saecki:
Good point, so for the paged target also `typst-svg`, and then another target for `typst-html`?

saecki:
If this is merged, this code needs to be updated:
https://github.com/google/oss-fuzz/blob/eb47f56bd15626ab4ae8727bc435148dd2f9857c/projects/typst/build.sh#L23

.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         run: |
           sudo dpkg --add-architecture i386
           sudo apt update
-          sudo apt install -y gcc-multilib libssl-dev:i386 pkg-config:i386
+          sudo apt install -y gcc-multilib g++-multilib libssl-dev:i386 pkg-config:i386
```
laurmaedje, line 42: what's this for?

saecki: The tests for the ubuntu 32-bit targets failed to compile because `libfuzzer` was missing some c++ dependencies. I think this is because I've added `lib.rs` file to the `fuzz` test suite:

https://github.com/typst/typst/actions/runs/18529922065/job/52817573555

Issue #7113, opened 2025-10-12
CJK-Latin spaces should not be added to footnote numbers and superscripts (v0.14.0-rc.1 regression)


YDX-2147483647:
### Description

In [CSL 1.0.2 spec](https://docs.citationstyles.org/en/stable/specification.html#the-root-element-cs-style), there're two classes of style: `in-text` and `note`. Most styles belong to `in-text`, but this issue occurs only for `note`.

For `note` styles, https://github.com/typst/typst/releases/tag/v0.14.0-rc.1 introduces additional spaces around foonote markers (the numbers).

v0.13.1 (Expected):
<img width="280" alt="Image" src="https://github.com/user-attachments/assets/450a75f4-88ea-45cd-9f4c-654e2ac48353" />

v0.14.0-rc.1:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/e9b2995e-da82-4a5a-9888-63bf8839e9c2" />

<details><summary>Full code</summary>

````typst
#set page(height: auto, width: 240pt, margin: 15pt)

孔乙己@key\与上大人@key。

孔乙己@key \
上大人@key

#let bib = ```bib
@misc{key,
  author = {Author},
  title = {Title},
  date = {2025-06-17},
}
```.text

#let csl = ```csl
<?xml version='1.0' encoding='utf-8'?>
<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0">
  <info>
    <title>Minimal CSL</title>
    <id>minimal-csl</id>
  </info>
  <citation et-al-min="4" et-al-use-first="3">
    <layout delimiter="; " suffix=".">
      <text value="This doesn’t matter."/>
    </layout>
  </citation>
  <bibliography>
    <layout>
      <text value="This doesn’t matter either."/>
    </layout>
  </bibliography>
</style>
```.text

#bibliography(bytes(bib), style: bytes(csl))
````

</details> 

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

**Comment by YDX-2147483647** [comment]
After investigations, I realize that this issue is about superscripts, not citations.

## New minimal example

```typst
#set text(fallback: false)

孔乙己#super[1]与上大人#super[2]。

孔乙己#super[A]与上大人#super[B]。
```

v0.13.1:

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/83b232eb-3707-48ac-879c-0caaa07b12a7" />

v0.14.0-rc.1:

[...]

PR #7157, opened 2025-10-16
Fix CJK-Latin spacing incorrectly added around super/subscripts

hongjr03:
This pr fixes https://github.com/typst/typst/issues/7113.

This solution checks whether text has `shift_settings` (which is only set by `#super[]`, `#sub[]`, and similar elements that render as actual superscripts/subscripts). This fix
tracks super/subscript boundaries to only prevent spacing at the edges while
maintaining proper spacing for mixed CJK-Latin text inside.

### DIFF
--- a/crates/typst-layout/src/inline/prepare.rs
+++ b/crates/typst-layout/src/inline/prepare.rs
@@ -189,7 +189,7 @@ fn add_cjk_latin_spacing(items: &mut [(Range, Item)]) {
         // between the super/subscript and the following text.
         // But don't reset if we're staying within a super/subscript.
         if is_sup_or_sub_script
-            && items.peek().map_or(true, |(_, next_item)| {
+            && items.peek().is_none_or(|(_, next_item)| {
                 next_item
                     .text()
                     .and_then(|t| t.styles.get(TextElem::shift_settings))


laurmaedje:
What if there are both CJK and Latin characters in the superscript?

hongjr03:
> What if there are both CJK and Latin characters in the superscript?

The inner spacing should be preserved. I’ll fix it right away.

laurmaedje:
I had the feeling that this could be simplified, so I gave it a shot. This required refactoring the original implementation a bit first, which I did in https://github.com/typst/typst/pull/7175/commits/4c0499f6c6c2a0cd8fac30367abef663b657ced5, to then add the shift handling in https://github.com/typst/typst/pull/7175/commits/0d63032a2db30472daeb88cae2ed6c2000074bcf. I've opened https://github.com/typst/typst/pull/7175 with this alternative approach. I kept your test and added you as a co-author.

hongjr03:
> I had the feeling that this could be simplified, so I gave it a shot. This required refactoring the original implementation a bit first, which I did in [4c0499f](https://github.com/typst/typst/commit/4c0499f6c6c2a0cd8fac30367abef663b657ced5), to then add the shift handling in [0d63032](https://github.com/typst/typst/commit/0d63032a2db30472daeb88cae2ed6c2000074bcf). I've opened #7175 with this alternative approach. I kept your test and added you as a co-author.

Alright! Should I close this pr or not?

laurmaedje:
I'll just go ahead and close it now that I've merged the other one. Thanks for the quick fix that I could base my patch on!

PR #7164, opened 2025-10-16
Increase max width of example image in accessibility guide

EpicEricEE:
This image was scaled down too much when viewed on mobile (see https://staging.typst.app/docs/guides/accessibility/#color-use-and-contrast). Now it matches the other images on that page.

| Before | After |
| --- | --- |
|  |  |


laurmaedje:
Thanks!

PR #7169, opened 2025-10-17
Document the OpenType contextual alternates (`calt`) feature

### LINKED ISSUE #6963: Add `text.contextual-alternates` for the OpenType feature `calt`
### Description

I suggest adding a `contextual-alternates` parameter to `text` for the [OpenType feature `calt`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-calt).

```typst
#set text(contextual-alternates: false)
// Default: true, as suggested by the OpenType spec.
```

> **Tag: `calt`**
>
> **Friendly name:** Contextual Alternates
>
> Registered by: Adobe
>
> **Function:** In specified situations, replaces default glyphs with alternate forms which provide better joining behavior. Used in script typefaces which are designed to have some or all of their glyphs join.
>
> Example: In Caflisch Script, o is replaced by o.alt2 when followed by an ascending letterform.
Recommended implementation: A lookup table for the 'calt' feature specifies the context in which each substitution occurs, and maps one or more default glyphs to replacement glyphs (GSUB lookup type 6).
>
> Application interface: Discretionary feature: can be applied to glyph runs based on document markup, user control or other application criteria.
>
> **UI suggestion:** This feature should be active by default.
>
> Script/language sensitivity: Can be used with scripts that have optional, cursive styles.
>
> Feature interaction: This feature may be used in combination with other substitution (GSUB) features, whose results it may override.

### Use Case

[Fira Code](https://github.com/tonsky/FiraCode) and many other nerd fonts use `calt` to implement ligatures. It's surprises some people that `set text(ligatures: false)` does not work, because it only turns off `liga` and `clig`.

```typst
#show raw: set text(font: "Fira Code")
`<=`
#show raw: set text(ligatures: false)
`<=`
#show raw: set text(features: (calt: 0))
`<=`
```

Image

### Reference implementation

[VS Code config `editor.fontLigatures` controls `liga` and `calt`.](https://code.visualstudio.com/updates/v1_40#_font-feature-settings)
I don't think VS Code has a better design or Fira Code implemented it right.
However the truth is, it makes some people expect `set text(ligatures: false)` will disable all kinds of ligatures.
We don't need to follow VS Code, but we can provide a friendly interface for these people.

---

This was discussed before in #1783, #2871, and [How do I disable change of `!=` to `≠` in raw text / code? - Questions - Typst Forum](https://forum.typst.app/t/how-do-i-disable-change-of-to-in-raw-text-code/3684), but no actions were taken.
If this is the right way to go, I can submit a PR this time.

YDX-2147483647:
Also document how to disable features that are enabled by default.

Resolves #6963

(The font _Cascadia Code_ was just added in https://github.com/typst/typst-dev-assets/pull/13.)

### DIFF
--- a/crates/typst-library/src/text/mod.rs
+++ b/crates/typst-library/src/text/mod.rs
@@ -644,8 +644,8 @@ pub struct TextElem {
     /// A fine ligature.
     /// ```
     ///
-    /// Note that some programming fonts exploit other OpenType font features
-    /// to implement “ligatures,” including the contextual alternates (`calt`)
+    /// Note that some programming fonts use other OpenType font features to
+    /// implement "ligatures," including the contextual alternates (`calt`)
     /// feature, which is also enabled by default. Use the general
     /// [`features`]($text.features) parameter to control such features.
     #[default(true)]


crates/typst-library/src/text/mod.rs
```diff
@@ -643,6 +643,11 @@ pub struct TextElem {
     /// #set text(ligatures: false)
     /// A fine ligature.
     /// ```
+    ///
+    /// Note that some programming fonts exploit other OpenType font features
+    /// to implement “ligatures,” including the contextual alternates (`calt`)
```
laurmaedje:
```suggestion
    /// to implement "ligatures," including the contextual alternates (`calt`)
```

crates/typst-library/src/text/mod.rs
```diff
@@ -643,6 +643,11 @@ pub struct TextElem {
     /// #set text(ligatures: false)
     /// A fine ligature.
     /// ```
+    ///
+    /// Note that some programming fonts exploit other OpenType font features
```
laurmaedje: do we want to have the strong wording "exploit" or would a softer wording like "use" suffice?

YDX-2147483647: Well, I can't remember why I used _exploit_ at that time…
Changed.

laurmaedje:
Thanks!

Issue #7185, opened 2025-10-19
Ignored element in `measure` when targeting HTML


MDLC01:
### Description

Consider this:

```typ
#context measure(image("example.png"))
```

This produces the following warning with 0.14.0-rc.2 when compiling to HTML:

```
warning: elem was ignored during paged export
  ┌─ test.typ:1:17
  │
1 │ #context measure(image("example.png"))
  │                  ^^^^^^^^^^^^^^^^^^^^
```

The result of the call to `measure` is `(width: 0pt, height: 0pt)`.

Compiling to PDF instead of HTML (even with the `html` feature enabled) works as intended.

This is a regression from Typst 0.13.1.

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
Measure is just completely broken in HTML right now. It's not really a regression w.r.t. to measure, it just happened to work because `image` did not have an HTML show rule previously. `#context measure[*strong*]`, for instance, is already broken in 0.13.1.

Probably an easy fix though. I think we just need to set the target to `Paged` here:

https://github.com/typst/typst/blob/87f7a7310f7be523b0b156392bf99da520f5cbc0/crates/typst-library/src/layout/measure.rs#L93

As is done for `html.frame`:

https://github.com/typst/typst/blob/87f7a7310f7be523b0b156392bf99da520f5cbc0/crates/typst-html/src/convert.rs#L130

PR #7186, opened 2025-10-19
Fix `measure` with HTML export

MDLC01:
Closes https://github.com/typst/typst/issues/7185.

As suggested in https://github.com/typst/typst/issues/7185#issuecomment-3419572989, this makes `measure` use paged export for its content. As measuring doesn't really make sense for HTML elements, I don't think there is a need for an additional `target` parameter.

I added a test, but I'm not sure how to make it not export an empty HTML file.

### DIFF
--- a/tests/suite/layout/measure.typ
+++ b/tests/suite/layout/measure.typ
@@ -102,7 +102,7 @@
 
 --- measure-html html ---
 #context {
-    let dims = measure(image("/assets/images/monkey.svg"))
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

---

> I added a test, but I'm not sure how to make it not export an empty HTML file.

There isn't a way at the moment, but it's not too bad. I want to refactor the test suite a bit in the future and then we can get rid of it.

PR #7192, opened 2025-10-20
Improve documentation for `state.update`

Andrew15-5:
This addresses some points from https://forum.typst.app/t/why-doesnt-state-behave-like-the-documentation-says/6499.

### DIFF
--- a/crates/typst-library/src/introspection/state.rs
+++ b/crates/typst-library/src/introspection/state.rs
@@ -353,29 +353,35 @@ impl State {
 
     /// Updates the value of the state.
     ///
-    /// Returns an _invisible_ piece of content that doesn't affect document
-    /// layout (similar to [`metadata`]). The update will be in effect at the
-    /// position where the returned content is inserted into the document. If
-    /// you don't put the output into the document, nothing will happen! This
-    /// would be the case, for example, if you write
-    /// `{let _ = state("key").update(7)}`. Updates only happen if they are
-    /// associated with a location in the document. That is, only state updates
-    /// that are _in the document_ update the said state.
+    /// Returns an invisible piece of [content] that must be inserted into the
+    /// document to take effect. This invisible content tells Typst that the
+    /// specified update should take place wherever the content is inserted into
+    /// the document.
     ///
-    /// Here is another way of looking at it. State is a part of your document,
-    /// it runs like a thread embedded in the document content. The value of a
-    /// state is the result of all state updates that happened in the document
-    /// up until that point. That’s why `state.update` returns an invisible
-    /// sliver of content that you need to return and include in the document —
-    /// a state update that is not "placed" in the document does not happen, and
-    /// "when" it happens is determined by where you place it. For example, only
-    /// figures that are placed into the document increase the value of the
-    /// figure counter. That’s also why you need context to read state, you need
-    /// to use the current document position to know where on the state’s
-    /// “thread” you are.
+    /// State is a part of your document and runs like a thread embedded in the
+    /// document content. The value of a state is the result of all state
+    /// updates that happened in the document up until that point.
+    ///
+    /// That's why `state.update` returns an invisible sliver of content that
+    /// you need to return and include in the document — a state update that is
+    /// not "placed" in the document does not happen, and "when" it happens is
+    /// determined by where you place it. That's also why you need [context] to
+    /// read state: You need to use the current document position to know where
+    /// on the state's "thread" you are.
+    ///
+    /// Storing a state update in a variable (e.g.
+    /// `{let my-update = state("key").update(c => c * 2)}`) will have no effect
+    /// by itself. Only once you insert the variable `[#my-update]` somewhere
+    /// into the document content, the update will take effect — at the position
+    /// where it was inserted. You can also use `[#my-update]` multiple times at
+    /// different positions. Then, the update will take effect multiple times as
+    /// well.
     ///
     /// In contrast to [`get`]($state.get), [`at`]($state.at), and
-    /// [`final`]($state.final), this function does not require [context].
+    /// [`final`]($state.final), this function does not require [context]. This
+    /// is because, to create the state update, we do not need to know where in
+    /// the document we are. We only need this information to resolve the
+    /// state's value.
     #[func]
     pub fn update(
         self,


benderydt:
I feel that this update stills puts too much emphasis on the unusual case. I tried my hand at rewriting the documentation but I don't know how to format it. See my answer in the forum thread: https://forum.typst.app/t/why-doesnt-state-behave-like-the-documentation-says/6499/5

laurmaedje:
@bluss Perhaps it would make sense to include [your explanation from Mastodon](https://mastodon.social/@bluss@social.linux.pizza/115180635979202911) (or an adaption thereof) instead?

bluss:
Yes it would be good to include it, if I don't add it feel free to take it. The difference between that and what's here isn't all to big.

crates/typst-library/src/introspection/state.rs
```diff
@@ -353,12 +353,14 @@ impl State {
...
-    /// applied in layout order and in that case, Typst wouldn't know when to
-    /// update the state.
+    /// Returns an _invisible_ piece of content that doesn't affect document
+    /// layout (similar to [`metadata`]). The update will be in effect at the
+    /// position where the returned content is inserted into the document. If
+    /// you don't put the output into the document, nothing will happen! This
+    /// would be the case, for example, if you write
+    /// `{let _ = state("key").update(7)}`. State updates are always applied in
+    /// layout order and in that case, Typst wouldn't know when to update the
+    /// state.
```
bluss: If we would adjust this doc I think we should remove the old "Typst wouldn't know when to update the state" and replace it with an explanation that updates only happen if they are associated with a location in the document - only state updates that are in the document, update state. How perfectly tautological...

laurmaedje:
@benderydt do you think the latest version makes it clearer?

benderydt:
First of, thanks for tagging me! I'm not a frequent github user, so I missed this mention.

The short answer to the question is: No, it didn't make it that much cleaner. The long alternative explanation (76b3834) is fine, but the short reference is still baffling for a Typst beginner like me. The problem is that the explanation focuses too much on the unusualness of `state.update()` and doesn't consider the normal use case. It doesn't help that the only piece of code in the explanation is the equally baffling `{let _ = state("key").update(7)}`. I have since found the `let _`-idiom in the context of `array.pop`, and there it made sense, but why the hell would one write `let _ = state.update()`? What's the point?

I did another try below, but the best change would related to the formatting, which is done by `tidy`, I suppose? 

Right now, the format of the reference documentation is

+ general explanation of the function
+ explanation of the return type, if any
+ detailed explanation of the arguments

It would be nicer, and especially in this case, if the return could wait until all the arguments have been explained.

Thus, for `state.update`, this would be:

+ `Update`: Updates the value of the state.
    ....
+ arguments: discussion about `any` and `function`
+ only now do we discuss the peculiarities of `update`s return type.

```
    /// Updates the value of the state.
    ///
    /// Returns an _invisible_ piece of content that must be inserted into the document. 
    /// This invisible content tells Typst that at _this_ position in the document _this_ update 
    ///  must take place. 
    ///
    /// Simply writing `state("key").update(7)`in content mode will insert the invisible code.
    /// But, you can save the return of `update` for later use by binding to it, like so:
    /// `let my-update = state("key").update(c => c * 2)`
    /// This means that the update will not yet happen because it is not in the document.
    /// It's only when you include `#my-update` in the document, that the update will take place.
    /// If you include `#my-update` twice, the update will happen twice. So the state will be quadrupled.
    ///
    /// In contrast to [`get`]($state.get), [`at`]($state.at), and
    /// [`final`]($state.final), this function does not require [context].
```

laurmaedje:
Changing the general format of the documentation for a single function is not feasible. But I'm not sure the "content" return in the signature is really a big problem. I think we can fix the situation by improving the text.

I gave it another shot below, remixing previous texts. What do you all think about this edit? (in particular @benderydt)

> Updates the value of the state.
>
> Returns an invisible piece of content that must be inserted into the document to take effect. This invisible content tells Typst that the specified update should take place wherever the content is inserted into the document.
> 
> State is a part of your document and runs like a thread embedded in the document content. The value of a state is the result of all state updates that happened in the document up until that point. 
> 
> That's why `state.update` returns an invisible sliver of content that you need to return and include in the document — a state update that is not "placed" in the document does not happen, and "when" it happens is determined by where you place it. That's also why you need context to read state: You need to use the current document position to know where on the state's "thread" you are.
> 
> Storing a state update in a variable (e.g. `let my-update = state("key").update(c => c * 2)`) will have no effect by itself. Only once you insert the variable `#my-update` somewhere into the document content, the update will take effect — at the position where it was inserted. You can also use `#my-update` multiple times at different positions. Then, the update will take effect multiple times as well.
>
> In contrast to `get`, `at`, and `final`, this function does not require context. This is because, to create the state update, we do not need to know where in the document we are. We only need this information to resolve the state's value.

<details>
<summary>Code version</summary>

```
/// Updates the value of the state.
///
/// Returns an invisible piece of [content] that must be inserted into the
/// document to take effect. This invisible content tells Typst that the
/// specified update should take place wherever the content is inserted into
/// the document.
/// 
/// State is a part of your document and runs like a thread embedded in the
/// document content. The value of a state is the result of all state updates
/// that happened in the document up until that point. 
/// 
/// That's why `state.update` returns an invisible sliver of content that you
/// need to return and include in the document — a state update that is not
/// "placed" in the document does not happen, and "when" it happens is
/// determined by where you place it. That's also why you need [context] to
/// read state: You need to use the current document position to know where on
/// the state's "thread" you are.
/// 
/// Storing a state update in a variable (e.g. 
/// `{let my-update = state("key").update(c => c * 2)}`) will have no effect
/// by itself. Only once you insert the variable `[#my-update]` somewhere into
/// the document content, the update will take effect — at the position where
/// it was inserted. You can also use `[#my-update]` multiple times at different
/// positions. Then, the update will take effect multiple times as well.
///
/// In contrast to [`get`]($state.get), [`at`]($state.at), and
/// [`final`]($state.final), this function does not require [context]. This is
/// because, to create the state update, we do not need to know where in the
/// document we are. We only need this information to resolve the state's
/// value.
```
</details>

It might be good to have an example as well, but I'm not sure what and this could also be done in a separate PR.

laurmaedje:
Thanks everybody!

Andrew15-5:
I have a grammar question: why everywhere I see Typst, there is a capital letter after a colon? This is not a thing in Russian, nor British English. And even in American English it's not required when there is only one explanatory full sentence followed [^1][^2][^3]. According to Grammarly, only APA format requires capitalized word for one sentence [^3], but APA and other format guidelines are not standard English, AFAIK. So, unless Typst follows APA format in full, I don't think capitalization should take place with one explanatory sentence.

[^1]: https://www.grammarbook.com/blog/commas/capitalization-with-colons/
[^2]: https://www.grammar-monster.com/punctuation/colons_capitalization_after_a_colon.html
[^3]: https://www.grammarly.com/blog/punctuation-capitalization/capitalization-after-colons/

laurmaedje:
All the linked sources seem to say that it's fine to capitalize if a full sentence follows (even if not necessary), which mirrors my personal preference.

Andrew15-5:
I've added a section about this in https://forum.typst.app/t/typst-documentation-editing-typst-app-docs/5705#p-17155-grammar-and-style-of-writing-11.

PR #7193, opened 2025-10-21
[PoC] Exclude incompatible package versions from import autocompletion

Andrew15-5:
https://github.com/Leedehai/typst-physics/issues/69
This triggered me, and I decided to try and implemented. And...you can't. Because the current API doesn't keep the minimal required compiler version for each package version. Not sure where to put it, so I expanded the `PackageSpec`, though technically I can pass a third tuple value.

To test it out, I had to use Tinymist, which lead to:
- using https://github.com/ParaN3xus/typst's nightly branch as the base for this PR/branch
- using https://github.com/Myriad-Dreamin/tinymist's nightly as the base for actual PoC.

So I have
- https://github.com/Andrew15-5/tinymist/tree/latest-package
- https://github.com/Andrew15-5/typst/tree/latest-package

To try this out, first compile the LSP server:

```sh
git clone --branch latest-package https://github.com/Andrew15-5/tinymist
cd tinymist
cargo build --release --package tinymist
```

In Neovim, I have something like this:

```lua
local lspconfig = require "lspconfig"
lspconfig.tinymist.setup {
  cmd = { "/path/to/tinymist/target/release/tinymist", "lsp" },
}
```

Additionally, since this Tinymist is already 0.14.0, and so is Typst, we need to make a package that requires 0.15.0 (there is still no package that requires 0.14.0, btw). To do this:

```sh
cd /tmp
mkdir -p public/preview
curl https://packages.typst.org/preview/index.json > public/preview/index.json
prettier --write public/preview/index.json
```

Now run a local server:

```sh
cargo binstall -y static-web-server
```

```sh
static-web-server -p 3000
```

Test out that everything works with:

```sh
export TYPST_DEFAULT_REGISTRY=http://127.0.0.1:3000
nvim "$(date +%FT%H:%M:%SZ)".typ
```

Write `#import "@preview/phy` and get the expected latest 0.9.6 for `physica`:

image

Now open `public/preview/index.json`, copy version 0.9.6 down, and change it to 0.10.0.

Close Neovim and run `nvim "$(date +%FT%H:%M:%SZ)".typ` again, and write `#import "@preview/phy` again as well. You will get 0.10.0:

image

As expected. Now change `"compiler": "0.14.0",` (for it) to `"compiler": "0.15.0",`.

Re-run `nvim "$(date +%FT%H:%M:%SZ)".typ`, and write `#import "@preview/phy`. Tada! It shows 0.9.6:

image

Now the users won't have to:

1. import latest available version
2. get the error that the compiler is incompatible with this package
3. figure out that you should try previous version
4. find out what's the previous version
5. fixed, continue with the workflow

Instead users will:

1. import latest available version
2. it's the latest compatible, continue with the workflow

And package authors won't have to worry about users having this issue.

The only problem is that people that
- want the latest package version
- **and** are not on the latest and greatest Typst version
- **and** are not aware that latest package version uses/is specific for the latest Typst version
- **and** only rely on autocompletion instead of manually writing the version
will be confused as to why the package was updated, but they can't "see it."

So yeah, I think the impact is minimal, and the gains are much greater, i.e. net positive.

Which brings us back to the new API problem. With that decided, making sure that all IDE have this feature is easy and already implemented.

Not sure if I can make a test for it.


laurmaedje:
A `PackageSpec` is supposed to uniquely identify a package rather than being a "query" for a package. Adding a `compiler` field there is a problematic change.

Andrew15-5:
Well yeah, but this doesn't answer what are the valid alternatives to this.

laurmaedje:
Because I don't know. I didn't look at it deeply and don't really have time right now.

Andrew15-5:
Oh, in that case, I'll be passively waiting. I assumed if you leave feedback after a while (considering all other my PRs), it means you've looked at it deeply _enough_ to leave some feedback that I can work with. It'd be nice if you can hint at it (i.e., not having time for further discussion) right away.

laurmaedje:
Thing is, my time is limited and I do not consider this fix a priority. I could muster the time to skim through the description and to determine that the diff is problematic, but I think it's a net-negative for the project if I spend more time on this, given the many other priorities. If you find a good fix, that's nice, but I can't find it for you.

Andrew15-5:
> I do not consider this fix a priority.

Neither do I.

> I could muster the time to skim through the description and to determine that the diff is problematic, but I think it's a net-negative for the project if I spend more time on this, given the many other priorities.

I feel like you're implying that I'm forcing you to spend time on this, which is false. I already explained the situation above.

> If you find a good fix, that's nice, but I can't find it for you.

There is no fix to be found, the discussion is how exactly the API should change, if `PackageSpec` should be left untouched. I can try guessing for alternatives, until I hit a jackpot, but that's not exactly productive, given the many other priorities.

We are both busy, and that's normal. But being upfront about it still helps.

laurmaedje:
Given that I will not have time to discuss this in the foreseeable future, I'll close it.

PR #7198, opened 2025-10-21
Fix logical ordering of broken grid cell regions

### LINKED ISSUE #5723: Section headers in table switch order based on rendered position not code position
I have section headers in a table to display two columns (maybe not the most elegant solution, but this is beside this issue...).
When one section spans more than one page, the order between sections is changed.

I am not entirely sure if this is intended behavior or a bug. At least the result is unintuitive for me.

# MWE

The following code produces three sections and, as the first header does not span more than one page, the order is correct (`1. First Header -> 2. Second Header -> 3. Third Header`).

```typst
#set heading(numbering: "1.1.")

#table(
  columns: (auto, auto),
  [
    = First Header, should be Section 1 


    If we increase the  `#lorem(100)` to a larger number (eg 500), then the *Second Header* moves from 2 to 3 in enumeration.

    
    #lorem(100)

    = Second Header, should be Section 2
  ],
  [
    = Third Header, should be Section 3
  ],
)
```

When I change `#lorem(100)` -> `#lorem(500)`, the first section spans more than one page and the order becomes `1. First Header -> 2. Third Header -> 3. Second Header`. This is as the third header remains on the first page, whereas the second header was pushed to the second header, although it is before the third header in the code.

### LINKED ISSUE #7188: In pagebroken grid, counter updates happen on page 2 before page 1
### Description

Typst version: 0.14.0-rc.2 (Typst webapp)

```typ
// This code places a number on each word.
// Problem: In Typst 0.13.1, the counter starts on page 1 and continues to page 2.
// In typst 0.14, word number 0 and number 1 are on page 2. Then the counting continues to page 1 only after page 2..

#set page(height: 10cm)

/// "minimal reproducer" based on a custom simple line numbering function
#let wordnumbering(body, numbering: "1", pad: 0.6em, style: x => x, wordpattern: "\w+\.?") = {
  let linenumber = counter("_linenumbered")
  context {
    let wordlabel = <_word>
    show wordlabel: mt => {
      linenumber.step()
      box({
        let dx = 0pt
        let dy = 0pt
        place(end + bottom, dx: dx, dy: dy, style(std.numbering(numbering, linenumber.get().first())))
      })
    }
    
    // mark each word start so that we know when we pass a new line
    show regex(wordpattern): it => it + [#metadata(none)#wordlabel] 
    body
  }
}
#grid(
  columns: (7cm, auto),
  column-gutter: 2em,
  { 
    set text(size: 14pt, fill: blue)
    show: wordnumbering.with(style: word => text(black, strong(super(word))))
    lorem(70)
  },
)
```

The code tries to number each word (black).
Problem: The low numbers appear on page 2, then the counter continues to page 1.
Expected result is like the result in Typst 0.13.1, where the first words are on page 1, and then the count continues on page 2.

Picture below: 0.14 RC2 (Webapp)

typst rendered output

Picture below: 0.13.1 (Webapp)

typst rendered output

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

saecki:
Fixes #5723, #7188

In the introspector logical children are currently inserted directly after the start tag of the parent element.
Before #6619 broken grid cells weren't handled using the logical parenting mechanism at all, that meant the closing tag would appear somewhere on another page, which could include a lot of unrelated content and would also screw up the nesting of tags. #6619 tried to fix this by making all but the first regions of broken grid cells logical children. This however assumed they would be inserted after the contents of the first region which is directly included inside the tags of the grid cell. This explains the incorrect ordering of the counter in #7188.

There were some other issues, like #7124 where gird cells were missing from the introspector.

This PR fixes #7188 by also setting the parent of the first region to the grid cell and then prepending the start and end tag that don't span any other content.

In the future the insertions (logical children) might be placed at the end of the parent element, but that would require keeping track of the index of end tags inside the introspector which currently doesn't happen. It might however happen if something like a `within` selector is added, then this subject could be revisited.


saecki:
Fixing this issue uncovered another somewhat related bug with figure captions. The problem is that `close_group` for `GroupKind::Figure` is called before `close_group` of `GroupKind::FigureCaption`, because the figure placement uses a floating `place`. In that case the list of captions is empty when the `build_figure` is called, and when the caption is later added to that list, it is ignored.

This only showed up now, because I fixed another issue. Previously LogicalChild and LogicalParent weren't part of this list here:
https://github.com/typst/typst/blob/c2cbc2c02481df438b1d56a13fdde5d0a9328ce4/crates/typst-pdf/src/tags/tree/mod.rs#L602

That was an issue for lists though, since they internally use grid layout. It previously didn't show up, since the first region of the grid was inside the "normal" introspection tags, not inside a logical child. So the `ListItemLabel` would still find the `List` as its semantic parent, since the start tag would always be in the first region (I could however imagine that there would have been some edge cases where that wouldn't be the case). Now that a logical child is also used for the first region, this change was necessary for lists to work properly.

This also means that previously figure captions of figures which used place for the figure placement, didn't properly handle the captions, because they didn't find a Figure as their semantic parent. So they were just be added somewhere inside the figure tag.

The fix is to only call `build_figure` after the entire tree has been visited. While I couldn't think of cases where similar problems could occur for `build_grid` and `build_table`. All of these are now done right before the final tag tree is resolved.

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -489,7 +496,8 @@ impl std::fmt::Debug for GroupKind {
             Self::TermsItemLabel(..) => "TermsItemLabel",
             Self::TermsItemBody(..) => "TermsItemBody",
             Self::BibEntry(..) => "BibEntry",
-            Self::Figure(..) => "Figure",
+            Self::FigureWrapper(..) => "FigureBody",
+            Self::Figure(..) => "FigureBody",
```
laurmaedje: Why is this `FigureBody` for both?

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -620,4 +630,20 @@ impl GroupKind {
             GroupKind::Standard(_, lang) => lang,
         })
     }
+
+    /// Whether this group should be ignored when searching for a semantic
+    /// parent or child group.
+    pub fn is_non_semantic(&self) -> bool {
+        // While paragraphs, do have a semantic meaning, they are automatically
```
laurmaedje: Unnecessary comma after paragraphs

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -620,4 +630,20 @@ impl GroupKind {
             GroupKind::Standard(_, lang) => lang,
         })
     }
+
+    /// Whether this group should be ignored when searching for a semantic
+    /// parent or child group.
+    pub fn is_non_semantic(&self) -> bool {
```
laurmaedje: I think flipping the name and return value would be clearer, in particular, as all uses use it with `!`.

tests/ref/pdftags/figure-tags-inline-equation-with-caption.yml
```diff
@@ -6,25 +6,22 @@
           /Placement: Block
           /K:
             - Content: page=0 mcid=8
-    - Tag: Div
```
laurmaedje, line 9: why did these tests change? Since the relevant figures are not floating, I don't understand how this is related.

crates/typst-pdf/src/tags/groups.rs
saecki: A combination of copy paste and forgetting to update the string :facepalm:

tests/ref/pdftags/figure-tags-inline-equation-with-caption.yml
saecki: I've moved this change into it's own commit.
It's only partly related, since the figure now also uses the `is_semantic` check to look for the first semantic child. The wrapping paragraph is looked through and the formula is found as the first single semantic child. The unnecessary grouping tag can then be omitted.

PR #7199, opened 2025-10-22
Specify the signature of `layout.func` in documentation

YDX-2147483647:
This PR, though trivial, should finally complete the checklist from #408.


laurmaedje:
Thanks! It's very nice that all functions are dealt with now.That said, I'm not sure whether that covers the intent of https://github.com/typst/typst/issues/408 or whether the issue author wanted a more structured view of the function signatures. For now, I'd prefer to keep the issue open (perhaps you want to comment there), so I edited your description to remove the "close". (I couldn't find another way to unlink the issue.)

Issue #7165, opened 2025-10-16
Replacing show rules


laurmaedje:
### Description

Add a way for show rules to fully replace the element instead of just redefining its appearance. Full replacement would have several additional consequences:

- The element isn't observed through introspection anymore, e.g. in queries
- Show-set rules for the element do not apply to the nested content anymore
- The element is not considered for accessibility tagging

### Use Case

Two example use cases:
- Replacing an elements with a modified version of the same element
- Using markup without imposing its semantics and styling on the result (e.g. raw DSLs)

I'm sure there are more.

This came up before a couple times:
- https://github.com/typst/typst/issues/6548
- https://github.com/typst/typst/issues/7054
- https://github.com/typst/typst/issues/7000#issuecomment-3358087155

MDLC01:
Regarding the accessibility tagging/semantics part, the current situation already allows stripping semantics in some situations. For example, using `show emph: it => html.span(style: "font-style: italic;", it.body)` removes the "emphasization" semantics in HTML export.

laurmaedje:
Yeah, in HTML the story is a bit different because it doesn't use invisible tags for accessibility.

gabriel-araujjo:
I see a possible solution in creating a naked element, whose semantics would be to strip the parent container.

To completely remove an element, one could use something like:

```typst
#show element: naked(none)
```

laurmaedje:
@gabriel-araujjo My current thinking is something like `show selector.replacing(element): it => ..`. Having the replacing nature as part of the selector rather than in the RHS imo more clearly models that this show rule is behaving fundamentally differently. It's also clear how to implement it that way. Having it on the RHS opens a few extra question (e.g. having `show element: _ => naked(none)` might already be a bit tricky and what about `show element: naked(none) + [Hi]`?)

sijow:
Maybe it would make sense to use another name altogether for these rules, e.g.  `#replace element: ...` ? (`show` seems a bit of a misnomer when the element is truly replaced)

laurmaedje:
Perhaps, it's a _bit_ of a misnomer, but the bar for new syntax is very high and it's conceptually close enough imo. (To the point, that the difference is often not visible.) Also, `show` is already used for three different mechanisms (`show: thing`, `show thing: thing`, `show heading: set(..)`) so it doesn't feel too out of place to extend it a bit more.

MDLC01:
I am in favor of adding a new keyword here, for two main reasons:

[...]

Issue #4511, opened 2024-07-06
Add fill argument to #hide


lazyeights:
### Description

In some documents, especially legal documents, it is important to redact information with the content removed from the output pipeline so it can not be recovered. At the same time, it is important to provide a visual indication of the redacted information, and to do so in various formats. The most basic redaction is a black line that starts within a text paragraph to the end of the line. There is no indication of word length within the redacted text.

Some prior solutions tried to use the `highlight` function, but it does not hide the content (and does not work well with `hide` presently) and also does not hide the size of the redacted words, like here:

![CleanShot 2024-07-05 at 21 24 48@2x](https://github.com/typst/typst/assets/167045/34297db4-c880-475e-87a6-ed8b1070de2e)
 

### Use Case

The `hide` function seems to do exactly what is required, but its content is literally hidden and does not provide a visual indication of a deliberate redaction. It would be important for legal text if `hide` includes a `fill` argument that is customizable to provide this. `#hide(fill: black, content)` is the minimum needed, and should be easy to implement.

lazyeights:
It would probably make sense to also include the other standard highlight arguments like stroke, etc. For example, a draft document could mark proposed redactions like this:

![CleanShot 2024-07-05 at 21 49 34@2x](https://github.com/typst/typst/assets/167045/25777d5e-c089-4d33-bdda-ef169d1f845f)

ludwig-austermann:
Isn't something like this already working?
```typst
// simple redaction
#let redact(body, color: black) = box(hide(body), fill: color)
// redaction as in the issue
#let redactre(body, color: black) = {
  show regex("[^\s]"): it => redact(it)
  body
}
```

Enivex:
> Isn't something like this already doing the job?
> ```typst
> // simple redaction
> #let redact(body, color: black) = box(hide(body), fill: color)
> // redaction as in the issue
> #let redactre(body, color: black) = {
>   show regex("[^\s]"): it => redact(it, color: color)
>   body
> }
> ```

That does not hide the size of redacted words

ludwig-austermann:
Then I do not understand what the intent is. By using `"."` as the regex expression, however, everything would be marked.

lazyeights:
The regex approach still leaks information in the PDF output because the boxes contain information about the size of the glyphs or words. See https://www.wired.com/story/redact-pdf-online-privacy/

[...]

PR #7201, opened 2025-10-22
Improve documentation for hide

Andrew15-5:
Current [docs for `hide`](https://typst.app/docs/reference/layout/hide) are not clear enough, IMO. This change explicitly states that nothing will be included in the output PDF, when passed to `hide`. Nothing "invisible" etc. Same for HTML and other formats.

[Apparently](https://discord.com/channels/1054443721975922748/1399326777540608041/1430486372816523326), it might not be the case for HTML and SVG...

### DIFF
--- a/crates/typst-library/src/layout/hide.rs
+++ b/crates/typst-library/src/layout/hide.rs
@@ -4,7 +4,7 @@ use crate::introspection::Tagged;
 /// Hides content without affecting layout.
 ///
 /// The `hide` function allows you to hide content while the layout still "sees"
-/// it. This is useful to create blank space that is exactly as large as some
+/// it. This is useful for creating blank space that is exactly as large as some
 /// content.
 ///
 /// # Example
@@ -14,10 +14,10 @@ use crate::introspection::Tagged;
 /// ```
 ///
 /// # Redaction
-/// This function may also be useful to redact content as its arguments are
+/// This function may also be useful for redacting content as its arguments are
 /// neither present visually nor accessible to Assistive Technology. That said,
 /// there can be _some_ traces of the hidden content (such as a bookmarked
-/// heading in the PDF document outline).
+/// heading in the PDF's Document Outline).
 ///
 /// Note that, depending on the circumstances, it may be possible for content to
 /// be reverse engineered based on its size in the layout. We thus do not


laurmaedje:
I wonder whether the wording is a little too strong now. If, for instance, the `hide` contains a heading, there can be a trace: In the PDF outline.

@saecki Thoughts?

saecki:
It might make sense to make this more about AT, as visually there won't be anything to see by definition.
Otherwise the "leave no trace" wording seemed fine to me. The special case with bookmarked headings is somewhat unfortunate anyway.

Andrew15-5:
Any special cases can be called out explicitly, while leaving the "strong" affirmative wording, providing insight and fine/precise behavior for a well-informed decision to use `hide` for a specific use case, knowing what to expect from it.

Documentation shouldn't be vague or ambiguous, since it doesn't help anyone.

Andrew15-5:
This is probably better, but I'm still a bit confused. Like, there is https://github.com/typst/typst/issues/4511, which is literally about redacting. But I don't get if it includes removing selectable text and any other traces, or is it just for PNG output or something. I want to link to it.

Another point, is that the docs say that it can be useful for redacting content, yet it's not a perfect redaction. So this says to me a contradiction.

Lastly, https://github.com/typst/typst/issues/7165 says to me that `hide` can be just `show <thing>: none`, which will replace element with `none` and therefore will be absent in the Document Outline, tags, queries, etc. So, my point is: can't we just eliminate the remaining pieces either directly, or through https://github.com/typst/typst/issues/7165 and then confidently say that there will be no trace? Though the reverse engineering shenanigans for me sound a kind of far-fetched, I maybe see it as a potential attack in very niche cases and with low success rate. But it'll be the only one, really. So we can, at some point, just highlight it, to keep users informed and aware, but otherwise say that there are literally no traces, other than "side-effects" for reverse engineering.

laurmaedje:
The edit does seem better to me.

> So, my point is: can't we just eliminate the remaining pieces either directly, or through #7165 and then confidently say that there will be no trace?

Perhaps, but this would be out of scope for `hide` and it also depends on the use case I guess whether you even want that. Hence, why it _might_ be useful for redaction, but why I don't want to provide strong guarantees.

> Another point, is that the docs say that it can be useful for redacting content, yet it's not a perfect redaction. So this says to me a contradiction.

I don't think it's a contradiction. Whether it's useful depends on your needs and threat model.

laurmaedje:
I pushed another commit expanding the explanation a bit more. What do you think?

Andrew15-5:
Looks almost perfect. I've added last touches and their explanation.

Useful links:
- https://lingohelp.me/preposition-after-adjective/useful-for-or-useful-in-or-useful-to/
- https://forum.wordreference.com/threads/useful-in-or-for.3838091/
- https://pdfa.org/glossary-of-pdf-terms/#:~:text=Refers%20to%20the%20PDF%20feature%20called%20Document%20Outline%2C%20which%20is%20commonly%20known%20as%20Bookmarks%2e

crates/typst-library/src/layout/hide.rs
```diff
@@ -14,10 +14,10 @@ use crate::introspection::Tagged;
 /// ```
 ///
 /// # Redaction
-/// This function may also be useful to redact content as its arguments are
+/// This function may also be useful for redacting content as its arguments are
 /// neither present visually nor accessible to Assistive Technology. That said,
 /// there can be _some_ traces of the hidden content (such as a bookmarked
-/// heading in the PDF document outline).
+/// heading in the PDF's Document Outline).
```
laurmaedje, line 20: why title case here?

Andrew15-5: Because it's like that in the [glossary](https://pdfa.org/glossary-of-pdf-terms/#:~:text=Refers%20to%20the%20PDF%20feature%20called%20Document%20Outline%2C%20which%20is%20commonly%20known%20as%20Bookmarks%2e). I guess a proper noun, or something.

laurmaedje (approved):

laurmaedje:
Thanks!

crates/typst-library/src/layout/hide.rs
```diff
@@ -3,16 +3,25 @@ use crate::introspection::Tagged;
...
 /// ```
+///
+/// # Redaction
+/// This function may also be useful for redacting content as its arguments are
+/// neither present visually nor accessible to Assistive Technology. That said,
+/// there can be _some_ traces of the hidden content (such as a bookmarked
+/// heading in the PDF's Document Outline).
+///
+/// Note that, depending on the circumstances, it may be possible for content to
+/// be reverse engineered based on its size in the layout. We thus do not
```
bastienvoirin, line 23:
```suggestion
/// be reverse-engineered based on its size in the layout. We thus do not
```

Andrew15-5: It's not 100% accepted, as [TechTarget article](https://www.techtarget.com/searchsoftwarequality/definition/reverse-engineering) always uses the dash, [Meriam-Webster dictionary](https://www.merriam-webster.com/dictionary/reverse%20engineer) doesn't use the dash for any of the RE terms. And the [Wikipedia article](https://en.wikipedia.org/wiki/Reverse_engineering) only doesn't use the dash for "reverse engineering", yet still has 2 instances of "reverse engineered". So it's all over the place, same as backend/frontend, codebase, fullstack, etc.

Issue #7183, opened 2025-10-19
CLI argument `--pdf-standard` does not seem to support multiple PDF standards


ramkumarkb:
### Description

I am testing the `v0.14-rc2` release. The compiler  `help` says that - 

--pdf-standard <PDF_STANDARD>
One (or multiple comma-separated) PDF standards that Typst will enforce conformance with

But when I tried the below - 

```
typst compile --root ./ --pdf-standard a-3b,ua-1 ./my-project/test_pdfua.typ
error: Typst currently only supports one PDF substandard at a time
```

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

ramkumarkb:
@Enivex  - I am not sure if this is a `docs` issue. A PDF document needs to both PDF/A and PDF/UA-1 standard. 
In the `v0.14-rc2` - am able to choose either one and that is a problem. 

If I choose `ua-1`, then I do not get the blue notification bar (that the doucment is of Archive standard) when I open the document in Adobe Reader.

LaurenzV:
> A PDF document needs to both PDF/A and PDF/UA-1 standard.

Do you mean in the general case? If so, that's not true, it is perfectly fine for a document to be PDF/UA-1 but not PDF/A.

But to answer your question, yes, choosing two standards at the same time is currently not supported, unfortunately. Although it shouldn't be too hard to add. There is also an upstream issue about this: https://github.com/LaurenzV/krilla/issues/101

Enivex:
> @Enivex  - I am not sure if this is a `docs` issue. A PDF document needs to both PDF/A and PDF/UA-1 standard. 
> In the `v0.14-rc2` - am able to choose either one and that is a problem. 
> 
> If I choose `ua-1`, then I do not get the blue notification bar (that the doucment is of Archive standard) when I open the document in Adobe Reader. 

I guess it would be docs or feature request depending on the point of view, but not a bug at least.

ramkumarkb:
@LaurenzV  - Thank you for the feedback.  

Yes, in general - PDF/UA-1 and PDF/A are distinct.  However, as per what I understand of the EAA conformance -  a doc needs to be *both* PDF/A and PDF/UA-1. 

Some references (am sure you might have seen them) - 
1. EN 301 549 (see Section 10): https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf 
[...]

PR #7202, opened 2025-10-22
Document limitation to one PDF standard at a time

reknih:


mohe2015:
Is there a (current) technical reason?

laurmaedje:
It's just not implemented yet. We first need to add support in https://github.com/LaurenzV/krilla.

Marcono1234:
See also #7183

PR #7207, opened 2025-10-23
Add check for maximum parsing depth

saecki:
This should fix stack overflows in the parser, due to deeply nested expressions.
It's possible that I missed some code path that can infinitely recurse, but I think this *should* catch most issues and missing checks can easily be added later on.

A couple of fuzz reproduction cases are also added.

Maybe you can take a look @isuffix?


isuffix:
I'll look at this over the weekend!

crates/typst-syntax/src/parser.rs
```diff
@@ -1958,7 +1981,7 @@ struct PartialState {
 }
 
 /// The Memoization interface.
-impl Parser<'_> {
+impl<'s> Parser<'s> {
```
saecki: This is completely unrelated and should be reverted

crates/typst-syntax/src/parser.rs
```diff
@@ -1,14 +1,17 @@
...
-use typst_utils::default_math_class;
+use typst_utils::{default_math_class, defer};
 use unicode_math_class::MathClass;
 
 use crate::set::{SyntaxSet, syntax_set};
 use crate::{Lexer, SyntaxError, SyntaxKind, SyntaxMode, SyntaxNode, ast, set};
 
+// Picked by gut feeling.
+const MAX_DEPTH: usize = 256;
+
```
isuffix, line 14: Seems fine. The biggest real document I have didn't break until `MAX_DEPTH < 20`.

crates/typst-syntax/src/parser.rs
```diff
@@ -46,6 +49,9 @@ fn markup(p: &mut Parser, at_start: bool, wrap_trivia: bool, stop_set: SyntaxSet
 /// Parses a sequence of markup expressions.
 fn markup_exprs(p: &mut Parser, mut at_start: bool, stop_set: SyntaxSet) {
     debug_assert!(stop_set.contains(SyntaxKind::End));
+
+    let Some(p) = &mut p.check_depth_until(stop_set) else { return };
+
```
isuffix, line 53: nit: I think it looks better to remove the leading blank line for the `check_depth_until` calls.

crates/typst-syntax/src/parser.rs
```diff
@@ -1563,6 +1583,8 @@ struct Parser<'s> {
     /// backtracking similar to packrat parsing. See comments above in
     /// [`expr_with_paren`].
     memo: MemoArena,
+    /// The current expression nesting depth.
+    depth: usize,
```
isuffix: I don't personally care, but if we're following #6026 and #6942, we should consider using `u32` or `u64` here.

crates/typst-syntax/src/parser.rs
isuffix: If you're adding this for consistency with the other impl blocks you should also add it to the error impl block.

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2111,66 @@ impl Parser<'_> {
...
+    /// also generates nicer error messages.
+    fn check_depth_until(&mut self, stop_set: SyntaxSet) -> Option<&mut Self> {
+        if self.depth < MAX_DEPTH {
+            return Some(self);
+        }
+
+        self.depth_check_error(Some(stop_set));
+
+        None
+    }
```
isuffix, line 2140:
```suggestion
    fn check_depth_until(&mut self, stop_set: SyntaxSet) -> Option<&mut Self> {
        if self.depth < MAX_DEPTH {
            Some(self)
        } else {
            self.depth_check_error(Some(stop_set));
            None
        }
    }
```

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2111,66 @@ impl Parser<'_> {
...
+    fn increase_depth(&mut self) -> Option<impl DerefMut<Target = Self>> {
+        if self.depth < MAX_DEPTH {
+            self.depth += 1;
+            return Some(defer(self, |p| p.depth -= 1));
+        }
+
+        self.depth_check_error(None);
+
+        None
+    }
```
isuffix, line 2153:
```suggestion
    fn increase_depth(&mut self) -> Option<impl DerefMut<Target = Self>> {
        if self.depth < MAX_DEPTH {
            self.depth += 1;
            Some(defer(self, |p| p.depth -= 1))
        } else {
            self.depth_check_error(None);
            None
        }
    }
```

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2111,66 @@ impl Parser<'_> {
...
+
+    /// Generate an error for an exceeded maximum depth check.
+    ///
+    /// This function has to guarantee some sort of forward progress, otherwise
+    /// the parser might loop indefinitely. One token is eaten in all cases, if
+    /// that token is an opening delimiter, try to balance the opening and
+    /// closing grouping delimiters before continuing.
```
isuffix: I think this comment works better inside the function just before the loop.

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2111,66 @@ impl Parser<'_> {
...
+            }
+            p.eat();
+            balance == 0
+        };
+        self.with_nl_mode(AtNewline::Continue, |p| {
+            while {
+                let balanced = eat_balanced(p);
+                let at_stop = stop_set.is_none_or(|s| p.at_set(s));
+                !(balanced && at_stop || p.end())
+            } {}
```
isuffix: I think a `loop` with an explicit break is easier to follow here, especially because it removes the negation.

This also makes it slightly better to inline `eat_balanced`, but I'm fine with it as-is.

```suggestion
            loop {
                let balanced = eat_balanced(p);
                let at_stop = stop_set.is_none_or(|s| p.at_set(s));
                if balanced && at_stop || p.end() {
                    break;
                }
            }
```

crates/typst-syntax/src/parser.rs
```diff
@@ -46,6 +49,9 @@ fn markup(p: &mut Parser, at_start: bool, wrap_trivia: bool, stop_set: SyntaxSet
 /// Parses a sequence of markup expressions.
 fn markup_exprs(p: &mut Parser, mut at_start: bool, stop_set: SyntaxSet) {
     debug_assert!(stop_set.contains(SyntaxKind::End));
+
+    let Some(p) = &mut p.check_depth_until(stop_set) else { return };
```
isuffix: You can remove `&mut` on the `check_depth_until` calls.

crates/typst-syntax/src/parser.rs
saecki: Oh, no this was by accident.

crates/typst-syntax/src/parser.rs
saecki: `eat_balanced` must be called before the `stop_set` check, so that would require everything to be inlined.

crates/typst-syntax/src/parser.rs
saecki: Good point :+1:

crates/typst-syntax/src/parser.rs
saecki: So do you think we should use a smaller value?

isuffix: No, sorry. I just meant to affirm the choice.

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2108,66 @@ impl Parser<'_> {
...
+            None
+        }
+    }
+
+    /// Generate an error for an exceeded maximum depth check.
+    fn depth_check_error(&mut self, stop_set: Option<SyntaxSet>) {
+        let m = self.marker();
+
+        let mut balance: usize = 0;
+        let mut eat_balanced = |p: &mut Parser| -> bool {
```
laurmaedje: is there a particular reason this is in a closure?

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2108,66 @@ impl Parser<'_> {
...
+
+        // This function has to guarantee some sort of forward progress,
+        // otherwise the parser might loop indefinitely. One token is eaten in
+        // all cases, if that token is an opening delimiter, try to balance the
+        // opening and closing grouping delimiters before continuing.
+        self.with_nl_mode(AtNewline::Continue, |p| {
+            loop {
+                let balanced = eat_balanced(p);
+                let at_stop = stop_set.is_none_or(|s| p.at_set(s));
+                if balanced && at_stop || p.end() {
```
laurmaedje: could you add parentheses here to clarify precedence?

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2108,66 @@ impl Parser<'_> {
...
+                    break;
+                }
+            }
+        });
+
+        self.wrap(m, SyntaxKind::Text);
+        let node = self.nodes.last_mut().unwrap();
+        node.convert_to_error("maximum parsing depth exceeded");
```
laurmaedje: do we have no better way to create an error than wrapping in text and converting? seems a tad hacky.

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2108,66 @@ impl Parser<'_> {
         }
         self.nodes.drain(start..end);
     }
+
+    /// Check if the maximum depth has been exceeded. If so, generate an error
+    /// and try to make a best effort recovery using the `stop_set` as a guide.
+    ///
+    /// This function isn't strictly necessary, but it is an optimization and
```
laurmaedje: the second sentence gives just enough detail to spark curiosity, but not enough to resolve my resulting confusion :p what would the alternative to the function be?

crates/typst-syntax/src/parser.rs
```diff
@@ -2088,4 +2108,66 @@ impl Parser<'_> {
...
+            Some(self)
+        } else {
+            self.depth_check_error(Some(stop_set));
+            None
+        }
+    }
+
+    /// Check if the maximum depth has been exceeded. If so, generate an error
+    /// and try to make a best effort recovery.
+    /// Otherwise increase the depth and return a handle that automatically
```
laurmaedje: Could you either add a blank line / parbreak or rewrap?

crates/typst-syntax/src/parser.rs
saecki: The alternative would be to not have this function at all. That leads to very poor performance in some cases though. For example with deeply nested unary operators. I'll expand a bit on that.

crates/typst-syntax/src/parser.rs
saecki: Not anymore, I'll inline it

crates/typst-syntax/src/parser.rs
saecki: Now we do :)

laurmaedje:
Perfect, looks great now.

PR #7208, opened 2025-10-23
Add `lang` attribute to the default root HTML element

### LINKED ISSUE #7203: `#set text(lang: …)` should affect `<html lang="…">` to conform WCAG 2.1 level A
### Description

> Note: I [mentioned it on the forum](https://forum.typst.app/t/typst-0-14-just-entered-its-final-testing-period-your-help-is-wanted/6363/10?u=y.d.x) before.

In [Natural Language – Accessibility – Typst Documentation](https://staging.typst.app/docs/guides/accessibility/#natural-language):

> In order for screen readers to pronounce your document correctly and translation software to work properly, you must indicate in which natural language your document is written. Use the rule [`#set text(lang: "..")`](https://staging.typst.app/docs/reference/text/text/#parameters-lang) at the very start of your document or your template's capability to set a language.

This should also apply to HTML.

```typst
#set text(lang: "zh")
```

To be specific, the above typst document should generate the HTML below, with the `lang` attribute of `<html>` set to `zh`.

```html
<!DOCTYPE html>
<html lang="zh">
  <head>
    <meta charset="utf-8">
  </head>
  <body></body>
</html>
```

### Use Case

_Language of Page_ (`<html lang="…">`) is marked as level A in [W3C's Web Content Accessibility Guidelines (WCAG) 2.1](https://www.w3.org/TR/WCAG21/#language-of-page).

However in typst v0.14.0-rc.2, it looks like the only way to set `lang` for `<html>` is to recreate the whole document with `html.html(lang: "…", { html.head(…); html.body(…) })`.
Unfortunately, this will discard metadata in the [auto-generated `<head>`](https://github.com/typst/typst/blob/f1e8fd52f05861a711979f47fc865a6fcbeee6bd/crates/typst-html/src/document.rs#L183-L184).
In other words, to support WCAG 2.1 level A, we have to clear all built-in elements added by typst for accessibility…

YDX-2147483647:
Resolves #7203

Please see the comment on `text.region`: https://github.com/typst/typst/issues/7203#issuecomment-3437595764

---

Note that this PR only changes the default root HTML `<html>` element. It does not change the result for `A #text(lang: "zh")[人]`.

I think we can support the default root element first, because in v0.14.0-rc.2, setting the root's `lang` requires to recreate the whole document and discard all auto-generated metadata in `<head>`, but setting a `<span>`'s lang is as simple as replacing `text` with `html.span`.

### DIFF
--- a/tests/ref/html/text-lang-region.html
+++ b/tests/ref/html/text-lang-region.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="zh-TW">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
+    <nav role="doc-toc">
+      <h2>目錄</h2>
+      <ol style="list-style-type: none"></ol>
+    </nav>
+  </body>
+</html>
--- a/tests/ref/html/text-lang-unknown-region.html
+++ b/tests/ref/html/text-lang-unknown-region.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="zh-XX">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
+    <nav role="doc-toc">
+      <h2>目录</h2>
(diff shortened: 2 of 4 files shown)


laurmaedje:
Thanks!

Issue #5886, opened 2025-02-18
Also output a depfile with `--make-deps` when the typst compiler fails


tovrstra:
### Description

The current implementation of `--make-deps` only writes a depfile after a successful build. When the build fails, no such file is written. The proposal is to write a depfile regardless of the return code of Typst. Not all dependencies may be known yet when an error occurs, so the depfile would be potentially incomplete (but still useful) in case of a failed build. 

### Use case

This feature would be useful for integrating Typst with [StepUp](https://reproducible-reporting.github.io/stepup-core/stable/), an experimental publication build tool I'm developing. (Warning: dragons ahead. It is quite fresh but has been incredibly useful for a few users. I'm also using it for course materials and to automate grading and feedback for large groups.) With #5873 by @mtoohey31 just merged, the integration will be easier and more robust. There's just one edge case that I can't get to work yet because the depfile is only written if the Typst compilation succeeds. I'll summarize the edge case below, taken from the discussion in #3890:

> For a conventional Makefile or ninja.build file, it is sufficient to write a depfile only when compilation succeeds. Tools like make and ninja only use depfiles to add dependencies on source files, i.e. files authored by a human. They don't support depfiles that add dependencies on the output of other build rules or steps. If compilation fails, the user must fix the source files and rerun the build tool, which will try to compile with updated sources.
> 
> StepUp supports more general dependencies in depfiles, also dependencies on the output of other steps in the build process. (With make or ninja, such dependencies must be declared up front. With ninja, you can get around this limitation with a generator rule and dynamic depfiles, which I've been using for a while. Both still have their limitations.)
>
> A common chain of dependencies goes like this (a bit simplified):
>
> 1. raw data file -> some python script -> reduced file e.g. npz
> 2. reduced data file -> some other Python script (using matplotlib) -> `figure.svg`
> 3. (`figure.svg`), article source -> latex or typst -> article pdf
>
> Parentheses are used to emphasize that the dependency on `figure.svg` is not known a priori (unless I would implement heuristics, e.g. grepping for `#image`, which is not a robust solution).
>
[...]

PR #7209, opened 2025-10-23
Add outputs to JSON dependency file

### LINKED ISSUE #7194: Extend the `--deps-format json` to include output files
### Description

I've just tested the change in PR #7022. First off all: thanks @magistau and @laurmaedje because that PR is a great improvement! Not having to deal with Make assumptions and conventions facilitates the interaction with typst and getting this information also after a failed build helps a lot.

I would suggest to also include output file information when specifying the JSON format. For example, as follows in case of SVG outputs:

```json
{"input": ["document.typ", "figure.svg", ...], "output": ["page-1.svg", "page-2.svg", ...]}
```

There are few variations to consider, e.g. shorter

```json
{"i": ["document.typ", "figure.svg", ...], "o": ["page-1.svg", "page-2.svg", ...]}
```

or even shorter

```json
[["document.typ", "figure.svg", ...], ["page-1.svg", "page-2.svg", ...]]
```

### Use Case

The output files are not always trivially predictable. When typst generates separate output files for every page, the only reliable way to know all the outputs is to run typst.

The list of output files is useful information to include for build tools that _can_ use it. I'm developing such a build tool, [StepUp](https://reproducible-reporting.github.io/stepup-core/stable/), which we use for preparing papers , course materials, etc. However, I believe there can be other use cases too.

With the output information, build tools can do the following, which would be difficult (fragile mainly) to implement otherwise:

- **Automatic removal of outdated outputs.** For example, when a document compiled into separate SVG files becomes shorter and gets recompiled, knowledge of the old and new output files allow the build tool to remove the stale outputs.
- **Automatic dependency tracking.** When later steps in the build use the numberd SVGs (for separate pages), knowledge of typst's outputs allows the build tool to execute everything in order.

(StepUp supports both if it receives a list of generated outputs after typst exits.)

tovrstra:
Fixes #7194

I've generated some examples with their corresponding dependency files in all supported formats (for comparison) and the JSON is shown before (as in 0.14) and after (as in this PR). Practically all of these results seem fine, except maybe one, see first bullet in the discussion below.

---

**demo.typ**

```typ
#image("monkey.svg")
```

- Make: `demo.pdf: monkey.svg demo.typ`
- zero: `monkey.svg\0demo.typ\0`
- JSON before: `{"inputs":["monkey.svg","demo.typ"]`
- JSON after: `{"inputs":["monkey.svg","demo.typ"],"outputs":["demo.pdf"]}`

---

**lorem.typ** with SVG pages output (`typst compile lorem.typ lorem{p}.svg`)

```typ
#lorem(1000)
```

- Make: `lorem1.svg lorem2.svg: lorem.typ`
- zero: `lorem.typ\0`
- JSON before: `{"inputs":["lorem.typ"]}`
- JSON after: `{"inputs":["lorem.typ"],"outputs":["lorem1.svg","lorem2.svg"]}`

---

**lorem.typ** with PDF output to stdoutt (`typst compile lorem.typ -`)

```typ
#lorem(1000)
```

- Make: *no* dep file created, `error: failed to create dependency file (make dependencies contain the output path, but the output was stdout)`
- zero: `lorem.typ\0`
- JSON before: `{"inputs":["lorem.typ"]}`
- JSON after: `{"inputs":["lorem.typ"],"outputs":[]}`

---

**error.typ**

```typ
#thereisnosuchthing
```

- Make: *no* dep file created
- zero: `error.typ\0`
- JSON before: `{"inputs":["error.typ"]}`
- JSON after: `{"inputs":["error.typ"],"outputs":null}`

---

Discussion of the results:

- With the `zero` and `json` formats, the dependency file is always written, which is good. Some build systems can make use of this information to decide which dependencies need to be added to the dependency graph. (I've written one such build system, StepUp, but there are certainly others.)

- Including the outputs in the JSON has some use cases, in which I'm particularly interested, hence this PR. For some discussion on why (not) to include it, see #7194. I think it is generally useful for custom scripts that run typst and want to perform some operation on all its output files, without (i) removing old files first or (ii) reimplementing the logic of Typst. For build tools that construct there dependency graph on the fly, which is how StepUp works, this is also the safest future-proof way to determine which outputs Typst has written.

- When running some experiments with this feature, it turned out to be more convenient to have an empty list of `outputs` when there are none, e.g. when writing to stdout or when there is an error. That way, scripts consuming the JSON file don't need to implement a conditional to process the `outputs` field. (Such scripts can still check the exitcode of Typst if they want to handle failed compilations.) This also simplifies the code in typst-cli a tiny bit.

---

Outdated discussion (other PR merged)

- I believe, if no Make dependency file can be written, an empty file could be breaking one of Make's assumptions, see https://github.com/typst/typst/issues/5886#issuecomment-3423311187 Instead of an empty file, it seems better not to write it all. In any case, this is an unrelated issue and can be dealt with in a separate PR.
    - Issue: #7232 
    - PR: #7246

### DIFF
--- a/crates/typst-cli/src/deps.rs
+++ b/crates/typst-cli/src/deps.rs
@@ -42,7 +42,7 @@ fn write_deps_json(
 
     let inputs = relative_dependencies(world)?
         .map(|dep| to_string(dep, "input"))
-        .collect::<Result<Vec<_>, _>>()?;
+        .collect::<Result<_, _>>()?;
 
     let outputs = outputs
         .map(|outputs| {
@@ -55,7 +55,7 @@ fn write_deps_json(
                         Output::Stdout => None,
                     }
                 })
-                .collect::<Result<Vec<_>, _>>()
+                .collect::<Result<_, _>>()
         })
         .transpose()?;
 


tovrstra:
@laurmaedje This should be ready for review.

laurmaedje:
I've pondered for a bit and decided that I'd like to avoid a future breaking change, but don't want to add the new feature this very late in the release candidate cycle. For that reason, I've opened a minimized PR (https://github.com/typst/typst/pull/7212) instead that only moves into the `inputs` field and we can proceed with the `outputs` more calmly, discussing the `null` thing and how to deal with stdout (I think it's currently inconsistent with make deps).

tovrstra:
To facilitate the discussion, I've updated the PR after merging with main, and also added examples with all formats to the first post + some discussion.

crates/typst-cli/src/deps.rs
```diff
@@ -26,24 +26,51 @@ pub fn write_deps(
...
+    dest: &Output,
+    outputs: Option<&[Output]>,
+) -> io::Result<()> {
     let inputs = relative_dependencies(world)?
         .map(|dep| {
             dep.into_string().map_err(|dep| {
                 io::Error::new(
                     io::ErrorKind::InvalidData,
-                    format!("{dep:?} is not valid utf-8"),
+                    format!("input {dep:?} is not valid utf-8"),
```
bastienvoirin:
```suggestion
                    format!("input {dep:?} is not valid UTF-8"),
```

crates/typst-cli/src/deps.rs
```diff
@@ -26,24 +26,51 @@ pub fn write_deps(
...
+        for output in outputs {
+            match output {
+                Output::Path(path) => {
+                    outputs2.push(
+                        path.as_os_str()
+                            .to_str()
+                            .ok_or_else(|| {
+                                io::Error::new(
+                                    io::ErrorKind::InvalidData,
+                                    format!("output {path:?} is not valid utf-8"),
```
bastienvoirin:
```suggestion
                                    format!("output {path:?} is not valid UTF-8"),
```

crates/typst-cli/src/deps.rs
tovrstra: I've checked other error messages throughout the project and most seem to be using `utf-8` in lowercase, not all. While I'm not against uppercase, consistency is also nice. I'm going to let the maintainers decide on this one.

tovrstra:
I've merged with main to include the recently merged #7246, and I've updated the original post accordingly. The behavior of the make format is now a bit more internally consistent.

laurmaedje:
> When running some experiments with this feature, it turned out to be more convenient to have an empty list of outputs when there are none

I can see it, though it feels like we're taking information away. (I know, we can use the exit code.) I guess it depends on the consumer how annoying it is to convert null to `[]`. I am honestly unsure which is better.

> e.g. when writing to stdout or when there is an error. 

Isn't the stdout handling a bit inconsistent with make deps output?

tovrstra:
> > When running some experiments with this feature, it turned out to be more convenient to have an empty list of outputs when there are none
> 
> I can see it, though it feels like we're taking information away. (I know, we can use the exit code.) I guess it depends on the consumer how annoying it is to convert null to `[]`. I am honestly unsure which is better.

You're right. It would not be annoying really, and it's generally better to make the deps file more informative. Something along the following lines is easy to implement in Typst and easy enough to handle by scripts reading the JSON deps file:

- `"outputs": ["file1", "file2", ...]` if all went well and output files were written.
- `"outputs": []` if all went well and output was not written to file(s).
- `"outputs": none` if something went wrong and no output was written. This also implies the list of inputs could be incomplete.

This covers all possibilities if Typst is not writing output files when it encounters an error in the input. Is that a safe assumption (also for future releases)?

> > e.g. when writing to stdout or when there is an error.
> 
> Isn't the stdout handling a bit inconsistent with make deps output?

I think so, also with the modification suggested in this comment. However, this seems beneficial. It is an opportunity to make the JSON format maximally informative. E.g. partial dependency information is breaking Make's assumption, so this cannot be put in a Make deps file. With the JSON format, we don't have to worry about such restrictions, so it can serve use cases where Make's (or Redo's) assumptions are not relevant and possibly restrictive.

laurmaedje:
> This covers all possibilities if Typst is not writing output files when it encounters an error in the input. Is that a safe assumption (also for future releases)?

I think it should be fine.

> I think so, also with the modification suggested in this comment. However, this seems beneficial. It is an opportunity to make the JSON format maximally informative.

Alright!

tovrstra:
Original post has been updated to reflect the latest commit. (The case `error.typ` now has `"outputs": null`.)

crates/typst-cli/src/deps.rs
laurmaedje: let's stay with `"utf-8"`

laurmaedje:
Thanks!

PR #7211, opened 2025-10-24
Fix and improve regex documentation

Andrew15-5:
- For some reason, the period was inlined with the example before, but now it's a standalone period: https://staging.typst.app/docs/reference/foundations/regex/.
- Also it doesn't make sense to use colon if the example is hidden (is there no open-by-default syntax?).
- The line width was > 80.
- Use best practice, and explicitly state what it is.

Also not sure about saying that it also include `\t` and `\n`, although in string formatting you mustn't escape the backslash for them, which is kinda out of scope for regex.

### DIFF
--- a/crates/typst-library/src/foundations/str.rs
+++ b/crates/typst-library/src/foundations/str.rs
@@ -993,7 +993,7 @@ impl Regex {
         /// Typst, you need to escape the backslash itself (e.g., writing
         /// `{regex("\\\\")}` for the regex `\\`). Regex escape sequences that
         /// are not valid Typst escape sequences (e.g., `\d` and `\b`) can be
-        /// entered into strings directly, but it's a good practice to still
+        /// entered into strings directly, but it's good practice to still
         /// escape them to avoid ambiguity (i.e., `{regex("\\b\\d")}`). See the
         /// [list of valid string escape sequences]($str/#escapes).
         ///


crates/typst-library/src/foundations/str.rs
```diff
@@ -955,14 +955,17 @@ impl Regex {
...
+        /// sequences that are also valid in Typst (e.g. `[\\]`), you need to
+        /// escape twice. Thus, to match a verbatim backslash, you would need to
+        /// write `{regex("\\\\")}`. To prevent undesired side effects, it's a
+        /// good practice to always escape the backslash in regex tokens that
+        /// use it.
         ///
         /// If you need many escape sequences, you can also create a raw element
         /// and extract its text to use it for your regular expressions:
-        /// ```{regex(`\d+\.\d+\.\d+`.text)}```.
+        /// `{regex(`\d+\.\d+\.\d+`.text)}`.
```
YDX-2147483647: This becomes two parts in the generated HTML.

```html
<code>{regex(</code>\d+.\d+.\d+<code>.text)}</code>.
```


```suggestion
        /// ``{regex(`\d+\.\d+\.\d+`.text)}``.
```

Using two backquotes works.

```html
<code><span class="typ-func">regex</span><span class="typ-punct">(</span><span class="typ-raw">`\d+\.\d+\.\d+`</span><span class="typ-punct">.</span>text<span class="typ-punct">)</span></code>.
```

crates/typst-library/src/foundations/str.rs
```diff
@@ -933,7 +933,7 @@ fn string_is_empty() -> EcoString {
 /// #"a,b;c".split(regex("[,;]"))
 ///
 /// // Works with show rules.
-/// #show regex("\d+"): set text(red)
+/// #show regex("\\d+"): set text(red)
```
YDX-2147483647, line 969: According to GitHub Search, [740 files use `\d`](https://github.com/search?q=lang%3Atypst+%22%5C%5Cd%22&ref=opensearch&type=code), but only [49 files use `\\d`](https://github.com/search?q=lang%3Atypst+%22%5C%5C%5C%5Cd%22&ref=opensearch&type=code).

Andrew15-5: Maybe can be a separate PR, not sure.

crates/typst-library/src/foundations/str.rs
Andrew15-5: Sad that curly braces don't handle this, and weird that double backticks is actually a thing.

crates/typst-library/src/foundations/str.rs
Andrew15-5: @laurmaedje, what do you think? Technically, if this PR brings correctness, might as well make all other regex expressions follow the outlined best practices.

laurmaedje:
> Also it doesn't make sense to use colon if the example is hidden (is there no open-by-default syntax?).

The example was only folding at all because it used three backticks. So that should be fixed now.

> Use best practice, and explicitly state what it is.

I can see the upside of using explicit escaping, but I'm not sure I would certify it as best practice. I would probably reach for a raw text + `.text` before writing tons of backslashes. It's always super confusing to read...

Andrew15-5:
> I'm not sure I would certify it as best practice

Then how would you certify it? Must users learn every possible regex token with a backslash and every possible character that can be constructed from a backslash and a letter (and other chars) in a string to correctly avoid incorrect regex in Typst? I don't think so, which is why following this best practice fixes this issue. At the expense of one or two extra backslashes.

> I would probably reach for a raw text + `.text` before writing tons of backslashes.

It's already documented below (5-backslash-token example). What are you implying?

> It's always super confusing to read...

I wouldn't say always or super, especially if everyone will get used to writing it like that and reading the double backslashes more. But the change from JavaScript is real. Introduce the same sugar syntax for regex constructor `/whatever/` and this will be a non-issue.

laurmaedje:
> It's already documented below (5-backslash-token example). What are you implying?

I am implying that I think it's a better practice that tons of double backslashes.

Andrew15-5:
OK, so at what point should you use ```regex(`\...`.text)``` over `regex("\\...")`? Even if it's just `\d`? In what way is this supposed to be highlighted in the docs? There are 2 separate paragraphs right now.

It's also not a ton, it's definitely shorter if you do `("\\d\\.")` rather than ``(`\d\.`.text)`` (8 vs. 11 chars). Or `("\\d+")` vs. ``(`\d+`.text)`` (6 vs. 10 chars).

What about the other usages? https://github.com/typst/typst/pull/7211#discussion_r2459015773

crates/typst-library/src/foundations/str.rs
```diff
@@ -955,14 +955,17 @@ impl Regex {
...
-        /// escape sequences. To produce regex escape sequences that are also valid in
-        /// Typst (e.g. `[\\]`), you need to escape twice. Thus, to match a verbatim
-        /// backslash, you would need to write `{regex("\\\\")}`.
+        /// Most regex escape sequences just work because they are not valid
+        /// Typst escape sequences (e.g., `\d`, `\b`). To produce regex escape
+        /// sequences that are also valid in Typst (e.g. `[\\]`), you need to
+        /// escape twice. Thus, to match a verbatim backslash, you would need to
+        /// write `{regex("\\\\")}`. To prevent undesired side effects, it's a
+        /// good practice to always escape the backslash in regex tokens that
+        /// use it.
```
laurmaedje: If we intend to guide users towards double-escaping, I think the paragraph should lead to that differently. I also think "to prevent undesired side effects" is a bit vague. How about, this:

> Both Typst strings and regular expressions use backslashes for escaping. To produce a regex escape sequence that is also valid in Typst, you need to escape the backslash itself (e.g. writing `{regex("\\\\")}` for the regex `\\`). Regex escape sequences that are not valid Typst escape sequences (e.g., `\d` and `\b`) can be entered into strings directly, but it's good practice to still escape them to avoid ambiguity (i.e. `{regex("\\d")}`).

crates/typst-library/src/foundations/str.rs
laurmaedje: Fine by me

Andrew15-5: Sorry, in a separate or this PR is fine? Both? I brought up both options.

laurmaedje: I'm not sure how many places are affected? The GitHub search showed mostly third-party repos. In principle, I'd be okay with having this in the same PR.

crates/typst-library/src/foundations/str.rs
Andrew15-5: > I also think "to prevent undesired side effects" is a bit vague.

Yeah.

> How about, this

I think it's great. So is it just a replacement for the whole paragraph?

I also kinda want to clear that regex is RegEx, which is short for regular expression. I guess it's kinda obvious, but I have a feeling that something like "(RegEx)" once in the introductory sentence or in _the_ paragraph will be enough.

Like, the freshest example is that I didn't see that SUMO or SuMo is short for Support Mozilla for several times, until I read "(SUpport MOzilla)", and _then_ it hit me.

laurmaedje: > So is it just a replacement for the whole paragraph?

Exactly.

> I guess it's kinda obvious

I think it really is. The "RegEx" captialization is a bit odd to me and feels out of place to be honest.

Andrew15-5: > I think it really is. The "RegEx" captialization is a bit odd to me and feels out of place to be honest.

Eh... Then whatever, I guess.

crates/typst-library/src/foundations/str.rs
Andrew15-5: After looking at

```sh
rg -g '*.{typ,yml}' '\\[^ u]'
```

I don't see any edge cases. In

```sh
rg -g '*.rs' '\\[^ u]'
```

```sh
rg -g '*.rs' '//.*?\\[^ u]' | rg '\\.'
```

I see only some stuff in `str.rs`. About 7 lines.

```sh
rg --stats -g '*.{typ,yml}' 'regex.*?\\'
```

Gives another 39.

So less than 50 potential lines to change, I guess.

laurmaedje: You can then go ahead and change it. Though I think it would be worthwhile to keep a test case with the not-double-escaped version, specifically to test for that behavior.

Andrew15-5:
How about adding the link to [a valid list of escape characters in a string](https://typst.app/docs/reference/foundations/str/#escapes)?

Though "Just like in markup, you can escape a few symbols in strings:" sounds like you can write `\n` in markup, and it will be translated to a newline.

laurmaedje:
We can link there. The differences between string and markup escape sequences could be addressed in a separate PR.

Andrew15-5:
Why this exists and what do I do with it?

https://github.com/typst/typst/blob/3c8084bde1c78abd3d967f48d3f9c1fe057dcdad/tests/suite/foundations/str.typ?plain=1#L208

Andrew15-5:
I don't know where to place the test.

---

Wow, platform-dependent newline, how nice.

https://github.com/typst/typst/actions/runs/19194712375/job/54873981911?pr=7211

This must be a bug, since the source file doesn't have Carriage Return.

Though there is "/r" and not "\r"? On an empty line too?

crates/typst-library/src/foundations/str.rs
Andrew15-5: [Done](https://github.com/typst/typst/pull/7211/commits/0818cb306c9c7b9f8c5062427547991c7b66b223) and [done](https://github.com/typst/typst/pull/7211/commits/161c7157bd0e173f06cb5a32952f7dd2cae7df9d).

Also looked at `regex` in some kind of wrappers, but there is none of them, until now.

Andrew15-5:
> We can link there. The differences between string and markup escape sequences could be addressed in a separate PR.

Not 100% sure if the link is valid. Also: https://github.com/typst/typst/issues/7339.

laurmaedje:
> This must be a bug, since the source file doesn't have Carriage Return.

Maybe the Windows checkout does have it because Git?

crates/typst-library/src/foundations/str.rs
```diff
@@ -955,14 +955,18 @@ impl Regex {
...
-        /// Most regex escape sequences just work because they are not valid Typst
-        /// escape sequences. To produce regex escape sequences that are also valid in
-        /// Typst (e.g. `[\\]`), you need to escape twice. Thus, to match a verbatim
-        /// backslash, you would need to write `{regex("\\\\")}`.
+        /// Both Typst strings and regular expressions use backslashes for
+        /// escaping. To produce a regex escape sequence that is also valid in
+        /// Typst, you need to escape the backslash itself (e.g., writing
+        /// `{regex("\\\\")}` for the regex `\\`). Regex escape sequences that
+        /// are not valid Typst escape sequences (e.g., `\d` and `\b`) can be
+        /// entered into strings directly, but it's a good practice to still
```
laurmaedje:
```suggestion
        /// entered into strings directly, but it's good practice to still
```
this was changed from my proposal, but this one reads better to me (ChatGPT agrees and English is the one thing it's good at...)

crates/typst-library/src/foundations/str.rs
```diff
@@ -955,14 +955,18 @@ impl Regex {
...
+        /// escape them to avoid ambiguity (i.e., `{regex("\\b\\d")}`). See the
+        /// [list of valid string escape sequences]($str.escapes).
```
laurmaedje:
```suggestion
        /// [list of valid string escape sequences]($str/#escapes).
```

Andrew15-5:
> > This must be a bug, since the source file doesn't have Carriage Return.
> 
> Maybe the Windows checkout does have it because Git?

Well, AFAIK, people are always saying that you should preserve whatever the EOL file use, so changing the platform with correctly setup git config upon cloning won't change the EOL, nor will just downloading an archive of a repository. So either it's incorrectly set `lfrc` thing or idk, some yet another thing.

crates/typst-library/src/foundations/str.rs
Andrew15-5: Thanks, sometimes my English is at its best, sometimes it isn't. Generally speaking, you have to use an article, otherwise it would sound strange. And...it probably works most of the time, but I think it doesn't always apply, I feel like. So, as always, I'm guided by my 6th sense. Also, even though LLMs generally should excel at English, I think they are still sucky at some details. Maybe I'm misremembering.

Yeah, I've tweaked it a bit, because of what you've said.

Andrew15-5:
Didn't know that it's [`waiting-on-author`](https://github.com/typst/typst/issues?q=state%3Aopen%20label%3Awaiting-on-author). I was waiting on review, i.e., further comments.

Now it's blocked on https://github.com/typst/typst/pull/7418.

Andrew15-5:
@laurmaedje, can you re-run the CI or is it not a thing?

laurmaedje:
I don't think that will help without you rebasing on main to get the `.gitattributes` file. And once you rebase it will rerun anyway.

Andrew15-5:
Right... I thought it would run CI from main.

laurmaedje:
Thanks!

crates/typst-library/src/foundations/str.rs
```diff
@@ -993,7 +993,7 @@ impl Regex {
         /// Typst, you need to escape the backslash itself (e.g., writing
         /// `{regex("\\\\")}` for the regex `\\`). Regex escape sequences that
         /// are not valid Typst escape sequences (e.g., `\d` and `\b`) can be
-        /// entered into strings directly, but it's a good practice to still
+        /// entered into strings directly, but it's good practice to still
```
Andrew15-5, line 996: You've said that the new one reads better?

laurmaedje: Sorry, I meant that the old one reads better. My other comment was incorrectly written. But ChatGPT indeed agreed that "good practice" is preferrable over "a good practice".

Andrew15-5: To be honest, in practice, there is little to no difference anyway. "Good practice" sounds like a "proper noun", but "a good practice" is more of "one of many good practices."

Yeah, having two "this" that refer to different things is very confusing indeed. Which is why I understood it differently.

> this was changed from my proposal, but this one reads better to me

Issue #7225, opened 2025-10-25
specify compiler options via in-file metadata header


gasche:
### Description

Many options that affect the build of a Typst document can be passed via command-line options to `typst compile` or `typst watch`. For example: `--root <root>`, `--ppi <PPI>`, `--features <feature>`. But sometimes it would be much easier, in particular for tooling support, to be able to specify those options directly in the file, in some header metadata:

```typst
#document-header(
  root: "..",
  features: ["html", "ally"],
)

(rest of the document)
```

The document-header comment must be placed first in the file (with optional comments before and inside), so that it can be located without interpreting the whole file. (A bespoke lexical syntax for metadata could also be designed, see [Pandoc metadata block](https://pandoc.org/MANUAL.html#extension-yaml_metadata_block).)

If options are provided via both the command-line and the document header, then the file header settings take precedence.

### Use Case

My Emacs `typst-ts-watch` mode fails on the following error for a given file:

```
error: failed to load file (access denied)
    ┌─ slides.typ:652:14
    │
652 │ #bibliography("../suspended.bib", title: none, style: "./gasche-author-date.csl")
    │               ^^^^^^^^^^^^^^^^^^
    │
    = hint: cannot read file outside of project root
    = hint: you can adjust the project root with the --root argument
```

Unfortunately `typst-ts-watch` does not currently provide support for specifying the `--root` parameter when it invokes `typst watch`. I will of course request such support in that particular tool, but there are dozens of tools out there (editor modes, linters, build system integrations, etc.) that will each need to be extended to support specific command-line options for the `typst` CLI, while these options could be passed once and for all in the file itself.

cady-b:
This would violate the security model[^security]: https://typst.app/open-source#faq:~:text=We%20take%20special%20care%20to%20ensure
It would effectively allow any document to read arbitrary files on the system (and, for example, embed them invisibly into the output, thereby exfiltrating data).

However, you can already ([since v0.3.0](https://typst.app/docs/changelog/0.3.0/#command-line-interface)) set the `TYPST_ROOT` environment variable to change the root without messing about with the final stages of your tooling. `TYPST_FEATURES` is available to enable features.  
Version 0.14.0 introduced the `typst info` subcommand, which allows you to inspect this and other information. The manual and `--help` also show which flags are available (next to their corresponding arguments), i.e., https://man.archlinux.org/man/extra/typst/typst-compile.1.en

[...]

PR #7227, opened 2025-10-25
(Proof of Concept) Add `sys.defaults` element

T0mstone:
I couldn't help myself and implemented a proof of concept for some of what I mentioned in https://github.com/typst/typst/issues/7225#issuecomment-3446661343 and https://github.com/typst/typst/issues/4174#issuecomment-3446668549.

This adds the `sys.defaults` element with fields `format` and `features` that correspond to the `--format` and `--features` CLI options. When the CLI options are not specified, these values serve are used as the fallback, rather than a hard-coded PDF and no-features as it was before.

I left out adding something for `--ppi` since that might be better as a target-specific API like described in #4174 rather than living in `sys`.

It took quite a bit of rearchitecting to make this work and the `features` thing still doesn't work yet because features have to be available earlier than the output format (they are already needed at world creation, but the world is needed to evaluate the document to get `sys.values.features`). One idea would be to perform the first evaluation with a world that doesn't have the features enabled, but that also wouldn't really work, e.g. if you wanted to enable the `html` feature, you'd certainly make use of it in the document, but the first "no feature" run that would be needed to determine that you have `"html" in sys.defaults.features` would already error out due to your use of the `html` API, preventing Typst from finding out that it would actually have been ok...

Anyway, other than that, the code is quite ad-hoc, hence why I'm only calling this a proof of concept, tho at least `sys.defaults.format` does fully work.

### DIFF
--- a/crates/typst-realize/src/lib.rs
+++ b/crates/typst-realize/src/lib.rs
@@ -609,6 +609,8 @@ fn visit_styled<'a>(
         } else if elem == DocumentElem::ELEM {
             if let Some(info) = s.kind.as_document_mut() {
                 info.populate(&local)
+            } else if let RealizationKind::SysDefaults { .. } = s.kind {
+                // do nothing
             } else {
                 bail!(
                     style.span(),
@@ -621,7 +623,11 @@ fn visit_styled<'a>(
                 info.populate_locale(&local)
             }
         } else if elem == PageElem::ELEM {
-            if !matches!(s.kind, RealizationKind::LayoutDocument { .. }) {
+            if !matches!(
+                s.kind,
+                RealizationKind::LayoutDocument { .. }
+                    | RealizationKind::SysDefaults { .. }
+            ) {
                 bail!(
                     style.span(),
                     "page configuration is not allowed inside of containers"


T0mstone:
ok I suppose I should have run the tests first...😅

T0mstone:
Yeah okay, I'm too tired for this now. Clearly the extra realization step I introduced causes some issues, but idk what exactly the root cause is. Maybe I'll find the issue later.

If it turns out to be irredeemable, then oh well, at least future work can look to this to see how some parts could work and what issues can come up.

laurmaedje:
We actually have somewhat different plans. Internally, this goes under "output API" and it's also what will power multi-file HTML output.

Essentially, the idea is to have a new `output` element (or maybe multiple for different targets) that describes a compilation output (e.g. a PDF, a HTML document, ..). Its properties would subsume the defaults in this PR. Usage of the output elements would of course be optional and could be completed with CLI flags. Multi-file outputs could result in a directory (could be a mix of HTML, PDF, and even plain text files).

I intend to write a blog post about this concept, but haven't yet gotten around to it. However, I'm fairly certain that we'll do _something_ in this direction.

T0mstone:
Sounds exciting. I'll close this then.

Issue #5695, opened 2025-01-14
Make `sym.angle` callable to produce angle brackets


MDLC01:
### Description

Make `sym.angle` callable to wrap its argument in angle brackets, similar to, e.g., [`math.floor`](https://typst.app/docs/reference/math/lr#functions-floor).

### Use Case

Wrapping a math expression in angle brackets is currently very cumbersome: `$lr(angle.l x angle.r)$`. This proposal would make it easier: `$angle(x)$`.

### Potential issue

The way callable symbols currently work is by associating some characters with a function. To implement this feature request, the character ∠ (`sym.angle`) would have to be associated with a function wrapping a body inside angle brackets. This is somewhat weird, as ∠ doesn't have much to do with angle brackets semantically.

Enivex:
I don't think this is a good idea. Both because the symbols are different, and also because I'm not convinced no one is using the angle symbol in a function-like manner. (Though I was unable to find an explicit example.)

laurmaedje:
I agree that using a non-matching symbol is problematic and that one might want to use the angle function-like. Moreover, floor(x) is good since it's semantic but angle(x) is not, so one should still define a semantic wrapper ideally. The probability of anyone doing this though is actually reduced if we add angle(x).

MDLC01:
Ignoring the issue of not having a semantic name, an alternative that does not require associating a function with an unrelated symbol is to use `angles` instead.

balthild:
I guess it is used as Dirac bra-ket notation? Maybe it's better to call it `braket(x)` in this situation, as well as `bra(x)` and `ket(x)`.

FYI the `physics` package for LaTeX calls it `\braket`, but it is `⟨x|x⟩` rather than `⟨x|x⟩`.

http://mirrors.ctan.org/macros/latex/contrib/physics/physics.pdf

![Image](https://github.com/user-attachments/assets/cc2dc530-21fb-43d3-83c0-4766ec6c2842)

MDLC01:
In my case, I mainly use angle brackets tu denote tuples, (e.g., ⟨𝑥, 𝑦⟩).

Enivex:
It's used for a lot of things. Inner products, dual pairings, averages, tuples, etc.

toniuyt123:
`#let bracket(body) = $angle.l #body angle.r$`

I use this workaround.

MDLC01:
[...]

PR #7228, opened 2025-10-25
Make more delimiter symbols callable

MDLC01:
Closes https://github.com/typst/typst/issues/5695.

This PR makes almost all symbols with the Opening Unicode math class[^1] callable to produce a `math.lr` expression.

I omitted three symbols that have the Opening math class because they are not yet in Codex: U+298B ⦋ LEFT SQUARE BRACKET WITH UNDERBAR (matching U+298C), U+2993 ⦓ LEFT ARC LESS-THAN BRACKET (matching U+2994), and U+2995 ⦕ DOUBLE LEFT ARC LESS-THAN BRACKET (matching U+2996).

I also omitted U+3014 and U+3018, which are commented our of the math class file.[^1] They are both CJK brackets which shouldn't be used in math.[^2]

In the future, we might also want to make top and bottom brackets (such as U+23DE ⏞ TOP CURLY BRACKET) callable to produce results similar to `math.underbrace` and `math.overbrace` with an empty `annotation` (related to https://github.com/typst/typst/pull/4637).

To simplify the source code, I created a declarative macro to easily create left/right functions. To keep the code simple, I decided to slightly alter the documentation for the body parameter of existing left/right functions, changing, e.g., "The expression to floor." to simply "The expression to wrap." I also changed the example for `math.norm` to present a more common use-case.

[^1]: https://www.unicode.org/Public/math/revision-15/MathClass-15.txt
[^2]: Section 2.5 in https://www.unicode.org/reports/tr25/tr25-15.pdf.


crates/typst-library/src/math/lr.rs
```diff
@@ -35,84 +35,101 @@ pub struct MidElem {
...
-    /// The expression to take the norm of.
-    body: Content,
-) -> Content {
-    delimited(body, '‖', '‖', size)
+    /// Takes the absolute value of an expression.
+    ///
+    /// ```example
+    /// $ abs(x/2) $
+    /// ```
+    pub abs: '|', '|';
```
Leedehai: I recall Typst intentionally decided to not make `|...|` an LR group, unless explicitly called by `abs()`. This is because pairing the vertical bars is difficult.

laurmaedje:
This is reaching dimensions where it might make more sense to generate the functions dynamically, as is done for the typed HTML API in `typst-html/typed.rs`. That would feel a bit cleaner to me and also be better for compile times and (probably just a tiny bit) for binary size.

MDLC01:
I'm wondering whether it would make more sense to store those left/right pairs in Codex somehow

MDLC01:
in the end, I decided to add the pairs that don't yet have a name in Codex because I see no reason not to, and that way it will already work when we give them names in Codex.

laurmaedje:
Do you think the accents could also make use of the auto-generated function approach? (Note necessarily in this PR, could also be separate.)

MDLC01:
Yes they probably should as well.

laurmaedje:
I pushed two commits. I'm happy to merge now and have the rest separately, but if you want to do it all together, that's fine with me too.

crates/typst-library/src/math/lr.rs
```diff
@@ -193,9 +200,7 @@ fn create_lr_func(
             move |_: &mut Engine, _: Tracked<Context>, args: &mut Args| {
                 let size = args.named("size")?;
                 let body = args.expect("body")?;
-                args.take().finish()?;
-                let output = delimited(body, left.clone(), right.clone(), size);
-                ::typst_library::foundations::IntoResult::into_result(output, args.span)
+                Ok(delimited(body, left.clone(), right.clone(), size).into_value())
```
MDLC01: Oh I forgot to clean up the code I copied from the `#[func]` macro, sorry for that.

But do you not need the `args.take().finish()?` part here to prevent the user from passing too many arguments?

tests/suite/math/delimited.typ
```diff
@@ -181,3 +181,15 @@ $
...
+$ bracket.l(x, size: #200%) $
+
+--- math-lr-sym-call-extra-arg ---
+// Error: 16-28 unexpected argument: nope
+$ bracket.l(x, nope: "nope") $
+
+--- math-lr-sym-func-repr ---
+// The outline thing is just a roundabout way to force a cast from symbol to
+// function...
+#test(repr(outline(indent: sym.chevron.l.curly).indent), "(..) => ..")
```
MDLC01, line 200: Hmm, I kinda hopped there wouldn't be a way to get the function without immediately calling it.

The title and docs are empty, which may appear weirdly in an LSP documentation.

MDLC01:
I'll try to implement the accent change tonight

tests/suite/math/delimited.typ
laurmaedje: Adding the symbol itself as the title and some short docs would be pretty simple I guess.

crates/typst-library/src/math/lr.rs
laurmaedje: It's done automatically one level higher up, see the test I added for this.

MDLC01:
Turns out it's not as easy as I expected. I'll come back to it tomorrow.

crates/typst-library/src/math/lr.rs
```diff
@@ -107,18 +127,131 @@ pub fn abs(
...
+        .collect()
+});
+
+/// Creates metadata for an L/R wrapper function.
+fn create_lr_func_data(
+    left: EcoString,
+    right: EcoString,
+    bump: &'static Bump,
+) -> NativeFuncData {
+    let title = bump.alloc(format!("{left}{right} Left/Right"));
```
laurmaedje: This could use https://docs.rs/bumpalo/latest/bumpalo/macro.format.html + https://docs.rs/bumpalo/latest/bumpalo/collections/string/struct.String.html#method.into_bump_str to avoid the intermediate allocation.

crates/typst-library/src/math/accent.rs
```diff
@@ -106,89 +127,127 @@ impl Accent {
...
-                /// The size of the accent, relative to the width of the base.
-                #[named]
-                size: Option<Rel<Length>>,
-                /// Whether to remove the dot on top of lowercase i and j when
-                /// adding a top accent.
-                #[named]
-                dotless: Option<bool>,
-            ) -> Content {
-                let mut accent = AccentElem::new(base, Accent::new($primary));
+// Keep it synced with the documenting table above.
```
MDLC01, line 135: Ideally I would like to generate the table from this slice automatically, but this can be done in a future PR.

docs/src/lib.rs
```diff
@@ -905,8 +905,7 @@ fn symbols_model(resolver: &dyn Resolver, group: &GroupData) -> SymbolsModel {
                 }),
                 value: value.into(),
                 // Matches casting `Symbol` to `Accent`
-                accent: value_char
-                    .is_some_and(|c| typst::math::Accent::combine(c).is_some()),
+                accent: typst::math::Accent::combine(value).is_some(),
```
MDLC01: I'm not sure this was true before (and it's not now either) because the cast uses `Accent::new`, not `Accent::combine`.

crates/typst-library/src/math/accent.rs
```diff
@@ -106,89 +127,127 @@ impl Accent {
...
+/// Creates metadata for an accent wrapper function.
+fn create_accent_func_data(accent: char, bump: &'static Bump) -> NativeFuncData {
+    let title = bumpalo::format!(in bump, "Accent ({})", accent).into_bump_str();
+    let docs = bumpalo::format!(in bump, "Adds the accent {} on an expression.", accent)
+        .into_bump_str();
+    NativeFuncData {
+        function: NativeFuncPtr(bump.alloc(
+            move |_: &mut Engine, _: Tracked<Context>, args: &mut Args| {
+                let base = args.expect("base")?;
+                let size = args.named("size")?;
```
MDLC01, line 178: I'm not sure why a test is failing, as I should be reading the argument properly here...

crates/typst-library/src/math/accent.rs
```diff
@@ -82,9 +88,24 @@ pub struct AccentElem {
...
+        }
+        bail!("expected a single codepoint")
+    }
+
+    /// Normalize an accent to a combining one.
+    pub fn combine(s: &str) -> Option<char> {
+        ACCENTS
+            .iter()
+            .copied()
+            .find(|(_, names)| names.contains(&s))
```
laurmaedje: The test failure seems unrelated to the arguments. The problem is that `names` is empty here for `dot.triple` and `dot.quad`. I think this should also allow for `s` itself to be in the accent list as the char.

MDLC01: Thanks! That was indeed the problem

crates/typst-library/src/math/accent.rs
```diff
@@ -82,9 +88,26 @@ pub struct AccentElem {
...
+    }
+
+    /// Normalize an accent to a combining one.
+    pub fn combine(s: &str) -> Option<char> {
+        ACCENTS
+            .iter()
+            .copied()
+            .find(|(accent, names)| {
+                accent.encode_utf8(&mut [0; 4]) == s || names.contains(&s)
+            })
```
MDLC01: This is a linear search, but the array is small so I don't think this is problematic.

Alternatively, I could use a `HashMap` like for the functions, or a sorted list.

MDLC01:
Actually, the failing test was due to the fact that `dash` is the em dash, which was in the match expression in `symbol.rs` before, but wasn't in `accent.rs`.

crates/typst-library/src/math/accent.rs
laurmaedje: Linear search is okay

crates/typst-library/src/math/accent.rs
```diff
@@ -106,89 +129,127 @@ impl Accent {
...
+    ('\u{20dc}', &[]),
+    ('\u{030a}', &["∘", "○"]),
+    ('\u{030b}', &["˝"]),
+    ('\u{030c}', &["ˇ"]),
+    ('\u{20d6}', &["←"]),
+    ('\u{20d7}', &["→", "⟶"]),
+    ('\u{20e1}', &["↔", "⟷"]),
+    ('\u{20d0}', &["↼"]),
+    ('\u{20d1}', &["⇀"]),
+];
```
laurmaedje: Why not use chars here in the array instead of strings? They are all single codepoints after all. Same with lr.

Maybe I'm missing something, but I gave this a rough try in https://github.com/typst/typst/commit/ca137bfd830a83e9b50e3b6f623493c680b01d6e and it seems simpler to me.

MDLC01: This is simply for consistency with Codex symbols, which are represented as strings.

And the reason why the first element of the pair is a char is because I don't think it makes sense for it to be a string given that it acts as a combining character.

Also, I like using a pair instead of a single array, because it makes it more obvious that the first element serves a different purpose: it is the character that will actually be combined, while the other ones are simply aliases. Essentially, this represents a map from non-combining accents to their combining form.

laurmaedje: I'm not arguing against the pair, just against the strings. If we never expect strings with more than one codepoint here, in my opinion, commit https://github.com/typst/typst/commit/ca137bfd830a83e9b50e3b6f623493c680b01d6e makes things simpler and more efficient.

MDLC01: It doesn't seem highly unlikely to me that we will need a string at some point in the future. I also don't see a major benefit in terms of code simplicity. But given that this is easy to change when/if needed, I'm fine with either solution.

When you say "more efficient", are you thinking of `char` to UTF-8 conversion and the size of a `char` compared to the size of an EcoString? Or am I missing something?

laurmaedje: > When you say "more efficient", are you thinking of char to UTF-8 conversion and the size of a char compared to the size of an EcoString? Or am I missing something?

Yeah, that. Of course, the efficiency point is barely relevant, but it just seems overall more straightfoward to me. Basically, I think it's good practice to use the most narrow type that expresses the concept we need. Just like using a `NonZeroUsize` instead of a `usize` is better for a page number.

MDLC01: Let's go with ca137bfd830a83e9b50e3b6f623493c680b01d6e then

laurmaedje:
Code-wise this is ready to merge, but I was just wondering whether we should also extend this to symbols of type fence? I'm not necessarily arguing for it, but I'd like to know the reasoning for/against, if any.

I'd also like to get @isuffix's input on the whole change (I saw the eyes emoji reaction, but didn't know how to interpret it.)

crates/typst-library/src/math/lr.rs
```diff
@@ -138,11 +139,23 @@ pub fn norm(
 /// Gets the Left/Right wrapper function corresponding to a left delimiter, if
 /// any.
 pub fn get_lr_wrapper_func(left: char) -> Option<Func> {
-    FUNCS.get(&left).map(Func::from)
+    match left {
+        // Unlike `round`, `abs`, and `norm`, floor` and `ceil` are of type
```
MDLC01:
```suggestion
        // Unlike `round`, `abs`, and `norm`, `floor` and `ceil` are of type
```

crates/typst-library/src/math/lr.rs
```diff
@@ -138,11 +139,23 @@ pub fn norm(
...
+        l => FUNCS.get(&l).map(Func::from),
+    }
 }
 
 /// The delimiter pairings supported for use as callable symbols.
 const DELIMS: &[(char, char)] = &[
+    // The `ceil` and `floor` pairs are omitted here because they are handled
+    // manually.
     ('⌈', '⌉'),
     ('⌊', '⌋'),
```
MDLC01: Unless I'm missing something, they look very much not omitted.

MDLC01:
> I was just wondering whether we should also extend this to symbols of type fence?

I thought about adding fences but decided not to for now because I felt like it is less clear what, e.g., `fence.dotted(x)` does. I don't have a strong opinion though.

crates/typst-library/src/math/lr.rs
laurmaedje: oops

isuffix:
This looks good to me, appreciate your work on this change!

I think I'm also in favor of making fences callable (what else would `$fence(x)$` be for?), but I'm not strongly either way.

laurmaedje:
I think I'm also in favor of making fences callable, too. It seems consistent and I don't really see a downside.

MDLC01:
I made all fences callable,[^1] with the exception of U+2982 ⦂ Z NOTATION TYPE COLON because I don't think it is ever used as a delimiter. Also, it does not have a name in Codex.

Also note that U+299A ⦚ VERTICAL ZIGZAG LINE is the only other fence that doesn't have a name in Codex. Instead, we have `fence.l` and `fence.r` for U+29D8 ⧘ LEFT WIGGLY FENCE and U+29D9 ⧙ RIGHT WIGGLY FENCE respectively. I added it anyway.

[^1]: You can find all fences in the following list by searching for the word "F" (this will yield false-positives, but a reasonable enough amount that it can be sorted by hand). Turns out there isn't a lot. Additionally, Typst doesn't define any new fences as part of the symbols it overrides the math class of. https://www.unicode.org/Public/math/latest/MathClassEx-15.html

laurmaedje:
Thank you!

Issue #1880, opened 2023-08-07
Footnotes on headings are duplicated in table of contents


jdonland:
### Description

A footnote on a heading will be duplicated in the table of contents.

### Typst version

- [X] I am using the latest version of Typst

moinsr:
Citations are duplicated too, would be nice to have an optional filter for that stuff in outlines.

moinsr:
I'll leave a crude workaround for now
```
#{
  set footnote.entry(
    separator: none
  )
  show footnote.entry: hide
  show ref: none
  show footnote: none
  
  outline()
  pagebreak()
  outline(
    target: figure.where(kind: image),
  )
  // etc.
}
```

laurmaedje:
This also applies to figure captions.

Also linking https://github.com/typst/typst/issues/1295, which is a bit related.

laurmaedje:
Just completely leaving things out seems somewhat problematic as a default to me. Unless they are at the end of the heading or caption, they would just leave a missing piece. For citations, it's relatively easy to get that behaviour if you want it (as you show above), but unless we add an extra toggle, it would be easy to opt-out of them being shown by default. For footnotes, it's unfortunately trickier to get rid of them since `show footnote: none` only hides the number, not the footnote at the bottom (Whether that should be a considered a bug or not, I am not sure. At first glance yes, but looking deeper `show heading: none` also leaves headings queriable and that seems like a good thing to me. We probably just need a way to control this since both things are sometimes desirable. For footnotes it's less obvious that introspection is involved, but for endnotes it would again be clearer.).

Caellian:
Very untested workaround:

<details>

```typst
#let custom-outline(
  spacing: 6pt,
  ..targets
) = context {
  let entries = ()

  let enumerate-target = (target) => {
    let counter = counter(target)
    return query(target).map(it => (
      counter.at(it.location()).at(0), it
    ))
  }
  for target in targets.pos() {
    entries = (..entries, ..enumerate-target(target))
  }

  let footnote-counter = counter(footnote)
  let footnote-lookup = query(footnote).map(it => it.location())

  // Nested entry items aren't querried so they don't have a location (yet?), so
  // sanitize-entry-body has to be impure
  let child-mutations = state("outline-entry-body-sanitizaion-mutations", (:))

  let sanitize-entry-body(
    body,
    location
  ) = {
    // directly forwarded component functions
    let forward = (text, ref)
    if type(body) == str {
      return body
    }
    if type(body) != content {
[...]

PR #7236, opened 2025-10-27
Strip footnote from outline entries

gabriel-araujjo:
This PR adds a solution for issue #1880.

The approach is quite naive, but it works for the most common use case:

```typst
#outline()

= Heading #footnote[note]
```


laurmaedje:
This will break very easily in various cases, so I'd rather not introduce this hack. With https://github.com/typst/typst/issues/7165, it would be easy to get the desired fix cleanly, though as mentioned in https://github.com/typst/typst/issues/1880#issuecomment-2582564120, I'm not 100% sure what the best behavior would be. Thank you still for the PR!

PR #7237, opened 2025-10-27
Various `str` constructor improvements

MDLC01:
This adds a hint for when trying to convert a positive integer to base 1 (unary).

I also made the hint that triggers when specifying a base with a value is not an integer trigger as well when `base` is explicitly set to `10`.

Additionally, I slightly improved the documentation to be more explicit.

### DIFF
--- a/crates/typst-library/src/foundations/str.rs
+++ b/crates/typst-library/src/foundations/str.rs
@@ -855,6 +855,8 @@ cast! {
     v: Str => Self::Str(v),
 }
 
+/// Similar to `Option<i64>`, but the default value casts to `10` rather than
+/// `none`, so that the right default value is documented.
 #[derive(Debug, Copy, Clone)]
 pub enum Base {
     Default,


crates/typst-library/src/foundations/str.rs
```diff
@@ -154,21 +154,28 @@ impl Str {
         value: ToStr,
         /// The base (radix) to display integers in, between 2 and 36.
         #[named]
-        #[default(Spanned::new(10, Span::detached()))]
-        base: Spanned<i64>,
+        #[default(Spanned::new(Base::Default, Span::detached()))]
+        base: Spanned<Base>,
```
laurmaedje, line 158: This could probably just use an `Option` to handle the absence case

crates/typst-library/src/foundations/str.rs
```diff
@@ -154,21 +154,28 @@ impl Str {
...
                 }
                 s
             }
             ToStr::Int(n) => {
-                if base.v < 2 || base.v > 36 {
+                let b = base.v.value();
+                if b == 1 && n > 0 {
+                    bail!(
+                        base.span, "base must be between 2 and 36";
+                        hint: "generate a unary representation with `\"1\" * {}`", n
```
laurmaedje, line 172: is that how unary works. if consistent, base 1 would only have the digit 0, wouldn't it? this seems soo niche to me, that I'm not sure the hint really makes sense.

crates/typst-library/src/foundations/str.rs
MDLC01: The idea behind this `Base` type is that the default is still displayed as `10` in the documentation.

crates/typst-library/src/foundations/str.rs
MDLC01: Unary is indeed not a regular base in that it is not a positional numeral system. Its main use that I am aware of is in complexity theory, when you want the time complexity of a Turing machine to be in the value of an input instead of its logarithm (because the _size_ of an integer encoded in binary is in its log base-2).

Although any symbol can be used, I've almost only seen "1".

https://en.wikipedia.org/wiki/Unary_numeral_system

The hint in indeed quite niche, but I like it because a user that wants a unary representation of a number, upon realizing that the `str` constructor does not allow that, might re-implement string multiplication instead of using the native syntax.

crates/typst-library/src/foundations/str.rs
laurmaedje: aha. seems slightly hacky, but I see now why.

crates/typst-library/src/foundations/str.rs
laurmaedje: I see.

crates/typst-library/src/foundations/str.rs
```diff
@@ -848,6 +855,29 @@ cast! {
...
+            Self::User(b) => b,
+        }
+    }
+}
+
+cast! {
+    Base,
+
+    self => self.value().into_value(),
+
```
laurmaedje: could you remove the extra blank lines here?

crates/typst-library/src/foundations/str.rs
MDLC01: I added a doc comment to `Base` to clarify that.

laurmaedje:
Thanks!

PR #7238, opened 2025-10-27
Error when passing both `end` and `count` to `{str,array}.slice`

MDLC01:
`end` and `count` are said to be mutually exclusive in the documentations, so a runtime error is probably for the best.


laurmaedje:
Thanks!

PR #7241, opened 2025-10-27
Use ems instead of points for default `math.cancel` metrics

MDLC01:
As per [the discussion on Discord](https://discord.com/channels/1054443721975922748/1175894504146993304/1432379451009667072), this PR changes `math.cancel`'s default metrics to use ems instead of points. This makes the line scale appropriately with font size, whereas before it could become overly thick with small font sizes.

This affects the length and thickness of the line, including in math scripts. In the image below, the new metrics (in green) are overlaid over the old metrics (in magenta) for in `display`, `inline`, `script`, and `sscript`.

image

Code (for Typst 0.14.0):
```typ
#set page(width: auto, height: auto, margin: 0.5cm)

#let test = $
  display(cancel(x)) \
  inline(cancel(x)) \
  script(cancel(x)) \
  sscript(cancel(x))
$

#place(text(fill: rgb("#ff00ff"), test))

#set math.cancel(length: 100% + 0.3em, stroke: 0.05em)
#text(fill: rgb("#00ff00"), test)
```


laurmaedje:
Thanks! Makes sense as a simple start to potentially more ems everywhere.

PR #7242, opened 2025-10-27
Minor fixes to documentation groups

MDLC01:
This PR changes the name of the Calculation group to "Calculations" (plural) because it can be used of calculations in general, instead of referring to a single specific calculation.

I also took the opportunity to use title case for the Standard library group, which I forgot in https://github.com/typst/typst/pull/6936.

Additionally, I removed all NBSPs from every file unless they were used properly.

### DIFF
--- a/docs/reference/groups.yml
+++ b/docs/reference/groups.yml
@@ -138,7 +138,7 @@
     the constants `pi`, `tau`, `e`, and `inf`.
 
 - name: std
-  title: Standard library
+  title: Standard Library
   category: foundations
   path: ["std"]
   details: |


laurmaedje:
I'm not sure I agree that it should be in plural. Fwiw, the page title (in the tab bar) would also be broken because it is "Calculation Functions" right now. Though that's also a problem for "Variants Functions" in math.

MDLC01:
I don't have a strong opinion so I can leave that part out of this PR

laurmaedje:
Let's leave that out then.

laurmaedje:
Thanks!

PR #7246, opened 2025-10-28
Do not create empty Make deps file when writing to standard output

### LINKED ISSUE #7232: Empty make dependency file when output is written to stdout
### Description

This issue can be reproduced with any example, e.g. `lorem.typ`

```typ
#lorem(1000)
```

Run typst (0.14) as follows

```bash
typst compile lorem.typ --deps lorem.dep --deps-format make -
```

This ends with the following error message as expected:

```text
error: failed to create dependency file (make dependencies contain the output path, but the output was stdout)
```

In addition, an empty `lorem.dep` file is written (zero bytes). As far as I understand, this goes against Make's assumption that dependency files are not written when a build fails.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

tovrstra:
Fixes #7232

Pattern from https://stackoverflow.com/a/50732452

### DIFF
--- a/crates/typst-cli/src/deps.rs
+++ b/crates/typst-cli/src/deps.rs
@@ -64,7 +64,7 @@ fn write_deps_make(
     dest: &Output,
     outputs: &[Output],
 ) -> io::Result<()> {
-    let mut dest = dest.open()?;
+    let mut buffer = Vec::new();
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks!

Issue #4011, opened 2024-04-25
CJK brackets at the beginning of paragraph


minamotorin:
### Description

When paragraph indentation is enabled, positioning of brackets at the beginning of paragraph doesn't follow [§3.1.5 of Requirements for Japanese Text Layout](https://www.w3.org/TR/2020/NOTE-jlreq-20200811/#positioning_of_opening_brackets_at_line_head).
Example code is below.

```typst
// This program is public domain (CC0).
#set page(
  width: 210mm,
  margin: (left: 30mm, right: 210mm-30mm-10em),
)
#set par(
  leading: 1em
)
#show par: set block(
  spacing: 2.5em
)

「日日日日日日日日日「日

#set par(first-line-indent: 1em)

「日日日日日日日日「日

#h(-0.5em)「日日日日日日日日「日
```

![brac](https://github.com/typst/typst/assets/76122224/8c71aede-8f58-410c-bffe-8893b8c31209)

The first paragraph is an example of paragraph indentation is **not** enabled, and the second one is an example of paragraph indentation is enabled.

The third paragraph is an example of the expected result that follows [§3.1.5 of Requirements for Japanese Text Layout](https://www.w3.org/TR/2020/NOTE-jlreq-20200811/#positioning_of_opening_brackets_at_line_head).

Thanks.

### Operating system

macOS

### Typst version

- [X] I am using the latest version of Typst

Leedehai:
https://typst.app/docs/reference/model/par/#parameters-first-line-indent mentioned "Only the first line of a consecutive paragraph will be indented (not the first one in a block or on the page)." -- a limitation, yes.. Typst wants to change that but it's not trivial.

laurmaedje:
I think the issue here differs from #311. If I'm understanding correctly, it's more that a bracket shouldn't disrupt the grid-like alignment of characters.

minamotorin:
The key point of this issue is that the space between left edge of paragraph and “「” is effectively 1.5 em, not the 1 em specified by `#set par(first-line-indent: 1em)`.

peng1999:
> The third paragraph is an example of the expected result that follows [§3.1.5 of Requirements for Japanese Text Layout](https://www.w3.org/TR/2020/NOTE-jlreq-20200811/#positioning_of_opening_brackets_at_line_head).

[...]

PR #7247, opened 2025-10-28
Fix auto hanging-indent for `〔1〕` heading numbering

### LINKED ISSUE #6527: Indentation on multiline headings with full-width character
### Description

```typst
#set heading(numbering: "1)")

= A very, very, very, very, very, very, very, very, very, very, very, very, very, very, very long heading

#set heading(numbering: "1）") // fullwidth brace

= A very, very, very, very, very, very, very, very, very, very, very, very, very, very, very long heading
```

Image

Note: The red content was marked by myself.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

YDX-2147483647:
Resolves a part of #6527.

Previously, if the heading numbering starts or ends with a CJ punctuation and its width is allowed to be shrunk or stretched, then `hanging-indent: auto` resolves to the wrong length, resulting in multi-line headings misaligned.

This PR amends the logic of `measure(heading.numbering).width` to fix that.

```typst
#set page(height: auto, width: 12em, margin: 1em)
#set text(lang: "zh", region: "CN", font: "Noto Serif CJK SC", fallback: false)

#set heading(numbering: "〔1〕")
= 犹可说也不可说也

#set heading(numbering: "一、")
= 犹可说也不可说也

#set heading(numbering: "一口")
= 犹可说也不可说也

#set heading(numbering: "一、")
#show heading: it => {
  show h.where(amount: 0.3em): none
  it
}
= 犹可说也不可说也
```

| Before (v0.15.1) | After (this PR) |
|---|---|
| 图片 | 图片 |

- The first two examples are improved
- The third example is already optimal and unchanged
- The fourth example is still not optimal and may be dealt with future PRs. At present, it can be circumvented by `#set heading(hanging-indent: 2em)`.

## Previous discussions

See [Typographic measure theory](https://typst.app/project/r0nMXEsKNnLfq8VZ4913fg) (or the [backup on the forum](https://forum.typst.app/t/joke-typographic-measure-theory-with-an-additional-joke/6656)) for more testing.

[Discord, 2025-10-30](https://discord.com/channels/1054443721975922748/1088371867913572452/1433351468496326738):

> Hi! There is a 0.3-em spacing between the heading number and the body. At present, the spacing is [weak](<https://typst.app/docs/reference/layout/v/#parameters-weak>).
Does anyone know why it has to be weak? Could it be changed to a strong spacing? How about `h(0.3em, weak: true) + h(0pt, weak: false)`?
> 图片

> It is weak to eat up a potential leading space in the body. For inline gaps that are intended to be precise, I prefer weak for that reason.


laurmaedje:
What's the state of this PR?

YDX-2147483647:
It fixes the original issue, but does not work well with `show h.where(amount: 0.3em): none`.
In practice, there's no space after `一、`, so `show h.where(…): none` is usually (if not always) necessary.
Therefore, this PR alone does not fix the real problem, yet introduces a breaking change.

Also the implementation depends on [quirks of `measure`](https://forum.typst.app/t/joke-typographic-measure-theory-with-an-additional-joke/6656/2).

I guess there exists a better fix, so I've marked it as a draft.

laurmaedje:
I think the supposed quirk of measure is rather a quirk of either horizontal spacing or CJ punctuation adjustment. `h(0pt)` will be discarded during inline layout, while it seems like `h(0.1pt)` will prevent the adjustment.

From a Western perspective, the patch here seems unproblematic, but if it doesn't really solve the problem for CJ users, that's unfortunate. I can't properly judge it, but from gut feeling perhaps the CJ punctuation adjustment is the root cause and needs change? I'd overall like to see the CJK handling in inline layout reworked as it has various subtle issues and is not that cleanly implemented.

YDX-2147483647:
I haven't carefully inspected how the CJ punctuation adjustment is implemented; but based on what DeepWiki told me, I think it looks logical. It's just that the punctuation adjustment doesn't rhyme with higher-level adjustment (e.g., justification).

Higher-level adjustment is irrelevant in this case, because the issue occurs even if justification is disabled and linebreaks are controlled manually.


I believe the root cause lies in this comment:
https://github.com/typst/typst/blob/bf946178ec3cb34c06b4666ea237a025b4ca4aa0/crates/typst-layout/src/rules.rs#L267-L268

1. We don't have a locator after the numbering, so we have to resort to `measure`.
2. The `measure` infrastructure can't measure things _in context_. It has to isolate the object before measurement.
3. The space occupied by the content may vary depending on the preceding and following content.

laurmaedje:
Having a locator wouldn't really change anything about measurement for this case. It lets things resolve "in context" in the sense that introspections resolve properly, but it does not let measure peak ahead and see that a weak space would not collapse.

Overall, I'm not sure how to proceed here. As I mentioned, the current diff looks unproblematic, but you mentioned that it does not fully address the issue. I think we should come to a decision here on whether we actually think we can merge something here or whether to close this.

YDX-2147483647:
Oh, my opinion has changed a bit. Someone once came across a use case in their real life that involves full-width punctuations and don't need to remove the 0.3em space (but I can't recall the speicifc case right now…).
In other words, this PR does indeed solve some real problems, so I now think it's worth merging.

I'll check my chat history, replace the currently suboptimal test case with that case, and then mark this PR as ready for review.

**Update:**
That case is `numbering: "〔1〕"`, reported by 盒子Bruce on 2025-06-17.

**Update 2:**
Somehow, this PR does not fix the case for `〔1〕`.
heading-hanging-indent-auto-cj-punct

**Update 3:**
The cause looks like to be https://github.com/typst/typst/issues/4011

<details>
<summary>Test</summary>

```typst
#set page(height: auto, width: 120pt, margin: 10pt)
#set text(lang: "zh", region: "CN", font: "Noto Serif CJK SC", fallback: false)

#set text(10pt)
#show heading: set text(10pt, weight: "regular")
#set page(width: 12em, margin: 1em)
#set heading(numbering: "〔1〕")

= #context here().position().x;犹可说也不可说也

〔1〕#h(0.3em, weak: true)#context here().position().x;犹可说也不可说也

#h(0.00001pt)〔1〕#h(0.3em, weak: true)#context here().position().x;犹可说也不可说也

#block(inset: (left: 22.71pt))[
  #h(-22.71pt)〔1〕#h(0.3em, weak: true)#context here().position().x;犹可说也不可说也
]

#context here().position().x
- #context measure("〔1〕" + h(0.3em)).width
- #context measure("〔").width
- #context measure("1").width
- #context measure("〕").width
```

图片



</details>

YDX-2147483647:
Hi! I've revised the code and update the PR description. It's ready for review now.

laurmaedje:
Thanks!

YDX-2147483647:
~~Don't merge yet!~~ In the previous version, the two spaces are different! One positive, one negative.

**Update:** Oh, I see what's going here… Never mind.

laurmaedje:
I felt a weak space was a _bit_ cleaner since it doesn't just cancel out, but actually disappears during layout.

Issue #7249, opened 2025-10-28
Typst uses fake superscripts for some characters in Source Serif 4 despite available typographic glyphs


apcamargo:
### Description

When using the [Source Serif 4](https://github.com/adobe-fonts/source-serif/releases/tag/4.005R) font, Typst appears to substitute fake superscript numerals even though the font includes proper typographic superscript glyphs.

In the example below, typographic superscripts are correctly used when the superscript contains only lowercase letters (`test`), but Typst switches to fake superscripts when the superscript includes numerical characters or parentheses.

<img width="236" alt="Image" src="https://github.com/user-attachments/assets/31348aa2-af91-4c6d-9d25-6261d9d8d573" />

As discussed in #7122, Typst is expected to fall back to fake superscripts only when real superscript glyphs are unavailable. However, Source Serif 4 provides genuine superscript glyphs for both numerals and parentheses, so this fallback should not occur.

For reference, here is how the output should look when the correct typographic superscripts are used:

<img width="246" alt="Image" src="https://github.com/user-attachments/assets/06abdd99-7c8f-4467-9cc2-d2ced73f6f90" />

### Reproduction URL

https://typst.app/project/rp5IR98kh7MhJX32ewRNHC

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

Enivex:
> For reference, here is how the output should look when the correct typographic superscripts are used:

How is this generated?

apcamargo:
> How is this generated?

I used Affinity Designer for this one, as it doesn't generate fake superscripts.

Enivex:
Based on https://discord.com/channels/1054443721975922748/1088371867913572452/1432761680089186455 this is a font issue

khaledhosny:
See my comment in https://github.com/typst/typst/pull/5777#pullrequestreview-3389349264, I don’t believe this is a font issue. The lack of Unicode mapping for superscript glyphs is irrelevant (the typst code does not check for it, and even if it did that would be a typst bug in itself).

frankrolf:
Thanks @khaledhosny – I explain why most subscripts don’t have Unicode mapping in my comment here: https://github.com/adobe-fonts/source-serif/issues/146#issuecomment-3461957850

frankrolf:
FWIW @Enivex the Discord you link to is not publicly accessible, so it might be worth pasting the reasoning here.

PR #7258, opened 2025-10-29
More robust detection of `subs` / `sups` font support

MDLC01:
Fixes https://github.com/typst/typst/issues/7249. Requires https://github.com/typst/typst-dev-assets/pull/17 for a regression test.

As explained in https://github.com/typst/typst/pull/5777#discussion_r2469986086 and https://github.com/typst/typst/pull/5777#discussion_r2469989087, the issue was that we only considered the first subtable from the first lookup when testing for coverage.

As suggested in https://github.com/typst/typst/pull/5777#discussion_r2473621965, I now check for coverage given a set of glyphs instead of testing for each individual codepoint in the string. I reorganized the code a bit to implement this change. The drawback is that we now have to shape before testing for coverage, in which case we shape a second time. In the future, if we decide to implement synthesized smallcaps similarly, we will shape the same text three times if the font defines both `sups` and `smcp` and the user writes `#super[smallcaps[...]]`.

https://github.com/typst/typst/pull/5777#discussion_r2473627804 also suggested that we should take the script and language into account. I'm not sure how to do that when testing for coverage. However, we do it when reshaping because we just shape from scratch.

I would appreciate a review from @khaledhosny because I'm still learning about all this so I can't say for sure I didn't miss something.

## Before/after comparison

```typ
#set text(font: "Source Serif 4")
#set super(typographic: true)
Foo#super[test] \
Foo#super[test1] \
Foo#super[(test)] \
// Source Serif 4 does not support `sups` for backticks, so this should be synthesized.
Foo#super[test\`]
```

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/tests/suite/text/shift.typ
+++ b/tests/suite/text/shift.typ
@@ -99,5 +99,5 @@ Foo#super[test] \
 Foo#super[test1] \
 Foo#super[(test)] \
 // Source Serif 4 does not support `sups` for backticks, so this should be
-// syntheszed.
+// synthesized.
 Foo#super[test\`]


crates/typst-layout/src/inline/shaping.rs
```diff
@@ -897,56 +897,57 @@ fn shape_segment<'a>(
...
+            );
+            ctx.features.pop();
+        } else {
+            // We have to synthesize the glyphs :(
+            let script_metrics = settings.kind.read_metrics(font.metrics());
+            script_shift = settings.shift.unwrap_or(script_metrics.vertical_offset);
+            script_compensation = script_metrics.horizontal_offset;
+            scale = settings.size.unwrap_or(script_metrics.height);
+        }
+    };
```
MDLC01, line 949: I felt like this was now small enough to be inlined here, but I can extract it back into a function if wanted.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -1079,62 +1080,78 @@ fn shape_segment<'a>(
...
+        return false;
+    };
+    let is_covered = |g| {
+        feature_table
+            .lookup_indices
+            .into_iter()
+            .flat_map(|lookup_index| {
+                gsub.lookups
+                    .get(lookup_index)
+                    .expect("font provided lookup index should be valid")
```
MDLC01, line 1141: I'm unsure whether `expect` can fail if the font is malformed, or if that would have been caught earlier (maybe when parsing the font).

I can look into it later.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -1079,62 +1080,78 @@ fn shape_segment<'a>(
...
+            })
+            .any(|coverage| coverage.contains(g))
+    };
+    glyphs
+        .glyph_infos()
+        .iter()
+        // The cast should always work as per `glyph_id`'s doc comment.
+        // I'm surprised `hb_glyph_info_t::as_glyph`, which performs this exact
+        // cast, isn't public...
+        .all(|g| is_covered(GlyphId(g.glyph_id as u16)))
```
MDLC01, line 1154: The fact that I have to do this cast manually makes me not-so-confident that this is right.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -1079,62 +1080,78 @@ fn shape_segment<'a>(
...
+        buffer.script(),
+        buffer.language().as_ref(),
+        features,
+    );
+
+    rustybuzz::shape_with_plan(font.rusty(), &plan, buffer)
+}
+
+/// Tests whether a feature covers all the glyphs of a buffer for a font.
+fn are_glyphs_covered(glyphs: &GlyphBuffer, font: &Font, feature: Tag) -> bool {
```
khaledhosny, line 1127: This still does not take script and language into account.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -897,56 +897,57 @@ fn shape_segment<'a>(
...
-        ctx.features.pop();
-    }
+    // Now that we know which glyphs we use, test whether the font provides
+    // `subs`/`sups` version if needed.
+    let mut script_shift = Em::zero();
+    let mut script_compensation = Em::zero();
+    let mut scale = Em::one();
+    if let Some(settings) = ctx.shift_settings {
+        if settings.typographic
+            && are_glyphs_covered(&buffer, &font, settings.kind.feature())
```
khaledhosny, line 927: Thinking about it more. This approach might make other things worse. For example if “fi” is used as superscript, or if `smcp` or `onum` features are enabled. With the new code, the glyphs will be reported uncovered (and rightly so) by `sups` feature, however in practice the `sups` feature actually works in these cases because the `sups` lookups come first in the font before `liga`, `smcp` and `onum` lookups, so when the feature is activated its substitution is applied first. But this depends on the font, and nothing prevents the font from having `sups` lookups later.

A possibly more robust solution is to actually shape the text with the `sups`/`subs` feature once enabled and once disabled, and if all the glyph IDs are different then use the feature, otherwise do the fallback. This has the advantage of checking what the font is actually doing (with all the complexities of OpenType script, language, and lookup order interactions) instead of second-guessing it, but it might be significantly slower (needs to be measured).

Alternatively, you can go back to checking coverage of unshaped glyphs as it seems to work better in this case, until someone hits a font where it doesn’t.

MDLC01: I think the solution you proposed in your second paragraph can still cause an issue in the following case. If a font provides:
- A glyph A for Latin Small Letter I without `sups`.
- A glyph B for Superscript Latin Small Letter I without `sups`, and the same glyph B for Latin Small Letter I with `sups`.
- A glyph C for Superscript Latin Small Letter I with `sups`.

When we want to shape `"iⁱ"` in superscript (Latin Small Letter I followed by Superscript Latin Small Letter I), the following happens:
- We shape it without `sups` and get the glyphs A B.
- We shape it with `sups` and get the glyphs B C.

There is a glyph in common, so we reject, but actually the font fully supported that.

In this case, it might look like we can compare only the glyphs with the same indices in their respective sequences, but in general, shaping with and without `sups` may yield a different number of glyphs, so we have to compare all pair of glyphs.

Maybe as you mentioned it would be better to just check coverage of unshaped glyphs.

An alternate solution when we want to shape superscript text with a set of features `F` might be the following:
- Find which features `F_before` have their lookups before `sups` in the font.
- Shape with `F_before`.
- Test for coverage for `sups`; if `sups` is fully covered at this point:
    - Shape with `F ∪ {"sups"}`.
- Otherwise:
    - Shape with `F` and synthesize the superscripts after.

Does this seem like a reasonable approach to you?

khaledhosny: > I think the solution you proposed in your second paragraph can still cause an issue in the following case. If a font provides:
> 
> * A glyph A for Latin Small Letter I without `sups`.
> 
> * A glyph B for Superscript Latin Small Letter I without `sups`, and the same glyph B for Latin Small Letter I with `sups`.
> 
> * A glyph C for Superscript Latin Small Letter I with `sups`.
> 
> 
> When we want to shape `"iⁱ"` in superscript (Latin Small Letter I followed by Superscript Latin Small Letter I), the following happens:
> 
> * We shape it without `sups` and get the glyphs A B.
> 
> * We shape it with `sups` and get the glyphs B C.
> 
> 
> There is a glyph in common, so we reject, but actually the font fully supported that.
> 
> In this case, it might look like we can compare only the glyphs with the same indices in their respective sequences, but in general, shaping with and without `sups` may yield a different number of glyphs, so we have to compare all pair of glyphs.

My suggestion was not to compare the set of glyphs in each case (that obviously wouldn’t work), but to compare the output of each code point. This can be tricky, as you say, but not impossible. One would need to take clustering into account when doing the comparison (e.g. segment each buffer based on HarfBuzz clusters, and compare these).



> An alternate solution when we want to shape superscript text with a set of features `F` might be the following:
> 
>  * Find which features `F_before` have their lookups before `sups` in the font.
> 
>  * Shape with `F_before`.
> 
> * Test for coverage for `sups`; if `sups` is fully covered at this point:
>       
>   * Shape with `F ∪ {"sups"}`.
> 
> * Otherwise:
>       
>   * Shape with `F` and synthesize the superscripts after.
> 
> 
> Does this seem like a reasonable approach to you?

The order in which lookups are applied is not only determined by the font. The shaper have some features that gets their lookups applied in specific order (depends on the script). It would be a non trivial logic to replicate.

khaledhosny:
> [#5777 (comment)](https://github.com/typst/typst/pull/5777#discussion_r2473627804) also suggested that we should take the script and language into account. I'm not sure how to do that when testing for coverage. However, we do it when reshaping because we just shape from scratch.

You would need to get the `Script` from `gsub.scripts` (falling back to `DFLT` script), then get the `Language` from `Script.languages` (falling back to `Script.default_language`), then get the feature indices from `Language.feature_indices` and locate the feature with the requested feature tag.

crates/typst-layout/src/inline/shaping.rs
laurmaedje: It can definitely panic. Font parsing is very lazy with ttf-parser.

laurmaedje:
I renamed this and opened https://github.com/typst/typst/pull/7462 because this PR is (1) not complete yet and (2) anyways a bit too large and dangerous for inclusion in a patch release, so I'd wanted to have a more minimal fix that can be backported either way. We can then carefully do the comprehensive fix here.

laurmaedje:
I'll close this for now due to inactivity. If you end up getting back to this, I'm happy to reopen or have a new PR!

PR #7271, opened 2025-10-30
Move equation `Eval` impl to `math.rs`

isuffix:
I've wanted to do this for a while, but it was always part of a larger commit/change, and I realized I can just factor it out as its own PR.


laurmaedje:
My thinking was that the dollars itself are still part of the markup syntax rather than the math syntax and only the contents (AST node `Math`) are where the math syntax begins.

isuffix:
The main benefit to me is to make it easier to reason about how equations evaluate. I figure if you want to know how `EquationElem` is generated, you're going to start in `math.rs`. Also, equations can be written directly in code, so they don't necessarily come from markup.

laurmaedje:
Fine by me!

PR #3854, opened 2024-04-02
Warn about unknown font families during parsing

niklasmohrin:
First progress on #1967

Edit: This started out as a PR to error when `fallback: true`, but we decided to first add a warning when encountering an unknown font, regardless of `fallback`. Original description below:

---

As discussed in the issue, I will try to emit errors first before thinking about warnings. This PR introduces a check for the easiest situation:

```typst
#text(font: "non-existing", fallback: false, "Hello")
```

yields

```plain
error: font family not found: non-existing
  ┌─ ex2.typ:1:6
  │
1 │ #text(font: "non-existing", fallback: false, "Hello")
  │       ^^^^^^^^^^^^^^^^^^^^
  │
  = hint: set `fallback: true` to use another font instead
```

However, the slightly more complicated case

```typst
#text(font: "Linux Libertine", fallback: false, text(font: "non-existing", "Hello"))
```

does not produce an error. This is because the check is currently in `TextElem::construct` and only inspects the direct arguments, not the inherited style.

Another example of this is

```typst
#text(font: "non-existing", text(fallback: false, "Hello"))
```

I think this is a bit unsatisfying, but better than the status-quo. Do you think it will be difficult to do the check later when style is inherited? Or am I missing an obvious way to inspect parent styles in the constructor here?

I had some trouble with getting the span of the `font` argument, because as far as I can tell, the `TextElem` does not know the spans of its arguments after construction, so I had to get the span just before setting the style.

Issue #6010, opened 2025-03-05
Disable _font not found_ warning if alternative font is found.


jounathaen:
### Description

I have written a template that uses the CMU fonts. However, as on some systems there is the alternative Font "New Computer Modern" installed, that is a perfectly valid alternative for all use-cases I have. So in the template I have the following code:

```typst
  set text(
    font: (
      "CMU Sans Serif",
      "New Computer Modern Sans",
    ),
    size: 12pt,
  )
```

Unfortunately, since https://github.com/typst/typst/issues/2818 this produces the following warning, since on this system, only the CMU fonts are installed:

```
warning: unknown font family: new computer modern sans
    ┌─ template/foobar.typ:105:10
    │  
105 │       font: (
    │ ╭───────────^
106 │ │       "CMU Sans Serif",
107 │ │       "New Computer Modern Sans",
108 │ │     ),
    │ ╰─────^
```

Maybe I'm following an antipattern here, but I think it'd be great if the warning is only thrown if no valid font (for actually used character sets) is found.

### Use Case

Described above.

janosh:
also, would be great if the warning could specify which fallback font is being used instead such that users can change to that one to silence the warning

YDX-2147483647:
## Previous discussions on the Forum

[Is it possible to suppress the warning about Noto CJK / Source Han duality? - Questions - Typst Forum](https://forum.typst.app/t/is-it-possible-to-suppress-the-warning-about-noto-cjk-source-han-duality/2207?u=y.d.x)

> I have a rough feeling that there are two kinds of font fallback.
> 
> * **Fallback for environments**: the same document for different browsers, OS’s , machines, etc.
>   You want to select one available font in a list of similar fonts.
> * **Fallback for graphemes**: different languages (or scripts?) in the same document.
>   Your document contains Latin, Han, and Emoji characters, and you want to combine several completely different fonts to cover them all.
>
> The font fallback mechanism of CSS is mainly/originally designed for the first kind, while that of typst mainly designed for the second kind?


A (not quite usable) workaround: `typst compile … --input HOSTNAME=$(hostname)`.

## A possible interface?

```typst
#set text(
  font: (
    (name: "CMU Sans Serif", if-unknown: "ok"),
    "New Computer Modern Sans",
    (name: "…", if-unknown: "warning"),
  ),
)
```

YDX-2147483647:
[...]

PR #7272, opened 2025-10-30
Do not warn if at least one font in `text.font` is available

gasche:
See the discussion in https://github.com/typst/typst/pull/3854#issuecomment-3444779769


gasche:
I rebased the PR to please some CI checks, with only minor changes.

laurmaedje:
Since this is problematic as is, and we'll likely need something that includes some form of user input as to how the fallback is intended to be used (see https://github.com/typst/typst/issues/6010#issuecomment-3554848249), I'll close this for now. We can have a new PR once we've settled on a design. Thanks for your work nonetheless!

PR #7276, opened 2025-10-31
Always treat `TextElem` fragments as spaced in math

### LINKED ISSUE #7270: Single-letter strings inside math mode are not sensitive to spaces around them
### Description

Single-letter strings in math mode are not sensitive to spaces around them like multi-letter strings. This can be seen with the following snippet, where we're trying to use strings to typeset units:
```typst
$0.3 "W"$

$0.3"W"$

$0.3 "kW"$

$0.3"kW"$
```
Which produces the following output:

Image

Is this intended behavior? I find this to be odd. I expected single-letter and multi-letter strings to have the same behavior with surrounding spaces. Someone mentioned it might be a remnant of #5738. 

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

isuffix:
Closes #7270

Seems like this was just a behavior from before #5738 that was never scrutinized in the transition.


laurmaedje:
Thanks!

PR #7277, opened 2025-10-31
Support for multiple bibliographies

thuiop:
This implements support for multiple bibliographies, as wanted in #1097 and discussed on the Discord.

It introduces two new parameters to the `#bibliography` element:
- `target` indicates what is covered by the bibliography. It can either be a selector of `cite` elements, or auto (default case). In the latter case, priority is given to the bibliographies with selectors; then, each citation is assigned to the next auto bibliography which contains the citation, or the previous one if it cannot be found.
-  `kind` can be either "shared" or "standalone" (actually anything other than "shared" right now); in the first case, the numbering will keep across bibliographies, in the second one the bibliography will start from one.

This requires https://github.com/typst/hayagriva/pull/409 to be merged in order for the numbering to work correctly.


crates/typst-library/src/foundations/selector.rs
```diff
@@ -141,6 +141,36 @@ impl Selector {
             Self::Regex(_) | Self::Before { .. } | Self::After { .. } => false,
         }
     }
+
+    /// Changes the element type of the selector
+    pub fn change_element(&self, elem: Element) -> Self {
```
laurmaedje: This does not seem to be in use anymore.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -119,6 +119,21 @@ pub struct BibliographyElem {
...
+    ///
+    /// This can be:
+    /// - A selector of `cite` elements. The citations it selects will be
+    ///   included in the bibliography.
+    pub target: Smart<Selector>,
+
+    /// When using multiple bibliographies in a single document, indicates
+    /// whether this bibliography should start numbering its citations after
+    /// the ones from previous bibliographies (`shared`) or start at 1 (`standalone`).
+    #[default("shared".into())]
```
laurmaedje, line 136: I'd have imagined `kind` to default to `auto` as well. And `standalone` wouldn't mean start at 1, but rather "count together with all bibliographies that also have `kind: "standalone"`. Basically each key is one counting track and `auto` is just the default track.

It still doesn't feel 100% great though as `auto` implies some extra smartness compared to a string, but there wouldn't be any.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,108 @@ pub struct BibliographyElem {
...
+        Ok(query
+            .into_iter()
+            .map(|elem| elem.to_packed::<Self>().unwrap().clone())
+            .collect())
+    }
+
+    #[comemo::memoize]
+    pub fn assign_citations(
+        introspector: Tracked<Introspector>,
+    ) -> FxHashMap<Span, Span> {
```
laurmaedje: One thing that is important across the PR: `Span`s are used to point back to a source code location (e.g. for diagnostics) rather than being used as unique identifiers. Instead, this is precisely the purpose of `Location`s.

Multiple citations or bibliographies can have the same span if they are created at the same place in the source code, for instance, in a function.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,108 @@ pub struct BibliographyElem {
...
-        Ok(elem.to_packed::<Self>().unwrap().clone())
+        // Find the bibliography for the remaining citations. Priority order:
+        // 1. First following auto bibliography containing the label
+        // 2. First preceding auto bibliography containing the label
+        for citation in citations {
+            if !citation_map.contains_key(&citation.span()) {
+                let citation_key = citation.to_packed::<CiteElem>().unwrap().key;
+                let citation_location = citation.location().unwrap();
+                if let Some(next_bib) = introspector
+                    .query(&Self::ELEM.select().after(citation_location.into(), false))
```
laurmaedje: We can rely on the introspector here to do these queries, but alternatively, we could also query for `selector.or(cite, bibliography)` at the top, to get both fused, in the correct order. I am not sure which is easier to work with, but the latter would be more efficient.

laurmaedje (commented):
I didn't yet have a chance to do a proper full review, but I have a few initial comments already. Feel free to act on these already, but of course you can also wait for the full review (will get to that next week).

crates/typst-library/src/model/bibliography.rs
thuiop: I actually did it with Location initially but I think there was some issue. I will try again.

thuiop: Well, it seems to work with Location without issue now.

crates/typst-library/src/model/bibliography.rs
thuiop: Ah, thanks, I wanted to do something like that but could not figure out how to assert whether a citation was before or after a given bibliography as I queried them separately.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,99 @@ pub struct BibliographyElem {
...
+                }
+            }
+        }
+
+        // Find the bibliography for the remaining citations. Priority order:
+        // 1. First following auto bibliography containing the label
+        // 2. First preceding auto bibliography containing the label
+        let citations_and_bib = introspector
+            .query(&Selector::Or(eco_vec![CiteElem::ELEM.select(), Self::ELEM.select()]));
+        for (i, citation) in citations_and_bib.iter().enumerate() {
```
laurmaedje, line 281: This is a bit problematic as the linear scans for the prev/following bibliography result in overall quadratic complexity. It would be good if we could get this down to $O(n \log n)$. This will probably need one extra data structure containing the bibliographies (e.g. a vector with bibliographies and indices which we can binary search.) Maybe there's a better approach.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,99 @@ pub struct BibliographyElem {
...
+            if citation.elem() == CiteElem::ELEM && !citation_map.contains_key(&citation.location().unwrap()) {
+                let citation_key = citation.to_packed::<CiteElem>().unwrap().key;
```
laurmaedje: The elem check followed by to_packed + unwrap can in multiple places be replaced by `if let Some(elem) = content.to_packed::<..>() { ...`

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,99 @@ pub struct BibliographyElem {
...
+                        .entry(citation.location().unwrap())
+                        .or_insert(bibliography_location);
+                }
+            }
+        }
+
+        // Find the bibliography for the remaining citations. Priority order:
+        // 1. First following auto bibliography containing the label
+        // 2. First preceding auto bibliography containing the label
+        let citations_and_bib = introspector
```
laurmaedje: these queries will result in a conflict with my changes on main (introspector warnings). You'll have to move these outside of the memoized function, but resolving it shouldn't be too bad I think.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,99 @@ pub struct BibliographyElem {
...
+                        }
+                        else {
+                            None
+                        }
+                    })
+                    .filter(|bibliography| {
+                        bibliography.target.get_ref(StyleChain::default()).is_auto()
+                            && bibliography.sources.derived.has(citation_key)
+                    })
+                    .next()
```
laurmaedje: The code duplication with the previous branch would ideally be avoided.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -593,6 +697,10 @@ impl Works {
             })
             .at(elem.span())
     }
+
+    pub fn hanging_indent(&self, elem: &Packed<BibliographyElem>) -> bool {
```
laurmaedje, line 765: Please add brief doc comments to public functions.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -580,7 +681,10 @@ impl Works {
         elem: &Packed<BibliographyElem>,
         styles: StyleChain,
     ) -> SourceResult<&'a [(Option<Content>, Content, Location)]> {
-        self.references
+        self.works
+            .get(&elem.location().unwrap())
+            .unwrap()
```
laurmaedje: these unwraps are not ideal. I think they maybe (?) could panic if a bibliography appears in the very last introspection iteration (see the added tests from https://github.com/typst/typst/pull/7364)

crates/typst-library/src/model/bibliography.rs
```diff
@@ -637,194 +746,255 @@ struct CiteInfo {
...
+                let style = match first.style.get_ref(StyleChain::default()) {
+                    Smart::Auto => bibliography_style.get(),
+                    Smart::Custom(style) => style.derived.get(),
+                };
 
-        let locale = locale(
-            self.bibliography.lang.unwrap_or(Lang::ENGLISH),
-            self.bibliography.region.flatten(),
-        );
+                self.infos.entry(bibliography_location).or_insert(vec![]).push(
```
laurmaedje:
```suggestion
                self.infos.entry(bibliography_location).or_default().push(
```

crates/typst-realize/src/lib.rs
```diff
@@ -1087,18 +1088,39 @@ fn finish_par(mut grouped: Grouped) -> SourceResult<()> {
...
+        }
+    }
+
+    for key in key_order {
+        let sub_children = map.get(&key).unwrap();
+        let span = Span::find(sub_children.iter().map(|c| c.span()));
+        let elem = CiteGroup::new(sub_children.clone()).pack().spanned(span);
+        visit(s, s.store(elem), trunk)?;
+    }
+    return Ok(());
```
laurmaedje: I'm not sure it makes sense to keep together two citations which are interrupted by one with a different bibliography. Segmenting based on neighbouring citations with different bibliographies seems sufficient, simpler, and more predictable to me.

crates/typst-library/src/model/bibliography.rs
```diff
@@ -150,19 +165,99 @@ pub struct BibliographyElem {
...
+                        .or_insert(next_bib.location().unwrap());
+                }
+                else if let Some(prev_bib) = citations_and_bib[..i]
+                    .iter()
+                    .rev()
+                    .filter_map(|content| {
+                        if content.elem() == Self::ELEM {
+                            Some(content.to_packed::<Self>().unwrap().clone())
+                        }
+                        else {
```
laurmaedje: The formatting is a bit off here. Could you format the files with rustfmt?

laurmaedje (commented):
Thanks for your work, this is going into a good direction! :)

I left some comments. Besides that, it would be good if you could add some tests for the new functionality. Currently, CI is failing because of the missing hayagriva change. I'd like to wait with merging this until this PR is merge-ready. Until then, you could point from `Cargo.toml` to your hayagriva fork with the relevant change.

The PR will also need a merge or rebase because it has some conflicts with the introspection warnings changes.

thuiop:
I have been taking a crack at reconciling this with the changes on main, but it is actually not trivial. Since you cannot do the introspection work in the memoized functions anymore, you basically need to do it every time the `Works::generate` function is called, which is not ideal. Also, I relied on hashmaps for the mapping between citations/groups and bibliographies, which now have to be passed to the memoized function, which does not play well with them as `FxHashMap` is not hashable (which could be implemented but does add a lot of boilerplate). Am I missing something here?

laurmaedje:
My thought was to do just the queries outside and pass them down and everything else remains as before. This is not _hugely_ different[^1] from doing the queries internally as even if it's memoized these calls are tracked and will be validated by comemo. The introspector also has an internal cache, so it should be fairly cheap.

I didn't realize though that there is also a variable number of queries happening in `assign_citations`. Given this, it might make sense to handle this more elaborately. It's absolutely possible to have memoized functions that perform internal queries, it's just a bit more ceremony as the `Engine` needs to be destructured and rebuilt at the memoization boundary (because comemo only supports top-level `Tracked` types as direct arguments to memoized functions, not types containing `Tracked` types). You can find examples of this basically anywhere `Protected::from_raw(introspector)` is called. Feel free to give that a shot. If you run into problem with this, let me know and I'll give it a shot myself.

[^1]: It's a bit less efficient because it does not profit from comemo's internal constraint validation acceleration cache. Honestly, I'm not sure how much this matters here in practice; it might very well matter.

thuiop:
Ok, then I guess doing this trick with Engine is the way to go. I will give it a try.

thuiop:
Ok, I think I see how it works, but in those cases you actually lose the span, right? And just query the introspector directly as before.

laurmaedje:
> Ok, I think I see how it works, but in those cases you actually lose the span, right? And just query the introspector directly as before.

Not sure what you mean. Which span? The introspector may only be directly queried in `impl Introspect for`. The `CiteGroupIntrospection` is special and used instead of `QueryIntrospection` only for diagnostic purposes. Such an introspection _may_ be necessary for diagnostic, but otherwise a `QueryIntrospection` would be the right tool.

thuiop:
> Not sure what you mean. Which span? The introspector may only be directly queried in `impl Introspect for`. The `CiteGroupIntrospection` is special and used instead of `QueryIntrospection` only for diagnostic purposes. Such an introspection _may_ be necessary for diagnostic, but otherwise a `QueryIntrospection` would be the right tool.

Well, QueryIntrospection takes in a span for diagnostic purposes, right? E.g. here https://github.com/typst/typst/blob/7b072b953ea444d5ce80fe4c439f530bdc56e522/crates/typst-library/src/model/bibliography.rs#L561C5-L561C8 it is the span of where the `generate` function is called. But if you memoize, you cannot pass around the span, so what span should be used? In places where you do the trick with the engine, I do not see where the span comes from ultimately.

laurmaedje:
Ah, yes, I see now. That's indeed annoying and a unique problem. I'll need a bit to think about how to deal with this.

thuiop:
Ok, for now I used a `Span::detached` to create a dummy span to use, so that I can focus on your other remarks while this gets figured out.

crates/typst-library/src/model/bibliography.rs
thuiop: I tried to factorize the code a bit, not completely happy about it but there is no more duplication.

crates/typst-library/src/model/bibliography.rs
thuiop: For now I changed it so that it scans only the bibliography elements

crates/typst-library/src/model/bibliography.rs
thuiop: I removed the unwrap there and returned the error instead

laurmaedje:
I looked into the span situation.

In `BibliographyElem::assign_citations`:

- The `QueryIntrospection` that searches for `CiteElem | BibliographyElem` could be moved outside of the memoized part and use the span of the source citation/bibliography. 

- The related `QueryIntrospection` for `Self::ELEM` could be fully removed in favor of filtering the existing `citations_and_bib` result to save one introspection.

- The `QueryIntrospection` for `bibliography_selector` can use `bibliography.span()`

In `Generator::new`:

- The `BibliographyElem::find` introspection and `CiteGroupIntrospection` introspection can be moved into `Works::generate` and passed down.

This should at least work and unblock the PR. The overhead might be a bit higher than before, but the core logic could stay memoized. I might look into perf impact separately and would potentially adjust queries to return prehashed results to reduce hashing overhead at the memoization boundary, but that's not in scope for the PR.

thuiop:
Hi, sorry I have been a bit out of touch these past few weeks. I am ok with making these changes; I think there are other instances that should be moved upwards outside of the memoization? E.g. calling `assign_citations` should be done in `Works::generate` instead of `Works::generate_impl`, and so on.

thuiop:
I did what seemed best; not entirely sure about the correct span to give when calling `assign_citations` during grouping though.

breezykermo:
It's great to see that this is being worked on again! The authors may already be aware of this: but since [bundling was merged](https://github.com/typst/typst/pull/7964) there is a new urgency for this, because AFAIK this extends to Typst documents that are being exported to the bundle format. This means that sites like [this one](https://lolm.ohrg.org/), which has per-page bibliographies, cannot be created with the Typst bundle format currently. (This site was produced with [Rheo](https://github.com/freecomputinglab/rheo).)

laurmaedje:
@breezykermo This is mostly blocked on me right now, but it's a priority for 0.15 and it's still in my plans to get this in.

laurmaedje:
Hi! I've opened #8307 to complete this. The design is still mostly the same, but I wanted to do some [refactoring first](https://github.com/typst/typst/pull/8283) and also change up how the citation grouping and assignment interacts, so in the end it made more sense to open a new PR. Thank you for your work nonetheless! It was quite useful to build upon and the design idea with having a selector that `target`s citations was very nice.

thuiop:
> Hi! I've opened #8307 to complete this. The design is still mostly the same, but I wanted to do some [refactoring first](https://github.com/typst/typst/pull/8283) and also change up how the citation grouping and assignment interacts, so in the end it made more sense to open a new PR. Thank you for your work nonetheless! It was quite useful to build upon and the design idea with having a selector that `target`s citations was very nice.

Hi Laurenz! Makes sense to me; glad that a similar design was retained. Looking forward to see the feature implemented in the next version!

PR #7278, opened 2025-10-31
Test runner changes

saecki:
Continuing from #7239. This is the first step in adding support for PDF and SVG tests. These tests will only store a hash of the reference output to reduce the storage needed for tests. I *think* the PDFs generated should be fully deterministic, but maybe @LaurenzV knows more?

The plan is to add a test tool later on that can build the test suite at another git revision on test failure to compare the previous and current reference output. Then static HTML is generated that allows conveniently inspecting the changes/diffs. The tool will be usable both in PRs and locally.

This PR changes a few things in the test suite:
- Move the `render` references to a sub directory
- Replace `render` with the `paged` attribute which always runs `render`, `pdf`, and `svg` output.
- Replace `nopdfua` with `pdfstandard` which also applies to the `pdf` output
- Require at least one of `paged`, `html`, or `pdftags` on a test, there is no default target anymore.
- The `pdf` output always runs both the standard and the PDF/UA-1 export
- `pdftags` can either be used in conjunction with `paged` or alone. If used alone only a tag tree reference is stored.
- `pdftags` are now generated from a finished PDF document using `hayro`
  - This allows reading a few extra attributes:
    - `/Lang`, `Alt`, `ActualText`, and `E` of marked content sequences
    - `/Contents` and `/URI` of link annotations.
  - There are also some other changes to the tag tree
    - The structure element types and attribute names now match the PDF spec exactly
    - `BBox`es are formatted a little more compact
- Add a `--stages` flag to only run specific which test targets/outputs
  - This is useful to update only one test output
- Fix running only parser tests with `--no-default-features`

Currently I've disabled `pdf` and `svg` references since there currently is no easy way to get the previous reference output, if there is a hash mismatch.

I had to update `krilla` because empty table cell headers were previously omitted from PDFs, and it now requires rust `1.89`.

### DIFF
--- a/tools/test-helper/src/extension.ts
+++ b/tools/test-helper/src/extension.ts
@@ -124,7 +124,7 @@ class TestHelper {
     const lenses = [];
     for (let nr = 0; nr < document.lineCount; nr++) {
       const line = document.lineAt(nr);
-      const re = /^--- ([\d\w-]+)(( [\d\w-]+)*) ---$/;
+      const re = /^--- ([\d\w-]+)(( [\d\w-\(\)]+)*) ---$/;
       const m = line.text.match(re);
       if (!m) {
         continue;
@@ -401,12 +401,7 @@ type Bucket = "store" | "ref";
 type Format = "html" | "render";
 
 function getUri(name: string, bucket: Bucket, format: Format) {
-  let path;
-  if (bucket === "ref" && format === "render") {
-    path = `tests/ref/${name}.png`;
-  } else {
-    path = `tests/${bucket}/${format}/${name}.${EXTENSION[format]}`;
-  }
+  let path = `tests/${bucket}/${format}/${name}.${EXTENSION[format]}`;
   return vscode.Uri.joinPath(getWorkspaceRoot(), path);
 }
 
@@ -420,8 +415,8 @@ async function getWebviewContent(
     stderr: string;
   }
 ): Promise<string> {
+  const showRender = attrs.includes("paged");
   const showHtml = attrs.includes("html");
-  const showRender = !showHtml || attrs.includes("render");
 
   const stdout = output?.stdout
     ? `<h2>Standard output</h2><pre class="output">${escape(


tests/src/output.rs
```diff
@@ -0,0 +1,427 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:032x}", self.0)
+    }
+}
+
+impl FromStr for HashedRef {
+    type Err = EcoString;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if !s.len() == 32 {
```
laurmaedje:
```suggestion
        if s.len() != 32 {
```

laurmaedje: good candidate for a clippy lint :D

tests/src/output.rs
```diff
@@ -0,0 +1,427 @@
...
+impl FromStr for HashedRefs {
+    type Err = EcoString;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let refs = s
+            .lines()
+            .map(|line| {
+                let mut parts = line.split_whitespace();
+                let Some(hash) = parts.next() else {
+                    bail!("found empty line");
```
laurmaedje: without the semicolons, this formats a little more nicely

tests/src/run.rs
```diff
@@ -125,33 +138,131 @@ impl<'a> Runner<'a> {
...
+
+    /// Run test specific to document format.
+    fn run_test<T: OutputType>(
+        &mut self,
+        doc: &Option<T::Doc>,
+    ) -> Option<(T::Doc, T::Live)> {
+        let live_path = T::OUTPUT.live_path(&self.test.name);
+
+        let output =
+            doc.clone()
```
laurmaedje: if we clone anyway, might as well take an owned doc. will also format more nicely.

tests/src/run.rs
```diff
@@ -230,9 +327,100 @@ impl<'a> Runner<'a> {
...
+            let hashed_refs = hashes.get_mut(source_path).unwrap();
+            let ref_path = T::OUTPUT.hashed_ref_path(source_path.as_rootless_path());
+            if skippable {
+                hashed_refs.remove(&self.test.name);
+                log!(
+                    into: self.result.infos,
+                    "removed reference hash ({})", ref_path.display()
+                );
+            } else {
+                eprintln!("update");
```
laurmaedje: there are a few leftover eprintlns here

tests/src/collect.rs
```diff
@@ -63,60 +65,208 @@ impl Display for FilePos {
...
 }
 
-#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
-pub struct Attrs {
-    pub large: bool,
-    pub pdf_ua: bool,
-    pub targets: Targets,
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[allow(non_camel_case_types)]
+pub enum PdfStandard {
```
laurmaedje: can't we use the existing types from `typst_pdf`?

tests/src/collect.rs
```diff
@@ -216,46 +366,116 @@ impl Collector {
         }
     }
 
+    #[cfg(not(feature = "default"))]
+    fn walk_references(&mut self) {}
+
     /// Walks through all reference output and ensures that a test exists for
     /// each one.
+    #[cfg(feature = "default")]
     fn walk_references(&mut self) {
```
laurmaedje, line 398: the indent is getting pretty massive here. perhaps this could be split up a bit?

tests/src/collect.rs
```diff
@@ -357,34 +579,83 @@ impl<'a> Parser<'a> {
...
+            if !self.s.at(' ') {
+                self.error("expected a space after an attribute");
+            }
+
+            match attr_name {
+                "paged" => {
+                    // The paged target will test render, pdf, and svg by
+                    // default.
+                    self.set_attr(attr_name, &mut stages, TestStages::PAGED);
+                    stages |= TestStages::RENDER | TestStages::PDF | TestStages::SVG;
```
laurmaedje: this is a bit duplicate with the code in `CliArguments::stages`. And also a bit with `TestStages::paged_stages`, but that one is a bit different (which is a bit confusing to me).

overall, it might be cleaner if the parsing code is not concerned with how different targets imply each other

tests/src/run.rs
```diff
@@ -1,20 +1,28 @@
...
+    FileSize, NoteKind, Test, TestOutput, TestStage, TestStages, TestTarget,
+};
 use crate::logger::TestResult;
+use crate::output::{FileOutputType, HashOutputType, HashedRefs, OutputType};
 use crate::world::{TestWorld, system_path};
+use crate::{ARGS, custom, output};
+
+type OutputHashes = FxHashMap<&'static VirtualPath, HashedRefs>;
+
+static HASHES: LazyLock<[RwLock<OutputHashes>; 2]> =
```
laurmaedje: having this as a global does not feel _great_ to me. I don't have any actionable suggestions, but was wondering whether you considered and discarded any other options?

tests/src/run.rs
```diff
@@ -23,6 +31,24 @@ pub fn run(test: &Test) -> TestResult {
...
+        hashed_refs.sort();
+        if !hashed_refs.changed {
+            continue;
+        }
+
+        let ref_path = T::OUTPUT.hashed_ref_path(source_path.as_rootless_path());
+        if hashed_refs.is_empty() {
+            std::fs::remove_file(ref_path).ok();
+        } else {
+            std::fs::write(ref_path, hashed_refs.to_string()).unwrap();
```
laurmaedje, line 41: if hash refs are enabled and pdf/svg dirs do not exist yet, it crashes here

tests/src/pdftags.rs
```diff
@@ -0,0 +1,487 @@
...
+            pages,
+            buf: String::new(),
+            indent: 0,
+            pending_indent: false,
+            maybe_space: None,
+        }
+    }
+}
+
+struct PageContent<'a> {
```
laurmaedje: the ordering of structs/impls is a bit weird here. I'd probably put the two struct definitions right below `struct Formatter` as they are just small helpers. but between the inherent impl and write impl is a bit strange.

tests/src/pdftags.rs
laurmaedje, line 1: the file uses panics quite heavily for the expected structure. I wonder whether results would be more appropriate here? if there's a bug in Typst or hayro, that would be a bit friendlier and not crash the whole test suite.

tests/suite/layout/grid/headers.typ
laurmaedje, line 1: I feel like writing `pdftags pdfstandard(ua-1)` might get old pretty quickly. Are we sure this is the right default?

tests/src/args.rs
```diff
@@ -70,6 +78,33 @@ pub struct CliArguments {
     // how you would expect and I'm too lazy to try to fix it.
 }
 
+impl CliArguments {
+    pub fn should_run(&self, stage: impl Into<TestStages>) -> bool {
+        static CACHED: AtomicU8 = AtomicU8::new(0xFF);
```
laurmaedje, line 84: can't help but micro-optimize, eh? 😂

tests/ref/pdftags/bibliography-basic.yml
```diff
@@ -8,29 +8,33 @@
     - Tag: Link
       /K:
         - Content: page=0 mcid=2
-        - Annotation: page=0 index=0
+        - Annotation: page=0 subtype=Link
+          /Contents: "[1]"
```
laurmaedje, line 12: yeah okay, now I see the point of the parsing. this is indeed nice to have.

tests/src/collect.rs
```diff
@@ -117,6 +117,26 @@ impl TestStages {
...
+            }
+            bitflags::bitflags_match!(flag, {
+                TestStages::PAGED => Display::fmt(&TestTarget::Paged, f).unwrap(),
+                TestStages::RENDER => Display::fmt(&TestOutput::Render, f).unwrap(),
+                TestStages::PDF => Display::fmt(&TestOutput::Pdf, f).unwrap(),
+                TestStages::PDFTAGS => Display::fmt(&TestOutput::Pdftags, f).unwrap(),
+                TestStages::SVG => Display::fmt(&TestOutput::Svg, f).unwrap(),
+                TestStages::HTML => Display::fmt(&TestTarget::Html, f).unwrap(),
+                _ => unreachable!(),
+            })
```
laurmaedje: could skip the unwraps like this:
```rs
bitflags::bitflags_match!(flag, {
    TestStages::PAGED => Display::fmt(&TestTarget::Paged, f),
    TestStages::RENDER => Display::fmt(&TestOutput::Render, f),
    TestStages::PDF => Display::fmt(&TestOutput::Pdf, f),
    TestStages::PDFTAGS => Display::fmt(&TestOutput::Pdftags, f),
    TestStages::SVG => Display::fmt(&TestOutput::Svg, f),
    TestStages::HTML => Display::fmt(&TestTarget::Html, f),
    _ => unreachable!(),
})?
```

tests/src/output.rs
```diff
@@ -205,9 +205,16 @@ impl OutputType for Pdf {
         if doc.info.title.is_none() {
             doc.info.title = Some("<test>".into());
         }
-        let standards = PdfStandards::new(test.attrs.pdf_standard.as_slice()).unwrap();
-        let options = PdfOptions { standards, ..Default::default() };
-        typst_pdf::pdf(doc, &options)
+
+        // Always run the default PDF export and PDF/UA-1 export, if another
```
laurmaedje: The comment could also motivate why we're doing this (to detect crashes, right?)

tests/src/output.rs
saecki: Oh wow :sweat_smile:

tests/src/collect.rs
saecki: That would break compilation with `--no-default-features`, or it would require a lot of `#[cfg(...)]` because the `pdf_standard` is stored on the `Attrs` struct.

tests/src/run.rs
saecki: Yeah I also encountered that, but the same goes for `render`, `pdftags` and `html`. I don't think there is any benefit in adding an if statement for a one time action. Once we decide to add the first test, the directory will be tracked by git...

tests/suite/layout/grid/headers.typ
saecki: I somewhat agree, but I think the `pdfstandard` for `paged` tests shouldn't be PDF/UA-1 and I didn't want to have different standards for `pdf` and `pdftags` references. I also didn't want _automagically_ to change the standard of a `paged` test when `pdftags` was added, so I arrived here. What do you think :thinking:

tests/src/args.rs
saecki: MUST ... AVOID ... LOCK

tests/src/run.rs
saecki: I wanted to get rid of the clone anyway, it was only required because a title was set if it was missing from the document info. I've just moved the title assignment right after where the paged document is compiled and now we can just pass along the references.

tests/src/run.rs
saecki: We could just have this in a local in `fn test()` and pass a reference to each `TestRunner`

tests/src/collect.rs
saecki: I've refactored this now a little bit, let me know if this seems fine.

tests/src/pdftags.rs
saecki: Will do tomorrow.

saecki: Done. There are still a few unwraps, but I think those are fine.

tests/src/run.rs
saecki: Done

saecki:
> One general remark is that it would be nice if you could add brief doc comments to at least all structs and ideally also methods.

Done

> Regarding merge: I'd like to keep the commits in principle. However, I think in the way it's currently set up, not all intermediate commits pass the tests. (Most importantly because of the last "Disable hashed references for now" commit.) Do you think it would be feasible to rebase in a way that lets each commit pass in isolation? Otherwise, I'm also fine with squashing. Let me know what you prefer.

The commits should all pass CI now
 
> To get the fuzzer tests to pass, I guess we also need to update the nightly toolchain used for fuzzing? I'd need to find a good nightly and sync the nightly versions across here and the web app. Perhaps, it would be worthwhile to decouple the whole krilla update from this PR. So basically: Update nightly toolchain for tests -> Update krilla -> Test runner changes. I can take care of the first PR.

I've opened a separate PR to `Update krilla`

tests/src/collect.rs
laurmaedje: The whole features stuff is a bit annoying tbh. I'm not sure it's quite worth it to keep.

saecki: Ok, apparently I broke it again with my latest change. Let's just remove it :+1:

tests/suite/layout/grid/headers.typ
laurmaedje: hmmmmmmm okay, einen tod muss man sterben

saecki:
The test suite might be a little bit slower now that svg and pdf are also tested and due to #7365.
Running `cargo testit --stages=render,html,pdftags` should be just as fast as before.

laurmaedje:
This LGTM apart from one thing: The VS Code test helper extension has some issues now. This line needs adjustment with the parameters

https://github.com/typst/typst/blob/bc3d0e580d436f9c0628cdb07fa1b6f07aa9e655/tools/test-helper/src/extension.ts#L127

And the preview also does not work correctly for tests that are have both paged and html. The extension does some light parsing of the attributes which needs to be adjusted.

saecki:
I think that should fix it

Issue #7210, opened 2025-10-23
Add a `filter` method to `dictionary`


Mc-Zen:
### Description

Usage:
```typ
#{
  let dict = (:)
  dict.filter((key, val) => ..)
}
```
or
```typ
#{
  let dict = (:)
  dict.filter(((key, val)) => ..)
}
```
up to discussion. 

### Use Case

I think this could be useful in many scenarios, for example to filter out key-value pairs with a certain value etc. Currently, this can be achieved through `dict.pairs().filter().to-dict()` which is not bad but can be improved upon.

PR #7284, opened 2025-11-01
Implement `filter` and `map` for `dictionary` and `arguments`

MDLC01:
Closes https://github.com/typst/typst/issues/6006.
Closes https://github.com/typst/typst/issues/7210.

I decided to only pass values to the functions, for consistency with `array`, which also only passes values. I also implemented the methods for `arguments`, as suggested in https://github.com/typst/typst/issues/6006#issuecomment-3383005372.

Applying `arguments.filter` or `arguments.map` erases a lot of span information, but I don't think there is much that can be done about it.

### DIFF
--- a/crates/typst-library/src/foundations/args.rs
+++ b/crates/typst-library/src/foundations/args.rs
@@ -275,6 +275,7 @@ cast! {
 }
 
 impl Args {
+    /// Tests whether there is no positional nor named argument.
     pub fn is_empty(&self) -> bool {
         self.items.is_empty()
     }
@@ -418,7 +419,7 @@ impl Args {
             .map(|arg| {
                 let mapped_value = mapper.call(engine, context, [arg.value.v])?;
                 Ok(Arg {
-                    span: Span::detached(),
+                    span: arg.span,
                     name: arg.name,
                     value: Spanned::detached(mapped_value),
                 })
--- a/crates/typst-library/src/foundations/dict.rs
+++ b/crates/typst-library/src/foundations/dict.rs
@@ -272,7 +272,7 @@ impl Dict {
     /// Produces a new dictionary with only the pairs from the original one for
     /// which the given function returns `{true}`.
     ///
-    /// ```example
+    /// ```example:"Basic usage"
     /// #{
     ///   (a: 0, b: 1, c: 2)
     ///     .filter(v => v > 0)


Mc-Zen:
This potentially also closes #7210. Here I proposed a filter method with pairs as arguments but I think the idea came from an application where only the value mattered. Still it should be worth to think again, whether passing values it the best for _filtering_ dictionaries.

crates/typst-library/src/foundations/args.rs
```diff
@@ -354,6 +358,64 @@ impl Args {
...
+        let mut run_test = |v: &Value| {
+            test.call(engine, context, [v.clone()])?
+                .cast::<bool>()
+                .at(test.span())
+        };
+        self.into_iter()
+            .filter_map(|arg| {
+                run_test(&arg.value.v).map(|b| b.then_some(arg)).transpose()
+            })
+            .collect()
```
MDLC01, line 398: Should we apply the span of the call to `filter` to the returned `Args` (as its `span` field)?

Same question below for `arguments.map`.

crates/typst-library/src/foundations/args.rs
```diff
@@ -354,6 +358,64 @@ impl Args {
...
+        engine: &mut Engine,
+        context: Tracked<Context>,
+        /// The function to apply to each value.
+        mapper: Func,
+    ) -> SourceResult<Args> {
+        self.into_iter()
+            .map(|arg| {
+                let mapped_value = mapper.call(engine, context, [arg.value.v])?;
+                Ok(Arg {
+                    span: Span::detached(),
```
MDLC01: Should we keep the initial span here? Or provide the span of the `mapper` argument? Or keep it detached?

MDLC01:
That's right. I've linked it in the PR.

Regarding whether keys should be passed as well, I haven't given many details as to why I took this decision in the OP, so I'll do it below.

1. **Consistency with arrays.** Arrays don't pass the element indices to the functions provided to `array.filter` and `array.map`. Not passing the keys for `dictionary.filter` and `dictionary.map` allows us "unify" the functions that exist on both `array` and `dictionary` in `arguments`.

2. **Consistency with existing functions in other languages.** This kind of function comes from functional languages like Haskell and OCaml. In both those languages, the "default" (i.e., the behavior of the function with the most simple name) considers values only.[^1]

4. **Mapping the keys of a dictionary is ambiguous.** For `dictionary.map` and `arguments.map`, if the mapper returns a key-value pair, the behavior is ambiguous in case multiple pairs have the same key.

5. **Anything more complex can be easily reproduced.** Using `dictionary.pairs`, one can easily reproduce any more complex mapping or filtering function on dictionaries. Note that `arguments.pairs` does not exist, but that can be added easily if wanted.

[^1]: See [#6006 (comment)](https://github.com/typst/typst/issues/6006#issuecomment-2937169248).

Mc-Zen:
The reasoning for `map` makes total sense. For consistency, it thus follows that `filter` should work the same (although the third argument does not apply here).

laurmaedje:
I agree that the arguments presented in https://github.com/typst/typst/pull/7284#issuecomment-3518568075 for `dict.map` only receiving the value are quite strong. One thing I'm still struggling to see a use case for is `map` and `filter` on `arguments`. Do you have any examples in mind or is this purely for consistency?

MDLC01:
This is mostly for consistency. I also like the idea of making `arguments` a datatype that can be used for more than just argument sinks, but I can understand that this may be out of scope.

crates/typst-library/src/foundations/dict.rs
```diff
@@ -266,6 +268,55 @@ impl Dict {
             .map(|(k, v)| Value::Array(array![k.clone(), v.clone()]))
             .collect()
     }
+
+    /// Produces a new dictionary with only the pairs from the original one for
+    /// which the value passes test.
```
laurmaedje: "passes the test" or passes `test` or maybe just "for which the given function returns true" for consistency with the array docs

crates/typst-library/src/foundations/dict.rs
```diff
@@ -266,6 +268,55 @@ impl Dict {
...
+    ///
```
laurmaedje, line 274: Maybe we want to add a second example here for how to filter pairs with `.pairs()` and `.to-dict()`? and same for `map`.

laurmaedje:
That makes sense. I can see it, considering that `arguments` also already has `at`. I guess `.len()` would make sense, too.

Regarding spans: The way spans currently work for arguments handled as values is already a bit crazy, e.g. in:
```typ
#let args = arguments(-1, a: 0, b: 1, 2)
#calc.max(..args)
```
where it annotes `a: 0` with unexpected argument and a trace on `calc.max`. But I guess it makes sense and with the trace it's okay.

If the span is detached, that kinda sucks:
```
error: unexpected argument: a

help: error occurred in this call of function `max`
  ┌─ hi.typ:6:1
  │
6 │ #calc.max(..args)
  │  ^^^^^^^^^^^^^^^^
```

The top-level error message should at least have some span. I guess we could do a fallback there to some other span where this error is produced. But it would be best to just have some span. To me, the mapper span seems less useful than the existing arg span since it will be the same for all args.

crates/typst-library/src/foundations/args.rs
MDLC01: Answered by https://github.com/typst/typst/pull/7284#issuecomment-3548518223

MDLC01:
Unless I missed something, this should be almost ready to be merged now. Sorry for taking so long just for a merge commit.

One last thing I'm not sure about is your opinion on the question raised in https://github.com/typst/typst/pull/7284#discussion_r2515589054?

crates/typst-library/src/foundations/args.rs
```diff
@@ -271,6 +275,10 @@ cast! {
 }
 
 impl Args {
+    pub fn is_empty(&self) -> bool {
```
laurmaedje, line 279: Could you add a doc comment here?

crates/typst-library/src/foundations/dict.rs
```diff
@@ -266,6 +268,64 @@ impl Dict {
             .map(|(k, v)| Value::Array(array![k.clone(), v.clone()]))
             .collect()
     }
+
+    /// Produces a new dictionary with only the pairs from the original one for
+    /// which the given function returns `{true}`.
+    ///
+    /// ```example
```
laurmaedje: Let's add a title here as well. The current policy is to always have titles for all examples as soon as there is more than one example.

laurmaedje:
Thanks!

PR #7289, opened 2025-11-03
Update link on selector page to point to element functions

bountonw:
Currently the selector docs page of the website mentions element functions, but points to the functions page and not the element functions sub-heading. However, other pages that refer to element functions do point to the element functions section of the function page. This PR updates the link on the selector docs page to point to the element functions sub-section of the functions page.


bountonw:
I may have mis-understood the syntax, But I hope this is the start of a good thing.

laurmaedje:
Thanks!

PR #7290, opened 2025-11-03
Let examples distinguish ±dx and ±dy in the docs for `place` and `move`

YDX-2147483647:
Both `place` and `move` accept `dx` and `dy`, but there are many convertions for x y coordinates.
This PR adds visual hints that can imply the convention that typst takes.

## [`place.dx`](https://typst.app/docs/reference/layout/place/#parameters-dx)

- Make the values of `dx` and `dy` different so that they can be distinguished.
- Make the color and font size vary along the line, to distinguish between ±.
- Change `A` to `T`, because `A` is too easy to collide with other `A`'s.

Before:
图片

After:
图片

## [`move`](https://typst.app/docs/reference/layout/move/#example)

- Make the values of `dx` and `dy` different so that they can be distinguished.
- Make the styles of the two rectangles different to clarify the direction of movement.
- ~~Center it.~~

Before:
图片

After:
图片

### DIFF
--- a/crates/typst-library/src/layout/transform.rs
+++ b/crates/typst-library/src/layout/transform.rs
@@ -9,7 +9,6 @@ use crate::layout::{Abs, Alignment, Angle, HAlignment, Length, Ratio, Rel, VAlig
 ///
 /// # Example
 /// ```example
-/// >>> #set page(width: auto)
 /// #rect(inset: 0pt, fill: gray, move(
 ///   dx: 4pt, dy: 6pt,
 ///   rect(


crates/typst-library/src/layout/transform.rs
```diff
@@ -9,8 +9,9 @@ use crate::layout::{Abs, Alignment, Angle, HAlignment, Length, Ratio, Rel, VAlig
 ///
 /// # Example
 /// ```example
-/// #rect(inset: 0pt, move(
-///   dx: 6pt, dy: 6pt,
+/// >>> #set page(width: auto)
```
laurmaedje: I think making the page width auto might make the image scale up and look blurry in the docs.

crates/typst-library/src/layout/place.rs
```diff
@@ -145,9 +145,14 @@ pub struct PlaceElem {
     ///
     /// ```example
     /// #set page(height: 100pt)
-    /// #for i in range(16) {
-    ///   let amount = i * 4pt
-    ///   place(center, dx: amount - 32pt, dy: amount)[A]
+    /// #for x in range(-8, 8) {
```
laurmaedje, line 148: This is nice. The colors are Typst-y .:D

crates/typst-library/src/layout/transform.rs
YDX-2147483647: Okay, then I'll delete it.

laurmaedje:
Thanks!

Issue #4702, opened 2024-08-07
Render actual text rather than paths in SVG (possibly with embedded fonts)


ongun-kanat:
### Description

While rendering text and fonts as paths in SVG images guarantee a consistent output, it makes them unselectable and unsearchable. The rendered path is also limited by the capability of the rendering engine and the font libraries Typst uses which are currently not as complete as the text rendering infrastructure that come with the OS and browsers.

An option in the typst CLI to opt for SVG `<text>` element output for the textual data would be very useful, especially when the file is to be viewed in the browser.

An optional improvement over this is embedding the actual fonts within the SVG file. This has a potential to bloat the SVG file, but could be useful for things like figures.

### Use Case

I'm using Typst to generate figures containing shapes and tables. I would prefer the text in them to be selectable when viewed in a browser. 

SVG files that use `<text>` elements are also smaller than the SVG files whose text is converted into paths.

Enivex:
Someone more knowledgeable can correct me if I'm wrong, but I do not believe it's possible to embed fonts in an SVG, at least not in a way that's usable outside of a browser. Therefore such SVGs would be non-portable.

(Just a wild thought: Would it be possible to put a transparent text layer on top, just for selectability purposes?)

ongun-kanat:
Yeah it is true that the embedding is done via CSS styling. There are even tools that analyze the local fonts and embed the TTF / WOFF files inside SVGs:

- https://github.com/BTBurke/svg-embed-font
- https://github.com/phauer/svg-buddy

Displaying those fonts correctly of course requires a CSS capable renderer. However, `<style>` tag is a part of the SVG standard (optional for 1.1, mandatory for 2.0): https://svgwg.org/svg2-draft/styling.html#StyleElement . The level of implementation varies quite a bit of course.

I am okay with the invisible text layer as well. As I said it is optional to embed fonts.

Just outputting the SVG with the text tag and hoping the user's system has the font is also an option but I can see that it can cause some unwanted results when the font is fed via CLI.

Enivex:
SVG 2.0 isn't finalized (and I'm not entirely convinced it ever will be)

LaurenzV:
The main problem is that exporting the text as strings in `<text>` elements would only work for the very simplest texts, but Typst supports a lot more (think subscripts, superscripts, specifying OpenType feature to use when shaping the text, and also math mode which (I think) has very complex positioning rules), and I don't think you can reliably represent those in SVG. 

[...]

Issue #344, opened 2023-03-26
Line numbering for Raw Text / Code


nightingale-m:
Currently, there is no way to set line numbers for [Raw Text / Code](https://typst.app/docs/reference/text/raw/). It would also be cool to be able to specify which number it starts from.

Perhaps specify this in the parameter `numbering`, which defaults to `none` but can be changed to integer?

victor-sm:
@mpizenberg shared a solution on [how to accomplish this](https://typst.app/project/ptjv9k7bXqL3zj71rOmzQD). A native solution might still be nice, though.

mpizenberg:
Yep, this works for me, but I agree something more native would be nice :)

classabbyamp:
> @mpizenberg shared a solution on [how to accomplish this](https://typst.app/project/ptjv9k7bXqL3zj71rOmzQD)

I cannot access this, could you share it directly or via a gist or similar?

mpizenberg:
Here you go: https://gist.github.com/mpizenberg/c6ed7bc3992ee5dfed55edce508080bb

mpizenberg:
*Goal*: being able to add line numbers, which are correct even in case of long lines that need wrapping.

*Strategy*: duplicate the code block, once for getting the line numbers correct, and the other for syntax highlighting. The idea is to split lines and prefix each line with its line number such that line wrapping should be respected.

```
#show raw.where(block: true): it => { set par(justify: false); grid(
  columns: (100%, 100%),
  column-gutter: -100%,
  block(width: 100%, inset: 1em, for i, line in it.text.split("\n") {
    box(width: 0pt, align(right, str(i + 1) + h(2em)))
    hide(line)
    linebreak()
  }),
  block(radius: 1em, fill: luma(246), width: 100%, inset: 1em, it),
)}
```

willakat:
Hi,
I was missing some of the features from [lstlistings](https://ctan.gutenberg-asso.fr/macros/latex/contrib/listings/listings.pdf) and put this together to start learning typst. Posting it here in case it can help someone or even maybe serve to help spark a "native" typst feature.

```
#let code(
    caption: none,             // content of caption bubble (string, none)
    bgcolor: rgb("#fefae0"),   // back ground color (color)
    strokecolor: 1pt + maroon, // frame color (color)
    hlcolor: auto,             // color to use for highlighted lines (auto, color)
    width: 100%,
    radius: 3pt,
    inset: 5pt,
[...]

PR #7295, opened 2025-11-03
Add support for unselectable text

### LINKED ISSUE #2249: Add option to make text non-selectable

andylizi:
Add a new parameter `set text(selectable: false)` that resolves #2249.

This particular design was previously proposed in #5789 and https://github.com/typst/typst/pull/5938#issuecomment-2694306451.

This is my first PR in Typst and I'm not very familiar with the codebase. Please let me know if there's anything I missed!

### Testing

Manually check the PDFs.

```shell
cargo testit --pdf --exact text-unselectable image-svg-text-unselectable
```

### Example Use Cases

- https://github.com/typst/typst/issues/344#issuecomment-2515223243
- https://github.com/typst/typst/issues/2249#issuecomment-2334476486

### Limitations

- SVG export currently does not support selectable text at all (#4702)
- Unselectable text in HTML requires the use of CSS, which is out of scope for the initial phase.
- Are there any implications regarding accessibility (e.g., the artifact tag?).


LaurenzV:
Seems like a good addition overall, but I'm wondering whether we should just disallow that option for PDF/UA-1? Since it's really not encouraged to do this normally.

andylizi:
> I'm wondering whether we should just disallow that option for PDF/UA-1? Since it's really not encouraged to do this normally.

I'm not very familiar with accessibility best practices, but the two use cases I've seen (code block line numbers, watermarks) might both be considered artifacts. Perhaps we could automatically put unselectable text in `pdf.artifact()`?

However, are *all* use cases of unselectable text artifacts? I can't think of any exceptions at the moment, but presumably they could exist?...

laurmaedje:
Regarding accessibility: cc @saecki @reknih

saecki:
> I'm not very familiar with accessibility best practices, but the two use cases I've seen (code block line numbers, watermarks) might both be considered artifacts. Perhaps we could automatically put unselectable text in `pdf.artifact()`?
> 
> However, are _all_ use cases of unselectable text artifacts? I can't think of any exceptions at the moment, but presumably they could exist?...

Considering that the text is inaccessible when drawn as plain glyphs, it should always be considered an artifact. Failing to mark it as an artifact would imply that it is somehow real content, but since it isn't accessible it wouldn't comply with the PDF/UA-1 spec.

> I'm wondering whether we should just disallow that option for PDF/UA-1? Since it's really not encouraged to do this normally.

Yeah it feels a little too easy to degrade text this way. I'm wondering whether it may be a better idea to make this a semantic element, to make the consequences more obvious. Something like a more generalized `pdf.artifact` that has an `unselectable` parameter maybe?

andylizi:
> I'm wondering whether it may be a better idea to make this a semantic element, to make the consequences more obvious. Something like a more generalized `pdf.artifact` that has an `unselectable` parameter maybe?

Just to throw some ideas out there, how about a new element under the Visualize section, along with `line`, `rect`, and friends. Here's what the docs say about Visualize:

> All shapes and paths drawn by Typst are automatically marked as artifacts to make them invisible to Assistive Technology (AT) during PDF export. However, their contents (if any) remain accessible.

Seems to fit pretty well in our case. We could call it `text-path` or `outlined-text` or some other bikesheddable name.

A potential drawback is that we'd lose the most natural way to handle text in SVG images, `show image: set text(selectable: false)`. Although it's not clear to me what additional use cases this would support, except for consistency with regular text, so maybe it's alright not to support it.

---

> I'm wondering whether we should just disallow that option for PDF/UA-1? Since it's really not encouraged to do this normally.

The current documentation for `pdf.artifact` explicitly mention watermarks as one example of artifact usage, it'd seem strange if we can't actually create such a (well-functioning) watermark in PDF/UA-1.

reknih:
My two cents would be to leave the function as-is and force the user to manually wrap it in an artifact.

Furthermore, I am not even sure that it would be allowed in UA either way. It goes against the PDF 2.0 spirit of having the artifacts in tree and being able to tag their insides, should the AT user choose to inspect the artifact.

Finally, a document using this feature will for sure fail [WCAG 2.2 clause 1.4.5](https://www.w3.org/WAI/WCAG21/Understanding/images-of-text.html) and documents using the function will thus fail to achieve WCAG AA conformance as well as conformance with standards based on it (e.g. EN 301 549). So that must be documented.

andylizi:
> force the user to manually wrap it in an artifact

Okay, I've implemented the check. There are still a few design decisions left to resolve through.

1. Should `pdf.artifact` be required even when the export target is not PDF or when using `--no-pdf-tags`? Right now unselectable text only works for PDFs, and the only implementation method would also make the text inaccessible, hence `pdf.artifact`. But in the future, if HTML export can use the `user-select: none` CSS property to create unselectable text that kept being accessible, it won't make much sense to mandate `pdf.artifact` in that case.
    - Furthermore, there is a note to consider adding `aria-hidden="true"` to `pdf.artifact` semantics in the future, which would make even less sense for unselectable text in HTML.
2. How should this requirement interact with embedded text in SVG images? Right now we have to wrap the entire image in an artifact tag. Not sure how desirable this would be in practice.

The current wording of the diagnostics could probably be improved as well.

---

> Furthermore, I am not even sure that it would be allowed in UA either way. It goes against the PDF 2.0 spirit of having the artifacts in tree and being able to tag their insides, should the AT user choose to inspect the artifact.

I found this specification on [Well-Tagged PDF in PDF 2.0](https://pdfa.org/wp-content/uploads/2024/02/Well-Tagged-PDF-WTPDF-1.0.pdf) on the PDF Association website, and it seems to suggest watermarks are a valid use of the artifact tag:

> A watermark is tagged with an **Artifact** structure element with a **Subtype** attribute with a value of _Watermark_.

Adobe guidelines also [talk about](https://helpx.adobe.com/ca/acrobat/using/creating-accessible-pdfs.html) watermarks not being part of the tag tree:

> You can add a watermark to a tagged PDF without adding it to the tag tree. Not having a watermark appear in the tag tree is helpful for people who are using screen readers, because they won’t hear the watermark read as a document content.
>
> The best way to add a watermark that doesn’t interfere with screen readers is to insert an untagged PDF of the watermark into a tagged PDF.

reknih:
On 1: I would not require it if tagging was disabled.

> I found this specification on [Well-Tagged PDF in PDF 2.0](https://pdfa.org/wp-content/uploads/2024/02/Well-Tagged-PDF-WTPDF-1.0.pdf) on the PDF Association website, and it seems to suggest watermarks are a valid use of the artifact tag:

I am aware that watermarks are a valid use of artifacts, my remark was on outlining the text within: This takes away the user's possibility to inspect the artifact in PDF 2.0. Put differently, what purpose does a "confidential" watermark serve when blind users can prove in court that they could not have been aware of it?

laurmaedje:
The motivating examples here are line numbers and water marks. Both of these should be emitted as normal, accessible text with an artifact tag and a subtype. There is a `Watermark` subtype in PDF 1.7, see Table 330 PDF 32000:1-2008 and there is a `LineNum` subtype in PDF 2.0, see Table 363 in ISO/DIS 32000-2. Both of these appear along the top-level artifact type `Pagination`.

Thus, it should be the responsibility of the viewer to turn the text unselectable in these cases instead of us outlining it. This also makes sense in thus far that both line numbers and watermark content _can_ be relevant information for people with vision impairment.

Currently, these subtypes are not exposed on `pdf.artifact`. There was work in this direction in https://github.com/LaurenzV/krilla/pull/273 though it looks a little stalled.

In any case, this leaves the actual `set text(selectable: false)` with somewhat thin motivation (there might be other cases?).

w1th0utnam3:
> In any case, this leaves the actual `set text(selectable: false)` with somewhat thin motivation (there might be other cases?).

I was looking for a feature like this when I was using icon fonts such as FontAwesome which result in random Unicode sequences when copy & pasting. Using fonts for this (instead of embedding the SVGs) is very convenient because you can more easily adapt size, color and alignment to the document text. However, a better solution than non-selectable text would be to turn it into a path and even add alt-text to it, so I guess it's also not a very strong use case.

andylizi:
I can think of other use cases where **text is intended as graphical elements rather than readable content**. For example, ASCII diagrams:

```
                    ┌───────────────────────┐
                    │   Is it supposed to   │
                    │        move?          │
                    └───────────┬───────────┘
                                │
                ┌───────────────┴───────────────┐
                │                               │
              YES                              NO
                │                               │
    ┌───────────▼───────────┐       ┌───────────▼───────────┐
    │     Is it moving?     │       │     Is it moving?     │
    └───────────┬───────────┘       └───────────┬───────────┘
                │                               │
        ┌───────┴───────┐               ┌───────┴───────┐
        │               │               │               │
       YES              NO              YES              NO
        │               │               │               │
        │     ┌─────────▼─────────┐     │     ┌─────────▼─────────┐
        │     │     Use WD-40     │     │     │      You're       │
        │     └───────────────────┘     │     │       fine        │
        │                               │     └───────────────────┘
        │                               │
        │                     ┌─────────▼─────────┐
        │                     │   Use Duct Tape   │
        │                     └───────────────────┘
        │
        └──────────────►  You're fine
```

Here, it might be desirable to make Unicode box-drawing characters unselectable without affecting the labels inside.

There is also the [treet](https://typst.app/universe/package/treet/) package that generate tree lists like this:

```
root_folder
├─ sub-folder
│ ├─ 1-1
│ │ └─ 1.1.1
│ └─ 1.2
│ ├─ 1.2.1
│ └─ 1.2.2
└─ 2
```

Another more niche example is the "glitch text" effect.

> H̶͉̆e̸̠̚r̸̤̒ẻ̵̩'̸͕͝s̸̝̑ ̷̞̂à̸̹n̸̛̺ ̵̭͆e̷̞̕x̷̩̏ā̵̞m̵̞̃p̷̲̕l̶͇̽e̵̡̓.̵̙̍ ̷̟̀I̸̱͝'̶̹͋m̸̥̅ ̵͚̋s̵̱̕ȗ̷͈r̵͎̊ȅ̵̡ ̷͒ͅy̶͈̐o̶̢̽ṷ̸̓'̶͖͆v̸̫͐è̸͕ ̴͚̚s̵̛̳ë̶̬́e̴̢͌n̶̞̄ ̴̳͝t̴͉͛h̵͕̽i̵̜̇s̴̮̄ ̷̦̍ơ̶ͅn̶͉̽ ̸̛͙t̷͌ͅh̸̦͘ë̷̳́ ̷̬̇í̸͉ǹ̸̼t̴͙͑ë̵̝r̸͕͘n̸̠͝e̶̳̎t̸̝̾ ̵̡̛a̵̫͌t̷̙͛ ̴̳̀s̵̢̊ó̴̜m̷̭̂e̷͔̿ ̴̼̆p̵͍̐ő̵͕ǐ̶̢ň̴͕t̸̙́.̷̗̎

I saw this used in some web novels to add flavor to scenes, but it always makes me wince because of how terrible they are for screen readers. But with unselectable text, maybe it'd be possible to keep the visual presentation without affecting accessibility.

reknih:
> I can think of other use cases where text is intended as graphical elements rather than readable content. For example, ASCII diagrams:

You can actually make these accessible by adding an alternative description using the `figure` function:

```typ
#figure(
  alt: "Here's an example I'm sure you've seen on the internet at some point rendered with a lot of accents and decorators so it looks spooky",
)[
  H̶͉̆e̸̠̚r̸̤̒ẻ̵̩'̸͕͝s̸̝̑ ̷̞̂à̸̹n̸̛̺ ̵̭͆e̷̞̕x̷̩̏ā̵̞m̵̞̃p̷̲̕l̶͇̽e̵̡̓.̵̙̍ ̷̟̀I̸̱͝'̶̹͋m̸̥̅ ̵͚̋s̵̱̕ȗ̷͈r̵͎̊ȅ̵̡ ̷͒ͅy̶͈̐o̶̢̽ṷ̸̓'̶͖͆v̸̫͐è̸͕ ̴͚̚s̵̛̳ë̶̬́e̴̢͌n̶̞̄ ̴̳͝t̴͉͛h̵͕̽i̵̜̇s̴̮̄ ̷̦̍ơ̶ͅn̶͉̽ ̸̛͙t̷͌ͅh̸̦͘ë̷̳́ ̷̬̇í̸͉ǹ̸̼t̴͙͑ë̵̝r̸͕͘n̸̠͝e̶̳̎t̸̝̾ ̵̡̛a̵̫͌t̷̙͛ ̴̳̀s̵̢̊ó̴̜m̷̭̂e̷͔̿ ̴̼̆p̵͍̐ő̵͕ǐ̶̢ň̴͕t̸̙́.̷̗̎
]
```

AT will treat this element as a leaf, i.e. it will read the alt text and, by default, not expose the capability to drill down into the contents. However, the original text stays selectable and can be copied from the PDF.

A limitation is that due to #7001, this pattern cannot be used for inline content. IMO, we might need a function to provide an alternative description that is not tightly coupled to figures.

This will satisfy [WCAG 2.2 Success Criterion 1.1.1 Non-text Content](https://www.w3.org/TR/WCAG22/#non-text-content) where non-text content [is defined to include ASCII art](https://www.w3.org/TR/WCAG22/#dfn-non-text-content). Also see [F72: Failure of Success Criterion 1.1.1 due to using ASCII art without providing a text alternative](https://www.w3.org/WAI/WCAG22/Techniques/failures/F72).

w1th0utnam3:
> A limitation is that due to #7001, this pattern cannot be used for inline content. IMO, we might need a function to provide an alternative description that is not tightly coupled to figures.

I also encountered this problem in my icon font use case.

reknih:
> I was looking for a feature like this when I was using icon fonts such as FontAwesome which result in random Unicode sequences when copy & pasting.

For this use case, I'd say that we have a multi-layered problem:

1. For icon font glyphs that match existing Unicode codepoints semantically, the icon font should map them to the appropriate code point. E.g., `fa-flag` should be mapped to U+1F3F3 Waving White Flag or U+1F3F4 Waving Black Flag so it is natively announced correctly and can be copy-pasted even if the icon font is not available at the target. Not all icon font glyphs may be assignable to well-known codepoints, so they have to be assigned to codepoints in the Unicode Private Use Area (PUA) and be made accessible using the second technique.

2. For icons that cannot be represented using Unicode codepoints outside the PUA, an alternative description is required. For block-level content, it can be provided using the above technique with a figure. For inline icon use, we need #7001 fixed or a dedicated alt text function that allows us to attach alternative descriptions to any content.

3. There are two cases where we may need to override what codepoint the icon font matches a glyph to:
   1. The font matches its icons that are representable in non-PUA Unicode codepoints to the PUA. In that case, we'd want to override the codepoint for an icon glyph with its well-known codepoint.

       See Section 8.5.2 of PDF/UA-2:

       > As no pre-defined meaning is associated with Unicode values in the Private Use Area (PUA), PUA values in content streams shall be used only if no other valid Unicode value is available.

   2. The font matches some or all of its icons that are not representable in non-PUA Unicode codepoints to inappropriate codepoints outside of the PUA. In that case, we'd want to map the codepoint back to the PUA. In addition, an alternative description is required, as described above.
    
   For the first of these problems, we can use the PDF capability `/ActualText` to override the codepoint for one or multiple glyphs locally. This matches the guidance in PDF/UA-2, Section 8.5.3:
   
   > In all cases where real content maps to PUA values an ActualText or Alt entry shall be present.

   (in the spec, real content refers to content not wrapped in an artifact)

   If `ActualText` is used, readers will copy and paste and expose to AT the contents of `ActualText` instead of the underlying code point. If the `ActualText` codepoint is not in the PUA, the icon is now accessible. `ActualText` is not currently exposed, but could be with a `pdf.actual-text()[]` function. To my knowledge, there is no corresponding mechanism in HTML or SVG. However, the reader support is spotty. Given that in almost all cases, we want to map a single glyph to the same codepoint across all of its uses, we may want to rewrite the font's `ToUnicode` cmap instead, a global lookup table that governs what Unicode codepoint is assigned to a particular glyph. The API design for this would be more complex, but we could generate a HTML webfont that incorporates the changes, so it would not be limited to PDF.

   Under PDF/UA-2 Section 8.7, `ActualText` and `Alt` cannot contain PUA codepoints. Since, at the same time, we may need to reassign codepoints to the PUA, we must have a mechanism to modify the `ToUnicode` cmap (see 2.ii. above) to accommodate this use case.

TL;DR: Don't outline icon fonts, instead, you need alt text and/or code point reassignment through `ActualText` and/or the `ToUnicode` cmap.

laurmaedje:
@reknih and I discussed this a bit more and reached the following conclusion:

Whether text is selectable is conceptually orthogonal to how it behaves w.r.t text extraction and accessibility. This means that we should encode it in a way that attempts to mirror this. If extractability is _also_ undesirable that's a separate concern and the job for a figure with alt text, a future non-figure alt text mechanism, or `pdf.artifact`. This interpretation is also consistent with how `user-select: none` works in browsers, which means it cleanly maps to HTML export.

The closest we can get to this ideal is to outline the text (the only way to consistently make it unselectable afaik), but then automatically add `/ActualText` to the outlined portion. The `/ActualText` should be added on the structure element (`/Type /StructElem`) rather than in the property dictionary of the marked content operator (`/Span <<..>> BDC`) as this has better viewer support. From the viewers I tested, the result unfortunately only remains accessible in Acrobat and pdf.js, not in Quartz and pdfium, but at least the information is there in principle.

Going through the use cases discussed above:
- A watermark could use `selectable: false` or keep it as `true`, depending on preference, and additionally use `pdf.artifact(kind: "watermark")`
- Icon font usage could use `selectable: false` if desired and combine it with a future alt text mechanism
- ASCII art could use either
  - `selectable: false` on the entire art and `figure(alt: ..)` around everything
  - `selectable: false` only on the box-drawing characters and more fine-grained alt text using a future alt text mechanism for portions of the diagram
- Spooky text could use `selectable: false` if desirable and a figure or future alt text mechanism around everything

@LaurenzV Do you think this is reasonable and how should this be approached w.r.t to krilla? If we go down this road, I think it would make some sense to let krilla handle this automatically, especially since the same handling should probably apply within `krilla-svg` (thinking of `embed_text`). If you would want to keep this behavior optional, `outlined` could become an enum with three modes (or an enum with two modes wrapped in `Option`).

LaurenzV:
> The /ActualText should be added on the structure element (/Type /StructElem) rather than in the property dictionary of the marked content operator (/Span <<..>> BDC) as this has better viewer support. 

Curious which viewers only support one but not the other?

Allowing to write `ActualText` on structure elements sounds fine to me, probably just a single setter that needs to be exposed, right?

> Do you think this is reasonable and how should this be approached w.r.t to krilla? 

I'm unsure I fully understand what changes you are proposing to krilla apart from the `ActualText` thing. 🤔 You mean krilla should detect things like watermarks automatically somehow?

reknih:
> > The /ActualText should be added on the structure element (/Type /StructElem) rather than in the property dictionary of the marked content operator (/Span <<..>> BDC) as this has better viewer support.
> 
> Curious which viewers only support one but not the other?

At least pdf.js. We only tested Quartz, Acrobat, (those support both) and Pdfium (supports neither)

> Allowing to write `ActualText` on structure elements sounds fine to me, probably just a single setter that needs to be exposed, right?

Wouldn't that lead to a lot of structure elements for multi-character words?

> You mean krilla should detect things like watermarks automatically somehow?

No, the API is all we need

laurmaedje:
> I'm unsure I fully understand what changes you are proposing to krilla apart from the ActualText thing. 🤔 You mean krilla should detect things like watermarks automatically somehow?

I was proposing that krilla's `outlined: bool` would be expanded to support three states (normal text, paths with actual text, paths without actual text) and that krilla-svg would likewise supports the same three options for what is now `embed_text: bool`. Then, krilla would automatically take care of writing the actual text (though I'm not familiar enough with the code to judge whether krilla could cleanly do this automatically if the `/ActualText` shall go into the tag tree instead of into the content stream.

If we do this fully in Typst (adding just a tag tree setter in krilla), I don't think there'd be a way to give text in SVGs the same treatment.

LaurenzV:
Hmm okay. Would be good to see it, but don't think I have much reason to be opposed!

laurmaedje:
@andylizi Any thoughts on the points above? If this is getting too complex and beyond the amount of work you wanted to invest, I fully understand. Then, I would just cross-link the collected points of discussion in the issue, close the PR, and leave this as future work.

andylizi:
@laurmaedje The new approach sounds much better, and I'd be happy to take a stab at it at some point!

I'm a bit ambivalent about closing this PR. On one hand, most of the Typst-side code could be reused (merge conflicts aside). On the other hand, since the bulk of the work will need to land in krilla first, the Typst PR likely won't be ready anytime soon. So closing it to reduce clutter and avoid leaving it in limbo probably makes sense.

(In that regard, I won't delete the branch -- I hope I'd remember to not accidentally delete it in the future xD)

P.S.: I don't have much domain knowledge around krilla or the `/ActualText` stuff, and I've been a bit busy lately. So while I'd like to come back to this, if anyone wants to take it on in the meantime, please feel free to go ahead!

PR #7298, opened 2025-11-04
Rename `global_attributes` to camelCase when serializing as `docs.json`

### LINKED ISSUE #7297: `global_attributes` is not renamed to camelCase in `docs.json`
In the typst-docs crate, most types set `#[serde(rename_all = "camelCase")]`, except `GroupModel`, which has just added a snake case field `global_attributes` in #7083.

https://github.com/typst/typst/blob/17a7890b2dc47da390d194d3593ed9a8b5668169/docs/src/model.rs#L169-L170

https://github.com/typst/typst/blob/17a7890b2dc47da390d194d3593ed9a8b5668169/docs/src/model.rs#L136-L144

This isn't really a bug, but the inconsistency of naming in `docs.json` may cause bugs in downstream projects.
Could `global_attributes` be renamed to camelCase as well?

YDX-2147483647:
Resolves #7297

I'm not sure if this is appropriate (for example, you may have to change the closed-source part), but it's too simple.
To resolve the problem quickly, I'm submitting this PR.

### DIFF
--- a/docs/src/model.rs
+++ b/docs/src/model.rs
@@ -135,7 +135,7 @@ pub struct StrParam {
 
 /// Details about a group of functions.
 #[derive(Debug, Serialize)]
-#[serde(rename_all = "camelCase")] 
+#[serde(rename_all = "camelCase")]
 pub struct GroupModel {
     pub name: EcoString,
     pub title: EcoString,


laurmaedje:
Thanks! This was an oversight.

PR #7296, opened 2025-11-04
Fix frame size produced by `math.accent`

mkorje:
Closes #5944.

The frame size is not calculated to incorporate the accent if either:
- the accent was not scaled, and so the glyph's width may not be reliable; or
- the accent's width is less than the base's, even if its positioning would place part of the accent's bounding box outside the resulting frame.

I think this behaviour addresses the concerns in the linked issue quite well, ensuring large accents have the correct frame size but avoiding unpleasant excess spacing in normal sized accents. See the new `math-accent-frame-size` test below.

<details>
<summary>Updated tests</summary>

```typ
--- math-accent-bottom-sized ---
// Test bottom accent size.
$accent(sum, \u{0330}), accent(sum, \u{0330}, size: #50%), accent(H, \u{032D}, size: #200%)$
```
| Old | New |
| - | - |
| math-accent-bottom-sized-OLD | math-accent-bottom-sized-NEW |
| math-accent-bottom-sized-MARKED-OLD | math-accent-bottom-sized-MARKED-NEW |

---

```typ
--- math-accent-frame-size ---
// Test final frame size.
$ x + dash(A, size: #2em) + y $
$ x + dash(M) + y $
```
| Old | New |
| - | - |
| math-accent-frame-size-OLD | math-accent-frame-size-NEW |
| math-accent-frame-size-MARKED-OLD | math-accent-frame-size-MARKED-NEW |

---

```typ
--- math-accent-sized ---
// Test accent size.
$tilde(sum), tilde(sum, size: #50%), accent(H, hat, size: #200%)$
```
| Old | New |
| - | - |
| math-accent-sized-OLD | math-accent-sized-NEW |
| math-accent-sized-MARKED-OLD | math-accent-sized-MARKED-NEW |

</details>

(I'm doing this as, when working on https://github.com/typst/typst/issues/6649, the under/over things were overlapping...)

Issue #3784, opened 2024-03-24
Multiple underbrace text not on the same height resp. horizontally aligned


Fabioni:
### Description

I think it would be nicer if the text of the `underbrace`s would be on the same height. Notice that the word "distributional" is slightly higher than "aleatoric" and "epistemic"

This is from Typst:
<img width="526" alt="image" src="https://github.com/typst/typst/assets/45362676/f517d5b1-b89a-4fb7-9244-b1b476ccb4c2">
using:
```typst
$ p(y^*|x^*, cal(D)) = integral_theta integral_mu underbrace(p(y^*|mu), "aleatoric") thick underbrace(p(mu|x^*, theta), "distributional") thick underbrace(p(theta|cal(D)), "epistemic") thick d mu thin d theta $ 
```

This is from Latex:
<img width="590" alt="image" src="https://github.com/typst/typst/assets/45362676/2cddb528-f014-4941-9b11-b89c28e60e76">

### Operating system

Web app

### Typst version

- [X] I am using the latest version of Typst

Leedehai:
TL;DR: it looks quite contrived.

---

It appears this is the result of the different glyph renderings of the underbraces. If you look at it closely, you can notice the underbraces all look a bit differently.

In fact, if I make the middle term narrower, the middle underbrace will use _nearly_ the same glyph as the other two, and then the heights of all three _look_ aligned:

```typ
// Fig 1 (original)
$ underbrace(p(y^*|mu), "aleatoric") thick underbrace(p(mu|x^*, theta), "distributional") thick underbrace(p(theta|cal(D)), "epistemic") $ 
```
```typ
// Fig 2 (modified)
$ underbrace(p(y^*|mu), "aleatoric") thick underbrace(p(mu|x^*), "distributional") thick underbrace(p(theta|cal(D)), "epistemic") $ 
```
Fig 1: <img width="260" src="https://github.com/typst/typst/assets/18319900/9f1cb8e7-c3f1-4ae7-9bd1-92f9a08ac485"> Fig 2: <img width="260" src="https://github.com/typst/typst/assets/18319900/e8d24b44-c44d-40a0-835d-4705b2834abc">

Each term is made of three frames stacked vertically: the term itself, the underbrace, and the annotation text.

From left to right, the underbrace frames' `height()`s are:
* Fig 1: `3.4699999999999998pt`, `2.4399999999999995pt`, `3.38pt`;
[...]

PR #7299, opened 2025-11-04
Layout Under/Over elements in math as "accents"

### LINKED ISSUE #6649: Layout Under/Over elements in math as "accents"
The [Under/Over](https://typst.app/docs/reference/math/underover) elements in math (excluding `math.underline` and `math.overline`) are currently laid out in a very "custom" way, different from everything else in math: they are laid out in a stack with a hard-coded gap of `0.25em`

This is very different from the way other engines do it. Both LuaTeX and the MathML spec lay them out as _accents_. You can see this mainly by way of the smaller gap between the base and the overbrace.[^1]

<details>

<summary>LuaTeX</summary>

```tex
\makeatletter
\renewcommand\@xipt{10}
\makeatother
\documentclass[10pt]{article}
\usepackage[paperheight=2cm,paperwidth=4cm,margin=0cm]{geometry}
\geometry{textwidth=\paperwidth, textheight=\paperheight, noheadfoot, nomarginpar}
\setlength{\topskip}{0mm}
\setlength{\parindent}{0mm}
\usepackage{amsmath}
\usepackage{unicode-math}
\usepackage{fontspec}
\usepackage[default]{fontsetup}

\begin{document}

\[ \]

\[ \overbrace{x + \cdots + x}^{k \text{ times}} \]

\end{document}
```

Image

</details>

<details>

<summary>MathML</summary>

```html
<math display="block">
  <mover>
    <mrow>
      <mi>x</mi>
      <mo>+</mo>
      <mo>...</mo>
      <mo>+</mo>
      <mi>x</mi>
    </mrow>
    <mover>
      <mo>⏞</mo>
      <mrow>
        <mi>k</mi>
        <mspace width="0.1111111111111111em"></mspace>
        <mtext>times</mtext>
      </mrow>
    </mover>
  </mover>
</math>
```

From https://fred-wang.github.io/MathFonts/mozilla_mathml_test/?fontFamily=NewComputerModern.

</details>

<details>

<summary>Typst</summary>

```typ
--- playground ---
$ overbrace(x + dots.c + x, k "times") $
```

Image

</details>

This came up while working on MathML export, where there is the choice of whether to use `accent="true"` with them in `mover`. Note that the Under/Over element functions would still exist, just that they would share a similar layout to accent (like how matrices, vectors and cases all share the same layout internals).

https://github.com/typst/typst/issues/422 is basically then a prerequisite for this change, which I should have a PR for very soon.

[^1]: It's interesting to note the different ways engine represent an over/under with an annotation. In LuaTeX it is the base with an accent, where you then attach a limit for the annotation. Whereas in the example for MathML it is the base with an accent, but the accent is itself the over/under as a base with a limit for the annotation.

mkorje:
Closes #6649. ~~This PR depends on #7296.~~ I've unblocked this.

I've opted to use the internal layout for accents and limits rather than simply turning under/over element directly into an AccentElem/AttachElem, since I wanted to keep the width of the glyph the same and didn't want show rules on those elements to affect under/over. TLDR, the only visible change is that the under/over things are closer to their body.

<details>

<summary>Updated tests</summary>

```typ
--- math-delim-show-rule-3 ---
#show "⏟": set text(fuchsia)
$ underbrace(1 + 1 = 2, "obviously") $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-delim-show-rule-3-OLD | math-delim-show-rule-3-OLD | math-delim-show-rule-3-NEW |

---
            
```typ
--- math-underover-line-bracket ---
// Test lines and brackets.
$ x = overbracket(
  overline(underline(x + y)),
  1 + 2 + ... + 5,
) $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-underover-line-bracket-OLD | math-underover-line-bracket-OLD | math-underover-line-bracket-NEW |

---
            
```typ
--- math-underover-shells ---
// Test tortoise shell brackets.
$ undershell(
  1 + overshell(2 + ..., x + y),
  "all stuff"
) $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-underover-shells-OLD | math-underover-shells-OLD | math-underover-shells-NEW |

---
            
```typ
--- math-underover-brace ---
// Test braces.
$ x = underbrace(
  1 + 2 + ... + 5,
  underbrace("numbers", x + y)
) $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-underover-brace-OLD | math-underover-brace-OLD | math-underover-brace-NEW |

---
            
```typ
--- math-underover-brackets ---
// Test brackets.
$ underbracket([1, 2/3], "relevant stuff")
          arrow.l.r.double.long
  overbracket([4/5,6], "irrelevant stuff") $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-underover-brackets-OLD | math-underover-brackets-OLD | math-underover-brackets-NEW |

---
            
```typ
--- math-underover-parens ---
// Test parentheses.
$ overparen(
  underparen(x + y, "long comment"),
  1 + 2 + ... + 5  
) $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-underover-parens-OLD | math-underover-parens-OLD | math-underover-parens-NEW |

---
            
```typ
--- math-underover-multiline-annotation ---
// Test that multiline annotations do not change the baseline.
$ S = overbrace(beta (alpha) S I, "one line")
    - overbrace(mu (N), "two" \  "line") $
$ S = underbrace(beta (alpha) S I, "one line")
    - underbrace(mu (N), "two" \  "line") $
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-underover-multiline-annotation-OLD | math-underover-multiline-annotation-OLD | math-underover-multiline-annotation-NEW |

---
            
```typ
--- math-spacing-decorated ---
// Test spacing for operators with decorations and modifiers on them
#set page(width: auto)
$a equiv b + c - d => e log 5 op("ln") 6$ \
$a cancel(equiv) b overline(+) c arrow(-) d hat(=>) e cancel(log) 5 dot(op("ln")) 6$ \
$a overbrace(equiv) b underline(+) c grave(-) d underbracket(=>) e circle(log) 5 caron(op("ln")) 6$ \
\
$a attach(equiv, tl: a, tr: b) b attach(limits(+), t: a, b: b) c tilde(-) d breve(=>) e attach(limits(log), t: a, b: b) 5 attach(op("ln"), tr: a, bl: b) 6$
```

| Old (Pre #7327) | Old (Post #7327) | New |
| - | - | - |
| math-spacing-decorated-OLD | math-spacing-decorated-OLD | math-spacing-decorated-NEW |

---
            
</details>


laurmaedje:
Could you elaborate on which kinds of consequences this would have in practice? I read through https://github.com/typst/typst/issues/6649, but I'm not yet sure what the "real world diff" is.

mkorje:
It'll just be a slight difference in the gaps, both between the base and the under/over glyph, and between the under/over glyph and the annotation (if it exists). Instead of the currently hard-coded 0.25em for both of these gaps, it will instead use the accent layout calculated gap for the first, and then the attach layout calculated gap for the second. This matches what LuaTeX does.

There was some discussion on Discord about gap being too small now with this change, but I think things will be more amenable with #7327. Fixing #7327 increases the two gaps mentioned previously around some under/over glyphs (as the bounding box's height is not too short anymore), and then this PR will shift things again. For the default math font NewCM it wil most definitely make the first gap smaller (in comparison to pre-#7327) but the second will probably end up around the same as before #7327.

mkorje:
I've removed the dependency on https://github.com/typst/typst/pull/7296 and rebased to incorporate https://github.com/typst/typst/pull/7327. Have a look at the test images to get a better idea of the history of what's changed here. We should now match LuaTeX exactly, so in particular the examples from #3784 will be the same as LuaTeX now.

crates/typst-layout/src/math/underover.rs
```diff
@@ -1,24 +1,21 @@
...
-    OverbraceElem, OverbracketElem, OverlineElem, OverparenElem, OvershellElem,
+    Accent, OverbraceElem, OverbracketElem, OverlineElem, OverparenElem, OvershellElem,
     UnderbraceElem, UnderbracketElem, UnderlineElem, UnderparenElem, UndershellElem,
 };
 use typst_library::text::TextElem;
 use typst_library::visualize::{FixedStroke, Geometry};
 use typst_syntax::Span;
 
+use crate::math::accent::place_accent;
+use crate::math::attach::layout_attachments;
```
laurmaedje:
```suggestion
use super::accent::place_accent;
use super::attach::layout_attachments;
```

laurmaedje:
Thank you!

Looking at the diff of `math-underover-brace`, I was a bit skeptical because "numbers" really gets pretty close, but it certainly does match LuaLaTeX. It's also primarily extra tight when there's text above the brace as opposed to math content, which shouldn't be the main use case. The gap was also indeed a bit overlarge before in some cases and #7327 improved things for this PR. Overall, I think it's sometimes a bit better, sometimes a bit worse, but always fine. Given that, I think matching what other implementations do is both reasonable and justifiable.

The code changes look correct. I've compared the old and new accent code path and think the accent handling itself should behave exactly the same in the non-underover case.

For underover, of course behavior changes a bit, but it seems to have improved. I found the following case where the new output looks more correct:
```typ
$ x = underbrace(
  "complexity",
  // The alignment points are not shared between body and annotation.
  underbrace(&1+2 \ 3+&4, a b+&c \ d+&e)
) $
```

I've added a test to pin down this improved behavior. :)

PR #7310, opened 2025-11-05
Fix Safari WASM compiler crash

koalazub:
A little while ago I posted an issue about Safari [experiencing compiler crashes](https://github.com/typst/webapp-issues/issues/734#issuecomment-3370777309).

Safari Technology Preview 231 on macOS crashes when loading typst.app with runtime errors indicating out-of-bounds memory access. This occurs because Safari's WebAssembly implementation contains unresolved bugs when handling modules compiled with multivalue and reference-types features.

Although fixes have been merged in WebKit (https://github.com/WebKit/WebKit/pull/43753) and workarounds implemented in wasm-bindgen:
https://github.com/wasm-bindgen/wasm-bindgen/issues/4471
https://github.com/wasm-bindgen/wasm-bindgen/pull/4472
https://github.com/wasm-bindgen/wasm-bindgen/pull/4612

These will take time to propagate to Safari users. As noted in the wasm-bindgen discussion, "it's going be a while before it is widely adopted" and older Safari versions are unlikely to receive backported fixes.

Rust 1.82 onwards enables these features by default for WASM targets, causing previously working applications to crash in Safari while continuing to function correctly in Chrome and Firefox.

This change disables the problematic WASM features through explicit compiler flags, restoring Safari compatibility without affecting other browsers. Safari also implements a more restrictive default stack size than other browsers, and since dynamic stack growth through the stacker library is unavailable on WASM targets, the stack size must be statically configured to prevent stack overflow during compilation of complex documents.

This requires testing in the actual web application environment. The fix should be validated in a staging or development deployment of typst.app using Safari Technology Preview to confirm it resolves the crash before merging to production.


laurmaedje:
Thanks for the investigation. These changes would have to be made in our private Rust code base rather than here to take effect. 

Do you have a reliable reproduction for https://github.com/typst/webapp-issues/issues/734#issuecomment-3370777309? Otherwise, it may be hard to test whether this fix does indeed work.

koalazub:
> Thanks for the investigation. These changes would have to be made in our private Rust code base rather than here to take effect.
> 
> Do you have a reliable reproduction for [typst/webapp-issues#734 (comment)](https://github.com/typst/webapp-issues/issues/734#issuecomment-3370777309)? Otherwise, it may be hard to test whether this fix does indeed work.

Ah yup I discovered that the web app itself is closed and made a mention at the end about needing to have it tested internally. 

By reliable do you mean the reproduction link shared in https://github.com/typst/webapp-issues/issues/734#issuecomment-3370777309? 

If you'd like the link mentioned in it it's https://typst.app/project/pOpDhGqhfbJfYA3GjGjPjS

laurmaedje:
> By reliable do you mean the reproduction link shared

Oh, I missed that link. But more generally I was wondering whether the crash happens always or sometimes.

koalazub:
> > By reliable do you mean the reproduction link shared
> 
> Oh, I missed that link. But more generally I was wondering whether the crash happens always or sometimes.

Yes this happens 100% of the time. I've had to swap to Chrome in order to use typst.app. There was a short time prior to macOS 26 that that wasn't the case. But as mentioned in the PR, this was as stop-gap to prevent certain wasm features from being enabled in Safari until the team have addressed the issues. The bug specifically is Safari's JavaScript engine (JavaScriptCore) incorrectly handling WebAssembly modules that use the multivalue and reference-types features

koalazub:
@laurmaedje I can see you marked this as a draft, but what's the plan here? This is a ready fix and not just a plan

laurmaedje:
I only marked this as draft because merging it here would have no effect. I need to test it with the web app and apply the changes there.

laurmaedje:
Unfortunately, neither the cargo config changes nor the wasm-bindgen bump seem to fix the issue. I'll have to debug more, but it's pretty tricky as WebKit gives _very_ little information as to what's going wrong.

koalazub:
> Unfortunately, neither the cargo config changes nor the wasm-bindgen bump seem to fix the issue. I'll have to debug more, but it's pretty tricky as WebKit gives _very_ little information as to what's going wrong.

Understandable. I tried to update the config whilst in the dark regarding the web app. So it was just that - a stab in the dark to see if that would ameliorate the issue. 

Thanks for validating the outcome

laurmaedje:
We're still in the process of trying to determine the cause, but since this fix doesn't seem to be the solution (and since it would have to be applied elsewhere anyway), I'll already close this PR. Thanks for your work on this!

PR #7314, opened 2025-11-05
Allow `typst init` to target the current directory

### LINKED ISSUE #7259: Allow `typst init` to target the current directory (CLI)
When creating a new typst project with `typst init`, for example like this,
```shell
typst init @preview/example-template new-exciting-project
```
you are not allowed to target the current directory. The third argument (`DIR`) should be optional, or allow the user to specify `.` to use the current directory. I think the `uv` project manager for python is a great example:

> [Creating a new project](https://docs.astral.sh/uv/guides/projects/#creating-a-new-project)
> You can create a new Python project using the uv init command:
> ```shell
> uv init hello-world
> cd hello-world
> ```
> Alternatively, you can initialize a project in the working directory:
> ```shell
> mkdir hello-world
> cd hello-world
> uv init
> ```

Currently, the CLI gives the following error message when I try to create a new typst project in an empty directory:
```shell
mkdir test-project
cd test-project
typst init @local/ubc-report .

error: project directory already exists (at .)
```

Typst version: `typst 0.14.0 (unknown hash)`

CrumblyLiquid:
Fixes #7259

It just removes the check for existing directory.

Because `CopyOptions` sets the `overwrite` flag by default, the copying `fs_extra::dir::copy` function will error if it tries to overwrite an existing file.

The error seems to be pretty clear to me. For example: `error: failed to create project directory (Path "./main.typ" exists)`

### DIFF
--- a/crates/typst-library/src/loading/xml.rs
+++ b/crates/typst-library/src/loading/xml.rs
@@ -99,6 +99,7 @@ fn convert_xml(node: roxmltree::Node) -> Value {
         return Value::Array(children);
     }
 
+    let namespace = node.tag_name().namespace();
     let tag: Str = node.tag_name().name().into();
     let attrs: Dict = node
         .attributes()
@@ -106,6 +107,7 @@ fn convert_xml(node: roxmltree::Node) -> Value {
         .collect();
 
     Value::Dict(dict! {
+        "namespace" => namespace,
         "tag" => tag,
         "attrs" => attrs,
         "children" => children,
--- a/tests/suite/loading/xml.typ
+++ b/tests/suite/loading/xml.typ
@@ -2,20 +2,22 @@
 // Test reading XML data.
 #let data = xml("/assets/data/hello.xml")
 #test(data, ((
+  namespace: none,
   tag: "data",
   attrs: (:),
   children: (
     "\n  ",
-    (tag: "hello", attrs: (name: "hi"), children: ("1",)),
+    (namespace: none, tag: "hello", attrs: (name: "hi"), children: ("1",)),
     "\n  ",
     (
+      namespace: none,
       tag: "data",
       attrs: (:),
       children: (
         "\n    ",
-        (tag: "hello", attrs: (:), children: ("World",)),
+        (namespace: none, tag: "hello", attrs: (:), children: ("World",)),
         "\n    ",
-        (tag: "hello", attrs: (:), children: ("World",)),
+        (namespace: none, tag: "hello", attrs: (:), children: ("World",)),
         "\n  ",
       ),
     ),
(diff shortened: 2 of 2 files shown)


laurmaedje:
I'm fine with this on a semantical level. On a technical level, I have one further concern: This can currently lead to partial failure, where `typst init` fails because a file already exists, but a partially scaffolded project remains. It would be nice if this had more transactional semantics. I'm not sure how feasible that is, but we should at least look into it.

CrumblyLiquid:
There are two things that can be done here (I think):

1. Check the directory trees beforehand for any file collisions (this only covers file/directory collisions).
2. Keep a list of already created files and directories and remove them if failure is encountered (this covers both file/directory collisions and filesystem related errors that are outside of our control)

Only one of those or both could be used. Either way it might be better to implement it with `std::fs` rather than use `fs_extra` which is only used for recursive copying as far as I can tell and I couldn't find any function in it that would do recursive directory tree traversal so both options would have to be implemented manually I think.

I also think that the performance of `1.` would be fine as long as the template is somewhat structured into directories and not a bunch of files in the root directory.

What's your opinion on this?

saecki:
> There are two things that can be done here (I think):
> 
>     1. Check the directory trees beforehand for any file collisions (this only covers file/directory collisions).

I think I would prefer this.
 
> Only one of those or both could be used. Either way it might be better to implement it with `std::fs` rather than use `fs_extra` which is only used for recursive copying as far as I can tell and I couldn't find any function in it that would do recursive directory tree traversal so both options would have to be implemented manually I think.
> 
> I also think that the performance of `1.` would be fine as long as the template is somewhat structured into directories and not a bunch of files in the root directory.
> 
> What's your opinion on this?

The performance should be a non issue in any case.

I took a look at the `fs_extra` crate, and after re-reading the documentation of `CopyOption::content_only` 3 times, I'm still not sure what it does. I'm also not a fan of the API in general since it uses boolean flags for mutually exclusive options such as `overwrite` and `skip_exist`. And it seems to do a great big deal more than what we need.

We already have `walkdir` in the dependency tree multiple times, so we could use that. I would honestly be in favor of just writing our own `copy_dir` function, maybe as part of `typst-utils`, that:

1. Indexes the files in the `package` directory.
2. Checks if there are any collisions of files in the `project` directory.
3. Copies all files, and just creates all missing directories.

I think this should be quite straight forward. If you'd like I'd be up to implementing this myself.

laurmaedje:
Hi @CrumblyLiquid, are you still interested in working on this? If not, that's also totally fine, but then I would close the PR.

CrumblyLiquid:
Yes, I'll do it :)

CrumblyLiquid:
Well TIL that force pushing to sync up with main closes the PR :D (I'll reopen it once I have commits)

PR #7318, opened 2025-11-06
Update Czech translations

CrumblyLiquid:
Adds the remaining translations for the Czech language.

Note: Even though some people use "email", the proper form would be "e-mail" as "email" could also mean enamel in Czech ([source](https://prirucka.ujc.cas.cz/?slovo=e-mail)).


laurmaedje:
Thanks!

PR #7324, opened 2025-11-07
Improve regex text extraction example

3w36zj6:
This pull request fixes the broken example in the [`/reference/foundations/regex/#constructor-regex`](https://typst.app/docs/reference/foundations/regex/#constructor-regex) documentation. Previously, the code was presented as block content even though it doesn't work on its own, and there was an inconsistency with a period at the end.

Since Typst's `regex` is difficult to demonstrate in isolation, I've added simple examples using the show rule with both an escape sequence and a raw element to make things clearer.

### DIFF
--- a/crates/typst-library/src/foundations/str.rs
+++ b/crates/typst-library/src/foundations/str.rs
@@ -994,8 +994,17 @@ impl Regex {
         /// backslash, you would need to write `{regex("\\\\")}`.
         ///
         /// If you need many escape sequences, you can also create a raw element
-        /// and extract its text to use it for your regular expressions:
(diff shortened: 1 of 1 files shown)


laurmaedje:
There is already an existing PR: https://github.com/typst/typst/pull/7211

3w36zj6:
Oops, I understand the situation now.

After comparing this PR with the previous one, I noticed that the ranges of changes are slightly different. If you think the regex example code is useful, I plan to rebase this PR after the previous one is merged, so that there won't be any conflicts. What do you think?

laurmaedje:
Given that there is already a show rule example at the top of the page, I don't think we need the extra example and thus we're covered with #7211. I'll close this, but thanks anyway!

PR #7233, opened 2025-10-26
Allow dots in language tags

frozolotl:
Not every supported language can currently be highlighted without workarounds. One such example are Jinja2 HTML templates:

    ```html.j2
    <tbody>
      {% for row in data.rows %}
      <tr>
          {% for column in row %}
          <td>{{ column }}</td>
          {% endfor %}
      </tr>
      {% endfor %}
    </tbody>
    ```

This is due to dots not being valid language tags in 0.14.0.
**This change extends the syntax to allow dots at the start and after.**

---

Here's also a little list of syntaxes that are supported, but can't be used with the fence syntax.
With this change, all of them should be usable.

<div style="background-color:white;color:black;"><table><thead><tr><th>name</th><th>extensions</th></tr></thead><tbody><tr><td>Git Link</td><td>.git</td></tr><tr><td>JavaScript (Rails)</td><td>js.erb</td></tr><tr><td>CMake C Header</td><td>h.in</td></tr><tr><td>CMake C++ Header</td><td>hh.in</td></tr><tr><td>CMake C++ Header</td><td>hpp.in</td></tr><tr><td>CMake C++ Header</td><td>hxx.in</td></tr><tr><td>CMake C++ Header</td><td>h++.in</td></tr><tr><td>CMakeCache</td><td>CMakeCache.txt</td></tr><tr><td>HTML (EEx)</td><td>html.eex</td></tr><tr><td>HTML (EEx)</td><td>html.leex</td></tr><tr><td>Regular Expressions (Elixir)</td><td>ex.re</td></tr><tr><td>go.mod</td><td>go.mod</td></tr><tr><td>go.mod</td><td>go.work</td></tr><tr><td>go.sum</td><td>go.sum</td></tr><tr><td>HTML (Jinja2)</td><td>htm.j2</td></tr><tr><td>HTML (Jinja2)</td><td>html.j2</td></tr><tr><td>HTML (Jinja2)</td><td>xhtml.j2</td></tr><tr><td>HTML (Jinja2)</td><td>xml.j2</td></tr><tr><td>resolv</td><td>resolv.conf</td></tr><tr><td>Todo.txt</td><td>todo.txt</td></tr><tr><td>Todo.txt</td><td>done.txt</td></tr><tr><td>debsources</td><td>sources.list</td></tr></tbody></table></div>

---

There are a few languages that can't be addressed with the proposed syntax, but these are very minor and a change to allow them would likely involve some breakage:

[...]

Issue #7266, opened 2025-10-30
Show-set rules should not apply recursively


laurmaedje:
### Description

Let show-set rules only apply to the selected element itself (if show & set target the same element) or the closest descendants of the selected elements(if different kinds of elements are targeted). The exact semantics might need some extra figuring out.

### Use Case

In most use cases of show-set rules, we want to affect a specific element. For example, `show figure: set block(breakable: false)` is intended to affect precisely the figure block and `show quote: set block(spacing: ..)` should affect precisely the quote block. For rules such as `show heading.where(level: 1): set heading(supplement: [..])`, the recursive nature is also not needed, even if it rarely interferes.

Set rules currently always apply to a full subtree of content, but in the case of show-set rules this is often simply not desirable. In fact, I can barely think of cases where it's actually desirable for the show rule to affect the whole subtree. (`show ..: set text(..)` might be considered an exception, but text is always a leaf, so the semantics of both options are pretty much the same.) 

**I would be interested to hear about it if people have use cases where the current behavior is _better_ than the proposed one.**

MDLC01:
`show heading: set par(justify: false)` might be a weird case as well, as headings don't actually create paragraphs. Same with `align`.

MDLC01:
The main problem I have with this proposal is that I fear the semantics would be harder to grasp for most users. Especially because a show rule might itself create multiple nested instances of an element.

Instead, a similar solution that could work in most cases (i.e., other than `text`, `par.justify`, `align`, and maybe other similar elements) is to apply the set rule in `show x: set elem(..)` exactly to the content _created in the show rule_ for `x`. For example:

```typ
#show x: it => {
  elem() // The set rule applies to this.
  it // The set rule does not apply to this.
  it.field // The set rule does not apply to this either.
  foo(it.field) // The set rule applies to the elements created by `foo`, but not to `it.field`.
  it.helper() // Depends on the implementation of `x.helper`.
}
```

I think this would be easier to understand: the show-set rule `show x: set y(..)` applies to the `y`s created in `x`'s show rule.

Enivex:
A question on discord made me realize a relevant example. I'm not sure if this deserves its own issue or not.

If one wants to use a different font for italic text only, then `#show emph: set text(font: "other font name")` is at first glance the way to go. However, this will not work correctly, because a double `emph` will display upright text using the wrong font.

Lege19:
> A question on discord made me realize a relevant example. I'm not sure if this deserves its own issue or not.
>
[...]

PR #7337, opened 2025-11-08
Parse language tags until the first whitespace

frozolotl:
Alternative to #7233.

---

Not every supported language can currently be highlighted without workarounds. One such example are Jinja2 HTML templates:

    ```html.j2
    <tbody>
      {% for row in data.rows %}
      <tr>
          {% for column in row %}
          <td>{{ column }}</td>
          {% endfor %}
      </tr>
      {% endfor %}
    </tbody>
    ```

This is due to dots not being valid language tags in 0.14.0. This change extends the syntax to parse language tags until the first whitespace was reached.


bastienvoirin:
To address point 2 in https://github.com/typst/typst/pull/7233#issuecomment-3456293243, i.e. _using the [info string](https://spec.commonmark.org/0.31.2/#info-string) not only for the language but for a_ comma-separated or semicolon-separated _list of key-value pairs for other metadata_ than the language, in order to _highly customize `raw` using plugins_:

````
```html;highlight=1-10,12;escape-between=$$;line-numbers
<!DOCTYPE html>
<html>

</html>
```
````

1. If https://github.com/typst/typst/issues/7266 is solved, would something like

    ```typ
    // pseudo-code
    #show raw: it => {
        let lang = it.lang.split(";").first()
        raw(lang: lang, ..it) // the "..it" is pseudo-code
    }
    ```
    work then?

2. Couldn't we define _now_ (in this PR or a follow-up one) a separator like `;` and split the info string (the first part would be the language tag and the rest given as an _argument sink_ to `raw` or a `metadata` dictionary argument)?

**Regardless of the approach, I believe an argument sink in `raw` (for custom arguments without the `Unexpected argument` error) would unlock deep and powerful customization at the syntax-sugar-level, i.e. ```` ```lang;meta=val;etc text``` ```` instead of ```#raw(lang: lang, ..., text)```.**

laurmaedje:
Could you also update the docs here?

https://github.com/typst/typst/blob/64641d687fcd658c75fee84bd5d40f9bb7f8404f/crates/typst-library/src/text/raw.rs#L78

---

@bastienvoirin The way to do it would be with this mechanism: https://github.com/typst/typst/issues/7165. Having a built-in metadata mechanism there feels a bit ad-hoc to me. I think it's probably better left to users if they want to.

frozolotl:
> Could you also update the docs here?

Done.

laurmaedje:
Thank you!

isuffix:
Apologies for bringing this up for a feature that's already merged, but I don't think the breakage caused by this change is acceptable, and I'd like to discuss that.

My fear is that this will cause content changes to raw blocks that authors won't actually notice when updating their documents, and that this will affect not just public code and packages that are visible on GitHub, but the private documents of individual authors.

For example, I've written text like ```` ```$f(x)$``` ```` in some of my own documents discussing math syntax, which renders as `$f(x)$` in v0.14, but with this change would silently convert to rendering **no text at all** since the entire content of the raw block would become the language tag. This literally already happened in a PR that linked here above (https://github.com/continuous-foundation/scienceicons/pull/9), and so I don't think it will be uncommon at all.

Also, that PR got lucky in that their usage caused an actual error, but the nature of us reinterpreting the meaning of the raw syntax means that a lot of documents won't error, but will silently produce different output and many authors may not notice the changed content for a long time. This is the kind of insidious problem that causes me to lose trust in tools that I use, and I don't want us to pay that same price in broken trust with users of Typst accepting updates.

The reason I'm bringing this up now is because I've recently been working on enabling the ability to add warnings directly in the syntax tree when parsing/lexing. With that ability we can add warnings in v0.15 for any raw blocks that would be affected by this, giving us a chance to alert users to the new behavior before launching it for real in the next version.

So I think we should revert this change temporarily, delay its rollout for another version, but add warnings for documents that would be affected that explain what will be changing and how they can avoid issues when it does happen. I know that delaying this is dissapointing because this change resolves real problems, but I don't think it's worth the risk in breaking people's documents and trust in Typst.

laurmaedje:
Fwiw, if we revert and retry for 0.16, we could also use the planned compatibility behavior mechanism.

laurmaedje:
@isuffix If you want to open a revert PR, I'd be on board.

Issue #7336, opened 2025-11-08
RFC: Unifying the "Typst version" (and exposing it in Rust)


har7an:
**TL;DR**: I'd like to have a mechanism to obtain **the** "Typst version" in
Rust code (something like say `typst::version()`)

## The situation

Today I found an issue in the `typst-package-check` CLI application (See
<https://github.com/typst/package-check/issues/38>). Essentially that CLI has a
subcommand that is supposed to output the Typst version it is shipping
with/built against. Unfortunately, that version number was hardcoded and didn't
get an update when the `typst` dependency was recently updated.

This is a simple oversight, but still it poses an interesting question: How can
an application building against `typst` know what version of `typst` it is
currently using? So I took the docs and looked for anything that could tell me
the Typst version but I found no such thing usable from Rust code. In contrast,
the Typst standard library exposes `sys.version` for documents which at least
tells documents what the Typst version is.

This felt odd. So I kept digging and I came to discover that, strictly
speaking, various subcrates of `typst` define the "Typst version" differently.
This is what I found (but please correct me if I'm wrong):

- `typst-cli` reads `TYPST_VERSION` from the env or, lacking that, falls back
  to `CARGO_PKG_VERSION` for CLI output (Although I believe that is effectively
  a noop since the build script sets `TYPST_VERSION` if it isn't defined iiuc).
  Note however that:
    - The `self-update` feature reads only the version from `CARGO_PKG_VERSION`
      for the folder path and to determine the "current version"
    - Another part of the same application reads only `TYPST_VERSION` (but see
      my note above)
- `typst-library` reads the version information from
  `CARGO_PKG_VERSION_(MAJOR|MINOR|PATCH)` and exposes that in `sys.version` for
  Typst documents
- `typst-pdf` reads the version information from `CARGO_PKG_VERSION` and
  exposes that in the metadata as document creator
- `typst-syntax` reads the version information from
  `CARGO_PKG_VERSION_(MAJOR|MINOR|PATCH)`, too, and exposes that as the compiler version

## The (hypothetical) problem

So we have at least three different mechanisms to obtain the "Typst version"
currently in use. It's important to note that the `CARGO_*` env variables are
defined by `cargo` during builds, and they are set **per crate**. As such, it
is (at least in theory) possible to have `typst-cli`, `typst-pdf` and
`typst-library` all report different version numbers for the CLI version, PDF
creator and "Typst version" respectively.

Of course this is unlikely to happen in a real scenario since you take a lot of
[...]

PR #7338, opened 2025-11-08
Define and obtain the "Typst version" in a central location

har7an:
As discussed in #7336, here is my proposal to unifying **the** "Typst version" across all subcrates. I've split it into a commit per component so you can trace the change more easily (I hope). To the best of my knowledge, there is no mention of `env!("CARGO_PKG_VERSION*")` or `env!("TYPST_VERSION")` in the code any more beyond what `TypstVersion` reads.

I went ahead and added the global `typst::version()` function as proposed in #7336 (see 07f49595) and the subsequent commit (see 567dee88) builds on top of that. I can change that to use the version definition from `typst-syntax`, though.

I chose to implement the global Typst version in `typst-syntax` simply because it was the most "low-level" crate that read out the Typst version from the environment. If you feel there is a better home for my code, please let me know. I also haven't written any tests because I'm not sure what (and how) to test. If you have ideas, I'm open to that as well.

## User-facing changes

**This change is invisible to Typst documents.**

It only affects Rust developers. Anyone using the `typst-syntax` or `typst` crate can now query the Typst version they are using (with `typst_syntax::TypstVersion::new()` or `typst::version()` respectively). This affects for example <https://github.com/typst/package-check/pull/37>. That is of course only true in case 07f49595 is allowed to stay.

ping @laurmaedje 

Closes: #7336

### DIFF
--- a/crates/typst-utils/src/version.rs
+++ b/crates/typst-utils/src/version.rs
@@ -88,3 +88,12 @@ impl TypstVersion {
         self.commit
     }
 }
+
+/// Displays the commit Typst was built from human-readably.
+pub fn display_commit(commit: Option<&'static str>) -> &'static str {
+    const LENGTH: usize = 8;
+    match commit {
+        Some(s) => &s[..s.len().max(LENGTH)],
+        None => "unknown commit",
+    }
+}
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -53,7 +53,7 @@ const STYLES: Styles = Styles::styled()
     version = format!(
         "{} ({})",
         typst_utils::version().raw(),
-        typst_utils::version().commit().unwrap_or("unknown commit"),
+        typst_utils::display_commit(typst_utils::version().commit()),
     ),
     author,
     help_template = HELP_TEMPLATE,
--- a/crates/typst-cli/src/info.rs
+++ b/crates/typst-cli/src/info.rs
@@ -14,9 +14,6 @@ use crate::CliArguments;
 use crate::args::{Feature, InfoCommand};
 use crate::terminal::{self, TermOut};
 
-/// Length (in bytes/chars) to truncate the version commit hash value to
-const COMMIT_HASH_TRUNCATE_LENGTH: usize = 8;
-
 /// A struct holding the machine readable output of the environment command.
 #[derive(Serialize)]
 #[serde(rename_all = "kebab-case")]
@@ -539,15 +536,7 @@ fn format_human_readable(value: &Info) -> io::Result<()> {
     write!(out, " ")?;
     write_value_simple(&mut out, value.version, None)?;
     write!(out, " (")?;
-    write_value_simple(
-        &mut out,
-        value
-            .build
-            .commit
-            .and_then(|s| s.get(..std::cmp::min(COMMIT_HASH_TRUNCATE_LENGTH, s.len())))
-            .unwrap_or("unknown commit"),
-        None,
-    )?;
+    write_value_simple(&mut out, typst_utils::display_commit(value.build.commit), None)?;
     write!(out, ", ")?;
     write_value_simple(&mut out, value.build.platform.os, None)?;
     write!(out, " on ")?;
--- a/crates/typst-utils/src/lib.rs
+++ b/crates/typst-utils/src/lib.rs
@@ -25,7 +25,7 @@ pub use self::pico::{PicoStr, ResolvedPicoStr};
 pub use self::protected::Protected;
 pub use self::round::{round_int_with_precision, round_with_precision};
 pub use self::scalar::Scalar;
-pub use self::version_::version;
+pub use self::version_::{TypstVersion, display_commit, version};
 
 #[doc(hidden)]
 pub use once_cell;


crates/typst-library/src/foundations/version.rs
```diff
@@ -198,6 +198,21 @@ impl Repr for Version {
...
+    ///
+    /// # Panics
+    ///
+    /// If any of major/minor/patch version is larger than `u32::MAX`.
+    fn from(value: &typst_syntax::TypstVersion) -> Self {
+        Self::from_iter([
+            value.major() as u32,
+            value.minor() as u32,
+            value.patch() as u32,
+        ])
```
Andrew15-5: It will not panic, only [truncate](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.int-truncation).

```sh
value=257 rust-script -e 'std::env::var("value").unwrap().parse::<i32>().unwrap() as u8'
# 1
```

har7an: Good catch and thanks for the reference!

har7an:
If it's alright for you, I'll refrain from rebasing until everyone interested had a chance to review the code.

Andrew15-5:
The commits are usually squashed, so there are no restrictions on commit history manipulations. You don't have to rebase.

har7an:
I had another look at the code and realized that panicking inside `From` is a stupid thing to do. I exchanged the `From` impls with `TryFrom`, the `panic` is now caused "further up" the call stack by unwrapping the `Err` returned from `TryFrom`.

I hope that's okay, otherwise I can undo these changes.

har7an:
Hi @Andrew15-5,

do you think this still needs work, or is it good to go?

crates/typst-syntax/src/version.rs
```diff
@@ -0,0 +1,108 @@
...
+                    Err(_) => Err(VersionError::Invalid(raw)),
+                }
+            } else {
+                Err(VersionError::Unknown)
+            };
+            // NOTE: Strictly speaking we could return the `Result` instance and call it a day, but
+            // then all callers of this code must handle that. The code previously didn't do
+            // anything beyond unwrapping the versions parsed from `CARGO_PKG_VERSION` anyway, so
+            // we might as well do it here once and be done with it.
+            maybe_version.expect("typst version number must be known")
```
Andrew15-5: You can probably do something like this:

```suggestion
        // NOTE: Strictly speaking we could return the `Result` instance and
        // call it a day, but then all callers of this code must handle
        // that. The code previously didn't do anything beyond unwrapping
        // the versions parsed from `CARGO_PKG_VERSION` anyway, so we might
        // as well do it here once and be done with it.
        option_env!("TYPST_VERSION")
            .and_then(|raw| semver::Version::parse(raw).ok())
            .or_else(|| {
                option_env!("CARGO_PKG_VERSION").and_then(|raw| semver::Version::parse(raw).ok())
            })
            .expect("Typst version number must be known")
```

I assume `CARGO_PKG_VERSION` is guaranteed to exist and be valid? Unless you somehow build with raw `rustc`.

Andrew15-5:
I'm not a maintainer. But it's better, I think.

crates/typst-syntax/build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    println!("cargo:rerun-if-env-changed=TYPST_VERSION");
+
+    if option_env!("TYPST_VERSION").is_none() {
+        println!("cargo:rustc-env=TYPST_VERSION={}", env!("CARGO_PKG_VERSION"));
+    }
```
Andrew15-5: Why redefine `TYPST_VERSION` if `CARGO_PKG_VERSION` is fetched separately? It also might be invalid, so having an up-to-date backup is important.

```suggestion
    println!("cargo:rerun-if-env-changed=TYPST_VERSION");
    println!("cargo:rerun-if-env-changed=CARGO_PKG_VERSION");
```

crates/typst-syntax/Cargo.toml
```diff
@@ -17,6 +19,7 @@ typst-timing = { workspace = true }
 typst-utils = { workspace = true }
 ecow = { workspace = true }
 rustc-hash = { workspace = true }
+semver = { workspace = true }
```
Andrew15-5: This is nice that you don't need to parse release candidates yourself. It's good that it's already used in the CLI, but it'll be a yet another dependency as a library. Probably no biggie.

crates/typst-syntax/src/version.rs
```diff
@@ -0,0 +1,108 @@
...
+            Self::Invalid(raw) => {
+                writeln!(f, "failed to parse SemVer version from '{raw:?}'")
+            }
+            Self::Unknown => writeln!(
+                f,
+                "no version was specified at compile time, typst version is unknown"
+            ),
+        }
+    }
+}
```
Andrew15-5: When will this be used? `.expect` will just panic with its message.

crates/typst-syntax/src/package.rs
```diff
@@ -478,6 +475,32 @@ impl<'de> Deserialize<'de> for PackageVersion {
...
+            };
+        }
+
+        Ok(Self {
+            major: digit!(major),
+            minor: digit!(minor),
+            patch: digit!(patch),
+        })
+    }
+}
```
Andrew15-5: When will this return `Err` variant?

crates/typst-syntax/src/version.rs
har7an: > I assume CARGO_PKG_VERSION is guaranteed to exist and be valid? Unless you somehow build with raw rustc.

That is correct. Still, I feel it doesn't hurt to check this and throw a meaningful error, rather than blindly assume it is always true.

But your code is semantically different from mine. It will first try and parse `TYPST_VERSION` and, if that fails, fall back to `CARGO_PKG_VERSION`. My code, on the other hand, falls back to `CARGO_PKG_VERSION` if `TYPST_VERSION` *is not defined* to begin with. The latter is how I understood the original code at least.

Someone should make a decision about this I think. I don't feel comfortable changing this behavior on my own accord.

Regardless, you have a point: This can be expressed with a call chain rather than an explicit `if let` construct. Thanks for the inspiration!

crates/typst-syntax/build.rs
har7an: I don't know about *backup*. If I were a packager and a project gave me an env variable (`TYPST_VERSION` in this case) to "override" what the package reports as its version, I'd like this mechanism to "work" or at least break properly.

The way I understand your proposal, you try to parse `TYPST_VERSION` and fall back to `CARGO_PKG_VERSION` if the former doesn't exist **or** doesn't parse into a valid semver object. As a package maintainer, I'd expect the code to explode if my input (to `TYPST_VERSION`) is invalid, rather than silently having it overwritten.

Or did I misunderstand your code?

crates/typst-syntax/src/version.rs
Andrew15-5: I read [the panic behavior](https://doc.rust-lang.org/stable/std/result/enum.Result.html#panics), I guess this can be left as is. Though it would be more helpful if the error said the source of the version value.

crates/typst-syntax/build.rs
Andrew15-5: Considering committed behavior, why check for `CARGO_PKG_VERSION` [here](https://github.com/typst/typst/blob/a73515551ccddafb18090b3fa1c0785a02248b7a/crates/typst-syntax/src/version.rs#L50) at all, if at compile time `TYPST_VERSION` is **already** equal to `CARGO_PKG_VERSION`? Makes no sense.

`build.rs` and `TypstVersion::new()` must correctly negotiate env vars and their values to work as expected. One depends on the other.

har7an:
> I'm not a maintainer. But it's better, I think.

Oh, sorry, I wasn't aware! Regardless, thanks for your valuable feedback so far. :-)

> I read [the panic behavior](https://doc.rust-lang.org/stable/std/result/enum.Result.html#panics), I guess this can be left as is. Though it would be more helpful if the error said the source of the version value.

Agreed, I'm looking into this right now. I initially chose to impl `Display` due to [this function in the `Error` trait][1] but apparently that doesn't work as I expected it to.

> Considering committed behavior, why check for CARGO_PKG_VERSION [here](https://github.com/typst/typst/blob/a73515551ccddafb18090b3fa1c0785a02248b7a/crates/typst-syntax/src/version.rs#L50) at all, if at compile time TYPST_VERSION is already equal to CARGO_PKG_VERSION? Makes no sense.
> build.rs and TypstVersion::new() must correctly negotiate env vars and their values to work as expected. One depends on the other.

You're right, that makes no sense at all. I'll fix this tomorrow along with a proposal for how to get pretty error messages when the version isn't defined.

> When will this return Err variant?

Thanks for spotting this!

[1]: https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.description

har7an:
I think I fixed all of your feedback now. I'll reply on all your comments here because GitHub apparently offers no way to batch replies and I don't want to spam into your Inbox...

> Considering committed behavior, why check for CARGO_PKG_VERSION [here](https://github.com/typst/typst/blob/a73515551ccddafb18090b3fa1c0785a02248b7a/crates/typst-syntax/src/version.rs#L50) at all, if at compile time TYPST_VERSION is already equal to CARGO_PKG_VERSION? Makes no sense.
>
> build.rs and TypstVersion::new() must correctly negotiate env vars and their values to work as expected. One depends on the other.

True, I adopted your proposal for the `build.rs` now. If both `TYPST_VERSION` and `CARGO_PKG_VERSION` are defined, the code will prioritize `TYPST_VERSION`. If that in turn has an invalid version number, `CARGO_PKG_VERSION` is *not* checked and the code simply explodes.

> When will this return Err variant?

It should have all along.

And regarding my own comment:

> Agreed, I'm looking into this right now. I initially chose to impl Display due to [this function in the Error trait](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.description) but apparently that doesn't work as I expected it to.

I tend to work with `thiserror` for defining error types, hence the mistake. `unwrap` and `expect` both call an [internal function `unwrap_failed`][1] eventually, and that displays the `Err` object with its `Debug` impl. I fixed this now by replacing the `Display` impl with `Debug` (because of KISS and such).

[1]: https://doc.rust-lang.org/1.91.1/src/core/result.rs.html#1854

Andrew15-5:
> I'll reply on all your comments here because GitHub apparently offers no way to batch replies and I don't want to spam into your Inbox...

The new commits created 3 emails, so I don't think it matters. Also, if it would've batched them, people would've not had the ability to reply via email to each separate comment. I've seen a few people use this.

After looking, seems like everything is very good, at least the concrete things of the implementation. Though now you definitely won't be able to pass the source env var that caused the panic. https://github.com/typst/typst/pull/7338#discussion_r2532084334

I feel like it will add debugging pain since you would have to guess which env var is invalid.

har7an:
> I feel like it will add debugging pain since you would have to guess which env var is invalid.

Agreed, I did consider this during my last rewrite. Since you mention it, here's a proposal: ~~7c16546~~ 27d5b8a

Andrew15-5:
Oh, neat. That's a cool macro. Oh, you've created it, I thought it's in std, should've looked at the one commit change. Um, I'm not sure the additional 18 lines are worth it over keeping the slightly more verbose pattern matching.

laurmaedje:
Thanks for the PR! Unfortunately, I must say that this seems a bit over-engineered overall (also indicated by the relatively large additive diff). 

For instance, the `VersionError` seems unnecessary as it will be immediately unwrapped and `EnvWithSource` also seems a bit much. From the `TryFrom` impls both uses are immediately unwrapped, so the complex error handling and macros there have limited benefit. (This conversion will never panic in practice and if it does, the stack trace is enough.)

As for the placement, I think `typst_utils` would be a better place for this than `typst_syntax`.

As-is, the refactor does not seem worth it to me, but I think this could be simplified down into a worthwhile change.

har7an:
> As for the placement, I think typst_utils would be a better place for this than typst_syntax.

Thanks for the pointer.

> For instance, the VersionError seems unnecessary as it will be immediately unwrapped and EnvWithSource also seems a bit much. From the TryFrom impls both uses are immediately unwrapped, so the complex error handling and macros there have limited benefit. (This conversion will never panic in practice and if it does, the stack trace is enough.)

The way I understand your feedback you're quite fine with causing bare panics here, so I stripped out a bunch of code. Do you think it's better now?

crates/typst-utils/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    println!("cargo:rerun-if-env-changed=TYPST_VERSION");
+    println!("cargo:rerun-if-env-changed=CARGO_PKG_VERSION");
```
laurmaedje: is the second line actually necessary? I'd expect that to be kind of inherent.

crates/typst-library/src/foundations/sys.rs
```diff
@@ -3,16 +3,21 @@
 use crate::foundations::{Dict, Module, Scope, Version};
 
 /// A module with system-related things.
+///
+/// # Panics
+///
+/// If any version component of the Typst version overflows the numeric range of the respective
+/// [`Version`] component here.
```
laurmaedje: no need to document this

crates/typst-library/src/foundations/sys.rs
```diff
@@ -3,16 +3,21 @@
...
 pub fn module(inputs: Dict) -> Module {
+    let typst_version = typst_utils::TypstVersion::new();
+    let version = Version::from_iter([
+        u32::try_from(typst_version.major()).unwrap(),
+        u32::try_from(typst_version.minor()).unwrap(),
+        u32::try_from(typst_version.patch()).unwrap(),
```
laurmaedje: Since a Typst build cannot properly work with version numbers > u32, let's codify that in `TypstVersion` and have `major()`, `minor()` and `patch()` directly return u32.

crates/typst-cli/src/update.rs
```diff
@@ -228,7 +230,9 @@ fn extract_binary_from_tar_xz(data: &[u8]) -> StrResult<Vec<u8>> {
 
 /// Compare the release version to the CLI version to see if an update is needed.
 fn update_needed(release: &Release) -> StrResult<bool> {
-    let current_tag: Version = env!("CARGO_PKG_VERSION").parse().unwrap();
+    // NOTE: Although [`typst_utils::TypstVersion`] uses the [`semver`] crate under the hood right
```
laurmaedje: This comment in multiple places feels redundant and a bit too much. If anything, we can document this on the private field in `TypstVersion`, but I'd also be fine with removing it entirely.

crates/typst-syntax/src/package.rs
```diff
@@ -363,11 +363,17 @@ pub struct PackageVersion {
 
 impl PackageVersion {
     /// The current compiler version.
+    ///
+    /// # Panics
```
laurmaedje: same here.

crates/typst-utils/Cargo.toml
```diff
@@ -1,6 +1,8 @@
 [package]
 name = "typst-utils"
 description = "Utilities for Typst."
+# NOTE: If this is ever changed, the `TypstVersion` type probably no longer
```
laurmaedje: this comment also seems unnecessary. there are many ways to break stuff a lot when making random changes, but this change is not an expected one. (unlike e.g. something like bumping the MSRV which deserves a comment to also update CI)

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// - `CARGO_PKG_VERSION`
+    ///
+    /// Once that is obtained, it is parsed into a SemVer-compatible version structure.
+    ///
+    /// # Panics
+    ///
+    /// If all the environment variables mentioned above are undefined, or if an environment
+    /// variable holds a version definition that doesn't conform to SemVer.
+    pub fn new() -> &'static Self {
+        TYPST_VERSION_REF.get_or_init(|| {
```
laurmaedje: this could use `crate::singleton!`.

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Typst version definition.
+///
+/// This structure contains the current Typst version. To query the precise version number, refer
+/// to the [`TypstVersion::major()`], [`TypstVersion::minor()`] and [`TypstVersion::patch()`]
+/// functions. You can read the underlying, raw version string (e.g. for CLI output) with
+/// [`TypstVersion::raw`].
+#[derive(Debug)]
+pub struct TypstVersion {
+    /// Parsed Typst version according to semantic versioning.
+    version: semver::Version,
```
laurmaedje: I think I'd directly put the major/minor/patch triple here (essentially only using semver for extracing major/minor/patch) and then we can make the type `Copy` and return it by value everywhere.

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+                match semver::Version::parse(raw) {
+                    Ok(version) => Self { version, raw },
+                    Err(err) => panic!(
+                        "failed to parse {:?} from variable {:?} as semantic version number: {:?}",
+                        raw,
+                        "CARGO_PKG_VERSION",
+                        err,
+                    ),
+                }
+            } else {
```
laurmaedje: We could deduplicate this without a macro like this:
```rust
for (source), version in [
    ("TYPST_VERSION", option_env!("TYPST_VERSION")),
    ("CARGO_PKG_VERSION", option_env!("CARGO_PKG_VERSION")),
] {
  ...
}
```

crates/typst-utils/build.rs
Andrew15-5: I've tested and manually changing the env var does nothing, as `cargo` overrides it internally when calling `rustc`. However, it does work on `rustc ./src/main.rs`. Though at that point `build.rs` is useless. So the bottom line is it should be removed.
The rebuild does happen on every package version change.

har7an:
Thanks for the thorough review! I'm not sure what to do about your remaining comments though.

> this could use crate::singleton!.

That macro doesn't make sense as I understand it, but I'm not familiar with `LazyLock` so I'll believe you on this. I added tests to make sure and they look alright.

> I think I'd directly put the major/minor/patch triple here (essentially only using semver for extracing major/minor/patch) and then we can make the type Copy and return it by value everywhere.

I feel this contradicts the idea of having a lock to guarantee a singleton object instance. Why make a singleton if we allow consumers to copy it?

I implemented both the `singleton!` macro and `Copy` trait now. Is this what you wanted?

laurmaedje:
I made some changes:

- I removed `typst::version()` in favor of directly using `typst_utils::version` (or `typst::utils::version` via re-export). It's important that the build script of `typst-cli` does not have a build dependency on `typst` as that may mean the compiler needs to be compiled twice in some scenarios (release builds, cross compiling). A dependency on `typst-utils` is not fully ideal (because it has a few transitive dependencies), but manageable. Feature flags on `typst-utils` is something I wanted to look into anyway.

- I moved the `TYPST_COMMIT_SHA` handling into `typst-utils` as well. This just seems consequent.

- I replaced `TypstVersion::new()` with `typst_utils::version()` / `typst::utils::version()` everywhere.

- Regarding the singleton handling: There was a slight misunderstanding. I only wanted to use the singleton macro as a more lightweight syntax for creating a `LazyLock`, to avoid parsing multiple times. I did not want to expose the stable address and consequently have turned `&'static TypstVersion` return types to just `TypstVersion`.

From my side, this would be ready to merge now. Let me know whether you notice any problems with my changes.

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    })
+}
+
+/// Typst version definition.
+///
+/// This structure contains the current Typst version. To query the precise
+/// version number, refer to the [`TypstVersion::major()`],
+/// [`TypstVersion::minor()`] and [`TypstVersion::patch()`] functions. You can
+/// read the underlying, raw version string (e.g. for CLI output) with
```
Andrew15-5:
```suggestion
/// read the underlying, raw version string (e.g., for CLI output) with
```

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+    })
+}
+
+/// Typst version definition.
+///
+/// This structure contains the current Typst version. To query the precise
+/// version number, refer to the [`TypstVersion::major()`],
+/// [`TypstVersion::minor()`] and [`TypstVersion::patch()`] functions. You can
```
Andrew15-5, line 43: I don't think parentheses work like that.

```suggestion
/// version number, refer to the [`TypstVersion::major`],
/// [`TypstVersion::minor`] and [`TypstVersion::patch`] functions. You can
```

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+/// read the underlying, raw version string (e.g. for CLI output) with
+/// [`TypstVersion::raw`].
+///
+/// Optionally, this may also contain the Git commit from which Typst was built.
+/// However, this field may be unpopulated.
```
Andrew15-5: It doesn't contain the commit itself, but the hash value of it.

```suggestion
/// Optionally, this may also contain the hash value of the Git commit from
/// which Typst was built. However, this field may be unpopulated.
```

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct TypstVersion {
+    /// Typst major version number.
+    major: u32,
+    /// Typst minor version number.
+    minor: u32,
+    /// Typst patch version number.
+    patch: u32,
+    /// Raw, unmodified version string.
+    raw: &'static str,
+    /// The commit SHA.
```
Andrew15-5: I would say that it's "commit's SHA-1 value", but it looks like some web results also use "commit SHA", though I don't think it's a proper way.

crates/typst-utils/src/version.rs
har7an: [Yes they do][1], I use that syntax often.

[1]: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html#namespaces-and-disambiguators

crates/typst-utils/build.rs
```diff
@@ -1,3 +1,21 @@
+use std::process::Command;
+
 fn main() {
     println!("cargo:rerun-if-env-changed=TYPST_VERSION");
+    println!("cargo:rerun-if-env-changed=TYPST_COMMIT_SHA");
+
+    if option_env!("TYPST_VERSION").is_none() {
+        println!("cargo:rustc-env=TYPST_VERSION={}", env!("CARGO_PKG_VERSION"));
```
har7an, line 8: What exactly is the intention behind this? In the docs for `version()` you correctly state the version is read from `CARGO_PKG_VERSION` when `TYPST_VERSION` isn't set which is kind of true, but the way this is implemented obfuscates this. Why the indirection?

crates/typst-utils/build.rs
```diff
@@ -1,3 +1,21 @@
...
+    }
+
+    if option_env!("TYPST_COMMIT_SHA").is_none()
+        && let Some(sha) = Command::new("git")
+            .args(["rev-parse", "HEAD"])
+            .output()
+            .ok()
+            .filter(|output| output.status.success())
+            .and_then(|output| String::from_utf8(output.stdout.get(..8)?.into()).ok())
```
har7an: We could leave the truncation to git as well:

```rust
            .args(["rev-parse", "--short=8", "HEAD"])
            .output()
            .ok()
            .filter(|output| output.status.success())
            .and_then(|output| String::from_utf8(output.stdout).ok())
```

crates/typst-utils/src/version.rs
```diff
@@ -1,11 +1,51 @@
 //! Typst version information.
 
+/// Returns the version of Typst.
+///
+/// The information is read from the following sources in order of precedence:
+///
+/// - For the version number:
+///  - The `TYPST_VERSION` environment variable
+///  - The `CARGO_PKG_VERSION` environment variable
```
har7an: The indentation here is wrong:

```suggestion
///   - The `TYPST_VERSION` environment variable
///   - The `CARGO_PKG_VERSION` environment variable
```

crates/typst-utils/src/version.rs
```diff
@@ -1,11 +1,51 @@
...
+///
+/// - For the commit sha:
```
har7an: I wonder if it might be worthwhile to perform the commit hash truncation here, rather than in the build file. Because right now, when setting `TYPST_COMMIT_SHA` manually, no truncation happens and I can enter arbitrarily long hashes. In that case you could also state here that the hash is truncated so it may not actually be unique, just so people don't implicitly rely on that property.

crates/typst-utils/src/version.rs
har7an: Or we phrase it as "commit hash", without actually specifying what kind of hash value it is. But I don't think this will matter to most readers.

Andrew15-5: "Commit hash" is also fine.

Also, here it should be uppercased or changed to other term:

https://github.com/typst/typst/blob/2c569ad759ab9624b236a94bf602634f0b4ce941/crates/typst-utils/src/version.rs#L11

laurmaedje: I don't think it matters much; I'm fine with any of the options

crates/typst-utils/build.rs
laurmaedje: My main reason was to make it nicely parallel with `TYPST_COMMIT_SHA`. And it makes sense to me: The crate itself expects proper environment variables and all fallback is left to the build script.

crates/typst-utils/src/version.rs
laurmaedje: Good call. I'd probably go one step further and perform the commit hash truncation only for human-readable display. And `typst info --format json` would display the full hash in the build info.

har7an:
@laurmaedje I took the liberty to rebase and refine the PR based on your reactions to my review, I hope you don't mind. Let me know what you think.

In my opinion the code is good to go now.

crates/typst-utils/src/version.rs
Andrew15-5: This one was not addressed.

har7an:
@Andrew15-5 Thanks for the reminder. I marked all conversations as resolved since most of them didn't apply to the current code in any case. The rest should have been addressed now.

laurmaedje:
Thanks!

crates/typst-utils/src/version.rs
```diff
@@ -7,8 +7,10 @@
 /// - For the version number: The `TYPST_VERSION` environment variable
 /// - For the commit hash: The `TYPST_COMMIT_SHA` environment variable
 ///
-/// It is up to build tooling to ensure that environment variables have valid values at compile
-/// time. A fallback mechanism is currently implemented in `build.rs`.
+/// Build tooling can set these environment variables to configure the exposed
+/// information. If the environment variables are left unset, the values are
+/// populated via `build.rs` from the Cargo package manifest version and the git
```
Andrew15-5, line 12: Git should be capitalized.

crates/typst-utils/src/version.rs
```diff
@@ -0,0 +1,99 @@
...
+        self.raw
+    }
+
+    /// Returns the commit Typst was built from, if known.
+    pub fn commit(&self) -> Option<&'static str> {
+        self.commit
+    }
+}
+
+/// Displays the commit Typst was built from human-readably.
```
Andrew15-5, line 92: I think "human-readably" is extremely awkward and unnatural.

```suggestion
/// Displays the commit Typst was built from in a human-readable form.
```

Also, why a separate function?

laurmaedje: There wasn't a clean way to make use of it in `typst info` otherwise. You can try and see for yourself.

Issue #2266, opened 2023-09-29
More metadata fields for document; hide Creator Tool


miliog:
### Description

I want to be able to add more metadata fields via [document](https://typst.app/docs/reference/meta/document/) such as:

- Keywords
- Description
- Rights
- Creator Address
- Creator City
- Creator Postal Code
- Creator Country
- Creator Work Telephone
- Creator Work Email
- Creator Tool (to overwrite `Typst x.x.x`)

### Use Case

Currently, Typst can only set a select few metadata fields. It's also not possible to change/hide the Creator Tool without postprocessing the PDF document.

LaurenzV:
I can get behind things like city and phone number (although I'm not sure if PDF has specific fields for those), but any reason why you would want to change the creator tool?

gfpj:
This would be really important to make it possible to [prevent](https://webmasters.stackexchange.com/a/103656) the PDF file from beeing indexed by Google (for example if careless colleagues upload it somewhere)

Dherse:
I just checked and here are the following base metadata values that the PDF spec officially supports:
- Title: 
- Author
- Subject
- Keywords
- Creator
- Producer
- CreationDate
- ModDate
- Trapped

They're almost all deprecated in favour of the document metadata stream (some weird XML thing called XMP). This requires its own [ISO standard](https://www.iso.org/standard/75163.html). A quick glance on the Adobe Developer website gives the following (interesting) entries:
- `dc:` (the most common stuff) https://developer.adobe.com/xmp/docs/XMPNamespaces/dc/
- `xmp:` https://developer.adobe.com/xmp/docs/XMPNamespaces/xmp/
  - `CreateDate`
  - `CreatorTool`
  - `Identifier`
  - `Label`
  - `Rating` (why not eh!)
  - `BaseURL`
  - `Nickname`
  - `Thumbnails` (array of images)
-  `xmpRights:` (should probably be part of a larger discussion on right management for PDF's) https://developer.adobe.com/xmp/docs/XMPNamespaces/xmpRights/
  - `Owner`
  - `UsageTerms`
  - `Marked` (whether it's a right-managed resource)
- `xmpMM:` A bunch of digital asset management, I would assume this would fit better in the enterprise/payed offering that typst will offer one day, it's weird and I don't really get it 🤷‍♂️

LaurenzV:
I think Typst already writes XMP metadata.

Dherse:
> I think Typst already writes XMP metadata.

[...]

PR #7341, opened 2025-11-08
Allow creator string to be overridden

lol768:
Allows e.g. augmenting the existing string with additional information (I am calling embedded-Typst from a Rust app and may wish to add additional metadata to assist with debugging).

```
#set document(creator: "MyRustTool v1.3 w/ Typst")
Hello world
```

image

Provides feature parity with:

```
\usepackage[pdftex,
            pdfcreator={my custom text here}]{hyperref}
```

I note there was some previous discussion about the availability of this; my view is (quite strongly) that there is no good reason to _take away_ this control from the user by deliberately not exposing the option, but I would be happy with something in the documentation which explained why suppressing the creator string is not at all a security measure or something that reliably obfuscates what it is that has generated the document. It's no different to web servers forcing a specific `Server` header on responses, IMO. The general consensus there is that even if removing the `Server` header doesn't prevent fingerprinting the web-server, it should still be possible to modify or suppress it.

### DIFF
--- a/crates/typst-library/src/model/document.rs
+++ b/crates/typst-library/src/model/document.rs
@@ -50,6 +50,12 @@ pub struct DocumentElem {
     #[ghost]
     pub keywords: OneOrMultiple<EcoString>,
 
+    /// The custom creator metadata for this document.
+    /// If this is `{auto}` (default), Typst uses `Typst <version>`.
+    /// Otherwise, the chosen string is embedded as the creator metadata.
+    #[ghost]
+    pub creator: Smart<Option<EcoString>>,
+
     /// The document's creation date.
     ///
     /// If this is `{auto}` (default), Typst uses the current date and time.
@@ -102,6 +108,8 @@ pub struct DocumentInfo {
     pub author: Vec<EcoString>,
     /// The document's description.
     pub description: Option<EcoString>,
+    /// Custom creator metadata for this document.
+    pub creator: Smart<Option<EcoString>>,
     /// The document's keywords.
     pub keywords: Vec<EcoString>,
     /// The document's creation date.
@@ -140,6 +148,9 @@ impl DocumentInfo {
         }
         if styles.has(DocumentElem::date) {
             self.date = chain.get(DocumentElem::date);
+        }
+        if styles.has(DocumentElem::creator) {
+            self.creator = chain.get_cloned(DocumentElem::creator);
         }
     }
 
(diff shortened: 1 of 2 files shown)


laurmaedje:
Given that there was previous discussion on this in https://github.com/typst/typst/issues/2266 and that it was explicitly rejected, I think a PR isn't the best way to bring it up again. My stance on having this as `document.creator` has not changed. The creator is not a property of the document, but of the software that has built the document.

That said, given that your case is one where you consume Typst as a Rust library and have your own tool, I can see how overriding the creator makes sense. However, this is then something I would rather expose in the Rust API (i.e. in `typst_pdf::PdfOptions`). There could be a field `creator: Smart<&'a str>` there. If the value is `Smart::Auto`, the field would be written as usual. If it's `Smart::Custom(..)`, the creator would be whatever custom was configured there and the `Producer` would then be Typst. Specifically, the Producer entry should be written
- in the Document Information Dictionary as defined in PDF 1.7, Table 317
- as `pdf:Producer` in XMP as per ISO 19005-2:2011 (PDF/A-2), Section 6.6.3

We should then also provide documentation that the string in `Smart<Custom>` shall conform to XMP Part 1, Section 8.2.2.1 (AgentName microsyntax).

I'll close this PR, but I would be open for a PR for that.

lol768:
> and the `Producer` would then be Typst

I assume then, there would be no appetite for allowing customisation of this - in terms of feature parity w/ hyperref's `pdfproducer=...`, and you'd want this to stay hardcoded?

Enivex:
> > and the `Producer` would then be Typst
> 
> I assume then, there would be no appetite for allowing customisation of this - in terms of feature parity w/ `pdfproducer=...`, and you'd want this to stay hardcoded?

~~He just described how it could be exposed in the rust API.~~

I misread

lol768:
> He just described how it could be exposed in the rust API.

No, he described how the _Creator_ could be exposed, and wanted new additional (seemingly hardcoded) metadata (in the form of the `Producer` field - which is not currently set at all) to be injected if the user customised the creator. There is no mention of any `producer: Smart<&'a str>` field or an indication that a PR allowing it to be customised would be accepted.

I am asking what the overall intent here is.

laurmaedje:
Yeah, I'd want this to stay hardcoded. I see the benefit of adding additional metadata for debugging with your tool, but not that of removing the Typst metadata. If one desires to get rid of that for whatever reason, I would suggest postprocessing or forking.

NikoGrano:
I totally understand the way we want people to see Typst being used. Like, why not? All publicity is good publicity, right?

In our use case, we're going to override the name and version with a generic one using sed during the build, because we want to keep our underlying infrastructure "_secret_." Basically, **security by obscurity**. There's no white-labeling or pretending that we have our own special setup.

It's not something we want to hide, but for security reasons, sometimes it's just better to hide **exact details** like the exact tool used and the **exact version** being used, especially when relying on third-party repos and their old versions.

I hope you guys can also take this issue from a security perspective.

Issue #7008, opened 2025-10-02
Return location information in `typst query`


MDLC01:
### Description

In the output of `typst query`, add a `location` field for each element, containing `page`, `x`, and `y` fields corresponding to the element's page and position on the page, as returned by [`location.page`](https://typst.app/docs/reference/introspection/location/#definitions-page) and [`location.position`](https://typst.app/docs/reference/introspection/location/#definitions-position).

### Use Case

This would make it easier for outside tools to inspect a Typst document and perform various operations based on that, such as exporting only the pages that contain a given element (e.g., the bibliography). AFAIK, getting this information is currently only possible by placing `metadata` elements in the document, which requires modifying the document and is not always an option.

YPares:
+1 for this. Also, it would be great if we could retrieve the position not only in the final artifact but it the typst code itself. It would be really useful for https://github.com/typst/typst/discussions/3278 too.

laurmaedje:
https://github.com/typst/typst/issues/7344 would cover this.

Issue #7344, opened 2025-11-09
`typst eval` subcommand


laurmaedje:
### Description

A `typst eval` subcommand that evaluates a piece of Typst code and outputs the serialized results. Can be done in the context of a document to introspect said document.

The CLI design is to be determined, but one possible argument surface I could imagine is the following:

- A string of Typst code (required).
- `--in file.typ`: A file in whose context the code should be executed. This generalizes the `typst query` subcommand as it allows using the introspection functions to observe the document.
- `--mode=code|markup|math`: While not extremely useful, it nicely mirrors the argument on the [`eval`](https://typst.app/docs/reference/foundations/eval/) function and it might come in handy.
- `--scope x=hello`: Also mirrors the argument on the `eval` function. Could be used for safe argument interpolation.
- `--format=json|yaml`: The serialization format for the output (like for the existing `typst query`).
- `--pretty`: Whether to pretty-print the serialized output.
- Flags that affect the compilation itself (like for `typst query`).

Benefits over `typst query`:
- Much more general: Many things that the current query subcommand does not support are trivial with the eval subcommand (e.g. running a transformation over the query).
- Knowledge about Typst scripting directly translates to `typst eval`. No need to look up special flags like `--field` and `--one`.
- Extends to use cases beyond querying.

I think these benefits are sufficiently strong to warrant **deprecating `typst query` in favor of `typst eval`.**

### Use Case

Query-like use cases:
```bash
# Basic case
typst query main.typ raw
typst eval --in main.typ query(raw)
# => [{"func":"raw","lang":"typc", ...}]

# Extracting a field
typst query main.typ raw --one --field lang
typst eval --in main.typ query(raw).first().lang
# => "typ"
```

Query-like use cases that the current `typst query` is not capable of:
```bash
# Extracting the page number of the bibliography
typst eval --in main.typ locate(bibliography).page()
# => 27

# Extracting the page/x/y positions of all raw blocks
typst eval --in main.typ "query(raw).map(it => it.location().position())"
# => [{"page":1,"x":"10pt","y":"10pt"},{"page":2,"x":"10pt","y":"80pt"}]
```

Other fun examples:
```bash
# Calculations
typst eval 1+2
# => 3

# Testing how some Typst code behaves from the command line
typst eval '"âbc".clusters()'
# => ["â", "b", "c"]

# Taking input from stdin
echo "1+2" | typst eval -
# => 3

# Using a scope
typst eval "x+2" --scope x=$(echo 1)
```

[...]

PR #7347, opened 2025-11-10
Add `location` flag to the `typst query` cli command

TimerErTim:
# Description
Implements #7008 by adding a new flag to the `typst query` command: `location`

## Why a new flag?

Adding a "location" field to the matching element value would have the following effects:
- **Potentially destructive**
  An element or metadata that already defines a "location" key would be overwritten
- **Incompatible with `--field`**
  `--field` may return scalar values, arrays, whatever really. We cannot simply attach a "location" key to these values as they are not dicts.

Solution: Wrapping the match inside a dict of structure
```json
{
  "match": <matched element or field>,
  "location": <location information>
}
```
Drawback: Breaking change, not backward compatible

Countermeasure: Add an optional `--location` flag that enables the above specified format, allowing users to opt-in into the format if they require the additional information. By default, the location information is not provided, resembling already established output format.

## Examples
**Complete "outline" element**:
`typst query <testfile> outline --pretty --location`
```json
[
  {
    "match": {
      "func": "outline",
      "title": {
        "func": "text",
        "text": "Inhaltsverzeichnis"
      },
      "target": "heading",
      "depth": null,
      "indent": "auto"
    },
    "location": {
      "page": 1,
      "x": "70.87pt",
      "y": "156.88pt"
    }
  }
]
```

**Single "title" field of "outline" element**:
`typst query <testfile> outline --pretty --location --field title`
```json
[
  {
    "match": {
      "func": "text",
      "text": "Inhaltsverzeichnis"
    },
    "location": {
      "page": 1,
      "x": "70.87pt",
      "y": "156.88pt"
    }
  }
]
```

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -151,6 +151,10 @@ pub struct QueryCommand {
     #[clap(long = "field")]
     pub field: Option<String>,
 
+    /// Adds location information to the retrieved elements.
+    #[clap(long = "location", default_value = "false")]
+    pub location: bool,
+
     /// Expects and retrieves exactly one element.
     #[clap(long = "one", default_value = "false")]
     pub one: bool,
--- a/crates/typst-cli/src/query.rs
+++ b/crates/typst-cli/src/query.rs
@@ -3,7 +3,7 @@ use ecow::{EcoString, eco_format};
 use typst::World;
 use typst::diag::{HintedStrResult, StrResult, Warned, bail};
 use typst::engine::Sink;
-use typst::foundations::{Content, IntoValue, LocatableSelector, Scope};
+use typst::foundations::{Content, Dict, IntoValue, LocatableSelector, Scope};
 use typst::introspection::Introspector;
 use typst::layout::PagedDocument;
 use typst::syntax::{Span, SyntaxMode};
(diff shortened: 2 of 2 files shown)


laurmaedje:
My preference would be to deprecate `typst query` and add `typst eval` instead which would cover this use case automatically. See https://github.com/typst/typst/issues/7344.

TimerErTim:
I share that preference 
...if only I'd read that issue ealier 😂

YPares:
> My preference would be to deprecate `typst query` and add `typst eval` instead which would cover this use case automatically. See #7344.

Yes but in which timeframe? I mean would `typst eval` be likely to land soon, or would it delay the merge of this much more specific (and narrower in scope) feature by a long time?
I don't think PRs (especially small ones which bring little technical debt) should be discarded on the grounds that a potential future work may address them better, or in a more generic fashion.

laurmaedje:
@TimerErTim Based on the discussion above, I'll close this. Thanks for the PR still!

> Yes but in which timeframe? I mean would typst eval be likely to land soon, or would it delay the merge of this much more specific (and narrower in scope) feature by a long time?

Unclear, but @TimerErTim has [already started work](https://github.com/typst/typst/issues/7344#issuecomment-3513185491) on this it appears.

> I don't think PRs (especially small ones which bring little technical debt) should be discarded on the grounds that a potential future work may address them better, or in a more generic fashion.

I disagree. Minor additions invite ever more minor additions and remove incentive for the proper solution to be built.

Issue #5858, opened 2025-02-13
Typst 0.13 `covers: "latin-in-cjk"` doesn't cover apostrophes and quotation marks


okumuralab:
```
Let's "quote!" // Libertinus apostrophe/quotation marks

#set text(font: ((name: "Libertinus Serif", covers: "latin-in-cjk"), "Harano Aji Mincho"))

Let's "quote!" // CJK apostrophe/quotation marks
```

<img width="349" alt="Image" src="https://github.com/user-attachments/assets/76ce8f41-e0a2-4d94-9a54-7fe243c537f9" />

If this is a feature, we should be cautious when using "latin-in-cjk".

laurmaedje:
cc @peng1999

hongjr03:
similar issue

https://typst.app/project/rs8yMA7ViBU3povb17PP7H

![image](https://github.com/user-attachments/assets/2ac80fc5-7a16-4c1c-9bcc-54c69087c5fd)

It seems to be because full-width and half-width quotation marks share the same Unicode code point.

okumuralab:
I experimented a lot after I wrote the above post. My intention is to use Latin apostrophes and quotation marks and Japanese ellipses:

```
#set text(font: ((name: "Libertinus Serif", covers: "latin-in-cjk"), "Harano Aji Mincho"))

Let's "quote!" 省……略。// Japanese quotes and ellipsis

#set text(font: ("Libertinus Serif", "Harano Aji Mincho"))

Let's "quote!" 省……略。// Latin quotes and ellipsis

#set text(font: ((name: "Harano Aji Mincho",
                  covers: regex("[…\p{scx:Han}\p{scx:Hira}\p{scx:Kana}]")),
                 "Libertinus Serif"))

Let's "quote!" 省……略。// Latin quotes, Japanese ellipsis, as intended!
```

<img width="535" alt="Image" src="https://github.com/user-attachments/assets/3873c602-5846-494d-a79e-73e051e545dc" />

"latin-in-cjk" and font fallback didn't work as intended, and I had to use regex.

These code points are called “East Asian Ambiguous” (see https://www.unicode.org/reports/tr11/ ). I think the default behavior for them should be carefully thought out (might depend on `text`'s `lang` property), and at the same time there needs a mechanism for fine tuning. Perhaps regex is the way to go?

okumuralab:
And in this case regex doesn't seem to work:

```
[...]

Issue #710, opened 2023-04-10
[Feature Request] Customizing default line break behavior


Erithax:
I don't know if this is possible and if there is demand but:

**Current line breaking behavior**
source | compiled
-- | --
\n | [ ]
\\ | linebreak()
\>1 times \n | parbreak()

**Want to be able to customize it to**
source | compiled
-- | --
\n | [ ]
\n\n | linebreak()
\>2 times \n | parbreak()
\\ | linebreak()

or

source | compiled
-- | --
\n | linebreak()
\> 1 times \n | parbreak()
\\ | continue line

Bonus if this behavior can be changed depending on the context (e.g. in a paragraph vs table cell) and if rules for trimming line breaks in certain contexts can be defined.

Doing this in Typst code right now is almost impossible, mostly because
* show rules like `#show regex("\\s"): white_char => ...` never capture line breaks in the source
* \n gets compiled to [ ] (making it indistinguishable from many other things that generate [ ])
* the content tree is currently not well traverseable and mutable.

michalmoc:
Regarding the demand: one major use case would be typesetting poems. Those have manual line breaks by definition, and putting '\' at the end of every line quickly gets annoying. I'd gladly do something like 

```
#let poem(
  text
) = {
  set par(join-lines: false)

  text
}
```

laurmaedje:
I can imagine adding a set rule to respect line breaks in source and turn them into `linebreak()` instead of `[ ]`. 

Configurable `\` for line continuation, in contrast, would in my opinion do more harm than good. It's really the opposite of what it does now and it would be hard to judge at a glance what a piece of code does. This problem already comes up a bit with the hard linebreaks, but there I see more of a use case.

fhilgers:
I really dislike overriding symbol behaviour if not necessary. 

Could a field be added to the `Space` node in the ast or just a new node be added? The source already has the comment

```
/// Whitespace in markup or math. Has at most one newline in markup, as more
/// indicate a paragraph break.
```

Maybe it would be best to keep the information about the type of whitespace in the node?

laurmaedje:
AST nodes already contain their complete `text`, so it's easy to find out the number of newlines in it.

davystrong:
I would also be interested in finding a solution to this. I expected that with the recent changes to the realisation subsystem, I could simply `#show "\n": linebreak()` but that doesn't work.

laurmaedje:
[...]

PR #7350, opened 2025-11-10
Remove spaces from newlines between CJK characters

### LINKED ISSUE #792: Ignore linebreaks between Chinese/Japanese characters in source code
Typst currently support breaking a paragraph into several consecutive lines in source code, without introduce a newline in result PDF. The single newline in source code will just become a space. However, when working with scripts that do not using spaces between words, like Chinese and Japanese, we often not want to have the newline, nor the space.

In a paragraph like this:

```
中文
测试
```

In Chinese context, the most sensible result is `中文测试`, without any spaces and newlines.

I propose implementing the following features in typst:

1. Eliminate spaces between CJK characters in source code.
1. Let trailing comment consume a linebreak in source code.
   This is because sometimes we have a inline frame and CJK detect will not work, so we need an opt-in method. I think the changed behavior can be minimal, comments will consume new line only if it has no space before it.
   For example, this code
   ```
   abc //
   def
   ```
   becomes `abc def`
   and
   ```
   abc//
   def
   ```
   becomes `abcdef`.
   Latin users often use the former one so they will not be affected.

isuffix:
This PR makes spaces due to newlines between CJK characters collapse to avoid creating a space when rendered. I've done so by splitting the `Space` syntax kind into two kinds to determine whether a space had a newline, and then by modifying the space-collapsing algorithm during Typst's realization step.

This is the behavior I mentioned in https://github.com/typst/typst/issues/792#issuecomment-2468896842 (I have since changed my username from `wrzian` to `isuffix`).

Closes #792 

Besides the below tradeoffs, I also have a few `TODO` comments that I would like input on.

## Tradeoffs

This _is_ a robust solution, but it makes multiple choices with tradeoffs that I'm not sure are desireable. I do not speak any CJK languages, so I'd appreciate feedback from the community about what is/isn't desired :)
CC @peng1999, @YDX-2147483647, @account-login

The main alternative design would be to solely resolve this in the parser or the AST, like YDX mentions in https://github.com/typst/typst/issues/792#issuecomment-2351027959. That may improve or harm any of the tradeoffs below based on your opinion.

Each of the tradeoffs is exemplified in one or more new test cases:

### Tradeoff: Collapsing happens during realization

Because collapsing happens after realization, a single newline space in the document may or may not collapse if its neighbors evaluate to CJK/non-CJK text dynamically.

Additionally, since a space element can itself be stored in a variable, static CJK characters can have different spacing due to stored space variables.

This is obviously one of the more contentious tradeoffs, but I think it's mostly fine. The first case is reasonable, and I doubt the second case is likely to affect real documents. But I am ok with changing either.

<details>
<summary>Test and output: <code>space-collapsing-cjk-dynamic</code></summary>

```typ
--- space-collapsing-cjk-dynamic ---
// Test cjk space collapsing with dynamic variables.
#let foo = [水果] // collapses
#foo
#foo

#let foo = [fruit] // doesn't collapse
#foo
#foo

#let one-newline = [
]
#let no-newline = [ ]
啊#one-newline;啊 // collapses

啊#no-newline;啊 // doesn't collapse
```

</details>

### Tradeoff: Normal spaces collapse only if they are adjacent to a newline space

Spaces that are not from newlines are kept, as in `空 格` (see the test case dropdown), but when adjacent to a newline space they will collapse.

While the basic behavior of collapsing a newline space or a newline space followed by a comment are straightforward, it's less clear whether a normal space followed by a comment and a newline space should combine as one space and collapse together, or if they should act separately. In addition, it's unclear if spaces with different styling should be able to combine and collapse together.

Currently, all three of the cases in this codeblock will combine and collapse their spaces. For me, I think the first is good, and the second is probably desired, but I'm less sure about the third case. (these are rendered in the dropdown below)
```typ
注//
释

空格 //
注释

*空格 *
换行
```

<details>
<summary>Tests and output: <code>issue-792-space-collapsing-cjk</code> and <code>space-collapsing-cjk-strong</code></summary>

```typ
--- issue-792-space-collapsing-cjk ---
// Test how spaces with/without newlines collapse in CJK text.

// No space from just a newline/comments
换
行

注//
释

多行/*
*/注释

// Should have a space from a space character
空 格

// With both a space and a newline it still collapses
空格 //
注释
```

```typ
--- space-collapsing-cjk-strong ---
// Test cjk space collapsing with strong emphasis.
*空 *格

*换*
行

// This space still collapses because it is followed by a newline space.
*空格 *
换行
```

</details>

### Tradeoff: Treating space values as equal

This is the one I'm least certain about, and would be improved by ignoring spaces in the parser/AST.

There are a few other test cases (`list-indent-trivia-nesting`, `list-indent-bracket-nesting`) that implicitly expect space elements to be equal to each other regardless of newlines. I feel like I also generally expect this behavior (I wrote those tetsts), so breaking them feels odd. I added a custom `PartialEq` implementation to `SpaceElem` that always returns true to make this work. However, I'm not sure if this is sound with the way Comemo caches data.

I'm totally ok with removing this if we stay with space-collapsing during realization, but it will require modifying the other test cases if we do, and we would probably also want to modify the `repr` of `SpaceElem`.

<details>
<summary>Test: <code>space-eq-newline</code></summary>

```typ
--- space-eq-newline ---
// Test whether spaces with/without newlines compare equal.
#let parbreak = [

]
#let one-newline = [
]
#let no-newline = [ ]
// parbreak is not equal
#assert.ne(one-newline, parbreak)
// spaces are equal despite newlines
// TODO: Would this break comemo?
#assert.eq(one-newline, no-newline)
```
</details>


YDX-2147483647:
Thank you for implementing this!

As for space-collapsing-cjk-strong (`*空格 *`), I don't think it's very important, because I guess no one would write it like this.
As for other aspects, I can take a closer look when I have time.

Besides, the [cjk-unbreak package](https://typst.app/universe/package/cjk-unbreak) has accumulated a few test cases in https://github.com/KZNS/cjk-unbreak/blob/2ea9b0ce3654ab537116499f63aa4077165192bc/test.typ. They might be relevant.

**Update on 2025-12-17:** The [cjk-spacer](https://typst.app/universe/package/cjk-spacer) package published recently is also relevant.
The two packages are developed by Chinese and Japanese people respectively. I guess there might be subtle differences, although I haven't compared them carefully yet.

crates/typst-syntax/src/lexer.rs
```diff
@@ -1061,6 +1039,15 @@ fn count_newlines(text: &str) -> usize {
     newlines
 }
 
+/// Whether a character is part of the Chinese, Japanese, or Korean scripts.
+#[inline]
+pub fn is_cjk(c: char) -> bool {
+    matches!(
+        c.script(),
+        Script::Han | Script::Hiragana | Script::Katakana | Script::Hangul
```
YDX-2147483647: I suggest we also include punctuation marks, for the following use case.

Typst source:

```
我就站住，
豫备她来讨钱。
“你回来了？”
她先这样问。
```

Result: 

> 我就站住，豫备她来讨钱。“你回来了？”她先这样问。

The list of CJK punctuation marks can be found in [clreq](https://www.w3.org/TR/clreq/#tables_of_chinese_punctuation_marks) and [jlreq](https://www.w3.org/TR/jlreq/#character_classes). (I don't know if K should be included here.)

This might be more complicated than it seems to be. For example, the following three characters are widely used in Chinese documents, but their categories are different. See [regex.pdf](https://ydx-2147483647.github.io/typst-set-font/docs/regex.pdf) for further comparisons.

- U+3001 `、` 顿号 (secondary comma) matches `\p{Script_Extensions=Han}`.

- U+FF0C `，` 逗号 (regular full-width comma) matches `\p{Script_Extensions=Common}`.

- U+201C `“` 上双引号 (left double quotation mark) matches `\p{Script_Extensions=Common}`, and it is also used in Latin documents.

They all match `\p{General_Category=Punctuation}` and `\p{Script=Common}`.

ponte-vecchio: For Korean, they should be included, along with its half-width counter parts of the Latin punctuations where possible---both of which are used.

peng1999: Please don't include Hangul here. Modern Korean uses spaces to separate words, so it's not relavant here. The function is better named as `is_cj`.

I used 'CJK' in the title of #792, that was a mistake. This issue only affects Chinese and Japanese.

isuffix: Thank you for the links, they're very helpful!

I can change to just CJ removing Hangul. I'll plan to move the function out of the lexer and leave the lexer behavior untouched (unless that should change too?)

However, I'd appreciate more thoughts on what to do for punctuation. It seems we have two categories of codepoints: non-ambiguous CJ punctuation and ambiguous CJ punctuation, such as left/right quotes. For non-ambiguous, I guess we can just treat them like all other CJ codepoints for collapsing, but I'm less sure what to do for ambiguous punctuation.

I presume it would be a good behavior for ambiguous punctuation followed by a CJ character (or vice-versa) to collapse a newline space, but what about an ambiguous punctuation next to another ambiguous punctuation? Are there any characters that would be likely to be split across lines? Should we look at the text language to determine this?

Also, some of the non-ambiguous CJ punctuation overlap in usage with Hangul. Should we be taking that into account?

This is also relevant for #5858, which has some related discussion around quotation marks.

isuffix: Another approach would be to just set the space collapsing behavior based on the text language, or an explicit property of say, `par()` (similar to the request in #710). This is more coarse-grained, but would simplify many of these considerations. Another tradeoff :face_exhaling:

ponte-vecchio: > Please don't include Hangul here. Modern Korean uses spaces to separate words, so it's not relavant here. The function is better named as `is_cj`.
> 
> I used 'CJK' in the title of #792, that was a mistake. This issue only affects Chinese and Japanese.

To clarify, my post was in relation to inclusion of punctuation marks. I also agree that Hangul should not be included here.

ponte-vecchio: > Thank you for the links, they're very helpful!
> 
> I can change to just CJ removing Hangul. I'll plan to move the function out of the lexer and leave the lexer behavior untouched (unless that should change too?)
> 
> However, I'd appreciate more thoughts on what to do for punctuation. It seems we have two categories of codepoints: non-ambiguous CJ punctuation and ambiguous CJ punctuation, such as left/right quotes. For non-ambiguous, I guess we can just treat them like all other CJ codepoints for collapsing, but I'm less sure what to do for ambiguous punctuation.
> 
> I presume it would be a good behavior for ambiguous punctuation followed by a CJ character (or vice-versa) to collapse a newline space, but what about an ambiguous punctuation next to another ambiguous punctuation? Are there any characters that would be likely to be split across lines? Should we look at the text language to determine this?
> 
> Also, some of the non-ambiguous CJ punctuation overlap in usage with Hangul. Should we be taking that into account?
> 
> This is also relevant for #5858, which has some related discussion around quotation marks.

This is the thing here. Modern Korean does use spaces as claimed by @peng1999, but to omit it altogether would create an awkward situation w.r.t its punctuations---also shared by Chinese and Japanese.

YDX-2147483647: > I presume it would be a good behavior for ambiguous punctuation followed by a CJ character (or vice-versa) to collapse a newline space, but what about an ambiguous punctuation next to another ambiguous punctuation?

I made a survey on LaTeX's treatment for linbreaks next to CJ punctuation marks, and appended the result to [my original post](https://github.com/typst/typst/issues/792#issuecomment-2303677367). It's quite chaos, so I don't recommend that we copy.

I suggest removing the space if and only if at least one side is a CJ character or a non-ambiguous CJ punctuation.

> Are there any characters that would be likely to be split across lines?

As far as I know, no.

> Should we look at the text language to determine this?

If the linebreak behavior can be controlled by `set text(control-linebreak: …)`, then we can make `auto` resolve to something depends on `text.lang`.
Otherwise, I'm against doing so.

---

Besides, @ponte-vecchio could you explain the usage in Korean?
If I understand it correctly, you use both `,` and `，`, but do you put spaces before or after them?

**Edit:** I've checked [klreq](https://www.w3.org/TR/klreq/).

> …in the case of punctuation, half-width punctuation (pause (comma) U+002C, full stop (period) U+002E) is used for horizontal writing, and full-width punctuation (ideographic comma U+3001, ideographic full stop U+3002) are used for vertical writing. ([§2.1.3](https://www.w3.org/TR/klreq/#fonts-horizvertpunct))

And according to [Figure 31](https://www.w3.org/TR/klreq/#punc-process:~:text=Figure%2031%20Brackets%20and%20quotation%20marks%20in%20vertical%20and%20horizontal%20writing), in horizontal writing, you put spaces after `,` and around `“…”`.

tats-u:
https://github.com/biomejs/biome/issues/7304#issuecomment-3219738064

Note: I personally prefer the `Emoji_Presentation` property to the `Emoji` property

crates/typst-syntax/src/lexer.rs
isuffix: The [most recent push](https://github.com/typst/typst/compare/dd58d2b3591d6a35b73e1d6ba81211ae0260919b..a70304f3dc97115a4c6db361bd9184bf4d712e0e) updated this.

We now collapse a newline space if either side matches the following function.

I'm certain this has edge problems, but it should be a good starting point. LMK what to change :)

```rust
/// Whether a character is part of the Chinese or Japanese scripts, or is a
/// non-ambiguous full-width punctuation.
///
/// These characters cause adjacent newline spaces to destruct in the source.
fn is_cj_or_cjk_punct(c: char) -> bool {
    match c.script() {
        Script::Han | Script::Hiragana | Script::Katakana => true,
        // TODO: Is this correct?
        // Attempting to check for just punctuation that is non-ambiguously CJK.
        // The `Script::Common` check rules out some cases, but I still need to
        // determine if there are any single-width characters to include or
        // double-width characters to exclude.
        Script::Common => c.width() == Some(2) && !c.is_emoji_char(),
        _ => false,
    }
}
```

isuffix:
@tats-u thanks! The link to the [CSS WG Draft](https://github.com/w3c/csswg-drafts/issues/5086) is also helpful.

crates/typst-syntax/src/lexer.rs
YDX-2147483647: I don't think `c.width() == Some(2)` is a good idea. According to https://docs.rs/unicode-width, it uses `East_Asian_Width` if it exists, and [`'\u{115F}'` HANGUL CHOSEONG FILLER](https://util.unicode.org/UnicodeJsps/character.jsp?a=115F) and [`'\u{17A4}'` KHMER INDEPENDENT VOWEL QAA](https://util.unicode.org/UnicodeJsps/character.jsp?a=17A4) also have width 2. Perhaps detecting whether `East_Asian_Width` exists is a better idea.

Also, there are three multi-character punctuation marks in Chinese: `——`, `……`, `⋯⋯`. The first two are commonly used.

Here's a list of widths of [punctuation marks mentioned in CLReq](https://www.w3.org/TR/clreq/#table_of_punctuation_marks).

<details>
<summary>~50 lines output + main.rs + Cargo.toml</summary>

```rust
'。'.width() = Some(2); // IDEOGRAPHIC FULL STOP
'．'.width() = Some(2); // FULLWIDTH FULL STOP
'，'.width() = Some(2); // FULLWIDTH COMMA
'、'.width() = Some(2); // IDEOGRAPHIC COMMA
'：'.width() = Some(2); // FULLWIDTH COLON
'；'.width() = Some(2); // FULLWIDTH SEMICOLON
'！'.width() = Some(2); // FULLWIDTH EXCLAMATION MARK
'‼'.width() = Some(1); // DOUBLE EXCLAMATION MARK — ✅ Rarely used
'？'.width() = Some(2); // FULLWIDTH QUESTION MARK
'⁇'.width() = Some(1); // DOUBLE QUESTION MARK — ✅ Rarely used
'⸺'.width() = Some(1); // TWO-EM DASH — ❓ Not sure
'～'.width() = Some(2); // FULLWIDTH TILDE
'-'.width() = Some(1); // HYPHEN-MINUS — ✅ Ambiguous
'–'.width() = Some(1); // EN DASH — ✅ Ambiguous
'—'.width() = Some(1); // EM DASH — ✅ Ambiguous
'·'.width() = Some(1); // MIDDLE DOT — ❓ Used in people names; Not sure if it's ambiguous
'・'.width() = Some(2); // KATAKANA MIDDLE DOT
'‧'.width() = Some(1); // HYPHENATION POINT — ✅ Rarely used
'/'.width() = Some(1); // SOLIDUS — ❓ Commonly used; Ambiguous?
'／'.width() = Some(2); // FULLWIDTH SOLIDUS
'「'.width() = Some(2); // LEFT CORNER BRACKET
'」'.width() = Some(2); // RIGHT CORNER BRACKET
'『'.width() = Some(2); // LEFT WHITE CORNER BRACKET
'』'.width() = Some(2); // RIGHT WHITE CORNER BRACKET
'“'.width() = Some(1); // LEFT DOUBLE QUOTATION MARK — ✅ Ambiguous
'”'.width() = Some(1); // RIGHT DOUBLE QUOTATION MARK — ✅ Ambiguous
'‘'.width() = Some(1); // LEFT SINGLE QUOTATION MARK — ✅ Ambiguous
'’'.width() = Some(1); // RIGHT SINGLE QUOTATION MARK — ✅ Ambiguous
'（'.width() = Some(2); // FULLWIDTH LEFT PARENTHESIS
'）'.width() = Some(2); // FULLWIDTH RIGHT PARENTHESIS
'《'.width() = Some(2); // LEFT DOUBLE ANGLE BRACKET
'》'.width() = Some(2); // RIGHT DOUBLE ANGLE BRACKET
'〈'.width() = Some(2); // LEFT ANGLE BRACKET
'〉'.width() = Some(2); // RIGHT ANGLE BRACKET
'【'.width() = Some(2); // LEFT BLACK LENTICULAR BRACKET
'】'.width() = Some(2); // RIGHT BLACK LENTICULAR BRACKET
'〖'.width() = Some(2); // LEFT WHITE LENTICULAR BRACKET
'〗'.width() = Some(2); // RIGHT WHITE LENTICULAR BRACKET
'〔'.width() = Some(2); // LEFT TORTOISE SHELL BRACKET
'〕'.width() = Some(2); // RIGHT TORTOISE SHELL BRACKET
'［'.width() = Some(2); // FULLWIDTH LEFT SQUARE BRACKET
'］'.width() = Some(2); // FULLWIDTH RIGHT SQUARE BRACKET
'｛'.width() = Some(2); // FULLWIDTH LEFT CURLY BRACKET
'｝'.width() = Some(2); // FULLWIDTH RIGHT CURLY BRACKET
'＿'.width() = Some(2); // FULLWIDTH LOW LINE
'﹏'.width() = Some(2); // WAVY LOW LINE
'●'.width() = Some(1); // BLACK CIRCLE — ✅ Rarely used
'•'.width() = Some(1); // BULLET — ✅ Rarely used
```

```rust
use unicode_width::UnicodeWidthChar;

fn main() {
    let chars = vec![
        // Copied from https://www.w3.org/TR/clreq/#table_of_punctuation_marks
        '。', '．', '，', '、', '：', '；', '！', '‼', '？', '⁇', '⸺',
        // Skip multi-character punctuation marks
        // '——',
        // '……',
        // '⋯⋯',
        '～', '-', '–', '—', '·', '・', '‧', '/', '／', '「', '」', '『', '』', '“', '”', '‘', '’',
        '（', '）', '《', '》', '〈', '〉', '【', '】', '〖', '〗', '〔', '〕', '［', '］', '｛',
        '｝', '＿', '﹏', '●', '•',
    ];

    for c in chars {
        println!(
            "'{c}'.width() = {:?}; // {:?}",
            c.width(),
            unicode_names2::name(c).unwrap()
        );
    }
}
```

```toml
[package]
name = "char-width"
version = "0.1.0"
edition = "2024"

[dependencies]
unicode-width = "0.2.2"
unicode_names2 = "2.0.0"
```

</details>

tats-u: @YDX-2147483647 According to https://www.unicode.org/Public/17.0.0/ucd/EastAsianWidth.txt, The EAW of U+17A4 is N, not W or F. Did you confuse it with something different?

tats-u: The Script property of U+115F is Hangul and EAW of it is W.

https://www.unicode.org/Public/17.0.0/ucd/Scripts.txt

YDX-2147483647: > The EAW of U+17A4 is N, not W or F. Did you confuse it with something different?

Hi! I didn't make it clear. What I mean is as follows.

1. `c.width()` uses a complex rule to determine the width, and EAW is one of the factors.
2. The full rule is documented on https://docs.rs/unicode-width, and it says that U+17A4 and U+115F will give width 2. (For these two specific characters, EAW does not contribute to `c.width()`.)
3. Therefore, I don't think `c.width() == Some(2)` is a good criterion.

tats-u: I agree with your opinion that `unicode_with` is not suitable for determining whether the character is CJ(K).
We should use the EAW property directly. Also, U+FF61 HALFWIDTH IDEOGRAPHIC FULL STOP is a (legacy) Japanese character but whose Script is not Katakana but Common and whose EAW is H. All non-Hangul/Korean characters whose EAW is H must be treated as Chinese/Japanese.

tats-u: If we have issues with determining whether characters are Korean, we should test their test cases in FIrefox and report them in https://bugzilla.mozilla.org/ (its bug tracker):

```html
<div lang=ja><!-- ← or zh -->
  Test
  Case
  Here
</div>
```

Removing a newline around a punctuation is enabled only in Chinese and Japanese.

tats-u:
> The link to the https://github.com/w3c/csswg-drafts/issues/5086 is also helpful.

It should be just a link to tracker. No concrete specification is stipulated in CSS WG Draft (https://drafts.csswg.org/css-text-4/#line-break-transform) now.

Prettier's issue (fixed):

- https://github.com/prettier/prettier/issues/6516

"K" should be excluded from the title, too.

FYI, the following JS expression returns `[]`, which means that no "Korean-dedicated punctuation chracters (whose Script is Hangul and General Category is Punctuation)" exists:

```js
Iterator.from((function*() {for (let i = 0; i <= 0x10ffff; i++) yield i;})()).filter(cp => /[\p{P}&&\p{sc=Hang}]/v.test(String.fromCodePoint(cp))).toArray()
```

isuffix:
I rebased off main and have refactored the space collapsing algorithm quite a bit. It now both depends on #7609 and reifies the kinds of actions that the algorithm takes in a new enum. This shouldn't change any of this PR's high-level behavior (we still discard newline spaces if either side is a space-discarding character), but the new organization really helps me keep the full algorithm in my head.

Additionally, I've updated the space-discarding character set to include common-script characters only if their East Asian Width is F/W/H but they are not emoji (although I still anticipate that this can be improved). This is aided by the new test `newline-space-discarding-edge-cases` which lets us check characters at a very fine grained level.

To determine the East Asian Width, I've moved to using the `icu_properties` crate which seems to be the highest quality source of the property in Rust right now. I've also moved the Script and Emoji checks to use `icu_properties` for consistency. However, we are still inexplicably using version 1.4 while `icu` is on 2.0 since May (now 2.1), so I'd appreciate suggestions as to whether it should be updated (it does have breaking changes). If we did want to stick to crates from `unicode-rs`, I believe we could ask the maintainer to add the EAW to the [`unicode-properties` ](https://docs.rs/unicode-properties/0.1.4/unicode_properties/) crate. However, it was very annoying that I could not find any real guidance online as to whether we should prefer `icu` vs. `unicode-rs`.

crates/typst-realize/src/spaces.rs
```diff
@@ -0,0 +1,207 @@
...
+            matches!(
+                EAW_DATA.get(c),
+                EastAsianWidth::Halfwidth
+                    | EastAsianWidth::Fullwidth
+                    | EastAsianWidth::Wide
+            ) && !EMOJI_DATA.contains(c)
+        }
+        _ => false,
+    }
+}
```
isuffix: Here is the new space-discarding character check.

tests/suite/text/space.typ
```diff
@@ -58,3 +58,208 @@ LLLLLLLLLLLLLLLLLL R _L_
...
+// Test newline space discarding for edge case characters.
+// Characters inspired by clreq and jlreq:
+// https://www.w3.org/TR/clreq
+// https://www.w3.org/TR/jlreq
+
+// Whether each string should discard an adjacent newline space.
+#let should-discard = (
+  // Basic characters in different languages
+  ("A", false),
+  ("漢", true),
```
isuffix: Let me know what characters I should add to this test and whether any should change.

tats-u: - <span lang=ja>ｱ</span> U+FF71 true ←EAW is H
- <span lang=ja>㊙</span> U+3299 ? (false in my opinion) ←This is a default _text_ presentation character like ©. Without a succeeding U+FE0F, it should not be displayed as emoji if a proper Japanese font is assigned. However, the current Firefox treats both 2 symbols as emoji because it treats every emoji character (a character that has an Emoji property) as emoji. There is Emoji_Presentation property.
- <span lang=ko>한</span> U+D55C false ←https://github.com/prettier/prettier/issues/6516
- ₩ U+20A9 ? (false in my opinion) ←EAW is H but Korea-dedicated. Since the EAW of ¥ U+00A5 for Japan and PRC is Na, ₩ should not be treated as Chinese-or-Japanese, either. Shamefully `&& c != '\u{20A9}'` is needed.

It is recommended to also check the character two positions away as necessary to check emojiness of the adjacent grapheme more accurately.

crates/typst-realize/src/spaces.rs
tats-u: There seems to be some other East Asian scripts that prefers without-space-style e.g. Yi.
Also you should add `// Especially Hangul` above `_ => false,`.

OverflowCat: > There seems to be some other East Asian scripts that prefers without-space-style e.g. Yi.

So imo the feature is more about scripts that does not use spaces in writing than CJK characters only. This reminds me of Tangut -- don't know if they are Script::Han, though. Also, do we need to take those languages that do not use space within sentence level into account? e.g. Tibetan only uses spaces after a punctuation mark like [`། `](https://en.wiktionary.org/wiki/%E0%BC%8D).

tats-u: The advantage of relying only on EAW for non-Hangul scripts recognition is that it eliminates the need to worry about about which scripts must be covered, including minor scripts like Yi or Tangut.

tats-u: > 't know if they are Script::Han, though.

A dedicated Script Property Value `Tangut` is assigned to Tangut since Unicode 9 (2016).

https://www.unicode.org/standard/supported.html

laurmaedje:
> However, we are still inexplicably using version 1.4 while icu is on 2.0 since May (now 2.1), so I'd appreciate suggestions as to whether it should be updated (it does have breaking changes).

The reason is just that nobody got to it. But there is https://github.com/typst/typst/pull/7412. There's still some unresolved questions though.

> However, it was very annoying that I could not find any real guidance online as to whether we should prefer icu vs. unicode-rs.

Definitely ICU.

isuffix:
The first push rebases off the updated #7609, the second/third just add some of the mentioned edge case characters, although I haven't updated their implementation yet.

isuffix:
I'd like to move this PR forward, but first—based on past discussions on Discord—we are going to move the definition for which characters should discard inferred spaces to the Codex crate.

[I wrote an update](https://discord.com/channels/1054443721975922748/1277628305142452306/1532111311482130572) in the [Codex forge](https://discord.com/channels/1054443721975922748/1277628305142452306) on Discord, and I would appreciate if any feedback could be posted there. But I am copying the message here for reference:



> 
> I'd like to push [PR 7350](<https://github.com/typst/typst/pull/7350>) (CJ newline space discarding) over the finish line for initial inclusion in Typst. However, we first need to address how to add the character definitions for this behavior to Codex. I'm posting here to get feedback before I submit a new PR to Codex and update the existing PR.
> 
> I know that there was still concern about defining the feature as a static property on specific characters vs. basing this on a show/set rule or adding a general mechanism for customizing newlines in markup, but I still think that the current proposal makes for a sensible default that we can add further customization to in the future.
> 
> For reference, you may want to read [my previous comment](https://discord.com/channels/1054443721975922748/1176062736514429008/1453446046847078615) in the CJK forge along with [Laurenz's response](https://discord.com/channels/1054443721975922748/1176062736514429008/1457647880193052694).
> 
> ### Adding to Codex
> 
> To add this to Codex I think it makes sense to create a new Cargo feature like `numeral-systems` and `styling` that we define in its own file under `src`. My current thought is to call the feature `space-discarding`, since the definition is intended to determine whether an inferred space due to a newline should be kept or discarded.
> 
> Please let me know if this feature name is fine or if you have any suggestions.
> 
> ### What will be added
> 
> As for what will be added under that feature, I'd still like to expose a simple function for whether to discard a newline space, but I think my original `is_space_discarding(char) -> bool` function has two problems that I'd like to fix.
> 
> The first problem is that we should really be considering grapheme clusters instead of codepoints, and I'd like the function to take `&str` instead of `char` to do so. The second problem is that what exactly the function is meant to represent feels ambiguous to me, and I think we should change how we define it to make it easier to reason about.
> 
> #### Taking `char` vs. `&str`
> 
> Instead of codepoints, what we really want to analyze are the grapheme clusters that surround the inferred space. This is because Unicode is complex and the immediately surrounding codepoints could be something like variation selectors that have a base codepoint that is what we actually want to make our determination based on. Accepting `char` as our argument would require every caller to implement that logic when it should really be the library which handles analyzing grapheme clusters.
> 
> And to avoid needing to expose two different functions for analyzing the initial/final grapheme clusters of a string, I think it would be best to have one function which accepts the two strings that come before and after the inferred space.
> 
> The new function would look like: `discard_space_between(before: &str, after: &str) -> bool`
> 
> I haven't actually implemented any grapheme cluster analysis yet (my current draft still just uses the first/last codepoint), but I think doing so is worthwhile and it's worth moving to `&str` now so that we can more easily add this in the future. And if we decide against doing any grapheme cluster analysis, changing from `&str` to `char` will likely be easier for users of the function than changing from `char` to `&str`.
> 
> #### Defining the function more precisely
> 
> I think only considering the high-level behavior of keeping or discarding spaces is likely to lead to ambiguity and confusing arguments. Instead of focusing the behavior of Typst itself, I think we should base the function on the relevant properties of the writing system/language the surrounding characters are from, and focus on mapping characters (or grapheme clusters) to the correct writing system/language. This would also make it easy to improve the function's accuracy by adding a `lang` parameter in the future to handle characters like puncuation whose writing system is ambiguous.
> 
> I think the property we should define and use is **whether a writing system/language uses spaces between words.** Chinese and Japanese dont, but Korean and many other writing systems do. Then the function would decide to discard an inferred space if either surrounding character is unambiguously from a writing system which does not uses spaces between words. This matches the current implementation in the PR, and I find it to be a reasonable default.
> 
> For the initial implementation I would leave it there, but I know that abugida-based writing systems such as Thai, Lao, Burmese, etc. use spaces between phrases and sentences but not between words. And I know other writing systems may also prefer different definitions, e.g. Vietnamese which uses spaces between _syllables_. But I think word-spacing is a good place to start and we can add more properties to account for other writing systems over time.
> 
> Please let me know if you have any feedback on this. If there are no concerns, I'll plan to submit this design as a PR to Codex and base PR 7350 on the new definition sometime next week.

isuffix:
This is now based on https://github.com/typst/codex/pull/182!

Since this updates the Codex dependency, I had to add a new commit to remove some deprecated symbol tests.

crates/typst-library/src/text/space.rs
```diff
@@ -6,13 +6,28 @@ use crate::foundations::{
 };
 
 /// A text space.
-#[elem(Unlabellable, PlainText, Repr)]
-pub struct SpaceElem {}
+#[elem(Unlabellable, PlainText, Repr, PartialEq)]
+pub struct SpaceElem {
+    #[required]
+    pub had_newline: bool,
```
laurmaedje, line 12: currently, this property is exposed, as observed e.g. by
```typ
#let space = [ ].func()
#space()
```
or
```typ
#panic([ ].had-newline)
```

it should be made `#[internal]` in my opinion.

then, it will need to be populated somehow. if we add a typical `Construct for SpaceElem`, that's technically also a bit breaking if people call `#space()` after getting ahold of `[ ].func()`, though that's kinda on them. but we can also just avoid that by adding `#[parse(false)]` instead of `Construct for SpaceElem`.

crates/typst-syntax/src/kind.rs
```diff
@@ -586,9 +591,11 @@ impl SyntaxKind {
             Self::BlockComment => None, // none
 
             Self::Markup => Known(Markup),
-            Self::Text => Text,        // none: Raw | markup: expr
-            Self::Space => Space,      // code/math: part | markup: expr/part
-            Self::Linebreak => Parent, // math/markup: expr
+            Self::Text => Text, // none: Raw | markup: expr
+            // TODO: Add tests for these separately, and re-check the valid modes
```
laurmaedje, line 595: Is this TODO for this PR or later?

tests/suite/symbols/symbol.typ
```diff
@@ -176,20 +176,6 @@
   `symbol("🖅")`.text,
 )
 
---- symbol-sect-deprecated paged ---
```
laurmaedje, line 179: if I'm seeing things correctly, these are the only tests that check that codex deprecations works correctly, so it would better it these would be updated to another deprecated symbol instead (we could also consider for codex to separately expose a perma-deprecated testing symbol separately, which we hook up in the test namespace, but that sounds like a bit of effort already and unrelated to the current work).

crates/typst-realize/src/spaces.rs
```diff
@@ -1,4 +1,5 @@
-//! The space collapsing infrastructure for realization.
+//! The space collapsing and discarding infrastructure for realization.
+use codex::space_discarding::discard_space_between;
```
laurmaedje, line 2: the codex import can go with the other external crate imports, but a blank line should follow the leading doc comment.

crates/typst-realize/src/spaces.rs
```diff
@@ -7,17 +8,52 @@ use typst_library::layout::HElem;
...
+    /// When destructive, we skip any future spaces.
     Destructive,
-    /// Normal elements. Spaces are only kept if supported on both sides.
-    Supportive,
-    /// Adjacent spaces collapse as one with the styles of the first space.
-    Space,
+    /// When supportive, we usually keep future spaces, but we will skip newline
+    /// spaces if our text ends in a space-discarding character.
+    Supportive { text: Option<&'a str> },
+    /// A current space that did not have a newline and remembers the preceding
```
laurmaedje, line 22: "did not have a newline" together with `had_newline: bool` does not really make sense

crates/typst-realize/src/spaces.rs
```diff
@@ -91,32 +130,77 @@ pub(crate) fn collapse_state(content: &Content, styles: StyleChain) -> SpaceStat
...
+fn for_space(state: SpaceState, had_newline: bool) -> (SpaceAction, SpaceState) {
+    match state {
+        SpaceState::Destructive => (SpaceAction::Skip, SpaceState::Destructive),
+        SpaceState::Supportive { text: prev_text } => {
+            (SpaceAction::Keep, SpaceState::Space { prev_text, had_newline })
+        }
+        SpaceState::Space { prev_text, had_newline: mut prev_nl } => {
+            prev_nl |= had_newline;
+            (SpaceAction::Skip, SpaceState::Space { prev_text, had_newline: prev_nl })
+        }
```
laurmaedje, line 191: no need for the mutation here
```rs
        SpaceState::Space { prev_text, had_newline: prev_nl } => {
            let had_newline = prev_nl || had_newline;
            (SpaceAction::Skip, SpaceState::Space { prev_text, had_newline })
        }
```
or inlining the or into the `SpaceState::Space` construct, but I presume you wanted to avoid it formatting very differently then the other branches?

tests/suite/text/space.typ
```diff
@@ -71,3 +71,97 @@ LLLLLLLLLLLLLLLLLL R _L_
...
+*空格 *
+换行
+
+// The second space here also collapses because it follows a newline space.
+空格
+* 换行*
+
+--- newline-space-discarding-regex-show-rule paged ---
+// Test newline space discarding with regex show rules.
+#show regex("注 释|换行 newline|newline 换行"): set text(red)
```
laurmaedje, line 138: I think this test would profit from a fourth case with Latin + Latin (that turns red) and a Chinese + Chinese comment on the first line. That would complete the 2x2 matrix.

The line comment after "注" also seems unrelated to the test's purpose as this kind of setup is already tested above.

laurmaedje (commented):
Just some nits and a concern about `had-newline` being publicly exposed. Otherwise, this looks good!

crates/typst-library/src/text/space.rs
isuffix: I'll add `#[internal]` and `#[parse(false)]` :+1:

crates/typst-syntax/src/kind.rs
isuffix: For this PR :sweat_smile:. Updating.

tests/suite/symbols/symbol.typ
isuffix: It's actually not too hard to iterate over all of the symbols by parsing their `repr`. I've added a test which annotates all the present deprecation warnings, although it won't test anything when none are present.

LMK if I should try to add the perma-deprecated symbol idea.

crates/typst-realize/src/spaces.rs
isuffix: Inlining is nice. I assume the mutation was leftover from some earlier structure where it was cleaner.

tests/suite/symbols/symbol.typ
laurmaedje: If there's still a deprecated symbol that we can instead use, I would prefer that for this PR. If there is none, then we can pursue the perma-deprecated symbol thing.

isuffix: There are still some deprecated symbols. The new test case should be easier to update with `--update` going forward.

isuffix:
This should be ready for another review.

isuffix:
I realized that space-discarding around inline elements wasn't consistent, so I fixed it with some new tests and small refactors. For example, in `あ\n#box()\nあ` we would discard the second space but keep the first one instead of discarding both newline spaces.

The fix changes to consistently use the `for_text` function for all supportive element transitions, which also led me to rename it from `for_text` to `for_supportive`. And then I created a parallel `for_destructive` function for the `Destructive` state transitions.

PR #7354, opened 2025-11-11
Fix that `text.font` does not return `covers` info

RungeCC:
After https://github.com/typst/typst/pull/5305#diff-0adb942019d59d0f74ea4230b8e4f398efeba0b944786b42f30edfeeaa141943R832 font can have a `covers`, however, currently `text.font` does not return `covers` info, see [here](https://typst.app/project/RzBZqrJdkOBU1N04d2XjAj), which is weird. 

This PR is a simple fix.


laurmaedje:
Thanks for the fix! Could you add a test for this in `tests/suite/text/font.typ`?

RungeCC:
Sure, I add tests and find previously a special case is missing to handle.

laurmaedje:
Thanks!

PR #7357, opened 2025-11-11
Better handling of rectangle strokes

saecki:
This is somewhat of a continuation of #5688

It changes/improves multiple things for rectangle corners with radii:
- Add support for stroke caps
  - If the radius is larger than the stroke-width, otherwise I found too many edge cases
- Differentiate between missing adjacent strokes and a stroke of width zero
- Fix a bug with computation of the `mid` corner control point. This can be observed on stroked segments.

And also a bug with dashed strokes where a change of the cap or thickness was ignored.

The test cases are a bit much, but I think they capture quite a large amount of behavior.

This is the document I used for testing:
https://typst.app/project/rd7nScpashEKJcfsI0nNrw

<table>
<tr>
<th> main <a href="https://github.com/user-attachments/files/23482661/test-main.pdf">PDF</a> </th>
<th> pr <a href="https://github.com/user-attachments/files/23482663/test-pr.pdf">PDF<a> </th>
</tr>
<tr>
<td width="50%">
test-main
</td>
<td widht = "50%">
test-pr
</td>
</tr>
</table>


tests/suite/visualize/rect.typ
```diff
@@ -127,3 +127,375 @@
...
+  test-block(cap: "square", radius: 12pt, dash: "loosely-dashed"),
+  test-block(cap: "round", radius: 12pt, dash: "loosely-dashed"),
+)
+
+--- rect-stroke-variations-with-radius-and-adjacent-zero-width-stroke ---
+#set page(width: 12cm)
+#let bg-rect(inner) = {
+  rect(inset: 0pt, outset: 0pt, fill: luma(200), inner)
+}
+#let test-block(cap: "butt", dash: none, adjacent: none, radius: 0pt) = {
```
laurmaedje: To cut down on the duplication, you can import from a shared file marked with `// SKIP`, see e.g. `tests/suite/introspection/switch.typ`.

Maybe we can also get the reference images a _bit_ smaller in terms of KB?

crates/typst-library/src/layout/point.rs
```diff
@@ -60,6 +60,16 @@ impl Point {
         Abs::raw(self.x.to_raw().hypot(self.y.to_raw()))
     }
 
+    // TODO: this is a bit awkward on a point struct.
```
laurmaedje, line 63: is this TODO for the PR or more long-term?

laurmaedje (commented):
looking the PDFs, the changes look very nice!!

crates/typst-library/src/layout/point.rs
saecki: This was meant more long term. There are also some other methods that should rather go onto a vector struct. But I think it's somewhat out of scope for this PR to introduce that and do all the refactoring.

tests/suite/visualize/rect.typ
saecki: Done. I also slightly reduced the size of the images.

PR #7362, opened 2025-11-12
Add `typst eval` CLI command

### LINKED ISSUE #7008: Return location information in `typst query`

### LINKED ISSUE #7344: `typst eval` subcommand

TimerErTim:
# Description

## Deprecates `query` subcommand
Display a deprecation warning when the `query` subcommand is invoked and also in clap help.

## No tests?
I have found no tests for individual CLI commands. So I decided to not design a new testsuite for raw CLI subcommands as that would be very much out of scope for this PR.

Closes #7344 
Closes #7008

### DIFF
--- a/crates/typst-cli/src/eval.rs
+++ b/crates/typst-cli/src/eval.rs
@@ -1,6 +1,3 @@
-use crate::args::{EvalCommand, Target};
-use crate::world::SystemWorld;
-use crate::{compile::print_diagnostics, set_failed};
 use comemo::Track;
 use ecow::eco_format;
 use typst::diag::{HintedStrResult, SourceResult, Warned};
@@ -9,6 +6,11 @@ use typst::syntax::{Span, SyntaxMode};
 use typst::{World, engine::Sink, introspection::Introspector, layout::PagedDocument};
 use typst_eval::eval_string;
 use typst_html::HtmlDocument;
+
+use crate::args::{EvalCommand, Target};
+use crate::compile::print_diagnostics;
+use crate::set_failed;
+use crate::world::SystemWorld;
 
 /// Execute a query command.
 pub fn eval(command: &'static EvalCommand) -> HintedStrResult<()> {
@@ -46,7 +48,7 @@ pub fn eval(command: &'static EvalCommand) -> HintedStrResult<()> {
                     &[]
                 }
             };
-            // Collect additional warnings from code evaluation
+            // Collect additional warnings from evaluating the expression.
             warnings.extend(sink.warnings());
 
             print_diagnostics(
@@ -74,7 +76,7 @@ pub fn eval(command: &'static EvalCommand) -> HintedStrResult<()> {
     Ok(())
 }
 
-/// Evaluates the expression with code SyntaxMode and no scope.
+/// Evaluates the expression with code syntax mode and no scope.
 fn evaluate_expression(
     expression: String,
     sink: &mut Sink,
--- a/crates/typst-cli/src/query.rs
+++ b/crates/typst-cli/src/query.rs
@@ -33,7 +33,7 @@ pub fn query(command: &'static QueryCommand) -> HintedStrResult<()> {
             .map(|output| output.map(|document| document.introspector)),
     };
 
-    // Add deprecation warning to warnings
+    // Add deprecation warning.
     warnings.push(deprecation_warning(command));
 
     match output {
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -103,10 +103,10 @@ impl SystemWorld {
                 .ok_or(WorldCreationError::InputOutsideRoot)?;
             FileId::new(None, main_path)
         } else if matches!(input, Some(Input::Stdin)) {
-            // Return the special id of STDIN
+            // Return the special id of STDIN.
             *STDIN_ID
         } else {
-            // Return the special id of EMPTY/no input at all otherwise
+            // Return the special id of EMPTY/no input at all otherwise.
             *EMPTY_ID
         };
 
@@ -430,9 +430,9 @@ fn system_path(
 
 /// Reads a file from a `FileId`.
 ///
-/// If the ID represents stdin it will read from standard input,
-/// else if it represents empty/no input at all it will return an empty vector,
-/// otherwise it gets the file path of the ID and reads the file from disk.
+/// - If the ID represents stdin it will read from standard input.
+/// - If it represents empty/no input at all it will return an empty vector.
+/// - Otherwise, it gets the file path of the ID and reads the file from disk.
 fn read(
     id: FileId,
     project_root: &Path,
--- a/crates/typst-library/src/foundations/mod.rs
+++ b/crates/typst-library/src/foundations/mod.rs
@@ -81,8 +81,9 @@ use ecow::EcoString;
 use typst_syntax::{Spanned, SyntaxMode};
 
 use crate::diag::{SourceResult, StrResult, bail};
+use crate::engine::Engine;
+use crate::introspection::Introspector;
 use crate::{Feature, Features};
-use crate::{engine::Engine, introspection::Introspector};
 
 /// Hook up all `foundations` definitions.
 pub(super) fn define(global: &mut Scope, inputs: Dict, features: &Features) {


MDLC01:
This would also close https://github.com/typst/typst/issues/7008

crates/typst-cli/src/query.rs
```diff
@@ -17,19 +17,26 @@ use crate::world::SystemWorld;
...
             .map(|output| output.map(|document| document.introspector)),
         Target::Html => typst::compile::<HtmlDocument>(&world)
             .map(|output| output.map(|document| document.introspector)),
     };
 
+    // Add deprecation warning to warnings
+    warnings.push(
+        SourceDiagnostic::warning(Span::detached(), "`query` is deprecated")
+            .with_hint("use `eval` instead"),
+    );
```
laurmaedje: it would be nice if the hint could show the exact eval command you need to use to replace your specific invocation of query.

crates/typst-library/src/foundations/mod.rs
```diff
@@ -66,6 +66,7 @@ pub use self::target_::*;
 pub use self::ty::*;
 pub use self::value::*;
 pub use self::version::*;
+use comemo::Track;
 pub use typst_macros::{scope, ty};
 use typst_syntax::SyntaxMode;
```
laurmaedje: could you move the import down below to the existing `use comemo::TrackedMut`? might as well fix the typst_syntax one while at it.

crates/typst-cli/src/args.rs
```diff
@@ -713,14 +795,27 @@ fn parse_page_number(value: &str) -> Result<NonZeroUsize, &'static str> {
...
+fn string_input_value_parser() -> impl TypedValueParser<Value = StringInput> {
     clap::builder::OsStringValueParser::new().try_map(|value| {
         if value.is_empty() {
             Err(clap::Error::new(clap::error::ErrorKind::InvalidValue))
         } else if value == "-" {
-            Ok(Input::Stdin)
+            Ok(StringInput::Stdin)
         } else {
-            Ok(Input::Path(value.into()))
+            Ok(StringInput::String(value.to_string_lossy().into()))
```
laurmaedje: I'm not sure we want to be lossy here. Is this the default behavior of clap when a string is required or does it error?

crates/typst-cli/src/eval.rs
```diff
@@ -0,0 +1,148 @@
...
+    World,
+    diag::{HintedStrResult, SourceResult, Warned},
+    engine::Sink,
+    foundations::{Binding, Context, Scope, StyleChain, Value},
+    introspection::Introspector,
+    layout::PagedDocument,
+    syntax::Span,
+};
+use typst_eval::eval_string;
+use typst_html::HtmlDocument;
```
laurmaedje: could you flatten these imports to make them consistent with the code base's style?

crates/typst-cli/src/world.rs
```diff
@@ -26,6 +26,11 @@ use crate::package;
 static STDIN_ID: LazyLock<FileId> =
     LazyLock::new(|| FileId::new_fake(VirtualPath::new("<stdin>")));
 
+/// Static `FileId` allocated for empty.
+/// This is to ensure that a file is read in the correct way.
+static EMPTY_ID: LazyLock<FileId> =
```
laurmaedje, line 31: I assume the idea of this is that we need a `SystemWorld` to properly compile the `statement`, but at the same time it requires a main document and there's not really a way around that?

I think "Static `FileId` allocated for empty." is a bit obtuse, maybe this could be better explained here.

crates/typst-cli/src/eval.rs
```diff
@@ -0,0 +1,148 @@
...
+    mode: SyntaxMode,
+    scope: Scope,
+    world: &dyn World,
+    introspector: &Introspector,
+) -> SourceResult<Value> {
+    eval_string(
+        &typst::ROUTINES,
+        world.track(),
+        // TODO: propagate warnings
+        Sink::new().track_mut(),
```
laurmaedje: Retaining the warnings here and printing them should be _fairly_ straightforward I think and might be good to have

laurmaedje (commented):
Thanks for the PR, it looks excellent!

I left a few minor comments, but the main thing to happen before merge is that we're confident in the API. What I posted in the issue was my initial idea, but upon playing around with your interpretation of it, a few questions have arisen.

First of all, I'm not sure it's desirable to interpret `scope` values as Typst code. I'd have expected the scope to always contain strings. Otherwise, it's not much different from just interpolating the variables into the main string. The idea was to offer a safe way to interpolate arbitrary strings. However, I'm not certain about this either. It feels somewhat redundant with `sys.inputs`. There have been various discussions about that one also supporting paths and bytes in some capacity (see https://github.com/typst/typst/issues/5382#issuecomment-3150286257) and those might crop up here, too. Since `sys.inputs` already works with `typst eval` out of the box, perhaps it's best to just rely on it and remove `scope` after all?

For `mode`, I'm also open whether it's actually pulling its weight then. (Should we remove `scope`, I don't feel quite as strong the urge for consistent with the `eval` function.)

What do you think?

crates/typst-cli/src/args.rs
TimerErTim: Clap’s StringValueParser checks whether the input is valid UTF-8, and if it isn’t, it displays the command’s help/usage message. This means there’s no lossy conversion—invalid input results in an explicit error, which appears to be the standard for Clap.

I will use the `NonEmptyStringValueParser` instead of the `OsStringValueParser`. This should solve the issue

crates/typst-cli/src/world.rs
TimerErTim: Yes, the assumption is correct. I thought about implementing a new `World` struct, but that seemed way to repetitive because most of the functionality of `SystemWorld` would still be needed.

In the end, adding another `FileInput` special case seemed to be the easiest, least intrusive solution.

I will elaborate on the comments though.

TimerErTim:
Thank you for the review and the ideas.

## `--scope` argument

I think having `--scope` values interpreted as Typst code is quite valuable. It allows simple and expressive usage such as:

```bash
typst eval x+y --scope x=2 --scope y=6
# 8
```

This also closely mirrors the familiar `eval` function inside Typst itself.

### Against interpreting scope values as strings

If `--scope` values were interpreted as plain strings (which you *can* emulate using `--scope 'x="2"'`), the result would be:

```bash
# "26"
```

This means we would lose a lot of flexibility, especially when leveraging shell interpolation, e.g. `--scope x=$(echo 1)`.

Additionally—and as you already mentioned—there would be almost no functional difference compared to `--input`, making `--scope` essentially redundant.

Even though I don’t currently have a practical use case for it, interpreting scope values as Typst code also enables injecting math contexts or markup content, allowing for things like:

```bash
typst eval x+y --scope 'x="2"' --scope y='$6$'
# {"func":"sequence","children":[{"func":"text","text":"2"},{"func":"equation","block":false,"body":{"func":"text","text":"6"}}]}
```

I have no idea why you would ever need this—but experience shows that users come up with all sorts of unanticipated (yet valid) uses.

### Against dropping `--scope` entirely

Most of the points above apply here as well. Without `--scope`, we would lose a great deal of flexibility. With only `--input`, examples like the following become much harder to write:

```bash
typst eval 'if x == 0 { query(heading) } else { query(heading.where(level: x)) }' \
  --scope x=$(<some script yielding a level>) \
  --in ...
# headings of all levels or only specific levels
```

Without `--scope`, the same dynamic value would have to be interpolated multiple times or parsed manually from `sys.input` via JSON or similar. This adds complexity and reduces expressiveness.

## `--mode` argument

From a practical standpoint, this argument can likely be omitted. Inside Typst itself it makes sense, since the evaluator actually renders math or content blocks. But from the CLI’s perspective, the output becomes something like:

```json
{
  "func": "equation",
  "block": false,
  "body": {
    "func": "frac",
    "num": { ... },
    "denom": { ... }
  }
}
```

(example generated by the current implementation using
`typst eval '#a/#b' --mode math --scope a='$2 times 3$' --scope b='6' --pretty`)

This form is not particularly meaningful for most CLI use cases.

### Alternative

If we remove the `--mode` argument and always evaluate in “code” mode, but a rare situation arises where a user actually *does* want the math-mode AST, they can simply wrap the expression:

```bash
typst eval '$#a/#b$' --scope a='$2 times 3$' --scope b='6' --pretty
```

This seems like a reasonable workaround, especially since such cases should be rare.

### Downsides of having `--mode`

That said, keeping the argument incurs very little maintenance cost. The biggest issue is the ambiguity and conversion between `args::SyntaxMode` and `syntax::SyntaxMode`.

Depending on how significant you consider that friction, offering a `--mode` argument (defaulting to `"code"`) could still be reasonable—especially since it mirrors the `#eval(...)` behavior from inside Typst, making the CLI more familiar for existing users.

## TL;DR

**`--scope`**
**+ Strong yes** — significant flexibility, meaningful difference from `--input`, powerful shell integration.

**`--mode`**
**\~ Mild benefit, low maintenance cost** — can reasonably go either way depending on project values and philosophy.

laurmaedje:
Thanks for the your detailed response. It's worth noting that the repetition could also be worked around like this:

```bash
typst eval "{ let x = $(echo 2); x + x + 2 }"
```

However, this easily makes for a code injection opportunity! I'm not sure what attack surface we're considering for `typst eval` yet. But having such an injection hazard is not great.

This is less bad with scope, _but_ I think it's still somewhat problematic and could be a pitfall. Something like `--scope x="$(cat user-input)"` could be attacked with `" + read("secret/file") + "`. And even if you don't have an attacker, you'd have to deal with Typst's escape sequences in strings.

Of course, you can avoid this by using `sys.inputs` if you want a string. But the convenience of `--scope` will make it the first thing people reach for.

Considering that 
- `sys.inputs` avoids such security problems by default
- you can still use `eval(sys.inputs.x)` explicitly within your code to emulate `--scope`

it might be a safer and better default. If you're expecting integers, I think it's also reasonable to be explicit about that and use `int` for parsing the inputs.

There is also the question of whether `--scope` would interact with `--mode`. (In the built-in `eval` function this is not relevant because the dictionary always contains values.)

TimerErTim:
Good point. I honestly did not think about potential attack surface on this one, but given your explanation, I agree that `--input` is enough and users can optin to risky input by using `eval(sys.inputs.<whatever>)`.

The repition mitigation with `{ let x = eval(sys.inputs.x); x + x + 2 }` works, but I feel like this is a lot of boilerplate for something we easily can have with the current design. Especially considering multiple `--scope`/`--input` arguments. I wonder if there could be a cleaner/less verbose/less boilerplaty way? 🤔

## Summary

Given your arguments:
- I agree with removing `--scope` in favor of the safer `--inputs` alternative.
- Maybe there is a better way for defining variables in the scope of the evaluated statement? 
  Your suggestion (`let` binding) seems to fit the Typst design philosophy perfectly though, so there may very well be no further room for improvement.
- I assume the `--mode` argument is off the table then?

TimerErTim:
> There is also the question of whether `--scope` would interact with `--mode`. (In the built-in `eval` function, this is not relevant because the dictionary always contains values.)

I think (and that is how I implemented it for now) `--mode` should only affect the received expression. `--scope` is unaffected by this and always evaluated as code. The reason for this is behaviour parity with the Typst `#eval()` function, where values in the scope Dict are always interpreted as code (unless put into math using `$...$` or content `[...]` mode, same as in CLI).

laurmaedje:
I'd say lets remove both `--scope` and `--mode` then. The downsides of `--scope` seem significant and after all it's always easier to add something later if we reconsider it than to remove it.

TimerErTim:
Done

crates/typst-cli/src/args.rs
```diff
@@ -178,6 +181,40 @@ pub struct QueryCommand {
...
+/// Evaluates a piece of Typst code.
+#[derive(Debug, Clone, Parser)]
+pub struct EvalCommand {
+    /// The piece of Typst code to evaluate. Use `-` to read input from stdin.
+    #[clap(value_parser = string_input_value_parser())]
+    pub expression: StringInput,
+
+    /// The file to evaluate the code in. Can be used to introspect the document.
+    #[clap(long = "in", value_hint = ValueHint::FilePath)]
+    pub r#in: Option<PathBuf>,
```
laurmaedje: Didn't notice this before, but this should probably also support `-` for stdin.

crates/typst-cli/src/args.rs
```diff
@@ -178,6 +181,40 @@ pub struct QueryCommand {
     pub process: ProcessArgs,
 }
 
+/// Evaluates a piece of Typst code.
+#[derive(Debug, Clone, Parser)]
+pub struct EvalCommand {
+    /// The piece of Typst code to evaluate. Use `-` to read input from stdin.
+    #[clap(value_parser = string_input_value_parser())]
+    pub expression: StringInput,
```
laurmaedje: Meanwhile, I just realized that this should probably _not_ support `-` as normal shell interpolation suffices here. For the path, it's different because using shell interpolation wouldn't work (as the data isn't the path).

Does that make sense? I'm not too familiar with shell conventions, but it would make sense if `-` is only used as a stand-in for paths.

I'm sorry for only noticing this just now after proposing the feature in the first place and nitpicking the string input parser.

laurmaedje (commented):
I made a few minor changes myself. I also noticed one more thing just now and left two comments. Other than that, this looks ready to merge.

crates/typst-cli/src/args.rs
TimerErTim: Hey, all right, no worries :)

A quick back and forth between ChatGPT and me showed that `-` for the statement is indeed uncommon practice and is usually reserved for file paths. I will change that now

TimerErTim:
Thanks a lot for the help along the way and the documentation/messages improvement from your side.

laurmaedje:
Thank you, great work!

PR #7363, opened 2025-11-12
Bump MSRV to 1.89

saecki:
This is mainly for #7278

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.88.0
+      - uses: dtolnay/rust-toolchain@1.89.0
       - uses: Swatinem/rust-cache@v2
       - run: cargo check --workspace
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@1.91.0
+    - uses: dtolnay/rust-toolchain@1.91.1
```
laurmaedje: do we need this? it's now out of sync with the non-MSRV ci workflow.

saecki: Oh right I missed the other one.
But I thought we might as well use the bug fix release: https://releases.rs/docs/1.91.1/

laurmaedje: the flake also specifies the version 😅

if you want to upgrade, we would also need to upgrade there. after editing the file you need to run `nix flake update rust-manifest`.

The Dockerfile does not seem to specify a version at all actually... but that's a separate issue and we otherwise didn't update it either.

saecki: Ok, I'll just update the MSRV then :)

PR #7371, opened 2025-11-13
Improve error for incomplete datetime

### LINKED ISSUE #7321: Improve error for incomplete datetime
### Description

Constructing a datetime without the full set of either `year+month+day` or `hour+minute+second` produces an unhelpful error.

Just `#datetime()` gives: *"Error: at least one of date or time must be fully specified"*
But adding year like `#datetime(year: 2025)`, gives a less helpful error: *"date is incomplete"*
And then adding an hour like `#datetime(year: 2025, hour: 1)`  changes over to: *"time is incomplete"*

It would be nice if we could give some hints guiding the user to a valid datetime, e.g. for `#datetime(year: 2025)`, produce a hint: *"add `month` and `day` fields to get a valid date"*

This is somewhat related to #4107

### Typst version

RungeCC:
May close #7321.

I'm not sure if this coding style is consistent with the rest of `typst`, let me know if I messed up.

Some notes about this PR:

- `format_missing_fields` currently does not assume at most two missing fields, although this is the case for datetime. Perhaps our diagnostic messages elsewhere might also use it so it shall be move to somewhere outside? I'm not sure.
- I'm not sure if we should issue both warnings for incomplete `date` and `time` in examples like `datetime(year: 10, hour: 12)`, as this would also alter the error reporting logic, this PR just keeps unchanged.

Any suggestions are welcome.

### DIFF
--- a/crates/typst-library/src/foundations/datetime.rs
+++ b/crates/typst-library/src/foundations/datetime.rs
@@ -2,6 +2,7 @@ use std::cmp::Ordering;
 use std::hash::Hash;
 use std::ops::{Add, Sub};
 
+use arrayvec::ArrayVec;
 use ecow::{EcoString, EcoVec, eco_format};
 use time::error::{Format, InvalidFormatDescription};
 use time::macros::format_description;
@@ -275,16 +276,17 @@ impl Datetime {
         #[named]
         second: Option<u8>,
     ) -> HintedStrResult<Datetime> {
-        fn format_missing_arguments(args: Vec<&str>) -> EcoString {
+        fn format_missing_args(args: ArrayVec<&str, 3>) -> EcoString {
             match args.as_slice() {
-                [] => EcoString::new(),
-                [arg] => eco_format!("argument {arg}"),
-                [arg1, arg2] => eco_format!("arguments {arg1} and {arg2}"),
+                [] => unreachable!(),
+                [arg] => eco_format!("the {arg} argument"),
+                [arg1, arg2] => eco_format!("the {arg1} and {arg2} arguments"),
                 [args @ .., tail] => {
-                    eco_format!("arguments {}, and {tail}", args.join(", "))
+                    eco_format!("the {}, and {tail} arguments", args.join(", "))
                 }
             }
         }
+
         let time = match (hour, minute, second) {
             (Some(hour), Some(minute), Some(second)) => {
                 match time::Time::from_hms(hour, minute, second) {
@@ -294,16 +296,15 @@ impl Datetime {
             }
             (None, None, None) => None,
             (hour, minute, second) => {
-                let time_args = [
+                let args = [
                     if hour.is_none() { Some("`hour`") } else { None },
                     if minute.is_none() { Some("`minute`") } else { None },
                     if second.is_none() { Some("`second`") } else { None },
                 ];
-                let missing_args =
-                    format_missing_arguments(time_args.into_iter().flatten().collect());
                 bail!(
                     "time is incomplete";
-                    hint: "add {missing_args} to get a valid time"
+                    hint: "add {} to get a valid time",
+                    format_missing_args(args.into_iter().flatten().collect())
                 )
             }
         };
@@ -317,16 +318,15 @@ impl Datetime {
             }
             (None, None, None) => None,
             (year, month, day) => {
-                let date_args = [
+                let args = [
                     if year.is_none() { Some("`year`") } else { None },
                     if month.is_none() { Some("`month`") } else { None },
                     if day.is_none() { Some("`day`") } else { None },
                 ];
-                let missing_args =
-                    format_missing_arguments(date_args.into_iter().flatten().collect());
                 bail!(
                     "date is incomplete";
-                    hint: "add {missing_args} to get a valid date"
+                    hint: "add {} to get a valid date",
+                    format_missing_args(args.into_iter().flatten().collect())
                 )
             }
         };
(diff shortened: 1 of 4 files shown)


crates/typst-library/src/foundations/datetime.rs
```diff
@@ -274,7 +274,29 @@ impl Datetime {
...
+                0 => String::new(),
+                1 => format!("{} field", fields[0]),
+                _ => format!(
+                    "{} and {} fields",
+                    fields[0..len - 1].join(", "),
+                    fields[len - 1]
+                ),
+            }
+        }
+        macro_rules! missing_fields_from {
```
xkevio: I get why but I feel like a macro for this is overkill. Also, I don't know if that is just a general styling convention but usually even local macros live outside of functions (since they aren't available outside anyway unless `#[macro_export]` is used.

crates/typst-library/src/foundations/datetime.rs
```diff
@@ -274,7 +274,29 @@ impl Datetime {
         /// The second of the datetime.
         #[named]
         second: Option<u8>,
-    ) -> StrResult<Datetime> {
+    ) -> HintedStrResult<Datetime> {
+        fn format_missing_fields(fields: Vec<&str>) -> String {
```
xkevio: If we stay with the `Vec` approach, I suppose it is nicer if this is a `&[&str]`.

crates/typst-library/src/foundations/datetime.rs
RungeCC: It seems depending on whether we want to reuse it, like the following macro `layout` is inside function `layout_attach`.

https://github.com/typst/typst/blob/57f78a678b380a58085f562e32a0af05f2896a00/crates/typst-layout/src/math/attach.rs#L60-L66

If we want to reuse this functionality somewhere else, I will pick it out together with local function `format_missing_fields`.

It's not too complicated to avoid using macros here (with only a few `fields`), and I agree we can switch to a more straightforward approach.

crates/typst-library/src/foundations/datetime.rs
RungeCC: Em, It seems that here after `flatten`, the size of `[&str]` can no longer be known at compile time, so we need to use a length-variable container here.

crates/typst-library/src/foundations/datetime.rs
```diff
@@ -274,7 +274,29 @@ impl Datetime {
...
+            let len = fields.len();
+            match len {
+                0 => String::new(),
+                1 => format!("{} field", fields[0]),
+                _ => format!(
+                    "{} and {} fields",
+                    fields[0..len - 1].join(", "),
+                    fields[len - 1]
+                ),
+            }
```
fmease:
```suggestion
            match fields.as_slice() {
                [] => String::new(),
                [field] => format!("{field} field"),
                [fields @ .., field] => format!(
                    "{} and {field} fields",
                    fields.join(", "),
                ),
            }
```

RungeCC: Thank you!

RungeCC:
I've updated this PR accordingly, macro is removed.

crates/typst-library/src/foundations/datetime.rs
```diff
@@ -274,7 +274,16 @@ impl Datetime {
...
         #[named]
         second: Option<u8>,
-    ) -> StrResult<Datetime> {
+    ) -> HintedStrResult<Datetime> {
+        fn format_missing_fields(fields: Vec<&str>) -> String {
+            match fields.as_slice() {
+                [] => String::new(),
+                [field] => format!("{field} field"),
+                [fields @ .., field] => {
+                    format!("{} and {field} fields", fields.join(", "))
```
laurmaedje: I'd like to have an Oxford comma before "and" in the case of three fields. Probably, it's easiest to just match on all cases then.

crates/typst-library/src/foundations/datetime.rs
```diff
@@ -274,7 +274,16 @@ impl Datetime {
         /// The second of the datetime.
         #[named]
         second: Option<u8>,
-    ) -> StrResult<Datetime> {
+    ) -> HintedStrResult<Datetime> {
+        fn format_missing_fields(fields: Vec<&str>) -> String {
```
laurmaedje: This should return an `EcoString` and use `eco_format!`

crates/typst-library/src/foundations/datetime.rs
```diff
@@ -274,7 +274,16 @@ impl Datetime {
...
+            match fields.as_slice() {
+                [] => String::new(),
+                [field] => format!("{field} field"),
```
laurmaedje: technically, this should probably not say "field", but "argument"

crates/typst-library/src/foundations/datetime.rs
RungeCC: Ah, I've tried but how to do? IIUC, that means we would not use `hint` here, since a `HintedString` can not be converted into an `EcoString`.

crates/typst-library/src/foundations/datetime.rs
RungeCC: Em, currently actually there're at most 2 missing arguments, since otherwise will always fall into `(None, None)` branch with message `"at least one of date or time must be fully specified"`,   shall I also add hints for it?

crates/typst-library/src/foundations/datetime.rs
RungeCC: Oh, this is referring to line 278.

RungeCC:
I've updated this PR accordingly. I also add hints for `(None, None)` branch; tests are updated accordingly.

I'm not sure shall I mark `unreachable` for actual unreachable arms (`args.len() == 0` or `args.len() >= 3`).

laurmaedje:
Thanks!

> I'm not sure shall I mark unreachable for actual unreachable arms

Indeed the empty case is better marked as unreachable I think. The case > 3 does not need that as the code _would_ produce sensible results. I made some minor touch-ups myself anyway and made the change alongside these.

isuffix:
Thank you for this!

Issue #7253, opened 2025-10-29
`columns` could be better documented


CarloLucibello:
### Description

The `columns` docstring could provide a usage example (there are none at the moment) and mention the `colbreak` function.

A possible example could be
```typst
#columns[$
  h^1 &= phi(W^1 x + b^1)\
  h^2 & = phi(W^2 h^1 + b^2)\
  ...\
  h^L &= phi(W^L h^(L-1) + b^L)\
$
#colbreak()
 Some image
]
```

cady-b:
These things are mentioned in the guides section of the docs, which go more in-depth on how different elements interact; specifically the Page Setup Guide: https://typst.app/docs/guides/page-setup/#columns

But I agree that `columns` should have better documentation as well.  I assume that's been lacking because there are still some open issues around how they work, but I'm not sure when that'll be resolved.

If someone would like to take on getting the docs improved in the meantime, I think the [Typst Discord](https://typst.app/open-source#support) documentation forge would be the place to discuss that.

laurmaedje:
Closed by https://github.com/typst/typst/pull/7372

PR #7372, opened 2025-11-13
Add documentation for column breaks in multi-column layout example

Ri-Nai:
related to https://github.com/typst/typst/issues/7253


laurmaedje:
Thanks!

PR #7374, opened 2025-11-13
Setup LD_LIBRARY_PATH in flake.nix

oliverpool:
To test a PR, I cloned the repo and used nix to setup a dev shell.

Everything compiled fine, however running the executable yields:
```
error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
```

[Setting the `LD_LIBRARY_PATH` env var](https://discourse.nixos.org/t/program-compiled-with-rust-cannot-find-libssl-so-3-at-runtime/27196) fixes this issue (when running inside the devShell).

I wasn't able to set [`OPENSSL_NO_VENDOR = true;`](https://github.com/NixOS/nixpkgs/blob/c2448301fb856e351aab33e64c33a3fc8bcf637d/pkgs/by-name/ty/typst/package.nix#L37C5-L37C22) so that the binary does not load the libssl.so...

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -152,6 +152,7 @@
             ];
 
             RUST_SRC_PATH = "${rust-toolchain.rust-src}/lib/rustlib/src/rust/library";
+            LD_LIBRARY_PATH = lib.makeLibraryPath [ pkgs.openssl ];
 
             packages = [
               # A script for quickly running tests.


frozolotl:
This shouldn't need to be necessary. I believe something is wrong with your setup.

Perhaps you configured cargo to use `mold` as the linker, instead of using `mold-wrapped`?

oliverpool:
> I believe something is wrong with your setup.

I am using the devShell. rust (and cargo) are not available/configured globally.

```
~/f►cd typst                                                                                                                                                                       15:57
direnv: loading ~/f/typst/.envrc
direnv: using flake
direnv: nix-direnv: Using cached dev shell
~/f/typst►cargo run c ../typst-converge-bug/main.typ                                                                                                                        (main) 15:57
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/typst c ../typst-converge-bug/main.typ`
target/debug/typst: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
~/f/typst[127]►
```

laurmaedje:
@drupol As resident Nix expert, do you have an idea what's going on?

drupol:
OK I can reproduce the issue by doing:

```
cargo run c ../typst-converge-bug/main.typ
```

```
   Compiling typst v0.14.0 (/home/pol/Code/tmp/typst/crates/typst)
   Compiling typst-kit v0.14.0 (/home/pol/Code/tmp/typst/crates/typst-kit)
   Compiling tempfile v3.16.0
   Compiling sigpipe v0.1.3
   Compiling shell-escape v0.1.5
   Compiling fs_extra v1.3.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4m 57s
     Running `target/debug/typst c ../typst-converge-bug/main.typ`
target/debug/typst: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory

typst is 📦 v0.14.0 via 🦀 v1.91.0 via ❄  impure (nix-shell-env) [w c | (empty) (no description) main] took 4m58s 
```

drupol:
I also asked in the Nix Rust room and I got this answer from @alyssais:

```
Alyssa: Usually I'd consider that a last resort after actually linking the library to the binary
Pol: What other option do we have here?
Alyssa: `CARGO_TARGET_${stdenv.hostPlatform.rust.cargoEnvVarTarget}_RUSTFLAGS = "-lssl"`
Alyssa: but the 100% best way to do this is to fix upstream to properly link the library
```

laurmaedje:
Can't we use the existing crate feature that vendors OpenSSL?

oliverpool:
Other workarounds (seem to work independently):

## Manually activate the `vendor-openssl` feature

```sh
cargo run --features vendor-openssl c ../typst-converge-bug/main.typ
```

## Enable the `vendored` feature of `openssl` in typst-kit

https://github.com/typst/typst/blob/a67dc63461c8f702a72d99bb8d8188efac558693/crates/typst-kit/Cargo.toml#L35


```toml
openssl = { workspace = true, optional = true, features = ["vendored"] }
```

My rust and nix experience is to limited to know which solution is better.

drupol:
Is this crate using a "vendored" `openssl` ?

laurmaedje:
> Enable the vendored feature of openssl in typst-kit

It should not be enabled by default outside of Nix. It's intentionally opt-in.

alerque:
@oliverpool I don't think either of those other work-arounds make any sense, the dev shell provides the Nix system openssl already just like it is supplied for the Flake, and I don't see any reason why it should be vendored by default there any more than anywhere else.

I'm also unable to replicate this problem on several systems using the `nix develop` shell (not on NixOS, just a Arch Linux host system). I suspect you have something (env vars?) setup in your system that are overriding the system and Flake defaults.

laurmaedje:
I'll close this since https://github.com/typst/typst/pull/7512 is merged.

Issue #7219, opened 2025-10-24
Overhang disabled if word has label


bluss:
### Description

Typst version: 0.14.0

If I label a word like `just,` then punctuation overhang is disabled for the relevant comma. It's expected that adding a label doesn't change the layout.

```typ
#set par(justify: true)
Regular:
#block(stroke: 0.5pt + blue, width: 4cm)[This is a text with just, the right length.]
Label the word:
#block(stroke: 0.5pt + blue, width: 4cm)[This is a text with just,<label> the right length.]
```

<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/8a836dc1-de46-4bb4-856b-83682859d0ac" />

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

PR #7376, opened 2025-11-13
Fix punctuation overhang after labeled words

Ri-Nai:
related to https://github.com/typst/typst/issues/7219

- skip zero-advance glyphs when finding the first/last glyph for line overhang
- restore comma and dash overhang when a preceding word injects label-driven zero-width glyphs
- keep existing behavior for lines that already start or end with regular glyphs


crates/typst-layout/src/inline/line.rs
```diff
@@ -719,16 +717,26 @@ impl<'a> Items<'a> {
...
-
     /// Access the first item (skipping tags) mutably, if it is text.
     pub fn leading_text_mut(&mut self) -> Option<&mut ShapedText<'a>> {
         self.0.iter_mut().find(|(_, item)| !item.is_tag())?.1.text_mut()
     }
 
+    /// Access the first glyph with non-zero advance in the line.
+    pub fn leading_glyph(&self) -> Option<(&ShapedText<'a>, &ShapedGlyph)> {
+        for (_, item) in self.0.iter() {
+            let Item::Text(text) = &**item else { continue };
```
laurmaedje: This will skip over non-tag non-text items, which shouldn't be the case.

crates/typst-layout/src/inline/line.rs
```diff
@@ -719,16 +717,26 @@ impl<'a> Items<'a> {
...
+            if let Some(glyph) = text
+                .glyphs
+                .iter()
+                .find(|glyph| glyph.x_advance.at(glyph.size) > Abs::zero())
```
laurmaedje: Why exactly zero-width glyphs. Could you elaborate on this condition?

Ri-Nai:
Thanks for looking! This change is scoped specifically to the regression where labels insert a zero-advance glyph between the word and the punctuation, causing the overhang logic to "miss" the comma.

To verify exactly what was happening in the renderer, I added some temporary debug printing to inspect the frame structure for the example.

```typst
#set par(justify: true)
Regular:
#block(stroke: 0.5pt + blue, width: 4cm)[This is a text with just, the right length.]
Label the word:
#block(stroke: 0.5pt + blue, width: 4cm)[This is a text with just,<label> the right length.]
```

Without a label, we get two clean text nodes:
```text
==== baseline ====
  Text "This is a text with just, " (glyphs: 26)
  Text "the right length." (glyphs: 17)
```

With a `<label>` present, the frame contains the tag pair plus a **standalone space** which gets collapsed:
```text
==== labelled ====
  Tag Start("text", …)
  Text "This is a text with just," (glyphs: 25)
  Tag End(…)
  Text " " (glyphs: 1)
  Text "the right length." (glyphs: 17)
```

That lone space glyph (the `Text " "` above) is kept for copy/paste support, but its advance is collapsed to zero (trimmed during `collect_range` / `ShapedText::build`). Because the tag mechanism splices it between the word and the comma, the old overhang logic stopped on that zero-width glyph and never saw the punctuation.

The new helper skips such glyphs to find the comma again. I agree that longer-term we need a broader tag story, but this unblocks the overhang regression without touching other inline layout code.

---
> This will skip over non-tag non-text items, which shouldn't be the case.

I’ve tightened both helpers so they now `break` immediately at the first non-tag/non-text item. This matches the previous semantics while still skipping the zero-advance glyphs caused by tags. I've pushed the update.

---

> Why exactly zero-width glyphs. Could you elaborate on this condition?

This is directly related to the frame structure I described in the general comment. The "zero-width glyphs" here are precisely those artifacts (like the collapsed space glyph in `Text " "`) that appear at the boundary due to the tag split.

Since `ShapedText::build` zeroes their advance (to keep them invisible but copyable), the old logic saw them as the "leading/trailing glyph" and returned a width of 0, effectively blocking the overhang check. The condition `x_advance > 0` allows the iterator to skip these invisible artifacts and land on the actual punctuation (the comma) to correctly calculate its overhang.

Enivex:
Any particular reason why you closed your PR?

Ri-Nai:
> Any particular reason why you closed your PR?

It's been a few months with no activity, and I unfortunately no longer have the time to maintain or update this PR. I also felt the implementation might still need some work. I decided to close it for now, but anyone is welcome to continue where I left off."

PR #7386, opened 2025-11-15
Allow the integer constructor to parse different bases

### LINKED ISSUE #6851: Integer constructor can't parse different bases from string
### Description

When passing a string to Integer's constructor, it can't parse anything other than base 10 as it uses the default i64 `from_str` function.

I would expect it to parse anything that is a valid Integer literal.

### Use Case

Probably the same as having a string to integer constructor but now any integer literal could be parsed so basically feature parity with integer literals.

I'm happy to try implementing this if there aren't any issues with the idea.

CrumblyLiquid:
Fixes #6851

This PR adds a new `base` argument for the integer constructor which allows it to parse strings with any base between 2 and 36.

So now you can do for example `#int("ff", base: 16)`

This should not be a breaking change since the default base is 10.

I'm not sure about the wording of the errors and I also use the `Base` struct from `str.rs` - maybe it should be moved since it's now used in multiple places?

### DIFF
--- a/crates/typst-library/src/foundations/int.rs
+++ b/crates/typst-library/src/foundations/int.rs
@@ -2,11 +2,11 @@ use std::num::{
     IntErrorKind, NonZeroI64, NonZeroIsize, NonZeroU32, NonZeroU64, NonZeroUsize,
 };
 
-use ecow::{EcoString, EcoVec, eco_format, eco_vec};
+use ecow::{EcoString, eco_format};
 use smallvec::SmallVec;
 use typst_syntax::{Span, Spanned};
 
-use crate::diag::{SourceDiagnostic, SourceResult, StrResult, bail, error};
+use crate::diag::{At, HintedString, SourceResult, StrResult, bail, error};
 use crate::foundations::{
     Base, Bytes, Cast, Decimal, Repr, Str, Value, cast, func, repr, scope, ty,
 };
@@ -85,22 +85,24 @@ impl i64 {
                         // Parse the digits part into u64
                         //  => abs(i64::MIN) fits into u64
                         let bigger = u64::from_str_radix(s, radix)
-                            .map_err(|e| parse_str_error(e.kind(), base, value.span))?;
+                            .map_err(|e| parse_str_error(e.kind(), base))
+                            .at(value.span)?;
 
                         // Number wouldn't fit into i64
                         if bigger > i64::MIN.unsigned_abs() {
                             return Err(parse_str_error(
                                 &IntErrorKind::NegOverflow,
                                 base,
-                                value.span,
-                            ));
+                            ))
+                            .at(value.span);
                         }
 
                         bigger.wrapping_neg() as i64
                     }
                     // Positive
                     None => i64::from_str_radix(&s, radix)
-                        .map_err(|e| parse_str_error(e.kind(), base, value.span))?,
+                        .map_err(|e| parse_str_error(e.kind(), base))
+                        .at(value.span)?,
                 }
             }
         })
@@ -430,39 +432,27 @@ pub fn convert_float_to_int(f: f64) -> StrResult<i64> {
 }
 
 #[cold]
-fn parse_str_error(
-    kind: &IntErrorKind,
-    base: Spanned<Base>,
-    span: Span,
-) -> EcoVec<SourceDiagnostic> {
+fn parse_str_error(kind: &IntErrorKind, base: Spanned<Base>) -> HintedString {
     let base = base.v.value();
-    let error = match kind {
-        IntErrorKind::Empty if base == 10 => {
-            error!(span, "expected an integer string (got an empty string)")
-        }
-        IntErrorKind::Empty => {
-            error!(span, "expected a base {} integer string (got an empty string)", base)
-        }
-        IntErrorKind::InvalidDigit if base == 10 => {
-            error!(span, "invalid digits for an integer")
-        }
-        IntErrorKind::InvalidDigit => {
-            error!(span, "invalid digits for a base {} integer", base)
-        }
+    match kind {
+        IntErrorKind::Empty => error!("string must not be empty"),
+        IntErrorKind::InvalidDigit => match base {
+            10 => error!("string contains invalid digits"),
+            _ => error!("string contains invalid digits for a base {base} integer"),
+        },
         IntErrorKind::PosOverflow => error!(
-            span, "invalid integer string";
-            hint: "number is too large to fit in a signed 64-bit integer";
+            "integer value is too large";
+            hint: "value does not fit into a signed 64-bit integer";
             hint: "try using a floating point number"
         ),
         IntErrorKind::NegOverflow => error!(
-            span, "invalid integer string";
-            hint: "number is too small to fit in a signed 64-bit integer";
+            "integer value is too small";
+            hint: "value does not fit into a signed 64-bit integer";
             hint: "try using a floating point number"
         ),
         IntErrorKind::Zero => unreachable!(),
-        _ => error!(span, "invalid integer string"),
-    };
-    eco_vec![error]
+        _ => error!("invalid integer"),
+    }
 }
 
 macro_rules! signed_int {
--- a/tests/suite/foundations/int.typ
+++ b/tests/suite/foundations/int.typ
@@ -37,19 +37,19 @@
 #int(10pt)
 
 --- int-constructor-str-empty paged ---
-// Error: 6-8 expected an integer string (got an empty string)
+// Error: 6-8 string must not be empty
 #int("")
 
 --- int-constructor-str-empty-based paged ---
-// Error: 6-8 expected a base 16 integer string (got an empty string)
+// Error: 6-8 string must not be empty
 #int("", base: 16)
 
 --- int-constructor-bad-value paged ---
-// Error: 6-12 invalid digits for an integer
+// Error: 6-12 string contains invalid digits
 #int("nope")
 
 --- int-constructor-bad-value-based paged ---
-// Error: 6-11 invalid digits for a base 3 integer
+// Error: 6-11 string contains invalid digits for a base 3 integer
 #int("123", base: 3)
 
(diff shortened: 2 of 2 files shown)


crates/typst-library/src/foundations/int.rs
```diff
@@ -42,26 +43,73 @@ type i64;
...
+                // Replace minus sign at the start with `-`
+                let s: &str = if s.as_str().starts_with(repr::MINUS_SIGN) {
+                    &s.as_str().replacen(repr::MINUS_SIGN, "-", 1)
+                } else {
+                    &s
+                };
+                match std::primitive::i64::from_str_radix(s, b.try_into().unwrap()) {
+                    Ok(n) => n,
+                    Err(e) => {
+                        let error = match e.kind() {
```
laurmaedje: could you extract the error formatting into a separate function? you can look across the code base for similar examples so that it's consistent.

crates/typst-library/src/foundations/int.rs
```diff
@@ -42,26 +43,73 @@ type i64;
...
+                            }
+                            IntErrorKind::PosOverflow => {
+                                "the integer is too large to fit into a signed 64-bit integer"
+                            }
+                            IntErrorKind::NegOverflow => {
+                                "the integer is too small to fit into a signed 64-bit integer"
+                            }
+                            // This shouldn't ever happen :)
+                            IntErrorKind::Zero => "zero couldn't be represented",
+                            _ => "unknown error",
```
laurmaedje: for things that are truly unreachable at runtime, we can also use `unreachable!()`. will it truly never happen?

crates/typst-library/src/foundations/int.rs
```diff
@@ -42,26 +43,73 @@ type i64;
...
+                let s: &str = if s.as_str().starts_with(repr::MINUS_SIGN) {
+                    &s.as_str().replacen(repr::MINUS_SIGN, "-", 1)
+                } else {
+                    &s
+                };
+                match std::primitive::i64::from_str_radix(s, b.try_into().unwrap()) {
+                    Ok(n) => n,
+                    Err(e) => {
+                        let error = match e.kind() {
+                            IntErrorKind::Empty => "can't parse an empty string",
```
laurmaedje: this error message is phrased differently from the others. It speaks in terms of behavior of the machine (_can't parse_) while the others state something about the integers. would be great if you could make this more consistent.

crates/typst-library/src/foundations/int.rs
```diff
@@ -42,26 +43,73 @@ type i64;
...
+                n
+            }
+            ToInt::Str(s) => {
+                let b = base.v.value();
+                if !(2..=36).contains(&b) {
+                    bail!(base.span, "base must be between 2 and 36");
+                }
+
+                // Replace minus sign at the start with `-`
+                let s: &str = if s.as_str().starts_with(repr::MINUS_SIGN) {
```
laurmaedje: this is a bit inefficient. could you we do the sign handling separately to avoid the allocation? (I presume we might run into issues with max/min integers like we did before?)

crates/typst-library/src/foundations/int.rs
isuffix: `unreachable!()` should be fine. Docs say that variant is just for NonZero kinds:
https://doc.rust-lang.org/std/num/enum.IntErrorKind.html#variant.Zero

CrumblyLiquid: Should the `otherwise` case also be marked as `unreachable!()`? `IntErrorKind` is marked as non-exhaustive so there might be more cases added in the future. If that happened would `unreachable!()` be okay, or should it say something more specific? I did the second option but I'm not sure.

crates/typst-library/src/foundations/int.rs
CrumblyLiquid: Yeah, I thought about it then and couldn't come up with a good solution, not sure why I didn't consider parsing into a bigger type (parse into the bigger type, check bounds and then cast into the smaller type). This should be now fixed and it doesn't require special check for different bases.

crates/typst-library/src/foundations/int.rs
isuffix: No it shouldn't, that's kind of the point of `non_exhaustive`. If the std lib adds a new error kind, we should be able to handle it gracefully.

crates/typst-library/src/foundations/int.rs
isuffix: @laurmaedje what about this? It still may allocate, but only if we're starting from a negative int generated within Typst that's more than 15 chars (more than 19 would start to error). IMO cleaning up the code is worth it.
```rust
let mut s: EcoString = s.into();
if s.starts_with(repr::MINUS_SIGN) {
    s = s.replacen(repr::MINUS_SIGN, "-", 1);
}
i64::from_str_radix(&s, radix)
    .map_err(|e| parse_str_error(e.kind(), value.span))?
```

crates/typst-library/src/foundations/int.rs
```diff
@@ -42,26 +43,66 @@ type i64;
...
+    ) -> SourceResult<i64> {
+        Ok(match value.v {
+            ToInt::Int(n) => match base.v {
+                Base::User(_) => bail!(base.span, "base is only supported for strings"),
+                _ => n,
+            },
+            ToInt::Str(s) => {
+                let base_value = base.v.value();
+                let radix: u32 = match base_value {
+                    2..=36 => base_value.try_into().unwrap(),
```
isuffix:
```suggestion
                    2..=36 => base_value as u32
```

crates/typst-library/src/foundations/int.rs
```diff
@@ -384,16 +428,22 @@ pub fn convert_float_to_int(f: f64) -> StrResult<i64> {
...
+            "the integer is too large to fit into a signed 64-bit integer"
+        }
+        IntErrorKind::NegOverflow => {
+            "the integer is too small to fit into a signed 64-bit integer"
+        }
+        IntErrorKind::Zero => unreachable!(),
+        _ => "unhandled number parsing error",
+    };
+
+    eco_vec![SourceDiagnostic::error(span, eco_format!("invalid integer: {error}"))]
```
isuffix: This is an improvement, but I think it could be worded better by using hints instead of repeating the error after "invalid integer." We can also pass in the `base` to get a better error for non-decimal bases.

Here's an idea for what that might look like:
(feel free to use this and edit if you dislike any of the wordings)
```rust
#[cold]
fn parse_str_error(
    kind: &IntErrorKind,
    base: Base,
    span: Span,
) -> EcoVec<SourceDiagnostic> {
    let error = match kind {
        IntErrorKind::Empty if base.value() == 10 => error!(span, "expected an integer"),
        IntErrorKind::Empty => error!(span, "expected a base {} integer", base.value()),
        IntErrorKind::InvalidDigit if base.value() == 10 => {
            error!(span, "invalid digits for an integer")
        }
        IntErrorKind::InvalidDigit => {
            error!(span, "invalid digits for a base {} integer", base.value())
        }
        IntErrorKind::PosOverflow => error!(
            span, "invalid integer";
            hint: "number is too large to fit in a signed 64-bit integer";
            hint: "try using a floating point number",
        ),
        IntErrorKind::NegOverflow => error!(
            span, "invalid integer";
            hint: "number is too small to fit in a signed 64-bit integer";
            hint: "try using a floating point number",
        ),
        IntErrorKind::Zero => unreachable!(),
        _ => error!(span, "invalid integer"),
    };
    eco_vec![error]
}
```

An example diff of the test cases would look like:
```diff
 --- int-constructor-str-empty paged ---
-// Error: 6-8 invalid integer: empty string isn't a valid integer
+// Error: 6-8 expected an integer
 #int("")
 
 --- int-constructor-str-empty-base-16 paged ---
-// Error: 6-8 invalid integer: empty string isn't a valid integer
+// Error: 6-8 expected a base 16 integer
 #int("", base: 16)
 
 --- int-constructor-bad-value paged ---
-// Error: 6-12 invalid integer: invalid digit found while parsing integer
+// Error: 6-12 invalid digits for an integer
 #int("nope")
 
 --- int-constructor-bad-value-base-8 paged ---
-// Error: 6-11 invalid integer: invalid digit found while parsing integer
+// Error: 6-11 invalid digits for a base 8 integer
 #int("888", base: 8)
 
 --- int-constructor-str-too-large paged ---
-// Error: 6-27 invalid integer: the integer is too large to fit into a signed 64-bit integer
+// Error: 6-27 invalid integer
+// Hint: 6-27 number is too large to fit in a signed 64-bit integer
+// Hint: 6-27 try using a floating point number
 #int("9223372036854775808")
 
 --- int-constructor-str-too-small paged ---
-// Error: 6-28 invalid integer: the integer is too small to fit into a signed 64-bit integer
+// Error: 6-28 invalid integer
+// Hint: 6-28 number is too small to fit in a signed 64-bit integer
+// Hint: 6-28 try using a floating point number
 #int("-9223372036854775809")
```

isuffix (commented):
Thank you for working on this! I'm quite happy with this change, and I appreciate your eye for detail in the tests.

I just have a small style note and an idea for improving the test error messages.

crates/typst-library/src/foundations/int.rs
CrumblyLiquid: Fixed, thanks!

crates/typst-library/src/foundations/int.rs
CrumblyLiquid: Thanks! I changed the error messages a bit to mention "integer string" since "invalid integer" as an error seemed a bit misleading to me.

laurmaedje: I wasn't entirely happy with either of the proposed error messages because "invalid integer" sounds like an integer is expected, but "invalid integer string" sounds strange to me as well. I gave it another shot. I realize that this is all personal preference in the end, but I'm quite happy with the wording now and I think it's reasonably consistent with the rest of Typst.

laurmaedje:
Thanks!

PR #7389, opened 2025-11-16
Fix lack of fallback to float for large integer literals

### LINKED ISSUE #7388: Integer literals exceeding i64 range are parsed as 0 instead of float
### Description

> [A positive integer greater than the maximum value and a negative integer less than or equal to the minimum value cannot be represented as an integer literal, and are instead parsed as a float. ](https://typst.app/docs/reference/foundations/int/)

But 

```typst
#sys.version

#(9223372036854775807) + 1 = #(9223372036854775808) \
#(-9223372036854775807) - 1 = #(-9223372036854775808)  \
```

will result in

```
0.14.0
9223372036854775807 + 1 = 0
−9223372036854775807 - 1 = 0
```

Note: Tests indicate that in version `0.13.x`, integer literals exceeding the i64 range were correctly parsed as floats. 

### Reproduction URL

https://typst.app/project/rKXwApSkX5W9wsO4YznacB

### Operating system

_No response_

### Typst version

YDX-2147483647:
Fixes #7388
Relates-to: #5969

### DIFF
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -1,5 +1,6 @@
-use ecow::{EcoString, eco_format};
 use std::num::IntErrorKind;
+
+use ecow::{EcoString, eco_format};
 use typst_utils::default_math_class;
 use unicode_ident::{is_xid_continue, is_xid_start};
 use unicode_math_class::MathClass;


laurmaedje:
cc @isuffix who had opened https://github.com/typst/typst/pull/5969

laurmaedje:
I'm wondering whether we actually _want_ to keep this behavior. Perhaps, we should rather turn this into a hard error and require the dot to explicitly turn it into a float.

YDX-2147483647:
Actually, I also find the behaviour weird. There should be a warning at least, like the warning for `decimal(1.0)`.

- If the typst code is generated by some program, then it's more appropriate to generate json/cbor/...  and use data loading functions.
- If the typst code is written by hand… Oh I can't image how can that be. @kokic, do you have a real use case?
- Nevertheless, the v0.13.1 behaviour can be retained by `#let n = yaml(bytes("9223372036854775808"))`.

isuffix:
Apologies for the regression! I should've checked the behavior in `ast.rs` more carefully.

However, I do agree with Laurenz that this is weird. I would consider it an odd piece of trivia if `#type(1000000000000000000)` differed from `#type(10000000000000000000)` (there's an extra zero). I don't think erroring on a too-large integer (with hints to either add a dot or use `bytes`) would be an undue burden.

Another option discussed in #5462 is to just warn on this case, but I don't think changing the type to a float ever really makes sense, so I would prefer an error and require an explicit dot.

Separately, we should just add methods `int.min()` and `int.max()` (or maybe `int.MIN`?) to help with the negative minimum use-case.

laurmaedje:
Given the precedent of `float.nan`, I think it would just be `int.min`. Uppercase is not really a thing in Typst (which is sometimes annoying because it means variables unnecessarily collide with types ...)

laurmaedje:
Let's turn it into a hard error.

laurmaedje:
Actually, let's first merge the fix and then introduce the hard error in a separate PR. That way I could cherry-pick should we decide to cut a point release.

crates/typst-syntax/src/lexer.rs
```diff
@@ -1,4 +1,5 @@
 use ecow::{EcoString, eco_format};
+use std::num::IntErrorKind;
```
laurmaedje, line 1: could you move this into a separate import block?

laurmaedje:
Thanks! Feel free to open a second PR for introducing a hard error. :)

isuffix:
@YDX-2147483647 Did you want to open another PR for adding `int.min/max` and a hard error for this case? Otherwise I could do so.

YDX-2147483647:
I have no plan to do it. Please go ahead.

PR #7392, opened 2025-11-16
Fix `calc.quo` flooring for negative integers

### LINKED ISSUE #6157: `calc.quo` gives inconsistent output depending on input type
### Description

The calculation function `quo` returns different values when passing an `int` or `float` to it. 

```typ
#calc.quo(-4, 5)  // gives 0
#calc.quo(-4, 5.) // returns -1
```

Being defined as _floored division_ in the documentation, the first case returns a wrong value. 

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

andylizi:
Fix #6157.

This can be considered a breaking change because `calc.quo` was not implemented correctly for negative integers. The documentation promises `quo(a, b) = floor(a/b)`, but the old test suite contradicts this:

```typ
--- calc-quo ---
// Test the `quo` function.
#test(calc.quo(1, 1), 1)
#test(calc.quo(5, 3), 1)
#test(calc.quo(5, -3), -1)  // ❌ should be -2 because `floor(5 / -3) = floor(-1.6) = -2`.
```

The old behavior on integers is truncated division (rounding toward zero), not floored division.


crates/typst-library/src/foundations/calc.rs
```diff
@@ -1035,7 +1035,15 @@ pub fn quo(
     let divided = dividend
         .apply2(
             divisor.v,
-            |a, b| Some(DecNum::Int(a / b)),
+            |a, b| {
+                let q = a.checked_div(b)?;
```
PgBiel, line 1050: Turns out the division by zero check above wasn't enough. This currently crashes the compiler: 

```typ
#calc.quo(int("-9223372036854775808"), -1)
```

Surprisingly, the same happens for `div_euclid`:

```typ
#calc.div-euclid(int("-9223372036854775808"), -1)
```

So it would indeed be nice to apply this change (using `checked`) to both quo and div_euclid.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -1035,7 +1035,15 @@ pub fn quo(
...
+                // Round towards negative infinity if the fraction is negative.
+                Some(DecNum::Int(if (a < 0) != (b < 0) && a % b != 0 {
```
PgBiel, line 1052: Maybe this is a bit better?

```suggestion
                Some(DecNum::Int(if q < 0 && a % b != 0 {
```

By the way, this is the official implementation in the currently nightly-only Rust `std` function for this[^1]:

```rs
pub const fn div_floor(self, rhs: Self) -> Self {
    let d = self / rhs;
    let r = self % rhs;

    // If the remainder is non-zero, we need to subtract one if the
    // signs of self and rhs differ, as this means we rounded upwards
    // instead of downwards. We do this branchlessly by creating a mask
    // which is all-ones iff the signs differ, and 0 otherwise. Then by
    // adding this mask (which corresponds to the signed value -1), we
    // get our correction.
    let correction = (self ^ rhs) >> (Self::BITS - 1);
    if r != 0 {
        d + correction
    } else {
        d
    }
}
```

[^1]: https://doc.rust-lang.org/nightly/std/primitive.i64.html#method.div_floor

crates/typst-library/src/foundations/calc.rs
PgBiel: I suppose we can now include integers in this list (let's just say "number" in general). Same for div-euclid.
```suggestion
/// error if the resulting number is larger than the maximum 64-bit signed
/// integer or smaller than the minimum for that type.
```

PgBiel (commented):
Thanks!

crates/typst-library/src/foundations/calc.rs
andylizi: Thanks! I submitted a PR to fix the separate bug: #7419

crates/typst-library/src/foundations/calc.rs
andylizi: Ouch, I somehow missed just checking the quotient :joy:, thanks for pointing it out!

After playing around with the generated assembly for a bit, I feel like using `(a < 0) != (b < 0)` might still have a slight advantage here. It can be reliably optimized to `(a ^ b) < 0` which is a single instruction (`xor rsi, rdi`) with identical performance to `q < 0` (`test rax, rax`), and there is the additional benefit that it's not dependent on the division result. x86 `div` is infamous for its high latency and less dependency might help out-of-order execution [^1].

The `std` version is very clever, but might be unnecessarily clever for our use case here (the explanation comment is definitely not optional, for instance). The current code can be optimized to be branchless and performant as well, and the branches would also very predictable (I'd guess it won't be negative in 99%+ of cases for typesetting).

<details>
<summary>Current version</summary>

```asm
  cmp     rdx, 1    ; Check remainder
  mov     rcx, rax
  adc     rcx, -1   ; Subtract 1 if carry (remainder != 0)
  xor     rsi, rdi  ; Check different signs
  cmovns  rcx, rax  ; Conditional move to use the subtracted quotient
  mov     rax, rcx
  ret
```
</details>

<details>
<summary><code>std</code> version</summary>

```asm
  xor     rsi, rdi
  sar     rsi, 63
  test    rdx, rdx
  cmove   rsi, rdx
  add     rsi, rax
  mov     rax, rsi
  ret
```
</details>

[^1]: In practice it probably doesn't matter if it's optimized to branchless conditional moves anyway, but the lack of dependency would still help in case branches are emitted in specific situations.

crates/typst-library/src/foundations/calc.rs
PgBiel: Can't it be part of this PR? Both are fixing the same bugs, so it's easier for me to review that way.

andylizi: It's not just `calc.quo` and `calc.div-euclid` that panic on overflow, there's also `calc.rem`, `calc.rem-euclid`, and even `calc.gcd`. This PR is fixing the flooring behavior in `calc.quo`, which feels like a completely separate bug to the overflow problem that is present everywhere `a / b` or `a % b` is used.

laurmaedje: Separate PR sounds good then.

PgBiel: I wasn't aware those other functions were also affected, so that sounds good indeed. Thank you for the explanation.

crates/typst-library/src/foundations/calc.rs
andylizi: I'm leaning towards keeping the current algorithm for the above reasons, curious to hear your thoughts

PgBiel: Yeah let's keep it. Thanks for checking!

tests/suite/foundations/calc.typ
```diff
@@ -126,11 +126,16 @@
 // Test the `quo` function.
 #test(calc.quo(1, 1), 1)
 #test(calc.quo(5, 3), 1)
-#test(calc.quo(5, -3), -1)
+#test(calc.quo(5, -3), -2)
+#test(calc.quo(-5, 3), -2)
+#test(calc.quo(-5, -3), 1)
+#test(calc.quo(6, -3), -2)
```
PgBiel, line 144: For good measure, I think the example in the linked issue (with 0 and -1, both should be the same now) should also be added.

andylizi: Done!

PR #7407, opened 2025-11-18
Fix that there're no autocomplete suggestions in math functions

### LINKED ISSUE #6312: typst-ide: No autocomplete suggestions in math functions
### Description

When the cursor is inside the arguments of a function in math mode, `typst_ide::autocomplete` gives no suggestions in neither explicit nor implicit mode, although in general math functions compose almost without limitations and completions within one should probably be equivalent to those offered at any other position within math mode.

For example when position is: `$ sqrt(|) $` there are no suggestions. This is also observable in the web app.

If the above sounds acceptable, I'm happy to work on a patch.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

RungeCC:
Fix #6312

Note about this PR:

## Changes in `typst_syntax`

+ Add a new method (called `mode`) to `typst_syntax::node::LinkedNode`, following the suggestion https://github.com/typst/typst/pull/6415#issuecomment-2961739270, it returns an `Option<SyntaxMode>`. If some unexpected conditions are encountered, it will return `None`.
  + Also, I'm not sure if it shall be a method or a free function.
+ It's archieved by explicitly exhaustively (namely, never use `_`) matching `self.kind()`, leaf nodes and containers are equally handled. Some nodes can not determined on their own, I leave comments for each exception case. 
  + For leaf node, checking their parents' `mode` is enough except for `Ident`, which needs to check whether it's preceded by a `Hash`(`#`), as in #6415. 
  + For container (node that could have children), there're three exceptions: `FuncCall`, `Args` and `FieldAccess` (that's also the main reason why auto-completion wasn't working before). IIUC, mode of `FuncCall` or `FieldAccess` is determined by the left most child -- `callee` of a `FuncCall` or the leading term of a `FieldAccess`.
+ Add a test for new `mode`, consider various cases, but not exhaustive (cover every `SyntaxKind`), since I'm not sure whether this is necessary and the current test function is already huge.
  + I put all testcases in a single test function, not sure if I should split them up, and if so, at what granularity.

## Changes in `typst_ide`

+ Now, as metioned in https://github.com/typst/typst/pull/6415#issuecomment-2971884637, now `complete_code` is done before `complete_math` and `complete_markup`, so codes for check interpolation are all removed. Also, in `globals` and `complete_math|markup|code`, use abovemetioned `LinkedList::mode` to check `mode`.
+ A related test is added.

### DIFF
--- a/crates/typst-syntax/src/node.rs
+++ b/crates/typst-syntax/src/node.rs
@@ -903,10 +903,12 @@ impl LinkedNode<'_> {
     fn as_parent(&self) -> Option<SyntaxMode> {
         self.parent().and_then(|parent| parent.mode())
     }
+
     #[inline]
     fn as_parent_or_default(&self, default: SyntaxMode) -> Option<SyntaxMode> {
         self.parent().map_or(Some(default), |parent| parent.mode())
     }
+
     /// Get the `SyntaxMode` of this node.
     ///
     /// Would be `None` if can not determine.


RungeCC:
CC @cAttte, @IgKh.

isuffix:
I've looked this over and I'm happy with it :+1:

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+            SyntaxKind::RefMarker => Some(SyntaxMode::Markup),
+            SyntaxKind::Heading => Some(SyntaxMode::Markup),
+            SyntaxKind::HeadingMarker => Some(SyntaxMode::Markup),
+            SyntaxKind::ListItem => Some(SyntaxMode::Markup),
+            SyntaxKind::ListMarker => Some(SyntaxMode::Markup),
+            SyntaxKind::EnumItem => Some(SyntaxMode::Markup),
+            SyntaxKind::EnumMarker => Some(SyntaxMode::Markup),
+            SyntaxKind::TermItem => Some(SyntaxMode::Markup),
+            SyntaxKind::TermMarker => Some(SyntaxMode::Markup),
+            SyntaxKind::Equation => Some(SyntaxMode::Math),
```
laurmaedje, line 953: I'm not sure this is correct. The contained `Math` node is definitely math, but the equation itself should probably be classified Markup/Code depending on the parent?

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
     }
 }
 
+impl LinkedNode<'_> {
+    /// Get the `SyntaxMode` of this node.
+    ///
+    /// Would be `None` if can not determine.
+    pub fn mode(&self) -> Option<SyntaxMode> {
```
laurmaedje, line 915: could you reorder the match arms to exactly match `enum SyntaxKind`?

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+            SyntaxKind::RightBracket => self.parent().and_then(|parent| parent.mode()),
+            SyntaxKind::LeftParen => self.parent().and_then(|parent| parent.mode()),
+            SyntaxKind::RightParen => self.parent().and_then(|parent| parent.mode()),
+            SyntaxKind::Comma => self.parent().and_then(|parent| parent.mode()),
+            SyntaxKind::Semicolon => self.parent().and_then(|parent| parent.mode()),
+            SyntaxKind::Colon => self.parent().and_then(|parent| parent.mode()),
+
+            // Either in code or in markup.
+            SyntaxKind::Star => self.parent().and_then(|parent| parent.mode()),
+            // Either in code or in markup.
```
laurmaedje: I think this should be "Either in markup or in math."

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+            SyntaxKind::Import => Some(SyntaxMode::Code),
+            SyntaxKind::Include => Some(SyntaxMode::Code),
+            SyntaxKind::As => Some(SyntaxMode::Code),
+
+            SyntaxKind::Code => Some(SyntaxMode::Code),
+            // `Ident` is in math if it's parent is in math and it's previous sibling is not a `Hash`.
+            // Otherwise, it's in code.
+            SyntaxKind::Ident => {
+                if self.parent().and_then(|parent| parent.mode())
+                    == Some(SyntaxMode::Math)
```
laurmaedje, line 1035: can this actually happen without a hash? don't we generate `MathIdent` in math?

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+            SyntaxKind::Parenthesized => Some(SyntaxMode::Code),
+            SyntaxKind::Array => Some(SyntaxMode::Code),
+            SyntaxKind::Dict => Some(SyntaxMode::Code),
+            SyntaxKind::Named => Some(SyntaxMode::Code),
+            SyntaxKind::Keyed => Some(SyntaxMode::Code),
+            SyntaxKind::Unary => Some(SyntaxMode::Code),
+            SyntaxKind::Binary => Some(SyntaxMode::Code),
+            // Mode of FieldAccess and FuncCall is determined by the leftmost leaf
+            // `callee` of `FuncCall` and leftmost expr of a `FieldAccess` chain.
+            // It's either in code or in math, the later case is only possible if
```
laurmaedje:
```suggestion
            // It's either in code or in math, the latter case is only possible if
```

crates/typst-syntax/src/node.rs
```diff
@@ -978,6 +1174,402 @@ impl std::error::Error for Unnumberable {}
 mod tests {
     use super::*;
     use crate::Source;
+    #[test]
+    fn test_linked_node_mode() {
```
laurmaedje, line 1224: This test code is quite repetitive. I think just extracting a helper function that collapses each case down to a single line would go a long way. E.g. `test("#! typ", 0, None)` with an `fn test(..)` that's marked `#[track_caller]`. If you search around the codebase, you'll see a couple instances of this pattern.

crates/typst-syntax/src/node.rs
laurmaedje: For the delimiters (brackets and dollars in particular), it's also not entirely clear to me which they should belong to.

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+            SyntaxKind::Int => Some(SyntaxMode::Code),
+            SyntaxKind::Float => Some(SyntaxMode::Code),
+            SyntaxKind::Numeric => Some(SyntaxMode::Code),
+            SyntaxKind::Str => Some(SyntaxMode::Code),
+            SyntaxKind::CodeBlock => Some(SyntaxMode::Code),
+            SyntaxKind::ContentBlock => Some(SyntaxMode::Markup),
+            SyntaxKind::Parenthesized => Some(SyntaxMode::Code),
+            SyntaxKind::Array => Some(SyntaxMode::Code),
+            SyntaxKind::Dict => Some(SyntaxMode::Code),
+            SyntaxKind::Named => Some(SyntaxMode::Code),
```
laurmaedje: This can also occur in a math call. Would that be a problem?

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+                    SyntaxKind::MathIdent => SyntaxMode::Math,
+                    _ => SyntaxMode::Code,
+                })
+            }
+            SyntaxKind::FuncCall => {
+                self.leftmost_leaf().map(|leaf| match leaf.kind() {
+                    // Eagerly return
+                    SyntaxKind::MathIdent => SyntaxMode::Math,
+                    _ => SyntaxMode::Code,
+                })
```
laurmaedje, line 1085: this is probably fine in practice, but I wonder whether it's a bit overzealous. the leftmost _leaf_ could be way down the hierarchy (though deep nesting is probably only possible in code and then it would not be a math ident. but I did a bit of a double-take here.

crates/typst-syntax/src/node.rs
```diff
@@ -898,6 +898,202 @@ impl LinkedNode<'_> {
...
+            }
+            // `Args` is always within a `FuncCall`.
+            SyntaxKind::Args => self.parent().and_then(|parent| parent.mode()),
+            SyntaxKind::Spread => Some(SyntaxMode::Code),
```
laurmaedje: same as `Named`, this can also be in math args. not sure how to classify that.

crates/typst-syntax/src/node.rs
```diff
@@ -978,6 +1174,402 @@ impl std::error::Error for Unnumberable {}
 mod tests {
     use super::*;
     use crate::Source;
+    #[test]
```
laurmaedje, line 1223:
```suggestion

    #[test]
```

crates/typst-syntax/src/node.rs
laurmaedje: cc @isuffix

RungeCC: Hmm, I'm not sure. It seems that there's room for choice if there's always a `Math` node within the `Equation` node.

crates/typst-syntax/src/node.rs
RungeCC: IIUC, this at least happens at `r` in case `$arrow.r$`, which parent is a `FieldAccess`. Not sure are there any other cases.

crates/typst-syntax/src/node.rs
RungeCC: You're right. Sorry about that I missed this and `Spread` (and also `Dots`), due to I misunderstand `FuncCall` in math.

IIUC, for both `Named` and `Spread` when in math they're always within some `Args`, so I think we should just take the parent mode.

RungeCC:
> * This case:
>   ```typst
>   $ abs(|) $
>   ```
Ah, I find out it runs into `complete_param` and failed due to a missing branch `MathIdent` in `resolve_global_callee`. And my original testcase is actually wrong since $f(...)$ is always not a `FuncCall`.

However, there still another problem, whether we want to completely use `complete_math` even when  `complete_param` is possible? For example, in

```typst
$ attach(|) $
```

shall we complete `"b", "tb"` etc or just do scope completion in math? I'm not sure.

RungeCC:
> * This case:
>   ```typst
>   $ f(|) $
>   ```

I think this could be solve with:

```diff
-    if matches!(
-        ctx.leaf.kind(),
-        SyntaxKind::Text | SyntaxKind::MathText | SyntaxKind::MathIdent
-    )
+   // Ensure that we are not just behind a left parenthesis,
+   // which shall be handled in the "Anywhere" case.
+    if matches!(
+        ctx.leaf.kind(),
+        SyntaxKind::Text | SyntaxKind::MathText | SyntaxKind::MathIdent
+    ) && ctx.leaf.text() != "("
```

However, this seems weird to me, since before we never try to complete barely `$f(|)$` (originally, this failed since we do not handle `MathDelimited` case in `complete_math`, `f(x|)` works since such a `MathText` is always grouped within some `Math` node), though we complete `f(|)` here `f` is a function defined somewhere above, but that's `complete_params` but not `complete_math`. On the other hand, it seems that we should also handle case like $f(x, |)$, here space ` ` is a `MathText` parent `MathDelimited`, and so on and so on, I'm not sure whether all of them shall be handled.

Notice originally the similar case `$ f(x,|) $` also have a completion at `from` 5 (not 6), since `,` is a MathText and grouped with `x` in a `Math` node, so can successfully run into `complete_math -> math_completion -> scope_completion`. However, it seems there's no completion in webapp, is this also incorrect?

And this:

```typst
$ abs(|) $
```

can be solved with eagerly returned if in math:

```rust
    // If we find out the current piece to be complete it's in math mode,
    // it shall be completed in math, so quit for calling
    // `math_complete` later.
    if deciding.mode() == Some(SyntaxMode::Math) {
        return false;
    }
```

I'm not sure shall we also quit if we call `complete_params` in math mode.

Notice that the later change will cause no completion for named parameters of math functions such as `attach, limits`.

If those changes look good, I'll apply them later.

crates/typst-syntax/src/node.rs
RungeCC: Um, I also not sure about that could this nesting actually happens in math mode.

Or we could just look for the prev sibling and check is it a Hash.

crates/typst-syntax/src/node.rs
laurmaedje: Given that `eval("$...$", mode: "math")` won't work, I think it should be code or markup depending parent.

IgKh:
@RungeCC apologies in my delay in looking into this. I have also tested the PR (in my case by building Katvan against your branch) and it did not work, but I see you are already aware of this.

Regarding "true" math function calls, which are the topic of my original issue, I think that in general patching `resolve_global_callee` to make the case properly catch inside `param_completions` is probably the more correct thing to do - as it gives couple of benefits: named parameters as you have said, but also awareness of the number of positional parameters.

But I think it then does need some more changes there:
* Positional parameter completions should probably use `math_completions`, as `param_value_completions` is missing many relevant completions.
* In math function calls (only!) there is an additional allowed parameter separator, which is `;` (the semicolon).  It would need consideration at least in `complete_params` (as an additional deciding node), and in the end of `param_completions` (for "enriching" with a space). However I'm not super confident about just adding it outright in the same path that also handles code mode function calls, which is mostly what left my original stab at the issue as a draft.

Some additional test cases that could be helpful:
* `$sqrt(|)$` (One positional parameter)
* `$frac(1,|)$` (More than one positional parameter)
* `$vec(1,2,3,4,5,|)$` (Variadic)
* `$mat(1,2;|)$` and `$mat(1,2; |)$` (The special array-forming syntax with a semicolon)

crates/typst-syntax/src/node.rs
```diff
@@ -1047,13 +1050,24 @@ impl LinkedNode<'_> {
             SyntaxKind::Parenthesized => Some(SyntaxMode::Code),
             SyntaxKind::Array => Some(SyntaxMode::Code),
             SyntaxKind::Dict => Some(SyntaxMode::Code),
-            SyntaxKind::Named => Some(SyntaxMode::Code),
+            // Either in code or in math,
+            // it's in math only when it's within an `Args` who is in math.
+            SyntaxKind::Named => {
```
laurmaedje, line 1057: couldn't this just be `as_parent()` or am I missing something?

laurmaedje:
Hey @RungeCC, just wanted to know whether you're still interested in completing this.

RungeCC:
I'm very sorry that I haven't had time to work on this pull request recently, and I apologize for wasting the reviewers' time. I will close this PR.

isuffix:
> I'm very sorry that I haven't had time to work on this pull request recently, and I apologize for wasting the reviewers' time. I will close this PR.

You haven't wasted anyone's time! I'm probably going to rework this soon (with co-author credit) to get it over the finish line. The syntax mode accessor was a smart contribution; I think it's the right fix for this problem, and I wouldn't have come up with it myself.

Issue #1920, opened 2023-08-12
Spaces after « and before » should be nobreak in french


crowdagger:
### Description

(Hi, first time submitting an issue for this project, hope I'm doing it things more or less right!)

When setting lang to french, spaces after and before the french « quotes » (or "guillemets") « and » should be unbreakable. Currently, this is not the case: 

```
« Bonjour, ceci est un dialogue » 

"Bonjour, ceci est un dialogue"
```
![Screenshot 2023-08-12 at 14-08-44 Test – Typst](https://github.com/typst/typst/assets/1961791/6957e2c6-0e7f-404b-8323-32d5648da161)

As the example above shows, using the smart quote feature it works correctly, but some french layouts (e.g. Bépo) allow to easily type « and » and if you are used to do it the results won't be great.

(If you agree this should be fixed and can point me in the right direction, I'd be happy to try and start working on a PR ^^)

### Operating system

Web app, Windows

### Typst version

- [X] I am using the latest version of Typst

Omikhleia:
Slightly relates to https://github.com/typst/typst/issues/1194#issuecomment-1561194499 (where I mentioned in passing that "the use of a `\u{00A0}` space in French quotes is not really respecting the true typography")

My 2 cents on the topic:

The "guillemet space" shall always be unbreakable indeed, and is "usually" smaller than the regular inter-word space, with also a bit less stretch/shrinkability. Babel (LaTeX) adopted a definition recommended by Thierry Bouche years ago (and SILE followed it too more recently), which as far as I understand is typographically sound for `fr` and `fr-CA` (but note that `fr-CH` may prefer a thin space there, or even no guillemet space at all).

Likewise, for the record, French requires a non-breakable space (but usually stretch/shrinkable as an inter-word space for justification purposes) before a colon -- and a thin space before certain "high" punctuation signs `;!?` (with a number of exceptions...), unbreakable too and usually non-variable, ideally around half of the inter-word space, at least in `fr`; `fr-CA` usually does not add that space before the so-said high punctuations)

Obviously/unfortunately, few typesetting software follow these usual/old conventions -- as noted, LaTeX with Babel does, as well as SILE -- But many just use some unbreakable _fixed_ spaces in these cases. FWIW, LibreOffice and MSWord do that (i.e. insert the same fixed nbsp everywhere), though they do implement some of the differences between French from France, Canada and Switzerland ;)

yodatak:
[...]

Issue #6774, opened 2025-08-16
In Chinese typsetting, Interpunct (U+00B7) should not appear at line start and should have consistent width


chiefaccelerators:
### Description

Right now the Chinese punctuation handling is mostly okay, but interpunct is one exception. Per clreq:

> The width of interpuncts varies in different regions. In principle, in Hong Kong and Taiwan, interpuncts should have the same dimensions as a Hanzi in both vertical writing mode and horizontal writing mode. In Mainland China, interpuncts take up half the space of a Hanzi.

> interpuncts should not appear at the line start.

Interpunct can use various codepoints, but most Chinese input methods produce U+00B7, so it's what I'm testing:

<img width="1407" height="658" alt="Image" src="https://github.com/user-attachments/assets/517965d0-1d36-43df-a560-fcd757567ea8" />

Granted, the width thing is also partly a font issue and can be easily patched with a show rule. But it would save some trouble if typst can give it a consistent width based on the current lang and region setting.

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

YDX-2147483647:
Perhaps the width of interpuncts can be adjusted. At present, clreq descriptions are inconsistent. (I created https://github.com/w3c/clreq/issues/705.) Therefore, I suggest postponing the implementation.

Nevertheless, the prohibition rules for line start can be fixed first.
The only existing test case that contains interpuncts is text-chinese-basic.
https://github.com/typst/typst/blob/560e49b67ca738e2b2768d7eda3bccdb47570dcc/tests/suite/layout/inline/cjk.typ#L3-L10

cocoAutumn:
Even in the national standard itself, the witdh of interpunct is still not well-defined.
https://openstd.samr.gov.cn/bzgk/gb/newGbInfo?hcno=22EA6D162E4110E752259661E1A0D0A8
>（横排文稿标点符号的位置和书写形式）
5.1.7 间隔号标在需要隔开的项目之间，占【半个字】位置，上下居中，不出现在一行之首。（half-width）

But in the following examples they are all full-width.
<img width="1920" height="1039" alt="Image" src="https://github.com/user-attachments/assets/73be26f1-5167-421e-912e-dc10463698b1" />
Since different fonts often make interpunct only half-width or even thinner, in many plain-text environment people may use U+30FB (Katakana Middle Dot) to ensure the full-width, which is not easy to input for non-Japanese users.

YDX-2147483647:
[...]

Issue #1009, opened 2023-04-27
Language-aware line breakpoints


peng1999:
### Description

Requirements for Chinese Text Layout have  [Prohibition Rules for Line Start and Line End](https://www.w3.org/TR/clreq/#prohibition_rules_for_line_start_end), and  Japanese  also has [similar](https://www.w3.org/TR/jlreq/#characters_not_starting_a_line) [rules](https://www.w3.org/TR/jlreq/#characters_not_ending_a_line). This is not implemented correctly in current typst, which uses vanilla UAX#14 directly by `xi-unicode`. 

UAX#14 can be used directly in most latin languages, but is not enough for CJK. In UAX#14 §8 Customization:
> A real-world line breaking algorithm has to be tailorable to some degree to meet user or document requirements. …
> In Japanese, for example, tighter and looser specifications of prohibited line breaks may be used.

The most noticeable gap between vanilla UAX#14 and Chinese/Japanese convention is the handling of quotation marks. Here is an example:

<details><summary>An example</summary>

```typst
#set page(width: 5em + 2em, margin: (x: 1em))
#set text(font: "Noto Serif CJK SC")
测试文本，“测”
```

Current line breaking: 
![](https://cdn.discordapp.com/attachments/1056574658771681361/1101024331942858873/rendered.png)

Desired line breaking:
![](https://cdn.discordapp.com/attachments/1056574658771681361/1101024377115512832/rendered.png)

</details>

ICU has a tailored line breaking [rule](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/brkitr/rules/line_cj.txt) for Chinese and Japanese, but unfortunately this is a C library and cannot used in typst.

Related: #276 

### Typst version

- [X] I am using the latest version of Typst

peng1999:
> There is an official [icu4x](https://github.com/unicode-org/icu4x) implementation in Rust by Unicode.

Unfortunately icu4x does not support Chinese/Japanese line break tailoring like ICU. See unicode-org/icu4x#3284

PR #7412, opened 2025-11-18
Update ICU dependencies

mohtakver:
This PR updates the various ICU4X dependencies from version 1.4 to 2.1.1. This change will fix issue #1920:

``` typst
#lorem(14) A test » \  
#lorem(14) A test !
```
content

I would not be surprised if it fixes other things as well given such a big version jump.

In this PR I opted to use the built-in compiled ICU data instead of the the runtime-loaded data from `typst-assets`. Correct me if I'm wrong, but the `icu4x-datagen` commands used to generate the data in that crate don't generate anything different from the built-in stuff, e.g. both cover all locales. Using the compiled data results in cleaner code, faster runtime, and importantly, more up to date CLDR language support.

### DIFF
--- a/crates/typst-layout/src/inline/linebreak.rs
+++ b/crates/typst-layout/src/inline/linebreak.rs
@@ -38,7 +38,7 @@ static SEGMENTER: LazyLock<LineSegmenterBorrowed> =
 
 /// The Unicode line break properties for each code point.
 static LINEBREAK_DATA: LazyLock<CodePointMapDataBorrowed<LineBreak>> =
-    LazyLock::new(|| CodePointMapData::new());
+    LazyLock::new(CodePointMapData::new);
 
 // Zero width space.
 const ZWS: char = '\u{200B}';


eltos:
Here is a test specifically for #1920 for you to cherry-pick @mohtakver, in case you find it's useful: https://github.com/typst/typst/commit/3539ef3dffa59cd895c768decf46a775ac67e4b2

laurmaedje:
The current blob data in `typst-assets` uses a [patched ICU data generator](https://github.com/typst/icu4x/tree/cj-patch). See https://github.com/typst/typst/issues/1009 and https://github.com/typst/typst-assets/blob/0bbb3f7ebe775b980a3ca4ae9c49d9ecdb7e021a/src/lib.rs#L14-L45 for background. I'm honestly not sure whether there was an attempt to upstream the patch in https://github.com/peng1999/icu4x/commit/b9beb6cbf633d61fc3d7983e5baf7f4449fbfae5.

I fear that, as-is, this may regress Typst's CJK handling. Two CJK test references have been changed in this PR, which is probably related to that.

laurmaedje:
It would also be worth investigating how this affects Typst's binary size. I'm not sure how exactly the blob data provider avoids bloat (if at all). The data generation commands inspect the built binary to determine exactly what it references. It's possible that the blob data provider simply relies on dead code elimination, but I wouldn't bet on that working perfectly.

laurmaedje:
cc @peng1999 if you have the time, your input would be appreciated on the CJK topic. :)

peng1999:
> I fear that, as-is, this may regress Typst's CJK handling. Two CJK test references have been changed in this PR, which is probably related to that.

Yes, that is a regression. The two tests shouldn’t be modified. 

> I'm honestly not sure whether there was an attempt to upstream the patch in [peng1999/icu4x@b9beb6c](https://github.com/peng1999/icu4x/commit/b9beb6cbf633d61fc3d7983e5baf7f4449fbfae5).

See https://github.com/typst/typst/pull/1355#discussion_r1210117454 (you should expand the resolved review to show the comment). I think icu4x 2.0 supports locale in segmenter, so patched datagen is no longer needed.

YDX-2147483647:
> I think icu4x 2.0 supports locale in segmenter, so patched datagen is no longer needed.

I created a minimal example of  using `icu_segmenter` 2.1.2 in https://github.com/typst/typst/issues/6774#issuecomment-3793985598.
Even if `LineBreakOptions`'s `content_locale` is set to `zh`, the segementer without patch is still problematic. Specifically, `：“` should be allowed to break into `：\n“`, but not.

So, **what should we do now?**
Request changes in icu4x, or update the patch to icu4x 2.1?

peng1999:
related issue: https://github.com/unicode-org/icu4x/issues/5595

laurmaedje:
> So, what should we do now?
> Request changes in icu4x, or update the patch to icu4x 2.1?

I think ideally both. The first for the long-term and the second to unblock us.

I'll close this PR since OP does not seem to be involved anymore.

@YDX-2147483647 If you'd be interested in picking this up, then feel free to! Otherwise, I will try to take a look sometime soon-ish.

YDX-2147483647:
> If you'd be interested in picking this up

The icu4x repo and its relationship with the Unicode line breaking algorithm are more complicated than I thought.
I haven't taken any substantive actions at present, and it's likely that I'm incapable of doing it in the near future.

laurmaedje:
Sure, no problem. Thanks for the quick response!

PR #7415, opened 2025-11-19
Pass the inferred writing script to HarfBuzz, making `locl` effective

### LINKED ISSUE #7396: The inferred writing script is not passed to HarfBuzz, making `locl` ineffective
### Description

The inferred writing script (e.g., latn/hani) is not passed to `rustybuzz::UnicodeBuffer`, making the [OpenType `locl` (Localized forms) feature](https://learn.microsoft.com/typography/opentype/spec/features_ko#tag-locl) ineffective.

## Outcome of this issue

When using certain fonts and languages, the position of period marks (U+3002 `。` ideographic full stop) between generated texts are not stable.

In the example below, the first period is placed at the center of the square space, but the second period is placed at the lower left corner.

```typst
#set text(lang: "zh", region: "TW", font: "Noto Serif CJK SC")
#set heading(numbering: "1")
= Heading <a>

句號。@a。@a 何故？
```

Image

The problem happens if the glyph for `。` should be a non-default localized form provided by the `locl` feature.

## Background and explanation

Typographical rules may have regional differences. For example, when writing Chinese horizontally, the period mark is placed at the _lower left_ corner in the square space in Chinese Mainland, but placed at the _center_ in Taiwan and Hong Kong.

Some fonts handle these differences with the `locl` feature. For instance, Noto Sans/Serif CJK, the language-specific version (also the version shipped on typst.app), contains all variants of the period mark, and the layout engine can select the needed glyph for each language from the font.

Note that `locl` requires the text to be properly tagged with both the language system and the script. If not, the default glyph will be used.

In the [GSUB table](https://learn.microsoft.com/en-gb/typography/opentype/spec/gsub#gsub-table-organization) of Noto CJK:

1. There are a default **script** (DFLT) and 6 non-default scripts (cyrl, grek, hang, hani, kana, latn).
2. The DFLT script contains only the default **language system**, and the non-default scripts also contain 5 non-default language systems (JAN, KOR, ZHH, ZHS, ZHT).
3. The `locl` **feature** is only available for non-default scripts and language systems.

In Typst:

- The `locl` **feature** is somewhat supported and enabled by default.

- The **language system** can be tagged by applying `set text(lang: "ja", region: "JP")` (JAN), ko-KR (KOR), zh-HK (ZHH), zh-CN (ZHS), or zh-TW (ZHT).

- The **script** can be manually tagged by applying `set text(script: "hani")` etc.

  However, ideally, the default `set text(script: auto)` should just work and Typst should **infer** the script from the used characters.

Typst has problems on inferring the script from characters. Specifically, the [Unicode script property](https://unicode.org/reports/tr24/#Script) of `。` is *Common*, which means it may be used with multiple scripts.

- When `。` is shaped in `句號。`, HarfBuzz can infer from the context that the script is hani, so `locl` is applicable and the correct localized glyph will be used.
- But if `。` is shaped alone, as the `。` in `@a。@a`, then HarfBuzz cannot determine the script and will use the DFLT script, so `locl` is not available and the default glyph will be used.

In the latter case, Typst is responsible for telling HarfBuzz that it is hani. However, Typst v0.14.0 only does so if the end author `set text(script: "hani")` manually. This should be improved.

https://github.com/typst/typst/blob/0da0165954e027ba48f7ba4a03e3b7b5b35ea8f6/crates/typst-layout/src/inline/shaping.rs#L960-L964

## Further examples

Period marks between `strong`, `emph`, `highlight`, `ref`, or colored texts are also affected, as shown below.

```typst
#set text(lang: "zh", region: "CN", font: "Noto Serif CJK TC")
桑之未落，其叶沃若。于嗟鸠兮，无食桑葚！\
桑之未落，#text(red)[其叶沃若]。#text(blue)[于嗟鸠兮]，无食桑葚！\
桑之未落，#highlight[其叶沃若]。#highlight[于嗟鸠兮]，无食桑葚！\
#show ref: _ => [于嗟鸠兮]
桑之未落，*其叶沃若*。@key，无食桑葚！
```

Image

Characters that are not specific to a single script are affected, including `、，。．？！：；≤≥≮≯`.

## Workaround

```typst
#show "。": set text(script: "hani")
```

## Notes

This issue was originally reported on 2025-08-14. Nothing was actionable until we tested the behavior of HarfBuzz (with the hb-view CLI), Firefox, and XeLaTeX / LuaLaTeX, and contacted an expert on typography by email.

Previous discussions was recorded in https://github.com/typst-doc-cn/clreq/issues/41, in Chinese, but with many screenshots.

### Reproduction URL

_No response_

### Operating system

Web app, Windows, Linux

### Typst version

YDX-2147483647:
This PR lets the inline shaping engine infer the scripting script (e.g., hani/latn/…) from the surrounding context of the text being shaped, and pass it to HarfBuzz if appropriate.

It will make the OpenType `locl` (Localized Forms) feature also effective for edge cases.
For instance, the second period mark in the example below uses the wrong glyph (corner-justified form) with v0.14.0, but will use the correct glyph (centered form) with this PR.

```typst
#set text(lang: "zh", region: "TW", font: "Noto Serif CJK SC")
#set heading(numbering: "1")
= Heading <a>

句號。@a。@a 何故？
```

Before:

Image

After: (the screenshot is trimmed)

图片

Fixes #7396

## Updating Noto CJK in dev assets

> [!NOTE]
>
> No test is added because typst-dev-assets has no font supporting `locl`.
> The latest Noto CJK supports `locl`, but the current version in typst-dev-assets are too old. See https://github.com/typst/typst/issues/7396#issuecomment-3541814065 for details.
>
> Discussions were moved to https://github.com/typst/typst-dev-assets/issues/18.


crates/typst-layout/src/inline/shaping.rs
```diff
@@ -725,10 +734,32 @@ pub fn shape_range<'a>(
...
+        // original (though generic) script.
+        let prior_script = std::iter::once(prior_script)
+            .chain(
+                // First backward then forward
+                (text[..range.start].chars().rev())
+                    .chain(text[range.end..].chars())
+                    .map(|c| c.script()),
+            )
+            .find(|&sc| !is_generic_script(sc))
+            .unwrap_or(prior_script);
```
laurmaedje, line 751: Is there some precedent for this particular fallback approach in other software? Sometimes, we have to go with a custom approach, but generally I would always try to look for prior art.

laurmaedje: Generally speaking, since we already do script segmentation below, we probably don't need to rely on Harfbuzz's segment property guessing at all... and could always provide the appropriate script here. Though I'm not sure whether that has any unintended consequences if applied naively.

YDX-2147483647: > Is there some precedent for this particular fallback approach

Well, I am not sure…

Regarding the overall direction, I discussed the issue briefly with [a member of W3C Chinese Layout Task Force](https://www.w3.org/TR/clreq/#contributors:~:text=LI%2C%20Bobby%20TUNG%2C-,Eric%20Q%2E%20LIU,-%2C%20Fuqiao%20XUE%2C "Eric Q. Liu") via email. I didn’t show him the code or any detailed algorithm, but he inferred that it should be fixed by the high-level engine, not the font or HarfBuzz.

As for the specific fallback approach, to be honest, I (kind of) came up with it on my own.

The root cause of the issue is that markups can interrupt the segmentation algorithm, making information beyond `range` necessary for `shape_range(text, range, ...)`. If there wasn’t markup, those ranges would be passed to `shape_range` as a whole and everything would be okay.

To get around that, this PR incorporates information beyond `range` into `shape_range` and `shape_segment` as if the ranges had never been split by markups. In this sense, this PR does not propose any new approach.

Nonetheless, AI agents ([Claude](https://github.com/YDX-2147483647/typst/agents/pull/1) in English and [Kimi](https://www.kimi.com/share/19cdc30d-8982-8434-8000-00001c72aa92) in Chinese) claim that the current matches GTK/Pango, ICU, Skia,  Qt QTextEngine, etc. I am not capable of verifying their claims. Here’s what I know:

- [UAX #24: Unicode Script Property](https://unicode.org/reports/tr24/#Script_Values):

  > A value of **Inherited** means that the character is treated as if it had the Script property value of a preceding base character.

  This PR matches the above description, except that it also looks subsequent characters. That’s necessary because `typst-layout/src/inline/shaping.rs` does not distinguish between Inherited and Common, and `（` in `“（示例……）”` is Common and hence requires looking forward.

- [ICU4C `uscript_nextRun`](https://github.com/unicode-org/icu/blob/426cea1b85e82e632dc5c0b35c7d329c0eb4af7b/icu4c/source/common/usc_impl.cpp#L257-L345) does not consider information beyond the range, and it has a subtle treatment for paired characters.

- Pango’s codebase is too complicated for me to understand, but it looks like that it also has [special treatment for paired characters](https://github.com/GNOME/pango/blob/2fb8d7782ec0e166aeeb961402db8afae521d574/pango/pango-script.c#L209-L268).

  [How the script of `。` is determined in `<strong>示例</strong>。`? - DeepWiki | Search](https://deepwiki.com/search/strongstrong-opentype-locl-scr_27c75f52-a322-418a-a562-35c5f60fdf2d?mode=fast) (in Chinese)

- I didn’t check [Skia](https://rust-skia.github.io/doc/skia_safe/type.Shaper.html) or [ICU4J](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/lang/UScriptRun.html#getScriptLimit()) in details.

> we probably don't need to rely on Harfbuzz's segment property guessing

I agree. [WebKit developers met a bug about in 2013](https://bugs.webkit.org/show_bug.cgi?id=110145): “Leaving direction to HarfBuzz to guess is *really* bad, but will do for now.”

However, considering that there are currently not many related issues, I think it's not worth refactoring at this time. It would be better to keep it as it is.

khaledhosny: Letting HarfBuzz guess any segment properties is really bad, production code should never do that (in my opinion it is even a mistake that HarfBuzz exposes this API).

Script segmentation should be done to full paragraph text at once, just like bidi. So inline markup should not interrupt it.

Another implementation that might be easier to follow, is [Raqm’s](https://github.com/HOST-Oman/libraqm/blob/00814288bec3deb7dc49201e135d572c730a96d3/src/raqm.c#L2032).

laurmaedje:
I feel like the diff is a bit too ad-hoc to be able to land with minor modifications. Given Khaled's comment, it seems the proper way forward is to implement proper full-paragraph script detection akin to BiDi detection. I would thus be inclined to close this.

YDX-2147483647:
I see. I'm fine with that. In any case, one can switch the version of Noto CJK.

Then I guess typst/typst-dev-assets#18 will also have to be postponed?

laurmaedje:
> I see. I'm fine with that. In any case, one can switch the version of Noto CJK.

I'll close this then. Thanks for your understanding.

> Then I guess https://github.com/typst/typst-dev-assets/issues/18 will also have to be postponed?

Not necessarily.

PR #7418, opened 2025-11-19
Fix potential EOL issues in CI tests

Andrew15-5:
Should fix CI errors in https://github.com/typst/typst/pull/7211.

See https://github.com/actions/checkout/issues/135.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


laurmaedje:
I'm not sure we really want to do this because contributors on Windows will likely not have this configured and then they could run into issues with the test suite. (The test suite not working on Windows has happened in the past.)

Since we've managed pretty well with the autocrlf so far, it might be better to work around this in the very rare case where it matters.

Not sure.

Andrew15-5:
[`man git-config`](https://www.man7.org/linux/man-pages/man1/git-config.1.html#:~:text=Setting%20this%20variable%20to%20%22true%22%20is%20the%20same%20as%20setting%20the):
> Setting this variable to "true" is the same as setting the text attribute to "auto" on all files and core.eol to "crlf". Set to true if you want to have CRLF line endings in your working directory and the repository has LF line endings. This variable can be set to input, in which case no output conversion is performed.

`crlf` is one of the most common first issues that everyone is facing when starting using Git. I think it's on the contributors to know the basic setup for Git.

The question you need to answer is, "Do you want to have CRLF line endings in your working directory while the repository has LF line endings?"

No, so it must be something other than `true`. There is `input`, though I'm under the impression that `false` and `input` are identical. At least if `core.eol` is unset. Which it is for me, and probably for default setup as well.

andylizi:
Chiming in as a Windows user, `autocrlf=true` has long been a painful footgun[^1][^2] especially for beginners who aren't familiar with line ending issues. Most resources I've found say `autocrlf` mainly exists to paper over legacy or broken tooling that can't handle LF correctly or that requires CRLF by design[^3] (e.g., batch files). From the official Git book[^4]:

> This is a subtle but incredibly annoying fact of cross-platform work; many editors on Windows silently replace existing LF-style line endings with CRLF, or insert both line-ending characters when the user hits the enter key. Git can handle this by auto-converting CRLF line endings [...]

But that's largely a thing of the past. Modern editors, even Windows Notepad, have handled LF properly for years. I don't see anything in Typst development today that would require the use of native EOL.

> Since we've managed pretty well with the autocrlf so far, it might be better to work around this in the very rare case where it matters.

The bigger issue from my perspective is consistency. Having files appear one way locally and another way in CI invites subtle differences that can be hard to debug. #7211 feels like less of a edge case and more like showing the kinds of problems this inconsistency can cause.

> [...] because contributors on Windows will likely not have this configured and then they could run into issues with the test suite

People in actions/checkout#135 suggested using `.gitattributes` to define line ending rules at the repo level rather than relying on global user settings. That seems like the ideal long-term approach, but this PR would already be a solid improvement.

[^1]: https://stackoverflow.com/a/2825829
[^2]: https://markentier.tech/posts/2021/10/autocrlf-true-considered-harmful/
[^3]: https://bornsql.ca/blog/the-hell-of-git-line-endings-and-the-not-so-simple-fix/
[^4]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

Andrew15-5:
Consistency is indeed the key. I'm down to use the config file instead.

I guess to prevent issues and enforce consistency, `core.eol` indeed should be set to `lf` as well, as was provided in the examples in the https://github.com/actions/checkout/issues/135.

laurmaedje:
I think the `.gitattributes` solution sounds nice: If I understand correctly, it will need no further settings from contributors and things will just work.

Andrew15-5:
My only small concern is binary files. `* text` should treat them also as text files, but it looks like it only matters for EOL. It can potentially break something, if we had such binaries and parsers that can't deal with EOL in a binary format, but I think there is no such thing, so it's fine. Otherwise, I'd list all the file extensions that actually should be addressed, i.e., `*.rs`, `*.typ`, `*.toml`, `*.nix`, `*.md`, etc. 

Like, technically, you shouldn't touch binary files.

---

https://github.com/typst/typst/actions/runs/19711894242/job/56474477185?pr=7418

yep...

---

Yeah, I guess, if the only binary format is PNG, then just excluding that is better, than listing all of the text formats.

laurmaedje:
Hmm, that's annoying. Having to maintain a list of binary file formats is also a bit annoying ... Now I'm not sure again which solution to use.

Andrew15-5:
Well, how often do you add new binary formats into the repository? And how much is it _actually_ annoying to maintain the list? A comment mentioning updating the file can be added as a tiny helper.

Nothing will break if the solution will be changed down the road (they are interchangeable). And it really only matters for CI, which is the reason it's being brought up.

From your messages, I'd go with the YAML workflow action approach and call it a day. Literally nothing changes for anyone, except the CI (getting more robust). That's my take, can't really say anything more than this.

andylizi:
If we use `* text=auto` instead of `* text`, that should keep the automatic detection behavior.

> When text is set to "auto", Git decides by itself whether the file is text or binary. If it is text and the file was not already in Git with CRLF endings, line endings are converted on checkin and checkout as described above. Otherwise, no conversion is done on checkin or checkout

https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvalueauto

laurmaedje:
> From your messages, I'd go with the YAML workflow action approach and call it a day. Literally nothing changes for anyone, except the CI (getting more robust). That's my take, can't really say anything more than this.

There would be a change: A Windows user with a stock installation would get failing tests immediately after checkout, precisely because CI would stop failing for scenarios that are problematic in the default configuration.

> If we use * text=auto instead of * text, that should keep the automatic detection behavior.

That seems worth a shot.

Andrew15-5:
> There would be a change: A Windows user with a stock installation would get failing tests immediately after checkout, precisely because CI would stop failing for scenarios that are problematic in the default configuration.

Sure, but then again, the additional EOL config that probably everyone that used Git for a while knows about will basically mirror the CI config and so it won't fail. I'm thinking that it's not realistic to base around stock config. Long contributors would have things set up already. And the major thing, is that those who are new or lazy, would not run the tests locally, so CI is basically the only place where the tests will run.

laurmaedje:
Thanks, also to @andylizi for the insights!

PR #7419, opened 2025-11-19
Fix panic when division overflows in `calc`

andylizi:
As pointed out in https://github.com/typst/typst/pull/7392#discussion_r2538385818. There is only one case where overflow happens: `i64::MIN.checked_div(-1)`[^1]

[^1]: https://doc.rust-lang.org/nightly/std/primitive.i64.html#method.checked_div

### DIFF
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -921,8 +921,9 @@ pub fn rem(
     dividend
         .apply2(
             divisor.v,
-            // `checked_rem` can only overflow on `i64::MIN % -1` which is mathematically zero.
-            |a, b| a.checked_rem(b).or(Some(0)).map(DecNum::Int),
+            // `checked_rem` can only overflow on `i64::MIN % -1` which is
+            // mathematically zero.
+            |a, b| Some(DecNum::Int(a.checked_rem(b).unwrap_or(0))),
             |a, b| Some(DecNum::Float(a % b)),
             |a, b| a.checked_rem(b).map(DecNum::Decimal),
         )
@@ -1007,8 +1008,9 @@ pub fn rem_euclid(
     dividend
         .apply2(
             divisor.v,
-            // `checked_rem_euclid` can only overflow on `i64::MIN % -1` which is mathematically zero.
-            |a, b| a.checked_rem_euclid(b).or(Some(0)).map(DecNum::Int),
+            // `checked_rem_euclid` can only overflow on `i64::MIN % -1` which
+            // is mathematically zero.
+            |a, b| Some(DecNum::Int(a.checked_rem_euclid(b).unwrap_or(0))),
             |a, b| Some(DecNum::Float(a.rem_euclid(b))),
             |a, b| a.checked_rem_euclid(b).map(DecNum::Decimal),
         )


crates/typst-library/src/foundations/calc.rs
```diff
@@ -893,8 +893,7 @@ pub fn odd(
 /// The value `calc.rem(x, y)` always has the same sign as `x`, and is smaller
 /// in magnitude than `y`.
 ///
-/// This can error if given a [`decimal`] input and the dividend is too small in
-/// magnitude compared to the divisor.
+/// This can error if the dividend is too small in magnitude compared to the divisor.
```
PgBiel: Seems like this needs an update now.
```suggestion
/// This can error if the dividend is too small in magnitude compared to the
/// divisor, or if the result of integer division would be larger than the
/// maximum 64-bit signed integer.
```

crates/typst-library/src/foundations/calc.rs
```diff
@@ -933,6 +932,8 @@ pub fn rem(
 /// The result of this computation is that of a division rounded to the integer
 /// `{n}` such that the dividend is greater than or equal to `{n}` times the divisor.
 ///
+/// This can error if the dividend is too small in magnitude compared to the divisor.
```
PgBiel: This seems more accurate for this function.
```suggestion
/// This can error if the resulting number is larger than the maximum value or
/// smaller than the minimum value for its type.
```

crates/typst-library/src/foundations/calc.rs
```diff
@@ -1012,8 +1013,8 @@ pub fn rem_euclid(
 /// Calculates the quotient (floored division) of two numbers.
 ///
 /// Note that this function will always return an [integer]($int), and will
-/// error if the resulting [`float`] or [`decimal`] is larger than the maximum
-/// 64-bit signed integer or smaller than the minimum for that type.
+/// error if the resulting number is larger than the maximum 64-bit signed integer
+/// or smaller than the minimum for that type.
```
PgBiel:
```suggestion
/// error if the resulting number is larger than the maximum 64-bit signed
/// integer or smaller than the minimum for that type.
```

crates/typst-library/src/foundations/calc.rs
PgBiel, line 932: Seems like this might be misleading now. Maybe we can try to return a different Result in each case instead (with this string for the decimal case and `too_large` for the integer case)?

PgBiel, line 565: Let's mention the error explicitly here now.

```suggestion
/// Calculates the greatest common divisor of two integers.
///
/// This will error if the result of integer division would be larger than the
/// maximum 64-bit signed integer.
```

PgBiel (commented):
Looks good, thank you! My review will focus mostly on docs. I'll note that a few suggestions are only to reflow comments so they adhere to our limit of up to 80 columns on each comment line.

crates/typst-library/src/foundations/calc.rs
andylizi: Seems more tricky than I thought... To summarize the error paths we have here:
1. `i64::MIN / -1` will overflow because the result mathematically will be one larger than `i64::MAX`. This is the only case where it happens.
1. For `decimal`, it will return `CalculationResult::Overflow` when calculating the quotient of a very large number divided by a very small number.
1. `i64::MIN % -1` will overflow. This is a bit unintuitive because in mathematical terms the result should be zero[^1]. The Rust docs explained it as "invalid due to implementation artifacts"[^2]
1. For `decimal`, it will return `CalculationResult::Overflow` when calculating the remainder of a very large number divided by a very small number. This can only happen because internally "If the dividend is smaller than the divisor then try to scale that up first"[^3].

For Case 1 and 2, "the result is too large" message makes total sense: it's literally what happens. For Situation 4, the existing message "dividend too small compared to divisor" makes sense as well.

But for Case 3 (i.e. the integer case you mentioned,) "the result is too large" doesn't seem to fit well here. For users who aren't familiar with the quirks of x86 integer division instructions, I'm worried it'd be confusing. But just like you've pointed out, "dividend too small compared to divisor" doesn't work here either.

[^1]: https://blog.regehr.org/archives/175
[^2]: https://doc.rust-lang.org/std/primitive.i64.html#method.strict_rem
[^3]: https://github.com/paupino/rust-decimal/blob/1.39.0/src/ops/rem.rs#L84

andylizi: Hmm, what do you think about using "division overflowed" for all four cases? It's how both Rust docs and `rust_decimal` describe it.

andylizi: Thanks!

crates/typst-library/src/foundations/calc.rs
andylizi: Good catch, thanks!

crates/typst-library/src/foundations/calc.rs
PgBiel: I'm not sure, "divison overflowed" is not very instructive. I'd like a friendlier error message if possible. 
An alternative is to match on the error, and if the type is decimal then return "dividend too small compared to divisor", if the type is  integer then "the result is too large", and for `i64::MIN % -1` we could consider just returning zero or using a more generic error message.

andylizi: Just returning zero is a great idea, no reason to limit ourselves to quirky x86 behavior after all. Let's do that!

I've pushed the changes, but I was not able to write a unit test for the "dividend too small compared to divisor" error due to  paupino/rust-decimal#764

laurmaedje:
Thanks!

Issue #7411, opened 2025-11-18
Error when using page numbering with link to labeled text


T0mstone:
### Description

This compiles with 0.13.1, but produces a weird error with 0.14.0.

```typst
#set page(numbering: (n, total) => none)
#text("text")<label>
#link(<label>)[link]
```

The error in question:
```
error: missing argument: total
  ┌─ bug.typ:1:21
  │
1 │ #set page(numbering: (n, total) => none)
  │                      ^^^^^^^^^^
```

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
The link resolves the page numbering without total for its alt text. The page numbering function must be able to handle that.

T0mstone:
The `page.numbering` docs still say "If a function is passed, it will always receive both numbers.", so this turns out to have been a docs issue.

PR #7420, opened 2025-11-19
Fix `page.numbering` docs

T0mstone:
See https://github.com/typst/typst/issues/7411#issuecomment-3552352586

### DIFF
--- a/crates/typst-library/src/layout/page.rs
+++ b/crates/typst-library/src/layout/page.rs
@@ -218,8 +218,8 @@ pub struct PageElem {
     /// 1. The first number is the current page number.
     /// 2. The second number is the total number of pages. In a numbering
     ///    pattern, the second number can be omitted. If a function is passed,
-    ///    it may receive one argument in the context of links or references,
-    ///    or two arguments when producing the visible page numbers.
+    ///    it will receive one argument in the context of links or references,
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/layout/page.rs
```diff
@@ -218,7 +218,7 @@ pub struct PageElem {
     /// 1. The first number is the current page number.
     /// 2. The second number is the total number of pages. In a numbering
     ///    pattern, the second number can be omitted. If a function is passed,
-    ///    it will always receive both numbers.
+    ///    it must be able to handle both cases of receiving one or two arguments.
```
laurmaedje: I'm not sure "or" is accurate here, but "and" doesn't sound just right here. Perhaps "it must be able to handle both the case of receiving and of receiving two arguments"?

T0mstone: Is it also guaranteed that the single-arg version is only used for metadata and the actual numbers on the page will always have both? If so, that might also be good to add.

laurmaedje: Actually, the single-arg version will also be used by `ref(<label>, form: "page")`, so the docs were already wrong pre-0.14.

T0mstone:
How about this then?

laurmaedje:
That's good, thanks!

PR #2687, opened 2023-11-14
Allow `elem` synthesized fields to take a default value

Dherse:
Fixes #2530 and Fixes #2165.

PR #7422, opened 2025-11-19
Rehome lost test :)

isuffix:
He needed some help finding his friends, but now he's home :)

The test was added in #2825 as cleanup from the #2687 changes, but got mixed up when #3922 reorganized the tests.


laurmaedje:
Thanks! 😄

PR #7423, opened 2025-11-19
Improve the font features interface

### LINKED ISSUE #7385: Select additional salt glyphs
### Description

Fonts can have multiple alternative glyphs defined via stylistic alternatives (salt feature), but it seems that Typst always selects only the first one, even if we explicitly declare the second alternative via e.g. `#text(features: ("salt=2",))[text]`.

### Use Case

[Here](https://github.com/slavonic/Mezenets) is an example of a font that has multiple defined alternative glyphs, e.g. for the codepoint U+1CF7D.

cady-b:
## Overview

**This PR**:
1. Corrects our model of "Stylistic Alternates", [`text.alternates`](https://typst.app/docs/reference/text/text/#parameters-alternates), to a non-boolean feature (closes #7385).
2. Overhauls the parsing of raw [`text.features`](https://typst.app/docs/reference/text/text/#parameters-features), assuring Typst documents adhere to the invariants defined in the OpenType specification.

Both points have been discussed extensively in the linked issue, which acts as the motivation and design foundation for this PR.

Implementing Part 1 to match was fairly trivial; I expect mostly nits there. Part 2, even though secondary to the original issue, turned out a lot more involved. Laurenz has given [clear guidance](https://github.com/typst/typst/issues/7385#issuecomment-3544123126), which I hope to have captured most of. I still expect to iterate on the implementation.

### Stylistic Alternates
I've kept the doc-comment concise as this is quite technical and probably irrelevant to most users. The other part of this PR has resulted in new hints that should help guide users to a solution (though atm we don't give hints customized to specific features, i.e. with `salt`). I think https://github.com/typst/typst/issues/7385#issuecomment-3538693150 (a short section in the new docs on how fonts work) is worth considering.

<details>
<summary>outdated notes (see below)</summary>

Sadly we have no visual tests at the moment. The additional fonts from `typst-dev-assets` that have `salt` tables are (according to otfinfo): IBM Plex, Roboto, and Ubuntu. I've inspected the candidates with fontforge and couldn't find any `salt` mappings to multiple GIDs. The font that was brought up in the original issue is OFL, and making a bespoke one specifically for testing Typst would also be possible though.
</details>

Nvm, I apparently missed exactly one case in Libertinus Serif. There is a visual test now :)
The font is set explicitly because it does matter in this case and if it's ever changed, this test would have to be revised as well.

### Tag parsing
In general, this parser is build to accommodate UX, specifically error reporting. We _could_ throw "invalid byte at offset n" and tell the user to read the OpenType spec (which is more or less what `harfrust` does), but that doesn't feel very Typsty™︎.

That comes with the problem of the concepts themselves being quite technical however; ideally we should use as simple language as possible to describe the issue at hand, but I think it's fair assume some understanding of basic Typst Features such as array indexing and Unicode to not give watered down errors.

---

Currently the PR contains a hint on how the original issue's OP tried to remedy the situation `#text(features: ("salt=2",))`:
```
error: feature tag must be one to four bytes in length
  ┌─ demo.typ:1:16
  │
1 │ #text(features: ("salt=2",))[]
  │                 ^^^^^^^^^^^
  │
  = hint: found 6 bytes
  = hint: occurred in tag at index 0, 'salt=2'
  = hint: to set features with custom values, consider supplying a `dictionary`
```

But I'm not sure if it's not rather too pedantic. For the only existing test that changed (`#set text(features: ("tag", false))`), it actually sounds reasonable, but in a number of the new tests, it's not really applicable. I'm not sure how users tend to interact with this feature in practice.

We might still want to consider supporting this syntax in the future (see https://github.com/typst/typst/issues/7385#issuecomment-3539357642).

---

Tag parsing is fully tested only in rust, as it's not about the Typst API <sub><sup>(and bc rust allows me to more easily test many invocations 💔)</sup></sub>. The lossy parsing bug (https://github.com/typst/typst/issues/7385#issuecomment-3541931358) and some other sample cases are covered in the Typst test suite however.

## A disclaimer

I did have a look at https://github.com/googlefonts/fontations/blob/main/font-types/src/tag.rs. While our implementation is not a copy and somewhat different due to the focus on UX, given the constraints, we did converge on similar looking solutions; the test cases also appear similar. Fontations is dual licensed under Apache 2.0 and MIT, I can add it to the `NOTICE` file if required.

### DIFF
--- a/crates/typst-library/src/text/mod.rs
+++ b/crates/typst-library/src/text/mod.rs
@@ -610,7 +610,6 @@ pub struct TextElem {
     /// #set text(alternates: true)
     /// 0, a, g, ß
     /// ```
-    #[default(Alternates::default())]
     #[ghost]
     pub alternates: Alternates,
 
@@ -1259,8 +1258,7 @@ cast! {
         }) {
             bail!(
                 "feature tag may contain only printable ASCII characters";
-                hint: "you may refer to https://typst.app/tools/ascii-table/";
-                hint: "found invalid cluster {}", cluster.repr();
+                hint: "found invalid cluster `{}`", cluster.repr();
             )
         }
 
@@ -1302,7 +1300,7 @@ cast! {
                 // Append a hint if the value is a string containing the
                 // assignment operator `=` or another type was supplied.
                 if v.cast::<Str>().map_or(true, |v| v.as_str().contains('=')) {
-                    e.with_hint("to set features with custom values, consider supplying a `dictionary`")
+                    e.with_hint("to set features with custom values, consider supplying a dictionary")
                 } else {
                     e
                 }
@@ -1321,7 +1319,7 @@ cast! {
 }
 
 fn tag_hint_helper(index: usize, key: &impl Repr) -> EcoString {
-    eco_format!("occurred in tag at index {index}, {}", key.repr())
+    eco_format!("occurred in tag at index {index} (`{}`)", key.repr())
 }
 
 impl Fold for FontFeatures {
--- a/tests/suite/layout/inline/text.typ
+++ b/tests/suite/layout/inline/text.typ
@@ -66,39 +66,38 @@ fi vs. #text(features: (liga: 0))[No fi]
 
 --- text-features-non-ascii paged ---
 // Error: 21-30 feature tag may contain only printable ASCII characters
-// Hint: 21-30 you may refer to https://typst.app/tools/ascii-table/
-// Hint: 21-30 found invalid cluster "ƒ"
-// Hint: 21-30 occurred in tag at index 0, "ƒeat"
+// Hint: 21-30 found invalid cluster `"ƒ"`
+// Hint: 21-30 occurred in tag at index 0 (`"ƒeat"`)
 #set text(features: ("ƒeat",))
 
 --- text-features-bad-padding paged ---
 // Error: 21-30 spaces may only appear as padding following a feature tag
-// Hint: 21-30 occurred in tag at index 0, " tag"
+// Hint: 21-30 occurred in tag at index 0 (`" tag"`)
 #set text(features: (" tag",))
 
 --- text-features-empty-array paged ---
 // Error: 21-26 feature tag must be one to four characters in length
 // Hint: 21-26 found 0 characters
-// Hint: 21-26 occurred in tag at index 0, ""
+// Hint: 21-26 occurred in tag at index 0 (`""`)
 #set text(features: ("",))
 
 --- text-features-overlong-dict paged ---
 // Error: 21-41 feature tag must be one to four characters in length
 // Hint: 21-41 found 15 characters
-// Hint: 21-41 occurred in tag at index 0, "verylongfeature"
+// Hint: 21-41 occurred in tag at index 0 (`"verylongfeature"`)
 #set text(features: (verylongfeature: 0))
 
 --- text-features-array-kv paged ---
 // Error: 21-32 feature tag must be one to four characters in length
 // Hint: 21-32 found 6 characters
-// Hint: 21-32 occurred in tag at index 0, "feat=2"
-// Hint: 21-32 to set features with custom values, consider supplying a `dictionary`
+// Hint: 21-32 occurred in tag at index 0 (`"feat=2"`)
+// Hint: 21-32 to set features with custom values, consider supplying a dictionary
 #set text(features: ("feat=2",))
 
 --- text-features-bad-nested-type paged ---
 // Error: 21-35 expected string, found boolean
-// Hint: 21-35 occurred in tag at index 1, false
-// Hint: 21-35 to set features with custom values, consider supplying a `dictionary`
+// Hint: 21-35 occurred in tag at index 1 (`false`)
+// Hint: 21-35 to set features with custom values, consider supplying a dictionary
 #set text(features: ("tag", false))
 
 --- text-tracking-negative paged ---


crates/typst-library/src/text/mod.rs
```diff
@@ -1166,6 +1169,21 @@ impl Resolve for TextDir {
...
+pub struct Alternates(u32);
+
+cast! {
+    Alternates,
+    self => self.0.into_value(),
+    v: bool => Self(v as u32),
+    v: i64 => if let Ok(v) = v.try_into() {
+        Self(v)
+    } else {
+        bail!("value must be non-negative and cannot exceed {}", u32::MAX)
```
cady-b: Not sure what the most friendly way to convey this is; u32 is a hard requirement by OpenType.

crates/typst-library/src/text/mod.rs
```diff
@@ -724,7 +726,8 @@ pub struct TextElem {
     /// - If given an array of strings, sets the features identified by the
     ///   strings to `{1}`.
     /// - If given a dictionary mapping to numbers, sets the features
-    ///   identified by the keys to the values.
+    ///   identified by the keys to the values. This allows interacting with
+    ///   non-boolean features such as `swsh`.
```
cady-b, line 729: "non-boolean" is also used by MDN / W3C: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-feature-settings#values

We don't have a font with `swsh` sadly

crates/typst-library/src/text/mod.rs
laurmaedje: typically, we just say "number too large" 

https://github.com/typst/typst/blob/bed7b077bd5bce45c4e6fad45f95b0db56068603/crates/typst-library/src/foundations/int.rs#L438

just specifying `v: u32 => Self(v)` would do that and be good enough, imo

crates/typst-library/src/text/mod.rs
```diff
@@ -1231,6 +1249,50 @@ pub enum NumberWidth {
...
+                ),
+                None => e
+            })
+        }
+
+        let mut within_padding = false;
+        for (i, &v) in v.as_bytes().iter().enumerate() {
+            match v {
+                0x20 if i == 0 => {
+                    bail!("feature tag must contain at least one non-space character")
```
laurmaedje: this would also be thrown for `" abc"` which is in conflict with the error message. I think the message "spaces may only appear as padding at the end of a feature tag" would be good here, too.

crates/typst-library/src/text/mod.rs
```diff
@@ -1231,6 +1249,50 @@ pub enum NumberWidth {
...
+                }
+                0x20 => within_padding = true,
+                0x21 ..= 0x7E => if within_padding {
+                    bail!("spaces may only appear as padding at the end of a feature tag")
+                },
+                .. 0x20 | 0x7F .. => bail!(
+                    "feature tag must be printable ASCII";
+                    hint: "found: U+{v:04X}"
```
laurmaedje: `\u{..}` might be more Typst-y than U+...

crates/typst-library/src/text/mod.rs
```diff
@@ -1231,6 +1249,50 @@ pub enum NumberWidth {
...
+
+        if !(1..=4).contains(&v.len()) {
+            let e = error!(
+                "feature tag must be one to four bytes in length";
+                hint: "found {} bytes", v.len()
+            );
+
+            return Err(match v.grapheme_indices(true).find(|(_, v)| v.len() > 1) {
+                Some((i,v)) => e.with_hint(
+                    eco_format!("found multi-byte cluster '{v}' in grapheme {}", i+1)
```
laurmaedje: `'{v}'` should be `` `{}` `` and then `v.repr()`. also, this would not be needed at all if we did check for ASCII first and throw that error then. so perhaps reorder the ASCII check to the top and just drop this?

crates/typst-library/src/text/mod.rs
```diff
@@ -1244,18 +1306,26 @@ cast! {
...
-            Ok((Tag::from_bytes_lossy(tag.as_bytes()), 1))
-        })
+        .enumerate()
+        .map(|(i, v)| Ok((
+            v.clone().cast::<Tag>()
+                .hint(eco_format!(
+                    "occurred in tag at index {i}, '{}'",
+                    v.cast::<EcoString>().unwrap_or_default()
+                ))
+                .hint("to set features with custom values, consider supplying a `dictionary`")?,
```
laurmaedje: maybe only add this hint if the tag contains an `=`?

crates/typst-library/src/text/mod.rs
```diff
@@ -1244,18 +1306,26 @@ cast! {
...
         .into_iter()
-        .map(|v| {
-            let tag = v.cast::<EcoString>()?;
-            Ok((Tag::from_bytes_lossy(tag.as_bytes()), 1))
-        })
+        .enumerate()
+        .map(|(i, v)| Ok((
+            v.clone().cast::<Tag>()
+                .hint(eco_format!(
+                    "occurred in tag at index {i}, '{}'",
```
laurmaedje: here, we should also use `repr` and not single quotes (since these are not valid for strings in Typst)

crates/typst-library/src/text/mod.rs
```diff
@@ -1244,18 +1306,26 @@ cast! {
...
+                    v.cast::<EcoString>().unwrap_or_default()
```
laurmaedje: the unwrap_or_default will produce a rather weird hint `occurred in tag at index 0, ''`. we should not add the value if it's not a string maybe? also maybe ``occurred in tag `"wrong"` at index 0`` would be a bit friendlier than the comma?

crates/typst-library/src/text/mod.rs
```diff
@@ -1244,18 +1306,26 @@ cast! {
...
         .into_iter()
-        .map(|(k, v)| {
-            let num = v.cast::<u32>()?;
-            let tag = Tag::from_bytes_lossy(k.as_bytes());
-            Ok((tag, num))
-        })
+        .enumerate()
+        .map(|(i, (k, v))| Ok((
+            k.clone().into_value().cast::<Tag>().hint(
+                eco_format!("occurred in tag at index {i}, '{k}'")
```
laurmaedje: this could also happen for the value. I suppose we could move it into a helper function.

maybe there is also a potential for sharing code with the `Array` case

crates/typst-library/src/text/mod.rs
```diff
@@ -1485,4 +1556,36 @@ mod tests {
     fn test_text_elem_size() {
         assert_eq!(std::mem::size_of::<TextElem>(), std::mem::size_of::<EcoString>());
     }
+
+    #[test]
+    fn test_text_tag_parsing() {
```
laurmaedje, line 1556: why test this from Rust? as it does not test the error message, we need to test pretty much everything from Typst anyways and then it seems redundant.

crates/typst-library/src/text/mod.rs
cady-b: Hmm, I see.

The ASCII check when we're iterating the bytes is the actual validation step. The multi-byte cluster hint is mostly a courtesy specifically for errors with length, not for correctness:
```typ
--- text-features-non-latin paged ---
// Error: 21-30 feature tag must be one to four bytes in length
// Hint: 21-30 found 5 bytes
// Hint: 21-30 found multi-byte cluster 'а' in grapheme 3
// Hint: 21-30 occurred in tag at index 0, 'feаt'
// Hint: 21-30 to set features with custom values, consider supplying a `dictionary`
#set text(features: ("feаt",))
```
Though I think I'd rather remove it entirely and keep the rest as is than make big changes here like iterate twice to first check for valid ASCII, but I'll think about it a bit more.

crates/typst-library/src/text/mod.rs
cady-b: I just learned `repr()` is easy to call from rust :sweat_smile: The unwrap is there because casting to a string first was the only way I knew how to print a `Value`.

 With `repr()` being infallible and always relevant to the source, however, it might be fine to print unconditionally? The type error is already be annotated (i.e. `expected string, found boolean`) either through `cast!` of `Tag` taking `Str` or the dictionary enforcing `Str`.

---

The rationale behind the phrasing is that both strings and arrays can reasonably have an index. We explicitly mean the nth tag of the array, not within the tag (i.e. "wrong") at byte / character n.

I guess it's not too difficult to deduce; I'll change it if you'd prefer.

crates/typst-library/src/text/mod.rs
cady-b: I'm not sure what "this" is referring to, sadly, given my understanding of the above case.

If you'd prefer a helper function, I could use a pointer on how those are typically introduced (I assume just sprinkling a `fn helper()` in the module somewhere close isn't it :sweat_smile:)

crates/typst-library/src/text/mod.rs
cady-b: These tests are to ensure the parsing code handles all distinct cases correctly from the spec's point of view; for that, the error messages are irrelevant.

The cast to Tag isn't directly accessible from Typst, only through `text.features`. As such, having all these cases would be quite convoluted. The test suite contains some sample cases with error messages and makes sure checks of some sort are happening internally. I think I'll add a few more as well, but adding everything would bloat the file quite a bit. If you want me to add tests for every case, I'm not sure if that'd be fine or if we'd want a new file in the directory or whatever.

crates/typst-library/src/text/mod.rs
cady-b: I noticed that my initial idea for the ASCII check after this part (one above in the review) was actually bad; reinterpreting utf-8 as a codepoint is incorrect, even for single-byte values. To report something sensible, it might indeed make sense to first check that the graphemes are single-byte before dropping to ASCII level checks.

That'd leave us with a more specialized range check later (where that mistake was). Now just for the control characters within ASCII, which are disallowed in Tags. Would you be comfortable linking to https://typst.app/tools/ascii-table/ in a hint? It certainly helped me a lot figuring all this out!

laurmaedje: I'm not sure I understand the problem yet. I'd just check & report `feature tag must be printable ASCII` first if it's not printable ASCII (without a hint; I don't think a hint is needed as you need to kinda go out of your way to even reach this). And then afterwards check for the length and no need to consider multi-byte clusters since we're already restricted to printable ASCII.

crates/typst-library/src/text/mod.rs
laurmaedje: Calling repr always is okay

crates/typst-library/src/text/mod.rs
laurmaedje: I meant that `v.cast::<u32>()?` could also profit from the hint because otherwise you also get `number too large` without any indication of which tag it happened in.

otherwise, yeah helper function would just be an `fn` somewhere close below

crates/typst-library/src/text/mod.rs
laurmaedje: It's fine to leave it in Rust

crates/typst-library/src/text/mod.rs
cady-b: I think we're more or less on the same page.

An ascii check should come first and all logic after should assume ascii as a given (especially the length check). That part is clear to me and would fit into either version; I was pondering later implications of checking _printable_ ascii already.

I do really like how the match statement (after the length check) that is iterating the bytes is explicitly handling the entire range at the moment; to me that feels reassuring. If we check not only for ascii but printable ascii up top, this would fall apart as the range is partially checked by the time we get there, and we wouldn't want a duplicate error in dead code, of course. Also, checking for ascii is nice via `str::is_ascii()`, while a printable check is more manual.

I did play around with it though, and must say the proposed version is quite okay still. It wouldn't be my preference, but I will push that version, and you can take a look.

---

Regarding the hint, I'm not sure whether I understand. If we check for ascii first, obviously the length check doesn't need to go through the clusters again. But I do like having the hint that's currently there around, just moved to the new check. A Unicode character sneaking in isn't much more esoteric than the rest of these, I'd say. I'll keep it in the next version I push, and then you can see it in context with everything else applied. If you still prefer, I'll gladly drop it.

cady-b:
I've pushed some revisions. The changes in tag parsing are in a separate commit to facilitate another round of review. To keep the semantics, it'd just have to be fixed up to the one below.

Re the phrasing from https://github.com/typst/typst/pull/7423#discussion_r2547066187; I've kept it for now since there were no objections since my comment, but I can still change it ofc. To be honest, I've been staring at it for so long that my judgement may be very off by now. I think the current version may not be super clear in the regard I mentioned either.

edit: rebased and updated for [1993b1a](https://github.com/typst/typst/commit/1993b1a61a04bb2d3925e27b3f4f8b278bc325e8)

laurmaedje:
I'm pretty happy with it now. I only made a few very minor final tweaks. In particular, I removed the link to the ASCII table in the end because it didn't seem _that_ useful specifically for "printable ASCII" and I made sure that there are backticks around all `repr()` output in the diagnostics (as usual).

Thanks!

PR #7429, opened 2025-11-21
Don't wrap `math.op` in unnecessary frame fragment

mkorje:
Changed test has slightly better (tighter) spacing now
| Old | New | Diff |
| - | - | - |
| old | new | diff |

(green is new and red is old in the diff)


laurmaedje:
Nice!

PR #7435, opened 2025-11-22
Add a math IR

mkorje:
Depends on ~~#7473~~, ~~#7605~~, #7659, ~~[typst-dev-assets#19](https://github.com/typst/typst-dev-assets/pull/19)~~

**Things left to do:**
- [x] sort out glyph stretching
  - [x] add tests to clarify mixed lr/stretch behaviour
  - [x] resolve the now failing tests (see 2. below)
- [x] fix dtls feature
- [x] fix flac feature
- [x] figure out correct alignment point behaviour
  - [x] make implementation actually correct
  - [x] refactor multiline handling to share with mathml output in #7436 
        **NOTE:** will maybe defer this to #7436 or another PR
- [x] use `MathBuffer` from #7473
- [x] tidy up commit history before merge
- [x] double-check styles passed to `<frag>.font(...)` calls
- [x] comments
- [x] double check for any redundant code & visibility of things.
- [x] performance

**CHANGES:** There are (hopefully) only two changes in behaviour with this PR. Tests cases to observe these changes were added in the third last commit https://github.com/typst/typst/pull/7435/commits/0b13df4df1357661831f33fa64d53a97ccd8e231 and then updated in the final commit https://github.com/typst/typst/pull/7435/commits/ac35ce90a61ba37e1f16dbeecce4a5e6c0c4197e.
1. Nested `math.lr` calls now affect the size of the `math.mid` elements in their body. See the test `math-lr-nested-mid`. This is an improvement/fix.
2. There have been some slight (user-facing) changes to how nested `math.stretch` and `math.lr` calls interact with regard to relative sizes and attachments: sizes are folded inward now, and `math.lr` only has an effect on things that can stretch vertically.
    + **Positive:** See the tests `math-stretch-nested-vertical-attach`, `math-stretch-lr-nested-vertical-attach`, `math-stretch-lr-nested-vertical`, `math-stretch-vertical-large-class`. This is an improvement/fix.
    + ~~**Negative:** A relative length at the outer most `math.stretch`/`math.lr` call may not always be preserved. Thus, (when there are attachments) the relative length won't be relative to the size of the attachments. See the tests `math-stretch-lr-nested-horizontal-attach`, `math-stretch-nested-horizontal-attach`, `math-stretch-horizontal-attach`. This change is either bad or of no concern.
    The new behaviour is a bit inconsistent, but this nested stuff is all fairly niche anyway.~~ Fixed


mkorje:
This is ready for a first review.

~~Only problem left is the handling of the `dtls` feature to style `dotless.i` and `dotless.j`. I'm applying the feature unconditionally atm, but that is an issue for fonts that don't support the feature. (e.g. In Libertinus Math, `dotless.i` is impossible to actually get :sweat_smile:.)~~

crates/typst-layout/src/math/mat.rs
```diff
@@ -1,156 +1,47 @@
...
-    };
-    layout_delimiters(ctx, styles, frame, open, close, span)
-}
-
-/// Lays out a [`MatElem`].
-#[typst_macros::time(name = "math.mat", span = elem.span())]
-pub fn layout_mat(
-    elem: &Packed<MatElem>,
+/// Layout a [`TableItem`].
+#[typst_macros::time(name = "math.vec", span = props.span)]
```
laurmaedje: This is a bit odd. This will just create three nested timing scopes. Same in at least one other place.

crates/typst-layout/src/math/line.rs
laurmaedje, line 1: The name of this file could be adjusted. `mat.rs`, too, maybe more.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -0,0 +1,650 @@
...
+    }
+
+    pub fn styles(&self) -> Option<StyleChain<'a>> {
+        match self {
+            Self::Component(comp) => Some(comp.styles),
+            _ => None,
+        }
+    }
+
+    pub fn get_mid_stretched(&self) -> Option<bool> {
```
laurmaedje: I'd drop the `get_` prefix on these

crates/typst-library/src/math/ir/item.rs
```diff
@@ -0,0 +1,650 @@
...
+    pub open: Option<MathRun<'a>>,
+    pub close: Option<MathRun<'a>>,
+    pub body: MathRun<'a>,
+    pub balanced: bool,
+    pub target: Rel<Abs>,
+    pub short_fall: Em,
+}
+
+impl<'a> FencedItem<'a> {
+    #[allow(clippy::too_many_arguments)]
```
laurmaedje: we could also allow that module-wide

crates/typst-library/src/math/ir/item.rs
```diff
@@ -0,0 +1,650 @@
...
+impl<'a> RadicalItem<'a> {
+    pub(crate) fn create(
+        radicand: MathRun<'a>,
+        index: Option<MathRun<'a>>,
+        sqrt: MathRun<'a>,
+        styles: StyleChain<'a>,
+        span: Span,
+        bump: &'a Bump,
+    ) -> MathItem<'a> {
+        let kind = MathKind::Radical(bump.alloc(Self { radicand, index, sqrt }));
```
laurmaedje: I fear that the current implementation might leak memory because `Bumpalo` does not call `Drop` on the contents of an arena, but this internally holds an `EcoVec` that would never be freed again. The cleanest solution is probably to use `bumpalo::Box` to get `Drop` (thought that might conflict with the `Clone` impl).

crates/typst-library/src/math/ir/run.rs
```diff
@@ -0,0 +1,280 @@
...
+                    MathItem::Component(new_comp)
+                }
+                other => other.clone(),
+            })
+            .collect();
+
+        MathRun { items, styles: self.styles.chain(style) }
+    }
+
+    pub fn is_multiline(&self) -> bool {
```
laurmaedje: This seems to have grown a bit: Is that new logic a fix or was it not necessary in the old code?

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -0,0 +1,1321 @@
...
+        locator: &mut SplitLocator<'a>,
+        styles: StyleChain<'a>,
+    ) -> SourceResult<MathRun<'a>> {
+        let mut context = Context::new(engine, locator, &self.arenas);
+        context.resolve_into_run(&elem.body, styles)
+    }
+}
+
+/// The math IR builder.
+struct Context<'a, 'v, 'e> {
```
laurmaedje: I think the naming is a bit confusing with different `MathContext`s in `typst-layout` and `typst-html` and then a generic `Context` here. Having the layout and HTML things be named the same seems fine to be as they fulfill the same role, but maybe `MathResolver` would work well for the `Context` here. The `Arenas` could just be directly passed into it like for `typst_realize`, turning the current `MathResolver` obsolete. Or we could keep it and rename it. But right now it's inconsistent with `typst_realize`, so if we like that, we might want to introduce something similar there, too.

laurmaedje (commented):
I took an initial look just now and I think this looks great! 

The changes look overall quite straightforward and clear, in particular I like how `typst-layout/src/math` is really only concerned with layout now. Also lovely how simple https://github.com/typst/typst/pull/7436 got as a result of this (I know that it's not fully complete, but what's there looks great, too).

I left a few comments, but barely have remarks. One thing I would like to see before merge are a few more doc comments.

I didn't double-check every piece of logic as that would just take way too much time, but I took a look at the big picture. If you want an extra pair of eyes on particular pieces of the code, let me know.

crates/typst-layout/src/math/mat.rs
mkorje: I forgot to tidy those up, will fix that.

crates/typst-library/src/math/ir/run.rs
mkorje: To keep the behaviour of linebreaks in Left/right  I need to recurse into the FencedItem's contents in the IR. In typst-layout before the fragments were just pushed flatly so the linebreaks were at the top level.

crates/typst-library/src/math/ir/resolve.rs
mkorje: How about just `Resolver`? The reason I even created the separate `MathResolver` was because it felt a bit clunky for the caller to have to deal with the `Arenas`. It feels a bit nicer, in my opinion at least, that you just need to do `MathResolver::new()` and then call resolve on it.

crates/typst-library/src/math/ir/item.rs
mkorje: I'm going to switch the MathRun's items to be held in the arenas. That should fix things, I think.

crates/typst-library/src/math/ir/resolve.rs
mkorje: Actually, as I need to use the arenas in typst-layout's MathContext now I might just remove the MathResolver struct and call this Context MathResolver instead.

crates/typst-library/src/math/ir/item.rs
mkorje: I'd appreciate it if you could double check that I haven't missed any field and their types in `item.rs` that could cause a memory leak since `Drop` is not called.

`Augment` (in `TableItem`) and `FixedStroke` (in `CancelItem`) both might be problematic, not sure how to deal with them though. There's also `CancelAngle` in `CancelItem` preventing `Copy` from being derived for it.

crates/typst-library/src/math/ir/resolve.rs
laurmaedje: That seems to _resolve_ the topic.

crates/typst-library/src/math/ir/item.rs
mkorje: I've culled the `MathRun` struct entirely now btw.

laurmaedje: > Augment (in TableItem) and FixedStroke (in CancelItem) both might be problematic, not sure how to deal with them though. There's also CancelAngle in CancelItem preventing Copy from being derived for it.

Those (and also e.g. `GlyphItem::text`) are indeed problematic. I'd suggest removing the `Clone` impls and using bump boxes everywhere like this:
```rust
use bumpalo::boxed::Box as BumpBox;

...

#[derive(Debug)]
pub enum MathKind<'a> {
    ...
    Glyph(BumpBox<'a, GlyphItem>),
    ...
}
```
From a very quick experiment, this seems to work.

laurmaedje: The bump allocated `&'a [..]` slices would need to become `BumpBox<'a, [..]>` then or use `BumpVec`.

laurmaedje:
> The new behaviour is a bit inconsistent, but this nested stuff is all fairly niche anyway.

I tried to understand what's happening, but I'm not sure I really get it. Could you describe what is now inconsistent and how it was more consistent before in a few more words?

laurmaedje:
The main thing that's still to sort out is the bump allocator handling (see https://github.com/typst/typst/pull/7435#discussion_r2662023662).

Beyond that, things look good (the changes since my last review seem to be _relatively_ minor in the grand scheme of things; mostly related to the MathRun struct). Though it's hard to get a proper diff without unrelated changes from main. But, as mentioned before, if there's something in particular that you'd like to have reviewed with some extra scrutinity, let me know.

crates/typst-library/src/math/ir/item.rs
mkorje: Should all be fixed now. Please double-check though that I haven't missed anything!

mkorje:
> I tried to understand what's happening, but I'm not sure I really get it. Could you describe what is now inconsistent and how it was more consistent before in a few more words?

Hopefully this comparison makes it clearer. The inconsistency I was referring to is the 2nd arrow on the first line.

| | |
|-|-|
| `main` | playground |
| This PR | playground |
```typ
$
  stretch(stretch(->, size: #2em), size: #200%) "vs" stretch(stretch(->, size: #2em), size: #200%)^"looooooong"
$
but
$ stretch(->, size: #200%) "vs" stretch(->, size: #200%)^"looooooong" $
```

mkorje:
> Beyond that, things look good (the changes since my last review seem to be _relatively_ minor in the grand scheme of things; mostly related to the MathRun struct). Though it's hard to get a proper diff without unrelated changes from main. But, as mentioned before, if there's something in particular that you'd like to have reviewed with some extra scrutinity, let me know.

I've finished off everything now. Once you've had a look at the recent changes I'll rebase and clean up the commit history. I'll go through it again myself later today and flag anything specific I'd like you to have a look at.

tests/ref/render/math-primes-scripts.png
mkorje, line 1: What's changed in this test:
One of the `<` which is wrapped in `op` gets large math class. Previously we only centered a large math class glyph if its default class was large. No idea if this was intentional, but this is the only consequence in the test suite. Its nearly impossible to see, but one of the `<` is now centered vertically.

crates/typst-layout/src/math/text.rs
```diff
@@ -113,70 +103,107 @@ fn layout_inline_text(
...
+            let text = item
+                .text
+                .graphemes(true)
+                .flat_map(|c| undo_dotless(c).unwrap_or(c).chars())
+                .collect();
+            (styles, text)
+        }
+    } else {
+        (styles, item.text.clone())
+    };
```
mkorje: This is sort of ugly, but was the only way I could figure out how to deal with the styling of `dotless.i` and `dotless.j` with the `dtls` feature. During IR creation, I don't know the font so just assume that it has `dtls` and swap the dotless glyphs to their non-dotless form and then style it. Then here during layout I deal with undoing this if the font doesn't actually support the `dtls` feature.

crates/typst-layout/src/math/run.rs
mkorje, line 1: This filename is a bit inaccurate now that `MathRun` no longer exists. Any ideas for a better name?

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -0,0 +1,1315 @@
...
+                for item in fenced.body.as_slice() {
+                    if item.mid_stretched() == Some(true) {
+                        item.set_stretch(stretch);
+                    }
+                }
+            } else {
+                one.set_y_stretch(StretchInfo::new(height.abs.into(), DELIM_SHORT_FALL));
+            }
+            return Ok(());
+        }
```
mkorje, line 726: This nested business is sort of ugly, but necessary to keep the behavior of nested `lr` working.

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -0,0 +1,1315 @@
...
+        {
+            if let Some(d) = try_dotless(c) {
+                enable_dtls = true;
+                c = d;
+            }
+            to_style(c, MathStyle::select(c, variant, bold, italic)).collect()
+        } else {
+            cluster.into()
+        };
+
```
mkorje: This has slightly changed compared to before. Instead of iterating over every char in the cluster and applying styling, I'm just applying it if it is a single character. This was motivated by needing to "undo" the dotless thing later on in layout.

crates/typst-layout/src/math/mod.rs
```diff
@@ -397,254 +399,186 @@ impl<'a, 'v, 'e> MathContext<'a, 'v, 'e> {
...
+            ctx.fragments.iter().rposition(|f| !matches!(f, MathFragment::Tag(_)))
+            && matches!(ctx.fragments[i], MathFragment::Align)
+        {
+            // Skip a single alignment point (if one exists) when placing
+            // spacing on the left.
+            ctx.fragments.insert(i, frag);
+        } else {
+            ctx.push(frag);
+        }
+    }
```
mkorje, line 502: This _should_ be the translation of the change in behaviour from https://github.com/typst/typst/pull/7473 to this PR.

crates/typst-layout/src/math/run.rs
laurmaedje: The name seems fine to me as it's conceptually still laying out a run of math items.

crates/typst-layout/src/math/text.rs
laurmaedje: It's hard for me to judge as I'm not as familiar with the logic, but would it be possible to put the responsibility of dealing with the dotless form entirely on the post-IR backend (i.e. layout and html), such that we don't have to undo it? Or are there negative consequences from having the unprocessed form in the `GlyphItem`?

crates/typst-library/src/math/ir/resolve.rs
laurmaedje: Is this user-visible?

crates/typst-library/src/math/ir/resolve.rs
laurmaedje: Yeah, the nested handling is a bit annoying, but not the fault of this PR

tests/ref/render/math-primes-scripts.png
laurmaedje: Seems fine

laurmaedje:
> Hopefully this comparison makes it clearer. The inconsistency I was referring to is the 2nd arrow on the first line.

It is very minor indeed, but it still feels a bit wrong to regress this. Could we perhaps fix it by maintaining two fields in `StretchInfo`: One pre-folded that's used if there is no `relative_to_size` and one that's used if there is one? Or something like that; maybe it could even be expressed in a smarter way by directly combining with `relative_to` or `font_size`. (I must admit, I don't entirely understand how everything works together, in particular, when only `set_stretch_relative_to` is called and when also `set_stretch_font_size`).

laurmaedje:
> Once you've had a look at the recent changes I'll rebase and clean up the commit history.

Sounds good. If we're going to keep the commit history, could you make sure that each commit by itself passes the tests? That makes bisecting nicer later on.

crates/typst-library/src/math/ir/resolve.rs
mkorje: I don't think so. Something like the grapheme cluster `g̈` (0067 ( g ) LATIN SMALL LETTER G, 0308 ( ◌̈ ) COMBINING DIAERESIS) might have, but I don't think it's possible to get that to work on the latest release.
```typ
#let cluster = symbol("g̈")
$ bb(cluster) $
```
The above triggers `Error: shaping the text `"𝕘\u{308}"` yielded more than one glyph`.

mkorje: Regardless, I think this is the better behaviour for now until we sort out https://github.com/typst/codex/issues/134 properly.

crates/typst-layout/src/math/text.rs
mkorje: That should work. I was worried about needing to pass the styling information, but I have the styles anyway! I will try it out.

crates/typst-library/src/math/ir/resolve.rs
laurmaedje: This would change behavior and compiles on 0.14.2:
```typ
#let s = symbol("u\u{fe00}")
$ #s $
```

I suppose, that's a bit of garbage-in garbage-out, but I wonder whether there could possibly be a situation where a sensible thing could be in the cluster and yet styling still makes sense.

crates/typst-layout/src/math/mod.rs
mkorje: The new test from that PR is failing, I will fix that up when I rebase things.

mkorje:
> It is very minor indeed, but it still feels a bit wrong to regress this. Could we perhaps fix it by maintaining two fields in `StretchInfo`: One pre-folded that's used if there is no `relative_to_size` and one that's used if there is one?

That worked! Behaviour should be the same as `main` now.

crates/typst-layout/src/math/text.rs
mkorje: I got it all working.

crates/typst-library/src/math/ir/resolve.rs
mkorje: Since I got dtls handling moved to post-IR, I reverted this and added a test.

mkorje:
Everything has been addressed and fixed. Let me know when I can clean up the commit history.

laurmaedje:
I just merged https://github.com/typst/typst/pull/7540, which might introduce minor conflicts. Other than that, I don't think anything should block the cleanup. From my side, this PR would be ready to land as soon as it's ready from yours.

mkorje:
The `math-align-spacing-ignorant` test has changed a bit, need to investigate...

| | `math-align-spacing-ignorant` |
|-|-|
| Pre #7473 | pre |
| `main` | main |
| This PR | pr |

laurmaedje:
Thank you so much. Great work! 🧮 🎉

PR #7436, opened 2025-11-22
MathML Core output for HTML export

mkorje:
Supersedes #7206.

### Stylesheet

To have the output in a browser match Typst's paged output exactly, we'd ideally override parts of the [user agent stylesheet](https://www.w3.org/TR/mathml-core/#user-agent-stylesheet) instead of having to supply things to each instance of an element. Since doing the former isn't possible yet and the latter isn't great (we are initially focusing on semantics after all), I'm ignoring all of this stuff. There is one exception, though: I need to disable the `text-transform: math-auto` property for `mi` (by using the `mathvariant="normal"` attribute) to stop the browser from auto italicizing things.

### What Should the Width of a Space Be?

We occasionally need to emit a single space, but we cannot obtain the font's space width! So we may need to choose something. Some choices I've found elsewhere:
- `4/18em` in https://github.com/jgm/texmath/blob/aee6b494595a58dbbf769264061876778f7aa269/src/Text/TeXMath/Shared.hs#L251C1-L251C37

### Specifications

- [MathML Core](https://www.w3.org/TR/mathml-core/)
- [MathML 4.0](https://www.w3.org/TR/mathml4/)
- [MathML 3.0](https://www.w3.org/TR/MathML3/)

### Other Implementations

- [LuaMML](https://github.com/latex3/luamml/)
- [LaTeXML](https://math.nist.gov/~BMiller/LaTeXML/)
- [MathJax](https://www.mathjax.org/)
- [KaTeX](https://katex.org/)
- [Temml](https://temml.org/)
- [AsciiMath2ML](https://johtela.github.io/asciimath2ml/)
- [TeXZilla](https://fred-wang.github.io/TeXZilla/)
- [texml-to-html](https://github.com/AmerMathSoc/texml-to-html)

### Fonts

- https://mathfonts.github.io/

## Browser Problems

Below I collect issues in the main 3 browser engine's MathML implementations that cause the MathML generated by this PR to display incorrectly. Whilst I do think we should work around these where possible, for my initial implementation I have not done so.

### WebKit


crates/typst-html/src/mathml.rs
```diff
@@ -17,7 +22,556 @@ pub(crate) fn convert_math_to_nodes(
...
+    if let Some((tag, other_children)) = match (t, b) {
+        (None, None) => None,
+        (Some(t), None) => Some((tag::mover, eco_vec![t])),
+        (None, Some(b)) => Some((tag::munder, eco_vec![b])),
+        (Some(t), Some(b)) => Some((tag::munderover, eco_vec![b, t])),
+    } {
+        let mut children = eco_vec![base];
+        children.extend(other_children);
+        base = HtmlElement::new(tag).with_children(children).into();
+    }
```
isuffix, line 910: I got curious about this from my other work on scripts with primes, and when both under/over and sub/super scripts are present, Firefox and Chrome seem to render better when doing under/over inside sub/super. In particular if you have `attach(sum, tr: 2, t: n)`, placing the `mover` outside the `msup` causes the `n` to be centered around `sum^2`, which looks quite bad.



```html
<math display="block">
  <mover><msup><mo>∑</mo><mn>2</mn></msup><mi>n</mi></mover>
  <mo>≠</mo>
  <msup><mover><mo>∑</mo><mi>n</mi></mover><mn>2</mn></msup>
</math>
```

However, this makes me worried about how users would typeset annotations on something with a normal attachment, e.g. `attach(limits(x^2), b: "squared")`. So I think we should consider updating the attachment model to improve this generally, although maybe after this PR or after my other fix for scripts lands.

mkorje:
Should be ready for an initial review now.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,614 @@
...
+
+    // TODO: maybe only add this if the base is an i or j, or only add when disabling.
+    // let dtls = if elem.dotless.get(styles) { "on" } else { "off" };
+    // if !accent.is_bottom() {
+    //     base.to_packed_mut::<HtmlElem>().unwrap().push_attr(
+    //         crate::attr::style,
+    //         eco_format!("font-feature-settings: 'dtls' {};", dtls),
+    //     );
+    // }
+
```
mkorje: Firefox seems to be the only browser that applies the `dtls` tag. We could add some inline styles to make it work on other browsers, and this would also make disabling it work too on Firefox.

But maybe this is erring away from the semantics focus, and so should be deferred?

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,614 @@
...
+    if let Some(close) = item
+        .close
+        .as_ref()
+        .map(|x| ctx.handle_into_node_with_only_form(x, Form::Postfix))
+        .transpose()?
+    {
+        children.push(close);
+    }
+
+    ctx.push(HtmlElement::new(tag::mrow).with_children(children));
```
mkorje: We shouldn't _always_ need to wrap this in an `mrow`, but browsers don't follow the spec and insert an inferred `mrow` in situations like `<msqrt>...</msqrt>`, `<mtd>...</mtd>`, and the top-level `<math>...</math>`. See, for example, https://bugzilla.mozilla.org/show_bug.cgi?id=236963 and related issues.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,614 @@
...
+    // if !accent.is_bottom() {
+    //     base.to_packed_mut::<HtmlElem>().unwrap().push_attr(
+    //         crate::attr::style,
+    //         eco_format!("font-feature-settings: 'dtls' {};", dtls),
+    //     );
+    // }
+
+    // TODO: convert accent char to non-combining, then lookup with postfix (or infix?) form for stretchy
+    // Should surpress "Text run starts with a composing character" warnings from validator.
+    let accent = ctx.handle_into_node_with_only_form(&item.accent, Form::Postfix)?;
```
mkorje: The comment here is outdated and really just there to flag that, for the HTML produced by
```typ
$ hat(x) $
```
the validator https://validator.w3.org/nu/#file warns "Warning: Text run starts with a composing character."

No idea if this is of any real concern.

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <math display="block">
      <mover accent="true">
        <mi>𝑥</mi>
        <mo>̂</mo>
      </mover>
    </math>
  </body>
</html>
```

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,614 @@
...
+    let MathItem::Component(comp) = item else {
+        match item {
+            MathItem::Spacing(amount, _, _) => ctx.push(
+                HtmlElement::new(tag::mspace)
+                    .with_attr(attr::width, eco_format!("{}", css::length(*amount))),
+            ),
+            MathItem::Space => ctx.push(HtmlElement::new(tag::mspace).with_attr(
+                attr::width,
+                eco_format!("{}", css::length(Em::new(0.2222222222222222))),
+            )),
```
mkorje: Need to decide what we should pick for the length of a space.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,608 @@
...
+        // it can be nested in MathML Core (as long as it is itself
+        // within a MathML token element).
+        MathKind::Box(_) => {}
+        MathKind::External(item) => {
+            ctx.engine.sink.warn(warning!(
+                props.span,
+                "{} was ignored during MathML export",
+                item.content.elem().name()
+            ));
+        }
```
mkorje: This all needs sorting:
- Warn for math elements that are ignored OR consider a polyfill.
- Enforce that phrasing content is the only non-math content allowed.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,608 @@
...
+    ctx: &mut MathContext,
+    props: &MathProperties,
+) -> SourceResult<()> {
+    // math.op should be mi instead of mtext.
+    // TODO: reconsider if this is the best way to do this...
+    let tag = match props.class {
+        MathClass::Large => tag::mi,
+        _ => tag::mtext,
+    };
+    ctx.push(
```
mkorje: Maybe this isn't the best way to differentiate things, but it seems nicest (and good enough), especially since `math.op` can take arbitrary content...

mkorje: This is irrelevant now (or will be, when I push a new commit soon). `math.op` is `mo` now.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,620 @@
...
+  padding: 0;
+}
+
+mtable.multiline-equation.aligned-equation mtd:nth-child(odd) {
+  text-align: -webkit-right;
+}
+
+mtable.multiline-equation.aligned-equation mtd:nth-child(even) {
+  text-align: -webkit-left;
+}";
```
mkorje: Need to finalise this: classes' names, vertical spacing, etc.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,620 @@
...
+    let mut fence = false;
+    let mut separator = false;
+    let mut largeop = false;
+    let tag = match props.class {
+        MathClass::Normal
+        | MathClass::Alphabetic
+        | MathClass::Special
+        | MathClass::GlyphPart
+        | MathClass::Space => {
+            let mathvariant = will_auto_transform(text).then_some("normal");
```
mkorje, line 774: We could add something like this to the `style` tag instead of needing to add `mathvariant="normal"` explicitly always:
```css
mi {
  text-transform: none;
}
```

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,751 @@
...
+    let separator =
+        (separator != is_separator(text)).then(|| eco_format!("{}", separator));
+
+    let largeop = (largeop != info.properties.contains(Properties::LARGEOP))
+        .then(|| eco_format!("{}", largeop));
+    // We don't use movablelimits as we handle the positioning ourselves.
+    let movablelimits = (info.properties.contains(Properties::MOVABLELIMITS))
+        .then(|| eco_format!("false"));
+
+    // TODO: symmetric, maxsize
```
mkorje: We don't need to do anything for `maxsize` I think, not sure about `symmetric` though yet, need to double-check.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,751 @@
...
+                .filter(|item| !item.is_ignorant())
+                .filter(|item| !is_space_like(item))
+                .count()
+                == 1
+                && group
+                    .items
+                    .iter()
+                    .filter(|item| !item.is_ignorant())
+                    .any(|item| is_embellished_operator(item))
+        }
```
mkorje, line 1293: This is a little sloppy sorry, but I just wanted something quick that works.
Should really be one pass to check that, ignoring any tags, there is only: one embellished operator, and zero or more space-like elements (see https://www.w3.org/TR/mathml-core/#dfn-embellished-operator).

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+                None
+            }
+        })
+    })? {
+        ctx.push(match node {
+            // TODO: add math-shift: compact
+            HtmlNode::Element(elem) => elem
+                .with_optional_attr(attr::scriptlevel, scriptlevel.clone())
+                .with_optional_attr(attr::displaystyle, displaystyle)
+                .into(),
```
mkorje: To do this we need to emit inline styles: `style="math-shift: compact"`.

mkorje: Or maybe we can try to be smarter about it and check if a `ScriptsItem` actually exists somewhere within.

crates/typst-html/src/tag.rs
```diff
@@ -151,6 +151,49 @@ pub fn is_escapable_raw(tag: HtmlTag) -> bool {
     matches!(tag, self::textarea | self::title)
 }
 
+/// Whether this is a foreign tag whose associated element is from the MathML
+/// namespace or the SVG namespace.
```
laurmaedje: Since all of these are MathML, this seems a bit wrong. Is what's meant that SVG tags _could_ also go in here?

crates/typst-html/src/encode.rs
```diff
@@ -338,8 +340,12 @@ fn allows_pretty_inside(tag: HtmlTag) -> bool {
...
 /// In contrast to `allows_pretty_inside`, which is purely spec-driven, this is
 /// more subjective and depends on preference.
-fn wants_pretty_around(tag: HtmlTag) -> bool {
-    allows_pretty_inside(tag) || tag::is_metadata_content(tag) || tag == tag::pre
+fn wants_pretty_around(element: &HtmlElement) -> bool {
+    (allows_pretty_inside(element.tag)
+        && (element.tag != tag::mathml::math
+            || element.attrs.get(attr::mathml::display).is_some_and(|v| v == "block")))
+        || tag::is_metadata_content(element.tag)
+        || element.tag == tag::pre
```
laurmaedje: Something like this might be a little more readable:
```rs
match element.tag {
    tag::mathml::math => {
        element.attrs.get(attr::mathml::display).is_some_and(|v| v == "block")
    }
    tag::pre => true,
    t if tag::is_metadata_content(t) => true,
    t => allows_pretty_inside(t),
}
````

File: crates/typst-html/src/encode.rs [hunk 15]
```diff
@@ -162,19 +162,20 @@ fn write_children(w: &mut Writer, element: &HtmlElement) -> SourceResult<()> {
...
     let pretty_inside = allows_pretty_inside(element.tag)
         && element.children.iter().any(|node| match node {
-            HtmlNode::Element(child) => wants_pretty_around(child.tag),
+            HtmlNode::Element(child) => wants_pretty_around(child),
             HtmlNode::Frame(_) => true,
             _ => false,
         });
 
     w.pretty &= pretty_inside;
+    w.pretty |= element.tag == tag::mathml::math;
```
laurmaedje: It's a bit unfortunate that this is hard-coded, but I'm also not sure I like the resulting formatting where inline math is forcibly multi-line. E.g. here:
```typ
The value of $x$ is $1$.
```
```html
<p>The value of <math>
        <mi>𝑥</mi>
      </math> is <math>
        <mn>1</mn>
      </math>.</p>
```

I think, I'd just remove this. For longer inline maths, the lines will get long, but the same may happen for other inline-level content.

crates/typst-html/src/encode.rs
```diff
@@ -329,6 +330,7 @@ impl RawMode {
 /// should also respect the `style` tag in the future.
 fn allows_pretty_inside(tag: HtmlTag) -> bool {
     (tag::is_block_by_default(tag) && tag != tag::pre)
+        || tag::is_foreign(tag)
```
laurmaedje: I wonder whether this is actually safe? For example, `<mtext>` probably (?) doesn't really want extraneous spacing inside, in practice it's just that it usually won't have children with `wants_pretty_around`. but ideally, the function would only return true for elements that truly don't care about additional spaces and newlines.

Kreijstal:
interested, are mantainers interested in pulling?

isuffix:
Laurenz updated to the `waiting-on-review` label last week. Maintainers are interested and aware of this. If you want to follow more closely, extra discussion usually happens in Discord under the Forge channels.

crates/typst-html/src/lib.rs
```diff
@@ -146,3 +149,9 @@ pub struct FrameElem {
     #[required]
     pub body: Content,
 }
+
+/// Checks whether the given element is an inline EquationElem.
+fn is_inline_equation(elem: &Content) -> bool {
+    elem.to_packed::<EquationElem>()
+        .is_some_and(|elem| elem.block.as_option().is_some_and(|block| !block))
```
laurmaedje: This `.as_option` is not fully safe as it does not respect `set math.equation(block: ..)`. We can make this work correctly by passing the style chain through all the way until here.

crates/typst-html/src/mathml.rs
```diff
@@ -1,13 +1,614 @@
...
+}
+
+trait HtmlNodesExt {
+    fn into_node(self) -> HtmlNode;
+}
+
+impl HtmlNodesExt for EcoVec<HtmlNode> {
+    fn into_node(mut self) -> HtmlNode {
+        if self.len() == 1 {
+            self.pop().unwrap()
```
laurmaedje: `self.into_iter().next().unwrap()` would be a preferable for `EcoVec` because it does not force a reallocation if the vector is shared. for normal `Vec`, I'm not sure whether it would make a difference, but even though it's longer, I still see the into_iter variant more often.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+    let css = if target != ctx.css { target } else { ctx.css };
+    if let Some(node) = ctx.with_css(css, |ctx| -> SourceResult<Option<HtmlNode>> {
+        Ok(match &comp.kind {
+            MathKind::Glyph(item) => {
+                Some(handle_glyph(item, ctx, props, position)?.into())
+            }
+            MathKind::Radical(item) => Some(handle_radical(item, ctx, props)?.into()),
+            MathKind::Accent(item) => Some(handle_accent(item, ctx, props)?.into()),
+            MathKind::Scripts(item) => Some(handle_scripts(item, ctx, props)?),
+            MathKind::Primes(_) => None,
```
laurmaedje: this is still missing and I couldn't find a comment regarding it. only relevant for >4 primes.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+        }
+        MathClass::Relation => {}
+        MathClass::Diacritic => form = Some(Form::Postfix),
+        MathClass::Binary => form = Some(Form::Infix),
+        MathClass::Vary | MathClass::Unary => form = Some(Form::Prefix),
+        MathClass::Punctuation => separator = true,
+        MathClass::Fence => fence = true,
+        MathClass::Large => {
+            largeop = true;
+            form = Some(Form::Prefix);
```
laurmaedje, line 787: why is the form forced to prefix? wouldn't e.g. infix be sensible in cases like this:
```typ
$ a plus.o.big x $
```

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+    lspace: Option<Em>,
+    rspace: Option<Em>,
+    position: NodePosition,
+    limits: bool,
+}
+
+#[derive(Copy, Clone, PartialEq)]
+struct CssContext {
+    /// True if math-style: normal, false if math-style: compact
+    math_style_normal: bool,
```
laurmaedje: I think an enum would be clearer for this, same for shift. I'd probably have both `MathShift` and `MathStyle` (even if they have the same variants) for clarity.

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+            ctx.push(HtmlNode::Tag(tag.clone()));
+            return Ok(());
+        }
+    };
+
+    let props = &comp.props;
+    let embellished = is_embellished_operator(item);
+
+    let target = CssContext::from_math_size(props.size);
+    let scriptlevel = (target.math_depth != ctx.css.math_depth.min(2))
```
laurmaedje, line 528: would be good to have a comment for the `min(2)` logic

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+
+/* Equations */
+mtable.multiline-equation mtd {
+  math-depth: inherit;
+  math-style: inherit;
+  math-shift: inherit;
+  padding: 0;
+}
+mtable.multiline-equation.aligned-equation mtd:nth-child(odd) {
+  text-align: -webkit-right;
```
laurmaedje: why (only) with vendor prefix?

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+    _props: &MathProperties,
+) -> SourceResult<HtmlElement> {
+    let (tag, attr) = if item.position == Position::Below {
+        (tag::munder, attr::accentunder)
+    } else {
+        (tag::mover, attr::accent)
+    };
+
+    let base =
+        ctx.with_css(ctx.css.shift_compact(), |ctx| ctx.handle_into_node(&item.base))?;
```
laurmaedje: for the `with_css` calls it would be nice to document which CSS rules each emulates. IIUC some of these relate to CSS provided by Typst and some to UA styles?

crates/typst-html/src/mathml.rs
```diff
@@ -0,0 +1,896 @@
...
+/* Other rules for scriptlevel, displaystyle and math-shift */
+mroot {
+  math-shift: inherit;
+}
+mroot > :first-child,
+munder > :nth-child(2),
+munderover > :nth-child(2),
+munder[accentunder=\"true\" i] > :first-child {
+  math-shift: compact
+}
```
laurmaedje: Is the `mover[accent=\"true\" i]` case missing here because it's already in the UA style? I don't really understand why that one is in the UA style, but not this one here... but oh well. Also worth a comment, though perhaps outside of the string.

crates/typst-layout/src/math/run.rs
```diff
@@ -70,6 +68,33 @@ pub fn layout_multiline(
     ))
 }
 
+/// Layout an [`AlignedRow`].
+pub fn layout_aligned_row(
+    row: &AlignedRow,
+    ctx: &mut MathContext,
+    styles: StyleChain,
+) -> SourceResult<Vec<MathRun>> {
+    let items: Vec<&MathItem> = row.iter().collect();
```
laurmaedje: This is a bit wasteful as the `AlignedRow` already holds a ready-to-use vector.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -393,6 +393,9 @@ pub struct MathProperties {
     pub lspace: Option<Em>,
     /// The amount of spacing to the right of this item.
     pub rspace: Option<Em>,
+    /// Whether this item is at the start of a left-aligned column and
+    /// semantically infix.
```
laurmaedje: perhaps "_but_ semantically infix" would make this a bit clearer

crates/typst-html/src/mathml.rs
```diff
@@ -380,12 +380,16 @@ fn handle_glyph(
...
 
-    let lspace = (props.lspace.unwrap_or_default().get() != info.lspace
-        && !matches!(position, NodePosition::Only(_)))
-    .then(|| eco_format!("{}em", props.lspace.unwrap_or_default().get()));
-    let rspace = (props.rspace.unwrap_or_default().get() != info.rspace
-        && !matches!(position, NodePosition::Only(_)))
-    .then(|| eco_format!("{}em", props.rspace.unwrap_or_default().get()));
+    let lspace = props
+        .lspace
+        .filter(|l| l.get() != info.lspace)
```
laurmaedje: removing this makes quite the difference. is it really just that what Typst resolves and what the MathML Core Operator dictionary prescribes is almost always the same because both are based on TeX?

crates/typst-library/src/math/ir/process.rs
```diff
@@ -280,6 +281,9 @@ fn spacing<'a>(
 
     let script = |f: &MathItem| f.size().is_some_and(|s| s <= MathSize::Script);
 
+    l.set_rspace(Some(Em::zero()));
+    r.set_lspace(Some(Em::zero()));
```
laurmaedje: in which cases is the field still left as `None`?

crates/typst-library/src/math/ir/item.rs
```diff
@@ -1130,6 +1158,10 @@ pub struct StretchInfo {
     /// A buffer to store the latest stretch added, in case it needs to be
     /// relative to something else.
     buffer: Option<Rel<Abs>>,
+    /// Whether this stretch should be represented semantically.
```
laurmaedje: What does it mean if a stretch is non-semantic?

crates/typst-html/src/mathml.rs
laurmaedje: @mkorje any thoughts on this proposal?

crates/typst-html/src/mathml.rs
laurmaedje: If we can, I think we should make it work. I don't think this is necessarily in conflict with the overall semantics focus.

crates/typst-html/src/mathml.rs
laurmaedje: Worth leaving a comment about in the source then.

crates/typst-html/src/mathml.rs
laurmaedje: If we can easily convert to the non-combining one, it makes the validator happy, and the rendering is good, too, I think we should do it. Any downsides?

crates/typst-html/src/mathml.rs
laurmaedje: Hard-coding the space width as `(4/18)em` seems fine to me. It's a bit unfortunate that this will be emitted at full precision, but that's the fault of `css::length` not rounding and outside of the scope of this PR.

crates/typst-html/src/mathml.rs
laurmaedje: - Warning for ignored math items would definitely be good. I would also suggest converting them to the closest thing alongside the warning, e.g. for `cancel`, at least drawing the not cancelled thing.
- Enforcing phrasing content can be delayed to later. There is a separate task for this in the NLNet grant and we don't enforce content models anywhere else either yet.

crates/typst-html/src/mathml.rs
laurmaedje: If you need any input, let me know.

crates/typst-html/src/mathml.rs
laurmaedje: Are there any downsides to using the CSS?

crates/typst-html/src/mathml.rs
laurmaedje: Once you've double checked, this could also go into an in-source comment even if there isn't any code related to it. The fact that there is _no_ code is just as important in a way.

crates/typst-html/src/mathml.rs
laurmaedje: No worries

crates/typst-html/src/mathml.rs
laurmaedje: Why _inline_ styles specifically. Can we use the style sheet? What would be an example where this makes a difference?

laurmaedje (commented):
First of all: Thank you for this, it's excellent work! I've played around with it, went through the Firefox test suite in all major browsers, read a bit into the spec, and followed along with your code and commits. I can't say I checked every single edge case condition, but I think I have a good idea about how everything fits together. It's quite clean and seems to work very well overall. :)

### A few rendering issues I noticed

Just a few things I noticed while going through the test cases. Maybe you already know these.

- Cases are center-aligned instead of left aligned. The `columnalign` attribute used in the Firefox tests is non-standard, but since `mtd` is just `display: table-cell`, normal CSS `text-align` seems to work from a quick test. Worth checking whether that would be a solution.
- There's an extra space in `$ dif x $` only in Firefox. No idea why.
- Something weird is going on for `$ abs(phi) $` in Chrome. Marking the `<mo>` elements as `stretchy="false"` seems to fix the issue. That's what MathJax does. Writing `stretchy="false"` on parens also improves the appearance of `f(x)` in Chrome.

### A general note

On a general note, I think some more comments would be nice. In particular two kinds:

- On things like `CssContext`, `EmbelishmentContext`, and `NodePosition` just high-level overviews over their purpose. For example, for `CssContext`, I first thought it was related to the CSS that's emitted in the `<style>` tag, but IIUC it's basically simulating how specific properties propagate and whether it is necessary to write them. Just putting things like that in a comment would be very helpful.

- On logic that decides when a particular attribute is emitted. This is often quite subtle as it depends on an interplay of MathML core rules, CSS rules/simulation, Typst & UA stylesheets, etc. As an example of what would've been helpful to me:
  ```rs
  // When top/bottom attachment was requested, but we're in the compact style
  // and the operator would allow under/over to be moved to script position,
  // prevent it.
  let movablelimits = ..
  ```

Figuring out how things play together would've been a little easier with some more explanations. I will gladly admit that I'm not the best at doing this myself, but since you've already done the work of figuring out when exactly it makes sense to write each attribute, I think it's worthwhile to keep that knowledge, close to the code, for posterity and to help other people work on it as well.

### One architectural choice I want to bring up

Currently, the MathML export directly operates on `HtmlElement`s. We could instead make it produce `HtmlElem`s, which would themselves be realized. The main thing this would mean is that the HTML elements produced by MathML creation are visible to realization, e.g. also to `show html.elem`, which would be consistent with how all other built-in elements are translated.

It also plays a bit into how `html.elem` interacts with MathML (currently, it will always give `elem was ignored during MathML export`), though it should be possible to support in both scenarios. While MathML puts a lot of restrictions of what may go in there, in theory it would be nice if packages could emit something like `html.elem("mmultiscripts", ..)` or similar.

If we went the route of using `HtmlElem` instead, we could also move equation handling into a show rule instead of having it hard-coded in conversion. One issue you'd run into is that you can't call `converter.locator.next(&elem.span())`, but this would be safe to replace with `Locator::synthesize(elem.location().unwrap())`.

One other consequence would be that equations would already be realized into HTML elements by the time paragraph collection runs, so the paragraph grouping would have to detect math elements without `display="block"` instead of equations.

Overall, this would feel a little more consistent to me than the current approach, but it might create additional problems. In particular, with the current paragraph handling in HTML, it might cause issues with extra `<p>` elements appearing, but I plan to fix that soon. It would also negatively impact performance but I'm not too concerned with that at the moment. 

We don't have to make this change necessarily, but it's worth discussing.

crates/typst-html/src/mathml.rs
mkorje: I suppose one downside would be that if you stripped the `style` tag from the head of the produced HTML, upright characters would get auto-italicized and one may consider this a change in semantics (in comparison to the other style stuff which is purely visual/alignment related).

crates/typst-html/src/tag.rs
mkorje: Yes. Given the function is checking for "foreign tags", I figured the comment should mention the full scope (though I see how this is confusing).

crates/typst-html/src/mathml.rs
mkorje: Woops I missed that, will fix.

crates/typst-html/src/mathml.rs
mkorje: This seemed to be the only thing that actually worked across all browsers. Here's my understanding of the options:
```css
  /* text-align: right; */ /* doesn't actually work */
  text-align: -webkit-right; /* this hack seems to work across all browsers */
  /* For the future: https://github.com/w3c/mathml-core/issues/156 */
  /* justify-items: end; */
```

crates/typst-html/src/mathml.rs
mkorje: AFAIK writing something like that is "incorrect". The `.big` modifies have Large math class as opposed to binary, and are all called `N-ARY ... OPERATOR ...` in Unicode, and so should only be used when they _are_ prefix. e.g.
```typ
$ plus.o.big_i x_i $
```
Basically, Large math class is never binary. If you wanted the bigger character you'd need to set the class to binary yourself.

crates/typst-html/src/mathml.rs
mkorje: Yes. Though I suppose what we currently do is the odd one out. Both the UA stylesheet and TeX do not apply a cramped style for bottom accents, but we do...

mkorje: I could open a separate PR changing this?

mkorje: There's also one other place where we differ. For `math.root`, we only apply cramped styles to the base, not the index. Whereas for MathML Core cramped styles are applied to both. I suppose we could also switch our behaviour to match that of UA stylesheet.
Making this change doesn't really matter, since I doubt someone has ever written a root with an index that is a superscript...

crates/typst-html/src/mathml.rs
mkorje: Yes, we need to emulate the UA styles with our own style changes.

crates/typst-html/src/mathml.rs
mkorje: Yep.

crates/typst-html/src/mathml.rs
mkorje: This comment is basically for when a user uses one of the size functions `display()`, `sscript()`, etc. with `cramped` not set to the default. So we need to emit things explicitly. There's no way to set `math-shift: compact` other than with CSS (unlike the `math-depth` and `math-style` which can be modified with the `displaystyle` and `scriptlevel` attributes).

crates/typst-library/src/math/ir/item.rs
mkorje: Basically the issue I was having is that the IR treats all stretching equally. So a stretch set for a large op in display is the same as an explicit user's call to stretch. But for MathML we need to differentiate the two, since the large op stretching is done automatically.

So a "semantic" stretch is really an explicit one, whereas a "non-semantic" stretch is one which is done automatically for display purposes. (As I'm writing this, I realize that "explicit" is probably a better name. Though that would mean `stretch(->)^"blahhhh"` is "explicit", which might be a bit confusing since the size is not explicitly set.)

crates/typst-html/src/mathml.rs
mkorje: One thing is adding some vertical spacing between rows in multiline equations. `border-spacing: 0 0.5em;` isn't great since it adds spacing before and after the equation, but also adding padding to each cell feels kinda bad.

crates/typst-html/src/mathml.rs
mkorje: We discussed this a bit a while ago. The current situation, whilst it can cause slightly worse behaviour for the `sum` example (I think both look equally bad though, regardless), is at least always correct for annotations. Trying to do something smart here seems quite tricky and would be fragile. Maybe we open an issue and see if it's possible to improve it later?

crates/typst-html/src/mathml.rs
mkorje: I believe the issue is with converting to a non-combining one. The spec has this list: https://www.w3.org/TR/mathml-core/#combining-character-equivalences, but it doesn't cover everything (for example, the top and bottom things from `under/over`). Also, from https://mathfonts.github.io/math-accents, the non-combining ones look incorrect sometimes.

crates/typst-html/src/lib.rs
mkorje: I had to go all the way back to realization, so I'm not 100% sure what I've done is all correct. Please double check.

crates/typst-html/src/mathml.rs
mkorje: Fixed, but it is a little ugly emitting an `mrow` of `mi`. A single `mi` would probably be nicer, but we don't know what the single prime item will end up as after realization. (Note that either of these two HTML code outputs yield identical results on chromium and firefox it seems.)

On another note, it's not clear whether primes should use an `mo` or a `mi`... It also appears that on chromium and firefox they are visually identical. Primes have normal math class.

crates/typst-html/src/mathml.rs
mkorje: Only issue I see is that it would become quite noisy to have a `font-feature-settings` inline css on _every_ single top accent.

crates/typst-html/src/mathml.rs
mkorje: Done for the ignored math items.

For the arbitrary content, do you want to just ignore everything and warn, or just allow everything?

crates/typst-html/src/encode.rs
mkorje: I agree it probably isn't safe. I would guess that everything that isn't a token element is safe to format then? Token elements are `mtext`, `mi`, `mn`, `mo`, `mspace`, `ms`.

crates/typst-library/src/math/ir/process.rs
mkorje: I realized that this change is actually not needed because the way I was emitting `lspace`/`rspace` was slightly wrong. It is gone now.

crates/typst-html/src/mathml.rs
mkorje: See also https://github.com/w3c/mathml-core/issues/311

crates/typst-html/src/tag.rs
laurmaedje: How about acknowledging this explicitly:
```rs
/// Whether this is a foreign tag whose associated element is from another
/// namespace.
///
/// Currently includes MathML tags, but could be extended to SVG.
```

crates/typst-html/src/encode.rs
laurmaedje: Sounds sensible

crates/typst-html/src/lib.rs
laurmaedje: Looks correct

crates/typst-html/src/mathml.rs
laurmaedje: > but we don't know what the single prime item will end up as after realization

That's only because the IR forgets it though, isn't it? In principle, it's just a number and we could emit whatever we want there.

crates/typst-html/src/mathml.rs
laurmaedje: hmm okay I see

crates/typst-html/src/mathml.rs
```diff
@@ -357,7 +358,21 @@ fn handle_realized(
             MathKind::Multiline(item) => Some(handle_multiline(item, ctx, props)?.into()),
 
             // Polyfill required for MathML Core.
-            MathKind::SkewedFraction(_) | MathKind::Cancel(_) | MathKind::Line(_) => None,
+            MathKind::SkewedFraction(item) => Some(ignored_math_item(
+                ctx,
+                &item.numerator,
+                props.span,
+                "math.frac.where(style: \"skewed\")",
```
laurmaedje: if we use code syntax it should be in backticks in diagnostics. even for `math.overline`,  probably. alternatively, we can use more human-readable names like "skewed fraction was ignored ..."

crates/typst-library/src/math/ir/item.rs
```diff
@@ -623,10 +623,11 @@ impl<'a> SkewedFractionItem<'a> {
         denominator: MathItem<'a>,
         slash: MathItem<'a>,
         styles: StyleChain<'a>,
+        span: Span,
     ) -> MathItem<'a> {
         let kind =
             MathKind::SkewedFraction(Box::new(Self { numerator, denominator, slash }));
-        let props = MathProperties::default(styles);
+        let props = MathProperties::default(styles).with_span(span);
```
laurmaedje: looking at this, maybe we should require a span in the constructors of `MathProperties` to make the scenario with a missing span opt-in

crates/typst-html/src/mathml.rs
laurmaedje: I'm fine with aligning behavior with TeX and MathML Core on the layout side, too

crates/typst-library/src/math/ir/item.rs
laurmaedje: I agree that `explicit` is better

crates/typst-html/src/mathml.rs
laurmaedje: Would it have to be _inline_ CSS?

crates/typst-html/src/mathml.rs
laurmaedje: Okay, let's stay with the combining one then.

crates/typst-html/src/mathml.rs
laurmaedje: For now, I'd lead towards just allowing everything. Same as the rest of HTML export.

crates/typst-html/src/mathml.rs
laurmaedje: What's the issue with padding?

crates/typst-html/src/mathml.rs
laurmaedje: Let's keep it in the HTML then

crates/typst-html/src/mathml.rs
laurmaedje: Also worth a comment then (outside of the string)

crates/typst-html/src/mathml.rs
mkorje: I suppose nothing really, other than just feeling :p. Though it's hidden in the stylesheet anyway, so this is really irrelevant.

Question is then what should it be set to?

crates/typst-html/src/mathml.rs
mkorje: I opened #8082

crates/typst-html/src/mathml.rs
mkorje: i'll go with the human-readable names then instead.

crates/typst-html/src/mathml.rs
mkorje: oh, ofc. I can add it to the styles, and then either create a class or use inline css to disable it.

crates/typst-html/src/mathml.rs
mkorje: Yes, but we have this behaviour in typst-layout where we pack the primes closer together than they would be if placed one after the other.

crates/typst-html/src/mathml.rs
mkorje: I'll deal with allowing it then with the structural change.

crates/typst-html/src/mathml.rs
```diff
@@ -44,6 +42,9 @@ mtable.multiline-equation mtd {
   math-shift: inherit;
   padding: 0;
 }
+mtable.multiline-equation mtr:not(:last-child) mtd {
+  padding-bottom: 0;
+}
```
mkorje: Need to decide what to put as the value here (instead of 0).

crates/typst-library/src/math/ir/item.rs
```diff
@@ -1148,9 +1148,9 @@ impl Stretch {
     }
 
     /// Whether the stretch along the given axis should be represented
-    /// semantically.
-    pub fn is_semantic(self, axis: Axis) -> bool {
-        self.0.get(axis).is_some_and(|info| info.semantic)
+    /// explicitally.
```
laurmaedje: love the word "explicitally" 😂

mkorje: That's what I get for doing a find and replace on `semantic` :laughing:

crates/typst-html/src/mathml.rs
laurmaedje: I suppose if we're gonna do this, might as well pick a round value that looks close enough to Typst's layout output?

crates/typst-html/src/mathml.rs
mkorje: I opened #8103

crates/typst-html/src/mathml.rs
mkorje: I've gone with 0.5em for now.

crates/typst-html/src/lib.rs
```diff
@@ -150,8 +149,12 @@ pub struct FrameElem {
...
-        .is_some_and(|elem| !elem.block.get(styles))
+/// Checks whether the given HTML element is an inline equation.
+fn is_inline_equation(elem: &HtmlElem) -> bool {
+    elem.tag == tag::mathml::math
+        && elem
+            .attrs
+            .as_option()
+            .as_ref()
+            .is_none_or(|f| f.get(attr::mathml::display).is_none())
 }
```
laurmaedje, line 160: is there a particular reason why you removed the `styles` handling that you introduced earlier? in principle, the `attrs` suffer from the same problem when using `as_option`

crates/typst-html/src/mathml.rs
```diff
@@ -1,13 +1,1256 @@
...
+                HtmlElement::new(tag::mspace)
+                    .with_attr(attr::width, eco_format!("{}", css::length(*amount))),
+            );
+            return Ok(());
+        }
+        MathItem::Space => {
+            // We hard-code the space width as `(4/18)em`.
+            ctx.push(HtmlElement::new(tag::mspace).with_attr(
+                attr::width,
+                eco_format!("{}", css::length(Em::new(0.2222222222222222))),
```
laurmaedje, line 443: let's maybe replace the constant with `4.0 / 18.0`

crates/typst-html/src/mathml.rs
```diff
@@ -1,13 +1,1256 @@
...
+    // We clamp the tracked depth at 2 so we don't emit unnecessary
+    // `scriptlevel="2"` everywhere.
+    let scriptlevel = (target.math_depth != ctx.css.math_depth.min(2))
+        .then(|| eco_format!("{}", target.math_depth));
+    let displaystyle = (target.math_style != ctx.css.math_style)
+        .then(|| eco_format!("{}", target.math_style == MathStyle::Normal));
+    let mut properties = css::Properties::new();
+    if target.math_shift != ctx.css.math_shift {
+        let value =
+            if target.math_shift == MathShift::Normal { "normal" } else { "compact" };
```
laurmaedje, line 466: for such cases, an exhaustive match would be nicer imo. could also do with a method for the to string conversion

crates/typst-html/src/mathml.rs
```diff
@@ -1,13 +1,1256 @@
...
+fn handle_primes(
+    item: &PrimesItem,
+    ctx: &mut MathContext,
+    props: &MathProperties,
+    position: NodePosition,
+) -> SourceResult<HtmlElement> {
+    // Try to collapse the resolved item of primes into a single `mo`.
+    if let MathItem::Component(comp) = &item.primes
+        && let MathKind::Group(group) = &comp.kind
+        && let Some(text) = primes_text(&group.items)
```
laurmaedje, line 933: why do we resolve the primes first (into a special HTML only field) and then unpack it again instead of just directly generating the correct prime text here, skipping symbol elem?

crates/typst-html/src/lib.rs
mkorje: No real reason. I think it might have been to match what the phrasing content check does? Regardless, I can go back.

laurmaedje (commented):
The PR is in a great state now I think! Basically mergeable except for another rebase with the test references + the unresolved infrastructural thing. Fixing the paragraph stuff in HTML is close to top of my todo list now, so I hope to have that resolved soon. Then, we can see whether the infrastructural change is fine with that change.

crates/typst-html/src/mathml.rs
```diff
@@ -1,13 +1,1256 @@
...
+pub(crate) const MULTILINE_EQUATION_CLASS: &str = "multiline-equation";
+pub(crate) const ALIGNED_CLASS: &str = "aligned";
+pub(crate) const LEFT_ALIGN_CLASS: &str = "left-align";
+pub(crate) const RIGHT_ALIGN_CLASS: &str = "right-align";
+pub(crate) const CASES_CLASS: &str = "cases";
+pub(crate) const FLUSHED_CLASS: &str = "flushed";
+pub(crate) const LEFT_FLUSH_CLASS: &str = "left-flush";
+pub(crate) const RIGHT_FLUSH_CLASS: &str = "right-flush";
+
+/// How Typst overrides the [MathML Core User Agent Stylesheet][UA].
```
laurmaedje, line 30: I really appreciate the detailed comment here.

crates/typst-html/src/mathml.rs
mkorje: The issue is we have no idea what happens to the primes after realization. Creating a special item for HTML export that includes all the primes means we can at least try to get the nice structure of a single `mo` in mathml export.

I don't really like this at all, but idk how else to do it.

mkorje: Oh wait I think I get what you mean. Remove symbolelem/glyph item middle man entirely and just store a string in primes directly like glyphitem? That makes a lot more sense ahah.

crates/typst-html/src/mathml.rs
mkorje: I'm glad! I thought it might also be cleaner to use some string formatting so that the class names aren't repeated, and are 'controlled' directly by the const class names. But that needs some const formatting crate added...

laurmaedje: It could also be a `EcoString` in a lazy lock. in the end, it becomes an eco string anyway, so in some sense this would even be more efficient as a single eco string could be shared.

crates/typst-html/src/mathml.rs
mkorje: I went with implement the display trait for a `to_string` method.

crates/typst-html/src/lib.rs
mkorje: I've gone back to what I had before with styles.

crates/typst-html/src/mathml.rs
mkorje: I went ahead and did that. It's in a separate commit which I'll squash later.

crates/typst-html/src/mathml.rs
mkorje: Fixed. I also made it so that typst-layout also uses this for consistency.

laurmaedje:
On the questions you DMed:

> box within math causes nested paragraphs. e.g. in `mathml-ignored-external-content` we get a `<p><span style="display: inline-block">b</span></p>` appearing.

This is caused by the `grid` in this case. It is allowed to mix inline and blocky HTML elements in a realization, but block-level Typst elements will force paragraphs.

The same problem would exist in layout, but there we use math realization, which does not include the paragraph grouping rules. Perhaps we want to do the same within MathML elements? Adding to the existing logic that selects between `html_block_fragment` and `html_inline_fragment` based on the tag. This might work, but I'm not sure.

> With some tests that have show rules on math stuff, I'm getting weird nested math tags appearing. e.g. in math-attach-show-limit `<mo lspace="0em">∫</mo>` becoming `<mo lspace="0em"><math><mtext>∫</mtext></math></mo>`. I believe this might have something to do with the MathML HTML elements going through HTML fragment realization.

I suspect this is because mathy elements (like `math.attach`) are wrapped in equations during realization (see `content.can::<dyn Mathy>()`). This is so that a lone `math.attach` outside of math mode results in an equation.

You can cause similar problems with anything that triggers a nested realization, e.g. something like this:
```typ
#show math.frac: it => html.elem("mrow", it)
$ a/b $
```

In layout, this is avoided by the code following `Transparently recurse into equations nested in math` during math realization. However, in HTML we're currently not in a math realization, so this is not happening. Potentially, this would thus also be solved by the same thing I proposed for the previous point.

There is a bit of a divergence between layout and HTML here because the math layout code drives realization and picks math realization while HTML export (as I suggested) uses a show rule and thus gets normal realization. I did not anticipate this consequence of my suggested design. My mind is not yet entirely cleared up on what is conceptually _right,_ but what I suggested above is definitely worth a shot and I'll try to think some more about this.

> I'm not quite sure how the new Display enum thing works exactly? It seems to leave out that you can have `<display-outside> || <display-inside>` values. [...] The reason this comes up is that the MathML spec extends the CSS with the values block math and inline math.

Since it is supposed to be able to express every variantion of `display`, additional variants for block and inline math should be fine.

> Now the CSS spec basically has for every combination a short display form, but one example that doesn't is `block ruby`.

That's a good catch. I expressed this as `<div><ruby>..</ruby></div>` because `display: block` ruins the ruby, but I didn't realize that `display: block ruby` is a possibility. We could switch to that.

> How should we handle `#box($ a $)`? With the current infrastructure you added it'd be natural to make that emit a plain math tag without `display="true"`, but that isn't quite right, since we want the display version.

> Also, there's added difficulty here since you can use (and we do use) the display attribute on the math tag, which sets the css display property. You can see what I ended up doing [in this commit](https://github.com/typst/typst/pull/7436/changes/8bcb4b363fa985a1406efa92a06b184d46002aea), which does seem to work, but doesn't seem like what you'd have intended/expected.

I have this comment in `convert.rs`:
```rs
/// Paragraphs only allow "Phrasing content". The default display properties
/// that exists for valid phrasing content are `none`, `inline`, `inline-block`,
/// `contents`, and `ruby`. These can all be left as-is.
```

However, `<math>` would be an exception here. It is valid phrasing content and has a display mode that cannot be left as-is.

I think it would be preferrable if `box(..)` and `block(..)` would only touch the CSS `display` property and not the `<math>` `display` attribute. It's also important that `box` can revert any change `block` makes and vice versa.

If possible, I'd make it such that the equation itself determines the value of the `display` attribute and `box/block` can then adjust the CSS `display` property (which should override the UA style sheet for `math[display=block]`?)

I tested and to me it seems that for `box($ ... $)`

- `<math display="block" style="display: inline math">` is the cleanest way to express what we want. It works in Chrome, but not in Firefox and Safari because they do not support `display: math` at all.

- `<math display="block" style="display: inline">` works in Firefox and Safari, but not in Chrome because it then uses inline layout instead of math layout.

- `<math display="block" style="display: inline; display: inline math;">` works in all three. (Using the typical polyfill mechanism of specifying the property twice with the unsupported new thing last.). The current CSS properties struct can't express duplicate properties, but if we think this makes sense, we could potentially make it work.

Any other ideas?

> And on the Typst doc regarding setting the row/col spacing of math/vec/cases to match Typst defaults, I'll just copy here my reply:

> > The only downside I suppose is that it would be quite complex, since we would need to have the padding be conditional for the outer edges. So this would add a lot more CSS. That's why I was a bit apprehensive.

I don't have a strong opinion on this. I'm fine with leaving it out as well. Your call.

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -189,6 +191,8 @@ fn resolve_realized<'a, 'v, 'e>(
     } else if let Some(elem) = elem.to_packed::<BoxElem>() {
         let locator = ctx.locator.next(&elem.span());
         ctx.push(BoxItem::create(elem, styles, locator));
+    } else if let Some(body) = (ctx.engine.routines.html_mathml_body)(elem, styles) {
```
laurmaedje: I would suggest moving this to the bottom, to not pay for the indirect call needlessly in the middle.

crates/typst-html/src/mathml.rs
```diff
@@ -1048,6 +1050,31 @@ fn handle_fenced(
...
+        .unwrap()
+        .as_ref()
+        .clone()
+        .with_body(body)
+        .pack()
+        .spanned(item.elem.span());
+    if let Some(location) = item.elem.location() {
+        content.set_location(location);
+    }
+    Ok(content)
```
laurmaedje: I would just write it like this:
```rs
let mut elem = item.elem.to_packed::<HtmlElem>().unwrap().clone();
elem.body.set(body);
Ok(elem.pack())
````
There shouldn't be a need to set the location or span since they should be retained when cloning.

Comment (mkorje) [reply crates/typst-html/src/mathml.rs]: I did not realise you could do `elem.body.set(body)`!

File: crates/typst-html/src/convert.rs [hunk 42]
```diff
@@ -208,6 +208,15 @@ fn handle_html_elem(
             // create inline-level content next to block-level content
             // without a paragraph automatically appearing.
             *converter.quoter = SmartQuoter::new();
+        } else if tag::is_foreign(elem.tag) {
```
laurmaedje: Let's use a more math specific check here. e.g. `tag::is_mathml`

laurmaedje:
Excellent work! I'm excited to see this merged. 🎉

PR #7443, opened 2025-11-23
Add colors to `--help/-h`

starsep:
User can disable colors via `NO_COLOR=1` environment variable if needed

Output of `typst -h`:
| before | after | `NO_COLOR=1` |
| - | - | - |
| typst-before | typst-colors | typst-no-color |

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -43,10 +43,8 @@ const AFTER_HELP: &str = color_print::cstr!("\
 ");
 
 const STYLES: Styles = Styles::styled()
-    .header(AnsiColor::Yellow.on_default().effects(Effects::BOLD))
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks! We like this in principle, but prefer different colors and styles. I've updated the PR to that effect.

image

alerque:
Isn't there a way to get a preconfigured set of styles that will match whatever the system is already using for man pages? Having less arbitrary color preferences and more matching terminal themes and how they already render similar lexical items seems like a good thing to me. I don't know what the incantation is off the top of my head but I thought I saw a project with that setup somewhere.

laurmaedje:
Honestly, I doubt that many people will configure how their man pages will look, so for many people this would just default to the preferences of whoever set _that_ default. I also think a distinct look gives a bit of character to the software. The blue-ish colors chosen here are pretty close to Typst's brand colors and I think that's quite nice.

PR #7446, opened 2025-11-24
Improve how `counter.display` auto-selects the numbering

mkorje:
From #6781


crates/typst-library/src/introspection/counter.rs
```diff
@@ -260,22 +261,56 @@ impl Counter {
...
-                } else if func == FootnoteElem::ELEM {
-                    Some(styles.get_cloned(FootnoteElem::numbering))
-                } else {
-                    None
-                }
-            }
+            CounterKey::Page => loc.page_numbering(engine, span),
+            CounterKey::Selector(Selector::Elem(func, _)) => engine
+                .introspect(QueryUniqueIntrospection(Selector::Location(loc), span))
+                .map_or_else(
```
mkorje: ~~Ignoring the error here is probably a problem, I think? But I need to deal with the case where no specific piece of content could be found at the location...~~

I used `QueryFirstIntrospection` instead since it returns `Option<Content>`

crates/typst-library/src/introspection/counter.rs
```diff
@@ -260,22 +261,56 @@ impl Counter {
...
+                        } else {
+                            None
+                        }
+                    },
+                    |content| {
+                        if func == HeadingElem::ELEM {
+                            content
+                                .to_packed::<HeadingElem>()
+                                .and_then(|elem| elem.numbering.as_option().clone())
+                                .flatten()
```
mkorje: I needed to do the initial comparison as I only want to get the numbering field if the content found at the location is actually the same as the target func

laurmaedje: I see!

laurmaedje:
Thanks!

PR #7448, opened 2025-11-24
Fix crash in PDF tagging with oob grid line end

### LINKED ISSUE #7398: Compiler crash in grid tagging
### Description

This code results in a crash in PDF export:
```typ
#rect(inset: 9pt, stroke: none)[
  #table(
    columns: (1fr, 1fr), align: left, stroke: none, inset: 10pt, gutter: -5pt,
    [column 1], [column 2],
    table.hline(stroke: 0.5pt),
    [
      some text
    ],
    table.vline(stroke: 0.5pt, end: 3),
    [
      some text
    ]
)]
```
```
thread 'main' panicked at crates/typst-pdf/src/tags/context/grid.rs:112:37:
index out of bounds: the len is 4 but the index is 4
```
(Bug report submitted via web app feedback.)

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

saecki:
Fix #7398.

This fixes the crash, but it still allows oob values for the `end`. I wonder if it would make sense to make this a hard error after we make the point release :thinking:


tests/suite/layout/grid/stroke.typ
```diff
@@ -437,3 +437,13 @@
 --- grid-vline-position-top paged ---
 // Error: 24-27 expected `start`, `left`, `right`, or `end`, found top
 #table.vline(position: top)
+
+--- grid-line-end-oob paged ---
```
laurmaedje:
```suggestion
--- issue-7398-grid-line-end-oob paged ---
```

Issue #7301, opened 2025-11-04
Crash when determining link parent in PDF export


aersam:
### Description

We generate lot's of PDF's using typst, but sometimes we randomly get this error:

```
panicked at ...\typst-pdf-0.14.0\src\link.rs:87:55: link parent
```

### Operating system

Windows

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
Would you be able to provide the source code that led to the crash?

Also a note for the future: The issue contains a lot of irrelevant information, which diverts from the point being made. The whole PyO3 thing is unrelated. It's best to keep the issue focused to the problem itself. I edited it now to make it a bit clearer.

aersam:
> Would you be able to provide the source code that led to the crash?
> 
> Also a note for the future: The issue contains a lot of irrelevant information, which diverts from the point being made. The whole PyO3 thing is unrelated. It's best to keep the issue focused to the problem itself. I edited it now to make it a bit clearer.

Ok, I was able to create a pretty minimal repo:

```typst
#link("asf")[#none\ #none]
```

saecki:
The problem is that the space which has the `LinkElem::curent` style applied is extracted from within the `link-marker` tags: https://github.com/typst/typst/blob/bd17979ef1ffa04428f00b69bee7af8b6880384b/crates/typst-realize/src/lib.rs#L834

Pascal-So:
I stumbled across another example of this same panic which I managed to reduce to this snippet:

```typst
#footnote(numbering: it => "", [asdf])
```

Note however that the following code works just fine??

```typst
#place(footnote(numbering: it => "", [asdf]))
```

Not sure if this is the same underlying issue of if it's a different case. Hope it helps. Tested in typst 0.14.0.

saecki:
> I stumbled across another example of this same panic which I managed to reduce to this snippet:
> 
> #footnote(numbering: it => "", [asdf])

This seems to be the same root cause. Since there is no other content inside the link body, the link target is applied to a space element.

laurmaedje:
Here's a spaceless reproduction:
```typ
#link("asdf", linebreak())
```
(Should be saved without any trailing whitespace.)

saecki:
[...]

Issue #7257, opened 2025-10-29
Crash with heading and par rules


knuesel:
### Description

The following code

```typ
#show par: none

#show heading: it => {
  // No problem when the following two lines are exchanged
  block()
  it.body
}

= Section
#lorem(30)
```

gives

```
thread 'main' panicked at crates/typst-pdf/src/tags/tree/build.rs:181:5:
tags weren't properly closed
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: typst_pdf::tags::tree::build::build
             at /home/j/repos/typst/crates/typst-pdf/src/tags/tree/build.rs:181:5
   3: typst_pdf::tags::init
             at /home/j/repos/typst/crates/typst-pdf/src/tags/mod.rs:28:9
   4: typst_pdf::convert::setup
             at /home/j/repos/typst/crates/typst-pdf/src/convert.rs:102:16
   5: typst_pdf::convert::convert
             at /home/j/repos/typst/crates/typst-pdf/src/convert.rs:44:34
   6: typst_pdf::pdf
             at /home/j/repos/typst/crates/typst-pdf/src/lib.rs:32:5
   7: typst::compile::export_pdf
             at /home/j/repos/typst/crates/typst-cli/src/compile.rs:377:18
   8: typst::compile::export_paged
             at /home/j/repos/typst/crates/typst-cli/src/compile.rs:341:13
   9: typst::compile::compile_and_export::{{closure}}
             at /home/j/repos/typst/crates/typst-cli/src/compile.rs:313:53
  10: core::result::Result<T,E>::and_then
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/result.rs:1381:22
  11: typst::compile::compile_and_export
             at /home/j/repos/typst/crates/typst-cli/src/compile.rs:313:26
  12: typst::compile::compile_once
             at /home/j/repos/typst/crates/typst-cli/src/compile.rs:248:43
  13: typst::compile::compile::{{closure}}
             at /home/j/repos/typst/crates/typst-cli/src/compile.rs:43:38
  14: typst::timings::Timer::record
[...]

PR #7450, opened 2025-11-24
Generate internal compiler errors instead of panicking

saecki:
Avoid crashes: #7301 #7257

But proper fixes are still outstanding.

This adds some utilities to generate internal compiler errors  instead of panicking. These errors can capture a stack trace which is nice for debugging. Generating a backtrace is only attempted when `debug_assertions` are enabled and when `RUST_BACKTRACE` is set. In release mode debug symbols are stripped anyway, so the backtrace only displays addresses.
The backtrace isn't captured using `Backtrace::force_capture`, because the paths in the backtrace are absolute and thus machine dependent, I experimented with sanitizing the paths, but that was somewhat hacky since there is no proper API to access the individual frames.

### DIFF
--- a/crates/typst-pdf/src/tags/tree/build.rs
+++ b/crates/typst-pdf/src/tags/tree/build.rs
@@ -19,7 +19,10 @@ use ecow::EcoVec;
 use krilla::tagging::{ArtifactType, ListNumbering, Tag, TagKind};
 use rustc_hash::FxHashMap;
 use smallvec::SmallVec;
-use typst_library::diag::{SourceDiagnostic, SourceResult, bail, error};
+use typst_library::diag::{
+    At, ExpectInternal, SourceDiagnostic, SourceResult, assert_internal, bail, error,
+    panic_internal,
+};
 use typst_library::foundations::{Content, ContextElem};
 use typst_library::introspection::Location;
 use typst_library::layout::{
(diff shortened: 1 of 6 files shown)


laurmaedje (commented):
It would be nice if the error had a precise line/column number regardless of the backtrace like panics do. if the backtrace API can't do that in release, I suppose a macro still could with `file!()` and `line!()`. though that would only nicely work for the `assert` variant, not for the `expect` variant.

saecki:
Using `#[track_caller]` and `std::panic::Location::caller()` works!

crates/typst-library/src/diag.rs
```diff
@@ -890,3 +891,44 @@ pub fn format_xml_like_error(format: &str, error: roxmltree::Error) -> LoadError
 
     LoadError { pos: pos.into(), message }
 }
+
+/// Assert a condition, on failuer generate an internal compiler error with the
```
istudyatuni:
```suggestion
/// Assert a condition, on failure generate an internal compiler error with the
```

laurmaedje:
The Windows tests strike again, c.f. https://github.com/typst/typst/pull/6433#issuecomment-3242140547

crates/typst-pdf/src/tags/tree/build.rs
```diff
@@ -178,18 +180,24 @@ pub fn build(document: &PagedDocument, options: &PdfOptions) -> SourceResult<Tre
...
+        assert_internal(tree.stack.is_empty(), "tags weren't properly closed")
+            .at(tree.groups.get(last.id).span)?;
+    }
+    assert_internal(tree.unfinished_stacks.is_empty(), "tags weren't properly closed")
+        .at(Span::detached())?;
+    assert_internal(
+        Some(&GroupId::ROOT) == tree.progressions.last(),
+        "tags weren't properly closed",
+    )
+    .at(Span::detached())?;
```
laurmaedje, line 194: I only just noticed the very small differences here.

1. The `tree.groups.get(last.id).span` should be fine since even if that somehow panics for a malformed ID, we'd have panicked anyway because the stack can't be empty if `tree.stack.last() is Some(last)`. In fact `assert_internal` could actually just be a new `panic_internal` here?

2. `tree.progressions.first()` became `Some(&GroupId::ROOT)`. Is that 100% always the same? Otherwise, I'd prefer to not touch it for this PR as it seems slightly unrelated and I want to really keep diffs as minimal as possible for PRs that will be backported. If it's just a nice clean-up, it could also easily be changed in a separate PR.

saecki: > I only just noticed the very small differences here.
> 
>    1. The `tree.groups.get(last.id).span` should be fine since even if that somehow panics for a malformed ID, we'd have panicked anyway because the stack can't be empty if `tree.stack.last() is Some(last)`. In fact `assert_internal` could actually just be a new `panic_internal` here?

Yeah I've noticed that but was a bit lazy. It's a good point though.

>    2. `tree.progressions.first()` became `Some(&GroupId::ROOT)`. Is that 100% always the same? Otherwise, I'd prefer to not touch it for this PR as it seems slightly unrelated and I want to really keep diffs as minimal as possible for PRs that will be backported. If it's just a nice clean-up, it could also easily be changed in a separate PR.

It should be 100% the case, and that assertion is also made somewhere else, but let's just keep things the same.

PR #7451, opened 2025-11-24
Fix crashes on array comparison

### LINKED ISSUE #6285: Typst panic on comparison function
### Description

The `sorted` function on arrays sometimes panics with the message
```
thread 'main' panicked at library/core/src/slice/sort/shared/smallsort.rs:865:5:
user-provided comparison function does not correctly implement a total order
```
when a `key` argument is provided. It seems that #5627 resolves this, but for a different purpose, i.e. to allow custom comparison functions in typst. Here the bug is a panic even without such custom comparison. I'm not sufficiently familiar with the codebase so might as well confirm. In any case, using a `key` argument to sort the array should never raise this error because built-in typst values should have correctly implemented total orders.

Somewhat minimized example:
```typst
#let translation-table = state("translations", ())
#let add-entry(zh, en) = translation-table.update(it => {
  it.push((zh, en))
  it
})
#let translate(zh, en) = [#zh #en #add-entry(zh,en)]

#translate[语境][context]
#translate[原始表达式][raw term]
#translate[判值相等][judgmental equality]
#translate[元理论][metatheory]
#translate[归纳递归类型族][indexed inductive-recursive type]
#translate[归纳类型族][indexed inductive type].
#translate[语境性][contextual]
#translate[宇宙][universe]
#translate[累积][cumulation]
#translate[正规形式][normal form]
#translate[中性形式][neutral form]
#translate[更名][renaming]
#translate[典范形式][canonical form]
#translate[语义][semantics]
#translate[解释][interpretation]
#translate[模型][model]
#translate[自洽][consistent]


#translate[我][] // Empty! Seems to panic only when the items are added in specific orders.


#translate[相容][consistent]
#translate[独立][independent]
#translate[反模型][contermodel]

#context {
  let final = translation-table.final().sorted(key : ((zh, en)) => en.text)
  for (zh, en) in final [
    / #zh: #en

  ]
}
```

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

RungeCC:
Close #6285.

This is done by move the `None` branch of `match by` into closure `are_in_order`.

+ The error message change in case `#(1pt, 2pt, 3em).sorted()` may relate to some internal differences between stdsort and glidesort; maybe for some specific error types, we should match their output with regular expressions (instead of a simple comparison), but this is beyond the scope of this PR.
+ <del>The added test is pretty verbose, but because it's related to the internal implementation of stdsort, I'm not sure if it's possible to simplify.</del> Now use reproducer from https://github.com/typst/typst/issues/6285#issuecomment-2887982762.
+ As my comment https://github.com/typst/typst/issues/6285#issuecomment-3523601099 says, this PR focuses on solving the crash, more fine-grained diagnostics when encountering non-total-order comparison function may be considered in the future.

### DIFF
--- a/tests/suite/foundations/array.typ
+++ b/tests/suite/foundations/array.typ
@@ -478,30 +478,8 @@
 #((k: "a", v: 2), (k: "b", v: 1)).sorted(it => it.v)
 
 --- issue-6285-crashed-with-sorting-non-total-order paged ---
-// Error: 2-23:11 cannot compare string and none
-#(
-  "context",
-  "raw term",
-  "judgmental equality",
-  "metatheory",
-  "indexed inductive-recursive type",
-  "indexed inductive type",
-  "contextual",
-  "universe",
-  "cumulation",
-  "normal form",
-  "neutral form",
-  "renaming",
-  "canonical form",
-  "semantics",
-  "interpretation",
-  "model",
-  "consistent",
-  none,
-  "consistent",
-  "independent",
-  "contermodel",
-).sorted()
+// Error: 2-66 cannot compare string and none
+#(("a", "b", "c", "d", "e", "z") * 3 + ("c", none, "a")).sorted()
 
 --- issue-3014-mix-array-dictionary paged ---
 // Error: 8-17 expected expression, found named pair


laurmaedje:
Closing in favor of https://github.com/typst/typst/pull/7520. The approach there is simpler and should be safer to backport (less chance for regressions). It's probably also more performant since it avoids an unnecessary extra allocation for retaining the indices which aren't needed. The change in error message is also avoided, though that wouldn't have been a problem.

Thanks for the PR nonetheless!

PR #7456, opened 2025-11-25
Fix symbols with variation selectors not working as accents (#7437)

### LINKED ISSUE #7437: Update to 14.0 broke arrow.l.r accents
### Description

A change in the representation of `math.arrow.l.r` from 13.1 to 14.0 broke the use of `arrow.l.r` as an accent.

**Version 13.1**
The codepoint representation `#str.codepoints(math.arrow.l.r)` gives `("↔",)` and both `$arrow.l.r(A)$` and `$accent(A, arrow.l.r) $` render correctly.

**Version 14.0**
The codepoint representation `#str.codepoints(math.arrow.l.r)` gives `("↔", "\u{fe0e}")`, `$arrow.l.r(A)$` compiles but renders incorrectly as ↔(A) and `$accent(A, arrow.l.r)$` doesn't compile with error "_Expected exactly one character_".

### Reproduction URL

https://typst.app/project/rfRqM6hO3ILjR9psnRMBja

marrable:
Fixes #7437

### Problem
The codex bump in cbc9ae1ec (#6875) intentionally added variation selectors to certain symbols to distinguish text vs emoji presentation. For example, `arrow.l.r` now contains `("↔", "\u{fe0e}")` instead of just `("↔",)`.

This broke code paths that assumed symbols are single codepoints:
1. `Symbol::func()` in `symbol.rs` - used when calling a symbol as a function (e.g., `$arrow.l.r(A)$`)
2. Accent cast in `accent.rs` - used when passing a symbol to `accent()` (e.g., `$accent(A, arrow.l.r)$`)
3. char cast in `str.rs` - used when converting strings to chars

### Fix
Updated all three locations to accept one cluster (base character + optional variation selectors) and extract only the base character. 

### Rationale
Considered alternatives:
- Strip variation selectors at source (in codex): would break the text/emoji presentation feature that was intentionally added
- New GraphemeCluster type: considered overkill for an urgent regression fix

### Changes
`crates/typst-library/src/foundations/symbol.rs`: Updated Symbol::func()
`crates/typst-library/src/math/accent.rs`: Updated Accent cast
`crates/typst-library/src/foundations/str.rs`: Updated char cast
`tests/suite/math/accent.typ`: Added regression test math-accent-bidirectional-arrow

### New Files
`tests/ref/render/math-accent-bidirectional-arrow.png`: Generated reference

### DIFF
--- a/docs/src/lib.rs
+++ b/docs/src/lib.rs
@@ -912,7 +912,7 @@ fn symbols_model(resolver: &dyn Resolver, group: &GroupData) -> SymbolsModel {
                 }),
                 value: value.into(),
                 // Matches casting `Symbol` to `Accent`
-                accent: value_char
+                accent: typst::foundations::Symbol::strip_text_presentation(value)
                     .is_some_and(|c| typst::math::Accent::combine(c).is_some()),
                 alternates: symbol
                     .variants()


MDLC01:
For the `accent` constructor, I'm wondering whether instead it should consider the whole cluster when choosing the accent, such that for example `emoji.arrow.r` would not work as an accent because of the emoji variation selector.

MDLC01:
For the `accent` constructor, I'm wondering whether instead it should consider the whole cluster when choosing the accent, such that for example `emoji.arrow.r` would not work as an accent because of the emoji variation selector.

laurmaedje:
At the very least, we'd want it to work in the case without the variation selector.

However, for the generic `char` cast I'm not sure we should silently drop the variation selector. Probably, the fix should be scoped to just accents. There might be further places where it's relevant, but generically on `char` seems to broad for me. I'll need to take a closer look.

laurmaedje:
I checked exactly which code paths would be affected by the added char cast:

- The `to-unicode` function on strings:
  https://github.com/typst/typst/blob/fd048b7d71f6f78413764eb650a13fac88c800c7/crates/typst-library/src/foundations/str.rs#L293-L299

- CSV delimiters:
  https://github.com/typst/typst/blob/fd048b7d71f6f78413764eb650a13fac88c800c7/crates/typst-library/src/loading/csv.rs#L135-L143

- Accents:
  https://github.com/typst/typst/blob/fd048b7d71f6f78413764eb650a13fac88c800c7/crates/typst-library/src/math/accent.rs#L240-L248

- Matrix delimiters:
  https://github.com/typst/typst/blob/fd048b7d71f6f78413764eb650a13fac88c800c7/crates/typst-library/src/math/matrix.rs#L277-L283

- The typed HTML accesskey attribute accepts a single-codepoint string or an array thereof: 
  https://github.com/typst/typst-assets/blob/d6a48348174f11bed3ab9f5dccdb6f4e894528d0/files/html/data.rs#L580-L584.

---

This reveals an interesting inconsistency between how casting of accents and matrix delimiters works, but that's besides the point. I think my take-away is that the char cast _is_ too generic, as it shouldn't apply to either of `to_unicode` or HTML `accesskey`. The CSV delimiter is also a stretch. Accents and matrix delimiters should probably be handled consistently, but I don't think typical matrix delimiter symbols have text presentation variation selectors attached anyway.

For a minimal fix, I'd probably handle only accents and then we can look into matrix delimiters and the inconsistency separately.

marrable:
Thanks for the feedback @MDLC01 & @laurmaedje. I've updated the PR to further limit the scope: 
- Removed the modification to the generic `char` cast in `str.rs`
- Removed the direct `v: char` branch in the `Accent` cast in `accent.rs`
- `Symbol::func()` still uses `extract_base_char()` to handle the `$arrow.l.r(A)$` syntax

One question: should `extract_base_char()` stay in `str.rs`, or would another 
location make more sense given it's only used by `symbol.rs` and `accent.rs` now?

MDLC01:
I may be wrong, but I think some characters have an emoji presentation by default. As such, I think it makes more sense to remove the variation selector for the default presentation of the base codepoint instead of always the text presentation.

marrable:
> I may be wrong, but I think some characters have an emoji presentation by default. As such, I think it makes more sense to remove the variation selector for the default presentation of the base codepoint instead of always the text presentation.

@MDLC01 that's true, and users can arbitrarily construct those as symbols and pass them, but I think this is addressed in part by the [comments](https://github.com/typst/typst/pull/7456/files#diff-edee70ab1f43567233279d9d53ef4daf9c86b663871de24d471a6ee08b531513R105) in the tests by @laurmaedje for future work:
```rust
// Since we already support arbitrary characters as accents, we might want to
// support cluster like this one, too, but it should render as an emoji instead
// of normalizing into the same accent as arrow.l.r, so it's better to keep
// this an error for now."
```
Outside of symbol construction, I don't see how an emoji-default character with a text variation selector would arise, assuming `sym` and `emoji` are kept separate. Adding something like a unicode property lookup to handle default presentations might be unnecessary complexity.

laurmaedje:
I opened https://github.com/typst/typst/pull/7481 as an alternative. I played around with this PR and as you saw even pushed some commits, but I wasn't entirely satisfied with both the user-facing semantics and how the responsibilities were distributed in the code. The design space here is actually much more subtle than it seems at first! I've laid out my precise reasoning in the PR description.

crates/typst-library/src/math/accent.rs
```diff
@@ -240,9 +240,10 @@ fn create_accent_param_info() -> Vec<ParamInfo> {
 cast! {
     Accent,
     self => self.0.into_value(),
-    v: char => Self::new(v),
```
laurmaedje, line 243: I hadn't spotted this removal during earlier review, but this actually breaks string accents such `$ accent(.., "↔") $`.

marrable:
@laurmaedje yeah it definitely didn't feel like there was a surface-level regression fix that was fully satisfactory. I'm glad you could find less of a band aid fix! Good to see the stack of new tests in the alternative PR too 👍

Issue #7248, opened 2025-10-28
Links to labelled elements sometimes do not work in Chromium- and Quartz-based viewers


DavZim:
### Description

When using typst 0.14.0 and opening the PDF file with a Chromium-based PDF viewer (Chrome, Edge, etc), links that go down do not work, whereas links that go up do work.
When using Adobe Acrobat Reader or using typst 0.13.1 everything works as expected (same with using the typst.app playground).

```typst
= Top Section <top-of-doc>

Compiler Version: #sys.version

// the link down only works when compiled with 0.13.1 but not with 0.14.0
// when opened with Chromium-based browsers

The following link does not work on Chromium-based browsers: #link(<bottom-of-doc>)[Go to Page 1]

#pagebreak()

= Page 1 <bottom-of-doc>

#link(<top-of-doc>)[Go to Top Section] // this always works
```

Note that when the document is opened with Adobe Acrobat, it works.

Or in summary

| Compiler | Adobe Acrobat Reader | Chromium-based Viewer |
|----|----|----|
| 0.13.1 | Link down and up work (GOOD) | Link down and up work (GOOD) |
| 0.14.0 | Link down and up work (GOOD) | Link down is broken (BAD), link up works |

[output_0_14_0.pdf](https://github.com/user-attachments/files/23185565/output_0_14_0.pdf)
[output_0_13_1.pdf](https://github.com/user-attachments/files/23185564/output_0_13_1.pdf)

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
A few things interplay to produce this bug:

- The fact that the headings are labelled
- The ordering in which the labels appear in the document
- A bug in krilla

 For link destinations that have a label, we write a named destination into the PDF. [Krilla](https://github.com/LaurenzV/krilla) collects the named destinations in the order in which the links appear in the document, and then sorts them for a consistent order, in which it is then written into a PDF name tree. It [sorts by PDF ID](https://github.com/LaurenzV/krilla/blob/2df4b1e1fc079657accdb16da65c837d5b96a90a/crates/krilla/src/chunk_container.rs#L255-L256), which does give a consistent order. Unfortunately though, the PDF spec requires the name tree entries to appear alphabetically.

Most PDF readers probably just ignore that and collect the name tree into a hash map or similar, but Chromium seems to rely it (which is fair enough). Probably it performs a binary search or something and then doesn't find a match. So this is unfortunately indeed a regression in Typst.

[...]

PR #7458, opened 2025-11-25
Bump krilla

laurmaedje:
Resolves https://github.com/typst/typst/issues/7248 via https://github.com/LaurenzV/krilla/pull/304

In krilla, two SVG bugs were fixed since the last bump, so that's nice, too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1438,7 +1438,7 @@ dependencies = [
 [[package]]
 name = "krilla"
 version = "0.5.0"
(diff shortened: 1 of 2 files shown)


LaurenzV:
Have you double-checked that it fixes that specific example? I've only tested it with my thesis.

laurmaedje:
> Have you double-checked that it fixes that specific example? I've only tested it with my thesis.

I have

PR #7471, opened 2025-11-25
Use original logical index of line fallback text item

### LINKED ISSUE #7301: Crash when determining link parent in PDF export

saecki:
Fixes #7301.
Should be merged after #7450.

The `ShapedText` stores the `StyleChain` which contains the `LinkElem::current` property. Previously the fallback text item received a custom `LogicalIndex` which would always move it to the end of the line. And thus `modify_frame` would generate the `FrameItem::Link` outside of the `LinkMarker` introspection tags.
The fix just stores the original `LogicalIndex` in addition to the fallback text `ItemEntry`.

### DIFF
--- a/crates/typst-layout/src/inline/line.rs
+++ b/crates/typst-layout/src/inline/line.rs
@@ -222,9 +222,9 @@ fn collect_items<'a>(
 
     // Add fallback text to expand the line height, if necessary.
     if !items.iter().any(|item| matches!(item, Item::Text(_)))
-        && let Some(fallback) = fallback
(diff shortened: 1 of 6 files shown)


laurmaedje:
That makes much more sense!

PR #7473, opened 2025-11-26
Fix spacing around alignment points

mkorje:
When adding spacing around say a binary operator `+`, we would insert the space on the left after the last real (non align-point, etc.) item. This would skip over multiple alignment points preceding the operator, leading to cases where the right/left alignment wasn't actually happening (see the `math-align-spacing` test below, the one thing it should get right is having the c's be right aligned and the e's left aligned). This PR adds a check so that only a single previous alignment point is skipped.

Basically, the behaviour of spacing around alignment points can be summarised as follows now: 
> always put the correct spacing between fragments separated by an alignment point, and always put the space on the left of the alignment point

<details>

<summary>Updated tests</summary>

## math-align-spacing
| Old | New |
| - | - |
| old | math-align-spacing-NEW |
```typ
--- math-align-spacing paged ---
// Test how spacing is inserted around alignment points.
#set page(width: auto)
#grid(
  columns: 2,
  stroke: 1pt,
  inset: 1em,
  $
    a & + b + & c \
    a & + b   &   & e & + d \
    a & + b + & c & e & + d \
      &       & c &   & + d \
      & = 0
  $,
  $
    a & + b + & c \
    a & + b   &   & e & + d \
    a & + b + & c &   & + d \
      &       & c & e & + d \
      & = 0
  $,

  $
    a & + b + & c \
    a & + b   &   &   & + d \
    a & + b + & c & e & + d \
      &       & c & e & + d \
      & = 0
  $,
  $
    a & + b + & c \
    a & + b   &   & e & + d \
    a & + b + & c & e & + d \
      &       & c & e & + d \
      & = 0
  $,
)
```

## math-pagebreaking
| Old | New |
| - | - |
| math-pagebreaking-OLD | math-pagebreaking-NEW |
```typ
--- math-pagebreaking paged ---
// Test breaking of equations at page boundaries.
#set page(height: 5em)

$ a &+ b + & c \
  a &+ b   &   && + d \
  a &+ b + & c && + d \
    &      & c && + d \
    &= 0 $
```

## math-pagebreaking-numbered
| Old | New |
| - | - |
| math-pagebreaking-numbered-OLD | math-pagebreaking-numbered-NEW |
```typ
--- math-pagebreaking-numbered paged ---
// Test breaking of equations with numbering.
#set page(height: 5em)
#set math.equation(numbering: "1")

$ a &+ b + & c \
  a &+ b   &   && + d \
  a &+ b + & c && + d \
    &      & c && + d \
    &= 0 $
```

</details>


crates/typst-layout/src/math/run.rs
```diff
@@ -90,7 +90,14 @@ impl MathRun {
                 if let Some(i) = last
                     && let Some(s) = spacing(&resolved[i], space.take(), &fragment)
                 {
-                    resolved.insert(i + 1, s);
+                    let idx = if matches!(resolved.last(), Some(MathFragment::Align)) {
```
laurmaedje: Could it happen that the presence of a trailing tag makes the different between `resolved.last()` matching here and not? Maybe not, but if it can happen that would be a bug since tags should never affect layout.

laurmaedje (commented):
This test case still seems off:
```typ
$
  a & + & c & e \
  a &   & c & + d
$
```

crates/typst-layout/src/math/run.rs
mkorje: No idea if it would actually happen, but it could yeah.

Having to special case tags would be annoying... Any ideas of a nicer way to do it?

mkorje: I added a struct to make it easy in the first commit. Tags shouldn't affect layout now here.

crates/typst-layout/src/math/run.rs
```diff
@@ -90,7 +103,14 @@ impl MathRun {
...
+                        // Only one alignment point should be skipped when
+                        // inserting spacing.
+                        resolved.last_index().unwrap()
+                    } else {
+                        i + 1
+                    };
```
laurmaedje:
```suggestion
                    // Only one alignment point should be skipped when inserting
                    // spacing.
                    let idx = resolved
                        .last_index()
                        .filter(|&k| matches!(resolved[k], MathFragment::Align))
                        .unwrap_or(i + 1);
```
That way we could skip the unwrap and only search once. The `last()` method can be removed then.

crates/typst-layout/src/math/run.rs
```diff
@@ -456,3 +478,41 @@ fn spacing(
...
+
+    /// Returns the physical index of the last non-Tag fragment.
+    fn last_index(&self) -> Option<usize> {
+        self.0.iter().rposition(|f| !matches!(f, MathFragment::Tag(_)))
+    }
+}
+
+impl Deref for MathBuffer {
+    type Target = Vec<MathFragment>;
+    fn deref(&self) -> &Self::Target {
```
laurmaedje, line 505:
```suggestion
    type Target = Vec<MathFragment>;
    
    fn deref(&self) -> &Self::Target {
```

laurmaedje (commented):
I've finally understood what's happening here. The change looks good to me! I've just got one comment on the implementation and one style nit.

laurmaedje:
Thinking a bit more, I think there is still an extreme edge case: There could be alignment points _between_ `last` and `resolved.last_index()`. The `last` variable is set when we see spacing, linebreak, non-ignorant frame or glyph. It is left untouched when we see tag, space, alignment point, or ignorant frame. 

- Tags are not an issue because of MathBuffer.
- Space is not an issue becasue we don't push them into `resolved` except through the spacing code path where we updated `last`.
- Alignment point is not an issue because you added that check.
- But an ignorant frame could be in between the `last` index and `resolved.last_index()`. An ignorant frame is only created through a `PlaceElem` I think.

I tried to create an example with this, but didn't manage to get it to display something wrong. I just got this:
```typ
#let p = place[]

$
  a + & b + & c & e & + d \
  a + &     & #p c &   & + d
$
```
which at least triggers the `last_index` being an ignorant frame with an alignment point before it.

I'm not sure whether this could ever become a problem in practice, but at least it seems unintended.

mkorje:
Modifying the buffer to instead check if the fragment isn't ignorant instead of just a tag should work then, right?

laurmaedje:
Yeah, that should work!

mkorje:
Alright should be fixed now. I modified your example and added it as a test, you can see the diff in the last commit for the image.

laurmaedje:
Thank you!

PR #7474, opened 2025-11-26
Fix quoting of empty smartquotes in other languages

### LINKED ISSUE #6853: Incorrect French smartquotes after `html.frame` with HTML export
### Description

Consider the following code:

```typ
#set text(lang: "fr")
"#box(html.frame[foo])"
```

In the generated HTML, the `<body>` consists of the following paragraph:

```html
<p>« <span style="display: inline-block">(omitted for brevity)</span>« </p>
```

Notice how the closing smartquote is in the wrong direction.

When setting `text.lang` to `"en"` instead, the closing smartquote is correct:

```html
<p>“<span style="display: inline-block">(omitted for brevity)</span>”</p>
```

Tested on c7938e93. Paged export works as intended.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

aspiringLich:
Fixes #6853, although the scope is not limited to HTML.

This bug affected smartquotes in languages which defined the smartquote with a NNBSP (\u{202F}) for spacing, namely French and Latin ([source](https://github.com/typst/typst/blob/main/crates/typst-library/src/text/smartquote.rs#L247)). When encountering a second smartquote without any textual content between, the quoting logic would consider the NNBSP from the opening quote as whitespace and not actually close it. This behavior did not match the behavior for languages that did not have a NNBSP in their quotes, like English.

I updated the `smartquote-empty` test to test for multiple languages. Here is a before and after, note the sixth and seventh set of quotes for french and latin:

smartquote-comparison

I just implemented the least intrusive / simplest fix. I could also consider NBSP (\u{00A0}), but I think this is probably enough as-is. Something significantly more robust would probably require redoing the whole quoting system.

### DIFF
--- a/tests/suite/text/smartquote.typ
+++ b/tests/suite/text/smartquote.typ
@@ -198,4 +198,4 @@ $i$'s $i$ 's
 
 --- issue-6853-smartquotes-around-html-frame html ---
 #set text(lang: "fr")
-"#box(html.frame[foo])"
+"#box(html.frame[])"


laurmaedje:
I'm not really comfortable with this fix. While a case such as `"See\u{202F}"Hello" world"` is rather unlikely, it's entirely possible for a narrow non-breaking space to exist before a quote for another reason than due to a preceding quote. While I acknowledge that a proper fix is more involved and potentially involves larger change, I'd like to avoid adding a hacky fix like this into the mix.

Generally, I think the smart quoter would have to be aware of the stuff it generated itself and not consider that at all, so that regardless of which string is used as the quote, we do not run into this issue.

For the `html.frame` issue specifically, I imagine it could be reasonable to emit an OBJECT REPLACEMENT CHARACTER for that either way, which would likely also resolve the particular issue, though not the one with empty quotes.

At some point, smartquotes will likely get more attention again since there are a few other issues (c.f. https://github.com/typst/typst/issues/7615).

Due to above, I will close this fix. Thank you for the PR nonetheless! If you have suggestions for a more robust fix, I'm happy to comment on them, though unfortunately I can't currently look into it myself.

bastienvoirin:
> At some point, smartquotes will likely get more attention again since there are a few other issues (c.f. https://github.com/typst/typst/issues/7615).

If this kind of issue accumulates, it could become useful to create an umbrella issue for smarquotes, like https://github.com/typst/typst/issues/7802 for underlines, overlines, strikethrough, and highlights.

laurmaedje:
It was my impression that https://github.com/typst/typst/issues/7615 was intended as such an issue.

PR #7476, opened 2025-11-27
Avoid redundant SVG groups

saecki:
~This removes a lot of other redundant groups from SVGs.~  I'm actually not sure, if some of the groups might be necessary, I'll have to do some more investigation.
This also removed groups that previously had the `typst-group` class attribute. But I'm wondering whether we should write `typst-group` classes at all, since they aren't really semantic, and groups may be inlined by the compiler at will.

EDIT: This PR will now only generate groups for `hard` frames, which do have some semantic significance (although I'm still not sure we really need them). All soft frames that don't require a groups to have attributes will be omitted.


laurmaedje:
> But I'm wondering whether we should write typst-group classes at all, since they aren't really semantic, and groups may be inlined by the compiler at will.

I agree that these should not have existed in the first place.

crates/typst-svg/src/lib.rs
```diff
@@ -283,22 +283,27 @@ impl<'a> SVGRenderer<'a> {
...
-        self.xml.end_element();
     }
 
     /// Render a group. If the group has `clips` set to true, a clip path will
     /// be created.
     fn render_group(&mut self, state: &State, group: &GroupItem) {
-        self.xml.start_element("g");
-        self.xml.write_attribute("class", "typst-group");
+        let mut started_group = false;
+        let mut group_writer = LazyCell::new(|| {
```
laurmaedje, line 292: could this use the `deferred` mechanism? I remember that we talked about this, but not the conclusion.

saecki: No, it doesn't really apply here. The (somewhat) manual implementation here is replaced later on with a custom `LazyElement`, which handles this a bit more elegantly.

PR #7478, opened 2025-11-27
Fix caching of pdf tagging text attribute parameters

saecki:
I had a messed up understand of what the font index actually is. This was just plain wrong :sweat_smile:


laurmaedje:
I think we want to have collision-safety here, right? If so, `typst_utils::hash128` would be more appropriate because rustc_hash can have collisions easily. Both will be quite cheap because `Font` objects are prehashed anyway.

laurmaedje:
Actually, do we need `TextParams` as a struct at all? We could just as well hash the font and size into the same hash, couldn't we?

saecki:
Good point

PR #7481, opened 2025-11-28
Fix `arrow.l.r` not working as an accent

laurmaedje:
This is an alternative to #7456. I played around with that PR and even pushed some commits, but I wasn't entirely satisfied with both the user-facing semantics and how the responsibilities were distributed in the code. The design space here is actually much more subtle than it seems at first!

What struck me as strange is that `Accent::combining` (previously `Accent::combine`) expected _its caller_ to handle variation selectors. The purpose of the function is to select an appropriate combining accent based on what the user entered. For example, it turns `"^"` into U+0302 Combining Circumflex Accent. Given that it already lists different variants with different options / names for various accents, I think its an appropriate expectation that it takes the full symbols value and handles variation selectors internally (they are just another option after all), making an `&str` input more appropriate.

This change to `Accent::combining` also let me find a place in `typst-docs` that had to be fixed, but was missing from #7456. The reason why this was still wrong was precisely because part of `Accent::combining`'s logic had to live outside of it previously.

Similarly, the purpose of `Accent::normalize` (previously `Accent::new`) is to select the appropriate combining accent or fall back to just the character itself. Thus, it must inherit the input type of `Accent::combining`. The string cast in `cast! { Accent, .. }` can then directly dispatch to this and no function with char input is needed. (This string cast was missing in #7456, but that's fixable.)

The second place where base extraction happens in #7456 is in `Symbol::func`. This function internally calls `get_accent_func` and `get_lr_wrapper_func` whose purpose it is to select a function belonging to a symbol. Given that symbols can be whole clusters rather than just chars since 0.14, having an `&str` argument on these seems more appropriate than a `char` here as well, since in theory other such getters could desire different normalization behavior and I think they should all have the same signature to reduce coupling with `Symbol::func`. This works out very neatly because then `get_accent_func` gets the variation selector handling for free by virtue of calling `Accent::combining`.

So that's the reasoning for changing the API from `char` to `&str` in various places. What's remaining is the variation selector logic itself. We now have two places that get an `&str` input, but internally have a list of chars: `Accent::combining` and `get_lr_wrapper_func`. In `get_lr_wrapper_func`, we could strip the variation selector, but it would be pointless because there isn't any text presentation selector defined for any of the L/R symbols we match on. On the other hand, in `get_accent_func`, there is only `arrow.l.r`.

The two options we have here are (a) to somehow strip presentation selectors (probably only the text one) for all accents, or (b) just adding `"↔\u{fe0e}"` to the name list of the left-right arrow accent. The second option is certainly simpler. It might seem more brittle, because you have to manually add these, but I expect very little change here honestly. More importantly, I think it results in clearer user-exposed semantics. Stripping any kind of variation selector is a problem because then `emoji.arrow.l.r` will suddenly work (c.f. https://github.com/typst/typst/pull/7456#issuecomment-3574320319). Stripping only the text presentation selector reaches the desired result, but it's still kind of nonsensical for symbols that don't have a text presentation selector defined. All of this is observable in raw string use as in `$ accent(.., "↔\u{fe0e}") $` and I think erroring for ill-defined sequences makes more sense than silently stripping.

I also added a bunch of tests to ensure all styles of accent usage (there are five!) work appropriately.

Fixes https://github.com/typst/typst/issues/7437.


MDLC01:
I'm a bit late to the party but I like that you chose option (b).

PR #7486, opened 2025-11-29
Change `1 of I` to `1 to 1`

mattalxndr:
Typo, I think.


MDLC01:
I suspect this is a way to illustrate how different numeral systems can be used in the same numbering kind.

laurmaedje:
Yes, this is indeed intentional.

PR #7488, opened 2025-11-29
Return friendly error when loading JSON has a leading UTF‑8 BOM

### LINKED ISSUE #3157: `json(path)` and `cbor(path)` do not support UTF8 with BOM
### Description

`json(path)`, `cbor(path)` cannot read a file with utf8-bom.

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

### LINKED ISSUE #5440: Notify when JSON encoding is not UTF-8
### Description

Currently, typst will output a vague error when reading a file with a byte order mark (BOM).
```
watching report.typ
writing to report.pdf

[17:42:00] compiled with errors

error: failed to parse JSON (expected value at line 1 column 1)
    ┌─ report.typ:216:23
    │
216 │ #let data-stock = json("data/stock.json")
    │                        ^^^^^^^^^^^^^^^^^
```

Adding a more descriptive error in this case would be ideal in order to give the user some direction on how to correct it—this could be done by detecting a BOM and indicating it:

```
watching report.typ
writing to report.pdf

[17:42:00] compiled with errors

error: failed to parse JSON (byte order mark present at start of file)
    ┌─ report.typ:216:23
    │
216 │ #let data-stock = json("data/stock.json")
    │                        ^^^^^^^^^^^^^^^^^
```

While this is the best solution, this already approaches the amount of work required to automatically handle BOM itself anyways (#3157). Alternatively, the error message could simply indicate that there was something unexpected at the beginning of the file stream and allude to an encoding error:

```
watching report.typ
writing to report.pdf

[17:42:00] compiled with errors

error: failed to parse JSON (expected value at line 1 column 1, verify encoding is UTF-8)
    ┌─ report.typ:216:23
    │
216 │ #let data-stock = json("data/stock.json")
    │                        ^^^^^^^^^^^^^^^^^
```

### Use Case

While not an issue in a lot of cases, some software does export files with the BOM (Microsoft Excel CSV export). The existing error message doesn't indicate any specific problem with the file, so it's up to a user to investigate and become aware of how typst interacts with BOM text files.

3w36zj6:
close #5440
close #3157

This pull request improves the handling of JSON files in the loading library by adding explicit detection and error reporting for files that start with a UTF-8 Byte Order Mark (BOM), which is not allowed by the JSON specification.

Although the JSON specification does not allow BOM, we could consider a more forgiving approach that removes the BOM when encountering BOM-prefixed JSON files. However, in my opinion, most major programming languages throw an error for JSON files with a BOM. To maintain consistent and clear behavior, this PR opts to throw a friendly error message in such cases.

### DIFF
--- a/crates/typst-library/src/loading/json.rs
+++ b/crates/typst-library/src/loading/json.rs
@@ -1,7 +1,7 @@
 use ecow::eco_format;
 use typst_syntax::Spanned;
 
-use crate::diag::{At, LineCol, LoadError, LoadedWithin, SourceResult};
+use crate::diag::{At, LineCol, LoadError, LoadedWithin, SourceResult, bail};
 use crate::engine::Engine;
 use crate::foundations::{Str, Value, func, scope};
 use crate::loading::{DataSource, Load, Readable};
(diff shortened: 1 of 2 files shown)


laurmaedje:
Thanks!

PR #7490, opened 2025-11-30
Show full font file path in `typst fonts --variants`

### LINKED ISSUE #6943: Allow fonts subcommand to also display the full filename path
### Description

Allowing the typst fonts subcommand to also display the full filename path of each registered font is extremely helpful for debugging, for several reasons.

For prior/example code – `tinymist` has that functionality about around here – https://github.com/Myriad-Dreamin/tinymist/blob/51f9cdec81a8eeb20fa8cb30538b251ac8928b64/tools/editor-tools/src/features/font-view/components/font-list.ts#L39

### Use Case

Allowing typst fonts to show the full font file path helps debugging by:

* Disambiguation: Distinguishes between fonts with the same family name.
* Source clarity: Reveals whether a font came from system, .local, or custom directories.
* Troubleshooting: Identifies why the the particular font was loaded.
* Cross-platform checks: Confirms which file is picked up on different systems.

### Example

i have the following output from `typst 0.13.1`, and dont know what is going on:

```
Clarendon Sys
- Style: Italic, Weight: 500, Stretch: FontStretch(1000)
- Style: Italic, Weight: 400, Stretch: FontStretch(1000)
- Style: Italic, Weight: 400, Stretch: FontStretch(1000)
- Style: Normal, Weight: 300, Stretch: FontStretch(1000)
- Style: Italic, Weight: 300, Stretch: FontStretch(1000)
- Style: Italic, Weight: 300, Stretch: FontStretch(1000)
- Style: Normal, Weight: 500, Stretch: FontStretch(1000)
- Style: Italic, Weight: 700, Stretch: FontStretch(1000)
- Style: Normal, Weight: 700, Stretch: FontStretch(1000)
- Style: Italic, Weight: 800, Stretch: FontStretch(1000)
- Style: Italic, Weight: 500, Stretch: FontStretch(1000)
- Style: Italic, Weight: 800, Stretch: FontStretch(1000)
- Style: Italic, Weight: 700, Stretch: FontStretch(1000)
- Style: Normal, Weight: 400, Stretch: FontStretch(1000)
- Style: Normal, Weight: 400, Stretch: FontStretch(1000)
- Style: Normal, Weight: 500, Stretch: FontStretch(1000)
- Style: Normal, Weight: 300, Stretch: FontStretch(1000)
- Style: Normal, Weight: 800, Stretch: FontStretch(1000)
- Style: Normal, Weight: 700, Stretch: FontStretch(1000)
- Style: Normal, Weight: 800, Stretch: FontStretch(1000)
```

### Reasoning

the addition might be non-invasive and non- api breaking as a quick-win to be included in the next release (ie. 0.14 ?)

3w36zj6:
close #6943

This pull request adds full font file paths to the detailed information displayed by `typst fonts --variants`.

If this change is acceptable, we can also consider implementing json output functionality for integration with external tools, as proposed in https://github.com/typst/typst/issues/6943#issuecomment-3553447343.

### DIFF
--- a/crates/typst-cli/src/fonts.rs
+++ b/crates/typst-cli/src/fonts.rs
@@ -1,3 +1,5 @@
+use std::path::Path;
+
 use typst::text::FontVariant;
 use typst_kit::fonts::Fonts;
 
@@ -15,20 +17,18 @@ pub fn fonts(command: &FontsCommand) {
         println!("{family}");
         if command.variants {
             for index in indices {
-                if let Some(font_info) = fonts.book.info(index) {
-                    let FontVariant { style, weight, stretch } = font_info.variant;
-                    let path = fonts
-                        .slots
-                        .get(index)
-                        .and_then(|slot| {
-                            slot.path().map(|p| p.to_string_lossy().to_string())
-                        })
-                        .unwrap_or_else(|| "<embedded>".to_string());
+                let Some(font_info) = fonts.book.info(index) else { continue };
+                let FontVariant { style, weight, stretch } = font_info.variant;
+                let path = fonts
+                    .slots
+                    .get(index)
+                    .and_then(|slot| slot.path())
+                    .unwrap_or_else(|| Path::new("<embedded>"))
+                    .display();
 
-                    println!(
-                        "- Style: {style:?}, Weight: {weight:?}, Stretch: {stretch:?}, Path: {path}"
-                    );
-                }
+                println!(
+                    "- Style: {style:?}, Weight: {weight:?}, Stretch: {stretch:?}, Path: {path}",
+                );
             }
         }
     }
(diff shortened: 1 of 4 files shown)


laurmaedje (commented):
Implementation-wise, instead of mirroring the families API with `families_with_slots` and extra structs, I'd suggest changing the return type of `FontBook::families` to `impl Iterator<Item = (&str, impl Iterator<Item = usize>)> + '_`, returning just the font indices. Then, we can get the font info via `book.info(index)` and the path via `fonts.slots[index].path()` and there's no need for new types.

The existing call sites of `book.families()` would need to be updated to manually access the `book.info` for the given IDs, but that shouldn't be a big trouble.

The test is a bit lengthy as-is. I think with this change, we could just skip it, since there wouldn't really be new functionality, just the existing one exposed differently.

laurmaedje:
Thanks!

PR #7493, opened 2025-11-30
Allow trailing semicolon for hints in `bail!`, `error!`, and `warning!` macros

isuffix:
It was annoying me how using the `hint: "..."` syntax in the `bail!` macro would error with a trailing semicolon (especially because one of the examples in the docstring used one!), so I rolled up my sleaves and figured out how to change it. Quite tricky, but works great now. It even errors if you use a trailing comma instead of a trailing semicolon on a hint.

I updated all the macro callsites to use trailing semicolons (if the macro had hints), and I fixed some other formatting issues like splitting long lines or adding trailing commas (if used multiline with no hints).

I also updated the comments on the macros in `diag.rs`, LMK if any of those need rewording.

### DIFF
--- a/crates/typst-cli/src/compile.rs
+++ b/crates/typst-cli/src/compile.rs
@@ -119,7 +119,7 @@ impl CompileConfig {
                 _ => bail!(
                     "could not infer output format for path {}.\n\
                      consider providing the format manually with `--format/-f`",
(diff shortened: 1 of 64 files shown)


laurmaedje:
Thanks, this was annoying me as well!

Issue #2613, opened 2023-11-07
Controlling colors and lines of the references as in hyperref


Atreyu-94:
### Description

It would be great to be able to control the colors of the references/citations as you can control them in hyperref/pdfLaTeX. This is very useful for creating documentation or educational content, for example:

|Lines|Colors|
|----------|--------|
|![imagen](https://github.com/typst/typst/assets/39834909/b1c82954-6ceb-4267-99f2-48021da34b4f) | ![imagen](https://github.com/typst/typst/assets/39834909/72db1ec2-c7af-47c6-aa2e-8017371bacbd) |

Some features of `hyperref`:
- Have some color around the link when the pdf file is opened in a viewer
- Have no color around the link if it is printed on a paper (the same with text color that disappears when printed).

This was discussed here: #2585 

### Use Case

- When we are using a PDF document to give a presentation or make an educational document, this is often important to know exactly which elements of the document are referenced.
- It is important that these links/colors that are not designed for printing do not appear in printed documents.
- Many times we want to differentiate between references, to know if it is a cite, a cross-reference, a link or simply a local file.

Enivex:
It seems to me like the actual request here is for support for the optional content group PDF feature

Antoinemarteau:
It would be amazing to be able to collapse / expand section of text / proofs in a typst generated PDF.

Enivex:
> It would be amazing to be able to collapse / expand section of text / proofs in a typst generated PDF.

At most you would be able to hide them. The remaining text wouldn't reflow. That's just inherently not possible in the PDF format.

Antoinemarteau:
Ah, ok, and would it be possible that some text is collapsed by default, and you could expand it over the exiting follow up text momentarily ?

Enivex:
> Ah, ok, and would it be possible that some text is collapsed by default, and you could expand it over the exiting follow up text momentarily ?

I don't know the exact limitations of OCGs, but that sounds more reasonable.

PR #7494, opened 2025-11-30
Link boxes like LaTeX's hyperref

akretz:
It would be cool if typst was able to produce link boxes similar to LaTeX's hyperref. There is interest in this feature and the usually suggested workarounds use Box elements around links. These workarounds have the downside that, unlike hyperref boxes, the boxes also appear in printed documents.
https://forum.typst.app/t/how-to-add-color-box-to-a-non-citation-reference/3735
#2585
#2613

The possibility to add boxes to links has been implemented in the PDF backend with LaurenzV/krilla#301. With this, it is possible to add borders to links in PDF which behave like hyperref link borders. I'm not sure what the best way is to integrate this in typst as I don't really know the typst codebase. This PR is a draft where I have added a `border` attribute to the `link` element and propagated it all the way to the PDF backend:

```typst
#link("https://example.com", border: blue)
```

Does this approach make sense? It works as intended in my document.

### DIFF
--- a/crates/typst-layout/src/rules.rs
+++ b/crates/typst-layout/src/rules.rs
@@ -224,9 +224,10 @@ const LINK_RULE: ShowFn<LinkElem> = |elem, engine, styles| {
     let body = elem.body.clone();
     let dest = elem.dest.resolve(engine, span)?;
     let alt = dest.alt_text(engine, styles, span)?;
+    let border = elem.border.get_cloned(styles);
     // Manually construct link marker that spans the whole link elem, not just
     // the body.
-    Ok(LinkMarker::new(body, Some(alt))
+    Ok(LinkMarker::new(body, Some(alt), border)
         .pack()
         .spanned(span)
         .set(LinkElem::current, Some(dest)))
--- a/crates/typst-library/src/foundations/content/mod.rs
+++ b/crates/typst-library/src/foundations/content/mod.rs
@@ -467,7 +467,7 @@ impl Content {
     /// Link the content somewhere.
     pub fn linked(self, dest: Destination, alt: Option<EcoString>) -> Self {
         let span = self.span();
-        LinkMarker::new(self, alt)
+        LinkMarker::new(self, alt, None)
             .pack()
             .spanned(span)
             .set(LinkElem::current, Some(dest))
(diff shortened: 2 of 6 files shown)


isuffix:
Nice work! :tada:

One note is that we should probably allow the various [stroke options](https://typst.app/docs/reference/visualize/stroke/), and maybe things like [`rect.radius`](https://typst.app/docs/reference/visualize/rect/#parameters-radius). We could document this as a `pdf-hover` argument that takes a dictionary and forwards the parameters to `rect`? Although I don't know how flexible these boxes are on the pdf side.

RungeCC:
> One note is that we should probably allow the various [stroke options](https://typst.app/docs/reference/visualize/stroke/), and maybe things like [rect.radius](https://typst.app/docs/reference/visualize/rect/#parameters-radius). We could document this as a pdf-hover argument that takes a dictionary and forwards the parameters to rect? Although I don't know how flexible these boxes are on the pdf side.

As a part of pdf annotation, `/Border` only support `h-radius`, `v-radius`, border style is also restricted. See PDF32000 sec 12.5: Annotations for more details. 

So it could not fully support `stroke` or `rect.radius` options, though maybe there're some other way to implement those in PDF backend. However, it seems fair to allow to set `dashed` or thickness.

akretz:
Thanks for the quick feedback! I guess using a `Stroke` attribute doesn't make so much sense here because many of its properties would be unused. The way I read the PDF standard it could make sense to introduce another `AttributeBorder` struct with attributes

* paint
* thickness
* dash
* h-radius
* v-radius

and pass that to the `link` element.

laurmaedje:
I'm not sure how to expose this, but I must admit that I'm not a big fan of exposing something that is so clearly PDF-specific as just `link.border`. (For instance, `heading.bookmarked` is also fairly PDF-specific, but it could at least somehow be interpreted as a more general semantic property; and I'm also not a big fan of it anyway.)

laurmaedje:
As mentioned, I'm not comfortable with the design as-is, so I'll close this for now. For design discussion, https://github.com/typst/typst/issues/2613 would be suitable and once we have settled on something we could reopen this or have a new PR. I'm sorry that landing this didn't turn as hoped, thank you nonetheless!

PR #7500, opened 2025-12-01
Change color of values in `typst info`

laurmaedje:
This is easier to read (at least in my terminal, YMMV), more in line with Typst's brand colors and also in line with the colors we opted for in https://github.com/typst/typst/pull/7443.


laurmaedje:
Before:

image

After:

image

PR #7505, opened 2025-12-01
Use phrasing content for paragraph grouping in HTML export

TheZoq2:
This was briefly discussed in Discord (https://discord.com/channels/1054443721975922748/1054443722592497796/1444060883826376898) and while we didn't came to any sort of conclusion there, I thought i'd open a PR for a proposed solution.

Barring a better source of "inline elements", I think Phrasing Content is the closest we can get.

The reason I started looking into this in the first place is that with an incorrect list of "inline elements", the html output will insert `<p>` tags in places where they should not be allowed. For example, the following code

```typ
#html.elem("div",
  html.elem("span", {
    {
      html.elem("input")
      html.elem("span", [span])
    }
  })
)
```

produces 
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <div><span><input><p><span>span</span></p></span></div>
  </body>
</html>
```

due to `input` not being in the current list of inline elements. This in turn is invalid HTML, at least according to https://validator.w3.org/nu/#textarea

With this fix, it will instead generate 
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <div><span><input><span>span</span></span></div>
  </body>
</html>
```
which is valid.

### DIFF
--- a/crates/typst-html/src/lib.rs
+++ b/crates/typst-html/src/lib.rs
@@ -119,7 +119,7 @@ impl HtmlElem {
         }
     }
 
-    /// Checks whether the given element is an inline-level HTML element.
+    /// Checks whether the given element is "phrasing content" in HTML.
     fn is_phrasing(elem: &Content) -> bool {
         elem.to_packed::<HtmlElem>()
             .is_some_and(|elem| tag::is_phrasing_content(elem.tag))
--- a/crates/typst-library/src/routines.rs
+++ b/crates/typst-library/src/routines.rs
@@ -120,10 +120,10 @@ pub enum RealizationKind<'a> {
     /// This the root realization for HTML. Requires a mutable reference to
     /// document metadata that will be filled from `set document` rules.
     ///
-    /// The `is_inline` function checks whether content consists of an inline
-    /// HTML element. It's used by the `PAR` grouping rules. This is slightly
-    /// hacky and might be replaced by a mechanism to supply the grouping rules
-    /// as a realization user.
+    /// The `is_phrasing` function checks whether content consists of a
+    /// "phrasing content" HTML element. It's used by the `PAR` grouping rules.
+    /// This is slightly hacky and might be replaced by a mechanism to supply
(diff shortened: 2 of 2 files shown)


laurmaedje:
I traced back the history of `is_inline_by_default` and it turns out that I added it in the very initial PR that added HTML export. I know that I didn't think too deeply about it there and just shipped something. Phrasing content, as what's contained in HTML `<p>` elements, indeed seems like the correct guideline for what should go into a Typst paragraph.

The `is_..._by_default` functions are supposed to reflect an element's default CSS `display` value. The other two are only used for pretty printing where that's precisely what we want because the CSS `display` property determines how spaces interact with the element and pretty-printing introduces additional spaces.

For paragraph grouping, however, it's just the wrong tool. I've added all the content model functions in https://github.com/typst/typst/pull/7563 and I would suggest rebasing this PR on top of it and simply deleting `is_inline_by_default` and using `is_phrasing_content` in the paragraph grouping code. The `is_inline` function used there can just be renamed to `is_phrasing` then.

TheZoq2:
Thanks, done! I did some LSP refactoring and rough find and replace to try to get rid of a few more mentions of "inline" around HTML stuff though I'm not sure if I went far enough and/or should go further with that. Without knowing anything about the rest of the code it seems like I eventually run into a typst notion if of inline (for example `FragmentKind::Inline` in `typst-library/routines.rs`) and I presume that should still be "inline"

TheZoq2:
I blanket accepted all the tests, presumably these will cause subtle differences in output but I think that is unavoidable with this patch

laurmaedje:
Thanks!

laurmaedje:
The PR introduced a few new issues (see https://github.com/typst/typst/issues/7974).

I've opened a proposal for how we could fix this: https://github.com/typst/typst/issues/8159

PR #7506, opened 2025-12-01
Add support for tiling offset

dayala1:
Current tilings can define a relative starting point, but it is not possible to do small (or not so small) adjustements to their position. This pull request adds the optional "offset" parameter to tilings so that it is possible to do so. For example:

 ```typst
 // Tiling without offset
#let pat1 = tiling(
  size: (30pt, 30pt),
  relative:"parent"
)[
  #place(rect(width: 15pt, height: 15pt, fill: black))
]

// Tiling with offset
#let pat2 = tiling(
  size: (30pt, 30pt),
  relative:"parent",
  offset: (-7.5pt,  -7.5pt)
)[
   #place(rect(width: 15pt, height: 15pt, fill: black))
]

No offset:
#block(fill: pat1, width: 50%, height: 60pt)

#v(1em)

Offset:
#block(fill: pat2, width: 50%, height: 60pt)
```

Gives the following output:

image

### CODE REVIEW

**Comment by laurmaedje** [comment]
This reminds me of https://typst.app/universe/package/modpattern/, which IIUC correctly does this, but in user space, by repeating the pattern 9 times around the core pattern zone.

Am I correct in assuming that it's basically that but natively?

**Comment by dayala1** [comment]
In a way it is similar. In the package, repeating the pattern in surrounding tiles allows patterns to tile as if the contents went putside the tile edges, enabling the displacement since when displacing, elements move outside the tile but are "saved" by the repetition. Of course at the cost of multiplying the elements that need to be drawn and rendered even if seamless tiling wasnt necessary.

Of course the underlaying mechanism is unrelated. This PR enables the offset of the tiles being drawn. The package just moves the elements within the tile, but thanks to the repetition, they still appear as tiling in the same way

The pull request uses the straightforward transformation of the tiling pattern in the apis that display it, which as can be seen required very little code.

**Comment by dayala1** [comment]
@laurmaedje Added offset to svg tiling key along with size and spacing. Also, renewed hash of tiling-related tests and added offset tests.

File: crates/typst-render/src/paint.rs [hunk 1]
```diff
@@ -245,8 +247,8 @@ pub fn to_sk_paint<'a>(
                 sk::FilterQuality::Nearest,
                 1.0,
                 fill_transform
-                    .pre_scale(1.0 / state.pixel_per_pt, 1.0 / state.pixel_per_pt)
-                    .pre_translate(offset.x.to_f32(), offset.y.to_f32()),
+                    .pre_translate(offset.x.to_f32(), offset.y.to_f32())
```
saecki: Why did you swap the order of transforms here?
~I've noticed that the strokes on the PDF version seems thinner than the SVG one. Did you figure out why that could be the case?~ This was a bug with zathura

crates/typst-render/src/paint.rs
```diff
@@ -233,9 +233,11 @@ pub fn to_sk_paint<'a>(
 
             let offset = match relative {
                 RelativeTo::Self_ => {
-                    gradient_map.map(|(offset, _)| -offset).unwrap_or_default()
+                    let base_offset =
+                        gradient_map.map(|(offset, _)| -offset).unwrap_or_default();
+                    Point::new(base_offset.x + tilings.dx(), base_offset.y + tilings.dy())
                 }
-                RelativeTo::Parent => Point::zero(),
+                RelativeTo::Parent => Point::new(tilings.dx(), tilings.dy()),
```
saecki: Shouldn't the offset also be applied to tilings that aren't relative to the parent?

crates/typst-svg/src/paint.rs
```diff
@@ -86,15 +86,17 @@ impl SVGRenderer<'_> {
 
     pub(super) fn push_tiling(&mut self, tiling: &Tiling, ts: Transform) -> DedupId {
         let tiling_size = tiling.size() + tiling.spacing();
+        let tiling_offset = tiling.offset();
         // Unfortunately due to a limitation of `xmlwriter`, we need to
         // render the frame twice: once to allocate all of the resources
         // that it needs and once to actually render it.
         let rendered = self.render_tiling_frame(&State::new(tiling_size), tiling.frame());
 
         // Use the rendered SVG and the tiling size as a key, since the `Tiling`
```
saecki: The comment should also be updated

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -236,6 +279,21 @@ impl Tiling {
...
+
+    /// Return the horizontal offset of the tiling in absolute units.
+    pub fn dx(&self) -> Abs {
+        self.0.offset.x
+    }
+
+    /// Return the vertical offset of the tiling in absolute units.
+    pub fn dy(&self) -> Abs {
+        self.0.offset.y
+    }
```
saecki: I'm not sure these getters pull their weight. And from their name it's not directly obvious that they only refer to the X/Y component of the offset. Just using `offset().x` seems clearer to me.

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -269,14 +327,22 @@ impl Repr for Tiling {
         let mut out =
             eco_format!("tiling(({}, {})", self.0.size.x.repr(), self.0.size.y.repr());
 
-        if self.0.spacing.is_zero() {
+        if !self.0.spacing.is_zero() {
```
saecki, line 333: Good catch!

saecki (commented):
I think having this natively using transforms instead of repetition would be nice. Here are some remarks.

crates/typst-render/src/paint.rs
dayala1: If the translation took place before, it was affected by the scaling, so I changed it to translate first, scale afterwards.

Here is an example of how the pdf export and svg are and how the png one would be without the inversion:

pdf and svg:

image

png:

image

I just checked and the strokes, at least in the examples I'm testing, seem to be the same for patterns in both formats. Still, I need to verify tests since with the relative setting, the formats etc I need to make sure I didn't mess anything up

crates/typst-render/src/paint.rs
dayala1: Unless I'm missing something the offset is applied in both cases. In the tests I have done the offset is indeed aplpied in both. However, both when building and with current release typst it is impossible for me to observe any difference between "parent" and "self" options, no matter what blocks setups I create. The documentation also has no example, so I wonder if I'm doing things wrong or the parent option is bugged.

I have tested with blocks and text.

crates/typst-library/src/visualize/tiling.rs
dayala1: Yep, was conflicted about it, so I will remove it

crates/typst-render/src/paint.rs
saecki: The reason this won't affect the tests, is because the test suite uses 1 pixel_per_point for the references. The `base_offset` as you've called it will have to be applied in the previous order and the tiling offset in the changed one.
Alternatively the tiling offset could also be pre-multiplied with the pixel per point.

Have a look at the following test:
```typ
#set page(width: 100pt, height: 100pt)

#rect(
  width: 100%,
  height: 100%,
  fill: tiling(relative: "self", size: (30pt, 30pt))[
    #square(width: 100%, height: 100%, stroke: 1pt, fill: green)
  ],
  stroke: (
    thickness: 10pt,
    paint: tiling(relative: "self", size: (30pt, 30pt))[
      #square(width: 100%, height: 100%, stroke: 1pt, fill: blue)
    ]
  ),
)
```

The images were compiled with `--ppi=400`, if you pass `--ppi=72` (1 pixel per point), it will render like the test suite.
|`main`|PR|
|---|---|
| test | test |

Although unrelated to this PR, this reminds me of a change to gradients recently: https://github.com/typst/typst/pull/7863#discussion_r2899492214. Do you think we should also change the tiling bounding boxes of strokes @eltos?

crates/typst-render/src/paint.rs
saecki: > Unless I'm missing something the offset is applied in both cases. In the tests I have done the offset is indeed aplpied in both.

You're right, I've just missed it above.
~You can also directly add the offset by converting it to a point: `tilings.offset().to_point()`.~
The offset shouldn't be added at all here, as mentioned in the other comment, it should be applied separately to the transform.

> However, both when building and with current release typst it is impossible for me to observe any difference between "parent" and "self" options, no matter what blocks setups I create. The documentation also has no example, so I wonder if I'm doing things wrong or the parent option is bugged.
> 
> I have tested with blocks and text.

I think for text the default is `"parent"` and it can't be set to `"self"`. Here's an example of how relative changes the behavior of the tiling:
```typ
#let t(stroke-color, ..args) = tiling(size: (30pt, 30pt), ..args)[
  #set line(stroke: stroke-color)
  #place(top + left, line(start: (0%, 0%), end: (100%, 100%), stroke: 1pt))
  #place(top + left, line(start: (0%, 100%), end: (100%, 0%), stroke: 1pt))
]

#set page(fill: t(blue), width: 120pt, height: 230pt, margin: 10pt)
#rect(
  width: 100pt, height: 100pt,
  fill: t(red, relative: "parent"),
  stroke: 1pt + green,
)
#v(weak: true, 10pt)
#rect(
  width: 100pt, height: 100pt,
  fill: t(red, relative: "self"),
  stroke: 1pt + green,
)
```
test

eltos:
> Although unrelated to this PR, this reminds me of a change to gradients recently: https://github.com/typst/typst/pull/7863#discussion_r2899492214. Do you think we should also change the tiling bounding boxes of strokes @eltos?


Tilings have no angle and always a fixed absolute size, so there are no issues when they are used on strokes where one bounding box dimension may collapse, and hence there is no immediate need to do such a change also for tilings. It could be done for consistency though. If done, the only effect as far as I can tell would be an offset displacement of the tiling. In @saecki's [example above](https://github.com/typst/typst/pull/7506#discussion_r3021889884), the black lines on fill and stroke tiling would then no longer be aligned.
I don't have a strong opinion other than doing such change after this PR to allow the offset to be manipulated if required.

Here's an example of tilings as strokes, as you can see, currently the tiling anchor is the starting point of the line/curve (green dots):




```typ
#for cap in ("square", "butt", "round"){
  box(line(length: 10pt, stroke: (
    thickness: 20pt,
    paint: tiling(size: (8pt, 10pt))[
  #place(line(start: (0%, 0%), end: (100%, 100%), stroke: red))
  #place(line(start: (0%, 100%), end: (100%, 0%)))
],
    cap: cap
  )), width: 30pt)
}
```




```typ
#set page(width: auto, height: auto, margin: 5pt)
#let pat = tiling(size: (8pt, 10pt))[
  #place(line(start: (0%, 0%), end: (100%, 100%), stroke: red))
  #place(line(start: (0%, 100%), end: (100%, 0%)))
]
#grid(columns: (100pt, 100pt, 100pt), rows: 100pt,
  for i in range(0,12){
    place(dx: 50pt + 12pt*calc.cos(i*30deg), dy: 50pt + 12pt*calc.sin(i*30deg),
      line(length: 36pt, angle: i*30deg, stroke: pat + 10pt)
    )
  },
  for i in range(0,12){
    place(dx: 50pt + 12pt*calc.cos(i*30deg), dy: 50pt + 12pt*calc.sin(i*30deg),
      curve(
        curve.line((36pt * calc.cos(i*30deg), 36pt * calc.sin(i*30deg))),
        stroke: pat + 10pt,
      )
    )
  },
  for i in range(0,12){
    place(dx: 50pt + 12pt*calc.cos(i*30deg), dy: 50pt + 12pt*calc.sin(i*30deg),
      curve(
        curve.quad((18pt * calc.cos((i+1)*30deg), 18pt * calc.sin((i+1)*30deg)), (36pt * calc.cos(i*30deg), 36pt * calc.sin(i*30deg))),
        stroke: pat + 10pt,
      )
    )
  },
)
```

crates/typst-render/src/paint.rs
eltos: > the test suite uses 1 pixel_per_point for the references

Side note @saecki: Maybe now with the improved test suit you could use 144ppi (2ppp) or the like to avoid such issues to go unnoticed. If you use the same hash-based method as for PDF/SVG/etc it would not bloat the repo size.

dayala1: @saecki Perfect, added the fix to the PR, separating the offset transformation. Result seems to be the same as main, and offset works fine. Also tested the example with self and parent relativeness.

saecki:
> Tilings have no angle and always a fixed absolute size, so there are no issues when they are used on strokes where one bounding box dimension may collapse, and hence there is no immediate need to do such a change also for tilings. It could be done for consistency though. If done, the only effect as far as I can tell would be an offset displacement of the tiling. In @saecki's [example above](https://github.com/typst/typst/pull/7506#discussion_r3021889884), the black lines on fill and stroke tiling would then no longer be aligned.
> I don't have a strong opinion other than doing such change after this PR to allow the offset to be manipulated if required.

Ok, thanks for the detailed reply! If there are no issues with the current behavior, I think it's nice to have them aligned :)

crates/typst-render/src/paint.rs
saecki: > Side note @saecki: Maybe now with the improved test suit you could use 144ppi (2ppp) or the like to avoid such issues to go unnoticed. If you use the same hash-based method as for PDF/SVG/etc it would not bloat the repo size.

That would be nice indeed. The problem is that typst-render currently isn't Cross-Platform deterministic. Though some solution for this is definitely on the bucket list

saecki (approved):

saecki:
Thanks for staying on top of this!
We've talked about the PR, and thought that it would also be useful to specify relative offsets, e.g. `(50%, 50%)`.
In the Rust code that would mean the type of `offset` would change to `Axes<Rel>`, which would be resolved relative to the `size + spacing` in the `construct` function. `TilingInner::offset` would stay `Size` (`Axes<Abs>`).

dayala1:
> Thanks for staying on top of this! We've talked about the PR, and thought that it would also be useful to specify relative offsets, e.g. `(50%, 50%)`. In the Rust code that would mean the type of `offset` would change to `Axes<Rel>`, which would be resolved relative to the `size + spacing` in the `construct` function. `TilingInner::offset` would stay `Size` (`Axes<Abs>`).



> Thanks for staying on top of this! We've talked about the PR, and thought that it would also be useful to specify relative offsets, e.g. `(50%, 50%)`. In the Rust code that would mean the type of `offset` would change to `Axes<Rel>`, which would be resolved relative to the `size + spacing` in the `construct` function. `TilingInner::offset` would stay `Size` (`Axes<Abs>`).

Good idea, could be useful for adjusting tilings independently of size. Made the change.

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -218,13 +218,12 @@ impl Tiling {
             bail!(spacing.span, "tile spacing must be finite");
         }
 
-        // Ensure that offset is absolute.
-        if !offset.v.x.em.is_zero() || !offset.v.y.em.is_zero() {
-            bail!(offset.span, "tile offset must be absolute");
-        }
-
```
saecki: We actually still want to disallow font-relative lengths, so this check should stay.

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -212,10 +253,17 @@ impl Tiling {
...
+
         Ok(Self(Arc::new(TilingInner {
-            size: frame.size(),
+            size,
             frame: LazyHash::new(frame),
-            spacing: spacing.v.map(|l| l.abs),
+            spacing,
+            offset: offset
+                .v
+                .map(|l| l.resolve(styles))
```
saecki: Since font-relative lengths are ruled out, we don't really need the resolve step, but I guess it's still the cleanest way to map the `Rel<Length>` to `Rel<Abs>`.

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -212,10 +253,17 @@ impl Tiling {
...
+        let size = frame.size();
+        let spacing = spacing.v.map(|l| l.abs);
+
         Ok(Self(Arc::new(TilingInner {
-            size: frame.size(),
+            size,
             frame: LazyHash::new(frame),
-            spacing: spacing.v.map(|l| l.abs),
+            spacing,
+            offset: offset
```
saecki: I'd move this into a separate variable to keep things clean.

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -78,6 +79,31 @@ use crate::visualize::RelativeTo;
 /// )
 /// ```
 ///
+/// The [`offset`]($tiling.offset) parameter allows you to offset the starting
+/// position of the tiling. This shifts the entire tile grid without affecting
+/// the tile size or spacing. Positive x values move the pattern to the right,
+/// and positive y values move it down. Relative values are resolved against the tile size plus spacing
```
saecki: Could you wrap this line at column 80?

saecki (commented):
I've got some final remarks, and then I think this should be good to go!

crates/typst-library/src/visualize/tiling.rs
dayala1: Done

crates/typst-library/src/visualize/tiling.rs
dayala1: I will then keep it as is if it is acceptable. I am not too familiar with the language and the framework, so I more or less work by checking similar examples.

crates/typst-library/src/visualize/tiling.rs
dayala1: Done.

crates/typst-library/src/visualize/tiling.rs
dayala1: Done.

dayala1:
> I've got some final remarks, and then I think this should be good to go!

Took care of the comments. Thanks for the review!

saecki (approved):
Thanks!

saecki:
Just realized there weren't any tests for the relative offset.

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -78,6 +79,32 @@ use crate::visualize::RelativeTo;
 /// )
 /// ```
 ///
+/// The [`offset`]($tiling.offset) parameter allows you to offset the starting
+/// position of the tiling. This shifts the entire tile grid without affecting
+/// the tile size or spacing. Positive x values move the pattern to the right,
+/// and positive y values move it down. Relative values are resolved against
+/// the tile size plus spacing
```
laurmaedje:
```suggestion
/// the tile size plus spacing.
```

crates/typst-library/src/visualize/tiling.rs
```diff
@@ -78,6 +79,32 @@ use crate::visualize::RelativeTo;
...
+///   #place(dx: 5pt, dy: 5pt, circle(radius: 5pt, fill: blue))
+/// ]
+///
+/// #grid(
+///   columns: 2,
+///   column-gutter: 10pt,
+///   rect(fill: pat, width: 100pt, height: 80pt, stroke: 1pt),
+///   rect(fill: pat-offset, width: 100pt, height: 80pt, stroke: 1pt),
+/// )
+/// ```
```
laurmaedje, line 110: I feel like the example is not so helpful. It shows that you can offset something, but it doesn't immediately make clear the usefulness over just changing the `dx` / `dy` values in `place`. It would be nice if that could come out more clearly.

On another note, in the tiling docs, it might make sense to move the examples for the parameters to the parameters docs themselves, but since this PR is following the current style, I'd leave that for another PR.

laurmaedje (commented):
Just a few notes on the docs, then this is good to go from my side as well.

crates/typst-library/src/visualize/tiling.rs
dayala1: How about this example? Trying to use dx and dy results in the objects going outside the tiling area:

```typst

#let pat = tiling(size: (20pt, 20pt))[
  #place(circle(radius: 10pt, fill: blue))
]
///
#let pat-offset = tiling(
  size: (20pt, 20pt),
  offset: (50%, 50%),
)[
  #place(circle(radius: 10pt, fill: blue))
]
///
#grid(
  columns: 2,
  column-gutter: 10pt,
  rect(fill: pat, width: 100pt, height: 80pt, stroke: 1pt),
  rect(fill: pat-offset, width: 100pt, height: 80pt, stroke: 1pt),
)
```

RESULT:

image

What happens if dx/dy (10pt) is used in an attempt. Not included in the example, but could be if considered illustrative:

image

laurmaedje: That's much better! In the top-level text, the second example (with dx/dy) is a bit much. If we would move it down into the parameter docs, we could have separate named examples, but I'm also fine with leaving it out. I would, however, mention the difference to dx/dy in prose.

dayala1: Aye. For the moment, did as suggested, and included the example with a mention of the difference with respect to positioning inner contents of the tile.

saecki:
Was the last open comment addressed?
The docs have been converted from markdown to Typst in #8138, so a few more doc changes are needed, and the hashes have to be updated again.
After that I think this can finally be merged :)

dayala1:
> Was the last open comment addressed? The docs have been converted from markdown to Typst in #8138, so a few more doc changes are needed, and the hashes have to be updated again. After that I think this can finally be merged :)

Yep, last one was addressed. Updated the doc, renewed hashes. Thanks for the reviews and attention to detail.

laurmaedje:
Thank you!

PR #7508, opened 2025-12-01
Fix semicolon on integer value  hints

isuffix:
I saw that you got this for the other recent PRs, but you missed this one ;)


laurmaedje:
Thanks :)

PR #7364, opened 2025-11-12
Detailed non-convergence diagnostics for introspections

laurmaedje:
### Description

This PR adds rich diagnostics for non-convergence of the document. Previously, only a single global warning would be emitted. Now, there are individual warnings for each specific introspection that did not converge.

This is implemented by tracing all introspections that occur during compilation (in a manner similar to how warnings are accumulated), re-running them against the documents for all iterations, and checking whether they converged.

Each introspection is now represented by a type that knows how to resolve itself against a document and how to diagnose its own non-convergence. This makes it easy to introduce new introspections with high-quality diagnostics.

Closes https://github.com/typst/typst/issues/4528.

### Design

Previously, when considering the problem of introspection diagnostics, I struggled to find a way to retain spans (pointing to source code locations) for diagnostics without them interfering with memoization and query deduplication (important for efficiency). The crucial insight of this PR is that we can decouple the high-level introspections (which are recorded with spans) from the low-level methods on the introspector (spanless and deduplicated).

Aside from the ability to record spans, this decoupling also lets us produce more high-level diagnostics. For example, while the `QueryIntrospection` maps directly to `introspector.query`, the `StateAtIntrospection` internally uses `introspector.query` to resolve the state's value, but its convergence behavior and its diagnostics are at the level of abstractions of state.

### Examples

Many of these examples use a `switch` function, which invokes a callback with the current layout iteration number. It's definition is as follows:

<details>
<summary>The <code>switch</code> function</summary>

```typ
#let switch(f) = context {
  // The `here()` trick is just to produce a counter unique to this context
  // block.
  let c = counter(metadata.where(value: here()))
  let i = c.final().first()
  let n = i + 1
  c.update(if n < 5 { i + 1 } else { i })
  context f(n)
}
```
</details>

<details open>
<summary>Position</summary>

```typ
#switch(n => v(n * 10pt) + [= Heading])
#context locate(heading).position().y
```
```
warning: document did not converge within five attempts
 = hint: see 1 additional warning for more details
 = hint: see https://typst.app/help/convergence for help

warning: heading position did not stabilize
  ┌─ <stdin>:3:9
  │
2 │ #switch(n => v(n * 10pt) + [= Heading])
  │                             --------- heading was created here
3 │ #context locate(heading).position().y
  │          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  │
  = hint: the following positions were observed:
    - run 1: page 1 at (0pt, 0pt)
[...]

PR #7512, opened 2025-12-02
Remove Nix flake

laurmaedje:
This PR removes the Nix flake from the repository. I'm aware that this will disappoint some people. However, I have (and have had for a long time) the feeling that the flake is not pulling its weight for Typst. Let me elaborate on why I think that:

- The main benefit is reproducibility and a streamlined dev environment. However, Typst's dev environment is already as streamlined as it gets: You basically only need a recent Rust toolchain. We take care to keep Typst simple to build. So basically you're switching out `cargo` for `nix`. But since Typst is a Rust project, `cargo` is the tool more people are likely to have installed anyway.

- The flake requires frequent attention. Looking [at the PR log](https://github.com/typst/typst/pulls?q=is%3Apr+Nix), the Nix flake tends to break from time to time. When there are issues, they are also not always simple to fix (see e.g. https://github.com/typst/typst/pull/7374). There have been efforts to add Nix to CI to avoid this, but this just shifts the problem: The problems don't go away, they only need to be fixed immediately. Moreover, it severely slowed down CI. The flake also needs to be updated alongside the Rust toolchain because everything is pinned. Overall, the maintenance effort is not _that_ high, but also not minimal.

Note that this does not go against Nix as a technology. I think it's cool tech and if Typst was harder to build, I think it would have its place in the Typst repository as well. But in our case, I feel like the flake just introduces unnecessary coupling between Typst and Nix, because the benefits of Nix are not really needed for Typst. In the context of Typst, it feels like a case of a "solution in search of a problem" to me.


alerque:
> * The main benefit is reproducibility and a streamlined dev environment.

For who? Not in my book. The main benefit of the Flake is that I can run any arbitrary version of Typst from anywhere. This makes it much easier to *USE* Typst in other projects. This goes for both CI jobs and local use, because it makes it a lot easier to have Typst *as a tooling dependency*.

I am not saying you need to run a heavy CI job on the regular (although on my own projects with proper CI caching the Nix flake test builds tend to finish before my regular test suite, so that can be looked into) but removing the Flake would fundamentally change how and where people are able to *use* it not just develop on it.

jamesrswift:
I use neither nix nor cargo, but it would seem sensible to remove as many development pinch points while in 0.x.x

PgBiel:
> In the context of Typst, it feels like a case of a "solution in search of a problem" to me.

I think this particular conclusion is dangerous because there are several large contributors using it (such as @tingerrr, @frozolotl, @mkorje).
But I think there is a better conclusion to consider, which is related to @alerque's concern above. Let me provide my view.

The main problem, I think, is about the dev shell, not the flake itself. **The dev shell doesn't build Typst.** It just installs cargo, rust-analyzer, rustc. It's true that it tends to install versions of those tools that are compatible with that of Typst. But it seems to have its own problems separate from Typst, and its effective usage - e.g., while editing the code base - are much more dependent on what the dev has installed on their own machine, and which OS/distro they are using etc. Most importantly, **problems in the dev shell can't be (easily) caught by CI,** as they are often system-dependent, so adding CI wouldn't make a difference.

Most importantly, taking a look at the log, the following PRs were only related to the dev shell: https://github.com/typst/typst/pull/7374 (mentioned in this PR's description),  https://github.com/typst/typst/pull/6845 (actually this one did a little bit more, but the claim of the flake being "huge" seems to not hold without the dev shell), https://github.com/typst/typst/pull/6826, https://github.com/typst/typst/pull/6618, https://github.com/typst/typst/pull/6107

The main benefit **for the Typst project** of having a flake is precisely being able to use it to _compile and use Typst_ in a way that is automatically using the correct toolchain versions to build it, even for arbitrary Typst commits, and with full caching support and other benefits. With an increasing number of servers using tools for increased reproducibility and reduced downtime, such as NixOS itself, and considering how Typst-nix tools appear to have hundreds of users (see https://github.com/loqusion/typix), I think we shouldn't close the door for these users. It's a great tool in Typst's belt at the moment.

In contrast, the dev shell is used only by _contributors_. It's of course nice to help out contributors and I'm not opposed to it. But if the conclusion is that it is taking too much maintenance time, it should be OK to remove it, and all of those PRs could be closed / not merged. But it's essential to note here that the flake has two parts - nix package and dev shell - which have entirely different purposes (even if related).

In other words, the flake is not currently "broken" despite the open PR - just the dev shell isn't producing expected results on certain systems. I think that the only maintenance needed for the flake outside of the dev shell would be to update the rust toolchain from time to time. Other PRs were just formatting and reorganization (https://github.com/typst/typst/pull/6827).

Case in point, the last time the flake had an actual "bug fix" was over a year ago (https://github.com/typst/typst/pull/5056) just to make it work on MacOS too.

>  Moreover, it severely slowed down CI.

It's worth noting that this observation is orthogonal to the proposal of running CI once a week, in a separate runner. Of course, running it together with main CI is indeed gonna make it slower for no benefit as it doesn't cache intermediate compilation steps.

So, in conclusion, I think a more reasonable course of action might be to just remove the dev shell and close PRs (currently just one) related to it.

alerque:
Most of what @PgBiel says is spot on. At least keep the Flake, whatever happens with the dev shell.

Also I think it's worth reiterating that the dev shell isn't even broken. Not only is the Flake working for everyone right now (and deployed in production environments) but the dev shell also works for many existing users of it (myself included). The open PR is addressing an edge case with unusual system configurations, and I don't think the dev shell should be supported on a 100% guaranteed to work bases, only as a best effort from existing consumers and contributors. It's not hurting anything that the dev shell is there and works for most people but doesn't cover some possible system edge cases.

Vortriz:
While I am not a contributor to Typst itself, I am a NixOS user (and inturn a Nix user). I think that the package and devshell both serve an important purpose for any Nix user developing for Typst (there are prominent contributors who use Nix, as stated by @PgBiel) and also for anyone wanting to use Typst main branch version as a dependency in another project (although I am not sure how many such projects exist).

If Nix is making the CI slow, I think it's very much fine to remove it from the CI.

tingerrr:
For what it's worth, I've never been a big fan of how flake maintenance is put on the project authors like this, even if they hardly use it themselves. I'm referring to the primary maintainers employed by the Typst company here, where @PgBiel seems to be the only nix user.

I understand why this is and how this is key to it allowing reproducibility, but it does either place additional burden on maintainers or require them to trust contributor submissions that they can't review fully.

However...

Typst is a large project, not being familiar with some parts of it is normal at this point, even for the original authors. Maintenance of some parts of the codebase (even if it's just devops/packaging stuff) has to be delegated. @PgBiel is not the only trusted contributor and nix user, he may be the only one employed by the Typst Company. But there are others whose judgment can be relied on.

Nix is fairly widely used, I would even say that it is popular enough that removing the flake would actually be similar to removing the docker packaging workflow for it. I'd expect there to be a non-trivial amount of complaints about this. The fact that the flake is on the official repository also means that nix users can trust that these receive proper community reviews and are safe to use. Removing the flake like this may cause nix users to opt for less thoroughly reviewed flakes, posing a supply chain risk for these users.

All in all, I would say see if maintenance can be delegated and if there are caching strategies to perhaps make testing the flake in CI more appealing.

PgBiel:
Yeah I should highlight I definitely don't want to "override" anyone's position here as a team member, and the Nix contributors are much more knowledgeable than me (I'm just a "casual user"). So, to make it clear, **my opinions here are my own.**

I'm mostly defending flakes for their utility in CI and server workflows. I only use Typst personally (not at scale, or in build servers) so I don't find much use for the flake myself. But I see that there could be large-scale Typst users (perhaps paying users) that find legitimate use in using Flakes to more easily integrate Typst into their existing deployment workflow. And the supply-chain part is certainly important. Here, it would be akin to having a trusted Docker deployment workflow.

That's also why the dev shell is more of a bonus in my view as it isn't used for that purpose, and can be maintained in a separate repository without much loss.

But the "maintenance burden," even if small, still exists. However, at the moment, the latest meaningful PR (besides formatting changes) has been over a year ago. So that's pretty light in my view, but I can see how the fact it still requires some delegation might be a bit annoying.

laurmaedje:
Upon reflection, I think why I want to do this is actually twofold: One part is the maintenance overhead, which is real, but also not _that_ large. It _is_ true that maintenance can be delegated (even though I would not want to prioritize time of paid staff for this). 

But perhaps the even bigger reason is that I dislike the coupling on a conceptual level. Leaving the dev shell out of the picture, the main purpose that seems to crystallize in the discussion above is running any version of Typst from anywhere. 

But _anywhere_ is really _anywhere where Nix is installed._ It's already possible to build Typst anywhere where cargo is installed, so this is just shifting the problem by one layer. To me, needing Rust 1.89 for Typst is the same as needing whatever version of Nix is required to build the flake.[^1] And Rust is literally the _only_ thing you need to build Typst. It's that simple. The argument to have a flake would apply to every other Rust tool (e.g. [`ripgrep`](https://github.com/BurntSushi/ripgrep) or [`uv`](https://github.com/astral-sh/uv)) in the same way.

That's, to be honest, my main problem and that's what I mean when I say "solution in search of a problem". Of course, it may add some convenience for people that frequently use Nix, but in the same way we could add conveniences for many other kinds of usage patterns. We could have a dedicated GitHub action and various other conveniences. But we don't because of separation of concerns. Development and packaging are kept separate, with the exception of a minimal common denominator: The pre-built binaries.

There is one further exception, which is the `Dockerfile`. But, to be honest, I dislike this one as well. I believe it would be fairly easy to move out of the repository into a `typst/docker` repository. Similarly, I would also be much happier with a `typst/flake` that is delegated and maintained as a best effort, as it properly addresses separation of concerns and decouples development and packaging. It would also be simple to add a disclaimer to that repo's README stating that it is a best-effort. I know that this has been discussed before, but I'm not sure what the final verdict on the feasability was.

Finally, I want to make a note on the flake being deployed in production environments or by paying users. This is something I would precisely _not want_ with the flake as-is, as it is not tested and only maintained as a best effort. It's not a guarantee I would want to give.

[^1]: Probably the minimum required Nix version does not jump as often as the MSRV, but still.

Andrew15-5:
> And Rust is literally the _only_ thing you need to build Typst. It's that simple.

Nuh-uh, the notorious OpenSSL dependency when not using the vendored feature was always a problem with any Rust project that had this. I already forgot all the details, but it's easily fixable if you enable the feature. I guess it probably is also handled in the flake. It's a bit more involved in NixOS (basically with one extra env var set?), but the issue happens on any Linux distribution that doesn't have the dynamic library installed (I think it's usually `openssl-dev`?).

That's my only problem usually. Other than that, `cargo build` is all you need. Even though I use NixOS btw, I haven't set up any of the Nix-based Rust toolchain, so I'm using `rustup` from Nixpkgs and run/update toolchains (nightly, stable, etc.) with `rustup`.

breezykermo:
For what it's worth, I am also a NixOS/Nix user who depends on the flake to build the CLI regularly on my systems. I do, however, agree with the points made by @laurmaedje that if Nix (and/or Docker) are not officially supported by the core team, then they should be moved to a different repository that explicitly states it is best-effort or community maintained.

As Nix users are likely to also want the devShell (I use both the derivation and the devShell for development), I personally would prefer for it to be moved as-is to a separate repo, rather than just keeping the derivation in the main typst repo. It is true that this will mean that there is a burden on the (Nix) user to search for the right commit to build older versions of typst. But if the core team is not maintaining the Nix flake explicitly and/or it is removed from CI, it will already be the case that there is no guarantee you can build an older version of Typst via the Nix flake in a particular commit.

alerque:
What @breezykermo suggests is a major regression and should only be a last resort. Please @laurmaedje give this some consideration time. It sounds like you're frustrated with something because you don't know how it works. Most of your objections don't really cover the nature of the thing you're working with. Having a flake available for an app is not even in the same league as having a rust toolchain and the ability to checkout and build whatever.

Moving this to an external repository is a *huge* detriment to the usability of the Flake. It makes it nearly impossible to do things like test feature branches. Right now testing a PR is as easy as `nix run github:sile-typesetter/refs/pull/<PR_NO/head`. No installation, no checking for what toolchain is required by that branch, no messing up whatever version is already installed with `cargo install`, no changing the working directory contents or adding worktrees, just one command to *run* any version. I can run anybody's fork of Typst at and commit or branch or PR with just one invocation. Test driving various proposed features and trying them on production projects is so much easier with a Flake directly linked to the Git source tree! If you force an external 3rd party repo to host the Flake it will be virtually impossible to sync up with *any* arbitrary commit.

There is more too, but I don't have time to get into the discussion points right now. I would ask that this be left to sit around a bit before yanking out a working solution people already rely on.

oliverpool:
I am sorry that #7374 triggered this feeling.

For the record I would probably have not tested the #7364 pr without the flake (even thought I had a small hiccup, it made it quite easy for a non-rustacean like me to compile and test it to give feedback on my usecase).

Suggestion: add a prominent disclaimer at the start of flake.nix that this isn’t officially supported (and mark issues/pr related to this as “community-supported” or something).

thuiop:
As a Nix user, the flake really made it very simple to work with, so I would be sad to see it go. I am however sympathetic to the maintenance burden it adds.

drupol:
Hey @laurmaedje,

I fully understand the maintenance burden this introduces and I respect the desire to decouple packaging from the core development.

However, I believe there is a slight misunderstanding in the comparison: "_But anywhere is really anywhere where Nix is installed. It's already possible to build Typst anywhere where cargo is installed, so this is just shifting the problem by one layer._": This is not entirely the case.

While Cargo handles crate dependencies beautifully, it still relies on the host system's state (_system libraries like OpenSSL, linker versions, libc, external tools_). Nix doesn't just "_shift_" the dependency, it hermetically seals the build environment. It ensures that if it builds on one machine with Nix, it will mathematically guaranteed build on any machine with Nix, regardless of the underlying OS state. It effectively solves the "_works on my machine_" problem in a way **Cargo cannot do alone**.

That being said, looking at the friction this has caused in the project history and the fact that it seems to be slowing down the core team's workflow, I agree that removing it is likely the best course of action. If the flake is becoming a distraction rather than a utility for the maintainers, it makes sense to let the community maintain it downstream.

alerque:
> the friction this has caused in the project history and the fact that it seems to be slowing down the core team's workflow

What burden? How is it slowing anything down? @PgBiel dived into some of the history above and I think it's pretty clear it just works the vast majority of the time, and most of the issues that occasionally come up are not even about the flake they are about the dev shell. Either way neither of the components is in a position to break or hold back the main app through other use cases. At most it would be broken for Nix Flake users, and we're in a position to submit fixes if that part breaks.

It's already been discussed that dropping (or running on a schedule instead of on PRs) the CI job is just fine if that's what is considered the slow down.

There is no reason for the Flake to "slow down the team's workflow". Just leave it here and change the `on:` trigger for the CI job.

laurmaedje:
I've looked a bit more into this and to me it appears that an appropriately engineered solution with an external `typst/flake` could work quite well, capturing the main benefits of the internal flake cited above.

The invocation would be slightly longer:

```bash
nix run github:typst/flake --override-input typst github:typst/typst/abc123
```

But the main concern, if I understand correctly, would be that this would not guarantee a clean build for any commit. However, I think this can be achieved through the flake adjusting its derivation based on the contents of the `typst` input. For instance, the Rust toolchain could be selected based on the `rust-version` in `Cargo.toml`. And decisions like "do we need OpenSSL" could be made based on checks like "is the Typst version >= 0.12 and is the `native-tls` dependency at this particular version in `Cargo.lock`". Basically any change in Typst that needs a change to the build should have some visible effect in the repo (typically in a TOML file) that should be fairly easy to observe and react to from within Nix.

I think this could, in some ways, be even more robust that then current setup, because:

- If the current flake is broken for a commit (totally possible since no CI), it _stays_ broken for that commit, violating the "can build any commit" assumption. With a setup as above, this could be fixed in the flake after the fact.

- The separate flake could have CI both for the main branch and regular CI for checking whether it's still compatible with the latest Typst main branch. If it isn't, it can be fixed retroactively, making all commits on the main branch work correctly. Commits on PRs would typically just work based on the content-based derivation, and if they don't, it's likely that the current flake wouldn't work either since the flake likely wouldn't have been updated atomically. Since there seems to be a bit of misunderstanding regarding CI in this thread, I'd like to clarify that **the current Nix flake is 100% untested. There is zero CI for it.**

Of course, this introduces more complexity to the flake. But it removes complexity from the main project. I think it could be a worthwhile trade-off. It allows us to fully externalize maintenance to trusted community members and gives the flake room to breath, grow, be tested and developed independently of the limitations imposed on it by being in the main repo (e.g. no CI).

To be honest, the only reason the flake exists in the main repo in the first place is because I didn't yet know how to say "No" when it was PRed just days after our open source release. If it hadn't been added then, I'm fairly certain I wouldn't merge a PR for it now, and I would be [in](https://github.com/gleam-lang/gleam/pull/1405) [good](https://github.com/ziglang/zig/pull/12283) [company](https://github.com/alacritty/alacritty/pull/8662).

drupol:
I once again support the decision **to remove the Nix flake from this repository**. If the core team does not possess the time/expertise/will/whatever to maintain it, or simply does not wish to allocate resources to it, it is far better to remove it than to host a broken or unverified implementation. An external, community-maintained `typst/flake` is likely the best path forward in this context.

I would also like to (sadly) highlight that Nix flakes are technically still **an experimental feature** of the Nix project. It is perfectly reasonable that a core team focused on Rust and Typst would not want to shoulder the burden of tracking an experimental feature of an external build system. (even if that build system is universal and provides the unique benefit of working identically across all Linux distributions and macOS!)

However, I feel compelled to address the reasoning used to justify this separation beyond the maintenance aspect, specifically regarding the "utility" of Nix and the comparisons made to other projects.

Dismissing a technology or labelling it a "solution in search of a problem" simply due to a lack of familiarity is a dangerous mindset for a large open-source project. While `cargo` is great, it does not solve the problem of environmental reproducibility (e.g., system libraries like OpenSSL, `libc` versions, linker variations, etc.). Nix solves this not just for the Rust toolchain, but for the entire operating system stack. The "problem" exists, even if it is not currently causing friction on the core team's specific machines.

Regarding the links provided to show you are in "good company":

I reviewed those discussions. In almost every instance, the rejection of Nix was not based on technical demerit, but rather on the same "maintenance burden" argument and/or a philosophical desire to enforce the language's own build system as the source of truth. While it is valid to prioritise maintainer focus, citing these examples essentially highlights a trend where strict reproducibility is sacrificed for maintainer comfort.

Being in "good company" with projects that choose to limit their official build scope is fine, but it shouldn't be confused with Nix being technically redundant. It just means those projects, like Typst, have decided that strict downstream reproducibility is not a core project responsibility.

Perhaps it would have been more accurate to simply state that Nix development builds (snapshots) is not a priority for Typst, nor is providing a working development environment; that would have been a clearer stance.

Moving this to `typst/flake` is the right move. Not because Nix is unnecessary, but because it requires maintainers who understand *why* it is necessary.

laurmaedje:
> Dismissing a technology or labelling it a "solution in search of a problem" simply due to a lack of familiarity is a dangerous mindset for a large open-source project. 

I really didn't intend to take a stab at Nix. Please disregard the "solution in search of a problem" bit.

> While it is valid to prioritise maintainer focus, citing these examples essentially highlights a trend where strict reproducibility is sacrificed for maintainer comfort.

I wouldn't say "maintainer comfort" is the right mindset either. It's simply that not every project needs to make use of every cool piece of tech, even if it has some benefit. I think it's okay for a project to say "this is out of scope" to some new tech _even if_ there are technical benefits to be had. The word "comfort" does not quite capture this.

That's what these other maintainers are doing and I find it perfectly acceptable. I would do the same now if a PR was opened now. Since a flake already existed, there is more pain in undoing this decision, which is why I'm trying to put effort into find a good compromise in my previous messages. But it's the same underlying reason and it's not just about the sum of individual minutes spent looking at Nix PRs or bumping Nix alongside Rust.

drupol:
Thank you for the clarification regarding the earlier comment, I really appreciate that.

However, I must insist on one nuance regarding the "_cool piece of tech_" phrasing. Nix is not just "cool" or "new" technology in the sense of a trendy feature add-on, or simply nice to see in an open-source repo. It is currently the only widely adopted tool (among very few others) capable of strictly solving the problem of hermetic environmental reproducibility. I personally feel it is somewhat dismissive to label it merely as a "cool piece of tech".

> I think it's okay for a project to say "this is out of scope"

Do you really believe that the ability to **reliably** build Typst from source is something that should be considered "out of scope"? Because this is precisely one of the main reason why we use Nix in this repo.

Anyway, we can agree to disagree and I think moving the Nix configuration to an external repository is indeed the correct technical decision. However, I must warn you of a specific technical regression this will introduce.

Moving the Nix files to a separate repository will significantly hinder the ability to create **snapshot builds**. Since the source code will no longer be co-located with the Nix files, the Typst source becomes a "flake input". In the context of Nix, every input change requires a corresponding hash verification. This makes generating builds for arbitrary commits or PRs much harder than the current setup, where the flake resides inside the repository and implicitly uses the local source without requiring external input locking or hash updates.

Creating the development environment (`devShell`) shouldn't be an issue, but the build workflow for snapshots will suffer significantly.

**Edit**: building the a snapshot could be achieved by doing: `nix-build -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixpkgs-unstable.tar.gz --expr 'let src = builtins.fetchTarball "https://github.com/typst/typst/archive/master.tar.gz"; pkgs = import <nixpkgs> {}; in pkgs.typst.overrideAttrs (self: { inherit src; cargoDeps = pkgs.rustPlatform.fetchCargoVendor { inherit src; hash = "<replace-me>"; }; doInstallCheck = false; })'`

laurmaedje:
In the end, I just dislike having something in the repo that's not tested and not officially supported. It's just not the Typst way. At the same time, even though I understand and appreciate the benefits of Nix, I don't want to officially support it either. I think the external flake is a reasonable compromise and I want to move forward with that. 

I've created a [typst/typst-flake](https://github.com/typst/typst-flake) repository for the new external flake. @PgBiel is already added to it and if trusted community members are interested in helping out with setting it up and maintaining it, I'd be happy to invite them. I'll wait a bit before merging this PR, so that the new flake can be in place before the old one is removed.

We'll probably proceed similarly with a `typst/typst-docker` repository.

drupol:
> In the end, I just dislike having something in the repo that's not tested and not officially supported.

If the lack of testing was the primary concern, we could have simply addressed that. Adding a GitHub workflow to build `typst` via Nix is trivial and would have legitimized the setup immediately.

In fact, such a workflow would have caught the recent build issue in release mode much earlier (see: https://discord.com/channels/1054443721975922748/1445830066259890289/1446281031605817536), even before releasing 0.14.1.

However, I won't debate this further as you have already created the new repository. Let's focus our energy there to make the best of it.

laurmaedje:
You're right that debating this further doesn't make much sense, but I'd like to clarify that the recent issue with the test suite in nixpkgs is really completely unrelated and it doesn't make sense to bring it up here. It's an issue with _running_ (not building) the test suite that happens _in release mode only._ If the normal GitHub action would run in release, it would catch it. Same with Nix: If the hypothetical Nix flake CI would've run the test suite, and testing would've happened in release mode, it would've caught it, but in debug mode it would not have. But probably it would _not_ have run in release mode, since in the brief time we did have it a longer while ago, it was already by far the slowest part of CI and release builds are much slower than debug builds.

Issue #7518, opened 2025-12-02
Crash on 0-byte attached PDF file


Andrew15-5:
### Description

```typ
// #pdf.embed("file")
#pdf.embed("file", bytes(""))
```

```
thread 'main' (1769815) panicked at /build/typst-0.14.0-vendor/krilla-0.5.0/src/stream.rs:240:12:
attempt to divide by zero
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This works in v0.13.1, since it used `miniz_oxide::deflate::compress_to_vec_zlib` instead of krilla for file compression (and it didn't divide by zero or handled it correctly).

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

Andrew15-5:
Oops, the auto-closing... But https://github.com/typst/typst/pull/7519 should actually fix this.

PR #7519, opened 2025-12-02
Bump krilla

Andrew15-5:
Fixes https://github.com/typst/typst/issues/7518.

Bring fix from https://github.com/LaurenzV/krilla/pull/319.

I tried

```toml
krilla = { git = "https://github.com/LaurenzV/krilla", rev = "84c6332", default-features = false, features = ["raster-images", "comemo", "rayon", "pdf"] }
```

But it creates 2 versions of the dependency in the lock file and causes errors. With patch it compiles cleanly.

@LaurenzV, will you create a new release just for this or nah?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1439,8 +1439,7 @@ dependencies = [
 [[package]]
 name = "krilla"
 version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0ddfec86fec13d068075e14f22a7e217c281f3ed69ddcb427bf3f5d504fd674"
+source = "git+https://github.com/LaurenzV/krilla?rev=84c6332#84c6332a4d6459427b02bb40c8eb7e46b6a173d1"
 dependencies = [
  "base64",
  "bumpalo",
@@ -1471,8 +1470,7 @@ dependencies = [
 [[package]]
 name = "krilla-svg"
 version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f485e1a850201a01dcd8d73e7cf09f2cd4c4cc85c2cd296359094d49336d8ef7"
+source = "git+https://github.com/LaurenzV/krilla?rev=84c6332#84c6332a4d6459427b02bb40c8eb7e46b6a173d1"
 dependencies = [
  "flate2",
  "fontdb",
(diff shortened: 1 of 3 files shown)


LaurenzV:
I feel like this is minor enough that it can wait until the next release.

laurmaedje:
> But it creates 2 versions of the dependency in the lock file and causes errors. With patch it compiles cleanly.

You need to change the version of `krilla-svg`, too, then it should work.

With a `[patch]`, I can't merge it because the `typst` crate would become non-functional as a dependency. It only affects the workspace.

Andrew15-5:
> You need to change the version of `krilla-svg`, too, then it should work.

Indeed. However, I then don't get why patching `krilla` without `krilla-svg` also just works? Like, I'm completely confused by this behavior between 2 linked packages and patch/git. And I thought that I probably understood how it all works...

saecki:
> Indeed. However, I then don't get why patching `krilla` without `krilla-svg` also just works? Like, I'm completely confused by this behavior between 2 linked packages and patch/git. And I thought that I probably understood how it all works...

The purpose of `[patch]` is to also override transitive dependencies, this is by design.

laurmaedje:
Thanks!

Andrew15-5:
I don't think "bump" PRs should create new tests.

Andrew15-5:
> The purpose of `[patch]` is to also override transitive dependencies, this is by design.

Huh? So it also automatically patches `krilla-svg` to the same git revision, even though I didn't ask it to? That's..kinda complex.

laurmaedje:
> I don't think "bump" PRs should create new tests.

I don't see a problem with that.

saecki:
> > The purpose of `[patch]` is to also override transitive dependencies, this is by design.
> 
> Huh? So it also automatically patches `krilla-svg` to the same git revision, even though I didn't ask it to? That's..kinda complex.

No it patches the `krilla` dependency of `krilla-svg`

Andrew15-5:
> I don't see a problem with that.

The description of fixing a bug by (updating a fixed dependency + a new test for that) makes total sense.
The description of updating a dependency that also for some reason includes a new test is weird. Version update is just a version update, unless it breaks something, and you have to update the code. But it doesn't break anything.

laurmaedje:
Package bumps frequently have varied effects, from fixing one bug to fixing multiple bugs to adding new features or changing behavior. I consistently name them "Bump x" to make it a clear that a version bump happened and various changes may result from that. Meanwhile the "Fix x" naming I reserve for the actual fix (in this case in krilla). This is a simple and consistent rule, which is why I also apply it for a bump whose only effect is to bring in a single bug fix.

In my opinion, it makes sense to add a test in the moment these changes hit typst/typst. If many bugs are fixed the moment a bump happens, I'd also add various different tests.

Andrew15-5:
Thanks for the explanations.

Issue #7475, opened 2025-11-26
Over/underlines extend to CJK-Latin spacing unexpectedly


YDX-2147483647:
## Description

If an [`overline`](https://typst.app/docs/reference/text/overline/) or [`underline`](https://typst.app/docs/reference/text/underline/) starts or ends with a Han character, and it's adjacent to a Latin letter, then the line will extend to cover the CJK-Latin spacing, which is unwanted.

May relate to #1716.

Workaround: putting an explicit space.

```typst
#set text(font: "Noto Sans CJK SC")


- #underline[五日]M // 😓
- #underline[五日] M // ✅
- #underline[五日] // ✅
```

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/d82c7bab-897c-4105-9cb9-14d69e110716" />

### More examples

```typst
#set text(font: "Noto Sans CJK SC")

- M#underline[五日]M
- M #underline[五日] M
- 日#underline[Mn]日
- 日 #underline[Mn] 日
```

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/8e77689c-d8c7-4a54-abf5-3286c50630db" />

### Cause

Runge commented on 2025-11-26 (translated from Chinese):

> This is basically because `add_cjk_latin_spacing` does not actually insert spacing. Instead, it adds padding to the Han characters by setting its `x_advance` and `offset`.

### Desired result

Translating from Chinese in https://github.com/w3c/clreq/issues/716#issuecomment-3578811704:

> The underline is by default of **the same length as the character frame**. When two underlines are adjacent, the length of the adjacent side is shortened to visually separate and distinguish them, while the length of the other side remains unchanged.

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

---

> This issue was originally reported by 差 on 2025-11-26.

RungeCC:
More explanation on this:

> This is basically because add_cjk_latin_spacing does not actually insert spacing. Instead, it adds padding to the Han characters by setting its x_advance and offset.

Typst (in `typst_layout`) calls `decorate` to add decoration elements such as underline _after_ shaping frame is built, in which step new `Glyph` is created whose `x_advance` and `x_offset` are based on shaped `ShapedGlyph`, already adjusted in `preparation`. During `preparation`, cjk-spacings are inserted as `Glyph` `offset` and/or `x_advance`, so `width` will contain them, and during iteration of `decorate`, `glyph` metrics will also contain them.

PR #7521, opened 2025-12-03
Fix unwanted CJK-Latin spacing in text decorations

Ri-Nai:
**Description:**
Related to #7475 where underlines and highlights extend into the automatic spacing added between CJK and Latin characters.

**Implementation Details:**
Instead of modifying the underlying shaping logic (which inserts the spacing directly into glyph advances/offsets), this PR adjusts the decoration rendering in `deco.rs`.

It introduces `adjust_cjk_latin_spacing` which heuristicsally detects if spacing was applied to the first/last glyphs of a run. It checks if the glyph is a CJK script and if the x-offset or advance includes the specific `0.25em` padding used by the shaper.

It's just a **temporary** patch. While refactoring the CJK-Latin spacing implementation in the shaper might be the "pure" fix, it is a significant architectural change with high regression risks.

### DIFF
--- a/crates/typst-layout/src/inline/deco.rs
+++ b/crates/typst-layout/src/inline/deco.rs
@@ -182,24 +182,24 @@ fn adjust_cjk_latin_spacing(width: Abs, text: &TextItem) -> (Abs, Abs) {
     let mut leading = Abs::zero();
     let mut trailing = Abs::zero();
 
-    if let Some(first) = text.glyphs.first() {
-        if glyph_primary_char(text, first).is_some_and(is_of_cj_script) {
-            let offset = first.x_offset.at(text.size);
-            if offset + tolerance >= spacing {
-                leading = spacing.min(offset);
-            }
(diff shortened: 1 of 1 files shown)


laurmaedje:
> It's just a temporary patch. While refactoring the CJK-Latin spacing implementation in the shaper might be the "pure" fix, it is a significant architectural change with high regression risks.

I must admit that I'm concerned with the maintainability of the inline layouter. It has accumulated quite a bit of somewhat hacky code, in particular for CJK handling, which makes it quite hard to cleanly refactor (especially without good understanding of CJK languages since I can barely judge whether something is an improvement or a regression).

laurmaedje:
Thanks for working on this, but I'll close it as I think this patch is too hacky and introduces too much maintainability burden. What it really shows me is that the CJK-Latin spacing should not be considered part of the glyph advance in the first place, but rather be handles as a separate spacing item or similar, that's inserted.

PR #7523, opened 2025-12-03
Add `block: true` in docs example for `read` function

ecstrema:
A friend was learning about typst, and wondering why his code looked like this:

image

Turns out he had followed the editor hover for read:

image

Since this is importing an HTML file, it is very probably more than a single line (plus, `read`ing a file for a single line would be a little overkill I believe).

Adding `block: true` fixed it:

image


laurmaedje:
Thanks!

PR #7528, opened 2025-12-04
Friendly hint and docs clarification for `array.sorted`

cady-b:
## Background

`0.14.1` (#7520) changed the algorithm used by `array.sorted`; from `slice::sort_by` ([`driftsort`](https://github.com/Voultapher/driftsort)) to [`glidesort`](https://github.com/orlp/glidesort). This slightly alters which elements are actually compared[^ds-gs], which causes inputs that previously passed silently despite not technically having been valid to now error more frequently (but still not always).

Previous discussion on Discord: [Typst › contributors 🗪](https://discord.com/channels/1054443721975922748/1088371867913572452/1445920024157098159)

[^ds-gs]: https://github.com/Voultapher/sort-research-rs/blob/main/writeup/driftsort_introduction/text.md#components-from-glidesort-which-are-not-incorporated

## Samples

| input                                                               | 0.14.0             | 0.14.1             |
| ------------------------------------------------------------------- | ------------------ | ------------------ |
| `((1, []), (1, []))`                                                | :x:                | :x:                |
| `((1, []), (2, []))`                                                | :white_check_mark: | :white_check_mark: |
| `range(1,8).map(v => (v, []))`                                      | :white_check_mark: | :white_check_mark: |
| `range(1,9).map(v => (v, []))`                                      | :white_check_mark: | :x:                |
| `((1, 1), (2, 1), (3, 1), (4, []), (5, 1), (6, 1), (7, 1), (8, 1))` | :white_check_mark: | :white_check_mark: |
| `((1, 1), (2, 1), (3, 1), (4, 1), (5, []), (6, 1), (7, 1), (8, 1))` | :white_check_mark: | :x:                |

## New Hint

Especially the volatility makes debugging this issue in one's document difficult. To counteract the source of confusion, I propose a new hint be added to the error generated in this code path:

```
error: cannot compare content and content
  ┌─ a.typ:1:1
  │
1 │ #range(1,9).map(v => (v, [])).sorted()
  │  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  │
  = hint: consider choosing a `key` or defining the comparison with `by`
```

It's not too noisy if the user knows something about the input array was wrong (see the changes for existing tests), but helps those determined to sort the given input find their way to a solution. We present keywords which help experienced users recall their options and allow newer ones to look up the most relevant information in the documentation.

The hint isn't applied to the branch where `by` is given because at that point we can't do much; the user has to figure out how to solve the error case. Same for an error within `key`.

Personally, I really don't like that invalid inputs _sometimes_ work, but also see that checking this more extensively would be pretty difficult. Maybe something to consider for the future alongside https://github.com/typst/typst/issues/6285#issuecomment-3523601099.

## Doc Comment change

Not much to say here. If you believe it doesn't matter as much as I feel rn but like the hint, I'll gladly drop this part.

### DIFF
--- a/crates/typst-library/src/foundations/array.rs
+++ b/crates/typst-library/src/foundations/array.rs
@@ -981,10 +981,18 @@ impl Array {
                     match (key_of(a.clone()), key_of(b.clone())) {
                         (Ok(a), Ok(b)) => ops::compare(&a, &b).unwrap_or_else(|err| {
                             if result.is_ok() {
-                                result = Err(HintedString::from(err).with_hint(match key {
-                                    None => "consider choosing a `key` or defining the comparison with `by`",
-                                    Some(_) => "consider defining the comparison with `by` or choosing a different `key`"
-                                })).at(span);
+                                result =
+                                    Err(HintedString::from(err).with_hint(match key {
+                                        None => {
+                                            "consider choosing a `key` \
+                                             or defining the comparison with `by`"
+                                        }
+                                        Some(_) => {
+                                            "consider defining the comparison with `by` \
+                                             or choosing a different `key`"
+                                        }
+                                    }))
+                                    .at(span);
                             }
                             Ordering::Equal
                         }),


laurmaedje:
Thanks! This is nice. :)

PR #7529, opened 2025-12-04
Update numbering.rs to support Armenian letter numbering

ditsak:
Added support for upper and lower case Armenian numbering(ա, բ, գ / Ա, Բ, Գ)


crates/typst-library/src/model/numbering.rs
```diff
@@ -522,6 +530,90 @@ impl NumberingKind {
...
+                    ("ա", 1),
+                    ("0", 0),
+                ],
+                n,
+            ),
+            Self::UpperArmenian => additive(
+                &[
+                    ("Ք", 9000),
+                    ("Փ", 8000),
+                    ("Վ", 7000),
```
laurmaedje: The codepoint used for 7000 is the same one as used for 3000. Should it be Ւ instead? Same for lowercase.

laurmaedje:
Could you please also add a test in `tests/suite/model/numbering.typ`?

crates/typst-library/src/model/numbering.rs
ditsak: Oh, yes, for 7000 it should be Ւ, sorry for the typo.
As for the 10.000 and 20.000, those better not be included since those weren't ever really used, and those two aren't part of the original alphabet. I'll try to fix it right now

laurmaedje:
Thanks!

PR #7530, opened 2025-12-05
Fix wording in `grid.footer` docs

ultimatile:
In [Grid Footer](https://typst.app/docs/reference/layout/grid/#definitions-footer) document, it currently states:

> Just like the [grid.header](https://typst.app/docs/reference/layout/grid/#definitions-header) element, the footer can repeat itself on every page of the table.

It would be more precise to say:

> Just like the [grid.header](https://typst.app/docs/reference/layout/grid/#definitions-header) element, the footer can repeat itself on every page of the grid.


laurmaedje:
Thanks!

PR #7535, opened 2025-12-06
Add `body-indent` property to headings

Dawnfz-Lenfeng:
This PR adds the `body-indent` property to the heading element, allowing users to customize the spacing between the numbering and the heading body.

Currently, `list` and `enum` both support the `body-indent` parameter to adjust the space between the marker/number and the content. However, heading lacked this property and relied on a **hardcoded** spacing of 0.3em.

Previously, to adjust this spacing (e.g., to increase it to 1em), users had to write a complex show rule that manually reconstructed the heading:

```typst
#show heading: it => block({
  if it.numbering != none {
    counter(heading).display(it.numbering)
    h(1em)
  }
  it.body
})
```
With this change, users can now achieve the same result simply and intuitively, consistent with other list-like elements:

```typst
#set heading(body-indent: 1em)
```

The default value is set to 0.3em to preserve existing behavior and ensure no breaking visual changes for users relying on the default spacing.


laurmaedje:
Thanks for the PR, but I'll close this for the same reason as https://github.com/typst/typst/pull/5583#issuecomment-2545401437.

TLDR: I want to implement a more general solution and by now also have somewhat concrete ideas, which I'll need to write down at some point.

PR #7538, opened 2025-12-06
Detect Hebrew Geresh/Gershayim in smartquotes

### LINKED ISSUE #6596: Smartquotes incorrectly apply to characters intended as Hebrew Geresh/Gershayim
### Description

Relates to previous discussion in https://github.com/typst/typst/pull/4842 and https://github.com/typst/typst/issues/3332.

As discussed before, Hebrew has both the single and double "merkhaʾot" (the good ol' quotes, as used in many other languages) and the Gereh and Gershayim characters (used for enumeration, abbreviations, as a diacritic, etc). The latter two are absolutely not quotes - i.e. they aren't paired - but due to the twists and turns of history are typed on Hebrew keyboards using the exactly same keys; that is, `'` for Geresh and `"` for Gershayim.

In a sentence that includes characters intended to be used as Geresh/Gershayim, the smartquotes features interferes with them and leads to incorrect results. As a fairly involved example, let's look at:

```typ
#set text(lang: "he")
⁧הוא אמר: "אתמול נסעתי לנתב"ג עם סרג'".⁩
```

If we look at the content emitted by current Typst main branch for this, the quote characters are replaced with (right to left, of course):

1. `U+201C Left Double Quotation Mark` (OK)
2.  `U+201D Right Double Quotation Mark` (WRONG! this is a Gershayim, and shouldn't have matched with 1)
3. `U+2019 Right Single Quotation Mark` (WRONG! this is a Geresh)
4. `U+201C Left Double Quotation Mark` (WRONG! this is the correct pair of 1, and should have been a U+201D)

Even without interfering with actual quotes, this causes problems as the the wrong glyph may be selected and this interferes with anything that will try to use the standard boundary analysis algorithm on the tagged content (leading to wrong word boundaries, word counts, etc).

I think that the rule that can be used to distinguish them is the one laid out in [UAX #29](https://www.unicode.org/reports/tr29):
- When a single quote directly follows a Hebrew letter (as the UAX defines it), it is a Geresh (rule WB7a).
- When a double quote is sandwiched between two Hebrew letters, it is a Gershayim (rules WB7b and WB7c).

I think that perhaps the `SmartQuoter` struct needs to bake-in this logic (as it does for the prime heuristic), and ideally emit the `U+05F3` / `U+05F4` characters for Geresh and Gershayim - or at very least, leave them alone.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

IgKh:
Fixes #6596

As described in the linked issue, with a typical Hebrew keyboard layout the Apostrophe (U+0027) and Quotation Mark (U+0022) characters are used to write the "regular" quotes, but also the Hebrew-specific Geresh (U+05F3) and Gershayim (U+05F4). While transforming the characters intended as quotes into the proper curly quote codepoints is good, doing so for the characters intended to be a Geresh or a Gershayim is very problematic.

This PR adds a heuristic to the `SmartQuoter` type that attempts to distinguish between the two cases.

My original intent was to use the same heuristic that is used as part of the Unicode Text Segmentation algorithm (UAX 29) - any instance of `'` and `"` which would not be a word boundary as per the algorithm is a Geresh/Gershayim. While it is a simple heuristic with no obvious false negatives, it is unfortunately not appropriate due to a very blatant false positive. 

According to Academy rules [[1]](https://hebrew-academy.org.il/topic/hahlatot/punctuation/#target-3465), prefix letters that are part of the "Moses and Caleb" mnemonic [[2]](https://www.hebrewversity.com/prefixes-in-hebrew-moses-and-caleb/) stay outside and immediately before the opening quote, tripping the simple heuristic. There is even an example of this in the existing test case - `on the 'telephone.'` in Hebrew would be `⁧ב'טלפון'.⁩`

Due to the recursive composability of prefixes in Hebrew, detecting and handling this case directly would require variable amount of lookback text, which is quite problematic considering the `SmartQuoter` call sites. Therefore I attempted a more complex heuristic that tries to take into consideration the actual use cases of Geresh and Gershayim in the language.

#### Geresh
For the use of Geresh as a punctuation mark (as used for abbreviation and counters [[3]](https://hebrew-academy.org.il/topic/hahlatot/punctuation/#target-3469)), Academy rules are that it comes at the end of the word but before any further punctuation. Therefore the heuristic is that any single smartquote which is immediately after a Hebrew letter but not immediately before any other letter (and is not closing an opening single quote) is a Geresh.

For the use of Geresh as a diacritic (as used in the transliteration of phonemes that are not normally part of the Hebrew phonology), any appearance of a single smartquote immediately after a Hebrew letter such that the pair appears in the standard transliteration table (Table A in pages 3 and 4 of [[4]](https://hebrew-academy.org.il/wp-content/uploads/taatik-latinit-ivrit-2025-1.pdf)) is a Geresh. This is done even if at the end of the word while the top quote in the stack is a single quote as I believe this is a very strong convention in modern Hebrew writing, to the extent that several Hebrew fonts have a ligature for these pairings.

**NOTE**: There is a false negative here for the use of Geresh as a diacritic in Yiddish (and in Hebrew loan-words originating from Yiddish), as without full lookback there is a conflict with the "Moses and Caleb" rule. It yields the wrong result on example (1) in [1].

#### Gershayim
Gershayim are only used as punctuation, for acronyms. The heuristic at the moment is just that every double smartquote sandwiched between two Hebrew letters is Gershayim.

This is rather problematic since it suffers from the same false positive described above, though it is probably less common to run into the issue with double quotes compared to single quotes. The Academy rules are that Gershayim as punctuation should only appear just before the last letter in the acronym, ~~but testing for this requires a two-character lookahead instead of just one. It is easy to provide it, but makes for a messy code. If it is OK - I'll change to take `after` as `Option<&str>` instead of `Option<char>`.~~ (Not actually possible/reasonable, due to arbitrary amount of suffixes that can be added after the acronym)


laurmaedje:
Thanks for the PR and the comprehensive explanation!

https://github.com/typst/typst/pull/4849 got rid of lookahead because it's quite hard to do correctly. Lookbehind is not _that_  problematic, because the content before where we currently are is already processed. But naive lookahead as implemented here is not really robust because it does not consider the varied ways in which the following content could end up being an adjacent character without it looking like it: For instance, due to being in a `<span>` in HTML , invisible introspection tags in between, default ignorables, etc. 

This in combination with having false positives/negatives in the Hebrew logic itself seems quite problematic to me.

To be fair, other smartquote issues (https://github.com/typst/typst/issues/7615) might force us to abandon the nice no-lookahead-approach that's currently implemented after all, but it's complicated and the lookahead as implemented here is something I would rather not reintroduce.

IgKh:
Thanks for the review @laurmaedje, I understand your position.

I'm not sure where it leaves this PR though. There isn't much point keeping it open if it pends on a yet unspecified future overhaul of the smart quoter, and I'd assume this would be one of those things where you'd prefer the design to happen in-house rather than in the community.

For the record, as input for that future design - I think that the Hebrew-specific logic can be done without false positives/negatives with one letter of lookahead and variable lookback (until the preceding word boundary).

With that said, I do still feel that applying smartquotes on Hebrew in the current state does more harm than good, particularly with the focus on accessibility. I wonder if an alternative PR to make smartquotes off by default for Hebrew is worthwhile. While disappointing, especially given work of previous contributors, it will put Typst in the same company as Microsoft Word which doesn't apply its' smart quotes to Hebrew text (despite otherwise being the gold standard on Hebrew word processing).

laurmaedje:
> I'm not sure where it leaves this PR though. There isn't much point keeping it open if it pends on a yet unspecified future overhaul of the smart quoter, and I'd assume this would be one of those things where you'd prefer the design to happen in-house rather than in the community.

Yeah, I suppose you are right. Let's close it then. Thank you still for your work! I think it could come in handy down the road.

> For the record, as input for that future design - I think that the Hebrew-specific logic can be done without false positives/negatives with one letter of lookahead and variable lookback (until the preceding word boundary).

Thanks, that's valuable info.

> With that said, I do still feel that applying smartquotes on Hebrew in the current state does more harm than good, particularly with the focus on accessibility.

That's fair enough, though it might be a bit of back-and-forth if we end up improving smartquotes later on. I wonder what other Hebrew Typst users think about this.

PR #7540, opened 2025-12-06
Respect text stroke style in math's horizontal bar

### LINKED ISSUE #7510: #text(stroke:red) has no effect on fraction lines in math mode
(fill:red) works, but (stroke:red) does not work. Is this a bug, or expected behavior?

Image

Leedehai:
Fixes #7510. The description below is specific to `math.frac`, but the same logic applies to `math.underline`, `math.overline`, `math.sqrt`, and `math.root` as well. This PR addressed them all.

The horizontal line's thickness is specified by `MathConstants.fraction_rule_thickness`, an attribute in OpenType (see the [spec](https://learn.microsoft.com/en-us/typography/opentype/spec/math)'s `fractionRuleThickness`). This has two implications:
* If we use a pre-composed glyph (e.g. [U+2015](https://www.compart.com/en/unicode/U+2015)) instead of a stroke to draw the fraction's horizontal line, then we will be unable to set the thickness according to that font attribute, and we will be at the mercy of that glyph's font design.
  * Therefore, in this PR, we stick to drawing the stroke, not a pre-composed glyph.
* Giving the fraction's horizontal bar a style of stroke: `red + .5pt` means adding an additional thickness of `.5pt` on top of `MathConstants.fraction_rule_thickness`.
  * But it's okay, since it's the user that explicitly wants to add more thickness. We are not obliged to keep the original thickness described by `MathConstants.fraction_rule_thickness`.


Leedehai:
I'd like to get some validation on this general approach before also handling the under/over bar and extracting the common code :)

laurmaedje:
I'm not sure about the approach. I would expect the stroke to also apply it at the sides:

image

Maybe we should use a rectangle instead of a line when there is a stroke?

Thoughts? @mkorje

mkorje:
I agree it should apply to the sides.

I tried a rectangle a few days ago and it works, but it causes all the tests to fail (the ones involving fractions) because the output is noticably thicker in the PNGs. (I'm guessing this is the rasterisation treating lines differently from rectangles?) Not sure if this is an issue, though. It looks all correct in pdf/svg export modes.

EDIT: Oh I didn't see you mentioned only doing this when there is a stroke, in that case it should all be fine.

laurmaedje:
@Leedehai Then, I'd say let's try the approach where we use a rectangle only if a stroke is defined.

Leedehai:
No problem, I'll modify. For the horizontal line of the square root though, I'd like to note we shouldn't add a stroke to the left side edge of the "rectangle".

laurmaedje:
True. I suppose there we could use a path. But it might still be rather hard to make it compose well with the stroked glyph.

Leedehai:
Done.

For convenience of code review, let me enlarge the ref images of the newly added tests:

* `math-frac-line-fill-stroke`:
  

* `math-underover-line-fill-stroke`:
  

* `math-root-line-fill-stroke`:

crates/typst-layout/src/math/frac.rs
```diff
@@ -144,18 +144,20 @@ fn layout_vertical_frac_like(
...
+            Some(stroke) => Geometry::Rect(Size::new(line_width, thickness))
+                .filled_and_stroked(
+                    text_fill.clone(),
                     FixedStroke::from_pair(
-                        styles.get_ref(TextElem::fill).as_decoration(),
-                        thickness,
+                        stroke.paint.clone().unwrap_or(text_fill),
+                        stroke.thickness.resolve(styles).unwrap_or(Abs::pt(1.0)),
                     ),
                 ),
```
mkorje, line 153:
```rust
                .filled_and_stroked(
                    text_fill.clone(),
                    stroke.clone().resolve(styles).unwrap_or_default(),
                ),

```
(It won't let me add a suggested change over deleted lines)

crates/typst-layout/src/math/underover.rs
```diff
@@ -239,17 +239,21 @@ fn layout_underoverline(
...
+
+    let text_fill = styles.get_ref(TextElem::fill).as_decoration();
+    let line = match styles.get_ref(TextElem::stroke) {
+        Some(stroke) => Geometry::Rect(Size::new(line_width, thickness))
+            .filled_and_stroked(
+                text_fill.clone(),
+                FixedStroke::from_pair(
+                    stroke.paint.clone().unwrap_or(text_fill),
+                    stroke.thickness.resolve(styles).unwrap_or(Abs::pt(1.0)),
+                ),
```
mkorje:
```suggestion
                stroke.clone().resolve(styles).unwrap_or_default(),
```

crates/typst-layout/src/math/root.rs
```diff
@@ -124,6 +124,33 @@ pub fn layout_root(
 
     frame.push_frame(sqrt_pos, sqrt);
     frame.push(line_pos, line);
+
+    // The horizontal line of the root symbol is drawn with a geometry object,
+    // not a text glyph, so the text's stroke style was not automatically applied and we
+    // need to apply it by drawing around the horizontal line.
+    if let Some(stroke) = styles.get_ref(TextElem::stroke) {
```
mkorje: I think you can simplify this by making use of the functions in `typst-layout/src/shapes.rs`, e.g. the `styled_rect` function.

Leedehai:
Done. I also renamed the return object of `styled_rect()` from `rects` to `sides` in deco.rs while I'm at it, because "rects" (short for "rectangles") is quite misleading.

crates/typst-layout/src/math/root.rs
```diff
@@ -124,6 +125,28 @@ pub fn layout_root(
 
     frame.push_frame(sqrt_pos, sqrt);
     frame.push(line_pos, line);
+
+    // The horizontal line of the root symbol is drawn with a geometry object,
+    // not a text glyph, so the text's stroke style was not automatically
+    // applied and we need to apply it by drawing around the horizontal line.
+    if let Some(stroke) = styles.get_ref(TextElem::stroke) {
```
mkorje: For consistency with `TextElem::fill` we should get the stroke from the sqrt MathFragment. Just adding this method to `MathFragment` should work
```rs
    pub fn stroke(&self) -> Option<FixedStroke> {
        match self {
            Self::Glyph(glyph) => glyph.item.stroke.clone(),
            _ => None,
        }
    }
```

crates/typst-layout/src/math/root.rs
```diff
@@ -54,10 +55,11 @@ pub fn layout_root(
     }
     .at(size);
 
+    let text_fill = styles.get_ref(TextElem::fill).as_decoration();
```
mkorje: This should try get the fill from the sqrt MathFragment first:  `sqrt.fill().unwrap_or_else(|| ...)`

tests/ref/render/math-root-line-fill-stroke.png
mkorje, line 1: I'm guessing that this image changed since we are now using a rectangle?

crates/typst-layout/src/math/root.rs
```diff
@@ -126,29 +127,24 @@ pub fn layout_root(
     frame.push(line_pos, line);
```
mkorje: When there is a stroke this should not be pushed, since later on a rectangle gets created.

tests/ref/render/math-root-line-fill-stroke.png
Leedehai: That's right, I enlarged the new ref image and verified the looks is intended and the same as before.
Screenshot 2025-12-14 at 14 04 12

Leedehai:
Done. The below is the ref image of the root resulting from the newest commit, and it has our intended look just like before:
Screenshot 2025-12-14 at 14 13 00

crates/typst-layout/src/math/root.rs
```diff
@@ -55,18 +55,15 @@ pub fn layout_root(
...
-        Geometry::Line(Point::with_x(line_width)).stroked(FixedStroke::from_pair(
-            sqrt.fill().unwrap_or(text_fill.clone()),
-            thickness,
-        )),
-        span,
-    );
 
     let target = radicand.height() + thickness + gap;
     sqrt.stretch_vertical(ctx, target, Abs::zero());
+    let sqrt_fill = sqrt.fill();
```
mkorje:
```suggestion
```

crates/typst-layout/src/math/root.rs
```diff
@@ -145,6 +138,15 @@ pub fn layout_root(
         frame.push_multiple(sides.into_iter().map(|shape| {
             (line_pos - Point::with_y(thickness / 2.0), FrameItem::Shape(shape, span))
         }));
+    } else {
+        let line = FrameItem::Shape(
+            Geometry::Line(Point::with_x(line_width)).stroked(FixedStroke::from_pair(
+                sqrt_fill.unwrap_or(text_fill.clone()),
```
mkorje:
```suggestion
                text_fill,
```

Leedehai:
Done. Thanks!

laurmaedje:
Thank you! Turned out nicely. :)

laurmaedje:
It looks like this doesn't work together with other changes on main. Could you rebase?

Leedehai:
Done. All tests are passing. For review's convenience, let me enlarge the ref images at the moment:

* `math-frac-line-fill-stroke`


* `math-root-line-fill-stroke`


* `math-underover-line-fill-stroke`

tests/ref/render/math-frac-line-fill-stroke.png
mkorje, line 1: Any idea why this changed?

Leedehai: My guess would be the update of the CM font files (https://github.com/typst/typst/pull/7597) that was merged  after this PR's creation.

Following up: after this PR's merging, another update of the CM font files (https://github.com/typst/typst/pull/7663) changed this image yet again, as noted by the comment there. Weird indeed.

PR #7544, opened 2025-12-07
Introduce `ext` attribute on `#[scope]` instead of hardcoding primitives

laurmaedje:
This makes it possible to use other types than `bool`, `i64`, and `f64` as native types with a scope definition without them being defined in `typst-library`.

### DIFF
--- a/crates/typst-macros/src/scope.rs
+++ b/crates/typst-macros/src/scope.rs
@@ -1,13 +1,15 @@
 use heck::ToKebabCase;
 use proc_macro2::TokenStream;
 use quote::quote;
+use syn::parse::{Parse, ParseStream};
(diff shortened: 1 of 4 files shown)


crates/typst-macros/src/scope.rs
```diff
@@ -113,6 +115,19 @@ pub fn scope(_: TokenStream, item: syn::Item) -> Result<TokenStream> {
     })
 }
 
+/// The `..` in `#[scope(..)]`.
+struct Meta {
+    /// Whether this the scope should be implemented through an extension
```
bastienvoirin, line 120:
```suggestion
    /// Whether this scope should be implemented through an extension
```

or

```suggestion
    /// Whether the scope should be implemented through an extension
```

PR #7552, opened 2025-12-09
Fix wording in syntax documentation

chanbai1999:
This PR fixes two wording and clarity issues in the documentation:

Math mode description ambiguity
The original text said math mode is entered by wrapping an equation in [$] characters, which may be mistaken as using a single literal character.
→ Updated to “paired [$..$] delimiters” to clearly indicate that both opening and closing delimiters are required.

Inaccurate explanation in the Identifiers section
The previous comment suggested that -π is invalid because it means -1 * π.
→ Updated to clarify that the leading “-” is a unary minus operator, and identifiers cannot start with a hyphen, so -π is not a valid identifier.

These edits improve precision and avoid potential misunderstandings for readers.


laurmaedje:
I'm not really convinced that these changes are an improvement. The PR description also feels AI written which makes me a bit skeptical.

Issue #971, opened 2023-04-25
Paths are unhygienic — strings and paths should be separate types


jhpratt:
### Description

Assume the following directory layout and file contents.

```text
├── demo.rs
├── include-code.typ
├── main.typ
└── sections
    └── demo.typ
```

`demo.rs`:
```rust
fn main() {}
```

`include-code.typ`:
```text
#let include-code(lang: "rust", code_path) = raw(read(code_path), lang: lang)
```

`main.typ`:
```text
= Demo
#include "sections/demo.typ"
```

`sections/demo.typ`:
```text
#import "../include-code.typ": *

#include-code("../demo.rs")
```

Running `typst c main.typ` gives an error.

```text
error: file not found (searched at ../demo.rs)
  ┌─ include-code.typ:1:54
  │
1 │ #let include-code(lang: "rust", code_path) = raw(read(code_path), lang: lang)
  │                                                       ^^^^^^^^^

help: error occurred in this call of function `include-code`
  ┌─ sections/demo.typ:3:1
  │
3 │ #include-code("../demo.rs")
  │  ^^^^^^^^^^^^^^^^^^^^^^^^^^

help: error occurred while importing this module
  ┌─ main.typ:2:9
  │
2 │ #include "sections/demo.typ"
  │          ^^^^^^^^^^^^^^^^^^^
```

The reason for this makes sense _given knowledge of internals_. The path "../demo.rs" is merely a string, and thus unmodified when passed to `include-code.typ`. This is despite `demo.typ` and `include-code.typ` being in different directories. For this reason, any situation where paths are expected are fundamentally unhygienic.

My recommendation is to make paths and strings separate types, such that paths can be made hygienic. A path could reasonably be constructed from a string using a built-in `path` function.

### Use Case

^^

user202729:
For comparison, in TeX unless special handling is used (`subfiles` package I believe) all paths are from the "root" (current working) directory. So passing it through different files is not a problem.

The same situation is true in Python and `open()` function. On the other hand `import .abc` in Python is relative to the current file in the package. And the `Path` type in Python doesn't affect that at all.

Ideas? Would making `#import` be relative and everything else be absolute fix the problem?

laurmaedje:
[...]

PR #7555, opened 2025-12-10
Add file `path` type

laurmaedje:
This PR adds a Typst-exposed `path` type. It closes https://github.com/typst/typst/issues/971.

## Purpose

The primary purpose of this type is to allow for construction of paths in one file that correctly resolve in another. All functionality that used to accept a path string now accepts a string _or_ a path.

The path type is designed to allow for operations like file existence checks or path enumeration. However, those additional features are not part of this initial PR. This PR also neither introduces syntax for the path type, nor does it remove support for using strings as paths. It's an additive change that allows explicit construction of a path via the `path` constructor. Special syntax for paths is not explicitly planned, but also not entirely off the table.

The PR ships with two (relatively minor) breaking changes (see below).

## Usage

Once constructed, a path can be freely used across the project and all packages, so that e.g. a package can read from a path passed to it by the project. An example usage of this would look like this:

```typ
// This is main.typ, with data.json next to it.
#import "@local/my-pkg:0.1.0": process
#let data-path = path("data.json")
#process(data-path)
```

## Roots

The project and each package are separate _roots_ that are fully isolated. The path type always represents an absolute path within the project or a package. It validates upon construction that its associated root is not escaped:

```typ
#path("../secret.txt")
```
```
error: path would escape the project root
  ┌─ hi.typ:1:6
  │
1 │ #path("../secret.txt")
  │       ^^^^^^^^^^^^^^^
  │
  = hint: cannot access files outside of the project sandbox
  = hint: you can adjust the project root with the --root argument
```

## Breaking changes

This PR introduces two breaking changes in Typst's path handling:

1. The Typst CLI does not accept non-Unicode input paths anymore.[^1] This change is necessary to let Typst switch to a fully portable, custom path representation. So far, Typst internally used Rust's `PathBuf` which has platform-specific behavior, leading to subtle differences in behavior across platforms (see point 2).

   Non-Unicode paths are exceedingly rare nowadays and popular tools such as `cargo` also require UTF-8 paths, so we're in good company with this change. There is some further discussion about this in the [docs of the `camino` crate](https://docs.rs/camino/latest/camino/). (We don't depend on `camino` since it still has OS-dependent behavior though.)

2. Typst now rejects paths containing backslashes in source files. These used to be allowed, but with varying behavior across platforms. On Windows, they would indicate path separators while on Linux they would be part of a component. In the web app, backslashes cannot occur in file names.

    I considered deprecating paths with backslashes before removing support. However, if a package would use backslashes, it would already be broken today across every non-Windows system, including the web app. Meanwhile, for document authors, updating paths to use forward slashes should be a very straightforward migration that's aided by the following error message:

    ```
    error: path must not contain a backslash
      ┌─ hi.typ:1:8
      │
    1 │ #import "path\\to\\file.typ"
      │         ^^^^^^^^^^^^^^^^^^^^
      │
      = hint: use forward slashes instead: `"path/to/file.typ"`
      = hint: in earlier Typst versions, backslashes indicated path separators on Windows
      = hint: this behavior is no longer supported as it is not portable
    ```

    Shaking out any non-portable usage of backslashes now and not burdening the new path implementation with platform-specific behavior seemed more important to me than letting document authors update a bit more slowly.
    
    Edit: Paths with backslashes will of course continue to be accepted by the CLI on Windows. The restriction only applies to "virtual" paths within Typst files. Real paths are segmented using Rust's platform-specific component logic when they are converted to virtual paths.

## Notes

If you want to take a look at the diff, I recommend going commit by commit.

[^1]: More precisely, it does not accept paths with non-UTF-8 components in the part that exceeds the project root. The path prefix that is shared with the project root may still contain non-UTF-8 components as it does not end up in Typst's path type.


gwennlbh:
does this mean the CLI won't take in backslash-separated paths? on windows (powershell), autocompleting paths with Tab turns all slashes into backslashes (even forward ones) 😅

laurmaedje:
Good point, that wasn't clear in my PR description. No worries, backslashes will continue to work in the CLI on Windows. The restriction only applies to paths within Typst (sandboxed _virtual_ paths). Real paths are segmented using Rust's normal, system-specific `Path` component logic when they are converted to virtual paths.

MDLC01:
In the "path would escape the project root" error, it might be useful to display the path in the hints, in case it's not a literal in the source code.

sijow:
> On its own, code in a package cannot construct a path that lives in the project or another package. For this reason, it can also not load files from the project directory.

This will limit the functionality of packages that process files that refer to other files. For example a Markdown cell in a Jupyter notebook can include images from the notebook directory with ``. Calling `#callisto.render(nb: path("notebook.ipynb"))` will fail to render these images...

This may seem like an edge case but many data formats include relative paths to other files, for example HTML, 3D graphics (text formats with external texture files), various medical imaging and GIS formats, HDF5 (a generic data format used in many science/engineering applications)...
 
We could use hacks to work around this, for example I can make `callisto.render()` accept an optional callback:

```
#callisto.render(nb: path("notebook.ipynb"), reader: name => path(name))
```

but maybe this could be made easier?

sijow:
A possible API:

```typ
#let nb = path("notebook.ipynb", share-directory: true)
#callisto.render(nb: nb)
```

(The user could also do `#set path(share-directory: true)` once at the beginning of the document.)

QuadnucYard:
When will path operations (`join`, `parent`, `stem`, `extension`, etc.) come?

Some solutions to "package access user workspace" I thought of:
- Join user user-provided assets root path with the file names required by the package.
- Provide a global path constant for the workspace root for packages to use.

Andrew15-5:
> ```text
> = hint: use forward slashes instead: `"path/to/file.typ"`
> = hint: in earlier Typst versions, backslashes indicated path separators on   Windows
> = hint: this behavior is no longer supported as it is not portable
> ```

Oh... I thought the code indeed has a double space. It doesn't.

>    Edit: Paths with backslashes will of course continue to be accepted by the CLI on Windows. The restriction only applies to "virtual" paths within Typst files. Real paths are segmented using Rust's platform-specific component logic when they are converted to virtual paths.

Can you give a small example of what this means? I don't get how only virtual paths cannot have it, but actual paths can, but only on Windows...even though supposedly virtual paths are always constructed from real ones (i.e., from `str`ing)?

https://github.com/typst/typst/blob/a2aa1274b9af95b132599fb47a7e495a15f9e3aa/crates/typst-library/src/foundations/path.rs?plain=1#L111-L112

I don't get it, it seems like this prevents the whole point of paths, making the example error on project root access:

```typ
// This is main.typ, with data.json next to it.
#import "@local/my-pkg:0.1.0": process
#let data-path = path("data.json")
#process(data-path)
```

If this also is true for the `package-func(read: path => read(path))` hack that allows packages to read stuff from project root, then it's definitely a breaking change, and a way to do the same (but better?) must exist. From my last usage, this will affect not only Alexandria, but also my long-used `code` helper that will read the file content for a path, and then generate the listing (with provided path string as its header) + `pdf.attach`.

```typ
#let code = code-inner.with(read: path => read(path))
```

This is a very powerful feature-hack that must not go away, even if just for a few months or one release.

---

The unification and simplification of paths is very nice, since indeed it should immensely help with removing a ton of edge cases and in general, when working with paths in any way later down the road.

> When will path operations (`join`, `parent`, `stem`, `extension`, etc.) come?

If the `read: path => read(path)` still works, then I guess not very soon, but I'm pretty sure it all a must-have, right with file/dir/link/etc. existence check and directory walking. If not, then at least some of that will have to be a part of this PR.

laurmaedje:
Oops, the docs contain a wrong sentence that is still partly left from the old docs. I'll fix that (Edit: Done).

To clarify: It _is_ possible for a package to read from a path pointing into the project. That's the whole point of this. It's just not possible for the package to, on its own, _create_ a path living in the project. It needs to be handed such a path by the project.

If we expose path modification operators (like joining), it would also be possible for packages to read from a whole directory tree when it's passed a single path. And unless we protect against joining `..` also from files upwards of that directory (not beyond the root though).

For this reason, it needs careful consideration how the path modification operations should be exposed and which kinds of cross-root operations we want to allow. There is no concrete timeline for that yet, it might happen soon, but could also take a while.

---

> Can you give a small example of what this means? I don't get how only virtual paths cannot have it, but actual paths can, but only on Window

Example: `C:\\Users\\Laurenz\\Desktop\\Project\\src\\main.typ` would be normalized within the root `C:\\Users\\Laurenz\\Desktop\\Project` to `/src/main.typ` through `VirtualPath::virtualize`. And when you call `.realize` on the virtual path `/src/main.typ`, it's joined to the physical root and becomes `C:\\Users\\Laurenz\\Desktop\\Project\\src\\main.typ` once again. The backslashes never appear in the virtual path because the real path is segmented into components using Rust's path API and those components are joined by `/` to create the virtual path. The backslash only works on Windows because `std::path::Path::components` splits by a platform-dependent path separator.

Andrew15-5:
> It _is_ possible for a package to read from a path pointing into the project. That's the whole point of this. It's just not possible for the package to, on its own, _create_ a path living in the project. It needs to be handed such a path by the project.

This doesn't answer if `read: path => read(path)` will still work or will it break. It doesn't use `path` type, so you can easily do whatever path operations through strings, and still access any project-root stuff like that.


> Example: `C:\\Users\\Laurenz\\Desktop\\Project\\src\\main.typ` would be normalized within the root `C:\\Users\\Laurenz\\Desktop\\Project` to `/src/main.typ` through `VirtualPath::virtualize`. And when you call `.realize` on the virtual path `/src/main.typ`, it's joined to the physical root and becomes `C:\\Users\\Laurenz\\Desktop\\Project\\src\\main.typ` once again. The backslashes never appear in the virtual path because the real path is segmented into components using Rust's path API and those components are joined by `/` to create the virtual path. The backslash only works on Windows because `std::path::Path::components` splits by a platform-dependent path separator.

No, I mean a Typst example. Conceptually, it's not that hard. What will and won't work on which platform within Typst code? The example error shows that in Typst, you are not allowed to use backslashes, yet it is somehow "allowed" on Windows platform. Is it strictly for CLI path arguments or for Typst code as well?

laurmaedje:
> This doesn't answer if `read: path => read(path)` will still work or will it break.

It will continue to work, otherwise this would be included in the "Breaking changes" section. The breaking changes section is intended to be comprehensive. Even if this closure is passed to a package, the `read` operation happens in the project and not a package.

> It doesn't use path type, so you can easily do whatever path operations through strings, and still access any project-root stuff like that.

Yes, if you provide such a closure explicitly, you can.

> No, I mean a Typst example. Conceptually, it's not that hard. What will and won't work on which platform within Typst code? 

Within Typst, backslashes will work on no platform, forward slashes on all. It's always the same, that's the point. It being allowed on Windows is strictly for CLI path arguments.

Andrew15-5:
> otherwise this would be included in the "Breaking changes" section. The breaking changes section is intended to be comprehensive.

Making sure you didn't forget about it, since it's also missing from outside the "Breaking changes" section.

---

Great. I don't have any more questions. Though I haven't looked at the diff yet. Not sure if I'll have time for this in the upcoming weeks. Either way, the docs can be adjusted later.

Andrew15-5:
Oh, no, actually, I have another question, after I was thinking of the next path-related features. When you create a `path` object, does it error if the target does not exist upon construction? Or is it currently the same as string, i.e., only errors when trying to access the path target within `read`, `raw`, etc.?

laurmaedje:
The path does not need to exist. If it had to, https://github.com/typst/typst/pull/7556 would be pointless. :p

blaggacao:
Out of a conversation over at https://github.com/typst/typst/pull/7556, I want to submit in this thread:

image

When I click on this in VSCode, I'm directly led to the site where the wrong string representation of the file is put.

Is that still ensured with the current `path()` construction, or would be extra amendments to the PR necessary (e.g. passing position context) to make sure, the _site of path declaration_ is also reported at the _site of loading_, if loading fails?

TheAlgorythm:
> If we expose path modification operators (like joining), it would also be possible for packages to read from a whole directory tree when it's passed a single path. And unless we protect against joining .. also from files upwards of that directory (not beyond the root though).

`path` could have an optional argument `allow-modification`. The second issue could be solved with my simple [path_ratchet](https://crates.io/crates/path_ratchet) crate.

blaggacao:
Post Datum: 

I havn't checkt, so this concern may be invalid.

You click on an image in tinimist IDE preview, then you land in the implementation details of some library instead at the site of path definition.

Is the necessary foundations in place for an IDE to implement correct UX behaviour?

laurmaedje:
> When I click on this in VSCode, I'm directly led to the site where the wrong string representation of the file is put.

You are always led to the place where the wrong string is used. That's also the case now. If you call `image` with a wrong path, the error conceptually directs you to the `image` call rather than the path itself; but both happen to be in the same place. It will remain the same with the path type. You will be led to the place where the path is wrongly used, not where the path is constructed. Often, both will still be the same; in particular when a string is coerced to a path. I think that's reasonable.

blaggacao:
I agree that this can be a reasonable trade-off, in the short term.

In case this ends up evolving into heavily used core infrastructure, extra effort on position reporting might be due.

RenanElfo:
I don't know if GitHub threads are a good place for compliments, but I'd like to say I'm very happy with this new feature and I'm really looking forward to Typst 0.15. Good job folks!

PR #7556, opened 2025-12-10
Add `path.exists` function

### LINKED ISSUE #2025: Check if file exists

laurmaedje:
This PR adds support for a `path.exists` function that checks whether a file or directory exists at a path. It depends on https://github.com/typst/typst/pull/7555.

The implementation right now uses the existing `World` infrastructure. It tries to load the path, and matches on the result to make a decision. In particular, it depends on World implementors correctly returning `FileError::IsDirectory` for directories.

I am not 100% decided whether we should keep it this way or whether we should extend the `World` interface to make this more explicit (maybe with a default impl mirroring what I've done here), also allowing it to be more efficient. We should also consider how the `World` interface needs to be adapted to support [directory walking](https://github.com/typst/typst/issues/2123). Maybe there is some overlap.

The function also still needs some more documentation.

This closes https://github.com/typst/typst/issues/2025.

### DIFF
--- a/crates/typst-library/src/foundations/path.rs
+++ b/crates/typst-library/src/foundations/path.rs
@@ -1,7 +1,11 @@
 use ecow::{EcoString, eco_format};
 use typst_syntax::{FileId, PathError, RootedPath, Spanned, VirtualRoot};
 
-use crate::diag::{At, HintedStrResult, HintedString, SourceResult, error};
+use crate::World;
+use crate::diag::{
+    At, FileError, HintedStrResult, HintedString, SourceResult, StrResult, error,
+};
+use crate::engine::Engine;
 use crate::foundations::{Repr, Str, cast, func, scope, ty};
 
 /// A file system path.
@@ -158,6 +162,17 @@ impl RootedPath {
         path: Spanned<PathOrStr>,
     ) -> SourceResult<RootedPath> {
         path.v.resolve_if_some(path.span.id()).at(path.span)
+    }
+
+    /// Checks whether a file or directory exists at this path.
+    #[func]
+    pub fn exists(self, engine: &mut Engine) -> StrResult<bool> {
+        match engine.world.file(self.intern()) {
+            Ok(_) => Ok(true),
+            Err(FileError::IsDirectory) => Ok(true),
+            Err(FileError::NotFound(_)) => Ok(false),
+            Err(err) => Err(err.into()),
+        }
     }
 }
 
--- a/tests/suite/foundations/path.typ
+++ b/tests/suite/foundations/path.typ
@@ -4,6 +4,15 @@
   "path(\"/tests/suite/foundations/hi/there.txt\")",
 )
 
+--- path-file-exists paged ---
+#assert(path("./path.typ").exists())
+
+--- path-dir-exists paged ---
+#assert(path("/tests").exists())
+
+--- path-none-exists paged ---
+#assert(not path("./unknown.typ").exists())
+
 --- path-escapes paged ---
 // Error: 7-29 path `"../../../../file.txt"` would escape the project root
 // Hint: 7-29 cannot access files outside of the project sandbox


blaggacao:
Regardless of impl concerns: Is there really a use case, where separating path construction and validation makes sense?

My imagination falls short, atm. I can imagine, however, that cost of exposing this api might outweigh the benefit.

laurmaedje:
> Regardless of impl concerns: Is there really a use case, where separating path construction and validation makes sense?

Could you elaborate? The idea is to check something "does this image exist" before calling the `image` function and displaying a placeholder otherwise. If the path construction failed if that path did not exist, that use case would become impossible.

blaggacao:
Got ya. What would this placeholder be useful for? - I have little depth of context / authority to hold a strong opinion.

Without that context, I could believe that it would be beneficial to fail on construction if the path does not exists. - as an implementation of a "fail early and hard" principle.

Since paths can't break out the project root, I reason that users of `path`  are supposed to have _always_ intricate knowledge of the repository scope, i.e. are supposed to _never_ construct a non-existing path.

If they do, we _can_ assume it's a genuine mistake (e.g. typing error), hence omitting to fail upon construction delays the user facing error / fix.

---

In a scenario where paths could indeed break out the project root scope (which they can't), I _could_ imagine a case where a user would want to construct and inventorize on purpose a path that does not (yet - in the context of construction) exist.

---

To a certain extend and in this line of thinking, it feels like this feature "contradicts" the enforcement repository scope.

QuadnucYard:
IMO, if we were to use the design that `path` is always valid, we could have `path.exists(str)` instead to check existence.

To avoid double-checking, we may need something like `path.try-new(str)`, which returns none if the path is not valid.

blaggacao:
> IMO, if we were to use the design that path is always valid, we could have path.exist(str) instead to check existence.

I find this an interesting proposal, indeed. This would have good use cases at the boundary between old string paths and new paths and could be a valuable tool for library (package) authors.

E.g.

```
# project

p = path(./assets)
```

```
# lib / package

path.exists(p + "/images")

# maybe a dir walking follows to enumerate images
```

laurmaedje:
Considering that https://github.com/typst/typst/issues/2025 has thirty upvotes and many use cases, people seem to have lots of use cases for checking if a path exists. 

Personally, I don't really see the point in failing on path construction. It complicates the API design for existence checks (e.g. `path.exists(str)` would again suffer from not being able to pass that to a package, i.e. you could not write a wrapper for `exists` that resolves properly) and also the implementation. It's also unlike how any other language treats paths.

To me, the path is just a description of something that might be in the file system and the _load_ operation is what fails when the file does not exist.

blaggacao:
> Considering that https://github.com/typst/typst/issues/2025 has thirty upvotes and many use cases, people seem to have lots of use cases for checking if a path exists.

Just to quickly counter the impression it might give: that's technically unrelated to the proposal, to fail on construction.

> Personally, I don't really see the point in failing on path construction.

Besides the principled concern "fail fast and hard", I explained, there's also an error context coming to my mind. If the path fails later, it is harder to trace where the path had been construed, and to ease this, some additional position context would have to be passed into the path type. And knowing quickly what went wrong where with proper error context is not something of 0 value in a typo scenario, for example.

image

In VSCode, when I click here, I'm directly led to the correct code line where the wrong path had been added.

> It complicates the API design for existence checks (e.g. path.exists(str) would again suffer from not being able to pass that to a package, i.e. you could not write a wrapper for exists that resolves properly) and also the implementation

I can't give a proper opinion, but if this is the tradeoff and preferences, then I indeed have nothing useful to add.

> It's also unlike how any other language treats paths.

Indeed.

---

Overall, I think I made my point well enough for your appreciation 😄 - Thanks for this little back & forth. I'll not push the proposal further.

tovrstra:
One small suggestion (for a future PR if any). If this gets added (hopefully), it would be beneficial to extend the JSON dependencies format with a list of files whose existence was checked.

For example, a document `summary.typ` has the following:

```typst
#let cases = if path("cases.json").exists() {
  json("cases.json")
} else {
  ("case1", "case2")
}
```

Then `typst compile summary.typ --deps deps.json --deps-format json` could produce one of the following two:

- if `cases.json` exists:

    ```json
    {
      "inputs": ["summary.typ", "cases.json"],
      "checks": ["cases.json"],
      "outputs": ["summary.pdf"],
    }
    ```

- if `cases.json` does not exist:

    ```json
    {
      "inputs": ["summary.typ"],
      "checks": ["cases.json"],
      "outputs": ["summary.pdf"],
    }
    ```

This would make it possible for a build system to decide that a document needs to be recompiled when a file `cases.json` is created. (When it does not exists yet, it cannot be included in the list of inputs.) It might also be useful to decide that it is worth trying to recompile when the `cases.json` file gets removed.

Beiri22:
I like the recent movement on this topic. Any clues on the decision, yet?

laurmaedje:
I'm pretty busy right now, but this is still on my radar.

laurmaedje:
The PR here is too hacky. This will need proper `World` support. I'll definitely revisit this later, but close it for now.

PR #7572, opened 2025-12-12
Fix technical details in v0.14.2 changelog

Andrew15-5:
Apart from 4 GB not being equal 2³² bytes (and whitespace missing between value and unit), I don't understand why the limit is 4 GiB, when https://github.com/WebAssembly/memory64/blob/main/proposals/memory64/Overview.md was already merged, as well as https://github.com/wasmi-labs/wasmi/issues/1357. Furthermore, all major browsers [support Memory64](https://caniuse.com/?search=memory64), apart from Safari (which is a reoccurring issue). For CLI specifically I don't see how it can be a 4 GiB limit.


Enivex:
Typst is not yet using Memory 64. Note that it's somewhat slower, so it's not actually a win unless you need that memory.

Andrew15-5:
All the underlying stuff has support for it. Does

https://github.com/typst/typst/blob/839c6f76f331dd2f30ede9affa9d734f88828d83/crates/typst-library/src/foundations/plugin.rs?plain=1#L484

indicate that it's not used? If `I64` is to be used, will Typst also support Memory64?

Regardless, the text refers to "WebAssembly linear memory" without any notes/remarks (other than "on a technical level", whatever this vague part means), which means in the latest spec this 4 GiB limit is outdated. This means that some sort of note about "Typst doesn't use Memory64 yet" should be included, to set the context straight.

laurmaedje:
This is unnecessarily nitpicky and not worth changing a released changelog in my book.

laurmaedje:
Regarding why the limit is 4GB: Memory64 supports [need to be enabled specifically](https://docs.rs/wasmi/latest/wasmi/struct.Config.html#method.wasm_memory64) in wasmi.

Andrew15-5:
> This is unnecessarily nitpicky

No, you've said "technical details", which means you have to be correct on these technical details (otherwise, what's even the point?). And according to SI (BIPM) and NIST bodies, 4 gigabytes are completely different from 4 gibibytes, by 294967296 bytes, or about 281 MiB. This is a technical error, which is very easy to fix, and being "a released changelog" doesn't mean you can't fix mistakes.

If it wasn't "technical" at all, then there is a bit more room to let it slide, as the majority of regular users don't know about GiB in the first place, or think that it's interchangeable with GB.

PR #7573, opened 2025-12-12
Clarify documentation of `x` and `y` in stroke dictionary

eltos:
Clarified documentation for 'x' and 'y' stroke parameters.

The documentation was ambiguous, because "horizontal stroke" referred to the stroke on the horizontally adjacent sides, the stroke lines of which are in turn drawn vertically.

### DIFF
--- a/docs/guides/tables.md
+++ b/docs/guides/tables.md
@@ -560,7 +560,7 @@ table with only horizontal lines, you can do this:
 ```
 
 This turns off all vertical strokes and leaves the horizontal strokes in place.
-To achieve the reverse effect (only horizontal strokes), set the stroke argument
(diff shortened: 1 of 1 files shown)


Enivex:
If it's ambiguous which sides they apply to, maybe the names x and y aren't optimal?

eltos:
It was only the documentation which was ambiguous, because the horizontally adjacent sides have vertical stroke lines and vice versa.

Maybe x/y isn't perfect, but at least its consistent with table/grid/etc. One could think about lr/tb (similar to math attach), but i don't think it's better for table/grid with midlines. Also it would be breaking. Just fixing the documentation seems reasonable to me.

laurmaedje:
Thanks!

Issue #5700, opened 2025-01-15
"." is not a punctuation in math


arnaudgolfouse:
### Description

A single dot is not in the punctuation class in math mode, as shown by the following example:
```typst
$ forall x. x = 1 $

#show ".": math.class("punctuation")[.]

$ forall x. x = 1 $
```

![image](https://github.com/user-attachments/assets/27c6e1ad-bd8e-4a6d-9148-a458dffe6591)

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

MDLC01:
For information, Unicode indeed classifies the period as a punctuation.[^1]

[^1]: https://www.unicode.org/Public/math/latest/MathClassEx-15.html

Enivex:
> For information, Unicode indeed classifies the period as a punctuation.[^1]
> 
> [^1]: https://www.unicode.org/Public/math/latest/MathClassEx-15.html

Typst uses that list, so I'm not sure what's going on here.

EpicEricEE:
> Typst uses that list, so I'm not sure what's going on here.

I'm not sure about the exact reasoning, but some characters don't use the unicode math class:
https://github.com/typst/typst/blob/a4ac4e656267e718a5cf60d1e959f74b2b7346f3/crates/typst-layout/src/math/fragment.rs#L279-L285

I could imagine that it's for numbers, so that there is no space after the decimal separator, though I was under the impression that numbers are handled differently anyway, so I don't know.

Enivex:
You're right that the show rule breaks decimal numbers, so that may be why. It seems like something that could be fixed though.

mkorje:
I've tracked down all the changes that have happened here:
+ During lexing, grapheme clusters and numbers (including with periods) are grouped together as one text element. In the layout of text in math, there is a branch for numbers triggered by `text.chars().all(|c| c.is_ascii_digit())`. Thus numbers with a period in them went through the normal text layout. And as numbers with a period necessarily have more than one grapheme cluster, the created frame got set as spaced. This ensured `$1.3$` has no spacing, however things like `$1.3 degree$` had unwanted spacing (in this case, between the 1.3 and the degree symbol). At this point in time, a period had math class punctuation, so `$1. 3$` for example had spacing after the period.
+ In #1855, the class for a period was changed from punctuation to normal. See the PR for the reasoning given. Now `$1. 3$` had no spacing.
[...]

Issue #1093, opened 2023-05-03
decimal separator in math mode for spanish


asaltanubes:
### Description

In Spanish the decimal separator is a comma instead of a period, in latex when using Babel in Spanish mode when in math mode the decimal separator is changed automatically. It would be nice to have the same treatment when using text(lang: "en")

### Use Case

#set text(lang: "es")
$3.1415$
should output something like 
![image](https://user-images.githubusercontent.com/56401901/235989705-c196c021-92a1-4e72-ae05-629b457d7ae7.png)
instead of 
![image](https://user-images.githubusercontent.com/56401901/235990025-60118022-1224-42c2-b44f-119d878e1f13.png)

owiecc:
This is not limited to Spanish language: https://en.wikipedia.org/wiki/Decimal_separator#Conventions_worldwide

asaltanubes:
Found a solution, it isnt the cleanest but it works.
`
#show regex("\d+\.\d+"): it => {show ".": ","
  it}
`
if only wanted in math mode
`
#show math.equation: it => {
show regex("\d+\.\d+"): it => {show ".": ","
  it}
it
}
`

binarynoise:
The math-mode rule seems broken, it requires semicolons or line breaks before the `it`s now

asaltanubes:
True single quote code block from github deleted the newlines. I fixed it.

DegrangeM:
Thanks for the code, even though it seems important to be avaible natively in typst (french here).

FeldrinH:
https://github.com/typst/typst/issues/448 is more or less the same issue, but without a concrete proposal.

Blezz-tech:
I hope this feature will be added in the near future

remiangot:
It should also handle spaces and separator for thousands like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

30000.65 → "30,000.65" where English is the default language, or
30000.65 → "30.000,65" where German is the default language, or
30000.65 → "30 000,65" where French is the default language

mkpoli:
An alternative solution using a function
```typst
#let n(num, decimal: ".", thousands: ",") = {
  let parts = str(num).split(".")
  let decimal_part = if parts.len() == 2 { parts.at(1) }
  let integer_part = parts.at(0).rev().clusters().enumerate()
    .map((item) => {
      let (index, value) = item
[...]

PR #7577, opened 2025-12-13
Add `math.num` to render number vs number-like text separately

### LINKED ISSUE #7574: Footnotes in math are not superscript
### Description

```typst
#set page(height: 5cm, width: 5cm)

$ "Math text" #footnote[Footnote] $

$ #[Also text #footnote[Also footnote]] $
```

Rendered as

Image

Those `1` and `2` are not superscript, but they should be.

### Reproduction URL

https://typst.app/project/rXgMyui38t7LEw376Dhy2y

### Operating system

Web app

### Typst version

Leedehai:
This PR adds a new math element `math.number` with the introduction of struct `NumberElem`, so that number (e.g. `$2.1$`) and number-like text (e.g. `$"2.1"$`) can be differentiated. A number can contain multiple digits and a decimal point in the middle, as controlled by the lexer's [`math_text()`](https://github.com/typst/typst/blob/839c6f76f331dd2f30ede9affa9d734f88828d83/crates/typst-syntax/src/lexer.rs#L692).

This way, as demonstrated in the new ref image `math-number-parsed.png`, we can apply a different rendering procedure to the numbers (see the changes in this PR's math/text.rs).

This PR fixes #7574 (see the new ref image `issue-7574-footnote-inside-math.png`). The footnote notation (e.g. `1` and `note 2`) doesn't really have the semantics of a number in math, and thus should be rendered as text. See comments [this](https://github.com/typst/typst/issues/7574#issuecomment-3648933177) and [this](https://github.com/typst/typst/issues/7574#issuecomment-3649249176) on that issue.

This PR changed the ref image `enum-numbering-closure-nested-complex.png`, because a numbered list isn't in a math context, and therefore the bullet point indexes should be rendered as text.

This PR changed the code snippet of test `math-attach-nested-base` in `attach.typ`, because `[0]` is a content block holding a text of zero, while `[$0$]` forces the zero to be a number in math context. This difference is also specifically tested in the new test  `math-number-parsed` (see `a0` and `b0` there).

CC @mkorje


MDLC01:
Would this PR make it possible to change the math class of the period back to Punctuation without affecting numbers (https://github.com/typst/typst/issues/5700)?

Leedehai:
@MDLC01 this can be done with ease 🎊  by adding these diffs on top of this PR:

```diff rust
--- a/crates/typst-utils/src/lib.rs
+++ b/crates/typst-utils/src/lib.rs
@@ -373,7 +373,9 @@ pub fn default_math_class(c: char) -> Option<MathClass> {
 
         // Better spacing.
         // https://github.com/typst/typst/pull/1855
-        '.' | '/' => Some(MathClass::Normal),
+        '/' => Some(MathClass::Normal),
+
+        '.' => Some(MathClass::Punctuation),

--- a/crates/typst-layout/src/math/text.rs
+++ b/crates/typst-layout/src/math/text.rs
@@ -126,7 +126,10 @@ pub fn layout_number(
 
         // This won't panic as ASCII digits and '.' will never end up as
         // nothing after shaping.
-        let glyph = GlyphFragment::new_char(ctx, styles, c, span).unwrap();
+        let mut glyph = GlyphFragment::new_char(ctx, styles, c, span).unwrap();
+        if c == '.' {
+            glyph.class = MathClass::Normal;
+        }
         fragments.push(glyph.into());
     }
     let frame = MathRun::new(fragments).into_frame(styles);
```
and I can verify it works as intended :) However, I'd say it's worth opening a separate PR after this current one's acceptance, because this additional change will break these tests and their fix needs more thinking:
* `math-class-exceptions`, especially what to do with `d_0.d_1d_2`. We may want to do `d_0\.d_1d_2`.
* `math-mat-align-explicit-mixed`, which I'm not sure how to modify, because `18&&.02` and the like is not parsed by the the lexer's [`math_text()`](https://github.com/typst/typst/blob/839c6f76f331dd2f30ede9affa9d734f88828d83/crates/typst-syntax/src/lexer.rs#L692) as one number due to alignment point `&&` in the middle, and thus the dot (as a `MathClass::Punctuation`) will introduce a space there. We might need to modify the lexer to make it right (or resign ourselves to this fact).
* `issue-2044-invalid-parsed-ident`.

Mc-Zen:
I wonder what the consequences would be for numerous packages that generate numbers in equations from strings.

crates/typst-syntax/src/ast.rs
```diff
@@ -868,8 +868,8 @@ impl<'a> MathText<'a> {
     pub fn get(self) -> MathTextKind<'a> {
         let text = self.0.text();
         if text.chars().next().unwrap().is_numeric() {
-            // Numbers are potentially grouped as multiple characters. This is
-            // done in `Lexer::math_text()`.
+            // A number is potentially a groupe of multiple characters. This is
+            // done in [`Lexer::math_text()`].
```
mkorje:
```suggestion
            // A number is potentially a group of multiple characters. This is
            // done in [`Lexer::math_text()`].
```

crates/typst-library/src/math/number.rs
```diff
@@ -0,0 +1,43 @@
...
+/// ```
+#[elem(Mathy, Repr, PlainText)]
+pub struct NumberElem {
+    /// The number.
+    #[required]
+    pub text: EcoString,
+}
+
+impl NumberElem {
+    /// Create a new packed symbol element.
```
mkorje:
```suggestion
    /// Create a new packed number element.
```

crates/typst-library/src/math/number.rs
```diff
@@ -0,0 +1,43 @@
+use crate::foundations::{Content, NativeElement, Packed, PlainText, Repr, elem};
+use crate::math::Mathy;
+use ecow::{EcoString, eco_format};
+
+/// A number in math.
+///
+/// A number is made of one or more ASCII digits and a possibly a decimal
+/// point in the middle.
```
mkorje: This isn't entirely accurate, it's really whether it has one of the general categories for numbers (https://doc.rust-lang.org/std/primitive.char.html#method.is_numeric).

crates/typst-layout/src/math/text.rs
```diff
@@ -68,56 +68,74 @@ fn layout_inline_text(
...
+        // than a single char after styling.
+        let style = MathStyle::select(unstyled_c, variant, bold, italic);
+        let c = to_style(unstyled_c, style).next().unwrap();
+
+        // This won't panic as ASCII digits and '.' will never end up as
+        // nothing after shaping.
+        let mut glyph = GlyphFragment::new_char(ctx, styles, c, span).unwrap();
+        if c == ',' {
+            glyph.class = MathClass::Normal;
         }
```
mkorje, line 133: This doesn't feel like a great way to do this, though I don't have a solution right now. 

I was also a bit concerned about whether this is missing some chars, but comparing https://en.wikipedia.org/wiki/Decimal_separator#Unicode_characters to https://www.unicode.org/Public/math/revision-15/MathClassEx-15.html the only other one is fullstop `.`, but we override the math class of that to normal atm.
EDIT: U+00B7 · MIDDLE DOT might need to be overridden from binary.

tests/suite/math/attach.typ
```diff
@@ -165,7 +165,7 @@ $ x_1^i x_2^lambda x_2^(2alpha) x_2^(k+1) x_2^(-p_(-1)) x_j^gamma \
 $ attach(a^b, b: c) quad
   attach(attach(attach(attach(attach(attach(sum, tl: 1), t: 2), tr: 3), br: 4), b: 5), bl: 6) $
 
-#let a0 = math.attach(math.alpha, b: [0])
+#let a0 = math.attach(math.alpha, b: [$0$])
```
mkorje:
```suggestion
#let a0 = math.attach(math.alpha, b: $0$)
```

tests/ref/render/enum-numbering-closure-nested-complex.png
mkorje, line 1: The fact that the numbering's numbers are no longer coerced to a math number automatically here might be a negative change? Doesn't really bother me, but thought it could use some discussion.

tests/suite/math/number.typ
```diff
@@ -0,0 +1,21 @@
+// Test numbers in math.
+
+--- math-number-parsed paged ---
+// Test the number is correctly rendered, separately from number-like text.
+#show math.number: set text(red)
+
+#let a0 = math.attach(math.alpha, b: [$0$])
```
mkorje:
```suggestion
#let a0 = math.attach(math.alpha, b: $0$)
```

mkorje (commented):
I left a couple comments, mainly to do with docs.

Maybe `math.num` would be better than `math.number`? It's a bit shorter and `num` is already used by packages like zero.

mkorje:
> I wonder what the consequences would be for numerous packages that generate numbers in equations from strings.

The plan is for `math.number` to translate to `mn` in MathML. The [spec](https://www.w3.org/TR/mathml4/#presm_not_single_mn) gives some details on what should go in an `mn`, and there's also some examples provided [here](https://w3c.github.io/mathml-docs/intent-core-properties/#Units-and-currency). There's also [this old note](https://www.w3.org/Math/Documents/Notes/units.html). So I would imagine number formatting packages should emit an equation to match whatever the translation eventually looks like to MathML. Though I'll defer this to #3269, #1093, and #7436.

crates/typst-library/src/math/number.rs
```diff
@@ -0,0 +1,43 @@
...
+///
+/// If you want to make a string that doesn't fit the above definition
+/// to be rendered like a number, you may use function `number()`.
```
laurmaedje:
```suggestion
/// If you want string that doesn't fit the above definition
/// to be rendered like a number, you can use the `number` function.
```
\+ rewrapping

crates/typst-library/src/math/number.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl NumberElem {
+    /// Create a new packed symbol element.
+    pub fn packed(text: impl Into<EcoString>) -> Content {
+        Self::new(text.into()).pack()
+    }
+}
+
+impl Repr for NumberElem {
+    /// Use a custom repr that matches normal content.
```
laurmaedje: What's the motivation for this?

tests/ref/render/math-number-parsed.png
laurmaedje, line 1: The fact that the page almost overflows is not so nice. Perhaps `#set page(width: auto)`?

tests/suite/math/number.typ
```diff
@@ -0,0 +1,21 @@
...
+#let b0 = math.attach(math.beta, b: [0])
+$
+  "num"  & 2, 2.1, n_2, n/2, mat(1, 2), a0 \
+  "text" & "2"#footnote[footnote], "2.1", n_"2", n/#[2], mat("1", #[2]), b0
+$
+
+--- math-number-coerced paged ---
+// Test the number-like text is rendered like a number.
```
laurmaedje:
```suggestion
// Test that number-like text is rendered like a number.
```

Mc-Zen:
I was worried because packages like Zero generate numeral strings and put them into equations. For this case, it seems to be a breaking change if `$"1.2"$` is now interpreted as text instead of math. Did I understand correctly?

laurmaedje:
> For this case, it seems to be a breaking change if `$"1.2"$` is now interpreted as text instead of math. Did I understand correctly?

It would now be layouted as math text instead of a math number. This trigger slightly different code internally, but the result is very similar. The existing number layout code was originally mostly intended as a fast path to avoid costly inline layout. For numbers, it almost yields the same results as proper inline layout. The only observable differences for existing code should be the ones listed in https://github.com/typst/typst/pull/7577#pullrequestreview-3639418093.

Now, on a semantic level of course packages should migrate to the `math.num` element, but that's more to get proper semantics for new stuff like MathML export, `show math.num`, etc. Existing equation layout should barely be affected as far as I can tell, so as far as breakage goes, it should be very minor.

Mc-Zen:
Thank you, I am reassured. This will give packages time to migrate.

crates/typst-library/src/math/number.rs
Leedehai: Nice catch! Deleted this comment.

crates/typst-layout/src/math/text.rs
Leedehai: I'm not sure what to do, so let me keep the code as is. Feel free to re-open.

tests/ref/render/enum-numbering-closure-nested-complex.png
Leedehai: Yeah, as noted in the PR description. But happy to discuss.

In details, this is because the [`#[parse(...)]`](https://github.com/typst/typst/blob/7eae2ca319b8ae589235bb4a2e08cdb60701af79/crates/typst-library/src/math/matrix.rs#L190-L214) of `MatElem.rows` uses `Value::display()` to convert the arguments, and `Value::display()` converts many non-`Content` arguments (e.g. those from `args.pos()` in this example) to `TextElem`: https://github.com/typst/typst/blob/7eae2ca319b8ae589235bb4a2e08cdb60701af79/crates/typst-library/src/foundations/value.rs#L196-L198 If we want to change this, I think we can add a new method `Value::display_math()` that does this differently and replace all `Value::display()` calls in crates/typst-library/src/math/{matrix.rs, frac.rs}.

To note, if the input of `Value::display()` is already a `Content` (say, a content whose `.func()` equals `math.num`), then the output is unchanged: https://github.com/typst/typst/blob/7eae2ca319b8ae589235bb4a2e08cdb60701af79/crates/typst-library/src/foundations/value.rs#L202 Therefore, the two numbers in the `$ ... mat(1, 2) ... $` call in this PR's test `math-num-parsed` are rendered as `math.num`s.

crates/typst-library/src/math/number.rs
Leedehai: That "above definition" was deleted because ASCII digits are not necessarily the only digit characters (see https://doc.rust-lang.org/std/primitive.char.html#method.is_numeric), so I reworded this line to: _If you want to render a string like a number, you may use function `num`._ (`number` is renamed `num` according to another review comment).

tests/ref/render/math-number-parsed.png
Leedehai: Done.

crates/typst-library/src/math/number.rs
Leedehai: The motivation for the `Rerpr` trait? I see `TextElem`, `SymbolElem`, `SpaceElem` all have `eco_format!("[{}]", self.text)`.

If you're asking about the comment `Use a custom repr that matches normal content.` -- I copied that from `SymbolElem`'s impl, but yeah let me delete this copy since it doesn't make much sense to me.

Leedehai:
Resolved review comments and renamed `math.number` to `math.num` as requested. I also ran `git rebase main` to sync this branch with main.

crates/typst-layout/src/math/text.rs
```diff
@@ -68,57 +68,75 @@ fn layout_inline_text(
...
+    .into_frame();
+
+    Ok(FrameFragment::new(styles, frame)
+        .with_class(MathClass::Alphabetic)
+        .with_text_like(true)
+        .with_spaced(true))
+}
+
+/// Layout a [`NumElem`].
+pub fn layout_number(
```
mkorje: Maybe we call this `layout_num` to match all the other layout functions?

tests/ref/render/enum-numbering-closure-nested-complex.png
mkorje: cc @laurmaedje

crates/typst-library/src/math/number.rs
laurmaedje: "may use function `num`" is still missing an article. "may use the `num` function" would read better.

crates/typst-library/src/math/number.rs
laurmaedje: I think this case is different. SymbolElem has this because it's internal and not really supposed to be observable. SpaceElem was also originally not intended to be publicly exposed and has no binding in std. It's also not as semantic as `NumElem`. For `NumElem`, using the default generated `Repr` would be more appropriate in my opinion.

laurmaedje:
I think this is unfortunate and potentially a blocker:
```typ
#show math.num: set text(red)
$ 12 #12 $
```

12 in red followed by 12 in black


Changing this might turn out rather hacky with the current infrastructure though ...

A long-term potential fix would be a math-specific show rule for integers, floats, and decimals. But this would depend on type/content unification...

tests/ref/render/enum-numbering-closure-nested-complex.png
laurmaedje: If we only add this for mat and frac, this would not handle top-level integers like `$ #1, #2 $`. I'm afraid that turning those into numbers would result in more hacks (see https://github.com/typst/typst/pull/7577#issuecomment-3728016523).

Leedehai:
Yeah, I agree it's a bit unfortunate to not be able to handle `$ #12 $` in the current approach.

Indeed, `#` switches away from math mode to code mode and starts evaluating the expression of `12`, but interpreting the result of this evaluation as `math.num` is the sensible/intuitive expectation, and making it happen is rather hacky. 

Closing this PR for now, feel free to follow up with comments :)

PR #7581, opened 2025-12-14
When testing, don't print ANSI sequence if the output is not TTY

Leedehai:
When the output destination is not a terminal, let's not print the ANSI escape sequence to draw underline `\x1B[4m`/`\x1B[0m`.

|                           |  Before this PR | After this PR |
|:---------------:|----------------|-------------|
|terminal              | | same as left |
|test-helper view|| |


laurmaedje:
Thanks!

PR #7596, opened 2025-12-17
Customize Typst registry URL

podsvirov:
Support for configuring the Typst registry URL from the `TYPST_REGISTRY_URL` environment variable.

Closes: #7602

### DIFF
--- a/crates/typst-kit/src/package.rs
+++ b/crates/typst-kit/src/package.rs
@@ -40,7 +40,7 @@ pub fn default_package_path() -> Option<PathBuf> {
     dirs::data_dir().map(|data_dir| data_dir.join(DEFAULT_PACKAGES_SUBDIR))
 }
 
-/// Attempts to infer the Typst registry url from the current
+/// Attempts to infer the Typst registry URL from the current
 /// environment.
 ///
 /// This simply return [`DEFAULT_REGISTRY`] or value of


crates/typst-kit/src/package.rs
```diff
@@ -39,6 +40,15 @@ pub fn default_package_path() -> Option<PathBuf> {
     dirs::data_dir().map(|data_dir| data_dir.join(DEFAULT_PACKAGES_SUBDIR))
 }
 
+/// Attempts to infer the Typst registry url from the current
```
bastienvoirin:
```suggestion
/// Attempts to infer the Typst registry URL from the current
```

laurmaedje:
Thanks for the PR. We do want to support custom registries and richer configuration down the road, but with a proper comprehensive design rather than with an ad-hoc env var that overrides the meaning of `@preview`.

podsvirov:
> We do want to support custom registries

This PR is not about that. I did not want create custom registry. I want to allow Typst users configure  URL of the  Typst registry mirror.

laurmaedje:
That's also in scope of the design work I'm talking about. I would just prefer to not ship a band-aid design.

podsvirov:
I may have been a bit hasty in suggesting changes, but this solution is very simple and it works well.

Can you provide links to discuss issues or other suggestions for implementing this functionality?

laurmaedje:
There's not a ton of concrete stuff yet, but most of what is there is spread out over the [issues with the `packages` label](https://github.com/typst/typst/issues?q=is%3Aissue%20state%3Aopen%20label%3Apackages) (in particular https://github.com/typst/typst/issues/1982) and the [Packages, Templates forge thread on Discord](https://discord.com/channels/1054443721975922748/1176122103355953162).

As for why not merge the simple solution: It's a general design philosophy of us to avoid shipping a band-aid fix in an area where we want to have richer functionality down the road. In my experience, it tends to invite many more small changes, which accumulate to an inconsistent design while also reducing the urgency to design a comprehensive solution (for mirrors / custom non-preview registries, Git dependencies, etc.)

podsvirov:
Thanks for the links and additional description of the principles of this project.

I hope that waiting of the functionality that we have been talking about will not take very long. Judging by #1982, users already  waiting for this for more than two years.

PR #7597, opened 2025-12-18
Bump New CM fonts to version 7.1.0

### LINKED ISSUE #1528: More consistent empty set signs for default font
### Description

See discussion in #1526.

As @Enivex mentioned,

 > the correct way of dealing with this is using the cv01 font feature, as the font provides an alternative version of the symbol. Though I do not believe this feature is currently accessible in Typst.

So the idea is fixing default math font with font features instead of directly swapping the symbols for all the fonts.

### Use Case

That issue (need to use `diameter` instead of `nothing` for default math font) might be really confusing for newcomers, so it should be at least mentioned somewhere until fixed.

### LINKED ISSUE #7322: Some integral signs have improper placing of integral bounds
### Description

So far, I have found two integral signs that exhibit this behaviour, namely:

| Name | Unicode | Typst symbol representation |
| :--- | :---: | :---: |
| Integral around a point operator | [2A15](https://www.compart.com/en/unicode/U+2A15) | - |
| Integral with double stroke | [2A0E](https://www.compart.com/en/unicode/U+2A0E) | `integral.dash.double` |

 MWE:
```typst
#let pointint = symbol("\u{2A15}")

$
  integral^1_0 quad "cf." quad integral.dash.double^1_0 pointint^1_0
$

$
  f(a) = 1 / (2 π i ) pointint_gamma f(z) / (z - a) dif z
$

$
  integral.dash.double_(C_1) g(z) dif z + integral.dash.double_(C_2) g(z) dif z + ...
$
```

Image

While one may consider the assertion of `class("large" ... )` this also yields a behaviour inconsistent with the default `integral` symbol. MWE:
```typst
#let pointint = math.class("large", symbol("\u{2A15}"))

$
  integral^1_0 quad "cf." quad pointint^1_0
$

$
  f(a) = 1 / (2 π i ) pointint_gamma f(z) / (z - a) dif z
$
```

Image

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

mkorje:
~~Depends on https://github.com/typst/typst-assets/pull/14.~~
Closes #7322 and closes #1528.

The updated test is due to the default `emptyset` glyph being swapped with the one in `cv01`.
| Old | New |
| - | - |
| old | new |


Enivex:
Strangely the change to the empty set glyph is not mentioned in the change log.

laurmaedje:
Thanks!

PR #7600, opened 2025-12-18
Nix flake build fixes

NyCodeGHG:
I wanted to try out the development version of typst and noticed that the package provided in the flake didn't build.

It looks like max was accidentally used instead of min in https://github.com/typst/typst/pull/7338, which made the build fail.
Since the flake is not using nixpkgs' rust compiler, but one from the rust project, [it defaults to linking using rust-lld since Rust 1.90.0](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/), but nixpkgs requires use of a wrapped linker, otherwise rpaths are not set and it leads to problems like `error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory`.
#7374 is the wrong solution to this problem, there's no need for `LD_LIBRARY_PATH`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -79,6 +79,15 @@
               pkgs.pkg-config
               pkgs.openssl.dev
             ];
+
+            env = {
+              RUSTFLAGS =
+                if pkgs.stdenv.hostPlatform.rust.rustcTargetSpec == "x86_64-unknown-linux-gnu" then
+                  # Upstream defaults to lld on x86_64-unknown-linux-gnu, we need to use the system linker
+                  "-Clinker-features=-lld -Clink-self-contained=-linker"
+                else
+                  null;
+            };
           };
 
           # Derivation with just the dependencies, so we don't have to keep


PartyWumpus:
have a look at https://github.com/typst/typst/pull/7512

laurmaedje:
Thanks for catching the bug. I'd be happy to merge the min/max mixup fix in a separate PR.

NyCodeGHG:
Oh, I see.
I opened a new PR: https://github.com/typst/typst/pull/7601

PgBiel:
@NyCodeGHG I'd be happy to accept your contribution at the flake's new home, at https://github.com/typst/typst-flake. The repository is still on the process of being created, but I'll be merging my own PR soon (you can also make a PR to my branch if you want).

In particular, it seems that this would solve https://github.com/typst/typst-flake/pull/1#discussion_r2599040253.

PR #7603, opened 2025-12-18
Add Galician smartquotes

deivisCode:


laurmaedje:
Somewhat hard to judge for me, but it seems to match what's said on https://gl.wikipedia.org/wiki/Comi%C3%B1as. Thanks!

PR #7605, opened 2025-12-19
Make `math.lr.size` argument relative only to the height of the wrapped content

mkorje:
From working on #7435, I found it a little bit weird that the delimiters are taken into account when calculating the "relative to" size used in `math.lr` (i.e. the size of the body). This PR makes it so that they are not included.

In tests where the body content is smaller than the default (i.e. non-stretched) size of the delimiter, I've had to increase the relative lengths so that they still pass.

The updated test image for `math-lr-nested` reflects that, since there is no body, relative lengths are just relative to 0. I considered carving out a special case for this situation, so that the layout ends up the same as `math.stretch`. But I didn't like the inconsistency this would cause, and you can always just use (and should prefer, IMO, as it is more "correct") `math.stretch` if you want this behaviour.

I realise though that this change might, to some, make things less intuitive. For example, relative lengths in nested `math.lr` calls won't "stack" like before: `$ lr(lr((a/b), size: #4em), size: #50%) $` ≠ `$ lr((a/b), size: #2em) $`. However, again, I didn't really like the idea of trying to "fold" the sizes, since it would change the behaviour of `math.lr` depending on its nesting. I'd appreciate any thoughts/comments/concerns anyone has.


laurmaedje:
I think this makes sense as it brings the reality in line with what's documented. I could see an argument to be made for _always_ interpreting the ratio relative to the base delimiter size rather than body content, but the current implementation just seems inconsistent, so I think this is a good change.

As for nested L/R: It's just generally quite weird and somewhat buggy (in particular https://github.com/typst/typst/issues/6940). It should probably just not be supported at all (e.g. @isuffix's proposal would be to pass the delimiters as separate arguments, which would alleviate the issue), but that's not for this PR to sort out. The one thing is that the current behavior is emergent rather than designed, the other is that it's not clear whether it is even desirable. I thus think the change in behavior here is defensible. How that fits in with `stretch` is another question, but I'd say also to be sorted out separately.

Thanks!

PR #7606, opened 2025-12-19
Fix uneven CJK-Latin spacing in justified paragraphs

### LINKED ISSUE #6062: CJK-latin glues stretch only before latin characters
### Description

When paragraphs are justified, glues between CJK and latin characters stretch only before latin characters. They should equally stretch.

```
#set par(justify: true)
#box(width:4em)[第1回だ。]
```

Image

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

ryuryu-ymj:
Fixes #6062

This PR ensures that CJK-Latin spacing is applied evenly both before and after Latin glyphs when par is justified.

Previously, Latin glyphs followed by CJK characters were not marked as justifiable, leading to uneven spacing distribution.

## Changes
- Updated the `add_cjk_latin_spacing` function to set `is_justifiable` to `true` for Latin glyphs that are followed by CJ characters.

## Testing
```typst
#set par(justify: true)
あaあ#linebreak(justify: true)
ああaa aaああ#linebreak(justify: true)
```

**before**
before

**after**
after


laurmaedje:
Thanks!

For the record, my understanding of what's happening here: The `is_justifiable` property only affects whether spacing can be inserted to the right of the glyph if the line is underfull (i.e. its stretchability does not suffice to fully justify). It might be clearer if it was called "can_justify_after" or something like that, but that's unrelated to this PR.

Extra available spacing is distributed to the right of all justifiable glyphs. Justifiable glyphs include spaces, CJ glyphs, and CJK punctuation. In Latin text, the extra spacing is thus distributed among spaces, and in CJ text, it is distributed to the right of all letters since spaces aren't usually used. However, what we really want is to distribute spacing _around_ the inner CJK glyphs. When two CJK glyphs are next to each other, this is naturally the case since the previous one will get spacing to the right. But for Latin followed by CJK, it is not, and that's what this patch changes.

The implementation seems reasonable to me as it fixes the problem and I don't see any major potential for issues arising from it (it only affects the code path where we overstretch the line and only allows spacing between Latin and CJK, which is precisely what we want). However, I think the overall implementation (not related to this PR), could probably be improved to make this a bit clearer (cf. also https://github.com/typst/typst/pull/7521#issuecomment-3720703364).

Issue #5454, opened 2024-11-23
Security: Restrict Arbitrary File Reading Part 2


ally-petitt:
## Background

It is known that previous versions of `typst` allowed for reading arbitrary files as discussed in issue [#219](https://github.com/typst/typst/issues/219), revealing potentially sensitive data in the filesystem. Below is the code snippet used to illustrate this.

```
# adjust for the right amount of ..
echo '#read("../../../etc/passwd")' >repro.typ
typst repro.typ
```

This issue was closed after @laurmaedje [commented](https://github.com/typst/typst/issues/219#issuecomment-1623202981_) that doing so was now forbidden.
            
 > I'm closing this as for now as reading outside of the directory the Typst file resides in is forbidden now.

## The Bypass
There is a way to circumvent this protection in Typst version 0.12.0 through the use of symlinks. Creating a file `passwd` that symlinks to `/etc/passwd` and reading it from the project directory `/tmp` will result in reading `/etc/passwd`, despite it not being within the project root.

As a proof of concept, I have created `passwd` and `main.typ` in the `/tmp` directory, which will be the project root for this example.
```
virusspec@4eff4abe169b:/tmp$ ls -l
lrwxrwxrwx 1 virusspec virusspec     9 Nov 22 21:23 passwd -> /etc/passwd
-rw-rw-r-- 1 virusspec virusspec    41 Nov 22 21:26 main.typ
```

The contents of `main.typ` are below.
```
#let text = read("passwd")
#raw(text)
```

Executing `typst compile --root /tmp main.typ` results in a PDF file that contains the contents of `/etc/passwd`. 

It can be verified that this was not intentional by changing `main.typ` to read from `/etc/passwd` directly and observing the subsequent error message asserting that reading outside of the project root is prohibited.

```
virusspec@4eff4abe169b:/tmp$ typst compile --root /tmp test.typ
error: failed to load file (access denied)
  ┌─ test.typ:1:17
  │
1 │ #let text = read("../etc/passwd")
  │                  ^^^^^^^^^^^^^^^
  │
  = hint: cannot read file outside of project root
  = hint: you can adjust the project root with the --root argument
```

## Remediation
This vulnerability can be mitigated by checking that the resolved symlink path is within the project root rather than relying entirely on the filename supplied by the user.

In bash, this can be done with `readlink` and it appears that there is a [Rust equivalent](https://doc.rust-lang.org/std/fs/fn.read_link.html) `std::fs::read_link`.

### Additional Notes
[...]

PR #7608, opened 2025-12-19
fix: prevent symlink path traversal in file access

Ayanprogrammer11:
## Summary
This PR fixes a symlink path traversal vulnerability in the CLI file access mechanism.
## Problem
The `VirtualPath::resolve()` function correctly blocks lexical path traversal (`../`), but as noted in the code comments, paths could still escape the project root through symbolic links.
**Attack scenario:**
1. Attacker creates symlink: `ln -s /etc/passwd secret.txt`
2. Typst document reads: `#read("secret.txt")`
3. Result: Contents of `/etc/passwd` are exposed.
## Solution
Added canonicalization check in `system_path()` that:
1. Canonicalizes both the resolved path and project root.
2. Verifies the resolved path is still within the root.
3. Returns `AccessDenied` if a symlink escapes the boundary.
## Testing
- All existing CLI tests pass!
- Manual verification confirms symlink attacks are blocked.

### DIFF
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -423,9 +423,19 @@ fn system_path(
         root = &buf;
     }
 
-    // Join the path to the root. If it tries to escape, deny
-    // access. Note: It can still escape via symlinks.
-    id.vpath().resolve(root).ok_or(FileError::AccessDenied)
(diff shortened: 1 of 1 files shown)


laurmaedje:
This is known and, barring further decisions, intentional behavior. See https://github.com/typst/typst/issues/5454#issuecomment-2496006467.

PR #7609, opened 2025-12-20
Small refactor for space collapsing

isuffix:
While working on #7350, I realized there was an opportunity for a nice refactor independent of adding any features. So I'm making this PR, on which #7350 will depend.

This adds a new state `Invisible` that allows us to centralize the logic in `find_regex_match_in_elems` and make if read similarly to the logic in `collapse_spaces`.

I split out the updated `SpaceState` and state functions into `space_collapse.rs` because it aids readability to open it side-by-side with the loops in `lib.rs` and makes it easier to compare the state functions. I wasn't sure whether to also move the `collapse_spaces` function over, so I left that in `lib.rs`

This also adds a test case (`show-text-styled-space`) that confused me for a while because I was only thinking about collapsing not grouping. I find the space behavior in grouping rules very difficult to reason about, and if you agree I'd like to work on changing it. But I'll leave that for later.

### DIFF
--- a/crates/typst-realize/src/lib.rs
+++ b/crates/typst-realize/src/lib.rs
@@ -35,6 +35,9 @@ use typst_library::text::{LinebreakElem, SmartQuoteElem, SpaceElem, TextElem};
 use typst_syntax::Span;
 use typst_utils::{ListSet, SliceExt, SmallBitSet};
 
+mod spaces;
+use spaces::{SpaceState, collapse_spaces, collapse_state_textual};
+
 /// Realize content into a flat list of well-known, styled items.
 #[typst_macros::time(name = "realize")]
 pub fn realize<'a>(
@@ -177,17 +180,6 @@ struct RegexMatch<'a> {
     id: RecipeIndex,
     /// The recipe that matched.
     recipe: &'a Recipe,
-}
-
-/// State kept for space collapsing.
-#[derive(Debug, Copy, Clone, Eq, PartialEq)]
-enum SpaceState {
-    /// A following space will be collapsed.
-    Destructive,
-    /// A following space will be kept unless a destructive element follows.
-    Supportive,
-    /// A space exists at this index.
-    Space(usize),
 }
 
 impl<'a> State<'a, '_, '_, '_> {
@@ -1153,10 +1145,11 @@ fn visit_textual(s: &mut State, start: usize) -> SourceResult<bool> {
 /// elements.
 ///
 /// Collects the element's merged textual representation into the bump arena.
+///
 /// This merging also takes into account space collapsing so that we don't need
 /// to call `collapse_spaces` on every textual group, performing yet another
-/// linear pass. We only collapse the spaces elements themselves on the cold
-/// path where there is an actual match.
+/// linear pass. We only collapse the space elements on the cold path when there
+/// is an actual match.
 fn find_regex_match_in_elems<'a>(
     s: &State,
     elems: &[Pair<'a>],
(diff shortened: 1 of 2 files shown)


crates/typst-realize/src/space_collapse.rs
```diff
@@ -0,0 +1,62 @@
...
+        SpaceState::Supportive
+    }
+}
+
+/// Space collapsing state for textual elements used during regex matching.
+pub(crate) fn collapse_state_textual<'a>(
+    content: &'a Content,
+    styles: StyleChain<'_>,
+) -> (SpaceState, &'a str) {
+    // Roughly ordered from most to least common.
```
laurmaedje: unless this provably has an effect on performance, I'd prefer having these in the same order in both functions, for clarity.

crates/typst-realize/src/space_collapse.rs
laurmaedje, line 1: I kinda dislike underscores in Rust file names. I'd prefer either `collapse.rs` or `space.rs`, probably the former.

crates/typst-realize/src/lib.rs
```diff
@@ -1367,6 +1353,7 @@ fn visit_regex_match<'a>(
 /// vicinity of destructive elements.
 fn collapse_spaces(buf: &mut Vec<Pair>, start: usize) {
```
laurmaedje: I think it would indeed make sense to move this as well since it was literally the name of the module.

crates/typst-realize/src/lib.rs
```diff
@@ -1163,18 +1156,28 @@ fn find_regex_match_in_elems<'a>(
...
+            SpaceState::Destructive => {
+                if state == SpaceState::Space {
+                    buf.pop();
+                }
+                SpaceState::Destructive
+            }
+            SpaceState::Supportive => SpaceState::Supportive,
+            SpaceState::Space => {
+                if state != SpaceState::Supportive {
+                    continue;
```
laurmaedje, line 1176: This and the `SpaceState::Invisible` branch now continue before the `if styles != current && !buf.is_empty() {` business as opposed to continuing afterwards previously. I'm not sure whether that could have any unintended consequences, but might be worth double checking if it's not intentional.

crates/typst-realize/src/lib.rs
```diff
@@ -4,6 +4,10 @@
 //! particular, show rules to produce well-known elements that can be processed
 //! further.
 
+mod space_collapse;
+
+use self::space_collapse::{SpaceState, collapse_state, collapse_state_textual};
```
laurmaedje: could you move self imports below external crate imports into a separate group?

crates/typst-realize/src/space_collapse.rs
isuffix: I'm going with `spaces.rs`

crates/typst-realize/src/lib.rs
isuffix: This change also gave me a lot of grief reasoning through it, but I think it's correct.

First, the `Invisible` case already continued at the beginning of the loop, so nothing's changed there. I'm not certain, but I believe tag elements can never have different styles than the element they precede, so this should never trigger the if-statement on its own. Another non-invisible element with different styles would need to be present, and that element's presence would trigger the if-statement itself.

Second, spaces will only `continue` if they're being collapsed or discarded. If it's being discarded, that space shouldn't affect regex searching, and if it's being collapsed, its styles don't matter because we keep the styles of the first space in a run.

Also, I'm quite unsure, but I believe that interior spaces with different styles than adjacent text aren't ever included in a group passed to `find_regex_match_in_elems` because of the general text grouping rules, but I don't have a strong-enough understanding of those rules to be certain. This is part of what the `show-text-styled-space` test is for.

laurmaedje: I see. The behavior tested by the two new tests is indeed confusing and I would consider it a bug.

laurmaedje:
Thanks!

PR #7614, opened 2025-12-21
Fix closing curves in smooth mode.

Emm54321:
Closing a curve in smooth mode when the end point already is at the start point can cause ugly glitches and is useless anyway.

### DIFF
--- a/crates/typst-layout/src/shapes.rs
+++ b/crates/typst-layout/src/shapes.rs
@@ -333,15 +333,35 @@ impl<'a> CurveBuilder<'a> {
     /// Close the curve if it was opened.
     fn close(&mut self, mode: CloseMode) {
         if self.is_started && !self.is_empty {
-            if mode == CloseMode::Smooth
-                && (!self.last_point.x.approx_eq(self.start_point.x)
-                    || !self.last_point.y.approx_eq(self.start_point.y))
-            {
-                self.cubic(
-                    self.last_control_from,
-                    self.start_control_into,
-                    self.start_point,
-                );
+            if mode == CloseMode::Smooth {
+                fn is_closed(p: Point, q: Point) -> bool {
+                    p.x.approx_eq(q.x) && p.y.approx_eq(q.y)
+                }
+                fn is_aligned(a: Point, b: Point, c: Point) -> bool {
+                    let u = a - b;
+                    let v = c - b;
+                    let ux = u.x.to_raw();
+                    let uy = u.y.to_raw();
+                    let vx = v.x.to_raw();
+                    let vy = v.y.to_raw();
+                    let det = ux * vy - uy * vx;
+                    let nu = ux * ux + uy * uy;
+                    let nv = vx * vx + vy * vy;
+                    det * det <= 1e-8 * nu * nv
+                }
+                if !is_closed(self.last_point, self.start_point)
+                    || !is_aligned(
+                        self.start_control_into,
+                        self.last_point,
+                        self.last_control_from,
+                    )
+                {
+                    self.cubic(
+                        self.last_control_from,
+                        self.start_control_into,
+                        self.start_point,
+                    );
+                }
             }
             self.curve.close();
             self.last_point = self.start_point;


Enivex:
Maybe I'm missing something, but wouldn't the correct way to handle this be to still use close mode smooth, but disregard the last point? Or warn that the user is doing something wrong.

Otherwise the end result is not going to look smooth.

Emm54321:
If the tangents don't match, the path won't indeed look smooth, but the user probably doesn't expect a loop to be added in this case. This could indeed be considered a user error. However, a more frequent case is when the tangents match but a small flat loop is added (e.g. if you approximate a circle with bezier curves).

Enivex:
> If the tangents don't match, the path won't indeed look smooth, but the user probably doesn't expect a loop to be added in this case. 

That's the only way to get a smooth curve. If I specify a smooth closed curve then the result should be smooth.

>  However, a more frequent case is when the tangents match but a small flat loop is added (e.g. if you approximate a circle with bezier curves).

Isn't it a user error to specify the end point in this case?

In any case, if handling this is desirable then the if statement should check for whether both the points _and_ tangents match.

(I really think it's ultimately user error though. Garbage in, garbage out.)

Emm54321:
Yes, but smooth is the default. And if your curve is already closed and smooth, you don't expect a curve.close() element to add a glitch at the closing point, and you may not realize why your curve doesn't look like what you expect. IMHO, the smooth closing was there to allow avoiding to put the last curve part if it could be interpolated, not to add a loop when there is an angle by design.

Emm54321:
Adding tangent checks could indeed be better.

Enivex:
>  not to add a loop when there is an angle by design.

If there is an angle, then the curve is not smooth.

Emm54321:
Well, the close element is there to add a path to the starting point. And this path can be either a line or a curve, but one can argue that adding this path can be avoided if the path is already closed. Adding a loop when "smooth" is explicitly specified would make sense, but I don't think it does if it is not. And besides, those loops when there is an angle are usually ugly.

Enivex:
> Well, the close element is there to add a path to the starting point. And this path can be either a line or a curve, but one can argue that adding this path can be avoided if the path is already closed.

It cannot be avoided if you want the curve to be smooth (as a closed curve).

> Adding a loop when "smooth" is explicitly specified would make sense, but I don't think it does if it is not.

Then choose "straight" instead of "smooth"?

> And besides, those loops when there is an angle are usually ugly.

Not if that is what you actually want

Emm54321:
The main problem is that if you design a nice smooth closed curve, you have to specify curve.close(mode: "straight") for it not to look bad, which is counter-intuitive and you may struggle to understand what is wrong. OTOH, if your closed curve is designed with an angle, there is nothing surprising about it not being smooth, and as you say it's garbage in, garbage out. But checking for matching tangents in the added check is easy.

Emm54321:
I've added a check for the tangents, but I still think that adding and extra curve when the curve is already closed doesn't make much sense and is probably not expected. Maybe it should just be documented.

```typst
// Glitch at the top without the fix.
#curve(
  curve.quad((1cm, 0cm), (1cm, 1cm)),
  curve.quad((1cm, 2cm), (0cm, 2cm)),
  curve.quad((-1cm, 2cm), (-1cm, 1cm)),
  curve.quad((-1cm, 0cm), (0cm, 0cm)),
  curve.close(),
)

#curve(
  curve.quad((1cm, 0cm), (1cm, 1cm)),
  curve.quad((0cm, 1cm), (0cm, 0cm)),
  curve.close(mode: "straight"),
)

// Do we really want the small loop added?
#curve(
  curve.quad((1cm, 0cm), (1cm, 1cm)),
  curve.quad((0cm, 1cm), (0cm, 0cm)),
  curve.close(mode: "smooth"),
)
```

eltos:
Idea: Change the default closing mode to `auto`, which uses "straight" (or essentially nothing) if the final point already coincides with the start point, and "smooth" otherwise. This would avoid the unexpected behaviour of the extra curve, while still allowing it in case mode is explicitly set to "smooth".

Enivex:
I honestly really don't think there is a problem to solve here, other than perhaps better documentation.

> The main problem is that if you design a nice smooth closed curve, you have to specify curve.close(mode: "straight") for it not to look bad

Only if you explicitly the last curve segment to the end point.

Emm54321:
> > The main problem is that if you design a nice smooth closed curve, you have to specify curve.close(mode: "straight") for it not to look bad
> 
> Only if you explicitly the last curve segment to the end point.

Yes, but :
- you have to do that if you don't want mirrored control points.
- closing the curve by yourself is quite natural if you are used to other similar curve building patterns since they usually only close curves by a straight line.
- you really think that it's ok  for : 
  ```typst
  #curve(
    curve.quad((1cm, 0cm), (1cm, 1cm)),
    curve.quad((1cm, 2cm), (0cm, 2cm)),
    curve.quad((-1cm, 2cm), (-1cm, 1cm)),
    curve.quad((-1cm, 0cm), (0cm, 0cm)),
    curve.close(),
  )
  ```
  to output garbage, and that people have to guess that either adding
  `mode: "straight"` or removing the last `quad` is the way to fix that?
  I mean, I was confused by a similar situation event though I wrote the curve code
  and knew how it works.

Either the `auto` suggestion or having `close` not adding anything when the curve is already closed (and documenting that) seems much less confusing.

Enivex:
I think the status quo is fine, yes. With perhaps the exception of special casing the situation where the start and end point are the same and the tangents also point in the same direction. (Since floating point error will lead to trouble).

laurmaedje:
To me, the `curve` is a low-level primitive that should have a simple, well-defined contract that maps cleanly to the underlying constructs. Smartness based on coinciding points or tangents seems inappropriate to me. It feels like turning a precise, well-defined API into a "let's paper over the user's impreciseness" type of API.

If anything, I think maybe the mistake was to default to smooth mode as it doesn't directly map to underlying drawing functionality in export targets and is unlike what other software does.

We could still change that, though if we want to do that I would delay that until after the compatibilty mechanism described in https://laurmaedje.github.io/posts/evolving-typst/ has landed.

Emm54321:
Changing the default would may indeed be the best solution.

laurmaedje:
I'll close this then. Let's revisit changing the default once the compatibility mechanism is in place. That shouldn't be too far away.

PR #7616, opened 2025-12-22
Add horizontal-rule element

### LINKED ISSUE #5318: Support horizontal rules
### Description

Typst should support a horizontal rule type, similar to markdown and HTML. While the output can be approximated with a line, it really should be a semantic type that can be styled appropriately.

I am happy to put up a PR for this if someone will point me in the right direction.

### Use Case

I want to use something like `#horizontal_rule()` so that the semantics are decoupled from display.

LegNeato:
Supports stroke customization via the `stroke` parameter and can be styled using set/show rules.

Closes #5318


linux-user36:
Maybe it should be called `thematic-break` instead, both to be more neutral towards potential vertical rules in vertical text (should that become supported), and because it needn't necessarily be a straight line: In my experience, plenty of prose books instead use asterisms, fleurons, or similar typographic marks. The HTML tag's official name has also been changed from "horizontal rule" to "thematic break", as far as I'm aware.

LegNeato:
The styling and rendering currently requires it to be a line, but perhaps it might be better as a block element so you can put whatever you want in it and style the block however.

I'll wait for a maintainer to chime in.

Enivex:
`separator` would be less of a mouthful than `thematic-break`

laurmaedje:
I absolutely agree that the element should be more neutral and semantic in its definition. I also agree that `thematic-break` is a bit of a mouthful. I think `separator` is not so bad, but I would like to reserve it for a potential future element that forms a composable building block used in place of some existing `separator` concepts (`figure.separator`, `footnote.separator`, `terms.separator`).

I would propose `divider`. I think it's similarly neutral as `separator`, and it is less in conflict with existing and planned Typst features. (To me it even feels a tad more semantically fitting but that's just vibes-based at this point.)

As for the interface: The `divider` would not be coupled to being a line and would not have a `stroke` field. Rather, it's default show rule would simply be a line. I would propose the following default styles:

```typ
#show divider: set block(spacing: 2em)
#show divider: line(length: 100%, stroke: 0.05em)
```

The block spacing rule gives the divider a bit of room to breath, while the default stroke is less obnoxious than 1pt (it matches the `math.cancel` stroke). There was an attempt to express more defaults in em, so this is a good place to continue with that.

Implementation-wise, this would not need any new block layout routine, but merely a default show rule (in `typst-layout/src/rules.rs`) and a `ShowSet` impl for `DividerElem` (in `typst-library/src/model/divider.rs`). It should also be added to `typst-html/src/rules.rs` with a mapping to `<hr>`.

laurmaedje:
@LegNeato are you still interested in working on this?

laurmaedje:
I'll close this due to inactivity, but if someone else wants to move this forward, I'd be happy to accept a PR for the `divider` design proposed above.

Issue #2967, opened 2023-12-15
Adjacent Footnotes Should be Separated by a Comma


adamkrellenstein:
### Description

This typst code: `Foobar#footnote[baz]#footnote[qux]`

Produces this result:
![Screenshot 2023-12-14 at 8 39 29 PM](https://github.com/typst/typst/assets/6189829/f20a869f-2078-42e8-8c25-b6982c0c76dd)

There should be a comma between the `1` and the `2`, at least optionally. If there's already a way to make that happen, I wasn't able to find it.

### Use Case

This would be useful whenever the author has multiple footnotes for the same content.

Leedehai:
LaTeX also has the [same issue](https://tex.stackexchange.com/questions/28465/multiple-footnotes-at-one-point), and a temporary fix could be
```typ
Foobar#footnote[content]$""^,$#footnote[content].
```
<img width="97" alt="Screenshot 2023-12-14 at 21 53 10" src="https://github.com/typst/typst/assets/18319900/3d3be4bb-da44-4ff1-89ee-88c1ea8c57fe">

Indeed a very useful feature request. Typst doesn't have to stop where LaTeX is.

MDLC01:
I think a better temporary fix would be
```typ
Foobar#footnote[baz]#super[,]#footnote[qux]
```
or, if you want to force a space.
```typ
Foobar#footnote[baz]#super[,#" "]#footnote[qux]
```

This is better because it does not use the math font. It also seems more semantically correct to me.

Marcono1234:
Could a potential solution be to make the `body` argument of `footnote` an argument sink (that is, allow >= 1 values)? If more than one footnote is specified, Typst could then insert a separating `,` (maybe configurable?). To me that seems more reliable than having to implement special casing for footnotes to inspect their surrounding content.
For example something like `#footnote([baz], [qux])`.


Here is a workaround function for now (based on the previous comments):
```typst
// Create multiple footnotes, separated by comma
#let footnotes(note, ..additional-notes, separator: ",") = {
  footnote(note)

  for note in additional-notes.pos() {
    super(separator)
    footnote(note)
  }
}


some text #footnotes("first", [second], "third")
```

MDLC01:
For reference, [`cite`](https://typst.app/docs/reference/model/cite/) already does something similar with adjacent citations

laurmaedje:
[...]

PR #7617, opened 2025-12-22
Preview pdftags tests' YAML output in the test-helper extension

Leedehai:
This PR allows previewing the output and reference of tests with attribute`pdftags` (e.g. `footnote-tags-basic`) in the test-helper extension.

Such result files are YAML files, so essentially this PR adds the support of previewing YAML files.

Screenshot 2025-12-21 at 19 49 46

---

Also, from a better web design, this PR adds a small `5px` vertical padding to the body of the web view, so that there will be a small gap between the end of previewing box and the bottom of the web view (see below), so that readers know there's no more to scroll down.
Screenshot 2025-12-21 at 19 55 25

### DIFF
--- a/tools/test-helper/src/extension.ts
+++ b/tools/test-helper/src/extension.ts
@@ -407,7 +407,7 @@ enum Format {
 const FORMAT_TO_FILE_EXTENSION = {
   [Format.HTML]: "html",
   [Format.RENDER]: "png",
-  [Format.PDFTAGS]: "yml"
+  [Format.PDFTAGS]: "yml",
 };
 
 function getUri(name: string, bucket: Bucket, format: Format) {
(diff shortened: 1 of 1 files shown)


Leedehai:
I found this useful with the my investigation into the tests related to footnotes in issue #2967 :)

saecki:
Looks good! I just wanted to jump in and say that I'm currently working on a test diffing tool, along with testing SVG and PDF output, which can also produce text diffs. Since it basically produces static HTML, we can probably reuse parts of it in the test helper extension. Just so we don't start working on the same thing in parallel :)

Here's an example of a [report.html](https://github.com/user-attachments/files/24290225/report.html)

Leedehai:
Oh, I was wondering if we could merge this PR first? @saecki  I'd really appreciate the convenience with the YAML outputs :) (I'm working on #7627, for example)

laurmaedje:
Thank you, I think the two of us are the only people that use the test helper :D

PR #7454, opened 2025-11-24
Clamp PDF page size to 3x3 units

saecki:
Fix #7292 by clamping the page size to 3x3 units. This is the minimum required until PDF 1.7, PDF 2.0 lifts the requirement.
This is similar to how `typst-render` clamps the image size to 1x1 pixels.

PR #7619, opened 2025-12-22
Clamp SVG size to 1x1pt

saecki:
Complementing #7454. While there SVGs with a width and height of 0 are legal, I ran into similar issues with different SVG viewers and `resvg`.

### DIFF
--- a/crates/typst-svg/src/lib.rs
+++ b/crates/typst-svg/src/lib.rs
@@ -231,6 +231,10 @@ impl<'a> SVGRenderer<'a> {
         size: Size,
         write_custom_attrs: impl FnOnce(&mut XmlWriter),
     ) {
+        // Clamp the size of SVGs to at least one pt. resvg and probably also
+        // other SVG parsers don't handle SVGs with 0 sized dimensions.
+        let size = size.max(Size::splat(Abs::pt(1.0)));
+
         self.xml.start_element("svg");
         write_custom_attrs(&mut self.xml);
         self.xml.write_attribute_fmt(


laurmaedje:
If they are technically legal, should we really do this? For PNG and PDF, we do it because zero sized ones are illegal. (At least I think that's also the case for PNG.)

saecki:
In PDF 2.0 the explicit restriction was removed too, but since krilla and also some other software doesn't handle zero sized pages in PDF well, we still added the limitation.
I think with SVG there is an argument to be made, since you could have `overflow` set to visible.
This however brings us in the weird position that some SVGs produced by typst can not be included since `resvg` rejects them.

reknih (commented):
I would not clamp to 1x1 for SVGs for which either side is actually zero units wide. I don't think that would be the intent of the author: My intent for writing zero on either axis would be to make a file that does not "print" anything or affect spacing along the axis where I put the zero. e.g., I could try to insert 300px worth of horizontal space by inserting a 300x0 SVG. Alternatively, I could want to use a 0x0 SVG to insert the space that goes above and below an image without actually inserting an image.

Now, I get that all of this is cursed and we don't want to support it. But I would just outright say that instead of silently changing the nature of the image against the intent of the user (i.e. print an error message).

The story is different for SVGs with a non-zero but smaller than one size on each axis. Here, the difference incurred by clamping is quantitative, not qualitative. Hence, I think it's acceptable to clamp to 1 for these files if necessary.

PR #7620, opened 2025-12-22
String interpolation

tingerrr:
This is a draft, I only opened this because it's easier to check out this way. The exact syntax and semantics for non-string values will be discussed in the syntax forge.


laurmaedje:
@tingerrr I didn't leave any reviews thus far since this is still marked as draft. Do you still have plans to work on this? If yes, we can keep it open, but I'm going through older PRs trying to establish which ones may actually still move forward.

tingerrr:
Yeah, I do plan to finish this, but if somebody else wants to base their work of this and finish it, I wouldn't mind either.

tingerrr:
I think this can be reviewed now. As always, I suggest going commit by commit, note that some of the commits in this branch are actually part of #7800, so you may want to look at this first.

If any of the commit messages are unclear let me know, but I hope they explain my reasoning behind the structure of this patch.

PR #7627, opened 2025-12-26
Separate adjacent footnotes with comma

### LINKED ISSUE #2967: Adjacent Footnotes Should be Separated by a Comma

Leedehai:
Fixes #2967.

This is done following an approach similar to `CiteGroup`: create a `FootnoteGroup` that collects `FootnoteElem`s (i.e. the numbers in the superscript positions) into a vector, and realize them at once in `Packed<FootnoteGroup>::realize()` with the intra-number separator added.

As an intermediate step, in order to ensure correctness, I ran `cargo testit` with an empty string
```rs
// in method Packed<FootnoteGroup>::realize()
let separator = TextElem::new("".into()).pack();
```
as the intra-number separator. All `paged` tests are passing, meaning the visual output remains the same, as expected.

Now, with the configurable separator, we can do:
```typst
// Default separator: ","
Hi#footnote[A]#footnote[B]#footnote[C]

#set footnote.group(separator: "&")
Hi#footnote[D]#footnote[E]

#set footnote.group(separator: none) // Same effect as before this PR.
Hi#footnote[F]#footnote[G]
```
Outputs:
|`paged` | `html` |
|:------:|:------:|
||Screenshot 2025-12-28 at 01 51 12|

---

**Credit.** This PR was created based on @MDLC01 's [previous work](https://github.com/typst/typst/issues/2967#issuecomment-3281488614) so let me make a note this line should be present in the squashed commit when merging: 
```
Co-authored-by: Malo <57839069+MDLC01@users.noreply.github.com>
```


Enivex:
Is the separator configurable?

Leedehai:
> Is the separator configurable?

Yes, configurable with `#set footnote.group(separator: ...)`.

crates/typst-library/src/model/footnote.rs
```diff
@@ -173,6 +175,38 @@ impl Count for Packed<FootnoteElem> {
...
+    /// By default, it is a comma. If set to `{none}`, there's no separator.
+    ///
+    /// ```example
+    /// #set footnote.group(separator: "&")
+    ///
+    /// The coldest winter is the summer in San Francisco
+    /// #footnote[Referring to the city's characteristic summer fog.]
+    /// #footnote[This quip is often misattributed to Mark Twain.].
+    /// ```
+    #[default(Some(TextElem::packed(",\u{200B}")))]
```
EpicEricEE: I feel like not allowing line breaks between footnotes (i.e. removing the zero-width space or replacing it with a word joiner) would be a better default

Image

Image

Leedehai:
Well, sorry, only commit [5a04fe6](https://github.com/typst/typst/pull/7627/commits/5a04fe6c422891cedf7facfffc55fc5e1fd16eaf) is the commit that I pushed after addressing EpicEricEE's comment. All other commit between that comment and this commit were made earlier, but they appear below that comment because I rebased this branch with `main` to sync...

Leedehai:
Done.

Well, sorry, in fact only commit [5a04fe6](https://github.com/typst/typst/pull/7627/commits/5a04fe6c422891cedf7facfffc55fc5e1fd16eaf) is the commit that I pushed after addressing EpicEricEE's comment. All other commits between that comment and this commit were made earlier, but they appear below that comment in timeline because I rebased this branch with `main` in order to sync...

crates/typst-library/src/model/footnote.rs
```diff
@@ -173,6 +175,38 @@ impl Count for Packed<FootnoteElem> {
...
+    ///
+    /// ```example
+    /// #set footnote.group(separator: "&")
+    ///
+    /// The coldest winter is the summer in San Francisco
+    /// #footnote[Referring to the city's characteristic summer fog.]
+    /// #footnote[This quip is often misattributed to Mark Twain.].
+    /// ```
+    #[default(Some(TextElem::packed(",")))]
+    pub separator: Option<Content>,
```
YDX-2147483647: Perhaps the default `separator` should be `auto`, which resolves to `","`/`" "`/`none` depending on `footnote.numbering`?

For example:

```rust
match footnote.numbering {
  "1" | "a" | "A" | "i" | "I" | "α" | "Α" => ",",
  "①" | "⓵" => " ", // The circles has already avoided ambiguity. Adding a comma is redundant
  "*" => " ", // It would be confusing if we put another symbol (the comma) between `*` and `†`
  "一" | "壹" => none, // Irrelevant: No one will use them for footnote.numbering
  "あ" | "い" | "ア" | "イ" | "א" | "가" | "ㄱ" | "١" | "۱" | "१" | "১" | "ক" => ???, // I have no idea…
  complex_patterns | function => none, // This is a safer fallback for patterns like `1)` / `[1]` / `[一]`
}
```

EDIT: I've never read any book that uses `一` or `壹` to number footnotes.
My estimation: Among published Chinese books, ~70% use `①`, ~20% use `[1]`, ~9% use `1`, ~1% use `*`/`1)`/`(1)`, <0.05% use others.
(a bit off the topic...)

Leedehai: For 一/壹/あ/イ/ア the normal choice is  "、". If it's none, readers can't disambiguate "九十" ("9,10" vs "90") for example.

The below is from AI, to which I asked for the most common and appropriate footnote number separator.
- For Korean 가/ㄱ, AI said it's interpunct "ㆍ".
- For Hebrew א, AI said it's the standard comma.
- For Arabic ١ (U+0661) and Persian ۱ (U+06F1), AI said it's the Arabic comma "،". 
- For Hindi १ and Bengali ১/ক, AI said it's the standard comma.

Leedehai:
Do we need to do something special in the PDF/HTML layout rules if the language is RTL, e.g. Arabic? Or the layout system already handled it? @laurmaedje and the team :)

Leedehai:
Hi @laurmaedje is the feature proposed by this PR still desirable? (I'm fine either way)

laurmaedje:
I think the general feature of separating with a delimiter in some way is probably desirable. What I'm not yet sure about is the best way to expose it in terms of configuration. I'm somewhat shying away from having a full-blown footnote group concept, but I'm also not sure whether another solution is actually better. I didn't manage to find time for this before 0.15, but I'll try to take a look in the upcoming weeks.

laurmaedje (commented):
I took a closer look now. The feature definitely seems desirable, but I have a few comments:

- First of all, I would like to avoid exposing a `footnote.group` construct. I think the user needn't be concerned with it (`CiteGroup` is also an internal concept). Instead, I would suggest having `#set footnote(separator: "..")` directly.

- I'm somewhat concerned with maintaining yet another list of numbering -> default separator, especially in typst/typst. If @MDLC01 thinks this is something that would fit into `codex` alongside the numeral systems, then I'd be fine with that.[^1]

- On the code side, there's quite a bit of duplication in the HTML and paged show rule. It would be good to factor out the shared part into some method on `FootnoteGroup`.

- Reversing the superscripts in RTL seems wrong since content should be in logical order. If the goal is to have "1" followed by "2" to be rendered as "2,1", the appropriate fix would probably be to surround the individual subscripts with LRI (Left-to-Right Isolate) and PDI (Pop Directional Isolate) characters. Though ideally we would get confirmation from at least one native speaker of an RTL language on how consecutive numbered footnotes should be displayed in their language (maybe @emilyyyylime?).

I also noticed that the the implementation does not work in combination with `show footnote` because then the groups are not collected anymore and hence there won't be a separator. From an implementation point of view, this felt odd to me, but I guess from a user point of view, it is not too unexpected that built-in toggles stop doing something if you fully override the show rule, especially if we move the separator to `footnote` and keep the footnote group element internal.

[^1]: On a side note: I was also wondering whether the smart quote list might fit into codex, considering its expanded scope?

crates/typst-library/src/model/footnote.rs
```diff
@@ -177,6 +179,87 @@ impl Count for Packed<FootnoteElem> {
...
+                        | NumberingKind::KatakanaAiueo
+                        | NumberingKind::KatakanaIroha => Some(TextElem::packed("、")),
+                        NumberingKind::KoreanJamo | NumberingKind::KoreanSyllable => {
+                            Some(TextElem::packed("·"))
+                        }
+                        NumberingKind::EasternArabic
+                        | NumberingKind::EasternArabicPersian => {
+                            Some(TextElem::packed("،"))
+                        }
+                        NumberingKind::Symbol => Some(TextElem::packed(" ")),
```
laurmaedje, line 251: A space seems a little odd, maybe this should default to no separator instead?

Leedehai:
Hi, thanks! All points SGTM.

As for the RTL.. maybe let's postpone it until a later PR so as to keep this PR small, and also to collect thoughts from native speakers?

laurmaedje:
> As for the RTL.. maybe let's postpone it until a later PR so as to keep this PR small, and also to collect thoughts from native speakers?

Let's can see whether clarity is reached while you're still working on the PR and, if not, we can also postpone it.

MDLC01:
> I'm somewhat concerned with maintaining yet another list of numbering -> default separator, especially in typst/typst. If @MDLC01 thinks this is something that would fit into `codex` alongside the numeral systems, then I'd be fine with that.

There is still the idea floating around of reworking numbering patterns. Maybe this concept of separator can be explored when we get to that? Also, which existing list of numbering -> default separator are you referring to?

crates/typst-library/src/model/footnote.rs
```diff
@@ -177,6 +179,87 @@ impl Count for Packed<FootnoteElem> {
...
+            Numbering::Func(_) => None,
+            Numbering::Pattern(pattern) => match pattern {
+                // If a footnote marker's either end is normal text
+                // (e.g. "#1", "1)", "[1]"), it's generally unnecessary
+                // to add a separator.
+                NumberingPattern { suffix, .. } if !suffix.is_empty() => None,
+                NumberingPattern { pieces, .. } => match pieces.as_slice() {
+                    [(prefix, ..), ..] if !prefix.is_empty() => None,
+                    [.., (.., kind)] => match kind {
+                        NumberingKind::LowerSimplifiedChinese
```
laurmaedje, line 236: @MDLC01 This is the list.

Edit: With "yet another" I did not mean that there is an existing separator list. Just that it's another list where per-language decisions need to be made.

laurmaedje:
Which rework do you mean? Do you mean a general one of numbering patterns? That seems somewhat orthogonal to me and it would be unfortunate to block this PR on some rework that will maybe happen down the road.

MDLC01:
> Which rework do you mean? Do you mean a general one of numbering patterns?

Yes.

> That seems somewhat orthogonal to me and it would be unfortunate to block this PR on some rework that will maybe happen down the road.

I don't think it needs to block this PR. The list can stay here for now. But if you prefer to move it to Codex immediately I'm fine with that as well. It should probably be a simple method on `NamedNumeralSystem`.

laurmaedje:
> But if you prefer to move it to Codex immediately I'm fine with that as well. It should probably be a simple method on NamedNumeralSystem.

Sounds good.

laurmaedje:
@Leedehai Hey, do you still plan to finalize work on this? I fully understand if you're busy (it's been myself often enough); I just wanted to friendlily bump the PR. I think the feature is desirable and if you don't want to work on it anymore, I would probably finish it myself at some point.

Leedehai:
Hi! A busy month indeed, thanks for the reminder. Will get to it this weekend.

PR #7629, opened 2025-12-27
Spot colors

reknih:
This PR adds spot colors to precisely request a given pigment in a professional print environment.

I have added the new type `spot-colorant`. Once constructed with a colorant name and a fallback color, the type has a single `.tint(Ratio)` method to construct a new spot color. `spot-colorant` accepts the special values `"all"` (all available colorants) and `none` (no colorants) that are specified in PDF.

For this feature, I had to break out process colors from the Color enum. I have tried my best to dispatch to small, Copyable process types early while keeping the existing API in place.

There is a behavioral change for gradients: They will now eagerly convert to the mixing space instead of doing so during export. This means that errors in conversion (which conversions to spot spaces can yield) will result in an error with a proper span when constructing the gradient instead of an unspanned error or panic during export. This means that `gradient.stops` will now return the spots in the mixing color space instead of their original space. The change is also a speedup since the conversion does not have to be performed for each sample.

Gradients gained the possibility to choose `auto` (new default) as a mixing space. `auto` will result in a spot color if all stops are spot colors using the same colorant and the old default of Oklab otherwise.

For the moment, all exporters will just export the fallback. I [have written the necessary code for Krilla](https://github.com/LaurenzV/krilla/pull/323) to support Separation color spaces. Once that is merged, I will integrate it here and set this to Ready to Review.

Open questions:
- Should the colorant name and fallback color be exposed on `spot-colorant`
- Should the computed fallback color be exposed on spot colors?
- The overall size of `SpotColorant` makes `Value` larger than the 32 bytes allowed by the tests. How should this be addressed?


baron-de-ropp:
This is literally the last thing I need to stop using latex, and I just discovered typst yesterday. Thank you for doing the Lord's work.

baron-de-ropp:
Hey! The CI is failing because SpotColor increases the Value enum size beyond 32 bytes.

Fix: Wrap SpotColor in Arc in the Color enum:

Spot(Arc<SpotColor>)  // instead of Spot(SpotColor)
Then wrap all SpotColor::new(...) calls in Arc::new(...). This matches the pattern used by Gradient which also wraps its variants in Arc.

I'm shit at rust tho so double check and make sure i'm not a jagoff.

crates/typst-library/src/visualize/color.rs
```diff
@@ -210,30 +234,31 @@ pub enum Color {
...
+    MAROON: Rgb -> 0.521569, 0.078431, 0.294118, 1.0,
+    RED: Rgb -> 1.0, 0.254902, 0.211765, 1.0,
+    ORANGE: Rgb -> 1.0, 0.521569, 0.105882, 1.0,
+    YELLOW: Rgb -> 1.0, 0.8627451, 0.0, 1.0,
+    OLIVE: Rgb -> 0.239216, 0.6, 0.4392157, 1.0,
+    GREEN: Rgb -> 0.1803922, 0.8, 0.2509804, 1.0,
+    LIME: Rgb -> 0.0039216, 1.0, 0.4392157, 1.0;
+
     /// The module of preset color maps.
     pub const MAP: fn() -> Module = || typst_utils::singleton!(Module, map()).clone();
```
PgBiel: I wonder if we could get `#[scope]` to work across multiple impl blocks so we can split this into  `scope_with_color_definitions! { Color, ... }` and `#[scope] impl Color { ... }` for more flexibility (would also allow multiple such macros).

But that seems complex enough to warrant a separate PR, and in the worst case scenario could require linking hacks[^1] to work properly.

[^1]: https://docs.rs/inventory/

crates/typst-library/src/visualize/color.rs
```diff
@@ -815,121 +787,83 @@ impl Color {
...
         Ok(match self {
-            Self::Luma(_) => bail!(
-                span, "cannot saturate grayscale color";
-                hint: "try converting your color to RGB first";
+            Self::Process(c) => Self::Process(c.saturate(span, factor)?),
+            Self::Spot(_) => bail!(
+                span,
+                "cannot saturate spot color";
+                hint: "try using the fallback color instead";
             ),
```
PgBiel, line 962: This probably doesn't matter too much, but it's worth noting that there are some other operations which use the fallback automatically, notably `to_hex` and other `to_` conversions, instead of throwing errors. Not sure if this would be too bad of an inconsistency, but it could be acceptable if we assume that those conversions are generally lossy anyway.

In addition, this error should be noted in the method's docs. (Other methods might be missing some additional documentation too.)

crates/typst-library/src/visualize/color.rs
```diff
@@ -1755,6 +2147,194 @@ impl Cmyk {
...
+        /// Name of the spot colorant to use.
+        ///
+        /// In production, this name will be manually checked and matched to a
+        /// colorant, so this value needs to be unambiguous. It's best to
+        /// reference a color from a registry like PANTONE, HKS, RAL, Toyo &
+        /// DIC, etc.
+        ///
+        /// Values in here may be treated case-sensitively during production:
+        /// `{"PANTONE 2221 C"}` and `{"PANTONE 2221 C"}` may be treated as
+        /// separate colors. Ensure that you are using a consistent naming
```
PgBiel: It seems like those are the exact same names. Perhaps you meant something like "c" instead of "C"?

```suggestion
        /// `{"PANTONE 2221 C"}` and `{"PANTONE 2221 c"}` may be treated as
        /// separate colors. Ensure that you are using a consistent naming
```

crates/typst-library/src/visualize/color.rs
```diff
@@ -1158,118 +1075,79 @@ impl Color {
...
-        match &mut self {
-            Color::Cmyk(_) => {}
-            Color::Luma(c) => c.alpha = alpha,
-            Color::Oklab(c) => c.alpha = alpha,
-            Color::Oklch(c) => c.alpha = alpha,
-            Color::Rgb(c) => c.alpha = alpha,
-            Color::LinearRgb(c) => c.alpha = alpha,
-            Color::Hsl(c) => c.alpha = alpha,
-            Color::Hsv(c) => c.alpha = alpha,
+    pub fn with_alpha(&mut self, alpha: f32) -> Self {
```
PgBiel: Since there are now multiple color spaces / types without an alpha channel, including any spot colors, perhaps we could list them more prominently within `color` docs, at the top.

PgBiel (commented):
Very cool stuff! I took a look around, and wanted to leave some comments.

crates/typst-library/src/visualize/color.rs
reknih: I think it's valid to allow conversions but fail on color modification operations. I think that otherwise it's too easy to make an error where you silently fall back to a process color which could have a very different print result.

crates/typst-library/src/visualize/color.rs
reknih: Pantone themselves seems to go with an uppercase C: https://www.pantone.com/eu/en/color-finder/2221-C

PgBiel: Yeah, I meant that this docs example seems to want to show how input might be invalid if you change a letter's case, but the two examples are identical

reknih: oh, got it, fixed

urshofer:
@reknih Are there still plans to merge this PR into main ? It would be such a nice feature, we use it in a project, and it works well!

laurmaedje:
The PR will at some point be merged, yes.

crates/typst-pdf/src/convert.rs
```diff
@@ -570,8 +572,13 @@ fn convert_error(
             "{prefix} the PDF is missing a CMYK profile";
             hint: "CMYK colors are not yet supported in this export mode";
         ),
-        ValidationError::InconsistentSeparationFallback(_) => {
-            unreachable!("separation color spaces are not currently used")
+        ValidationError::InconsistentSeparationFallback(colorant) => {
+            let repr = SpotColorantName::from_krilla(colorant).repr();
+            error!(
+                Span::detached(),
+                "spot colorant {repr} appeared with multiple distinct fallback colors";
```
laurmaedje:
```suggestion
                "spot colorant `{repr}` appeared with multiple distinct fallback colors";
```

crates/typst-pdf/src/paint.rs
```diff
@@ -1,25 +1,27 @@
...
-use krilla::color::{self, cmyk, luma, rgb};
+use krilla::color::separation::SeparationSpace;
+use krilla::color::{self, cmyk, luma, rgb, separation};
 use krilla::num::NormalizedF32;
 use krilla::paint::{
     Fill, LinearGradient, Pattern, RadialGradient, SpreadMethod, Stop, Stroke,
     StrokeDash, SweepGradient,
 };
 use krilla::surface::Surface;
+use std::f64::consts::PI;
```
laurmaedje: this should be moved back

crates/typst-pdf/src/paint.rs
```diff
@@ -280,7 +302,14 @@ fn convert_gradient_stops(gradient: &Gradient) -> Vec<Stop> {
     let mut stops = vec![];
 
     let mut add_single = |color: &Color, offset: Ratio| {
-        let (color, opacity) = convert_solid(&color.to_space(gradient.space()));
+        let (color, opacity) = match color {
```
laurmaedje: it looks like this could just use `convert_solid`

crates/typst-svg/src/paint.rs
```diff
@@ -436,14 +443,16 @@ impl From<&Gradient> for GradientKind {
...
 impl SvgDisplay for Color {
     fn fmt(&self, f: &mut impl SvgWrite) {
-        match *self {
-            c @ Color::Rgb(_)
-            | c @ Color::Luma(_)
-            | c @ Color::Cmyk(_)
-            | c @ Color::Hsv(_) => {
+        let process = self.to_process();
+
+        match process {
```
laurmaedje:
```suggestion
        match self.to_process() {
```

tests/suite/visualize/color.typ
```diff
@@ -220,15 +220,15 @@
 #color.mix((red, 1, 2))
 
 --- color-mix-bad-space-type eval ---
-// Error: 31-38 expected `rgb`, `luma`, `cmyk`, `oklab`, `oklch`, `color.linear-rgb`, `color.hsl`, or `color.hsv`, found string
+// Error: 31-38 expected `rgb`, `luma`, `cmyk`, `oklab`, `oklch`, `color.linear-rgb`, `color.hsl`, `color.hsv`, or `color.spot`, found string
```
laurmaedje: this seems incorrect, since it expects the _constructor_ `color.hsv`, but an _instance_ of `color.spot`, yet both are listed here equivalently

tests/suite/visualize/color.typ
```diff
@@ -327,3 +327,75 @@
 // output with a PDF comparison script.
 #rect(fill: red.transparentize(50%))
 #image("/assets/images/tiger.jpg", width: 45pt)
+
+--- spot-color-basic paged ---
+// Test basic spot color creation and rendering
```
laurmaedje: Please terminate the comments with full stops

tests/suite/visualize/gradient.typ
```diff
@@ -634,6 +634,7 @@ $ A = mat(
 #test(gradient.linear(red, green, blue).kind(), gradient.linear)
 
 --- gradient-stops eval ---
+// Error: 2-90 Assertion failed: (\n  (oklab(65.95%, 0.2, 0.108), 0%),\n  (oklab(73.96%, -0.179, 0.129), 50%),\n  (oklab(56.22%, -0.05, -0.17), 100%),\n) != (\n  (rgb("#ff4136"), 0%),\n  (rgb("#2ecc40"), 50%),\n  (rgb("#0074d9"), 100%),\n)
```
laurmaedje: this test seems to be failing in disguise

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
...
+    }
+}
+
+/// A spot colorant from which spot colors can be created.
+///
+/// Use spot colors to request a precise pigment in a professional print
+/// environment. Once you have created a spot colorant, you can create
+/// colors using its @color.spot.tint[`tint` method].
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[ty(scope, name = "spot")]
```
laurmaedje:
```suggestion
#[ty(scope, name = "spot", title = "Spot Colorant")]
```
this way, also the long type name becomes (lowercased) "spot colorant" and error messages will say "expected spot colorant, found X" instead "expected spot, found X"

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
...
+    /// Returns the fallback process color with the tint applied.
+    pub fn fallback(&self) -> ProcessColor {
+        self.colorant.fallback.lighten(Ratio::one() - self.tint)
+    }
+}
+
+impl Repr for SpotColor {
+    fn repr(&self) -> EcoString {
+        eco_format!(
+            "spot-color({}, colorant: {})",
```
laurmaedje: this seems either outdated or random

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
...
+            "color.spot({}, {})",
+            self.name.repr(),
+            self.fallback.into_value().repr()
+        )
+    }
+}
+
+/// The name of a spot colorant.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum SpotColorantName {
```
laurmaedje: I would at least like to consider dropping `NoColorant` here and renaming `AllColorants` to `All` and then using `None` in the `name` field of `SpotColorant`. Having custom none variants (especially when they map from Typst's `none`) feels a bit like a code smell.

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
...
+    /// No colorant (equivalent to `none`).
+    NoColorant,
+    /// The special "All" colorant that applies to all separation colorants.
+    AllColorants,
+    /// A custom colorant name (e.g., "PANTONE 2221 C").
+    Custom(EcoString),
```
laurmaedje: should we use backticks for the strings here? Also, it's a bit unclear whether "All" is a PDF or Typst name here.

docs/components/base.typ
```diff
@@ -73,6 +73,24 @@
...
+  } else if n == 4 {
+    html.h4
+  } else if n == 5 {
+    html.h5
+  } else {
+    html.h6
+  }
+
+  fn(..args)
+}
```
laurmaedje, line 81:
```suggestion
#let html-headings = (html.h1, html.h2, html.h3, html.h4, html.h5, html.h6)
#let html-heading-n(n, ..args) = {
  assert(n > 0)
  let fn = html-headings.at(n - 1, default: html.h6)
  fn(..args)
}
```

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1283,16 +1296,115 @@ cast! {
...
 /// different gradient types.
 #[comemo::memoize]
-fn process_stops(stops: &[Spanned<GradientStop>]) -> SourceResult<Vec<(Color, Ratio)>> {
-    let has_offset = stops.iter().any(|stop| stop.v.offset.is_some());
+fn process_stops(
+    stops: &[Spanned<GradientStop>],
+    space: Spanned<Smart<ColorSpace>>,
+) -> SourceResult<(Vec<(Color, Ratio)>, ColorSpace)> {
+    let space_span: Option<_> =
+        if space.span.is_detached() { None } else { Some(space.span) };
```
laurmaedje:
```suggestion
        (!space.span.is_detached()).then_some(space.span);
```

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1283,16 +1296,115 @@ cast! {
...
+    let space = match verdict {
+        ColorSpaceVerdict::AutoWithoutPrejudice => ColorSpaceVerdict::fallback_space(),
+        ColorSpaceVerdict::Space(s) => s,
+        ColorSpaceVerdict::MayBeSpotOrOklab(colorant) => ColorSpace::Spot(colorant),
+    };
+
+    let stops_iter = stops.iter().map(|s| -> SourceResult<Spanned<GradientStop>> {
+        Ok(Spanned::new(
+            GradientStop {
+                color: s.v.color.to_space(&space).at(s.span).map_err(|diagnostics| {
```
laurmaedje: creating an intermediate `let color = s.v. ....` would make this more readable

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1310,13 +1422,14 @@ fn process_stops(stops: &[Spanned<GradientStop>]) -> SourceResult<Vec<(Color, Ra
             last_stop = stop.get();
         }
 
-        let out = stops
-            .iter()
-            .map(|Spanned { v: GradientStop { color, offset }, span }| {
+        let out = stops_iter
+            .map(|stop| {
+                let Spanned { v: GradientStop { color, offset }, span } = stop?;
+
```
laurmaedje:
```suggestion
```

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1283,16 +1296,115 @@ cast! {
...
+    /// What color space to use to mix the gradient.
+    enum ColorSpaceVerdict {
+        /// The space parameter was `auto` and we did not yet process any stops.
+        AutoWithoutPrejudice,
+        /// We settled on a color space, either by user input or by seeing
+        /// multiple non-matching spaces that we will mix in Oklab.
+        Space(ColorSpace),
+        /// So far, we only have seen stops with the same spot colorant. If no
+        /// other stops occur, we can mix using the colorant.
+        MayBeSpotOrOklab(SpotColorant),
```
laurmaedje: I find the `OrOklab` part confusing. Yes, if other stuff happens, we will go to oklab. But the enum really encodes the _current_ state of a state machine and the crucial part is `MayBeSpot` or even just `Spot`.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1283,16 +1296,115 @@ cast! {
...
+    stops: &[Spanned<GradientStop>],
+    space: Spanned<Smart<ColorSpace>>,
+) -> SourceResult<(Vec<(Color, Ratio)>, ColorSpace)> {
+    let space_span: Option<_> =
+        if space.span.is_detached() { None } else { Some(space.span) };
+
+    /// What color space to use to mix the gradient.
+    enum ColorSpaceVerdict {
+        /// The space parameter was `auto` and we did not yet process any stops.
+        AutoWithoutPrejudice,
```
laurmaedje: I don't get what `WithoutPrejudice` is trying to tell me.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -1283,16 +1296,115 @@ cast! {
...
+                // color, so we can go to the fallback state.
+                _ => ColorSpaceVerdict::fallback(),
+            },
+        };
+
+        // `ColorSpaceVerdict::Space` and `has_offset == true` are terminal
+        // states.
+        if matches!(verdict, ColorSpaceVerdict::Space(_)) && has_offset {
+            break;
+        }
```
laurmaedje: The comment sounds like both (individually) are terminal states (because plural "are"), but the condition says that they form a terminal state _together._

crates/typst-library/src/visualize/color.rs
```diff
@@ -66,6 +66,9 @@ static TO_SRGB: LazyLock<Arc<moxcms::Transform8BitExecutor>> = LazyLock::new(||
 /// - Linear RGB through the @color.linear-rgb[`color.linear-rgb` function]
 /// - HSL through the @color.hsl[`color.hsl` function]
 /// - HSV through the @color.hsv[`color.hsv` function]
+/// - Color spaces described by Spot colorants through the @color.spot[`color.spot` function]
```
laurmaedje: Please wrap at column 80

laurmaedje: Also, do we say "Spot colorants" or "spot colorants"? Because other parts of the docs also have "spot" lowercased.

crates/typst-library/src/visualize/color.rs
```diff
@@ -66,6 +66,9 @@ static TO_SRGB: LazyLock<Arc<moxcms::Transform8BitExecutor>> = LazyLock::new(||
...
+///
+/// All color spaces except for CMYK and any Spot colorants have alpha channels.
```
laurmaedje: The "any" reads a bit weirdly, I'd drop it

crates/typst-library/src/visualize/color.rs
```diff
@@ -1282,118 +1206,82 @@ impl Color {
...
-        match &mut self {
-            Color::Cmyk(_) => {}
-            Color::Luma(c) => c.alpha = alpha,
-            Color::Oklab(c) => c.alpha = alpha,
-            Color::Oklch(c) => c.alpha = alpha,
-            Color::Rgb(c) => c.alpha = alpha,
-            Color::LinearRgb(c) => c.alpha = alpha,
-            Color::Hsl(c) => c.alpha = alpha,
-            Color::Hsv(c) => c.alpha = alpha,
+    pub fn with_alpha(&mut self, alpha: f32) -> Self {
```
laurmaedje: This should take `self` and then it doesn't need to clone.

crates/typst-library/src/visualize/color.rs
```diff
@@ -291,28 +303,28 @@ pub enum Color {
 
 #[scope]
 impl Color {
+    pub const BLACK: Self = Self::Process(ProcessColor::BLACK);
```
laurmaedje: I think it suffices to have these constants here and inline the process color definitions.

The only use of any of the `ProcessColor` constants is in `impl Frame`, but that could just as well use `Color`.

crates/typst-library/src/visualize/color.rs
```diff
@@ -342,13 +354,16 @@ impl Color {
         color: Color,
     ) -> SourceResult<Color> {
         Ok(if let Some(color) = args.find::<Color>()? {
-            Color::Luma(color.to_luma())
+            Self::Process(ProcessColor::Luma(color.to_luma()))
```
laurmaedje: The `Self::Process(ProcessColor::Luma(..))` is duplicated across both branches and could go over the whole if. Same may apply for other constructors.

crates/typst-library/src/visualize/color.rs
```diff
@@ -938,121 +902,89 @@ impl Color {
...
-            Self::Hsl(c) => Self::Hsl(c.darken(factor)),
-            Self::Hsv(c) => Self::Hsv(c.darken(factor)),
+            Self::Process(c) => Self::Process(c.darken(factor)),
+            Self::Spot(c) => Self::Spot(c.darken(factor)),
         }
     }
 
     /// Increases the saturation of a color by a given factor.
+    ///
+    /// Only process colors can be saturated. If you want to saturate a spot
```
laurmaedje, line 947: I don't think the term "process color" has been properly introduced in the documentation.

crates/typst-library/src/visualize/color.rs
```diff
@@ -1066,84 +998,70 @@ impl Color {
...
-                c.lightness,
-                c.alpha,
-            )),
-            Self::Hsv(c) => Self::Hsv(Hsv::new(
-                RgbHue::from_degrees(c.hue.into_degrees() + 180.0),
-                c.saturation,
-                c.value,
-                c.alpha,
-            )),
+        #[default(Smart::Auto)]
```
laurmaedje:
```suggestion
        #[default]
```
For types with an obvious `Default` impl, just this suffices.

crates/typst-library/src/visualize/color.rs
```diff
@@ -1066,84 +998,70 @@ impl Color {
...
-            ColorSpace::Hsv => {
+            ProcessColorSpace::Hsv => {
                 let hsv = self.to_hsv();
                 let rotated = hsv.shift_hue(angle.to_deg() as f32);
-                Self::Hsv(rotated).to_space(self.space())
+                Self::Process(ProcessColor::Hsv(rotated))
+                    .to_space(&self.space())
+                    .at(span)?
             }
             _ => bail!(span, "this colorspace does not support hue rotation"),
```
laurmaedje: while we're at it, might add that missing space

crates/typst-library/src/visualize/color.rs
```diff
@@ -1165,10 +1083,14 @@ impl Color {
         colors: Vec<WeightedColor>,
         /// The color space to mix in. By default, this happens in a perceptual
         /// color space (@color.oklab[`oklab`]).
+        ///
+        /// All colors will be converted into this color space. To mix spot
+        /// colors sharing a single colorant without converting them into
+        /// process colors, provide their shared colorant as the mixing space.
         #[named]
-        #[default(ColorSpace::Oklab)]
+        #[default(ColorSpace::Process(ProcessColorSpace::Oklab))]
```
laurmaedje: why do the other methods take `auto` and then deal with spot colors automatically, but this one expects you to pass the spot colorant as the mixing space?

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+    }
+
+    pub fn to_hsv(&self) -> Hsv {
+        match self {
+            Self::Process(c) => c.to_hsv(),
+            Self::Spot(c) => c.fallback().to_hsv(),
+        }
+    }
+
+    /// Convert to ProcessColor, using fallback for spot colors.
```
laurmaedje:
```suggestion
    /// Convert to `ProcessColor`, using fallback for spot colors.
```

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Process(a), Self::Process(b)) => PartialEq::eq(a, b),
+            (Self::Spot(a), Self::Spot(b)) => PartialEq::eq(a, b),
+            _ => false,
+        }
+    }
+}
+
+impl Eq for Color {}
```
laurmaedje: can't these be derived?

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+        match self {
+            Self::Process(c) => Repr::repr(c),
+            Self::Spot(c) => Repr::repr(c),
+        }
+    }
+}
+
+fn hue_angle(degrees: f32) -> Angle {
+    Angle::deg(f64::from(degrees).rem_euclid(360.0))
+}
```
laurmaedje: this definition is a bit randomly in the middle of the color impls

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+
+impl Hash for Color {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        core::mem::discriminant(self).hash(state);
+        match self {
+            Self::Process(c) => c.hash(state),
+            Self::Spot(c) => c.hash(state),
+        }
+    }
+}
```
laurmaedje: this could also be derived

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+
+impl From<Luma> for ProcessColor {
+    fn from(c: Luma) -> Self {
+        Self::Luma(c)
+    }
+}
+
+impl From<Luma> for Color {
```
laurmaedje: these could all be replaced by
```rs
impl<C> From<C> for Color
where
    C: Into<ProcessColor>,
{
    fn from(color: C) -> Self {
        Self::Process(color.into())
    }
}
```

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+                // Duplicate number for shorthand notation, i.e. `a` -> `aa`
+                values[elem] += values[elem] * 16;
+            }
+        }
+
+        Ok(Self::from_u8(values[0], values[1], values[2], values[3]))
+    }
+}
+
+impl From<Luma> for ProcessColor {
```
laurmaedje: I think I'd prefer if the whole `ProcessColor` type (including its definition, inherent and trait impls) would move fully below all of the `Color` impls and, in particular, its trait impls below its inherent impl(s)

crates/typst-library/src/visualize/color.rs
```diff
@@ -1402,51 +1290,662 @@ impl Color {
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        core::mem::discriminant(self).hash(state);
+        match self {
+            Self::Process(c) => c.hash(state),
+            Self::Spot(c) => c.hash(state),
+        }
+    }
+}
+
+impl FromStr for Color {
```
laurmaedje, line 1409: Would be nice to move this to `FromStr for ProcessColor` and then just forward here

crates/typst-library/src/visualize/color.rs
```diff
@@ -1704,106 +2216,15 @@ impl PartialEq for Color {
...
-impl From<Hsv> for Color {
-    fn from(c: Hsv) -> Self {
-        Self::Hsv(c)
-    }
+cast! {
+    ProcessColor,
+    self => Color::from(self).into_value(),
+    v: Color => match v {
+        Color::Process(c) => c,
+        Color::Spot(_) => bail!("spot color found where process color is required")
```
laurmaedje, line 2163: could you add a test for this. error message does not seem to be exercised by the test suite.

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
     }
 }
 
+/// A Spot color.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct SpotColor {
+    /// How much of the colorant to apply to the medium.
+    pub tint: Ratio,
+    /// Which colorant to apply.
+    pub colorant: Arc<SpotColorant>,
```
laurmaedje: I think the opposite field order would be a bit more natural, also in `new`

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
...
+    /// How to render this color if the specified colorant is not available.
+    pub fallback: ProcessColor,
+}
+
+impl Repr for SpotColorant {
+    fn repr(&self) -> EcoString {
+        eco_format!(
+            "color.spot({}, {})",
+            self.name.repr(),
+            self.fallback.into_value().repr()
```
laurmaedje: `ProcessColor` already implements `Repr`
```suggestion
            self.fallback.repr()
```

crates/typst-library/src/visualize/color.rs
```diff
@@ -1881,6 +2302,196 @@ impl Cmyk {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[ty(scope, name = "spot")]
+pub struct SpotColorant {
+    /// Name of the spot colorant to use.
+    pub name: SpotColorantName,
+    /// How to render this color if the specified colorant is not available.
+    pub fallback: ProcessColor,
+}
+
+impl Repr for SpotColorant {
```
laurmaedje, line 2311: please organize the impls as usual, with inherent first and trait impls later.

crates/typst-library/src/visualize/color.rs
laurmaedje: please move the name to its own section below the SpotColorant impls

crates/typst-library/src/visualize/color.rs
```diff
@@ -1920,8 +2531,49 @@ cast! {
...
+    }
+}
+
+cast! {
+    ColorSpace,
+    self => match self {
+        Self::Process(s) => s.variant_constructor(),
+        Self::Spot(s) => s.into_value()
+    },
+    spot_colorant: SpotColorant => Self::Spot(spot_colorant),
```
laurmaedje:
```suggestion
    spot: SpotColorant => Self::Spot(spot),
```

crates/typst-library/src/visualize/color.rs
```diff
@@ -66,75 +66,61 @@ static TO_SRGB: LazyLock<Arc<moxcms::Transform8BitExecutor>> = LazyLock::new(||
...
+/// throughout the printing process instead of ahead of time.
 ///
 /// = Example <example>
 /// ```example
 /// #rect(fill: aqua)
 /// ```
 ///
-/// = Predefined colors <predefined-colors>
-/// Typst defines the following built-in colors:
+/// = Predefined colors <predefined-colors> Typst defines the following built-in
```
laurmaedje: The rewrap broke the docs here

crates/typst-library/src/visualize/color.rs
```diff
@@ -2488,11 +2342,42 @@ impl SpotColorant {
...
+            Self::Custom(name) => Str::from(name.clone()).repr(),
+        }
+    }
+}
+
+cast! {
+    SpotColorantName,
+    self => match self {
+        Self::AllColorants => "all".into_value(),
+        Self::Custom(name) => name.as_str().into_value(),
```
laurmaedje: I think the `as_str` can be dropped here

crates/typst-library/src/visualize/color.rs
```diff
@@ -2556,9 +2494,9 @@ cast! {
         Self::Process(s) => s.variant_constructor(),
         Self::Spot(s) => s.into_value()
     },
-    spot_colorant: SpotColorant => Self::Spot(spot_colorant),
+    spot: SpotColorant => Self::Spot(spot),
     v: Value => {
-        let expected = "expected `rgb`, `luma`, `cmyk`, `oklab`, `oklch`, `color.linear-rgb`, `color.hsl`, `color.hsv`, or `color.spot`";
+        let expected = "expected `rgb`, `luma`, `cmyk`, `oklab`, `oklch`, `color.linear-rgb`, `color.hsl`, `color.hsv`, or spot color";
```
laurmaedje: this should say "spot colorant" rather than "spot color"

tests/suite/visualize/gradient.typ
laurmaedje: This one is still wrong

baron-de-ropp:
YES ITS IN MAIN!!!!! 🐔🐓🎉🎉🎉

docs/assets/docs.css
```diff
@@ -702,6 +732,12 @@ h1 .pill {
   padding: 4px 12px;
 }
 
+h3 .pill {
+  font-size: 20px;
+  border-radius: 8px;
+  padding: 4px 8px;
+}
+
```
isuffix, line 740: This font-size makes the parameter name pills look plain goofy. Here's a screenshot of the `str` constructor:

Image

(same on firefox/chrome)

I assume there was another part of the docs that this change improved? Can we just scope it to that location?

isuffix: I looked more and I now see that this is being used for the spot colorant type (a nested type using a pill as the header in an `h3`), and it looks like not a single other pill on the `color` page is inside an `h3`, so that would explain why this wasn't noticed.

I'm opening a PR to change the targeting to exclude the normal pills in `additional-info` divs.

PR #7632, opened 2025-12-28
Fix Greek Numbering to Modern Greek

purpl3F0x:
Although "ϛ" is technically correct, as numbering, it is not used in modern Greek and instead "στ" is used for the number 6. 

This commit addresses this issue. The 6000 unit is left as is.


laurmaedje:
Thanks for the PR! Could you please point to an authoritative source that backs this? Otherwise, it is hard for me to judge.

laurmaedje:
There is also the issue that the test are failing. These would need to be adjusted.

purpl3F0x:
> Thanks for the PR! Could you please point to an authoritative source that backs this? Otherwise, it is hard for me to judge.
Here some sources:

Ministry of Education Elementary book (online version): https://ebooks.edu.gr/ebooks/v/html/8547/2334/Grammatiki-Neas-Ellinikis-Glossas_A-B-G-Gymnasiou_html-apli/index_C_04.html#:~:text=6-,%CF%83%CF%84%CE%84(%CF%82)%CE%84,-%CE%AD%CE%BE%CE%B9


6th grade denoted as ΣΤ formally
https://www.public.gr/product/books/sholika-biblia/st-dimotikoy/geografia-st-dimotikou-10-0236/1617807

also:

https://en.wikipedia.org/wiki/Greek_numerals

also: pic from the first book I found on my desk

{9F2D90AB-6608-4E3E-A9D5-9C9D0F758815}




> There is also the issue that the test are failing. These would need to be adjusted.

I saw that I was waiting for general feedback on the PR before fixing.

laurmaedje:
Thanks! Looking at these links and doing some further research, I _think_ this change makes sense.

One interesting thing I found though is that CSS `list-style: lower-greek` seems to just _skip_ the six?! 

```html
<ol style="list-style-type: lower-greek">
  <li>1</li>
  <li>2</li>
  <li>3</li>
  <li>4</li>
  <li>5</li>
  <li>6</li>
  <li>7</li>
  <li>8</li>
</ol>
```
[greek.html](https://github.com/user-attachments/files/24742394/greek.html)

purpl3F0x:
> Thanks! Looking at these links and doing some further research, I _think_ this change makes sense.
> 
> One interesting thing I found though is that CSS `list-style: lower-greek` seems to just _skip_ the six?!
> 
> ```
> <ol style="list-style-type: lower-greek">
>   <li>1</li>
>   <li>2</li>
>   <li>3</li>
>   <li>4</li>
>   <li>5</li>
>   <li>6</li>
>   <li>7</li>
>   <li>8</li>
> </ol>
> ```
> 
> [greek.html](https://github.com/user-attachments/files/24742394/greek.html)

Seems to also be an issue in word, and you need to define a custom numbering
https://learn.microsoft.com/en-us/answers/questions/4868606/custom-list-number-style

I also found this on the latex babel-greek package
https://mirrors.mit.edu/CTAN/macros/latex/contrib/babel-contrib/greek/babel-greek-doc.html#:~:text=Standing%20for%206,used%20as%20replacement.

I have to also note that (as a greek) I have never seen the symbol "ϛ" anywhere being used.

laurmaedje:
I found some another relevant discussions here: https://github.com/jgm/pandoc/issues/1243. And apparently, the W3C has chosen your variant as [`greek-lower-modern`](https://www.w3.org/TR/predefined-counter-styles/#greek-styles).

For practical use, I'm convinced that this is a good change and I'll go ahead. In the future, when we have named numberings in addition to just the shorthands I could see having an option for both.

Thanks!

PR #7637, opened 2025-12-30
Move footnote link to the superscript position

### LINKED ISSUE #7631: In HTML export, a footnote's link should be on the superscript position
### Description

Hi - in the current typst-html code, a footnote's link is a normal piece of text containing a superscript indicating that footnote's number:
```rust
// crates/typst-html/src/rules.rs, see FOOTNOTE_RULE and FOOTNOTE_ENTRY_RULE
let sup = SuperElem::new(num).pack().spanned(span);
let link = LinkElem::new(dest.into(), sup)
```
So typst code `A#footnote[text]` is compiled into an HTML structure like:

```html
<!-- Status quo -->
A<a id="loc-1" href="#loc-2" role="doc-noteref"><sup>1</sup></a>
```

However, since the footnote number is on the superscript position, the link should be placed on the superscript position as well:
```html
<!-- Proposed -->
A<sup><a id="loc-1" href="#loc-2" role="doc-noteref">1</a></sup>
```

In summary, the visual effects:

|        | Status quo | Proposed |
|:---:|:--------:|:---------:|
|Visual output|Image|Image|

---
For example, the footnote links in Wikipedia pages are all on the superscript position (note in Wikipedia's CSS configs, the underline appears when one hovers the mouse above the footnote):

Image

---
If agreed, it seems like an easy change and I'm happy to make a PR.

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

Leedehai:
Let the superscript wrap the link, instead of letting the link wrap the superscript, in both HTML and PDF exports.

Fixes #7631. In HTML, the new structure is something like `<sup><a>1</a></sup>` instead of `<a><sup>1</sup></a>` (see that issue's description for a visual look), matching the examples in the HTML spec.

### DIFF
--- a/crates/typst-html/src/rules.rs
+++ b/crates/typst-html/src/rules.rs
@@ -336,6 +336,7 @@ const QUOTE_RULE: ShowFn<QuoteElem> = |elem, _, styles| {
 
 const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
     let span = elem.span();
+
     // The footnote number that links to the footnote entry.
     let link = elem.realize(engine, styles)?;
     let sup = SuperElem::new(link)
--- a/crates/typst-layout/src/rules.rs
+++ b/crates/typst-layout/src/rules.rs
@@ -396,8 +396,8 @@ const QUOTE_RULE: ShowFn<QuoteElem> = |elem, _, styles| {
 const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
     // The footnote number that links to the footnote entry.
     let link = elem.realize(engine, styles)?;
-    let sup = PdfMarkerTag::Label(SuperElem::new(link).pack().spanned(elem.span()));
-    Ok(HElem::hole().clone() + sup)
+    let sup = SuperElem::new(link).pack().spanned(elem.span());
+    Ok(HElem::hole().clone() + PdfMarkerTag::Label(sup))
 };
 
 const FOOTNOTE_ENTRY_RULE: ShowFn<FootnoteEntry> = |elem, engine, styles| {
--- a/crates/typst-library/src/model/footnote.rs
+++ b/crates/typst-library/src/model/footnote.rs
@@ -148,9 +148,8 @@ impl Packed<FootnoteElem> {
         let counter = Counter::of(FootnoteElem::ELEM);
         let num = counter.display_at(engine, loc, styles, numbering, span)?;
         let alt = FootnoteElem::alt_text(styles, &num.plain_text());
-        Ok(DirectLinkElem::new(loc.variant(1), num, Some(alt))
-            .pack()
-            .spanned(span))
+        let dest = loc.variant(1);
+        Ok(DirectLinkElem::new(dest, num, Some(alt)).pack().spanned(span))
     }
 
     /// Returns the location of the definition of this footnote.


crates/typst-html/src/rules.rs
```diff
@@ -337,18 +337,16 @@ const QUOTE_RULE: ShowFn<QuoteElem> = |elem, _, styles| {
 const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
     let span = elem.span();
     let (dest, num) = elem.realize(engine, styles)?;
-    let sup = SuperElem::new(num).pack().spanned(span);
-
-    // Link to the footnote entry.
-    let link = LinkElem::new(dest.into(), sup)
+    let link = LinkElem::new(dest.into(), num)
```
PgBiel: Let's keep the comment there, for good measure...
```suggestion

    // Link to the footnote entry.
    let link = LinkElem::new(dest.into(), num)
```

crates/typst-layout/src/rules.rs
```diff
@@ -403,11 +403,17 @@ const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
...
+    let prefix = PdfMarkerTag::Label(
+        DirectLinkElem::new(dest, sup, Some(alt)).pack().spanned(span),
+    );
     Ok(Content::sequence([
         HElem::new(elem.indent.get(styles).into()).pack(),
-        PdfMarkerTag::Label(prefix),
+        prefix,
         HElem::new(number_gap.into()).with_weak(true).pack(),
         body,
     ]))
```
PgBiel: If I understood correctly, `realize` and the footnote entry show rules were changed to have different behavior between PDF and HTML. But I wonder if it wouldn't be fine to do the same for PDF? I suppose the superscript itself wouldn't be represented in the tag tree, right? So it'd be fine to tag a superscript of a link instead of the link itself...

I'll defer the final word on this to @saecki in case my assumption is wrong.

Leedehai: That's my initial thought as well but it turned out the tag tree would change if we align the PDF behavior with HTML -- the annotation node will appear after.

For example, test `footnote-tags-basic`:

<details>

<summary>Original PDF behavior, untouched by this PR</summary>

```yaml
- Tag: P
  /K:
    - Content: page=0 mcid=0
    - Tag: Lbl
      /K:
        - Tag: Link
          /K:
            - Annotation: page=0 subtype=Link
              /Contents: "Footnote 1"
            - Tag: Span
              /BaselineShift:   3.500
              /LineHeight:   6.000
              /K:
                - Content: page=0 mcid=1
                  /Lang: "en"
    - Tag: Note
      /Id: "Note 1"
      /K:
        - Tag: Lbl
          /K:
            - Tag: Link
              /K:
                - Annotation: page=0 subtype=Link
                  /Contents: "1"
                - Tag: Span
                  /BaselineShift:   2.975
                  /LineHeight:   5.100
                  /K:
                    - Content: page=0 mcid=3
                      /Lang: "en"
        - Content: page=0 mcid=4
    - Content: page=0 mcid=2
```

</details>

<details>

<summary>New, if we also align PDF's behavior with HTML</summary>

```yaml
- Tag: P
  /K:
    - Content: page=0 mcid=0
    - Tag: Lbl
      /K:
        - Tag: Link
          /K:
            - Annotation: page=0 subtype=Link
              /Contents: "Footnote 1"
            - Tag: Span
              /BaselineShift:   3.500
              /LineHeight:   6.000
              /K:
                - Content: page=0 mcid=1
                  /Lang: "en"
    - Tag: Note
      /Id: "Note 1"
      /K:
        - Tag: Lbl
          /K:
            - Tag: Link
              /K:
                - Tag: Span
                  /BaselineShift:   2.975
                  /LineHeight:   5.100
                  /K:
                    - Content: page=0 mcid=3
                - Annotation: page=0 subtype=Link
                  /Contents: "1"
        - Content: page=0 mcid=4
    - Content: page=0 mcid=2
```

</details>

---
**(update Jan 1, 2026)** I saw the `krilla` crate (which, you know, is in charge of PDF generation since Typst 0.14) has a [snippet](https://github.com/LaurenzV/krilla/blob/7622bbda2ab9e14f7d554a94530b8d465560ed2c/crates/krilla/src/interchange/tagging/generate.toml#L295-L300) that advised the annotation should be the first child of the `Link` tag -- I didn't find the source of this advice, though, possibly in ISO 14289 which is behind a quite expensive pay wall :)

PgBiel: Hmm... Not sure if that would be harmful by itself. I suppose a more notable change is that the `/Lang: "en"` part is gone. Let's wait for further comments on the matter

Leedehai: **TL;DR**: I suggest we don't touch the PDF tags tree reference files in this PR, i.e. I suggest we merge this PR as is.

---
If interested, this is the code that aligns the PDF's behavior with the HTML, which you can patch on top of this PR and play with the `footnote-tags-basic` test:

<details>

<summary>expand for code</summary>

```rust
// crates/typst-layout/src/rules.rs

const FOOTNOTE_ENTRY_RULE: ShowFn<FootnoteEntry> = |elem, engine, styles| {
    let span = elem.span();
    let number_gap = Em::new(0.05);
    let (dest, num, body) = elem.realize(engine, styles)?;
    let alt = num.plain_text();

    /**
     * (Version 1) This PR: producing the same pdf tags tree as before:
     * - the "/Lang" part is present,
     * - "Annotation" is the first child under "Link":
     */
    // let sup = SuperElem::new(num).pack().spanned(span);
    // let prefix = PdfMarkerTag::Label(
    //    DirectLinkElem::new(dest, sup, Some(alt)).pack().spanned(span),
    // );

    /**
     * (Version 2a) Patched on top of this PR to align PDF with HTML: producing the new tags tree
     * (see comment above):
     * - the "/Lang" part is missing,
     * - "Annotation" appears after the "Span"
     */
     let link = PdfMarkerTag::Label(
         DirectLinkElem::new(dest, num, Some(alt)).pack().spanned(span),
     );
     let prefix = SuperElem::new(link).pack().spanned(span);

    /** (Version 2b) Alternative version, producing the same new tags tree as 2a */
    // let link = DirectLinkElem::new(dest, num, Some(alt)).pack().spanned(span);
    // let prefix = PdfMarkerTag::Label(SuperElem::new(link).pack().spanned(span));

    Ok(Content::sequence([
        HElem::new(elem.indent.get(styles).into()).pack(),
        prefix,
        HElem::new(number_gap.into()).with_weak(true).pack(),
        body,
    ]))
};
```

</details>

---
As for the missing `/Lang: "en"` part, my investigation reveals this is because:
* in [`typst_pdf::tags::groups::Groups::propagate_lang()`](https://github.com/typst/typst/blob/a87f4b15ca86a0b2f98948d8f393608070ed731e/crates/typst-pdf/src/tags/groups.rs#L59), the `group.kind()` is `GroupKind::Link` instead of `GroupKind::TextAttr`, and thus `group.kind().lang_mut()` is evaluated to [`Some(None)`](https://github.com/typst/typst/blob/a87f4b15ca86a0b2f98948d8f393608070ed731e/crates/typst-pdf/src/tags/groups.rs#L591) instead of [`None`](https://github.com/typst/typst/blob/a87f4b15ca86a0b2f98948d8f393608070ed731e/crates/typst-pdf/src/tags/groups.rs#L595), leading to a different code path at this [branching statement](https://github.com/typst/typst/blob/a87f4b15ca86a0b2f98948d8f393608070ed731e/crates/typst-pdf/src/tags/groups.rs#L63), which I didn't pursue further.
* Therefore, in [`typst_pdf::tags::text()`](https://github.com/typst/typst/blob/a87f4b15ca86a0b2f98948d8f393608070ed731e/crates/typst-pdf/src/tags/mod.rs#L174), the variable [`lang` (of type `Option<Locale>`)](https://github.com/typst/typst/blob/a87f4b15ca86a0b2f98948d8f393608070ed731e/crates/typst-pdf/src/tags/mod.rs#L192-L194) becomes `None` instead of `Some(Locale { ... })`, and therefore variable `lang_str` becomes `None` instead of `Some("en")`.
* Therefore, in `krilla`, `krilla::surface::Surface::start_tagged()` accepts a different `content` argument (of type `ContentTag`), and therefore `krilla::interchange::tagging::ContentTag::write_properties()` doesn't write the `Lang` property, and therefore in the resulting pdf tags tree the `/Lang: "en"` part becomes missing instead of being present.

I think this deviates far enough from the purpose of this PR, which was aimed at addressing the HTML export. Therefore, I'd suggest submitting this PR as is (i.e. keep the original behavior of the PDF, don't update the reference  YAML files), and figure out what to do with the aforementioned PDF export's code paths later.

laurmaedje:
There is one issue with this: The code now emits 
```html
<sup id="fn1"><a href="#loc-2" role="doc-noteref">1</a></sup>
```
rather than
```html
<sup><a id="fn1" href="#loc-2" role="doc-noteref">1</a></sup>
```

This goes against both the [HTML spec guidance](https://html.spec.whatwg.org/commit-snapshots/0aa021ab4f21a6550f1591a9cc98449aaea9eefa/#footnotes) and the [WAI ARIA Digital Publishing guidance](https://www.w3.org/TR/dpub-aria-1.1/#doc-noteref) and splits up the `id` and `role` attribute, which is unfortunate.

This, alongside consistency with PDF export and the fact that it's nice to be able to style the whole DOM structure via the `doc-noteref` role (without relying on CSS contains rules), were the three reasons why I implemented it as it currently is.

Given the HTML spec guidance and real world use (e.g. Wikipedia), I see how the current solution is not ideal. Looking more closely at Wikipedia, it actually puts both the ID and a class on the `<sup>`. I think this could potentially be a sweet spot for Typst too, as it makes it again easier to style and keeps ID and role together, which IMO is more semantically appropriate. Since the `href` is more an implementation detail of the display representation, I think it's less important to have it on the same element as `id` and `role`.
```html
<sup id="fn1" role="doc-noteref"><a href="#loc-2">1</a></sup>
```


---

If we change this here, I agree that it would in principle be nice if the PDF export would behave equivalently, keeping consistency. IIUC the only issue is PDF tagging, not visual output. @saecki Could you take a look at whether there is a conceptual issue or whether it's merely an implementation issue?

crates/typst-layout/src/rules.rs
saecki: > I'll defer the final word on this to @saecki in case my assumption is wrong.

Sorry, I somehow managed to completely miss this :/


> If I understood correctly, realize and the footnote entry show rules were changed to have different behavior between PDF and HTML. But I wonder if it wouldn't be fine to do the same for PDF? I suppose the superscript itself wouldn't be represented in the tag tree, right? So it'd be fine to tag a superscript of a link instead of the link itself...

The sub/superscript are currently represented by a `Span` with a `BaseLineshift` and `LineHeight`, and in PDF 2.0 there is `TextPosition` with `sup` and `sub`. But the `Span` will always be generated inside the `Link` tag due to how the text attributes are generated in the PDF export.

> That's my initial thought as well but it turned out the tag tree would change if we align the PDF behavior with HTML -- the annotation node will appear after.

Having the `Annotation` first is only an artifact of having a nested group within the `Link`; because the `Annotation` needs to be a direct child of the `Link` tag, it is pushed before the nested child group is inserted. Having the annotation directly after the link content is actually the intended way, following the [Tagged PDF Best Practice Guide](https://pdfa.org/download-area/publications/Tagged-PDF-Best-Practice-Guide.pdf) `4.2.12.2 Creation`. But either way is just a recommendation, and it doesn't really matter.

> As for the missing /Lang: "en" part, my investigation reveals this is because: ...

Yeah, you've correctly deduced this. It's a result of this TODO:
https://github.com/typst/typst/blob/f20691020af4672b8dd76410a63c3674f6f1a3c3/crates/typst-pdf/src/tags/groups.rs#L60

As a note: the additional `/Lang` attributes are just redundant, not wrong, and don't change any behavior so it should be ok.

saecki: So just to be clear, the changes to the tag tree would be fine.

Leedehai:
Done. Now the emitted HTML puts the id and role on the outer `sup`, i.e. something like
```html
<sup id="fn1" role="doc-noteref"><a href="#loc-2">1</a></sup>
```

crates/typst-layout/src/rules.rs
```diff
@@ -397,19 +397,17 @@ const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
     let span = elem.span();
     let (dest, num) = elem.realize(engine, styles)?;
     let alt = FootnoteElem::alt_text(styles, &num.plain_text());
-    let sup = SuperElem::new(num).pack().spanned(span).linked(dest, Some(alt));
-    Ok(HElem::hole().clone() + PdfMarkerTag::Label(sup))
+    let sup = PdfMarkerTag::Label(
+        SuperElem::new(num).pack().spanned(span).linked(dest, Some(alt)),
```
laurmaedje: The change here seems incomplete. This is still `link(super(..))` rather than `super(link(..))`.

crates/typst-layout/src/rules.rs
```diff
@@ -397,19 +397,17 @@ const FOOTNOTE_RULE: ShowFn<FootnoteElem> = |elem, engine, styles| {
     let span = elem.span();
     let (dest, num) = elem.realize(engine, styles)?;
     let alt = FootnoteElem::alt_text(styles, &num.plain_text());
```
laurmaedje: I think the link creation could in principle now also be moved into the `elem.realize` using `DirectLinkElem`, so that it's guaranteed the same between layout and HTML.

The only thing kept out would be the `SuperElem` because of its role assignment in HTML.

laurmaedje:
I was a bit concerned that the link target in paged export might now be too small, but it appears that it works perfectly with both typographic and synthesized scripts thanks to @MDLC01's work on sub-/superscripts. 🎉

<table>
<thead>
<th>Code</th>
<th>Output</th>
</thead>
<tbody>
<tr>
  <td>

```typ
#[
  #set super(typographic: true)

  1) Hello #super(link("..")[abc])

  2) Hello #link("..", super[abc])
]

#[
  #set super(typographic: false)

  3) Hello #super(link("..")[abc])

  4) Hello #link("..", super[abc])
]
```

</td>
<td>
image
</td>
</tbody>
</table>

Leedehai:
Done.

laurmaedje:
Thank you! This is a nice change, I'm happy with it.

PR #3522, opened 2024-02-28
Add word-break attribute for text

RanolP:
Fixes #1164. In the future, we'll need to implement [overflow-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) to cover some edge cases (we can see that the texts are clipped in the new test image).

Issue #674, opened 2023-04-08
Hard wrap table cell contents (and other blocks) without hyphenating


rktjmp:
Given a table where cells may contain long contiguous strings (eg. email addresses), the wider cells tend to "overflow" and run into their sibling cells.

The overflow can be hard clipped with `box` or `block`, at the loss of content, or hyphenation can be forced via `text(hyphenate: true)[...]` or `par(justify: true)`, but this isn't always appropriate, eg: terms may contain hyphens themselves, where `abc-xyz` is meaningfully different to `ab-c-xy-z`.

It would be great if we could force content to wrap without hyphenation.

See below where `alan.mathison.turing@university-of-cambridge.com` may be read as `alan.mathison.-turing@univer-sity-of-cam-bridge.com`.

![image](https://user-images.githubusercontent.com/866010/230729634-0a259e4b-d0da-4a16-87d9-31b7d121c465.png)

```typst
#set page(
  width: 240pt,
  height: auto,
)

The email runs out of the table cell.

#table(
  columns: 2,
  [email], [name],
  [alan.mathison.turing\@university-of-cambridge.com], [Alan Mathison Turing]
)

The email is constrained by the cell but the hyphens may be mis-read:

#table(
  columns: 2,
  [email], [name],
  text(hyphenate: true)[alan.mathison.turing\@university-of-cambridge.com], [Alan Mathison Turing]
)

Ideally:

#table(
  columns: (1fr, 1fr),
  [email], [name],
  [alan.mathison.turi
  ng\@university-of
  -cambridge.com], [Alan Mathison Turing]
)
```

Semi-related: https://github.com/typst/typst/issues/335

f3rn0s:
Just to add context back into this from #862, this happens with page content and raw blocks as well:

![232671950-5f907833-9d3e-4909-a9fa-1ba9d95a1d1e](https://github.com/typst/typst/assets/1351279/948297ba-5855-4aa0-8b23-90eb08252cae)

Enivex:
You can always add a zws anywhere you'd like to allow line break without hyphenation

aszenz:
Just ran into this, I set `text(hyphenate:true)` but it doesn't work for numbers, so how can i break apart long numbers in table cells, this example has a lot of overflow happening:
 
![Screenshot from 2024-09-06 01-21-25](https://github.com/user-attachments/assets/966f58e4-0212-4908-853f-1fbe13dbc586)

Fethbita:
[...]

PR #7642, opened 2026-01-01
feat: implement cjk-breaking property for text (closes #1164)

### LINKED ISSUE #1164: Not-so-prefered word wrapping on Korean words
### Description

Here's a 2 dummy Korean sentences:

> 뉴질랜드 북섬 호크스베이 지역에 위치한 포랑아하우에 있는 언덕인 타우마타와카탕이항아코아우아우오타마테아투리푸카카피키마웅아호로누쿠포카이웨누아키타나타후는 줄여서 타우마타와카탕이항아코아우아우오타마테아포카이웨누아키타나타후이며, 더 줄여서 타우마타라고 불리기도 한다.

> 세종대왕의 맥북프로 던짐 사건은 역사서적인 조선왕조실록에 기록된 일화로, 15세기 조선시대 세종대왕이 새로 개발한 훈민정음(한글)의 초고를 작성하던 중, 문서 작성 중단에 대한 담당자에게 분노하여 맥북프로와 함께 그를 방으로 던진 사건입니다.

And here is the render result with #335 sample texts:

image

Typst break some words that should not broken:
- 세종대왕이 => 세종대왕/이
- 맥북프로와 => 맥북/프로와
- 타우마타와카탕이항아코아우아우오타마테아포카이웨누아키타나타후이며 => 타우마타와카탕이항아코/아우아우오타마테아포카이웨누아키타나타후이며

Korean generally does not prefer breaking words, but prefers line break before a word instead. Something like [`word-break: keep-all` option in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) would be good?

Example of expected word break:

```
뉴질랜드 북섬 호크스베이 지역에 위치한 포랑아하우에 있는 언덕인
타우마타와카탕이항아코아우아우오타마테아투리푸카카피키마웅아호로누쿠포카이웨누아키타나타후는 줄여서
타우마타와카탕이항아코아우아우오타마테아포카이웨누아키타나타후이며, 더 줄여서 타우마타라고 불리기도
한다.

세종대왕의 맥북프로 던짐 사건은 역사서적인 조선왕조실록에 기록된 일화로, 15세기 조선시대
세종대왕이 새로 개발한 훈민정음(한글)의 초고를 작성하던 중, 문서 작성 중단에 대한 담당자에게 분노하여
맥북프로와 함께 그를 방으로 던진 사건입니다.
```

Or more narrow width:
```
뉴질랜드 북섬 호크스베이 지역에 위치한 포랑아하우에 있는 언덕인
타우마타와카탕이항아코아우아우오타마테아투리푸카카피키마웅아호로
누쿠포카이웨누아키타나타후는 줄여서
타우마타와카탕이항아코아우아우오타마테아포카이웨누아키타나타후이
며, 더 줄여서 타우마타라고 불리기도 한다.

세종대왕의 맥북프로 던짐 사건은 역사서적인 조선왕조실록에 기록된
일화로, 15세기 조선시대 세종대왕이 새로 개발한
훈민정음(한글)의 초고를 작성하던 중, 문서 작성 중단에 대한
담당자에게 분노하여 맥북프로와 함께 그를 방으로 던진 사건입니다.
```

Maybe related: #335 

### Reproduction URL

https://typst.app/project/rRVJfqU5x1LjoJOOtx0ZoU

### Browsers

Firefox

### OS

Linux

lamininA1:
This PR implements the requested `text.cjk-breaking` property to control line breaks in CJK text. This functionality is essential for high-quality CJK typography (especially Korean), as discussed in #1164.

### Context & Rationale
This implementation addresses the long-standing issue #1164 and supersedes previous attempts (such as #3522). It resolves previous design concerns raised by maintainers in the following ways:

1.  **Typst-native Naming:** Instead of adopting generic CSS terminology like `word-break` (which was discouraged in #3522), I introduced `cjk-breaking`. This avoids ambiguity and aligns with existing Typst properties like `cjk-latin-spacing`, clearly signaling that this property affects CJK scripts only.
2.  **Safe Scope:** The logic is strictly guarded by `unicode-script`. It only modifies behavior for **Han, Hangul, Hiragana, and Katakana**. Latin scripts (English, etc.) and other languages are completely unaffected.
3.  **Extensible Design:** Instead of a boolean, I used an `enum` (`CjkBreaking`) as suggested in previous discussions. This allows for future extensibility while keeping the API clean.
4.  **Safe Default:** The default value is `distribute` (current behavior), ensuring **no breaking changes** for existing documents. Users must explicitly opt-in to `keep-all`.

### Implementation Details
- Added `CjkBreaking` enum to `TextElem`.
- Added `is_cjk_script` helper using `unicode-script`.
- Modified inline layout preparation (`prepare.rs`) to suppress break opportunities between consecutive CJK characters when `keep-all` is active.

### Usage
This feature is controlled via the `cjk-breaking` property on the `text` element.

**1. Default Behavior (`distribute`)**
Line breaks are allowed between any characters, which is standard for Chinese and Japanese but often undesirable for Korean words.

```typ
// Default behavior (same as before)
#set text(lang: "ko")
한글 텍스트의 줄바꿈 방식을 테스트합니다.
```
2. Opt-in Behavior (keep-all) Line breaks are prohibited between consecutive CJK characters. Words are kept together, breaking only at spaces or punctuation.

```typ
// Keep CJK words together
#set text(lang: "ko", cjk-breaking: "keep-all")
한글 텍스트의 줄바꿈 방식을 테스트합니다.
```

Visual Comparison
I have verified this behavior with Korean, Chinese, and Japanese text, as well as mixed Latin text to ensure safety.

image

Relation to #674 (Overflow Handling)
I am aware that enabling keep-all can lead to content overflow if a single word exceeds the line width. However, as noted by @peng1999 in #1164, this should be treated independently from the overflow handling mechanism (#674).

Standard Behavior: Even in CSS, word-break: keep-all naturally causes overflow without overflow-wrap.

Scope: This PR strictly focuses on implementing the policy of keeping CJK words together. Handling the exception (overflow) is a separate concern that belongs to #674 or future work.

Therefore, strictly following the "Let's not let perfect be the enemy of good" philosophy, I propose merging this feature to solve the immediate readability issues for CJK users.

Checklist
- [x] The code compiles and runs.

- [x] I have tested the behavior with CJK texts.

- [x] No breaking changes for non-CJK languages.


lamininA1:
I have further expanded the logic in a separate branch to fully utilize the `enum` design.

With this update, the `cjk-breaking` property now supports three distinct behaviors:

1. **`distribute`** (Default): Breaks anywhere (standard CJK behavior).
2. **`keep-all`** (Current PR): Keeps words together but respects punctuation rules.
3. **`whitespace`** (New): A stricter mode that forces breaks *only* at whitespace characters.

I have implemented this in the following branch:
[Link to your branch](https://github.com/lamininA1/typst_kor/tree/feature/korean-keep-all)

Since this is a direct enhancement of the current PR, I would prefer to push these changes here to complete the feature set. Would that be acceptable?

laurmaedje:
Thanks for the PR. Unfortunately, I'll reject it based on the same concerns as raised in https://github.com/typst/typst/pull/3522#issuecomment-2053584199.

I'm still concerned with how to expose this. 

- The `cjk-latin-spacing` precedent is not that strong because that property is also a bit of a band-aid addition. I think we need a more general mechanism for language-specific configuration rather than adding evermore properties on `text`.

- The enum options are now in an in-between state between the CSS terminology and something custom. I personally find it hard to judge the naming, but I'm not entirely convinced by it.

> Let's not let perfect be the enemy of good

Letting perfect be the enemy of the good is sort of one of our principles and I think so far it's worked quite well because it keeps pressure for a comprehensive solution on.

Last but not least: This is not the primary factor, but I'm also a bit skeptical because of the PR description and the broken accepted test image (that overflows a lot). It kind of gives me AI vibes (also with "Link to your branch").

lamininA1:
Sorry for the AI vibes.. English is not my mother language. So I used AI a lot because I want it to look professional. Those description is AI stuff.. that's true. But the test image that I posted in here itself is valid. Although there is some minor issue, such as overflowing(which is expected behavior. Even CSS has the overflow when you use `keep-all` property), If there is some space between the word, this function working properly..  and for the Korean Users this word-breaking issue is major issue.

I understand your opinion. Typst is really fantastic program and I really love it. I hope someday we can fix this. Thank you for the review!

Issue #7269, opened 2025-10-30
Typst ignores the transparency of translucent gradients in PDF images


apcamargo:
### Description

I have a PDF figure with a translucent gradient ([`translucent-gradient.pdf`](https://github.com/user-attachments/files/23241793/translucent-gradient.pdf)) that I'm using in a Typst document. The rendered PDF ignores the transparency of the gradient in the source PDF and outputs an opaque gradient.

This is how it should look like:

<img width="260" alt="Image" src="https://github.com/user-attachments/assets/261bae70-496f-46b6-9903-ceb6187ce8b4" />

This is how it is appears in the Typst-rendered PDF:

<img width="260" alt="Image" src="https://github.com/user-attachments/assets/4782a438-ba37-4dc2-8059-e1cd1816d96a" />

This is how I'm inserting the figure into the document:

```
#figure(
  image("figures/figure-4.pdf", width: 97.5%),
  scope: "parent",
  placement: top,
  caption: [Lorem ipsum dolor sit amet.]
)
<fig-4>
```

Interestingly, this issue only happens when I apply the transparency to the box as a whole. When I set the transparency directly in the color within the gradient, it works as expected. Here is a PDF of the working version: [`translucent-gradient-working.pdf`](https://github.com/user-attachments/files/23242060/translucent-gradient-working.pdf).

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

apcamargo:
I found that the issue only happens when the transparency is applied to the whole rectangle. When I reduce the opacity in the color within the gradient, it works as expected. I updated the issue description and uploaded a PDF that works as expected in Typst.

LaurenzV:
I'm not sure I can reproduce, in the webapp I get this in the preview:

<img width="1920" height="941" alt="Image" src="https://github.com/user-attachments/assets/a9728e45-8552-4f28-98b3-22ab104ddb7e" />

and this in the final PDF:

<img width="1920" height="1043" alt="Image" src="https://github.com/user-attachments/assets/414062d7-4c9a-46d2-bcec-40755e8f51cf" />

Is that not the expected look?

Enivex:
Could be an issue with whatever pdf reader is being used

apcamargo:
I tested it in multiple PDF readers. I don't think that's the issue

apcamargo:
[...]

Issue #5509, opened 2024-12-03
SVG plot missing lines when rendered in Apple Preview, and Apple Safari but renders correctly in Typst Preview and Adobe Acrobat.


maxall41:
### Description

I have an SVG line plot created with matplotlib that renders fine in the typst preview. The lone SVG file (Not in a PDF) also renders correctly in Safari, Chrome, and Firefox. But when I export a PDF containing this SVG file and try to view it with Apple Safari or Apple Preview the lines in the plot do not show up. This does not occur with Adobe Acrobat. This issue also persists when enabling PDF/A.

Reproduction:
```
#figure(
  image("test.svg", width: 100%),
  caption: [
    caption
  ],
) <test>
```

May be related to https://github.com/typst/typst/issues/1634. SVG and PDF attached.

[Reproduce.pdf](https://github.com/user-attachments/files/17985653/Reproduce.pdf)
![test.svg](https://github.com/user-attachments/assets/c5c53ca5-f7e6-48e3-a3e7-cc5bc49b8a0f)

Edit: I almost certain this has something todo with clip paths because I was able to use Claude to convert the clip paths in the SVG into normal paths and that fixed it.

### Operating system

macOS

### Typst version

- [x] I am using the latest version of Typst

LaurenzV:
Probably similar to https://github.com/typst/typst/issues/4901#issue-2506849398.

LaurenzV:
Would be fixed by https://github.com/typst/typst/pull/5420.

reverendpaco:
[x.pdf](https://github.com/user-attachments/files/19141065/x.pdf)
![Image](https://github.com/user-attachments/assets/e3f498f1-3825-4d19-ab3b-bed5211b64b2)

Relevant reproduction of this issue.  PDF shows properly outside of Preview but missing path from Preview.   SVG generated from D2.

Only reason I'm piling on here is that _this_ SVG does not use clip-paths.  I understand this may be a bug with Apple.

reverendpaco:
Update:   

I detected the exact source of the problem:   the `mask` attribute.  

When I remove this attribute (which is on all the path elements of my SVG), the pdf now displays properly in PDFs.  So, yes inasmuch as clipping and masking are related, this problem seems to share the same DNA


<img width="823" alt="Image" src="https://github.com/user-attachments/assets/c9aad07d-7b4d-4076-a84b-7d7ec73078ba" />

LaurenzV:
Yep, the root cause are masks.

PR #7657, opened 2026-01-06
Cache SVG graphics

LaurenzV:
This PR just serves as a reference for how SVG handling should be updated once krilla is bumped the next time.

To provide some context: In typst v0.13 and earlier, we used svg2pdf to render SVGs into XObjects and embed those into the PDF. However, as can be seen in #5509, Apple Preview had an annoying bug where SVG images with masks would often render incorrectly. After some investigation, I realized that this bug could be circumvented by rendering SVGs directly into the page stream, which is what has been implemented in Typst v0.14. However, the problem with that is that 1) It makes it impossible to cache the SVGs and 2) It also doesn't isolate the contents of the graphics from the main page contents, which I think would be the correct behavior.

After doing more investigation, prompted by #7269, I realized there is another thing that can be done to fix the behavior inside of Preview even when using XObjects, which has been [ implemented in Krilla](https://github.com/LaurenzV/krilla/commit/a83e51fa26203cae3e13d1d19b21443f39581b0f). Therefore, when bumping krilla the next time, we can go back to the old behavior of caching SVG graphics.

The differences in behavior can now be observed in the following files:
v0.13: 
[v0_13.pdf](https://github.com/user-attachments/files/24450124/v0_13.pdf)

As you can see, this PDF doesn't render correctly in Preview.

Current main: 
[main.pdf](https://github.com/user-attachments/files/24450132/main.pdf)

This PDF does render correctly, but since images aren't deduplicated the file is larger.

This PR: 
[newest.pdf](https://github.com/user-attachments/files/24450142/newest.pdf)

The PDF renders correctly in Preview AND we have small file sizes due to deduplicating.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1439,6 +1439,7 @@ dependencies = [
 [[package]]
 name = "krilla"
 version = "0.6.0"
+source = "git+https://github.com/LaurenzV/krilla?rev=7622bbd#7622bbda2ab9e14f7d554a94530b8d465560ed2c"
 dependencies = [
  "base64",
  "bumpalo",
@@ -1469,6 +1470,7 @@ dependencies = [
 [[package]]
 name = "krilla-svg"
 version = "0.3.0"
+source = "git+https://github.com/LaurenzV/krilla?rev=7622bbd#7622bbda2ab9e14f7d554a94530b8d465560ed2c"
 dependencies = [
  "flate2",
  "fontdb",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,8 +77,8 @@ image = { version = "0.25.5", default-features = false, features = ["png", "jpeg
 indexmap = { version = "2", features = ["serde"] }
 infer = { version = "0.19.0", default-features = false }
 kamadak-exif = "0.6"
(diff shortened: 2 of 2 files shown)


LaurenzV:
Unfortunately this causes a regression in my master's thesis when opening the PDF in preview, so I'll have to investigate more. :(

LaurenzV:
Here's a sample document generated with Typst that now renders wrongly in Preview again... But it's a different bug, because it's not directly related to soft masks but instead to the fact that we are using layer isolation for XObjects now.  [main.pdf](https://github.com/user-attachments/files/24451384/main.pdf)

I've reported it to Apple via Feedback Assistent, but no idea whether they actually look into those.

laurmaedje:
That's unfortunate :(

laurmaedje:
Is there a reasonable path forward for the PR, i.e. any potential to have a workaround on the krilla side? I doubt Apple will do something and even if they do, old Apple OS releases are famously still in use for a long time after release.

LaurenzV:
It's been a few weeks but I think I didn't see an immediate solution. :/ Feel free to close this PR if you want to clean up.

laurmaedje:
Alright, sad. :/ It's a bit unfortunate, but leaving the PR lying around won't help either, so I'll go ahead and close it.

PR #7658, opened 2026-01-06
Accept multiple values in `array.insert`

### LINKED ISSUE #7649: Accept multiple values to insert in `array.insert`
### Description

This is probably best explained with an example:

```typ
#let arr = ("A", "B", "C")
#arr.insert(1, "x", "y", "z")
#arr // -> ("A", "x", "y", "z", "B", "C")
```

### Use Case

This seems like a simple extension to the existing `array.insert` function that is always good to have when you need it.

calm329:
Closes #7649

This PR extends `array.insert` to accept multiple values at once, allowing users to insert several elements at a specific position in a single call.

## Example
```typ
#let arr = ("A", "B", "C")
#arr.insert(1, "x", "y", "z")
#arr // -> ("A", "x", "y", "z", "B", "C")
```

Changes

- Modified array.insert to accept variadic arguments after the index
- Updated method dispatch in typst-eval to pass args directly
- Added tests for multiple value insertion (start, middle, end, negative index)

Backward Compatibility

Single-value insertion continues to work as before:
#arr.insert(1, "x") // still works

### DIFF
--- a/crates/typst-library/src/foundations/array.rs
+++ b/crates/typst-library/src/foundations/array.rs
@@ -267,11 +267,21 @@ impl Array {
         let first: Value = args.expect("value")?;
         // Any additional values
         let rest = args.all::<Value>()?;
-        // Insert in reverse order at the same index to maintain order
-        for value in rest.into_iter().rev() {
-            self.0.insert(i, value);
+
+        // Fast path: single value insertion
+        if rest.is_empty() {
+            self.0.insert(i, first);
+            return Ok(());
         }
-        self.0.insert(i, first);
+
+        // Multiple values: build new array in one pass to avoid repeated shifts
+        let values_count = 1 + rest.len();
+        let mut new = EcoVec::with_capacity(self.0.len() + values_count);
+        new.extend(self.0[..i].iter().cloned());
+        new.push(first);
+        new.extend(rest);
+        new.extend(self.0[i..].iter().cloned());
+        self.0 = new;
         Ok(())
     }
 


cady-b:
I'm a bit worried about performance with this, especially with the musl allocator used in the official builds.

It shouldn't have big effects on existing code at least, but from my looking into this earlier, I feel calling [`.insert()`](https://github.com/typst/ecow/blob/9639e9f01af758d7fde9f9e03ae89d8cbe487ae4/src/vec.rs#L304-L336) in a loop isn't the ultimate way to go about it. The `std::vec` has methods such as `splice`, [which could be used](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.splice:~:text=insert%20new%20items) and is very versatile, but would also be a pain to implement in `ecow::EcoVec`; I'm not sure if someone wants to tackle that or if implementing something simpler would end up being more worthwhile.

calm329:
> I'm a bit worried about performance with this, especially with the musl allocator used in the official builds.
> 
> It shouldn't have big effects on existing code at least, but from my looking into this earlier, I feel calling [`.insert()`](https://github.com/typst/ecow/blob/9639e9f01af758d7fde9f9e03ae89d8cbe487ae4/src/vec.rs#L304-L336) in a loop isn't the ultimate way to go about it. The `std::vec` has methods such as `splice`, [which could be used](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.splice:~:text=insert%20new%20items) and is very versatile, but would also be a pain to implement in `ecow::EcoVec`; I'm not sure if someone wants to tackle that or if implementing something simpler would end up being more worthwhile.

Good point! I've updated the implementation to avoid the loop.

For single-value insertion, it still uses the existing `self.0.insert()` directly (no regression for existing code).

For multiple values, it now builds a new array in one pass:
- pre-allocates exact capacity needed
- extends with elements before insertion point
- adds the new values
- extends with elements after insertion point

What do you think about it?

calm329:
Hi @laurmaedje I’ve opened this PR and would appreciate a review when you have a chance. Thanks!

laurmaedje:
Based on the discussion in https://github.com/typst/typst/issues/7649, this is rather unlikely to land with this API. For now, I would close it, as we'll need to reach a design decision first. Should we in the end come to the decision of supporting it like this after all, we can still reopen it. Thanks nonetheless for your work on it!

PR #7659, opened 2026-01-07
Add more math tests in different fonts

mkorje:
~~Depends on https://github.com/typst/typst-dev-assets/pull/20.~~

More tests will be added soon. Clearly copy and pasting the same thing ten times isn't particularly sustainable. But I don't want to try to modify the testing code without some guidance or prior discussion.


istudyatuni:
Probably test generation support would help with this. Something like this:

```typ
#let fonts = ("font 1", "font 2")
#for font in fonts [
  #let test-name = font.replace(regex("[^\w]+"), "-")
  --- math-font-#{test-name} ---
  #show math.equation: set text(font: font, fallback: false)
  // math expression
]
```

laurmaedje:
Why not just use a single test with a for loop?

mkorje:
Mainly because I wanted to make them all separate tests. I figured it'd make comparing images easier.

mkorje:
> Probably test generation support would help with this. Something like this:
> 
> ```typ
> #let fonts = ("font 1", "font 2")
> #for font in fonts [
>   #let test-name = font.replace(regex("[^\w]+"), "-")
>   --- math-font-#{test-name} ---
>   #show math.equation: set text(font: font, fallback: false)
>   // math expression
> ]
> ```

Will this work with how the tests are parsed/collected though?

istudyatuni:
> Will this work with how the tests are parsed/collected though?

Do you mean with the current implementation? If I remember correctly, it isn't supported, so will be just ignored

It would probably be easier to support programmatic definition of tests, like

```typ
#for /*  ..  */ {
  make-test("math-..")[
    #set ...
    // ...
  ]
}
```

By "to support" I mean "to add new functionality to test runner"

laurmaedje:
I would rather not extend the test runner. The extra complexity doesn't seem worth it to me just to have the images be separate.

alerque:
This seems more like an exercise of the fonts than a test of the layout engine. I would suggest limiting the test cases to only one instance of each unique thing being tested. If specific fonts had different features being put though the paces that's fine, but just doing everything in a matrix of all the fonts doesn't seem to make sense.

laurmaedje:
> [...] but just doing everything in a matrix of all the fonts doesn't seem to make sense.

Since fonts like to be quirky, it can easily happen that some modification to a code path seems fine with a particular font, but would break with another.

Having tests to avoid such regressions that only happen in particular combinations of font + layout engine seems sensible to me. Of course, there is the question where to draw the line and what parts to exercise multiple times vs. just once.

tests/ref/render/emph-syntax.png
mkorje, line 1: Unpredictable font fallback strikes again. This time it's the fullstop on the third last line.

tests/ref/render/math-font-fallback-class.png
mkorje, line 1: Fira Math was removed hence the change here.

tests/ref/render/math-equation-font.png
mkorje, line 1: Fira Math was removed hence the change here.

tests/ref/render/issue-7113-cjk-latin-spacing-shift.png
mkorje, line 1: Same here with the fullstop.

tests/ref/render/issue-2214-baseline-math.png
mkorje, line 1: Font fallback seems to have chosen Libertinus Math now for the double-struck W in text.

mkorje:
> If the main goal is avoiding regressions, this may actually be a good fit for the upcoming hash-based tests! These are not suitable for rendering because it's not 100% deterministic, but they will be used for PDF and SVG and we'll have comparison tooling. For now, we had no plans to get rid of the image-based tests entirely because they are easier to work with and _do_ test the rendering itself, but if the tests can't be reasonably reduced in size without compromising on coverage, it might be worth looking into how can also make this infrastructure work for such cases. Since we want to test layout here rather than rendering and SVG and PDF output test this as well.

This sounds like a much better idea! I'll reopen this then once the new hash-based testing stuff is done.

Issue #3147, opened 2024-01-07
Ability to crop an image


kuba375:
### Description

I suggest that the `image()` function should have a `crop` (or another name) argument that would crop the image. It could work a bit like the `pad()` function, for example

```typst
#image(
     "figure.jpg", 
     crop: 5pt,
)
```
would crop the image by 5pt from each side. 

Or maybe it could be a separate function, like so

```typst
#crop(
     x: 5pt,
     top: 2pt,
     image("figure.jpg"),
)
```

### Use Case

The ability to crop an image, or anything else, is useful on it's own, but when I'm writing a math paper and have a figure generated from python, the figures tend to have a white background spanning unnecessarily far. Im sure there will be other use cases tho

Enivex:
This can already be achieved by wrapping the image in a block with negative inset and clip: true

XcantloadX:
> This can already be achieved by wrapping the image in a block with negative inset and clip: true

Is it possible to crop an image into a circle?

ntjess:
Yes, you can add a `radius` to the box used during clipping:

```typst
box(pic, radius: 50%, clip: true)
```

lumpiluk:
Is it possible to crop an image with an arbitrary shape? (Our university's style guide is very particular about slopes…)

laurmaedje:
That's not currently possible. You'd have to do that in another software separately.

mpf:
Is it possible to clip from one side only?

ntjess:
Yes, if you specify a negative `inset` on the clipping box:

```typst
#let img = image("/image.png", height: 3in, fit: "cover")
#show link: it => underline(text(blue, it))

#grid(columns: 2, column-gutter: 1em)[
  Image from #link("https://unsplash.com/photos/a-lynx-standing-in-the-middle-of-a-forest-6njkAeHHMOE")[Unsplash]
  #img
][
  \ // Linebreak to align with with bottom image
  #stack(spacing: 5pt)[
    #box(img, clip: true, inset: (bottom: -1.5in), stroke: red)
  ][
    #box(img, clip: true, inset: (top: -1.5in), stroke: red)
  ] 
]
```

![image](https://github.com/user-attachments/assets/4c261e4d-0c1f-4ea4-bd63-514e061c719f)

w1th0utnam3:
[...]

PR #7666, opened 2026-01-08
Add a clipping example to the image docs

omentic:
See also https://github.com/typst/typst/issues/3147. I went looking for how to do this and pretty immediately found the GitHub issue and the (excellent) example within, but this is a common enough use case that I think it should be in the docs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3153,7 +3153,7 @@ dependencies = [
 [[package]]
 name = "typst-dev-assets"
 version = "0.15.1"
-source = "git+https://github.com/typst/typst-dev-assets?rev=79ba7c3#79ba7c354e39d9cd0a074bc41becc05f0119eee3"
+source = "git+https://github.com/typst/typst-dev-assets?rev=5a10d90#5a10d90b044e030d793e700c7c5052759621ae38"
 
 [[package]]
 name = "typst-docs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ typst-syntax = { path = "crates/typst-syntax", version = "0.15.1" }
 typst-timing = { path = "crates/typst-timing", version = "0.15.1" }
 typst-utils = { path = "crates/typst-utils", version = "0.15.1" }
 typst-assets = { git = "https://github.com/typst/typst-assets", rev = "94dcb99" }
-typst-dev-assets = { git = "https://github.com/typst/typst-dev-assets", rev = "79ba7c3" }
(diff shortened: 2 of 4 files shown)


flokl:
I don't think this is a good idea, as this code doesn't actually crop the image. The proposed code only hides the "cropped" parts of the image, and the PDF file still contains the original image file. This could mislead users into thinking they have removed sensitive parts of the image, which isn’t the case.

PgBiel:
Agreed with @flokl -- there should be a very clear disclaimer (perhaps the example should also be hidden by default with this warning above it).

Also, `box` is indeed used to make it inline (that line should **not** be changed). If that's not what you want, you should use `block` instead. So the docs should recommend `block` for clipping and other purposes.

omentic:
If I'm understanding the terminology right, is "clip" more accurate? Let me know if this wording is better, also.

laurmaedje (commented):
As is, the example is hard to follow, for one because it's mixed with the existing example, but also because it's difficult to see the effect of the different numbers. The example in https://github.com/typst/typst/issues/3147#issuecomment-2435675631 is a bit better for that.

But it also appears that adding two images, one with `inset: (right: -50%)` and one with `inset: (right: -50%)`, does not necessarily compose together into one image, so the numbers are actually really subtle to get right. Don't have time to look to deeply, but maybe it depends on the aspect ratio?

For this to be included in the official docs, the explanation would need to be easy to follow and actionable without being too long, ideally in a separate `# Clipping` section rather than attached to an existing example.

laurmaedje:
@omentic are you still interested in working on this?

omentic:
Yes, but I've been busy with schoolwork. I'll update this when I've got some time and ping for re-review.

omentic:
I have updated this to have a separate clipping example (taken directly from @ntjess) and mention clipping in the docs on `box` and `block` too. It now depends on https://github.com/typst/typst-dev-assets/pull/22.

(cc. @laurmaedje for review)

blaggacao:
In my opinion, this is the wrong strategy to solve the problem, however practicable.

The proper solution is to implement an image `crop:` attribute that works like `inset:` but dies the right thing.

That's what people and AI alike seem to expect.

Deviating from expectations via a sanctioned work-around is in my opinion worse than leaving the problem painful until a proper solution can be implemented.

Potentially, the original issue needs to be formulated in a more actionable manner. Right now it discusess mostly the work arounds and advanced use cases which seem to mask the original issue - as this PR seems to proof.

laurmaedje:
I think it's fine to add this to the docs. I rewrote the example and description the bit and downscaled and recompressed the image to reduce bloat.

Thanks!

PR #7671, opened 2026-01-10
Fix copy-paste typos in `duration` methods documentation

joshtriplett:
The documentation for the `duration` methods all mentioned "rather than
the second component of the duration", even for the methods that aren't
about the seconds. Change them to mention the correct component for each
method.

In addition, "second component" sounds like a reference to the 2nd
positional component, so change it to "seconds component".

Add commas to improve readability.


laurmaedje:
Thanks!

PR #7674, opened 2026-01-10
Add an illustration for `page.footer-descent` in docs

YDX-2147483647:
图片

A few days ago, someone read the docs in Chinese via a machine translator, and this sentence becomes quite confusing.

- Original:
  The amount the footer is _lowered into_ the bottom margin.

- Translated (and translated back):
  The distance from the footer _to_ the bottom margin.

This illustration would avoid the confusion.


laurmaedje:
Thanks, that's a nice illustration. :)

PR #7675, opened 2026-01-10
Prefer newlines over `<br>` when emitting code blocks in HTML export

wylited:
Currently, Typst-HTML render code blocks like
```c
int something(){
   return 1;
}
```

into 

```html
<pre><code>void something(){<br>   return 1;<br>}</code></pre>
```

However, it is a bad practice to use `<br>` inside `<pre>` and `<code>` blocks as it is expected to be preformatted.

This breaks syntax highlighters like highlight.js (refer to https://github.com/highlightjs/highlight.js/issues/2559).

And can break older browsers as when users try to copy code from a block, the result will not be clean.

This PR fixes it by adding case handling for code tag elements and updates the test for code blocks to be correct.

The example will render as
```html
<pre><code>void something(){
   return 1;
}</code></pre>
```
which will fix the technical issues mentioned.

### DIFF
--- a/crates/typst-html/src/convert.rs
+++ b/crates/typst-html/src/convert.rs
@@ -109,11 +109,10 @@ fn handle(
         // Nothing to do for zero-sized spacing. This is sometimes used to
         // destruct spaces, e.g. in footnotes. See [`HElem::hole`].
     } else if let Some(elem) = child.to_packed::<LinebreakElem>() {
-        if converter.whitespace == Whitespace::Pre {
-            converter.push(HtmlNode::text("\n", elem.span()));
-        } else {
-            converter.push(HtmlElement::new(tag::br).spanned(elem.span()));
-        }
+        converter.push(match converter.whitespace {
+            Whitespace::Normal => HtmlElement::new(tag::br).spanned(elem.span()).into(),
+            Whitespace::Pre => HtmlNode::text("\n", elem.span()),
+        });
     } else if let Some(elem) = child.to_packed::<SmartQuoteElem>() {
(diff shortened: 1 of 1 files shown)


laurmaedje:
I'm not sure but converting a `<br>` back into a plain newline feels a _bit_ wrong to me. As is, this would also affect code and br tags generated via `html.elem`, right?

wylited:
> I'm not sure but converting a \<br> back into a plain newline feels a bit wrong to me
> As is, this would also affect code and br tags generated via html.elem, right?

Thanks for letting me know, as it's my first time contributing I am a bit unfamiliar with the codebase and hadn't considered this edge case, I have corrected this issue in the latest commit and updated the PR title to reflect this.

Now,

`````typst
#html.elem("code")[
  we need the\
  to work here
]

```
however, here we expect a \n
instead of a br
```
`````

correctly compiles to:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <p><code>we need the<br>to work here</code></p>
    <pre><code>however, here we expect a \n
instead of a br</code></pre>
  </body>
</html>
```

laurmaedje:
I'll need to check the edge cases, but at first glance, this seems like a nicer solution!

It looks like a commit from your other PR ended up here again, that should be reverted.

wylited:
Yep, sorry I've corrected that. Thanks.

laurmaedje:
As long as there are no conflicts, there is no need to merge with main all the time. :)

laurmaedje:
Thanks!

PR #7676, opened 2026-01-11
Implement field access for `arguments`

Jollywatt:
This adds field access to `arguments` to more closely align them with `dictionary`, which has precedent in #7284. 

This is motivated by the desire to reduce the syntactical noise of scripting things like `args => args.named().left + args.named().right` or `args => args.at("left") + args.at("right")` which would become `args => args.left + args.right`.

However, this also implements `in`, so that `key in args` is equivalent to `key in args.named()`. This may be unexpected if someone could think that `value in args` means `value in args.pos()`, but I personally think the proposed behaviour is as you'd expect.

### DIFF
--- a/crates/typst-eval/src/access.rs
+++ b/crates/typst-eval/src/access.rs
@@ -84,7 +84,11 @@ pub(crate) fn access_dict<'a>(
             let span = access.target().span();
             if matches!(
                 value, // those types have their own field getters
-                Value::Symbol(_) | Value::Content(_) | Value::Module(_) | Value::Func(_)
+                Value::Symbol(_)
+                    | Value::Content(_)
+                    | Value::Module(_)
+                    | Value::Func(_)
+                    | Value::Args(_)
             ) {
                 bail!(span, "cannot mutate fields on {ty}");
             } else if typst_library::foundations::fields_on(ty).is_empty() {
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -387,6 +387,13 @@ fn missing_field_call_error(target: Value, field: Ident) -> SourceDiagnostic {
                 field.as_str(),
             ));
         }
+        Value::Args(ref args) if matches!(args.field(&field), Ok(Value::Func(_))) => {
+            error.hint(eco_format!(
+                "to call the function stored in a named argument, surround \
+                the field access with parentheses, e.g. `(args.{})(..)`",
+                field.as_str(),
+            ));
+        }
         _ if target.field(&field, ()).is_ok() => {
             error.hint(eco_format!(
                 "did you mean to access the field `{}`?",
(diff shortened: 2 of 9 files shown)


MDLC01:
I'm not sure I like `in` for `arguments` because, contrary to other methods on `arguments`, it doesn't mirror what `in` does on arrays.

laurmaedje:
I'm also not sure about `in`. I'd say, let's split this PR up so that we can land the uncontroversial thing (field access), without blocking on a decision about `in`.

Jollywatt:
Now this just implements field access.
P.S. I snuck in another change: the wording of the missing key error message to use "named argument" rather than "key".

```
"arguments do not contain key {..}" -> "no named argument {..}"
```

isuffix:
We should also probably add support for calling fields on arguments in the same way as dictionaries. It should suffice to edit `typst-eval/src/call.rs::eval_field_call` and `missing_field_call_error` and add a test or two.

Similar tests are in `suite/foundations/dict.typ` and `suite/scripting/methods.typ`. If you're willing, it would also probably be good to move the dict method call tests to `methods.typ` to reduce the number of places to look for them (also `array-bad-method-lvalue` and `array-unknown-method-lvalue`)

isuffix:
Another trickier question is whether we want to allow field mutation on arguments in general, i.e. `args.field += value`. This does kind of make sense to me, but it would require many more changes in `typst-eval`.

laurmaedje:
@Jollywatt with these additional complications that have come up, do you still plan to work on this?

Jollywatt:
> @Jollywatt with these additional complications that have come up, do you still plan to work on this?

Thanks for the reminder, I had just forgotten about this…

I've taken on most of @isuffix's suggestions so that more error messages are uniform:
```typ
#let dict = (foo: () => "result", bar: 42)
#let args = arguments(foo: () => "result", bar: 42)

// type dictionary has no method `bar`
// hint: did you mean to access the field `bar`?
#dict.bar()
// type arguments has no method `bar`
// hint: did you mean to access the field `bar`?
#args.bar()
// type dictionary has no method `foo`
// hint: to call the function stored in the dictionary, surround the field access with parentheses, e.g. `(dict.foo)(..)`
#dict.foo()
// type arguments has no method `foo`
// hint: to call the function stored as a named argument, surround the field access with parentheses, e.g. `(args.foo)(..)`
#args.foo()
#(dict.foo)() // "result"
#(args.foo)() // "result"
```
But I haven't touched anything regarding making `arguments` mutable.
```typ
#{
	// cannot mutate fields on arguments
	args.bar = 3
	// cannot mutate a temporary value
	args.named().bar = 3
}
```

Jollywatt:
This has been waiting for review for a month, and now has some merge conflicts with main, so I will just rebase the changes…

laurmaedje:
Thanks for your work on this! I still wanted to land this in 0.15 and wasn't sure whether we'd manage to in time with more review rounds, so I asked @isuffix to bring it over the finish line in https://github.com/typst/typst/pull/8407. Me giving timely reviews would have been the better way, but unfortunately I didn't manage to.

PR #7679, opened 2026-01-12
Fix image glyph scaling issues in SVG

saecki:
This basically removes a duplicate and somewhat broken implementation of the ttf image glyph conversion, that was part of `typst-svg` and reuses the existing implementation that was already part of `typst-library` and used by `typst-render`.


crates/typst-library/src/text/font/color.rs
```diff
@@ -307,19 +318,43 @@ fn make_svg_unsized(svg: &mut String) {
         }
     }
 
-    // Remove the `viewBox` attribute.
-    if let Some(range) = viewbox_range {
-        svg.replace_range(range.clone(), &" ".repeat(range.len()));
-    }
+    let svg_body_start = s.cursor();
+    let Some(svg_body_end) = memchr::memmem::rfind(svg.as_bytes(), b"</svg>") else {
```
laurmaedje: could this just be `svg.rfind("</svg>")`? since everything is strings.

crates/typst-svg/src/text.rs
```diff
@@ -1,221 +1,157 @@
...
+use typst_library::visualize::{FillRule, Paint, RelativeTo};
 
-use crate::{SVGRenderer, State, SvgMatrix, SvgPathBuilder};
+use crate::{DedupId, SVGRenderer, State, SvgMatrix, SvgPathBuilder};
+
+/// Represents a glyph to be rendered.
+#[derive(Clone)]
+pub enum RenderedGlyph {
+    /// A frame that contains an image glpyh.
+    Image(GlyphFrame),
```
laurmaedje: Maybe call this variant `Frame` instead of `Image` since it can also be the tofu variant?

crates/typst-svg/src/text.rs
```diff
@@ -1,221 +1,157 @@
...
-        let path = convert_outline_glyph_to_path(&text.font, glyph_id, scale)?;
-        let id = self
-            .glyphs
-            .insert_with((&text.font, glyph_id, scale), || RenderedGlyph::Path(path));
+        x_offset: Abs,
+        y_offset: Abs,
+        id: DedupId,
+    ) {
+        // Apply the transform here, because the state transform is used to draw
+        // stokes and fills with gradients and tilings.
```
laurmaedje: stokes -> strokes

crates/typst-svg/src/text.rs
```diff
@@ -1,221 +1,157 @@
...
+        let state = state.pre_concat(Transform::translate(x_offset, y_offset));
+
+        let Some(glyph_size) = text.font.ttf().glyph_bounding_box(glyph_id) else {
+            // TODO: This shouldn't happen, because the glyph has been
```
laurmaedje: The "TODO" part seems not so actionable. Should this be just a normal comment?

crates/typst-svg/src/text.rs
```diff
@@ -1,221 +1,157 @@
...
-        self.xml.end_element();
-
-        Some(())
+        glyph_id: GlyphId,
+        x_offset: Abs,
+        y_offset: Abs,
+    ) {
+        if should_outline(&text.font, glyph_id) {
+            // Pre-scale outlined glyphs, so strokes and fill patterns don't
+            // need to considerate text size glyph scaling.
```
laurmaedje: considerate -> consider

crates/typst-utils/src/hash.rs
```diff
@@ -13,6 +13,30 @@ pub fn hash128<T: Hash + ?Sized>(value: &T) -> u128 {
     state.finish128().as_u128()
 }
 
+/// Calculate a 128-bit siphash of a value. To make the hash stable between
+/// 64-bit and 32-bit architectures, usize is hashed as u64.
+pub fn stable_hash128<T: Hash + ?Sized>(value: &T) -> u128 {
```
laurmaedje: I wonder whether we should just make the default `hash128` function do that? I'd imagine the performance impact to be pretty small and that way we dodge such issues right away in the future. What do you think?

saecki: Sure, sounds good!

PR #7680, opened 2026-01-12
SVG improvements

saecki:
This is a major refactor of `typst-svg` to improve/reduce the size of generated SVGs, and make them deterministic.

Apart from the commit titles, here a list of changes, which is also included in the `Refactor typst-svg` commit:
- Custom `SvgElement` wrapper around `XmlWriter`
    - Removes the need of `end_element` due to `Drop` impl
    - Use `itoa` and `ryu` for number formatting
    - Round floats to 9 decimal digits
        - this removes long formatted numbers due to rounding errors
    - `SvgWrite` and `SvgDisplay` traits analogous to `std::fmt`
        - Force going through custom number formatting
        - Avoid going through `std::fmt::Formatter`, which has a cost (due to vtable lookups)
    - Round numbers used in colors instead of using fixed decimal format specifiers
        - Avoid redundant decimals such as `0.00000`
- Use `Point`, `Size`, `Angle`, etc. for path computations
    - These primitves are backed by f64, which removes some inaccuracies
      that lead to long formatted numbers

### DIFF
--- a/crates/typst-svg/src/lib.rs
+++ b/crates/typst-svg/src/lib.rs
@@ -21,17 +21,23 @@ use typst_library::layout::{
     Transform,
 };
 use typst_library::visualize::{Geometry, Gradient, Tiling};
-use xmlwriter::{Options, XmlWriter};
+use xmlwriter::XmlWriter;
 
 use crate::paint::{GradientRef, SVGSubGradient, TilingRef};
 use crate::text::RenderedGlyph;
 use crate::write::{SvgDisplay, SvgElem, SvgIdRef, SvgTransform, SvgUrl, SvgWrite};
 
+const XML_WRITE_OPTIONS: xmlwriter::Options = xmlwriter::Options {
+    use_single_quote: false,
+    indent: xmlwriter::Indent::Spaces(2),
+    attributes_indent: xmlwriter::Indent::None,
+};
+
 /// Export a frame into a SVG file.
 #[typst_macros::time(name = "svg")]
 pub fn svg(page: &Page) -> String {
     let mut renderer = SVGRenderer::new();
-    let mut xml = XmlWriter::new(Options::default());
+    let mut xml = XmlWriter::new(XML_WRITE_OPTIONS);
     let mut svg = svg_header(&mut xml, page.frame.size());
 
     let state = State::new(page.frame.size());
@@ -44,7 +50,7 @@ pub fn svg(page: &Page) -> String {
 #[typst_macros::time(name = "svg frame")]
 pub fn svg_frame(frame: &Frame) -> String {
     let mut renderer = SVGRenderer::new();
-    let mut xml = XmlWriter::new(Options::default());
+    let mut xml = XmlWriter::new(XML_WRITE_OPTIONS);
     let mut svg = svg_header(&mut xml, frame.size());
 
     let state = State::new(frame.size());
@@ -65,7 +71,7 @@ pub fn svg_html_frame(
     let mut renderer = SVGRenderer::with_options(Some(introspector));
     let mut xml = XmlWriter::new(xmlwriter::Options {
         indent: xmlwriter::Indent::None,
-        ..Default::default()
+        ..XML_WRITE_OPTIONS
     });
     let mut svg = svg_header_with_custom_attrs(&mut xml, frame.size(), |svg| {
         if let Some(id) = id {
@@ -106,7 +112,7 @@ pub fn svg_merged(document: &PagedDocument, gap: Abs) -> String {
         + document.pages.iter().map(|page| page.frame.height()).sum::<Abs>();
 
     let mut renderer = SVGRenderer::new();
-    let mut xml = XmlWriter::new(Options::default());
+    let mut xml = XmlWriter::new(XML_WRITE_OPTIONS);
     let mut svg = svg_header(&mut xml, Size::new(width, height));
 
     let mut y = Abs::zero();


crates/typst-svg/src/lib.rs
```diff
@@ -212,11 +212,11 @@ impl<'a> SVGRenderer<'a> {
             introspector,
             glyphs: Deduplicator::new('g'),
             clip_paths: Deduplicator::new('c'),
-            gradient_refs: Deduplicator::new('g'),
             gradients: Deduplicator::new('f'),
+            gradient_refs: Deduplicator::new('t'),
```
laurmaedje: is it intentional that now `gradient_refs` and `tilings` both use `t`?

crates/typst-svg/src/text.rs
```diff
@@ -131,24 +139,27 @@ impl SVGRenderer<'_> {
...
-                aspect_ratio,
-                self.text_paint_transform(&state, &stroke.paint),
-            );
-        }
-        self.xml.end_element();
+        svg.elem("use")
+            .attr("xlink:href", SvgIdRef(id))
+            .attr("x", x_offset.to_pt())
+            .attr("y", y_offset.to_pt())
+            .with(|svg| {
```
laurmaedje: I think storing the `use` element in a binding might be a bit clearer here (as is done for `let mut defs = svg.elem("defs");`). The `with` thing seems a bit over-engineered to me.

crates/typst-svg/src/text.rs
```diff
@@ -170,48 +181,42 @@ impl SVGRenderer<'_> {
...
+        defs.attr("id", "glyph");
 
         let glyphs = std::mem::take(&mut self.glyphs);
         for (id, glyph) in glyphs.iter() {
             let Some(glyph) = glyph else { continue };
 
-            self.xml.start_element("symbol");
-            self.xml.write_attribute("id", &id);
-            self.xml.write_attribute("overflow", "visible");
+            let symbol = &mut defs.elem("symbol");
```
laurmaedje: Personally, I'm not a big fan of temporary lifetime extension. To me `let mut symbol = defs.elem("symbol");` is easier to read and, in particular, it's clearer to reason about when the `Drop` impl will actually run.

crates/typst-svg/src/write.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// Write an integer.
+    fn push_int(&mut self, num: i64) {
+        let mut buf = itoa::Buffer::new();
+        self.push_str(buf.format(num));
+    }
+
+    /// Convenience method to write a list of space separated numbers.
+    fn push_nums(&mut self, nums: impl IntoIterator<Item = f64>) {
```
laurmaedje, line 127: how about this?
```rust
fn push_comma_list<T: SvgDisplay>(&mut self, items: impl IntoIterator<Item = T>) {
```
maybe it's unnecessary if this only ever happens with numbers though

crates/typst-svg/src/path.rs
```diff
@@ -0,0 +1,197 @@
+use ecow::EcoString;
+use typst_library::layout::{Abs, Angle, Point, Ratio, Size};
+use typst_utils::defer;
+
+use crate::write::{SvgFormatter, SvgWrite};
+
+/// A builder for SVG path using relative coordinates.
+pub struct SvgPathBuilder {
+    pub path: EcoString,
```
laurmaedje: having this private + a `fn finish(self) -> EcoString` would seem cleaner to me

crates/typst-svg/src/paint.rs
```diff
@@ -122,114 +128,103 @@ impl SVGRenderer<'_> {
...
-                        self.xml
-                            .write_attribute_fmt("stroke", format_args!("url(#{id})"));
-                        self.xml.write_attribute("stroke-width", "0");
-                        self.xml.write_attribute("shape-rendering", "optimizeSpeed");
-                        self.xml.end_element();
+                        pattern
+                            .elem("path")
+                            .attr("d", builder.path)
+                            .attr("fill", SvgUrl(id))
+                            // .attr("stroke", SvgUrl(id))
```
laurmaedje: what's up with this?

crates/typst-svg/src/lib.rs
```diff
@@ -473,210 +468,26 @@ impl<T> Deduplicator<T> {
...
-    fn with_translate(pos: Point) -> Self {
-        Self {
-            path: eco_format!("M {} {}", pos.x.to_pt(), pos.y.to_pt()),
-            scale: Ratio::one(),
-            last_close_point: pos,
-            last_point: Point::zero(),
-        }
-    }
+        // SAFETY: The digits are all valid ASCII hex characters.
+        let str = unsafe { std::str::from_utf8_unchecked(&digits) };
```
laurmaedje: doesn't seem worth it to me to use unsafe here. couldn't you push char by char and already skip the zeros in the for loop? alternatively `std::str::from_utf8(..).unwrap()` is also fine by me.

crates/typst-svg/src/write.rs
```diff
@@ -0,0 +1,269 @@
...
+impl Drop for LazySvgElem<'_, '_> {
+    fn drop(&mut self) {
+        if self.initialized {
+            self.inner.xml.end_element();
+        }
+    }
+}
+
+pub trait SvgWrite: Sized {
+    /// Write a string, escaping of quotes is handled.
```
laurmaedje: what does this comment mean? I can't see any quote escaping.

crates/typst-svg/src/lib.rs
saecki: No, I'm just stupid :melting_face:

crates/typst-svg/src/write.rs
saecki: It also happens with other items, but then the items are usually heterogeneous. I also though about implementing `SvgDisplay` for a set of tuples (maybe up to 6 elements). So one could use `f.push((a, b, c, d))`. To make that more useful we could add some formatting wrappers, such as `Em`, `Pt`, `Percent`. But that may be a bit over the top :thinking:

crates/typst-svg/src/write.rs
saecki: Quote escaping is handled by `xmlwriter`, that should probably be clarified.
I just wanted to mention that, because `xmlwriter` does that in a second pass, after the data is appended to the buffer and it's a bit hidden.

crates/typst-svg/src/write.rs
laurmaedje: let's skip it for now

crates/typst-svg/src/paint.rs
saecki: Oh, it seemed odd to me to set a stroke here. This feels like it might be much more computationally expensive than just a fill, and I couldn't see any difference. I'll remove it completely if the comparison tool doesn't complain.

saecki:
Here's a diff generated with the [typst-svg-cmp.sh](https://github.com/user-attachments/files/24588975/typst-svg-cmp.sh) comparing against https://github.com/typst/typst/commit/11fd40743fb38c375fe28fd15a6c9a5b592d9061: [diff.zip](https://github.com/user-attachments/files/24589025/diff.zip)

I think the only significant change is `tiling-small`, but that is actually a fix:
tiling-small

crates/typst-svg/src/paint.rs
saecki: I decided to add `stroke="none"`

crates/typst-svg/src/lib.rs
saecki: I thought, since there are quite a lot of `DedupId`s I'd make the hex encoding fast; converting and pushing char by char, is a lot more work. This optimizes quite nicely to a few SIMD instructions, but I guess it's not *that* important, so I'll just put an `unwrap`.

PR #7682, opened 2026-01-12
Hint about unsupported network access when trying to load from an URL-like path

### LINKED ISSUE #7669: Add hint when attempting to fetch an image from a URL
### Description

I want to use # image ("") to display web pages on HTTP, but I found that there is a problem. Does typst not support it? I suggest adding this feature!

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

wylited:
Fixes #7669 as recommended.

I don't suppose so, but let me know if more cases should be covered, for example `ftp://`?

### DIFF
--- a/crates/typst-cli/tests/smoke.rs
+++ b/crates/typst-cli/tests/smoke.rs
@@ -185,6 +185,17 @@ fn test_path_to_package() {
     output.stderr.must_contain("error: panicked with: (7, 42)");
 }
 
+#[test]
+fn test_network_access_hint() {
+    // Using a CLI test because the error message differs across operating
+    // systems. If the test runner could handle that, we could migrate to a
+    // normal test.
+    let project = tempfs();
+    let main = project.write("main.typ", "#image(\"https://example.org/image.png\")");
+    let output = exec().arg("compile").arg(main).must_fail();
+    output.stderr.must_contain("hint: network access is not supported");
+}
+
 /// Executes a command with the Typst CLI.
 fn exec() -> Command {
     Command::new(env!("CARGO_BIN_EXE_typst"))


crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -72,6 +72,11 @@ pub struct ImageElem {
     #[required]
     #[parse(
         let source = args.expect::<Spanned<DataSource>>("source")?;
+        if let DataSource::Path(path) = &source.v {
+            if path.starts_with("http://") || path.starts_with("https://") {
+                bail!(source.span, "network access is not supported"; hint: "download the image and place it as a file next to the document");
```
cady-b: This currently uses `bail!` which would immediately return from the function and is a hard error.

I'm not certain this is desired as some targets (i.e. Linux) are fairly lax in what they allow within paths, we wouldn't want to impose restrictions beyond those necessitated by the security model. It seems currently Typst converts the `//` to `/`, which is sensible regarding correctness in POSIX / with multi-platform stuff. It also causes the tests added with this PR to result in valid reads however; placing a file at `http[s]:/example.com/image.png` lets them compile just fine. The patch is a regression in that sense as we now get an error.

We could just emit a `warning!` into `engine.sink.warn()` and let the fs error separately. I believe the hint could also be attached directly to the result of the read below, but I don't know how to do that on a `SourceResult`; it might have to be implemented within the cli's `World` itself.

wylited: Thanks for the suggestion, I believe I have corrected the regression introduced by the patch in e688ff0. 
The hint should also be attached directly to the result of the read below.

Also by the POSIX comment, were you referring to how we could have a directory named http://, and thus it would be a regression as it wouldn't allow that image to be included?
Does adding the warning directly to the source result allow us to only emit the warning iff the path is not found?

Thank you again.

laurmaedje: That looks like a better approach. We could also move this into `impl Load for Spanned<&DataSource>` to make it apply more generally to functions that load data.

cady-b: The POSIX (stand-in for Linux as that's not as nicely documented) comment was me reasoning about pathname resolution, where `//` in the middle of a path is considered equivalent to `/`[^posix-pathname]. Thus, our behavior of treating something that may look invalid at first as a valid path is okay. `https://` isn't a directory we can actually create, but `https:/` is, and it does resolve via `http://` as demonstrated :sweat_smile:

I've had a look at the revised version of the PR. The regression was indeed resolved, and we do only get a hint iff the file really couldn't get loaded—that's great! Sadly the bit I said about `SourceResult` (the alias isn't loaded here, so `Result<T, EcoVec<SourceDiagnostic>>`) applies. Each `SourceDiagnostic` is a separate error / warning with an origin, hints, and other stuff attached; we're currently overwriting all the ones returned from this function with our custom data. I don't think more than one error will be returned in practice, but if a different FS error comes up, we may lose valuable information, especially since we're overwriting the `message` and not just appending hints.

I'm still not sure how to go about this issue; going as far as the CLI's `World` like I initially thought seems to be unnecessary. We can intercept the plain errors here: https://github.com/typst/typst/blob/3badcac5bba55bd0e76ff065a872fe3dfc7d940e/crates/typst-library/src/loading/mod.rs#L87-L101

Especially the `Result` generated on `let data = …` would be relevant to have this hint on. But that will apply to all ways to load data from a file, not just `image`. Alternatively, we can just not overwrite the `message` of the `SourceDiagnostic` (-s) we get back.

[^posix-pathname]: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap04.html#tag_04_16 (though it appears to not be explicitly stated and rather assumed from the other definitions in this section. All systems I know about behave like this at least).

cady-b: oops it looks like Laurenz was faster, the github UI hadn't refreshed yet :sweat_smile:

laurmaedje: oh, I didn't see the part that overrides the messages. that should indeed not be done. adding a hint should be sufficient. I also think just one hint ("network access is not supported") is sufficient and that also resolves the issue of the second hint being image specific.

In `impl Load for Spanned<&DataSource>` it's not yet a `SourceResult`, but rather a `FileResult`, so there we can use the existing `.hint("...")` infrastructure.

wylited: I've added the hint into `impl Load for Spanned<&DataSource>` now.

crates/typst-library/src/loading/mod.rs
```diff
@@ -88,7 +88,14 @@ impl Load for Spanned<&DataSource> {
         match &self.v {
             DataSource::Path(path) => {
                 let file_id = self.span.resolve_path(path).at(self.span)?;
-                let data = world.file(file_id).at(self.span)?;
+                let data = world.file(file_id).at(self.span).map_err(|mut diags| {
+                    if path.starts_with("http://") || path.starts_with("https://") {
```
istudyatuni:
```suggestion
                    const PROTOCOLS: &[&str] = &["http", "https"];
                    if PROTOCOLS.iter().any(|p| path.starts_with(format!("{p}://"))) {
```

laurmaedje: Seems a bit like over-engineering to me tbh

istudyatuni: For 2 protocols - probably yes. Anyway, this is just a suggestion

laurmaedje:
Thanks!

Issue #7677, opened 2026-01-12
Easier petite caps access, and other cases (titling case and unicase) 


jassielof:
Similar to how there's a small caps function for the respective font features, I feel if it would be possible to also provide a petite caps function for pcap font feature.

Either a function or cramble it as an extension on the small caps function with a petite boolean. **As both small and petite are closely related**.

The dedicated function would be well _much clearer_ and that, but _I think_ would require duplication of the same functions, probably.

As for the petite boolean property, would feel a bit hidden by the main small caps function but yeah it's a bit assumed that petites are much smaller, and the same configuration would be shared with a show-where rule, such as `#show smallcaps.where(petite: true)`. 

And related to that I'm a bit working on the fake small caps function too so is related, in terms of faking it, as they are simply much smaller (petite) caps.

Related: #7009.

## Use Case

Basically, for petite caps.

Some fonts that support it are EB Garamond and Mrs Eaves.

## Relevant

It might also be relevant to probably consider other case variations such as unicase and titling case (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-variant-caps). Which thinking about it, could lead to a single element like case? That would have small, petite, uni, titling, lower, upper, etc., instead of all being divided.

Could be indirectly relevant #2405, in the case that it also support fake petite caps, as these petite caps tend mostly to be of lower case height or x-height. 

## References

- https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#pcap
  - https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#c2pc
- https://www.wikiwand.com/en/articles/Small_caps
- Unicase: https://learn.microsoft.com/en-us/typography/opentype/spec/features_uz#unic
- Titling: 
  - https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#titl
  - https://www.wikiwand.com/en/articles/Titling_capitals

### Titling

Titling seems like a sane default that could be used pre-checking for headings and the title element.

It could be easily explained in the docs or possibly added as a boolean property, and to disabled it with a boolean or `show title: set text(feature: ...)`

Issue #2405, opened 2023-10-16
```ex``` as a unit

### ISSUE THREAD

**Issue by JamesxX** [typst/typst#2405]
### Description

```em``` units are used for lengths involving spacing or that which otherwise depends on the horizontal size of the font. For those cases involving vertical spacings, the unit typically used is ```ex``` (though people will sometimes substitute that for ```em```).

```ex``` is roughly the height of an 'x' in the current font.

### Use Case

Measuring vertical spacing where the relationship with the font height is important.

**Comment by Enivex** [comment]
Relevant: https://discord.com/channels/1054443721975922748/1088371919725793360/1121918053735731340

**Comment by rwmpelstilzchen** [comment]
For reference, see [the Wikipedia article about x-height](https://en.wikipedia.org/wiki/X-height).

This is relevant for all kinds of typesetting matters, but also specifically for [this](https://github.com/lonkaars/typst-metalogo/blob/08e330be164790c2374684b5e5aa9f86aa2fe108/metalogo.typ#L1-L2)… 🙃

**Comment by EpicEricEE** [comment]
> but also specifically for [this](https://github.com/lonkaars/typst-metalogo/blob/08e330be164790c2374684b5e5aa9f86aa2fe108/metalogo.typ#L1-L2)… 🙃

Fwiw, for simple contained cases like this, you should be able to get away with `measure` and `top-edge: "x-height"`.
In more general cases like global set rules for spacings, this won't work well though.

**Comment by Bugg4** [comment]
I would find `ex` units very useful. Are there any plans to work on this?

**Comment by garulfogb** [comment]
It would also be helpful to define interline and interparagraph spacings.
Relative to horizontal font size: em
Relative to vertical font size: ex

**Comment by YDX-2147483647** [comment]
Here's the workaround [suggested by Eric Biedert above](https://github.com/typst/typst/issues/2405#issuecomment-2308846441). Explanations can be found on the [Forum](https://forum.typst.app/t/how-to-setup-the-interlinespacing-parameters-as-function-of-x-height/6662/3?u=y.d.x).

```typst
// Tested against typst v0.14.0
#show: rest => context {
  let x-height = measure({
    set text(bottom-edge: "baseline", top-edge: "x-height")
    [x]
  }).height
  let ex = x-height / 1em.to-absolute() * 1em // Make it relative to the font size.


  // Now use ex freely
  set par(leading: 1.2 * ex)

  rest
}
```

This workaround does not involve the following issue mentioned by Lauren Mädje, because it always uses the font when `x-height` is measured.

[...]

Issue #394, opened 2023-03-28
Fake text weight, text style request


1216892614:
Classic Chinese fonts often do not include weight and style. Fake text weight, text style can really help.

rapiz1:
xelatex has the `AutoFakeBold` option built-in. But I didn't find where I can set options for fonts in typst. Maybe adding a new argument for text, named `auto-fake-bold`. @laurmaedje What do you think?

laurmaedje:
I can imagine having support for synthesized bold & italic, but it's non-trivial to implement.

rapiz1:
I think an alternative solution is to expose stroke for text to users, then user-defined functions can be used to make fakebold. However, I found that TextRenderingMode is set in pdf-writer and not exposed to typst.

rapiz1:
> I think an alternative solution is to expose stroke for text to users, then user-defined functions can be used to make fakebold.

## Fakebold Demo 

![image](https://user-images.githubusercontent.com/20105061/234943768-f2e8f9ce-9569-41fa-b29c-09047c1cb0a1.png)

Left: fake bold(Stroke::default() in this case)
Right: original

Just played with the source and it seems working.

## Proposed Solution

Add `stroke` parameter to `text`, just as the one in `overline`.

When `stroke` is not `None`, call `Content::set_text_rendering_mode(pdf_writer::types::TextRenderingMode::FillStroke);` and set stroke.

@laurmaedje Does this idea look fine to you? If so, I continue to work on this and open a PR.

laurmaedje:
I'm not sure this works as you want it to since glyph forms are designed for being outlined, not stroked. Moreover, we also need to be able to render to an image, not only as PDF.

rapiz1:
https://ctan.org/pkg/xfakebold uses stroke to make fakebold. It doesn't work for all fonts. But in this way users get what they expect for latex.

Export for image may be an issue. There could be some inconsistency. Do you think it's a must to keep the rendered result exactly the same?

Enter-tainer:
For export to image, tiny_skia has `stroke_path`, https://docs.rs/tiny-skia/latest/tiny_skia/struct.PixmapMut.html#method.stroke_path. Maybe it is possible to make fakebold path by using it.

> glyph forms are designed for being outlined, not stroked

Indeed, that's why it is called "fake bold".  MS Word employs similar methods when a font does not have a bold variant available.

laurmaedje:
[...]

PR #7684, opened 2026-01-12
Add fake small caps support

### LINKED ISSUE #7009: Simulate fake small caps
### Description

Add the possibility for Typst to support synthetical (fake) small caps.

#### Proposed design

```typ
#smallcaps(
  synth: auto | never | always = never
)
```

- Auto: Check whether or not the font has support for it and use it accordingly.
- Never (default): Don't use fake small caps, fallback to current behavior (up to the font if it has support for it).
- Always: Forces the use of fake small caps, even if the font has support for it.

LaTeX can simulate this by manipulating the font scale with a XeLaTeX or LuaLaTeX compiler with either `\relscale{0.75}` or `\scalebox{0.75}[0.75]`.

### Use Case

Yes, they lack the nuanced stroke weight and spacing of true small caps (affecting legibility and hierarchy), **but** they can be useful as valuable fallbacks.

It also might be _somewhat useful_ for the HTML export, considering CSS has it via `font-synthesis-small-caps: auto;`.

jassielof:
# Fake small caps

Support for synthesized (fake) small capitals when fonts don't provide native OpenType small capitals glyphs. A new `typographic` parameter is added to the `smallcaps` function, consistent with the existing `sub` and `super` functions.

Fixes #7009. 

## Example

```typst
#set text(font: "Some Font Without Smallcaps")

// Will automatically synthesize since the font lacks smcp
#smallcaps[Hello World]

// Force synthesis even if font has smcp
#smallcaps(typographic: false)[Hello World]

// Works with 'all' parameter
#smallcaps(all: true)[UNICEF and UN]
```

## Changes

Fonts without `smcp` will now synthesize instead of showing regular text.

## Considerations

This is an initial implementation. Some features I'm doubting are:

- **Configurable scaling** (that only affect the small caps) such as:
  - Size/scaling, similar to sub/super scripts.
  - Expansion/width (such as `scale()`)
  - Glyph stroke (stroke)
  - Vertical shift (baseline)
- **Warning when synthesis is used (deferred due to potential noise):** I decided to not include a warning at least for now since it would be noisy/inconsistent compared with sub/super scripts, which don't warn when synthesizing either.

I'd like to know if this is fine as is, or well, the features from above are needed.

### HTML

The following example could be used for small caps support in HTML, as well as fake ones, considering it can be easily style-inlined.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>IBM Plex Serif - Fake Small Caps Demo</title>
    <link rel="preconnect" href="<https://fonts.googleapis.com>">
    <link rel="preconnect" href="<https://fonts.gstatic.com>" crossorigin>
    <link href="https://fonts.googleapis.com/css2?family=IBM+Plex+Serif:wght@400;600&display=swap" rel="stylesheet">
    <style>
        body {
            font-family: 'IBM Plex Serif', serif;
            max-width: 800px;
            margin: 50px auto;
            padding: 20px;
            line-height: 1.6;
            color: #333;
        }
        h1 {
            font-weight: 600;
            margin-bottom: 30px;
        }
        .example {
            margin: 30px 0;
            padding: 20px;
            background: #f5f5f5;
            border-left: 4px solid #0066cc;
        }
        code {
            background: #e0e0e0;
            padding: 2px 6px;
            border-radius: 3px;
            font-family: monospace;
        }
    </style>
</head>
<body>
    <h1>IBM Plex Serif with Fake Small Caps</h1>

    <div class="example">
        <h3>Normal Text</h3>
        <p>This is regular text using IBM Plex Serif font.</p>
    </div>

    <div class="example">
        <h3>Fake Small Caps (Inline Style)</h3>
        <p style="font-variant: small-caps;">This text uses font-variant: small-caps, which creates fake small caps since IBM Plex Serif lacks true small caps.</p>
    </div>

    <div class="example">
        <h3>Manual Fake Small Caps (Better Control)</h3>
        <p><span style="text-transform: uppercase; font-size: 0.8em; letter-spacing: 0.05em;">This approach</span> gives you more control over the appearance by combining uppercase transformation with size reduction and letter spacing adjustments.</p>
    </div>

    <div class="example">
        <h3>Implementation Notes</h3>
        <p>Since IBM Plex Serif doesn't have OpenType small caps features, you have two options:</p>
        <ul>
            <li><code>font-variant: small-caps</code> - Browser synthesizes small caps automatically</li>
            <li><code>text-transform: uppercase</code> with reduced <code>font-size</code> - Manual approach with better control</li>
        </ul>
        <p>The manual approach typically looks better as you can fine-tune size (usually 70-85% of normal), letter-spacing, and ensure consistent baseline alignment.</p>
    </div>

    <div class="example">
        <h3>Mixed Example</h3>
        <p><span style="text-transform: uppercase; font-size: 0.75em; letter-spacing: 0.08em;">Business Intelligence</span> for <span style="text-transform: uppercase; font-size: 0.75em; letter-spacing: 0.08em;">Event Salons</span></p>
    </div>
</body>
</html>

```

If irrelevant I could move it to the issue, I guess.

### Dependent on other issues and relevant

These could be possibly migrated once they are implemented:

- [#7677](https://github.com/typst/typst/issues/7677): Petite caps, either adding a new `petitecaps()` function or extending it with `smallcaps(petite: boolean)`
- [#394](https://github.com/typst/typst/issues/394): Fake weight which might help specially on serif fonts (stroke thickness works fine for sans-serif fonts), since these tend to maintain its horizontal traces thinness.
- [#2405](https://github.com/typst/typst/issues/2405): As better measurement and default scaling alternative instead of em-unit, as small caps tend to use lower case height as reference (mostly).


Enivex:
Regarding petite caps, I think anything other than a separate function would make it very inconvenient.

Alternatives to `petitecaps()` would be `smallercaps()` or `tinycaps()`. But if petite caps is standard terminology I guess we should go with that.

fake-small-caps-test.typ
jassielof, line 1: Both of these Typst and PDF will obviously be deleted; they are just for my quick testing.

fake-small-caps-test.pdf
jassielof, line 1: Both of these Typst and PDF will obviously be deleted; they are just for my quick testing.

jassielof:
> Regarding petite caps, I think anything other than a separate function would make it very inconvenient.
> 
> Alternatives to `petitecaps()` would be `smallercaps()` or `tinycaps()`. But if petite caps is standard terminology I guess we should go with that.

Uhh, it's up to the team at the end, which way they'll think or find it better. I shared most of my thoughts in the relevant issue.

Petite caps is the usual way to call them. And I just suggested adding it as an extra option to small caps, as they are closely related (smaller and just a different font feature), to avoid duplication probably, but yeah.

laurmaedje:
As long as there are no conflicts, there is no need to merge main constantly. :)

jassielof:
Sob, I'm sorry 😓

laurmaedje:
No worries! Just thought I'd mention it so you can skip the unnecessary work of keeping it up to date all the time.

mewmew:
Just a minor typo (?) in the HTML example.

```diff
-    <link href="<https://fonts.googleapis.com/css2?family=IBM+Plex+Serif:wght@400;600&display=swap>" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=IBM+Plex+Serif:wght@400;600&display=swap" rel="stylesheet">
```

laurmaedje:
Hi! Thanks for your work on this. I'm sorry for the long radio silence. 

Looking at the implementation, this seems to make quite intricate changes which add significant complexity to the core shaping loop. I am somewhat concerned with the maintainability aspect of these changes. Could you motivate why these changes need to live there? I also wonder whether you explored any alternative designs. Naively, it feels like this could be handled at a bit higher level.

jassielof:
I initially tried to simply have workarounds like the Cuti package from Universe (uses RegEx), and so I thought maybe I could just put that into Rust, but it breaks the text flow, so I just found that the sub/superscripts are also in it and well added it there.

Also, once you find the approach suitable, would it be fine for me to also implement the Petite Caps feature? since it's related this, both the feature with fake support, either in this PR or a new one.

laurmaedje:
I'm still concerned with the complexity of the implementation. It's somewhat hard for me to judge how a cleaner implementation would play out without actually doing the work, but my gut feeling leans quite strongly towards this being too complex and a maintainability burden. As-is, I would thus tend to close this.

jassielof:
That's ok, I'll close it then, thanks for taking the time to review it.

Either way, about the petite caps, do you find those fine to add? Or not enough use case to add a dedicated function?

laurmaedje:
I wouldn't really want a new function (also `petitecaps` reads badly and `petite-caps` would be inconsistent). If anything, maybe it could be integrated into the smallcaps function. But then, there's even more: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-variant-caps. Not sure how to model this in terms of API design.

jassielof:
Yeah, I mentioned them in #7677, maybe a unified case or caps function? 🗣️
It would default to the most used one, being small caps, then with something enum-like, one could call the rest. 

Would be a breaking change in a way, but it would ease all the special caps or cases of course.

PR #7685, opened 2026-01-13
Use correct font constants for fractions without a rule

mkorje:
For fractions without a fraction rule, the layout is actually slightly different according to the spec[^1] (called a "stack"). In particular, different constants are used. This PR implements this.

Most fonts appear to make the constants for a "stack" such that its layout exactly matches a fraction but with the fraction rule invisible, hence no tests have needed updating.

Note that this construct is currently only visible to the user through the `math.binom` function (cf. #7686).

[^1]: See https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#stacktopshiftup-stacktopdisplaystyleshiftup-stackbottomshiftdown-stackbottomdisplaystyleshiftdown-stackgapmin-stackdisplaystylegapmin for a detailed explanation.

### DIFF
--- a/crates/typst-layout/src/math/fraction.rs
+++ b/crates/typst-layout/src/math/fraction.rs
@@ -17,55 +17,59 @@ pub fn layout_fraction(
     styles: StyleChain,
     props: &MathProperties,
 ) -> SourceResult<()> {
-    let constants = ctx.font().math();
-    let size = styles.resolve(TextElem::size);
-    let axis = constants.axis_height.at(size);
-    let thickness = constants.fraction_rule_thickness.at(size);
-    let math_size = props.size;
-    let shift_up = match math_size {
-        MathSize::Display => constants.fraction_numerator_display_style_shift_up,
-        _ => constants.fraction_numerator_shift_up,
-    }
-    .at(size);
-    let shift_down = match math_size {
-        MathSize::Display => constants.fraction_denominator_display_style_shift_down,
-        _ => constants.fraction_denominator_shift_down,
-    }
-    .at(size);
-    let num_min = match math_size {
-        MathSize::Display => constants.fraction_num_display_style_gap_min,
-        _ => constants.fraction_numerator_gap_min,
-    }
-    .at(size);
-    let denom_min = match math_size {
-        MathSize::Display => constants.fraction_denom_display_style_gap_min,
-        _ => constants.fraction_denominator_gap_min,
-    }
-    .at(size);
-
     let num = ctx.layout_into_fragment(&item.numerator, styles)?.into_frame();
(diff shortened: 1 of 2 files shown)


crates/typst-layout/src/math/fraction.rs
```diff
@@ -77,7 +81,41 @@ pub fn layout_fraction(
...
+            _ => constants.stack_bottom_shift_down,
+        }
+        .at(size);
+        let gap_min = match math_size {
+            MathSize::Display => constants.stack_display_style_gap_min,
+            _ => constants.stack_gap_min,
+        }
+        .at(size);
+
+        let gap = (shift_up - num.descent()) + (shift_down - denom.ascent());
```
laurmaedje, line 102: here, the gap does not yet factor in `gap_min`, but above `num_gap` and `denom_gap` do. This seems slightly inconsistent.

crates/typst-layout/src/math/fraction.rs
```diff
@@ -77,7 +81,41 @@ pub fn layout_fraction(
...
+
+        let width = num.width().max(denom.width()) + 2.0 * item.padding.at(size);
+        let height = num.height() + gap.max(gap_min) + denom.height();
+        let size = Size::new(width, height);
+
+        let num_pos = Point::with_x((width - num.width()) / 2.0);
+        let denom_pos =
+            Point::new((width - denom.width()) / 2.0, height - denom.height());
+
+        let baseline = num.ascent() + shift_up + (gap_min - gap).max(Abs::zero()) / 2.0;
```
laurmaedje, line 112: All the other computations seem very parallel to the previous code just simplified, but the baseline computation seems very different. I don't fully understand it. Why `(gap_min - gap).max(Abs::zero()) / 2.0`?

crates/typst-layout/src/math/fraction.rs
mkorje: It's because above there is a separate gap for both `num` and `denom`, but here there is only a single gap

mkorje: And for the above code everything is around the axis-height, where the fraction rule sits. Here that isn't the case anymore.

crates/typst-layout/src/math/fraction.rs
mkorje: If the gap is too small, then we need to use `gap_min` instead. To do that, the space needing to be added (`gap_min - gap`) has to be evenly distributed both above and below. All that really means is adjusting the baseline (since the position of both the num and denom are relative to the edges of the frame). `num.ascent() + shift_up` gets us to the baseline _only if_ the gap wasn't too small. If it was too small, then the baseline is actually a further `(gap_min - gap) / 2.0` away.

laurmaedje:
Thanks!

PR #7687, opened 2026-01-13
Fix clippy warning

saecki:


laurmaedje:
Is this a Rust 1.92 warning? Generally speaking, I'd prefer to only address warnings alongside Rust bumps as otherwise it's not enforced anyway and just adds more PR churn.

saecki:
> Is this a Rust 1.92 warning?

Yeah, I think so. 

Do you set a rustup override for the typst directory to use the same rust version as in the CI? Otherwise I find it somewhat annoying to mentally filter out the clippy warnings that aren't relevant.

laurmaedje:
> Do you set a rustup override for the typst directory to use the same rust version as in the CI?

Since I only really work on Typst (I don't have energy to code otherwise), I'm just always globally on the version that's also in Typst.

> Otherwise I find it somewhat annoying to mentally filter out the clippy warnings that aren't relevant.

Yeah that makes sense. I think locally overriding (if that's easily possible?) is the cleanest solution though since fundamentally the project will sometimes lag behind the very latest release.

saecki:
`rustup override set <toolchain>` is the command, for anyone reading this :)

PR #7691, opened 2026-01-13
Accept `Duration` type timezone offset values

### LINKED ISSUE #6277: Support half-hour timezones
### Description

Hi!

It would be really call if typst's `datetime.today()` supported my Island's timezone, which is not an intedger 

### Use Case

I live in the wonderful Island of [Newfoundland](https://en.wikipedia.org/wiki/Newfoundland_(island)), which follows the [Newfoundland Time Zone](https://en.wikipedia.org/wiki/Newfoundland_Time_Zone), it would be wonderful if we could have non-intedger timezone support in typst, as occasionally the current day (i.e. `datetime.today()`) is wrong due to us having a -3.5 hour offset timezone (except for DST, which is 2.5 hours)

t-rapp:
Support timezone offset values in `#datetime.today()` which are not full hours. For example this applies to the Newfoundland Time Zone (UTC-03:30 / UTC-02:30), or Nepal Standard Time (UTC+05:45).

Fixes #6277.

### DIFF
--- a/crates/typst-library/src/foundations/datetime.rs
+++ b/crates/typst-library/src/foundations/datetime.rs
@@ -629,5 +629,5 @@ cast! {
     TodayOffset,
     self => self.0.into_value(),
     v: Duration => Self(v),
-    hours: i64 => Self(Duration::construct(0, 0, hours, 0, 0)),
+    hours: i64 => Self(time::Duration::hours(hours).into()),
 }


tests/suite/foundations/datetime.typ
```diff
@@ -65,6 +65,11 @@
 #test(datetime.today().display(), "1970-01-01")
 #test(datetime.today(offset: auto).display(), "1970-01-01")
 #test(datetime.today(offset: 2).display(), "1970-01-01")
+#test(datetime.today(offset: 24).display(), "1970-01-02")
+#test(datetime.today(offset: -24).display(), "1969-12-31")
+#test(datetime.today(offset: duration(hours: 5, minutes: 45)).display(), "1970-01-01")
+#test(datetime.today(offset: duration(days: 2)).display(), "1970-01-03")
+#test(datetime.today(offset: duration(days: -2)).display(), "1969-12-30")
```
laurmaedje, line 72: These tests are a tad misleading because the actual CLI function using `FixedOffset` does not allow a duration `>= 24h`. Perhaps, we should rather base the TestWorld's `today` output on a time different than midnight such that an offset < 24h can have an effect in both directions. 

I also noticed https://github.com/typst/typst/issues/7843 while playing with this. I'm good with merging the fix as-is for now, but aligning the CLI and test impls a bit more could be a good followup.

laurmaedje:
Thanks!

tests/suite/foundations/datetime.typ
t-rapp: Thanks for bringing the limitation with `FixedOffset` to my attention, I somehow missed that.

I assume aligning CLI and test `World` impls would mean the offset calculation is also done with `chrono` types in the tests?

laurmaedje: That would be an option. Alternatively, we could impose the limitation separately in the tests though at that point we're not testing all that much I suppose.

I could also imagine moving the chrono based today impl to typst-kit to share it, but that would be a slightly larger task and might involve some API design bikeshedding. And it would need to support both current and fixed time of course.

While looking at that, I just noticed that we're depending on Local timezone even when SOURCE_DATE_EPOCH is set, which I guess is not great for reproducibility.

But anyways, I'm going a bit off-topic. There are more things to improve here, but moving the tests to a reproducible chrono based impl would be a reasonable first step I think.

PR #7700, opened 2026-01-14
Fix prime and top attachment merging

### LINKED ISSUE #7647: Prime and top attachments can swap order when merged
### Description

The below output has the prime on the inside despite being added as the outer attachment.

```typ
#set page(width: 1in, height: auto)
#let base = $attach(a, t: b)$
$ attach(base, tr: ') $
```

Image

The issue is that when we merge the `t` and `tr` fields, we always add them as `tr + t`, cementing the order. The fix would be to not merge a primed `tr` into the base when the base has a `t`, but it's not trivial to _also_ guarantee that we are typesetting `t` without limits, as that depends on the laid out base, which already had its fields merged.

cc @mkorje

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

isuffix:
Closes #7647

This effectively inlines the recursion that was previously happening stage-by-stage through realization and back.

I made a previous PR at https://github.com/mkorje/typst/pull/1, but we decided to wait for the `math-ir` branch to merge.

I am not applying @mkorje's suggestion from the issue.

### DIFF
--- a/tests/ref/pdf/hashes.txt
+++ b/tests/ref/pdf/hashes.txt
@@ -990,6 +990,7 @@ e8d812e8211e884238df314992db0d2d math-attach-kerning-mixed
 1faea4ab2fa0ed56bd56393e07ee2043 math-attach-mixed
 5c675003241e27c125ba12cde2bbfda5 math-attach-nested
 daa8ce444236a701cdf9bf5f74cb08ec math-attach-nested-base
+1ed090bfe9a0a4606d4bf08338ee6184 math-attach-nested-base-bad-placement
 546f2fdc7c346dba07a85cb02594a5a4 math-attach-nested-deep-base
 79cd3f10cb464b2bd5583a8a72607d8e math-attach-postscripts
 e120f62f76c1a20a4f33805cfc842753 math-attach-prescripts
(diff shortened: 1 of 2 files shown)


isuffix:
Update: I found an issue with the case below, should be a quick fix.
```typ
$
  attach(attach(a, b: 1, tr: '), t: 2)
  quad
  attach(attach(attach(a, b: 1), tr: '), t: 2)
$
```

isuffix:
@mkorje
While fxing this issue, I looked at the rendering more and really started to dislike the look of this implementation. Joining the primes to the top attachment as content only looks good when the prime comes first, i.e. `tr + t`. The other ordering of `t + tr` just looks bad/incorrect.

Ex: (first is `t + tr`)


So I think I do agree with [your earlier comment](https://github.com/typst/typst/issues/7647#issuecomment-3706903510) that we should avoid merging any `tr` primes if the inner attach has a `t`. However it does still look bad if the base had limits. Luckily, the new implementation already makes it easy to check if the base had limits!

I'm now having the function optimistically merge the `tr` element inward before the recursive call, but potentially put it back in the outer attachments after we layout the inner base and we know the base didn't have limits.

isuffix:
I found (and fixed) another issue with attachment merging (hopefully the last one). But for this, I've just added it as two additional commits.

We have quite bad placement when using nested attachments like so:
```typ
--- math-attach-nested-base-bad-placement paged ---
// This case has bad placement with nested attachments.
$
  // `b:2` in the middle attach stops `b:4` from moving inward.
  attach(attach(attach(a, t: 1), b: 2), t: 3, b: 4)
  quad
  attach(attach(b, t: 1, b: 2), t: 3, b: 4)
$
```


I assume this wasn't found earlier because its the left-associative form of attachment which is normally right-associative, but I think its still worth improving.

To solve it, we need to consider not just adjacent attachments, but the list of all attachments from the outside in. To do so, I changed from passing a single set of attachments at each recursive call with a stack-allocated linked list pointing back up the call chain to the top. A [stack-allocated linked list](https://rust-unofficial.github.io/too-many-lists/infinity-stack-allocated.html#the-stack-allocated-linked-list) does require some complication in Rust, but I've addressed that in the comments and I think its just nice for this case to avoid extra allocations.

laurmaedje:
Thanks!

PR #7701, opened 2026-01-14
Update Latvian translations

kristoferssolo:
Added Latvian translations to the missing fields

### DIFF
--- a/crates/typst-library/translations/lv.txt
+++ b/crates/typst-library/translations/lv.txt
@@ -6,6 +6,6 @@ heading = Sadaļa
 outline = Saturs
 raw = Saraksts
 page = lpp.
-footnote = Kājene
+footnote = Vēre
 email = E-pasts
 telephone = Tālrunis


crates/typst-library/translations/lv.txt
```diff
@@ -6,6 +6,6 @@ heading = Sadaļa
 outline = Saturs
 raw = Saraksts
 page = lpp.
-# footnote =
-# email =
-# telephone =
+footnote = Kājene
```
laurmaedje: Are you sure this is the correct term? According to online translation, this is "Footer" rather than "Footnote". Hard for me to judge.

kristoferssolo: My bad. It indeed translates to "footer". I've checked 2 major terminology sources ([letonika](https://www.letonika.lv/groups/default.aspx?r=1107&q=piez%C4%ABme&id=2071502&g=2) and [termini.go.lv](https://termini.gov.lv/atrast/footnote)) and it translates as "Vēre".

laurmaedje:
Thanks!

PR #7702, opened 2026-01-15
Allow specifying test stage on diagnostic note

saecki:
This is enforced for diagnostics that are only generated in a specific output/target that isn't hit by all possible branches of the stage tree:
```txt
        ╭─> render
 paged ─┼─> pdf ───> pdftags
        ╰─> svg
 html  ───> html
```
An example would be a `PDF/UA-1` error, that is only emitted in the `pdf` stage, but the test has a `paged` attribute, and will also run the `render` and `svg` stages.
The error annotation must specify the specific stage like this:
```
// Error: [pdf] 3:6-5:6 PDF/UA-1 error: PDF artifacts may not contain links
```

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -99,11 +99,13 @@ There are, broadly speaking, three kinds of tests:
   target.
 
 - Tests that ensure the code emits particular diagnostic messages: Those have
-  inline annotations like `// Error: 2-7 thing was wrong`. An annotation can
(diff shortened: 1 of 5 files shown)


saecki:
During the implementation I realized that we only really need to change 2 tests, which could also just be annotated with `pdf` instead of `paged`.

PR #7709, opened 2026-01-16
Ensure errors are emitted in all test stage branches

saecki:
Supersedes #7702.

This ensures that errors are emitted in all branches of the stage tree:
```txt
        ╭─> render
 paged ─┼─> pdf ───> pdftags
        ╰─> svg
 html  ───> html
```
Since currently only the `paged`, `html`, and `pdf` stages can emit errors, this should only happen in a few combinations:
- A test is annotated with `html` and `paged`, but only produces an error in one of the two targets.
- A test is annotated with `paged`, but only produces an error in the `pdf` output.

This was only an issue for two tests in the whole test suite. Which arguably should have been `pdf` tests to begin with, but the `pdf` attribute wasn't available.

With these changes, the test suite is guaranteed to pass with any combination of the `--stages` flag.


tests/src/run.rs
```diff
@@ -148,35 +146,37 @@ impl<'a> Runner<'a> {
...
                 continue;
             }
+
+            log!(
+                self,
+                "diagnostic is only emitted in [{seen}] but the test is run for [{}]",
+                self.test.attrs.stages
+            );
+
+            log!(self, "consider narrowing the test to these attributes: [{seen}]");
```
laurmaedje: I think printing this below the error with 4 space indent would be clearer as it's, at least to me, conceptually hierarchically lower than the error itself. it belongs to it.

tests/src/collect.rs
```diff
@@ -107,6 +107,14 @@ pub trait TestStage: Into<TestStages> + Display + Copy {}
 bitflags! {
     /// The stages a test in ran through. This combines both compilation targets
     /// and output formats.
+    ///
+    /// Here visual representation of the stage tree:
+    /// ```md
```
laurmaedje:
```suggestion
    /// Here's a visual representation of the stage tree:
    /// ```
```
is this really markdown?

saecki: This should probably be `txt`, otherwise it's assumed to be a doc test.

laurmaedje: ah

tests/src/run.rs
saecki: I didn't really like the way any variation of this looked. I just decided to make the error message a bit more concise an fit everything into one line.

Issue #7582, opened 2025-12-14
typst-kit: Allow adding custom fonts from bytes, similar to default embedded fonts


ThiefMaster:
When using typst embedded in custom code using the Rust API, it's already trivial to provide the main file from some non-filesystem source (e.g. `include_str!()`). This is great because it allows to produce a single binary containing everything needed w/o any auxiliary files.

However, doing the same with fonts is extremely cumbersome and requires extremely ugly code due to the use of private fields:

<details>

```rust
static FONT_DATA: OnceLock<(
    LazyHash<FontBook>,
    Arc<Vec<FontSlot>>,
    usize,
    Arc<Vec<Font>>,
)> = OnceLock::new();

impl TypstWrapperWorld {
    fn new(source: &str, inputs: Dict) -> Self {
        let (book, fonts, custom_font_offset, custom_fonts) = FONT_DATA.get_or_init(|| {
            let mut fonts = FontSearcher::new().include_system_fonts(false).search();
            let mut offset = 0;
            loop {
                if fonts.book.info(offset).is_none() {
                    break;
                }
                offset += 1;
            }
            let mut custom_fonts = Vec::new();
            for font in Font::iter(Bytes::new(include_bytes!(
                "../typst/LiberationSans-Regular.ttf"
            ))) {
                fonts.book.push(font.info().clone());
                custom_fonts.push(font);
            }
            (
                LazyHash::new(fonts.book),
                Arc::new(fonts.fonts),
                offset,
                Arc::new(custom_fonts),
            )
        });
        Self {
            source: Source::detached(source),
            library: LazyHash::new(Library::builder().with_inputs(inputs).build()),
            book: book.clone(),
            fonts: Arc::clone(fonts),
            custom_font_offset: *custom_font_offset,
            custom_fonts: Arc::clone(custom_fonts),
        }
    }
}

impl typst::World for TypstWrapperWorld {
    fn font(&self, id: usize) -> Option<Font> {
        if id >= self.custom_font_offset {
            self.custom_fonts.get(id - self.custom_font_offset).cloned()
        } else {
            self.fonts[id].get()
        }
    }

    // ...
}

```

</details>

It would be nice if the `FontSearcher` (or probably more likely the `Fonts` object it produces) had an `add_from_bytes` function which adds the provided font (coming e.g. from `include_bytes!()`) to the collection.

I wouldn't mind sending a PR implementing this if you're open to this functionality.

laurmaedje:
[...]

Issue #6044, opened 2025-03-11
Add implementation of SystemWorld to typst-kit


yalap13:
### Description

Adding an implementation of `SystemWorld` (as the one in typst-cli) in typst-kit would be useful for using typst in other projects. I am aware this has been mentioned previously (ie [this comment](https://github.com/typst/typst/issues/4653#issuecomment-2279889262)) but I have not found any updates afterwards. Is this still planned? 

### Use Case

Using typst in other projects as a library with a clean implementation of the `World` trait.

laurmaedje:
Not sure whether the whole `SystemWorld` makes sense (it might), but it could be modularized and parts could be in typst-kit.

yalap13:
It might not be necessary to have the complete `SystemWorld` indeed but if there were more utilities (kind of like `FontSearcher`) to make the implementation of the `World` trait easier, it would be very helpful.

PR #7710, opened 2026-01-16
Rework and expand `typst-kit`

laurmaedje:
Typst Kit provides reusable building blocks to downstream Typst integrations. This PR improves the API of existing functionality and adds a bunch of additional functionality. The original motivation for this was reusing some stuff from the CLI in the upcoming Typst-based docs.

### Overview
I have attempted to design APIs that are as general and flexible as possible, supporting varied use cases. I'll give a brief overview over the changes and additions in each subsystem below. I've also remodelled the feature flags for more flexibility and naming consistency.

#### Fonts
Font loading is now a bit more generic: The new `FontStore` can still load embedded, path, and system fonts, but things can be composed somewhat more freely than before, and a font can also be loaded generically through a trait.

#### Files
The largest (and probably most useful) additions is the new `FileStore`, which is backed by a `FileLoader`. It takes care of creating, caching, and manages Typst source files, while delegating the raw loading to an underlying loader. From what I've seen, getting this right is (understandibly) the most difficult part about building a `World`. The new implementation is also a bit more efficient and correct than the old CLI's. When a file is used both as a source file and raw bytes, it's only read once now and the backing storage is reused as much as possible.

#### Packages
The previous `PackageStorage` is now broken up into three package providers: `FsPackages`, `UniversePackages`, and `SystemPackages` (which combines the two). Instead of being coupled to a fixed downloader implementation, `UniversePackages` depends on a generic `impl Downloader`.

#### Diagnostics
`typst-kit` now has a `diagnostics::emit` function, which emits formatted CLI-style diagnostics to a colorized writable. It's backed by the new `DiagnosticsWorld` (an extension to the `World` which can resolve names of files). This is similar in spirit to `typst_ide::IdeWorld` which adds extra requirements to the `World` for usage with IDE functionality.

#### Downloader
As already mentioned, the concrete provided downloader (based on `ureq`) is now decoupled from abstract downloading. I've also decoupled downloading itself from progress reporting by adding a new `ProgressDownloader` that wraps a generic downloader. Now, individual download calls do not decide whether progress reporting is desired. Instead, they provide a dynamic key, which the progress downloader can use to decide whether it wants to emit something (e.g. printing to the console). I've also fixed a bug with the download status printing, which would swallow the initial "downloading ..." line on the first update and show the status line twice when finishing.

#### Watcher
The file watching implementation of the CLI is now exposed in `typst-kit`.

#### Server
The live-reloading HTTP server implementation of the CLI is now also exposed in `typst-kit`.

### Breakage

Given that the PR reworks the full API surface, it's obviously breaking. Beyond that, I've changed the default features from all to none. It is actually not that common to need all and selecting exactly what one needs is encouraged by this new default.

Beyond `typst-kit`, in the process of adjusting the feature flags, I've also renamed the CLI's `embed-fonts` feature flag to `embedded-fonts`.

### Checking the API surface

For the best experience when browsing the new API's docs, I recommend the `docsrs` cfg because it gives nice docs for the future flag:
```bash
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc -p typst-kit --no-deps
```

### Why now?
I fell down a rabbit hole while working on the Docs-in-Typst. While it took a couple of days, I am quite happy with the new APIs and in particular the `FileStore` should make it considerably easier to build a `World`.

### Issues
- Closes https://github.com/typst/typst/issues/7582
- Given that the `SystemWorld` is pretty thin now, I would say it's also fair to say that this closes https://github.com/typst/typst/issues/6044


laurmaedje:
cc @tingerrr who originally introduced `typst-kit`, in case you want to take a look :)

crates/typst-kit/Cargo.toml
```diff
@@ -35,27 +42,42 @@ ureq = { workspace = true, optional = true }
...
-embed-fonts = ["fonts", "dep:typst-assets", "typst-assets/fonts"]
+# Enables `packages::SystemPackages`.
+system-packages = ["dep:dirs", "universe-packages"]
 
-# Whether to vendor OpenSSL. Not applicable to Windows and macOS builds.
+# Enables `packages::UniversePackages`.
+universe-packages = ["dep:flate2", "dep:tar", "dep:fastrand"]
+
+# Enables `diagnostics::emit`.
+emit-diagnostics = ["dep:codespan-reporting"]
```
tingerrr, line 65: Perhaps we should make it clear that these are _terminal_ diagnostics.

laurmaedje: Technically, it's not bound to a terminal as a writable color stream suffices, but in practice yeah. 

Right now though, the feature flags use the exact same words as the struct/fn item paths (with duplicate words like "packages" removed and reverse order, as it sounds more natural), which I think is quite nice. How would you name this? Or do you just mean that the docs should clarify this?

laurmaedje: I've kept the name for consistency, but added some additional explanation to each feature flag.

PR #7712, opened 2026-01-16
Use libm to ensure cross-platform deterministic results

saecki:
This ensures floating point math is deterministic across platforms, by adding lints to forbid non-deterministic functions inside the typst workspace. Dependencies can still use them, and that's why `typst-render` still won't be deterministic in the foreseeable future. But floating point computations with `Value`s, `typst-layout`, `typst-svg`, and `typst-pdf` *should* be.

### DIFF
--- a/crates/typst-utils/src/scalar.rs
+++ b/crates/typst-utils/src/scalar.rs
@@ -43,6 +43,9 @@ impl Scalar {
 
     /// Raises a number to an integer power.
     pub fn powi(self, mut b: i32) -> Self {
+        // Ported from https://github.com/llvm/llvm-project/blob/0ee439b/compiler-rt/lib/builtins/powidf2.c
+        // Copyright: The LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+        // See NOTICE for full attribution.
         let mut a = self.get();
         let recip = b < 0;
         let mut r = 1.0;
--- a/NOTICE
+++ b/NOTICE
@@ -105,3 +105,289 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ================================================================================
+
+
+================================================================================
+`Scalar::powi` implementation in `crates/typst-utils/src/scalar.rs` is ported
+from here:
+https://github.com/llvm/llvm-project/blob/0ee439b/compiler-rt/lib/builtins/powidf2.c
+
+------------------------------------------------------------------------------
+The LLVM Project is under the Apache License v2.0 with LLVM Exceptions:
+------------------------------------------------------------------------------
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+    1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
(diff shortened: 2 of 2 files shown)


Enivex:
Has there been any further discussions regarding moving to fixed point?

saecki:
We quickly touched the topic, but didn't really get into further discussion

crates/typst-library/src/layout/point.rs
```diff
@@ -57,7 +57,11 @@ impl Point {
 
     /// The distance between this point and the origin.
     pub fn hypot(self) -> Abs {
-        Abs::raw(self.x.to_raw().hypot(self.y.to_raw()))
+        // The `sqrt` function is defined by IEEE-754 and thus deterministic.
+        // In addition this should be faster than `f64::hypot` or `libm::hypot`.
```
laurmaedje, line 61: why would std and libm implement something suboptimal? or do we provide different guarantees?

crates/typst-library/src/visualize/stroke.rs
```diff
@@ -62,7 +62,7 @@ pub struct Stroke<T: Numeric = Length> {
     /// The stroke's line dash pattern.
     pub dash: Smart<Option<DashPattern<T>>>,
     /// The miter limit.
-    pub miter_limit: Smart<Scalar>,
+    pub miter_limit: Smart<Ratio>,
```
laurmaedje, line 65: I never knew what this is supposed to represent. 

If it's conceptually actually a ratio, we should probably expose it like that in the future, either in addition or with a deprecation period for the float? (separately from this PR)

crates/typst-utils/src/scalar.rs
```diff
@@ -35,6 +35,34 @@ impl Scalar {
...
     }
+
+    /// Returns the square root of this scalar.
+    pub fn sqrt(self) -> Self {
+        Self::new(self.get().sqrt())
+    }
+
+    /// Raises a number to an integer power.
+    pub fn powi(self, mut b: i32) -> Self {
+        let mut a = self.get();
```
laurmaedje, line 49: where does this algorithm come from? a link to something would be good because it's very dense code.

probably it's a well known algorithm, but if not, does it need some sort of link or NOTICE entry?

crates/typst-library/src/layout/point.rs
saecki: I think they provide additional guarantees for very large/small numbers, but since this is only used for layouting, where very large numbers shouldn't really appear, and very small numbers will be rounded to `0`, this should be fine.
As for the performance: `libm` can't guarantee the presence of an `FSQRT` instruction, so I assume it implements something that would be faster than a dot product followed by `libm::sqrt`. The rust standard library also delegates to `cmath` (the platform specific `libm`), I'm guessing because of the additional guarantees.

crates/typst-library/src/visualize/stroke.rs
saecki: Yeah, we probably should: https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/stroke-miterlimit#:~:text=ratio

crates/typst-utils/src/scalar.rs
saecki: Oh, good catch! I made a mental note to check how to attribute this code, but completely forgot. I got it from the llvm project: https://github.com/llvm/llvm-project/blob/0ee439b/compiler-rt/lib/builtins/powidf2.c#L17

Should I just copy the copyright notice above the code inside the function. Or should it be added to the `NOTICE`

bastienvoirin: See also https://en.wikipedia.org/wiki/Exponentiation_by_squaring#With_constant_auxiliary_memory, but there aren't many references listed there.

laurmaedje: See https://github.com/typst/typst/blob/9a4316c6b112e7ed1b0ab85c2607753a6fe04481/crates/typst-syntax/src/path.rs#L314-L316 and https://github.com/typst/typst/blob/9a4316c6b112e7ed1b0ab85c2607753a6fe04481/NOTICE#L14-L17 for how I've most recently done this.

Not all attributions are done exactly like this, but this is how I want to do it going forward for license that are compatible with this (link only in NOTICE and with full revision, still short attribution text in source with reference to NOTICE). You'll need to add a new NOTICE section for Apache-2 with LLVM exception.

saecki: Ok, I've added the attribution

PR #7716, opened 2026-01-18
Add support for discovering Adobe's Creative Cloud fonts as system fonts

### LINKED ISSUE #5336: Support fonts activated through Adobe Fonts
### Description

In the Creative Cloud App, it says "You can now use Adobe Fonts in other apps on your device" after installing a font, so it should be fine.

Feature request received via email by a person that achieved this via symlinks in 0.11.1, but apparently this no longer works in 0.12.

### Use Case

This is useful for people that are subscribed to Adobe Fonts.

jassielof:
# Adobe fonts discovery support as part of *system* fonts

This is supposed to be only supported and affecting Windows and MacOS systems, as those are the platforms [Adobe Creative Cloud officially supports](https://helpx.adobe.com/download-install/apps/system-requirements/creative-cloud-requirements.html).

The fonts, at least on Windows, are stored as IDs, such as `16179`, with the hidden attribute, and without any file extension.

Typst uses the fonts DB crate, so `load_fonts_dir()` won't work as *I think*, as it expects the directory to have the fonts with its respective extensions, so it's needed to read each file, making the assumption that they are valid font files *with the naming convention as mentioned before*.

Fixes #5336.

## Paths

Considering both platform "data directories" are `~/Library/Application Support/` for MacOS and `%APPDATA%` for Windows, as resolved by `dirs::data_dir()`.

Adobe related font directories are (appened to the data directory):

- Core Sync: `Adobe/CoreSync/plugins/livetype/`
- Root: `Adobe/`

### Relevant subdirectories

> MacOS *relevant* directories are prefixed with a dot.
> Windows font files have no extensions, and are hidden via file attributes.

For Core Sync, the *relevant* subdirectories are:

- `r/`: Installed fonts
- `w/`: Available fonts for installation, these need to be installed from Creative Cloud first, it basically moves them to the `r/` directory.
- The rest of subdirectories seem to be irreleavnt.

For the *Root* directory, the *relevant* subdirectory is:

- `User Owned Fonts`: Uploaded fonts by the user

> Information about the fonts was mainly gathered from the relevant [issue](https://github.com/typst/typst/issues/5336) by <lime@github.com>.
> What each subdirectory exactly contains and is used for isn't publicly documented by Adobe.

## Licensing and permissions

As the licensing information is explained in for example [Adobe Garamond](https://fonts.adobe.com/fonts/adobe-garamond):

> **The full Adobe Fonts library is cleared for both personal and commercial use.**
> As with everything from Adobe Fonts, you can use these fonts for:
>
> - Design Projects
> - Website Publishing
> - **PDFs**: *Embed fonts in PDFs for viewing and printing*
> ...

Also explained in <https://helpx.adobe.com/fonts/using/font-licensing.html>:

> **Can I embed the fonts in a PDF file or ebook? Can they be used in printed books?**
> Yes, Adobe Fonts are licensed for embedding in ebook formats that protect font data, such as EPUB, iBooks, Kindle (MOBI), Adobe’s Digital Publishing Suite (DPS), and PDFs.
> ...

This means that using these fonts in Typst documents is allowed, as Typst doesn't distribute the fonts, nor fully embeds them, it subsets them.

### Implementation

The respective font discovery implementation is located in `./crates/typst-kit/src/fonts.rs` with methods `load_adobe_fonts()` and `FontSearcher::search_with()`.

By default, Typst will only read the installed (`r/`) and user owned fonts (`User Owned Fonts/`).

**NOTE: I have tested this only on my machine, Windows, it would be nice if some MacOS user that also has Creative Cloud could test this.**

### DIFF
--- a/crates/typst-kit/src/fonts.rs
+++ b/crates/typst-kit/src/fonts.rs
@@ -152,23 +152,7 @@ pub fn system() -> impl Iterator<Item = (FontPath, FontInfo)> {
 
         // Add Adobe Fonts on Windows and macOS.
         #[cfg(any(target_os = "windows", target_os = "macos"))]
-        if let Some(data_dir) = dirs::data_dir() {
-            let adobe_base = data_dir.join("Adobe");
-
-            #[cfg(target_os = "macos")]
-            let dot = ".";
-            #[cfg(target_os = "windows")]
-            let dot = "";
-
-            let subdirs = [
-                format!("CoreSync/plugins/livetype/{dot}r"),
-                format!("{dot}User Owned Fonts"),
-            ];
-
-            for subdir in subdirs {
-                load_adobe_fonts(db, &adobe_base.join(subdir));
-            }
-        }
+        load_adobe_fonts(db);
     })
 }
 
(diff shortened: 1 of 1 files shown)


reknih:
Can confirm it works on macOS.

Proxima Nova picked up by typst fonts
Creative Cloud Desktop app with Proxima Nova in Adobe Fonts

reknih:
The relevant document for licensing are the [terms of use](https://wwwimages2.adobe.com/content/dam/cc/en/legal/servicetou/Adobe-Fonts-Product-Specific-Terms-en_US-20241007.pdf). Clause 3.1 (A) covers our use case:

> You may use Desktop Fonts to design and develop Documents and you may embed copies of the Desktop Fonts into your Document for the purpose of printing and viewing the document. The font must be subset to include only the glyphs necessary for displaying the work, and the Document must obfuscate or protect its embedded font data from deliberate or inadvertent discovery or misuse. No other embedding rights are implied or permitted under this license.

This means that for example something like using it as a web font or embedding the full font file would not be allowed, but currently, if Typst embeds a font, it also subsets it, so we are good to go

laurmaedje:
Thanks!

PR #7718, opened 2026-01-18
Fix grammar in README

Howard-aile:


README.md
```diff
@@ -239,7 +239,7 @@ instant preview. To achieve these goals, we follow three core design principles:
   (`\usepackage{knob}`).
 
 - **Performance through Incrementality:**
-  All Typst language features must accommodate for incremental compilation.
+  All Typst language features must accommodate incremental compilation.
```
cady-b: I don't think this is necessarily an _error_: https://ell.stackexchange.com/questions/232081/accommodate-vs-accommodate-for

Personally, I definitely prefer the version with "for"; it'd be good to know what others think of this.

Howard-aile: lol, i agree with u now.

laurmaedje:
Thanks!

PR #7720, opened 2026-01-18
Add wrap-float support for text wrapping around placed elements

boergens:
## Summary

This PR adds support for text wrapping around floating placed elements, similar to LaTeX's `wrapfig` package. This is a commonly requested feature that enables more sophisticated document layouts where text flows around figures, images, or other content.

## Usage

The feature adds a `wrap: true` parameter to the `place` function:

```typst
#place(
  top + right,
  float: true,
  wrap: true,
  clearance: 10pt,
  box(width: 35%)[
    #image("figure.png", width: 100%)
    #text(size: 9pt)[Caption]
  ]
)

#lorem(100) // Text wraps around the placed element
```

## Key Features

- **Flexible positioning**: Supports `top + left`, `top + right`, `bottom + left`, `bottom + right` alignments
- **RTL support**: Respects text direction for logical alignment (start/end)
- **Column support**: Works with both column-scoped and parent-scoped floats
- **Clearance control**: The existing `clearance` parameter controls spacing between the float and wrapped text
- **Justified text**: Properly handles justified text with variable line widths
- **Page breaks**: Floats that don't fit are correctly deferred to the next page

## Technical Approach

The implementation uses a three-phase approach:

1. **Collection & Positioning**: Wrap-floats are collected during flow layout and positioned at their target location
2. **Exclusion Zones**: A `ParExclusions` structure tracks regions where text cannot flow
3. **Variable-Width Line Breaking**: The Knuth-Plass algorithm was extended to support per-line width constraints, enabling text to wrap around exclusion zones

The layout uses iterative refinement to handle the circular dependency between paragraph height and float positioning, converging in 1-2 iterations for typical documents.

## Warnings

The implementation includes helpful warnings for edge cases:
- Float wider than 2/3 of the available width
- Text gap narrower than 1/6 of the available width  
- Text overflow when content doesn't fit in the narrow gap

## Test Coverage

47 new test cases covering:
- Basic left/right/top/bottom positioning
- Multiple floats on the same page
- RTL text direction
- Multi-column layouts
- Page break behavior
- Interaction with footnotes and citations
- Edge cases (zero-height floats, very narrow gaps, etc.)

## Files Changed

Key changes:
- `crates/typst-library/src/layout/place.rs`: Added `wrap` parameter
- `crates/typst-library/src/layout/regions.rs`: Added `WrapFloat` and `ParExclusions` types
- `crates/typst-layout/src/flow/distribute.rs`: Wrap-float positioning and exclusion computation
- `crates/typst-layout/src/inline/linebreak.rs`: Variable-width Knuth-Plass implementation
- `crates/typst-layout/src/inline/mod.rs`: Measure/commit split for paragraph layout with exclusions

### DIFF
--- a/crates/typst-layout/src/flow/distribute.rs
+++ b/crates/typst-layout/src/flow/distribute.rs
@@ -105,16 +105,17 @@ impl WrapState {
     ///
     /// Returns `None` if there are no exclusions overlapping the paragraph,
     /// which is an optimization to avoid unnecessary work in the common case.
-    fn exclusions_for(&self, par_y: Abs, par_height_estimate: Abs) -> Option<ParExclusions> {
+    fn exclusions_for(
+        &self,
+        par_y: Abs,
+        par_height_estimate: Abs,
+    ) -> Option<ParExclusions> {
         if self.floats.is_empty() {
             return None;
         }
-        let excl = ParExclusions::from_wrap_floats(par_y, par_height_estimate, &self.floats);
-        if excl.is_empty() {
-            None
-        } else {
-            Some(excl)
-        }
+        let excl =
+            ParExclusions::from_wrap_floats(par_y, par_height_estimate, &self.floats);
+        if excl.is_empty() { None } else { Some(excl) }
     }
 
     /// Find the bottom of existing floats on the given side.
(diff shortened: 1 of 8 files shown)


laurmaedje:
I'm not sure what exactly this test case is trying to test:



Moreover, if you can't be bothered to delete an inapplicable `CLAUDE.md` file with wrong instructions about Typst's issue tracker setup, I see no reason to spend any time looking into the remainder of the changes.

I should also note that, even if the PR were human-built, a topic such as this is way too large for a first contribution and would need a tight feedback loop during development.

Issue #7653, opened 2026-01-05
`first-line-indent` triggered after column break


Enivex:
### Description

```
#set page("a8", flipped: true,columns: 2)
#set par(first-line-indent: (amount:1em,all: false))

#lorem(6)

#lorem(6)

#colbreak()

#lorem(6)

#lorem(6)

```
results in

<img width="773" height="579" alt="Image" src="https://github.com/user-attachments/assets/d19b7320-31c7-4bc7-bd18-66ac7396fa67" />

As you can see, the first line in the first paragraph of the second column is indented. This is unlike the behavior for a page break, where the first line in the first paragraph of the second page is _not_ indented

<img width="393" height="541" alt="Image" src="https://github.com/user-attachments/assets/097d1b1d-b75c-4739-8a8d-78ba1221fefc" />

### Typst version

- [ x] I am using the latest version of Typst

laurmaedje:
The big question is whether this should only apply to explicit or also incidental column breaks. For pages, it currently only applies to actual `pagebreak`, not when a paragraph happens to be first on the page. See e.g. here:
```typ
#set par(first-line-indent: 2em)
#set page(height: 31cm)

#(15 * (lorem(85) + parbreak()))
```

https://github.com/typst/typst/pull/7722 is an appropriate fix if we only want this for explicit breaks. If we also want to handle incidental page and column breaks, it's a different story.

Enivex:
The problem with doing so for incidental page and column breaks is that it may then no longer be clear whether it is a continuation of the previous paragraph, or a new paragraph. For explicit column- or pagebreaks you are presumably not filling up the entire page, so it's more immediately clear.

I would think it should only apply to explicit breaks, but I guess we should look at prior art. What do other typesetting applications do?

Edit: It could also be another toggle the user can control, of course.

laurmaedje:
That makes sense. Looking into prior art would be worthwhile, but in that case I'd say let's move on with #7722 either way.

PR #7722, opened 2026-01-19
Avoid undesirable `first-line-indent` after column break

lltsdyp:
## Overview
This pr is related to #7653 , which reports an unexpected triggering of `first-line-indent` at the beginning of the column:

``` typst
#set page("a8", flipped: true,columns: 2)
#set par(first-line-indent: (amount:1em,all: false))

#lorem(6)

#lorem(6)

#colbreak()

#lorem(6)

#lorem(6)
```

Before fixup:
image

After fixup:
image


crates/typst-layout/src/inline/mod.rs
```diff
@@ -246,7 +246,7 @@ fn configuration(
 /// inline layout that isn't a semantic paragraph.
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub enum ParSituation {
-    /// The paragraph is the first thing in the flow.
+    /// The paragraph is the first thing in the page or column.
```
Enivex: This looks like an incorrect description to me. A paragraph can be the first thing in a page or column and still follow other paragraphs.

lltsdyp: Do you think it would be clearer to update the description of `ParSituation::Consecutive` to indicate that it refers to a paragraph that does not start at the beginning of a page or column?

Enivex: What's wrong with the original description? I'm not sure why this has to change. That being said, I'll defer to someone who actually knows anything about the code.

lltsdyp: It seems that the issue stems from the fact that the beginning of a new column is not currently treated as the beginning of the flow.
Based on that understanding, I marked the first `ParElem` after `ColbreakElem` as `ParSituation::First`.
Given this, I wonder whether the current description of `ParSituation::Consecutive` could be clarified, as it may otherwise be a bit misleading.

laurmaedje:
Whether this fix is appropriate sort of depends on what we consider the desirable behavior for implicit page and column breaks. See https://github.com/typst/typst/issues/7653#issuecomment-3871291941 for more details.

laurmaedje:
I've adapted the comment and added a test. 

Thanks!

PR #7733, opened 2026-01-22
Ensure math accent is in front of the base

mkorje:
| Before | After |
| - | - |
| old | new |

This matches how normal accent layout works when done by the shaper.


laurmaedje:
Thanks! :)

PR #7734, opened 2026-01-22
Add Slovak and Polish translations

zouharvi:
For most languages in `translations/` the `page` field is lowercased, which I don't know why but I followed the consistency.

For Polish, you might be sceptical that the loanword `Listing` is used, but [it is really the case](https://galaxy.agh.edu.pl/~estrzalk/inz.pdf).


crates/typst-library/translations/pl.txt
```diff
@@ -4,8 +4,8 @@ equation = Równanie
 bibliography = Bibliografia
```
mgazeel:
```suggestion
bibliography = Piśmiennictwo
```

This should be: "Piśmiennictwo" or "Literatura", the AGH guidelines mention that "Bibliografia" is only a possible alternative.

Canonically "Piśmiennictwo" is the Polish word and used. See for example works in the UJ repository:
https://ruj.uj.edu.pl/entities/publication/12f1a8c6-6e89-44f3-b80d-e272a33795d4

crates/typst-library/translations/pl.txt
```diff
@@ -4,8 +4,8 @@ equation = Równanie
...
 heading = Sekcja
```
mgazeel:
```suggestion
heading = Rozdział
```

Just like in the Slovak version this is "Kapitola", it should be "Rozdział".

Definition from SJP PWN (Polish Science Publishing Authority):
Rozdział - "1. «jedna z części, na jakie dzieli się dzieło literackie, naukowe itp.»"
(One of the parts into which a literary or scientific work is divided into)

https://sjp.pwn.pl/slownik-synonimow/rozdzia%C5%82.html

crates/typst-library/translations/pl.txt
```diff
@@ -4,8 +4,8 @@ equation = Równanie
...
 outline = Spis treści
-raw = Program
+raw = Listing
```
mgazeel, line 7:
```suggestion
raw = Wypis
```

Hard disagree, this choice seems exclusive to AGH.

Works published by the nationwide PWN (Polish Science Publishing Authority) typically omit this completely:
https://ksiegarnia.pwn.pl/Wprowadzenie-do-algorytmow,1041498749,p.html

The closest Polish equivalent is "Wypis"

crates/typst-library/translations/pl.txt
```diff
@@ -4,8 +4,8 @@ equation = Równanie
...
 page = strona
-# footnote =
-# email =
-# telephone =
+footnote = Przypis
+email = E-mail
```
mgazeel:
```suggestion
email = Adres mailowy
```

E-mail is not a word. "Adres mailowy" reffers to an email address.

zouharvi:
Thanks @mgazeel!

crates/typst-library/translations/pl.txt
```diff
@@ -1,11 +1,11 @@
 figure = Rysunek
 table = Tabela
 equation = Równanie
-bibliography = Bibliografia
-heading = Sekcja
+bibliography = Piśmiennictwo
```
fuine, line 4: I'd actually argue that Bibliografia would be correct here, as the most generic term, which is also widely used. See for example this link: https://sjp.pwn.pl/poradnia/haslo/literatura-a-bibliografia;9543.html in which this quote stands out the most:

> Szczegółowe omówienie różnych typów bibliografii, w tym także specyficznych zestawień źródłowych w systemie „autor – rok” lub „autor – numer”, zawarto w nowo opublikowanej nakładem Wydawnictwa Naukowego PWN książce Edycja tekstów (rozdział 13, Bibliografia).

As you can see, the chapter on writing bibliographies in the popular book on text editing/typesetting is called "Bibliografia" and the expert also talks about different types of "bibliografii"

crates/typst-library/translations/pl.txt
```diff
@@ -1,11 +1,11 @@
...
+heading = Rozdział
 outline = Spis treści
-raw = Program
+raw = Wypis
```
fuine: I think "Listing" is the more popular option here, fwiw it's used here at the Warsaw University of Technology. You can also find this word in different dictionaries:
https://sjp.pl/listing
https://sjp.pwn.pl/slowniki/listing.html

crates/typst-library/translations/pl.txt
mgazeel: I have seen all three of the ones I mentioned used in scientific works commonly "Literatura" (in STEM), "Piśmiennictwo" (in Medicine, Law, etc.)

Notice that "Bibliografia" is used mostly in Master/Bachelor theses as in STEM these are written in Polish since most STEM journals require English. 

If there was a way to include alternatives I agree the Bibliografia should be there but as a default I think a Polish word should be included.

Remember that you can always freely change these so if your making a theme for your university where another term is used this is not a problem!

crates/typst-library/translations/pl.txt
mgazeel: Yeah, this is a tough one.

All the works I've seen published by PWN or papers simply do not use any qualifier before a raw block.

If both AGH and PW use Listing, I do agree it may be a better choice.

zouharvi: Thanks for the context! Keeping Listing.

crates/typst-library/translations/pl.txt
zouharvi: Thanks for the context! Keeping Piśmiennictwo.

laurmaedje:
Just to double-check: Are "Rozdział" and "Kapitola" really the correct terms? I obviously can't judge, but it translates as "Chapter" while the intended meaning is "Section". In the German translation, we also use "Abschnitt" instead of "Kapitel". And, for instance, DeepL translates "See Section 5.3 of the whitepaper." as "Zobacz sekcję 5.3 dokumentu technicznego.".

laurmaedje:
Also, CI is failing because a newline is missing at the end of the file(s).

zouharvi:
Thanks for the feedback and for looking into it!
- I based this translation on the existing Czech translation in `cs.txt` which also contains _"Kapitola"_. The meaning and usage seems to be similar for these, so either it should be _"Kapitola"_ for both Czech and Slovak or _"Sekce"_/_"Sekcia"_ (or _"Časť"_ in Slovak) for both. In practice, both are used: search for _"kapitola 7.3"_ [here](https://dspace.cuni.cz/bitstream/handle/20.500.11956/30637/120008139.pdf?sequence=1&isAllowed=y), _"kapitole 4.2.2"_ [here](https://dspace.cuni.cz/bitstream/handle/20.500.11956/99725/130232096.pdf?sequence=1&isAllowed=y)\*, or _"Sekcii 1.4"_ [here](https://dspace.cuni.cz/bitstream/handle/20.500.11956/119445/120361824.pdf?sequence=1&isAllowed=y)\*. Another term _"Podkapitola"_ (subchapter) is also possible, but less common and sensitive to the actual level. To me, _"Sekce"_/_"Sekcia"_ sound like translationese.
- For Polish, [this thesis template](https://ics.p.lodz.pl/~anetap/Diploma/SzablonPracyDyplomowej.pdf) describes that it should be divided to _"Rozdział"_ and _"Podrozdział"_ (subchapter). My understanding is that if we have only one word, then it should be _"Rozdział"_. I have not encountered _"Sekcja"_.

*The _"-e"_ or _"-i"_ suffix marks the locative case. I.e. _"In chapter 5.4..."_ is _"V kapitole 5.4..."_ but _"Chapter 5.4 shows..."_ is _"Kapitola 5.4 ukazuje..."_. Similar problems arise with plural forms. Declensions and various word forms are beyond the scope of this PR (though they do bear some relation to `form` argument of `cite`).

mgazeel:
From the Polish side: we dont really differentiate betwen "Chapter" and "Section" like English does.

Books have "Rozdziały" but then whitepapers also have "Rozdziały".

The word "Sekcja" does exist in Polish but it means division of eg. an organization.

"Klub Sportowy - Sekcja Piłkarska" = "Sport Club - Football Division"

Hence what DeepL gives very English carbon copy vibes, and doesnt make much sense.

laurmaedje:
Okay, thanks for the clarifications and the contribution!

PR #7737, opened 2026-01-22
Hashed reference test preparations

saecki:
These are a few fixes/improvements to merge before finally enabling testing of PDFs and SVGs using hashed references.

### DIFF
--- a/tests/src/collect.rs
+++ b/tests/src/collect.rs
@@ -468,17 +468,11 @@ impl Collector {
             }
 
             let path = entry.path();
-            let sub_path = path.strip_prefix(REF_PATH).unwrap();
-            let sub_dir = sub_path.components().next().unwrap();
-            let Some(output) =
-                TestOutput::from_sub_dir(sub_dir.as_os_str().to_str().unwrap())
-            else {
-                self.errors.push(TestParseError {
-                    pos: FilePos::new(path, 0),
-                    message: "unknown reference file".into(),
-                });
-                continue;
-            };
+            let output = (path.strip_prefix(REF_PATH).ok())
+                .and_then(|sub_path| sub_path.components().next())
+                .and_then(|sub_dir| sub_dir.as_os_str().to_str())
+                .and_then(TestOutput::from_sub_dir);
+            let Some(output) = output else { continue };
 
             match output.kind() {
                 TestOutputKind::File => self.check_dangling_file_references(path, output),
@@ -498,29 +492,29 @@ impl Collector {
         let name = &*stem;
 
         let Some((pos, attrs)) = self.seen.get(name) else {
-            self.errors.push(TestParseError {
-                pos: FilePos::new(path, 0),
-                message: "dangling reference output".into(),
-            });
+            self.errors.push(TestParseError::new(
+                FilePos::new(path, 0),
+                TestParseErrorKind::DanglingFile,
+            ));
             return;
         };
 
         if !attrs.stages.contains(output.into()) {
-            self.errors.push(TestParseError {
-                pos: FilePos::new(path, 0),
-                message: "dangling reference output".into(),
-            });
+            self.errors.push(TestParseError::new(
+                FilePos::new(path, 0),
+                TestParseErrorKind::DanglingFile,
+            ));
         }
 
         let len = path.metadata().unwrap().len() as usize;
         if !attrs.large && len > crate::REF_LIMIT {
-            self.errors.push(TestParseError {
-                pos: pos.clone(),
-                message: format!(
+            self.errors.push(TestParseError::new(
+                pos.clone(),
+                format!(
                     "reference output size exceeds {}, but the test is not marked as `large`",
                     FileSize(crate::REF_LIMIT),
                 ),
-            });
+            ));
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/src/collect.rs
```diff
@@ -296,7 +296,14 @@ impl TestOutput {
         }
     }
 
-    /// The path at which the live output will be stored for inspection.
+    /// The path at which the live output will be stored.
+    pub fn hash_path(&self, hash: impl Display, name: &str) -> PathBuf {
+        let ext = self.extension();
+        PathBuf::from(format!("{STORE_PATH}/by-hash/{hash}/{name}.{ext}"))
```
laurmaedje: is there a particular reason to have a directory for file instead of directly using the hash as the file name (or as part of the file name)?

tests/src/run.rs
```diff
@@ -254,7 +256,31 @@ impl<'a> Runner<'a> {
...
+                        std::fs::write(&live_path, live_data).unwrap();
+                    }
+                    TestOutputKind::Hash => {
+                        // Write the file to a path of its hash.
+                        let hash = T::make_hash(live);
+                        let hash_path = T::OUTPUT.hash_path(hash, &self.test.name);
+                        std::fs::create_dir_all(hash_path.parent().unwrap()).unwrap();
+                        std::fs::write(&hash_path, live_data).unwrap();
+
+                        // Create a link in the store directory.
```
laurmaedje, line 261: Are there upsides / downsides to the direction of the symlink? from dir -> by-hash vs the other way around?

tests/src/run.rs
```diff
@@ -254,7 +256,31 @@ impl<'a> Runner<'a> {
...
+                        std::fs::remove_file(&live_path).ok();
+
+                        let relative_path = hash_path.strip_prefix(STORE_PATH).unwrap();
+                        let link_path = Path::new("..").join(relative_path);
+
+                        #[cfg(target_family = "unix")]
+                        std::os::unix::fs::symlink(&link_path, &live_path).unwrap();
+                        #[cfg(target_family = "windows")]
+                        std::os::windows::fs::symlink_file(&link_path, &live_path)
+                            .unwrap();
```
laurmaedje, line 271: on other platforms, this will basically degrade gracefully into the store not working?

tests/src/collect.rs
```diff
@@ -459,25 +461,33 @@ impl Collector {
...
-            let sub_path = path.strip_prefix(crate::REF_PATH).unwrap();
+            let sub_path = path.strip_prefix(REF_PATH).unwrap();
             let sub_dir = sub_path.components().next().unwrap();
             let Some(output) =
                 TestOutput::from_sub_dir(sub_dir.as_os_str().to_str().unwrap())
             else {
+                self.errors.push(TestParseError {
+                    pos: FilePos::new(path, 0),
+                    message: "unknown reference file".into(),
+                });
```
laurmaedje: this caused an error on my system because of `.DS_Store`. do we need this?

tests/src/collect.rs
```diff
@@ -940,12 +939,42 @@ fn selected(name: &str, abs: PathBuf) -> bool {
...
 }
 
 impl Display for TestParseError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        write!(f, "{} ({})", self.message, self.pos)
+        write!(f, "{} ({})", self.kind, self.pos)
+    }
+}
+
+pub enum TestParseErrorKind {
```
laurmaedje, line 952: could you add brief docs to all types?

tests/src/tests.rs
```diff
@@ -157,13 +159,42 @@ fn clean() {
...
+                            dangling_hashes.entry(&error.pos.path).or_default();
+                        if let Some(idx) = NonZeroUsize::new(error.pos.line) {
+                            line_nrs.push(idx);
+                        }
+                    }
+                    TestParseErrorKind::Other(_) => (),
+                }
+            }
+
+            // Remove dangling hashes from file.
```
laurmaedje, line 181: Should this code perhaps live alongside the parsing code for the hashed refs file?

tests/src/collect.rs
saecki: It's not strictly necessary, and I'm not so much worried about hash collisions, but I've found it reassuring to see the name of the test in the path when looking for a specific hash. It also allows finding out which test a file belongs to, when browsing the store directory.

tests/src/run.rs
saecki: I think the other way around would always overwrite old live output, and there would be two different links pointing to the same file.

tests/src/run.rs
saecki: Yeah, the test suite always uses the hashed path. The links are just for more convenient manual inspection.

tests/src/collect.rs
saecki: Oh, no we don't really. I just thought it was a nice addition, but that's annoying.

tests/src/collect.rs
laurmaedje: I see. I wonder whether it would make sense to just have hash+name in the filename? The extra directories seem like unnecessary file system churn to me. And having the file their directly might also be nicer if you look directly into the directory (which is admittedly probably a rare thing to do).

tests/src/run.rs
laurmaedje: That makes sense

tests/src/collect.rs
saecki: Yeah, I like that :+1:

tests/src/tests.rs
saecki: I've now reused the already existing parsing logic, which should make this a bit more robust.

PR #7738, opened 2026-01-22
Add eval tests and empty attribute

saecki:
See: https://discord.com/channels/1054443721975922748/1217423226410373240/1461767283881017606

This adds the `eval` and `empty` attributes to the test suite.
The `eval` stage is the first target agnostic part of compilation which produces content. An `eval` test can be used for scripting tests that don't produce any content, which previously had to be either `paged` or `html` tests.
Some scripting don't produce any visual output, but still require realization/layouting to be run. For example because they have code inside a show-rule. These tests can now use the `empty` attribute, to indicate that they shouldn't generate any visual output.
The test runner will detect if a test unexpectedly produces empty output/doesn't produce any output, and will suggest using the correct test attributes. For example an `eval` test that produces non-empty content will fail, and suggest chaning the test to a `paged` or `html` test with the `empty` attribute.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -60,7 +60,7 @@ and SVGs. To make them go faster while developing, you can pass the `--stages`
 flag to only run certain targets/outputs. The available stages are the following
 ones:
 
-- `eval`: Evaluate the source code, this is (document-) target agnostic.
+- `eval`: Evaluate the source code. This is (document-) target agnostic.
 - `paged`: Compile the paged target and produce `render`, `pdf`, and `svg`
   outputs.
   - `render`: Produce `render` (`png`) output.


crates/typst/src/lib.rs
```diff
@@ -88,6 +88,35 @@ where
...
+
+    Warned { output, warnings: sink.warnings() }
+}
+
+/// Compile content into a fully layouted document.
+///
+/// - Returns `Ok(document)` if there were no fatal errors.
+/// - Returns `Err(errors)` if there were fatal errors.
+#[typst_macros::time]
+pub fn realize<D>(world: &dyn World, content: Content) -> Warned<SourceResult<D>>
```
laurmaedje: Realize is not quite the right term here as it performs both realization (as in `typst_realize::realize`) and compilation of the document in the target format. I'm also not sure I want to expose this in `typst` as the API here is somewhat more stable than the public API surface at large. 

Let's discuss this later.

saecki:
> * For the second error, the `empty` attribute doesn't actually help because of the first error. I would suggest hiding the second here if the first is emitted. For the first, I think starting with something other than the targets would be friendlier, e.g. `test produced empty content, but was marked [paged, render, pdf, svg]`.
> ```
> ❌ newline-continuation-if-else-comment (tests/suite/syntax/newlines.typ:69)
>   [paged, render, pdf, svg] test produced empty content
>     hint: consider making this an `eval` test
>   expected non-empty output [render] newline-continuation-if-else-comment
>     hint: consider adding the `empty` attribute
> ```

I've added two fields to the test runner, which collect the unexpected empty/non-empty output, and print a summarized error after all stages have been run.

> * For this error, I'm not sure why it repeats "issue-measure-image" in the error message. Also, the error message could be made more consistent with the other one, e.g. `test produced empty [render] output, but was not marked empty`.
> ```
> ❌ issue-measure-image (tests/suite/visualize/image.typ:366)
>   expected non-empty output [render] issue-measure-image
>   hint: consider adding the `empty` attribute
> ```

Fixed

> * Ideally, `html` would also have a notion of `empty`, but if there isn't a clean way to do this, it's not urgent.

This also became an issue when implementing the more robust empty/non-empty error reporting, so I've added a somewhat hacky solution for now.

crates/typst/src/lib.rs
saecki: I've just copy pasted the relevant code into `tests` for now. We're parsing and evaluating sources to content multiple times now, but it's quite cheap. I couldn't find a measurable difference when running the test suite.

saecki:
Seems like deleting the base branch closed the PR.

saecki:
Oops just noticed, I messed up the `--stages` flag. Should work as expected again.

tests/src/run.rs
```diff
@@ -697,3 +693,62 @@ impl<'a> Runner<'a> {
...
+            Route::default().track(),
+            &main,
+        )?
+        .content();
+
+        // Promote delayed errors.
+        let delayed = sink.delayed();
+        if !delayed.is_empty() {
+            return Err(delayed);
+        }
```
laurmaedje: This is not needed.

tests/src/run.rs
```diff
@@ -697,3 +693,62 @@ impl<'a> Runner<'a> {
         if line == 1 { format!("{col}") } else { format!("{line}:{col}") }
     }
 }
+
+/// A bunch of copy pasted code from the `typst` crate, so we don't have to
+/// change the public API.
+mod compile {
```
laurmaedje: maybe `mod eval` would make more sense?

saecki:
> Could you also already update the README here to mention the eval tests?

Oh right

tests/README.md
```diff
@@ -60,6 +60,7 @@ and SVGs. To make them go faster while developing, you can pass the `--stages`
 flag to only run certain targets/outputs. The available stages are the following
 ones:
 
+- `eval`: Evaluate the source code, this is (document-) target agnostic.
```
laurmaedje:
```suggestion
- `eval`: Evaluate the source code. This is (document-) target agnostic.
```
Or semicolon. Since it's two main clauses.

tests/README.md
```diff
@@ -82,21 +92,27 @@ specified. For instance, `--- my-test html ---` adds the `html` target to
...
+- `paged`: Tests paged output: `render`, `pdf`, `svg`
 - `html`: Tests HTML output against a reference HTML file.
+- `pdf`: Tests the PDF output specifically. The `pdf` stage is currently the
+  only fallible output, due to tagged PDF.
 - `pdftags`: Tests the output of the PDF tag tree.
 - `pdfstandard({standard})`: Sets the PDF standard used for testing PDFs and the
   PDF tag tree.
 - `large`: Permits a reference image size exceeding 20 KiB. Should be used
   sparingly.
+- `empty`: Indicates that a test shouldn't produce any non-trivial output, if it
```
laurmaedje:
```suggestion
- `empty`: Indicates that a test shouldn't produce any non-trivial output. If it
```
same here

tests/README.md
```diff
@@ -82,21 +92,27 @@ specified. For instance, `--- my-test html ---` adds the `html` target to
...
 There are, broadly speaking, three kinds of tests:
 
 - Tests that just ensure that the code runs successfully: Those typically make
   use of `test` or `assert.eq` (both are very similar, `test` is just shorter)
   to ensure certain properties hold when executing the Typst code. Generic
-  scripting tests that don't depend on the target should use `paged` as the test
-  target.
+  scripting tests that don't depend on the target should use the `eval`
+  attribute, if possible. If they rely on the evaluated content being compiled,
+  for example due to test code in show-rules, they should `paged|html empty`.
```
laurmaedje:
```suggestion
  for example due to test code in show-rules, they should use `paged|html empty`.
```

laurmaedje:
Just noticed a bit of a quirk:
```typ
--- playground paged ---
#(1 + "2")
```
```
❌ playground (tests/suite/playground.typ:2)
  missing output [render]
  not annotated
    Error [eval]: 3-10 cannot add integer and string
    Error [paged]: 3-10 cannot add integer and string
0 passed, 1 failed, 3217 skipped
```

This should probably only display the missing annotation once, for the earliest stage. Maybe this actually worked before we reverted change to the public API and before compile reran eval?

saecki:
Oh yeah, that worked properly before. Because the errors in the `eval` stage actually only got emitted there. I didn't think about that :/

PR #7739, opened 2026-01-22
Enable PDF and SVG testing using hashed references

saecki:
This is almost ready, just needs a little more manual testing...


tests/src/args.rs
```diff
@@ -55,12 +55,17 @@ pub struct CliArguments {
     pub num_threads: Option<usize>,
     /// Don't generate a HTML test report.
     #[arg(long)]
-    pub no_report: bool,
+    no_report: bool,
```
laurmaedje, line 58: any particular reason to make this one private?

tests/rerun/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "rerun"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
```
laurmaedje: could be removed

.github/workflows/ci.yml
```diff
@@ -31,13 +30,19 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
+        # TODO: This was initially useful for testing cross platform
+        # compatibility, but it might be nice to test more platforms.
+        arch: [amd64, arm64]
```
laurmaedje: this seems fairly unrelated and would double the amount of CI minutes. are we sure we need this?

tests/src/logger.rs
```diff
@@ -94,10 +95,20 @@ impl<'a> Logger<'a> {
 
         self.print(move |out| {
             if !result.errors.is_empty() {
-                writeln!(out, "❌ {test}")?;
-                if !crate::ARGS.compact {
+                if ARGS.use_github_annotations {
```
laurmaedje, line 115: that's nice :)

tests/src/report/mod.rs
```diff
@@ -21,284 +24,78 @@ impl TestReport {
...
-                        }
-                        while right.len() < left.len() {
-                            right.push(Line::EMPTY);
-                        }
-
-                        let left_line_nr = change.old_index().unwrap();
-                        left.push(line_unchanged(left_line_nr, &change));
+    missing_live.sort_by(|a, _, b, _| a.cmp(b));
+
+    // Find the git revisions that in which the hash references that are missing
```
laurmaedje: `s/that //`

tests/src/report/report.css
laurmaedje, line 1: let's use two space indent for CSS and JS

tests/src/run.rs
```diff
@@ -719,16 +722,72 @@ pub fn read_git_file(revision: &str, ref_path: &Path) -> Option<Vec<u8>> {
...
+                .map_err(|err| eco_format!("failed to parse date: {err}"))?;
+
+            let text = s.after();
+
+            Ok((hash.into(), date.unix_timestamp(), text.into()))
+        })
+        .collect::<StrResult<_>>()
+}
+
+/// Runa  git command
```
laurmaedje: "Run a git command".

tests/src/run.rs
```diff
@@ -719,16 +722,72 @@ pub fn read_git_file(revision: &str, ref_path: &Path) -> Option<Vec<u8>> {
     git_command(&["show", &rev_file]).ok()
 }
 
-/// Read a file from a specific git revision.
-pub fn git_command(args: &[&str]) -> Result<Vec<u8>, String> {
+/// Runs git blame on the file path and returns a list of of git revision,
```
laurmaedje: let's maybe put all the functions that call `git` into a separate `git.rs` file? So that it's clearer which things interact with it.

tests/src/report/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+
+pub fn write(reports: &mut [TestReport]) {
+    let report_path = Path::new(STORE_PATH).join("report.html");
+    let missing_path = Path::new(STORE_PATH).join("missing.txt");
+
+    if !ARGS.gen_report() {
+        _ = std::fs::remove_file(report_path);
+        _ = std::fs::remove_file(missing_path);
+        return;
+    }
```
laurmaedje: this is a bit strange. I might disable report generation precisely to keep my old report. I would suggest removing this and putting the whole `report::write` into `if ARGS.gen_report() { ... }`

tests/src/report/html.rs
```diff
@@ -0,0 +1,680 @@
...
+    pub static VIEW_DIFFERENCE: SvgIcon  = SvgIcon("M5 4a4 4 0 1 0 1.693 7.625c.14-.065.142-.258.016-.346A4 4 0 0 1 5 8c0-1.357.676-2.556 1.709-3.28.126-.088.124-.28-.016-.345A4 4 0 0 0 5 4m6 0a4.01 4.01 0 0 0-4 4c0 2.202 1.798 4 4 4s4-1.798 4-4-1.798-4-4-4m0 1.2c1.554 0 2.8 1.246 2.8 2.8s-1.246 2.8-2.8 2.8A2.79 2.79 0 0 1 8.2 8c0-1.554 1.246-2.8 2.8-2.8");
+
+    pub static ALIGN_TOP: SvgIcon = SvgIcon("M2 3.016v1.25h12v-1.25zm6 3-.443.441-2 2-.442.441.885.885.441-.441.934-.934v4.582h1.25V8.408l.932.934.443.441.883-.885-.442-.441-2-2Z");
+    pub const ALIGN_HORIZON: SvgIcon = SvgIcon("M7.375 10.174v4.857h1.25v-4.857zM8 9.504l-2.441 2.441-.444.444.885.882.441-.44.934-.935v-1.722h1.25v1.722l.934.934.441.441.883-.882-.442-.444Zm-.625-8.486v3.134l-.934-.931L6 2.779l-.885.883.444.443 2 2L8 6.547l.441-.442 2.002-2 .442-.443L10 2.78l-.441.442-.934.931V1.018ZM2 7.75V9h12V7.75Z");
```
laurmaedje: why is one const but all the other static?

tests/src/report/mod.rs
laurmaedje, line 1: Some more brief doc comments would be nice across the whole module tree

tests/Cargo.toml
```diff
@@ -25,10 +25,14 @@ typst-render = { workspace = true }
 typst-kit = { workspace = true }
 typst-svg = { workspace = true }
 typst-utils = { workspace = true }
+base64 = { workspace = true }
 bitflags = { workspace = true }
 clap = { workspace = true }
+codespan-reporting = { workspace = true }
```
laurmaedje: this does not seem to be used?

tests/src/run.rs
```diff
@@ -463,6 +534,23 @@ impl<'a> Runner<'a> {
...
     }
 
+    fn read_old_live_data<T: HashOutputType>(
+        &mut self,
+        old_hash: HashedRef,
+    ) -> Old<Vec<u8>> {
+        let old_hash_path = T::OUTPUT.hash_path(old_hash, &self.test.name);
+
+        let old_live_data = std::fs::read(&old_hash_path).inspect_err(|_| {
+            // TODO: Here we could check out a previous version of the
```
laurmaedje: is this todo still applicable?

tests/src/args.rs
saecki: The `gen_report()` method should be used instead. It will also return false when the `--update` flag is passesd. So directly using `no_report` would be kind of a footgun.

tests/src/run.rs
saecki: No, not really

tests/src/report/html.rs
saecki: Good question :D

tests/src/report/mod.rs
saecki: My initial thought was to avoid having stale reports, but keeping the report around is definitely useful.

saecki:
> Very nice! The smartness of the rerun tool is pretty amazing. I expected to have to enter the base revision manually.
> 
> Some initial comments:
> 
> * The testing README deserves an update :)

Done

> * It would be nice if test runner would emit a hint if any live output is missing to use the rerun tool

It now prints:             `"  run `cargo testit regen` to generate missing old live output"`

> * If the rerun tool can't be part of `cargo testit` (can it?), maybe we should have `cargo rerun` (or maybe another name that is a bit more descriptive).

The wrapper I've added now prompts: `generate missing live output [y/N]?` 

> On the report UX:
> 
> * The tool takes up a lot of space vertically even for small changes. Maybe it would be useful to group by test case rather than by file? And perhaps also support filtering by format.
> 
> * PDFs are hard to read because they are lacking a white background
> 
> * SVGs appear twice (once rendered and once with text) in the test report is a bit confusing. Perhaps this could be a toggle on one entry?

I think all of these have been addressed.

> * It would also be kinda nice if the PNG diffs were higher-res... with the store we _could_ have them in higher res. But I guess this is not really nicely feasible in the current setup. I think it's fine for now. Maybe we'll get rid of the PNGs in the repo anyway in the future ...

That would require somehow keeping track of old live output from the `render` output. Just storing the hashes could fail, since the `render` output isn't cross-platform deterministic. Maybe we would find out when the reference file was committed and then generate old output into a `by-rev` directory, where old live output is generated qualified with the git revision. This is just a half-baked idea, and should be explored in another PR, if at all.

.github/workflows/ci.yml
```diff
@@ -51,24 +53,32 @@ jobs:
...
-      - name: Update test artifacts
-        if: failure()
-        run: |
-          cargo test --workspace --test tests ${{ matrix.bits == 32 && '--target i686-unknown-linux-gnu' || '' }} -- --update
-          echo 'updated_artifacts=1' >> "$GITHUB_ENV"
-      - name: Upload updated reference output (for use if the test changes are desired)
-        if: failure() && env.updated_artifacts
+
+      - name: Upload test report
+        if: failure() || success() && github.event_name == 'pull_request'
```
laurmaedje: if github supports parentheses it would be good to have some here and above

.github/workflows/ci.yml
```diff
@@ -51,24 +53,32 @@ jobs:
...
-        if: failure()
+        env:
+          # Only generate annotations from one matrix configuration.
+          USE_GITHUB_ANNOTATIONS: ${{ startsWith(matrix.os, 'ubuntu-') && matrix.bits == 64 }}
+
+      - name: Generate missing live output and regenerate the test report
+        if: always()
+        run: cargo testit regen --rerun ${{ matrix.bits == 32 && '--target=i686-unknown-linux-gnu' || '' }}
+
+      - name: Upload test store
```
laurmaedje, line 67: I wonder whether we should even keep that. With the test report and the `--update` flag in place it seems to serve a very limited purpose. It was originally introduced in https://github.com/typst/typst/pull/5021. But the whole "Upload updated reference output (for use if the test changes are desired)" does not really make sense anymore with the hashed references.

tests/wrapper/src/main.rs
```diff
@@ -0,0 +1,276 @@
...
+    if args.peek().is_some_and(|arg| arg == "regen") {
+        args.next();
+        let cmd = parse_regen_command_args(args);
+        regen(cmd);
+        return;
+    }
+
+    // Forward the args to real test runner.
+    let status = Command::new("cargo")
+        .args(["test", "--workspace", "--test=tests", "--"])
```
laurmaedje, line 26: isn't this lacking the `--target` handling?

tools/test-helper/src/extension.ts
laurmaedje, line 1: this file is a bit of a battlefield because every editor seems to have different default auto-formatting. currently, we use the VS Code style formatting and if I save, it reformats a bunch of stuff. to keep it consistent with how it was before, I just pushed these changes.

tests/wrapper/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "test-wrapper"
+version = { workspace = true }
+rust-version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
```
laurmaedje, line 6: please add `publish = false`

tests/src/logger.rs
```diff
@@ -106,7 +151,37 @@ impl<'a> Logger<'a> {
             eprintln!("  pass the --update flag to update the reference output");
         }
 
-        self.failed == 0
+        let mut prompt_regen = false;
+        if ARGS.gen_report() {
+            prompt_regen = report::write(reports).unwrap_or(false);
+
+            if ARGS.open_report {
```
laurmaedje, line 158: we already depend on the `open` crate for the CLI. could use that here, too.

tests/README.md
laurmaedje, line 1: I made a few tweaks and rewrapped and pushed that.

.github/workflows/ci.yml
saecki: Maybe we could do it only on failure. I think it's useful to debug platform specific issues.

laurmaedje: that seems reasonable

tests/wrapper/src/main.rs
saecki: It kind of is, but it's also just not really needed I think. The `regen` command only has the `--target` option because of the CI. And for the first time the tests are run without the wrapper in the CI: https://github.com/typst/typst/blob/a9f884657e3f337c58585d51ae14b2272461bfc7/.github/workflows/ci.yml#L54

tests/src/logger.rs
saecki: Oh, nice

tests/wrapper/src/main.rs
laurmaedje: I see. Maybe it would still be good to handle the case.

saecki: I don't really like this flag at all, since it's conecptually one level higher than all the other ones., Maybe we can set the `CARGO_TARGET` environment variable at the start of the workflow and avoid explicitly passing the `--target` to cargo everywhere?

laurmaedje: Yes, that seems preferrable indeed. I dislike it, too.

laurmaedje:
Aside from clippy and two missing hashes due to new tests, there is one interesting CI failure: The change in https://github.com/typst/typst/pull/7786 affected a test, so the hash committed here was outdated after rebasing, but the live output could not be regenerated for the test report. However, I presume that this is merely the case because the codebase did not have the new test runner at that commit, otherwise it would have worked?

saecki:
> Aside from clippy and two missing hashes due to new tests, there is one interesting CI failure: The change in #7786 affected a test, so the hash committed here was outdated after rebasing, but the live output could not be regenerated for the test report.

I think it worked as expected :thinking: The report from [this CI run](https://github.com/typst/typst/actions/runs/21713886465?pr=7739) shows two failing cases. The first one is the test that changed (here the old live output was successfully generated), and the second one is the test that was added, but doesn't have references for PDF and SVG yet:
image

To make this more clear, I've now changed it so there is an `added` note next to the output:
image

When output is missing this would have already been explicitly shown:
image

Or were you talking about another CI run?

EDIT: When rebasing this doesn't work because the commit in which the hash that now changed was committed doesn't produce the the old output anymore. When merging (which is what I think the github CI does) the old commit is preserved, and the old output can be generated. I don't think there is a way  around this.
For reference images it would be the same. The rebased commit would fail CI because the reference image doesn't match the output that is (now) produced.

laurmaedje:
I was talking about this GitHub annotation in the review panel:
```
counter-display-matching-numbering-page (tests/suite/introspection/counter.typ:132)
  mismatched output
    live      | tests/store/pdf/counter-display-matching-numbering-page.pdf
    old       | bda57ff38abdf81881da51d6c54497e7
    new       | 85528bdbf5d23526460ea120192bccf3
    missing old live output tests/store/by-hash/bda57ff38abdf81881da51d6c54497e7_counter-display-matching-numbering-page.pdf
```
But I guess the "missing old live output" referred to the first run before regen?

saecki:
> But I guess the "missing old live output" referred to the first run before regen?

Oh yeah, that's a bit unfortunate. In theory we could generate the annotations after the missing live output has been generated, if the test suite failed. But since the test suite is only rerun if there is a `missing.txt` file, we can't just move the `USE_GITHUB_ANNOTATIONS` to the `cargo testit regen` command, because then we may generate no annotations at all. The live output doesn't improve the annotations in any other way, so I think it's fine for now.

saecki:
Almost forgot to include the `--base-revision` flag when CI in a PR passes. This way we get a report with the changes from the PR.

saecki:
// STOPSHIP: fix difference image view mode

saecki:
Ok, I've switched the image diffs to use a HTML canvas. I've been considering doing that for a while, but always managed to do without it. The final straw here was that the CSS `mix-blend-mode` doesn't work when different ancestors of HTML elements have a `scale` transform.
To make things somewhat performant, I've implemented lazy (re-)drawing of the canvas elements, which turned out to be faster in load time (and while actually using it), than the `` elements.
It's also a lot more flexible and finally enabled doing proper additive blending of the images. It would now be trivial to implement a `swipe` diff mode like on github, but that's an improvement left for the future :)

laurmaedje:
🎉

tests/src/git.rs
```diff
@@ -0,0 +1,107 @@
...
+) -> StrResult<Vec<(EcoString, i64, EcoString)>> {
+    let blame = git()
+        .args(["blame", "-e", "--date=iso8601-strict"])
+        .args(revision)
+        .arg(path)
+        .run()?;
+    let blame = std::str::from_utf8(&blame).map_err(|err| err.to_string())?;
+
+    // Format of git blame is:
+    // `<hash> <padded_nr>) <text>
```
istudyatuni, line 36:
```suggestion
    // (<hash> <padded_nr>) <text>
```

tests/src/git.rs
```diff
@@ -0,0 +1,107 @@
...
+                bail!("expected `(`");
+            }
+
+            // email
+            if !s.eat_if('<') {
+                bail!("expected email start (`<`)");
+            }
+            s.eat_until('>');
+            if !s.eat_if('>') {
+                bail!("expected email end (`<`)");
```
istudyatuni, line 55:
```suggestion
                bail!("expected email end (`>`)");
```

tests/src/git.rs
saecki: The format at the time of writing this comment was indeed `<hash> <padded_nr>) <text>`, because of the `git blame -s` flag. But the comment is completely outdated, good catch though :+1:

Issue #6430, opened 2025-06-11
Make line breaks a part of the display math delimiters, while allowing spaces in inline math


SeSodesa:
Inspired by [a conversation that was had on the Typst Discord server](https://discord.com/channels/1054443721975922748/1088371867913572452/1373939555350216796), I am now maing this pull request, since this suggestion seemed to garner some popularity. The basic gist is this: currently inline and display mathematics is differentiated by adding any whitespace after the opening dollar sign of math mode and before the closing dollar sign. In other words, `$x$` would result in an inline equation, whereas `$ x $` would produce a centered display equation.I am proposing a change, where both `$x$` and `$ x $` would produce an inline equation, and to produce a displayed equation, it would be compulsory to use newlines `\n`, carriage returns `\r`, or whatever combination of these functions as a line break on a given OS as display math delimiters:
```typst
$
	display math
$
```
In other words, the new display math delimiters would be `$\n` and `\n$` (with carriage returns added as necessary). The additional indentation is simply my personal preference, and would be allowed but not required.

This has at least 3 benefits that come to mind:

1. It is more explicit, and makes Typst source code more accessible in the official sense of the word. Displayed equations would always look like displayed equations, both in the source code and in whatever output format Typst supports. This would make the intention completely unambiguous.

2. Significant white space that is not displayed as explicitly as newlines are is always a bit dangerous, and it is easy to forget to add it, or accidentally omit it. Adding this requirement would probably reduce the chance of accidental inline and/or display mathematics being left in a document.

3. This would make mathematics markup behave more like `raw` markup currently does with `` `inline` `` and
   ``````typst
   ```
   block
   ```
   ``````
   which would increase the consistency of the Typst language.

The downside would be that this would be a breaking change for a lot of documents. At the same time, if a published document was already written with an older Typst version, having to compile it with the same compiler version would probably not be a huge issue, as the old versions are as easily accessed as newer ones.

I might implement this myself, but I have not yet had the time. I assume it would be a rather simple change.

MDLC01:
I don't have a strong opinion on whether this should be accepted or not, but I think it should be seriously evaluated because the presented arguments seem quite strong.

If it is decided that this should be implemented, I would suggest having at least one Typst version (if not two or even three) where uses of the inline display syntax (i.e., spaces but no line break) produces a warning, before actually introducing the change.

Enivex:
This would also make it easier to support arguments to math.equation in the syntax

laurmaedje:
[...]

PR #7741, opened 2026-01-22
Make `$  $` act like a block equation

### LINKED ISSUE #4791: Empty equation with spaces is not recognized as block equation
### Description

```typ
$  $
```
The ast for code above is
```
Markup: 5 [
    Equation: 4 [
        Dollar: "$",
        Math: 0,
        Space: "  ",
        Dollar: "$",
    ],
    Space: "\n",
]
```
And for some reason it is not considered as block equation. I guess it is related some edge case in parser so it is not correctly recognized.
```
[src\pretty\mod.rs:331:9] equation.block() = false
```

### Reproduction URL

_No response_

### Operating system

### Typst version

isuffix:

---

Changes the parsing behavior of equation syntax so empty equations that contain whitespace do parse as block-level.

Closes #4791

This is related to #6430, but that change will be considered separately.

### DIFF
--- a/crates/typst-syntax/src/ast.rs
+++ b/crates/typst-syntax/src/ast.rs
@@ -851,10 +851,33 @@ impl<'a> Equation<'a> {
 
     /// Whether the equation should be displayed as a separate block.
     pub fn block(self) -> bool {
-        let is_space = |node: Option<&SyntaxNode>| {
-            node.map(SyntaxNode::kind) == Some(SyntaxKind::Space)
-        };
-        is_space(self.0.children().nth(1)) && is_space(self.0.children().nth_back(1))
+        // The parser likes to group adjacent trivia, so if the equation's body
+        // is an empty Math node, e.g. in `$ /**/ $`, the CST will look like:
+        // `Eqn [ Math(<empty>) Space(" ") Trivia("/**/") Space(" ") ]`.
+        // Yet we still want to treat this as a block, so we check the following
+        // node by skipping an empty math node at the start.
+        let mut front_iter = self.0.children().enumerate().skip(1);
+        let mut front_pair = front_iter.next();
+        if let Some((_, front)) = front_pair
+            && front.is_empty()
+            && front.kind() == SyntaxKind::Math
+        {
+            front_pair = front_iter.next();
+        }
+
+        if let Some((front_idx, front)) = front_pair
+            && let Some((back_idx, back)) = self.0.children().enumerate().nth_back(1)
+        {
+            front.kind() == SyntaxKind::Space
+                && back.kind() == SyntaxKind::Space
+                // If the front and back spaces are the same one, require that
+                // it has multiple chars or, if one char, that it is a newline.
+                // `parse::<char>` is `Ok` if there was only a single char.
+                && (front_idx != back_idx
+                    || front.text().parse::<char>().map_or(true, is_newline))
+        } else {
+            false
+        }
     }
 }
 
--- a/tests/suite/math/equation.typ
+++ b/tests/suite/math/equation.typ
@@ -1,6 +1,86 @@
 // Test alignment of block equations.
 // Test show rules on equations.
 
+--- math-equation-block-syntax eval ---
+// Block equations require whitespace on both sides.
+#assert($ x $.block)
+#assert($ x y z $.block)
+#assert($
+x
+$.block)
+#assert($
+x
+y
+$.block)
+
+// Just whitespace only creates a block equation if it contains multiple
+// characters or a single newline.
+#assert($  $.block)
+#assert($       $.block)
+#assert(not $ $.block) // A single space doesn't make a block.
+#assert($
+$.block)
+#assert($
+
+
+$.block)
+
+// If either side lacks whitespace, we have an inline equation.
+#assert(not $x$.block)
+#assert(not $x
+$.block)
+#assert(not $
+x$.block)
+
+// Even if the inside has lots of whitespace.
+#assert(not $x
+
+
+y$.block)
+
+// And inline comments _do_ interrupt the block status.
+#assert($ /**/ $.block)
+#assert(not $/**/ $.block)
+#assert(not $ /**/$.block)
+#assert(not $/**/
+$.block)
+
+--- math-equation-block-syntax-unicode eval ---
+// Test equation block edge cases using odd unicode whitespace characters.
+// Just a representative sample, not exhaustive.
+#let spaces = ("\t", "\u{A0}", "\u{2009}", "\u{205F}", "\u{3000}")
+#let newlines = ("\r", "\u{B}", "\u{2029}")
+#let non-spaces = ("\u{200B}", "\u{200C}", "\u{2060}")
+
+// To debug these, add `message: str(thing.to-unicode(), base: 16)`.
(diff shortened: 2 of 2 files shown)


crates/typst-syntax/src/parser.rs
```diff
@@ -223,8 +223,10 @@ fn equation(p: &mut Parser) {
 /// Parses the contents of a mathematical equation: `x^2 + 1`.
 fn math(p: &mut Parser, stop_set: SyntaxSet) {
     let m = p.marker();
-    math_exprs(p, stop_set);
-    p.wrap(m, SyntaxKind::Math);
+    let count = math_exprs(p, stop_set);
+    if count != 0 {
+        p.wrap(m, SyntaxKind::Math);
+    }
 }
```
isuffix, line 228: Note that this interacts with the implementation of `Equation::block()` in the AST, but that did not require changes. And by not creating a Math element, we depend more on the `Default` impl of the `Math` struct in the AST when calling `Equation::body()`.

https://github.com/typst/typst/blob/4baa244fc8dc1fe46e122a32a8834ad2e5dff612/crates/typst-syntax/src/ast.rs#L818-L831

laurmaedje: The `Default` impls were primarily intended as "don't crash on malformed ASTs because eval is called for IDE stuff" rather than being relied on for compilation itself, so I'm not sure I'm 100% comfortable with this.

Is there an alternative design that is more explicit?

isuffix: How about just returning an `Option<Math>` in `Equation::body` and handling it in eval?

Enivex:
Is there a particular reason why this should be considered block level?

isuffix:
> Is there a particular reason why this should be considered block level?

Because it has whitespace after the first dollar sign and whitespace before the last dollar sign. That they're the same space is a bit odd, but I don't think it should affect the rule. Note that this PR also changes the behavior of `$\n$` for the same reason, and I think we can agree that that one feels much more blocky.

Enivex:
I would find it stranger if `$$` is inline, `$ $` is display, `$ x$` is inline, `$x $` is inline and `$ x $` is display.

The odd one out here is `$ $`.

This is not an artificial thing either. Suppose my editor is set to match `$` but not spaces. If I want to write `$ x $` my file would go through the chain `$$` `$ $` `$ x$` and `$ x $`. With this change my preview would be terribly unstable, since the equation would oscillate between inline and display twice.

laurmaedje:
Worth mentioning that this change also turns `$  $` (two spaces) block-level, which is currently also treated as inline.

crates/typst-syntax/src/parser.rs
laurmaedje: I'm not sure I fully understand this proposal. Let's discuss it later.

laurmaedje:
@Enivex how would you treat this?
```typ
$
$
```
and this
```typ
$  $
```

Enivex:
> @Enivex how would you treat this?
> 
> ```typst
> $
> $
> ```
> 
> and this
> 
> ```typst
> $  $
> ```

This is hard to answer without straying into the territory of <https://github.com/typst/typst/issues/6430>, but they are both more blocky than `$ $`. The first one has only one whitespace character, but that character is a newline. The second has two spaces instead of one.

(My actual hope is that only newlines will trigger block equations in the future, or that there's some other non whitespace based way to indicate it.)

crates/typst-syntax/src/parser.rs
isuffix: I've updated so that this function is unchanged and we implement the new behavior in the AST.

isuffix:
I've updated this so that `$ $` is not a block, but `$\n$` and `$  $` (two spaces) are.

crates/typst-syntax/src/ast.rs
```diff
@@ -851,10 +851,30 @@ impl<'a> Equation<'a> {
 
     /// Whether the equation should be displayed as a separate block.
     pub fn block(self) -> bool {
-        let is_space = |node: Option<&SyntaxNode>| {
-            node.map(SyntaxNode::kind) == Some(SyntaxKind::Space)
-        };
-        is_space(self.0.children().nth(1)) && is_space(self.0.children().nth_back(1))
+        // The `.unwrap()` calls in this function won't panic because there will
+        // always be an extra dollar-sign node on the opposing side.
```
laurmaedje: Is this guaranteed to hold even in a potentially erroneous CST? Since the AST is also used for parses with errors in IDE tooling.

crates/typst-syntax/src/ast.rs
```diff
@@ -851,10 +851,30 @@ impl<'a> Equation<'a> {
...
+
+        front.kind() == SyntaxKind::Space
+            && back.kind() == SyntaxKind::Space
+            // If the front and back spaces are the same one, require that it
+            // has multiple characters or, if one char, that it is a newline.
+            && (front_idx != back_idx || {
+                match front.text().char_indices().next_back() {
+                    Some((0, c)) => is_newline(c), // A single char must be a newline.
+                    _ => true
+                }
```
laurmaedje: could we do this instead?
`front.text().parse::<char>().map_or(true, is_newline)`

crates/typst-syntax/src/ast.rs
```diff
@@ -851,10 +851,30 @@ impl<'a> Equation<'a> {
...
+        let (mut front_idx, mut front) = front_iter.next().unwrap();
+        // The parser likes to group adjacent trivia, so if the equation's body
+        // is an empty Math node, e.g. in `$ /**/ $`, the CST will look like:
+        // `Eqn [ Math(<empty>) Space(" ") Trivia("/**/") Space(" ") ]`.
+        // Yet we still want to treat this as a block, so we check the following
+        // node by skipping an empty math node at the start.
+        if front.is_empty() && front.kind() == SyntaxKind::Math {
+            (front_idx, front) = front_iter.next().unwrap();
+        }
+        let (back_idx, back) = self.0.children().enumerate().nth_back(1).unwrap();
```
laurmaedje: I would find this grouping more readable because it emphasizes the general parallelity of front & back
```rust
// The `.unwrap()` calls in this function won't panic because there will
// always be an extra dollar-sign node on the opposing side.

// The parser likes to group adjacent trivia, so if the equation's body
// is an empty Math node, e.g. in `$ /**/ $`, the CST will look like:
// `Eqn [ Math(<empty>) Space(" ") Trivia("/**/") Space(" ") ]`.
// Yet we still want to treat this as a block, so we check the following
// node by skipping an empty math node at the start.
let mut front_iter = self.0.children().enumerate().skip(1);
let (mut front_idx, mut front) = front_iter.next().unwrap();
if front.is_empty() && front.kind() == SyntaxKind::Math {
    (front_idx, front) = front_iter.next().unwrap();
}

let (back_idx, back) = self.0.children().enumerate().nth_back(1).unwrap();
```

laurmaedje: There are also various possible renditions of this with `skip_while` or `find`, but I guess you already thought of those and preferred this one. I don't have a strong preference, the mutable version is also fine with me.

laurmaedje:
Thanks!

PR #7742, opened 2026-01-22
Improve errors for incomplete math operators and disallow linebreaks in operands

### LINKED ISSUE #1435: _ and ^ eat newline in math
### Description

```
$
    a = b_ \
    c = d
$
```
will render as 
image

The same is the case if _ is replaced with ^

This is not ideal, because it makes the output jump around whenever you add a sub- or superscript at the end of an existing line in an equation.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

isuffix:

---

This PR tightens up our handling of general operators in math (primes will be dealt with separately).

Closes #1435

### First commit
In this commit, we now error more predictably when an operator has a missing side, and the error message we give is much better. This resolves the odd behavior of `$(1/)$` parsing the close-paren as the fraction's denominator, now giving the error, "expected an expression to the right of the operator".

I implemented this by making better use of the `stop_set` argument in `math_expr_prec`, ensuring that we thread it through recursive operator calls when parsing an operator's right operand, and erroring if we start the call at the stop set.

This also improves the error for unclosed delimiters used for grouping. So `$($` alone is still fine, but `$1/($` is an error. This was done by passing in the precedence of the surrounding expression to `math_delimited` and checking that it was higher than zero if we didn't find a closing delimiter.

And I removed the `MATH_EXPR` syntax set because the `unexpected` error message case was really only used for these operators, and I would prefer to give a more explicit error for them.

### Second commit
This is simpler and just disallows linebreaks and alignment points as direct operands, requiring parentheses or escaping to place them inside a math operator. Doing so on their own was never actually useful, so this should just improve the editing experience in math.
 
I did this by adding `Linebreak` and `MathAlignPoint` to a new `TOP_LEVEL` syntax set in `math_exprs` that we handle explicitly at the top-level, but treat as part of the stop set for all calls to `math_expr_prec`. This causes us to error whenever an operator would be adjacent to one of these kinds, for the same reason as we error for `$/$`, but it does not change the behavior if these would just join as part of the top-level math sequence.

### DIFF
--- a/crates/typst-syntax/src/parser.rs
+++ b/crates/typst-syntax/src/parser.rs
@@ -3,7 +3,7 @@ use std::ops::{DerefMut, Index, IndexMut, Range};
 
 use ecow::{EcoString, eco_format};
 use rustc_hash::{FxHashMap, FxHashSet};
-use typst_utils::{default_math_class, defer};
+use typst_utils::{default_math_class, defer, matching_delim};
 use unicode_math_class::MathClass;
 
 use crate::set::{SyntaxSet, syntax_set};
@@ -456,16 +456,15 @@ fn math_delimited(p: &mut Parser, prec: u8) {
         // If we had no closing delimiter, just produce a math sequence.
         if prec > 0 {
             // Unless we were to the right of an operator, then we error.
-            let common_closing = match p[m].text().as_str() {
-                "(" => Some(')'),
-                "[" => Some(']'),
-                "{" => Some('}'),
-                _ => None,
-            };
             p[m].convert_to_error("unclosed delimiter");
             p[m].hint("delimiters must be correctly matched when used for grouping");
-            if let Some(c) = common_closing {
-                p[m].hint(eco_format!("try adding a closing delimiter: `{c}`"));
+            let open = p[m].text().as_str().chars().next().unwrap();
+            if let Some(close) = matching_delim(open) {
+                p[m].hint(eco_format!("try adding a closing delimiter: `{close}`"));
+                p[m].hint(eco_format!(
+                    "or escape the delimiter with a backslash \
+                     to display it verbatim: `\\{open}`"
+                ));
             }
         }
         p.wrap(m, SyntaxKind::Math);
--- a/crates/typst-library/src/math/matrix.rs
+++ b/crates/typst-library/src/math/matrix.rs
@@ -1,6 +1,6 @@
 use smallvec::{SmallVec, smallvec};
 use typst_syntax::Spanned;
-use typst_utils::{Numeric, default_math_class};
+use typst_utils::{Numeric, default_math_class, matching_delim};
 use unicode_math_class::MathClass;
 
 use crate::diag::{At, HintedStrResult, StrResult, bail};
@@ -300,21 +300,6 @@ impl Delimiter {
     pub fn get(self) -> Option<char> {
         self.0
     }
-
-    pub fn find_matching(self) -> Self {
-        match self.0 {
-            None => Self::none(),
-            Some('[') => Self(Some(']')),
-            Some(']') => Self(Some('[')),
-            Some('{') => Self(Some('}')),
-            Some('}') => Self(Some('{')),
-            Some(c) => match default_math_class(c) {
-                Some(MathClass::Opening) => Self(char::from_u32(c as u32 + 1)),
-                Some(MathClass::Closing) => Self(char::from_u32(c as u32 - 1)),
-                _ => Self(Some(c)),
-            },
-        }
-    }
 }
 
 /// A pair of delimiters (one closing, one opening) used for matrices, vectors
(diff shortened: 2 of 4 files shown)


crates/typst-syntax/src/parser.rs
```diff
@@ -438,6 +444,20 @@ fn math_delimited(p: &mut Parser) {
...
+            let common_closing = match p[m].text().as_str() {
+                "(" => Some(')'),
+                "[" => Some(']'),
+                "{" => Some('}'),
+                _ => None,
+            };
+            p[m].convert_to_error("unclosed delimiter");
+            p[m].hint("delimiters must be correctly matched when used for grouping");
+            if let Some(c) = common_closing {
+                p[m].hint(eco_format!("try adding a closing delimiter: `{c}`"));
```
laurmaedje: perhaps we should add a second hint like this?
```
or escape the delimiter with a backslash to display it verbatim: `\\{c}`
```
though I suppose this hint would also make sense on some of the other errors and I'm not sure whether it's too much overall

tests/suite/math/call.typ
```diff
@@ -17,6 +17,12 @@ $func(a$
 // Error: 5-6 unclosed delimiter
 $sin(x$
 
+--- math-call-missing-operator-sides paged ---
+// Error: 11 expected an expression to the right of the operator
+// Error: 15 expected an expression to the right of the operator
+// Error: 22 expected an expression to the right of the operator
+$func(1^2_, 1/; , , √ )$
```
laurmaedje, line 24: _much_ better than the old one:
```
error: unclosed delimiter
  ┌─ math.typ:1:5
  │
1 │ $func(1^2_, 1/; , , √ )$
  │      ^
```

crates/typst-syntax/src/parser.rs
```diff
@@ -286,30 +286,36 @@ fn math_expr_prec(p: &mut Parser, min_prec: u8, stop_set: SyntaxSet) {
...
-        _ => p.expected("expression"),
+        // Infix/postfix operators here produce an error, but we'll still parse
+        // the operator in the loop below for resilience.
+        SyntaxKind::Hat | SyntaxKind::Slash | SyntaxKind::Underscore => {
+            p.expected("an expression to the left of the operator");
+        }
+
+        // Any other kinds must have been due to an error.
+        SyntaxKind::Error => p.eat(),
+        _ => panic!("the lexer doesn't produce any other syntax kinds in math"),
```
laurmaedje: `panic!` -> `unreachable!`?

crates/typst-syntax/src/parser.rs
```diff
@@ -438,6 +444,20 @@ fn math_delimited(p: &mut Parser) {
         p.wrap(m, SyntaxKind::MathDelimited);
     } else {
         // If we had no closing delimiter, just produce a math sequence.
+        if prec > 0 {
+            // Unless we were to the right of an operator, then we error.
+            let common_closing = match p[m].text().as_str() {
+                "(" => Some(')'),
+                "[" => Some(']'),
+                "{" => Some('}'),
```
laurmaedje: I have a bit of a Déjà vu moment as if we already had some code at some point that does this for all parens, but I may be misremembering.

mkorje: There's some code for this in typst-library, tje `Delimiter::find_matching` function: https://github.com/typst/typst/blob/main/crates%2Ftypst-library%2Fsrc%2Fmath%2Fmatrix.rs#L304

crates/typst-syntax/src/parser.rs
isuffix: Done

crates/typst-syntax/src/parser.rs
isuffix: Should I factor this out to `typst-utils`?

laurmaedje: actually, I think there is no particular reason why `default_math_class` is in `typst-utils` instead of `typst-syntax`. Almost the whole crate tree depends on typst syntax, so maybe it fits better there?

laurmaedje: let's just put it in typst-utils, not sure where exactly to put it in typst-syntax and it's actually not strictly syntactical

isuffix:
Oof, sorry :grimacing:. We'll need to rebase `math-syntax` in `typst/typst` to get these commits to go away

crates/typst-syntax/src/parser.rs
isuffix: I figure these kinds of general hints deserve a more general fixup kind of improvement like the Rust compiler gives, but I'm fine to leave them in until we have such a functionality since they are useful.

crates/typst-syntax/src/parser.rs
isuffix: Done in the most recent commit.

laurmaedje:
I updated the `math-syntax` branch to the latest main.

laurmaedje:
Thank you!

Issue #3206, opened 2024-01-16
Interaction between lists and paragraphs


Enivex:
Currently, a tight list attaches to the preceding paragraph when there is no whitespace, and a new paragraph is always created after the list.

![image](https://github.com/typst/typst/assets/2191754/35c5d8f8-5d9d-477c-a808-6383ce15ed5e)

In this case, this behavior is not desirable, because everything here is really part of the _same_ paragraph.

For comparison, the behavior in LaTeX is that the next line is also part of the paragraph (new paragragraphs are indented)
![image](https://github.com/typst/typst/assets/2191754/0c680a74-a9c4-4752-be7e-645d99a8c35a)

_unless_ whitespace is added
![image](https://github.com/typst/typst/assets/2191754/7f4900b4-5d8c-4e6d-8c17-1e6c92f32969)

BrandonHenke:
This is related to how typst handles paragraphs. A list is a different element than a paragraph, so, no matter how many lines are placed between the list and the next piece of content, it will treat the next piece of content as its own paragraph.

A paragraph is only indented if it follows another paragraph (and has first-line-indent != 0 and some alignment stuff). Typst doesn't consider a list a paragraph. Hence, the content that comes after the list will be its own paragraph, and that paragraph will not be indented.

I've forked the repo, because I think the paragraph element is too limited as it is. I'm attempting to change that.

Enivex:
I modified the title to reflect that this is a wider issue

Enivex:
Another example: Figures that are not `top` or `bottom`

laurmaedje:
I'm not really convinced that a figure really makes sense as part of a paragraph.

BrandonHenke:
> I'm not really convinced that a figure really makes sense as part of a paragraph.

How sensical you or anyone else thinks it is doesn't matter.
This is a typesetting software.
Disallowing some sort of formatting because it doesn't make sense to someone would be limiting.

A figure is a block of content.
A paragraph is a string of connected ideas.
If one of those ideas is best given as an equation block, code block, figure block, etc., then discriminating between content blocks is limiting to the string of connected ideas used to form the paragraph.

BrandonHenke:
I'm sorry if I seemed rude in my previous comment. I in no way meant it antagonistically. My apologies!

OrangeX4:
For some people who need a workaround:

```typst
#set page(width: 30em)
#set par(first-line-indent: (amount: 2em, all: true))
[...]

PR #7747, opened 2026-01-23
Make `par.first-line-indent` foldable

### LINKED ISSUE #7052: Make `par.first-line-indent` fold

YDX-2147483647:
In Typst 0.14, the following typst code emits an error (Dictionary does not contain key "amount").
In this PR, it enables `all` without affecting the `amount`. It would be helpful for templates and also handy for one-off toggling.

```typst
#set par(first-line-indent: (all: true))
```

Resolves #7052
Continues #7123 (which was rejected because AI didn't fully implement it)

### Values from `context`

[GitHub search](https://github.com/search?q=par.first-line-indent+lang%3Atypst&ref=opensearch&type=code) shows that people are assuming `type(par.first-line-indent.amount) == length`. (https://github.com/typst/typst/issues/3206#issuecomment-3013274959 has a concrete example.)
Therefore, this PR keeps this assumption.

Note that this behaviour is _not_ consistent with `grid.stroke`, which is also both foldable and smart.

```typst
// Before all set rules
#context par.first-line-indent // (amount: 0pt, all: false)

// After partial set rules
#set par(first-line-indent: 2em)
#context par.first-line-indent.all // false
```

```typst
// Before all set rules
#context grid.stroke // (:) — an empty dictionary

// After partial set rules
#set grid(stroke: red)
#context grid.stroke.thickness // auto, not 1pt
```

<details><summary>Outdated content</summary>

### Repeated code

The ad hoc `take` function that `cast!` `Dict -> FirstLineIndent` is copied from `stroke.rs`.
I'm not sure if it's worth merging the two functions.

https://github.com/typst/typst/blob/7cfde04f4bc8fb9ae8a65dc8e10809ad5a83c098/crates/typst-library/src/visualize/stroke.rs#L379-L388

</details>


tests/suite/model/par.typ
```diff
@@ -156,6 +156,32 @@ starts a paragraph, also with indent.
...
+
+#set par(first-line-indent: (all: true))
+#check((amount: 2em, all: true))
+
+/// The following three ways should be the same.
+#set par(first-line-indent: 7em)
+#check((amount: 7em, all: true))
+#set par(first-line-indent: (amount: 1em))
+#check((amount: 1em, all: true))
+#set par(first-line-indent: (amount: 8em, all: auto))
```
laurmaedje: `all` should not be able to take `auto`

crates/typst-library/src/model/par.rs
```diff
@@ -412,6 +412,11 @@ pub struct ParElem {
     ///
     /// Even the first one.
     /// ```
+    #[fold]
+    #[default(FirstLineIndent {
+        amount: Smart::Custom(Default::default()),
+        all: Smart::Custom(Default::default()),
```
laurmaedje: It's a bit unfortunate that the defaults are technically specified twice: Once here and then once in `FixedFirstLineIndent::default()`. If the default were not just the type defaults, this would be more obvious.

YDX-2147483647:
I reorganized it and replaced "FixedFirstLineIndent" with accessors. It's cleaner now.

crates/typst-library/src/model/par.rs
```diff
@@ -652,8 +661,39 @@ cast! {
 impl From<FirstLineIndent> for Dict {
     fn from(indent: FirstLineIndent) -> Self {
         dict! {
-            "amount" => indent.amount,
-            "all" => indent.all,
+            "amount" => indent.amount(),
+            "all" => indent.all(),
```
laurmaedje: These properties should not be included in the dictionary if they are `None`, rather than then using the default, since the `None` represents the strongly-typed version of an absent key.

YDX-2147483647: As explained in the PR description, I want `context par.first-line-indent` to have defaults filled.
`impl From<FirstLineIndent> for Dict` was added in #5768, and it looks like the only usage was in `cast!`, so I implemented the logic here.

```rust
cast! {
    FirstLineIndent,
    self => Value::Dict(self.into()),
```

I've just removed `impl From<FirstLineIndent> for Dict`. How about now?

laurmaedje: > As explained in the PR description, I want context par.first-line-indent to have defaults filled.

That shouldn't be an issue, since `context par.first-line-indent` will observe the folded value and since the default is initialized with `Some(_)` fields, they cannot be `None` after folding in practice. My comment was only about having the correct semantical behavior of the code. The code for folding across the code base is already a bit all over the place and if we give up on the semantics, I fear at some point nobody will understand it anymore, including me.  

If you reintroduce the `impl From<FirstLineIndent> for Dict` with the `None` handling, I think it should all work as you expect.

At some point, I might create some additional infrastructure to make foldable properties less confusing to implement.

YDX-2147483647: > That shouldn't be an issue

Indeed! I was misled by my initial implementation. (In that version, the struct `FirstLineIndent` and the parameter `par.first-line-indent` have different default values. Therefore, that didn't work.)

Now updated.

> create some additional infrastructure to make foldable properties less confusing to implement

I'm looking forward to it. It did take me some time to figure out the relationship between `Fold` and [`AlternativeFold`](https://docs.rs/typst-library/0.14.2/typst_library/foundations/trait.AlternativeFold.html), what should be stored if a key is not given, …

laurmaedje: The diff now is unfortunately still not quite what I meant. 🫠

Now, `None` values would be turned into Typst `none` values in the dictionary rather than being absent.

What I meant is that they would not be inserted at all, like in `IntoValue for Sides<Option<T>>`.

YDX-2147483647: Oh, I guess this version would match what you meant…

laurmaedje:
Thanks!

crates/typst-library/src/model/par.rs
laurmaedje: Yup

PR #7752, opened 2026-01-24
Hints for invalid characters in code mode

### LINKED ISSUE #7749: Add a helpful message to fix `#` error in code mode
### Description

[Context](https://discord.com/channels/1054443721975922748/1088371867913572452/1464003120785915996)

Image

The current ``error: the character `#` is not valid in code`` is kind of unhelpful, especially after seeing https://forum.typst.app/t/most-basic-page-numbering-on-footer/7727/5?u=andrew.

How about adding `` hint: try removing `#` ``? The solution shouldn't even be hard to come up with (in [`crates/typst-syntax/src/lexer.rs`](https://github.com/typst/typst/blob/24b493814e15281dc0835571b3c55b2866856dc2/crates/typst-syntax/src/lexer.rs?plain=1#L795)).

```typ
#{
  #let a = 5
  #a
}
```

### Use Case

This makes the intention behind the error clear, and will actually fix some portion of cases of this error, just because an extra `#` was added into otherwise working code.

isuffix:
Closes #7749


MDLC01:
I'm not sure always suggesting to remove the character is a good idea. In `#!not-a-shebang`, suggesting using `not` instead (because `!` is used for that in many programming languages) would be more useful in my opinion.

isuffix:
Great point! I've updated the imlpementation to only recommend removing `#`, and added much more explicit hints for other operators like `&&` and `||`:
```typ
--- syntax-bugs-extra-hash paged ---
#{
  // Error: 3-4 the character `#` is not valid in code
  // Hint: 3-4 you are already in code mode
  // Hint: 3-4 try removing the `#`
  #
}

--- syntax-bugs-bad-boolean-ops paged ---
// Test writing invalid boolean operators:
#{
  // Error: 3-5 `&&` is not valid in code
  // Hint: 3-5 in Typst, `and` is used for logical AND
  &&
  // Error: 3-4 the character `&` is not valid in code
  &
  // Error: 3-5 `||` is not valid in code
  // Hint: 3-5 in Typst, `or` is used for logical OR
  ||
  // Error: 3-4 the character `|` is not valid in code
  |
  // Error: 3-4 the character `!` is not valid in code
  // Hint: 3-4 in Typst, `not` is used for negation
  // Hint: 3-4 or did you mean to write `!=` for not-equal?
  !
  // Error: 3-4 the character `!` is not valid in code
  // Hint: 3-4 in Typst, `not` is used for negation
  // Hint: 3-4 or did you mean to write `!=` for not-equal?
  // Error: 5-6 unexpected equals sign
  ! =
  // Error: 3-4 the character `~` is not valid in code
  // Hint: 3-4 in Typst, `not` is used for negation
  ~
  // Error: 3-5 `~=` is not valid in code
  // Hint: 3-5 in Typst, `!=` is used for not-equal
  ~=
}
```

tests/suite/syntax/shebang.typ
```diff
@@ -4,4 +4,6 @@
 #!typst compile
 
 // Error: 2-3 the character `!` is not valid in code
+// Hint: 2-3 in Typst, `not` is used for negation
+// Hint: 2-3 or did you mean to write `!=` for not-equal?
```
laurmaedje: these hints are perhaps a bit misleading, because neither `not` nor `!=` will work directly in a hash expression. do you think we could detect this scenario?

isuffix: Yes, very easy to do in the lexer.

isuffix:
@MDLC01 and @Andrew15-5 do the new errors look alright?

Andrew15-5:
Yes.

crates/typst-syntax/src/lexer.rs
```diff
@@ -792,7 +792,51 @@ impl Lexer<'_> {
...
+                    '~' if no_hash && self.s.eat_if('=') => {
+                        let error = self.error("`~=` is not valid in code");
+                        self.hint("in Typst, `!=` is used for not-equal");
+                        error
+                    }
+                    '~' if no_hash => {
+                        let error = self.error("the character `~` is not valid in code");
+                        self.hint("in Typst, `not` is used for negation");
+                        error
+                    }
```
MDLC01: AFAIK, `~` is usually for bitwise negation, so suggesting using `not` instead might be a bit misleading.

crates/typst-syntax/src/lexer.rs
```diff
@@ -792,7 +792,51 @@ impl Lexer<'_> {
...
+                        let error = self.error("the character `!` is not valid in code");
+                        self.hint("in Typst, `not` is used for negation");
+                        self.hint("or did you mean to write `!=` for not-equal?");
+                        error
+                    }
+                    '~' if no_hash && self.s.eat_if('=') => {
+                        let error = self.error("`~=` is not valid in code");
+                        self.hint("in Typst, `!=` is used for not-equal");
+                        error
+                    }
```
MDLC01: I'm not aware of a language that uses `~=` for not equal so I'm unsure this hint is really relevant.

MDLC01:
Otherwise, the hints look great!

Andrew15-5:
About `~=`. I know about Lua, but looks like MATLAB also uses it. https://en.wikipedia.org/wiki/Relational_operator#Standard_relational_operators

Though wouldn't say that it's well known or anything.

crates/typst-syntax/src/lexer.rs
bastienvoirin: [Lua uses `~=` for not equal](https://www.lua.org/pil/3.2.html).

isuffix:
I'll remove the hint for a lone tilde, but I'll keep the hint for `~=` given that Lua is a reasonably popular language.

MDLC01 (approved):

tests/suite/syntax/bugs.typ
```diff
@@ -1,14 +1,67 @@
...
+  // Error: 3-5 `~=` is not valid in code
+  // Hint: 3-5 in Typst, `!=` is used for not-equal
+  ~=
+}
+
+--- syntax-bugs-bad-boolean-after-hash paged ---
+// We give different hints than above directly after a hash
+// Error: 2-3 the character `!` is not valid in code
+// Hint: 2-3 try removing the `!`
+#!
```
laurmaedje: based on our recent discussion on having hints primarily as explanation and not giving a suggestion which we're not sure of, this might not be ideal. we don't really know what the user wanted to do and removing the `!` might be a nonsense suggestion if they specifically wanted an exclamation point. in some cases, escaping the hash might be what they want. I'd suggest removing the hint (or thinking through all the potential things a user might have wanted to express).

tests/suite/syntax/bugs.typ
laurmaedje, line 1: didn't notice this before, but the file wasn't really intended for tests like this. it was for parser bugs / crashes with long reproductions found by the fuzzer. maybe just add a new file `suite/syntax/code.typ`. some of the tests would also fit into `suite/syntax/embedded.typ`, but I guess not all.

crates/typst-syntax/src/lexer.rs
```diff
@@ -792,7 +792,46 @@ impl Lexer<'_> {
 
             c if is_id_start(c) => self.ident(start),
 
-            c => self.error(eco_format!("the character `{c}` is not valid in code")),
+            // Everything else is an error, but we try to give good hints for
+            // commonly confusing operators.
+            c => {
+                // Don't give hints like `!` -> `!=` immediately after a hash.
```
laurmaedje: I think all of this would deserve to go into its own function

tests/suite/syntax/bugs.typ
isuffix: I've renamed `bugs.typ` to `panics.typ` containing just the previous test, and I've merged the other tests with the tests in `embedded.typ` into a new `code.typ` file with some improved names.

tests/suite/syntax/bugs.typ
isuffix: We're now giving the following hints for embedded code vs. code-in-code by checking whether the previous character was a hashtag. This isn't a 100% accurate method, but it is easy and the false-positive case should be rare enough that I'm happy with it.
```typ
// Error: 2-3 the character `&` is not valid in code
// Hint: 2-3 the preceding hash is causing this to parse in code mode
// Hint: 2-3 try escaping the preceding hash: `\#`
#&

// Error: 4-5 the character `&` is not valid in code
#{ & }

Here's the bad false-positive case, but it should be uncommon:
// Error: 5-6 the character `&` is not valid in code
// Hint: 5-6 the preceding hash is causing this to parse in code mode
// Hint: 5-6 try escaping the preceding hash: `\#`
#{ #& }
```

I don't like that the span on the hints still points to the current character, but there's no simple way to fix it, and I think its fine.

laurmaedje:
Thanks!

PR #7762, opened 2026-01-26
Mention `std` module in error for unknown variable in math

isuffix:
This was a nice hint I realized we could add for a common error in math. I'm merging it onto the `math-syntax` branch as part of the exhaustive math syntax refactor even though it's self-contained.


crates/typst-library/src/foundations/scope.rs
```diff
@@ -438,8 +438,13 @@ fn unknown_variable_math(var: &str, in_global: bool) -> HintedString {
         ));
     } else if in_global {
         res.hint(eco_format!(
-            "`{var}` is not available directly in math, \
-             try adding a hash before it: `#{var}`",
+            "`{var}` is not available directly in math, but is in the standard library",
+        ));
+        res.hint(eco_format!(
+            "to access `{var}` in math mode you can use the `std` module: `std.{var}`",
+        ));
```
laurmaedje, line 445: maybe we should reverse the order? I'd think that typically definitions that are not available in math mode are more useful through code mode. an exception might be `text`.

isuffix: I swapped the order.

laurmaedje:
actually, why not land this directly in main? is there a particular reason to leave this in the math-syntax branch?

isuffix:
It just has the potential to cause a merge conflict later, but I'm fine with moving it to main as long as we rebase math-syntax occasionally.

laurmaedje:
I'd prefer to move it to main.

Issue #7672, opened 2026-01-10
Typst compiler hang when I try to insert svg


AprilGrimoire:
### Description

```typst
#image("Gamma.svg")
```

<details>
<summary>Debugger output</summary>

```
(base) ➜  typst git:(main) gdb target/debug/typst
GNU gdb (Gentoo 16.3 vanilla) 16.3
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://bugs.gentoo.org/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from target/debug/typst...
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/april/CrossDistrio/april/Working/typst/target/debug/typst.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) r compile ~/Courses/ModuliOfCurves/final-exam-cut.typ
Starting program: /home/april/CrossDistrio/april/Working/typst/target/debug/typst compile ~/Courses/ModuliOfCurves/final-exam-cut.typ
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib64/libthread_db.so.1".
[New Thread 0x7ffff76336c0 (LWP 265228)]
[New Thread 0x7ffff74326c0 (LWP 265229)]
[New Thread 0x7ffff72316c0 (LWP 265230)]
[New Thread 0x7ffff70306c0 (LWP 265231)]
[New Thread 0x7ffff6e2f6c0 (LWP 265232)]
[New Thread 0x7ffff6c2e6c0 (LWP 265233)]
[New Thread 0x7ffff6a2d6c0 (LWP 265234)]
[New Thread 0x7ffff682c6c0 (LWP 265235)]
[New Thread 0x7ffff662b6c0 (LWP 265236)]
[New Thread 0x7ffff642a6c0 (LWP 265237)]
[New Thread 0x7ffff62296c0 (LWP 265238)]
[...]

Issue #7565, opened 2025-12-11
Infinite compile time when including SVG with diacritic


Evidlo:
### Description

I managed to produce an SVG in Inkscape with the string `hello x̂` which causes the compiler to hang if the string uses the CMU Serif typeface:

```
= SVG is below
// #image("test_works.svg")
#image("test_broken.svg")
```

The key seems to be that Inkscape puts the ^ combining character into a separate tag, which Typst doesn't like.

`test_broken.svg` excerpt
``` xml
<tspan
         x="144.58371"
         y="205.5264"
         id="tspan4">
    <tspan
           style="font-style:italic;font-family:'CMU Serif';-inkscape-font-specification:'CMU Serif';stroke:none"
           id="tspan2">hello x</tspan>
    <tspan
           style="font-style:italic;font-family:'CMU Serif';-inkscape-font-specification:'CMU Serif';stroke:none"
           id="tspan3">̂</tspan>
</tspan>

```

[test_works.svg](https://github.com/user-attachments/assets/0f9feb2c-b46d-4896-a1fe-a6160052df6e)
[test_broken.svg](https://github.com/user-attachments/assets/4c4cdf0d-cbea-49b4-82c8-df6ce1c586bb)

### Operating system

Debian

### Typst version

0.14

cady-b:
This is being tracked in #7178. Basically it can't resolve the glyph like that and fallback handling of our usvg dependency has issues.

Enivex:
> This is being tracked in #7178. Basically it can't resolve the glyph like that and fallback handling of our usvg dependency has issues.

It seems to be related, but possibly not entirely the same thing?

cady-b:
~~I'm note sure actually; if the user is experiencing the infinite loop i think it should be that, but I'm also seeing a crash instead.~~

Nvm, I copied the snippet weirdly. Now I'm having it looping and it works if I change out the font to Libertinus Serif.

---

Edit: it works if the `<tspan>` are directly next to each other but not if there is whitespace between them. I think that's actually a pretty unusual sequence (though we shouldn't crash ofc). The OPs actual file does indeed just hang as normal (`test_broken.svg`; the attachment, ***not*** the pretty printed excerpt in the issue description!). Here a sort of MRE:

crash:
```svg
<text style="font-family:'CMU Serif'">
  <tspan>x</tspan> <tspan>̂</tspan>
</text>
```
loop:
```svg
<text style="font-family:'CMU Serif'">
[...]

Issue #7279, opened 2025-10-31
When svg `font-family` is specified to a non-existent one, compiler may compile forever


nik-nul:
### Description

Original svg file will lead to a extremely long compilation time (more than 6min, which is possibly forever):
<img width="827" height="708" alt="Image" src="https://github.com/user-attachments/assets/9357106a-57cc-4444-babd-861fa36647d7" />

While if I replace the specified `font-family` name from `Microsoft Yahei` (a font does not exist on my machine) to `Noto Sans` (a font that does exist), the compilation will end instantly:

<img width="827" height="731" alt="Image" src="https://github.com/user-attachments/assets/c6270480-3345-4470-9c83-f78b32af333f" />

You may test with this svg: [https://github.com/user-attachments/assets/4ee92706-2433-43ce-8dea-34451a69a71f](https://github.com/user-attachments/assets/4ee92706-2433-43ce-8dea-34451a69a71f), it is exactly `forever.svg` in image 1.

OS: 
```
Linux x1c6 6.1.0-40-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.153-1 (2025-09-20) x86_64 GNU/Linux
```

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

laurmaedje:
Possibly related to https://github.com/typst/typst/issues/7178?

laurmaedje:
I could not reproduce this because I didn't know which fonts to install, but it's likely fixed by https://github.com/typst/typst/pull/7766. If it's not fixed, please let me know.

Issue #7178, opened 2025-10-17
Deadlock on Excalidraw's SVG


michalfita:
### Description

This code:
```typst
#import "@preview/typslides:1.3.0": *

// Project configuration
#show: typslides.with(
  ratio: "16-9",
  theme: "purply",
  font: "Excalifont",
  font-size: 18pt,
  link-style: "color",
  show-progress: true,
)

#slide(title: "Original process", outlined: true)[
  #image("odk-original-process.excalidraw.svg")
]
```
with this SVG image
![Image](https://github.com/user-attachments/assets/73671910-0471-4b09-856f-5b5a93103314)
sends CLI `typst` into spiral of deadlock on Ubuntu 24.04.

I'm using Typst 0.13.1 installed via `cargo install`. It deadlocks VS Code's `tinymist` extension as well.

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

Enivex:
~~Obviously it shouldn't crash, but if I recall correctly Excalidraw SVGs won't work anyway. They're created in a way that's only compatible with a browser.~~

Edit: Apparently incorrect

michalfita:
Line from `top` after a few minutes of running w/o any message on the output:
```
2099404 user   20   0 1170480  46164  20432 R 100,0   0,2   7:05.32 typst
```
Other SVG files exported from Excalidraw work fine.

Enivex:
Have you tried using the development version?

LaurenzV:
I cannot reproduce this on the webapp at least.

michalfita:
Now I wonder if font matters...

michalfita:
typst 0.13.1 (87f7a731) deadlocks as well on this example.

Changing font doesn't matter.

LaurenzV:
If possible to might help to remove parts of the SVG as much as possible to see where exactly it trips up.

michalfita:
Writing a SVG bisector sounds like a interesting challenge, I'm afraid I don't have capacity to do so. This happened while working on presentation for job interview on Monday.

LaurenzV:
No worries! Maybe someone else can reproduce the issue.

michalfita:
In the act of _desperation_ I narrowed it down to 🤣 

<details><summary>Narrowing details</summary>
<p>
`µ` (U+00B5) character.

Commenting it out in the SVG stops deadlocking `typst` and it renders everything as it's supposed to.
</p>
</details>

[...]

PR #7766, opened 2026-01-26
Fix deadlock with font fallback in SVGs

laurmaedje:
Confirmed fixed, could reproduce:
- Fixes https://github.com/typst/typst/issues/7178
- Fixes https://github.com/typst/typst/issues/7565

Very likely fixed[^1], but could not reproduce due to unknown required set of fonts:
- Closes https://github.com/typst/typst/issues/7279
- Closes https://github.com/typst/typst/issues/7672

[^1]: Since these issues are very likely to be the same, I'll also close them. If they end up being different, we can still reopen them.


laurmaedje:
Looks like GitHub actions is having issues ...

cady-b:
:partying_face: this is incredible; so happy these are finally fixed!

Issue #7045, opened 2025-10-07
Document the embedded New Computer Modern does not have NewCM10-Book (wght=450) variants


YDX-2147483647:
### Description

It looks like that the embedded font _New Computer Modern_ is not as comprehensive as that shipped with TeX Live.
Specifically, typst-asset only contains NewCM10-Regular (wght=400), but not NewCM10-Book (wght=450). The latter is the default font for `\usepackage{newcomputermodern}` in LaTeX.
For special characters like U+00B7 MIDDLE DOT `·`, the difference is quite visible.

Please refer to this topic on the forum: [Why does a centered dot render with different spacing in markup? - Questions - Typst Forum](https://forum.typst.app/t/why-does-a-centered-dot-render-with-different-spacing-in-markup/6292)

### Where to document it

- https://typst.app/docs/reference/text/text/#parameters-font
- https://typst.app/docs/guides/guide-for-latex-users/#latex-look

I suggest adding to either of the above sections, or including it in a new Font Setup Guide.

### Links

- I suggest not rushing to document this details, as #6466 might affect the wording.

- This issue also affects #6774 in a certain sense.

Enivex:
I don't quite see why this needs to be documented?

YDX-2147483647:
Because it's said that [_New Computer Modern_ is embedded in the CLI](https://typst.app/docs/reference/text/text/#parameters-font:~:text=are%20Libertinus%20Serif%2C-,New%20Computer%20Modern,-%2C%20New%20Computer), but what's not said is that it's only a partial version.

In fact, I've also encountered another strange problem where NCM Math of typst and matplotlib were inconsistent (although I didn't have time to record it then).
I hope that the specific version of fonts embedded in typst could be better documented.

Enivex:
It's not a partial version, just not all the weights

YDX-2147483647:
> not all the weights

Yes, that's what I mean.

sijow:
The Book weight is also the default now when doing `\usepackage[default]{fontsetup}`. It would probably make sense to include it in the embedded fonts (that would add about 1MB to have both normal and italic styles).

MDLC01:
I am wondering whether it would be better to simply embed all variants of NewCM in the executable. It is kind of unexpected for the user that only some variants of a family are included. For reference, the full NewCM family is about 21.3 Mo in size and the Typst 0.14.2 executable is around 46.2 Mo.

[...]

PR #1774, opened 2023-07-21
Ensure normal weight math font face is consistent

damaxwell:
By default, Typst uses New Computer Modern Math at weight 450 for its math font.  This results in NewComputerModernMath-Book as the default font.  However, sometimes NewComputerModernMath-Regular (font weight 400) sneaks in.
```
$ log x (log x)/ y $
```
image

There are two fonts in the file:
```
name                                 type              encoding         emb sub uni object ID
------------------------------------ ----------------- ---------------- --- --- --- ---------
ABCDEF+NewCMMath-Book                CID Type 0C (OT)  Identity-H       yes yes yes      3  0
ABCDEF+NewCMMath-Regular             CID Type 0C (OT)  Identity-H       yes yes yes      8  0
```
The `log` in the beginning is Book weight, and the `log` in the numerator is Regular weight (look at the thinnest spots on the `o`s and the upper-right decorations on the `g`s).  Both `x`s, on the other hand, are Book weight; this isn't a design choice that stuff in denominators should have lesser weight.  Instead, there are currently multiple pathways for rendering text in math, and one of them is accidentally overriding the initial weight of 450.

This PR ensures that the meaning of "normal" font weight is tied to the font weight initially selected for the math font.

A number of reference test images needed to be updated.  The only changes were that text in some circumstances became a little heavier.

## After
image

```
name                                 type              encoding         emb sub uni object ID
------------------------------------ ----------------- ---------------- --- --- --- ---------
ABCDEF+NewCMMath-Book                CID Type 0C (OT)  Identity-H       yes yes yes      3  0
```

PR #7771, opened 2026-01-27
Mention font weight in the section "How do I get the 'LaTeX look?'" in guide for LaTeX users

jhvoigts:
Typst's default font weight in math mode is set to `450` instead of `400` (or `regular`) which is used for normal text. LaTeX uses the lighter font weight by default. So, to achieve the "LaTeX look", it needs to be reduced. This PR adds a respective note to the guide for LaTeX users.

Besides that, the fact that in math mode, the font weight is different, is not documented. Maybe, this should be made explicit in the docs.

### DIFF
--- a/docs/guides/guide-for-latex-users.md
+++ b/docs/guides/guide-for-latex-users.md
@@ -627,6 +627,7 @@ The example below
   [first-line-indent]($par.first-line-indent)
 - [sets the font]($text.font) to "New Computer Modern", an OpenType derivative of
   Computer Modern for both text and [code blocks]($raw)
+- decreases the [font weight]($text.weight) in math mode
 - disables paragraph [spacing]($block.spacing)
 - increases [spacing]($block.spacing) around [headings]($heading)
 
@@ -635,6 +636,7 @@ The example below
 #set par(leading: 0.55em, spacing: 0.55em, first-line-indent: 1.8em, justify: true)
 #set text(font: "New Computer Modern")
 #show raw: set text(font: "New Computer Modern Mono")
+#show math.equation: set text(weight: "regular")
 #show heading: set block(above: 1.4em, below: 1em)
 ```
 


YDX-2147483647:
Double-check is needed here… See #7045.

- 450 is the default when doing `\usepackage{newcomputermodern}`, but I can't remember whether it applies to text or text+math.

- According to [Jeremie Knuesel's reply](https://github.com/typst/typst/issues/7045#issuecomment-3381808634), 450 is also the default when doing `\usepackage[default]{fontsetup}`. ~~I have't check it, though.~~

  <details><summary>Confirmed</summary>
  
  [`newcm-doc.pdf` §2](https://mirrors.ctan.org/fonts/newcomputermodern/doc/newcm-doc.pdf#section.2)
  
	> **2 How to load the fonts**
	>
	> The simpler way to load the fonts is through the `fontsetup` package. The command
	>
	> ```latex
	> \usepackage[default]{fontsetup}
	> ```
	>
	> will load the Book weight of the NewCM family,
	>
	> ```latex
	> \usepackage[olddefault]{fontsetup}
	> ```
	>
	> will load the Regular weight, and
	>
	> ```latex
	> \usepackage[sansdefault]{fontsetup}
	> ```
	>
	> will load the Sans Serif NewCM family.
	
	Book weight = 450, Regular weight = 400.

  </details>

- _Computer Modern_ is the default font for LaTeX standard classes without special packages. The difference between _Computer Modern_ and _<ins>New</ins> Computer Modern_ might be more significant than  _New Computer Modern_ 400/450.

---

#### Edit

The 450/400 problem was also mentioned before in #1774.

Git blame tells that 450 dates back to 2023-02-26 https://github.com/typst/typst/commit/7597b997890647aa0546ec7bfef1eae4e33cedcb.
> New default style

[February 25, 2023 — Changelog](https://typst.app/docs/changelog/earlier/#february-25-2023):
> New default font for math: Book weight of New Computer Modern Math

jhvoigts:
Indeed, you are right, I just tested both `fontsetup` and loading `newcomputermodern` and they result in the higher font weight. I was unaware that, as I never came across these before.

As far as I can tell, math equations with `#show math.equation: set text(weight: "regular")` look identical to LaTeX's default Computer Modern (ref https://discord.com/channels/1054443721975922748/1465808867068739585).

As New Competer Modern with weight 400 seems to mimic LaTeX's default, I think it is appropriate to recommend this setting to achieve the "LaTeX look".

docs/guides/guide-for-latex-users.md
```diff
@@ -627,6 +627,7 @@ The example below
   [first-line-indent]($par.first-line-indent)
 - [sets the font]($text.font) to "New Computer Modern", an OpenType derivative of
   Computer Modern for both text and [code blocks]($raw)
+- decreases [font weight]($text.weight) in math mode
```
laurmaedje:
```suggestion
- decreases the [font weight]($text.weight) in math mode
```

laurmaedje:
Thanks!

Issue #5982, opened 2025-03-01
`stroke` documentation does not provide definitions for dash pattern shorthands


wisp3rwind:
### Description

The [documentation for `stroke`](https://typst.app/docs/reference/visualize/stroke/#constructor-dash) mentions the several predefined patterns for `dash`, i.e.

- "solid"	
- "dotted"	
- "densely-dotted"	
- "loosely-dotted"	
- "dashed"	
- "densely-dashed"	
- "loosely-dashed"	
- "dash-dotted"	
- "densely-dash-dotted"	
- "loosely-dash-dotted"

but no details about them. I think it would be useful if it additionally included

- an example of each
- their definition, as a starting point for modified dash patterns, i.e. I would appreciate if the docs showed that `densely-dotted` is `("dot", 1pt)` and so on.

This is a follow-up to [this discussion](https://forum.typst.app/t/custom-underline-strokes-based-on-default-stroke-thickness/3042) in the forum.

YDX-2147483647:
> * their definition, as a starting point for modified dash patterns, i.e. I would appreciate if the docs showed that `densely-dotted` is `("dot", 1pt)` and so on.

I think documenting how to get it programmatically would suffice, similar to paper sizes (https://github.com/typst/typst/issues/3361#issuecomment-3113093105).

```typst
#stroke(dash: "loosely-dashed").dash
```

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/8e51625c-154a-45d4-bc6b-236e0e2494c3" />

wisp3rwind:
> I think documenting how to get it programmatically would suffice, similar to paper sizes ([#3361 (comment)](https://github.com/typst/typst/issues/3361#issuecomment-3113093105)).
> 
> #stroke(dash: "loosely-dashed").dash

Thanks for looking into this! Pointing out introspection is definitely a big help already (I didn't think of that possibility when I had the question originally). I think that personally, I'd still prefer having the explicit values in the docs, saving the need to manually inspect the resolved `dash`.

YDX-2147483647:
Okay, but #7772 cannot cover it, because I heard that Typst staffs are working on refactoring the infrastructure for Typst (#6704), so doing it now will lead to conflicts.

PR #7772, opened 2026-01-28
Illustrate predefined patterns for `stroke.dash` in docs

YDX-2147483647:
Resolves part of #5982

- Document how to inspect a predefined pattern
- Display predefined patterns for stroke.dash in docs

图片

Inspired by [Table: Colors in Northern Kurdish - Wiktionary](https://en.wiktionary.org/wiki/pirteqal%C3%AE#See_also) and [15.3.2 Graphic Parameters: Dash Pattern - TikZ & PGF Manual v3.1.11a](https://texdoc.org/serve/tikz/0#subsubsection.15.3.2).

### DIFF
--- a/crates/typst-library/src/visualize/stroke.rs
+++ b/crates/typst-library/src/visualize/stroke.rs
@@ -160,7 +160,6 @@ impl Stroke {
         /// ```
         ///
         /// ```preview:"Predefined patterns"
-        ///
         /// #set page(width: auto, height: auto)
         ///
         /// #let patterns = (


cady-b:
I think the teal (`eastern`) doesn't pass WCAG AA for body text on white like here. Otherwise this would be great to have in the docs :+1:

YDX-2147483647:
Indeed! It should pass now.

图片

[WebAIM: Contrast Checker (1D8290 vs. FFFFFF)](https://webaim.org/resources/contrastchecker/?fcolor=1D8290&bcolor=FFFFFF)

laurmaedje:
Thanks!

PR #7774, opened 2026-01-28
Update Vietnamese translations

HSGamer:
This completes the missing fields of the translation.

### DIFF
--- a/crates/typst-library/translations/vi.txt
+++ b/crates/typst-library/translations/vi.txt
@@ -4,7 +4,8 @@ equation = Phương trình
 bibliography = Tài liệu tham khảo
 heading = Phần
 outline = Mục lục
-raw = Đoạn mã
+# may be wrong
+raw = Chương trình
 page = trang
 footnote = Chú thích
 email = Email


crates/typst-library/translations/vi.txt
```diff
@@ -4,9 +4,8 @@ equation = Phương trình
 bibliography = Tài liệu tham khảo
 heading = Phần
 outline = Mục lục
-# may be wrong
-raw = Chương trình
+raw = Chỉ mục
```
laurmaedje: could you provide a source/reference for this change?

HSGamer: I cannot find a source for this. But since raw is usually used to write part of code, "Đoạn mã" is better in this case. "Chương trình" feels like we are writing a full app. 
This is specifically for programming terms. Not sure about general terms since we use different words for different cases.

laurmaedje: The term used here should specifically be for code or data listings in technical works, hence in English "Listing" rather than "Code" or "Program". Without any source, I would rather not change this (potentially back and forth). Are there any Vietnamese publications of record containing such listings that could be used as a reference in place of a proper style guide?

HSGamer: It's almost impossible to find a Vietnamese publication that use the exact word for that case. And there is no exact guide on what exact word to use in that case. 
Like I said, we here usually use different words for different cases. For example, the code to demonstrate an algorithm is called "Giải thuật", the code as a part of a function of an app is called "Đoạn mã",the code to demonstrate a data structure is called "Cấu trúc dữ liệu". There is no general way to translate a "Listing" to something that matches all the cases. 
We probably have to use `figure-supplement` for a lot of cases here and there so translating this doesn't mean anything.
We don't even add the whole code to our papers, instead we draw flow diagrams and make pseudo-code for specific algorithms. 
If you are reluctant to change this, then it's fine I can revert it back. But in my opinion, it's definitely not "Chương trình", which means "program".

laurmaedje: I'm not relucant to change it per se, in particular if the current wording is unusual. I just need something to base my decision on, as I can't judge it myself, and anybody can open a PR.

HSGamer: Neither do I. There is no style guidelines for that specific word. 
Anyways, I reverted it since there is no point to discuss as both I and you don't have a reference to base on.

HSGamer: It still doesn't solve the main problem though.
If you ask for official guidelines, well, there is none here. 
But judging by the usual usage of the code/raw block you mentioned, I would prefer "Đoạn mã" (part of code) or "Mã nguồn" (source code) than "Chương trình" (full program/application).

laurmaedje: I'm not asking for official guidelines, just links to some papers or publications that contains such listings, to see the wording used in practice. If you say "usual usage", there's gotta be something that can be used as a reasonable precedent. Of course, the higher-profile the publication, the better, but I only wanted to see _usage_ not any official guidance.

HSGamer: I guess it's a deal breaker then. I'm just a student who is about to make a thesis. The guidelines of my university limits the papers to be English, and most of the sources I usually use are in English, so for me this PR is meaningless. I don't have access to any high-profile publication since it's usually behind a paywall, and obviously I don't want to buy a subscription to any publication in Viet Nam just to prove a point here.
Like I said before, there is no obvious word for the usage of "listing" you mentioned. We would probably use different words for different context. Like "thuật toán" ([Source](https://ctujsvn.ctu.edu.vn/index.php/ctujsvn/article/view/3319/500)), "giải thuật" ([Source](https://gwdu.ptit.edu.vn/slink/file/68906db4c7ff8fe064cc8091/De%20an%20tot%20nghiep_PhuongHV_Update%20chinh%20sua.pdf)), and more that I cannot access the publication. Some may just capture the screenshot of the code and called it an image or "Hình" ([Source](https://dlib.ptit.edu.vn/bitstream/HVCNBCVT/3649/1/Do%20an%20Hoang%20Anh%20Duc%20D19%20CNPM.pdf)). We would probably use `figure-supplement` here and there for those different contexts.
I'm here just as a native Vietnamese, who have thoughts to try correcting the words. If you want to have any example of any publication before checking the PR, I will leave this PR to someone who have more access to those paywalled publication instead.
I revert the `raw` translation to the old one since it looks like I cannot proof that. And well, If you ask for proof for the remaining lines, I will close the PR instead.

laurmaedje:
Thanks.

PR #7781, opened 2026-01-30
Explain the differences between `attach()` and `accent()`

### LINKED ISSUE #7650: The `attach()` page should mention the `accent()` page
### Description

When one tries to typeset an accent but does not know how, they might stumble into the `attach()` function documentation, think that is the correct way to write an accent and encounter what looks like bugs. It would be nice if the `attach()` function documentation mentioned that the correct way to typeset an accent is to use the `accent()` function, so that this situation could be avoided.

lltsdyp:
Closes #7650 .


laurmaedje:
Thanks!

Issue #6735, opened 2025-08-09
U+2E3A should not be overhung


chiefaccelerators:
### Description

To my knowledge, U+2E3A is used in Chinese typesetting only, which does not have punctuation overhang styling right now (#6582), so the hanging of this particular punctuation is really confusing. Plus, even if we bring it back, U+2E3A should not be overhung. Per [clreq](https://w3c.github.io/clreq/#hanging_punctuation_marks_at_line_end), hanging punctuation "helps to avoid moving characters or punctuation marks between lines and avoids inconsistency of space between the characters in different lines." The current hanging effect of U+2E3A defeats that purpose:

<img width="749" height="211" alt="Image" src="https://github.com/user-attachments/assets/64c6a9e5-43e9-4479-a583-ef849e3d7e05" />

Typst code:
```
#let ccwd = 12pt

// Set the text width to exact 30 Chinese characters.
#set page("a5", margin: (x: 50% - ccwd * 15))

#set text(
  lang: "zh",
  font: ("Source Han Serif SC"), // font converts 2 U+2014s to U+2E3A
  size: ccwd,
)

#set par(justify: true, first-line-indent: (amount: 2em, all: true))

// stripe background
#place(
  top + left,
	dx: ccwd,
  repeat(
      rect(width: ccwd, height: 6em, stroke: none, fill: rgb("#7092be70")),
			gap: ccwd,
			justify: false
    )
  )

#set text(overhang: false)
囗囗囗囗囗囗囗，囗囗囗囗囗囗囗囗。囗囗囗囗囗囗囗囗囗——囗囗囗囗 this is correct

#set text(overhang: true)
囗囗囗囗囗囗囗，囗囗囗囗囗囗囗囗。囗囗囗囗囗囗囗囗囗——囗囗囗囗 the grid is broken :(
```

### Operating system

macOS

### Typst version

- [x] I am using the latest version of Typst

YDX-2147483647:
This issue does not occur if we write the _real_ U+2E3A TWO-EM DASH (`⸺`).
It only happens for two consecutive U+2014 EM DASH (`——`).

> `font: ("Source Han Serif SC"), // font converts 2 U+2014s to U+2E3A`

According to the [official font readme `SourceHanSerifReadMe.pdf`](https://github.com/adobe-fonts/source-han-serif/raw/release/SourceHanSerifReadMe.pdf), this is implemented using the OpenType GSUB feature.

[...]

PR #7782, opened 2026-01-30
Configurable character protrusion for both margins

### LINKED ISSUE #261: Line layout, protrusion
I'm new to typst, so forgive me if I'm missing something.

Could you add more advanced, microtype/InDesign-like, things like protrusion?

Using Minion Pro, `typst`:

typst-prot

Here's lualatex with microtype:

tex-protrusion

### LINKED ISSUE #6582: Chinese punctuation overhang
### Description

Hi, I'm trying to enable punctuation overhang for Chinese punctuation. I noticed that this feature was already there but it was removed by #1211.

It was quoted that

> Most Chinese publications do not use hanging punctuation at line end.

But this is more of a stylish choice, not a rule, and I definitely see publications using overhang punctuation, especially in the indie publishing scene. InDesign also provides this option (Apply kinsoku hanging): https://helpx.adobe.com/indesign/using/composing-cjk-characters.html

Do you think it would be possible to provide some customization options for which punctuation should be used for hanging?

### Use Case

See for example the punctuation in [政见合辑](http://cnpolitics.github.io/compilation/pdf/cnpolitics-compilation-2014-s3-s4.pdf)

政见合辑

[Han.css](https://hanzi.pro/) also enables punctuation hanging by default for simplified Chinese, but not traditional Chinese.

Han

### LINKED ISSUE #7231: Support hanging-punctuation in the left margin
### Description

Punctuation such as quotation marks should hang into the left margin. (See _Thinking with Type_, Ellen Lupton, 3rd edition, p. 76) The `microtype` package does this in LaTeX.

Here is the current behavior:

Image

Here is what it looks like with hanging punctuation:

Image

### Use Case

This would be useful because hanging punctuation is good in both the left and the right margins. :)

-------------------------

I am sorry if I am submitting this issue to the wrong place—I'm new here. Please point me in the right direction if this is the wrong place. How new I am: I tried out Typst yesterday, fell in love with it, and started work on this feature shortly thereafter. I actually have a mock-up [in my fork](https://github.com/ashton314/typst). It took very little effort to get a hacky version working.

Please let me know where to go with this feature from here. It's working for me in this one little case. If there are tests I need to add, documentation, switches, etc. please let me know and give me a nudge in the right direction.

sicikh:
~~Depends on #8131.~~

Closes #261, closes #6582, closes #7231.

This PR changes the behaviour of the `text.overhang` property, such that:

1. Left and right margins can be used for overhang characters for LTR and RTL languages accordingly.

2. Users can now specify where overhang characters should happen:

    - enable or disable completely (`true` or `false`), 
    - by side (`left` or `right`),
    - by direction (`start` or `end`),
    - by individual _glyphs:_ by UTF-8 character or by glyph's name. The later is needed, for example, to reference ligatures or specific glyphs behind substitution tables, not available otherwise.

4. Old values `overhang: false` and `overhang: true` stay, but `true` changes its meaning from "enable default overhang punctuation only for the end margin" to "enable default protrusions for both margins".

    The default value changes from `true` to `end`, that works as previous `true`.  

5. Most beneficial result is that _overhang punctuation is now considered when paragraph is being laid out,_ what leads to better linebreaks and hyphenation. Now documents, that were using default values (and that will continue to use them), will benefit from this PR without additional changes from their side.

For the results see below:

image

<details>
<summary>
Code
</summary>

```typ
#set page(paper: "a4")

// Wiki suggested values + Hàn Thế Thành’s original settings
//
// - https://en.wikipedia.org/wiki/Optical_margin_alignment#
// - https://ftp.snt.utwente.nl/pub/software/tex/macros/latex/contrib/microtype/microtype-code.pdf, p. 155
#let protrusion-table = (
  "A": (20%, 20%),
  "F": (0%, 5%),
  "J": (5%, 0%),
  "K": (0%, 5%),
  "L": (0%, 5%),
  "C": (10%, 0%),
  "O": (10%, 10%),
  "T": (20%, 20%),
  "V": (20%, 20%),
  "W": (20%, 20%),
  "X": (5%, 5%),
  "Y": (20%, 20%),
  "k": (0%, 5%),
  "r": (0%, 5%),
  "t": (0%, 10%),
  "v": (5%, 5%),
  "w": (5%, 5%),
  "x": (5%, 5%),
  "y": (10%, 10%),
  "c": (10%, 10%),
  "o": (10%, 10%),
  ".": (0%, 100%),
  ",": (0%, 100%),
  ":": (0%, 100%),
  ";": (0%, 100%),
  "!": (0%, 20%),
  "?": (0%, 20%),
  "(": (20%, 0%),
  ")": (0%, 20%),
  "-": (0%, 75%),
  "\u{ad}": (0%, 75%),
  "–": (0%, 50%),
  "—": (0%, 25%),
  "“": (100%, 0%),
  "”": (0%, 100%),
)
#let italic-protrusion-table = (
  "p": (20%, 0%),
)
#let use-another-table-for-emph(body) = {
  show emph: set text(overhang: (map: italic-protrusion-table, default: false))
  body
} 

#let example(name) = [
  #place(top, float: true, scope: "parent", strong(name))
  
Shortly after this, when the sexton
came to pay them a visit, the father
broke out to him, and told him what
a bad hand his youngest son was
at everything: he knew nothing and
learned nothing. “Only think! when
I asked him how he purposed gaining a livelihood, he actually asked
to be taught to shudder.” “If that’s
all he wants,” said the sexton, “I can
teach him that; just you send him to
me, I’ll soon polish him up.” The father was quite pleased with the proposal, because he thought: “It will
be a good discipline for the youth.”
And so the sexton took him into his
house, and his duty was to toll the
bell. After a few days he woke him
at midnight, and bade him rise and
climb into the tower and toll. “Now,
my friend, I’ll teach you to shudder,”
thought he. He stole forth secretly
in front, and when the youth was
up above, and had turned round to
grasp the bell-rope, he saw, standing _opposite_ the hole of the belfry,
a white figure. “Who’s there?” he
called out, but the figure gave no answer, and neither stirred nor moved.
“Answer,” cried the youth, “or begone; you have no business here
at this hour of the night.” But the
sexton remained motionless, so that
the youth might think that it was a
ghost. The youth called out the second time: “What do you want here?
Speak if you are an honest fellow, or
]

#set page(width: 500pt, height: 21em, margin: 15pt)
#set par(justify: true)
#set text(size: 0.8em)

#grid(
  columns: (1fr, 1fr, 1fr, 1fr),
  gutter: 10pt,
  {
    set text(overhang: false)
    example[No protruding]
  },
  {
    set text(overhang: (map: protrusion-table, default: false))
    show: use-another-table-for-emph
    example[Character protruding]
  },
  {
    set text(overhang: false)
    set par(justification-limits: (tracking: (max: 0.02em, min: -0.02em)))
    example[Character level justification]
  },
  {
    set text(overhang: (map: protrusion-table, default: false))
    set par(justification-limits: (tracking: (max: 0.02em, min: -0.02em)))
    show: use-another-table-for-emph
    example[Protruding + justification]
  }
)
```

</details>

## API

```typ
// Disable overhang completely.
#set text(overhang: false)

// Enable overhang for both margins with default protrusions.
#set text(overhang: true)

// Enable overhang only for one margin:
#set text(overhang: left)
#set text(overhang: right)
#set text(overhang: start) // will be left margin for LTR text
#set text(overhang: end) // will be right margin for LTR text, also this is the new default value

// Enable overhang _per glyph_.

#let protrusion-table = (
  // (left, right) protrusions as `ratios-to-width`% + `relative-to-font`em.
  "«": (100%, 100%),
  "»": (100%, 100%),
  "”": (100%, 100%),
  "“": (100%, 100%),
  "A": (10%, 0%),
  "C": (1% + 0.001em, 0%)
  // If we would like to customize only one side,
  // and rely on default algorithm (more on that later) for another,
  // `auto` value is used.
  "-": (100%, auto),
  // If glyph in a font can not be accessed by UTF-8 character,
  // then we can specify exact name of the glyph specific to that font via "/glyph-name"
  // (like in "microtype").
  "/f_f_i": (10%, 10%),
)

// Note that this protrusion table can be shipped by third-party package
// for specific font and its style (this is what LaTeX package "microtype" does).

// `text.overhang` also can receive dictionary with two optional values:
// `map` - the table from character to protrusions for both sides,
// `default` - the behaviour for characters, that are not in the table.
//             Receives values, as `overhang` itself (false, true, left, right, ...)
//             If omitted, results in `false` 
//             (the protrusions will be read only from the specified table).
#set text(overhang: (map: protrusion-table, default: right)))
```

## Rationale

As written in https://github.com/typst/typst/issues/7231#issuecomment-3510208728, the main problem for this functionality is the API. https://github.com/typst/typst/issues/7231#issuecomment-3456621718 proposes some kind of automatic calculation for the amount of protrusion for characters.

Main argument against _disabling_ the possibility for customizing protrusions is that such algorithm, that can be implemented in the future, may yield suboptimal results. In some circumstances users may want to customize behaviour per character or per margin basis (see https://github.com/typst/typst/issues/7231#issuecomment-3507224012 and #6582).

Thus, the API should take into account, that such algorithm for calculating protrusions by default may be implemented later, but also should give users a possibility to directly adjust values.

Another option was considered to rather recieve function `(char, side) => protrusion`, but it has some technical limitations (see https://github.com/typst/typst/pull/7782#issuecomment-3832935206). Also manipulating dictionaries from the user-side is better than with the black-box functions (think of possible typst's "microtype" package with those protrusion tables). More importantly, is that compiler needs to see _keys of the table_ for identifying glyphs ids (see #8131).

With all of this, the design space is limited only by such protrusion table.

## TODO

- ~~Improve API such that table may contain keys for the characters, consisting of multiple UTF-8 characters (see https://github.com/typst/typst/pull/7782#issuecomment-3833147301), while disallowing incorrect usage.~~ Not needed anymore, since #8131 fixes overhang glyphs that are shaped from multiple characters.

    ~~The only case when we can not refer to a font's glyph via UTF-8 characters are ligatures. We should decide how to refer to them (see: https://tex.stackexchange.com/questions/235994/microtype-customizing-protrusion-for-ellipsis-and-multiplication-symbols/236661#236661 for how `microtype` does it).~~ Done in 10edfe9c.

- Improve error handling (nested dictionaries have bad ergonomics...).

- Design and implement folding for `text.overhang`.


terefang:
i would propose it differently:

lets keep an option that make simple things simple but still allows complex behavior:
```
#set text(overhang: bool | string | dict)
```

### for "bool" case

* false – simply disables overhang behavior and should be the default
* true – enables overhang behavior with whatever internal default can be agreed upon

this is actually a weak case and could be dropped entirely by using the case below.

### for the string/enum case

simple "left", "right", "both", "default", and "none" overhang layout with whatever characters agreed as the default.

### for the dict case

this follows the proposal of the OP but more refined:

```
#set text(overhang: (
  map: (
    "«": (100%, 100%),
    "»": (100%, 100%),
    "”": (100%, 100%),
    "“": (100%, 100%),
    "A": (10%, false),
  // customize protrusion only for the left margin
    "-": (100%, true),
  ),
  // regex match against characters.
  match: (
    "regex1": (100%, 100%),
    "regex1": (10%, false),
  ),
  // use built-in values for other characters 
  // (could be "left", "right", "both", "default", and "none") 
  rest: "right",
  // if omitted, results in `none`
))
```

what do you think ?

sicikh:
Thanks for the suggestions! My thoughts on this:

1. Why would we need `"default"` value? I don't think that is the right path — the more correct one will be using other kind of values as default (particularly, `right`/`"right"` or `true`/`"both"`).

2. After removing `"default"` from your design, for the string case we have only `"left"`, `"right"`, `"both"` and `"none"`. Initially I also was thinking about this, but then I came to this logic:

    - Users alredy may have `overhang: true` or `overhang: false` in their code. Thus, new `overhang` values must accept `bool`.

    - If `overhang` can accept `bool`, what these values will mean? `false`, obviously, should just disable overhang completely (yours `"none"`). Then `true` should mean "enable (full) overhang", that is yours `"both"`.

    - Then only `"left"` and `"right"` values remain for the string case. Why not just use alignment type, as done in `string.trim(at: start/end)`? 

    - As `true` changed its meaning from "enable (right) overhang" to "enable (full) overhang", the default value _may_ become `right`.

3. The proposal for the dict case is quite interesting, I also had some thoughts in this direction. But what do you think about the usecases for the regex matching?

4. After a day, I think that individual protrusions should receive `auto` (use default algorithm/table, does not matter) or ratios. Then we don't have two options for disabling overhanging with either `0%` and `false`, only the first one remains.

With this, the API for the dict case may look like the following:

```typst
#set text(overhang: (
  map: (
    "«": (100%, 100%),
    "»": (100%, 100%),
    "”": (100%, 100%),
    "“": (100%, 100%),
    // Enable overhang only for the left margin.
    "A": (10%, 0%),
    // Customize protrusion only for the left margin,
    // use default values for the right one.
    "-": (100%, auto),
  ),
  // Use built-in values for other characters on the right margin
  // (could be `left`, `right`, `true` or `false`).
  // If omitted, results in `false`.
  rest: right,
))
```

laurmaedje:
I'm not yet ready to comment on this at large, but still wanted to add three points to the discussion:

- We don't necessarily have to have `true` / `false` in the primary API. They could remain as deprecated aliases for something like `none` / `auto`.
- Stringly-typed values should be avoided if better values exist (like `left`, `right`, `none`, or `auto`)
- Rather than supporting some complex ad-hoc dictionary structure, we might want to just support passing a function that receives the char.

eltos:
You could use the **alignment** type, i.e. `left + right` instead of "both"/true, and `start`/`end`/`start + end` as a text-direction adaptive setting.

laurmaedje:
unfortunately, adding `left + right` does not work. you can only add alignments for different axes. the same thing came up for `block.sticky` where there was discussion of having `sticky: (top, bottom)` to have it sticky on both sides.

sicikh:
> * We don't necessarily have to have `true` / `false` in the primary API. They could remain as deprecated aliases for something like `none` / `auto`.

I think that we should not deprecate `bool` value in the `overhang` option: `overhang: false` disables overhang, `overhang: true` enables overhang for left and right margins.

> * Rather than supporting some complex ad-hoc dictionary structure, we might want to just support passing a function that receives the char.

That is quite an interesting idea! The question now what signature should this function have. I see two options:

1. Function receives character and side of the margin as `alignment`, returns `ratio` or `auto`:

    ```typ
    #let protrusion-table = (
      "«": (100%, 100%),
      "»": (100%, 100%),
      "”": (100%, 100%),
      "“": (100%, 100%),
      "A": (10%, 0%),
      "-": (100%, auto),
    )

    // text.overhang: bool | alignment | (str, alignment) => (auto | ratio)  
    #set text(overhang: (char, side) => 
      protrusion-table
        // enable default algorithm for the right margin, disable overhang for the left one 
        .at(char, default: (0%, auto))
        .at(if side == left { 0 } else { 1 })
    )
    ```

2. Function receives only character, returns an array of two protrusions (`ratio` or `auto`):

    ```typ
    // text.overhang: bool | alignment | str => [auto | ratio; 2]  
    #set text(overhang: char => 
      protrusion-table
        .at(char, default: (0%, auto))
    )
    ```
    
Currently, I'm leaning more towards the first option.

sicikh:
The situation is as follows.
    
1. Now `overhang` receives function instead of dictionary, and this is better approach. Currently, its signature is `(str, alignment) => (auto | ratio)`, but it can be changed easily. The `alignment` here is either `left` or `right`.

2. Previous changes clearly broke tests for RTL languages. The previous behaviour was to overhang punctuation _on the end margin_, not only on the right one.

    So, `overhang` can not receive just `left` or `right`, because this can not replicate previous behaviour — it must also take `start` and `end` values. Now the default value is `overhang: end` and all tests pass.
    
    I'm thinking, do we currently really need `left` and `right` values in these circumstances...

sicikh:
As I read in [Hàn Thế Thành's dissertation](https://www.tug.org/TUGboat/tb21-4/tb69thanh.pdf), this PR implements so-called "level 1 character protruding" (pp. 41-43), when "margin kerning" is used when lines are already laid. "Level 2", on the other hand, incorporates margin kerning into the layout algorithm.

Since very good results were obtained in #6161 when character-level justification was included in the layout algorithm without much changes, I think we can do this too. I will study this in more detail.

See the comparison between level 1 and level 2 character protruding from Hàn Thế Thành dissertation (p. 47) below:

image

sicikh:
_I think_, that now overhang characters' width, that they add to the line's one, is included into the layout algorithm. The results show exactly this, but not by amount, that was shown in the dissertation. I'll investigate it further. But the result is great nonetheless.

image

<details>
<summary>
Code
</summary>

```typ
#set page(paper: "a4")

// Wiki suggested values + Hàn Thế Thành’s original settings
//
// - https://en.wikipedia.org/wiki/Optical_margin_alignment#
// - https://ftp.snt.utwente.nl/pub/software/tex/macros/latex/contrib/microtype/microtype-code.pdf, p. 155
#let protrusion-table = (
  "A": (20%, 20%),
  "F": (0%, 5%),
  "J": (5%, 0%),
  "K": (0%, 5%),
  "L": (0%, 5%),
  "C": (10%, 0%),
  "O": (10%, 10%),
  "T": (20%, 20%),
  "V": (20%, 20%),
  "W": (20%, 20%),
  "X": (5%, 5%),
  "Y": (20%, 20%),
  "k": (0%, 5%),
  "r": (0%, 5%),
  "t": (0%, 10%),
  "v": (5%, 5%),
  "w": (5%, 5%),
  "x": (5%, 5%),
  "y": (10%, 10%),
  "c": (10%, 10%),
  "o": (10%, 10%),
  ".": (0%, 100%),
  ",": (0%, 100%),
  ":": (0%, 100%),
  ";": (0%, 100%),
  "!": (0%, 20%),
  "?": (0%, 20%),
  "(": (20%, 0%),
  ")": (0%, 20%),
  "-": (0%, 75%),
  "\u{ad}": (0%, 75%),
  "–": (0%, 50%),
  "—": (0%, 25%),
  "“": (100%, 0%),
  "”": (0%, 100%),
)
#let italic-protrusion-table = (
  "p": (20%, 0%),
)
#let use-another-table-for-emph(body) = {
  show emph: set text(overhang: (map: italic-protrusion-table, default: false))
  body
} 

#let example(name) = [
  #place(top, float: true, scope: "parent", strong(name))
  
Shortly after this, when the sexton
came to pay them a visit, the father
broke out to him, and told him what
a bad hand his youngest son was
at everything: he knew nothing and
learned nothing. “Only think! when
I asked him how he purposed gaining a livelihood, he actually asked
to be taught to shudder.” “If that’s
all he wants,” said the sexton, “I can
teach him that; just you send him to
me, I’ll soon polish him up.” The father was quite pleased with the proposal, because he thought: “It will
be a good discipline for the youth.”
And so the sexton took him into his
house, and his duty was to toll the
bell. After a few days he woke him
at midnight, and bade him rise and
climb into the tower and toll. “Now,
my friend, I’ll teach you to shudder,”
thought he. He stole forth secretly
in front, and when the youth was
up above, and had turned round to
grasp the bell-rope, he saw, standing _opposite_ the hole of the belfry,
a white figure. “Who’s there?” he
called out, but the figure gave no answer, and neither stirred nor moved.
“Answer,” cried the youth, “or begone; you have no business here
at this hour of the night.” But the
sexton remained motionless, so that
the youth might think that it was a
ghost. The youth called out the second time: “What do you want here?
Speak if you are an honest fellow, or
]

#set page(width: 500pt, height: 21em, margin: 15pt)
#set par(justify: true)
#set text(size: 0.8em)

#grid(
  columns: (1fr, 1fr, 1fr, 1fr),
  gutter: 10pt,
  {
    set text(overhang: false)
    example[No protruding]
  },
  {
    set text(overhang: (map: protrusion-table, default: false))
    show: use-another-table-for-emph
    example[Character protruding]
  },
  {
    set text(overhang: false)
    set par(justification-limits: (tracking: (max: 0.02em, min: -0.02em)))
    example[Character level justification]
  },
  {
    set text(overhang: (map: protrusion-table, default: false))
    set par(justification-limits: (tracking: (max: 0.02em, min: -0.02em)))
    show: use-another-table-for-emph
    example[Protruding + justification]
  }
)
```

</details>

Secondly, the idea of passing a function to the `overhang` parameter was interesting, but it has one significant technical limitation — calling a function requires a mutable reference to the `Engine`, while layout algorithm has only shared one. So I reverted the changes and left "ad-hoc" dictionary with two entries: the table itself ("map") and default behaviour ("default" field).

Also I changed protrusion value from `Ratio` to `Rel` — Donald Knuth commented on the Hàn Thế Thành's dissertation, that a more correct option would be to specify protrusion not in ratio to a glyph width, but in `em`. Fortunately, we can support both options.

My next steps will be to reassure, that the implementation is correct (currently I have some doubts), maybe make `overhang` foldable and improve error messages.

YDX-2147483647:
Hi! I see that `text.overhang.map` does not accept multi-character punctuation marks like `——` (CJK two-em dash, two consecutive U+2014).

> error: protrusion table keys must be single characters

Is that an essential limitation of the current design, or merely a superficial restriction of the interface (or data types)?
If the latter, I wish the restriction can be removed. It will provide a workaround for #6735.

----

Besides, I suggest updating the PR description or at least adding a note. The current version is outdated and misleading.

sicikh:
> Is that an essential limitation of the current design, or merely a superficial restriction of the interface (or data types)? If the latter, I wish the restriction can be removed. It will provide a workaround for #6735.

Hi! Yes, this is a current limitation of the interface, but not a technical one. I just haven't had time to think about how to describe the protrusions for several characters, but at the same time to disallow, for example, `"abcdef": (9999%, 9999%)`.

> Besides, I suggest updating the PR description or at least adding a note. The current version is outdated and misleading.

Also thought of this, will change it in a minute. UPD: done.

sicikh:
@YDX-2147483647 

> Is that an essential limitation of the current design, or merely a superficial restriction of the interface (or data types)?
If the latter, I wish the restriction can be removed. It will provide a workaround for https://github.com/typst/typst/issues/6735.

With time have passed and a bit of experiments, I think that is the essential limitation of the current design. But the specified issue is resolved via another (more robust) way, see #8131 and comments in the TODO section of the current PR description.

With that additional PR you can specify _exact glyph_ that will visually appear in the document by its UTF-8 character. So if you specify `"\u{2E3A}": (0%, 0%)`, glyph, that will be triggered by character, won't be overhang in either way, even if it was shaped by two em-dashes, that overhang by themselves alone.

The only glyphs that cannot be accessed by UTF-8 characters are ligatures. I think that we should adapt `microtype`'s way of defining them in the table via `"/glyph-name"`.

sicikh:
> The only glyphs that cannot be accessed by UTF-8 characters are ligatures. I think that we should adapt microtype's way of defining them in the table via "/glyph-name".

Added exactly this, also merged #8131. The PR description is also updated.

The PR is ready for review. I don't think I'll progress any further without additional help.

adamkrellenstein:
This would solve a case I'm hitting: optical margin alignment of capitals (W/A/V/T) at flush-left line starts, which the current punctuation-only overhang table doesn't cover.

sicikh:
I've merged current main with #8131, PR is no longer blocked and is ready for review.

laurmaedje (commented):
Thanks for the work you've put into this far. I've finally had time to consider this in more depth together with @reknih. The big picture of the code changes seems fine, but before diving into that, I would like to address some final API design concerns. I think the building blocks of the API are going into a good direction, but there are still some concerns and suggestions I have:

- Most importantly, the dissonance between the `default` and `map` APIs. The APIs cover two complementary parts of the glyph space and need to describe similar options, but their shape is completely different. The type of `default` is `start | end | left | right | false | true` while a map entry is `(auto | relative, auto | relative)`). However, the `default` API can mostly be expressed in terms of the `map` one:
    - `false` is equivalent to `(0pt, 0pt)`
    - `true` is equivalent to `(auto, auto)`
    - `start` is equivalent to `(auto, 0pt)`
    - ...

  For consistency, I would thus suggest adopting the map model for both. To make the common cases more convenient for both, I would additionally suggest allowing a single value instead of an array to have both uniform, e.g. `auto` instead of `(auto, auto)` and `0pt` instead of `(0pt, 0pt)`.

  If we think it's too awkward for something like `overhang: 10pt` to overhang _everything_ (a pretty useless operation which we probably want to avoid exposing), we could add `none` as an alternative to `0pt` and make the default API a subset of the map API that only allows combinations of `none` and `auto`, not concrete values.

  As far as I understand, the array form only supports start/end, not left/right, but that seems sufficient to me.

- I also don't really like the `default` and `map` names structure in general. It creates a lot of nesting and names aren't immediately clear. Perhaps we could rather have the map top-level, with a `rest` key for the default?

  ```typ
  #set text(overhang: (
    "-": (none, 50%),
    "W": 20%,
    rest: none,
  ))
  ```

  As with `default` in your PR, the value of the `rest` key could also be specified top-level on `overhang` as a short form. And that in turn could be shortened again if both sides are symmetric, so both `overhang: none` and `overhang: auto` would work for the current `false` and `true`. The compatibility behavior for `false` and `true` should remain, but give a deprecation warning that you should use `none` and `auto`, respectively.

- The PR is missing documentation for the `overhang` field. The docs should describe the expected structure in detail since the signature itself is not enough (it just says `dictionary`).

- I am not 100% sure about exposing the PostScript names, but I _think_ I can see it. However, if we keep this, it would need to be well-documented with an explanation how to determine these names.[^1]

- If we ever add optical margin overhang, we could migrate `auto` to this smarter default.

[^1]: One alternative would be to allow multi-letter strings in the overhang map, _shape_ them and see which glyph they result in. Though maybe that's a bit extreme and it would probably still not allow selecting every possible glyph. It would also introduce a collision with `rest`, which is currently non-colliding since it's neither a single char nor starts with a slash.

laurmaedje:
@sicikh Any thoughts on the points above? I think the PR is in a fairly good state already, so if you'd want to finish it, that would be great. Otherwise, I would probably try to finish it myself at some point.

PR #7783, opened 2026-01-30
HTML Export: Exclude metadata content from paragraph grouping

LiamSnow:
## Summary
Metadata elements (`<meta>, <link>, <script>, <style>`, ..) were incorrectly being wrapped in `<p>` tags during HTML. This issue was introduced in #7505.

## Problem
After #7505 changed paragraph grouping to use `is_phrasing_content`, metadata elements started being wrapped in `<p>` because they're technically phrasing content ([source](https://html.spec.whatwg.org/dev/dom.html#phrasing-content)). However, they should never be wrapped in `<p>` tags and doing so will break the page in browsers (they don't expect `<p>` tags in `<head>` and so will move that content to the `<body>`).

Ex.

```typst
#html.head({
  html.title("Test")
  html.meta(charset: "utf-8")
  html.meta(name: "viewport", content: "test")
})
```

incorrectly produces:
```html
<head>
  <title>Test</title>
  <p>
    <meta charset="utf-8">
    <meta name="viewport" content="test">
  </p>
</head>
```

which the browser (firefox in this case) converts to something completely wrong:
```html
<html>
  <head>
    <title>Test</title>
  </head>
  <body>
    <p>
      <meta charset="utf-8">
      <meta name="viewport" content="test">
    </p>
  </body>
</html>
```

## Fix
Exclude metadata content from the phrasing check used for paragraph grouping.

While this does fix the problem, it would probably be best to rename the function. I wanted to keep this PR minimal though.

### DIFF
--- a/crates/typst-html/src/lib.rs
+++ b/crates/typst-html/src/lib.rs
@@ -119,10 +119,13 @@ impl HtmlElem {
         }
     }
 
-    /// Checks whether the given element is "phrasing content" in HTML.
+    /// Checks whether the given element is "phrasing content" in HTML for
+    /// paragraph grouping purposes. Excludes metadata content since those
+    /// elements should not be wrapped in paragraphs.
     fn is_phrasing(elem: &Content) -> bool {
-        elem.to_packed::<HtmlElem>()
-            .is_some_and(|elem| tag::is_phrasing_content(elem.tag))
+        elem.to_packed::<HtmlElem>().is_some_and(|elem| {
+            tag::is_phrasing_content(elem.tag) && !tag::is_metadata_content(elem.tag)
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


laurmaedje:
The current situation is definitely bad, but this also feels a bit like a band-aid. HTML allows metadata content in paragraphs. The issue is more that currently everything that is allowed to be in a paragraph also always becomes a paragraph. This issue goes beyond the metadata tags.

I intend to take a stab at the paragraph situation soon (I also need it for Docs-in-Typst), but would rather not have lots of back-and-forth changes to it in the meantime. I'll thus close this and ask for a bit more patience. Thank you regardless!

LiamSnow:
Totally makes sense. Excited to see where typst-html goes!

laurmaedje:
I've opened a proposal for how we could fix the underlying issue: https://github.com/typst/typst/issues/8159

PR #7784, opened 2026-01-30
Disallow content body for raw HTML tags

### LINKED ISSUE #7724: Disallow content body for `html.style` and `html.script`
### Description

As the body of these is foreign content, Typst's normal content processing does not make any sense.

For `html.elem`, I would rather not special case the body handling, but not 100% sure.

### Use Case

See e.g. https://github.com/typst/typst/issues/7714

andylizi:
`html.script` and `html.style` now only accepts `str` instead of `content`.

This check is only done for `html.script` and `html.style`. It does currently not include untyped `html.elem("script")` per reasoning in #7724.

It does not include 'escapable' raw tags like `html.textarea` and `html.title`, because I think `html.title[A "clever" example]` should gets compiled to `<title>A “clever” example</title>`.

Fixes #7724.

### DIFF
--- a/crates/typst-html/src/typed.rs
+++ b/crates/typst-html/src/typed.rs
@@ -85,30 +85,25 @@ fn create_param_info(element: &'static data::ElemInfo) -> Vec<ParamInfo> {
     }
     let tag = HtmlTag::constant(element.name);
     if !tag::is_void(tag) {
-        params.push(if tag::is_raw(tag) {
-            ParamInfo {
-                name: "body",
-                docs: "The text content of the HTML element.",
-                input: CastInfo::Type(Type::of::<Str>()),
-                default: None,
-                positional: true,
-                named: false,
-                variadic: false,
-                required: false,
-                settable: false,
-            }
-        } else {
-            ParamInfo {
-                name: "body",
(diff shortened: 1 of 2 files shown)


crates/typst-html/src/typed.rs
```diff
@@ -83,23 +85,41 @@ fn create_param_info(element: &'static data::ElemInfo) -> Vec<ParamInfo> {
...
-            input: CastInfo::Type(Type::of::<Content>()),
-            default: None,
-            positional: true,
-            named: false,
-            variadic: false,
-            required: false,
-            settable: false,
+        params.push(if tag::is_raw(tag) {
+            ParamInfo {
+                name: "content",
```
laurmaedje: the term `content` for the _non_-content case is a bit confusing.

laurmaedje (commented):
On this one, I'd skip the deprecation period. We can just break the API as HTML export is in development and experimental.

andylizi:
Done, now the error just says "expected string, found content", do we want a more specific message for this case?

laurmaedje:
Thanks! I think the message is fine.

PR #7786, opened 2026-02-01
Fix duplicated `match` branch and typo in `typst-pdf` crate

OneProgGit:
There was a duplicated `match` branch in page.rs file of typst-pdf crate, and it couldn't be caused in any way. Also, there was a typo in word "optimization"

### DIFF
--- a/crates/typst-pdf/src/page.rs
+++ b/crates/typst-pdf/src/page.rs
@@ -23,7 +23,7 @@ impl PageLabelExt for PageLabel {
 
             let (prefix, kind) = pat.pieces.first()?;
 
-            // If there is a suffix, we cannot use the common style optimisation,
+            // If there is a suffix, we cannot use the common style optimization,
             // since PDF does not provide a suffix field.
             let style = if pat.suffix.is_empty() {
                 use krilla::page::NumberingStyle as Style;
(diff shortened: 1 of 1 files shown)


cady-b:
Huh, good catch! The error seems to have been introduced in #5059

Interestingly, it seems krilla is doing some work to make this not actually result in a change that's visible for non-technical consumers. This patch doesn't result in a change with `#set page(numbering: "a")`, but for `#set page(numbering: "A")`, we get the following diff (plus new checksums / pointers):
```diff
@@ -73,14 +73,16 @@ endobj
 9 0 obj
 <<
   /Type /PageLabel
-  /P (A)
+  /S /A
+  /St 1
 >>
 endobj

 10 0 obj
 <<
   /Type /PageLabel
-  /P (B)
+  /S /A
+  /St 2
 >>
 endobj
```
where `/S /A` is the pre-defined uppercase Arabic with a `/St …` starting index, and `/P (…)` a custom string computed from the numbering. The lowercase version uses `/S /a` both before and after the patch, so this sounds like an improvement! ^^

laurmaedje:
Thanks!

Worth mentioning that "optimisation" wasn't incorrectly spelled, just in the British variant. But we generally use the American spelling in the codebase, so I'm fine with the change.

PR #7798, opened 2026-02-03
Improve diagnostic for compiler error when spreading multiple dictionaries together

### LINKED ISSUE #7795: Cannot spread dictionaries into empty dictionary literal
### Description

We can spread one dictionary into another as long as the dictionary literal we're spreading into has at least one key-value pair:

```
#let default-data = (a: 1, b: 2, c: 3)
#let data = (..default-data, d: 4, e: 5)
```

However, it there is no key-value pair in target dict literal, compiler incorrectly assumes it's array we're spreading into, and throws error: "Cannot spread dictionary into array".

```
error: cannot spread dictionary into array
   ┌─ src/main.typ:30:13
   │
30 │ #let data = (..default-data, ..invoice-data)
   │              ^^^^^^^^^^^^^^
```

It kind of makes sense since there's no syntactical differentiation between empty array literal and empty dict literal (`()`). But I think compiler could be smart enough to determine type of resulting data structure from its operands, since it already knows their types. It would throw error only if you would mix arrays and dictionary spreads in single structure.

As a workaround, I currently have to add dummy key-value pair to literal to tell compiler it's dictionary, not array.

### Reproduction URL

https://typst.app/project/RzT3a1vnTHVNwLvFuCHXzO

### Operating system

_No response_

### Typst version

dylwil3:
If a user attempts to spread several dictionaries together without using the [special syntax](https://typst.app/docs/reference/foundations/dictionary#1) `(:..spread)` then they get a potentially confusing compiler error:

```typst
#{
  let x = (a: 1)
  let y = (b: 2)
  // This gives error "cannot spread dictionary into array"
  (..x,..y)
}
```

This PR detects when all the array items are spreads of dictionaries and suggests the special syntax as a hint in that case.

Closes #7795

Apologies for any gaffes, this is my first contribution - happy to make any adjustments if I've gone astray!

### DIFF
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -228,13 +228,10 @@ impl Eval for ast::Array<'_> {
 
         let mut vec = EcoVec::with_capacity(items.size_hint().0);
 
-        // We raise an error when one of the array items is
-        // the spread of a dictionary. If _all_ of the array
-        // items are spreads of dictionaries, the user probably
-        // wanted to write `(: ..dict_a, ..dict_b)` instead to
-        // create a dictionary, not an array. In this case we
-        // provide a hint to that effect. This boolean is to
-        // track whether we should emit that hint.
+        // We raise an error when one of the array items is the spread of a
+        // dictionary. If _all_ of the array items are spreads of dictionaries,
+        // the user probably wanted to write `(: ..dict_a, ..dict_b)` instead
+        // to create a dictionary, not an array.
         let mut items_seen_are_spread_dicts = true;
 
         while let Some(item) = items.next() {
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -3,7 +3,7 @@ use typst_library::diag::{At, SourceResult, bail, error, warning};
 use typst_library::engine::Engine;
 use typst_library::foundations::{
     Array, Capturer, Closure, ClosureNode, Content, ContextElem, Dict, Func,
-    NativeElement, Selector, Str, Type, Value, ops,
+    NativeElement, Selector, Str, Value, ops,
 };
 use typst_library::introspection::{Counter, State};
 use typst_syntax::ast::{self, AstNode};
@@ -249,7 +249,7 @@ impl Eval for ast::Array<'_> {
                         items_seen_are_spread_dicts = false;
                         vec.extend(array.into_iter())
                     }
-                    v if items_seen_are_spread_dicts && v.ty() == Type::of::<Dict>() => {
+                    v @ Value::Dict(_) if items_seen_are_spread_dicts => {
                         // Lookahead to see whether remaining items
                         // are spreads of dicts
                         if items.all(|it| {
@@ -258,7 +258,7 @@ impl Eval for ast::Array<'_> {
                             };
                             spd.expr()
                                 .eval(vm)
-                                .is_ok_and(|val| val.ty() == Type::of::<Dict>())
+                                .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
                         }) {
                             bail!(spread.span(), "cannot spread {} into array", v.ty(); hint: "open container with `(:` to create a dictionary")
                         } else {
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -228,22 +228,28 @@ impl Eval for ast::Array<'_> {
 
         let mut vec = EcoVec::with_capacity(items.size_hint().0);
 
-        let mut all_items_are_variable_spreads = true;
+        // We raise an error when one of the array items is
+        // the spread of a dictionary. If _all_ of the array
+        // items are spreads of dictionaries, the user probably
+        // wanted to write `(: ..dict_a, ..dict_b)` instead to
+        // create a dictionary, not an array. In this case we
+        // provide a hint to that effect. This boolean is to
+        // track whether we should emit that hint.
+        let mut items_seen_are_spread_dicts = true;
+
         while let Some(item) = items.next() {
             match item {
                 ast::ArrayItem::Pos(expr) => {
-                    all_items_are_variable_spreads = false;
+                    items_seen_are_spread_dicts = false;
                     vec.push(expr.eval(vm)?)
                 }
                 ast::ArrayItem::Spread(spread) => match spread.expr().eval(vm)? {
                     Value::None => {}
                     Value::Array(array) => {
-                        all_items_are_variable_spreads = false;
+                        items_seen_are_spread_dicts = false;
                         vec.extend(array.into_iter())
                     }
-                    v if all_items_are_variable_spreads
-                        && v.ty() == Type::of::<Dict>() =>
-                    {
+                    v if items_seen_are_spread_dicts && v.ty() == Type::of::<Dict>() => {
                         // Lookahead to see whether remaining items
                         // are spreads of dicts
                         if items.all(|it| {
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -246,21 +246,22 @@ impl Eval for ast::Array<'_> {
                         items_seen_are_spread_dicts = false;
                         vec.extend(array.into_iter())
                     }
-                    v @ Value::Dict(_) if items_seen_are_spread_dicts => {
-                        // Lookahead to see whether remaining items
-                        // are spreads of dicts
-                        if items.all(|it| {
+                    v @ Value::Dict(_)
+                        if items_seen_are_spread_dicts
+                        // Lookahead to see whether remaining items are spreads
+                        // of dicts
+                        && items.all(|it| {
                             let ast::ArrayItem::Spread(spd) = it else {
                                 return false;
                             };
                             spd.expr()
                                 .eval(vm)
                                 .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
-                        }) {
-                            bail!(spread.span(), "cannot spread {} into array", v.ty(); hint: "open container with `(:` to create a dictionary")
-                        } else {
-                            bail!(spread.span(), "cannot spread {} into array", v.ty())
-                        }
+                        }) =>
+                    {
+                        bail!(spread.span(), "cannot spread {} into array",
+                            v.ty();
+                        hint: "add a colon to create a dictionary instead `(: ..dict)`")
                     }
                     v => bail!(spread.span(), "cannot spread {} into array", v.ty()),
                 },
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -250,8 +250,7 @@ impl Eval for ast::Array<'_> {
                             let ast::ArrayItem::Spread(spd) = it else {
                                 return false;
                             };
-                            spd
-                                .expr()
+                            spd.expr()
                                 .eval(vm)
                                 .is_ok_and(|val| val.ty() == Type::of::<Dict>())
                         }) {
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -250,18 +250,20 @@ impl Eval for ast::Array<'_> {
                         if items_seen_are_spread_dicts
                         // Lookahead to see whether remaining items are spreads
                         // of dicts
-                        && items.all(|it| {
-                            let ast::ArrayItem::Spread(spd) = it else {
-                                return false;
-                            };
-                            spd.expr()
-                                .eval(vm)
-                                .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
-                        }) =>
+                        && items.all(|item| matches!(
+                            item,
+                            ast::ArrayItem::Spread(spread) if matches!(
+                                spread.expr().eval(vm),
+                                Ok(Value::Dict(_)),
+                            ),
+                        )) =>
                     {
-                        bail!(spread.span(), "cannot spread {} into array",
-                            v.ty();
-                        hint: "add a colon to create a dictionary instead `(: ..dict)`")
+                        let fixed =
+                            self.to_untyped().clone().into_text().replacen("(", "(: ", 1);
+                        bail!(
+                            spread.span(), "cannot spread {} into array", v.ty();
+                            hint: "add a colon to create a dictionary instead `{fixed}`";
+                        )
                     }
                     v => bail!(spread.span(), "cannot spread {} into array", v.ty()),
                 },
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -3,7 +3,7 @@ use typst_library::diag::{At, SourceResult, bail, error, warning};
 use typst_library::engine::Engine;
 use typst_library::foundations::{
     Array, Capturer, Closure, ClosureNode, Content, ContextElem, Dict, Func,
-    NativeElement, Selector, Str, Value, ops,
+    NativeElement, Selector, Str, Type, Value, ops,
 };
 use typst_library::introspection::{Counter, State};
 use typst_syntax::ast::{self, AstNode};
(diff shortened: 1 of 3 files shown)


crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,42 @@ impl Eval for ast::Array<'_> {
...
+                }
                 ast::ArrayItem::Spread(spread) => match spread.expr().eval(vm)? {
                     Value::None => {}
-                    Value::Array(array) => vec.extend(array.into_iter()),
+                    Value::Array(array) => {
+                        all_items_are_variable_spreads = false;
+                        vec.extend(array.into_iter())
+                    }
+                    v if all_items_are_variable_spreads
+                        && v.ty() == Type::of::<Dict>() =>
```
dylwil3: Is this the preferred way to check the type of something?

tests/suite/foundations/dict.typ
```diff
@@ -61,6 +61,17 @@
...
 
+--- spread-multiple-dicts-diagnostic paged ---
+
+#{
+  let x = (a: 1)
+  let y = (b: 2)
+  let z = (3,4)
+  // Error: 4-7 cannot spread dictionary into array
+  // Hint: 4-7 use `(:..spread)` syntax to spread multiple dictionaries together
+  (..x,..y)
```
dylwil3, line 88: I'd like to add more tests here, but I was slowed down because I couldn't get

```
cargo testit --exact spread-multiple-dicts-diagnostic --update
```

to automatically update the snapshot for the annotation (assuming that is what it's meant to do?)

Moreover, the test failure gives the ranges for the annotations in `column:row` form, whereas I think the test infrastructure wants byte offsets - so I had to do some manual counting to sort it out.

Any suggestions here?

crates/typst-eval/src/code.rs
isuffix: For content types like `TextElem`, yes, but for primitves like `Dict` you can match on the `Value` enum: `match v { Value::Dict(_) if ... => ... }` or as a boolean: `matches!(v, Value::Dict(_))`.

tests/suite/foundations/dict.typ
isuffix: I'm working on making `--update` work for error annotations! ...but it's not yet done. I do have a WIP branch that needs cleanup, but should function if you want to work on top of it.

Working on top of my branch would be somewhat complex, but maybe worthwhile: you would add my repo (<https://github.com/isuffix/typst>) as an upstream called `isuffix`, then rebase this branch on on top of `isuffix/test-notes-update` (and replace the `paged` attribute on the test header to `eval`). Then you could add tests with `--update` until satisfied, before cherry picking commits back on top of main for the PR.

Ok, typing it out is complicated, but honestly not that bad depending on your level of comfort with `git`. If you're using `jj` though, it would be a breeze.

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,41 @@ impl Eval for ast::Array<'_> {
     type Output = Array;
 
     fn eval(self, vm: &mut Vm) -> SourceResult<Self::Output> {
-        let items = self.items();
+        let mut items = self.items();
 
         let mut vec = EcoVec::with_capacity(items.size_hint().0);
-        for item in items {
+
+        let mut all_items_are_variable_spreads = true;
```
isuffix: This variable's name is a bit confusing to me. Part of it is that a spread doesn't need to be spreading a variable, it can be any expression, e.g. `(..(1, 2), ..none, ..(3, 4))`. But you're also setting it to false even when we spread an array, so I think it would be clearer if you rename it to something simpler. Maybe just "`at_start`" because the hint is only useful if the spread is applying to a dictionary at the start of the array? Although I guess that differs in the `(..none, ..dict)` case. Lmk your thoughts.

It may also help to add a comment explaining its use: `// We don't want to give the hint for spreading dicts if the spread occurs in the middle of a normal array.`

saecki:
I think this is a bit misleading, since `:..` isn't an operator, and this hint also applies for only one dictionary.
I would rather check if the spreaded item is a dictionary and suggest that the container must also be a dictionary, which can be indicated by adding the colon `(:)`. So, for the spread example `(: ..dict)`, and with multiple dicts it would be like this: ` (: ..dict-a, ..dict-b)`

The footnote should probably also be updated.

crates/typst-eval/src/code.rs
dylwil3: I renamed the variable to `items_seen_are_spread_dicts` and added an explanatory comment. The point is that (I think?) we only want to give the hint when _all_ the items are spreads of dictionaries (variable or otherwise).

tests/suite/foundations/dict.typ
dylwil3: thanks for the suggestion! I ended up just counting - it wasn't too bad when I realized what the ranges were actually referring to 😄 - I should've read the testing README more carefully first

crates/typst-eval/src/code.rs
dylwil3: nice! I had incorrectly assumed that only worked for dict literals

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
...
+                        // are spreads of dicts
+                        if items.all(|it| {
+                            let ast::ArrayItem::Spread(spd) = it else {
+                                return false;
+                            };
+                            spd.expr()
+                                .eval(vm)
+                                .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
+                        }) {
+                            bail!(spread.span(), "cannot spread {} into array", v.ty(); hint: "open container with `(:` to create a dictionary")
```
dylwil3: I tried a few different possible wordings and I wasn't really happy with any of them (including this one)... happy to accept suggestions (both for this and the changed docs footnote)

saecki: Maybe ``hint: "add a colon to create a dictionary instead `(: ..dict)`"``?

saecki: Id' also move the `hint: ` onto a new line

tests/suite/foundations/dict.typ
isuffix: Good news, my change got merged so this should be easier for next time :D

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
     type Output = Array;
 
     fn eval(self, vm: &mut Vm) -> SourceResult<Self::Output> {
-        let items = self.items();
+        let mut items = self.items();
 
         let mut vec = EcoVec::with_capacity(items.size_hint().0);
-        for item in items {
+
+        // We raise an error when one of the array items is
```
isuffix: For formatting in Rust, we wrap multiline comments and doc comments at 80, and code at 90. You should reformat this paragraph, the "lookahead" comment below, and the footnote for `Dict`.

If you want, you can likely install an editor extension to do this with a keybind, such as Rewrap Revived for VSCode (I use it with "Rewrap Whole Comment" toggled off).

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
...
+        // the spread of a dictionary. If _all_ of the array
+        // items are spreads of dictionaries, the user probably
+        // wanted to write `(: ..dict_a, ..dict_b)` instead to
+        // create a dictionary, not an array. In this case we
+        // provide a hint to that effect. This boolean is to
+        // track whether we should emit that hint.
```
isuffix:
```suggestion
        // create a dictionary, not an array.
```
These sentences are sufficiently clear from the code, you should leave them out.

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
...
+                        // Lookahead to see whether remaining items
+                        // are spreads of dicts
+                        if items.all(|it| {
+                            let ast::ArrayItem::Spread(spd) = it else {
+                                return false;
+                            };
+                            spd.expr()
+                                .eval(vm)
+                                .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
+                        }) {
```
isuffix: You can remove the duplicated `bail!` in the else-block by moving this into the guard of the match.

```suggestion
                    v @ Value::Dict(_) if items_seen_are_spread_dicts
                        && items.all(|it| {
                            matches!(
                                ast::ArrayItem::Spread(spread)
                                if spread
                                    .expr()
                                    .eval(vm)
                                    .is_ok_and(|value| matches!(value, Value::Dict(_))
                            )
                        })
                        =>
                        {
```
(modulo rustfmt)

crates/typst-eval/src/code.rs
isuffix: This should be reformatted to less than 90 columns. You can split the strings with a backslash if necessary.

tests/suite/foundations/dict.typ
```diff
@@ -61,6 +61,35 @@
   test((..(a: 1), b: 2), (a: 1, b: 2))
 }
 
+--- spread-dicts-in-array-diagnostic paged ---
+
```
isuffix: This is a small thing, but I'm a stickler. Please remove these leading newlines after the test headers.

tests/suite/foundations/dict.typ
```diff
@@ -61,6 +61,35 @@
...
+--- spread-dict-and-array-in-array-diagnostic paged ---
+// Here we should not emit a hint to use `(: <...>)` syntax
+// since the developer's intent is unclear
+#{
+  let x = (a: 1)
+  let y = (2,3)
+  // Error: 4-7 cannot spread dictionary into array
+  (..x,..y)
+}
+
```
isuffix, line 90: We should add one more test variant with a spread dict and a normal item:

```suggestion

--- spread-dict-and-array-in-array-diagnostic paged ---
// We should not emit a hint to use `(: <...>)` syntax
// since the developer's intent is unclear
#{
  let x = (a: 1)
  // Error: 4-7 cannot spread dictionary into array
  (..x, "item")
}
```

isuffix (changes requested):
I've reviewed the code more thoroughly now, and the overall approach seems fine. I do have a bunch of code organization/formatting changes for you, but I want to reiterate that I appreciate the contribution and your working on this!

crates/typst-eval/src/code.rs
dylwil3: I use helix so I tried just doing `:reflow 80` for the comments... hopefully that worked!

crates/typst-eval/src/code.rs
dylwil3: not sure why rustfmt didn't break here - I moved the whole `hint` part to the next line

dylwil3:
Thanks so much for your feedback and guiding me through my first contribution here! It's a great codebase. Apologies for the long delay in addressing your comments.

I realized after rebasing and force pushing that that may have been rude since it's harder in the GitHub UI to see changes since your last review. Let me know if you would prefer that I just merge in `main` in the future!

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,45 @@ impl Eval for ast::Array<'_> {
...
+
+        while let Some(item) = items.next() {
             match item {
-                ast::ArrayItem::Pos(expr) => vec.push(expr.eval(vm)?),
+                ast::ArrayItem::Pos(expr) => {
+                    items_seen_are_spread_dicts = false;
+                    vec.push(expr.eval(vm)?)
+                }
                 ast::ArrayItem::Spread(spread) => match spread.expr().eval(vm)? {
                     Value::None => {}
```
isuffix, line 244: I just realized there's an edge case here if we spread a none and then a dictionary: `(..none, ..(one: 1))`.

I think we should include the hint in this case, so the current code is correct, but we should add this as a test case.

crates/typst-eval/src/code.rs
isuffix: Yeah, it's unfortunate. My understanding is that when dealing with macros rustfmt can manage some reformatting but generally doesn't try as hard due to their flexibility.

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,45 @@ impl Eval for ast::Array<'_> {
...
+                                return false;
+                            };
+                            spd.expr()
+                                .eval(vm)
+                                .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
+                        }) =>
+                    {
+                        bail!(spread.span(), "cannot spread {} into array",
+                            v.ty();
+                        hint: "add a colon to create a dictionary instead `(: ..dict)`")
```
isuffix: We can actually improve this hint and get the full array expression as text by accessing the array's syntax node directly (via `self.to_untyped()`). And since we're producing an error (not a hot path), there's no issue with doing a mildly expensive clone.

```suggestion
                        let fixed =
                            self.to_untyped().clone().into_text().replacen("(", "(: ", 1);
                        bail!(
                            spread.span(), "cannot spread {} into array", v.ty();
                            hint: "add a colon to create a dictionary instead `{fixed}`";
                        )
```
Note that the clone is necessary because `AstNode::to_untyped()` returns a `&SyntaxNode`, but `SyntaxNode::into_text()` takes an owned `SyntaxNode`. And while `SyntaxNode::text()` exists and takes a reference, it doesn't produce the full text for nested trees, only for leaf nodes like identifiers.

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,45 @@ impl Eval for ast::Array<'_> {
...
+                        // Lookahead to see whether remaining items are spreads
+                        // of dicts
+                        && items.all(|it| {
+                            let ast::ArrayItem::Spread(spd) = it else {
+                                return false;
+                            };
+                            spd.expr()
+                                .eval(vm)
+                                .is_ok_and(|spd| matches!(spd, Value::Dict(_)))
+                        }) =>
```
isuffix: Alright, last notes on this block.

We can use the fancy `matches!(thing, pattern(...) if expr)` syntax here to remove the let-else-return statement and keep this as one expression. It's still ugly, but I think it's a bit nicer.

Also, we should keep the variable names as `item` and `spread` since these are the same expressions as used elsewhere in the loop.

```suggestion
                        && items.all(|item| matches!(
                            item,
                            ast::ArrayItem::Spread(spread) if matches!(
                                spread.expr().eval(vm),
                                Ok(Value::Dict(_)),
                            ),
                        )) =>
```

isuffix (commented):
No worries about the delay. Asynchronous review takes time, and I have plenty of tasks to work on otherwise. It's a double-edged sword kind of thing.

I just have two code suggestions and one more edge case to add a test for, otherwise this should be ready for @laurmaedje.

As for rebasing vs. merging main: I usually prefer to force-push with a rebase, but it's best to do so as two pushes: first with just the rebase, and second with your new changes. That makes it possible to view the new changes with the compare button in the github UI. But it's also fine if your workflow makes it easier to just merge main in. We generally squash PRs into one commit on main unless you keep individual commits high-quality and ensure the code passes tests on all of them, then Laurenz can keep them separate manually when merging. Hopefully github catches up with the stacked-diffs review methods that other tools are adding 🤞

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
...
+                        if items_seen_are_spread_dicts
+                        // Lookahead to see whether remaining items are spreads
+                        // of dicts
+                        && items.all(|item| matches!(
+                            item,
+                            ast::ArrayItem::Spread(spread) if matches!(
+                                spread.expr().eval(vm),
+                                Ok(Value::Dict(_)),
+                            ),
+                        )) =>
```
laurmaedje, line 259: for a diagnostic, this is maybe fine I guess, but overall, it feels a bit unsound because we're suddenly evaluating only parts of the AST while skipping others even though the things in between could have side effects.

maybe it's fine but thoughts @isuffix?

crates/typst-eval/src/code.rs
laurmaedje: we should make `text()` less of a footgun. maybe rename it or let it return `Option` or something. unrelated to the PR though.

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
...
+                    {
+                        let fixed =
+                            self.to_untyped().clone().into_text().replacen("(", "(: ", 1);
+                        bail!(
+                            spread.span(), "cannot spread {} into array", v.ty();
+                            hint: "add a colon to create a dictionary instead `{fixed}`";
```
laurmaedje:
```suggestion
                            hint: "add a colon to create a dictionary instead: `{fixed}`";
```

crates/typst-eval/src/code.rs
```diff
@@ -224,15 +224,47 @@ impl Eval for ast::Array<'_> {
...
+        let mut items = self.items();
 
         let mut vec = EcoVec::with_capacity(items.size_hint().0);
-        for item in items {
+
+        // We raise an error when one of the array items is the spread of a
+        // dictionary. If _all_ of the array items are spreads of dictionaries,
+        // the user probably wanted to write `(: ..dict_a, ..dict_b)` instead
+        // to create a dictionary, not an array.
+        let mut items_seen_are_spread_dicts = true;
```
laurmaedje: I think a name such as `all_dict_spreads` would be at least as clear while being a bit shorter

crates/typst-eval/src/code.rs
isuffix: I don't think it's unsound. We're not actually skipping any parts of the AST because `.all()` should exit early as soon as any array item isn't a spread, or produces an error or a non-dict. So there's never a time where we continue evaluating when we've already encountered an error. It's actually relatively conservative.

crates/typst-eval/src/code.rs
isuffix: How about `leaf_text`?

laurmaedje: could work.

alternatively maybe this?
- `as_text(&self) -> Option<&EcoString>`
- `to_text(&self) -> EcoString`
- `into_text(self) -> EcoString`

laurmaedje:
Thanks!

PR #7800, opened 2026-02-04
syntax: Allow specifying a preferred compiler version

tingerrr:
While this has currently no effect, it allows packages to already set a preferred compiler version before any breaking changes are introduced. This is integral for initial adoption of the compatibility feature, as introducing it with the breaking change would constitute a chicken and egg problem.

Consider a package targeting Typst 0.14, if Typst introduced a breaking change and the new manifest format in 0.15, then the package could not set the preferred compiler version to 0.14 without breaking itself for 0.14 also.

This would make it effectively a 0.15.0 targeting package without being able to use the new feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3326,6 +3326,7 @@ dependencies = [
  "serde_json",
  "tar",
  "tiny_http",
+ "toml",
  "typst-assets",
  "typst-bundle",
  "typst-library",
--- a/crates/typst-cli/src/eval.rs
+++ b/crates/typst-cli/src/eval.rs
@@ -10,7 +10,8 @@ use typst::foundations::{
 };
 use typst::routines::SpanMode;
 use typst::syntax::{
-    FileId, RangeMapper, RootedPath, Source, Span, SyntaxMode, VirtualPath, VirtualRoot,
+    FileId, PreferredCompilerVersion, RangeMapper, RootedPath, Source, Span, SyntaxMode,
+    VirtualPath, VirtualRoot,
 };
 use typst::text::{Font, FontBook};
 use typst::{World, engine::Sink, introspection::Introspector};
@@ -204,6 +205,13 @@ impl World for ExpressionWorld {
         } else {
             self.world.file(id)
         }
+    }
+
+    fn preferred_version(
+        &self,
+        root: &VirtualRoot,
+    ) -> FileResult<PreferredCompilerVersion> {
+        self.world.preferred_version(root)
     }
 
     fn font(&self, index: usize) -> Option<Font> {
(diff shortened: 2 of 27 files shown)


laurmaedje:
As [discussed on Discord](https://discord.com/channels/1054443721975922748/1457697035334717542/1468903578423922854), let's land this change and the actual compatibility behavior together and leave this as a draft for now.

tingerrr:
I've based #7620 on this and some work on my fork to see how one would use this for such a breaking change. It seems especially tough to do this for syntax related changes, because most of the lower level API in typst-syntax has no pre-existing context objects to attach this information to. It would also have to be tied to `Source` in someway to be used in reparsing, I think.

For anything between parsing and export a comemo-cached function using a `World` can be used to map `FileId` to `CompatOptions`. I'm unsure if we still have access to `World` on export, but this is probably easily added there too.

However, for syntax changes like string interpolation this is a little tougher, the main breaking point seems to be `World::file` and `World::source`. If you pass a flag like `string_interpolation: bool` down the syntax API, it inevitably bubbles up to these methods. And since you're inside `World` you likely can't construct a tracked instance without running into problems. Maybe tracking the inner file store/loader can solve this, but I don't know how recursive tracking interacts with comemo.

saecki:
I think the current plan is to introduce the compatibility behavior in `0.16`, but I think it would make sense to either land this, or a more general lenient manifest parsing mechanism with `0.15`.

tingerrr:
I'm fine with landing this on its own, I'll mark it as ready and you guys can decide what you prefer. If you make the parsing more lenient in some other way let me know and/or close this one so I can rebase the rest of the work on that.

tingerrr:
I think I arrived at a design that works quite well, but I'll need some more time to test it properly since I didn't write the string interpolation MVP with the ability to toggle it on and off in mind.

As of now, I went with a new method on `World`, `FileStore`, and `FileLoader` called `preferred_version` which maps `&VirtualRoot -> PreferredVersion`. Though, one might wonder if this should just return a manifest right away, since we throw away a lot of info otherwise.

Putting it on `FileStore` has the upside of making it available for parsing and for `World` to get in other stages, But this means we don't use comemo for caching the `root -> version` mapping, instead using slots similar to `typst_kit::files::FileSlot`.

laurmaedje:
Discussed this with @saecki. While there are some benefits to having 0.15 series compilers already recognize this syntax, I don't think it's too important as there are good alternatives (e.g. linting people to prefer `compiler = "0.15"` over `compiler = { minimum = "0.15" }` specifically for 0.15) and I would rather not rush the syntax change.

tingerrr:
Great, means I'm in no hurry!

PR #7809, opened 2026-02-04
Explicit error on both tagged-pdf and page ranges

### LINKED ISSUE #7743: Internal error when trying to export (tagged) PDF with a page range
### Description

When trying to export the project (link below) to a (tagged) PDF with a page range (either "current", or `1`, or `2`), I get the following errors:
1. `The compiler crashed 😢`
2. ```
   internal error: tree traversal didn't complete properly (occurred at /home/runner/.cargo/git/checkouts/typst-bdd39722d2f25244/a5cde71/crates/typst-pdf/src/tags/tree/mod.rs:78:9)
   ```

This seems to happen independently of the PDF version I choose.

The problem is no longer present when I switch to exporting the whole range / use quick export.

### Reproduction

Minimized reproducer:

```typ
#pagebreak()

a
```

I, too, think this is too small to be true, so just in case, here's the original source code as well

<details><summary>Code</summary>
<p>

````typ
#set enum(numbering: "a)i)")

== Aufgabe 8.2

#let comm_tbl = table.with(
  columns: 2,
  // align: center,
  stroke: (x, y) => (
    left: if x>0 {1pt},
    top: if y>0 {1pt}
  ),
  table.header[Sender][Nachrichteninhalt],
)

+ #comm_tbl()[
  UA ][`HELO example.com`][
  MTA][`250 haii`][
  UA ][`MAIL FROM: m.muster@meine-uni.de`][
  MTA][`250 sender <m.muster@meine-uni.de> ok`][
  UA ][`RCPT TO: komdat@syscom.aachen-rwth.de`][
  MTA][`250 recipient <komdat@syscom.aachen-rwth.de> ok`][
  UA ][`DATA`][
  MTA][`354 go ahead`][
  UA ][```
       MIME-Version: 1.0
       Content-Type: multipart/mixed;
       boundary="<BOUNDARY>"

       --<BOUNDARY>

       Content-Type: image/jxl
       Content-Transfer-Encoding: base64
       
       <DATEI>
       --<BOUNDARY>
       .
       ```][
  MTA][`250 ok: Message XXX accepted`][
  UA ][`QUIT`][
  MTA][`221 byee`
]

+ #comm_tbl()[
  S][`+OK POP3 server ready`][
  C][`user max`][
  S][`+OK`][
  C][`pass bmw`][
  S][`+OK user successfully logged in`][
  C][`list`][
  S][`1   498`][
  S][`.`][
  C][`retr 1`][
  S][`<message 1 contents>`][
  S][`.`][
  C][`dele 1`][
  C][`quit`][
  S][`+OK`
]

#pagebreak()

== Aufgabe 8.3

```
a) www.nehcaa.redivorp.de.   3600   IN   CNAME   nehcaa.redivorp.de.
b) git.nehcaa.redivorp.de.    200   IN   A       <an ipv4 addr>
c) nehcaa.redivorp.de.        600   IN   MX      10 mail.nehcaa.redivorp.de.
d) mail.nehcaa.redivorp.de.   600   IN   TXT     "v=spf mx ~all"
e) vpn.nehcaa.redivorp.de.    200   IN   A       <an ipv4 addr>
   vpn.nehcaa.redivorp.de.    200   IN   AAAA    <an ipv6 addr>
```

- Alle `Class` sind `IN`, da das der übliche Wert ist.
- TTL der Weiterleitung ist hoch gesetzt, da diese i.d.R. lange gültig bleibt.
- TTL des MX-Eintrages ist mittel hoch gesetzt, da die Domain des Mailservers sich i.d.R. nicht sehr oft ändert.
- Priorität des Mailservers ist auf den kleinstmöglichen Wert (höchste Priorität) gesetzt, da es z.Z. keine weiteren Server gibt.
- TTL der Git- bzw. VPN-Servers sind klein gesetzt, da die IP-Addressen sich öfters ändern können.
````

</p>
</details> 

### Operating system

Web app

### Typst version

### PR DESCRIPTION
**Pull request by saecki**
Closes #7743 by adding a check that emits a more descriptive error message. The original web-app issue that was closed as a duplicate should be reopened, and the web-app should gain a check that doesn't allow passing this combination to `typst-pdf`.

`typst-cli` already has a check for this in place: https://github.com/typst/typst/blob/bfe0a01df5df8a9dc58e53e9ad450f0deb8114e4/crates/typst-cli/src/compile.rs#L142

### DIFF
--- a/crates/typst-pdf/src/tags/mod.rs
+++ b/crates/typst-pdf/src/tags/mod.rs
@@ -3,10 +3,11 @@ use krilla::geom as kg;
 use krilla::page::Page;
 use krilla::surface::Surface;
 use krilla::tagging::{ArtifactType, ContentTag, SpanTag};
-use typst_library::diag::SourceResult;
+use typst_library::diag::{SourceResult, bail};
 use typst_library::layout::{FrameParent, PagedDocument, Point, Rect, Size};
 use typst_library::text::{Locale, TextItem};
 use typst_library::visualize::{Image, Shape};
+use typst_syntax::Span;
 
 use crate::PdfOptions;
 use crate::convert::{FrameContext, GlobalContext};
@@ -25,6 +26,10 @@ mod util;
 
 pub fn init(document: &PagedDocument, options: &PdfOptions) -> SourceResult<Tags> {
     let tree = if options.tagged {
+        if options.page_ranges.is_some() {
+            bail!(Span::detached(), "cannot enable tagged PDF and export a page range");
+        }
+
         tree::build(document, options)?
     } else {
         Tree::empty(document, options)
@@ -256,5 +261,29 @@ fn update_bbox(
         && gc.options.standards.config.validator() == Validator::UA1
     {
         bbox.expand_frame(fc, compute_bbox);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::num::NonZeroUsize;
+
+    use typst_library::layout::PageRanges;
+    use typst_utils::NonZeroExt;
+
+    use crate::tags;
+
+    #[test]
+    fn tagged_and_page_range() {
+        let options = crate::PdfOptions {
+            page_ranges: Some(PageRanges::new(vec![Some(NonZeroUsize::ONE)..=None])),
+            ..Default::default()
+        };
+        let res = tags::init(&typst_library::layout::PagedDocument::default(), &options);
+
+        assert_eq!(
+            res.err().unwrap().first().unwrap().message,
+            "cannot enable tagged PDF and export a page range"
+        );
     }
 }

### CODE REVIEW

File: crates/typst-pdf/src/tags/mod.rs [hunk 1]
```diff
@@ -25,6 +26,11 @@ mod util;
 
 pub fn init(document: &PagedDocument, options: &PdfOptions) -> SourceResult<Tags> {
     let tree = if options.tagged {
+        if options.page_ranges.is_some() {
+            panic_internal("cannot enable tagged PDF and export a page range")
```
laurmaedje: imo this should be a normal bail rather than an internal error since a consumer of `typst_pdf` can trigger this. an internal error should only be triggered by a bug in typst-pdf (or a bug in typst at large I guess).

ada4a:
Sorry for the unsolicited comment, but wouldn't it be safer long-term to disallow even creating `PdfOptions` with `page_ranges=Some(..)` and `tagged=true`? I see that `PdfOptions::new` performs a bunch of validation, so this rule could be enforced there as well.

Also, I'm not sure whether tagged PDFs are fundamentally incompatible with ranged exports, but if that is the case, one could consider going one step further and making the combination completely irrepresentible, by replacing the fields `page_range` and `tagged` with `tagged: Tagged`, with `Tagged` defined as follows:
```rs
/// <other docs>
///
/// Only the [No] variant has the option to specify a page range,
/// since tagged PDFs can only be created from the full document.
enum Tagged {
    Yes,
    No { page_range: Option<PageRange> },
}
```

saecki:
I've thought about this too, but since these options aren't conceptually related and this is rather a limitation of the Typst compiler, this would make creating the options more of a hassle for consumers of the `typst-pdf` crate.
There currently is no `PdfOptions::new`, but making the creation fallible would mean we either would have to introduce a `new` method that takes all arguments, and make the fields private, or add a builder. Which all felt like too much work, just to fail a little earlier...

ada4a:
Ah, sorry, misread `PdfStandards::new` as `PdfOptions::new`. Yes, introducing validation does make this more annoying than extending the pre-existing validation logic.

Thanks for the explanation!

Issue #6968, opened 2025-09-27
Sharp gradients overlap when printing


mxmerz:
### Description

I have a project where I use a sharp gradient with the colors `red, white, blue, white` repeated, and angle `45deg`. The PDF looks perfect, but when I print it, the red and blue stripes bleed into the white. At first I thought it was a printer issue, but the printer prints regular `rect` just fine, even rotated `rect`s.

PDF (web app project linked):
<img width="868" height="220" alt="Image" src="https://github.com/user-attachments/assets/96f6aa24-0ccb-4bab-9be3-f8c001cef7c0" />

Scan:
<img width="855" height="199" alt="Image" src="https://github.com/user-attachments/assets/c290f700-de79-4e39-a1a9-49611bd31a23" />

The problem also occurs with the example code from the documentation about [sharp gradients](https://typst.app/docs/reference/visualize/gradient/#definitions-sharp), where there is about a `1mm` wide strip between the "sharp" colors where they mix.

I use macOS, the problem occurs with both Preview.app and Adobe Acrobat Reader.

### Reproduction URL

https://typst.app/project/rGGLMy061Ordb9ndZuhciD

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

mxmerz:
Here is a web app link to the project I found the problem with, in case that helps with reproducing the error or debugging: https://typst.app/project/r85xqfZHS4ULVpMB8uf4f4

Enivex:
This problem may possibly already be resolved on main. Have you tested that?

mxmerz:
If that is what happens when you select "Development version" as the compiler in the web app, then yes, I tested that. Doesn't seem to make a difference.

Enivex:
Paging @LaurenzV

LaurenzV:
I mean, if I open the PDF on my PC it works fine in all viewers. Maybe a bug in the printer?

LaurenzV:
> At first I thought it was a printer issue, but the printer prints regular rect just fine, even rotated rects.

It could be that the bug in the printer simply occurs once you start using gradients, no?

mxmerz:
> I mean, if I open the PDF on my PC it works fine in all viewers.

Yeah, same for me, PDF looks perfect on screen in all viewers. The problem occurs only when I print the file, independent on which PDF viewer I use for printing.

> Maybe a bug in the printer? […] It could be that the bug in the printer simply occurs once you start using gradients, no?

[...]

Issue #6359, opened 2025-05-31
SVG Export size reductions


bluss:
### Description

Hi, I [recently ran `scour`](https://github.com/typst-community/rowmantic/commit/f865a7ed31c00346fb93efbfd76337a58061ef78) to reduce size of these typst outputs (to save space in typst/packages). The following are ideas for reducing size of SVG outputs:

- [ ] Use unique minimal ids
  - Typst 0.13.1 produces symbols with ids like `id="gAE8F30B35D6A8EB54F09046949FB529D"`and scour replaces it with `id="n"` and so on
  - This saved 8% file size on the two text filled tables
- [ ] Use less indentation
  - Typst 0.13.1 produces 4-space indentation, consider 0-space, 1-space or no whitespace at all
  - Going from 4-space to no whitespace saves 5-10% on the tested files.
- [ ] Use less coordinate precision
  - This is a lossy thing that scour does. It's probably not a good fit for Typst (?)

### Use Case

It would save some space and bandwidth

MDLC01:
At least part of this seems to be covered by https://github.com/typst/typst/pull/6247

Enivex:
For point 2, using tabs instead of spaces for indentation would save space sacrificing readability.

Dherse:
@Enivex iirc the IDs were indeed introduced because of incremental compilation (I want to say that they were introduced in one of the gradient PRs but frankly it's been too long). That being said, we might be able to rely on smaller hashes here? and manually detect collisions and do differentiation? That should work, right? Like use a 4-byte hash and then give them an extra byte if you have clashes 🤷‍♂

Enivex:
> @Enivex iirc the IDs were indeed introduced because of incremental compilation (I want to say that they were introduced in one of the gradient PRs but frankly it's been too long). That being said, we might be able to rely on smaller hashes here? and manually detect collisions and do differentiation? That should work, right? Like use a 4-byte hash and then give them an extra byte if you have clashes 🤷‍♂ 

Is file size important enough for this though? For those who truly need minimal svgs there's always external tools like the one mentioned above.

bluss:
Just punting this to external tools seems fair enough to me. I can see how none of the suggestions really work out, except for indentation.

Enter-tainer:
whitespace and long id can be well compressed by gzip if one want to serve these svg on web.

PR #7818, opened 2026-02-06
Fix excessive sampling of linear gradients in PDF and SVG

### LINKED ISSUE #6326: gradient.linear with sharp will alias depending on space used
### Description

When viewing a linear gradient inside a rectangle and using the sharp function, aliasing between the colors can show up if the `space` argument is used.

Version
```shell-session
$ typst --version
typst 0.13.1 (unknown hash)
```

Code

```typst
#rect(width: 100%, height: 5em, fill: gradient.linear(
  (oklch(51%, 0.208963, 29.2339deg)),
  (oklch(70%, 0.204259, 43.4910deg)),
  (oklch(81%, 0.168100, 76.7800deg)),
  (oklch(60%, 0.175100, 147.560deg)),
  (oklch(60%, 0.141400, 241.380deg)),
  (oklch(46%, 0.194300, 288.710deg)),
).sharp(6))

#rect(width: 100%, height: 5em, fill: gradient.linear(
  space: color.oklch,
  (oklch(51%, 0.208963, 29.2339deg)),
  (oklch(70%, 0.204259, 43.4910deg)),
  (oklch(81%, 0.168100, 76.7800deg)),
  (oklch(60%, 0.175100, 147.560deg)),
  (oklch(60%, 0.141400, 241.380deg)),
  (oklch(46%, 0.194300, 288.710deg)),
).sharp(6))
```

Screenshot from Zathura PDF viewer

Image

Screenshot from Okular PDF viewer

Image

For some reason this phenomena does not show up in typst.app.

### Reproduction URL

https://typst.app/project/rnzQ5cDvguAygjlMJI6Ti9

### Operating system

Linux

### Typst version

### LINKED ISSUE #7365: Excessive sampling of linear gradients in PDF
After rebasing my `test-suite` branch, I noticed some PDF tests getting noticable slower. I pinpointed it to this PR: #7326

This specific test now takes up twice the amount of time:
```typ
--- gradient-math-misc ---
// Test miscellaneous
#show math.equation: set text(fill: gradient.linear(..color.map.rainbow))
#show math.equation: box

$ hat(x) = bar x bar = vec(x, y, z) = tilde(x) = dot(x) $
$ x prime = vec(1, 2, delim: "[") $
$ sum_(i in NN) 1 + i $
$ attach(
  Pi, t: alpha, b: beta,
  tl: 1, tr: 2+3, bl: 4+5, br: 6,
) $
```

The gradient code in `typst-pdf` generates 31 samples between every stop of a gradient. Since the `rainbow` gradient already has a lot of stops, this seems to be the culprit. Maybe we can choose the number of samples based on the distance of the stops and or distance of the hue instead? In the rainbow case I think it would be fine to basically omit any extra sampled stops.

eltos:
### Description
**Commit a82dce9d52a83da6e9853b98e1e00d23a7697dbc**
Fixes duplicating stops at equal offset during interpolation, which caused sharp gradients in PDF to become blurry.
Also fixes duplicating stops in SVG where both stops of the (overlapping) window strides were added.
This results in visual changes (fixes) to the tests containing sharp gradients (see examples below).

- [x] Fixes #6326  
      Maybe also https://github.com/typst/typst/issues/6968, to be clearified  
	  Contributes to https://github.com/typst/typst/issues/6359

**Commit ba121d92ec65ff01840a1311db441b52f871b447**
Instead of always inserting a fixed number of in-between-stops (30 for PDF, 256/len for SVG), only insert actually required stops.
How many stops are required is determined by comparing the linearly rgb-interpolated color with the actual color at the midpoints, iteratively subdividing the distance between stops until the error falls below a threshold.
The threshold of 0.001 used results in no visible change. 
I also tested with 0.01 where there are very small almost unnoticable changes (see examples below). 
Since the step 0.001 to 0.01 reduces the size only little, I decided to use 0.001 in the end, but this can be adjusted if desired.

- [x] Fixes #7365
      Contributes to https://github.com/typst/typst/issues/6359

### Size statistics for `cargo testit gradient`

| | Before this PR | Duplication fix | **With th=0.001** | (With th=0.01)
|-|-|-|-|-|
| Stop count bloat factor in PDF export       |28.7   |16.2   |**15.0**   |14.6
| total size of PDF files with gradient tests |67.9 MB|63.2 MB|**2.9 MB** |2.7 MB
| Stop count bloat factor in SVG export       |30.2   |29.8   |**4.4**    |1.9
| total size of SVG files with gradient tests |13.1 MB|11.3 MB|**10.5 MB**|10.4 MB

_Stop count bloat factor denotes the number of stops after the interpolation, compared to the number of stops of the original gradient (mean of factors for all tests)_

### Changes (examples)

Changes due to sharp gradient fix:

gradient-linear-stroke-relative-parent first_commit
curve-stroke-gradient-sharp first_commit
issue-6680-gradient-linear-with-aspect-correction first_commit

---------

Changes in case of th=0.01:

grafik

grafik
grafik


LaurenzV:
I've mentioned this somewhere else, but in case it hasn't been tried, I would strongly suggest trying to use the [color](https://crates.io/crates/color) crate and the [gradient](https://docs.rs/color/latest/color/fn.gradient.html) method to create the optimal number and location of stops to approximate the color in SRGB. I don't know the theory behind it, but I think they use some more advanced approach to determine the best approximation for the given color space. This won't work for CMYK unfortunately, so we would need fallback behavior there, but I believe all other color spaces are supported.

eltos:
Thank you, I looked at how they do it and basically followed the same algorithm: bisecting the gradient until the Euclidean difference between the rgb-linearly interpolated midpoint color and the midpoint actual color between two stops falls below a threshold. I decided against using the color create to avoid adding it as dependency, to avoid the copying and type conversion necessary to call that method, to avoid the non-CMYK restriction and because in the end it's a rather simple algorithm.

LaurenzV:
Fair enough!

laurmaedje:
@eltos Is this ready for review? Since this and your other PR are still marked as draft.

eltos:
Not yet, there is a similar stop duplication behaviour in SVG export which I plan to cover, too. Please bear with me, I'll notify you once I come to it and the PR is ready.

laurmaedje:
Sure, no rush, it's just that sometimes people open a draft PR when they're actually waiting for a review, so I wanted to double check.

eltos:
@laurmaedje PR ready for review. SVG handling added and threshold lowered to 0.001, see details in updated PR description.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -819,6 +819,70 @@ impl Gradient {
...
 
+    /// Generates additional stops to replicate a gradient in
+    /// rgb color space interpolation by bisecting
+    pub fn generate_intermediate_stops_for_rgb_interpolation(
+        &self,
+        first: GradientStop,
+        second: GradientStop,
+    ) -> impl Iterator<Item = (Color, Ratio)> {
+        let threshold = 0.001;
+        let max_subdivisions = 64;
```
saecki: I think making these `const`s would improve readability. They could just stay here in the function.

eltos: done

saecki: Perfect, thanks!

laurmaedje:
Thank you!

PR #7820, opened 2026-02-07
Update test error annotations when passed `--update`

isuffix:
Update test error annotations so we can update them when passing `--update`!

There are four main changes that allow this:
1. Parsing annotations separately from the body to track line numbers that won't mutate as we write the body
2. Adding line/column info to note ranges to improve formatting so we display idiomatic ranges (e.g. `4-19` instead of `1:4-1:19` or `7` instead of `7-7`)
3. Local tracking (in `Note`) of which stages a test is seen in and its annotated/emitted/updated status
4. Pushing non-annotated notes and updating close matches in the notes vector so we have all annotated/emitted/updated notes in one container

As usual, this is best reviewed commit-by-commit. The earlier commits are generally smaller and serve to simplify/clarify moving annotation comment parsing into the new `notes.rs` file, to which the new annotation abstractions are added in the later commits.

<details><summary>
Here are some examples for checking <code>--update</code> or looking at range error messages
</summary>
<p>

```typ
--- testit-not-annotated eval ---
#let none = none

--- testit-not-emitted eval ---
// Error: 6-10 expected pattern, found `none`
// Hint: 6-10 keyword `none` is not allowed as an identifier; try `none_` instead
#let none_ = none

--- testit-wrong-message eval ---
// Error: 6-9 this message is wrong
#let 123 = none

--- testit-wrong-range eval ---
// Error: 12-16 expected pattern, found integer
#let 123 = none

--- testit-wrong-range-missing eval ---
// Error: expected pattern, found integer
#let 123 = none

--- testit-wrong-position eval ---
// Error: 3 expected pattern
#let

--- testit-wrong-kind eval ---
// Hint: 6-9 expected pattern, found integer
#let 123 = none

--- testit-wrong-line-implicit-extra eval ---
// Error: 6-9 expected pattern, found integer
// something
#let 123 = none

--- testit-wrong-line-implicit-swap eval ---
#let 123 = none
// Error: 6-9 expected pattern, found integer
// something

--- testit-wrong-line-explicit eval ---
// Error: 2:6-2:9 expected pattern, found integer
#let 123 = none
// something

--- testit-comparing eval ---
// Error: blank note
// Hint: 7-8 bad hint
// Error: 7 expected semicolon or line break
#{ 123 <abc> }
// Hint: 3-3 try wrapping your code in a markup block (`[ ]`)
// Hint: 7 labels can only be applied in markup mode

// Error: 7 expected semicolon or line break
// Error: 7 labels can only be applied in markup mode
// Hint: 7 try wrapping your code in a markup block (`[ ]`)
#{ 123 <abc> }
// something

#let not- annotated

--- testit-stages paged empty ---
// Try changing from `paged` to `eval`.
// Warning: 1-3 no text within stars
// Hint: 1-3 using multiple consecutive stars (e.g. **) has no additional effect
**

// Warning: 6-13 block may not occur inside of a paragraph and was ignored
#par(block())
// Warning: 18-25 block may not occur inside of a paragraph and was ignored
#show "oof": par(block())
oof
```

</details>

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -149,43 +149,70 @@ attributes are currently defined:
 
 There are, broadly speaking, three kinds of tests:
 
-- Tests that just ensure that the code runs successfully: Those typically make
-  use of `test` or `assert.eq` (both are very similar, `test` is just shorter)
-  to ensure certain properties hold when executing the Typst code. Generic
-  scripting tests that don't depend on the target should use the `eval`
-  attribute, if possible. If they rely on the evaluated content being compiled,
-  for example due to test code in show-rules, they should use `paged empty` or
-  `html empty`.
+1. Tests that just ensure that some code **runs successfully**:
+    ```typ
+    --- example-eval-test eval ---
+    // Test that `range` produces a normal array.
+    #test(type(range(1)), array)
+    #test(range(3), (0, 1, 2))
+    #test(range(3) + (3,), (0, 1, 2, 3))
+    ```
 
-- Tests that ensure the code emits particular diagnostic messages: Those have
-  inline annotations like `// Error: 2-7 thing was wrong`. An annotation can
-  start with either "Error", "Warning", or "Hint". The range designates the
-  code span the diagnostic message refers to in the first non-annotation line
-  below. If the code span is in a line further below, you can write ranges
-  like `3:2-3:7` to indicate the 2-7 column in the 3rd non-annotation line.
+    These typically call `test` or `assert.eq` (both are very similar, `test` is
+    just shorter) to ensure certain properties hold when executing Typst code.
 
-- Tests that ensure certain output is produced:
+    Most of these tests can use the `eval` attribute because they just check for
+    generic scripting behavior without depending on the compiled output itself.
+    But if a test does rely on the evaluated content being compiled beyond just
+    evaluation, for example to test code in show-rules, then it should use the
+    `paged empty` or `html empty` attributes.
 
-  - Visual output: When a test has the `paged` attribute, the compiler produces
-    paged output, renders it with the `typst-render` crate, and compares it
-    against a reference image stored in the repository. The test runner
-    automatically detects whether a test has visual output and requires a
-    reference image in this case.
 
-    To prevent bloat, it is important that the test images are kept as small as
-    possible. To that effect, the test runner enforces a maximum size of 20 KiB.
-    If you're updating a test and hit `reference output size exceeds`, see the
-    section on "Updating reference images" below. If truly necessary, the size
-    limit can be lifted by adding a `large` attribute after the test name, but
-    this should be the case very rarely.
+2. Tests that ensure the code **emits a particular diagnostic message**:
+    ```typ
+    --- example-diagnostic-test eval ---
+    // Test that parentheses without commas don't make arrays.
+    // Error: 4-19 cannot add array and integer
+    #{ (0, 1, 2) + (3) }
+    ```
 
-  - HTML output: When a test has the `html` attribute, the compiler produces
-    HTML output and compares it against a reference file stored in the
-    repository.
+    These have inline annotations like `// Error: 2-7 thing was wrong`. An
+    annotation can start with either "Error", "Warning", or "Hint". The range
+    designates the code span the diagnostic message refers to in the _first
+    non-annotation line_ below. If the code span is in a line further below, you
+    can write ranges like `3:2-3:7` to indicate the 2-7 column in the 3rd
+    non-annotation line.
 
-  - PDF tags output: When a test has the `pdftags` attribute, the compiler
-    produces a human readable tag tree formatted as a YAML document and compares
-    it against a reference file stored in the repository.
+    Similarly to 1, these tests should have either the `eval` attribute or a
+    pair like `paged empty` or `html empty`.
+
+3. Tests that ensure certain **output is produced**:
+    ```typ
+    --- example-visual-test paged ---
+    // Test how array values are displayed as content.
+    #range(3)
+    ```
+
+    - **Visual output:** When a test has the `paged` attribute, the compiler
+      produces paged output, renders it with the `typst-render` crate, and
+      compares it against a reference image stored in the repository. The test
+      runner automatically detects whether a test has visual output and requires
+      a reference image in this case.
+
+      To prevent bloat, it is important that the test images are kept as small
+      as possible. To that effect, the test runner enforces a maximum size of 20
+      KiB. If you're updating a test and hit `reference output size exceeds`,
+      see the section on "Updating reference images" below. If truly necessary,
+      the size limit can be lifted by adding a `large` attribute after the test
+      name, but this should be the case very rarely.
+
+    - **HTML output:** When a test has the `html` attribute, the compiler
+      produces HTML output and compares it against a reference file stored in
+      the repository.
+
+    - **PDF tag output:** When a test has the `pdftags` attribute, the compiler
+      produces a human readable tag tree formatted as a YAML document and
+      compares it against a reference file stored in the repository.
 
 If you have the choice between writing a test using assertions or using
 reference images, prefer assertions. This makes the test easier to understand


tests/src/collect.rs
```diff
@@ -678,37 +687,32 @@ impl<'a> Parser<'a> {
...
-                }
+            self.collector.tests.push(Test { name, attrs, body });
+        }
+    }
 
-                self.s.eat_until(is_newline);
-                if self.s.eat_newline() {
-                    self.line += 1;
-                }
+    /// Skips the preamble of a test.
```
laurmaedje: "of a test file" would be a bit clearer

tests/src/notes.rs
```diff
@@ -58,19 +44,45 @@ impl Display for NoteKind {
...
-        && let Ok(kind) = head.parse::<NoteKind>()
-    {
-        s.eat_if(' ');
-        Some(kind)
-    } else {
-        None
+    for (pattern, kind) in [
+        ("// Error:", NoteKind::Error),
+        ("// Warning:", NoteKind::Warning),
+        ("// Hint:", NoteKind::Hint),
```
laurmaedje, line 449: why this instead of the old `FromStr`?

tests/src/notes.rs
```diff
@@ -70,8 +90,48 @@ impl LineCol {
...
+    let mut body = String::with_capacity(full_body.len());
+    let mut annotated_line = 0;
+    let mut note_lines: Vec<(usize, usize, NoteKind, Scanner)> = vec![];
+    for (i, line) in full_body.lines().enumerate() {
+        let mut s = Scanner::new(line);
+        if let Some(kind) = parse_note_start(&mut s) {
+            // Notes annotate the next non-note line.
+            note_lines.push((i, annotated_line, kind, s));
+        } else {
+            body.push_str(line);
```
laurmaedje, line 419: could you elaborate on the idea behind excluding the annotations from the actual test source?

tests/src/notes.rs
```diff
@@ -64,6 +140,171 @@ impl Display for NoteKind {
...
+        stage: TestStages,
+        message: &EcoString,
+        span: Span,
+        world: &TestWorld,
+    ) -> Self {
+        let range = if let Some(id) = span.id()
+            && let Some(range) = world.range(span)
+        {
+            let source = world.source(id).unwrap();
+            let lines = source.lines();
```
laurmaedje: technically, it's a bit wonky if we create a `Source` for something like a JSON file even if it works. it would be a bit cleaner to go through `world.file(id)` and then `Bytes::lines`. note that if it's a source file, the `FileStore` will transparently reuse the backing storage and line information of the source file. and for other files it will also be memoized. 

c.f. also `typst-kit/src/diagnostics` (where I do this but with an extra cache for known sources because not all world implementors might have the nice bytes setup...).

tests/src/notes.rs
```diff
@@ -64,6 +140,171 @@ impl Display for NoteKind {
...
+
+/// The range of an annotated diagnostic as indices into a source file and as
+/// lines/columns for display.
+///
+/// We use an explicit enum instead of Option so we can implement Display.
+#[derive(Debug, Clone)]
+pub enum NoteRange {
+    None,
+    Some {
+        source: Source,
```
laurmaedje: having separate fields for `FileId` and `Lines` here also seems cleaner to me (see also https://github.com/typst/typst/pull/7820/changes/fb44d1f51396919018525f69240fbda2425fbef3#r2777548259)

tests/src/notes.rs
```diff
@@ -64,6 +140,171 @@ impl Display for NoteKind {
...
+        is_external: bool,
+        range: Range<usize>,
+        start: LineCol,
+        end: LineCol,
```
laurmaedje: technically, this could be a `Range<LineCol>`. might be worth considering.

tests/src/run.rs
```diff
@@ -247,54 +239,88 @@ impl<'a> Runner<'a> {
...
 
-            // Figure out if the diagnostic is only emitted in a specific in a
-            // specific output/target that isn't hit by all possible branches of
-            // the stage tree.
-            // See the doc comment on `TestStages` for an overview.
+            // Figure out if the diagnostic is only emitted in a specific
+            // output/target that isn't hit by all possible branches of the
+            // stage tree.
+            // E.g. If running Paged and the note was only emitted in Eval, it's
+            // still covered because Eval is required everything in Paged.
```
laurmaedje: something is off with the grammar here

tests/src/world.rs
```diff
@@ -79,38 +76,29 @@ impl World for TestWorld {
     }
 }
 
-impl TestWorld {
-    /// Retrieves line metadata for a file.
-    pub fn lines(&self, id: FileId) -> FileResult<Lines<String>> {
```
laurmaedje, line 82: related to https://github.com/typst/typst/pull/7820/changes/fb44d1f51396919018525f69240fbda2425fbef3#r2777548259: why remove this instead of just using it?

tests/README.md
```diff
@@ -149,43 +149,70 @@ attributes are currently defined:
...
+
+    These have inline annotations like `// Error: 2-7 thing was wrong`. An
+    annotation can start with either "Error", "Warning", or "Hint". The range
+    designates the code span the diagnostic message refers to in the _first
+    non-annotation line_ below. If the code span is in a line further below, you
+    can write ranges like `3:2-3:7` to indicate the 2-7 column in the 3rd
+    non-annotation line.
+
+    Similarly to 1, these tests should have either the `eval` attribute or a
+    pair like `paged empty` or `html empty`.
```
laurmaedje, line 187: `paged` or `html` without `empty` is also possible for warnings

tests/README.md
```diff
@@ -149,43 +149,70 @@ attributes are currently defined:
...
+    ```typ
+    --- example-visual-test paged ---
+    // Test how array values are displayed as content.
+    #range(3)
+    ```
+
+    - **Visual output:** When a test has the `paged` attribute, the compiler
+      produces paged output, renders it with the `typst-render` crate, and
+      compares it against a reference image stored in the repository. The test
+      runner automatically detects whether a test has visual output and requires
```
laurmaedje, line 199: while you're at it, could you expand this to also mentioned that `paged` will test SVG and PDF as well, via hashing?

tests/src/collect.rs
```diff
@@ -829,52 +835,62 @@ impl<'a> Parser<'a> {
...
         let mut file = None;
-        if self.s.eat_if('"') {
-            let path = self.s.eat_until(|c| is_newline(c) || c == '"');
-            if !self.s.eat_if('"') {
-                self.error("expected closing quote after file path");
-                return None;
+        if s.eat_if('"') {
+            let path = s.eat_until(|c| is_newline(c) || c == '"');
+            if !s.eat_if('"') {
+                return Err("expected closing quote after file path".to_string());
```
laurmaedje: what do you think about switching all these errors to `bail!` and `StrResult`? the final message would then be `EcoString` instead of `String` but that seems unproblematic to me

tests/src/collect.rs
```diff
@@ -989,8 +995,11 @@ pub struct TestParseError {
 }
 
 impl TestParseError {
-    pub fn new(pos: FilePos, kind: impl Into<TestParseErrorKind>) -> Self {
-        Self { pos, kind: kind.into() }
+    pub fn new(kind: impl Into<TestParseErrorKind>, path: &Path, line: usize) -> Self {
```
laurmaedje, line 779: what's the motivation for the change in signature?

tests/src/notes.rs
laurmaedje: I guess if the update is automatic, it could otherwise happen that the comment disrupts the test? e.g. if you somehow managed to get an error span within a raw block (not sure that's possible as text show rules don't seem to generate tracepoints).

tests/src/world.rs
```diff
@@ -79,38 +76,29 @@ impl World for TestWorld {
...
-}
-
 /// Shared foundation of all test worlds.
-struct TestBase {
+pub(crate) struct TestBase {
     library: LazyHash<Library>,
     book: LazyHash<FontBook>,
     fonts: Vec<Font>,
-    files: FileStore<TestFiles>,
+    pub files: FileStore<TestFiles>,
```
laurmaedje: I must admit I'm not a huge fan of having this half-public. It's not a big deal, but feels a bit off. I think I would already be more amenable to it if `files` was behind an accessor function rather than a plain `pub` field

tests/src/logger.rs
```diff
@@ -140,13 +163,27 @@ impl Logger {
...
+            selected,
+            passed,
+            failed,
+            skipped,
+            reports,
+            test_updates,
+            ..
+        } = self;
+
+        test_updates.into_par_iter().for_each(|(path, updates)| {
```
laurmaedje, line 176: with all it's doing here, maybe `Logger` is not the best name anymore, but not something that would necessarily have to change in this PR

laurmaedje (commented):
Seems to work great!

tests/src/collect.rs
isuffix: It's so that we can move the `Arc::new()` call into the function since most callers just have a `&Path` and not an `Arc<PathBuf>`. And for the callers that do have an `Arc` I try to just clone it and build the `TestParseError` manually, although that's not really necessary since test parse errors should be uncommon.

tests/src/notes.rs
isuffix: I'll switch to it. I don't know why but until I looked just now, I never realized that `Lines` is cheap to clone. I always thought it was just a String wrapper, but that's totally incorrect.

tests/src/notes.rs
isuffix: Removes a layer of indirection. Lets us remove the call to `is_id_continue`. Makes it very explicit exactly what syntax is allowed. Not great reasons, but just a little nicer.

tests/src/notes.rs
isuffix: The point is to remove a translation from line numbers in the old file to line numbers in the new file when updating. Only tracking line numbers in the test body's source removes that complication and makes me more confident in the updated code.

tests/src/collect.rs
isuffix: Love that!

tests/src/world.rs
isuffix: Because I was moving to use `Source` and wasn't thinking to directly store `FileId` and `Lines` from my previous misunderstanding: :sweat_smile: 
> I don't know why but until I looked just now, I never realized that Lines is cheap to clone. I always thought it was just a String wrapper, but that's totally incorrect.

isuffix:
This first push updates one comment and moves us to use `StrResult` with `bail!` for note parsing. Very happy with that suggestion :)

tests/src/notes.rs
isuffix: Also to avoid this issue where a line-offset range could point into another annotation:
```typ
--- example-test eval ---
// Error: 1:1-3:2 First
A
// Error: 1-2 Second
B
C
```
`First` is trying to annotate `A\nB\nC`, but actually annotates `A\n// Error: 1-2 Second\nB` because we don't remove `Second` from the test body and don't check for this case in the current code.

tests/src/world.rs
isuffix: I'm able to avoid it by using `FileId` and `Lines` instead of `Source`.

tests/src/world.rs
isuffix: Actually, with moving to `FileId` and `Source`, It really just isn't necessary. The place I would use it, `Note::emitted` now just reads:
```rust
let source = world.source(file).unwrap();
let lines = source.lines().clone();
```
which pretty much does the exact same thing. So we really just don't need this method.

isuffix:
This second push updates `NoteRange` to store `FileId` and `Lines` instead of `Source` and to use `Range<LineCol>` instead of two explicit `LineCol`. This let me remove the changes that require making `TestBase.file` public.

I also added a big comment explaining range syntax on `parse_line_col_range`. Though I may copy some of that into the readme.

tests/src/notes.rs
isuffix: Changed to:
```typ
// We don't call `.source()` because file may not be Typst code.
let lines = world.file(file).unwrap().lines().unwrap();
```

tests/src/run.rs
isuffix: I've rewritten it to this:
```rust
// Even if a diagnostic is emitted and annotated, it may only have
// been emitted in a subset of the ran stages. This may be an error,
// but only if that subset isn't "covered" by each ran stage.
// For example, if we have a `paged html` test that sees an error
// only in `eval`, this is fine because `eval` is covered by both
// `paged` and `html`. But we do not support errors occuring in
// only one of `paged` or `html` when both are given.
let ran_stages = self.test.attrs.implied_stages() & ARGS.required_stages();
// Whether the seen stages are implied by all ran stages.
let fully_covered =
    ran_stages.iter().all(|s| s.with_required().intersects(*seen));
if !fully_covered {
    consistent_set &= *seen;
    inconsistent_stages = true;
    let siblings = ran_stages & seen.with_siblings();
    log!(self, "only emitted in [{seen}] but expected in [{siblings}]");
}
```
@saecki I'm reasonably confident in this, but I would appreciate you double checking.

laurmaedje:
Thank you, great work!

PR #7822, opened 2026-02-07
Fix doc comment on `--open-report` and reorder before `--no-report` 

isuffix:
A small thing that seems to have slipped through.

For the ordering, I figure this way is more useful and a bit easier to understand.


laurmaedje:
Thanks!

PR #7823, opened 2026-02-07
Remove a `.clone()` and add a comment on eval tests

isuffix:
These changes were non-conflicting and easy to update together.

I added the comment because I had to convince myself of it by reading a good bit of surrounding code. Lmk if it's too much.


laurmaedje:
Huh nice. Tbh, @saecki and I didn't even realize that it's cached. I remember discussing that we're willing to accept the slowdown. :p

PR #7824, opened 2026-02-07
Fix figure alignment with floating placement

### LINKED ISSUE #7754: Figure placement set to auto centers caption and figure despite left alignment
This issue is explained in the end of:
[(https://forum.typst.app/t/how-to-align-figures-and-their-captions-to-the-left/2644/1)]

The following is not enough to left align figure and caption when figure placement is set to auto:

```
#show figure: set align(start)
#show figure: set block(width: 100%)
#show figure.caption: set align(left)

```

To override placement's centering the following workaround is needed. 
```
#show figure.where(kind: image): it => {
  set block(width: 100%)
  show image: set align(left)
  it
}
```

This is making things unnecessarily complicated and confusing. 

I would think this should override placement in terms of horizontal alignment:

```
#show figure: set align(start)
#show figure.caption: set align(left)

```

... which would also allow this to be easily changed later in the document. This is also valid for bottom and top configuration. Alignment only works (as I) expect for "none".

HollowNumber:
Fixes #7754. Necro of my previous pr which i closed due to erroneous code. My suggested fix is to add an internal field to `PlaceElem` (`read_horizontal_from_styles`) which signals whether to read horizontal alignment from the style chain during layout. This allows figures with `placement: auto` to respect show rule alignments (e.g., `#show figure: set align(start)`) while maintaining backward compatibility for plain `place(auto, float: true)` elements that default to center.

Added test `issue-7754` in `tests/suite/model/figure.typ` for verification.

### DIFF
--- a/crates/typst-layout/src/rules.rs
+++ b/crates/typst-layout/src/rules.rs
@@ -329,20 +329,24 @@ const FIGURE_RULE: ShowFn<FigureElem> = |elem, _, styles| {
 
     // Wrap in a float.
     if let Some(align) = elem.placement.get(styles) {
-        let alignment = styles.get(AlignElem::alignment);
-        let h_align = alignment.x().unwrap_or(HAlignment::Center);
-
         let full_alignment = match align {
-            // for auto placement, pass horizontal-only aligment
+            // For auto placement, pass Smart::Auto
+            // Horizontal alignment will be read from styles during layout
             Smart::Auto => Smart::Auto,
-            // For specific vertical placement, combine both
-            Smart::Custom(v_align) => Smart::Custom(h_align + v_align),
+
+            // For specific vertical placement, combine with horizontal alignment from styles
+            Smart::Custom(v_align) => {
+                let h_align =
+                    styles.get(AlignElem::alignment).x().unwrap_or(HAlignment::Center);
+                Smart::Custom(h_align + v_align)
+            }
         };
 
         realized = PlaceElem::new(realized)
             .with_alignment(full_alignment)
             .with_scope(elem.scope.get(styles))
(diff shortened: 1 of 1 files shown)


laurmaedje:
Thanks for the PR! 

Testing it, I get the feeling that the current behavior of `#place(auto, float: true)` is actually surprising and should be changed as well (backwards-incompatibly). There is no need for the internal field.

However, before I'm willing to make any change here, I would like to see a comprehensive overview over all the cases and how they behave before and after, to ensure the end result is truly consistent. If you're willing to put in the work to do this, we could try to move forward with this. Otherwise, I would close the PR and see when I find the time to do this myself.

HollowNumber:
Thanks! Just to confirm I'm understanding correctly: You would also want `place`  to respect set rules? Doing some local testing I can see that the same behaviour (mentioned in #7754) does indeed happen with place. I'd be happy to take a crack at extending this pr to a blanket solution :)

laurmaedje:
> Thanks! Just to confirm I'm understanding correctly: You would also want place to respect show rules?

In principle, yes, I _think_ I would also want it to respect alignment set rules (I assume you meant set rather than show).

But before any code changes would happen, I'd like to see in the PR thread a comment with examples that reasons through the different possible alignment cases of `place` (vertical/horizontal alignment present/missing/auto, etc.), how it reacts to alignment styles, and justification why the end result is consistent. This is what I what otherwise do myself and what I need to be confident in any particular change.

laurmaedje:
I'll close this for now due to inactivity. If you're still interested in working on this, it would be best to post a comprehensive design in an issue where we can discuss it and then we could have a new PR. Thanks still!

PR #7830, opened 2026-02-08
Add reference to `array.to-dict` in `dictionary` constructor docs

MeGaGiGaGon:
The `dictionary` docs currently have no references to `array.to-dict`, which is unfortunate for someone trying to convert an `array` to a `dictionary` for the first time.

I added the reference to the `dictionary` constructor docs, since to me the most likely pipeline for finding it would be trying `dictionary(..pairs)`, and on getting an error looking at the `dictionary` constructor docs, where you would find this new hint.

I did not add a reference to the overall docs since I did not see a clean place to fit it in, and unless you are specifically trying to convert an `array` to a `dictionary` (not sure how common that is) it would not be helpful to mention there.

### DIFF
--- a/crates/typst-library/src/foundations/dict.rs
+++ b/crates/typst-library/src/foundations/dict.rs
@@ -176,7 +176,7 @@ impl Dict {
     /// Note that this function is only intended for conversion of a
     /// dictionary-like value to a dictionary, not for creation of a dictionary
     /// from individual pairs. Use the dictionary syntax `(key: value)` instead.
-    /// Also see [`array.to-dict`].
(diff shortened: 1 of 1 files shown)


laurmaedje:
The addition seems a bit terse to me. If we add this, I would like to expand this to a full sentence, that does not leave the reader guessing what they'll find in the place where they should also look.

MeGaGiGaGon:
Sounds good, I originally left it short since to me `array.to-dict` says everything you would need inline (method on array for converting it to a dictionary), but I can see wanting it to be a full sentence.

laurmaedje:
Thanks!

PR #7836, opened 2026-02-09
Wait for images to be decoded before drawing image diff

saecki:
Fixes a bug where the images show in the initial viewport weren't displayed because the natural height/width of the images wasn't yet known and thus returned 0.


saecki:
Seems like a spurious CI failure

PR #7838, opened 2026-02-10
Convert diagnostic tests to use the eval target

isuffix:
This converts every test which has an annotated error that occurs only during `eval` to use that target instead of `paged`. This comes out to 709 tests total.

Note that many of the files are already eval-centric, e.g. `array.typ`, but some are more layout-centric with the surrounding tests generally using `paged`, e.g. `grid/footers.typ`. Let me know if we would prefer to have the layout-centric files still used `paged` for local consistency.

### DIFF
--- a/tests/suite/foundations/arguments.typ
+++ b/tests/suite/foundations/arguments.typ
@@ -14,12 +14,12 @@
 #test(args.at(2), 3)
 #test(args.at("a"), 2)
 
---- arguments-at-invalid-index paged ---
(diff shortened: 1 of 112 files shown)


laurmaedje:
Was just discussing this with @saecki yesterday.

In particular: should we perhaps _enforce_ an `eval` test if the eval stage already errored?

isuffix:
We probably should. Since Tobias already has a branch where this is a subset, I'll close this and let that go through

PR #7851, opened 2026-02-12
 Enforce eval attribute for tests that error in the eval stage 

saecki:
Continues #7838

Tests that error in the eval stage, will have have the `eval` attribute. This is in line with how tests that don't have error annotations are enforced to be `eval` tests or have the `empty` attribute.

### DIFF
--- a/tests/suite/foundations/arguments.typ
+++ b/tests/suite/foundations/arguments.typ
@@ -14,12 +14,12 @@
 #test(args.at(2), 3)
 #test(args.at("a"), 2)
 
---- arguments-at-invalid-index paged ---
+--- arguments-at-invalid-index eval ---
 #let args = arguments(0, 1, a: 2, 3)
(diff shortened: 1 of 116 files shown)


tests/src/run.rs
```diff
@@ -98,12 +98,12 @@ impl UnexpectedNonEmpty {
 enum UnexpectedEmpty {
     None,
     Output(TestStages),
-    Eval,
+    Eval(Result<(), ()>),
```
isuffix: I think it would be better to use a named boolean here instead of a Result.

isuffix:
Maybe not in this PR, but we should probably make it so that `--update` can also fix attribute issues like this (and maybe dangling result files?)

saecki:
> Maybe not in this PR, but we should probably make it so that `--update` can also fix attribute issues like this (and maybe dangling result files?)

Yeah, I've also thought about that :+1:

PR #7853, opened 2026-02-12
Update `World::today()` in tests to mimic CLI

t-rapp:
Create date values in tests with a chrono-based approach that is similar to the implementation in `typst-cli`.

Suggested in #7691.


t-rapp:
Have also added a commit that avoids the local timezone offset being added in CLI for reproducible builds. This was also mentioned in #7691. If considered better I can split this change into a separate PR.

laurmaedje:
A separate PR would be nice. Thanks for tackling this!

t-rapp:
Moved the commit which changes when local timezone is applied to #7856.

Regarding merging the implementations for `World::today()` in tests and CLI, and moving it into `typst-kit`: That would be even better but I did not know how to approach that (where to put what helper types).

laurmaedje:
Great thanks! We can take a look at a potential typst-kit design once these changes are landed.

laurmaedje:
Thanks!

Issue #6390, opened 2025-06-05
Make `math` size style functions actually override `show` rules


anormalperson8:
### Description

The functions `math.display`, `math.inline`, `math.script` and `math.sscript` allow manual configuration of the size of equation elements. However, the manual configuration does not override if there is an existing show rule.

Could there somehow be a style setting that keeps track of display style of a particular equation element so that show-set rules can be used instead, or that the style functions override everything including show rules?

### Use Case

```typ
#show math.frac: math.display // Show rule for fractions to be in display style

$1/2$ // A 1/2 fraction in display style

$inline(1/2)$ //  A 1/2 fraction still in display style
```

```typ
$1/2$ // A 1/2 fraction in inline style

$display(1/2)$ //  A 1/2 fraction in display style
```

MDLC01:
What I think is happening here is that the inner-most function takes precedence. Your first example can be "desugared" to the following:

```typ
$display(1/2)$ // A 1/2 fraction in display style

$inline(display(1/2))$ //  A 1/2 fraction still in display style
```

In the second case, the inner call takes precedence.

To be clear, I'm not saying this is good or bad. Just that i believe this is what is happening.

Enivex:
That is what's happening, and it's consistent with how typst works overall.

It seems to me that what you want is this https://github.com/typst/typst/issues/420 ?

anormalperson8:
Perhaps? I just find it weird that the `inline`, `display`, `script` and `sscript` is overridden by the show rule, even though intuitively I'm applying a layer of size style in the `$ ... $` clause.

Another thought: Would it be more convenient if this can be done via show-set rules?

PR #7855, opened 2026-02-13
Add `math.var` so that text in math mode uses the text font

### LINKED ISSUE #366: Text in math mode is not using the text font
I became aware of this while trying to understand https://github.com/typst/typst/issues/355

If I change the text font to "Comic Neue" (Only to get a distinctly different font)

`Hello world, $"Hello world"$, $#text(font: "Comic Neue")[Hello world]$`

will render as

image

This is not what I expect. In LaTeX, the input

`Hello world, $\text{Hello world}$`

will indeed render as

image,

correctly using the text font when displaying text in math mode.

----------

Why is this the correct behavior? Because it is common to write something like

`And therefore, if
$ f(a) > 0 quad "and" quad f'(a) > 0, $
we have`

where the text is part of the, well, text. When this is typeset using the math font it looks very awkward.

image

### LINKED ISSUE #1125: [RFC #6] Clearer syntax for displaying text/vars in math

### LINKED ISSUE #4303: Script function is ignored in text mode
### Description

Hello,
today I encountered an [issue](https://github.com/ludwig-austermann/typst-ouset/issues/1) with one of my packages. Essentially, the error occurred because of typst v0.11 and v0.10 behave differently. The following code
```typst
#set text(font: "Cascadia Code")

Hello #math.script("World") $script("World")$ #math.script($"World"$)

#math.frac("hello", "world") $"hello"/"world"$
```
yields:
| v0.11.1 | v0.10.0
|----|----|
| ouset test | ouset test(1) |

Notice, how `math.script` does not imply math mode, whereas `math.frac` does.

Either way, some behaviour would be wrong. In my opinion, the situation before v0.11 is correct, as it allows easier access to math mode — also inside code environments.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

### LINKED ISSUE #5081: Italic correction does not work with multi-letter italics in math
### Description

Reproduction (on 0.11.1):

```typ
$ P' $
$ italic("P")' $
$ italic("PP")' $
```

The first two lines are as expected. On the last line, the prime symbol appears too close to the second P.

Feel free to suggest a better title because I'm not sure how to describe the issue precisely.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

### LINKED ISSUE #5818: Script style not applied in new math text
### Description

When replacing `upright(B)` with just `"B"`, script or script-script style (with widened characters) is not applied. See for example

```typ
$ upright(B) "B" wide
  script(upright(B)) script("B") wide
  sscript(upright(B)) sscript("B") $
```

Image

This was already the case previously (with multi-character strings), but I only just noticed when replacing `upright` calls, so I'm not entirely sure if that's a bug or intended.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

### LINKED ISSUE #6737: Problems on setting CJK/Han font in math equations

mkorje:
The main change is the addition of a `math.var` element, which single letters symbols in math are automatically wrapped in. As a consequence, string literals (`"..."`) and content blocks (`#[...]`) in math use the body text font (in particular, equation numbers use the body font).

## Q&A

**(Subject to change)**

<ol>
<li>
How do I set the math font?

```typ
#show math.var: set text(font: "math font")
```
</li>
<li>
How do I set the text font in math?

```typ
#show math.equation: set text(font: "text font")
```
Note that it defaults to the body text font.
</li>
<li>
How do I use the body text font in math?
<br>
Use a string literal or content block:

```typ
$ "in text font" #[also in text font] $
```
</li>
<li>
How do I type a multi-character variable?

```typ
$ var("blah") $
// or if you use it often
#let blah = math.var("blah")
$ blah $
```
</li>
<li>
How do I type an upright variable?

```typ
$ upright(A) var(upright("blah")) $
```
</li>
<li>
What is the default font for <code>math.op</code>?
<br>
The body text font - this matches LaTeX.
</li>
<li>
How do I set the font for <code>math.op</code>?

```typ
#show math.op: set text(font: "font")
```
</li>
<li>
How do I use the math font for a specific operator?
<br>
Use <code>math.var</code> with <code>math.op</code>:

```typ
#let lim = math.op(math.var("lim"), limits: true)
```
</li>
<li>
How do I (a CJK user) set the font with <code>"latin-in-cjk"</code> covers?
<br>
If you only want to use CJK text in strings and content in math: if you've set the body font to a CJK font everything will just work.
<br>
Otherwise: do the same thing as before (as described <a href="https://typst-doc-cn.github.io/guide/FAQ/equation-chinese-font.html">here</a>).
</li>
</ol>
 
## Issues

- Closes #366
- Closes #1125
- Closes #6737
- Closes #5081 (no longer applicable)
- Closes #5818 (no longer applicable)
- Closes #4303 (reduced to being a duplicate of #6390)

## To-Do

- [ ] Discuss remaining failing tests
- [ ] Update docs


MDLC01:
It's great to see some movement in this area! Thank you for your continued work on math.

> How do I type an upright variable?

Does that mean `math.var` does not automatically make multi-letter variables upright by default?

> What is the default font for math.op?
> The body text font - this matches LaTeX.

I there a reason why this is the case, other than being LaTeX's default? To me, it seems that operators are "mathy" enough to be typeset in a math font.

> Closes #5081 (no longer applicable)

Why is it no longer applicable? Wouldn't https://github.com/typst/typst/issues/5081#issuecomment-2684645904 reproduce the issue?

Enivex:
> I there a reason why this is the case, other than being LaTeX's default? To me, it seems that operators are "mathy" enough to be typeset in a math font.

Math fonts aren't designed to typeset runs of letters

> The body text font - this matches LaTeX.

I'm not sure this is _quite_ true. It does use a text font, but I believe the text font used for regular text and math operators can technically be different. Sometimes you want to use a font which more closely matches the actual math font.

winstontsai:
>     6. What is the default font for `math.op`?
>        
>        The body text font - this matches LaTeX.

Not sure about 6. I think the text used for math operators should correspond more closely to `\mathrm` in LaTeX or [`upright`](https://typst.app/docs/reference/math/styles/#functions-upright) in Typst, which is math font. Not the same as body text.

sijow:
> How do I type an upright variable?
> `$ upright(A) var(upright("blah")) $`

I suppose `upright(var("blah"))` will also work? (It seems more logical to me and more consistent with `#let lim = math.op(math.var(...))`)

MDLC01:
> > I there a reason why this is the case, other than being LaTeX's default? To me, it seems that operators are "mathy" enough to be typeset in a math font.
> 
> Math fonts aren't designed to typeset runs of letters

But what if I want to display an operator in bold or sans? Text font aren't designed to typeset mathematical letters. Multi-letter `var`s are typeset using a math font with this PR so using a math font to display runs of letters will still be common.

Is there a technical reason why math fonts can't typeset runs of letters? Or is it just that ligatures, kerning, and such are often only present in text fonts? If it's the latter, I suspect this would be mainly due to the precedent set by LaTeX and we should just ask math font authors to clone metrics and ligatures from the corresponding text font.

peng1999:
> Not sure about 6. I think the text used for math operators should correspond more closely to `\mathrm` in LaTeX or [`upright`](https://typst.app/docs/reference/math/styles/#functions-upright) in Typst, which is math font. Not the same as body text.

`\mathrm` is usually a text font, not math font. Here is a quick survey of common LaTeX defaults:

Document class | Math mode normal font | \mathrm font | main text roman font
-- | -- | -- | --
article (using pdfLaTeX) | Computer Modern Math fonts: letters=Computer Modern Math Italic (cmm)，symbols=Computer   Modern Symbols (cmsy)，largesymbols=Computer Modern Extended (cmex) | Computer Modern Roman (cmr) | Computer Modern Roman(cmr)
article + unicode-math (XeLaTeX/LuaLaTeX, default recommended settings) | Latin Modern Math | Latin Modern Roman | Latin Modern Roman
IEEEtran.cls | Computer Modern Math fonts, same   as article | Computer Modern Roman (cmr) | Times Roman (ptm)

Therefore, using the body text font for math.op seems a reasonable default, and it is also consistent with how unicode-math treats \mathrm.

sijow:
From this table (especially the last row) it seems that care is usually taken to have the `\mathrm` font consistent with the math font, irrespective of the main text font, as @Enivex suggested might be done.

So if the user changes the main font, but keeps the default math font (New Computer Modern Math), the best might be to keep New Computer Modern for math text such as operators.

jhvoigts:
> Therefore, using the body text font for math.op seems a reasonable default, and it is also consistent with how unicode-math treats \mathrm.

I am unsure about this. Having a different font family for operators and variables seems very out of place to me. Even in the example you provided, `\mathrm` always uses a font from the same family, while with IEEEtran.cls, the main text font is a completely different font family.
At least in my quick testing with `pdflatex`, after loading a different text font, `\mathrm` matches the operator font but not necessarily the main text font. See this minimal example:
```tex
\documentclass{article}

\usepackage{amsmath}
\usepackage{palatino}

\begin{document}
\begin{align*}
	\cos(x) \\
	\mathrm{cos}(x) \\
	\text{cos}(x)
\end{align*}
\end{document}
```
image

Thus, with Typst's default settings, I would expect that operators and variables use NCM, while text inside math uses Libertinus.

Also, do the CM and LM Math fonts even contain the upright character variants? If not, then that explains why the roman fonts are used.

sijow:
So it looks like we have (without manual override of the `math.op` font):

- The font for body text
- The font for math symbols and multi-letter variables
- The font for math operators and for text in math

while LaTeX has

- The font for body text and for text in math (with `\text` or `\textrm`)
- The font for math symbols
- The font for math multi-letter variables and operators (with `\mathrm`, typically used also by `\DeclareMathOperator`)

I think I find the LaTeX division more logical: sentences in math should use the body font, and `var("speed")` should default to upright and probably the same font as operators. This should be a text font that can be set separately from the body font (though like @MDLC01 I wonder if this could eventually be done with a math font...)

winstontsai:
What font would/should numbers use? Should they match the operator font?

jhvoigts:
> What font would/should numbers use? Should they match the operator font?

They should match the math/operator font to remain visually consistent. In my LaTeX example above, `1234` uses CM and `\text{1234}` uses Palatino.
In general, everything mathy should match while actual text should remain in its own font. That includes text that happens to intertwine with the math.

jhvoigts:
What I was just wondering about: What about numbers in math? Should they maybe also be their own element, e.g., `math.num`?

PR #7856, opened 2026-02-13
Avoid local timezone offset being added for reproducible builds

t-rapp:
Always use UTC timezone in the CLI implementation of `World::today()` when a fixed date is requested. Only apply the local timezone offset when the date is not fixed, and no offset argument is set.

Mentioned in #7691.


laurmaedje:
Thank you!

Issue #6597, opened 2025-07-13
Gradient fill does not respect negative heights for `rect` with PDF export


Mc-Zen:
### Description

```typ
#set rect(fill: gradient.linear(angle: 90deg, ..color.map.viridis))

#rect()\
#rect(height: -1cm, )
```
The result with SVG and PNG export is 

<img width="110" height="132" alt="Image" src="https://github.com/user-attachments/assets/3b40f286-b0cd-43c5-9d8d-1dc0d273b956" />

but with PDF it's 

<img width="115" height="142" alt="Image" src="https://github.com/user-attachments/assets/ec2cedfc-e836-4eb3-a51e-072c6db62ce8" />

This only happens with negative height, not with negative width (and `angle: 0deg` of course). 

### Typst version

- [x] I am using the latest version of Typst

LaurenzV:
I tried it with the newest development version and it seems to work there. Can reproduce it with 0.13.1, though.

Mc-Zen:
Nice

Mc-Zen:
I think this is broken again. It works in the development version of the web app but not on main.

LaurenzV:
It seems you are right, bisecting.

LaurenzV:
Ouch, seems like it was https://github.com/typst/typst/commit/2809dac6a783318c1b07e237a0fba284c445fee2. We really need tests for the PDF export. :/

LaurenzV:
I'm pretty sure the fix is correct though and it's just pure chance that it worked before: https://github.com/typst/typst/blob/c7938e93dcd63464912345c3d7363f5961c2f47d/crates/typst-pdf/src/shape.rs#L82-L87

Here is where we transform negative rectangles, but we also need to apply the transform to the paint, not just the shape outline.

Andrew15-5:
> Here is where we transform negative rectangles, but we also need to apply the transform to the paint, not just the shape outline.

https://github.com/typst/typst/blob/02cd1c13de50363010b41b95148233dc952042c2/crates/typst-pdf/src/paint.rs?plain=1#L180-L195

This? The `y` already carries the negative value, so I assume the necessary transformation is already being applied.

Andrew15-5:
@LaurenzV, do you have any hints?

LaurenzV:
I think size is always positive? But yes somewhere there you probably need to apply a negation depending on the sign of the with/height of the original rectangle.

Andrew15-5:
[...]

Issue #2282, opened 2023-10-01
Gradient - Tracking issue


Dherse:
# Gradient progress report:
- [x] Part 1: Color rework #2171 
- [x] Part 1a: Color fixes #2262 
- [x] Part 2: Linear gradients #2279
- [ ] Part 3: Radial gradients
- [ ] Part 4: Conic gradients
- [ ] Part 5: Gradients on text
- [ ] Part 6: Additional gradient features

The vast majority of those features (all of them in SVG) are available on my [gradient](https://github.com/Dherse/typst/tree/gradient) branch. However, in this branch the layout engine is completely broken, which I have fixed in this PR, but didn't bother to on my experimental branch.

This is just a tracking issue I can point to in other gradient PRs ;)

Dherse:
Also closes #443 (with Part 8).

Dherse:
Also closes #260. I'd even go as far as stating that it already closes it.

Dherse:
It was decided in discussion with @laurmaedje that we would not support (at least now) glyph-by-glyph gradients as they're extremely heavy file-size wise, kind of ugly, and difficult to get right.

stellarpower:
This looks fab! Would building it be as simple as pointing rust to a crate from your fork? Afraid I've not much experience with the language. I would like some gradients in my document, so I can do some basic testing.

laurmaedje:
@stellarpower Most of the gradient work has already landed on main, so if you just compile Typst from main you can just use it.

stellarpower:
As of a5966636, it seems that I can fill a box with a gradient (`fill` can take a `gradient`) - but the `stroke` for `line`s doesn't seem to support this; is this worth adding as a subitem above? I imagine single straight lines ought to be a simpler starting point than arbitrary paths or font glyphs, and would enable people to take it for a spin in more situations!

Dherse:
@stellarpower With regards to line, they generally you work, you can find an example [here](https://github.com/typst/typst/blob/main/tests/typ/visualize/gradient-stroke.typ) but I generally recommend keeping simple linear gradients on line as line technically have a zero width which makes other gradient look... funky?

>  I imagine single straight lines ought to be a simpler starting point than arbitrary paths or font glyphs

Truth be told, there is essentially no difference wrt. the "complexity" of the stroke for gradient fills. Currently, gradient fills in parent mode will work on any path you throw at them, it's only when applied in `relative: "self"` or `relative: "auto"` on lines that it can get tricky.

[...]

PR #7863, opened 2026-02-13
Fix gradient stroke for line and curve

### LINKED ISSUE #5647: Path stroke gradients do not work as expected
### Description

Tested with `typst 0.12.0 (dhb06dbf9)`. On current master, drawing gradient strokes with `path` got fixed (with the new curve API): it is possible to stroke straight lines (angle 0deg) with a gradient now. But drawing the path flipped, still does not flip the gradient.

The following code:
```typst
#let red-blue = gradient.linear(red, blue, relative: "self")

// Ok
#path((0cm, 0cm), (5cm, 0cm), stroke: red-blue + 5pt)

// Expected blue to red gradient, got red to blue
#path((5cm, 0cm), (0cm, 0cm), stroke: red-blue + 5pt)
```

Results in this rendering:
grafik

### Reproduction URL

_No response_

### Operating system

Web app, Linux

### Typst version

### LINKED ISSUE #5705: Border Stroke Gradient Rendering Issue
### Description

The rendering for gradient stroke rendering is currently wrong for right and bottom

```typst
#rect(
  width: 100pt,
  height: 100pt,
  radius: 20pt,
  inset: 0pt,
  fill: black,
  stroke: (
    top: gradient.linear((gray, 0%), (black, 75%), (black, 100%), angle: 90deg) + 3pt,
    right: gradient.linear((gray, 0%), (black, 75%), (black, 100%), angle: 180deg) + 3pt,
    bottom: gradient.linear((gray, 0%), (black, 75%), (black, 100%), angle: 270deg) + 3pt,
    left: gradient.linear((gray, 0%), (black, 75%), (black, 100%), angle: 0deg) + 3pt,
  )
)
```

Image

### Reproduction URL

_No response_

### Operating system

Windows

### Typst version

### LINKED ISSUE #6068: Gradients on `curve` elements don't use the correct bounding box when last segment is `curve.line`

eltos:
Fixes issues with gradient strokes for lines and curves:
- Take the stroke width into account when calculating the bounding box to avoid zero-sized boxes
- Avoid mirroring the gradient for negative bounding box sizes when the angle is outside the first quadrant
- Fixes #6068
- Fixes #5705
- Closes #5647 as "expected behaviour", since the gradient is now always drawn according to the angle specified. The previous behaviour of line strokes using a mirrored gradient when end is left of start was IMHO confusing, especially for gradient angles other than 0deg, and anyhow not consistently applied to paths/curves. See output of _gradient-line-stroke_ test below for the comparison of this behaviour before/after.

### Tests added

_Left / right shows before / after this PR. PDF screenshots are from Adobe Acrobat (report.html does not render them correctly with my firefox)_

**issue-6068-curve-stroke-gradient**
PNG
grafik
SVG
canvas
PDF
canvascanvas

**issue-5705-gradient-border-stroke paged**
PNG

SVG

PDF

_In the test images below, each "star" consists of lines/curves at different angles, all using the same stroke gradient._

**gradient-line-stroke**
PNG

SVG

PDF
grafik

**gradient-curve-stroke**
PNG

SVG

PDF
canvas

**gradient-curve-stroke-quad**
PNG

SVG

PDF
canvas

### DIFF
--- a/crates/typst-library/src/visualize/curve.rs
+++ b/crates/typst-library/src/visualize/curve.rs
@@ -487,11 +487,6 @@ impl Curve {
             Point::new(Abs::raw(bbox.x1), Abs::raw(bbox.y1)),
         )
     }
-
-    /// Computes the size of the bounding box of this curve.
-    pub fn bbox_size(&self) -> Size {
-        self.bbox(None).size()
-    }
 }
 
 impl Curve {


crates/typst-library/src/visualize/curve.rs
```diff
@@ -547,6 +523,10 @@ impl Curve {
     /// When this curve is stroked with `stroke`, would the stroke contain
     /// `point`?
     pub fn stroke_contains(&self, stroke: &FixedStroke, needle: Point) -> bool {
+        Shape::contains(&self.to_kurbo_stroke(stroke), point_to_kurbo(needle))
```
saecki: Could you prefix `Shape` with `kurbo::` for consistency and use an underscore import: `use kurbo::Shape as _`, so it's only to bring the trait methods in scope.

crates/typst-pdf/src/paint.rs
```diff
@@ -65,13 +65,31 @@ fn convert_paint(
...
+    let mut size = Size::zero();
+    let mut offset = Point::zero();
+    if let Some(s) = shape {
+        size = s.geometry.bbox_size();
+
+        // Edge cases for strokes.
+        if matches!(s.geometry, Geometry::Line(..) | Geometry::Curve(..)) {
+            (offset, size) = s.geometry.bbox_size_with_stroke(s.stroke.as_ref());
+            // avoid mirroring gradient if line angle results in negative sizes
+            if size.x.signum() < 0.0 {
```
saecki: ~Does the `signum` handle a case I'm not aware of? Negative inifinity and signed `-0.0` should also be handled by a direct comparison.~

With the changes to `bbox_size_with_stroke` these extra checks could be completely removed.

crates/typst-library/src/visualize/shape.rs
```diff
@@ -423,4 +424,32 @@ impl Geometry {
             Self::Curve(curve) => curve.bbox_size(),
         }
     }
+
+    /// The bounding box offset and size
+    /// taking the stroke width of the shape into account
+    pub fn bbox_size_with_stroke(&self, stroke: Option<&FixedStroke>) -> (Point, Size) {
```
saecki: I feel like this function should be renamed to `bbox_with_stroke` and return a `Rect`. Since it's a bounding box, the size should never be negative, and that would remove the extra code path in `typst-pdf` to avoid mirrored gradients.

crates/typst-library/src/visualize/shape.rs
```diff
@@ -423,4 +424,32 @@ impl Geometry {
...
+        let bbox = self.bbox_size();
+        let stroke_width = stroke.map(|s| s.thickness).unwrap_or(Abs::zero());
+        match self {
+            Self::Line(line) => {
+                let a = Angle::atan2(line.y.to_raw(), line.x.to_raw());
+                let padding = Size::new(
+                    line.x.signum() * stroke_width * 0.5 * a.sin().abs(),
```
saecki: Does this handle square stroke caps?

crates/typst-pdf/src/text.rs
```diff
@@ -33,16 +32,17 @@ pub(crate) fn handle_text(
...
-            paint::convert_stroke(gc, s, true, surface, fc.state(), Size::zero())
-        }) {
-            Some(stroke?)
-        } else {
-            None
-        };
+    let stroke = if let Some(stroke) = t
+        .stroke
+        .as_ref()
+        .map(|s| paint::convert_stroke(gc, s, true, surface, fc.state(), None))
```
saecki: I think moving the code from the `map` inside the if let would be more readable.

crates/typst-library/src/visualize/shape.rs
eltos: This could be changed, but the other bbox methods also return negatively sized bboxes. We should keep it consistent, shouldn't we?

Changing this would also disable the mirroring behaviour for negatively sized rects, see #6597. I'm fine with that, but it should be considered. What do you think?

crates/typst-pdf/src/text.rs
eltos: I only changed `Size::zero()` for `None` here, rest was `cargo fmt`. But I can do that as refactor.

saecki: Yeah that would be appreciated :)
The line-wrap heuristics of `rustfmt` are a bit unfortunate sometimes.

crates/typst-library/src/visualize/shape.rs
saecki: > This could be changed, but the other bbox methods also return negatively sized bboxes. We should keep it consistent, shouldn't we?

I think the only other `bbox` and `bbox_size` methods are for `Curve` and there `bbox_size` is a wrapper of the `bbox` method, which always computes a rectangle with positive dimensions. 

> Changing this would also disable the mirroring behaviour for negatively sized rects, see https://github.com/typst/typst/issues/6597. I'm fine with that, but it should be considered. What do you think?

Hmm I see, currently this seems to be handled a little bit differntly in every export. The whole concept of negative sizes for bounding boxes seems to be somewhat of a pain point.

Maybe we can move the logic into `typst-library` and  introduce a `BBox` type to make this more robust. It could have a `Point` and `Size` internally and explicit accessors to either get a positive `Rect` with `min` and `max` coordinates, or the possibly negative `Size` and the offset `Point`.
For `Curve` the bounding box is already always positive, for `Line` we could just use this code, the other ones would stay the same https://github.com/typst/typst/blob/88356d0b708393dc96d8dd6ac88d96cc6e190e06/crates/typst-library/src/visualize/shape.rs#L404

Then we could get rid of the `bbox_size`, and looking at it more closely, I think we should remove `bbox_size_with_stroke` and instead make `Geometry::bbox()` take an optional `stroke` parameter, like you added for `Curve::bbox()`. Finally we could add a `Shape::bbox()` method, which uses the shape's `stroke` field, since that seems to always be passed anyway.

This refactor can be done in a follow-up PR, it just feels like this would remove a lot of the complications.

saecki: Or maybe we could remove handling of negative dimensions alltogether, depending on the outcome of https://github.com/typst/typst/issues/7901

crates/typst-library/src/visualize/shape.rs
eltos: I have update this to use kurbo in order to handle both square and round caps. Also added a test for it:

crates/typst-library/src/visualize/shape.rs
eltos: > Changing this would also disable the mirroring behaviour for negatively sized rects

I was wrong, this behaviour originates from `bbox_size` not `bbox_size_with_stroke`. Hence, I applied the rename, signature change and merge with `bbox` as you initially proposed. Further changes touching `bbox_size` can be kept for another PR, depending on outcome of #7901.

crates/typst-pdf/src/paint.rs
eltos: done

eltos:
@saecki ready for 2nd review

crates/typst-pdf/src/paint.rs
```diff
@@ -65,23 +65,35 @@ fn convert_paint(
...
-    // Edge cases for strokes.
-    if size.x.is_zero() {
-        size.x = Abs::pt(1.0);
-    }
+    let mut bbox = if let Some(s) = shape {
+        match s.geometry {
+            Geometry::Line(_) | Geometry::Curve(_) => s.geometry.bbox(s.stroke.as_ref()),
+            // Special handling for fill of rectangles (mirrors gradients for negative sizes)
+            Geometry::Rect(_) => {
+                Rect::from_pos_size(Point::zero(), s.geometry.bbox_size())
```
saecki: The documentation on `Rect` says that `min` is always smaller than `max`, and the `from_pos_size` function doesn't check that. Maybe this could return a `Point` and `Size` tuple for now, so that we don't misuse the documented API of `Rect`.
The same also applies to the code in `typst-svg`.

saecki: ~~I've just noticed that by not passing the stroke, the bbox will be wrong for rectangles created by the `typst_layout::shapes::simple_rect` function such as:~~

```typ
#rect(
  width: 100pt,
  height: 100pt,
  stroke: gradient.linear(green, black, yellow, angle: 90deg).sharp(8) + 20pt,
)
```

~~The gradient is not spreading to the edge of the stroke (the weird clipping in the converted PDF is an issue that should be fixed in a newer hayro version)
[report.html](https://github.com/user-attachments/files/25694315/report.html)~~

~~So essentially the special handling for rectangles also has to check if the paint is converted for the stroke or the fill of the rectangle shape.~~

EDIT: It's probably desirable to have the same transform as the rect fill here, so let's just leave this as is :thinking:

crates/typst-render/src/shape.rs
```diff
@@ -67,43 +67,28 @@ pub fn render_shape(canvas: &mut sk::Pixmap, state: State, shape: &Shape) -> Opt
...
             let dash = dash.as_ref().and_then(to_sk_dash_pattern);
-
-            let bbox = shape.geometry.bbox_size();
-            let offset_bbox = if !matches!(shape.geometry, Geometry::Line(..)) {
-                offset_bounding_box(bbox, *thickness)
-            } else {
-                bbox
-            };
-
+            let bbox = shape.geometry.bbox(shape.stroke.as_ref());
```
saecki: (This comment should go on line 44, but I can't add comments to lines not modified in this PR)

I've found another issue only present in `typst-render` for rectangles that don't have a radius. This is the `issue-5705-gradient-border-stroke` test without the `radius`:

```
#rect(
  width: 100pt,
  height: 100pt,
  stroke: (
    top: gradient.linear(green, black, yellow, angle: 90deg) + 20pt,
    right: gradient.linear(green, black, yellow, angle: 180deg) + 20pt,
    bottom: gradient.linear(green, black, yellow, angle: 270deg) + 20pt,
    left: gradient.linear(green, black, yellow, angle: 0deg) + 20pt,
  )
)
```

[report.html](https://github.com/user-attachments/files/25694919/report.html)

crates/typst-pdf/src/paint.rs
eltos: Okay, reverted to use `Point` and `Size` again as in the initial commit.

crates/typst-render/src/shape.rs
eltos: Fixed and test updated

grafik

eltos:
@saecki ready for 3rd review

> Thanks! And sorry for the delay (we had quite a busy week).

No worries. I look forward to watching the talks from the meeting on your YouTube channel.

crates/typst-pdf/src/paint.rs
eltos: I thought about it again, and I think for consistency it would actually be better to use different bboxes, i.e. use the tight bbox-without-stroke for the fill gradient, and the extended bbox-with-stroke for the stroke gradient.
Here's an example with a `rect`, `rect(radius: 20pt)`, `curve` and `circle`:

| Before PR | Current state of PR | Proposed |
|-|-|-|
|before|fc70e77|proposed

Reasoning:
- Before PR: uses bbox-without-stroke for both fill and stroke, this breaks gradients on lines, curves etc. where the bbox size collapses
- Current state of PR: Uses bbox-with-stroke for strokes on lines and curves but also changes bbox for some fills. This leads to inconsistency because of the special handling for rects. For round rects the situation is more complex as fill and strokes are handled separately and the codepath that generates the fill is not aware of the strokes (I think the reason being that strokes can be different for the four edges). So using the bbox-with-strokes for the fill is currently not possible.
- Proposed: Keep using the bbox-without-stroke for fills as before, and use bbox-with-stroke for all strokes (even for `rect` to handle zero-sized rects with strokes). This results in different bboxes for fill and stroke, but is consistent across shapes, predictable and backwards compatible with respect to gradient fills.

Note that the test "gradient-fill-and-stroke", which with the current state of the PR still passes, would change with the proposed version.
canvas
As it's a bit difficult to see, here a variant with a sharp gradient and different colors:




It's a design decision (at least until linear gradient start/end point can be controlled, see https://github.com/typst/typst/issues/2282). 
Let me know what you think.

saecki: > Note that the test "gradient-fill-and-stroke", which with the current state of the PR still passes, would change with the proposed version.

That's what I thought of when I wanted to dismiss this comment, but I didn't think of the other cases where this already happens because the rectangle is decomposed into different shapes.

> I thought about it again, and I think for consistency it would actually be better to use different bboxes, i.e. use the tight bbox-without-stroke for the fill gradient, and the extended bbox-with-stroke for the stroke gradient.

I agree, we definitely either want the state `Before PR` or the `Proposed` one, not the in-between we currently have. Since the `Before PR` state wouldn't allow properly and consistently dealing with gradients on strokes, I think it's better to opt for your proposed changes.

Maybe the use case of `gradient-fill-and-stroke` can be revisited together with `relative` gradients and tilings at some point. But at least for the example shown, just removing the stroke and using an `outset` should be enough to replicate it.

eltos: Okay, I applied the proposed changes and replaced the `gradient-fill-and-stroke` test




Also did some more refactoring, including what you suggested earlier:
> Then we could get rid of the bbox_size, and [...]
> Finally we could add a Shape::bbox() method, which uses the shape's stroke field, since that seems to always be passed anyway.
> This refactor can be done in a follow-up PR, it just feels like this would remove a lot of the complications.

crates/typst-library/src/visualize/curve.rs
```diff
@@ -475,46 +475,22 @@ impl Curve {
...
+        Rect::new(
+            Point::new(Abs::raw(bbox.x0), Abs::raw(bbox.y0)),
+            Point::new(Abs::raw(bbox.x1), Abs::raw(bbox.y1)),
+        )
     }
 
     /// Computes the size of the bounding box of this curve.
     pub fn bbox_size(&self) -> Size {
-        self.bbox().size()
+        self.bbox(None).size()
```
saecki: I think we can now completely remove the `bbox_size` method

eltos: Done with ce72e88c39065a5b7def4d81645ad9f1f23c7116

laurmaedje:
Thank you!

PR #7864, opened 2026-02-13
Fix typo in equation docs

WakaJoekoe:


crates/typst-library/src/math/equation.rs
```diff
@@ -29,11 +29,11 @@ use crate::text::{FontFamily, FontList, FontWeight, LocalName, Locale, TextElem}
 /// #set text(font: "New Computer Modern")
 ///
 /// Let $a$, $b$, and $c$ be the side
-/// lengths of right-angled triangle.
+/// lengths of a right-angled triangle.
 /// Then, we know that:
 /// $ a^2 + b^2 = c^2 $
 ///
-/// Prove by induction:
+/// Proof by induction:
```
laurmaedje: This was intended as an imperative. I think it was correct.

WakaJoekoe: after rereading, i agree that it should be an imperative

Issue #7524, opened 2025-12-04
math vector function 'vec' not working properly with arguments that are variables


jevgarrido:
### Description

The following works:
```typst
#let val = 10;

$ mat(val) $
$ mat(#val) $
$ mat(#[#val]) $
$ mat(#{val}) $
$ mat(#{}val) $
```

However: 
```typst
#let val = 10;

$ vec(val) $ // <- Not working
$ vec(#val) $ // <- Not working
$ vec(#[#val]) $
$ vec(#{val}) $ // <- Not working
$ vec(#{}val) $
```

The error is "Expected content, found integer." for the three cases.

This bug is persistent in all versions of the compiler available on the web app at the moment (from 0.10.0 to 0.14.1)

Thank you!

### Operating system

Web app

### Typst version

- [x] I am using the latest version of Typst

Enivex:
It might be inconvenient, but it's not really a bug. You need to make your variable `$10$` instead.

cAttte:
isn't it a little weird that `mat` accepts numbers but `vec` doesn't 🤔

Enivex:
> isn't it a little weird that `mat` accepts numbers but `vec` doesn't 🤔 

It's too early in the morning, ok 😅

I must've just skimmed through that part.

I'm honestly more surprised that it actually works for mat than that it doesn't for vec

Issue #5608, opened 2024-12-19
Using a push module function leads to error


Gijom:
### Description

Hello there,

I noticed that when using a push function from a module into another function I get the error "variables from outside the function are read-only and cannot be modified".

Here is a minimal exemple.

First a module "module.typ":
```typst
#let push() = [
  This is a dummy function
]
```

Then try to use it in another module and in a function, for instance a "call.typ" module:
```typst
#import "module.typ" as m

#m.push()  //no problem there

#let mycall() = {
  m.push()  //here is the error
}

#mycall()
```

The exact error is:
```
error: variables from outside the function are read-only and cannot be modified
  ┌─ call.typ:6:2
  │
6 │   m.push()
  │   ^

help: error occurred in this call of function `mycall`
  ┌─ call.typ:9:1
  │
9 │ #mycall()
  │  ^^^^^^^^
```

My feeling is that for some reason it interprets m as being a global array which thus cannot be modified from the function.

Changing the function name to any other name (e.g. add) works like a charm. I was wondering if this is supposed to be normal or a bug.

### Operating system

Linux

### Typst version

- [X] I am using the latest version of Typst

laurmaedje:
You've discovered a cursed part of the Typst compiler. It needs to know upfront whether to evaluate the callee `abc` of `abc.def()` as mutable or not, and for this reason the names `"push" | "pop" | "insert" | "remove"` are currently special-cased when used in method call notation (so that array & dictionary methods work). I didn't find a better solution at the time.

PR #7865, opened 2026-02-13
Correct math unparsing

### LINKED ISSUE #7274: Problems with non-functions sharing the function call syntax in math mode
I'm adding this because it isn't tracked elsewhere, although most of the cases are present in the test suite.

In math mode the function call syntax needs to 'unparse' a call at runtime if the function identifier isn't actually a Typst function. E.g. `$pi(x,y)$` will render like `$pi\(x\,y\)$`.

However, unparsing is currently broken/incomplete when using any special syntax for arguments, i.e. named arguments, argument spreading, and 2D arguments. We currently give poor/nonsense errors for these cases because we try to evaluate the arguments before realizing we need to unparse. E.g. `$pi(x:y)$` gives "Error: unexpected argument: x" (this is distinct from #3802 because `pi` isn't a function).

Instead, we should either correctly unparse the special syntax, or give better error messages if we do recognize it being used with a non-function.

Also, when parsing a non-function identifier we have different behavior for delimiters: unclosed parentheses cause an error ( `$pi(x$` is an error, but not `$pi (x$` ) and parentheses don't pair with non-parentheses (`$pi(2,5]$`).

isuffix:
This is now a normal PR targeting `main`
~~**Part of the Math Syntax Refactor**
This PR is part of the exhaustive math syntax refactor. It should be considered exploratory, and not necessarily a final decision, even if merged to the `math-syntax` branch. Non-specific feedback for this change should go in one of its related issues.~~

---

As a quick refresher, recall these existing features of math arguments:
- Empty arguments create empty content: `$func(, , , )$`
- Semicolons create two-dimensional arguments: `$mat(1, , ; ,2, ; , ,3;)$`
- Spread arguments require the expression to directly follows the two dots: `$func(..spread, .. not-spread)$`
- Named arguments require the colon to directly follow the name: `$func(named-param: value, not-named : X)$`

---

This PR implements correct math unparsing by rewriting the entire math function call infrastructure and refactoring how we evaluate method function calls. It adds two new SyntaxKinds to distinguish math calling and math arguments, and now handles two-dimensional arguments in evaluation rather than parsing.

With my prior PR, #7742, this will fully resolve #7274.
#7524 is somewhat related, but is not resolved by this PR, although `$sin(#1)$` does now work.

The first commit adds and updates many test cases for math function calls that will be improved in the final commit (although the test ordering in that file is still quite messy). The second commit moves some code in `call.rs` around to improve locality, which I prefer instead of putting all `impl Trait` blocks before all free functions. Finally, the third commit implements everything and updates the tests. I apologize that it is quite large, but everything in it depends on everything else, so one-commit it shall stay.

### What is unparsing?
In Typst's math mode, we allow `$vec(1, 2)$` to call the `math.vec` function, but we also allow `$pi(1, 2)$` to render as the symbol `pi` next to `(1, 2)` as content. These two examples are syntactically indistinguishable, so we parse them into the same syntax tree and interpret that tree at runtime as either function arguments or content based on the type of the callee. To avoid parsing while evaluating, we currently fully parse the argument list as though we would call a function, and if we end up needing to render arguments as content, we "unparse" the argument structure and build content by manually inserting commas between arguments and wrapping in parentheses.

However, the old approach is wildly incomplete code-wise and suffers from architectural choices that make unparsing, named/spread argument syntax, and two-dimensional arguments fragile.

In the old approach:
- Unparsing errors with any two-dimensional arguments: `$pi(1; 2)$`
- Unparsing errors with any named arguments: `$pi(x: y)$`
- Unparsing includes spread arguments after spreading them, so `$pi(1, ..#none, 3)$` only renders one comma between `1` and `3`
- Unparsing does not respect spaces around arguments: `#repr($pi(, | ,)$)` vs. `#repr($pi (, | ,)$)`
- Punctuation added in unparsing (commas and parens) is not given span information
- Two-dimensional arguments error if named arguments are not directly at the beginning or directly behind a semicolon: `$mat(1, delim: "[", 2;)$` and `$mat(1, 2; 3, 4, delim: "[")$`
- Spread arguments with named components never work with two-dimensional arguments: `$mat(..#(delim: "["), 1, 2; 3, 4)$`
- EDIT: I forgot to mention that unparsed integer arguments like in `$sin(#1)$` would always fail because we try to cast them to content instead of displaying them like we do for `$#1$`

This PR makes almost all of these work correctly and produce the output you would expect.

The two exceptions are that we error (with useful hints) if we would unparse a call that uses either spread argument sytnax or named argument syntax (`$ pi(..items) pi(param: y) $`). This is because these features resemble code mode too closely and it would be confusing if this sytnax was allowed to be used with non-functions. However, this still follows the rules from above, where adding a single space before the colon or after the two dots will allow unparsing.

One thing that is not changing in this PR is that semicolons that directly follow embedded code expressions in arguments still attach to the code instead of creating two-dimensional arguments, e.g. `$mat(#"abc"; #"def"; ;)$` only has one row. This may change, but not in this PR.


isuffix:
One interesting note is that the new `math-call-symbol-{named,spread}-argument` tests are the first instances where a hint uses a different span than the error (besides `convergence-position` and `convergence-page-supplement` where the main error is `Detached`).

```typ
--- math-call-symbol-named-argument eval ---
// Error: 10-18 named-argument syntax can only be used with functions
// Hint: 3-6 `phi` is not a function
// Hint: 10-18 to render the colon as text, escape it: `alpha\: y`
$ phi(x, alpha: y) $
```

I tested, and this renders great on the command line, but it may be worth opening an issue in the WebApp repo if it doesn't render well there.

```txt
error: named-argument syntax can only be used with functions
  ┌─ <stdin>:1:9
  │
1 │ $ phi(x, alpha: y) $
  │   ---    ^^^^^^^^
  │   │       
  │   `phi` is not a function
  │
  = hint: to render the colon as text, escape it: `alpha\: y`
```

laurmaedje:
> I tested, and this renders great on the command line, but it may be worth opening an issue in the WebApp repo if it doesn't render well there.

Aware of that! It's not yet well-handled in the dev version on the web app, but we want to deal with that before the next release.

> besides convergence-position and convergence-page-supplement where the main error is Detached

The hint in `convergence-position` belongs to the "heading position did not stabilize" warning, which is not detached, so I think it's basically the same as yours (which looks lovely, btw).

crates/typst-syntax/src/ast.rs
```diff
@@ -990,6 +994,144 @@ impl MathShorthand<'_> {
...
 
+node! {
+    /// A function call in math: `mat(delim: "[", a, b; ..#($c$,), d)`.
+    struct MathCall
+}
+
+impl<'a> MathCall<'a> {
+    /// The function to call.
+    pub fn callee(self) -> MathCallee<'a> {
+        self.0.try_cast_first().unwrap()
```
laurmaedje: Why unwrap here instead of the usual `cast_first`?

crates/typst-syntax/src/parser.rs
laurmaedje, line 1: these changes make me happy. especially good riddance on the array wrapping, since there isn't at all an array here in the syntax...

tests/suite/foundations/dict.typ
```diff
@@ -193,8 +193,8 @@
     nonfunc: 1
   )
 
-  // Error: 8-15 type dictionary has no method `nonfunc`
-  // Hint: 8-15 did you mean to access the field `nonfunc`?
+  // Error: 3-15 expected function, found integer
```
laurmaedje: this error seems to imply that calling a function in a dictionary without parentheses _is_ possible. the previous error seems clearer to me.

tests/suite/math/call.typ
```diff
@@ -235,9 +264,14 @@ $ sin( ,/**/x/**/, , /**/y, ,/**/, ) $
...
 
+--- math-call-value-error eval ---
+// TODO: Why does this work for mat, but not vec?
+$ mat(#1) $
+// Error: 8-9 expected content, found integer
+$ vec(#1) $
+
 --- math-call-value-non-func paged ---
 $ sin(1) $
-// Error: 8-9 expected content, found integer
```
laurmaedje, line 240: this change is not listed in the PR description (unless I missed it), may be worth adding there and discussing (since there is a _bit_ of potential for confusion why `sin(#1)` works but `vec(#1)` does not). See also https://github.com/typst/typst/pull/7865#discussion_r2828042798

tests/suite/math/call.typ
```diff
@@ -333,50 +335,49 @@ $ pi.alt() $
 --- math-call-field-symbol-error eval ---
 #math.tilde($pi.alt$)
 // Error: 2-13 symbol ϖ is not callable
+// Hint: 2-13 try adding a space before the parentheses: `math.pi.alt ($tilde$)`
```
laurmaedje: The lack of the hash in the hint is a tad confusing. Overall, I'm not sure the hint is that useful here because to me, it's unclear what the user intent even is.

tests/suite/math/call.typ
```diff
@@ -333,50 +335,49 @@ $ pi.alt() $
...
 // Test accessing a field on a module in a call.
 $ std.color.map() $
 
---- math-call-field-dict-non-func eval ---
+--- math-call-field-dict-non-func paged ---
 // Test accessing a field on a dictionary in a call.
 #let pi = (alt: math.pi.alt)
-// Error: 6-9 type dictionary has no method `alt`
-// Hint: 6-9 did you mean to access the field `alt`?
 $ pi.alt() $
```
laurmaedje: Supporting math field calls on dictionaries is sort of problematic because it introduces the same ambiguity as in code mode. Consider this example:
```typ
#let dict = (text: sym.pi, at: "inaccessible")
$ dict.text("text") $ // → π(text)
$ dict.at("text") $   // → π
```

Symbols can get away with it because there are no methods on symbols. (Not sure whether I actively thought about that or whether that's a lucky coincidence.) But for modules, it's definitely intentionally that they support it with the trade-off that they cannot have any methods.

tests/suite/scripting/methods.typ
```diff
@@ -26,24 +26,24 @@
...
 #let numbers = ()
+// Error: 2-13 type array has no method `fun`
 #numbers.fun()
 
 --- method-unknown-but-field-exists paged ---
-// Error: 2:4-2:10 element line has no method `stroke`
-// Hint: 2:4-2:10 did you mean to access the field `stroke`?
 #let l = line(stroke: red)
+// Error: 2-10 expected function, found stroke
+// Hint: 2-10 try removing the parentheses: `l.stroke`
```
laurmaedje: similarly to https://github.com/typst/typst/pull/7865#discussion_r2827226326, I find this new error a bit confusing

crates/typst-syntax/src/ast.rs
```diff
@@ -990,6 +994,144 @@ impl MathShorthand<'_> {
...
+    Space(Space<'a>),
+    /// A comma separating arguments.
+    Comma(char, &'a SyntaxNode),
+    /// A semicolon separating arguments. This excludes semicolons that end
+    /// embedded code expressions.
+    Semicolon(char, &'a SyntaxNode),
+    /// The left paren at the start of the arguments.
+    LeftParen(char, &'a SyntaxNode),
+    /// The right paren at the end of the arguments.
+    RightParen(char, &'a SyntaxNode),
```
laurmaedje, line 1089: why do these store the chars? The chars seem to be fully defined by just the variant.

laurmaedje: also, the Comma, LeftParen, and RightParen variants seem to be handled equivalently in eval, so I wonder whether this enum is really the ideal public interface for `content_items`.

anyway, let's discuss this setup more later.

crates/typst-eval/src/call.rs
```diff
@@ -33,53 +33,270 @@ impl Eval for ast::FuncCall<'_> {
...
+    // Stacker is broken on WASM.
+    #[cfg(target_arch = "wasm32")]
+    return f();
+
+    #[cfg(not(target_arch = "wasm32"))]
+    stacker::maybe_grow(32 * 1024, 2 * 1024 * 1024, f)
+}
+
+/// Attempt to resolve a mutating method call by evaluating args and then
+/// attmempting to access the target mutably. If the target's type doesn't
```
laurmaedje: attmempting ->  attempting

crates/typst-eval/src/call.rs
```diff
@@ -33,171 +33,271 @@ impl Eval for ast::FuncCall<'_> {
...
-            field.as_str(),
-        ),
-        _ => error!(
-            field.span(),
-            "type {} has no method `{}`",
-            target.ty(),
-            field.as_str(),
-        ),
+        match target.field(&field, sink) {
+            Ok(callee) => callee,
```
laurmaedje: This is a bit dangerous as it means fields can now be called on potentially any type instead of the blessed list of the four types listed below. For instance, this is now possible
```typ
#let h = heading(supplement: x => x + 1)[]
#h.supplement(2)
```
where it previously generated an error. This results in ambiguity between calling a field and a method.

From the blessed list, the only type with methods is `function`, which is already not so great...

Was this intentional? 

Relates to https://github.com/typst/typst/pull/7865#discussion_r2827314215

crates/typst-eval/src/call.rs
```diff
@@ -33,171 +33,271 @@ impl Eval for ast::FuncCall<'_> {
...
+                            "try removing the parentheses: `{}`",
+                            access.to_untyped().clone().into_text(),
+                        ));
+                    } else {
+                        err.hint(eco_format!(
+                            "try adding a space before the parentheses: `{} {}`",
+                            access.to_untyped().clone().into_text(),
+                            args.to_untyped().clone().into_text(),
+                        ));
                     }
```
laurmaedje: as mentioned on the tests, these hints aren't super clear to me because the user intent is not so clear in the first place. let's discuss these more.

crates/typst-eval/src/call.rs
```diff
@@ -33,171 +33,271 @@ impl Eval for ast::FuncCall<'_> {
...
+                    Err(err) => FieldCallee::NonFunc(callee_value, err),
+                }
+            }
+            ast::MathCallee::FieldAccess(access) => {
+                let target_expr = access.target();
+                target_span = target_expr.span();
+                let field = access.field();
+                let target = target_expr.eval(vm)?;
+                if is_mutating_method(field.as_str())
+                    && matches!(target, Value::Array(_) | Value::Dict(_))
```
laurmaedje: The RHS of this `&&` introduced a change in behavior that isn't backed by a test.

This now works were it previously errored:
```typ
#let val = symbol(("push", "a"))
$ val.push(1) $
```

Was this intentional?

crates/typst-eval/src/call.rs
```diff
@@ -252,40 +352,162 @@ impl Eval for ast::Args<'_> {
...
-        body += arg;
-    }
-    if trailing_comma {
-        body += SymbolElem::packed(',');
     }
-
-    let formatted = callee.display().spanned(callee_span)
-        + LrElem::new(SymbolElem::packed('(') + body + SymbolElem::packed(')'))
+    if errors.is_empty() {
+        let parens = LrElem::new(SequenceElem::new(body).pack())
```
laurmaedje: The let binding seems unnecessary here. Overall, I'd probably write this as
```rust
if !errors.is_empty() {
    return Err(errors);
}

Ok(LrElem::new(SequenceElem::new(body).pack())
    .pack()
    .spanned(args.span()))
```

crates/typst-eval/src/call.rs
```diff
@@ -252,40 +352,162 @@ impl Eval for ast::Args<'_> {
...
+            }
+        }
+
+        if let Some(start) = two_dim_start
+            && start != pos.len()
+        {
+            drain_into_array(&mut pos, start, self.span());
+        }
+
+        named.extend(pos);
```
laurmaedje, line 471: since this changes the order, I wonder whether the order of named vs pos arguments is currently observable other than via `repr`. maybe actually not. but either way, we'd probably not want to give any guarantees on that anyway.

crates/typst-syntax/src/ast.rs
```diff
@@ -990,6 +994,144 @@ impl MathShorthand<'_> {
...
+/// a semicolon.
+#[derive(Debug, Copy, Clone, Hash)]
+pub struct MathArg<'a> {
+    /// The argument.
+    pub arg: Arg<'a>,
+    /// Whether the argument ends with a semicolon and should create
+    /// two-dimensional args. This excludes semicolons that end embedded code
+    /// expressions.
+    pub ends_in_semicolon: bool,
+}
```
laurmaedje, line 1068: I suppose a design with an enum of either an `Arg` or a `Semicolon` would also work on the call site in typst-eval. Did you consider and discard this alternative or would it be worth exploring? It would be more parallel with `MathArgItem`. Overall, a bit related to https://github.com/typst/typst/pull/7865#discussion_r2827698551.

crates/typst-eval/src/call.rs
```diff
@@ -252,40 +352,162 @@ impl Eval for ast::Args<'_> {
...
+                body.push(space.eval(vm)?.spanned(space.span()));
+            }
+            ast::MathArgItem::Comma(c, node)
+            | ast::MathArgItem::Semicolon(c, node)
+            | ast::MathArgItem::LeftParen(c, node)
+            | ast::MathArgItem::RightParen(c, node) => {
+                body.push(SymbolElem::packed(c).spanned(node.span()));
+            }
+            ast::MathArgItem::Arg(ast::Arg::Pos(expr)) => {
+                body.push(expr.eval(vm)?.display().spanned(expr.span()));
```
laurmaedje, line 500: Here's the root cause of https://github.com/typst/typst/pull/7865#discussion_r2827249833. Previously we used `args.all::<Content>()` and not `Value::display`.

tests/suite/math/call.typ
isuffix: spans generally don't include hashes because we don't wrap them when producing nodes. E.g. we always produce `Markup [ Hash, FuncCall [ ... ] ]` instead of `Markup [ CodeExpr [ Hash, FuncCall [ ... ] ] ]`, this causes me many annoyances in math AST manipulation :)

crates/typst-eval/src/call.rs
isuffix: The change is intentional, but not leaving out the tests. I had a number of tests in a draft branch for mutability that I left out to keep centralized with some future mutable method changes, but I'll bring the relevant ones over into the first commit here.

crates/typst-eval/src/call.rs
isuffix: I was aware of making this change, but did not realize it had a tangible effect. Let's discuss this more synchronously.

crates/typst-syntax/src/ast.rs
isuffix: Mainly developer comfort, it really simplifies the `Eval` impl to be able to match on the structure and bind the characters in one spot. This used to be a single `Punctuation` variant with a char, but I didn't want to match on that char being a semicolon in `arg_items` below so separated the variants.

crates/typst-syntax/src/parser.rs
isuffix: The wonderful part about the parser changes is that I only ever implemented them once like a month ago and haven't had to touch them to make anything else work!

tests/suite/foundations/dict.typ
isuffix: That's a good point. I still don't like the previous error (it doesn't really say that the field exists or _how_ to access it), but now we get to work to improve them both 💪

tests/suite/math/call.typ
isuffix: Added to the description, thanks for the catch. I think `sin(#1)` vs. `vec(#1)` is certainly unfortunate, but it's better for `sin(#1)` to match `$#1$`.

crates/typst-syntax/src/ast.rs
isuffix: Just because it didn't implement `Default`.

tests/suite/math/call.typ
isuffix: I'm removing this hint in general when we're outside of math.

isuffix:
Updates!
- The rebases are a bit complex, but [this difference link](https://github.com/typst/typst/compare/a28721bd330efa1e62dd71aaa3ff7f91ee806d9b..632e146f68c076ead2c2f11bd4d6895606b64afc) is the one to use if you want to see the overall diff from the previous version.
- This is now dependent on #7897 and includes it as the first commit.
- I've finally bit the bullet and fully reorganized `math/call.typ`. It now separates sections more nicely, doesn't include `-repr` at the end of test names, and moves some separate tests out to `interactions.typ` and `accent.typ`.

    The only really notable difference other than the reorganization and naming is the removal of `math-call-named-args`. This test is fully covered by `math-call-named-basic`, and doesn't add anything so I'm dropping it.
- A new commit follows the updates to `math/call.typ` with updates in `scripting/methods.typ`. I chose to keep this file and expand it massively. It now contains every test I've written for mutable methods everywhere, and duplicates most of them as `math-method-*` tests.
- I split out and reworked the general changes to missing method errors into a new commit. You can now see how this impacts `methods.typ` separately from the main commit. I believe this address most of the review comments, and I'm ready for more :muscle:
- I added another commit for a small change to errors for `push/pop` on dictionaries that I thought was nice.
- In the main commit, I moved the implementation of `Eval` for `MathCall` into a free function, `eval_math_call`. This avoids future rebase conflicts by preempting the same change in my planned adjustments to #7866.

crates/typst-syntax/src/ast.rs
isuffix: Yes, but I think it would be worse because it doesn't encode the property that a a semicolon _must_ follow an argument (albeit, maybe a named argument). Then the caller would have to either read the documentation, handle the case of two semicolons in a row, or be more mindful about structuring the algorithm so that case won't have an impact.

laurmaedje:
Great! I'll take a look at the updates soon.

I merged https://github.com/typst/typst/pull/7897, so you might want to rebase math-syntax and this off of main and I can then reset `math-syntax` to the relevant hash.

crates/typst-eval/src/call.rs
```diff
@@ -27,59 +27,313 @@ impl Eval for ast::FuncCall<'_> {
...
+    } else {
+        // Otherwise we cannot call this field and produce an error.
+        let full_text = || access.to_untyped().clone().into_text();
+        match target.field(&field, sink) {
+            // The field does exist.
+            Ok(callee_value) => match target {
+                Value::Dict(_) => {
+                    let mut err = error!(
+                        access.span(),
+                        "cannot directly call dictionary keys as methods";
```
laurmaedje: wouldn't "as functions" be more accurate?

laurmaedje: looking at the comment later which states "This lets us create a callable method with an arbitrary name without the inconvenience of creating a whole module", I think we have different definitions of the term "method".

I think for you, it's any kind of call where the callee is a field access, while for me, it's specifically calling a function in a way where the thing before the dot will be passed as the first argument.

tests/suite/scripting/methods.typ
```diff
@@ -25,27 +22,302 @@
...
+
+--- math-method-unknown-content eval ---
+#let pi = [pi]
+// Error: 3-9 element text has no method `fun`
+$ pi.fun() $
+
+--- method-invalid-content-field-func eval ---
+// This would be cursed to allow.
+// Error: 2-48 `supplement` is not a valid method for element `heading`
+#heading([], supplement: x => x + 1).supplement(2)
```
laurmaedje, line 70: previously, this generated a hint "Did you mean to access the field `supplement`?". even if you're not a fan of that, should it maybe generate the same hint as dictionaries do?

tests/suite/scripting/methods.typ
```diff
@@ -25,27 +22,302 @@
...
+
+--- math-method-invalid-content-field-func eval ---
+#let cancel = math.cancel($x$, angle: ang => ang + 90deg)
+// Error: 3-15 `angle` is not a valid method for element `cancel`
+$ cancel.angle(#30deg) $
+
+--- method-invalid-non-func eval ---
+// Error: 2-26 `stroke` is not a valid method for element `line`
+// Hint: 2-26 this looks like a method call, but `line(stroke: red).stroke` produced type `stroke`
```
laurmaedje: tbh, this is more confusing to me than the existing
```
Error: element line has no method `stroke`
Hint: Did you mean to access the field `stroke`?
```
let's discuss this syncly

tests/suite/scripting/ops.typ
```diff
@@ -512,3 +517,11 @@
 // (since then it doesn't resolve to the standard library version anymore).
 #let rect = ""
 #(rect = "hi")
+
+--- ops-assign-shadow-eval-order eval ---
+/// Test shadowing a variable while assigning to it and calling a method on it.
```
laurmaedje: There's an unnecessary extra `/` here

tests/suite/scripting/methods.typ
```diff
@@ -1,17 +1,14 @@
-// Test method calls.
+// Test method calls in code and in math.
+
+// A pattern used in some of the `method-accent-*` tests is to create a symbol
+// with an accent character (`sym.tilde`) as a field. This lets us create a
+// callable method with an arbitrary name without the inconvenience of creating
+// a whole module.
 
 --- method-whitespace eval ---
```
laurmaedje: As discussed, let's replace "method" with "field-call"

laurmaedje (commented):
As discussed, I still had some pending comments, but we basically discussed everything syncly already, so this is just for completeness. 

On the core logic in call.rs, the changes look fine to me. I'd take one last look once we have finalized the error messages and hints, but otherwise it should be fine.

Let me know if further input from my side is needed!

crates/typst-eval/src/call.rs
isuffix: Updated to "as functions" and fixed that comment. LMK if I'm still using the method/field call language imprecisely anywhere else :+1:

tests/suite/scripting/methods.typ
isuffix: I'm now generating similar hints as for dictionaries:
```typ
--- field-call-invalid-content-field-func eval ---
// This would be cursed to allow.
// Error: 2-48 `supplement` is not a valid method for element `heading`
// Hint: 2-48 only built-in methods on `heading` can be directly called as functions
// Hint: 2-48 to call the function indirectly, wrap the field access in parentheses: `(heading([], supplement: x => x + 1).supplement)(..)`
#heading([], supplement: x => x + 1).supplement(2)
```

tests/suite/scripting/methods.typ
isuffix: Thoughts on this update?
```typ
--- field-call-invalid-non-func eval ---
// Error: 2-26 `stroke` is not a valid method for element `line`
// Hint: 2-26 only built-in methods on `line` can be directly called as functions
// Hint: 2-26 to access the `stroke` field, remove the function arguments: `line(stroke: red).stroke`
#line(stroke: red).stroke()
```

isuffix:
I've updated the error messages for field calls and resolved some of the other comments. LMK what else needs adjusting.

crates/typst-eval/src/call.rs
```diff
@@ -27,59 +27,307 @@ impl Eval for ast::FuncCall<'_> {
...
+            }
+        }
+        ast::MathCallee::FieldAccess(access) => {
+            let target_expr = access.target();
+            target_span = target_expr.span();
+            let field = access.field();
+            let target = target_expr.eval(vm)?;
+            if is_mutating_method(field.as_str())
+                && matches!(target, Value::Array(_) | Value::Dict(_))
+            {
```
laurmaedje, line 108: The fact that we eval first instead of accessing and then check for array/dict means than in some respects code is now _more_ limited than math:
```typ
// other.typ
#let push(it) = it

// main.typ
#import "other.typ"
#let meta = metadata(other)

// cannot mutate fields on content
#meta.value.push(2)

// just works
$ meta.value.push(2) $
```

This is not a huge problem per se, since in code it's really a bug than this doesn't work, but worth calling out. And it may make mutable methods harder to support in math down the road (though we really want to have our cake and eat it --- i.e. support them while fixing the bug).

crates/typst-eval/src/call.rs
```diff
@@ -27,59 +27,307 @@ impl Eval for ast::FuncCall<'_> {
...
+                // handled above.
+                let is_dict = matches!(target, Value::Dict(_));
+                let mut err = if is_dict {
+                    // Dictionaries get a specific error & hint because they're
+                    // the easiest to attempt this with, and users need to be
+                    // told directly why it's not allowed.
+                    error!(
+                        access.span(),
+                        "cannot directly call dictionary keys as functions";
+                        hint: "cannot use method syntax with dictionary keys, \
```
laurmaedje: I think the comma after "keys" isn't correct here, though "keys as keys" reads a bit weirdly. perhaps "... as they could ..." would be an alternative?

crates/typst-eval/src/call.rs
```diff
@@ -27,59 +27,307 @@ impl Eval for ast::FuncCall<'_> {
...
+                    )
+                };
+                if callee_value.clone().cast::<Func>().is_ok() {
+                    err.hint(eco_format!(
+                        "to call the function indirectly, {}wrap the field access \
+                            in parentheses: `{}({})(..)`",
+                        if in_math { "use code mode and " } else { "" },
+                        if in_math { "#" } else { "" },
+                        full_text()
+                    ));
```
laurmaedje, line 288: I think we're slowly converging:

- I like the main diagnostic message.
- The second hint I also like overall, but I think "indirectly" is a bit confusing in what exactly is meant and it's not defined anywhere what an "indirect call" is supposed to be. I would suggest either just dropping "indirectly" or replacing it with "stored in the `...` field" / "stored at the key"
- The first hint, I'm not a fan of. It feels a bit imprecise with the "method" and "function" terminology, but more importantly I don't think it's doing a lot for the user, especially in the non-function case (e.g. stroke). But even in the function case (e.g. supplement), the other hint feels much more actionable. I think this hint can be just be fully dropped to be honest.
- I think the extra explanation for dictionaries is nice, but less important than the call-to-action of calling the function, so I would move it down and have it as the last hint.

So in total, I would suggest:

```typ
// Error: 2-34 cannot directly call dictionary keys as functions
// Hint: 2-34 to call the function stored at the key, wrap the field access in parentheses: `((call-me: () => "maybe").call-me)(..)`
// Hint: 2-34 method syntax cannot be used with dictionary keys as they could conflict with built-in method names
#(call-me: () => "maybe").call-me()

// Error: 2-48 `supplement` is not a valid method for element `heading`
// Hint: 2-48 to call the function stored in the `supplement` field, wrap the field access in parentheses: `(heading([], supplement: x => x + 1).supplement)(..)`
#heading([], supplement: x => x + 1).supplement(2)

// Error: 2-26 `stroke` is not a valid method for element `line`
// Hint: 2-26 to access the `stroke` field, remove the function arguments: `line(stroke: red).stroke`
#line(stroke: red).stroke()
```

crates/typst-eval/src/call.rs
isuffix: Pinging #5608

I think it's fine to keep this since we do want to have this behavior eventually (including shipping to main). However if you think we may potentially want a different behavior, then we can definitely error here regardless of the target type until we know for sure we want to allow this.

isuffix: This is tested at `tests/suite/src/scripting/methods.typ:202`:
```typ
--- field-call-mut-access-module-push eval ---
// Edge case for module access that isn't fixed.
#import "module.typ"
// Works because the method name isn't categorized as mutating.
#test((module,).at(0).item(1, 2), 3)
// Doesn't work because of mutating name.
// Error: 7-16 cannot mutate a temporary value
#test((module,).at(0).push(2), 3)

--- math-field-call-mut-access-module-push eval ---
// This used to error, but math now correctly ignores the mutable method name.
#import "module.typ"
#let indirect = (mod: module)
#test($indirect.mod.push(#2)$, $#3$)
```

And line 286:
```typ
--- field-call-accent-eval-order-shadowed-push eval ---
// This differs because `push` is a mutating method name, even though the method
// on `sp` here isn't actually mutating.
#{
  let sp = symbol("p", ("push", sym.tilde))
  // Error: 8-15 type boolean has no method `push`
  test(sp.push(let sp = false), sym.tilde(none))
  test(sp, false)
}

--- math-field-call-accent-eval-order-shadowed-push eval ---
// Math doesn't support mutable methods and always evaluates arguments second.
#{
  let sp = symbol("p", ("push", sym.tilde))
  test($sp.push(#let sp = false;)$, $#sym.tilde(none)$)
  test(sp, false)
}
```

isuffix:
> What do you think? Is there anything really breaking in here that means we need to hold back?

I don't think there's anything holding it back. I'll change this to merge onto main. There are some small conflicts with tests when rebasing, but I'll include the fixes when force-pushing.

crates/typst-eval/src/call.rs
isuffix: I'd like to avoid "they" because even if it's not actually ambiguous, it still feels like it requires thought to read correctly. How about reordering the message to this?
"dictionary keys cannot be used with method syntax as keys could conflict with built-in method names"

laurmaedje: That works

crates/typst-eval/src/call.rs
isuffix: For "stored in the `...` field" / "stored at the key", what if we just avoid the extra distinction by saying "to call the stored function, ..."?

Otherwise, I'm good with all of these changes. Thanks for working these out so clearly.

laurmaedje: That would work I think and it'd be a bit more compact too. I'd like to have "field" in there at least once, but we do have "field access", so I think it's fine.

isuffix: I have already pushed that, so we should now be set.

laurmaedje:
Thank you!

PR #7866, opened 2026-02-13
Error on function literals in math

isuffix:

---

This PR causes function literals placed in math to produce errors instead of rendering the function name as monospace text. Function literals in markup mode still render as before.

This depends on PR #7865. Only the last commit is new.

This is in part a preparation for resolving The Math Mode Problem (cf. #7084), but that will come later.

Here is one of the test cases showing the new error and hint. I am not fully satisfied with the wording, so feel free to give suggestions :)
```typ
--- math-func-literal-parens eval ---
// Error: 3-6 this does not call the `mat` function
// Hint: 3-6 to call the `mat` function, write `mat()`
$ mat () $
```

This is implemented by breaking out the `Eval` impls of `MathIdent` and `FieldAccess` into free functions so that when called with `Eval::eval()`, they produce an error on function literals in math, but are still able to be evaluated in `MathCall::eval()` without producing an error.

I'm not quite happy with the amount of complication required for this approach to work with field accesses, and an alternative I considered was to encode the difference between `MathIdent` and `Ident` into a more generic `MathIdentWrapper` which could directly contain both `Ident` and `FieldAccess` nodes when used in math. I didn't go with that for now because it requires a lot of code churn, but it does feel a bit conceptually cleaner. I also believe it could simpify #7407, although have not tried it.


MDLC01:
For the hint, I think it should detect whether parentheses are opened after some whitespace and suggest "to call the `mat` function, remove any whitespace before the opening parenthesis". Right now, it's weird because if the user specified arguments inside parentheses but added a whitespace, the hint is just wrong by suggesting to use `()` with no arguments.

If no parentheses are present, I think the current hint could be rephrased to "to call the `mat` function, specify arguments in parentheses: `mat()`". This provides a little more information.

laurmaedje:
> I'm not quite happy with the amount of complication required for this approach to work with field accesses, and an alternative I considered was to encode the difference between MathIdent and Ident into a more generic MathIdentWrapper which could directly contain both Ident and FieldAccess nodes when used in math. I didn't go with that for now because it requires a lot of code churn, but it does feel a bit conceptually cleaner. 

I took a look to understand the current approach and am happy to learn about the alternative. Let's discuss this later.

isuffix:
I have now implemented the alternative design for `MathAccessWrapper`, and this PR is again just one extra commit ahead of #7865. However, this is not the end of the PR. I have some draft changes locally that should help give us higher fidelity errors, but I still need to work on them for a bit before they're ready. However, I've uploaded this commit because it's relatively self-contained and should be reviewable for now.

isuffix:
I've been working on other things for a while, but now it's time to get back to improving math syntax!

For this PR, I realized that there was a nice simplification if we just return a `MathError` enum from the `eval_display` method. This centralizes the check function literal displays for `MathIdent` and `MathFieldAccess` and doesn't require adding any new eval functions or traits! This also handles edge cases if an embedded code expression would render as a function literal (now erroring) or if a function literal would be passed as an argument to another function in math (now _not_ erroring), neither of which I was handling before :sweat_smile:. And of course, I've added a bunch of tests to cover all the new cases I could think up.

This PR also has a new second commit that adjusts the error range when we're hinting for delimiters to use a sub-range of the overall `Math` node that spans from the function literal to the delimiters, which should make the error more actionable. For example, in the `math-func-literal-parens` test, the main error range is now `3-9` for `mat ()` instead of `3-6` for just `mat`.

Unfortunately, I still do not have a fix that can add hints when a function literal is next to parens but is inside an operator, like in `math-func-literal-frac` for `$ 1 / frac (1,2) $`. However I think this PR should be good enough to merge on its own, and I will plan to try building those hints in a different PR, perhaps by sharing some infrastructure with other math syntax updates.

I would love to hear if anyone has feedback on any of the new error messages or test cases!

tests/suite/math/call.typ
```diff
@@ -365,3 +365,149 @@ $ int(
   // Error: 3-8 missing argument: value
   int()
 ) $
+
+--- math-func-literal-basic eval ---
+// Error: 3-6 this does not call the `mat` function
```
T0mstone, line 370: I feel like this doesn't fit the style of an error message. Plus, it's kind of redundant with the hint.
The error message should more closely stick to what exactly went wrong, so perhaps something like "cannot use function without call"?

cady-b: I agree that this feels more like a hint than an error, and with the others it's generally redundant; below, the "expected content, found function" is more useful as an error to me, but perhaps also difficult to understand for newcomers. Not entirely sure what I'd want it to be (the idea of functions having a sort of debug representation isn't the most straightforward), but perhaps something along the lines of "function type does not represent mathematical content"?

Though wording aside, this is a pretty sweet! A welcome change with really nice hints that'll probably prevent quite a few headaches :sweat_smile:

crates/typst-eval/src/math.rs
```diff
@@ -189,12 +216,119 @@ impl Eval for ast::MathRoot<'_> {
...
+}
+
+impl<'a> ExprExt<'a> for ast::Expr<'a> {
+    /// Evaluate the expression as content for math.
+    fn eval_display(self, vm: &mut Vm) -> Result<Content, MathError<'a>> {
+        let value = self.eval(vm)?;
+        // Symbols can cast to functions, but we don't error since they're also
+        // valid as content.
+        if !matches!(value, Value::Symbol(_))
+            && let Ok(func_value) = value.clone().cast::<Func>()
```
laurmaedje, line 230: I think with `Func::castable` we could avoid a clone in the common case here. For math, this is kind of on the hot path, right?

crates/typst-eval/src/math.rs
```diff
@@ -189,12 +216,119 @@ impl Eval for ast::MathRoot<'_> {
...
+                error.hint(eco_format!("evaluated to the `{name}` function"));
+            }
+            if kind == SyntaxKind::MathCall {
+                // `MathCall` is the only kind that can produce a function
+                // literal but cannot be called by adding trailing parentheses
+                // (writing `$func()()$` doesn't work), so we just return
+                // without adding extra hints.
+                return error;
+            }
+            func = node.full_text();
```
laurmaedje, line 298: This seems to happen equivalently in both branches.

tests/suite/math/call.typ
laurmaedje: How about "cannot display function" and add the current message as a hint?

crates/typst-eval/src/math.rs
isuffix: Technically we don't use the full text of the node if we return due to a `MathCall` like `$returnsfunc()()$`, but given that that's very unlikely and we're already producing an error, I'll move this up.

PR #7868, opened 2026-02-14
JSON Schema for Typst manifest

jassielof:
Adds a JSON Schema for the Typst package manifest.

This was previously published by me in the Schema Store (<https://github.com/SchemaStore/schemastore/pull/4662>) but later requested to be taken down (<https://github.com/SchemaStore/schemastore/pull/5125>), I also maintain it in my personal repository (<https://github.com/jassielof/json-schemas>) which also includes the Hayagriva one.

It should be compliant, so far, with the docs in the "packages" repository.

I'm just adding it to a schema directory (both), I also wasn't sure if you'd want it in the "packages" repository (given <https://github.com/typst/packages/pull/559>) so here I'm just adding it.

To serve this directly from this repository one can simply add a reference just as <https://github.com/SchemaStore/schemastore/pull/5086>.


schema/typst.schema.json
```diff
@@ -0,0 +1,282 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://jassielof.github.io/json-schemas/docs/typst.schema.json",
```
elegaanz: If this is to be merged in this repository, the ID should be changed.

schema/typst.schema.json
```diff
@@ -0,0 +1,282 @@
...
+    "title": "Typst Package Manifest",
+    "description": "(Community) Schema for typst.toml package manifest files, used by the Typst package manager and universe.\nhttps://github.com/typst/packages/blob/main/docs/manifest.md",
```
elegaanz: The `(Community)` should be removed too. And I'm not a big fan of putting a raw link here, it feels like this field should only contain raw text.

schema/typst.schema.json
```diff
@@ -0,0 +1,282 @@
...
+    "required": ["package"],
+    "properties": {
+        "package": {
+            "type": "object",
+            "description": "Contains metadata about the package.",
+            "required": ["name", "version", "entrypoint"],
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "The package's identifier in its namespace.\nSee [Naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).",
```
elegaanz: Will Markdown be rendered correctly by all consuming software? I couldn't find any mention of it in the JSON Schema specification. If not, I think it would be preferable to replace it with a raw link or just "See the naming rules in the official packaging documentation".

schema/typst.schema.json
```diff
@@ -0,0 +1,282 @@
...
+                    "pattern": "^[a-z\\d]+(-[a-z\\d]+)*$",
```
elegaanz: Package names can't start with numbers. Also, I have a preference for `0-9` instead of `\d` in that case as it is more explicit and avoids double-escaping. So I think the correct pattern would be:

```
^[a-z][a-z0-9]*(-[a-z0-9]+)*$
```

schema/typst.schema.json
```diff
@@ -0,0 +1,282 @@
...
+                    "license": "MIT",
+                    "description": "Calculate elementary arithmetic with functions."
+                }
+            ],
+            "additionalProperties": false
+        },
+        "template": {
+            "type": "object",
+            "description": "If present, declares this package as a template.\nTemplate packages must specify at least one category.",
+            "required": ["path", "entrypoint", "thumbnail"],
```
elegaanz: The thumbnail is only required for Universe submissions.

schema/typst.yaml
elegaanz, line 1: Even if I agree YAML is more convenient to work with, the duplication is not so great. We don't expect the schema to be changed or even read by a human that often, so removing it and keeping only the JSON version would be preferred.

schema/typst.schema.json
jassielof: Where is it going to be served from? from <typst.app> or directly from raw GitHub? as that's where the ID should point to.

schema/typst.schema.json
jassielof: Links were added as _best practice_ from the [Schema Store](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#useful-conventions), do you still want it out?

schema/typst.schema.json
jassielof: In both VS Code and Zed, using the Tombi LSP/extension, it's rendered correctly as Markdown:

image

image

schema/typst.schema.json
elegaanz: This would need input from @laurmaedje or @reknih but I think either `https://typst.app/typst.schema.json` or `https://typst.org/typst.schema.json` could work. It will probably not actually be deployed, as the spec clearly states that [this URI only serves as a unique identifier](https://json-schema.org/draft/2020-12/json-schema-core#name-the-id-keyword), and doesn't need to be an actual endpoint to fetch the schema.

schema/typst.schema.json
elegaanz: If it's recommended to do so, it's fine, it can stay as is. Thank you for pointing it out.

schema/typst.schema.json
elegaanz: I'm a bit worried that this won't work everywhere, as this seems to be a non-standard feature they decided to add, even if they are probably two of the most common consumers. In the worst case, reading a bit of raw Markdown is not too bad, so unless someone else from the team opposes it, it can stay as is.

schema/typst.schema.json
jassielof: I'll leave it with GitHub then, in the meantime.

elegaanz: After a bit more discussion we agreed on `https://typst.org/schema/json/package-manifest/v1`. Using `typst.org` makes it clear that it is independent from Typst GmbH and only related to the open-source side of Typst. The others components allow for clear disambiguation and hierarchy, in case we were to add more schemas in the future. And the `v1` component will also us to version these schemas if we evolve them in the future.

jassielof: Should I change it then? Or will it be changed in place on deployment? The ID.

elegaanz: Yes please, change it.

schema/typst.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft-07/schema#",
-  "$id": "https://raw.githubusercontent.com/typst/typst/refs/heads/main/schema/typst.schema.json",
+  "$id": "https://typst.org/schema/json/package-manifest/v1/typst.schema.json",
```
elegaanz: Sorry if I was unclear, but the URI I gave was supposed to be the whole thing:

```suggestion
  "$id": "https://typst.org/schema/json/package-manifest/v1",
```

reknih (commented):
Please also add tests to `.github/workflows/ci.yml`:

In `tests/packages`, you can find a few example packages. Please run the schema against their manifests and expect it to pass. To do so, you can use e.g. [`dsanders11/json-schema-validate-action`](https://github.com/dsanders11/json-schema-validate-action).

You can add an additional package that exercises more of the Typst Universe-specific fields. However, let's not build the machinery to test against expected-wrong files now.

Since neither the test packages nor the schema frequently change, running CI on this often is a waste. Consider using [`dorny/paths-filter`](https://github.com/dorny/paths-filter) to only run the tests if either the manifests or the schema changed, and `ubuntu-slim` runners to conserve resources. Perform a sparse checkout using the checkout action if possible. All actions must be pinned to a commit SHA matching a published tag.

In a comment in the workflow file, document how to test the schema locally.

jassielof:
So, @reknih, I've added the job.

Now, rather than checking out the packages repository I was thinking of simply copying some 3 or 4 manifests as they are versioned either way.

The action you suggested for schema validation doesn't work against TOML so I'm using the Python one, this one natively supports TOML thanks to it being from the std. lib., I also didn't add a setup Python as the [Ubuntu Slim reference says that it's LTS and it's preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md), so it's mostly managed by Ubuntu LTS and its security stuff. There's the obvious `setup-python` action but I don't think there's much benefit (?), obviously I'll add it if you request it.

Lastly, the schema is just checked, not linted or thoroughly tested like Schema Store or other linting with Ajv, this can as strict or naggy with checks as you want it (from the time I previously submitted the Hayagriva one), the manifest schema is relatively simple so in a way it'd mostly provide no benefit for the time being, Hayagriva in any case would (and has been an issue for me, heh...)

reknih:
Does the default Schema Store lint block submission there? If so, please add it to CI to avoid unpleasant surprises. The Python thing is fine (but can you pin the installed package by hash)?

Please do not check out the packages repository. If you vendor packages or manifests into the Typst repository, you also need to update the NOTICE file if these packages require attribution in their license. And don't vendor packages under copyleft licenses :). Alternatively, you can just write one or two examples yourself.

.github/workflows/ci.yml
```diff
@@ -25,6 +25,44 @@ jobs:
...
+        id: changes
+        with:
+          filters: |
+            data:
+              - 'schema/typst.schema.json'
+              - 'tests/packages/**/typst.toml'
+      # Test locally by installing `check-jsonschema` from PyPI.
+      # Test that the schema itself is just valid with `--check-metaschema`.
+      # Test that the testing packages are valid with `--schemafile`.
+      - if: steps.changes.outputs.data == 'true'
```
reknih: Add a `name` key here since the automatically generated name is otherwise misleading.

jassielof:
The installed Python package can't be pinned as is unless a lock file is added, such as a `uv.lock`, `pylock`, or `requirements.txt`. As for the linting stuff, Ajv-CLI seems good for this, **but again** `npm install` doesn't have an option to install by hash unless there's a `package-lock.json`. I can try adding it under the `tools` directory, with `pyproject.toml`/`uv.lock` and `package.json`/`package-lock.json` and then it'd be installed globally to be used within the respective job. Ajv is basically what Schema Store heavily uses, and can be naggy on everything, then manually the checks can be reviewed from the Schema Store best practices.

jassielof:
Should be done now @reknih, I had to add npm and uv lock files, but dependencies are frozen at least, aside that it now uses the strict Ajv check which is better than the previous.

reknih:
Thanks, but one thing is still unclear to me: Do the checks Schema Store run match with the Ajv strict set?

The lockfile situation is a bit unfortunate, but at least they are not top-level. I defer to @laurmaedje's input on whether they carry their weight.

jassielof:
Checks should match, as they use Ajv strict too, but if we wanted to use theirs on this CI, it'd be needed to bring their custom CLI check and custom pre-commit checks too, you can check what they basically do is in https://github.com/SchemaStore/schemastore/blob/master/cli.js#L672C21-L672C21 and https://github.com/SchemaStore/schemastore/blob/master/cli.js#L810 basically. Hayagriva would be (and is) the one who's going to cause me more (is) causing me issues because of Ajv, at least this one is simple.

For example, the basic check passes:

```
schemastore on  master [+⇣] via  v26.7.0
❯ bun cli.js check --schema-name=typst.json
===== VALIDATE PRECONDITIONS =====
✔️ Directory structure conforms to expected layout
✔️ catalog.json validates against its schema
✔️ catalog.json has no fields that break guidelines
✔️ catalog.json has no duplicate "fileMatch" values
✔️ catalog.json has no invalid schema URLs
✔️ catalog.json has all local entries that exist in file total
✔️ schema-validation.jsonc validates against its schema
✔️ schema-validation.jsonc has no invalid schema names
✔️ schema-validation.jsonc has no invalid schema URLs
✔️ schema-validation.jsonc has no invalid skiptest[] entries
===== VALIDATE SCHEMAS =====
✔️ Completed "pre-checks"
✔️ Completed "Ajv validation"
===== REPORT =====
Out of 2156 TOTAL schemas:
- 977 (45%) are SchemaStore URLs
- 1179 (55%) are External URLs

Out of 935 TESTED schemas:
- 319 (34%) are validated with Ajv's strict mode
- 697 (75%) have tests
- 269 (29%) have negative tests

Out of 1 TESTED schemas:
- Total  2020-12: 0
- Total  2019-09: 0
- Total draft-07: 1
- Total draft-06: 0
- Total draft-04: 0
- Total draft-03: 0

```

But if we try the strict check

```
❯ bun cli.js check-strict --schema-name=typst.json
✖ Running "strict metaschema check" on file "./src\schemas\json\typst.json"
---
>> Failed to validate file "src\schemas\json\typst.json" against schema file "./src/schemas/json/metaschema-draft-07-unofficial-strict.json"
>> Showing first error out of 1 total error(s)
{
  instancePath: '/description',
  schemaPath: '#/pattern',
  keyword: 'pattern',
  params: { pattern: '\\nhttps?://.' },
  message: 'must match pattern "\\nhttps?://."'
}
---

```

It fails because it checks against its own custom draft https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/metaschema-draft-07-unofficial-strict.json, we could make Ajv compare it against this if you want.

laurmaedje:
The amount of surrounding infrastructure with Node.js + Python toolchains and lockfiles and the big-ish chunk of Actions YAML is a bit concerning to me (especially since we would be likely to duplicate all that in hayagriva). I want to look into alternative approaches, but I don't expect you to invest more time into this; I'll do so myself when I find a bit of time.

I'm sorry for the back and forth. We should have better aligned what we expect from this PR within the team before making you implement a particular testing strategy.

PR #7870, opened 2026-02-14
feat(pdf): add --no-embed-fonts CLI flag

### LINKED ISSUE #6466: Ability to disable font embedding
### Description

Mainly for legal reasons, you can't embed some fonts due to their license. Some fonts can be used in PDF, but not embedded, which hurts reproducibility, but keeps a court away.

One thing to note, is that, IIRC, some fonts can be used only in certain environments/software, so the unembedding won't help with that, I guess.

### Use Case

Use of proprietary fonts or any font that is not allowed to be (fully) embedded in the PDF file. I reckon this also can significantly drop the file size in some cases.

https://forum.typst.app/t/how-to-turn-off-typst-embed-font-to-the-output-pdf/4627

balcsida:
## Summary

- Adds `--no-embed-fonts` CLI flag to `typst compile` that produces PDFs without embedded font data
- Adds `no_embed_fonts: bool` field to `PdfOptions` and wires it through to krilla's `SerializeSettings`
- Handles new krilla `ValidationError` variants (`FontsNotEmbedded`, `InconsistentSeparationFallback`)
- Points krilla dependency to local fork with identity glyph mapping support

## Motivation

Font embedding accounts for the majority of PDF file size in text-heavy documents. For documents using only web-safe fonts (Arial, Courier New, Times New Roman), this flag reduces PDF size by ~70% (e.g., 120K → 33K for a resume) while maintaining correct rendering on systems with those fonts installed.

Depends on LaurenzV/krilla#332 for correct identity glyph mapping.

## Test plan

- [x] `typst compile --no-embed-fonts resume.typ` produces a valid PDF
- [x] Text renders correctly in macOS Preview, Acrobat Reader, and Chrome
- [x] No font warnings in Acrobat Reader when using web-safe fonts
- [x] Flag has no effect when not specified (default behavior unchanged)

Fixes #6466

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1434,7 +1434,7 @@ dependencies = [
 [[package]]
 name = "krilla"
 version = "0.6.0"
-source = "git+https://github.com/LaurenzV/krilla?rev=84c6332#84c6332a4d6459427b02bb40c8eb7e46b6a173d1"
+source = "git+https://github.com/balcsida/krilla?rev=a844a2f#a844a2f463c041df0bcdd8570f2fb64dd2e3f746"
 dependencies = [
  "base64",
  "bumpalo",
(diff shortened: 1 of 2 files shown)


laurmaedje:
Thanks for the PR, but unfortunately we have decided that we do not want to support this in Typst. I've [closed the corresponding issue](https://github.com/typst/typst/issues/6466).

For the future, please also keep in mind the following from [`CONTRIBUTING.md`](https://github.com/typst/typst/blob/main/CONTRIBUTING.md):

PR #7873, opened 2026-02-15
Infer dict type from spread operands in collection literals

### LINKED ISSUE #7795: Cannot spread dictionaries into empty dictionary literal

omair445:
## Problem

When a collection literal contains only spread items and no explicit key-value pairs, the parser defaults to treating it as an array. This means `(..dict1, ..dict2)` fails with:

```
error: cannot spread dictionary into array
```

The workaround is to use the `(:` prefix: `(: ..dict1, ..dict2)`, but this feels unintuitive since the compiler already knows the types of the spread operands at eval time.

## Fix

Instead of erroring when an array literal encounters a dictionary spread, the evaluator now detects this case and dynamically switches to building a dictionary result. This only happens when there are no positional (array) items — if you mix positional items with dict spreads, you still get a clear error.

**What works now:**
```typst
#let x = (a: 1, b: 2)
#let y = (c: 3)
#let merged = (..x, ..y)  // produces (a: 1, b: 2, c: 3)
```

**Still errors (as expected):**
- Spreading an array into a dict literal: `(..(1, 2), a: 1)`
- Mixing positional items with dict spreads: `(1, ..(a: 2))`

## Tests

Added test case `spread-only-dicts` covering the basic scenarios. All existing dict and array tests pass.

Closes #7795

### DIFF
--- a/crates/typst-eval/src/code.rs
+++ b/crates/typst-eval/src/code.rs
@@ -2,8 +2,8 @@ use ecow::{EcoVec, eco_vec};
 use typst_library::diag::{At, SourceResult, bail, error, warning};
 use typst_library::engine::Engine;
 use typst_library::foundations::{
-    Capturer, Closure, ClosureNode, Content, ContextElem, Dict, Func,
-    NativeElement, Selector, Str, Value, ops,
+    Capturer, Closure, ClosureNode, Content, ContextElem, Dict, Func, NativeElement,
+    Selector, Str, Value, ops,
 };
 use typst_library::introspection::{Counter, State};
 use typst_syntax::ast::{self, AstNode};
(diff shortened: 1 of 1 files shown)


laurmaedje:
I'm not _necessarily_ against making `(..dict)` work, I'm just a bit concerned on the AST level of how to represent that. Either, we'd need a new node or an `Array` is not actually guaranteed to be an array, which leaves me a bit uneasy. Also, what would need to remain either way is the `(:)` empty dict syntax, so we couldn't fully get rid of the colon syntax. 

To an extent, such an AST trouble often reflects on how easy it is to build a mental model for some piece of syntax, though I'm not sure how much this applies here. In any case, I'm not ready to move forward with this currently.

The PR description also looks a bit AI-y which is against our contribution guidelines.

PR #7874, opened 2026-02-16
Move multiline math handling to IR

mkorje:
As promised, here is the follow up to #7435 that moves the multiline handling to the IR.

The temporary regression https://github.com/typst/typst/pull/7435#issuecomment-3726965436 was actually a fix :sweat_smile:. I have no idea how this got through https://github.com/typst/typst/pull/7473, even though it clearly doesn't follow the description of the behaviour! Handling this in the IR has made this much easier to reason about. The actually correct this time matching the description description is

> The behavior of spacing around alignment points is subtle and differs from
> the `align` environment in amsmath. The current policy is:
> > always put the correct spacing between items separated by an alignment
> > point, and move the spacing between items in different columns of a
> > (right-aligned, left-aligned) pair to the right-aligned column
>
> This is handled in the `split_at_align` function.

There are two other test changes to note.

Previously, trailing linebreaks in `mat`/`vec`/`cases` wouldn't produce a warning, since the warning was emitted at a later stage after they were stripped. Now the warning is given, e.g.
```typ
--- math-mat-linebreaks-trailing paged ---
// Warning: 7-10 linebreaks are ignored in cells
// Hint: 7-10 use commas instead to separate each line
$ mat(a \ ) $
```

For some reason the `math-mat-align-explicit-mixed` test previously had this extra space in the last column. I can't see how this could be the intended behavior, nor can I remember what change I made exactly resolved this; nevertheless, it has been fixed.


crates/typst-library/src/math/ir/item.rs
```diff
@@ -447,24 +438,43 @@ pub struct GroupItem<'a> {
...
-    /// will follow the group of items.
-    pub(crate) fn create<I>(
-        items: I,
-        closing_exists: bool,
+    /// Rows correspond to linebreaks in the source. Columns within each row
+    /// correspond to alignment points. All rows are padded to have the same
+    /// number of columns.
+    pub rows: BumpVec<'a, BumpVec<'a, MathItem<'a>>>,
+    /// The number of columns originally in each row before padding.
+    pub row_lengths: &'a [usize],
```
mkorje:
```suggestion
    pub(crate) row_lengths: &'a [usize],
```

crates/typst-layout/src/math/fenced.rs
```diff
@@ -1,9 +1,9 @@
 use typst_library::diag::SourceResult;
 use typst_library::foundations::StyleChain;
 use typst_library::layout::{Abs, Axis};
-use typst_library::math::ir::{FencedItem, MathProperties};
+use typst_library::math::ir::{FencedItem, MathItem, MathProperties};
 
-use super::MathContext;
+use super::{MathContext, fragment::MathFragment};
```
laurmaedje: please split this up into separate imports, to avoid the nested one

tests/suite/math/text.typ
```diff
@@ -27,6 +27,12 @@ $bird "or" 🐦‍⬛$
 // Hint: 1-16 rendering may be poor
 $bird "or" 🐦‍⬛$
 
+--- math-text-styled-num paged ---
+// Test that styled numbers are treated the same as unstyled ones.
+$ a"123.4"b quad a "123.4" b $
+#show math.text: math.bold
```
laurmaedje:
```suggestion
#show text: math.bold
```
the `math.` doesn't really do anything, but can be a little confusing.

overall, this test is a tad cursed because `show text:` is always a tad cursed, but I guess it's okay to have some tests for it since it exists

tests/ref/pdf/hashes.txt
```diff
@@ -1216,7 +1222,7 @@ cb770b0275b8af4253d6b578b94d8d1f math-stretch-shorthand
 7dd24d787c4357b19a38722c2383b9cd math-stretch-vertical-attach
 c113f8791cf1d538ca3a5f2d33b54944 math-stretch-vertical-large-class
 595c703d9e102fc27435077b51c706a3 math-stretch-vertical-scripts
-f2796fce601c2cd7f6dbd9a6cd2fe294 math-style
+7f05d553e1ae544a8b37e32a937a6543 math-style
```
laurmaedje, line 1229: I took a look why this changed and the reason is that the letters in "𝟷𝟸𝟹" and "𝟺𝟻𝟼" are now each encoded separately instead of as a single text string. Similarly, in the SVG they each use a separate `<g>` group with a separate transform. Do you know why this happened?

laurmaedje: One interesting consequence is that Firefox does not recognize them as a consecutive word when trying to double-click-select. Though this seems to be a Firefox-specific quirk, other viewers do just fine.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -602,9 +614,12 @@ impl<'a> SkewedFractionItem<'a> {
 
 /// A 2D collection of math items laid out as a table/matrix.
 #[derive(Debug)]
+#[allow(clippy::type_complexity)]
 pub struct TableItem<'a> {
     /// The cells of the table, organized by row.
-    pub cells: BumpBox<'a, [BumpBox<'a, [MathItem<'a>]>]>,
+    ///
+    /// Each cell is split at alignment points into sub-columns.
+    pub cells: BumpBox<'a, [BumpBox<'a, [BumpBox<'a, [MathItem<'a>]>]>]>,
```
laurmaedje: This type is getting a little crazy. I think it might be worth factoring out some types to make this more readable. The bump box + slice setup makes it especially heavy, but even `Vec<Vec<Vec<MathItem<'a>>>` would be a bit much imo. The outer two are fine, but maybe math items split at alignment points could have their own type?

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -221,22 +235,46 @@ fn resolve_text<'a, 'v, 'e>(
...
+            *local_styles.get_or_insert_with(|| {
+                let local = [
+                    TextElem::top_edge.set(TopEdge::Metric(TopEdgeMetric::Bounds)),
+                    TextElem::bottom_edge
+                        .set(BottomEdge::Metric(BottomEdgeMetric::Bounds)),
+                    TextElem::overhang.set(false),
+                ]
+                .map(|p| p.wrap());
+                ctx.chain_styles(styles, local)
+            })
```
laurmaedje: This + the option could be replaced with a `LazyCell`. In fact the lazy cell could even be `static`.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -837,26 +844,26 @@ impl<'a> PrimesItem<'a> {
 pub struct TextItem<'a> {
     /// The text content.
     pub text: &'a str,
+    /// Whether the text is a number.
+    pub num: bool,
```
laurmaedje: What do you think about already creating a `NumItem` instead? Even if we don't have one on the Typst side yet. Seems a bit cleaner to me.

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -221,22 +235,46 @@ fn resolve_text<'a, 'v, 'e>(
...
-        styles,
-        elem.span(),
-        &ctx.arenas.bump,
-    ));
+    let item = if !elem.text.contains(is_newline) {
+        create_item(&elem.text)
+    } else {
+        let bump = &ctx.arenas.bump;
+        let rows = BumpVec::from_iter_in(
+            elem.text.split_terminator(is_newline).map(|line| {
```
laurmaedje: Upon testing
```typ
$ "L""L" $
$ "L\n""L" $
```
seems to still behave the same, but I'm a bit confused why since this is using `split_terminator`. Do you have any insight for me?

crates/typst-library/src/math/ir/item.rs
```diff
@@ -237,6 +237,16 @@ impl<'a> MathItem<'a> {
...
 
+    /// If this is a multiline item, sets the align field to true.
+    pub(crate) fn with_multiline_align(mut self) -> Self {
+        if let Self::Component(comp) = &mut self
+            && let MathKind::Multiline(multiline) = &mut comp.kind
+        {
+            multiline.align = true;
+        }
+        self
+    }
```
laurmaedje: I think the name is not super clear. Something like `with_multiline_centering` would be much more immediate. The fields could be called `centered` or something.

crates/typst-layout/src/math/mod.rs
```diff
@@ -124,9 +125,15 @@ pub fn layout_equation_block(
     let item = resolve_equation(elem, engine, &mut locator, &arenas, styles)?;
 
     let mut ctx = MathContext::new(engine, &mut locator, regions.base(), font.clone());
-    let full_equation_builder = ctx
-        .layout_into_fragments(&item, styles)?
-        .multiline_frame_builder(styles);
+    let full_equation_builder = if let MathItem::Component(MathComponent {
+        kind: MathKind::Multiline(multi),
+        ..
```
laurmaedje, line 127: is it safe to just ignore the props and styles here?

laurmaedje: also, the pattern of matching on the item in some specific cases (here and in L/R) even though `layout_into_fragments` also works seems easy to miss in case it's needed elsewhere. maybe it's fine, but let's discuss this syncly.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -957,6 +964,77 @@ impl<'a> ExternalItem<'a> {
...
+}
+
+/// The body of a [`FencedItem`].
+#[derive(Debug)]
+pub enum FencedBody<'a> {
+    /// Owned body.
+    Owned(MathItem<'a>),
+    /// Shared body stored in [`SharedFenceSizing`].
+    Shared { index: usize, sizing: Rc<SharedFenceSizing<'a>> },
```
laurmaedje, line 1006: The whole ref-counting + interior mutability thing seems a bit dangerous as it could easily lead to sizing information being mutated and then read from a place that's not supposed to observe it. Even if that doesn't happen today, I'm overall a bit concerned with such a pattern. Let's discuss this syncly as well.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -36,12 +39,49 @@ pub enum MathItem<'a> {
...
+    pub(crate) fn wrap(
+        mut items: BumpVec<'a, MathItem<'a>>,
+        styles: StyleChain<'a>,
+    ) -> MathItem<'a> {
+        if items.len() == 1 {
+            items.pop().unwrap()
+        } else {
+            GroupItem::create(items.into_boxed_slice(), styles)
+        }
+    }
```
laurmaedje: it would be nice to have some guidance in the doc comments as to when `group` should be used vs `wrap`

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -877,20 +915,91 @@ fn resolve_lr<'a, 'v, 'e>(
...
+        ctx.items.insert(insert_pos, item);
+    }
 
-    ctx.items.insert(start + start_idx, item);
     Ok(())
 }
 
+/// Splits a fenced item's body by alignment points and linebreaks into
+/// segments.
+fn split_into_fenced_segments<'a>(
```
laurmaedje: This code looks a bit like it's basically undoing the previous preprocessing. Is that accurate? If yes, it feels a bit unfortunate.

crates/typst-library/src/math/ir/item.rs
laurmaedje, line 1: A few general notes on the bump allocation: 
- It does add quite a bit of noise to the code. 
- Moreover, when multiple `BumpVec` are created at the same time (e.g. with a nested loop), the reallocation can cause a fair amount of "dead memory" (since the arena only ever grows forward). Though, I suppose this is limited to 2x overhead since Vec grows exponentially. And you're also using `with_capacity` quite actively, which is good. 
- Finally, I am concerned with the lack of type safety around memory leaks. It would be fairly easy to accidentally introduce a leak.

In any case, I was wondering whether you had measured how much the bump allocation overall gives us compared to plain `Vec` in terms of speedup (ideally we'd also compare memory consumption, but that's not so easy)? I just wonder whether it's really pulling its weight.

The existing uses of bump allocation outside of math fall into two categories:
- Realization: Here, it's actually needed to make it work lifetime-wise. This one is not about performance.
- Flow layout: Here, it's just an optimization iirc, but I think it predates the benchmark infrastructure and I'm not sure if/how much I actually benchmarked it.

If the bump allocation _is_ a big win perf-wise, the leak safety thing is solvable I think. It could probably be made safer by writing a thin wrapper over bumpalo, but that's very orthogonal to this PR of course.

crates/typst-layout/src/math/fenced.rs
```diff
@@ -44,14 +43,14 @@ pub fn layout_fenced(
         ctx.push(open);
     }
 
-    // Check if the body needs re-layout, since stretch info was updated after
-    // initial layout.
-    if has_mid_stretched {
-        let body = ctx.layout_into_fragments(&item.body, styles)?;
-        ctx.extend(body);
+    // Check if the body needs re-layout, since stretch info was updated or we
+    // deferred initial layout.
```
laurmaedje, line 48: I didn't realize this during the original PR, but I think this is a bit unsound since it performs relayout (and with the earlier stuff potentially also laying out out of order?) without manually managing `Locator`s, rather relying on the `&mut MathContext` to produce new locators. this can subtly mess with introspection.

is relayout here truly necessary or is it more an artifact of the current implementation?

crates/typst-layout/src/math/run.rs
```diff
@@ -1,149 +1,126 @@
...
-            let mut pos = Point::with_y(size.y);
-            if alignments.points.is_empty() {
-                pos.x = align.position(alignments.width - sub.width());
-            }
-            size.x.set_max(sub.width());
-            size.y += sub.height();
-            frames.push((sub, pos));
+/// Build and stack line frames for rows of aligned sub-columns.
+pub fn stack_sub_column_rows(
+    rows: impl IntoIterator<Item = (Vec<Vec<MathFragment>>, (Abs, Abs), (Abs, Abs))>,
```
laurmaedje: this type is getting pretty crazy, too

crates/typst-layout/src/math/run.rs
```diff
@@ -1,149 +1,126 @@
...
+}
 
-            let mut pos = Point::with_y(size.y);
-            if alignments.points.is_empty() {
-                pos.x = align.position(alignments.width - sub.width());
-            }
-            size.x.set_max(sub.width());
-            size.y += sub.height();
-            frames.push((sub, pos));
+/// Build and stack line frames for rows of aligned sub-columns.
```
laurmaedje: the term "sub-column" vs just "column" (or I guess "cell" if it's just the part in a single row?) isn't entirely clear to me

crates/typst-layout/src/math/run.rs
```diff
@@ -255,56 +229,71 @@ impl MathRunFrameBuilder {
...
-    for i in 1..points.len() {
-        let prev = points[i - 1];
-        points[i] += prev;
-    }
-    AlignmentResult {
-        width: points.last().copied().unwrap_or(pending_width),
-        points,
+/// Build a frame from sub-column fragments positioned at alignment points.
+fn sub_columns_into_line_frame(
+    sub_cols: Vec<Vec<MathFragment>>,
```
laurmaedje: I think the inner `Vec<MathFragment>` might really benefit from its own type, too, after all. I recall that a `MathRun` type existed at some point, but not quite sure what the motivation was to remove it.

crates/typst-library/src/math/ir/item.rs
mkorje: Sounds like a good idea. Will add it.

crates/typst-layout/src/math/run.rs
mkorje: The term exists to describe the situation in mat/vec/cases, where you can use alignment points to align cells in a given column. So we are breaking up cells further into a vec of items broken up by alignment points. It's not exactly a great name, but all I could come up with at the time.

crates/typst-library/src/math/ir/resolve.rs
mkorje: It isn't actually the newline causing it to shift, but the centering on the math axis!
```typ
$ "L\n\n""L\n""L" $
```
Should just need to make the centering conditional on there actually being more than one row after trimming a trailing linebreak.

crates/typst-library/src/math/ir/item.rs
mkorje: I went with `NumberItem` instead of abbreviating it.

crates/typst-layout/src/math/fenced.rs
mkorje: It is a consequence of the current implementation where the stretching information is resolved at once, so that in typst-layout when we lay get to laying out a `GlyphItem` we know all the stretch information, and so we never need to re-stretch something.
Here it's needed for `mid()` items since we only get to update their stretch info after an initial layout of the body, and with that set we need to relayout.


As an aside, I tried using the codex review thing a while ago and it actually pointed this out, but I dismissed it as I thought it wasn't an issue :sweat_smile: :
> **P1 Badge Avoid consuming locators during shared fence sizing**
> 
> When `item.body.sizing()` is present, `relative_to_from_sizing` measures delimiter height by calling `ctx.layout_into_fragments(item, styles)` for **every** shared body cell, but those cells are then laid out again for actual output in `layout_fenced`. Because this measurement pass runs on the live `MathContext`, any cell content that allocates locator entries (e.g. text/box/external content with tags) advances locator state twice, which can shift or duplicate location IDs for subsequent rendered content in multiline `lr` expressions.
> 
> Useful? React with 👍 / 👎.

tests/ref/pdf/hashes.txt
mkorje: It's because previously the check for whether text is a number happened after math styling, so "𝟷𝟸𝟹" wouldn't go through the "Small optimization for numbers." path in `layout_text`. I figured the behaviour now is more consistent. (Actually, this is maybe bad because I have this comment `This won't panic as ASCII digits and '.' will never end up as nothing after shaping` in that path...) But this means that all numbers in math have had each letter encoded separately in PDF/SVG...

crates/typst-library/src/math/ir/resolve.rs
mkorje: Yes, that is accurate. IIRC I did it this way as it seemed to be the simplest thing, otherwise I'd need to basically duplicate the `MathItem::group()` + `build_multiline` logic and make it somewhat custom to work for fenced things. I can try doing that though and see how it looks.

crates/typst-layout/src/math/mod.rs
mkorje: > is it safe to just ignore the props and styles here?

Oops the styles shouldn't, but props is fine as it isn't needed in `built_multiline`. I've fixed this.

> also, the pattern of matching on the item in some specific cases (here and in L/R) even though

This is really the only place where it's used I think (its not in L/R anywhere?), and it's needed cause we need to get a MathRunFrameBuilder, and so can't use the any of the `layout_into_...` things.

crates/typst-library/src/math/ir/item.rs
mkorje: I've gone with a `TableCell` type alias for `[BumpBox<'a, [MathItem<'a>]>]`.

crates/typst-layout/src/math/run.rs
mkorje: I've added a type alias for `Vec<MathFragment>` called `MathRun`, and added to its doc comment an explanation of the sub-column thing and have used it throughout. Lmk if it's clearer now.

crates/typst-layout/src/math/run.rs
mkorje: Along with the other changes described below, I've made a struct to represent this type.

crates/typst-library/src/math/ir/item.rs
mkorje: I ran the benchmark on this PR vs this PR with a commit switching to plain old `Vec` everywhere (https://github.com/mkorje/typst/commit/69d94006c0e4df2fada9d9f663841941158cae69), see runs 187 & 188.

crates/typst-library/src/math/ir/resolve.rs
mkorje: I've made the change, see the latest commit.

crates/typst-library/src/math/ir/item.rs
```diff
@@ -71,6 +74,10 @@ impl<'a> MathItem<'a> {
 
     /// Wraps the given items into a group item, or returns the single item if
     /// there is only one.
+    ///
+    /// Use this function instead of [`Self::group`] when the items you have
```
laurmaedje: I wonder whether the already processed / not already processed state could be represented at the type system level

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -236,27 +249,15 @@ fn resolve_text<'a, 'v, 'e>(
     let italic = styles.get(EquationElem::italic).or(Some(false));
 
     // Create item with correct styles and properties.
-    let mut local_styles = None;
-    let mut create_item = |text: &str| {
+    let local_styles =
+        LazyCell::new(|| ctx.chain_styles(styles, TEXT_BASE_LOCAL_STYLES.clone()));
```
laurmaedje: I think we could immediately chain here (without `ctx.chain_styles`) and avoid the the `clone()`

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -261,12 +261,13 @@ fn resolve_text<'a, 'v, 'e>(
         TextItem::create(styled_text, num, styles, elem.span(), &ctx.arenas.bump)
     };
 
-    let item = if !elem.text.contains(is_newline) {
-        create_item(&elem.text)
+    let text = elem.text.strip_suffix(is_newline).unwrap_or(&elem.text);
```
laurmaedje: I think this would benefit from a comment. 

I also wonder how this code path would react to `\r\n`, which is in principle supported in Typst.

crates/typst-layout/src/math/text.rs
```diff
@@ -23,34 +23,45 @@ pub fn layout_text(
...
+    props: &MathProperties,
+) -> SourceResult<()> {
+    let text = item.text;
+    let span = props.span;
+    // Small optimization for numbers. Note that this lays out slightly
+    // differently to normal text and is worth re-evaluating in the future.
+    let mut fragments = vec![];
+    for c in text.chars() {
+        // This won't panic as ASCII digits and '.' will never end up as
+        // nothing after shaping.
```
laurmaedje: Yeah, that's a bit of an issue with the other change. I'd prefer to have some non-panicking solution.

crates/typst-layout/src/math/run.rs
```diff
@@ -14,6 +14,15 @@ use super::fragment::MathFragment;
...
 const TIGHT_LEADING: Em = Em::new(0.25);
 
+/// A list of math fragments between alignment points and/or linebreaks.
+///
+/// For multiline equations this represents a distinct "cell", the list of
+/// fragments at a specific row and column. For tables, this represents a
+/// "sub-column", one of the parts that make up a cell in a table since
+/// alignment points can be used to align fragments within a cell for an
+/// individual column.
+pub type MathRun = Vec<MathFragment>;
```
laurmaedje, line 24: with the two type aliases, it's already much more readable, but unfortunately an alias does not help with type safety.

I think it might make sense to turn this into a newtype instead. Moreover, the `TableCell` construct from the other commit might make sense as a generic construct for "runs split at alignment points".

crates/typst-layout/src/math/mod.rs
laurmaedje: > its not in L/R anywhere?

It used to be in `resolve_lr`, but it's gone as of https://github.com/typst/typst/pull/7874/commits/fae9377638aa00fa97cd7a0d34e9cf429dc9ba5a

mkorje: ah, okay

crates/typst-library/src/math/ir/resolve.rs
laurmaedje: might be able to use `split_newlines`

crates/typst-library/src/math/ir/resolve.rs
laurmaedje: basically this
```rust
let local_styles = ctx.arenas.bump.alloc(styles).chain(&*TEXT_BASE_LOCAL_STYLES);
```

laurmaedje: in combination
```rust
let local_styles = ctx.store_chain(styles).chain(&*TEXT_BASE_LOCAL_STYLES);
```
and then
```rust
/// Lifetime-extends some style chain.
fn store_chain<'c>(&self, styles: StyleChain<'c>) -> &'a StyleChain<'c> {
    self.arenas.bump.alloc(styles)
}
```

crates/typst-library/src/math/ir/item.rs
```diff
@@ -20,8 +19,47 @@ use crate::math::{
...
+        }
+    }
+
+    /// Unwraps this into its inner [`MathItem`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if this is a [`Linebreak`](RawMathItem::Linebreak) or
+    /// [`Align`](RawMathItem::Align).
+    pub(crate) fn into_item(self) -> MathItem<'a> {
```
laurmaedje: I think a name like `unwrap_item` or just `unwrap` would make the panicky nature clearer.

or honestly, I'd probably go with returning an `Option` and then `.into_item().unwrap()`

mkorje: I've gone with returning an `Option`.

mkorje:
All ready to merge on my end now.

laurmaedje:
Thank you!

Issue #7668, opened 2026-01-09
Inclusion of png in pdf without reencoding


Enivex:
### Description

As evidenced by https://github.com/josch/img2pdf, it is possible to directly embed raw png data in pdf without decoding and re-encoding. This has some benefits

- Performance
- Smaller file sizes (possibly significantly so) when using optimized pngs (e.g. from https://github.com/oxipng/oxipng )

(Could also be relevant for some other formats if typst currently does not do so)

> For JPEG, JPEG2000, non-interlaced PNG, TIFF images with CCITT Group 4 encoded data, and JBIG2 with single-page generic coding (e.g. using jbig2enc), img2pdf directly embeds the image data into the PDF without re-encoding it. It thus treats the PDF format merely as a container format for the image data. In these cases, img2pdf only increases the filesize by the size of the PDF container (typically around 500 to 700 bytes). Since data is only copied and not re-encoded, img2pdf is also typically faster than other solutions for these input formats.

jqwv:
This was considered as [not possible for JPEG](https://github.com/typst/typst/issues/5278#issuecomment-3116724364) because the file can be invalid. Without decoding it, it would not be possible to validate if "file.jpg" or "file.png" are expected image files.
As mentioned in the linked issue, currently an inconsistent "hack" exists to already allow what you request: embed the file in a PDF, then all next times embed the PDF instead of the PNG/JPG. For PDF embedding, the file contents are currently not checked and just copied instead, even if the PDF is invalid.

Related to this (because not many people know this): PNG and JPG have the same API (iteration by scanlines). A very efficient implementation could subsample an image (for cropped images, or even scaled images) without decoding it completely. This is done by skipping each Y scanlines and each X columns. The problem is that only the C implementations support this. The rust image people all have ignored this. As a result, decoding JPG and PNG in Rust always requires a full decompression, I think even to RGBA even if it's greyscale or monochrome, even if we could crop and/or scale an image from a big 50 MB image with only constant memory requirement.

Enivex:
Even if you need to verify that an image is valid by decoding, you can still embed after doing that instead of reencoding.

jqwv:
Yes, this is what is done with JPG. But _decoding_ the image is required. Reencoding is not.
Your issue asked for "without decoding".

Enivex:
> Yes, this is what is done with JPG. But _decoding_ the image is required. Reencoding is not. Your issue asked for "without decoding".

[...]

PR #7876, opened 2026-02-16
typst-pdf: Attach known image types without compression

krkk:
Most of the image files already have the data compressed, so sprinkling deflate on top of it won't do much besides making it take longer to compile the document.

Initially, #6256 had it right, but then it got flipped during [refactoring]. I believe it wasn't intentional, hence my PR. :^)
(interesting that clippy doesn't catch equal match cases)

[refactoring]: https://github.com/typst/typst/pull/6256/changes/cd659bf460e820ac1c2eadacd34e007d135294d8#diff-2615574cce919d378aeda60ca1604075f0665af57065a39286a0e363578eae78L107-R107

### DIFF
--- a/crates/typst-pdf/src/attach.rs
+++ b/crates/typst-pdf/src/attach.rs
@@ -113,7 +113,7 @@ fn should_compress(data: &[u8]) -> Option<bool> {
             | "image/heif"
             | "image/avif"
             | "image/jxl"
-            | "image/vnd.djvu" => None,
+            | "image/vnd.djvu" => Some(false),
             _ => None,
         },
         infer::MatcherType::Text => None,


krkk:
waaaaait this is only for `#pdf.attach()`, `#image()` has its own code path. huhhhh, so i guess its almost useless now.
oh well  ¯\\_(ツ)_/¯

saecki:
> Most of the image files already have the data compressed, so sprinkling deflate on top of it won't do much besides making it take longer to compile the document.
> 
> Initially, #6256 had it right, but then it got flipped during [refactoring](https://github.com/typst/typst/pull/6256/changes/cd659bf460e820ac1c2eadacd34e007d135294d8#diff-2615574cce919d378aeda60ca1604075f0665af57065a39286a0e363578eae78L107-R107). I believe it wasn't intentional, hence my PR. :^) (interesting that clippy doesn't catch equal match cases)

Hmm, have you noticed or measured the performance impact of the deflate compression? Because in my testing I've had some cases where images weren't very efficiently encoded and this made quite the difference in size. Note that `None` in this case means: use automatic compression, and if the threshold isn't met the image is written as is.

laurmaedje:
> Note that None in this case means: use automatic compression, and if the threshold isn't met the image is written as is.

That might be worth a comment, because I also thought it's a typo 😅

saecki:
Yeah, maybe we can also return a `Smart<bool>` from `should_compress` :thinking:

Enivex:
> waaaaait this is only for `#pdf.attach()`, `#image()` has its own code path. huhhhh, so i guess its almost useless now.
> oh well  ¯\\_(ツ)_/¯

Relevant https://github.com/typst/typst/issues/7668

laurmaedje:
Since this is intentional, let's close this. We could switch to `Smart` in a separate PR. 

Thanks for the PR anyway!

PR #7877, opened 2026-02-16
Ensure lorem produces exact word count by skipping hyphens

### LINKED ISSUE #6186: #lorem has a glitch at the 194th word
### Description

When experimenting with word counting (using `typst-wordometer`), I noticed that `#lorem` produces an empty word with a dot at position 194 (counting from 1), and a hyphen wenn generating at least 195 words on that position. Word counting will usually not count the hypen and neither the empty sentence, so word counters's result differs from the number of words supplied to `#lorem`.

Steps to reproduce:
```typst
#lorem(193)

#lorem(194)

#lorem(195)
```

I believe this to happen again at any position where a hyphen is in the text.

Expected behavior: In my opinion, hyphens should not count as "words", or the documentation should make it clear that `#lorem` can not actually be trusted to produce the exact number of words specified.

Suggested fix: Return hypens additionally, without counting them as words.

### Reproduction URL

https://typst.app/project/rspchq7jyWPp2u0Ffhy6jx

### Operating system

Web app

### Typst version

3w36zj6:
close #6186

This issue was initially marked as an upstream problem, but the author of lipsum suggested using `MarkovChain::iter`. 
Therefore, this PR adds a custom implementation to Typst that takes the word count into account.

```typ
#import "@preview/wordometer:0.1.4": word-count

#word-count(total => [
  #lorem(193) \
  #[`#lorem(193)`: *#total.words words*] <no-wc>
], exclude: <no-wc>)

#word-count(total => [
  #lorem(194) \
  #[`#lorem(194)`: *#total.words words*] <no-wc>
], exclude: <no-wc>)

#word-count(total => [
  #lorem(195) \
  #[`#lorem(195)`: *#total.words words*] <no-wc>
], exclude: <no-wc>)
```

image

### DIFF
--- a/crates/typst-library/src/text/lorem.rs
+++ b/crates/typst-library/src/text/lorem.rs
@@ -1,5 +1,6 @@
 use std::sync::LazyLock;
 
+use ecow::EcoString;
 use lipsum::{LIBER_PRIMUS, LOREM_IPSUM, MarkovChain};
 
 use crate::foundations::{Str, func};
@@ -24,22 +25,21 @@ pub fn lorem(
     /// The length of the blind text in words.
     words: usize,
 ) -> Str {
-    generate_lorem(words).into()
+    lorem_impl(words).into()
 }
 
-/// Generate `n` words of lorem ipsum text, treating `--` as a non-word
+/// Generates `n` words of Lorem Ipsum text, treating `--` as a non-word
 /// separator that is replaced with an en-dash (`–`).
 ///
 /// This reimplements the joining logic from lipsum's private `join_words`
 /// function, but skips `--` entries in the Markov chain output without
 /// counting them toward the requested word count.
-fn generate_lorem(n: usize) -> String {
-    // Ported from https://github.com/mgeisler/lipsum/blob/0.9.1/src/lib.rs
-    // MIT License
-    // Copyright (c) 2017 Martin Geisler
+fn lorem_impl(n: usize) -> EcoString {
+    // Based on the `lipsum` crate (MIT).
+    // Copyright (c) 2017 Martin Geisler.
     // See NOTICE for full attribution.
     if n == 0 {
-        return String::new();
+        return EcoString::new();
     }
 
     static LOREM_CHAIN: LazyLock<MarkovChain<'static>> = LazyLock::new(|| {
@@ -55,7 +55,7 @@ fn generate_lorem(n: usize) -> String {
     // Punctuation characters which end a sentence.
     const PUNCTUATION: [char; 3] = ['.', '!', '?'];
 
-    let mut sentence = String::new();
+    let mut sentence = EcoString::new();
     let mut word_count = 0;
     let mut needs_cap = false;
 
--- a/NOTICE
+++ b/NOTICE
@@ -20,8 +20,10 @@ The MIT License applies to:
   based on the `TokenSet` from rust-analyzer
   (https://github.com/rust-lang/rust-analyzer/blob/master/crates/parser/src/token_set.rs)
 
-* The lorem ipsum generator in `crates/typst-library/src/text/lorem.rs` which
-  is based on the `lipsum` crate (https://github.com/mgeisler/lipsum)
+* The Lorem Ipsum generator in `crates/typst-library/src/text/lorem.rs` which
+  is based on the `lipsum` crate
+  (https://github.com/mgeisler/lipsum/blob/0.9.1/src/lib.rs)
+  Copyright (c) 2017 Martin Geisler
 
 The MIT License (MIT)
 
--- a/tests/suite/text/lorem.typ
+++ b/tests/suite/text/lorem.typ
@@ -37,4 +37,3 @@
 #test(count(193), 193)
 #test(count(194), 194)
 #test(count(195), 195)
-


crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+}
+
+// https://docs.rs/lipsum/0.9.1/src/lipsum/lib.rs.html#403-413
+thread_local! {
+    static LOREM_CHAIN: RefCell<MarkovChain<'static>> = RefCell::new({
+        let mut chain = MarkovChain::new();
+        chain.learn(LOREM_IPSUM);
+        chain.learn(LIBER_PRIMUS);
+        chain
+    });
```
istudyatuni: If I understand correctly, this code will run on each invocation of typst-cli/downstream tools? Maybe it's better to use `OnceCell` to defer initialization to first use of `generate_lorem`?

saecki: No, this code will be lazily run once on every thread. Though depending on how the rayon manages its thread-pool this may still be a waste of resources. Instead this should use a `LazyLock` like most other parts of the Typst code base.

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+                sentence.push('\u{2013}');
+                first = false;
+                continue;
+            }
+
+            word_count += 1;
+
+            if first {
+                sentence.push_str(&capitalize(word));
+                needs_cap = sentence.ends_with(punctuation);
```
saecki: This check can be pulled out of the if statement, since it's present in both branches. I'd just put it at the end of the loop.

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+        let mut needs_cap = false;
+        let mut first = true;
+
+        while word_count < n {
+            let Some(word) = iter.next() else { break };
+
+            // Skip `--` without counting it as a word; append an en-dash
+            // to the output instead.
+            if word == "--" {
+                if !first {
```
saecki: This check can be lifted out of the if statement as the first thing right after getting the next word. Since the first word will always be `Lorem`, we can remove the `first` variable and use `word_count == 0`/`word_count > 0` instead.

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+                    sentence.push(' ');
+                }
+                sentence.push('\u{2013}');
+                first = false;
+                continue;
+            }
+
+            word_count += 1;
+
+            if first {
```
saecki: Since `"Lorem"` is already captialized, we can remove this branch of the if statement.

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+                sentence.push_str(&capitalize(word));
```
saecki: Now that this is the only occurence of `capitalize`, I would just inline it and push directly into the `sentence` string, instead of creating an intermediate `String` buffer.

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+        return String::new();
+    }
+
+    // https://docs.rs/lipsum/0.9.1/src/lipsum/lib.rs.html#344-382
+    LOREM_CHAIN.with(|chain| {
+        let chain = chain.borrow();
+        let mut iter = chain.iter_from(("Lorem", "ipsum"));
+
+        // Punctuation characters which end a sentence.
+        let punctuation: &[char] = &['.', '!', '?'];
```
saecki: For readability I would make this a constant, and while we're at it I'd convert it to an array : `[char; 3]` instead.

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
...
+            if word == "--" {
+                if !first {
+                    sentence.push(' ');
+                }
+                sentence.push('\u{2013}');
+                first = false;
+                continue;
+            }
+
+            word_count += 1;
```
saecki: Finally I'd also move this to the end of the loop

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,96 @@ pub fn lorem(
     /// The length of the blind text in words.
     words: usize,
 ) -> Str {
-    lipsum::lipsum(words).replace("--", "–").into()
+    generate_lorem(words).into()
+}
+
+// https://docs.rs/lipsum/0.9.1/src/lipsum/lib.rs.html#403-413
+thread_local! {
+    static LOREM_CHAIN: RefCell<MarkovChain<'static>> = RefCell::new({
```
saecki: To make attributing everything easier, could you move this static into the `generate_lorem` function, and then add a copyright notice like this in the code: https://github.com/typst/typst/blob/c817e592b735ef77af5eed079d411369cf433362/crates/typst-utils/src/scalar.rs#L46
And add an entry here: https://github.com/typst/typst/blob/c817e592b735ef77af5eed079d411369cf433362/NOTICE#L22

crates/typst-library/src/text/lorem.rs
```diff
@@ -61,59 +42,64 @@ fn generate_lorem(n: usize) -> String {
...
 
-            if first {
-                sentence.push_str(&capitalize(word));
-                needs_cap = sentence.ends_with(punctuation);
-                first = false;
-            } else {
+        // Skip `--` without counting it as a word; append an en-dash
+        // to the output instead.
+        if word == "--" {
+            if word_count > 0 {
```
saecki: Could you move this check above the `if word == "--"` statement and then remove the duplicate one below?

crates/typst-library/src/text/lorem.rs
```diff
@@ -61,59 +42,64 @@ fn generate_lorem(n: usize) -> String {
...
-        sentence
-    })
+        if needs_cap {
+            // Capitalize the first character in a string.
+            let idx = match word.chars().next() {
+                Some(c) => c.len_utf8(),
+                None => 0,
+            };
+            sentence.push_str(&word[..idx].to_uppercase());
+            sentence.push_str(&word[idx..]);
```
saecki: Sorry for not noticing in the first round, but there was still an unnecessary allocation here, that can be avoided quite idiomatically:
```suggestion
            // Capitalize the first character of the word.
            if let Some(c) = word.chars().next() {
                sentence.extend(c.to_uppercase());
                sentence.push_str(&word[c.len_utf8()..]);
            }
```

crates/typst-library/src/text/lorem.rs
```diff
@@ -20,5 +24,82 @@ pub fn lorem(
...
     words: usize,
 ) -> Str {
-    lipsum::lipsum(words).replace("--", "–").into()
+    generate_lorem(words).into()
+}
+
+// Ported from https://github.com/mgeisler/lipsum/blob/0.9.1/src/lib.rs
+// MIT License
+// Copyright (c) 2017 Martin Geisler
+// See NOTICE for full attribution.
```
saecki: This is just personal preference, but I think the comment looks a bit odd above the doc comment. Could you move it as the first thing inside the function?

saecki (approved):

laurmaedje:
Thanks!

PR #7878, opened 2026-02-16
Add mise installation instructions to README

3w36zj6:
[mise](https://mise.jdx.dev/) is a tool manager that has recently gained widespread use. Like Nix or Docker, it enables reproducible builds by pinning specific versions on a per-project basis. Furthermore, mise makes it easy to integrate with the Typst ecosystem (such as Tinymist and typstyle), as well as with external tools for data acquisition and processing. 

Many popular tools in recent years provide installation instructions using mise. Therefore, I think it's helpful to mention mise in the README.


laurmaedje:
Thanks for the PR, but this is too niche for inclusion in the README.

PR #7882, opened 2026-02-17
Allow Fractional widths in block elements

### LINKED ISSUE #7880: Fractional units should convert to relative units when used as a block width
### Description

Currently: A block cannot have a fractional width but can have a relative width

This rules seems strange:
If a user of a block uses `width: 1fr`, this could be transparently converted to `width: 100%`. This is what I would expect, Since blocks are always the same thing on their line, `2fr`, `3fr`, etc., should be safely convertible to `100%`.

### Use Case

I am writing a template for an exam class. I want the user to be able to specify an "answer box". I want it to be as flexible as possible. Currently I must inspect their `width` argument and convert it manually to the appropriate unit type, but I don't see a reason why this should be the case.

siefkenj:
Fixes #7880

Based on my reading of the code, any fractional width for a block is already treated as a 100% relative width. This PR makes fractional widths valid for a block (as well as a few other types that depended on the width type of block).

### DIFF
--- a/crates/typst-layout/src/flow/block.rs
+++ b/crates/typst-layout/src/flow/block.rs
@@ -29,7 +29,7 @@ pub fn layout_single_block(
     let inset = elem.inset.resolve(styles).unwrap_or_default();
 
     // Build the pod regions.
-    let pod = unbreakable_pod(&width.into(), &height, &inset, styles, region.size);
+    let pod = unbreakable_pod(&width, &height, &inset, styles, region.size);
 
     // Layout the body.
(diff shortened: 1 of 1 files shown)


laurmaedje:
I'm happy to consider the design proposal, but not right away and I'm not ready to commit to anything now. I'd prefer to let this sit for a while as I have some tangentially related thoughts about fr, which may interact with this in the future (see https://github.com/typst/typst/issues/7880#issuecomment-3914432511). For this reason, I'll close the PR, but thanks for the contribution nonetheless.

PR #7884, opened 2026-02-17
Fix where selector matching for auto on synthesized

### LINKED ISSUE #7825: Inconsistent behavior of auto in selectors
### Description

I think the issue here is best shown with an example
```typst
#[
  #show heading.where(supplement: auto): "Case 1 Matched"
  // explicit auto
  // does not match
  #heading(supplement: auto)[Case 1]
  // does not match
  // auto by default
  = Case 1
]

#[
  #show heading.where(supplement: [Section]): "Case 2 Matched"
  // explicit auto
  // does match
  #heading(supplement: auto)[Case 2]
  // auto by default
  // does match
  = Case 2
]

#[
  #show heading.where(bookmarked: auto): "Case 3 Matched"
  // explicit auto
  // does match
  #heading(bookmarked: auto)[Case 2]
  // auto by default
  // does match
  = Case 3

  // and for anyone wondering, setting the selector to use `true` or `false` does not match
]
```
So, the behavior of `auto` in `where` selectors differs between `supplement` and `bookmarked`.

Since there are so many parameters which use `auto`, I have not been through them all to figure out which ones have which behavior.

I don't think this is documented anywhere.

### Use Case

I think it would be quite nice to be able to write `heading.where(supplement: auto)` instead if `heading.where(supplement: [Section])`, because it is more obvious that you are selecting headings where the supplement is the _default_. And it also means that if Typst decides to change what `supplement: auto` gets elaborated to in the future (to be fair, this seems quite unlikely), the selector will not break.

ericceg:
## Summary

Fixes #7825. Previously, matching in `verdict` was done only using the pre-synthesized clone, where `auto` may already be rewritten to a concrete value (e.g. for `heading.supplement`). That made selectors like `heading.where(supplement: auto)` fail.

Selector matching now checks both:
- the pre-synthesized clone and
- the original element (when they differ).

This keeps synthesized-field matching behavior while allowing selectors to match raw constructor values like `auto`.

## Test

Added test `issue-7825-show-where-auto-on-synthesized-field` in `show-where.typ` that verifies `heading.where(supplement: auto)` matches both explicit and implicit auto headings.

### DIFF
--- a/crates/typst-realize/src/lib.rs
+++ b/crates/typst-realize/src/lib.rs
@@ -448,17 +448,14 @@ fn verdict<'a>(
 
     for (r, recipe) in styles.recipes().enumerate() {
         // We're not interested in recipes that don't match.
-        if !recipe
-            .selector()
-            .is_some_and(|selector| {
-                // Match against both the pre-synthesized clone and the original
-                // element so selectors can still target raw constructor values
-                // like `auto` even if synthesis normalizes them.
-                selector.matches(elem, Some(styles))
-                    || (!std::ptr::eq(elem, original)
-                        && selector.matches(original, Some(styles)))
-            })
-        {
+        if !recipe.selector().is_some_and(|selector| {
+            // Match against both the pre-synthesized clone and the original
+            // element so selectors can still target raw constructor values
+            // like `auto` even if synthesis normalizes them.
+            selector.matches(elem, Some(styles))
+                || (!std::ptr::eq(elem, original)
+                    && selector.matches(original, Some(styles)))
+        }) {
             continue;
         }
 


laurmaedje:
I'm not comfortable with this fix. It's not clearly motivated why this is the appropriate way to deal with the issue.

The description (and potentially implementation) also feels like AI, which is against our contribution guidelines.

ericceg:
Thanks for the review. This being my first shot at a contribution I thought it might be smart to keep changes at one place in the codebase, but I see your concerns. The proper way to fix this is probably to somehow change the pre-synthesis step to remember whether a field was synthesized from `auto` to something more concrete (e.g. `[Section]` for `heading.supplement`). I guess I'll leave this to someone more experienced.

laurmaedje:
Yeah, it's definitely not the easiest place to start. Sorry if I came across as a little harsh.

Issue #7844, opened 2026-02-10
Incorrect conic gradient on text in SVG


saecki:
### Description

The `gradient-conic-text` test doesn't render properly in SVG:

<img width="1077" height="384" alt="Image" src="https://github.com/user-attachments/assets/dd10826b-6792-467e-b18a-8b60a66e25ee" />

### Typst version

- [ ] I am using the latest version of Typst

PR #7886, opened 2026-02-18
Upgrade resvg and tiny-skia

wmedrano:
Closes at #7844

# Blockers
- [ ] Upstream dependency krilla needs to be updated as well. Current draft uses my patched version
- [ ] Figure out why tests/ref/render/image-decode-bad-svg.png started appearing


wmedrano:
This PR is almost ready.

- I want to get rough approval before sending out the krilla updates just in case this stalls indefinitely.
- Not sure why image-decode-bad-svg.png appeared. Any tips on debugging would be appreciated.
  - Perhaps the SVG is valid in the new resvg version and I just need to do a new "bad svg"?

wmedrano:
Updated the example bad svg to be a different bad svg

laurmaedje:
In principle I'm happy to update resvg, but updating krilla was recently stalled in https://github.com/LaurenzV/krilla/pull/328 because hayro also needs some updates. I'm not sure whether updating hayro in Typst is currently feasible or whether it's still in an in-flux state. 

cc @LaurenzV

LaurenzV:
Should be feasible I think, can make a new release soon-ish.

saecki:
I think `resvg` and `tiny-skia` are also bumped in #8294, though it doesn't bump the `png` dependency, which results in both versions being present.
Since those changes are also the main changes in this PR, maybe we could rebase this PR once #8294 is merged and rename it to `Bump png` instead?

laurmaedje:
I would rather have #8294 bump `png` as well. Seems more like an oversight. cc: @reknih

saecki:
I don't think this fixes the linked issue. The issue is about the SVG export, not how SVGs are rendered. The PNG output of that test was fine to begin with.

laurmaedje:
Let's close this in favor of #8294 then.

Thank you for the PR nonetheless!

PR #7892, opened 2026-02-18
Add HTML iframe preview in test reports

saecki:
This adds a visual HTML preview inside an `<iframe>` for the test report:

image

### DIFF
--- a/tests/src/report/html.rs
+++ b/tests/src/report/html.rs
@@ -61,6 +61,12 @@ impl Html {
         write!(self.buf, r#" {name}="{val}""#).ok();
     }
 
+    fn write_bool_attr(&mut self, name: &str) {
+        assert!(self.in_attribute_list);
+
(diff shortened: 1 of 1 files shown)


tests/src/report/html.rs
```diff
@@ -1210,6 +1209,30 @@ fn image_diff(
     });
 }
 
+fn html_diff(parent: &mut HtmlElem, diff: &FileDiff<diff::Html>) {
+    let iframe = |parent: &mut HtmlElem<'_>, data_url: &str| {
+        parent
+            .iframe()
+            .class("html-frame")
```
laurmaedje, line 1230: it would be good to add the [`sandbox` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox) (with no / empty value) to ensure the contained HTML is maximally restricted in what it is allowed to do.

PR #7893, opened 2026-02-18
feat: merge strokes

### LINKED ISSUE #7839: Add merging/sum capabilities for strokes
### Description

I'm currently working on a package which allows some styling through a config dictionary parameter. Local changes can be made by passing the config and overwriting default configurations.

I wish to merge stroke settings together via addition `a + b` (similar behavior to merging dictionaries). But this doesn't work.

Example: `stroke(black + 0.5pt) + stroke(paint: red, dash: "dashed")` results to an error funnily saying `cannot add stroke and stroke` (the desired result would be `stroke(paint: red, dash: "dashed", thickness: 0.5pt)`).

### Use Case

I wish to merge stroke settings together. For example I have `stroke(black + 0.5pt)` as the default settings. If the user wants to set it to `stroke(paint: red, dash: "dashed")`, it would desirably result in `stroke(paint: red, dash: "dashed", thickness: 0.5pt)`.

What are your thoughts on this?

statxc:
# Add stroke merging/addition capability

## Problem

Users could not merge `Stroke` objects using the `+` operator, which made it difficult to build configuration systems where default stroke settings can be overridden by user-specified properties.

**Example that failed before this PR:**
```typst
#let default = stroke(black + 0.5pt)
#let user = stroke(paint: red, dash: "dashed")
#let merged = default + user  // Error: cannot add stroke and stroke
```

The desired behavior is to merge the stroke properties, with the right-hand side overriding the left-hand side where specified, resulting in:
```typst
stroke(paint: red, thickness: 0.5pt, dash: "dashed")
```

## Use Case

This feature is particularly useful for package authors who want to provide configurable styling:

1. Define default stroke settings (e.g., `stroke(black + 0.5pt)`)
2. Allow users to override specific properties (e.g., `stroke(paint: red, dash: "dashed")`)
3. Merge them to get a stroke with both default and user-specified properties

## Solution

Added support for stroke addition in the `add` function (`crates/typst-library/src/foundations/ops.rs`) by leveraging the existing `Fold` trait implementation that `Stroke` already has.

**Fixes #7839**

### DIFF
--- a/crates/typst-library/src/foundations/ops.rs
+++ b/crates/typst-library/src/foundations/ops.rs
@@ -7,7 +7,7 @@ use typst_utils::Numeric;
 
 use crate::diag::{HintedStrResult, StrResult, bail};
 use crate::foundations::{
-    Datetime, IntoValue, Regex, Repr, SymbolElem, Value, format_str,
+    Datetime, Fold, IntoValue, Regex, Repr, SymbolElem, Value, format_str,
 };
 use crate::layout::{Alignment, Length, Rel};
 use crate::text::TextElem;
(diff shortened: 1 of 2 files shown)


statxc:
@laurmaedje Please review my PR. Thanks.

PR #7894, opened 2026-02-18
 Add highlight color to grouping parens in math 

isuffix:
~~This PR is part of the exhaustive math syntax refactor. It should be considered exploratory, and not necessarily a final decision, even if merged to the `math-syntax` branch. Non-specific feedback for this change should go in one of its related issues.~~

---

This PR adds syntax highlighting to parentheses that are used for grouping in math mode. This will help to distinguish delimiters that will render as text from ones that are used purely for grouping with operators.

(I'm including this PR in the math syntax refactor because it will be depended-on by other changes to grouping parentheses, so it's better to keep it on the `math-syntax` branch)

One possible point of contention is which color to use for grouping parens. I'm currently using the same color as other math operators and escaped characters, and while I can see this being confusing at times, I think it looks alright and works to show that the parentheses are grouping text *for* the corresponding operator.

**Old highlighting:**

**New highlighting:**

The first commit improves both the markup and math highlighting test cases with more examples, and separates identifier highlighting in math into a different test case. This does incur a mild increase in file sizes, but I think it's worth including these cases.

The second commit adds highlighting for grouping parentheses in math. I went with the scope name `punctuation.math.typst` because other parentheses and braces are given the `punctuation.typst` scope, although I'm really just choosing based on instinct and the surrounding code.


laurmaedje:
> I'm including this PR in the math syntax refactor because it will be depended-on by other changes to grouping parentheses, so it's better to keep it on the math-syntax branch

Since this is not at all breaking and quite useful, maybe we can land it on main and rebase math-syntax after that?

laurmaedje:
Thanks!

PR #7895, opened 2026-02-19
Improved list layout

### LINKED ISSUE #529: Unable to center math blocks in a numbered list
Not sure if this a bug or actually an expected behavior. It seems the width of a numbered list is not always 100% but instead restricted to the maximum width of its lines.

Therefore, if you don't have a long (>=100%) line in a numbered list, your math blocks in that list will not be placed at the horizontal center of the page. These math blocks are instead centered relative to the the parent numbered list. This is some what conterintuitive (at least from a previous LaTeX user's perspective).

With a long line:

Without a long line:

To reproduce:
```
= Question 1

+ #set enum(numbering: "(a)")
  + The constructed sequent is given as follows:
    $ q arrow p, r arrow not p, q and r tack.long ((p and q) arrow not r) or ((p and r) arrow not q) $
    //The constructed sequent has assumptions $X={A_1,A_2,A_3}$ and a conclusion $C$ where:
    Without the long line:
    $ A_1 &= q arrow p\
      A_2 &= r arrow not p\
      A_3 &= q and r\
      C space &= ((p and q) arrow not r) or ((p and r) arrow not q)
    $
```

### LINKED ISSUE #1204: Bullet point is not aligned with text at beginning of list item
### Description

For a bullet list item with extra line height, the bullet point is not aligned with the text as one would expect for "usual" items.

```typ
- abc
- def #box(stroke: 1pt, inset: 3mm, baseline: 3mm)[test]
- #lorem(30)
```
produces:
image

As the third bullet point shows, the issue is not the total height of the item but really its line height.

### Reproduction URL

_No response_

### Operating system

Web app

### Typst version

PgBiel:
Fixes #529, #1204

Depends on #8150.

Reworks list (and enum, **but not `terms`**) layout by 1. adding a form of baseline alignment between the list item's marker and its body, allowing proper alignment between e.g. the marker and a math equation, and 2. allowing the list to expand to the full surrounding width when it isn't infinite, which fixes center-aligned things inside it.

The example image below shows the core of the changes: a centered equation in a list with baseline alignment (the marker matches the baseline height of the math text).

image

## Behavior changes

1. **Baseline alignment:** The list marker for each item will now be moved downwards in attempt to match the baseline of the body. For example, if the text starts a bit below, the marker will move a bit below. However, if it starts a bit above, the text will be moved down instead of the marker.

  ~~It is worth noting that baseline alignment will consider the very first (visible) object inside the body. This would correspond to an equation's first row, or to a rectangle around text, etc., so the results won't always be what one might expect. Still, we can always improve the heuristics for specific use cases.~~

2. **Centering and expansion:** Lists now always occupy the full page width (unless it has `width: auto`), ceasing the need for hacks like `#h(1fr)` or `show list: block.with(width: 100%)` etc. from #529. This means that anything that is center-aligned inside the list body, such as equations, will still be center-aligned. (But also right-aligned, and so on.) Of course, the marker itself isn't aligned.

  There is an important consequence of this. `#align(right)[- my list]` will **no longer** align the whole list to the right - the marker remains to the left (`start`), because the list always occupies the full width, so alignment doesn't affect the marker. To move the whole list to the right, you need to **restore the old list behavior** of not expanding to the full width. To do this, one can just surround it with a `block`: `#align(right, block[- my list])`.

## Implementation details

- **Lists are no longer laid out using grids.** Instead, they use a `stack` combined with a custom layouter (for marker horizontal alignment and baseline alignment). `stack`, differently from a `grid` with auto columns, propagates to its children whether they can `expand` to fill the page width or not. This automatically fixes centering in lists (#529).
- The custom layouter measures all list markers to assign the same (max) width to all of them, allowing horizontal centering of enum numbers to `start` or `end`.
- For baseline alignment of frames, we calculate the baseline of both (how far down from the top the text / math baseline is) and come to a difference. If this difference is positive, we move the marker down. If it is negative, we move the body down instead.
- To calculate the baseline of each frame, we first check if the frame has a natural baseline already and align with it (this should match the baseline of the first text/math content in the list item, per #8150). Otherwise (if there is no text or the content is "too weird"), the marker just goes to the top.

## TODOs

- [x] Reimplement layout with `stack`, fixing centering
- [x] Horizontal marker alignment
- [x] Initial baseline alignment impl
- [x] Initial tests
- [x] Support vertical alignment through `enum.number-align`
- [x] Some edge-case tests


mewmew:
Great to see work on this @PgBiel!

As a small reminder, the following [Typst roadmap](https://typst.app/docs/roadmap/) bullet point can be crossed out once this is merged.
> * Fix issues with list (in particular baselines & alignment)

crates/typst-layout/src/lists.rs
```diff
@@ -195,10 +197,33 @@ fn layout_items(
...
+        // Infinite space or `width: auto` used. Both would prevent the list
+        // from expanding to fit, breaking alignment. Therefore, restrict the
+        // list size to the size of the largest item, prompting list items to
+        // align between themselves instead of relative to the full page width.
+        let mut measured_body_size = Abs::zero();
+        for item in &items {
+            let body = crate::layout_frame(
+                engine,
+                &item.body,
+                locator.next(&item.body.span()),
```
laurmaedje: This is a problem. The measurement needs to use the exact same locator as the proper layout or else we'll have issues with introspection.

PgBiel: Bummer, I assume `locator.next(&())` would still produce a different locator, and the stack would still produce yet another locator, right?...

Maybe we should store the locator inside each item?

laurmaedje: Yeah, the way to go is to produce a locator upfront and store it, like in grid layout.

crates/typst-layout/src/lists.rs
```diff
@@ -138,27 +144,449 @@ pub fn layout_enum(
...
+                width,
+                (marker.size().y + self.marker_offset.y)
+                    .max(body_frame.height() + self.body_offset.y),
+            ));
+
+            let mut body_pos = self.body_offset;
+            if self.list.is_rtl {
+                // In RTL cells expand to the left, thus the position must
+                // additionally be offset by the cell's width.
+                body_pos.x = width - (body_pos.x + body_frame.width());
```
laurmaedje: Given,
```
body_pos.x + body_frame.width()
= self.body_offset.x + body_frame.width()
= width
```
doesn't this always compute to `Abs::zero()`?

crates/typst-layout/src/lists.rs
```diff
@@ -138,27 +144,449 @@ pub fn layout_enum(
...
+            }
+
+            // Only place the marker on the first non-empty frame.
+            if i == self.first_frame {
+                let mut marker_pos = self.marker_offset;
+                if self.list.is_rtl {
+                    marker_pos.x = width - (marker_pos.x + self.list.marker_width);
```
laurmaedje:
```suggestion
                    marker_pos.x = width - (self.list.marker_width + marker_pos.x);
```
I think this would be a bit clearer since the marker offset (via `marker_pos`) is on the right for RTL

crates/typst-layout/src/lists.rs
```diff
@@ -138,27 +144,449 @@ pub fn layout_enum(
...
+        // Collect the item's frames. Here, we add the marker to the first
+        // non-empty frame, and additionally indent the whole body so it appears
+        // after the marker.
+        let mut frames = vec![];
+        for (i, body_frame) in body_fragment.into_iter().enumerate() {
+            let width = self.body_offset.x + body_frame.width();
+            let mut frame = Frame::soft(Size::new(
+                width,
+                (marker.size().y + self.marker_offset.y)
+                    .max(body_frame.height() + self.body_offset.y),
```
laurmaedje: would be nice to extract a `height` variable here

crates/typst-layout/src/lists.rs
```diff
@@ -138,27 +144,449 @@ pub fn layout_enum(
...
+            engine,
+            &self.item.body,
+            self.body_locator.relayout(),
+            self.styles,
+            regions,
+        )?;
+
+        // First non-empty frame (ignores frames with only tags due to a forced
+        // region break).
+        self.first_frame = if should_skip_first_frame(&fragment) { 1 } else { 0 };
```
laurmaedje:
```suggestion
        if should_skip_first_frame(&fragment) {
            self.first_frame = 1;
        }
```
would be a bit simpler in my opinion and should be the same (?)

crates/typst-layout/src/lists.rs
```diff
@@ -138,27 +144,449 @@ pub fn layout_enum(
...
+    styles: StyleChain<'a>,
+    /// The first body's non-empty frame.
+    first_frame: usize,
+    /// The item's regions, but adapted for the body's expected width.
+    body_regions: Regions<'a>,
+    /// Total body indent from the left of the region, as well as its offset from the top of the region.
+    ///
+    /// Vertical offset is always zero without baseline alignment.
+    body_offset: Point,
+    /// Resolved marker indent from the left of the region, as well as its offset from the top of the region.
```
laurmaedje: could you wrap these comments at column 80?

crates/typst-layout/src/lists.rs
```diff
@@ -138,27 +144,449 @@ pub fn layout_enum(
...
+    }
+}
+
+/// Layout list items.
+///
+/// This is done in 4 steps:
+///
+/// 1. Compute marker widths for horizontal marker alignment;
+/// 2. If necessary (when within a `width: auto` block or page), compute body
+/// widths as well for horizontal body alignment to work;
```
laurmaedje:
```suggestion
///    widths as well for horizontal body alignment to work;
```

crates/typst-library/src/model/enum.rs
```diff
@@ -166,15 +166,24 @@ pub struct EnumElem {
...
     ///
-    /// By default, this is set to `{end + top}`, which aligns enum numbers
-    /// towards end of the current text direction (in left-to-right script, for
-    /// example, this is the same as `{right}`) and at the top of the line. The
-    /// choice of `{end}` for horizontal alignment of enum numbers is usually
-    /// preferred over `{start}`, as numbers then grow away from the text
-    /// instead of towards it, avoiding certain visual issues. This option lets
-    /// you override this behaviour, however. (Also to note is that the
+    /// By default, this is set to `{end}`, which aligns enum numbers
+    /// horizontally towards end of the current text direction (in left-to-right
```
laurmaedje:
```suggestion
    /// horizontally towards the end of the current text direction (in a left-to-right
```
(+ rewrap)

crates/typst-library/src/model/enum.rs
```diff
@@ -166,15 +166,24 @@ pub struct EnumElem {
...
+    /// script, for example, this is the same as `{right}`). In addition, the
+    /// lack of a vertical alignment places each number vertically just above
+    /// the baseline of the item, as if it were part of its first line of text.
+    ///
+    /// The choice of `{end}` for horizontal alignment of enum numbers is
+    /// usually preferred over `{start}`, as numbers then grow away from the
+    /// text instead of towards it, avoiding certain visual issues. This option
```
laurmaedje: I'd drop ", avoiding certain visual issues"

crates/typst-library/src/model/enum.rs
```diff
@@ -166,15 +166,24 @@ pub struct EnumElem {
...
+    /// lets you override this behaviour, however. (Also to note is that the
     /// @list[unordered list] uses a different method for this, by giving the
-    /// `marker` content an alignment directly.)
+    /// `marker` content an alignment directly, albeit it possesses a similar
```
laurmaedje: I'm not sure I understand what this sentence is trying to tell me with "`marker` content ... directly". Isn't `marker-align` the exact analogous thing?

laurmaedje (commented):
I have a few minor nitpicks on the code, but more importantly, I found a bug that is not present on main:

When wrapping a list in an auto-sized container like a `block`, it seems that the body region is not computed correctly. List or body indent or a very wide marker then lead to overflowing content while negative indent leads to a gap at the right end.

Here's an example:

```typ
#set page("a5")
#set list(indent: 100pt)
// here are more things to try
// #set list(indent: -50pt)
// #set list(marker: [AAAAAAAAAAAAAAAAAAAAA])
#block[
  - #lorem(20)
]
```

I think the measurement code path does not take into account the indents and markers at all.

---

I had already merged with main in preparation of merge, but feel free to force push that away if you prefer to rebase.

crates/typst-layout/src/lists.rs
PgBiel: Yeah, actually! The math itself is consistent with the grid implementation - but this is indeed always zero, as the body expands to the right, which becomes the left of the frame in RTL. Nice catch!

PgBiel: My goodness, why does my own reply look AI generated to me? Am I going insane? :joy:

Guess I will have to abandon my dear em dashes... :pensive:

crates/typst-layout/src/lists.rs
PgBiel: Welp, completely missed those, sorry haha

crates/typst-library/src/model/enum.rs
PgBiel: Yeah that's bogus, dropping it.

(Although it's much less relevant for lists, but still very much the same thing.)

PgBiel:
> When wrapping a list in an auto-sized container like a block, it seems that the body region is not computed correctly. 

Nice catch. I missed the need for auto column width redistribution when mimicking grid column measurement :sweat_smile:

crates/typst-layout/src/lists.rs
xkevio: > My goodness, why does my own reply look AI generated to me? Am I going insane? :joy:
> 
> Guess I will have to abandon my dear em dashes... :pensive: 

Actually, that was an en-dash 🤓

PgBiel:
I've added the same algorithm as grids now: if the sum of marker width + body width surpasses the available width (total - indents, i.e. total - fixed-width columns), then each is given (up to) half of the available width as needed.

I've also refactored the `ListLayouter`, splitting measurement into methods as well.

Please take a look at the latest commit to confirm. :)

PR #7897, opened 2026-02-19
Replace `Default` impls for `AstNode` with new `placeholder` method

isuffix:
This adds a new `placeholder` method to the `AstNode` trait that formalizes the previously implicit guarantee that AST nodes would not panic if constructed incorrectly.

### DIFF
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -192,7 +192,7 @@ pub fn eval_closure(
             (closure.name(), closure.params(), closure.body())
         }
         ClosureNode::Context(ref node) => {
-            (None, ast::Params::default(), node.cast().unwrap())
+            (None, ast::Params::placeholder(), node.cast().unwrap())
         }
     };
 
--- a/crates/typst-syntax/src/ast.rs
+++ b/crates/typst-syntax/src/ast.rs
@@ -77,12 +77,20 @@ using the lazy interface would only need to traverse each node once, improving
 throughput at the cost of initial latency and development flexibility.
 */
 
+// The AST should never panic when traversing a CST, even if the CST is in an
+// invalid structure, e.g. if we have a syntax error or make an incorrect edit
+// from the `typst-ide` crate. We disallow common panics in this file using
+// these lints, and we provide an alternative to panicking with the
+// `AstNode::placeholder()` method.
+#![deny(clippy::unwrap_used, clippy::expect_used, clippy::panic, clippy::unreachable)]
+
 use std::num::NonZeroUsize;
 use std::ops::Deref;
 use std::path::Path;
 use std::str::FromStr;
 
 use ecow::EcoString;
+use typst_utils::NonZeroExt;
 use unscanny::Scanner;
 
 use crate::package::PackageSpec;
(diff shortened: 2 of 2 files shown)


crates/typst-eval/src/call.rs
```diff
@@ -192,7 +192,7 @@ pub fn eval_closure(
             (closure.name(), closure.params(), closure.body())
         }
         ClosureNode::Context(ref node) => {
-            (None, ast::Params::default(), node.cast().unwrap())
+            (None, ast::Params::placeholder(), node.cast().unwrap())
         }
     };
```
isuffix, line 197: This was the only external user that required `Default` as far as I can tell. I tried to refactor this to remove the call to `placeholder`, but I couldn't do so without some serious refactorings to `ClosureNode` that I don't think are worth it.

crates/typst-syntax/src/ast.rs
```diff
@@ -77,12 +77,19 @@ using the lazy interface would only need to traverse each node once, improving
 throughput at the cost of initial latency and development flexibility.
 */
 
+// The AST should never panic when parsing a CST, even if the CST is in an
+// invalid structure, e.g. if incorrectly edited from the `typst-ide` crate. We
+// disallow common panics in this file using these lints, and we provide an
+// alternative to panicking with the `AstNode::placeholder()` method.
+#![deny(clippy::unwrap_used, clippy::expect_used, clippy::panic, clippy::unreachable)]
```
isuffix: Just added this. Should help with some of the mental overhead of this requirement.

Thanks to this, I also discovered `NonZeroUsize::ONE`. Neat.

crates/typst-syntax/src/ast.rs
```diff
@@ -100,6 +107,13 @@ pub trait AstNode<'a>: Sized {
     fn span(self) -> Span {
         self.to_untyped().span()
     }
+
+    /// A placeholder for this node type. If the underlying CST did not come
+    /// from the parser, the AST is not guaranteed to be valid. But instead of
+    /// panicking in that case, we return these placeholder nodes.
```
laurmaedje: This does not seem entirely accurate. Even if the CST comes from the parser, it can be invalid if it contains error nodes. (This is precisely what happens in practice in typst-ide.)

crates/typst-syntax/src/ast.rs
```diff
@@ -125,13 +139,13 @@ impl SyntaxNode {
     }
 
     /// Get the first child of AST type `T` or a placeholder if none.
-    fn cast_first<'a, T: AstNode<'a> + Default>(&'a self) -> T {
-        self.try_cast_first().unwrap_or_default()
+    fn cast_first<'a, T: AstNode<'a>>(&'a self) -> T {
+        self.try_cast_first().unwrap_or(T::placeholder())
```
laurmaedje: Maybe we should use unwrap_or_else instead in these places. Should be a little easier to optimize for the compiler.

isuffix:
I've rebased off main, updated the comments, and used `unwrap_or_else` wherever we construct a placeholder.
This is ready for review.

laurmaedje:
Thanks!

PR #7899, opened 2026-02-20
Add support for namespaces when decoding XML

### LINKED ISSUE #7898: Include namespace URI in XML parsing output
### Description

At the moment, the XML parser takes namespace information into account when processing prefixed elements (such as `<foo:bar/>`). However, it seems like this information gets lost in the output and cannot be used when iterating over the document structure. In the `<foo:bar/>` case, we end up with a Typst array element that indicates that the tag name was `bar`, but the namespace information is lost.

To enable use cases that work on XML-encoded data structures, it would probably be nice if the namespace URI would be included if present in the respective tag or attribute.

### Use Case

One particular use case I have in mind is the processing of DocBook documents. However, adding this feature should probably be useful for other XML-based inputs as well.

JKRhb:
Potentially resolves #7898.

The [`hello.xml`](https://github.com/typst/typst-dev-assets/blob/main/files/data/hello.xml) file in the [`typst-dev-assets`](https://github.com/typst/typst-dev-assets) unfortunately does not contain any namespaces yet, which is why the existing test does not really indicate whether the changes work as intended. However, I assume we should be able to extend `hello.xml` with namespace information without breaking the previous implementation.

### DIFF
--- a/tests/suite/loading/xml.typ
+++ b/tests/suite/loading/xml.typ
@@ -2,27 +2,27 @@
 // Test reading XML data.
 #let data = xml("/assets/data/hello.xml")
 #test(data, ((
+  namespace: none,
   tag: "data",
   attrs: (:),
   children: (
     "\n  ",
-    (tag: "hello", attrs: (name: "hi"), children: ("1",), namespace: none),
+    (namespace: none, tag: "hello", attrs: (name: "hi"), children: ("1",)),
     "\n  ",
     (
+      namespace: none,
       tag: "data",
       attrs: (:),
       children: (
         "\n    ",
-        (tag: "hello", attrs: (:), children: ("World",), namespace: none),
+        (namespace: none, tag: "hello", attrs: (:), children: ("World",)),
         "\n    ",
-        (tag: "hello", attrs: (:), children: ("World",), namespace: none),
+        (namespace: none, tag: "hello", attrs: (:), children: ("World",)),
         "\n  ",
       ),
-      namespace: none,
     ),
     "\n",
   ),
-  namespace: none,
 ),))
 
 // Test reading through path type.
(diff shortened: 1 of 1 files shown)


JKRhb:
> However, I assume we should be able to extend `hello.xml` with namespace information without breaking the previous implementation.

Testing the implementation locally with the following adjusted XML document actually works:

```xml
<data xmlns="http://example.org" xmlns:foo="urn:foo">
  <hello name="hi">1</hello>
  <data>
    <foo:hello>World</foo:hello>
    <foo:hello>World</foo:hello>
  </data>
</data>
```

The `namespace` fields are filled with `http://example.org` or `urn:foo` depending on whether the tags contain a namespace prefix or not.

JKRhb:
I just opened https://github.com/typst/typst-dev-assets/pull/21 to update `hello.xml` as indicated above.

SeSodesa:
If you wrote a pull request, you might just as well add a unit test that checks that the simple document you provided works.

JKRhb:
> If you wrote a pull request, you might just as well add a unit test that checks that the simple document you provided works.

Thank you for the suggestion :) I now added a separate test case via https://github.com/typst/typst/pull/7899/commits/49711ea957f957d1062ebda3193e83ceb4fe8389, annotated with a TODO that the XML document could be moved to the typst-dev-assets.

laurmaedje:
Thanks!

I've decided to keep the inline test instead of adjusting the file in typst-dev-assets. The files there largely predate the capability to even load inline data, so being able to test at all is their main purpose (though testing that loading from a file works is another motivation). But for a more specific test like namespaces, having it inline is simpler to maintain. I'd also like to keep a test without namespaces to test for the `none` case, so it's actually good to keep `hello.xml` as is.

PR #7907, opened 2026-02-24
Link "See also" page of symbol shorthands to the docs page showing them

omentic:
Tiny little fix. The docs actually link to the wrong page here: shorthands are documented on the main symbols page.


YDX-2147483647:
For your information, shorthands are also shown on the page of [General Symbols (`/symbols/sym/`)](https://typst.app/docs/reference/symbols/sym/).
图片

(It's not obvious that the cards are clickable…)

laurmaedje:
Thanks!

PR #7909, opened 2026-02-25
Fix docs of `array.chunks`

knuesel:
I think "discard" is more correct since whether=yes corresponds to exact=true.


laurmaedje:
Thanks!

PR #7916, opened 2026-02-27
Add link to web app in tutorial

### LINKED ISSUE #7587: Add a link to the web app on the tutorial
### Description

Say a newcomer want’s to learn to use Typst and lands on the [tutorial](https://typst.app/docs/tutorial/). They are greeted by a link to the [CLI tool](https://github.com/typst/typst), which is not suitable for many beginners, but no link to the web app, which is the appropriate place for most people to start. I suggest adding a link that targets a new window/tab (so the tutorial stays open).

Where to link exactly?

- One option is a blank canvas, like [Typst Playground](https://typst.app/play/) but without any preexisting code.  
  ✅ Minimal friction. No need to sign in or alter existing text.  
  ❌ The user does not create an account, so they are limited in what they can do.
- Another is the sign-up or the sign-in page.  
  ✅ With an account, the user can produce documents (including experiments done when going through the tutorial).  
  ❌ Signing up is a hurdle many users would not be willing to pass when they don’t know if Typst is for them, and upon seeing a sign-up page they might just go away.

I think the first option is much better: less hurdles, more retention. IMHO being inclusive involves not ‘scaring off’ newbies with commitments like signing up before they are convinced it’s worth it.

The first part, [Writing in Typst](https://typst.app/docs/tutorial/writing-in-typst/) also lacks a link to the web app when it mentions it at the beginning.

3w36zj6:
close #7587

This PR adds a link to the web app at the top of the tutorial documentation.

The issue suggested linking to either the Playground or the sign up page. I agree with choosing the Playground, as it allows users to start writing immediately and smoothly create an account later if they want to.

### DIFF
--- a/docs/tutorial/1-writing.md
+++ b/docs/tutorial/1-writing.md
@@ -5,9 +5,9 @@ description: Typst's tutorial.
 # Writing in Typst
 Let's get started! Suppose you got assigned to write a technical report for
 university. It will contain prose, maths, headings, and figures. To get started,
-you create a new project on the Typst app. You'll be taken to the editor where
-you see two panels: A source panel where you compose your document and a
-preview panel where you see the rendered document.
+[open Typst Playground](https://typst.app/play/) and create a new project.
+You'll be taken to the editor where you see two panels: A source panel where you
+compose your document and a preview panel where you see the rendered document.
 
 ![Typst app screenshot](1-writing-app.png)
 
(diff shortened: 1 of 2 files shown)


laurmaedje:
The playground isn't really suitable for the current tutorial because you cannot add additional files. We're also in the process of rewriting the whole tutorial, so don't want to invest anymore into the old one. Thanks anyway!

PR #7919, opened 2026-02-28
Expose local timezone offset

### LINKED ISSUE #7112: Expose local timezone offset
### Description

It would be useful to have a method `datetime.localoffset() -> int`, which returns the local offset to UTC. `datetime.today(offset: auto)` already uses this offset internally.

`datetime.today(offset: datetime.localoffset())` should be equivalent to `datetime.today(offset: auto)`, whatever type may be used in the future. See #6277.

This seems like a fairly small change that just exposes already existing internals. Ideally, there would be an API to access the whole timezone data base, but that might be too much.

### Use Case

This would allow parsing dates given in UTC and displaying them in local time in a document, in a similar vein to #4107.

In my particular use case, I want to convert markdown files with a yaml metadata header to pdf using typst. The header contains date/time information given in UTC, which I want to display in the document in local time. Both pyrunner and jogs have issues with date/time access, so the only way to do this currently seems to be using `--input` arguments.

3w36zj6:
close #7112

This pull request introduces `datetime.local-offset` to retrieve the local timezone offset. 

Since the underlying logic for this feature already existed internally for `datetime.today`, the implementation has been refactored to share the same processing.

```typ
#set page(height: auto)

#let metadata = toml(bytes("createdAt = 2026-03-01T00:00:00Z"))

/ UTC: #metadata.createdAt.display()
/ Local: #(metadata.createdAt + datetime.local-offset()).display()
```

image

As a side note, while working on this, I noticed that `toml()` doesn't resolve timezones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2993,6 +2993,7 @@ dependencies = [
  "sigpipe",
  "tar",
  "tempfile",
+ "time",
  "toml",
  "typst",
  "typst-dev-assets",
--- a/crates/typst-cli/Cargo.toml
+++ b/crates/typst-cli/Cargo.toml
@@ -28,6 +28,7 @@ typst-svg = { workspace = true }
 typst-timing = { workspace = true }
 typst-utils = { workspace = true }
 chrono = { workspace = true }
+time = { workspace = true }
 clap = { workspace = true, features = ["string"] }
 clap_complete = { workspace = true }
 codespan-reporting = { workspace = true }
--- a/crates/typst-cli/src/world.rs
+++ b/crates/typst-cli/src/world.rs
@@ -5,6 +5,7 @@ use std::sync::{LazyLock, OnceLock};
 
 use chrono::{DateTime, Datelike, FixedOffset, Local, Utc};
 use ecow::{EcoString, eco_format};
+use time::Duration as TimeDuration;
 use typst::diag::{FileError, FileResult};
 use typst::foundations::{Bytes, Datetime, Dict, Duration, IntoValue, Repr};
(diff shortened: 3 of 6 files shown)


t-rapp:
The example above assumes that the local timezone offset which is in effect when the document is compiled, is the same as the local timezone offset for the given date (2026-03-01). This will not be valid in all cases for locations with Daylight Saving Time (DST).

I'm a bit skeptic that exposing the plain _current_ timezone offset is a good idea. A timezone offset should be relative to a specific date. Not sure how far Typst wants to go with timezone support, maybe some inspiration can be taken from PostgreSQL [datetime types](https://www.postgresql.org/docs/current/datatype-datetime.html) and [operations](https://www.postgresql.org/docs/current/functions-datetime.html)?

3w36zj6:
Oops, I had overlooked the DST consideration. While this is indeed problematic if treated as a feature for reliably converting arbitrary datetimes, the responsibility of this PR is merely to expose logic that already exists internally, and it is useful enough for many use cases, so I believe it is fine as a short-term solution. That said, the DST caveat should be explicitly documented.

Ideally, I would also hope to see timezone-aware operations on the `datetime` type implemented in the future.

laurmaedje:
I'm also not yet sure how far we want to go with timezone support, but I think as-is, this approach is a bit too simplistic. It indeed feels like a stop-gap short-term solution, which we are generally not big fans of shipping. I'm also not too happy with the proliferation of methods on the `World` trait.

While I appreciate that there are some genuine use cases, we'd thus rather not move forward with this. Thank you still for your work on it!

PR #7923, opened 2026-02-28
Improve error when function receives too many compiler-generated args

### LINKED ISSUE #2102: Make error more descriptive when non-sink argument is used instead of expected sink argument
The issue #2101 was created due to the next error message:

```
error: unexpected argument
   ┌─ /file.typ:42:21
   │
42 │   heading(numbering: it => it, name, ..args)
   │                      ^^
```

It shows that the argument is unexpected, but in reality it is of course expected, and instead of a normal argument, it should be a [sink argument](https://typst.app/docs/reference/types/arguments/):

```typ
42 │   heading(numbering: (..it) => it, name, ..args)
```

The error did throw me off, but after an [answer](https://github.com/typst/typst/issues/2101#issuecomment-1712825858) and some "digging" through the [documentation](https://github.com/typst/typst/issues/2101#issuecomment-1712842441), the error was resolved. In any case, the error message should be more descriptive and user-friendly.

vasilytrofimchuk:
Closes #2102

When `fold`, `reduce`, or similar higher-order functions call a closure internally, they pass arguments via `Args::new(func_span, values)`. This means every compiler-generated argument gets the same span as the function itself. If the closure doesn't accept enough parameters, `finish()` reports "unexpected argument" pointing at the function definition — which is confusing, since the user never wrote an extra argument.

The fix detects this case by checking `arg.span == self.span` in `finish()`. When the spans match, it means the argument was compiler-generated (created by `Args::new` with the function's span as fallback), so we emit a more descriptive error:

> function received more arguments than it expected
> hint: consider adding an argument sink to capture them

For user-written arguments (where `arg.span` points to the actual source location), behavior is unchanged.

Updated the existing `fold`/`reduce` tests and added a new test case for closures with too few parameters.

### DIFF
--- a/crates/typst-library/src/foundations/args.rs
+++ b/crates/typst-library/src/foundations/args.rs
@@ -251,6 +251,18 @@ impl Args {
     /// argument.
     pub fn finish(self) -> SourceResult<()> {
         if let Some(arg) = self.items.first() {
+            // When the argument's span matches the args span, it means the
+            // argument was compiler-generated and has no real source location
+            // (its span defaulted to the function's span). In this case,
+            // produce a more descriptive error message instead of the
+            // confusing "unexpected argument" pointing at the function.
+            if arg.span == self.span {
+                bail!(
+                    arg.span,
+                    "function received more arguments than it expected";
+                    hint: "consider adding an argument sink to capture them"
+                )
+            }
             match &arg.name {
                 Some(name) => bail!(arg.span, "unexpected argument: {name}"),
                 _ => bail!(arg.span, "unexpected argument"),
--- a/tests/suite/foundations/array.typ
+++ b/tests/suite/foundations/array.typ
@@ -279,8 +279,14 @@
 #test((1, 2, 3, 4).fold(0, (s, x) => s + x), 10)
 
 --- array-fold-closure-without-params eval ---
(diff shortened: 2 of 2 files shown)


isuffix:
Just comparing spans equal is unfortunately fraught due to `eval` setting all spans to the outer string's span, so the solution causes these examples to produce two different errors:
```typ
// Error: 26-27 unexpected argument
#let id = x => x; #id(1, 2)

// Error: 7-33 function received more arguments than it expected
// Hint: 7-33 consider adding an argument sink to capture them
#eval("let id = x => x; id(1,2)")
```

However, since internal arguments are always `Span::detached()`, fixing this case should just require adding `self.span.is_detached() && ...` to the condition.

isuffix:
Also, since I can see that Claude co-wrote this commit, could you clarify what amount of AI you're using, i.e. did AI write the code, the PR description, the commit message, etc.?

laurmaedje:
The PR description also reads fairly AI-y to me (which is not permitted as per the contribution guidelines at time of PR creation), but since [we're fully banning AI-implemented contributions](https://github.com/typst/typst/pull/7932) and the implementation was clearly done by Claude, I'll go ahead and close this.

PR #7926, opened 2026-03-01
Accept `bytes` for `text.font`.

### LINKED ISSUE #2469: Load fonts from relative paths
### Description

It would be nice if fonts could be loaded from typ files for use in templates and icon packages.

Something like:
```typst
#loadfont(name: "Explicit Font Name", "./font.ttf")
```

Not sure if this covered by #100, but it's related.

### Use Case

I want to use a icon font, and instead of having it installed locally (which isn't ideal for packages), I'd like to include it with a relative path.

SLUCHABLUB:
I've added the ability to pass `bytes` objects to the `font` parameter of `text`. This can either be a single face in the form of a `.ttf`/`.otf` file or a falimy in the form of a `.ttc`/`otc` file.

This would close #2469.

## An Example

```typst
// Load a `bytes` object from a .ttf/.ttc/.otf/.otc file.
#let my_font = read("path/to/my/font.ttf", encoding: none)

#set text(font: my_font)
```

## Breakdown

- I broke the resolution of a `FontFamily` to a `Font` into a function (`FontFamily::font`).
- I added an optional field `faces` to `FontFamily`. This field would only be set if the `FontFamily` was parsed from bytes, otherwise, it would like before only be a reference of sort (via a name) to a collection of faces in the world.
- I added casting from `bytes` to a `FontFamily`  and from a `(bytes: ..., covers: ...)` dictionary analogous to with `name`.
- I added a test to test passing a read `.ttf` file to `text.font`.

## Open Questions

Whilst I've tested that it works for `.ttc`/`.otc` files locally, I was not able to write a test for it since there is currently no such files in the assets repository. Should I make a PR to the assets repo to add this?


laurmaedje:
Thanks for the PR! While this is a nice proof of concept, to really land font loading from paths in Typst, we need a much more comprehensive design. There need to be proper mechanisms in place configure various things about how families and styles are grouped etc (such that this works properly with things like strong emphasis even if not using a `.ttc` font collection). Most likely, this will involve adding a new `font` type to the language (See also: https://github.com/typst/typst/issues/2469#issuecomment-2705747813).

It's not out of the question for this to be in part a community contribution (though it's more likely to end up being done by the team because so much design is involved), but it's definitely too involved for a first contribution. I'll close this PR, but thank you nonetheless!

PR #7929, opened 2026-03-02
Fix wording in visualize accessibility note

gomazarashi:
In the accessibility section of `visualize.md`, I changed `model` to `module` in the sentence:

`If you are using the functions in this model`

In this context, the sentence appears to refer to where the functions belong, so `module` seemed more appropriate, and I thought `model` might be a typo.  
If `model` was intentional here, that is my misunderstanding, and I am happy for this PR to be closed.

This is a wording-only change and does not affect functionality or behavior.

### DIFF
--- a/docs/reference/library/visualize.md
+++ b/docs/reference/library/visualize.md
@@ -10,7 +10,7 @@ All shapes and paths drawn by Typst are automatically marked as
 [artifacts]($pdf.artifact) to make them invisible to Assistive Technology (AT)
 during PDF export. However, their contents (if any) remain accessible.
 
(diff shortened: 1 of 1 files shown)


laurmaedje:
This seems to indeed by a typo, but "module" is actually not factually correct either since it's not a module at the Typst language level. The correct term would be "category".

gomazarashi:
@laurmaedje 

Thank you for the correction. I’ve updated it to `category` and pushed a commit.

I’d appreciate it if you could take a look when you have time.

laurmaedje:
Thanks!

PR #7933, opened 2026-03-02
Underline rich test diff file path

MDLC01:
Just a small improvement to make it easier to locate the path.

### DIFF
--- a/tests/src/logger.rs
+++ b/tests/src/logger.rs
@@ -184,7 +184,13 @@ impl Logger {
             eprintln!(
                 "  pass '--update' to update error annotations or reference outputs"
             );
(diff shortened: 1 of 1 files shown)


istudyatuni:
Maybe it's better to use `color-print` dependency, since it's already used

https://github.com/typst/typst/blob/0eca2a07c15c3d42c71ceb0474a97e2abd574af0/Cargo.toml#L51

MDLC01:
I followed what was done in https://github.com/typst/typst/pull/6281. I'm not sure adding `color-print` as a dependency of `typst-test` is really necessary here.

However, I'll add an `is_terminal` check (as already done for failing test paths in #7581).

saecki (approved):

laurmaedje:
Thanks!

PR #7936, opened 2026-03-02
Get numeral systems from Codex

MDLC01:
As was previously discussed, this proposes to move numbering kinds to Codex. Numbering patterns stay in Typst, but numbering kinds and their "chars" are moved to Codex as _numeral systems_ with _shorthands_.

This is still a draft as the corresponding Codex PR (https://github.com/typst/codex/pull/145) should be merged first. Additionally, I made the decision to now raise an error when a numeral system cannot represent a number (e.g., zero) instead of arbitrarily falling back to Western Arabic numerals or a hyphen. Currently, the error is reported with a detached span, which is highly suboptimal.

For now, I do not make use of the names assigned to the various numeral systems in Codex and instead only use the shorthands. This means there should not be any observable behavior other than some numbers no longer being representable (hence the updated tests).


MDLC01:
The tests for the various numeral systems should probably be moved to Codex, and the test for deprecated symbol variants will always be annoying to maintain so I think it should be removed altogether.

laurmaedje:
> Currently, the error is reported with a detached span, which is highly suboptimal.

You'll probably have to add a `Span` argument to `Numbering::apply`.

laurmaedje:
> The tests for the various numeral systems should probably be moved to Codex

Let's keep some basic smoke tests.

> and the test for deprecated symbol variants will always be annoying to maintain so I think it should be removed altogether.

I can see how it is annoying, but having deprecation completely untested is also annoying. Maybe we could have a test-only perma-deprecated symbol? Though that would not test the codex integration if it's only on the Typst side...

MDLC01:
I bumped to the latest Codex commit and added span information. I'm not sure why the `enum.item` element in the added test doesn't get a span. Regarding the page numbering test, it is ambiguous what span should be reported. Maybe the place where the page numbering is declared?

I also want to emphasize that the new errors constitute a breaking change that will cause documents that currently compile to error. I believe this is for the better, as the current fallback to Arabic numerals is somewhat arbitrary. But given that there is no general error catching mechanism in Typst, nor a way to test whether a numeral system can represent a given number, this might be undesirable.

> Let's keep some basic smoke tests.

I have just kept the existing `numbering` test.

> I can see how it is annoying, but having deprecation completely untested is also annoying. Maybe we could have a test-only perma-deprecated symbol? Though that would not test the codex integration if it's only on the Typst side...

Let's keep it the way it is for now then. We can always make the change in a future PR if we want to.

crates/typst-layout/src/lists.rs
```diff
@@ -107,15 +108,19 @@ pub fn layout_enum(
...
             parents.pop();
             content
         } else {
             match numbering {
-                Numbering::Pattern(pattern) => {
-                    TextElem::packed(pattern.apply_kth(parents.len(), number))
-                }
-                other => other.apply(engine, context.track(), &[number])?.display(),
+                Numbering::Pattern(pattern) => TextElem::packed(
+                    pattern.apply_kth(parents.len(), number).at(Span::detached())?,
```
laurmaedje: > I'm not sure why the enum.item element in the added test doesn't get a span.

The item does have a span, but it's not used here.

crates/typst-library/src/model/numbering.rs
```diff
@@ -240,680 +262,16 @@ cast! {
...
     self => {
         let mut pat = EcoString::new();
-        for (prefix, kind) in &self.pieces {
+        for (prefix, system) in &self.pieces {
             pat.push_str(prefix);
-            pat.push(kind.to_char());
+            pat.push_str(
+                system
+                .shorthand()
+                .expect("it is not possible to construct numbering systems that don't have a shorthand within Typst for now"),
```
laurmaedje: the indent is a bit off here because there is no autoformat

crates/typst-pdf/src/convert.rs
```diff
@@ -98,7 +98,12 @@ fn convert_pages(gc: &mut GlobalContext, document: &mut Document) -> SourceResul
         if let Some(label) = typst_page
             .numbering
             .as_ref()
-            .and_then(|num| PageLabel::generate(num, typst_page.number))
+            .and_then(|num| {
+                PageLabel::generate(num, typst_page.number)
+                    .at(Span::detached())
```
laurmaedje: this error will be a bit hard to understand since it doesn't have any span or indication why it happened. perhaps we should at least add a hint that this is related to page numbering?

Let's also add a test for this. I think it should happen when you set a page numbering, but also a custom header/footer to avoid layout already failing.

tests/suite/model/outline.typ
```diff
@@ -184,7 +184,7 @@
 
 --- outline-heading-start-of-page paged ---
 #set page(width: 140pt, height: 200pt, margin: (bottom: 20pt), numbering: "1")
-#set heading(numbering: "(1/a)")
+#set heading(numbering: "(a/1)")
```
laurmaedje, line 187: why was this changed?

tests/suite/model/numbering.typ
```diff
@@ -94,3 +94,29 @@
...
+#numbering("①", 51)
+
+--- enum-numbering-too-high paged ---
+// Error: the number 11 is too large to be represented with the `arabic.oo` numeral system
+#set enum(numbering: "⓵")
+11. Test
+
+--- page-numbering-too-high paged ---
+// Error: the number 100 is too large to be represented with the `arabic.o` numeral system
+#set page(numbering: "①")
```
laurmaedje, line 118: > Maybe the place where the page numbering is declared?

I think reporting this at the page numbering definition would be fine

tests/suite/model/outline.typ
MDLC01: This was changed because there is a level-4 heading without a level-2 heading at some point. The full number for this heading is thus `(2, 3, 0, 1)`. Because the `latin` numeral system cannot display zero, this would produce an error.

laurmaedje:
Thank you, nice work!

MDLC01:
> I think reporting this at the page numbering definition would be fine

Regarding this, I did not find a way to get the span of the corresponding set rule. I don't have the time to think about it for now, so if anyone is interested to look into it that feel free to do so.

PR #7937, opened 2026-03-02
Retain test report artifacts longer

saecki:
The previous 3 `retention-days` where a bit to short of a time frame to review the PR.
This also avoids uploading the same exact test report 3 times, when the CI passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
             tests/store/pdftags
             tests/store/html
           compression-level: 9
-          retention-days: 30
+          retention-days: 7
 
       - name: Upload test report
         # When the CI passes, only upload one test report, since all should be the same.
@@ -85,7 +85,6 @@ jobs:
           name: tests-report-${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.bits }}
           path: tests/store/report.html
           compression-level: 9
-          retention-days: 30
 
   checks:
     name: Check clippy, formatting, and documentation


.github/workflows/ci.yml
```diff
@@ -74,15 +74,18 @@ jobs:
             tests/store/render
             tests/store/pdftags
             tests/store/html
-          retention-days: 3
```
laurmaedje, line 77: ah now I see where you got the 3 from. the idea here was that the store can be fairly large and it's mostly for immediate use by the contributor. for review, it's fairly useless.

I think we can leave this one at 3 or maybe 7 ...

.github/workflows/ci.yml
```diff
@@ -74,15 +74,18 @@ jobs:
...
-        if: failure() || (success() && github.event_name == 'pull_request')
+        # When the CI passes, only upload one test report, since all should be the same.
+        if: failure() || (success() && github.event_name == 'pull_request' && startsWith(matrix.os, 'ubuntu-') && matrix.bits == 64)
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: tests-report-${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.bits }}
           path: tests/store/report.html
-          retention-days: 3
+          compression-level: 9
+          retention-days: 30
```
laurmaedje: ... while removing this entirely to get the default of 90 days. sometimes, it may be even more than a month until a review happens. should we somehow get into issues with artifact storage we can still reduce it.

PR #7943, opened 2026-03-04
docs: list all valid elements for counter constructor (#2728)

### LINKED ISSUE #2728: List all elements that can be used in `counter` constructor in documentation
https://typst.app/docs/reference/meta/counter/:
> Counts through pages, elements, and more.
>
> With the counter function, you can access and modify counters for pages, headings, figures, and more. Moreover, you can define custom counters for other things you want to count.

What does "counters for pages, headings, figures, and more" mean? Are there other elements that can be used inside `counter()` or "more" refers to non-element things? I find it really frustrating every time I want to know if I can use `counter` for something built-in. IMO, documentation **must** state every single built-in element that can be used in the `counter` constructor. What about `enum`? Yes? No? I don't know! Documentation should answer this question, but it doesn't at the moment of writing.

ionodeionode:
Fixes #2728

### DIFF
--- a/crates/typst-library/src/introspection/counter.rs
+++ b/crates/typst-library/src/introspection/counter.rs
@@ -321,15 +321,14 @@ impl Counter {
     pub fn construct(
         /// The key that identifies this counter globally.
         ///
-        /// - If it is a string, creates a custom counter that is only affected
+    /// - If it is a string, creates a custom counter that is only affected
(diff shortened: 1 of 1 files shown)


laurmaedje:
The PR is kinda full of random changes and, for this reason, does not seem worth investing time into reviewing, so I'll close this.

PR #7944, opened 2026-03-04
Document how to get the ISO week year

### LINKED ISSUE #7927: Get the year field of the ISO week date
### Description

Hi,

I'm speaking of https://en.wikipedia.org/wiki/ISO_week_date.

The below script prints the date and relevant year and week of the last month of 2025.

```typst
#{
  let current = datetime(year: 2025, month: 12, day: 1)
  let n = 0
  while current.year() == 2025 {
    if (current.weekday() == 1) {
      current.display() + " " + current.display("[year]/[week_number padding:zero]") + linebreak()
    }
    current = current + duration(days: 1)
  }
}
```

That renders as:

Image

See 2025-12-29, in week based year, it's the first week of 2026. I don't see any way to get the year that corresponds to that week using the formatter only. It would require me to compute it myself (for example, if month is 12 and week is 1, I should replace year with the next year).

Likewise, I believe it would be wise to either add a "week_year" component or a "repr:week" modifier to the year.

For example, in Java, I would use:

```java
final var year = d.get(WeekFields.ISO.weekBasedYear());
final var week = d.get(WeekFields.ISO.weekOfWeekBasedYear());
```

I have zero experience with Rust, but if you believe this feature is a good idea, and you want me to do it, I'd be glad to give it a try.

Best regards,

Adrien H.

### Use Case

I came across this feature while I was trying to generate a PDF of a week-based timesheet.

AdrienHorgnies:
I believe it is useful as https://typst.app/docs/reference/foundations/datetime/#format specifies how to get the ISO week number of an ISO week date but the year documentation doesn't help on how to get the ISO week year counterpart.

My personal use case is that I format my timesheets to PDF before sending them to my freelance contract broker and I need/want to put the ISO week year along the ISO week in the title: `= Timesheet #start.display("[year base:iso_week]/[week_number padding:none]")`. And without the feature being documented, I was left in the dark before browsing the time rust crate.

closes #7927

### DIFF
--- a/crates/typst-library/src/foundations/datetime.rs
+++ b/crates/typst-library/src/foundations/datetime.rs
@@ -130,8 +130,8 @@ use crate::foundations::{
 ///   - `padding`: Can be either `zero`, `space` or `none`. Specifies how the
 ///     second is padded.
 ///
-/// See https://time-rs.github.io/book/api/format-description.html for the
-/// complete list of supported syntax.
+/// [See here](https://time-rs.github.io/book/api/format-description.html#components)
+/// for more details on the supported syntax.
 ///
 /// Keep in mind that not always all components can be used. For example, if you
 /// create a new datetime with `{datetime(year: 2023, month: 10, day: 13)}`, it


eltos:
Would it be useful to also mention and link to the original format description docs of the `time` crate? This way people would be able to access the source documentation in case more information is desired, for example about defaults for modifiers or in case more specialised components and/or modifiers are missing here, like e.g. `"[ordinal padding:zero]"`. 

https://time-rs.github.io/book/api/format-description.html#components

AdrienHorgnies:
Why not, but is there an easy way to get the correct version for the crate (dynamically from the Cargo.toml) ? Otherwise, we risk people believing something is supported while it's not.

laurmaedje:
Thanks!

PR #7947, opened 2026-03-05
Fix rect radius with negative dimensions

### LINKED ISSUE #7901: Rects with negative dimensions have no radius.
### Description

This is probably related to clamping the radius to the 50% of the minimum dimension and not checking for negative values.

```typ
#set rect(fill: red, radius: 25%)
#rect(height: 1cm)\
#rect(height: -1cm)
```

Image

### Operating system

_No response_

### Typst version

collide2954:
Fixes #7901.

The radius clamping logic in `clip_rect` and `segmented_rect` computes `base_radius` as `size.x.min(size.y) / 2.0`. When a rect has negative dimensions, this produces a negative base radius, which causes the corner radius to get clamped to zero.

The fix uses `size.x.abs().min(size.y.abs())` so the radius is computed from the actual magnitude of the dimensions regardless of sign.

Added a visual test with both positive and negative height rects to verify rounded corners are preserved.

Thank you! Please let me know if there are any questions or issues.


bastienvoirin:
The test image doesn't look right, does it?

issue-7901-rect-negative-dimensions-radius

The corners of the bottom rectangle look "rotated" but I would expect both rectangles to be identical.

laurmaedje:

PR #7948, opened 2026-03-05
Fix footnote markers in math not rendering as superscript

### LINKED ISSUE #7574: Footnotes in math are not superscript

collide2954:
Fixes #7574.

`layout_inline_text()` in `math/text.rs` has a fast path for all-ASCII-digit strings that skips `layout_inline()`. The problem is that `layout_inline()` is where `ShiftSettings` get applied for super/subscript positioning. Footnote markers like "1" hit this fast path and never get shifted up.

The fix adds a check for `ShiftSettings` to the fast-path condition so it falls through to the normal text layout path when superscript styling is needed.


laurmaedje:

PR #7950, opened 2026-03-05
Use em length for default footnote entry separator line thickness

MDLC01:
This makes the footnote separator line thickness relative to the text size to prevent the line from being too thick for documents with small font size and too thin for documents with large font size. `clearance`, `gap`, and `indent` already use ems. This is the last parameter of `footnote` and `footnote.entry` to use an absolute length.

Note that those ems are relative to the outer text size, not the text size within `footnote.entry` (i.e., `show footnote.entry: set text(size: ..)` does not affect the thickness of the line, but `set text(size: ..)` does).

Related to: https://github.com/typst/typst/pull/7241.


laurmaedje:
Thanks!

Issue #7890, opened 2026-02-18
Inconsistency between conic gradients in PNG/SVG and PDF


saecki:
### Description

Discord thread: https://discord.com/channels/1054443721975922748/1239578999664152699/1473376620550291559

With the switch to krilla, the way conic gradients in PDF adapt to the container aspect ratio has changed:

| `0.13` | `0.14`/`main`|
|---|---|
|<img width="200" height="509" alt="Image" src="https://github.com/user-attachments/assets/464b8e2d-e887-4e71-9a04-690ed44a3306" />| <img width="200" height="509" alt="Image" src="https://github.com/user-attachments/assets/87cfc87c-7d9a-4f2e-8a5f-7117da51e293" />|

<details>
<summary>Source code</summary>

```typ
#grid(
  columns: 2,
  gutter: 5pt,
  rect(width: 70pt, height: 70pt, fill: gradient.linear(angle: 135deg, ..color.map.turbo)),
  rect(width: 25pt, height: 70pt, fill: gradient.linear(angle: 135deg, ..color.map.turbo)),
  rect(width: 70pt, height: 25pt, fill: gradient.linear(angle: 135deg, ..color.map.turbo)),
)

#grid(
  columns: 2,
  gutter: 5pt,
  rect(width: 70pt, height: 70pt, fill: gradient.conic(angle: 135deg, ..color.map.turbo)),
  rect(width: 25pt, height: 70pt, fill: gradient.conic(angle: 135deg, ..color.map.turbo)),
  rect(width: 70pt, height: 25pt, fill: gradient.conic(angle: 135deg, ..color.map.turbo)),
)

#grid(
  columns: 2,
  gutter: 5pt,
  rect(width: 70pt, height: 70pt, fill: gradient.radial(..color.map.turbo)),
  rect(width: 25pt, height: 70pt, fill: gradient.radial(..color.map.turbo)),
  rect(width: 70pt, height: 25pt, fill: gradient.radial(..color.map.turbo)),
)
```
</details>

The output of PNG and SVG aligns with the `0.13` PDF, and the different export formats should be visually the same. The question is what behavior we actually want.

Was the old behavior intentional? Because it seems like aspect ratio correction is applied in reverse. If no correction was applied at all, the 45deg angle should exactly meet the corner of the container.

Pinging @LaurenzV @Dherse (feel free to ignore if you're buisy)

### Typst version

- [x] I am using the latest version of Typst

eltos:
[...]

PR #7952, opened 2026-03-05
Fix conic gradient angles in typst-svg and typst-render

saecki:
As there seemed to be a general consensus regarding conic gradient angles in #7890 and the linked discord discussion, I've replicated the behavior of `krilla` for `typst-render` and `typst-svg`:
<table>
<tr>
<th>0.14</th>
<th>PR</th>
</tr>
<tr>
<td width="50%">
0.14 PNG
</td>
<td>
PR PNG
</td>
</tr>
<tr>
<td width="50%">
0.14 SVG
</td>
<td>
PR SVG
</td>
</tr>
</table>

(There is a bug in the firefox canvas API that sometimes incorrectly displays the angle of linear gradients of SVG images, so the test report might show an incorrect image diff)


laurmaedje:
@eltos Since you've done quite a bit of work in this area and are by now much more familiar with it than me, I'd be grateful if you could double-check this. Of course only if you have the time!

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -931,13 +931,19 @@ impl Gradient {
...
+
+                // Transform the angle computed from the ratio positions back
+                // into the screen space, to correct for the aspect ratio.
+                let sample_angle = -Gradient::correct_aspect_ratio(
+                    Angle::atan2(y, x),
                     Ratio::new((width / height) as f64),
                 );
-                ((-Angle::atan2(y, x) + Angle::rad(PI) + angle).to_rad() % TAU) / TAU
+
+                (sample_angle + Angle::rad(PI) + conic.angle).normalized().to_rad() / TAU
```
eltos: I stumbled about the definition of the angle, as `t` should be related to the difference between `sample_angle` minus `conic.angle`. But for some reason the stops of conic gradients are defined anti-clockwise, while the starting angle is defined clockwise from 180deg up. So it's alright.

tests/suite/visualize/gradient.typ
```diff
@@ -309,6 +309,33 @@
   fill: gradient.conic(..color.map.rainbow, space: color.hsv, angle: 90deg),
 )
 
+--- gradient-linear-angled-aspect-ratio paged ---
+#grid(
+  columns: 2,
+  gutter: 5pt,
+  rect(width: 70pt, height: 70pt, fill: gradient.linear(angle: 135deg, ..color.map.inferno)),
+  rect(width: 25pt, height: 70pt, fill: gradient.linear(angle: 135deg, ..color.map.inferno)),
+  rect(width: 70pt, height: 25pt, fill: gradient.linear(angle: 135deg, ..color.map.inferno)),
```
eltos: Here you could use a sharp gradient to make it easier to confirm and understand the test.

crates/typst-svg/src/paint.rs
```diff
@@ -167,14 +168,24 @@ impl SVGRenderer<'_> {
                     pattern.attr("y", "-0.5");
 
                     // The rotation angle, negated to match rotation in PNG.
-                    let angle = -Gradient::correct_aspect_ratio(conic.angle, *ratio);
+                    let angle = -conic.angle;
                     let center = conic.center;
 
+                    // The aspect-ratio for angle correction needs to be
+                    // inverted for conic gradients.
+                    let correct_ratio = aspect_ratio.recip();
```
eltos: The gradient is now deformed with inverse ratio, and when applied as fill, in `shape_paint_transform` a transform is added which undoes this deformation such that the result has the original angle. On first thought, that seemed convoluted, and I was wondering why not write it without deformation, and instead fix the transform in `shape_paint_transform`? However, after considering this, I realize it's required to handle gradient center correctly, and anyway the same approach is taken for linear gradients above.

So I just suggest giving this comment a bit more meaning by explaining this reasoning (something along deform with inverse aspect-ratio such that conic angle turns out right when the gradient is applied as fill with associated transform in shape_paint_transform)

tests/suite/visualize/gradient.typ
```diff
@@ -309,6 +309,33 @@
...
+)
+
+--- gradient-conic-angled-aspect-ratio paged ---
+#grid(
+  columns: 2,
+  gutter: 5pt,
+  rect(width: 70pt, height: 70pt, fill: gradient.conic(angle: 135deg, ..color.map.inferno)),
+  rect(width: 25pt, height: 70pt, fill: gradient.conic(angle: 135deg, ..color.map.inferno)),
+  rect(width: 70pt, height: 25pt, fill: gradient.conic(angle: 135deg, ..color.map.inferno)),
```
eltos: Here you could optionally add an offset to one of them to also check that the offset is correct (which it is).

crates/typst-library/src/visualize/gradient.rs
eltos: EDIT: CSS also defines the colors of conic gradients clockwise from the starting point, maybe you want to reconsider this.

laurmaedje:
Thanks a lot @eltos!

crates/typst-library/src/visualize/gradient.rs
saecki: Good point, I've changed it to also be clockwise.

crates/typst-svg/src/paint.rs
saecki: Yeah, the whole process seems somewhat convoluted but necessary.
I was a bit lazy when writing this comment, it's updated now :sweat_smile:

PR #7958, opened 2026-03-07
Prevent panic on empty tables

Genius740Code:
Panic request can be caused when processing empty tables.

This fix adds better error handling to stop panic.


saecki:
Have you tried if the latest `main` branch panics without this change? If so, could you share a minimal example where this panic happens? Because we already have tests for empty tables that currently pass:
https://github.com/typst/typst/blob/7d34dd27d62d595cf4acbc87e545a76cff129094/tests/suite/pdftags/table.typ#L2
https://github.com/typst/typst/blob/7d34dd27d62d595cf4acbc87e545a76cff129094/tests/suite/layout/table.typ#L4

saecki:
A table tag should also still be generated, even if the table has no cells. So instead the problematic piece of code should be skipped or adapted for empty tables, but the function should not early return.

laurmaedje:
Since the fix is not appropriate and there wasn't a further response, I'll go ahead and close this.

PR #7961, opened 2026-03-09
Improve wording in doc of math function call syntax

knuesel:
A very minor fix, but I think the current wording might suggest that "They support positional and named arguments, as well as argument spreading" is special for math and not true for code mode.


laurmaedje:
Thanks!

PR #7963, opened 2026-03-10
Bump krilla

laurmaedje:
Need https://github.com/LaurenzV/krilla/commit/37136e940d7695aad5c8a49e24584941d4078ac7 for bundle export

### DIFF
--- a/crates/typst-pdf/src/convert.rs
+++ b/crates/typst-pdf/src/convert.rs
@@ -506,6 +506,9 @@ fn convert_error(
             "{prefix} the PDF is missing a CMYK profile";
             hint: "CMYK colors are not yet supported in this export mode";
         ),
+        ValidationError::InconsistentSeparationFallback(_) => {
+            unreachable!("separation color spaces are not currently used")
+        }
         ValidationError::ContainsNotDefGlyph(f, loc, text) => error!(
             to_span(*loc),
             "{prefix} the text `{}` could not be displayed with {}",
--- a/crates/typst-pdf/src/tags/groups.rs
+++ b/crates/typst-pdf/src/tags/groups.rs
@@ -170,6 +170,7 @@ impl Groups {
                 TagKind::Annot(_) => Never,
                 TagKind::Figure(_) => Never,
                 TagKind::Formula(_) => Never,
+                TagKind::Form(_) => Never,
                 TagKind::NonStruct(_) => Never,
                 TagKind::Datetime(_) => Never,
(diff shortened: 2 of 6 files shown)


crates/typst-pdf/src/tags/resolve/mod.rs
```diff
@@ -391,7 +391,8 @@ fn element_kind(tag: &TagKind) -> ElementKind {
         | TagKind::Link(_)
         | TagKind::Annot(_)
         | TagKind::Figure(_)
-        | TagKind::Formula(_) => ElementKind::Inline,
+        | TagKind::Formula(_)
+        | TagKind::Form(_) => ElementKind::Inline,
```
laurmaedje, line 395: according to the PDF 1.7 spec, figure, formula, and form are all illustration elements, so I just put form with the other two

crates/typst-pdf/src/tags/groups.rs
```diff
@@ -170,6 +170,7 @@ impl Groups {
                 TagKind::Annot(_) => Never,
                 TagKind::Figure(_) => Never,
                 TagKind::Formula(_) => Never,
+                TagKind::Form(_) => Never,
```
laurmaedje, line 173: is this correct?

crates/typst-pdf/src/convert.rs
```diff
@@ -506,6 +506,9 @@ fn convert_error(
             "{prefix} the PDF is missing a CMYK profile";
             hint: "CMYK colors are not yet supported in this export mode";
         ),
+        ValidationError::InconsistentSeparationFallback(_) => {
+            unreachable!("separation color spaces are not currently used")
```
laurmaedje, line 510: a proper message will come with https://github.com/typst/typst/pull/7629. I couldn't already copy it here because it depends on extra code from that PR. but the error should be truly unreachable since we don't use separation color spaces.

tests/ref/pdf/hashes.txt
laurmaedje, line 1: These changes are additional `/CS` entries due to https://github.com/LaurenzV/krilla/commit/9d5c6c7d9deb0dad33f1a0858aec616e4cada4cf. Should be fine.

crates/typst-pdf/src/tags/groups.rs
saecki: Yes, anything remotely semantic should probably never be broken up :+1:

Issue #7735, opened 2026-01-22
Multi-file output


laurmaedje:
This issue proposes a design which would allow users to emit multiple files from a single Typst project. The main goal of this work is to enable emitting web sites with multiple pages, but the design enables further use cases. The feature would initially be guarded behind a feature flag, similarly to HTML export.

## Design

Currently, Typst supports four export formats spread across two targets: The formats PDF, PNG, and SVG are backed by the paged target and the HTML format has its own target.

This design introduces a new format and target called `bundle`. When this target is selected (from the CLI via `-f bundle`), multiple output files can be created from a single Typst file/project, all emitted to one output directory.

The design centers around the `document` function. Currently, this function can only be used in set rules. The function is extended to be directly callable. A single call to the `document` function results in one output file within the output directory.

The `document` function gains six new parameters:

- `path: string | path` (required): This parameter designates the path in the output directory of the document. It may contain interior slashes, which result in nested directories. The extension of the path is used to infer the format.

- `body: content` (required): The content that is processed to produce the file.

- `format: "pdf" | "png" | "svg" | "html" | auto` (optional): The export format. If `auto` or omitted, inferred from the path.

- `ppi: int | auto` (optional): Defines the default PPI (pixels per inch) used when rastering things. Initially, this would be used for raster export. In the future, it may also be used to define the PPI for raster effects in PDF export or for image recoding in PDF export.

- `pages: array | auto` (optional): The pages to output for paged documents. The array may contain integers indicating single pages and pairs (arrays of length two) of integers defining ranges.

- `pdf: dictionary` (optional): PDF-specific options, which includes `standard: string | auto`, and `tagged: bool | auto`. This property supports folding. We might want to split up `standard` into `version` and `standards`. This was discussed before, but not decided. If we do so, it should be split up in the CLI as well.

These configuration parameters mirror the document options exposed in the CLI (`--ppi`, `--pages`, `--pdf-standard`, `--no-pdf-tags`), with format-specific options in a dictionary, and options that may apply to multiple formats top-level. This extends the existing mapping between `document.date` and `--creation-timestamp`. Notably, we only expose export options, not environment parameters such as `--root` or `--font-path`.

In addition to the `document` function, a new `asset` function is introduced. This function can be used to emit a file whose contents are a raw string or bytes, in no particular format.

The existing experimental `target` function (guarded behind the `html` feature) returns `"bundle"` when the bundle target is selected. When within a `document`, it returns the appropriate target for that document.

### Use Cases

[...]

Issue #7342, opened 2025-11-08
Allow using page number template for PDF to produce multiple PDF files


Andrew15-5:
### Description

Previously mentioned in [the forum](https://forum.typst.app/t/splitting-pdf-files-or-pulling-individual-pdf-page-from-pdf-file/6118/7?u=andrew), it makes no sense to restrict users from using `page-{0p}-of-{t}.pdf` as the output template name, since it's already available for PNG and SVG. And now there is [another similar post](https://forum.typst.app/t/will-it-be-possible-to-create-multiple-pdf-outputs-from-a-single-call-to-typst/6791?u=andrew).

Doing [this](https://forum.typst.app/t/splitting-pdf-files-or-pulling-individual-pdf-page-from-pdf-file/6118/8?u=andrew):

```sh
for page in $(seq 1 3); do typc --pages $page a.typ page-$page.pdf; done
```

or using other tools to split a single file into multiple is the current workaround.

But since there is `--pages`, I feel like there is little work needed to add this natively. Basically all the API/infrastructure already exist.

### Use Case

Support more automation/generation workflows.

laurmaedje:
This is different, but somewhat related, so linking it here: https://github.com/typst/typst/issues/7329

dayala1:
Having a way to define the creation of different pdfs with a function similar to breakpage would be ideal for us.

For example at our university we need to create a document for each of the students in a csv file. Being able to do so automatially with a single typ file would be wonderful.

Andrew15-5:
A similar issue was mentioned in [is-it-possible-to-create-exams-with-student-names-automatically-printed-on-them/7206](https://forum.typst.app/t/is-it-possible-to-create-exams-with-student-names-automatically-printed-on-them/7206), though that is a more general feature, while this one only refers to the already existing page number template value.

PR #7964, opened 2026-03-10
Bundle export

laurmaedje:
This pull request implements a new experimental export target that can emit multiple files from a single Typst project. The implementation mostly follows the design laid out in https://github.com/typst/typst/issues/7735.

The new export target is primarily useful for multi-page websites, but is not limited to that. You can create bundles containing any combination of HTML pages, PDFs, PNGs, SVGs, and arbitrary assets.

The primary motivation to ship this new target now is to use it in the ongoing work on porting the Typst documentation to Typst. However, it's also a big step in generally making HTML export more practically useful.

### Example

```bash
typst compile main.typ --format bundle --features bundle,html
```

```typ
#document("index.html", title: "Home")[
  #title()
  - #link(<blog>)[Go to blog]
]

#document("blog.html", title: "Blog")[
  #title()
  Welcome to my blog!

  ...

  This blog also exists as a
  #link(<blog-pdf>)[single PDF].
] <blog>

#document("blog.pdf", title: "Blog")[
  ...
] <blog-pdf>

#asset(
  "favicon.ico",
  read("images/favicon.ico", encoding: none),
)
```

In the example, we create two HTML documents: A home page and a blog. The home page links to the blog through a label link. Typst's built-in linking mechanism natively supports cross-document links and resolves the correct relative paths for you. The bundle also contains a PDF version of the blog, which is linked from the HTML version. In practice, you could now share the content between the HTML and PDF version by storing it in a variable and using it in both. This is omitted here for brevity. Finally, the bundle contains an icon asset for the website. In this case, we're providing the asset's data by reading a file from disk. Alternatively, it's also possible to generate asset data from within Typst (e.g. via a function like `json.encode`).

### Scope

Compared to the design issue, this is a _slightly_ scoped down version. Specifically, for now, I omitted the `ppi`, `pages`, and `pdf` parameters to `document`. Instead, export of individual documents respects the usual CLI parameters. In principle, I still want to expose these configuration options at the Typst level, but it's not blocking for landing the work itself and also not necessary for the docs work. Moreover, to make CLI settings and in-document settings interact cleanly, a few things need sorting out first.[^1]

[As mentioned in the issue](https://github.com/typst/typst/issues/7735#issuecomment-3908841853), I've also decided to accept that introspections are always global to the bundle for the time being. Often, this is exactly what you want, but in some cases, it's unwieldy. I don't think there's a simple clean "fix" for that. Rather, improving this will be more of a long-term process of evolving the introspection subsystem.

### Notes

The history is carefully crafted with `jj` :), so if you want to take a look at what's changed / how it's implemented, I invite you to read the commits in isolation.

### Linked issues

Closes https://github.com/typst/typst/issues/7735

[^1]: Primarily that `SOURCE_DATE_EPOCH` is not perfectly handled at the moment. It behaves exactly like `--creation-timestamp` while in reality it should behave more like if the system clock happened to be it at its timestamp. I'll open a separate issue about that.

### DIFF
--- a/docs/reference/export/bundle.md
+++ b/docs/reference/export/bundle.md
@@ -1,0 +1,107 @@
+<div class="info-box">
+
+Bundle export is only available for experimentation behind a feature
+flag. Do not use this feature for production use cases. In the CLI, you can
+experiment with it by passing `--features bundle` or setting the
+`TYPST_FEATURES` environment variables to `bundle`. To use both the `bundle` and
+the `html` feature at the same time, specify both separated with a comma (i.e.
+`bundle,html`). In the web app, bundle export is not available at this time.
+</div>
+
+With Typst's bundle export, you can emit multiple output files from a single
+Typst project. Bundle output is useful for creating multi-page web sites with
+HTML export, but it is not limited to HTML export. You can create bundles
+containing any combination of [HTML pages]($html), [PDFs]($pdf), [PNGs]($png),
+[SVGs]($svg), and arbitrary [assets]($asset).
+
+# Exporting as a bundle
+## Command Line
+Pass `--format bundle` to the `compile` or `watch` subcommand. Note that you
+must also pass `--features bundle` or set `TYPST_FEATURES=bundle` to enable this
+experimental export target.
+
+When using `typst watch`, Typst will launch a live-reloading HTTP server serving
+your files. You can configure it as follows:
+
+- Pass `--port` to change the port. (Defaults to the first free port in the
+  range 3000-3005.)
+- Pass `--no-reload` to disable injection of a live reload script into HTML
+  pages. (The HTML that is written to disk isn't affected either way.) Non-HTML
+  documents do not support live reload.
+- Pass `--no-serve` to disable the server altogether.
+
+## Web App
+Not currently available.
+
+# Creating files
+A bundle is a collection of files. Each of these bundle files falls into one of
+two categories: Document or asset. A [`document`] takes [content] that is
+exported with one of Typst's other export formats. Meanwhile, an [`asset`] takes
+raw [byte data]($bytes) of your choice that will be written to disk as-is. Both
+elements take the desired output path as their first argument.
+
+The example below shows a basic example of how bundle export could be used in
+practice:
+
+```typ
+#document("index.html", title: [Home])[
+  #title()
+  - #link(<blog>)[Go to blog]
+]
+
+#document("blog.html", title: [Blog])[
+  #title()
+  Welcome to my blog!
+
+  ...
+
+  This blog also exists as a
+  #link(<blog-pdf>)[single PDF].
+] <blog>
+
+#document("blog.pdf", title: [Blog])[
+  ...
+] <blog-pdf>
+
+#asset(
+  "favicon.ico",
+  read("images/favicon.ico", encoding: none),
+)
+```
+
+In the example, we create two HTML documents: A home page and a blog. The home
+page links to the blog through a label link. Typst's built-in linking mechanism
+natively supports [cross-document links]($link/#links-in-bundle-export) and
+resolves the correct relative paths for you. The bundle also contains a PDF
+version of the blog, which is linked from the HTML version. In practice, you
+could now share the content between the HTML and PDF version by storing it in a
+variable and using it in both. This is omitted here for brevity. Finally, the
+bundle contains an icon asset for the website. In this case, we're providing the
+asset's data by reading a file from disk. Alternatively, it's also possible to
+generate asset data from within Typst (e.g. via a function like
+[`json.encode`]).
+
+Documents and assets are normal elements, so you can use them with Typst's usual
+scripting, styling, and introspection mechanisms. For more details, refer to the
+[`document`] and [`asset`] documentation.
+
+# Introspection
+Introspections always observe the full bundle rather than individual documents.
+For instance, querying for headings will give you all headings in all documents
+rather than the ones in the current document. Similarly, labels are global to
+the bundle—you can locate and [link to labels in other
+documents]($link/#links-in-bundle-export). Counters and states are likewise
+global.[^1] In particular, if you enable things like heading numbering, the
+numbering will progress consecutively across the full bundle.
+
+If you're using bundle export to build one conceptual work that is split up
+across multiple output files, this is typically what you want. If, however,
+you're using bundle export to export multiple conceptually separate works, you
+might want introspections to consider each document in isolation. Currently,
+you'll have to do this manually (e.g. by resetting counters, adjusting
+selectors, etc.) We [plan to provide more tools][introspection-comment] for
+managing the precise scope of introspection in the future.
+
+[^1]: An exception to this forms the page counter, which is (naturally) per document.
+
+[introspection-comment]: https://github.com/typst/typst/issues/7735#issuecomment-3908841853
--- a/docs/reference/export/html.md
+++ b/docs/reference/export/html.md
@@ -27,10 +27,13 @@ templates and show rules. This way, the document's contents can be fully
 agnostic to the export target and content can be shared between PDF and HTML
 export.
 
-Currently, Typst will always output a single HTML file. Support for outputting
-directories with multiple HTML documents and assets, as well as support for
-outputting fragments that can be integrated into other HTML documents is
-planned.
+In the `html` export format, Typst will output a single HTML file. You can also
+use Typst to create a web site comprising multiple HTML documents and additional
+assets. Typst calls such a collection a _bundle_ and supports this use case with
+the dedicated [`bundle` target]($bundle). Refer to its documentation to learn
+more. Note that both in the `html` and the `bundle` target, Typst will currently
+always emit a standalone HTML file. Support for emitting fragments that can be
+integrated into other HTML documents is planned for the future.
 
 Typst currently does not output CSS style sheets, instead focussing on emitting
 semantic markup. You can of course write your own CSS styles and still benefit
@@ -44,7 +47,7 @@ Pass `--format html` to the `compile` or `watch` subcommand or provide an output
 file name that ends with `.html`. Note that you must also pass `--features html`
 or set `TYPST_FEATURES=html` to enable this experimental export target.
 
-When using `typst watch`, Typst will spin up a live-reloading HTTP server. You
+When using `typst watch`, Typst will launch a live-reloading HTTP server. You
 can configure it as follows:
 
 - Pass `--port` to change the port. (Defaults to the first free port in the


Andrew15-5:
One step closer to docs in Typst!

---

After reading a bit about counters in https://github.com/typst/typst/issues/7735#issuecomment-3908841853, I'm wondering if this "solves" https://github.com/typst/typst/issues/7342 as well. If you somehow can get `counter(page.from(<doc>)).get().first()`, then you can combine https://github.com/typst/typst/issues/7735#issuecomment-3785176582 with https://github.com/typst/typst/issues/7342#issue-3604100263 and loop over `document` with different `document.page` and `document.path`. This should be enough to be a cumbersome workaround for the shell script, but run `pages` times faster, as the compiler is only run once.

crates/typst-svg/src/lib.rs
```diff
@@ -46,29 +46,48 @@ pub fn svg(page: &Page) -> String {
     xml.end_document()
 }
 
-/// Export a frame into a SVG file.
-#[typst_macros::time(name = "svg frame")]
-pub fn svg_frame(frame: &Frame) -> String {
-    let mut renderer = SVGRenderer::new();
+/// Export a page into a SVG file as part of a bundle.
```
cady-b: wouldn't it be "an" SVG (es-vee-gee)?

https://en.wikipedia.org/w/index.php?title=SVG&oldid=1335782831 Has 11 matches for "an SVG" and one match for "a SVG"

According to grep, there is also a match for this in line 36 of this file and `crates/typst-svg/src/write.rs:243`. Throughout the rest of the codebase, we use "an".

tests/suite/introspection/query.typ
```diff
@@ -201,6 +201,150 @@
   ([Frog], [GiraffeCat], [Iguana])
 )
 
+--- query-within paged html bundle ---
+// The within selector is not yet publicly exposed, but already used internally,
+// so it's good have some tests. Since it's not yet in the public API, we have
```
cady-b: good [to] have some test; I think 😅

tests/suite/layout/flow/footnote.typ
```diff
@@ -408,3 +408,14 @@ C
     Hello #footnote[Footnote 1]
   ]
 })
+
+--- footnote-bundle bundle ---
+// These two documents have a shared footnote counter.
+#document("a.html", footnote[A])
+#document("b.pdf", footnote[B])
+
+// From now on, each documents has its own.
```
cady-b: each document***s***?

crates/typst-svg/src/lib.rs
laurmaedje: just a typo

crates/typst-library/src/introspection/introspector.rs
```diff
@@ -57,8 +57,9 @@ pub trait Introspector: Send + Sync {
     /// [`Locator`](crate::introspection::Locator) docs for more details.
     fn locator(&self, key: u128, base: Location) -> Option<Location>;
 
-    /// Returns the total number pages in the document.
-    fn pages(&self) -> Option<NonZeroUsize>;
+    /// Returns the total number pages in the document which contains the given
```
cady-b: As discussed on discord (https://discord.com/channels/1054443721975922748/1080791870982078504/1480956063431004190):

> "Returns the total number [of] pages in the document ~~which~~ [that] contains the given location."

preferring "that" over "which" to make it clearer we're dealing with a restrictive clause: https://www.merriam-webster.com/grammar/when-to-use-that-and-which

docs/reference/export/bundle.md
```diff
@@ -0,0 +1,105 @@
...
+- Pass `--no-reload` to disable injection of a live reload script into HTML
+  pages. (The HTML that is written to disk isn't affected either way.) Non-HTML
+  documents do not support live reload.
+- Pass `--no-serve` to disable the server altogether.
+
+## Web App
+Not currently available.
+
+# Creating files
+Bundle files fall into two categories: Documents and assets. A [`document`]
```
reknih: Since we either are on a definition spree or will be soon: What is a _bundle file?_

docs/reference/export/bundle.md
```diff
@@ -0,0 +1,105 @@
...
+containing any combination of [HTML pages]($html), [PDFs]($pdf), [PNGs]($png),
+[SVGs]($svg), and arbitrary [assets]($asset).
+
+# Exporting as a bundle
+## Command Line
+Pass `--format bundle` to the `compile` or `watch` subcommand. Note that you
+must also pass `--features bundle` or set `TYPST_FEATURES=bundle` to enable this
+experimental export target.
+
+When using `typst watch`, Typst will spin up a live-reloading HTTP server
```
reknih: I know this is copied from `html.md`. Yet, I noticed the word _launch_ is lava to us (occurs only two times in the whole repo)

docs/reference/export/bundle.md
reknih, line 1: Maybe the HTML page would like a link to this?

docs/reference/export/bundle.md
```diff
@@ -0,0 +1,105 @@
...
+  pages. (The HTML that is written to disk isn't affected either way.) Non-HTML
+  documents do not support live reload.
+- Pass `--no-serve` to disable the server altogether.
+
+## Web App
+Not currently available.
+
+# Creating files
+Bundle files fall into two categories: Documents and assets. A [`document`]
+takes [content] that is exported with one of Typst's single-file export formats.
```
reknih: Typst's single-file -> other? idk

crates/typst-library/src/model/asset.rs
reknih, line 1: Maybe this should document the dev server behavior:

- No live reload for assets
-  `Content-Type` header derived from extension

docs/reference/export/bundle.md
```diff
@@ -0,0 +1,105 @@
...
+global.[^1] In particular, if you enable things like heading numbering, the
+numbering will progress consecutively across the full bundle.
+
+If you're using bundle export to build one conceptual work that is split up
+across multiple output files, this is typically what you want. If, however,
+you're using bundle export to export multiple conceptually separate works, you
+might want introspections to consider each document in isolation. Currently,
+you'll have to do this manually (e.g. by resetting counters, adjusting
+selectors, etc.) It is planned to provide more tools for managing the precise
+scope of introspection in the future.
```
reknih: Where can people contribute / leave input? Also, passive voice misuse.

crates/typst-library/src/model/asset.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use crate::diag::bail;
+use crate::foundations::{BundlePath, Bytes, ShowFn, Str, cast};
+use crate::introspection::Locatable;
+
+/// Raw byte data in a bundle.
+///
+/// This function creates a single file in a [bundle] from [raw byte
+/// data]($bytes). Unlike [documents]($document), assets are not processed
+/// further, their data will simply be emitted as-is.
```
reknih: What processing (other than live reload script injection) do documents undergo?

crates/typst-library/src/model/asset.rs
```diff
@@ -0,0 +1,83 @@
+use typst_macros::elem;
+
+use crate::diag::bail;
+use crate::foundations::{BundlePath, Bytes, ShowFn, Str, cast};
+use crate::introspection::Locatable;
+
+/// Raw byte data in a bundle.
+///
+/// This function creates a single file in a [bundle] from [raw byte
+/// data]($bytes). Unlike [documents]($document), assets are not processed
```
reknih:
```suggestion
/// Adds a custom file to a bundle.
///
/// This function creates a single file from [raw byte
/// data]($bytes) in a [bundle]. Unlike [documents]($document), assets are not processed
```

crates/typst-library/src/model/asset.rs
```diff
@@ -0,0 +1,83 @@
...
+/// further, their data will simply be emitted as-is.
+///
+/// The `asset` function can be combined with [`read`] to copy a file from the
+/// project into the output bundle. Notably, the path in the output bundle and
+/// the path in the project can be different.
+///
+/// ```typ
+/// #asset("styles.css", read("styles.css"))
```
reknih, line 20: Comment or explanation, in particular why the duplicated path?
```suggestion
/// // Copy the file `styles.css` into the bundle
/// #asset("styles.css", read("styles.css"))
```

crates/typst-library/src/model/asset.rs
```diff
@@ -0,0 +1,83 @@
...
+///   asset("meta.json", json.encode(meta))
+/// }
+///
+/// #document("doc.pdf")[
+///   = Introduction
+///   = Conclusion
+/// ]
+/// ```
+///
+/// This would result in the following `meta.json` file:
```
reknih:
```suggestion
/// This would emit a `meta.json` file with the following contents in the resulting bundle:
```

crates/typst-library/src/model/asset.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// The path in the bundle at which the asset should be placed.
+    ///
+    /// May contain interior slashes, in which case intermediate directories
+    /// will be automatically created.
+    #[required]
+    pub path: BundlePath,
+
+    /// The raw data that will be written into the file at the specified path.
+    ///
+    /// If a string is given, it will be written as UTF-8.
```
reknih:
```suggestion
    /// If a string is given, it will be encoded using UTF-8.
```

Also, do we link to Wikipedia for UTF-8 or assume familiarity?

crates/typst-library/src/model/document.rs
```diff
@@ -1,19 +1,21 @@
...
-    Args, Array, Construct, Content, Datetime, OneOrMultiple, Smart, StyleChain, Styles,
-    Value, cast, elem,
+    Array, BundlePath, Cast, Content, Datetime, OneOrMultiple, Packed, ShowFn, ShowSet,
+    Smart, StyleChain, Styles, Target, Value, cast, elem,
 };
+use crate::introspection::Locatable;
 use crate::text::{Locale, TextElem};
 
-/// The root element of a document and its metadata.
+/// The top-level element.
```
reknih: Maybe something like the root element of each file with its metadata? The new description is not great I feel like.

At any rate, these are bad for SEO. Ahrefs complains about meta description too short for many reference pages and Google's SERP replaces many of these descriptions or (arguabely worse) just renders them.

crates/typst-library/src/model/document.rs
```diff
@@ -1,19 +1,21 @@
...
 ///
-/// All documents are automatically wrapped in a `document` element. You cannot
-/// create a document element yourself. This function is only used with
-/// [set rules]($styling/#set-rules) to specify document metadata. Such a set
-/// rule must not occur inside of any layout container.
+/// # Single-file export
```
reknih: Many users will come to this reference page for accessibility and have no clue that they are in single-file export mode. They might think that they have multiple source files and thus are in multi-file export. I would either enumerate the targets in the title or in the first sentence.

reknih: Likewise, on the PDF, HTML, PNG, and SVG pages, I would insert something that calls them single-file export modes if we want to use this terminology.

crates/typst-library/src/model/document.rs
```diff
@@ -1,19 +1,21 @@
...
+/// In single-file export, you cannot create a document element directly.
```
reknih: > In single-file export, you cannot create a document element directly.


SpongeBob driving a car, dressed up as a posh person with a top hat, caption: But of course

crates/typst-library/src/model/document.rs
```diff
@@ -1,19 +1,21 @@
...
+/// Instead, this function is only used with [set rules]($styling/#set-rules) to
```
reknih: Passive voice misuse

crates/typst-library/src/model/document.rs
```diff
@@ -23,9 +25,64 @@ use crate::text::{Locale, TextElem};
 /// ```
 ///
 /// Note that metadata set with this function is not rendered within the
-/// document. Instead, it is embedded in the compiled PDF file.
-#[elem(Construct)]
+/// document. Instead, it is embedded in the compiled output. Metadata support
+/// varies across export formats.
```
reknih: Where can I find more info on that?

crates/typst-library/src/model/document.rs
```diff
@@ -23,9 +25,64 @@ use crate::text::{Locale, TextElem};
...
+///
+/// # Bundle export
+/// In [bundle export]($bundle), a document element represents a single output
+/// file in one of Typst's supported export formats. When creating a document,
```
reknih:
```suggestion
/// In [bundle export]($bundle), a document element represents a single file
/// in one of Typst's supported export formats in the bundle output. When creating a document,
```

crates/typst-library/src/model/document.rs
```diff
@@ -23,9 +25,64 @@ use crate::text::{Locale, TextElem};
...
+///   #title()
+///   - My
+///   - Famous
+///   - List
+/// ] <list>
+/// ```
+///
+/// Document elements pick up metadata from top-level `{set document}` rules
+/// within them. This means that documents written for single-file export can be
+/// used with explicit `document` elements while properly retaining metadata.
```
reknih: Document quirk of overriding args?

crates/typst-library/src/model/link.rs
```diff
@@ -96,6 +100,46 @@ use crate::text::{LocalName, TextElem};
...
+/// Cross-document links are emitted as relative paths (potentially with
+/// fragments). Typst automatically assigns anchor names per document based on
+/// the same rules as in HTML export. In HTML and SVG documents, these are
+/// emitted as `id` attributes on elements. In PDF documents, they are emitted
+/// as _named destinations._ PNG documents do not support linking.
+///
+/// Note that links always use full relative paths. In some scenarios (primarily
+/// for multi-page web sites), this may not be desirable. For instance, you may
+/// want to generate a `/blog/index.html` document while wanting to link to it
+/// as just `/blog`. Currently, Typst does not have a way to directly hook into
```
reknih:
```suggestion
/// as just `/blog`. Furthermore, your web server might treat `/blog` and `/blog/` as interchangeable and serve `/blog/index.html` for both. If a user then navigates to `/blog`, relative links to other pages generated by Typst will no longer work. Currently, Typst does not have a way to directly hook into
```

crates/typst-bundle/src/export.rs
```diff
@@ -0,0 +1,129 @@
...
+use typst_library::introspection::Location;
+use typst_library::model::{LateLinkResolver, PagedFormat};
+use typst_pdf::PdfOptions;
+use typst_syntax::{Span, VirtualPath};
+use typst_utils::Scalar;
+
+use crate::{Bundle, BundleDocument, BundleFile};
+
+/// A raw mapping from paths to bytes.
+pub type VirtualFs = IndexMap<VirtualPath, Bytes>;
```
saecki: Should this use `FxBuildHasher`?

crates/typst-library/src/model/asset.rs
laurmaedje: Compilation.

crates/typst-library/src/model/asset.rs
laurmaedje: I think that would fit better into planned CLI documentation. That would let us document it in one place instead of spreading it across the document and asset docs.

crates/typst-library/src/model/document.rs
```diff
@@ -1,53 +1,170 @@
...
+    /// [`path`'s]($document.path) file extension.
+    ///
+    /// This property is only supported in the [bundle] target.
+    pub format: Smart<DocumentFormat>,
+
     /// The document's title. This is rendered as the title of the PDF viewer
     /// window or the browser tab of the page.
     ///
+    /// By default, the configured title is not visibly rendered in the
+    /// document. You can add the title to the document's contents by using the
```
hpcfzl, line 149:
```suggestion
    /// document. You can add the title to the document's contents by using
```

Sorry, I know I'm late for this, but I thought I would point out this extra `the`.

PR #7979, opened 2026-03-14
Fix missing file autocompletions for `read` function

### LINKED ISSUE #7859: No file autocompletions for `read`
### Description

Using typst.app 

BUG:
Typing a file path into a function like #read("filepath")  the app does not give tab completion suggestions.

Expected Behavior:
Typing a filepath should provide a pop up in a similar way to existing functions such as #bibliography(filepath) or #image(filepath)

### Reproduction URL

_No response_

### Browsers

Firefox

### OS

Windows

mango766:
## Summary

- Adds the missing `(Some("read"), "path") => &[]` entry to the `path_completion` function in `crates/typst-ide/src/complete.rs`
- This restores file path autocompletions when typing inside `#read("")`, which regressed when the `read` function's first parameter was renamed from `source` to `path`
- Adds a corresponding test case to `test_autocomplete_file_path`

## Context

Fixes #7859.

The `path_completion` function maps `(function_name, parameter_name)` pairs to file extension filters for IDE autocompletion. When the `read` function's parameter was renamed from `source` to `path`, the completion mapping was not updated. Functions like `embed` and `attach` that also use `"path"` as their parameter name have explicit entries, but `read` was missing one.

The empty slice `&[]` means "suggest all file types", which is the correct behavior for `read` since it can read any file.

## Test plan

- [x] Added test case: `#read("")` now offers completions for all file types (images, CSVs, etc.)
- [ ] Verify `cargo test -p typst-ide test_autocomplete_file_path` passes (requires rustc >= 1.89)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -859,6 +859,7 @@ fn path_completion(func: &Func, param: &ParamInfo) -> Option<&'static [&'static 
         (Some("cite"), "style") => &["csl"],
         (Some("raw"), "syntaxes") => &["sublime-syntax"],
         (Some("raw"), "theme") => &["tmtheme"],
+        (Some("read"), "path") => &[],
         (Some("embed"), "path") => &[],
         (Some("attach"), "path") if *func == typst::pdf::AttachElem::ELEM => &[],
         (None, "path") => &[],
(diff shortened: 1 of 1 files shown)


laurmaedje:

PR #7982, opened 2026-03-14
Add divider element

### LINKED ISSUE #5318: Support horizontal rules

jakmaz:
This PR implements the `divider` element based on the design outlined by @laurmaedje in #7616. It continues the work started by @LegNeato

The divider is an empty element without its own styling. It has a default show rule that specifies its behavior.

Default behavior:
- 2 em spacing around
- Renders as 0.05em line
- HTML export to `<hr>`

By default, it basically behaves as:
```typst
#show divider: set block(spacing: 2em)
#show divider: line(length: 100%, stroke: 0.05em)
```

Users can customize it with their own show rules, for example:
```typst
#show divider: block(spacing: 1em)[
	#line(length: 80%, stroke: 2pt + blue)
]
```

Closes #5318


jakmaz:
Hey! It's my first contribution here, hope it meets your expectations :)
Let me know if you need any changes/improvements.

tests/suite/model/divider.typ
```diff
@@ -0,0 +1,62 @@
...
+#box(width: 150pt, stroke: 1pt, inset: 10pt)[
+  Content before
+  #divider()
+  Content after
+]
+
+--- divider-show-custom-line paged ---
+// Test custom line styling
+#set page(width: 200pt)
+#show divider: block[#line(length: 100%, stroke: 2pt + red)]
```
laurmaedje:
```suggestion
#show divider: block(line(length: 100%, stroke: 2pt + red))
```

tests/suite/model/divider.typ
```diff
@@ -0,0 +1,62 @@
...
+Before
+#divider()
+After
+
+--- divider-show-centered paged ---
+// Test centered, shorter divider via show rule.
+#set page(width: 200pt)
+#show divider: block(width: 100%, spacing: 1em)[
```
laurmaedje: these could also be simplified by avoiding markup mode

crates/typst-library/src/model/divider.rs
```diff
@@ -0,0 +1,38 @@
...
+/// ]
+/// First part
+/// #divider()
+/// Second part
+/// ```
+#[elem(ShowSet)]
+pub struct DividerElem {}
+
+impl ShowSet for Packed<DividerElem> {
+    fn show_set(&self, _styles: StyleChain) -> Styles {
```
laurmaedje:
```suggestion
    fn show_set(&self, _: StyleChain) -> Styles {
```

crates/typst-library/src/model/divider.rs
```diff
@@ -0,0 +1,38 @@
+use crate::foundations::{Packed, ShowSet, Smart, StyleChain, Styles, elem};
+use crate::layout::{BlockElem, Em};
+
+/// A divider.
```
laurmaedje: I think something like "A thematic break that separates sections of content" would be more useful here and then we can drop the first line of the next paragraph.

crates/typst-library/src/model/divider.rs
```diff
@@ -0,0 +1,38 @@
...
+///
+/// # Example
+/// ```example
+/// Introduction
+/// #divider()
+/// Body
+/// ```
+///
+/// # Styling
+/// The divider can be styled through show rules.
```
laurmaedje, line 18: I would suggest explaining two things here:

- Since the divider shows as a line by default, you can use `show divider: set line(..)` to adjust the stroke's color and width (Edit: This working depends on one of my other comments).
- You can also fully replace the divider with something like a floral or asterisks, but then you should wrap that in a `block`

If you want a centered floral or asterisks, I would also suggest splitting it into two rules such as
```typ
#show divider: set align(center)
#show divider: block[∗ ∗ ∗]
```
This is more composable as we can still change the alignment later.

Notably, you _don't_ need a block for the line (as in your example), because the line is already block-level.

You could also adjust the tests to test the `set line` case and split your `***` tests into two rules to test overriding `set align`.

crates/typst-library/src/model/divider.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Creates a thematic break that separates sections of content. By default,
+/// it renders as a horizontal line, but it can be customized through show
+/// rules.
+///
+/// # Example
+/// ```example
+/// Introduction
+/// #divider()
+/// Body
+/// ```
```
laurmaedje: I feel like it should be possible to make this example a bit more realistic without making it overly long. A divider is often used in a book if there is a change in scene or time, so maybe something like this:
```typ
She left without a word.
#divider()
Three days later, she returned.
```

crates/typst-layout/src/rules.rs
```diff
@@ -236,6 +237,23 @@ const DIRECT_LINK_RULE: ShowFn<DirectLinkElem> = |elem, _, _| {
...
+            thickness: Smart::Custom(Em::new(0.05).into()),
+            ..Default::default()
+        })
+        .pack()
+        .spanned(span);
+
+    Ok(BlockElem::new()
+        .with_body(Some(BlockBody::Content(line)))
+        .pack()
+        .spanned(span))
```
laurmaedje: This block is not necessary as a `LineElem` already is block-level.

crates/typst-layout/src/rules.rs
```diff
@@ -236,6 +237,23 @@ const DIRECT_LINK_RULE: ShowFn<DirectLinkElem> = |elem, _, _| {
     Ok(elem.body.clone().linked(dest, elem.alt.clone()))
 };
 
+const DIVIDER_RULE: ShowFn<DividerElem> = |elem, _, _| {
+    let span = elem.span();
+    let line = LineElem::new()
+        .with_length(Ratio::one().into())
+        .with_stroke(Stroke {
+            thickness: Smart::Custom(Em::new(0.05).into()),
```
laurmaedje: I didn't think about this in my original design, but could you move the length and thickness part of the line into the `ShowSet` rule? That would make it possible to override these settings via `show divider: set line(..)`.

jakmaz:
Thanks for the in-depth feedback and suggestions! That's really encouraging

laurmaedje:
Thank you!

PR #7985, opened 2026-03-15
Fix duplicated "the" typos in comments

phychi:


laurmaedje:
There is another one, see here: https://github.com/typst/typst/pull/7964#discussion_r2933726373. Would be nice to have that fixed as well.

phychi:
Do whatever you want — sometimes when I'm free, I just clone a project like this and use VSCode to search for duplicate words.

laurmaedje:
I'd rather have a comprehensive fix.

PR #7988, opened 2026-03-15
Fix regression in supplement of Finnish raw-kind figures

### LINKED ISSUE #7987: Finnish raw-kind figure supplement translation regression
### Description

As was originally discussed in https://github.com/typst/typst/issues/3452 and then fixed in https://github.com/typst/typst/pull/3474, the Finnish translation of raw figure supplements should be "Listaus" and not "Esimerkki".

This regressed back to "Esimerkki" at some point. Not all code listings are literal examples, but could also include algorithms or pseudocode. A more general term is needed here, such that it takes a smaller stance of what the intent of the presentation of the listing of the raw lines is. It would be better to state what type of figure is in question by default.

I believe the new location for Finnish translations is in `crates/typst-library/translations/fi.txt`. It could be fixed there.

### Operating system

Linux

### Typst version

SeSodesa:
Fixes #7987.


laurmaedje:
Seems like it regressed in https://github.com/typst/typst/pull/3728. There was another similar issue (https://github.com/typst/typst/pull/3728#issuecomment-2177050029) previously. Something seems to have gone wrong in that PR. Maybe it would be worthwhile to double-check that there aren't anymore unintended changes in there.

Anyway, thanks!

PR #7989, opened 2026-03-15
Add currently missing Finnish translations

SeSodesa:
These include translations for the term footnote, email and telephone. Notice that these are not capitalized by default. This is intentional, as in Finnish references to text objects such as figures are written in lower-case, unless they start a sentence.


laurmaedje:
Fwiw, the three added translations here are not used in normal text, only in alt text, as a standalone phrase, e.g. `Footnote 3` for a `footnote` or `Email hello@typst.app` for `#link("mailto:hello@typst.app")`.

SeSodesa:
In alt texts I think the right thing to do is to use the lower-cased forms. They do not form full sentences by themselves (you need a verb for that).

laurmaedje:
In English, it's also not a full sentence we still use upper case (and to me, it feels much better there). Currently, all languages use upper case terms. Diverging is of course possible if one language works differently, but it should only be done if it's clearly different for that language. Obviously, I can't judge Finnish, but is it really different here from all the other languages?

SeSodesa:
I don't have a really strong preference, but it would certainly be incorrect if in the midst of text there would be something like

> Lähetä sähköpostia osoitteeseen Sähköposti email@something.org.

instead of

> Lähetä sähköpostia osoitteeseen sähköposti email@something.org.

These mean "Send e-mail to the address email email@something.org." and I have substituted the alt text + the address in place of the hypothetical link element.

But I am open to capitalizing them to maintain consistency with other languages. I think a screen reader does not care about the case of the first letter.

laurmaedje:
Since it's not used in prose, let's capitalize it.

SeSodesa:
Should be done now.

laurmaedje:
Thanks!

PR #7992, opened 2026-03-17
Fix duplicate words in comments

hpcfzl:
Having commented https://github.com/typst/typst/pull/7964#discussion_r2933726373, I felt somewhat obligated to submit the change alongside @phychi's [findings](https://github.com/typst/typst/pull/7985). Through Regex `the[\s/]*the`, I have confirmed there's no more `the the`, neither in the same nor separate lines, in `.rs` files.

### DIFF
--- a/crates/typst-library/src/foundations/content/raw.rs
+++ b/crates/typst-library/src/foundations/content/raw.rs
@@ -124,7 +124,7 @@ impl RawContent {
     ///
     /// # Safety
     /// - The reference count must be zero.
-    /// - The raw content must be be of type `E`.
+    /// - The raw content must be of type `E`.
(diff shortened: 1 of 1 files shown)


cady-b:
I ran some additional checks (`rg  --pcre2 -iU "\b([a-z]+)[\s/]+\1\b"`). Yielded many false-positives, but also a few more true errors:

```
tests/src/git.rs
22:/// Runs git blame on the file path and returns a list >>of of<< git revision,

crates/typst-pdf/src/tags/groups.rs
442:    /// into table, or next >>to to<< the figure tag.

crates/typst-library/src/foundations/content/raw.rs
127:    /// - The raw content must >>be be<< of type `E`.
```

laurmaedje:
Thanks!

PR #7995, opened 2026-03-18
Clarify that `"page"` form references require a page numbering

cady-b:
Even though we do emit a hint, it was noted by caltorix [on discord](https://discord.com/channels/1054443721975922748/1088371919725793360/1483780845352259585) that it wouldn't hurt to mention in the docs as well.

### DIFF
--- a/crates/typst-library/src/model/reference.rs
+++ b/crates/typst-library/src/model/reference.rs
@@ -44,7 +44,8 @@ use crate::text::TextElem;
 /// page number at its location. You can use the
 /// [page's supplement]($page.supplement) to modify the text before the page
 /// number. Unlike a `{"normal"}` reference, the label can be attached to any
-/// element.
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/model/reference.rs
```diff
@@ -44,7 +44,7 @@ use crate::text::TextElem;
 /// page number at its location. You can use the
 /// [page's supplement]($page.supplement) to modify the text before the page
 /// number. Unlike a `{"normal"}` reference, the label can be attached to any
-/// element.
+/// element if the corresponding [page's numbering]($page.numbering) is set.
```
laurmaedje:
```suggestion
/// element as long as the corresponding [page's numbering]($page.numbering) is set.
```
(might need rewrapping)

laurmaedje:
Thanks!

PR #8000, opened 2026-03-18
More compact diagnostic tracebacks

laurmaedje:
## Description

This PR changes how diagnostic tracebacks are rendered. The current tracebacks are very verbose, making it hard to see the forest for the trees. This is especially bad when you have multiple diagnostics with traces, where there isn't a clear visual hierarchy, with color being basically the only indicator of tracepoints being conceptually nested.

The refreshed diagnostic style is more compact and focuses on what's essential. It display what kind of tracepoint it is (function call, show rule, import, include) with some details (function name, shown element, import/include path) and then underlined the path at which the tracepoint happened. Directly below, it shows only the first line of the relevant source. If there are multiple lines, it shows the last character and an ellipsis in between.

## Example

It's best to just take a look.

### Source
```typ
// main.typ
#show strong: _ => include "chap" + "ter1.typ"
*Slightly unusual
 strong text*

// chapter1.typ
#import "system.typ": my-figure
#my-figure(
  "tigers.jpg"
)

// system.typ
#let my-figure(p) = image(p)
```

### New output
The new output, see description above for details

### Old output
The old output, see description above for details

### DIFF
--- a/crates/typst-cli/tests/smoke.rs
+++ b/crates/typst-cli/tests/smoke.rs
@@ -196,6 +196,41 @@ fn test_network_access_hint() {
     output.stderr.must_contain("hint: network access is not supported");
 }
 
+#[test]
+fn test_tracepoints() {
+    let project = tempfs();
+    let main = project.write(
+        "main.typ",
+        r#"#show strong: _ => include "chap" + "ter1.typ"
+           *Slightly unusual
+            strong text*"#,
+    );
+    project.write(
+        "chapter1.typ",
+        r#"#import "system.typ": my-figure
+           #my-figure(
+             "tigers.jpg"
+           )"#,
+    );
+    project.write("system.typ", "#let my-figure(p) = image(p)");
+    let output = exec().arg("compile").arg(&main).must_fail();
+    output
+        .stderr
+        .must_contain("while calling `my-figure` at")
+        .must_contain("chapter1.typ:2:12")
+        .must_contain("my-figure(…)");
+    output
+        .stderr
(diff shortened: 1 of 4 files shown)


MDLC01:
> If there are multiple lines, it shows the last character and an ellipsis in between.

Does "the last character" refer to the last character of the line, or the last character of the affected span?

laurmaedje:
> Does "the last character" refer to the last character of the line, or the last character of the affected span?

Of the affected span.

PgBiel:
This is definitely a great improvement! Especially when there are several levels of errors going through files, some of them third-party.

Though I wonder if showing more context sometimes to solve some specific problem could still be useful. Reminds me of how `grep` has an option like `-C 10` to show 10 lines around each match. Maybe something similar could exist for errors, and could be as simple as a toggle to expand errors, which can be helpful when there are many files involved. But maybe I'm going a bit too far in my thinking, and this is probably good enough for most cases anyway (assuming you can easily access non-package source files)...

laurmaedje:
> Though I wonder if showing more context sometimes to solve some specific problem could still be useful. [...] Maybe something similar could exist for errors, and could be as simple as a toggle to expand errors, which can be helpful when there are many files involved.

I can see how that could be helpful, but I am not sure whether it would see real use compared to just opening the file (recompiling with the option, potentially stopping your watch session, is a fair bit of friction). For this PR, I think it's out of scope either way. Fwiw, it's not like the old diagnostic was guaranteed to show everything relevant either.

elegaanz (commented):
Looks good to me :)

PR #8004, opened 2026-03-20
feat: The "unknown variable" error has been eliminated; `$xyz$ output x y z` instead of the error.

benzeneringlq:
My edit: if a word is an unknown variable in math mode, show it as a sequence of symbols to prevent compilation errors.
My edition will not change compile behavior if original compile sucess.
i edit source code ,change error behavior
example:
image

### DIFF
--- a/crates/typst-eval/src/math.rs
+++ b/crates/typst-eval/src/math.rs
@@ -1,11 +1,12 @@
 use ecow::eco_format;
-use typst_library::diag::{At, SourceResult};
+use typst_library::diag::SourceResult;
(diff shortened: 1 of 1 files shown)


laurmaedje:
This would be a huge forward compatibility hazard, as every addition to the standard library would become a breaking change.

benzeneringlq:
Thanks for your reply, is me consider less.

Enivex:
It would also encourage using italic math characters in a way you are not supporting to use them. This is a pet peeve of mine in latex.

PR #8006, opened 2026-03-20
Infer CSS and JS mime types based on extension in watch server

laurmaedje:
Not really a scalable solution, but these two are very important and a better long-term solution needn't block them.

The `infer` crate is good for binary formats, but not so much for text formats. There is [`mime_guess`](https://github.com/abonander/mime_guess), but I haven't checked whether it would be good for us.

Note that this only affects the live reload server, not the actual output. This is also why I don't really want to add a mime field to the `asset` function: We wouldn't have a clean way to export it, so it would only really help for the watch server, and might actually make things more confusing why things suddenly stop working once an exported bundle is deployed on a real web server.

### DIFF
--- a/crates/typst-kit/src/server.rs
+++ b/crates/typst-kit/src/server.rs
@@ -196,6 +196,8 @@ fn select_mime_type(path: &str, buf: &[u8]) -> Option<&'static str> {
         Some("pdf") => Some("application/pdf"),
         Some("png") => Some("image/png"),
         Some("svg") => Some("image/svg+xml"),
+        Some("css") => Some("text/css"),
+        Some("js") => Some("text/javascript"),
         _ => infer::get(buf).map(|ty| ty.mime_type()),
     }
 }


bastienvoirin:
Could we add at least `Some("json") => Some("application/json"),` until a more scalable solution is found? Both

```js
import data from "/data.json";
```
and
```js
import data from "/data.json" with { type: "json" };
```
fail because of the empty content type header that is rejected:

> Le chargement du module à l’adresse « http://127.0.0.1:3000/data.json » a été bloqué en raison d’un type MIME interdit («  »).

(Loading the module at "http://127.0.0.1:3000/data.json" has been blocked because of a forbidden MIME type ("  ").)

```
GET /data.json HTTP/1.1
Host: 127.0.0.1:3000
...
Accept: application/json,*/*;q=0.5
...
```

```
HTTP/1.1 200 OK
Server: tiny-http (Rust)
Date: ...
Content-Length: ...
```

laurmaedje:
Yes, I'm okay with adding some more common ones as a stopgap.

laurmaedje:
I opened https://github.com/typst/typst/pull/8650.

PR #8007, opened 2026-03-20
Add test sources to test `report.html`

PgBiel:
Adds test sources to the test report file, so you can just share `report.html` around with full information, without having to also share or take a look at the test `suite` folder to retrieve sources:

image

Updated video:

https://github.com/user-attachments/assets/cf1c7c8d-7266-47af-be60-17206b97f294

# Extra features

There's also a button to toggle the source for each test (it is hidden by default), and another button on the sidebar to show the sources for all tests.

# TODO

- [x] Add a button to hide the test sources (maybe hidden by default)

### DIFF
--- a/tests/src/report/html.rs
+++ b/tests/src/report/html.rs
@@ -657,13 +657,16 @@ fn sidebar(parent: &mut HtmlElem, reports: &[TestReport]) {
     parent.div().class("sidebar-setting").with(|div| {
         div.text("Test Sources");
         div.fieldset().class("control-group").with(|fieldset| {
-            icon_button(
-                fieldset,
-                "global-view-test-sources",
-                "Show test sources",
-                icons::SOURCE,
-                false,
-            );
+            fieldset.label().class("icon-toggle-button").with(|label| {
+                label
+                    .input()
+                    .type_("checkbox")
+                    .id("global-view-test-sources")
+                    .title("Show test sources")
+                    .checked(false)
+                    .disabled(false);
+                svg_icon(label, icons::SOURCE);
+            });
         });
     });
 
@@ -760,11 +763,10 @@ fn test_report_header(
         });
 
     parent.fieldset().class("control-group").with(|fieldset| {
-        // The source is shown by default without JavaScript.
         fieldset
             .button()
             .class("icon-button test-report-source-toggle")
-            .aria_expanded(true)
+            .aria_expanded(false)
             .aria_controls(display!("test-report-source-{test_idx}"))
             .title("View test source")
             .with(|button| svg_icon(button, icons::SOURCE))
@@ -775,6 +777,7 @@ fn test_report_source(parent: &mut HtmlElem, test_report: &TestReport, test_idx:
     parent
         .div()
         .class("test-report-source")
+        .hidden(true)
         .id(display!("test-report-source-{test_idx}"))
         .with(|div| {
             div.table().class("text-diff").with(|table| {
--- a/tests/src/report/report.css
+++ b/tests/src/report/report.css
@@ -225,7 +225,7 @@ h2.diff-container-header {
 }
 
 .test-report-source {
-  border-bottom: 2rem solid var(--layer-2);
+  border-bottom: 1.25rem solid var(--layer-2);
 }
 
 .icon-button.test-report-toggle {
@@ -349,12 +349,12 @@ table.text-diff {
   border-spacing: 0;
   table-layout: fixed;
 
-  td.line-body:nth-child(2) {
-    border-right: var(--border-panel);
+  td.line-gutter:nth-child(3) {
+    border-left: var(--border-panel);
   }
 
-  td.diff-gap:nth-child(1) {
-    border-right: var(--border-panel);
+  td.diff-gap:nth-child(2) {
+    border-left: var(--border-panel);
   }
 }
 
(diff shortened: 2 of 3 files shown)


tests/src/report/html.rs
```diff
@@ -743,6 +758,43 @@ fn test_report_header(
                 }
             }
         });
+
+    parent.fieldset().class("control-group").with(|fieldset| {
+        // The source is shown by default without JavaScript.
```
saecki: The problem with this is that on large reports, or just reports that have complex/large SVG files the sources will be visible initially, until they are loaded, and then everything will shift around.

tests/src/report/report.css
```diff
@@ -371,6 +375,12 @@ tr {
   max-width: 50%;
 }
 
+.col-source-line-body {
```
saecki, line 378: Can't put a comment on lines that haven't been touched. But if you replace these lines, the duplicate border on the right of the source text is fixed.

https://github.com/typst/typst/pull/8007/files#diff-d0e620d6652cf7a167e45f1250b7a65359407bd5fee6f267e0348145aed0c66fR352-R358

```css
  td.line-gutter:nth-child(3) {
    border-left: var(--border-panel);
  }

  td.diff-gap:nth-child(2) {
    border-left: var(--border-panel);
  }
```

tests/src/report/report.js
```diff
@@ -134,6 +150,11 @@ for (const mode of diff_modes) {
     })
 }
 
+document.getElementById("global-view-test-sources")
+  .addEventListener("click", () => {
+    changeGlobalSourceVisibility(true)
```
saecki: I think it would make sense for this to work like a checkbox with a tri-state. So if no sources are visible: show all, if all sources are visible: hide all, and if there is a mix: hide all.

tests/src/report/report.css
```diff
@@ -224,6 +224,10 @@ h2.diff-container-header {
   }
 }
 
+.test-report-source {
+  border-bottom: 2rem solid var(--layer-2);
```
saecki: I think the wide border looks a bit odd, though I'm not how else to handle this.
Just put a normal border `var(--border-panel)` looks fine with the image diff, but when there is a text diff below, it's easy to miss :thinking:

tests/src/report/html.rs
PgBiel: For me at least everything loads instantly.

My main concern is that, without JavaScript - that is, on the site's "simplest" form -, it should always be possible to see the sources.

Mostly because my browser sometimes blocks JS automatically for privacy reasons, so I thought this would be a bit more pleasant of an experience in that case... (There's also `<noscript>` i guess.)

This doesn't matter that much though, mostly just following my own preference here. So, let me know if you'd rather have me change this (but please try it out first to make sure).

tests/src/report/report.js
PgBiel: Right. My concern was that the icon doesn't really reflect this. But I suppose I can just appropriately change the icon color

tests/src/report/html.rs
saecki: > My main concern is that, without JavaScript - that is, on the site's "simplest" form -, it should always be possible to see the sources.

If there wasn't any drawbacks, I would agree. But there are so many other things that are broken without javascript, I don't think it's worth making the intended user experience worse for this.

> This doesn't matter that much though, mostly just following my own preference here. So, let me know if you'd rather have me change this (but please try it out first to make sure).

I've tried it. Try looking at the test report with `cargo testit --base-revision 091b5ff15`. Gradients are somewhat slow to render, which makes this very noticeable. But also just large reports in general will make this noticable.

tests/src/report/report.js
saecki: Yeah that would work

tests/src/report/report.css
PgBiel: I've reduced it to 1.25rem. I think this should be a good compromise for the time being
image

tests/src/report/html.rs
PgBiel: Thank you for testing it! I will be removing it then.

tests/src/report/report.js
```diff
@@ -35,6 +36,11 @@ const reportFiles = []
  * @typedef {"visual" | "text"} DiffMode
  */
 
+let activeTestSources = 0
+
+// Avoid implicit statefulness by the browser
+globalSourceToggle.checked = false
```
saecki, line 42: I would actually like this to be the case. I often rely on this implicit state (which only seems to work in firefox) to stay on the right format tab. To "reset" the document you can still use `CTRL+F5`.
But it's nothing important and to make this work here the toggles for the sources would also have to be checkboxes. The same problem also exists for the details-like toggles that show/hide the test reports. So this could perhaps be addressed in another PR.

PgBiel: It can be done, surely, but it requires extra work. Without this line, the checkbox would be visually checked, but the test sources wouldn't be shown. So some code would be needed to check for all existing checkboxes (which would entail changing the tests to use checkboxes, as you mention) and show the appropriate test sources. But that would probably be it.

saecki: Yeah, exactly. While registering the event handlers: check the state of the toggles and count the active sources. Then set the global toggle based on that.

PR #8008, opened 2026-03-20
Ignore query parameters in watch server

laurmaedje:
So that when these happen to be populated, we don't get stray 404s.

### DIFF
--- a/crates/typst-kit/src/server.rs
+++ b/crates/typst-kit/src/server.rs
@@ -115,17 +115,22 @@ fn start_server(port: Option<u16>) -> StrResult<(SocketAddr, tiny_http::Server)>
 
 /// Handles a request.
 fn handle(req: Request, reload: bool, bucket: &Arc<RouterBucket>) -> io::Result<()> {
-    let path = req.url();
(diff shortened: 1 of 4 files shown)


crates/typst-kit/src/server.rs
```diff
@@ -115,17 +115,22 @@ fn start_server(port: Option<u16>) -> StrResult<(SocketAddr, tiny_http::Server)>
 
 /// Handles a request.
 fn handle(req: Request, reload: bool, bucket: &Arc<RouterBucket>) -> io::Result<()> {
-    let path = req.url();
+    let base = url::Url::parse("http://localhost").unwrap();
+    let Ok(url) = base.join(req.url()) else {
+        return req.respond(Response::empty(StatusCode(400)));
+    };
+
+    let path = url.path();
```
laurmaedje, line 123: Just `let path = req.url().split('?').next().unwrap();` would also do the job afaict, but since `url` is already in our dep tree, it feels cleaner to not drop to the HTTP implementation level unnecessarily.

PR #8009, opened 2026-03-20
Update docs for command line queries to use `typst eval`

laurmaedje:
`typst query` was deprecated in favor of `typst eval` in https://github.com/typst/typst/pull/7362, but the `query` docs weren't updated to that effect.

### DIFF
--- a/crates/typst-library/src/introspection/query.rs
+++ b/crates/typst-library/src/introspection/query.rs
@@ -100,10 +100,10 @@ use crate::introspection::Introspector;
 ///
 /// # Command line queries
 /// You can also perform queries from the command line, using the `typst eval`
-/// command. This command evaluates arbitrary Typst code, potentially in the
-/// context of a document, and outputs the resulting value in serialized form.
-/// It takes the code to evaluate as its first argument and (optionally) the
-/// path to a document via `--in`.
+/// command. This command evaluates Typst code, potentially in the context of a
+/// document, and outputs the resulting value in serialized form. It takes the
+/// code to evaluate as its first argument and (optionally) the path to a
+/// document via `--in`.
 ///
 /// Consider the following `example.typ` file which contains some invisible
 /// [metadata]:
@@ -128,12 +128,8 @@ use crate::introspection::Introspector;
 /// `{query(<note>)}` with access to the resulting document.
 ///
 /// **Note:** The code is surrounded with quotes to avoid special characters
-/// being interpreted by the shell. How to quote strings in the shell depends on
-/// your platform. The example above works on Unix platforms. If you're on
-/// Windows, you'll need to use double quotes instead of single quotes.
-/// ```cmd
-/// typst eval "query(<note>)" --in example.typ
-/// ```
+/// being interpreted by the shell. How to quote strings depends on your
+/// platform/shell.
 ///
 /// ## Retrieving a specific field
 ///


crates/typst-library/src/introspection/query.rs
```diff
@@ -120,23 +124,33 @@ use crate::introspection::Introspector;
...
 /// ```
 ///
-/// If you are interested in just a single element, you can use the `--one`
-/// flag to extract just it.
+/// If you are interested in just a single element, you can also use the
+/// [`first()`]($array.first) method to extract just it.
 ///
 /// ```sh
-/// $ typst query example.typ "<note>" --field value --one
+/// $ typst eval 'query(<note>).first().value' --in example.ty
```
xkevio:
```suggestion
/// $ typst eval 'query(<note>).first().value' --in example.typ
```

crates/typst-library/src/introspection/query.rs
```diff
@@ -99,18 +99,22 @@ use crate::introspection::Introspector;
 /// ```
 ///
 /// # Command line queries
-/// You can also perform queries from the command line with the `typst query`
-/// command. This command executes an arbitrary query on the document and
-/// returns the resulting elements in serialized form. Consider the following
-/// `example.typ` file which contains some invisible [metadata]:
+/// You can also perform queries from the command line, using the `typst eval`
+/// command. This command evaluates arbitrary Typst code, potentially in the
```
YDX-2147483647: > This command evaluates _arbitrary_ Typst code

The word _arbitrary_ might be inaccurate? At least `read` cannot be used.

```sh
$ typst eval 'read("a.txt")'
error: cannot access file system from here
```

crates/typst-library/src/introspection/query.rs
```diff
@@ -120,23 +124,33 @@ use crate::introspection::Introspector;
 /// ]
 /// ```
 ///
+/// This command tells Typst to compile `example.typ` and then run the code
+/// `{query(<note>)}` with access to the resulting document.
+///
+/// **Note:** The code is surrounded with quotes to avoid special characters
+/// being interpreted by the shell. How to quote strings in the shell depends on
+/// your platform. The example above works on Unix platforms. If you're on
+/// Windows, you'll need to use double quotes instead of single quotes.
```
YDX-2147483647: > If you're on Windows, you'll need to use double quotes instead of single quotes.

Not really. It applies to [Command Prompt (cmd.exe)](https://en.wikipedia.org/wiki/Cmd.exe), but not [PowerShell (pwsh)](https://en.wikipedia.org/wiki/PowerShell).

In PowerShell, `$a` in double-quoted strings also means the variable named `a`. (note the syntax highlighting)

```pwsh
> typst eval "$a$"
error: unclosed delimiter

> typst eval '$a$'
{"func":"equation","block":false,"body":{"func":"symbol","text":"a"}}
```

Image

Both cmd.exe and PowerShell[^1] are ubiquitous on Windows.
[Since Windows 10 build 14971, PowerShell replaced Command Prompt as the default command shell exposed by File Explorer.](https://en.wikipedia.org/wiki/PowerShell#:~:text=Since%20Windows%2010%20build%2014971%2C%20PowerShell%20replaced%20Command%20Prompt%20as%20the%20default%20command%20shell%20exposed%20by%20File%20Explorer%2E%5B7%5D%5B8%5D)

[^1]: To whom it may concern: I know I have confused _PowerShell_ with _Windows PowerShell_.

laurmaedje: Do you have a suggestion for how to phrase this? I find it a bit tricky.

YDX-2147483647: Well, I suggest omiting the details.

> **Note:** The code is surrounded with quotes to avoid special characters being interpreted by the shell.

- If the reader understands what CLI means, know how to open a shell and change to the directory that contains `example.typ`, and can interpret `$` in the code block correctly, then quoting shouldn't be a problem.
  (PowerShell tends to use `>` instead of `$`.)

- For most use cases of `typst eval`, both single and double quotes (in bash/zsh/pwsh) are okay. One may even want to use double quotes for commands like `typst eval "[It's]"`.

- Even if the argument isn't quoted properly, it's usually harmless.

- To be honest, I didn't know how cmd.exe interprets double quotes until 3 days ago.
  Obviously one shouldn't learn cmd.exe by reading https://typst.app/docs/.

Relevant materials:

- [White spaces and quotation marks - Quoting differences in scripting languages | Microsoft Learn](https://learn.microsoft.com/en-us/cli/azure/use-azure-cli-successfully-quoting?view=azure-cli-latest&tabs=cmd1%2Cbash2%2Cbash3#white-spaces-and-quotation-marks):

  > - If you provide a parameter value that contains white space, wrap the value in quotation marks.
  > - In Bash and PowerShell, if your variable value contains single quotes, wrap the value in double quotes, and vice-versa.
  > - In Bash, double quotes that are escaped, are treated as part of the string.
  > - In Windows Command Prompt, quotes inside variable values are treated as part of the value.

- [Course Overview + Introduction to the Shell · Missing Semester](https://missing.csail.mit.edu/2026/course-shell/#exercises):

  Quoting is included in the first lecture of the course, but the difference between single and double quotes is left as an exercise.

  > What’s the difference between `'single quotes'`, `"double quotes"`, and `$'ANSI quotes'`? Write a command that echoes a string containing a literal `$`, a `!`, and a newline character. See [Quoting](https://www.gnu.org/software/bash/manual/html_node/Quoting.html).

---

Perhaps the Typst CLI can improve the error message.

PR #8011, opened 2026-03-21
Fix crash with partially matching symbol in text show rule

### LINKED ISSUE #7991: Footnote numbering + asterisk show rule crash
### Description

I noticed when I was switching between `ast` variants I got a notification about LSP crashing, and also the preview stopped updating, which is how I found this.

I was trying to recreate the looks of an asterisk footnote, but `strong` doesn't change things much and `bold` doesn't change things at all. So maybe I'll try stroke for a bold asterisk.

```typ
#show "*": it => it
#set footnote(numbering: n => $ast.basic$)
#footnote[]
```

```
thread 'main' (3883033) panicked at crates/typst-realize/src/lib.rs:1349:64:
```

https://github.com/typst/typst/blob/b33de9de113c91c184214b299bd7a8eb3070c3ab/crates/typst-realize/src/lib.rs?plain=1#L1349

No crash with v0.13.1.

### Operating system

Linux

### Typst version

vlad2305m:
Fixes #7991.

In `typst-realize`:
Currently the `#show` dispatch code cannot process symbols if they are not matched fully (happens in math mode).
The fix is to treat symbols in the same way as text elements in `visit_regex_match`.
Symbols represent a single grapheme cluster, so they should probably never be partially matched (#8065).
Caveats:
 Introduces minor code duplication, which should go away with #8065


laurmaedje:
Thanks for the investigation! I think the fix isn't quite ideal though because it will mean show rules such as the one below will simply not work, just because of the text presentation selector:
```typ
#show "*": set text(red)
$ast.basic$
```
This would be a bit surprising to me. 

My suggestions for now would be to align the behavior of symbols with that of text. We treat a symbol just like text and slice it up into two separate symbols. This is not ideal but in line with the behavior for text. Outside of math, it will just become text anyway and in math, symbols containing only default ignorables (which the text presentation selector is) will be ignored anyway. Also, in the future, the special casing for symbols might disappear completely in favor of them becoming text earlier in the pipeline through other mechanisms, and this approach would be compatible with that anticipated change (relates to https://github.com/typst/typst/pull/7855).

Separately, we can consider consistently restricting regex show rules to not slice up grapheme clusters, uniformly for text and symbols. I'll open a separate issue about that (edit: opened https://github.com/typst/typst/issues/8058).

vlad2305m:
Implemented the suggested fix.
My initial rationale was to allow having different show rules for `"*"` and `sym.ast.basic`, but parity with text is indeed better.

Thank you for maintaining this awesome project!

laurmaedje:
Thanks!

PR #8020, opened 2026-03-24
fix: handle empty string in --font-path gracefully

### LINKED ISSUE #6183: Error when giving an empty list of extra font paths
### Observed Behavior
Typst accepts extra font paths either by the commandline option `--font-path <DIR>` or the environment variable `TYPST_FONT_PATHS`.
However, if the argument is an empty string, it fails with an error:

```bash
$ typst --version
typst 0.13.1 (unknown hash)

# Works when no extra fonts are given
$ typst fonts
Arimo Nerd Font
...

# Fails, as if the argument containing an empty string was mistaken for no argument at all.
$ typst fonts --font-path ""
error: a value is required for '--font-path <DIR>' but none was supplied

For more information, try '--help'.

# Fails with same error as with `--font-path ""`, which is additionally misleading
$ env "TYPST_FONT_PATHS=" typst fonts
error: a value is required for '--font-path <DIR>' but none was supplied

For more information, try '--help'.

# The font path doesn't need to exist, as long as the list isn't empty
$ typst fonts --font-path "/does/not/exist"
Arimo Nerd Font
...

# The font path doesn't need to exist, as long as the list isn't empty
$ env "TYPST_FONT_PATHS=/does/not/exist" typst fonts
Arimo Nerd Font
...
```

The above happens for the subcommands `compile`, `watch`, `query` and `fonts` (those are all the commands which recognize the `--font-path` flag).

### Expected Behavior
I would like Typst to accept an empty string for `--font-path` and `TYPST_FONT_PATHS`. This can happen for example when the environment variable is calculated as part of a script and it happens to resolve to an empty list.

Additionally, if there is an error message, it should correctly state whether the invalid value came from `--font-path` or `TYPST_FONT_PATHS`.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

majiayu000:
Fixes #6183

## Summary

Passing an empty string via `--font-path ""` or `TYPST_FONT_PATHS=""` currently errors with "a value is required..." because clap's default `PathBufValueParser` (which delegates to `OsStringValueParser`) rejects empty `OsStr` values.

This adds a small custom `FontPathParser` that accepts any `OsStr` input (including empty) and converts it to `PathBuf`. Empty paths are then filtered out before font scanning so they don't cause unnecessary directory lookups.

## Changes

- **`crates/typst-cli/src/args.rs`**: Add `FontPathParser` implementing `TypedValueParser<Value = PathBuf>` that accepts empty strings. Use it for the `font_paths` field.
- **`crates/typst-cli/src/fonts.rs`**: Skip empty paths in the font discovery loop.
- **`crates/typst-cli/tests/smoke.rs`**: Add `test_fonts_empty_path` covering both `--font-path ""` and `TYPST_FONT_PATHS=""`.

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -468,6 +468,7 @@ pub struct FontArgs {
         env = "TYPST_FONT_PATHS",
         value_name = "DIR",
         value_delimiter = ENV_PATH_SEP,
+        value_parser = font_path_value_parser(),
     )]
     pub font_paths: Vec<PathBuf>,
 
@@ -768,6 +769,14 @@ fn parse_page_number(value: &str) -> Result<NonZeroUsize, &'static str> {
     } else {
         NonZeroUsize::from_str(value).map_err(|_| "not a valid page number")
     }
+}
+
+/// The clap value parser used by `FontArgs.font_paths`.
+///
+/// Unlike the default `PathBufValueParser`, this accepts empty strings so that
+/// `--font-path ""` and `TYPST_FONT_PATHS=""` don't produce an error.
+fn font_path_value_parser() -> impl TypedValueParser<Value = PathBuf> {
+    clap::builder::OsStringValueParser::new().map(PathBuf::from)
 }
 
 /// The clap value parser used by `SharedArgs.input`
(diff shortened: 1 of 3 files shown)


laurmaedje:

PR #8024, opened 2026-03-25
Move single line test annotations to their annotated line

isuffix:
This PR moves every test annotation that used the `<line>:<col>-<line>:<col>` range notation but only annotated a single line to instead use the `<col>-<col>` notation by moving them to just before the line they're annotating.

This matches the output of annotations added with `--update` and should increase the consistency of annotations across the test suite.

### DIFF
--- a/tests/suite/foundations/array.typ
+++ b/tests/suite/foundations/array.typ
@@ -152,14 +152,14 @@
 
 --- array-bad-method-lvalue eval ---
 // Test bad lvalue.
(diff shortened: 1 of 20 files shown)


isuffix:
You can check this with ripgrep: `rg --pcre2 '// \w+: (\d+):\d+-\1:\d+'` giving no matches.

laurmaedje:
Some of these were intentional to increase readability of the code itself, but I don't have a strong opinion, so I'm fine with this.

Issue #7980, opened 2026-03-14
Different environment variable checks between `info` and `compile`/`watch` commands


Andrew15-5:
### Description

```sh
 typst info 2>&1 | grep System
  System fonts   on
```
```sh
 env TYPST_IGNORE_SYSTEM_FONTS=yes typst info 2>&1 | grep System
  System fonts   off
```
```sh
 env TYPST_IGNORE_SYSTEM_FONTS=yes typst compile -
error: invalid value 'yes' for '--ignore-system-fonts'
  [possible values: true, false]
For more information, try '--help'.
```

Moreover, the help message provides less information, not more:

```
      --ignore-system-fonts
          Ensures system fonts won't be searched, unless explicitly included via
          `--font-path` [env: TYPST_IGNORE_SYSTEM_FONTS=]
```

Same with `1` value.

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

PR #8030, opened 2026-03-27
Fix inconsistent environment variable checks in `typst info`

VanillaViking:
Fixes: #7980

By using `FalseyValueParser` for parsing FontArgs, environment variable checks are consistent between info, compile and watch commands.

Please let me know if any documentation needs to be changed to reflect this fix. Thanks :grin:

### DIFF
--- a/crates/typst-cli/src/info.rs
+++ b/crates/typst-cli/src/info.rs
@@ -415,7 +415,7 @@ fn parse_bool(cmd: &Command, val: &str, key: &'static str) -> Option<bool> {
         Ok(bool) => Some(bool),
         Err(_) => {
             crate::print_error(&format!(
-                "invalid value `{val}` for `{key}`. expected `true` or `false`."
+                "invalid value `{val}` for `{key}`, expected `true` or `false`."
             ))
             .map_err(|e| eco_format!("{e}"))
             .expect("failed to print error");


istudyatuni:
Maybe it's better to use [BoolishValueParser](https://docs.rs/clap/latest/clap/builder/struct.BoolishValueParser.html) to reject values, that are unrelated to true/false, like "anything", "", "100". FalseyValueParser will accept them as `true`

saecki:
I think it would rather make sense to align the behavior of `typst info` with `compile` and `watch` than the other way around.

laurmaedje:
> I think it would rather make sense to align the behavior of typst info with compile and watch than the other way around.

Agreed.

crates/typst-cli/src/info.rs
```diff
@@ -315,11 +315,18 @@ pub fn info(command: &InfoCommand) -> StrResult<()> {
...
-        // This is only an error if `v` is not valid UTF-8, which it
-        // always is.
-        FalseyValueParser::new().parse_ref(&cmd, None, v.as_ref()).ok()
-    };
+    let boolish =
+        |v: &String| match BoolValueParser::new().parse_ref(&cmd, None, v.as_ref()) {
+            Ok(bool) => Some(bool),
+            Err(_) => {
+                crate::print_error(&format!(
+                    "Invalid environment variable value: `{v}`. Expected: [true, false]"
```
saecki: This error message is a bit too imprecise. It's unclear for which variable the value was specified.

I'd also accept a name here, and change the message to something like:
```"invalid value `1` for `TYPST_IGNORE_SYSTEM_FONTS`, expected `true` or `false`."```

We also usually lowercase all error messages. I see the ```"Unknown runtime feature: `{feature}`"``` message is uppercase, but that should be changed separately.

File: crates/typst-cli/src/info.rs [hunk 2]
```diff
@@ -315,11 +315,18 @@ pub fn info(command: &InfoCommand) -> StrResult<()> {
         .map(PathBuf::from)
         .collect::<_>();
 
-    let boolish = |v: &String| {
-        // This is only an error if `v` is not valid UTF-8, which it
-        // always is.
-        FalseyValueParser::new().parse_ref(&cmd, None, v.as_ref()).ok()
-    };
+    let boolish =
```
saecki: Since this closure doesn't capture any variable, I'd just make it a function named `parse_bool` and put it next to the `parse_features` one.

crates/typst-cli/src/info.rs
laurmaedje: > We also usually lowercase all error messages. I see the "Unknown runtime feature: `{feature}`" message is uppercase, but that should be changed separately.

I already changed that on main.

VanillaViking:
Thanks for the review! I just made the changes you've requested :+1:

crates/typst-cli/src/info.rs
```diff
@@ -415,6 +409,21 @@ fn get_vars() -> StrResult<Environment> {
     })
 }
 
+/// Turns an environment variable string into a boolean value.
+fn parse_bool(cmd: &Command, val: &str, key: &'static str) -> Option<bool> {
```
saecki, line 413: Oh, never mind `cmd` was captured. In principle we could just "parse" the boolean ourselves in the match statement. But using `BoolValueParser` is probably a good idea to keep the parsing behavior aligned with clap. So all is good.

PR #8035, opened 2026-03-28
docs: list all countable elements in counter constructor documentation

### LINKED ISSUE #2728: List all elements that can be used in `counter` constructor in documentation

tranhoangtu-it:
Closes #2728

The `counter` documentation said "pages, headings, figures, and more" without specifying what "more" means. This explicitly lists all built-in elements with automatic counting support:

- `page`
- `heading` (supports multiple levels)
- `figure` (with automatic numbering)
- `math.equation`
- `footnote`

Also clarifies that any locatable element, label, or string key can be used.

### DIFF
--- a/crates/typst-library/src/introspection/counter.rs
+++ b/crates/typst-library/src/introspection/counter.rs
@@ -29,18 +29,9 @@ use crate::routines::Routines;
 /// Counts through pages, elements, and more.
 ///
 /// With the counter function, you can access and modify counters for pages,
-/// headings, figures, equations, and footnotes. These are the built-in
-/// elements that have automatic counting support:
-///
-/// - [`page`]: page counter
-/// - [`heading`]: heading / section counter (supports multiple levels)
-/// - [`figure`]: figure counter (with automatic numbering via `figure.caption`)
-/// - [`math.equation`]: equation counter
-/// - [`footnote`]: footnote counter
-///
-/// Beyond these, you can pass any [locatable]($location/#locatable)
-/// element or [label] to count its occurrences. You can also define
-/// fully custom counters using a string key.
+/// headings, figures, equations, and more. You can also create fully custom
+/// counters using a string key, or pass any [locatable]($location/#locatable)
+/// element or [label] to count its occurrences.
 ///
 /// Since counters change throughout the course of the document, their current
 /// value is _contextual._ It is recommended to read the chapter on [context]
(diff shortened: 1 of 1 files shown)


YDX-2147483647:
>  without specifying what "more" means. 

Perhaps providing too much information in the first section is overwhelming.

I suggest we move the details down to the [Constructor section](https://typst.app/docs/reference/introspection/counter/#constructor), that is:
https://github.com/typst/typst/blob/9e4d9cf223cb4f8063fd6c0d70d7c8ac458bba34/crates/typst-library/src/introspection/counter.rs#L337-L352

By moving it down, we also have  more space to explain `counter(figure.where(kind: table))`.

*Disclaimer: I am NOT a staff of Typst GmbH.*

laurmaedje (commented):
> I suggest we move the details down to the [Constructor section](https://typst.app/docs/reference/introspection/counter/#constructor), that is: [...]

I think that's a good idea!

tranhoangtu-it:
Thanks for the suggestion @YDX-2147483647 and @laurmaedje\! I've moved the countable elements list from the top-level documentation to the constructor's key parameter docs. The overview now stays concise while the constructor section provides the detailed reference.

tranhoangtu-it:
The requested changes have been pushed — moved the countable elements list to the constructor's key parameter docs as suggested. Could you remove the \ label when you get a chance? Thanks\!

crates/typst-library/src/introspection/counter.rs
```diff
@@ -348,6 +349,15 @@ impl Counter {
         ///   - provide a [`where`]($function.where) selector:
         ///     counts a type of element with specific fields,
         ///   - provide a [`{<label>}`]($label): counts elements with that label.
+        ///
+        /// The following built-in elements have automatic counting support
+        /// (their counter is stepped at each occurrence):
+        /// - [`page`]: page counter
+        /// - [`heading`]: heading / section counter (supports multiple levels)
+        /// - [`figure`]: figure counter (with automatic numbering via
```
YDX-2147483647, line 357:
```suggestion
        /// - [`figure`] of each [`kind`]($figure.kind): figure counter (with automatic numbering via
```

I suggest mentioning `kind` here. Figures of the same kind share a common counter (`counter(figure.where(kind: …))`), which is a bit more complex than `counter(heading)`, `counter(page)`, etc.  [The section under `kind`](https://typst.app/docs/reference/model/figure/#parameters-kind) also provides further guidance on how to use figure counters.

laurmaedje:
@tranhoangtu-it Do you still plan to complete this? Otherwise, I would close the PR.

PR #8037, opened 2026-03-28
docs(accent): Describe math mode accent functions

TheJanzap:
The examples in the documentation for `accent` show that in math mode, an accent can be applied by just using the accent name as a function. I added a sentence describing this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2703,9 +2703,9 @@ checksum = "55937e1799185b12863d447f42597ed69d9928686b8d88a1df17376a097d8369"
 
 [[package]]
 name = "tar"
-version = "0.4.43"
+version = "0.4.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


crates/typst-library/src/math/accent.rs
```diff
@@ -21,6 +21,9 @@ pub const ACCENT_SHORT_FALL: Em = Em::new(0.5);
 
 /// Attaches an accent to a base.
 ///
+/// In math mode, each accent name is its own function and can be used
+/// directly without `accent`.
```
laurmaedje: The wording is not fully precise, since technically the accents are `symbol`s rather than `function`s. How about something like this?

```rs
/// In math mode, common accents are also available as named [symbols]($symbol)
/// that can be directly called (like [functions]($function)) to attach them to
/// some content.
```

TheJanzap: Much better, fixed it :)

laurmaedje:
Thanks!

PR #8039, opened 2026-03-28
Fix line breaking for hosts starting with a digit

I-Al-Istannen:
## The what

This PR implements more general URL detection during line breaking, so the custom URL-breaking algorithm is _actually applied_ to links. Discussed briefly [on Discord](https://discord.com/channels/1054443721975922748/1088371867913572452/1486415096858284173), though I am not sure this qualifies as "clean fix"… I am also very happy to remove more comments, things were just new to me and I am not sure if they are new to the next person too. In my own code I would probably delete a few of them.

## The why

Sometimes bibliographies (or text!) contain URLs, which need to be broken nicely. This is already implemented in Typst by adjusting `UAX #14` and detecting links by:
- having an `UAX #14` break after `://` (i.e. the URL scheme)
- having an `UAX #14` break before `www.`

While this mostly works for URLs, it falls flat for IPs and domains starting with a digit: `UAX #14` does not produce a break after a `/` if the slash is followed by a digit, trying to keep fractions together. This causes `http://1.1.1.1` or `http://1up.example` to be ignored by the custom URL-breaking heuristic.

As a picture:

<table>
<tr>
 <th>Before</th>
 <th>After</th>
<tr>
 <td> image </td>
 <td> image </td>
</table>

## The how

This PR tries to detect a URL scheme and the separating `://`, e.g. `http://`, and then applies the custom line breaking heuristic. It does so by scanning the next 40 characters (chosen by taking the maximum registered schema length and rounding up, accounting for a +3 due to `://`) for the scheme separator `://` and then ensuring the portion up until that separator conforms to the scheme grammar.

It still relies on `UAX #14` placing a break right before the URL, but that should be fine, as long as URLs aren't directly conjoined with neighbouring words.


crates/typst-layout/src/inline/linebreak.rs
```diff
@@ -773,6 +773,52 @@ fn breakpoints(p: &Preparation, mut f: impl FnMut(usize, Breakpoint)) {
...
+}
+
+/// Guesses whether the passed in text starts with a URL scheme followed by `://`, as defined by
+/// https://www.rfc-editor.org/rfc/rfc3986#section-3.1.
+fn starts_with_url_scheme(text: &str) -> bool {
+    // We set a max length for the tail to avoid quadratic complexity in the worst case
+    // (i.e. prevent every step from scanning the whole text for the separating colon).
+    // Most URL schemes are very short, 40 was chosen as it contains all current schemes
+    // https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml.
+    let end = match text.char_indices().nth(40) {
```
laurmaedje: I'm not super comfortable with decoding the next fourty chars for every break opportunity. Do you know whether UAX #40 already uses lookahead? That would be interesting to know.

If we accept lookahead, at least it seems possible to retain the next searched `://` across multiple breakpoints if the breakpoint is still in front of it. This would turn it from a potentially quadratic (and therefore capped to 40) search into two zipped linear traversals of the string. 

Not sure how easy it is to make this clean without too much code, but worth exploring.

I-Al-Istannen: Yea, understandable. I will give it a shot :)

I-Al-Istannen: UAX does lookahead in some cases (for more complicated break rules), but not as far as necessary for URLs, I think.

I refactored it to a version that first searches for `://` and then walks back to find the longest potential scheme start. Things like `This is a url.http://` will treat `url.http` as a scheme then, which seems to work okay in practice. If we wanted to more reliably detect URLs that are adjoint to text and where no break opportunity falls before them, we need a different approach anyways.

This is maybe a bit worse than the solution you envisioned, as it still scans (potentially the whole string) first, but it at least only reads the whole thing twice at worst. Once for finding the `://` and then a backwards search to find where the scheme begins. I first prototyped a version that fused both into a single loop, but that was quite unwieldy and hard to read. I hope that memchr is fast enough that doing it once "wastefully" is okay. The string is already materialized as a str and we do not need to work on anything but chars, as URL schemes must be ascii.

laurmaedje:
I think the approach is fine now. It's not perfect, but good enough in terms of the complexity/gain trade-off. I also ran a benchmark and the impact is miniscule.

I just refactored it a bit and also added a test case that's still suboptimal, but it practically be very rare.

Thanks for your work and patience!

I-Al-Istannen:
Thanks a lot!

PR #8042, opened 2026-03-30
Fix SimSun-ExtB being incorrectly merged with SimSun

### LINKED ISSUE #6205: Typst chooses different fonts when setting `font` as `SimSun` in `text`
### Description

Recently, when using typst, I encountered a confusing behavior: when I used `set text` to specify the font as `SimSun`, typst would embed a different font into the compiled PDF instead of `SimSun`. This behavior only occurred when `set text(font: "SimSun")` was used; specifying any other font worked normally.

Initially, I tried replacing the `SimSun` font file several times, but this behavior would still occur intermittently (it wasn't consistently replaced, nor was it consistently replaced by the same other font). Subsequently, I was determined to investigate further and after a binary search of the font list on my machine, I discovered that this behavior was confusingly related to some kind of metadata of the parent folder where the font file was located. This is the minimal reproducible example I provided [case.tar.gz](https://github.com/user-attachments/files/19861818/case.tar.gz).

The composition of this test case is as follows:

```
.
├── fonts
│   ├── LiSu.ttf
│   ├── W
│   │   └── simsunb.ttf
│   └── l
│       └── SimSun.ttc
├── test.sh
├── test.typ
└── update.sh
```

```typst
// test.typ
#set text(font: "simsun")
#set text(lang: "zh", region: "CN")
这是一段测试文字。 // This is a piece of test text.
```

```bash
# test.sh
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
typst compile --ignore-system-fonts --font-path="$1" $SCRIPT_DIR/test.typ - | pdffonts -
```

```bash
# update.sh
mv "$1" /tmp/tmpfile
mv /tmp/tmpfile "$1"
rm -rf /tmp/tmpfile
```

The three font files in the example are all necessary, and `simsunb.ttf` and `SimSun.ttc` must be located in two different subdirectories, not directly in the `fonts` folder. On my machine, typst's behavior is as follows:

```bash
origami@pc ./case % ./update.sh fonts/l/
origami@pc ./case % ./test.sh fonts
name                                 type              encoding         emb sub uni object ID
------------------------------------ ----------------- ---------------- --- --- --- ---------
FCEZEI+LiSu                          CID TrueType      Identity-H       yes yes yes      4  0
origami@pc ./case % ./update.sh fonts/W
origami@pc ./case % ./test.sh fonts
name                                 type              encoding         emb sub uni object ID
------------------------------------ ----------------- ---------------- --- --- --- ---------
IWJAHX+SimSun                        CID TrueType      Identity-H       yes yes yes      4  0
```

Depending on the different "last updated times" of the `W` folder and the `l` folder, the font that typst chooses to embed in the PDF also changes.

The typst version I am using is as follows:

```bash
# typst --version
typst 0.13.1 (8ace67d9)
```

My system is archlinux with gnome. I have tested this example using docker in debian 12, and the behavior is consistent with archlinux.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

DzmingLi:
`typographic_family()` strips the suffix "ExtB" from "SimSun-ExtB", treating it as an "ExtraBold" style variant. This incorrectly merges SimSun-ExtB (a CJK  Unified Ideographs Extension B supplementary font covering U+20000–U+2A6DF)   with SimSun (the standard font covering the CJK basic block U+4E00–U+9FFF) into a single family. Since both register as Normal/400 variants,        book.select() may non-deterministically pick SimSun-ExtB, which lacks basic CJK characters, causing fallback to an unrelated system font.

The fix adds a family exception for SimSun-ExtB, consistent with similar entries for Archivo Narrow, Noto Sans Display, etc.
                                                                                
Closes #6205

### DIFF
--- a/crates/typst-library/src/text/font/exceptions.rs
+++ b/crates/typst-library/src/text/font/exceptions.rs
@@ -327,6 +327,10 @@ static EXCEPTION_MAP: phf::Map<&'static str, Exception> = phf::phf_map! {
         .family("Latin Modern Sans 12"),
     "LMSans17-Oblique" => Exception::new()
         .family("Latin Modern Sans 17"),
+    // SimSun-ExtB is a CJK Extension B font, not an "ExtraBold" variant of
+    // SimSun. Without this exception, `typographic_family()` strips the "ExtB"
+    // suffix and merges it with SimSun, causing wrong font selection.
+    "SimSun-ExtB" => Exception::new().family("SimSun-ExtB"),
     // STKaiti is a set of Kai fonts. Their weight values need to be corrected
     // according to their PostScript names.
     "STKaitiSC-Regular" => Exception::new().weight(400),


laurmaedje:
The whole name processing logic is very unfortunate and should be reconsidered at some point. Thanks for the fix!

PR #8049, opened 2026-03-30
 Warn on inconsistent equation spacing, e.g. `$x $` or `$ x$`

isuffix:

---

The first commit is what I hope is an actually clean refactoring of `Equation::block` (see discussions in #7741), via matching on the structure of the children slice.

The second commit adds a warning when equations have inconsistent spacing around their dollar signs, such as `$x $` or `$x $`. We only warn if the _presence_ of whitespace differs, not based on the content of that whitespace, so `$ x\n$` is still fine. It also fixes a few instances of inconsistent spacing in the tests.

One test that was more interesting is `math-linebreaking-trailing-linebreak`, which has a linebreak at the end of an inline equation. This is an interesting edge case because the backslash for the linebreak needs to be followed by whitespace (or it would be a character escape), so linebreak syntax just cannot be used at the end of inline equations without triggering this warning. We could remove the warning by using `#linebreak()` or `#[\ ]`, but it's questionable whether ending an inline equation with a linebreak is useful at all, so I left it with the warning and added a comment explaining this.

A choice that could differ is which span we warn on. Currently we target for the entire equation, which will noisily underline the whole thing in yellow in most editors, but we could choose to instead only target the first/last dollar sign(s). However, we do want users to notice the warning, so I'm fine with it being noisy.

I'm also open to changing the wording. I think it's currently OK, but could likely be better.

### Rationale

This was mentioned in #6430 at the very end of [this comment](https://github.com/typst/typst/issues/6430#issuecomment-3907959175), but I'll expand on the rationale a bit.

We don't want to _error_ on inconsistent spacing because we expect it to be common when initially typing an equation and you want to see the live preview without needing to have correct spacing just yet. However, we recognize that Typst's dollar sign syntax is unfamiliar, and it can be common for users to accidentally create inline equations when they intended to have a block. So we give a (pretty loud) warning that explains the syntax and hopefully gets users clarify their intent.

### DIFF
--- a/crates/typst-eval/src/math.rs
+++ b/crates/typst-eval/src/math.rs
@@ -1,5 +1,5 @@
 use ecow::eco_format;
-use typst_library::diag::{At, SourceResult};
+use typst_library::diag::{At, SourceResult, warning};
 use typst_library::foundations::{Content, NativeElement, Symbol, SymbolElem, Value};
 use typst_library::math::{
     AlignPointElem, AttachElem, EquationElem, FracElem, LrElem, PrimesElem, RootElem,
@@ -14,7 +14,23 @@ impl Eval for ast::Equation<'_> {
 
     fn eval(self, vm: &mut Vm) -> SourceResult<Self::Output> {
         let body = self.body().eval(vm)?;
-        let block = self.block();
+        let block = match self.block() {
+            ast::EquationBlock::Consistent { block } => block,
+            ast::EquationBlock::Inconsistent => {
+                vm.engine.sink.warn(warning!(
+                    self.span(),
+                    "inconsistent spacing next to opening and closing dollar signs";
+                    hint: "a block-level equation requires whitespace both after the \
+                           opening dollar sign and before the closing dollar sign";
+                    hint: "an inline equation should not have whitespace on either side";
+                    hint: "this is being treated as an inline equation";
+                ));
+                // We treat inconsistently spaced equations as inline since one
+                // of the sides didn't have a space. This avoids shifting the
+                // layout when writing `$a + $` before typing `b`.
+                false
+            }
+        };
         Ok(EquationElem::new(body).with_block(block).pack())
     }
 }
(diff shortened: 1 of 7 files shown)


crates/typst-eval/src/math.rs
```diff
@@ -14,7 +14,23 @@ impl Eval for ast::Equation<'_> {
...
         let body = self.body().eval(vm)?;
-        let block = self.block();
+        let block = match self.block() {
+            ast::EquationBlock::Consistent { block } => block,
+            ast::EquationBlock::Inconsistent => {
+                vm.engine.sink.warn(warning!(
+                    self.span(), "inconsistent spacing inside dollar signs";
+                    hint: "this is being treated as an inline equation";
+                    hint: "a block-level equation requires whitespace inside \
+                           both dollar signs";
```
laurmaedje: The phrasing "inside" seems not quite precise because whitespace _inside_ of the dollar sign itself is not really possible. It's next to a dollar sign or maybe in between the dollar signs.

One possible precise version would be "a block-level equation requires whitespace both after the opening and before the closing dollar sign". I can't quite think of a shorter more precise one, unfortunately. Maybe someone else has an idea.

What we choose here may also affect the main message and other hints of course.

isuffix: I also couldn't find a shorter precise wording. I've updated it to:
```rust
"inconsistent spacing next to opening and closing dollar signs";
hint: "a block-level equation requires whitespace both after the \
        opening dollar sign and before the closing dollar sign";
hint: "an inline equation should not have whitespace inside \
        either dollar sign";
hint: "this is being treated as an inline equation";
```
Thoughts?

laurmaedje: I'm still kinda stumbling over the "inside" word for the inline hint. Perhaps `an inline equation should not have whitespace on either side`. That still makes use of the context with the previous hint, but feels less wrong to me.

laurmaedje:
Thanks! :)

PR #8052, opened 2026-03-31
Update `error!` macros to allow spanned hints as: `hint[span]: "..."`

isuffix:
This adds syntax to the `error!`, `bail!`, and `warning!` macros for spanned hints.

I had to reconfigure the macro internals a bit to allow for a recursive macro call --- moving from `.with_hint()` taking the diagnostic by value to mutating the diagnostic --- but the result should be the same.

I also updated the three cases of manual calls that were able to use the spanned hint macro plus a single warning that didn't need the new syntax, but could still use the `warning!` macro.

### DIFF
--- a/crates/typst-library/src/diag.rs
+++ b/crates/typst-library/src/diag.rs
@@ -35,7 +35,7 @@ use crate::{World, WorldExt};
 /// bail!(
 ///     span, "returning a {} error", "formatted";
 ///     hint: "with multiple hints";
-///     hint: "the hints can have {} too", "formatting";
+///     hint[hint_span]: "hints can have custom spans and {}", "formatting";
 /// ); // SourceResult
 /// ```
 #[macro_export]
@@ -80,7 +80,7 @@ macro_rules! __bail {
 /// error!(
 ///     span, "an error with a {} message", "formatted";
 ///     hint: "with multiple hints";
-///     hint: "the hints can have {} too", "formatting";
+///     hint[hint_span]: "hints can have custom spans and {}", "formatting";
 /// ); // SourceDiagnostic
 /// ```
 #[macro_export]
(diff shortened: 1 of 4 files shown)


crates/typst-library/src/diag.rs
```diff
@@ -130,22 +145,26 @@ macro_rules! __error {
...
+        let mut warning = $crate::diag::SourceDiagnostic::warning(
             $span,
             $crate::diag::eco_format!($fmt $(, $arg)*)
-        ) $(.with_hint($crate::diag::eco_format!($hint $(, $hint_arg)*)))*
-    };
+        );
+        // We use a recursive macro for hints to allow for optional spans.
+        $($crate::diag::error!(hint$([$hint_span])?: warning, $hint $(, $hint_arg)*);)*
+        warning
+    }};
```
laurmaedje, line 167: any particular reason not keep the old approach and use `with_spanned_hint`? that would keep the amount of generated code down.

isuffix: Mainly because I wasn't able to get it to work with a single level of recursion. You can't return a partial expression like `.xyz()` from a macro, so I went with the current form, which only uses one level of recursion in `error!`.

However, this comment made me realize that we can still use the `with_hint()` form if we use multiple levels of recursion and pass an accumulating error expression into each macro call. I just tried writing it out, and it totally works, but it would generate a linear amount of recursive macro calls for the number of hints, which is concerning.

Looking it up I found that this kind of macro recursion is mentioned in The Little Book of Rust Macros as the [Push-down Accumulation pattern](https://lukaswirth.dev/tlborm/decl-macros/patterns/push-down-acc.html), and looks like a subset of the [Incremental TT Muncher pattern](https://lukaswirth.dev/tlborm/decl-macros/patterns/tt-muncher.html).

The book has this to say about [TT muncher performance](https://lukaswirth.dev/tlborm/decl-macros/patterns/tt-muncher.html#performance):

> TT munchers are inherently quadratic

> if you can write a macro using normal repetition via `*` or `+`, that should be preferred to a TT muncher.

So I think it's probably sensible to expand a bit of extra code instead of adding more recursion macros. But I haven't measured it yet, so it could go either way.

laurmaedje: ah okay, no it's fine then. I didn't realize that you can't return a partial expression.

laurmaedje:
Thanks!

PR #8054, opened 2026-03-31
Implement syntax sugar for automatic links with body

### LINKED ISSUE #8051: URL body syntax sugar
### Description

Concise syntax sugar à la Asciidoc for naming links to external webpages, as proposed in #1667:

```typst
This is a https://www.example.com[link].
```

### Use Case

With HTML and bundle export making Typst evermore viable for building websites, I expect links like these to become increasingly relevant, just as in Markdown. I find the Asciidoc variant most intuitive, since it fits very well into the existing syntax of Typst, specifically paralleling the syntax of `@references[with supplement]`. Balanced brackets are currently legal inside links, so this change would be (slightly) breaking.

As to the concern raised in #1667 of linking to local files, I believe this has been solved by #7964, though I can't say for sure if it really solves all use cases, since I haven't tested it (the development version on the web app is still a little behind).

linux-user36:
Resolves #8051. I've added minimal test and docs coverage. I haven't written any Rust before, so I heavily based the implementation off the similar syntax already used for references with supplement.


laurmaedje:
Thanks for providing an implementation! Without judgement of the underlying proposal, I just wanted to say though that it's a little early for a PR. We'll consider the proposal (though it may take some time; and I'd also like to see a bit more community input), but a syntax change is always a larger thing and not something that's just a quick merge away.

linux-user36:
No worries! It wasn't my intention to rush anything, but rather to finally get acquainted with the codebase a little. I know you guys are often busy, so I didn't want the effort of implementation to become a bottleneck. :-)

laurmaedje:
As explained in https://github.com/typst/typst/issues/8051#issuecomment-5177279878 in some more detail, we have decided that we do not want to support this syntax sugar, so I'll close this PR. Thanks for your work nonetheless!

Issue #7087, opened 2025-10-10
French words not hyphenated when containing apostrophes


Krasjet:
### Description

In French, verbs are sometimes prefixed by a pronoun with an `'`, like "je m'attends", "ils m'acceptent". Hyphenation seems to work incorrectly when I render a document in typst. For example:

```typst
#set page(width: 5em, height: auto)
#set text(lang: "fr", hyphenate: true)

J'étais sûre qu'ils ne m'accepteraient pas.

J'étais sûre qu'ils ne m' accepteraient pas.
```

<img width="110" height="369" alt="rendered image" src="https://github.com/user-attachments/assets/61a72eaf-7eaa-4896-bc7b-1788116cfc9b" />

There is a hyphenation rule to `accepteraient`, but it isn't applied when it's prefixed by `m'`. You can see when I separate the prefix manually, the hyphenation applies.

I'm not sure if it's a problem with hypher or typst. I tried the [online tool](https://typst.app/tools/hyphenate/) and it seems to work fine.

<img width="1727" height="628" alt="hyphenation online" src="https://github.com/user-attachments/assets/25c50313-e5e3-4bda-bc07-6208f19aed75" />

### Operating system

Linux

### Typst version

- [x] I am using the latest version of Typst

Pipcount:
The issue is (I think) that a `segment` with an `'` is rejected on line 762 here:
https://github.com/typst/typst/blob/689fe17d27fddc5d9d6a8067dc67dfe5e4cee777/crates/typst-layout/src/inline/linebreak.rs#L759-L767

vmartel08:
As a workaround for now, hyphenation works if you bypass the [smartquotes](https://typst.app/docs/reference/text/smartquote/) by using the apostrophe symbol.

> U+02BC : MODIFIER LETTER APOSTROPHE {apostrophe}

```typ
#set page(width: 5em, height: auto)
#set text(lang: "fr", hyphenate: true)

J'étais sûre qu'ils ne mʼaccepteraient pas.

```

<img width="122" height="230" alt="Image" src="https://github.com/user-attachments/assets/37204eed-5205-46d0-a975-d41d0e9d1f3a" />

Discussed in https://forum.typst.app/t/is-hyphenation-disabled-in-words-with-apostrophes/6611/2

Edit: As for **apostrophes**, this is also the case for **double quotes**.

```typ
#set page(width: 5em, height: auto)
#set text(lang: "fr", hyphenate: true)

[...]

Issue #1389, opened 2023-06-01
Doesn't is hyphenated as does-n't in english


Enivex:
### Description

```typst
#set text(lang:"en",hyphenate:true)
#block(width:10pt)[doesn't]
```
yields

![image](https://github.com/typst/typst/assets/2191754/cd939687-dbab-4c85-929f-97a1700a7661)

Note that doesn't shouldn't have any hyphenation points, here's the word with \usepackage{showhyphens} in LaTeX

![image](https://github.com/typst/typst/assets/2191754/e06d3b54-dfab-4bff-8c5f-4200d995ab80)

### Typst version

- [X] I am using the latest version of Typst

nino:
Providing an additional example, useful for testing edge cases:

```typst
#set text(lang:"en",hyphenate:true)
#block(width: 10pt)[You're]
#block(width: 10pt)["You're]
```

yields

![CleanShot 2023-11-03 at 15 59 56@2x](https://github.com/typst/typst/assets/125832/46534191-7586-4bb5-86df-ff06cf081630)

So the `"` seems to make "You're" hyphenatable.

Enivex:
<img width="1251" alt="Screenshot 2023-11-25 204456" src="https://github.com/typst/typst/assets/2191754/44c39afd-56a5-4f39-b136-7ff1720c0f53">

Another example of strange hyphenation. Should also be releated to the lack of lefthyphenmin and righthyphenmin, as described in https://github.com/typst/typst/issues/1941#issuecomment-1693037879

laurmaedje:
Typst does respect the hyphenmins. The problem is rather that it hyphenates the wrong strings. Rather than hyphenating proper words, it hyphenates anything between two break opportunities. This means the input to the hyphenation here is "(Onenote)", which the hyphenator doesn't deal with well. Still a bug that needs fixing of course. Just wanted to point that out.

Enivex:
> Typst does respect the hyphenmins. The problem is rather that it hyphenates the wrong strings. Rather than hyphenating proper words, it hyphenates anything between two break opportunities. This means the input to the hyphenation here is "(Onenote)", which the hyphenator doesn't deal with well. Still a bug that needs fixing of course. Just wanted to point that out.

I see! I just parroted that from the other issue since it sounded like a plausible explanation.

PR #8063, opened 2026-04-03
Fix hyphenation for Indic scripts by including combining marks in word detection

### LINKED ISSUE #8062: Hyphenation skips words containing Virama and combining marks, breaking Indic script support
### Description

Words containing Unicode combining marks — particularly Virama (the consonant joiner used in virtually all Indic scripts) — are silently skipped by the hyphenation engine. This causes hyphenation to produce no output for the majority of words in Malayalam, Devanagari, Tamil, Telugu, Kannada, and other Indic scripts.

I was testing latest hypher where I added support for 12 Indic languages. 

The words that are candidates for hyphenation is filtered as follows

https://github.com/typst/typst/blob/d6848a802e86a6269300f9768c054a641c2da77f/crates/typst-layout/src/inline/linebreak.rs#L760-L768

 

`char::is_alphabetic` maps to the Unicode Alphabetic derived property. Virama (് U+0D4D, ् U+094D, ் U+0BCD, and equivalents) is classified as `General_Category=Mn (Non-spacing Mark)` and is explicitly excluded from Alphabetic in the Unicode standard — it is a conjunct-forming control character, not a vowel sign. As a result, any segment containing a Virama fails this check and is skipped entirely.
Since Virama is used to form consonant clusters in Indic scripts, and consonant clusters are ubiquitous in these languages, this effectively disables hyphenation for the entire script family.

## Proposed Fix

Replace the `is_alphabetic` guard with a two-property check using `icu_properties` (already a Typst dependency). The `Alphabetic` property covers all letter categories (`Ll`, `Lu`, `Lt`, `Lm`, `Lo`) and already includes most Indic vowel signs via `Other_Alphabetic`. Adding the Mark group (`Mn`, `Mc`, `Me`) brings in Virama and other combining marks that are genuine word constituents but are intentionally excluded from `Alphabetic` by the Unicode spec.

### Operating system

Linux

### Typst version

santhoshtr:
Replace `char::is_alphabetic()` with a proper `is_word_char()` check that includes both the Unicode `Alphabetic` property and `General_Category =Mark`. The std method excludes combining marks like Virama (U+0D4D) and vowel signs (matras) that are integral to Indic scripts, causing words in Malayalam, Devanagari, Tamil, Telugu, and others to be silently skipped by the hyphenation engine.

Cache the ICU property data in `LazyLock` to avoid per-character reconstruction.

Fixes #8062


laurmaedje:
Thanks for the PR! There's also https://github.com/typst/typst/issues/7087 (and https://github.com/typst/typst/pull/8070), which is quite related. Do you happen to know how other engines handle this?

santhoshtr:
Interesting. So, now I am wondering why we have that is_alphabet check at all:

https://github.com/typst/typst/blob/d6848a802e86a6269300f9768c054a641c2da77f/crates/typst-layout/src/inline/linebreak.rs#L760-L768

Do you know the reason for that `segment.chars().all(char::is_alphabetic)` check? Can't we just pass all the segmented words for hyphenation ?
I tried to see what tests it breaks. There is a test to see if `doesn't` get hyphenated as `does-n't`

image

This is the only test that is failing. TeX does not hyphenate `doesn't` as `does-n't`. This can be tested as follows:
```
$ tex                      
This is TeX, Version 3.141592653 (TeX Live 2025/dev/Debian) (preloaded format=tex)

*\showhyphens{communication}

Underfull \hbox (badness 10000) detected at line 0
[] \tenrm com-mu-ni-ca-tion

*\showhyphens{doesn't}          

Underfull \hbox (badness 10000) detected at line 0
[] \tenrm doesn't
```

if we can remove that is_alphabet check, everything becomes simple - Indic languages and french will work well.

laurmaedje:
It was introduced in https://github.com/typst/typst/pull/4498 because there were various false positives. See the issues linked there.

quibo:
Good morning.  I was bounced here from #8070.
Please remove the test on "doesn't".
"Doesn't" is hyphenated as "does-n't" and I am sorry if Knuth got this wrong.  I can see the reasoning in keeping such a short sequence of letters together, but you can't throw that against large problems as my apostrophes or santhoshtr's viramas and matras (of which I know nothing).

santhoshtr:
Hi, 
Thanks for the bug reports behind the `is_alphabet check`. After going through them, especially your [reasoning](https://github.com/typst/typst/issues/1389#issuecomment-1826419071) 
> Typst does respect the hyphenmins. The problem is rather that it hyphenates the wrong strings. Rather than hyphenating proper words, it hyphenates anything between two break opportunities. This means the input to the hyphenation here is "(Onenote)", which the hyphenator doesn't deal with well. Still a bug that needs fixing of course. Just wanted to point that out.

— my understanding is we can definitely remove that `is_alphabet` condition.  The paranthesis or punctuation inside the words are handled in TeX and other applications as follows. Let us take the case of `(Doesn't)`

In Knuth’s hyphenation algorithm, the core algorithm itself **does not handle punctuation at all**. It is strictly a mathematical pattern-matching engine designed to operate exclusively on lowercase letters. This behaviour is same as hypher. 

To deal with punctuation, the system relies on a **pre-processing (tokenization) step** that strips away punctuation before feeding the resulting "pure" words to the algorithm, and then reassembles everything afterward.

### Character Categorization
In TeX, every character is assigned a specific code. Letters (a-z, A-Z) are given a lowercase equivalent (`\lccode`). Punctuation marks like `(`, and `)`` normally have an `\lccode` of 0, meaning the system does not consider them to be letters.

### Splitting into "Words"
When the text engine encounters the string `(Doesn't)`, it scans left to right, looking for contiguous blocks of valid letters.  Because punctuation marks act as boundaries, the engine breaks `(Doesn't)` into the following sequence of parts:

*   `(` — Punctuation (Ignored by the hyphenator)
*   **`Doesn`t** — Contiguous letters (Word chunk 1)
*   `)` — Punctuation (Ignored by the hyphenator)

### Applying the Algorithm

These three segments will be passed to hypher and it does the right thing, including the application of min character conditions.

However, the extra check for all characters being alphabet will prevent them to pass to hypher.

https://github.com/typst/typst/blob/d6848a802e86a6269300f9768c054a641c2da77f/crates/typst-layout/src/inline/linebreak.rs#L760-L768

I guess you meant to filter out punctuation characters (like paranthesis with length ==1) and did not mean to check if all characters in the word belong to alphabet class

```
cargo run --bin hypher -- en "doesn't"         
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/hypher en 'doesn'\''t'`
does-n't
```

_this binary is not part of hyphr, I wrote for testing_

### Reassembly

After the algorithm evaluates the letter chunks, the system pieces the original string back together exactly as it was, maintaining all capitalization and punctuation. 

### French apostrophes

For example, in French (`l'arbre`), Italian (`un'amica`), or Catalan (`l·l`), the word iterator won't break them as seperate words. Hypher would do proper hyphenation and there won't be any issue for these languages. For Malayalam, too, the word iterator behavior is correct and will recognize virama and matra as part of words as expected.

## Fix

1. We need to update the test case for doesn't - We expect `does-n't` 
2. Remove is_alphabet check, introduce a simple test if the segment is a single char - if so, no need to pass to hypher 

I just pushed the above fix and following is a test case and output:

Typst code
```typst
#set page(paper: "a5")

#set text(hyphenate: true)
#block(width: 0pt, "doesn't")
#block(width: 0pt, "(OneNote)")
#block(width: 0pt, "(present)")

#set text(lang: "de")
#block(width: 0pt, "(bzw.)")

#set text(lang: "fr")
#block(width: 0pt, "l'arbre")
#block(width: 0pt, "(l'arbre)")

#set text(lang: "it")
#block(width: 0pt, "un'amica")

#set text(lang: "ca")
#block(width: 0pt, "l·l")
```
Output:



(if the above reasoning and patch make sense, I will do edit the PR description to cover general cases)

laurmaedje:
> "Doesn't" is hyphenated as "does-n't" and I am sorry if Knuth got this wrong.

@quibo Do you have a source for that? Trying to research this does not give particularly conclusive results, but my gut feeling and the opinion of AIs I asked is that contractions cannot be hyphenated. It just looks strange.

At the very least, Merriam-Webster does not show a possible hyphenation on its page: https://www.merriam-webster.com/dictionary/doesn%27t

I'm not saying that the current algorithm is good and that any word containing an apostrophe shouldn't be hyphenated, but just regressing the "doesn't" case doesn't seem desirable either.

---

> I guess you meant to filter out punctuation characters (like paranthesis with length ==1) and did not mean to check if all characters in the word belong to alphabet class

@santhoshtr Given what you describe, doesn't TeX also categorize into letters and not letters (at least somewhat similar to the `is_alphabetic`) check? I'm not sure how you arrived at the `len() > 1` check. One the hand hand, this checks for byte length, so is always true for things beyond ASCII, but even if it were `chars().count() > 1`, I'm not sure whether `split_word_bounds` is guaranteed to separate things that are not really words into separate chunks. And then we still have the "doesn't" case, which in my opinion is a problem and which TeX somehow doesn't run into.

santhoshtr:
Hi, 
After re-reading and thinking about this issue, I found a critical flaw in my reasoning. I assumed `split_word_bounds` method will split `doesn't` as `doesn` and `t`. It does not. Based on my false assumption, I assumed that this split is analogous to what TeX does for segmentation and then hyphenation on segments. 

 ```rust
use unicode_segmentation::UnicodeSegmentation;
fn main() {
  let cases = ["(Doesn't)", "doesn't", "(OneNote)", "(present)", "l'arbre",  
     "(l'arbre)", "un'amica", "l·l", "well-known",
    "3.14", "e.g.",           "U.S.A.", "can't", "fo'c's'le", "naïve", "(bzw.)"];
  for c in &cases {
    let segs: Vec<&str> = c.split_word_bounds().collect();
    println!("{:>12}  ->  {:?}", c, segs);
  }
  // Devanagari/Malayalam combining marks
  let indic = "कार्य";
  println!("{:>12}  ->  {:?}", indic,
  indic.split_word_bounds().collect::<Vec<_>>());
}
```
```
cargo run --quiet 2>&1 | tail -25)
⎿     (Doesn't)  ->  ["(", "Doesn't", ")"]
doesn't  ->  ["doesn't"]
(OneNote)  ->  ["(", "OneNote", ")"]
(present)  ->  ["(", "present", ")"]
l'arbre  ->  ["l'arbre"]
(l'arbre)  ->  ["(", "l'arbre", ")"]
un'amica  ->  ["un'amica"]
  l·l  ->  ["l·l"]
well-known  ->  ["well", "-", "known"]
 3.14  ->  ["3.14"]
 e.g.  ->  ["e.g", "."]
U.S.A.  ->  ["U.S.A", "."]
can't  ->  ["can't"]
fo'c's'le  ->  ["fo'c's'le"]
naïve  ->  ["naïve"]
(bzw.)  ->  ["(", "bzw", ".", ")"]
कार्य   ->  ["कार\u{94d}य"]
```

UAX 29 rules WB6/WB7 (and WB11/12, MidNumLet, etc.) keep apostrophes, middle dots, and intra-word punctuation inside the segment. So `doesn't`, `l'arbre`, `un'amica`, `l·l` each come out as one segment containing the punctuation — and you then hand that whole thing to hypher, which produces `does-n't`. 

> @santhoshtr Given what you describe, doesn't TeX also categorize into letters and not letters (at least somewhat similar to the is_alphabetic) check? 

Yes it does, and I wrongly assumed split_word_bounds is equivalent to that. Tex segments by finding longest text run where char satisfies `lchar !=0` 

btw, @laurmaedje, you were right about segment.len() > 1 is byte length. I should have used `chars().count() > 1`

I studied the TeX behavior more deeply again and I found some more nuances. Earlier I said "TeX splits a word at punctuation and hyphenates each part." That is not fully accurate: TeX uses category codes to find the **first** letter run after a space and stops at the first non-letter — it does *not* restart for later runs. As a consequence, letter-connecting punctuation such as the apostrophe in French `l'organisation` is **not** hyphenated, matching base TeX (this is the limitation `babel-french` exists to work around, See https://texdoc.org/serve/babel-french/0 page 15).  So I deliberately follows the TeX behavior rather than the more permissive split.

In the updated PR, I replace the per-segment filter with `first_word`, which:

- treats a character as part of a word if it is a **letter or a combining mark** (`is_alphabetic() || General_Category = Mark`), so viramas and vowel signs stay attached to their base letters;
- scans for the **first** run of word characters in a line-break chunk, skipping any leading punctuation;
- hyphenates only that first word.

This mirrors TeX, which hyphenates the first letter run following a space and leaves anything after interior punctuation alone. The `Mark` check uses the baked `icu_properties` general-category data (the `compiled_data` feature is already enabled transitively), so no change to the bundled ICU blob is needed.

`first_word` runs only on the text between two line-break opportunities, so for ordinary space-separated prose its behavior is unchanged — the difference is only visible when a single chunk contains interior punctuation.

### Behavior vs. TeX

Verified against `tex \showhyphens` (cmr, default `\lefthyphenmin=2`, `\righthyphenmin=3`):

| Input                            | TeX                  | This PR             |
| -------------------------------- | -------------------- | ------------------- |
| `communication`                  | com-mu-ni-ca-tion    | ✓ same              |
| `(communication)`                | (com-mu-ni-ca-tion)  | ✓ same              |
| `(((communication`               | (((com-mu-ni-ca-tion | ✓ same              |
| `1communication`                 | 1com-mu-ni-ca-tion   | ✓ same              |
| `communication,communication`    | …,communication      | ✓ second left alone |
| `(communication)(communication)` | …)(communication)    | ✓ second left alone |
| `co)mmunication`                 | co)mmunication       | ✓ not hyphenated    |
| `doesn't`                        | doesn't              | ✓ not hyphenated    |
| `l'arbre`                        | l'arbre              | ✓ not hyphenated    |

The `doesn't` reference image is restored to its original (unbroken) state — the earlier `does-n't` regression is gone, because `doesn't` truncates to the first run `doesn`, which has no valid break. 

### Dependency

Requires `hypher` 0.1.7, which adds patterns for Assamese, Bengali, Gujarati, Hindi, Kannada, Malayalam, Marathi, Oriya, Panjabi, Sanskrit, Tamil and Telugu. Bumped in a separate commit; the filter fix alone has no observable effect for Indic without these patterns.

If this does not matter to you, I can seperate the hypher bump to another PR.

### Tests

- `hyphenate-first-word-only` — only the first word of a chunk is hyphenated (`communication,communication`).
- `hyphenate-indic` — Devanagari hyphenates with combining marks intact (विद्यालय → वि-द्या-लय, पुस्तकालय → पु-स्त-का-लय).

laurmaedje:
I tested with LuaLaTeX and `\usepackage{showhyphenation}` and it gave me this:

Bildschirmfoto 2026-08-04 um 13 39 24

This is not quite base TeX, but still interesting since it differs from the behavior you describe. In particular, the first one, but also the second since I did not load `babel-french`.

As-is, the PR may degrade hyphenation for long French words that have an apostrophe close the start if I understand correctly and somehow LuaLaTeX does not suffer from this. Before coming to a decision here, we would need to find out how that difference comes to be.

---

As a side note: Some parts of your latest response are clearly AI generated, while other's aren't, but all use "I". I understand using AI for research, but just forwarding AI responses to me as if they were yours feels a bit disrespectful. Especially since I put a certain trust in what you say given that you have a longer history working on hyphenation patterns.

santhoshtr:
@laurmaedje Yes, I have used AI tools to understand typst code to prepare the PR and articulate my non-native-English thoughts better for others. I think the requirements for Indic hyphenation is documented clear here. Feel free to merge, not-merge  partially or full as per the AI policy of the project. Thanks!

laurmaedje:
@santhoshtr Thanks for your response. I think (regardless of AI use) the existing investigation on what to do is very valuable, but not yet fully complete (the underlying behavior of LuaLaTeX in my latest comment is not yet clear). For this reason, I'm not yet comfortable with merging. I you'd rather not invest more time here, I fully understand that. I would then close the PR; the discussion remains useful for future work on the problem.

PR #8067, opened 2026-04-03
Fix incremental parsing of unclosed strings

### LINKED ISSUE #7393: Incremental parsing of import with string is inconsistent
## Steps to reproduce on the web app
1.  Go to https://typst.app/play and partially type out an import statement
    ```typ
    #import "@preview/unify:0.1.0
    ```
    Note the missing ending quote!!
2.  Put in the ending quote
    ```typ
    #import "@preview/unify:0.1.0"
    ```
Expectation: the package `unify` is imported.
What actually happens: `Error: Expected expression`.

Image

## Reproduction with the Rust crate
```rust
use typst::syntax::{FileId, Source, VirtualPath};

fn main() {
    // s1 parses the whole thing at once
    let s1 = Source::new(
        FileId::new_fake(VirtualPath::new("/test1")),
        "#import \"@preview/unify:0.1.0\"".to_owned(),
    );

    // whereas s2 parses an intermediate step missing an ending quote first
    let mut s2 = Source::new(
        FileId::new_fake(VirtualPath::new("/test2")),
        "#import \"@preview/unify:0.1.0".to_owned(),
    );
    // and then incrementally add the ending quote
    s2.replace("#import \"@preview/unify:0.1.0\"");

    println!("{:?}", s1.root());
    println!("{:?}", s2.root());
}
```
Output:
```txt
Markup: 30 [Hash: "#", ModuleImport: 29 [Import: "import", Space: " ", Str: "\"@preview/unify:0.1.0\""]]
Markup: 30 [Hash: "#", ModuleImport: 6 [Import: "import", Error: "" (expected expression)], Error: "" (expected semicolon or line break), Space: " ", SmartQuote: "\"", Ref: 8 [RefMarker: "@preview"], Text: "/unify", Text: ":", Text: "0.1.0", SmartQuote: "\""]
```

### Typst version

I found this bug on Typst 0.14.0.

isuffix:
Closes #7393

The root of this issue was not in the reparser, but in how we would _intially_ parse unclosed strings at the end of code expressions.

When parsing the unclosed string in `#import "str`, we would produce this tree:
```
Markup: [
    Hash: "#",
    ModuleImport: [
        Import: "import",
        Error: "" (expected expression),
    ],
    Error: "" (expected semicolon or line break),
    Space: " ",
    SmartQuote: "\"",
    Text: "str",
]
```
The unclosed string is not included in the code expression as part of either error, and its contents are actually parsed as markup!

However, this only reveals itself as a problem when we later add a closing quote and attempt to reparse `#import "str"`. The reparser starts from the location the quote was added (treating the new quote as markup) and only reparses a small section of markup expressions, leaving the code mode errors unchanged.

### Fix

The first commit upgrades the testing framework for reparsing making it more explicit and more ergonomic to edit. The second commit adds tests and implements the fix for unclosed strings in the parser.

There are two changes in the second commit that need to happen together:

First is the change to `Parser::expected` to consume lexing errors instead of skipping them. This is the core fix and resolves every case of this error I was able to think of when looking through the parser.

However this is not particularly principled, and we really ought to add some kind of fuzz testing for incremental parsing to make sure this doesn't crop up again.

Second is the change to the `ATOMIC_CODE_PRIMARY` check in `embedded_code_expr`. This check called `p.unexpected()` and relied on the call to `p.expected("expression")` at the end of  `code_primary` (via `code_expr_prec`) not consuming erroneous tokens, which became an incorrect assumption given the first change. Luckily, moving the `p.unexpected()` call inwards to `code_primary` fully preserves the original behavior (since `p.unexpected()` calling `trim_errors` would remove the empty error caused by `p.expected("expression")`). This also removed any need for the `ATOMIC_CODE_PRIMARY` syntax set, so I removed it from `set.rs` for clarity.

### DIFF
--- a/crates/typst-syntax/src/parser.rs
+++ b/crates/typst-syntax/src/parser.rs
@@ -572,13 +572,7 @@ fn embedded_code_expr(p: &mut Parser) {
         }
 
         let stmt = p.at_set(set::STMT);
-        let at = p.at_set(set::ATOMIC_CODE_EXPR);
         code_expr_prec(p, true, 0);
-
-        // Consume error for things like `#12p` or `#"abc\"`.#
-        if !at {
-            p.unexpected();
-        }
 
         // Note: 2d math arguments rely on the `directly_at` check.
         let semi = (stmt || p.directly_at(SyntaxKind::Semicolon))
@@ -592,7 +586,7 @@ fn embedded_code_expr(p: &mut Parser) {
 
 /// Parses a single code expression.
 fn code_expr(p: &mut Parser) {
-    code_expr_prec(p, false, 0)
+    code_expr_prec(p, false, 0);
 }
 
 /// Parses a code expression with at least the given precedence.
@@ -722,6 +716,9 @@ fn code_primary(p: &mut Parser, atomic: bool) {
         | SyntaxKind::Numeric
         | SyntaxKind::Str
         | SyntaxKind::Label => p.eat(),
+
+        // Consume erroneous tokens for things like `#12p`, `#]`, or `#"abc\"`.
+        _ if atomic => p.unexpected(),
 
         _ => p.expected("expression"),
     }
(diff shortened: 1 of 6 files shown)


crates/typst-syntax/src/parser.rs
```diff
@@ -1984,9 +1981,24 @@ impl Parser<'_> {
         }
     }
 
-    /// Produce an error that the given `thing` was expected.
+    /// Produce an error that the given `thing` was expected. If the parser is
+    /// at an erroneous token, this will instead eat that token and continue.
     fn expected(&mut self, thing: &str) {
-        if !self.after_error() {
+        if self.token.kind.is_error() {
```
laurmaedje, line 1987: This also affects (improves!) the errors we give for things like `#(1+12p)` or `#(1+~)`. Might be nice to add this as a test.

laurmaedje:
Thank you!

PR #8068, opened 2026-04-04
Refactor the incremental reparser algorithm

isuffix:
This PR is dependent on #8067. While working on that, I found a nice way to clarify the incremental reparser algorithm. I also added a few tests for incremental reparsing of markup added at the start/end of the text, and I rewrote/expanded many comments.

The key refactor is pulling out the `overlapping_children` function that was previously implicit in the loop at the start of `try_reparse`. This lets us separate finding the children to investigate from actually recursing inwards, and to me it really clears up the organization of the code.

I also broke out the markup expression reparsing into the `expand_and_reparse_markup` function. This is less clearly an improvement, as the function takes nearly every variable in its parent as an argument, but I like separating it because it doesn't recurse back into `try_reparse` and I find the markup reparsing to be too complicated to leave in the main function. Separating it also lets us separate the usage of the mutable children slice from the non-mutable slice with the call to `node.children().as_slice()` in the new function.

### DIFF
--- a/crates/typst-syntax/src/reparser.rs
+++ b/crates/typst-syntax/src/reparser.rs
@@ -339,6 +339,11 @@ mod tests {
         test("abc~def~gh~", Edit::Range(5..6), "+", Incr("abc~d+f~"));
         test("~~~~~~~", Edit::Range(3..4), "A", Incr("~~~A~~"));
         test("abc~~", Edit::Match("b"), "", Incr("ac~"));
+        // Newlines behave slightly differently to tildes:
+        test("~b~c~d~e", Edit::At(0), "a", Incr("a~b"));
+        test("a~b~c~d~", Edit::End, "e", Incr("~d~e"));
+        test("\nb\nc\nd\ne", Edit::At(0), "a", Incr("a\nb\n"));
+        test("a\nb\nc\nd\n", Edit::End, "e", Incr("c\nd\ne"));
         test("#var. hello", Edit::Match(" "), " ", All);
         test("#var;hello", Edit::Range(9..10), "a", All);
         test("https:/world", Edit::After("/"), "/", All);
--- a/crates/typst-syntax/src/reparser.rs
+++ b/crates/typst-syntax/src/reparser.rs
@@ -28,7 +28,30 @@ pub fn reparse(
     })
 }
 
-/// Try to reparse inside the given node.
+/// Try to reparse inside the given node, returning the range that was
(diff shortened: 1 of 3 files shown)


crates/typst-syntax/src/reparser.rs
```diff
@@ -192,15 +225,34 @@ fn try_reparse(
...
-    (first.start <= second.start && second.start <= first.end)
-        || (second.start <= first.start && first.start <= second.end)
+/// The indices and start offset of the children which overlap a replaced range.
+/// Returns `None` if the children don't fully cover the range.
+fn overlapping_children(
+    node: &SyntaxNode,
+    range: Range<usize>,
+    mut offset: usize,
+) -> Option<(Range<usize>, usize)> {
+    if !node.is_inner() || offset > range.start || range.end > offset + node.len() {
```
laurmaedje, line 239: The fact that the input range is a byte range, but the output range is a node index range could be a source of confusion. A newtype _just here_ might be a bit ad-hoc, but worth documenting.

crates/typst-syntax/src/reparser.rs
```diff
@@ -37,75 +60,83 @@ fn try_reparse(
...
         }
-
-        cursor += child.len();
     }
 
-    // Try to reparse a range of markup expressions within markup. This is only
-    // possible if the markup is top-level or contained in a block, not if it is
-    // contained in things like headings or lists because too much can go wrong
-    // with indent and line breaks.
-    if overlap.is_empty()
```
laurmaedje, line 101: this is_empty check is gone now. was it irrelevant / is it now implicit in another way?

laurmaedje (commented):
Nice change! I agree that the separation aids understanding. To the point that I couldn't even bother trying to understand/recall exactly how the old code worked during review.

crates/typst-syntax/src/reparser.rs
isuffix: Yes. `overlapping_children` will never return an empty array if it returns `Some`. I've added a comment saying so, and this can be seen statically by `index += 1` always coming immediately after the `start = index` assignment in the function.

crates/typst-syntax/src/reparser.rs
isuffix: I'll just add a sentence to the doc comment explaining it.

laurmaedje:
Thanks!

PR #8069, opened 2026-04-05
Add hint for unary operators used as atomic expressions

### LINKED ISSUE #7511: Make `#-30deg` a valid syntax
### Description

When we want an [angle](https://typst.app/docs/reference/layout/angle/) in markup or math mode, we have to switch to code mode with `#`.

At present, `#30deg` works, but `#-30deg` does not. Could we make the latter syntax also valid?

```log
error: unexpected minus
  ┌─ <stdin>:1:1
  │
1 │ #-30deg
  │  ^
```

### Use Case

This is useful for functions designed to be used in math mode. For instance, the `edge` function from [fletcher](https://typst.app/universe/package/fletcher) receives arguments in math mode, like `f`, `->`, and `<-`. To simplify writing math symbols, we sometimes use `$edge(…)$` rather than `$#edge(…)$`. 

```typst
#import "@preview/fletcher:0.5.8": diagram, edge
#diagram($
  (ker f)^complement edge(f, ->, bend: #{-30deg}) & im f
$)
#diagram($
  A edge(->, shift: #3pt) edge(<-, shift: #(-3pt)) & B
$) 
```

Image

However, `-30deg` and `-3pt` have to be wrapped with `#(…)` or `#{…}`.
We can get rid of the burden by making `#-30deg` a valid syntax.

linux-user36:
(Maybe) closes #7511 by supplying a hint:

```
error: unexpected minus
  ┌─ ../unary-hint.typ:1:1
  │
1 │ #-30deg
  │  ^
  │
  = hint: to use a unary operator, wrap the entire expression in parentheses
```


linux-user36:
Now also works for `#context -30` and the like.

crates/typst-syntax/src/parser.rs
isuffix, line 611: I think we should move the hint here to centralize our behavior for unary operators. I also think the current `unexpected` message is slightly better than the `expected` message, but I'm not fully satisfied with the wording.
```suggestion
    if p.at_set(set::UNARY_OP) {
        if !atomic {
            let op = ast::UnOp::from_kind(p.current()).unwrap();
            p.eat();
            code_expr_prec(p, atomic, op.precedence());
            p.wrap(m, SyntaxKind::Unary);
        } else {
            p.unexpected();
            p.hint("to use a unary operator, wrap the entire expression in parentheses");
        }
    } else {
        code_primary(p, atomic);
    }
```

crates/typst-syntax/src/parser.rs
```diff
@@ -575,8 +575,8 @@ fn embedded_code_expr(p: &mut Parser) {
         let at = p.at_set(set::ATOMIC_CODE_EXPR);
```
isuffix:
```suggestion
        let at = p.at_set(set::CODE_EXPR);
```
Using `CODE_EXPR` here lets you remove the `&& !p.at_set(set::UNARY_OP)` check, and will avoid an improper error with my change to `p.unexpected()` in the other suggestion.

`CODE_EXPR` is just `ATOMIC_CODE_EXPR` plus both `Underscore` and `UNARY_OP`, and the new change leaves both explicitly handled in `code_primary` and `code_expr_prec` respectively.

isuffix (changes requested):
This is a great hint, should be very helpful!

Unfortunately, this is probably going to conflict with my PR #8067 because I'm moving the `let at = p.at_set(...);` and `if at { p.unexpected(;) }` sections to `code_primary`.  To help reduce the conflict, and for general locality, I would recommend moving the error to the `if p.at_set(set::UNARY_OP)` branch in `code_expr_prec`. See my two suggestions below.

You should also add some tests to the test suite for the new behavior. The process for this is to start by finding a `.typ` file under `tests/suite` that tests something similar and add a new test case nearby. For this change, I would suggest editing `tests/suite/scripting/ops.typ` around line 362 and insert a new test like:
```typ
--- ops-unary-embedded-error eval ---
// Error: 2-3 unexpected minus
// Hint: 2-3 to use a unary operator, wrap the entire expression in parentheses
#-30deg
// Error: 2-3 unexpected plus
// Hint: 2-3 to use a unary operator, wrap the entire expression in parentheses
#+.5
// Error: 2-5 unexpected operator `not`
// Hint: 2-5 to use a unary operator, wrap the entire expression in parentheses
#not
// Error: 23-26 unexpected operator `not`
// Hint: 23-26 to use a unary operator, wrap the entire expression in parentheses
$harpoon(i, dotless: #not false)$
```
The range in the error/hint annotation comments is the character column, which should be visible in your editor somewhere. Also note that `eval` in the test header limits this to evaluation and won't produce any images to compare against.

You can compile and run the test suite with `cargo testit`, or just run a set of tests like `cargo testit ops` (for all tests with `ops` in their name). Unfortunately, because these changes are in `typst-syntax` they'll take longer to recompile than others. More info is in the [testing README](https://github.com/typst/typst/tree/main/tests).

If you change the error message or add a new test, you can update the error annotations automatically with: `cargo testit --update` (or just `-u`).

You should add your own test for `context` expressions with unary operators.

crates/typst-syntax/src/parser.rs
linux-user36: Changed it now. This also changed a test diagnostic for `underscore-invalid`, it actually emits two errors now, but I can't think of any clean solution that would clearly be better.

isuffix: Ah, that's because `Underscore` is only handled in `code_primary` with an `if !atomic` guard, meaning I lied above, apologies.

I think the best solution is actually to port over half of my change from #8067: delete the `at` variable in `embedded_code_expr` entirely and add an `_ if atomic => p.unexpected(),` branch to the match statement in `code_primary`. And if you copy the comment I wrote there exactly (``// Consume erroneous tokens for things like `#12p`, `#]`, or `#"abc\"`.``), then we might be able to get away with no rebase conflicts!

linux-user36: That seems quite pretty indeed! Everything should be working as intended now. Thanks for guiding me so thoroughly!

laurmaedje:
Thanks! That's a nice hint.

PR #8070, opened 2026-04-06
Fix hypenation bug in words with apostrophe

quibo:
Words with apostrophes should be hyphenated (example: "tomorrow's weather").  This is currently not done.

The patch to have that is super simple, just replace a `segment.chars().all` with `segment.chars().any`.

Please see [a discussion in the forum](https://forum.typst.app/t/is-hyphenation-disabled-in-words-with-apostrophes/6611/3) and [this issue](https://github.com/typst/typst/issues/7087).


laurmaedje:
I'm not sure this is the ideal fix. It will reintroduce the issues resolved by https://github.com/typst/typst/pull/4498 in the first place. There is also another open PR (https://github.com/typst/typst/pull/8063) where we're currently discussing this is greater detail. I'll close this one here in favor of that one. Thanks nonetheless!

PR #8079, opened 2026-04-08
Add breakable image support for cross-page slicing

kesavkolla:
# Summary

* Adds a breakable property to ImageElem (default false) that allows images to be sliced across page boundaries when they are taller than the remaining space on the current page
* When enabled, the image is split at page breaks and continues on subsequent pages, similar to how browsers handle images when printing to PDF
* Includes a new `layout_image_breakable` function that handles the multi-region slicing logic with proper clipping and translation

# Test Plan

- [x] Added test cases for breakable images (image-breakable, image-breakable-fits, image-breakable-false)
- [x] Updated PDF and SVG reference hashes


Enivex:
Closing for the same reason as the first one

PR #8082, opened 2026-04-09
Modify where cramped styles are applied in math

mkorje:
This brings us more in line with TeX and MathML Core (see https://github.com/typst/typst/pull/7436#discussion_r3027058216).


laurmaedje:
I'm surprised that this affects zero tests. Perhaps that's an indication that we should add some?

mkorje:
> I'm surprised that this affects zero tests. Perhaps that's an indication that we should add some?

Done

laurmaedje:
Thanks!

PR #8088, opened 2026-04-10
Retain metadata and spans in text show rules

laurmaedje:
This PR ensures that metadata and spans of textual elements matched by a text/regex show rule is retained as much as possible. This improves diagnostics and IDE functionality.

### IDE
For IDE functionality, this means that clicking on word processed by a text show rule at least goes _roughly_ to the right spot. For instance, clicking on "word" here in the preview at least goes to "my" instead of just giving up:

```typ
#show "word": underline
This is \ my word.
```

It's still not quite precise because the text elements are sliced up and text show rules can't currently apply span offsets (the `styled` element would make the `.text` field not immediately accessible) that are otherwise used for this kind of problem.

### Diagnostics
For diagnostics, it means that a construct such as
```typ
#show "mistake": _ => panic("found a forbidden word!")
I made a mistake  
```
now gives a traceback where previously it would give no indication whatsover where in the markup this happened
```
error: panicked with: "found a forbidden word!"
  ┌─ hi.typ:1:22
  │
1 │ #show "mistake": _ => panic("found a forbidden word!")
  │                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  while showing text element at hi.typ:2:0
    I made a mistake
```

### Commits
1. Small refactoring that should not affect behavior but simplifies the code
2. Test for behavior that was already in place but is now ensured
3. The main change


isuffix (commented):
Overall this looks good. I agree that removing the failsafe should be sound (although a `debug_assert!(m.is_none());` may be smart), and I think the `slice_textual` function is a nice simplification!

I do think that there's a refactor we could do to improve the inner loop, and I also think some of the comments could be a bit clearer. I've added those changes to a commit on my fork: <https://github.com/typst/typst/commit/96fa8af41dac24acf159e2bef0ab3dedf5b1650a>. The main idea of the loop refactor is that if we move the assignment of `cursor = elem_range.end` above the calls to `visit`, then we can use a `continue` statement when visiting elements that are fully outside the matched range and keep the slicing cases more grouped.

laurmaedje:
I like that refactoring. I'll add that commit to the PR.

Issue #7956, opened 2026-03-06
The size field in math.accent does not behave as expected


Lama3L9R:
### Description

The size should grow linearly, but it's actually not. The following code produces the exact same result, despite having different sizes.

```typst
math.accent(body, math.hat, size: 100%, dotless: false)
math.accent(body, math.hat, size: 133.3%, dotless: false)
```

And different symbols behave differently as well. The following code produces the exact same result, despite having different sizes.

```typst
math.accent(body, math.arrow, size: 0%, dotless: false)
math.accent(body, math.arrow, size: 100%, dotless: false)
math.accent(body, math.arrow, size: 122.2%, dotless: false)
```

### Operating system

macOS

### Typst version

- [ ] I am using the latest version of Typst

Enivex:
This is just how math fonts work. There are only a discrete number of versions of each glyph.

XanderLeaDaren:
I encountered a pretty close issue when typing vectors with arrow above. The arrow is not always stretching to “encompass” everything below.

For now, I tweaked with horizontal space, defining a function:

```
#let vect(name) = $arrow(#h(1.4pt) name #h(1.4pt))$
$ vect(u) = arrow(u) $
$ vect(V) != arrow(V) $
$ vect(A B) approx arrow(A B)$
```

<img width="308" height="203" alt="Image" src="https://github.com/user-attachments/assets/0ced7935-c36a-4bb5-a9fb-5536a709552c" />

or using `$ arrow(V, size: #1.05em) $` for example, but that seems rather complicated, right?

Enivex:
The exact width will depend on the font. It's not necessarily the case that it'll cover the entire span. This is intentional.

You can do something like `#set math.arrow(size: 120%)`  (tweak to your liking) if you really want.

YDX-2147483647:
This problem is encountered repeatedly:

- [How to define uniformly larger accents in math? - Questions - Typst Forum](https://forum.typst.app/t/how-to-define-uniformly-larger-accents-in-math/983) (2024-10)
- [Why the size of math accent are not the good size? - Questions - Typst Forum](https://forum.typst.app/t/why-the-size-of-math-accent-are-not-the-good-size/8014) (2026-02)
- #7796 (2026-02)
[...]

Issue #7796, opened 2026-02-03
Glitch when resize accent


uwni:
### Description

```typst
$ tilde(E, size: #1.0000000000001em) = tilde(E, size: #1em) = tilde(E) $
$ tilde(e, size: #1.0000000000001em) = tilde(e, size: #1em) = tilde(e) $
```
produces 

<img width="610" height="352" alt="Image" src="https://github.com/user-attachments/assets/d00cb71d-432b-4fe5-a185-9db0845779d2" />

accordingly. 
Even if we don't determine to use which width as the default one. At least, in my opinion, `.0000000000001em` shouldn't differ so much like that.

### Typst version

- [x] I am using the latest version of Typst

uwni:
In addition
```typst
#set rect(height: 1em, stroke: none, fill: blue)
$
  tilde(std.rect(width: #1em))\
  tilde(std.rect(width: #1.000000000000001em))
$
```
will give

<img width="316" height="592" alt="Image" src="https://github.com/user-attachments/assets/0c633929-a9e1-458d-b26d-3eaa364484f4" />

## Solutions

Maybe we can always make `exact_frame_width` be true when a size (like `1em` or `100%`) is specified. That should be considered as the users' intention.

mkorje:
This is basically due to a combination of the accent short fall that is hardcoded in and there being discrete size variants of the tilde accent in fonts.

Just making the accent size match exactly the width of the base is undesirable a lot of the time. And the 0.0000001em difference is just enough for the target size to miss the previous tilde variant and move on to the next (larger) one.

laurmaedje:
Closed through https://github.com/typst/typst/pull/8096 by documenting this behavior.

Issue #7999, opened 2026-03-18
`stretch(->)` with STIX Two Math emits a warning despite correct rendering


FA555:
### Description

I ran into this warning when using `math.stretch(->)` with STIX Two Math:

```
warning: glyph has assembly parts with overlap less than minConnectorOverlap
 = hint: its rendering may appear broken - this is probably a font bug
 = hint: please file an issue at https://github.com/typst/typst/issues
````

The rendered result *looks fine*, but the warning is reproducible.

Here is a minimal example that triggers it:

```typ
#show math.equation: set text(font: "STIX Two Math")

$ stretch(->)^#h(237 / 175 * 1em + 1e-15em) $
```

While this does **not** produce the warning:

```typ
#show math.equation: set text(font: "STIX Two Math")

$ stretch(->)^#h(237 / 175 * 1em) $
```

I also tested `$ stretch(->)^#lorem(20) $`. The warning still appears and the result still looks correct.

---

I tested PDF, PNG, and SVG as target formats, getting the same result.

I tested both of these fonts, getting the same result:

- [`STIXTwoMath-Regular.otf`](https://github.com/stipub/stixfonts/blob/master/fonts/static_otf/STIXTwoMath-Regular.otf)
- [`STIXTwoMath-Regular.ttf`](https://github.com/stipub/stixfonts/blob/master/fonts/static_ttf/STIXTwoMath-Regular.ttf)

I tested this with both of these Typst distributions, getting the same result:

- Typst (stable) 0.14.2
- Current typst/typst main: 3c73671d3a240b3a6afca010ba586b04b0616b95

---

A screenshot of a minimal reproducible example is attached:

![MSE](https://github.com/user-attachments/assets/1a991937-a2b0-4f52-a6d3-a6961aceed19)

### Operating system

macOS

### Typst version

- [x] I am using the latest version of Typst

Enivex:
This is why the warning says "may". It simply cannot guarantee that the result will look correct since the font has issues.

@mkorje will know, but this is unlikely to be a bug.

mkorje:
What @Enivex said is spot on. So I'd say this isn't a bug (though I can understand that it can be a bit annoying).

MDLC01:
Is this an issue that should be reported to STIX Two Math? If not, then it seems unfortunate that this warning triggers for no reason and asks the user to open an issue here.

mkorje:
[...]

PR #8096, opened 2026-04-12
Document that the resulting size may not match the given `{accent,stretch}.size`

YDX-2147483647:
This was discussed many times before. Links can be found in https://github.com/typst/typst/issues/7956#issuecomment-4081957252.

If there is no plan to change this behavior, then I think #7796 can be closed by this PR.

Reference:

- https://learn.microsoft.com/en-us/typography/opentype/spec/math#math-variants
- https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#size-variants

### DIFF
--- a/crates/typst-library/src/math/accent.rs
+++ b/crates/typst-library/src/math/accent.rs
@@ -72,8 +72,25 @@ pub struct AccentElem {
 
     /// The size of the accent, relative to the width of the base.
     ///
-    /// ```example
+    /// ```example:"Basic usage"
     /// $dash(A, size: #150%)$
+    /// ```
+    ///
+    /// Note that the resulting accent may not have the exact desired size. For
+    /// example, an arrow may be either a pre-defined short glyph, or a long
+    /// glyph assembled from building blocks (arrowhead + line) provided by the
+    /// font. The sizes of the two possibilities may not cover the entire span.
+    /// Consequently, arrows of certain intermediate sizes cannot be constructed.
+    ///
+    /// ```example:"Size of arrow growing discontinuously"
+    /// >>> #set par(spacing: 0.3em)
+    /// #for i in range(6) {
+    ///   $ arrow(#box(
+    ///     width: 0.4em + 0.3em * i,
+    ///     fill: aqua,
+    ///     height: 0.4em,
+    ///   )) $
+    /// }
     /// ```
     #[default(Rel::one())]
     pub size: Rel<Length>,
--- a/crates/typst-library/src/math/attach.rs
+++ b/crates/typst-library/src/math/attach.rs
@@ -132,10 +132,9 @@ pub struct StretchElem {
     /// #for size in (
     ///   100%, // short
     ///   101%, 200%, // tall
-    ///   201%, 300%, 400%, 500%, 600%, // grande
+    ///   201%, 300%, 400%, 500%, 600%, // taller
     /// ) {
     ///   $stretch(integral, size: #size)$
-    ///   h(-0.3em, weak: true)
     /// }
     /// ```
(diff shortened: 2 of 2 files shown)


crates/typst-library/src/math/attach.rs
```diff
@@ -119,6 +119,25 @@ pub struct StretchElem {
...
+    /// <<< #set align(center)
+    /// #for size in (
+    ///   100%, // short
+    ///   101%, 200%, // tall
+    ///   201%, 300%, 400%, 500%, 600%, // grande
+    /// ) {
+    ///   $stretch(integral, size: #size)$
+    ///   h(-0.3em, weak: true)
+    /// }
+    /// ```
```
YDX-2147483647, line 139: Image

crates/typst-library/src/math/accent.rs
```diff
@@ -72,9 +72,26 @@ pub struct AccentElem {
...
+    /// ```example:"Size of arrow growing discontinuously"
+    /// >>> #set par(spacing: 0.3em)
+    /// #for i in range(6) {
+    ///   $ arrow(#box(
+    ///     width: 0.4em + 0.3em * i,
+    ///     fill: aqua,
+    ///     height: 0.4em,
+    ///   )) $
+    /// }
+    /// ```
```
YDX-2147483647, line 94: Image

crates/typst-library/src/math/attach.rs
```diff
@@ -119,6 +119,25 @@ pub struct StretchElem {
...
+    /// cover the entire span. In the example below, when the `size` parameter
+    /// is increased from `{101%}` to `{200%}`, the selected glyph remains the
+    /// same, so the actual size does not change.
+    ///
+    /// ```example:"Size of ∫ growing discontinuously"
+    /// >>> #set align(center)
+    /// #for size in (
+    ///   100%, // short
+    ///   101%, 200%, // tall
+    ///   201%, 300%, 400%, 500%, 600%, // grande
```
laurmaedje: is "grande" a specific term here? because the English word would be just "grand", right?

crates/typst-library/src/math/attach.rs
```diff
@@ -119,6 +119,25 @@ pub struct StretchElem {
...
+    /// ) {
+    ///   $stretch(integral, size: #size)$
+    ///   h(-0.3em, weak: true)
```
laurmaedje: I think dropping the negative spacing does not harm the output and would simplify the example a bit

crates/typst-library/src/math/attach.rs
YDX-2147483647: I thought it's a term because physics (the LaTeX package) uses `\grande`. Now I don't think so.

How about short, tall, taller?

laurmaedje:
Thanks! Very nice as well.

YDX-2147483647:
> If there is no plan to change this behavior, then I think #7796 can be closed by this PR.

So, is there any plan about #7796?

crates/typst-library/src/math/attach.rs
MDLC01: Why are two of them slanted? Is this just a feature of the font?

YDX-2147483647: I guess so. Actually the short ∫ is also slanted.

`#show math.equation: set text(stylistic-set: (2,))` can make them upright:

图片

However, turning on ss02 would trigger a warning:
> Glyph has assembly parts with overlap less than minConnectorOverlap

It's similar to #7999, but this time the rendering is indeed problematic.

laurmaedje:
I closed it.

PR #8098, opened 2026-04-12
Replace qcms with moxcms

LaurenzV:
We already have moxcms in the dependency tree and it's also been [adopted by the image crate](https://github.com/image-rs/image/blob/f2a0cbd9ae7e02481928c7d9d04d338578bad4e9/Cargo.toml#L45), so we might as well get rid of qcms now.

Changes are all only small color differences.


laurmaedje:
Thanks!

PR #8108, opened 2026-04-14
Allow negative fractions

fpdotmonkey:
Previously negative fractions were clamped to 0.  This led to unexpected behavior when using negative fraction that didn't align with the intuition that the width of a fraction is proportional to its value divided by the total of all fractions in a row.  This removes the clamping and makes negative fractions well-defined behavior.
    
There was a minor issue in testing where most people use #v for observing the behavior of fr.  It appears that it doesn't support negative widths in general.  Using #h instead lead to visually strange, though correct, output.  So I employed colored boxes which easily show correct behavior.


sijow:
The proposed doc fix doesn't seem to describe the current behavior:

```typ
#set page(width: 5cm)
a #h(1fr)           b\
a #h(1fr)#h(0fr)    b\
a #h(1fr)#h(-1fr)   b\
a #h(1fr)#h(-0.5fr) b\
a #h(1fr)#h(-2fr)   b\
a #h(2fr)#h(-2fr)   b\
```

image

The current behavior seems a bit erratic, but the fact that `h(x*1fr)` and `h(-x*1fr)` appear to cancel out is useful in various cases (for example [here](https://forum.typst.app/t/what-is-v-1fr-why-does-it-work/8534) and [here](https://discord.com/channels/1054443721975922748/1088371919725793360/1501127917743112292)). Another use case:  when a "slide" function is implemented as `v(4fr) + it + v(6fr)` to set an alignment of "40% vertically", and some odd slide should be top aligned, the user can write `v(-4fr)` at the top of the `it` value.

fpdotmonkey:
You know what, I did misunderstand how it works.  It seems there is clipping to 0 as I describe, but that occurs after the fractions are resolved into a ratio (or so, I need to check the source to confirm.

Observe,

```typst
#set page(margin: (left: 33%, right: 33%))
a #h(1fr)           b\
a #h(0fr)           b\
a #h(-1fr)          b\
a #h(1fr)#h(0fr)    b\
a #h(1fr)#h(-1fr)   b\
#box(stroke: red)[a #h(1fr)#h(-0.5fr) #box(stroke: green)[b]]\
#box(stroke: red)[a #h(1fr)#h(-2fr)   #box(stroke: green)[b]]\
a #h(2fr)#h(-2fr)   b\
```

image

If you try to work out the widths of the `h(1fr) h(-0.5fr)` case, it's `h(1fr / (1fr + -0.5fr)) h(-0.5fr / (1fr + -0.5fr))` = `h(200%) h(-100%)`.  We can observe in the image that we're seeing a 200% width, so the -100% is seemingly being discarded.  This also explains why `h(-1fr)` is the same as `h(1fr)`.  This doesn't explain `h(1fr) h(-1fr)` since it would incur a divide by 0, but presumably there is explicit handling for that.

fpdotmonkey:
Converted to draft since this is now incorrect until I fix it.

sijow:
Before you spend too much time on documenting the current behavior in all details, it would probably be best to discuss how it should behave ideally... I doubt that the green boxes being out of the red and pushed in the margin is what we want! Maybe the layout forge on discord is a good place to start the discussion.

laurmaedje:
@fpdotmonkey Do you still plan to work on this? If yes, we can keep it open, but I'm going through older PRs trying to establish which ones may actually still move forward.

fpdotmonkey:
@laurmaedje yes I'll pick this back up

crates/typst-library/src/layout/fr.rs
```diff
@@ -54,7 +56,7 @@ impl Fr {
     pub fn share(self, total: Self, remaining: Abs) -> Abs {
         let ratio = self / total;
         if ratio.is_finite() && remaining.is_finite() {
-            (ratio * remaining).max(Abs::zero())
```
laurmaedje, line 57: This was last changed in the opposite direction in https://github.com/typst/typst/commit/b9c0fd87d304942db315ca6fdbccddf233d045cd, but I cannot recover why I made this change. The old test suite passes even without this, so it may just have been me avoiding negatives there for no particular reason.

crates/typst-library/src/layout/fr.rs
```diff
@@ -11,7 +11,9 @@ use crate::layout::Abs;
 /// Defines how the remaining space in a layout is distributed.
 ///
 /// Each fractionally sized element gets space based on the ratio of its
-/// fraction to the sum of all fractions.
+/// fraction to the sum of all fractions.  If some fractions are negative,
```
laurmaedje, line 14:
```suggestion
/// fraction to the sum of all fractions. If some fractions are negative,
```

laurmaedje (commented):
> There was a minor issue in testing where most people use #v for observing the behavior of fr. It appears that it doesn't support negative widths in general. 

What do you mean by this? Negative `v` definitely works at least sometimes. Here's an example:
```typ
#set block(width: 100%)
#block(fill: blue.transparentize(30%), height: 4cm)
#v(-2cm)
#block(fill: red.transparentize(30%), height: 4cm)
```

Regarding testing, it would be good add tests for all layouters that support fr (basically "Find all references" on `fn share`). That would be `h`, `box(width: _)`, `v`, `block(height: _)`, `grid` rows and columns, and `stack` spacing. The individual tests should be added to the existing tests of the respective layouter.

fpdotmonkey:
> What do you mean by this? Negative v definitely works at least sometimes. Here's an example:

Excuse me, it works for cm (ish; the spacing definitely looks wrong), but not for em.  For example, this listing I would expect would produce something that looks like "1 3 2 4", which it does for cm, but not for em.

```typst
#box[1 #v(2cm) 2 #v(-1cm) 3 #v(2cm) 4]
#box[1 #v(2em) 2 #v(-1em) 3 #v(2em) 4]
```

image

laurmaedje:
It also works with `em` (try setting `#set text(size: 1cm)` and you'll see both being the exact same). It's just that without `weak: true`, `-1em` is not enough to cancel out the default paragraph spacing.

PR #8111, opened 2026-04-14
Fix equal negative scale failing in PNG export

### LINKED ISSUE #7872: Scaling by `-100%` on both axes hides the content in PNG output
### Description

I expect both "hey"s to be visible.

```typ
#rect(scale(x: -100%, y: -101%)[hey])
#rect(scale(x: -100%, y: -100%)[hey])
```

The first hey is visible inside its rectangle. The second one is not: the rectangle is empty.

Seems to only occur with text and in PNG output (not PDF output).

Tested with both 0.14.2 and Feb 5, 2026 development version.

### Operating system

Web app

### Typst version

eltos:
Fixes #7872

`pixglyph::Glyph::rasterize` only supports positive rendering sizes; for negative sizes the alternative render-as-path code path has to be used. While we already checked for `text.size < Abs::zero()`, this check did not take the transform scaling into account. The fix is to instead check for `ppem < 0.0`, since that is the size actually passed to `pixglyph::Glyph::rasterize` in line 124.

_Cases like `scale(x: -100%, y: -101%)[hey]` worked because pixglyph also does not support non-equal transform scalings, and we check for this explicitly (but missed the case of equal negative transform scalings)._


laurmaedje:
Thanks!

PR #8116, opened 2026-04-16
Fix math function tooltips/completions, and add `MathFieldAccess` and `LinkedNode::mode`

### LINKED ISSUE #6312: typst-ide: No autocomplete suggestions in math functions

isuffix:
This PR fixes a number of issues with IDE functionality in Typst and adds two major refactorings to simplify future maintenance of these features.

The major refactorings are the addition of the `MathFieldAccess` syntax kind and the `LinkedNode::mode` method. `MathFieldAccess` is meant to further distinguish math from code in the syntax tree (in the same vein as `Ident` vs. `MathIdent`), which both simplifies the implementation of `LinkedNode::mode` and will be useful for future math syntax refactorings. The implementation of `LinkedNode::mode` supersedes PR #7407, but I've changed the implementation to use a data-oriented approach by having the `LinkedNode` method match on an enum, `ModeSearch,` which is only dependent on the `SyntaxKind`.

However, I only implemented the refactorings _after_ having resolved most of the issues via the prior commits. I did this so that I could better understand where those errors were actually coming from and so we could more clearly see the benefits of the refactorings to code clarity without them also affecting behavior. Although the addition of `LinkedNode::mode` did allow for fixing completions for named and spread arguments in math which I couldn't find a way to resolve beforehand.

- Closes #6312
- Resolves a regression added in #7865 where we weren't calling `trace` on identifiers
- Adds many tooltip and completion tests for function calls in math and code
- Adds `SyntaxKind::MathFieldAccess`
- Adds the method `LinkedNode::mode(&self) -> Option<SyntaxMode>`
- Refactors a bunch of stuff in `typst-ide`

### DIFF
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -42,7 +42,14 @@ impl Eval for ast::FuncCall<'_> {
             } else {
                 (target_expr.eval(vm)?, None)
             };
-            match eval_field_callee(vm, access, target, false)? {
+            match eval_field_callee(
+                vm,
+                access.to_untyped(),
+                field.as_str(),
+                field.span(),
+                target,
+                false,
+            )? {
                 FieldCallee::Func(func) => {
                     let args = match maybe_args {
                         Some(args) => args,
@@ -91,7 +98,7 @@ fn eval_math_call(vm: &mut Vm, math_call: ast::MathCall) -> SourceResult<Value> 
     vm.engine.route.check_call_depth().at(span)?;
 
     let math_call_result = match callee {
-        ast::MathCallee::MathIdent(ident) => {
+        ast::MathAccess::MathIdent(ident) => {
             let callee_value = ident.eval(vm)?;
             // We need to call `trace_at` for the callee manually because we did
             // not evaluate via `ast::Expr::eval()`.
@@ -101,7 +108,7 @@ fn eval_math_call(vm: &mut Vm, math_call: ast::MathCall) -> SourceResult<Value> 
                 Err(err) => FieldCallee::NonFunc(callee_value, err),
             }
         }
-        ast::MathCallee::FieldAccess(access) => {
+        ast::MathAccess::MathFieldAccess(access) => {
             let target_expr = access.target();
             target_span = target_expr.span();
             let field = access.field();
@@ -126,7 +133,14 @@ fn eval_math_call(vm: &mut Vm, math_call: ast::MathCall) -> SourceResult<Value> 
                         math_call.to_untyped().clone().into_text();
                 );
             }
-            eval_field_callee(vm, access, target, true)?
+            eval_field_callee(
+                vm,
+                access.to_untyped(),
+                field.as_str(),
+                field.span(),
+                target,
+                true,
+            )?
         }
     };
 
@@ -225,13 +239,12 @@ enum FieldCallee {
 ///   e.g. `(at: x => ...).at(key)`.
 fn eval_field_callee<'a, 'b>(
     vm: &'a mut Vm<'b>,
-    access: ast::FieldAccess,
+    access: &SyntaxNode,
+    field: &str,
+    field_span: Span,
     target: Value,
     in_math: bool,
 ) -> SourceResult<FieldCallee> {
-    let field_node = access.field();
-    let field_span = field_node.span();
-    let field = field_node.as_str();
     let sink = (&mut vm.engine, field_span);
 
     let mut is_method_call = false;
@@ -251,7 +264,7 @@ fn eval_field_callee<'a, 'b>(
         target.field(field, sink).at(field_span)?
     } else {
         // Otherwise we cannot call this field and produce an error.
-        let full_text = || access.to_untyped().clone().into_text();
+        let full_text = || access.clone().into_text();
         match target.field(field, sink) {
             // The field does exist.
             Ok(callee_value) => {
@@ -479,7 +492,7 @@ impl Eval for ast::MathArgs<'_> {
 fn unparse_math_args(
     vm: &mut Vm,
     args: ast::MathArgs,
-    callee: ast::MathCallee,
+    callee: ast::MathAccess,
 ) -> SourceResult<Content> {
     let mut body = Vec::new();
     let mut errors = EcoVec::new();
@@ -726,6 +739,9 @@ impl<'a> CapturesVisitor<'a> {
 
             // Don't capture the field of a field access.
             Some(ast::Expr::FieldAccess(access)) => {
+                self.visit(access.target().to_untyped());
+            }
+            Some(ast::Expr::MathFieldAccess(access)) => {
                 self.visit(access.target().to_untyped());
             }
 
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -313,9 +313,7 @@ fn eval_field_callee<'a, 'b>(
         }
     };
 
-    if vm.inspected == Some(access.span()) {
-        vm.trace(callee_value.clone());
-    }
+    vm.trace_at(access.span(), &callee_value);
 
     match callee_value.clone().cast::<Func>() {
         Ok(func) if is_method_call => Ok(FieldCallee::Method(func, target)),
--- a/crates/typst-eval/src/call.rs
+++ b/crates/typst-eval/src/call.rs
@@ -93,6 +93,9 @@ fn eval_math_call(vm: &mut Vm, math_call: ast::MathCall) -> SourceResult<Value> 
     let math_call_result = match callee {
         ast::MathCallee::MathIdent(ident) => {
             let callee_value = ident.eval(vm)?;
+            // We need to call `trace_at` for the callee manually because we did
+            // not evaluate via `ast::Expr::eval()`.
+            vm.trace_at(ident.span(), &callee_value);
             match callee_value.clone().cast::<Func>() {
                 Ok(func) => FieldCallee::Func(func),
                 Err(err) => FieldCallee::NonFunc(callee_value, err),
--- a/crates/typst-eval/src/vm.rs
+++ b/crates/typst-eval/src/vm.rs
@@ -20,7 +20,8 @@ pub struct Vm<'a> {
     pub flow: Option<FlowEvent>,
     /// The stack of scopes.
     pub scopes: Scopes<'a>,
-    /// A span that is currently under inspection.
+    /// A span that is currently under inspection. If this is `Some`, we're in
+    /// tracing mode, and will record every value the given span sees.
     pub inspected: Option<Span>,
     /// Data that is contextually made accessible to code behind the scenes.
     pub context: Tracked<'a, Context<'a>>,
@@ -55,9 +56,7 @@ impl<'a> Vm<'a> {
     /// This will insert the value into the top-most scope and make it available
     /// for dynamic tracing, assisting IDE functionality.
     pub fn bind(&mut self, var: ast::Ident, binding: Binding) {
-        if self.inspected == Some(var.span()) {
-            self.trace(binding.read().clone());
-        }
+        self.trace_at(var.span(), binding.read());
 
         // This will become an error in the parser if `is` becomes a keyword.
         if var.get() == "is" {
@@ -73,12 +72,22 @@ impl<'a> Vm<'a> {
         self.scopes.top.bind(var.get().clone(), binding);
     }
 
-    /// Trace a value.
+    /// Helper to only call [`Self::trace`] for a value if we're inspecting its
+    /// span. This method (or `trace`) should be called for every value produced
+    /// by an expression.
+    pub fn trace_at(&mut self, span: Span, value: &Value) {
+        if self.inspected == Some(span) {
+            self.trace(value.clone());
+        }
+    }
+
+    /// Trace a value. Tracing powers IDE tooltips and hover info. This method
+    /// should be called for every value produced by an expression.
     #[cold]
     pub fn trace(&mut self, value: Value) {
         self.engine
             .sink
-            .value(value.clone(), self.context.styles().ok().map(|s| s.to_map()));
+            .value(value, self.context.styles().ok().map(|s| s.to_map()));
     }
 }
 
--- a/crates/typst-ide/src/complete.rs
+++ b/crates/typst-ide/src/complete.rs
@@ -300,6 +300,8 @@ fn complete_math(ctx: &mut CompletionContext) -> bool {
         ctx.leaf.parent_kind(),
         Some(SyntaxKind::Equation)
             | Some(SyntaxKind::Math)
+            | Some(SyntaxKind::MathCall)
+            | Some(SyntaxKind::MathArgs)
             | Some(SyntaxKind::MathFrac)
             | Some(SyntaxKind::MathAttach)
     ) {
@@ -321,10 +323,7 @@ fn complete_math(ctx: &mut CompletionContext) -> bool {
     }
 
     // Behind existing atom or identifier: "$a|$" or "$abc|$".
-    if matches!(
-        ctx.leaf.kind(),
-        SyntaxKind::Text | SyntaxKind::MathText | SyntaxKind::MathIdent
-    ) {
+    if matches!(ctx.leaf.kind(), SyntaxKind::MathText | SyntaxKind::MathIdent) {
         ctx.from = ctx.leaf.offset();
         math_completions(ctx);
         return true;
@@ -663,6 +662,11 @@ fn show_rule_recipe_completions(ctx: &mut CompletionContext) {
 }
 
 /// Complete call and set rule parameters.
+///
+/// FUTURE: Make this work for math functions. This will require a much deeper
+/// refactoring of `param_completions` below, including handling 2d arguments
+/// correctly and ensuring we add a hash in math when suggesting to insert
+/// values that aren't strings.
 fn complete_params(ctx: &mut CompletionContext) -> bool {
     // Ensure that we are in a function call or set rule's argument list.
     let (callee, set, args, args_linked) = if let Some(parent) = ctx.leaf.parent()
@@ -868,16 +872,16 @@ fn path_completion(func: &Func, param: &ParamInfo) -> Option<&'static [&'static 
 
 /// Complete in code mode.
 fn complete_code(ctx: &mut CompletionContext) -> bool {
-    if matches!(
+    debug_assert!(!matches!(
         ctx.leaf.parent_kind(),
         None | Some(SyntaxKind::Markup)
             | Some(SyntaxKind::Math)
+            | Some(SyntaxKind::MathCall)
+            | Some(SyntaxKind::MathArgs)
             | Some(SyntaxKind::MathFrac)
             | Some(SyntaxKind::MathAttach)
             | Some(SyntaxKind::MathRoot)
-    ) {
-        return false;
-    }
+    ));
 
     // An existing identifier: "{ pa| }".
     // Ignores named pair keys as they are not variables (as in "(pa|: 23)").
@@ -1680,7 +1684,10 @@ mod tests {
 
     #[test]
     fn test_autocomplete_hash_expr() {
+        test("#", -1).must_include(["int", "if conditional"]);
         test("#i", -1).must_include(["int", "if conditional"]);
+        test("$#$", -2).must_include(["int", "if conditional"]);
+        test("$#i$", -2).must_include(["int", "if conditional"]);
     }
 
     #[test]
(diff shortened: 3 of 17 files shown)


crates/typst-syntax/src/kind.rs
```diff
@@ -520,3 +525,295 @@ impl SyntaxKind {
...
+            SyntaxKind::Strong => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::Emph => ModeSearch::Known(SyntaxMode::Markup),
+            // TODO: Should we treat raw as 'code | markup'?
+            SyntaxKind::Raw => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::RawLang => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::RawDelim => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::RawTrimmed => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::Link => ModeSearch::Known(SyntaxMode::Markup),
+            // TODO: Should we treat labels as 'code | markup'?
+            SyntaxKind::Label => ModeSearch::Known(SyntaxMode::Markup),
```
isuffix, line 579: Both raw text and labels can be included directly in code, but #7407 treated them as markup only, so I wasn't sure if we should change them.

crates/typst-eval/src/vm.rs
```diff
@@ -20,7 +20,8 @@ pub struct Vm<'a> {
     pub flow: Option<FlowEvent>,
     /// The stack of scopes.
     pub scopes: Scopes<'a>,
-    /// A span that is currently under inspection.
+    /// A span that is currently under inspection. If this is `Some`, we're in
+    /// tracing mode, and will record every value  the given span sees.
```
laurmaedje: `/value  the/value the/`

crates/typst-ide/src/tooltip.rs
```diff
@@ -58,7 +58,9 @@ fn expr_tooltip(world: &dyn IdeWorld, leaf: &LinkedNode) -> Option<Tooltip> {
     }
 
     let expr = ancestor.cast::<ast::Expr>()?;
-    if !expr.hash() && !matches!(expr, ast::Expr::MathIdent(_)) {
+    // We only analyze embeddable code expressions or math expressions that
+    // access variables.
+    if !expr.hash() && !matches!(expr, ast::Expr::MathIdent(_) | ast::Expr::MathCall(_)) {
         return None;
     }
```
laurmaedje, line 72: I'd suggest this. This would make it simpler to parse mentally imo:
```rust
let analyze =
    expr.hash() || matches!(expr, ast::Expr::MathIdent(_) | ast::Expr::MathCall(_));
if !analyze {
    return None;
}
```

crates/typst-ide/src/complete.rs
```diff
@@ -663,6 +662,11 @@ fn show_rule_recipe_completions(ctx: &mut CompletionContext) {
 }
 
 /// Complete call and set rule parameters.
+///
+/// FUTURE: Make this work math functions. This will require a much deeper
```
laurmaedje: `/work math/work for math/`

crates/typst-eval/src/call.rs
```diff
@@ -101,11 +108,17 @@ fn eval_math_call(vm: &mut Vm, math_call: ast::MathCall) -> SourceResult<Value>
                 Err(err) => FieldCallee::NonFunc(callee_value, err),
             }
         }
-        ast::MathCallee::FieldAccess(access) => {
+        ast::MathAccess::MathFieldAccess(access) => {
             let target_expr = access.target();
             target_span = target_expr.span();
             let field = access.field();
-            let target = target_expr.eval(vm)?;
+            let target = match target_expr {
```
laurmaedje: This seems a bit duplicate with `Eval for ast::MathFieldAccess<'_>`. What do you think about this change: https://github.com/typst/typst/commit/0096e3c158d35fdde7bb24b7c5898ea460476a79.

Or does it not make sense with planned changes in https://github.com/typst/typst/pull/7866?

(If you think it makes sense, feel free to dissolve this change into your commit structure, no need to keep my commit.)

crates/typst-ide/src/complete.rs
```diff
@@ -51,15 +51,20 @@ pub fn autocomplete(
...
+            || complete_rules(&mut ctx)
+            || complete_params(&mut ctx))
+    {
+        // Only attempt the general completions after the more specific ones.
+        let _ = match mode {
+            SyntaxMode::Markup => complete_markup(&mut ctx),
+            SyntaxMode::Math => complete_math(&mut ctx),
+            SyntaxMode::Code => complete_code(&mut ctx),
+        };
+    }
```
laurmaedje: how do you feel about short-circuiting with `?`

I'd propose this variant:
```rust
// Already ensures we aren't in a comment.
let mode = ctx.leaf.mode()?;

_ = complete_field_accesses(&mut ctx)
    || complete_open_labels(&mut ctx)
    || complete_imports(&mut ctx)
    || complete_rules(&mut ctx)
    || complete_params(&mut ctx)
    // Only attempt the general completions after the more specific ones.
    || match mode {
        SyntaxMode::Markup => complete_markup(&mut ctx),
        SyntaxMode::Math => complete_math(&mut ctx),
        SyntaxMode::Code => complete_code(&mut ctx),
    };
```

crates/typst-ide/src/complete.rs
```diff
@@ -296,33 +276,7 @@ fn markup_completions(ctx: &mut CompletionContext) {
...
-    }
-
-    // Behind existing interpolated identifier: "$#pa|$".
-    if ctx.leaf.kind() == SyntaxKind::Ident {
-        ctx.from = ctx.leaf.offset();
-        code_completions(ctx, true);
-        return true;
-    }
-
+    debug_assert_eq!(ctx.leaf.mode(), Some(SyntaxMode::Math));
```
laurmaedje: I'd add an extra blank after this

crates/typst-ide/src/complete.rs
```diff
@@ -898,13 +850,6 @@ fn complete_code(ctx: &mut CompletionContext) -> bool {
         return true;
     }
 
-    // A potential label (only at the start of an argument list): "(<|".
-    if ctx.before.ends_with("(<") {
-        ctx.from = ctx.cursor;
-        ctx.label_completions();
-        return true;
-    }
```
laurmaedje, line 897: do we have a test for this working even without the branch here?

crates/typst-syntax/src/kind.rs
```diff
@@ -523,3 +525,298 @@ impl SyntaxKind {
...
+            SyntaxKind::Heading => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::HeadingMarker => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::ListItem => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::ListMarker => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::EnumItem => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::EnumMarker => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::TermItem => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::TermMarker => ModeSearch::Known(SyntaxMode::Markup),
+
+            SyntaxKind::Equation => ModeSearch::Known(SyntaxMode::Math),
```
laurmaedje, line 591: I'm not sure this change (and the corresponding whitespace change above) makes sense. See https://github.com/typst/typst/pull/7407#discussion_r2576550495.

crates/typst-syntax/src/kind.rs
```diff
@@ -523,3 +525,298 @@ impl SyntaxKind {
...
+            SyntaxKind::MathAlignPoint => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathCall => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathArgs => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathDelimited => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathAttach => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathPrimes => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathFrac => ModeSearch::Known(SyntaxMode::Math),
+            SyntaxKind::MathRoot => ModeSearch::Known(SyntaxMode::Math),
+
+            SyntaxKind::Hash => ModeSearch::Known(SyntaxMode::Code),
```
laurmaedje, line 606: Similarly to the equation, I'm not sure this makes sense since `eval("#", mode: "code")` is an error. In particular, you marked `LeftBracket` as `Parent`. Is it not the exact opposite of `#` in its effect?

crates/typst-syntax/src/kind.rs
```diff
@@ -523,3 +525,298 @@ impl SyntaxKind {
...
+            SyntaxKind::SmartQuote => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::Strong => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::Emph => ModeSearch::Known(SyntaxMode::Markup),
+            // TODO: Should we treat raw as 'code | markup'?
+            SyntaxKind::Raw => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::RawLang => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::RawDelim => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::RawTrimmed => ModeSearch::Known(SyntaxMode::Markup),
+            SyntaxKind::Link => ModeSearch::Known(SyntaxMode::Markup),
+            // TODO: Should we treat labels as 'code | markup'?
```
laurmaedje, line 578: If we treat what mode it correctly `eval`s as our source of truth, then both raw and label should be parent indeed.

crates/typst-syntax/src/kind.rs
```diff
@@ -523,3 +525,298 @@ impl SyntaxKind {
...
+            SyntaxKind::LeftParen => ModeSearch::Parent, // code | math
+            SyntaxKind::RightParen => ModeSearch::Parent, // code | math
+            SyntaxKind::Comma => ModeSearch::Parent,     // code | math
+            // TODO: Semicolon after code is also embedded like a hash, so this
+            // is incorrect...
+            SyntaxKind::Semicolon => ModeSearch::Parent, // code | math
+            SyntaxKind::Colon => ModeSearch::Parent,     // code | math | markup
+            SyntaxKind::Star => ModeSearch::Parent,      // code | markup
+            SyntaxKind::Underscore => ModeSearch::Parent, // code | math
+            SyntaxKind::Dollar => ModeSearch::Known(SyntaxMode::Math),
```
laurmaedje, line 620: this would potentially also be parent then.

crates/typst-syntax/src/node.rs
```diff
@@ -751,6 +752,42 @@ impl<'a> LinkedNode<'a> {
...
+
+    /// Get the [`SyntaxMode`] of this node's parent.
+    ///
+    /// Unlike some other `LinkedNode` methods, this does not treat all trivia
+    /// the same: it returns `None` for comments and returns `Some` for
+    /// whitespace (based on the parent's mode). The only other way this would
+    /// return `None` is when inside a partial tree, i.e. one not rooted in
+    /// `Markup`, `Math`, or `Code`.
+    ///
+    /// Also note that errors inherit the mode of their parent.
```
laurmaedje: I don't think we need to copy the comment here.

laurmaedje (commented):
Nice work! Just a few comments. 

On the mode suggestions for a couple of the "boundary" syntax nodes, I'm not entirely certain, but I feel like it's not quite consistent as is.

crates/typst-syntax/src/kind.rs
```diff
@@ -520,3 +525,298 @@ impl SyntaxKind {
...
+            SyntaxKind::LeftBrace => ModeSearch::Parent, // code | markup
+            SyntaxKind::RightBrace => ModeSearch::Parent, // code | markup
+            SyntaxKind::LeftBracket => ModeSearch::Parent, // code | markup
+            SyntaxKind::RightBracket => ModeSearch::Parent, // code | markup
+            SyntaxKind::LeftParen => ModeSearch::Parent, // code | math
+            SyntaxKind::RightParen => ModeSearch::Parent, // code | math
+            SyntaxKind::Comma => ModeSearch::Parent,     // code | math
+            // TODO: Semicolon after code is also embedded like a hash, so this
+            // is incorrect...
+            SyntaxKind::Semicolon => ModeSearch::Parent, // code | math
```
isuffix, line 616: I added this comment, but I'm now actually unsure if it's correct. If we want to treat a hashtag as code, we must be doing so because we're targeting the position after the hash (`#|`), since the position before the hash (`|#`) would have to be math or markup. If we follow that rule, the position after an embedded semicolon, like `[#123;|]` would again be the mode of its parent. So the current definition should be fine and just needs to be annotated as `code | math | markup`. LMK if that makes sense or not.

crates/typst-eval/src/call.rs
isuffix: That's very nice, I'll include it! I think I had a blind spot here due to considering adding mutable methods to math in the future, but that wouldn't actually conflict here.

crates/typst-ide/src/complete.rs
isuffix: Looks like no? I can't come up with a way in which this would run without first matching to `complete_open_labels`. Both were initially added by you in this commit: https://github.com/typst/typst/commit/5a6cadefda9a65a83e678e0d541a89bb672962ee, but that didn't add any tests. The `test_autocomplete_ref_function` test was added much later (in #6586), and while it could be matched by this code, `complete_open_labels` gets there first.

isuffix:
I think I've resolved all the non-mode related comments. I also pushed a refactor that didn't change any behavior, but adds `use ModeAfter::*;` and `use SyntaxMode::{...};` statements to `mode_search` to make it easier to read/edit.

For the various mode comments, I think what we really want out of the method is a way to determine the `SyntaxMode` that we would be in immediately after a given node. To me this is justified by the idea that returning anything but `Code` for `#` would be ludicrous; like what would be the use of such a method? I did consider that we could add a `Side` parameter, or maybe define `mode` in terms of a cursor position instead of on the `LinkedNode`, but those all feel a little cumbersome for helping with completions.

This would also resolve what to return when something is embedded with a hash, like `#ident`: we return `Code` because immediately after `ident` you can write `.field` and stay in code mode, which is exactly the completion we want to offer. And while we could make a new distinction between atomic/non-atomic expressions, I don't think it would be worth it.

To clarify this, how would you feel about renaming `LinkedNode::mode` to `LinkedNode::mode_after`? And then `ModeSearch` could be `ModeAfter` and `SytnaxKind::mode_search` could be `mode_after`, although I feel like `parent_mode` could just stay the same.

Finally, for `Label`/`Raw`/`Equation`/`Ident` (and also `Str`), I think we're just missing a new enum variant for `Embeddable` syntax kinds. Those kinds can all exist in either markup or math on their own, but are also available directly in code, or embedded as code into markup/math with a hashtag. And I think the resolution for them is to say that they inherit from their parent _unless_ they're embedded into markup/math with a hashtag. Then with that addition,  for dollar signs we can start to distinguish opening dollar signs from closing ones, where opening dollars will return `Math` (because after them we're in math mode) and closing dollars will return the mode of the overall `Equation`.

laurmaedje:
I think what you say makes sense. I was only thinking about the modes in the abstract, not with their concrete use for completions in mind. I'm on board with the changes.

One thing I don't fully get yet is how `Ident` relates to `Label`/`Raw`/`Equation`. Isn't `Ident` like any other expression? It requires a hash to be in markup, while label/raw/equation don't. In math, idents don't but that would be `MathIdent` then.

isuffix:
`Ident` differs from `Label`/`Raw`/`Equation`/`Str` in that it can't be a _top-level_ expression in markup or math, but unlike say, `Float`, `Ident` is available outside of code mode when in a `Named` kind under `MathArgs`. This makes it somewhat similar to how `Colon` is mainly code-mode, but can be part of markup under `TermItem`. However `Ident` can't just use `Parent` because it can still be embedded with a hashtag like the other embeddable kinds.

isuffix:
Updated with those changes. I also realized that there were issues with a few other kinds like `ContentBlock` (mainly code, but can be markup under `Ref`) and `Space` (needs special handling due to how we use it at the edges of `Equation`), so I updated those and added tests for everything I could think of.

Also, I expanded the `code | math | markup` comments to include the parent kind which produces each other mode. There's a large explanation comment at the top, but LMK it's perhaps too much info or would be too hard to keep up-to-date.

crates/typst-syntax/src/kind.rs
```diff
@@ -520,3 +525,315 @@ impl SyntaxKind {
...
+            Self::Space => Space,      // code/math: part | markup: expr/part
+            Self::Linebreak => Parent, // math/markup: expr
+            Self::Parbreak => Known(Markup),
+            Self::Escape => Parent, // math/markup: expr
+            Self::Shorthand => Known(Markup),
+            Self::SmartQuote => Known(Markup),
+            Self::Strong => Known(Markup),
+            Self::Emph => Known(Markup),
+            Self::Raw => Embeddable,    // code/markup: expr
+            Self::RawLang => Parent,    // code/markup: Raw
```
laurmaedje: After the `RawLang`, are we really in code or markup? We're still in the raw block after all. So perhaps we could handle this like we handle comments, by returning `None` from `LinkedNode::mode_after.`

The other raw parts are also a bit tricky.

crates/typst-syntax/src/kind.rs
```diff
@@ -520,3 +525,315 @@ impl SyntaxKind {
...
+        // kinds would be too long (otherwise, the list of kinds should be
+        // exhaustive). `expr` is a kind which can be a top-level expression in
+        // its mode, and `part` is a kind which is just a part of an expression.
+        match self {
+            Self::End => Comment,
+            Self::Error => Parent, // code/math/markup
+
+            Self::Shebang => Comment,
+            Self::LineComment => Comment,
+            Self::BlockComment => Comment,
```
laurmaedje: Aren't we, in principle, back in the parent mode right after a block comment? Though changing this would also have an effect on the autocomplete changes.

crates/typst-ide/src/complete.rs
```diff
@@ -51,15 +51,20 @@ pub fn autocomplete(
         explicit,
     )?;
 
-    let _ = complete_comments(&mut ctx)
-        || complete_field_accesses(&mut ctx)
+    // Getting the syntax mode also ensures we are not in a comment.
+    let mode = ctx.leaf.mode_after()?;
```
laurmaedje, line 55: I wonder if the "mode after" framing will cause any issues when attempting to trigger completions at a position that is not on a node boundary.

crates/typst-syntax/src/kind.rs
isuffix: I think returning `None` makes sense for the internal kinds. For now, I'll switch to doing so except for the overall `Raw` kind and the closing raw delimiter.

However, you should also know that tinymist has a very similar construct as an `InterpretMode` enum with variants for `Code`/`Math`/`Markup` _and_ `Raw`/`String`/`Comment` (defined in `tinymist-analysis/src/syntax/matcher.rs:316`). If you want, we could try to unify with them by outputting a similar enum, but their code would probably benefit from `mode_after` regardless.

Also note that `Raw` does trigger some special completions for tinymist, but I can't find an instance of them using `String`/`Comment` to produce any particular kind of completion.

crates/typst-syntax/src/kind.rs
isuffix: I guess we kind of are, but I don't think we can do any better without adding a representation for the delimiter of a comment or changing the whole `mode_after` interface to use cursors instead of whole nodes. So I think it's probably best to just not do anything for this :/

crates/typst-ide/src/complete.rs
isuffix: I think there are really only a few edge cases like raw/strings/comments which would be problematic and could be handled by special completions before accessing `mode_after`.

isuffix: But since we don't have any of those completions ourselves, I think it's fine for now?

crates/typst-syntax/src/kind.rs
laurmaedje: I don't quite follow. What would be wrong about setting `BlockComment` to `Parent`?

isuffix: Nothing wrong with it, it would just affect autocompletion if we did.

isuffix:
Rebased off main and updated to return `None` for the bodies of raw text. I also reverted one of the comments in the first commit that would conflict with #8188

laurmaedje:
Thank you!

PR #8125, opened 2026-04-16
More examples on how to use stops within gradients

### LINKED ISSUE #3109: Improve gradient docs
### Description

image

there are 2 problems:

1. the stops arg doesnt have a name in the signature. It takes some time for me to figure out
2. the stops accept `array` but the discription of format is missing. Proving more examples should be better.

ctx: I'm trying to do this: https://typst.app/project/r0ylIFbUVdjsbK8498-qVT

image

```typst
#let grad = gradient.linear((black, 0%),(black, 50%),(white, 50%),(white, 75%), (black, 75%),(black, 100%), dir: ttb)
#rect(fill: grad, width: 100%, height: 4pt)
```

### Reproduction URL

_No response_

### Operating system

### Typst version

tomeichlersmith:
Instead of adding one sloppy (pun intended) example under the "stops" parameter description, I think it makes more sense to (a) explain under the "stops" parameter that it is our name for the (unnamed) positional arguments and (b) add more examples with different manually-defined "stops" in the section about stops.

I added three examples showing my own personal progression in understanding. First realizing that I can just provide my own list of colors (instead of a pre-defined color map). Then realizing that I could shift the boundaries of the colors by using manually-defined stops. Finally, doing something slightly more complicated by preventing a gradient except within a small transition region of a progress bar.

This resolves #3109

### DIFF
--- a/crates/typst-library/src/visualize/gradient.rs
+++ b/crates/typst-library/src/visualize/gradient.rs
@@ -71,18 +71,18 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
 /// stops evenly which is a reasonable default in order to provide a smooth
 /// gradient.
 ///
-/// ```example:"Smooth Gradient between Red and Blue"
+/// ```example
 /// #rect(
-///   width: 100%, 
+///   width: 100%,
 ///   height: 20pt,
 ///   fill: gradient.linear(red, blue)
 /// )
 /// ```
 ///
-/// We can use manually-defined stops in order to end the gradient "early"
+/// We can use manually-defined stops in order to end the gradient early
 /// rather than in a smooth way.
 ///
-/// ```example:"Smooth Gradient with Manually-Defined Stops"
+/// ```example
 /// #block(
 ///   width: 100%,
 ///   inset: 4pt,
@@ -104,11 +104,11 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
 /// ```
 ///
 /// We can go even further.
-/// Sometimes it is preferable to have a "gradient" that is not smooth
-/// and instead "blocky" or "stepwise". For example, you can display a progress
+/// Sometimes it is preferable to have a gradient that is not smooth
+/// and instead blocky or stepwise. For example, you can display a progress
 /// bar using a simple linear gradient with manually defined stops.
-/// 
-/// ```example:"Simple Progress Bar"
+///
+/// ```example
 /// #block(
 ///   width: 100%,
 ///   inset: 4pt,
@@ -119,7 +119,7 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
 /// )[We're 2/3 done!]
 /// ```
 ///
-/// You could also use [`stack`]($stack) (with `dir: ltr`) to make a progress
+/// You could also use [`stack`]($stack) (with `{dir: ltr}`) to make a progress
 /// bar (which is probably more sensible), but this at least helps display how
 /// to write custom stops.
 ///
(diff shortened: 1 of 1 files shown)


crates/typst-library/src/visualize/gradient.rs
```diff
@@ -67,7 +67,62 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
...
+///
+/// ```example
+/// #rect(
+///   width: 100%,
+///   height: 20pt,
+///   fill: gradient.linear(red, blue)
+/// )
+/// ```
+///
+/// We can use manually-defined stops in order to end the gradient early
```
laurmaedje, line 81: This phrasing sounds a bit like ending early is the primary point of manually-defined stops though in reality it's just one small use case.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -67,7 +67,62 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
...
+/// ```example
+/// #block(
+///   width: 100%,
+///   inset: 4pt,
+///   fill: gradient.linear(
+///     (red, 0%),
+///     (blue, 50%),
+///     (blue, 100%)
+///   ),
+///   align(
```
laurmaedje, line 93: The content with multiple alignments distracts a bit from the point of the example in my opinion.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -67,7 +67,62 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
...
+///       align(
+///         left
+///       )[I'm already fully blue]
+///     )
+///   )
+/// )
+/// ```
+///
+/// We can go even further.
+/// Sometimes it is preferable to have a gradient that is not smooth
```
laurmaedje, line 106: For a stepwise gradient, the more idiomatic approach would in principle be `.sharp(..)`.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -67,7 +67,62 @@ use crate::visualize::{Color, ColorSpace, WeightedColor};
...
+///     (gray, 100%*2/3), (gray, 100%)
+///   ),
+/// )[We're 2/3 done!]
+/// ```
+///
+/// You could also use @stack[`stack`] (with `dir: ltr`) to make a progress
+/// bar (which is probably more sensible due to the
+/// @gradient:note-on-file-sizes[file size implications of using `gradient`]),
+/// but this at least helps display how
+/// to write custom stops.
```
laurmaedje, line 125: This justification for having an example feels out of place. If we think a stack is better for a progress bar than we should try to find a different example for gradients.

crates/typst-library/src/visualize/gradient.rs
```diff
@@ -213,7 +268,7 @@ impl Gradient {
     pub fn linear(
         args: &mut Args,
         span: Span,
-        /// The color @gradient:stops[stops] of the gradient.
+        /// The unnamed (positional) arguments are the color @gradient:stops[stops] of the gradient.
```
laurmaedje, line 271: The phrasing here feels a bit out of place as generally the first docs line always specifies just _what_ the argument is instead of being a free-standing sentence. I think the previous sentence was fine and reasonable to understand given examples.

To resolve https://github.com/typst/typst/issues/3109, the per-stop array format should be explained here. It suffices to do it once and then, the other constructors can link here.

laurmaedje (commented):
https://github.com/typst/typst/pull/8645 addresses the problem a bit better in my opinion, so I have decided to merge that one and close this. Thanks still!

PS: Since I was already mostly done reviewing this before realizing that the other PR exists, I've left my comments anyway.

PR #8127, opened 2026-04-16
Add unit spacing hint for args, arrays, and dicts

### LINKED ISSUE #1287: More useful error message for lengths with space
Currently
```
#box(stroke: 1 pt, [a #sym.zws]) 
```
yields the not very helpful error message
```
Error: expected comma
   ╭─[<user input>:8:15]
   │
 8 │ #box(stroke: 1 pt, [a #sym.zws])
   │                 
   │                  
───╯
```
It would be helpful if the error message could tell me something like "Did you mean 1pt?". For an end user, the error message is confusing, because there is a comma right after.

GabrielDTB:
Closes https://github.com/typst/typst/issues/1287


GabrielDTB:
Hey, I've been working on a few other small issues, so I want to make sure my approach on this PR one wasn't way off base as I prepare to PR those other changes. Any comments appreciated if anyone has the time :)

crates/typst-syntax/src/parser.rs
```diff
@@ -2033,6 +2043,60 @@ impl Parser<'_> {
...
+    /// If the current token is a unit name (e.g. `pt`) and the node before
+    /// the most recent error is a number, hint that they should be combined.
+    fn hint_unit_spacing(&mut self) {
+        if !self.at(SyntaxKind::Ident) {
+            return;
+        }
+        let unit = match self.current_text() {
+            "pt" | "mm" | "cm" | "deg" | "rad" | "em" | "fr" => self.current_text(),
+            _ => return,
+        };
```
isuffix: This match-statement is duplicated from similar code in the lexer, which we generally want to avoid. Probably the best way to factor it out would be to add a `FromStr` impl to `Unit` in `ast.rs` (with an `Err` type of `()`). Could you add that as a separate commit and change this `match` and the one in the lexer to use `.parse::<ast::Unit>()` instead?

isuffix (commented):
Thank you for contributing this! Apologies for the wait on review.

I think that the main changes could be a bit better, and I'll try to guide you through what I think should could be improved. But know that I appreciate your work here, especially adding tests! It seems like you've really considered this, so forgive me I ever come off as unappreciative :)

There's a major simplification we can make to the organization of the hinting code. Since we always parse commas via `p.expect(SyntaxKind::Comma)`, I think it's ok for us to just resolve the error automatically in `Parser::expect` by adding a branch like `else if kind == SyntaxKind::Comma && <check-for-unit-spacing> { ... }`. This avoids duplicating the `if p.hint_percent_unit_spacing()` statements in `args` and `parenthesized_or_array_or_dict`. It also lets us recognize the problematic case _before_ we've actually added the error message, so we don't have to skip past it when searching for a prior integer/float, and would make it so the numeric suffix (both units and the error for a percent) is directly in the parser's `token` field.

As for the `<check-for-unit-spacing>` part, I think that would make sense as a `Parser` method taking `&self` and returning `Option<&str>` populated with the number's text only if we recognize that the current `self.token.node` is a numeric suffix (unit or percent) that comes after a number (similar to `preceding_number` plus the start of `hint_unit_spacing`). I think the creation of the hint itself likely makes more sense inside `expect` after checking this function, so the new branch would look like: `else if kind == SyntaxKind::Comma && let Some(number) = self.check_unit_spacing() { <add-error-and-hint> }`.

Finally, I think the error messages **"expected comma"** and **"the character `%` is not valid in code"** ought to be placed on the suffix itself and worded as **"expected comma, but found numeric suffix"**. So I would fill in the `<add-error-and-hint>` part with `self.token.node = SyntaxNode::Error(SyntaxError("<message>"), self.current_text())` and then `self.token.node.hint("<hint-message>")`. Although note that because `number` will be a `&` based on `&self`, we'll need to build the hint message with `eco_format!` in a separate statement before we can assign to `self.token.node`.

GabrielDTB:
Thank you for the code review! It's definitely a lot cleaner this way than before.

I have a question on the approach: By attaching the error to the unit, we get better error message squiggles (hey, look at the unit, this is the important bit). But there's not really an error on the unit, right? (well, except for percent, but for different reasons). `pt` is fine on its own. I could imagine this causing problems somehow, like if you inspected the parse tree backwards (sort of like how we're doing here). Does this matter?

Also, I wasn't sure on was what to do with the `%` case. Normally, `%` on its own is just an error, and here we're replace that error with the `expected comma, but found numeric suffix` error. If we return up without eating it, then it seems like we could cause the same issue that prompted this handling in `expected`:

```rust
    fn expected(&mut self, thing: &str) {
        if self.token.kind.is_error() {
            // If we encounter an erroneous token when something was expected,
            // we need to actually consume the token. If we don't, and we then
            // proceed to exit our current lexing mode, future incremental
            // reparsing could fail to lex the token in the correct mode.
            //
            // Example: When parsing an unclosed string in `#import "str`,
            // we need to make sure the string is lexed as code, because if
            // it were lexed as markup, a future insert of a closing quote
            // would only be adjacent to markup text, and wouldn't lex as a
            // string when reparsing, causing a difference between the full
            // parse and the incremental parse.
            self.trim_errors();
            self.eat();
```

So, I figured we should at least eat this error. If we don't eat the following comma we would get a spurious `unexpected comma` error, which isn't present for the other units, so I eat that too.

crates/typst-syntax/src/parser.rs
```diff
@@ -1973,13 +1973,49 @@ impl Parser<'_> {
...
+                self.current_text(),
+            );
+            self.token.node.hint(hint);
+            // Match `expected`'s handling of errors, but also eat the following
+            // comma to suppress a spurious `unexpected comma` error.
+            if self.token.kind.is_error() {
+                self.trim_errors();
+                self.eat();
+                self.eat_if(SyntaxKind::Comma);
+            }
```
isuffix: Here, we should make sure that `token.kind` doesn't get out of sync with `token.node.kind()`, so I think we should either not eat anything and assign `self.token.kind = SyntaxKind::Error;` or we should eat the error in all cases. The latter feels preferrable to me.

isuffix (changes requested):
I agree that the error is due to the space between the unit and the number lacking a comma, but the point of this change is to produce a better message for the likely cause of the error, even if doing so interprets the cause incorrectly for some cases. I think using `pt` or another suffix as a variable itself and missing a comma after a number is reasonably unlikely, so it's fine if the error message is more confusing as long as it's roughly understandable.

For percent signs, turning them into an error token should never cause the issue mentioned by that comment since we only turn them into errors when we expected a comma, which we would never expect if we could be at a lexing mode border, although this reasoning is a bit involved. But I think eating the error plus comma is still fine. IMO the code around parsing parenthesized items is a bit fragile in general (I haven't really ever tried to refactor the main loops), so I don't think this change particularly affects the code quality. And eating the error ensures that we do always produce an error. If we were concerned with code clarity, we could make it a separate parser method, but I think it's fine this way.

We could try to reduce false positives by checking that the space between the number and suffix doesn't span a newline, since if you write `(1\n%)`, the two tokens are getting reasonably distant. This should just require adding `&& !self.had_newline()` to the if-statement. I think I'm in favor of this, but it's not strictly necessary. Although regardless, we should add something like `(1\n%)` as a test case.

isuffix:
To be clear, I think the only changes this needs are to add `&& !self.had_newline()` to the if-statement in `Parser::expect`, and to always call `self.trim_errors(); self.eat(); self.eat_if(SyntaxKind::Comma);` at the end of the block.

crates/typst-syntax/src/parser.rs
```diff
@@ -1973,13 +1973,49 @@ impl Parser<'_> {
...
             self.eat_and_get().expected(kind.name());
+        } else if kind == SyntaxKind::Comma
+            && let Some(number) = self.check_unit_spacing()
+        {
+            let hint = eco_format!(
+                "if you meant to use a unit, try `{number}{}` instead",
+                self.current_text(),
+            );
+            self.token.node = SyntaxNode::error(
+                SyntaxError::new("expected comma, but found numeric suffix"),
```
isuffix: Also this line will need to remove the `SyntaxError::new()` call due to the syntax warning update:


```suggestion
                "expected comma, but found numeric suffix",
```

isuffix (approved):
This looks good! Thank you

laurmaedje:
I discussed this again with @isuffix and I am a bit concerned with the non-local complexity this introduces in a generic part of the parser for a hint that is specifically about numbers. I think the only case we really care about is that we have a number followed by one space followed by a unit suffix. This local distinction could be made in the lexer and I would be more comfortable with having this logic sit there.

laurmaedje:
@GabrielDTB Hey, do you still plan to work on this? If not, that's okay, too, then I would make the changes myself.

GabrielDTB:
@laurmaedje  I do not, sorry. I thought about the edge cases a while back, got stuck between whether they matter or not, and honestly just didn't feel like figuring that out. Unfortunately I don't recall well what the edges were, but it seemed a bit thorny for what's supposed to be a pure diagnostic improvement, at least from my understanding of how things fit together at the time.

PR #8131, opened 2026-04-17
More robust overhang

### LINKED ISSUE #6735: U+2E3A should not be overhung

### LINKED ISSUE #7219: Overhang disabled if word has label

### LINKED ISSUE #8130: Show-rule disables overhanging punctuation
### Description

Show-rule (even if not applied) disables overhanging punctuation:

```typst
#set page(margin: 2pt, width: auto, height: auto)
#set par(justify: true)
#set text(size: 11pt)
#let f(body) = {
  show "x": "y"
  body
}

#block(stroke: 0.5pt + blue, width: 10pt, [#h(1fr);a, b])
#block(stroke: 0.5pt + blue, width: 10pt, [#h(1fr);#f[a,] b])
```

Image

### Operating system

Web app

### Typst version

sicikh:
Continues work of #7376.

Closes #7219, closes #8130. Also closes #6735 (see below).

For more info, see https://github.com/typst/typst/pull/7376#issuecomment-3551550938.

- Skip zero-advance glyphs when finding the first/last glyph for line overhang.
- Restore overhang when a preceding word injects label-driven zero-width glyphs.
- Keep existing behavior for lines that already start or end with regular glyphs.

New change, not included in the #7376 and what is resolving #6735:

When a glyph is shaped, its first character (`glyph.c`) may differ from the expected character (for example, `sym.wj + sym.dash.em` results in a glyph in which `glyph.c` is equal to `\u{2060}`). In this example, the old function-based implementation, which checks only the `glyph.c`, cannot verify that a particular character is a em-dash, and thus disables overhang for it. And vice versa, with font that substitutes multiple characters with different glyph, `glyph.c` can be misleading — two em-dashes in Noto Serif SC are substituted with U+2E3A, which should not be overhang, while `glyph.c` is the first em-dash, which causes the whole glyph to overhang (even with the wrong ratio!).

To solve this problem, each character in the overhang table is mapped to a glyph id in the glyph font, and only then are these ids compared to check if we should enable overhang.


sicikh:
> And vice versa, with font that substitutes multiple characters with different glyph, glyph.c can be misleading — two em-dashes in Noto Serif SC are substituted with U+2E3A, which should not be overhang, while glyph.c is the first em-dash, which causes the whole glyph to overhang (even with the wrong ratio!).

This should be tested, but current test runner has specific version of Noto Serif SC without such substituion. Code for future test is under the spoiler below (adapted from original issue). See https://github.com/typst/typst-dev-assets/issues/18.

<details>
<summary>
Test case
</summary>

```
#let ccwd = 12pt
#set page(height: auto, width: ccwd * 30 + 20pt, margin: 10pt)
#set text(
  lang: "zh",
  font: ("Noto Serif CJK SC"), // font converts 2 U+2014s to U+2E3A
  size: ccwd,
  cjk-latin-spacing: auto
)
#set par(justify: true, first-line-indent: (amount: 2em, all: true))
#place(
  top + left,
  dx: ccwd,
  box(width: ccwd * 30,
    repeat(
      rect(width: ccwd, height: 2.4em, stroke: none, fill: rgb("#7092be70")),
      gap: ccwd,
      justify: false
    )
  )
)

囗囗囗囗囗囗囗，囗囗囗囗囗囗囗囗。囗囗囗囗囗囗囗囗囗------囗囗囗囗
```

</details>

crates/typst-layout/src/inline/line.rs
```diff
@@ -668,27 +666,43 @@ fn add_par_line_marker(
...
-        // Punctuation.
-        '.' | ',' => 0.8,
-        ':' | ';' => 0.3,
-
-        // Arabic
-        '\u{60C}' | '\u{6D4}' => 0.4,
-
-        _ => 0.0,
-    }
+fn overhang(glyph: &ShapedGlyph) -> Abs {
```
laurmaedje, line 677: I'm not so comfortable with this from a performance point of view since we're iterating the full table and doing the font lookup every time in most cases. If `glyph.c` is unreliable, here would be two alternatives worth exploring:

- When creating `c` in the first place, using the first character that is not default ignorable instead of the first one
- Using `glyph.range` together with the shaped text's `text` to access the full cluster here

laurmaedje:
@sicikh Since there is still an unresolved review comment: Do you still plan to work on this?

crates/typst-layout/src/inline/line.rs
sicikh: I'm sorry for the delay.

I have explored both suggested alternative methods and have found them unsatisfactory:

- Writing the needed value of the first letter in `glyph.c` does not solve the problem that one set of characters turns into a completely different glyph (as in #6735).

- Reading the original sequence of characters that led to the creation of the glyph is subject to the same problem.

So, in any case, we are interested in real `ShapedGlyph` (and its `glyph_id`), that will end up in the document, and not in characters that shaped it. That is true especially with the case of using custom protrusion tables (#7782). 

I agree that the current solution is really naive. In d39d1aa I implemented table memoization with binary search (in 8c12fe4 changed to simple iteration), and it yielded good results.

Using a simple benchmark, the following data was obtained:

```typst
#set par(justify: true)

#for i in range(50) {
    lorem(50 * i)
    parbreak()
}
```

Total `overhang` calls: 4268.

Total `overhang` time (whole compilation took 2 s), multiplication `factor * glyph_size` included:

- before PR: ~440 us,
- naive version: 2 ms,
- table memoization: ~440 us too.

sicikh:
Sorry for the delay! I've answered to the review comment: https://github.com/typst/typst/pull/8131#discussion_r3692669691. In short, we _really_ need to compute overhang table with data from the glyph's font, so obvious optimization is to memoize it, which is done in https://github.com/typst/typst/pull/8131/commits/d39d1aa5ebcbac7885bf0551a3a6e83b6ae36f13.

crates/typst-layout/src/inline/line.rs
laurmaedje: I ran a benchmark and on a very text heavy document, it resulted in a 0.1% instruction count regression. Not nothing, but not too bad. That's probably just the baseline overhead of `comemo::memoize` doing a hash table lookup. We could probably avoid that by storing the data in the font, but not really worth it if we want to add more stuff on top in #7782.

crates/typst-layout/src/inline/line.rs
```diff
@@ -765,6 +797,23 @@ impl<'a> Items<'a> {
...
+    pub fn trailing_visible_glyph(&self) -> Option<(&ShapedText<'a>, &ShapedGlyph)> {
+        self.iter()
+            .rev()
+            .take_while(|item| matches!(item, Item::Tag(_) | Item::Text(_)))
+            .find_map(|item| {
+                let text = item.text()?;
+                text.glyphs
+                    .iter()
+                    .rev()
+                    // Skip non-positive advance artifacts from tag splits so
```
laurmaedje: What do you mean with "tag splits"?

sicikh: See https://github.com/typst/typst/pull/7376#issuecomment-3551550938.

I think the second example is meant in this case:

> With a `<label>` present, the frame contains the tag pair plus a **standalone space** which gets collapsed:
> ```text
> ==== labelled ====
>  Tag Start("text", …)
>  Text "This is a text with just," (glyphs: 25)
>  Tag End(…)
>  Text " " (glyphs: 1)
>  Text "the right length." (glyphs: 17)
> ```
> 
> That lone space glyph (the `Text " "` above) is kept for copy/paste support, but its advance is collapsed to zero (trimmed during `collect_range` / `ShapedText::build`). Because the tag mechanism splices it between the word and the comma, the old overhang logic stopped on that zero-width glyph and never saw the punctuation.
> 
> The new helper skips such glyphs to find the comma again. I agree that longer-term we need a broader tag story, but this unblocks the overhang regression without touching other inline layout code.

crates/typst-layout/src/inline/line.rs
sicikh: I think that the work on improving performance should be continued in #7782, where margin kerning is integrated into the logic of the line-breaking algorithm.

crates/typst-layout/src/inline/line.rs
laurmaedje: I see. I looked into it a bit more and it turns out that the advanced-based condition is not needed since the glyphs we want to filter out are already trimmed through the `Glyphs` struct. It's thus sufficient to take the first/last text item that has any untrimmed glyph instead of always taking the first/last even if it has no glyph.

This is to the best of my judgement desirable in all use cases of `leading_text` / `trailing_text`, so I've refactored it such that these always skip over fully-trimmed text items. I've added `Item::is_skippable` to that effect.

I've dropped `leading_visible_glyph` in favor of this (even if it's a bit more type-safe since it gives direct access to the glyph that is guaranteed to exist) because a mutable version is not cleanly possible since the text and glyph would alias.

laurmaedje:
Thanks!

YDX-2147483647:
This PR closes #6735, but it looks like that it shouldn't?

Here're the examples in #6735 compiled with ~~540db575~~ 921bb831. You can see that the results are still not expected.

图片
图片

Note that I have not checked the details of this PR. Please correct me if I missed something.

## Update

Sorry I messed up my `$env:Path`. I was testing the wrong version.

PR #8138, opened 2026-04-20
Typst-based documentation, including PDF version

### LINKED ISSUE #551: Multiple problems on the Symbols documentation page
### Description

There are multiple issues with the Symbols documentation (not to be confused with General Symbols). Namely, the two definitions (`sym` and `emoji`) have inappropriate descriptions, with no final period (the first line of the descriptions on their respective pages should be used instead). Additionally, the "Within Markup Mode" and "Within Math Mode" sub-sections are not listed in the "On This Page" section.

### Reproduction URL

https://staging.typst.app/docs/reference/symbols/

### Browsers

_No response_

### OS

_No response_

### LINKED ISSUE #741: Selecting a PR link icon in the changelogs reveals the hidden PR number in an ugly way
### Description

It selects the text under the link icon, which doesn't look great.

Text is selected under the link icon.

### Reproduction URL

https://typst.app/docs/changelog/0.14.1/

### Browsers

Firefox

### OS

_No response_

### LINKED ISSUE #813: Some symbols do not have a Unicode name in the symbol list
### Description

`sym.earth` does not display its Unicode name in the Symbol list:

Screenshot of the `earth` popup, which does not display the Unicode name.

### Reproduction URL

https://typst.app/docs/reference/symbols/sym/?query=earth

### Browsers

Firefox, Chrome

### OS

_No response_

### LINKED ISSUE #2089: Typst documentation available offline?
### Description

It'd be great if the documentation could be made available offline, as a PDF file.

### Use Case

Personally, I'd find this very useful whilst travelling: train/plane/etc.

### LINKED ISSUE #6704: Make Typst Docs Entirely in Typst!
### Description

Current, Typst uses Markdown to write its documentation. However, as a markup language, it's a shame to use other markup languages to write its docs.

There is already someone doing it (see [typst-doc-cn/tutorial](typst-doc-cn/tutorial)), and there's no technical problem doing so (use [Myriad-Dreamin/shiroa](Myriad-Dreamin/shiroa). The final effect looks fascinating!

So I here recommend the development team to migrate all Typst's docs to Typst, using [shiroa](Myriad-Dreamin/shiroa), which will make Typst more _Typst-ic_. Hope this issue can be taken into consideration.

### LINKED ISSUE #6930: The documentations' example for "style" parameter in the text function uses the wrong font
### Description

The example in the documentation for the [style parameter ](https://typst.app/docs/reference/text/text/#parameters-style)shows two examples with two different fonts, one with the "Italic" style and "Libertinus Serif" font, and the other with "Oblique" style and "DejaVu Sans" font. However, both outputs use the "Libertinus Serif" font.

Image

### Reproduction URL

https://typst.app/project/r6LngUL1yFYlVozxrHfEHI

### Browsers

Firefox

### OS

Windows

### LINKED ISSUE #7200: Handle warnings when generating the documentation
### Description

At present, when compiling examples in the documentation, warnings are just ignored.
https://github.com/typst/typst/blob/f1e8fd52f05861a711979f47fc865a6fcbeee6bd/docs/src/html.rs#L467

If we edit the above line and check the warnings, we'll find that the current docs have indeed triggered several warnings, and some of them are actually errors like #6930.

<details><summary>Example rust code that prints warnings</summary>

```rust
    let result = typst::compile::<PagedDocument>(&world);
    if !result.warnings.is_empty() {
        println!("Warnings: {:?}", result.warnings);
        println!("Source:\n```typst\n{}```\n", text);
    }
    let mut document = match result.output {
```

</details> 

<details><summary>List of warnings in the current documentations</summary>

(f1e8fd52f)

Warnings: [SourceDiagnostic { severity: Warning, span: Span(321775443720561), message: "unknown font family: dejavu sans", trace: [], hints: [] }]
Source:

```typst
#text(font: "Libertinus Serif", style: "italic")[Italic]
#text(font: "DejaVu Sans", style: "oblique")[Oblique]
```

Warnings: [SourceDiagnostic { severity: Warning, span: Span(288599812058642), message: "unknown font family: noto sans", trace: [], hints: [] }]
Source:

```typst
#set text(font: "Noto Sans", 20pt)
#set text(number-type: "lining")
Number 9.

#set text(number-type: "old-style")
Number 9.
```

Warnings: [SourceDiagnostic { severity: Warning, span: Span(287617076148568), message: "unknown font family: noto sans", trace: [], hints: [] }]
Source:

```typst
#set text(font: "Noto Sans", 20pt)
#set text(number-width: "proportional")
A 12 B 34. \
A 56 B 78.

#set text(number-width: "tabular")
A 12 B 34. \
A 56 B 78.
```

Warnings: [SourceDiagnostic { severity: Warning, span: Span(283988146145573), message: "the `path` function is deprecated, use `curve` instead", trace: [], hints: [] }]
Source:

```typst
#path(
  fill: blue.lighten(80%),
  stroke: blue,
  closed: true,
  (0pt, 50pt),
  (100%, 50pt),
  ((50%, 0pt), (40pt, 0pt)),
)
```

Warnings: [SourceDiagnostic { severity: Warning, span: Span(289253003410841), message: "the `path` function is deprecated, use `curve` instead", trace: [], hints: [] }]
Source:

```typst
// We use `.with` to get a new
// function that has the common
// arguments pre-applied.
#let star = path.with(
  fill: red,
  closed: true,
  (25pt, 0pt),
  (10pt, 50pt),
  (50pt, 20pt),
  (0pt, 20pt),
  (40pt, 50pt),
)

#star(fill-rule: "non-zero")
#star(fill-rule: "even-odd")
```

Warnings: [SourceDiagnostic { severity: Warning, span: Span(1), message: "layout did not converge within 5 attempts", trace: [], hints: ["check if any states or queries are updating themselves"] }]
Source:

```typst
// This is bad!
#let x = state("key", 1)
#context x.update(x.final() + 1)
#context x.get()
```

Warnings: [SourceDiagnostic { severity: Warning, span: Span(1), message: "layout did not converge within 5 attempts", trace: [], hints: ["check if any states or queries are updating themselves"] }]
Source:

```typst
= Real
#context {
  let elems = query(heading)
  let count = elems.len()
  count * [= Fake]
}
```

</details>

## Proposal

- For deliberately triggered warnings, change ` ```example ` to ` ```example:expect-warning `, and display them below the rendering results on the website.

  A prototype example for the [query function](https://typst.app/docs/reference/introspection/query/#caution):
Image

- For deprecation warnings, ignore them.

- For other warnings, panics when compiling them.

laurmaedje:
This PR ports the official documentation to be itself written in Typst. This affects both the longer-form content (like language explanations and guides) and in-source documentation for the standard library.

### The highlights of this work

- There is now an official PDF version of the documentation. You can view a [preview version here](https://files.typst.org/typst-docs-preview.pdf) (Note: The file weighs ~30MB).

- Both the website and PDF version of the documentation can now be fully built from the open-source code in this repository.

- Both versions can now be edited with hot reload, even for in-source (i.e. in-Rust) documentation. The previous (closed source) build process was a much larger hassle. This helps both with contributing and reviewing docs changes.

- We are not anymore limited by Markdown for documentation content. While this PR's focus is purely infrastructural, this paves the road for expanding and improving the docs content in various ways.

### How to use it?

Analogously to the test runner shell alias `cargo testit`, there is now a `cargo docit` command that lets you work with and build the docs. There are more details in `docs/README.md` (in the PR branch), but here are the most relevant things you'll want to do:

```bash
# Build HTML version into docs/dist/site
cargo docit compile
# Build PDF version into docs/dist/docs.pdf
cargo docit compile --format pdf
# Serve HTML version on http://localhost:3000
cargo docit watch
# Continuously build PDF version into docs/dist/docs.pdf
cargo docit watch --format pdf
```

The docs build system is its own crate with a small subslice of the main CLI's command surface because it can't be built with a stock Typst CLI. It depends on various docs-specific additions.

### Status

The PR is almost finished. There are four main things left to be done:

- I have to integrate this into typst.app's build process.

- The PR is sort-of blocked on https://github.com/typst/typst/issues/5907. Since this issue is crucial to fix anyway, I'll try to do so before landing this. The PR contains a workaround in a WIP commit, but I'd like to drop that one.

- I wanted to get the PR up today, but I do want to self-review and polish some more. There'll probably be lots of typos.

- Adding new fonts affected some tests. I'll need to double-check those. The relevant commit is marked as WIP.

### On the implementation

The PR (and its commit history) is roughly structured into the following steps:

- Preparatory work: Minor changes across the codebase that I needed.
- Fixes for various minor issues with the old docs content that I noticed during conversion.
- Deletion of the old docs infrastructure.
- Mass conversion of the docs (both the Markdown files and the in-source doc comments). I wrote a custom tool to perform the mass conversion, which I've published at https://github.com/typst/docs-md-to-typ for posterity.
- The Rust side of the new docs infrastructure.
- Typst-based docs components. There's some supporting Rust code in `docs/src`, but the bulk is implemented in Typst.
- Typst markup/code that ties together the mass-converted contents and new components to realize the docs content.
- CI + README

I've documented the more complex subsystems (e.g. the cross referencing system) in the Typst sources where they are implemented (e.g. `docs/components/linking.typ`). Ideally, we'd want to add more docs on subsystems relevant for docs writing and also on how to _write_ good docs in the first place to the docs README. I've left this as future work.

### How can you help?

Play around with it and let me know about issues! I've tried my best to avoid regressions in the HTML version and to have the PDF version somewhat polished, but there will definitely be issues with both versions that went unnoticed. If you spot something, just leave a review comment here or, if the PR is already merged, open an issue.

### Linked issues

- Closes https://github.com/typst/typst/issues/2089
- Closes https://github.com/typst/typst/issues/6704
- Closes https://github.com/typst/typst/issues/7200
- Fixes https://github.com/typst/typst/issues/6930
- Fixes https://github.com/typst/webapp-issues/issues/551
- Fixes https://github.com/typst/webapp-issues/issues/741
- Fixes https://github.com/typst/webapp-issues/issues/813


Enivex:
Should also close https://github.com/typst/webapp-issues/issues/206 I think?

Edit: At least the pdf version looked like it had vector graphics, so I assume the html one does as well.

MDLC01:
It looks like in the PDF, links are converted to a link followed by a footnote containing the same link. I suspect this is not the intended behavior.

laurmaedje:
> It looks like in the PDF, links are converted to a link followed by a footnote containing the same link. I suspect this is not the intended behavior.

That's intended. The whole PDF manual is designed to be usable when printed out and hyperlinks without full link text are not accessible in this case.

Andrew15-5:
Looks like a level has been skipped.

image

Enivex:
> Note: The file weighs ~30MB

I did some experimenting, and found that it compresses very well. 7zip brings it to less than 11 MB

Which makes me think that we could get pretty close to that by using PDF Object Streams. Most of the file is in uncompressed plain text.

linux-user36:
The release dates in the changelog up to version 0.13.0 all say "October 24, 2025"; I guess a copy-paste error?

MDLC01:
The `tiling` constructor example code has undesirable text wrap in the PDF file (p. 442) because the preview is displayed on the side instead of below the code. On the web version, the preview is properly displayed below the code.

MDLC01:
All symbols are marked as deprecated on the web version in the two symbol lists as well as the shorthand list.

MDLC01:
For footnote numbering, I think lowercase Roman numerals would be better than the current uppercase Latin. The first time I saw a superscript "A", it wasn't clear to me that this was referring to a footnote.

For link footnotes, a nice thing would be to deduplicate them within a page, but I suspect this would be non-trivial to do so it doesn't fit within this PR. An example of where deduplication would be useful is in the first page of the 0.14.0 changelog, where there are two links to https://github.com/LaurenzV.

laurmaedje:
> For footnote numbering, I think lowercase Roman numerals would be better than the current uppercase Latin.

I tried that but the letters seemed _very_ small. I'll give it another shot. Maybe it's fine.

MDLC01:
Multiline default parameter values look a bit weird both in the PDF and in the web version, but I don't have a good solution. An example of a multiline default value is for the `text.costs` parameters.

Notably, they look different in the PDF and in HTML: in the PDF, they extend above the heading, while on the web version, they extend below the heading.

Enivex:
The way paper sizes are presented in the PDF could probably be improved. The same description is repeated for every single variant

image

laurmaedje:
> Multiline default parameter values look a bit weird both in the PDF and in the web version, but I don't have a good solution. An example of a multiline default value is for the text.costs parameters.

Yeah, I noticed that discrepancy. I think the look is overall fine, though the discrepancy isn't great. But with some planned changes to baseline handling in #7895, the PDF version will likely align with the HTML version in the future, so I accepted this as okay for now.

laurmaedje:
> The way paper sizes are presented in the PDF could probably be improved. The same description is repeated for every single variant

Yeah. I didn't do anything special, so for now, it's just like in the web version, though I see how it's a bit worse for the print version. I didn't want to do too many drive-by improvements, mostly for time reasons, but also to keep the PR focused on infrastructure. But definitely room for improvement there.

MDLC01:
> it's just like in the web version

Note that, on typst.app (i.e., the docs for 0.14.2), the "View options" `<details>` is empty and there is a special "View paper sizes" `<details>` with a comma-separated list of valid paper sizes.

Enivex:
Could I suggest using a header where the full names of all ancestors are mentioned?

image

It would be helpful for getting oriented that it's visible that "3" refers to "Reference", "5" refers to "Foundations", and "2" refers to "array". Those could also be links to get to the start of that chapter, section or subsection.

docs/assets/index.typ
```diff
@@ -0,0 +1,60 @@
...
+  "HKGrotesk-MediumItalic.ttf",
+  "HKGrotesk-Regular.ttf",
+  "HKGrotesk-SemiBold.ttf",
+  "HKGrotesk-SemiBoldItalic.ttf",
+  "HKGrotesk-Thin.ttf",
+)
+
+#for path in fonts {
+  asset(
+    "assets/" + path,
```
MDLC01:
```suggestion
    "assets/fonts/" + path,
```

This is currently inconsistent with `base.css`, preventing the fonts from loading properly.

MDLC01:
Typed HTML function parameter default values are all listed as `none`. For reference, on typst.app (i.e., docs for Typst 0.14.2), no default is listed.

docs/components/footnote.typ
```diff
@@ -0,0 +1,43 @@
...
+// adds this extra feature.
+
+// Creates a listing of all footnotes that are descendants of the element
+// with the location `scope`.
+#let footnote-container(scope) = context {
+  let notes = query(docs.selector-within(footnote, scope))
+  if notes.len() == 0 { return }
+  html.elem("section", attrs: (role: "doc-endnotes"), {
+    for note in notes {
+      let num = counter(footnote).at(note.location()).first()
```
MDLC01:
```suggestion
      let num = counter(footnote).display(note.numbering, at: note.location())
```

Currently, `num` is simply an integer so the numbering format is not respected.

Right now, this does not create a visible issue, but if the `set footnote(numbering: "A")` rule were moved from `paged-styling` to `styling` in the future, this would become noticeable.

YDX-2147483647:
Does it run on Windows?

```sh
$ cargo docit compile
  Finished `dev-fast` profile [optimized + debuginfo] target(s) in 14m 21s
     Running `target\dev-fast\typst-docs.exe compile`
error: path must not contain a backslash
   ┌─ docs/components/live.typ:20:56
   │
20 │ #let live-docs-at-path(path) = docs.docs-in-source(read(path))
   │                                                         ^^^^
   │
   = hint: use forward slashes instead: `"/crates/typst-library/src/foundations/args.rs"`
   = hint: in earlier Typst versions, backslashes indicated path separators on Windows
   = hint: this behavior is no longer supported as it is not portable

  while calling `live-docs-at-path` at docs/components/live.typ:42:13
    live-docs-at-path(path)
  while calling `live-docs` at docs/components/category.typ:457:4
    live-docs(ty-info.docs, ty-info.def-site)
  while calling `ty-section` at docs/components/category.typ:667:6
    ty-section(route, name, value, info, deprecation-info)
  while including `foundations.typ` at docs/content/reference/library/index.typ:5:1
    include "foundations.typ"
  while including `library/index.typ` at docs/content/reference/index.typ:22:1
    include "library/index.typ"
  while including `reference/index.typ` at docs/content/index.typ:3:1
    include "reference/index.typ"
  while including `content/index.typ` at docs/main.typ:8:1
    include "content/index.typ"

error: process didn't exit successfully: `target\dev-fast\typst-docs.exe compile` (exit code: 1)
```

A dirty fix:

```diff
diff --git a/docs/src/reflect.rs b/docs/src/reflect.rs
index be6ac279b..f9cf113be 100644
--- a/docs/src/reflect.rs
+++ b/docs/src/reflect.rs
@@ -111,7 +111,7 @@ fn describe_symbol(symbol: &Symbol) -> Dict {
 /// Provides details where a native definition is located in the sources.
 fn describe_def_site(site: DefSite) -> Dict {
     dict! {
-        "path" => site.path,
+        "path" => site.path.replace("\\", "/"),
         "key" => site.key,
     }
 }
 ```

---


In addition, on my machine, the `dev-fast` profile seems to compile more slowly and consume more storage than `dev`.
I didn't carefully control the environment, but this should indicate that the difference between them is insignificant?

|Profile| Duration of compilation after `cargo check && cargo clean` |Size of `target/` after compilation|
|---|---|---|
|dev-fast| 14m 21s| 7.05 GiB| 
|dev| 9m 52s| 5.97 GiB|

**Comment by laurmaedje** [comment]
> In addition, on my machine, the dev-fast profile seems to compile more slowly and consume more storage than dev.

Do you mean just the Rust build being slower? That's expected. The dev-fast build is a trade off to get slower Rust compilation but reasonably fast Typst compilation. The docs are too heavy to compile with a plain debug build, it gets really slow. But a full release build is way too slow on the Rust side.

**Comment by MDLC01** [comment]
This PR seems to fix the following issues as well:
- https://github.com/typst/webapp-issues/issues/551
- https://github.com/typst/webapp-issues/issues/741
- https://github.com/typst/webapp-issues/issues/813

File: docs/content/reference/library/symbols.typ [hunk 3]
```diff
@@ -0,0 +1,303 @@
...
+      html.ul(class: "symbol-grid")
+    })
+  }))
+
+  html.template(id: "flyout-sym-row", {
+    html.li(html.button(html.span(class: "sym")))
+  })
+}
+
+// Once list entry in symbol list / cell in a symbol grid.
```
MDLC01:
```suggestion
// One entry in a symbol list or cell in a symbol grid.
```

I am not sure this is the correct fix, but right now I am struggling to understand the sentence.

laurmaedje:
> For footnote numbering, I think lowercase Roman numerals would be better than the current uppercase Latin. The first time I saw a superscript "A", it wasn't clear to me that this was referring to a footnote.

> I tried that but the letters seemed very small. I'll give it another shot. Maybe it's fine.

I tried again, but it's really too small in my opinion. What about `[A]` as the numbering?

laurmaedje:
> Note that, on typst.app (i.e., the docs for 0.14.2), the "View options" <details> is empty and there is a special "View paper sizes" <details> with a comma-separated list of valid paper sizes.

Okay, fair enough. I did not recall that.

MDLC01:
> I tried again, but it's really too small in my opinion. What about `[A]` as the numbering?

I think `[A]` would be better than `A`.

~What about `[i]`? This is even clearer imo and the additional brackets make the thing a bit wider and higher.~ I tried it and it looks weird.

Another solution is `*`, but when superscripted, the asterisk is smaller than a superscript `i`, so I don't think that would be acceptable.

docs/content/reference/library/symbols.typ
```diff
@@ -0,0 +1,303 @@
...
+      if class != none { math-class-names.at(class) }
+    },
+    data-override: symbol-overrides.at(value, default: none),
+    data-deprecation: deprecation,
+  )
+
+  let body = symbol-overrides.at(value, default: value)
+
+  context if target() == "paged" {
+    box(width: 5em, h(1fr) + text(font: symbol-fonts, body) + h(1em))
```
MDLC01:
```suggestion
    let style = if value in symbol-overrides {
      (fill: rgb("#565565"), weight: 500, style: "italic")
    }
    box(width: 5em, h(1fr) + text(font: symbol-fonts, ..style, body) + h(1em))
```

This makes it consistent with the web version.

docs/content/reference/library/symbols.typ
```diff
@@ -0,0 +1,306 @@
...
+  context if target() == "paged" {
+   let style = if value in symbol-overrides {
+      (fill: colors.dark-gray.shade-10, weight: 500, style: "italic")
+    }
+    box(width: 5em, h(1fr) + text(font: symbol-fonts, ..style, body) + h(1em))
+    if title == auto {
+      raw(full)
+    } else {
+      text(fill: colors.text.syntax.teal, raw(title))
+    }
```
MDLC01, line 171:
```suggestion
    let wrapper = if deprecation != none { strike } else { it => it }
    if title == auto {
      wrapper(raw(full))
    } else {
      text(fill: colors.text.syntax.teal, wrapper(raw(title)))
    }
```

Currently, the PDF version does not show any deprecation information for symbols. This simply strikes them, but ideally the deprecation messages would be displayed as well.

Given that there will have to be a separate PR to add support for modules, this can also be left as a future work.

docs/components/pill.typ
```diff
@@ -0,0 +1,78 @@
...
+    let inner = raw(name)
+    let body = if linked { link(def-dest(ty), inner) } else { inner }
+    box(
+      fill: pill-fill.at(c),
+      inset: (x: 0.3em),
+      outset: (y: 0.3em),
+      radius: 0.3em,
+      // Make the pill contents a little smaller because the pill itself also
+      // adds visual weight.
+      text(0.9em, body),
```
MDLC01:
```suggestion
      text(0.9em, weight: "regular", body),
```

For consistency with the web version, where the pill text is always in regular weight, even in headings:

https://github.com/typst/typst/blob/9b2d69be8c2ab3aff537a5fad609599713e1cd31/docs/assets/docs.css#L501

MDLC01:
I love that arguments for nested function definitions (e.g., arguments to the `footnote.entry` function, or to `array.push`) are indented in the PDF version. That makes the docs so much more readable. I wonder whether something similar could be done in the web version as well.

laurmaedje:
> I wonder whether something similar could be done in the web version as well.

I would leave that out of this PR but we can experiment with that after it has landed!

docs/content/changelog/index.typ
```diff
@@ -0,0 +1,151 @@
...
+#docs-chapter(
+  title: "Changelog",
+  route: "/changelog",
+  description: "Learn what has changed in the latest Typst releases and move your documents forward.",
+)[
+  Learn what has changed in the latest Typst releases and move your documents forward. This section documents all changes to Typst since its initial public release.
+
+  #context if target() == "paged" {
+    info[
+      Some changelog items contain references like this one: #pr(5017). These refer to a _pull request_ or _issue_ on #link("https://github.com")[GitHub] related to the change.
```
istudyatuni, line 82:
```suggestion
      Some changelog items contain references like this one: #pr(5017). These refer to a _pull request_ or _issue_ on #link("https://github.com/typst/typst")[GitHub] related to the change.
```

MDLC01: Some references are not to the typst/typst repository though.

istudyatuni: Missed that, thanks! Then maybe something like this:
```suggestion
      Some changelog items contain references like this one: #pr(5017). These refer to a _pull request_ or _issue_ on #link("https://github.com")[GitHub] related to the change, if not stated explicitly, on #link("https://github.com/typst/typst")[Typst repo].
```

YDX-2147483647:
Various minor issues: (✅ = resolved)

1. (Web+PDF) ✅ Can `cargo docit` be executed only from the repo root? It seems that the following does not work.

	```sh
	$ cd docs
	$ cargo docit c -f pdf
	error: file not found (searched at .\docs\main.typ)
	error: process didn't exit successfully: `…\typst\target\dev-fast\typst-docs.exe c -f pdf` (exit code: 1)
	```


2. (Web) ✅ Does `cargo docit` support building a website under a sub-folder base (e.g. `https://owner.github.io/repo/`)?

   The old CLI supports it.
   https://github.com/typst/typst/blob/78251637643fc4c8fbc8ec174dd622ff23e06861/docs/src/main.rs#L117-L125

3. (PDF) It looks like `docs.pdf` omits the concept of part (language/library/export)?

   <details>
   <summary>Screenshot</summary>
   图片
   </details>

    If that's intentional, the sentence below needs to be revised.
    https://github.com/typst/typst/blob/251ba8b1550180fef36fe80873e91655f1318309/docs/content/reference/index.typ#L12
   


4. (Web) The new CSS style for defaults does not fit complex values well.

	Example: [`par.justification-limits`](http://localhost:3000/reference/model/par/#parameters-justification-limits)
	
	The style in PDF looks better.
	
	<details>
	<summary>Screenshot</summary>
	Web:
	图片
	PDF:
	图片
	</details>


5. (Web) Preview-only examples are no longer collapsible. (Perhaps on purpose?)

	Example: `par.leading` [localhost:3000](http://localhost:3000/reference/model/par/#parameters-leading) vs. [typst.app](https://typst.app/docs/reference/model/par/#parameters-leading)

6. (Web) HTML typed APIs now pollute the search result.
	
	For example, the search result for `span` was `{table,grid}.{,cell.}{rowspan,colspan}`, but now it becomes `html.{,col.,colgroup.}span`, `{td,th}.{rowspan,colspan}`. 
	https://github.com/typst/typst/issues/8269#issuecomment-3345321770

7. (Web) In the `font-family` for emoji, _Segoe UI Symbol_ has been moved before _Segoe UI Emoji_, causing inconsistencies on Windows.
	
    <details>
	<summary>Example</summary>
	
	
	Before: https://typst.app/docs/reference/symbols/emoji/?query=plane
	图片
	
	After: http://localhost:3000/reference/symbols/emoji/?query=plane
	图片
	
	
	</details>


8. (Web) A similar font inconsistency happens to Greek letters in `<code>`, but I don't know why...
		
	<details>
	<summary>Example</summary>
	
	Before: https://typst.app/docs/reference/model/quote/#example
	图片
	
	After: http://localhost:3000/reference/model/quote/#example
	图片
	
	
	</details>

9. (PDF) In the example for `text.font`, 置 in the source code uses the TC (Traditional Chinese) / JP (Japanese) glyph, but that example is actually for SC (Simplified Chinese).
	
	<details>
	<summary>Screenshot</summary>
	图片
	图片
	</details>

	The list of bibliography styles has the similar problem. For example, _China National Standard GB/T 7714-2005
(numeric, 中文)_  and _SIST02 (日本語)_.
	These texts should be language-tagged in the future.

10. (PDF) Some table columns are too far apart, making it difficult to match the rows.

	<details>
	<summary>Screenshot</summary>
	图片
	图片
	</details>
	
11. (PDF) Twemoji does not cover new emoji symbols. For instance, `dancing.ballet` and `bigfoot`.

	<details>
	<summary>Screenshots</summary>
	
	https://github.com/typst/typst/blob/251ba8b1550180fef36fe80873e91655f1318309/docs/content/reference/library/symbols.typ#L49-L56
	图片
	图片

	</details>

MDLC01:
3. This roughly corresponds to https://github.com/typst/typst/issues/8140.

4. See https://github.com/typst/typst/pull/8138#issuecomment-4284241166.

docs/content/reference/library/symbols.typ
```diff
@@ -0,0 +1,307 @@
...
+  "special": "Special",
+)
+
+// Fonts that are used to display the symbols.
+#let symbol-fonts = (
+  // TODO: Maybe prefer Libertinus Serif for the full default look?
+  // Note: Keep in sync with docs.css.
+  fonts.body,
+  "New Computer Modern Math",
+  "Twitter Color Emoji",
```
MDLC01, line 55: Those two fonts are not imported in `assets/index.typ` so they are not available.

docs/content/reference/library/symbols.typ
```diff
@@ -0,0 +1,307 @@
...
+  "space": "Space",
+  "unary": "Unary",
+  "vary": "Vary",
+  "special": "Special",
+)
+
+// Fonts that are used to display the symbols.
+#let symbol-fonts = (
+  // TODO: Maybe prefer Libertinus Serif for the full default look?
+  // Note: Keep in sync with docs.css.
```
MDLC01, line 52: `docs.css` contains many more fonts that are not listed in this Typst file:

https://github.com/typst/typst/blob/251ba8b1550180fef36fe80873e91655f1318309/docs/assets/docs.css#L1398-L1402

docs/assets/base.css
```diff
@@ -0,0 +1,528 @@
...
+
+@font-face {
+  font-family: "NewComputerModernMath";
+  font-weight: normal;
+  font-display: swap;
+  src:
+    local("NewComputerModernMath-Regular"),
+    local("NewComputerModernMath"),
+    url("/assets/fonts/NewCMMath-Regular.otf") format("truetype");
+}
```
MDLC01, line 147: Right now, New Computer Modern Math is imported here but not added as an asset in `assets/index.typ` (note that New Computer Modern Math is not even present in typst-dev-assets).

I think it might be beneficial to generate those `@font-face` entries procedurally based on the fonts listed in `asssets/index.typ`. This would ensure that every font that is imported here is actually available as an asset, and, conversely, that every font that is added as an asset is actually usable.

MDLC01 (commented):
Some thoughts about fonts.

docs/assets/docs.css
```diff
@@ -0,0 +1,1525 @@
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  padding: 8px;
+  overflow: hidden;
+}
+
+.symbol-grid *[data-deprecation] button {
+  background: #f7f7f9;
+}
```
MDLC01:
```suggestion
```

I was always bothered by the fact that deprecated symbols are a subtly different shade of gray. I thought there was an issue on typst/webapp-issues but I can't find it.

docs/content/reference/library/symbols.typ
laurmaedje: Some of those are system fonts which can make sense in a browser but are not available in the sandboxed docs build system.

docs/assets/base.css
laurmaedje: Since `base.css` is derived from styles in our proprietary code base, that would make it harder to ensure these don't go out of sync.

MDLC01: I see. But right now it is similarly hard to ensure `base.css` doesn't go out of sync with `assets/index.typ`.

crates/typst-syntax/src/node.rs
```diff
@@ -159,10 +159,32 @@ impl SyntaxNode {
...
     pub fn synthesize(&mut self, span: Span) {
+        self.synthesize_with(|_| span);
+    }
+
+    /// Set a custom span for each node gives its range.
+    ///
+    /// Should be called with `offset = 0` on the root node.
+    pub fn synthesize_with(&mut self, mut f: impl FnMut(Range<usize>) -> Span) {
+        self.synthesize_with_impl(0, &mut f);
+    }
```
isuffix: Instead of taking an arbitrary function, could we just take a `&RangeMapper` and split the functions? This makes more sense to me since it would justify defining `RangeMapper` in `span.rs`, since it could otherwise go anywhere.

This would also reduce the amount of external callers of `Span::from_range`, as we could otherwise make it private to the crate aside from loaded external file diagnostics. I'm thinking about that because I've been working on adding sub-ranges to diagnostics as mentioned in #8188, and if we do, then we would also have an extra `u64` to use for external diagnostic ranges, so they can take 46 bits per start/end index instead of 23 since external files are reasonably likely to exceed 8MB. Typst sources are less likely to exceed 8MB, and since we still want raw-ranges for this `synthesize_with` functionality, we can keep them as-is with 23 bits. But it would be nice to limit their use to this crate to help simplify the public Span interfaces.

crates/typst-eval/src/lib.rs
```diff
@@ -111,7 +110,15 @@ pub fn eval_string(
         SyntaxMode::Math => parse_math(string),
     };
 
-    root.synthesize(span);
+    match spans {
+        SpanMode::Uniform(span) => root.synthesize(span),
```
isuffix, line 115: We should add an `if span.is_detached() => {}` branch to avoid work for some callers.

docs/content/reference/library/symbols.typ
laurmaedje: Those are used for text rendering in the PDF version and not for the web version, so the assets shouldn't matter.

docs/assets/base.css
laurmaedje: I fixed the missing fonts. I would be okay with auto-generating the CSS, but didn't want to spend time on this myself. If you'd like, you could submit a PR later.

laurmaedje:
@YDX-2147483647 

> 1. Can cargo docit be executed only from the repo root? It seems that the following does not work.

Fixed.

> 2. Does cargo docit support building a website under a sub-folder base?

I had initially added support for this as `--base`, but have now replaced it with something else. The docs are now usable as a Typst package and you can provide your own main file that imports `#import "@typst/docs:0.0.0"`. The docs are essentially a Typst template now. :D The entrypoint function lets you customize the sub-folders, add extra chapters, and any future configuration would also go there. Note though that you need the special docs CLI, it's not a package that works with the normal compiler.

> 3. It looks like `docs.pdf` omits the concept of part (language/library/export)?

I don't see a good way to express that in the PDF outline. Not sure whether to still have this concept in the PDF or to reword the text (falling under https://github.com/typst/typst/issues/8140). I will leave this as future work since the "parts" concept is already not particularly strong in the web docs.

> 4. The new CSS style for defaults does not fit complex values well.

Fixed.

> 5. Preview-only examples are no longer collapsible. (Perhaps on purpose?)

It was not intentional, but I think it makes sense to treat them differently from examples. There are only two uses of this that are collapsible in the current docs (`par.leading` and `page.footer-descent`). For both of these, the preview is part of the explanation more than it is an example.

> 6. HTML typed APIs now pollute the search result.

Fixed.

> 7. In the font-family for emoji, Segoe UI Symbol has been moved before Segoe UI Emoji, causing inconsistencies on Windows.

Fixed. The `emoji` class was missing from the symbol grid.

> 8. A similar font inconsistency happens to Greek letters in `<code>`, but I don't know why...

Probably also because the web fonts were not linked correctly, which should be fixed now.

> 9. In the example for `text.font`, 置 in the source code uses the TC (Traditional Chinese) / JP (Japanese) glyph, but that example is actually for SC (Simplified Chinese).

Fixed.

> 10. Some table columns are too far apart, making it difficult to match the rows.

Fixed.

> 11. Twemoji does not cover new emoji symbols. For instance, dancing.ballet and bigfoot.

I don't see an immediate way to deal with this, so I'll leave it as future work.

laurmaedje:
@Enivex 

> Could I suggest using a header where the full names of all ancestors are mentioned?

I could see this in principle, but I don't have time to work on this at the moment. I would leave it as future work.

istudyatuni:
> Twemoji does not cover new emoji symbols

Twemoji with Unicode 17 support is at https://github.com/jdecked/twemoji

istudyatuni:
Not sure if this should be on the list

image

Enivex:
> Not sure if this should be on the list
> 
> image

That's not really the fault of the documentation 

See https://github.com/typst/codex/pull/79

laurmaedje:
I'll squash most of the later commits before merging, but in case someone wants to inspect a particular fix, the commit history should still be available [here](https://github.com/typst/typst/commits/8f28a72eab4023376c0cdcc41fa00e5b9ed5408e). (Though I'm not sure whether GitHub retains it indefinitely.)

crates/typst-eval/src/lib.rs
```diff
@@ -140,12 +144,18 @@ pub fn eval_string(
...
+        SyntaxMode::Math => {
+            let span = match spans {
+                SpanMode::Uniform(span) => span,
+                SpanMode::Mapped { .. } => Span::detached(),
+            };
+            Value::Content(
+                EquationElem::new(root.cast::<ast::Math>().unwrap().eval(&mut vm)?)
+                    .with_block(false)
+                    .pack()
+                    .spanned(span),
```
isuffix: For this, I think you can just replace `.spanned(span)` with `.spanned(root.span())` and then we don't need to use a detached span

crates/typst-macros/src/util.rs
```diff
@@ -261,3 +277,24 @@ pub mod kw {
     syn::custom_keyword!(parent);
     syn::custom_keyword!(ext);
 }
+
+/// Extract the first line of documentation.
+pub fn oneliner(docs: &str) -> String {
+    let paragraph = docs.split("\n\n").next().unwrap_or_default();
```
istudyatuni, line 283:
```suggestion
    let Some(paragraph) = docs.split("\n\n").next() else {
        return "".to_string();
   };
```

crates/typst-svg/src/image.rs
```diff
@@ -53,43 +53,91 @@ pub fn convert_image_scaling(scaling: Smart<ImageScaling>) -> Option<&'static st
...
+            ImageKind::Pdf(pdf) => {
+                (WebImageFormat::Svg, Bytes::from_string(pdf_to_svg(pdf)))
+            }
+        };
+        Self { format, data }
+    }
+
+    /// Turns the image into a base64 URL.
+    ///
+    /// The format of the URL is `data:{mime};base64,`.
```
istudyatuni, line 133:
```suggestion
    /// The format of the URL is `data:{mime};base64,[encoded data]`.
```

crates/typst-utils/src/lib.rs
```diff
@@ -437,3 +437,38 @@ pub fn defer<T, F: FnOnce(&mut T)>(
 
     DeferHandle { thing, deferred: Some(deferred) }
 }
+
+/// Describes the source code where something is defined.
+///
+/// This does not include a concrete line number because, due to the way Rust
```
istudyatuni, line 443: It's possible to get correct `line!()` by using `track_caller`, but I tried it, and it doesn't work for everything, since it can be used only on functions, and some macros generate impls with consts

laurmaedje:
Plsss no more reviews, I just want to merge 😅

crates/typst-utils/src/lib.rs
laurmaedje: Interesting. Anyway, the key-based system has other benefits. It's stable over reorderings and stuff. The more accurate line number would have made a more robust system less pressing, but it would still suffer fundamental problems.

PR #8143, opened 2026-04-21
Expose `bibliography.entry` for entry-level bibliography customization

hongjr03:
This PR exposes bibliography items as a public `bibliography.entry` element so that they can be customized with show rules.

At the moment, `bibliography` can only be customized as a whole. Its individual entries are generated internally and are not exposed as public elements, so there is currently no way to write entry-level show rules for bibliography items. In [#942](https://github.com/typst/typst/issues/942#issuecomment-1666261368), @laurmaedje suggested that a `bibliography.entry` or `bibliography.item` function could provide this kind of customization.

With this change, each generated bibliography item becomes a `bibliography.entry`, which exposes the rendered `key`, `prefix`, and `body` of the entry. It also provides `inner(prefix, body)` and `indented(prefix, body, gap: ..., indent: ...)` as helper functions for rebuilding an entry while preserving bibliography-specific layout behavior. 

For example, users can now write rules like this to get a paragraph like bibliography entry and can manipulate its hanging-indent:

```typ
#show bibliography.entry: it => {
  set par(hanging-indent: 2em)
  par({
    it.prefix
    h(.5em)
    it.body
  })
}

@fukushima1982 and @lecun1998 and @krizhevsky2012 and @simonyan2014.

#bibliography(
  bytes(
    ```bib
    @incollection{fukushima1982,
      author    = {Fukushima, K. and Miyake, S.},
      title     = {Neocognitron: A Self-Organizing Neural Network Model for a Mechanism of Visual Pattern Recognition},
      booktitle = {Competition and Cooperation in Neural Nets},
      publisher = {Springer},
      address   = {Berlin, Heidelberg},
      year      = {1982},
      pages     = {267--285}
    }

    @article{lecun1998,
      author  = {LeCun, Y. and Bottou, L. and Bengio, Y. and Haffner, P.},
      title   = {Gradient-Based Learning Applied to Document Recognition},
      journal = {Proceedings of the IEEE},
      year    = {1998},
      volume  = {86},
      number  = {11},
      pages   = {2278--2324}
    }

    @inproceedings{krizhevsky2012,
      author    = {Krizhevsky, A. and Sutskever, I. and Hinton, G. E.},
      title     = {ImageNet Classification with Deep Convolutional Neural Networks},
      booktitle = {Advances in Neural Information Processing Systems},
      year      = {2012},
      volume    = {25},
      pages     = {1097--1105}
    }

    @article{simonyan2014,
      author  = {Simonyan, K. and Zisserman, A.},
      title   = {Very Deep Convolutional Networks for Large-Scale Image Recognition},
      journal = {arXiv preprint arXiv:1409.1556},
      year    = {2014}
    }
    ```.text,
  ),
)
```

image


xkevio:
I will refer to my previous attempt to do the exact same thing as you did, though I have not fully compared the code. Have you checked layout iterations in case of grid-based bibliography styles?

https://github.com/typst/typst/pull/5932

(also your example is a little weird with calling `par` directly while you claim that you have an `indented` function)

hongjr03:
> I will refer to my previous attempt to do the exact same thing as you did, though I have not fully compared the code. Have you checked layout iterations in case of grid-based bibliography styles?
> 
> #5932
> 
> (also your example is a little weird with calling `par` directly while you claim that you have an `indented` function)

Thanks for pointing this out. I had read the issue discussion, but I had missed your earlier PR [#5932](https://github.com/typst/typst/pull/5932), and it is very relevant here. My current implementation also computes prefix alignment from realized `bibliography.entrys`, so I need to check carefully whether it runs into the same issue.

As for the example: the intention there was not to suggest that `indented` is unnecessary. I was trying to show that with original `bibliography` to control hanging indentation is not really possible. But I agree that this is confusing in the context of this PR.

hongjr03:
I think this still has the same core problem after checking it more carefully. It seems to need infras we don't have yet for doing this kind of layout work locally. :(

I'll close this for now.

xkevio:
Sorry, I did not mean to discourage you and the work that went into this. I am excited that there is enough interest in this feature that at least two people have attempted it so far! ^^

Unfortunately, we will probably have to wait for local introspection for this to make it in.

hongjr03:
> Sorry, I did not mean to discourage you and the work that went into this. I am excited that there is enough interest in this feature that at least two people have attempted it so far! ^^
> 
> 
> 
> Unfortunately, we will probably have to wait for local introspection for this to make it in.

No worries at all — we're aiming on the same goal :)

PR #8147, opened 2026-04-21
Do not emit note-style citations when already in a footnote

benzeneringlq:
Add a boolean variable so that Cite knows whether it is in a footnote.
```
foo #footnote[See, for example: #cite(<alpha>)]
foo #footnote[See, for example: #cite(<alpha>)]
foo #footnote[See, for example: #cite(<alpha>)]

foo #footnote[See, for example: #cite(<beta>)]
foo #footnote[See, for example: #cite(<beta>)]
foo #footnote[See, for example: #cite(<beta>)]

foo #footnote[See, for example: #cite(<alpha>)]
```

image

### DIFF
--- a/crates/typst-library/src/model/bibliography.rs
+++ b/crates/typst-library/src/model/bibliography.rs
@@ -771,7 +771,8 @@ impl<'a> Generator<'a> {
                 subinfos,
                 span: first.span(),
                 footnote: normal
-                    && style.settings.class == citationberg::StyleClass::Note,
+                    && style.settings.class == citationberg::StyleClass::Note
+                    && !first.infootnote.unwrap_or(false),
             });
 
             driver.citation(CitationRequest::new(
--- a/crates/typst-library/src/model/cite.rs
+++ b/crates/typst-library/src/model/cite.rs
@@ -7,7 +7,7 @@ use crate::foundations::{
 };
 use crate::introspection::Locatable;
 use crate::model::bibliography::Works;
-use crate::model::{CslSource, CslStyle};
+use crate::model::{CslSource, CslStyle, FootnoteElem};
 use crate::text::{Lang, Region, TextElem};
 
 /// Cite a work from the bibliography.
@@ -115,6 +115,11 @@ pub struct CiteElem {
     #[internal]
     #[synthesized]
     pub region: Option<Region>,
+
+    /// Whether this citation is inside a footnote.
+    #[internal]
+    #[synthesized]
+    pub infootnote: bool,
 }
 
 impl Synthesize for Packed<CiteElem> {
@@ -122,6 +127,7 @@ impl Synthesize for Packed<CiteElem> {
         let elem = self.as_mut();
         elem.lang = Some(styles.get(TextElem::lang));
         elem.region = Some(styles.get(TextElem::region));
+        elem.infootnote = Some(styles.get(FootnoteElem::infootnote));
         Ok(())
     }
 }
--- a/crates/typst-library/src/model/footnote.rs
+++ b/crates/typst-library/src/model/footnote.rs
@@ -79,6 +79,10 @@ pub struct FootnoteElem {
     #[default(Numbering::Pattern(NumberingPattern::from_str("1").unwrap()))]
     pub numbering: Numbering,
 
+    #[internal]
+    #[parse(Some(false))]
+    pub infootnote: bool,
+
     /// The content to put into the footnote. Can also be the label of another
     /// footnote this one should point to.
     #[required]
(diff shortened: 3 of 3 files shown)


nicl:
Just to say, I don't know this codebase well-enough to vouch for the specific code changes, but the behaviour/bug addressed is incredibly important to my workflow and I'm sure others too.

So if there is anything that I or others can do to help get this reviewed and merged, please shout.

laurmaedje:
Could you provide references for this being the desirable default behavior? It seems at least a bit opinionated. I also wonder whether it would be expected for parentheses to show up in some way if this is the default.

For some bibliography styles (e.g. `chicago-notes`), you can already achieve the effect of this PR in user space like this:
```typ
#show footnote.entry: set cite(form: "full")
```

If implemented in the compiler, it would be nice if we could also use a show-set rule as that means it remains overridable by users. To do that, we'd probably need a new `form` or other configuration option though since we can't select `"normal"` without getting the automatic footnote.

nicl:
@laurmaedje just to add my 2c. Journals in the humanities often prefer footnoted citations, and these are very often discursive, i.e. they contain nested citations. Certainly in my field at least (theology), none of them would follow the current behaviour but would be consistent with this proposed fix. To be honest, I am struggling to imagine a case where the existing behaviour would be desirable (i.e. are we convinced the status quo is intentional behaviour rather than simply being a bug?).

Are there any known academic examples that actually follow the *current* behaviour?

But very happy to be corrected on this!

This is a screenshot of a randomly selected article from a prominent journal in the field (https://www.jstor.org/stable/48804214?seq=3) just to illustrate the norm (i.e. frequent nested citations, without extra parentheses etc):

Screenshot 2026-05-08 at 11 40 38

zacharybutler:
I can second @nicl on this. I am a PhD student working in New Testament Textual Criticism. I would love to use Typst for my dissertation, but this footnote behavior is a huge hurdle. I have not been able to figure out how to make it accomplish the format demonstrated in the example picture above.

dayala1:
The problem was unclear to me. You mentioned nested references, but I dont see anything in the examples that is necessarily nested. I see footnotes that contain references, but is the problem that those split the footnote into several ones?

Or is the issue that having a show rule to automatically place references as a footnote, the inclusion of a reference in a footnote should not create a different footnote e.g. have the show rule be aware that the citation is already in a footnote)?

nicl:
@dayala1 Thanks for the questions. Yes, by 'nested citations' is meant citations nested within a footnote. As you can see from the example screenshots, Typst's current behaviour (of creating multiple footnotes) here is quite unusual --- it neither meets expectations, nor what is standard in parts of academia that actually use this kind of footnote.

To add, the screenshot linked in @benzeneringlq 's original PR description is very helpful in displaying and contrasting the current (I think buggy) behaviour and the desired behaviour.

dayala1:
Something strange is indeed happening. 

When citation form is set to default, the footnote breaks into several ones, and the first citation has a different style:

```typst
#set cite(style: "chicago-notes")

foo #footnote[See, for example: #cite(<maclure2013clarity>)]
foo #footnote[See, for example: #cite(<maclure2013clarity>)]
foo #footnote[See, for example: #cite(<maclure2013clarity>)]


#bibliography("bibliography.bib", style: "chicago-notes")
```

image

When it is set to "prose", it does not break, but the style of the first one is still different. Also, unclosed parentheses appear:

```typst
#set cite(form: "prose", style: "chicago-notes")

foo #footnote[See, for example: #cite(<maclure2013clarity>)]
foo #footnote[See, for example: #cite(<maclure2013clarity>)]
foo #footnote[See, for example: #cite(<maclure2013clarity>)]

#bibliography("bibliography.bib", style: "chicago-notes")
```

image

When it is set to author, the name appears differently in the first one:

image

When it is set to full (or year), it works well:

image

Reference used:

```bib
@article{maclure2013clarity,
  title={‘Clarity bordering on stupidity’: where’s the quality in systematic review?},
  author={MacLure, Maggie},
  journal={Knowledge Production},
  pages={55--80},
  year={2013},
  publisher={Routledge}
}
```

It would also be nice to have a way to have full citations without the bibliography at the end, but it is possible to place it in a 0 height clipping block or similar.

benzeneringlq:
@dayala1  Thanks for your comment and tests
### the first citation has a different style
The difference in the first citation is actually by design. Under the Chicago style, the first mention of a source provides the full bibliographic details, while subsequent citations are automatically shortened to save space and avoid redundancy.
That's also why forcing form: "full" or form: "year" made them look uniform—it simply overrode the style's built-in logic.
See : https://www.chicagomanualofstyle.org/tools_citationguide/citation-guide-1.html#:~:text=Sample%20notes%20show%20full%20citations%20followed%20by%20shortened%20citations%20for%20the%20same%20sources.
###  unclosed parentheses appear
This is indeed a bug. See: https://github.com/typst/hayagriva/issues/138
But it has nothing to do with "citations in footnotes"—it can be reproduced in the main text as well.
###  full citations without the bibliography at the end
The following code might achieve this.
`#show bibliography: none`
`#set cite(form: "full", style: "chicago-notes")`
`#bibliography("bibliography.bib")`

PR #8149, opened 2026-04-21
Fix gradients in LinearRGB, CMYK and Luma spaces for SVG and PDF export 

### LINKED ISSUE #8139: PDF gradients regression
While looking at page 414 in the PDF linked in https://github.com/typst/typst/pull/8138 I noticed the following regression from 0.13 to 0.14, which has seemingly gone unnoticed

```
#set page(height: auto,margin:0pt)
#set text(fill: white, font: "IBM Plex Sans", 8pt)
#set block(spacing: 0pt)
#let spaces = (
  ("Oklab", color.oklab),
  ("Oklch", color.oklch),
  ("sRGB", color.rgb),
  ("linear-RGB", color.linear-rgb),
  ("CMYK", color.cmyk),
  ("Grayscale", color.luma),
  ("HSL", color.hsl),
  ("HSV", color.hsv),
)

#for (name, space) in spaces {
  block(
    width: 100%,
    inset: 4pt,
    fill: gradient.linear(
      red,
      blue,
      space: space,
    ),
    strong(upper(name)),
  )
}
```

PDF export

Image

Linear RGB, CMYK and grayscale are (presumably) incorrect

PNG export (looks the same in SVG as well I believe)
Image

PDF export seems OK in 0.13, so the regression seems to have happened with the transition to krilla in 0.14.

Edit: OKLAB was borked in PDF too, but that seems to have been fixed on main. I've updated the picture to one from a pdf generated by main.

Edit 2: I added the `upstream` label, but I don't know for sure that the issue is krilla itself.

eltos:
Fixes #8139

Changes:
- Use sRGB instead of LinearRgb when determining the intermediate stops, as sRGB is the native color space used by SVG and PDF (this only concerns finding the intermediate stop positions, stop colors always use the gradient space, see below) a25282a6
- Enable the code path which generates intermediate stops in PDF also for LinearRGB. 765115b8
- Convert all stop colors to the target space before writing them to SVG/PDF. This fixes an issue in Luma interpolation, where the first and last stop would be emitted as-is, resulting in colorful edges for the grayscale gradient. 06b236a6
- For PDF export, encode colors in the target space where possible (avoids conversion of CMYK and Luma colors to RGB). 420c4620
- In `sample_stops` make sure the first stop is also converted to the mixing space. This fixes the yellow slice at the start of the Luma gradient not being grayscale in PNG export c02bcfd5
- Add a test containing all the color spaces to replace the dedicated tests for HSV, HSL, Oklch and OKlab. The test suite missed LinearRGB, CMYK and Luma (the ones which regressed), so having a test with all spaces should prevent future regressions 8551fc333fa75e407203bdcb0b1fa88040054beb d6886c1e ff859459
 
The added test before/after this PR:
| PNG | SVG | PDF |
|-|-|-|
|||

The other tests changed because of the fix to interpolate in sRGB instead of LinearRgb, which caused the intermediate stop calculation to return slightly more/less stops depending on the gradient. Visually, they all look identical with one exception:

test `issue-gradient-cmyk-encode` in PDF export (before/after PR):

If looking carefully, you can notice the upper gradient changed slightly. While barely noticeable, if compared to the middle and lower ones, I think that's a fix.


eltos:
I found this comment in the gradient docs which should probably be updated too, but I don't want to mess with #8138:

> ### Note on file sizes
> Gradients can be quite large, especially if they have many stops. This is because gradients are stored as a list of colors and offsets, which can take up a lot of space. If you are concerned about file sizes, you should consider the following:
> - SVG gradients are currently inefficiently encoded. This will be improved in the future.
> - PDF gradients in the color.oklab, color.hsv, color.hsl, and color.oklch color spaces are stored as a list of color.rgb colors with extra stops in between. This avoids needing to encode these color spaces in your PDF file, but it does add extra stops to your gradient, which can increase the file size.



- Drop the first point, was resolved in #7818
- Add LinearRGB to the spaces listed in the second point. ~The comment about additional stops also applies to CMYK, though CMYK colors are stored as such and not converted to RGB if all stops in the original gradient were CMYK colors.~

EDIT: Now that all stop colors are always converted to the gradient's color space, CMYK to RGB conversion is never necessary for PDF export. I updated the PR to always encode CMYK colors as such, which also removes the need for the intermediate stops.

saecki (approved):
Thanks for fixing this! The separate commits made following the changes much easier :)

laurmaedje:
> but I don't want to mess with https://github.com/typst/typst/pull/8138

Don't worry, you can update the docs. I'll rebase my PR before merging and regenerate the commit where I auto-update the docs with the converter. It's no issue.

laurmaedje:
Since you made these nice individual commits, it might make sense to retain them. I'd be happy to fast-forward merge the PR. However, I also want the test suite to pass on every commit that lands in main, so the test changes would have to happen alongside the code changes. If you want to rebase, I'm happy to land it this way. That said, if you don't think it makes sense, I can also just squash the PR.

eltos:
I have updated the note in the gradient docs bd86d35.

You can just squash it, some of the commits are tiny (single line) changes. Good for review, but probably not worth to keep on main for posterity. They are archived here in this PR anyway, just in case.

laurmaedje:
Thanks!

Issue #7319, opened 2025-11-06
Option to not justify math text


amogusussy:
### Description

When `#par(justify:true)` is set, all text, including the math, is justified. This often causes math to break into pieces mid formula.

For example:
<img width="704" height="67" alt="Image" src="https://github.com/user-attachments/assets/90e524b8-5819-4418-be39-d14ac0808e55" />

The `>2^1024` breaks midway through the math, which makes the text harder to read.

Adding an option to not justify/wrap text would make this easier to read:

<img width="640" height="67" alt="Image" src="https://github.com/user-attachments/assets/7ed044ca-72ba-4c07-bf76-d318b538a1c6" />

### Use Case

It would make math blocks easier to read.

Enivex:
This is not really related to justification, but line-breaking in general. In theory you could box inline equations to fix this, but this is currently broken.

<https://github.com/typst/typst/issues/4796>

That being said, I agree that I almost never want inline equations to break.

Issue #493, opened 2023-04-01
Set the baseline of a box to the top text line


lvjr:
```typst
Some text some text Some text some text Some text some text
Some text some text Some text some text Some text some text
#box[One\ Two\ Three]
Some text some text Some text some text Some text some text
Some text some text Some text some text
```
![image](https://user-images.githubusercontent.com/4239949/229258236-b5942b90-9e95-46fc-9e0f-44722704c804.png)

Typst sets the baseline of a box to the bottom text line. But I could not change it to the top text line (like `\vtop` in TeX).

laurmaedje:
We'll add configuration for that in the future.

wisp3rwind:
Relatedly, it might be interesting to have functionality like the LaTeX [xcoffins](https://ctan.org/pkg/xcoffins) project (which gives access to both first and last baseline of box content, and allows arranging boxes based on various handles).

PR #8150, opened 2026-04-22
Expose baselines for box and block

### LINKED ISSUE #4796: Boxing an inline equation shifts the baseline
### Description

```typ
$p/q$ #box($p/q$)
```
image

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

### LINKED ISSUE #5855: When setting text bottom-edge, box gets misaligned
### Description

```
foo #box[foo] foo

#set text(bottom-edge: "descender")

foo #box[foo] foo
```

Image

### Reproduction URL

_No response_

### Operating system

macOS

### Typst version

PgBiel:
Needed for #7895. Fixes #4796 (which might also address https://github.com/typst/typst/issues/7319#issuecomment-3498996342), #5855. Partially addresses https://github.com/typst/typst/issues/493.

Boxes and blocks will now propagate the baseline of what's inside them - usually text or math - allowing them to properly align with their surrounding text / math. For this to work, text and math content now expose their baselines. For paragraphs, the baseline will be set to that under the first line of text. For math equations, the global baseline will be set to the baseline of the first row. And for content with more than one paragraph / equation, the baseline of the first one will be used. With that, several little annoyances when using `box` shall be gone.

This means that:
- Boxed inline equations will no longer be vertically shifted compared to non-boxed equations (#4796). You can now peacefully use `box($a x + b y = 10 - c z$)` to ensure that pesky long inline equation won't break across lines.
- Same for any boxed text whatsoever.
- Which implies you can now add a stroke (or use other `box` effects) around the boxed text, and it will now remain aligned!
    image

For blocks, for now, the only meaningful change is when a `block` is used inside an equation: it will now be properly aligned with the rest. For example, this is how `$ #block($y$) + y $` currently looks like:

image

And this is how it will now look like:

image

In the future (in #7895), block baselines will also be used for list marker alignment.

In addition, `box(scale[Text])` will also use the correct baseline after the scaling **with `reflow: true`**. However, `box(rotate[Text])` will stay as it is today regardless of `reflow: true` as it is not yet very clear where the baseline should go after, say, a 180-degree rotation of a paragraph.

## Configuration

Despite all the goodies listed above, it's possible the old behavior is still desired. In that case, `box(baseline: ...)` now accepts new values (defaulting to `auto`):

- `box(baseline: bottom)` ensures text aligns with `bottom` of the box - **this was the previous default behavior.**
- `box(baseline: top)` and `box(baseline: horizon)` also allow moving the baseline to the top or middle of the box.

Note that `box(baseline: 5pt)` (to shift the baseline by that height compared to where it normally is) is still supported, and both options can be specified simultaneously through a **dictionary:** `box(baseline: (at: bottom, shift: 5pt))`.

This is a **breaking change** if you're reading `context box.baseline` as this will now always return a dictionary.

## TODO

- [x] Address remaining tests.


MDLC01:
Note that https://github.com/typst/typst/issues/5855 is not properly linked as being closed by this PR.

PgBiel:
Fixed, thanks!

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -590,6 +591,18 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
                     let pos = Point::new(x, y);
                     offset += frame.height();
 
+                    // The baseline of the whole region will be the set to the
+                    // baseline of the first in-flow frame. For example, of the
+                    // first paragraph, if there is more than one. But also,
+                    // inside the paragraph itself, this will be the first line
+                    // (since each line is laid out as a separate frame).
+                    if !baseline_set {
```
laurmaedje, line 599: could we use `if !output.has_baseline()` here?

crates/typst-layout/src/math/run.rs
```diff
@@ -241,13 +241,29 @@ pub struct MathRunFrameBuilder {
 
 impl MathRunFrameBuilder {
     /// Consumes the builder and returns a [`Frame`].
-    pub fn build(self) -> Frame {
+    fn build(self, mut set_baseline: bool) -> Frame {
         let mut frame = Frame::soft(self.size);
         for (sub, pos) in self.frames.into_iter() {
+            if set_baseline && sub.has_baseline() {
```
laurmaedje, line 247: same here

crates/typst-layout/src/transforms.rs
```diff
@@ -204,11 +213,41 @@ fn measure_and_layout(
...
+        // For now, however, rotation with reflow, in particular, does not yet
+        // preserve baselines as it's not yet clear that such a baseline change
+        // would be an expected outcome of rotation. (For example, a 180-degree
+        // rotation would cause the baseline to be above the text!)
+        let new_baseline = if frame.has_baseline() && keep_baseline {
+            let p = Point::new(frame.width() / 2.0, frame.baseline()).transform(ts);
+            Some(p.y)
+        } else {
+            None
+        };
```
laurmaedje: just personal preference, but I've kinda grown accustomed to `then` (and sometimes clippy even nudges towards it)
```suggestion
        let new_baseline = (frame.has_baseline() && keep_baseline)
            .then(|| Point::new(frame.width() / 2.0, frame.baseline()).transform(ts).y);
```

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
     /// The height of the box.
     pub height: Smart<Rel<Length>>,
 
-    /// An amount to shift the box's baseline by.
+    /// The vertical position of the box's baseline. This is used to align the
+    /// box with the text surrounding it in a paragraph, as the baseline is
+    /// meant to go right below text by default.
+    ///
+    /// By default, the box's baseline will match the baseline of its contents
+    /// (for example, of the text or equation inside it) - this is the `auto`
```
laurmaedje:
```suggestion
    /// (for example, of the text or equation inside it) - this is the `{auto}`
```

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// option. However, the baseline can be adjusted in two ways. The first
+    /// one is to simply pick a vertical [alignment]($alignment), such as
+    /// `{top}`, `{horizon}` or `{bottom}`, to place the baseline at that
+    /// position, relative to the total height of the box's contents.
```
laurmaedje:
```suggestion
    /// position, relative to the total height of the box.
```
This seems more accurate since inset is also part of the calculation.

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    ///
+    /// The other way to adjust it is to simply shift the default baseline
```
laurmaedje: I think we can drop the "simply"

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// vertically by some amount, specified as a [relative length]($relative).
+    /// For example, the value of `{2pt}` will move it up by that exact
```
laurmaedje:
```suggestion
    /// For example, a value of `{2pt}` will move it up by that exact
```

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// [length]($length) (causing the contents to go _down_, as the align
+    /// point moves _up_), whereas `{-40%}` will shift the baseline _down_ by
```
laurmaedje: "align point" does not feel like a complete term. maybe "alignment point"

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// 40% of the inner contents' total height (thus causing the contents to
```
laurmaedje: here, it's again the question of whether "contents' total height" is accurate because of inset

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// move _up_).
+    ///
+    /// Note that both options can be specified at the same time through a
+    /// dictionary with keys `at` and `shift`, respectively. For example, by
```
laurmaedje:
```suggestion
    /// dictionary with the keys `at` and `shift`, respectively. For example, when
```

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// (for example, of the text or equation inside it) - this is the `auto`
+    /// option. However, the baseline can be adjusted in two ways. The first
+    /// one is to simply pick a vertical [alignment]($alignment), such as
+    /// `{top}`, `{horizon}` or `{bottom}`, to place the baseline at that
+    /// position, relative to the total height of the box's contents.
+    ///
+    /// The other way to adjust it is to simply shift the default baseline
+    /// vertically by some amount, specified as a [relative length]($relative).
+    /// For example, the value of `{2pt}` will move it up by that exact
+    /// [length]($length) (causing the contents to go _down_, as the align
```
laurmaedje: since we already linked "relative length", maybe we can drop this one

crates/typst-library/src/layout/container.rs
```diff
@@ -46,12 +46,34 @@ pub struct BoxElem {
...
+    /// point moves _up_), whereas `{-40%}` will shift the baseline _down_ by
+    /// 40% of the inner contents' total height (thus causing the contents to
+    /// move _up_).
+    ///
+    /// Note that both options can be specified at the same time through a
```
laurmaedje: I think we can also skip the "Note that" without much loss. (Though I also tend to use that phrase a lot...)

crates/typst-library/src/layout/container.rs
```diff
@@ -491,6 +513,81 @@ cast! {
...
+    at: Option<Smart<VAlignment>>,
+    /// Amount to shift the baseline by.
+    shift: Option<Rel<Length>>,
+}
+
+cast! {
+    BaselinePos,
+    self => Value::Dict(self.into()),
+    at: Smart<VAlignment> => Self { at: Some(at), shift: Default::default() },
+    shift: Rel<Length> => Self { at: Default::default(), shift: Some(shift) },
```
laurmaedje: I think `None` would be clearer than `Default::default()` here since we mean unspecified rather than the default value of that field

tests/src/world.rs
```diff
@@ -251,3 +254,31 @@ fn selector_within(selector: LocatableSelector, ancestor: LocatableSelector) ->
         ancestor: Arc::new(ancestor.0),
     }
 }
+
+fn layout_bounds(
```
laurmaedje, line 273: could you move this below `fn bounds`?

tests/src/world.rs
```diff
@@ -251,3 +254,31 @@ fn selector_within(selector: LocatableSelector, ancestor: LocatableSelector) ->
...
+        frame.mark_box_in_place();
+    }
+
+    Ok(fragment)
+}
+
+/// Display boundaries and baseline around some content's frames.
+#[func]
+fn bounds(content: Content) -> SourceResult<Content> {
+    let block = BlockElem::new().with_body(Some(BlockBody::Content(content)));
```
laurmaedje: Using a `BlockElem` to transport the content here is a bit unidiomatic. I would suggest something like this:
```rs
/// Display boundaries around and the baseline of some content's frames.
#[func]
fn bounds(content: Content) -> SourceResult<Content> {
    Ok(BlockElem::multi_layouter(Packed::new(BoundsElem::new(content)), layout_bounds)
        .pack())
}

#[elem]
struct BoundsElem {
    #[positional]
    #[required]
    body: Content,
}

fn layout_bounds(
    elem: &Packed<BoundsElem>,
    engine: &mut Engine,
    locator: Locator,
    styles: StyleChain,
    regions: Regions,
) -> SourceResult<Fragment> {
    let mut fragment = layout_fragment(engine, &elem.body, locator, styles, regions)?;
    for frame in &mut fragment {
        frame.mark_box_in_place();
    }
    Ok(fragment)
}
```

tests/suite/layout/inline/baseline.typ
```diff
@@ -11,6 +11,123 @@ that ever learned to mimic a human voice.
...
+
+#bounds([abc] + box(skew(ax: 30deg, reflow: true)[Hi\ World!]))
+#bounds([abc] + box(skew(ay: 30deg, reflow: true)[Hi\ World!]))
+
+--- baseline-shapes paged ---
+#bounds[abc #box(circle[Hey]) #box(circle())]
+#bounds[abc #box(square[hey!]) #box(square())]
+#bounds[abc #box(ellipse[hey!]) #box(ellipse(width: 2em, height: 1em))]
+
+
```
laurmaedje:
```suggestion
```

tests/suite/layout/container.typ
```diff
@@ -290,6 +290,10 @@ First!
   image("/assets/images/rhino.png", width: 30pt)
 )
 
+--- box-baseline-context paged empty ---
+
```
laurmaedje:
```suggestion
```

laurmaedje: we might want to add a few assertions to check that folding works correctly

laurmaedje (commented):
Looking good! I have some nitpicks, but all minor.

- The test case `block-clip-svg-glyphs` is interesting. The auto baseline is pushed beyond the height of the box because already the first line overflows. I wonder whether we should restrict the baseline to not exceed the box's height. But I'm not sure.

- Do you know what happened for these five tests? The PNG is the same, but PDF and SVG differ slightly:

  - `line-numbers-page-scope-quasi-empty-first-column`
  - `math-equation-number-align-multiline-expand`
  - `math-lr-ignore-ignorant`
  - `show-text-regex-character-class`
  - `spacing-h-and-v`

  I'm not sure whether we discussed these before, I might've forgotten.

- Also, is the TODO "Address remaining tests" still relevant?

crates/typst-layout/src/flow/distribute.rs
PgBiel: That would change the semantics. Here, I'm setting the baseline of `output` to the baseline of the very first frame in it. If the very first frame has no baseline, `output` also will not. In contrast, your change would force `output` to have a baseline if the first N frames did not have a baseline, but the (N + 1)th did.

In particular, for example, if the subflow had two frames - one from a `grid` and one from a paragraph -, the baseline would be set to that of the paragraph below the grid, which would be unexpected.

crates/typst-layout/src/math/run.rs
PgBiel: Here I suppose we could argue similarly as in my other comment, that if the first frame doesn't have a baseline, then the whole equation shouldn't have.

Now, this could be problematic if this is done within the equation itself. For example, if we placed some "baseline-less" object inside math for some reason, such as a grid, we could still want the equation to use the baseline of something after this grid.

But that's not a problem here as the frames in this case are always entire rows, as documented in the definition of the builder type. So this will only discard the baseline of the whole equation if the first row doesn't have a baseline, which would make sense. (However, I doubt it would ever not have a baseline... but, at least in isolation, this seems sound, and is better than the alternative of potentially moving the baseline to the second row in weird edge cases.)

Maybe adding this observation as a comment could be nice.

crates/typst-layout/src/transforms.rs
PgBiel: This feels a bit less explicit to me, but that's fair enough.

crates/typst-library/src/layout/container.rs
PgBiel: Right, this was a copy-paste from `FirstLineIndent` for par. Will fix

crates/typst-layout/src/flow/distribute.rs
laurmaedje: Fair enough!

laurmaedje:
Would be great if you could also rebase on top of main since there are conflicts.

PgBiel:
> The test case block-clip-svg-glyphs is interesting. The auto baseline is pushed beyond the height of the box because already the first line overflows. I wonder whether we should restrict the baseline to not exceed the box's height. But I'm not sure.

Yeah, that's an interesting question. I suppose we generally can't prevent the baseline from moving out of bounds... But your proposal does make some sense, if we consider that baseline is strictly a property of layout, not of render or purely visual. Worth thinking a bit about.

> Do you know what happened for these five tests? 

`line-numbers-page-scope-quasi-empty-first-column` was already discussed - it's unrelated to the baseline changes, more related to how line numbers themselves work, so I accepted it. (I was planning on open an issue to discuss this more deeply.)

The core problem here is that the `text` inside `#box[text]` is considered a separate line of text, so `Hello #box(baseline: -5pt)[world]` generates two separate line numbers. It might be worth thinking about whether we should restrict line numbers to semantic `par` by default - although note that it is already possible today with `#show par: set par.line(numbering: "1.")`. (I believe my work on line numbers predated semantic paragraphs, otherwise this might have been discussed earlier.)

For the rest, I had found that the changes were mostly in subtle position changes of empty shapes, likely spaces, so I don't think anything major was spotted there. In any case, I'll take a second look to confirm.

> Also, is the TODO "Address remaining tests" still relevant?

No.

PgBiel:
> `math-equation-number-align-multiline-expand`
> `math-lr-ignore-ignorant`
> `show-text-regex-character-class`
> `spacing-h-and-v`

For all of these, seems like only empty groups were affected, as seen below. I believe these were minor changes in the spacing due to the baseline update.

With that, I think we're only missing the decision on `block-clip-svg-glyphs`.

image

laurmaedje:
I tested what CSS does:

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <style>
      div {
        display: flex;
        align-items: baseline;
        gap: 1em;
      }

      span {
        border: 1pt solid red;
      }

      .b {
        font-size: 2em;
        height: 10px;
      }
    </style>
  </head>
  <body>
    <div>
      <span class="a">A</span>
      <span class="b">B</span>
    </div>
  </body>
</html>
```

grafik

It seems like the current behavior is consistent with CSS. I'd be inclined to leave it as-is. It also requires less special casing in the code.

laurmaedje:
Nice to have this merged. One step closer to having the list issues fixed, too. :)

Issue #8148, opened 2026-04-21
calc.norm is prone to overflow


Enivex:
### Description

<img width="1180" height="152" alt="Image" src="https://github.com/user-attachments/assets/1c5aa9b5-33d5-4448-821b-8a09958ab9ca" />

This can easily be circumvented by using a strategy like the one described in <https://timvieira.github.io/blog/numerically-stable-p-norms/>

### Typst version

- [x] I am using the latest version of Typst

aaron-jack-manning:
This can be closed now.

PR #8156, opened 2026-04-22
Fix overflow of `norm` in certain ranges

aaron-jack-manning:
Fix for #8148.

### DIFF
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -1153,9 +1153,7 @@ pub fn norm(
         bail!(p.span, "p must be greater than zero");
     }
 
-    // Create an iterator over the absolute values.
     let abs = values.iter().map(|v| f64::abs(*v));
-
     let max = abs.clone().max_by(|a, b| a.total_cmp(b)).unwrap_or(0.0);
 
     Ok(if p.v.is_infinite() {
@@ -1166,8 +1164,9 @@ pub fn norm(
         // This is a special case to avoid division by zero in the trick below.
         0.0
     } else {
-        // Compute `max * (sum_i (x_i / max)^p)^(1 / p)` instead of raising `x_i^p` directly, to avoid overflowing.
-        // This is described further in https://timvieira.github.io/blog/numerically-stable-p-norms/
+        // Compute `max * (sum_i (x_i / max)^p)^(1 / p)` instead of raising
+        // `x_i^p` directly, to avoid overflowing. This is described further in
+        // <https://timvieira.github.io/blog/numerically-stable-p-norms/>.
         max * libm::pow(abs.map(|v| libm::pow(v / max, p.v)).sum::<f64>(), 1.0 / p.v)
     })
 }
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -1154,13 +1154,21 @@ pub fn norm(
     }
 
     // Create an iterator over the absolute values.
-    let abs = values.into_iter().map(f64::abs);
+    let abs = values.iter().map(|v| f64::abs(*v));
+
+    let max = abs.clone().max_by(|a, b| a.total_cmp(b)).unwrap_or(0.0);
 
     Ok(if p.v.is_infinite() {
         // When p is infinity, the p-norm is the maximum of the absolute values.
-        abs.max_by(|a, b| a.total_cmp(b)).unwrap_or(0.0)
+        max
+    } else if max == 0.0 {
+        // When the maximum absolute value is 0, the norm is just 0.
+        // This is a special case to avoid division by zero in the trick below.
+        0.0
     } else {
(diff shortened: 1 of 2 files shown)


Enivex:
You should probably add some tests

saecki:
One problem that should be considered, is that we lose precision. I played around a little with the tests:
```typ
#test(calc.norm(p: 10, 1210, 10, 20), 1210)
```
This will produce `1210.0000000000005` on `main` but `1210` with this PR.

I'm honestly not sure what's more important here, but I think that should be discussed before making a change. Maybe we can somehow retain the precision for smaller numbers and use the stable p-norm for larger cases?

Enivex:
> One problem that should be considered, is that we lose precision. I played around a little with the tests:
> ```typ
> #test(calc.norm(p: 10, 1210, 10, 20), 1210)
> ```
> This will produce `1210.0000000000005` on `main` but `1210` with this PR.
> 
> I'm honestly not sure what's more important here, but I think that should be discussed before making a change. Maybe we can somehow retain the precision for smaller numbers and use the stable p-norm for larger cases?

The real norm is 1210.0000000000000001843552... so the PR is closer to being correct here.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -1159,11 +1159,17 @@ pub fn norm(
     // Create an iterator over the absolute values.
     let abs = values.into_iter().map(f64::abs);
 
+    let max = abs.clone().max_by(|a, b| a.total_cmp(b)).unwrap_or(0.0);
```
bluss, line 1157: don't know if it's a focus here, but abs is an IntoIter, so cloning the iterator clones the vector in this case. Would be nicer to use `abs = values.iter().map(|elt| f64::abs(*elt))` as the iterator, which can be cloned cheaply? That's my old Rust intuition for this.

aaron-jack-manning: You are right, it is better that way.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -1157,13 +1157,19 @@ pub fn norm(
     }
 
     // Create an iterator over the absolute values.
-    let abs = values.into_iter().map(f64::abs);
+    let abs = values.iter().map(|elt| f64::abs(*elt));
```
saecki: Just a nitpick, but I was shortly confused where `elt` came frome. I'd just call it `v` as below.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -1157,13 +1157,19 @@ pub fn norm(
...
+
+    let max = abs.clone().max_by(|a, b| a.total_cmp(b)).unwrap_or(0.0);
 
     Ok(if p.v.is_infinite() {
         // When p is infinity, the p-norm is the maximum of the absolute values.
-        abs.max_by(|a, b| a.total_cmp(b)).unwrap_or(0.0)
+        max
+    } else if max == 0.0 {
+        0.0
     } else {
```
saecki, line 1166: Maybe we can link the article from the issue here, to have a little more context.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -1157,13 +1157,19 @@ pub fn norm(
...
-        libm::pow(abs.map(|v| libm::pow(v, p.v)).sum::<f64>(), 1.0 / p.v)
+        // We compute: max * (sum_i (x_i / max)^p)^(1 / p). This is to reduce
+        // the range of overflow compared to computing x_i^p directly.
```
saecki: Maybe something like this would be clearer:
```suggestion
        // Compute `max * (sum_i (x_i / max)^p)^(1 / p)` instead of raising
        // `x_i^p` directly, to avoid overflowing.
```

saecki (commented):
I tried to find instances where a similar approach is used, but couldn't find anything.
I've also found the claim the article makes about `scipy` to be false: https://timvieira.github.io/blog/numerically-stable-p-norms/#:~:text=Remarks
Even looking back to 2014 `scipy` simply forwarded to `numpy` for non-special values of `p`.

Though I think my initial skepticism was a bit ill-founded, and I feel more comfortable with this approach now.

Here are a couple of remarks regarding the changes.

Enivex:
I don't know how much we care about performance here, but a potential alternative to dividing by the exact maximum is dividing by the closest power of 2.

It would have to be a very contrived document for this to actually have any bearing on compilation time though. It's not like typst is being used for HPC

Edit: Upon further reflection this would just reintroduce overflow, but to a somewhat lesser extent.

saecki:
> I don't know how much we care about performance here, but a potential alternative to dividing by the exact maximum is dividing by the closest power of 2.
>
> It would have to be a very contrived document for this to actually have any bearing on compilation time though. It's not like typst is being used for HPC
>
> Edit: Upon further reflection this would just reintroduce overflow, but to a somewhat lesser extent.

Definitely an interesting idea, but yeah it would introduce another edge case.

saecki (approved):

laurmaedje:
Thanks!

PR #8164, opened 2026-04-24
Bump New CM fonts to version 8.0.0

### LINKED ISSUE #7887: Potential accent misalignment in subscript
### Description

When putting an accent on a subscript the alignment changes as follows:

`
$hat(G) quad  m_hat(G)$ 
`

Image

To me this seems like undesired behavior, i.e. in the script style G the accent seems to attach too far to the right.

The following also showcases this behavior:

`$hat(G) script(hat(G))  sscript(hat(G))$`

Image

### Suggestion
A quick fix is to ignore top accent attachment points for font sizes <= script size and use the existing centered alignment instead as fallback.
However, the result also seems suboptimal:

Image

That being said, I am not sure what the desired behavior is but I think it would be nice to have accents "look the same" for all the different font sizes.

### Typst version

mkorje:
Closes #7887. (Please let me know if I missed any other issues.) Depends on https://github.com/typst/typst-assets/pull/18.

A _lot_ of tests have been updated: most are pdf and svg tests, where glyph ids appear to have changed slightly (no visual changes though). Visual changes (determined based on a change in the png tests) occurred only in the following tests:
- Changes due to "latin letters in NewCM10-Regular and accented and built composites re-traced from public metafont sources and hand adjusted glyph by glyph"
  - enum-numbering-closure-nested-complex
  - justify-knuth-story
  - math-op-scripts-vs-limits
  - math-op-set-font
  - script-metrics-bundled-fonts
  - shorthand-nbsp-width
  - show-text-regex
  - text-call-body
  - transform-tex-logo
- Changes due to "Knuthian Calligraphic capitals have been restored as the default calligraphic capitals in Regular, Book and Bold Math fonts"
  - math-font-features-switch
  - math-style
  - math-style-cluster
  - math-style-exceptions
  - math-style-script
  - math-text-single-grapheme-cluster
- The glyphs for `shell.l.stroked` and `shell.l.filled` (and the `.r` variants) have changed.
  - math-lr-sym-call
- I do not know what's happened here. I can't see any visual difference in any of png, svg, pdf. Looking at the svg source though, it appears the paths for some glyphs has changed slightly.
  - math-attach-followed-by-func-call
  - math-frac-line-fill-stroke


laurmaedje:
Thanks!

PR #8167, opened 2026-04-25
Optimize `SyntaxNode::into_text` and `SyntaxNode::errors` methods with `traverse` helper

isuffix:
As far as I can tell, we weren't using `.collect()` for any principled reason, so these methods should be more straightforwardly performant.

I did keep the optimization of `.into_text()` returning the leaf `EcoString` without cloning, but I don't think it really matters and it may make sense to just return `String` instead.


isuffix:
Sorry, uploaded with a wrong commit.

PR #8171, opened 2026-04-26
Retry stretching in math without `flac` and `ssty` features

### LINKED ISSUE #6660: Delimiters do not stretch correctly in script size for some math fonts
### Description

When writing delimiters in script size, they do not stretch when using some fonts, particularly Noto Sans Math and STIX Two Math, as shown:

Image

At first, I thought it may be with the fonts just not supporting stretchable delimiters in script size. However, when I tried it in LaTeX with the same fonts, the delimiters (here being parentheses) are stretched properly:

Image

(As shown are STIX Two Math and Noto Sans Math, respectively.) I tried using `math.lr` function with the `size` parameter to manually adjust the size of the delimiters, but that seems to do nothing in script size using those fonts.

### Reproduction URL

https://typst.app/project/rKKdGPpFwa2kL0tyQoI9c9

### Operating system

Linux, Web app

### Typst version

### LINKED ISSUE #7267: Some stretchable accents in Stix Two Math broke between 0.12 and 0.13
### Description

0.12:

Image

0.13 (and 0.14)

Image

As far as I can tell this did not happen with other fonts, but I do not know whether this is a typst issue or a font issue.

### Reproduction URL

_No response_

### Operating system

_No response_

### Typst version

### LINKED ISSUE #8157: Incorrect radical sizing for `sqrt` with STIX Two Math for tall content in super- and subscripts
## Behavior

The following code produces an unexpected result when STIX Two Math is used:

```typ
#let test = $ lr(a / b|)_sqrt(c / d) $

#test

#show math.equation: set text(font: "STIX Two Math")
#test
````

Output:

Image

The square root in the subscript is rendered too short: the denominator of the fraction extends below the radical.

The same issue also occurs with other tall content, such as `lr(a / b|)_sqrt(display(sum))`, and it also appears in superscripts.

The formula looks fine with New Computer Modern Math.

I could not find an existing issue describing the same behavior.

I could not determine whether this is a font issue or a Typst bug, but LaTeX produces the expected output, as shown below.

## Comparison with LaTeX

The equivalent LaTeX code produces the expected result, both with LuaLaTeX and XeLaTeX:

```tex
\documentclass{article}
\usepackage{unicode-math}

\begin{document}
$$
    \left. \frac{a}{b} \right|_{\sqrt{\frac{c}{d}}}
$$

{
\setmathfont{STIX Two Math}
$$
    \left. \frac{a}{b} \right|_{\sqrt{\frac{c}{d}}}
$$
}
\end{document}
```

Output:

Image

## Tested configurations

I reproduced this unexpected behavior with all of the following:

- Output formats: PDF, PNG, SVG
- Font distributions:
  - [STIX Two Math 2.13 b171 (latest non-interim, OTF)](https://github.com/stipub/stixfonts/blob/c4afdf3fa5390159ef24aca1db5e957487c23897/fonts/static_otf/STIXTwoMath-Regular.otf)
  - STIX Two Math 2.12 b168 (bundled with my MacTeX distribution, OTF)
- Typst distributions:
  - Typst 0.14.2
  - Current `main`: https://github.com/typst/typst/commit/78251637643fc4c8fbc8ec174dd622ff23e06861

## MRE WebApp URL

https://typst.app/project/RtWopRndyXijyXikW19k6W

### Operating system

macOS

### Typst version

mkorje:
Closes #8157, closes #7267, closes #6660.

The problem in all the above issues was that some fonts give stretchy glyph's a `ssty`/`flac` variant but don't provide the ability for them to stretch in the MATH table. Instead of just not applying `ssty`/`flac` to things that might be stretched, we now perform a fallback if we find the resulting glyph can't be stretched but we need it to. That way, if a font does supply constructions for glyph's variants we can use them, and if not, we only don't use the exact variant if we can't reach the requested stretch length.

~~NOTE: Before merging this we should do a benchmark run, since the `GlyphFragment` creation logic has changed a fair bit (in particular, its memoization).~~ Performance is no issue, ~8% faster than main on the math heavy test (test run 200, compared with 196). Not too shabby.


crates/typst-layout/src/math/fragment/glyph.rs
```diff
@@ -242,40 +243,128 @@ pub struct GlyphFragment {
...
+            class,
+            math_size,
+        )
+        .map(|glyph| glyph.with_span(span))
     }
 
-    /// Selects a font to use and then shapes text.
-    #[comemo::memoize]
+    /// Creates a glyph fragment from the given text, stretching it if needed.
     pub fn new(
```
laurmaedje, line 67: Not _really_ related to the PR I guess, but it feels a bit odd that `new` and `new_char` take such different sets of parameters. It seems like they should only differ in char vs string.

crates/typst-layout/src/math/shaping.rs
```diff
@@ -30,24 +54,76 @@ pub fn shape(
...
+pub fn fallback<F>(mut features: Vec<Feature>, mut retry: F)
+where
+    F: FnMut(&[Feature]) -> bool,
+{
+    const FLAC: Tag = Tag::from_bytes(b"flac");
+    const SSTY: Tag = Tag::from_bytes(b"ssty");
+
+    // (flac, ssty) combinations to try.
+    const OPTIONS: [(bool, u32); 6] =
+        [(true, 2), (true, 1), (true, 0), (false, 2), (false, 1), (false, 0)];
```
laurmaedje, line 75: Why exactly `0, 1, 2`? Are these all possible ones or the commonly used ones? If it's the latter, it feels a bit unfortunate to hard code this.

crates/typst-layout/src/math/fragment.rs
```diff
@@ -478,6 +551,19 @@ impl Debug for GlyphFragment {
...
+
+impl PlannedGlyph {
+    fn with_span(mut self, span: Span) -> Self {
+        self.glyph = self.glyph.with_span(span);
+        self
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct FrameFragment {
```
laurmaedje: The `FrameFragment` is kinda lost in the middle here. I think we should consider reorganizing the file or splitting it up across glyph and frame or moving the stretching logic into a separate one.

crates/typst-layout/src/math/shaping.rs
mkorje: _Technically_, I think others can be used, as per the [spec](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#ssty):
>  This feature can have a parameter indicating the script level: 1 for simple subscripts and superscripts, 2 for second level subscripts and superscripts (that is, scripts on scripts), and so on. (Currently, only the first two alternates are used).

But for all purposes these are the only ones (0 here represents disabling it). We don't provide any way to use further levels anyways (unless a user sets it explicitly with `text.features`, I guess).

crates/typst-layout/src/math/fragment.rs
mkorje: Moving all the glyph stuff to a `glyph.rs` feels clean. But then leaving the `FrameFragment` and `MathFragment` in `fragment.rs` is maybe a little odd?

crates/typst-layout/src/math/fragment/glyph.rs
mkorje: I think it was the same even before when they were more similar, I just needed a decent name and couldn't come up with anything. `new_char` is used in three places:
- `text.rs`: shortcut for creating a number's glyph
- `scripts.rs`: creating a prime glyph
- `table.rs`: creating a parenthesis glyph to use a padding for table cell heights.

Any ideas for a better name?

crates/typst-layout/src/math/fragment.rs
laurmaedje: > But then leaving the FrameFragment and MathFragment in fragment.rs is maybe a little odd?

I think it's okay since they are smaller.

crates/typst-layout/src/math/shaping.rs
laurmaedje: Ah, okay, that makes sense. I misunderstood the purpose of the feature.

crates/typst-layout/src/math/fragment/glyph.rs
laurmaedje: Maybe `GlyphFragment::synthetic` since it doesn't come from a text or symbol element?

laurmaedje: Aligning the use of MathContext vs Engine across the two could still be nice.

crates/typst-layout/src/math/fragment.rs
mkorje: Would you be happy with making `fragment` its own small module `fragment/mod.rs` so that `glyph.rs` is still "owned" by it? Or would you rather keep it all at the top level?

laurmaedje: I'm okay with having a `fragment` folder

crates/typst-layout/src/math/fragment/glyph.rs
mkorje: > Aligning the use of MathContext vs Engine across the two could still be nice.

Sure. `new` changed just for memoization reasons, IIRC.

crates/typst-layout/src/math/shaping.rs
mkorje: We could add two enums for the flac and ssty possible values?

laurmaedje: I think it's fine, but a comment would be nice

mkorje: Added.

laurmaedje:
If you rebase (since this will conflict with the `math.class` changes), would you mind squashing into 2+ commits where the first commit cleanly pulls out the submodule and the rest contains the logic changes? The rest all being one commit is totally fine by me (though the `GlyphFragment` changes could also cleanly remain separate), but it would be nice to have the module split come first and be separated.

crates/typst-layout/src/math/shaping.rs
```diff
@@ -30,24 +54,78 @@ pub fn shape(
...
 
-    shape_impl(&mut ctx, text, families.into_iter());
+    shape_text(&mut ctx, text, families.into_iter());
 
     Some((ctx.font?, ctx.glyphs))
 }
 
+/// Calls `retry` with feature slices that turn `flac`/`ssty` on or off in
+/// different combinations until the closure returns `true`.
+pub fn fallback<F>(mut features: Vec<Feature>, mut retry: F)
```
laurmaedje: The name `fallback` is perhaps a bit generic

crates/typst-layout/src/math/shaping.rs
```diff
@@ -30,24 +54,78 @@ pub fn shape(
...
+    // used, with 0 here representing disabling the ssty feature.
+    const OPTIONS: [(bool, u32); 6] =
+        [(true, 2), (true, 1), (true, 0), (false, 2), (false, 1), (false, 0)];
+
+    let had_flac = features
+        .iter()
+        .rev()
+        .find(|f| f.tag == FLAC)
+        .is_some_and(|f| f.value != 0);
+    let had_ssty = features
```
laurmaedje: The name sounds like it's a bool, but it's not

mkorje:
> If you rebase (since this will conflict with the `math.class` changes), would you mind squashing into 2+ commits where the first commit cleanly pulls out the submodule and the rest contains the logic changes? The rest all being one commit is totally fine by me (though the `GlyphFragment` changes could also cleanly remain separate), but it would be nice to have the module split come first and be separated.

Sure, will do in a couple mins.

crates/typst-layout/src/math/shaping.rs
mkorje: Maybe `feat_fallback` then?

crates/typst-layout/src/math/shaping.rs
mkorje: I renamed it to `prev_ssty`.

mkorje:
Rebasing was slightly painful :sweat_smile:. I applied the name changes I suggested in response to your latest comments. I'm going to sleep so feel free to fix things up yourself so it can get merged.

laurmaedje:
> Rebasing was slightly painful 😅

I'm sorry. 🙈

Thanks for taking care of it! And for the PR. :)

PR #8172, opened 2026-04-26
Migrate from `rustybuzz` and `ttf-parser` to `harfrust` and `fontations`

mkorje:
An attempt at doing this migration. Other PRs: https://github.com/typst/pixglyph/pull/4.

The first (and only, atm) commit does the smallest change possible needed to replace `rustybuzz` with `harfrust`. The second commit will try to replace as much of `ttf-parser` with `read-fonts` and/or `skrifa` (tbd).

Note that we won't be able to get rid of `ttf-parser` completely just yet, since `read-fonts` lacks support for the MATH table (see https://github.com/googlefonts/fontations/pull/1269).

We may also want to wait for `resvg` to be ported (see https://github.com/linebender/resvg/pull/922).

Performance/binary size analysis needs to be done.

Similar in spirit to these changes, we might want to investigate migrating to `fontique` from `fontdb` (see the discussion in https://xi.zulipchat.com/#narrow/channel/465085-resvg/topic/Moving.20resvg.20to.20fontations.2Fharfrust/with/572344735).

### DIFF
--- a/crates/typst-layout/src/inline/shaping.rs
+++ b/crates/typst-layout/src/inline/shaping.rs
@@ -5,8 +5,8 @@ use std::sync::Arc;
 
 use az::SaturatingAs;
 use comemo::Tracked;
-use rustybuzz::{BufferFlags, Feature, ShapePlan, UnicodeBuffer};
-use ttf_parser::Tag;
+use harfrust::{BufferFlags, Feature, ShapePlan, UnicodeBuffer};
+use read_fonts::types::Tag;
 use ttf_parser::gsub::SubstitutionSubtable;
 use typst_library::World;
 use typst_library::engine::Engine;
@@ -835,7 +835,7 @@ struct ShapingContext<'a> {
     styles: StyleChain<'a>,
     size: Abs,
     variant: FontVariant,
-    features: Vec<rustybuzz::Feature>,
+    features: Vec<Feature>,
     fallback: bool,
     dir: Dir,
     shift_settings: Option<ShiftSettings>,
@@ -958,13 +958,13 @@ fn shape_segment<'a>(
     buffer.push_str(text);
     buffer.set_language(language(ctx.styles));
     if let Some(script) = ctx.styles.get(TextElem::script).custom().and_then(|script| {
-        rustybuzz::Script::from_iso15924_tag(Tag::from_bytes(script.as_bytes()))
+        harfrust::Script::from_iso15924_tag(Tag::new(script.as_bytes()))
     }) {
         buffer.set_script(script)
     }
     buffer.set_direction(match ctx.dir {
-        Dir::LTR => rustybuzz::Direction::LeftToRight,
-        Dir::RTL => rustybuzz::Direction::RightToLeft,
+        Dir::LTR => harfrust::Direction::LeftToRight,
+        Dir::RTL => harfrust::Direction::RightToLeft,
         _ => unimplemented!("vertical text layout"),
     });
     buffer.guess_segment_properties();
@@ -997,15 +997,15 @@ fn shape_segment<'a>(
         &ctx.features,
     );
 
-    if has_shift_feature {
-        ctx.features.pop();
-    }
-
     // Shape!
-    let buffer = rustybuzz::shape_with_plan(font.rusty(), &plan, buffer);
+    let buffer = font.shaper().shape_with_plan(&plan, buffer, &ctx.features);
     let infos = buffer.glyph_infos();
     let pos = buffer.glyph_positions();
     let ltr = ctx.dir.is_positive();
+
+    if has_shift_feature {
+        ctx.features.pop();
+    }
 
     // Whether the character at the given offset is covered by the coverage.
     let is_covered = |offset| {
@@ -1155,11 +1155,16 @@ fn determine_shift(
             // OpenType feature instead of synthesizing if possible), we add
             // "subs"/"sups" to the feature list if supported by the font.
             // In case of a problem, we just early exit
-            let gsub = font.rusty().tables().gsub?;
-            let lookups = gsub.features.find(settings.kind.feature())?.lookup_indices;
+            let gsub = font.ttf().tables().gsub?;
+            let lookups = gsub
+                .features
+                .find(ttf_parser::Tag::from_bytes(
+                    &settings.kind.feature().to_be_bytes(),
+                ))?
+                .lookup_indices;
             text.chars()
                 .all(|c| {
-                    let Some(i) = font.rusty().glyph_index(c) else { return false };
+                    let Some(i) = font.ttf().glyph_index(c) else { return false };
                     lookups
                         .into_iter()
                         .flat_map(|i| gsub.lookups.get(i))
@@ -1197,13 +1202,13 @@ fn determine_shift(
 #[comemo::memoize]
 pub fn create_shape_plan(
     font: &Font,
-    direction: rustybuzz::Direction,
-    script: rustybuzz::Script,
-    language: Option<&rustybuzz::Language>,
-    features: &[rustybuzz::Feature],
+    direction: harfrust::Direction,
+    script: harfrust::Script,
+    language: Option<&harfrust::Language>,
+    features: &[Feature],
 ) -> Arc<ShapePlan> {
-    Arc::new(rustybuzz::ShapePlan::new(
-        font.rusty(),
+    Arc::new(harfrust::ShapePlan::new(
+        &font.shaper(),
         direction,
         Some(script),
         language,
--- a/crates/typst-library/src/text/font/color.rs
+++ b/crates/typst-library/src/text/font/color.rs
@@ -16,7 +16,8 @@ use crate::visualize::{
 
 /// Whether this glyph should be rendered via simple outlining instead of via
 /// `glyph_frame`.
-pub fn should_outline(font: &Font, glyph_id: GlyphId) -> bool {
+pub fn should_outline(font: &Font, id: u16) -> bool {
+    let glyph_id = GlyphId(id);
     let ttf = font.ttf();
     (ttf.tables().glyf.is_some() || ttf.tables().cff.is_some())
         && !ttf
--- a/crates/typst-library/src/text/font/mod.rs
+++ b/crates/typst-library/src/text/font/mod.rs
@@ -41,12 +41,12 @@ struct FontInner {
     ttf: ttf_parser::Face<'static>,
     /// The underlying fontations face.
     fontations: read_fonts::FontRef<'static>,
-    /// The underlying rustybuzz face.
-    rusty: rustybuzz::Face<'static>,
-    // NOTE: `ttf` and `rusty` reference `data`, so it's important for `data`
-    // to be dropped after them or they will be left dangling while they're
-    // dropped. Fields are dropped in declaration order, so `data` needs to be
-    // declared after `ttf` and `rusty`.
+    /// Cached shaper data for the font.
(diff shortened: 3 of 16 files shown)


crates/typst-library/src/text/font/mod.rs
```diff
@@ -141,11 +152,21 @@ impl Font {
...
+    fn fontations(&self) -> &read_fonts::FontRef<'_> {
         // We can't implement Deref because that would leak the
         // internal 'static lifetime.
-        &self.0.rusty
+        &self.0.fontations
+    }
+
+    /// Get a configured shaper for the font.
+    pub fn shaper(&self) -> harfrust::Shaper<'_> {
+        // RESOLVE: Should this be cached?
```
mkorje, line 164: Since variable fonts aren't supported and we don't set the point size (Should we? We weren't before, but that might have just been a limitation of `rustybuzz`.), maybe this should be stored directly in `FontInner`.

Enivex:
The resvg PR has been open for nearly a year, so I wouldn't hold my breath

laurmaedje:
> The resvg PR has been open for nearly a year, so I wouldn't hold my breath

We aren't. We're just switching anyway.

Enivex:
https://github.com/harfbuzz/harfrust/releases/tag/0.8.4

Enivex:
Looks like there's finally movement in https://github.com/linebender/resvg/pull/922

Enivex:
https://github.com/linebender/resvg/pull/922 has now been merged

Edit: The 0.48 release includes this change

dfrg:
> Note that we won't be able to get rid of ttf-parser completely just yet, since read-fonts lacks support for the MATH table (see https://github.com/googlefonts/fontations/pull/1269).

Finally coming in https://github.com/googlefonts/fontations/pull/2105

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -1176,18 +1178,21 @@ fn determine_shift(
...
-                            lookup.subtables.into_iter::<SubstitutionSubtable>()
-                        })
-                        .any(|subtable| subtable.coverage().contains(i))
+                    let Some(i) = font.glyph_index(c) else { return false };
+                    glyph.clear();
+                    glyph.insert(i);
+                    let mut reachable = indices.clone();
+                    gsub.closure_lookups(&glyph, &mut reachable).is_ok()
+                        && !reachable.is_empty()
                 })
```
mkorje, line 1196: This is suboptimal to say the least...

crates/typst-library/src/text/font/color.rs
mkorje, line 1: Most of things I noted from in the [pixglyph PR](https://github.com/typst/pixglyph/pull/4#issuecomment-5727889210) apply here too, and in `typst-render` and `typst-svg`.

crates/typst-library/src/text/font/color.rs
```diff
@@ -360,101 +374,151 @@ fn fixup_svg(svg: &mut String, view_box: Rect) {
     }
 }
 
-struct ColrBuilder<'a>(&'a mut String);
```
mkorje, line 363: Everything from this point onward was taken from `resvg` (with minor modifications): https://github.com/linebender/resvg/pull/922. Need to properly attribute it.

crates/typst-library/src/text/font/color.rs
```diff
@@ -586,30 +690,106 @@ impl GlyphPainter<'_> {
...
+        let outlined = if let Some(outliner) = self.font.outline_glyphs().get(glyph_id) {
+            let mut builder = ColrBuilder::new(self.path_buf);
+            let size = skrifa::instance::Size::unscaled();
+            let ok = outliner
+                .draw(
+                    skrifa::outline::DrawSettings::unhinted(size, self.location)
+                        .with_path_style(skrifa::outline::pen::PathStyle::HarfBuzz),
+                    &mut builder,
+                )
+                .is_ok();
```
mkorje, line 711: This is the only place where outlining happens outside of the method on `FontInstance`. Would be good to make this use that somehow so things are consistent w.r.t. the path style.

crates/typst-svg/src/paint.rs
```diff
@@ -232,7 +232,7 @@ impl SVGRenderer<'_> {
                         // Add the path to the pattern.
                         pattern
                             .elem("path")
-                            .attr("d", builder.finsish())
+                            .attr("d", builder.finish())
```
mkorje, line 235: Surely this spelling mistake isn't intentional. I decided to fix it.

crates/typst-layout/src/inline/shaping.rs
```diff
@@ -1218,23 +1223,17 @@ fn determine_shift(
...
 ) -> Arc<ShapePlan> {
-    Arc::new(rustybuzz::ShapePlan::new(
-        font.rusty(),
-        direction,
-        Some(script),
-        language,
-        features,
-    ))
+    Arc::new(ShapePlan::new(font.shaper(), direction, Some(script), language, features))
 }
```
mkorje, line 1232: Maybe we return a `ShapeOptions` directly (which includes a `ShapePlane`) instead?

crates/typst-library/src/text/font/info.rs
```diff
@@ -102,51 +100,48 @@ impl FontInfo {
...
-            }
+        for (c, _) in skrifa.charmap().mappings() {
+            codepoints.push(c);
         }
+        // TODO: less comprehensive as before...
+        // for subtable in ttf.tables().cmap.into_iter().flat_map(|table| table.subtables) {
+        //     if subtable.is_unicode() {
+        //         subtable.codepoints(|c| codepoints.push(c));
+        //     }
+        // }
```
mkorje: Walking all the tables like `ttf-parser` was doing is slightly painful with `read-fonts`, so I haven't added that yet. What `skrifa`'s `charmap` does is just pick one "best" subtable.

crates/typst-library/src/text/font/color.rs
```diff
@@ -498,26 +560,37 @@ impl<'a> GlyphPainter<'a> {
...
+        self.svg.write_attribute("cx", &c1.x);
+        self.svg.write_attribute("cy", &c1.y);
+        self.svg.write_attribute("r", &r1);
+        self.svg.write_attribute("fr", &r0);
+        self.svg.write_attribute("fx", &c0.x);
+        self.svg.write_attribute("fy", &c0.y);
         self.svg.write_attribute("gradientUnits", &"userSpaceOnUse");
-        self.write_spread_method_attribute(gradient.extend);
+        self.write_spread_method_attribute(extend);
         self.write_transform_attribute("gradientTransform", gradient_transform);
```
mkorje, line 592: gpt-6 astra flagged this on review with the following comment. I have no idea about this part of the code (I took it from `resvg` after all).

> Skrifa can supply negative radii when normalizing valid COLR color stops. I reproduced `r0 = -256`, which this code writes as `fr="-256"`. Usvg clamps that to zero without adjusting the stops, changing the gradient’s colors and geometry. The color line needs truncation and reparameterization before SVG serialization.

crates/typst-library/src/text/font/info.rs
mkorje: As mentioned in another issue though, the coverage shouldn't just include single chars, but also clusters (i.e. a char and a variation selector).

crates/typst-library/src/text/font/metrics.rs
```diff
@@ -33,27 +36,69 @@ pub struct FontMetrics {
...
+                .map(|os2| {
+                    let mut ascent = os2.s_typo_ascender() as f32;
+                    if let (Ok(mvar), true) = (skrifa.mvar(), !coords.is_empty()) {
+                        use skrifa::raw::tables::mvar::tags::*;
+                        let metric_delta = |tag| {
+                            mvar.metric_delta(tag, coords).unwrap_or_default().to_f64()
+                                as f32
+                        };
+                        ascent += metric_delta(HASC);
+                    }
```
mkorje, line 62: `skrifa` only returns the OS/2 ascender/descender values if USE_TYPO_METRICS is set by the font, but before we were taking these unconditionally. This is also ugly and can certainly be made nicer.

crates/typst-library/src/text/font/mod.rs
```diff
@@ -220,33 +248,44 @@ impl FontInstance {
...
-            .glyph_ver_advance(GlyphId(glyph))
-            .map(|units| self.to_em(units))
+    pub fn y_advance(&self, gid: impl Into<GlyphId>) -> Option<Em> {
+        // TODO manually: skrifa doesn't provide these: https://github.com/googlefonts/fontations/pull/1552
+        // self.0
+        //     .rusty
+        //     .glyph_ver_advance(GlyphId(glyph))
+        //     .map(|units| self.to_em(units))
+        None
     }
```
mkorje, line 276: This would need to be done manually, but is going to be tedious... (Though surprisingly this causes no test failures - it is only used to set the `y_advance` of a math variant glyph.)

crates/typst-library/src/text/font/info.rs
```diff
@@ -165,41 +160,11 @@ impl FontInfo {
...
-        if code < 128 { code as char } else { TABLE[(code - 128) as usize] }
-    }
-
-    coded.iter().copied().map(char_from_mac_roman).collect()
+pub(super) fn find_name(skrifa: &skrifa::FontRef, id: StringId) -> Option<String> {
+    skrifa
+        .localized_strings(id)
+        .english_or_first()
+        .map(|name| name.to_string())
 }
```
mkorje, line 168: The methods `skrifa` provides here are a bit nicer and will prefer an English name first (and so the behaviour is slightly different, and better imo?). One issue though is that it doesn't check that the record itself decodes. So this could return an empty string, I believe.

crates/typst-library/src/text/font/mod.rs
```diff
@@ -66,19 +71,24 @@ impl Font {
...
-        //   to the self-lifetime in `ttf()`.
+        //   to the self-lifetime in `ttf()`. TODO: this comment.
         let slice: &'static [u8] =
             unsafe { std::slice::from_raw_parts(data.as_ptr(), data.len()) };
 
-        let ttf = ttf_parser::Face::parse(slice, index).ok()?;
-        let info = FontInfo::from_ttf(&ttf)?;
+        let skrifa = skrifa::FontRef::from_index(slice, index).ok()?;
+        let harfrust = harfrust::ShaperData::new(&skrifa);
+        let info = FontInfo::from_skrifa(&skrifa)?;
```
mkorje, line 80: `skrifa` doesn't validate fonts at all, in contrast to `ttf-parser`, which for example checked that the units per em was valid.

mkorje (commented):
I've finished the migration and gotten this PR in a working state. Still a number of things to do though, and I've left some comments.

Note that `fontations` is currently undergoing an API change (pointed out to me in https://github.com/googlefonts/fontations/issues/2156#issuecomment-5774962878). A number of things are moving from `skrifa` into `read-fonts`, and this work should be finished in a few weeks time. This PR will probably change a fair bit then once that work is done, so probably best to hold off on a review until then. (I will start to get the code working using this new API, which is under a feature flag, on another branch.)

A bunch of tests are failing. Most are svg tests, due to glyph ids changing. A more important change though is that `skrifa` doesn't always return `None` when outlining empty glyphs (e.g. spaces). So we have a bunch of `<symbol><path d="M 0 0"/></symbol>` showing up now. (I'd imagine this change is wrong, though.)
There are also some render tests and one pdf test failing, which I need to look into.

PR #8173, opened 2026-04-26
Allow Citations to Define Their Locator Types in Bibliographies

gozelski:
This PR adds the ability to pass Locator types to `cite` supplements. such that the following is possible:

`cite(<SourceA>, supplement: ("page", [50]))` 

~~This PR adds support for setting locator types to better conform to citation styles. An example of this would be MLA, where the CSL defines delimitation by a comma unless the locator type is of _line_, _page_, or _timestamp_.

An excerpt from [modern-language-association.csl](https://github.com/citation-style-language/styles/blob/77bcd6d4052c606156de90f9f48405250b64a4ea/modern-language-association.csl#L1146-L1159):
```xml
      <choose>
        <if locator="line page timestamp" match="any">
          <group delimiter=" ">
            <text macro="author-short"/>
            <text variable="locator"/>
          </group>
        </if>
        <else>
          <group delimiter=", ">
            <text macro="author-short"/>
            <text macro="label-locator"/>
          </group>
        </else>
      </choose>
```
 This, to the best of my knowledge, is a faithful implementation of MLA. The issue simply lies in Typst's lack of locator typing.
 
 Example citation:
 ```typst
 @SourceA[50]
 ```
 The current output, _before the PR_:
```
(SourceA, 50)
```
 MLA desired output (as it occurs with this PR):
```
(SourceA 50)
```
 

This PR isn't the most elegant way to solve this, as it is a mess of stacked heuristics, but it does _work_.
It is, however, somewhat opinionated to certain syntaxes and may break for styles that both present lines, pages, and timestamps differently enough with unique syntax.

This doesn't change language design or add explicit typing to sources. 
You may want to make this an argument extending cite's supplement field, and make it settable.

Apart from possibly strengthening the "intelligence" of the heuristic, is there a different approach that you would like for this feature, if desired at all?

A feature that gives the same _outcome_ as this PR is a necessity for a majority of my use of Typst, or anyone writing MLA papers in Typst. This could also be implemented by simply setting the type as an argument to `cite`, and simply make a set rule. 

Let me know if you would be more in favor for adding a field for the supplement type either in cite (If on cite itself, or a part of supplement).


gozelski:
I would like to note that `citationberg::taxonomy::Locator::Timestamp` does get applied to timestamp-like locators, but the locator type itself does not seem to apply correctly.

laurmaedje:
I must admit that I'm not super comfortable with the heuristic approach, both because of its inherent heuristicness, but also because the plain_text mechanism is already an unreliable hack in itself.

In principle, a well-defined approach with a proper API seems preferable. What kind of specific API would you imagine? And would that API be convenient enough to use in practice or would it be very annoying because you constantly need to specify things inline?

gozelski:
For an API it seems simple enough to have s field in CiteElem such that you can pass the locator type as a string that maps to the LocatorType enum as so:
```typst
#cite(<MySource>, supplement_type: "page", supplement: [50])

```

or as an optional single or tuple such that the supplement is (content, page), if that is allowed in Typst. 
```typst
#cite(<MySource>, supplement: ([50], "page"))

```
I am not yet sure how the in-line version should look. I will get back to you on that.

laurmaedje:
Related issue: https://github.com/typst/typst/issues/8609

landenhord:
> or as an optional single or tuple such that the supplement is (content, page), if that is allowed in Typst. 
> 
> ```typst
> #cite(<MySource>, supplement: ([50], "page")
> ```

To add my two cents, I think that the tuple type API seems promising. 

To propose an inline version, I was thinking something like

```typst
@Source[locator_type: locator]

@SourceA[page: 50]
@SourceB[chapter: 10]
```

Although I don't know if something like that would be reasonably parseable, especially when things like timestamps get involved.

gozelski:
> To add my two cents, I think that the tuple type API seems promising. 

Yeah, it seems like the best way to add this functionality without introducing a new field that is so closely tied to the value of `supplement`.


> ```typst
> @Source[locator_type: locator]
> 
> @SourceA[page: 50]
> @SourceB[chapter: 10]
> ```

My main concern with this format, and why I haven't yet proposed my own, is the very ambiguity present, if through a timestamp or some other piece of text that includes a colon.

Then again, if it would be within the content block (already losing the promise of []), it could be a good option.

I wonder if it is on the table to present it something like:

```typst
@SourceA(page)[25]
```

or

```typst
@SourceB.page[25]
```
(Edit: I forgot that source labels may possibly have periods and other special characters. If this style is given more thought, that would need to be ironed out)

It would be nice to have an inline syntax in Typst, but having at least the functional syntax would open the door to it being at least settable. 

@laurmaedje I am curious what you think of any of the three options so far, if any of them fit or if they all strive off of the path you envision for Typst's syntax.  I will preemptively add the tuple syntax, let me know if you dislike the tuple approach. Thanks!

gozelski:
I went for using `Repr` and `cast!` explicitly. It may be bit weird to implement `Repr` for `Locator` here as this is not crate portable, this only works because `Repr` is in the same `typst-library` crate. The only other place that seems it could be would be in its own module of `typst-library`. Let me know if that is preferred or if it is preferred to implement this some way else entirely. 
This seems way better than relying on something like `serde` with a `serde_*` library. 

I will try to refactor the `FromValue` implementation for `CitationSupplement` at the very least, as it currently does not mention that the Locator is incorrect.

laurmaedje:
Parsing things from a content block (as in `@source[page: 50]`) would be quite hacky, so I would rather avoid that. More generally, I would rather not add anything locator-specific to the generic `ref` function as it is used for more than just citations.

I think it's best to create a `cite`-only API. This API should be compatible with set rules such that `ref`-generated `cite` elements can pick up defaults defined via set rule.

It's true that the locator kind and supplement are closely related and one does not really make sense without the other. This is a good motivation for keeping them in one field.

I would thus suggest expanding the `supplement` field of `cite` (not `ref`) to also allow passing a dictionary with two keys:

- `kind`: The locator kind, as specified in CSL, as a stringy enum.
- `value`: The locator value. This can be content or an integer. (I would allow the integer form since locators are number variables in CSL terms. Morevero, the integer form could use `LocatorPayload::Str` instead of `LocatorPayload::Transparent`.)

Passing just content or an integer would be equivalent to passing a dictionary with just a `value`.

The dictionary type would _fold,_ such that two properties combine into one, like below:

```typ
#set cite(supplement: (kind: "timestamp"))
#cite(supplement: [10:24])
```

This kind of dictionary with folding is an established pattern, c.f. e.g. `par.first-line-indent`. It's also one of the reasons why I prefer a dictionary over an array even though it is more verbose: It gives a well-defined way to give just the `kind` component. If you're frequently citing with a particular locator kind, but don't want to set it globally, my suggestion would be to define a custom function. That will be shorter in any case because just `supplement: ` is already pretty long, so the array doesn't seem like a big win.

This design could be implemented well on top of the existing diff. If it sounds good to you, I would suggest to change the implementation before I review the code itself so that we can review everything in one go.

If you run into problems with the `Fold` implementation, let me know. It's always a bit of a pain and fairly verbose. I intend to make that easier sometime.

PR #8174, opened 2026-04-26
Optimize `SyntaxNode::into_text` and `SyntaxNode::errors` methods with `traverse` helper

isuffix:
Ok, this version of #8167 should actually be correct. I had accidentally swapped `ControlFlow::Break` and `ControlFlow::Continue` for `NodeKind::Inner` inside `errors` in the previous commit when I submitted it, but then I realized that my scheme for skipping non-erroneous inner nodes was actually not workable at all for `errors`. I've fixed that by just having the function return `std::slice::Iter` instead of `ControlFlow`, which I think is a bit easier to reason about.

---

As far as I can tell, we weren't using `.collect()` for any principled reason, so these methods should be more straightforwardly performant.

I did keep the optimization of `.into_text()` returning the leaf `EcoString` without cloning, but I don't think it really matters and it may make sense to just return `String` instead.

### DIFF
--- a/crates/typst-syntax/src/node.rs
+++ b/crates/typst-syntax/src/node.rs
@@ -103,10 +103,19 @@ impl SyntaxNode {
     pub fn into_text(self) -> EcoString {
         match self.0 {
             NodeKind::Leaf(leaf) => leaf.text,
-            NodeKind::Inner(inner) => {
-                inner.children.iter().cloned().map(Self::into_text).collect()
-            }
             NodeKind::Error(node) => node.text.clone(),
+            NodeKind::Inner(_) => {
+                let mut text = EcoString::with_capacity(self.len());
+                self.traverse(|node| {
+                    match &node.0 {
+                        NodeKind::Leaf(leaf) => text.push_str(&leaf.text),
+                        NodeKind::Inner(_) => {}
+                        NodeKind::Error(node) => text.push_str(&node.text),
+                    }
+                    node.children()
+                });
+                text
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/typst-syntax/src/node.rs
```diff
@@ -103,10 +103,19 @@ impl SyntaxNode {
     pub fn into_text(self) -> EcoString {
         match self.0 {
             NodeKind::Leaf(leaf) => leaf.text,
-            NodeKind::Inner(inner) => {
-                inner.children.iter().cloned().map(Self::into_text).collect()
-            }
             NodeKind::Error(node) => node.text.clone(),
+            NodeKind::Inner(_) => {
+                let mut text = String::with_capacity(self.len());
```
saecki: Why aren't you creating an `EcoString` here? This will force reallocation, because the conversion can't reuse the existing heap allocation.

crates/typst-syntax/src/node.rs
```diff
@@ -230,6 +237,23 @@ impl SyntaxNode {
         Ok(())
     }
 
+    /// Traverse the tree in-order, calling `f` on each node and recursing on
+    /// the returned nodes. Note that `f` can prune the traversal at any point
+    /// by yielding `[].iter()` instead of the actual children slice of an inner
+    /// node.
+    fn traverse(&self, mut f: impl FnMut(&Self) -> std::slice::Iter<'_, Self>) {
```
saecki, line 244: Since it already has to return a concrete iterator type, could it not directly return a slice?

saecki (commented):
Nice, I always like optimizations :)

crates/typst-syntax/src/node.rs
isuffix: To optimize the hot-loop of the recursive call, so we aren't doing as much work by checking the ref-count and whether the string is inline vs. allocated. And there's a nearly guaranteed branch mispredict if the EcoString transitions from inline to allocated.

But it's not like I've measured it, so I can convert to `EcoString` if that's easier.

crates/typst-syntax/src/node.rs
isuffix: I use `std::slice::Iter` to match the return type of `SyntaxNode::children`

crates/typst-syntax/src/node.rs
laurmaedje: The usual pattern in the code base is to use `EcoString` regardless.

isuffix: It is done

laurmaedje:
Thanks!

PR #8178, opened 2026-04-27
Add more structured `css::Properties`

saecki:
This adds a more structured approach of defining CSS properties. The initial work has been done by Laurenz in https://github.com/typst/typst/commit/f4102f7588e9446640e6deb0ca56490ecb603663
Currently they are just converted to inline styles, but in the future the styles might be used to generate a stylesheet. 

Additionally a few other things are changed:
- `DeprecationSink` has been renamed to a more general `WarningSink`
- A `NumericLength` trait has been introduced, mainly for use with `Rel<Length: NumericLength>`
  - This allows implementing `Numeric` for `Ratio`.


crates/typst-library/src/diag.rs
```diff
@@ -369,25 +369,22 @@ impl From<SyntaxError> for SourceDiagnostic {
...
 
-/// Destination for a deprecation message when accessing a deprecated value.
-pub trait DeprecationSink {
-    /// Emits the given deprecation message into this sink alongside a version
-    /// in which the deprecated item is planned to be removed.
-    fn emit(self, message: &str, until: Option<&str>);
+/// Destination for a warning message.
+pub trait WarningSink {
+    /// Emits the message as a warning.
+    fn emit(&mut self, message: HintedString);
```
saecki, line 375: I've removed the `impl Into<HintedString>` to make it dyn compatible. Otherwise `CssWriter` would have to be generic over a `WarningSink` impl.
Alternatively we could create a temporary `Vec` for the `PropertiesBuilder` that the `CssWriter` mutably borrows.

Issue #5906, opened 2025-02-20
Unclosed delimiter error message points to wrong location


Evidlo:
### Description

I recently ran into this issue in my thesis document and spent about 20 minutes trying to find the real problem, which was located far away from the highlighted line.

Below is a stripped down reproduction:

```
- $3x + 123
- more content
- this is some content

$G_1, G_2$
```

```
[evan@blackbox tmp] typst compile test.typ 
error: unclosed delimiter
  ┌─ test.typ:5:9
  │
5 │ $G_1, G_2$
  │          ^

```

I'm not sure if there's any way to fix this with some fancy heuristics?

(Typst 0.12)

### Operating system

Linux

### Typst version

- [ ] I am using the latest version of Typst

PartyWumpus:
It is technically correct though. The first math block *is* from `$3x to the second dollar sign. 

I would guess this is more difficult to do any sort of heuristics for compared to brackets because the opening symbol is the same as the closing one.

PR #8188, opened 2026-04-28
Add syntax warnings

isuffix:
This PR allows us to add warning messages to the syntax tree generated when parsing, and it moves the warnings for `**` and `__` to the parser.

Since warnings do not act like errors in stopping the evaluation of the file, I've chosen to implement them as transparent wrappers around existing syntax nodes, accessible only via the `errors_and_warnings` method on `SyntaxNode`. This method was expanded from the `errors` method, and required adding a new `Erroneous` struct to differentiate between errors and warnings instead of just having a single `bool` for errors. And I renamed `SyntaxError` to `SyntaxDiagnostic` and gave it a boolean to distinguish the two (no enum because it felt weird to have a pure duplicate of `diag::Severity` housed in `typst-syntax`).

Also, the debug output for warnings inspired the new debug output for errors, but is a bit weirder since warnings directly contain a child node. LMK if anyone has thoughts on improving the debug outputs, or if we should just radically alter them at a later time.

The first four commits help prepare for the addition of warnings:
- The first commit adds a test for the `Debug` impl of `SyntaxNode` which we should've had a _long_ time ago
- The second updates the `Debug` impl for error nodes to include hints, but falls back to a shorter version if the error only contains a message (no text or hints)
- The third renames `ErrorNode` variables so that we can call errors `err` and warnings `warn` since calling both `node` would be confusing
- Finally, the fourth commit makes a few larger changes to remove/simplify a number of the private methods on structs in `node.rs` that I thought deserved it. This notably includes removing the `SyntaxError::new` method in favor of just having `SyntaxNode::error` take a message as a string to create errors (since `SyntaxError` is renaming to `SyntaxDiagnostic` and because I just prefer `SyntaxNode::error` taking a string)

Then the fifth commit adds warnings themselves.

Note that I am also planning on separately adding an abstraction for attaching sub-ranges to diagnostics. This would allow warnings and errors to target ranges of text that aren't necessarily contained directly by a Span in the syntax tree, greatly increasing the utility of syntax warnings (since this PR otherwise has little real benefit). However I am still working on that, but I wanted to push this set of changes now since they are ready to review.

### DIFF
--- a/crates/typst-eval/src/lib.rs
+++ b/crates/typst-eval/src/lib.rs
@@ -68,9 +68,16 @@ pub fn eval(
     let root = source.root();
     let mut vm = Vm::new(engine, context.track(), scopes, root.span());
 
-    // Check for well-formedness unless we are in trace mode.
-    let errors = root.errors();
+    // Check for errors or warnings in the syntax tree before evaluating it.
+    // However, if we're inspecting a span, we keep going with evaluation
+    // regardless of syntax errors.
+    let (errors, warnings) = root.errors_and_warnings();
+    for warning in warnings {
+        vm.engine.sink.warn(warning.into());
+    }
     if !errors.is_empty() && vm.inspected.is_none() {
+        // We _could_ also return the warnings here with the errors, but we want
+        // to only use the sink for warnings for consistency.
         return Err(errors.into_iter().map(Into::into).collect());
     }
 
@@ -96,7 +103,7 @@ pub fn eval(
 pub fn eval_string(
     world: Tracked<dyn World + '_>,
     library: &LazyHash<Library>,
-    sink: TrackedMut<Sink>,
+    mut sink: TrackedMut<Sink>,
     introspector: Tracked<dyn Introspector + '_>,
     context: Tracked<Context>,
     string: &str,
@@ -116,9 +123,14 @@ pub fn eval_string(
         SpanMode::Mapped { id, mapper } => root.synthesize_mapped(id, mapper),
     }
 
-    // Check for well-formedness.
-    let errors = root.errors();
+    // Check for errors or warnings in the syntax tree before evaluating it.
+    let (errors, warnings) = root.errors_and_warnings();
+    for warning in warnings {
+        sink.warn(warning.into());
+    }
     if !errors.is_empty() {
+        // We _could_ also return the warnings here with the errors, but we want
+        // to only use the sink for warnings for consistency.
         return Err(errors.into_iter().map(Into::into).collect());
     }
 
--- a/crates/typst-syntax/src/node.rs
+++ b/crates/typst-syntax/src/node.rs
@@ -690,7 +690,17 @@ impl ErrorNode {
 
 impl Debug for ErrorNode {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        write!(f, "Error: {:?} ({})", self.text, self.error.message)
+        if self.text.is_empty() && self.error.hints.is_empty() {
+            write!(f, "Error: {:?}", self.error.message)
+        } else {
+            let mut out = f.debug_struct("Error:");
+            out.field("text", &self.text);
+            out.field("message", &self.error.message);
+            for hint in &self.error.hints {
+                out.field("hint", hint);
+            }
+            out.finish()
+        }
     }
 }
 
@@ -1148,7 +1158,7 @@ Markup: 14 [
             "\
 Markup: 1 [
     Hash: \"#\",
-    Error: \"\" (expected expression),
+    Error: \"expected expression\",
 ]"
         );
         // A syntax error with multiple hints:
@@ -1157,7 +1167,12 @@ Markup: 1 [
             "\
 Markup: 2 [
     Hash: \"#\",
-    Error: \"#\" (the character `#` is not valid in code),
+    Error: {
+        text: \"#\",
+        message: \"the character `#` is not valid in code\",
+        hint: \"the preceding hash is causing this to parse in code mode\",
+        hint: \"try escaping the preceding hash: `\\\\#`\",
+    },
 ]"
         );
     }
--- a/crates/typst-syntax/src/node.rs
+++ b/crates/typst-syntax/src/node.rs
@@ -73,7 +73,7 @@ impl SyntaxNode {
         match &self.0 {
             NodeKind::Leaf(leaf) => leaf.len(),
             NodeKind::Inner(inner) => inner.len,
-            NodeKind::Error(node) => node.len(),
+            NodeKind::Error(err) => err.len(),
         }
     }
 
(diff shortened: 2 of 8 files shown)


crates/typst-eval/src/lib.rs
```diff
@@ -68,10 +68,18 @@ pub fn eval(
...
-    let errors = root.errors();
-    if !errors.is_empty() && vm.inspected.is_none() {
-        return Err(errors.into_iter().map(Into::into).collect());
+    // Check for errors or warnings in the syntax tree before evaluating it.
+    // However, if we're inspecting a span, we keep going with evaluation
+    // regardless of syntax errors (although we will still add warnings).
+    let errors_and_warnings = root.errors_and_warnings().into_iter();
+    if root.erroneous().errors && vm.inspected.is_none() {
+        // Returning errors and warnings together should preserve ordering.
+        return Err(errors_and_warnings.map(Into::into).collect());
```
laurmaedje: currently, we never emit warnings as part of the `Err` payload. With this, they'll show up in the error instead of the `warnings` field of `Warned` in the final output, which is somewhat unexpected.

isuffix:
As discussed synchronously, I have made a few changes:

- Moved the `Debug` impls back to their original locations for cohesion across the codebase
- In `typst-eval`, moved warnings to always be added to the `Sink` instead of directly returning them in the `SourceResult`. This required splitting the return of `errors_and_warnings` into two vectors for convenience. This is maybe inefficient, but it was easy and I doubt it will really matter
- Renamed `Erroneous` to `Diagnosis` (cute!), however I am somewhat concerned that we never actually call `.diagnosis()` outside of `node.rs`, but I guess it's fine
- Moved warnings out of the top `NodeKind` enum and created a separate `NodeWrapper` enum.

I also added a commit that renames the `NodeKind` enum to `Node` to help avoid confusion with `SyntaxKind`, and because it obviously represents a kind of node because it's an enum. This also works well with the new `node()` and `node_mut()` methods for descending into `NodeWrapper`.

isuffix:
Oops! I wasn't careful about cleaning up from previewing `docs-in-typst` and jj accidentally pushed the docs website with +109,492 lines :flushed: (now fixed)

crates/typst-syntax/src/node.rs
```diff
@@ -91,27 +163,29 @@ impl SyntaxNode {
...
+            Node::Inner(_) => &EMPTY,
+            Node::Error(err) => &err.text,
         }
     }
 
     /// Extract the text from the node.
     ///
     /// Builds the string if this is an inner node.
     pub fn into_text(self) -> EcoString {
+        // This isn't fully efficient for warnings, but the efficient
```
saecki, line 222: Not fully related to this PR, but is there really a benefit that `into_text` takes the node by value? Cloning of `EcoString` should be cheap, either just a bitwise copy for the inline representation up to 15 bytes, which should be a good amount of leaf nodes, and an increment of the atomic refcount for other ones. Since the returned string never seems to be modified, increasing the refcount seems like a non issue.
Most use sites of `into_text` seem to clone the `SyntaxNode` anyway, which clones the `LeafNode::text`, and also needlessly increments and decrements the refcount of inner and error nodes.

Maybe the functions should have more descriptive names like `leaf_text` and `full_text` to clear up the intent?

crates/typst-syntax/src/node.rs
```diff
@@ -283,27 +372,28 @@ impl SyntaxNode {
 
     /// Whether this is a leaf node.
     pub(super) fn is_leaf(&self) -> bool {
-        matches!(self.0, NodeKind::Leaf(_))
+        matches!(self.node(), Node::Leaf(_))
+        // TODO: Should we also treat non-empty errors as leaves?
```
saecki, line 444: I think a problem could be that the error nodes always return `SyntaxKind::Error`, which could negatively impact completion?

crates/typst-syntax/src/node.rs
```diff
@@ -657,68 +754,165 @@ impl Debug for InnerNode {
...
+            && self.child.spanless_eq(&other.child)
+    }
+}
+
+impl Debug for WarningWrapper {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        /// This helper lets us output `hint: "msg"` instead of `"hint: msg"`
+        /// while using `debug_set`.
+        /// FUTURE: In Rust 1.93, we can use `fmt::from_fn` instead!
+        struct FieldHelper<'a>(&'static str, &'a EcoString);
```
saecki: You can also use `typst_utils::debug`. With Rust 1.93 all uses of it can be replaced.

saecki (commented):
Not sure, if there was any more feedback from Laurenz, but apart from the few remarks this looks good to me!

Could this in combination with the sub ranges be used to better report unclosed delimiters? For example adding a hint to the opening delimiter, or the location where it was popped off the stack (https://github.com/typst/typst/issues/5906).

laurmaedje:
> Not sure, if there was any more feedback from Laurenz, but apart from the few remarks this looks good to me!

Aside from the `typst_utils::debug` thing, which I also noticed, I didn't have any issues with the code. However, a benchmark run revealed a ~4% slowdown on an eval-heavy document, so that's a bit of a blocker. Ian and I discussed this yesterday and I gave him access to the benchmark infra to play around with what we could do.

crates/typst-syntax/src/node.rs
isuffix: I agree. I'll open another PR that renames them after this one.

I will mention that Laurenz once commented that we should really have a `summarize` method that would add ellipses in a smart location instead of generating the full text, perhpas based on an expected or maximum length. I would be in favor of it, but I'm not sure how exactly I would go about implementing it. Thoughts?

laurmaedje:
Oh, I just saw that this was mislabelled. Are we on the same page that we'll want to change the code to avoid the perf regression in some way (e.g. something like your `top-level-spans` branch)?

isuffix:
Yes exactly. I've been a bit busy this week, but have been planning to bring the commits from  that branch into this PR, just been doing some cleanup on it.

isuffix:
I've added the four commits from my `top-level-spans` branch which should resolve the performance regression. The first is a small change I wanted to do, the second and third were actually just part of #8254 that needed to come earlier to work well, and the fourth is the change that moves the `Span` and `SyntaxKind` to the top level of `SyntaxNode`.

The reason `SyntaxKind` has to be inside the enum variants (not a field of `SyntaxNode` like the `Span`) is so that Rust will optimize the representation to store the kind and the enum's tag next to each other using the padding bytes. If we instead store the kind outside the enum, Rust doesn't do this and we add 8 bytes to the representation. This does lead to an annoying situation where warnings wrap the `Node` type, but also store their own top-level `SyntaxKind`, which means they actually duplicate the kind. But as I explain in the comment, trying to deduplicate the syntax kinds would require a whole other enum type, which makes mutable access too painful in my opinion.

This should now be ready to review.

laurmaedje:
Thank you!

crates/typst-syntax/src/node.rs
saecki: I'm not sure how I'd approach this, but it would definitely be nice to have some syntax aware summary function, that'll also handle cases like: https://github.com/typst/typst/issues/8339

PR #8201, opened 2026-05-01
Add repeatable figure captions

### LINKED ISSUE #5057: Add ability to repeat the caption (or a modified caption) of a multi-page table in subsequent pages

hongjr03:
Closes typst/typst#5057.

This adds support for repeating figure captions when a breakable figure spans multiple pages. It is opt-in through `figure.caption(repeat: true)`, so existing figures keep their current layout behavior. The caption also gets a synthesized `continued` field while it is being laid out on continuation pages, which lets users customize continuation captions with the usual show rule.

This is what `figure.caption` now looks like:

```typ
figure.caption(
  repeat: false,
  continued: false,
  body,
)
```

An example:

```typ
#show figure: set block(breakable: true)

#show figure.caption.where(continued: true): it => [
  Continued table #context it.counter.display(it.numbering)
]

#figure(
  table(
    columns: 2,
    table.header([*A*], [*B*]),
    ..range(20).map(n => ([#n], [#n])).flatten(),
  ),
  caption: figure.caption(
    repeat: true,
    position: top,
    [Repeated caption],
  ),
)
```

table-figure-caption-repeat

Also, this PR revisits typst/typst#5191, which already explored repeatable figure captions. I saw that the comments there pointed out that repetition is most useful when continuation captions can be customized. The `continued` field is intended to cover that case while staying within the existing show rule model.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2866,7 +2866,7 @@ dependencies = [
 
 [[package]]
 name = "test-wrapper"
-version = "0.14.2"
+version = "0.15.0"
 
 [[package]]
 name = "thin-vec"
@@ -3057,7 +3057,7 @@ checksum = "8e28f89b80c87b8fb0cf04ab448d5dd0dd0ade2f8891bae878de66a75a28600e"
 
 [[package]]
 name = "typst"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "arrayvec",
  "comemo",
@@ -3076,15 +3076,15 @@ dependencies = [
 
 [[package]]
 name = "typst-assets"
-version = "0.14.2"
-source = "git+https://github.com/typst/typst-assets?rev=955ae8d#955ae8db08d6e36694eb9cc61e61f586032ea572"
+version = "0.15.0"
+source = "git+https://github.com/typst/typst-assets?rev=c0ae970#c0ae970f64413cf8e39cecd56d15ee50de42eaf4"
 dependencies = [
  "bitflags 2.11.1",
 ]
 
 [[package]]
 name = "typst-bundle"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "comemo",
  "ecow",
@@ -3106,7 +3106,7 @@ dependencies = [
 
 [[package]]
 name = "typst-cli"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "chrono",
  "clap",
@@ -3152,12 +3152,12 @@ dependencies = [
 
 [[package]]
 name = "typst-dev-assets"
-version = "0.14.2"
-source = "git+https://github.com/typst/typst-dev-assets?rev=0967e5e#0967e5ebc4ef264eee89dee938b29033c91d8e2c"
+version = "0.15.0"
+source = "git+https://github.com/typst/typst-dev-assets?rev=a2e29ed#a2e29ed9393121735ce56d9d06deeab701cf35e0"
 
 [[package]]
 name = "typst-docs"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "az",
  "base64",
@@ -3195,7 +3195,7 @@ dependencies = [
 
 [[package]]
 name = "typst-eval"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "comemo",
  "ecow",
@@ -3213,7 +3213,7 @@ dependencies = [
 
 [[package]]
 name = "typst-fuzz"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "comemo",
  "libfuzzer-sys",
@@ -3229,7 +3229,7 @@ dependencies = [
 
 [[package]]
 name = "typst-html"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "az",
  "bumpalo",
@@ -3250,7 +3250,7 @@ dependencies = [
 
 [[package]]
 name = "typst-ide"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "comemo",
  "ecow",
@@ -3270,7 +3270,7 @@ dependencies = [
 
 [[package]]
 name = "typst-kit"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "chrono",
  "codespan-reporting",
@@ -3305,7 +3305,7 @@ dependencies = [
 
 [[package]]
 name = "typst-layout"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "az",
  "bumpalo",
@@ -3339,7 +3339,7 @@ dependencies = [
 
 [[package]]
 name = "typst-library"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "arrayvec",
  "az",
@@ -3405,7 +3405,7 @@ dependencies = [
 
 [[package]]
 name = "typst-macros"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "heck",
  "proc-macro2",
@@ -3415,7 +3415,7 @@ dependencies = [
 
 [[package]]
 name = "typst-pdf"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "az",
  "bytemuck",
@@ -3442,7 +3442,7 @@ dependencies = [
 
 [[package]]
 name = "typst-realize"
-version = "0.14.2"
+version = "0.15.0"
 dependencies = [
  "arrayvec",
  "bumpalo",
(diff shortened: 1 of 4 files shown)


hongjr03:
The last workaround in #5057 creates several caption-like regions around a table: the normal figure caption, a continuation label at the top, and optionally a continuation label at the bottom. So I wonder whether a future design should keep `caption` as exactly one `figure.caption` with a `position`, or whether it should eventually allow multiple `figure.caption` elements, each with its own placement and repetition condition.

For example:

```typ
caption: (
  figure.caption(position: bottom)[Main caption],
  figure.caption(position: top, repeat: rest)[Continuation of table @tbl],
  figure.caption(position: bottom, repeat: except-last)[Continued on next page],
)
```

Not necessarily this exact API, but #5057 seems to suggest that repeated figures may need more than a single positioned caption.

SchrodingerBlume:
Nice work, the continued synthesized field is a clever fit with Typst's show rule model. That said, I had a few thoughts on the API and performance side:

1. On the expressiveness of `continued: bool`

One concern about the API: `continued: bool` can only distinguish "first page" from "any continuation page". It can't express things users may reasonably want on continuation captions, such as "is this the last page", "which continuation index this is", or "N of M pages". Since `figure.caption` is public API, locking `continued` in as a bool now would make it a breaking change to evolve into something richer later (e.g. `none | (index, total, is-last) or similar`). Would it be worth considering a more structured type up front, even if only `true`/`false`-equivalent values are populated for now? That would keep the door open for finer-grained show rules without an API break.

SchrodingerBlume:
2. On show-rule granularity for multiple repeat figures

Since `figure.caption.where(continued: true)` matches globally, there doesn't seem to be a clean way to customize continuation captions for one specific repeat figure without also affecting all others in the document. Users would have to fall back on filtering by `kind`, `supplement`, etc., which gets awkward when several long tables coexist and each wants its own continuation style. I don't think this needs to block the PR, but it might be worth calling out as a known limitation, or thinking about whether the API should eventually allow per-figure continuation customization (e.g. passing the continuation body/function directly on the figure).

SchrodingerBlume:
3. On layout performance

The layout path looks like it lays out each continuation caption at least twice: once in the measurement loop to collect `caption_heights`, and again in the final assembly loop. For captions containing context expressions, counters, or other non-trivial content, this could become noticeably expensive on long tables that span many pages. A few options worth considering: assume a uniform height from the first measurement and only re-measure if the body actually overflows; or introduce a measure-only fast path that skips full frame construction. Probably fine to land as-is, but might be worth a comment in the code flagging the cost so it's on the radar for future optimization.

hongjr03:
> Nice work, the continued synthesized field is a clever fit with Typst's show rule model. That said, I had a few thoughts on the API and performance side:
> 
> 1. On the expressiveness of `continued: bool`
> 
> One concern about the API: `continued: bool` can only distinguish "first page" from "any continuation page". It can't express things users may reasonably want on continuation captions, such as "is this the last page", "which continuation index this is", or "N of M pages". Since `figure.caption` is public API, locking `continued` in as a bool now would make it a breaking change to evolve into something richer later (e.g. `none | (index, total, is-last) or similar`). Would it be worth considering a more structured type up front, even if only `true`/`false`-equivalent values are populated for now? That would keep the door open for finer-grained show rules without an API break.

I think this is probably not a problem if the goal is just to let the caption behave differently on continuation pages. We already have the counter there, so users can distinguish the repeated caption from the original one without needing more metadata. As the example in the description:

```typ
#show figure.caption.where(continued: true): it => [
  Continued table #context it.counter.display(it.numbering)
]
```

So if `continued` is only meant to tell us whether this is a continuation caption, I think a bool is enough for now.

> 2. On show-rule granularity for multiple repeat figures
> 
> Since `figure.caption.where(continued: true)` matches globally, there doesn't seem to be a clean way to customize continuation captions for one specific repeat figure without also affecting all others in the document. Users would have to fall back on filtering by `kind`, `supplement`, etc., which gets awkward when several long tables coexist and each wants its own continuation style. I don't think this needs to block the PR, but it might be worth calling out as a known limitation, or thinking about whether the API should eventually allow per-figure continuation customization (e.g. passing the continuation body/function directly on the figure).

The tension is that we want selector-based customization for repeated captions, but we may also want more per-figure customization. If we want both, this may eventually call for a function-shaped API instead of only a selector-based one, but I have not fully worked that out yet.

> 3. On layout performance
> 
> The layout path looks like it lays out each continuation caption at least twice: once in the measurement loop to collect `caption_heights`, and again in the final assembly loop. For captions containing context expressions, counters, or other non-trivial content, this could become noticeably expensive on long tables that span many pages. A few options worth considering: assume a uniform height from the first measurement and only re-measure if the body actually overflows; or introduce a measure-only fast path that skips full frame construction. Probably fine to land as-is, but might be worth a comment in the code flagging the cost so it's on the radar for future optimization.

The current implementation lays out continuation captions once to determine how much vertical space to reserve for the figure body, and then lays them out again when assembling the final frames. Because the body layout needs the caption heights up front, while the caption itself can be affected by those styling rule or contexual elements. I'm still thinking about whether there is a safe small optimization here.

Anyway, thanks for the detailed review — really appreciate it!

laurmaedje:
Hi! Thanks for the PR.

In principle, relying on `grid` (like in #5191) seems simpler than having a dedicated multi-layouter for figures. Writing a layouter is non-trivial and for maintainability reasons I would like to keep their number as low as possible. Using existing lower-level layout primitives also means such a feature would be easier to reproduce in user space when building something analogous to figures. Synthesized fields are also always something that I see as a last resort; they are definitely not my favorite feature.

The unsolved problem in #5191 was how to handle continuations, but it feels solvable without giving up on the `grid` approach. I'd like to see more exploration of the design space for how to handle continuations in the alternative approach before shipping a figure-specific custom layouter. Maybe we can find a unified approach that works for figure captions but also for other repetitions of content (like in a table header)? 

Just an idea off the top of my head: Maybe a generic element that takes either a variadic number of content arguments and picks the N-th in the N-th time it is placed in the document or it takes a function that gets which occurance it is as an integer. (I thought approximately 3 minutes about this, so there might be some flaw in this, but seems worth exploring).

laurmaedje:
I suppose such a generic element would also need a key (e.g. the figure's location) to work out. Only elements with the same key would be counted together. This already does not feel super nice, but I still think that some nice generic design exists. What also feels _kinda_ related is repeated content in slideshows and polylux's `only` mechanism.

hongjr03:
Thanks, that direction makes sense. The grid-based version could then be something like this in principle:

```typ
grid.header(repeatable(key: <tbl>, n => {
  if n == 0 { [Main caption] } else { [Continuation of table @tbl] }
}))
```

That seems closer to the `grid` approach from #5191. The `key` would avoid unrelated repeated elements sharing the same counter.

The remaining case for #5057 is probably continuations that depend on whether the current fragment is the final one. But it also feels somewhat separable from the basic “N-th placement” primitive, so maybe the first step is just to see whether the keyed occurrence model works well for first/rest continuation captions with grid.

Also, #5057 may need both axes: where the repeated content is placed, and how it varies by occurrence. For example, a repeated table could reasonably want the main caption at the bottom, a continuation label at the top of continuation fragments, and a “continued on next page” label at the bottom of non-final fragments.

If that direction works out well, figure captions could potentially become just another consumer of the same mechanism rather than requiring a dedicated layouter. I’ll try to prototype this and see what a workable API could look like.

PR #8207, opened 2026-05-03
Implement column balancing

### LINKED ISSUE #466: Balancing the contents of multiple columns to the same vertical space
I remember that in LaTeX the output of multicolumns will be balanced to the same vertical space in all columns to have enough space below that could start a new single column layout or an image or so.

In Typst the output for

```typst
#set par(justify: true)

#set page(columns: 2)

#lorem(500)
```

is currently unbalanced which looks not so professional, I think.

balanced

Maybe a parameter like `balance-columns` or `columns: (balance: bool)` can be added to make this possible.

eltos:
Closes #466

API implemented as suggested in https://github.com/typst/typst/issues/466#issuecomment-1685034064
```typ
// Whether to equalize the height of columns by breaking columns early.
#set columns(balanced: true)
```

Example (from tests added):

-----------------

The implementation then follows advice from https://github.com/typst/typst/pull/5115#issuecomment-2449817143
> I would try to implement this roughly as follows:
> - Layout a region normally first
> - Then, if we're in balanced mode, estimate the needed column sizes based on the items in the distributor (basically somehow extracting the `used` sizes in `Distributor::finalize` before expansion is applied, but (probably) after fr spacing
> - Store this information somewhere in the `Composer` (like the insertions are)
> - Relayout (via `Stop::Relayout(PlacementScope::Parent)`)

With these remarks:
- I grouped the `columns` related parameters to `fn layout_fragment_impl` as it appears the comemo macro supports only up to 12 arguments, see https://github.com/typst/comemo/blob/5944487f5e9d11949038662df4b307ec56f153b3/src/input.rs#L270.
  **EDIT**: As the follow-up PR #8353 adds yet another parameter, that PR contains a refactor grouping those into a dedicated struct.
- I track used sizes in `Distributor::run` instead of `Distributor::finalize`, since I needed them there anyway to decide if content still fits into the region (see below)
- The most simple approach of shrinking the region size before laying out contents in `Composer::column_contents` for the relayout run turned out to not work well with footnotes, and not give the freedom required to optimize balancing results. Therefore, the distributor keeps track of used sizes along the way in various subroutines of `Distributor::run` and checks not only if content still fits into the region, but also if it would exceed the height demanded by column balancing (via `Composer::fits`). For breakable blocks I instead temporarily shrink the region size before passing it down to `multi.layout`. This is possible here, as there is no issue with footnotes on this level, and then it's just easier and cleaner.
- I chose to include bottom floats but not footnotes in the balancing logic, since I think this is the most favourable handling (e.g. for auto placed figures in a text). So footnotes stay at the bottom, but bottom floats get lifted to the balanced height. This can be seen in the `columns-balanced-figures` test.

Added some tests with all possible combination with figures, floats, footnotes, blocks, tables, and pagebreaks I could think of.

Limitations / TODOs:
- Balancing uses a simple 1/n rule to distribute heights, and columns are broken when the height limit is reached. There is no glue-streching to actually make the bottom of the columns align perfectly with each other. (out of scope for PR)
- Using `#set page(columns: n)` mid-page will balance but still start a new page. (out of scope for PR)
- Balancing does not work across nested column environments (only the innermost columns get balanced)
- [x] ~~If distribution cannot work perfectly, e.g. when there are 3 columns but 32 lines, the extra 2 lines will usually end up in the last column (10+10+12), even though it might be nicer to have the last column one line shorter (11+11+10).~~ 
  **EDIT: With 70cf35c192f2ebf7ff7a2b9da8c18edd98868a6a the result will now be 11+11+10 lines.**
- [x] ~~Edge cases with 2 or 3 lines to be distributed over 6 columns give unsatisfactory results, see [this comment below](https://github.com/typst/typst/pull/8207#issuecomment-4873692995)~~
  **EDIT: Fixed in 70cf35c192f2ebf7ff7a2b9da8c18edd98868a6a**
- [x] ~~Currently, relayout is called only once, even though changing the column breaking may change the amount of size to distribute. You can see this when setting `page(height: auto)` for the `columns-balanced-tables` test, then when the table is broken and the header is repeated, this additional required space was not known in the initial iteration before balancing, and the result becomes uneven.~~
  **EDIT: With 891e6df84473cbf3b59538d3d1cde5613c069778 balancing is repeated if the total height increased due to a previous balanding run, thus the table example is nicely balanced**
- The improved balancing logic is not yet used for breakable blocks, see https://github.com/typst/typst/pull/8207#discussion_r3560233284  
  **EDIT:** This will be addressed in a follow-up PR, see #8652


dayala1:
It feels like an option to have several iterations will be a natural step towards smoothing some edges. Maybe a second iteration redistributing for the non-last columns the length of the overflowing content for the last one? Maybe just in cases where the last column overflows above a given threshold. Or just iteratively adding length if the last column overflows.

Sorry if some of this doesn't make much sense, I don't know much about the inner implementation but it is a feature I would really like to see added to typst.

eltos:
> It feels like an option to have several iterations will be a natural step towards smoothing some edges. Maybe a second iteration redistributing for the non-last columns the length of the overflowing content for the last one? Maybe just in cases where the last column overflows above a given threshold. Or just iteratively adding length if the last column overflows.
> 
> Sorry if some of this doesn't make much sense, I don't know much about the inner implementation but it is a feature I would really like to see added to typst.

This is basically covered by what I meant with "🚧 Draft PR: I want to play a bit more and see if the simple 1/n balancing algorithm can be improved...", but thank you for your early feedback @dayala1.

Re-iterating is only required if the total height changed due to balancing (repeated table header example). Otherwise, the distribution can be optimized from the start just by tweaking the break condition a bit (see 6bb1e899), yielding the behaviour depicted in the screenshot below. I have implemented both improvements and updated the PR description accordingly. 

canvas

eltos:
I kindly suggest reviewing this together with the closely related https://github.com/typst/typst/pull/8353

jqwv:
Thank you for implementing this! I guess, the following code produces an unexpected end:

```typst
#set page(columns: 6)
#set text(lang: "la")
#set par(justify: true)
#set columns(balanced: true)
#lorem(2000)
```

I would expect the last full page to have the same number of lines in all columns and then break the last part on the next page and then put it in the first column instead on the last one.

balanced

E: Maybe also the last full page has an unoptimized number of lines per column value.


E2: The following does not work but is requested in the issue and is one main use case:

```typst
#set page(columns: 6)
#set text(lang: "la")
#set par(justify: true)

#set columns(balanced: true)

#lorem(800)
#set page(columns: 1)
#lorem(100)
```

Maybe it is a typst limitation of `columns` in general and is not related to this PR.
Previously, doing a pagebreak was required here because of the missing balancing.
I guess with `balanced: true` the linebreak should not be done after a page-columns change.
The same problem appear before a page-columns change. At best it would be possible to change the layout without forcing a pagebreak. Otherwise we cannot reuse the white space for different content.

eltos:
For your first example it might be possible to improve the balancing logic for a better result, at least putting the trailing lines in the first and not last column(s). I will look into it when I find the time, maybe (note so self) "place at least 1 child in a column, even if that exceeds the 1/n target" and "accept content even if exceeds target height when otherwise empty space would be more that X% of item height" with "100%/n" or "100%/(n-1)" instead of current "50%".

But I am not sure how to get a satisfactory result when trying to balance 2 or 3 lines into 6 columns while also avoiding widows. Maybe you better not balance the columns in this case.

For the second example, it is indeed not possible to change page level columns mid-page. This was not requested in the original issue and I would consider this a separate issue and out of scope for this PR. What was requested by [someone commenting in the issue](https://github.com/typst/typst/issues/466#issuecomment-2481023464) was inserting balanced multi-column blocks mid page, and this should work with the current PR, but I could add a test for it.
```typ
#lorem(14)

#columns(2, gutter: 4pt, balanced:true)[
  #lorem(20)
]

#lorem(20)

#columns(3, gutter: 4pt, balanced:true)[
  #lorem(20)
]

#lorem(20)
```

jqwv:
I think the 6 columns test should definitly work. In worst case, we would have to try different possibilities for example K-1,K,K+1 as line numbers and select the one which gives the expected result.

I wrote in the linked issue: "to have enough space below that could start a new single column layout or an image or so."
But I understand that it is not related to this issue. Maybe I will open a distinct issue, once this PR is merged.

Otherwise, I think it works good and fast!

eltos:
> I wrote in the linked issue: "to have enough space below that could start a new single column layout or an image or so."

Fair enough, using the space below should be possible with `#place(bottom, float: true, scope: "parent", ...)`. I used this in one of my templates to emulate a "widetext" environment. But yeah for doing it via `#set page` specifically I'd suggest to open a separate issue.

jqwv:
The reason why the content on the last page is in the last column instead in the first, can be simplified using this code:

```typst
#set page(columns: 3)

#set columns(balanced: true)

#rect(width: 100%, height: 1cm)
```

If we set `balanced: false`, the content is placed correctly.

The following code creates a case that is unclear: the contents get balanced correctly, but the first and third column is used. Maybe it is more expected that the first and second column is used. Maybe we could check what other tools like latex with multicols or InDesign do here, or even just HTML/CSS if it supports balancing.

```typst
#set page(columns: 3)

#set columns(balanced: true)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 3cm)
```

It seems to be a bit random whether it chooses the second or third col:

```typst
#set page(columns: 3)

#set columns(balanced: true)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 2cm) // switch height between 2 and 4 cm and the col jumps
```

Regarding the algorithm: it seems to balance quite well. Probably good enough. But I saw the following case:

```typst
#set page(columns: 3)

#set columns(balanced: true)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 3cm)

#rect(width: 100%, height: 2cm)

#rect(width: 100%, height: 4cm)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 5cm)

#rect(width: 100%, height: 3cm)

#rect(width: 100%, height: 2cm)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 2cm)

#rect(width: 100%, height: 3cm)

#rect(width: 100%, height: 2cm)

#rect(width: 100%, height: 4cm)

#rect(width: 100%, height: 1cm)

#rect(width: 100%, height: 3cm)
```

Here, we can see that col3.last is bigger than col3.first and we could place the col3.first as col2.last and get a better balancing (better means here = the maximum height of all balanced columns is smaller than before). I think this check can be added to each col neighbours. Not sure if there is some big disadvantage.
Maybe, implementing this, directly solves the issue where it has one text line missing from my first comment.

The following example has the same problem:

```typst
#set page(columns: 3)
#set columns(balanced: true)

#rect(width: 100%, height: 2cm)

#lorem(20)

#rect(width: 100%, height: 4cm)

#lorem(15)

#rect(width: 100%, height: 3cm)

#lorem(25)
```

The following is an AI example, maybe it is indeed unexpected (it does not use the third col at all):
```typst
#set page(columns: 3)
#set columns(balanced: true)

- #lorem(6)
- #lorem(20)
- #lorem(3)
- #lorem(15)
- #lorem(12)
- #lorem(18)
- #lorem(7)
- #lorem(22)
```

Other AI examples did all pass.
But it seems this problem appears with all kind of contents like also raw code: the first col gets one third and the second two third of the content. The third col is empty.

eltos:
I experimented with some slight variations of the balancing algorithm to have leftover lines in the left columns instead of the right columns
- Variant 🅰️ results in columns mostly being filled from left to right (analogous to the reading order), and the leftover space cumulates in the last column. It is predictable and as far as I know it matches the behaviour of CSS and LaTeX. But if there are many columns, the last column might have a lot less lines than the other ones, or might even be empty. This might be desirable or not. Also, this variant can result in the max column height to be larger than necessary, if a large element happens to be at the bottom of a column (see one of the rect examples below).
- Variant 🅱️ results in overall less difference between column heights, with leftover space distributed between columns. It yields an overall more "balanced" result in most cases, but how exactly the content is distributed also depends on the heights of individual elements and thus is less predictable. It works well for up to including 5 columns, but for the 6 column example with the widow it is still not perfect. While it tries to make the column heights only reduce from left to right, that is not guaranteed and intermediate columns may actually be shorter, as is the case for the widow example.

Both variants are better than what I had before for 3 or more column texts, and both resolve the issue with the widow you posted above (see comparison below). But I am unsure which variant is typographically more desirable.

@jqwv @dayala1 What do you think is more favourable?

<details>
<summary>Comparison</summary>
Left is previous state, middle is variant 🅰️, right is variant 🅱️ 

The examples with rects of different size are a bit pointless, as it is impossible to balance them, and it is a matter of taste what is the least ugly result. I have included them in the comparison below, but I would focus on the common problem of text balancing, as in such corner-cases you will have to reside to `#colbreak()` to get whatever look you want.






</details>

eltos:
> The following is an AI example, maybe it is indeed unexpected (it does not use the third col at all):
> [...]
> Other AI examples did all pass.
> But it seems this problem appears with all kind of contents like also raw code: the first col gets one third and the second two third of the content. The third col is empty.

Thanks for checking @jqwv, the list not getting balanced was a bug. I applied balancing only in `multi` but forgot to handle `multi_spill` too. Fixed and test added (not yet pushed).

jqwv:
Thank you for the updates and the `multi_spill` fix! I think the middle variant in the comparison image is the expected result. As I understand, the last column can only be empty, when there is very few content per column like in your comparison example. If this is true, then this is expected and not a problem. I will recompile when you add the new code with the fix and test again. But it looks very good already.

dayala1:
Having both those options seems really good to me. A for big documents where you want nothing weird to happen, B for cases where it is important to avoid blank spaces. If I had to choose strictly one I would say A, but having modularity when it comes to important parts of the algorithm seems fun.

eltos:
> I asked AI to create such a balancing algo in python for a simple rectangle case and svg output with binary search requirement.

I can also ask AI myself, but the implementation I use is not binary search or iterative, but rather a single run (maybe a second if balancing changed the total height), and unless Laurenz is of different opinion I intend to keep it that way. One cannot know the heights of all elements a-priori, as this depends on the layouting, which in turn depends on where columns get broken. A binary search would therefore have to include the complete layouting logic in the loop, which will probably slow down compilation significantly. Also, it is much simpler to use a breaking criterion than iterating the layout.

eltos:
I agree that 🅰️ is preferable, thank you both for your feedback.

> As I understand, the last column can only be empty, when there is very few content per column like in your comparison example. If this is true, then this is expected and not a problem. 

This is true. The examples are minimized to avoid bloating the reference image file size.

> I will recompile when you add the new code with the fix and test again. But it looks very good already.

I have updated the PR accordingly. Feel free to test it again.

jqwv:
> One cannot know the heights of all elements a-priori

Ok, I thought the column-balancing happens in a re-layout. I assumed that all heights are known from a first pass.

eltos:
You are correct that it happens in a single (or maybe two) re-layout runs, but currently only the total height required is known from the previous run. I wasn't very precise with my wording. I wanted to say that it's more involved than a list of fixed-size content as in your AI example. There are things like breakable blocks, fractional and relative sizes, weak spaces which can collapse, paragraph leading vanishing at a column break, floating figures and footnotes the position of which may change, tables headers which might or might not get repeated, widow and orphan prevention rules, and more. That's why I said the layout composition has to be in the loop.

jqwv:
I compiled the new updates. All my `lorem` tests look expected now.

---

The following example was already in your comparison and does not balance. I think it is ok to not balance here, just as a note:

```typst
#set page(columns: 2)
#set columns(balanced: true)

#rect(width: 100%, height: 2cm)
#rect(width: 100%, height: 3cm)
```
---

The following looks unexpected:

```typst
#set page(columns: 4)
#set columns(balanced: true)

- lists start right
```
Though, this seems to effect only the first two elements. When there are at least 3, it seems to work.

---

All other things I tested look good and expected, thank you!

crates/typst-layout/src/flow/compose.rs
```diff
@@ -664,27 +715,38 @@ impl<'a, 'b> Insertions<'a, 'b> {
...
     }
 
     /// The combined height of the top and bottom area (including clearances).
     /// Subtracting this from the total region size yields the available space
     /// for distribution.
     fn height(&self) -> Abs {
+        self.top_size + self.bottom_size + self.footnote_size
+    }
+
+    /// The combined height of the top and bottom area for floats (including clearances) but excluding footnotes.
```
laurmaedje: Please wrap this at column 80.

crates/typst-layout/src/flow/compose.rs
```diff
@@ -635,10 +676,20 @@ struct Insertions<'a, 'b> {
...
     width: Abs,
     skips: Vec<Location>,
 }
 
+/// State for column balancing
+#[derive(Default)]
+pub struct ColumnBalancing {
+    // The height used by the inner contents (e.g. text) during column layouting.
+    pub(crate) used_height: Abs,
+    // The balanced height of the columns
```
laurmaedje: This could be a triple slash comment. Could you also terminate all comments with full stops?

crates/typst-layout/src/flow/compose.rs
```diff
@@ -166,6 +170,15 @@ impl<'a, 'b> Composer<'a, 'b, '_, '_> {
             inner.next();
         }
 
+        // Column balancing with re-layout
+        if self.config.columns.balanced && self.work.children.is_empty() {
```
laurmaedje: Does this intentionally use `self.work.children.is_empty()` rather than `self.work.done()`?

crates/typst-layout/src/flow/compose.rs
```diff
@@ -166,6 +170,15 @@ impl<'a, 'b> Composer<'a, 'b, '_, '_> {
...
+            let column_height = balancing_height / self.config.columns.count as f64;
+            if self.column_balancing.column_height.is_none_or(|h| h < column_height) {
```
laurmaedje: Do you think this could potentially retrigger more than once or even somehow keep looping in an extreme edge case?

crates/typst-layout/src/flow/compose.rs
```diff
@@ -244,6 +265,26 @@ impl<'a, 'b> Composer<'a, 'b, '_, '_> {
...
+    /// The height limit if column balancing is active
+    pub fn column_balancing_limit(&self) -> Option<Abs> {
+        if self.column < self.config.columns.count - 1 {
+            self.column_balancing.column_height
+        } else {
+            None
+        }
+    }
+
+    /// If the amount fits into the region, taking into account column balancing limits
```
laurmaedje:
```suggestion
    /// Whether the amount fits into the region, taking into account column balancing limits
```

crates/typst-layout/src/flow/compose.rs
```diff
@@ -244,6 +265,26 @@ impl<'a, 'b> Composer<'a, 'b, '_, '_> {
...
+    pub fn fits(&self, regions: Regions, amount: Abs) -> bool {
+        let mut fits = regions.size.y.fits(amount);
+        if let Some(target) = self.column_balancing_limit() {
+            // Add elements as long as the balancing target is not reached. By not including
+            // the amount here, we avoid protruding items to cumulate in the last column.
+            fits &= target.fits(self.column_balancing.used_height);
+        }
+        fits
```
laurmaedje: In principle, this could be lazy if `fits` is already false. E.g. something like this:
```rs
regions.size.y.fits(amount)
    && self
        .column_balancing_limit()
        .is_none_or(|target| target.fits(self.column_balancing.used_height))
```

crates/typst-layout/src/flow/compose.rs
```diff
@@ -664,27 +715,38 @@ impl<'a, 'b> Insertions<'a, 'b> {
...
 
     /// Produce a frame for the full region based on the `inner` frame produced
     /// by distribution or column layout.
-    fn finalize(self, work: &mut Work, config: &Config, inner: Frame) -> Frame {
+    fn finalize(
+        self,
+        work: &mut Work,
+        config: &Config,
+        inner: Frame,
+        position_bottom_floats: Option<Abs>,
```
laurmaedje: This variable name is not 100% clear imo, because it's not clear that it's the _bottom_ end of the bottom floats. It could also be the Y (top) position where the bottom floats start. Maybe just column_height would actually be clearer?

laurmaedje: I also think there's actually a bug hidden here because things can overflow with footnotes:
```typ
#set page(columns: 2)
#set columns(balanced: true)

#lorem(5)
#footnote(lorem(300))
#lorem(300)
```

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -427,6 +449,7 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
 
         // Push an item for the frame.
         self.regions.size.y -= frame.height();
+        self.composer.column_balancing.used_height += frame.height();
```
laurmaedje: This might be problematic because a sticky frame that's moved to another column via a snapshot/restore could maybe be counted twice. I'm not entirely sure, but please double-check. 

My general suggestion would be as follows:
- Moving the `used` summation to a new field of the `Distributor` instead of directly modifying the `Composer`
- Then, we can also snapshot and restore it there in the `DistributorSnapshot`
- And to avoid some code duplication, we can encapsulate this pattern in a new method:
  ```rs
  self.regions.size -= x;
  self.used_height += x;
  ```

crates/typst-layout/src/flow/distribute.rs
laurmaedje: In principle a check like this should probably also trigger if the column is limited. I feel like we may want to have more principled methods for accessing regions with different amounts of space factored in and used for different purposes. This would also be a way to avoid the code duplication below (for constructing the `pod`).

It might also be a chance to remove `self.composer.fits` in favor of such a region access + a normal `.fits` call, which would feel cleaner to me.

If you have any thoughts, let me know and we can discuss it a bit. Maybe, we could even always "synthesize" the region from the `used_height` and the immutable base region. Then, we would also automatically get rid of the duplication I mentioned in the third bullet point of https://github.com/typst/typst/pull/8207#discussion_r3560209314.

crates/typst-layout/src/flow/mod.rs
```diff
@@ -238,23 +232,29 @@ pub fn layout_flow<'a>(
 fn configuration<'x>(
     shared: StyleChain<'x>,
     regions: Regions,
-    columns: NonZeroUsize,
-    column_gutter: Rel<Abs>,
+    columns: (NonZeroUsize, bool, Rel<Abs>),
```
laurmaedje: I think the struct from the other PR would already be nice to have here

crates/typst-library/src/layout/columns.rs
```diff
@@ -5,14 +5,11 @@ use crate::layout::{Length, Ratio, Rel};
...
-/// remaining height on the page. Support for balanced columns is planned for
-/// the future.
-///
-/// When arranging content across multiple columns, use @colbreak to explicitly
-/// continue in the next column.
+/// The `column` function lets you separate the contents of any container into
+/// multiple columns. By default, columns take up the height of their container or the
+/// remaining height on the page and are filled up one after another.
+/// Use @colbreak to explicitly end a column and continue in the next one.
+/// Use @columns.balanced to automatically equalize the height of the columns.
```
laurmaedje: Since we're already on the columns page, I would write something like
```
Use the @columns.balanced[`balanced` parameter] to ...
```

tests/suite/layout/columns.typ
```diff
@@ -129,3 +129,143 @@ Hallo
 A
 #colbreak(weak: true)
 B
+
+--- columns-balanced paged ---
+#set page(width: 300pt)
+#set par(justify: true)
+#rect(inset: 0pt, outset: 0pt, stroke: yellow,
+    columns(3, balanced: true, lorem(40))
```
laurmaedje: Please use two spaces of indent here.

tests/suite/layout/columns.typ
```diff
@@ -129,3 +129,143 @@ Hallo
...
+
+--- columns-balanced-lists paged ---
+// Check that breakable blocks are balanced correctly via `multi` and `multi_spill`
+#set page(width: 350pt, columns: 4)
+#set columns(balanced: true)
+#{
+    8 * [- #lorem(5)]
+}
+
+
```
laurmaedje:
```suggestion
```

crates/typst-library/src/layout/columns.rs
```diff
@@ -65,9 +62,36 @@ pub struct ColumnsElem {
     pub count: NonZeroUsize,
 
     /// The size of the gutter space between each column.
+    ///
+    /// #example(
+    ///   ```
+    ///   >>> #set page(background: place(center + horizon, dy: -1cm, rect(
+    ///   >>>   width: 1cm, height: 1cm, stroke: (x: green + 2pt), inset: 0pt,
+    ///   >>>   line(length: 1cm, stroke: green + 2pt)
+    ///   >>> )))
```
laurmaedje: While I appreciate the intent, I think it's a bit confusing that the example output does not match the source code.

crates/typst-library/src/layout/columns.rs
```diff
@@ -65,9 +62,36 @@ pub struct ColumnsElem {
...
     #[default(Ratio::new(0.04).into())]
     pub gutter: Rel<Length>,
 
+    /// Whether to equalize the height of columns by breaking columns early.
+    ///
+    /// #example(
+    ///   ```
+    ///   #set page(columns: 2, height: 5cm)
+    ///   #set par(justify: true)
+    ///   >>> #set columns(gutter: 1cm)
```
laurmaedje: This had me confused for a second because it makes it seem like the gutter setting in the `gutter` example has no effect.

laurmaedje (commented):
Thanks for the PR! It's really good work; focused and clean. I have a few notes and found an edge case bug, but this seems very solid already and the results speak for themselves. :)

crates/typst-layout/src/flow/distribute.rs
eltos: > I feel like we may want to have more principled methods for accessing regions with different amounts of space factored in and used for different purposes.

I agree. There is also the last comment by jqwv mentioning an issue with balancing a single list item. The cause being the `multi` and `multi_spill` code paths not using the improved `composer.fits` function, but a simple shrunken region size with the balanced height as a hard limit. Moving the `composer.fits` logic into a generalized region concept with both, a height target and height limit, could solve this too.

laurmaedje: Nice, then feel free to give it a shot.

crates/typst-layout/src/flow/compose.rs
git-staus: >Could you also terminate all comments with full stops?

Sorry for chiming in with something unrelated to the review, but I just wanted to mention that `cargo clippy -- -W clippy::doc-paragraphs-missing-punctuation` catches just this. Sadly there is no lint to catch `//` instead of `///` though.

adaltare:
Hey @eltos, I just want to thank you personally for spending your time on this long standing issue! I have been patiently waiting for balanced columns for nearly two years now. It is a must have for the historical restoration work I do. I was dreading that this feature would never be added to Typst, or worse that I would have to learn Rust to solve this problem myself. You guys are geniuses! I see that you are also adding column separators, which is going to make my life so much easier too.

**So seriously, you rock, and I can not thank you enough! Keep up the fantastic work!**

crates/typst-layout/src/flow/compose.rs
eltos: Done

crates/typst-layout/src/flow/compose.rs
eltos: > This could be a triple slash comment. Could you also terminate all comments with full stops?

Done

crates/typst-layout/src/flow/compose.rs
eltos: No, changed it.

crates/typst-layout/src/flow/compose.rs
eltos: I can't think of a way. Height increase due to early column wrapping can't happen more often than the number of columns for sure. And since the initial run will underestimate the required height, it will initially distribute to the same or more than finally required columns, already triggering the maximum additional amount to be added. For repeated table headers I have checked this, weak spacing is anyway trimmed at a column break, and if someone does something fancy with context, that would be on a different re-layout level and not change heights down here.

crates/typst-layout/src/flow/compose.rs
eltos: Applied.

crates/typst-layout/src/flow/compose.rs
eltos: Of course, I applied the suggestion.

crates/typst-layout/src/flow/compose.rs
eltos: Fixed and example added:
```typ
--- columns-balanced-footnote-overflow paged ---
#set page(width: 300pt, height: 150pt, columns: 2)
#set columns(balanced: true)
#lorem(10)
#footnote(lorem(30))
#lorem(45)
```

crates/typst-layout/src/flow/distribute.rs
eltos: I double-checked, `column_balancing.used_height` is always modified when `region.size.y` is modified, so if it was counted twice, the same would apply to the region size (which is not handled by snapshot/restore either). However, restore is only called from within `Distributor.finalize` and after restore, the only way `used_height` (and `region.size.y`) are modified is through `Distributor.trim_spacing`. This is correct behaviour and since `Distributor.trim_spacing` removes the `Item::Abs` from the `items` vector, it is never counted twice (snapshot/restore only restores the current index within `items`, not the `items` vector itself).

I cannot just _move_ the summation to the `Distributor`, as part of the sum (float_height) is added in `compose.rs`. I could duplicate the `used` field there, and then copy the value over at some point. But IMHO this would be more confusing. Also, `Distributor` already modifies `compose.work`, so having it modify `composer.column_balancing.used_height` does not change the picture.

Given that there is no need, and I think it would make things more involved than cleaner, I only applied the 3rd point you suggested. Let me know if that is okay.

Note that the region generalization point mentioned below is still open, the new `use_height` function will probably be touched again.

crates/typst-layout/src/flow/mod.rs
eltos: I pulled it in.

crates/typst-library/src/layout/columns.rs
eltos: Changed.

tests/suite/layout/columns.typ
eltos: Done.

crates/typst-library/src/layout/columns.rs
eltos: Okay, removed it.

crates/typst-library/src/layout/columns.rs
eltos: Fair enough, reduced the gutter. I don't want to remove it, because the default is so small it looks like regular spaces in a single column example.

crates/typst-layout/src/flow/distribute.rs
eltos: > Moving the `composer.fits` logic into a generalized region concept with both, a height target and height limit

This one's still on the **🚧 TODO** list. The current idea is
- Add a `balanced` height field to `Regions`
- Move the `Composer.fits` function to `Regions`, and refactor usages of `Regions.is_full` and `Regions.size.y.fits`
- Maybe even replace the (remaining) `size` field with a `used_height` and fixed `width`, moving the `Distributor.use_height` logic into `Regions` and refactoring all 102 usages of `Regions.size`

eltos:
> So seriously, you rock, and I can not thank you enough! Keep up the fantastic work!

@adaltare Thank you for the warm and motivating words

crates/typst-layout/src/flow/distribute.rs
laurmaedje: My intuition was that this should be possible without changing the definition of the `Regions` themselves, by creating a new copy with reduced `size`. Is that not possible?

If doing this means coupling balancing to the generic Regions definition, then I'm not sure I'm in favor.

eltos: > By creating a new copy with reduced size. Is that not possible?

I'm confused. Isn't this [the current approach with "pod"](https://github.com/typst/typst/pull/8207/changes#diff-77365ab089296da5f259515979412a7bb66db669bc5ace6249df8ed15b88dd44R342)?


```typ
// For column balancing, reduce the region size for layout
let mut pod = self.regions;
if let Some(lim) = self.composer.column_balancing_limit() {
    let remaining = lim - self.composer.column_balancing.used_height;
    pod.size.y.set_min(remaining);
}
```

The limitation being that it's a hard limit which does not distinguish between the "maximum height allowed" and the "optimal height targeted". By not allowing contents to exceed the "target height" slightly, which would overall yield a better balancing result, that approach favors contents which don't quite fit, to accumulate in the last column. The example being the "single list item" being balanced into the last rather than first column, as [mentioned](https://github.com/typst/typst/pull/8207#issuecomment-4920720141).

> It might also be a chance to remove self.composer.fits in favor of such a region access + a normal .fits call, which would feel cleaner to me.
> [...]
> I feel like we may want to have more principled methods for accessing regions with different amounts of space factored in and used for different purposes. This would also be a way to avoid the code duplication below (for constructing the pod).

I interpreted your earlier comment such that -- instead of creating a new copy "pod" -- you suggested factoring in a "target height"  and moving the "composer.fits" logic into a generalized "region.fits" method. Since the `multi` layout path is complex with many branches and possible recursion, having that incorporated into `Regions` seems cleaner to me than passing along a `balanced_height` variable to all the subroutines, though I guess it would not be impossible.

Since it seems I misunderstood, could you please elaborate @laurmaedje?

laurmaedje: > I'm confused. Isn't this [the current approach with "pod"](https://github.com/typst/typst/pull/8207/changes#diff-77365ab089296da5f259515979412a7bb66db669bc5ace6249df8ed15b88dd44R342)?

My intent with this was not to improve the balancing, but merely to make the logic harder to misuse and more maintainable (less state to keep in sync). I was not aware that a separate balanced height allows making the algorithm produce better results. In fact, I misunderstood the purpose of the logic of `Composer::fits`. But I understand it now.

Overall, I just thought it might be nice to pull the pod creation out into a method and use it consistently, such that the distributor does not modify its `self.regions` at all (it stays immutable) and instead materializes a new one from `self.regions` plus the used height whenever `self.regions` would currently be used. Then, `use_height` would only add to one `Length` variable. Though I'm not sure how much overhead this would create (would need to benchmark) and how nice it would actually be ...

crates/typst-layout/src/flow/distribute.rs
laurmaedje: > I double-checked, column_balancing.used_height is always modified when region.size.y is modified, so if it was counted twice, the same would apply to the region size (which is not handled by snapshot/restore either).

Unlike the used height, the region size isn't used anymore during and after finalization. If a heading is put at the bottom of the column, its height is considered via `use_height` and then we restore because it was an orphan, doesn't the heading then factor into `column_balancing.used_height`?

My intent with the suggestions (similar to the other comment) was mostly to keep all state as local as possible to make it easier to reason about things. The used float height is purely additive to the used height in the column contents, right? I was thinking that no state-based mutation is necessary as conceptually we're just summing values and maybe it's possible to cleanly express this with return types instead of mutation. It's hard for me to judge though whether this would feel clean without actually trying it out. I also don't want to send you on a wild goose chase.

eltos: I have implemented the refactor in e30ce29d0784097f086420f37e588a4bc5e135f0 and I agree that it is more maintainable that way. 

I also applied another refactor that was made possible by the local tracking of used height: 0517cc00f0cf1f4f8f6a388c48c340bf71f887ff

This should be resolved now.

crates/typst-layout/src/flow/distribute.rs
eltos: Thank you for the clarification, and sorry for the misunderstanding.

With 5dc30faaecfedabda7cb8da7668e98c4de4cd365 the balancing height reduction is done before the `is_full` check.

I have also implemented the "pod creation" refactor in 07782dbf67cd593efa75ea2523a27ac62f221d16, but pushed to a separate branch [`eltos-columns-balanced-optional-refactor`](https://github.com/eltos/typst/tree/eltos-columns-balanced-optional-refactor) as I don't think it's better. Feel free to have a look, benchmark and cherry-pick or discard as you prefer.

eltos:
Thank you for the productive review @laurmaedje 
With my last commits, all review comments should be addressed. 

However, the point of improving of balancing in the `multi` code path is still open (the example of a single list item ending up in the last rather than first column). Since you mentioned doubts about my proposal, I would first like to discuss it before going down a path you don't like ;) 
If you want, we can also carry this over to a separate PR.

Context:
```typ
/// Processes a breakable block.
    fn multi(&mut self, multi: &'b MultiChild<'a>) -> FlowResult<()> {
        let mut pod = self.regions;

        // For column balancing, reduce the region size for layout
        if let Some(lim) = self.target {
            let remaining = lim - self.used.y;
            pod.size.y.set_min(remaining); // <-- TODO: Setting the target as a hard limit does not always give the best balancing result
        }

        // Skip directly if the region is already (over)full. `line` and
        // `single` implicitly do this through their `fits` checks.
        if pod.is_full() {
            return Err(Stop::Finish(false));
        }

        // Lay out the block.
        let (frame, spill) = multi.layout(self.composer.engine, pod)?;
```

I have been thinking about it again. Since in the `multi` function, a breakable block can basically contain anything, even a columns environment recursively, I cannot think of a feasible way to know a better "hard limit" for balancing a priori. This would basically require knowing the size of all the building blocks inside the breakable block. 

So to improve the balancing beyond the current "hard limit = balancing target" approach, I don't think there is a way other than to pass down both heights, the hard limit (as before) and in addition the (smaller) balancing target.
This could be done as part of the `Regions` struct, which I assume could turn out to be a clean solution. Maybe the concept of a "target" height could even be useful beyond column balancing one day.
Alternatively it could also be done by adding a `target_height` parameter to all the involved functions.
In either way, it will probably be a larger refactor to ensure the target height is handled properly for all manipulations and checks where `regions.size` is involved. Let me know if any of this seems reasonable enough for you so I can give it a go, or if you prefer a different approach. 

**EDIT:** Actually, if you don't mind, I would prefer a follow up PR, also to unlock the other PR with the column separator.

laurmaedje:
> Actually, if you don't mind, I would prefer a follow up PR, also to unlock the other PR with the column separator.

I'm not sure about the approach yet (though I see the benefit; it's just really wide-ranging in its effects), but in any case I agree that a separate PR would be a better place. This PR looks like it's in a pretty good spot now. The refactors look nice from a first glance. I still need to double-check the changes, which I won't get to anymore today, but will try to do soon.

oliverpool:
With this PR my usecase of balanced table would also be perfectly addressed, thank you:
image

(I just wanted to provide some user-feedback before the merge, I hope it isn't adding too much noise)

crates/typst-layout/src/flow/compose.rs
```diff
@@ -69,6 +70,7 @@ pub struct Composer<'a, 'b, 'x, 'y> {
     page_base: Size,
     page_insertions: Insertions<'a, 'b>,
     column_insertions: Insertions<'a, 'b>,
+    pub(crate) column_balancing_height: Option<Abs>,
```
laurmaedje: This can now be made private :)

crates/typst-layout/src/flow/compose.rs
```diff
@@ -728,6 +773,9 @@ impl<'a, 'b> Insertions<'a, 'b> {
         // surprised and considered this strange. In LaTeX, it can be changed
         // with `\usepackage[bottom]{footmisc}`. We could also consider adding
         // configuration in the future.
+
+        let mut offset_bottom =
```
laurmaedje: Might be nice to have these variables be called differently, e.g. `float_offset_bottom` and `footnote_offset_bottom`.

The definition of `float_offset_bottom` could even use the initial value of `footnote_offset_bottom`, though that would disrupt the nice structure where the variable are defined right above where they will be used. So not sure about that one, I'll let you make the call.

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -192,8 +210,8 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
                     if weakness <= prev_weakness
                         && (weakness < prev_weakness || amount > prev_amount)
                     {
-                        self.regions.size.y -= amount - prev_amount;
                         *item = Item::Abs(amount, weakness);
+                        self.use_height(amount - prev_amount);
```
laurmaedje, line 214: I would move this line up by above the assignment, back to where the previous one was.

crates/typst-layout/src/flow/distribute.rs
```diff
@@ -270,19 +288,29 @@ impl<'a, 'b> Distributor<'a, 'b, '_, '_, '_> {
         Abs::zero()
     }
 
+    /// Whether the amount fits into the remaining region, taking into account column balancing limits.
```
laurmaedje: Could you wrap the latest changes to column 80. There is also a new comment without a trailing period.

crates/typst-layout/src/flow/mod.rs
```diff
@@ -352,6 +359,13 @@ impl<'a, 'b> Work<'a, 'b> {
     }
 }
 
+#[derive(Hash)]
```
laurmaedje, line 363: Could you add brief doc comments here?

crates/typst-library/src/layout/columns.rs
```diff
@@ -5,14 +5,12 @@ use crate::layout::{Length, Ratio, Rel};
...
-/// The `column` function lets you separate the interior of any container into
-/// multiple columns. It will currently not balance the height of the columns.
-/// Instead, the columns will take up the height of their container or the
-/// remaining height on the page. Support for balanced columns is planned for
-/// the future.
-///
-/// When arranging content across multiple columns, use @colbreak to explicitly
-/// continue in the next column.
+/// The `column` function lets you separate the contents of any container into
+/// multiple columns. By default, columns take up the height of their container or the
```
laurmaedje: This also overflows column 80.

crates/typst-library/src/layout/columns.rs
```diff
@@ -65,9 +63,32 @@ pub struct ColumnsElem {
...
     /// The size of the gutter space between each column.
+    ///
+    /// #example(
+    ///   ```
+    ///   #set page(columns: 2, height: 5cm)
+    ///   #set par(justify: true)
+    ///   #set columns(gutter: 15mm)
+    ///   #lorem(30)
+    ///   ```
+    /// )
```
laurmaedje:
```suggestion
    /// ```example
    /// #set page(columns: 2, height: 5cm)
    /// #set par(justify: true)
    /// #set columns(gutter: 15mm)
    /// #lorem(30)
    /// ```
```
same below

tests/ref/svg/hashes.txt
```diff
@@ -314,7 +328,7 @@ cae9e4e87ce63e0a3be8cde298cf5427 gradient-math-conic
 de4eb48f3300499bec530dd5b48c9a4a gradient-math-misc
 b0ce942acd695b5909c8204775be8ddc gradient-math-radial
 ec0a500c006cdc8066a2e791722b2b8c gradient-math-root
-344c967eb31d9b931ab6744b242edb4d gradient-math-underover
+cbf1869a9e8236f466068a8061eabb43 gradient-math-underover
```
laurmaedje, line 331: I'm a bit confused by how 0517cc00f0 can have any effect on any test. Isn't it supposed to be a 100% equivalence refactor. Does this have something to do with the order of float operations?

laurmaedje (commented):
This looks very good now. :) Just some very minor nits and a question about the changed SVG hash. Once addressed, this should be good to merge.

crates/typst-layout/src/flow/distribute.rs
laurmaedje: I would have created an 
```rs
fn regions(&self) -> Regions {
    self.regions.shrunken_by(self.used.y)
}
```
instead of having this copied everywhere, but given that this would not eliminate the `if let Some(lim) = self.target {`, stuff in `multi` and `multi_spill`, I'm also inclined to consider this not worthwhile at the moment.

crates/typst-layout/src/flow/compose.rs
eltos: I decided to keep the order

crates/typst-layout/src/flow/distribute.rs
eltos: The compiler didn't like that:
> cannot borrow `*self` as mutable more than once at a time [E0499]

It knows that the `self.regions` modification is disjoint with the `self.items` borrow earlier up, but with `self.use_height(...)` it doesn't know for sure and rejects.

laurmaedje: ah okay

tests/ref/svg/hashes.txt
eltos: I was also confused, but yes the result is slightly different due to floating point precision:
```
used:      Size(13.333pt, 10.438pt)
self.used: Size(13.333pt, 10.438000000000002pt)
````
How it affects the ID hashes in that particular SVG I don't know.

laurmaedje:
Thank you, awesome work and I agree that the review rounds were very productive. :)

PR #8209, opened 2026-05-04
Enable splitting more of typst-library into separate crates

davidv1992:
For the dutch kiesraad's use of typst, there is a desire to reduce the number of dependencies used by typst. This PR is meant as an illustration for how this reduction can be achieved by splitting some of the functionality in typst-library into their own crates. The first commit provides the ground work for making these splits possible, and the second commit shows how this can be applied by splitting of the loading of advanced data formats into its own crate.

Library users of typst get the benefit of reduced dependencies by depending on the individual crates. This requires some small duplications of code from the main typst crate on their end, but as this crate is small that duplication will be limited and is deemed acceptable, at least by the kiesraad.

On the other hand, the typst project does not have to deal with feature flags and the combinatorial explosion of trying to ensure everything works with some subset of flags enabled. The dependency structure of the crates will inherently enforce that the split-off parts of the code base will function independently by making use of the split-off functionality in typst-library and typst-render impossible by default.

As a side benefit, work in the split-off crates will likely have shorter compile-test cycle times, as fewer crates will have to be recompiled for changes in those crates as compared to changes to typst-library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3067,6 +3067,7 @@ dependencies = [
  "typst-html",
  "typst-layout",
  "typst-library",
+ "typst-loading",
  "typst-macros",
  "typst-realize",
  "typst-syntax",
@@ -3345,10 +3346,8 @@ dependencies = [
  "az",
  "bitflags 2.11.1",
  "bumpalo",
- "ciborium",
  "codex",
  "comemo",
- "csv",
  "ecow",
  "either",
  "flate2",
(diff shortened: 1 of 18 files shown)


laurmaedje:
Just marked as a draft since it's more illustrative that meant for merging as-is. I'm quite busy right now, but will give feedback later.

davidv1992:
Updated this to be based on commit 3e32cb92 to make the comparison with #8459 easier.

laurmaedje:
See https://github.com/typst/typst/pull/8459#issuecomment-4957810596.

PR #8211, opened 2026-05-04
Allow escaping characters in panic strings

### LINKED ISSUE #5219: Escaping characters in `panic` not working
### Description

It seems like I cannot escape quotations marks (`"`) in `panic` calls.

Let's say I want to panic with the message `Please provide an identifier like "math"`.
I expected this would be done like this:
```typ
#panic("Please provide an identifier like \"math\"")
```
However, when you compile the above code, you get:
```
error: panicked with: "Please provide an identifier like \"math\""
  ┌─ test.typ:1:1
  │
1 │ #panic("Please provide an identifier like \"math\"")
  │  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
In the first line it would be great to have `Please provide an identifier like "math"`.

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

GabrielDTB:
Closes https://github.com/typst/typst/issues/5219

You rarely, if ever, want to present the literal contents of a string in a panic call (especially as a library author). A panic like:

```typ
#panic("msg\n\thint: suggestion")
```

would normally be presented as:

```
error: panicked with: "msg\n\thint: suggestion"
```

With this PR, it is instead presented as:

```
error: panicked with: msg
        hint: suggestion
```


isuffix (commented):
The code itself is fine, but we may want to consider escaping the output strings if we do want this change.

In particular, this seems to allow emitting control characters like ansi escape sequences from panic strings directly, which can affect the visual output of errors in the terminal, and is liable to allow code to break some terminal emulators.

Running `echo '#panic("hello \u{1B}[31mworld")' | cargo run -- compile - /tmp/test.png` does change "world" to red:



so this is definitely a possible concern.

I'm not sure whether that's really part of our threat model (and I would reckon this is probably already possible in some form) but this change would make it far easier to do so. Neither of the `codespan-reporting` nor `termcolor` crates we use for error output seem to support escaping control sequences themselves (given a quick look), so perhaps this could be an upstream concern?

laurmaedje:
> I'm not sure whether that's really part of our threat model (and I would recon this is probably already possible in some form)

I'm not sure either. With `assert(message: "..")` it's already possible I think. I looked into this once, noticed that rustc suffers from the same problem and moved on for the time being...

laurmaedje:
I think this change makes sense and while the concern about terminals is real, this doesn't really change anything about it, since `assert` already suffers from it.

Fwiw, I also checked how `console.log` behaves in the browser and it's essentially the same. It can print any kind of value, but strings are printed without quotes.

Thanks!

PR #8212, opened 2026-05-05
Fix error message for attempting to assign to temporary return value

### LINKED ISSUE #5034: Confusing error when using certain methods as an assignee
### Description

Some methods don't produce mutable references, hence their outputs can't be assigned to. The compiler wrongly reports this as the methods not existing.

Example:
```typst
#{
 let s = "string"
 s.first() = "S"
}
```

### Reproduction URL

_No response_

### Operating system

Linux

### Typst version

GabrielDTB:
Closes https://github.com/typst/typst/issues/5034

Attempting something like `s.first() = "e"`, where s is a string, currently gives an error about the method not existing. The issue is actually that the return value isn't mutable. This PR changes the error to give the actual cause.


isuffix (approved):
Thank you, this change looks quite nice!

I will caution that I have been working on a more general fix for mutable methods which would change their evaluation order and refactor this code (although I've been focusing on other things recently), so you should let me know if you're planning on working more on this code so we don't duplicate work :)

laurmaedje:
Thanks!

PR #8224, opened 2026-05-07
Page Setup Guide fixes

### LINKED ISSUE #8176: Incorrect mention of "footnote" in Page Setup Guide
https://typst.app/docs/guides/page-setup/

> Alternating alignment between even and odd pages is not currently supported using this property. To do this, you'll need to specify a custom footer with your footnote [...]

I don't know why this mentions footnotes.

### LINKED ISSUE #8202: Suboptimal example document structure in Page Setup Guide
### Description

The second example in the [Add columns](https://typst.app/docs/guides/page-setup/#columns) section should use the `title` element and has a level-2 heading without a level-1 heading.

Corrected code:

```typ
#set page(columns: 2)
#set par(justify: true)

#place(
  top + center,
  float: true,
  scope: "parent",
  title[
    Impacts of Odobenidae
  ],
)

= About seals in the wild
#lorem(80)
```

MDLC01:
Closes https://github.com/typst/typst/issues/8202. I added a hidden line to reduce the size of the level-1 heading. I am not fully sure it is best to keep it hidden.

Also closes https://github.com/typst/typst/issues/8176.

### DIFF
--- a/docs/content/guides/page-setup.typ
+++ b/docs/content/guides/page-setup.typ
@@ -188,7 +188,7 @@ This is one of many numbered pages.
 
 Go to the @numbering.numbering[`{numbering}` function reference] to learn more about the arguments you can pass here.
 
-In case you need to right- or left-align the page number, use the @page.number-align[`number-align`] argument of the @page[`{page}`] set rule. Alternating alignment between even and odd pages is not currently supported using this property. To do this, you'll need to specify a custom footer with your footnote and query the page counter as described in the section on conditionally omitting headers and footers.
+In case you need to right- or left-align the page number, use the @page.number-align[`number-align`] argument of the @page[`{page}`] set rule. Alternating alignment between even and odd pages is not currently supported using this property. To do this, you'll need to specify a custom footer and query the page counter as described in the section on conditionally omitting headers and footers.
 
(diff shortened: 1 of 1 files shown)


docs/content/guides/page-setup.typ
```diff
@@ -284,19 +284,18 @@ Very commonly, scientific papers have a single-column title and abstract, while
...
   #set par(justify: true)
 
   #place(
     top + center,
     float: true,
     scope: "parent",
-    text(1.4em, weight: "bold")[
-      Impacts of Odobenidae
-    ],
+    title[Impacts of Odobenidae],
```
laurmaedje: Typically, `set document(title: [Impacts of Odobenidae])` + `title()` would be more idiomatic, but it's more ceremony and may distract from the purpose of the example. What do you think, which one is preferable?

MDLC01: I've changed the example to use that instead. I agree that it may distract a bit from the example, but I don't want to give users the impression that using `title[..]` directly is the most idiomatic solution. Especially because I feel like new users often have troubles distinguishing what is idiomatic code and what has been simplified into less idiomatic code in examples.

laurmaedje:
Thanks!

PR #8225, opened 2026-05-07
Don't call `par`s `block`s in the "Advanced Styling" tutorial

### LINKED ISSUE #8146: Docs: The Advanced Styling tutorial still says `par`s are `block`s
> The title, the grid, paragraphs, and all other elements that Typst arranges from the top to the bottom of the page are called _blocks._  Each block is controlled by the [`block`] function.

Should probably rather say

> The title, the grid, and all other elements that Typst arranges from the top to the bottom of the page, expect for paragraphs, are called _blocks._  Each block is controlled by the [`block`] function.

Waiting on #8138 to PR (unless the completely new tutorials will already be a thing for 0.15.0)

cady-b:
Resolves #8146


docs/content/tutorial/3-advanced.typ
```diff
@@ -140,7 +140,7 @@ This looks right. Let's also add the author list: Since we are writing this pape
...
 
-Looking at the authors and the title, they are a bit too close together. You can address this by using another show-set rule to configure the space below the title. The title, the grid, paragraphs, and all other elements that Typst arranges from the top to the bottom of the page are called _blocks._ Each block is controlled by the @block function. It controls behaviors like their distance and whether a block can contain a page break. That means that we can write another show-set rule that selects the title to set the block spacing:
+Looking at the authors and the title, they are a bit too close together. You can address this by using another show-set rule to configure the space below the title. The title, the grid, and all other elements that Typst arranges from the top to the bottom of the page, except for paragraphs, are called _blocks._ Each block is controlled by the @block function. It controls behaviors like their distance and whether a block can contain a page break. That means that we can write another show-set rule that selects the title to set the block spacing:
```
laurmaedje: I feel like parentheses may read a bit better than commas here because the insertion is still part of the last item of the preceding comma-separated list.

cady-b: Fair! I'll try to keep that in mind since it's different from my personal preference :sweat_smile:

laurmaedje:
Thanks!

Issue #8190, opened 2026-04-29
More configurable cross-page links in bundle export


YDX-2147483647:
### Description

The new bundle feature makes it easy to create cross-page links.
In the following example, `#outline()` generates `foo/index.html#loc-1`. Is it possible to change it to `foo/#loc-1`?

```typst
#document("index.html")[
  #outline()
]

#document("foo/index.html")[
  = Foo
]
```

```sh
typst watch main.typ dist/ --features html,bundle --format bundle
```

```html
    <nav role="doc-toc">
      <h2>Contents</h2>
      <ol style="list-style-type: none">
        <li><a href="foo/index.html#loc-1">Foo</a></li>
      </ol>
    </nav>
```

### Use Case

Well, shorter URLs look nicer.

[Hypertext Style: Cool URIs don't change](https://www.w3.org/Provider/Style/URI) by W3C also recommends to remove the `.html` suffix.

> **What to leave out**
>
> - **File name extension.** This is a very common one `.cgi`, even `.html` is something which will change. You may not be using HTML for that page in 20 years time, but you might want today's links to it to still be valid. The canonical way of making links to the W3C site doesn't use the extension. ([how?](https://www.w3.org/Provider/Style/URI#remove))
> - …
> 
> So a better example from our site is simply
> ```
> http://www.w3.org/1998/12/01/chairs
> ```
> a report of the minutes of a meeting of W3C chair people.

### My attempt

I've tried `query(heading).first().location()`, but it seems that there's no way to retrieve the URL from a `location`.

And I see that links generated in #8138 do not have `/index.html`. If I understand correctly, then cross-page links in #8138 are constructed directly from URL strings embedded in `metadata`, which are stored by the rust program in advance. That's a different story…

laurmaedje:
It's planned to make this more configurable, but it was deliberately left out of scope for now. There are various desirable configurations (absolute/relative paths, with/without full path) and it's not yet clear how best to expose them.

The current choice is the least opinionated: It's always relative (so it's independent of whether it is hosted on a subpath) and always full paths (so it's independent of the web server configuration).

There is also some documentation for this in the section "Links in bundle export" on the `link` function.

As for #8138: It does use native links, but it uses a patched show rule for links (see `PATCHED_LINK_RULE` in `docs/src/world.rs`) because it wants to make an opinionated choice.

eira-fransham:
[...]

PR #8227, opened 2026-05-08
Add trailing / to requests in bundle index.html

istudyatuni:
When `index.html` is opened in browser, and url doesn't end with `/`, relative links doesn't work correctly. Example (link resolution is the same in JS):

```js
console.log(new URL("article", "https://example.org/foo/").href)
// https://example.org/foo/article

console.log(new URL("article", "https://example.org/foo").href)
// https://example.org/article
```

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references#current_directory_relative)

Repro:

```typ
#document("index.html")[
    #link("foo", "foo")
]
#document("foo/index.html")[
    this link is invalid:
    #link("bar.html", "bar")

    when current url is `/foo` (not `/foo/`), link points to `/bar.html`
]
#document("foo/bar.html")[
    bar
]
```

tested in chrome and firefox

### DIFF
--- a/crates/typst-kit/src/server.rs
+++ b/crates/typst-kit/src/server.rs
@@ -161,10 +161,18 @@ fn handle(req: Request, reload: bool, bucket: &Arc<RouterBucket>) -> io::Result<
 
 /// Handles for the `/` route. Serves the compiled HTML.
 fn handle_body(req: Request, reload: bool, mut body: HttpBody) -> io::Result<()> {
+    // requests for index.html without trailing / should be redirected for
+    // adding /, e.g. /foo -> /foo/, because otherwise relative links on the
+    // page doesn't work
+    let mut should_redirect_index = false;
     let (data, mime) = match &mut body {
         HttpBody::Html(html) => {
             if reload {
                 inject_live_reload_script(html);
+            }
+            let url = req.url();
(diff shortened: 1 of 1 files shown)


YDX-2147483647:
Perhaps the better way is this:

```typst
#document("index.html")[
    #link(<foo>, "foo")
]
#document("foo/index.html")[
    #link(<bar>, "bar")
]<foo>
#document("foo/bar.html")[
    bar
]<bar>
```

Relates-to: #8190

istudyatuni:
Then url in browser's bar would end with `index.html`. Though in this case typst can generate link without trailing `index.html`

Upd. Oh, linked issue is about this

laurmaedje:
While I was first a bit skeptical as it seemed a bit ad-hoc, this behavior actually directly mirrors the defaults of nginx and `python -m http.server`. So it's fine from that perspective. The diff itself, however, is a _bit_ hacky.[^1] I've created a cleaner patch in https://github.com/typst/typst/pull/8721, so I'll close this. Thanks though!

[^1]: The string checks are only an approximation of information that is already available in another place (whether we read from an implied `index.html`). I also chose HTTP status code 301 instead of 307 as that's what nginx and Python yield.

PR #8231, opened 2026-05-08
Make duplicate path errors in bundle export delayed

### LINKED ISSUE #8163: State cannot be used for generating document paths for multiple documents in bundle export
### Description

I'm experimenting with the experimental bundle export.

I'm using the current Typst `main` branch (78251637643fc4c8fbc8ec174dd622ff23e06861) and the compiler options `--features html,bundle --format bundle`.

**How to reproduce**
Consider the following Typst code:
```typ
#let doc-name-state = state("doc-name", "unknown")

#let make-doc(name) = context {
  doc-name-state.update(name)
  context {
    let filename = doc-name-state.get() + ".html"
    document(filename)[This is document #filename]
  }
}

#make-doc("index")
// #make-doc("content")
```
This creates a file at `index.html` with the content "This is document index.html", which works as expected.

To reproduce this bug, uncomment the line that's commented out in the example.

**Expected behaviour**
I would expect this to create two files: `index.html` and `content.html`, with the respective content.

**Actual behaviour**
But what actually happens is that the following error message is shown:
```
error: path `unknown.html` occurs multiple times in the bundle
  ┌─ test.typ:7:4
  │
7 │     document(filename)[This is document #filename]
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  │     │
  │     path is already in use here
  │
  = hint: document paths must be unique in the bundle
```

According to the code I've written, it should not be possible for `unknown.html` to ever be generated, but Typst tries to generate it twice.

**Note**
In this very simplified example, it is trivial to fix this issue by using `name` instead of `filename` to generate the path. In this case, `filename` inside of the generated documents then correctly shows the updated state, which indicates to me that the issue only affects the `document` call itself. But this could also be a more general issue with bundle export.

Still, this bug prevents more advanced usages of bundle export. I came across this issue while writing some code to split each section into its own document and generate the path of the document based on the section heading and its numbering. Using state is already a workaround, since `document`'s `path` (unlike `title`) does not accept `content`s.

### Operating system

Linux

### Typst version

luanst:
This PR fixes #8163: duplicate file names in bundle export are now only reported once the Typst project converges (or doesn't converge within five attempts).

The same also applies to errors from disallowed top-level elements, although the delayed errors seem way less useful there.

My approach for fixing this was rather straightforward: instead of immediately returning all the errors, delay them via `engine.sink.delay`. For my first contribution to Typst this was way easier to figure out than expected :)


laurmaedje:
Thanks!

This would probably have been fine as-is, but I decided to adjust it such that only the duplicate path error is delayed and not the other one. The problem in https://github.com/typst/typst/issues/8163 is real and deserves a fix. However, delayed errors do have some downsides as they can swallow mistakes and my mental model for when errors should be delayed is not yet fully complete, so I wanted to be conservative here. I'll have to form a more complete mental model at some point later and then we can revisit what should be delayed across the code base.

PR #8233, opened 2026-05-08
Escape potentially dangerous characters before printing user-provided strings

MDLC01:
Following the concern expressed in https://github.com/typst/typst/pull/8211#pullrequestreview-4230670199, this PR replaces potentially dangerous characters with the REPLACEMENT CHARACTER in user-provided strings. The following characters are considered dangerous characters for this purpose:
- C0 controls and DELETE from the [C0 Controls and Basic Latin] block
- C1 controls from the [C1 Controls and Latin-1 Supplement] block

[C0 Controls and Basic Latin]: https://www.unicode.org/charts/PDF/U0000.pdf
[C1 Controls and Latin-1 Supplement]: https://www.unicode.org/charts/PDF/U0080.pdf

We could also add BiDi control characters to this list, but I think this is fine for now.

Below is a prototype of an panic message a package could print to mislead a user into executing arbitrary code:

```typ
#let clear-screen = "\u{001B}[2J"
#let move-cursor-to-top-left = "\u{001B}[H"
#let red = "\u{001B}[31m"
#let reset = "\u{001B}[0m"
#let message = "Typst warning: One of your packages contains a security issue. Update it as soon as possible by executing the following command:\n\t~/.cache/typst/packages/preview/whatever/1.0.0/security-update\n\n"
#let security = clear-screen + move-cursor-to-top-left + red + message + reset

#assert(false, message: security)
```

Result (without colors):

```
Typst warning: One of your packages contains a security issue. Update it as soon as possible by executing the following command:
        ~/.cache/typst/packages/preview/whatever/1.0.0/security-update

  ┌─ /whatever
  │
7 │ #assert(false, message: security)
  │  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/crates/typst-utils/src/lib.rs
+++ b/crates/typst-utils/src/lib.rs
@@ -30,6 +30,7 @@ pub use self::version_::{TypstVersion, display_commit, version};
 #[doc(hidden)]
 pub use once_cell;
 
+use std::borrow::Cow;
 use std::fmt::{Debug, Display, Formatter};
 use std::hash::Hash;
 use std::iter::{Chain, Flatten, Rev};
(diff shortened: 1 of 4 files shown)


crates/typst-utils/src/lib.rs
```diff
@@ -472,3 +472,45 @@ macro_rules! display_possible_values {
...
+fn is_dangerous_character(c: char) -> bool {
+    matches!(c, '\u{0000}'..='\u{001F}' | '\u{007F}' | '\u{0080}'..='\u{009F}')
+}
+
+/// Escapes a user-provided string to be displayed in the terminal.
+///
+/// This replaces potentially dangerous characters with the REPLACEMENT
+/// CHARACTER.
+pub fn escape_user_string<'a>(s: &'a str) -> impl AsRef<str> + 'a {
+    enum Escaped<'a> {
```
bluebear94: Why invent our own type instead of using `Cow<'a, str>`?

MDLC01: Good catch! I initially wanted to use `EcoString`s but since they don't support `replace` with a function as pattern, I ended up reimplementing `Cow` without realizing it.

saecki:
This was discussed a bit on the discord, and for me this feels a bit to much of an ad-hoc solution. There are many more cases where problematic characters could end up in the diagnostic output, which is why I think this should be handled by `typst-kit`. Specifically either directly `codespan-reporting` or we should do some sort of filtering/replacement before printing any messages/source code.

laurmaedje:
I tend to agree with @saecki

MDLC01:
That makes sense. I opened https://github.com/typst/typst/issues/8256 to track this issue.

PR #8236, opened 2026-05-09
Bump MSRV to 1.92 and Rust in CI to 1.95

laurmaedje:
Since krilla has bumped its MSRV to 1.92, we'll have to bump it sooner or later anyways.


isuffix:
Local Clippy will finally be quiet!

laurmaedje:
I was just blissfully unaware since my machine was still on 1.91. :D

laurmaedje:
Fwiw @isuffix, you can set a per project toolchain override: https://github.com/typst/typst/pull/7687#issuecomment-3743210187

Issue #8237, opened 2026-05-09
Bump to ureq 3


laurmaedje:
There was an attempt, but it didn't go over the finish line: https://github.com/typst/typst/pull/6366

Issue #7957, opened 2026-03-07
Bump TOML crate & spec to 1.0+spec-1.1.0


YDX-2147483647:
### Description

[TOML spec v1.1.0](https://toml.io/en/v1.1.0) has been published on 2025-12-28 ([1.1.0 changelog](https://github.com/toml-lang/toml/releases/tag/1.1.0), [all changelog](https://github.com/toml-lang/toml/blob/main/CHANGELOG.md)). Most notably, an inline table can now span multiple lines.

The `toml` crate has adapted to the new spec. I suggest we update it from 0.8 (the current version, see below) to 1.0+spec-1.1.0.
According to [its changelog](https://github.com/toml-rs/toml/blob/main/crates/toml/CHANGELOG.md), the breaking change happened in 0.9.0, when serde and std were refactored as features.

https://github.com/typst/typst/blob/364ece3cb37975f509a779fc3ddc929285971d73/Cargo.toml#L133

### Use Case

Use TOML spec v1.1.0.

### Further info

See [All releases of TOML // Lib.rs](https://lib.rs/crates/toml/versions) and [`from_str` in `toml` - Rust](https://docs.rs/toml/latest/toml/fn.from_str.html)

The update might affect datetime/timezone support (#7920), because the crate bumped `toml_datetime` in 1.0.0+spec-1.1.0.

I can submit a PR, but the `toml` crate published 7 versions (1.0.{0,1,…,6}+spec-1.1.0) since 2026-02-11.
Perhaps it's better to update it later? I don't know.

### Attempt

```diff
-toml = { version = "0.8", default-features = false, features = ["parse", "display"] }
+toml = { version = "1.0", default-features = false, features = ["serde", "parse", "display"] }
```

```log
    Updating serde_spanned v0.6.8 -> v1.0.4
    Updating toml v0.8.19 -> v1.0.6+spec-1.1.0
    Updating toml_datetime v0.6.8 -> v1.0.0+spec-1.1.0
      Adding toml_parser v1.0.9+spec-1.1.0
      Adding toml_writer v1.0.6+spec-1.1.0
    Updating winnow v0.7.1 -> v0.7.15
```

#### Failed tests

The new version of the crate might have a bug for `toml-decode-integer-too-{large,small}`.

```log
❌ toml-invalid (tests/suite\loading\toml.typ:44)
  mismatched range (tests/suite\loading\toml.typ:44):
    annotated | "assets\data\bad.toml" 1:16-2:1 | \n
[...]

Issue #7834, opened 2026-02-09
Update to `icu4x` 2.x


laurmaedje:
More complicated than it appears on the surface.

Prior art: https://github.com/typst/typst/pull/7412

laurmaedje:
There was an attempt in https://github.com/typst/typst/pull/7412, but it regressed CJK handling. This should, in theory, be fixable by adjusting that patch to construct a locale-aware segmenter. Then, there still the issue that https://github.com/unicode-org/icu4x/issues/5595 is also a regression compared to the current main. However, it appears that a new segmenter is in the works and it will fix that (https://github.com/unicode-org/icu4x/issues/5595#issuecomment-4406517158). So if that lands any time soon, we might have an easy time updating.

isuffix:
FYI, I am looking into this somewhat.

MDLC01:
https://github.com/typst/typst/pull/8282 will require the next version of `icu_collator`, containing the fix to https://github.com/unicode-org/icu4x/issues/7972, when it releases.

PR #8238, opened 2026-05-09
Bump various dependencies

laurmaedje:
This bumps a few dependencies, mostly ones that don't really have any user-visible effect.

I checked the other possible upgrades and this is what's left:

- Will have user-facing changes and should be bumped in its own PR:
  - `toml` (https://github.com/typst/typst/issues/7957)
  - `two-face`
- Need to make a coordinated bump:
  - `hayro`
  - `krilla`
  - `kurbo`
  - `moxcms`
  - `png`
  - `resvg`
  - `roxmltree`
  - `tiny-skia`
  - `usvg`
- More complex to migrate:
  - `ureq` (https://github.com/typst/typst/issues/8237)
- Blocked:
  - `icu4x` (https://github.com/typst/typst/issues/7834)
- Not sure whether we'll migrate away and rather stay on known good version for now:
  - `codespan-reporting`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,7 +146,7 @@ url = "2.5"
 usvg = { version = "0.45", default-features = false, features = ["text"] }
 utf8_iter = "1.0.4"
 walkdir = "2"
-wasmi = { version = "0.51.5", default-features = false, features = ["simd"] }
+wasmi = { version = "1.0.9", default-features = false, features = ["simd"] }
 web-sys = "0.3"
 xmlwriter = "0.1.0"
 xz2 = { version = "0.1", features = ["static"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -89,8 +89,8 @@ native-tls = "0.2"
 notify = "8"
 once_cell = "1"
 open = "5.0.1"
-openssl = "0.10.72"
-oxipng = { version = "9.0", default-features = false, features = ["filetime", "parallel", "zopfli"] }
+openssl = "0.10.79"
+oxipng = { version = "10", default-features = false, features = ["filetime", "parallel", "zopfli"] }
```
saecki: @Enivex found that the `zopfli` feature doesn't enable Zopfli compression on its own, and since it's quite slow we might not want it anyway. The `zopfli` crate also doesn't appear anywhere else in the dependency tree, so this would remove it as a dependency.

PR #8246, opened 2026-05-11
Documentation symbol list improvements including submodule support

### LINKED ISSUE #8245: Some links are not displayed as links in the PDF documentation
### Description

In Section 3.9.2 (General Symbols), the link "math" in the second paragraph is not at all displayed like a link. I am unable to figure out why this is the case.

MDLC01:
This contains multiple improvements to the symbol list / grid in the documentation, including support for submodules.

Commits are independent so I can remove one if necessary.

Regarding submodule support, there is no indication that, e.g., `gender` refers to a submodule in `gender.female`. I am not sure how to indicate that in a concise way, especially in the PDF version.

Fixes https://github.com/typst/typst/issues/8245.
Closes https://github.com/typst/typst/issues/8270.
Closes https://github.com/typst/typst/issues/8274.

### DIFF
--- a/docs/assets/docs.js
+++ b/docs/assets/docs.js
@@ -467,7 +467,9 @@ function setUpSymbolFlyout(symbolGrid) {
     ".info .sym-deprecation .text",
   );
   /** @type {HTMLElement | null} */
-  const foName = flyout.querySelector(".sym-name code");
+  const foName = flyout.querySelector(".info .sym-name");
+  /** @type {HTMLElement | null} */
+  const foNameCode = flyout.querySelector(".info .sym-name code");
   /** @type {HTMLElement | null} */
   const foUnicName = flyout.querySelector(".info .unic-name");
   /** @type {HTMLElement | null} */
@@ -533,8 +535,8 @@ function setUpSymbolFlyout(symbolGrid) {
     item.ariaHasPopup = "true";
     flyoutOpenId = item.id;
     flyout.style.display = "block";
-    const name = item.id.replace(/^symbol-/, "");
     const deprecation = item.dataset.deprecation;
+    const codexName = item.dataset.codexName;
     const unicName = item.dataset.unicName;
     const latexName = item.dataset.latexName;
     const codepoint = item.dataset.value?.charCodeAt(0) ?? 0;
@@ -563,7 +565,12 @@ function setUpSymbolFlyout(symbolGrid) {
       foDeprecationText.textContent = deprecation.replaceAll("`", "");
     }
 
-    foName.textContent = name;
+    if (codexName) {
+      foName.style.display = "block";
+      foNameCode.textContent = codexName;
+    } else {
+      foName.style.display = "none";
+    }
     foUnicName.textContent = unicName ?? "";
     foCodepoint.textContent = codepointText;
     foAccent.style.display = accent ? null : "none";
@@ -585,7 +592,7 @@ function setUpSymbolFlyout(symbolGrid) {
       foMathClass.style.display = "none";
     }
 
-    const nameListener = () => copyText(name);
+    const nameListener = () => copyText(codexName);
     foCopySymNameBtn.addEventListener("click", nameListener);
     listeners.push({ target: foCopySymNameBtn, listener: nameListener });
 
@@ -988,7 +995,7 @@ function searchSymbols(symbols, query) {
   for (const element of symbols) {
     let hit = false;
     for (const s of [
-      element.id.replace(/^symbol-/, ""),
+      element.dataset.codexName,
       element.dataset.unicName,
       element.dataset.latexName,
       element.dataset.value,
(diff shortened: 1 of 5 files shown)


MDLC01:
Marking this as a draft because I may add other various fixes

docs/content/reference/library/symbols.typ
```diff
@@ -195,13 +200,19 @@
   }
 }
 
-// A list / grid of symbols.
-#let symbol-list(mod, shorthands: none, emoji: false) = {
+// Computes the entries to display in a symbol list / grid.
+#let symbol-list-entries(mod, prefix, shorthands, emoji) = {
```
laurmaedje: The `emoji` parameter is not used

docs/assets/docs.js
```diff
@@ -724,9 +724,6 @@ async function setUpGlobalSearch() {
     const items = matches.map((hit) => {
       const item = index.items[hit];
       let url = item.route;
-      if (item.kind == "Symbols") {
```
laurmaedje, line 731: The idea here was that if you search for "epsilon" in the global search, then it will also give you a hit for "General symbols" as all symbols are indexed and then it's nice to not have to write it again.

So there is a bit of tension between searches for literally "symbol" vs searches for a particular symbol.

docs/assets/docs.js
```diff
@@ -973,27 +973,55 @@ function searchSymbols(symbols, query) {
     query.replace(/^\\?[uU]\{?\+?/, "").replace(/\}$/, ""),
     16,
   );
-  let char = codepoint ? String.fromCharCode(codepoint) : null;
+  const char = codepoint && 0 <= codepoint && codepoint <= 0x10FFF
```
laurmaedje: The maximum codepoint is 0x10FFFF rather than 0x10FFF.

docs/assets/docs.js
```diff
@@ -973,27 +973,55 @@ function searchSymbols(symbols, query) {
...
+        // We have a match if all parts of the query are present somewhere in
+        // the name.
+        hit ||= queryParts.every(part => {
+          if (part.length <= 3) {
+            // For small parts, we want them to be present as is in the name.
+            return nameParts.includes(part);
+          } else {
+            // For larger parts, we allow matching anywhere within the name.
+            return processedName.includes(part);
+          }
```
laurmaedje: I'm not sure this logic is great as is. Currently, searching for `alp` yields no hits and only once I type `h`, I get a match for `alpha`. What was the motivation for these changes compared to the old algorithm?

laurmaedje: The results for `=>` also seem worse than before.

laurmaedje (commented):
Nice changes! Most commits are just fine, just the search changes I'm not sure about. We can iterate on that in this PR, but I would also be happy to merge the rest already if the search stuff was extracted into its own PR.

docs/content/reference/library/symbols.typ
```diff
@@ -258,6 +240,26 @@
...
 }
 
+// A list / grid of symbols with their names.
+#let symbol-name-list(mod, emoji: false) = {
+  let entries = symbol-list-entries(mod, none, emoji)
+  symbol-list(entries, emoji: emoji)
+}
+
+// A list / grid of symbols with their shorthands.
+#let symbol-shorthand-list(shorthands, emoji: false) = {
```
laurmaedje: The emoji parameter could be removed here.

docs/assets/docs.js
MDLC01: Nice catch!

docs/assets/docs.js
MDLC01: > What was the motivation for these changes compared to the old algorithm?

The motivation what that since the results are not ordered by relevance, having a single-letter part in the search would burry more relevant results under a bunch of less relevant results. The threshold of 3 was chosen based on what felt to work best.

For `=>`, that's a case I had not considered.

In both cases, I agree that the proposed change is not optimal. The solution would be to order search results.

docs/assets/docs.js
MDLC01: That makes sense. What about prefilling the search box unless the search is a prefix of "symbols"?

laurmaedje: Sounds fine.

docs/assets/docs.js
laurmaedje: The docs that are currently live _do_ order search results. I did not replicate this behavior partly because I didn't want to implement it (it used a JS library before), but also because I felt like it made the results harder to scan. But I might have been wrong on that. Searching for `par` now on main is much worse than in the live version...

MDLC01:
I used the newly acquired power provided to me by Jujutsu to extract the symbol list search changes into a separate branch that I will push when I have taken your suggestions into account.

Regarding the prefilling feature, the behavior I implemented is that the symbol search box is filled unless the query is a prefix of "symbols" or "emojis".

laurmaedje:
Thank you!

Issue #6590, opened 2025-07-11
Support for downloading fonts on-demand automatically


mewmew:
### Description

A GitHub issue to track the bullet already added to the [Typst roadmap](https://typst.app/docs/roadmap/). I tried to search for a meta issue that tracked this development, but was unable to find one on GitHub, so created this one.

This issue is meant to track discussions for implementing *Support for downloading fonts on-demand automatically*.

### Use Case

When using Typst, it is not too uncommon for a font or font variant to be missing from the local system (e.g. missing bold version for Greek letters, smallcaps, etc).

Using `--font-path` does resolve the issue, at least partially.

However, since the way that fonts are bundled and packaged varies across different operating systems, the user experience could be improved. Both for Typst package and template developers and Typst users.

Being able to automatically download the fonts required for a given Typst project (in a similar way that Typst packages are downloaded and cached) would be incredible. And it would also lead to reproducible builds of PDFs across operating systems (if deterministically setting `date`).

jtkiley:
I would like this a lot. My devcontainers are more complicated than I'd like, in part because I have a script to pull down and install some fonts that I like to use, and I install fontconfig via apt.

I use quarto a lot, and I've been trying to move more things from LaTeX as a PDF backend to typst. For websites, quarto can pull fonts by name from Google Fonts with no container-level customization. Having typst, which always needs external fonts (at least to my taste), be able to handle that would be really nice. I'm sure it's easier in a host OS that are already full of preferred fonts, but containers in particular benefit from this kind of automation. That's both because you can make things that just work, and it helps keep containers slim by only getting things on demand.

flashymittens:
Whatever the resolution, I hope it will be an optional feature.

haydntrowell:
How about treating fonts as a special package type with their own namespace, e.g.:

```typst
#import "@fonts/inter:4.1.0"
```

Rather than importing Typst code, importing a font package would register the contained font files with the compiler:

```typst
#import "@fonts/inter:4.1.0"

#set text(font: "Inter")
```

The package manifest could use a dedicated `[font]` section alongside the existing `[package]` metadata, for example:

```toml
[package]
name = "inter"
version = "4.1.0"

[font]
family = "Inter"
files = [
  "Inter-Regular.ttf",
  "Inter-Italic.ttf",
  "Inter-Bold.ttf",
  "Inter-BoldItalic.ttf",
]
```

[...]

Issue #2599, opened 2023-11-06
Google Fonts support


Enivex:
### Description

Being able to specify a font from Google Fonts, and have the compiler pull it from the internet if not already present.

https://developers.google.com/fonts/docs/developer_api

### Use Case

Google Fonts is a large curated database of fonts, and being able to pull fonts from there would instantly give typst users access to thousands of fonts, without having to have them installed as system fonts.

alerque:
It would also open up a huge privacy hole. You could mitigate this from the web side with a cache on *your* server side, but the CLI would leave lots of exposure. If you go down this road figure out something more generic for remote fonts such that alternative providers could be used and no network actions would be taken without the knowledge of somebody rendering on their own system.

realpixelcode:
> It would also open up a huge privacy hole. You could mitigate this from the web side with a cache on _your_ server side, but the CLI would leave lots of exposure. If you go down this road figure out something more generic for remote fonts such that alternative providers could be used and no network actions would be taken without the knowledge of somebody rendering on their own system.

Instead of fetching the respective font from Google Fonts during each compilation, Typst could instead download the font once and then include the `.ttf` file (or whatever) in the project source. There could be a dedicated `.fonts` folder where Typst would look for the desired font before downloading it.

laurmaedje:
Compiler-wise this is not desirable I think. Web-app wise we have our own fonts infrastructure, which also requires other metadata than Google fonts, privacy issues aside.

PR #8247, opened 2026-05-12
Add Auto Font Downloading Support via Google Fonts

etbala:
Implements #6590.

Downloads fonts from Google Fonts and stores them in the cache directory (`cache_dir/fonts/google/{font-family}/.ttfs`). This feature is implemented as **opt-in**. typst will not attempt to download from Google Fonts unless you add the `--auto-download-fonts` flag. Once the fonts are downloaded to the cache, they are considered system fonts (and therefore can be excluded using `--ignore-system-fonts`, and are loaded before compilation along with system installed fonts).

Because font loading is currently all done prior to compiling, instead of completely refactoring font loading to allow it during compilation, I let the full compilation complete, then check for font not found warnings. If there are any of these warnings, then we attempt to fetch them from Google Fonts, and if any fonts are successfully fetched we recompile. Is this the most efficient approach? Probably not. However, this double compilation shouldn't trigger often since once a font is downloaded it will be considered a system font, and if it isn't found in Google Fonts then nothing is recompiled.

Also, Google Font's API is case sensitive, so I pushed the lowercase normalization of font family names down into FontBook. This does alter the warning messages to show user-provided-case names instead of lowercase family names.

### DIFF
--- a/crates/typst-cli/Cargo.toml
+++ b/crates/typst-cli/Cargo.toml
@@ -66,6 +66,7 @@ features = [
     "datetime",
     "emit-diagnostics",
     "system-downloader",
+    "font-downloader",
     "watcher",
     "timer",
 ]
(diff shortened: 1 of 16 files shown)


Enivex:
Closing, since https://github.com/typst/typst/issues/2599 was already rejected

PR #8249, opened 2026-05-12
Automatically generate CSS style sheet

saecki:
This implements automatic CSS style sheet resolution and generation from CSS properties specified on individual elements. The HTML tag, aria role, and existing classes are used to generate selectors that target property groups.
A more detailed explanation of the algorithm can be found in `crates/typst-utils/src/css/resolve.rs`

The style sheet is currently inserted last in the `head` element. This will cause it to override styles with the same specificity. I originally wanted to change this around, so the style sheet would be inserted before the first user provided one. But the docs sort of rely on the generated styles to be inline styles, so they should take precedence.
This will be a bit flaky in any case, due to how selectors may change with seemingly unrelated DOM changes. But it might make sense, to have a mechanism that controls where the generated style sheet is inserted.

## Limitations
Currently the only styles that are resolved in the CSS style sheet are the ones added by native show rules. In addition there are currently only a few style properties generated by the native show rules.
Users can currently only specify an inline `style` string in `html.elem` and the typed html API. With a typed, or at least more structured CSS API, user provided style properties could be included during the resolution as well. Though I think the raw `style` attribute should stay available for specifying inline styles.

## Additions to `typst-utils`
- The `Id` and `IdVec` types from `typst-pdf` have been moved to `typst-utils
- `IdMap` and `IdRange` types as well as a `KeyFor` trait are added
- An `IdQueue` based on a binary heap is added


crates/typst-html/src/css/resolve.rs
```diff
@@ -1,38 +1,1407 @@
-use ecow::eco_format;
+//! This modules tries to resolve a [`Stylesheet`] from the CSS properties that
+//! are specified for each element. Existing classes, aria roles, and tags are
```
bastienvoirin:
```suggestion
//! are specified for each element. Existing classes, ARIA roles, and tags are
```

PR #8250, opened 2026-05-12
Within selector for introspection

### LINKED ISSUE #1926: Ancestry/inside/within selector
### Description

Currently, it is only possible to determine whether an element occurs after or before a specific element. However, it is not easily possible to determine whether an element is "inside" a specific element.

For example, take the following document:

```
= Chapter 1 <C1>
This <Label> is inside Chapter 1

== Section 1
This <OtherLabel> is inside Chapter 1 and Section 1, but not Section 2

== Section 2
And <YetAnotherLabel> is inside Chapter 1 and Section 2, but not Section 1
```

It becomes very difficult to determine whether an element is inside a chapter. This is *especially* true when combined with calls to `state()`, where one might be interested in seeing if a state update was made in a specific chapter or not.

### Use Case

The primary use-case would be to determine whether or not an element is specifically inside another element. Take the following document as an example:

```
#let s = state("state", 0)

#let valueForChapter(chapter) = {
  locate(loc => {
    let tmp = query(selector(<StateUpdate>).after(chapter), loc)
    if tmp == () {
      // The chapter has no value associated. We'll return 0
      0
    }
    else {
      s.at(tmp.first().location())      
    }
  }) 
}

Expected: 10 / Actual: #valueForChapter(<H1>)

Expected: 0 / Actual: #valueForChapter(<H2>)

Expected: 20 / Actual: #valueForChapter(<H3>)

= H1 <H1>
The value of H1 is 10
#s.update(10)
<StateUpdate>

= H2 <H2>
This chapter does not have a value associated.

= H3 <H3>
The value of H3 is 20
#s.update(20)
<StateUpdate>
```

With a `.inside()` method, it would become trivial to determine if H2 has a value associated to it or not. Without this method, one would need to use something like `selector(<StateUpdate>).after(<H2>).before(<H3>)`, which in turn becomes problematic inside loops.

laurmaedje:
This selector is super useful, so I don't want to hold it back even though it's currently only supported in introspections, not in show rules. It's the same as `before` and `after` in this regard, but unlike these, it can be supported in show rules in the future.

Closes https://github.com/typst/typst/issues/1926. The show rule part will be tracked in https://github.com/typst/typst/issues/8251.

### DIFF
--- a/crates/typst-library/src/foundations/selector.rs
+++ b/crates/typst-library/src/foundations/selector.rs
@@ -197,6 +197,9 @@ impl Selector {
 
     /// Returns a modified selector that will only match elements that occur
     /// before the first match of `end`.
+    ///
+    /// _Note:_ This selector is currently only supported with introspection
+    /// functions, not in show rules.
     #[func]
     pub fn before(
         self,
@@ -217,6 +220,9 @@ impl Selector {
 
     /// Returns a modified selector that will only match elements that occur
     /// after the first match of `start`.
+    ///
+    /// _Note:_ This selector is currently only supported with introspection
+    /// functions, not in show rules.
     #[func]
     pub fn after(
         self,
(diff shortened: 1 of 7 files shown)


bastienvoirin:
Does it mean that for document-aware outlines in bundle export we can now write
```typ
#show document: it => context {
  outline(target: selector(heading).within(here()))
  it.body
}
```

laurmaedje:
Not quite like that because you're discarding the document element, but in principle something like that can work yes. 

Alternatively, you can also directly use the document's location with `within` to find things in a specific document. What is not yet exposed is a _nice_ way to get the location of the current documentation from within it, but it's possible by querying for the last document before `here()`. So you could write something like this:

```typ
#let local-outline() = context {
  let doc = query(selector(document).before(here())).last()
  outline(target: selector(heading).within(doc.location()))
}

#document("a.pdf")[
  #local-outline()

  = A
  = B
]

#document("b.pdf")[
  #local-outline()

  = C
  = D
  = E
]
```

Worth saying that querying documents is not the best for incremental compilation ... it's not a disaster, but the outline will likely be recomputed for every edit to the document its contained in.

bastienvoirin:
Awesome, thank you! What would a better way with respect to incremental compilation look like?

laurmaedje:
> Awesome, thank you! What would a better way with respect to incremental compilation look like?

Something that doesn't let the introspection observe the document's body. Internally, there is something like `here().document()` which gives the document's location right away without looking at the document itself. The location is stable across edits while the body isn't. But I'm not sure I like that API, so it's not currently exposed.

PR #8254, opened 2026-05-12
Add sub-ranges for diagnostics and hints

isuffix:
This PR adds sub-ranges for diagnostics, although does not add any uses of them yet. A following PR will add warnings for raw syntax that will make use of sub-ranges and sub-ranges on hints.

This PR is dependent on #8188. The first new commit is "Don't use raw range spans for imported files in `definition.rs`"

**Why are sub-ranges necessary?**
When we assign spans to a concrete syntax tree, we assign targetable numbers that correspond to a range in the source. However, spans only exist on full nodes in the tree, and do not allow targeting parts of a node, or ranges of text across multiple nodes. Adding sub-ranges gives us much more flexibility to create good diagnostic messages that target exactly the desired text.

**Why not use raw-range spans instead?**
Sub-ranges are relative to the overall range for a span so they aren't invalidated when we insert new text into the syntax tree when parsing incrementally. Absolute raw-ranges would need to be searched for and globally updated on every reparse.

**What else changed?**
- The new interface in `synthesize_with` evolves a lot over all of the commits, although I am open to changing it. However, leaving `InnerNode::synthesize_with` as a separate function caused too many non-local changes, so I just inlined it to `synthesize_with`.
- I updated the `RangeMapper` quite a bit and notably removed the `Option` from the range return by using `partition_point` and returning a result on construction if any invariants weren't followed.
- I added the `SpanNumber` type that solely interacts with the `Source::range` since sub-ranges add a new way to produce ranges that would have made `Source::range` taking a whole `Span` easy to use incorrectly.
- `SpanNumber` also made it easier to distinguish the kinds of `LinkedNode::find` operation, and I made it work for raw-range spans.
- Of course, there needed to be a way to get a `SpanNumber`, and I settled on a `.get()` method returning the new unpacked `SpanKind` enum that can be explicitly matched against.
- Sub-ranges get their own type as `SubRange` and use a `NonZeroU32` for the end index for null-optimization and to ensure they are non-empty ranges. Most uses take an `Option<SubRange>` which is conviently produced by `SubRange::new` requiring a non-empty range.
- However, sub-ranges are not meant to be stored in the document content, so I added `DiagSpan` and `DiagSpanKind` to distinguish diagnostic spans with sub-ranges from normal spans.
- `DiagSpan` containing an extra 64 bits also allows us to store larger ranges than just 23 bits for externally loaded files by storing a 46-bit start index within the `Span` and another 46-bit end index in the extra space.
- And this comes with no compromise for raw-range span sizes except fixing a bug where the start index could be `2^23` instead of saturating at `2^23-1` like the end index.
- In the syntax tree, only warnings get sub-ranges since they wrap other nodes instead of encompassing a node entirely like errors.
- However in the final commit, both warnings and errors are allowed sub-ranges on hints.

### DIFF
--- a/crates/typst-library/src/diag.rs
+++ b/crates/typst-library/src/diag.rs
@@ -17,7 +17,7 @@ use std::string::FromUtf8Error;
 use az::SaturatingAs;
 use comemo::Tracked;
 use typst_syntax::package::{PackageSpec, PackageVersion};
-use typst_syntax::{Lines, Span, Spanned, SyntaxDiagnostic, VirtualRoot};
+use typst_syntax::{DiagSpan, Lines, Span, Spanned, SyntaxDiagnostic, VirtualRoot};
 use utf8_iter::ErrorReportingUtf8Chars;
 
 use crate::engine::Engine;
@@ -301,8 +301,9 @@ impl<T> Warned<T> {
 pub struct SourceDiagnostic {
     /// Whether the diagnostic is an error or a warning.
     pub severity: Severity,
-    /// The span of the relevant node in the source code.
-    pub span: Span,
+    /// The span of a relevant node in a Typst source file or the relevant byte
+    /// range of an external file.
+    pub span: DiagSpan,
     /// A diagnostic message describing the problem.
     pub message: EcoString,
     /// The trace of function calls leading to the problem.
@@ -328,10 +329,10 @@ pub enum Severity {
 
 impl SourceDiagnostic {
     /// Create a new, bare error.
-    pub fn error(span: Span, message: impl Into<EcoString>) -> Self {
+    pub fn error(span: impl Into<DiagSpan>, message: impl Into<EcoString>) -> Self {
         Self {
             severity: Severity::Error,
-            span,
+            span: span.into(),
             trace: eco_vec![],
             message: message.into(),
             hints: eco_vec![],
@@ -339,10 +340,10 @@ impl SourceDiagnostic {
     }
 
     /// Create a new, bare warning.
-    pub fn warning(span: Span, message: impl Into<EcoString>) -> Self {
+    pub fn warning(span: impl Into<DiagSpan>, message: impl Into<EcoString>) -> Self {
         Self {
             severity: Severity::Warning,
-            span,
+            span: span.into(),
             trace: eco_vec![],
             message: message.into(),
             hints: eco_vec![],
@@ -838,14 +839,14 @@ fn load_err_in_text(
     match loaded.source.v {
         LoadSource::Path(file_id) => {
             if let Some(range) = pos.range(&lines) {
-                let span = Span::from_range(file_id, range);
+                let span = DiagSpan::from_range(file_id, range);
                 return SourceDiagnostic::error(span, message);
             }
 
             // Either `ReportPos::None` was provided, or resolving the range
             // from the line/column failed. If present report the possibly
             // wrong line/column in the error message anyway.
-            let span = Span::from_range(file_id, 0..loaded.data.len());
+            let span = DiagSpan::from_range(file_id, 0..loaded.data.len());
             if let Some(pair) = pos.line_col(&lines) {
                 message.pop();
                 let (line, col) = pair.numbers();
--- a/crates/typst-library/src/diag.rs
+++ b/crates/typst-library/src/diag.rs
@@ -310,12 +310,12 @@ pub struct SourceDiagnostic {
     pub trace: EcoVec<Spanned<Tracepoint>>,
     /// Additional hints to the user.
     ///
-    /// - When the span is detached, these are generic hints. The CLI renders
-    ///   them as a list at the bottom, each prefixed with `hint: `.
+    /// - When the span is `None`, these are generic hints. The CLI renders them
+    ///   as a list at the bottom, each prefixed with `hint: `.
     ///
     /// - When a span is given, the hint is related to a secondary piece of code
     ///   and will be annotated at that code.
-    pub hints: EcoVec<Spanned<EcoString>>,
+    pub hints: EcoVec<Spanned<EcoString, DiagSpan>>,
 }
 
 /// The severity of a [`SourceDiagnostic`].
@@ -356,8 +356,12 @@ impl SourceDiagnostic {
     }
 
     /// Adds a single hint specific to a source code location to the diagnostic.
-    pub fn spanned_hint(&mut self, hint: impl Into<EcoString>, span: Span) {
-        self.hints.push(Spanned::new(hint.into(), span));
+    pub fn spanned_hint(
+        &mut self,
+        hint: impl Into<EcoString>,
+        span: impl Into<DiagSpan>,
+    ) {
+        self.hints.push(Spanned::new(hint.into(), span.into()));
     }
 
     /// Adds a single hint to the diagnostic.
@@ -367,7 +371,11 @@ impl SourceDiagnostic {
     }
 
     /// Adds a single hint specific to a source code location to the diagnostic.
-    pub fn with_spanned_hint(mut self, hint: impl Into<EcoString>, span: Span) -> Self {
+    pub fn with_spanned_hint(
+        mut self,
+        hint: impl Into<EcoString>,
+        span: impl Into<DiagSpan>,
+    ) -> Self {
         self.spanned_hint(hint, span);
         self
     }
@@ -393,7 +401,7 @@ impl From<SyntaxDiagnostic> for SourceDiagnostic {
             span,
             message,
             trace: eco_vec![],
-            hints: hints.into_iter().map(Spanned::detached).collect(),
+            hints,
         }
     }
 }
--- a/crates/typst-syntax/src/node.rs
+++ b/crates/typst-syntax/src/node.rs
@@ -1,3 +1,4 @@
+use std::cell::LazyCell;
 use std::fmt::{self, Debug, Display, Formatter};
 use std::ops::{Deref, Range};
 use std::rc::Rc;
@@ -7,7 +8,10 @@ use ecow::{EcoString, EcoVec, eco_format, eco_vec};
 use typst_utils::debug;
 
 use crate::kind::ModeAfter;
-use crate::{FileId, RangeMapper, Span, SpanKind, SpanNumber, SyntaxKind, SyntaxMode};
+use crate::{
+    DiagSpan, FileId, RangeMapper, Span, SpanKind, SpanNumber, SubRange, SyntaxKind,
+    SyntaxMode,
+};
 
 /// A node in the untyped syntax tree.
 #[derive(Clone, Eq, PartialEq, Hash)]
@@ -139,7 +143,24 @@ impl SyntaxNode {
     pub fn warn(&mut self, message: impl Into<EcoString>) {
         let kind = self.kind();
         let child = std::mem::replace(&mut self.data, Node::Leaf(EcoString::new(), kind));
-        let warn = Arc::new(WarningWrapper::new(child, message.into()));
+        let warn = Arc::new(WarningWrapper::new(child, None, message.into()));
+        self.data = Node::Warning(warn, kind);
+    }
+
+    /// Add a warning around this node at a particular sub-range of the node's
+    /// text. Panics if the range is empty or exceeds the length of the wrapped
+    /// text.
+    #[track_caller]
+    pub fn warn_at(
+        &mut self,
+        Range { start, end }: Range<usize>,
+        message: impl Into<EcoString>,
+    ) {
+        assert!(end <= self.len()); // This isn't checked by `SubRange::new`.
+        let sub_range = SubRange::new(start, end).expect("a valid sub-range");
+        let kind = self.kind();
+        let child = std::mem::replace(&mut self.data, Node::Leaf(EcoString::new(), kind));
+        let warn = Arc::new(WarningWrapper::new(child, Some(sub_range), message.into()));
         self.data = Node::Warning(warn, kind);
     }
 
@@ -293,7 +314,8 @@ impl SyntaxNode {
 
     /// Set a synthetic span for the node and all its descendants.
     pub fn synthesize(&mut self, span: Span) {
-        self.synthesize_with(0, &|_, _| span);
+        // Sub-ranges are removed since the overall range is not accurate.
+        self.synthesize_with(0, &|_, _| span, &|_, sub_range| *sub_range = None);
     }
 
     /// Set a raw range span for each node.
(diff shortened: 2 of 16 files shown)


crates/typst-syntax/src/span.rs
```diff
@@ -1,76 +1,123 @@
...
+///
+/// This is mainly used externally as an input to the
+/// [`Source::range`](crate::Source::range) method for efficiently finding the
+/// byte range of a span.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct SpanNumber(pub(crate) u64);
+
+/// The possible kinds of span.
+#[derive(Debug)]
+pub enum SpanKind {
```
saecki, line 76: The `SpanKind` is a nice touch. The previous way of checking for the different span types felt a little un-rusty in some places :)

isuffix:
Rebased off main

crates/typst-syntax/src/span.rs
```diff
@@ -33,24 +33,56 @@ use crate::FileId;
 /// performance when text is inserted somewhere in the document other than the
 /// end.
 ///
-/// Span ids are ordered in the syntax tree to enable quickly finding the node
-/// with some id:
-/// - The id of a parent is always smaller than the ids of any of its children.
-/// - The id of a node is always greater than any id in the subtrees of any left
-///   sibling and smaller than any id in the subtrees of any right sibling.
+/// Span numbers are ordered in the syntax tree to enable quickly finding the
+/// node a known span:
```
laurmaedje: small grammar error here

crates/typst-syntax/src/span.rs
```diff
@@ -223,44 +234,79 @@ impl RangeMapper {
...
-                } else if offset <= at + segment.len() {
-                    Ordering::Equal
-                } else {
-                    Ordering::Less
-                }
-            })
-            .ok()?;
-        let (at, segment) = &self.0[idx];
-        Some(segment.start + (offset - at))
+    /// Map a single offset, prefering the second index if at a boundary.
```
laurmaedje: prefering -> preferring

crates/typst-syntax/src/span.rs
```diff
@@ -223,44 +234,79 @@ impl RangeMapper {
...
+    fn map_start(&self, offset: usize) -> usize {
+        let idx = self.vec.partition_point(|&Mapping { old, new: _ }| old <= offset);
+        // Subracting by 1 is valid: vec is non-empty, index 0 has `old == 0`,
```
laurmaedje: Subracting -> Subtracting

crates/typst-syntax/src/span.rs
```diff
@@ -1,37 +1,53 @@
...
-/// Spans come in two flavors: Numbered spans and raw range spans. The
-/// `WorldExt::range` function automatically handles both cases, yielding a
-/// `Range<usize>`.
+/// Spans internally distinguish between four kinds of values, these are
+/// accessible as the [`SpanKind`] or [`DiagSpanKind`] enums via the
+/// [`Span::get`] or [`DiagSpan::get`] methods.
+/// 1. They can be detached, originating from nowhere or from the compiler
+///    itself.
+/// 2. They can be numbered values, corresponding to a node in a Typst source
+///    file's concrete syntax tree. These are the most common type of Span and
```
laurmaedje: > most commons type of "spans"

would sound more natural imo

crates/typst-syntax/src/span.rs
```diff
@@ -1,76 +1,123 @@
...
+
+/// The possible kinds of span.
+#[derive(Debug)]
+pub enum SpanKind {
+    /// A span that does not point into any file.
+    Detached,
+    /// A numbered span.
+    Number { id: FileId, num: SpanNumber },
+    /// A raw byte range in a file.
+    Range { id: FileId, start: usize, end: usize },
```
laurmaedje: Is there a particular reason to keep `start` and `end` separate instead of `range: Range<usize>`?

crates/typst-library/src/diag.rs
```diff
@@ -301,20 +301,21 @@ impl<T> Warned<T> {
...
     ///
-    /// - When the span is detached, these are generic hints. The CLI renders
-    ///   them as a list at the bottom, each prefixed with `hint: `.
+    /// - When the span is `None`, these are generic hints. The CLI renders them
+    ///   as a list at the bottom, each prefixed with `hint: `.
     ///
     /// - When a span is given, the hint is related to a secondary piece of code
     ///   and will be annotated at that code.
-    pub hints: EcoVec<Spanned<EcoString>>,
+    pub hints: EcoVec<(EcoString, Option<DiagSpan>)>,
```
laurmaedje: I see why you used `Option<DiagSpan>` here, but it feels inconsistent with other uses since `DiagSpan` also already has a detached mode. What do you think about having `Spanned<EcoString, DiagSpan>` here and making `Spanned<T, S = Span>`?

crates/typst-syntax/src/node.rs
```diff
@@ -800,12 +889,28 @@ pub struct SyntaxDiagnostic {
...
     pub message: EcoString,
     /// Additional hints to the user indicating how this issue could be avoided
     /// or worked around.
-    pub hints: EcoVec<EcoString>,
+    pub hints: EcoVec<(EcoString, Option<DiagSpan>)>,
+}
+
+/// Map a vector of hints with optional sub-ranges to one with optional
+/// diagnostic spans derived from a parent span.
+fn build_diagnostic_hints(
```
laurmaedje: could you move the helpers below their use sites, also in `span.rs`?

crates/typst-syntax/src/node.rs
laurmaedje, line 919: not new in this PR, but a blank line is missing here

crates/typst-syntax/src/source.rs
```diff
@@ -113,16 +115,32 @@ impl Source {
     ///
     /// Returns `None` if the span does not point into this source file.
     pub fn find(&self, span: Span) -> Option<LinkedNode<'_>> {
+        // TODO: Maybe rename this to `find_node` to encourage using `.range()`
+        // instead of `.find()?.range()` for errors?
```
laurmaedje: Is this really a concern? Diagnostics give you a `DiagSpan` and turning that into a `Span` for use with this function seems rather difficult.

crates/typst-syntax/src/span.rs
```diff
@@ -213,7 +405,17 @@ impl<T: Debug> Debug for Spanned<T> {
 /// source file (for instance, Typst code in a doc comment with start-of-line
 /// slashes).
 #[derive(Hash)]
-pub struct RangeMapper(Vec<(usize, Range<usize>)>);
+pub struct RangeMapper {
+    vec: Vec<Mapping>,
+    pub(crate) total: usize,
```
laurmaedje: I find this crate-public field a bit unaesthetic. I think a read-only method would be nicer.

crates/typst-syntax/src/span.rs
laurmaedje: would it make sense to use `SpanNumber` here? maybe there are more places we can add it for more type safety.

crates/typst-syntax/src/span.rs
```diff
@@ -165,6 +213,150 @@ impl Span {
...
+        }
+    }
+
+    /// Create a new diagnostic span from a source span and optional sub-range.
+    ///
+    /// The preferred interface for converting a Span into DiagSpan is using
+    /// `span.into()`, which calls this with `sub_range: None`.
+    pub(crate) fn from_source(span: Span, sub_range: Option<SubRange>) -> Self {
+        let extra = sub_range.map_or(0, |SubRange { start, end }| {
+            ((start as u64) << 32) | (end.get() as u64)
```
laurmaedje, line 261: these as casts could be made safer as `u64::from`. maybe there are more options to tighten conversion safety in other place of the file.

crates/typst-syntax/src/span.rs
```diff
@@ -165,6 +213,150 @@ impl Span {
...
+                } else {
+                    let sub_range = {
+                        let start = (extra >> 32) as u32;
+                        let end = NonZeroU32::new(extra as u32); // `as` truncates
+                        end.map(|end| SubRange { start, end })
+                    };
+                    DiagSpanKind::Number { id, num, sub_range }
+                }
+            }
+            SpanKind::Range { id, start, end: _ } if extra != 0 => {
```
laurmaedje: instead of the if guard and the `unwrap`, I would suggest unifying into one match branch and having an if let inside. I think this would be clearer even if it weren't for the unwrap.

crates/typst-syntax/src/span.rs
```diff
@@ -165,6 +213,150 @@ impl Span {
...
+            span: Span::pack(id, Span::EXTERNAL_BASE + start),
+            extra: end,
+        }
+    }
+
+    /// Create a new diagnostic span from a source span and optional sub-range.
+    ///
+    /// The preferred interface for converting a Span into DiagSpan is using
+    /// `span.into()`, which calls this with `sub_range: None`.
+    pub(crate) fn from_source(span: Span, sub_range: Option<SubRange>) -> Self {
```
laurmaedje: While I get where you're coming from, I think `from_span` would be clearer as it kind of sounds like it takes a `Source`.

crates/typst-syntax/src/span.rs
```diff
@@ -165,6 +213,150 @@ impl Span {
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub struct DiagSpan {
+    span: Span,
+    extra: u64,
+}
+
+/// The possible kinds of a diagnostic span.
+#[derive(Debug)]
+pub enum DiagSpanKind {
+    Detached,
```
laurmaedje, line 228: Even though we already have plenty of comments also on `Span`, just oneliners on these variants might be nice since `DiagSpanKind` is of the primary touch points people will have with the span API.

crates/typst-syntax/src/span.rs
```diff
@@ -165,6 +213,150 @@ impl Span {
...
+    /// The preferred interface for converting a Span into DiagSpan is using
+    /// `span.into()`, which calls this with `sub_range: None`.
+    pub(crate) fn from_source(span: Span, sub_range: Option<SubRange>) -> Self {
+        let extra = sub_range.map_or(0, |SubRange { start, end }| {
+            ((start as u64) << 32) | (end.get() as u64)
+        });
+        Self { span, extra }
+    }
+
+    /// Convert to the easier to use `DiagSpanKind` enum.
```
laurmaedje: We may want to repeat the ``To access a range, you may want to use `WorldExt::range` instead.`` comment here since this is the more direct touchpoint.

crates/typst-syntax/src/span.rs
```diff
@@ -223,61 +425,102 @@ impl RangeMapper {
...
+    /// Map a single offset, prefering the first index if at a boundary.
+    ///
+    /// This will panic if `offset` is 0.
+    fn map_end(&self, offset: usize) -> usize {
+        debug_assert_ne!(offset, 0);
+        let idx = self.vec.partition_point(|&Mapping { old, new: _ }| old < offset);
+        // Unlike `map_start`, this can yield index 0 when `offset == 0`, making
+        // `idx - 1` potentially panicking.
+        let Mapping { old, new } = &self.vec[idx - 1];
+        new + (offset - old)
```
laurmaedje, line 549: since this and map_start have the exact same code except for the closure, maybe these two can be unified through a function that takes a closure?

crates/typst-syntax/src/span.rs
isuffix: Only in an older version where I had a `derive(Copy)` annotation. I'll change to use `Range<usize>`.

crates/typst-syntax/src/source.rs
isuffix: Good point! That resolves my concern.

crates/typst-syntax/src/span.rs
isuffix: Rewording, but I'm going with: "These are the most common span type and are explained more below."

crates/typst-syntax/src/span.rs
isuffix: I don't think they should be unified. To me, the reasoning that makes doing `idx - 1` safe for `map_start` but unsafe for `map_end` is too subtle to combine into a unified function.

crates/typst-syntax/src/span.rs
isuffix: I'll move to it. I don't know that there are that many more places that would be helpful.

Moving `InnerNode::upper` or `Span::number` to `SpanNumber` feels reasonable, but it would require a lot of unpacking and re-packing whenever we do arithmetic on the numbers in `numberize`. Perhaps we could move some of the `numberize` functionality to `SpanNumber`, like a `distribute` method that returns an iterator of spans with the correct stride? Maybe for a later PR.

crates/typst-library/src/diag.rs
isuffix: I'm ok with it, but it does require something tricky for `Spanned::detached()` to work.

We could `impl<T> Spanned<T, Span> [ pub fn detached() -> Self { ... } }`, but then we need to use a different method name for `DiagSpan` (`detached_diag`?) to avoid ambiguity errors in the uses in `typst-library`.

We could also define a private trait and give a trait bound like `impl<T, S: SpanDetached> Spanned<T, S> { fn detached() -> Self { Self::DETACHED } }`. However, this makes `rustc` (not clippy) give us a [`private_bounds`](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#private-bounds) warning, which we can totally `#[allow]`, but gives me pause.

crates/typst-syntax/src/span.rs
isuffix: There are actually a host of pedantic clippy lints for this, but all of them felt overly cautious to me and IMO hurt readability. It's up to you, but I don't really like them.

crates/typst-library/src/diag.rs
isuffix: I moved to this in a new commit with `#[expect(private_bounds)]` and a new trait:
```rust
trait SpanDetached {
    const SPAN_DETACHED: Self;
}
impl SpanDetached for Span {
    const SPAN_DETACHED: Self = Self::detached();
}
impl SpanDetached for DiagSpan {
    const SPAN_DETACHED: Self = Self::detached();
}
```

isuffix:
With the new commit, this should be ready.

laurmaedje:
For fast-forward merging, it's a bit unfortunate if the last two commits go back and forth on the `hints` representation. I would suggest squashing the last two.

laurmaedje:
Thank you, nice work! :)

PR #8255, opened 2026-05-12
Implement `Debug` for `typst_kit::package` types

tingerrr:
Stumbled across these when refactoring tytanic, ironically I likely won't use them because the typst-kit stuff now does most of what I wrote there. But these seem useful to me nontheless.

### DIFF
--- a/crates/typst-kit/src/packages.rs
+++ b/crates/typst-kit/src/packages.rs
@@ -1,5 +1,6 @@
 //! Package loading.
 
+use std::fmt::Debug;
 use std::path::{Path, PathBuf};
 
 use ecow::eco_format;
@@ -30,6 +31,7 @@ use {
 /// With default configuration, this loads packages from the same sources as the
 /// CLI.
 #[cfg(feature = "system-packages")]
+#[derive(Debug)]
 pub struct SystemPackages {
     data: Option<FsPackages>,
     cache: Option<FsPackages>,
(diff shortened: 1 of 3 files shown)


crates/typst-kit/src/packages.rs
```diff
@@ -1,6 +1,9 @@
 //! Package loading.
 
-use std::path::{Path, PathBuf};
+use std::{
+    fmt::Debug,
+    path::{Path, PathBuf},
+};
```
laurmaedje: could you split these into separate lines?

crates/typst-kit/src/timer.rs
```diff
@@ -13,7 +13,7 @@ use typst_library::World;
 use typst_library::diag::{StrResult, bail};
 use typst_syntax::Span;
 
-/// Allows to record timings of function executions.
```
laurmaedje, line 16: The original wording seems fine to me.

crates/typst-kit/src/packages.rs
tingerrr: In the future we should probably add a rustfmt.toml when https://github.com/rust-lang/rustfmt/issues/6666 is resolved.

crates/typst-kit/src/timer.rs
tingerrr: Both LanguageTool and harper flag this as a mistake and suggest adding a subject or using present participle. Though, it's no major mistake, so it's likely fine to keep. I've reverted it.

crates/typst-kit/src/packages.rs
saecki: In case you're interested, you can configure the imports granularity in rust analyzer. I have this in my config far a long time:
https://github.com/saecki/dotfiles/blob/main/.config%2Fnvim%2Flua%2Fconfig%2Flsp%2Fserver%2Frust_analyzer.lua#L340-L344

laurmaedje:
Thanks!

PR #8257, opened 2026-05-12
Revert raw language tag changes temporarily with warnings for re-enabling in the future

isuffix:
This PR temporarily reverts the change to raw language tag parsing from #7337 to the current (0.14) behavior for 0.15. But since the change from #7337 is still desireable (and should land in 0.16), I've added warnings to raw blocks whose language tag would differ with the new behavior so that users are aware of the change before the switch. See [my comment on the PR](https://github.com/typst/typst/pull/7337#issuecomment-4337833842) for more motivation.

This PR is dependent on #8254, which is further dependent on #8188.
The first new commit is **"Convert single `raw-blocky` test into multiple eval tests"**

Here is an example of how the new warning will render in the CLI:
(this makes use of both warning sub-ranges and sub-ranges on hints!)
```txt
warning: no whitespace between language tag and text
  ┌─ <stdin>:1:14
  │
1 │ #let raw = ```lang.tag++ test```
  │               ^^^^^^^^^^
  │               │
  │               if the current behavior is correct, please add a space after `lang`
  │               otherwise, add a space or newline after the initial backticks
  │
  = hint: currently, Typst is treating `lang` as the language tag
  = hint: in the next version of Typst, this will change and we will treat all text until
the first whitespace as the language tag
```

You can see more warning cases in `tests/suite/raw.typ`, which I cleaned up by splitting the monolithic `raw-blocky`into individual tests alongside some new tests for this behavior. I'm not happy that there are three distinct, but similar warnings messages for the different cases, but I think each does a reasonable job of explaining the context and offering suggestions.

I also rewrote and cleaned up the top-level `raw` documentation, including greatly expanding the Syntax section and adding new sections for the language tag changes and an example I felt was missing on embedding strings in code mode using `.text` on raw syntax.

<details><summary>
You can open this dropdown to see the new raw docs (minus example images)
</summary>
<table>
<td>

## `raw`
Raw text with optional syntax highlighting.

Displays the text verbatim and in a monospace font. This is typically used to embed computer code into a document.

Text given to this element will ignore markup syntax, such as `*strong*` or `_emphasis_`, and will be displayed verbatim. If you would like to display content with a monospace font while still allowing markup syntax, instead of using [`raw`](https://typst.app/docs/reference/text/raw/#syntax), you can explicitly set the text font to a monospace font with the [`text.font`](https://typst.app/docs/reference/text/text/#parameters-font) parameter.

Raw elements are mainly produced with their [dedicated syntax](https://typst.app/docs/reference/text/raw/#syntax) by enclosing text with either one or three-plus backtick characters (``` ` ```) on both sides. When using three or more backticks, text immediately after the initial backticks will be treated as a [language tag](https://typst.app/docs/reference/text/raw/#parameters-lang) used for syntax highlighting, and the raw text begins after the first whitespace.

## Example
````typ
Adding `rbx` to `rcx` gives
the desired result.

What is ```rust fn main()``` in Rust
would be ```c int main()``` in C.

```rust
fn main() {
    println!("Hello World!");
}
```

This has ``` `backticks` ``` in it
(but the spaces are trimmed). And
``` here``` the leading space is
also trimmed.
````

You can also construct a [`raw`](https://typst.app/docs/reference/text/raw/) element programmatically from a string (and provide the language tag via the optional [`lang`](https://typst.app/docs/reference/text/raw/#parameters-lang) parameter).

```typ
#raw("fn " + "main() {}", lang: "rust")
```

If no syntax highlighting is available by default for your specified language tag (or if you want to override the built-in definition), you may provide a custom syntax specification file to the [`syntaxes`](https://typst.app/docs/reference/text/raw/#parameters-syntaxes) parameter.

## Styling
By default, the `raw` element uses the `DejaVu Sans Mono` font (included with Typst), with a smaller font size of `0.8em` (that is, 80% of the global font size). This is because monospace fonts tend to be visually larger than non-monospace fonts.

You can customize these properties with show-set rules:

````typ
// Switch to Cascadia Code for both
// inline and block raw.
#show raw: set text(font: "Cascadia Code")

// Reset raw blocks to the same size as normal text,
// but keep inline raw at the reduced size.
#show raw.where(block: true): set text(1em / 0.8)

Now using the `Cascadia Code` font for raw text.
Here's some Python code. It looks larger now:

```py
def python():
  return 5 + 5
```
````

In addition, you can customize the syntax highlighting colors by setting a custom theme through the [`theme`](https://typst.app/docs/reference/text/raw/#parameters-theme) parameter.

For complete customization of the appearance of a raw block, a show rule on [`raw.line`](https://typst.app/docs/reference/text/raw/#definitions-line) could be helpful, such as to add line numbers.

Note that in raw text, typesetting features like [hyphenation](https://typst.app/docs/reference/text/text/#parameters-hyphenate), [overhang](https://typst.app/docs/reference/text/text/#parameters-overhang), [CJK-Latin spacing](https://typst.app/docs/reference/text/text/#parameters-cjk-latin-spacing), and (for raw blocks) [justification](https://typst.app/docs/reference/model/par/#parameters-justify) will be disabled by default.

## Syntax
This function has dedicated syntax that produces a raw element in both markup and code mode. You can enclose text in one or three-plus backtick characters (``` ` ```) on both sides to make it raw. The number of backticks must be the same on both sides, and the enclosed text cannot contain a group of that many backticks in a row. Writing just two backticks (``` `` ```) produces an empty raw element.

This is intended to be similar to syntax found in other markup languages like Markdown, but has some notably different behaviors which make it more flexible: writing two backticks in a row makes it easy to produce an empty raw element, single backticks can enclose text spanning multiple lines without removing indentation, and the three-plus backtick syntax still interprets language tags when used inline.

Raw text enclosed in _single_ backticks has no way to specify a language tag and is always treated as inline for use within a paragraph, i.e. the [`block`](https://typst.app/docs/reference/text/raw/#parameters-block) parameter is `false`. However, the text may span multiple lines and will include the line breaks and indentation verbatim in its content (this is similar to the behavior of [strings](https://typst.app/docs/reference/foundations/str/)).

Raw syntax using _three or more_ backticks is meant to be more flexible and has the following smart behaviors for including text and trimming whitespace:

- **The text starts after the initial set of backticks and only ends when it encounters a group of the same number of backticks**

    This allows for enclosing text which itself contains backticks, which you can do by increasing the number of enclosing backticks until they exceed the largest group of backticks in the text itself. To enclose text containing a group of three backticks, use four or more backticks on either side.

- **The raw element can be inline or block-level based on linebreaks**

    If the text inside three or more backticks contains a linebreak character, it will produce a block-level element. If the text does not, it will produce an inline element. This sets the [`block`](https://typst.app/docs/reference/text/raw/#parameters-block) parameter to `true` or `false` accordingly.

- **Text immediately after the initial backticks (up to the first whitespace) is treated as a "language tag" for interpreting the raw text as code in that language**

    If any non-whitespace text follows on the initial line, a single (ASCII) space may be added after the language tag or immediately after the initial backticks to avoid treating the following text as part of the language tag. This space will be trimmed from the raw text if present.

    The specific rules for which text can be treated as the language tag are planned to change, and are [explained in detail below.](https://typst.app/docs/reference/text/raw/#language-tag-changes)

- **Common indentation in the text is trimmed so raw syntax can be used directly at any indentation level**

    In particular, the line which has any non-whitespace characters (including the line containing the closing backticks) that contains the least amount of leading whitespace characters determines the number of characters trimmed from the beginning of every line. Lines which are only whitespace will remove the same number of characters until they are empty, but do keep extra trailing whitespace.

    Note that this check treats tabs and spaces as equivalent characters for simplicity, and that it operates on numbers of [Unicode code points](https://www.unicode.org/glossary/#code_point), i.e. characters, not on byte lengths.

- **The initial and final lines trim whitespace specially**

    If all characters on the the initial line following the initial backticks or language tag are whitespace, they are trimmed from the raw text. Otherwise a single space character (if present) will be trimmed immediately following the initial backticks or language tag.

    If the final line (up to the closing backticks) is entirely whitespace, it is trimmed from the raw text. Otherwise, if the last non-whitespace character of the final line is a backtick, then one space character (if present) will be trimmed from the end.

    Trimming single spaces allows writing content that starts or ends with a backtick by adding a space inside the enclosing backticks. Trimming a space from the start allows using a space to separate a language tag from content on the initial line or to avoid including a language tag at all.

### Embedding strings with raw syntax
A common use-case for raw syntax is to embed data as strings with formatting by accessing the `.text` field on raw content to get the underlying string. This may also be paired with the [`bytes`](https://typst.app/docs/reference/foundations/bytes/) constructor to convert the string to bytes.

````typ
An inline YAML dictionary via `.text`

#yaml(bytes(
  ```yaml
  Magic:
    limited-by: Mana
  Pokémon:
    limited-by: Energy
  Yu-Gi-Oh:
    limited-by: false
  ```.text
  //  ^^^^ used as a string
))
````

### Language tag changes

When using raw syntax with three or more backticks, text immediately after the initial backticks (up to the first whitespace) is treated as a [language tag](https://typst.app/docs/reference/text/raw/#parameters-lang). However in the current version of Typst, only text that would be a valid Typst identifier is treated as the language tag. So if a raw block starts with `C++`, the identifier `C` will be the language tag, and the raw text will start with `++`. If a raw block starts with `++C`, it will have no language tag and the raw text will start with `++C`.

If you do want to use something like `C++`, `++C`, or `html.j2` as a language tag in the current version, you will need to use the [`lang`](https://typst.app/docs/reference/text/raw/#parameters-lang) parameter and pass in the raw text as a string. This can be aided by using a `set` rule, or by writing a raw block with no tag and using `.text` to access the contents as a string, but it can still be annoying.

In the next version of Typst, _all text_ up to the first whitespace or backtick will be treated as the language tag, so tags like `C++` or `html.j2` or `$!#%@` can be written without issue. Although tags including spaces or backticks will still need to be set manually via the [`lang`](https://typst.app/docs/reference/text/raw/#parameters-lang) parameter.

However, this may cause some documents relying on the behavior of raw text starting like `C++` or `++C` to change, so we are giving a warning for those cases. To silence the warning and keep the current behavior, you can add a single space after the intended language tag or, if no tag is intended, after the initial backticks.

</td>
</table>
</details> 

Please comment if you have any feedback on the warning messages or documentation!

### DIFF
--- a/crates/typst-library/src/text/raw.rs
+++ b/crates/typst-library/src/text/raw.rs
@@ -31,13 +31,19 @@ use crate::visualize::Color;
 /// Raw text with optional syntax highlighting.
 ///
 /// Displays the text verbatim and in a monospace font. This is typically used
-/// to embed computer code into your document.
+/// to embed computer code into a document.
 ///
-/// Note that text given to this element cannot contain arbitrary formatting,
-/// such as `[*strong*]` or `[_emphasis_]`, as it is displayed verbatim. If
-/// you'd like to display any kind of content with a monospace font, instead of
-/// using @raw, you should change its font to a monospace font using the @text
-/// function.
+/// Text given to this element will ignore markup syntax, such as `[*strong*]`
+/// or `[_emphasis_]`, and will be displayed verbatim. If you would like to
+/// display content with a monospace font while still allowing markup syntax,
+/// instead of using @raw, you can explicitly set the text font to a monospace
+/// font with the @text.font parameter.
+///
+/// Raw elements are mainly produced with their @raw:syntax[dedicated syntax] by
+/// enclosing text with either one or three-plus backtick characters (``` ` ```)
+/// on both sides. When using three or more backticks, text immediately after
+/// the initial backticks will be treated as a @raw.lang[language tag] used for
+/// syntax highlighting, and the raw text begins after the first whitespace.
 ///
 /// = Example <example>
 /// ````example
@@ -60,32 +66,16 @@ use crate::visualize::Color;
 /// ````
 ///
 /// You can also construct a @raw element programmatically from a string (and
-/// provide the language tag via the optional @raw.lang[`lang`] argument).
+/// provide the language tag via the optional @raw.lang[`lang`] parameter).
 ///
 /// ```example
 /// #raw("fn " + "main() {}", lang: "rust")
 /// ```
 ///
-/// = Syntax <syntax>
-/// This function also has dedicated syntax. You can enclose text in 1 or 3+
-/// backticks (``` ` ```) to make it raw. Two backticks produce empty raw text.
-/// This works both in markup and code.
-///
-/// When you use three or more backticks, you can additionally specify a
-/// language tag for syntax highlighting directly after the opening backticks.
-/// Within raw blocks, everything (except for the language tag, if applicable)
-/// is rendered as is, in particular, there are no escape sequences.
-///
-/// The language tag ends at the first whitespace or backtick. If your text
-/// starts with something that looks like an identifier, but no syntax
-/// highlighting is needed, start the text with a single space (which will be
-/// trimmed) or use the single backtick syntax. If your text should start or end
-/// with a backtick, put a space before or after it (it will be trimmed).
-///
 /// If no syntax highlighting is available by default for your specified
 /// language tag (or if you want to override the built-in definition), you may
 /// provide a custom syntax specification file to the @raw.syntaxes[`syntaxes`]
-/// field.
+/// parameter.
 ///
 /// = Styling <styling>
 /// By default, the `raw` element uses the `DejaVu Sans Mono` font (included
@@ -114,15 +104,150 @@ use crate::visualize::Color;
 /// ````
 ///
 /// In addition, you can customize the syntax highlighting colors by setting a
-/// custom theme through the @raw.theme[`theme`] field.
+/// custom theme through the @raw.theme[`theme`] parameter.
 ///
 /// For complete customization of the appearance of a raw block, a show rule on
 /// @raw.line could be helpful, such as to add line numbers.
 ///
-/// Note that, in raw text, typesetting features like
+/// Note that in raw text, typesetting features like
 /// @text.hyphenate[hyphenation], @text.overhang[overhang],
-/// @text.cjk-latin-spacing[CJK-Latin spacing] (and @par.justify[justification]
-/// for @raw.block[raw blocks]) will be disabled by default.
+/// @text.cjk-latin-spacing[CJK-Latin spacing], and (for raw blocks)
+/// @par.justify[justification] will be disabled by default.
+///
+/// = Syntax <syntax>
+/// This function has dedicated syntax that produces a raw element in both
+/// markup and code mode. You can enclose text in one or three-plus backtick
+/// characters (``` ` ```) on both sides to make it raw. The number of backticks
+/// must be the same on both sides, and the enclosed text cannot contain a group
+/// of that many backticks in a row. Writing just two backticks (``` `` ```)
+/// produces empty raw text.
+///
+/// Notable differences from Markdown include that single backticks can enclose
+/// text spanning multiple lines without removing indentation, and that the
+/// three-plus backtick syntax still interprets language tags when used inline.
+///
+/// Raw text enclosed in _single_ backticks has no way to specify a language tag
+/// and is always treated as inline for use within a paragraph, i.e. the
+/// @raw.block[`block`] parameter is `{false}`.
+///
+/// Raw syntax using _three or more_ backticks has the following properties:
+///
+/// - *After the initial backticks, the raw block is only terminated by a
+///   sequence of the same number of backticks*
+///
+///   To include text containing a sequence of backticks, the initial and final
+///   backticks must have at least one more backtick than the sequence.
+///
+/// - *If the raw text contains a linebreak, it will be block-level, otherwise
+///   it will be inline*
+///
+///   This sets the @raw.block[`block`] parameter to `{true}` or `{false}`
+///   accordingly.
+///
+/// - *Text immediately after the initial backticks, up to the first whitespace,
+///   is treated as a _language tag_ used for syntax highlighting*
+///
+///   The specific rules for which text can be treated as the language tag are
+///   planned to change, and are @raw:language-tag-changes[explained in detail
+///   below.]
+///
+/// - *The initial and final lines have special trimming behavior*
+///
+///   For the initial line, if all characters following the initial backticks or
+///   language tag are whitespace, the entire line will be trimmed. However, if
+///   there are non-whitespace characters on that line, only a single space
+///   immediately following the initial backticks or language tag will be
+///   trimmed if present.
+///
+///   If the final line is entirely whitespace up to the closing backticks, it
+///   will be trimmed. Otherwise, if the last non-whitespace character of the
+///   final line is a backtick, then one space character will be trimmed from
+///   the end of the line if present.
+///
+/// - *Common indentation at the beginning of lines is trimmed*
+///
+///   Typst will remove initial whitespace at the beginning of lines in the raw
+///   text that is shared between all lines, i.e. common indentation. Although
+///   this excludes text on the line with the initial backticks.
+///
+///   Typst first finds the line with the fewest initial whitespace characters
+///   that contains some non-whitespace characters, including the line with the
+///   closing backticks. Then Typst trims characters from every line equal to
+///   the number of initial whitespace characters in that line. Lines which are
+///   only whitespace will remove the same number of characters until they are
+///   empty, but will keep any extra trailing whitespace.
+///
+///   #let code-point = "https://www.unicode.org/glossary/#code_point"
+///
+///   Note that this check treats tabs and spaces as equivalent characters for
+///   simplicity, and that it operates on numbers of #link(code-point)[Unicode
+///   code points], i.e. characters, not on byte lengths.
+///
+/// These properties of the three-plus backtick syntax allow for some use cases
+/// that may not be obvious:
+///
+/// - To write text containing a sequence of backticks, enclose it with one or
+///   more backticks than the sequence:
+///   ````` ```` enclosed```backticks```` `````
+///
+/// - To write text that starts or ends with a backtick, add a space inside the
+///   opening and closing backticks: ```` ``` `backticks` ``` ````
+///
+/// - To write inline text highlighted with a language tag, add a space between
+///   the language tag and the text ````rust ```rust fn main() {}``` ````
+///
+/// - To write inline text without any language tag, add a space after the
+///   initial backticks: ```` ``` text``` ```` or use the single backtick
+///   syntax: ``` `text` ```
+///
+/// == Embedding strings with raw syntax <embedding-strings>
+/// A common use-case for raw syntax is to embed data as strings with formatting
+/// by accessing the `.text` field on raw content to get the underlying string.
+/// This may also be paired with the @bytes constructor to convert the string to
+/// bytes.
+///
+/// ````example
+/// An inline YAML dictionary via `.text`
+///
+/// #yaml(bytes(
+///   ```yaml
+///   Magic:
+///     limited-by: Mana
+///   Pokémon:
+///     limited-by: Energy
+///   Yu-Gi-Oh:
+///     limited-by: false
+///   ```.text
+///   //  ^^^^ used as a string
+/// ))
+/// ````
+///
+/// == Language tag changes <language-tag-changes>
+///
+/// When using raw syntax with three or more backticks, text immediately after
+/// the initial backticks (up to the first whitespace) is treated as a
+/// @raw.lang[language tag]. However in the current version of Typst, only text
+/// that would be a valid Typst identifier is treated as the language tag. The
+/// first character not valid for an identifier will be interpreted as starting
+/// the raw text.
+///
+/// For example, in the current verion of Typst, if a raw block starts with
+/// `C++`, the identifier `C` will be the language tag, and the raw text will
+/// start with `++`. If a raw block starts with `++C`, it will have no language
+/// tag and the raw text will start with `++C`.
+///
+/// To use language tags that are not valid as identifiers in the current
+/// version of Typst, you must use the @raw.lang[`lang`] parameter, either by
+/// calling the constructor with a string: ```typ #raw("text", lang: "...")```,
+/// or by writing a set rule: ```typ #set raw(lang: "...")```.
+///
+/// In the next version of Typst, _all text_ up to the first whitespace or
+/// backtick will be treated as the language tag, allowing a wider character set
+/// for language tags. Tags including spaces or backticks will still need to be
+/// set manually via the @raw.lang[`lang`] parameter.
+///
+/// Typst will alert you if your raw blocks will be interpreted differently in
+/// the next Typst version by emitting a warning.
 #[elem(
     scope,
     title = "Raw Text / Code",
@@ -196,7 +321,10 @@ pub struct RawElem {
     #[default(false)]
     pub block: bool,
 
-    /// The language to syntax-highlight in.
+    /// The language to interpret the raw text as for syntax highlighting.
+    ///
+    /// In @html[HTML export], this sets the `data-lang` attribute of the
+    /// generated @html.code element.
     ///
     /// Apart from typical language tags known from Markdown, this supports the
     /// `{"typ"}`, `{"typc"}`, and `{"typm"}` tags for
--- a/crates/typst-syntax/src/lexer.rs
+++ b/crates/typst-syntax/src/lexer.rs
@@ -186,8 +186,7 @@ impl Lexer<'_> {
 /// Raw.
 impl Lexer<'_> {
     /// We parse entire raw segments in the lexer as a convenience to avoid
-    /// going to and from the parser for each raw section. See comments in
-    /// [`Self::blocky_raw`] and [`Self::inline_raw`] for specific details.
+    /// going to and from the parser for each raw section.
     fn raw(&mut self) -> (SyntaxKind, SyntaxNode) {
         let start = self.s.cursor() - 1;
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by laurmaedje** [comment]
The output looks great! As a side note: For stacked PRs, it would be great if you could adjust the base branch to the branch it's based on so that there is a clear diff and dependency chain.

**Comment by isuffix** [comment]
Rebased off main and fixed the conflict in "Move raw lexer functions into their own impl block"

**Comment by isuffix** [comment]
Rebased off main to simplify the commits :)

File: crates/typst-syntax/src/lexer.rs [hunk 1]
```diff
@@ -221,31 +220,120 @@ impl Lexer<'_> {
...
+    /// In the next version of Typst, we will parse all text up to the first
+    /// whitespace or backtick, so tags like `C++` or `html.j2` or `$!#%@` can
+    /// be written without issue
+    ///
+    /// However, this may cause some documents relying on the behavior of raw
+    /// text starting like `C++` or `++C` to change, so we are giving a warning
+    /// for those cases.
+    fn raw_lang_tag<'a>(s: &mut Scanner<'a>) -> (Option<&'a str>, Option<usize>) {
+        let start = s.cursor();
+        // TODO: Change the backtick behavior?
```
laurmaedje: What does this TODO mean?

crates/typst-syntax/src/lexer.rs
```diff
@@ -221,31 +220,120 @@ impl Lexer<'_> {
...
+            // Future tag is always longer than current tag, if empty, we have
+            // no current tag either.
+            return (None, None);
+        }
+        s.jump(start);
+        let mut tag = None;
+        if s.eat_if(is_id_start) {
+            s.eat_while(is_id_continue);
+            tag = Some(s.from(start));
+        }
```
laurmaedje: This could be written fairly nicely with `.then(|| ..)`

crates/typst-syntax/src/lexer.rs
```diff
@@ -221,31 +220,120 @@ impl Lexer<'_> {
...
-        push_raw(SyntaxKind::RawDelim, &self.s);
-
-        if backticks >= 3 {
-            self.blocky_raw(end - backticks, &mut push_raw);
+        let mut nodes = vec![SyntaxNode::leaf(SyntaxKind::RawDelim, ticks)];
+
+        let mut tag = None;
+        let mut diff_future_tag_len = None;
+        if ticks.len() >= 3 {
+            (tag, diff_future_tag_len) = Self::raw_lang_tag(&mut s);
```
laurmaedje: I think I would inline the function as I'm not sure it's really a well segmented conceptually. The signature is a bit overloaded with the three return values. And it doesn't do the full parsing either since the opening backticks are already processed.

Since `fn raw` would become quite long, I would suggest moving the hint logic into a separate function instead.

crates/typst-syntax/src/lexer.rs
```diff
@@ -221,31 +220,120 @@ impl Lexer<'_> {
...
+                let tag_range = backticks..backticks + tag.len();
+                raw.hint_at(tag_range.clone(), eco_format!(
+                    "if the current behavior is correct, please add a space after `{tag}`"
+                ));
+                raw.hint_at(
+                    tag_range,
+                    "otherwise, add a space or newline after the initial backticks",
+                );
+            } else {
+                raw.warn_at(future_range, "no whitespace before the raw text");
```
laurmaedje: Drop the "the", it's cleaner.

crates/typst-syntax/src/lexer.rs
```diff
@@ -221,31 +220,120 @@ impl Lexer<'_> {
...
+            // Empty with no tag/ws is only possible with two backticks, which
+            // is handled above.
+            raw.warn("empty raw text");
+            raw.hint(eco_format!("Typst is treating `{tag}` as the language tag"));
+            let tag_range = backticks..backticks + tag.len();
+            raw.hint_at(
+                tag_range,
+                "to treat this as text, add a space after the initial backticks",
+            );
+            // TODO: Should we hint to add a space or linebreak to avoid the
```
laurmaedje: I don't think we need to.

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+/// @text.cjk-latin-spacing[CJK-Latin spacing], and (for raw blocks)
+/// @par.justify[justification] will be disabled by default.
+///
+/// = Syntax <syntax>
+/// This function has dedicated syntax that produces a raw element in both
+/// markup and code mode. You can enclose text in one or three-plus backtick
+/// characters (``` ` ```) on both sides to make it raw. The number of backticks
+/// must be the same on both sides, and the enclosed text cannot contain a group
+/// of that many backticks in a row. Writing just two backticks (``` `` ```)
+/// produces an empty raw element.
```
laurmaedje: produces an empty raw element -> produces empty raw text

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///
+/// This is intended to be similar to syntax found in other markup languages
+/// like Markdown, but has some notably different behaviors which make it more
+/// flexible: writing two backticks in a row makes it easy to produce an empty
+/// raw element, single backticks can enclose text spanning multiple lines
+/// without removing indentation, and the three-plus backtick syntax still
+/// interprets language tags when used inline.
```
laurmaedje: I think this is a bit much. Maybe we could trim it down to this:
```
/// Notable differences to Markdown include that single backticks can enclose
/// text spanning multiple lines without removing indentation and that
/// three-plus backtick syntax still interprets language tags when used inline.
```

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///
+/// Raw text enclosed in _single_ backticks has no way to specify a language tag
+/// and is always treated as inline for use within a paragraph, i.e. the
+/// @raw.block[`block`] parameter is `{false}`. However, the text may span
+/// multiple lines and will include the line breaks and indentation verbatim in
```
laurmaedje: This feels redundant with the previous paragraph. I'd probably keep this and drop the previous one entirely.

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+/// its content (this is similar to the behavior of @str[strings]).
+///
+/// Raw syntax using _three or more_ backticks is meant to be more flexible and
+/// has the following smart behaviors for including text and trimming
+/// whitespace:
```
laurmaedje: I don't like the "is meant" as this is not the place to defend design rationale. I would also prefer a more neutral tone over the word "smart". We could just trim this down to: `Raw syntax using _three or more_ backticks has the following properties:`

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///
+/// - *The text starts after the initial set of backticks and only ends when it
```
reknih: The items in this list do not share a level of abstraction. For example, the first item is the full behavioral definition where the paragraph below is essentially guidance on how to escape a sequence of backticks (conceptually a `details` element). However, the second paragraph is more like a heading with a full definition below.

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///   If the text inside three or more backticks contains a linebreak character,
+///   it will produce a block-level element. If the text does not, it will
+///   produce an inline element. This sets the @raw.block[`block`] parameter to
+///   `{true}` or `{false}` accordingly.
+///
+/// - *Text immediately after the initial backticks (up to the first whitespace)
+///   is treated as a "language tag" for interpreting the raw text as code in
+///   that language*
+///
+///   If any non-whitespace text follows on the initial line, a single (ASCII)
```
reknih: ASCII is not a concept we introduce in the docs and inconsistent with the use of _linebreak character_ above. I'd recommend sticking to _space_ or _space character._

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///   trimmed from the beginning of every line. Lines which are only whitespace
+///   will remove the same number of characters until they are empty, but do
+///   keep extra trailing whitespace.
+///
+///   #let code-point = "https://www.unicode.org/glossary/#code_point"; Note that
+///   this check treats tabs and spaces as equivalent characters for simplicity,
+///   and that it operates on numbers of #link(code-point)[Unicode code points],
+///   i.e. characters, not on byte lengths.
+///
+/// - *The initial and final lines trim whitespace specially*
```
reknih: Doesn't this kinda belong in the last point?

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///   tag. This space will be trimmed from the raw text if present.
+///
+///   The specific rules for which text can be treated as the language tag are
+///   planned to change, and are @raw:language-tag-changes[explained in detail
+///   below.]
+///
+/// - *Common indentation in the text is trimmed so raw syntax can be used
+///   directly at any indentation level*
+///
+///   In particular, the line which has any non-whitespace characters (including
```
reknih: Avoid parentheses in favor of commas or changing word order

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///   `{true}` or `{false}` accordingly.
+///
+/// - *Text immediately after the initial backticks (up to the first whitespace)
+///   is treated as a "language tag" for interpreting the raw text as code in
+///   that language*
+///
+///   If any non-whitespace text follows on the initial line, a single (ASCII)
+///   space may be added after the language tag or immediately after the initial
+///   backticks to avoid treating the following text as part of the language
+///   tag. This space will be trimmed from the raw text if present.
```
reknih: We try to avoid passive voice. Common replacements include "Typst will" or "Typst does", "You can", "You must", or imperative voice, depending on the use case.

crates/typst-library/src/text/raw.rs
```diff
@@ -114,15 +104,152 @@ use crate::visualize::Color;
...
+///   directly at any indentation level*
+///
+///   In particular, the line which has any non-whitespace characters (including
+///   the line containing the closing backticks) that contains the least amount
+///   of leading whitespace characters determines the number of characters
+///   trimmed from the beginning of every line. Lines which are only whitespace
+///   will remove the same number of characters until they are empty, but do
+///   keep extra trailing whitespace.
+///
+///   #let code-point = "https://www.unicode.org/glossary/#code_point"; Note that
```
reknih:
```suggestion
///   #let code-point = "https://www.unicode.org/glossary/#code_point"
///
///   Note that this check treats tabs and spaces as equivalent characters
```

reknih (commented):
I feel like overall, the additions could be tightened a bit and focus more on being a reference for the language and syntax. In particular, the docs should not be preoccupied with justifying a design decision (though I can see why doing this for the reasoning behind the planned changes is sensible).

I think that how these docs deal with terminology can be improved. For example, I'd like to propose raw text as the term for the text that is actually rendered (or available in `.text` as a string) and raw block for the syntactic construct.

I have tried to boil this down a bit in [this gist](https://gist.github.com/reknih/c9acdc3f1294503c24ad2289209bb6ca). The biggest weakness of my version is that in the triple backtick section, I still make claims about _raw blocks_ only applicable to the triple backtick version without having a term to qualify that. Tell me if you think this leans a bit too terse, but the best outcome could lie somewhere inbetween.

laurmaedje:
> For example, I'd like to propose raw text as the term for the text that is actually rendered (or available in .text as a string) and raw block for the syntactic construct.

I would (and do typically) use "raw text" for a generic raw element since "raw block" sounds like it is a raw element with `block: true`.

crates/typst-syntax/src/lexer.rs
isuffix: The new behavior is to parse text up to the first whitespace or backtick as the language tag, but I don't know if the backtick part of that was ever discussed. It didn't have any rationale in the prior PR, so I wasn't sure what the purpose was.

Like, it's theoretically useful to have a language tag like ``a`b`c``, and it seems a bit arbitrary to say that it should stop at the backtick when `!@#$%^&` would be a totally valid language tag.

But a backtick would always have to be preceded by some other character to be part of the tag since you can't make a language tag start with a backtick, as continuous backticks would just be treated as the opening delimiter, and whitespace would be interpreted as ending the language tag.

crates/typst-library/src/text/raw.rs
isuffix: :sweat_smile: This was an artifact that I forgot to remove from fighting with my automatic line wrapper not recognizing Typst syntax.

isuffix:
Pushed the updated implementation in the lexer, still working on the prose.

crates/typst-library/src/text/raw.rs
isuffix: I don't think so. The previous point is only meant to be about common indentation, and your gist also only leads with "initial whitespace," while the behavior of the first and last lines is about more than that.

And in your gist, the paragraph you combined this with really feels like it should be separated as another header to me if we want it to better match the other bullet points:
> The initial and final lines have special trimming behavior. For raw blocks with non-whitespace characters on the initial line, ...

vs.
> **The initial and final lines have special trimming behavior**
>
> For raw blocks with non-whitespace characters on the initial line, ...

isuffix:
The feedback is very helpful, thank you! I've updated the raw documentation by including or rewriting the suggestions and adding some new parts.

Apologies for rebasing off main at the same time as the update, [the most recent push](https://github.com/typst/typst/compare/fc1f690fa517c3e32fd8c1f609ce344076722c92..e35ea5790e5b62b7f009accaec6499de76ec7a51) should now contain only the diff to `raw.rs`.

I agree that the three-plus backtick list should be more purely reference text, however I feel like dropping the inline examples would hurt clarity, so I added a new list of _use cases that may not be obvious_ following the reference list. I wasn't sure if this should be its own section or not, or maybe if it could be a dropdown in HTML. Would appreciate thoughts :)

Also in the three-plus backtick list, I moved the _initial and final lines_ item to come after the _language tag_ item so that I could remove the paragraph on trimming a single space from the _language tag_ item, since it was just duplicating the explanation of the initial line behavior.

I've also kept the `C++` and `++C` example in the _language tag changes_ section, I think it really helps the clarity to have an example there, and should make it more useful if someone gets confused by the warning and goes to that section to figure it out.

crates/typst-library/src/text/raw.rs
reknih: Fair enough

laurmaedje:
Thank you!

Issue #1474, opened 2023-06-11
Render trailing punctuation marks after a block equation on the same line as the equation


MDLC01:
### The problem

When a block equation is part of a sentence and followed by a punctuation mark, like a period, you have to put it *inside* the equation for it to not be rendered on the line bellow the equation (see examples bellow). This is frustrating because (1) the period is not technically part of the equation, and (2) the math font is used instead of the text font.

#### Example 1

Notice how the period is *inside* the equation in the Typst code, and rendered correctly: on the same line as the equation, right after it.

```typst
Binomial coefficients are defined for all $n in NN$ and $k in [|0, n|]$ by the formula
$ binom(n, k) := (n!) / (k! (n - k)!). $
```

![image](https://github.com/typst/typst/assets/57839069/bc9dd8ad-1eaf-4c70-8433-e8f42f24cfba)

#### Example 2

Notice how the period is *atfer* the equation in the Typst code, and rendered incorrectly: on the line bellow the equation.

```typst
Binomial coefficients are defined for all $n in NN$ and $k in [|0, n|]$ by the formula
$ binom(n, k) := (n!) / (k! (n - k)!) $.
```

![image](https://github.com/typst/typst/assets/57839069/8f699080-6929-420a-817b-1a985c81902a)

### The solution

Although fixing #1125 would likely provide a way to solve (2), it still won't solve (1).

My suggestion it that punctuation characters (which I recognize might not be well defined for non-latin languages) that are right after the closing `$` of a block equation should be displayed on the same line as the equation. A potential problem with this specific solution would be languages that require a space before some punctuation marks (like french with the colon), and it could maybe increase the complexity of the parser.

### Use case

Having cleaner markup because your punctuation marks not included as part of an equation just for them to be rendered at the right place.

Enivex:
See my comment here https://github.com/typst/typst/issues/3936#issuecomment-2057008693

onomou:
I wrote an example here https://github.com/typst/typst/issues/3936#issuecomment-2057470137 that illustrates one logical place that a resolution to this issue would be useful.

MDLC01:
This was recently [discussed on Discord](https://discord.com/channels/1054443721975922748/1176478139757629563/1337417057670664285). In particular, [this Stack Exchange answer](https://tex.stackexchange.com/questions/7542/for-formal-articles-should-a-displayed-equation-be-followed-by-a-punctuation-to/170691#170691) provides additional context and recommandations for ponctuations after equations.

[...]

PR #8258, opened 2026-05-12
Use math in docs!

MDLC01:
With https://github.com/typst/typst/pull/7436 now merged, we can use math equations in the documentation.

I took the opportunity to properly define the less common mathematical functions from `calc`.

As suggested by Laurenz on Discord,[^1] I used New Computer Modern Math for the math font. Notably, this is a serif font, while the text font is a sans serif font. Noto Sans Math is a good candidate for a font that looks like the text font.

Example of a result (here in the PDF documentation):
The doc for `calc.norm`, containing a block mathematical equation defining the p-norm and the special case of the Euclidean norm when p = 2.

[^1]: https://discord.com/channels/1054443721975922748/1088371867913572452/1496184132445929613

### DIFF
--- a/crates/typst-library/src/foundations/calc.rs
+++ b/crates/typst-library/src/foundations/calc.rs
@@ -155,7 +155,7 @@ pub fn pow(
     }
 }
 
-/// Raises a value to some exponent of e.
+/// Raises a value to some exponent of $e$.
 ///
 /// ```example
 /// #calc.exp(1)
@@ -201,9 +201,9 @@ pub fn sqrt(
     Ok(value.v.float().sqrt())
 }
 
-/// Calculates the real nth root of a number.
+/// Calculates the real $n$#super[th] root of a number.
 ///
-/// If the number is negative, then n must be odd.
+/// If the number is negative, then $n$ must be odd.
 ///
 /// ```example
 /// #calc.root(16.0, 4) \
@@ -213,7 +213,7 @@ pub fn sqrt(
 pub fn root(
     /// The expression to take the root of.
     radicand: f64,
-    /// Which root of the radicand to take.
+    /// The value of $n$.
     index: Spanned<i64>,
 ) -> SourceResult<f64> {
     if index.v == 0 {
@@ -300,7 +300,7 @@ pub fn tan(
 /// ```
 #[func(title = "Arcsine")]
 pub fn asin(
-    /// The number whose arcsine to calculate. Must be between -1 and 1.
+    /// The number whose arcsine to calculate. Must be between $-1$ and $1$.
     value: Spanned<Num>,
 ) -> SourceResult<Angle> {
     let val = value.v.float();
@@ -318,7 +318,7 @@ pub fn asin(
 /// ```
 #[func(title = "Arccosine")]
 pub fn acos(
-    /// The number whose arccosine to calculate. Must be between -1 and 1.
+    /// The number whose arccosine to calculate. Must be between $-1$ and $1$.
     value: Spanned<Num>,
 ) -> SourceResult<Angle> {
     let val = value.v.float();
@@ -344,7 +344,8 @@ pub fn atan(
 
 /// Calculates the four-quadrant arctangent of a coordinate.
 ///
-/// The arguments are `(x, y)`, not `(y, x)`.
+/// Note that contrary to the usual convention, this function accepts $(x, y)$
+/// instead of $(y, x)$.
 ///
 /// ```example
 /// #calc.atan2(1, 1) \
@@ -352,9 +353,9 @@ pub fn atan(
 /// ```
 #[func(title = "Four-quadrant Arctangent")]
 pub fn atan2(
-    /// The X coordinate.
+    /// The $x$ coordinate.
     x: Num,
-    /// The Y coordinate.
+    /// The $y$ coordinate.
     y: Num,
 ) -> Angle {
     Angle::atan2(y.float(), x.float())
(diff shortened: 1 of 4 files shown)


crates/typst-library/src/foundations/calc.rs
```diff
@@ -344,24 +344,33 @@ pub fn atan(
 
 /// Calculates the four-quadrant arctangent of a coordinate.
 ///
-/// The arguments are `(x, y)`, not `(y, x)`.
+/// The four-quadrant arctangent of $(x, y)$ is defined as the argument of the
+/// complex number $x + i y$.
```
T0mstone, line 348: I think this should also specify the branch cut, i.e. the function's range.

MDLC01: Good catch! I thought `angle`s would be considered up to equivalence but apparently not (e.g., `-180deg != 180deg`). I'll change that.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -549,18 +589,23 @@ pub fn fact(
...
-///   "perm"(5, 3) &= #calc.perm(5, 3) $
+/// #calc.perm(5, 3)
 /// ```
 #[func(title = "Permutation")]
 pub fn perm(
-    /// The base number. Must be non-negative.
+    /// The value of $n$. Must be non-negative.
     base: u64,
-    /// The number of permutations. Must be non-negative.
+    /// The value of $k$. Must be non-negative.
```
istudyatuni: I think it's better to leave this as is, or at least combine both variants, because it's would be easier to understand purpose of the argument. If only "value of k" is shown, then you will need to scroll up, read description and search what is "k". Same applies to other similar changes

MDLC01: The current description is incorrect: $k$ is not the number of permutations (this is what the function returns) but the size of the permutations. I clarified the description.

crates/typst-library/src/foundations/calc.rs
```diff
@@ -344,24 +344,34 @@ pub fn atan(
...
 ) -> Angle {
     Angle::atan2(y.float(), x.float())
 }
 
 /// Calculates the hyperbolic sine of a hyperbolic angle.
 ///
+/// The hyperbolic sine of $x$ is defined as follows:
+/// $
+///   (e^x - e^(-x)) / 2 thin "."
+/// $
```
mkorje: Is there a reason you've used `thin` and the punctuation as strings throughout? I'd have thought just `.` and `\,` would be fine.

Though this has revealed a tiny bug that needs fixing, woops:
```html
<math display="block">
  <mfrac>
    ...
  </mfrac>
  <mspace width="0.1667em"></mspace>
  <mn>.</mn>
</math>
```

mkorje:
The (inline) CSS style sheet produced by the MathML export needs to be included in the doc's assets somehow. Without it, the cases aren't aligned properly (for example).

crates/typst-library/src/foundations/calc.rs
Enivex: Yeah I typically do not see spacing before punctuation in equations.

MDLC01: For the thin space, I took advice from [this StackExchange answer](https://tex.stackexchange.com/questions/7542/for-formal-articles-should-a-displayed-equation-be-followed-by-a-punctuation-to/170691#170691).

For the quotes, I used them because the punctuation is conceptually text and not math (see https://github.com/typst/typst/issues/1474).

Enivex: As pointed out by some of the comments, the thin space is not as common as they make it out to be. I've honestly never seen it.

laurmaedje:
> The (inline) CSS style sheet produced by the MathML export needs to be included in the doc's assets somehow. Without it, the cases aren't aligned properly (for example).

I think we probably want to automatically inject the style sheet into a user-generated head similar to what's being done in https://github.com/typst/typst/pull/8249.

MDLC01:
> The (inline) CSS style sheet produced by the MathML export needs to be included in the doc's assets somehow.

I'm not sure I understand exactly what this means. Does MathML export produce a stylesheet? Where can I find it, and how can I include it in the pages?

crates/typst-library/src/foundations/calc.rs
MDLC01: I removed the thin spaces before punctuation marks.

laurmaedje:
> I'm not sure I understand exactly what this means. Does MathML export produce a stylesheet? Where can I find it, and how can I include it in the pages?

MathML export produces an inline `<style>` sheet, but not if a custom `<head>` is provided (which the docs do). I would argue that it should always insert into the custom head like #8249 will. If we agree on that, fixing this would be out of scope for this PR and should be done separately.

crates/typst-library/src/foundations/calc.rs
mkorje: Using a string is more semantically correct, but it still _feels_ weird 😅. Once the bug is fixed, it should end up in an `mtext`. I tried the MathCAT screen reader online version thing (but only the math element, not the text before it - need to figure out how to do this), and whether its in an mn, mo, or mtext, it just ignores it.

MDLC01: > Using a string is more semantically correct, but it still _feels_ weird 😅.

Hence my proposal at https://github.com/typst/typst/issues/1474 (in situations where there is only a trailing punctuation mark, which is not the case with `cases`).

laurmaedje:
Regarding the whole punctuation debate: Since all these cases are preceded by a sentence ending with a colon, I would be inclined to fully remove it. It's not that the equations are in the middle of otherwise unterminated sentences.

MDLC01:
I removed all punctuation marks

laurmaedje:
Thanks!

PR #8259, opened 2026-05-12
Include file path in error message when loading a binary file fails

### LINKED ISSUE #8135: Include the file path in the error message “the PDF could not be loaded”
### Description

I suggest we include the file path in the error message “the PDF could not be loaded”.

Typst v0.14.2:

```log
error: the PDF could not be loaded
  ┌─ a.typ:2:1
  │
2 │ #image(file)
  │  ^^^^^^^^^^^
  │
  = hint: perhaps the PDF file is malformed
```

Proposed:

```log
error: the PDF could not be loaded (path/to/the.pdf)
  ┌─ a.typ:2:1
  │
2 │ #image(file)
  │  ^^^^^^^^^^^
  │
  = hint: perhaps the PDF file is malformed
```

### Use Case

If the PDF images are inserted programatically (e.g., in a loop), then it would help determine which one of the PDFs is malformed.

A similar error message already includes the path:

```log
error: file not found (searched at path/to/the.pdf)
```

---

Tip: You can create an empty file (`touch a.pdf`) to create a malformed PDF.

By the way, the _malformed_ PDF in my case wasn't actually malformed. I started `typst watch` and exported the PDFs with matplotlib. I guess the operating system (Windows) led to a certain race condition, triggering this issue. However, I can't reproduce it, so I'm not expecting it to be fixed.

Xiretza:
Fixes #8135.


tests/suite/loading/invalid.cbor
```diff
@@ -0,0 +1 @@
+B
```
Xiretza: Not sure how happy you are with binary-ish test files in your repository (see the xz debacle); in this case I simply put a single `0x42` byte in a file as a sort of "nothing up my sleeve" sign, but I'm open to better ideas.

istudyatuni: You can name it "invalid-cbor.txt"

saecki: > Not sure how happy you are with binary-ish test files in your repository (see the xz debacle); in this case I simply put a single `0x42` byte in a file as a sort of "nothing up my sleeve" sign, but I'm open to better ideas.

We have other binary test files in the `typst-dev-assets` repository. But since we only need any invalid CBOR file, maybe you could just load a file that's in another format?

crates/typst-library/src/loading/cbor.rs
```diff
@@ -77,8 +77,8 @@ pub fn cbor(
 ) -> SourceResult<Value> {
     let loaded = source.load(engine.world)?;
     ciborium::from_reader(loaded.data.as_slice())
-        .map_err(|err| eco_format!("failed to parse CBOR ({err})"))
-        .at(source.span)
+        .map_err(|err| LoadError::new_in_binary("failed to parse CBOR", err))
```
saecki: It's a bit unfortunate that the error forwards its `Display` impl to `Debug`, but since it's currently only 4 variants, I think we could match on those and make the error message a little nicer. Similar to whats done in `format_xml_like_error`

crates/typst-library/src/diag.rs
```diff
@@ -710,23 +710,36 @@ pub type LoadResult<T> = Result<T, LoadError>;
...
     message: EcoString,
 }
 
 impl LoadError {
-    /// Creates a new error from a position in a file, a base message
+    /// Creates a new error from a position in a text file, a base message
     /// (e.g. `failed to parse JSON`) and a concrete error (e.g. `invalid
     /// number`)
-    pub fn new(
+    pub fn new_in_text(
```
saecki: I think I'd just call the two constructors `LoadError::text` and `LoadError::binary`

crates/typst-library/src/diag.rs
```diff
@@ -762,7 +775,11 @@ where
 
     fn within(self, loaded: &Loaded) -> Self::Output {
         let LoadError { pos, message } = self.into();
-        load_err_in_text(loaded, pos, message)
+        if let Some(pos) = pos {
+            load_err_in_text(loaded, pos, message)
+        } else {
+            load_err_in_binary(loaded, None::<ReportPos>, message)
```
saecki: I think you can replace the `impl Into<ReportPos>` with `ReportPos` in both `load_in_text` and `load_in_binary`, then this can just be `None`

crates/typst-library/src/visualize/image/mod.rs
```diff
@@ -266,17 +266,23 @@ impl Packed<ImageElem> {
                         // TODO: the `DecyptionError` is currently not public
                         LoadPdfError::Decryption(_) => {
                             bail!(
```
saecki, line 268: I think `return Err(...)` instead of `bail!` would be cleaner

crates/typst-library/src/loading/cbor.rs
```diff
@@ -1,7 +1,7 @@
 use ecow::eco_format;
 use typst_syntax::Spanned;
 
-use crate::diag::{At, SourceResult};
+use crate::diag::{At, LoadError, LoadedWithin as _, SourceResult};
```
saecki: This isn't a common name so the `as _` isn't really needed

saecki (commented):
I know this was still marked as a draft, but after I commented the curiosity got the best of me. Here are a couple of remarks :)

crates/typst-library/src/visualize/image/mod.rs
Xiretza: I actually switched from that to `bail!()` because it's not simply a `return Err(x)`; but rather a `return Err(eco_vec![x]);` - I found it rather ugly, but I can switch back if desired.

saecki: Right, didn't think of that. Then let's keep it this way 👍

crates/typst-library/src/diag.rs
```diff
@@ -999,7 +1019,19 @@ pub fn format_xml_like_error(format: &str, error: roxmltree::Error) -> LoadError
         err => eco_format!("failed to parse {format} ({err})"),
     };
 
-    LoadError { pos: pos.into(), message }
+    LoadError { pos: Some(pos.into()), message }
+}
+
+/// Format a user-facing error encountered while parsing a CBOR file ([`ciborium::de::Error`]'s
+/// [`Display`] implementation just forwards to [`Debug`]).
+pub fn format_cbor_error(error: ciborium::de::Error<std::io::Error>) -> LoadError {
```
saecki: The `format_xml_like_error` only lives here, because it's used by both XML and SVG. I'd move this into the `cbor` module.

tests/suite/loading/invalid_cbor.txt
saecki, line 1: We still don't add assets for the test suite in this repository directly. So I'd either create an accompanying PR in the `typst-dev-assets` repository, or just load an arbitrary other file that will cause a cbor parse error.

crates/typst-library/src/diag.rs
```diff
@@ -999,7 +1019,19 @@ pub fn format_xml_like_error(format: &str, error: roxmltree::Error) -> LoadError
...
+    let err = match error {
+        ciborium::de::Error::Io(e) => eco_format!("IO error ({e})"),
```
saecki: The nested parentheses feel a little weird, what do you think about using a colon instead  `"IO error: ..."`?

crates/typst-library/src/diag.rs
```diff
@@ -999,7 +1019,19 @@ pub fn format_xml_like_error(format: &str, error: roxmltree::Error) -> LoadError
...
     };
 
-    LoadError { pos: pos.into(), message }
+    LoadError { pos: Some(pos.into()), message }
+}
+
+/// Format a user-facing error encountered while parsing a CBOR file ([`ciborium::de::Error`]'s
+/// [`Display`] implementation just forwards to [`Debug`]).
+pub fn format_cbor_error(error: ciborium::de::Error<std::io::Error>) -> LoadError {
+    let err = match error {
```
saecki: We can avoid allocating the intermediate `EcoString` by using `typst_utils::display`:
```
    let err = typst_utils::display(|f| match error {
        Error::Io(e) => write!(f, "IO error ({e})"),
        Error::Syntax(_) => f.write_str("syntax error"),
        Error::Semantic(_, s) => f.write_str(&s),
        Error::RecursionLimitExceeded => f.write_str("recursion limit exceeded"),
    });
```
When inside the `cbor` module, I think it's ok to import the `Error` type.

crates/typst-library/src/diag.rs
```diff
@@ -710,23 +710,36 @@ pub type LoadResult<T> = Result<T, LoadError>;
 /// [`FileId`]: typst_syntax::FileId
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct LoadError {
-    /// The position in the file at which the error occurred.
-    pos: ReportPos,
+    /// The position in the file at which the error occurred, or `None` for binary sources.
+    pos: Option<ReportPos>,
```
saecki: I think we could make it a little more obvious that this position is only suited for text files.
What do you think of renaming the type to `TextPos` and the field as well to `text_pos`?

saecki (commented):
Thanks! I've left a few nitpicky comments, but otherwise I think this is looking great :)

crates/typst-library/src/diag.rs
Xiretza: I was on the fence about that, will move.

tests/suite/loading/invalid_cbor.txt
Xiretza: I opened typst/typst-dev-assets#26. Just reusing another file would probably cause someone grief sooner or later if they have to update it for unrelated reasons.

crates/typst-library/src/diag.rs
Xiretza: Since it's only to be used for diagnostics rather than a generic span, I used `ReportTextPos`, does that sound good?

crates/typst-library/src/loading/cbor.rs
```diff
@@ -77,8 +78,22 @@ pub fn cbor(
...
     ciborium::from_reader(loaded.data.as_slice())
-        .map_err(|err| eco_format!("failed to parse CBOR ({err})"))
-        .at(source.span)
+        .map_err(format_cbor_error)
+        .within(&loaded)
+}
+
+/// Format a user-facing error encountered while parsing a CBOR file ([`ciborium::de::Error`]'s
+/// [`Display`] implementation just forwards to [`Debug`]).
+pub fn format_cbor_error(error: Error<std::io::Error>) -> LoadError {
```
saecki: The CI seems to fail, because `Display` isn't in scope. I don't think we need the doc links here, so removing the square brackets is fine.
Also while we're at it, can you make the function private?

saecki (commented):
Alright, one final nitpick from me :)

crates/typst-library/src/loading/cbor.rs
Xiretza: I'd actually already committed that, but I hadn't noticed the push had failed :roll_eyes:

Fixed!

saecki (approved):
Thanks! From my side this is only blocked on https://github.com/typst/typst-dev-assets/pull/26

laurmaedje:
Thanks!

PR #7311, opened 2025-11-05
Mention `SOURCE_DATE_EPOCH` in docs for `today`

pavelzw:
#3809 doesn't have any documentation yet.

PR #8277, opened 2026-05-13
Fix URI inside documentation for creation-timestamp

claudydoba:
URI `https://reproducible-builds.org/specs/...` changed to use `docs` instead of `specs`.

Related:
- PR (https://github.com/typst/typst/pull/7311)
- Commit 7ed54a5

### DIFF
--- a/crates/typst-cli/src/args.rs
+++ b/crates/typst-cli/src/args.rs
@@ -412,7 +412,7 @@ pub struct WorldArgs {
 
     /// The document's creation date formatted as a UNIX timestamp.
     ///
-    /// For more information, see <https://reproducible-builds.org/specs/source-date-epoch/>.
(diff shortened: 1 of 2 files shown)


laurmaedje:
https://reproducible-builds.org/specs/source-date-epoch/ exists and contains the specification for `SOURCE_DATE_EPOCH`. Meanwhile, https://reproducible-builds.org/docs/source-date-epoch/ links to it:

> However, if you find that it’s ideal for your use-case, please feel free to jump straight to our [published specification](https://reproducible-builds.org/specs/source-date-epoch/).

The spec seems like a more canonical source, so I would leave this as is.

claudydoba:
Oh, it's true. When I checked the site, it was inaccessible, maybe it was down or something. I agree the spec seems more appropriate in this case.

I wrongly thought it had changed. I hadn't seen the spec site before, so I assumed it had been moved.

Thank you for reviewing it.

PR #8278, opened 2026-05-13
Add copy button to code examples in documentation

MDLC01:
This closes https://github.com/typst/typst/issues/8264 by adding a Copy button to all code examples (only those that have a preview).

The Copy button is added within the JavaScript script. Alternatively, it could be added in the Typst template directly, but this seemed less easy, and also the current solution has the advantage of not showing a Copy button to users that have JavaScript disabled, or when the page has not fully loaded yet.

Current result:

A side-by-side example with code and preview has a small copy button at the top right of the code.

Feel free to suggest changes to the styling.

### DIFF
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -301,15 +301,45 @@ sup[role="doc-noteref"] {
 
 /* Code example components */
 
-pre > button.copy {
-  top: -8px;
-  right: -8px;
-  padding: 2px;
-  background: none;
-  border: 0px;
+pre {
   position: relative;
+}
+
+pre > button.copy {
+  top: 8px;
+  right: 8px;
+  padding: 8px;
+  background: #fdfdfd;
+  border: 1px solid var(--border-secondary);
+  color: #565565; /* Should be var(--text-secondary) */
+  position: absolute;
   float: right;
   cursor: pointer;
+  opacity: 0;
+  transition: opacity 0.2s;
+}
+
+pre:hover > button.copy,
+pre > button.copy:active,
+pre > button.copy:focus-visible {
+  opacity: 1;
+}
+
+pre > button.copy:hover {
+  background: var(--action-secondary-hover);
+}
+
+pre > button.copy:focus-visible {
+  background: #fdfdfd;
+  transition: opacity 0s;
+}
+
+pre > button.copy:disabled {
+  color: rgba(25, 24, 31, 0.5) /* Should be var(--text-disabled) */
+}
+
+pre > button.copy > svg {
+  display: block;
 }
 
 .code-definition + h4,
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -311,7 +311,7 @@ pre > button.copy {
   padding: 8px;
   background: #fdfdfd;
   border: 1px solid var(--border-secondary);
-  color: #565565; /* Should be var(--text-secondary) */
+  color: var(--text-primary);
   position: absolute;
   float: right;
   cursor: pointer;
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -319,7 +319,6 @@ pre > button.copy {
   transition: opacity 0.2s;
 }
 
-pre:hover > button.copy,
 pre > button.copy:active,
 pre > button.copy:focus-visible {
   opacity: 1;
@@ -336,6 +335,18 @@ pre > button.copy:focus-visible {
 
 pre > button.copy:disabled {
   color: rgba(25, 24, 31, 0.5) /* Should be var(--text-disabled) */
+}
+
+@media (hover: hover) {
+  pre:hover > button.copy {
+    opacity: 1;
+  }
+}
+
+@media (hover: none) {
+  pre.tapped > button.copy {
+    opacity: 1;
+  }
 }
 
 pre > button.copy > svg {
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -316,11 +316,11 @@ pre > button.copy {
   float: right;
   cursor: pointer;
   opacity: 0;
-  transition: opacity 0.2s;
+  transition: opacity 0.2s ease-in-out;
 }
 
 pre > button.copy:active,
-pre > button.copy:focus-visible {
+pre > button.copy:focus {
   opacity: 1;
 }
 
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -334,7 +334,7 @@ pre > button.copy:focus-visible {
 }
 
 pre > button.copy:disabled {
-  color: rgba(25, 24, 31, 0.5) /* Should be var(--text-disabled) */
+  color: var(--text-disabled);
 }
 
 @media (hover: hover) {
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -301,6 +301,17 @@ sup[role="doc-noteref"] {
 
 /* Code example components */
 
+pre > button.copy {
+  top: -8px;
+  right: -8px;
+  padding: 2px;
+  background: none;
+  border: 0px;
+  position: relative;
+  float: right;
+  cursor: pointer;
+}
+
 .code-definition + h4,
 .previewed-code + h4 {
   margin-block-start: 28px;
--- a/docs/assets/docs.js
+++ b/docs/assets/docs.js
@@ -428,18 +428,13 @@ function setUpPreviewSplits() {
  * Sets up the Copy button on example codes.
  */
 function setUpPreviewCopy() {
-  for (const pre of document.querySelectorAll(".previewed-code > pre")) {
+  for (const button of document.querySelectorAll("pre > button.copy")) {
+    const pre = button.parentElement;
     const code = pre.innerText;
-    const button = document.createElement("button");
-    button.classList.add("copy");
-    const img = document.createElement("img");
-    img.src = copyIconSrc;
-    img.alt = "Copy";
-    button.appendChild(img);
     button.addEventListener("click", () => {
       copyText(code);
     });
-    pre.prepend(button);
+    button.disabled = false;
   }
 }
 
--- a/docs/assets/docs.js
+++ b/docs/assets/docs.js
@@ -431,6 +431,14 @@ function setUpPreviewCopy() {
   for (const button of document.querySelectorAll("pre > button.copy")) {
     const pre = button.parentElement;
     const code = pre.innerText;
+    // Display the Copy button for 30s when the `<pre>` is tapped on touch
+    // screens.
+    let timeoutId;
+    pre.addEventListener("click", () => {
+      pre.classList.add("tapped");
+      clearTimeout(timeoutId);
+      timeoutId = setTimeout(() => pre.classList.remove("tapped"), 30_000);
+    });
     button.addEventListener("click", () => {
       copyText(code);
     });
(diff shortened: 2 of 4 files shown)


docs/assets/docs.css
```diff
@@ -279,6 +279,17 @@ sup[role="doc-noteref"] {
 
 /* Code example components */
 
+pre > button.copy {
+  top: -8px;
+  right: -8px;
+  padding: 2px;
```
reknih: We generally have a padding of 8px around icon buttons for a total touch target of 32x32px

docs/assets/docs.js
```diff
@@ -422,6 +424,25 @@ function setUpPreviewSplits() {
...
+ * Sets up the Copy button on example codes.
+ */
+function setUpPreviewCopy() {
+  for (const pre of document.querySelectorAll(".previewed-code > pre")) {
+    const code = pre.innerText;
+    const button = document.createElement("button");
+    button.classList.add("copy");
+    const img = document.createElement("img");
+    img.src = copyIconSrc;
+    img.alt = "Copy";
```
reknih: Instead of constructing a button on hover, I'd put it in the HTML source. If you'd like, you can give it the `disabled` attribute that this script then removes once it mounts the handlers

docs/assets/docs.js
```diff
@@ -422,6 +424,25 @@ function setUpPreviewSplits() {
...
 
+/**
+ * Sets up the Copy button on example codes.
+ */
+function setUpPreviewCopy() {
+  for (const pre of document.querySelectorAll(".previewed-code > pre")) {
+    const code = pre.innerText;
+    const button = document.createElement("button");
+    button.classList.add("copy");
+    const img = document.createElement("img");
```
reknih: Constructing the graphic as an `img` element instead of an `svg` will render all `currentColor` instances in the image to black. When constructing an SVG using a single `use` element referencing the file path within, the current text color will be used instead. `use-icon` is the right function to produce this markup.

However, that's also not what I'm looking for: The icon should have the color `#565565`. In the web app, this is the design token `var(--text-secondary)`. In `docs/assets/base.css`, the value is not defined (even though it should be, but in another PR) so the literal should be used.

docs/assets/docs.js
reknih, line 1: The icon being visible all the time adds to much visual clutter I think. Instead, it should behave like this:

- It should be invisible by default
- It should be visible on hover of the code example. When displaying, it should fade its opacity in .2s or less using an appropriate easing function. This should be accomplished using CSS. When a button is active, it is not hidden, even if the pointer left the code example element while holding down the button.
- It should become visible when receiving keyboard focus and use the focus ring from the design system. The focus ring should not be visible if the focus was not created by keyboard navigation (`:focus-visible`). When a focus ring is visible, the element gains the background `var(--layer-1)` (also not in the design token list; `#fdfdfd`). When the element receives keyboard focus, no transition shall be played.

  This is the focus ring from the design system where var(--border-focus) is `rgba(0,122,255,0.500)`:

  ```sass
  @mixin effect-focus-raw
    outline: 3px solid var(--border-focus)
    outline-offset: 0px
  
  @mixin effect-focus
    &:focus
      @include effect-focus-raw
    // As of writing Safari doesn't support :focus-visible yet,
    // and just ignores all rules that contain :focus-visible
    &:focus-visible, &:focus
      outline: none
    &:focus-visible
      @include effect-focus-raw
  ```
- On devices without hover (`@media (hover: none)`), the button should be displayed after the user tapped on the code example. This should result in the same animation as for hover. I think that the button should stay visible even after the example is tapped again but fade out after about 30s. To detect this tap and remember it, some JavaScript is required. You should listen to the `MediaQueryList` returned by `matchMedia` to attach and remove this behavior so that the website correctly incorporates events like attaching and detaching a Magic Keyboard from an iPad. Gaining the hover capability should immediately hide all buttons.

docs/assets/docs.css
```diff
@@ -279,6 +279,17 @@ sup[role="doc-noteref"] {
...
+  background: none;
+  border: 0px;
+  position: relative;
+  float: right;
+  cursor: pointer;
+}
```
reknih, line 354:
```suggestion
  color: #565565; /* Should be var(--text-secondary) */
  border-radius: 6px;
}
```

docs/assets/docs.css
```diff
@@ -279,6 +279,17 @@ sup[role="doc-noteref"] {
...
+
 .code-definition + h4,
```
reknih, line 356: If you follow my suggestion for page load behavior:

```suggestion
pre > button.copy:disabled {
  color: rgba(25,24,31,0.500) /* Should be var(--text-disabled) */
}

.code-definition + h4,
```

reknih (commented):
In addition to my review comments: Are there any instances in the docs where the icon overlaps with text? If it does and if that looks bad, we may instead want to use a `GroupButton` or a `Pane` style.

**`GroupButton`:**

| Property | Variable | Value |
|:------|:-------|:------|
| Background | `var(--layer-1)` | `#fdfdfd` |
| Border | `1px solid var(--border-secondary)` | `#bdbfcc` |
| Total height |  | 32px |

**`Pane`:**

| Property | Variable | Value |
|:------|:-------|:------|
| Background | `var(--layer-2-80)` | `rgba(239,240,243,0.8)` |
|  `backdrop-filter` | | `blur(3px)` |
| Border | `1px solid var(--border-secondary)` | `#bdbfcc` |
| Total height |  | Flexible |
| Box shadow [^1] | `-2px -2px 8px rgba(var(--text-secondary-rgb), 0.1)` | `rgba(86, 85, 101, 0.1)`|

Both components should have 6px border radius.

Also verify the button has `background: var(--action-secondary-hover)` on hover and `background: var(--action-secondary-active)` (`#d7d9e0`) on active.

[^1]: only in light mode, but we only have light mode atm

MDLC01:
I tried to follow everything you said. Regarding the focus ring, I think it already works as intended as-is. About the `(hover: none)` support, I needed a little less JavaScript that you suggested so I suspect I may have missed something.

I found a place where the button overflows with the code in the first example of "A toy template" in the "Making a template" tutorial. I opted for the `GroupButton` style.

I had to modify the Copy icon to make it visible (https://github.com/typst/typst-dev-assets/pull/25). I'm not sure why it still doesn't look right though.

The Copy icon, with the inside filled with plain black.

saecki:
Yeah, I remember I also had an issue with the icon in the test report. I fixed it, but don't remember how exactly, you can probably just copy the path from here: https://github.com/typst/typst/blob/de6f400976f9bf6ab8b923d13a068722959d0070/tests/src/report/html.rs#L1360

docs/assets/docs.css
```diff
@@ -279,6 +279,58 @@ sup[role="doc-noteref"] {
...
+  right: 8px;
+  padding: 8px;
+  background: #fdfdfd;
+  border: 1px solid var(--border-secondary);
+  color: #565565; /* Should be var(--text-secondary) */
+  position: absolute;
+  float: right;
+  cursor: pointer;
+  opacity: 0;
+  transition: opacity 0.2s;
```
reknih: Add `ease-in-out` or a custom easing function here (Firefox has a visual editor)

docs/assets/docs.css
```diff
@@ -279,6 +279,58 @@ sup[role="doc-noteref"] {
...
+}
+
+pre > button.copy:active,
+pre > button.copy:focus-visible {
```
reknih: I think this can be forced open on all focus states.

```suggestion
pre > button.copy:focus {
```

docs/assets/docs.css
```diff
@@ -279,6 +279,58 @@ sup[role="doc-noteref"] {
...
+  color: rgba(25, 24, 31, 0.5) /* Should be var(--text-disabled) */
+}
+
+@media (hover: hover) {
+  pre:hover > button.copy {
+    opacity: 1;
+  }
+}
+
+@media (hover: none) {
```
reknih, line 346: Smart to do this instead of the `matchMedia`

docs/assets/docs.css
saecki: I think this would mean clicking it would force it to stay visible, because it has an invisible focus. I think I did the same in the test report

docs/assets/docs.css
```diff
@@ -279,6 +279,58 @@ sup[role="doc-noteref"] {
...
+  position: relative;
+}
+
+pre > button.copy {
+  top: 8px;
+  right: 8px;
+  padding: 8px;
+  background: #fdfdfd;
+  border: 1px solid var(--border-secondary);
+  color: #565565; /* Should be var(--text-secondary) */
```
reknih:
```suggestion
  color: var(--text-primary);
```

In the component, text-primary looks better.

reknih (commented):
Otherwise ready for rebase & merge

MDLC01:
https://github.com/typst/typst-dev-assets/pull/25 needs to be merged before this PR

laurmaedje:
> https://github.com/typst/typst-dev-assets/pull/25 needs to be merged before this PR

Done

MDLC01:
This should now be ready to merge.

This was my first pull request completed with Jujutsu so I hope I did not make any mistake. I double checked and I think everything is right.

docs/assets/docs.css
```diff
@@ -287,6 +287,58 @@ sup[role="doc-noteref"] {
...
+  background: var(--action-secondary-hover);
+}
+
+pre > button.copy:focus-visible {
+  background: #fdfdfd;
+  transition: opacity 0s;
+}
+
+pre > button.copy:disabled {
+  color: rgba(25, 24, 31, 0.5) /* Should be var(--text-disabled) */
```
laurmaedje: Could you turn this into a `var` and define the appropriate variable?

docs/assets/docs.css
laurmaedje: Is this still relevant?

docs/components/example.typ
```diff
@@ -159,6 +159,22 @@
...
+      }
+      let copy-button = html.button(
+        class: "copy",
+        disabled: true,
+        use-icon(16, "copy", "Copy"),
+      )
+      let reconstructed = html.elem(
+        "pre",
+        attrs: pre.attrs,
+        copy-button + pre.body
```
laurmaedje:
```suggestion
        copy-button + pre.body,
```

laurmaedje (commented):
It seems like the copy button only works on top-level examples, but not on parameter examples. It just doesn't copy the text for me.

docs/assets/docs.css
MDLC01: I think this was changed in https://github.com/typst/typst/pull/8278#discussion_r3298383131.

MDLC01:
Regarding the issue with copying parameter examples, replacing `copyText(code);` with `copyText(pre.innerText);` solves the issue. The reason is that apparently, hidden elements (including the `<pre>` that are in hidden `<detail>` elements) have no `innerText`.

~~I vaguely remember not using `pre.innerText` in the callback intentionally, but I can't remember why. Testing it again, I can't find find any issue with it. So I think we can keep it like that for now.~~ I think I remember now: this is a residue from when the Copy button was constructed in JavaScript, and due to the way it was implemented, the button was adding additional `innerText`, so I had to get the `innerText` before adding the button. This is no longer an issue as the button adds no additional `innerText`.

laurmaedje:
Thank you!

PR #8285, opened 2026-05-13
Replace `ArcExt::take` with `Arc::unwrap_or_clone` from Rust 1.76

isuffix:
This replaces a 4-year-old trait extenstion that was originally added in https://github.com/typst/typst/commit/5becb32ba463d6b0ace914ab06bb237483a94fbc with the equivalent standard library method. `Arc::unwrap_or_clone` was stabilized in Rust 1.76, although I do admit to disliking the new name.


laurmaedje:
Thanks!

PR #8287, opened 2026-05-14
Fix non-ASCII paths for live reload server

### LINKED ISSUE #8084: Files with non-ASCII characters are not served properly by the `typst watch` server in bundle export
### Description

Document:

```typ
// main.typ
#document("é.html")[Hello]
```

Compile command:

```sh
typst watch main.typ --format=bundle --features=html,bundle
```

Now, try to open http://127.0.0.1:xxxx/é.html in your browser (making sure the port is the one provided by the watcher). I get an error with both Firefox and Chrome.

### Operating system

Windows

### Typst version

andylizi:
Fix #8084 by percent-decoding the paths before routing.

No new dependency is added since `percent-encoding` is already in use by `url`.

Tested locally using the example in the issue.

### DIFF
--- a/crates/typst-kit/src/server.rs
+++ b/crates/typst-kit/src/server.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 
 use ecow::eco_format;
 use parking_lot::{Condvar, Mutex, MutexGuard};
+use percent_encoding::percent_decode_str;
 use tiny_http::{Header, Request, Response, StatusCode};
 use typst_library::diag::{StrResult, bail};
 use typst_library::foundations::Bytes;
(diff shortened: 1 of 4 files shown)


crates/typst-kit/src/server.rs
```diff
@@ -147,12 +148,16 @@ fn handle(req: Request, reload: bool, bucket: &Arc<RouterBucket>) -> io::Result<
     };
 
     let path = url.path();
+    let Ok(path) = percent_decode_str(path).decode_utf8() else {
```
laurmaedje, line 151: Technically, I think this could conflate `%2F` with a literal slash `/`, such that `a%2Fb` resolves to `a/b.html`. I'm not sure whether this is correct given https://datatracker.ietf.org/doc/html/rfc3986, but I tried with a couple of big web sites and half the internet treats it the same, so I suppose it's fine.

laurmaedje:
Thanks!

PR #8288, opened 2026-05-14
(Feature) Decimal literals

### LINKED ISSUE #8093: `decimal` Literals
### Description

It would be helpful to have a more concise way to specify numbers of type `decimal`. JavaScript solved a similar issue with [BigInts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) by using a postfix.
```
decimal("0.5") // current
0.5d           // proposed
```

### Use Case

We are currently not using decimals in some places it would make sense to, simply because `decimal("0.5")` is too long and would visually polute the codebase.

Lege19:
Fix #8093

With this change `1.2345d` is equivalent to the longer `decimal("1.2345")`

I also updated the diagnostic previously emitted on `decimal(1.2345)` to suggest `1.2345d` instead of `decimal("1.2345")`, and updated the documentation to recommend using the new literal syntax.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3389,6 +3389,7 @@ name = "typst-syntax"
 version = "0.14.2"
 dependencies = [
  "ecow",
+ "rust_decimal",
  "rustc-hash",
  "serde",
  "toml",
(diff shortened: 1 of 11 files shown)


Lege19:
I can't find how to draft this again on mobile. I've realised there's probably a bug which I'll fix in the morning.

PR #8289, opened 2026-05-15
Fixed single decimal outputting `<mn>` in MathML

### LINKED ISSUE #8261: String `"."` in math becomes an `mn` in MathML export
### Description

From #8258, the Typst code
```typ
$ (e^x - e^(-x)) / 2 thin "." $
```
produces
```html
<math display="block">
  <mfrac>
    ...
  </mfrac>
  <mspace width="0.1667em"></mspace>
  <mn>.</mn>
</math>
```

### Operating system

_No response_

### Typst version

TheMagisterial:
This PR fixes #8261 

I added two checks:

+ During the check of each character, the number of decimal occurences are counted. If the count of decimals is anything other than zero or one, the "number" check fails and an `mtext` is returned.
+ After iterating on each character, the text length is compared against the decimal. If they are equal that means either (1) there are more than one decimals or (2) the text length is one and the only character is a decimal.

Sample inputs:

```
$"1.2.3"$ produces <mtext>
$"1..3"$ produces <mtext>
$"1.3"$ produces <mn>
$"13"$ produces <mn>
$"."$ produces <mtext>
```


Enivex:
Are the following handled correctly currently?
- Empty strings
- Ellipsis
- Period used as punctuation
- Dots used as thousands separators in some languages

Possibly out of scope for this pr, but I wanted to put it out there

TheMagisterial:
> Are the following handled correctly currently?
> 
> * Empty strings
> * Ellipsis
> * Period used as punctuation
> * Dots used as thousands separators in some languages
> 
> Possibly out of scope for this pr, but I wanted to put it out there

@Enivex all great thoughts and I 100% agree on all points. Below is what is addressed with this PR:

- [x] Empty Strings: ~~now empty strings are returned as `<mtext>` elements instead of `<mn>`~~ return no new MathML element.
- [x] Ellipsis: in testing, the ellipsis produces `<mtext>` now instead of `<mn>` as well
- [x] Period used as punctuation is innately handled correctly as the presence of non-numeric characters short-circuits to being an `<mtext>`
- [ ] Dots as thousands separators...: This is a great candidate for a new issue as the current implementation does not factor in any i18n at present. I am not skilled enough with this repository yet to make the judgement call about exactly *where* that checking should live. I would like to seek formal input from seasoned contributors on this one, but regardless I believe a new issue/PR would be best for that.

In the meantime, I am working on a another commit to address the failing tests (I did not understand how the test framework worked until tonight after playing around with it some more. Now that I understand how it works I will be addressing that).

TheMagisterial:
Following up on tests:

I added/updated the following tests and checked them locally against commit de6f400 and my changes:

```typ
--- math-string-as-empty html ---
// Testing that empty strings produce no element in MathML
$ "" $

--- math-string-as-ellipsis html ---
// Testing that ellipses produce `<mtext>` in MathML, not `<mn>`
$ "..." $

--- math-string-as-non-numeric html ---
// Testing that various "bogus" strings produce `<mtext>` in MathML, not `<mn>`
$ "1..1" ".1.1" "1.1." "1..1..1" "1.1.1" ".." $

--- math-multiline-string paged html ---
$ "a\nb\n\n\n" "a\nb\n\n" "a\nb\n" "a\nb" $

--- math-multiline-string-alignment paged html ---
$ "L\n\n""L\n""L"  #$
```


```diff
diff --git "a/.\\tester-de6f400.html" "b/.\\tester-pr-8261.html"
index 159e09ed5..1f0bd7ea8 100644
--- "a/.\\tester-de6f400.html"
+++ "b/.\\tester-pr-8261.html"
@@ -86,16 +86,21 @@
     </math>
     <p>-- math-string-as-ellipsis html --</p>
     <math display="block">
-      <mn>...</mn>
+      <mtext>...</mtext>
     </math>
     <p>-- math-string-as-non-numeric html --</p>
     <math display="block">
-      <mn>1..1</mn>
-      <mn>.1.1</mn>
-      <mn>1.1.</mn>
-      <mn>1..1..1</mn>
-      <mn>1.1.1</mn>
-      <mn>..</mn>
+      <mtext>1..1</mtext>
+      <mspace width="0.2222em"/>
+      <mtext>.1.1</mtext>
+      <mspace width="0.2222em"/>
+      <mtext>1.1.</mtext>
+      <mspace width="0.2222em"/>
+      <mtext>1..1..1</mtext>
+      <mspace width="0.2222em"/>
+      <mtext>1.1.1</mtext>
+      <mspace width="0.2222em"/>
+      <mtext>..</mtext>
     </math>
     <p>-- math-multiline-string paged html --</p>
     <math display="block">
@@ -112,12 +117,12 @@
         </mtr>
         <mtr>
           <mtd>
-            <mn></mn>
+            <mtext></mtext>
           </mtd>
         </mtr>
         <mtr>
           <mtd>
-            <mn></mn>
+            <mtext></mtext>
           </mtd>
         </mtr>
       </mtable>
@@ -134,7 +139,7 @@
         </mtr>
         <mtr>
           <mtd>
-            <mn></mn>
+            <mtext></mtext>
           </mtd>
         </mtr>
       </mtable>
@@ -173,7 +178,7 @@
         </mtr>
         <mtr>
           <mtd>
-            <mn></mn>
+            <mtext></mtext>
           </mtd>
         </mtr>
       </mtable>

```

I will commit these test updates shortly but wanted to at least throw them here for documentation.

Enivex:
Regarding ellipsis, I was thinking of numbers like `$pi = 3.14...$`

TheMagisterial:
Ah, my misunderstanding. 

I actually think this poses a larger investigation; evidently, plain-written values count as numerics (e.g., "twelve" counts as an `<mn>`). But should these rules apply to quoted strings? To me, I feel that any text inside double quotes should just be interpreted as `<mtext>` instead of trying to determine if it is numeric. Automated determination of `<mtext>` vs. `<mn>`, in my opinion, should only apply to content outside strings within Math blocks.

I need to do some more reading (outside the scope of this PR) about the spec as it relates to Typst's strings inside math. It appears `<mtext>` should be used as a container for arbitrary text, so should we be considering ALL content inside of double-quotes as `<mtext>` by default?

@Enivex @mkorje thoughts on this?

crates/typst-library/src/math/ir/resolve.rs
```diff
@@ -282,7 +282,12 @@ fn resolve_text<'a, 'v, 'e>(
     // Create item with correct styles and properties.
     let local_styles = ctx.store_chain(styles).chain(&*TEXT_BASE_LOCAL_STYLES);
     let mut create_item = |text: &str| {
-        let num = text.chars().all(|c| c.is_ascii_digit() || c == '.');
+        let mut decimal_count = 0;
+        let num = text.chars().all(|c| {
+            decimal_count += if c == '.' { 1 } else { 0 };
```
mkorje: I think this is a bit cleaner
```rs
if c == '.' {
    decimal_count += 1;
}
```

mkorje:
> * Dots as thousands separators...: This is a great candidate for a new issue as the current implementation does not factor in any i18n at present. I am not skilled enough with this repository yet to make the judgement call about exactly _where_ that checking should live. I would like to seek formal input from seasoned contributors on this one, but regardless I believe a new issue/PR would be best for that.

> I actually think this poses a larger investigation; evidently, plain-written values count as numerics (e.g., "twelve" counts as an `<mn>`). But should these rules apply to quoted strings? To me, I feel that any text inside double quotes should just be interpreted as `<mtext>` instead of trying to determine if it is numeric. Automated determination of `<mtext>` vs. `<mn>`, in my opinion, should only apply to content outside strings within Math blocks.

Regarding this, the intention is to add a `math.num` (or related) element to Typst precisely for this purpose. Numbers during parsing (like `10`) would get wrapped in this element, instead of text. And everything in quotes in math would become `mtext`. Because baking in all possible i18n at parsing isn't really feasible, users can then wrap numbers explicitly (e.g. `$ num("10,01") $`).

The addition of a `math.num` element is currently blocked on some complex issues (see https://github.com/typst/typst/pull/7577#issuecomment-3728016523). The change you've made suffices for the current state of things though, and is basically what I had in mind.

crates/typst-library/src/math/ir/resolve.rs
isuffix: Not that I prefer it, but I recently learned (from a pedantic [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html?search=Bool+int#bool_to_int_with_if)) that you could also write `decimal_count += (c == '.') as u64;`

I kind of hate it, but it is neat!

TheMagisterial: For cleanliness, I will be changing the lines to match the pattern @mkorje prefers, but yes @isuffix I LOVE esoteric one-liners like that. Definitely noted for future implementations.

mkorje: > Not that I prefer it, but I recently learned (from a pedantic [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html?search=Bool+int#bool_to_int_with_if)) that you could also write `decimal_count += (c == '.') as u64;`
> 
> I kind of hate it, but it is neat!

That is very cursed aha.

laurmaedje: I do this quite often actually, e.g. here:

https://github.com/typst/typst/blob/69ca2783dc0be23632427ede5a1b9e0857ade11d/crates/typst-library/src/introspection/introspector.rs#L252

laurmaedje:
Thanks!

